PR #2435, opened 2018-08-02
Swap cache-db-path and cache-chunk-path in cache.md

sparkyman215:
Have the db path option come first in doc, as the chunk path references db and isn't needed if the preceding (db path) command is used.


ncw:

PR #2436, opened 2018-08-02
webdav: sharepoint recursion with different depth

### LINKED ISSUE #2426: ls or copy file error for Onedrive Business(using webdav sharepoint)
@hensur , @ncw 

this issue related to:
onedrive: Onedrive Business Support for organizations in an unamanged state (mostly universities) #1975  [#1975](https://github.com/ncw/rclone/issues/1975)

This is a problem to ls or copy file in webdav (onedrive sharepoint) , for example:
When ls, copy or sync from a source directory, it work fine.
rclone ls od1:song
rclone copy od1:song/  ./mylocal

But when ls or copy from a source file,, it has error:
rclone  ls od1:song/my.mp3
rclone copy od1:song/my.mp3  ./mylocal
ERROR : : error reading source directory: directory not found

Using rclone --dump headers --log-level DEBUG ls od1:song/my.mp3 to debug:
DEBUG : HTTP REQUEST (req 0xc4203ea300)
DEBUG : PROPFIND /personal/xxx_xxx_edu/Documents/song/my.mp3 HTTP/1.1
Host: xxx-my.sharepoint.com
DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
DEBUG : HTTP RESPONSE (req 0xc4203ea300)
DEBUG : HTTP/1.1 404 NOT FOUND
 DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
ERROR : : error listing: directory not found
Failed to ls: directory not found

Using rclone --dump headers --log-level DEBUG ls od1:song/  to debug:
DEBUG : HTTP REQUEST (req 0xc420310e00)
DEBUG : PROPFIND /personal/xxx_xxx_edu/Documents/song/  HTTP/1.1
DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
DEBUG : HTTP RESPONSE (req 0xc420310e00)
DEBUG : HTTP/1.1 207 MULTI-STATUS
DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
  4750661 my.mp3
  3898990 test.mp3

hensur:
this change adds the depth parameter to listAll and readMetaDataForPath.
this allows recursive calls of these methods with a different depth
header.

Sharepoint won't list files if the depth header is != 0. If that is the
case, it will just return a error 404 although the file exists.
Since it is not possible to determine if a path should be a file or a
directory, rclone has to make a request with depth = 1 first. On success
we are sure that the path is a directory and the listing will work.
If this request returns error 404, the path either doesn't exist or it
is a file.
To be sure, we can try again with depth set to 0. If it still fails, the
path really doesn't exist, else we found our file.

fixes #2426

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -46,7 +46,8 @@ import (
 const (
 	minSleep      = 10 * time.Millisecond
 	maxSleep      = 2 * time.Second
-	decayConstant = 2 // bigger for slower decay, exponential
+	decayConstant = 2   // bigger for slower decay, exponential
+	defaultDepth  = "1" // depth for PROPFIND
 )
 
 // Register with Fs
@@ -103,17 +104,18 @@ type Options struct {
 
 // Fs represents a remote webdav
 type Fs struct {
-	name        string        // name of this remote
-	root        string        // the path we are working on
-	opt         Options       // parsed options
-	features    *fs.Features  // optional features
-	endpoint    *url.URL      // URL of the host
-	endpointURL string        // endpoint as a string
-	srv         *rest.Client  // the connection to the one drive server
-	pacer       *pacer.Pacer  // pacer for API calls
-	precision   time.Duration // mod time precision
-	canStream   bool          // set if can stream
-	useOCMtime  bool          // set if can use X-OC-Mtime
+	name               string        // name of this remote
+	root               string        // the path we are working on
+	opt                Options       // parsed options
+	features           *fs.Features  // optional features
+	endpoint           *url.URL      // URL of the host
+	endpointURL        string        // endpoint as a string
+	srv                *rest.Client  // the connection to the one drive server
+	pacer              *pacer.Pacer  // pacer for API calls
+	precision          time.Duration // mod time precision
+	canStream          bool          // set if can stream
+	useOCMtime         bool          // set if can use X-OC-Mtime
+	retryWithZeroDepth bool          // some vendors (sharepoint) won't list files when Depth is 1 (our default)
 }
 
 // Object describes a webdav object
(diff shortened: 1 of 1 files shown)


ncw:
I wonder if it might be better to set the depth as part of the sharepoint quirks:

https://github.com/ncw/rclone/blob/f9eb9c894de0840c25beaf8e288c7dfc31b72214/backend/webdav/webdav.go#L361-L370

So just set a depth amount which is 1 by default but 0 for sharepoint.

What do you think?

hensur:
I think that would break other default behavior, like the recursive ls.
I just hardcoded the `defaultDepth` to `"0"` and tried `lsd` and `ls`. Both returned an empty list.

With depth set to `"0"` sharepoint (and nextcloud) only return the requested directory (or file). The listing of the requested file or directory itself is ignored by rclone.
If a directory is requested with `depth = "1"` both return the directory and its files and subdirectories, as expected.
However, if a file is requested with `depth = "1"`, nextcloud will return its properties (it does the same with `depth = "0"`). Sharepoint on the other hand will return an error 404, not found.
Only if I set the depth to `"0"`, sharepoint will behave the same as nextcloud and return the file's properties.

To circumvent this "specialty" of sharepoint, I just call the list again if it has failed with error 404, but with the depth set to `"0"`.
I don't think there is any other way to determine if the requested path is supposed to be a file or a directory. If this second request also fails, there is probably no file or directory at the requested path.

ncw:
> I don't think there is any other way to determine if the requested path is supposed to be a file or a directory. If this second request also fails, there is probably no file or directory at the requested path.

Ah, I see - thanks for explaining. I should have read your comment more carefully as you explain it there.

I'll merge this now and we can give it whirl :-)

Issue #1603, opened 2017-08-15
Support OVH Public Cloud Archive


wlhlm:
In addition to their Swift-compatible cloud storage, OVH offers a [Cloud Archive](https://www.ovh.co.uk/public-cloud/storage/cloud-archive/) service, which "freezes" data for very low storage costs, but data can take multiple hours to be available for retrieval. Cloud Archive uses the same Swift API albeit with some modifications. For example, in order to create containers on Cloud Archive, an additional header has to be sent:
```
X-Storage-Policy: PCA
```
Otherwise, it defaults to normal object storage for the container.

This can be surprising when uploading files larger than 5GiB (by default), since this means the files get split into segments and rclone creates a new container ("\<containername\>"+"_segments") in which these segments are stored. Since rclone doesn't send the special header this newly created container uses the standard Swift object storage instead of the Cloud Archive service, which has higher storage costs. Luckily, inbound traffic for normal object storage is free, though it can get costly if you don't notice right away (since retention cost is significantly higher).

A workaround is to manually create the _segments container in the web interface where you can choose storage type and rclone will happily use that.

Another thing that's different is, since data takes hours to be available for retrieval, trying to read cold data returns status code `429 Too Many Requests`. Upon the first request, a retrieval job is initiated and there have been addition to the swift API to query when the objects will be available.

- [Cloud Archive developer guide](https://docs.ovh.com/gb/en/cloud/storage/pca/dev/)
- [API perculiarities](https://docs.ovh.com/gb/en/cloud/storage/pca/api/)

ncw:
What would be the minimum changes needed for rclone to support this?

A new flag `--swift-storage-policy` which would add the paramenter to a `X-Storage-Policy` header on container create.  Would that make is useful enough?

The user could attempt to read cold data, get a 429 error, wait for a bit, try again and then it will work?

wlhlm:
> A new flag --swift-storage-policy which would add the paramenter to a X-Storage-Policy header on container create. Would that make is useful enough?

Yes, I think that is enough for basic support. It mostly works now by using the generic Swift driver and manually creating the _segments-container beforehand.

> The user could attempt to read cold data, get a 429 error, wait for a bit, try again and then it will work?

Yes, that works.

DougParker1992:
I agree, I just had this happen to me. For now I can just manually created the segment container for cold storage but it would be ideal if rclone supported a flag that adds a parameters for the X-Storage-Policy Header.

ncw:
[...]

PR #2439, opened 2018-08-03
Swift: Add storage_policy

rubenbe:
Add support for swift storage policies.
Tested against OVH and I can use this to create PCA containers.
This implements the container creation mentioned in #1603.
Is there a way to add specific automated tests for this? I didn't immediately find a good place it.

### DIFF
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -121,6 +121,20 @@ func init() {
 				Value: "admin",
 			}},
 		}, {
+			Name:    "storage_policy",
+			Help:    "The storage policy to use when creating a new container",
+			Default: "",
+			Examples: []fs.OptionExample{{
+				Help:  "Default",
+				Value: "",
+			}, {
+				Help:  "OVH Public Cloud Storage",
+				Value: "pcs",
+			}, {
+				Help:  "OVH Public Cloud Archive",
+				Value: "pca",
+			}},
+		}, {
 			Name:     "chunk_size",
 			Help:     "Above this size files will be chunked into a _segments container.",
 			Default:  fs.SizeSuffix(5 * 1024 * 1024 * 1024),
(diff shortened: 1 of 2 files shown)


backend/swift/swift.go
```diff
@@ -120,6 +120,10 @@ func init() {
 				Help:  "Admin",
 				Value: "admin",
 			}},
+		}, {
```
ncw: It would be nice to have some example storage policies here so users know what to type.

Either some `Examples` or another line in the Help with some text.  Seperate it with `\n` otherwise it will appear in `rclone --help`

backend/swift/swift.go
```diff
@@ -583,7 +588,9 @@ func (f *Fs) Mkdir(dir string) error {
 		_, _, err = f.c.Container(f.container)
 	}
 	if err == swift.ContainerNotFound {
-		err = f.c.ContainerCreate(f.container, nil)
+		headers := swift.Headers{}
+		headers["X-Storage-Policy"] = f.opt.StoragePolicy
```
ncw: This will be adding a blank storage policy header when it isn't defined - I think it would be better not to supply the header at all in that case - what do you think?

backend/swift/swift.go
rubenbe: I'll add some examples.
I could only find OVH implementing its cold storage according to the "swift standard" [0]
It seems that Rackspace does not offer cold storage.
And IBM has implemented this as a separate "location" [1], e.g. eu-vault / eu-cold

[0]https://docs.openstack.org/swift/latest/overview_policies.html
[1]https://console.bluemix.net/docs/services/cloud-object-storage/basics/classes.html#use-storage-classes

OTH, this is the equivalent of the storage classes on S3 and GCS, where these options are non-advanced too. So for the sake of consistency I'd keep this option out of the advanced options.

backend/swift/swift.go
rubenbe: You're right, it's cleaner from an API point of view.
Although strictly seen it doesn't make any difference for swift, as an undefined header and empty string are handled identically.
https://github.com/openstack/swift/blob/86d78f6667392b2ee0ca017da8f54ff153aca491/swift/proxy/controllers/container.py#L61

rubenbe:
I did not generate all the docs, because my pandoc version insisted on rewriting almost all files.

Issue #2008, opened 2018-01-24
--track-renames doesn't work with 'copy' command on Google Drive


jschwalbe:
_Version:_
rclone v1.39
- os/arch: darwin/amd64
- go version: go1.9.2

_Command:_
`rclone copy /Volumes/WD3TB/TV/Aerial\ America/ gdrive:/TV/WD3TB/Aerial\ America/ -vv --track-renames --size-only --transfers 1`

Here's the setup: I previously uploaded a file (name: `file`) to Google Drive via rclone. Then I change the name on my local computer to `samefile`. Here's the output (slightly trimmed for conciseness):
```
2018/01/23 20:03:15 DEBUG : Using config file from "/Users/username/.rclone.conf"
2018/01/23 20:03:15 DEBUG : rclone: Version "v1.39" starting with parameters ["rclone" "copy" "-vv" "/Volumes/WD3TB/TV/Aerial America/" "gdrive:/TV/WD3TB/Aerial America/" "-vv" "--track-renames" "--size-only" "--transfers" "1"]
<snipped a few pacer rate limited errors>
2018/01/23 20:03:20 INFO  : Google drive root 'TV/WD3TB/Aerial America': Modify window is 1s
2018/01/23 20:03:20 INFO  : Google drive root 'TV/WD3TB/Aerial America': Making map for --track-renames
2018/01/23 20:03:20 INFO  : Google drive root 'TV/WD3TB/Aerial America': Finished making map for --track-renames
2018/01/23 20:03:20 INFO  : Google drive root 'TV/WD3TB/Aerial America': Waiting for checks to finish
2018/01/23 20:03:20 INFO  : Google drive root 'TV/WD3TB/Aerial America': Waiting for renames to finish
2018/01/23 20:03:22 INFO  : samefile: Copied (new)
2018/01/23 20:03:31 DEBUG : America in Color - The 1920s.mp4: Sending chunk 0 length 8388608
^C
```

As you can see, rclone copied `samefile` over as if it were a new file. Then it begins sending over the next file (which also is already uploaded, just with a different name). 

Curiously, this also happens when a file is moved to another folder (which I thought was supposed to be covered as well by `--track-renames`). For example: Directory1/file is already on Google Drive. On my computer move file to Directory1/subdirectoryA/file and rclone begins a brand new upload (a duplicate copy). 

Any thoughts? Am I using `--track-renames` incorrectly? Thanks much

ncw:
[...]

PR #2445, opened 2018-08-06
docs: remove references to copy and move for --track-renames

X1011:
this change was omitted in the fix for #2008


ncw:
Thank you very much - yes that was missed :-)

PR #2447, opened 2018-08-06
Add reveal command

kir4h:
This PR adds the `reveal` command, which is the opposite of the existing `obscure` command.

It will take as an argument an obscured password (that is, a password as stored in rclone.config) and will return the actual password value. 

I'm taking for granted this command doesn't exist just because it was not implemented, and not because we don't want it to be easily available. In the end, my understanding is that obscurity is just an extra step for avoiding visual eavesdropping.

Lot of global flags are available for this command and are useless, but since obscure command also has them I didn't go further in removing them.

### DIFF
--- a/cmd/all/all.go
+++ b/cmd/all/all.go
@@ -42,6 +42,7 @@ import (
 	_ "github.com/ncw/rclone/cmd/purge"
 	_ "github.com/ncw/rclone/cmd/rc"
 	_ "github.com/ncw/rclone/cmd/rcat"
+	_ "github.com/ncw/rclone/cmd/reveal"
 	_ "github.com/ncw/rclone/cmd/rmdir"
 	_ "github.com/ncw/rclone/cmd/rmdirs"
 	_ "github.com/ncw/rclone/cmd/serve"
--- a/cmd/reveal/reveal.go
+++ b/cmd/reveal/reveal.go
@@ -1,0 +1,30 @@
+package reveal
+
+import (
+	"fmt"
+
+	"github.com/ncw/rclone/cmd"
+	"github.com/ncw/rclone/fs/config/obscure"
+	"github.com/spf13/cobra"
+)
+
+func init() {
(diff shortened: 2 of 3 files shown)


ncw:
> I'm taking for granted this command doesn't exist just because it was not implemented, and not because we don't want it to be easily available. In the end, my understanding is that obscurity is just an extra step for avoiding visual eavesdropping.

I didn't implement it just to add a small extra step in un-obscuring the passwords....

I'm fully aware that the rclone password obscuration isn't secure, it is just to protect against visual eavesdropping like you said (and a good way of putting it).

What do you think about making the command hidden - add `Hidden: true` in the command definition then it won't appear in the help but will still be usable?

> Lot of global flags are available for this command and are useless, but since obscure command also has them I didn't go further in removing them.

Yes that is a different project!

kir4h:
> What do you think about making the command hidden - add Hidden: true in the command definition then it won't appear in the help but will still be usable?

Added! In the end, my goal was to provide with a mechanism that will allow recovering this initial password. The scenario would be using auto generation of password but lack of awareness from the user that those passwords will be stored in an obscured format, ending up on a scenario where the original password is unknown by the user itself.

Just thinking, maybe a disclaimer when using auto-generated passwords would be good?

ncw:
> Added

Thanks!

> Just thinking, maybe a disclaimer when using auto-generated passwords would be good?

Do you mean telling the user to make a copy of the password as it will be stored obscured in the password file?

kir4h:
> Do you mean telling the user to make a copy of the password as it will be stored obscured in the password file?

Yes. Just a proposal to minimize the use cases of the need for a "reveal" option :)

ncw:
> Yes. Just a proposal to minimize the use cases of the need for a "reveal" option :)

Good idea. Fancy sending a PR for the relevant strings in the code?

I'll merge this now - thank you very much for your contribution :-)

PR #2450, opened 2018-08-08
Add disclaimer about generated passwords being stored in an obscured format

kir4h:
This PR adds an extra message when creating a new crypt remote and choosing to auto-generate a password for encyption. It warns the user that the "plaintext" version of the password won't be stored so that he remembers to keep it safe somewhere.

@ncw Feel free to comment about the sentence, I am open to any change to make it as clear as possible or if you think the sentence needs to be introduced at some other stage. I have kept line length to 120, not sure what is the convention.

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -770,7 +770,9 @@ func ChooseOption(o *fs.Option, name string) string {
 				}
 				password = base64.RawURLEncoding.EncodeToString(pw)
 				fmt.Printf("Your password is: %s\n", password)
-				fmt.Printf("Use this password?\n")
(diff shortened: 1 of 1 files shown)


ncw:
I think the wording looks fine :-)

Can you add hard linebreaks `\n` to keep the line lengths to about 70 chars?  That is what I normally do for the terminal messages.

Thank you

kir4h:
> Can you add hard linebreaks \n to keep the line lengths to about 70 chars? That is what I normally do for the terminal messages.

Done!

ncw:
Super - thank you :-)

PR #2456, opened 2018-08-10
vfs: "Seek" in the --buffer-size data

B4dM4n:
Try to fulfill `Seek` calls on read only files by discarding bytes in the current buffer data.

Previously all `Seek` calls would discard the whole buffer and open a new connection every time.
Now they will consume the buffer until the right position is reached and only open a new connection if the position is outside the buffer data.

It is also possible to `Seek` backwards, but only inside the current buffer block.

### DIFF
--- a/fs/asyncreader/asyncreader.go
+++ b/fs/asyncreader/asyncreader.go
@@ -180,6 +180,76 @@ func (a *AsyncReader) WriteTo(w io.Writer) (n int64, err error) {
 	}
 }
 
+// SkipBytes will try to seek 'skip' bytes relative to the current position.
+// On success it returns true. If 'skip' is outside the current buffer data or
+// an error occurs, Abandon is called and false is returned.
+func (a *AsyncReader) SkipBytes(skip int) (ok bool) {
+	a.mu.Lock()
+	defer func() {
+		a.mu.Unlock()
+		if !ok {
+			a.Abandon()
+		}
+	}()
+
+	if a.err != nil {
+		return false
+	}
+	if skip < 0 {
+		// seek backwards if skip is inside current buffer
+		if a.cur != nil && a.cur.offset+skip >= 0 {
+			a.cur.offset += skip
+			return true
+		}
+		return false
+	}
+	// early return if skip is past the maximum buffer capacity
+	if skip >= (len(a.ready)+1)*BufferSize {
+		return false
+	}
+
+	refillTokens := 0
+	for {
+		if a.cur.isEmpty() {
+			if a.cur != nil {
+				a.putBuffer(a.cur)
+				refillTokens++
+				a.cur = nil
+			}
+			select {
+			case b, ok := <-a.ready:
+				if !ok {
+					return false
+				}
+				a.cur = b
+			default:
+				return false
+			}
+		}
+
+		n := len(a.cur.buffer())
+		if n > skip {
+			n = skip
+		}
+		a.cur.increment(n)
+		skip -= n
+		if skip == 0 {
+			for ; refillTokens > 0; refillTokens-- {
+				a.token <- struct{}{}
+			}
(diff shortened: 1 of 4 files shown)


fs/asyncreader/asyncreader.go
```diff
@@ -180,6 +180,77 @@ func (a *AsyncReader) WriteTo(w io.Writer) (n int64, err error) {
...
+		// seek backwards if skip is inside current buffer
+		if a.cur != nil && a.cur.offset+skip >= 0 {
+			a.cur.offset += skip
+			a.mu.Unlock()
+			return true
+		}
+		goto abandon
+	}
+	// early return if skip is past the maximum buffer capacity
+	if skip >= a.buffers*BufferSize {
```
ncw: Might this be better as this to only skip data which has been read?

```go
if skip >= len(a.ready)*BufferSize {
```

Or maybe that should be`len(a.ready)+1` - not sure.

fs/asyncreader/asyncreader.go
```diff
@@ -180,6 +180,77 @@ func (a *AsyncReader) WriteTo(w io.Writer) (n int64, err error) {
 	}
 }
 
+// SkipBytes will try to seek 'skip' bytes relative to the current position.
+// On success it returns true. If 'skip' is outside the current buffer data or
+// an error occurs, Abandon is called and false is returned.
+func (a *AsyncReader) SkipBytes(skip int) bool {
+	a.mu.Lock()
```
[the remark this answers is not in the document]
B4dM4n: Good point, using named return values in this case simplifies the return statements a lot.

fs/asyncreader/asyncreader.go
B4dM4n: Good point, `len(a.ready)+1` is closer to the real buffer length than `a.buffers`.

B4dM4n:
I came up with this while digging into straces of mediainfo. 

It makes some seeks in the file header when analysing files. I was curious if rclone would utilize the buffer data for these close jumps. It didn't, so I added it :-)

The benefit is probably pretty small and only noticable in specific use cases, but if it can save some API calls and helps some users it is worth it.

animosity22:
I'd love to test this out as I think a nice buffer would help with the open/close/open/close dance that plex does at times with files.

If you had a lot of extra memory, I think this would work for a nice big buffer size setting if you had spare memory/bandwidth/etc.

B4dM4n:
> I'd love to test this out as I think a nice buffer would help with the open/close/open/close dance that plex does at times with files.

Since the buffer is not preserved between open calls, this change will not help in this situation.

It should however help when you jump `1 min` forward and the needed file location is already in the buffer.

BinsonBuzz:
if this is implemented, would a mount like this make sense to have a big seek buffer?

`rclone mount --allow-other --dir-cache-time 72h --vfs-read-chunk-size 128M --vfs-read-chunk-size-limit off --buffer-size 2G --log-level INFO gdrive_media_vfs: /mnt/user/mount_rclone/google_vfs`

Would the previous need to have read-chunk-size bigger than buffer-size to reduce API calls go away?

B4dM4n:
> if this is implemented, would a mount like this make sense to have a big seek buffer?

You should not depend to heavily on seeking the buffer data. It only works when seeking forwards. If you want to use a big buffer, you should first ensure that the file access pattern of the program can utilize the buffer.

Only one `Seek` backwards is enough to discard the whole buffer.

> Would the previous need to have read-chunk-size bigger than buffer-size to reduce API calls go away?

This was never a hard requirement. You could always set `buffer-size` bigger than `vfs-read-chunk-size`. This can lead to a higher number of API calls, which in turn could result in rate limiting on some remotes and therefore slow down opening files or reading data.

ncw:

PR #2457, opened 2018-08-10
docs: clearify buffer-size is per transfer/filehandle

B4dM4n:
Since there seems be some confusion how much memory `--buffer-size` uses, I suggest we change the command line help, to make it clearer it's per `--transfer`.

The full documentation already makes this clear, but the CLI help is probably the first place people will read and should express the same.

I also added a paragraph for `mount` and `cmount` to the documentation that they behave somewhat differently.

Should the CLI help also include this difference?
Is it possible change the help text for `mount` and `cmount` only?


ncw:

PR #2460, opened 2018-08-12
jottacloud: Handle empty time values

mpolden:
I encountered a time parsing bug while testing the new Jottacloud backend. If a
mount point in Jottacloud has never been written to, the `modification` XML
element will be an empty string.

Example from my initial testing. My `jotta` remote was configured to use the
`/Sync` mount point, which I had never used before.

```
$ rclone ls jotta:
2018/08/11 22:48:36 ERROR : : error listing: couldn't list files: parsing time "" as "2006-01-02-T15:04:05Z0700": cannot parse "" as "2006"
2018/08/11 22:48:36 Failed to ls: couldn't list files: parsing time "" as "2006-01-02-T15:04:05Z0700": cannot parse "" as "2006"
```

### DIFF
--- a/backend/jottacloud/api/types.go
+++ b/backend/jottacloud/api/types.go
@@ -12,41 +12,35 @@ const (
 	timeFormat = "2006-01-02-T15:04:05Z0700"
 )
 
-// Time represents represents date and time information for the
-// Jottacloud API, by using a custom RFC3339 like format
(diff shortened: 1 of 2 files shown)


mpolden:
Any chance of getting this merged? The only failing build is Go 1.7.6, which failed on something unrelated.

ncw:
Apologies for the delay - I've been travelling.

I'll merge this now - it looks great - thanks :-)

Issue #2091, opened 2018-02-23
Support for Azure Archive Blob Storage


kflu:
Opening a feature request per @ncw suggestion. [Here's][2] the original discussion thread. 

I would like to use rclone to back up to Azure [**Archive** blob storage][1] (cheapest cold storage service as far as I know). Since Azure blob storage container cannot set default blob tier to Archive, the clients must explicitly set the tier for the uploaded blobs to Archive after or during the uploads. However, it appears that rclone's Azure Blob storage implementation doesn't support this as of now.

It would be great to allow a configuration that [sets the tier of the blob][3] when uploading it. 

As Azure Archive Blob Storage is [so cheap][4] (half of Glacier's pricing at the moment), this would be quite a popular use scenario for rclone.

I'm not a go-lang dev, but I think it's not too hard for someone fluent in Go :) ? I'd be happy to test it if someone implements it!

[1]: https://azure.microsoft.com/en-us/blog/announcing-the-public-preview-of-azure-archive-blob-storage-and-blob-level-tiering/
[2]: https://forum.rclone.org/t/support-for-azure-archive-blob-storage/4891
[3]: https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-tier
[4]: https://azure.microsoft.com/en-us/pricing/details/storage/blobs/

ncw:
Unfortunately this isn't supported yet by the libary rclone uses to access the azure blob storage.  The relevant issue is here: https://github.com/Azure/azure-sdk-for-go/issues/808

Otherwise it would be quite easy to add!

kflu:
If we don't want to wait for the SDK support, another option would be to call the [Set Blob Tier REST API](https://docs.microsoft.com/en-us/rest/api/storageservices/set-blob-tier):

The URL to call is `https://myaccount.blob.core.windows.net/mycontainer/myblob?comp=tier`, then set the desired tier as one of the HTTP header: `x-ms-access-tier`.

kflu:
Or, is there a way to configure rclone to invoke a command for every file it had just uploaded. And I would write a small executable `set-blob-tier` and hook them up?

kflu:
Hi Nick @ncw , looks like the SetTier is supported in the base library. Is this a good time to incorporate it? Refer to: https://github.com/Azure/azure-storage-blob-go/issues/1

hashbackup:
[...]

PR #2461, opened 2018-08-13
AzureBlob: Add SetBlobTier feature

sandeepkru:
azureblob: Added blob tier feature, a new configuration of azureblob-access-tier is added to tier blobs between Hot, Cool and Archive. Addresses #2901


sandeepkru:
@ncw - These are basic changes, Can you please take a look at my open questions mentioned in #2091. 

Also, I wanted to check with you on adding validations of tier on `sync`, `copy`, `move` etc. operations. For example if user invokes `sync` operation with desired access tier and if blobs exist in different access tier, ideally we should tier those blobs to desired tier. - This doesn't exist yet.

I verified code, based on my understanding all validations whether to skip file or not exists here
https://github.com/ncw/rclone/blob/751bfd456f459cb2046af249d91e20be8fbc8ea2/fs/operations/operations.go#L106

Since this is not common to all backends, adding `access_tier` matching condition is not valid here. Do you have any suggestions on where to add this?

Also, I've not yet updated documentation, I'll update in next iteration along fixes with other open questions.

backend/azureblob/azureblob.go
```diff
@@ -1199,8 +1219,28 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
...
 	}
+	// Refresh metadata on object
 	o.clearMetaData()
-	return o.readMetaData()
+	err = o.readMetaData()
+	if err != nil {
+		return err
+	}
+
+	// Now, set blob tier based on configured access tier
```
ncw: Presumably we only need to actually set the blob tier if it differs from the default?  If so then we can save an HTTP roundtrip for the default case.

backend/azureblob/azureblob.go
```diff
@@ -1199,8 +1219,28 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
...
+	if o.fs.opt.AccessTier == string(azblob.AccessTierNone) || o.fs.opt.AccessTier == string(o.AccessTier()) {
+		return nil
+	}
+
+	desiredAccessTier := azblob.AccessTierType(o.fs.opt.AccessTier)
+	err = o.fs.pacer.Call(func() (bool, error) {
+		_, err := blob.SetTier(ctx, desiredAccessTier)
```
ncw: Can we not set the tier as part of the upload?

ncw:
> Can you please take a look at my open questions mentioned in #2091.

I'll answer those over there

> Also, I wanted to check with you on adding validations of tier on sync, copy, move etc. operations. For example if user invokes sync operation with desired access tier and if blobs exist in different access tier, ideally we should tier those blobs to desired tier. - This doesn't exist yet.

I think we should not move blobs between tiers and document that we don't - I don't want to further complicate the sync routines!  I think in practice this won't be a common operation - what do you think?

sandeepkru:
> Can you write some docs in docs/content/azureblob.md about the --azureblob-access-tier please?

Sure, will do today, I was just waiting to get clear understanding on what we support with this.

> I think we should not move blobs between tiers and document that we don't - I don't want to further complicate the sync routines! I think in practice this won't be a common operation - what do you think?

Sounds good to me, we can document this functionality clearly. I was just thinking use cases of `daily sync` where if user decides to change tier of all blobs to `archive` (for all the blobs already exist in Azure). With current implementation, user needs to re-upload everything just to change tier if they intend to use `rsync`. I was also thinking people may come-up with feature ask soon.

ncw:
> Sounds good to me, we can document this functionality clearly. I was just thinking use cases of daily sync where if user decides to change tier of all blobs to archive (for all the blobs already exist in Azure). With current implementation, user needs to re-upload everything just to change tier if they intend to use rsync. I was also thinking people may come-up with feature ask soon.

On the issue (I think) I suggested that this might be another command - I think that is probably the right approach here.

I think I'm awaiting some docs before merging?

sandeepkru:
> I think I'm awaiting some docs before merging?
Yes, this PR is still pending from my end. I've changes changes ready with error handling on `Open` and documentation, will send out new iteration today.

> On the issue (I think) I suggested that this might be another command - I think that is probably the right approach here.
Correct, you had this recommendation, and I wasn't sure if it is a good idea to have generic command to support only one back end. I'll see if I can add those changes as part of same PR or will work as separate item.

ncw:
> I'll see if I can add those changes as part of same PR or will work as separate item.

A new PR would be my preference so we can get this bit merged ASAP!

sandeepkru:
@ncw - Please take a look if you get a chance, please let me know if you have any suggestions before merge.
Also, the failing test is a known one.

ncw:

I'll merge it in a moment.

PR #2463, opened 2018-08-14
Jottacloud: Fix wrong error check

olihey:
Add JMd5 header if MD5 is available


ncw:
I would write `err == nil && md5 != ""` probably here.

There are backends which support md5 but don't necessarily have an md5 for each object (s3 is an example)  - these return no error but "" for the hash.

olihey:
@ncw Changed it. Wasn't aware of this.

ncw:
Thank you - I've merged this :-)

It will likely conflict with your other PR but merging this is helpful

PR #2465, opened 2018-08-14
crypt: ObjectInfo should return NotSupported instead of empty string …

olihey:
@ncw I am not sure if that makes sense at all but returning a empty string as hash and NO error seems a bit dnagerous. I would assume the crypt backend should return the hash.ErrUnsupported for hash values


olihey:
Closed since this is expect behavior according to @ncw

Issue #2467, opened 2018-08-14
Add support for merging remotes


felixbrucker:
I use rclone to export remotes/part of remotes via the http serve command. This works quite well.
However my remotes use the same structure and just add content to it. As such it would be great to be able to export that as a single directory via http serve.

As of right now the only way to achieve this (at least that i know of) is to use the fuse mount option, use a 3rd party fuse based merge-fs like mergerfs etc, and then export that merged fuse mount. This is a lot of overhead and not working on machines without fuse support.

I searched the issues and there is one similar to this, but proposing to implement this via the fuse mount, which doesn't really improve the situation described above, i'll link it here for reference: https://github.com/ncw/rclone/issues/909

I know this probably is more work to do (and there exists a workaround mentioned above), so i'll leave it here and see if anybody wants to have this as well.

vb0:
If I understand the request correctly this could be extremely useful in many cases. For example if you want to backup one large disk/folder to a couple (smaller) portable drives, a very common scenario for offline backups. You don't need to split the source, to take care what folders go on which drive, just connect the drives and let the software pick a place to put stuff on either drive.

felixbrucker:
I thought more of a read-only approach to just export multiple remotes as a single directory structure which should be easier. However a way to write to such a merged remote could be done as well.

ncw:
This would be best implemented as a remote which wraps other remotes.  You could read from all remotes, but only write to one of them.

You'd also want to define what happens when the same file is in all the backend remotes.

It doesn't sound too tricky - fancy having a go?

vb0:
Write support (which isn't in the original scope of this issue, I didn't want to hijack it) is actually a little more complicated than writing just to one remote - it involves writing to any of the remotes if it involves an existing files you want to change. And this is not only for delete and move but also for copy when you want to overwrite existing files. We need probably to get some inspiration for all the corner cases from mergerfs or similar.

felixbrucker:
@ncw Unfortunately i have never used go and looking at the code i don't understand much, so i'll pass.

However i'm putting up a bounty of 20$ via PayPal or BTC to get this implemented, everybody is welcome to implement this feature and claim the bounty.

felixbrucker:
@ncw I took a deeper look at go and rclone and got it working read-only in a "make it work" fashion.

Some stuff i'm not sure about how to handle correctly are:

[...]

PR #2474, opened 2018-08-19
config: add List value types

B4dM4n:
To help #2467 and possibly replace list parsing methods in other remotes, this adds two `[]string` value types for config values, `CommaSepList` and `SpaceSepList`.

They use the `encoding/csv` package for parsing, which allows to use quoted entries inside the list.

### DIFF
--- a/fs/config/config_list.go
+++ b/fs/config/config_list.go
@@ -1,0 +1,94 @@
+package config
+
+import (
+	"bytes"
+	"encoding/csv"
+	"fmt"
+)
+
+// CommaSepList is a comma separated config value
+// It uses the encoding/csv rules for quoting and escaping
+type CommaSepList []string
+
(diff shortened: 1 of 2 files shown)


ncw:
That is great - thank you @B4dM4n :-)  I like the example tests - I always forget to make those and they are really helpful.

ncw:
Merged too soon!  These need to be in `fs` otherwise the core rclone config can't use them.  `fs` can't import `fs/config` as it creates a circular dependency.

Have fixed in 1b8f4b616c06fd54edd0290776e68d487f63e1e3

B4dM4n:
I was not aware of that import cycle. Sorry for the inconveniences.

Testing it in a `backend` worked fine :smile:

ncw:
@B4dM4n no worries - all working now!

PR #2475, opened 2018-08-19
jottacloud: Implement MimeTyper and Abouter interface.

buengese:
Stuff I forgot to implement in the initial pull request. In theory Jottacloud also supports link sharing but there is no up to date documentation on how to do it.

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -87,6 +87,7 @@ type Options struct {
 type Fs struct {
 	name        string
 	root        string
+	user        string
 	opt         Options
 	features    *fs.Features
 	endpointURL string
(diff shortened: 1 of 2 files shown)


backend/jottacloud/jottacloud.go
```diff
@@ -634,6 +638,15 @@ func (f *Fs) DirMove(src fs.Fs, srcRemote, dstRemote string) error {
 	return nil
 }
 
+// About gets quota information
+func (f *Fs) About() (*fs.Usage, error) {
+	usage := &fs.Usage{
```
ncw: This is more useful if it is "live" so that a person using a jottacloud mount can run `df` and see the amount changing.  Mount caches the value for the same length of time as the directory cache time.

backend/jottacloud/jottacloud.go
```diff
@@ -250,6 +252,8 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	f.features = (&fs.Features{
 		CaseInsensitive:         true,
 		CanHaveEmptyDirectories: true,
+		ReadMimeType:            true,
```
ncw: This probably belonged in the previous commit, but it isn't a big deal!

buengese:
I addressed the comments btw. so this should be mergeble.

ncw:
Thanks for that :-)

I'll merge it now.

PR #2476, opened 2018-08-19
box: Implement link sharing

buengese:


backend/box/box.go
```diff
@@ -120,6 +120,7 @@ type Object struct {
 	size        int64     // size of the object
 	modTime     time.Time // modification time of the object
 	id          string    // ID of the object
+	publicLink  string    // Public Link for the object
```
ncw: Is the publicLink mostly empty?  I'm just wondering how much this will bloat memory usage as sometimes rclone has millions of Object~s in memory.

backend/box/box.go
```diff
@@ -293,6 +294,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 			}
 			return nil, err
 		}
+		f.features.Fill(&newF)
```
ncw: What is this change for?  Hasn't features been fill-ed already?

backend/box/box.go
```diff
@@ -838,6 +840,46 @@ func (f *Fs) DirMove(src fs.Fs, srcRemote, dstRemote string) error {
...
+	if err == nil {
+		fs.Debugf(f, "attempting to share directory '%s'", remote)
+
+		opts = rest.Opts{
+			Method:     "PUT",
+			Path:       "/folders/" + id,
+			Parameters: fieldsValue(),
+		}
+	} else {
+		fs.Debugf(f, "attempting to share signle file '%s'", remote)
```
ncw: typo here :-)

backend/box/box.go
buengese: It should be empty for every object that isn't shared.

backend/box/box.go
buengese: In the case the case that root is a file a copy of the fs struct is created and used but operations like public link take the function pointers from the Feature list which still use the old object. I wasn't sure if this was intentional or not. In my own manual test share TestBox:Testfile f.Root as accessed from PublicLink will be "Testfile" which would make the Parameter to PublicLink redundant.

buengese:
Please not that I'm on vacation right now. I will address the points in the review but it might take a few days.

backend/box/box.go
buengese: No pr without typos from me apparently.

ncw:
> Please not that I'm on vacation right now. I will address the points in the review but it might take a few days.

No worries - happy vacation!

backend/box/box.go
ncw: Oh - good catch.  I think I've made that mistake before and likely it is in other backends too...

backend/box/box.go
ncw: I think if it is mostly empty then that is fine.

buengese:
I'm pretty sure this has been done for a while unless there's something else you want me to change.

ncw:
Hey that looks great thank you!

I'll merge it now :-)

jonathanking:
How difficult is it to access the raw data URL instead of the box preview page? I'd like to share a link to a file without going through the Box preview.

ncw:
> How difficult is it to access the raw data URL instead of the box preview page? I'd like to share a link to a file without going through the Box preview.

What we provide is the shared link.  I had a quick look through [the docs](https://developer.box.com/reference) to see if there was something better but I couldn't see one - can you?

PR #2477, opened 2018-08-19
(minor) onedrive: Add back the check for DirMover interface

Cnly:
It seems I've forgotten to uncomment this check.


ncw:
That looks great - thanks :-)

Issue #2046, opened 2018-02-04
Copy to gdrive with gdocs conversion


dav1303:
Hi

In some cases users want to let Microsoft Office because of ans oldest version or not want to pay for a new license.

Is it possible to add à flag to launch the conversion of an office
Document to is équivalent un gdocs when making a copy ?

Regards

B4dM4n:
This sounds like a good idea and would complete document handling for drive.

The problem is that rclone can't verify the checksum or size when doing server side conversions. You would have to run it with `--ignore-size` to upload files.
It should nevertheless be possible to implement this feature.

I suggest `--drive-import-formats` as a new parameter to specify the wanted conversions. By default this has to be disabled, because the document conversion is not lossless and can not be reverted.

The `--drive-formats` should probably be renamed to `--drive-export-formats` to make the difference between both clearer.

dav1303:
@B4dM4n Did you plan to integrate this feature ?

B4dM4n:
Yes, I'm working on this (see [B4dM4n/rclone#drive_import_formats](https://github.com/B4dM4n/rclone/tree/drive_import_formats))
It should be working already, but I would like to implement some tests to verify this.

@ncw I would like to add some tests to `drive_test.go` but I know it is generated from `gen_tests.go`. Any advice how I could add something like `TestInternal` to run before `TestFinalise`?

dav1303:
Hi

Please let me know when your test are fine on your side and i can also test
myself.

Regards

dav1303:
@B4dM4n have you finish your tests ?

Can i help you on tests ?

dav1303:
Hello

Some news around that enhancement ?

Thanks

B4dM4n:
@dav1303 
Sorry for the delay, but I didn't have time in the last weeks to work on this issue.

The problem I'm facing is mapping the real filename to a Google document name and back to a real filename for copy and sync to work properly.
I think I found a solution and will try to implement it in time for rclone v1.43.

dav1303:
Hi
No problem 
Take your time i'll just want to follow.
Regards

B4dM4n:
There has been progress for this issue:
#2479 is a work in progress rewrite of the drive documents handling. 
If anyone wants to help testing the new implementation, there is a beta branch containing the changes: 
https://beta.rclone.org/branch/v1.42-162-g6a66967c-drive-documents/

**Keep in mind this is not tested thoroughly yet and therefore be considered experimental.**

[...]

Issue #1349, opened 2017-04-17
Option to create links to google documents instead of converting it


pckapps:
When syncing with Google Drive, would it be possible to add an option so that links to google documents are created instead of converting the files? Double clicking on the file should then open the browser with the relevant file. I believe this is how it works when using the official Drive client by Google. This should be similar to how "--drive-skip-gdocs" works.

ncw:
How does that actually work?  What type of file is created and what is in it?  I suspect this intercepted by the google drive client somehow.

pckapps:
Yes, the Google Drive client creates a file called "WorkbookName.gsheet". Inside is simple json, something like this:

```
{"url": "https://docs.google.com/open?id=aaabbbccc",
"doc_id": "aaabbbccc",
"email": "myname@gmail.com",
"resource_id": "document:aaabbbccc"}
```

My suggestion is to create a similar file that natively opens the link in the browser. It's not ideal to edit my Google documents in Libre office and uploading it back to Google Drive.

What do you think?

agilmor:
The solution/alternative that [odeke-em/drive](https://github.com/odeke-em/drive) client does is to create a ``.desktop`` for all google docs with this fields:
```
Type=Link
URL=https://drive.google.com/...
```
I guess that it's the same idea, but for Linux.

This also works for other Drive but not Docs files like Zoho's ones, etc...
I'm movimng from odeke-em/drive to rclone and I'm missing this feature... so +1 to this feature!!

Thanks!

crysman:
:+1:  for this, as **this is how the official Google Drive Stream client behaves on Windows**. Would be great if this client behaved the same way! I am now looking for some good working alternative to Google Drive Stream, but not yet successfully because of this :(

ncw:
Does anyone want to work on this issue?

I'm happy to talk people through what needs to be done.

B4dM4n:
I have looked a bit into this issue while working on Google documents uploads.

It seems like there are different concepts of `link` files:

Windows and macOS have `.url` files.
```ini
[InternetShortcut]
URL=https://rclone.org/
```
Source: [stackoverflow](https://superuser.com/questions/689441/how-do-i-create-a-url-file-on-os-x)

macOS also has `.webloc` files.
```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

Issue #1065, opened 2017-01-24
Standalone Google Apps Script


peter-kehl:
Hi Nick,

would it be possible to support "standalone Google Apps Script" files from Google drive, please? Those are server-side javascripts created and run at https://script.google.com.

Google provides limited API to the source of those scripts. They show up at
https://drive.google.com/drive/u/0/search?q=type:script.

To complicate,
1. each such script ("project") contains of one or more scripts with extension ".gs" (as listed at https://script.google.com).
2. Those ".gs" scripts don't show up as separate files via Google API. Only the parent ("project") shows up.
3. Since those scripts don't count towards user's size quota, the parent shows up with size 0.

See also
http://stackoverflow.com/questions/18445863/how-to-download-google-apps-script
https://developers.google.com/apps-script/import-export#export_projects_from_drive

ncw:
What does rclone do with them at the moment?  (try with the -v flag).

Is this just a question of adding a mapping for `application/vnd.google-apps.script+json` exports?

See: https://github.com/ncw/rclone/blob/master/drive/drive.go#L60

Not sure about the parent vs the contents - how do you find the names of the contents if you can't see them?

peter-kehl:
ls lists the file (the container). Copying all files (not sure what rclone command it was, rclone docs are confusing) copies it as an empty file, with .desktop file:

cat PowerMute.desktop 
[Desktop Entry]
Icon=application-vnd.google-apps.script
Name=PowerMute
Type=Link
URL=https://script.google.com/d/19QeOin6dGLEuScs7TrMwlfsFna-U1vB4ITbVSv-C4-qRq8EoaPnlGUlV/edit?usp=drivesdk

Copy of the file fails:
rclone -v copy  Gdrive:PowerMute .
2017/01/27 18:17:41 rclone: Version "v1.35-DEV" starting with parameters ["rclone" "-v" "copy" "Gdrive:PowerMute" "."]
2017/01/27 18:17:42 Local file system at /home/pkehl/gdrive: Modify window is 1ms
2017/01/27 18:17:43 Local file system at /home/pkehl/gdrive: Waiting for checks to finish
2017/01/27 18:17:43 PowerMute: Modification times differ by -169h22m32.373s: 2017-01-24 06:18:02.373 +0000 UTC, 2017-01-17 15:55:30 +1100 AEDT
2017/01/27 18:17:43 Local file system at /home/pkehl/gdrive: Waiting for transfers to finish
[...]

PR #2479, opened 2018-08-21
WIP: drive: upload documents and document links

B4dM4n:
This has become a bigger change than thought at first.

To address #1349 #2046 and #1065 I rewrote many parts of the drive remote since the current state became to cluttered with all features stacked on top of it.

The two major things rewrote are the mime-type handling and splitting `Object` into smaller functional types.

The mime-type handling is now done by the `mime` package which makes use of the system wide mime-types. The basic mime-types that got used before are still registered.

The `Object` type is now split into 3 types and one `baseObject`.
- `baseObject` is embedded by all other objects and contains the shared fields
- `Object` now only represents a generic file without document handling 
- `documentObject` handles google docs. This still has as `Size` of `-1`, which gets updated after read.
- `linkObject` contains the new link file handling. Currently implemented are `.desktop`, `.link.html` (added `.link` since `.html` is a valid export format), `.url` and `.webloc` files. I could only test `.link.html` until now, but will test `.desktop` and `.url` soon.

Since this is a big change and not yet complete I don't think we should merge this before the `v1.43` release.
@ncw How do you like the split `Object` design. Are you fine with returning other concrete types besides `*Object` as `fs.Object`?

**TODO:**
- [x] Add documentation for `--drive-import-formats`
- [ ] Add documentation for rules of allowed conversions / updates
- [x] Add documentation for link files
- [x] Test `.url` link files on Windows
- [x] Test `.desktop` link files on Linux
- [x] Test `.url` and `.webloc` link files on MacOS
- [x] Test `application/vnd.google-apps.script` export/import
- [ ] Handle filename changes during `sync`?

**Examples:**
By default the conversion to Google Docs is only allowed when the uploaded file gets the same file type (determined by `--drive-export-formats`) as the source.

To upload `.doc` files as Google Docs use `.doc` as export and import format:
```
rclone copy --drive-import-formats=.doc --drive-export-formats=.doc /folder TestDrive:
```

To allow upload of multiple file types as Google Docs, `--drive-allow-import-name-change` has to be used:
```
rclone copy --drive-import-formats=.doc,.txt --drive-allow-import-name-change /folder TestDrive:
```

Using the link files works by specifying the link file extension as a `--drive-export-formats`:
```
rclone copy --drive-export-formats=.link.html TestDrive: /folder
```
This also works for `mount`:
```
rclone mount --drive-export-formats=.link.html TestDrive: /folder
```


ncw:
> This has become a bigger change than thought at first.

It did didn't it ;-)

> Since this is a big change and not yet complete I don't think we should merge this before the v1.43 release.

That sounds sensible to me.

> How do you like the split Object design. Are you fine with returning other concrete types besides *Object as fs.Object?

Absolutely.  As long as it satisfies the interface (and any optional interfaces we need on that Object) then I'm fine with that.

I think if the split object design makes the code cleaner then let's go with it.  In drive, gdocs and drive objects are quite different things so I think having two object types makes perfect sense.

If you want to push this to a branch on `ncw/rclone` then the CI will build a beta and we can get other people testing things when you are ready.

I think clear docs are going to be quite important for this feature!

I see this deprecates `--drive-formats` - will this continue to work as before so we don't break peoples scripts?  Do you think at some point we will remove it?  Or make it set both `import` and `export` formats?

B4dM4n:
> Absolutely. As long as it satisfies the interface (and any optional interfaces we need on that Object) then I'm fine with that.

Perfect. I put interface checks at the bottom of `drive.go` for all types as before.

> If you want to push this to a branch on ncw/rclone then the CI will build a beta and we can get other people testing things when you are ready.

I will do that. Just need to make sure people know that the new features (and maybe the whole drive remote) are experimental and can contain serious bugs.

> I think clear docs are going to be quite important for this feature!

Definitely. There are many details hidden from the users behind the `--drive-*-formats` flags. When used without `--drive-allow-import-name-change` the conversions should work seamless and otherwise produce understandable error messages.

> I see this deprecates --drive-formats - will this continue to work as before so we don't break peoples scripts? Do you think at some point we will remove it? Or make it set both import and export formats?

`--drive-formats` will work as before, even in combination with `--drive-import-formats`. There is a [migration](https://github.com/ncw/rclone/pull/2479/files#diff-c0b169225d4835a062c24b8bd8662f87R819) logic that transfers the value from `opt.Extensions` to `opt.ExportExtensions`, but only one of the options is allowed to be set.

At some point `--drive-formats` should be removed. It's name could imply that this is the default value for imports and exports. I will set `OptionHideConfigurator` in the next push to hide it during configuration.

`--drive-import-formats` should never get a default value, since the conversion is not lossless reversible and `rclone move` would destroy the original files this way.

ncw:
Happy for you to merge those commits after I get the release out the door (should be today hopefully!)

ncw:
OK the build is done! Happy you for you to merge away :-)

ncw:
Is this ready to merge @B4dM4n ?

B4dM4n:
There didn't arise new bugs since the last iteration in my testing and no other reported problems.

I rebased the code and it should be ready for beta.

ncw:
OK I'll merge that now and we can keep an eye on bug reports and the integration tests :-)

Thank you :-)

mhagnumdw:
Is the `--drive-export-formats=.link.html` option documented somewhere?

ncw:
See: https://rclone.org/drive/#import-export-of-google-documents

Issue #1320, opened 2017-04-10
Upload file to remote storage from URL


ext4:
Directly uploading to remote storage from a URL would be a useful feature, for example:

`rclone copy http:https://example.com/some_large_file.tar remote:/Files`

This would be useful for systems without enough free space to save the entire file before upload (for example a VPS).

I understand that currently `rclone` needs to know the size of the file before upload (https://github.com/ncw/rclone/issues/230#issuecomment-161762623), however properly configured webservers should send a [`Content-Length`](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13) header which indicates the size of the file.

ncw:
Nice idea!

I think that would be best implemented as a separate command, say

    rclone copyurl https://example.com/file.tar remote:File/file.tar

Yes rclone does need to know the size of files in advance, however for a lot of remotes, you can just use 0 and it will work.  This doesn't work with b2 for instance.  You mostly likely will get a Content-Length as you suggest though.

chmercesmoreira:
This is amazing, currently I'm running a VPS from OVH that has only 10gb of disk space (~8gb available for use), this surely will help a lot.

pmwoodward3:
+1 this would be awesome, would be even better if you could pass a list of URLs.

chmercesmoreira:
I'm dreaming:
Imagine if we could integrate [jdownloader2](https://support.jdownloader.org/Knowledgebase/Article/View/40/0/myjdownloaderorg-api-docu) with rclone?
Something like:
`rclone jd2 http://linkToCloudProvider remote:path`
I know, I know, I'm thinking too much big haha.

pmwoodward3:
I think a URL list and/or the ability to take an open directory and directly send that to a cloud destination would be amazing :)

desimaniac:
+1

I need this so bad!

scriptzteam:
nice idea :)

xiaolei0125:
+1
nice idea

gidoBOSSftw5731:
I understand this ISNT golang, but I think we should all take https://github.com/golang/go/wiki/NoPlusOne into account.

ncw:
Anyone want to work on the `copyurl` command?  Happy to walk people through what needs to be done?

gidoBOSSftw5731:
I can try to work on it @ncw, but my experience is limited (all ive done is setup [this](https://imagen.click) and some tinkering in golang and py.)

[...]

PR #2481, opened 2018-08-22
cmd: add copyurl command

kfur:
Implement #1320

### DIFF
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -328,7 +328,7 @@ func (s *server) postObject(w http.ResponseWriter, r *http.Request, remote strin
 		}
 	}
 
-	err := operations.UploadHttpBody(s.f, r.Body, r.ContentLength, remote)
+	err := operations.UploadHTTPBody(s.f, r.Body, r.ContentLength, remote)
 	if err != nil {
 		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -6,27 +6,23 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
 	"regexp"
 	"strconv"
 	"strings"
-	"time"
-
 	"github.com/ncw/rclone/cmd"
 	"github.com/ncw/rclone/cmd/serve/httplib"
 	"github.com/ncw/rclone/cmd/serve/httplib/httpflags"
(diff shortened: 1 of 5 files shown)


ncw:
Can you run the code through goimports then the CI will pass - thanks :-)

kfur:
@ncw It seems that CI passed

cmd/serve/restic/restic.go
```diff
@@ -326,46 +323,16 @@ func (s *server) postObject(w http.ResponseWriter, r *http.Request, remote strin
...
-		// Size unknown use Rcat
-		_, err := operations.Rcat(s.f, remote, r.Body, time.Now())
-		if err != nil {
-			fs.Errorf(remote, "Post request rcat error: %v", err)
-			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
-			return
-		}
+	err := operations.UploadHTTPBody(s.f, r.Body, r.ContentLength, remote)
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
ncw: This if clause needs

    fs.Errorf(remote, "Post request rcat error: %v", err)

If it is to be equivalent to the previous code.

It also needs

    accounting.Stats.Error(err)

(see below for the explanation)

fs/operations/operations.go
```diff
@@ -1315,6 +1315,48 @@ func NeedTransfer(dst, src fs.Object) bool {
...
+			closeErr := in.Close()
+			if closeErr != nil {
+				fs.Errorf(remote, "Post request: close failed: %v", closeErr)
+				if err == nil {
+					err = closeErr
+				}
+			}
+			ok := err == nil
+			accounting.Stats.DoneTransferring(remote, err == nil)
+			if !ok {
```
ncw: I think this `if` statement and the `account.Stats.Error` is wrong (it was wrong in the original code too!).

I think we should move it up into the caller so L333 in the restic.go or thereabouts.

You don't need to call it at all for `copyurl` since you are returning an error straight away from your code.

fs/operations/operations_test.go
```diff
@@ -468,6 +470,20 @@ func TestRmdirsLeaveRoot(t *testing.T) {
 	)
 }
 
+func TestUploadHTTPBody(t *testing.T) {
```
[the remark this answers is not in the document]
kfur: Do i need check size of `obj` in test case with unknown length ?

ncw:

I'll merge that now!

kfur:
@ncw thanks for grate software and your help !

kfur:
@ncw thanks for greate software and your help !

kfur:
@ncw thanks for great software and your help !

PR #2487, opened 2018-08-25
vfs: add vfs/poll-interval rc command

B4dM4n:
This rc command allows to query and edit the `--poll-interval` value.
The polling function will be restarted if the new interval is greater than 0.
```
rclone rc vfs/poll-interval interval=2m
```
To disable polling `interval=0` can be used.
```
rclone rc vfs/poll-interval interval=0
```
Without any parameter set, only the current status will be returned:
```json
{
	"enabled": true,
	"interval": {
		"raw": 60000000000,
		"seconds": 60,
		"string": "1m0s"
	},
	"supported": true
}
```

A use case for this command could be to disable change polling temporarily while is script is renaming entries in a large folder over multiple minutes. During this time, the directory cache would be reset many times and cause the folder (and subfolders) to be read again.

**Edit:**
After I pushed this PR I realized a design issue involving missed changes when the polling function gets restarted on every change.
I decided to change the `pollInterval` parameter of `ChangeNotify` to a `chan time.Duration`, to provide a way of updating a existing polling function. This channel also functions as a new way to stop the function by closing it.

### DIFF
--- a/backend/amazonclouddrive/amazonclouddrive.go
+++ b/backend/amazonclouddrive/amazonclouddrive.go
@@ -1240,24 +1240,38 @@ func (o *Object) MimeType() string {
 // Automatically restarts itself in case of unexpected behaviour of the remote.
 //
 // Close the returned channel to stop being notified.
-func (f *Fs) ChangeNotify(notifyFunc func(string, fs.EntryType), pollInterval time.Duration) chan bool {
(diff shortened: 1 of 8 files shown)


ncw:
Is this ready for merge?

B4dM4n:
Sorry for the delay.

I was thinking about adding `-1` as a restart value for the polling function, but it now doesn't seem necessary to me anymore.
This is working fine for me as it is, so I will merge it now.

PR #2488, opened 2018-08-26
Fix directory listing when -L is enabled and a bad symlink is encountered

connesc:
Currently, when -L is enabled, listing a directory fails if a bad symlink is found.
This PR simply ignore dangling symlinks.

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -16,8 +16,10 @@ import (
 	"unicode/utf8"
 
 	"github.com/ncw/rclone/fs"
+	"github.com/ncw/rclone/fs/accounting"
 	"github.com/ncw/rclone/fs/config/configmap"
 	"github.com/ncw/rclone/fs/config/configstruct"
+	"github.com/ncw/rclone/fs/fserrors"
 	"github.com/ncw/rclone/fs/hash"
 	"github.com/ncw/rclone/lib/readers"
 	"github.com/pkg/errors"
@@ -280,6 +282,13 @@ func (f *Fs) List(dir string) (entries fs.DirEntries, err error) {
 			// Follow symlinks if required
 			if f.opt.FollowSymlinks && (mode&os.ModeSymlink) != 0 {
 				fi, err = os.Stat(newPath)
+				if os.IsNotExist(err) {
+					// Skip bad symlinks
+					err = fserrors.NoRetryError(errors.Wrap(err, "symlink"))
+					fs.Errorf(newRemote, "Listing error: %v", err)
+					accounting.Stats.Error(err)
+					continue
+				}
 				if err != nil {
 					return nil, err
 				}


ncw:
This seems like a reasonable plan...  However I usually do what rsync does in this sort of case, so what does rsync do?

connesc:
Thanks for you answer.

Well, I have not found any reference to this particular case in the [rsync man page](https://download.samba.org/pub/rsync/rsync.html). Most options are related to recreation of symlinks at the destination, which is not yet addressed by rclone.

So, I just have tested with rsync (v3.1.3). I have started with the following `/tmp/src` directory, `baz` being the dangling symlink:

```
$ ls -l /tmp/src
total 4
lrwxrwxrwx 1 connesc connesc  3 août  26 17:57 bar -> foo
lrwxrwxrwx 1 connesc connesc  3 août  26 17:57 baz -> qux
-rw-r--r-- 1 connesc connesc 13 août  26 17:57 foo
```

Then, I have run rsync like this:

```
$ rsync -vrL /tmp/src/ /tmp/dst/
sending incremental file list
symlink has no referent: "/tmp/src/baz"
created directory /tmp/dst
./
bar
foo

sent 177 bytes  received 88 bytes  530.00 bytes/sec
total size is 8  speedup is 0.03
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1189) [sender=3.1.3]
```

Finally, the `/tmp/dst` directory is like this:

```
$ ls -l /tmp/dst
total 8
-rw-r--r-- 1 connesc connesc 13 août  26 17:57 bar
-rw-r--r-- 1 connesc connesc 13 août  26 17:57 foo
```

So, rsync essentially behaves like proposed, but exits with a non-zero code. Do you think we have to mimic this behavior as well?

connesc:
Sounds very good to me. Thanks for your help, I was not sure how to do it myself.
I have added those changes to the PR.

Now, my underlying use case (`rclone serve`) is just working as expected.

However, the error reporting also triggers retries for other commands, such as `rclone copy` or `rclone sync`. Ideally, this could be avoided because we know the error will persist. Do you think we can improve this easily?

Another good side of reporting the error is that it make `rclone sync` skip the deletion step, which is consistent with `rsync --delete`.

ncw:
> However, the error reporting also triggers retries for other commands, such as rclone copy or rclone sync. Ideally, this could be avoided because we know the error will persist. Do you think we can improve this easily?

Well spotted!

That is easy to fix :-)

When creating the error, use `fserrors.NoRetryError(errors.Wrap(err, "symlink"))` to wrap the error and it will signal to the high level loop that this error isn't worth retrying for.

connesc:
Thanks again for your help!
However, this does not seem to work, probably because `NoRetryError`s are only taken into account when they are returned, but not when they are passed to `accounting.Stats.Error()` (see [here](https://github.com/ncw/rclone/blob/007c7757d434deaf53d8b969353d85843506456e/cmd/cmd.go#L305)).

The reason behind this may be that multiple errors could be accounted, so just checking the last one might not be very relevant.
So, my proposal would be to add the following methods to `accounting.StatsInfo`, each of them returning `true` if at least one `(Fatal|NoRetry)Error` has been reported:

```go
HadFatalError() bool
HadNoRetryError() bool
```

Would this make sense to you?

ncw:
Sorry for the delay in responding - I had to get the v1.43 release out the door.

I think this PR is perfect now - I'll merge it.

The other PR needs a bit more discussion which we can do there.

PR #2491, opened 2018-08-27
[WIP] Add union remote

### LINKED ISSUE #2467: Add support for merging remotes

felixbrucker:
This PR adds support for a new remote: union.

The union remote "merges" multiple remotes into one logical structure. The content of directories with the same name are merged into a single directory. The ordering of the remotes decides which remote "overwrites" others if some files in a directory have the same name. The next remote in the list takes precedence. The last remote is used to write files, all other remotes are read-only.

Closes #2467


ncw:
This is looking nearly complete :-)  How do you think you are doing?

It looks like you'll need to run [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) on the files to fix the formatting of the includes as it does it slightly differently to go fmt.

felixbrucker:
Not sure if i did that correctly, i configured goimports style in goland

What about the integration test?
2 issues i see:
- one failure because i'm wrapping multiple fs and thus the returned fs is different from mine
- currently using `(*local.Object)` as Nil object for the test setup because my union fs test remote wraps three local fs's. If i just input Nil it can't find the local fs backend during the tests

ncw:
I've managed to fix the integration test with some fixes to the test framework.

I've squashed your commits and added one of my own which fixes the fs and sync integration tests and pushed these to the union branch.

https://github.com/ncw/rclone/tree/union
https://github.com/ncw/rclone/commits/union

Can you take a look over that and see if you are happy with it?

I think there are a few doc bits which need to be done, but otherwise it is looking complete :-)

https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend

Do you want to send a PR (on top of the union branch) for those last doc bits? I'm happy to do it if you haven't got time.

Thanks

Nick

felixbrucker:
Thanks, LGTM!

The only thing left regarding docs i found was the missing "union.md" in the make_manual.py, i ignored all other places which only list real filesystems and do not include alias as well as union isn't a "real" cloud storage provider but an abstraction. Did i miss some place regarding docs? In either case feel free to add the missing parts if necessary, greatly appreciated!

ncw:
Great!

I'l give it the once over and double check missing bits then merge :-)

felixbrucker:
Nice thanks!

ncw:
I've merged that now - thank you very much indeed for your contribution - very useful :-)

9e3ea3c6acbf1f427d82914ac8cc4ef91281ea8d

felixbrucker:
Nice, thanks!

PR #2493, opened 2018-08-27
cache: Fixes #2458

### LINKED ISSUE #2458: [Feature Request] VFS - cache files opened in read mode
Sorry for my bad english.
VFS does not cache files opened in read mode except in "--vfs-cachemode full".
The problem is that in full mode, --vfs-read-chunk-size is not applied and the entire file is downloaded.
It would be great if the VFS caches read mode file without full mode. (adding vfs-cachemode readwrite?)

dcpu:
Improve cache backend performance.

### DIFF
--- a/backend/cache/handle.go
+++ b/backend/cache/handle.go
@@ -304,11 +304,11 @@ func (r *Handle) Close() error {
 	close(r.preloadQueue)
 	r.closed = true
 	// wait for workers to complete their jobs before returning
-	waitCount := 3
+	waitCount := 300
 	for i := 0; i < len(r.workers); i++ {
 		waitIdx := 0
 		for r.workers[i].isRunning() && waitIdx < waitCount {
-			time.Sleep(time.Second)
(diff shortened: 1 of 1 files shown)


ncw:
Can you write the commit message in the [form recommended here](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages) please?

Does this fix #2458 ?  If so can you add "Fixes #2458" to the commit message too please.

I'd love a bit more explanation in the commit message too if possible - thanks!

The code itself looks correct, I'd just like to know a bit more about what problem it is fixing.

dcpu:
The loop interval for checking cache worker termination is too long.
Creates a latency of at least one second, even if the job finishes sooner.
This patch reduces loop interval to 10ms.

dcpu:
Close as https://github.com/ncw/rclone/pull/2506/commits/ec6aad36b0c7ef613b7a79a92f4bb93be068565f is better way to fix this.

PR #2495, opened 2018-08-27
cache: Not sending info request for cached chunk

dcpu:
Improve cache backend performance.

### DIFF
--- a/backend/cache/object.go
+++ b/backend/cache/object.go
@@ -208,8 +208,10 @@ func (o *Object) SetModTime(t time.Time) error {
 
 // Open is used to request a specific part of the file using fs.RangeOption
 func (o *Object) Open(options ...fs.OpenOption) (io.ReadCloser, error) {
-	if err := o.refreshFromSource(true); err != nil {
-		return nil, err
+	if o.Object == nil {
+		if err := o.refreshFromSource(true); err != nil {
+			return nil, err
+		}
 	}
 
 	var err error
--- a/backend/cache/object.go
+++ b/backend/cache/object.go
@@ -208,13 +208,17 @@ func (o *Object) SetModTime(t time.Time) error {
 
 // Open is used to request a specific part of the file using fs.RangeOption
 func (o *Object) Open(options ...fs.OpenOption) (io.ReadCloser, error) {
+	var err error
+
 	if o.Object == nil {
-		if err := o.refreshFromSource(true); err != nil {
-			return nil, err
-		}
+		err = o.refreshFromSource(true)
+	} else {
+		err = o.refresh()
+	}
+	if err != nil {
+		return nil, err
 	}
 
-	var err error
 	cacheReader := NewObjectHandle(o, o.CacheFs)
 	var offset, limit int64 = 0, -1
 	for _, option := range options {


ncw:
I'm not sure this is worth doing at most it will save a few cycles and stop contention on the refreshMutex.  Have you measured an improvement with this patch?

dcpu:
rclone always sends an info request to the source remote, regardless of whether the chunk is already cached.
Therefore, access to the cached chunk is delayed by latency between source remote.
This patch prevents this. The difference is noticable when you access already cached chunks.

dcpu:
I wrote a test program for random read performance benchmarks.
Mount cache remote and run 

randtest [file to cache mount]

```
package main

import (
	"log"
	"math/rand"
	"os"
	"time"
)

func check(e error) {
	if e != nil {
		panic(e)
	}
}

func main() {
	var err error
	b := make([]byte, 16384)
	fn := os.Args[1]
	f, err := os.Open(fn)
	check(err)
	fi, err := f.Stat()
	check(err)
	log.Printf(fn)

	start := time.Now()
	fs := fi.Size()
	for i := 0; i < 30; i++ {
		cstart := time.Now()
		of := rand.Int63n(fs)
		_, err = f.Seek(of, 0)
		check(err)
		_, err = f.Read(b)
		check(err)
		celapsed := time.Since(cstart)
		log.Printf("offset %d took %s", of, celapsed)
	}
	f.Close()
	elapsed := time.Since(start)
	log.Printf("took %s", elapsed)
}
```

Here is my results from 140M file on gdrivecache mount.

Original

1m 8s non cached
42s cached

With #2493

42s non cached
12s cached

With #2493 #2495

32s non cached
0.6s cached

ncw:
Thanks for doing the test :-)

I was looking at the source of `refreshFromSource` - I can see I mis-read it now and your change does make a difference.

So it looks like your change is equivalent to calling `o.refreshFromSource(false)` rather than `o.refreshFromSource(true)` - would you agree?

```go
// refreshFromSource requests the original FS for the object in case it comes from a cached entry
func (o *Object) refreshFromSource(force bool) error {
	o.refreshMutex.Lock()
	defer o.refreshMutex.Unlock()
	var err error
	var liveObject fs.Object

	if o.Object != nil && !force {
		return nil
	}
```

@remusb - can you help with this?  Is this sensible or will we miss updates from the source?

dcpu:
Yes. They are equivalent.

ncw:
@B4dM4n could you take a look at this change?

B4dM4n:
This looks good to me.
Using `o.refresh()` is the right way to check for expiration when `o.Object != nil`.

What I didn't figure out yet is why this change impacts seek performance. `Object.Open` should only be called once, but it looks like it is called on every seek.

I will investigate this further.

B4dM4n:
Got it, `cache.Object.Open` returns a `readers.NewLimitedReadCloser` not a `cache.Object.Handle` as I assumed.

This change definitely improves the cache performance.

ncw:
Thanks for your review @B4dM4n 

I'll merge this now then - thank you very much @dcpu for the contribution.

PR #2497, opened 2018-08-27
s3: add support for KMS Key ID - fixes #2217

### LINKED ISSUE #2217: KMS for S3 file creation
An issue was reported on the [forum](https://forum.rclone.org/t/kms-for-s3-file-creation/5255) where a KMS policy is preventing rclone from working with AWS S3. The policy reported as causing the issue was: 

```
{  
   “Version”:“2012-10-17”,
   “Id”:“PutObjPolicy”,
   “Statement”:[  
      {  
         “Sid”:“DenyUnEncryptedObjectUploads”,
         “Effect”:“Deny”,
         “Principal”:"",
         “Action”:         “s3:PutObject”,
         “Resource”:"arn:aws:s3:::YourBucket/",
         “Condition”:{  
            “StringNotEquals”:{  
               “s3:               x-amz-server-side-encryption”:               “aws:kms”
            }
         }
      }
   ]
}
```

I am submitting the issue on behalf of the original reporter and do not have answers to the standard issue questions.

It appears rclone needs to support the `--server-side-encryption aws:kms` and `--sse-kms-key-id <value>` flags as available in the AWS S3 CLI. See [AWS S3 Sync](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html) for an example.

I’m not 100% sure, but a quick look at the SDK looks like the s3crypto service may contain what’s needed to implement this: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3crypto/

Good general AWS documentation about S3 and KMS encryption is available at:
- https://docs.aws.amazon.com/kms/latest/developerguide/services-s3.html#sse
- https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html
- https://aws.amazon.com/blogs/security/how-to-prevent-uploads-of-unencrypted-objects-to-amazon-s3/

bsteiss:
s3: add support for KMS Key ID

This code should add support aws:kms and the kms key id to the backend/s3.go.
From my testing this has worked in my testing environment.
The config command seems to reflect the new options.

Fixes #2217


ncw:
That looks good!  Can' you drop the test logs from your commit and `go fmt` the files and force push.  The CI should pass then I think!

ncw:
Run `go fmt backend/s3/s3.go` commit and that should fix the CI

bsteiss:
So it looks like it's failing because I am using SSEKMSKeyId instead of SSEKMSKeyID.  However when I build rclone with make using SSEKMSKeyID it fails.  Not sure what you need me to do.

backend/s3/s3.go
```diff
@@ -569,6 +583,7 @@ type Options struct {
 	LocationConstraint   string        `config:"location_constraint"`
 	ACL                  string        `config:"acl"`
 	ServerSideEncryption string        `config:"server_side_encryption"`
+	SSEKMSKeyId          string        `config:"sse_kms_key_id"`
```
ncw: If you change this one to `SSEKMSKeyID` then the linter should be satisfied (cross fingers!).

backend/s3/s3.go
```diff
@@ -1429,6 +1444,9 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
 	if o.fs.opt.ServerSideEncryption != "" {
 		req.ServerSideEncryption = &o.fs.opt.ServerSideEncryption
 	}
+	if o.fs.opt.SSEKMSKeyId != "" {
+		req.SSEKMSKeyId = &o.fs.opt.SSEKMSKeyId
```
ncw: You need to leave `req.SSEKMSKeyId` alone and the linter shouldn't complain about it since we didn't implement it.

bsteiss:
That looks like it worked.  Thanks for the help I'm not a developer so I was totally out of my league doing this.

ncw:
That looks great now - thanks :-)

I'll merge it now!

Issue #2407, opened 2018-07-20
cache: pre-cache Plex On Deck content


drpau:
Considering plex integration is available albiet in its infancy, can we look at seeing how to have rclone get an upcoming list of what is "On Deck" and download / pre-cache this data?

Similar/Same idea to the unionfs script i have seen floating around which does much the same thing, but it requires a unionfs mount and the script to run.

Thanks

ncw:
@remusb?

B4dM4n:
Implementing more features based on Plex seems like a bad idea to me. This would increase the complexity of the `cache` even more would force us to keep up to date with the Plex development to keep this feature working. `rclone` should remain to be a common tool for working with storage providers. Besides that, there are other media solutions like Plex (for example Emby) which could also benefit from such a feature.

How about a new `rc` command to trigger caching of specific files:
```
rclone rc cache/fetch chunks=10 file=movie1.mkv file2=movie2.mkv
```
This could would keep `rclone` general and provide a way for external scripts/tools to fill the cache.

drpau:
@jkaberg fair point, im sure rclone in all efforts are trying to improve this just as continued development.  This isnt particarly for load times, as a lof o people who use rclone may just have poor bandwidth.  This saves the use of bandwidth being used to watch media, rather than having the prepopulated (eg when lots of people at home at night, bandwidth is important, but when its not being used during the day, bandwidth can be used).

@B4dM4n i hear your point, but there is already plex integration.  This service could be dependant on if plex integration is active, then action it.  Your idea of doing a combined call for scripts isnt a bad option, but this could just be done at the OS level rather than integration into rclone.  My thoughts would be that if you are going to amend rclone, you would amend it for the integration to get it done right rather than dependant on something external.  just my opinion but.

drpau:
@remusb @ncw - any chance of this being considered?
Appreciate may be a bit of work to do - but............ would be great to have this done.

memnos:
This feature would be great.
Let's hope to have an answer.

B4dM4n:
The command is not merged yet. We need to wait for a review of the code.

If you want to try it now, you would have to install Go and compile rclone manually.

memnos:
Ok, tnx

memnos:
I think now command has been committed.
I tried to run the .py file after changing settings and got no errors.
But how i know it is working?
What is it caching exactly?

B4dM4n:
This could be my bad. I messed up the error return value of the fetch command.
[...]

PR #2506, opened 2018-08-30
 cache: add cache/fetch rc method

B4dM4n:
This change adds a `cache/fetch` rc method to allow caching of specific file chunks.
The command is inspired by the feature request in #2407.

An example request to fetch the first 10 and the last chunk looks like this:
```
rclone rc cache/fetch chunks=:10,-1 file=hello file2=home/goodbye
```

The syntax for the `chunks` parameter is based on the Python slice syntax. It allows indexing from the beginning of the file using positive indices like `:10` and from the back with negative indices like `-10:`.

### DIFF
--- a/backend/cache/cache.go
+++ b/backend/cache/cache.go
@@ -6,10 +6,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"os/signal"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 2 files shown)


ncw:
Do you think these are best left until after 1.43 which I plan for the weekend?

B4dM4n:
In theory those two commits are unrelated and will work independently.

The rc method itself is not a complex feature. The hardest part to get right was parsing the index list. Otherwise it only creates a object handle and requests the needed chunks. Without the second commit, it just fetches `opt.TotalWorkers` additional chunks.

During my tests I noticed that `scaleWorkers(1)` was not working as intended. When scaling down from more than one worker, the workers get shut down by sending `r.preloadQueue <- -1`, but the workers never got deleted from the `r.workers` list. This was not a big issue by itself, but many methods relied on `len(r.workers)` to determine the number of running workers, including `Handle.queueOffset` where I noticed it in the logs. So one worker had to process `len(r.workers)` chunks every time `Handle.queueOffset` was called. Calling `scaleWorkers()` more time would make the situation worse, as it also used `len(r.workers)`, but this does not happen currently. `scaleWorkers()` is only used to increase the number of workers.
Correcting this also fixed what @dcpu intended to improve with #2493.

I tried to replicate the old behavior as close as possible, but if you don't want to take the risk of breaking something, this should wait until 1.43 as the current implementation seems to work.

ncw:
1.43 is out and I'm happy for you to merge these now :-)

I've given them the once over.

PR #2508, opened 2018-08-30
stats: handle FatalError and NoRetryError when reported to stats

connesc:
This is a proposal to properly handle `FatalError` and `NoRetryError` when they are not directly returned by the command, but just reported to stats.

See https://github.com/ncw/rclone/pull/2488#issuecomment-416675577 for the addressed use-case.

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/ncw/rclone/fs"
+	"github.com/ncw/rclone/fs/fserrors"
 	"github.com/ncw/rclone/fs/rc"
 )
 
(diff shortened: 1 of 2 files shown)


fs/accounting/stats.go
```diff
@@ -234,13 +241,25 @@ func (s *StatsInfo) String() string {
 	)
 
 	if !fs.Config.StatsOneLine {
+		errorFlags := make([]string, 0, 2)
+		if s.fatalError {
+			errorFlags = append(errorFlags, "fatal")
```
ncw: I'm not sure appending "fatal" and "noretry" has enough information for the user to make sense of what is going on.

Maybe we should tell the user explicitly what is going to happen, so "Fatal error detected" or "Not retrying" (but only if there are only NoRetry errors)

connesc:
OK, got it, thanks for the clarification!

I was actually thinking that if at least one NoRetryError happen, then there is no chance that another attempt can succeed. Now I understand that the purpose of retries is not necessarily to succeed, but to go as far as possible. This makes more sense because otherwise, NoRetryError would have been essentially the same as FatalError.

I will fix this :wink:

connesc:
Well, I have turned the !NoRetryError logic into a simple `retryError` flag, which behaves similarly to the `fatalError` flag. I think it helps keeping things simple.

I have also improved the formatted error details as suggested.
Let me know if it's better now.

ncw:
Apologies for the delay!

I think that looks great now thank you - I will merge it now :-)

PR #2509, opened 2018-08-31
Add SetTier command 

sandeepkru:
Added new command `settier` which changes object's tier on remote if supported, added optional interfaces to `SetTierer` and `GetTierer` to support this feature.

### DIFF
--- a/cmd/settier/settier.go
+++ b/cmd/settier/settier.go
@@ -1,0 +1,54 @@
+package settier
+
+import (
+	"github.com/ncw/rclone/cmd"
+	"github.com/ncw/rclone/fs/operations"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	cmd.Root.AddCommand(commandDefintion)
+}
+
+var commandDefintion = &cobra.Command{
+	Use:   "settier tier remote:path",
+	Short: `Changes storage class/tier of objects in remote.`,
+	Long: `
+rclone settier changes storage tier or class at remote if supported.
+Few cloud storage services provides different storage classes on objects,
+for example AWS S3 and Glacier, Azure Blob storage - Hot, Cool and Archive,
+Google Cloud Storage, Regional Storage, Nearline, Coldline etc.
+
+Note that, certain tier chages make objects not available to access immediately.
+For example tiering to archive in azure blob storage makes objects in frozen state,
+user can restore by setting tier to Hot/Cool, similarly S3 to Glacier makes object
+inaccessible.true
+
+You can use it to tier single object
+
+    rclone settier Cool remote:path/file
+
+Or use rclone filters to set tier on only specific files
+
+	rclone --include "*.txt" settier Hot remote:path/dir
+
+Or just provide remote directory and all files in directory will be tiered
+
+    rclone settier tier remote:path/dir
+`,
+	Run: func(command *cobra.Command, args []string) {
+		cmd.CheckArgs(2, 2, command, args)
+		tier := args[0]
+		input := args[1:]
+		fsrc := cmd.NewFsSrc(input)
+		cmd.Run(false, false, command, func() error {
+			isSupported := fsrc.Features().SetTier
+			if !isSupported {
+				return errors.Errorf("Remote %s does not support settier", fsrc.Name())
+			}
+
+			return operations.SetTier(fsrc, tier)
+		})
+	},
+}
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -177,6 +177,32 @@ func parsePath(path string) (container, directory string, err error) {
 	return
 }
 
+// validateAccessTier checks if azureblob supports user supplied tier
+func validateAccessTier(tier string) bool {
+	switch tier {
+	case string(azblob.AccessTierHot),
+		string(azblob.AccessTierCool),
+		string(azblob.AccessTierArchive):
+		// valid cases
+		return true
+	default:
+		return false
+	}
+}
+
+// validAccessTiers returns list of supported storage tiers on azureblob fs
+func validAccessTiers() []string {
+	validTiers := [...]azblob.AccessTierType{azblob.AccessTierHot, azblob.AccessTierCool,
+		azblob.AccessTierArchive}
+
+	var tiers [len(validTiers)]string
+
+	for i, tier := range validTiers {
+		tiers[i] = string(tier)
+	}
+	return tiers[:]
+}
+
 // retryErrorCodes is a slice of error codes that we will retry
 var retryErrorCodes = []int{
 	401, // Unauthorized (eg "Token has expired")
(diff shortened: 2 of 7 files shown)


sandeepkru:
@ncw - I duplicated some test code in `InternalTest` at `azureblob` remote. Please let me know if I need to remove this and add part of that test as separate integration test? Otherwise, I think, I've all basic working parts as you suggested except enhancements to `lsjson` and other `list` commands.

cmd/settier/settier.go
```diff
@@ -0,0 +1,54 @@
...
+`,
+	Run: func(command *cobra.Command, args []string) {
+		cmd.CheckArgs(2, 2, command, args)
+		tier := args[0]
+		input := args[1:]
+		fsrc := cmd.NewFsSrc(input)
+		cmd.Run(false, false, command, func() error {
+			isSupported := fsrc.Features().SetTier
+			if !isSupported {
+				return errors.Errorf("Remove %s does not support settier", fsrc.Name())
```
ncw: That should say "remote" not "remove"

ncw:
> I duplicated some test code in InternalTest at azureblob remote. Please let me know if I need to remove this and add part of that test as separate integration test?

I think that if we are to support tiers in other remotes, then having those tests as part of the integration test suite would be the best way to go.  The backend would need to indicate which tiers should be tested in fstests.Opt - if these were empty then the tests should be skipped.

To make the SetTier/GetTier more general purpose, do you think we need a ListTiers?  Then the user could run `rclone settier remote:` and it would print the possible tiers?

> Otherwise, I think, I've all basic working parts as you suggested except enhancements to lsjson and other list commands.

I think what you've done is excellent.

Let's just tidy this lot up and merge it before we do anything else.

I'm planning on making the 1.43 release over the weekend, so I won't merge this before then as I'm trying to stabilise the release right now.

Would you like to be an rclone maintainer?  Maybe with responsibility for the azureblob backend?  If so drop me an email to nick@craig-wod.com and we can discuss.

sandeepkru:
> I think that if we are to support tiers in other remotes, then having those tests as part of the integration test suite would be the best way to go. The backend would need to indicate which tiers should be tested in fstests.Opt - if these were empty then the tests should be skipped.

Yup, this is better approach, I'll change and send out new iteration today.

> To make the SetTier/GetTier more general purpose, do you think we need a ListTiers? Then the user could run rclone settier remote: and it would print the possible tiers?

This is definitely more flexible and users will make less mistakes also this way. Thanks for the suggestion, will get this done in next iteration.

> I'm planning on making the 1.43 release over the weekend, so I won't merge this before then as I'm trying to stabilise the release right now.

Thanks for the heads up, and agree its better to get that stabilized than merging any new changes

> Would you like to be an rclone maintainer? Maybe with responsibility for the azureblob backend? If so drop me an email to nick@craig-wod.com and we can discuss.

That would be great opportunity, I can share responsibility for azureblob, will drop an email.

Thanks for all suggestions ! Will send out new iteration today.

ncw:
@sandeepkru the 1.43 release is now out  - hooray :-)

Have you got some more things to add to this PR?

I haven't seen an email from you yet?

-- Nick

sandeepkru:
Sorry, I was away and didn't get a chance to work/contact on this. Planning to work on this today.

ncw:
No worries!  I've been very busy too.

sandeepkru:
> To make the SetTier/GetTier more general purpose, do you think we need a ListTiers? Then the user could run rclone settier remote: and it would print the possible tiers?

@ncw - Adding `ListTiers` is making slightly convoluted, current implementation of `SetTier/GetTier` are on optional at object level (as suggested earlier), ideally `ListTiers` do we need to do at `Fs` instead of object level, this is where its getting into confusing state, having `SetTier/GetTier` at object level and this one at `Fs` level.
If we make at object level, I don't really see a clean implementation ( choose one object from `Fs`(?), what if there are no objects?) 

Please let me know your thoughts, if you agree to exclude `ListTiers`, I'll refactor internal test and integration test, and will send out new iteration so that we can close this. Thanks !

ncw:
OK let's not do `ListTiers` for the moment and concentrate on getting this merged.

If you could re-factor the integration tests that would be perfect.

You or I can then add an implementation for s3 and gcs...

sandeepkru:
> OK let's not do ListTiers for the moment and concentrate on getting this merged.

I still added `ListTiers` to make integration tests easy and hoping we can extended if required.

> If you could re-factor the integration tests that would be perfect.

Added an integration test and removed all unnecessary things introduced at azureblob internal test.

> You or I can then add an implementation for s3 and gcs...

Sounds good, I can give a try on S3 and GCS, once we close this.

sandeepkru:
@ncw - Can you please take a look at latest changes if you get a chance.

ncw:
I'm not 100% happy with `ListTiers` on Object, I think it would be better on `Fs`, but I'm happy to merge this and iterate later. It certainly works for the test suite.

Also I think that trying to set all the tiers on the object is likely to fail on some remotes so we might need to configure that.

Anyway this code does the job, so I'll merge it and we can see what the implementation for GCS/S3 looks like which might guide the design more :-)

ncw:
I noticed that the integraton tests failed for SetTier this morning 

https://pub.rclone.org/integration-tests/current-test.txt

```
    --- FAIL: TestIntegration/TestSetTier (0.04s)
        fstests.go:1083: 
            	Error Trace:	fstests.go:1083
            	Error:      	Expected nil, but got: Failed to set Blob Tier: -> github.com/ncw/rclone/vendor/github.com/Azure/azure-storage-blob-go/2018-03-28/azblob.NewResponseError, /home/rclone/go/src/github.com/ncw/rclone/vendor/github.com/Azure/azure-storage-blob-go/2018-03-28/azblob/zz_generated_response_error.go:28
            	            	===== RESPONSE ERROR (ServiceCode=InvalidHeaderValue) =====
            	            	Description=The value for one of the HTTP headers is not in the correct format.
            	            	RequestId:1a0ced0d-d01e-0045-6d1e-4b4785000000
            	            	Time:2018-09-13T05:01:05.8884209Z, Details: 
            	            	   Code: InvalidHeaderValue
            	            	   HeaderName: x-ms-access-tier
            	            	   HeaderValue: Archive
            	            	   PUT https://rclone.blob.core.windows.net/rclone-test-mikomuy3wuqoquz5quhucim0/file%20name.txt?comp=tier&timeout=61
            	            	   Authorization: REDACTED
            	            	   User-Agent: [Azure-Storage/0.1 (go1.11; linux)]
            	            	   X-Ms-Access-Tier: [Archive]
            	            	   X-Ms-Client-Request-Id: [b004e25a-4815-4951-4605-0f6cc414089a]
            	            	   X-Ms-Date: [Thu, 13 Sep 2018 05:01:05 GMT]
            	            	   X-Ms-Version: [2018-03-28]
            	            	   --------------------------------------------------------------------------------
            	            	   RESPONSE Status: 400 The value for one of the HTTP headers is not in the correct format.
            	            	   Content-Length: [329]
            	            	   Content-Type: [application/xml]
            	            	   Date: [Thu, 13 Sep 2018 05:01:05 GMT]
            	            	   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
            	            	   X-Ms-Error-Code: [InvalidHeaderValue]
            	            	   X-Ms-Request-Id: [1a0ced0d-d01e-0045-6d1e-4b4785000000]
            	            	   X-Ms-Version: [2018-03-28]
            	            	
            	            	
            	Test:       	TestIntegration/TestSetTier
        fstests.go:1085: 
            	Error Trace:	fstests.go:1085
            	Error:      	Not equal: 
            	            	expected: "Archive"
            	            	actual  : "Cool"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-Archive
            	            	+Cool
            	Test:       	TestIntegration/TestSetTier
```

Is that because my account isn't capable of doing `Archive`?  We probably only need to swap betwen Hot and Cool anyway to show that it works - we could configure that in the fstests.Opt?

sandeepkru:
> Is that because my account isn't capable of doing Archive? 

[Tiering Support](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers#storage-accounts-that-support-tiering) - As per this - `You may only tier your object storage data to hot, cool, or archive in Blob storage or General Purpose v2 (GPv2) accounts. General Purpose v1 (GPv1) accounts do not support tiering. However, customers can easily convert their existing GPv1 or Blob storage accounts to GPv2 accounts through a simple one-click process in the Azure portal.`

> We probably only need to swap betwen Hot and Cool anyway to show that it works - we could configure that in the fstests.Opt?

Agree, I'll fix integration test.

PR #2511, opened 2018-09-01
Remove leftover files

toddy15:
Hi,

you might want to remove those two files which seem to be leftover after a merge conflict.


ncw:
Thank you for that - a nice tidy up :-)

Issue #2514, opened 2018-09-02
Avoid creation of directory %USERPROFILE%\.config\rclone when using --config or RCLONE_CONFIG


albertony:
I'm trying out version 1.43 on Windows, and want to specify a custom location for the config file (to run it as "stealth" as possible). I can then use command line argument `--config=<configfile>` or environment variable `set RCLONE_CONFIG=<configfile>`. But even if I do, rclone creates an empty directory `%USERPROFILE%\.config\rclone`. I find that I have to specify environment variable XDG_CONFIG_HOME to avoid this. When looking into the source file `fs/config/config.go` and function `makeConfigPath`, I'm wondering if this could be avoided with a minor code change: At [line 121](https://github.com/ncw/rclone/blob/c6c74cb86992303fec6584c6e0561a0efae1a849/fs/config/config.go#L121-L128) it sets local variables xdgdir and xdgcfgdir based on the value of homedir, and then at [line 151](https://github.com/ncw/rclone/blob/c6c74cb86992303fec6584c6e0561a0efae1a849/fs/config/config.go#L151-L157) it creates the actual folder path given by xdgcfgdir, before at line [167](https://github.com/ncw/rclone/blob/c6c74cb86992303fec6584c6e0561a0efae1a849/fs/config/config.go#L167-L173) it checks for `--config` or `RCLONE_CONFIG`.. When user has specified `--config` or `RCLONE_CONFIG`, should this take priority so the logic searching for home and xdgdir can then just be skipped?

ncw:
> When user has specified --config or RCLONE_CONFIG, should this take priority so the logic searching for home and xdgdir can then just be skipped?

Yes it should.

Would you like to send a pull request with this change?

albertony:
Sure, I am not very familiar with Go, but I can try..

albertony:
[...]

PR #2515, opened 2018-09-02
Avoid creating empty default configuration directory when user supplies path to config file

albertony:
Fixes issue #2514.

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -153,7 +153,7 @@ func makeConfigPath() string {
 	// variable.  We can't use pflag for this because it isn't initialised
 	// yet so we search the command line manually.
 	_, configSupplied := os.LookupEnv("RCLONE_CONFIG")
-	if (!configSupplied) {
+	if !configSupplied {
 		for _, item := range os.Args {
 			if item == "--config" || strings.HasPrefix(item, "--config=") {
 				configSupplied = true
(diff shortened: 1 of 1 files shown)


albertony:
Do you understand the Travis build failed, because I don't?

albertony:
One of the Travis builds failed in the test WebDav test, but I don't see that this can be caused by my changes..

ncw:
I've given it a rebuild and we'll see what happens!

fs/config/config.go
```diff
@@ -147,38 +149,43 @@ func makeConfigPath() string {
...
+			return cfgpath
+		}
 	}
+
+	// Assume .rclone.conf can be written to user's home directory.
+	if homeconf != "" {
+		return homeconf
+	}
+
+	// Default to ./.rclone.conf (current working directory) if everything else fails.
```
ncw: I think this needs to be bracketed in `if !configSupplied {` otherwise if homeconf is "" but --config is supplied it still outputs the message.

You can simulate this quite easily by compiling without cgo `CGO_ENABLED=0 go install` and `unset HOME` - if you supply `--config` then rclone still gives the error.

albertony:
You are right, I reproduced it and fixed it. Thanks.

ncw:
> You are right, I reproduced it and fixed it. Thanks.

I'll merge that now - thank you very much :-)

> Edit: The same WebDav test failed again..

Grrr, have made an issue about this #2517 - I haven't managed to get it to fail locally though which is annoying.

PR #2526, opened 2018-09-04
cache: documentation fix - Fixes #2519

### LINKED ISSUE #2519: --cache-total-chunk-size= missing in v1.43
Rclone doesn't seem to recognize --cache-total-chunk-size= in v1.43 when using rclone mount. It worked in v1.42. I am not sure if this was done on purpose, but it was not mentioned in the changelogs and the docs still mention it.

v1.43

Tested on Ubuntu server and Windows 10, 64 bit

Google Drive

rclone mount 

It immediately fails to mount with an unknown flag "--cache-total-chunk-size=10G"

darthShadow:
Also, added a minor fix for code format in `CONTRIBUTING.md`

Signed-off-by: Anagh Kumar Baranwal <anaghk.dos@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ Make sure you
 
 When you are done with that
 
-  git push origin my-new-feature
+    git push origin my-new-feature
 
 Go to the Github website and click [Create pull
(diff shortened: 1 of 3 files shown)


ncw:
Nice one - thank you very much :-)

PR #2530, opened 2018-09-04
Implemented PublicLink support for Jottacloud

albertony:
Title basically says it all..
Did not see any functionality in rclone for removing the public link again? That would be simple to implement for Jottacloud (just set mode "disableShare" instead of "enableShare")..

### DIFF
--- a/backend/jottacloud/api/types.go
+++ b/backend/jottacloud/api/types.go
@@ -233,16 +233,17 @@ GET http://www.jottacloud.com/JFS/<account>/<device>/<mountpoint>/.../<file>
 
 // JottaFile represents a Jottacloud file
 type JottaFile struct {
-	XMLName    xml.Name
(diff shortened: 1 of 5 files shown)


albertony:
Squashed commits

ncw:
Code looks great - thank you.

Can you rebase this on master please which will remove the conflict.

And if you haven't already can you check the integration test works

    cd backend/jottacloud
    go test -v

Thank you :-)

albertony:
Now... (Hope I did it correct, as I'm not that experienced with Git)

ncw:
I'm not super keen on the `--jottacloud-unlink` flag, but I'm going to merge as-is since I can't think of anything better :-)

Thank you for this - very useful.

albertony:
I can agree with that. unlink as separate command would probably be cleanest, but overkill if only one or very few backends support it..

PR #2540, opened 2018-09-05
S3: Use (custom) pacer, to retry operations when reasonable - fixes #2503

### LINKED ISSUE #2503: s3 add pacer to retry directory listings
When doing large (75M files) copies to AWS S3 (region is in another country, with ~30ms latency and occasionally slightly glitchy networks), rclone will report "error reading destination directory: SerializationError: failed to decode REST XML response caused by: unexpected EOF".  It's not wrong, but it is just a transient error and a retry would very likely succeed and allow the whole operation to succeed cleanly

1.42

Ubuntu 16.04 64 bit

AWS S3

rclone copy --retries 1 --size-only --immutable --ignore-checksum --max-delete 0 --no-gzip-encoding /some/local/dir aws-${REGION}:somebucketname

Sorry, it'd be huge.  The relevant lines are:
2018/08/30 08:08:06 ERROR : <redacted directory name>: error reading destination directory: SerializationError: failed to decode REST XML response
caused by: unexpected EOF

It seems to me that making the directory listing code use low-level retries might help, or else making S3 use Pacer to handle retries.  But my Go isn't strong enough yet to help write a patch .  Happy to test any potential fixes in real-world scenarios, if it'd help.

craig-miskell-fluxfederation:
I can confirm this works well in my situation, fixing the transient problems I was having.  

I'm not 100% sure I've got the right Go idiom in all cases, and if not I'd be very happy to be shown the error of my ways


lib/pacer/pacer.go
```diff
@@ -309,6 +317,42 @@ func (p *Pacer) drivePacer(retry bool) {
...
+//
+// Call with p.mu held
+func (p *Pacer) s3Pacer(retry bool) {
+	oldSleepTime := p.sleepTime
+	if retry {
+		if p.attackConstant == 0 {
+			p.sleepTime = p.maxSleep
+		} else {
+			if p.sleepTime == 0 {
+				p.sleepTime = 1
```
ncw: I think this should be p.sleepTime = p.minSleep

I think minSleep should be something like `10ms` for the amazon pacer so we sleep for 10ms minimum on an error, and that is increased exponentially.

The constant `1` is 1 nS so not a lot of sleeping, and still only 1 second after 30 retries!

lib/pacer/pacer.go
```diff
@@ -309,6 +317,42 @@ func (p *Pacer) drivePacer(retry bool) {
...
+		}
+		if p.sleepTime > p.maxSleep {
+			p.sleepTime = p.maxSleep
+		}
+		if p.sleepTime != oldSleepTime {
+			fs.Debugf("pacer", "Rate limited, increasing sleep to %v", p.sleepTime)
+		}
+	} else {
+    p.sleepTime = (p.sleepTime<<p.decayConstant - p.sleepTime) >> p.decayConstant
+		if p.sleepTime != oldSleepTime {
```
ncw: This should probably have this I think.

```go
		if p.sleepTime < p.minSleep {
			p.sleepTime = 0
		}
```

ncw:
Note that one of my comments is hidden by default so you'll need to click the "show outdated" button to see it!  I think that must be because I commented on the changeset that was updated by the whitespace fixes.

craig-miskell-fluxfederation:
You make some excellent points.  I implemented a test method for S3Pacer, and it promptly showed their correctness.  I also took one of them a small step further (in the retry case, making the attack increase occur only on the second retry; the first retry just takes us to minSleep).

Ran some short tests, and it seems to be behaving.  I also finally read CONTRIBUTING.md, and have squashed commits etc for a better review experience.  Apologies for earlier mess.

ncw:
The integration tests are failing because the code needs to run through `go fmt`

If you can fix that then I'll do a final review before merging :-)

craig-miskell-fluxfederation:
Heh, just got into the office and noticed that.  Pushed fixes (and learned that I can use 'go fmt', rather than having to get my tabs right with vim :))

ncw:
This looks great now - thank you very much :-)

Issue #2532, opened 2018-09-04
[Feature request] Support --fast-list with Jottacloud


albertony:
By adding query parameter "mode=list" to the API request for a folder, you get all files and subfolders recursive, as a filedirlist XML. Is this what is needed to implement the ListR function, enabling the --fast-list flag?

ncw:
Yes that is exactly what is needed for `ListR` and `--fast-list` support :-)

How does it work if you have 1,000,000 files - does it page the listing?

Would you like to have a go at this?

albertony:
It does not page the results, as far as I know. I don't know if this means we have to do some kind of stream parsing for large XMLs, are there other backends doing something like that?

ncw:
All the other backends use some sort of paging mechanism to give you 1,000 files in a chunk.

The XML parser in the REST module does do streaming though, but the result will be passed back in-memory.

Then again for `--fast-list` it does explicitly state in the help that it will use more memory.

I'd say implement it and see - we can always write caveats in the docs!

PR #2545, opened 2018-09-06
Implemented --fast-list for Jottacloud (#2532)

albertony:
Basic implementation. No paging mechanism, since the API method does not support it to my knowledge. Not tested for all kinds of operations, but ls and tree seem to work.

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ncw/rclone/fs/fserrors"
 	"github.com/ncw/rclone/fs/fshttp"
 	"github.com/ncw/rclone/fs/hash"
+	"github.com/ncw/rclone/fs/walk"
 	"github.com/ncw/rclone/lib/pacer"
 	"github.com/ncw/rclone/lib/rest"
 	"github.com/pkg/errors"
@@ -227,9 +228,14 @@ func errorHandler(resp *http.Response) error {
 	return errResponse
 }
 
+// filePathRaw returns an unescaped file path (f.root, file)
+func (f *Fs) filePathRaw(file string) string {
+	return path.Join(f.endpointURL, replaceReservedChars(path.Join(f.root, file)))
+}
+
 // filePath returns a escaped file path (f.root, file)
 func (f *Fs) filePath(file string) string {
-	return rest.URLPathEscape(path.Join(f.endpointURL, replaceReservedChars(path.Join(f.root, file))))
+	return rest.URLPathEscape(f.filePathRaw(file))
 }
 
 // filePath returns a escaped file path (f.root, remote)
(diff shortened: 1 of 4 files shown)


albertony:
Whats this error?

type listFileDirFn func(fs.DirEntry) error

 fs.Object

ncw:
This is looking good!  I gave the code a once over and I couldn't see any obvious problems with it :-)

----

Can you add fast list to the integration tests here:

https://github.com/ncw/rclone/blob/57b9fff9047c680fa8737f2bc0c53a8423ed7727/fstest/test_all/test_all.go#L76-L78

This will mean --fast-list gets tested in the daily integration tests.  This should really be worked out automatically - I'm planning a revamp of the integration tests!

----

Can you try running the integration tests locally to check they pass for Jottacloud with `--fast-list`?

Running this should do it

    cd backend/jottacloud
    go test -v
    cd ../../fs/sync
    go test -v -remote TestJottacloud:
    go test -v -remote TestJottacloud: -fast-list
    cd ../../fs/operations
    go test -v -remote TestJottacloud:
    go test -v -remote TestJottacloud: -fast-list

(See writing a new backend [integration tests](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend) for more info)

----

Can you squash your code into two commits please, one for the `--fast-list` and one for the `About` changes.

----

Also can you put as the first line of your commits `jottacloud: ....`  That first line goes into making the changelog so you'll make my life easier if you put a user readable summary there - thanks! Here is the [commit message guide](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).

Thank you

Nick

albertony:
Thank you for your patience and guidance :+1:! I hope I have managed to do what you described.

The fs/operations test TestCopyFile fails on my machine, but then it does so also on latest master on my machine:

```
Listing wrong, want
  sub/file2 (14) or
  sub/file2 (14) got
  Not found "sub/file2"
```

ncw:
That looks great now - thanks :-)

I can see from the daily integration tests that that test is passing:  https://pub.rclone.org/integration-tests/

So probably it will work if you retry it again.  The `test_all` framework does retries as all remotes can be unreliable!

I'll merge this now - keep any eye out for failures in https://pub.rclone.org/integration-tests/current-summary.txt tomorrow!

albertony:
Seems there is a failure in the integration tests..
Althought the current-summary.txt indicates its with the PublicLink (which is not not supported on master):
```
FAIL	github.com/ncw/rclone/backend/jottacloud	45.815s
--- FAIL: TestIntegration (17.66s)
    --- FAIL: TestIntegration/TestPublicLink (0.26s)
```

The current-test.txt reports error from TestListRSubdir, which is more understandable:

```
--- FAIL: TestIntegration/TestFsListRSubdir (0.19s)
	fstests.go:505: 
			Error Trace:	fstests.go:505
										fstests.go:513
			Error:      	Not equal: 
							expected: "hello? sausage/êé/Hello, 世界/ \" ' @ < > & ? + ≠/z.txt"
							actual  : "z.txt"
							
							Diff:
							--- Expected
							+++ Actual
							@@ -1 +1 @@
							-hello? sausage/êé/Hello, 世界/ " ' @ < > & ? + ≠/z.txt
							+z.txt
			Test:       	TestIntegration/TestFsListRSubdir
	require.go:738: 
			Error Trace:	fstests.go:506
										fstests.go:513
			Error:      	"[]" should have 0 item(s), but has 1
			Test:       	TestIntegration/TestFsListRSubdir
```

Not exactly sure what this test does, but it seems it has created a subdir path with special characters, and tests that it gets the same path back. I tried the following on my local master:

```
rclone mkdir "jottatest:/Test/folder/hello? sausage/êé/Hello, 世界/ \" ' @ ^< ^> ^& ? + ≠"
rclone ls jottatest:/Test/folder
        0 hello? sausage/êé/Hello, 世界/ " ' @ < > & ? + ≠/z.txt
rclone ls jottatest:/Test/folder --fast-list
        0 hello? sausage/êé/Hello, 世界/ " ' @ < > & ? + ≠/z.txt
```
This seems to OK. But probably the test does something else/more. Do you have any pointers as to how I can run the actual test?

ncw:
> Althought the current-summary.txt indicates its with the PublicLink (which is not not supported on master):

They are a bit difficult to read the output of the tests - something I'm planning to fix soon!  Which test failed goes at the end so it is

```
--- FAIL: TestIntegration (45.79s)
    --- FAIL: TestIntegration/TestFsListRSubdir (0.19s)
FAIL
FAIL	github.com/ncw/rclone/backend/jottacloud	45.815s
```

----

To replicate just this fault run

    go test -v -run 'TestIntegration/Test(Setup|Init|FsMkdir|FsPutFile2|FsListRSubdir)$' 

(the backend tests are somewhat interdependent to keep them simple)

----

I see what is happening.  The path names returned are supposed to be relative to the remote passed in, not the dir passed into ListR.

So you are returning `z.txt` whereas we are expecting the full path from the root for the objects `hello? sausage/êé/Hello, 世界/ \" ' @ < > & ? + ≠/z.txt` even though we are listing the directory `hello? sausage/êé/Hello, 世界/ \" ' @ < > & ? + ≠`.

In rclone the `Remote()` of the objects are always relative to the root of the`Fs`.

ncw:
You won't be able to see this problem using the command line because it will only ever call ListR with dir="".

The problem is that `ListR("subdir", callback)` should return objects with `Remote()` giving

- subdir/file1
- subdir/file2

Not

- file1
- file2

Like it is at the moment.  I hope that makes sense!

albertony:
Ahh.. Slowly it is making sense.. :) Think I have fixed it now, created new pull request (#2558).

albertony:
Just for the record: I just found an alternative to the API method I used to implement --fast-list. I used `?mode=list` but there is also a `?mode=liststream` which gives much less verbose xml response: No indentation or line endings, lists only completed files (not deleted, corrupt or incomplete), less information about each file etc. Unfortunately it only works at the mountpoint level, which corresponds to the root level of the remote, not on any folder like the existing mode=list. And also since the xml format is different it does not work with the current implementation. If listing from remote root level is a common use case and it turns out to bee too slow, we could implemented the mode=liststream as an addition to the existing function (e.g. branch out into liststream mode when listing from root level, else use the list mode). Still no paging mechanism with liststream mode either, as far as I can tell.

ncw:
I've merged your PR now - thank you.

Noted about the listing from the root.  I would say it is probably less common to list from the root - people tend to organise their stuff in subdirectories.  There used to be a similar option for google drive but I took it out as I didn't want to maintain two listing mechanisms.

Issue #2538, opened 2018-09-05
rclone ncdu - Invalid argument


qntf:
#### What is the problem you are having with rclone?

running
`rclone ncdu myremote:`
causes:
```
2018/09/05 15:20:39 invalid argument
```
and fails.

all other `rclone` functions with that remote are working well, including the `rclone serve`.`

#### What is your rclone version (eg output from `rclone -V`)

```
rclone v1.43
- os/arch: linux/amd64
- go version: go1.11
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Arch Linux 64-bit

####  Which cloud storage system are you using? (eg Google Drive)
Google Drive
(also replicated with OneDrive)

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`rclone ncdu myremote:`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
2018/09/05 15:23:42 DEBUG : rclone: Version "v1.43" starting with parameters ["rclone" "-vv" "ncdu" "myremote:"]
2018/09/05 15:23:42 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2018/09/05 15:23:42 invalid argument
```

B4dM4n:
I'm closing this issue as the problem seems be resolved. The fixed error message should should be much clearer and point to the source of the problem.

Feel free to reopen this or create a new issue if the problem persists.

PR #2547, opened 2018-09-06
ncdu: return error instead of log.Fatal in Show

B4dM4n:
I don't think this `log.Fatal(err)` in ncdu is intentional, but better ask than revert later.

This makes a termbox error much more recognizable.
#2538 could be an error returned by termbox, which is simply a syscall error message.


ncw:
Good catch - thank you :-)

Issue #2517, opened 2018-09-03
Fix integration test for serve webdav


ncw:
This test has become unreliable on go1.11

```
2018/09/03 09:59:16 NOTICE: Config file "/notfound" not found - using defaults
2018/09/03 09:59:16 NOTICE: Local file system at /tmp/rclone656136016: WebDav Server started on http://127.0.0.1:51778/
2018/09/03 09:59:20 Error on serving HTTP server: http: Server closed
--- FAIL: TestWebDav (3.40s)
    webdav_test.go:75: 
        ----------
        2018/09/03 09:59:18 NOTICE: Config file "/notfound" not found - using defaults
        2018/09/03 09:59:18 ERROR : does not exist: error listing: directory not found
        2018/09/03 09:59:19 ERROR : : error listing: directory not found
        2018/09/03 09:59:20 ERROR : : error listing: directory not found
        --- FAIL: TestIntegration (1.84s)
            --- FAIL: TestIntegration/TestFsPutError (0.10s)
                fstests.go:400: 
                    	Error Trace:	fstests.go:400
                    	Error:      	Expected nil, but got: &webdav.Object{fs:(*webdav.Fs)(0xc000168c80), remote:"hello? sausage/êé/Hello, 世界/ \" ' @ < > & ? + ≠/z.txt", hasMetaData:true, size:4096, modTime:time.Time{wall:0x0, ext:63671565558, loc:(*time.Location)(0xc000267200)}, sha1:""}
                    	Test:       	TestIntegration/TestFsPutError
                fstests.go:401: 
                    	Error Trace:	fstests.go:401
                    	Error:      	Not equal: 
                    	            	expected: *errors.fundamental(object not found)
                    	            	actual  : <nil>(<nil>)
                    	Test:       	TestIntegration/TestFsPutError
        FAIL
        exit status 1
        FAIL	github.com/ncw/rclone/backend/webdav	1.847s
        ----------
    <autogenerated>:1: 
        	Error Trace:	webdav_test.go:77
        	Error:      	Received unexpected error:
        	            	exit status 1
        	Test:       	TestWebDav
        	Messages:   	Running webdav integration tests
FAIL
FAIL	github.com/ncw/rclone/cmd/serve/webdav	3.426s
```

Travis: https://travis-ci.org/ncw/rclone/jobs/423852271

ncw:
[...]

PR #2548, opened 2018-09-06
Implement optional permanent deletion in Jottacloud (not using trash)

albertony:
When deleting, the item will be moved to trash by default, as before, but now you can specify option `--jottacloud-use-trash=false` to permanently delete it instead.

PS: I used the same naming of the option as in Drive (`--drive-use-trash=false`), but it is kind of annoying to have such boolean arguments where the default is true so you must always set the `=false` value instead of just the `--option` , if you know what I mean. Mega has such a negated option `--mega-hard-delete`, but then that name wasn't 100% fit..

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -71,6 +71,11 @@ func init() {
 			Help:     "Files bigger than this will be cached on disk to calculate the MD5 if required.",
 			Default:  fs.SizeSuffix(10 * 1024 * 1024),
 			Advanced: true,
+		}, {
+			Name:     "hard_delete",
+			Help:     "Delete files permanently rather than putting them into the trash.",
+			Default:  false,
+			Advanced: true,
 		}},
 	})
 }
@@ -81,6 +86,7 @@ type Options struct {
 	Pass               string        `config:"pass"`
 	Mountpoint         string        `config:"mountpoint"`
 	MD5MemoryThreshold fs.SizeSuffix `config:"md5_memory_limit"`
+	HardDelete         bool          `config:"hard_delete"`
 }
 
 // Fs represents a remote jottacloud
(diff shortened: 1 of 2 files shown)


albertony:
Another mysterious Travis error.. Let me know if there is something I can do.

ncw:
> When deleting, the item will be moved to trash by default, as before, but now you can specify option --jottacloud-use-trash=false to permanently delete it instead.
> 
> PS: I used the same naming of the option as in Drive (--drive-use-trash=false), but it is kind of annoying to have such boolean arguments where the default is true so you must always set the =false value instead of just the --option , if you know what I mean. Mega has such a negated option --mega-hard-delete, but then that name wasn't 100% fit..

drive used to delete things permanently by default, but I got lots of user complaints about that so I had to change it, hence the strange double negative.

In subsequent remotes I used `--xxx-hard-delete` and that would be my preferred name for this new flag.

```
      --b2-hard-delete                           Permanently delete files on remote removal, otherwise hide files.
      --mega-hard-delete                         Delete files permanently rather than putting them into the trash.
```

I'm planning to make a `--hard-delete` option too at some point which will set this for all backends.

> Another mysterious Travis error.. Let me know if there is something I can do.

I've restarted the osx build and it seems to be building now.  The failure looked like some sort of glitch at travis (which aren't too uncommon alas).

albertony:
>In subsequent remotes I used --xxx-hard-delete and that would be my preferred name for this new flag.

Agreed, just pushed an update.

>I'm planning to make a --hard-delete option too at some point which will set this for all backends.

I was thinking the same (not that I planned to do it, but that it would be cleaner..)

>I've restarted the osx build and it seems to be building now. 

Thanks.

albertony:
Squashed commits

ncw:
Grrr!  There is a ticket about the webdav build #2517 which I haven't managed to solve yet.

I suspect the test is racy somehow - making the webdav object bigger will probably fix it.

Anyway, your code looks great and has passed the tests so I'll merge it now :-)

PR #2550, opened 2018-09-07
Update azure sdk to latest version

sandeepkru:
Azure SDK latest version fixes couple of bug fixes and adds transactional MD5 to stageblock API

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -6,7 +6,7 @@ bazil.org/fuse/fuseutil
 cloud.google.com/go/compute/metadata
 # github.com/Azure/azure-pipeline-go v0.0.0-20180607212504-7571e8eb0876
 github.com/Azure/azure-pipeline-go/pipeline
-# github.com/Azure/azure-storage-blob-go v0.0.0-20180712005634-eaae161d9d5e
+# github.com/Azure/azure-storage-blob-go v0.0.0-20180906215025-bb46532f68b7
 github.com/Azure/azure-storage-blob-go/2018-03-28/azblob
 # github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78
 github.com/Azure/go-ansiterm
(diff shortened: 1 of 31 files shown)


vendor/modules.txt
```diff
@@ -6,7 +6,7 @@ bazil.org/fuse/fuseutil
 cloud.google.com/go/compute/metadata
 # github.com/Azure/azure-pipeline-go v0.0.0-20180607212504-7571e8eb0876
 github.com/Azure/azure-pipeline-go/pipeline
-# github.com/Azure/azure-storage-blob-go v0.0.0-20180712005634-eaae161d9d5e
+# github.com/Azure/azure-storage-blob-go v0.0.0-20180906215025-bb46532f68b7
```
ncw: I see there is an official 0.2.0 release: https://github.com/Azure/azure-storage-blob-go/releases released 15 hours ago!

Do you think it is worth updating to that?

sandeepkru: Agree, that was my intention, will pull the right version.

ncw:
The diff of the go.mod still looks odd.  There are some strange extra dependencies and the version is still 0.0.0.  I have to say I'm still learning `go mod` so not sure why that is!

sandeepkru:
> The diff of the go.mod still looks odd. There are some strange extra dependencies and the version is still 0.0.0. I have to say I'm still learning go mod so not sure why that is!

I'm not yet successful pulling `v0.2.0` , I tried `go get package@version` and `go get package@sha1`, neither of those are successful, and I even tried `go mod edit -replace=old[@v]=new[@v] -dropreplace=old[@v]` no luck with this either. I'm also still trying to figure out `go mod`, not sure about additional strange dependencies.

ncw:
Well if you think you've got the new code then we should merge this as-is and worry about go mod later!

ncw:
OK. I'm happy for you to merge this when you are satisfied :-)

sandeepkru:
`go.mod` has some updates related to `tools` like `lint, mod, errcheck`. I don't know why,  as far as azure changes concerned, it pull all changes as part of `v.0.2.0` which is the `tip` today and locally tested changes, I believe its safe to merge.

ncw:
OK I'll merge this now then :-)

Issue #2215, opened 2018-04-05
Plex integration with HTTPS requires a valid certificate if securitymode=required


kelinger:
From this discussion:  [https://forum.rclone.org/t/is-this-normal-plex-websocket-eof/5240](url)

#### What is the problem you are having with rclone?

Basically, when using HTTPS because the Plex Server has "required" enabled for secure connections, the Plex clients have no issues connecting to the server, even if the server is using a self-signed certificate (browsers, of course, complain, but even plex.tv integration still works without complaint).  However, rclone itself does not integrate but gives the very cryptic error:

`ERROR : plex: websocket.Dial ws://1.2.3.4:32400/:/websockets/notifications?X-Plex-Token=---redacted---: unexpected EOF`

Since even those using valid certificates are most likely connecting from rclone to Plex via the LAN instead of the WAN, the certificate may still appear invalid for 127.0.0.1 or internal IP addresses (which, as far as I know, aren't supported by Let's Encrypt for obvious reasons).  Of course, one could workaround this issue with creating DNS or host-file entries but, ideally, there would be a flag/option to tell rclone to ignore certificate errors.

#### What is your rclone version (eg output from `rclone -V`)

Happens on all versions with Plex-integration support up to and including current beta.  I'm using v1.40-043-g92c5aa37β right now.

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Ubuntu 17.10 64-bit

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

Plex integration configuration in rclone.conf

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

`2018/04/01 13:36:02 INFO  : GoogleCache: Connected to Plex server: https://10.10.10.11:32400
2018/04/01 13:36:02 ERROR : plex: websocket.Dial ws://10.10.10.11:32400/:/websockets/notifications?X-Plex-Token=---redacted---: unexpected EOF`

(note that 10.10.10.11 is the Plex server which, in my case, is NOT the same server running rclone mount)

remusb:
Wasn't this fixed with: https://github.com/ncw/rclone/commit/71b4f1ccabf4423af6cd42f8498319e349b6400f

BinsonBuzz:
[...]

PR #2554, opened 2018-09-08
docs: add section for .plex.direct urls to cache

B4dM4n:
A discussed in #2215, this adds a section to the cache docs to explain the usage of `.plex.direct` URL's.

### DIFF
--- a/docs/content/cache.md
+++ b/docs/content/cache.md
@@ -182,6 +182,28 @@ and password) in your remote and it will be automatically enabled.
 Affected settings:
 - `cache-workers`: _Configured value_ during confirmed playback or _1_ all the other times
 
+##### Certificate Validation #####
+
+When the Plex server is configured to only accept secure connections, it is
+possible to use `.plex.direct` URL's to ensure certificate validation succeeds.
+These URL's are used by Plex internally to connect to the Plex server securely.
+
+The format for this URL's is the following:
+
+https://ip-with-dots-replaced.server-hash.plex.direct:32400/
+
+The `ip-with-dots-replaced` part can be any IPv4 address, where the dots
+have been replaced with dashes, e.g. `127.0.0.1` becomes `127-0-0-1`.
+
+To get the `server-hash` part, the easiest way is to visit
+
+https://plex.tv/api/resources?includeHttps=1&X-Plex-Token=your-plex-token
+
+This page will list all the available Plex servers for your account
+with at least one `.plex.direct` link for each. Copy one URL and replace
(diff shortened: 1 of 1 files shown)


docs/content/cache.md
```diff
@@ -182,6 +182,28 @@ and password) in your remote and it will be automatically enabled.
 Affected settings:
 - `cache-workers`: _Configured value_ during confirmed playback or _1_ all the other times
 
+##### Certificate Validation #####
+
+When the Plex server is configured to only accept secure connections, it is
+possible to use `.plex.direct` URL's to ensure certificate validation succeeds.
```
evulhotdog: Shouldn't this say required to use the *.plex.direct URL's for the certificate to evaluate at all without issue?

docs/content/cache.md
```diff
@@ -182,6 +182,28 @@ and password) in your remote and it will be automatically enabled.
...
+The format for this URL's is the following:
+
+https://ip-with-dots-replaced.server-hash.plex.direct:32400/
+
+The `ip-with-dots-replaced` part can be any IPv4 address, where the dots
+have been replaced with dashes, e.g. `127.0.0.1` becomes `127-0-0-1`.
+
+To get the `server-hash` part, the easiest way is to visit
+
+https://plex.tv/api/resources?includeHttps=1&X-Plex-Token=your-plex-token
```
evulhotdog: Can you explain where to grab the Plex Token from as well?

docs/content/cache.md
B4dM4n: Not necessarily, the Plex server can also be configured to use a user provided certificate in addition to the `.plex.direct` one. Another possibility is to put a reverse proxy in front of the Plex server with a custom certificate.

There is also the option to use #2555 once it's merged (which also needs to be documented). Maybe it's better to wait until it is done and then update this PR.

docs/content/cache.md
evulhotdog: While you and I understand, as you mentioned, it won't hurt to make the documentation more complete by referring where to get each value, etc.

B4dM4n: Sure, but where to stop adding things like this to the documentation.

Anybody who wants to use Plex more than "install on my NAS, add files and play" should know how to find and use the Plex-Token, since it is the key to all further usage.

Finding these `.plex.direct` URL's is a bit harder and they are barely documented otherwise, so adding this information adds a real value for others.

ncw:
This looks like an improvement - thanks @B4dM4n - will merge now :-)

PR #2555, opened 2018-09-08
cache: add plex_insecure option to skip certificate validation

### LINKED ISSUE #2215: Plex integration with HTTPS requires a valid certificate if securitymode=required

B4dM4n:
As requested in #2215, this adds a `plex_insecure` config setting to the cache backend, which will turn off all certificate verification's when connecting to the Plex server.

Fixes #2215

### DIFF
--- a/backend/cache/cache.go
+++ b/backend/cache/cache.go
@@ -82,6 +82,10 @@ func init() {
 			Hide:     fs.OptionHideBoth,
 			Advanced: true,
 		}, {
+			Name:     "plex_insecure",
+			Help:     "Skip all certificate verifications when connecting to the Plex server",
+			Advanced: true,
+		}, {
 			Name:    "chunk_size",
 			Help:    "The size of a chunk. Lower value good for slow connections but can affect seamless reading.",
 			Default: DefCacheChunkSize,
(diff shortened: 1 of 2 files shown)


ncw:
That looks very useful - thank you :-)

PR #2556, opened 2018-09-08
webdav: add workaround for missing mtime - fixes #2420

### LINKED ISSUE #2420: Listing with webdav not working with NextCloud external storages
When listing a webdav (NextCloud) remote I get the following error

```
rclone ls myremote:
2018/07/27 23:44:13 ERROR : : error listing: couldn't list files: parsing time "" as "Mon Jan _2 15:04:05 MST 2006": cannot parse "" as "Mon"
2018/07/27 23:44:13 Failed to ls: couldn't list files: parsing time "" as "Mon Jan _2 15:04:05 MST 2006": cannot parse "" as "Mon"
```

I tried the current release v1.42 and also a build from the current revision 9c90b5e both on Windows 10 with amd64 architecture.
I found out the problem is that I have an external storage configured in NextCloud and the corresponding file entry has no modification time:

```
  <d:response>
    <d:href>/remote.php/webdav/ftp/</d:href>
    <d:propstat>
      <d:prop>
        <d:resourcetype>
          <d:collection/>
        </d:resourcetype>
        <d:quota-used-bytes>5</d:quota-used-bytes>
        <d:quota-available-bytes>-3</d:quota-available-bytes>
        <d:getetag>"5581aa796f9be"</d:getetag>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
    <d:propstat>
      <d:prop>
        <d:getlastmodified/>
        <d:getcontentlength/>
        <d:getcontenttype/>
      </d:prop>
      <d:status>HTTP/1.1 404 Not Found</d:status>
    </d:propstat>
  </d:response>
```

This bug is somewhat related to #1952, however the problem here is not the wrong time format but the missing modification time.
I can confirm that it works with the following dirty hack in UnmarshalXML in webdav/api/types.go

```
if (v=="") {
	*t = Time(time.Now())
	return nil
}
```

However, I think the proper way of solving this might be evaluating the d:status tag and see if it is OK before parsing the properties.

buergi:
Workaround for #2420, maybe there is a better fix, but this will do for the moment.
I replaced the proposed Now() with unix time 0, as I think this is less arbitrary and prevents possible files without an mtime from constant overwriting during sync.

### DIFF
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -151,6 +151,12 @@ func (t *Time) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
 		return err
 	}
 
+	// If time is missing then return the epoch
+	if v == "" {
+		*t = Time(time.Unix(0, 0))
+		return nil
+	}
+
 	// Parse the time format in multiple possible ways
 	var newT time.Time
 	for _, timeFormat := range timeFormats {


ncw:
Code looks good!  Can you run through `go fmt` to fix the formatting - thanks :-)

Nice idea about the 1970 times BTW

ncw:
I merged this already in 753cc63d96ceab60f49547b382c36f96ac42a141 - not sure why the PR is still open!  Closing now.

PR #2558, opened 2018-09-08
jottacloud: fix for --fast-list

albertony:
Fixes TestIntegration/TestFsListRSubdir, discovered after merge of #2545.


ncw:
That is looking great now - thanks :-)

I'll merge it.

Issue #2575, opened 2018-09-13
Rclone RC - Add support for CORS to http interface


frenos:
Hey,
just a little feature request for the rc http interface.

I'm currently building a small angular app to display the status of rclone and in the future also allow changing settings. It's supposed to fully run in your browser and communicate directly with rclone, without any backend etc.

Currently we can't access the http interface directly from a browser because every browser sends an "OPTIONS" request first to check for CORS.

Can we get a either a flag to add our specific sites or a flag to turn CORS off (e.g. set Access-Control-Allow-Origin: *)?

#### What is your rclone version (eg output from `rclone -V`)
rclone v1.43.1
- os/arch: linux/amd64
- go version: go1.11

ncw:
I don't know a great deal about CORS but in principle I'm fine with this.  Perhaps a flag to control the CORS setting for rc would be the best option? Is turning it off unconditionally safe?

Another thing that comes to mind is that rclone rc could serve your html/js/css files which would mean that CORS wouldn't be necessary?

frenos:
If we could give a folder for rclone rc to serve, that would also work fine. Depends on what is easier to implement.
I don't think allowing everything will worsen security as currently rclone takes post request from anywhere, it's just the browser that blocks them. So if I turn off all the security features in the browser it works fine. Currently I proxy all requests through a small python server on my dev machine.

So to implement CORS with the same security implications as now (e.g. no security and allow everyone to make requests) we need two things:

1. Also allow OPTIONS requests, the response body doesn't matter.
2. Send the header "Access-Control-Allow-Origin: *" with every response.

I started looking at the source but have never programmed in go. If I can get something done I'll open a pull request. But I am of course grateful for everyone taking time to implement this or even give hints where to look in the code.

ncw:
Here is a rough outline of what would need to be done.

# For step 1 - OPTIONS

The http handler is here

https://github.com/ncw/rclone/blob/b3d8bc61aceabae9e1dc927a60c21f8444add1ba/fs/rc/rc.go#L80

What is in that function needs factoring into a different function `handlePOST` say, and that function needs to dispatch on `r.Method` for `POST` and `OPTIONS` - you'd write the `handleOPTIONS` function.

This would open the door to serving GET requests for serving the index etc...

# For step 2 - add CORS

What you want to do is in the main handler function add something like:

```go
                w.Header().Add("Access-Control-Allow-Origin", "*")
```

ncw:
I'll close this since it is done!

PR #2579, opened 2018-09-14
WIP: rc: basic CORS support #2575 

frenos:
Implemented basic CORS support as discussed in the #2575.
Tested with Angular-httpclient in Firefox and Chrome.
There is one TODO left I can't solve with my current 5minute GO skills, the writeError function.

I'll squash and rebase if this is solved.
Thank you for your time, any help is appreciated.

### DIFF
--- a/fs/rc/rc.go
+++ b/fs/rc/rc.go
@@ -94,19 +94,7 @@ func (s *server) handler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	if r.Method != "POST" {
-		writeError(errors.Errorf("method %q not allowed - POST required", r.Method), http.StatusMethodNotAllowed)
-		return
-	}
-
-	// Find the call
-	call := registry.get(path)
-	if call == nil {
-		writeError(errors.Errorf("couldn't find method %q", path), http.StatusMethodNotAllowed)
-		return
-	}
-
-	// Parse the POST and URL parameters into r.Form
+	// Parse the POST and URL parameters into r.Form, for others r.Form will be empty value
 	err := r.ParseForm()
 	if err != nil {
 		writeError(errors.Wrap(err, "failed to parse form/URL parameters"), http.StatusBadRequest)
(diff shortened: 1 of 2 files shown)


fs/rc/rc.go
```diff
@@ -130,6 +117,81 @@ func (s *server) handler(w http.ResponseWriter, r *http.Request) {
...
+
+	fs.Debugf(nil, "rc: %q: reply %+v: %v", path, out, err)
+	err = WriteJSON(w, out)
+	if err != nil {
+		// can't return the error at this point
+		fs.Errorf(nil, "rc: failed to write JSON output: %v", err)
+	}
+}
+func (s *server) handleOptions(w http.ResponseWriter, r *http.Request, path string, in Params) {
+	//TODO: find elegant way to extract this from handlers
```
ncw: I think this is probably overkill for `handleOptions` - I don't think it needs to return a JSON formatted error.

fs/rc/rc.go
```diff
@@ -130,6 +117,81 @@ func (s *server) handler(w http.ResponseWriter, r *http.Request) {
...
+			"input": in,
+		})
+		if err != nil {
+			// can't return the error at this point
+			fs.Errorf(nil, "rc: failed to write JSON output: %v", err)
+		}
+	}
+
+	// For options request we do nothing, just noop and reflect client input
+	call := registry.get("rc/noop")
```
ncw: My reading of what OPTIONS is supposed to do indicates it should return an empty body.  So I think just returning rather than running `rc/noop` will do, which also means no possibility of returning an error so no error handling needed!

fs/rc/rc.go
frenos: You are right, if we also want to support GET in a later version we would just send a 404 and not format that as json in the body, correct?

frenos:
I have changed the code based on your feedback. I would like to add the second feature we mentioned in the issue (rclone can serve html via GET itself) later if I have more time to tinker with it.

fs/rc/rc.go
ncw: Yes that is correct.

ncw:

PR #2581, opened 2018-09-15
Add --drive-base-url option to override base url for Google Drive backend

kelvl:
I have a multihoming setup at home with multiple outgoing internet connections. 

It was useful for me to be able to override the base url and point it to a haproxy load balancer.

I'm not sure if this is a useful feature for everyone but it's relatively out of the way for people that don't need it -- happy to keep this on my own fork as well!

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -68,26 +68,26 @@ var (
 		RedirectURL:  oauthutil.TitleBarRedirectURL,
 	}
 	mimeTypeToExtension = map[string]string{
-		"application/epub+zip":                            "epub",
-		"application/msword":                              "doc",
-		"application/pdf":                                 "pdf",
-		"application/rtf":                                 "rtf",
-		"application/vnd.ms-excel":                        "xls",
-		"application/vnd.oasis.opendocument.presentation": "odp",
-		"application/vnd.oasis.opendocument.spreadsheet":  "ods",
-		"application/vnd.oasis.opendocument.text":         "odt",
+		"application/epub+zip":                                                      "epub",
+		"application/msword":                                                        "doc",
+		"application/pdf":                                                           "pdf",
+		"application/rtf":                                                           "rtf",
+		"application/vnd.ms-excel":                                                  "xls",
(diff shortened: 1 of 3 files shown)


ncw:
I wonder if [go's proxy support](https://golang.org/pkg/net/http/#ProxyFromEnvironment) would work for you instead of this?

That could send all http traffic via your load balancer and works with rclone as-is.

kelvl:
I tried doing that at first but because Google's APIs uses SSL, the proxy servers need to do SSL tunneling. This together with go trying to reuse HTTP connections between requests, HAProxy wasn't able to inspect the traffic inside the tunnel to effectively load balance across the multiple internet facing interfaces.

The current set up that I have terminates SSL on HAProxy itself which allows it to inspect the individual HTTP requests that rclone makes and send them down different backends even though rclone is doing KeepAlives between itself and HAProxy. 

This worked much better than the proxy solution at least for my specific usecase! 
Please feel free to close if you think this is much too specific for general users!

ncw:
I was just wondering if we could generalise this solution somehow.

Thinking about it some more, quite a few remotes have the `endpoint` flag

```
      --azureblob-endpoint string                Endpoint for the service
      --b2-endpoint string                       Endpoint for the service.
      --qingstor-endpoint string                 Enter a endpoint URL to connection QingStor API.
      --s3-endpoint string                       Endpoint for S3 API.
      --swift-endpoint-type string               Endpoint type to choose from the service catalogue (OS_ENDPOINT_TYPE) (default "public")
```

So could you re-work this slightly so that the config parameter is called `endpoint`.  I think it should probably take the full endpoint URL so the default if blank would be `https://www.googleapis.com/drive/v3` that would fit in with the other endpoints better I think - what do you think - would that work for you?

Note that you'll need to gofmt stuff with go1.11 as the formatting changed slightly from go1.10 (see c19d1ae9a55fbe080cd0c646c3ddc11a6f44e97a)

ncw:
@kelvl just wondering if you were planning to finish this off?

ncw:
I'm going to close this since I haven't heard from you.  If you'd like to re-submit then please open a new PR - thank you :-)

PR #2583, opened 2018-09-16
change too long names cutting mechanism - fixes #2490

### LINKED ISSUE #2490: [enhancement] show beginning of the file's name instead of dots
Using iterm2 on a Mac with latest (1.42) to transfert files to Hubic.
When the file's name is too long, the last part of the file's name is shown but not beginning, it's replaced by dots. It can be very difficult to know which file is currently being uploaded.

1 rclone 2018-08-22 20-23-30

This ticket is to change the UI to <beginning><...><end> like
ThisisATestFile...EndOfFilesName.xxx

asiamarek:
Fix for the https://github.com/ncw/rclone/issues/2490 issue
I wasn't sure how I should split the visible characters ( quoting the issue description `This ticket is to change the UI to <...> like ThisisATestFile...EndOfFilesName.xxx`) so I implemented the version where a visible prefix is equal or  one character longer than a visible suffix.

### DIFF
--- a/fs/accounting/accounting.go
+++ b/fs/accounting/accounting.go
@@ -256,11 +256,14 @@ func (acc *Account) String() string {
 			etas = "0s"
 		}
 	}
+
 	name := []rune(acc.name)
 	if fs.Config.StatsFileNameLength > 0 {
 		if len(name) > fs.Config.StatsFileNameLength {
-			where := len(name) - fs.Config.StatsFileNameLength
-			name = append([]rune{'.', '.', '.'}, name[where:]...)
+			suffixLength := fs.Config.StatsFileNameLength / 2
+			prefixLength := fs.Config.StatsFileNameLength - suffixLength
+			suffixStart := len(name) - suffixLength
+			name = append(append(name[:prefixLength], '…'), name[suffixStart:]...)
 		}
 	}
 


fs/accounting/accounting.go
```diff
@@ -256,11 +256,14 @@ func (acc *Account) String() string {
...
+
 	name := []rune(acc.name)
 	if fs.Config.StatsFileNameLength > 0 {
 		if len(name) > fs.Config.StatsFileNameLength {
-			where := len(name) - fs.Config.StatsFileNameLength
-			name = append([]rune{'.', '.', '.'}, name[where:]...)
+			prefixLength := (fs.Config.StatsFileNameLength + 1) / 2
+			suffixLength := fs.Config.StatsFileNameLength / 2
+			suffixStart := len(name) - suffixLength
+			name = append(append(name[:prefixLength], rune('…')), name[suffixStart:]...)
```
ncw: You don't need to say `rune('…')` -  `'…'` is already a `rune`.

fs/accounting/accounting.go
```diff
@@ -256,11 +256,14 @@ func (acc *Account) String() string {
 			etas = "0s"
 		}
 	}
+
 	name := []rune(acc.name)
 	if fs.Config.StatsFileNameLength > 0 {
 		if len(name) > fs.Config.StatsFileNameLength {
-			where := len(name) - fs.Config.StatsFileNameLength
-			name = append([]rune{'.', '.', '.'}, name[where:]...)
+			prefixLength := (fs.Config.StatsFileNameLength + 1) / 2
```
ncw: I think you'd be better off writing

```
suffixLength := fs.Config.StatsFileNameLength / 2
prefixLength := fs.Config.StatsFileNameLength - suffixLength 
```

That is clearer to me that `suffixLength + prefixLength ==  fs.Config.StatsFileNameLength` which is the invariant we are looking for I think.

I think what you wrote does the same thing, I just have more trouble convincing myself of it ;-)

asiamarek: As far as i know it does :) 
(if the length is odd lets say 7, then 7/2 = 3 and 8/2 = 4, so it adds up to 7, and when it is even like 6 is, then 6/2 = 3 and 7/2 =3 so it adds up to 6)
But I agree it would look much simpler your way :)

ncw:
Perfect - thank you very much :-)

PR #2584, opened 2018-09-16
jottacloud: Fix handling of reserved characters. fixes #2531

### LINKED ISSUE #2531: Jottacloud backend and special characters
I have a folder with name "Notepad++" in my Jottacloud account, but trying to "ls" it with rclone fails:
```
ERROR : Notepad++: error listing: directory not found
Failed to ls: directory not found
```
I guess it is because the backend implementation replaces special characters (including "+") with their FULLWIDTH Unicode equivalent. Shouldn't such characters be URL (percent) encoded instead? I know for a fact that in the Jottacloud API you can refer to paths with "+" in them using "%2B", "#" with "%23" etc..

buengese:
#2531 has been open long enough. Quick and simple fix for this.


ncw:
That look fine to me - thank you.  I gave it a quick test locally :-)

PR #2585, opened 2018-09-17
cache: Remove entries that no longer exist in the source

dcpu:
If the file is deleted from another rclone instance, it will not be removed from the cache. 
The entry keep remains and causes a refresh error.
In this case expire will not work. Because it triggers refresh not remove it.

This change will remove entries that no longer exist in the source.


ncw:
@remusb have you got time to look at this?

Issue #2573, opened 2018-09-12
Google Drive v3 API speed limit


Dulanic:
Issue: Rclone is limited to 25MB/sec using Gdrive using google api v3.

This began back /w this commit: https://github.com/ncw/rclone/commit/07f20dd1fd6f89b9ebdefdc8964a8e9ee9aa9892

I am posting this here as it has been discussed on the forums and B4dM4n created a beta version that implemented a workaround (That has since disappeared) that forced rclone to use v2 API which didn't have this speed cap.

Most talk of this topic seems to have just died down with no implementation of the --drive-v2-download-min-size flag created previously by B4dM4n . This flag should be implemented via a PR so it can be implemented as a wok around to the v3 API speed limit.

Link to forum discussion:
https://forum.rclone.org/t/slow-download-from-gdrive-to-local/5733/79

![image](https://user-images.githubusercontent.com/2948206/45442737-79cbb180-b688-11e8-8b53-3a8369ddce43.png)

ncw:
@B4dM4n - what do you think, want to include the v2 flag?

animosity22:
@ncw  - Does the copy command use a different API than the mount command? I can max out a copy from a mount but not from a rclone copy command which is strange:

Same file as one is a copy from a mount and one is a rclone copy command:

![image](https://user-images.githubusercontent.com/5088712/45590961-012c5580-b913-11e8-9b6d-328bc133b3f8.png)

NurdTurd:
@animosity22 I want a response on this too. I've had this happened to me as well before.

Dulanic:
@animosity22 So do you use cache/vfs? Im wondering if your situation might not speed up in the 2nd command because it already pulled down the data from the copy. I dont think this is so much a copy vs mount thing, I think if you did the mount copy 1st /w sync youd see the same 25MB cap like I do.

animosity22:
@Dulanic  - I'm not using a cache mount so nothing is pulled cached or stored prior. You can see my connection is pulling well above 25MB/s.

![image](https://user-images.githubusercontent.com/5088712/45630810-7e68df00-ba67-11e8-8960-86c880a6f115.png)

Dulanic:
@animosity22 This damn thing gets more and more confusing, but it is has to be related to the API. I know when @B4dM4n created a "beta" test version /w the v2 API it fixed everything for me. Wish I had kept that version, and not it's deleted.

animosity22:
[...]

PR #2590, opened 2018-09-17
drive: add workaround for slow downloads

B4dM4n:
Add `--drive-v2-download-min-size` flag to allow downloading files via the drive v2 API. If files are greater than this flag, a download link is generated when needed.
The flag is disabled by default.

Workaround for #2573


B4dM4n:
Sure, I will split the commit.

Rebuilding the vendor tree after a rebase is probably the fastest way to get a clean result.

ncw:
Thank you :-)

ncw:
Do you want to advertise that in the thread on the forum?

PR #2592, opened 2018-09-18
azureblob and fstests - Exclude archive tier from SetTier integration tests

sandeepkru:
azureblob and fstests - Exclude archive tier from SetTier integration tests, added ExcludedSettings to fstests.Opt, a key-value map to hold various settings, where key is setting name and value is list of settings

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -191,19 +191,6 @@ func validateAccessTier(tier string) bool {
 	}
 }
 
-// validAccessTiers returns list of supported storage tiers on azureblob fs
-func validAccessTiers() []string {
-	validTiers := [...]azblob.AccessTierType{azblob.AccessTierHot, azblob.AccessTierCool,
-		azblob.AccessTierArchive}
-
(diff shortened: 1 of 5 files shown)


sandeepkru:
@ncw - I settled on below approach, please let me know if you have any concerns, I can change this is to be an option of simple slice of strings, but I thought this would be more generic and can be used in different purposes.
```console
	// Describes map of key value pair settings, key is setting name and value is
	// list of settings as strings which can be excluded from tests
	ExcludedSettings map[string][]string
```

ncw:
I think I prefer the less generic approach of

```golang
    TiersToTest []string
```

That has the advantage that it is fully typed and we can use the compiler/go-oracle to find uses.

The test should break if the interfaces are found but `TiersToTest` is empty.

sandeepkru:
Sounds good, with this `ListTiers` not used anywhere hence I removed that code as well. It can be introduced with `S3 and GCS` changes at `Fs` level correctly.

ncw:

I'll merge it now.

PR #2595, opened 2018-09-18
fs/operations: optimize choosing the main folder in dedup - fixes #2568

### LINKED ISSUE #2568: rclone dedupe, choose the main folder efficiently
Let's say 'Folder' and 'Folder (1)' exist in a certain directory, the webUI etc will show them as the same name, just 'Folder', but File Stream etc can't have the same filename so they show as '(1)' etc. Let's say I have 1000 files in 'Folder' and later on 'Folder (1)' is uploaded that has 5 files. I want to run dedupe so that I will end up with just 'Folder' that has 1005 files, or less if some of those 5 files already existed in the original folder. You'd think the way rclone would do this is to treat the 'Folder' as the original one, and copy the contents of 'Folder (1)' into 'Folder' and delete 'Folder (1)'. Instead I see rclone is copying the contents of 'Folder' into 'Folder (1)' and then deleting 'Folder'. This means there's 1000 move (or 'merge' as the CLI shows) operations instead of 5 (takes much longer, more API calls, etc.). It also means the activity pane in the webUI will now show only 'moved x files' and you'll lose the activity of the original 'Folder' because it was deleted.

rclone v1.43.1

Windows 10 64bit

Google Drive

rclone dedupe GoogleDrive: --dedupe-mode largest -vv

asiamarek:
Fix to  #2568 issue.

I used a different function ('Size()') than @ncw suggested in the issue comments, but if I understand the code correctly, it should give us exactly the information we need.

### DIFF
--- a/fs/operations/dedupe.go
+++ b/fs/operations/dedupe.go
@@ -209,6 +209,20 @@ func dedupeMergeDuplicateDirs(f fs.Fs, duplicateDirs [][]fs.Directory) error {
 	}
 	for _, dirs := range duplicateDirs {
 		if !fs.Config.DryRun {
+
+			//Find the largest directory and merge duplicates into it
+			largestDir := dirs[0]
+			largestDirId := 0
+			counter := 0
+			for _, dir := range dirs {
+				if dir.Size() > largestDir.Size() {
+					largestDir = dir
+					largestDirId = counter
+				}
+				counter++
+			}
+			dirs[0], dirs[largestDirId] = dirs[largestDirId], dirs[0]
+
 			fs.Infof(dirs[0], "Merging contents of duplicate directories")
 			err := mergeDirs(dirs)
 			if err != nil {


fs/operations/dedupe.go
```diff
@@ -209,6 +209,20 @@ func dedupeMergeDuplicateDirs(f fs.Fs, duplicateDirs [][]fs.Directory) error {
 	}
 	for _, dirs := range duplicateDirs {
 		if !fs.Config.DryRun {
+
+			//Find the largest directory and merge duplicates into it
+			largestDir := dirs[0]
+			largestDirId := 0
+			counter := 0
+			for _, dir := range dirs {
+				if dir.Size() > largestDir.Size() {
```
ncw: Unfortunately there aren't many backends which support the `Size()` attribute and `drive` isn't one of them :-(

I think you'll find they are all -1

So I think doing a f.List(dir.Remote()) will be necessary to find the size.

fs/operations/dedupe.go
```diff
@@ -209,6 +209,20 @@ func dedupeMergeDuplicateDirs(f fs.Fs, duplicateDirs [][]fs.Directory) error {
 	}
 	for _, dirs := range duplicateDirs {
 		if !fs.Config.DryRun {
+
+			//Find the largest directory and merge duplicates into it
```
ncw: I think I'd factor this out into a function `findLargestDirectory(dirs []fs.Directory) (largestDirId int, err error)` then it will be easy to write a unit/integration test for it.

ncw:
Are you going to have a go at fixing this up @asiamarek ?  Or would you like me to help?

ncw:
@asiamarek wrote:
> Hi, if thats not a problem I can continue working on this once I'm back from my vacation (around week or so).

That would be fine - whatever works for you :-)

ajh0912:
Just wondering if there's any more progress on this, thanks

ncw:
@asiamarek How are you doing with this - need any help?

ajh0912:
Hi, any progress? (I appreciate that it's all voluntary)

ncw:
I'm going to close this since I haven't heard from you in a while @asiamarek .  If you (or anyone else) would like to pick it up in a new pull request, I'd be delighted!

Issue #878, opened 2016-11-14
Improve verbose output when using --delete-excluded


diamondsw:
> What is your rclone version (eg output from `rclone -V`)
1.34
> Which OS you are using and how many bits (eg Windows 7, 64 bit)
macOS 10.12.1, 64-bit
> Which cloud storage system are you using? (eg Google Drive)
Amazon Drive
> The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
/usr/local/bin/rclone copy --verbose --stats 0 --dry-run --exclude-from /Users/jochs/Desktop/rclone_filter.txt --delete-excluded "/Volumes/Shares/Archive" "amazon:/Backup/Archive"
> A log from the command with the `-v` flag (eg output from `rclone -v copy /tmp remote:tmp`)
.DS_Store: Excluded from sync (and deletion)                                                                                                                    
.VolumeIcon.icns: Excluded from sync (and deletion)                                                                                                             
Media/.DS_Store: Excluded from sync (and deletion)                                                                                                              
Computing/.DS_Store: Excluded from sync (and deletion)                                                                                                          
Documents/.DS_Store: Excluded from sync (and deletion)                                                                                                          

In short, rclone still prints "(and deletion)" even with the '--delete-excluded' option. This makes it sound as if excluded files won't be deleted, although I expect they actually would be. Holding off on using this option until it's clarified. ;)

vesper8:
+1 I agree this should be changed, I use --delete-excluded and got confused by this message in the log

ncw:
@ssaqua fancy fixing up the log messages?

ssaqua:
Sure, will start a PR soon.

tgallant:
looks like this can be closed

PR #2597, opened 2018-09-20
list: change debug logs for excluded items

ssaqua:
Addresses discussion started in #878.

I thought it might be better to drop "from sync (and deletion)" altogether since listing is not just restricted to the sync command (`rclone -vv --exclude a tree remote:dir` would output the "Excluded from sync (and deletion)" line for example).

With files setup as:

```
dir/a
dir/b

remote/dir/a
remote/dir/b
```

The output from a `rclone -vv --exclude a --delete-excluded sync dir remote:dir` seems clear enough:

```
2018/09/20 23:17:33 DEBUG : a: Excluded
2018/09/20 23:17:33 INFO : Google drive root 'dir': Waiting for checks to finish
2018/09/20 23:17:33 DEBUG : b: Size and modification time the same (differ by -852.223µs, within tolerance 1ms)
2018/09/20 23:17:33 DEBUG : b: Unchanged skipping
2018/09/20 23:17:33 INFO : Google drive root 'dir': Waiting for transfers to finish
2018/09/20 23:17:33 INFO : Waiting for deletions to finish
2018/09/20 23:17:34 INFO : a: Deleted
```

### DIFF
--- a/fs/list/list.go
+++ b/fs/list/list.go
@@ -28,7 +28,7 @@ func DirSorted(f fs.Fs, includeAll bool, dir string) (entries fs.DirEntries, err
 	// starting directory, otherwise ListDirSorted should not be
 	// called.
 	if !includeAll && filter.Active.ListContainsExcludeFile(entries) {
-		fs.Debugf(dir, "Excluded from sync (and deletion)")
+		fs.Debugf(dir, "Excluded")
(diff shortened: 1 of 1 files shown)


ncw:
Thanks for doing that :-)  The logs make much more sense now.

Issue #2006, opened 2018-01-23
DirChangeNotify misses changes


B4dM4n:
As I read through #1937 I became aware of the following statement about `DirChangeNotify` in the drive remote from @remusb:
> moving a file doesn't seem to expire both parent folders, only the destination. At the moment this makes things a bit tricky as cache would best work if it doesn't expire its entries at all but it seems some cases aren't included in the notifications

I tried moving some files between folders using a master build of rclone to replicate the behavior.
My setup was the following:
```sh
$ rclone mount -vv --dir-cache-time 1h mydrive: /mnt/a
$ rclone mount -vv --dir-cache-time 1h mydrive: /mnt/b
$ find /mnt/a
/mnt/a/folder1
/mnt/a/folder1/file1
/mnt/a/folder1/file2
/mnt/a/folder2
$ find /mnt/b
/mnt/b/folder1
/mnt/b/folder1/file1
/mnt/b/folder1/file2
/mnt/b/folder2
$ mv /mnt/a/folder1/file1 /mnt/a/folder2
# wait for "Checking for changes on remote" in /mnt/b
$ find /mnt/b
/mnt/b/folder1
/mnt/b/folder1/file1
/mnt/b/folder1/file2
/mnt/b/folder2
/mnt/b/folder2/file1
```

> What is your rclone version (eg output from `rclone -V`)

rclone v1.39-072-g33b19c13
- os/arch: linux/amd64
- go version: go1.9.2

> Which OS you are using and how many bits (eg Windows 7, 64 bit)

Linux (NixOS)
> Which cloud storage system are you using? (eg Google Drive)

Google Drive

While digging through `backend/drive/drive.go` i found the source of this bug. In [drive.go#1153](https://github.com/ncw/rclone/blob/master/backend/drive/drive.go#L1153) `DirChangeNotify` tries to use `f.dirCache.GetInv` to get the old path for the changed file. This can never work, because `dirCache` is, as the name implies, only a directory cache, not a file cache.

~~The amazondrive remote is using the same method in  [amazonclouddrive.go#1264](https://github.com/ncw/rclone/blob/master/backend/amazonclouddrive/amazonclouddrive.go#L1264). It should be affected as well.~~ amazondrive is not affected

What would be the best fix for this bug? Should `dircache.DirCache` be changed into a full file tree cache?

ncw:
Excellent work to find the cause of that bug!

[...]

PR #2606, opened 2018-09-25
drive: improve directory notifications in ChangeNotify

B4dM4n:

When moving a directory in drive, most of the time only a notification for the directory itself is created, not the old or new parents.

This tires to find the old path for the changed directoy in the `dirCache` and the new path with the `dirCache` of the new parent, which can result in two notifications for a moved directory.

It is somewhat related to #2006, but affects directories instead of files.

#### Checklist


ncw:
Thank you :-)

PR #2616, opened 2018-09-28
onedrive: ignore OneNote files by default - fixes #211

### LINKED ISSUE #211: onedrive: Ignore OneNote links
In OneDrive, OneNote links live under Documents folder and they seem to be a URL links to the actual onenote.com link. Currently rclone gives an error when encountered such files:

2015/11/10 11:18:27 Documents/Quick Notes: Failed to open: itemNotFound: The specified item does not exist, or does not have content.
2015/11/10 11:18:27 Documents/Computing: Failed to open: itemNotFound: The specified item does not exist, or does not have content.
2015/11/10 11:18:27 Documents/Programming: Failed to open: itemNotFound: The specified item does not exist, or does not have content.

It would be nice to ignore them.

Cnly:
OneNote files are a kind of files that look like files, but behave like folders. Operations like `Open` and `Update` fail on them. This PR hides them from directory listing, and prevent invalid operations on them, returning more user-friendly error messages. 

Since hiding these files from listing may also prevent them from being deleted, the flag `--onedrive-expose-onenote-files` is also added, which allows OneNote files to show up in directory listing. Illegal operations on them are still rejected even if this flag is set.

This should fix #211.

### DIFF
--- a/backend/onedrive/api/types.go
+++ b/backend/onedrive/api/types.go
@@ -9,6 +9,9 @@ import (
 
 const (
 	timeFormat = `"` + time.RFC3339 + `"`
+
+	// PackageTypeOneNote is the package type value for OneNote files
+	PackageTypeOneNote = "oneNote"
 )
 
 // Error is returned from one drive when things go wrong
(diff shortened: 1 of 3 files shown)


ncw:
Is this ready to merge now?

Cnly:
@ncw I've been waiting for feedback but there seems to be none... Do you think we should post this onto the forum and see how things go before v1.44? I'm seeing no problems on my side, though.

Modelmat:
I can test in 12-13 hours :)

ncw:
@Cnly wrote:
> I've been waiting for feedback but there seems to be none... Do you think we should post this onto the forum and see how things go before v1.44? I'm seeing no problems on my side, though.

It is a simple enough change and we still have time in the release cycle to fix stuff so I think we should probably merge it now.

Testing would be appreciated @Modelmat :-)

Modelmat:
Okay, this works fine on my end, with the error message. Thanks!

Cnly:
@Modelmat Thanks for your help! I'll merge this now.

Issue #1788, opened 2017-10-31
Dropbox: error reading destination directory: too_many_requests


mathieucarbou:
> What is your rclone version (eg output from `rclone -V`)
> Which OS you are using and how many bits (eg Windows 7, 64 bit)

```
rclone v1.38-049-gd0b9baabβ
- os/arch: linux/amd64
- go version: go1.9.2
```

> Which cloud storage system are you using? (eg Google Drive)

`Dropbox`

> The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`./rclone --log-level "INFO" --transfers 3 --delete-excluded --config "$CONFIG" --exclude-from "EXCLUDES" sync /share/homes/mathieu/Backups dropbox:NAS/Backups`

> A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
2017/10/31 09:39:58 INFO  : 
Transferred:      0 Bytes (0 Bytes/s)
Errors:                 0
Checks:             10893
Transferred:            0
Elapsed time:        7m1s

2017/10/31 09:40:40 ERROR : mathieu@MACBOOKPRO/mathieu/Pictures/Lightroom/medias/medias Smart Previews.lrdata/8/8AB4: error reading destination directory: too_many_requests/
2017/10/31 09:40:42 ERROR : mathieu@MACBOOKPRO/mathieu/Pictures/Lightroom/medias/medias Smart Previews.lrdata/4/4A14: error reading destination directory: too_many_requests/.
2017/10/31 09:40:46 ERROR : mathieu@MACBOOKPRO/mathieu/Pictures/Lightroom/medias/medias Smart Previews.lrdata/D/D9B3: error reading destination directory: too_many_requests/...
2017/10/31 09:40:48 ERROR : mathieu@MACBOOKPRO/mathieu/Pictures/Lightroom/medias/medias Smart Previews.lrdata/C/C8B8: error reading destination directory: too_many_requests/..
2017/10/31 09:40:58 INFO  : 
Transferred:      0 Bytes (0 Bytes/s)
Errors:                 4
Checks:             10893
Transferred:            0
Elapsed time:        8m1s
```

I think this issue can also be in relation to #1354. Dropbox seems to have a rate-limited API. See:

- https://github.com/saguiitay/DropboxRestAPI/issues/12
- https://www.dropbox.com/developers/documentation/http/documentation

[...]

Issue #1798, opened 2017-11-05
Sync: identical file downloaded/uploaded when only modification time differs


bzamecnik:
I tried to synchronize a DropBox directory to local and vice versa. For some files --dry-run reported "Not updating modification time as --dry-run" but then real "sync" started transferring the file, even though the content was already identical, just to update the modification time. I can understand that possibly a cloud service API might not allow setting just mtime, but local filesystem certainly does.

> What is your rclone version (eg output from `rclone -V`)

```
rclone v1.38
- os/arch: darwin/amd64
- go version: go1.9
```

> Which OS you are using and how many bits (eg Windows 7, 64 bit)

OS X El Capitan

> Which cloud storage system are you using? (eg Google Drive)

DropBox

> The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

```
rclone sync -v dropbox:some/path/ /Volumes/external_disk/some/path/
```

> A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

ncw:
Can you paste a log with `-vv` of the problem?

bzamecnik:
OK. Probably I wrongly observed the behavior. For dropbox -> local it now modified the mtime without download. For local -> dropbox it said it can't update mtime so it reuploaded it.

Same mtime:

```
$ rclone sync -vv dropbox:tmp/rclone-1798/ .
2017/11/05 21:10:27 DEBUG : Using config file from "/Users/bzamecnik/.rclone.conf"
2017/11/05 21:10:27 DEBUG : rclone: Version "v1.38" starting with parameters ["rclone" "sync" "-vv" "dropbox:tmp/rclone-1798/" "."]
2017/11/05 21:10:28 INFO  : Local file system at /Users/bzamecnik/tmp/rclone-1798: Modify window is 1s
2017/11/05 21:10:29 DEBUG : J.S.Bach - Tocata - D Moll.mp3: Size and modification time the same (differ by 0s, within tolerance 1s)
2017/11/05 21:10:29 DEBUG : J.S.Bach - Tocata - D Moll.mp3: Unchanged skipping
2017/11/05 21:10:29 INFO  : Local file system at /Users/bzamecnik/tmp/rclone-1798: Waiting for checks to finish
2017/11/05 21:10:29 INFO  : Local file system at /Users/bzamecnik/tmp/rclone-1798: Waiting for transfers to finish
[...]

PR #2622, opened 2018-10-03
dropbox: Fix Issue #1788 Retry-After

mathieucarbou:

See #1788

No - just on #1798 and https://github.com/dropbox/dropbox-sdk-go-unofficial/issues/33

#### Checklist

- [ ] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

Didn't do anything of that because this PR is a PoC to test the Retry-After support vs `--tpsLimit`.
It also depends on Dropbox's SDK support.
This is just a working base.

#### Disclaimer

Not a Go dev, and do not know the rclone internals (such as the impact on the pacer or whether we still need it), so I strongly suggest a big review for this one, or take the code and update it.


ncw:
This code looks very useful thanks - I think we can merge that with a little bit of work :-)

We'll need to wait until the commit has been accepted upstream then use `go mod` to [update the vendor directory](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#updating-a-dependency) (in a separate commit).

So when it is upstream can you make this into two commits
- update the vendor directory
- add the fix

Can you checkout the [commit message guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages) too please.


Thank you

mathieucarbou:
Ok! Let's wait for the Dropbox team then, hoping their reaction time will be less than a year... 😁😁

ncw:
@mathieucarbou 
> btw, could you please help me figuring out the impact on the pacer code ? I am not quite sure that the place where I put the sleep is the right place considering the whole code is wrapped by the pacer code.

This is what the b2 backend does...  (I don't really like having to divide the value by 2 - that is a bit hacky - should probably fix that!)

https://github.com/ncw/rclone/blob/ddbd4fd881a5068ab115017a3b6189d0e4111bbc/backend/b2/b2.go#L220-L225

mathieucarbou:
@ncw ok I'll do that when I'll cleanup the PR.

BTW, supposing the Dropbox Team is really not responsive on this issue, do you think there is a way in Go to "patch" or override (but only for your project) the files that go downloads from GithHub for the dependencies ? A little like I did, but what I did directly impacts and modify the dropbox dependencies so this quite bad...

I think this all comes down to: what could we do to patch the dropbox files or replace them in case the Dropbox Team is not responsive and we want an rclone version with the Retry-After support ?

ncw:
> BTW, supposing the Dropbox Team is really not responsive on this issue, do you think there is a way in Go to "patch" or override (but only for your project) the files that go downloads from GithHub for the dependencies ? A little like I did, but what I did directly impacts and modify the dropbox dependencies so this quite bad...
> 
> I think this all comes down to: what could we do to patch the dropbox files or replace them in case the Dropbox Team is not responsive and we want an rclone version with the Retry-After support ?

Yes there is a way to do this... It involves forking the project and adjusting the `go.mod` file and adding a [replace directive](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive).

That said I've found the Dropbox team does merge external patches as you can see from the changelog (and my patches there!).

mathieucarbou:
Dropbox PR update: https://github.com/dropbox/dropbox-sdk-go-unofficial/pull/46
This PR updated

ncw:
> I've spent 1h figuring out how to point to my dropbox fork's branch in go.mod and I gave up. Dependency management in GO is really a hell compared to java.

The easiest way to get the HEAD of the sdk into rclone is to do this

```
 GO111MODULE=on go get -u github.com/dropbox/dropbox-sdk-go-unofficial@master
 GO111MODULE=on go mod vendor
```

Then you commit the vendor stuff and the go.mod,go.sum stuff in a separate commit.

Ideally we would persuade @diwakergupta to make a release, then we would drop the `@master` from the above and just do a normal [update a module](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#updating-a-dependency).

Can you make sure the vendor commit is separate from the actual code commit please?

mathieucarbou:
@ncw : done!

diwakergupta:
@mathieucarbou @ncw Pushed a new release, enjoy!

mathieucarbou:
@ncw : I have updated the first commit to point to tag [v5.0.0](https://github.com/dropbox/dropbox-sdk-go-unofficial/tree/v5.0.0)

ncw:
@mathieucarbou that looks perfect now - I will merge in just a moment - thank you.

@diwakergupta thank you very much for your help :-)

PR #2634, opened 2018-10-09
backend: unify NewFs path handling for wrapping remotes

B4dM4n:

This change unifies the way the wrapping remotes join the configured root path with the path provided to `NewFs` and allows to use `\\server\` paths on Windows. 7b975bc1ff7c4f4ff900db0a2b1e2e254cdbd57a fixed this for the `alias` remote, but `crypt` and `cache` could not handle them until now.

The new function `fspath.JoinRootPath` that is used to join the paths is equivalent to `path.Join`, but if the first non empty element starts with `//` this is preserved. Since all path are already `/` separated (by `fspath.Parse`) there is no need to use `filepath.Join`.

No

#### Checklist

### DIFF
--- a/backend/alias/alias.go
+++ b/backend/alias/alias.go
@@ -2,13 +2,12 @@ package alias
 
 import (
 	"errors"
-	"path"
-	"path/filepath"
 	"strings"
 
 	"github.com/ncw/rclone/fs"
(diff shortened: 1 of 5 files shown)


ncw:
Very nice - thank you :-)

PR #2648, opened 2018-10-13
Add uPnP/DLNA server

nicolov:
Add `rclone serve dlna` to serve the contents of a remote as an uPnP/DLNA media server that can be used with Smart TVs, Xbox, etc..

Most of the code is just lifted from https://github.com/anacrolix/dms (new BSD license), but I can clean it up and add tests if there's interest in merging this (not sure a media server belongs in rclone, but it's pretty convenient).


ncw:
Very nice!

I'm in the middle of making the 1.44 release at the moment - I'll look at this in detail after that is out the door.

A few notes

- You'll need to [vendor the external modules you are using](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency)
- Can you rebase it to master to pick up the gimports fixes
- Can you see if you can use the cmd/httplib library for making the http server?  This will get you lots of flags for free.

See

https://github.com/ncw/rclone/blob/e1503add41020fafe7e02fa7bb2e7c69dc6c1b4f/cmd/serve/http/http.go#L62-L71

nicolov:
Sorry for the delay. I've updated to master and fixed the vendor code. I took a look and don't think I would save much code by reusing the HTTP code. The listen port doesn't really matter, as the DLNA will advertise it to clients anyway, and the authentication is done differently too. There still are a bunch of `make check` issues on transitive dependency, what should we do about them?

ncw:
> Sorry for the delay. I've updated to master and fixed the vendor code

Great!

Can you rebase it on master rather than using a merge commit?

> I took a look and don't think I would save much code by reusing the HTTP code. The listen port doesn't really matter, as the DLNA will advertise it to clients anyway, and the authentication is done differently too.

OK

> There still are a bunch of `make check` issues on transitive dependency, what should we do about them?

I think what is left is

```
cmd/serve/dlna/dlna.go:166:17:	defer in.Close()
cmd/serve/dlna/dlna.go:176:10:	w.Write(server.rootDescXML)
```

Which is errcheck complaining about errors which haven't been checked in your code.

For the first use `fs.CheckClose`

```
func CheckClose(c io.Closer, err *error)
    CheckClose is a utility function used to check the return from Close in a
    defer statement.
```

For the second, check the error!

----

Can we do any tests on this?

ncw:
It looks like there is another error to check here

```
cmd/serve/dlna/dlna.go:140:19:	s.initServicesMap()
```

Then hopefully the build will go green :-)

Can you fill out some more instructions in that file too?  Those instructions will be the `-h` for the command and become the man page and the web page.

I'll do a full review once the build has gone green.

nicolov:
I've reformatted the commits as requested.

Oh, I'm sorry that it doesn't work on your devices. I've happily been using it on VLC on AppleTV and nPlayer on iOS since I posted the PR. Are you ok with merging like this for now, or do you want more extensive compatibility tests?

nicolov:
I've added a small tweak to make it work with VLC on Android and Linux.

ncw:
Thanks for doing the code tweaks :-)

I've had an experiment with it and I got it to work with vlc on linux which is good :-)  No sign of it on my Sony or Samsung devices and on BubbleUPNP it sees it but can't browse.

I'll merge this now - do you want to annouce it on [the rclone forum](https://forum.rclone.org/) - then we can get some users testing it?

beredim:
I tried using this (v. 1.46) today unsuccessfully on

- LG TV, LibreELEC (both fail to detect the server), 
- also BubbleUPNP on Android (as @ncw above, server is detected but failure to browse)
- LocalCast on Android also fails to detect the server.

1. I couldn't even get it to work on VLC (v. 3.0.6 on Ubuntu 18.04) which should have worked. Is View --> Playlist --> Local Network --> UPnP the correct sequence or am I missing something ?

2. Should we continue the discussion here or on the forum? I did start a forum thread, just thought of commenting here also to make sure you get pinged.

ncw:
Yes i think this needs a bit more work...  @nicolov have you got time to look at this some more?

nicolov:
Interesting, as I've been using it with no problem on both Apple TV and
nPlayer on iOS.

I'll try to see why the discovery is not working on Kodi (on which
LibreElec is based).

ncw:
I've just merged  95194ad which makes it work with my Samsung TV :-)

@beredim can you have a go with https://beta.rclone.org/v1.46.0-069-g229898dc-beta/ (uploaded in 15-30 mins)

nicolov:
@ncw How can you be so productive? I barely had the time to look at this, and you run the whole project!

beredim:
> @beredim can you have a go with https://beta.rclone.org/v1.46.0-069-g229898dc-beta/ (uploaded in 15-30 mins)

Sure. I'll check tomorrow evening and comment.

beredim:
It's getting better:

- LibreELEC/Kodi still not detecting the server
- LG TV still not detecting the server (it's a 2016 webos 3 model)
- However, BubbleUPNP, LocalCast, and VLC* on Android all now appear to work

Maybe, if/when @sretlawd 's #3045 gets merged it will solve even more issues.

* VLC 3.0.6 (snap) on Ubuntu 18.04 still doesn't work, but I think it's a VLC on Ubuntu bug (since it appears to also fail to detect other dlna servers)

Issue #2650, opened 2018-10-14
Jottacloud failing some integration tests


ncw:
The Jottacloud backend is consistently failing [these integration tests](https://pub.rclone.org/integration-tests/2018-10/2018-10-14-164449/summary-v1.43-170-g5dac8e05-beta.txt) but only with `--fast-list`

```
2018/10/14 16:44:37   * /home/rclone/go/src/github.com/ncw/rclone/fs/operations/operations.test -test.timeout 30m0s -remote TestJottacloud: -fast-list -test.run '^(TestRmdirsNoLeaveRoot|TestRmdirsLeaveRoot|TestListFormat)$'
2018/10/14 16:44:37     * Failed tests: [TestRmdirsNoLeaveRoot TestRmdirsLeaveRoot TestListFormat]
```

```
2018/10/14 16:44:37   * /home/rclone/go/src/github.com/ncw/rclone/fs/sync/sync.test -test.timeout 30m0s -remote TestJottacloud: -fast-list -test.run '^(TestCopyEmptyDirectories|TestSyncIgnoreErrors|TestSyncAfterRemovingAFileAndAddingAFileSubDir|TestSyncAfterRemovingAFileAndAddingAFileSubDirWithErrors)$'
2018/10/14 16:44:37     * Failed tests: [TestCopyEmptyDirectories TestSyncIgnoreErrors TestSyncAfterRemovingAFileAndAddingAFileSubDir TestSyncAfterRemovingAFileAndAddingAFileSubDirWithErrors]```
```

So to test these locally

```
cd fs/operations
go test -v -remote TestJottacloud: -fast-list -run '^(TestRmdirsNoLeaveRoot|TestRmdirsLeaveRoot|TestListFormat)$'
```

and

```
cd fs/sync
go test -v -remote TestJottacloud: -fast-list -run '^(TestCopyEmptyDirectories|TestSyncIgnoreErrors|TestSyncAfterRemovingAFileAndAddingAFileSubDir|TestSyncAfterRemovingAFileAndAddingAFileSubDirWithErrors)$'
```

@albertony do you fancy having a look at these since you did the `--fast-list` support?

Unfortunately I haven't had time for the 1.44 release

albertony:
Think I fixed it, thanks for the description of how to reproduce.

PR #2651, opened 2018-10-14
jottacloud: fix bug in --fast-list related to empty folders (#2650)

albertony:
Failing integration tests described in #2650 should be fixed now. Seems the problem was with the empty folders in the test case. It was caused by some incorrect logic in my previous integration test fix (#2558) - which I struggled a bit with, so hopefully all tests works now.

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -464,12 +464,12 @@ func (f *Fs) listFileDir(remoteStartPath string, startFolder *api.JottaFolder, f
 		if folder.Deleted {
 			return nil
 		}
-		folderPath := path.Join(folder.Path, folder.Name)
-		remoteDirLength := len(folderPath) - pathPrefixLength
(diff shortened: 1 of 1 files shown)


ncw:
Thanks for fixing this :-)

Alas, this seems to break the normal integration tests

```
cd backend/jottacloud/
go test -v
```

```
    --- FAIL: TestIntegration/TestFsListRSubdir (0.46s)
        require.go:738: 
            	Error Trace:	fstests.go:719
            	            				fstests.go:726
            	Error:      	"[ hello? sausage hello? sausage/êé hello? sausage/êé/Hello, 世界 hello? sausage/êé/Hello, 世界/ " ' @ < > & ? + ≠]" should have 0 item(s), but has 5
            	Test:       	TestIntegration/TestFsListRSubdir
```

ncw:
I ran all the tests - looks great thank you :-)

PR #2680, opened 2018-10-21
Added docs to use the progress flag for real time statistics

### LINKED ISSUE #2507: Stats default and/or documentation
I was a bit confused on how to get progress reports / stats from rclone. Just running a `sync` command did not give any output. Adding `-v` seems to show every file copy, which is a bit verbose. I was looking for a progress indication or regular summary, and found `--stats`. The documentation for that option says the default is 1 minute, so I expected to see stats output after a minute, but that never happened.

Looking more, I found #1180, which tells me that the default stats output happens at a more verbose log level, so you have to pass `-v`, or `--stats-log-level NOTICE`. However:
 - Typically, I'd think you want stats output, and not a full file list, so having to add `-v` seems counterproductive. In my setup, this even mixed up stats and file listing output in a way that didn't look too clear (I believe stats were written on top of the file list, with the file list showing through), but that might be something weird in my terminal setup.
 - Changing the stats log level seems a bit user-unfriendly, though it is clean from an implementation perspective.
 - It would have been nice if this was documented better. In particular, perhaps the `--stats` option should not that the default is 1 minute, but the output is hidden by default needing either `-v`, or `--stats-log-level NOTICE`. Doing so would have prevented some user surprise on my end.

rclone v1.42
- os/arch: linux/arm
- go version: go1.10.1

darthShadow:
* Added the corresponding documentation
* Also fixed the python version for the `make_backend_docs.py` & `make_manual.py` scripts

Fixes #2507

### DIFF
--- a/bin/make_backend_docs.py
+++ b/bin/make_backend_docs.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 """
 Make backend documentation
 """
--- a/bin/make_manual.py
+++ b/bin/make_manual.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 """
 Make single page versions of the documentation for release and
 conversion into man pages etc.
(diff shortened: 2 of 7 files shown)


ncw:
Can you split this into two commits please?  One for the python things and the other for the doc changes?  Happy for them to be in the same PR - thanks!

darthShadow:
@ncw Done. Thanks.

ncw:
That is great thank you :-)

PR #2683, opened 2018-10-22
sftp: Ensure file hash checking is really disabled

jfautley:

The Hash() function still attempts to run a remote SSH command to check the SHA/MD5 checksum of a file, even if the `disable_hashcheck` option is set in the configuration file. This change fixes that behaviour and makes `disable_hashcheck` work as expected again.


ncw:

Issue #2604, opened 2018-09-25
--cache-db-path and --cache-chunk-path are using the default value of --cache-dir not the value of --cache-dir set on the command line when using vfs


kieranlee1970:
Added a vfs flag to my rclone mount with cache and crypt and. instead of using the --cache-chunk-path I set. the cache-backend files are being stored in /tmp/rclone

From: https://forum.rclone.org/t/cach-chunk-path-ignored-when-vfs-set/6907

ncw:
I've managed to replicate this problem.

The problem is that `--cache-db-path` and `--cache-chunk-path` are using the default value of `--cache-dir` not the value of `--cache-dir` set on the command line.

This is due to the initialisation sequence of rclone... and the cache backend using the value of `fs.CacheDir` in an `init()` function when `--config-dir` hasn't been parsed yet.

This can be fixed by making the defaults "" and filling them up in `NewFs` - I think re-ordering the init sequence is probably too hard.

Though could set the default and see if it changed in `NewFs` which would mean the default would appear in the command line flags.

Also need to check use of `fs.CacheDir` in the other remotes.

ivandeex:
@kieranlee1970 Does the problem reproduce with rclone 1.54?

ivandeex:
I am going to close this stale ticket.
Also, the cache backend has been obsoleted and replaced by VFS in recent versions of rclone.

PR #2690, opened 2018-10-23
Fixed the docs generation to use `$HOME` instead of specific user directory

### LINKED ISSUE #2687: Replace `home/ncw` with `$HOME`?
Hello, @ncw !

I saw `/home/ncw` in documentation many times. `git grep 'home/ncw'` on you repository and replace `home/ncw` with `$HOME`, maybe?

Thanks.

darthShadow:

Fixes the docs generation to use the generic `$HOME` & `$USER` variables instead of specific values for reference.

Fixes #2687 

#### Checklist

---

Some observations:

* Doesn't work properly when built on Mac.
* User is still set statically in some places where the internal go function is called first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ MANUAL.txt:	MANUAL.md
 	pandoc -s --from markdown --to plain MANUAL.md -o MANUAL.txt
 
 commanddocs: rclone
-	rclone gendocs docs/content/commands/
+	XDG_CACHE_HOME="" XDG_CONFIG_HOME="" HOME="\$$HOME" USER="\$$USER" rclone gendocs docs/content/commands/
 
 backenddocs: rclone bin/make_backend_docs.py
 	./bin/make_backend_docs.py
--- a/docs/content/cache.md
+++ b/docs/content/cache.md
@@ -420,7 +420,7 @@ The remote name is used as the DB file name.
 - Config:      db_path
 - Env Var:     RCLONE_CACHE_DB_PATH
 - Type:        string
-- Default:     "/home/ncw/.cache/rclone/cache-backend"
+- Default:     "$HOME/.cache/rclone/cache-backend"
 
 #### --cache-chunk-path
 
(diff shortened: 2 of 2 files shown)


darthShadow:
So, this fixes the cases where we directly refer to the environment variables for the `$HOME` or the `$USER` but fails where we use the internal function `user.Current().HomeDir`.

Let me know if you have an idea on how to fix that too.

ncw:
A very clever solution - well done :-)

Note that the bits that you've changed in cache.md are autogenerated also by `make backenddocs` - a similar fix to the commanddocs may work there too.  However because of #2604 you'll probably need to fix the below first...

> So, this fixes the cases where we directly refer to the environment variables for the `$HOME` or the `$USER` but fails where we use the internal function `user.Current().HomeDir`.

I think these are the only uses of user.Current which need addressing

```
$ git grep -n user.Current
backend/sftp/sftp.go:160:       usr, err := user.Current()
fs/config/config.go:107:        usr, err := user.Current()
```

I'm thinking maybe that we should set a special environment variable `RCLONE_MAKING_DOCS` and if that is set then get it to return $HOME and $USER directly.

That might require a bit of factoring of user.Current()...  I'd factor this from fs/config/config.go makeConfigPath into a GetHome function and put a test for `RCLONE_MAKING_DOCS` and just return `$HOME` if it is set.

```
	usr, err := user.Current()
	var homedir string
	if err == nil {
		homedir = usr.HomeDir
	} else {
		// Fall back to reading $HOME - work around user.Current() not
		// working for cross compiled binaries on OSX.
		// https://github.com/golang/go/issues/6376
		homedir = os.Getenv("HOME")
	}
```

And make another GetUser function that the sftp backend can call which does similar.

What do you think?

darthShadow:
Sounds good. I will try to get this done by this weekend.

Is it ok if I create a separate helpers module in `lib` for the above two functions?

ncw:
> Sounds good. I will try to get this done by this weekend.

:-)

> Is it ok if I create a separate helpers module in `lib` for the above two functions?

That is probably a good idea if they are imported by a backend.

`lib/user` or something like that?

You could have a go at writing a test for them too!

ncw:
How are you doing with this?  Shall I merge the Makefile part of this and we can leave the other bits for another PR or do you want to carry on with this one?

darthShadow:
I am actually swamped with work right now so I doubt I will be able to do much work on this for at least the next 2 weeks. You can go ahead & merge this set of changes and I will create a separate PR for the `user.Current` changes.

ncw:
No wories - work comes first.

I'll merge this now then.  The changes in `docs/content/cache.md` will get overwritten next time `make docs` is used, but I'm OK with that for the moment!

ncw:
Thank you for this - it is very useful and a great improvement :-)

PR #2695, opened 2018-10-24
yandex: Major rewrite

### LINKED ISSUE #2586: Yandex.Disk: implement full support for Yandex REST API
Current **rclone 1.43** version doesn't [support ](https://rclone.org/overview/#optional-features) all available Yandex.Disk features.
Original code for rclone [Yandex disk backend](https://github.com/dibu28/yandex) was written by @dibu28 in late 2015 and is a bit outdated.
I'd suggest implementing support for at least moving/copying files and directories using [Yandex REST API](https://tech.yandex.com/disk/api/concepts/about-docpage/?ncrnd=4445) (link to rclone forum [thread](https://forum.rclone.org/t/unable-to-move-file-on-yandex-disk-falls-back-to-download-upload-delete/6801))

- [copy a file or folder](https://tech.yandex.com/disk/api/reference/copy-docpage/)
- [move a file or folder](https://tech.yandex.com/disk/api/reference/move-docpage/)

buengese:
This is a very WIP version of my Yandex rewrite. This aims to fix #2586 and make Yandex more maintainable by rewriting it using rclone rest implementation. The optional interfaces asked for in #2586 are not yet implemented however I will do so in the next days. The basic functionality should be working by now.

The main reason for opening the PR this early is that I definitely need help with testing this one. ~~Due to some weirdness with my ISP my connections to Yandex is absolutely abysmal and I cannot even run the integration tests without them constantly timing out or failing for other connection related reasons.~~ 

#### Checklist


ncw:
> This is a very WIP version of my Yandex rewrite. This aims to fix #2586 and make Yandex more maintainable by rewriting it using rclone rest implementation. The optional interfaces asked for in #2586 are not yet implemented however I will do so in the next days. The basic functionality should be working by now.

That is looking very nice, using the rest module with pacer - all very modern!

I haven't given it a thorough going over, but it is looking good :-)

> The main reason for opening the PR this early is that I definitely need help with testing this one. Due to some weirdness with my ISP my connections to Yandex is absolutely abysmal and I cannot even run the integration tests without them constantly timing out or failing for other connection related reasons.

What would you like me to do - run the integration tests and spot problems?  I can have a go with that if you want.

buengese:
Copy, Move, DirMove, PublicLink and About are now implemented and everything should be working as expected now. So if anyone wants to do testings now is the right time. 

It is expected that the Copy, Move, Purge and especially  DirMove integration tests may fail. These operations need some time to complete on Yandex and my not be reflected in List when checked in tests. In my local tests i fixed this by adding delays of up to 5 seconds on these operations but I've left those out for now. The Yandex documentation says you are supposed to track the status of these operations until they are complete but i consider this pretty useless because in all my testing I've never seen one of those operations fail.

The code still still needs some cleanup and especially the error handling needs a few more improvements.

Snousmoumryk:
> This PR changes how authorization works.

Can stable and beta work simultaneously with their tokens in different config files?
Do not want to interfere with regular backups.

buengese:
Sure just name the remote different.

Snousmoumryk:
While creating new remote [yantest] got error in the log:
``ERROR` : Failed to save new token in config file: section 'yantest' not found``
But when the dialog ended, the new remote [yantest] was in config file.

buengese:
This is part of the auth workaround. For now you can ignore this it should work anyway.

Snousmoumryk:
v1.44-029-g1e47b247-beta
**Passed**
- File: upload/download (copy) - both size/mod time and size/MD5
- Directory: upload/download (copy) - both size/mod time and size/MD5
- Server side file: copy/move/moveto
- Server side directory: copy/move/mkdir/rmdir/purge

**Failed** (not implemented yet?)
- Server side file: delete
- Server side directory: delete

ncw:
> Oh yeah you have to run config again. This PR changes how authorization works.

Can you tell if the user is using the old Auth?  If so a helpful message like this one would be great

https://github.com/ncw/rclone/blob/948a5d25c226829e896a6fdea96d191ac9aa5717/backend/onedrive/onedrive.go#L406-L408

buengese:
@ncw I'm not sure if the OAuth changes are final. Maybe you know a better solution. See review.

@Snousmoumryk I cannot reproduce your issue with delete it works fine for me. Can you please send me log?

backend/yandex/yandex.go
```diff
@@ -57,6 +61,17 @@ func init() {
...
 			if err != nil {
 				log.Fatalf("Failed to configure token: %v", err)
+				return
+			}
+			token, err := oauthutil.GetToken(name, m)
+			if err != nil {
+				fmt.Printf("Couldn't read OAuth token (this should never happen).")
+				return
+			}
+			token.TokenType = "OAuth"
```
buengese: Yandex expects the OAuth token type to be "OAuth" however the token returned during  config is a standart "Bearer" token. Right know I rewrite the token during config to change it's type. The old implementation solved this problem by setting the Authorization header manually.

Snousmoumryk:
> I cannot reproduce your issue with delete it works fine for me. Can you please send me log?

Checked test scripts with fresh eyes
> - **Failed**: Server side file: delete

Sorry, my fault, copy-paste catch - no errors found, test **Passed**.
> - **Failed**: Server side directory: delete

```
>rclone lsd yantest:/tmpdir
           0 2018-10-29 10:34:44        -1 tmpsubdir

>rclone ls yantest:/tmpdir
        0 fileintmpdir.txt
        0 tmpsubdir/fileintmpsubdir.txt

>rclone delete -vv --dump headers yantest:/tmpdir		
		
>rclone lsd yantest:/tmpdir
           0 2018-10-29 10:34:44        -1 tmpsubdir

>rclone ls yantest:/tmpdir
```
According to [rclone delete](https://rclone.org/commands/rclone_delete/) doc page */tmpdir* content is expected to be deleted.
*/tmpsubdir* is part of */tmpdir* content, isn't it?

Log file [rclone-subdir-delete-failed.zip](https://github.com/ncw/rclone/files/2524447/rclone-subdir-delete-failed.zip)

ncw:
> According to [rclone delete](https://rclone.org/commands/rclone_delete/) doc page _/tmpdir_ content is expected to be deleted.
> _/tmpsubdir_ is part of _/tmpdir_ content, isn't it?

`rclone delete` only deletes objects but leaves the directory structure alone.  If you want to delete a directory and all of its contents use `rclone purge`

buengese:
I consider the code now ready for review. Only minor changes left. I'll update the docs once the code is 100% finished.

Snousmoumryk:
> I consider the code now ready for review. Only minor changes left.

Great news @buengese! 👍 
Looking forward to the final release.

> `rclone delete` only deletes objects but leaves the directory structure alone.  If you want to delete a directory and all of its contents use `rclone purge`.

@ncw Maybe it makes sense to add these two sentences to the help/online docs of the `delete` command?

It took me a while to accept the logic of when an object attribute (a path in terms of file system) remains after the object itself is deleted. I guess I'm not the only one confused.

ncw:
> @ncw Maybe it makes sense to add these two sentences to the help/online docs of the `delete` command?
> 
> It took me a while to accept the logic of when an object attribute (a path in terms of file system) remains after the object itself is deleted. I guess I'm not the only one confused.

Yes more docs are better I'll add them.

`delete` is something of a throwback command... You can use it with filters which is really useful - eg delete all `*.jpg` files, but it only works with files.

buengese:
@ncw Don't want to annoy you. Just a reminder the main thing this PR is waiting for is a review from you. I'd like to get this done during next week or so because I won't have much time after that.

ncw:
I have had a look through the code - it looks great :-)  Nothing I particularly think we should change.

Some things to discuss:

- If you use this with a config from the old yandex backend then when you run the tests you get `oauth2: token expired and refresh token is not set` errors.  Can we look for the string "refresh_token" in the token and make the user reconfigure the backend if it isn't set? Maybe like this code from the onedrive backend (but return an error don't log.Fatalf - must fix that!)

https://github.com/ncw/rclone/blob/9322f4baef55380a88d2a5d5ce2fa1c12f7e51f0/backend/onedrive/onedrive.go#L406-L408

- ListR support is now gone?  I don't have a strong view on this but I just thought I'd ask you about it.
- Can you update `docs/content/overview.md` with the updated optional features.

- I ran the unit and integration tests with a couple of failures which it would be nice to clean up:

```
go install github.com/ncw/rclone/fstest/test_all
test_all -backends yandex
```

image

```
    --- FAIL: TestIntegration/TestFsDirMove (18.53s)
        fstests.go:906: 
            	Error Trace:	fstests.go:906
            	Error:      	Not equal: 
            	            	expected: *errors.fundamental(directory not found)
            	            	actual  : <nil>(<nil>)
            	Test:       	TestIntegration/TestFsDirMove
        fstest.go:304: Sleeping for 1s for list eventual consistency: 1/6
        fstest.go:298: Sleeping for 6s just to make sure
```

and

```
"./sync.test -test.v -test.timeout 30m0s -remote TestYandex: -test.run '^(TestSyncDeleteDuring)$'" - Starting (try 5/5)
=== RUN   TestSyncDeleteDuring
2018/11/10 14:53:40 ERROR : potato2: Failed to copy: [423 - DiskResourceLockedError] Resource is locked. (Ресурс заблокирован. Возможно, над ним выполняется другая операция.)
2018/11/10 14:53:40 ERROR : Yandex rclone-test-comohod0varihiw9hutaduv3: not deleting directories as there were IO errors
--- FAIL: TestSyncDeleteDuring (9.56s)
    run.go:173: Remote "Yandex rclone-test-comohod0varihiw9hutaduv3", Local "Local file system at /tmp/rclone525352071", Modify Window "1ns"
    require.go:794: 
        	Error Trace:	sync_test.go:579
        	            	[423 - DiskResourceLockedError] Resource is locked. (Ресурс заблокирован. Возможно, над ним выполняется другая операция.)
        	Test:       	TestSyncDeleteDuring
FAIL
"./sync.test -test.v -test.timeout 30m0s -remote TestYandex: -test.run '^(TestSyncDeleteDuring)$'" - Finished ERROR in 11.328096301s (try 5/5): exit status 1: Failed [TestSyncDeleteDuring]
```

I think we are very close - excellent work :-)

buengese:
Thanks for the Review.

> If you use this with a config from the old yandex backend then when you run the tests you get oauth2: token expired and refresh token is not set errors. Can we look for the string "refresh_token" in the token and make the user reconfigure the backend if it isn't set? Maybe like this code from the onedrive backend (but return an error don't log.Fatalf - must fix that!)

I'll definetly have another look at the OAuth Stuff.

>  I ran the unit and integration tests with a couple of failures which it would be nice to clean up:

The Test failures are somewhat expected because certain operations just take some time to complete on Yandex. The tests can be fixed by adding something like `time.Sleep(5 * time.Second)`
to the operations that cause the failure or changing the tests to wait a certain time for operations to finish. I've intentionally not done this because I deemed it unnecessary. Also the wait time depends on the size of the operation e.g. moving a directory with 1000s of files can take up to 30 or mor e seconds. The API returns an URL which can be used to track the status of these operations but for now I'm just ignoring it.

>  ListR support is now gone

ListR has the same problem as the operations above with changes taking some time to be reflected in the file list. With ListR this may also extend to operations like upload or deleting a single file which should be reflected instantly in the normal Listing. I can bring it back if you want.

> Can you update docs/content/overview.md with the updated optional features.

Updating the docs is still on my todo list.

ncw:
> I'll definetly have another look at the OAuth Stuff.

I think just checking the value of "token" in the `configmap.Mapper` passed into `NewFs` and seeing if it has a string "refresh_token" within it and returning an error if it doesn't will be sufficient.

> The Test failures are somewhat expected because certain operations just take some time to complete on Yandex (but they will complete eventually). The tests can be fixed by adding something like `time.Sleep(5 * time.Second)`
> to the operations that cause the failure or changing the tests to wait a certain time for operations to finish. I've intentionally not done this because I deemed it unnecessary. Also the wait time depends on the size of the operation e.g. moving a directory with 1000s of files can take up to 30 or mor e seconds. The API returns an URL which can be used to track the status of these operations but for now I'm just ignoring it.

Ah!  It looks like DirMove is the only troublesome one. We should implement the polling of the URL because `rclone mount` for instance will expect the directory to have moved after `DirMove` returns.  Onedrive has an identical concept

https://github.com/ncw/rclone/blob/c4b0a37b212bb147c118949d960baf2fa0ba31c2/backend/onedrive/onedrive.go#L793



> ListR has the same problem as the operations above with changes taking some time to be reflected in the file list. With ListR this may also extend to operations like upload or deleting a single file which should be reflected instantly in the normal Listing. I can bring it back if you want.

I'm not too bothered either way!

> Updating the docs is still on my todo list.

:-)

buengese:
>Ah! It looks like DirMove is the only troublesome one. We should implement the polling of the URL because rclone mount for instance will expect the directory to have moved after DirMove returns. Onedrive has an identical concept

Ah ok in this case I'll implement it. It actually DirMove and Purge that need this.

The Error in the Sync test is actually related to mkDirs. I'm going to revert mkDirs to the way it was in the original implementation where it pretty much ignores all errors. I'm really not happy with that because this means mkDirs will fail silently but the Yandex docs don't provide a comprehensive list of all Errors and when they're expected to occur.

buengese:
Sorry for the delay. This should finally be done now.

backend/yandex/yandex.go
```diff
@@ -448,48 +444,185 @@ func (f *Fs) PutStream(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption
...
+		RootURL: location,
+		Method:  "GET",
 	}
-	return mkDirFullPath(f.yd, root)
+	deadline := time.Now().Add(fs.Config.Timeout)
+	for time.Now().Before(deadline) {
+		var resp *http.Response
+		var body []byte
+		err = f.pacer.Call(func() (bool, error) {
+			resp, err = f.srv.Call(&opts)
```
ncw: This needs `NoResponse` in opts.

backend/yandex/yandex.go
```diff
@@ -448,48 +444,185 @@ func (f *Fs) PutStream(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption
...
+		Parameters: url.Values{},
+	}
+
+	opts.Parameters.Set("path", path)
+	opts.Parameters.Set("permanently", strconv.FormatBool(hardDelete))
+
+	var resp *http.Response
+	var body []byte
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.Call(&opts)
```
ncw: This needs `NoResponse` in opts.

backend/yandex/yandex.go
```diff
@@ -501,14 +634,242 @@ func (f *Fs) Purge() error {
...
+	}
+
+	opts.Parameters.Set("from", src)
+	opts.Parameters.Set("path", dst)
+	opts.Parameters.Set("overwrite", strconv.FormatBool(overwrite))
+
+	var resp *http.Response
+	var body []byte
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.Call(&opts)
```
ncw: This needs `NoResponse` in opts.

ncw:
Found the leak

```patch
diff --git a/backend/yandex/yandex.go b/backend/yandex/yandex.go
index 73767a107..f237d6834 100644
--- a/backend/yandex/yandex.go
+++ b/backend/yandex/yandex.go
@@ -979,6 +979,7 @@ func (o *Object) setCustomProperty(property string, value string) (err error) {
 		Method:     "PATCH",
 		Path:       "/resources",
 		Parameters: url.Values{},
+		NoResponse: true,
 	}
 
 	opts.Parameters.Set("path", o.filePath())
```

That is what is probably a mis-feature in the rest library

```go
// CallJSON runs Call and decodes the body as a JSON object into response (if not nil)
//
// If request is not nil then it will be JSON encoded as the body of the request
//
// If response is not nil then the response will be JSON decoded into
// it and resp.Body will be closed.
//
// If response is nil then the resp.Body will be closed only if
// opts.NoResponse is set.
```

So if response is `nil` then you are in charge of closing the body.

buengese:
Second time I forgot about this. Thanks would've probably taken me quite some time to notice.

sergeevabc:
@buengese, thank you for improving compatibility with Yandex.

ncw:
Thank looks great now - thank you very much :-)

I'll merge it now.  Check out https://pub.rclone.org/integration-tests/current/ to see how the integration tests fared tomorrow morning!

Snousmoumryk:
@ncw @buengese 
Thank you for your hard work.
I highly appreciate your efforts to fully support Yandex.Disk features.

Issue #2512, opened 2018-09-01
rclone dedupe creating more duplicates that then need to be deduped again


ghost:
#### What is the problem you are having with rclone?

rclone dedupe is creating duplicates if the following situation is given:

file
file
file
file-1
file-2

So when rclone dedupe is being ran on that, this is what happens:

file
file-1
file-1
file-2
file-2

Naturally, this is silly. When I run dedupe again, this is what I get:

file
file-1
file-1-1
file-2
file-2-1

What rclone dedupe *should* do is this:

file
file-1
file-2
file-3
file-4

#### What is your rclone version (eg output from `rclone -V`)

rclone v1.42
- os/arch: openbsd/amd64
- go version: go1.10.1

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

OpenBSD

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone dedupe --dedupe-mode rename remote:

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

First run: rclone -vv dedupe --dedupe-mode rename gdrive:testfiles
         
2018/09/01 14:59:52 DEBUG : rclone: Version "v1.42" starting with parameters ["rclone" "-vv" "dedupe" "--dedupe-mode" "rename" "gdrive:testfiles"]
2018/09/01 14:59:52 DEBUG : Using config file from "/home/nrt/.config/rclone/rclone.conf"
2018/09/01 14:59:52 DEBUG : pacer: Rate limited, sleeping for 1.869441003s (1 consecutive low level retries)
2018/09/01 14:59:52 DEBUG : pacer: low level retry 1/10 (error googleapi: Error 403: Rate Limit Exceeded, rateLimitExceeded)
2018/09/01 14:59:52 DEBUG : pacer: Rate limited, sleeping for 2.785122608s (2 consecutive low level retries)
2018/09/01 14:59:52 DEBUG : pacer: low level retry 2/10 (error googleapi: Error 403: Rate Limit Exceeded, rateLimitExceeded)
2018/09/01 14:59:54 DEBUG : pacer: Rate limited, sleeping for 4.884547223s (3 consecutive low level retries)
2018/09/01 14:59:54 DEBUG : pacer: low level retry 3/10 (error googleapi: Error 403: Rate Limit Exceeded, rateLimitExceeded)
[...]

PR #2706, opened 2018-10-27
dedupe: check for existing filename before renaming a dupe file

ssaqua:

Fixes dedupe with rename strategy choosing names of files that already exist on the remote as described in #2512. Wasn't sure about adding tests since the existing test case doesn't seem to cover this scenario. 

With files in a remote folder as:
```
file
file
file-1
file-2
```

the first two should be renamed to file-3 and file-4

#### Checklist

### DIFF
--- a/fs/operations/dedupe.go
+++ b/fs/operations/dedupe.go
@@ -18,16 +18,32 @@ import (
 )
 
 // dedupeRename renames the objs slice to different names
-func dedupeRename(remote string, objs []fs.Object) {
-	f := objs[0].Fs()
+func dedupeRename(f fs.Fs, remote string, objs []fs.Object) {
 	doMove := f.Features().Move
 	if doMove == nil {
 		log.Fatalf("Fs %v doesn't support Move", f)
 	}
 	ext := path.Ext(remote)
 	base := remote[:len(remote)-len(ext)]
+
+outer:
 	for i, o := range objs {
-		newName := fmt.Sprintf("%s-%d%s", base, i+1, ext)
+		suffix := 1
+		newName := fmt.Sprintf("%s-%d%s", base, i+suffix, ext)
+		_, err := f.NewObject(newName)
+		for ; err != fs.ErrorObjectNotFound; suffix++ {
+			if err != nil {
+				fs.CountError(err)
+				fs.Errorf(o, "Failed to check for existing object: %v", err)
+				continue outer
+			}
+			if suffix > 100 {
+				fs.Errorf(o, "Could not find an available new name")
+				continue outer
+			}
+			newName = fmt.Sprintf("%s-%d%s", base, i+suffix, ext)
+			_, err = f.NewObject(newName)
+		}
 		if !fs.Config.DryRun {
 			newObj, err := doMove(o, newName)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


ncw:
Can you have a go at adding a test?

What you could do is modify this test

https://github.com/ncw/rclone/blob/f7c31cd21028dfb5ada17a5a81a450b177f9cce2/fs/operations/dedupe_test.go#L150

But create a "one-1.txt" file and check it doesn't get overwritten.

You'll need to run those tests with

    go test -v -run TestDeduplicate -remote drive:

Where drive: is a google drive remote for rclone to use.

fs/operations/dedupe.go
```diff
@@ -18,16 +18,23 @@ import (
...
-		log.Fatalf("Fs %v doesn't support Move", f)
+		log.Fatalf("Fs %v doesn't support Move", info)
 	}
 	ext := path.Ext(remote)
 	base := remote[:len(remote)-len(ext)]
 	for i, o := range objs {
-		newName := fmt.Sprintf("%s-%d%s", base, i+1, ext)
+		suffix := 1
+		newName := fmt.Sprintf("%s-%d%s", base, i+suffix, ext)
+		_, err := f.NewObject(newName)
```
ncw: There are 3 possible returns we are interested in here
- `err == nil` - all good!
- `err == fs.ErrorObjectNotFound` - retry
- `err != nil` - some other error

You are ignoring the `some other error` case.

fs/operations/dedupe.go
```diff
@@ -18,16 +18,23 @@ import (
...
+		for err != fs.ErrorObjectNotFound {
```
ncw: Being super paranoid I would put a limit on this loop - say 10 or 100 times maximum!

fs/operations/dedupe.go
```diff
@@ -18,16 +18,23 @@ import (
 )
 
 // dedupeRename renames the objs slice to different names
-func dedupeRename(remote string, objs []fs.Object) {
-	f := objs[0].Fs()
-	doMove := f.Features().Move
+func dedupeRename(f fs.Fs, remote string, objs []fs.Object) {
+	info := objs[0].Fs()
```
ncw: Now that you are passing in `f` you can remove this line and just use `f` in place of `info`

ssaqua:
Thanks for the feedback, just wanted to double check this approach with you before checking in an updated commit.

```Go
outer:
    for i, o := range objs {
        suffix := 1
        newName := fmt.Sprintf("%s-%d%s", base, i+suffix, ext)
        _, err := f.NewObject(newName)
        for ; err != fs.ErrorObjectNotFound; suffix++ {
            if err != nil {
                fs.CountError(err)
                fs.Errorf(o, "Failed to check for existing object: %v", err)
                continue outer
            }   
            if suffix > 100 {
                fs.Errorf(o, "Could not find an available new name")
                continue outer
            }   
            newName = fmt.Sprintf("%s-%d%s", base, i+suffix, ext)
            _, err = f.NewObject(newName)
        }   
        ...
    }
```

For the handling of the possible returns from `NewObject(newName)`, the `if err != nil` block should handle the case where an error is not of the `fs.ErrorObjectNotFound` type that we are looking for. The `if suffix > 100` block should put some upper limit for how long the loop will look for new names. Is the error handling for each of these blocks appropriate? (just skipping over to the next file)

ncw:
> Thanks for the feedback, just wanted to double check this approach with you before checking in an updated commit.
> For the handling of the possible returns from `NewObject(newName)`, the `if err != nil` block should handle the case where an error is not of the `fs.ErrorObjectNotFound` type that we are looking for. The `if suffix > 100` block should put some upper limit for how long the loop will look for new names. Is the error handling for each of these blocks appropriate? (just skipping over to the next file)

Yes I think that looks fine :-)

ssaqua:
Should be ready for another review now 🙂

ncw:
That looks great now :-)  Apologies for the delay, will merge now.

PR #2717, opened 2018-11-02
feat: add --md-only flag

giacomoguiulfo:
Signed-off-by: Giacomo Guiulfo <giacomoguiulfo@gmail.com>

Add support for a global flag, "md-only". If enabled, only metadata of the files will be uploaded to the cloud backends. Also, a metadata header "Mdonly" will be included so that a cloud backend is aware that metadata-only put(s) were intended.

Partially. This PR represent a subset of the changes discussed in https://forum.rclone.org/t/min-age-and-max-age-support-for-atime-and-ctime/2122/2.

#### Checklist

- [ ] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -14,6 +14,7 @@ What happens if you CTRL-C a multipart upload
 */
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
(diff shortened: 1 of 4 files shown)


ncw:
What is the purpose of this change?  It seems to upload 0 size objects only?  Why would you want to do that?  Or have I mis-understood - thanks!

giacomoguiulfo:
@ncm Hello Nick. Yes, the purpose of this change is to be able to upload 0 byte objects with their proper metadata attributes (system attributes, xattr, etc.) similarly to how `cp --attributes-only` works. There is a thread which discusses this topic further https://askubuntu.com/questions/56792/how-to-copy-only-file-attributes-metadata-without-actual-content-of-the-file.

I have to admit that clouds don't really leverage anything similar to this as a feature. So far, its purpose is to help us build one of the components we are developing here at Scality for [Zenko](https://github.com/scality/Zenko).

In short, this component will `rclone copy` metadata from a filesystem, and then a s3-like backend will "read" the filesystem using this metadata as reference. This allows you to manage data from the cloud, without having to copy or move it there. You can find more detailed information about this [here](https://github.com/scality/Zenko-RFCs/pull/5).

Let me know your thoughts about it, thanks!

giacomoguiulfo:
Closing this since it's not a general use case.

PR #2724, opened 2018-11-05
enable softfloat on MIPS arch

sedlund:
MIPS does not have a floating point unit.  Enable softfloat to binaries run on devices that do not have MIPS_FPU enabled in their kernel.

#### Checklist

- [ x] All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).


Cnly:
Hi @sedlund, it seems the `GOARCH` value you defined (`MIPS`) is not in https://golang.org/doc/install/source#environment. Can you confirm it is working for the target devices? Thanks!

sedlund:
sorry, I messed up the case of the GOARCH target.  Updated.

ncw:
@sedlund this looks useful thank you :-)

- Do you know what the consequences of enabling "softfloat" are?  Does it mean that processors with floating point processors will use emulation?
- I think it would be a good idea to do mipsle also
- can you `go fmt` the file to fix the build please?

sedlund:
1. From my understanding of patches from https://github.com/golang/go/issues/18162 it seems the softfloat flag causes the compiler to emit code that does not use hardware FPU.
1. Added mipsle
1. Ran go fmt - travis passed, AppVeyor might have some other problem?

Cnly:
> probably need to enable for mipsle as well (#2725) . I do not have a device to verify with though.

Just confirmed on a mipsle device provided by @tony-sunjian:

```
$ ./rclone -V
rclone v1.44-042-g4a6cfbeb-patch-1-beta
- os/arch: linux/mipsle
- go version: go1.11.2
```

sedlund:
I think I defeated git finally.  :)  Let me know if it works for you.

PR #2729, opened 2018-11-07
s3: support aws config role_arn, credential_source, etc...

eriksw:
When the env_auth option is enabled, the AWS SDK's session constructor
now loads configuration from ~/.aws/config and environment variables,
and credentials per the selected (or default) AWS_PROFILE's settings.

This is accomplished by **NOT** including any Credential provider in the
aws.Config passed to the session constructor: If the Config.Credentials
is non-nil, that will always be used and the user's configuration re
role_arn, credential_source, source_profile, etc... from the shared
config will be completely ignored.

(The conditional creation and configuration of the stscreds Credential
provider is complicated enough that it is not worth re-creating that
logic.)

This change makes it possible for rclone to assume a role before accessing s3, which is essential for some cross-account use cases.

Note: The handling of `credential_source` and `role_arn` by aws-sdk-go from the shared config file is relatively recent. It was not present in the version of aws-sdk-go in rclone 1.44, but the version used by rclone master as of now does include these features.

See: https://github.com/ncw/rclone/issues/1352

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -804,8 +804,21 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 		WithHTTPClient(fshttp.NewClient(fs.Config)).
 		WithS3ForcePathStyle(opt.ForcePathStyle)
 	// awsConfig.WithLogLevel(aws.LogDebugWithSigning)
-	ses := session.New()
-	c := s3.New(ses, awsConfig)
+	awsSessionOpts := session.Options{
+		Config: *awsConfig,
+	}
+	if opt.EnvAuth && opt.AccessKeyID == "" && opt.SecretAccessKey == "" {
+		// Enable loading config options from ~/.aws/config (selected by AWS_PROFILE env)
+		awsSessionOpts.SharedConfigState = session.SharedConfigEnable
+		// The session constructor (aws/session/mergeConfigSrcs) will only use the user's preferred credential source
+		// (from the shared config file) if the passed-in Options.Config.Credentials is nil.
+		awsSessionOpts.Config.Credentials = nil
+	}
+	ses, err := session.NewSessionWithOptions(awsSessionOpts)
+	if err != nil {
+		return nil, nil, err
+	}
+	c := s3.New(ses)
 	if opt.V2Auth || opt.Region == "other-v2-signature" {
 		fs.Debugf(nil, "Using v2 auth")
 		signer := func(req *request.Request) {


ncw:
This looks very useful - thank you :-)

- Can you adjust [this paragraph](https://github.com/ncw/rclone/blob/master/docs/content/s3.md#authentication) in the docs if necessary?
- In your opinion will this cause any backwards compatibility problems?
- I gave tha appveyor build a kick since it appeared to have broken with networking issues.

And thanks for getting your head round the s3 authentication - it is very complicated!

ncw:
> A useful complimentary change to this PR would be an rclone backend config option to specify the particular aws config profile; that'd eliminate the need to specify any AWS SDK-related environment variables at all to use configurations from the standard shared aws config.

So you'd set `env_auth` and `profile` and that would be enough?  That seems like a good change.


> Alternatively, it'd also be possible to handle a _subset_ of the standard-config/default-session-constructor-handled cases by adding more s3 backend options to rclone. I actually implemented [eriksw@bd20f49](https://github.com/eriksw/rclone/commit/bd20f4960d05b35cf514bbe86505070f937c79b4) before realizing that the necessary bits for my own use case actually had been landed in aws-sdk-go, and that it was rclone's populating the Credentials of the config that was breaking things.

What options would you suggest?  I don't have a problem adding more backend options.  If they are of less general use they can be marked as `Advanced` and won't bother users unnecessarily.

eriksw:
Re the docs, I'm not sure that actually needs any adjustment with _this_ change since this is an expansion of what named profile auth options are supported.

It might make sense to mention https://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html but even that is incomplete (it's missing role_arn + credential_source, which is mutually exclusive with role_arn + source_profile); I haven't yet found a doc from AWS that actually lays out the more interesting auth possibilities in one place. (There's _yet another_ interesting one that aws-sdk-go doesn't support yet but boto3 and thus awscli does: credential_process.)

eriksw:
As far as backwards compatibility, I don't think this should cause any issues. It'd be much more straightforward if env_auth was explicitly mutually-exclusive with supplying creds in the backend config, but because I leave the existing Credentials logic entirely in place when there's a key id or secret key defined, I can't think of any scenarios where after merging this things would stop working if they had ever worked. The credentials chain setup done by session.NewSessionWithOptions when Credentials is nil supports a superset of what the one rclone provided did, the sole exception being the static creds.

ncw:
Thank you for going through the backwards compatibility concerns and the docs.

I'll merge this as-is - if you wanted to send a PR with docs tweaks that would be great (but if not also great!).

Thank you for a great change

Nick

PR #2743, opened 2018-11-17
s3: add config info for Wasabi's US-West endpoint

hptasins:
Wasabi has two location, US East and US West, with different endpoint URLs.
When configuring S3 to use Wasabi, provide the endpoint information for both
locations.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -291,7 +291,11 @@ func init() {
 				Provider: "DigitalOcean",
 			}, {
 				Value:    "s3.wasabisys.com",
-				Help:     "Wasabi Object Storage",
+				Help:     "Wasabi US East endpoint",
+				Provider: "Wasabi",
+			}, {
+				Value:    "s3.us-west-1.wasabisys.com",
+				Help:     "Wasabi US West endpoint",
 				Provider: "Wasabi",
 			}},
 		}, {


backend/s3/s3.go
```diff
@@ -269,10 +269,21 @@ func init() {
 				Value: "s3.tor01.objectstorage.service.networklayer.com",
 				Help:  "Toronto Single Site Private Endpoint",
 			}},
+		}, {
+			Name:     "endpoint",
+			Help:     "Endpoint for Wasabi S3 API.\nRequired when using Wasabi.",
+			Provider: "Wasabi",
+			Examples: []fs.OptionExample{{
+				Value:    "s3.wasabisys.com",
```
ncw: Just wondering why you duplicated the "endpoint" setting - I think stuffing it into the existing one would work just as well wouldn't it?  When the config runs it would just give the Wasabi endpoints

backend/s3/s3.go
```diff
@@ -289,10 +300,6 @@ func init() {
 				Value:    "sgp1.digitaloceanspaces.com",
 				Help:     "Digital Ocean Spaces Singapore 1",
 				Provider: "DigitalOcean",
-			}, {
```
ncw: So instead of deleting this stanza, add a new one with the other endpoint?

backend/s3/s3.go
hptasins: I was following the style of the IBM COS section, but that's a much longer list.  Should have looked closer at the Digital Ocean case which is a better match.  I'll change it and push a new version.

hptasins:
I pushed an update to address the feedback.  Please take another look.

ncw:
Perfect - thank you very much :-)

PR #2807, opened 2018-12-01
Add packr2 for static assets

poundifdef:

The purpose of this change is to:

- Move the HTML template for the `serve http` command into an external file, rather than an inline string
- Introduce a dependency on [packr2](https://github.com/gobuffalo/packr/tree/master/v2), which will statically compile templates so that they are included with the rclone binary.
- With this, we will be able to more easily add new templates and static assets to enhance the web-based GUI

This change has been discussed here: https://github.com/gobuffalo/packr/tree/master/v2

#### Checklist


poundifdef:
Okay - the diff looks much better now, though I'm not _quite_ sure if the build new scripts are correct (with the new `packr2` command), or what exactly the issue is with travis. Will continue to dig, would love to know if you're able to point me in the right direction here!

ncw:
I've done a bit of research on packr2.

I don't think the way it works will work for rclone.  I want users to be able to `go get github.com/ncw/rclone` and get essentially the release build.  packr2 seems to assume that you'll build the assets as part of the release process.  Which means for the `go get` install rclone won't have the assets.

I'd rather use a tool which turns that on its head and assumes that the generated assets will be checked in but when developers are changing assets and just want to load them off disk, then they use a different procedure.  Also a tool which supports `go generate` would be the go standard way of doing it.

This tool seems to fit the bill
- https://github.com/shurcooL/vfsgen

It supports `go generate` for making the static assets.  It supports a build tag for dev use, and it has a nice feature that it will serve the static assets gzipped which is cool.

It is also much ligher weight than packr2 which has loads of dependencies.

What do you think about using that instead?

poundifdef:
Sure thing! I'm not so familiar with `go generate` and how specifically we would apply it to the build process here, especially since the vfsgen docs make some assumptions about build tags and how they're used. However! Let me refactor some code and get something working and see if it looks better.

ncw:
> Sure thing! I'm not so familiar with `go generate` and how specifically we would apply it to the build process here, especially since the vfsgen docs make some assumptions about build tags and how they're used. However! Let me refactor some code and get something working and see if it looks better.

Just shout if you need help :-)

The idea is that running `go generate` in that directory would be something that only people who are updating the assets would need to do.  Also only those people would need to use the build tag.

poundifdef:
Ah - so you're not anticipating using the //go:generate directive? Would I check in the generated static asset .go file?

ncw:
> Ah - so you're not anticipating using the //go:generate directive? Would I check in the generated static asset .go file?

Yes we'd want the go generate directive and yes we'd check in the generated file.

poundifdef:
Closed this PR, created a new one here, based on feedback: https://github.com/ncw/rclone/pull/2844

PR #2816, opened 2018-12-06
Fix install for Synology

Webreaper:
7z check doesn't work due to misplaced comma, so installation fails on Synology.

As discussed here: https://forum.rclone.org/t/rclone-install-broken-for-synology-again/7863/5


ncw:
That looks great thank you :-)

I'll merge it then push it up to the live website.

PR #2830, opened 2018-12-10
Revert #2622

mathieucarbou:


ncw:
We've fixed this upstream now, so I'll close this

PR #2844, opened 2018-12-16
Vfsgen for templates

poundifdef:

The purpose of this change is to:

- Move the HTML template for the `serve http` command into an external file, rather than an inline string
- Introduce a dependency on [vfsgen](https://github.com/shurcooL/vfsgen), which will statically compile templates so that they are included with the rclone binary.
- With this, we will be able to more easily add new templates and static assets to enhance the web-based GUI 

This change was originally discussed here: https://github.com/gobuffalo/packr/tree/master/v2
This PR is based on feedback from a previous iteration: https://github.com/ncw/rclone/pull/2807

#### Checklist


poundifdef:
@ncw Hello! Per your comments on a previous PR, I've updated this to use vfsgen. However, the Travis build seems to be failing, but I'm not quite sure why! I didn't see anything obvious in the logs, but I must have missed something.

Is this PR going in the right direction? Are you able to point me to where the build might be failing?

Many thanks!

cmd/serve/httplib/serve/dir.go
```diff
@@ -77,26 +79,36 @@ func (d *Directory) Serve(w http.ResponseWriter, r *http.Request) {
 	defer accounting.Stats.DoneTransferring(d.DirRemote, true)
 
 	fs.Infof(d.DirRemote, "%s: Serving directory", r.RemoteAddr)
+
+	var indexPage = GetTemplate(w)
```
ncw: I would say we don't want to be instantiating the template every time this function is called.

I would instantiate it in `httplib.NewServer` and store the indexTemplate in the `Server` struct.

poundifdef:
@ncw Hello! Thank you for the feedback - I've made the changes to instantiate the template once. The only way I could think to pass that value was via a new parameter in the `NewDirectory()` function. Does this work for you?

(Certainly I'll squash/rebase prior to merging!)

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"html/template"
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+)
+
+// GetTemplate eturns the HTML template for serving directories via HTTP
+func GetTemplate() (tpl *template.Template, e error) {
```
ncw: It is conventional in go to call errors `err` rather than `e`

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+	tpl = nil
```
ncw: You can rely on variables (including return parameters) being initialized to zero, so these two lines are unnecessary

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+	e = nil
+
+	templateFile, err := Assets.Open("index.html")
+	if err != nil {
+		e = errors.New("Failed to open template")
```
ncw: Idiomatic go would be to use `errors.Wrap` to add some context to the error, like this

```go
return nil, errors.Wrap(err, "get template open")
```

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+		return
+	}
+
+	defer func() { _ = templateFile.Close() }()
```
ncw: rclone has a utility function for this

```
defer fs.CheckClose(templateFile, &err)
```

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+
+	templateBytes, err := ioutil.ReadAll(templateFile)
+	if err != nil {
+		e = errors.New("Failed to read template file")
+		return
```
ncw: `return nil, errors.Wrap(err, "get template read")`

cmd/serve/httplib/serve/data/data.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	var templateString = string(templateBytes)
+
+	tpl, err = template.New("index").Parse(templateString)
+	if err != nil {
+		e = err
```
ncw: `return nil, errors.Wrap(err, "get template parse")`

cmd/serve/httplib/serve/dir_test.go
```diff
@@ -2,23 +2,30 @@ package serve
...
 	"net/http/httptest"
 	"net/url"
 	"testing"
 
+	"github.com/ncw/rclone/cmd/serve/httplib/serve/data"
 	"github.com/stretchr/testify/assert"
 )
 
+func GetTemplate() *template.Template {
+	htmlTemplate, _ := data.GetTemplate()
```
ncw: Check the error in the test here so

```go
func GetTemplate(t *testing.T) *template.Template {
    htmlTemplate, err := data.GetTemplate()
    require.NoError(t, err)
    return htmlTemplate
}
```

cmd/serve/httplib/serve/dir_test.go
```diff
@@ -2,23 +2,30 @@ package serve
...
+	return htmlTemplate
+}
+
 func TestNewDirectory(t *testing.T) {
-	d := NewDirectory("z")
+	d := NewDirectory("z", GetTemplate())
```
ncw: This then becomes

```
d := NewDirectory("z", GetTemplate(t))
```

cmd/serve/httplib/serve/dir.go
```diff
@@ -77,26 +79,12 @@ func (d *Directory) Serve(w http.ResponseWriter, r *http.Request) {
 	defer accounting.Stats.DoneTransferring(d.DirRemote, true)
 
 	fs.Infof(d.DirRemote, "%s: Serving directory", r.RemoteAddr)
+
+	var indexTemplate = d.HTMLTemplate
+
 	err := indexTemplate.Execute(w, d)
```
ncw: I don't think `indexTemplate` is used later in the function, so I'd just write

```go
err := d.HTMLTemplate.Execute(w, d)
```

poundifdef:
@ncw Cool beans. I've updated the code to reflect your more idiomatic suggestions, and also squashed the commits. Thanks for taking so much time with this!

ncw:
That looks great now - I'll merge it and we can work on the next part!

It could probably (at some point when we've worked out the workflow) do with some docs in the CONTRIBUTING file to explain how to work on (for development) and rebuild the assets.

PR #2856, opened 2018-12-21
Update drive.md

animosity22:
Fixed the Google Drive API documentation

#### Checklist

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -516,9 +516,7 @@ second that each client_id can do set by Google.  rclone already has a
 high quota and I will continue to make sure it is high enough by
 contacting Google.
 
-However you might find you get better performance making your own
-client_id if you are a heavy user. Or you may not depending on exactly
-how Google have been raising rclone's rate limit.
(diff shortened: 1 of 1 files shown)


ncw:
Thank you for that :-)

Do you think that I should change the default pacer for drive?  At the moment it allows 100 transactions per second - I could easily change that to 10 per second.

animosity22:
@ncw  -  I think that would make sense if the quota defaults for a new API are 1000 per 100 seconds, it seems that 10 would be a much better number. I'm not sure what the rclone key itself is.

ncw:
@animosity22 I've been running a little experiment on this. It looks like for uploading files 10 tps is exactly the right limit for your own key.

image

This is what the rclone key looks like.  It has an overall limit of 1200 QPS and a max/user/100s of 1000 just like a standard key

image

So I think setting the pacer to 10 transactions per second by default for drive would make very little difference in throughput but should reduce the likelihood of a ban enormously.

It should help with the enormous error rate in the rclone key too!

I wonder whether the pacer minimum time should be configurable?

ncw:
I turned the pacer discussion into a new issue #2880

PR #2875, opened 2018-12-29
fstests: fix TestFsName fails when using remote:with/path

Cnly:
This PR makes it possible to specify a subfolder to run the tests in.

Before this, the command `go test -v -remote remote:some/path` causes the `FsName` test to fail because `remote.Name() == "remote:"` is not equal to `remoteName == "remote:some/path"`


ncw:
Good idea - thanks :-)

Issue #2778, opened 2018-11-27
@saisimha Sync only working with parent Folders


saisimha:
Hi, So after a little bit of more digging I was able to narrow down the issue.

So, Sync on Shared Folders is only working one Root Parent Folder. if I try to sync the subFolder, rclone is not able to see the files and Files are being uploaded again .

ncw:
Could you answer these questions for me please? - Thanks!

#### What is your rclone version (output from `rclone version`)



#### Which OS you are using and how many bits (eg Windows 7, 64 bit)



####  Which cloud storage system are you using? (eg Google Drive)



#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)



#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

saisimha:
```

C:\WINDOWS\system32>set BaseDir="C:\rclone"

C:\WINDOWS\system32>cd "C:\rclone"

C:\rclone>echo current directory:
current directory:

C:\rclone>rclone --checksum -vv sync --transfers 12 "K:\Test\Two02" "onedrive:Two02"
2018/11/27 22:01:33 DEBUG : rclone: Version "v1.45" starting with parameters ["rclone" "--checksum" "-vv" "sync" "--transfers" "12" "K:\\Test\\Two02" "onedrive:Two02"]
2018/11/27 22:01:33 DEBUG : Using config file from "C:\\Users\\Sai\\.config\\rclone\\rclone.conf"
2018/11/27 22:01:37 DEBUG : Test_Folder 02/Test_Folder 02 - Sample_Text_04.txt: Size and SHA-1 of src and dst objects identical
2018/11/27 22:01:37 DEBUG : Test_Folder 02/Test_Folder 02 - Sample_Text_03.txt: Size and SHA-1 of src and dst objects identical
2018/11/27 22:01:37 DEBUG : Test_Folder 02/Test_Folder 02 - Sample_Text_06.txt: Size and SHA-1 of src and dst objects identical
2018/11/27 22:01:37 DEBUG : Test_Folder 02/Test_Folder 02 - Sample_Text_05.txt: Size and SHA-1 of src and dst objects identical
2018/11/27 22:01:37 DEBUG : Test_Folder 02/Test_Folder 02 - Sample_Text_05.txt: Unchanged skipping
[...]

PR #2876, opened 2018-12-29
onedrive: fix broken support for "shared with me" folders - fixes #2536, #2778

### LINKED ISSUE #2536: OneDrive and shared folders - listing issue
I just updated to the latest beta of rclone ( v1.43-036-g89be5cad-beta) on what Microsoft Graph code was merged into v 1.43, however on my personal OneDrive doing an rclone lsd on a Shared Folder only works on the first level of directories, example: `rclone lsd myonedrive:sharedfolder1/`.

Trying a `rclone lsd myonedrive:sharedfolder1/sharedfolder2/` delivers this error:
```
2018/09/05 00:32:31 ERROR : : error listing: directory not found
2018/09/05 00:32:31 Failed to lsd: directory not found.
```
I did checked and sharedfolder2 in fact exists in the directory tree (this works well ok on non-shared folders).

The same happens when trying to copy files from `myonedrive:sharedfolder1/sharedfolder2/` to my local instance as it says
```
2018/09/05 00:34:31 ERROR : : error listing: directory not found
2018/09/05 00:34:31 Failed to lsd: directory not found
```

Any ideas on why listing directories is not working properly on shared folders?

This may be related to what is described here [https://forum.rclone.org/t/rclone-is-not-working-outside-of-the-root-of-crypt-remote/6697/2](url) for crypt folders.

Cnly:
This PR should fix the compatibility issues for "shared with me" folders in OneDrive Personal.

With these changes, in some infrequent cases when using OneDrive Personal, an extra query will be made when reading the metadata of a path. Also, moving between a shared folder and a normal (local) folder will fall back to copying+deleting because the `Update` API does not allow moving between drives. Otherwise rclone should behave the same as before.

Some extra work may be needed in order to automate the testing for "shared with me" folders in the CI platforms:
1. The tests should be run in a shared folder in addition to a normal one, for example with `go test -v -remote onedrive:shared_folder/deeper/path` (See also https://github.com/ncw/rclone/pull/2875)
1. It would be the best if operations like `Move` and `Copy` can be automatically tested between shared folders and normal folders, because shared folders are actually considered as something in a different drive.

@ncw What changes do you think we should make to the automated tests?

### DIFF
--- a/backend/onedrive/api/types.go
+++ b/backend/onedrive/api/types.go
@@ -285,6 +285,7 @@ type AsyncOperationStatus struct {
 
 // GetID returns a normalized ID of the item
 // If DriveID is known it will be prefixed to the ID with # seperator
+// Can be parsed using onedrive.parseNormalizedID(normalizedID)
 func (i *Item) GetID() string {
 	if i.IsRemote() && i.RemoteItem.ID != "" {
 		return i.RemoteItem.ParentReference.DriveID + "#" + i.RemoteItem.ID
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -334,26 +334,82 @@ func shouldRetry(resp *http.Response, err error) (bool, error) {
 	return authRety || fserrors.ShouldRetry(err) || fserrors.ShouldRetryHTTP(resp, retryErrorCodes), err
 }
 
-// readMetaDataForPath reads the metadata from the path
-func (f *Fs) readMetaDataForPath(path string) (info *api.Item, resp *http.Response, err error) {
-	var opts rest.Opts
-	if len(path) == 0 {
-		opts = rest.Opts{
-			Method: "GET",
-			Path:   "/root",
-		}
-	} else {
-		opts = rest.Opts{
-			Method: "GET",
-			Path:   "/root:/" + rest.URLPathEscape(replaceReservedChars(path)),
-		}
-	}
+// readMetaDataForPathRelativeToID reads the metadata for a path relative to an item that is addressed by its normalized ID.
+// if `relPath` == "", it reads the metadata for the item with that ID.
+func (f *Fs) readMetaDataForPathRelativeToID(normalizedID string, relPath string) (info *api.Item, resp *http.Response, err error) {
+	opts := newOptsCall(normalizedID, "GET", ":/"+rest.URLPathEscape(replaceReservedChars(relPath)))
 	err = f.pacer.Call(func() (bool, error) {
 		resp, err = f.srv.CallJSON(&opts, nil, &info)
 		return shouldRetry(resp, err)
 	})
 
 	return info, resp, err
+}
+
+// readMetaDataForPath reads the metadata from the path (relative to the absolute root)
+func (f *Fs) readMetaDataForPath(path string) (info *api.Item, resp *http.Response, err error) {
+	firstSlashIndex := strings.IndexRune(path, '/')
+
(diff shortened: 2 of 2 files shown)


ncw:
> With these changes, in some infrequent cases when using OneDrive Personal, an extra query will be made when reading the metadata of a path. Also, moving between a shared folder and a normal (local) folder will fall back to copying+deleting because the `Update` API does not allow moving between drives. Otherwise rclone should behave the same as before.

Those sound like great changes :-)

> Some extra work may be needed in order to automate the testing for "shared with me" folders in the CI platforms:
> 
> 1. The tests should be run in a shared folder in addition to a normal one, for example with `go test -v -remote onedrive:shared_folder/deeper/path` (See also #2875)
> 2. It would be the best if operations like `Move` and `Copy` can be automatically tested between shared folders and normal folders, because shared folders are actually considered as something in a different drive.
> 
> @ncw What changes do you think we should make to the automated tests?

Can we make shared folders via the API?  If so then I think some backend specific tests with shared folders we make would probably be the best solution.  See [drive_internal_test.go](https://github.com/ncw/rclone/blob/master/backend/drive/drive_internal_test.go) for an example of how to do this.

If we have to make the shared folders in advance then that is going to make things trickier!

Cnly:
> Can we make shared folders via the API?

The problem is we need to have a folder from another drive. So I'm actually not expecting those specific tests to just work whenever one runs `go test`. I think if we can have the tests run on the CIs or pub.rclone.org it would already be great.

ncw:
> > Can we make shared folders via the API?
> 
> The problem is we need to have a folder from another drive. So I'm actually not expecting those specific tests to just work whenever one runs `go test`. I think if we can have the tests run on the CIs or pub.rclone.org it would already be great.

We could make a specially named remote which we use as "the second account" - could the tests then use that to automatically create shared things?

Cnly:
> We could make a specially named remote which we use as "the second account"

rclone determines whether an item is shared by another user by looking for the `remoteItem` JSON object in the server's response to a get item query. Therefore, if we simulate a new account using a local folder, rclone won't be able to know it because the server will not include the `remoteItem` part when rclone looks for it.

Also, it seems that the API cannot add a shared folder to a user's directory tree. It can create sharing links and access the items directly through those links, but it's not the same thing as "shared with me" folders which appear in a user's OneDrive root directory.

ncw:
> Also, it seems that the API cannot add a shared folder to a user's directory tree. It can create sharing links and access the items directly through those links, but it's not the same thing as "shared with me" folders which appear in a user's OneDrive root directory.

Not being able to create "shared with me" directories over the API is a show stopper I think...

Could we create some "shared with me" directories in advance in the test account and if they were present run the tests?

Cnly:
> Could we create some "shared with me" directories in advance in the test account and if they were present run the tests?

I think that's it. This way the tests can be run at least once every day. Am I right?

ncw:
> I think we'd better also add tests for moving and copying between shared and normal folders, but we can't specify the subdirs to run those tests now. Another thing I see is that the Union remote seems to lack similar tests between the remotes being unified, too. Maybe these two can be dealt with together?

Well my preference is for the tests to create any infrastructure they need during the tests if at all possible.  It sounds like it isn't possible here, but if we can rely on the minimum amount of supporting infrastructure for which the tests can be `t.Skip()`-ed if it isn't found.

Do you want to have a go at this as part of this PR?  Or would you rather I merge this one and we can think about more tests as part of a different PR?

PR #2898, opened 2019-01-06
Add cookie support with cmdline switch --cookie

qip:
Cookies are handled by cookiejar in memory with fshttp module through
the entire session.
One useful scenario is, with HTTP storage system where index server
adds authentication cookie while redirecting to CDN for actual files.
Also, it can be helpful to reuse fshttp in other storage systems
requiring cookie.

To make fshttp module support cookiejar.

No.

#### Checklist

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -87,7 +87,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.IntVarP(flagSet, &fs.Config.MaxBacklog, "max-backlog", "", fs.Config.MaxBacklog, "Maximum number of objects in sync or check backlog.")
 	flags.BoolVarP(flagSet, &fs.Config.StatsOneLine, "stats-one-line", "", fs.Config.StatsOneLine, "Make the stats fit on one line.")
 	flags.BoolVarP(flagSet, &fs.Config.Progress, "progress", "P", fs.Config.Progress, "Show progress during transfer.")
-	flags.BoolVarP(flagSet, &fs.Config.Cookie, "cookie", "", fs.Config.Cookie, "Enable session cookiejar.")
+	flags.BoolVarP(flagSet, &fs.Config.Cookie, "use-cookies", "", fs.Config.Cookie, "Enable session cookiejar.")
 }
 
 // SetFlags converts any flags into config which weren't straight foward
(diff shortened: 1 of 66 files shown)


qip:
I'm not a golang expert, so this commit has only passed existing tests, anyone would like to spare some time helping me a little?
Anyway, I have a nginx index server which returns 301 to actual files with set-cookie header, and this commit works with that with HTTP backend, so I assume it's functional... Too bad not an actual test case :(

fs/config/configflags/configflags.go
```diff
@@ -87,6 +87,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.IntVarP(flagSet, &fs.Config.MaxBacklog, "max-backlog", "", fs.Config.MaxBacklog, "Maximum number of objects in sync or check backlog.")
 	flags.BoolVarP(flagSet, &fs.Config.StatsOneLine, "stats-one-line", "", fs.Config.StatsOneLine, "Make the stats fit on one line.")
 	flags.BoolVarP(flagSet, &fs.Config.Progress, "progress", "P", fs.Config.Progress, "Show progress during transfer.")
+	flags.BoolVarP(flagSet, &fs.Config.Cookie, "cookie", "", fs.Config.Cookie, "Enable session cookiejar.")
```
ncw: I think my preference for the name of this flag would be "use-cookies"

qip: Actually I'm working on keeping cookies in an external cookiejar file as another option.
Now that rethinking about the naming convention, we could use `--use-cookies` to enable cookie support, and `--cookie-jar-file filename` to load and save cookies (which implies `--use-cookies`); Or, use `--cookies` to enable session cookie, and `--cookies filename` to enable cookie with loading and saving cookies.
How do you think?

qip:
Thank you for reviewing!

> Can you drop all the `docs` changes - I update the docs as part of the release process normally.

Sure, I'll update it together with other updates in next commit.

> Do you think this should be enabled normally?

I think this adds a little bit of overhead, and it doesn't help for backends with proper API like Box or Google Drive, so I'd say we enable it manually by the switch.

ncw:
The CI is complaining about the formatting.  Can you run `fs/fshttp/http.go` through `go fmt` or `goimports`?

qip:
Oh now I see what's the diff in CI log is about. Thanks for the reminder!

ncw:
> Oh now I see what's the diff in CI log is about. Thanks for the reminder!

It is annoying difficult to see! I'm going to fix this shortly by using golangci-lint instead of the rather bodged up system I'm using at the moment!

That looks great now - will merge shortly!

PR #2925, opened 2019-01-18
Add support for '-l' option

yairm:
Add support for the '-l' option, translating symbolic links. 

[issue 1152](https://github.com/ncw/rclone/issues/1152)

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -2,6 +2,7 @@
 package local
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -28,6 +29,7 @@ import (
 
 // Constants
 const devUnset = 0xdeadbeefcafebabe // a device id meaning it is unset
+const linkSuffix = ".rclonelink"    // The suffix added to a translated symbolic link
 
 // Register with Fs
 func init() {
@@ -48,6 +50,13 @@ func init() {
 			Default:  false,
 			NoPrefix: true,
 			ShortOpt: "L",
+			Advanced: true,
+		}, {
+			Name:     "links",
+			Help:     "Translate symlinks to/from regular files with a '" + linkSuffix + "' extension",
+			Default:  false,
+			NoPrefix: true,
+			ShortOpt: "l",
 			Advanced: true,
 		}, {
 			Name: "skip_links",
@@ -93,12 +102,13 @@ check can be disabled with this flag.`,
 
 // Options defines the configuration for this backend
 type Options struct {
-	FollowSymlinks bool `config:"copy_links"`
-	SkipSymlinks   bool `config:"skip_links"`
-	NoUTFNorm      bool `config:"no_unicode_normalization"`
-	NoCheckUpdated bool `config:"no_check_updated"`
-	NoUNC          bool `config:"nounc"`
-	OneFileSystem  bool `config:"one_file_system"`
+	FollowSymlinks    bool `config:"copy_links"`
+	TranslateSymlinks bool `config:"links"`
+	SkipSymlinks      bool `config:"skip_links"`
+	NoUTFNorm         bool `config:"no_unicode_normalization"`
+	NoCheckUpdated    bool `config:"no_check_updated"`
+	NoUNC             bool `config:"nounc"`
+	OneFileSystem     bool `config:"one_file_system"`
 }
 
 // Fs represents a local filesystem rooted at root
@@ -120,13 +130,14 @@ type Fs struct {
 
 // Object represents a local filesystem object
 type Object struct {
-	fs      *Fs    // The Fs this object is part of
-	remote  string // The remote path - properly UTF-8 encoded - for rclone
-	path    string // The local path - may not be properly UTF-8 encoded - for OS
-	size    int64  // file metadata - always present
-	mode    os.FileMode
-	modTime time.Time
-	hashes  map[hash.Type]string // Hashes
+	fs             *Fs    // The Fs this object is part of
+	remote         string // The remote path - properly UTF-8 encoded - for rclone
+	path           string // The local path - may not be properly UTF-8 encoded - for OS
+	size           int64  // file metadata - always present
+	mode           os.FileMode
+	modTime        time.Time
+	hashes         map[hash.Type]string // Hashes
+	translatedLink bool                 // Is this object a translated link
 }
 
 // ------------------------------------------------------------
@@ -166,13 +177,27 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	if err == nil {
 		f.dev = readDevice(fi, f.opt.OneFileSystem)
 	}
-	if err == nil && fi.Mode().IsRegular() {
+	if err == nil && f.isRegular(fi.Mode()) {
 		// It is a file, so use the parent as the root
 		f.root = filepath.Dir(f.root)
 		// return an error with an fs which points to the parent
 		return f, fs.ErrorIsFile
 	}
 	return f, nil
+}
+
+// Determine whether a file is a 'regular' file,
+// Symlinks are regular files, only if the TranslateSymlink
+// option is in-effect
+func (f *Fs) isRegular(mode os.FileMode) bool {
+	if !f.opt.TranslateSymlinks {
+		return mode.IsRegular()
+	}
+
+	// fi.Mode().IsRegular() tests that all mode bits are zero
+	// Since symlinks are accepted, test that all other bits are zero,
+	// except the symlink bit
+	return mode&os.ModeType&^os.ModeSymlink == 0
 }
 
 // Name of the remote (as passed into NewFs)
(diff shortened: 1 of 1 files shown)


ncw:
Can you rebase this on master, fixup the conflict(s) and force push the result?

The CI doesn't appear to want to run unless the PR merges without conflict.

Thanks

backend/local/local.go
```diff
@@ -48,6 +49,13 @@ func init() {
 			NoPrefix: true,
 			ShortOpt: "L",
 			Advanced: true,
+		}, {
+			Name:     "translate_links",
```
ncw: To stay in line with `rsync` this needs to be called `links`

docs/content/local.md
```diff
@@ -120,6 +120,73 @@ $ rclone -L ls /tmp/a
         6 b/one
 ```
 
+#### --translate-links, -l ####
```
ncw: Very nice docs :star:

yairm:
changed 'translated_links' to 'links'

Regarding rebase.

I tried to rebase, however, I don't know how to resolve the conflicts in the Update() function. 
The nopWriterCloser doesn't seem to work well with the new 'File' type that was introduced

Should I rebase, anyway, making the code non-compiling?

ncw:
It doesn't look like your rebase has worked...

You need to do something like this

```
git co master
git pull
git co TranslateSymlinks
git rebase master
# fix conflicts
```

Here is [github's help on fixing rebase conflicts](https://help.github.com/articles/resolving-merge-conflicts-after-a-git-rebase/) and here is the article on [fixing merge conflicts](https://help.github.com/articles/resolving-a-merge-conflict-using-the-command-line/) - I hope that helps!

Once you've rebased your branch then force push it.

Remember if anything goes wrong, you can just do `git rebase --abort` to start again.

If you want me to do the rebase I can do it then push it back to your fork.

yairm:
I think there was a slight misunderstanding.
I think I know how to "git rebase", however, the issue was that I didn't know the best way to resolve the merge issues that I encountered.

At your request, I performed the rebase, and I'd like your comments regarding this code in the update function:

`	if !o.translatedLink {
		f, err := file.OpenFile(o.path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
		if err != nil {
			return err
		}
		// Pre-allocate the file for performance reasons
		err = preAllocate(src.Size(), f)
		if err != nil {
			fs.Debugf(o, "Failed to pre-allocate: %v", err)
		}
		out = f
	} else {
		out = nopWriterCloser{new(strings.Builder)}
	}
`

I introduced the nopWriterCloser, to facilitate calculating the hash without any additional code changes to the hash logic.
The main idea, is that the contents of the symbolic link acts as a file, and the hash is calculated without knowing whether it is calculating the hash of a file or of a symlink.

However, the new preallocate() call, doesn't accept a writerCloser, and therefore, I did the above trickery.

backend/local/local.go
```diff
@@ -827,15 +914,22 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
...
+			return err
+		}
+		// Pre-allocate the file for performance reasons
+		err = preAllocate(src.Size(), f)
+		if err != nil {
+			fs.Debugf(o, "Failed to pre-allocate: %v", err)
+		}
+		out = f
+	} else {
+		out = nopWriterCloser{new(bytes.Buffer)}
```
ncw: Then this can be `ioutil.NopCloser(&symlinkData)`...

backend/local/local.go
```diff
@@ -827,15 +914,22 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
...
-	}
-
-	// Pre-allocate the file for performance reasons
-	err = preAllocate(src.Size(), out)
-	if err != nil {
-		fs.Debugf(o, "Failed to pre-allocate: %v", err)
+	// If the object is a regular file, create it.
+	// If it is a translated link, just read in the contents, and
+	// then create a symlink
+	if !o.translatedLink {
```
ncw: I would put this above the if

    var symlinkData bytes.Buffer

backend/local/local.go
```diff
@@ -850,6 +944,27 @@ func (o *Object) Update(in io.Reader, src fs.ObjectInfo, options ...fs.OpenOptio
...
+		if err == nil {
+			// Remove any current symlink or file, if one exsits
+			if _, err := os.Lstat(o.path); err == nil {
+				if removeErr := os.Remove(o.path); removeErr != nil {
+					fs.Errorf(o, "Failed to remove previous file: %v", removeErr)
+					return removeErr
+				}
+			}
+			// Use the contents for the copied object to create a symlink
+			nwc, _ := out.(nopWriterCloser)
```
ncw: And you can replace these two lines with

    err = os.Symlink(symlinkData.String(), o.path)

And your `nopWriterCloser` becomes redundant (clever though it is!)

backend/local/local.go
yairm: The ioutil.NopCloser is a ReadCloser (while the nopWriterCloser was a writeCloser).

The io.copy(out, in) copies data from 'in' to 'out', and we can't set 'out' to the suggested nopCloser, since it's of the wrong type (it's a reader and not a writer).

What am I missing?

yairm:
@ncw :
1. Please see my comment above to your suggestion
2. I think the code is ready for code review.
3. Regarding testing, if you could provide me with an example for a test that uses '-l', I can follow suit, and add more comprehensive tests.

ncw:
> 1. Please see my comment above to your suggestion

I replied inline - I don't know if you saw that.

> 2. I think the code is ready for code review.

OK

> 3. ci/circleci failed - I looked at the reason, and I don't understand the reason for the failure. I have a feeling it is not related to the change I'm introducing..

That is a recently introduced problem which I'll fix - don't worry about it!

> 4. Regarding testing, if you could provide me with an example for a test that uses '-l', I can follow suit, and add more comprehensive tests.

The way to write the tests is to use the internal integration tests framework - here is an example

https://github.com/ncw/rclone/blob/ea12d76c0354c8c099e253a7701f7dc7d848f95c/backend/drive/drive_internal_test.go#L272

I would say this will be reasonably tricky though if you don't know the integration test framework...

If you'd like me to do this bit I will as part of reviewing the code.

yairm:
> > 1. Please see my comment above to your suggestion
> 
> I replied inline - I don't know if you saw that.
Just saw it - 
Removed the type assertion.   

> 
> I would say this will be reasonably tricky though if you don't know the integration test framework...
> 
> If you'd like me to do this bit I will as part of reviewing the code.
Taking on your suggestion to write the tests, thank you!

yairm:
A few thoughts:

Thought #1:
The List() function must add an .rclonesuffix to symlinks it encounters.

The reason is, that otherwise, the operations of 'copyto', etc.. wouldn't work correctly.
Those commands rely on the names returned from List(), and use them to create the destination files in the remote.

Thought #2:
Due to the above, I've made a slight change change to the documentation. What we really want to say is that '-l/--links' changes the way the local backend "sees" symbolic links. The local backend "sees" them as text files with a .rclonelink suffix.

This is apparent from the following example:

$ rclone ls ./
2019/01/26 22:24:23 NOTICE: linktotestfile: Can't follow symlink without -L/--copy-links

$ rclone -l ls ./ 
   9 **linktotestfile.rclonelink**

I didn't push the change to the documentation, not to disturb the code review process.

Thought #3:
Currently, I don't think anything is stopping someone from using both '-L' and '-l'.
I'm not sure if the code deals with this in a consistent way (preferring always one option over the other).
I think they should be mutually exclusive, and so documented.

ncw:
> A few thoughts:
> 
> Thought 1:
> The List() function must add a .rclonesuffix to symlinks it encounters.
> 
> The reason is, that otherwise, the operations of 'copyto', etc.. wouldn't work correctly.
> Those commands rely on the names returned from List(), and use them to create the destination files in the remote.
> 
> This is the current state of the code, in the pull request.

I've made tests for that
 
> Thought 2:
> Due to the above, I've made a slight change change to the _documentation._ What we really want to say is that '-l/--links' changes the way the local backend "sees" symbolic links. The local backend "sees" them as text files with a .rclonelink suffix.
> 
> This is apparent from the following example:
> 
> $ rclone ls ./
> 2019/01/26 22:24:23 NOTICE: linktotestfile: Can't follow symlink without -L/--copy-links
> 
> $ rclone -l ls ./
> 9 **linktotestfile.rclonelink**
> 
> I didn't push the change to the documentation, not to disturb the code review process.

OK - they can come as an update
 
> Thought 3:
> Currently, I don't think anything is stopping someone from using both '-L' and '-l' on a single command.
> I'm not sure if the code deals with this in a consistent way (preferring always one option over the other).
> I think they should be mutually exclusive, and so documented.

Hmm, good point.

I'll add a bit of code to check and a test.

----

I have discovered one small gotcha, running `os.Chtimes` sets the modification time on the destination of the link, not the link file itself (which is what we want for `-l` mode).  Unfortunately there isn't an easy way of doing this without delving into `sys`.  I'll have a go at fixing this shortly...

The testing is going well - I'm using a coverage guided approach and I think I'm just about there!

ncw:
OK I've done the tests and fixed the above and merged into the fix-1152-links branch :-)

I'll close this PR now.  If you want to send further changes can you do them against that branch please?

I want to merge this to master ASAP in time for the release on the 7th Feb.

yairm:
Thank you for helping me with the tests, and for finding the os.Chtimes issue(!)

I'm not sure about the wording in the documentation. I think I'm happy with the current version.

I would like to thank you for this experience, and for your wonderful rclone project.

ncw:
> Thank you for helping me with the tests, and for finding the os.Chtimes issue(!)
> 
> I'm not sure about the wording in the documentation. I think I'm happy with the current version.

If you find something you are unhappy with then please send a PR :-)

> I would like to thank you for this experience, and for your wonderful rclone project.

It has been a pleasure - looking forward to your next contribution :-)

I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.46

PR #2938, opened 2019-01-27
Swift: introduce application credential auth support

### LINKED ISSUE #2777: Add Openstack application credential support

1.45

Openstack application credential is not supported: https://docs.openstack.org/keystone/queens/user/application_credentials.html

Here is the [code example](https://github.com/gophercloud/gophercloud/blob/3a7818a07cfc862267a0b03635075a444c92dcda/openstack/auth_env.go#L41..L43) on how it is done in gophercloud

Handle application credentials. [API reference](https://developer.openstack.org/api-ref/identity/v3/index.html?expanded=authenticating-with-an-application-credential-detail#application-credentials).

kayrus:
Resolves #2777


kayrus:
Split the commit. In addition, added the `auth v3` note into the doc.

kayrus:
Is there a way to force recheck without the commit? appveyor.com is quite buggy.

ncw:
That looks great now - will merge in a second.

I broke the appveyor build :-( will fix it up shortly!

PR #2940, opened 2019-01-28
WIP: Memory reduction

splitice:
Work in progress to reduce memory usage within the cache remote.

Currently:
 - re-ordering object struct (minor)
 - investigating mmap


splitice:
- Added buffer re-use within cache instead of mmap. This way native array slices can continue to be used.
- Memory cache now takes a copy of the chunk so that the chunk can be reduced in size (previously required full chunk size memory consumption regardless).

splitice:
Memory usage from 2GB during active plex streaming to 600MB. Still much more possible.

Also currently upload is not optimized only download.

splitice:
Support for direct writes into the targeted buffer within cache added. This makes uploads use less temporary buffer allocations.

splitice:
@ncw The difference with this PR is night and day.

Any chance you could check it (and definitely clean it up!)

ncw:
Firstly - excellent work on reducing memory use :-) That has been a problem with the cache for a long time.

Can you
- rebase on master (that will fix the appveyor tests)
- drop the lib/mmap and asyncreader commits as I don't think you are using them
- squash your commits
- rename your commit `cache: ...`
- it would be really helpful for future readers if you could explain in your commit message more about what you changed.  I like the before/after style here as it focuses the mind on exactly what you changed: eg 06ae4258be1833968e9d3d3c07daa726f3a3f086
- add a seperate commit "vendor: add github.com/90TechSAS/go-buffer-pool-recycler" following the instructions here: https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency - that commit should go before your other commit - that should fix the travis build hopefully

Then I'll look through the code one more time :-)

Thanks

Nick

PS Perhaps I should use the buffer pool recycler in the asyncreader instead of lib/mmap?

splitice:
@ncw 

1. I'll periodically squash when my commit list gets bigger if that's ok with you. I don't think I'm done yet. I plan to squash everything into 1 or a few commits upon completion before merge with appropriate messages for the changes that make the cut. I've largely been pushing to take advantage of your travis-ci builds on the PR.
2. I did a (largely) artificial stress test today and got the memory growth up to 700MB. So I wouldn't consider it solved (just largely improved). I believe the growth in this case was in that the upload through cache path still uses go allocated arrays. Not really sure about how to solve this since those buffers are provided into getChunk. Perhaps mmap.
3. Could you merge the master branch with the mmap branch? This is based off that so I can bring that up easier.
4. Ideally all large buffer allocations (including those from the pool) with mmap. Not sure if mmaped allocations can be sliced in go.
5. With the new dependency perhaps it would be better just to roll our own, a buffer pool with mmap's could be used in other layers - not just cache.

One thing I've also observed is that it would be better I think if there was a disk chunk size and a read chunk size. It's better for the filesystem to have larger chunks, but it's suboptimal to have to read each chunk in it's entirely before it can be passed on. I've been running 64MB chunks and this results in a read latency that ranges between 1-3 seconds (RAID 10 volume over Infiniband)

splitice:
Regarding point 4 it looks like the answer is yes - but not safely. 

https://stackoverflow.com/questions/39215392/go-how-to-convert-unsafe-pointer-into-pointer-to-array-with-unknown-length

Derfinately something to be carefully defensively architected around.

splitice:
After completing this merge it's not working as well. Really highlighting that I need to know more Go before jumping into something like this.

Perhaps instead you could take a look at perhaps implementing buffer re-use within cache. This experiment showed substantial improvements from doing so :)

ncw:
> After completing this merge it's not working as well

In what way?  Maybe it is something I broke as part of changes to lib/mmap and lib/pool.

You seem to have set the duration to 0 in pool.New - I think you'll want something higher than that otherwise the memory will never be flushed would be my guess.

```
pool.New(0, int(r.cfs.opt.ChunkSize), int(r.cfs.opt.TotalWorkers), fs.Config.UseMmap)
```

I'm not sure cache is safe to use with lib/mmap so `false` rather than `fs.Config.UseMmap` might be better.

I note you found a bug in lib/pool - thanks!  That is probably a consequence of setting the cache flush interval to 0 which I'm not sure is a good idea!



> Perhaps instead you could take a look at perhaps implementing buffer re-use within cache. This experiment showed substantial improvements from doing so :)

I think you are doing fine and showing great go skills :-)

ncw:
Are you going to work on this some more?  Or shall we close it for now?

PR #2942, opened 2019-01-28
jottacloud: add support for 2-factor authentification fixes #2722

### LINKED ISSUE #2722: Add 2FA support for jottacloud
v1.43 AFAIK the feature is not added in 1.44

Backing up to jottacloud with two factor enabled
Now, this results in: 
Failed to create file system for "jottapriv:freenas": couldn't get account info: failed to get endpoint url: error 401: org.springframework.security.authentication.BadCredentialsException: Invalid. Basic-auth disabled when 2FA is activated. (Unauthorized)

Handle Two factor authentication for jottacloud
The official (beta) cli-client jotta-cli (https://docs.jottacloud.com/jottacloud-command-line-tool) handles this. The login process triggers a two factor SMS, and it asks you for the OTP before trying to connect. I would guess it gets some kind of device token or other to authenticate, as it stores this for future connections.

buengese:
Quick PR that implements support for 2fa to the jottacloud backend. Should also improve error handling during config a bit.

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


ncw:
That looks nice an easy!

I'll merge that in just a moment :-)

Thank you.

Issue #2832, opened 2018-12-11
Can't copy Google documents to OneDrive


izzobot:
#### What is the problem you are having with rclone?

I have a daily sync that copies all of my google documents from Google Drive to OneDrive but for a while now it's not been working properly, no files are copied/synced, as you can see by the logs below.
It works fine if I transfer from Google Drive to local.
Please let me know if you need any more information or if I have misconfigured/misunderstood something.

Thank you.

#### What is your rclone version (output from `rclone version`)

rclone v1.45
- os/arch: linux/amd64
- go version: go1.11.2

(I have also tried the latest beta and I get the same result)

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

4.18.20-unRAID GNU/Linux
Windows 10, 64 bit
(It occurs on both)

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive and OneDrive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`rclone sync "goo:Google Suite/Docs/document.docx" "one:temp/" -vv`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
rclone sync "goo:Google Suite/Docs/document.docx" "one:temp/" -vv
2018/12/11 00:43:50 DEBUG : rclone: Version "v1.45" starting with parameters ["rcloneorig" "--config" "/boot/config/plugins/rclone/.rclone.conf" "sync" "goo:Google Suite/Docs/document.docx" "one:temp/" "-vv"]
2018/12/11 00:43:50 DEBUG : Using config file from "/boot/config/plugins/rclone/.rclone.conf"
2018/12/11 00:43:55 INFO  : One drive root 'temp': Waiting for checks to finish
2018/12/11 00:43:55 INFO  : One drive root 'temp': Waiting for transfers to finish
2018/12/11 00:43:55 DEBUG : pacer: Rate limited, sleeping for 1.749245704s (1 consecutive low level retries)
2018/12/11 00:43:55 DEBUG : pacer: low level retry 1/10 (error googleapi: Error 403: Rate Limit Exceeded, rateLimitExceeded)
2018/12/11 00:43:55 DEBUG : pacer: Rate limited, sleeping for 2.148953723s (2 consecutive low level retries)
2018/12/11 00:43:55 DEBUG : pacer: low level retry 2/10 (error googleapi: Error 403: Rate Limit Exceeded, rateLimitExceeded)
[...]

PR #2947, opened 2019-02-02
Refine handling of unknown-sized uploads - #2832

Cnly:
This PR implements the changes disscussed in https://github.com/ncw/rclone/issues/2832#issuecomment-451862366. It:
- Makes clear in docs that `Put` and `Update` won't be called from outside a backend with a size < 0
- Makes `operations.Copy` call `operations.Rcat` if the input size is -1 which will buffer it on disk if necessary or use PutStream if available
- Ensures in the tests that `Fs.Put()` and `Object.Update()` don't panic when `src.Size() == -1`.

### DIFF
--- a/fstest/fstests/fstests.go
+++ b/fstest/fstests/fstests.go
@@ -1407,6 +1407,53 @@ func Run(t *testing.T, opt *Opt) {
 
 		})
 
+		// TestFsUploadUnknownSize ensures Fs.Put() and Object.Update() don't panic when
+		// src.Size() == -1
+		t.Run("FsUploadUnknownSize", func(t *testing.T) {
+			skipIfNotOk(t)
+
+			t.Run("FsPutUnknownSize", func(t *testing.T) {
+				defer func() {
+					assert.Nil(t, recover(), "Fs.Put() should not panic when src.Size() == -1")
+				}()
+
+				contents := fstest.RandomString(100)
+				in := bytes.NewBufferString(contents)
+
+				obji := object.NewStaticObjectInfo("unknown-size-put.txt", fstest.Time("2002-02-03T04:05:06.499999999Z"), -1, true, nil, nil)
+				obj, err := remote.Put(in, obji)
+				if err == nil {
+					require.NoError(t, obj.Remove(), "successfully uploaded unknown-sized file but failed to remove")
+				}
+				// if err != nil: it's okay as long as no panic
+			})
+
+			t.Run("FsUpdateUnknownSize", func(t *testing.T) {
+				unknownSizeUpdateFile := fstest.Item{
+					ModTime: fstest.Time("2002-02-03T04:05:06.499999999Z"),
+					Path:    "unknown-size-update.txt",
+				}
+
+				testPut(t, remote, &unknownSizeUpdateFile)
+
+				defer func() {
+					assert.Nil(t, recover(), "Object.Update() should not panic when src.Size() == -1")
+				}()
+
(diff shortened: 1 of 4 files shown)


fstest/fstests/fstests.go
```diff
@@ -1407,6 +1407,63 @@ func Run(t *testing.T, opt *Opt) {
 
 		})
 
+		// TestPutUnknownSize ensures Fs.Put() and Object.Update() don't panic when
+		// src.Size() == -1
+		t.Run("FsPutUnknownSize", func(t *testing.T) {
+
+			testPutUnknownSize := func() {
```
ncw: These two functions could be separate tests, or they could be subtests

    t.Run("Put", func(t*testing.T) {
    t.Run("Update", func(t*testing.T) {

fstest/fstests/fstests.go
```diff
@@ -1407,6 +1407,63 @@ func Run(t *testing.T, opt *Opt) {
...
+					Path:    "unknown-size-put.txt",
+				}
+
+				defer func() {
+					assert.Nil(t, recover(), "Fs.Put() should not panic when src.Size() == -1")
+				}()
+
+				contents := fstest.RandomString(100)
+				buf := bytes.NewBufferString(contents)
+				uploadHash := hash.NewMultiHasher()
```
ncw: You don't appear to use the uploadHash...

You could check it, however I think in this test just checking the object returned from Put has Size() 100 would be a sufficient test.

If you are doing that then you can lose the fstest.Item construction too.

fstest/fstests/fstests.go
```diff
@@ -1407,6 +1407,63 @@ func Run(t *testing.T, opt *Opt) {
...
+				in := io.TeeReader(buf, uploadHash)
+
+				unknownSizePutFile.Size = -1
+				obji := object.NewStaticObjectInfo(unknownSizePutFile.Path, unknownSizePutFile.ModTime, unknownSizePutFile.Size, true, nil, nil)
+				_, err := remote.Put(in, obji)
+				if err != nil {
+					// It's okay as long as no panic
+				}
```
ncw: After testing the size, removing the object would be polite just in case we add more tests after this one

fstest/fstests/fstests.go
```diff
@@ -1407,6 +1407,63 @@ func Run(t *testing.T, opt *Opt) {
...
+					Path:    "unknown-size-update.txt",
+				}
+
+				testPut(t, remote, &unknownSizeUpdateFile)
+
+				defer func() {
+					assert.Nil(t, recover(), "Object.Update() should not panic when src.Size() == -1")
+				}()
+
+				newContents := fstest.RandomString(200)
```
ncw: Similar comments apply here to above
- hashreader unused
- just check the size
- delete the object when one

ncw:
I've squashed your last two commits and merged - thank you very much :-)

I want to get this into the integration tester so we can see what blows up tomorrow morning at

https://pub.rclone.org/integration-tests/current/

> I removed the hash part. I think only checking the size is not much better than not checking at all. And since we're not calling `Put` and `Update` from the outside with `-1` sizes, it seems these tests are more of some guidelines than some mandatory standards (except for that `Put` and `Update` shouldn't panic)... That's why I intended not to have checks on the files.

OK that sounds fine :-)

ncw:
This test flushed out the mega backend as being non-conformant (it hung) fixed in a0016e00d1f720503052e207e44ca2b8fc1651de

PR #2961, opened 2019-02-09
b2: allow manual configuration of backblaze downloadUrl - fixes #2808

### LINKED ISSUE #2808: Request to have manual configuration of Backblaze downloadUrl
rclone v1.45
- os/arch: linux/amd64
- go version: go1.11.2

When moving files, rclone downloads file and re-uploads for backblaze, which results in bandwidth usage during downloads.

Recently cloudflare and backblaze made a bandwidth alliance, so if you could possible separate api url and ask for download url in backblaze configuration during setup, this problem can be solved by providing the download url for cloudflare enabled url rather than direct api.

trettet:

Backblaze introduced free egress for B2 Cloud Storage when files are [downloaded through the Cloudflare network](https://www.backblaze.com/blog/backblaze-and-cloudflare-partner-to-provide-free-data-transfer/). This requires using a custom url or endpoint for downloads instead of the `downloadUrl` provided by `b2_authorize_account`.

https://github.com/ncw/rclone/issues/2808

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -125,6 +125,14 @@ minimum size.`,
 			Help:     `Disable checksums for large (> upload cutoff) files`,
 			Default:  false,
 			Advanced: true,
+		}, {
+			Name: "download_url",
+			Help: `Custom endpoint for downloads.
+
+This is usually set to a Cloudflare CDN URL as Backblaze offers
+free egress for data downloaded through the Cloudflare network.
+Leave blank if you want to use the endpoint provided by Backblaze.`,
+			Advanced: true,
 		}},
 	})
 }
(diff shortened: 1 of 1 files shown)


ncw:

Presumably this will only work for public buckets?

I'll merge it now.

trettet:
It should normally have an HTTP Error 401 when using it for private buckets, but with rclone, I don't really know why it doesn't. It even works for snapshots.

ncw:
> So, yes, this would also work on private buckets without any further changes.

Interesting!

How do you get the Cloudflare endpoint you need?

trettet:
There is a guide posted by Backblaze [here](https://help.backblaze.com/hc/en-us/articles/217666928-Using-Backblaze-B2-with-the-Cloudflare-CDN). This guide assumes that your `downloadUrl` does not change.

**Tl;dr:** You need to set a CNAME record on Cloudflare for your `downloadUrl` and make sure to have the orange cloudflare icon for that record so traffic passes through their network.

PR #2962, opened 2019-02-09
pacer: modularize the package 

B4dM4n:

A while ago I tried to use the `lib/pacer` package for a personal project and found it hard to use without rclone config system. So I decided to improve reusability by removing the dependency on the `fs` package.

This change splits the sleep time calculation from the large `pacer.Pacer` struct into smaller parts and moves the rclone specific behavior like logging to the `fs` package. The new `Calculator` types also accept only their specific configuration options to make it clearer which pacer uses which settings.

All the default numbers and pacing algorithms should be the same as before. The existing tests only got modified to match the new API. I also added a `TestCallParallel` to check the parallel connection limit and modified the `TestGoogleDrivePacer` to test the new burst behavior.

The new `lib/errors` package provides a new `Walk` function to handle error chains similar to `fserrors.Cause` with added support for [xerrors.Unwrap](https://godoc.org/golang.org/x/xerrors#Unwrap) (if this PR is accepted, `fserrors` can be updated [too](https://github.com/ncw/rclone/compare/master...B4dM4n:fserrors-walk)).

No

#### Checklist

### DIFF
--- a/backend/amazonclouddrive/amazonclouddrive.go
+++ b/backend/amazonclouddrive/amazonclouddrive.go
@@ -155,7 +155,7 @@ type Fs struct {
 	noAuthClient *http.Client       // unauthenticated http client
 	root         string             // the path we are working on
 	dirCache     *dircache.DirCache // Map of directory path to directory id
-	pacer        *pacer.Pacer       // pacer for API calls
+	pacer        *fs.Pacer          // pacer for API calls
 	trueRootID   string             // ID of true root directory
 	tokenRenewer *oauthutil.Renew   // renew the token on expiry
 }
@@ -273,7 +273,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		root:         root,
 		opt:          *opt,
 		c:            c,
-		pacer:        pacer.New().SetMinSleep(minSleep).SetPacer(pacer.AmazonCloudDrivePacer),
+		pacer:        fs.NewPacer(pacer.NewAmazonCloudDrive(pacer.MinSleep(minSleep))),
 		noAuthClient: fshttp.NewClient(fs.Config),
 	}
 	f.features = (&fs.Features{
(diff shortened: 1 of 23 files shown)


ncw:
Actually looking at the CI results - there are some race conditions which need fixing....

B4dM4n:
I hope the tests are now stable on the CI server.

The race was easy to fix. I put the racing variable inside the wait lock

Fixing the failing `TestGoogleDrivePacer` was trickier. It looks like millisecond precision sleeps and time measurements are very fragile on the CI server. I didn't want to increase the test duration by using hundreds of milliseconds or seconds, so I decided to skip the exact timing test and reduced it to counting the sleep calls.

ncw:
I've just realised why the CI keeps failing.  It is because the GITHUB_USER and GITHUB_PASS needed to avoid the rate limiting aren't set for pull requests :-(

ncw:

ncw:
One of the pacer tests seems to be intermittent on macOS

- [1](https://travis-ci.org/ncw/rclone/jobs/498369957)
- [2](https://travis-ci.org/ncw/rclone/jobs/496642891)
- [3](https://travis-ci.org/ncw/rclone/jobs/496026500)
- [4](https://travis-ci.org/ncw/rclone/jobs/494762524)

```
--- FAIL: TestGoogleDrivePacer (0.02s)
    pacer_test.go:245: 
        	Error Trace:	pacer_test.go:245
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestGoogleDrivePacer
        	Messages:   	test: {calls:12 want:2}, got: 1
FAIL
```

Any ideas @B4dM4n ?

ncw:
@B4dM4n I fixed the macOS pacer test in f6eb5c69838e53c9d0c4041bab3283a023e3e15a - I realised  that the test didn't need to sleep to perform its function which removes all the indeterminacy.

Issue #87, opened 2015-07-27
Incremental sync


jmsanzg:
When rclone is installed on a system with a bad connectivity where connections go down from time to time, could be perfect if rclone could do small incremental syncs of big files like rsync does so the file could be uploaded/downloaded in small chuncks.

ncw:
rclone can't do an incremental sync like rsync - that just isn't possible with any cloud storage system I know.

However some cloud storage systems (drive and s3 for definite) can resume uploads which I think would satisfy your needs.

hitl0r:
+1 for this one. I am uploading huge files which would take +50 hours to upload to amazon s3  - but my line gets disconnected by the ISP every 24hours. (standard procedure)
Thus the rclone upload never finishes and my backup is stuck at the same position forever.

noahmehl:
@ncw for instance: https://developers.google.com/drive/v3/web/manage-uploads#resumable

Is this a possibility in rclone?

DurvalMenezes:
This would definitely be a great feature for rclone.

DurvalMenezes:
It would be a great feature for rclone, and would make it much more similar to a "rsync for the cloud". Hope it can be done to some "unlimited for a fixed price" cloud, which is something I would really like to use here.

felixbuenemann:
Amazon Cloud Drive definitely supports resuming uploads, it is possible to pause all uploads in the official Amazon Drive Mac app and resume uploading at a later time and upload will resume at eg. 50%.

I think as a first step it would be great to support resuming uploads on network errors, so rclone doesn't need to re-upload the complete file.

A second step could be some sort of state file written at Ctrl+C that could be used to resume uploads. For example for S3 this needs to contain the Bucket, Key and UploadId.

felixbuenemann:
I have looked into the ACD REST API docs, but it does not document how resuming works.

However the Mac .NET client also uses the REST API and has very verbose log, so I can somewhat reconstruct how it works:

```
# Generate a unique localId, eg uuidv4, the Mac client uses a 32 byte hex string
# Let's use "xyz" in this example.

# Start upload and pass the localId (or cancellation token)
POST /nodes?suppress=deduplication&localId=xyz
# Cancel the POST request, either manually or due to a network error

# Get the state of the upload
GET /resume?localId=xyz
# Returns either 404 NotFound, 204 NoContent or 200 OK
# on 204 the client retries with some delay until it gets 200
# response JSON contains:
[...]

PR #2996, opened 2019-02-23
[WIP] atexit: allow functions to be unregistered

Cnly:
This PR refines the `atexit` library mainly to provide the possibility for backends to gracefully cancel transfers when rclone is interrupted by the user.

Forum post that initiated this PR: https://forum.rclone.org/t/onedrive-tmp-issue/8263

### DIFF
--- a/lib/atexit/atexit.go
+++ b/lib/atexit/atexit.go
@@ -13,23 +13,32 @@ import (
 )
 
 var (
-	fns          []func()
+	fns          = make(map[FnHandle]bool)
+	fnsMutex     sync.Mutex
 	exitChan     chan os.Signal
 	exitOnce     sync.Once
 	registerOnce sync.Once
 )
 
-// Register a function to be called on exit
-func Register(fn func()) {
-	fns = append(fns, fn)
+// FnHandle is the type of the handle returned by function `Register`
+// that can be used to unregister an at-exit function
+type FnHandle *func()
+
+// Register a function to be called on exit.
+// Returns a handle which can be used to unregister the function with `Unregister`.
+func Register(fn func()) FnHandle {
+	fnsMutex.Lock()
+	fns[&fn] = true
+	fnsMutex.Unlock()
+
 	// Run AtExit handlers on SIGINT or SIGTERM so everything gets
 	// tidied up properly
 	registerOnce.Do(func() {
 		exitChan = make(chan os.Signal, 1)
 		signal.Notify(exitChan, os.Interrupt) // syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT
 		go func() {
-			sig, closed := <-exitChan
-			if closed || sig == nil {
+			sig := <-exitChan
+			if sig == nil {
 				return
 			}
 			fs.Infof(nil, "Signal received: %s", sig)
@@ -38,6 +47,15 @@ func Register(fn func()) {
 			os.Exit(0)
 		}()
 	})
+
+	return &fn
+}
+
+// Unregister a function using the handle returned by `Register`
+func Unregister(handle FnHandle) {
+	fnsMutex.Lock()
+	defer fnsMutex.Unlock()
+	delete(fns, handle)
 }
 
 // IgnoreSignals disables the signal handler and prevents Run from beeing executed automatically
@@ -53,8 +71,10 @@ func IgnoreSignals() {
 // Run all the at exit functions if they haven't been run already
 func Run() {
 	exitOnce.Do(func() {
-		for _, fn := range fns {
-			fn()
+		fnsMutex.Lock()
+		defer fnsMutex.Unlock()
+		for fnHandle := range fns {
+			(*fnHandle)()
 		}
 	})
 }


lib/atexit/atexit.go
```diff
@@ -53,8 +71,10 @@ func IgnoreSignals() {
 // Run all the at exit functions if they haven't been run already
 func Run() {
 	exitOnce.Do(func() {
-		for _, fn := range fns {
-			fn()
+		fnsMutex.Lock()
+		defer fnsMutex.Unlock()
+		for fnHandle, _ := range fns {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for fnHandle := range ...` (from `golint`)

Cnly:
@ncw I think we should add a new function in `atexit`, say `atexit.Atomic(func ()) err`. It guarantees the specified function is either finished or aborted when rclone is interrupted. I thought of this when implementing graceful cancel for OneDrive:

Before doing multipart upload, rclone needs to create an upload session:

```go
session, err := o.createUploadSession(modTime)
	if err != nil {
		return nil, err
	}
	uploadURL := session.UploadURL
```

But cancelling an upload session requires the `uploadURL` returned by `createUploadSession`. Therefore, if rclone is interrupted during the upload session is being created, there will be no way to gracefully cancel the process and a `~tmp` file will still be created. With `atexit.Atomic`, `createUploadSession` can be guaranteed to provide an `uploadURL` or fail with an error and we should have no problem anymore.

What's your thoughts on this?

ncw:
> @ncw I think we should add a new function in `atexit`, say `atexit.Atomic(func ()) err`. It guarantees the specified function is either finished or aborted when rclone is interrupted. I thought of this when implementing graceful cancel for OneDrive:

I'm not sure I understand how that would work.  I can see how you might run the `func()` atomically, but I'm struggling to work out what happens if it is aborted.  What action should it take at that point?

> Before doing multipart upload, rclone needs to create an upload session:
> 
> ```go
> session, err := o.createUploadSession(modTime)
> 	if err != nil {
> 		return nil, err
> 	}
> 	uploadURL := session.UploadURL
> ```
> 
> But cancelling an upload session requires the `uploadURL` returned by `createUploadSession`. Therefore, if rclone is interrupted during the upload session is being created

Do you mean we've sent the request to onedrive, but the user CTRL-C's rclone before we read the response and setup the atexit?  I think getting rid of race opportunities like that completely will be very hard!

>, there will be no way to gracefully cancel the process and a `~tmp` file will still be created. With `atexit.Atomic`, `createUploadSession` can be guaranteed to provide an `uploadURL` or fail with an error and we should have no problem anymore.

Do you have a sketch of the implementation?  That might make it clearer to me what you are thinking of.

ncw:
I'm going to merge this now as I've realised it fixes #2997 too with this change:

image

Cnly:
Oh, no wonder I didn't figure out what was this `if closed` thing for... The registered functions won't run on my machine with it. :D
I'll open another thread to discuss `Atomic`.

ncw:
> Oh, no wonder I didn't figure out what was this `if closed` thing for... The registered functions won't run on my machine with it. :D

They were only running half the time on mine...

Probably indicates that we need some tests for lib/atomic which might be a little tricky to write...   The way it is normally done is running a new copy of the test binary with an environment variable, then an init function which reads the variable and runs something which we can test the signal arrived and the library cleared up properly.

> I'll open another thread to discuss `Atomic`.

Great

Cnly:
Recently I thought about `Atomic` again and it seems that we don't need such a function. Channels are doing a pretty good job I think. I have implemented the graceful exit feature for OneDrive at https://github.com/ncw/rclone/commit/e87b4e5385903588734905addc99763e914ee6ff, and, do you have a list of backends that would leave trash behind without this feature? Maybe we just need to worry about them.

ncw:
> Recently I thought about `Atomic` again and it seems that we don't need such a function. Channels are doing a pretty good job I think. I have implemented the graceful exit feature for OneDrive at [cae787a](https://github.com/ncw/rclone/commit/cae787ad91f1a7aca23b27e630feb99ed2efd7e1),

I'm unsure why you need the channel in that code, why not do it something like this with a closure? Registering the handler before you call the `createUploadSession` doesn't buy you anything does it?

```go
	// Create upload session
	fs.Debugf(o, "Starting multipart upload")
	session, err := o.createUploadSession(modTime)
	if err != nil {
		return nil, err
	}
	uploadURL := session.UploadURL
        cancelFuncHandle := atexit.Register(func(){
		fs.Debugf(o, "Cancelling multipart upload")
		cancelErr := o.cancelUploadSession(uploadURL)
		if cancelErr != nil {
			fs.Logf(o, "Failed to cancel multipart upload: %v", cancelErr)
		}
        })
```

> and, do you have a list of backends that would leave trash behind without this feature? Maybe we just need to worry about them.

I think any backend which supports multipart uploads...

I went through recently attempting to put integration tests into all the backends for chunked uploads so this list is probably reasonably accurate

```
$ find backend -name "*_test.go" | xargs grep -l Chunk
backend/drive/drive_test.go
backend/s3/s3_test.go
backend/dropbox/dropbox_test.go
backend/swift/swift_test.go
backend/azureblob/azureblob_test.go
backend/qingstor/qingstor_test.go
backend/onedrive/onedrive_test.go
backend/b2/b2_test.go
```

Cnly:
> I'm unsure why you need the channel in that code, why not do it something like this with a closure? 

I used the channel because I wanted no `~tmp` files to be left behind in one case where the user interrupts when rclone is waiting for the response of a `createUploadSession` request. If the user is syncing a lot of small files, considering that the transfers are parallel and every file requires a `createUploadSession`, I think it's quite possible that there will be a bunch of `~tmp` files unhandled when rclone is interrupted, if we're not using some mechanism like that channel.

Maybe I should've also make it clearer that once the `createUploadSession` is sent, a `~tmp` file is already created.

> I think any backend which supports multipart uploads...

Maybe we can just deal with the backends that will have side effects, like OneDrive which has `~tmp` files?

ncw:
> > I'm unsure why you need the channel in that code, why not do it something like this with a closure?
> 
> I used the channel because I wanted no `~tmp` files to be left behind in one case where the user interrupts when rclone is waiting for the response of a `createUploadSession` request.

I think I'm missing something here :-)

If the user interrupts the `createUploadSession` request , we won't have sent the `uploadURL`  to the channel so we won't be able to cancel it?

So yes there is a window of opportunity for ~tmp files, but I don't think having the channel narrows it.

> If the user is syncing a lot of small files, considering that the transfers are parallel and every file requires a `createUploadSession`, I think it's quite possible that there will be a bunch of `~tmp` files unhandled when rclone is interrupted, if we're not using some mechanism like that channel.
> 
> Maybe I should've also make it clearer that once the `createUploadSession` is sent, a `~tmp` file is already created.
> 
> > I think any backend which supports multipart uploads...
> 
> Maybe we can just deal with the backends that will have side effects, like OneDrive which has `~tmp` files?

The side effects may not be visible on some remotes, but normally you pay for unfinished multipart uploads on all of them.

Cnly:
> So yes there is a window of opportunity for ~tmp files, but I don't think having the channel narrows it.

You're right; we need `uploadURL` to cancel a session. Why don't you think the channel cannot narrow it? If the channel is there, when the user interrupts, this line

```go
		uploadURL, ok := <-uploadURLChan
```

blocks rclone from completing the `gracefulCancel` functions registered by the OneDrive backend, and thus blocks rclone from calling `os.Exit(0)`. This way, rclone won't exit until all `createUploadSession` requests are complete and all the upload sessions are cancelled.

> The side effects may not be visible on some remotes, but normally you pay for unfinished multipart uploads on all of them.

Okay, I think this makes it necessary to implement this feature on more backends...

ncw:
> > So yes there is a window of opportunity for ~tmp files, but I don't think having the channel narrows it.
> 
> You're right; we need `uploadURL` to cancel a session. Why don't you think the channel cannot narrow it? If the channel is there, when the user interrupts, this line
> 
> ```go
> 		uploadURL, ok := <-uploadURLChan
> ```
> 
> blocks rclone from completing the `gracefulCancel` functions registered by the OneDrive backend, and thus blocks rclone from calling `os.Exit(0)`. This way, rclone won't exit until all `createUploadSession` requests are complete and all the upload sessions are cancelled.

Oh, I see! That's what I was missing, sorry to be slow on the uptake. The channel receive probably needs a comment that it will block the atexit process.

Yes that narrows the window to zero which is great.

> > The side effects may not be visible on some remotes, but normally you pay for unfinished multipart uploads on all of them.
> 
> Okay, I think this makes it necessary to implement this feature on more backends...

From memory most of the systems will expire half completed uploads after a while and most also have API for listing them.

The atexit hook could be used for writing a resume file too couldn't it.

Cnly:
> Oh, I see! That's what I was missing, sorry to be slow on the uptake. The channel receive probably needs a comment that it will block the atexit process.

Ah, I think I can understand. When I was writing the code I also stuck there for a while thinking about why this was able to do the job :P I should have considered adding comments in situations like this.

> The atexit hook could be used for writing a resume file too couldn't it.

Surely, which may even be better than the current solution. I think we can make this into rclone's `lib`. But we may also need to allow the user to choose from saving a resume file or cancelling the operation.

There seems to be pretty much new stuff here in this closed PR now. Maybe we should move elsewhere for further discussion?

ncw:
> There seems to be pretty much new stuff here in this closed PR now. Maybe we should move elsewhere for further discussion?

Do you want to make another issue about cancelling or resuming multipart transfers?

Cnly:
I think I'll merge the graceful cancel feature for OneDrive first, then perhaps I can also work on other backends. But for resuming, I saw issues like https://github.com/ncw/rclone/issues/87 and there are also discussions related to some more general metadata storage. I'm not so sure how we should proceed on that.

ncw:
> I think I'll merge the graceful cancel feature for OneDrive first, then perhaps I can also work on other backends.

That sounds great - very interested to see how it turns out

> But for resuming, I saw issues like #87 and there are also discussions related to some more general metadata storage. I'm not so sure how we should proceed on that.

Resuming will require some more infrastructure in the backends, I think - it is quite a bit more complicated.

PR #3003, opened 2019-02-26
Koofr backend

jkmcnk:

Implements a backend for the Koofr cloud storage service.

No

#### Checklist


backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+
+func (f *koofrfs) Hashes() hash.Set {
+	return hash.Set(hash.MD5)
+}
+
+func (f *koofrfs) fullPath(part string) string {
+	return path.Join("/", f.root, part)
+}
+
+func NewFs(name, root string, m configmap.Mapper) (ff fs.Fs, err error) {
```
golangcibot: exported function `NewFs` should have comment or be unexported (from `golint`)

backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+)
+
+func init() {
+	fs.Register(&fs.RegInfo{
+		Name:        "koofr",
+		Description: "Koofr",
+		NewFs:       NewFs,
+		Options: []fs.Option{
+			{
+				Name:     "baseurl",
```
ncw: This should be called `endpoint` for consistency with the other backends

backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+				Name:       "password",
+				Help:       "Your Koofr password for rclone (generate one at https://app.koofr.net/app/admin/preferences/password)",
+				IsPassword: true,
+				Required:   true,
+			},
+		},
+	})
+}
+
+type options struct {
```
ncw: I'd like to keep the backends consistent so this should be called `Options` as it is in the other backends.

backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+	BaseURL  string `config:"baseurl"`
+	MountID  string `config:"mountid"`
+	User     string `config:"user"`
+	Password string `config:"password"`
+}
+
+type koofrfs struct {
```
ncw: Likewise this should be called `Fs`

backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+	name     string
+	mountID  string
+	root     string
+	opt      options
+	features *fs.Features
+	client   *koofrclient.KoofrClient
+}
+
+type object struct {
```
ncw: And this called `Object`

Those 3 things are public so the backends can be used without the rest of the rclone framework.

backend/koofr/koofr.go
```diff
@@ -0,0 +1,461 @@
...
+func (o *object) fullPath() string {
+	return o.fs.fullPath(o.remote)
+}
+
+func (o *object) Storable() bool {
+	return true
+}
+
+func (o *object) SetModTime(mtime time.Time) error {
+	return nil
```
ncw: It is a shame koofr can't set the modified time of a file.

Is there metadata that rclone could use to save the modified time?

ncw:
I had a quick go with the backend and managed to successfully upload and download 1000 files :-)

It gave me these rather unfriendly looking errors which I'm not sure what they mean - a retry cleared them up.

```
2019/02/26 15:05:49 ERROR : sewewot7c/mefipuz7ka/dutamor1: Failed to copy: Invalid response status! Got 400, expected [200 201]; headers: map[Cache-Control:[no-cache, no-store, must-revalidate] Content-Length:[0] Date:[Tue, 26 Feb 2019 15:05:49 GMT] Expires:[0] Pragma:[no-cache] Set-Cookie:[PLAY_SESSION=; Expires=Mon, 25 Feb 2019 15:05:50 GMT; Path=/; HTTPOnly] X-Request-Id:[a2f2d2ed-0059-43fb-925a-421a9156f185] X-User-Id:[64ee1c28-c9ac-4f50-97d0-b24cfe05482f]], content: 
```

The backend doesn't implement the pacer so doesn't do low-level retries.

It also uncovered a bug as the errors that httpclient return are not comparable as they don't contain a slice.  This needs this fix

```patch
diff --git a/fs/fserrors/error.go b/fs/fserrors/error.go
index a2ede7416..39fd9fb64 100644
--- a/fs/fserrors/error.go
+++ b/fs/fserrors/error.go
@@ -194,7 +194,7 @@ func Cause(cause error) (retriable bool, err error) {
                        // this case.
                        err = prev
                }
-               if err == prev {
+               if reflect.DeepEqual(err, prev) {
                        // Unpack any struct or *struct with a field
                        // of name Err which satisfies the error
                        // interface.  This includes *url.Error,
@@ -215,7 +215,7 @@ func Cause(cause error) (retriable bool, err error) {
                                }
                        }
                }
-               if err == prev {
+               if reflect.DeepEqual(err, prev) {
                        break
                }
        }
```

To fix.

I note that Koofr supports Webdav - does rclone work with the webdav interface? What advantage does using the API give? MD5 support seems like one of them.

I can't find docs for the API anywhere - can you link them?

backend/koofr/koofr.go
jkmcnk: lol. I have only unexported the types after the golangci linter complained extensively about undocumented exported stuff, as I hate the noise of warnings during builds. :D

you really should add something to the CONTRIBUTING.md about ignoring the damn linter. ;)

kudos for the PR submission policy, though.

backend/koofr/koofr.go
jkmcnk: arbitrary metadata is only available on files stored on our servers, but since we also allow acces to some third-party cloud storage providers as well as files from users machines via the same API, this would not be a universal solution. we'll look into what can be done about this on our side.

jkmcnk:
wrt to the integration tests, I've ran

test_all -backends koofr

and it passes with flying colours; is there more to the integration tests? (but, more importantly, we pushed our local windows test build to our CEO who used to use rclone via our webdav interface, and I still have the job, so I suppose it's mostly working)

the advantage over the webdav interface is speed. the webdav interface is quite limited wrt to concurrency as it strictly adheres to the webdav locking spec.

the unfriendly looking errors: any idea which part of rclone do they come from? the Put or Copy methods?

API docs and some examples can be found at http://k00.fr/cug7u

ncw:
> wrt to the integration tests, I've ran
> 
> test_all -backends koofr
> 
> and it passes with flying colours; is there more to the integration tests?

No that is it - and good news :-)

> (but, more importantly, we pushed our local windows test build to our CEO who used to use rclone via our webdav interface, and I still have the job, so I suppose it's mostly working)

Excellent :-)

> the advantage over the webdav interface is speed. the webdav interface is quite limited wrt to concurrency as it strictly adheres to the webdav locking spec.

It appears to support MD5 hashes also which isn't a standard part of the webdav spec IIRC.
 
> the unfriendly looking errors: any idea which part of rclone do they come from? the Put or Copy methods?

That was from `Put` - I just tried to upload a directory of 1000 smallish files with the default concurrency.  Every 1 in 20 or so files gave that error, but uploaded fine on a retry.

> API docs and some examples can be found at http://k00.fr/cug7u

Cool - thanks.  I note that koofr could support some more optional methods
- `PublicLink` to provide a public link to a file or directory
- `About` to provide used/free/quota space (which then makes `df` on `rclone mount`  work nicely)
- `Purge` - maybe DELETE folder deletes a folder and its contents?  If so this would implement Purge

jkmcnk:
ok, the 400 errors are a legit response of our API on attempting to create a folder which already exists. it should have been a 409, but it's a pain to change the API with a ton of clients out there. so, I have fixed this in the our rclone backend.

fwiw, attempts to create existing dirs are due to concurrent Put()s with each one try to create the complete directory hierarchy which ends up in races between a fileinfo in one thread, reporting a nonexisting folder, then a mkdir in another thread, followed by yeat another mkdir in the first one, with this last one erroring out.

I'll also do the reverse type renaming, and see about the implementation of the 3 optional methods. Closing this PR, will submit another one tomorrow.

ncw:
> ok, the 400 errors are a legit response of our API on attempting to create a folder which already exists. it should have been a 409, but it's a pain to change the API with a ton of clients out there. so, I have fixed this in the our rclone backend.

That makes sense.
>
> fwiw, attempts to create existing dirs are due to concurrent Put()s with each one try to create the complete directory hierarchy which ends up in races between a fileinfo in one thread, reporting a nonexisting folder, then a mkdir in another thread, followed by yeat another mkdir in the first one, with this last one erroring out.
> 
> I'll also do the reverse type renaming, and see about the implementation of the 3 optional methods. Closing this PR, will submit another one tomorrow.

Great :-)

PR #3036, opened 2019-03-08
support --no-check-certificate on copyurl command

breunigs:

rclone will currently always fail when copying a file via `rclone copyurl <url> <remote>`. `--no-check-certificate` exists, but is only used for http fs as far as I can tell. This patch extends the usage to `copyurl` as well, so that `rclone copyurl --no-check-certificate <url> <remote>` will allow to copy from broken servers. 

Not that I am aware.

#### Checklist

### DIFF
--- a/fs/fshttp/http.go
+++ b/fs/fshttp/http.go
@@ -28,7 +28,7 @@ const (
 
 var (
 	transport    http.RoundTripper
-	noTransport  sync.Once
+	noTransport  = new(sync.Once)
 	tpsBucket    *rate.Limiter // for limiting number of http transactions per second
 	cookieJar, _ = cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
 )
@@ -121,9 +121,15 @@ func dialContextTimeout(ctx context.Context, network, address string, ci *fs.Con
 	return newTimeoutConn(c, ci.Timeout)
 }
 
+// ResetTransport resets the existing transport, allowing it to take new settings.
+// Should only be used for testing.
+func ResetTransport() {
+	noTransport = new(sync.Once)
+}
+
 // NewTransport returns an http.RoundTripper with the correct timeouts
 func NewTransport(ci *fs.ConfigInfo) http.RoundTripper {
-	noTransport.Do(func() {
+	(*noTransport).Do(func() {
 		// Start with a sensible set of defaults then override.
 		// This also means we get new stuff when it gets added to go
 		t := new(http.Transport)
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net/http"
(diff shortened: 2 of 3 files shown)


fs/operations/operations.go
```diff
@@ -1399,7 +1400,12 @@ func RcatSize(fdst fs.Fs, dstFileName string, in io.ReadCloser, size int64, modT
 
 // CopyURL copies the data from the url to (fdst, dstFileName)
 func CopyURL(fdst fs.Fs, dstFileName string, url string) (dst fs.Object, err error) {
-	resp, err := http.Get(url)
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: fs.Config.InsecureSkipVerify},
+	}
+	client := &http.Client{Transport: tr}
```
ncw: I think using the rclone client might be best here

```go
    client := fshttp.NewClient(fs.Config)
```

That will make --no-check-certificate work.  It will also make a host of other things work like `--timeout`, `--dump headers`, etc so it is probably the best solution

ncw:
@breunigs did you see my comments on the above?

fs/fshttp/http.go
```diff
@@ -121,6 +121,11 @@ func dialContextTimeout(ctx context.Context, network, address string, ci *fs.Con
 	return newTimeoutConn(c, ci.Timeout)
 }
 
+func ResetTransport() {
+	var newOnce sync.Once
+	noTransport = newOnce
```
golangcibot: assignment copies lock value to noTransport: sync.Once contains sync.Mutex (from `govet`)

fs/fshttp/http.go
```diff
@@ -121,6 +121,11 @@ func dialContextTimeout(ctx context.Context, network, address string, ci *fs.Con
 	return newTimeoutConn(c, ci.Timeout)
 }
 
+func ResetTransport() {
```
golangcibot: exported function `ResetTransport` should have comment or be unexported (from `golint`)

fs/fshttp/http.go
```diff
@@ -121,9 +121,15 @@ func dialContextTimeout(ctx context.Context, network, address string, ci *fs.Con
 	return newTimeoutConn(c, ci.Timeout)
 }
 
+// NewTransport resets the existing transport, allowing it to take new settings.
```
golangcibot: comment on exported function `ResetTransport` should be of the form `ResetTransport ...` (from `golint`)

breunigs:
@ncw I did, but I was away from any keyboard. Your suggestion is very reasonable, but as you can see I was struggling to keep the test in `operations_test` instead of making it a very small unit test in `httpfs_test`. I can move it to unexport the `ResetTransport` function, if that's a better trade off.

ncw:
I think the `ResetTransport` looks like a useful function for testing so i'm going to merge this as-is.

Thank you very much and apologies for the delay - I've got a bit of a backlog.

PR #3045, opened 2019-03-12
dlna: add connection manager service description

sretlawd:
The UPnP MediaServer spec says that the ConnectionManager service is
required, and adding it was enough to get dlna support working on my
other TV (LG webOS 2.2.1).

Improve device compatibility & spec compliance by adding an additional service descriptor for ConnectionManager v1.  Through lots of trial and error, I found this to be the missing piece to
get dlna working on my LG webOS 2.2.1 TV.

#### Checklist


sretlawd:
Ah - I should add this doesn't actually implement the ConnectionManager interface, it just advertises it, but that's all that was necessary to get it working on my TV.  I can look at what a minimal actual implementation would be, just to avoid false advertising.

ncw:
> The UPnP MediaServer spec says that the ConnectionManager service is
> required, and adding it was enough to get dlna support working on my
> other TV (LG webOS 2.2.1).
> 
> Improve device compatibility & spec compliance by adding an additional service descriptor for ConnectionManager v1. Through lots of trial and error, I found this to be the missing piece to
> get dlna working on my LG webOS 2.2.1 TV.

Nice one :-)

> Ah - I should add this doesn't actually implement the ConnectionManager interface, it just advertises it, but that's all that was necessary to get it working on my TV. I can look at what a minimal actual implementation would be, just to avoid false advertising.

What does the `ConnectionManager` interface do?  Is it likely to cause problems advertising an interface which isn't there?

sretlawd:
> What does the `ConnectionManager` interface do? Is it likely to cause problems advertising an interface which isn't there?

I don't think it can cause problems not to implement it, but there's probably some device out there that won't work until it is implemented.  Looking at a couple other open source implementations:  Emby Server implements only the getProtocolInfo method, which basically returns which mime types can be played over http.  Minidlnad implements that one plus returns dummy data for a couple other methods.  Doesn't look important at all, it's just typical overengineered design by committee stuff, like all of UPnP..

tabnk:
This PR partially fixed issue #3050. Oculus Go can now view and play media files from rclone (DLNA) via Google Drive. 

However, LG TV (LG LM6200 model 2012) TV crash (restart itself) when rclone (DLNA) was selected.

beredim:
@tabnk
LG LM6200 would be a pre-webos era model, correct?
Does it have latest f/w installed, and how does it behave with other DLNA servers?

tabnk:
@beredim yes, it is pre-webos model.  Latest firmware installed. 

Serviio, Universal Media Server etc work fine my LG TV.

ncw:
OK I'll merge this as it is definitely an improvement - thank you :-)

PR #3058, opened 2019-03-18
webdav: new timeformat for SFCC - fixes #3048

### LINKED ISSUE #3048: WebDAV: Another timeFormat
<!--

Welcome :-) We understand you are having a problem with rclone; we want to help you with that!

If you've just got a question or aren't sure if you've found a bug then please use the rclone forum:

    https://forum.rclone.org/

instead of filing an issue for a quick response.

If you think you might have found a bug, please can you try to replicate it with the latest beta?

    https://beta.rclone.org/
    
If you can still replicate it with the latest beta, then please fill in the info below which makes our lives much easier.  A log with -vv will make our day :-)

Thank you

The Rclone Developers

-->

Trying to copy from webdav - in this case, Salesforce Commerce Cloud log webdav, date is parsed incorrectly.

Looks like it's a timeFormat timestamp with additional hour offset:

"Wed, 13 Mar 2019 15:19:49 GMT-00:00"



rclone v1.46                                                                                                                                                                                                                                                                    os/arch: linux/amd64                                                                                                                                                                                                                                                          go version: go1.11.5


Ubuntu Linux 18.04

WebDAV

rclone copy remote1: remote2:remotepath/ -vv

> Couldn't parse Date: from server remote1: "Wed, 13 Mar 2019 15:19:49 GMT-00:00": parsing time "Wed, 13 Mar 2019 15:19:49 GMT-00:00" as "Mon Jan _2 15:04:05 2006": cannot parse ", 13 Mar 2019 15:19:49 GMT-00:00" as " "

danielloader:

Add additional timezone format to be be able to parse the webdav modification times that Salesforce Commerce Cloud uses

https://github.com/ncw/rclone/issues/3048

#### Checklist

- [*] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [*] I have added tests for all changes in this PR if appropriate.
- [*] I have added documentation for the changes if appropriate.
- [*] All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).

### DIFF
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -19,6 +19,9 @@ const (
 	// The same as time.RFC1123 with optional leading zeros on the date
 	// see https://github.com/ncw/rclone/issues/2574
 	noZerosRFC1123 = "Mon, _2 Jan 2006 15:04:05 MST"
+	// The same as time.RFC1123 with suffixed timezone offset
+	// see https://github.com/ncw/rclone/issues/3048
+	TZOffsetRFC1123 = "Wed, 13 Mar 2019 15:19:49 GMT-00:00"
 )
 
 // Multistatus contains responses returned from an HTTP 207 return code
@@ -172,6 +175,7 @@ var timeFormats = []string{
 	time.UnixDate,  // Wed May 17 15:31:58 UTC 2017 (as used in an internal server)
 	noZerosRFC1123, // Fri, 7 Sep 2018 08:49:58 GMT (as used by server in #2574)
 	time.RFC3339,   // Wed, 31 Oct 2018 13:57:11 CET (as used by komfortcloud.de)
+	TZOffsetRFC1123,// Wed, 13 Mar 2019 15:19:49 GMT-00:00 (as used by Salesforce Commerce Cloud)
 }
 
 var oneTimeError sync.Once
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -21,7 +21,7 @@ const (
 	noZerosRFC1123 = "Mon, _2 Jan 2006 15:04:05 MST"
 	// The same as time.RFC1123 with suffixed timezone offset
 	// see https://github.com/ncw/rclone/issues/3048
-	TZOffsetRFC1123 = "Wed, 13 Mar 2019 15:19:49 GMT-00:00"
+	TZOffsetRFC1123 = "Mon, 02 Jan 2006 15:04:05 MST-07:00"
 )
 
 // Multistatus contains responses returned from an HTTP 207 return code
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -175,7 +175,7 @@ var timeFormats = []string{
 	time.UnixDate,  // Wed May 17 15:31:58 UTC 2017 (as used in an internal server)
 	noZerosRFC1123, // Fri, 7 Sep 2018 08:49:58 GMT (as used by server in #2574)
 	time.RFC3339,   // Wed, 31 Oct 2018 13:57:11 CET (as used by komfortcloud.de)
-	TZOffsetRFC1123,// Wed, 13 Mar 2019 15:19:49 GMT-00:00 (as used by Salesforce Commerce Cloud)
+	tzOffsetRFC1123,// Wed, 13 Mar 2019 15:19:49 GMT-00:00 (as used by Salesforce Commerce Cloud)
 }
 
 var oneTimeError sync.Once
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -170,12 +170,12 @@ func (t *Time) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
 
 // Possible time formats to parse the time with
 var timeFormats = []string{
-	timeFormat,     // Wed, 27 Sep 2017 14:28:34 GMT (as per RFC)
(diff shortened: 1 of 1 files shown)


backend/webdav/api/types.go
```diff
@@ -19,6 +19,9 @@ const (
 	// The same as time.RFC1123 with optional leading zeros on the date
 	// see https://github.com/ncw/rclone/issues/2574
 	noZerosRFC1123 = "Mon, _2 Jan 2006 15:04:05 MST"
+	// The same as time.RFC1123 with suffixed timezone offset
```
golangcibot: comment on exported const `TZOffsetRFC1123` should be of the form `TZOffsetRFC1123 ...` (from `golint`)

backend/webdav/api/types.go
```diff
@@ -19,6 +19,9 @@ const (
...
+	// see https://github.com/ncw/rclone/issues/3048
+	TZOffsetRFC1123 = "Wed, 13 Mar 2019 15:19:49 GMT-00:00"
```
ncw: That doesn't look like a [go magic time value](https://golang.org/pkg/time/#Parse)?  I think you need something like `"Mon, 02 Jan 2006 15:04:05 MST-07:00"`

danielloader: I did wonder how it read a human readable time stamp, makes sense now!
Changing it.

danielloader:
Seems to fail the automatic checks and build, related to the comments added?

ncw:
> Seems to fail the automatic checks and build, related to the comments added?

That is the linter complaining about public symbols...  I suggest you rename your constant to `tzOffsetRFC1123` which is private and it will stop complaining!

danielloader:
Made the change requested for the variable name.

ncw:
To get the CI to pass, you'll need to run `go fmt backend/webdav/api/types.go`

Is `rclone lsl -vv --max-depth 1 webdav:` showing sensible time values now?

danielloader:
I've built the binary and tried doing a lsl -vv and I still get the same issue with parsing, not sure what's wrong at this point:
```

$ ./rclone lsl webdav:log_archive -vv                                                                                                                                                                                                  2019/03/19 11:24:50 DEBUG : rclone: Version "v1.46-DEV" starting with parameters ["./rclone" "lsl" "staging-iceland-co-uk:log_archive" "-vv"]                                                                                                                                   2019/03/19 11:24:50 DEBUG : Using config file from "/home/logs/.config/rclone/rclone.conf"                                                                                                                                                                                  2019/03/19 11:24:51 DEBUG : Couldn't parse Date: from server example.com: "Tue, 19 Mar 2019 11:24:50 GMT-00:00": parsing time "Tue, 19 Mar 2019 11:24:50 GMT-00:00" as "Mon Jan _2 15:04:05 2006": cannot parse ", 19 Mar 2019 11:24:50 GMT-00:00" as " "             2019/03/19 11:24:52 DEBUG : Couldn't parse Date: from server example.com: "Tue, 19 Mar 2019 11:24:52 GMT-00:00": parsing time "Tue, 19 Mar 2019 11:24:52 GMT-00:00" as "Mon Jan _2 15:04:05 2006": cannot parse ", 19 Mar 2019 11:24:52 GMT-00:00" as " "
--



```

ncw:
It looks like you can't parse a timezone twice: https://play.golang.org/p/cQ6C1P4qBBR - the syntax is correct but it just doesn't work.

So we are going to have to think of a different idea...

You could try something like this (untested)

```diff
diff --git a/backend/webdav/api/types.go b/backend/webdav/api/types.go
index 8fe89279d..5662ad800 100644
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -190,6 +190,13 @@ func (t *Time) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
 		return nil
 	}
 
+	// If time has two timezones, eg Salesforce Commerce Cloud
+	// chop one of them off
+	// See: https://github.com/ncw/rclone/issues/3048
+	if strings.HasSuffix("GMT-00:00") {
+		v = v[:len(v)-6]
+	}
+
 	// Parse the time format in multiple possible ways
 	var newT time.Time
 	for _, timeFormat := range timeFormats {
```

danielloader:
I've implemented the fix you suggested (after reading the docs to realise I needed to pass a variable to check first!) - that said, no dice after building it - still fails. 

Going to assume it's unusual for a webdav environment to have two types of timestamp in one request?

ncw:
> I've implemented the fix you suggested (after reading the docs to realise I needed to pass a variable to check first!) - that said, no dice after building it - still fails.

Did you try printing out the input? and the output?

What was the error message?

> Going to assume it's unusual for a webDAV environment to have two types of timestamp in one request?

Yes, but there are a lot of WebDAV implementations out there!

> As mentioned in the issue thread, put in a request for support to confirm or deny if this is a correct implementation or a bug they can fix.
> 
> In the mean time, is this an edge case worth checking for going forth?

If we can fix it easily then I'm happy to take a fix.

ncw:
How are you doing @danielloader ?  This commit looks done, but I think we were wondering whether to merge it at all?

What do you think?

danielloader:
Wow times flown, only just seen this notification lurking. I'd say it's no longer relevant as they've fixed it upstream to be compliant!

ncw:
> Wow times flown, only just seen this notification lurking. I'd say it's no longer relevant as they've fixed it upstream to be compliant!

The best kind of bug fix :-)

Issue #3088, opened 2019-04-04
Add support for AWS S3 Glacier Deep Archive Storage Class


m3nu:
#### What problem are you are trying to solve?
AWS recently launched a cheaper storage tier for S3, called "Glacier Deep Archive". 

#### How do you think rclone should be changed to solve that?
Rclone already [supports](https://rclone.org/s3/) setting different S3 storage classes. It would be useful to add support for the [new storage class](https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-class-intro.html#sc-glacier) `DEEP_ARCHIVE`.

m3nu:
Simply adding the `DEEP_ARCHIVE` option seems to work. From what I understand, it's just passed as a header.

Restoring files needs to be done from the web interface. The docs already link to it, so I didn't change much there.

PR #3089, opened 2019-04-04
s3: Add support for Glacier Deep Archive and Intelligent-Tiering

Deewiant:

Support for AWS S3 storage classes that currently aren't supported.

#3088 mentions Glacier Deep Archive. Intelligent-Tiering hasn't been discussed as far as I can tell, but it seems like a no-brainer to support all possible storage classes.

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate. (`GLACIER` didn't come with tests so I figured this doesn't need any either, but I tested `DEEP_ARCHIVE` manually since I actually wanted to use it, and it works with these changes.)

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -653,6 +653,9 @@ isn't set then "acl" is used instead.`,
 			}, {
 				Value: "DEEP_ARCHIVE",
 				Help:  "Glacier Deep Archive storage class",
+			}, {
+				Value: "INTELLIGENT_TIERING",
+				Help:  "Intelligent-Tiering storage class",
 			}},
 		}, {
 			// Mapping from here: https://www.alibabacloud.com/help/doc-detail/64919.htm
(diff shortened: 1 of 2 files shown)


ncw:
Apologies for the delay in responding, I've been on vacation.

I merged #3088 to get it in the next release.

I don't want to update the SDK this close to the release though so I'll merge this one after the release.

Unless you think this could work without an SDK update?  If so then can you drop the other parts of the PR?

Deewiant:
I don't know whether the SDK update is required, I simply assumed that the SDK would need to support something before `rclone` can use it. If #3091 works by itself then fair enough. Unfortunately I won't have the opportunity to test that for some time — I can only confirm that with the SDK update it does work.

ncw:
I've updated the SDK and we've had another PR for DEEP ARCHIVE so I've dropped those parts of your PR.  I've pushed the rebased remaining to your branch, and hopefully it will go green so I can merge - thank you :-)

ncw:
Looks good - will merge now - thank you very much :-)

Deewiant:
Thanks for picking this up, and apologies for never getting back to it myself.

PR #3120, opened 2019-04-18
SFTP: add about support

### LINKED ISSUE #3107: SFTP support for `about` command

```
rclone v1.46
- os/arch: windows/amd64
- go version: go1.11
```

I'd love to use the `about` command for the SFTP remote type. Currently:

```
.\rclone about remote:
2019/04/10 14:59:54 Failed to about: sftp://user@remote: doesn't support about
```

Similarly to how rclone uses md5 and sha1sum commands if it can to generate hashes, it would be great if it could try and use the `df` command to query space for the current device.

Since a remote system can have multiple disks, i'd suggest returning the information for whatever disk is currently mounted to the target remote folder. This can be done by supplying a path argument to `df`.

Not all of the information for command would be available, but `about` doesn't guarantee all info is available.

gary-kim:
Fixes #3107
This adds support for using about with SFTP remotes.

Currently, this does not check to ensure that it is possible to invoke `df`. Should this be added, and if so, when should the check occur?

The `-k` flag in `df -k` should ensure even non-GNU `df` returns with 1k-blocks.

I've tested this on my Ubuntu Linux server. It still needs to be tested on other platforms.

Also, for some reason, [my Travis build](https://travis-ci.com/gary-kim/rclone/builds/108866098) is passing but the one here is not?

Add support for the `about` command for SFTP remotes

#3107 

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -14,6 +14,7 @@ import (
 	"os/user"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -813,6 +814,47 @@ func (f *Fs) Hashes() hash.Set {
 	return set
 }
 
+// About gets usage stats
+func (f *Fs) About() (*fs.Usage, error) {
+	c, err := f.getSftpConnection()
+	if err != nil {
+		return nil, errors.Wrap(err, "About get SFTP connection")
+	}
+	session, err := c.sshClient.NewSession()
+	f.putSftpConnection(&c, err)
+	if err != nil {
+		return nil, errors.Wrap(err, "About put SFTP connection")
+	}
+
+	var stdout, stderr bytes.Buffer
+	session.Stdout = &stdout
+	session.Stderr = &stderr
+	escapedPath := shellEscape(f.root)
+	if f.opt.PathOverride != "" {
+		escapedPath = shellEscape(path.Join(f.opt.PathOverride, f.root))
+	}
+	if len(escapedPath) == 0 {
+		escapedPath = "/"
+	}
+	err = session.Run("df -k " + escapedPath)
+	if err != nil {
+		_ = session.Close()
+		return nil, errors.Wrap(err, "About invocation of df failed. Your remote may not support about.")
+	}
+	_ = session.Close()
+
+	usageTotal, usageUsed, usageAvail := parseUsage(stdout.Bytes())
+	if usageTotal < 0 || usageUsed < 0 || usageAvail < 0 {
+		return nil, errors.Wrap(err, "About failed to parse information")
+	}
+	usage := &fs.Usage{
+		Total: fs.NewUsageValue(usageTotal),
+		Used:  fs.NewUsageValue(usageUsed),
+		Free:  fs.NewUsageValue(usageAvail),
+	}
+	return usage, nil
+}
+
 // Fs is the filesystem this remote sftp file object is located within
 func (o *Object) Fs() fs.Info {
 	return o.fs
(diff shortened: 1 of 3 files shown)


backend/sftp/sftp.go
```diff
@@ -903,6 +944,29 @@ func parseHash(bytes []byte) string {
...
+	lines := strings.Split(string(bytes), "\n")
+	if len(lines) < 2 {
+		return -1, -1, -1
+	}
+	split := regexp.MustCompile(" +").Split(lines[1], -1)
+	if len(split) < 6 {
+		return -1, -1, -1
+	}
+	fmt.Print(split)
+	spaceTotal, err := strconv.ParseInt(split[len(split) - 5], 10, 64)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/sftp/sftp.go
```diff
@@ -903,6 +944,29 @@ func parseHash(bytes []byte) string {
...
+	spaceUsed, err := strconv.ParseInt(split[len(split) - 4], 10, 64)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/sftp/sftp.go
```diff
@@ -813,6 +814,46 @@ func (f *Fs) Hashes() hash.Set {
...
+		escapedPath = shellEscape(path.Join(f.opt.PathOverride, f.root))
+	}
+	if len(escapedPath) == 0 {
+		escapedPath = "/"
+	}
+	err = session.Run("df -k " + escapedPath)
+	if err != nil {
+		_ = session.Close()
+	}
+	_= session.Close()
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/sftp/sftp.go
```diff
@@ -813,6 +814,46 @@ func (f *Fs) Hashes() hash.Set {
...
+	}
+	_ = session.Close()
+
+	usageTotal, usageUsed, usageAvail := parseUsage(stdout.Bytes())
+	if usageTotal == -1 || usageUsed == -1 || usageAvail == -1 {
+		return nil, errors.Wrap(err, "About failed to parse information")
+	}
+	usage := &fs.Usage{
+		Total: fs.NewUsageValue(usageTotal),
+		Used: fs.NewUsageValue(usageUsed),
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ncw:
It looked like the build got stuck for go1.9 - I gave it a kick...

backend/sftp/sftp.go
```diff
@@ -903,6 +944,34 @@ func parseHash(bytes []byte) string {
...
+// Parses the byte array output from the SSH session
+// returned by an invocation of df into
+// the disk size, used space, and avaliable space on the disk, in that order.
+// Only works when `df` has output info on only one disk
+func parseUsage(bytes []byte) (int64, int64, int64) {
+	lines := strings.Split(string(bytes), "\n")
+	if len(lines) < 2 {
+		return -1, -1, -1
+	}
+	split := regexp.MustCompile(" +").Split(lines[1], -1)
```
ncw: If you are splitting on whitespace only then I would have thought `strings.Fields` would work here wouldn't it?

backend/sftp/sftp.go
```diff
@@ -903,6 +944,34 @@ func parseHash(bytes []byte) string {
 	return strings.Split(string(bytes), " ")[0] // Split at hash / filename separator
 }
 
+// Parses the byte array output from the SSH session
+// returned by an invocation of df into
+// the disk size, used space, and avaliable space on the disk, in that order.
+// Only works when `df` has output info on only one disk
+func parseUsage(bytes []byte) (int64, int64, int64) {
```
ncw: I'd really like to see a test for parseUsage.  I'd write a table driven test like the ones here: https://github.com/ncw/rclone/blob/master/backend/sftp/sftp_internal_test.go

We can then get the output of `df -k` from various OSes as test cases.

ncw:
> Currently, this does not check to ensure that it is possible to invoke `df`. Should this be added, and if so, when should the check occur?

Perhaps what we should do is attempt to execute `df` in the `About` as we are at the moment, but if it fails, then don't call it again.

> The `-k` flag in `df -k` should ensure even non-GNU `df` returns with 1k-blocks.
> 
> I've tested this on my Ubuntu Linux server. It still needs to be tested on other platforms.

Sure.

backend/sftp/sftp.go
gary-kim: Sorry, I didn't actually know about `strings.Fields`. I've changed it to use `strings.Fields` now.

backend/sftp/sftp.go
gary-kim: Okay, I put 2 test cases in.

gary-kim:
> Perhaps what we should do is attempt to execute `df` in the `About` as we are at the moment, but if it fails, then don't call it again.

For now, I added a error for not being able to invoke `df` that mentions it may be because the remote does not support it.

ncw:

> For now, I added an error for not being able to invoke df that mentions it may be because the remote does not support it

I think that is probably OK, so I'm going to merge it now - thank you very much :-)

PR #3134, opened 2019-04-25
moveto: fix case-insensitive same remote move

### LINKED ISSUE #3104: moveto deletes file when trying to do a case-sensitive rename.

When using the `moveto` command to rename the file by changing it's case, the file gets deleted instead of renamed in some environments.

v1.46.0-112-g1c301f9f-beta

Windows 7, 64 bit

Windows Local, Jottacloud

rclone moveto R:\test.bin R:\TEST.bin

```
2019/04/08 19:02:14 DEBUG : rclone: Version "v1.46.0-112-g1c301f9f-beta" starting with parameters ["[...]\\rclone.exe" "moveto" "R:\\test.bin" "R:\\TEST.bin" "-vv"]
2019/04/08 19:02:14 DEBUG : Using config file from "[...]\\rclone.conf"
2019/04/08 19:02:14 DEBUG : test.bin: Size and modification time the same (differ by 0s, within tolerance 100ns)
2019/04/08 19:02:14 DEBUG : test.bin: Unchanged skipping
2019/04/08 19:02:14 INFO  : test.bin: Deleted
2019/04/08 19:02:14 INFO  :
Transferred:            0 / 0 Bytes, -, 0 Bytes/s, ETA -
Errors:                 0
Checks:                 2 / 2, 100%
Transferred:            0 / 0, -
Elapsed time:       200ms

2019/04/08 19:02:14 DEBUG : 2 go routines active
```

gary-kim:
Fixes #3104

Fix a bug where moveto would delete a file when trying to change the case of a name on a remote that is case insensitive.

#3104 

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"path"
 	"path/filepath"
 	"sort"
@@ -423,8 +424,8 @@ func Move(fdst fs.Fs, dst fs.Object, remote string, src fs.Object) (newDst fs.Ob
 	}
 	// See if we have Move available
 	if doMove := fdst.Features().Move; doMove != nil && (SameConfig(src.Fs(), fdst) || (SameRemoteType(src.Fs(), fdst) && fdst.Features().ServerSideAcrossConfigs)) {
-		// Delete destination if it exists
-		if dst != nil {
+		// Delete destination if it exists and is not the same file as src (could be same file while seemingly different if the remote is case insensitive)
+		if dst != nil && (!fdst.Features().CaseInsensitive || strings.ToLower(dst.Remote()) != strings.ToLower(src.Remote())) {
 			err = DeleteFile(dst)
 			if err != nil {
 				return newDst, err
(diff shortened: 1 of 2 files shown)


fs/operations/operations.go
```diff
@@ -398,8 +398,8 @@ func Move(fdst fs.Fs, dst fs.Object, remote string, src fs.Object) (newDst fs.Ob
 	}
 	// See if we have Move available
 	if doMove := fdst.Features().Move; doMove != nil && (SameConfig(src.Fs(), fdst) || (SameRemoteType(src.Fs(), fdst) && fdst.Features().ServerSideAcrossConfigs)) {
-		// Delete destination if it exists
-		if dst != nil {
+		// Delete destination if it exists and is not the same file as src (could be same file while seemingly different if the remote is case insensitive)
+		if dst != nil && (!fdst.Features().CaseInsensitive || strings.ToLower(dst.String()) != strings.ToLower(src.String())) {
```
ncw: I think you want `dst.Remote()`  instead of `dst.String()` same with `src`.

gary-kim: Okay, this change has been committed.

gary-kim:
I'll try to write a test for this. Give me a bit to figure it out, I'm kind of busy at the moment and I'm not sure how to write a test for this scenario.

ncw:
> I'll try to write a test for this. Give me a bit to figure it out, I'm kind of busy at the moment and I'm not sure how to write a test for this scenario.

No worries!

I'd start by copying this test (or maybe adding a bit to the end of it) which tries to do the move from "upper" to "UPPER" and make sure the right thing happens.  This will get run on the integration tests and check all the different backends.

https://github.com/ncw/rclone/blob/99bc013c0a23d9dd5062600429a1488ae00cd277/fs/operations/operations_test.go#L713

gary-kim:
Okay, so while writing the test, I noticed the issue does not seem to be fully resolved. It turns out this is not quite as simple to solve. It will require changes to a lot of functions. A lot of functions can't handle this situation. `moveOrCopyFile` needs to be changed to know that a move operation does not necessarily mean the source can be deleted. `Move` still needs the above change. `NeedTransfer` needs a case for when the dst and src are the same file but with different case. On top of all that, even once everything else is fixed, OneDrive will return a `nameAlreadyExists` error when attempting to move a file in this way. I wouldn't be surprised if other issues come up as well. It may be easier to add a special case for changing the case of files on a remote.

ncw:
> If this can't be fixed before the next release, maybe a temporary error should be added for this case?

I think a partial fix is better than no fix :-)

Maybe we should merge your original code as it prevents data loss and think more about the other cases.

The `NeedsTransfer` case is probably less important because the march routine should be doing case insensitve/case sensitive match-ups as appropriate already.

gary-kim:
> Maybe we should merge your original code as it prevents data loss and think more about the other cases.

The current version seems to still delete files. I'm about to push a change that at least prevents data loss. It still cannot rename the file though.

> The `NeedsTransfer` case is probably less important because the march routine should be doing case insensitve/case sensitive match-ups as appropriate already.

I thought it needs to be changed because even once the deleting issue is solved, this is shown:
```
019/05/20 11:32:18 DEBUG : testfile.txt: Size and modification time the same (differ by 0s, within tolerance 1s)
2019/05/20 11:32:18 DEBUG : testfile.txt: Unchanged skipping
```

I was thinking a bit of a messy way to get by the issue was to rename the file to a random name then move it back. That would also avoid the `nameAlreadyExists` issue.

ncw:
Is this ready to merge - what do you think?

gary-kim:
At the moment, this kind of just pushes the problem back to be dealt with later. It's better than nothing but it also feels weird to call this a fix. 
What would you think about having this case simply move the file twice to rename? The idea would be to find a random file name that is not used, move it there, then move it back with the proper capitalization. It's messy but it would work.

ncw:
> At the moment, this kind of just pushes the problem back to be dealt with later. It's better than nothing but it also feels weird to call this a fix.

OK.

> What would you think about having this case simply move the file twice to rename? The idea would be to find a random file name that is not used, move it there, then move it back with the proper capitalization. It's messy but it would work.

I think that would be fine.  I'd probably take the existing name and add some random junk on the end for the first rename so if it went wrong half way through the user wouldn't be completely lost.  Do do the move recursing is probably the best thing to do.

gary-kim:
The most recent commit can now handle case insensitive renames and includes a test to ensure the issues does not happen again.

ncw:
> Also, I'm not sure why but the build is complaining about an ineffectual assignment in the box backend?

That was a short lived update from golang-ci which has now been reverted.  I've kicked the build again!

gary-kim:
@ncw 
Looks like the build was successful this time. 
The pull request is ready to be reviewed.

ncw:
This looks good now - thank you very much :-)

I'll merge it now.

PR #3142, opened 2019-04-26
mega: add cleanup support

### LINKED ISSUE #3138: Empty Trash (in particular in Mega)
While using cloud storage, one of the challenges is always being under storage limit. Good practice is to delete not needed material.

Each of the remotes have a different approach, but, usually, deleted items are temporarily stored in a "to be finally deleted" area -TBFDA, that is usually purged after x days of staying there.

The space used by that TBFDA is counted towards storage limits. So, if one is short, the best is to force an inmediate purge TBFDA. "native" web interfaces allow to do it; unsure if APIs provide the function.

I did not find a mean to force purge of TBFDA in rclone for Mega.

In Mega, there are several "system" folders, one is /Root and other is /Trash
The only way to do this that I found in CMD was to use megatools, that allows to list /Trash and to delete item by item in /Trash.

I have not found any way to use (list, delete items) mega /Trash with rclone (but documentation is so vast that I might be wrong....). rlcone ls megaremote:File is always refering to /Root/File.  But in W10 CMD has also some issues. When there are too many items to delete including non-ANSI names.... web interface is the only resort I found.

I believe rclone can profit from stablishing a framework to purge TBFDA, so that internally could use the best way for each remote, and avoid the pitfalls with non-ANSI names; besides, it could be done with just a single login/logout session.

(I have been pested lately for what it seems throttling by mega in terms of how often one logins/logouts, as some rclone operations produce sucesive login/logout - this is still under investigation).

For mega, the algorithm is not complex:
a) login
b) list recursively contents of /Trash
c) delete each item under /Trash and any folder under /Trash
d) list remaining empty space (optional)
e) logout

TIA,
CT

gary-kim:
Fixes #3138

Add `cleanup` support for Mega remote

#3138 

#### Checklist

Also, I really need to remember to run golint.

### DIFF
--- a/backend/mega/mega.go
+++ b/backend/mega/mega.go
@@ -402,6 +402,27 @@ func (f *Fs) clearRoot() {
 	//log.Printf("cleared root directory")
 }
 
+// CleanUp deletes all files currently in trash
+func (f *Fs) CleanUp() (err error) {
+	trash := f.srv.FS.GetTrash()
+	items := []*mega.Node{}
+	_, err = f.list(trash, func(item *mega.Node) bool {
+		items = append(items, item)
+		return false
+	})
+	if err != nil {
+		return errors.Wrap(err, "CleanUp failed to list items in trash")
+	}
+	// similar to f.deleteNode(trash) but with HardDelete as true
+	for _, item := range items {
+		err = f.pacer.Call(func() (bool, error) {
+			err = f.srv.Delete(item, true)
+			return shouldRetry(err)
+		})
+	}
+	return err
+}
+
 // Return an Object from a path
 //
 // If it can't be found it returns the error fs.ErrorObjectNotFound.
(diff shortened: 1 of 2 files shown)


backend/mega/mega.go
```diff
@@ -402,6 +402,27 @@ func (f *Fs) clearRoot() {
 	//log.Printf("cleared root directory")
 }
 
+// Deletes all files currently in trash
```
golangcibot: comment on exported method `Fs.CleanUp` should be of the form `CleanUp ...` (from `golint`)

ncw:
Nice!  Did you run the integration tests for this?  `go test` in the `backend/mega` directory with a `TestMega` remote setup?

gary-kim:
I ran the integration test before committing. There was one error but it was from `PublicLink`. That will need to be fixed separately.

ncw:
> I ran the integration test before committing. There was one error but it was from `PublicLink`. That will need to be fixed separately.

The PublicLink test has been broken for a long time.  The feature works, but the test is broken - ideas to fix appreciated!

ncw:
I've given the travis build a kick - it has been really unreliable recently :-(

gary-kim:
Looks like the build passed now. Any idea why it seems to be failing so often?

caspertone2003:
> Looks like the build passed now. Any idea why it seems to be failing so often?

Would not wish to be pushy... are you going to provide a test/branch release with this feature any time soon so I can try it?
Thanks

gary-kim:
Sorry, this is on my own repo since I don't have write access here so I can't use the beta release system as it is. I'll try to trigger a build for an unofficial version.

gary-kim:
Okay, I built an unofficial version. Available here: https://nc.ydgkim.com/s/ebCoLfGKHaKcQH5
Switch back to ncw's version as soon as possible.

ncw:
Sorry for the delay merging this.  The code is great - I've just been a bit overloaded - sorry!

I'll merge it now - thank you very much.

PR #3149, opened 2019-04-30
fserrors: use errors.Walk for the wrapped error types

B4dM4n:

This change uses the `lib/errors` in the `fs/fserrors` package to test for specific errors in an error chain.

All wrapped errors in the `fs/fserrors` package now have a `Cause() error` method to allow nested errors to be also checked with `errors.Walk()`.

Uses features introduced by #2962

#### Checklist

### DIFF
--- a/fs/fserrors/error.go
+++ b/fs/fserrors/error.go
@@ -5,11 +5,10 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"reflect"
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


ncw:
That looks much neater now - thank you :-) Will merge now.

I don't know if you noticed but I fixed a bug in lib/errors in 7963320a29853da9342306ede7dfb483f67c340c - it is possible to make errors which are uncomparable and panic, so doing `err1 == err2` is potentially not safe!  It won't panic if one of err1, err2 is comparable and the other isn't as they are definitely different at that point. So this is a bit of a corner case but interesting!

PR #3177, opened 2019-05-08
Adds a public IP flag for ftp.  Closes #3158

### LINKED ISSUE #3158: serve ftp: advertise configured passive addr
vsftpd alllows us to set a advertised ip address for PASSIVE connections so that if a ftp server is located behind a firewall and the client is also behind its own firewall that the connection can be advertised to the client correctly.  An example would be a ftp server on GCP which has a VMs external ip address behind a device.

pasv_address=X.X.X.X

can we add a similar parameter to rclone serve ftp to allow this?

pasv_address
Use this option to override the IP address that vsftpd will advertise in response to the PASV command. Provide a numeric IP address, unless pasv_addr_resolve is enabled, in which case you can provide a hostname which will be DNS resolved for you at startup.
Default: (none - the address is taken from the incoming connected socket)

The issue is that the address currently advertised on rclone is the address on the incoming socket which confuses a client to connecting to the internal interface of a VM on GCP for example resulting in failures.

calisro:
#### Checklist

- [x ] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [x ] All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).

### DIFF
--- a/cmd/serve/ftp/ftp.go
+++ b/cmd/serve/ftp/ftp.go
@@ -78,6 +78,7 @@ func newServer(f fs.Fs, opt *ftpopt.Options) (*server, error) {
 		},
 		Hostname:     host,
 		Port:         portNum,
+		PublicIp:     opt.PublicIP,
 		PassivePorts: opt.PassivePorts,
 		Auth: &Auth{
 			BasicUser: opt.BasicUser,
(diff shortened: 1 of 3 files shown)


ncw:
That looks great - thank you.

Will merge now

PR #3190, opened 2019-05-12
install: linux skip man pages if no mandb - fixes #3175

### LINKED ISSUE #3175: bash: line 157: mandb: command not found
When i installed rclone i got this error message:
`bash: line 157: mandb: command not found`

Instead of showing this error message it must be easy to check if mandb exists and instead show  a nice message and tell the user that rclone will still work but the help can be limited when the `man` command is not available.

rclone v1.47.0

Linux dietpi 4.14.98-v7+

didil:
Skip man docs during install if linux mandb is not installed

https://github.com/ncw/rclone/issues/3175
#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -152,9 +152,13 @@ case $OS in
     chown root:root /usr/bin/rclone.new
     mv /usr/bin/rclone.new /usr/bin/rclone
     #manuals
-    mkdir -p /usr/local/share/man/man1
-    cp rclone.1 /usr/local/share/man/man1/
-    mandb
+    if ! [ -x "$(command -v mandb)" ]; then
+        echo 'mandb not found. The rclone man docs will not be installed.'
+    else 
+        mkdir -p /usr/local/share/man/man1
+        cp rclone.1 /usr/local/share/man/man1/
+        mandb
+    fi
     ;;
   'freebsd'|'openbsd'|'netbsd')
     #bin


docs/content/install.sh
```diff
@@ -152,9 +152,13 @@ case $OS in
     chown root:root /usr/bin/rclone.new
     mv /usr/bin/rclone.new /usr/bin/rclone
     #manuals
-    mkdir -p /usr/local/share/man/man1
-    cp rclone.1 /usr/local/share/man/man1/
-    mandb
+    if ! [ -x "$(command -v mandb)" ]; then
```
ncw: This runs the mandb command twice which isn't desirable since it can take quite a long time...

How about using the detection we use above for discovering the zip tools?

```sh
    trash=`hash $tool 2>>errors`
    if [ "$?" -eq 0 ]; then
        unzip_tool="$tool"
        break
    fi
```

didil: Sure I can do that, but I don't think it runs the command twice, I think it only checks that it exists.
if you replace mandb with a script that has a side effect such as deleting a file, you'll see that it doesn't run

ncw: Sorry I mis-read the man page you are right.

I'll merge what you've done now :-)

ncw:
Thank you for that.  I've pushed it live so if there is a problem we should get feedback soon!

didil:
Great !

Issue #2098, opened 2018-02-26
Consider supporting compression for crypt remotes


jikamens:
It's relatively common to compress plaintext data before encrypting it.

This is because you can't really compress it after encrypting it because good compression randomizes the data from the point of view of the compression algorithm and makes it compress very poorly.

Furthermore, since the data is randomized by compression, it can't be de-duplicated by the server, so even if the storage remote backing the crypt remote supports de-duplication, you still end up paying to store every encrypted byte.

It therefore might be useful if the crypt remote type supported (optional) compression which could potentially provide savings in terms of storage cost, upload and download bandwidth, and upload and download time.

If anybody ever ends up acting on this feature request, two additional things t consider:

1) Make sure a mix of compressed and uncompressed encrypted files is supported -- i.e., indicate in the header of the uploaded file whether it is compressed or uncompressed and react appropriately when downloading -- so that there is a backward-compatible path to enabling compression without having to re-upload all the files in the remote;

2) Dynamically determine while uploading whether individual files should be compressed or whether they're already compressed so additional compression won't help. For example, I played around with this a bit and found that a good algorithm is something like this:

* Read and compress the first 10MB of the file (big enough chunk to get a good idea whether compression will be useful on the entire file).
* If the file size is less than 10MB, then:
  * If the compressed blob is smaller than the uncompressed data, then upload the file compressed.
  * Else upload the file uncompressed.
* Else if the compressed 10MB blob is 25% or more smaller than the uncompressed blob, then upload the file compressed.
* Else upload the file uncompressed.

ncw:
I think I'd implement this as a layer over the encrypted remote and I'd encrypt in chunks and save in many different files.  Rclone relies on being able to seek to individual points in the file which you'd lose if you just compressed the file.

jikamens:
That's going to cause problems for things like file versioning on B2.

Also, it's probably going to cause problems similar to #1627, so you'd want to make sure to fix #1627 as well as avoid a similar situation here.

id01:
I think doing something similar to dictzip (https://github.com/dictzip/dictzip-go) would allow for more seekable gzip compression without needing to handle multiple files; there are tools for other compression algorithms (like pixz) that take advantage of compression blocks to provide seekable compression as well. Might want to do something similar.

ncw:
[...]

PR #135, opened 2015-09-12
ACD: Retry when we get a "500 Internal server error".

klauspost:
It seems like this happens randomly, and retrying works fine for that.

Issue #3154, opened 2019-05-01
ci: cmd/mount test causing build timeouts


ncw:
The cmd/mount test has been hanging on Travis for unknown reasons.

```
?   	github.com/[secure]/rclone/cmd/mkdir	[no test files]

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

In a successful run it takes about 10s

```
?   	github.com/[secure]/rclone/cmd/mkdir	[no test files]
ok  	github.com/[secure]/rclone/cmd/mount	10.648s
```

ncw:
I ran the tests 100 times locally with no hangs

ncw:
Adding `-v` to `go test` makes it less likely to happen but it still happens.

I observed in real time the hang after the "mkdir" test.

gary-kim:
I'm testing something at the moment on Travis that seems like it fixes the issue.

ncw:
> I'm testing something at the moment on Travis that may fix this issue.

That would be fantastic :-)

ncw:
I finally managed to reproduce this locally....

With

```patch
diff --git a/Makefile b/Makefile
index 3cefc2625..492730185 100644
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,10 @@ test:       rclone
 
 # Quick test
 quicktest:
-       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) $(GO_FILES)
+       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -count=1 $(GO_FILES)
 
 racequicktest:
-       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -cpu=2 -race $(GO_FILES)
+       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -count=1 -cpu=2 -race $(GO_FILES)
 
 # Do source code quality checks
 check: rclone
```

And

    ( for i in `seq 1000`; do echo `date -Is`: Trial $i ; make quicktest ; done ) 2>&1 | tee quicktest.log

When it hangs the process graph looks like this

```
 3217 pts/6    S+     0:00      |   |   \_ make quicktest
[...]

PR #3194, opened 2019-05-13
press: Added experimental compression remote - implements #2098, #135…

id01:

Adding compression, a wanted feature.

Yes. Discussion was [here](https://github.com/ncw/rclone/issues/2098), and there were also [these](https://github.com/ncw/rclone/issues/675) [requests](https://github.com/ncw/rclone/issues/1356) from a while ago.

#### Checklist

Note: I also added a second commit that lets lib/env/env_test.go work on Windows so that AppVeyor will pass.


backend/press/alg_lz4.go
```diff
@@ -0,0 +1,89 @@
...
+	_, err = h.Write(compressedBytes[4:])
+	if err != nil {
+		return 0, err
+	}
+	if binary.LittleEndian.Uint32(hash) != h.Sum32() {
+		return 0, errors.New("XXHash checksum invalid")
+	}
+	dst := make([]byte, BlockSize*2)
+	decompressed, err := lz4.Decode(dst, compressedBytes)
+	out.Write(decompressed)
```
golangcibot: Error return value of `out.Write` is not checked (from `errcheck`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+
+// CompressFile compresses a file. Argument "size" is ignored.
+func (c *Compression) CompressFile(in io.Reader, size int64, out io.Writer) error {
+	// Initialize buffered writer
+	bufw := bufio.NewWriterSize(out, int(c.maxCompressedBlockSize()*uint32(c.NumThreads)))
+
+	// Get blockData, copy over header, add length of header to blockData
+	blockData := make([]byte, 0)
+	header := c.getHeader()
+	bufw.Write(header)
```
golangcibot: Error return value of `bufw.Write` is not checked (from `errcheck`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+		// Process writers in order
+		for i := 0; i < c.NumThreads; i++ {
+			if compressionResults[i] != nil {
+				// Get current compression result, get buffer, and copy buffer over to output
+				res := <-compressionResults[i]
+				close(compressionResults[i])
+				if res.buffer == nil {
+					return res.err
+				}
+				io.Copy(bufw, res.buffer)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+			if DEBUG {
+				log.Printf("%d", eofAt)
+				log.Printf("%v", blockData)
+			}
+			break
+		}
+	}
+
+	// Write footer
+	bufw.Write(c.getFooter())
```
golangcibot: Error return value of `bufw.Write` is not checked (from `errcheck`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+			// Spawn thread to decompress block
+			if DEBUG {
+				log.Printf("Spawning %d", i)
+			}
+			go func(i int, currBlock uint32, in io.Reader) {
+				var block bytes.Buffer
+				var res DecompressionResult
+
+				// Decompress block
+				d.decompressBlockRange(in, &block)
```
golangcibot: Error return value of `d.decompressBlockRange` is not checked (from `errcheck`)

backend/press/compression_test.go
```diff
@@ -0,0 +1,147 @@
...
+	compressedFile, err := ioutil.TempFile(os.TempDir(), "rclone_compression_test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	outHasher := md5.New()
+
+	// Compress file and hash it (size doesn't matter here)
+	testFileReader, testFileWriter := io.Pipe()
+	go func() {
+		io.Copy(io.MultiWriter(testFileHasher, testFileWriter), testFile)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

backend/press/compression_test.go
```diff
@@ -0,0 +1,147 @@
...
+		testFileWriter.Close()
```
golangcibot: Error return value of `testFileWriter.Close` is not checked (from `errcheck`)

backend/press/compression_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}()
+	comp.CompressFile(testFileReader, 0, compressedFile)
```
golangcibot: Error return value of `comp.CompressFile` is not checked (from `errcheck`)

backend/press/alg_exec.go
```diff
@@ -0,0 +1,124 @@
+package press
+
+// This file implements shell exec algorithms that require binaries.
+// Requires alg_gzip.go for gzip-wrapped files (cough cough lzma).
+import (
+	"os/exec"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/alg_gzip.go
```diff
@@ -0,0 +1,49 @@
+package press
+
+// This file implements the gzip algorithm.
+import (
+	"compress/gzip"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/alg_lz4.go
```diff
@@ -0,0 +1,89 @@
+package press
+
+// This file implements the LZ4 algorithm.
+import (
+	"io"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/alg_snappy.go
```diff
@@ -0,0 +1,34 @@
+package press
+
+// This file implements compression/decompression using snappy.
+import (
+	"bytes"
+	"io"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+
+/*
+NOTES:
+Structure of the metadata we store is:
+gzipExtraify(gzip([4-byte header size][4-byte block size] ... [4-byte block size][4-byte raw size of last block]))
+This is appended to any compressed file, and is ignored as trailing garbage in our LZ4 and SNAPPY implementations, and seen as empty archives in our GZIP and XZ_IN_GZ implementations.
+*/
+
+import (
+	"log"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/press.go
```diff
@@ -0,0 +1,837 @@
+// Package press provides wrappers for Fs and Object which implement compression.
+package press
+
+import (
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/compression_test.go
```diff
@@ -0,0 +1,147 @@
+package press
+
+import (
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/press_test.go
```diff
@@ -0,0 +1,90 @@
+// Test Crypt filesystem interface
+package press
+
+import (
+	"os"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/press/alg_exec.go
```diff
@@ -0,0 +1,124 @@
...
+	"io"
+	"bytes"
+)
+
+// XZ command
+const xzcommand = "xz" // Name of xz binary (if available)
+
+// Header we add to an exec file. We don't need this.
```
golangcibot: comment on exported var `ExecHeader` should be of the form `ExecHeader ...` (from `golint`)

backend/press/alg_gzip.go
```diff
@@ -0,0 +1,49 @@
...
+	"bufio"
+	"io"
+)
+
+// Header we add to a gzip file. We're contatenating GZIP files here, so we don't need this.
```
golangcibot: comment on exported var `GzipHeader` should be of the form `GzipHeader ...` (from `golint`)

backend/press/alg_lz4.go
```diff
@@ -0,0 +1,89 @@
...
+
+BD byte:
+	Reserved = 0
+	Block Max Size = 101 (or 5; 256kb)
+	Reserved = 0000
+
+Header checksum byte (xxhash(flags and bd byte) >> 1) & 0xff
+*/
+
+// Header and footer of our LZ4 file
```
golangcibot: comment on exported var `LZ4Header` should be of the form `LZ4Header ...` (from `golint`)

backend/press/alg_lz4.go
```diff
@@ -0,0 +1,89 @@
...
+var LZ4Header = []byte{0x04, 0x22, 0x4d, 0x18, 0x70, 0x50, 0x84}
+var LZ4Footer = []byte{0x00, 0x00, 0x00, 0x00} // This is just an empty block
```
golangcibot: exported var `LZ4Footer` should have comment or be unexported (from `golint`)

backend/press/alg_snappy.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/golang/snappy"
+)
+
+// Header we add to a snappy file. We don't need this.
```
golangcibot: comment on exported var `SnappyHeader` should be of the form `SnappyHeader ...` (from `golint`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+	GzipMin
+	GzipDefault
+	GzipMax
+	XzInGzMin
+	XzInGz
+	LZ4
+	Snappy
+)
+
+// Constants
```
golangcibot: comment on exported const `DEBUG` should be of the form `DEBUG ...` (from `golint`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+		case GzipStore, GzipMin, GzipDefault, GzipMax: return decompressBlockRangeGz(in, out)
+		case XzInGzMin: return decompressBlockRangeExecGz(in, out, d.c.BinPath, []string{"-dc1"})
+		case XzInGz: return decompressBlockRangeExecGz(in, out, d.c.BinPath, []string{"-dc"})
+		case LZ4: return decompressBlockLz4(in, out, int64(d.c.BlockSize))
+		case Snappy: return decompressBlockSnappy(in, out)
+	}
+	panic("Compression mode doesn't exist") // If none of the above returned
+}
+
+// Wrapper function for decompressBlockRange that implements multithreading
```
golangcibot: comment on exported type DecompressionResult should be of the form "DecompressionResult ..." (with optional leading article) (from `golint`)

backend/press/compression.go
```diff
@@ -0,0 +1,684 @@
...
+		*d.cursorPos += offset
+	} else if whence == io.SeekEnd {
+		*d.cursorPos = d.decompressedSize + offset
+	}
+
+	// Return
+	return offset, nil
+}
+
+// Decompresses a file. Argument "size" is very useful here.
```
golangcibot: comment on exported method `Compression.DecompressFile` should be of the form `DecompressFile ...` (from `golint`)

backend/press/press.go
```diff
@@ -0,0 +1,837 @@
...
+func (o *Object) Hash(ht hash.Type) (string, error) {
+	return "", hash.ErrUnsupported
+}
+
+// UnWrap returns the wrapped Object
+func (o *Object) UnWrap() fs.Object {
+	return o.Object
+}
+
+// Combines a Reader and a Closer to a ReadCloser
```
golangcibot: comment on exported type ReadCloserWrapper should be of the form "ReadCloserWrapper ..." (with optional leading article) (from `golint`)

backend/press/press.go
```diff
@@ -0,0 +1,837 @@
...
+func combineReaderAndCloser(dataSource io.Reader, closer io.Closer) *ReadCloserWrapper {
+	rc := new(ReadCloserWrapper)
+	rc.dataSource = dataSource
+	rc.closer = closer
+	return rc
+}
+func (w *ReadCloserWrapper) Read(p []byte) (n int, err error) {
+	return w.dataSource.Read(p)
+}
+func (w *ReadCloserWrapper) Close() error {
```
golangcibot: exported method `ReadCloserWrapper.Close` should have comment or be unexported (from `golint`)

backend/press/press.go
```diff
@@ -0,0 +1,847 @@
...
+	remote := opt.Remote
+	if strings.HasPrefix(remote, name+":") {
+		return nil, errors.New("can't point press remote at itself - check the value of the remote setting")
+	}
+	wInfo, wName, wPath, wConfig, err := fs.ConfigFs(remote)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse remote %q to wrap", remote)
+	}
+	// Strip trailing slashes if they exist in rpath
+//	rpath = strings.TrimRight(rpath, "\\/")
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ncw:
I cherry-picked the build fix - thank you :-)

Will review the press code later today.

backend/press/press.go
```diff
@@ -0,0 +1,870 @@
...
+Size() function is really inefficient because it requires opening and seeking through the file
+Files are now always compressed - the heuristic test function is defunct (for now)
+**/
+
+// Globals
+// Register with Fs
+func init() {
+	// Build compression mode options. Show XZ options only if they're supported on the current system.
+	compressionModeOptions := []fs.OptionExample{{ // Default compression mode options
+			Value: "lz4",
```
golangcibot: File is not `goimports`-ed (from `goimports`)

id01:
:+1: 

Just added some small logging, ux, and code structure changes. Ran some tests on my local machine and pushed. Shouldn't really impact the functionality. Now just awaiting your review :)

id01:
After looking back at the compression.go file, I found some panics on gzip failure along with certain functions that could be shortened and/or made more modular. Also, the testing file could have benefited from subtests instead of individual test cases. Pushed both of those things in another (smaller) commit.

ncw:
This is a very impressive piece of work - well done :-)

All the integration tests pass which is great!

I've been running some tests of my own and apart from this error they've worked very well.

```
rclone -P copy 2001/ presstestswift:2001test

Gave lots of

2019-05-19 14:23:01 ERROR : XXX/2001-10-23-115227.jpg: Unable to determine size: invalid seek position
```

However there is a problem.  I made a press remote on a swift cluster and copied a small directory structure full of jpegs to it.

Just listing the directory is extremely slow.  This is of course the reading of the sizes from the file itself.

```
$ time rclone -vv size presstestswift:2001test
Total objects: 1128
Total size: 460.350 MBytes (482712279 Bytes)

real	4m38.029s
user	0m14.635s
sys	0m4.643s
```

vs this on the underlying remote

```
$ time rclone size ospnickcaa2:press/2001test
Total objects: 1128
Total size: 460.506 MBytes (482875686 Bytes)

real	0m1.430s
user	0m0.131s
sys	0m0.029s
```

I think this would make syncing to a press remote unacceptably slow :-(

Other things I observed
- not being able to skip in-compressible objects is causing more CPU work than necessary so I think we should fix that
- I'm not mad keen on the external dependency of the `xz` command but as long as we document it the I'm OK with it.

I note your original scheme of adding the length into the file name.  This speeds up the listings at a cost of increasing the file name length.  I think we are going to have to do something like this in order to make sync speeds acceptable.

Another scheme might be to store metadata separately (one per file or per directory or some other scheme) which would have the advantage that we could store length, original hashes, original file name etc. That would fix one of the complaints people have about crypt that it doesn't allow hashes.

What do you think?

id01:
I'm not surprised; on certain remotes, I believe opening anywhere on the file at all will require a pretty large read.

I was thinking a similar thing as your second idea; as it's much easier to use deterministic filenames, it would be possible to use per-file metadata files to store hashes, block data, and file extension of compressed/uncompressed files. This would both make the remote faster and allow for more features, while still keeping the files independent.

Originally, I was thinking of adding this as a separate mode, as it would double the file count (which may be problematic with certain remotes), but if you want, I can remove the internal metadata storage scheme entirely.

id01:
I added mimetypes and hashes, along with the size and block data, to the metadata files (which are basically gzip-compressed gobs). Externally storing block data also means that xz compression can now work without wrapping it in gzip. Unfortunately, gzipping the metadata files tends to make most of them slightly larger, but as block data size scales at 4 bytes per 256KB block, compression would likely help for larger files, though I have yet to test with copying a large file. At the moment, it's pretty debatable whether gzip compression of the metadata files is actually a good idea.

I worked on the press.go code yesterday; I'll clean up some unused code in the compression.go code today.

ncw:
> I added mimetypes and hashes, along with the size and block data, to the metadata files (which are basically gzip-compressed gobs).

Great!

> Externally storing block data also means that xz compression can now work without wrapping it in gzip. Unfortunately, gzipping the metadata files tends to make most of them slightly larger, but as block data size scales at 4 bytes per 256KB block, compression would likely help for larger files, though I have yet to test with copying a large file. At the moment, it's pretty debatable whether gzip compression of the metadata files is actually a good idea.

I think keeping the metadata uncompressed is probably best which keeps things simple.
 
> I worked on the press.go code yesterday; I'll clean up some unused code in the compression.go code today.

What do you think performance is going to be like compared to keeping the metadata in the file?  It will certainly be faster, but how much?

id01:
I just cleaned up the compression code and uploaded.

I just tested with a 10GB vdi (along with different portions of it); it generates a 250KB uncompressed/170KB compressed metadata file due to the block indexing data. It seems like gzipping saves about 33% on large files, at the cost of 16-20 bytes of overhead on small files (and some computational power). However, seeing as the metadata for a single large file is compressible enough to compensate for thousands of small files, I'll probably leave metadata compression in.

I'm pretty sure that in certain remotes, keeping metadata in the file led to very slow lookups with large files as you need to download the entire file (or a large portion) just to get a part of it. With metadata, we have at most a 1MB download for huge files and a <1KB download for normal-sized files, which should speed things up by quite a bit.

Also, looks like this commit was affected by #3154.

id01:
Well, I couldn't really think of any other changes to make at this point, so I suppose I'm just waiting for your review again.

I would run a few speed tests myself, but I don't have a data storage cluster at home, and my internet is really slow. :P Maybe that's a good thing...

id01:
I decided to do a benchmark using Dropbox, and I'm getting the same problem - listing a directory of 313 files takes about 3 minutes. I monitored the network bandwidth going in and out, and it seems like now, instead of the bandwidth requirement of downloading large files, it's bottlenecked by the latency of opening each of these files - I only had 5KB upload/download max when listing. Switching to opening metadata files didn't seem to solve it, as we still have to open the same amount of files.

As appending sizes to filenames will cause a similar number of requests due to the number of lists that will be required to list a wrapped directory (unless said listing result is cached), it seems like the best way to solve this problem will be to keep a single database of sorts of the entries in every directory and cache it on the client-side.

id01:
Actually, adding a database will make the put operations much slower, as it will probably need to download the entire database and reupload it. Maybe it will be possible to split the database into multiple files?

Another, simpler, solution is if we put the sizes in the filenames, then cache the sizes upon listing or calling Size(). This way, we will be able to make the Size() calls much faster. Unfortunately, it's also a pretty ugly solution, as we have to be careful not to make duplicate files.

ncw:
Sorry for the radio silence, too much going on!

I was trying to reproduce the tests above to see what the speed looked like, but I got this error

```
 rclonevv -vv --low-level-retries 1 --retries 1 copy README.md presstestswift:
2019/06/06 16:09:48 DEBUG : rclone: Version "v1.47.0-DEV" starting with parameters ["rclone" "-vv" "-vv" "--low-level-retries" "1" "--retries" "1" "copy" "README.md" "presstestswift:"]
2019/06/06 16:09:48 DEBUG : Using config file from "/home/ncw/.rclone.conf"
2019/06/06 16:09:49 DEBUG : README.md: Couldn't find file - need to transfer
2019/06/06 16:09:49 DEBUG : pacer: low level retry 1/1 (error Put https://example.com/swift/v1/id/press/README.md.gz: net/http: HTTP/1.x transport connection broken: http: ContentLength=5268 with Body length 1676)
```

What that looks like to me is that in the src Object passed to the swift backend returned the wrong `Size()` when called so it returned 5268 bytes (which presumably is the uncompressed size) rather than 1676 bytes.

You can pass the size as -1 but this is only supported by certain backends.  Some backends **need** to know the size of the file being uploaded.  These are the ones not supporting [StreamUpload](https://rclone.org/overview/#optional-features).

I guess you don't know the size of the compressed file in advance...

`operations.Rcat` is a wrapper around the stream upload which can stream a file to any backend and spool it to disk if necessary which you could use.

> I decided to do a benchmark using Dropbox, and I'm getting the same problem - listing a directory of 313 files takes about 3 minutes. I monitored the network bandwidth going in and out, and it seems like now, instead of the bandwidth requirement of downloading large files, it's bottlenecked by the latency of opening each of these files - I only had 5KB upload/download max when listing. Switching to opening metadata files didn't seem to solve it, as we still have to open the same amount of files.

I think it will be quicker than your original scheme but you are right the round trip times are killing us.

> As appending sizes to filenames will cause a similar number of requests due to the number of lists that will be required to list a wrapped directory (unless said listing result is cached), it seems like the best way to solve this problem will be to keep a single database of sorts of the entries in every directory and cache it on the client-side.

A per directory cache file might work.  However simultaneous access to a directory is really going to ruin your day.

> Actually, adding a database will make the put operations much slower, as it will probably need to download the entire database and reupload it. Maybe it will be possible to split the database into multiple files?

Or we could use a log structured db where you create new directory entries in new files, then consolidate the entries as part of another process.

> Another, simpler, solution is if we put the sizes in the filenames, then cache the sizes upon listing or calling Size(). This way, we will be able to make the Size() calls much faster. Unfortunately, it's also a pretty ugly solution, as we have to be careful not to make duplicate files.

Putting the sizes in the filenames means we could cache them in the Object.  It will make `NewObject` difficult though as that requires you to open an object by name that you don't have a size for.  That would require a directory listing.  `NewObject` doesn't get called much though - only if you are using `--no-traverse` so it doesn't matter if it is a bit slow.  Objects are normally created from directory listings.

Rclone has ways of dealing with duplicate objects see `rclone dedupe` so there are ways of getting rid of them.

backend/press/press.go
```diff
@@ -0,0 +1,1239 @@
...
+	*entries = append(*entries, f.newDir(dir))
+}
+
+// Get an object from a metadata file
+func (f *Fs) addMetaParallel(out chan fs.DirEntry, mo fs.Object) {
+	meta := readMetadata(mo)
+	origFileName, _, err := processFileName(mo.Remote())
+	if err != nil {
+		fs.Errorf(mo, "Not a metadata file: %v", err)
+	//	return
```
golangcibot: File is not `goimports`-ed (from `goimports`)

id01:
> Sorry for the radio silence, too much going on!

It's good :)

> Rclone has ways of dealing with duplicate objects see `rclone dedupe` so there are ways of getting rid of them.

The problem is that if I allow duplicate files, the duplicate files will be unreliably created and created upon any update of any compressed objects.

I've just coded up two solutions - the size-in-filenames solution, and a multithreaded solution.

The multithreaded solution is on the master branch of my fork. It's something like THC Hydra does - as the latency from connecting/disconnecting to download the metadata for every Object slows us down greatly, I sent the download requests in parallel so that I can make greater use of my bandwidth. With this solution, I got up to 50KB/s, 10 times the bandwidth of before, and was able to list 98 entries in about 6 seconds:
```
time ./rclone ls press_over_dropbox:test_press
   524288 xzedx
   524288 xzedy
... snip ...
   524288 xzgte
   524288 xzgur

real	0m5.958s
user	0m0.469s
sys	0m0.120s
```

This is much faster than before, though still 8.9 times slower than listing on the base dropbox remote, and takes quite a bit of bandwidth, as it sends out many requests.

I've also implemented another version that stores file sizes in names; this will store the name and size of the metadata objects gotten from listing, and load the actual metadata object and the metadata itself on-demand. Unfortunately, this also means that we can't update an object, but instead have to reupload the object, not only if it switches between the compressed/uncompressed modes, but also if it changes in size. Not sure how this will affect performance. Fortunately, it's a lot faster and takes a lot less bandwidth than even the multithreaded version of directly loading metadata from files:
```
time ./rclone ls press_over_dropbox:test_press_2
   524288 xzedx
   524288 xzedy
... snip ...
   524288 xzgun
   524288 xzgur

real	0m0.847s
user	0m0.195s
sys	0m0.007s
```
This is on branch cache-compression of my fork.

ncw:
> > Rclone has ways of dealing with duplicate objects see `rclone dedupe` so there are ways of getting rid of them.
> 
> The problem is that if I allow duplicate files, the duplicate files will be unreliably created and created upon any update of any compressed objects.

In rclone terms, update means upload a new version.  Almost no cloud storage systems allow changing existing objects.
 
> I've just coded up two solutions - the size-in-filenames solution, and a multithreaded solution.
> 
> The multithreaded solution is on the master branch of my fork. It's something like THC Hydra does - as the latency from connecting/disconnecting to download the metadata for every Object slows us down greatly, I sent the download requests in parallel so that I can make greater use of my bandwidth. With this solution, I got up to 50KB/s, 10 times the bandwidth of before, and was able to list 98 entries in about 6 seconds:
> 
> ```
> time ./rclone ls press_over_dropbox:test_press
>    524288 xzedx
>    524288 xzedy
> ... snip ...
>    524288 xzgte
>    524288 xzgur
> 
> real	0m5.958s
> user	0m0.469s
> sys	0m0.120s
> ```
> 
> This is much faster than before, though still 8.9 times slower than listing on the base dropbox remote, and takes quite a bit of bandwidth, as it sends out many requests.

That is an improvement, but at a cost!

> I've also implemented another version that stores file sizes in names; this will store the name and size of the metadata objects gotten from listing, and load the actual metadata object and the metadata itself on-demand. Unfortunately, this also means that we can't update an object, but instead have to reupload the object, not only if it switches between the compressed/uncompressed modes, but also if it changes in size. Not sure how this will affect performance.

Re-uploading an object is fine for an update.  That is the expected behaviour for rclone.

> Fortunately, it's a lot faster and takes a lot less bandwidth than even the multithreaded version of directly loading metadata from files:
> 
> ```
> time ./rclone ls press_over_dropbox:test_press_2
>    524288 xzedx
>    524288 xzedy
> ... snip ...
>    524288 xzgun
>    524288 xzgur
> 
> real	0m0.847s
> user	0m0.195s
> sys	0m0.007s
> ```
> 
> This is on branch cache-compression of my fork.

Very nice!  That looks like the winner to me :-)

backend/press/press.go
```diff
@@ -0,0 +1,1369 @@
...
+		if err != nil {
+			return "", "", 0, errors.New("Could not decode size")
+		}
+	}
+	// Return everything
+	return name, extension, size, nil
+}
+
+// Processes a metadata filename. Returns the original filename.
+func processMetadataName(metaFileName string) (origFileName string, err error) {
```
golangcibot: `processMetadataName` is unused (from `deadcode`)

backend/press/press.go
```diff
@@ -0,0 +1,1369 @@
...
+// Converts an int64 to hex
+func int64ToHex(number int64) string {
+	intBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(intBytes, uint64(number))
+	return hex.EncodeToString(intBytes)
+}
+
+// Converts hex to int64
+func hexToInt64(hexNumber string) (int64, error) {
+	intBytes := make([]byte, 8)
```
golangcibot: ineffectual assignment to `intBytes` (from `ineffassign`)

backend/press/press.go
```diff
@@ -0,0 +1,1369 @@
...
+	}
+	nameWithSize := compressedFileName[:extensionPos]
+	extension = compressedFileName[extensionPos:]
+	// Separate the name with the size if this is a compressed file. Otherwise, just return nameWithSize (because it has no size appended)
+	var name string
+	var size int64
+	if extension == uncompressedFileExt {
+		name = nameWithSize
+		size = -2
+		err = nil
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

id01:
@ncw finished it up. Made the Update() function prefer a move-then-update over a reupload-then-remove, but default to reuploading, mainly for things like versioning on some remotes.

ncw:
> finished it up. 

That is great!

> Made the Update() function prefer a move-then-update over a reupload-then-remove, but default to reuploading, mainly for things like versioning on some remotes.

:-)

v1.48 is coming out at the weekend.  I could merge this for that.  I'd like to get it into the integration tests running against google drive and a bucket based backend (swift most likely).

I need to do a few more tests to make sure it works for me and review the code.

Something went wrong with your merge as github now says it is unmergeable!  Can you `git rebase master` fix any merge conflicts and push --force it up again?

id01:
> Something went wrong with your merge as github now says it is unmergeable! Can you `git rebase master` fix any merge conflicts and push --force it up again?

Strange; pulling from id01/master into ncw/master works for me without any conflicts.

ncw:
It is the rebase which is failing...

Can you rebase it on master (it might be easiest to cherry-pick the commit into a new branch based of master).

If you could split it into two commits one for the vendor stuff, eg

```
vendor: add dependencies for press backend

-  github.com/OneOfOne/xxhash
-  github.com/gabriel-vasile/mimetype
-  github.com/golang/snappy
-  github.com/id01/go-lz4
```

And one for the backend that would be perfect - thanks.

id01:
Ah, looks like it's having problems with the cherry picking of the env_test patch upstream. I'm going to have to push the updates to a second fork then manually copy over the changes to the original fork.

id01:
There we go; should be good.

ncw:
> There we go; should be good.

Can you split into two commits as per https://github.com/ncw/rclone/pull/3194#issuecomment-501677481 ?

I can do this if you are having trouble with it.

id01:
Ah, sorry; thought that was a suggestion of how to fix the rebase issues.

ncw:
I've been running some more tests and found some more failures :-(  I made a press remote against a swift backend (though I think any bucket based backend would do - minio for instance).

I made this small patch then

```
diff --git a/backend/press/press_test.go b/backend/press/press_test.go
index db4110ca7..f26a3fca7 100644
--- a/backend/press/press_test.go
+++ b/backend/press/press_test.go
@@ -7,6 +7,7 @@ import (
        "testing"
 
        _ "github.com/ncw/rclone/backend/local"
+       _ "github.com/ncw/rclone/backend/swift"
        "github.com/ncw/rclone/fstest"
        "github.com/ncw/rclone/fstest/fstests"
 )
```

    go test -v presstestswift:

Gives

```
            --- FAIL: TestIntegration/FsMkdir/FsPutFiles/FsCopy (0.17s)
                require.go:794: 
                    	Error Trace:	fstests.go:1005
                    	            	object not found
                    	            	github.com/ncw/rclone/fs.init.ializers
                    	            		/home/ncw/go/src/github.com/ncw/rclone/fs/fs.go:57
                    	            	runtime.main
                    	            		/opt/go/go1.12/src/runtime/proc.go:188
                    	            	runtime.goexit
                    	            		/opt/go/go1.12/src/runtime/asm_amd64.s:1337
                    	Test:       	TestIntegration/FsMkdir/FsPutFiles/FsCopy
                    	Messages:   	Error: object not found
```

There are other failures here: running this in `fs/operations`, `fs/backend`, `fs/sync`

    go test -v -remote presstestswift: -verbose -list-retries 1

Which I didn't have time to dig in to, but the feature reporting seems to be wrong

We should probably add some wrapped remotes other than local into the integration tests

```
 - backend:  "press"
   remote:   "TestPressLz4:"
   subdir:   false
   fastlist: true
 - backend:  "press"
   remote:   "TestPressGzip:"
   subdir:   false
   fastlist: true
 - backend:  "press"
   remote:   "TestPressXZ:"
   subdir:   false
   fastlist: true
```

So add a `TestPressDrive` and a `TestPressSwift` here too as representatives of extra tests that would be useful.

id01:
@ncw ah, yeah; I see; tested on B2 and found the problem. Seems like bucket-based remotes use Hash in the ObjectInfo along with Size and Name, so I needed to override that too.

Running operations and sync with other remotes uncovered a ton of bugs that didn't really affect the basic tests (mainly regarding duplicate files being created where they shouldn't be), but did affect certain remote and integration tests, especially with bucket-based remotes. After a bunch of patches, they began working better. Unfortunately, these patches made certain functions like Move() and Copy() quite a bit slower.

Also, I'm still having an evasive bug with bucket-based remotes. There's a bug that only shows up at random when running the full sync or operations test on b2 and causes the occasional random operation to fail. And running the failed tests individually always gives a pass. However, this may be a problem with the b2 backend, not press, as the raw b2 backend is failing the operation and sync tests as well.

ncw:
Note that I have just merged f78cd1e0434ee597fb8937ba3cda0d5fd58becfe which changes the Fs interface to thread it with context.  It took me 5 or 10 minutes to fixup an unmerged backend I have (google photos) with help from the compiler so it shouldn't take you too long hopefully.

> @ncw ah, yeah; I see; tested on B2 and found the problem. Seems like bucket-based remotes use Hash in the ObjectInfo along with Size and Name, so I needed to override that too.

OK

> Running operations and sync with other remotes uncovered a ton of bugs that didn't really affect the basic tests (mainly regarding duplicate files being created where they shouldn't be), but did affect certain remote and integration tests, especially with bucket-based remotes. After a bunch of patches, they began working better. Unfortunately, these patches made certain functions like Move() and Copy() quite a bit slower; testing is taking seconds per test now.

Seconds per test on the bucket based remotes doesn't sound too bad...
 
> Also, I'm still having an evasive bug with bucket-based remotes. There's a bug that only shows up at random when running the full sync or operations test on b2 and causes the occasional random operation to fail. And running the failed tests individually always gives a pass. However, this may be a problem with the b2 backend, not press, as the raw b2 backend is failing the operation and sync tests as well.

If you look at the integration tests: https://pub.rclone.org/integration-tests/current/ you'll see that the b2 tests usually pass after a retry or two.  This may be a problem in the b2 backend but it might be something timing dependent, or networking issues...

> Finally, due to the speed loss because of the amount of checking and moving that is needed to be done to prevent duplicate files from being created with many operations where they shouldn't be, the compression remote is becoming pretty slow. Perhaps storing the size in the filename isn't the best idea after all?

I see what you mean.  What do you think plan B is?

id01:
Sorry for not responding earlier; I'm currently in the process of shrinking my PC for college, and as such, it's very frequently non-functional.

If seconds per test isn't too bad, I suppose I'll just stick with this solution then.

Maybe it's due to networking issues on my side. My internet isn't great.

An alternate solution would be something similar to before, but with some way to cache large numbers of object sizes into one file per directory for listing purposes. However, this can have issues with race conditions.

ncw:
> Sorry for not responding earlier; I'm currently in the process of shrinking my PC for college, and as such, it's very frequently non-functional.

Me too - I've been super busy!

> If seconds per test isn't too bad, I suppose I'll just stick with this solution then.
> 
> Maybe it's due to networking issues on my side. My internet isn't great.
> 
> An alternate solution would be something similar to before, but with some way to cache large numbers of object sizes into one file per directory for listing purposes. However, this can have issues with race conditions.

That is a good idea but getting more complicated...

What do you think - should we merge this as-is?  Or do you want to do more to it?

I'm aware that I probably need to do another review.

We need to document the format of the meta files somewhere too.

id01:
I haven't merged the branch I made to master of my fork yet.

Also, I believe the first solution of storing filesizes in metadata files would be feasible if enough remotes allow a multi-download command, similar to when Drive allows you to download a zip file by selecting multiple files in the GUI. If this is possible, we could even have a setting to store metadata in a separate remote that supports multi-downloads

ncw:
> I haven't merged the branch I made to master of my fork yet.

OK

> Also, I believe the first solution of storing filesizes in metadata files would be feasible if enough remotes allow a multi-download command, similar to when Drive allows you to download a zip file by selecting multiple files in the GUI. If this is possible, we could even have a setting to store metadata in a separate remote that supports multi-downloads (or even just a faster one like SSH).

Not many remotes allow you to download multiple files at once - it doesn't fit well into what you can do with http unless you zip them.

thestigma:
I hope it's not inappropriate for me to comment here, but I just want to say that I hope there is still ongoing development on this branch. This (specifically the ability to join many small files into one for the sake of Rclone efficiency) is something I am very excited about and feel is a big limitation currently due to the inherent limitations of transferring small files via Rclone.

If I can be of any service in testing or something of that nature, just let me know and I will assist in any way I am able.

id01:
This branch isn't about joining small files into larger ones, but is more about reducing the size of large files. However, this branch would greatly benefit from some code that combines small files.

I was hoping on working on this feature this summer, but I've been busier than expected preparing for college and everything. I might get back to it later during the school year or during winter break.

ncw:
> This branch isn't about joining small files into larger ones, but is more about reducing the size of large files. However, this branch would greatly benefit from some code that combines small files.
> 
> I was hoping on working on this feature this summer, but I've been busier than expected preparing for college and everything. I might get back to it later during the school year or during winter break.

That would be great :-)

Skylion007:
Are there any plans to add support for zstd compression? It's on the Pareto frontier for fast decompression and as such would be pretty helpful to add.  There is a helpful wrapper for it in Go as well: https://github.com/DataDog/zstd

id01:
I haven't contributed for quite a while; I've just got into college and the school year has been very busy, and I believe I need the winter break to unwind. It would be nice if someone could pick up where I left off, as I'm probably going to only be able to contribute mid to late next year if the workload stays the same.

ncw:
@id01 thanks for getting this far - it is a great piece of work that could be picked up by someone else. @ivandeex this looks like a thing that might interest you?

ivandeex:
> @ivandeex this looks like a thing that might interest you?
Hi @ncw Up until March I won't have enough spare time for projects, even for planned improvements in my remotes. Sorry about that.

buengese:
Just fyi I had played around with this a bit. The version in my press branch is rebased and should pass all the tests but I think this might need some more work. I may or may not find the time to do this so if someone else is working on this he could also start from there.

PR #3202, opened 2019-05-16
fs: add --ignore-case-sync for forced case insensitivity - fixes #2773

### LINKED ISSUE #2773: Add support for forced case insensitivity
1.35

Preventing clone/sync of duplicate files, for which only the casing differs, regardless if the source/target file-systems having cases sensitive filesystems (e.g. most/all linux systems)

e.g. files “Application A.doc” and “application a.doc” are now treated as two different files, I hope to be able to set a flag so they are treated as one and the same.

Add a flag that can either be combined or superseded with/by "--ignore-existing" so a file will NOT be copied/synced when the existing filename is the same, even if the casing is different 
(e.g. --ignore-existing-case-insensitive)

Forum link: https://forum.rclone.org/t/copy-sync-how-to-ignore-casing-when-checking-for-duplicates/7715

garry415:
This change adds the ability to prevent the clone/sync of duplicate files for which only the casings differs, regardless if the source/target file-systems have case-sensitive filesystems (e.g. most/all linux systems)

e.g. files "Application A.doc" and "application a.doc" are now treated as the same file if the flag is set.

Issue link: https://github.com/ncw/rclone/issues/2773
Forum link: https://forum.rclone.org/t/copy-sync-how-to-ignore-casing-when-checking-for-duplicates/7715

#### Checklist

- [X ] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [X ] I have added tests for all changes in this PR if appropriate.
- [X ] I have added documentation for the changes if appropriate.
- [X ] All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [X ] I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -500,6 +500,12 @@ Do a trial run with no permanent changes.  Use this to see what rclone
 would do without actually doing it.  Useful when setting up the `sync`
 command which deletes files in the destination.
 
+### --ignore-case-sync ###
+
+Using this option will cause rclone to ignore the case of the files 
+when synchronizing so files will not be copied/synced when the
+existing filenames are the same, even if the casing is different.
+
 ### --ignore-checksum ###
 
 Normally rclone will check that the checksums of transferred files
--- a/fs/config.go
+++ b/fs/config.go
@@ -62,6 +62,7 @@ type ConfigInfo struct {
 	MaxDepth               int
 	IgnoreSize             bool
 	IgnoreChecksum         bool
+	IgnoreCaseSync         bool
 	NoTraverse             bool
 	NoUpdateModTime        bool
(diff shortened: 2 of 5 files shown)


ncw:
The code looks really good - thank you :-)

Can you have a go at writing a test?

I'd start with a test like this

https://github.com/ncw/rclone/blob/cff85f0b9551881be535a4847b56742451e6b476/fs/sync/sync_test.go#L1317

Create two files  with different cases one in local and one in remote and check the sync does what you expect with the flag set.

ncw:

I'll merge.

(Apologies for the slow turnaround!)

PR #3204, opened 2019-05-17
jottacloud: Add support for selecting device and mounpoint fixes #3069

### LINKED ISSUE #3069: Jottacloud: Add mountpoint "Backed up files"
rclone v1.46
- os/arch: freebsd/amd64
- go version: go1.11.5

As I've previously only used the Jottacloud desktop client to back up my files, all my files are now in the section "Backed up files", under a device name which gets automatically created from the computer name running the client.
I do not use "Sync" as you can only use it for ONE folder and you also can't use file/folder exclusions.
I also do not use the "Archive" since the desktop client cannot directly upload to Archive anyway, files can only be moved to "Archive" from "Backed up files" manually through the Jottacloud website which renders the feature pretty much useless. You also can't move them back to "Backed up files" once done.

As I now want to use rclone instead, and rclone for some reason doesn't have that section available under mount points, I'd end up uploading everything again and then have duplicates as rclone doesn't checksum against the files already there under "Backed up files".
That's just not very feasible with large backups.

Add "Backed up files" to the mount point selection in rclone config for Jottacloud remotes.
Also adding the ability to add devices (computer name) would probably solve the bug about not being to able to use rclone from two rclone sessions at the same time since they would have separate device names.

Jottaclouds own command line client has both these features but unfortunately it really blows in every other way.

buengese:
I finally found the time to implement support for selecting custom devices and mountpoints. This is done with an additional step during config. Old config files stay compatible and will automatically use the default Jotta/Archive which was the only option in the past. This should help everyone who has used the official client in the past. I'd ask everyone who has waited for this feature to test it because I only did some quick rough testing.

Should be mostly ready and working. I'll cleanup the code some more and update the docs this weekend.

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -41,7 +41,7 @@ const (
 	maxSleep                    = 2 * time.Second
 	decayConstant               = 2 // bigger for slower decay, exponential
 	defaultDevice               = "Jotta"
-	defaultMountpoint           = "Sync" // nolint
+	defaultMountpoint           = "Archive"
 	rootURL                     = "https://www.jottacloud.com/jfs/"
 	apiURL                      = "https://api.jottacloud.com/files/v1/"
 	baseURL                     = "https://www.jottacloud.com/"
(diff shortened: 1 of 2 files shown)


ncw:
Thanks for doing this.

I pushed this to a branch in the master repo so it makes a beta and I'll post a link for testers in the issue

https://beta.rclone.org/branch/v1.47.0-075-g854d083d-jotta_mountpoints-beta/ (uploaded in 15-30 mins)

buengese:
Because no one seems to have really noticed this and it really shouldnt break anything i would be in favor of merging this as it is right now.

ncw:
OK I will merge now  - thank you very much.

PR #3220, opened 2019-05-22
Distinguish between objects and directories when synchronizing filesystems

forgems:
#### This change allows to synchronize between two bucket based cloud storages that use key/value storage

This change adds sorting by object type to DirEntries sorting. Additionally it distinguishes between objects and directories in matchListings function. By adding this directories with the same name as object name are  distinguished from objects, and can be synchronized. This allows to synchronize files between cloud storages that are key/value stores. For example when the source has the following structure (taken with rclone ls -R --files-only)
```
index
index/test
index/test2
```
previously it couldn't synchronize the `index` file because it had the same name as directory `index/`. After this change the `index` file and `index` directory are distinguished and are not reported as duplicates. This could be a problem when you are synchronizing the files from Key/Value storage to local filesystem, because it doesn't support duplicate directory entries, and right now it reports the error about it. 

The relevant discussion is here
https://forum.rclone.org/t/problem-when-cloning-one-cloud-storage-to-another-cloud-storage/10146/5

#### Checklist

### DIFF
--- a/fs/direntries_test.go
+++ b/fs/direntries_test.go
@@ -1,0 +1,97 @@
+package fs
+
+import (
+	"fmt"
+	"io"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/ncw/rclone/fs/hash"
+	"github.com/stretchr/testify/assert"
+)
+
+type FakeObject string
+
+func (o FakeObject) ModTime() time.Time {
+	return time.Now()
+}
+
+func (o FakeObject) String() string {
+	return fmt.Sprintf("FakeObject(%s)", string(o))
+}
+
+func (o FakeObject) Remote() string {
+	return string(o)
+}
+
+func (o FakeObject) Size() int64 {
+	return 0
+}
+
+func (o FakeObject) SetModTime(t time.Time) error {
+	return nil
+}
+
+func (o FakeObject) Open(options ...OpenOption) (io.ReadCloser, error) {
+	return nil, nil
+}
+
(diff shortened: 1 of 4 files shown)


fs/direntries_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"io"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/ncw/rclone/fs/hash"
+	"github.com/stretchr/testify/assert"
+)
+
+type FakeObject string
```
ncw: You should be able to use `fstest/mockobject` here I think rather than build your own FakeObject

ncw: If it is giving you a circular import then change the package to `fs_test`

ncw:
Can you have a go at fixing up the travis output?  This seems to have broken the `TestSyncUTFNorm` test.

fs/direntries_test.go
forgems: Yes. I've got circular imports when I tried to use mockobject and mockdir. 
Thanks I'll change it.

forgems: Fixed

forgems:
Hi @ncw.  I've fixed the issue with `TestSyncUTFNorm`. Unfortunately, it resulted in some code duplication in `matchListings` (I couldn't reuse `CompareDirEntries` function) because it compares transformed names from `matchEntry` instead of remote names.

ncw:
I think this looks great now - I'll merge now.

Apologies for the delay!

PR #3225, opened 2019-05-24
Fixes 3224

### LINKED ISSUE #3224: sync fails to S3 Glacier and Glacier Deep Archive tiers
Creating an issue to track this bug.
https://forum.rclone.org/t/suspected-bug-in-s3-or-compatible-sync-logic-to-glacier/10187

When syncing from a S3 bucket to a S3 bucket if rclone detects that the source object timestamp has changed and the file is the same size and etag is matching in source and destination it attempts to update the metadata that stores the original modtime.
If the S3 object to be updated (copied) is in Glacier or Glacier Deep Archive tier then updating the modtime updates fails since you can't do a server side copy of an object that is in Glacier tier.
The error message is:
Failed to set modification time: InvalidObjectState: Operation is not valid for the source object's storage class
The behavior should be updated to upload the object with the new modtime rather than attempt to do the server side copy/modtime update.

pharveybattelle:
Fixes #3224
https://forum.rclone.org/t/suspected-bug-in-s3-or-compatible-sync-logic-to-glacier/10187
#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -47,6 +47,7 @@ import (
 	"github.com/ncw/rclone/lib/rest"
 	"github.com/ncw/swift"
 	"github.com/pkg/errors"
+	"github.com/ncw/rclone/fs"
 )
 
 // Register with Fs
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -47,7 +47,6 @@ import (
 	"github.com/ncw/rclone/lib/rest"
 	"github.com/ncw/swift"
 	"github.com/pkg/errors"
-	"github.com/ncw/rclone/fs"
 )
 
 // Register with Fs
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1736,6 +1736,9 @@ func (o *Object) SetModTime(modTime time.Time) error {
 	if o.fs.opt.SSEKMSKeyID != "" {
 		req.SSEKMSKeyId = &o.fs.opt.SSEKMSKeyID
 	}
+	if o.fs.opt.StorageClass == "GLACIER" || o.fs.opt.StorageClass == "DEEP_ARCHIVE" {
+		return fs.ErrorCantSetModTime 
+  	}    
 	if o.fs.opt.StorageClass != "" {
 		req.StorageClass = &o.fs.opt.StorageClass
 	}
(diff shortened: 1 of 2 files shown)


backend/s3/s3.go
```diff
@@ -1736,6 +1736,9 @@ func (o *Object) SetModTime(modTime time.Time) error {
 	if o.fs.opt.SSEKMSKeyID != "" {
 		req.SSEKMSKeyId = &o.fs.opt.SSEKMSKeyID
 	}
+	if o.fs.opt.StorageClass == "GLACIER" || o.fs.opt.StorageClass == "DEEP_ARCHIVE" {
+		return fs.ErrorCantSetModTime
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/s3/s3.go
```diff
@@ -47,6 +47,7 @@ import (
 	"github.com/ncw/rclone/lib/rest"
 	"github.com/ncw/swift"
 	"github.com/pkg/errors"
+	"github.com/ncw/rclone/fs"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ncw:
That looks great!

Can you add a small bit [to the docs](https://github.com/ncw/rclone/blob/master/docs/content/s3.md#modified-time) 

Maybe something like

> If the modification time needs to be updated, then rclone will use the COPY command to update the metadata in-place.  Note this isn't possible with GLACIER storage, so in this case rclone will have to upload the file again.

pharveybattelle:
> That looks great!
> 
> Can you add a small bit [to the docs](https://github.com/ncw/rclone/blob/master/docs/content/s3.md#modified-time)
> 
> Maybe something like
> 
> > If the modification time needs to be updated, then rclone will use the COPY command to update the metadata in-place.  Note this isn't possible with GLACIER storage, so in this case rclone will have to upload the file again.

done

ncw:
I've merged that now - apologies for the delay.

Thank you very much for that - much appreciated :-)

pharveybattelle:
> I've merged that now - apologies for the delay.
> 
> Thank you very much for that - much appreciated :-)

Thanks ncw for walking me though the process as a noob

ncw:
> Thanks ncw for walking me though the process as a noob

My pleasure!

PR #3226, opened 2019-05-25
drive: add --drive-size-as-quota to show storage quota usage for file…

### LINKED ISSUE #3135: gDrive rclone ls with display quotaBytesTotal instead of fileSize
At now `rclone ls` command on Google Drive displays only real file size (`fileSize` argument from Google API).

Is this possible to switch `rclone` display number value from `fileSize` to `quotaBytesTotal` (or add `quotaBytesTotal` as second number in output), for understanding file quota usage in my gDrive?

If not possible, can you accept this as feature request, so implement something like `rclone ls --size=quota`?

Same feature also will be very useful for other storage providers too.

garrmcnu:

Add --drive-size-as-quota to show storage quota usage for file size.

Fixes #3135

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -368,6 +368,14 @@ will download it anyway.`,
 			Help:     "Keep new head revision of each file forever.",
 			Advanced: true,
 		}, {
+			Name:    "size_as_quota",
+			Default: false,
+			Help: `Show storage quota usage for file size.
+
+The storage used by a file is the size of the current version plus any
+older versions that have been set to keep forever.`,
+			Advanced: true,
+		}, {
 			Name:     "v2_download_min_size",
 			Default:  fs.SizeSuffix(-1),
 			Help:     "If Object's are greater, use drive v2 API to download.",
@@ -424,6 +432,7 @@ type Options struct {
 	ChunkSize                 fs.SizeSuffix `config:"chunk_size"`
 	AcknowledgeAbuse          bool          `config:"acknowledge_abuse"`
 	KeepRevisionForever       bool          `config:"keep_revision_forever"`
+	SizeAsQuota               bool          `config:"size_as_quota"`
 	V2DownloadMinSize         fs.SizeSuffix `config:"v2_download_min_size"`
 	PacerMinSleep             fs.Duration   `config:"pacer_min_sleep"`
 	PacerBurst                int           `config:"pacer_burst"`
@@ -631,6 +640,9 @@ func (f *Fs) list(dirIDs []string, title string, directoriesOnly, filesOnly, inc
 	}
 	if f.opt.SkipChecksumGphotos {
 		fields += ",spaces"
+	}
+	if f.opt.SizeAsQuota {
+		fields += ",quotaBytesUsed"
 	}
 
 	fields = fmt.Sprintf("files(%s),nextPageToken", fields)
@@ -1007,13 +1019,17 @@ func (f *Fs) newBaseObject(remote string, info *drive.File) baseObject {
 	if f.opt.UseCreatedDate {
 		modifiedDate = info.CreatedTime
 	}
+	size := info.Size
+	if f.opt.SizeAsQuota {
+		size = info.QuotaBytesUsed
+	}
 	return baseObject{
 		fs:           f,
 		remote:       remote,
 		id:           info.Id,
 		modifiedDate: modifiedDate,
 		mimeType:     info.MimeType,
-		bytes:        info.Size,
+		bytes:        size,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


MurzNN:
Thanks for the patch, I try to compile `rclone` with it, but can't see `--drive-size-as-quota` argument in  `./rclone help flags drive` output, also adding it to command line shows the error:
```
$ ./rclone size gdrive-murznn:"/temp" --drive-size-as-quota
Error: unknown flag: --drive-size-as-quota
Usage:
  rclone size remote:path [flags]

Flags:
  -h, --help   help for size
      --json   format output as JSON

2019/05/26 19:39:07 Fatal error: unknown flag: --drive-size-as-quota
```
But in sources I see all your changes, applied to this merge (eg `size_as_quota` in `backend/drive/drive.go`).
Maybe I need to add some options to `go build` for apply your patch?

MurzNN:
Also adding `size_as_quota = true` to `.config/rclone/rclone.conf` don't change `rclone size` results on gdrive folders, containing google photos files, that use zero quota.

MurzNN:
Ow, sorry, seems `go build` compile `rclone` binary from other folder (default repository), I'll try to recompile with your patches.

garrmcnu:
Thanks for taking the time to test, it should appear in the `help flags` output, but yes does need to be built from the default repository location. Do you see the flag with an updated build?

MurzNN:
I solve my problem with compiling `rclone` from git, and test your PR - it works well with `size` and `ls` commands, thanks!

MurzNN:
After some other tests I detect that sometimes `size` command returns `-1` and `-2` as total size for shared folders, instead of zero. This is because Google Docs files returns `-1` as size instead of zero. Can you add workaround and force `0` size for Google Docs files?

docs/content/drive.md
```diff
@@ -804,6 +804,18 @@ Keep new head revision of each file forever.
 - Type:        bool
 - Default:     false
 
+#### --drive-size-as-quota
+
+Show storage quota usage for file size.
+
+The storage used by a file is the size of the current version plus any
```
ncw: This is an auto generated file, and the help here needs to go into the `Help:     "Show storage quota usage for file size.",` line in drive.go

MurzNN:
Issue with `-1` size is reproduced also without `--drive-size-as-quota` flag in `ls` and `size` commands, so this is more general issue, and maybe solution better to do in separate PR.

docs/content/drive.md
garrmcnu: Will update that, thanks!

garrmcnu:
Thanks for the feedback, yes the -1 isn't related to quota flag, looks like the size is set to -1 for all files that are Google documents, set by https://github.com/ncw/rclone/blob/e7c20e0bce14ae62957febbb32300c354cc9b1da/backend/drive/drive.go#L1059

garrmcnu:
Have updated the help message (I had missed that file is auto generated).

On the -1 size for Google documents, setting `baseObject.bytes` to 0 should fix that?
Happy to add that change, or is it cleaner to put it in a separate PR?

garrmcnu:
Not sure why the OS X build failed? looks like an inactivity timeout?

ncw:
> Not sure why the OS X build failed? looks like an inactivity timeout?

I gave it a kick...

> On the -1 size for Google documents, setting `baseObject.bytes` to 0 should fix that?
> Happy to add that change, or is it cleaner to put it in a separate PR?

I think I'd prefer it in this one as it is part of the same change.  It should be quick though I think!

garrmcnu:
> > Not sure why the OS X build failed? looks like an inactivity timeout?
> 
> I gave it a kick...
> 
Great, tests are passing now, thanks!

> > On the -1 size for Google documents, setting `baseObject.bytes` to 0 should fix that?
> > Happy to add that change, or is it cleaner to put it in a separate PR?
> 
> I think I'd prefer it in this one as it is part of the same change. It should be quick though I think!

sounds good......however seems there's always a catch :-)
setting the size to 0 caused a test to fail, `TestIntegration/FsMkdir/FsPutFiles/Internal/DocumentImport` failed with a "sizes differ" error, sizeDiffers() ignores negative sizes (so 0 means 0, but -1 means undefined/unknown).
https://github.com/ncw/rclone/blob/e7c20e0bce14ae62957febbb32300c354cc9b1da/fs/operations/operations.go#L103-L108

One option would be to leave the reported size as -1, so the `ls` command would still show this and file can still be identified as a Google document, but for the `size` command should only count non-negative sizes towards the total? This should mean the reported total size would match the quota used.

ncw:
Apologies for the delay in responding

> One option would be to leave the reported size as -1, so the `ls` command would still show this and file can still be identified as a Google document, but for the `size` command should only count non-negative sizes towards the total? This should mean the reported total size would match the quota used.

Yes that is a good idea and I see you've added a commit for that.

I'll merge those now  - thank you very much :-)

Issue #3050, opened 2019-03-14
Serve DLNA not wokrking on Oculus Go and LG TV (2012 model)


tabnk:
#### What is the problem you are having with rclone?

DLNA not working  on Oculus Go and LG TV (2012 model). 

On oculus go, it shown rclone (DLNA) but no media was found.

On LG TV, it shown rclone (DLNA) but TV auto restart when selected.

#### What is your rclone version (output from `rclone version`)

Ver 1.46.

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Windows 10, 64-bit

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive/Team Drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone serve dlna remote:Backup\Media

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

019/03/15 02:37:43 DEBUG : Adding path "vfs/forget" to remote control registry
2019/03/15 02:37:43 DEBUG : Adding path "vfs/refresh" to remote control registry
2019/03/15 02:37:43 DEBUG : Adding path "vfs/poll-interval" to remote control registry
2019/03/15 02:37:43 NOTICE: Google drive root 'Backup/Media': Serving HTTP on [::]:7879
2019/03/15 02:37:43 Started SSDP on Ethernet

ncw:
The situation has been improved by #3045 but it isn't 100% - is that right @tabnk and @sretlawd ?

tabnk:
@ncw Yes, Oculus Go work and no crash.  Only issue left is my LG TV (PRE-WEBOS) which doesn’t work and cause my TV to restart itself.

sretlawd:
I did do a small implementation of the ConnectionManager interface, just doing what Emby does, the other day @sretlawd/rclone@07dfb81faf95b10c2cb707f55a6ed56bdedeadbb.  I don't expect it to make much of a difference, but there's always the possibility.  There's a lot of finicky clients out there.

@tabnk can you get a packet trace of the requests from your TV?  I got my devices working by running wireshark, capture tcp 7879 and udp 1900, and seeing what it's trying to do.

tabnk:
@sretlawd I’m will try wireshark now.

tabnk:
@sretlawd  

Frame 41: 60 bytes on wire (480 bits), 60 bytes captured (480 bits) on interface 0
[...]

PR #3227, opened 2019-05-26
Dlna compat logging

sretlawd:
Improved compatibility with various devices and greatly improved logging for diagnosis purposes.

Did some baseline refactoring, which I went back and tried to place in separate commits.  Static assets are included via vfsgen, the root descriptor XML has been changed to a text/template, and the http servemux has been reworked a little bit to clean it up and add logging and error handling.

I added bits of XML and headers and such based on what I've seen in other implementations (minidlna, emby, and simpleDLNA.)  I can now pause, fast forward, rewind, seek, etc., when watching videos on my Samsung TV, which is a big improvement.  I suspect some of the other changes will also improve device compatibility - the <iconList></iconList> without any <icons> that was in the descriptor seems like the kind of thing that could crash a poorly written client, for example.

Some discussion in #2648 and #3050.

#### Checklist


ncw:
That is really nice - thank you :-)

I like the use of vfsgen

Some things which occurred to me while reading the code
- the templates could be stored with vfsgen too to save a bit of binary space
- it looks like you've got to the point where you could integrate `cmd/httplib` which would save you a bit of code and make the serve experience more consistent,eg 

https://github.com/ncw/rclone/blob/da3b685cd847deae498dd19603a4f62f0191e4f6/cmd/serve/http/http.go#L64

I'll merge this now - thanks :-)

sretlawd:
Ah, I was still trying to think of what I could add as far as tests.  :-)  Have been sitting on it for a couple months and figured I should at least get it out there for testing.  Good idea on moving the template.

I had the same thought on httplib - right now media files are being served using the query string, at /res?path=path/to/file.ext, and I think it makes sense to just use /res/path/to/file.ext and route that to httplib.  But maybe there’s a not so obvious reason for using the query string (caching behavior?), so figured I should look at other implementations before changing that, just to be sure.

ncw:
> Ah, I was still trying to think of what I could add as far as tests. :-)

;-) Well if you can think of some then I'll merge them!

> Have been sitting on it for a couple months and figured I should at least get it out there for testing. 

It might be worth dropping a note in [the rclone forum](https://forum.rclone.org/) if you want more testers.

> Good idea on moving the template. I had the same thought on httplib - right now media files are being served using the query string, at /res?path=path/to/file.ext, and I think it makes sense to just use /res/path/to/file.ext and route that to httplib. But maybe there’s a not so obvious reason for using the query string (caching behavior?), so figured I should look at other implementations before changing that, just to be sure.

Sure!

PR #3245, opened 2019-06-03
Use rclone.conf from rclone executable directory if already existing

albertony:

To run rclone in "portable" mode, as in "portable app", by putting rclone.exe together with rclone.conf in a folder on a memory stick. Rclone will pick up the local config file without having to always remember to specify the --config command line argument, and not having to resort to launcher script like described in [forum](https://forum.rclone.org/t/how-to-change-the-path-to-the-config-file-rclone-cong-permanently/10324/5).

The idea came from [this](https://forum.rclone.org/t/how-to-change-the-path-to-the-config-file-rclone-cong-permanently/10324/5) forum post, although it does not describe the solution itself. 

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -452,7 +452,11 @@ Specify the location of the rclone config file.
 Normally the config file is in your home directory as a file called
 `.config/rclone/rclone.conf` (or `.rclone.conf` if created with an
 older version). If `$XDG_CONFIG_HOME` is set it will be at
-`$XDG_CONFIG_HOME/rclone/rclone.conf`
+`$XDG_CONFIG_HOME/rclone/rclone.conf`.
+
+If there is a file `rclone.conf` in the same directory as the rclone
+executable it will be preferred. This file must be created manually
+for Rclone to use it, it will never be created automatically.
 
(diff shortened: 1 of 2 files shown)


albertony:
> This is a nice idea :-)

There are a few other applications that does similar things, although with variations (e.g. ILSpy checks for file ILSpy.xml to store settings in, VSCode checks for folder "data" to store settings, cache, extensions etc in, Notepad++ checks for dummy file "doLocalConf.xml" to indicate it should store settings, plugins etc in various hard coded locations beneath the application folder). Command line argument is "required", but to also be able to run it "portable" without having to remember to always use the same argument, or to resort to launcher script, is nice.

> I'm a little concerned it might interfere with the already complicated finding the config file code, but it looks OK to me.

Sure, I can agree with that. I think it should be safe though (and it is a smaller change than my previous pr in the same code, #2515 :)

> I've given the CI another kick to see if we can get it to pass!

Didn't seem to help, unfortunately...

ncw:
Finally!  I'm getting closer to figuring out why the CI sometimes fails but not there yet! (#3154 )

I'll merge this now - thank you very much for this.

PR #3249, opened 2019-06-07
cmd: add support for private repositories in `serve restic` - fixes #3247

### LINKED ISSUE #3247: serve restic: Support for private repositories in restic http server functionality

```
rclone v1.47.0
- os/arch: linux/amd64
- go version: go1.11.6
```

Use `rclone serve restic` as restic backend and limit users to their own repositories.

It would be great if `rclone serve restic` supported the `--private-repos` flag like the original rest server from restic does:
```
      --private-repos       users can only access their private repo
```
This would limit users to a subpath of `/<username>/`

apollo13:
This is WIP to make it easier to discuss changes.

### DIFF
--- a/cmd/serve/httplib/httplib.go
+++ b/cmd/serve/httplib/httplib.go
@@ -2,6 +2,7 @@
 package httplib
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
@@ -114,6 +115,11 @@ type Server struct {
 	HTMLTemplate    *template.Template // HTML template for web interface
 }
 
+type contextUserType struct{}
+
+// ContextUserKey is a simple context key
+var ContextUserKey = &contextUserType{}
+
 // singleUserProvider provides the encrypted password for a single user
 func (s *Server) singleUserProvider(user, realm string) string {
 	if user == s.Opt.BasicUser {
@@ -167,6 +173,7 @@ func NewServer(handler http.Handler, opt *Options) *Server {
 				}
 				authenticator.RequireAuth(w, r)
 			} else {
+				r = r.WithContext(context.WithValue(r.Context(), ContextUserKey, username))
 				oldHandler.ServeHTTP(w, r)
 			}
 		})
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -29,14 +29,16 @@ import (
 )
 
 var (
-	stdio      bool
-	appendOnly bool
+	stdio        bool
+	appendOnly   bool
+	privateRepos bool
 )
 
 func init() {
 	httpflags.AddFlags(Command.Flags())
 	Command.Flags().BoolVar(&stdio, "stdio", false, "run an HTTP2 server on stdin/stdout")
 	Command.Flags().BoolVar(&appendOnly, "append-only", false, "disallow deletion of repository data")
+	Command.Flags().BoolVar(&privateRepos, "private-repos", false, "users can only access their private repo")
 }
 
 // Command definition for cobra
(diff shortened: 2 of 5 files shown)


ncw:
That is looking very good.

What more do you want to add to it?

apollo13:
Tests I guess, just not sure how to write them yet 😃

ncw:
Check out the these tests for some ideas https://github.com/ncw/rclone/blob/master/cmd/serve/restic/restic_appendonly_test.go

apollo13:
@ncw I have added tests as good as I could. One thing that I noticed though: Passing the username ala `r.Header.Set("Remote-User", username)` is pretty much dangerous since a user could probably override by manually sending this header. I do not see how the current code would allow for this, but it seems massively dangerous! http headers are not to be trusted after all :)

apollo13:
I mean technically we are only setting it after we verified the user and only then ever call the old handler, but it still feels wrong. Is there nothing else we can use to pass a "context" on or so?

apollo13:
I have pushed a follow up commit which uses go contexts to pass the information along, this should be 100% safe

apollo13:
One thing to note: My tests only test the restic handler, I couldn't figure out how to get access to the httplib handler to test the whole basic auth flow.

apollo13:
And the following change:
```
diff --git a/cmd/serve/restic/restic.go b/cmd/serve/restic/restic.go
index c5543fcc2..28c560947 100644
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -215,10 +215,20 @@ func (s *server) handler(w http.ResponseWriter, r *http.Request) {
        remote := makeRemote(path)
        fs.Debugf(s.f, "%s %s", r.Method, path)
 
-       v := r.Context().Value(httplib.ContextUserKey)
-       if privateRepos && (v == nil || !strings.HasPrefix(path, "/"+v.(string)+"/")) {
-               http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
-               return
+       if privateRepos {
+               // Check against the cert CN if client cert auth is enabled
+               if s.Opt.ClientCA != "" {
+                       if !strings.HasPrefix(path, "/"+r.TLS.PeerCertificates[0].Subject.CommonName+"/") {
+                               http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
+                               return
+                       }
+               } else { // Otherwise check the username from basic auth
+                       v := r.Context().Value(httplib.ContextUserKey)
+                       if v == nil || !strings.HasPrefix(path, "/"+v.(string)+"/") {
+                               http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
+                               return
+                       }
+               }
        }
 
        // Dispatch on path then method
```
would most likely allow validation against the client cert's CN -- but I did not test that.

ncw:
> @ncw I have added tests as good as I could. One thing that I noticed though: Passing the username ala `r.Header.Set("Remote-User", username)` is pretty much dangerous since a user could probably override by manually sending this header. I do not see how the current code would allow for this, but it seems massively dangerous! http headers are not to be trusted after all :)

We could make sure the header is "" when auth fails... Point taken however...

> I have pushed a follow up commit which uses go contexts to pass the information along, this should be 100% safe

That is probably the proper way of doing it.  I did consider that when writing my original advice to set "Remote-User" however I thought it would be a lot more complicated - I didn't know about the `Request.WithContext` method!


> One thing to note: My tests only test the restic handler, I couldn't figure out how to get access to the httplib handler to test the whole basic auth flow.

That should probably be in the (non existent ;-) httplib tests anyway!

> If you are okay with my latest changes I'll squash the commits.

Yes I think they look great.  If you are happy with them then I'll merge.


> And the following change:
> ...
> would most likely allow validation against the client cert's CN -- but I did not test that.

I wonder whether it might be better to set the Context key in httplib than check it here?  That would save code duplication.

Shall we leave that to a separate PR?  Or do you want to get it in this one?

apollo13:
>  I didn't know about the `Request.WithContext` method!

me neither, everything is new for me there :)

> Shall we leave that to a separate PR? Or do you want to get it in this one?

Separate since it might require some thinking, ie what happens if you gave client ca and basic auth (probably stupid, but well…). I squashed and the commit should be ready to go.

ncw:
> Separate since it might require some thinking, ie what happens if you gave client ca and basic auth (probably stupid, but well…).

Indeed, possible but unlikely? I think as long as we document which one takes precedence then it would be fine.

> I squashed and the commit should be ready to go.

Thank you for this excellent piece of work - much appreciated :-)

Will merge now.

PR #3256, opened 2019-06-12
Combine HTTP and WebDAV Serves

### LINKED ISSUE #2824: allow webdav to also serve http methods.
reference: https://forum.rclone.org/t/webdav-serve-http-pages-also/7878/3

Enhancement to allow webdav to optionally also serve GET/HEAD/PUT requests like a normal http server essentially combining the two functions into one so that there is no need to run separate ports to have both services running. This is typical with nginx implementations of davs and would be helpful.

gary-kim:
This is still very messy and needs a lot of testing.
I'll need to go back to clean things up, finish up implementing everything, add tests, and add documentation. Targeting the v1.49.0 release.
- [x] Complete implementation
- [x] Cleanup
- [x] Add Tests
- [x] Add documentation

[Travis Build](https://travis-ci.com/gary-kim/rclone/builds/115400252)

Fixes #2824 

Combine methods for WebDAV and HTTP serves.

#2824 

#### Checklist


cmd/serve/webdav/webdav.go
```diff
@@ -109,17 +113,25 @@ func newWebDAV(f fs.Fs, opt *httplib.Options) *WebDAV {
...
-
-	w.Server = httplib.NewServer(handler, opt)
+	w.webdavhandler = webdavHandler
+	w.httpserver = rhttp.NewServer(f, opt)
+	w.Server = httplib.NewServer(http.HandlerFunc(w.handler), opt)
 	return w
 }
 
+func (w *WebDAV) handler(rw http.ResponseWriter, r *http.Request) {
+	if (r.Method == "GET" || r.Method == "HEAD") && strings.HasSuffix(r.URL.Path, "/") {
```
ncw: So this just serves the directories?

Is that enough to be useful?

gary-kim: > So this just serves the directories?

Serving files is already handled by the WebDAV handler so I think the http handler only has to directories.

gary-kim:
> Do you think this should be unconditionally enabled in the webdav backend?

While I can't see much of a reason to disable it, just in case, I'll add a flag for people who would like to disable it.


> If you wanted tests you could import the http tests and run them here?

Sounds like a good idea. I'll do that.

ncw:
> > Do you think this should be unconditionally enabled in the webdav backend?
> 
> While I can't see much of a reason to disable it, just in case, I'll add a flag for people who would like to disable it.

It doens't give you any more access to anything, so I think your plan of a disable flag is perfect.

> > If you wanted tests you could import the http tests and run them here?
> 
> Sounds like a good idea. I'll do that.

:-)

cmd/serve/webdav/webdav_test.go
```diff
@@ -17,6 +17,17 @@ import (
 	"github.com/ncw/rclone/fstest"
 	"github.com/stretchr/testify/assert"
 	"golang.org/x/net/webdav"
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ncw:
I was hoping you could import the http tests in go `import` sense rather than duplicate the tests?  Is that possible?

gary-kim:
> I was hoping you could import the http tests in go `import` sense rather than duplicate the tests? Is that possible?

I'm not sure how to import a test function from another testing file. 

If it can be imported, the `checkGolden` and `TestGet` functions would not have to be duplicated.

ncw:
> > I was hoping you could import the http tests in go `import` sense rather than duplicate the tests? Is that possible?
> 
> I'm not sure how to import a test function from another testing file.

I think you have to factor it into another package.

> If it can be imported, the `checkGolden` and `TestGet` functions would not have to be duplicated.
> The golden files would still need to be duplicated because the exact responses of the webdav and http serves are different.

OK let's leave it as is for the moment.

I'll merge this - thank you very much :-)

Issue #3219, opened 2019-05-22
--backup-dir doesn't work for single file copy/move


Cnly:
#### What is the problem you are having with rclone?

`rclone.1` [says](https://github.com/ncw/rclone/blob/a0d4fdb2fa34c7f6ac5a86021d16403713499aaf/rclone.1#L4547) `--backup-dir` works for "sync, copy or move", but in fact it only works for `sync`. From the code it seems this option is only implemented for `sync`, and if that is the case the doc should be refined.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.47.0-074-g81fad0f0-beta
- os/arch: linux/amd64
- go version: go1.12.5
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

`Debian GNU/Linux 10 (buster), x86_64`

####  Which cloud storage system are you using? (eg Google Drive)

Any backends should be the same.

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`rclone -vv move myfile remote:afolder --backup-dir remote:backupdir`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
2019/05/22 16:21:52 DEBUG : rclone: Version "v1.47.0-074-g81fad0f0-beta" starting with parameters ["rclone" "-vv" "move" "myfile" "remote:afolder" "--backup-dir" "remote:backupdir"]
2019/05/22 16:21:52 DEBUG : Using config file from "/home/me/.config/rclone/rclone.conf"
2019/05/22 16:21:56 DEBUG : myfile: Sizes differ (src 30217 vs dst 3218)
2019/05/22 16:21:56 DEBUG : myfile: Starting multipart upload
2019/05/22 16:21:56 DEBUG : myfile: Uploading segment 0/30217 size 30217
2019/05/22 16:21:57 INFO  : myfile: Copied (replaced existing)
2019/05/22 16:21:57 INFO  : myfile: Deleted
2019/05/22 16:21:57 INFO  :
Transferred:       29.509k / 29.509 kBytes, 100%, 5.984 kBytes/s, ETA 0s
Errors:                 0
Checks:                 2 / 2, 100%
Transferred:            1 / 1, 100%
Elapsed time:        4.9s

2019/05/22 16:21:57 DEBUG : 9 go routines active
[...]

PR #3266, opened 2019-06-17
[WIP] Implement --suffix without --backup-dir for current dir

### LINKED ISSUE #2801: Use same or relative path for backups when files will be modified/deleted: --suffix without --backup-dir
I think it would be awesome if `rclone` could make backups in the same folder as the original file and/or a folder relative to the same folder.

For example:

- backup `remote:drive:/path/to/file.txt` to `remote:drive:/path/to/file.txt.20181124`
- backup `remote:drive:/path/to/file.txt` to `remote:drive:/path/to/backups/file.txt.20181124`

I was hoping to add this enhancement to the code myself but I cannot figure out the code. I know, at minimum, lines 147-148 in `rclone/fs/config/configflags/configflags.go` would need to be modified to allow `--suffix` without `--backup-dir`. I can tell some other files need to be updated like `sync.go` but can't make sense of `rclone`'s code structure to know everywhere that would need to be updated. If someone is willing to point me in the right direction I would add and test the enhancement myself then submit a pull request.

yparitcher:

allow --suffix by itself to add suffix and backup files in current directory

discussed in and Fixes #2801
#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -880,11 +880,23 @@ The default is `bytes`.
 
 ### --suffix=SUFFIX ###
 
-This is for use with `--backup-dir` only.  If this isn't set then
-`--backup-dir` will move files with their original name.  If it is set
-then the files will have SUFFIX added on to them.
+When using `sync`, `copy` or `move` any files which would have been
+overwritten or deleted will have the suffix added to them.  If there 
+is a file with the same path (after the suffix has been added), then 
+it will be overwritten.
 
-See `--backup-dir` for more info.
+The remote in use must support server side move or copy and you must
+use the same remote as the destination of the sync.
+
+This is for use with files to add the suffix in the current directory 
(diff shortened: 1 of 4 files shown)


yparitcher:
I have started implementing #2801 however when using --suffix without --backup-dir on a folder (not a file) that already has a file with the suffix added on the remote only, it just gives it 2 suffixes and does not backup the newly overwritten file. which is probably not the intended behavior.

This is shown in the second half of the test case.
https://github.com/yparitcher/rclone/blob/ba2199d6f1dabfd4206d64d3d6fb32432fcb084f/fs/sync/sync_test.go#L1332-L1355
which is why it currently fails.

i therefore think to restrict using --suffix by itself to only files (not folders), however as i am not familiar with rclone code, i am not sure what is the right opperation to use to check if it run on a file.

Please advise,
Y

yparitcher:
depends on #3219

yparitcher:
superseded by  #3272

PR #3276, opened 2019-06-19
 Add tracking for transfer history

ajankovic:

Currently accounting is keeping track of transfers by only showing transfers that are currently underway. All completed transfers are removed from this list and only reference to them is incremented counter.

This makes it difficult to keep track of what was transferred or the per transfer reason of why it failed and at what point. So history of transfers is kept in memory and made available to the `core/transferred` call.

#### Implemented changes

- Change accounting to keep track of transfers that have been completed
- Add rc call for getting list of recently completed transfers
- Transfer history retention time is configurable
- If file transfers are part of an async job they are marked with jobid
- Added per file transfer error description

#### Checklist

- [] I have added documentation for the changes if appropriate.

### DIFF
--- a/cmd/serve/httplib/serve/serve.go
+++ b/cmd/serve/httplib/serve/serve.go
@@ -76,21 +76,10 @@ func Object(w http.ResponseWriter, r *http.Request, o fs.Object) {
 		return
 	}
 	accounting.Stats.Transferring(o.Remote())
-	in := accounting.NewAccount(file, o) // account the transfer (no buffering)
 	defer func() {
-		closeErr := in.Close()
-		if closeErr != nil {
-			fs.Errorf(o, "Get request: close failed: %v", closeErr)
-			if err == nil {
-				err = closeErr
-			}
-		}
-		ok := err == nil
-		accounting.Stats.DoneTransferring(o.Remote(), ok)
-		if !ok {
-			accounting.Stats.Error(err)
-		}
+		accounting.Stats.DoneTransferring(r.Context(), o.Remote(), err)
 	}()
+	in := accounting.Stats.NewAccount(r.Context(), file, o) // account the transfer (no buffering)
 
 	w.WriteHeader(code)
 
--- a/fs/accounting/accounting.go
+++ b/fs/accounting/accounting.go
@@ -2,6 +2,8 @@
 package accounting
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"sync"
@@ -14,9 +16,71 @@ import (
 	"github.com/pkg/errors"
 )
 
+type jobCtx int64
+
+const jobKey jobCtx = 1
+
+// WithJobID returns copy of the parent with assigned jobID.
+func WithJobID(parent context.Context, jobID int64) context.Context {
+	return context.WithValue(parent, jobKey, jobID)
+}
+
+// JobIDFromContext returns job ID from the context if it's available.
+func JobIDFromContext(ctx context.Context) (int64, bool) {
+	jobID, ok := ctx.Value(jobKey).(int64)
+	return jobID, ok
+}
+
 // ErrorMaxTransferLimitReached is returned from Read when the max
 // transfer limit is reached.
 var ErrorMaxTransferLimitReached = fserrors.FatalError(errors.New("Max transfer limit reached as set by --max-transfer"))
+
+// AccountSnapshot represents state of an account at point in time.
+type AccountSnapshot struct {
+	Name      string `json:"name"`
+	Size      int64  `json:"size"`
+	Bytes     int64  `json:"bytes"`
+	Checked   bool   `json:"checked"`
+	Timestamp msTime `json:"timestamp"`
+	Error     error  `json:"-"`
+	JobID     int64  `json:"jobid"`
+}
+
+// MarshalJSON implements json.Marshaler interface.
+func (as AccountSnapshot) MarshalJSON() ([]byte, error) {
+	err := ""
+	if as.Error != nil {
+		err = as.Error.Error()
+	}
+	type Alias AccountSnapshot
+	return json.Marshal(&struct {
+		Error string `json:"error"`
+		Alias
+	}{
+		Error: err,
+		Alias: (Alias)(as),
+	})
+}
+
+// NewCheckingSnapshot creates snapshot for checked object.
+func NewCheckingSnapshot(obj fs.Object) AccountSnapshot {
+	return AccountSnapshot{
+		Name:      obj.Remote(),
+		Size:      obj.Size(),
+		Bytes:     obj.Size(),
+		Checked:   true,
+		Timestamp: msTime(time.Now().UTC()),
+	}
+}
+
+// NewNamedCheckingSnapshot creates snapshot for checked object.
+func NewNamedCheckingSnapshot(name string) AccountSnapshot {
+	return AccountSnapshot{
+		Name:      name,
+		Checked:   true,
+		Timestamp: msTime(time.Now().UTC()),
+	}
+}
 
 // Account limits and accounts for one transfer
 type Account struct {
@@ -25,12 +89,14 @@ type Account struct {
 	// in http transport calls Read() after Do() returns on
 	// CancelRequest so this race can happen when it apparently
 	// shouldn't.
-	mu      sync.Mutex
-	in      io.Reader
-	origIn  io.ReadCloser
-	close   io.Closer
-	size    int64
-	name    string
+	mu     sync.Mutex
+	in     io.Reader
+	origIn io.ReadCloser
+	close  io.Closer
+	size   int64
+	name   string
+	jobID  int64
+
 	statmu  sync.Mutex    // Separate mutex for stat values.
 	bytes   int64         // Total number of bytes read
 	max     int64         // if >=0 the max number of bytes to transfer
@@ -47,7 +113,8 @@ const averagePeriod = 16 // period to do exponentially weighted averages over
 
 // NewAccountSizeName makes a Account reader for an io.ReadCloser of
 // the given size and name
-func NewAccountSizeName(in io.ReadCloser, size int64, name string) *Account {
+func NewAccountSizeName(ctx context.Context, in io.ReadCloser, size int64, name string) *Account {
+	jobID, _ := JobIDFromContext(ctx)
 	acc := &Account{
 		in:     in,
 		close:  in,
@@ -58,15 +125,15 @@ func NewAccountSizeName(in io.ReadCloser, size int64, name string) *Account {
 		avg:    0,
 		lpTime: time.Now(),
 		max:    int64(fs.Config.MaxTransfer),
+		jobID:  jobID,
 	}
 	go acc.averageLoop()
-	Stats.inProgress.set(acc.name, acc)
 	return acc
 }
 
 // NewAccount makes a Account reader for an object
-func NewAccount(in io.ReadCloser, obj fs.Object) *Account {
-	return NewAccountSizeName(in, obj.Size(), obj.Remote())
+func NewAccount(ctx context.Context, in io.ReadCloser, obj fs.Object) *Account {
+	return NewAccountSizeName(ctx, in, obj.Size(), obj.Remote())
 }
 
 // WithBuffer - If the file is above a certain size it adds an Async reader
@@ -219,7 +286,6 @@ func (acc *Account) Close() error {
 	}
 	acc.closed = true
 	close(acc.exit)
-	Stats.inProgress.clear(acc.name)
 	if acc.close == nil {
 		return nil
 	}
@@ -346,8 +412,22 @@ func (acc *Account) RemoteStats() (out map[string]interface{}) {
 		percentageDone = int(100 * float64(a) / float64(b))
 	}
 	out["percentage"] = percentageDone
+	if acc.jobID > 0 {
+		out["jobid"] = acc.jobID
+	}
 
 	return out
+}
+
+// Snapshot produces stats for this account at point in time.
+func (acc *Account) Snapshot() AccountSnapshot {
+	a, b := acc.progress()
+	return AccountSnapshot{
+		Name:      acc.name,
+		Size:      b,
+		Bytes:     a,
+		Timestamp: msTime(time.Now().UTC()),
+	}
 }
 
 // OldStream returns the top io.Reader
--- a/fs/accounting/mstime.go
+++ b/fs/accounting/mstime.go
@@ -1,0 +1,27 @@
+package accounting
+
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
+// msTime is helper type for representing unix epoch timestamp in milliseconds.
+type msTime time.Time
+
+func (t msTime) MarshalJSON() ([]byte, error) {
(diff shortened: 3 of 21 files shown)


ncw:
I gave the build a kick - hopefully it will complete this time.

fs/accounting/accounting.go
```diff
@@ -14,23 +16,86 @@ import (
...
+
+func (as AccountSnapshot) MarshalJSON() ([]byte, error) {
+	err := ""
+	if as.Error != nil {
+		err = as.Error.Error()
+	}
+	type Alias AccountSnapshot
+	return json.Marshal(&struct {
+		Error string `json:"error"`
+		Alias
```
golangcibot: structtag: struct field Error repeats json tag "error" also at accounting.go:56 (from `govet`)

fs/accounting/accounting.go
```diff
@@ -14,23 +16,86 @@ import (
...
+	}{
+		Error: err,
+		Alias:    (Alias)(as),
```
golangcibot: File is not `goimports`-ed (from `goimports`)

fs/accounting/accounting.go
```diff
@@ -14,23 +16,86 @@ import (
...
+	Name      string `json:"name"`
+	Size      int64  `json:"size"`
+	Bytes     int64  `json:"bytes"`
+	Checked   bool   `json:"checked"`
+	Timestamp msTime `json:"timestamp"`
+	Error     error  `json:"error"`
+	JobID     int64  `json:"jobid"`
+}
+
+func (as AccountSnapshot) MarshalJSON() ([]byte, error) {
```
golangcibot: exported method `AccountSnapshot.MarshalJSON` should have comment or be unexported (from `golint`)

fs/config/configflags/configflags.go
```diff
@@ -100,6 +100,8 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.IntVarP(flagSet, &fs.Config.MultiThreadStreams, "multi-thread-streams", "", fs.Config.MultiThreadStreams, "Max number of streams to use for multi-thread downloads.")
 	flags.DurationVarP(flagSet, &fs.Config.RcJobExpireDuration, "rc-job-expire-duration", "", fs.Config.RcJobExpireDuration, "expire finished async jobs older than this value")
 	flags.DurationVarP(flagSet, &fs.Config.RcJobExpireInterval, "rc-job-expire-interval", "", fs.Config.RcJobExpireInterval, "interval to check for expired async jobs")
+	flags.DurationVarP(flagSet, &fs.Config.TransferredExpireDuration, "accounting-transferred-expire-duration", "", fs.Config.TransferredExpireDuration, "remove transferred accounting stats older than this value")
```
mmatczuk: Would it be possible to reuse RcJobExpireDuration and RcJobExpireInterval and auto delete them when job is deleted?

fs/accounting/mstime.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
+// msTime is helper type for representing unix epoch timestamp in milliseconds.
+type msTime time.Time
+
+func (t msTime) MarshalJSON() ([]byte, error) {
```
mmatczuk: Could we have a test for that?

fs/accounting/accounting.go
```diff
@@ -58,15 +125,15 @@ func NewAccountSizeName(in io.ReadCloser, size int64, name string) *Account {
...
 	}
 	go acc.averageLoop()
-	Stats.inProgress.set(acc.name, acc)
 	return acc
 }
 
 // NewAccount makes a Account reader for an object
-func NewAccount(in io.ReadCloser, obj fs.Object) *Account {
-	return NewAccountSizeName(in, obj.Size(), obj.Remote())
+func NewAccount(ctx context.Context, in io.ReadCloser, obj fs.Object) *Account {
```
mmatczuk: This function is now only used in test I'd suggest to drop it.

fs/accounting/accounting.go
```diff
@@ -47,7 +113,8 @@ const averagePeriod = 16 // period to do exponentially weighted averages over
 
 // NewAccountSizeName makes a Account reader for an io.ReadCloser of
 // the given size and name
-func NewAccountSizeName(in io.ReadCloser, size int64, name string) *Account {
+func NewAccountSizeName(ctx context.Context, in io.ReadCloser, size int64, name string) *Account {
```
mmatczuk: This function is only used inside the package maybe we should not export it to avoid confusion with `Stats.NewAccount` function.

fs/accounting/accounting.go
```diff
@@ -14,23 +16,87 @@ import (
...
+		Alias: (Alias)(as),
+	})
+}
+
+// NewCheckingSnapshot creates snapshot for checked object.
+func NewCheckingSnapshot(obj fs.Object) AccountSnapshot {
+	return AccountSnapshot{
+		Name:      obj.Remote(),
+		Size:      obj.Size(),
+		Bytes:     obj.Size(),
```
mmatczuk: It would be better if `"bytes": total transferred bytes for this file,` would be the actually transferred bytes. If we have a transferred entry we can assume success if error is "".

fs/accounting/stats.go
```diff
@@ -63,6 +65,38 @@ Returns the following values:
...
+` + "```" + `
+{
+	"transferred":  an array of completed transfers (including failed ones):
+		[
+			{
+				"name": name of the file,
+				"size": size of the file in bytes,
+				"bytes": total transferred bytes for this file,
+				"checked": if the transfer is only checked (skipped, deleted),
+				"timestamp": integer representing millisecond unix epoch,
```
mmatczuk: Of what event?

fs/accounting/stats.go
```diff
@@ -63,6 +65,38 @@ Returns the following values:
...
+
+` + "```" + `
+{
+	"transferred":  an array of completed transfers (including failed ones):
+		[
+			{
+				"name": name of the file,
+				"size": size of the file in bytes,
+				"bytes": total transferred bytes for this file,
+				"checked": if the transfer is only checked (skipped, deleted),
```
mmatczuk: I'd suggest renaming that to `checked_only` for clarity, technically every file is checked...

mmatczuk: Or dropping that, we can figure out that is was checked only if `bytes = 0` and `error = ""`, thoughts?

fs/config/configflags/configflags.go
ajankovic: It would, but it's a different thing. Completed transfers are not necessarily related to jobs.

fs/config.go
```diff
@@ -133,6 +135,8 @@ func NewConfig() *ConfigInfo {
 	c.MultiThreadStreams = 4
 	c.RcJobExpireDuration = 60 * time.Second
 	c.RcJobExpireInterval = 10 * time.Second
+	c.TransferredExpireDuration = 60 * time.Second
```
ncw: Maybe we could re-use the `RcJobExpireDuration` and `Interval` for this to save on flag explosion?

fs/operations/operations.go
```diff
@@ -1178,14 +1176,7 @@ func Cat(ctx context.Context, f fs.Fs, w io.Writer, offset, count int64) error {
 				size = count
 			}
 		}
-		in = accounting.NewAccountSizeName(in, size, o.Remote()).WithBuffer() // account and buffer the transfer
-		defer func() {
```
ncw: Where does `in` get closed?

Maybe in `accounting.Stats.DoneTransferring`?  That seems a bit too magic for me.

fs/accounting/mstime.go
```diff
@@ -0,0 +1,27 @@
+package accounting
+
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
+// msTime is helper type for representing unix epoch timestamp in milliseconds.
```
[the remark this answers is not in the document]
ajankovic: Sure I just thought that this was a more standardized way. I will remove it.

fs/accounting/accounting.go
```diff
@@ -14,23 +16,87 @@ import (
 	"github.com/pkg/errors"
 )
 
+type jobCtx int64
```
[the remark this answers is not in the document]
ajankovic: That would make accounting directly depend on jobs which I wanted to avoid. In POC this was more generic "group" identifier that jobs happen to be using. But since I couldn't find other uses I went for actual jobid here.

Maybe `accounting` can depend on `jobs` package I guess, since we are adding reference to jobid in accounting anyways.

fs/config.go
ajankovic: Yeah I had suggestion like that above. It comes down to, if perhaps someone would need to have different schedule for transferred files than with jobs. And since we don't have such requirement, I guess we can reuse them.

ajankovic:
> I don't like the magic closing of file handles - it is important that that is very obvious so someone looking at the code can see - yes that file handle got closed.

Yeah i understand what you are saying. But I had the same feeling for the previous implementation :)

It was "magical" to me that call to `accounting.NewAccount` is doing side effects and adding itself to the private field of `Stats` and then caller had to keep track of it even though `Stats` had reference to it. It was like spilling internal stuff onto caller. So it wasn't really "magical" but I would call it unnecessary and complicated.

The reason proposed implementation is less magical in my opinion is that you are now calling `Stats.NewAccount` which suggests that object which is handing you the `Account` has reference to it and it's responsible for it's disposal. Maybe `Stats.TrackedReader` or something similar would be more appropriate but I didn't want to make drastic changes.

The idea is to keep all interface for accounting on the `Stats` object and you just make proper API interaction with it by `Transfering` and `DoneTransfer`. If you need to track byte reading you ask for proper object that is capable of tracking reads from `Stats` which should keep track of closing when transfer is complete. Maybe all of this needs even bigger overhaul but this was the least intrusive way of dealing with it that I found.

Another technical reason for moving it to `Stats` is to force ownership of `Account` because if account is instantiated without `Stats` knowing about it we can't get transferring information when transfer is complete. And also (this is probably a non-issue but) I was worried about deadlocks because `Account` is still using `Stats` directly for other operations and centralizing it's closing would help.

ajankovic:
In general I also have concern for global nature of `Stats` object. Especially when running in server (`rcd`) mode. Because if you want server to handle multiple transactions at different times the tracking process is still running. The `elapsedTime` will tick, the `speed` will reduce, the errors will remain, it just doesn't have concept of distinctive transactions. It becomes averaged, not so accurate, representation of stats for the entire servers lifetime. Maybe opening up reset stats would help but that has problems of it's own, main being it requires synchronous actions. Actually in general, synchronous actions are needed because of `accounting` handling errors and some operations will fail if `Stats` reports an error.

I understand CLI usage was the main thing when this was design so with your permission I would probably make a new PR that would address this with some better implementation that targets long running server that responds to remote calls.

cmd/serve/httplib/serve/serve.go
```diff
@@ -76,21 +76,10 @@ func Object(w http.ResponseWriter, r *http.Request, o fs.Object) {
 		return
 	}
 	accounting.Stats.Transferring(o.Remote())
-	in := accounting.NewAccount(file, o) // account the transfer (no buffering)
 	defer func() {
-		closeErr := in.Close()
```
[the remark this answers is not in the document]
ajankovic: `DoneTransferring` will handle closing and error reporting on them. You just take `NewAccount` from `Stats` that keeps track of it.

ncw:
> > I don't like the magic closing of file handles - it is important that that is very obvious so someone looking at the code can see - yes that file handle got closed.
> 
> Yeah i understand what you are saying. But I had the same feeling for the previous implementation :)
> 
> It was "magical" to me that call to `accounting.NewAccount` is doing side effects and adding itself to the private field of `Stats` and then caller had to keep track of it even though `Stats` had reference to it. It was like spilling internal stuff onto caller. So it wasn't really "magical" but I would call it unnecessary and complicated.

Not ideal certainly!
 
> The reason proposed implementation is less magical in my opinion is that you are now calling `Stats.NewAccount` which suggests that object which is handing you the `Account` has reference to it and it's responsible for it's disposal. Maybe `Stats.TrackedReader` or something similar would be more appropriate but I didn't want to make drastic changes.
> 
> The idea is to keep all interface for accounting on the `Stats` object and you just make proper API interaction with it by `Transfering` and `DoneTransfer`. If you need to track byte reading you ask for proper object that is capable of tracking reads from `Stats` which should keep track of closing when transfer is complete. Maybe all of this needs even bigger overhaul but this was the least intrusive way of dealing with it that I found.
> 
> Another technical reason for moving it to `Stats` is to force ownership of `Account` because if account is instantiated without `Stats` knowing about it we can't get transferring information when transfer is complete. And also (this is probably a non-issue but) I was worried about deadlocks because `Account` is still using `Stats` directly for other operations and centralizing it's closing would help.

Maybe it is worth thinking about the use cases... I think there are two main use cases, firstly just accounting something that is going on and secondly accounting a transfer.

The two uses cases look like

```go
accounting.Stats.Transferring(remote)
// stuff
accounting.Stats.DoneTransferring(remote, true)
```
and
```go
accounting.Stats.Transferring(o.Remote())
in := accounting.NewAccount(file, o)
///stuff
in.Close()
accounting.Stats.DoneTransferring(o.Remote(), ok)
```

It is a pretty clunky interface...

What about if we put everything on the accounting Object, so the two use cases could be something like this.  So de-emphasise the role of the global statsl

I like the New/Done pairing it is obvious that we closed what we opened.

```go
acc := accounting.Stats.NewTransfer(remote, nil)
// stuff
acc.Done(ok)
```
and
```go
acc := accounting.Stats.NewTransfer(file, o)
///stuff
acc.Done(ok)
```

> In general I also have concern for global nature of `Stats` object. Especially when running in server (`rcd`) mode. Because if you want server to handle multiple transactions at different times the tracking process is still running. The `elapsedTime` will tick, the `speed` will reduce, the errors will remain, it just doesn't have concept of distinctive transactions. It becomes averaged, not so accurate, representation of stats for the entire servers lifetime. Maybe opening up reset stats would help but that has problems of it's own, main being it requires synchronous actions. Actually in general, synchronous actions are needed because of `accounting` handling errors and some operations will fail if `Stats` reports an error.
> 
> I understand CLI usage was the main thing when this was designed so with your permission I would probably make a new PR that would address this with some better implementation that targets long running server that responds to remote calls.

Yes global stats needs addressing at some point!

Nanodragon999:
Just an idea i had but, maybe adding support for exporting the transfer history to text/DB like sqlite etc/csv/json etc with some other details (such as path, filename, size, time it got transferred) would be very useful coupled with this :+1:

ncw:
I'm sorry I haven't managed to merge this yet, I'm just not quite happy with the design.

I'd like to see the error from closing the wrapped stream propagated to the outside, preferably with a method called `Close()`.

I think the idea of making the dependency betwen the Account and the Stats structs explicit is a good one - I like that and that will let you add the transfer history.

So can you have a go at breaking this commit up a bit please, maybe one commit changing the interface, one commit refactoring the rc stuff and a commit adding the trasfer history tracking?

Thank you.

PS. I'm going to on vacation for a while so will be a bit quiet!

ajankovic:
@ncw I've created new pull request as I have reconsidered approach for solving this. See #3370 .

ncw:
OK I'll close this one and we'll continue in  #3370

PR #3277, opened 2019-06-19
fs/config: Add method to reload configfile from disk

### LINKED ISSUE #3268: Multiple calls to LoadConfig() to force a reload of the config file
I'm trying to use Rclone as a library but running into some issues that I wanted to run by you guys.

I'm implementing Rclone into a long running process in such a way that the Rclone config might change while the process is running. I'm trying to pick up changes of the config file in order to always display the latest remotes using `config.FileSections()`. However after it cached the config it will use the cache, even if the underlying config file changes (by manually editing the file or using `rclone config`)

My first instinct was to just call `LoadConfig()` before `FileSections()` but I noticed some other methods being called in `LoadConfig()` and I therefor was not sure if it's safe to call `LoadConfig()` multiple times. Is it safe to do so? And if not could you recommend an other way to ensure I always have the most up to date data from the config file?

maran:
Fixes #3268

Allowing a configfile to be re-read from disk when you expect there might have been changes that were written from an other please.

Yes in #3268

#### Checklist

### DIFF
--- a/fs/config/config_test.go
+++ b/fs/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


fs/config/config_test.go
```diff
@@ -292,3 +293,20 @@ func TestMatchProvider(t *testing.T) {
...
+	}))
+	b, err := ioutil.ReadFile(ConfigPath)
+	assert.NoError(t, err)
+
+	b = bytes.Replace(b, []byte("refresh_test"), []byte("refreshed_test"), 1)
+	err = ioutil.WriteFile(ConfigPath, b, 0644)
+	assert.NoError(t, err)
+
+	assert.NotEqual(t, []string{"refreshed_test"}, configFile.GetSectionList())
+	FileRefresh()
```
golangcibot: Error return value of `FileRefresh` is not checked (from `errcheck`)

ncw:
Very nice - thank you :-)

PR #3280, opened 2019-06-20
march: return errors when listing dirs

### LINKED ISSUE #3172: Behaviour for HTTP API is not the same as with CLI API

Sending requests that should fail (missing local dir, access denied to remote, etc.) to rclone HTTP server are not failing. They end with empty object and 200 code result. Tested with both curl and rclone rc.

Example:

```
a> rclone rcd -vv --rc-no-auth
b> rclone rc sync/move srcFs=/non-existent/ dstFs=created-but-inaccessible-remote:bucket/test
{}
b> curl --verbose -H "Content-Type: application/json" -d '{"srcFs":"local-dir-that-exists", "dstFs":"created-but-inaccessible-remote:bucket/test"}' http://localhost:5572/sync/move
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 5572 (#0)
> POST /sync/move HTTP/1.1
> Host: localhost:5572
> User-Agent: curl/7.58.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 65
> 
* upload completely sent off: 65 out of 65 bytes
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: 
< Access-Control-Allow-Origin: *
< Date: Tue, 07 May 2019 08:11:03 GMT
< Content-Length: 3
< Content-Type: text/plain; charset=utf-8
< 
{}

```

Same request made with CLI is reporting an error (with retries) and exiting with `1`. I've taken a look at the code and it seems that CLI calls are wrapped with `cmd.Run` function which integrates with `accounting` and has retries. This functionality is not available within handler for HTTP server so requests never fail. Although available in the server error logs.

Which leads me to another "API design" question, why is for example `sync.MoveDir` not returning an error on failure? It seems currently the only way to know if there was a failure is to use `accounting.Stats`. From how these top level functions are used it seems like entire call chain depends on these top level functions to return errors.

rclone v1.47.0-DEV
- os/arch: linux/amd64
- go version: go1.12.4

Ubuntu 18.04

AWS S3 but it shouldn't matter to this issue.

HTTP Server logs:

```
2019/05/07 10:28:09 DEBUG : rclone: Version "v1.47.0-DEV" starting with parameters ["rclone" "rcd" "-vv" "--rc-no-auth"]
2019/05/07 10:28:09 NOTICE: Serving remote control on http://127.0.0.1:5572/
2019/05/07 10:28:14 DEBUG : rc: "sync/move": with parameters map[dstFs:bla:bucket/test srcFs:/non-existent/]
2019/05/07 10:28:14 DEBUG : Using config file from "/home/alek/.config/rclone/rclone.conf"
2019/05/07 10:28:15 ERROR : S3 bucket bucket path test/: Failed to update region for bucket: reading bucket location failed: AccessDenied: Access Denied
	status code: 403, request id: 5F13BD3082D862ED, host id: FlUy7pvhPWARqV4qVNBGRhKbrIykPXQnFAfRoU/6CK6HGs3qyWaZq/u57YLYOjqbn0yAa0WyoKM=
2019/05/07 10:28:15 ERROR : S3 bucket bucket path test/: Failed to update region for bucket: reading bucket location failed: AccessDenied: Access Denied
	status code: 403, request id: F45BAF0FA7E4A5D9, host id: SRxpFTaW90nTzWMToU5XBAeHI4lepEBM3mtTJrBwHx2L99FMQhqz2rLEtrcX+qDzKm6bcPx4u4g=
2019/05/07 10:28:15 ERROR : : error reading source directory: directory not found
2019/05/07 10:28:15 INFO  : S3 bucket bucket path test/: Waiting for checks to finish
2019/05/07 10:28:15 INFO  : S3 bucket bucket path test/: Waiting for transfers to finish
2019/05/07 10:28:15 DEBUG : rc: "sync/move": reply map[]: <nil>
2019/05/07 10:28:57 DEBUG : rc: "sync/move": with parameters map[dstFs:bla:scylla-manager-aleksandar/bash srcFs:tmp]
2019/05/07 10:28:57 ERROR : : error reading source directory: directory not found
2019/05/07 10:28:57 INFO  : S3 bucket scylla-manager-aleksandar path bash/: Waiting for checks to finish
2019/05/07 10:28:57 INFO  : S3 bucket scylla-manager-aleksandar path bash/: Waiting for transfers to finish
2019/05/07 10:28:57 DEBUG : rc: "sync/move": reply map[]: <nil>
```

ajankovic:

Partially fixes #3172

As described in linked issue errors in file system access are not returned when copying/moving. This change corrects that by catching any listing errors.

#### Checklist

### DIFF
--- a/fs/march/march.go
+++ b/fs/march/march.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/pkg/errors"
+
 	"github.com/ncw/rclone/fs"
 	"github.com/ncw/rclone/fs/filter"
 	"github.com/ncw/rclone/fs/list"
@@ -110,7 +112,7 @@ type listDirJob struct {
 }
 
 // Run starts the matching process off
-func (m *March) Run() {
+func (m *March) Run() error {
 	m.init()
 
 	srcDepth := fs.Config.MaxDepth
@@ -121,6 +123,10 @@ func (m *March) Run() {
 	if filter.Active.Opt.DeleteExcluded {
 		dstDepth = fs.MaxLevel
 	}
+
+	var mu sync.Mutex // Protects vars below
+	var jobError error
+	var errCount int
 
 	// Start some directory listing go routines
 	var wg sync.WaitGroup         // sync closing of go routines
@@ -138,7 +144,16 @@ func (m *March) Run() {
 					if !ok {
 						return
 					}
-					jobs := m.processJob(job)
+					jobs, err := m.processJob(job)
+					if err != nil {
+						mu.Lock()
+						// Keep reference only to the first encountered error
+						if jobError == nil {
+							jobError = err
+						}
+						errCount++
+						mu.Unlock()
+					}
 					if len(jobs) > 0 {
 						traversing.Add(len(jobs))
 						go func() {
@@ -178,6 +193,11 @@ func (m *March) Run() {
 	traversing.Wait()
 	close(in)
 	wg.Wait()
+
+	if errCount > 1 {
+		return errors.Wrapf(jobError, "march failed with %d error(s): first error", errCount)
+	}
+	return jobError
 }
 
 // Check to see if the context has been cancelled
@@ -339,8 +359,9 @@ func matchListings(srcListEntries, dstListEntries fs.DirEntries, transforms []ma
 // more jobs
 //
 // returns errors using processError
-func (m *March) processJob(job listDirJob) (jobs []listDirJob) {
+func (m *March) processJob(job listDirJob) ([]listDirJob, error) {
 	var (
+		jobs                   []listDirJob
 		srcList, dstList       fs.DirEntries
 		srcListErr, dstListErr error
 		wg                     sync.WaitGroup
@@ -367,14 +388,14 @@ func (m *March) processJob(job listDirJob) (jobs []listDirJob) {
 	if srcListErr != nil {
 		fs.Errorf(job.srcRemote, "error reading source directory: %v", srcListErr)
 		fs.CountError(srcListErr)
-		return nil
+		return nil, srcListErr
 	}
 	if dstListErr == fs.ErrorDirNotFound {
 		// Copy the stuff anyway
 	} else if dstListErr != nil {
 		fs.Errorf(job.dstRemote, "error reading destination directory: %v", dstListErr)
 		fs.CountError(dstListErr)
-		return nil
+		return nil, dstListErr
 	}
 
 	// If NoTraverse is set, then try to find a matching object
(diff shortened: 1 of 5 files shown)


ncw:
The build for this is failing on go1.9

```
# github.com/ncw/rclone/fs/march
fs/march/multierror.go:29:8: undefined: strings.Builder
```

I try to keep support for 2 years worth of go releases so that rclone can keep building for distros which are a bit behind.

ajankovic:
Build failure is now different and vague I guess it needs another run.

fs/march/multierror.go
```diff
@@ -0,0 +1,40 @@
+package march
+
+import (
+	"bytes"
+	"fmt"
+	"sync"
+)
+
+type multierror struct {
```
mmatczuk: Why do we need a new multierror type? 

How about using an existing package like one from Uber or HashiCorp?

fs/march/march.go
```diff
@@ -122,6 +122,7 @@ func (m *March) Run() {
 		dstDepth = fs.MaxLevel
 	}
 
+	merr := &multierror{}
```
mmatczuk: If you used `github.com/uber-go/multierr` you could avoid allocation here.

fs/march/multierror.go
ajankovic: We need protection for concurrent use I don't think those packages provide that?

fs/march/march.go
ajankovic: What about concurrent usage needed below?

fs/march/multierror.go
ncw: I think I'd be happy just showing the last error (or the first).  More than a couple of errors are going to make the error messages unreadable.

ajankovic:
Ok I replaced multierror with first occurring error and error count.

PTAL

fs/march/march.go
```diff
@@ -178,6 +192,11 @@ func (m *March) Run() {
 	traversing.Wait()
 	close(in)
 	wg.Wait()
+
+	if errCount > 1 {
+		return fmt.Errorf("%+v (and %d others)", jobError, errCount-1)
```
ncw: Can you make that something like `return errors.Wrapf(err, "march failed with %d errors: first error", errCount)` - rclone uses pkg/errors for error annotations and that will enable consumers to unwrap the error.

fs/march/march.go
```diff
@@ -3,6 +3,7 @@ package march
 
 import (
 	"context"
+	"github.com/pkg/errors"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

ncw:
That is perfect now - thank you very much :-)

I'll merge now.

PR #3290, opened 2019-06-24
Calculate correct Content-Length on MultiPart Uploads

fionera:
Add the Size of the Boundaries etc to the Content-Length so its a valid value.

No.

#### Checklist

### DIFF
--- a/backend/pcloud/pcloud.go
+++ b/backend/pcloud/pcloud.go
@@ -9,11 +9,9 @@ package pcloud
 // FIXME mime type? Fix overview if implement.
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


lib/rest/rest.go
```diff
@@ -292,11 +292,39 @@ func (api *Client) Call(opts *Opts) (resp *http.Response, err error) {
...
 	contentType := writer.FormDataContentType()
 
+	// Create a Multipart Writer as base for calculating the Content-Length
+	buf := &bytes.Buffer{}
+	dummyMultipartWriter := multipart.NewWriter(buf)
+	err := dummyMultipartWriter.SetBoundary(writer.Boundary())
+	if err != nil {
+		return nil, "", nil, err
+	}
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

lib/rest/rest.go
```diff
@@ -292,11 +292,39 @@ func (api *Client) Call(opts *Opts) (resp *http.Response, err error) {
 // contentName - the name of the parameter for the file
```
ncw: Can you update the comment please as to what it returns?

I think the best name for it would be `overhead` as it is the overhead of the multpart encoding.

lib/rest/rest.go
```diff
@@ -292,11 +292,39 @@ func (api *Client) Call(opts *Opts) (resp *http.Response, err error) {
...
 //
 // NB This doesn't allow setting the content type of the attachment
-func MultipartUpload(in io.Reader, params url.Values, contentName, fileName string) (io.ReadCloser, string, error) {
+func MultipartUpload(in io.Reader, params url.Values, contentName, fileName string) (io.ReadCloser, string, *int64, error) {
```
ncw: Why is the return an `*int64`?  I would have thought an `int64` would be fine? That saves problems with `nil` pointers etc.

backend/pcloud/pcloud.go
```diff
@@ -1091,21 +1089,18 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 	// opts.Body=0), so upload it as a multpart form POST with
 	// Content-Length set.
 	if size == 0 {
-		formReader, contentType, err := rest.MultipartUpload(in, opts.Parameters, "content", leaf)
+		formReader, contentType, length, err := rest.MultipartUpload(in, opts.Parameters, "content", leaf)
```
ncw: can you rename `length` to `overhead`?

lib/rest/rest.go
```diff
@@ -403,10 +431,15 @@ func (api *Client) callCodec(opts *Opts, request interface{}, response interface
 			params.Add(opts.MultipartMetadataName, string(requestBody))
 		}
 		opts = opts.Copy()
-		opts.Body, opts.ContentType, err = MultipartUpload(opts.Body, params, opts.MultipartContentName, opts.MultipartFileName)
+
+		fileSize := *opts.ContentLength
```
ncw: I'd suggest something like this as supplying the ContentLength is optional...

```go
var overhead int64
opts.Body, opts.ContentType, overhead, err = MultipartUpload(opts.Body, params, opts.MultipartContentName, opts.MultipartFileName)
if *opts.ContentLength != nil {
    *opts.ContentLength += overhead
}
```

ncw:
I'll close this since it is now part of #3295

Issue #2178, opened 2018-03-24
implement link sharing for all remotes that support it


breunigs:
in #2176 link sharing support was added for some remotes. This is a tracking ticket to add it for all other remotes that support it:

- [ ] Amazon Drive
- [ ] Amazon S3
- [ ] Backblaze B2
- [ ] Box
- [x] Dropbox
- [ ] FTP
- [ ] Google Cloud Storage
- [x] Google Drive
- [ ] HTTP
- [ ] Hubic
- [ ] Microsoft Azure Blob Storage
- [ ] Microsoft OneDrive
- [ ] Openstack Swift
- [ ] pCloud
- [ ] QingStor
- [ ] SFTP
- [ ] WebDAV
- [ ] Yandex Disk
- [x] ~~The local filesystem~~

checkmark means that this remote was considered. If it's strike-through, then the remote cannot support it.

theblackhole:
This might help for Openstack Swift implementation :

From what I could understand, there are two types of urls :
- urls from public containers (eg https://storage.sbg1.ovh.net/v1/AUTH_<token>/<container-name>/path/to/file.txt).
- temporary url created for a file, seems to work on private containers : https://docs.ovh.com/gb/en/public-cloud/share_an_object_via_a_temporary_url/

ncw:
@theblackhole thanks for that.  The first is probably what we want I think.  Do you fancy making a pull request to implement?

theblackhole:
@ncw Unfortunatly I don't know Golang but I may try when I got time.

ncw:
@xiaolei0125 would you like to send a pull request for this?

If you implement the `PublicLink` optional method in the s3 backend with the code above that should be it!

For the time being put a really long expires on it and we can work out how to pass the desired expiry time in a second PR.

xiaolei0125:
@ncw  Ok,  I've been a bit busy recently,  I may try when I got time.

ivandeex:
I stumbled upon this ticket paging through "least recently updated, not new backend, not from ncw", preferably labeled "help wanted" (aka "we are not going to take on this anytime soon"). Such tickets are mostly rot, rusted, non-reproducible and forgotten, _but not this one_. It's in fact a _roadmap ticket_. Marking as important to not forget.

Also I'd like to propose a column in https://rclone.org/overview/#features marking backends with public links. Maybe _albertony_ or _edwardxml_ would be kind to tackle? (Deliberately no mentions yet to reduce noise).

cc @ncw

[...]

PR #3314, opened 2019-06-30
b2: Implement link sharing #2178

yparitcher:

Add link sharing to B2

#2178

#### Checklist

### DIFF
--- a/backend/b2/api/types.go
+++ b/backend/b2/api/types.go
@@ -189,6 +189,21 @@ type GetUploadURLResponse struct {
 	AuthorizationToken string `json:"authorizationToken"` // The authorizationToken that must be used when uploading files to this bucket, see b2_upload_file.
 }
 
+// GetDownloadAuthorizationRequest is passed to b2_get_download_authorization
+type GetDownloadAuthorizationRequest struct {
+	BucketID               string `json:"bucketId"`                       // The ID of the bucket that you want to upload to.
+	FileNamePrefix         string `json:"fileNamePrefix"`                 // The file name prefix of files the download authorization token will allow access to.
+	ValidDurationInSeconds int64  `json:"validDurationInSeconds"`         // The number of seconds before the authorization token will expire. The minimum value is 1 second. The maximum value is 604800 which is one week in seconds.
+	B2ContentDisposition   string `json:"b2ContentDisposition,omitempty"` // optional - If this is present, download requests using the returned authorization must include the same value for b2ContentDisposition.
+}
+
+// GetDownloadAuthorizationResponse is received from b2_get_download_authorization
+type GetDownloadAuthorizationResponse struct {
+	BucketID           string `json:"bucketId"`           // The unique ID of the bucket.
+	FileNamePrefix     string `json:"fileNamePrefix"`     // The file name prefix of files the download authorization token will allow access to.
+	AuthorizationToken string `json:"authorizationToken"` // The authorizationToken that must be used when downloading files, see b2_download_file_by_name.
+}
+
 // FileInfo is received from b2_upload_file, b2_get_file_info and b2_finish_large_file
 type FileInfo struct {
 	ID              string            `json:"fileId"`          // The unique identifier for this version of this file. Used with b2_get_file_info, b2_download_file_by_id, and b2_delete_file_version.
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -134,42 +134,53 @@ This is usually set to a Cloudflare CDN URL as Backblaze offers
 free egress for data downloaded through the Cloudflare network.
 Leave blank if you want to use the endpoint provided by Backblaze.`,
 			Advanced: true,
+		}, {
+			Name: "download_auth_duration",
+			Help: `Time before the authorization token will expire in s or suffix ms|s|m|h|d.
+
+The duration before the download authorization token will expire.
+The minimum value is 1 second. The maximum value is one week.`,
+			Default:  fs.Duration(7 * 24 * time.Hour),
+			Advanced: true,
 		}},
 	})
 }
 
 // Options defines the configuration for this backend
 type Options struct {
-	Account         string        `config:"account"`
-	Key             string        `config:"key"`
-	Endpoint        string        `config:"endpoint"`
-	TestMode        string        `config:"test_mode"`
-	Versions        bool          `config:"versions"`
-	HardDelete      bool          `config:"hard_delete"`
-	UploadCutoff    fs.SizeSuffix `config:"upload_cutoff"`
-	ChunkSize       fs.SizeSuffix `config:"chunk_size"`
-	DisableCheckSum bool          `config:"disable_checksum"`
-	DownloadURL     string        `config:"download_url"`
+	Account                       string        `config:"account"`
+	Key                           string        `config:"key"`
+	Endpoint                      string        `config:"endpoint"`
+	TestMode                      string        `config:"test_mode"`
+	Versions                      bool          `config:"versions"`
+	HardDelete                    bool          `config:"hard_delete"`
+	UploadCutoff                  fs.SizeSuffix `config:"upload_cutoff"`
+	ChunkSize                     fs.SizeSuffix `config:"chunk_size"`
+	DisableCheckSum               bool          `config:"disable_checksum"`
+	DownloadURL                   string        `config:"download_url"`
+	DownloadAuthorizationDuration fs.Duration   `config:"download_auth_duration"`
 }
 
 // Fs represents a remote b2 server
 type Fs struct {
-	name          string                       // name of this remote
-	root          string                       // the path we are working on if any
-	opt           Options                      // parsed config options
-	features      *fs.Features                 // optional features
-	srv           *rest.Client                 // the connection to the b2 server
-	bucket        string                       // the bucket we are working on
-	bucketOKMu    sync.Mutex                   // mutex to protect bucket OK
-	bucketOK      bool                         // true if we have created the bucket
-	bucketIDMutex sync.Mutex                   // mutex to protect _bucketID
-	_bucketID     string                       // the ID of the bucket we are working on
-	info          api.AuthorizeAccountResponse // result of authorize call
-	uploadMu      sync.Mutex                   // lock for upload variable
-	uploads       []*api.GetUploadURLResponse  // result of get upload URL calls
-	authMu        sync.Mutex                   // lock for authorizing the account
-	pacer         *fs.Pacer                    // To pace and retry the API calls
-	bufferTokens  chan []byte                  // control concurrency of multipart uploads
+	name            string                       // name of this remote
+	root            string                       // the path we are working on if any
+	opt             Options                      // parsed config options
+	features        *fs.Features                 // optional features
+	srv             *rest.Client                 // the connection to the b2 server
+	bucket          string                       // the bucket we are working on
+	bucketOKMu      sync.Mutex                   // mutex to protect bucket OK
+	bucketOK        bool                         // true if we have created the bucket
+	bucketIDMutex   sync.Mutex                   // mutex to protect _bucketID
+	_bucketID       string                       // the ID of the bucket we are working on
+	bucketTypeMutex sync.Mutex                   // mutex to protect _bucketType
+	_bucketType     string                       // the Type of the bucket we are working on
+	info            api.AuthorizeAccountResponse // result of authorize call
+	uploadMu        sync.Mutex                   // lock for upload variable
+	uploads         []*api.GetUploadURLResponse  // result of get upload URL calls
+	authMu          sync.Mutex                   // lock for authorizing the account
+	pacer           *fs.Pacer                    // To pace and retry the API calls
+	bufferTokens    chan []byte                  // control concurrency of multipart uploads
 }
 
 // Object describes a b2 object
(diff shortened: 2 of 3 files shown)


backend/b2/api/types.go
```diff
@@ -189,6 +189,21 @@ type GetUploadURLResponse struct {
 	AuthorizationToken string `json:"authorizationToken"` // The authorizationToken that must be used when uploading files to this bucket, see b2_upload_file.
 }
 
+// GetDownloadAuthorizationRequest is passed to b2_get_download_authorization
+type GetDownloadAuthorizationRequest struct {
+	BucketID               string `json:"bucketId"`               // The ID of the bucket that you want to upload to.
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/b2/b2.go
```diff
@@ -1125,6 +1149,65 @@ func (f *Fs) Hashes() hash.Set {
...
 }
 
+// getDownloadAuthorization returns authorization token for downloading
+// without accout.
+func (f *Fs) getDownloadAuthorization(remote string) (authorization string, err error) {
+	validDurationInSeconds := time.Duration(f.opt.DownloadAuthorizationDuration).Nanoseconds() / 1e9
+	if validDurationInSeconds == 0 {
+		return "", errors.New("Link command requires --b2-download-auth-duration")
+	}
+	if  validDurationInSeconds < 0 || validDurationInSeconds > 604800 {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	if validDurationInSeconds < 0 || validDurationInSeconds > 604800 {
```

backend/b2/api/types.go
```diff
@@ -189,6 +189,21 @@ type GetUploadURLResponse struct {
 	AuthorizationToken string `json:"authorizationToken"` // The authorizationToken that must be used when uploading files to this bucket, see b2_upload_file.
 }
 
+// GetDownloadAuthorizationRequest is passed to b2_get_download_authorization
+type GetDownloadAuthorizationRequest struct {
+	BucketID               string `json:"bucketId"`                       // The ID of the bucket that you want to upload to.
+	FileNamePrefix         string `json:"fileNamePrefix"`                 // The file name prefix of files the download authorization token will allow access to.
+	ValidDurationInSeconds int64    `json:"validDurationInSeconds"`       // The number of seconds before the authorization token will expire. The minimum value is 1 second. The maximum value is 604800 which is one week in seconds.
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	ValidDurationInSeconds int64  `json:"validDurationInSeconds"`         // The number of seconds before the authorization token will expire. The minimum value is 1 second. The maximum value is 604800 which is one week in seconds.
```

backend/b2/b2.go
```diff
@@ -1125,6 +1149,65 @@ func (f *Fs) Hashes() hash.Set {
...
+	bucketID, err := f.getBucketID()
+	if err != nil {
+		return "", err
+	}
+	opts := rest.Opts{
+		Method: "POST",
+		Path:   "/b2_get_download_authorization",
+	}
+	var request = api.GetDownloadAuthorizationRequest{
+		BucketID: bucketID,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/b2/b2.go
```diff
@@ -1125,6 +1149,65 @@ func (f *Fs) Hashes() hash.Set {
...
+	opts := rest.Opts{
+		Method: "POST",
+		Path:   "/b2_get_download_authorization",
+	}
+	var request = api.GetDownloadAuthorizationRequest{
+		BucketID:               bucketID,
+		FileNamePrefix:         remote,
+		ValidDurationInSeconds: validDurationInSeconds,
+	}
+	var response = api.GetDownloadAuthorizationResponse{
```
golangcibot: File is not `goimports`-ed (from `goimports`)

yparitcher:
@ncw 
Build stalled due to #3154, Please rerun the CI
thanks

backend/b2/b2.go
```diff
@@ -796,6 +804,22 @@ func (f *Fs) listBucketsToFn(fn listBucketFn) error {
 	return nil
 }
 
+// getbucketType finds the bucketType for the current bucket name
+// can be one of allPublic. allPrivate, or snapshot
+func (f *Fs) getbucketType() (bucketType string, err error) {
```
ncw: `Does `listBuckets` work with limited keys? I don't think it does...

Maybe this should be integrated into getBucketID which is very likely to have been called already.  We could then cache type in the `Fs` in case it is called repeatedly

backend/b2/b2.go
```diff
@@ -134,22 +134,30 @@ This is usually set to a Cloudflare CDN URL as Backblaze offers
 free egress for data downloaded through the Cloudflare network.
 Leave blank if you want to use the endpoint provided by Backblaze.`,
 			Advanced: true,
+		}, {
+			Name: "download_auth_duration",
+			Help: `Time before the authorization token will expire in s or suffix ms|s|m|h|d.
+
+The duration before the download authorization token will expire.
+The minimum value is 1 second. The maximum value is one week.`,
+			Advanced: true,
```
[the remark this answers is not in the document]
yparitcher: The B2 api does not Have a Default so i was not sure if i should put one, however if you think i should i will do so.

backend/b2/b2.go
yparitcher: with a Application Key limited to a bucket, `listBuckets` only list the bucket if the `bucketId` is included in the request which `listBucketsToFn` does.
if a key does not have the `listBuckets` permission at all then it wont work, same if the key does not have the `shareFiles` permission, however a default application key limited to a bucket has them.
i could not find a way to determine if a bucket was public or private without `listBuckets` 

when using the link command, as far as i can tell `getBucketID` is not called before `getBucketType`, however it might make sense to include `getBucketType` in `NewFs()` and cache the result.

please let me know what you think

backend/b2/b2.go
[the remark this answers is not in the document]
ncw: Thanks

backend/b2/b2.go
ncw: Good to know about the limited application key.

I think it would be a good idea to cache the bucket type.  If you wanted to add lots of links then using the API would be the way to do it, and if we don't cache the bucket type then we will be doing more calls to b2 than we need.

I would do something like this :-)

```diff
diff --git a/backend/b2/b2.go b/backend/b2/b2.go
index 65e643a11..499a417a3 100644
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -162,8 +162,9 @@ type Fs struct {
 	bucket        string                       // the bucket we are working on
 	bucketOKMu    sync.Mutex                   // mutex to protect bucket OK
 	bucketOK      bool                         // true if we have created the bucket
-	bucketIDMutex sync.Mutex                   // mutex to protect _bucketID
+	bucketIDMutex sync.Mutex                   // mutex to protect _bucketID and _bucketType
 	_bucketID     string                       // the ID of the bucket we are working on
+	_bucketType   string                       // the Type of the bucket we are working on
 	info          api.AuthorizeAccountResponse // result of authorize call
 	uploadMu      sync.Mutex                   // lock for upload variable
 	uploads       []*api.GetUploadURLResponse  // result of get upload URL calls
```

And

```go
// _getBucketInfo finds the bucket ID and type for the current bucket
// name
//
// must be called with the bucketIDMutex held
func (f *Fs) _getBucketInfo() (bucketID, bucketType string, err error) {
	if f._bucketID != "" {
		return f._bucketID, f._bucketType, nil
	}
	var foundBucket *api.Bucket
	err = f.listBucketsToFn(func(bucket *api.Bucket) error {
		if bucket.Name == f.bucket {
			foundBucket = bucket
		}
		return nil

	})
	if foundBucket == nil {
		return "", "", fs.ErrorDirNotFound
	}
	f._bucketID = foundBucket.ID
	f._bucketType = foundBucket.Type
	return f._bucketID, f._bucketType, nil
}

// getBucketID finds the ID for the current bucket name
func (f *Fs) getBucketID() (bucketID string, err error) {
	f.bucketIDMutex.Lock()
	defer f.bucketIDMutex.Unlock()
	bucketID, _, err = f._getBucketInfo()
	return bucketID, err
}

// getBucketType finds the type for the current bucket name
func (f *Fs) getBucketType() (bucketType string, err error) {
	f.bucketIDMutex.Lock()
	defer f.bucketIDMutex.Unlock()
	_, bucketType, err = f._getBucketInfo()
	return bucketType, err
}
```

yparitcher: i agree, however the `bucketID` is included in the response from `AuthorizeAccount` and therefore when using a single bucket ApplicationKey `getBucketID` does not make any call, if we include `getBucketType` (that is **not** included in the response from `AuthorizeAccount`) in one call it will require that every command will add a call to `listBuckets`, even ones that until now did not require it
therefore i think that while caching it, it should be kept as a separate function from `getBucketID`

if you agree i will finish this up.

yparitcher:
I tested Manually, however i was not able to run the integration tests, as they wont run with a applicationkey limited to a bucket.

backend/b2/b2.go
ncw: Ah I see what you mean.  Let's do what you suggest :-)

yparitcher: done in the latest

ncw:
That looks really good now - thank you.  Your solution to caching the bucketType looks very elegant - much better than my sketch!

I'll merge that now - thank you very much :-)

yparitcher:
> That looks really good now - thank you. Your solution to caching the bucketType looks very elegant - much better than my sketch!
> 
> I'll merge that now - thank you very much :-)

i just plagiarized `getBucketID`

yparitcher:
also 2 related points that might need their own issue:

1.  
`downloadurl` is usually going to be used with a public bucket (technically one can pass the auth parameter through the cdn, but that is unlikely), now that `BucketType`is available does it make sense to put a debug warning where` downloadurl `is used with a private bucket (i wouldn't change the behavior as some might want to use the corner case, just issue a warning or notice so it can be noticed by the user)

2.
 >  If you wanted to add lots of links then using the API would be the way to do it, and if we don't cache the bucket type then we will be doing more calls to b2 than we need.

the `downloadauthtoken` can be for a fileprefix so a better way of many links (eg. a folder), would be to make one call to b2 authorizing the folder, then rclone can locally add the same token to every file link drastically reducing api calls, however this would need a separate `PublicLinkFolder()` function. said function can also be used from the commandline if passed a folder.

what do you think?

ncw:
> `downloadurl` is usually going to be used with a public bucket (technically one can pass the auth parameter through the cdn, but that is unlikely), now that `BucketType`is available does it make sense to put a debug warning where`downloadurl`is used with a private bucket (i wouldn't change the behavior as some might want to use the corner case, just issue a warning or notice so it can be noticed by the user)

Mmm, that would mean reading the BucketType unconditionally wouldn't it?  I think `downloadurl` is a niche enough feature that we should expect the users using it to know what they are doing.

Perhaps a note in the config docs for downloadurl about it only working on a public bucket?

https://github.com/ncw/rclone/blob/3f5767b94ed471cc8c521c7c4f06291c097c1573/backend/b2/b2.go#L133-L135

> > If you wanted to add lots of links then using the API would be the way to do it, and if we don't cache the bucket type then we will be doing more calls to b2 than we need.
> 
> the `downloadauthtoken` can be for a fileprefix so a better way of many links (eg. a folder), would be to make one call to b2 authorizing the folder, then rclone can locally add the same token to every file link drastically reducing api calls, however this would need a separate `PublicLinkFolder()` function. said function can also be used from the commandline if passed a folder.
> 
> what do you think?

The `PublicLink` is actually used elsewhere with a path to a file or a directory, eg

https://github.com/ncw/rclone/blob/3f5767b94ed471cc8c521c7c4f06291c097c1573/backend/drive/drive.go#L2100-L2105

The docs in `fs/fs.go` don't state this though which is an oversight.

So could we make it so if the user passed a path to a directory then it authorized everything in the directory?  How would that work?  What users are expecting to see is that if you share a directory you get an http link which shows a directory listing with files you can just download.

yparitcher:
> Perhaps a note in the config docs for downloadurl about it only working on a public bucket?

will do.

> The `PublicLink` is actually used elsewhere with a path to a file or a directory, eg
> 
> https://github.com/ncw/rclone/blob/3f5767b94ed471cc8c521c7c4f06291c097c1573/backend/drive/drive.go#L2100-L2105
> The docs in `fs/fs.go` don't state this though which is an oversight.
> 
> So could we make it so if the user passed a path to a directory then it authorized everything in the directory? How would that work? What users are expecting to see is that if you share a directory you get an http link which shows a directory listing with files you can just download.

since B2 is bucket based there are not really any directories.
 right now if you pass it a folder it will authorize that folder, but the link wont do anything as it does not point to any file. (you can add the file path manually and it will work)
the only thing we can do for a folder is authorize once, then print out links for each file in the folder.
 or i can filter the command that it only works on valid objects, and folders wont work as a limitaion of a bucket based remote

let me know what you think

ncw:
> > Perhaps a note in the config docs for downloadurl about it only working on a public bucket?
> 
> will do.

Thank you

> since B2 is bucket based there are not really any directories.
> right now if you pass it a folder it will authorize that folder, but the link wont do anything as it does not point to any file. (you can add the file path manually and it will work)
> the only thing we can do for a folder is authorize once, then print out links for each file in the folder.

I don't really like that idea it isn't what people expect for PublicLink.  People expect one link for a directory.

> or i can filter the command that it only works on valid objects, and folders wont work as a limitaion of a bucket based remote

If you get a link for a directory then it is fairly easy to turn into a link for a file beneath that directory - maybe we should just document how to do that and leave the code how it is?

yparitcher:
> If you get a link for a directory then it is fairly easy to turn into a link for a file beneath that directory - maybe we should just document how to do that and leave the code how it is?

B2 gives a link like:
```
./rclone link remote:bucket/path/to/file.txt
https://f002.backblazeb2.com/file/bucket/path/to/file.txt?Authorization=xxxxxxxx
```
if given the folder:
```
./rclone link remote:bucket/path
```
Currently rclone returns
```
https://f002.backblazeb2.com/file/bucket/path?Authorization=YYYYYY
```
which obviously does not work as it is not a valid file.
but said auth `?Authorization=YYYYYY` will work on any file matching `path` so:
```
https://f002.backblazeb2.com/file/bucket/path/to/file1?Authorization=YYYYYY
https://f002.backblazeb2.com/file/bucket/path/file2?Authorization=YYYYYY
https://f002.backblazeb2.com/file/bucket/path/folder/file3?Authorization=YYYYYY
```
will all work.

while figuring this out i found a bug in the way i parsed paths for the auth.
submitting a PR with the fix and doc change

ncw:
> where is the best place to document this behavior? in the b2 doc or the link command doc?

I'd make a new subheading in the b2 docs, say `### B2 and rclone link`  or something like that.



> while figuring this out i found a bug in the way i parsed paths for the auth.
> submitting a PR with the fix and doc change

Ah the old missing root problem!  (I've made that mistake many times!)

Will merge in a moment - thank you

PR #3333, opened 2019-07-08
[WIP] union: add `--union-mirror`

### LINKED ISSUE #3308: union: --mirror option for union backend

```
rclone v1.48.0
- os/arch: linux/amd64
- go version: go1.12.6
```

Keep multiple remote (backup) targets in sync

Add a `mirror` option for the union backend which causes it to write any new or changed files to all listed backends rather than simply the last one.  This would allow the use of a single rclone target which puts my backup files on google drive, dropbox, all at once.  This mode of operation would allow users to configure the union backend to be "read top, write all", rather than "read top, write top".

gary-kim:
So, a few things that will need to be decided on. Some of the functions really don't work for a multi-remote backend. For example, for the `PutStream` function: should we check that all remotes support `PutStream` first then perform operations or just perform the operation on all of them and return any errors? I'm leaning towards checking first because in this case, it'd be better to have a failure rather then files only written to some of the remotes.

Add `--union-mirror` option for Union backend.

Closes #3308 

#### Checklist

### DIFF
--- a/backend/union/union.go
+++ b/backend/union/union.go
@@ -23,11 +23,18 @@ func init() {
 		Name:        "union",
 		Description: "A stackable unification remote, which can appear to merge the contents of several remotes",
 		NewFs:       NewFs,
-		Options: []fs.Option{{
-			Name:     "remotes",
-			Help:     "List of space separated remotes.\nCan be 'remotea:test/dir remoteb:', '\"remotea:test/space dir\" remoteb:', etc.\nThe last remote is used to write to.",
-			Required: true,
-		}},
+		Options: []fs.Option{
+			{
+				Name:     "remotes",
+				Help:     "List of space separated remotes.\nCan be 'remotea:test/dir remoteb:', '\"remotea:test/space dir\" remoteb:', etc.\nThe last remote is used to write to.",
+				Required: true,
+			},
+			{
+				Name:     "mirror",
+				Help:     "Mirror operations to all remotes",
+				Advanced: true,
+			},
+		},
 	}
 	fs.Register(fsi)
 }
@@ -35,6 +42,7 @@ func init() {
 // Options defines the configuration for this backend
 type Options struct {
 	Remotes fs.SpaceSepList `config:"remotes"`
+	Mirror  bool            `config:"mirror"`
 }
 
 // Fs represents a union of remotes
@@ -91,6 +99,16 @@ func (f *Fs) Features() *fs.Features {
 
 // Rmdir removes the root directory of the Fs object
 func (f *Fs) Rmdir(ctx context.Context, dir string) error {
+	var err error
+	if f.opt.Mirror {
+		for _, remote := range f.remotes {
+			remoteError := remote.Rmdir(ctx, dir)
+			if remoteError != nil {
+				err = remoteError
+			}
+		}
+		return err
+	}
 	return f.wr.Rmdir(ctx, dir)
 }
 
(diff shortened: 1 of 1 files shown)


Nanodragon999:
Just curious but, could you elaborate more on what this flag do?

ncw:
Apologies for the delay in responding, I've been on vacation.

I think this looks like an interesting change.

We'll need to make an integration test with the mirror flag on.

> So, a few things that will need to be decided on. Some of the functions really don't work for a multi-remote backend. For example, for the PutStream function: should we check that all remotes support PutStream first then perform operations or just perform the operation on all of them and return any errors? I'm leaning towards checking first because in this case, it'd be better to have a failure rather then files only written to some of the remotes.

For PutStream probably the best thing to do is in the initialization declare we support PutStream if all mirrors support it.  If we don't declare it then rclone will use a fallback path which will work for all backends.

In general trying to work everything out in the initialization will give the best results I think.

buengese:
@Nanodragon999 do you still intend to finish this? If not I would be willing to to the take over this sometime during the next week.

gary-kim:
> @gary-kim do you still intend to finish this? If not I would be willing to to the take over this sometime during the next week. (I'm interested in this feature too)

Sorry, my bad, this slipped my mind. I'm actually trying out a new IDE at the moment so I'll get back to this.

gary-kim:
@buengese I just remembered, I'm about to get really busy soon so feel free to take over. 
Thank you!

buengese:
Ok now that I had time to think a bit more about this I conclude there are quite a few considerations to be made for a proper mirror backend that I'd like to discuss.

1. I'm definitely going to separate this from the union backend keeping this together would only result in to separate codepaths for every single function. 
2. Regarding the available capabilities we can probably reuse some mechanism from the crypt backend.
3. There is the question of what to do with non writing operations. We could either always read always from the same backend or from random backend (or another strategy to balance the usage of the backends). In case we use a random backend there are again multiple options to implement this. Either we do the random selection in NewFs or we do it on each function. The latter would have the advantage of potentially balancing the load between backends. 
4. Do we do uploads sequentially or in parallel. This is especially important for PutStream given that or input is not guaranteed to be seekable.
5. If we use any other option than only reading from one backend in 3 we need to ensure that backends are already in sync when the mirror is created. We should probably compare them during configuration and ask the user to sync them/use empty dirs if they aren't identical.

@ncw, @gary-kim, @tsumare any thoughts on these?

gary-kim:
@buengese 
Your make a good point with the first point. Almost everything requires separate codepaths. Making another remote altogether doesn't seem like a bad idea.
For number 3, it may be worth it to just access the first remote in the list for now and add the load balance feature as an option later.
I'll come back in a bit to add on to this.

tsumare:
I would definitely suggest that there be at least the option to prefer a specified mirror (last? first?) for all read operations.  In the backup use case, I would like one of the backends to be Amazon S3's GLACIER storage type which is not readable without initiating a restore.

It might be worth considering checking either a random or specific mirror first, then falling back to others if the file is not present.  This eliminates some of the initial synchronization issue, even if it does not eliminate the possibility for conflicts.  In the conflict case, I believe all we can do is to pick one and hopefully be consistent in our choice.  Whether this is based on a filename hash for load balancing, or whether it is based on a preferred mirror doesn't seem too important to me.

Even if we enforce or warn about initial synchronization, we still should have this sort of consistent choice mechanism, as things could still fall out of sync later and we wouldn't want to mix reads.

It seems simplest to me to simply have a clearly defined preferred order of mirrors for all read-type operations.

PR #3350, opened 2019-07-15
Koofr backend: support for setting modification times

jkmcnk:

Support for setting modification times on Koofr backend.

No.

#### Checklist

### DIFF
--- a/backend/koofr/koofr.go
+++ b/backend/koofr/koofr.go
@@ -41,6 +41,12 @@ func init() {
 				Default:  "",
 				Advanced: true,
 			}, {
+				Name:     "setmtime",
+				Help:     "Does the backend support setting modification time. Set this to false if you use a mount ID that points to a Dropbox or Amazon Drive backend.",
+				Default:  true,
+				Required: true,
+				Advanced: true,
+			}, {
 				Name:     "user",
 				Help:     "Your Koofr user name",
 				Required: true,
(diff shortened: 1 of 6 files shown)


buengese:
Could you explain a bit more how this works? What if the user answers the setmtime option wrong? Could this be detected automatically?

jkmcnk:
> Could you explain a bit more how this works? What if the user answers the setmtime option wrong? Could this be detected automatically?

for this to be relevant, the user must meddle with the advanced options in the first place, to select a location that goes to a dropbox or amazon drive backed storage, and not to a Koofr-hosted one, so I don't find this such a problem. tbh, it could be detected automatically, but would require some network communication with the server just to return from the Precision() method: thus, I opted to avoid this.

buengese:
I see looks fine to be but I think we're gonna wait for ncw to return from vacation before merging this.

jkmcnk:
> I see looks fine to me but I think we're gonna wait for ncw to return from vacation before merging this.

ofc. we're in no hurry, but hopefully we get this merged in prior to next release.

ncw:
I think this looks sound to me.  Do the integration tests pass with and without the flag?

jkmcnk:
> I think this looks sound to me. Do the integration tests pass with and without the flag?
yes. except for the obviously wrong case of declaring that the backend supports setting mtime while directing the files to a dropbox backend connected via koofr.

ncw:
Great.  The code for this looks good - will merge now :-)

ncw:
Thank you very much

PR #3360, opened 2019-07-19
fichier: make FolderID int and adjust related code - fixes #3359

### LINKED ISSUE #3359: 1fichier: list error
Listing dirs from 1fichier 

rclone v1.48.0-067-g0a1169e6-beta
- os/arch: linux/amd64
- go version: go1.12.7

Linux 4.9.53-5.ph2-esx x86_64

1fichier

rclone lsd 1fichier:

2019/07/19 21:54:56 DEBUG : rclone: Version "v1.48.0-067-g0a1169e6-beta" starting with parameters ["rclone" "lsd" "1fichier:" "-vv"]
2019/07/19 21:54:56 DEBUG : Using config file from "/root/.config/rclone/rclone.conf"
2019/07/19 21:54:57 ERROR : : error listing: couldn't list folders: json: cannot unmarshal number into Go struct field Folder.pass of type string
2019/07/19 21:54:57 Failed to lsd with 2 errors: last error was: couldn't list folders: json: cannot unmarshal number into Go struct field Folder.pass of type string

buengese:
API now provides folder_id and pass as number which go doens't want to unmarshal into string so make them int instead. Requires a little more converting between string and int because of dir cache but not a big deal.

### DIFF
--- a/backend/fichier/api.go
+++ b/backend/fichier/api.go
@@ -88,7 +88,7 @@ func (f *Fs) listSharedFiles(ctx context.Context, id string) (entries fs.DirEntr
 	return entries, nil
 }
 
-func (f *Fs) listFiles(directoryID string) (filesList *FilesList, err error) {
+func (f *Fs) listFiles(directoryID int) (filesList *FilesList, err error) {
 	// fs.Debugf(f, "Requesting files for dir `%s`", directoryID)
 	request := ListFilesRequest{
 		FolderID: directoryID,
@@ -111,7 +111,7 @@ func (f *Fs) listFiles(directoryID string) (filesList *FilesList, err error) {
 	return filesList, nil
 }
 
-func (f *Fs) listFolders(directoryID string) (foldersList *FoldersList, err error) {
+func (f *Fs) listFolders(directoryID int) (foldersList *FoldersList, err error) {
 	// fs.Debugf(f, "Requesting folders for id `%s`", directoryID)
 
 	request := ListFolderRequest{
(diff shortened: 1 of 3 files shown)


backend/fichier/api.go
```diff
@@ -4,8 +4,8 @@ import (
 	"context"
 	"io"
 	"net/http"
-	"regexp"
 	"strconv"
+	"regexp"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"regexp"
	"strconv"
```

backend/fichier/structs.go
```diff
@@ -40,7 +40,7 @@ type GenericOKResponse struct {
 // MakeFolderRequest is the request structure of the corresponding request
 type MakeFolderRequest struct {
 	Name     string `json:"name"`
-	FolderID string `json:"folder_id"`
+	FolderID int `json:"folder_id"`
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	FolderID int    `json:"folder_id"`
```

ncw:
Great fix :-)

I was wondering - do we know the range of the integers?  If they are 64 bit then we should be using `int64` as the type.

The code as written at the moment will work fine with 64 bit platforms (where `int` is 64 bit) but could fail on 32 bit platforms where `int` is 32 bits.  There is a reducing number of these, but Raspberry Pi and NAS are both popular targets for rclone.

I see you've

buengese:
To be honest I kinda assumed FolderID would just be an incrementing counter unique to each user and no one would store 2 billion folders however this may not be the case. In my account no ID exceeded 2000000 but you are probably right. Didn't give it that much thought. While you're here you may also take a look at my other changes. Don't want to break anything in your absence.

ncw:
> To be honest I kinda assumed FolderID would just be an incrementing counter unique to each user and no one would store 2 billion folders however this may not be the case.

640k should be enough for everyone :-)

> In my account no ID exceeded 2000000 but you are probably right. Didn't give it that much thought. While you're here you may also take a look at my other changes. Don't want to break anything in your absence.

The underlying type of the JSON is floating point double, which can store integers to 2^56 or so so an int 64 is probably safest. However it may be completely unecessary... Are there any docs we could check?

The changes for this PR look excellent :-)

Issue #985, opened 2017-01-04
Total upload/time limits! [Feature Request]


doonze:
Loving rclone so far, but I would like to see a total session upload limit option. Pretty straight forward, a flag you can set to stop uploading at say 100GB. rclone will finish it's current uploads once that target is reached but will not take any more work, then close the session. Since it tracks total upload anyway, I think it would be a really easy thing to do. If no one else is interested, I can fork and write it for testing. But this should be a simple feature for someone already familiar with the code. I have a 1TB per month (soft) limit with my ISP, and I would rather keep them out of my business, so being able to control the amount uploaded would be nice. 

Also, while we're at it, a time based limit as well. I've seen that request before. If you could set something to only upload for 3 hours for example. That way with crontab (or the like) you could control the exact times your uploading. Should also be easy since total time is also already tracked. 
 
Biggest advantage is that currently if I want to cut it off, I have to of course kill it once my target time/size is reached. So I lose the current files it's uploading, and risk maybe leaving things in a funky state. I'm confident rclone will fix that on the next copy/sync, but it's a hacky way of imposing limits. I'd rather just be able to set them when I run it. With total limits, and bandwidth limits that are already in place, I could for example just set my big directory's to upload 750GB, at a low bandwidth limit, then just let it run all month till it hit the target. Then do it again next month. And I never have to worry about blowing my ISP's limits. 

Thoughts?

ncw:
Nice idea.  Fancy having a go?

doonze:
I can pull and fork it when I get a chance. Likely take me awhile to dig through the code, why I was hoping someone familiar with the code wanted to take a swing at it. Lol. But yeah, I'll take a look.

doonze:
Hmmm... Crash course in GO, still not sure of the logic flow your using, but I can at least read it now...

ncw:
Ok here is an outline of what you need to do...
  * in fs/config.go make a new flag (or flags) and stuff them into the Config struct
  * in fs/sync.go in the `run` method, you need to break out of the `Do the transfers` loop when your limits are reached.  Curernt transfers will finish neatly to completion if you do that, and there may be some objects in the pipeline, but that is the neatest way to do it.

Then write a test or two, do the docs and you are done!

doonze:
ah ha!! sync was the package I was not getting! I hasn't looked in there much cause I had figured it was like copy and others, just a package for that command. *Light bulb comes on* bbl.....

doonze:
[...]

PR #3371, opened 2019-07-23
Add a max session time limit

boosh:
It allows users to specify the maximum duration of a transfer session sort of as requested in #985, making it easier to script rclone to e.g. backup during the night. The setting I've added introduces a duration flag. After rclone has been running for that amount of time, no new transfers will be initiated, but those already started will be able allowed to complete. It may be useful to implement a further flag for a hard limit that will terminate in-progress transfers once the session limit is reached, but in another PR.

Note - Since `--max-transfer` doesn't appear to have tests, neither does this feature. I have tested the non-server copy manually though - I've never used server mode so would need some pointers to test that.

Sort of in #985

#### Checklist

### DIFF
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
@@ -102,7 +103,14 @@ func newSyncCopyMove(ctx context.Context, fdst, fsrc fs.Fs, deleteMode fs.Delete
 	if err != nil {
 		return nil, err
 	}
-	s.ctx, s.cancel = context.WithCancel(ctx)
+	// If a max session duration has been defined add a deadline to the context
+	if fs.Config.MaxDuration > 0 {
+		endTime := time.Now().Add(fs.Config.MaxDuration)
+		fs.Infof(s.fdst, "Transfer session deadline: %s", endTime.Format("2006/01/02 15:04:05"))
+		s.ctx, s.cancel = context.WithDeadline(ctx, endTime)
+	} else {
+		s.ctx, s.cancel = context.WithCancel(ctx)
+	}
 	if s.noTraverse && s.deleteMode != fs.DeleteModeOff {
 		fs.Errorf(nil, "Ignoring --no-traverse with sync")
 		s.noTraverse = false
@@ -194,6 +202,9 @@ outer:
 func (s *syncCopyMove) processError(err error) {
 	if err == nil {
 		return
+	}
+	if err == context.DeadlineExceeded {
+		err = fserrors.NoRetryError(err)
 	}
 	s.errorMu.Lock()
 	defer s.errorMu.Unlock()
@@ -741,6 +752,9 @@ func (s *syncCopyMove) run() error {
 		//delete empty subdirectories that were part of the move
 		s.processError(deleteEmptyDirectories(s.ctx, s.fsrc, s.srcEmptyDirs))
 	}
+
+	// Read the error out of the context if there is one
+	s.processError(s.ctx.Err())
 
 	// cancel the context to free resources
 	s.cancel()
--- a/fs/sync/sync_test.go
+++ b/fs/sync/sync_test.go
@@ -4,6 +4,7 @@ package sync
 
 import (
 	"context"
+	"fmt"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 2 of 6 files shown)


boosh:
OK thanks. Let me know if you want me to squash the commits.

boosh:
Ah it doesn't work if I move it to those places. I'll revert the change.

boosh:
`sync/run` is too high-level. I need to break during the transfer loop. Perhaps `moveOrCopy` didn't work because I tried syncing a directory? I expect that'd be the primary usecase since right now in-progress transfers aren't terminated once the session limit is reached. 

Is there anywhere else it needs adding to given it's mainly to help script backups (so several operations wouldn't be relevant)?

ncw:
Note that MaxTransfer does have tests here: fs/accounting/accounting_test.go :-)

I think that breaking the transfer in `Copy` is probably the wrong place.

What you want to do is [cancel the context](https://github.com/ncw/rclone/blob/a35aa1360e3e09d4973d52d995f7edaae803adb3/fs/sync/sync.go#L662) passed to the march routine.

I think using a deadline should work

```go
newCtx, cancel := context.WithDeadline(ctx, 100*time.Seconds) // config var goes here
// pass newCtx to march
// do the march
```

This will cause march routine to exit and cause the sync to come to a stop gracefully.

fs/sync/sync.go
```diff
@@ -16,6 +12,9 @@ import (
 	"github.com/ncw/rclone/fs/march"
 	"github.com/ncw/rclone/fs/operations"
 	"github.com/pkg/errors"
+	"path"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

boosh:
OK I've changed to add a deadline to the context. I've added a test but it'll be quite brittle. Is there any way of throttling transfers during testing to e.g. simulate large files, slow connections etc? At the moment I've just created 10000 test files. The test works on my mac today but obviously as computers get faster it'll break one day... It'd be better to only create a handful of files with a function that just makes each transfer sleep for a while to make them safely reproducible.

boosh:
Hmm the appveyor test failed - perhaps because it's on Windows? On my Mac no `context.DeadlineExceeded` is ever thrown so I can't require an error in the tests. I assume it's something to do with the test filesystem abstraction. For now I won't add an assertion to the error returned from `Sync` unless there's a cross-platform solution. The test makes sure that not all files were transferred so that should be good enough.

fs/sync/sync_test.go
```diff
@@ -977,6 +978,35 @@ func TestSyncWithUpdateOlder(t *testing.T) {
...
+	testFiles := make([]fstest.Item, 10000)
+
+	for i := 0; i < len(testFiles); i++ {
+		testFiles[i] = r.WriteFile(fmt.Sprintf("file%d", i), "***", t1)
+	}
+
+	fstest.CheckListing(t, r.Flocal, testFiles)
+
+	accounting.Stats.ResetCounters()
+	err = Sync(context.Background(), r.Fremote, r.Flocal, false)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

fs/sync/sync.go
```diff
@@ -657,6 +657,15 @@ func (s *syncCopyMove) run() error {
 
 	s.startTrackRenames()
 
+	// If a max session duration has been defined, change the context to have a deadline
+	if fs.Config.MaxSession > 0 {
+		endTime := accounting.Stats.GetStartTime().Add(fs.Config.MaxSession)
+		fs.Infof(s.fdst, "Transfer session deadline: %s", endTime.Format("2006/01/02 15:04:05"))
+		newContext, cancel := context.WithDeadline(s.ctx, endTime)
+		s.ctx = newContext
+		s.cancel = cancel
```
ncw: I don't think you want to overwrite the existing `s.ctx` or `s.cancel` as this context with deadline is just for March and its go-routines.

fs/sync/sync_test.go
```diff
@@ -977,6 +978,35 @@ func TestSyncWithUpdateOlder(t *testing.T) {
...
 
+// Test with a max transfer session
+func TestSyncWithMaxSession(t *testing.T) {
+	r := fstest.NewRun(t)
+	defer r.Finalise()
+
+	maxSession, err := time.ParseDuration("2s")
+	require.NoError(t, err)
+
+	fs.Config.MaxSession = maxSession
```
ncw: You should put defer in here to reset the parameter back to the default

```
defer func(){
    fs.Config.MaxSession = 0 // reset back to default
}()
```

fs/sync/sync_test.go
```diff
@@ -977,6 +978,35 @@ func TestSyncWithUpdateOlder(t *testing.T) {
...
+
+	testFiles := make([]fstest.Item, 10000)
```
ncw: Making 10000 files will kill the integration tests with other backends for certain! (These tests are run against all the backends once per day).

How about just using a few filesa nd temporarily setting a bwlimit to 1b/s or something like that?

ncw:
> Hmm the appveyor test failed - perhaps because it's on Windows?

I think that is because you didn't reset the deadline in your tests see the `deadline exceeded` messages.

> On my Mac no `context.DeadlineExceeded` is ever thrown so I can't require an error in the tests.

You should find the error in `fs.Stats.GetLastError()`

This should work on all platforms

fs/sync/sync_test.go
boosh: They're still flaky on my Mac - I thought I'd got down to only needing 5 test files but sometimes it can get through a few hundred :-(

boosh:
Am I missing some kind of initialisation?

ncw:
I merged a lot of things which changes the stats interface which has caused the breakage.

If you want me to finish it off then I can - it is very nearly there.

If you want to have a go yourself, then rebase it onto master and change the uses of `fs.Config.Stats` to `accounting.GlobalStats()`

boosh:
@ncw It'd be great if you could finish it off please. I'm running into problems rebasing. Thanks.

ncw:
> @ncw It'd be great if you could finish it off please. I'm running into problems rebasing. Thanks.

No problems, will do :-)

boosh:
@88lex not yet.

ncw:
I'm still working on this, it turns out it is much harder than I first thought!

88lex:
> I'm still working on this, it turns out it is much harder than I first thought!
> @88lex not yet.

Completely understood. For all good things one must wait.

ncw:
All fixed now!  I'll merge it for 1.51

PR #3373, opened 2019-07-24
Add a json format log

justina777:
#### What is the purpose of this change?
It allows to use a json format log to fit in log system such as logstash and sort of as requested in https://github.com/ncw/rclone/issues/2459. The setting I've added introduces a flag --use-json-log with boolean value. After enable the flag, the output of log of rclone changes to json format with fields level, msg, source and time. It may be useful to implement a further flag for control fields in json log.

#### Was the change discussed in an issue or in the forum before?
sort of https://github.com/ncw/rclone/issues/2459

#### Checklist

- [v] I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [v] I have added tests for all changes in this PR if appropriate.
- [v] I have added documentation for the changes if appropriate.
- [v] All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [v] I'm done, this Pull Request is ready for review :-)

PR #3384, opened 2019-07-29
add object and objectType in json logs

justina777:
improvements for json logs to indicate information more precisely.

https://github.com/rclone/rclone/pull/3373

#### Checklist

- [v] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [v] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [v] I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/fs/log.go
+++ b/fs/log.go
@@ -77,18 +77,17 @@ var LogPrint = func(level LogLevel, text string) {
 // LogPrintf produces a log string from the arguments passed in
 func LogPrintf(level LogLevel, o interface{}, text string, args ...interface{}) {
 	out := fmt.Sprintf(text, args...)
-	if o != nil {
-		out = fmt.Sprintf("%v: %s", o, out)
-	}
 
 	if Config.UseJSONLog {
 		fields := logrus.Fields{}
 		switch o.(type) {
 		case string:
 		default:
-			fields = logrus.Fields{
-				"object":     o,
-				"objectType": fmt.Sprintf("%T", o),
+			if o != nil {
+				fields = logrus.Fields{
+					"object":     fmt.Sprintf("%+v", o),
+					"objectType": fmt.Sprintf("%T", o),
+				}
 			}
 		}
 
@@ -107,6 +106,9 @@ func LogPrintf(level LogLevel, o interface{}, text string, args ...interface{}) 
 			logrus.WithFields(fields).Panic(out)
 		}
 	} else {
+		if o != nil {
+			out = fmt.Sprintf("%v: %s", o, out)
+		}
 		LogPrint(level, out)
 	}
 }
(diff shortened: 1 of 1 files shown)


ncw:
The logs for this look like

```json
{
  "level": "info",
  "msg": "README.md: Copied (replaced existing)",
  "object": {},
  "objectType": "*local.Object",
  "source": "operations/operations.go:401",
  "time": "2019-07-29T21:25:22.301502+01:00"
}
```

I think they should look like

```json
{
  "level": "info",
  "msg": "Copied (replaced existing)",
  "object": "README.md",
  "objectType": "*local.Object",
  "source": "operations/operations.go:401",
  "time": "2019-07-29T21:25:22.301502+01:00"
}
```

So I think you should do something like this (untested)

```go
// LogPrintf produces a log string from the arguments passed in
func LogPrintf(level LogLevel, o interface{}, text string, args ...interface{}) {
	out := fmt.Sprintf(text, args...)
	if Config.UseJSONLog {
		fields := logrus.Fields{
			"object":     fmd.Sprintf("%v", o),
			"objectType": fmt.Sprintf("%T", o),
		}
		switch level {
		case LogLevelDebug:
			logrus.WithFields(fields).Debug(out)
		case LogLevelInfo:
			logrus.WithFields(fields).Info(out)
		case LogLevelNotice, LogLevelWarning:
			logrus.WithFields(fields).Warn(out)
		case LogLevelError:
			logrus.WithFields(fields).Error(out)
		case LogLevelCritical:
			logrus.WithFields(fields).Fatal(out)
		case LogLevelEmergency, LogLevelAlert:
			logrus.WithFields(fields).Panic(out)
		}
	} else {
		if o != nil {
			out = fmt.Sprintf("%v: %s", o, out)
		}
		LogPrint(level, out)
	}
}
```

what do you think?

Do you think `o == nil` should have special handling?

justina777:
I updated the output format for object and added special handling for `o==null`.

fs/log.go
```diff
@@ -77,25 +77,38 @@ var LogPrint = func(level LogLevel, text string) {
 // LogPrintf produces a log string from the arguments passed in
 func LogPrintf(level LogLevel, o interface{}, text string, args ...interface{}) {
 	out := fmt.Sprintf(text, args...)
-	if o != nil {
-		out = fmt.Sprintf("%v: %s", o, out)
-	}
+
 	if Config.UseJSONLog {
+		fields := logrus.Fields{}
+		switch o.(type) {
```
ncw: This needs to do something for case string: doesn't it?  Maybe you don't need the type switch at all and can just do the same thing for all types?

justina777: Yes, it does. I thought that was to prevent redundant information in logs but it's fine to apply the format for all types. I'll update it.

ncw:
That looks great now - will merge.

Thank you very much - this is turning into a great feature :-)

PR #3336, opened 2019-07-09
Attempt to fix #3335

buengese:
This is an attemp to fix #3335.
With the current implementation march is unable to see subdirectories on a destination remote when used with no traverse. If these directories exist on the source they will be perceived as missing on the destination. The new march job that is created for srcOnly will then attempt to find the files contained in these directories in the root of the destination remote. I'm not entirely sure if this is intended behaviour but currently it causes unnecessary copies. I'm fixing this by setting dstRemote to be equal to srcRemote so that the correct path is checked even if it not necessarily exists. This will have no influence on runs without ``--no-traverse`` and fixes this unwanted behaviour. However in case that the path actually does not exist on destination those files will no longer be checked for in the root dir.

PR #3405, opened 2019-08-05
Significantly rework testing for march and fix checking subdirs when using --no-traverse

buengese:
Creating a test-case that would actually catch the no-traverse problem turned out to be a lot more work than I thought. In the end I decided to create to create 2 new test cases that actually invoke march in the same way syncCopyMove does. This requires a bit more helper code than I'm comfortable with but I don't really see a simpler way to catch these types of bugs. Also frustratingly the test framework didn't have a good way of comparing a DirEntries to a slice Items generated by working with the fake test remotes so I added one.  Closes #3336

### DIFF
--- a/fs/march/march.go
+++ b/fs/march/march.go
@@ -423,6 +423,7 @@ func (m *March) processJob(job listDirJob) ([]listDirJob, error) {
 		if recurse && job.srcDepth > 0 {
 			jobs = append(jobs, listDirJob{
 				srcRemote: src.Remote(),
+				dstRemote: src.Remote(),
 				srcDepth:  job.srcDepth - 1,
 				noDst:     true,
 			})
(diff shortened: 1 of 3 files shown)


ncw:
Ooops this PR fell off my radar sorry.

> Creating a test-case that would actually catch the no-traverse problem turned out to be a lot more work than I thought. In the end I decided to create to create 2 new test cases that actually invoke march in the same way syncCopyMove does. This requires a bit more helper code than I'm comfortable with but I don't really see a simpler way to catch these types of bugs.

I think you've done an excellent job with that and the coverage has increased from 33% to 80% for fs/march which is fantastic.

> Also frustratingly the test framework didn't have a good way of comparing a DirEntries to a slice Items generated by working with the fake test remotes so I added one. Closes #3336

Yes, you  would have though that would be there already!  It probably is already buried in some test or other...

Thank you for this - great work :-)

I'll merge it now.

PR #3410, opened 2019-08-05
Removing the shorthand symbol for rclone-web-gui

negative0:

This change removes the shorthand symbol w for rc-web-gui because it is reserved for rsync functionality.


ncw:
I've cherry picked the commit onto master (you put the old commit in here too!)

I'm just going to run that and a few other fixes through the CI then I'll send it to github

Issue #3130, opened 2019-04-24
Folder creation and deletion fails for Digital Ocean S3 spaces.


jpbakke:
#### What is the problem you are having with rclone?
When workin with S3 compatible digital ocean spaces, I cannot create (via mkdir) or delete (via purge, rmdir, rmdirs) folders.  If I use a sync or copy, the folder structure is correct.  If I purge a folder with files and subfolders, the files are deleted, but all the folders remain.  If I try to rmdir or rmdirs folders that are empty, I don't get any error but the folder are not removed.  If I manually use mkdir to create a new folder, that folder does not get created.  I've verified that I can create an delete folders without any issues in 'S3 Browser' (windows application) and using the linux s3cmd application.

#### What is your rclone version (output from `rclone version`)
1.47.0

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Linux Ubuntu 18.04 64-bit

####  Which cloud storage system are you using? (eg Google Drive)
Digital Ocean spaces

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
mkdir
purge
rmdir 
rmdirs

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp 
I've attached a full debug log of my results
[problem.txt](https://github.com/ncw/rclone/files/3113728/problem.txt)

The file sync and copy features work correctly.  But when I try to purge or delete a folder with contents, rclone leaves the folders behind no matter what I try.

-Jeff

jpbakke:
Further testing: the 'delete' command appears to fail for both folders (expected) AND a regular file.

jpbakke:
Install current beta version: v1.47.0-019-g3d475dc0-beta
Same problems... mkdir fails, rmdir fails.

ncw:
How did the folders get created?

S3 compatible storage systems don't actually have the concept of a folder.  They are key value stores.  By convention the key is a path name `dir/subdir/file` but there is no need to create a "directory" `dir/sub/`.

This works very well, but means you can't have empty directories.

To work around this, some tools create 0 length files called `dir/sub/` which are known as directory markers.

Rclone reads these directory markers as directories, but when you do `rmdir` it doesn't delete them as rclone's preferred mode of operation is not to use directory markers.

So if you upload things with rclone, it won't create directory markers (which saves objects and operations) and everything will work just fine.

jpbakke:
[...]

PR #3415, opened 2019-08-07
Fix TestFsListDirEmpty for bucket based remotes

yparitcher:

fix `TestFsListDirEmpty` to accept an error of `DirNotFound` on remotes that don't support empty directories like the `TestFsListDirNotFound` test.

add the remotes which support `EmptyDir` to the features doc.

I tested as best as i could with the integration tests on B2.
**please make sure all the bucket based remotes pass the integration tests before merging**

this came up while implementing B2 linkshare. #3392

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -629,6 +629,10 @@ func (f *Fs) list(ctx context.Context, dir string, recurse bool, prefix string, 
 		if err != nil {
 			return err
 		}
+		// Check that there are matching files
+		if len(response.Files) == 0 {
+			return fs.ErrorDirNotFound
+		}
 		for i := range response.Files {
 			file := &response.Files[i]
 			// Finish if file name no longer has prefix
@@ -1239,13 +1243,9 @@ func (f *Fs) PublicLink(ctx context.Context, remote string) (link string, err er
 	}
 	_, err = f.NewObject(ctx, remote)
 	if err == fs.ErrorObjectNotFound || err == fs.ErrorNotAFile {
-		err2 := f.list(ctx, remote, false, "", 1, f.opt.Versions, func(remote string, object *api.File, isDirectory bool) error {
-			err = nil
-			return nil
+		err = f.list(ctx, remote, false, "", 1, f.opt.Versions, func(remote string, object *api.File, isDirectory bool) error {
+			return errEndList
 		})
-		if err2 != nil {
-			return "", err2
-		}
 	}
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


backend/b2/b2.go
```diff
@@ -629,6 +629,10 @@ func (f *Fs) list(ctx context.Context, dir string, recurse bool, prefix string,
 		if err != nil {
 			return err
 		}
+		// Check that there are matching files
+		if len(response.Files) == 0 {
```
ncw: I'm unsure about this...

It potentially seems like a good idea (and one which could go in all the Bucket based remotes), however will it break stuff?

How does `rclone mount` behave with this patch if you make a new directory then list it for b2?

If we want all the bucket based remotes to behave like this then we should add an another integration test in fstest/fstests/fstests.go to list a non existent directory and make sure we get an error.



https://github.com/rclone/rclone/issues/3427

yparitcher: i have never used `rclone mount` so i am not sure about its behavior, will test later today.

as i have said before, i think the bucket based remotes should either support empty folders all the way (via hidden files, as B2 and S3 already do) or not support them at all (`ErrorDirNotFound`).
the current way of appearing to support but silently not supporting is leading to a lot of confusion. ( as seen in #3427 & #3130)

whichever way you decide to go will need integration with the rest of rclone. this might need some broader fixing. (mount & --create-empty-src-dirs to start)

i can back out that commit for now and open it separately for a general PR to fix BBR whichever way you decide. however please merge the test & doc change.
also the current bugs/limitations docs imply that rclone does not create empty dirs, while to the user it appears that is does.

please let me know what you think

yparitcher:
fixed empty bucket (root can be an empty dir)

backend/b2/b2.go
yparitcher: i ran a brief test with `mount -vv` and besides for debug errors (which i also got for `.Trash` and other system files my file browser was looking for) it seemed to work fine, created and removed directory, add and delete files

ncw:
OK what I've done is I've cherry pickets the tests patch and the docs patch.

I think we should do the list empty directory => DirNotFound for all BBR in a new issue which can include your b2 changes.  I think that needs a new integration test.

I didn't take the b2 changes as I'm just about to merge a very large change to all the BBR which allows them to work from the root so you don't have to pass in a bucket/container for read/write access.  This is something I've been meaning to do for ages and it regularized rclone's backends - previous to that there were all sorts of things you couldn't do to a BBR.

One advantage of that reworking is that all the listing routines now look extremely similar so changing them all should be straight forward!

yparitcher:
> OK what I've done is I've cherry pickets the tests patch and the docs patch.
> 
perfect

> I think we should do the list empty directory => DirNotFound for all BBR in a new issue which can include your b2 changes. I think that needs a new integration test.
> 
i will close this and open a new meta issue for BBR & empty directories

PR #3440, opened 2019-08-14
backend/local: fadvise run syscall on a dedicated go routine

mmatczuk:

It relates to #3419 and removes fadivise syscall from hot read path.

Before we issued an additional syscall periodically on a hot path.
This patch offloads the fadvise syscall to a dedicated go routine.

#### Checklist

### DIFF
--- a/backend/local/fadvise_unix.go
+++ b/backend/local/fadvise_unix.go
@@ -45,6 +45,35 @@ type fadvise struct {
 	lastPos    int64
 	curPos     int64
 	windowSize int64
+
+	freePagesCh chan offsetLength
+	doneCh      chan struct{}
+}
+
+type offsetLength struct {
+	offset int64
+	length int64
+}
+
+const (
+	defaultAllowPages      = 32
+	defaultWorkerQueueSize = 64
+)
+
+func newFadvise(o *Object, fd int, offset int64) *fadvise {
+	f := &fadvise{
+		o:          o,
+		fd:         fd,
+		lastPos:    offset,
+		curPos:     offset,
+		windowSize: int64(os.Getpagesize()) * defaultAllowPages,
+
+		freePagesCh: make(chan offsetLength, defaultWorkerQueueSize),
+		doneCh:      make(chan struct{}),
+	}
+	go f.worker()
+
+	return f
 }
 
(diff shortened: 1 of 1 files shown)


backend/local/fadvise_unix.go
```diff
@@ -77,12 +100,36 @@ func (f *fadvise) freePagesIfNeeded() {
...
+	default:
+		fs.Debugf(f.o, "Blocked on queueing pages to clean for file descriptor %d", f.fd)
+		f.freePagesCh <- p
 	}
+
 	f.lastPos = f.curPos
 }
 
+func (f *fadvise) worker() {
+	for {
```
ncw: this could be like this to simplify slightly

```go
for p := range f.freePagesCh {
    if err := ...
```

backend/local/fadvise_unix.go
```diff
@@ -77,12 +100,36 @@ func (f *fadvise) freePagesIfNeeded() {
 }
 
 func (f *fadvise) freePages() {
-	if err := unix.Fadvise(f.fd, f.lastPos, f.curPos-f.lastPos, unix.FADV_DONTNEED); err != nil {
-		fs.Debugf(f.o, "fadvise dontneed failed on file descriptor %d: %s", f.fd, err)
+	p := offsetLength{f.lastPos, f.curPos - f.lastPos}
+	select {
+	case f.freePagesCh <- p:
+	default:
+		fs.Debugf(f.o, "Blocked on queueing pages to clean for file descriptor %d", f.fd)
```
ncw: I discovered that for certain type of copy this debug was quite noisy!  Do we need it?  Can we just write to the channel not in a select?

Maybe we should make the channel a little bigger - maybe 16 entries?

mmatczuk: After experimentation I'd suggest sth like 64. I can sure skip the log.

ncw: SGTM

mmatczuk:
@ncw PTAL

ncw:
That is great, thank you I'll merge it now.

And than you very much with your help with performance analysis - I haven't spent very much time doing that so it is great that you have!

PR #3442, opened 2019-08-14
Refactor jottacloud backend

buengese:
As already announced yesterday I decide to refactor the jottacloud backend a bit. Mostly splitting the huge config function into 3 logical parts but also update some errors and and some internal documentation.

I also decided to remove username from the jottacloud config options. While we still need it to get our authentication token it doesn't make sense to store it beyond this point and removing it also removes the somewhat misleading cmdline option.

Because I removed a config option for jottacloud I ran `make backenddocs` which turned out to update a bunch of other backend docs too.


docs/content/sftp.md
```diff
@@ -187,7 +187,7 @@ SSH host to connect to
 
 #### --sftp-user
 
-SSH username, leave blank for current username, ncw
```
ncw: Ha!

We set USER in the makefile to try to avoid this but it obviously isn't working!

buengese:
Ah that makes sense. I removed all non Jottacloud doc updates. I also caught a tiny bug make backenddocs which caused the advanced options to be categorized as standard options if there were no standard options.

ncw:
LGTM please merge :-)

Issue #3420, opened 2019-08-08
Copyurl command get file name from server


nishkalprakash:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.48.0-132-gd0c65b4c-beta
- os/arch: windows/amd64
- go version: go1.12.3

#### What problem are you are trying to solve?
Copyurl could be able to get the file name from the server.

The url looks something like:
http://example.com/download.php?id=4460

On calling the url, it returns a status 302 and redirects to:
http://example.com/filename.txt?verify=1565166562-3oMSduA%2FXDukfrcavrhUdJn0FFN%2FlxKxmRxoJ2zQa8%3D

so if the following command was used:
rclone copyurl "http://example.com/download.php?id=4460" gd:folder1/

the file `filename.txt` would get saved in folder `folder1`.

#### How do you think rclone should be changed to solve that?
Maybe it could be made, if a destination ending in / was passed, rclone would use the leaf name of the file in the URL (after any redirection).

kfur:
@nishkalprakash  good proposal
but how do u think what must be expected behavior must be in case when url doesn't contain filename ?

nishkalprakash:
Post redirect if the url doesn't contain a file name then rclone can fallback to the same mechanism where a filename needs to be provided.
Maybe something like:
`rclone copyurl :http: --http-url http://example.com?id=123 --default-name file1.txt`

Here, lets suppose:
If the site redirects to:
http://example.com/**ActualFile.txt**?idtoken=123ab123ef
Then rclone saves the url as `ActualFile.txt`
else 
saves the file as `file1.txt`

if rclone cant find a name in the end and --default-name is missing then maybe an error can be shown:
ERROR - Unable to find a file name, please provide --default-name to avoid this error...
or something

nishkalprakash:
maybe we can just add a new flag to the existing `copyurl`, `--get-name-from-url` ...
`rclone` tries to get the name from the url after the last redirect, 
else does the same thing which rclone currently does with `copyurl`

kfur:
it was resolved i think

PR #3451, opened 2019-08-17
copyurl: add --auto-filename flag for using file name from url in destination path

kfur:
Add flag for `copyurl` command to allow users don't specify file name in destination path.

#3420

#### Checklist

### DIFF
--- a/cmd/copyurl/copyurl.go
+++ b/cmd/copyurl/copyurl.go
@@ -4,12 +4,18 @@ import (
 	"context"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	autoFilename = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefintion)
+	commandDefintion.Flags().BoolVarP(&autoFilename, "auto-filename", "a", autoFilename, "Get the file name from the url and use it for destination file path")
 }
 
 var commandDefintion = &cobra.Command{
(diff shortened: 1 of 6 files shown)


cmd/copyurl/copyurl.go
```diff
@@ -2,14 +2,19 @@ package copyurl
...
 	"context"
-
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	GetFileNameFromURL bool
```
golangcibot: exported var `GetFileNameFromURL` should have comment or be unexported (from `golint`)

fs/operations/rc.go
```diff
@@ -214,7 +214,7 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 		if err != nil {
 			return nil, err
 		}
-		_, err = CopyURL(ctx, f, remote, url)
+		_, err = CopyURL(ctx, f, remote, url, false)
```
ncw: We should add another parameter to here (see "rmdirs" just above) which defaults to false (just call in.GetBool and ignore the error)

It should be called the same name as the flag so `autoFilename` and needs to be documented in the copyurl docs (also in this file).

In an ideal world we'd add a bit more in `rc_test.go` in the `TestRcCopyurl` test.

fs/operations/rc_test.go
```diff
@@ -108,12 +108,34 @@ func TestRcCopyurl(t *testing.T) {
...
+		"autoFilename": false,
 	}
 	out, err := call.Fn(context.Background(), in)
 	require.NoError(t, err)
 	assert.Equal(t, rc.Params(nil), out)
 
-	fstest.CheckListingWithPrecision(t, r.Fremote, []fstest.Item{file1}, nil, fs.ModTimeNotSupported)
+	urlFileName := "filename.txt"
+	in = rc.Params{
+		"fs":     r.FremoteName,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

fs/operations/rc.go
kfur: Fixed. But it seems i don't understand difference between `rc_test.go` and `operations_test.go`, so maybe i something did wrong here.

fs/operations/rc.go
```diff
@@ -214,7 +214,12 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 		if err != nil {
 			return nil, err
 		}
-		_, err = CopyURL(ctx, f, remote, url)
+		autoFilename, err := in.GetBool("autoFilename")
+		if err != nil {
```
ncw: This makes the parameter required - we want to allow it to be missing for backwards compatibility, so this should be OK.

```
autoFilename, _ := in.GetBool("autoFilename")
```

fs/operations/rc.go
ncw: The rc_test.go tests the operation via the remote control interface which allows us to check the parameters are correct, etc.

fs/operations/rc.go
kfur: Fixed.

ncw:
Apologies for the delay.

That looks really good now - thank you.  Will merge :-)

PR #3457, opened 2019-08-19
mount: Fix typo in argument checking

pjw91:

The two `if` should check against `mountlib.ExtraOptions` and `mountlib.ExtraFlags` respectively, instead of checking `mountlib.ExtraOptions` twice.


ncw:
Good catch!

Thanks for the fix :-)

Issue #3423, opened 2019-08-09
Command genautocomplete output to stdout


pataquets:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.47.0-DEV

#### What problem are you are trying to solve?
Scripting the `genautocomplete` installation

#### How do you think rclone should be changed to solve that?
Just adding a `--stdout` switch or similar to command

ncw:
How about writing to stdout if the output file was `-`?

Fancy sending a PR for that?

pataquets:
I think my go-fu is not quite sufficient, although I didn't tried it. Just dropped the idea and moved on, since it's marginally useful for me, atm.
I've even tried to build rclone from source inside a Dockerfile (to send it a PR later) and failed at it, so I've left it for until my skills improve. I'm pretty sure it's quite standard build, but not enough for my knowledge and a couple how-to s out there 
 Any build advice?

ncw:
These instruction should work: https://rclone.org/install/#install-from-source

lasombra:
@ncw I'll take a look at this one, if you don't mind

Cnly:
Closing as the relevant PR has already been merged.

Issue #3029, opened 2019-03-06
 Failed to copy: can't copy - source file is being updated (mod time changed from 2018-07-03 15:14:11.819130985 -0400 EDT to 2018-07-03 15:14:11.821105924 -0400 EDT)


MohanaPriyaNarapareddy:

ncw:
Are you using glusterfs maybe?  This doesn't have completely consistent local timestamps.

You can use this to fix

      --local-no-check-updated           Don't check to see if the files change during upload

ncw:
I'm going to close this now as I haven't heard from you for a while.

crisanto-cc:
Solucionado, Gracias

Dryrtan:
Solved my problem, thanks

PR #3459, opened 2019-08-20
Add Docker workflow support

pataquets:
Add Dockerfile to enable image building.
Using the official Golang image, latest tag. More info at https://hub.docker.com/_/golang/

Just adding files, setting working dir and running make instructions.

Build:

```
$ docker build -t rclone .
```

Run:

```
$ docker run --rm -it [-v /host/path/rclone.conf/:/root/.rclone/rclone.conf] rclone [options...]
```

You might need to mount your host `rclone.conf` file inside the Docker container (`-v`switch), or bake it as a child image. Docker allows lots of workflows/usages.
I'm already using dockerized rclone, with a few aliases, thus not having to install it.
Also this would enable a quick start for newcome hackers.
Using this, I've been able to create Docker Compose manifests to quickly deploy DLNA and WebDAV servers using the `serve` command. I can send them also, if you're interested.


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang
```
ncw: Doesn't this make a container which contains the go compiler and the go source code?

That is more than is needed to run rclone, rclone can run from the scratch container, but ideally would have some SSL roots in too.

There is an example of how to do that here (scroll down to Dockerfile.multistage): https://www.callicoder.com/docker-golang-image-container-example/

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+COPY . /go/src/github.com/rclone/rclone/
+WORKDIR /go/src/github.com/rclone/rclone/
+
+RUN go build -v
```
ncw: Ideally this would use `make` rather than `go build` - then the git build tag etc will appear when you do `rclone version`.

There might be a more docker way of doing this as it is easy to express in just go commands.

It would be nice to run some tests here too `make quicktest` takes a couple of minutes and tests all the important things. If that fails then the build should fail.

Dockerfile
pataquets: Yes, it does.
I've used multi-stage builds in some cases, where size is priority. However, it adds [a slighty] more time to the build and Dockerfile complexity. Most of time, the tradeoff in the Dockerfile is the convenience as a developer tool. Usually developers focus on ease of use, efficency in repeating the build a lot and having debug tools at hand, whereas distribution usage priorize image size.
I've done it this way, since it's the middle spot, IMO. Since, I'm not as familiar with Golang development to have all considerations in mind, I can switch to a multi-stage build if you think it's better.

Dockerfile
ncw: > There might be a more docker way of doing this as it is easy to express in just go commands.

For instance if the build version came in as an environment variable it would be easy to use.

Dockerfile
ncw: I think a multi-stage build would be my preference.  The golang official image is 800 MB wheras a bare one should be ~10MB which is enough of a win in my mind to make it worth doing :-)

pataquets: Pushed new multi-staged build Dockerfile.
Tested it, and the single-stage vs ms build is (29s vs 33s), not a big deal.
We'll see if it makes local development workflow much worse or not. I can't say for sure.
Hopefully, I'll be able to try it myself with some easy issue, if time permits. :)

Dockerfile
pataquets: I was following the advice you gave me at #3423. Also, I wasn't aware there was a Makefile.
There is no single 'Docker Way' to build any software. I've dockerized software several times using make-based builds, and it's OK to do so, as long as it works.
I would even add that, the more similar all builds are done, the better, since you just want to make the build as reproduceable as possible and the Docker binary to be the same as with other packaging methods. So, if the build tool is `make`, we can switch to it. Whatever makes more sense to you.

About the tests, I'm in for testing builds. Sometime I test at least the most critical parts of the build, to make the build fail loudly. The caveats I see are:
* We have to ensure we cleanup after the tests, to keep image size.
* If it will not add too much time to the build, which is especially relevant for development workflow. The `make` tool is already included in the Golang image.
EDIT: The above caveats do not not apply with multi-stage build, but listing here for completeness.

Also, I've checked and the tests run, but fail. I've tried to run them in the v.1.48.0 tag, and they also fail. If there is something to fix before, or adding them is more involved, it can be left as a future iteration, too. This way, we can tackle #3460 and get familiar with the Automated Builds workflow too ;)

ncw: How did the tests fail? Can you paste the failure?  If I had to guess I'd say it would be the mount tests failing to make a FUSE mount in the container.

pataquets: Tried reproducing the failure, but now they run fine every time.
They run fine both before and after build. Which is preferred?

ncw: I normally do the tests before the build.

Are you ready to send the file?

pataquets: Added tests, they pass all the time.
Also, had to statically link the binary, do it won't fail on the Alpine container.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM golang AS builder
+
+COPY . /go/src/github.com/rclone/rclone/
+WORKDIR /go/src/github.com/rclone/rclone/
+
+RUN make quicktest
+RUN \
```
ncw: Can you run `make` instead of `go build -v` does that work?

That will build the correct tag into `rclone version`. You might need CGO_ENABLED=0 too.

That will also strip some debug symbols out of the binary too which make no difference unless you attach a debugger.

pataquets: It works, indeed.
Pushed.

ncw:
That looks great now - I'll merge that now.  Thank you very much :-)

I'll add some docs and makefile targets in a bit!

pataquets:
Great! Happy to give back for such a useful tool.
Point is, I wonder if with the build changes we've not deviated too much of the original issue's intent, at least from my PoV.
My original intent was just placing together the pieces to allow a quick jumpstart into hacking rclone, as per the linked issue #3029. However, looks like we've leaned more in favour of a new distribution/packaging method (once #3460 is done), which is also helpful. I can't tell as of now, let's see how it works, and we'll iterate, if needed.
I've already created an [Automated Build](https://hub.docker.com/r/pataquets/rclone-src/) (from my forked repo). Until now, I was using [an image built via `go get`](https://hub.docker.com/r/pataquets/rclone/), which worked fine, but needed manual triggers to keep up to date.
Tested the AB image with simple Docker Compose manifest I have to deploy a WebDAV server for one of my remotes using `rclone serve` and it works.
Next step would be IMO, following up #3460 and leaving it as a "devel" build by now and see how it works and how people use it or what feedback we get.

PR #3462, opened 2019-08-20
Autologin for web-gui and prefix patch for rcd

negative0:

Auto-login makes use of login_token to directly enable rclone-webui-react to login the user without entering any credentials in the browser.

Prefix patch for rcd enables user to use --rc-prefix command with web-gui


ncw:

Will merge now (apologies for the delay - a bit busy at work!)

PR #3466, opened 2019-08-21
Job id setting

mmatczuk:

The goal is to initialise `jobID` in `fs/rc/jobs/`. The intended use case is to always start with a new `jobID` to avoid conflicts upon restarts of rclone.  

No, I consider it low risk and the patch is really small.

#### Checklist

### DIFF
--- a/fs/rc/jobs/job.go
+++ b/fs/rc/jobs/job.go
@@ -59,6 +59,13 @@ func SetOpt(opt *rc.Options) {
 	running.opt = opt
 }
 
+// SetInitialJobID allows for setting jobID before starting any jobs.
+func SetInitialJobID(id int64) {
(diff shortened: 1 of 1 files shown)


mmatczuk:
No there is not, the point is that it's exported.

mmatczuk:
Theoretically it could be exported as a flag but I'm not sure it adds much value to the cli.

ncw:
OK, I'll merge it since it seems harmless enough :-)

mmatczuk:
Thanks.

davepallot:
This is very useful. How do you invoke this feature?

ncw:
@davepallot this interface is only for calling when using rclone as a Go library. It could have a flag easily enough though.

Issue #2582, opened 2018-09-15
box: Implement JWT and “as-user”


ncw:
Originally from [this forum thread](https://forum.rclone.org/t/box-com-capped-upload/6761).

- What are JWT: https://jwt.io/
- There is a go library for JWT: https://github.com/dgrijalva/jwt-go 
- https://developer.box.com/docs/setting-up-a-jwt-app
- https://developer.box.com/reference#as-user-1
- https://developer.box.com/docs/construct-jwt-claim-manually
-  https://medium.com/@daveym/authentication-fun-with-the-box-com-api-and-go-91901f2066cd
- https://github.com/daveym/box-reporter/blob/master/boxapi/client.go

> It’s actually easier to use than OAuth2, as you can get a json file from box which contains all the necessary parts, which you can easily parse. The only difference to OAuth2 is there is no refresh token, you just send another claim to get a new access token. This is why it works with multiple clients as the token only expires by explicit invalidation or expiration.

I think this end up with two extra config parameters
- `jwt` - a path to a jwt file from box
-  `impersonate`- a user id/name? to impersonate for the `as-user` field.

This will then bring it up to par with drive service accounts and drive impersonate.

----

From jmfrank63 in the forum.

JWT is widely supported.
- Amazon supports it via Amazon Cognito: https://aws.amazon.com/cognito/
- Google Cloud Storage supports it: https://www.cloudberrylab.com/blog/explaining-google-cloud-storage-authentication/
- Auth0 support it: https://auth0.com/docs/jwt
- Azure supports it as well: https://docs.microsoft.com/en-us/azure/api-management/policies/authorize-request-based-on-jwt-claims

Zapier has a good article when to use what kind of authentication: https://zapier.com/engineering/apikey-oauth-jwt/

In general OAuth2 is a great way for humans to authenticate, but if you want passwordless access JWT is a great extension to OAuth2.

Simplified you could say it is working similar to an ssh key authentication, with you holding the private key and the other side holding the public key.

...

I have a sneaky suspicion that the oauth2 library supports this already.  It certainly does for google drive and google cloud storage and we use this support already.

ncw:
This is now implemented and was released in v1.50

PR #3467, opened 2019-08-21
Allow JWT auth with Box

dbramwell:

Allow passing in a Box config.json file when configuring a Box remote. Also potentially paves the way for JWT auth in other apps.

[2582](https://github.com/rclone/rclone/issues/2582)

#### Checklist

Almost...

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -11,8 +11,12 @@ package box
 
 import (
 	"context"
+	"crypto/rsa"
+	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
@@ -20,6 +24,10 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/rclone/rclone/lib/jwtutil"
+
+	"github.com/youmark/pkcs8"
 
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/backend/box/api"
@@ -29,12 +37,14 @@ import (
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/config/obscure"
 	"github.com/rclone/rclone/fs/fserrors"
+	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/lib/dircache"
 	"github.com/rclone/rclone/lib/oauthutil"
 	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/rest"
 	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/jws"
 )
 
 const (
@@ -49,6 +59,7 @@ const (
 	listChunks                  = 1000     // chunk size to read directory listings
 	minUploadCutoff             = 50000000 // upload cutoff can be no lower than this
 	defaultUploadCutoff         = 50 * 1024 * 1024
+	tokenURL                    = "https://api.box.com/oauth2/token"
 )
 
 // Globals
@@ -73,9 +84,34 @@ func init() {
 		Description: "Box",
 		NewFs:       NewFs,
 		Config: func(name string, m configmap.Mapper) {
-			err := oauthutil.Config("box", name, m, oauthConfig)
-			if err != nil {
-				log.Fatalf("Failed to configure token: %v", err)
+			jsonFile, ok := m.Get("box_config_file")
+			boxSubType, boxSubTypeOk := m.Get("box_sub_type")
+			var err error
+			if ok && boxSubTypeOk && jsonFile != "" && boxSubType != "" {
+				boxConfig, err := getBoxConfig(jsonFile)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				privateKey, err := getDecryptedPrivateKey(boxConfig)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				claims, err := getClaims(boxConfig, boxSubType)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				signingHeaders := getSigningHeaders(boxConfig)
+				queryParams := getQueryParams(boxConfig)
+				client := fshttp.NewClient(fs.Config)
+				err = jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m, client)
+				if err != nil {
+					log.Fatalf("Failed to configure token with jwt authentication: %v", err)
+				}
+			} else {
+				err = oauthutil.Config("box", name, m, oauthConfig)
+				if err != nil {
+					log.Fatalf("Failed to configure token with oauth authentication: %v", err)
+				}
 			}
 		},
 		Options: []fs.Option{{
@@ -84,6 +120,19 @@ func init() {
 		}, {
 			Name: config.ConfigClientSecret,
 			Help: "Box App Client Secret\nLeave blank normally.",
+		}, {
+			Name: "box_config_file",
+			Help: "Box App config.json location\nLeave blank normally.",
+		}, {
+			Name:    "box_sub_type",
+			Default: "user",
+			Examples: []fs.OptionExample{{
+				Value: "user",
+				Help:  "Rclone should act on behalf of a user",
+			}, {
+				Value: "enterprise",
+				Help:  "Rclone should act on behalf of a service account",
+			}},
 		}, {
 			Name:     "upload_cutoff",
 			Help:     "Cutoff for switching to multipart upload (>= 50MB).",
@@ -96,6 +145,74 @@ func init() {
 			Advanced: true,
 		}},
 	})
+}
+
+func getBoxConfig(configFile string) (boxConfig *api.ConfigJSON, err error) {
+	file, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "box: failed to read Box config")
+	}
+	err = json.Unmarshal(file, &boxConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "box: failed to parse Box config")
+	}
+	return boxConfig, nil
+}
+
+func getClaims(boxConfig *api.ConfigJSON, boxSubType string) (claims *jws.ClaimSet, err error) {
+	val, err := jwtutil.RandomHex(20)
+	if err != nil {
+		return nil, errors.Wrap(err, "box: failed to generate random string for jti")
+	}
+
+	claims = &jws.ClaimSet{
+		Iss: boxConfig.BoxAppSettings.ClientID,
+		Sub: boxConfig.EnterpriseID,
+		Aud: tokenURL,
+		Iat: time.Now().Unix(),
+		Exp: time.Now().Add(time.Second * 45).Unix(),
+		PrivateClaims: map[string]interface{}{
+			"box_sub_type": boxSubType,
+			"aud":          tokenURL,
+			"jti":          val,
+		},
+	}
+
+	return claims, nil
+}
+
+func getSigningHeaders(boxConfig *api.ConfigJSON) *jws.Header {
+	signingHeaders := &jws.Header{
+		Algorithm: "RS256",
+		Typ:       "JWT",
+		KeyID:     boxConfig.BoxAppSettings.AppAuth.PublicKeyID,
+	}
+
+	return signingHeaders
+}
+
+func getQueryParams(boxConfig *api.ConfigJSON) map[string]string {
+	queryParams := map[string]string{
+		"client_id":     boxConfig.BoxAppSettings.ClientID,
+		"client_secret": boxConfig.BoxAppSettings.ClientSecret,
+	}
+
+	return queryParams
+}
+
+func getDecryptedPrivateKey(boxConfig *api.ConfigJSON) (key *rsa.PrivateKey, err error) {
+
+	block, rest := pem.Decode([]byte(boxConfig.BoxAppSettings.AppAuth.PrivateKey))
+	if len(rest) > 0 {
+		return nil, errors.Wrap(err, "box: extra data included in private key")
+	}
+
+	rsaKey, err := pkcs8.ParsePKCS8PrivateKey(block.Bytes, []byte(boxConfig.BoxAppSettings.AppAuth.Passphrase))
+	if err != nil {
+		return nil, errors.Wrap(err, "box: failed to decrypt private key")
+	}
+
+	return rsaKey.(*rsa.PrivateKey), nil
 }
 
 // Options defines the configuration for this backend
--- a/docs/content/box.md
+++ b/docs/content/box.md
@@ -12,7 +12,8 @@ Paths are specified as `remote:path`
 Paths may be as deep as required, eg `remote:directory/subdirectory`.
 
 The initial setup for Box involves getting a token from Box which you
-need to do in your browser.  `rclone config` walks you through it.
+can do either in your browser, or with a config.json downloaded from Box
+to use JWT authentication.  `rclone config` walks you through it.
 
 Here is an example of how to make a remote called `remote`.  First run:
 
@@ -37,7 +38,14 @@ Storage> box
 Box App Client Id - leave blank normally.
 client_id> 
 Box App Client Secret - leave blank normally.
-client_secret> 
+client_secret>
+Box App config.json location
+Leave blank normally.
+Enter a string value. Press Enter for the default ("").
+config_json>
+'enterprise' or 'user' depending on the type of token being requested.
+Enter a string value. Press Enter for the default ("user").
+box_sub_type>
 Remote config
 Use auto config?
  * Say Y if not sure
--- a/lib/jwtutil/jwtutil.go
+++ b/lib/jwtutil/jwtutil.go
@@ -1,0 +1,87 @@
+package jwtutil
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/hex"
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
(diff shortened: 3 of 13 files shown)


backend/box/box.go
```diff
@@ -73,9 +84,20 @@ func init() {
...
-				log.Fatalf("Failed to configure token: %v", err)
+			jsonFile, ok := m.Get("config_json")
+			boxSubType, boxSubTypeOk := m.Get("box_sub_type")
+			if ok && boxSubTypeOk && jsonFile != "" && boxSubType != "" {
+				boxConfig := getBoxConfig(jsonFile)
+				privateKey := getDecryptedPrivateKey(boxConfig)
+				claims := getClaims(boxConfig, boxSubType)
+				signingHeaders := getSigningHeaders(boxConfig)
+				queryParams := getQueryParams(boxConfig)
+				jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m)
```
golangcibot: Error return value of `jwtutil.Config` is not checked (from `errcheck`)

backend/box/box.go
```diff
@@ -98,6 +127,90 @@ func init() {
...
 }
 
+func getBoxConfig(configFile string) ConfigJSON {
+	jsonData, err := os.Open(configFile)
+	if err != nil {
+		log.Fatalf("Failed to read Box config: %v", err)
+	}
+	byteValue, _ := ioutil.ReadAll(jsonData)
+	var boxConfig ConfigJSON
+	json.Unmarshal(byteValue, &boxConfig)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,78 @@
...
+	client := &http.Client{}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Fatalf("Failed making auth request: %v", err)
+	}
+	if resp.StatusCode != 200 {
+		log.Fatalf("Failed making auth request: %v", resp.Status)
+	}
+	defer resp.Body.Close()
```
golangcibot: Error return value of `resp.Body.Close` is not checked (from `errcheck`)

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,78 @@
...
+	return hex.EncodeToString(bytes), nil
+}
+
+// Config configures rclone using JWT
+func Config(id, name string, claims *jws.ClaimSet, header *jws.Header, queryParams map[string]string, privateKey *rsa.PrivateKey, m configmap.Mapper) (*oauth2.Token, error) {
+	payload, err := jws.Encode(header, claims, privateKey)
+	if err != nil {
+		log.Fatalf("Failed to encode payload: %v", err)
+	}
+	req, err := http.NewRequest("POST", claims.Aud, nil)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,78 @@
...
+	q := req.URL.Query()
+	q.Add("grant_type", "urn:ietf:params:oauth:grant-type:jwt-bearer")
+	q.Add("assertion", payload)
+	for key, value := range queryParams {
+		q.Add(key, value)
+	}
+	queryString := q.Encode()
+
+	req, err = http.NewRequest("POST", claims.Aud, bytes.NewBuffer([]byte(queryString)))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/box/box.go
```diff
@@ -84,6 +107,13 @@ func init() {
 		}, {
 			Name: config.ConfigClientSecret,
 			Help: "Box App Client Secret\nLeave blank normally.",
+		}, {
+			Name: "config_json",
```
ncw: is `config_json` what users would call it?  Might `jwt_file` or something like that be better?

Also would users ever put it directly in the config file?

Apologies if these are silly questions, I dont know much about JWT!

backend/box/box.go
```diff
@@ -98,6 +128,93 @@ func init() {
...
+
+	key, err := pkcs8.ParsePKCS8PrivateKey(block.Bytes, []byte(boxConfig.BoxAppSettings.AppAuth.Passphrase))
+	if err != nil {
+		log.Fatalf("Decrypt failed: %v", err)
+	}
+
+	return key.(*rsa.PrivateKey)
+}
+
+// ConfigJSON defines the shape of a box config.json
```
ncw: These type definitions maybe should go in `api/types.go`?

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Config configures rclone using JWT
+func Config(id, name string, claims *jws.ClaimSet, header *jws.Header, queryParams map[string]string, privateKey *rsa.PrivateKey, m configmap.Mapper) error {
+	payload, err := jws.Encode(header, claims, privateKey)
+	if err != nil {
+		log.Fatalf("Failed to encode payload: %v", err)
+	}
+	req, err := http.NewRequest("POST", claims.Aud, nil)
+	if err != nil {
+		log.Fatalf("Failed to create new request: %v", err)
```
ncw: This should be returning errors to the user, not exiting out...

So

```go
if err != nil {
    return errors.Wrap(err, "jwtutil: failed to create new request")
}
```

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+	q.Add("grant_type", "urn:ietf:params:oauth:grant-type:jwt-bearer")
+	q.Add("assertion", payload)
+	for key, value := range queryParams {
+		q.Add(key, value)
+	}
+	queryString := q.Encode()
+
+	req, err = http.NewRequest("POST", claims.Aud, bytes.NewBuffer([]byte(queryString)))
+	if err != nil {
+		log.Fatalf("Failed to create new request: %v", err)
```
ncw: likewise here

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	client := &http.Client{}
```
ncw: We should probably pass in a *http.Client here so the caller can pass in [an rclone client](https://github.com/rclone/rclone/blob/master/fs/fshttp/http.go#L188) which will do debuging, rate limiting etc, etc

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Fatalf("Failed making auth request: %v", err)
```
ncw: return error

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if resp.StatusCode != 200 {
+		log.Fatalf("Failed making auth request: %v", resp.Status)
```
ncw: return error

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			fmt.Printf("Error closing response Body: %v", err)
```
ncw: return error by naming the return error in the function signature and set it here.

lib/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}()
+
+	result := &response{}
+	err = json.NewDecoder(resp.Body).Decode(result)
+	if err != nil || result.AccessToken == "" {
+		log.Fatalf("Failed to get token: %v", err)
```
ncw: return error

backend/box/box.go
```diff
@@ -98,6 +128,93 @@ func init() {
 	})
 }
 
+func getBoxConfig(configFile string) ConfigJSON {
+	jsonData, err := os.Open(configFile)
+	if err != nil {
+		log.Fatalf("Failed to read Box config: %v", err)
```
ncw: Ideally this should return the error upwards.

I'm trying to kill off the log.Fatals in the backends so it would be helpful if you could pass errors up and put the log.Fatal in the Config function where it will be easy to fix when you can return an error from there!

backend/box/box.go
```diff
@@ -98,6 +128,93 @@ func init() {
...
+	}
+	byteValue, _ := ioutil.ReadAll(jsonData)
```
ncw: error check?

backend/box/box.go
```diff
@@ -98,6 +128,93 @@ func init() {
 	})
 }
 
+func getBoxConfig(configFile string) ConfigJSON {
+	jsonData, err := os.Open(configFile)
```
ncw: This open appears to be missing a Close?

You might be better off using ioutil.ReadFile rather than Open/ReadAll/Close

backend/box/box.go
```diff
@@ -98,6 +128,93 @@ func init() {
...
+	var boxConfig ConfigJSON
+	err = json.Unmarshal(byteValue, &boxConfig)
+	if err != nil {
+		log.Fatalf("Failed to parse Box config: %v", err)
+	}
+	return boxConfig
+}
+
+func getClaims(boxConfig ConfigJSON, boxSubType string) *jws.ClaimSet {
+	val, _ := jwtutil.RandomHex(20)
```
ncw: ignored error?

dbramwell:
Great, thanks for the help. I'm very much a go newbie so I've probably completely missed every coding standard! I'll reply to the comments inline.

With regards to adding JWT auth to other backends, from my limited JWT knowledge the Config function in `jwtutil.go` should be generic enough for other backends to use. Each backend that supports JWT should simply have to call it with the correct `ClaimSet` etc. Having said that I'm not sure if any other storage providers support JWT auth in the same way?

backend/box/box.go
```diff
@@ -84,6 +107,13 @@ func init() {
 		}, {
 			Name: config.ConfigClientSecret,
 			Help: "Box App Client Secret\nLeave blank normally.",
+		}, {
+			Name: "config_json",
+			Help: "Box App config.json location\nLeave blank normally.",
+		}, {
+			Name:    "box_sub_type",
+			Help:    "'enterprise' or 'user' depending on the type of token being requested.",
```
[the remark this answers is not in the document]
dbramwell: Thanks, will do

backend/box/box.go
dbramwell: In the case of Box I think possibly, the "Authenticate with JWT" [tutorial](https://developer.box.com/docs/construct-jwt-claim-manually#section-1-read-json-config) refers to it as `config.json`, and when downloading the file from the box developer page when logged in it saves it as `123456__config.json` (random numbers).

`config.json` does seem rather generic for rclone purposes though. I think `jwt_file` might be misleading as the file doesn't contain the jwt but rather the config to create a jwt assertion which can then be exchanged for an oauth token. Maybe `box_config_file`?

backend/box/box.go
dbramwell: Yes they definitely should, thanks.

ncw:
> Great, thanks for the help. I'm very much a go newbie so I've probably completely missed every coding standard! I'll reply to the comments inline.

I think you've done very well :-)

> With regards to adding JWT auth to other backends, from my limited JWT knowledge the Config function in `jwtutil.go` should be generic enough for other backends to use. Each backend that supports JWT should simply have to call it with the correct `ClaimSet` etc. Having said that I'm not sure if any other storage providers support JWT auth in the same way?

I know drive/gcs support JWT but the SDK has it built in already.

I think that some webdav providers (dput?) can support JWT too.

ncw:
Sorry for the radio silence?  How are you getting on with the fixes?  Need help?

dbramwell:
Hey. Sorry I was on holiday last week so didn't have any time to work on it. I should get some time tomorrow evening. It all looks quite straight forward, will let you know if I need a hand though, thanks.

ncw:
Great - thanks.  Looking forward to merging this :-)

backend/box/box.go
```diff
@@ -73,7 +83,28 @@ func init() {
...
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				claims, err := getClaims(boxConfig, boxSubType)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				signingHeaders := getSigningHeaders(boxConfig)
+				queryParams := getQueryParams(boxConfig)
+				err = jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/box/box.go
```diff
@@ -73,7 +84,29 @@ func init() {
...
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				claims, err := getClaims(boxConfig, boxSubType)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				signingHeaders := getSigningHeaders(boxConfig)
+				queryParams := getQueryParams(boxConfig)
+				client := fshttp.NewClient(fs.Config)
+				err = jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m, client)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/box/box.go
```diff
@@ -73,9 +84,34 @@ func init() {
...
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				claims, err := getClaims(boxConfig, boxSubType)
+				if err != nil {
+					log.Fatalf("Failed to configure token: %v", err)
+				}
+				signingHeaders := getSigningHeaders(boxConfig)
+				queryParams := getQueryParams(boxConfig)
+				client := fshttp.NewClient(fs.Config)
```
dbramwell: Unsure if this is the correct way to create a http client or not

backend/box/box.go
```diff
@@ -98,6 +147,74 @@ func init() {
...
+	rsaKey, err := pkcs8.ParsePKCS8PrivateKey(block.Bytes, []byte(boxConfig.BoxAppSettings.AppAuth.Passphrase))
+	if err != nil {
+		return nil, errors.Wrap(err, "box: failed to decrypt private key")
+	}
+
+	return rsaKey.(*rsa.PrivateKey), nil
+}
+
 // Options defines the configuration for this backend
 type Options struct {
```
dbramwell: I moved all the types that I created into api/types. I attempted to move other types that already existed, like Options and Fs, but found it quickly escalated into me moving a huge part of the file (lots of Fs function definitions, etc) over. Unsure if this is the correct thing to do

dbramwell:
Hey @ncw , sorry for the delay with this.

I think i've addressed all of the issues. I'm unsure if about a couple of points which i've left comments about.

backend/box/box.go
ncw: That's perfect. It will get you an http client with all Rclone's customisations.

backend/box/box.go
ncw: I usually put just the types that go over the wire in the API in api/types.go.

This file should be mostly struct definitions maybe with a few methods for reading data. Certainly no business logic for rclone .

The point of doing this is so that the definitions can be reused else where.

dbramwell: That makes sense. Does that mean that some of the functions to do with parsing the config file should also go in api/types?

dbramwell:
Awesome. Thanks a lot for the guidance.

I've added the snips back in and squashed it all into 3 commits.

backend/box/box.go
ncw: They could do but I think what you've done is fine :-)

docs/content/box.md
```diff
@@ -31,13 +32,79 @@ Type of storage to configure.
 Choose a number from below, or type in your own value
 [snip]
 XX / Box
+ 1 / 1Fichier
+   \ "fichier"
+ 2 / A stackable unification remote, which can appear to merge the contents of several remotes
+   \ "union"
```
ncw: This still seems to have all the other cloud providers?

This is what it should look like

```
Type of storage to configure.
[snip]
XX / Box
   \ "box"
[snip]
Storage> box
```

dbramwell:
Haha, sorry. That makes so much more sense. I think when I first started on this the documentation still had the (out-dated) list of providers in there, hadn't realised that the list had been removed altogether. Still I'm disappointed in myself for not realising what I'd done made absolutely no sense. I blame the fact that it was before 9am!

ncw:
> Haha, sorry. That makes so much more sense. I think when I first started on this the documentation still had the (out-dated) list of providers in there, hadn't realised that the list had been removed altogether.

No worries!  I'd forgotten that this PR lived through the changes I made to the docs to remove the other providers.

That looks perfect now, I'll merge it :-)

Thank you very much for doing this.

jacquesvandermerwe:
@dbramwell  Thanks for this excellent addition. I have 2 questions, having looked at the code that I cannot figure out.

1. If the box_sub_type is 'user' so that the JWT can do 'As-User'  calls where do we define the user id to use for the 'As-User' call. It isn't in the config setup, is there a switch that is not documented?
2. Does rclone handle expiry for JWT? Since there isn't a refresh_token, what would happen after 60 minutes if the JWT token expires?

Thanks in advance

dbramwell:
Hey @jacquesvandermerwe 

1. Good point, reading the Box docs I'm missing that functionality. My organisation doesn't use the As-User stuff. I guess user_id should be an additional config option. We'd then want to add the As-User header in each of the requests if the user_id is set.

2. It doesn't for Box, not sure if it does anywhere else. We get around this by adding the config to our automated process. You can do config in a non-interactive way like this:
`rclone config create box box box_config_file config.json
rclone ls box:/foo`

I guess there's an opportunity here to dig a little deeper into Rclone and make it do the config stuff again when the token expires.

Hope this helps, sorry my answers are basically "It doesn't do what you want it to do"!

Issue #2907, opened 2019-01-11
vfs: with cache mode >= writes rename should rename any cached items too


ncw:
With `--vfs-cache-mode writes` or `--vfs-cache-mode full` when we do a rename, we rename directly on the cloud storage, and ignore the cache.

This isn't incorrect, however it means that a file which is, for example, written as `file.tmp` then renamed to `file` will need to be re-downloaded when `file` is opened again, rather than using `file.tmp` which is still on the disk.

Originally from: #2730

ncw:
I think this is working now.

PR #3480, opened 2019-08-26
Rename files in cache too if they exist

darthShadow:

Files in cache should be renamed too along with renames on the remote

#2907 

#### Checklist

### DIFF
--- a/vfs/cache.go
+++ b/vfs/cache.go
@@ -278,53 +278,40 @@ func (c *cache) exists(name string) bool {
 }
 
 // renames the file in cache
-func (c *cache) rename(name string, newName string) bool {
+func (c *cache) rename(name string, newName string) (err error) {
 	osOldPath := c.toOSPath(name)
 	osNewPath := c.toOSPath(newName)
 	sfi, err := os.Stat(osOldPath)
 	if err != nil {
-		return false
+		return errors.Wrapf(err, "Failed to stat source: %s", osOldPath)
 	}
 	if !sfi.Mode().IsRegular() {
 		// cannot copy non-regular files (e.g., directories, symlinks, devices, etc.)
-		fs.Errorf("rename: non-regular source file %s (%q)", sfi.Name(), sfi.Mode().String())
-		return false
+		return errors.Errorf("Non-regular source file: %s (%q)", sfi.Name(), sfi.Mode().String())
 	}
 	dfi, err := os.Stat(osNewPath)
 	if err != nil {
 		if !os.IsNotExist(err) {
-			fs.Errorf(name, "rename: Failed to stat destination: %v", err)
-			return false
+			return errors.Wrapf(err, "Failed to stat destination: %s", osNewPath)
 		}
 		parent := findParent(osNewPath)
 		err = os.MkdirAll(parent, 0700)
 		if err != nil {
-			return false
+			return errors.Wrapf(err, "Failed to create parent dir: %s", parent)
 		}
 	} else {
 		if !(dfi.Mode().IsRegular()) {
-			fs.Errorf("rename: non-regular destination file: %s (%q)", dfi.Name(), dfi.Mode().String())
-			return false
+			return errors.Errorf("Non-regular destination file: %s (%q)", dfi.Name(), dfi.Mode().String())
 		}
 		if os.SameFile(sfi, dfi) {
-			return true
+			return nil
 		}
-	}
-	if err = os.Link(osOldPath, osNewPath); err == nil {
-		err = os.Remove(osOldPath)
-		if err != nil {
-			fs.Errorf(name, "rename: Failed to remove source in cache: %v", err)
-			return false
-		}
-		fs.Infof(name, "Renamed in cache")
-		return true
 	}
 	if err = os.Rename(osOldPath, osNewPath); err != nil {
-		fs.Errorf(name, "Failed to rename in cache: %v", err)
-		return false
+		return errors.Wrapf(err, "Failed to rename in cache: %s to %s", osOldPath, osNewPath)
 	}
 	fs.Infof(name, "Renamed in cache")
-	return true
+	return nil
 }
 
 // _close marks name as closed - must be called with the lock held
--- a/vfs/cache.go
+++ b/vfs/cache.go
@@ -270,9 +270,60 @@ func (c *cache) exists(name string) bool {
 	if err != nil {
 		return false
 	}
-	if fi.IsDir() {
+	// checks for non-regular files (e.g. directories, symlinks, devices, etc.)
+	if !fi.Mode().IsRegular() {
 		return false
 	}
+	return true
+}
+
+// renames the file in cache
+func (c *cache) rename(name string, newName string) bool {
+	osOldPath := c.toOSPath(name)
+	osNewPath := c.toOSPath(newName)
+	sfi, err := os.Stat(osOldPath)
+	if err != nil {
+		return false
+	}
+	if !sfi.Mode().IsRegular() {
+		// cannot copy non-regular files (e.g., directories, symlinks, devices, etc.)
+		fs.Errorf("rename: non-regular source file %s (%q)", sfi.Name(), sfi.Mode().String())
+		return false
+	}
+	dfi, err := os.Stat(osNewPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fs.Errorf(name, "rename: Failed to stat destination: %v", err)
+			return false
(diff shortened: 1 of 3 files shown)


darthShadow:
I still need to handle cases where the file is moved into or out of a directory.

vfs/cache.go
```diff
@@ -276,6 +276,21 @@ func (c *cache) exists(name string) bool {
...
+	osNewPath := c.toOSPath(newName)
+	err := os.Rename(osPath, osNewPath)
+	if err != nil && !os.IsNotExist(err) {
+		fs.Errorf(name, "Failed to rename in cache: %v", err)
+		return err
+	} else {
+		fs.Infof(name, "Renamed in cache")
+		return nil
+	}
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

vfs/cache.go
```diff
@@ -276,6 +276,21 @@ func (c *cache) exists(name string) bool {
...
 
+// remove should be called if name is deleted
+func (c *cache) rename(name string, newName string) error {
+	osPath := c.toOSPath(name)
+	osNewPath := c.toOSPath(newName)
+	err := os.Rename(osPath, osNewPath)
+	if err != nil && !os.IsNotExist(err) {
+		fs.Errorf(name, "Failed to rename in cache: %v", err)
+		return err
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
+		if !(dfi.Mode().IsRegular()) {
+			fs.Errorf("rename: non-regular destination file: %s (%q)", dfi.Name(), dfi.Mode().String())
+			return false
+		}
+		if os.SameFile(sfi, dfi) {
+			return true
+		}
+	}
+	if err = os.Link(osOldPath, osNewPath); err == nil {
```
ncw: Why not use os.Rename always?  Isn't it equivalent to Link+Remove?

os.Link doesn't work on all windows fileystems (eg FAT)

os.Rename may have troubles renaming over an existing file on Windows.

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
 	if err != nil {
 		return false
 	}
-	if fi.IsDir() {
+	// checks for non-regular files (e.g. directories, symlinks, devices, etc.)
+	if !fi.Mode().IsRegular() {
```
ncw: Good catch!

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
 		return false
 	}
 	return true
 }
 
+// renames the file in cache
+func (c *cache) rename(name string, newName string) bool {
```
ncw: It would be more go style to return an annotated error, than a boolean here I think...

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
+	osOldPath := c.toOSPath(name)
+	osNewPath := c.toOSPath(newName)
+	sfi, err := os.Stat(osOldPath)
+	if err != nil {
+		return false
+	}
+	if !sfi.Mode().IsRegular() {
+		// cannot copy non-regular files (e.g., directories, symlinks, devices, etc.)
+		fs.Errorf("rename: non-regular source file %s (%q)", sfi.Name(), sfi.Mode().String())
```
ncw: ... so instead of fs.Errorf(...); return false

do
```go
return errors.Errorf("rename: non-regular source file %s (%q)", sfi.Name(), sfi.Mode().String()))
```

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
+		return false
+	}
+	dfi, err := os.Stat(osNewPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fs.Errorf(name, "rename: Failed to stat destination: %v", err)
```
ncw: ...and likewise wrap the errors when you have them so this should be

```go
return errors.Wrap(err, "rename: Failed to stat destination")
```

vfs/file.go
```diff
@@ -124,6 +124,14 @@ func (f *File) rename(ctx context.Context, destDir *Dir, newName string) error {
 			fs.Errorf(f.Path(), "File.Rename error: %v", err)
 			return err
 		}
+
+		// Rename in the cache too if it exists
+		if f.d.vfs.Opt.CacheMode >= CacheModeWrites && f.d.vfs.cache.exists(f.Path()) {
+			if ok := f.d.vfs.cache.rename(f.Path(), newPath); !ok {
```
ncw: Then this should be something like

```go
err = f.d.vfs.cache.rename(f.Path(), newPath)
if err != nil {
    fs.Infof(f.Path(), "File.Rename failed in Cache: %v", err)
}
```

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
 	return true
 }
 
+// renames the file in cache
+func (c *cache) rename(name string, newName string) bool {
+	osOldPath := c.toOSPath(name)
+	osNewPath := c.toOSPath(newName)
+	sfi, err := os.Stat(osOldPath)
+	if err != nil {
+		return false
```
ncw:
```go
return errors.Wrap(err, "rename: source file")
```

vfs/cache.go
```diff
@@ -270,12 +270,63 @@ func (c *cache) exists(name string) bool {
...
+		err = os.MkdirAll(parent, 0700)
+		if err != nil {
+			return false
+		}
+	} else {
+		if !(dfi.Mode().IsRegular()) {
+			fs.Errorf("rename: non-regular destination file: %s (%q)", dfi.Name(), dfi.Mode().String())
+			return false
+		}
+		if os.SameFile(sfi, dfi) {
```
ncw: How likely do you think `os.SameFile` is to return true here? I'm struggling to see how it might happen

darthShadow: Just thought it better to be safe than have an issue afterwards in case it ever happens. This avoids silent failures from rename in case both the source & destination point to the same file via links.

Source: https://forum.golangbridge.org/t/os-rename-and-hard-link-overwrite/7838

vfs/cache.go
darthShadow: Not much of difference either way. Didn't try it on windows, if it doesn't work there, I will just remove the section that tries to link.

As for rename, what do you think would be better? Remove the existing destination or do nothing and print an error log?

darthShadow:
Thanks for the review. As for the tests, I may need some help with that. Haven't written any tests before 😅

ncw:
> Thanks for the review. As for the tests, I may need some help with that. Haven't written any tests before

Take a look in the cache_test.go file there is a fair amount of machinery in there.  You'd write a test for the rename function in there.

It would be good if we could also adjust the rename file test  to check the correct things are happening in the cache but that might be a bit trickier.

https://github.com/rclone/rclone/blob/d1a39dcc4b5eb7a5f086f4fcd5734f1af06bbe80/vfs/dir_test.go#L469

ncw:
How are you doing with this?  Need help finishing it off?

darthShadow:
Need to get some time together to finish off the tests. Will be updating it by the weekend at the latest.

ncw:
> Need to get some time together to finish off the tests. Will be updating it by the weekend at the latest.

Great :-)  Let me know if you need any help.

darthShadow:
@ncw Sorry for the delay, but doesn't look like I will have any time in the next 2-3 months to write any tests and finish this. I have just pushed the latest changeset that I had started on and hope that you or someone else can use that and finish this PR.

ncw:
> @ncw Sorry for the delay, but doesn't look like I will have any time in the next 2-3 months to write any tests and finish this. I have just pushed the latest changeset that I had started on and hope that you or someone else can use that and finish this PR.

That is no problem.  I'll try to get this merged one way or another in 1.51

@bdutro do you fancy finishing this off?

bdutro:
@ncw, sure. How should I add my changes? It seems like the proper way is to do a pull request against @darthShadow's branch, which would then get added to this pull request when it's merged.

ncw:
This is now merged in #3688 - thank you @darthShadow for the initial work and @bdutro for finishing it off.

PR #3483, opened 2019-08-26
rcd: Added missing parameter for web-gui info logs.

negative0:

Added nil for the object type while logging to fs.Infof


ncw:

Issue #591, opened 2016-07-18
add  Cloud mail.ru


C-A-T9LIFE:
Please add  Cloud mail.ru  https://cloud.mail.ru .

ncw:
Does it have a public API somewhere?

C-A-T9LIFE:
Hey. githabe have on several projects with api
https://github.com/search?utf8=%E2%9C%93&q=cloud+mail.ru
https://github.com/ldinc/mcloudru-lib/tree/dev

xor2003:
This is working code:
https://github.com/MasterSoft24/CloudCross/tree/master/src/MailRu

sergevs:
Here is the API implementation which should be sufficient to develop cloud.mail.ru support for rclone: https://github.com/sergevs/GoMailRuCloud

I'm the API project owner and can answer any questions related to the API or provide any other kind of support by a request.

zeliboba:
I alos vote for cloud mail.ru support. Any update?

ivandeex:
Hi Nick. I am preparing a PR, which implements the _mailru_ backend for rclone. I am still writing the documentation. Planning to submit the PR this month. In the meantime everybody interested can download beta binaries from the releases page. https://github.com/ivandeex/rclone

ncw:
> Hi Nick. I am preparing a PR, which implements the _mailru_ backend for rclone. I am still writing the documentation. Planning to submit the PR this month. In the meantime everybody interested can download beta binaries from the releases page. https://github.com/ivandeex/rclone

I just had a quick look a the code - excellent job - very easy to read and similar to the existing backends :-)

I guess you are following the [integrating a new backend proceedure](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend)?

How are the integration tests progressing?

I'll put this on the 1.49 milestone.

ivandeex:
Closed by https://github.com/rclone/rclone/pull/3501

ncw:
I've merged this to master now which means it is in [the latest beta](https://beta.rclone.org) and will be released in v1.50

trochdewei:
I can not access cloud.mail.ru remote because of error `Failed to create file system for "mailcloud:": Failed to authorize: oauth2: server response missing access_token`

ncw:
> I can not access cloud.mail.ru remote because of error `Failed to create file system for "mailcloud:": Failed to authorize: oauth2: server response missing access_token`

[...]

PR #3501, opened 2019-08-31
Add Mailru backend

ivandeex:

This PR implements a remote for [Mail.ru Cloud](https://cloud.mail.ru/), a Russian cloud storage provider. All unit and integration tests pass for me. See https://travis-ci.org/ivandeex/rclone
I will submit a separate PR which will enable Mailru hash algorithm in rclone.

This backend was requested in issue #591 

#### Checklist


ivandeex:
This branch enables mailru hash in rclone globally: https://github.com/ivandeex/rclone/tree/mrhash
It has a single commit https://github.com/ivandeex/rclone/commit/31e30753728041b9df8cc8135b6f138ca665f0d7
If you agree, I can add the commit right into this PR.

docs/content/mailru.md
```diff
@@ -0,0 +1,274 @@
...
+        - Ignore and continue.
+
+#### --mailru-user-agent
+
+User agent used by client (internal)
+
+- Config:      user_agent
+- Env Var:     RCLONE_MAILRU_USER_AGENT
+- Type:        string
+- Default:     "CloudDiskOWindows 17.12.0009 beta WzBbt1Ygbm"
```
ncw: Is this necessary?  I prefer to use rclone's user agent if possible so cloud providers can identify problems, however if this is the only way then fine!

backend/mailru/mailru_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/rclone/rclone/backend/mailru"
+	"github.com/rclone/rclone/fstest"
+	"github.com/rclone/rclone/fstest/fstests"
+)
+
+// TestIntegration runs integration tests against the remote
+func TestIntegration(t *testing.T) {
+	remoteName := *fstest.RemoteName // from the -remote cli option
+	if remoteName == "" {
+		remoteName = "TestMailru:"
```
ncw: Just specify `RemoteName: "TestMailru:"` and when the `-remote` flag is passed it will get overridden automatically.

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+	return fmt.Sprintf("[%s]", f.root)
+}
+
+// Precision return the precision of this Fs
+func (f *Fs) Precision() time.Duration {
+	return time.Second
+}
+
+// Hashes returns the supported hash sets
+func (f *Fs) Hashes() hash.Set {
```
ncw: I think adding your patch to `fs/hash` and making this return an mrhash would be a good idea - it will mean rclone can check hashes on upload/download and using `rclone check`.

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+	if err = oauthutil.PutToken(f.name, f.m, t, false); err != nil {
+		return err
+	}
+
+	// Mailru API server expects access token not in the request header but
+	// in the URL query string, so we must use a bare token source rather than
+	// client provided by oauthutil.
+	//
+	// WARNING: direct use of the returned token source triggers a bug in the
+	// `(*token != *ts.token)` comparison in oauthutil.TokenSource.Token()
```
ncw: Do you want to fix up oauthutil?  You could do that in another commit in this PR if you wanted? Or make a new issue about it?

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+		}, {
+			Name:     "speedup_enable",
+			Default:  true,
+			Advanced: false,
+			Help: `Skip full upload if there is another file with same data hash.
+
+This feature is called "speedup" or "put by hash". It is especially efficient
+in case of generally available files like popular books, video or audio clips,
+because files are searched by hash in all accounts of all mailru users.
+Please note that rclone needs local memory and disk space to calculate
```
ncw: Getting the mrhash into the hash framework will mean that you can ask the source for the mrhash which will mean you no longer need to buffer things to calculate the hash in advance.

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+	case api.MkdirResultInvalidName:
+		return ErrorInvalidName
+	default:
+		return fmt.Errorf("mkdir error %d", status)
+	}
+}
+
+// Mkdir creates the container (and its parents) if it doesn't exist.
+// Normally it ignores the ErrorDirAlreadyExist, as required by rclone tests.
+// Nevertheless, such programs as borgbackup or restic use mkdir as a locking
```
ncw: How does borgbackup call rclone? Maybe a flag to `rclone mkdir` might be better?

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+}
+
+// Update an existing object
+// Copy the reader into the object updating modTime and size
+// The new object may have been created if an error is returned
+func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) error {
+	wrapIn := in
+	size := src.Size()
+	if size < 0 {
+		return errors.New("mailru does not support streaming uploads")
```
ncw: This will mean that `rclone mount` uploads have to be spooled to disk first...  It would be good to support if possible, but some backends just can't do it!

backend/mailru/mailru.go
```diff
@@ -0,0 +1,2356 @@
...
+		return err
+	}
+
+	var (
+		fileBuf  []byte
+		fileHash []byte
+		newHash  []byte
+	)
+
+	// Check whether file is eligible for speedup method (put by hash)
```
ncw: With mrhash in the hash framework you can just call `src.Hash(hash.Mrhash)` and if copying from the local backend, you'll get the hash which will be much quicker than spooling the file.

If you are copying from another backend (other than mailru or local) then you won't get the hash, and will need to spool it if you want it in advance.

If you want to see an example of another backend that does this see

https://github.com/rclone/rclone/blob/e2ce687f9377813039a1614658ef2d1952de7a90/backend/jottacloud/jottacloud.go#L1262-L1278

docs/content/mailru.md
ivandeex: I will recheck whether API works stable for non-mailru user agent.
Is there a generic flag to specify non-rclone user agent?

ncw: This is the generic one

      --user-agent string   Set the user-agent to a specified string. The default is rclone/ version (default "rclone/v1.49.1-DEV")

ivandeex: Can a generic flag like this be attached in the [remote] section?

ivandeex:
The protocol was originally rev-eng'ed by this guy https://github.com/yar229/WebDavMailRuCloud

backend/mailru/mailru.go
ivandeex: Will submit an issue.

backend/mailru/mailru.go
ivandeex: Unfortunately Mailru wants size and hash in advance. That's the base of their deduplication logic. OTOH they let you skip the whole body if they have one :)

ncw:
Let me know when you are ready for another round of review :-)

docs/content/mailru.md
ivandeex: I can't find a way to stick `--user-agent` to remote in rclone.conf. I will then keep `--mailru-user-agent` but make it by default empty and fall back to --user-agent.

backend/mailru/mailru.go
ivandeex: Let me investigate and submit a separate Issue. The issue will state that the mailru backend quirk should be removed when the issue resolves.

ncw:
That is looking really good now.  It passes all the tests and also my informal testing :-)

Shall I merge it now and we fixup anything else we find later?

image

ivandeex:
I have committed remaining enhancemented and run unit/integration test suites successfully.
Beta release that includes up-to-date changes can be found here https://github.com/ivandeex/rclone/releases/tag/v1.49.mr08
I will later submit tickets or PRs for two remaining issues:
- mkdir atomicity for borg backup done generally instead of mailru-specific quirk -- see https://github.com/rclone/rclone/pull/3501#discussion_r319996798 and https://borgbackup.readthedocs.io/en/stable/usage/general.html#file-systems
- replace _oauthutil_ workaround by fix -- see https://github.com/rclone/rclone/pull/3501#discussion_r319994696
Please review and merge.

ncw:
I've merged that now - thank you very much :-)

I squashed the commits locally and added [the missing docs](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend). 

- commit ba1daea07263f1e4b65898141cc3e9a0425e8803
- commit bdcd0b4c640b40fb1f21ff6ba44dbc9a6273517a

I'll manually close this pull request now.

Thank you very much for this :-)

Alexey-github:
https://github.com/rclone/rclone/issues/3988

PR #3503, opened 2019-09-01
Chunker

ivandeex:

Many cloud providers limit file size depending on account tier.
This PR creates Chunker, an overlay remote that transparently splits large files during put, finds paers during directory scans and transparently merges parts when a file is downloaded. Chunks are kept on the underlying remote as ordinary files. The overlay supports confirable chunk size and flexible naming schemes configured with user-friendly patterns. It has a small memory footprint and performs most actions on the fly without using temporary files. OTOH this method does not play well with remotes involving multipart uploads, resulting in premature EOF. One known workaround is to set a multipart cutoff larger that the chunk size. It is described in the Chunker documentation.

Yes. See https://github.com/ncw/rclone/issues/497

#### Checklist


ivandeex:
Integration tests performed successfully with Chunker overlay above the following remotes: local box mailru.
Unit tests have been slightly modified to let user set a custom remote, eg. to test against another underlying backend. Another custom option allows to enable/disable bad windows characters in file names under tests.
Another possible improvement is to extend FsPutFiles family of tests specifically for this overlay remote, namely: dynamically create Chunker overlay with a chunk size small enough and upload/download files larger than this size. This can be implemented by a later commit(PR).

docs/content/chunker.md
```diff
@@ -0,0 +1,237 @@
...
+d) Delete this remote
+y/e/d> y
+```
+
+## Specifying the remote ##
+
+In normal use, make sure the remote has a `:` in. If you specify the
+remote without a `:` then rclone will use a local directory of that
+name.  So if you use a remote of `/path/to/secret/files` then rclone
+will encrypt stuff to that directory.  If you use a remote of `name`
```
ncw: encrypt -> chunk ?

docs/content/chunker.md
```diff
@@ -0,0 +1,237 @@
...
+files are given their final names.
+
+By default chunks names are `BIG_FILE_NAME.rclone-chunk.001`,
+`BIG_FILE_NAME.rclone-chunk.002` etc, because the default chunk name
+format is `*.rclone-chunk.###`. You can configure another name format
+using the `--chunker-name-format` option. The format uses asterisk
+`*` as a placeholder for the base file name and one or more consecutive
+hash characters `#` as a placeholder for the chunk number. There must be
+one and only one asterisk. The number of consecutive hashes defines the
+minimum left of a chunk number. If a chunk number has less digits than
```
ncw: left -> length

docs/content/chunker.md
```diff
@@ -0,0 +1,237 @@
...
+#### Metadata
+
+By default chunker creates metadata object besides data chunks. It is named
+after the original file. Currenly then only supported format is `simplejson`,
+which keeps the following fields:
+
+- Name         - name of the composite file (equal to the meta file name)
+- Size         - total size of all chunks
+- PublicKey    - currently not used, always "null"
+- CreationDate - last modification time
```
ncw: Perhaps this should be called `ModificationDate` if that is what it is?

docs/content/chunker.md
```diff
@@ -0,0 +1,237 @@
...
+
+The list command might encounter files with missinng or invalid chunks.
+This boolean flag tells what rclone should do in such cases. By default
+when chunker detects a missing chunk it will silently ignore the whole
+group. You can use the `--chunker-fail-on-bad-chunks` option to make it
+fail loudly.
+
+#### Modified time and hashes
+
+Chunker stores modification times using the underlying remote so support
```
ncw: Can you explain how it does that?  On the first chunk?

docs/content/chunker.md
```diff
@@ -0,0 +1,237 @@
...
+depends on that. Hashes are not stored.
+
+### Limitations ###
+
+Chunker overlay internally streams a file being split or merged and
+consequently fails if an underlying remote switches to multipart mode,
```
ncw: Why does it fail?  I would have thought it would work properly whether the underlying remote is using chunked upload or not?

It is probably good advice to turn off the chunking though.

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+	"github.com/rclone/rclone/fs/config/configstruct"
+	"github.com/rclone/rclone/fs/fspath"
+	"github.com/rclone/rclone/fs/hash"
+)
+
+const (
+	// optimizeFirstChunk enables the following Put optimization:
+	// if a single chunk is expected, put the first chunk using base target
+	// name instead of temporary name, thus avoiding extra rename operation.
+	// WARNING: this optimization is not transaction safe!
```
ncw: Does this mean that chunker only works on backends which support rename?

This should be in the docs...

And we should check for it in NewFs with `operations.CanServerSideMove` and return an error as soon as possible

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+	c.chunkLimit = c.chunkSize
+	c.sizeLeft = c.sizeTotal
+
+	// Wrap the accounting back on
+	wrapIn := wrapBack(c)
+
+	baseRemote := remote
+	tempRemote := fmt.Sprintf("%s__tmp.%d.", baseRemote, time.Now().Unix())
+
+	// Prepare to rename
```
ncw: It would be better to use `operations.Move` here as if the underlying remote doesn't support Move it will fallback to Copy+Delete.  None of the bucket based remotes support Move.  Nearly every rclone backend supports Move or Copy now.

Use `operations.CanServerSideMove` to check that `operations.Move` doesn't do a download then re-upload.

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+	ChunkSize       fs.SizeSuffix `config:"chunk_size"`
+	NameFormat      string        `config:"name_format"`
+	StartFrom       int           `config:"start_from"`
+	MetaFormat      string        `config:"meta_format"`
+	FailOnBadChunks bool          `config:"fail_on_bad_chunks"`
+}
+
+// Fs represents a wrapped fs.Fs
+type Fs struct {
+	fs.Fs
```
ncw: You've embedded `fs.Fs` here which means you'll get its underlying methods.

This can be good or bad, however it looks like you've implemented methods you didn't need to (like `About`).

So either remove the extra methods, or make `fs.Fs` into a field.  I'd probably do the latter as then when new methods are made on the underlying backends we need to make a deliberate choice to support them here.

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+	return f.features
+}
+
+// String returns a description of the FS
+func (f *Fs) String() string {
+	return fmt.Sprintf("Chunked '%s:%s'", f.name, f.root)
+}
+
+// parseNameFormat converts pattern-based name format into Printf format and Regexp
+func (f *Fs) parseNameFormat(pattern string) error {
```
ncw: This should probably have a unit test which makes patterns and checks they perform as expected!

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+	return o.main.Size()
+}
+
+// Storable returns whether object is storable
+func (o *Object) Storable() bool {
+	if o.main != nil {
+		return o.main.Storable()
+	}
+	if o.chunks == nil {
+		panic("invalid chunked object")
```
ncw: Is this likely?

Maybe factor this into a method on object `firstChunk` since it seems to be used several times?

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+		}
+	}
+	if limit < 0 {
+		limit = o.size - offset
+	}
+
+	var mrc multiReadCloser
+	var readers []io.Reader
+
+	for _, chunk := range o.chunks {
```
ncw: I think opening all the chunks in advance will be very inefficient for small chunk sizes?

Wouldn't it be better to just have one chunk open at a time?

backend/chunker/chunker.go
```diff
@@ -0,0 +1,1412 @@
...
+func (oi *ObjectInfo) ModTime(ctx context.Context) time.Time {
+	if oi.newTime != nil {
+		return *oi.newTime
+	}
+	return oi.ObjectInfo.ModTime(ctx)
+}
+
+// Hash returns the selected checksum of the file
+// If no checksum is available it returns ""
+func (oi *ObjectInfo) Hash(ctx context.Context, hash hash.Type) (string, error) {
```
ncw: Could we put the Hash of the file into the metadata?

backend/chunker/chunker_test.go
```diff
@@ -0,0 +1,194 @@
...
+			ModTime: fstest.Time("2001-02-03T04:05:06.499999999Z"),
+			Path:    fmt.Sprintf("chunker-upload-%d.dat", fileSize),
+			Size:    fileSize,
+		})
+	})
+
+	_ = operations.Purge(context.Background(), remote, "")
+}
+
+// The below code is shamelessly copied from fstests.go, better have it exported.
```
ncw: Adding another commit to make them public would be fine :-)

backend/chunker/chunker_test.go
```diff
@@ -0,0 +1,194 @@
...
+			{Name: name, Key: "remote", Value: tempdir},
+		},
+		UnimplementableFsMethods:     []string{"OpenWriterAt"},
+		UnimplementableObjectMethods: []string{"MimeType"},
+	})
+}
+
+// TestLargeUpload tests large file upload. It runs when
+// both -remote and -upload-megabytes flags are set.
+func TestLargeUpload(t *testing.T) {
```
ncw: Note that you want to implement `fstests.InternalTester` by creating a function called `func (f *Fs) InternalTest(t *testing.T)` then this can be called as part of the integration tests so it doesn't have to do all it's own setup.

See the drive tests for an example

https://github.com/rclone/rclone/blob/e2ce687f9377813039a1614658ef2d1952de7a90/backend/drive/drive_internal_test.go#L271-L287

backend/chunker/chunker.go
ivandeex: Will add the check. What error code is best?

docs/content/chunker.md
ivandeex: yep

backend/chunker/chunker.go
ncw: I think an informative error for the user as in "can't use chunker on a backend which doesn't support move or copy"

docs/content/chunker.md
ivandeex: If metadata is "none", the modification time is taken from the first chunk. If metadata is On then it's taken from metadata.

docs/content/chunker.md
ivandeex: The format is obviously poor. It roots from https://github.com/yar229/WebDavMailRuCloud origins. My humble problem is a few terabytes of data kept using it :(
I propose to freeze this meta-format (as "wdmrcompat" or "yar229chunks" or "poormanjson", or choose a better name) and proceed to improving "simplejson", eg: remove PublicKey, replace CreationDate by ModificationTime, add HashSum+HashType. ok?

docs/content/chunker.md
ivandeex: The user report and logs can be found here: https://github.com/ivandeex/rclone/issues/19
I believe the Box layer was first to manifest the bug because its multipart cutoff is small compared to other backends.

backend/chunker/chunker.go
ivandeex: For the sake of transparency I am inclined to support at least About and Cleanup. Users should not revert to underlying remote to check free space if the mostly use overlay. Yet I am less inclined to transparently wrap SetTier and whatever pops up in future.
So, I'll factor fs.Fs into a field and keep only a subset of extra methods, to my liking. The holy GoF blessed us with composition over inheritance.

backend/chunker/chunker.go
ivandeex: Generally unlikely bug possible. Factoring this out seems a good idea.

backend/chunker/chunker.go
ivandeex: See above

backend/chunker/chunker.go
ivandeex: Ok, errors.New("can't use chunker on a backend which doesn't support move or **rename**")

backend/chunker/chunker.go
ivandeex: Right, it's suboptimal. Then I'll need something more intelligent than io.MultiReader(readers...), a custom reader that can open another chunk when previous chunk reading ends, and passes errors appropriately. Doesn't look an easy job. Can we have it a separate PR?

ivandeex: In a general case it depends on what is optimized: time to first byte or a total download time (it may be favorable to start parallel downloads asap).

docs/content/chunker.md
ncw: Oh, I see... Well it isn't terrible and if it is an existing "standard" then OK!  Your transition plan sounds fine to me, but we should try to finalise it before we get to the v1.50 release.

docs/content/chunker.md
ncw: Can you write that in the docs?

docs/content/chunker.md
ncw: Interestingly it manages to upload the entire first chunk OK, but fails on the second chunk with an EOF error. I don't see why that should be unless chunker is returning the EOF on a Read call?

```
019/08/04 07:49:32 DEBUG : mnt/engram/projects/archive/INDIVGraphTheory.tar.gz__tmp.1564919310.001: Multipart upload session started for 64 parts of size 32M
2019/08/04 07:49:32 DEBUG : mnt/engram/projects/archive/INDIVGraphTheory.tar.gz__tmp.1564919310.001: Cancelling multipart upload: multipart upload failed to read source: EOF
2019/08/04 07:49:33 DEBUG : mnt/engram/projects/archive/INDIVGraphTheory.tar.gz: Received error: multipart upload failed to read source: EOF - low level retry 1/10
```

backend/chunker/chunker.go
ncw: That should say "move or copy" not "move or rename"!  Remember if the remote can't move then it is implemented as copy then delete...

backend/chunker/chunker.go
ncw: :-) composition for the win!

ivandeex: Composition looses to fstests.go, as I can't drop anything:
    --- FAIL: TestIntegration/FsCheckWrap (0.00s)
        fstests.go:401: Missing Fs wrapper for DirCacheFlush
        fstests.go:401: Missing Fs wrapper for PublicLink
        fstests.go:401: Missing Fs wrapper for MergeDirs
        fstests.go:401: Missing Fs wrapper for UserInfo
        fstests.go:401: Missing Fs wrapper for Disconnect

ivandeex: I now embed wrapped Fs rather than inherit (will push commits later). But I cannot remove irrelevant methods from Chunker such as DirCacheFlush, PublicLink, MergeDirs, UserInfo, Disconnect because they are required by TestIntegration/FsCheckWrap.

ncw: You can mark these as unsupported in the test structure. See for example the test runner in the cache backend

zenjabba:
Is there a beta build of this version available yet?

ivandeex:
> Is there a beta build of this version available yet?

https://github.com/ivandeex/rclone/releases/tag/v1.49.mr08

ivandeex:
A quick way to reproduce chunker failure with multipart uploads described above https://github.com/rclone/rclone/pull/3503#discussion_r319979552 is to set box multipart cutoff to 50M, chunker chunk size to 100M and upload a 200M file.

Add this to `rclone.conf`:
```
[TestBoxCutoff50m]
type = box
upload_cutoff = 50m
token = {...}
[TestChunkerChunk100mCutoff50mBox]
type = chunker
remote = TestBoxCutoff50m:
chunk_size = 100m
```
and invoke this command:
```
time go test -count=1 -v ./backend/chunker -run=TestIntegration/FsMkdir/FsPutFiles/Internal/PutLarge -remote=TestChunkerChunk100mCutoff50mBox: -upload-kilobytes=204800
```

ivandeex:
New `simplejson` format supports hash sums and verification of chunk size/count.
The old version of `simplejson` is renamed to `wdmrcompat`.

Now integration tests run for a matrix of chunker parameters: chunking on/off (via `chunk_size`), and `meta_format`. See `fstest/test_all/config.yaml`.

Recommended `rclone.conf` snippets for new `config.yaml`:
```
[TestChunkerLocal]
type = chunker
meta_format = simplejson
remote = /tmp/rclone-chunker-test

[TestChunkerChunk3bLocal]
type = chunker
chunk_size = 3b
meta_format = simplejson
remote = /tmp/rclone-chunker-test

[TestChunkerNometaLocal]
type = chunker
meta_format = none
remote = /tmp/rclone-chunker-test

[TestChunkerChunk3bNometaLocal]
type = chunker
chunk_size = 3b
meta_format = none
remote = /tmp/rclone-chunker-test

[TestChunkerCompatLocal]
type = chunker
meta_format = wdmrcompat
remote = /tmp/rclone-chunker-test
```

Most tests use small test files, less than 100 bytes, frequently 3-5 bytes.
Running integration test suite on chunker remotes with chunk size set to
an exremely small value (3-4 bytes) will trigger multiple chunks in 99%
of tests. However, some tests create large files, which will result in
a few hundred thousands of chunks, hit OS resource limits and fail.
This commit adds a generic option in the test suite configuration
that can be used by tests to skip such cases.

ncw:
I've given the code a once over - it is looking good :-)  Apologies for the delay!

Can you rebase it then squash it into something like
- one commit for the chunker + docs
- one commit for the fstest.SizeLimit changes
- one commit for the test_all changes

If could arrange the commits so that rclone compiles for each one that helps bisection!

I'll merge it then.  We have a little while (until 12th October) until the 1.50 release which should hopefully give us a chance to shake any bugs out!

Thank you :-)

docs/content/chunker.md
ivandeex: fixed in the next round of commits

backend/chunker/chunker.go
ivandeex: implemented in the next round of commits

ivandeex:
> Can you rebase it then squash it into something like
> * one commit for the chunker + docs
> * one commit for the fstest.SizeLimit changes
> * one commit for the test_all changes
I will add documentation and squash/rebase commits in the next days.

ivandeex:
@ncw
I've squashed the PR to 7 commits here  https://github.com/rclone/rclone/pull/3503#commits-pushed-8820d22

A few questions need clearing:
- What is better for empty hashsum: no error or Unsupported? I don't fully understand how rclone pieces treat the difference. What's better returned for small files without metadata: empty sum or unsupported?
- You might decide to completely remove md5quick/sha1quick options. I described their pros and cons in the docs.
- I don't see a big difference in average download time between two strategies: fastopen (implemented initially) vs linear (requested by you). Maybe we'd better keep only one?
- Please notice the "tests memory pressure" patch. I stumbled upon the issue when I tested the hash_type/meta_format matrix on S3. Finally I just commented all extended tests out, yet I recommend to apply the patch - it will shorten the total test suite time in many cases (3h -> 1h on my vpc with 2g ram).
- I don't completely like the way I fixed synchronization tests for chunker - too much hacking the transfer count, yet I don't have a better way. Maybe provide chunk size to tests?

Also, some ideas for future:
- teach deletefile to remove hidden chunks for a file too
- a flag to reveal hidden chunks in listings
- a future metadata format (eg. call it "metajson") that embraces small files as well, to make hashsums completely consistent

Seemingly I addressed all your comments.
Please review.

ivandeex:
Aditional points collected after self-review. I will address them together with yours:
- Fs.Hashes() must must return HashSet[md5] for hash type "md5", [sha1] for "sha1", [md5,sha1] for quick modes, [none] otherwise. This change must be tested with extra test remotes based on S3 uncommented. (see code https://github.com/rclone/rclone/pull/3503/commits/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR992-R994)
- Object.SetTier() must either set tier for **all** chunks if file is composite, or SetTierer must be removed at all (see code https://github.com/rclone/rclone/pull/3503/commits/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR1776-R1781)
- In future optionally use DirCache since key chunker operations scan directories a lot

ivandeex:
@Loewkie I cannot reproduce your issue using the latest beta. Please submit as a separate ticket if the problem holds for you after v1.50.

Loewkie:
Thank you @ivandeex. I'll have a look.

ivandeex:
Also, some ideas for future:
- teach deletefile to remove hidden chunks for a file
- a flag to reveal hidden chunks in listings
- use DirCache as the key operations scan directories a lot (added on 2019-09-17)
- future metadata format (named eg. "metajson") that embraces small files making hashsums more consistent; simplejson and metajson should be interoperable at reading objects (metajson can become the default format in a future rclone release, but it's better to avoid rclone.conf migrations if possible) and reliably detect single-chunked objects with or without metadata (eg currently there is a "_TODO maybe it's a small single chunk_" in Fs.Object https://github.com/rclone/rclone/pull/3503/commits/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR625), the differenence is just about preferred put/update method.
- metajson can use metadata field "ver" to detect whether single-chunk object is allowed to have metadata, eg. ver=0 cannot (added on 2019-09-18)

ivandeex:
@ncw Gentle bump

comio:
Interesting work. Thanks @ivandeex
I will use this on top of my unreliable webdav

ncw:
Apologies for the delay

> I've squashed the PR to 7 commits here [#3503 (commits)](https://github.com/rclone/rclone/pull/3503#commits-pushed-8820d22)

Great - thank you :-)

> * What is better for empty hashsum: no error or Unsupported? I don't fully understand how rclone pieces treat the difference. What's better returned for small files without metadata: empty sum or unsupported?

It doesn't matter too much.  Empty is for when the hash is supported but you haven't got one for some reason (eg chunked files in S3).  If the hash can never be supported then unsupported.

> * You might decide to completely remove md5quick/sha1quick options. I described their pros and cons in the docs.

They look like they might be useful, I don't mind leaving them in.  I'll be guided by you here.

> * I don't see a big difference in average download time between two strategies: fastopen (implemented initially) vs linear (requested by you). Maybe we'd better keep only one?

My main concern with the fastopen is opening too many file handles at once when chunk is small compared to the file size.  If you are happy with the speed of linear then maybe we should remove the fastopen strategy.

> * Please notice the "tests memory pressure" patch. I stumbled upon the issue when I tested the hash_type/meta_format matrix on S3. Finally I just commented all extended tests out, yet I recommend to apply the patch - it will shorten the total test suite time in many cases (3h -> 1h on a VPC with 2G RAM), even on a travis box with 4G.

Ouch... That needs fixing properly in the s3 backend.  I think the thinking was that the large chunk sizes are needed because you can only have a limited number of chunks (10,000 I think).  However 2GB of memory is way too much! This will be used in `rclone mount` when uploading files so is a fairly common use.  

The swift backend probably does the same thing... Not sure about the other bucket based remotes.

> * I don't completely like the way I fixed synchronization tests for chunker - too much hacking the transfer count, yet I don't have a better way. Maybe provide chunk size to tests?

Yes that is not very sustainable!  I've spend quite a bit of time mucking about just getting those tests to pass, so maybe they should do something different, or maybe certain backends could opt out of the counting parts of those tests (eg union and chunker).

> * Fs.Hashes() should return HashSet[md5] for hash type "md5" or [sha1] for "sha1" **if and only if** the wrapped remote supports it too (so that advertised hash is consistently supported for files of any size), [none] otherwise. This change must be tested with extra-hash test remotes uncommented and should not imply significant changes or lots of new bypass conditions in the test suite. (see code [8820d22#diff-a132438692a9c451a2a1cf6aebcb3c3eR992-R994](https://github.com/rclone/rclone/commit/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR992-R994)).

You could move the failing test into a subtest then exclude it in `config.yaml` eg

https://github.com/rclone/rclone/blob/49d6d6425c2e9c54d5f41ad096179165674fe859/fstest/test_all/config.yaml#L58

That is the way I'm dealing with backend test failures at the moment

>   Note: with _iff_ condition, my tests only require to adjust move/copy counts in a predictable way which is expected, _without_ iff few more tests fail and I have to ignore counts in some cases as I can't predict reliably whetner backend tracks hashsums/renames.

Not sure what you mean by _iff_ condition, but I think that the counts are a smell of a bad test - I should think of a better way of doing this!

> * The `hash_type` option should not be advanced as users need it frequently (patch should update config section in docs)

OK

> * Object.SetTier() must either set tier for **all** chunks if file is composite, or SetTierer should be removed at all (see code [8820d22#diff-a132438692a9c451a2a1cf6aebcb3c3eR1776-R1781](https://github.com/rclone/rclone/commit/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR1776-R1781))

Happy for you to remove it, or fix it.  `SetTier` only works on a small number of backends anyway.

> * **Before** v1.50, add json field "ver" (named Version in struct) to metadata, which defaults to "0" so future releases can detect and automatically upgrade metadata or throw an error if version is not supported. Right now Open/Hash/ModTime/Update/etc (NewObject is lazy about metadata) **must** error out if "ver" is not "0" - it's **critical** that v1.50 will not corrupt file systems created by future releases

Great idea.

> * You might decide to completely remove _wdmrcompat_ to simplify code, simplify future effort of strategy-sing metadata logic, reduce tests matrix, and decrease support burden. This format is weak, I am steadily migrating my data from it.

I'll leave that to you to decide... I'm easy either way.

> * Documentation should state clearly that Chunker is in beta (like crypt) and include caveat emptor

Probably a good idea!

> Also, some ideas for future:

I normally stick these in a comment in the backend code.

> 
> * teach deletefile to remove hidden chunks for a file
> * a flag to reveal hidden chunks in listings
> * use DirCache as the key operations scan directories a lot (added on 2019-09-17)
> * future metadata format (named eg. "metajson") that embraces small files making hashsums more consistent; simplejson and metajson should be interoperable at reading objects (metajson can become the default format in a future rclone release, but it's better to avoid rclone.conf migrations if possible) and reliably detect single-chunked objects with or without metadata (eg currently there is a "_TODO maybe it's a small single chunk_" in Fs.Object [8820d22#diff-a132438692a9c451a2a1cf6aebcb3c3eR625](https://github.com/rclone/rclone/commit/8820d224d99e41e4895ee5892192f0446f317d79#diff-a132438692a9c451a2a1cf6aebcb3c3eR625)), the differenence is just about preferred put/update method.
> * metajson can use metadata field "ver" to detect whether single-chunk object is allowed to have metadata, eg. ver=0 cannot (added on 2019-09-18)
> * metadata logic is tightly coupled into main code a.t.m, moving it into a strategy pattern like upload/download readers would be nice

I had a look through the code and didn't see anything I particularly thought needed changing.

I'd like to merge this so we have time for a bit of testing before 1.50.

I also have some changes I'd like to merge which will require a bit of work on mailru/chunker before 1.50 too (see [filename-encode](https://github.com/rclone/rclone/tree/filename-encode) branch and how far I got so far with [the integration tests](https://pub.rclone.org/integration-tests/2019-09-21-132013/)) so getting this in sooner and fixing it up after would be my preference :-)

ivandeex:
Thanks. Please merge now as you decided.

I will then submit another PR with _clear points_ (please make sure you review it _before_ 1.50):
- to keep handling of empty vs unsupported hashsums as is
- md5quick/sha1quic will stay
- to remove fastopen downloader, keep linear downloader (even though there is no strategy switch anymore, I'll keep it as an separate structure created and returned by Object.Open, I think it looks better)
- "hash_type" will become "Advanced: false"
- Object.SetTier will be removed, the interface will be marked as unsupported in fstest.Opt
- "ver" will be added to meta json, marshaller will return a error if ver!=0 (and btw will do strict checks on other fields eg size>=0, nchunks>0, md5/sha1 are empty or valid hashes)
- wdmrccompat will be removed
- documentation will warn that chunker is beta and structures can break in future

Now on to _vague points_:
- Please see this commit https://github.com/ivandeex/rclone/commit/72c15b4ffb425933138e7c09005ca582d13994af to better understand what I mean by `if and only if (iff)`. It further munges the notorious rename counters, I just tried to replace nested ifs by ... (you'll see). It's ugly but introducing the `fs.ChunkSizer` interface to make a lone track-rename test happy, is much worse.
- Whether you fix enormous S3 buffer or not (it's up to you even to replace golang allocator by `--use-mmap`...), I still suggest to merge the
 memory pressure patch https://github.com/rclone/rclone/pull/3503/commits/f37c9229e9a9d4b1de8cd85673b6bc7e252c7726 and shift problematic tests to the end of execution.

ncw:
Ok sounds like a good plan!  I'll merge this now.  I'll try to can the s3 memory usage properly (see #/3568)

Can you paste suggested config for the backends in `config.yml` so I can add them to the integration test box?

ivandeex:
> Can you paste suggested config for the backends in `config.yml` so I can add them to the integration test box?

```
[TestChunkerLocal]
type = chunker
remote = /tmp

[TestChunkerChunk3bLocal]
type = chunker
remote = /tmp
chunk_size = 3b

[TestChunkerNometaLocal]
type = chunker
remote = /tmp
meta_format = none
hash_type = none

[TestChunkerChunk3bNometaLocal]
type = chunker
remote = /tmp
chunk_size = 3b
meta_format = none
hash_type = none

[TestChunkerMailru]
type = chunker
remote = TestMailru:

[TestChunkerChunk50bMailru]
type = chunker
remote = TestMailru:
chunk_size = 50b

[TestChunkerChunk50bYandex]
type = chunker
remote = TestYandex:
chunk_size = 50b

[TestChunkerChunk50bBox]
type = chunker
remote = TestBox:
chunk_size = 50b

[TestChunkerS3]
type = chunker
remote = TestS3:

[TestChunkerChunk50bS3]
type = chunker
remote = TestS3:
chunk_size = 50b

[TestChunkerChunk50bMD5HashS3]
type = chunker
remote = TestS3:
chunk_size = 50b
hash_type = md5

[TestChunkerChunk50bSHA1HashS3]
type = chunker
remote = TestS3:
chunk_size = 50b
hash_type = sha1

[TestChunkerChunk50bMD5QuickS3]
type = chunker
remote = TestS3:
chunk_size = 50b
hash_type = md5quick

[TestChunkerChunk50bSHA1QuickS3]
type = chunker
remote = TestS3:
chunk_size = 50b
hash_type = sha1quick

[TestChunkerCompatLocal]
type = chunker
remote = /tmp
meta_format = wdmrcompat
hash_type = none

[TestChunkerChunk3bCompatLocal]
type = chunker
remote = /home/deex/devel/rclone-vag1/_junk
chunk_size = 3b
meta_format = wdmrcompat
hash_type = none
```

Note: The next chunker PR will remove `wdmrcompat` and suggest removing two bottom remotes: `TestChunkerCompatLocal` and `TestChunker3bCompatLocal`

ncw:
Thanks.  I've added those.

Check out https://pub.rclone.org/integration-tests/current/ after 0800 GMT tomorrow.

PR #3504, opened 2019-09-02
vfs: add flag --vfs-case-insensitive for windows mounts

ivandeex:

This change solves a particular problem with rclone mount.
I use rclone on windows to mount a cloud file archive and need an automated scriptable way to verify file consistency. Windows provides `fciv` for that, a functional equivalent of `sha1sum` / `md5sum` Linux utilties but with a different format of directory checksums. Linux utilities use hex-encoded one-file-per-line format and `fciv` uses base64 encoded xml based format. `fciv ` is used daily by many windows devops to verify large archives for consistency. I use automated powershell scripts to run it against the archive.

However, there is a gotcha: fciv brings file names to lower case for compatibility with previous windows versions. This is not a problem if I mount archive using such programs as WebDrive or NetDrive but breaks with rclone. I tried to tweak a case-insensitivity option in WinFSP, the rclone mount layer for Windows, but to no avail. This patch fixed the problem for me and did not introduce problems with the mount. I believe it can solve other similar use cases, eg. with video or dlna players.

How to reproduce: 
- create a test webdav remote
- open powershell cli window on windows
```
choco install -y fciv
rclone mount webdav_test: m:
cd m:
ls
echo "abc" > TestCase1.txt
fciv.exe TestCase1.txt
```
- fciv fails trying to open `testcase1.txt\*`.
- stop rclone and proceed:
```
rclone mount webdav_test: m: --vfs-case-insensitive
cd m:
fciv.exe TestCase1.txt
```
Now fciv can successfully calculate checksum.

Note: I have not yet added documentation or unit tests for the flag as I think we need to discuss the hack first.

Probably yes, but I cannot find.

#### Checklist

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -40,6 +40,7 @@ build_script:
 
 test_script:
 - make GOTAGS=cmount quicktest
+- make GOTAGS=cmount racequicktest
 
 artifacts:
 - path: rclone.exe
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -58,7 +58,6 @@ strategy:
       GO_INSTALL_ARCH: 386
       BUILD_FLAGS: '-include "^windows/386" -cgo'
       MAKE_QUICKTEST: true
-      MAKE_RACEQUICKTEST: true
       DEPLOY: true
     other_os:
       imageName: ubuntu-16.04
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,6 +49,7 @@ strategy:
       GO_VERSION: latest
       BUILD_FLAGS: '-include "^windows/amd64" -cgo'
       MAKE_QUICKTEST: true
+      MAKE_RACEQUICKTEST: true
       DEPLOY: true
     windows_386:
       imageName: windows-2019
@@ -57,6 +58,7 @@ strategy:
       GO_INSTALL_ARCH: 386
       BUILD_FLAGS: '-include "^windows/386" -cgo'
       MAKE_QUICKTEST: true
+      MAKE_RACEQUICKTEST: true
       DEPLOY: true
     other_os:
       imageName: ubuntu-16.04
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -291,16 +291,19 @@ If an upload or download fails it will be retried up to
 rclone mount remote:path /path/to/mountpoint [flags]
 ```
 
-### Case Insensitive Mounts
+### Case Sensitivity
 
-Windows is not like other operating systems supported by rclone.
-Linux and MacOS file systems are case-sensitive: two files can differ
-only by case, and the exact case must be used when opening a file.
+Linux file systems are case-sensitive: two files can differ only
+by case, and the exact case must be used when opening a file.
 
+Windows is not like most other operating systems supported by rclone.
 File systems in modern Windows are case-insensitive but case-preserving:
 although existing files can be opened using any case, the exact case used
 to create the file is preserved and available for programs to query.
 It is not allowed for two files in the same directory to differ only by case.
+
+Usually file systems on MacOS are case-insensitive. It is possible to make MacOS
+file systems case-sensitive but that is not the default
 
 The `--vfs-case-insensitive` mount flag controls how rclone handles these
 two cases. If its value is `false`, rclone passes file names to the mounted
@@ -316,9 +319,13 @@ transparently fixup the name. This fixup happens only when an existing file
 is requested. Case sensitivity of file names created anew by rclone is
 controlled by an underlying mounted file system.
 
-If the flag is not provided on command line, then its default value depends on
-the operating system running rclone: `true` on Windows and `false` otherwise.
-If the flag is provided without a parameter, then its value is `true`.
+Note that case sensitivity of the operating system running rclone (the target)
+may differ from case sensitivity of a file system mounted by rclone (the source).
+The flag controls whether "fixup" is performed to satisfy the target.
+
+If the flag is not provided on command line, then its default value depends
+on the operating system where rclone runs: `true` on Windows and MacOS, `false`
+otherwise. If the flag is provided without a value, then it is `true`.
 
 ### Options
 
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -287,10 +287,38 @@ This mode should support all normal file system operations.
 If an upload or download fails it will be retried up to
 --low-level-retries times.
 
-
 ```
 rclone mount remote:path /path/to/mountpoint [flags]
 ```
+
+### Case Insensitive Mounts
+
+Windows is not like other operating systems supported by rclone.
+Linux and MacOS file systems are case-sensitive: two files can differ
+only by case, and the exact case must be used when opening a file.
+
+File systems in modern Windows are case-insensitive but case-preserving:
+although existing files can be opened using any case, the exact case used
+to create the file is preserved and available for programs to query.
+It is not allowed for two files in the same directory to differ only by case.
+
+The `--vfs-case-insensitive` mount flag controls how rclone handles these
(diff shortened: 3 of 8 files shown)


ncw:
This looks like a useful fix to me - thank you :-)

Some thoughts...
- Do you think the flag should default to `true` on windows?
- a linear search through d.items could be slow, but the way you've written it is that this is only called as a fallback so that is probably OK
- I'm not sure about returning ENOENT if the file is duplicated, maybe that should return EIO or maybe just return the first match?  What do you think?
- I did a quick review and I couldn't see any other places in the vfs that would need to be changed, but I could be wrong!

Docs and a test would be great!

ivandeex:
All my windows mounts have this flag for three months. No harm, runs good.
The question is - how to define "Windows"? An .exe build of rclone?
Will think about EIO (or EINVAL?)
Will do docs and a test.

ncw:
runtime.GOOS == "windows" is a pretty good test and it is also a constant.

ivandeex:
A note on error code.
I can't find a specification for stat() call on Windows. POSIX can be found here https://pubs.opengroup.org/onlinepubs/9699919799/ and concretely for stat() family here https://pubs.opengroup.org/onlinepubs/9699919799/functions/stat.html.
EIO seems like a good match but it is not yet defined in rclone/vfs/errors.go
EBADF is another candidate, its error message matches our case, it is already present in vfs/errors.go, but POSIX defines it only for fstatat(), not for stat(). I'd prefer EBADF.

ivandeex:
As we are going to make --vfs-case-insenstive=true the default on Windows and thus Windows diverges from Linux/Mac, I'll add racequicktest in Appveyor script.
```
test_script:
- make GOTAGS=cmount quicktest
- make GOTAGS=cmount **racequicktest**
```
Hope it includes full VFS tests too.

ncw:
> A note on error code.
> I can't find a specification for stat() call on Windows. POSIX can be found here https://pubs.opengroup.org/onlinepubs/9699919799/ and concretely for stat() family here https://pubs.opengroup.org/onlinepubs/9699919799/functions/stat.html.
> EIO seems like a good match but it is not yet defined in rclone/vfs/errors.go

I think if you just return a plain go error, then the FUSE layer will translate it to EIO.

> EBADF is another candidate, its error message matches our case, it is already present in vfs/errors.go, but POSIX defines it only for fstatat(), not for stat(). I can use EBADF or add EIO in errors.go if you like or if testing with WinFSP detects any problems.

Let's try just returning a nice go error `errors.Errorf("duplicate filename %q detected with --vfs-case-insensitive set", fileName)` - I'm pretty sure that will return EIO to userspace and make an ERROR log on linux & windows.

> As we are going to make --vfs-case-insenstive=true the default on Windows and thus Windows diverges from Linux/Mac, I'll add racequicktest in **Appveyor** script.

Good idea.

You could add it to Azure pipelines too with

```diff
diff --git a/azure-pipelines.yml b/azure-pipelines.yml
index 42a0cd138..512a7e0af 100644
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,6 +49,7 @@ strategy:
       GO_VERSION: latest
       BUILD_FLAGS: '-include "^windows/amd64" -cgo'
       MAKE_QUICKTEST: true
+      MAKE_RACEQUICKTEST: true
       DEPLOY: true
     windows_386:
       imageName: windows-2019
```

ivandeex:
The unit test creates test files whose names differ only by case. This pre-condition is expected to fail on Windows and I skip the test there. But it also fails on MacOS. Is MacOS filesystem case-insensitive-case preserving too? Can I skip the test on MacOS? Here is the log https://travis-ci.org/rclone/rclone/jobs/580412958

ivandeex:
TestVFSCaseInsensitive fails for Mailru backend. This means that Mailru cloud is case-insensitive. Thus, we need to identify all case-insensitive backends and ignore this test for them.

ivandeex:
Now case sensitivity test detects whether underlying remote file system is case-sensitive. If not, the test skips itself (on Windows, MacOS and some cloud providers).
The CaseInsensitive flag is by default set on Windows.
Though MacOS file system is not fully case-sensitive the CaseInsensitive flag is not set by default, as per documentation.
All test suites now pass OK. Please review.

ncw:
> The unit test creates test files whose names differ only by case. This pre-condition is expected to fail on Windows and I skip the test there. But it also fails on MacOS. Here is the log https://travis-ci.org/rclone/rclone/jobs/580412958
> Is MacOS filesystem case-insensitive-case preserving too? (I don't know this OS, I though it is based on case-sensitive BSD)

I had forgotten about macOS... The usual file systems for macOS are case-insensitive.  It is possible to make macOS file systems case-sensitive but that isn't the default and I understand it isn't very common.

See local backend setting of case insensitive flag

https://github.com/rclone/rclone/blob/b3db38ae3150aad2764a50728e9fddaa3a17ce31/backend/local/local.go#L252-L264

> Can I skip the test on MacOS?

Yes

> Do you think we should set CaseInsensitive to true on "darwin" as well?

I think we probably should, yes.

In which case the code and docs need a small tweak.

> TestVFSCaseInsensitive fails for Mailru backend. This means that Mailru cloud is case-insensitive on its own. I will mention this in documentation.

You should set the `Features` flag `CaseInsensitive` also.

> Thus, we need to identify all case-insensitive backends and ignore this test for them and for all overlay test remotes that use underlying case-insensitive backends for integration tests. :O

You can read the `f.Features().CaseInsensitive` flag to determine whether a remote is CaseInsensitive or not.

It isn't 100% reliable (see macOS discussion above!) but it works elsewhere in the tests.

> A simpler way is to skip the test if precondition fails. I have added a comment on the line that fails on case-insensitive backends (see https://github.com/rclone/rclone/pull/3504/files#diff-6cb1692bd88d435baf2c05ce9f4702d7R328):

That is another way too :-)

It might be easier to skip on the feature flag though.

ivandeex:
The f.Features().CaseInsensitive for Mailru will be updated by relevant PR.
The case sensitivity test passes using auto-detection.
--vfs-case-insensitive will be set by default on MacOS too, the docs have been updated.
All review comments have been addressed.

ncw:
That is great - thank you very much :-)

I'll merge this now.

Check https://pub.rclone.org/integration-tests/current/ tomorrow to see if any backends got broken!

ncw:
Grrr, palm clicked the submit button before I finished editing the commit message!

All merged now :-)

ncw:
It looks like [all the new tests passed](https://pub.rclone.org/integration-tests/current/) except for box

image

```
"./vfs.test -test.v -test.timeout 30m0s -remote TestBox: -verbose -test.run '^TestCaseSensitivity$'" - Starting (try 5/5)
=== RUN   TestCaseSensitivity
--- FAIL: TestCaseSensitivity (6.99s)
    run.go:175: Remote "box root 'rclone-test-nawawup8sufawaw1sawoxur3'", Local "Local file system at /tmp/rclone040288654", Modify Window "1s"
    run.go:269: Failed to put "FilEb" to "box root 'rclone-test-nawawup8sufawaw1sawoxur3'": Error "item_name_in_use" (409): Item with the same name already exists
FAIL
2019/09/05 11:13:23 DEBUG : box root 'rclone-test-nawawup8sufawaw1sawoxur3': Purge remote
"./vfs.test -test.v -test.timeout 30m0s -remote TestBox: -verbose -test.run '^TestCaseSensitivity$'" - Finished ERROR in 8.554789525s (try 5/5): exit status 1: Failed [TestCaseSensitivity]
```

So it looks like box didn't allow the file to be overwritten by a file of the new name.

Perhaps this is a bug in the box backend?

ivandeex:
The error happens here: https://github.com/rclone/rclone/blob/master/vfs/vfs_case_test.go#L27
We should either fix it or ignore it.

ivandeex:
I propose to return to your original poposal to skip this test if `f.Features().CaseInsensitive == true`.
Let's add this check and do `t.Skipf(...)` at the very top of TestCaseSensitivity, before test file creation.

PR #3516, opened 2019-09-05
vfs: completely skip TestCaseSensitivity on Box.com

ivandeex:

TestCaseSensitivity fails at the preparation stage for a Box.com backend when it tries to create two files with names different by case only. This is not a bug of Box but a rather tough variant of a case-insensitive file system. This PR skips this test early (even before preparation stage) for Box and all officially (according to features) case-insensitive backends.

See https://github.com/rclone/rclone/pull/3504#issuecomment-528364299 and https://github.com/rclone/rclone/pull/3504#issuecomment-528556158

#### Checklist


ncw:
Thank you :-)

Issue #3460, opened 2019-08-20
Create Automated Build on Docker Hub


pataquets:
Docker Hub allows you to create Automated Builds from source: https://docs.docker.com/docker-hub/builds/
It would add another packaging/distribution/installation method, whose buildings would be triggered automatically on each commit. It also allows to create different image tags automatically from git tags & branches.

By making the image build via an AB, you give the resulting image verifiability and auditability. Also, the build is fully automatic. You can have the latest image tag build from HEAD and individual image tags from git's release tags. Some people avoid non-verifiable (manually uploaded) images due to security & traceability reasons.

After #3459 is in, just a free Docker Hub account and a quick setup would do. Ping me if you need help.

ncw:
This seems like a good plan!

Would you help setup/maintain the system?

I made an rclone organization on docker hub - I can invite you there.

I'm not terribly experienced with docker though so could do with help!

> You can have the latest image tag build from HEAD and individual image tags from git's release tags. 


Can you explain how that would work?  Ideally I'd like a stream of releases and another stream of commits to master which I'd normally call the beta release.

> Some people avoid non-verifiable (manually uploaded) images due to security & traceability reasons.

That is probably a good reason to do this.

I'll comment some more on the PR

pataquets:
It can be done.
Docker images have tags, which are different versions, similar as git.
You can set up ABs to trigger builds and generate different image tags based on git branches and tags. You can create rules to tailor it to your needs.
One very obvious rule would be a 1 to 1 match of git tags and image tags, for having tagged images for each release, creating image tags from git branch and tag names.
A Docker image is made of `imagename:tag` and, if you omit the tag, it always assumes `latest`.
I would go for linking master (devel head) git branch to `:latest` image tag, but you might prefer to link it to the latest stable (v.1.48) release tag (which would involve updating the latest stable tag field on Docker Hub on each release).
With the exception of the `:latest` tag above, everything can be set up in a way that it's all triggered automatically. I expect no further involvement necessary, since I can't always commit time to maintainership, but it should be pretty easy.
As for now, if you feel like it, you can link the Github repo to a new AB and leave everything as default, so we get instantly a devel image. :)

pataquets:
Edited mentioning benefits for images listed on the wiki.

ncw:
I have merged #3459 which is a great start!

I've made the auto build work here: https://cloud.docker.com/u/rclone/repository/docker/rclone/rclone

[...]

PR #3521, opened 2019-09-08
Contrib: add Docker Compose sample manifests for DLNA and WebDAV servers

pataquets:
As discussed in #3460.
Docker Compose sample manifests.
See inline comments for details.
Usage should be fairly straightforward for Docker users, just doing `docker-compose -f file.yml up [-d]`.


ncw:
That is great - thank you :-)

Will merge now.

PR #3527, opened 2019-09-10
operations: dedupe on merge continues on errors like insufficientFilePermisson - fixes #3470

### LINKED ISSUE #3470: `Dedupe` could continue on errors like `insufficientFilePermisson`
rclone v1.48.0-211-g82c6c77e-beta
- os/arch: linux/amd64
- go version: go1.12.9

Rclone dedupe termintates completely if it encounters an insufficientFilePermisson error,
shouldn't it just skip the files it cannot edit/delete and continue operation on the rest of the files?

"rclone dedupe will terminate if any of the commands it does fail... This isn't ideal!
It could be made to continue some of the errors quite easily I think."

See: https://forum.rclone.org/t/dedupe-terminating-on-insufficientfilepermisson-error/11368

SezalAgrawal:

Rclone dedupe termintates completely if it encounters an insufficientFilePermisson error on merging duplicate folders and does not check for other duplicate files. The change in this PR logs this error and count of this error instead and continues with other dedupe tasks.

#3470 
#### Checklist


fs/operations/dedupe.go
```diff
@@ -242,31 +243,24 @@ func dedupeMergeDuplicateDirs(ctx context.Context, f fs.Fs, duplicateDirs [][]fs
...
-	// until all fixed
-	for {
-		duplicateDirs, err := dedupeFindDuplicateDirs(ctx, f)
-		if err != nil {
-			return err
-		}
-		if len(duplicateDirs) == 0 {
-			break
-		}
+	// Find duplicate directories first and fix them
```
ncw: I'm not sure why you changed this bit?

The original code will loop until all duplicates are found (which is necessary since we only fix one level at once).

Your change looks like it will only loop once?

SezalAgrawal: @ncw,

I tested with this directory structure:

**dummy/dummy_1.txt
dummy/dummy_2.txt
dummy/test/test1/a.txt
dummy/test/test1/b.txt**

This is the log I received:

2019/09/16 00:14:26 DEBUG : rclone: Version "v1.49.2-DEV" starting with parameters ["./test/rclone" "dedupe" "first" "remote2:" "-vv"]
2019/09/16 00:14:26 DEBUG : Using config file from "/Users/sezalagrawal/.config/rclone/rclone.conf"
2019/09/16 00:14:26 INFO  : Google drive root '': Looking for duplicates using first mode.
2019/09/16 00:14:26 DEBUG : remote2: Loaded invalid token from config file - ignoring
2019/09/16 00:14:27 DEBUG : remote2: Saved new token in config file
2019/09/16 00:14:56 INFO  : _dummy/test/test1: Merging contents of duplicate directories
2019/09/16 00:14:57 INFO  : dummy/test/test1: merging "a.txt"
2019/09/16 00:14:57 ERROR : merge duplicate dirs: MergeDirs move failed on "a.txt" in dummy/test/test1: googleapi: Error 403: Insufficient Permission: Request had insufficient authentication scopes., insufficientPermissions_
2019/09/16 00:14:57 INFO  : _dummy: Merging contents of duplicate directories
2019/09/16 00:14:57 INFO  : dummy: merging "dummy_1.txt"
2019/09/16 00:14:57 ERROR : merge duplicate dirs: MergeDirs move failed on "dummy_1.txt" in dummy: googleapi: Error 403: Insufficient Permission: Request had insufficient authentication scopes., insufficientPermissions_
2019/09/16 00:14:59 DEBUG : Untitled form: No export formats found for "application/vnd.google-apps.form"
.....

As you can see, it checks for each levels of directory structure. I removed the for loop because if there was some error (in this case Error 403: Insufficient Permission), the duplicates could not be removed and the for loop ran endlessly.

Can you tell me with what directory structure should I test?

SezalAgrawal: Hey @ncw, did you get a chance to look into it?

ncw: Sorry, I missed your message.

I think what you have written will work provided that the directories are returned in top down order (so parent before children) from `dedupeFindDuplicateDirs`.  This isn't guaranteed because the `dirs` map can be in any order.

https://github.com/rclone/rclone/blob/e14d968f8d5fdba0d695749648c2b1be05a04fb0/fs/operations/dedupe.go#L202-L221

So probably what we should do is sort the keys from the `dirs` map in `dedupeFindDuplicateDirs` and range through them in sorted order when we build `duplicateDirs`.  This will guarantee this ordering and we can remove the recursion.  This will make the process more efficient.

So something like

```go
        // make sure parents are before children
        duplicateNames := []string{}
	for name, ds := range dirs {
		if len(ds) > 1 {
			duplicateNames = append(duplicateNames, name)
		}
	}
        sort.Strings(duplicateNames)
	duplicateDirs := [][]fs.Directory{}
	for _, name := range duplicateNames {
  	    duplicateDirs = append(duplicateDirs, dirs[name])
	}
```

fs/operations/dedupe.go
```diff
@@ -235,7 +235,8 @@ func dedupeMergeDuplicateDirs(ctx context.Context, f fs.Fs, duplicateDirs [][]fs
 			fs.Infof(dirs[0], "Merging contents of duplicate directories")
 			err := mergeDirs(ctx, dirs)
 			if err != nil {
-				return errors.Wrap(err, "merge duplicate dirs")
+				fs.CountError(err)
```
golangcibot: Error return value of `fs.CountError` is not checked (from `errcheck`)

ncw:
That looks great now :-)

Sorry this PR has taken such a long time to complete.

I'll merge it now

PR #3544, opened 2019-09-17
Docker container fixes

mateidavid:

Specifically:
- enable Docker container userspace operation
- enable Docker userspace mount exposed to the host
- add more Docker image usage documentation

The changes affect only the Docker image. I haven't noticed any CI/CD pipeline checking the Docker image. (Perhaps that should be a task in itself.)

Yes: https://forum.rclone.org/t/run-docker-container-in-userspace/11734/7

#### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,10 +12,11 @@ RUN ./rclone version
 # Begin final image
 FROM alpine:latest
 
-RUN apk --no-cache add ca-certificates
+RUN apk --no-cache add ca-certificates fuse
 
-WORKDIR /root/
+COPY --from=builder /go/src/github.com/rclone/rclone/rclone /usr/local/bin/
 
-COPY --from=builder /go/src/github.com/rclone/rclone/rclone .
+ENTRYPOINT [ "rclone" ]
 
-ENTRYPOINT [ "./rclone" ]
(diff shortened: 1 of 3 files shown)


ncw:
That looks really good thank you and excellent documentation :-)

I'll merge that now!

mateidavid:
Thanks.

Just a quick note- I realized there's a typo in the doc file: It should read:

> You need to mount the host rclone config dir at `/config/rclone`.

instead of:

> You need to mount the host rclone config dir at `/config`.

However, this is seen in the example commands right below.

I'm not sure if this warrants another commit, let me know if you think so.

ncw:
> I'm not sure if this warrants another commit, let me know if you think so.

It is probably worth fixing up so if you want to send another PR it will be easy to merge!

jeroenvdheuvel:
@ncw can this change be merged into the `v1.49-fixes` branch? I'm currently using the beta image because there's no stable version with this fix.

The fix only breaks BC for people that use docker. However, it allows the broken experience to work when you're executing it via a non-root user.

ncw:
I'm planning on releasing v1.50 in the next week or two so I'd rather wait until then.

Issue #111, opened 2015-08-25
Feature request: Copy custom metadata on objects


mlanner:
For object storage systems that support custom (user defined) metadata types, include a flag (`-meta`?) that will also copy custom metadata from source to destination. For example, if copying/migrating data from Amazon S3 to Swift, or from Swift to Swift, it would be very handy to be able to maintain any custom metadata set on the objects.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
@mlanner nice idea - I've wanted exactly this before when moving stuff about in Swift.  Not sure how I'd implement it - I'll have a think.

mlanner:
@ncw How about something like: If you invoked the `-meta` flag in your Rclone command, then it will head the object, read the custom metadata and then store it in a temporary json blob. Once the object has been transferred and the md5sum has been checked, the metadata stored in json will be applied through a post to the object in the destination container/bucket. ?? It might require a few extra operations, but it would only take a hit if you invoke the `-meta` flag, and the operations would (perhaps) not be ridiculously expensive.

Only an idea. I'm sure there might be better ways to do this, as I'm not much of a developer myself.

jamshid:
I wonder if metadata could be stored in file attributes (https://en.wikipedia.org/wiki/Extended_file_attributes), so we don't need a separate/temporary json file.

ncw:
@jamshid not a bad idea.. I wonder if it can be done in a cross platform way from go...

ncw:
An implementation idea - we pass the source object to the Put methods.  Do a type assertion at that point to break out the metadata - not too tricky.

freedge:
> could be stored in file attributes

@jamshid  @ncw that's how it's done when using [JClouds with the "filesystem" blob store ](https://jclouds.apache.org/guides/filesystem/) implementation. For example:

```
$ attr -l ./Package_98/Package_98_version_13
Attribute "user.content-md5" has a 16 byte value for ./Package_98/Package_98_version_13
Attribute "user.user-metadata.sha1" has a 3 byte value for ./Package_98/Package_98_version_13
Attribute "user.content-type" has a 19 byte value for ./Package_98/Package_98_version_13
[...]

PR #3555, opened 2019-09-20
Add support of checksum check on crypted remote

dr1rrb:
This allow users to store the MD5 of the un-encrypted file content as a metadata on a azure blob storage, so we are still able to use the `--checksum` parameter for any command.

When a remote is encrypted, checksums is usually just ignored by all commands and the only way to check the diff between source and remote is to use the `cryptcheck`. This command will try to read some metadata at the beginning of the remote file, decrypt them, and then validate the hash. 

This however is not a suitable solution when a remote is (kind of) write-only. For instance if you configure your backups to use the [archive tier](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-storage-tiers#archive-access-tier) on a azure blob backend, you need re-hydrate your data before reading it ... which usually takes few hours!

#### Why ?
My own use-case is to backup my personal photos library:

* I obviously want to encrypt all my photos locally before sending them to azure; 
* As photos are [`--immutable`](https://rclone.org/docs/#immutable), the cheap archive tier fits perfectly;
* I would like to run a "full backup" once from time to time with checksum validation to make sure that the data does not get corrupted for any reason (which happened to me last winter due to a failing disk on my local server 😭).

Not yet ... that's why I opened this PR as a draft only

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -42,6 +42,7 @@ const (
 	decayConstant         = 1    // bigger for slower decay, exponential
 	maxListChunkSize      = 5000 // number of items to read at once
 	modTimeKey            = "mtime"
+	md5Key                = "md5"
 	timeFormatIn          = time.RFC3339
 	timeFormatOut         = "2006-01-02T15:04:05.000000000Z07:00"
 	maxTotalParts         = 50000 // in multipart upload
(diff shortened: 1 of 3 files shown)


ncw:
This is a very interesting idea :-) I think we need to a bit of work to fit that into rclone in a sustainable way.

I'm not that keen on storing plain MD5's though - that is an info leak you could drive a bus through!  What about storing them encrypted? Crypt could run hashes available on the source through secretbox then base64 that which I don't think would make them too big.

I've been thinking about making a metadata framework for rclone in #111 and other issues.  I think the idea I've settled on is that an Object would implement a `ReadMetadata` call if possible.  This would return a `map[string]string`.

Local backends could return username, userid permissions, xattrs.  Bucket based backends could return their metadata.  Then `Put` and `Update` could optionally take metadata to store on the new object.

I would propose not to implement a `SetMetadata` and sync metadata, this would just be on Put/Update of new objects. (Maybe - for discussion!)

With that in place crypt could be reading metadata of the underlying object and encrypting that (or not) and passing it through.  That would be the point to add the underlying hashes from the object.

What would you think of working on the metadata framework with me first?

ivandeex:
Can we give user a choice between md5/sha1? Seems not a big piece of work, both hashes are widely used by cloud providers.

ncw:
> Could you give user a choice between md5/sha1? Seems not a big piece of work, both hashes are widely used by cloud providers.

That seems OK!

There is some discussion needed on how to make this supportable by all rclone backends first though!

ivandeex:
My bad. I didn't notice you are using wrapped Fs as meta storage and though by mistake that you use some metadata container in rclone's crypt format.

dr1rrb:
Yeah @ncw  you're right that my approach is really limited and we should be able to store any kind of metadata, not only the original file hash like I did. I would also agree that meta should be encrypted in case of an encrypted backend.

I'll be glad to help you with that (even if go is not my first dev language), but I'll leave for vacation soon ... so I won't have time before November.

In the meantime, I'll close my draft as we all agree that it's not the right approach.

ncw:
Thanks for opening this discussion.  I'm going to be thinking about metadata in general soon and I've got this on my radar now :-)

PR #3560, opened 2019-09-20
Write the template to a buffer to catch render errors

### LINKED ISSUE #3559: superfluous response.WriteHeader call in httplib
rclone serve http throws errors when using _lots_ of connections

1.49.0

Debian 64bit

Local

rclone serve http local: --transfers=9999 --checkers=9999

http: superfluous response.WriteHeader call from github.com/rclone/rclone/cmd/serve/httplib/serve.Error (dir.go:72)

fionera:
Fix #3559

#3559

#### Checklist

### DIFF
--- a/cmd/serve/httplib/serve/dir.go
+++ b/cmd/serve/httplib/serve/dir.go
@@ -1,6 +1,7 @@
 package serve
 
 import (
+	"bytes"
 	"fmt"
 	"html/template"
 	"net/http"
@@ -69,7 +70,9 @@ func (d *Directory) AddEntry(remote string, isDir bool) {
 func Error(what interface{}, w http.ResponseWriter, text string, err error) {
 	fs.CountError(err)
 	fs.Errorf(what, "%s: %v", text, err)
-	http.Error(w, text+".", http.StatusInternalServerError)
+	if w != nil {
+		http.Error(w, text+".", http.StatusInternalServerError)
+	}
 }
 
 // Serve serves a directory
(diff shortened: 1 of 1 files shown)


cmd/serve/httplib/serve/dir.go
```diff
@@ -80,9 +83,14 @@ func (d *Directory) Serve(w http.ResponseWriter, r *http.Request) {
...
-	err := d.HTMLTemplate.Execute(w, d)
+	buf := &bytes.Buffer{}
+	err := d.HTMLTemplate.Execute(buf, d)
 	if err != nil {
 		Error(d.DirRemote, w, "Failed to render template", err)
 		return
 	}
+	_, err = buf.WriteTo(w)
+	if err != nil {
+		Error(d.DirRemote, nil, "Failed to drain template buffer", err)
```
terorie: This doesn't solve the original problem. As soon as the first `w.Write` attempt has been made, `w.WriteHeader` was automatically called so `serve.Error` shouldn't be called in this case.

fionera: serve.Error wont write to the http response tho, instead its only in the log

terorie: Maybe update the doc then? :) `// Error returns an http.StatusInternalServerError and logs the error`

fionera: Done :smile:

terorie:
Quick idea – We can also do zero allocs if we create a custom `io.Writer` that remembers if any write occurred.

```go
type WriteChecker struct {
    written bool
}

func (w *WriteChecker) Write(p []byte) (n int, err error) {
    w.written = true
    return len(p), nil
}

func (w *WriteChecker) HasWritten() bool {
    return w.written
}

// ...

var checker WriteChecker
err := d.HTMLTemplate.Execute(buf, io.MultiWriter(&checker, w))
if err != nil {
    // ...
    !if checker.HasWritten() {
        serve.Error(...)
    }
}
```

ncw:
It seems wrong to turn a streaming process into something which uses a memory buffer just to avoid an error message.

However I can't think of anything better so I'll merge this  - thank you :-)

PR #3561, opened 2019-09-21
s3: fix signature v2_auth headers

antrusd:
When used with `v2_auth = true`, `PresignRequest` doesn't return
signed headers, so remote dest authentication would be fail.
This commit copying back `HTTPRequest.Header` to `headers`.

Tested with RiakCS v2.1.0.

Signed-off-by: Anthony Rusdi <33247310+antrusd@users.noreply.github.com>


ncw:
That looks like a good fix - thank you :-)

PR #3563, opened 2019-09-23
Plugin support

terorie:

Adds support for loading out-of-tree storage backends and commands using [Go plugins](https://golang.org/pkg/plugin/) on Linux and macOS without needing to recompile rclone.

#### Behavior

Plugins with a file name matching `librcloneplugin_XXX.so` inside `$RCLONE_PLUGIN_PATH` will be loaded and accessible like any other storage backend. Because loading third-party code is a potential security risk, plugin support is disabled by default. On unsupported operating systems, this variable is silently ignored.
In case loading a plugin fails (e.g. mismatched source revisions of system rclone and rclone the plugin was compiled against), a warning gets printed and the plugin is ignored.

#### Creating plugins

Plugins are written exactly the same as in-tree storage backends or commands with the exception that the package name must be `main`. Here's a quick start [(source)](https://gist.github.com/terorie/21b517ee347828e899e1913efc1d684f):

```sh
# Fresh rclone installation
git clone https://github.com/rclone/rclone && cd rclone

# Build plugin
mkdir plugins && cd plugins
wget https://gist.githubusercontent.com/terorie/21b517ee347828e899e1913efc1d684f/raw/8b4669a53082fac6d62e73c108ac0943c84e9ac4/minimal.go
go build -buildmode=plugin -o librcloneplugin_minimal.so minimal.go
cd ..

# Start rclone with plugin support enabled
export RCLONE_PLUGIN_PATH=./plugins
go build
./rclone config
```

Should this go in the Wiki?

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate. (I don't think we can add meaningful tests)


backend/plugin/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+	dir := os.Getenv("RCLONE_PLUGIN_PATH")
+	if dir == "" {
+		return
+	}
+	// Get file names of plugin dir
+	f, err := os.Open(dir)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Failed to open plugin directory:", err)
+	}
+	defer f.Close()
```
golangcibot: Error return value of `f.Close` is not checked (from `errcheck`)

ncw:
Nice work - I've been wanting to experiment with this!

I'd be interested to know your motivation for this?

Slimming down rclone would be one, as would keeping an out of tree backend.  Though there might be a better solution for the latter.

> Plugins with a file name matching `librcloneplugin_XXX.so`

Conceivably we could have rclone commands which work as plugins so maybe this should be `librclonebackendplugin_XXX.so` ?

terorie:
@ncw
> I'd be interested to know your motivation for this?

We use rclone with a few backends that will probably never get merged because they are either too silly (Telegram & Discord), application-specific (Redis, SQL) or would cause licensing issues (GPL + Copyright).
In short, it's easier to maintain a single backend than an entire fork or to wait until it gets reviewed and merged upstream.

> Slimming down rclone would be one

Creating a plugin for each backend would cause considerable bloat, they are >5 MB at least. If we split off some backends from the main repo, maybe in the form of collection packages inspired by GStreamer: `rclone-plugins-community`, `rclone-plugins-experimental`.

> as would keeping an out of tree backend. Though there might be a better solution for the latter.

That sounds interesting, would you mind giving more details?
I agree that plugins have some disadvantages, like needing to be compiled against the exact same revision of rclone.

> Conceivably we could have rclone commands which work as plugins so maybe this should be librclonebackendplugin_XXX.so ?

I would prefer keeping the hardcoded prefix as-is and just encourage good naming practices for greater flexibility. Otherwise, something likes this will probably happen: `librclonebackendplugin_something-that-is-not-a-backend.so`.

Here's what I had in mind with naming:
 - `librcloneplugin_backend_mystorage.so` (single backend)
 - `librcloneplugin_command_qt5gui.so` (single command)
 - `librcloneplugin_bundle_closedsourcecompany.so` (third-party backend/command collection)

ncw:
> > I'd be interested to know your motivation for this?
> 
> We use rclone with a few backends that will probably never get merged because they are either too silly (Telegram & Discord), application-specific (Redis, SQL) or would cause licensing issues (GPL + Copyright).
> In short, it's easier to maintain a single backend than an entire fork or to wait until it gets reviewed and merged upstream.

OK fair enough!

> > Slimming down rclone would be one
> 
> Creating a plugin for each backend would cause considerable bloat, they are >5 MB at least. If we split off some backends from the main repo, maybe in the form of collection packages inspired by GStreamer: `rclone-plugins-community`, `rclone-plugins-experimental`.

Interesting idea too!

> > as would keeping an out of tree backend. Though there might be a better solution for the latter.
> 
> That sounds interesting, would you mind giving more details?

If you just want an out of tree backend then you can arrange your code like this

```
rclone
- rclone.go
- backend
  - backend.go
```

In `rclone.go` you copy rclone.go from the rclone repo and add the import for your new backend.

Bulding this will make an rclone binary with your new backend.

> I agree that plugins have some disadvantages, like needing to be compiled against the exact same revision of rclone.

I do think they are an interesting idea though.

> > Conceivably we could have rclone commands which work as plugins so maybe this should be librclonebackendplugin_XXX.so ?
> 
> I would prefer keeping the hardcoded prefix as-is and just encourage good naming practices for greater flexibility. Otherwise, something likes this will probably happen: `librclonebackendplugin_something-that-is-not-a-backend.so`.
> 
> Here's what I had in mind with naming:
> 
> * `librcloneplugin_backend_mystorage.so` (single backend)
> * `librcloneplugin_command_qt5gui.so` (single command)
> * `librcloneplugin_bundle_closedsourcecompany.so` (third-party backend/command collection)

Sounds fine!

Would you might writing this up somewhere? Maybe in backend/plugin and we can just put a link to the go doc version of that in CONTRIBUTING?

terorie:
@ncw I wrote some docs in `CONTRIBUTING.md`, assuming the next version to include plugin support is v1.50. Let me know what you think of them.

I just noticed since plugins could be used to load more than just backends, maybe it makes sense to move the source to somewhere else than `backend`, maybe `lib`?

ncw:
> I wrote some docs in `CONTRIBUTING.md`, assuming the next version to include plugin support is v1.50. Let me know what you think of them.

Very nice docs - thank you :-)

> I just noticed since plugins could be used to load more than just backends, maybe it makes sense to move the source to somewhere else than `backend`, maybe `lib`?

Yes you are right this isn't backend specific...

I like your idea of moving it to `lib/plugin`

I suggest that we do the `_ include` in the root "rclone.go" file.

If you do that then I'll merge.  This is a nice feature to have, I'm sure it will make for some interesting rclone experiments!

terorie:
Done, I moved everything to `lib/plugin` and updated references :+1:

ncw:
I have merged this manually in 44b603d2bd7f063ade2cd5a9abb27c5153b9cda4 to fixup `rclone.org` needing goimports running on it.

Thank you very much for this - I think there will be lots of exiting uses of it!

PR #3582, opened 2019-10-01
[WIP] Move config file handling to external packages

fionera:
Replace the current goconfig handling with an abstraction to allow use of other config types. I only did Search and Replace for the tests, so I dont know if anything works yet. Also I dont know if this is even a good idea. I use viper for a fork I maintain to have a map[string][]string type for remotes and more flexible handling of these, so I thought maybe upstream it

I think I mentioned it somewhere :smile: 

#### Checklist


fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+		newSection.Set(k, v)
+	}
+}
+
+func GetConfigKey(args ...string) string {
+	return strings.Join(args, ".")
+}
+
+var (
+	_ config.Provider     = (*viperConfig)(nil)
```
golangcibot: File is not `goimports`-ed (from `goimports`)

fs/config/provider/goconfig/provider.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"bytes"
+	"io"
+	"log"
+
+	"github.com/Unknwon/goconfig"
+	"github.com/rclone/rclone/fs/config"
+)
+
+func Register() {
```
golangcibot: exported function `Register` should have comment or be unexported (from `golint`)

fs/config/provider/goconfig/provider.go
```diff
@@ -0,0 +1,56 @@
...
+	config.RegisterConfigProvider(&config.ProviderDefinition{
+		NewFunc:   NewGoConfigProvider,
+		FileTypes: []string{"ini", "conf"},
+	})
+}
+
+func NewGoConfigProvider() config.Provider {
```
golangcibot: exported function `NewGoConfigProvider` should have comment or be unexported (from `golint`)

fs/config/provider/goconfig/section.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (s *section) Get(name string) interface{} {
+	return s.GetStringDefault(name, "")
+}
+
+func (s *section) GetString(name string) string {
+	return s.GetStringDefault(name, "")
+}
+
+func (s *section) GetStringDefault(name string, default_ string) string {
```
golangcibot: don't use underscores in Go names; method parameter default_ should be default (from `golint`)

fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+	"log"
+	"strings"
+
+	"github.com/rclone/rclone/fs/config"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	RemotesPrefix = "remotes"
```
golangcibot: exported const RemotesPrefix should have comment (or a comment on this block) or be unexported (from `golint`)

fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (c *viperConfig) GetString(key string) string {
+	return c.GetViper().GetString(key)
+}
+
+func (c *viperConfig) SetString(key string, value string) {
+	c.GetViper().Set(key, value)
+}
+
+func NewViperProvider() config.Provider {
```
golangcibot: exported function `NewViperProvider` should have comment or be unexported (from `golint`)

fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+	return &viperConfig{}
+}
+
+func Register() {
```
golangcibot: exported function `Register` should have comment or be unexported (from `golint`)

fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (c *viperConfig) GetViper() *viper.Viper {
+	return viper.GetViper()
+}
+
+func (c *viperConfig) GetRemotes() []string {
+	var remotes []string
+	remoteEntries := viper.GetStringMap(RemotesPrefix)
+	for key, _ := range remoteEntries {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for key := range ...` (from `golint`)

fs/config/provider/viper/config.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	sourceConfig := viper.Sub(GetConfigKey(RemotesPrefix, source))
+	newSection := viper.Sub(GetConfigKey(RemotesPrefix, destination))
+
+	for k, v := range sourceConfig.AllSettings() {
+		newSection.Set(k, v)
+	}
+}
+
+func GetConfigKey(args ...string) string {
```
golangcibot: exported function `GetConfigKey` should have comment or be unexported (from `golint`)

fs/config/provider/viper/config_section.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (s *section) Get(name string) interface{} {
+	return s.v.Get(GetConfigKey(s.basePath, name))
+}
+
+func (s *section) GetString(name string) string {
+	return s.v.GetString(GetConfigKey(s.basePath, name))
+}
+
+func (s *section) GetStringDefault(name string, default_ string) string {
```
golangcibot: don't use underscores in Go names; method parameter default_ should be default (from `golint`)

fs/config/provider/viper/config_section.go
```diff
@@ -0,0 +1,96 @@
...
+	if s.v.IsSet(GetConfigKey(s.basePath, name)) {
+		return s.v.GetString(GetConfigKey(s.basePath, name))
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

fstest/fstest.go
```diff
@@ -8,6 +8,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/rclone/rclone/fs/config/provider/goconfig"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ncw:
> Replace the current goconfig handling with an abstraction to allow use of other config types. I only did Search and Replace for the tests, so I dont know if anything works yet. Also I dont know if this is even a good idea.

The config system was designed to be replaceable by setting these global variables

https://github.com/rclone/rclone/blob/77a520c97c95cb4e51dfa1e4a17752630051a802/fs/config.go#L16-L28

Now whether that is enough configuration I'm not sure since I don't think anyone used that feature...

That doesn't work with `rclone config` but should be enough for `library` use of rclone.

I think your patch goes one step further so `rclone config` would work with `viper` - is that right?

> I use viper for a fork I maintain to have a map[string][]string type for remotes and more flexible handling of these, so I thought maybe upstream it

I thought about using viper at some point but decided it would be too difficult to retro-fit.  I don't really know how it works though - how are you using it?

fionera:
> The config system was designed to be replaceable by setting these global variables

yes I saw that but since I did not wanted to use a string as storage, it wasnt enought. I wanted to use configs like this: https://i.imgur.com/QxaJpAt.png

> I think your patch goes one step further so rclone config would work with viper - is that right?

Yes I completely moved the file handling to a different package and made it switchable. The access is now over interfaces which also makes mocking far easier. 

> I don't really know how it works though - how are you using it?

It is basicly a wrapper around yaml, toml etc.

ncw:
> yes I saw that but since I did not wanted to use a string as storage, it wasn't enough.

I'm trying to understand in what way wasn't it enough?  What you mean by "use a string as storage"?

fionera:
I needed a map value and since I hate using CLIs I wanted to be able to still write the config manually. In the screenshot is an example of this

fionera:
@ncw Would it be better to split this into multiple smaller PRs? Like moving the Config System itself into a different file, abstracting it more etc.?

ncw:
> @ncw Would it be better to split this into multiple smaller PRs? Like moving the Config System itself into a different file, abstracting it more etc.?

Sorry I've been a bit distracted trying to firm up the 1.50 release!

Yes smaller PRs would be much easier to review.

Maybe first step you could post a `Config interface {}` type which we could discuss?

fionera:
> Maybe first step you could post a Config interface {} type which we could discuss?

Sure thing :) Will close this and open an Issue

PR #3605, opened 2019-10-09
filter: prevent mix opts when filesfrom is present - fixes #3599

### LINKED ISSUE #3599: [USABILITY] --files-from override all other filtering, but does not stop you from using both
Referring to this thread:
https://forum.rclone.org/t/why-do-we-ignore-other-filtering-rules-when-using-the-files-from-flag/12065/5

**Short synopsis of conclusion:**
--files-from overrides all other filtering. We concluded that this was intended behavior for complexity reasons, but it would be reasonable (to prevent user mistakes) to throw an error if user tries to use --files from with other filtering.

mcaci:

I would like to address the issue #3599 where one should forbid mixing --files-from opt with other one

Here's the link to the discussion: https://forum.rclone.org/t/why-do-we-ignore-other-filtering-rules-when-using-the-files-from-flag/12065/5

#### Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [ x ] I have added tests for all changes in this PR if appropriate.
- [ x ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ x ] I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/fs/filter/filter.go
+++ b/fs/filter/filter.go
@@ -191,7 +191,12 @@ func NewFilter(opt *Opt) (f *Filter, err error) {
 			return nil, err
 		}
 	}
+
+	inActive := f.InActive()
 	for _, rule := range f.Opt.FilesFrom {
+		if !inActive {
+			return nil, fmt.Errorf("The usage of --files-from overrides all other filters, it should be used alone")
+		}
 		f.initAddFile() // init to show --files-from set even if no files within
 		err := forEachLine(rule, func(line string) error {
 			return f.AddFile(line)
--- a/fs/filter/filter_test.go
+++ b/fs/filter/filter_test.go
@@ -42,7 +42,58 @@ func testFile(t *testing.T, contents string) string {
 	return s
 }
 
-func TestNewFilterFull(t *testing.T) {
+func TestNewFilterForbiddenMixOfFilesFromAndFilterRule(t *testing.T) {
+	Opt := DefaultOpt
+
+	// Set up the input
+	Opt.FilterRule = []string{"- filter1", "- filter1b"}
+	Opt.FilesFrom = []string{testFile(t, "#comment\nfiles1\nfiles2\n")}
+
+	rm := func(p string) {
+		err := os.Remove(p)
+		if err != nil {
+			t.Logf("error removing %q: %v", p, err)
+		}
+	}
+	// Reset the input
+	defer func() {
+		rm(Opt.FilesFrom[0])
+	}()
+
+	_, err := NewFilter(&Opt)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "The usage of --files-from overrides all other filters")
+}
+
+func TestNewFilterWithFilesFromAlone(t *testing.T) {
+	Opt := DefaultOpt
+
+	// Set up the input
+	Opt.FilesFrom = []string{testFile(t, "#comment\nfiles1\nfiles2\n")}
+
+	rm := func(p string) {
+		err := os.Remove(p)
+		if err != nil {
+			t.Logf("error removing %q: %v", p, err)
+		}
+	}
+	// Reset the input
+	defer func() {
+		rm(Opt.FilesFrom[0])
+	}()
+
+	f, err := NewFilter(&Opt)
+	require.NoError(t, err)
+	assert.Len(t, f.files, 2)
+	for _, name := range []string{"files1", "files2"} {
+		_, ok := f.files[name]
+		if !ok {
+			t.Errorf("Didn't find file %q in f.files", name)
+		}
+	}
+}
+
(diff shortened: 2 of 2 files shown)


fs/filter/filter.go
```diff
@@ -192,6 +192,10 @@ func NewFilter(opt *Opt) (f *Filter, err error) {
 		}
 	}
 	for _, rule := range f.Opt.FilesFrom {
+		if !isFilesFromOptUsedAlone(f.Opt) {
```
ncw: I think you could simplify this a bit by using `f.InActive()` to check if anything had been added to the filter, then you wouldn't need the `isFilesFromOptUsedAlone` function.

Something like

```go
inActive := f.InActive()
for _, rule := range f.Opt.FilesFrom {
    if !inActive {
        return nil, fmt.Errorf("...")
```

mcaci: Hi @ncw, you're very welcome and thanks for your review and advice.

I must confess a bit of initial skepticism for replacing my function with the InActive method because with a raw comparison of the two functions I was fearing that some Opt(s) might have not been checked.

Not a situation I would have expected: I trust, of course, more your suggestion than my raw comparison, but to make things completely straight I added an additional test, that passed with the code change you propose.

This is the test:

```go
func TestNewFilterForbiddenMixOfFilesFromAndFilterRule(t *testing.T) {
	Opt := DefaultOpt

	// Set up the input
	Opt.FilterRule = []string{"- filter1", "- filter1b"}
	Opt.FilesFrom = []string{testFile(t, "#comment\nfiles1\nfiles2\n")}

	rm := func(p string) {
		err := os.Remove(p)
		if err != nil {
			t.Logf("error removing %q: %v", p, err)
		}
	}
	// Reset the input
	defer func() {
		rm(Opt.FilesFrom[0])
	}()

	_, err := NewFilter(&Opt)
	require.Error(t, err)
}
```

In conclusion, I totally approve your suggestion and I ask you whether you're ok that I also add this test case or not. In general I would say yes, but maybe you reckon that the new `TestNewFilterFull` and `TestNewFilterFullExceptFilesFromOpt` already cover this situation sufficiently.

What's your view on this?

fs/filter/filter.go
```diff
@@ -559,3 +565,4 @@ func (f *Filter) UsesDirectoryFilters() bool {
 	}
 	return true
 }
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

fs/filter/filter.go
mcaci: Alright, thank you for your advice. I have just committed the change and the new test case

fs/filter/filter_test.go
```diff
@@ -42,6 +42,28 @@ func testFile(t *testing.T, contents string) string {
...
+			t.Logf("error removing %q: %v", p, err)
+		}
+	}
+	// Reset the input
+	defer func() {
+		rm(Opt.FilesFrom[0])
+	}()
+
+	_, err := NewFilter(&Opt)
+	require.Error(t, err)
```
ncw: You could add after this to make sure we get the right error

    require.Contains(t, err.Error(), "some substring from the error")

fs/filter/filter.go
```diff
@@ -191,7 +191,13 @@ func NewFilter(opt *Opt) (f *Filter, err error) {
 			return nil, err
 		}
 	}
+
+	inActive := f.InActive()
 	for _, rule := range f.Opt.FilesFrom {
+		if !inActive {
+			return nil, fmt.Errorf("The usage of --files-from overrides all other filters, it should be used alone")
+
```
ncw: Extra new line

mcaci: Yes, I will remove it. No need for that.

fs/filter/filter_test.go
mcaci: Yes, it makes sense, I will add this assertion

fs/filter/filter_test.go
```diff
@@ -42,6 +42,28 @@ func testFile(t *testing.T, contents string) string {
...
+}
+
 func TestNewFilterFull(t *testing.T) {
```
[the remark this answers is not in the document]
mcaci: It's true, I realize indeed that by modifying the tests in the two new ones I've lost the coverage of the case with the filter `--files-from` alone. I like the suggestion to repurpose the `TestNewFilterFull` to test this case alone. I will correct my code this way. Thanks for spotting this :)

ncw:

I'll merge it now.

PS I have a note of the test that failed being unreliable already - not your code!

mcaci:
Alright I was going to have a look at this failure but thanks for informing me. Have a nice weekend! :)

PR #3623, opened 2019-10-13
rc docs: fix code section not rendering properly due to missing quotes

negative0:

Adding quotes to fix the rendering issues in rc docs section

Screenshot from 2019-10-13 10-31-04


ncw:
Thank you very much :-)

PR #3628, opened 2019-10-15
mount: replace use of WriteAt with Write for files opened with O_APPEND

bdutro:
os.File.WriteAt returns an error if a file was opened with O_APPEND. Replacing it with os.File.Seek and os.File.Write for cache modes >= writes will ensure that writes will always work if a file is opened with O_APPEND.

Writing to a file opened with O_APPEND on a mounted remote (with caching >= writes) returns an error because os.File.WriteAt does not allow writing to files opened with O_APPEND.

No

#### Checklist

### DIFF
--- a/cmd/mountlib/mounttest/write.go
+++ b/cmd/mountlib/mounttest/write.go
@@ -2,6 +2,7 @@ package mounttest
 
 import (
 	"os"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -126,6 +127,51 @@ func TestWriteFileDup(t *testing.T) {
 
 	err = fh.Close()
 	require.NoError(t, err)
+
+	run.waitForWriters()
+	run.rm(t, "to be synced")
+}
+
+// TestWriteFileAppend tests that O_APPEND works on cache backends >= writes
+func TestWriteFileAppend(t *testing.T) {
+	run.skipIfNoFUSE(t)
+
+	if run.vfs.Opt.CacheMode < vfs.CacheModeWrites {
+		t.Skip("not supported on vfs-cache-mode < writes")
+		return
+	}
+
+	// TODO: Windows needs the v1.5 release of WinFsp to handle O_APPEND properly.
+	// Until it gets released, skip this test on Windows.
+	if runtime.GOOS == "windows" {
+		t.Skip("currently unsupported on Windows")
+	}
+
+	filepath := run.path("to be synced")
+	fh, err := osCreate(filepath)
+	require.NoError(t, err)
+
+	testData := []byte("0123456789")
+	appendData := []byte("10")
+
+	_, err = fh.Write(testData)
+	require.NoError(t, err)
+
+	err = fh.Close()
+	require.NoError(t, err)
+
+	fh, err = osAppend(filepath)
+	require.NoError(t, err)
+
+	_, err = fh.Write(appendData)
+	require.NoError(t, err)
+
+	err = fh.Close()
+	require.NoError(t, err)
+
+	info, err := os.Stat(filepath)
+	require.NoError(t, err)
+	require.EqualValues(t, len(testData)+len(appendData), info.Size())
 
 	run.waitForWriters()
 	run.rm(t, "to be synced")
--- a/cmd/cmount/fs.go
+++ b/cmd/cmount/fs.go
@@ -366,7 +366,12 @@ func (fsys *FS) Write(path string, buff []byte, ofst int64, fh uint64) (n int) {
 	if errc != 0 {
 		return errc
 	}
-	n, err := handle.WriteAt(buff, ofst)
+	var err error
+	if fsys.VFS.Opt.CacheMode < vfs.CacheModeWrites || handle.Node().Mode()&os.ModeAppend == 0 {
+		n, err = handle.WriteAt(buff, ofst)
+	} else {
+		n, err = handle.Write(buff)
+	}
 	if err != nil {
 		return translateError(err)
 	}
--- a/cmd/mount/handle.go
+++ b/cmd/mount/handle.go
@@ -5,6 +5,7 @@ package mount
 import (
 	"context"
 	"io"
+	"os"
 
 	"bazil.org/fuse"
 	fusefs "bazil.org/fuse/fs"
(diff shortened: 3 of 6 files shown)


ncw:
This looks useful thanks :-)

Looks like the tests are failing on Windows

```
--- FAIL: TestMount (53.50s)
    --- FAIL: TestMount/CacheMode=writes (19.45s)
        --- FAIL: TestMount/CacheMode=writes/TestWriteFileAppend (1.06s)
            write.go:167: 
                	Error Trace:	write.go:167
                	            	write E:to be synced: Incorrect function.
                	Test:       	TestMount/CacheMode=writes/TestWriteFileAppend
FAIL
FAIL	github.com/rclone/rclone/cmd/cmount	54.061s
```

bdutro:
> Looks like the tests are failing on Windows

Yeah, I haven't figured out yet whether it's a problem between Windows and the way the test is written or if there's an underlying problem with how O_APPEND works on Windows.

cmd/mountlib/mounttest/write.go
```diff
@@ -130,3 +130,45 @@ func TestWriteFileDup(t *testing.T) {
...
+
+	run.waitForWriters()
+	err = run.vfs.CleanUp()
+	require.NoError(t, err)
+	run.vfs.FlushDirCache()
+
+	fh, err = osAppend(filepath)
+	require.NoError(t, err)
+
+	_, err = fh.Seek(int64(len(testData)) + 1, 0)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	_, err = fh.Seek(int64(len(testData))+1, 0)
```

bdutro:
@ncw, in my local testing on Windows, I'm seeing that the flags for the second Open() call with O_APPEND ends up getting turned into O_RDONLY somewhere between the call to os.Open() and the cmount fs Open() handler. My guess is that something weird is happening somewhere between Windows and WinFSP that's returning an invalid handle.

cmd/mountlib/mounttest/write.go
```diff
@@ -130,3 +130,42 @@ func TestWriteFileDup(t *testing.T) {
...
+
+	_, err = fh.Write(appendData)
+	require.NoError(t, err)
+
+	err = fh.Close()
+	require.NoError(t, err)
+
+	info, err := os.Stat(filepath)
+	require.NoError(t, err)
+	require.EqualValues(t, len(testData) + len(appendData), info.Size())
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	require.EqualValues(t, len(testData)+len(appendData), info.Size())
```

bdutro:
I opened a pull-request on WinFSP with a fix: https://github.com/billziss-gh/winfsp/pull/261

ncw:
Excellent sleuthing!

Wouldn't the patch [I posted above ](https://github.com/rclone/rclone/pull/3628#issuecomment-543217529) fix this too as it adds `FILE_WRITE_DATA` - or have I mis-understood something?

bdutro:
It would, but since this is really a problem between the Go libraries and WinFSP I figure it's better to have a solution that works for everyone instead of an rclone-only band-aid.

bdutro:
Also, your patch would make it possible to write anywhere in the file instead of just appending to the end, which goes against the intent of the O_APPEND flag.

billziss-gh:
I just saw @bdutro's PR on WinFsp and followed the link from there.

I am under the impression that if a caller specifies **ONLY** `FILE_APPEND_DATA` when opening a file (not `FILE_WRITE_DATA | FILE_APPEND_DATA`) then the OS will always send an "append-only" offset to the file system (i.e. `-1`).

There is some evidence for this in the documentation. See this [link](https://docs.microsoft.com/en-us/windows/win32/fileio/file-access-rights-constants).

> **FILE_APPEND_DATA**
> For a file object, the right to append data to the file. (For local files, write operations will not overwrite existing data if this flag is specified without FILE_WRITE_DATA.) For a directory object, the right to create a subdirectory (FILE_ADD_SUBDIRECTORY).

ncw:
> I just saw @bdutro's PR on WinFsp and followed the link from there.
> 
> I am under the impression that if a caller specifies **ONLY** `FILE_APPEND_DATA` when opening a file (not `FILE_WRITE_DATA | FILE_APPEND_DATA`) then the OS will always send an "append-only" offset to the file system (i.e. `-1`).
> 
> There is some evidence for this in the documentation. See this [link](https://docs.microsoft.com/en-us/windows/win32/fileio/file-access-rights-constants).
> 
> > **FILE_APPEND_DATA**
> > For a file object, the right to append data to the file. (For local files, write operations will not overwrite existing data if this flag is specified without FILE_WRITE_DATA.) For a directory object, the right to create a subdirectory (FILE_ADD_SUBDIRECTORY).

Thanks for that Bill.

So I think the correct solution here is to patch this

https://github.com/rclone/rclone/blob/22bf8589cdaaea6a2fa15a607f4fc2f330e16157/lib/file/file_windows.go#L43-L46

To add `FILE_WRITE_DATA`.

Though will that start with the file offset at the end of the file?

billziss-gh:
@ncw 

> Though will that start with the file offset at the end of the file?

I believe so.

----

Latest beta of WinFsp includes @bdutro's `FILE_APPEND_DATA` fix.

https://github.com/billziss-gh/winfsp/releases/tag/v1.5B3

bdutro:
> @ncw
> 
> > Though will that start with the file offset at the end of the file?
> 
> I believe so.

That might be a problem (from the [docs](https://golang.org/pkg/os/#File.Seek)):

>The behavior of Seek on a file opened with O_APPEND is not specified.

So if the file pointer doesn't start at the end it might not be possible to actually append to the file.

ncw:
I guess the solution to this is to accept the behaviour is different on Windows and adjust the tests accordingly.  What do you think @bdutro ?

bdutro:
I could just skip the test on Windows until rclone targets the next release of WinFsp.

ncw:
> I could just skip the test on Windows until rclone targets the next release of WinFsp.

Let's do that so we can get this merged.

I won't merge it until after 1.50 is released which should be at the weekend.

bdutro:
One of the modules_race tests timed out. Is there any chance that's unrelated to my change?

ncw:
> One of the modules_race tests timed out. Is there any chance that's unrelated to my change?

There is every chance!  I've got that test on my list of sometimes failing tests.

I think this all looks good now - I'll merge it - thank you :-)

PR #3648, opened 2019-10-21
dedupe: make largest directory primary

### LINKED ISSUE #2568: rclone dedupe, choose the main folder efficiently

sakshamkhanna:
Fixes  #2568

#### Checklist

### DIFF
--- a/fs/operations/dedupe.go
+++ b/fs/operations/dedupe.go
@@ -247,20 +247,82 @@ func (x *DeduplicateMode) Type() string {
 	return "string"
 }
 
+// Directory with entry count and links to parents
+type dedupeDir struct {
+	dir    fs.Directory
+	parent string
+	count  int
+}
+
+// Map of directories by ID with recursive counts
+type dedupeDirsMap map[string]*dedupeDir
+
+func (dm dedupeDirsMap) get(id string) *dedupeDir {
+	d := dm[id]
+	if d == nil {
+		d = &dedupeDir{}
+		dm[id] = d
+	}
+	return d
+}
+
+func (dm dedupeDirsMap) increment(parent string) {
+	if parent != "" {
+		d := dm.get(parent)
+		d.count++
+		dm.increment(d.parent)
+	}
+}
+
 // dedupeFindDuplicateDirs scans f for duplicate directories
-func dedupeFindDuplicateDirs(ctx context.Context, f fs.Fs) ([][]fs.Directory, error) {
+func dedupeFindDuplicateDirs(ctx context.Context, f fs.Fs) (duplicateDirs [][]*dedupeDir, err error) {
+	dirsByID := dedupeDirsMap{}
+	dirs := map[string][]*dedupeDir{}
+
 	ci := fs.GetConfig(ctx)
-	dirs := map[string][]fs.Directory{}
-	err := walk.ListR(ctx, f, "", true, ci.MaxDepth, walk.ListDirs, func(entries fs.DirEntries) error {
-		entries.ForDir(func(d fs.Directory) {
-			dirs[d.Remote()] = append(dirs[d.Remote()], d)
-		})
+	err = walk.ListR(ctx, f, "", true, ci.MaxDepth, walk.ListAll, func(entries fs.DirEntries) error {
+		for _, entry := range entries {
+			remote := entry.Remote()
+			parentRemote := path.Dir(remote)
+			if parentRemote == "." {
+				parentRemote = ""
+			}
+
+			// Obtain ID of the object parent, if known.
+			// (This usually means that backend allows duplicate paths)
+			// Fall back to remote parent path, if unavailable.
+			var parent string
+			if entryParentIDer, ok := entry.(fs.ParentIDer); ok {
+				parent = entryParentIDer.ParentID()
+			}
+			if parent == "" {
+				parent = parentRemote
+			}
+
+			var ID string
+			if entryIDer, ok := entry.(fs.IDer); ok {
+				ID = entryIDer.ID()
+			}
+			if ID == "" {
+				ID = remote
+			}
+
+			if fsDir, ok := entry.(fs.Directory); ok {
+				d := dirsByID.get(ID)
+				d.dir = fsDir
+				d.parent = parent
+				dirs[remote] = append(dirs[remote], d)
+			}
+
+			dirsByID.increment(parent)
+		}
 		return nil
 	})
 	if err != nil {
 		return nil, errors.Wrap(err, "find duplicate dirs")
 	}
-	// make sure parents are before children
+
+	// Make sure parents are before children
 	duplicateNames := []string{}
 	for name, ds := range dirs {
 		if len(ds) > 1 {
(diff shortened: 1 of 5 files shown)


sakshamkhanna:
Can someone please review the code before I add unit tests

sakshamkhanna:
I accidently force pushed to the fork. I have reverted back, can someone please re-open the pull request.

fs/operations/dedupe.go
```diff
@@ -245,6 +250,70 @@ func dedupeMergeDuplicateDirs(ctx context.Context, f fs.Fs, duplicateDirs [][]fs
...
+		if dirSizes[i] > largestDirSize {
+			largestDirSize = dirSizes[i]
+			largestDirIndex = i
+		}
+	}
+	return
+}
+
+// directorySize returns the size of directory. includes size of subDirectories if recursive == true
+func directorySize(ctx context.Context, f fs.Fs, dir fs.Directory, recursive bool) (size int64, err error) {
```
ncw: I note that there is a function which does pretty much exactly this already.  It would need a dir parameter added.

https://github.com/rclone/rclone/blob/18c029e0f0c3a5dd0c925e285ce39071cb8728ff/fs/operations/operations.go#L1032

fs/operations/dedupe.go
```diff
@@ -245,6 +250,70 @@ func dedupeMergeDuplicateDirs(ctx context.Context, f fs.Fs, duplicateDirs [][]fs
...
+	var dirSize int64
+	entries, err := f.List(ctx, dir.Remote())
```
ncw: However I'm not sure this actually works...

when using `dir.Remote()` you'll get the name of the directory not the ID so rclone will list the same directory each time.

Same with the `Count` function.

ivandeex:
@sakshamkhanna any updates?

ivandeex:
@sakshamkhanna 
Please address review comments after a year.

ivandeex:
@sakshamkhanna
Rebased on master. Fixed build. Fixed linter warnings.

ncw:
I think this PR has a sound idea - making the largest directory primary but the implementation needs fixing

- calling the listing routines on duplicate directories won't work as expected
- listing the directories is expensive - we've already listed them so why not re-use that data

So I think using a simpler approach of finding the directory with the most entries (rather than the largest size) will be more efficient and this can be done by looking through data we already have collected here

https://github.com/rclone/rclone/blob/8a429d12cfb181bf3a28ce4e794b3eada8bf7e0e/fs/operations/dedupe.go#L236-L240

(this could count the items in every directory very easily and make a map of them by ID which gets returned)

ivandeex:
@sakshamkhanna has not been active on github for half a year

ivandeex:
Rebased on current master

ivandeex:
@ncw

>     * listing the directories is expensive - we've already listed them so why not re-use that data
> So I think using a simpler approach of finding the directory with the most entries (rather than the largest size) will be more efficient and this can be done by looking through data we already have collected here

You in fact ask to replace directory-only walk (with `walk.ListDirs` parameter) inside  `dedupeFindDuplicateDirs` by full directory+file walk with `walk.ListAll`. Being in doubt I checked whether some backends which implement `ListR` on the API level can use it to optimize the amount of API traffic. _I would expect that at least Google Drive with its SQL-like queries could support that._
After looking at the `walk.ListR` code (with or without backend ListR support) and especially at the backend [Features.ListR](https://github.com/rclone/rclone/blob/v1.54.0/fs/fs.go#L478) function, which does not imply any API-based filtering, I'm left to admit that your request is absolutely valid.

I'm OK with that, just don't like that after this PR the now all-object walk inside `dedupeFindDuplicateDirs` will be immediately followed by another all-object walk in `Duplicate` (https://github.com/rclone/rclone/blob/v1.54.0/fs/operations/dedupe.go#L354). I don't see a way to avoid that without buffering the complete all-object tree in memory which is wasteful.

fs/operations/dedupe.go
ivandeex: I think it's impossible to satisfy your `directorySize` request with existing `fs.Fs` interfaces. See below.

fs/operations/dedupe.go
ivandeex: I think it's impossible to implement `Count` and satisfy your requirements with existing `fs.Fs` interfaces.
See my comment below.

ivandeex:
@ncw 

I created the following directory structure on Google Drive as an experiment:
```
test/dir1
test/dir1/x.txt
test/dir1          (same name, different ID)
test/dir1/y.txt    (in the second one)
```

If I call `fs.Walk` with `ListAll` argument as you instructed, the entries are returned in the following order:
- test/dir1 (first ID)
- test/dir1 (second ID)
- test/dir1/x.txt
- test/dir1/y.txt

We have to find a parent for `x` and `y` to correctly increment directory counters. We cannot use `path.Dir` as both parent paths are equal. Neither generic `Fs.Object` nor concrete implementation in `backend/drive/drive.go` provide a way to get parent ID(s) of an object.
Alternatively we can separately list directories with duplicate names after the first walk one by one to count entries, but this requires extra API calls and still needs a new interface like `Fs.ListByID` which would use directory ID instead of remote path. Neither `List` nor `ListR` allow that now.

In this PR I am adding a new experimental `fs.Object` interface `ParentIDer` with single function `ParentID` that returns parent directory ID of the Object. I implement it in the `drive` and `opendrive` backends. This is possible as drive's `Object` internally keeps `drive.File` in its `baseObject`, which has `Parents` field with the list of parent IDs. You recently wrote somewhere that GDrive does not support multiple parents anymore, so I use just the first one.

I did not add unit test with duplicate directory names because rclone's `Mkdir` and `NewObject` APIs do not allow to fix parent directory ID. Instead I manually created directory structure described above and ran dedupe against it. Additionally I ran integration tests against drive remote.

Please review and comment.
Thanks.

fs/operations/dedupe.go
ivandeex: Solved it differently. See my comment below.

fs/operations/dedupe.go
ivandeex: Solved it differently. See my comment below.

ivandeex:
I'm not sure whether a secondary optimization in one of subcommands is worth expanding the core internal API. I just don't see any way around it. Leaving for @ncw to decide. If he sees no other utility in new APIs, this pull request can be just closed.

ivandeex:
waiting for review

ncw (approved):
I think this approach is a good one.

We already added 2 additional methods to make dedupe work so I think adding another isn't bad. It is also a very useful method which I'm sure people would use if we were to expose it in `rclone lsf` and `rclone lsjson`.

I think you've done a good job here and we should merge the code so I'm happy for you to do that when you are ready :-)

Issue #3437, opened 2019-08-13
Switch config in use with HTTP call.


AndyIbanez:
#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.48.0
- os/arch: darwin/amd64
- go version: go1.12.6
```

#### What problem are you are trying to solve?

I want to host rclone in a server I do not control (somewhere like DigitalOcean) because I want to build an app that consumes it's JSON API. It's the only way I can think of to write an iOS App that supports rclone in any capacity.

I do no want to start the `rcd` service with my unencrypted config file for security reasons. So having the option to run it with an empty rclone config file and then switch it to the actual config when needed would be ideal.

We talked about it [here](https://forum.rclone.org/t/rcd-with-encrypted-config-file/11222/7).

#### How do you think rclone should be changed to solve that?

`rcd` should expose a REST API that allows us to change the config file. Something like `POST config/file`. It should of course also support encrypted config files, so maybe a way to pass it both the config file name and an optional password would be ideal.

ncw:
I think this should be reasonably straight forward...

Do you fancy having a go?  If so I can talk you through it.

AndyIbanez:
Unfortunately I'm not familiar with Go yet to try using it in an open source project. Sorry 😔

terorie:
@AndyIbanez I implemented a `config/upload` call with a _nosave_ option in #3657. Could you try test it?

ivandeex:
@AndyIbanez 
If we implement _config profiles_ (https://github.com/rclone/rclone/issues/2697) and add an _rc_ command to switch a profile, would it suite your case ??

Bremonterro1:
What is going on????

ivandeex:
Such API call might open a security hole unless used with strong password or mTLS.

ivandeex:
Depends on #2697

PR #3657, opened 2019-10-23
rc: add config/upload

terorie:
Adds an `config/upload` call to rc.
The call has two parameters: 
 - _content_: Content of the config file
 - _nosave_: Don't write updated config to file (optional)

Caveat: _nosave_ is not effective on subsequent calls to `config/update,delete,…` which could cause secret information to "leak" to disk.

Solves #3437

### DIFF
--- a/fs/config/rc_test.go
+++ b/fs/config/rc_test.go
@@ -116,6 +116,47 @@ func TestRc(t *testing.T) {
 		assert.Equal(t, "cabbage", obscure.MustReveal(FileGet(testName, "test_key2")))
 	})
 
+	t.Run("Upload", func(t *testing.T) {
+		// Call upload with implicit nosave == false
+		call := rc.Calls.Get("config/upload")
+		assert.NotNil(t, call)
+		in := rc.Params{
+			"content": "[" + testName + "]\n" +
+				"type = local\n" +
+				"test_key = rutabaga\n" +
+				"test_key2 = cabbage\n",
+		}
+		out, err := call.Fn(context.Background(), in)
+		require.NoError(t, err)
+		assert.Nil(t, out)
+
+		assert.Equal(t, "local", FileGet(testName, "type"))
+		assert.Equal(t, "rutabaga", FileGet(testName, "test_key"))
+		assert.Equal(t, "cabbage", FileGet(testName, "test_key2"))
+
+		// Call upload with nosave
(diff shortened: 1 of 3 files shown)


fs/config/rc_test.go
```diff
@@ -116,6 +116,47 @@ func TestRc(t *testing.T) {
...
+			"content": "[" + testName + "]\n" +
+				"type = local\n" +
+				"test_key = rutabaga\n" +
+				"test_key2 = cabbage\n",
+		}
+		out, err := call.Fn(context.Background(), in)
+		require.NoError(t, err)
+		assert.Nil(t, out)
+
+		assert.Equal(t, "local", FileGet(testName, "type"))
```
terorie: @ncw I had to remove `MustReveal` from the asserts which is present in other lines. Is this a problem?

terorie:
After reading through the forum, I think it makes more to change the method signature to `config/file`, exposing there parameters:
 - _path_: Path to config file
 - _password_: Password of config file

This would solve the _nosave_ problem, too.

fs/config/rc_test.go
ncw: No that is fine I think - passwords don't get revealed until they are used - they are kept encrypted in the config file

ncw:
Apologies for the delay in reviewing this - I wanted to get 1.50 out the door!

I think what you've done looks very good.

> After reading through the forum, I think it makes more to change the method signature to `config/file`, exposing there parameters:
> 
> * _path_: Path to config file
> * _password_: Password of config file

So is the idea rclone would load and use the config file at `path` so you wouldn't send the contents?

I think you are right, this meets the original requirements better.

I guess rclone already has the ability to add/update/remove config file entries so replacing the config file in its entirity like the code does at the moment isn't strictly necesary, though I'm sure some users would find it convenient.

I guess if we wanted to support both use cases we could have
- path - path to config file - if not set then use existing
- password - new config password, if not set then use existing, if "" then unset
- contents - new contents of config file, if not set then use existing

> This would solve the _nosave_ problem, too.

arnwas:
beep

terorie:
Sorry, I didn't have time to complete this. If you want this feature @arnwas feel free to do so yourself. You can use this as a basis: https://patch-diff.githubusercontent.com/raw/rclone/rclone/pull/3657.patch

PR #3669, opened 2019-10-26
config: do not open browser on headless on google

### LINKED ISSUE #3323: drive: config flow should not open browser when non auto config

Running the headless config for Google Drive launches the default browser automatically.

Reproduced with rclone 1.45 on Ubuntu 18 and on Mac OS. See https://forum.rclone.org/t/option-to-disable-opening-browser/10710/2 for details.

Yexiaoxing:
On google fs (drive, google photos, and google cloud storage), if
headless is selected, do not open browser.

This also supplies a new option "auth-no-open-browser" for authorize
if the user does not want it.

This should fix #3323.

No browser popup if headless and using drive

#3323

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -386,6 +386,8 @@ func doConfig(id, name string, m configmap.Mapper, oauthConfig *oauth2.Config, o
 	oauthConfig, changed := overrideCredentials(name, m, oauthConfig)
 	authorizeOnlyValue, ok := m.Get(config.ConfigAuthorize)
 	authorizeOnly := ok && authorizeOnlyValue != "" // set if being run by "rclone authorize"
+	authorizeNoAutoBrowserValue, ok := m.Get(config.ConfigAuthNoBrowser)
+	authorizeNoAutoBrowser := ok && authorizeNoAutoBrowserValue != ""
 
 	// See if already have a token
 	tokenString, ok := m.Get("token")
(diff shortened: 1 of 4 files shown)


lib/oauthutil/oauthutil.go
```diff
@@ -470,9 +472,13 @@ func doConfig(id, name string, m configmap.Mapper, oauthConfig *oauth2.Config, o
 		authURL = "http://" + bindAddress + "/auth?state=" + state
 	}
 
-	// Open the URL for the user to visit
-	_ = open.Start(authURL)
-	fmt.Printf("If your browser doesn't open automatically go to the following link: %s\n", authURL)
+	if !authorizeNoAutoBrowser && oauthConfig.RedirectURL != TitleBarRedirectURL {
```
ncw: Why does this have `&& oauthConfig.RedirectURL != TitleBarRedirectURL` ?

`TitleBarRedirectURL` still needs the browser window open doesn't it?

Yexiaoxing: Well this is actually a complicated logic :)

There are two scenarios that will open the browser,
- `rclone authorize` with no option given
- `rclone config`, and use auto config

In the second scenarios, if the `oauthConfig.RedirectURL` was `TitleBarRedirectURL`, and user chooses to use auto config, it will be changed to the internal web server instead of `TitleBarRedirectURL`. So if in this line `oauthConfig.RedirectURL == TitleBarRedirectURL`, it means the user has choosen not to use auto config, and not open the browser automatically.

PR #3684, opened 2019-10-29
Adding group parameter in status return

negative0:
The group parameter gives a id to uniquely identify a job and allow operations to be performed on it.
This will allow the job to be stopped using the rclone web ui.


ncw:
That looks great thank you - will merge now :-)

negative0:
Thanks. :-)

Issue #2153, opened 2018-03-18
Upload .deb and .rpm to packagecloud


ncw:
Upload the built .rpm and .deb to packagecloud.io

lskillen:
@ncw <3 rclone - Please feel free to consider [Cloudsmith](https://cloudsmith.io) as well for packaging. :-)

It's also commercial but has very generous free plans and offerings for open-source projects.

*Full disclosure*: I work for Cloudsmith, so might be biased (but only _slightly_, honest). We use rclone too!

lskillen:
@ncw Still interested? We added the support for "agnostic" distributions, so that you can upload packages that work for "any distribution", "any version" and "any architecture". You can also divide this into the stable and beta streams.

AzureRaptor:
@ncw I hope you can take a look at this.  Having to install (and "maintain") things via a 'curl'ed shell script is _reeeeallly_ kinda 1999.  ;}

azlux:
My personal repo is already set for that (apt only). Step to test it : 
```
echo "deb http://packages.azlux.fr/debian/ buster main" | sudo tee /etc/apt/sources.list.d/azlux.list
wget -qO - https://azlux.fr/repo.gpg.key | sudo apt-key add -
sudo apt update
sudo apt install rclone
```

A discussion is already in progress on the forum for rlcone to have is own repository and/or put unofficial repo into the docs : https://forum.rclone.org/t/apt-repository-for-rclone/13725/7

@AzureRaptor I don't like manual script to update soft on all my machine. Too touchy.

AzureRaptor:
@azlux Thank you kindly for this information!

I agree completely about manual scripts and maintenance of any kind.  I strongly prefer to use the distro's preferred tools for updating software and system maintenance, than a hodgepodge of sloppy methods I constantly have to tweak, babysit or juryrig.

AzureRaptor:
@azlux  Firstly I want to just thank you again for providing an apt repo for everyone! 

But, something seems to be up with https://packages.azlux.fr/  ... when I go there I see:
(aptitude also yields a similar error)

Secure Connection Failed

An error occurred during a connection to azlux.fr. Peer reports incompatible or unsupported protocol version. Error code: SSL_ERROR_PROTOCOL_VERSION_ALERT

    The page you are trying to view cannot be shown because the authenticity of the received data could not be verified.
    Please contact the website owners to inform them of this problem.


Maybe there's some problem going on with your hosting provider?

[...]

PR #3695, opened 2019-11-02
install.sh: Install to /usr/local/bin

nyuszika7h:

This PR changes the install script to use `/usr/local/bin` instead of `/usr/bin`. You should never install to `/usr/bin` directly, it should be reserved for the package manager. Installing to `/usr/bin` will also mess up a possible existing installation in there.

No, but this is a pretty standard thing.

#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -146,6 +146,10 @@ cd $unzip_dir/*
 
 case $OS in
   'linux')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+        echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
+    fi
     #binary
     cp rclone /usr/local/bin/rclone.new
     chmod 755 /usr/local/bin/rclone.new
@@ -161,6 +165,10 @@ case $OS in
     fi
     ;;
   'freebsd'|'openbsd'|'netbsd')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+        echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
+    fi
     #bin
     cp rclone /usr/local/bin/rclone.new
     chown root:wheel /usr/local/bin/rclone.new
@@ -171,6 +179,10 @@ case $OS in
     makewhatis
     ;;
   'osx')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+        echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
+    fi
     #binary
     mkdir -p /usr/local/bin
     cp rclone /usr/local/bin/rclone.new
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -147,7 +147,7 @@ cd $unzip_dir/*
 case $OS in
   'linux')
     rclonepath="$(command -v rclone)"
-    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin/rclone' ]; then
         echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
     fi
     #binary
@@ -166,7 +166,7 @@ case $OS in
     ;;
   'freebsd'|'openbsd'|'netbsd')
     rclonepath="$(command -v rclone)"
-    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin/rclone' ]; then
         echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
     fi
     #bin
@@ -180,7 +180,7 @@ case $OS in
     ;;
   'osx')
     rclonepath="$(command -v rclone)"
-    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin/rclone' ]; then
         echo "WARNING: Existing rclone installation found in PATH: $rclonepath"
     fi
     #binary
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -147,10 +147,10 @@ cd $unzip_dir/*
 case $OS in
   'linux')
     #binary
-    cp rclone /usr/bin/rclone.new
-    chmod 755 /usr/bin/rclone.new
-    chown root:root /usr/bin/rclone.new
-    mv /usr/bin/rclone.new /usr/bin/rclone
+    cp rclone /usr/local/bin/rclone.new
+    chmod 755 /usr/local/bin/rclone.new
+    chown root:root /usr/local/bin/rclone.new
+    mv /usr/local/bin/rclone.new /usr/local/bin/rclone
     #manuals
     if ! [ -x "$(command -v mandb)" ]; then
         echo 'mandb not found. The rclone man docs will not be installed.'
@@ -162,9 +162,9 @@ case $OS in
     ;;
   'freebsd'|'openbsd'|'netbsd')
     #bin
-    cp rclone /usr/bin/rclone.new
-    chown root:wheel /usr/bin/rclone.new
-    mv /usr/bin/rclone.new /usr/bin/rclone
+    cp rclone /usr/local/bin/rclone.new
+    chown root:wheel /usr/local/bin/rclone.new
+    mv /usr/local/bin/rclone.new /usr/local/bin/rclone
     #man
     mkdir -p /usr/local/man/man1
     cp rclone.1 /usr/local/man/man1/


ncw:
This has the potential to cause a backwards compatibility problem if a user has rclone installed in `/usr/bin/rclone` installed by a previous version of this script.  When they install it again they will install rclone in `/usr/local/bin` and PATH only knows which one will be chosen to run.  I think the one in `/usr/local/bin` according to my PATH.

Is `/usr/local/bin` is on the PATH by default in all the OSes?

Thoughts?

nyuszika7h:
Yes, as far as I know, both Linux and BSD have it in PATH by default. Usually (if not always) it comes before `/usr/bin` indeed, so the "local" installation would take precedence.

Yexiaoxing:
I would go for this PR - most distros prefer `/usr/local/bin`.

BTW it is better to display a warning or some notice if the user has previously installed in `/usr/bin`?

ncw:
A warning if there is an /usr/bin/rclone sounds like a great idea.

Fancy adding that to the PR @nyuszika7h ?

nyuszika7h:
I added a warning if it's found anywhere in PATH other than `/usr/local/bin`.

docs/content/install.sh
```diff
@@ -146,6 +146,10 @@ cd $unzip_dir/*
 
 case $OS in
   'linux')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
```
darthShadow: Correct me if I am wrong, but shouldn't the check be to `/usr/local/bin/rclone` instead of just `/usr/local/bin` ?

ncw: I think you are correct...

docs/content/install.sh
```diff
@@ -161,16 +165,24 @@ case $OS in
     fi
     ;;
   'freebsd'|'openbsd'|'netbsd')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
```
Cnly:
```suggestion
    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin/rclone' ]; then
```

Little typo here?

docs/content/install.sh
```diff
@@ -161,16 +165,24 @@ case $OS in
...
+    chown root:wheel /usr/local/bin/rclone.new
+    mv /usr/local/bin/rclone.new /usr/local/bin/rclone
     #man
     mkdir -p /usr/local/man/man1
     cp rclone.1 /usr/local/man/man1/
     makewhatis
     ;;
   'osx')
+    rclonepath="$(command -v rclone)"
+    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin' ]; then
```
Cnly:
```suggestion
    if [ -x "$rclonepath" ] && ! [ "$rclonepath" = '/usr/local/bin/rclone' ]; then
```

ncw:
@nyuszika7h if you can fix up the issue pointed out by @Cnly then I'll merge this - thanks :-)

Dibbyo456:
This will break php. In order to execute binary via php’s ‘exec()‘ without using an absolute path, the binary needs to be located in ‘/usr/bin’.
Php won’t find the binary if it’s located on ‘/usr/local/bin’

Dibbyo456:
This will break Cronjobs for many users (including me). If someone uses Rclone via Cronjob like this:
```
* * * * * rclone copy remote: dest:
```
It would stop working. Because in Ubuntu the default `shell` for crontab is `/bin/sh` and `$PATH` is `/usr/bin:/bin`. As you can see `/usr/local/bin` is not included the `$PATH` variable.

Either the users now needs to use full path of rclone or they needs to add `/usr/local/bin/` to their `$PATH` before running the command.

Now, the question is do we really need this? I understand that `/usr/local/bin` is the standard location of Rclone but doing this will create chaos.

nyuszika7h:
Sorry, I forgot about this.

@Dibbyo456 Seems like you're right, cron hardcodes PATH to `/usr/bin:/bin`. I would expect it is common practice to always specify the full path to executables in the crontab anyway. And you should generally test your software to see if it still works after updates. Maybe some sort of warning could be added, but IMO that's a poor excuse to keep the existing behavior which can conflict with package managers and cause confusion. `/usr/local/bin` has been a standard location forever, despite cron not including it in its hardcoded PATH.

ivandeex:
@ncw I suggest rejecting this for compatibility with existing rclone installations and OSes

nyuszika7h:
As I said before, IMO the counter-arguments are weak. Surely you're not using rclone in a critical production system, or if you are then you're able to test it in staging before and make a trivial one-line change to fix it?

ivandeex:
> [snip] in a critical production system, or if you are then you're able to test it in staging

Testing changes in _critical production_ systems is expensive, including changes deemed trivial. Cosmetic production changes cost too. Business folks usually require proofs to start spending.

> What I'm suggesting is completely standards compliant

Please provide links to POSIX, Unix SysV or similar authoritative document that proves your claims under a section marked as [SHOULD] or stronger.

@ncw It doesn't really matter whatever resolution you choose. rclone has more critical stuff on schedule.

I think this PR is of as low business value as Lilliput vs Blefuscu dispute regarding egg ends, pure bike-shedding example.

ivandeex:
@ncw I can't find if this `install.sh` used by rclone's Makefiles or obeyed by the homebrew formula? If yes, it's a `build` change. If not, it's a `doc fix`.

ivandeex:
Looking again at rclone install location in documentation, adopted ansible role, mentioned here `install.sh`, and `.deb` / `.rpm` produced by our release jobs.
Related to https://github.com/rclone/rclone/issues/4809 (adopting ansible role into project)
Related to https://github.com/rclone/rclone/issues/2153 (push rclone native deb/rpm on package cloud)

ivandeex:
https://github.com/rclone/rclone/pull/3695#issuecomment-591588788 is right, `cron` does not by default hardcode `/usr/local/bin` (http://manpages.ubuntu.com/manpages/xenial/en/man5/crontab.5.html)
I think it would be helpful to add a cron related warning in the docs.

ivandeex:
Looking into more places for info:

- adopted ansible role uses `/usr/local/bin` https://github.com/rclone/rclone/issues/4809
- ubuntu implementation of cron ignores `local` http://manpages.ubuntu.com/manpages/xenial/en/man5/crontab.5.html
- systemd used to require absolute paths but started accepting executables without path since recently, and they include `local` in search https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines (see also https://www.freedesktop.org/software/systemd/man/systemd.exec.html#%24PATH)
- OpenGroup (and their POSIX related Unix specification) leaves this topic up to distributions: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html#tag_08_03 (table of contents https://pubs.opengroup.org/onlinepubs/009695399/nframe.html)
- POSIX ignores the topic of default PATH https://unix.stackexchange.com/questions/538844/does-posix-guarantee-the-paths-to-any-standard-utilities#538858
- bash by default includes `local` but at least `command -p` omits it https://github.com/bminor/bash/blob/master/config-top.h
- wikipedia is not concrete on this https://en.m.wikipedia.org/wiki/Unix_filesystem#Conventional_directory_layout
- `restic` runs rclone from whatever system path https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#other-services-via-rclone
- assorted non-authoritative stuff and links from askubuntu https://askubuntu.com/questions/386629/what-are-the-default-path-values#386636 and stackoverflow https://unix.stackexchange.com/questions/4186/what-is-usr-local-bin
- BSD packaging uses `/usr/local` widely (old rant) https://www.aplawrence.com/Opinion/religion.html

phew

ivandeex:
In case this _install.sh_ is used by build system, let's add a changelog line pretending we let users know in advance

ncw:
I've read over the thread again and there are good arguments for and against the change.

Thank you for you work on exploring the options @ivandeex 

I think because we can't come to a consensus, and that this is a breaking change then I'm going to reject this pull request as it is.

It always breaks my heart rejecting pull requests that people have put time and energy in, but I wouldn't be doing my job properly unless I look out for backwards compatibility.

I would accept a PR that made the install location an option.

However I think a PR which installed the relevant .deb or .rpm if possible would be even better.

nyuszika7h:
If rclone really is used on critical production systems, then nobody is pressuring them to update immediately. They generally don't rush to update until there's an important security fix anyway. I understand that even small unexpected changes might cause them to waste more time, but such changes can very well happen in a new stable release of a Linux distribution as well, and it shouldn't be anything that requires hours to work around. If this was to be included an important security patch or LTS version that's releasing tomorrow or something I would have slightly more sympathy. People had to move on from Windows XP too, and this is a *much* smaller change in the grand scheme of things.

I wouldn't call it bikeshedding since this is definitely an FHS standard (https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html) despite some software like cron not having an entirely sane default. Maybe it's just me, but I've known for years to always use the absolute path or `/usr/bin/env` in cron because relative paths don't always work. Realistically I don't think this is going to cause breakage in important production systems, the worst that will happen is that average Joe is confused why his rclone mount is broken, but if they didn't even notice the change, then the more likely case is that the old version stays behind in `/usr/bin`, and properly coded programs will prefer the binary in `/usr/local/bin`, while others will continue to use the outdated one in `/usr/bin`.

You may save some people the trouble of making a one-line change, but IMO the current script will continue to cause headaches for people who rightfully expect it to install to `/usr/local/bin` and then manage to overwrite the package-manager-installed version which may then cause it to throw errors, or the locally installed version to be uninstalled when they try to get rid of the old version from the package manager.

&lt;/rant&gt;

Oh well, if even the maintainer of the project thinks it's not worth making the change, then I will at least try to submit a PR that allows customizing the installation path if someone else doesn't beat me to it, since right now it appears even the Makefile doesn't allow installing to /usr/local/bin. The most you can do without editing it is use `DESTDIR` which just installs to `$(DESTDIR)/usr/bin`.

ivandeex:
Related to #4857

PR #3701, opened 2019-11-07
sftp: Retry Creation of Connection - fixes #3656

### LINKED ISSUE #3656: SFTP Error doesn't seem to be retried

We are pulling data from an apparently rather unreliable data source, hence we enabled retrying with sleep in between. That also works for most errors, but one error leads to rclone immediately exit and seemingly not retrying:

`2019/10/23 10:45:37 Failed to create file system for "foo:": NewFs: couldn't initialise SFTP: EOF`

rclone v1.49.5
- os/arch: linux/amd64
- go version: go1.13.1

linux/amd64

SFTP

`rclone copy foo: s3:bar`

```
2019/10/23 12:13:21 DEBUG : rclone: Version "v1.49.5" starting with parameters ["/usr/bin/rclone" "copy" "foo:" "s3:bar" "--config" "/root/rclone/config" "--filter" "+ /bar/**" "--filter" "+ /baz/**" "--filter" "- *" "--retries-sleep" "10s" "--backup-dir" "s3:foo/backup" "--max-delete" "0" "--size-only" "-vv"]
2019/10/23 12:13:21 DEBUG : Using config file from "/root/rclone/config"
2019/10/23 12:13:27 DEBUG : sftp://xxx@yyy:22/: New connection 100.96.2.187:51900->176.9.117.162:22 to "SSH-2.0-OpenSSH_6.7p1 Debian-5"
2019/10/23 12:13:31 Failed to create file system for "foo:": NewFs: couldn't initialise SFTP: EOF
```

sbrandtb:

Increase stability of SFTP by retrying connection creation.

#3656 

Tried it out in real life - it fixes my problem.

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -29,15 +29,17 @@ import (
 	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/lib/env"
+	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/readers"
 	sshagent "github.com/xanzy/ssh-agent"
 	"golang.org/x/crypto/ssh"
-	"golang.org/x/time/rate"
 )
 
 const (
-	connectionsPerSecond    = 10 // don't make more than this many ssh connections/s
 	hashCommandNotSupported = "none"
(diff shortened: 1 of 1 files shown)


sbrandtb:
@ncw Great idea since it reduces complexity and makes SFTP now use the usual configurable retrying. Changed.

ncw:

I'll merge it now.

PR #3711, opened 2019-11-08
jottacloud: use new auth method used by official client

### LINKED ISSUE #3645: Unable to create new Jottacloud remote (2FA)

Unable to create new Jottacloud remote, seems they might have changed to only accept OTP

1.49.5 and v1.49.5-236-g2e505430-beta

Manjaro, 64 bit

Jottacloud

`rclone config`

```
~/rclone/rclone-v1.49.5-236-g2e505430-beta-linux-amd64 >>> ./rclone -vv config
2019/10/21 10:42:57 DEBUG : rclone: Version "v1.49.5-236-g2e505430-beta" starting with parameters ["./rclone" "-vv" "config"]
2019/10/21 10:42:57 NOTICE: Config file "/home/steven/.config/rclone/rclone.conf" not found - using defaults
No remotes found - make a new one
n) New remote
s) Set configuration password
q) Quit config
n/s/q> n
name> Jotta
Type of storage to configure.
 1 / 1Fichier
   \ "fichier"
 2 / Alias for an existing remote
   \ "alias"
 3 / Amazon Drive
   \ "amazon cloud drive"
 4 / Amazon S3 Compliant Storage Provider (AWS, Alibaba, Ceph, Digital Ocean, Dreamhost, IBM COS, Minio, etc)
   \ "s3"
 5 / Backblaze B2
   \ "b2"
 6 / Box
   \ "box"
 7 / Cache a remote
   \ "cache"
 8 / Citrix Sharefile
   \ "sharefile"
 9 / Dropbox
   \ "dropbox"
10 / Encrypt/Decrypt a remote
   \ "crypt"
11 / FTP Connection
   \ "ftp"
12 / Google Cloud Storage (this is not Google Drive)
   \ "google cloud storage"
13 / Google Drive
   \ "drive"
14 / Google Photos
   \ "google photos"
15 / Hubic
   \ "hubic"
16 / JottaCloud
   \ "jottacloud"
17 / Koofr
   \ "koofr"
18 / Local Disk
   \ "local"
19 / Mail.ru Cloud
   \ "mailru"
20 / Mega
   \ "mega"
21 / Microsoft Azure Blob Storage
   \ "azureblob"
22 / Microsoft OneDrive
   \ "onedrive"
23 / OpenDrive
   \ "opendrive"
24 / Openstack Swift (Rackspace Cloud Files, Memset Memstore, OVH)
   \ "swift"
25 / Pcloud
   \ "pcloud"
26 / Put.io
   \ "putio"
27 / QingCloud Object Storage
   \ "qingstor"
28 / SSH/SFTP Connection
   \ "sftp"
29 / Transparently chunk/split large files
   \ "chunker"
30 / Union merges the contents of several remotes
   \ "union"
31 / Webdav
   \ "webdav"
32 / Yandex Disk
   \ "yandex"
33 / http Connection
   \ "http"
34 / premiumize.me
   \ "premiumizeme"
Storage> 16
** See help for jottacloud backend at: https://rclone.org/jottacloud/ **

Edit advanced config? (y/n)
y) Yes
n) No
y/n> n
Remote config

Do you want to create a machine specific API key?

Rclone has it's own Jottacloud API KEY which works fine as long as one only uses rclone on a single machine. When you want to use rclone with this account on more than one machine it's recommended to create a machine specific API key. These keys can NOT be shared between machines.

y) Yes
n) No
y/n> y
2019/10/21 10:43:13 DEBUG : Trying to register device 'rclone-NR0XkIAZEjvVRpu5FPUTE'
2019/10/21 10:43:13 DEBUG : Saving config "client_id" = "1oj1g1ei7g9rgm2gcllbjc7j9o" in section "Jotta" of the config file
2019/10/21 10:43:13 DEBUG : Saving config "client_secret" = "qGoIEwoneg4aNHDiDqxT1ASANHdI5QAoSjKoGV6P6D8bt16QSv3vENu0" in section "Jotta" of the config file
2019/10/21 10:43:13 DEBUG : Got clientID '1oj1g1ei7g9rgm2gcllbjc7j9o' and clientSecret 'gosevbnk80pro1l1q4qopvrs7s'
Username> s****@****.org
Your Jottacloud password is only required during setup and will not be stored.
password:
2019/10/21 10:43:26 Failed to get oauth token: HTTP error 400 (400 Bad Request) returned body: "{\"error\":\"invalid_client_or_otp_missing_or_wrong\",\"error_description\":\"Client authentication failed (e.g., unknown client, no client authentication included, unsupported authentication method, or the resource owner has enabled OTP and the otp code is not supplied in the password grant request or the supplied otp code is wrong).\"}"
```

buengese:
This kinda needs some user testing. It doesn't break any of my stuff but you can never be sure with Jottacloud. Authentication now works by using a one time token you can generate in the Security Tab of the Jottacloud webui  ([Link](https://www.jottacloud.com/web/secure)). Will update docs once I confirm this doesn't break anything.
Fixes #3645

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -4,12 +4,13 @@ import (
 	"bytes"
 	"context"
 	"crypto/md5"
+	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
@@ -25,7 +26,6 @@ import (
 	"github.com/rclone/rclone/fs/config"
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
-	"github.com/rclone/rclone/fs/config/obscure"
 	"github.com/rclone/rclone/fs/encodings"
 	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/fshttp"
@@ -41,29 +41,25 @@ const enc = encodings.JottaCloud
 
 // Globals
 const (
-	minSleep                    = 10 * time.Millisecond
-	maxSleep                    = 2 * time.Second
-	decayConstant               = 2 // bigger for slower decay, exponential
-	defaultDevice               = "Jotta"
-	defaultMountpoint           = "Archive"
-	rootURL                     = "https://www.jottacloud.com/jfs/"
-	apiURL                      = "https://api.jottacloud.com/"
-	baseURL                     = "https://www.jottacloud.com/"
-	tokenURL                    = "https://api.jottacloud.com/auth/v1/token"
-	registerURL                 = "https://api.jottacloud.com/auth/v1/register"
-	cachePrefix                 = "rclone-jcmd5-"
-	rcloneClientID              = "nibfk8biu12ju7hpqomr8b1e40"
-	rcloneEncryptedClientSecret = "Vp8eAv7eVElMnQwN-kgU9cbhgApNDaMqWdlDi5qFydlQoji4JBxrGMF2"
-	configClientID              = "client_id"
-	configClientSecret          = "client_secret"
-	configDevice                = "device"
-	configMountpoint            = "mountpoint"
-	charset                     = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	minSleep          = 10 * time.Millisecond
+	maxSleep          = 2 * time.Second
+	decayConstant     = 2 // bigger for slower decay, exponential
+	defaultDevice     = "Jotta"
+	defaultMountpoint = "Archive"
+	rootURL           = "https://www.jottacloud.com/jfs/"
+	apiURL            = "https://api.jottacloud.com/"
+	baseURL           = "https://www.jottacloud.com/"
+	tokenURL          = "https://id.jottacloud.com/auth/realms/jottacloud/protocol/openid-connect/token"
+	cachePrefix       = "rclone-jcmd5-"
+	configDevice      = "device"
+	configMountpoint  = "mountpoint"
+	configVersion     = 1
 )
 
 var (
 	// Description of how to auth for this app for a personal account
 	oauthConfig = &oauth2.Config{
+		ClientID: "jottacli",
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  tokenURL,
 			TokenURL: tokenURL,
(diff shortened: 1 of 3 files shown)


ncw:
Do you want to push this to a branch on the main repo so it makes a beta - then we can ask for testers in the forum?

buengese:
Yeah will do so after I fixup one minor thing. Kinda forgot that we don't generate artifacts for pr builds

buengese:
There is Beta available [here](https://beta.rclone.org/branch/v1.50.1-035-g42c3365a-jotta_newauth-beta/) now. Contrary to what I initially said this does break existing configs.
@ncw I don't even have an account in the rclone forum so maybe you could post this in the appropriate place?

ncw:
> There is Beta available [here](https://beta.rclone.org/branch/v1.50.1-035-g42c3365a-jotta_newauth-beta/) now. Contrary to what I initially said this does break existing configs.

:-(  I'd like to avoid that if we possibly can! It creates a lot of user problems on upgrade!  If it is inevitable then a definitive error message telling the user what to do will stop the confusion (`rclone config reconnect remote:`)

> @ncw I don't even have an account in the rclone forum so maybe you could post this in the appropriate place?

Here is is: https://forum.rclone.org/t/jottacloud-new-access-method-testers-needed-please/12887

buengese:
The problem is the token refresh. Both API endpoint and the request itself changed. However I could just leave the old code in select a different code path for older configs. Maybe we should introduce some kind of internal config version parameter for jottacloud this way we could leave the old code in for now and warn the users when they use an outdated config. After all this is the fourth time we are breaking the jottacloud config.

ncw:
> The problem is the token refresh. Both API endpoint and the request itself changed. However I could just leave the old code in select a different code path for older configs. Maybe we should introduce some kind of internal config version parameter for jottacloud this way we could leave the old code in for now and warn the users when they use an outdated config. After all this is the fourth time we are breaking the jottacloud config.

I don't mind removing the old code paths as long as we can signal to the user that they need to re-auth reliably.

We could do this with a config version parameter I suppose, but we'll need to be careful with the default value.

backend/jottacloud/jottacloud.go
```diff
@@ -504,6 +516,19 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		return nil, err
 	}
 
+	var ok bool = true
```
golangcibot: ineffectual assignment to `ok` (from `ineffassign`)

rchiechi:
Using `rclone-v1.50.1-035-g42c3365a-jotta_newauth-beta-linux-amd64`, I was able to create a new Jotta remote, but after a few minutes it starting giving:
```
Failed to create file system for "jotta:": couldn't get customer info: Get https://api.jottacloud.com/account/v1/customer: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"unauthorized_client","error_description":"INVALID_CREDENTIALS: Invalid client credentials"}
```
Deleting it and re-creating it produced the same result.

buengese:
The beta build available here was somewhat outdated. Could you please try [this one](https://beta.rclone.org/branch/v1.50.1-037-g46516e1a-jotta_newauth-beta/). @rchiechi

StevenB83:
Using `v1.50.1-037-g46516e1a-jotta_newauth-beta` I was able to create a jotta remote, but lost auth after a while.
re-created and after an hour it still worked, shortly after it didn't
```
~/Downloads >>> ./rclone-linux-amd64 ls jotta:testing                                                                                                                                        
        3 newfile
~/Downloads >>> ./rclone-linux-amd64 ls jotta:testing                                                                                                                                        
2019/11/19 05:40:35 Failed to create file system for "jotta:testing": couldn't get customer info: Get https://api.jottacloud.com/account/v1/customer: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"unauthorized_client","error_description":"INVALID_CREDENTIALS: Invalid client credentials"}
```

rchiechi:
> The beta build available here was somewhat outdated. Could you please try [this one](https://beta.rclone.org/branch/v1.50.1-037-g46516e1a-jotta_newauth-beta/). @rchiechi

There was no Linux build available, so I tried `rclone-v1.50.1-037-g46516e1a-jotta_newauth-beta-freebsd-amd64`, but ran into the same problem (works at first, then throws INVALID_CREDENTIALS).

ncw:
The linux build didn't build because of this

```
-- START CODE QUALITY REPORT -------------------------------
##[error]backend/jottacloud/jottacloud.go:519:6: ineffectual assignment to `ok` (ineffassign)
	var ok bool = true
	    ^
```

buengese:
Yeah hadn't pushed a fix for that yet. Looks like token refresh is broken again. I'll fix it later today.

backend/jottacloud/jottacloud.go
```diff
@@ -41,29 +41,27 @@ const enc = encodings.JottaCloud
...
+	defaultMountpoint = "Archive"
+	rootURL           = "https://www.jottacloud.com/jfs/"
+	apiURL            = "https://api.jottacloud.com/"
+	baseURL           = "https://www.jottacloud.com/"
+	tokenURL          = "https://id.jottacloud.com/auth/realms/jottacloud/protocol/openid-connect/token"
+	registerURL       = "https://api.jottacloud.com/auth/v1/register"
+	cachePrefix       = "rclone-jcmd5-"
+	configDevice      = "device"
+	configMountpoint  = "mountpoint"
+	charset           = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
golangcibot: `charset` is unused (from `varcheck`)

backend/jottacloud/jottacloud.go
```diff
@@ -41,29 +41,27 @@ const enc = encodings.JottaCloud
...
+	minSleep          = 10 * time.Millisecond
+	maxSleep          = 2 * time.Second
+	decayConstant     = 2 // bigger for slower decay, exponential
+	defaultDevice     = "Jotta"
+	defaultMountpoint = "Archive"
+	rootURL           = "https://www.jottacloud.com/jfs/"
+	apiURL            = "https://api.jottacloud.com/"
+	baseURL           = "https://www.jottacloud.com/"
+	tokenURL          = "https://id.jottacloud.com/auth/realms/jottacloud/protocol/openid-connect/token"
+	registerURL       = "https://api.jottacloud.com/auth/v1/register"
```
golangcibot: `registerURL` is unused (from `varcheck`)

buengese:
Ok let's give this another try. New beta [here](https://beta.rclone.org/branch/v1.50.1-036-g88b76af0-jotta_newauth-beta/) with linux build this time :)

rchiechi:
> Ok let's give this another try. New beta [here](https://beta.rclone.org/branch/v1.50.1-036-g88b76af0-jotta_newauth-beta/) with linux build this time :)

This one works! In fact, I didn't even have to recreate the Jotta remote. It worked using the config file generated from the earlier beta. I tested it both on Linux and FreeBSD.

buengese:
I think we can merge this now.

ncw:
Great!  I'll do that now - thank you :-)

ncw:
@buengese I just had a go with this...

I have a couple of comments...

I think we should put some extra text in the config as this is a bit bare!

```
Remote config
 
Login Token>
```

Perhaps something as simple as this?

```
Remote config

Generate a personal login token here: https://www.jottacloud.com/web/secure
Login Token>
```

The second thing I noticed was that the mountpoint didn't default to `Sync` like it used to so all my files disappeared.  I'm not sure what it did default to though.  I ran through the config and explicitly chose the Sync mountpoint rather than the default and my files re-appeared.

Thoughts?

buengese:
Yeah the extra text would have been better kinda forgot about that. I didn't touch the mountpoint selection code and defaultMountpoint always used to be Archive.  ( see https://github.com/rclone/rclone/commit/4195bd78801cd138a3c81f922cffbe852621b3b0#diff-0d824e6a65d3ed8f6b67a62f9daad6edL48)

ncw:
> Yeah the extra text would have been better kinda forgot about that.

I added the extra text here: 33c80bbb96aa7dc3b20393553931f094f0e04f04 

> I didn't touch the mountpoint selection code and defaultMountpoint always used to be Archive. ( see [4195bd7#diff-0d824e6a65d3ed8f6b67a62f9daad6edL48](https://github.com/rclone/rclone/commit/4195bd78801cd138a3c81f922cffbe852621b3b0#diff-0d824e6a65d3ed8f6b67a62f9daad6edL48))

Ah OK. I probably got confused as my jotta remote has been reconfigured quite a few times!

StevenB83:
Thank you so much, seems to work great now!
For my use rclone is far superior to the official client! :1st_place_medal: 
Adding the url so they know to go there and generate token to the config is a good idea imo.

PR #3718, opened 2019-11-11
fix error counter - fixes #3650

### LINKED ISSUE #3650: RClone Counts Errors twice for Copy operations
RClone counts errors twice for the copy operations

RClone 1.49.3

Ubuntu 18.04 LTS, 64 bit

Developing a new backend for RClone

copy

image

This happens because `fs.CountError(err)` is called once from within the `Copy()` function in `operations.go` and is then again called as part of `tr.Done(err)`. This results in the errors being counted twice. 

It looks like `fs.CountError` is again being called once at the end of all files, though I couldn't trace from where it is called. For every error, the count is 2n+1.

ankur0493:
For few commands, RClone counts a error multiple times. This was fixed by
creating a new error type which keeps a flag to remember if the error has
already been counted or not.

#3650 

#### Checklist

- [ x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -82,7 +82,7 @@ func ShowVersion() {
 func NewFsFile(remote string) (fs.Fs, string) {
 	_, _, fsPath, err := fs.ParseRemote(remote)
 	if err != nil {
-		fs.CountError(err)
+		err = fs.CountError(err)
 		log.Fatalf("Failed to create file system for %q: %v", remote, err)
 	}
 	f, err := cache.Get(remote)
@@ -92,7 +92,7 @@ func NewFsFile(remote string) (fs.Fs, string) {
 	case nil:
 		return f, ""
 	default:
-		fs.CountError(err)
+		err = fs.CountError(err)
 		log.Fatalf("Failed to create file system for %q: %v", remote, err)
 	}
 	return nil, ""
@@ -107,13 +107,13 @@ func newFsFileAddFilter(remote string) (fs.Fs, string) {
 	if fileName != "" {
 		if !filter.Active.InActive() {
 			err := errors.Errorf("Can't limit to single files when using filters: %v", remote)
-			fs.CountError(err)
+			err = fs.CountError(err)
 			log.Fatalf(err.Error())
 		}
 		// Limit transfers to this file
 		err := filter.Active.AddFile(fileName)
 		if err != nil {
-			fs.CountError(err)
+			err = fs.CountError(err)
 			log.Fatalf("Failed to limit to single file %q: %v", remote, err)
 		}
 	}
@@ -135,7 +135,7 @@ func NewFsSrc(args []string) fs.Fs {
 func newFsDir(remote string) fs.Fs {
 	f, err := cache.Get(remote)
 	if err != nil {
-		fs.CountError(err)
+		err = fs.CountError(err)
 		log.Fatalf("Failed to create file system for %q: %v", remote, err)
 	}
 	return f
@@ -189,11 +189,11 @@ func NewFsSrcDstFiles(args []string) (fsrc fs.Fs, srcFileName string, fdst fs.Fs
 	fdst, err := cache.Get(dstRemote)
 	switch err {
 	case fs.ErrorIsFile:
-		fs.CountError(err)
+		_ = fs.CountError(err)
 		log.Fatalf("Source doesn't exist or is a directory and destination is a file")
 	case nil:
 	default:
-		fs.CountError(err)
+		_ = fs.CountError(err)
 		log.Fatalf("Failed to create file system for destination %q: %v", dstRemote, err)
 	}
 	return
@@ -239,7 +239,7 @@ func Run(Retry bool, showStats bool, cmd *cobra.Command, f func() error) {
 	SigInfoHandler()
 	for try := 1; try <= *retries; try++ {
 		err = f()
-		fs.CountError(err)
+		err = fs.CountError(err)
 		lastErr := accounting.GlobalStats().GetLastError()
 		if err == nil {
 			err = lastErr
@@ -386,12 +386,12 @@ func initConfig() {
 		fs.Infof(nil, "Creating CPU profile %q\n", *cpuProfile)
 		f, err := os.Create(*cpuProfile)
 		if err != nil {
-			fs.CountError(err)
+			err = fs.CountError(err)
 			log.Fatal(err)
 		}
 		err = pprof.StartCPUProfile(f)
 		if err != nil {
-			fs.CountError(err)
+			err = fs.CountError(err)
 			log.Fatal(err)
 		}
 		atexit.Register(func() {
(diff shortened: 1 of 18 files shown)


ankur0493:
@ncw This change currently just covers the `Copy` function. We can extend the same change to a lot of other places to ensure errors aren't counted multiple times. I'll proceed if the approach looks good.

fs/fserrors/error.go
```diff
@@ -230,6 +230,61 @@ func IsRetryAfterError(err error) bool {
...
+
+// Counted sets the isCounted variable on the error if it conforms to the
+// CountableError interface
+func Count(err error) {
+	if r, ok := err.(CountableError); ok {
+		r.Count()
+	}
+}
+
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

fs/fserrors/error.go
```diff
@@ -230,6 +230,61 @@ func IsRetryAfterError(err error) bool {
...
+// IsCounted returns true if err conforms to the CountableError interface
+// and has already been counted
+func IsCounted(err error) bool {
+	if r, ok := err.(CountableError); ok {
+		return r.IsCounted()
+	}
+	return false
+}
+
+// Counted sets the isCounted variable on the error if it conforms to the
```
golangcibot: comment on exported function `Count` should be of the form `Count ...` (from `golint`)

ncw:
Deduplication of the error is a very interesting approach that I hadn't though of.

I guess the question is: do the deduplication in the error itself; or do it in the `CountError` function?

I think doing it in the `CountError` function might be easier.

Let's say we kept a slice of the last 5 errors, we could ignore the error if it was one of the last 5 errors. (note that potentially we might need to unwrap the incoming error to match it).

Note that interface comparison is a little tricksy in go: https://play.golang.org/p/mzX8vbyeAar

I think just comparing with `==` will be OK as there are very few non pointer type errors (but there are a few).

ankur0493:
> Let's say we kept a slice of the last 5 errors, we could ignore the error if it was one of the last 5 errors. 
Keeping a record of last x errors will never be 100% reliable. Let's say I am running 20 transfer threads with very small files, and all my transfers are failing. There will probably still be a possibility that we end up counting an error twice. However if we do that deduplication in the error itself, we can be sure that we won't count an error again.

> https://play.golang.org/p/mzX8vbyeAar
In this example, we are essentially doing `string("hello") == string("hello")`, which will return true. Will we see same behavior with the error interface?
We are already using different error interfaces and they work quite well.

ncw:
> > Let's say we kept a slice of the last 5 errors, we could ignore the error if it was one of the last 5 errors.
> 
> Keeping a record of last x errors will never be 100% reliable. Let's say I am running 20 transfer threads with very small files, and all my transfers are failing. There will probably still be a possibility that we end up counting an error twice.

Hmm, possibly... However counting an error twice isn't catastrophic.  We could always size the internal buffer to `fs.Config.Transfers + fs.Config.Checkers`.

> However if we do that deduplication in the error itself, we can be sure that we won't count an error again.

Sure, but then we need to change every error site which it would be nice not to!

Thinking aloud:

It might be a better interface to do

```
err = fs.CountError(err)
```

Where CountError counts the error and returns a `Counted` wrapper for `err` if `err` isn't `Counted`.

This will mean that `errcheck` will find all the places we didn't update.

> > https://play.golang.org/p/mzX8vbyeAar
> 
> In this example, we are essentially doing `string("hello") == string("hello")`, which will return true. Will we see same behavior with the error interface?

Probably not. But we could do... For example https://golang.org/pkg/syscall/#EAGAIN

This would mean we count fewer errors that we should.

As long as we never count 0 errors when we had some errors this doesn't matter too much.

fs/operations/operations.go
```diff
@@ -685,7 +685,7 @@ func checkIdentical(ctx context.Context, dst, src fs.Object) (differ bool, noHas
 	if !same {
 		err = errors.Errorf("%v differ", ht)
 		fs.Errorf(src, "%v", err)
-		fs.CountError(err)
+		err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

fs/operations/operations.go
```diff
@@ -716,7 +716,7 @@ func (c *checkMarch) DstOnly(dst fs.DirEntry) (recurse bool) {
 		}
 		err := errors.Errorf("File not in %v", c.fsrc)
 		fs.Errorf(dst, "%v", err)
-		fs.CountError(err)
+		err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

fs/operations/operations.go
```diff
@@ -734,7 +734,7 @@ func (c *checkMarch) SrcOnly(src fs.DirEntry) (recurse bool) {
 	case fs.Object:
 		err := errors.Errorf("File not in %v", c.fdst)
 		fs.Errorf(src, "%v", err)
-		fs.CountError(err)
+		err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

fs/operations/operations.go
```diff
@@ -784,7 +783,7 @@ func (c *checkMarch) Match(ctx context.Context, dst, src fs.DirEntry) (recurse b
 		} else {
 			err := errors.Errorf("is file on %v but directory on %v", c.fsrc, c.fdst)
 			fs.Errorf(src, "%v", err)
-			fs.CountError(err)
+			err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

fs/operations/operations.go
```diff
@@ -796,7 +795,7 @@ func (c *checkMarch) Match(ctx context.Context, dst, src fs.DirEntry) (recurse b
 		}
 		err := errors.Errorf("is file on %v but directory on %v", c.fdst, c.fsrc)
 		fs.Errorf(dst, "%v", err)
-		fs.CountError(err)
+		err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

cmd/cmd.go
```diff
@@ -189,11 +189,11 @@ func NewFsSrcDstFiles(args []string) (fsrc fs.Fs, srcFileName string, fdst fs.Fs
 	fdst, err := cache.Get(dstRemote)
 	switch err {
 	case fs.ErrorIsFile:
-		fs.CountError(err)
+		err = fs.CountError(err)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

ankur0493:
@ncw 

> Thinking aloud:
>It might be a better interface to do
err = fs.CountError(err)
>Where CountError counts the error and returns a Counted wrapper for err if err isn't Counted.
> This will mean that errcheck will find all the places we didn't update.

This sounds like a good approach. I updated my pull request with changes wherever were needed. All the tests are also passing now.

ncw:
That is looking good now - I'll merge it - thank you :-)

PR #3719, opened 2019-11-11
[WIP] Implement Mirror Backend

buengese:
A few months ago I started the implementation of mirror backend separate as was requested multiple times. However due to other responsibilities I didn't have the time to finish it. Now that I have a lot more free time I intend to work on this again.
This is a first very WIP implementation of this I'm using internally. Looking at the other feature requests I think it probably doesn't make much sense to separate this from union backend which was my original intention. However until I have figured out all the implementation details I will keep this separated.

### DIFF
--- a/backend/mirror/mirror.go
+++ b/backend/mirror/mirror.go
@@ -55,25 +55,21 @@ type Fs struct {
 	hashSet  hash.Set     // intersection of hash types
 }
 
-// Object describes a union Object
+// Object describes a mirror object
 //
-// This is a wrapped object which returns the Union Fs as its parent
+// Will definitely have info but maybe not meta
 type Object struct {
-	fs.Object
-	fs *Fs // what this object is part of
+	fs      *Fs
+	remote  string
+	objects []fs.Object
 }
 
-// Wrap an existing object in the union Object
-func (f *Fs) wrapObject(o fs.Object) *Object {
-	return &Object{
-		Object: o,
-		fs:     f,
-	}
-}
-
-// Fs returns the union Fs as the parent
-func (o *Object) Fs() fs.Info {
-	return o.fs
+// Dir is the special Mirrored dir
+type Dir struct {
+	fs     *Fs
+	id     string
+	remote string
+	dirs   []*fs.Dir
 }
 
 // Name of the remote (as passed into NewFs)
@@ -88,7 +84,20 @@ func (f *Fs) Root() string {
 
 // String converts this Fs to a string
 func (f *Fs) String() string {
-	return fmt.Sprintf("union root '%s'", f.root)
+	return fmt.Sprintf("mirror root '%s'", f.root)
+}
+
+// Creates from the parameters passed in a half finished Object which
+// must have setMetaData called on it
+//
+// Used to create new objects
+func (f *Fs) createObject(remote string) (o *Object) {
+	// Temporary Object under construction
+	o = &Object{
+		fs:     f,
+		remote: remote,
+	}
+	return o
 }
 
 // Features returns the optional features of this Fs
@@ -154,14 +163,17 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (o fs.Objec
 		fs.Debugf(srcObj, "Can't copy - not same remote type")
 		return nil, fs.ErrorCantCopy
 	}
-	for _, wr := range f.remotes {
-		o, err = wr.Features().Copy(ctx, src, remote)
+
+	obj := f.createObject(remote)
+	for _, roSrc := range srcObj.getRemotes() {
+		roDst, err := roSrc.Fs().Features().Copy(ctx, roSrc, remote)
 		if err != nil {
 			return nil, err
 		}
+		obj.addRemote(roDst)
 	}
 
-	return f.wrapObject(o), nil
+	return obj, nil
 }
 
 // Move src to this remote using server side move operations.
@@ -173,7 +185,7 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (o fs.Objec
 // Will only be called if src.Fs().Name() == f.Name()
 //
 // If it isn't possible then return fs.ErrorCantMove
-func (f *Fs) Move(ctx context.Context, src fs.Object, remote string) (o fs.Object, err error) {
+/*func (f *Fs) Move(ctx context.Context, src fs.Object, remote string) (o fs.Object, err error) {
 	srcObj, ok := src.(*Object)
 	if !ok {
 		fs.Debugf(srcObj, "Can't move - not same remote type")
@@ -188,7 +200,7 @@ func (f *Fs) Move(ctx context.Context, src fs.Object, remote string) (o fs.Objec
 	}
 
 	return f.wrapObject(o), nil
-}
+}*/
 
 // DirMove moves src, srcRemote to this remote at dstRemote
 // using server side move operations.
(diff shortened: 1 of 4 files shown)


backend/mirror/mirror.go
```diff
@@ -0,0 +1,589 @@
...
+		}
+
+		// set the reader to our read memory block
+		out := bytes.NewReader(inData)
+		for _, remote := range f.remotes {
+			o, err = remote.Put(ctx, out, src, options...)
+			if err != nil {
+				return nil, err
+			}
+			out.Seek(0, 0)
```
golangcibot: Error return value of `out.Seek` is not checked (from `errcheck`)

jkaberg:
@buengese I'm somewhat interested in this, however my use case is to duplicate whatever is uploaded onto all configured remotes with the "mirror remote" - would this be feasible? Looking at the code it seem's you only have one writeable remote?

buengese:
This is exactly what this remote is supposed to do (write all, read all and write all, read one). Please ignore some of the commands in the code the basis for this is copied from the union remote which was read all, write one.

backend/mirror/mirror.go
```diff
@@ -0,0 +1,657 @@
...
+// Will definitely have info but maybe not meta
+type Object struct {
+	fs      *Fs
+	remote  string
+	objects []fs.Object
+}
+
+// Dir is the special Mirrored dir
+type Dir struct {
+	fs     *Fs
```
golangcibot: `fs` is unused (from `structcheck`)

backend/mirror/mirror.go
```diff
@@ -0,0 +1,657 @@
...
+	id     string
```
golangcibot: `id` is unused (from `structcheck`)

backend/mirror/mirror.go
```diff
@@ -0,0 +1,657 @@
...
+	remote string
```
golangcibot: `remote` is unused (from `structcheck`)

backend/mirror/mirror.go
```diff
@@ -0,0 +1,657 @@
...
+	dirs   []*fs.Dir
```
golangcibot: `dirs` is unused (from `structcheck`)

backend/mirror/mirror.go
```diff
@@ -0,0 +1,677 @@
...
+// will return the object and the error, otherwise will return
+// nil and the error
+func (f *Fs) PutStream(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) (fs.Object, error) {
+	pr, pw := io.Pipe()
+	tee := io.TeeReader(in, pw)
+
+	var obj1, obj2 fs.Object
+	g, ctx := errgroup.WithContext(ctx)
+	g.Go(func() (err error) {
+		defer pw.Close()
```
golangcibot: Error return value of `pw.Close` is not checked (from `errcheck`)

buengese:
Ok I have all the tests passing now for the first time.

Some points worth discussing
The current Implementation uses a virtual mirror object that maintains a slice of all the underlying objects. For this reason we list all remotes on a List() call and match up the underlying this has the advantage that we allways check if the underlying remotes are in sync however it comes at significant performance penalty. The alternative would be to only list one underlying remote (e.g. the first) and create the other objects using NewObject only when a write access happens.

In Put we cache the file so we can seek and upload it sequentially to all underlying remotes. This obviously breaks accounting for now.

I have limited PutStream to 2 underlying remotes because it creates extra goroutines  but the current approach could be extended to an arbitrary number of remotes. Once again this breaks accounting.

 For Open() i implemented simple round robin load balancing I intend to make this configurable and maybe also implement accounting based load balancing.

The optional interfaces are only enabled if all underlying remotes support them. I'd like to use them too if only some remotes underlying support them falling back to download and upload only for the remotes that don't support them however this might not be possible with the current backend interface.

Pinging @ncw

ncw:
> Ok all the tests are passing now.

:-)
 
> Some points worth discussing
> The current Implementation uses a virtual mirror object that maintains a slice of all the underlying objects. For this reason we list all remotes on a List() call and match up the underlying this has the advantage that we allways check if the underlying remotes are in sync however it comes at significant performance penalty. The alternative would be to only list one underlying remote (e.g. the first) and create the other objects using NewObject only when a write access happens.

Let's take a step back for the moment.... What is the purpose of the mirror backend? Is it to make sure that remote1: and remote2: are always in sync at all times? Say like a RAID1 mirror?

What happens when one remote is unavailable?  Do we carry on with only one for reads and error on writes?

In which case your listing both the remotes is probably what the user would expect.

> In Put we cache the file so we can seek and upload it sequentially to all underlying remotes. This obviously breaks accounting for now.

(you mean Update)

I think this should work like Put does and just Tee the input streams.  Rclone doesn't rely on seeking when uploading.

What happens when the upload to one of the remotes fails?  Rclone will retry the upload. Do we really want to upload to all remotes again?

The accounting is fixable by wrapping and unwrapping the accounting probably...

> I have limited PutStream to 2 underlying remotes because it creates extra goroutines but the current approach could be extended to an arbitrary number of remotes. Once again this breaks accounting.

Go routines are cheap :-)
 
> For Open() i implemented simple round robin load balancing I intend to make this configurable and maybe also implement accounting based load balancing.

(Aside: Linux RAID1 works like this - it will read only one of the disks unless that read fails in which case it will read the other.  Which disk is chosen at random (this doubles the bandwidth) though you can weight it if you RAID1 an SSD and an HDD.)

> The optional interfaces are only enabled if all underlying remotes support them. I'd like to use them too if only some remotes underlying support them falling back to download and upload only for the remotes that don't support them however this might not be possible with the current backend interface.

Hmmm!

Maybe we should make some use cases to guide our thinking...

## Use 1 - duplication

User wants to back files up both to s3 and azure.  They set up a mirror backend so only 1 rclone sync is needed.

Not sure this buys you a great deal over running two rclone syncs.

## Use 2 - reliability

User wants more reliability by being able to read files from two places at once.  Let's say s3 and drive.  Normally we read from drive since it is cheaper but if drive is down then we'll read from s3.

You can't do this any other way right now.

I'm not sure that your code would fulfil this as it stands though.

## Use 3 - caching

User wants a fast local cache but stores everything online at the same time.  User specifies reads should come from local cache but writes go to both.

This is a similar use use case to the union backend, except you copy files to the local dir then `rclone move` them to the remote dir and the union backend makes it seamless exactly where they are.

....

Any other use cases?

It is tempting to say we should merge this with the union backend but I'm not sure...

buengese:
The way this draft is laid out follows my own use case which is very RAID1 one like. I use rclone as a library to provide cloud storage to an internal tool. The mirroring is strictly for data redundancy and I'm not actually interested in any load balancing because one of the underlying remotes reads way faster than all the others (write all, read one - the exact opposite of union). However I'm still listing all the remotes to make sure they stay in sync because some of the underlying remotes are accessed by other tools / people. 

I think this most closely matches the deduplication use case you mentioned.

The currently code does not fullfill the reliability use case. This could be easily adapted by changing list to ignore errors caused by a remote not being available and always return successfully as long as one remote lists completely. This use case also wouldn't require listing all remotes.

To be honest I'm not quite sure how to understand your caching use case.

Regarding accounting: correctly wrapping the streams does fix the stats but it still doesn't correctly register the number of concurrent transfers which is what I was actually concerned about. (Unless I've misread the code)

ncw:
> The way this draft is laid out follows my own use case which is very RAID1 one like. I use rclone as a library to provide cloud storage to an internal tool. The mirroring is strictly for data redundancy and I'm not actually interested in any load balancing because one of the underlying remotes reads way faster than all the others (write all, read one - the exact opposite of union). However I'm still listing all the remotes to make sure they stay in sync because some of the underlying remotes are accessed by other tools / people.

Ah OK!
 
> I think this most closely matches the deduplication use case you mentioned.
> 
> The currently code does not fullfill the reliability use case. This could be easily adapted by changing list to ignore errors caused by a remote not being available and always return successfully as long as one remote lists completely. This use case also wouldn't require listing all remotes.

That might be a nice addon.
 
> To be honest I'm not quite sure how to understand your caching use case.

It is like your use case but you have one fast local copy which you prefer to read from and one "backup" copy.

> Regarding accounting: correctly wrapping the streams does fix the stats but it still doesn't correctly register the number of concurrent transfers which is what I was actually concerned about. (Unless I've misread the code)

Ah I see....

Yes think showing anything other than 1 transfer will be very hard!

ivandeex:
@buengese You forgot to comment on then reason for closing.

buengese:
True. This has been superseeded by vastly superior new union backend. Which can do everything this was supposed to do and much more.

PR #3726, opened 2019-11-13
koofr: set rclone-specific user agent for Koofr backend.

jkmcnk:

Merely setting a rclone specific user agent, so that we can discriminate between clients on server side.

No.

#### Checklist

### DIFF
--- a/backend/koofr/koofr.go
+++ b/backend/koofr/koofr.go
@@ -263,6 +263,7 @@ func NewFs(name, root string, m configmap.Mapper) (ff fs.Fs, err error) {
 	basicAuth := fmt.Sprintf("Basic %s",
 		base64.StdEncoding.EncodeToString([]byte(opt.User+":"+pass)))
 	client.HTTPClient.Headers.Set("Authorization", basicAuth)
+	client.SetUserAgent(fs.Config.UserAgent)
 	mounts, err := client.Mounts()
 	if err != nil {
 		return nil, err


ncw:
Ah, this means that the koofr backend isn't using rclone's transport so (for example) `-vv --dump bodies` won't work and the timeouts won't be set and the `User-Agent` isn't set!

Is it possible to get the koofr backend to use the client or the transport instead of doing this?

https://github.com/rclone/rclone/blob/2784c3234bcf601dd9d0a47c7c43c12517802f53/fs/fshttp/http.go#L193

or

https://github.com/rclone/rclone/blob/2784c3234bcf601dd9d0a47c7c43c12517802f53/fs/fshttp/http.go#L201

jkmcnk:
hi,

I believe this is feasible. perhaps two of our own deps that wrap go http.Client need some slight constructor API changes, but I think we allow for setting the base go http.Client as it is. will look into it tomorrow. forget this PR then.

if we use the rclone client, the user agent will be set to rclone as well automagically, I suppose, yes?

ncw:
> I believe this is feasible. perhaps two of our own deps that wrap go http.Client need some slight constructor API changes, but I think we allow for setting the base go http.Client as it is. will look into it tomorrow. forget this PR then.

I think that would be a good addition to your library.  Eg google app engine requires you to pass its transport into all http calls.

> if we use the rclone client, the user agent will be set to rclone as well automagically, I suppose, yes?

Yes it will.  It is set to something like `User-Agent: rclone/v1.50.1-DEV`

ncw:
This is superseded by acefa5c40d8c87fa10e4aa11a012d1f751ceb8d7 so I'll close this now

Issue #1034, opened 2017-01-15
[Feature request] deleted files in stats


matrix1000:
Hi Nick,

it would be very nice if you can add the information of deletet files in the stat.
Currently (v1.35) you can see in stats: Errors, Checks and Transferred.
It would be perfect if I can also see the deleted files if I am using sync.

Many thanks.

ncw:
Nice idea.

chrissv:
Any plans on this?
Right now I see:
```
Transferred:       99.401M / 742.480 GBytes, 0%, 504.050 kBytes/s, ETA 428h59m41s
Errors:                 0
Checks:             34102 / 34106, 100%
Transferred:           18 / 156654, 0%
Elapsed time:     3m21.9s
```

It would be nice to see:
```
Transferred:       99.401M / 742.480 GBytes, 0%, 504.050 kBytes/s, ETA 428h59m41s
Errors:                 0
Checks:             34102 / 34106, 100%
Transferred:           18 / 156654, 0%
Deleted:                0 / 1024, 0%
Elapsed time:     3m21.9s
```

ncw:
Yes, I agree this would be very nice.  This also came up in #2723 

I'll move it up the priority list.

Any one want to have a go at it?

chrissv:
I code in C, but the GO code has me somewhat baffled. Any pointers where to start, data structures which hold the information?

ncw:
> I code in C, but the GO code has me somewhat baffled. Any pointers where to start, data structures which hold the information?

Looking at the code, I see that the deletes are already being counted, they just aren't displayed to the user.  So it is a question adding `s.deletes` here

https://github.com/ncw/rclone/blob/ea12d76c0354c8c099e253a7701f7dc7d848f95c/fs/accounting/stats.go#L254-L259

I'd quite like not to make another line of stats so I'm not 100% sure how they should be shown.

anuar45:
I am going to use rclone on raspbian for onedrive. Can I pickup this?

ncw:
> I am going to use rclone on raspbian for onedrive. Can I pickup this?

By all means :-)

anuar45:
> 
> 
> > I am going to use rclone on raspbian for onedrive. Can I pickup this?
> 
> By all means :-)

Thanks. 

One question. Why are you not sure about another line in stats?

ncw:
> One question. Why are you not sure about another line in stats?

It just starts making it a bit long.

[...]

PR #3730, opened 2019-11-13
stats: show deletes in stats

anuar45:

This change is about enabling showing deletes in stats output. 
Additionally it is omit zero stats.

https://github.com/rclone/rclone/issues/1034

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -313,11 +313,23 @@ func (s *StatsInfo) String() string {
 			errorDetails = " (no need to retry)"
 		}
 
-		addStatln(buf, "Errors:        %10d%s", s.errors, errorDetails)
-		addStatln(buf, "Checks:        %10d / %d, %s", s.checks, totalChecks, percent(s.checks, totalChecks))
-		addStatln(buf, "Deleted:       %10d", s.deletes)
-		addStatln(buf, "Transferred:   %10d / %d, %s", s.transfers, totalTransfer, percent(s.transfers, totalTransfer))
-		addStatln(buf, "Elapsed time:  %10v", dtRounded)
+		// Add only non zero stats
+		if s.errors != 0 {
+			_, _ = fmt.Fprintf(buf, "Errors:        %10d%s\n",
+				s.errors, errorDetails)
+		}
+		if s.checks != 0 || totalChecks != 0 {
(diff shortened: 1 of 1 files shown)


fs/accounting/stats.go
```diff
@@ -313,16 +313,11 @@ func (s *StatsInfo) String() string {
...
-Checks:        %10d / %d, %s
-Transferred:   %10d / %d, %s
-Elapsed time:  %10v
-`,
-			s.errors, errorDetails,
-			s.checks, totalChecks, percent(s.checks, totalChecks),
-			s.transfers, totalTransfer, percent(s.transfers, totalTransfer),
-			dtRounded)
+		addStatln(buf, "Errors:        %10d%s", s.errors, errorDetails)
+		addStatln(buf, "Checks:        %10d / %d, %s", s.checks, totalChecks, percent(s.checks, totalChecks))
```
ncw: This should probably show `if s.checks != 0 || totalChecks != 0`

fs/accounting/stats.go
```diff
@@ -313,16 +313,11 @@ func (s *StatsInfo) String() string {
...
+		addStatln(buf, "Deleted:       %10d", s.deletes)
+		addStatln(buf, "Transferred:   %10d / %d, %s", s.transfers, totalTransfer, percent(s.transfers, totalTransfer))
```
ncw: This should show `if s.transfers != 0 || totalTransfer != 0`

fs/accounting/stats.go
```diff
@@ -342,6 +337,19 @@ Elapsed time:  %10v
 	return buf.String()
 }
 
+// Wrapper around Fprintf to filer zero stat lines
+func addStatln(buf *bytes.Buffer, f string, a ...interface{}) {
```
ncw: I think given that we need more complicated tests, I think I would inline this into `if x != 0 { _,_ = fmt.Printf("...%d...\n", x}`and just write it out 4 times rather than attempting to make it any more complicated

anuar45: Thanks, I see what you mean. Sorry for complicating. I will be back with fix.

ncw: Give me a ping on the issue when you want another review :-)

ncw:
That looks great now - thanks!

I'll merge it now.

PR #3733, opened 2019-11-14
accounting: allow MaxCompletedTransfers to be configurable

### LINKED ISSUE #3732: accounting: not configurable limit of started transfers

v1.5.0 introduced limit of number of started transfers in accounting statistics which leads to races if number of transfered objects exceeds this limit. This race happens when user starts transfer and then fetches accounting stats. In case when some file was already transfered between start and stat fetch, information about it may be already deleted due pruning on transfer finish.

latest master

### LINKED ISSUE #3734: accounting: `stats-reset` doesn't clear finished transfers
latest master

High memory usage in case when transfering huge amount of objects in huge number of different groups. 

Remove finished transfer statistics on `stats-reset`.

zimnx:
Two changes in single PR since they depend on each other.

1.
rclone library users might be intrested in changing default value to other, or even disabling it. With current version it's impossible which leads to races when number of uploaded objects exceeds default limit.

Fixes #3732

2.
In order to reduce memory usage `stats-reset` also clears finished transfers.
 
Fixes #3734 
No

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -13,8 +13,8 @@ import (
 	"github.com/rclone/rclone/fs/rc"
 )
 
-// Maximum number of completed transfers in startedTransfers list
-const maxCompletedTransfers = 100
+// MaxCompletedTransfers specifies maximum number of completed transfers in startedTransfers list
+var MaxCompletedTransfers = 100
 
 // StatsInfo accounts all transfers
 type StatsInfo struct {
@@ -624,11 +624,15 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
 	s.mu.Unlock()
 }
 
-// PruneTransfers makes sure there aren't too many old transfers
+// PruneTransfers makes sure there aren't too many old transfers by removing
+// single finished transfer.
 func (s *StatsInfo) PruneTransfers() {
+	if MaxCompletedTransfers < 0 {
+		return
+	}
 	s.mu.Lock()
 	// remove a transfer from the start if we are over quota
-	if len(s.startedTransfers) > maxCompletedTransfers+fs.Config.Transfers {
+	if len(s.startedTransfers) > MaxCompletedTransfers+fs.Config.Transfers {
 		for i, tr := range s.startedTransfers {
 			if tr.IsDone() {
 				s.removeTransfer(tr, i)
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -624,6 +624,20 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
 	s.mu.Unlock()
 }
 
+// PruneAllTransfers removes all finished transfers.
+func (s *StatsInfo) PruneAllTransfers() {
+	s.mu.Lock()
+	for i := 0; i < len(s.startedTransfers); i++ {
+		tr := s.startedTransfers[i]
+		if tr.IsDone() {
+			s.removeTransfer(tr, i)
+			// i'th element is removed, recover iterator to not skip next element.
+			i--
+		}
+	}
+	s.mu.Unlock()
+}
+
 // PruneTransfers makes sure there aren't too many old transfers by removing
 // single finished transfer.
 func (s *StatsInfo) PruneTransfers() {
--- a/fs/accounting/stats_groups.go
+++ b/fs/accounting/stats_groups.go
@@ -59,8 +59,10 @@ func resetStats(ctx context.Context, in rc.Params) (rc.Params, error) {
 	}
 
 	if group != "" {
-		groups.get(group).ResetCounters()
-		groups.get(group).ResetErrors()
+		stats := groups.get(group)
+		stats.ResetCounters()
+		stats.ResetErrors()
+		stats.PruneAllTransfers()
 	} else {
 		groups.clear()
 	}
@@ -190,8 +192,8 @@ Returns the following values:
 		Fn:    resetStats,
 		Title: "Reset stats.",
 		Help: `
-This clears counters and errors for all stats or specific stats group if group
-is provided.
+This clears counters, errors and finished transfers for all stats or specific 
+stats group if group is provided.
 
 Parameters
 
@@ -333,6 +335,7 @@ func (sg *statsGroups) clear() {
 	for _, stats := range sg.m {
 		stats.ResetErrors()
 		stats.ResetCounters()
+		stats.PruneAllTransfers()
 	}
 
 	sg.m = make(map[string]*StatsInfo)
--- a/docs/content/rc.md
+++ b/docs/content/rc.md
@@ -521,8 +521,8 @@ The value for "eta" is null if an eta cannot be determined.
 
 ### core/stats-reset: Reset stats. {#core/stats-reset}
 
-This clears counters and errors for all stats or specific stats group if group
-is provided.
+This clears counters, errors and finished transfers for all stats or specific
+stats group if group is provided.
 
 Parameters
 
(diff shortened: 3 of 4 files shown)


fs/accounting/stats.go
```diff
@@ -624,17 +624,28 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
 	s.mu.Unlock()
 }
 
-// PruneTransfers makes sure there aren't too many old transfers
-func (s *StatsInfo) PruneTransfers() {
+// pruneTransfers removes transfers that already finished
+//
+// Must be called with the lock held
+func (s *StatsInfo) pruneTransfers() {
```
mmatczuk: Usually on such occasions it's called `pruneTransfersLocked`, we could move it below the main function too.

zimnx: I followed `RemoveTransfer` convention. I'll leave it as it is.

fs/accounting/stats.go
```diff
@@ -624,17 +624,28 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
...
+		if tr.IsDone() {
+			s.removeTransfer(tr, i)
+			break
+		}
+	}
+}
+
+// PruneTransfers makes sure there aren't too many old transfers.
+// It purges transfers if number of startedTransfers exceeds `limit`.
+func (s *StatsInfo) PruneTransfers(limit int) {
```
mmatczuk: I think that naming / behaviour can be misleading here, how do ppl know what limit to put and how do they know that sth would be added to the number?

I'd suggest having two functions
* `PruneTransfers()` - the usual logic that uses `MaxCompletedTransfers`
* `PruneAllTransfers()` - remove all logic used by `clear`

fs/accounting/stats.go
```diff
@@ -624,17 +624,28 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
...
 	s.mu.Lock()
 	// remove a transfer from the start if we are over quota
-	if len(s.startedTransfers) > maxCompletedTransfers+fs.Config.Transfers {
-		for i, tr := range s.startedTransfers {
-			if tr.IsDone() {
-				s.removeTransfer(tr, i)
-				break
-			}
-		}
+	if len(s.startedTransfers) > limit+fs.Config.Transfers {
```
mmatczuk: Wouldn't it be better to use `max(limit, fs.Config.Transfers)` here?

cc: @ncw

fs/accounting/stats.go
mmatczuk: Isn't it that this function should keep 100 last transfers by default?

zimnx: It does since it removes single transfer (note break in loop) and it's called once one of the transfer finishes. During applying your comment below I noticed it can't be used in `clear` because of it and I'm preparing patch as we speak.

fs/accounting/stats.go
zimnx: good idea, applied.

fs/accounting/stats.go
```diff
@@ -624,11 +624,29 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
 	s.mu.Unlock()
 }
 
-// PruneTransfers makes sure there aren't too many old transfers
+// PruneAllTransfers removes all finished transfers.
+func (s *StatsInfo) PruneAllTransfers() {
+	s.mu.Lock()
+	for i := 0; i < len(s.startedTransfers); i++ {
```
mmatczuk: How about range iterator?

fs/accounting/stats.go
```diff
@@ -624,11 +624,29 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
...
+// PruneTransfers makes sure there aren't too many old transfers by removing
+// single finished transfer.
 func (s *StatsInfo) PruneTransfers() {
+	if MaxCompletedTransfers < 0 {
+		return
+	}
 	s.mu.Lock()
 	// remove a transfer from the start if we are over quota
-	if len(s.startedTransfers) > maxCompletedTransfers+fs.Config.Transfers {
+	if len(s.startedTransfers) > MaxCompletedTransfers+fs.Config.Transfers {
```
mmatczuk: Now I wonder shouldn't this be max?

fs/accounting/stats_groups.go
```diff
@@ -59,8 +59,10 @@ func resetStats(ctx context.Context, in rc.Params) (rc.Params, error) {
 	}
 
 	if group != "" {
-		groups.get(group).ResetCounters()
-		groups.get(group).ResetErrors()
+		stats := groups.get(group)
+		stats.ResetCounters()
+		stats.ResetErrors()
+		stats.PruneTransfers()
```
mmatczuk: @zimnx this should be `PruneAllTransfers`

fs/accounting/stats.go
```diff
@@ -624,11 +624,29 @@ func (s *StatsInfo) RemoveTransfer(transfer *Transfer) {
 	s.mu.Unlock()
 }
 
-// PruneTransfers makes sure there aren't too many old transfers
+// PruneAllTransfers removes all finished transfers.
```
mmatczuk: @zimnx this function is not used, it should be used in cleanup flow IMO.

fs/accounting/stats.go
zimnx: I guess that was initial intention, although maybe I'm missing something. @ncw ?

fs/accounting/stats.go
zimnx: in range iterator stop condition is not checked on each iteration, and it goes off slice boundaries when any element is removed.

fs/accounting/stats.go
zimnx: right, I screwed this when I was resolving conflicts. On it.

fs/accounting/stats.go
mmatczuk: Sometimes it works for instance with maps, but I guess it's not the case here. How about https://github.com/rclone/rclone/pull/3733/files#diff-3db88cbb7dcd4ae3715a81f3543ae867R650 then?

fs/accounting/stats.go
zimnx: done

ncw:
I think this is looking good now - was there anything else you wanted to add before merging?

zimnx:
> I think this is looking good now - was there anything else you wanted to add before merging?

@ncw There's one comment above regarding using `max` instead sum of `MaxCompletedTransfer` and minimum slice length, but it's not so critical. If sum is fine, there's nothing more to add from my side.

ncw:
> There's one comment above regarding using `max` instead sum of `MaxCompletedTransfer` and minimum slice length, but it's not so critical. If sum is fine, there's nothing more to add from my side.

I think it is OK as it is as there will be up to `fs.Config.Transfers` which we don't want to get rid of as they are in progress.

The logic is a bit sloppy either way but I don't think that matters too much!

I'll merge this now - thank you for your contribution :-)

zimnx:
@ncw can we ask minor release with this fix? Otherwise we would have to fork repo, and come back to original repo once this will be released.

ncw:
> @ncw can we ask minor release with this fix? Otherwise we would have to fork repo, and come back to original repo once this will be released.

I try very hard not to break stuff when issuing patch releases by only putting essential fixes in and don't add new features, so I don't think this qualifies unfortunately.

zimnx:
> > @ncw can we ask minor release with this fix? Otherwise we would have to fork repo, and come back to original repo once this will be released.
> 
> I try very hard not to break stuff when issuing patch releases by only putting essential fixes in and don't add new features, so I don't think this qualifies unfortunately.

Understood. Thanks!

Issue #3706, opened 2019-11-08
Rework Config System


fionera:
#### What problem are you are trying to solve?
Currently rclone uses `goconfig` to load a ini file. The implementation behind it does not allow the use of more complicated data structures, like array and maps. Currently rclone circumvents this with "CommaSepList" and "SpaceSepList". I started this because I have already changed to yaml files in my fork, to be able to use a map[string][]string structure.

#### How do you think rclone should be changed to solve that?
This is a Thread to discuss a good interface for Config and the implementation of it.

My first Suggestion would be smth like this:
```golang

// ProviderDefinition defines the supported FileTypes (the extension)
// that the Provider can handle. It will be created by calling the `NewFunc`.
// The `NewFunc` should not do any loading.
// The definitions would be registered on init, which allows endless modifications
type ProviderDefinition struct {
	NewFunc   func() Provider
	FileTypes []string
}

// The basic interface a config Provider has to implement
// String returns the raw Config
// Load reads the already decrypted config file content
// Save writes the unencrypted config file content
// GetRemoveConfig returns an implementation of the RemoteConfig interface
type Provider interface {
	String() string
	Load(io.Reader) error
	Save(io.Writer) error
	GetRemoteConfig() RemoteConfig
}

// GlobalConfigProvider is an extension for a config Provider
type GlobalConfigProvider interface {
	GetGlobalConfig() GlobalConfig
}

// GlobalConfig is an interface to allow the set of global settings like `bwlimit`
// It has to support loading and saving of string values by default
type GlobalConfig interface {
	StringSupporter
}

// RemoteConfig is an interface to use the config of the remotes
// GetRemotes returns a slice with all the remote names
// HasRemotes returns if the given remote exists in the config
// GetRemote returns the config section of the named remote or nil if it does not exist
// CreateRemote creates a remote and returns the section for it
// DeleteRemote deletes a remote
// RenameRemote renames a remote
// CopyRemote copies the config of the given remote to a clone with the given name
type RemoteConfig interface {
	GetRemotes() []string
	HasRemote(remote string) bool
	GetRemote(remote string) Section
	CreateRemote(remote string) Section
	DeleteRemote(name string)
	RenameRemote(oldName string, newName string)
	CopyRemote(source string, destination string)
}

// Section is the config of a single remote
// GetKeys returns all set options
// GetConfig returns the raw config
// Remove deletes the config option with the given name
// It has to support loading and saving of string values by default
type Section interface {
	GetKeys() []string
[...]

PR #3747, opened 2019-11-18
[WIP] Refactor config package to use an interface

fionera:
Allow replacing of the Config Library by wrapping calls with an interface

#3706 

#### Checklist


fs/config/configfile/config_file.go
```diff
@@ -0,0 +1,60 @@
...
+	HasSection(section string) bool
+	DeleteSection(section string)
+
+	GetKeyList(section string) []string
+	GetValue(section string, key string) (string, error)
+	MustValue(section string, key string, default_ ...string) string
+	SetValue(section string, key string, value string)
+	DeleteKey(section string, key string) bool
+
+	Load(io.Reader) error
```
ncw, line 20: I think having these take an `io.Reader` is probably an implementation detail, so maybe the interface should just be `Load() error) and `Save() error` - what do you think?

What is your plan for this? Do you want rclone to be able to encrypt its config? It doesn't seem quite flexible enough somehow.

ncw (commented):
Great work! I think this is getting there. I think something is not quite right about the loading and saving yet.

In an ideal world we'd be able to import parts of rclone and not have to have `github.com/Unknwon/goconfig` as a dependency.

This probably means the interface should go into `fs/config` and that actually setting the value of `configFile configfile.ConfigFile` should go elsewhere, probably into `cmd/cmd.go` which is kind of like rclone's main file.

fs/config/configfile/config_file.go
fionera: I added the `io.Reader` because with that I was able to not touch the encryption system. This would ofc. be a good opportunity to move that part too. Then I would add a NewGoConfig function which takes the reader and returns the ConfigFile

fionera:
Yes currently the config source is hardcoded. I was just pushing an example to check if this is ok for you. I am thinking about setting the LoadConfig and SaveConfig functions to nil and then setting them to the GoConfig ones somewhere at start. This would completely remove any dependency to goconfig for external projects.

fs/config/configfile/config_file.go
ncw: > Yes currently the config source is hardcoded. I was just pushing an example to check if this is ok for you.

It is a great idea.

> I am thinking about setting the LoadConfig and SaveConfig functions to nil and then setting them to the GoConfig ones somewhere at start. This would completely remove any dependency to goconfig for external projects.

Hmm, I think maybe we are mixing up two possible concepts

1. The setting and getting of config data
2. The serializing of the config data into a config file.

I'm presuming that if you swap out 1 you'll provide your own implementation of 2 - is that your understanding?

I'm thinking that what you might do is use some kind of database for reading and writing the config. In which case the serialization into a config file is irrelevant and un-needed.

In which case all the saving and loading to files (including the encryption) should be in `fs/config/configfile` and the interface should just have a `Load() error` and `Save() error` in it which likely would be ignored by someone implementing their own config system.

What do you think?

ivandeex:
@fionera Please address review comments

fs/config/configfile/config_file.go
fionera: Sorry for not responding :) Didnt had much time lately. I think your idea is good, but I dont know yet how I would like to implement the registering of the custom config backend. I will play a bit with the source when I find time the next days

ivandeex:
Marking as draft until this PR shows any signs of life.

fs/config/configfile/config_file.go
fionera: I would like to have the encryption seperated from the config interface/implementation since the implementation is so generic that you can use it for other providers too.

ncw:
Thanks for doing this @fionera and sorry for the delay.

I think it is a really nice idea. It unfortunately touches the very beating heart of rclone so it doesn't quite work as-is.

I'll fix up the patch and push it back here when I'm done with some more stuff on top.

ncw:
OK this looks good now!

I'll merge it.

Thank you @fionera for your code. You'll see that I've change the names of the interfaces but otherwise it should function exactly as your original code.

PR #3748, opened 2019-11-18
drive: add --drive-use-shared-date to use date file was shared

### LINKED ISSUE #3624: drive: New flag to track share date of files/dirs with `--drive-shared-with-me`
1.49.5

Generate a list of shared files and directories with `--drive-shared-with-me` tag based on share date. Since shared files/directories can be generated in a large number with a mixture of different file names and/or name patents. 

Traditionally, `--filter-from` might solve this problem, but when working with a number of over 100+ files/dirs it becomes really difficult to do.

`rclone` has already had `--max-age` tag to track creation/modification time. I think implement another tag for example `--max-share-age` to track the share date would solve this problem.

garrmcnu:

Add --drive-use-shared-date to use date file was shared instead of modified date.

Fixes #3624

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -327,6 +327,17 @@ or move the photos locally and use the date the image was taken
 (created) set as the modification date.`,
 			Advanced: true,
 		}, {
+			Name:    "use_shared_date",
+			Default: false,
+			Help: `Use date file was shared instead of modified date.
+
+Note that, as with "--drive-use-created-date", this flag may have
+unexpected consequences when uploading/downloading files.
+
+If both this flag and "--drive-use-created-date" are set, the created
+date is used.`,
+			Advanced: true,
+		}, {
 			Name:     "list_chunk",
 			Default:  1000,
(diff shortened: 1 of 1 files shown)


backend/drive/drive.go
```diff
@@ -326,6 +326,14 @@ Photos folder" option in your google drive settings. You can then copy
 or move the photos locally and use the date the image was taken
 (created) set as the modification date.`,
 			Advanced: true,
+		}, {
+			Name:    "use_shared_date",
+			Default: false,
+			Help: `Use date file was shared instead of modified date.
+
+Note that, as with "--drive-use-created-date", this flag may have
+unexpected consequences when uploading/downloading files.`,
```
ncw: Perhaps note here that `--drive-use-created-date` takes precedence if both set?

garrmcnu: Thanks for the feedback, that sounds good, will add that.

ncw:
Super thank you :-)  I'll merge this now.  Sorry I missed the notification that you updated the PR.

PR #3749, opened 2019-11-19
allow specification of config password dynamically - fixes #3694

### LINKED ISSUE #3694: allow specification of config password dynamically
rclone v1.50.0
- os/arch: darwin/amd64
- go version: go1.13.3

I want to secure my config with a password, but I never wish to either know the password, nor imbed it in a script, nor pass it in an environment variable.

There is a discussion in the Borg backup docs regarding the desirability for this feature, and for a model for implementing such.
See, for example: [here](https://borgbackup.readthedocs.io/en/stable/faq.html?highlight=borg_passcommand).  Look for the BORG_PASSCOMMAND info.

permezel:
#### Checklist

- [x ] I'm done, this Pull Request is ready for review :-)

よろしくお願いします。

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -880,6 +880,14 @@ Rclone will do its best to transfer the best file it has so in
 practice this should not cause a problem.  Think of `--order-by` as
 being more of a best efforts flag rather than a perfect ordering.
 
+### --password-command=STRING ###
+
+This flag supplies a program which should supply the config password
+when run. This is an alternative to rclone prompting for the password
+or setting the `RCLONE_CONFIG_PASS` variable.
+
+See the [Configuration Encryption](#configuration-encryption) for more info.
+
 ### -P, --progress ###
 
 This flag makes rclone update the stats in a static block in the
@@ -1361,13 +1369,13 @@ which will retrieve the password and print on standard output.  This
 script should have a fully specified path name and not rely on any
 environment variables.  The script is supplied either via
 `--password-command="..."` command line argument or via the
-`RCLONE_CONFIG_PASS_COMMAND` environment variable.
+`RCLONE_PASSWORD_COMMAND` environment variable.
 
 One useful example of this is using the `passwordstore` application
 to retrieve the password:
 
 ```
-export RCLONE_CONFIG_PASS_COMMAND="pass rclone/config"
+export RCLONE_PASSWORD_COMMAND="pass rclone/config"
 ```
 
 If the `passwordstore` password manager holds the password for the
@@ -1381,11 +1389,11 @@ script method of supplying the password enhances the security of
 the config password considerably.
 
 If you are running rclone inside a script, unless you are using the
-`RCLONE_CONFIG_PASS_COMMAND` method, you might want to disable 
+`--password-command` method, you might want to disable 
 password prompts. To do that, pass the parameter 
 `--ask-password=false` to rclone. This will make rclone fail instead
 of asking for a password if `RCLONE_CONFIG_PASS` doesn't contain
-a valid password, and `RCLONE_CONFIG_PASS_COMMAND` has not been supplied.
+a valid password, and `--password-command` has not been supplied.
 
 
 Developer options
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -1285,7 +1285,7 @@ your cloud services. This means that you should keep your
 
 If you are in an environment where that isn't possible, you can
 add a password to your configuration. This means that you will
-have to enter the password every time you start rclone.
+have to supply the password every time you start rclone.
 
 To add a password to your rclone configuration, execute `rclone config`.
 
@@ -1322,9 +1322,9 @@ c/u/q>
 ```
 
 Your configuration is now encrypted, and every time you start rclone
-you will now be asked for the password. In the same menu, you can
-change the password or completely remove encryption from your
-configuration.
+you will have to supply the password. See below for details.
+In the same menu, you can change the password or completely remove
+encryption from your configuration.
 
 There is no way to recover the configuration if you lose your password.
 
@@ -1356,11 +1356,36 @@ Then source the file when you want to use it.  From the shell you
 would do `source set-rclone-password`.  It will then ask you for the
 password and set it in the environment variable.
 
-If you are running rclone inside a script, you might want to disable 
+An alternate means of supplying the password is to provide a script
+which will retrieve the password and print on standard output.  This
+script should have a fully specified path name and not rely on any
+environment variables.  The script is supplied either via
+`--password-command="..."` command line argument or via the
+`RCLONE_CONFIG_PASS_COMMAND` environment variable.
+
+One useful example of this is using the `passwordstore` application
+to retrieve the password:
+
+```
+export RCLONE_CONFIG_PASS_COMMAND="pass rclone/config"
+```
+
+If the `passwordstore` password manager holds the password for the
+rclone configuration, using the script method means the password
+is primarily protected by the `passwordstore` system, and is never
+embedded in the clear in scripts, nor available for examination
+using the standard commands available.  It is quite possible with
+long running rclone sessions for copies of passwords to be innocently
+captured in log files or terminal scroll buffers, etc.  Using the
+script method of supplying the password enhances the security of
+the config password considerably.
+
+If you are running rclone inside a script, unless you are using the
+`RCLONE_CONFIG_PASS_COMMAND` method, you might want to disable 
 password prompts. To do that, pass the parameter 
 `--ask-password=false` to rclone. This will make rclone fail instead
 of asking for a password if `RCLONE_CONFIG_PASS` doesn't contain
-a valid password.
+a valid password, and `RCLONE_CONFIG_PASS_COMMAND` has not been supplied.
 
 
 Developer options
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -880,11 +880,23 @@ Rclone will do its best to transfer the best file it has so in
 practice this should not cause a problem.  Think of `--order-by` as
 being more of a best efforts flag rather than a perfect ordering.
 
-### --password-command=STRING ###
+### --password-command SpaceSepList ###
 
 This flag supplies a program which should supply the config password
 when run. This is an alternative to rclone prompting for the password
 or setting the `RCLONE_CONFIG_PASS` variable.
+
+The argument to this should be a command with a space separated list
+of arguments. If one of the arguments has a space in then enclose it
+in `"`, if you want a literal `"` in an argument then enclose the
+argument in `"` and double the `"`. See [CSV encoding](https://godoc.org/encoding/csv)
+for more info.
+
+Eg
+
+    --password-command echo hello
+    --password-command echo "hello with space"
+    --password-command echo "hello with ""quotes"" and space"
 
 See the [Configuration Encryption](#configuration-encryption) for more info.
 
--- a/fs/config.go
+++ b/fs/config.go
@@ -89,7 +89,7 @@ type ConfigInfo struct {
 	StreamingUploadCutoff  SizeSuffix
 	StatsFileNameLength    int
 	AskPassword            bool
-	PasswordCommand        string
+	PasswordCommand        SpaceSepList
 	UseServerModTime       bool
 	MaxTransfer            SizeSuffix
 	MaxBacklog             int
--- a/fs/config.go
+++ b/fs/config.go
@@ -89,6 +89,7 @@ type ConfigInfo struct {
 	StreamingUploadCutoff  SizeSuffix
 	StatsFileNameLength    int
 	AskPassword            bool
+	PasswordCommand        string
 	UseServerModTime       bool
 	MaxTransfer            SizeSuffix
 	MaxBacklog             int
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -275,10 +275,6 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 
 	if len(configKey) == 0 {
 		pwc := fs.Config.PasswordCommand
-
-		if pwc == "" {
-			pwc = os.Getenv("RCLONE_CONFIG_PASS_COMMAND")
-		}
 		if pwc != "" {
 			var stdout bytes.Buffer
 			var stderr bytes.Buffer
@@ -291,7 +287,7 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 
 			if err := cmd.Run(); err != nil {
 				// One does not always get the stderr returned in the wrapped error.
-				fs.Errorf(nil, "Using RCLONE_CONFIG_PASS_COMMAND returned: %v", err)
+				fs.Errorf(nil, "Using --password-command returned: %v", err)
 				if ers := strings.TrimSpace(stderr.String()); ers != "" {
 					fs.Errorf(nil, "--password-command stderr: %s", ers)
 				}
@@ -357,7 +353,7 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 		} else {
 			if len(configKey) == 0 {
 				if usingPasswordCommand {
-					return nil, errors.New("using password command derived password, unable to decrypt configuration")
+					return nil, errors.New("using --password-command derived password, unable to decrypt configuration")
 				}
 				if !fs.Config.AskPassword {
 					return nil, errors.New("unable to decrypt configuration and not allowed to ask for password - set RCLONE_CONFIG_PASS to your configuration password")
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -274,13 +274,11 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 	}
 
 	if len(configKey) == 0 {
-		pwc := fs.Config.PasswordCommand
-		if pwc != "" {
+		if len(fs.Config.PasswordCommand) != 0 {
 			var stdout bytes.Buffer
 			var stderr bytes.Buffer
 
-			args := strings.Fields(pwc)
-			cmd := exec.Command(args[0], args[1:]...)
+			cmd := exec.Command(fs.Config.PasswordCommand[0], fs.Config.PasswordCommand[1:]...)
 
 			cmd.Stdout = &stdout
 			cmd.Stderr = &stderr
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -14,6 +14,7 @@ import (
 	"log"
 	mathrand "math/rand"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -239,15 +240,7 @@ var errorConfigFileNotFound = errors.New("config file not found")
 // loadConfigFile will load a config file, and
 // automatically decrypt it.
 func loadConfigFile() (*goconfig.ConfigFile, error) {
-	envpw := os.Getenv("RCLONE_CONFIG_PASS")
-	if len(configKey) == 0 && envpw != "" {
-		err := setConfigPassword(envpw)
-		if err != nil {
-			fs.Errorf(nil, "Using RCLONE_CONFIG_PASS returned: %v", err)
-		} else {
-			fs.Debugf(nil, "Using RCLONE_CONFIG_PASS password.")
-		}
-	}
+	var usingPasswordCommand bool
 
 	b, err := ioutil.ReadFile(ConfigPath)
 	if err != nil {
@@ -280,6 +273,59 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 		return goconfig.LoadFromReader(bytes.NewBuffer(b))
 	}
 
+	if len(configKey) == 0 {
+		pwc := fs.Config.PasswordCommand
+
+		if pwc == "" {
+			pwc = os.Getenv("RCLONE_CONFIG_PASS_COMMAND")
+		}
+		if pwc != "" {
+			var stdout bytes.Buffer
+			var stderr bytes.Buffer
+
+			args := strings.Fields(pwc)
+			cmd := exec.Command(args[0], args[1:]...)
+
+			cmd.Stdout = &stdout
+			cmd.Stderr = &stderr
+
+			if err := cmd.Run(); err != nil {
+				// One does not always get the stderr returned in the wrapped error.
+				fs.Errorf(nil, "Using RCLONE_CONFIG_PASS_COMMAND returned: %v", err)
+				if ers := strings.TrimSpace(stderr.String()); ers != "" {
+					fs.Errorf(nil, "--password-command stderr: %s", ers)
+				}
+				return nil, errors.Wrap(err, "password command failed")
+			}
+			if pass := strings.Trim(stdout.String(), "\r\n"); pass != "" {
+				err := setConfigPassword(pass)
+				if err != nil {
+					return nil, errors.Wrap(err, "incorrect password")
+				}
+			} else {
+				return nil, errors.New("password-command returned empty string")
+			}
+
+			if len(configKey) == 0 {
+				return nil, errors.New("unable to decrypt configuration: incorrect password")
+			}
+			usingPasswordCommand = true
+		} else {
+			usingPasswordCommand = false
+
+			envpw := os.Getenv("RCLONE_CONFIG_PASS")
+
+			if envpw != "" {
+				err := setConfigPassword(envpw)
+				if err != nil {
+					fs.Errorf(nil, "Using RCLONE_CONFIG_PASS returned: %v", err)
+				} else {
+					fs.Debugf(nil, "Using RCLONE_CONFIG_PASS password.")
+				}
+			}
+		}
+	}
+
 	// Encrypted content is base64 encoded.
 	dec := base64.NewDecoder(base64.StdEncoding, r)
 	box, err := ioutil.ReadAll(dec)
@@ -310,6 +356,9 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 			fs.Debugf(nil, "using _RCLONE_CONFIG_KEY_FILE for configKey")
 		} else {
 			if len(configKey) == 0 {
+				if usingPasswordCommand {
+					return nil, errors.New("using password command derived password, unable to decrypt configuration")
+				}
 				if !fs.Config.AskPassword {
 					return nil, errors.New("unable to decrypt configuration and not allowed to ask for password - set RCLONE_CONFIG_PASS to your configuration password")
 				}
--- a/fs/config/config_test.go
+++ b/fs/config/config_test.go
@@ -270,6 +270,84 @@ func TestConfigLoadEncrypted(t *testing.T) {
 	assert.Equal(t, expect, keys)
 }
 
+func expectConfigValid(t *testing.T) {
+	var err error
+
+	configKey = nil // reset password
+
+	c, err := loadConfigFile()
+	require.NoError(t, err)
+
+	// Not sure why there is no "RCLONE_ENCRYPT_V0" here...
+	sections := c.GetSectionList()
+	var expect = []string{"nounc", "unc"}
+	assert.Equal(t, expect, sections)
+
+	keys := c.GetKeyList("nounc")
+	expect = []string{"type", "nounc"}
+	assert.Equal(t, expect, keys)
+}
+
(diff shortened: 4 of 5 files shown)


fs/config/config.go
```diff
@@ -233,16 +234,58 @@ var errorConfigFileNotFound = errors.New("config file not found")
...
 func loadConfigFile() (*goconfig.ConfigFile, error) {
-	envpw := os.Getenv("RCLONE_CONFIG_PASS")
-	if len(configKey) == 0 && envpw != "" {
-		err := setConfigPassword(envpw)
-		if err != nil {
-			fs.Errorf(nil, "Using RCLONE_CONFIG_PASS returned: %v", err)
-		} else {
-			fs.Debugf(nil, "Using RCLONE_CONFIG_PASS password.")
+	var passFail bool
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

fs/config/config_test.go
```diff
@@ -256,6 +256,68 @@ func TestConfigLoadEncrypted(t *testing.T) {
...
+	// using fs.Config.PasswordCommand, correct password
+	fs.Config.PasswordCommand = "echo asdf"
+	ExpectConfigValid(t)
+	// using fs.Config.PasswordCommand, incorrect password
+	fs.Config.PasswordCommand = "echo asdf-blurfl"
+	ExpectConfigInvalid(t)
+	fs.Config.PasswordCommand = ""
+
+	var err error
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

fs/config/config.go
```diff
@@ -233,14 +234,56 @@ var errorConfigFileNotFound = errors.New("config file not found")
...
+			cmd.Stdout = &out
+			cmd.Env = os.Environ()
+
+			if err := cmd.Run(); err != nil {
+				fs.Errorf(nil, "password-command execution error: %v", err)
+			}
+			if pass := out.String(); pass != "" && pass != "\n" {
+				pass := strings.TrimSuffix(pass, "\n")
+				err := setConfigPassword(pass)
+				pass = ""	// wipe(pass); wipe(&out)
```
golangcibot: ineffectual assignment to `pass` (from `ineffassign`)

fs/config/config.go
```diff
@@ -233,14 +234,56 @@ var errorConfigFileNotFound = errors.New("config file not found")
...
+
+		if envpw != "" {
+			err := setConfigPassword(envpw)
+			if err != nil {
+				fs.Errorf(nil, "Using RCLONE_CONFIG_PASS returned: %v", err)
+			} else {
+				fs.Debugf(nil, "Using RCLONE_CONFIG_PASS password.")
+			}
 		}
+		envpw = ""	// *** wipe(envpw)
```
golangcibot: ineffectual assignment to `envpw` (from `ineffassign`)

permezel:
That is not the way I read the docs.
exec.Command returns a cmd, and only the Path and Args have been set.
If Env is not set, it will use the envoronment from the current process.
I explicitly want os.Environ().
I prefer not to rely on squishy interface definitions which a) appear to be in dispute and b) may change.
Making things explicit is always better.

permezel:
type ExitError <https://golang.org/src/os/exec/exec.go?s=13315:13831#L445>
An ExitError reports an unsuccessful exit by a command.

type ExitError struct {
    *os <https://golang.org/pkg/os/>.ProcessState <https://golang.org/pkg/os/#ProcessState>

    // Stderr holds a subset of the standard error output from the
    // Cmd.Output method if standard error was not otherwise being
    // collected.
    //
    // If the error output is long, Stderr may contain only a prefix
    // and suffix of the output, with the middle replaced with
    // text about the number of omitted bytes.
    //
    // Stderr is provided for debugging, for inclusion in error messages.
    // Users with other needs should redirect Cmd.Stderr as needed.
    Stderr []byte <https://golang.org/pkg/builtin/#byte> // Go 1.6
}

So I guess it depends on how %v displays the ExitError.

At this point, I need to a) sleep and b) probably read a book on go programming.

permezel:
I lost the flag.

Now my tests fail:

=== RUN   TestConfigLoadEncryptedWithValidPassCommand
--- PASS: TestConfigLoadEncryptedWithValidPassCommand (0.01s)
=== RUN   TestConfigLoadEncryptedWithInvalidPassCommand
2019/11/24 01:43:13 ERROR : Couldn't decrypt configuration, most likely wrong password.
Enter configuration password:
password:2019/11/24 01:43:13 Failed to read line: EOF
exit status 1

It is not possible for me to tell, up where I set that flag, if the password really is correct.
The full test is much further below.
If it actually fails to decrypt, without that flag being introduced, it prompts for a password.

I could override the global flag setting, but using a local flag is much cleaner.

"passFail" was a poor choice, though.  "dontPrompt" would have been a better choice.

I also find that if I provide a password, if the config was not encrypted, I cannot tell that it was not needed.
I think that should be an error.  I think my config is encrypted, and am supplying the password, and nothing is letting me know that hey, perhaps you need to actually enable the encryption you think you have.

ncw:
> That is not the way I read the docs. exec.Command returns a cmd, and only the Path and Args have been set. If Env is not set, it will use the environment from the current process. I explicitly want os.Environ().

Isn't the environment from the current process exactly equal to `os.Environ()`?

> // Stderr holds a subset of the standard error output from the Cmd.Output method if standard error was not otherwise being collected.

I didn't know that - scratch my comment about unhelpful error messages :-)

> I could override the global flag setting, but using a local flag is much cleaner.

If that is the way the code turns out then I'll take a look at it when you update the PR :-)

permezel:
> On 2019-Nov-24, at 02:13, Nick Craig-Wood <notifications@github.com> wrote:
> 
> That is not the way I read the docs. exec.Command returns a cmd, and only the Path and Args have been set. If Env is not set, it will use the environment from the current process. I explicitly want os.Environ().
> 
> Isn't the environment from the current process exactly equal to os.Environ()?
> 
> 
Um… come to think of it, yes.
What I intended was to clear the environment, and for some reason thought “os.Env” was “OS’s ENV”,  not processes.
I yield.

> // Stderr holds a subset of the standard error output from the Cmd.Output method if standard error was not otherwise being collected.
> 
> I didn't know that - scratch my comment about unhelpful error messages :-)
> 
> 
I have had some experience now with the generation of these messages.

If the programme just returns with some exit code, then yes, you get this unhelpful status code which does not seem to be correlated with the exit code I try to supply.

If it writes to stderr, then you do, I think, get the stderr.  It was 2am this morning, so I might be making this all up.  I’ll play with it a bit more.

As the programme being invoked supposedly provides sensitive password info, I want to err on the side of providing less of the output in an error message, just in case there is sensitive stuff there.
By that I mean, stderr is ok, stdout probably not, when the command fails.
> I could override the global flag setting, but using a local flag is much cleaner.
> 
> If that is the way the code turns out then I'll take a look at it when you update the PR :-)
> 

Thanks for being patient with me.  Go is very weird.  Time I learned it.

ncw:
Sorry missed your reply!  Where are we with this pull request? I'd like to merge this for 1.51 (which I'm going to delay until the new year) as it is a great feature :-)

permezel:
Hi Nick.
This is a priority for me.
Unfortunately, it never pops to the top of the stack, due to other higher priority tasks never being completed.  How I ever actually manage to accomplish stuff, I can no longer understand.

I will endeavour to refresh my brain cells regarding what I need to do to re-issue a pull.

Cheers,
Damon

permezel:
Just pushed code up.
Got macos failure:

2019/12/27 09:19:05 Failed to get memory for buffer, waiting for 1ms: failed to allocate memory
622
 <https://github.com/rclone/rclone/runs/364887180#step:9:622>2019/12/27 09:19:05 Failed to get memory for buffer, waiting for 2ms: failed to allocate memory
623
 <https://github.com/rclone/rclone/runs/364887180#step:9:623>2019/12/27 09:19:05 Failed to free memory: failed to free memory
624
 <https://github.com/rclone/rclone/runs/364887180#step:9:624>2019/12/27 09:19:05 Failed to free memory: failed to free memory
625
 <https://github.com/rclone/rclone/runs/364887180#step:9:625>2019/12/27 09:19:05 Failed to free memory: failed to free memory
626
 <https://github.com/rclone/rclone/runs/364887180#step:9:626>--- FAIL: TestPool (0.77s)
627
 <https://github.com/rclone/rclone/runs/364887180#step:9:627>    --- FAIL: TestPool/canFail (0.22s)
628
 <https://github.com/rclone/rclone/runs/364887180#step:9:628>        --- FAIL: TestPool/canFail/Flusher (0.19s)
629
 <https://github.com/rclone/rclone/runs/364887180#step:9:629>            pool_test.go:114: 
630
 <https://github.com/rclone/rclone/runs/364887180#step:9:630>                	Error Trace:	pool_test.go:114
631
 <https://github.com/rclone/rclone/runs/364887180#step:9:631>                	            				pool_test.go:218
632
 <https://github.com/rclone/rclone/runs/364887180#step:9:632>                	Error:      	Not equal: 
633
 <https://github.com/rclone/rclone/runs/364887180#step:9:633>                	            	expected: 0
634
 <https://github.com/rclone/rclone/runs/364887180#step:9:634>                	            	actual  : 2
635
 <https://github.com/rclone/rclone/runs/364887180#step:9:635>                	Test:       	TestPool/canFail/Flusher
636
 <https://github.com/rclone/rclone/runs/364887180#step:9:636>FAIL
637
 <https://github.com/rclone/rclone/runs/364887180#step:9:637>FAIL	github.com/rclone/rclone/lib/pool	0.854s
638
 <https://github.com/rclone/rclone/runs/364887180#step:9:638>ok  	github.com/rclone/rclone/lib/random	0.088s
639
 <https://github.com/rclone/rclone/runs/364887180#step:9:639>ok  	github.com/rclone/rclone/lib/readers	0.090s
640
 <https://github.com/rclone/rclone/runs/364887180#step:9:640>ok  	github.com/rclone/rclone/lib/rest	0.088s
641
 <https://github.com/rclone/rclone/runs/364887180#step:9:641>?   	github.com/rclone/rclone/lib/terminal	[no test files]
642
 <https://github.com/rclone/rclone/runs/364887180#step:9:642>ok  	github.com/rclone/rclone/vfs	11.062s
643
 <https://github.com/rclone/rclone/runs/364887180#step:9:643>?   	github.com/rclone/rclone/vfs/test_vfs	[no test files]
644
 <https://github.com/rclone/rclone/runs/364887180#step:9:644>?   	github.com/rclone/rclone/vfs/vfsflags	[no test files]
645
 <https://github.com/rclone/rclone/runs/364887180#step:9:645>FAIL
646
 <https://github.com/rclone/rclone/runs/364887180#step:9:646>make: *** [quicktest] Error 1
647
 <https://github.com/rclone/rclone/runs/364887180#step:9:647>##[error]Process completed with exit code 2.




Doesn't look like something I had anything to do with....

fs/config.go
```diff
@@ -87,6 +87,7 @@ type ConfigInfo struct {
 	StreamingUploadCutoff  SizeSuffix
 	StatsFileNameLength    int
 	AskPassword            bool
+	PasswordCommand        string
```
ncw: I think it would be best to use `SpaceSepList` here instead of string.

The parser will then break it up into space separated parts, but also allow you to use " etc so you can have parameters with spaces in.  It uses the escaping rules from [encoding/csv](https://golang.org/pkg/encoding/csv/)

fs/config/config.go
```diff
@@ -274,6 +267,61 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
 		return goconfig.LoadFromReader(bytes.NewBuffer(b))
 	}
 
+	if len(configKey) == 0 {
+		pwc := fs.Config.PasswordCommand
+
+		if pwc == "" {
```
ncw: We shouldn't need to read the environment variable here - `flags.StringVarP` should have loaded it from the env var already?

Ah, I see the problem, the variable associated with `--password-command` is `RCLONE_PASSWORD_COMMAND` not `RCLONE_CONFIG_PASS_COMMAND`.

So really we should be using `RCLONE_PASSWORD_COMMAND` everywhere.

The tests can then set `fs.Config.PasswordCommand` directly.

fs/config/config_test.go
```diff
@@ -256,6 +256,84 @@ func TestConfigLoadEncrypted(t *testing.T) {
...
 
+func expectConfigValid(t *testing.T) {
+	var err error
+
+	configKey = nil // reset password
+
+	c, err := loadConfigFile()
+	require.NoError(t, err)
+
+	// Not sure why there is no "RCLONE_ENCRYPT_V0" here...
```
ncw: You'll only see that in the encrypted file - it will have been decrypted here

docs/content/docs.md
```diff
@@ -1282,11 +1282,36 @@ Then source the file when you want to use it.  From the shell you
...
+is primarily protected by the `passwordstore` system, and is never
+embedded in the clear in scripts, nor available for examination
+using the standard commands available.  It is quite possible with
+long running rclone sessions for copies of passwords to be innocently
+captured in log files or terminal scroll buffers, etc.  Using the
+script method of supplying the password enhances the security of
+the config password considerably.
+
+If you are running rclone inside a script, unless you are using the
+`RCLONE_CONFIG_PASS_COMMAND` method, you might want to disable
```
ncw: This should say "unless you are using the `--password-command` method" - all rclone flags have associated environment variables and the docs prefer to refer the the flag name.

docs/content/docs.md
```diff
@@ -1282,11 +1282,36 @@ Then source the file when you want to use it.  From the shell you
...
+script method of supplying the password enhances the security of
+the config password considerably.
+
+If you are running rclone inside a script, unless you are using the
+`RCLONE_CONFIG_PASS_COMMAND` method, you might want to disable 
 password prompts. To do that, pass the parameter 
 `--ask-password=false` to rclone. This will make rclone fail instead
 of asking for a password if `RCLONE_CONFIG_PASS` doesn't contain
-a valid password.
+a valid password, and `RCLONE_CONFIG_PASS_COMMAND` has not been supplied.
```
ncw: ..and `--password-command` has not been supplied.

fs/config/config.go
```diff
@@ -274,6 +267,61 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
...
+
+			args := strings.Fields(pwc)
+			cmd := exec.Command(args[0], args[1:]...)
+
+			cmd.Stdout = &stdout
+			cmd.Stderr = &stderr
+
+			if err := cmd.Run(); err != nil {
+				// One does not always get the stderr returned in the wrapped error.
+				fs.Errorf(nil, "Using RCLONE_CONFIG_PASS_COMMAND returned: %v", err)
```
ncw: "Using --password-command returned"

fs/config/config_test.go
```diff
@@ -256,6 +256,84 @@ func TestConfigLoadEncrypted(t *testing.T) {
...
+	oldConfig := fs.Config
+	ConfigPath = "./testdata/encrypted.conf"
+	defer func() {
+		ConfigPath = oldConfigPath
+		configKey = nil // reset password
+		fs.Config = oldConfig
+	}()
+
+	// using fs.Config.PasswordCommand, incorrect password
+	fs.Config.PasswordCommand = "echo asdf-blurfl"
```
ncw: This set fs.Config.PasswordCommand should be moved above so the reset can go in the defer.  If `expectConfigInvalid` panics then it won't reset the `PasswordCommand` otherwise.

permezel:
Hi.  There are two ways to set the config password command.  One is via an explicitly set and exported environment setting.  The other is via the —password-command= argument.  It appears I failed to adequately document this, but I refer to them both as the RCLONE_CONFIG_PASS_COMMAND method.
In all my manual testing to date, I use the “export RCLONE_CONFIG_PASS_COMMAND=‘..’” method.  It is much more convenient. 

If command line arguments are really turned into environment variables, well, I am shocked.  Is that a GO thing, or an rclone thing?

> On 2020-Jan-03, at 00:47, Nick Craig-Wood <notifications@github.com> wrote:
> 
> @ncw commented on this pull request.
> 
> Apologies for the delay responding, the macOS failure is one of those tests that fails occasionally
> 
> I think we are nearly there :-) There is a little confusion about what the environment variable should be called and some docs tips I've pointed out inline.
> 
> Thank you and looking forward to merging this for 1.51 :-)
> 
> In fs/config.go <https://github.com/rclone/rclone/pull/3749#discussion_r362489427>:
> 
> > @@ -87,6 +87,7 @@ type ConfigInfo struct {
>  	StreamingUploadCutoff  SizeSuffix
>  	StatsFileNameLength    int
>  	AskPassword            bool
> +	PasswordCommand        string
> I think it would be best to use SpaceSepList here instead of string.
> 
> The parser will then break it up into space separated parts, but also allow you to use " etc so you can have parameters with spaces in. It uses the escaping rules from encoding/csv <https://golang.org/pkg/encoding/csv/>

Which parser?  I admit that there are some issues one runs into for certain potential commands, but nothing which cannot be solved by wrapping the command in a script and then just invoking the script interpreter with the file.  “..COMMAND=‘bash  ./script.sh’”.
> In fs/config/config.go <https://github.com/rclone/rclone/pull/3749#discussion_r362490138>:
> 
> > @@ -274,6 +267,61 @@ func loadConfigFile() (*goconfig.ConfigFile, error) {
>  		return goconfig.LoadFromReader(bytes.NewBuffer(b))
>  	}
>  
> +	if len(configKey) == 0 {
> +		pwc := fs.Config.PasswordCommand
> +
> +		if pwc == "" {
> We shouldn't need to read the environment variable here - flags.StringVarP should have loaded it from the env var already?
> 
> Ah, I see the problem, the variable associated with --password-command is RCLONE_PASSWORD_COMMAND not RCLONE_CONFIG_PASS_COMMAND.
> 
> So really we should be using RCLONE_PASSWORD_COMMAND everywhere.
> 
> 
No.
> The tests can then set fs.Config.PasswordCommand directly.
> 
> In fs/config/config_test.go <https://github.com/rclone/rclone/pull/3749#discussion_r362491409>:
> 
> > @@ -256,6 +256,84 @@ func TestConfigLoadEncrypted(t *testing.T) {
>  	assert.Equal(t, expect, keys)
>  }
>  
> +func expectConfigValid(t *testing.T) {
> +	var err error
> +
> +	configKey = nil // reset password
> +
> +	c, err := loadConfigFile()
> +	require.NoError(t, err)
> +
> +	// Not sure why there is no "RCLONE_ENCRYPT_V0" here...
> You'll only see that in the encrypted file - it will have been decrypted here
> 
> In docs/content/docs.md <https://github.com/rclone/rclone/pull/3749#discussion_r362494881>:
> 
> > +```
> +export RCLONE_CONFIG_PASS_COMMAND="pass rclone/config"
> +```
> +
> +If the `passwordstore` password manager holds the password for the
> +rclone configuration, using the script method means the password
> +is primarily protected by the `passwordstore` system, and is never
> +embedded in the clear in scripts, nor available for examination
> +using the standard commands available.  It is quite possible with
> +long running rclone sessions for copies of passwords to be innocently
> +captured in log files or terminal scroll buffers, etc.  Using the
> +script method of supplying the password enhances the security of
> +the config password considerably.
> +
> +If you are running rclone inside a script, unless you are using the
> +`RCLONE_CONFIG_PASS_COMMAND` method, you might want to disable 
> This should say "unless you are using the --password-command method" - all rclone flags have associated environment variables and the docs prefer to refer the the flag name.
> 
> 

I think the name of the method can be simplified as I have, and not have to be the
RCLONE_CONFIG_PASS_COMMAND / —password-command method.
But when I sit down to process this again, I will have another think and try and clean it up.


> 
> In docs/content/docs.md <https://github.com/rclone/rclone/pull/3749#discussion_r362495005>:
> 
> >  password prompts. To do that, pass the parameter 
>  `--ask-password=false` to rclone. This will make rclone fail instead
>  of asking for a password if `RCLONE_CONFIG_PASS` doesn't contain
> -a valid password.
> +a valid password, and `RCLONE_CONFIG_PASS_COMMAND` has not been supplied.
> ..and --password-command has not been supplied.
> 
> 
.. either by setting the RCLONE_CONFIG_PASS_COMMAND environment variable or supplying the —password-command argument.

> 
> In fs/config/config.go <https://github.com/rclone/rclone/pull/3749#discussion_r362495532>:
> 
> > +		if pwc == "" {
> +			pwc = os.Getenv("RCLONE_CONFIG_PASS_COMMAND")
> +		}
> +		if pwc != "" {
> +			var stdout bytes.Buffer
> +			var stderr bytes.Buffer
> +
> +			args := strings.Fields(pwc)
> +			cmd := exec.Command(args[0], args[1:]...)
> +
> +			cmd.Stdout = &stdout
> +			cmd.Stderr = &stderr
> +
> +			if err := cmd.Run(); err != nil {
> +				// One does not always get the stderr returned in the wrapped error.
> +				fs.Errorf(nil, "Using RCLONE_CONFIG_PASS_COMMAND returned: %v", err)
> "Using --password-command returned"
> 
> 

Hmm.  Perhaps I need to distinguish both cases so I indicate which one failed.
If you supply both, one takes precedence: we only check the ENV if the command line arg was not supplied.

> In fs/config/config_test.go <https://github.com/rclone/rclone/pull/3749#discussion_r362496473>:
> 
> > +	err = os.Unsetenv("RCLONE_CONFIG_PASS_COMMAND")
> +	require.NoError(t, err)
> +}
> +
> +func TestConfigLoadEncryptedWithInvalidPassCommand(t *testing.T) {
> +	oldConfigPath := ConfigPath
> +	oldConfig := fs.Config
> +	ConfigPath = "./testdata/encrypted.conf"
> +	defer func() {
> +		ConfigPath = oldConfigPath
> +		configKey = nil // reset password
> +		fs.Config = oldConfig
> +	}()
> +
> +	// using fs.Config.PasswordCommand, incorrect password
> +	fs.Config.PasswordCommand = "echo asdf-blurfl"
> This set fs.Config.PasswordCommand should be moved above so the reset can go in the defer. If expectConfigInvalid panics then it won't reset the PasswordCommand otherwise.
> 
> 
Ah.  I think I understand the mechanism used now.  Cool.  Thanks.  I thought it was handled on a function level, not a statement level.

permezel:
> On 3 Jan 2020, at 00:47, Nick Craig-Wood <notifications@github.com> wrote:
> 
> In fs/config/config_test.go <https://github.com/rclone/rclone/pull/3749#discussion_r362496473>:
> 
> > +	err = os.Unsetenv("RCLONE_CONFIG_PASS_COMMAND")
> +	require.NoError(t, err)
> +}
> +
> +func TestConfigLoadEncryptedWithInvalidPassCommand(t *testing.T) {
> +	oldConfigPath := ConfigPath
> +	oldConfig := fs.Config
> +	ConfigPath = "./testdata/encrypted.conf"
> +	defer func() {
> +		ConfigPath = oldConfigPath
> +		configKey = nil // reset password
> +		fs.Config = oldConfig
> +	}()
> +
> +	// using fs.Config.PasswordCommand, incorrect password
> +	fs.Config.PasswordCommand = "echo asdf-blurfl"
> This set fs.Config.PasswordCommand should be moved above so the reset can go in the defer. If expectConfigInvalid panics then it won't reset the PasswordCommand otherwise.
> 
> 

Actually, I am no longer sure I understand this.  Made sense to me last night, but not today.

The fs.Config = oldConfig is supposed to reset it.
Anyway, I deleted the --password-command so it is no longer an issue.

ncw:
> Hi. There are two ways to set the config password command. One is via an explicitly set and exported environment setting. The other is via the —password-command= argument. It appears I failed to adequately document this, but I refer to them both as the RCLONE_CONFIG_PASS_COMMAND method. In all my manual testing to date, I use the “export RCLONE_CONFIG_PASS_COMMAND=‘..’” method. It is much more convenient. If command line arguments are really turned into environment variables, well, I am shocked. Is that a GO thing, or an 
rclone thing?

It is an rclone thing - see https://rclone.org/docs/#environment-variables .  So there is no need for you to read an environment variable explicitly in the code as `RCLONE_PASSWORD_COMMAND` if set will do the same as setting `--password-command`.

You can still use `RCLONE_PASSWORD_COMMAND` which will make it consistent with the rest of rclone.

> In fs/config.go <[#3749 (comment)](https://github.com/rclone/rclone/pull/3749#discussion_r362489427)>: > @@ -87,6 +87,7 @@ type ConfigInfo struct { StreamingUploadCutoff SizeSuffix StatsFileNameLength int AskPassword bool + PasswordCommand string I think it would be best to use SpaceSepList here instead of string. The parser will then break it up into space separated parts, but also allow you to use " etc so you can have parameters with spaces in. It uses the escaping rules from encoding/csv <https://golang.org/pkg/encoding/csv/>
> Which parser? I admit that there are some issues one runs into for certain potential commands, but nothing which cannot be solved by wrapping the command in a script and then just invoking the script interpreter with the file. “..COMMAND=‘bash ./script.sh’”.

Rclone has the mechanism already, I suggest we use it :-)  It would let you use `--pasword-command 'echo "hello world"'` for example.

> > So really we should be using RCLONE_PASSWORD_COMMAND everywhere.
> No.

I beg to disagree! (See above)

> Actually, I am no longer sure I understand this. Made sense to me last night, but not today. The fs.Config = oldConfig is supposed to reset it.

`fs.Config = oldConfig` will reset it, but not if there is a panic() before it gets there.  Putting it in the defer statement means it will be executed as part of a panic unwinding the stack.

It is go's version of RAII

```go
func() {
    obj := createTheObj()
   defer obj.Destroy()
}
```

The defer statement is executed after the function returns

> Anyway, I deleted the --password-command so it is no longer an issue.

Why? I think having both a flag (and an automatically read environment variable) is the best solution.  That will make it 100% consistent with the rest of rclone.

I don't want to just have an environment variable - everything (with the exception of the config password) is configurable via both the command line and an environment variable.

permezel:
I read the docs on env vars.  It states that for any cli option, there is an env var, and a mapping from cli option to env var.  It does not suggest that for any env var there is a --cli-option.  It then goes on to discuss "other env vars".

RCLONE_CONFIG_PASS_COMMAND is one of these "other" env vars.  Quite similar in fact to "RCLONE_CONFIG_PASS".
I do not find any reference to "--config-pass" in my current source.

Cheers,
=dap

ncw:

I'd like there to be a `--password-command` environment variable, but what I can do is merge what you've done and add it afterwards - how does that sound?

Thanks

permezel:
The current version only has the env var.
It is named RCLONE_CONFIG_PASSWORD_CMD  to avoid confusion with all the other passwords possible and to make it explicit as well as to match the existing env var.

You are welcome to do whatever you like with it.  I have found rclone to be much too flakey for my use.
Can’t remember the precise details now.  One issue was discovered when I was trying to determine why a sync (about 40GB to NAS and to Samsung fast removable SSD, I make two copies) took forever when mostly nothing changed.  I turned on -vvv and discovered that the time delta was deciding the files were off by about 1/2 second and copying them over and over.

If it cannot set the time reliably, what use is it?

The other had to do with truncated files being sent to S3 bucket.  Don’t really remember details, but I sync a NAS-resident Borg backup to S3 bucket encrypted by rclone.  Should be a simple issue.  No errors reported.  These issues were with the version in the Debian buster repo, not one I built.

ncw:
> The current version only has the env var. It is named RCLONE_CONFIG_PASSWORD_CMD to avoid confusion with all the other passwords possible and to make it explicit as well as to match the existing env var. You are welcome to do whatever you like with it.

I'll merge what you've done with a few fixes from me - thank you very much for contributing that.

> I have found rclone to be much too flakey for my use. Can’t remember the precise details now. One issue was discovered when I was trying to determine why a sync (about 40GB to NAS and to Samsung fast removable SSD, I make two copies) took forever when mostly nothing changed. I turned on -vvv and discovered that the time delta was deciding the files were off by about 1/2 second and copying them over and over. If it cannot set the time reliably, what use is it?

That is probably to do with the file system you were using. Rclone expects linux file systems to be able to set nS precision but for instance vfat only has 1 second precision.  You can use the `--modify-window` flag to work around.

> The other had to do with truncated files being sent to S3 bucket. Don’t really remember details, but I sync a NAS-resident Borg backup to S3 bucket encrypted by rclone. Should be a simple issue. No errors reported. These issues were with the version in the Debian buster repo, not one I built.

I'm sorry to hear that. The version of rclone in buster is 1.45 which is quite old now alas. If you wanted to try again I'd recommend trying the release version 1.50.2.

Thank you for your help

ncw:
All merged now - thank you for your help :-)

PR #3771, opened 2019-11-28
backend/local: add disable_checksum option

ajankovic:
Rclone will calculate checksums on read for all objects in local
backend. This is a problem in performance critical environments or for
large files.

We are introducing `disable_checksum option for the local backend to
disable hashing for file reads.

#### Checklist

- [] I have added documentation for the changes if appropriate.

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2040,7 +2040,7 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 	// read the md5sum if available for non multpart and if
 	// disable checksum isn't present.
 	var md5sum string
-	if !multipart || !o.fs.opt.DisableChecksum {
+	if !multipart && !o.fs.opt.DisableChecksum {
 		hash, err := src.Hash(ctx, hash.MD5)
 		if err == nil && matchMd5.MatchString(hash) {
 			hashBytes, err := hex.DecodeString(hash)


ncw:
> Rclone will calculate checksums on read for all objects in local
> backend. This is a problem in performance critical environments or for
> large files

Which rclone operation do you notice this in? sync/copy/mount to from which backend?

I've been trying to track down operations where rclone uses **all** the hashes - whirlpool is particularly expensive - might it be that is what you are seeing?

If you try [the latest beta](https://beta.rclone.org) do you notice the same problems?  This has be5392f4488d7e994f2b2a44794642eb9144605f in which speeds up mount.

Disabling the local hashes for the pursuit of speed stops rclone being able to integrity check transfers which I'm not wildly keen on!

ajankovic:
@ncw  You are right this pushed me to investigate more. It turns out hashing was performed by the `s3` backend. I've modified PR. Please check if this condition is correct because I think it's not.

https://github.com/rclone/rclone/blob/master/backend/s3/s3.go#L2043

Comment also says `"non multpart and if disable checksum"` but there is an `||` between those two.

ivandeex:
> Which rclone operation do you notice this in? sync/copy/mount to from which backend?

In fact, local backend will calculate any rclone-supported hash requested by current operation via fs.Hash API. I stumbled upon this while testing mailru backend, which internally requests file hash to attempt body-less "upload by hash". This resulted in long unexpected pre-copy delays on large local files. Since we have no standard API to ask source backend, whether fs.Hash is resource-expensive or not, this will waste cpu in vain if source file is 99.99% unique (eg part of ongoing restic backup or just encrypted).

> Disabling the local hashes for the pursuit of speed stops rclone being able to integrity check transfers which I'm not wildly keen on!

To solve above use case, I added a knob to mailru backend that disables pre-copy source hash requests reverting to on-the-fly hasing. It fixed pre-copy delays in the "sync large restic backups to mailru" use case. Having similar knob on the localfs side looks ok to me. It's up to users to use it or not.

ncw:
> @ncw You are right this pushed me to investigate more. It turns out hashing was performed by the `s3` backend.
> I've modified PR. Please check if this condition is correct because I think it's not.
> 
> https://github.com/rclone/rclone/blob/master/backend/s3/s3.go#L2043
> 
> Comment also says `"non multpart and if disable checksum"` but there is an `||` between those two.

Urggh, I think you are right. I'll merge the updated fix - thanks.

Issue #3752, opened 2019-11-21
asyncread: WriteTo EOF


buengese:
While debugging some stuff related to my mirror backend I noticed an inexplicable EOF error from io.Copy(). Looking into it this appears to be caused by rclone's asyncreader.WriteTo() method. The documentation for io.Copy() states that a successful copy should be indicated by err == nil and not err == EOF. When WriteTo() is available io.Copy() directly propagates the error from WriteTo(). So I don't think asyncreader.WriteTo() should return an EOF error.

@ncw

ncw:
Well spotted.

Want to send a fix?

buengese:
Sure

ncw:
I found [the official docs](https://golang.org/pkg/io/#WriterTo)

> WriterTo is the interface that wraps the WriteTo method.
>
> WriteTo writes data to w until there's no more data to write or when an error occurs. The return value n is the number of bytes written. Any error encountered during the write is also returned.

I think `until there's no more data to write` is a strong indicator that it should never return io.EOF.

ncw:
This is fixed now - thank you @buengese :-)

PR #3787, opened 2019-12-05
asyncreader: fix EOF error

buengese:
Quick fix for #3752.

### DIFF
--- a/fs/asyncreader/asyncreader.go
+++ b/fs/asyncreader/asyncreader.go
@@ -174,6 +174,9 @@ func (a *AsyncReader) WriteTo(w io.Writer) (n int64, err error) {
 	n = 0
 	for {
 		err = a.fill()
+		if err == io.EOF {
+			return n, nil
+		}
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ncw:
If you rebase on master it should fix the build by picking up cb97239a608eb64d101c324e747918a5d882866f

Can you do a test?

buengese:
Sure.

buengese:
This should be correct in all cases no. I did some additional testing comparing it to bufio.Reader and appears to be correct now. The double EOF just to ignore it isn't pretty but i couldn't think of a better way to do it without either breaking something or making some bigger changes to asyncreader.

ncw:
> This should be correct in all cases no. I did some additional testing comparing it to bufio.Reader and appears to be correct now. The double EOF just to ignore it isn't pretty but i couldn't think of a better way to do it without either breaking something or making some bigger changes to asyncreader.

That patch looks correct to me, and I think the minimal change is probably the best one.

I'll merge that now - thank you :-)

PR #3812, opened 2019-12-18
cmd: add --error-on-no-transfer option

jfautley:

This allows rclone to exit with a non-zero return code if no files are
transferred. This is useful when calling rclone as part of a workflow/script
pipeline as it allows the end user to stop processing if no files have been
transferred.

NB: Enabling this option will return in rclone exiting non-zero if there are no
transfers. Depending on how you're currently using rclone in your scripts,
this may break existing setups!

##### Example usage:
```
rclone --error-on-no-transfer copy remote:/myfiles s3:my-s3-bucket/files/from/me
RC=$?
if [ $RC -eq 0 ]; then
  echo "I'm going to process some files now"
elif [ $RC -eq 9 ]; then
  echo "There's nothing to do, goodbye"
  exit 0
else
  echo "Something went wrong in the transfer"
  exit 1
fi
```

No, but it probably should be... I've sort of stolen the `--stats-*` namespace for the command option as it seemed the most logical place to put it. I've also added a new exit code (the next sequential/available one in the list of current codes) to reflect "all good, but no files transferred".

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -66,6 +66,7 @@ const (
 	exitCodeNoRetryError
 	exitCodeFatalError
 	exitCodeTransferExceeded
+	exitCodeNoFilesTransferred
 )
 
 // ShowVersion prints the version to stdout
@@ -312,6 +313,7 @@ func Run(Retry bool, showStats bool, cmd *cobra.Command, f func() error) {
 		}
 	}
 	resolveExitCode(cmdErr)
+
 }
 
 // CheckArgs checks there are enough arguments and prints a message if not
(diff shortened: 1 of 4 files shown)


ivandeex:
I am sorry for accidentally closing this.

ncw:
Apologies for the delay looking at this, I'm a bit behind with the pull requests.

I think the idea is sound and useful.

I don't like the name of the flag though and I think the logic should be inverted so rclone exits with a non-zero error if no files were transferred.

How about using `--error-on-no-transfers` or something like that?

jfautley:
Makes sense, leave this with me and I'll take a look at refactoring in the next few days 😄

ncw:
> Makes sense, leave this with me and I'll take a look at refactoring in the next few days

:-)

jfautley:
OK, so it was a _little_ more than a few days, but what's a month or two between friends? 🤡

ncw:
That looks super now thank you.

I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.52

PR #3828, opened 2019-12-26
s3: introduce max_keys option for bucket listing

Kriechi:

The S3 ListObject API returns paginated bucket listings, with "max_keys" items for each GET call.
The default value is 1000 entries, but for buckets with millions of objects it might make sense to request more elements per request. This commit adds a "max_keys" option for the user to specify a higher (or lower) number of entries per request.

This commit does not add safe guards around this value - if a user decides to request a too large list, it might result in connection timeouts (on the server or client).

No.

#### Checklist

I ran `make backenddocs` which seems to have added a few more things - just let me know if I should remove them from this PR or keep them to have a fresh auto-generated documentation.


Kriechi:
The test failure on macOS seems unrelated to the changes in this PR: `TestJobsKickExpire`

ncw:
I'm under the impression that 1000 is the max number available in a listing. If you set more than 1000 that Amazon silently truncates it to 1000?

We called this parameter list-chunk in drive so we should call it the same think here.

https://rclone.org/drive/#drive-list-chunk

Kriechi:
@ncw I considered different names for this option as well - I then settled on the same name used by AWS in the official docs for AWS S3 - which to me made the most sense.
They also use the terms `max-items` and `page-size`.

https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html

I couldn't find anything regarding your proposed "truncation after 1000" in https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html.
Although, as you suggested, this page mentioned 1000 as maximum: https://docs.aws.amazon.com/cli/latest/reference/s3/ls.html

I haven't tested it against actual AWS S3, but at least Ceph RadosGW happily responds with 10.000 or 100.000 keys per request.

I'm open to renaming it - but I'd like to have the option of specifying lower/higher values, than the default 1000, if needed.

ncw:
> I considered different names for this option as well - I then settled on the same name used by AWS in the official docs for AWS S3 - which to me made the most sense.
> They also use the terms `max-items` and `page-size`.

I try to keep the names consistent between the backends, so I'd prefer something to match

     --azureblob-list-chunk int   Size of blob list. (default 5000)
      --drive-list-chunk int       Size of listing chunk 100-1000. 0 to disable. (default 1000)

I'm pretty sure AWS S3 truncates to 1,000 as I used to use 1,024 which caused a problem with wasabi 2944f7603d99aa417764c67b4061374979bf208c

Perhaps you could put a bit more help in the help text that this is truncated at 1000 by AWS and may cause errors with Wasabi, but works when set larger than 1000 with CEPH.

Kriechi:
Thanks @ncw, this makes sense - I updated the PR accordingly.

ncw:
Thanks for that - I'll merge it now :-)

I normally only regenerate the docs at the release, so I dropped the non relevant docs bits from the commit.

Thank you :-)

PR #3833, opened 2019-12-30
RFC: s3: add more progress indication during bucket listing

Kriechi:

Long-running commands, e.g., copy/sync on S3 backends are currently not providing any feedback to the user during the initial bucket listing. If the source bucket has e.g., 10 million objects, and each ListObjects S3 API call takes 500ms, with 1000 objects per response, we are potentially sitting and waiting for 1.5 hours without any progress or activity indication. 

Is it stuck? Do I have the right credentials? Is rclone waiting for I/O or CPU or disk?

Since the S3 API does not provide an easy way of accessing the total number of files up front, we have to simply list all them - so we can only display the number of already listed (received) files.

This at least provides some kind of metric and activity indicator to the user. This value is expected to increase every few seconds, every time the ListObjects returns another 1000 items (`MaxKeys`/`max-items`/`page-size`/`list-chunk`).

Please consider this PR as a proposal or request-for-comments. It solved an immediate need for myself - but I'm not sure if the current implementation fits into the bigger picture around statistics collections and UI/UX.

I will update documentation, commit message, etc., after agreeing on the implementation.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -43,6 +43,7 @@ import (
 	"github.com/ncw/swift"
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/accounting"
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/encodings"
@@ -1248,6 +1249,8 @@ type listFn func(remote string, object *s3.Object, isDirectory bool) error
 //
 // Set recurse to read sub directories
 func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBucket bool, recurse bool, fn listFn) error {
+	accounting.Stats(ctx).ListedFiles(f.Name(), 0)
+
 	if prefix != "" {
 		prefix += "/"
 	}
@@ -1357,6 +1360,7 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 					return err
 				}
 			}
+			accounting.Stats(ctx).ListedFiles(f.Name(), int64(len(resp.CommonPrefixes)))
 		}
 		for _, object := range resp.Contents {
 			remote := aws.StringValue(object.Key)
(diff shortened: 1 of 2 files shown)


ncw:

I like the idea of a "listing files" stat of some kind.

However I'm not that keen on adding stats counting into every backend.

Currently there are two listing interfaces

One which looks like this `ListR`

```go
// ListRCallback defines a callback function for ListR to use
//
// It is called for each tranche of entries read from the listing and
// if it returns an error, the listing stops.
type ListRCallback func(entries DirEntries) error

// ListRFn is defines the call used to recursively list a directory
type ListRFn func(ctx context.Context, dir string, callback ListRCallback) error
```

And the other `List` which looks like

```
        // List the objects and directories in dir into entries.  The
        // entries can be returned in any order but should be for a
        // complete directory.
        //
        // dir should be "" to list the root, and should not have
        // trailing slashes.
        //
        // This should return ErrDirNotFound if the directory isn't
        // found.
        List(ctx context.Context, dir string) (entries DirEntries, err error)
```

I think the ideal solution to this would be to change the List interface into something more like the ListR interface.

It would then be easy to create a shim function with worked the same as List and counted stats.

Changing the List interface for **every** backend is a lot of work though which is why I haven't done it previously (though it has occurred to me!).

It might be that having an optional interface say `ListC` for `List` with callback might be the way to go.  No all backends are capable of "paging" a listing of a single directory.

What do you think?

swimfish09:
rclone ListObject all the time, and memory usage became very high. Will it list all the objects and then upload?

ivandeex:
@Kriechi 
Kindly address the review comments

Kriechi:
Unfortunately I currently cannot continue on this PR at the moment - feel free to close.

ivandeex:
@ncw ?

ncw:
I like the goal, but I think the implementation would be better done by adding an additional ListP (list paged) interface.

I've written this up in a new issue: https://github.com/rclone/rclone/issues/4788

I'm going to close this issue as I don't think it is the right direction to go in as it isn't general purpose enough, sorry @Kriechi 

However I do like the idea and would appreciate help with #4788

Thank you

Issue #3753, opened 2019-11-21
[AWS] Support assume role via AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN environment variables when running on EKS


jlerche:
#### What is your current rclone version (output from `rclone version`)?
```
# ./rclone version
rclone v1.50.2
- os/arch: linux/amd64
- go version: go1.13.4
```

#### What problem are you are trying to solve?
We run workloads on AWS EKS using [IAM roles for (kubernetes) service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). When a kubernetes pod runs under a (kubernetes) service account with this feature, two environment variables get populated automatically:
```
# env | grep AWS
AWS_ROLE_ARN=<snip>
AWS_WEB_IDENTITY_TOKEN_FILE=<snip>
```
which can then be used to authenticate.

#### How do you think rclone should be changed to solve that?
As of https://github.com/rclone/rclone/commit/479c803fd9034a38d01be0aa8c55a8881b01406c#diff-37aff102a57d3d7b797f152915a6dc16 the aws sdk in `go.mod` supports this feature (we use it in our own code). It would be great if `rclone` could support it as well. With this feature supported, I would expect a minimal config like:
```
[remote]
type = s3
provider = AWS
env_auth = true
region = us-west-2
```
to work, assuming the two environment variables listed above are populated with valid values.

In principle, creating a provider via [`stscreds.NewWebIdentityRoleProvider`](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/#NewWebIdentityRoleProvider) and adding it to the [credentials provider chain here](https://github.com/rclone/rclone/blob/master/backend/s3/s3.go#L954) should work.

ncw:
Would you like to send a pull request for this?

I recently updated the SDK in master so it should just be a matter of adding that?

ncw:
This is fixed now so will close this issue.

PR #3837, opened 2019-12-31
use aws web identity role provider

tennix:

Support EKS iam role.

#3753 

#### Checklist

I've tested this manually in a kops cluster with kube2iam deployed.

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -987,7 +987,7 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 			ExpiryWindow: 3 * time.Minute,
 		},
 
-		// Pick up IAM role if we are in eks
+		// Pick up IAM role if we are in EKS
 		&stscreds.WebIdentityRoleProvider{
 			ExpiryWindow: 3 * time.Minute,
 		},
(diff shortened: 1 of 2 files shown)


ncw:
That looks very useful - thanks :-)

Can you update the auth docs here too please: https://github.com/rclone/rclone/blob/master/docs/content/s3.md#authentication

(All the other docs are auto generated from that one file)

tennix:
@ncw I've updated the document. The CI fails, but I don't think this is related to my changes :slightly_smiling_face:

ncw:
That looks great now - thanks!

Yes the CI is broken for some unknown reason :-(

backend/s3/s3.go
```diff
@@ -985,6 +986,11 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 			}),
 			ExpiryWindow: 3 * time.Minute,
 		},
+
+		// Pick up IAM role if we are in EKS
+		&stscreds.WebIdentityRoleProvider{
```
tirsen: Does creating this struct this way really work? It has quite a lot of non-public fields that are set when calling `NewWebIdentityRoleProvider`. As you can see that state is being used in the `Retrieve` function: https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/web_identity_provider.go#L67

ncw: @tirsen - is this causing you a problem?

Want to send a PR to fix?

tirsen: Yeah EKS role authentication doesn't work. I've checked all the env vars of the pod but everything seems to be in good order. Tracing through the code paths of `WebIdentityRoleProvider` I can see that the env vars will not be read if that struct is created that way.

Here is a fix:
https://github.com/rclone/rclone/pull/4023

PR #3856, opened 2020-01-05
operations: write debug message when hashes could not be checked

oschuett:
Adds a simple debug message that helps to track down cases where the hashes could not be checked.


ncw:
This will likely print two log message per file won't it?

    file: OK
    file: Hash could not be checked

Perhaps the OK message should be "OK", or "OK - couldn't check hash"?

oschuett:
Alright, I merged the two messages as you suggested.

ncw:
Thanks that looks great :-)

I'll merge it now.

PR #3887, opened 2020-01-18
backend/onedrive: add support "Retry-After" header

vmi:

Add support "Retry-After" header to backend/onedrive.

No.

#### Checklist

### DIFF
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"net/http"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


backend/onedrive/onedrive.go
```diff
@@ -380,13 +381,31 @@ var retryErrorCodes = []int{
 // shouldRetry returns a boolean as to whether this resp and err
 // deserve to be retried.  It returns the err as a convenience
 func shouldRetry(resp *http.Response, err error) (bool, error) {
-	authRetry := false
-
-	if resp != nil && resp.StatusCode == 401 && len(resp.Header["Www-Authenticate"]) == 1 && strings.Index(resp.Header["Www-Authenticate"][0], "expired_token") >= 0 {
-		authRetry = true
-		fs.Debugf(nil, "Should retry: %v", err)
+	if err == nil {
```
Cnly: It seems the function now just returns `false, nil` when `err == nil`, but the original version checks the status code and headers even in this case. Is this change intended?

ncw: There are 3 possible cases

1. `err == nil, `resp.StatusCode == 200` - everything OK
2. `err == nil, `resp.StatusCode != 200` - everything NOT OK - need to check StatusCode
3. `err != nil, `resp.StatusCode undefined` - everything NOT OK - need to check err whether we should retry it

Returning if `err == nil` misses out case 2 (which is the Retry-After case).

The "normal" case will be `err == nil` but some non 200 error in `*http.Response` so

vmi: Thanks comments. I pushed the fix again.

ncw:
I think that looks OK to me - how about you @Cnly ?

ncw:
I'll merge this now - thank you very much :-)

PR #3888, opened 2020-01-18
Password input fix

Ajaja:
Turning off ENABLE_LINE_INPUT disables "Enter" key. Only way to enter password is Ctrl-Z. That is weird.


Ajaja:
It was turned off in https://github.com/rclone/rclone/commit/251cfc100eb1240b7d90cd463be26a2de5476fae

ncw:
It looks like this has been reported in https://github.com/golang/go/issues/36609

I can't take a patch to the vendor directory, so I'll revert the problem commit and await a fix for that issue from upstream.

Thanks for reporting

ncw:
I reverted the commit in 18d26e2ddbab0f010a21ffa84beb00bc11526576

PR #3899, opened 2020-01-23
log: change cache invalidation to an info level

beyondmeat:

This is useful for users of plex_autoscan and plex_rcs.
log-level debug can create a very large log file, very quickly, 1G in less than an hour.

No, but not needed IMO.

#### Checklist

### DIFF
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -113,7 +113,7 @@ func (d *Dir) forgetDirPath(relativePath string) {
 	if dir := d.cachedDir(relativePath); dir != nil {
 		dir.walk(func(dir *Dir) {
 			// this is called with the mutex held
-			fs.Debugf(dir.path, "forgetting directory cache")
+			fs.Infof(dir.path, "forgetting directory cache")
 			dir.read = time.Time{}
 			dir.items = make(map[string]Node)
 		})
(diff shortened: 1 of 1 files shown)


beyondmeat:
I tried a few scripts that watch rclone logs so when files are uploaded it sends a scan request to Plex.

Eventually these scripts all crash because the logfile is constantly changing and eventually it gets a bad read, throws an error about invalid encoding. The second problem is the log file will be several gigs within an hour of truncating it due to all the debug information being written, especially during a Plex scan, it could be several gigs in minutes.

So my intention here is to get this useful info without needing all the debug istuff too. It is my understanding that the cache backed will output similar info when files are uploaded, and it outputs at the info level. So my thought here was to match the similar messages to output as Info for vfs.

darthShadow:
Can this be done as a flag, something like `--log-invalidations`?
This won't be useful for most users of rclone and will add unwanted cruft to the log files.

BTW, not sure if you are aware of it but Plex Autoscan already has an alternative to scanning log files, in the form of subscribing to change notifications itself from Google which is a much better solution to this problem.

Ideally, in the future, we could have an event channel in rclone which other programs could subscribe to and receive such events.

beyondmeat:
Your proposals sound like a better idea. I'll check out autoscan again, thank you

PR #3903, opened 2020-01-26
stats: Allow printing unscaled statistics

kir4h:

This PR introduces the `--stats-unscaled` flag which will skip scaling units when printing stats (no conversion to k|M|G|P|T). A possible use case is being able to generate a metric directly from parsing the log, that can be then consumed later on by other systems (i.e: grafana)

No, this is a minor change so according to the contribution guidelines I think no discussion is needed.

#### Checklist

~~Regarding the test, the `String` method for `StatsInfo` isn't tested (I was hoping to just edit current test to cover the new code). Not very confident in writing it, but if you point me to some example I can work on it (or maybe doing this in a separate commit/PR).~~

I have tried to do some testing on the String method. I'm not a big fan of testing the way I have done (every extra space will fail the test), so let me know if I should drop the commit (or please point me to the right direction to improve it =) )

Feel free to add any suggestion to the PR, any is welcome. 

PS: Just realized my commit should be `accounting: <text>` instead of `stats: <text>`. But on the other hand stats sounds more meaningful to the change. Should I rename it to `accounting`?

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -378,7 +378,7 @@ An example of timetable with WEEKDAY could be:
 `--bwlimit "Mon-00:00,512 Fri-23:59,10M Sat-10:00,1M Sun-20:00,off"`
 
 It mean that, the transfer bandwidth will be set to 512kBytes/sec on Monday.
-It will raise to 10Mbytes/s before the end of Friday. 
+It will raise to 10Mbytes/s before the end of Friday.
 At 10:00 on Sunday it will be set to 1Mbyte/s.
 From 20:00 at Sunday will be unlimited.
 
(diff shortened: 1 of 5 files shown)


ncw:
If this is for consumption programatically, then wouldn't it be easier to use `--use-json-log`? A bit of extra code could then put the stats counters into the json log as JSON objects.

I think that would be a more useful direction to go in rather than adjusting what is supposed to be human readable output.

What do you think?

kir4h:
Let me describe my current scenario:
* A cronjob running rclone for backups
* A [simple app](https://github.com/hoffie/multilog_exporter) tailing rclone's log file, and looking for a regexp. Once a given regexp is matched, a metric is issued and served under and endpoint
* Prometheus scraping the endpoint
* Grafana consuming the metric for visualization purposes

So in my scenario, having the ability to issue the size in bytes would be enough (by that I mean that there might also be a use case for the `no json` log to be this way; I can consume the same log as a human (bytes being harder to read than scaled uinits of course) and with such parser).

But on the other hand I agree that working on the json side is definitely the way to go, it's a much cleaner design than parsing strings with regex on the consumer side. With this PR, the json output would still output bytes, but everything is wrapped under the `msg` field, which totally defeats the advantages of json.

I propose to leave current PR as an option (in some scenarios outputting raw bytes in the human log might come handy, and given the huge number of flags I take it `rclone` aims for flexibility), but open a new ticket to rework this on the JSON side.

Taking a look at the original json PR, I see @wanderleihuttel was discussing that option and [you proposed to issue a new ticket for discussion](https://github.com/rclone/rclone/pull/3373#issuecomment-523395333), but I guess it was never created. 

I am willing to work on that issue if we create it, but we should agree on the specs when it comes to which fields should be present in json output (plus I have used rclone in a very basic way, so I am not yet fully familiar with all the concepts I found yesterday while writing the tests).

ncw:
I have merged #3968 which provides unscaled stats in the "stats" parameter when using `--use-json-log`

I think this provides a cleaner interface to the required data, so I'm going to close this one in favour of that.

Issue #3909, opened 2020-01-27
Use rolling average to improve ETA accuracy


sjpotter:
#### What is the problem you are having with rclone?

```
2020-01-27 17:55:29 NOTICE: Bandwidth limit set to 60M
Transferred:        1.503T / 1.544 TBytes, 97%, 33.027 MBytes/s, ETA 21m46s
Errors:                 0
Checks:                57 / 57, 100%
Transferred:           57 / 59, 97%
Elapsed time:  13h15m17.2s
Transferring:
 *            myfile 52% /87.671G, 3.922M/s, 3h2m17s
```

note how the "myfile" eta is correct, but the top ETA isn't.  This is becaues I was transferring all files at once to maximize my bandwidth, but when the last largest file is left, it calculates ETA by the same transferred speed on top.  Which is the average speed, which is useful to know, but isn't useful for calculating ETA.  At worst, ETA should be based on the current speed, which is easy to eyeball with one file left, but is hard to eyeball when one has many files going.  It be nice to have a "Realtime row" under Transfered with real time speed and a closer to real ETA moved to there from transferred.  should be a matter of just totalling the individual files speeds and calculating ETA with that

#### What is your rclone version (output from `rclone version`)

rclone v1.47.0
- os/arch: linux/amd64
- go version: go1.12.

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

ubuntu 19.10

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone move

animosity22:
Can you update your version and validate you are seeing the same thing? There have been a number of changes to that.

sjpotter:
same behavior with the beta (haven't tested with today's release)

animosity22:
@ncw - is this something you want to take a look at? I believe it's because of how we read the backlog in and not sure what would change here if anything.

ncw:
This is something it would be nice to fix I think.

We need to calculate a rolling average aggregate speed the same as we do for each individual transfer, and use that for the ETA calculations, rather than the long term average.

sjpotter:
that's what my PR does, just sums the rolling average of the individual transfers.  probably not 100% accurate, but closer than the existing.

ncw:
> that's what my PR does, just sums the rolling average of the individual transfers. probably not 100% accurate, but closer than the existing.

[...]

PR #3929, opened 2020-02-03
compute total running average speed and use that for ETA and display

sjpotter:

rclone only prints the average speed from the beginning of the connection which can be very different for long transfers over time from the actual speed one is getting at that moment.

This is easily seen by the ETA calculation being wrong in those cases.

i.e. if someone can get 2 MB/s per transfer, but can do 40 transfers in parallel to get 80MB/s they will see that as their average if its a very long transfer, but towards the end when only a few files are remaining, their actual transfer rate will be much lower, but ETA will still be calculated based on average from the beginning of time

use inProgress set to calculate the current transfer rate and use that for ETA as wel as printing

sort of: https://github.com/rclone/rclone/issues/3909

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -239,6 +239,20 @@ func percent(a int64, b int64) string {
 	return fmt.Sprintf("%d%%", int(float64(a)*100/float64(b)+0.5))
 }
 
+// runningAvgSpeed retuns the total running average speed of all in progress transfers
+func runningAvgSpeed(ip *inProgress) float64 {
+	ip.mu.Lock()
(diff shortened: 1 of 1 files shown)


ivandeex:
Hi! Do you need a hand rebasing this on master?

ivandeex:
I suspect this PR solves diferently the same problem as
#4329 (accounting: use **rolling** average for speed display)

ivandeex:
@sjpotter Are you still interested in getting this merged?

sjpotter:
the other pr can probably go ahead.

Issue #3928, opened 2020-02-03
Google Drive list starred files and folders


jmacul2:
#### What is your current rclone version (output from `rclone version`)?
Latest Version

#### What problem are you are trying to solve?
List files that are starred for quicker access to important items

#### How do you think rclone should be changed to solve that?
Add simple flag "--dive-starred-only" ?

Dibbyo456:
I believe it's possible to implement since `files.get()` does return something like this

```
{
 "kind": "drive#file",
 ...
 "labels": {
  "starred": false,
  "hidden": false,
  "trashed": false,
  "restricted": false,
  "viewed": false,
  "modified": true
},
```

jmacul2:
In the Google Drive api, its pretty simple with the q search param:  `q=starred=true`

https://developers.google.com/drive/api/v3/search-files

jmacul2:
My first attempt at go....hacking something together.  It appears to work.  It would really be awesome if somehow this could work for rclone mount....but I realize that might not be an option.

06019f0fbe19a984dfa891c757ff5c84b5082347

PR #3936, opened 2020-02-05
Added starred files only support for Google Drive backend #3928

jmacul2:
What is the purpose of this change?
Allowing start files only to display on the Google drive backend.

Was the change discussed in an issue or in the forum before?
Not that I know of....now referenced by #3928

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -679,13 +679,8 @@ func (f *Fs) list(ctx context.Context, dirIDs []string, title string, directorie
 			q = fmt.Sprintf("(mimeType='%s' or %s)", driveFolderType, q)
 		}
 		query = append(query, q)
-
-		s := "starred=" + strconv.FormatBool(f.opt.StarredOnly)
-		if f.opt.StarredOnly {
-			s = fmt.Sprintf("(mimeType='%s' or %s)", driveFolderType, s)
-		}
-		query = append(query, s)
 	}
+
 	// Search with sharedWithMe will always return things listed in "Shared With Me" (without any parents)
 	// We must not filter with parent when we try list "ROOT" with drive-shared-with-me
 	// If we need to list file inside those shared folders, we must search it without sharedWithMe
@@ -697,8 +692,13 @@ func (f *Fs) list(ctx context.Context, dirIDs []string, title string, directorie
 		if parentsQuery.Len() > 1 {
 			_, _ = parentsQuery.WriteString(" or ")
 		}
-		if f.opt.SharedWithMe && dirID == f.rootFolderID {
-			_, _ = parentsQuery.WriteString("sharedWithMe=true")
+		if dirID == f.rootFolderID {
+			if f.opt.SharedWithMe {
+				_, _ = parentsQuery.WriteString("sharedWithMe=true")
+			}
+			if f.opt.StarredOnly {
+				_, _ = parentsQuery.WriteString("starred=true")
+			}
 		} else {
 			_, _ = fmt.Fprintf(parentsQuery, "'%s' in parents", dirID)
 		}
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -283,6 +283,12 @@ commands (copy, sync, etc), and with all other commands too.`,
 			Help:     "Only show files that are in the trash.\nThis will show trashed files in their original directory structure.",
 			Advanced: true,
 		}, {
+		}, {
+			Name:     "starred_only",
+			Default:  false,
+			Help:     "Only show files that are starred.",
+			Advanced: true,
+		}, {
 			Name:     "formats",
 			Default:  "",
 			Help:     "Deprecated: see export_formats",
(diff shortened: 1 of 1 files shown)


backend/drive/drive.go
```diff
@@ -282,6 +282,12 @@ commands (copy, sync, etc), and with all other commands too.`,
 			Default:  false,
 			Help:     "Only show files that are in the trash.\nThis will show trashed files in their original directory structure.",
 			Advanced: true,
+		}, {
```
golangcibot: File is not `goimports`-ed (from `goimports`)

backend/drive/drive.go
```diff
@@ -684,8 +691,16 @@ func (f *Fs) list(ctx context.Context, dirIDs []string, title string, directorie
 		if parentsQuery.Len() > 1 {
 			_, _ = parentsQuery.WriteString(" or ")
 		}
-		if f.opt.SharedWithMe && dirID == f.rootFolderID {
-			_, _ = parentsQuery.WriteString("sharedWithMe=true")
+		if dirID == f.rootFolderID {
```
ncw: I don't think the logic is correct here...

if `dirID == f.rootFolderID` and `!f.opt.SharedWithMe` then it should be outputing `fmt.Fprintf(parentsQuery, "'%s' in parents", dirID)` but in this case it outputs nothing.

I think the condition needs to be

`if (f.opt.SharedWithMe || f.opt.StarredOnly) && dirID == f.rootFolderID {`

backend/drive/drive.go
```diff
@@ -684,8 +691,16 @@ func (f *Fs) list(ctx context.Context, dirIDs []string, title string, directorie
...
+			if f.opt.SharedWithMe {
+				_, _ = parentsQuery.WriteString("sharedWithMe=true")
+			}
+			if f.opt.StarredOnly {
+				if f.opt.SharedWithMe {
+					_, _ = parentsQuery.WriteString(" and ")
```
ncw: I'm not sure this `and` is correct without brackets - what do you think?

backend/drive/drive.go
jmacul2: Yes, perfect.  I was going to talk about this actually, but you beat me to it.  This logic works for me.

backend/drive/drive.go
jmacul2: It seems to work for me as is....but I'll defer to whatever seems best.

My bigger annoyance is that "starred" and "sharedwithme" seem not to play very well together.  "sharedwithme" seems to retain parent/child folder structures, but starred does not.  I guess it is more of how I wanted to use it.  For example, I had hoped that by "staring" a specific file, it would retain its parent/child folder relationships.  But I admit that would not follow how Google has it setup through its UI whereby when you star an item it only shows that item instead of its full file path.

ncw: If you want to show the starred files in their correct place then you want something like this

```go
		if f.opt.SharedWithMe && dirID == f.rootFolderID {
			_, _ = parentsQuery.WriteString("sharedWithMe=true")
		} else {
			_, _ = fmt.Fprintf(parentsQuery, "'%s' in parents", dirID)
		}
		if f.opt.StarredOnly {
			if parentsQuery.Len() > 1 {
				_, _ = parentsQuery.WriteString(" and ")
			}
			_, _ = parentsQuery.WriteString("(mimeType='" + driveFolderType + "' or starred=true)")
		}
```

That will make --shared-with-me incompatible with --starred-only though as it searches in given directories only.

If you wanted starred to work with sharedwith me then you'd want something like (untested)

```go
		if f.opt.SharedWithMe && dirID == f.rootFolderID {
			_, _ = parentsQuery.WriteString("sharedWithMe=true")
			if f.opt.StarredOnly {
				if parentsQuery.Len() > 1 {
					_, _ = parentsQuery.WriteString(" and ")
				}
				_, _ = parentsQuery.WriteString("starred=true")
			}
		} else {
			_, _ = fmt.Fprintf(parentsQuery, "'%s' in parents", dirID)
			if f.opt.StarredOnly {
				if parentsQuery.Len() > 1 {
					_, _ = parentsQuery.WriteString(" and ")
				}
				_, _ = parentsQuery.WriteString("(mimeType='" + driveFolderType + "' or starred=true)")
			}
		}
```

iamAnshMittal:
When will this get merged waiting for this like forever ???

ncw:
Good question?

@jmacul2 is this finished now? Do you need help finishing it off?

jmacul2:
This is finished for my purposes.  Seems to be working well as I have been using this fork since we lasted talked.

If someone wants to code review it, feel free.

iamAnshMittal:
Can you update your PR with latest rclone version and check if everything works fine ?

If It does tag "ncw" and tell him to merge it.

iamAnshMittal:
@ncw Can you merge it now ??

jmacul2:
Just to clarify, this change is implemented the same way that the UI is presented in google drive.  I thought it best to stay consistent. 

Not sure why its failing on 1.13....seems unrelated to the changes.

PR #3938, opened 2020-02-05
cmd/serve/ftp: replace deprecated os.SEEK_SET with io.SeekStart

alrs:
`os.SEEK_SET` is deprecated, and replaced with `io.SeekStart`.


ncw:
Thank you :-)

PR #3941, opened 2020-02-06
lib/oauthutil: replace deprecated oauth2.NoContext

alrs:
`oauth2.NoContext` is deprecated, this replaces it with `context.Background()`.

While we're in here, would it make sense to put a timeout on this context?


ncw:
> `oauth2.NoContext` is deprecated, this replaces it with `context.Background()`.

Thanks! I'll merge that now.
 
> While we're in here, would it make sense to put a timeout on this context?

A good question... The transaction should use rclone's http.Transport which has timeouts so it shouldn't be necessary.  That said rclone's timeouts are quite long by default 5m.

PR #3952, opened 2020-02-10
backend/premiumizeme: prune unused functions

alrs:
This removes the unused functions `Object.metaHash()`, `Object.srvPath()`, and `Fs.rootSlash()` from the `backend/premiumizeme` package.

### DIFF
--- a/backend/premiumizeme/premiumizeme.go
+++ b/backend/premiumizeme/premiumizeme.go
@@ -315,14 +315,6 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	return f, nil
 }
 
-// rootSlash returns root with a slash on if it is empty, otherwise empty string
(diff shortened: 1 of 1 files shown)


ncw:
Thank you - always useful to delete dead code!

Did you use a tool to find these?

alrs:
staticcheck.io

ncw:
golangci-lint which rclone uses supports `staticcheck` - I just tried it and there are a lot of warnings!

```diff
diff --git a/.golangci.yml b/.golangci.yml
index 8b8e9bd73..79357ca35 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,6 +11,7 @@ linters:
     - varcheck
     - govet
     - unconvert
+    - staticcheck
     #- prealloc
     #- maligned
   disable-all: true
```

Do you think it is worth getting to a point where we can enable it?

PR #3953, opened 2020-02-11
Experimental compression remote - continuation

buengese:
This is a continuation of #3194 by @id01. It should now pass all the tests even when running against bucket based remotes. There is still stuff that is far from optimal and especially the solution of including the original filesize in the filename does complicate a lot of things. 
See some off inline comments especially for PutStream for context. I also dropped a small number of features that caused problems to get it here.

I'm opening this new pull request now so that we can create a public beta because I'm almost sure that there are bugs left not uncovered by the tests. 

I'll probably be able to finish this work within the next week or so but no guarantees.


id01:
Whoohoo! It's nice to see the compression remote finally come to fruition :)

thestigma:
Very cool! Keeping my eye on this :)
I wonder if it would have good synergy with the planned "keep a copy of the original hash in the crypt format header " thing that is brewing from my discussion with Nick. It seems like it would compliment this very well - and possibly add some optimization shortcuts? 

We will see. Either way, very promising. Ideally all cloud-storage should be compressed pretty much as bandwidth compared to CPU power is unlikely to change any time in the coming decade or two.

ncw:
> This is a continuation of #3194 by @id01. It should now pass all the tests even when running against bucket based remotes.

Nice one - well done :-)

> There is still stuff that is far from optimal and especially the solution of including the original filesize in the filename does complicate a lot of things.

It does - we had a lot of discussions about that. Putting the filesize in allows single pass directory listings, but at the cost of complicating a lot of things.

I wonder whether there are too many compression options? More options means bloating the code more and potential to confuse the user?

id01:
@ncw the lzma-gzip-lz4 choice is pretty common (see Borg); some people actually requested adding zstd too. I was thinking that now that rclone is platform-dependent due to use of cgo, we could get rid of the third-party binaries entirely and simply integrate the XZ/LZ4 C libraries. I believe that LZ4 has already been integrated.

id01:
Oh wait; @buengese already removed XZ and LZ4 via binary >_<
Should probably figure out a way to add XZ back via library, but that could be something for later.

buengese:
I did. I don't intend to add any C library but I am considering zstd via [klauspost/compress](https://github.com/klauspost/compress). I think lz4, gzip and zstandard is would be a good selection for fast, balanced and strong compression. I'll also probably make compression level selection an advanced option.

id01:
@buengese lz4's base compression was implemented with cgo because anything else was much slower. There doesn't seem to be a good golang XZ library; I'll probably have to implement that myself using liblzma too (in time). XZ is a pretty good compression algorithm for high compression and beats zstd for high compression; ideally, we would have LZ4 for fast, zstd for moderate, and XZ for strong compression.

Also, now that we have LZ4 implemented via only C, maybe it's time to remove Snappy?

buengese:
I'm aware of that there aren't any great compression options in go and I'm fine with using cgo for lz4 for now. 
Regarding xz I'm not sure there is a need for compression that strong that it necessitates the inclusion of and LZMA implementation given that with transparent compression I expect access speeds to be more important than compression ratio.

id01:
I did. The ratios and times were worse.

id01:
I'd probably just do it in a future pull request, not this one :)

id01:
Also, LZMA level 1 is faster and much better than gzip level 9 ._.

buengese:
Hmm I see. I'm going to do some testing myself for now I'd prefer staying cgo free.

id01:
@buengese I was at first reluctant to use lz4 cgo, but after extensive testing. I realized that cgo gives a major performance boost (golang implementations of lz4 are slower than gzip), and a small compression boost. Besides, rclone already includes cgo for rclone mount.

ncw:
> I'm still trying to avoid cgo dependencies.

Yes please! CGO will mean that we can only build support for Windows/Linux/macOS without re-working the build system.

buengese:
Ok given the discussion here I decided to do some extensive benchmarking of compression in golang. I've considered a large number of libraries and a decent number of test files. Note that these benchmarks only cover compression but decompression is pretty much guaranteed to be faster so the compression benchmarks establish a good lower bound for speed.

You can see the results [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vSmbwg-QnUq_B6f37KFBjXi6OAvyARo_zqxSV19u6zrDVqRkrpr0DzYcoaUgm3FSIGDmrbP3e9ffXMW/pubhtml)
My takeaways are

- lz4 is a good choice for minimal and fast compression. 
- there is no reason to use cgo for lz4 even the single thread version is > 100mb/s so the network speed will always be the limiting factor
- we should probably swap out our current lz4 implementation for the one used in this benchmark
- pgzip by klauspost performs very well even one a high compression level and would be preferable over the compression/gzip implementation

- there is no pure go library available that can reach the compression ratios of lzma
- but also lzma is by far slowest 
- zstd offers better compression and higher speeds than gzip for compression levels > 4 

My preferred way of going forward would be: replace the current lz4 library with the one used in this benchmark, replace compression/gzip with pgzip, include zstandard as the highest level compression, remove snappy. Doing this would have an additional advantage all of these libraries implement proper concurrency and produce standard files that can be decompressed with the respective linux tools. 

One could still argue in favor of including libzma via cgo for achieving the best compression ratios possible or including gxz for achieving the best compression ratios possible via a pure go implementation but I'd argue that if this level of compression is required it should probably be done outside of rclone.

Please also note that libraries I'd use are licensed under different flavors of BSD but as far as my knowledge goes given that MIT is more free license we can still use them as long as we comply with the additional terms set by the BSD license.

ncw:
> Ok given the discussion here I decided to do some extensive benchmarking of compression in golang. I've considered a large number of libraries and a decent number of test files. Note that these benchmarks only cover compression but decompression is pretty much guaranteed to be faster so the compression benchmarks establish a good lower bound for speed.

That is a great piece of work - well done :-)

> My preferred way of going forward would be: replace the current lz4 library with the one used in this benchmark, replace compression/gzip with pgzip, include zstandard as the highest level compression, remove snappy. Doing this would have an additional advantage all of these libraries implement proper concurrency and produce standard files that can be decompressed with the respective linux tools.

I think the points you've higlighted

- concurrency
- standard files
- no cgo

Are very good points and I agree 100% with your selection.

So we have

- lz4 for fast compression
- pgzip for standard compression
- zstd for state of the art compression

If I was going to drop one of those, it would be zstd I think...

I'd rather start out with fewer compressors and add more later as it is difficult to remove them afterwards.


> One could still argue in favor of including libzma via cgo for achieving the best compression ratios possible or including gxz for achieving the best compression ratios possible via a pure go implementation but I'd argue that if this level of compression is required it should probably be done outside of rclone.

I think people will either be choosing lz4 to compress at wire speeds or gzip to compress well and slowly to save storage. zstd for super compression. I don't think we need any more options and I'm not keen on cgo at all!

> Please also note that libraries I'd use are licensed under different flavors of BSD but as far as my knowledge goes given that MIT is more free license we can still use them as long as we comply with the additional terms set by the BSD license.

Yes the licensing seems compatible. We should probably make a bit more effort in the rclone release to call out the licenses of the libraries....

id01:
Hi @buengese , you're getting caught up with a problem I tried to solve before. I tested LZ4 using this library https://github.com/pierrec/lz4 as I couldn't find another LZ4 standard compatible library that didn't use cgo. It was slower and worse than gzip at compression level 1. If you say that network speed is the limiting factor in a version of lz4 this slow, we might as well just use gzip, which provides much better compression. There have even been posts online saying that pierrec's lz4 was slower than the included gzip library, so I'm not the only one observing this. If really want to remove cgo, we will have to remove any lz4 compatibility.

id01:
As we already have multithreading implemented in rclone-compress, we can go through the compression benchmark and multiply the speed of every single-threaded library by 6.25 times to find the performance we should get with them. This includes snappy, gxz, gzstd, gzkp, cgzip, and clzma.

id01:
Should we optimize for transfer over the network, I'd say the best selection would be:

- gzstd for standard compression
- gxz for good compression

This would provide all three of the points mentioned above:

- concurrency (implemented in rclone-compress itself)
- standard files (both .gz and .xz files are pretty standard, while .zstd and .snappy files would be custom)
- no cgo (no fast lz4)

buengese:
I really don't see your problem with this LZ4 implementation and you clearly didn't read my benchmarks correctly. I ran LZ4 both with a single thread and all 8 achieving 194Mb/s and 943.34Mb/s respectively for enwik9 or 467Mb/s and 1705.72Mb/s for gob-stream. For all test files concurrent lz4 is the fastest by a significant margin. The cgo implementation might still be faster but I cannot think of a scenario where that really matters. 

I know that you already implemented concurrency in rclone but I intend to get rid of it.  This implementation has the annoying side effect that it produces files that aren't compatible with the standard tools and requires the metadata for compression. Both of these points I consider highly undesirable and in my opinion weigh higher than the fact that this approach might be able to extract slightly more performance than the concurrent libraries I'm suggesting given that proper concurrency is significantly harder.

id01:
@buengese the compression does not produce files that are incompatible with standard tools; I specifically made it so that it is compatible (except for Snappy, which has none). Concatenating gzip files produces a valid gzip file, while concatenating xz files produces a valid xz file. That is the basis of seekable compression. If you were to remove this implementation of concurrency, you would also remove seekability, which is the entire basis of rclone-compress being usable in rclone. Furthermore, this is the same approach used by concurrent gzip and xz libraries...

Also, maybe they improved the lz4 implementation since I last used it. But if your point about network is to be believed, lz4 and gzip-1 will both be bottlenecked by the network the same. Why not just make everything simpler and remove lz4?

buengese:
Seekability is actually a fair point and I admittedly didn't know that concatenating blocks valid for anything but gzip. I'd still prefer that this would be done on the library side but us requiring the metadata for seeking is a valid point for doing it inside rclone. I'll think about a little more but maintaining forks of the libraries just for this feature doesn't sound great either. I don't personally care for the inclusion of lz4 I'd keep it in for lower power devices given that is my tests it's still significantly faster than gzip in these scenarios but whatever. 

So your preferred choice would be

- gzip via compress/gzip with concurrency done by rclone (I'd still be in favor of using klauspost gzip in this case given the significant speed difference)
- xz via goxz 

Drop lz4 and snappy.

id01:
Many archive formats (lz4 streams, gzip files, xz files) are compatible with concatenation due to influence from gzip. I was originally thinking LZ4 could be used for local compressed backups, though now I think about it, there are probably better solutions than rclone for that. We could potentially include LZ4 through non-cgo means, especially if the performance of this LZ4 library has improved since last summer. Honestly, with the wrapper I have (I compressed and hashed each block individually to save on space and compute power instead of concatenating streams), we only need an LZ4 block implementation, not a stream one, which would make it much easier to (potentially) implement faster libraries.

However, if we are optimizing for transfer over the network, I would believe that gzip/xz would be the way to go.

id01:
I ran a simple benchmark.
These are very interesting results:

```
$ time xz -c enwik8 > enwik8.c.xz

real	1m2.888s
user	1m2.670s
sys	0m0.127s

$ time ~/go/bin/gxz -c enwik8 > enwik8.go.xz

real	0m15.045s
user	0m14.989s
sys	0m0.064s
$ time gzip -9 -c enwik8 > enwik8.gzip

real	0m6.958s
user	0m6.908s
sys	0m0.043s
$ ls -l enwik8*
-rw-r--r-- 1 nobody nobody 100000000 Jun  1  2011 enwik8
-rw-r--r-- 1 nobody nobody  26375764 Feb 16 13:28 enwik8.c.xz
-rw-r--r-- 1 nobody nobody  33318676 Feb 16 13:28 enwik8.go.xz
-rw-r--r-- 1 nobody nobody  36445248 Feb 16 13:30 enwik8.gzip
```

Looks like goxz doesn't compress as well, but compresses much faster.

buengese:
Yeah I noticed that too and you can see it in my compression benchmarks when comparing clzma with gxz. The difference between gzip level 9 and gxz varies quite a bit. For enwik9 it’s pretty small but for gob-stream it’s much larger. I’ve allready tried to tweak some of the internal parameters of gxz to increase compression but it didn’t seem to make much of a difference. I’ll have to look a little deeper into libzma to see what’s causing this difference.

id01:
Hey @buengese, any updates on gxz?

buengese:
I'd been rather busy last week so not much has changed here. I haven't really had time to look into the internals of gxz but I think I'm going to include it here as the high compression option. Even if the compression ratio isn't as good as could be we can still improve it late or swap it out for a better implementation. This week looks much better so I hope to get the rest of this pr done during this week though I can't make any promises that I'll have the time to look into improving gxz.

backend/press/alg_lz4.go
```diff
@@ -0,0 +1,224 @@
...
+
+// LZ4Header - Header of our LZ4 file
+//var LZ4Header = []byte{0x04, 0x22, 0x4d, 0x18, 0x70, 0x50, 0x84}
+
+// LZ4Footer - Footer of our LZ4 file
+var LZ4Footer = []byte{0x00, 0x00, 0x00, 0x00} // This is just an empty block
+
+const (
+	frameMagic     uint32 = 0x184D2204
+	frameSkipMagic uint32 = 0x184D2A50
```
golangcibot: `frameSkipMagic` is unused (from `deadcode`)

buengese:
I finally found the time to work on this again. The changes I made now are
- remove snappy
- swap out lz4 library
- use klauspost/compress for gzip
- integrate xz via ulikunitz/xz

I've also changed how the compression package is structured a bit and moved to using an interface for the various compression algorithms. To be honest I'm still not entirely happy with it we're definitely wasting performance and memory allocations are likely terrible. I'm still thinking about ways to improve this but maybe we can now make a beta from the current state.

ncw:
Those changes look good to me :-)

If you want to make a beta then push it to a branch on the main rclone repo.

Or were you thinking we should merge it now?

buengese:
Nah I just wanted your ok for pushing this into an upstream branch.
I'd prefer to leave it there at least for a few days to see if there are any obvious bugs left. Maybe make a forum post so we can get a few people to try it.

After that I'd be ok to merge it but still mark it as experimental/beta. I'm definetly going to some further optimization.

id01:
I agree. This is super experimental, especially with all the code surrounding the messy approach of keeping file sizes in file names.

id01:
@thestigma I had dedicated metadata files that had the file size in them. They contain some other things like block positions, etc. A copy is stored in the file header so that it can list without having to download many small files (which will cause very slow listing)

ncw:
> Nah I just wanted your ok for pushing this into an upstream branch.

That's great

> I'd prefer to leave it there at least for a few days to see if there are any obvious bugs left. Maybe make a forum post so we can get a few people to try it.

Do you want to do the honours of making a forum post?

ncw:
> I feel more and more like advanced features tend to run into the same issue of needing to be able to store extra metadata somewhere. Maybe this is not the time to implement a general metadata support system like that, but it sure would make future projects like this one much less complicated to implement and with fewer janky compromises.

You are 100% right about that @thestigma 

The problems are
- no general place to store metadata
- file name length limitations

We had the same problems with the chunker and crypt backends too - perhaps @ivandeex might want to comment here too.

pgalbavy:
Dumb question time; I have not read the code (my lazy). Will the compression type be selectable? I am thinking of backends that support special handling for certain types, like Google Cloud Storage for gzip transcoding and maybe others.

pgalbavy:
I've tried the beta (/rclone-v1.51.0-089-g07e2cea1-press-beta-linux-amd64) but it doesn't work as expected;

I configure it to use remote "gcs" and "gzip" but it copies the files to a lcoal directory called gcs:

```
./rclone --log-level DEBUG  -P copy /var/www/www.literature.org/index.html gcsz:www.literature.org/blah.html
2020/06/24 11:06:47 DEBUG : rclone: Version "v1.51.0-089-g07e2cea1-press-beta" starting with parameters ["./rclone" "--log-level" "DEBUG" "-P" "copy" "/var/www/www.literature.org/index.html" "gcsz:www.literature.org/blah.html"]
2020/06/24 11:06:47 DEBUG : Using RCLONE_CONFIG_PASS password.
2020/06/24 11:06:47 DEBUG : Using config file from "/home/peter/.rclone.conf"
2020-06-24 11:06:47 DEBUG : index.html: Need to transfer - File not found at Destination
====> 2020-06-24 11:06:47 DEBUG : Local file system at /home/peter/rclone-v1.51.0-089-g07e2cea1-press-beta-linux-amd64/gcs/www.literature.org/blah.html: File to upload is small (2101 bytes), uploading instead of streaming <====
2020-06-24 11:06:47 DEBUG : index.html9913000000000000.gz: MD5 = 514e9a87224f14d2554a5db2e2629b81 OK
2020-06-24 11:06:47 INFO  : index.html9913000000000000.gz: Copied (new)
2020-06-24 11:06:47 DEBUG : index.html: MD5 = 2a2b8f0dd07d8bee4faec4d43bbc1822 OK
2020-06-24 11:06:47 INFO  : index.html: Copied (new)
Transferred:        6.155k / 6.155 kBytes, 100%, 564.715 kBytes/s, ETA 0s
Transferred:            3 / 3, 100%
Elapsed time:         0.0s
2020/06/24 11:06:47 INFO  :
Transferred:        6.155k / 6.155 kBytes, 100%, 564.715 kBytes/s, ETA 0s
Transferred:            3 / 3, 100%
Elapsed time:         0.0s

2020/06/24 11:06:47 DEBUG : 2 go routines active

$ ls -lR gcs
gcs:
total 4
drwxrwxr-x 3 peter peter 4096 Jun 24 11:06 www.literature.org

gcs/www.literature.org:
total 4
drwxrwxr-x 2 peter peter 4096 Jun 24 11:06 blah.html

gcs/www.literature.org/blah.html:
total 8
-rw-rw-r-- 1 peter peter 2101 Jun 24 08:47 index.html9913000000000000.gz
-rw-rw-r-- 1 peter peter  198 Jun 24 08:47 index.html.meta
```


```
remote> 7
--------------------
[gcsz]
type = press
remote = gcs
compression_mode = gzip
--------------------
```

ncw:
@pgalbavy 
> remote = gcs

This needs to say `remote = gcs:`

pgalbavy:
Thanks, that worked, but it is not clear from the prompts at all.

ivandeex:
superceded by https://github.com/rclone/rclone/pull/4566
please close

buengese:
completely forgot this was still open.

PR #3956, opened 2020-02-11
fs/accounting: Restore "Max number of stats groups reached" log line

mmatczuk:

To restore a log line that was commented and merged.

#### Checklist


ncw:
Maybe this should be a DEBUG? I'm not sure it should be an INFO level which are supposed to be of general interest to users.

mmatczuk:
Sure this can be a debug.

ncw:
Looks good - thank you :-)

PR #3960, opened 2020-02-12
backend/s3: add max_retries to options

ajankovic:

Amazon S3 is built to handle different kinds of workloads.
In rare cases where S3 is not able to scale, for whatever reason, users will face status 500 errors.
Main mechanism for handling these errors are retries.
Amount of retries needed is different for each use case.

This change is making max_retries option for s3 backend configurable.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1200,12 +1200,17 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		return nil, err
 	}
 
+	pc := fs.NewPacer(pacer.NewS3(pacer.MinSleep(minSleep)))
+	// Set pacer retries to 0 because we are relying on SDK retry mechanism.
+	// Setting it to 1 because in context of pacer it means 1 attempt.
+	pc.SetRetries(1)
+
 	f := &Fs{
 		name:  name,
 		opt:   *opt,
 		c:     c,
 		ses:   ses,
-		pacer: fs.NewPacer(pacer.NewS3(pacer.MinSleep(minSleep))),
+		pacer: pc,
 		cache: bucket.NewCache(),
 		srv:   fshttp.NewClient(fs.Config),
 		pools: make(map[int64]*pool.Pool),
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -850,7 +850,6 @@ In Ceph, this can be increased with the "rgw list buckets max chunk" option.
 			Default:  memoryPoolFlushTime,
 			Advanced: true,
 			Help: `How often internal memory buffer pools will be flushed.
-
 Uploads which requires additional buffers (f.e multipart) will use memory pool for allocations.
 This option controls how often unused buffers will be removed from the pool.`,
 		}, {
@@ -866,7 +865,6 @@ This option controls how often unused buffers will be removed from the pool.`,
 const (
 	metaMtime           = "Mtime"                // the meta key to store mtime in - eg X-Amz-Meta-Mtime
 	metaMD5Hash         = "Md5chksum"            // the meta key to store md5hash in
-	maxRetries          = 10                     // number of retries to make of operations
 	maxSizeForCopy      = 5 * 1024 * 1024 * 1024 // The maximum size of object we can COPY
 	maxUploadParts      = 10000                  // maximum allowed number of parts in a multi-part upload
 	minChunkSize        = fs.SizeSuffix(1024 * 1024 * 5)
@@ -1095,7 +1093,7 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 		opt.ForcePathStyle = false
 	}
 	awsConfig := aws.NewConfig().
-		WithMaxRetries(maxRetries).
+		WithMaxRetries(fs.Config.LowLevelRetries).
 		WithCredentials(cred).
 		WithHTTPClient(fshttp.NewClient(fs.Config)).
 		WithS3ForcePathStyle(opt.ForcePathStyle).


docs/content/flags.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Global Flags"
 description: "Rclone Global Flags"
-date: "2020-02-01T10:26:53Z"
+date: "2020-02-12T07:00:07+01:00"
```
mmatczuk: ?

backend/s3/s3.go
```diff
@@ -844,15 +844,24 @@ In Ceph, this can be increased with the "rgw list buckets max chunk" option.
...
 				encoder.EncodeDot),
-		}},
+		}, {
+			Name: "max_retries",
+			Help: `Maximum number of retries.
+
+The maximum number of times that a request will be retried for failures.
+Defaults to 20. Setting it to -1 defers the max retry setting to the service
+specific configuration.`,
+			Default:  20,
```
mmatczuk: Should default to 10?

docs/content/flags.md
ajankovic: This automatically done when you generate docs.

ncw:
How many retries does S3 do by default?

All the S3 operations are wrapped in pacer calls which will retry them up to `--low-level-retries` times anyway.

So maybe we should be setting the retries to 0 (or 1)?

Thoughts?

ajankovic:
AWS SDK defaults to -1 which defers setting to service specific configuration. Default for rclone is 10 and it will remain 10 with this change.

Pacer retries are ok but using SDK backoff is more appropriate for S3 service in my opinion. I think we should be setting pacer retries to 0 just for S3 to remove complexity of understanding two levels of retries.

Also can you please inspect [behavior of pacer in s3](https://github.com/rclone/rclone/blob/5a231345741c4d9b646b57dbdd342c8e18933108/backend/s3/s3.go#L2367) with [this check in the loop](https://github.com/rclone/rclone/blob/1d73f071f658422b4ef547e8f7993e77d5823cda/fs/operations/operations.go#L441). Where is pacer going to backoff before making the next attempt? I think there is no backoff, but I might be missing something? And that's the primary reason for me to just use default one from the SDK.

mmatczuk:
Please restore the original commit and add a new one with new changes.

ajankovic:
> Please restore the original commit and add a new one with new changes.

Done.

ncw:
> AWS SDK defaults to -1 which defers setting to service specific configuration. Default for rclone is 10 and it will remain 10 with this change.

OK

> Pacer retries are ok but using SDK backoff is more appropriate for S3 service in my opinion. I think we should be setting pacer retries to 0 just for S3 to remove complexity of understanding two levels of retries.

I agree we don't want two things doing retries otherwise we'll be getting 10*10 retries! Does the S3 SDK retry all operations do you know? I suspect it probably only retries idempotent requests but I can't see any docs on what it retries - can you?

If we go with using S3 SDK retries then the default setting for s3 retries should be fs.Config.LowLevelRetries (which is also 10 by default!).

> Also can you please inspect [behavior of pacer in s3](https://github.com/rclone/rclone/blob/5a231345741c4d9b646b57dbdd342c8e18933108/backend/s3/s3.go#L2367) with [this check in the loop](https://github.com/rclone/rclone/blob/1d73f071f658422b4ef547e8f7993e77d5823cda/fs/operations/operations.go#L441). 

That loop can't be retried easily because the io.Reader `in` has been read. What `CallNoRetry` does is mark it for a retry at a higher level in `operations.Copy` so the `io.Reader` can be reset. Last time I looked at the SDK it has some logic to try to see if it can rewind an `io.Reader` which it won't be able to do in this case.

https://github.com/rclone/rclone/blob/964f1f6a7e4cf150331344c40efc516756683509/fs/operations/operations.go#L441-L442

> Where is pacer going to backoff before making the next attempt? I think there is no backoff, but I might be missing something? And that's contributing factor to just use default one from the SDK.

I think there will still be a backoff as normal.

Did you notice a problem with the retries BTW?

mmatczuk:
Hey guys,
I agree that having two rery mechanisms is not so good. The retry logic present in almost every SDK,.

> All of the methods of this package use exponential backoff to retry calls that fail with certain errors, as described in https://cloud.google.com/storage/docs/exponential-backoff. Retrying continues indefinitely unless the controlling context is canceled or the client is closed. See context.WithTimeout and context.WithCancel.

https://godoc.org/cloud.google.com/go/storage

So I have some questions:
* Why pacer does retries? 
* Why this is called low level retries, are there any high level retries?

I also have some suggestions:
* Rename `pacer.Call` to `pacer.CallWithRetry` 
* Rename `pacer.CallNoRetry` to `pacer.Call`
* In s3 replace all CallNoRetry to Call...
* Most probably in other backends we can do the similar

Do you want me to open some new issues?

ncw:
> I agree that having two rery mechanisms is not so good. The retry logic present in almost every SDK,.
> 
> > All of the methods of this package use exponential backoff to retry calls that fail with certain errors, as described in https://cloud.google.com/storage/docs/exponential-backoff. Retrying continues indefinitely unless the controlling context is canceled or the client is closed. See context.WithTimeout and context.WithCancel.
> 
> https://godoc.org/cloud.google.com/go/storage

That is the google cloud SDK which rclone doesn't use - it uses the lower level SDK that it is built on.
 
> So I have some questions:
> 
> * Why pacer does retries?

It has several functions
- pacing the calls when things are going well
- doing the exponential backoff when things are going badly
- doing the retries

> * Why this is called low level retries, are there any high level retries?

Yes, high level retries retry the whole sync. Controlled with the --retry flag

> I also have some suggestions:
> 
> * Rename `pacer.Call` to `pacer.CallWithRetry`
> * Rename `pacer.CallNoRetry` to `pacer.Call`
> * In s3 replace all CallNoRetry to Call...
> * Most probably in other backends we can do the similar

I'd like to leave these how they are - I don't think the names are a barrier to understanding.

backend/s3/s3.go
```diff
@@ -1180,12 +1180,18 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Initialize pacer to 0 retries because we are relying on SDK with
+	// MaxRetries for that
```
mmatczuk: Dot at the end maybe.

backend/s3/s3.go
```diff
@@ -1180,12 +1180,18 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
...
+	pc := fs.NewPacer(pacer.NewS3(pacer.MinSleep(minSleep)))
+	pc.SetRetries(1)
```
mmatczuk: 1 vs 0

ajankovic: It doesn't work with 0. Attempts would be a better naming in this case.

backend/s3/s3.go
ajankovic: Consistent with rclone comments where there is no dot.

mmatczuk:
Looks good to me.

ncw:

I'll merge it now.

PR #3963, opened 2020-02-13
oauthutil: give robust CLI command line

outloudvi:
It append "--" on the `rclone authorize` command line before `client_id`, in case that the `client_id` or `client_secret` has a prefix of "-" (OneDrive's does), and affects the argument parsing of `getopt` (see man page `getopt(1)`).

Nope.

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -432,7 +432,7 @@ func doConfig(id, name string, m configmap.Mapper, oauthConfig *oauth2.Config, o
 			fmt.Printf("For this to work, you will need rclone available on a machine that has a web browser available.\n")
 			fmt.Printf("Execute the following on your machine (same rclone version recommended) :\n")
 			if changed {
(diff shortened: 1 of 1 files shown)


ncw:
Great fix - thank you :-)

Cnly:
OneDrive's client secret may have more naughty characters than that. IIRC mine had something (don't remember exactly what, perhaps a `$`?) that required me to manually escape it before the command would run.

ncw:
Getting into tricky cross platform territory here! We could escape it with `"` which will work on windows/linux/macOS but won't solve the `$` problem.

Single quotes would be best for unix platforms, but they don't work on windows.

Note that putting things in any sort of quotes doesn't solve the leading `-` problem which this PR addresses!

outloudvi:
@ncw Spoken of Windows, is `--` supported on Windows? Maybe we shall limit the `--` feature just on UNIX, and keep Windows as is, since Windows doesn't take `-` seriously but `/`.

ncw:
> Spoken of Windows, is `--` supported on Windows?

Yes, this is parsed by rclone, not the shell so Windows will have the same problem and needs the `--`

PR #3965, opened 2020-02-13
Prune qingstor

alrs:
I linted this with `staticcheck`.

This prunes the unused `list()` function and satisfies the linter by removing a noop `fmt.Sprintf()`.


ncw:

PR #3973, opened 2020-02-17
backend/onedrive: support onedrive national - fixes #3804

### LINKED ISSUE #3804: OneDrive: support national versions of Office 365
**To contributors**: We have got several PRs on this issue: #3973 #4072 #4131, please think twice if you want to open a new PR on it. Instead, consider help review these PRs.

latest

National (Germany, Chinese, and US Government) version of Office 365 requires a different API endpoint. Please refer to https://docs.microsoft.com/en-us/graph/deployments

xytoki:

Make onedrive backend support national versions mentioned in [this Microsoft doc](https://docs.microsoft.com/en-us/graph/deployments).
#3804

#### Checklist

### DIFF
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -42,19 +42,47 @@ const (
 	minSleep                    = 10 * time.Millisecond
 	maxSleep                    = 2 * time.Second
 	decayConstant               = 2 // bigger for slower decay, exponential
-	graphURL                    = "https://graph.microsoft.com/v1.0"
-	configDriveID               = "drive_id"
-	configDriveType             = "drive_type"
-	driveTypePersonal           = "personal"
-	driveTypeBusiness           = "business"
-	driveTypeSharepoint         = "documentLibrary"
-	defaultChunkSize            = 10 * fs.MebiByte
-	chunkSizeMultiple           = 320 * fs.KibiByte
+
+	graphAPIEndpoint = "https://graph.microsoft.com"
+	authEndpoint     = "https://login.microsoftonline.com"
+
+	graphAPIEndpointUS = "https://graph.microsoft.us"
+	authEndpointUS     = "https://login.microsoftonline.us"
+
+	graphAPIEndpointDE = "https://graph.microsoft.de"
+	authEndpointDE     = "https://login.microsoftonline.de"
+
+	graphAPIEndpointCN = "https://microsoftgraph.chinacloudapi.cn"
+	authEndpointCN     = "https://login.chinacloudapi.cn"
+
+	configDriveID       = "drive_id"
+	configDriveType     = "drive_type"
+	driveTypePersonal   = "personal"
+	driveTypeBusiness   = "business"
+	driveTypeSharepoint = "documentLibrary"
+	defaultChunkSize    = 10 * fs.MebiByte
+	chunkSizeMultiple   = 320 * fs.KibiByte
 )
 
 // Globals
 var (
 	// Description of how to auth for this app for a business account
+	oauthEndpoint = &oauth2.Endpoint{
+		AuthURL:  authEndpoint + "/common/oauth2/v2.0/authorize",
+		TokenURL: authEndpoint + "/common/oauth2/v2.0/token",
+	}
+	oauthEndpointCN = &oauth2.Endpoint{
+		AuthURL:  authEndpointCN + "/common/oauth2/v2.0/authorize",
+		TokenURL: authEndpointCN + "/common/oauth2/v2.0/token",
+	}
+	oauthEndpointUS = &oauth2.Endpoint{
+		AuthURL:  authEndpointUS + "/common/oauth2/v2.0/authorize",
+		TokenURL: authEndpointUS + "/common/oauth2/v2.0/token",
+	}
+	oauthEndpointDE = &oauth2.Endpoint{
+		AuthURL:  authEndpointDE + "/common/oauth2/v2.0/authorize",
+		TokenURL: authEndpointDE + "/common/oauth2/v2.0/token",
+	}
 	oauthConfig = &oauth2.Config{
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  "https://login.microsoftonline.com/common/oauth2/v2.0/authorize",
@@ -78,8 +106,32 @@ func init() {
 		Description: "Microsoft OneDrive",
 		NewFs:       NewFs,
 		Config: func(name string, m configmap.Mapper) {
+			opt := new(Options)
+			err := configstruct.Set(m, opt)
+			if err != nil {
+				fs.Errorf(nil, "Couldn't parse config into struct: %v", err)
+				return
+			}
+
+			graphURL := graphAPIEndpoint + "/v1.0"
+			oauthConfig.Endpoint = *oauthEndpoint
+			switch opt.Region {
+			case "cn":
+				graphURL = graphAPIEndpointCN + "/v1.0"
+				oauthConfig.Endpoint = *oauthEndpointCN
+				break
+			case "us":
+				graphURL = graphAPIEndpointUS + "/v1.0"
+				oauthConfig.Endpoint = *oauthEndpointUS
+				break
+			case "de":
+				graphURL = graphAPIEndpointDE + "/v1.0"
+				oauthConfig.Endpoint = *oauthEndpointDE
+				break
+			}
+
 			ctx := context.TODO()
-			err := oauthutil.Config("onedrive", name, m, oauthConfig)
+			err = oauthutil.Config("onedrive", name, m, oauthConfig)
 			if err != nil {
 				log.Fatalf("Failed to configure token: %v", err)
 				return
@@ -223,6 +275,23 @@ func init() {
 			Name: config.ConfigClientSecret,
 			Help: "Microsoft App Client Secret\nLeave blank normally.",
 		}, {
+			Name:    "region",
+			Help:    "Choose national cloud region for OneDrive.",
+			Default: "global",
+			Examples: []fs.OptionExample{{
+				Value: "global",
+				Help:  "Microsoft Cloud Global",
+			}, {
+				Value: "us",
+				Help:  "Microsoft Cloud for US Government",
+			}, {
+				Value: "de",
+				Help:  "Microsoft Cloud Germany",
+			}, {
+				Value: "cn",
+				Help:  "Azure and Office 365 operated by 21Vianet in China",
+			}},
+		}, {
 			Name: "chunk_size",
 			Help: `Chunk size to upload files with - must be multiple of 320k (327,680 bytes).
 
@@ -303,6 +372,7 @@ listing, set this option.`,
 
 // Options defines the configuration for this backend
 type Options struct {
+	Region             string               `config:"region"`
 	ChunkSize          fs.SizeSuffix        `config:"chunk_size"`
 	DriveID            string               `config:"drive_id"`
 	DriveType          string               `config:"drive_type"`
@@ -420,7 +490,7 @@ func (f *Fs) readMetaDataForPathRelativeToID(ctx context.Context, normalizedID s
 	if relPath != "" {
 		relPath = "/" + withTrailingColon(rest.URLPathEscape(f.opt.Enc.FromStandardPath(relPath)))
 	}
-	opts := newOptsCall(normalizedID, "GET", ":"+relPath)
+	opts := newOptsCall(normalizedID, "GET", ":"+relPath, f.opt.Region)
 	err = f.pacer.Call(func() (bool, error) {
 		resp, err = f.srv.CallJSON(ctx, &opts, nil, &info)
 		return shouldRetry(resp, err)
@@ -557,6 +627,23 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		return nil, errors.New("unable to get drive_id and drive_type - if you are upgrading from older versions of rclone, please run `rclone config` and re-configure this backend")
 	}
 
+	rootURL := graphAPIEndpoint + "/v1.0" + "/drives/" + opt.DriveID
+	oauthConfig.Endpoint = *oauthEndpoint
+	switch opt.Region {
+	case "us":
+		rootURL = graphAPIEndpointUS + "/v1.0" + "/drives/" + opt.DriveID
+		oauthConfig.Endpoint = *oauthEndpointUS
+		break
+	case "de":
+		rootURL = graphAPIEndpointDE + "/v1.0" + "/drives/" + opt.DriveID
+		oauthConfig.Endpoint = *oauthEndpointDE
+		break
+	case "cn":
+		rootURL = graphAPIEndpointCN + "/v1.0" + "/me/drive"
+		oauthConfig.Endpoint = *oauthEndpointCN
+		break
+	}
+
 	root = parsePath(root)
 	oAuthClient, ts, err := oauthutil.NewClient(name, m, oauthConfig)
 	if err != nil {
@@ -569,7 +656,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		opt:       *opt,
 		driveID:   opt.DriveID,
 		driveType: opt.DriveType,
-		srv:       rest.NewClient(oAuthClient).SetRoot(graphURL + "/drives/" + opt.DriveID),
+		srv:       rest.NewClient(oAuthClient).SetRoot(rootURL),
 		pacer:     fs.NewPacer(pacer.NewDefault(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
 	}
 	f.features = (&fs.Features{
@@ -689,7 +776,7 @@ func (f *Fs) CreateDir(ctx context.Context, dirID, leaf string) (newID string, e
 	// fs.Debugf(f, "CreateDir(%q, %q)\n", dirID, leaf)
 	var resp *http.Response
 	var info *api.Item
-	opts := newOptsCall(dirID, "POST", "/children")
+	opts := newOptsCall(dirID, "POST", "/children", f.opt.Region)
 	mkdir := api.CreateItemRequest{
 		Name:             f.opt.Enc.FromStandardName(leaf),
 		ConflictBehavior: "fail",
@@ -721,7 +808,7 @@ type listAllFn func(*api.Item) bool
 func (f *Fs) listAll(ctx context.Context, dirID string, directoriesOnly bool, filesOnly bool, fn listAllFn) (found bool, err error) {
 	// Top parameter asks for bigger pages of data
 	// https://dev.onedrive.com/odata/optional-query-parameters.htm
-	opts := newOptsCall(dirID, "GET", "/children?$top=1000")
+	opts := newOptsCall(dirID, "GET", "/children?$top=1000", f.opt.Region)
 OUTER:
 	for {
 		var result api.ListChildrenResponse
@@ -870,7 +957,7 @@ func (f *Fs) Mkdir(ctx context.Context, dir string) error {
 
 // deleteObject removes an object by ID
 func (f *Fs) deleteObject(ctx context.Context, id string) error {
-	opts := newOptsCall(id, "DELETE", "")
+	opts := newOptsCall(id, "DELETE", "", f.opt.Region)
 	opts.NoResponse = true
 
 	return f.pacer.Call(func() (bool, error) {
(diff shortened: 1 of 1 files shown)


backend/onedrive/onedrive.go
```diff
@@ -42,7 +42,19 @@ const (
 	minSleep                    = 10 * time.Millisecond
 	maxSleep                    = 2 * time.Second
 	decayConstant               = 2 // bigger for slower decay, exponential
-	graphURL                    = "https://graph.microsoft.com/v1.0"
+
+	graphAPIEndpoint            = "https://graph.microsoft.com"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

xytoki:
I could only test this with onedrive global and `Azure and Office 365 operated by 21Vianet in China` since I don't have accounts from other national regions.

ncw:
What do you think of this @Cnly ?

backend/onedrive/onedrive.go
```diff
@@ -78,8 +106,32 @@ func init() {
...
+			err := configstruct.Set(m, opt)
+			if err != nil {
+				fs.Errorf(nil, "Couldn't parse config into struct: %v", err)
+				return
+			}
+
+			graphURL := graphAPIEndpoint + "/v1.0"
+			oauthConfig.Endpoint = *oauthEndpoint
+			switch opt.Region {
+			case "cn":
```
Cnly: Perhaps we can store the graph and auth endpoints in a map (or two maps)? That way we don't need so many switch cases or predefined `oauth2.Endpoint` instances.

backend/onedrive/onedrive.go
```diff
@@ -1799,8 +1886,8 @@ func (o *Object) ID() string {
 	return o.id
 }
 
-func newOptsCall(normalizedID string, method string, route string) (opts rest.Opts) {
-	id, drive, rootURL := parseNormalizedID(normalizedID)
+func newOptsCall(normalizedID string, method string, route string, Region string) (opts rest.Opts) {
```
Cnly:
```suggestion
func newOptsCall(normalizedID string, method string, route string, region string) (opts rest.Opts) {
```

Lower case by convention

backend/onedrive/onedrive.go
```diff
@@ -1818,7 +1905,22 @@ func newOptsCall(normalizedID string, method string, route string) (opts rest.Op
...
 // Such a normalized ID can come from (*Item).GetID()
-func parseNormalizedID(ID string) (string, string, string) {
+func parseNormalizedID(ID string, Region string) (string, string, string) {
+	graphURL := graphAPIEndpoint + "/v1.0"
+	switch Region {
+	case "us":
+		graphURL = graphAPIEndpointUS + "/v1.0"
+		break
+	case "de":
+		graphURL = graphAPIEndpointUS + "/v1.0"
```
Cnly:
```suggestion
		graphURL = graphAPIEndpointDE + "/v1.0"
```

Looks like a typo here. Consider using a map (same suggestion as above) to avoid these.

backend/onedrive/onedrive.go
```diff
@@ -1818,7 +1905,22 @@ func newOptsCall(normalizedID string, method string, route string) (opts rest.Op
...
+		break
+	case "cn":
+		if strings.Index(ID, "#") >= 0 {
+			s := strings.Split(ID, "#")
+			return s[1], "", ""
```
Cnly: I'm not sure what's happening here?

backend/onedrive/onedrive.go
```diff
@@ -1818,10 +1865,11 @@ func newOptsCall(normalizedID string, method string, route string) (opts rest.Op
 // parseNormalizedID parses a normalized ID (may be in the form `driveID#itemID` or just `itemID`)
 // and returns itemID, driveID, rootURL.
 // Such a normalized ID can come from (*Item).GetID()
-func parseNormalizedID(ID string) (string, string, string) {
+func parseNormalizedID(ID string, region string) (string, string, string) {
+	rootUrl := graphAPIEndpoint[region] + "/v1.0/drives"
```
golangcibot: var `rootUrl` should be `rootURL` (from `golint`)

xytoki:
Two maps and lowercases are used now.
However now there's a problem that in onedrive china version `rclone ls` command always list all files in the drive ignoring the user specified path,and fixing this is what the codes below do.
[https://github.com/rclone/rclone/blob/32719ee2c0276438fdac9d655e58d1a665c971e0/backend/onedrive/onedrive.go#L1917-L1923](https://github.com/rclone/rclone/blob/32719ee2c0276438fdac9d655e58d1a665c971e0/backend/onedrive/onedrive.go#L1917-L1923)
I think it's a dirty hack ( it's from [here](https://gist.github.com/ShadeShady/81b6fecca82c289e8e5ae0637918e7e6#file-rclone_onedrive_21vianet-patch-L104) ) so I'm looking deeper trying  to resolve it.

xytoki:
OK I know what happended.
Onedrive China and onedrive global behaves differently when querying files which not exists.
For example , GET
```
https://microsoftgraph.chinacloudapi.cn/v1.0/drives/<my-drive-id>/items/<root-itemid>:/ObviouslyNotExist:
```
in onedrive global should returns an error like
```json
{
  "error": {
    "code": "itemNotFound",
    "message": "The resource could not be found.",
    "innerError": {
      "request-id": "81b43054-01bc-4c24-bcd4-56d7878076f3",
      "date": "2020-02-18T01:29:30"
    }
  }
}
```
But in onedrive china,GET 
```
https://microsoftgraph.chinacloudapi.cn/v1.0/drives/<driveid>/items/<root-itemid>:/<anything>:
```
always returns info of the root folder like below,causing rclone listing all files,no matter <anything> exists or not.
```json
{
    "@odata.context": "https://microsoftgraph.chinacloudapi.cn/v1.0/$metadata#drives('<my-driveid>')/items/$entity",
    "createdDateTime": "2019-07-24T16:49:03Z",
    "id": "<my-itemid>",
    "lastModifiedDateTime": "2020-02-14T00:21:07Z",
    "name": "root",
    "webUrl": "<my-weburl>",
    "size": 974705440021,
    "parentReference": {
        "driveId": "<my-driveid>",
        "driveType": "business"
    },
    "fileSystemInfo": {
        "createdDateTime": "2019-07-24T16:49:03Z",
        "lastModifiedDateTime": "2020-02-14T00:21:07Z"
    },
    "folder": {
        "childCount": 5
    },
    "root": {}
}
```

xytoki:
Maybe something should be changed to avoid this. I'll try to look for solutions.

xytoki:
Maybe change `newOptsCall` and `parseNormalizedID` like this? I think it's also a 'hack'.
```go
func newOptsCall(normalizedID string, method string, route string, region string) (opts rest.Opts) {
	id, drive, rootURL := parseNormalizedID(normalizedID, region)
	if drive != "" {
		path:="/" + drive + "/items/" + id + route
		if region=="cn" {
			path="/items/" + id + route
		}
		return rest.Opts{
			Method:  method,
			RootURL: rootURL,
			Path:    path,
		}
	}
	return rest.Opts{
		Method: method,
		Path:   "/items/" + id + route,
	}
}

func parseNormalizedID(ID string, region string) (string, string, string) {
	rootURL := graphAPIEndpoint[region] + "/v1.0/drives"
	if region=="cn" {
		rootURL = graphAPIEndpoint[region] + "/v1.0/me/drive"
	}
	if strings.Index(ID, "#") >= 0 {
		s := strings.Split(ID, "#")
		return s[1], s[0], rootURL
	}
	return ID, "", ""
}
```

backend/onedrive/onedrive.go
```diff
@@ -415,12 +456,19 @@ func shouldRetry(resp *http.Response, err error) (bool, error) {
...
+//
 // If `relPath` == '', do not append the slash (See #3664)
 func (f *Fs) readMetaDataForPathRelativeToID(ctx context.Context, normalizedID string, relPath string) (info *api.Item, resp *http.Response, err error) {
 	if relPath != "" {
 		relPath = "/" + withTrailingColon(rest.URLPathEscape(f.opt.Enc.FromStandardPath(relPath)))
 	}
-	opts := newOptsCall(normalizedID, "GET", ":"+relPath)
+	path := ":" + relPath
+	if f.opt.Region == "cn" {
+		path = "/children" +relPath
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		path = "/children" + relPath
```

xytoki:
OK.I found that `drives/driveID/items/itemID/children/relativePath` is available, so I changed all relative code when `region` is `cn`.

xytoki:
I don't know whether APIs of other regions have the same behavior, so I only change this for China region.

backend/onedrive/onedrive.go
```diff
@@ -1715,12 +1784,16 @@ func (o *Object) uploadSinglepart(ctx context.Context, in io.Reader, size int64,
 	var resp *http.Response
 	var opts rest.Opts
 	leaf, directoryID, _ := o.fs.dirCache.FindPath(ctx, o.remote, false)
-	trueDirID, drive, rootURL := parseNormalizedID(directoryID)
+	trueDirID, drive, rootURL := parseNormalizedID(directoryID, o.fs.opt.Region)
 	if drive != "" {
+		path := "/" + drive + "/items/" + trueDirID + ":/" + rest.URLPathEscape(o.fs.opt.Enc.FromStandardName(leaf)) + ":/content"
+		if o.fs.opt.Region == "cn" {
+			path = "/" + drive + "/items/" + trueDirID + "/children/" + rest.URLPathEscape(o.fs.opt.Enc.FromStandardName(leaf)) + "/content"
```
Cnly: Does this mean the request will fail if the file is named `content`?

backend/onedrive/onedrive.go
```diff
@@ -432,14 +480,18 @@ func (f *Fs) readMetaDataForPathRelativeToID(ctx context.Context, normalizedID s
 // readMetaDataForPath reads the metadata from the path (relative to the absolute root)
 func (f *Fs) readMetaDataForPath(ctx context.Context, path string) (info *api.Item, resp *http.Response, err error) {
 	firstSlashIndex := strings.IndexRune(path, '/')
-
-	if f.driveType != driveTypePersonal || firstSlashIndex == -1 {
+	if (f.driveType != driveTypePersonal && f.opt.Region != "cn") || firstSlashIndex == -1 {
```
Cnly:
```suggestion
	if f.driveType != driveTypePersonal || firstSlashIndex == -1 {
```

As the difference between the Chinese version and the global is only `:/` and `/children/`, we can still catch the Chinese case in this block. The remaining code outside this block is intended for OneDrive Personal only, which has "shared with me" folders. They don't exist in OneDrive for Business.

backend/onedrive/onedrive.go
```diff
@@ -432,14 +480,18 @@ func (f *Fs) readMetaDataForPathRelativeToID(ctx context.Context, normalizedID s
...
 		var opts rest.Opts
 		if len(path) == 0 {
 			opts = rest.Opts{
 				Method: "GET",
 				Path:   "/root",
 			}
+		} else if f.opt.Region == "cn" {
```
Cnly: This new branch can be removed. We only need to use `:/` for the global version, or `/children/` for the Chinese version, in the next `else` branch.

backend/onedrive/onedrive.go
xytoki: I tested this case and the request succeeded.
```
[toki@xytoki-vm rclone]$ ./rclone copy local:/home/toki/content od:/ -vvv
2020/02/18 22:47:07 DEBUG : rclone: Version "v1.51.0-DEV" starting with parameters ["./rclone" "copy" "local:/root/content" "od:/" "-vvv"]
2020/02/18 22:47:07 DEBUG : Using config file from "/home/toki/.config/rclone/rclone.conf"
2020/02/18 22:47:08 DEBUG : content: Need to transfer - File not found at Destination
2020/02/18 22:47:08 DEBUG : content: Starting singlepart upload
2020/02/18 22:47:09 DEBUG : content: QuickXorHash = 0000000000000000000000000000000000000000 OK
2020/02/18 22:47:09 INFO  : content: Copied (new)
2020/02/18 22:47:09 INFO  : 
Elapsed time:         0.8s

[toki@xytoki-vm rclone]$ ./rclone ls od:/ | grep content
        0 content
```

backend/onedrive/onedrive.go
xytoki: Here `/children/` in the Chinese version won't work when called `/items/<id>/children/<dir>/<file>`.It only allows one level of subdirectory.
 So I think maybe the solution for OneDrive personal  "shared with me" folders is more suitable.

xytoki:
So now constants region codes are used.
Here /children/ in the Chinese version won't work when called /items/<id>/children/<dir>/<file>.It only allows one level of subdirectory.
So I think maybe the solution for OneDrive personal "shared with me" folders is more suitable.

backend/onedrive/onedrive.go
```diff
@@ -78,8 +93,17 @@ func init() {
 		Description: "Microsoft OneDrive",
 		NewFs:       NewFs,
 		Config: func(name string, m configmap.Mapper) {
+			opt := new(Options)
+			err := configstruct.Set(m, opt)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/onedrive/onedrive.go
Cnly: I don't think it's assumed in the code that `readMetaDataForPathRelativeToID` is called with only one level of subdirectory. The special handling for shared folders in `readMetaDataForPath` only makes sure that if we're querying an item inside a folder, we guarantee that it's queried with a path relative to the ID of *at least* the first level directory in its path. That is to say, if you call `readMetaDataForPath(ctx, "a/b/c")`, it's possible that the function in turn calls `readMetaDataForPathRelativeToID(ctx, idOfA, "b/c")`.

xytoki: Then maybe a brand new implementation is needed. I'll try to work for it.

xytoki: Now there are to solutions. The first is to implement a function to query subdirectories one by one, and another is to switch to `/me/drive` in the Chinese version.
I'll try the first one first.

Max-Sum:
The built-in client id does not seem to work with 21Vianet's onedrive. Do we need a set of ID/Secret for each region?

ncw:
Sorry this PR fell off my radar. Where are we at with it?

xytoki:
The APIs of the OneDrive China version have many incompatibilities with the global one, so maybe I need some help.
In addition, I don't have accounts of the other regions so I can't test if there are also any incompatibilities.

ncw:
@Cnly have you got time to help with this?

oriso:
@xytoki 
When file path or file name include brackets, it always failed to copy to onedrive. Btw, I'm using 21Vianet version onedrive.
Error is:

> BadRequest: Bad Request - Error in query syntax.

Any idea?

742481030:
> 这里来的

我看过此pr ,不能支持sharepoint 写死了me/root
我的方法是search sharepoint时候配置文件写入站点id,这样就能正常使用sharepoint了,我之前粗暴的把国际版api地址全部替换成国内版本,他是不能正常上传文件的 会报错文件不支持的文件路径,最后我只好存储了siteid,可能由于其他原因 世纪互联对于drives/driveid 的sharepoint api 没有写权限或者其他缘故

742481030:
> 更详细的信息：
> Onedrive China在查询文件时表现异常。
> `GET /drives/<drive-id>/root:/<path>` 并且
> `GET /drives/<drive-id>/items/<item-id>:/<path>`不起作用，
> 但是替换`/drives/<drive-id>`后`/me/drive`效果很好。
> 
> 例如，
> `https://graph.microsoft.com/v1.0/drives/<my-drive-id>/items/<itemid>:/ObviouslyNotExist: `onedrive global中的GET 应该返回错误，而获取普通文件应该返回文件信息。
> 
> 但是在onedrive china中，GET
> `https://microsoftgraph.chinacloudapi.cn/v1.0/drives/<driveid>/items/<itemid>:/<anything>:`
> 总是返回`itemid`如下所示url 中提供的信息，从而导致rclone列出所有文件，无论是否存在。

https://microsoftgraph.chinacloudapi.cn/v1.0/sites/<siteid>/items/<itemid>
就可以解决此问题

742481030:
> 内置客户端ID似乎不适用于21Vianet的onedrive。每个区域是否需要一组ID /秘密？

那样的话选择区域应当在客户端之前

NyaMisty:
Maybe we can minimize the problem and delay the testing of DE & US versions of OneDrive? In this way we can push forward this PR.

ivandeex:
@Cnly @NyaMisty could you please jump in and help out @xytoki ??

ivandeex:
Sorry, I am not a OneDrive backend guru. Just a few thoughts.

As far as I can understand from your discussions, after requests of Communist Party of China (unnamed government authority <s>or directly Xi Jinping</s>), Microsoft introduced tricky small incompatible changes into OneDrive China, so right now:

- backend developers have patches scattered arround backend code
- OneDrive China needs specific testing (e.g. registration using a mobile phone confirmed by Mr Xi)
- incompatible behavior changes

Did someone consider one of the following ways to untangle?
- fully clone `backends/onedrive` into `backends/onedrivechina`, add current China-related patches and let two codebases develop independently (one could even add chinese translation into documentation)
- have two separate test suites: `backend/onedrive/onedrive_test.go` vs `backend/onedrive/onedrive_china_test.go`

Thanks. Sorry for noice.

Cnly:
@ivandeex @NyaMisty I think the primary problem here is that we don't have a large enough group of regional OneDrive users or developers. As a result, even if we just talk about the Chinese version, we've only discovered a subset of the quirks required for it in this PR, and that's not enough knowledge to support it. For the same reason the tests we can run are also very limited... Given this situation, even if we separate out the source code, it doesn't look like the new branch can go much farther. :/

I'd be very happy to see if any regional users can jump in and help with the development/testing.

NyaMisty:
@ivandeex @Cnly 
Hi, sorry for the late reply :(
Firstly, I'm a user of 21vianet OneDrive, so I can provide you with a token for test (but of course privately).
Then I tried to compare multiple fork that modified to support 21vianet. It seems that all of them are simply replacing URL of each API request, and none of other logic are changed. And there're people saying the 21vianet's onedrive is only the old version of international OneDrive. So in this way, I'd suggest that we keep two versions in one backend and only add a specific test for 21vianet version.

As for code: 
- All of other fork simply replaced the `RootURL` of REST srv. 
- In this fork, I saw an additional parameter `region` is added to a function, but I noticed that almost every API call's URL is built by an `if drive == "" {} else {}` block 
  - (which is introduced from @Cnly 's fix 2 years ago.)
  - Maybe we can refactor the code and merge these if-blocks into a new function

Cnly:
@NyaMisty 

> Firstly, I'm a user of 21vianet OneDrive, so I can provide you with a token for test (but of course privately).

Thank you! That would be very helpful. By "token," do you mean a pair of Client ID and secret, or will you also be able to contribute a test account?



> Then I tried to compare multiple fork that modified to support 21vianet. It seems that all of them are simply replacing URL of each API request, and none of other logic are changed.

Do these forks actually work? Since in this PR we've found some tweaks needed for the Chinese version. I'm not sure if everything would work if they just change the API endpoint.



> So in this way, I'd suggest that we keep two versions in one backend and only add a specific test for 21vianet version.

Yeah, that'd be easier to maintain provided we don't need to insert conditional branches everywhere in the code.



> * In this fork, I saw an additional parameter `region` is added to a function, but I noticed that almost every API call's URL is built by an `if drive == "" {} else {}` block
>   * Maybe we can refactor the code and merge these if-blocks into a new function

Sure, we can probably refine the `newOptsCall` function for this.

BTW, would you like to continue the work of this PR? It's beginning to age and your help could really push it forward.

NyaMisty:
@Cnly 
- for the "token", I actually mean a pair of Client ID/Secret, which should be better way for rclone to test (compared with giving account's password)?
- for those forks, they in fact work in many cases, and is actually working fluently for my own use (I haven't noticed any bugs, as I only use them to ls & upload files with normal names)
  - [PR1](https://github.com/rclone/rclone/compare/5b09599a23e455f342f54cecaf2a3174d1e9bf79...NyaMisty:21net#diff-450423db929e61f8263e38700e24926f748ffc3962003040aad070565a1b9c7b)
  - [PR2](https://github.com/rclone/rclone/compare/e5190f14ce843c2bed0a49f083207945f98f8ab2...742481030:master)
- for the PR, I don't have much time recently, but I can spare some time next month (which is Chinese New Year then)

NyaMisty:
- for those forks, they in fact work in many cases, and is actually working fluently for my own use (I haven't noticed any bugs, as I only use them to ls & upload files with normal names)
  - https://github.com/rclone/rclone/compare/5b09599a23e455f342f54cecaf2a3174d1e9bf79...NyaMisty:21net#diff-450423db929e61f8263e38700e24926f748ffc3962003040aad070565a1b9c7b

ivandeex:
@Cnly I think we can close this in favor of #4963

ivandeex:
Obsoleted by #4963

Issue #3588, opened 2019-10-04
DLNA and external subtitles


kxzaon:
I'm testing the DLNA server that seems to be too good to be true but I have a small issue with this DLNA server. It doesn't seem to recognize my external subtitles.
On your website, at the DLNA page, I can read :"Rclone will list all files present in the remote, without filtering based on media formats or file extensions. Additionally, there is no media transcoding support. This means that some players might show files that they are not able to play back correctly."
However, in my debug log, I can see :

`2019/10/04 01:50:32 DEBUG : DLNA server on :7879: unrecognized file type: **************.srt`

I can provide more log if you want.
Is there a way to have external subtitles working ?
The synology DLNA server can see these files and sends them to my LG WebOS player, it is working perfectly.

Thanks for your time.

ncw:
Hmm, I think this is because the mime type of `.srt` files doesn't match

https://github.com/rclone/rclone/blob/d15425e8c8356524ab7c624b5173abaa07a6e077/cmd/serve/dlna/cds.go#L68

which means that this returns `nil, nil`

https://github.com/rclone/rclone/blob/d15425e8c8356524ab7c624b5173abaa07a6e077/cmd/serve/dlna/cds.go#L106-L109

which in turn means this skips the file and emits the debug

https://github.com/rclone/rclone/blob/d15425e8c8356524ab7c624b5173abaa07a6e077/cmd/serve/dlna/cds.go#L166-L174

I'm not sure of the consequence of widening that check though - @sretlawd what do you think?

sretlawd:
Yeah, external SRTs is actually a bit of a pet project of mine ...  :-)  Unfortunately it's not as simple as just listing the SRT file like any other, the subtitles have to be explicitly associated with the media file in the XML metadata.

I have a branch that serves external SRTs with the "modern method", it works with VLC and on my LG webOS 2 TV.  I've been trying to get it to work on my Samsung F series before submitting it, but I've spent untold hours on it and still havn't figured that one out.  It's possible that it may work on newer Samsungs though.

I'll rebase and send up what I have here in a sec, though.

ncw:
I've merged that to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.50

Thanks @sretlawd :-)

kxzaon:
I've tested it out this morning from 3 differents systems and the latest beta continues to give me this error when I try to connect my client (LG WebOS and VLC also) : 
[...]

PR #3976, opened 2020-02-18
dlna: cds: don't specify childCount at all when unknown

sretlawd:
Basically, solving #3541 with a different approach - bringing in
the upstream upnpav module, and changing ChildCount from int to a
*int to avoid childCount="0" in the XML output when that value is
simply unknown.

Current approach is leading to some recursion issues and according
to the DLNA spec it shouldn't be necessary, anyway.

Discussed in #3588 , @jorams reported an issue where the entire remote is being scanned in order to get the childCount metadata.

Alternative approach: could also just add a flag to ensure there's no recursion, but the CDS stuff is already quite messy, so this seems like a better idea and better handles the "top level directory with subdirectories for each media item" use case anyway.

It's possible some clients may have an adverse reaction to the change, but I think it's worth putting it out there, since it is spec compliant.

#### Checklist

### DIFF
--- a/cmd/serve/dlna/cds.go
+++ b/cmd/serve/dlna/cds.go
@@ -16,8 +16,8 @@ import (
 
 	"github.com/anacrolix/dms/dlna"
 	"github.com/anacrolix/dms/upnp"
-	"github.com/anacrolix/dms/upnpav"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


ncw:
Looks fine to me - will merge now :-)

Issue #3967, opened 2020-02-14
Inefficient memory allocations during S3 multipart uploads


zimnx:
#### What is the problem you are having with rclone?
When uploading big files to S3 storage, memory graphs looks like a rollercoaster. Each upload request allocates his own memory buffers for chunks, and once file is uploaded they are garbaged.
![heap_released_bytes](https://user-images.githubusercontent.com/3951334/74548379-0483a080-4f4e-11ea-8a07-b5af1a139da1.png)
In total allocation space per function graph, we can see that multipart upload allocated 14GB in total throught workflow. This should be constant and equal to number of concurrent requests times chunk size.
![Selection_027](https://user-images.githubusercontent.com/3951334/74548441-2b41d700-4f4e-11ea-8d89-77116c5159f0.png)

#### What is your rclone version (output from `rclone version`)
master

####  Which cloud storage system are you using? (eg Google Drive)
S3

ncw:
Rclone has a library for exactly this purpose: https://github.com/rclone/rclone/blob/master/lib/pool/pool.go

Note that chunk sizes can vary with size of file in S3 so using this would need a bit of care.

Fancy sending in a PR to use `lib/pool` in the s3 backend?

zimnx:
@ncw yep, I'll take care of it soon. Thanks for pool reference!

PR #3985, opened 2020-02-19
backend/s3: use memory pool for buffer allocations

zimnx:

Currently each multipart upload allocated his own buffers, which after
file upload were garbaged. Next files couldn't leverage already allocated
memory which resulted in inefficent memory management. This change
introduces backend memory pool keeping memory chunks which can be
used during object operations.

#3967

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -56,6 +56,7 @@ import (
 	"github.com/rclone/rclone/lib/bucket"
 	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/pool"
 	"github.com/rclone/rclone/lib/readers"
 	"github.com/rclone/rclone/lib/rest"
 	"golang.org/x/sync/errgroup"
@@ -841,11 +842,24 @@ In Ceph, this can be increased with the "rgw list buckets max chunk" option.
 			// - doubled / encoding
 			// - trailing / encoding
 			// so that AWS keys are always valid file names
-			Default: (encoder.EncodeInvalidUtf8 |
+			Default: encoder.EncodeInvalidUtf8 |
 				encoder.EncodeSlash |
-				encoder.EncodeDot),
-		}},
-	})
+				encoder.EncodeDot,
+		}, {
+			Name:     "memory_pool_flush_time",
+			Default:  memoryPoolFlushTime,
+			Advanced: true,
+			Help: `How often internal memory buffer pools will be flushed.
+
+Uploads which requires additional buffers (f.e multipart) will use memory pool for allocations.
+This option controls how often unused buffers will be removed from the pool.`,
+		}, {
+			Name:     "memory_pool_use_mmap",
+			Default:  memoryPoolUseMmap,
+			Advanced: true,
+			Help:     `Whether to use mmap buffers in internal memory pool.`,
+		},
+		}})
 }
 
 // Constants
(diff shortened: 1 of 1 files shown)


zimnx:
With this patch, memory consumption is flat and constant.
Selection_045
Total allocated memory by this pool equals to `transfers*uploadConcurrency*chunkSize` which is expected.
Selection_044

backend/s3/s3.go
```diff
@@ -2078,15 +2110,15 @@ func (o *Object) uploadMultipart(ctx context.Context, req *s3.PutObjectInput, si
 	if concurrency < 1 {
 		concurrency = 1
 	}
-	bufs := make(chan []byte, concurrency)
+	concurrencyShares := make(chan struct{}, concurrency)
```
ncw: You could use TokenDispenser here which is essentially this code packaged into a lib which will simplify the code a bit.

https://github.com/rclone/rclone/blob/481c8a40ea716bacb1f646604665f9859b65988f/lib/pacer/tokens.go#L10-L11

backend/s3/s3.go
```diff
@@ -2159,11 +2191,9 @@ func (o *Object) uploadMultipart(ctx context.Context, req *s3.PutObjectInput, si
...
 
 	for partNum := int64(1); !finished; partNum++ {
-		// Get a block of memory from the channel (which limits concurrency)
-		buf := <-bufs
-		if buf == nil {
-			buf = make([]byte, partSize)
-		}
+		// Get a block of memory from the pool and share from channel which limits concurrency.
+		share := <-concurrencyShares
+		buf := f.pool.Get()
```
ncw: There is a problem here in that `partSize` may not be equal to `f.opt.ChunkSize` at this point so the memory chunks in the pool will be the wrong size.

This will happen for large files bigger than 10000*5M = 48GiB files by default.

There are several things we could do here...

1. fall back to `make([]byte, partSize)`
2. make a local `Pool` and call `Flush` on it at the end.
3. make a map of `Pool`s in `f` by partSize

What do you think? 2. is simple and would probably work OK given that there will be ~10,000 chunks used

backend/s3/s3.go
```diff
@@ -859,6 +875,9 @@ const (
 	defaultUploadCutoff = fs.SizeSuffix(200 * 1024 * 1024)
 	maxUploadCutoff     = fs.SizeSuffix(5 * 1024 * 1024 * 1024)
 	minSleep            = 10 * time.Millisecond // In case of error, start at 10ms sleep.
+
+	memoryPoolFlushTime = fs.Duration(time.Minute) // flush the cached buffers after this long
+	memoryPoolUseMmap   = false
```
mmatczuk: Why this is false by default?

backend/s3/s3.go
zimnx: Nice catch. 1 and 2 would bring back initial issue this PR is trying to resolve and 3 would work, i'll implement that.

backend/s3/s3.go
zimnx: Default memory allocator is safer, only advances users knows what they are doing, so I;m leaving a way to set is via a configuration option.

backend/s3/s3.go
```diff
@@ -841,11 +842,26 @@ In Ceph, this can be increased with the "rgw list buckets max chunk" option.
...
-			Default: (encoder.EncodeInvalidUtf8 |
+			Default: encoder.EncodeInvalidUtf8 |
 				encoder.EncodeSlash |
-				encoder.EncodeDot),
-		}},
-	})
+				encoder.EncodeDot,
+		}, {
+			Name:     "memory_pool_flush_time",
+			Provider: "AWS",
```
ncw: I think this is relevant for all the providers so we should probably remove the `Provider` line here and below.

zimnx: Right, fixed

ncw:
That looks great now - I'll merge this now.

Thank you for a great contribution :-)

PR #3987, opened 2020-02-19
Prune azureblob

alrs:
Linted with `staticcheck`.

This removes the unused `Object.parseTimeString()` from `backend/azureblob`, as well as an unneeded `fmt.Sprintf()`.


ncw:
Thank you :-)

Issue #2889, opened 2019-01-04
Expensive daily s3 to disk replication


kardaj:
Hey there! This is an updated post from slack a couple weeks ago.

I have a couple of S3 buckets with hundred thousands of objects each and I'm syncing them to a local server on daily basis. I have been using `aws s3 sync` for a couple of years and in the last few months, I moved to `rclone` to improve sync reliability. This switch came with a noticeable bump in my bill. After breaking down the costs, each daily sync session is costing an average of 10$:
- 6 million HeadObject (2$)
- 1.2 million ListBucket (8$)
- 3.5 thousand GetObject(<0.01$)

From what I noticed, `rclone` is quite aggressive on ListBucket and HeadObject operations. Is there a way to tune down this behaviour through the configuration?

Here's some setup information:
```bash
$rclone version
rclone v1.44
- os/arch: linux/amd64
- go version: go1.11.1
```
The commands I'm using look like this: `rclone sync s3://bucket /s3/bucket --verbose`

ncw:
The first thing to do is to use `--size-only` or `--checksum` as a syncing method.  This will stop rclone reading the metadata to discover the modification time of the files.  This will help with the HeadObject calls.

If you want to reduce the ListBucket calls then try `--fast-list`.  This will use a more memory but it will use much fewer ListBucket calls.

If you want to see what operations rclone is doing then do `-vv --dump headers`.

There are some hints to these things in the rclone docs, but maybe there should be a optimizing s3 section in the s3 docs?

kardaj:
Thank you for the answer. I'll be testing the effects of this optimisation in the next few weeks. I think it's important information that should be present in the documentation as personally, the bill took me by surprise. There was another feedback from the slack channel that had a similar experience with Google Storage. Is the behaviour the same for all providers?

ncw:
> Thank you for the answer. I'll be testing the effects of this optimisation in the next few weeks.

Great - let us know how it goes

> I think it's important information that should be present in the documentation as personally, the bill took me by surprise

Sorry :-(  Let's try to draft some more words for the documents - do you want to have a go?


> There was another feedback from the slack channel that had a similar experience with Google Storage. Is the behaviour the same for all providers?

No, annoyingly the providers are all slightly different!  A constant is that `--fast-list` will help.  `--size-only` will help on s3 & swift, but not on b2 or google cloud storage (if I remember correctly).

dertel:
[...]

PR #4011, opened 2020-02-27
fs/operations: do ignore size when --ignore-size is specified along with --checksum

tiborvass:

This fixes a bug where two different but same-sized objects that cannot be compared
with checksum are considered equal even though --checksum and --ignore-size are
specified.

To test this, use an existing s3 bucket on AWS that was not created with rclone and
that has objects uploaded with multi-part (ETAG != MD5) then try to sync from it to
some destination, twice in a row. The second time it should copy again the objects
that do not have a hash, instead of relying on --size-only behavior.

If someone can add a test for this case, I'd appreciate it!

IMHO, --ignore-size (and not --size-only) should be the default fallback, when
--checksum is specified.

Unsure.

#### Checklist

### DIFF
--- a/backend/sharefile/api/types.go
+++ b/backend/sharefile/api/types.go
@@ -106,7 +106,7 @@ type UploadSpecification struct {
 type UploadFinishResponse struct {
 	Error        bool   `json:"error"`
 	ErrorMessage string `json:"errorMessage"`
-	ErrorCode    int    `json:"errorCode"`
+	ErrorCode    int    `json:"errorCode,string"`
 	Value        []struct {
 		UploadID    string `json:"uploadid"`
 		ParentID    string `json:"parentid"`
@@ -114,7 +114,7 @@ type UploadFinishResponse struct {
 		StreamID    string `json:"streamid"`
 		FileName    string `json:"filename"`
 		DisplayName string `json:"displayname"`
-		Size        int    `json:"size"`
+		Size        int    `json:"size,string"`
 		Md5         string `json:"md5"`
 	} `json:"value"`
 }


ncw:
> To test this, use an existing s3 bucket on AWS that was not created with rclone and
> that has objects uploaded with multi-part (ETAG != MD5) then try to sync from it to
> some destination, twice in a row. The second time it should copy again the objects
> that do not have a hash, instead of relying on --size-only behavior.

Is this using

    rclone copy --checksum --ignore-size s3:bucket /local/path

?

Can you describe the workflow that causes this problem. I don't understand why you would use `--ignore-size` - it is quite an esoteric option.

In general rclone prefers not to keep copying objects so you can run the command multiple times.

tiborvass:
@ncw Thanks for your response!

I was using `rclone sync --checksum --ignore-size s3:bucket s3:otherbucket` but the same would be true for local destination.

The workflow is that I'm copying/syncing to a bucket from another existing bucket that was not created with rclone and has objects that were multipart uploaded, think "backup bucket". By changing one byte in such an object and attempting to copy it to the backup bucket, you will notice that because it has the same size, it doesn't copy, even though: 1. the files are different, and 2. I specified --ignore-size.

Maybe this will help understand:
```
$ mkdir dir dir2
$ bucket=mybucket
$ dd if=/dev/urandom bs=9M count=1 | base64 > dir/file       # I believe default for multipart_threshold is 8M
$ aws s3 cp --recursive dir s3://$bucket/dir
$ aws s3 cp --recursive dir2 s3://$bucket/dir2            # create a second copy
$ dd if=/dev/urandom bs=9M count=1 | base64 > dir/file       # change pretty much all bytes
$ wc -c dir/file             # different content, but same size (and above multipart_threshold)
12748477
$ aws s3 cp --recursive dir s3://$bucket/dir        # update only dir, not dir2 which keeps old content
$ aws s3api head-object --bucket $bucket --key dir/file
{
    "AcceptRanges": "bytes",
    "LastModified": "Thu, 27 Feb 2020 21:24:45 GMT",
    "ContentLength": 12748477,
    "ETag": "\"547717adb79902aaf28e20b0eed60cb0-2\"",
    "VersionId": "null",
    "ContentType": "binary/octet-stream",
    "Metadata": {}
}
$ aws s3api head-object --bucket $bucket --key dir2/file
{
    "AcceptRanges": "bytes",
    "LastModified": "Thu, 27 Feb 2020 21:24:49 GMT",
    "ContentLength": 12748477,
    "ETag": "\"33391fd40656ffa4eb9dfa2be9336202-2\"",
    "VersionId": "null",
    "ContentType": "binary/octet-stream",
    "Metadata": {}
}
$ rclone -v --checksum --ignore-size copy s3:$bucket/dir s3:$bucket/dir2
2020/02/27 21:25:24 INFO  : S3 bucket $bucket path dir2: Waiting for checks to finish
2020/02/27 21:25:24 NOTICE: S3 bucket $bucket path dir2: --checksum is in use but the source and destination have no hashes in common; falling back to --size-only
2020/02/27 21:25:24 INFO  : S3 bucket $bucket path dir2: Waiting for transfers to finish
2020/02/27 21:25:24 INFO  :
Elapsed time:         0.0s
```

It should have transferred (IMHO even if --ignore-size is not specified, but this PR is about at least honoring --ignore-size).

Hopefully this helps!

> I don't understand why you would use --ignore-size - it is quite an esoteric option.
> 
> In general rclone prefers not to keep copying objects so you can run the command multiple times.

I must say I'm a bit surprised by that. I agree that --ignore-size is an esoteric option, but possibly not for the same reasons: I believe it should be the default. And as a user I expect rclone by default to prioritize copying the correct files before optimizing transfer, providing optimization opt-ins for people who have additional knowledge of their data (for instance if they know size changes over time, they can go with size comparison). In other words, overoptimizing to the point of incorrectness seems very weird to me.

tiborvass:
If it helps, s3cmd does want to copy.

```
$ s3cmd cp -n s3://$bucket/dir s3://$bucket/dir2
copy: 's3://$bucket/dir' -> 's3://$bucket/dir2'
WARNING: Exiting now because of --dry-run
```

thaJeztah:
ping @ncw PTAL 🙏🤗

thaJeztah:
@ncw could you have a look to see if there's any objections against this change?

ivandeex:
I rebased the fix on master.
When I force-pushed, github said "PR has no new commits" and closed it.
:rage:

ivandeex:
@ncw

This PR is unfortunately broken by me (can be revived by https://github.com/ivandeex/rclone/commit/6066adf261904f25bbb904f953c13350ceeee0a2) and forgotten by authors.

However it serves as a good example of uncontrolled complexity (see https://github.com/rclone/rclone/issues/4793#issuecomment-733647201) of the ["equal" method](https://github.com/rclone/rclone/blob/v1.53.3/fs/operations/operations.go#L136) and around, which used to grow organically in response to the stream of [change detection](https://github.com/rclone/rclone/issues?q=label%3A%22change+detection%22) feature requests, often serving contradictory user requirements such as _consistency_ (this PR https://github.com/rclone/rclone/pull/4011#issuecomment-592186479), _cost savings_ (see https://github.com/rclone/rclone/issues/2889#issuecomment-508683225) and many others (sometimes [unexplained ;)](https://github.com/rclone/rclone/issues/4793)).

This in turn has led to a number of change-related CLI flags (see https://rclone.org/flags), some of them having negative (_ignore_) or even double-negative (_noignore_) names, which interact in exotic ways with 30+ supported backends (https://rclone.org/overview/) and their (in)ability to provide required traits.

On top of that we have _filtering_ flags and two comnands: `sync` vs `copy`. I feel that many users won't feel difference between them and documentation doesn't stress enough. In the end we might expect that _any_ behavior can be requested, even that that feels esoteric or illogical.

This probably calls for some brushing (https://github.com/rclone/rclone/issues/4793#issuecomment-733628218):
- open a dedicated issue with bullets below
- assign ordering priorities (weights?) to flags and conditions
- describe which flag wins if they are used in combination
- probably develop a syntax similar (in idea) to Unix PAM (see https://linux.die.net/man/5/pam.conf or https://docs.ansible.com/ansible/latest/collections/community/general/pamd_module.html#examples) or NSS directives (https://linux.die.net/man/5/nsswitch.conf)
- see whether existing flags can be described by the syntax
- make it as simple as possible (think okkam's razor)
- revise test coverage, make sure existing shell wrappers around rclone don't break
- add an simple enough automata in _equal_ to handle the logic
- add a combined flag like `--change-criteria +size,-hash,!modtime,whatever` with a syntax that covers exusting and esoteric cases
- improve docs

This seems like a big effort so I'm not sure.

ncw:
> However it serves as a good example of uncontrolled complexity (see [#4793 (comment)](https://github.com/rclone/rclone/issues/4793#issuecomment-733647201)) of the ["equal" method](https://github.com/rclone/rclone/blob/v1.53.3/fs/operations/operations.go#L136) and around, which used to grow _organically_ in response to the stream of [change detection](https://github.com/rclone/rclone/issues?q=label%3A%22change+detection%22) feature requests, often serving contradictory user requirements such as _consistency_ (this PR [#4011 (comment)](https://github.com/rclone/rclone/pull/4011#issuecomment-592186479)), _cost savings_ (see [#2889 (comment)](https://github.com/rclone/rclone/issues/2889#issuecomment-508683225)) and many others (sometimes [unexplained ;)](https://github.com/rclone/rclone/issues/4793)), blurring the line between "bug" ang "feature".

I think we could inventory the flags to see where we are at.

However I don't want to break backwards compatibility, so if we want to change stuff then we need to do so carefull!

The `equal` function has been reasonably stable, I think the only thing added to it recently is --refresh-times - which is a very specialised options.

User demand is how rclone has gained new features, though we need to be careful to push back on ones which aren't of general interest.

> This in turn has led to a number of change-related CLI flags (see https://rclone.org/flags), some of them having negative (_no-check_), conditional (_x-if-y_) or even double-negative (_noignore_) names, which interact in exotic ways with 30+ supported backends (https://rclone.org/overview#features) and their (in)ability to provide required traits.

Indeed. N-flags * N-providers * N-options-per-provider is a big number.

> On top of that we have _filtering_ flags and two unobviously different commands: `sync` vs `copy`. I feel that many users won't feel difference between them and documentation doesn't stress enough.

More docs for `sync` and `copy` would be appreciated. The `sync` page has lots of warnings on now as a result of user accidents.

> In the end we might expect that _**any** behavior can be requested_, even that that feels esoteric or illogical.

Yes, users want to get their stuff done, it is our job to keep rclone consistent!
 
> This probably calls for some brushing ([#4793 (comment)](https://github.com/rclone/rclone/issues/4793#issuecomment-733628218)):

I'll just copy the comment here

> Idea: pause adding new --no-ignore-x-if-prefer-y flags for now. Add examples to docs for existing flags, they cover 95% of known use cases. For remaining 5%, create an issue to design a flag like --change-criteria +time,-size,whatever.

I think that is a good plan to control the complexity.

> * open a dedicated _issue_ with bullets below
> * assign ordering priorities (weights?) to flags and conditions
> * describe which flag wins if they are used in combination
> * probably develop a syntax similar (in idea) to Unix PAM (see https://linux.die.net/man/5/pam.conf or https://docs.ansible.com/ansible/latest/collections/community/general/pamd_module.html#examples) or NSS directives (https://linux.die.net/man/5/nsswitch.conf)
> * (I started thinking about CSS rule weights but stopped myself ;))
> * see whether existing flags can be described by the new syntax
> * make it as simple as possible (think okkam's razor)
> * revise test coverage, make sure existing shell wrappers around rclone don't break
> * add a simple enough automata in _equal_ to handle the logic
> * add a combined flag like `--change-criteria +size,-hash,!modtime,?whatever` with a syntax reach enough to cover existing and esoteric cases
> * improve docs: add examples
> * improve docs: map the combined flag on existing flags
> * improve docs (other)
> 
> This seems like a big effort covering <5% of user base so I'm not sure.

Making a generalised solution will mean that advanced users will have somewhere to go when then need to get stuff done, but hopefully a much smaller subset of flags will cover the main use cases (eg --checksum, --size-only).

I like that idea. We could implement it then deprecate flags which we don't like. (You can hide flags so they still work but are hidden).

> OTOH there is a _generalization_ PR #3782 (multiwrite union) that has been a [success](https://github.com/rclone/rclone/pull/3782#issuecomment-601892715).

Great idea.

I especially like your moratorium on adding more complexity to what we've got already.

Do you want to make an issue, then we can think about all the things which might need to be controlled?

Issue #3997, opened 2020-02-22
Override TOTAL in configuration


tYYGH:
#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.51.0
- os/arch: linux/amd64
- go version: go1.13.7
```

#### What problem are you are trying to solve?

Cloud providers bill based on usage, and crossing a usage threshold can provoke an automatic switch to the upper (and costlier) offer.

#### How do you think rclone should be changed to solve that?

I’ve searched the issues for my problem, and my issue is exactly #1956. It is closed, though.
I totally agree with the NO-FIX resolution, since the cloud providers just don’t provide the necessary data.

However, apparently Rclone now has the feature, at least for those providers that do provide the information: I did some tests, and both FREE and TOTAL spaces are correctly reported for:
* Google Drive
* Koofr
* MS OneDrive

Strangely, even though both FREE and TOTAL are correctly reported with the `about` command, I get wrong results with the `mount` command for:
* Mega: I get a TOTAL of 148G instead of 50G, and 13G FREE insead of 4G;
* Yandex: while the TOTAL is correct (10G), the FREE space is reported as 1P.

So, what I suggest is adding the possibility to configure an optional per-remote TOTAL that would be used for that remote in place of whatever is reported by the remote.
If the TOTAL is such configured for a remote, then the FREE space would be the simple calculation `TOTAL−USED=FREE`.

tYYGH:
Although I’m just discovering rclone, and I don’t know the Go language, I’m trying to see if I’d be able to bring the change myself.

I think I’m trying to get some bearings, but I’d be glad to get some help.
If I understand correctly:
* Based on `fs.go` found in `cmd/cmount/` and `cmd/mount/`, I think that the computation I’m interested in happens in `VFS.Statfs`, which itself (`vfs/vfs.go`) is using the `About` “feature” of the underlying backend.
* In `mega.go` for example, `features` is mostly filled by `Fill` in `fs/fs.go`, which sets `About` to the `About` function found in `mega.go` as an implementation of the `Abouter` interface.
* `fs/config/config.go` is the handling of the configuration. I appended that to the `const {…}` block:
```go
	// ConfigForceTotalFsSize is an optional override of the Total size reported by a filesystem
	ConfigForceTotalFsSize = "fs_total_bytes"
```
Then I should just have to call `FileGet(_a section_, ConfigForceTotalFsSize)` at the right places (i.e. for each backend)…

[...]

PR #4020, opened 2020-03-01
webdav: report full and consistent usage with `about`

tYYGH:

When running `rclone about webdav-remote:`, a typical answer would look like:

```
Total:   10G
Used:    90.119M
```

In such a situation, we have the data to compute the Free value too, so it should be done.
Besides, the current code does not allow either Free or Used to be equal to 0 (meaning “full” or “empty”): in this situation, it may happen that no value is reported. This is also fixed.

Yes: https://github.com/rclone/rclone/issues/3997

#### Checklist


ncw:

Will merge now.

PR #4022, opened 2020-03-01
s3: `rclone about` gives Used space (costly: needs full recursive listing)

tYYGH:

I needed a report on Used space for S3 remotes, so here it is.
Implementation relies on pre-existing code for ListObjectsInput creation, and listing of the buckets.
Thus, I chose not to use “V2” versions of API calls, instead aligning with existing API calls.

Not really, although it is still linked to my ultimate goal exposed here: https://github.com/rclone/rclone/issues/3997

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1329,19 +1329,8 @@ func (f *Fs) updateRegionForBucket(bucket string) error {
 	return nil
 }
 
-// listFn is called from list to handle an object.
-type listFn func(remote string, object *s3.Object, isDirectory bool) error
-
-// list lists the objects into the function supplied from
-// the bucket and directory supplied.  The remote has prefix
-// removed from it and if addBucket is set then it adds the
-// bucket to the start.
-//
-// Set recurse to read sub directories
-func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBucket bool, recurse bool, fn listFn) error {
-	if prefix != "" {
-		prefix += "/"
-	}
+// Return a ListObjectsInput structure, and a boolean indicating whether listings are URL-encoded.
+func (f *Fs) createListObjectsInput(bucket *string, directory string, recurse bool) (s3.ListObjectsInput, bool) {
 	if directory != "" {
 		directory += "/"
 	}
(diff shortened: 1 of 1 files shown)


ncw:
I think your use case for this is making `df` correct in `rclone mount`.

This essentially duplicates the `rclone size` command but in a backend specific way doesn't it?

I think a better way of doing this would be for a flag to the VFS layer which calls `rclone size` internally to provide the total amount.

This approach is very expensive though - many people have millions of objects in their s3... so we don't want call it by default... I'm not sure whether it is a good idea!

tYYGH:
> I think your use case for this is making `df` correct in `rclone mount`.

Exactly.

> This essentially duplicates the `rclone size` command but in a backend specific way doesn't it?

I did not know of `rclone size` but the name suggests exactly what I did, indeed.

> I think a better way of doing this would be for a flag to the VFS layer which calls `rclone size` internally to provide the total amount.
> 
> This approach is very expensive though - many people have millions of objects in their s3... so we don't want call it by default... I'm not sure whether it is a good idea!

Right. I’ve been thinking about that too, and I thought, that maybe this feature should even be limited (recommendations in the documentation) to usage in restricted conditions: 1° with a cache, and 2° with exclusive access, so that “±files” can be directly translated into “±bytes” onto the cached value of “Used”, at least until the next refresh of the cached Used size.

Anyway, all of that is all the more reason to go the way you suggest, with a VFS flag to use `rclone size` if/when the user knows what they’re doing.

----------

Considering on the one hand the requirement of contribution rules to have a single commit for each contribution, and on the other hand the fact that amending an already-pushed commit is bad practice, I suggest that we close this pull request, and that I create a new one, when I am ready with the alternate solution. Is that OK?

ncw:
OK agreed!

> Considering on the one hand the requirement of contribution rules to have a single commit for each contribution, and on the other hand the fact that amending an already-pushed commit is bad practice, I suggest that we close this pull request, and that I create a new one, when I am ready with the alternate solution. Is that OK?

I don't mind if you force push branches or add more than one commit. I do want one feature per commit though!

Anyway I'll close this one and you can open a new one since it is a quite unrelated implementation.

PR #4024, opened 2020-03-02
crypt: support timestamped filenames from --b2-versions

### LINKED ISSUE #1627: --b2-versions doesn't work for crypt backed by B2
Hidden file versions are not listed when I do `rclone ls --b2-versions` on a crypt bucket that is backed by a B2 bucket. This is highly inconvenient, to say the least. :-/

wlritchi:
Relies on identifying the version string that backend/b2 adds to the
filename. There are pathological cases where an obfuscated filename
could end in what looks like a version string purely by chance. In such
a case, this code will incorrectly leave that portion of the filename
obfuscated. There's not really a clean fix for that, other than keeping
track of versions out-of-band instead of in the filename string.

Fixes #1627

Background: Backblaze B2 supports multiple versions of a given filename. rclone supports this behind `--b2-versions` for read-only access, by appending a timestamp suffix to all but the most recent version of a file. The crypt backend supports encrypted or obfuscated filenames, but fails to decrypt them when a timestamp suffix has been added.

This change updates the crypt backend to identify when a file has a version string added to it. If it does, the version string will be stripped off before decrypting the filename, and added back to the decrypted name. This allows `--b2-versions` to be used with crypt remotes backed by B2.

This issue was discussed in #1627, where a fix broadly similar to this was suggested.

#### Checklist

- *Note:* I have added unit tests, but I'm not sure what the best way is to add integration tests for a crypt remote backed by b2. I imagine we want to add some that cover this use case.

### DIFF
--- a/backend/b2/api/types.go
+++ b/backend/b2/api/types.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -64,6 +65,16 @@ func (t *Timestamp) UnmarshalJSON(data []byte) error {
 }
 
 const versionFormat = "-v2006-01-02-150405.000"
+
+var /* const */ versionRegexp = regexp.MustCompile("-v\\d{4}-\\d{2}-\\d{2}-\\d{6}-\\d{3}")
+
+// HasVersion returns true if it looks like the passed filename has a timestamp on it.
+//
+// Note that the passed filename's timestamp may still be invalid even if this
+// function returns true.
+func HasVersion(remote string) bool {
+	return versionRegexp.MatchString(remote)
+}
 
 // AddVersion adds the timestamp as a version string into the filename passed in.
 func (t Timestamp) AddVersion(remote string) string {
--- a/backend/b2/api/types_test.go
+++ b/backend/b2/api/types_test.go
@@ -36,6 +36,24 @@ func TestTimestampUnmarshalJSON(t *testing.T) {
 	assert.Equal(t, (time.Time)(t1), (time.Time)(tActual))
 }
 
+func TestTimestampHasVersion(t *testing.T) {
+	for _, test := range []struct {
+		in       string
+		expected bool
+	}{
+		{"potato.txt", false},
+		{"potato", false},
+		{"", false},
+		{"potato-v1970-01-01-010101-123.txt", true},
+		{"potato-v2001-02-03-040506-123", true},
+		{"-v2001-02-03-040506-123", true},
+		{"-v9999-99-99-999999-999", true},
+	} {
+		actual := api.HasVersion(test.in)
+		assert.Equal(t, test.expected, actual, test.in)
+	}
+}
+
 func TestTimestampAddVersion(t *testing.T) {
 	for _, test := range []struct {
 		t        api.Timestamp
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -15,6 +15,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/pkg/errors"
+	"github.com/rclone/rclone/backend/b2/api"
 	"github.com/rclone/rclone/backend/crypt/pkcs7"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
@@ -500,6 +501,19 @@ func (c *cipher) decryptFileName(in string) (string, error) {
 		if !c.dirNameEncrypt && i != (len(segments)-1) {
 			continue
 		}
+
+		hasVersion := false
+		var t api.Timestamp
+		if i == (len(segments)-1) && api.HasVersion(segments[i]) {
+			var s string
+			t, s = api.RemoveVersion(segments[i])
+			// RemoveVersion can fail, in which case it returns segments[i]
+			if s != segments[i] {
+				segments[i] = s
+				hasVersion = true
+			}
+		}
+
 		if c.mode == NameEncryptionStandard {
 			segments[i], err = c.decryptSegment(segments[i])
 		} else {
@@ -509,6 +523,10 @@ func (c *cipher) decryptFileName(in string) (string, error) {
 		if err != nil {
 			return "", err
 		}
+
+		if hasVersion {
+			segments[i] = t.AddVersion(segments[i])
+		}
 	}
 	return strings.Join(segments, "/"), nil
 }
@@ -517,10 +535,18 @@ func (c *cipher) decryptFileName(in string) (string, error) {
 func (c *cipher) DecryptFileName(in string) (string, error) {
 	if c.mode == NameEncryptionOff {
 		remainingLength := len(in) - len(encryptedSuffix)
-		if remainingLength > 0 && strings.HasSuffix(in, encryptedSuffix) {
-			return in[:remainingLength], nil
+		if remainingLength == 0 || !strings.HasSuffix(in, encryptedSuffix) {
+			return "", ErrorNotAnEncryptedFile
 		}
-		return "", ErrorNotAnEncryptedFile
+		decrypted := in[:remainingLength]
+		if api.HasVersion(decrypted) {
+			_, unversioned := api.RemoveVersion(decrypted)
+			if unversioned == "" {
+				return "", ErrorNotAnEncryptedFile
+			}
+		}
+		// Leave the version string on, if it was there
+		return decrypted, nil
 	}
 	return c.decryptFileName(in)
 }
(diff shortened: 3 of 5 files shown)


backend/b2/api/types.go
```diff
@@ -64,6 +65,11 @@ func (t *Timestamp) UnmarshalJSON(data []byte) error {
 }
 
 const versionFormat = "-v2006-01-02-150405.000"
+var /* const */ versionRegexp = regexp.MustCompile("-v\\d{4}-\\d{2}-\\d{2}-\\d{6}-\\d{3}")
+
+func HasVersion(remote string) bool {
```
golangcibot: exported function `HasVersion` should have comment or be unexported (from `golint`)

backend/b2/api/types.go
```diff
@@ -65,6 +66,16 @@ func (t *Timestamp) UnmarshalJSON(data []byte) error {
 
 const versionFormat = "-v2006-01-02-150405.000"
 
+var /* const */ versionRegexp = regexp.MustCompile("-v\\d{4}-\\d{2}-\\d{2}-\\d{6}-\\d{3}")
```
ncw: Is this `/* const */` annotation understood by some tool? Not seen it before?

wlritchi: Honestly, it was the first suggestion I found on StackOverflow for a regexp constant, and I got the sense that it was more from the answerer's personal style guide, as a way to tell humans about the intent. I haven't worked with Go before so I have no stylistic opinions of my own. Is there something you'd prefer to indicate that kind of intent? Or just drop it and let the context explain it for humans?

wlritchi:
I'm inclined to agree that `lib/version` would be a better design; I'll look into refactoring the logic. Apart from that, I'm still a little hesitant on whether or not string manipulation of the filename is robust enough to use widely. I *think* it's alright when the backend is read-only, but definitely will have unexpected behaviour with write operations. I'd want some way to assure that the version tagging logic forces all version-supporting backends to read-only.

backend/b2/api/types.go
ncw: OK! I'd probably drop it if it doesn't have a meaning to a tool.

ncw:
> I'm inclined to agree that `lib/version` would be a better design; I'll look into refactoring the logic.

Great! Let me know if you need help.

> Apart from that, I'm still a little hesitant on whether or not string manipulation of the filename is robust enough to use widely. I _think_ it's alright when the backend is read-only, but definitely will have unexpected behaviour with write operations. I'd want some way to assure that the version tagging logic forces all version-supporting backends to read-only.

Hmm, that might be an argument for this being a flag, eg `--crypt-versions` then there is no mistaking the users intent.

That could set the crypt read only if we wanted too.

fdalvi:
Hello,

Firstly, thanks for your work on this @wlritchi and @ncw. I see that this PR has been inactive for a while, I would love to jump in and finish what @wlritchi had started if they don't have the cycles for this anymore.

While we wait for the answer, I thought it would be a good idea anyways to formalize what it means to refactor the versioning into its own library. Currently, three methods will be exposed by this library: `Add`, `Match` and `Remove`. `Add` needs a `versionFormat` string as required by the `time` library, `Match` would need a `versionRegexp` as with the current implementation, while `Remove` can work with either the format string or the regexp string. What would the best way be to pass these on to the library functions? There are many options:

1. `Add` and `Remove` take a filename string + a time format string, `Match` takes filename string + regexp string
2. `Add` takes a filename string + a time format string, `Match` and `Remove` take filename string + regexp string
3. `Add`, `Match` and `Remove` all take a time format strings, and we programmatically derive a regexp string or different matching code in `Match`/`Remove` (`Remove` already uses code that needs time format string alone)
4. `Add`, `Match` and `Remove` take a `versionDefinition` struct (which contain both a format string and a regexp matcher)

The advantage of **3** and **4** is a unified interface for all three functions; the disadvantage of **3** is that the automatic regexp generation/matching might break in the future with more complicated time string formats (lets say something that includes a three letter day of the week segment)

Let me know what you all think is a good way forward, and if I can work on this!

fdalvi:
I took a closer look and realized two things that I missed earlier:

- `Add` would need to take the timestamp as well along with the filename
- The "version timestamp" is something thats added by rclone rather than a backend, so we can hardcode both the `formatString` and the `regexpString` regardless of which backend we are using, so we do not need to pass either of these to any of the functions!

Let me know if I missed something.

ncw:
@fdalvi great if you could continue this - I suggest you put your work in a separate commit so we can get the attributions right!

> * The "version timestamp" is something thats added by rclone rather than a backend, so we can hardcode both the `formatString` and the `regexpString` regardless of which backend we are using, so we do not need to pass either of these to any of the functions!

Yes the idea of the library is to standardise the format strings so we don't need to pass them in anywhere.

I'd like to implement versions for the s3 backend so it would be great to re-use it there.

I see it the interface being quite simple

```go
// Add returns fileName modified to include t as the version
func Add(fileName string, t time.Time) string

// Remove returns a modified fileName without the version string and the time it represented
// If the fileName did not have a version then time.Time{} is returned along with an unmodified fileName
func Remove(fileName string) (time.Time, string)

// Match returns true if the fileName has a version string
func Match(fileName string) bool
```

ivandeex:
@fdalvi friendly ping

ivandeex:
@wlritchi This PR is abandoned for almost a year :(

ivandeex:
This PR is abandoned and needs a new maintainer.

domyd:
I'd like to give this a shot. I'm not very familiar with go, but this looks simple enough to do, and it's a feature I'm actively interested in using.

I'm not sure on how to update this PR though, since I can't push code here. Would I need to open up a new PR (containing this PR's commit, and my changes on top) or is there a Github feature that somehow allows this?

ivandeex:
@domyd 

Your effort is very appreciated!
The issue with backblaze versioning behind crypt is frequently mentioned by users, but we don't have a maintainer.

Obviously you cannot push to this PR, but you can open your _own_ rclone PR based on the code here.

_Alternatively_, you can open a pull request to the _wlritchi's fork_ that this PR stems from. I did so once: opened PR-for-PR https://github.com/denisneuling/rclone/pull/1 then was waiting for author to accept my changes until they appeared on the _rclone PR_ https://github.com/rclone/rclone/pull/3533#issuecomment-736681544. This way you'll have to wait for _wlritchi_ too. Given he hadn't replied to pings for quite a time I think it's gonna take long. I _don't_ recommend it.

The following steps are for your _OWN_ way:

1. Fork https://github.com/rclone/rclone to your github account by using the `Fork` button in the github web interface (on the top right)

2. Clone _your_ fork down to your computer.
```
git clone git@github.com:domyd/rclone
```

3. Add reference for the wlritchi's work in your local codebase and start working from there
```
git remote add wlritchi https://github.com/wlritchi/rclone
git fetch wlritchi
git checkout -b  b2-crypt-domyd wlritchi/crypt-support-b2-versions
```

4. Install latest golang 1.16 from https://golang.org/dl/ or use any version from 1.14 up if it's already installed

5. Familiarize yourself:
    - Take a look at _rclone backblaze_ docs: https://rclone.org/b2/#versions
    - Make sure you have read the discussion thread of this PR #4024 and its connected issue #1627
    - Try to understand what was done by your predecessors: https://github.com/rclone/rclone/pull/4024/files

Now you can commit changes on the `b2-crypt-domyd` branch!

----------------------

The rclone development workflow is described at https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request but instead of `go get -u github.com/rclone/rclone` you should do the steps above (because you need a concrete PR instead of rclone HEAD).
I wrote a short intro here https://github.com/rclone/rclone/pull/4075#issuecomment-729060366

1. After your changes, make sure that rclone builds locally:
```
go build .
```

2. Run `b2` and `crypt` unit tests locally as described here https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#testing
```
go test -v -count=1 ./backend/b2/...
go test -v -count=1 ./backend/crypt/...
```
Ensure that `TestB2` remote is configured in your `rclone.conf`. See logs and fix bugs, if any.

3. Periodically check code style, run the following and see what needs tweaking
```
gofmt -w ./backend/b2/*.go ./backend/crypt/*.go
git status; git diff
```
Then analyze diff and commit

4. When you are satisfied with your changes and the unit test passes, you will want to run integration tests as explained here https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#testing
```
go run ./fstests/test_all -backends b2,crypt
```

When you are ready, push your work on _your fork_ of rclone:
```
git push -u origin b2-crypt-domyd
```

Open your browser, login to github and go to your forked repo. You'll see a green button to open a pull request (it disappears 12-20 hours after the push, but then you can go to the branch and click the PR button there).
Name the PR like `crypt: support timestamped filenames from --b2-versions (attempt 2)` and fill in the form.

domyd:
@ivandeex Thank you very much for the detailed guide! I've opened up a new PR with the requested changes.

ncw:
Thanks @domyd - I'll close this PR now since the new one is under way.

PR #4029, opened 2020-03-02
Correct exitcode on Transfer Limit Exceeded flag. Fixes #3203

### LINKED ISSUE #3203: Incorrect Exit Code for Transfer Limit Exceeded

Incorrect Exit Code by rclone when the specified max-transfer limit is reached. As per the docs @ https://rclone.org/docs/#exit-code, the exit-code should be `8` instead of `7` which indicates a fatal error.

This is just my guess from a brief overview of the related code, but I believe that the error stems from https://github.com/ncw/rclone/blob/7d70e9266446901e93a888212edd06765821047b/fs/sync/sync.go#L178-L183 where the actual error is added to which fails the eventual comparison test for the error type.

```
rclone v1.47.0-062-g99bc013c-beta
- os/arch: linux/amd64
- go version: go1.12.5
```

`Debian 9 64-bit`

`Google Drive`

`rclone sync Drive: DriveCopy: --tpslimit=5 --tpslimit-burst=50 --max-transfer=700G`

Relevant Lines:

```
2019/05/16 07:54:40 ERROR : Cancelling sync due to fatal error: Max transfer limit reached as set by --max-transfer
2019/05/16 07:54:40 INFO  : Google drive root '': Waiting for checks to finish
2019/05/16 07:54:40 INFO  : Google drive root '': Waiting for transfers to finish
2019/05/16 07:54:41 INFO  : o6i2cgd36bmi2nva56ubo4a268/i28c365mh5iu29bbbq5h5bf960/q58f203e7ps6gblq0fu0f36954/o6omqo0u6b7sb4rjhdjf7qsmtg/dajheiohfhb8b7vag018ke
paucfq3fau0kb3l75fqstcii04ch32jcj7p79rnv86jd6v4v8atj3p81pcq460pge65ut8p7n02ptt7eo: Copied (server side copy)
2019/05/16 07:54:41 ERROR : Fatal error received - not attempting retries
2019/05/16 07:54:41 Failed to sync: Max transfer limit reached as set by --max-transfer
```

anuar45:

Send correct exit code(8) when transfer is limited by --max-transfer flag.

Issue #3203

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -7,6 +7,7 @@ package cmd
 // would probably mean bringing all the flags in to here? Or define some flagsets in fs...
 
 import (
+	stderrors "errors"
 	"fmt"
 	"log"
 	"math/rand"
@@ -442,7 +443,7 @@ func resolveExitCode(err error) {
 		os.Exit(exitCodeFileNotFound)
 	case unwrapped == errorUncategorized:
 		os.Exit(exitCodeUncategorizedError)
-	case unwrapped == accounting.ErrorMaxTransferLimitReached:
+	case stderrors.Is(err, accounting.ErrorMaxTransferLimitReached):
 		os.Exit(exitCodeTransferExceeded)
 	case fserrors.ShouldRetry(err):
 		os.Exit(exitCodeRetryError)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -7,7 +7,6 @@ package cmd
 // would probably mean bringing all the flags in to here? Or define some flagsets in fs...
 
 import (
-	stderrors "errors"
 	"fmt"
 	"log"
 	"math/rand"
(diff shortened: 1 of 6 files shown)


cmd/cmd.go
```diff
@@ -442,7 +443,7 @@ func resolveExitCode(err error) {
 		os.Exit(exitCodeFileNotFound)
 	case unwrapped == errorUncategorized:
 		os.Exit(exitCodeUncategorizedError)
-	case unwrapped == accounting.ErrorMaxTransferLimitReached:
+	case stderrors.Is(err, accounting.ErrorMaxTransferLimitReached):
```
ncw: We can't use `errors.Is` yet as we still need to support go1.11

However you shouldn't need it as we use `fserrors.Cause` which does a pretty similar thing.

fs/fserrors/error.go
```diff
@@ -116,6 +116,10 @@ func (err wrappedFatalError) Cause() error {
 	return err.error
 }
 
+func (err wrappedFatalError) Unwrap() error {
```
ncw: If we aren't using `errors.Is` we don't need the `Unwrap` as we have `Cause` which is what pkg/errors uses.

ncw:

I'll merge it.

PR #4038, opened 2020-03-04
backend/s3: Replace Deprecated session.New()

alrs:
There is a deprecation warning on `github.com/aws/aws-sdk-go/aws/session.New()`
>  Deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.

This PR removes the deprecated `New()` function and replaces it with `NewSession()`.

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1039,6 +1039,12 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 	def := defaults.Get()
 	def.Config.HTTPClient = lowTimeoutClient
 
+	// start a new AWS session
+	awsSession, err := session.NewSession()
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "NewSession")
+	}
+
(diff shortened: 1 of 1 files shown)


backend/s3/s3.go
```diff
@@ -1039,6 +1039,12 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 	def := defaults.Get()
 	def.Config.HTTPClient = lowTimeoutClient
 
+	// start a new AWS session
+	awsSession, err := session.NewSession()
+	if err != nil {
+		return nil, nil, fmt.Errorf("NewSession: %v", err)
```
ncw: In rclone we are using the pkg/errors package for error wrapping so that should be

```go
return nil, nil, errors.Wrap(err, "NewSession")
```

alrs: Fixed, thanks.

ncw:
That looks great now thanks - I'll merge it now.

PR #4041, opened 2020-03-07
Added maximum chunk size limit warning on docs

Dibbyo456:

If chunk size is more than 250M (262,144,000 bytes) then API throws the following error: 

`Microsoft.SharePoint.Client.InvalidClientQueryException: The request message is too big. The server does not allow messages larger than 262144000 bytes.`

No.

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -989,13 +989,14 @@ func (f *Fs) About(ctx context.Context) (*fs.Usage, error) {
 		return nil, errors.Wrap(err, "about call failed")
 	}
 	usage := &fs.Usage{}
-	if q.Available != 0 || q.Used != 0 {
-		if q.Available >= 0 && q.Used >= 0 {
-			usage.Total = fs.NewUsageValue(q.Available + q.Used)
-		}
-		if q.Used >= 0 {
-			usage.Used = fs.NewUsageValue(q.Used)
-		}
+	if q.Used >= 0 {
+		usage.Used = fs.NewUsageValue(q.Used)
+	}
+	if q.Available >= 0 {
+		usage.Free = fs.NewUsageValue(q.Available)
+	}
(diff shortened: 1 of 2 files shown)


ncw:
Thanks for doing this :-)

Unfortunately those docs are auto generated - the source that needs patching is here

https://github.com/rclone/rclone/blob/dfc7215bf9b7840f438d006bf7b399974f5ebf42/backend/onedrive/onedrive.go#L226-L230

Dibbyo456:
Done.

ncw:
Sorry for the delay!

Can you put exactly the same text in onedrive.go as onedrive.md otherwise when I come to make a release the text in onedrive.md will be overwritten by that in onedrive.go

Thanks!

Dibbyo456:
> Can you put exactly the same text in onedrive.go as onedrive.md otherwise when I come to make a release the text in onedrive.md will be overwritten by that in onedrive.go

Sorry to say but I am little confused here. The last time I wanted to change some docs and you wanted to me update `onedrive.md` on #3617. So why `onedrive.go` instead of `onedrive.md` this time?

ncw:
> Sorry to say but I am little confused here. The last time I wanted to change some docs and you wanted to me update `onedrive.md` on #3617. So why `onedrive.go` instead of `onedrive.md` this time?

Sorry for the confusion.

Part of onedrive.md is auto generated from onedrive.go - in particular the part between these lines

```
<!--- autogenerated options start - DO NOT EDIT, instead edit fs.RegInfo in backend/onedrive/onedrive.go then run make backenddocs -->
...
<!--- autogenerated options stop -->
```

This means that `rclone` has the docs within itself so you can type `rclone help backend onedrive` and those docs are shown during `rclone config`

The rest of it is edited by humans which was the section you added last time.

Ideally I'd use an include file but I don't think markdown supports include files :-(

PR #4052, opened 2020-03-10
OneDrive maximum chunk size limit warning

Dibbyo456:
If chunk size is more than 250M (262,144,000 bytes) then API throws the following error: 

`Microsoft.SharePoint.Client.InvalidClientQueryException: The request message is too big. The server does not allow messages larger than 262144000 bytes.`


ncw:
Perfect , thank you.

Apologies for the confusion.

I'll merge it now :-)

PR #4054, opened 2020-03-10
fs: Add configurable matching criteria for --track-renames

### LINKED ISSUE #2721: Make - - track-renames work with encrypted remotes
Hi, 
As title said can you update - - track-renames to work also on encrypted remotes?
Thx

### LINKED ISSUE #3696: Configurable matching criteria for --track-renames
Hello,

In cases of many new or renamed files, --track-renames can be very slow due to it's need to checksum all of these unknown items from a local disk. In my use case it would be acceptable to forgo checksums and limit the rename matching to mod-time and file size. This has been discussed on the forum and an extra flag such as --track-renames-strategy has been proposed.

Please see the discussion here:
https://forum.rclone.org/t/can-not-use-track-renames-when-syncing-to-and-from-crypt-remotes/11063/8

Thanks.

quexten:
This commit adds the --track-using-modtime flag which makes
rclone match renames using the file modification time instead
of file hashes. This enables support for tracking renames in
encrypted remotes. For two files to match, their modification
times must be within 1 millisecond of each other to account
for remote modification time inaccuracies.

Fixes #3696
Fixes #2721

This commit adds the `--track-using-modtime` flag which makes
rclone match renames using file modification time. It makes
`--track-renames` work with encrypted destinations.

It was discussed [here](https://forum.rclone.org/t/can-not-use-track-renames-when-syncing-to-and-from-crypt-remotes/11063).

#### Checklist

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -61,6 +61,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.BoolVarP(flagSet, &deleteAfter, "delete-after", "", false, "When synchronizing, delete files on destination after transferring (default)")
 	flags.Int64VarP(flagSet, &fs.Config.MaxDelete, "max-delete", "", -1, "When synchronizing, limit the number of deletes")
 	flags.BoolVarP(flagSet, &fs.Config.TrackRenames, "track-renames", "", fs.Config.TrackRenames, "When synchronizing, track file renames and do a server side move if possible")
+	flags.BoolVarP(flagSet, &fs.Config.TrackUsingModTime, "track-using-modtime", "", fs.Config.TrackUsingModTime, "When tracking renames, use the modtime for comparison instead of the file hash")
 	flags.IntVarP(flagSet, &fs.Config.LowLevelRetries, "low-level-retries", "", fs.Config.LowLevelRetries, "Number of low level retries to do.")
 	flags.BoolVarP(flagSet, &fs.Config.UpdateOlder, "update", "u", fs.Config.UpdateOlder, "Skip files that are newer on the destination.")
 	flags.BoolVarP(flagSet, &fs.Config.UseServerModTime, "use-server-modtime", "", fs.Config.UseServerModTime, "Use server modified time instead of object metadata")
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -61,7 +61,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.BoolVarP(flagSet, &deleteAfter, "delete-after", "", false, "When synchronizing, delete files on destination after transferring (default)")
 	flags.Int64VarP(flagSet, &fs.Config.MaxDelete, "max-delete", "", -1, "When synchronizing, limit the number of deletes")
 	flags.BoolVarP(flagSet, &fs.Config.TrackRenames, "track-renames", "", fs.Config.TrackRenames, "When synchronizing, track file renames and do a server side move if possible")
-	flags.StringVarP(flagSet, &fs.Config.TrackRenamesStrategy, "track-renames-strategy", "", fs.Config.TrackRenamesStrategy, "When tracking renames, use the modtime for comparison instead of the file hash")
+	flags.StringVarP(flagSet, &fs.Config.TrackRenamesStrategy, "track-renames-strategy", "", fs.Config.TrackRenamesStrategy, "Strategies to use when synchronizing using track-renames hash|modtime")
 	flags.IntVarP(flagSet, &fs.Config.LowLevelRetries, "low-level-retries", "", fs.Config.LowLevelRetries, "Number of low level retries to do.")
 	flags.BoolVarP(flagSet, &fs.Config.UpdateOlder, "update", "u", fs.Config.UpdateOlder, "Skip files that are newer on the destination.")
 	flags.BoolVarP(flagSet, &fs.Config.UseServerModTime, "use-server-modtime", "", fs.Config.UseServerModTime, "Use server modified time instead of object metadata")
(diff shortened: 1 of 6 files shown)


quexten:
As a note, for the test I just re-used the `--track-renames` test and modified it so it uses the new flag. 
I could use some help with the failing  build on the "modules_race" pipeline, as it builds fine on my repository.

ncw:
Thanks for taking a look at this and apologies for the delay looking at it.

I quite like the idea of making this a bit more general - see the comment here: https://forum.rclone.org/t/can-not-use-track-renames-when-syncing-to-and-from-crypt-remotes/11063/8

If we implemented a format string effectively for the rename hash then it would probably end up simplifying the code.

I note that the modtime comparison should be done according to the precision specified by `fs.GetModifyWindow` with the source and the dest Fs as parameters. Storing the modtime rendered as a unix time, divided by the precision (perhaps again divided by 2 as the comparison normally done is +/- precision) would be sufficient for matching I think.

fs/sync/sync.go
```diff
@@ -557,20 +560,44 @@ func (s *syncCopyMove) srcParentDirCheck(entry fs.DirEntry) {
...
+		}
+
+		renameID += fmt.Sprintf(",%s", hash)
 	}
-	if hash == "" {
-		return ""
+
+	if isUsingRenameStrategy("modtime", renameStrategy) {
+		modTime := obj.ModTime(s.ctx).Unix() / precision.Nanoseconds() / 2
+		renameID += fmt.Sprintf(",%s", modTime)
```
golangcibot: printf: Sprintf format %s has arg modTime of wrong type int64 (from `govet`)

quexten:
@ncw 
Thanks for the response!
I changed the parameter to now be a comma-separated list of requested strategies and moved the functions for generating the hash-string and the modtime string into 1 function which now just checks which strategies are requested and generates a comma separated string. So for the strategies hash,modtime it would be `size,hash,modtime`. I also changed the modtime comparison to use the precision as you suggested.
Let me know what you think.

fs/sync/sync.go
```diff
@@ -557,20 +560,44 @@ func (s *syncCopyMove) srcParentDirCheck(entry fs.DirEntry) {
...
+			return ""
+		}
+
+		renameID += fmt.Sprintf(",%s", hash)
 	}
-	if hash == "" {
-		return ""
+
+	if isUsingRenameStrategy("modtime", renameStrategy) {
+		modTime := obj.ModTime(s.ctx).Unix() / precision.Nanoseconds() / 2
```
ncw: If `precision.Nanoseconds()` is small (eg it is 1 on local unix drives) then it would be nice to include the nanoseconds too in the comparison.

So the calculation really wants to be something like this (with rounding)

```go
modTime := obj.ModTime(s.ctx)
divisor := precision.Nanoseconds() / 2
rounding := divisor / 2
if rounding > 0 {
    rounding--
}
timeHash := (modTime.Unix() * time.Nanosecond + modTime.Nanosecond() + rounding ) / divisor
```

Note that `modTime.Unix() * time.Nanosecond` potentially overflows an int64. It will do so when modTime.Unix > 2**63/1E9 = the year 2262 so that is probably OK... If it isn't we'll have to break out the large integer maths library.

Note also we should probably be rejecting "modtime" if the `Precision == fs.ModTimeNotSupported`

quexten: Okay, I have pushed another commit which resolves the issues you have pointed out. For the last suggestion, rejecting the modtime, I have handled it as it was previously handled when hashes are not available in either remote. So if one remote doesn't support modtime, (the window is ModTimeNotSupported) then `s.trackRenames` will be set to false.

fs/config/configflags/configflags.go
```diff
@@ -61,6 +61,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.BoolVarP(flagSet, &deleteAfter, "delete-after", "", false, "When synchronizing, delete files on destination after transferring (default)")
 	flags.Int64VarP(flagSet, &fs.Config.MaxDelete, "max-delete", "", -1, "When synchronizing, limit the number of deletes")
 	flags.BoolVarP(flagSet, &fs.Config.TrackRenames, "track-renames", "", fs.Config.TrackRenames, "When synchronizing, track file renames and do a server side move if possible")
+	flags.StringVarP(flagSet, &fs.Config.TrackRenamesStrategy, "track-renames-strategy", "", fs.Config.TrackRenamesStrategy, "When tracking renames, use the modtime for comparison instead of the file hash")
```
ncw: The help for this is outdated

quexten: Okay, updated the help in a new commit.

ncw:
That looks great - thank you very much :-)

I'll merge it now.

Mushin:
Will the --track-renames-strategy flag obey --modify-window ?

I'm trying to rename many files that will then update on an encrypted Gdrive. The mod-times at some point got changed, and are off by about a second, so I've been using --size-only. 

So I thought if I use the --track-renames with --modify-window, I could still rename the files server side?

ncw:
> Will the --track-renames-strategy flag obey --modify-window ?

Yes it will.

However I discovered a bug in the implementation after the integration tests failed: 93f5125f51866791b1a3f0d0e316123967db1453

So you might want to try [the latest beta](https://beta.rclone.org) in 15-30 mins.

Mushin:
> > Will the --track-renames-strategy flag obey --modify-window ?
> 
> Yes it will.
> 
> However I discovered a bug in the implementation after the integration tests failed: [93f5125](https://github.com/rclone/rclone/commit/93f5125f51866791b1a3f0d0e316123967db1453)
> 
> So you might want to try [the latest beta](https://beta.rclone.org) in 15-30 mins.

Great - thanks.

PR #4069, opened 2020-03-15
Fixes #2127 Ensures config directory is owned by the user and hence w…

### LINKED ISSUE #2127: install.sh creates .config/rclone directory owned by root
Running `sudo rclone version` creates a `.config/rclone` directory owned by root.

This should be fixed by
  * making version not need the config file (see also #2096)
  * fixing install.sh to not create the config file in the user's home directory
    * `rclone version --config /dev/null` seems like a reasonable work-around, or maybe touch an empty file first.

pendashteh:

Ensures the config directory is owned by the user and hence writable. Currently `rclone config` fails at writing configs at the end of process.

Yes. #2127 

#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -188,7 +188,9 @@ esac
 #update version variable post install
 version=`rclone --version 2>>errors | head -n 1`
 #create the directory for the default configuration folder
-mkdir $HOME/.config/rclone -p
+# Make sure we don't create a root owned .config/rclone directory #2127
+mkdir -p $HOME/.config/rclone
+chown -R $USER:$USER $HOME/.config/rclone
 
 printf "\n${version} has successfully installed."
 printf '\nNow run "rclone config" for setup. Check https://rclone.org/docs/ for more details.\n\n'


ncw:
I'm not sure this will work will it?

```bash
$ cat z.sh 
#!/bin/bash
echo USER=$USER
echo HOME=$HOME
```

```
$ sudo ./z.sh 
USER=root
HOME=/root
```

pendashteh:
This will fix the issue with $USER. I did not find any issue with $HOME.
```
$ cat ./test.sh
user=$USER
[ "$SUDO_USER" ] && user=$SUDO_USER
echo $user @ $HOME
```
```
$ ./test.sh
ubuntu @ /home/ubuntu
$ sudo ./test.sh
ubuntu @ /home/ubuntu
```

ncw:
$HOME doesn't work on my system (Ubuntu 19.10)

Here is something which works for me

```bash
user=${SUDO_USER:-${USER}}      
echo USER $user
home=$(eval echo ~${user})
echo HOME $home
```

Maybe we should just remove the mkdir? After all when the user runs rclone they will get made... The trouble with tinkering with install.sh is that it runs on a lot of different kinds of machine linux/macos/freebsd so it is easy to break without realizing.

pendashteh:
Script looks great, however if the folder is gonna gets created properly later on, let's just remove the line as you suggest.

ncw:
Do you want to update the PR to do that?

pendashteh:
This is not ready to be merged yet.

@ncw the actual problem was when after going through `rclone config` at the end I got the error message that said `~/.config/rclone` is not writable.

So there are two potential solutions:

1. To make sure that the config directory is created with the right permission at the install phase (current PR)
2. To leave it to the `rclone config` command to create the directory if does not exist. (Can you point me at part of the code where this happens?)

Which one do you think is a better/easier approach for now?

ncw:
This bug was introduced by d694bb30e5f31521bbb150c85aa4e89fabf5bd30

1. is a proper fix
2. which is reverting d694bb30e5f31521bbb150c85aa4e89fabf5bd30 is also possible and we know will work but takes away a feature someone cared enough to send a PR about.

If we can make this PR work 100% reliably then I'll merge it

Want to continue trying to make it work properly?

pendashteh:
Ok, cool. Let's do #1 then. And it's kind of urgent since it's a blocker for new user. After following instructions and copy/pasting keys etc (let's say for setting up Google Drive), you find yourself with an error message. I am guessing most new users give up at that point. I'll do another PR based on https://github.com/rclone/rclone/pull/4069#issuecomment-599267287

ncw:
Sounds good - thank you :-)

pendashteh:
@ncw made the PR https://github.com/rclone/rclone/pull/4096
closing this.

PR #4073, opened 2020-03-16
backend/b2: remove unused largeUpload.clearUploadURL()

alrs:
This prunes an unused function from `backend/b2`.


ncw:
Nice one - thank you :-)

PR #4076, opened 2020-03-18
Fix typo

ferferga:
Fix a tiny typo in the Contact page.


ferferga:
I also think that the wording here it's not appropiate. I will make another commit with my proposal, so you can decide if you want to merge both or only one :)

ncw:
That looks great, thank you. I'll merge both those changes :-)

Issue #3815, opened 2019-12-19
Document for dropbox not right


mbrother2:
## What is the problem you are having with rclone?
`rclone config` for Dropbox not right with document https://rclone.org/dropbox/

#### What is your rclone version (output from `rclone version`)
```
butdr@ubuntu1804:~$ rclone version
rclone v1.50.2
- os/arch: linux/amd64
- go version: go1.13.4
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Ubuntu 18.04 64 bit

####  Which cloud storage system are you using? (eg Google Drive)
Dropbox

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
`rclone config`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
butdr@ubuntu1804:~$ rclone -vv config
2019/12/19 15:51:49 DEBUG : rclone: Version "v1.50.2" starting with parameters ["rclone" "-vv" "config"]
2019/12/19 15:51:49 DEBUG : Using config file from "/home/butdr/.config/rclone/rclone.conf"
Current remotes:

Name                 Type
====                 ====
googledrive          drive

e) Edit existing remote
n) New remote
d) Delete remote
r) Rename remote
c) Copy remote
s) Set configuration password
q) Quit config
e/n/d/r/c/s/q> n
name> dropbox
Type of storage to configure.
Enter a string value. Press Enter for the default ("").
Choose a number from below, or type in your own value
 1 / 1Fichier
   \ "fichier"
 2 / Alias for an existing remote
   \ "alias"
 3 / Amazon Drive
   \ "amazon cloud drive"
 4 / Amazon S3 Compliant Storage Provider (AWS, Alibaba, Ceph, Digital Ocean, Dreamhost, IBM COS, Minio, etc)
   \ "s3"
 5 / Backblaze B2
   \ "b2"
 6 / Box
   \ "box"
 7 / Cache a remote
   \ "cache"
 8 / Citrix Sharefile
   \ "sharefile"
 9 / Dropbox
   \ "dropbox"
10 / Encrypt/Decrypt a remote
   \ "crypt"
11 / FTP Connection
   \ "ftp"
12 / Google Cloud Storage (this is not Google Drive)
   \ "google cloud storage"
13 / Google Drive
   \ "drive"
14 / Google Photos
   \ "google photos"
15 / Hubic
   \ "hubic"
16 / JottaCloud
[...]

PR #4077, opened 2020-03-18
Dropbox: Add info about required redirect url

glensc:
Updates documentation when creating your own Dropbox App:
- https://github.com/rclone/rclone/issues/3815#issuecomment-599202942


ncw:
That is great - thank you for doing that :-)

Issue #3886, opened 2020-01-18
Specify Interactive Mode in rclone.conf


jeff-geoff:
#### What is your current rclone version (output from `rclone version`)?

rclone v1.50.2
- os/arch: darwin/amd64
- go version: go1.13.4

#### What problem are you are trying to solve?
Accidental Deletion

#### How do you think rclone should be changed to solve that?

Add the ability to specify something like interactive-mode = true within rclone.conf

Enabling interactive mode would cause rclone to present Yes/No options to a user when conducting an operation that results in overwriting or deleting a file to prevent accidental data loss.

ncw:
The mode for this would be something like this from the `cp` man page

       -i, --interactive
              prompt before delete or overwrite

Some careful locking will be required given that rclone is concurrent but there is only one stdin/stdout for talking to the user.

Finding all the uses of `DryRun` and turning them into a function something like this would be necessary

```go
func SkipDestructive(text, remote string) (skip bool) {
    var flag string
    switch {
       case fs.Config.DryRun:
            flag := "--dry-run"
            skip = true
       case fs.Config.Interactive:
            flag = "--interactive"
            // ask user here
            interactiveMu.Lock()
            defer interactiveMu.Unlock()
            // use config.Choose maybe
        default:
            skip = false
    }
    if skip {
        fs.Logf(src, "Not %s as %s", o, action, flag)
    }
    
}
```

jeff-geoff:
Would it be easier to prompt once at the beginning and say eg “rclone purge will delete 850 files and 10 directories. Are you sure? Y/n”

I realize that doesn’t mimick cp

HowJMay:
May I try to take this issue?

ncw:
That would be great. Let me know if you need help.

fishbullet:
@ncw I'm trying this one, ok ?

ncw:
> I'm trying this one, ok ?

Super - thanks :-) . Let me know if you need help.

ncw:
I've merged this now - thank you @philidor-green for your work.

It will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.53

PR #4083, opened 2020-03-20
fs: interactive mode for desctructive commands #3886

philidor-green:
Abillity to denote interactive mode (ask user to continue potentially danger command)

#3886

#### Checklist


fishbullet:
@ncw Hey !
Could I request a help from you? 

I need understand if I'm on the right way. 

I've found hard to change all entires of `DryRun`, I added only for few command, could you please take a look and tell something? 

Thank you!

fishbullet:
/cc @ncw 
Please take a look.

fs/config/config.go
```diff
@@ -822,6 +822,30 @@ func Choose(what string, defaults, help []string, newOk bool) string {
 	}
 }
 
+// ChooseAnswer asks the user to enter yes or no
```
ncw: Apart from the `what` string this does the same as the `Confirm` function doesn't it?

https://github.com/rclone/rclone/blob/243a868a5b826ee385d4b22d0a8a27a6b034b462/fs/config/config.go#L726-L729

fs/operations/operations.go
```diff
@@ -2165,3 +2166,28 @@ func GetFsInfo(f fs.Fs) *FsInfo {
 	}
 	return info
 }
+
+var interactiveMutex sync.Mutex
+
+func skipDesctructive(ctx context.Context, action string) (string, bool) {
```
ncw: Destructive is spelled wrong here :-)

I think this function needs to take an action (eg `delete`, `move`, `copy`) and a subject (eg  `fs.LogDirName(f, dir)`)

I can see in a future commit (or maybe this one) we'll want to be able to give the user a menu something like this

```
About to ACTION SUBJECT
y) Yes ACTION 
n) No, do not ACTION (default)
s) Skip all future ACTION with no more questions
a) Allow all future ACTION with no more questions
```

So I think the function should look like

```
func skipDestructive(action, subject string) bool {
}
```

fs/operations/operations.go
```diff
@@ -2165,3 +2166,28 @@ func GetFsInfo(f fs.Fs) *FsInfo {
...
+		flag = "--interactive"
+		interactiveMutex.Lock()
+		defer interactiveMutex.Unlock()
+		skip = !config.ChooseAnswer(
+			fmt.Sprintf("rclone will %s. Are you sure? y/n\n", action),
+		)
+	default:
+		skip = false
+	}
+	message := fmt.Sprintf("Not %s as %s is set", action, flag)
```
ncw: This function should probably just do the logging rather than returning the message and just return a bool

fs/operations/operations.go
fishbullet: Could you please point me, how to code `skip all future` feature? I think only way in write to config at runtime. 

I added `ctx` here in the arguments because want to count a objects\dirs to be processed to show in the log message.

ncw: It would just be `skip all future` in this run of rclone.  So I'd make a map which is protected by your mutex

```
var (
    interactiveMutex sync.Mutex
    interactiveSkip = map[string]bool{}
    interactiveAllow = map[string]bool{}
```

Then see if `interactiveSkip[action]` or `interactiveAllow[action]` was set to immediately return `false` or `true`. When the user sets the option you set `interactiveSkip[action] = true`.

> I added ctx here in the arguments because want to count a objects\dirs to be processed to show in the log message.

I don't understand what you mean here? Anyway adding `ctx` is probably a good idea!

ncw:
Sorry this PR fell off my radar...

Do you want help finishing it off?

ncw:
I'm happy to finish it off if you haven't got time - I'd like to get it into 1.53 :-)

philidor-green:
Sorry for delay, dont have much time right now.

ncw:
@philidor-green no worries. I will finish it off. You've done nearly all the work - thank you :-)

PR #4084, opened 2020-03-22
Use custom http client so that --no-check-certificate is honored by oauth2 token fetch

mspieth:
When using onedrive and configuring a remote with --no-check-certificate, the token is not fetchable with error `x509: certificate signed by unknown authority`

This is because the token fetch in oauthutil does not use the provided http client.

This PR allows the oauth2 token fetch phase to honor the --no-check-certificate flag.

No

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -503,7 +503,8 @@ func doConfig(id, name string, m configmap.Mapper, oauthConfig *oauth2.Config, o
 	}
 
 	// Exchange the code for a token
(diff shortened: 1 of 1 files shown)


mspieth:
I don't understand the above failure. The patch does not change the functionality very much from the rc point of view AFAICT.

ncw:
Nice one - that looks really good thank you.

The test failure is one of those tests which fails occasionally alas. I go through killing them where I can but it is a lot of work :-(

I'll merge this now - thank you.

PR #4092, opened 2020-03-25
docs: unmystify how crypt stores encryption password in config

samcv:
Without explaining exactly how this is generated, it can be confusing
and worrying to not know how the password that encrypts your data is
stored.

This also brings peace of mind to the user that even though
the same password is obscured differently each time, all the data to
get back to the original password remains. Explaining how it works
is much better than the reader of the documentation having to trust
a blackboxy/magical mechanism.

No

#### Checklist


ncw:
Thanks for that - nice explanation :-)

I'll merge it now

PR #4096, opened 2020-03-28
Fixes #2127 Makes config folder writable for the user.

### LINKED ISSUE #2127: install.sh creates .config/rclone directory owned by root

pendashteh:
#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -188,7 +188,13 @@ esac
 #update version variable post install
 version=`rclone --version 2>>errors | head -n 1`
 #create the directory for the default configuration folder
-mkdir $HOME/.config/rclone -p
+# Make sure we don't create a root owned .config/rclone directory #2127
+mkdir -p $HOME/.config/rclone
+# Prepare values for $user and $home for the current user to get consitency when the command is run as 'sudo'
+user=${SUDO_USER:-${USER}}
+home=$(eval echo ~${user})
+# Make the user the owner of the config folder to ensure the 'write' permission
+chown -R $user:$user $home/.config/rclone
 
 printf "\n${version} has successfully installed."
 printf '\nNow run "rclone config" for setup. Check https://rclone.org/docs/ for more details.\n\n'


docs/content/install.sh
```diff
@@ -188,7 +188,13 @@ esac
...
 version=`rclone --version 2>>errors | head -n 1`
 #create the directory for the default configuration folder
-mkdir $HOME/.config/rclone -p
+# Make sure we don't create a root owned .config/rclone directory #2127
+mkdir -p $HOME/.config/rclone
+# Prepare values for $user and $home for the current user to get consitency when the command is run as 'sudo'
+user=${SUDO_USER:-${USER}}
+home=$(eval echo ~${user})
+# Make the user the owner of the config folder to ensure the 'write' permission
+chown -R $user:$user $home/.config/rclone
```
ncw: I'd probably just go for `chown $user $home/.config/rclone` here as I don't think the BSD's use a group with the same name as the user necessarily and `-R` is touching other files the user may have in there.

Also we probably shouldn't be doing this at all unless we actually created the directory so maybe put the mkdir to the chown in an if statement which only runs if the directory doesn't not exist?

ncw: Any thoughts on this?

Yexiaoxing: Maybe we should also throw a hint that the folder already exists. Something like it should work.

```bash
user=${SUDO_USER:-${USER}}
home=$(eval echo ~${user})

if [ ! -d "$home/.config/rclone" ]
then 
  mkdir -p $home/.config/rclone
  if [ $? -ne 0 ] 
  then 
    echo "Config folder $home/.config/rclone cannot be created. Please check if you have enough permission."
  else
    chown $user $home/.config/rclone
  fi
else
  echo "Config folder $home/.config/rclone already exists. If you have not installed rclone before, you may check its owner to avoid permission errors in future."
fi
```

ncw: While we are thinking about this I reverted d694bb30e5f31521bbb150c85aa4e89fabf5bd30 in 2c4aadb588b9db48a74f923e4f40ebe3e303b360 and pushed the install.sh up so that stuff isn't immediately broken for new users.

Yexiaoxing: Well the revert commit also raises a problem - what if .config is even not exist before? The .config folder will also be owned by root.

ivandeex:
@pendashteh Please address review comments

ivandeex:
This PR has review comments unanswered for a year. Marking as draft until it shows signs of life.

ivandeex:
This PR is abandoned and needs a new maintainer.
It will be closed in about three months.

ivandeex:
This PR is abandoned and can be closed in 2 months.

Issue #3762, opened 2019-11-24
`--files-from` does not list the files with leading/trailing whitespace in its name


geart891:
#### What is the problem you are having with rclone?
`--files-from` does not list the files with leading/trailing whitespace in its name. Since `--files-from` does not support **wildcards**, there is noway to list the files with the tag.

#### What is your rclone version (output from `rclone version`)
Version tested:
1.49.5
1.50.1
1.50.2

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Os tested:
Ubuntu x64
Windows 10 x64

####  Which cloud storage system are you using? (eg Google Drive)
Google Drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
`rclone lsf remote: --files-only --files-from list.txt`

`list.txt`
```
# Filenames with leading, trailing whitespace or both
 leading-whitespace-filename.txt
trailing-whitespace-filename.txt 
 leading-trailing-whitespace.txt 
```

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
$ rclone lsf remote: --files-only --files-from list.txt -vv                                                        2019/11/25 05:40:40 DEBUG : rclone: Version "v1.50.2" starting with parameters ["F:\\System\\rClone\\rclone.exe" "lsf" "remote:" "--files-only" "--files-from" "list.txt" "-vv"]
2019/11/25 05:40:40 DEBUG : Using config file from "C:\\Users\\Rhaps\\.config\\rclone\\rclone.conf"
2019/11/25 05:40:42 DEBUG :  leading-trailing-whitespace.txt : Excluded
2019/11/25 05:40:42 DEBUG :  leading-whitespace-filename.txt: Excluded
2019/11/25 05:40:42 DEBUG : trailing-whitespace-filename.txt : Excluded
2019/11/25 05:40:42 DEBUG : 5 go routines active
2019/11/25 05:40:42 DEBUG : rclone: Version "v1.50.2" finishing with parameters ["F:\\System\\rClone\\rclone.exe" "lsf" "remote:" "--files-only" "--files-from" "list.txt" "-vv"]
```

P/S: A small visual bug. 
It seems that the `Elapsed time:` line does not include an eol character.
Os: Windows 10 x64, git-bash MINGW64 Shell, lastest rclone.

[...]

PR #4102, opened 2020-04-01
filterflags: Added --process-raw flag

ankur0493:
--files-from parses input files by ignoring comments starting with # and ;
and stripping whitespace from start and end of strings. We want to add some solution to this so that expected behavior of `rclone lsf` being good input to `--files-from` is restored.

A --process-raw flag was added so that --files-from reads every line from the file ignoring
comment characters and not stripping whitespace while maintaining backwards
compatibility.

https://forum.rclone.org/t/rclone-skips-files-ending-in-unicode-whitespace-when-using-files-from/15285

#### Checklist

### DIFF
--- a/cmd/lsf/lsf.go
+++ b/cmd/lsf/lsf.go
@@ -132,13 +132,13 @@ Eg
     "this file contains a comma, in the file name.txt",6
 
 Note that the --absolute parameter is useful for making lists of files
-to pass to an rclone copy with the --files-from flag.
+to pass to an rclone copy with the --files-from-raw flag.
 
 For example to find all the files modified within one day and copy
 those only (without traversing the whole directory structure):
 
     rclone lsf --absolute --files-only --max-age 1d /path/to/local > new_files
-    rclone copy --files-from new_files /path/to/local remote:path
+    rclone copy --files-from-raw new_files /path/to/local remote:path
 
 ` + lshelp.Help,
 	Run: func(command *cobra.Command, args []string) {
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -1542,6 +1542,7 @@ For the filtering options
   * `--include`
   * `--include-from`
   * `--files-from`
+  * `--files-from-raw`
   * `--min-size`
   * `--max-size`
   * `--min-age`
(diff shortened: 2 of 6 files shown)


ncw:
I think the code looks great, thank you :-)

This needs some docs in `docs/content/filtering.md` and a callout in `docs/content/docs.md`  in the [filtering section](https://rclone.org/docs/#filtering).

I really don't like the flag name though! `--process-raw` is a bit too general for a global flag. It should really start with `--files-from-` to make it discoverable. So it could be `--files-from-raw` or something else - suggestions?

Can you put `Fixes #3762` in the commit message too?

I think you'll need to rebase this on master and force push as GitHub says this is unmergeable at the moment!

Thank you :-)

ankur0493:
> This needs some docs in `docs/content/filtering.md` and a callout in `docs/content/docs.md` in the [filtering section](https://rclone.org/docs/#filtering).

Sure, I will add that.

> I really don't like the flag name though! `--process-raw` is a bit too general for a global flag. It should really start with `--files-from-` to make it discoverable. So it could be `--files-from-raw` or something else - suggestions?

Well the trouble with having a flag starting with `--files-from` is that it cannot act as an option for `--files-from`, but will have to be a new flag in itself. We can go ahead with `--files-from-raw` as you originally proposed, but we will than have to post-process that in `NewFilter` and map it into `Opt.FilesFrom` and `Opt.ProcessRaw` (or similar) flags to minimize the code changes.

> Can you put `Fixes #3762` in the commit message too?
Sure, will edit with my next change. Which commit message will the merge commit picks up when the PR has multiple commits?

> I think you'll need to rebase this on master and force push as GitHub says this is unmergeable at the moment!

I had rebased it. You probably reviewed an older version. Right now it no longer complains of conflicts. :)

ankur0493:
@ncw I have made the changes suggested. Please suggest if any further change is required.

ncw:
I think that looks great thank you :-)  I'll merge tomorrow when Github isn't misbehaving!

ncw:
Thank you very much for an excellent contribution - all merged now :-)

Issue #4091, opened 2020-03-25
Box: "Enterprise" box_sub type does not refresh token


davidhewitt:
#### What is your current rclone version (output from `rclone version`)?

rclone v1.51.0

#### What problem are you are trying to solve?

I have a Box app, with a `config.json` downloaded for that app.

I can authenticate with `rclone config`, setting the `box_sub_type` to `enterprise` and providing the path to the config.json. `rclone` successfully authenticates and stores a config entry similar to the below:

```
[box]
type = box
box_config_file = config.json
box_sub_type = enterprise
token = {"access_token": "xyz...", "token_type": "bearer", "expiry": "<timestamp>"}
```

`rclone` will then operate correctly until after token `expiry`, from which point `rclone` fails with errors like `token expired and refresh token is not set`

#### How do you think rclone should be changed to solve that?

All information to re-authenticate is provided in `config.json`, so `rclone` should be able to generate a new token and proceed operation.

ncw:
I don't know very much about this...

@dbramwell you contributed the original code for this - is it working OK for you?

dbramwell:
When using the box config.json we don't get a refresh token back, so the usual way of refreshing doesn't work. We've been getting round it by automating the rclone config step as well and running that before each rclone box command.

Having rclone automatically ask for a new token using the config.json seems reasonable to me. If someone then removed the config file then rclone would be unable to get a new token but I guess that's fine. Another option would be for rclone to store the contents of the config.json, rather than its location? Not sure if that could be considered a security problem

ncw:
The box backend knows when the token expires - this bit of code gets called. We store the location of the `config.json` in the configuration file so we should have everything we need to auto renew.

https://github.com/rclone/rclone/blob/1af9fcbbfae66fc0de0aa44eb2dd3d8c3b185321/backend/box/box.go#L396-L400

Do you fancy having a go at this @dbramwell ?

dbramwell:
I was having a look the other day but found it quite hard to follow. I'll give it a go but will probably be back with questions.

ncw:
Sounds great @dbramwell ! Ping me here if you need any help - thanks.

abrar-hnxlabs:
@dbramwell  any ETA on this fix, I am also getting this error, I could also look into the code and fix this if you havent started yet.

dbramwell:
[...]

PR #4107, opened 2020-04-04
Box: Refresh token when using jwt auth - Fixes #4901 

dbramwell:

Refresh the auth token when using box, jwt and box config.json file

[#4091 ](https://github.com/rclone/rclone/issues/4091)

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -89,22 +89,7 @@ func init() {
 			boxSubType, boxSubTypeOk := m.Get("box_sub_type")
 			var err error
 			if ok && boxSubTypeOk && jsonFile != "" && boxSubType != "" {
-				boxConfig, err := getBoxConfig(jsonFile)
-				if err != nil {
-					log.Fatalf("Failed to configure token: %v", err)
-				}
-				privateKey, err := getDecryptedPrivateKey(boxConfig)
-				if err != nil {
-					log.Fatalf("Failed to configure token: %v", err)
-				}
-				claims, err := getClaims(boxConfig, boxSubType)
-				if err != nil {
-					log.Fatalf("Failed to configure token: %v", err)
-				}
-				signingHeaders := getSigningHeaders(boxConfig)
-				queryParams := getQueryParams(boxConfig)
-				client := fshttp.NewClient(fs.Config)
-				err = jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m, client)
+				err = refreshJWTToken(jsonFile, boxSubType, name, m)
 				if err != nil {
 					log.Fatalf("Failed to configure token with jwt authentication: %v", err)
 				}
@@ -161,6 +146,26 @@ func init() {
 				encoder.EncodeInvalidUtf8),
 		}},
 	})
+}
+
+func refreshJWTToken(jsonFile string, boxSubType string, name string, m configmap.Mapper) error {
+	boxConfig, err := getBoxConfig(jsonFile)
+	if err != nil {
+		log.Fatalf("Failed to configure token: %v", err)
+	}
+	privateKey, err := getDecryptedPrivateKey(boxConfig)
+	if err != nil {
+		log.Fatalf("Failed to configure token: %v", err)
+	}
+	claims, err := getClaims(boxConfig, boxSubType)
+	if err != nil {
+		log.Fatalf("Failed to configure token: %v", err)
+	}
+	signingHeaders := getSigningHeaders(boxConfig)
+	queryParams := getQueryParams(boxConfig)
+	client := fshttp.NewClient(fs.Config)
+	err = jwtutil.Config("box", name, claims, signingHeaders, queryParams, privateKey, m, client)
+	return err
 }
 
 func getBoxConfig(configFile string) (boxConfig *api.ConfigJSON, err error) {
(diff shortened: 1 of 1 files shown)


dbramwell:
@ncw I wanted to just use the Renewer from oauthutil, passing a different function to do the renew. But then I realised that the renewer in there only runs when [uploads are pending](https://github.com/rclone/rclone/blob/master/lib/oauthutil/renew.go#L41), so when running a simple `ls`, rclone still failed to refresh the token. So I created a new renewer that does it whether there are uploads pending or not. I'm not sure that this is the correct thing to do but it works.

I'm unsure where the token currently get's refreshed in `ls` cases (seemed to be somewhere in the `pacer` calls [here](https://github.com/rclone/rclone/blob/master/backend/box/box.go#L547). Maybe it's better for that to be altered for jwt and the `jwtutil.Renew` to be removed? I'm unsure about how it all works though.

ncw:
Apologies for the delay in looking at this.

> I'm unsure where the token currently get's refreshed in `ls` cases (seemed to be somewhere in the `pacer` calls [here](https://github.com/rclone/rclone/blob/master/backend/box/box.go#L547). Maybe it's better for that to be altered for jwt and the `jwtutil.Renew` to be removed? I'm unsure about how it all works though.

The token gets renewed by the oauth library golang.org/x/oauth2 - you use its http Transport and it magically refreshes the token when it expires.

The oauth library may be able to help you directly - it understands JWTs (unlike me :-)

> So I created a new renewer that does it whether there are uploads pending or not. I'm not sure that this is the correct thing to do but it works.

That will only work for uploads though as that is the only place we call the token renewer at the moment.

You could hook it into the Signer call for the rest library which is called before every HTTP transaction

https://github.com/rclone/rclone/blob/6893ce0bbf81f9d4264f12af3e62f21ae7ba5ced/backend/sugarsync/sugarsync.go#L422 (uploaded in 15-30 mins)

You have to be a little careful about recursion - I should probably add a "NoSigner" option to `rest.Opts`.

The most efficient hook might be to add a timer on refresh of the token to when it next needs refreshing since we know exactly when it expires. We should examine the expiry on startup and refresh if necessary then or set the timer, then set it again on every refresh.

dbramwell:
No problem, thanks for taking a look.

> That will only work for uploads though as that is the only place we call the token renewer at the moment.

I removed the check for whether uploads are taking place in the jwt renewer [here](https://github.com/rclone/rclone/pull/4107/files#diff-e57f15e014bec6db19b4be39ba7eff78R41). It does work, but I don't like the idea of kicking off an async process to renew the token when running one short command like `ls` or whatever.

I'm still looking at this when I get time, currently hoping to use `golang.org/x/oauth2` like you said. I think I tried this last time but hit an issue with it [not supporting keys with a passphrase](https://github.com/rclone/rclone/blob/master/vendor/golang.org/x/oauth2/jwt/jwt.go#L42). Will see where I get to and hopefully have changes by the weekend.

ncw:
> > That will only work for uploads though as that is the only place we call the token renewer at the moment.
> 
> I removed the check for whether uploads are taking place in the jwt renewer [here](https://github.com/rclone/rclone/pull/4107/files#diff-e57f15e014bec6db19b4be39ba7eff78R41), compared to the oauthutil one [here](https://github.com/rclone/rclone/blob/master/lib/oauthutil/renew.go#L41). It does work, but I don't like the idea of kicking off an async process to renew the token when running one short command like `ls` or whatever.

You are right. I missed the fact that we kick off a go routine to received the expired notification so ignore what I said above, sorry!

Maybe the neatest thing to do would be to add a mode to the existing token renewer so that it always renews the token when it expires. That could be as simple as calling `Start()` as part of the initialisation (so uploads is always > 0) with no code changes needed which should simplify the code and mean that we don't need to duplicate the token renewer code.

> I'm still looking at this when I get time, currently hoping to use `golang.org/x/oauth2` like you said. I think I tried this last time but hit an issue with it [not supporting keys with a passphrase](https://github.com/rclone/rclone/blob/master/vendor/golang.org/x/oauth2/jwt/jwt.go#L42). Will see where I get to and hopefully have changes by the weekend.

Thanks

dbramwell:
> Maybe the neatest thing to do would be to add a mode to the existing token renewer so that it always renews the token when it expires. That could be as simple as calling Start() as part of the initialisation (so uploads is always > 0) with no code changes needed which should simplify the code and mean that we don't need to duplicate the token renewer code.

Cool, I've taken that approach.

I took a look into using the jwt stuff in the oauth2 package and eventually hit the issue of not being able to set query params which will be solved by [this](https://github.com/golang/oauth2/pull/390/files), which once I saw I realised I'd definitely seen when I was trying to add the jwt stuff in the first place, I should have taken note! Once that's merged I think a lot my code from the Config function in jwtutil can be removed, and the renewer stuff could possibly be simplified too.

ncw:
> Cool, I've taken that approach.

Rather than changing the interface of the token renewer, I'd I've just called `Start()` from the box backend. I'm sorry about my stream of consciousness ramble above - I should have been more explicit!

> I took a look into using the jwt stuff in the oauth2 package and eventually hit the issue of not being able to set query params which will be solved by [this](https://github.com/golang/oauth2/pull/390/files), which once I saw I realised I'd definitely seen when I was trying to add the jwt stuff in the first place, I should have taken note! Once that's merged I think a lot my code from the Config function in jwtutil can be removed, and the renewer stuff could possibly be simplified too.

That is good news! I put a note on the PR asking for a status check.

dbramwell:
> Rather than changing the interface of the token renewer, I'd I've just called Start() from the box backend. I'm sorry about my stream of consciousness ramble above - I should have been more explicit!

No problem. Done!

I'm not sure why the two actions are failing, running on my local I don't seem to get errors.

ncw:
> I'm not sure why the two actions are failing, running on my local I don't seem to get errors.

Unfortunately they are tests which fail occasionally that I haven't had time to fix :-( 

Your changes look great now - thank you :-)

I'll merge them now.

PR #4116, opened 2020-04-10
s3: use single memory pool

mmatczuk:

Previously we had a map of pools for different chunk sizes.
In practice the mapping is not very useful and requires a lock.
Pools of size other that ChunkSize can only happen when we have a huge file (over 10k * ChunkSize).
We need to have a bunch of identically sized huge files.
In such case most likely ChunkSize should be increased.

The mapping and its lock is replaced with a single initialised pool for ChunkSize, in other cases pool is allocated and freed on per file basis.

No

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -941,19 +941,18 @@ type Options struct {
 
 // Fs represents a remote s3 server
 type Fs struct {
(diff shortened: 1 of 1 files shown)


ncw:
That looks sensible to me - I'll merge that - thanks

PR #4119, opened 2020-04-11
Implement option to view the Jottacloud trash

buengese:
As discussed in the forum I added a --jottacloud-trash-only flag that shows Trashed files in their original location. Normal operations should actually work on trashed files as long as the flag is included in the command.
I've also quickly added the cleanup command.

### DIFF
--- a/backend/jottacloud/api/types.go
+++ b/backend/jottacloud/api/types.go
@@ -164,6 +164,12 @@ type CustomerInfo struct {
 	IOSHash           string      `json:"ios_hash"`
 }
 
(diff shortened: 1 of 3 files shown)


ncw:
Nice one! I gave that the once over and LTGM.

Please merge :-)

albertony:
I guess documentation should be changed to CleanUp=Yes in https://rclone.org/overview/#optional-features then?

ncw:
> I guess documentation should be changed to CleanUp=Yes in https://rclone.org/overview/#optional-features then?

It should indeed...

PR #4123, opened 2020-04-11
copy: fix typo

dandv:
Just a drive-by typo fix. I presume the [md](https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone_copy.md) is autogenerated from this file?


ncw:
Yes you are correct :-)

Thanks for the drive by fix!

PR #4141, opened 2020-04-17
rc: adding mount command

negative0:
Adding mount support for remote control.
I need suggestions about the directory structure, and the change is being implemented correctly.
This is just an initial implementation. 

I see that every independent module like `fs` has a separate `rc.go` file. But the files in `cmd` don't have it. Instead inside their init function, they have an `AddCommand()` call. How should the `mount` endpoint addition for `rc` look like?

No

#### Checklist

### DIFF
--- a/cmd/mount2/mount.go
+++ b/cmd/mount2/mount.go
@@ -24,6 +24,10 @@ import (
 
 func init() {
 	mountlib.NewMountCommand("mount2", true, Mount)
+
+	// Add mount to rc
+	mountlib.AddRc("mount2", mount)
+
 }
 
 // mountOptions configures the options from the command line flags
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -39,6 +39,13 @@ var (
 	AsyncRead          = true        // do async reads by default
 )
 
+type (
+	// UnmountFn is called to unmount the file system
+	UnmountFn func() error
+	// MountFn is called to mount the file system
+	MountFn func(f fs.Fs, mountpoint string) (*vfs.VFS, <-chan error, func() error, error)
+)
+
 // Global constants
 const (
 	MaxLeafSize = 4095 // don't pass file names longer than this
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -39,6 +39,9 @@ func init() {
 		name = "mount"
 	}
 	mountlib.NewMountCommand(name, false, Mount)
+	// Add mount to rc
+	mountlib.AddRc("cmount", mount)
+
 }
 
 // mountOptions configures the options from the command line flags
--- a/cmd/mount/mount.go
+++ b/cmd/mount/mount.go
@@ -23,6 +23,8 @@ import (
 
 func init() {
 	mountlib.NewMountCommand("mount", false, Mount)
+	// Add mount to rc
+	mountlib.AddRc("mount", mount)
 }
 
 // mountOptions configures the options from the command line flags
--- a/cmd/mountlib/rc.go
+++ b/cmd/mountlib/rc.go
@@ -3,6 +3,7 @@ package mountlib
 import (
 	"context"
 	"log"
+	"sort"
 
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
@@ -16,6 +17,18 @@ var (
 	unmountFns map[string]UnmountFn
 )
 
+// AddRc adds mount and unmount functionality to rc
+func AddRc(mountUtilName string, mountFunction MountFn) {
+	if mountFns == nil {
+		mountFns = make(map[string]MountFn)
+	}
+	if unmountFns == nil {
+		unmountFns = make(map[string]UnmountFn)
+	}
+	// rcMount allows the mount command to be run from rc
+	mountFns[mountUtilName] = mountFunction
+}
+
 func init() {
 	rc.Add(rc.Call{
 		Path:         "mount/mount",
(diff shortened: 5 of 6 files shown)


cmd/mount2/mount.go
```diff
@@ -5,7 +5,10 @@
 package mount2
 
 import (
+	"context"
 	"fmt"
+	"github.com/rclone/rclone/cmd"
```
golangcibot, line 10: File is not `goimports`-ed (from `goimports`)

ncw:
That looks like a good start :-)

There are 3 possible mount commands confusingly and I try to move shared logic into `cmt/mountlib` so I think this logic should live there.

You'd want to make an `AddRc(MountFunction)` function in mountlib then just call it in the init methods for `cmd/mount` `cmd/mount2` `cmd/cmount` with the corresponding `Mount` function.

There is more we could do here like add access to parameters (vfs and mount)

Note that the `Mount()` function doesn't return so we might want to pass the `mount` function instead which is what the tests use.

negative0:
@ncw, thank you for taking the suggestions. I will update try and complete the changes.

negative0:
@ncw, This is ready for the review :)

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+		Title:        "Create a new mount point",
+		Help: `
+rclone allows Linux, FreeBSD, macOS and Windows to
+mount any of Rclone's cloud storage systems as a file system with
+FUSE.
+
+This takes the following parameters
+
+	- mountPoint: valid path on the local machine (required)
+	- remotePath - a remote path to be mounted (required)
```
ncw, line 87: This should be called `fs` to be consistent with the rest of the rc

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+	remotePath, err := in.GetString("remotePath")
+	if err != nil {
+		return nil, err
+	}
+
+	out = make(rc.Params)
+
+	tempArgs := []string{remotePath}
+	//append(tempArgs, remotePath)
+	fdst := cmd.NewFsDir(tempArgs)
```
ncw, line 153: If you name it `Fs` then you can use `rc.GetFs(in)` to parse it and turn it into an fs.Fs.

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+		Fn:           unMountRc,
+		Title:        "Unmount all active mounts",
+		Help: `
+rclone allows Linux, FreeBSD, macOS and Windows to
+mount any of Rclone's cloud storage systems as a file system with
+FUSE.
+
+This takes the following parameters
+
+	- mountPoint: valid path on the local machine (required)
```
ncw, line 107: These should be presented the other way round so

- fs:
- mountPoint:

As the mount command is traditionally presented `mount <device> <directory>`

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+    rclone rc mount/unmount mountPoint=/home/<user>/mountPoint remotePath=mydrive:
+		`,
+	})
+}
+
+// rcMount allows the umount command to be run from rc
+func unMountRc(_ context.Context, _ rc.Params) (out rc.Params, err error) {
+
+	out = make(rc.Params)
+	if currentUnmountFn != nil {
```
ncw, line 121: This is going to go wrong if there is more than one mount...

I know people want rclone to run multiple mounts at once so we need to make this work!

We probably need to store the unmount functions in a map keyed on remote string + mountpoint maybe?

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+		err := currentUnmountFn()
```
ncw, line 122: Just return the err - the handling code will deal with setting the status.

so

```go
err := currentUnmountFn()
if err != nil {
    return nil, err
}
```

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+		if err != nil {
+			out["status"] = "Failed"
+			out["err"] = err
+		} else {
+			out["status"] = "Success"
+		}
+	} else {
+		out["status"] = "No mount found"
```
ncw, line 130: I'd just return an error here `return nil, errors.New("mount not found")`

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+		return nil, err
+	}
+
+	out = make(rc.Params)
+
+	tempArgs := []string{remotePath}
+	//append(tempArgs, remotePath)
+	fdst := cmd.NewFsDir(tempArgs)
+
+	_, _, currentUnmountFn, err = currentMountFn(fdst, mountpoint)
```
ncw, line 155: So we should be stashing the unmountfn in a dictionary keyed off mountpoint probably...

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+	if err != nil {
+		log.Printf("mount FAILED: %v", err)
+		out["status"] = "Failed"
```
ncw, line 158: Just return the error here

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+
+	} else {
+		log.Printf("mount OK")
+		out["status"] = "Success"
```
ncw, line 162: No need for this line - the handler will deal with it.

cmd/mountlib/mount.go
```diff
@@ -51,6 +65,111 @@ func init() {
...
+
+	}
+	if err != nil {
+		log.Fatalf("Fatal error: %v", err)
```
ncw, line 166: Don't put fatal errors in the rc! Just return it :-)

negative0:
Thanks for the corrections. I will implement them and push them today or tomorrow.

ncw:
> Thanks for the corrections. I will implement them and push them today or tomorrow.

Thank you :-)

cmd/mountlib/mount.go
negative0: Can we include only the mountpoint? mountpoint will has to be unique in itself.

cmd/mountlib/mount.go
negative0: Umm, I don't think I understand what you mean

ncw: You don'e really need to return `{"status": "Success"}` as the error status of the `rc` call will indicate that, so in general the rc commands don't bother.

negative0: So should I do ```return nil, nil``` ?

negative0:
@ncw , please check once

ncw:
The CI has gone strange for this!

I pulled it locally to see what the problem was. I fixed a small linting issue and in the process moved the rc code to a separate file.

I've also added a commit with some tests in on the top - lets see if they pass on the CI!

ncw:
All passed now :-)

Thanks for making this feature - it will make users very happy!

To be as useful as it could be the `mount/mount` command really needs to take a vfs options structure and a mount options structure (which needs making). Do you fancy working on this? I could do but I want to concentrate on getting the 1.52 release out the door so am trying not to get distracted!

Being able to run the `rclone serve` commands over the rc would be very useful also.

PR #4143, opened 2020-04-17
New backend for seafile servers

### LINKED ISSUE #1877: Support for Seafile
Are there any plans on supporting Seafile? 

It's a open source cloud system similar to Dropbox and they seem to have an API available here: 
https://manual.seafile.com/develop/

I couldn't fine anything about it on here so I'd though to ask.

creativeprojects:

Implementation of a backend for Seafile 

Fix #1877 

#### Checklist

### DIFF
--- a/backend/seafile/object.go
+++ b/backend/seafile/object.go
@@ -1,0 +1,127 @@
+package seafile
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/hash"
+)
+
+// Object describes a seafile object (also commonly called a file)
+type Object struct {
+	fs            *Fs       // what this object is part of
+	id            string    // internal ID of object
+	remote        string    // The remote path (full path containing library name if target at root)
+	pathInLibrary string    // Path of the object without the library name
+	size          int64     // size of the object
+	modTime       time.Time // modification time of the object
+	libraryID     string    // Needed to download the file
+}
+
+// ==================== Interface fs.DirEntry ====================
+
+// Return a string version
+func (o *Object) String() string {
+	if o == nil {
+		return "<nil>"
+	}
+	return o.remote
+}
+
+// Remote returns the remote string
+func (o *Object) Remote() string {
+	return o.remote
+}
+
+// ModTime returns last modified time
+func (o *Object) ModTime(context.Context) time.Time {
+	return o.modTime
+}
+
+// Size returns the size of an object in bytes
+func (o *Object) Size() int64 {
+	return o.size
+}
+
+// ==================== Interface fs.ObjectInfo ====================
+
+// Fs returns the parent Fs
+func (o *Object) Fs() fs.Info {
+	return o.fs
+}
+
+// Hash returns the selected checksum of the file
+// If no checksum is available it returns ""
+func (o *Object) Hash(ctx context.Context, ty hash.Type) (string, error) {
+	return "", hash.ErrUnsupported
+}
+
+// Storable says whether this object can be stored
+func (o *Object) Storable() bool {
+	return true
+}
+
+// ==================== Interface fs.Object ====================
+
+// SetModTime sets the metadata on the object to set the modification date
+func (o *Object) SetModTime(ctx context.Context, t time.Time) error {
+	return fs.ErrorCantSetModTime
+}
+
+// Open opens the file for read.  Call Close() on the returned io.ReadCloser
+func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (io.ReadCloser, error) {
+	downloadLink, err := o.fs.getDownloadLink(ctx, o.libraryID, o.pathInLibrary)
+	if err != nil {
+		return nil, err
+	}
+	reader, err := o.fs.download(ctx, downloadLink, o.Size(), options...)
+	if err != nil {
+		return nil, err
+	}
+	return reader, nil
+}
+
+// Update in to the object with the modTime given of the given size
+//
+// When called from outside a Fs by rclone, src.Size() will always be >= 0.
+// But for unknown-sized objects (indicated by src.Size() == -1), Upload should either
+// return an error or update the object properly (rather than e.g. calling panic).
+func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) error {
+	// The upload sometimes return a temporary 500 error
+	// We cannot use the pacer to retry uploading the file as the upload link is single use only
+	for retry := 0; retry <= 3; retry++ {
+		uploadLink, err := o.fs.getUploadLink(ctx, o.libraryID)
+		if err != nil {
+			return err
+		}
+
+		uploaded, err := o.fs.upload(ctx, in, uploadLink, o.pathInLibrary)
+		if err == ErrorInternalDuringUpload {
+			// This is a temporary error, try again with a new upload link
+			continue
+		}
+		if err != nil {
+			return err
+		}
+		// Set the properties from the upload back to the object
+		o.size = uploaded.Size
+		o.id = uploaded.ID
+
+		return nil
+	}
+	return ErrorInternalDuringUpload
+}
+
+// Remove this object
+func (o *Object) Remove(ctx context.Context) error {
+	return o.fs.deleteFile(ctx, o.libraryID, o.pathInLibrary)
+}
+
+// ==================== Optional Interface fs.IDer ====================
+
+// ID returns the ID of the Object if known, or "" if not
+func (o *Object) ID() string {
+	return o.id
+}
--- a/backend/seafile/seafile.go
+++ b/backend/seafile/seafile.go
@@ -1,0 +1,1247 @@
+package seafile
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/pkg/errors"
+	"github.com/rclone/rclone/backend/seafile/api"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/config"
+	"github.com/rclone/rclone/fs/config/configmap"
+	"github.com/rclone/rclone/fs/config/configstruct"
+	"github.com/rclone/rclone/fs/config/obscure"
+	"github.com/rclone/rclone/fs/fserrors"
+	"github.com/rclone/rclone/fs/fshttp"
+	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/lib/bucket"
+	"github.com/rclone/rclone/lib/cache"
+	"github.com/rclone/rclone/lib/encoder"
+	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/random"
+	"github.com/rclone/rclone/lib/rest"
+)
+
+const (
+	librariesCacheKey = "all"
+	retryAfterHeader  = "Retry-After"
+)
+
+// This is global to all instances of fs
+// (copying from a seafile remote to another remote would create 2 fs)
+var (
+	rangeDownloadNotice sync.Once  // Display the notice only once
+	createLibraryMutex  sync.Mutex // Mutex to protect library creation
+)
+
+// Register with Fs
+func init() {
+	fs.Register(&fs.RegInfo{
+		Name:        "seafile",
+		Description: "seafile",
+		NewFs:       NewFs,
+		Options: []fs.Option{{
+			Name:     "url",
+			Help:     "URL of seafile host to connect to",
+			Required: true,
+			Examples: []fs.OptionExample{{
+				Value: "https://cloud.seafile.com/",
+				Help:  "Connect to cloud.seafile.com",
+			}},
+		}, {
+			Name:     "user",
+			Help:     "User name",
+			Required: true,
+		}, {
+			Name:       "pass",
+			Help:       "Password",
+			IsPassword: true,
+			Required:   true,
+		}, {
+			Name: "library",
+			Help: "Name of the library. Leave blank to access all non-encrypted libraries.",
+		}, {
+			Name:       "library_key",
+			Help:       "Library password (for encrypted libraries only). Leave blank if you pass it through the command line.",
+			IsPassword: true,
+		}, {
+			Name:     "create_library",
+			Help:     "Should create library if it doesn't exist",
+			Advanced: true,
+			Default:  false,
+		}, {
+			Name:     config.ConfigEncoding,
+			Help:     config.ConfigEncodingHelp,
+			Advanced: true,
+			Default: (encoder.EncodeZero |
+				encoder.EncodeCtl |
+				encoder.EncodeSlash |
+				encoder.EncodeBackSlash |
+				encoder.EncodeDoubleQuote |
+				encoder.EncodeInvalidUtf8),
+		}},
+	})
+}
+
+// Options defines the configuration for this backend
+type Options struct {
+	URL           string               `config:"url"`
+	User          string               `config:"user"`
+	Password      string               `config:"pass"`
+	LibraryName   string               `config:"library"`
+	LibraryKey    string               `config:"library_key"`
+	CreateLibrary bool                 `config:"create_library"`
+	Enc           encoder.MultiEncoder `config:"encoding"`
+}
+
+// Fs represents a remote seafile
+type Fs struct {
+	name                string       // name of this remote
+	root                string       // the path we are working on
+	libraryName         string       // current library
+	encrypted           bool         // Is this an encrypted library
+	rootDirectory       string       // directory part of root (if any)
+	opt                 Options      // parsed options
+	libraries           *cache.Cache // Keep a cache of libraries
+	librariesMutex      sync.Mutex   // Mutex to protect getLibraryID
+	features            *fs.Features // optional features
+	endpoint            *url.URL     // URL of the host
+	endpointURL         string       // endpoint as a string
+	srv                 *rest.Client // the connection to the one drive server
+	pacer               *fs.Pacer    // pacer for API calls
+	authMu              sync.Mutex   // Mutex to protect library decryption
+	createDirMutex      sync.Mutex   // Protect creation of directories
+	useOldDirectoryAPI  bool         // Use the old API v2 if seafile < 7
+	moveDirNotAvailable bool         // Version < 7.0 don't have an API to move a directory
+}
+
+// ------------------------------------------------------------
+
+// NewFs constructs an Fs from the path, container:path
+func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
+	// Parse config into Options struct
+	opt := new(Options)
+	err := configstruct.Set(m, opt)
+	if err != nil {
+		return nil, err
+	}
+	root = strings.Trim(root, "/")
+	isLibraryRooted := opt.LibraryName != ""
+	var libraryName, rootDirectory string
+	if isLibraryRooted {
+		libraryName = opt.LibraryName
+		rootDirectory = root
+	} else {
+		libraryName, rootDirectory = bucket.Split(root)
+	}
+
+	if !strings.HasSuffix(opt.URL, "/") {
+		opt.URL += "/"
+	}
+	if opt.Password != "" {
+		var err error
+		opt.Password, err = obscure.Reveal(opt.Password)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't decrypt user password")
+		}
+	}
+	if opt.LibraryKey != "" {
+		var err error
+		opt.LibraryKey, err = obscure.Reveal(opt.LibraryKey)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't decrypt library password")
+		}
+	}
+
+	// Parse the endpoint
+	u, err := url.Parse(opt.URL)
+	if err != nil {
+		return nil, err
+	}
+
+	f := &Fs{
+		name:          name,
+		root:          root,
+		libraryName:   libraryName,
+		rootDirectory: rootDirectory,
+		libraries:     cache.New(),
+		opt:           *opt,
+		endpoint:      u,
+		endpointURL:   u.String(),
+		srv:           rest.NewClient(fshttp.NewClient(fs.Config)).SetRoot(u.String()),
+		pacer:         getPacer(opt.URL),
+	}
+	f.features = (&fs.Features{
+		CanHaveEmptyDirectories: true,
+		BucketBased:             opt.LibraryName == "",
+	}).Fill(f)
+
+	ctx := context.Background()
+	serverInfo, err := f.getServerInfo(ctx)
+	if err != nil {
+		return nil, err
+	}
+	fs.Debugf(nil, "Seafile server version %s", serverInfo.Version)
+
+	// We don't support bellow seafile v6.0 (version 6.0 is already more than 3 years old)
+	serverVersion := semver.New(serverInfo.Version)
+	if serverVersion.Major < 6 {
+		return nil, errors.New("unsupported Seafile server (version < 6.0)")
+	}
+	if serverVersion.Major < 7 {
+		// Seafile 6 does not support recursive listing
+		f.useOldDirectoryAPI = true
+		f.features.ListR = nil
+		// It also does no support moving directories
+		f.moveDirNotAvailable = true
+	}
+
+	err = f.authorizeAccount(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if f.libraryName != "" {
+		// Check if the library exists
+		exists, err := f.libraryExists(ctx, f.libraryName)
+		if err != nil {
+			return f, err
+		}
+		if !exists {
+			if f.opt.CreateLibrary {
+				err := f.mkLibrary(ctx, f.libraryName, "")
+				if err != nil {
+					return f, err
+				}
+			} else {
+				return f, fmt.Errorf("library '%s' was not found, and the option to create it is not activated (advanced option)", f.libraryName)
+			}
+		}
+		libraryID, err := f.getLibraryID(ctx, f.libraryName)
+		if err != nil {
+			return f, err
+		}
+		f.encrypted, err = f.isEncrypted(ctx, libraryID)
+		if err != nil {
+			return f, err
+		}
+		if f.encrypted {
+			// If we're inside an encrypted library, let's decrypt it now
+			err = f.authorizeLibrary(ctx, libraryID)
+			if err != nil {
+				return f, err
+			}
+			// And remove the public link feature
+			f.features.PublicLink = nil
+		}
+	} else {
+		// Deactivate the cleaner feature since there's no library selected
+		f.features.CleanUp = nil
+	}
+
+	if f.rootDirectory != "" {
+		// Check to see if the root is an existing file
+		remote := path.Base(rootDirectory)
+		f.rootDirectory = path.Dir(rootDirectory)
+		if f.rootDirectory == "." {
+			f.rootDirectory = ""
+		}
+		_, err := f.NewObject(ctx, remote)
+		if err != nil {
+			if errors.Cause(err) == fs.ErrorObjectNotFound || errors.Cause(err) == fs.ErrorNotAFile {
+				// File doesn't exist so return the original f
+				f.rootDirectory = rootDirectory
+				return f, nil
+			}
+			return f, err
+		}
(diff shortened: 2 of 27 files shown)


backend/seafile/seafile.go
```diff
@@ -0,0 +1,1138 @@
...
+// callback returns an error then the listing will stop
+// immediately.
+func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) error {
+	if dir == "" && f.libraryName == "" {
+		libraries, err := f.listLibraries(ctx)
+		if err != nil {
+			return err
+		}
+		// Send the library list as folders
+		callback(libraries)
```
golangcibot, line 523: Error return value of `callback` is not checked (from `errcheck`)

backend/seafile/webapi.go
```diff
@@ -0,0 +1,888 @@
...
+	// Non-encrypted libraries are accepting the HTTP Range header,
+	// BUT encrypted libraries are simply ignoring it
+	if partialContent && resp.StatusCode == 200 {
+		// Partial content was requested through a Range header, but a full content was sent instead
+		f.rangeDownloadNotice.Do(func() {
+			fs.Logf(nil, "%s ignored our request of partial content. This is probably because encrypted libraries are not accepting range requests. Loading this file might be slow!", f.String())
+		})
+		if start > 0 {
+			// We need to read and discard the beginning of the data...
+			io.CopyN(ioutil.Discard, resp.Body, start)
```
golangcibot, line 457: Error return value of `io.CopyN` is not checked (from `errcheck`)

backend/seafile/seafile.go
```diff
@@ -0,0 +1,1138 @@
...
+			filePathInLibrary = path.Join(parentPathInLibrary, entryPath, entry.Name)
+		} else {
+			// In non-recursive mode, paths are build from the parameters
+			filePath = path.Join(parentPath, entry.Name)
+			filePathInLibrary = path.Join(parentPathInLibrary, entry.Name)
+		}
+		if entry.Type == api.FileTypeDir {
+			d := fs.
+				NewDir(filePath, time.Now()).
+				SetSize(int64(entry.Size)).
```
golangcibot, line 1015: unnecessary conversion (from `unconvert`)

creativeprojects:
For information, this is the docker-compose file I was using for testing locally:

```yaml
version: '2.0'
services:
  db:
    image: mariadb:10.1
    container_name: seafile-mysql
    environment:
      - MYSQL_ROOT_PASSWORD=db_dev
      - MYSQL_LOG_CONSOLE=true
    volumes:
      - ${SEAFILE_TEST_DATA}/seafile-mysql/db:/var/lib/mysql
    networks:
      - seafile-net

  memcached:
    image: memcached:1.5.6
    container_name: seafile-memcached
    entrypoint: memcached -m 256
    networks:
      - seafile-net

  seafile:
    image: seafileltd/seafile-mc:latest
    container_name: seafile
    ports:
      - "8088:80"
      - "8080:8080"
    volumes:
      - ${SEAFILE_TEST_DATA}/seafile-data:/shared
    environment:
      - DB_HOST=db
      - DB_ROOT_PASSWD=db_dev
      - TIME_ZONE=Etc/UTC
      - SEAFILE_ADMIN_EMAIL=me@example.com
      - SEAFILE_ADMIN_PASSWORD=asecret
      - SEAFILE_SERVER_LETSENCRYPT=false
      - SEAFILE_SERVER_HOSTNAME=127.0.0.1:8088
    depends_on:
      - db
      - memcached
    networks:
      - seafile-net

networks:
  seafile-net:
    ipam:
     config:
       - subnet: 10.5.0.0/16
         gateway: 10.5.0.1
```

Looking at it today, I can't remember why I fixed the IP range of the network. It might have something to do with the Seafile webdav server I was also running for comparison. Oh well it does work like that 👍

ncw:
That is looking very nice!

I haven't reviewed the code in detail yet, but at first glance it looks very good :-)

Can you split the go module and vendor changes into another commit please? ([docs](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency) )

It would be perfect if we could integrate your docker compose into the rclone test framework

Rclone has a system for starting docker containers to run test servers - see https://github.com/rclone/rclone/tree/master/fstest/testserver/init.d

You make a bash script which starts and stops the docker container and outputs configuration for the backend needed. Just call the file `TestSeafile` and it will automatically get called in the tests.

The examples in there all use a single docker container so it would need a bit of adapting for docker compose but I don't think it would be very hard.

creativeprojects:
> Can you split the go module and vendor changes into another commit please? ([docs](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency) )
> 

I'm not too sure how to do that. I'm not used to rewrite history in git. Would re-creating a new Seafile branch do the trick?

> The examples in there all use a single docker container so it would need a bit of adapting for docker compose but I don't think it would be very hard.

Right, I'll have a look at integrating docker-compose into these automated tests 👍

ncw:
> > Can you split the go module and vendor changes into another commit please? ([docs](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency) )
> 
> I'm not too sure how to do that. I'm not used to rewrite history in git. Would re-creating a new Seafile branch do the trick?

I can do it for you just before merging if you want. If you want to have a go [here is an article](https://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/) on what you need to do. Remember if everything goes wrong `git rebase --abort`!

If you do that then you'll need to `--force` push your branch up to github. That is the workflow we encourage from contributors normally, but I'm happy to help as I realise that git is hard!

> > The examples in there all use a single docker container so it would need a bit of adapting for docker compose but I don't think it would be very hard.
> 
> Right, I'll have a look at integrating docker-compose into these automated tests

That would be fantastic!

creativeprojects:
I think I nailed it 👍 

Except that I don't quite understand the tests failing on the CI build...

ncw:
I think you did nail it!

For the unreliable CI test see here: https://github.com/rclone/rclone/issues/4171

I'm having trouble running the docker-compose locally, I get these errors

```
2020/04/27 11:23:26 TestSeafile:: Failed to start test server: failed to run /home/ncw/go/src/github.com/rclone/rclone/fstest/testserver/init.d/TestSeafile start

The MYSQL_ROOT_PASSWORD variable is not set. Defaulting to a blank string.
The SEAFILE_TEST_DATA variable is not set. Defaulting to a blank string.
The SEAFILE_PORT variable is not set. Defaulting to a blank string.
The SEAFILE_ADMIN_EMAIL variable is not set. Defaulting to a blank string.
The SEAFILE_ADMIN_PASSWORD variable is not set. Defaulting to a blank string.

The Compose file '/home/ncw/go/src/github.com/rclone/rclone/fstest/testserver/init.d/seafile/docker-compose.yml' is invalid because:
services.seafile.ports is invalid: Port ranges don't match in length
```
I checked that the environment variables do get loaded here

```
# environment variables automatically loaded by docker-compose
. $(dirname "$0")/seafile/.env
```

But they don't seem to get passed on? Any ideas?

creativeprojects:
> But they don't seem to get passed on? Any ideas?

This is very weird. I've been putting all my environment variables in `.env` for years and I never had any issue with it.

I've moved them back into the startup script. It still works locally 👍

ncw:
> This is very weird. I've been putting all my environment variables in `.env` for years and I never had any issue with it.

Ah, I think the problem was that they weren't `export`-ed

The tests are running now - I'll report back when they've finished...

backend/seafile/seafile.go
```diff
@@ -0,0 +1,1144 @@
...
+
+	ctx := context.Background()
+	serverInfo, err := f.getServerInfo(ctx)
+	if err != nil {
+		return nil, err
+	}
+	fs.Debugf(nil, "Seafile server version %s", serverInfo.Version)
+
+	// We don't support bellow seafile v6.0 (version 6.0 is already more than 3 years old)
+	serverVersion := semver.New(serverInfo.Version)
```
buengese, line 192: Just a quick note: If this is the only place this is used, is there really a need to add yet another dependency to rclone? I'm a big fan of reducing dependency bloat. I realize that this is a very small and simple library but for something that could be done with one simple regex this still feels kinda unnecessary for me.

ncw:
The test was successful!

image


There were quite a few retries of tests (the `#2`s).

```
==> seafile-backend.seafile-TestSeafile-2.txt <==
"go test -v -timeout 30m0s -remote TestSeafile:
-test.run '^TestIntegration$/^FsMkdir$/^FsPutFiles$/^(FromRoot|FsIsFile|FsListDirRoot|FsListRDirRoot)$/^(List|ListEntries|ListR|Put)$/^Remove$'" - Starting (try 2/5)

==> seafile-fs.operations-TestSeafile-2.txt <==
"./operations.test -test.v -test.timeout 30m0s -remote TestSeafile:
-test.run '^(TestCopyFileBackupDir|TestCopyFileCompareDest|TestCopyFileCopyDest|TestMoveFileBackupDir)$'" - Starting (try 2/5)

==> seafile-fs.operations-TestSeafile-fastlist-2.txt <==
"./operations.test -test.v -test.timeout 30m0s -remote TestSeafile: -fast-list
-test.run '^(TestCopyFileBackupDir|TestCopyFileCompareDest|TestCopyFileCopyDest|TestMoveFileBackupDir)$'" - Starting (try 2/5)

==> seafile-fs.sync-TestSeafile-2.txt <==
"./sync.test -test.v -test.timeout 30m0s -remote TestSeafile:
-test.run '^(TestServerSideCopy|TestServerSideMove|TestServerSideMoveDeleteEmptySourceDirs|TestServerSideMoveWithFilter|TestSyncBackupDir|TestSyncCompareDest|TestSyncCopyDest)$'" - Starting (try 2/5)

==> seafile-fs.sync-TestSeafile-fastlist-2.txt <==
"./sync.test -test.v -test.timeout 30m0s -remote TestSeafile: -fast-list
-test.run '^(TestServerSideCopy|TestServerSideMove|TestServerSideMoveDeleteEmptySourceDirs|TestServerSideMoveWithFilter|TestSyncCompareDest|TestSyncCopyDest)$'" - Starting (try 2/5)
```

The sync retries and the operations retries are almost identical with and without --fast-list which maybe indicates a persistent problem?

Do they always fail like that for you? If so then they'll probably fail for users.... Those test retries are for working around networking problems and eventual consistency neither of which we should have here working on a private test instance.

backend/seafile/seafile.go
ncw: That is a reasonable point, especially if we only use the major version.

Conversely I note we already have quite a lot of semvar parsing code already here:

https://github.com/rclone/rclone/blob/master/fs/version/version.go

Which might be a good candidate for replacement by that library...

creativeprojects:
> The sync retries and the operations retries are almost identical with and without --fast-list which maybe indicates a persistent problem?
> 
> Do they always fail like that for you? If so then they'll probably fail for users.... Those test retries are for working around networking problems and eventual consistency neither of which we should have here working on a private test instance.

Yes they quite often do. But it's not always the same tests. And I've seen them running without error either.

Unless I missed something, my investigations were showing that the tests are too fast for a "normal" utilisation. By default Seafile will block at 300 requests per minute: If we throw 300 requests in 10 seconds, we have 50 seconds to wait. I tried to play with the parameters of the pacer with no success. Maybe it needs a specific pacer?

If we remove the request throttling, all the tests pass on the first attempt (and it's quite fast as well).
But the thing is, I never noticed there was such requests throttling until I started to investigate the failing tests. It's not a well advertised configuration setting. Which make me think it's kind of cheating if we remove it for the tests.

Maybe I should investigate a specific pacer?

backend/seafile/seafile.go
creativeprojects: Yes this is a good point, I never noticed the version class in the code.
I could update to use the in-house code, or alternatively replace the in-house implementation by the semver library?

ncw:
>  By default Seafile will block at 300 requests per minute: If we throw 300 requests in 10 seconds, we have 50 seconds to wait. I tried to play with the parameters of the pacer with no success. Maybe it needs a specific pacer?

300 requests / minute is 5 per second

Your current pacer settings are...

```
	minSleep          = 100 * time.Millisecond
	maxSleep          = 10 * time.Second
```

So 10 per second.  Raising minSleep to 200*time.Millisecond should be 5 per second

**However** if the requests are failing it means that we are not identifying the fact that they are retriable errors correctly. Rclone should see a rate limit error and back off and retry.

I see your `shouldRetry` function is copied from the box backend... Is that correct? Does seafile return 429 errors or something else? I haven't managed to capture one in a log yet!

BTW I ran it again and it worked perfectly in 4 minutes!

creativeprojects:
> I see your `shouldRetry` function is copied from the box backend... Is that correct? Does seafile return 429 errors or something else? I haven't managed to capture one in a log yet!
> 

Yes this is right, it is returning a 429 error with a Retry-After header. I had loads of these to start with.
And you're right the pacer does its work perfectly: I just tried it myself again and I haven't got any. It was an issue originally but it's no longer the case.

> BTW I ran it again and it worked perfectly in 4 minutes!

I know!

It sounds like some kind of concurrency issue somewhere...

creativeprojects:
Found it!

Seafile lets you create multiple library using the same name. It's probably checked on the client side, so the API is cool with it.

In a normal operation, we only have **one fs** at a time, and inside the same **fs** it's all managed with some mutex.

Now, at the start of the tests, there are more than one  **fs** instantiated, and there's a race condition to create the test library.

For now, I disabled automatic library creation, and I'm using the default library to run the tests instead of creating another one. So all the tests ran fine on my machine in 2 minutes.

In a normal usage, are there some cases where we would need more than one  **fs** pointing to the same server? I guess that would be the case if we have 2 remotes pointing to the same actual server. I need to see how to fix that in which case.

ncw:
> Found it!

Great!

> Seafile lets you create multiple library using the same name. It's probably checked on the client side, so the API is cool with it.

What is a Seafile library? Is it like an overall storage container?
 
> In a normal operation, we only have **one fs** at a time, and inside the same **fs** it's all managed with some mutex.
> 
> Now, at the start of the tests, there are more than one **fs** instantiated, and there's a race condition to create the test library.

Ah I see...

Presumably the library only needs to be created once?

> 
> For now, I disabled automatic library creation, and I'm using the default library to run the tests instead of creating another one. So all the tests ran fine on my machine in 2 minutes.
> 
> In a normal usage, are there some cases where we would need more than one **fs** pointing to the same server? I guess that would be the case if we have 2 remotes pointing to the same actual server. I need to see how to fix that in which case.

The user might be copying from one library to another. However if I'm understanding correctly it is unlikely that the user would be creating two libraries at the same time and causing the problem above?

Now the integration tests are passing nicely, I'll give the code another review now

ncw:
The integration tests failed for me with

```
    TestIntegration: fstests.go:389: 
        	Error Trace:	fstests.go:389
        	            				seafile_test.go:13
        	            	library 'My Library' was not found, and the option to create it is not activated (advanced option)
        	Test:       	TestIntegration
        	Messages:   	unexpected error: library 'My Library' was not found, and the option to create it is not activated (advanced option)
```

:-(

creativeprojects:
> What is a Seafile library? Is it like an overall storage container?

Yes, it's a bucket

> Presumably the library only needs to be created once?

Exactly. Parallels tests will start creating their own library and work in it. Until the point one test is copying files generated by a previous test - hence possibly in a different library

> The user might be copying from one library to another. However if I'm understanding correctly it is unlikely that the user would be creating two libraries at the same time and causing the problem above?

Yes I think it would be very unlikely. Also the auto-create library setting is _advanced_ so most people will leave it off.

> Now the integration tests are passing nicely, I'll give the code another review now

Doh! I forgot the default library is created the first you actually **login**. Back to the drawing board. Sorry about that :-|

backend/seafile/seafile.go
```diff
@@ -0,0 +1,1132 @@
...
+func (f *Fs) About(ctx context.Context) (usage *fs.Usage, err error) {
+	accountInfo, err := f.getUserAccountInfo(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	usage = &fs.Usage{
+		Used: fs.NewUsageValue(accountInfo.Usage), // bytes in use
+	}
+	if accountInfo.Total > 0 {
```
ncw, line 741: Does `Total <=0` here mean that there is no quota?

backend/seafile/seafile.go
```diff
@@ -0,0 +1,1132 @@
...
+		remote:        remote,
+		libraryID:     libraryID,
+		pathInLibrary: folder,
+		size:          size,
+		modTime:       modTime,
+	}
+	return object
+}
+
+// Check the interfaces are satisfied
```
ncw, line 1118: That is a nice set of optional interfaces!

No `DirMove` though? So not possible to rename directories?

backend/seafile/seafile_internal_test.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+func TestEmptySplitPathFromRootTarget(t *testing.T) {
+	fs := &Fs{
+		libraryName:   "",
+		rootDirectory: "",
+	}
+	libraryName, directory := fs.splitPath("")
+	if libraryName != "" {
+		t.Errorf("library name should be empty but found '%s'", libraryName)
```
ncw, line 15: In the rclone project we usually use assert and require from testify

So this could become

```go
assert.Equal(t, "", libraryName)
assert.NotEqual(t, "", directory)
```

Most of the time there are no need for comment strings as the error string takes you straight to the code

I'd have put all of the splitSpath tests into one table driven test - maybe like this one

https://github.com/rclone/rclone/blob/74d9dabdff69d7c275db1f09be9767b331d74d4d/fs/sizesuffix_test.go#L15-L34

You can add a `name` parameter to the struct and use an inner `t.Run` if you want like this one for more discoverability

https://github.com/rclone/rclone/blob/74d9dabdff69d7c275db1f09be9767b331d74d4d/fstest/fstests/fstests.go#L573-L597

backend/seafile/seafile_internal_test.go
```diff
@@ -0,0 +1,187 @@
...
+func TestSplitPathIntoSlice(t *testing.T) {
+	testData := map[string][]string{
+		"1":     {"1"},
+		"/1":    {"1"},
+		"/1/":   {"1"},
+		"1/2/3": {"1", "2", "3"},
+	}
+	for input, expected := range testData {
+		output := splitPath(input)
+		if !equalSlice(output, expected) {
```
ncw, line 169: You can just use `assert.Equal(t, expected, output)` (in that order!) here so the `equalSlice` function isn't necessary.

docs/content/seafile.md
```diff
@@ -0,0 +1,242 @@
...
+Invalid UTF-8 bytes will also be [replaced](/overview/#invalid-utf8),
+as they can't be used in JSON strings.
+
+### Seafile and rclone link ###
+
+Rclone supports generating share links for non-encrypted libraries only.
+They can either be for a file or a directory:
+
+```
+./rclone link seafile:seafile-tutorial.doc
```
ncw, line 223: Can we lose the `./` at the start of the example as the rest of the docs assume the user has rclone on the PATH?

docs/content/seafile.md
```diff
@@ -0,0 +1,242 @@
...
+```
+./rclone link seafile:dir
+http://my.seafile.server/d/9ea2455f6f55478bbb0d/
+```
+
+Please note a share link is unique for each file or directory. If you run a link command on a file/dir
+that has already been shared, you will get the exact same link.
+
+
+<!--- autogenerated options start - DO NOT EDIT, instead edit fs.RegInfo in backend/seafile/seafile.go then run make backenddocs -->
```
ncw, line 239: If you want to fill this bit in run `make backenddocs` then `git checkout` all of the other backend docs which got changes!  You don't need to though - I'll do it at release time!

fstest/test_all/config.yaml
```diff
@@ -264,3 +264,9 @@ backends:
    remote:   "TestMailru:"
    subdir:   false
    fastlist: false
+ - backend:  "seafile"
+   remote:   "TestSeafile:"
+   fastlist: true
+   ignore:
+     - TestVFSStatfs
```
ncw, line 271: I think this test was broken and now I've fixed it so you can probably remove this!

fstest/testserver/init.d/TestSeafile
```diff
@@ -0,0 +1,52 @@
...
+# environment variables passed on docker-compose
+export MYSQL_ROOT_PASSWORD=pixenij4zacoguq0kopamid6
+export SEAFILE_ADMIN_EMAIL=seafile@rclone.org
+export SEAFILE_ADMIN_PASSWORD=pixenij4zacoguq0kopamid6
+export SEAFILE_PORT=8088
+export SEAFILE_TEST_DATA=/tmp/seafile-test-data
+
+# local variables
+NAME=seafile
+SEAFILE_IP=127.0.0.1
```
ncw, line 14: This is binding to the world I noticed

```
tcp6       0      0 :::8088                 :::*                    LISTEN      62984/docker-proxy  
```

There are rather a lot of things going on on the integration server and I try to stop the docker stuff binding to IPs owned by the host. It shouldn't be accessible from outside the server. 

We could use the `docker_ip` function from `docker.bash` and just turn off the external port forwarding?

At minimum I'd like it only to bind to 127.0.0.1 not the external IPs.

Oh and while experimenting with this I discovered what you meant

```
$ rclone lsd seafile:
           0 2020-04-30 10:40:33        -1 TestSeafile
          12 2020-04-30 10:40:33        -1 TestSeafile
          30 2020-04-30 10:40:33        -1 TestSeafile
           0 2020-04-30 10:40:33        -1 TestSeafile
          40 2020-04-30 10:40:33        -1 TestSeafile
```



I note that 10.5.0.1 seems to work!

creativeprojects: Doing routing back to a container like you do only works on linux machines.
On Windows and Mac the containers are behind a NAT so you can't connect back to the container IP directly.
I couldn't run the tests locally with the default configuration.

I can bind the ports to 127.0.0.1 so that would work everywhere.
But if you prefer I will change the configuration to use routing and I'll setup and run the tests manually on my machine.

backend/seafile/seafile.go
creativeprojects: This is right, by default a user has no quota and it's not returning the size of the disk (returning -2 for no quota)

fstest/testserver/init.d/TestSeafile
ncw: Bind it to 127.0.0.1 is fine if that works for you.

I didn't know that about the direct routing only working on linux machines, maybe I should reconsider how those tests work...

backend/seafile/object.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Remote returns the remote string
+func (o *Object) Remote() string {
+	return o.remote
+}
+
+// ModTime returns last modified time
+func (o *Object) ModTime(context.Context) time.Time {
+	return time.Now()
```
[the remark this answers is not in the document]
creativeprojects: It's funny I remember having issues running the tests with the modification time (you can see it, but you cannot set it). I've put it back and it all works :-| I must be going mad

backend/seafile/seafile.go
creativeprojects: DirMove is a difficult one. The API works this way:

`move path1/dir1 to path2` will give you `path2/dir1` or `path2/dir1 (1)` if `dir1` was existing, but the API is also not telling you that it renamed it to `dir1 (1)`.

With this behaviour it's also impossible to simply change the name of the directory.

I can look into it again with a fresh mind though

ncw:
Sorry I broke master with a careless merge - can you rebase and try your tests again.

creativeprojects:
> I've put some comments inline with things I noticed.

Thanks for all these comments, I'm going through them :)

> If you didn't already it would be worth doing some uploads and downloads with `--dump openfiles` just to check there are no open file handles on exit - it is an easy mistake to make with the rest library. I did check it with mk1 eyeball but the `--dump openfiles` test is a good one too!

What should I look for?
I notice there are some leftover, but possibly just from the go runtime

ncw:
> > I've put some comments inline with things I noticed.
> 
> Thanks for all these comments, I'm going through them :)

Let me know when you are ready for another review :-) It feels like we are getting very close!

> > If you didn't already it would be worth doing some uploads and downloads with `--dump openfiles` just to check there are no open file handles on exit - it is an easy mistake to make with the rest library. I did check it with mk1 eyeball but the `--dump openfiles` test is a good one too!
> 
> What should I look for?

Sockets are the easiest thing to leak - they look like this in the `lsof` output

```
rclone  337948  ncw    9u     IPv4            3067009      0t0      TCP localhost:5572->localhost:36838 (ESTABLISHED)
```

> I notice there are some leftover, but possibly just from the go runtime

backend/seafile/seafile.go
creativeprojects: I think I found a way to make the DirMove working. I need a bit more testing but it looks good so far 👍

ncw:
Are you ready for another review?

creativeprojects:
Almost. I made 3 test remotes:
- Seafile v6 is working perfectly.
- Both v7 and v7 encrypted are sometimes returning annoying 500 errors when uploading a file. I'm trying to find a pattern but I think it's a concurrency issue in the server.

Maybe I could try to instantiate a pacer per remote, as opposed to a pacer per **fs**...

creativeprojects:
> Are you ready for another review?

Right, I think that's all good!

The tests do trigger some really random errors sometimes though. It's only happening in version 7 of the server (latest) and I don't know how to avoid them (errors 443, 500 and 520 - 443 being "out of quota": interesting since I haven't setup quota on users...)

ncw:
I think your code is looking great now and ready to merge :-) Can you squash it again into two commits (vendor + code) please? Also can you fix the commit text of the code commit I don't need to see the squashed history in there. Thanks!

I had a problem with the integration tests though. The integration test server runs all the tests at once, so the servers all need to be on different ports so they can run concurrently.

If you want to try this locally try

    test_all -backends seafile

Can you put the V6 server on a high port, not on :80 as that causes a problem on my local machine!

Note also that the `_connect` output is just supposed to be an IP:PORT - it isn't a URL. We could make a _connect_url if that was helpful.

Looking really good and it looks like we'll get it into 1.52 which is great :-)

creativeprojects:
> Note also that the `_connect` output is just supposed to be an IP:PORT - it isn't a URL. We could make a _connect_url if that was helpful.
> 
Oh I see I understand now! I had to put a timer after starting the container because the tests were too fast to start running.
What's happening is that the port is opened by the reverse-proxy in front of the server, even though the server is not started yet. So a TCP connection is good, but an HTTP connection would return 502 Bad Gateway.
I think that would be a good addition to the test framework (maybe something I can do later)

creativeprojects:
I think we're all good 👍 

Screenshot 2020-05-05 at 22 16 43

ncw:
> I think we're all good +1

I think so too :-)

I'll merge this now - thank you very much for a most excellent contribution :-)

Check the output of the integration tests tomorrow: https://pub.rclone.org/integration-tests/current/

I'll put you down in the [MAINTAINERS](https://github.com/rclone/rclone/blob/master/MAINTAINERS.md) file for the seafile backend if that is OK?

You are welcome to join the project if you would like. It is quite a busy project so I'm always trying to recruit new people to help! Let me know and I'll send you an invite if you want.

creativeprojects:
> I'll put you down in the [MAINTAINERS](https://github.com/rclone/rclone/blob/master/MAINTAINERS.md) file for the seafile backend if that is OK?
> 

Sure 👍 Thanks

> You are welcome to join the project if you would like. It is quite a busy project so I'm always trying to recruit new people to help! Let me know and I'll send you an invite if you want.

To be honest that was my first contribution of this size to an open source project: I needed the support for Seafile, and I wanted to learn Go. Mission accomplished!

Actually I really like Go: so, yes I can probably give some of my spare time :)

ncw:
> > I'll put you down in the [MAINTAINERS](https://github.com/rclone/rclone/blob/master/MAINTAINERS.md) file for the seafile backend if that is OK?
> 
> Sure  Thanks

I've done that.
 
> > You are welcome to join the project if you would like. It is quite a busy project so I'm always trying to recruit new people to help! Let me know and I'll send you an invite if you want.
> 
> To be honest that was my first contribution of this size to an open source project: I needed the support for Seafile, and I wanted to learn Go. Mission accomplished!

It was a really great contribution :-)
 
> Actually I really like Go: so, yes I can probably give some of my spare time :)

I've sent you an invite - thank you!

Take a look over the MAINTAINERS file  for what maintainers do... But really how much you have time for is up to you so if you want to work on something go with your interests!

It would be a good idea if you announce on [the rclone forum](https://forum.rclone.org/) that the seafile backend is in [the latest beta](https://beta.rclone.org) - we might get some more testers that way. I'll put a link to the page on twitter also.

Issue #3422, opened 2019-08-08
Allow uploading by folder id on Box


YenForYang:
#### What problem are you are trying to solve?
Nearly a year ago I posted an [Allow uploading by folder id on Google Drive](https://github.com/rclone/rclone/issues/2329), which eventually lead to my discovery of `RCLONE_DRIVE_ROOT_FOLDER_ID` and `--drive-root-folder-id`. 

Well, I've been using Box lately, which also has folder ids it seems. These IDs appear to be numeric e,g,
`https://XXXXX.app.box.com/folder/1234567890`

#### How do you think rclone should be changed to solve that?

I'm not aware of any root folder id variable or flag on `rclone` for the Box remote. But I'm guessing it can be added?

ncw:
This would be relatively easy to do - you'd need to replace `rootID` which is a constant `0` at the moment with an option `--box-root-folder-id`.

Note that you can see the IDs of folders in box with

    rclone lsf --dirs-only -F ip TestBox:

phillipahereza:
@ncw can I work on this?

ncw:
> @ncw can I work on this?

Absolutely :-)  Do you need any help more than the hint above?

phillipahereza:
@ncw I think I should be able to manage with the hint above, if not I'll hit you up

ncw:
Add it to the options in the box backend and it will appear in the configurator and as a command line option. It should probably be an advanced option.

Then people can put it in the config file or in the command line or as an environment variable whatever they choose.

ncw:
I've merged @Sunil-P 's fix to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.52

PR #4145, opened 2020-04-18
Added support for interchangable root folder for Box backend

Sunil-P:

Add support for non root folder id to Box, similar to google drive.

https://github.com/rclone/rclone/issues/3422

#### Checklist


backend/box/box.go
```diff
@@ -121,6 +120,11 @@ func init() {
 		}, {
 			Name: config.ConfigClientSecret,
 			Help: "Box App Client Secret\nLeave blank normally.",
+		}, {
+			Name:     "root_folder_id",
+			Help:     "Fill in for rclone to use a non root folder as its starting point.",
+			Default:  0,
```
ncw, line 126: The default should be "0" as it is a string

backend/box/box.go
```diff
@@ -399,7 +403,11 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		return err
 	})
 
-	// Get rootID
+	// Get rootFolderID
```
ncw, line 406: Can you add it to the `Options` struct like this

```go
type Options struct {
    // ...
    RootFolderID              string               `config:"root_folder_id"`
}
```

Instead of using `Get`, then you can use use `f.opt.RootFolderID` in place of `rootID`

Sunil-P:
@ncw  can you tell me why modules_race failed, and how to run this test on my local machine.
Also, is it possible to restart a test on the CI/CD here ? Unfamiliar with git actions..

ncw:
That looks great now - thank you :-) I'll merge it now.

Sunil-P:
@ncw Hi, can you tell me why build was failing ? I have been trying to restart builds manually and that hasn't been very fun

ncw:
> @ncw Hi, can you tell me why build was failing ? I have been trying to restart builds manually and that hasn't been very fun

There are some tests which sometimes fail. If you look at the logs of the builds you can see what they are. Rclone has maybe 2,000 tests - it only takes one to spoil the show!

I haven't had time to look through the persistent failures recently - any help much appreciated!

> Also is there any way to restart build without pushing commit ? Unfamiliar with git actions

There is a button you can press on the github actions page but it rebuilds everything not just the failing item.

You can run most of the tests locally either with `go test -v` in the directory you are changing. If you want to run what the CI does then do `make check` (you may have to do `make build_dep` first to get the tools) and also run `make quicktest`. That covers most of the tests.

PR #4155, opened 2020-04-22
Remove "iat" from box jwt claims and fix error handling

dbramwell:

When using box jwt auth from a server that has an incorrect timestamp. It can result in the following error:
`Current date/time MUST be after or equal to the not-before date\/time listed in the 'nbf' or 'iat' claim`

The `iat` claim doesn't actually have to be set on the request for box to accept it, and so it makes sense to remove it.

When I first hit this issue I wasn't receiving any error messages from rclone, so I've fixed that up too.

Nope

#### Checklist


ncw:
Seems like a pragmatic solution - thank you :-)

Issue #4114, opened 2020-04-09
[Feature Request] box.com authentication using app token


lonix1:
As [discussed here](https://forum.rclone.org/t/can-i-force-rclone-to-use-boxs-cli-api/15463/6) with @ncw, it would be helpful if we could authenticate with box.com using an app token. Right now auth is performed using oauth.

When running services on a headless server, it is more convenient (and often more common) to use an app token, which is generated by hand on the service's website, then saved in scripts/code and included in every REST call to the service.

Box [has such a feature](https://developer.box.com/guides/authentication/app-token/). It's typically used in calls made with `curl`, no oauth required.

If implemented, it would greatly simplify configuration in headless environments.

(As an aside, it is quite possible many other storage services also support this auth option.)

ncw:
Since you were currently working on box auth - I wondered if you'd like to look at this one @dbramwell  ?

It should be a question of adding the key straight to the client with `f.rest.SetHeader("Authorization", "Bearer "+f.opt.APIKey)` and disabling the oauth config.

lonix1:
My understanding of rclone is that you support the "lowest common denominator" - i.e. those features that exist on all storage services.

Is it even possible to support this auth mechanism, just for box?

dbramwell:
@ncw Sure, I'll take a look. Seem's simple enough, but as always, expect a bunch of questions!

ncw:
Thank you @dbramwell  and happy to answer questions :-)

dbramwell:
Might have jumped the gun a bit here. Between my permissions in Box at work and the £8/month cost of adding `App Token` functionality in a personal account, I have no way to create an App Token to test this.

lonix1:
@dbramwell You can get a free account on box. Their entry-level is a free 10GiB account.

dbramwell:
Ah Thanks. Think I was looking at "Admin Console" rather than "Dev Console". Interestingly now that I've created an app in the dev console I can also get into the admin console...

Anyway, I'll have a go at this at some point this week

lonix1:
The thing I don't like about box's dev workflow: all files are uploaded to the dev console and can only be accessed there. So you can't see them in the normal console (was very confusing at first).

So if you upload files and don't understand why they don't show, it's because they're in the dev console.

ncw:
I've put @dbramwell 's change into a branch so it builds a beta for you to try @lonix1 

[...]

PR #4157, opened 2020-04-23
box: allow authentication with access token

dbramwell:

Allow authentication with Box using an access token

[4114](https://github.com/rclone/rclone/issues/4114)

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -86,13 +86,16 @@ func init() {
 		Config: func(name string, m configmap.Mapper) {
 			jsonFile, ok := m.Get("box_config_file")
 			boxSubType, boxSubTypeOk := m.Get("box_sub_type")
+			boxAccessToken, boxAccessTokenOk := m.Get("access_token")
 			var err error
+			// If using box config.json, use JWT auth
 			if ok && boxSubTypeOk && jsonFile != "" && boxSubType != "" {
 				err = refreshJWTToken(jsonFile, boxSubType, name, m)
 				if err != nil {
 					log.Fatalf("Failed to configure token with jwt authentication: %v", err)
 				}
-			} else {
+				// Else, if not using an access token, use oauth2
+			} else if boxAccessToken == "" || !boxAccessTokenOk {
 				err = oauthutil.Config("box", name, m, oauthConfig)
 				if err != nil {
 					log.Fatalf("Failed to configure token with oauth authentication: %v", err)
@@ -113,6 +116,9 @@ func init() {
 		}, {
 			Name: "box_config_file",
 			Help: "Box App config.json location\nLeave blank normally.",
+		}, {
+			Name: "access_token",
+			Help: "Box App Primary Access Token\nLeave blank normally.",
 		}, {
 			Name:    "box_sub_type",
 			Default: "user",
@@ -245,6 +251,7 @@ type Options struct {
 	CommitRetries int                  `config:"commit_retries"`
 	Enc           encoder.MultiEncoder `config:"encoding"`
 	RootFolderID  string               `config:"root_folder_id"`
+	AccessToken   string               `config:"access_token"`
 }
 
 // Fs represents a remote box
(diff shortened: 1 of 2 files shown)


backend/box/box.go
```diff
@@ -402,24 +417,27 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		CanHaveEmptyDirectories: true,
 	}).Fill(f)
 	f.srv.SetErrorHandler(errorHandler)
+	f.srv.SetHeader("Authorization", "Bearer "+f.opt.AccessToken)
```
ncw, line 420: Do we want to do this always? It seems like we'd only want to set this if `f.opt.AccessToken` was valid

backend/box/box.go
```diff
@@ -246,6 +251,7 @@ type Options struct {
 	CommitRetries int                  `config:"commit_retries"`
 	Enc           encoder.MultiEncoder `config:"encoding"`
 	RootFolderID  string               `config:"root_folder_id"`
+	AccessToken   string               `config:"box_primary_access_token"`
```
ncw, line 254: Can you remove the `box_` from this otherwise the generated config flags will stutter `--box-box-primary-access-token`

Would calling it `access_token` be OK? Just to shorten it a bit?

backend/box/box.go
```diff
@@ -384,16 +390,25 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
...
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to configure Box")
+
+	boxAccessToken, boxAccessTokenOk := m.Get("box_primary_access_token")
+
+	client := &http.Client{}
+	var ts *oauthutil.TokenSource
+
+	// If not using an accessToken, create an oauth client and tokensource
+	if !boxAccessTokenOk || boxAccessToken == "" {
```
ncw, line 400: I would have thought `f.opt.AccessToken == ""` would be good enough here? A blank access token will never be valid will it? You can lose the `m.Get` above as well if so.

backend/box/box.go
```diff
@@ -384,16 +390,25 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	}
 
 	root = parsePath(root)
-	oAuthClient, ts, err := oauthutil.NewClient(name, m, oauthConfig)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to configure Box")
+
+	boxAccessToken, boxAccessTokenOk := m.Get("box_primary_access_token")
+
+	client := &http.Client{}
```
ncw, line 396: This should be `client := fshttp.NewClient(fs.Config)` so the default is to pick up rclone's http client, rather than the default http.Client. Using the default client means that `--tpslimit` and `--dump headers` etc won't work.

backend/box/box.go
dbramwell: Oh yeah, I'll remove `box_`. There's a "Secondary Access Token" in the box admin panel which is used to roll over the primary access token. Wasn't sure if people would get them confused. I guess I could change it to `access_token` and rely on people reading the help message?

backend/box/box.go
dbramwell: It still worked even when using another auth method and having an empty AccessToken, but you're right, I probably shouldn't assume that'll always be the case

dbramwell:
Awesome, no worries for the delay. I'll take a look over the weekend

ncw:
> Awesome, no worries for the delay. I'll take a look over the weekend

Thank you :-)

dbramwell:
Sorry for dragging my feet over this a little. I've made changes in line with what you requested. Thanks for the advice!

ncw:
I've fixed up the conflict, squashed the commits and fixed a bug spotted by the linter (the client wasn't being used after being created by the oauthutil).

I've merged this to the fix-4114-box-token branch so we can get the user to try it then I'll merge it to master.

Thank you very much for doing this - much appreciated :-)

PR #4158, opened 2020-04-23
drive: Added rc support for changing the service-account-file

darthShadow:

Add rc support for changing the service-account-file

https://forum.rclone.org/t/edit-service-account-while-mounted/15760

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -36,6 +36,7 @@ import (
 	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/fs/rc"
 	"github.com/rclone/rclone/fs/walk"
 	"github.com/rclone/rclone/lib/dircache"
 	"github.com/rclone/rclone/lib/encoder"
@@ -542,6 +543,8 @@ type Fs struct {
 	exportExtensions []string           // preferred extensions to download docs
 	importMimeTypes  []string           // MIME types to convert to docs
 	isTeamDrive      bool               // true if this is a team drive
+
+	configMapper configmap.Mapper // TODO: Figure a way around this
 }
 
 type baseObject struct {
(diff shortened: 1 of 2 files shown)


backend/drive/drive.go
```diff
@@ -542,6 +543,8 @@ type Fs struct {
 	exportExtensions []string           // preferred extensions to download docs
 	importMimeTypes  []string           // MIME types to convert to docs
 	isTeamDrive      bool               // true if this is a team drive
+
+	configMapper     configmap.Mapper   // TODO: Figure a way around this
```
darthShadow, line 547: Any way to avoid this and still be able to reference the mapper for creating the oAuthClient?

fs/fs.go
```diff
@@ -1286,6 +1286,12 @@ func NewFs(path string) (Fs, error) {
 	if err != nil {
 		return nil, err
 	}
+	// TODO: Fix import for rc
+	// f, err := fsInfo.NewFs(configName, fsPath, config)
+	// if err != nil && f.reload != nil {
+	// 	rc.AddOptionReload(f.name, &f.opt, f.reload)
+	// }
+	// return f, err
```
darthShadow, line 1294: I was planning to add it as a common option, but there is a cyclic import for the rc & fs libs. Thoughts on avoiding it or should I remove this for now?

darthShadow:
@ncw This is for a dynamic reload of the service account via rc. I have 2 problems right now and have added comments inline for both of them. Would you mind taking a look?

ncw:
I think this is a great idea, however I think doing it via the rc is the wrong framework... What if there is more than one drive backend in use?

This needs to in via the backend tools framework (which I have designed but not built yet!) so that we can specify a drive backend instantiation on the command line.

So I'd suggest trying to make it work with the rc as a prototype, then when I've merged the backend tools framework you can plug it into that which will give an `rc` command to access it (and also a command line tool which will be pretty useless in this case!).

darthShadow:
I did think about the multiple drive backends, which is why the reload command is added per fs with the config remote name as the key. However, I have not tested that scenario myself so I assume you are right and I missed something which will stop it from working.
The only case that would break the above scenario is when the same config remote is specified multiple times and I am not sure how to go about handling that.

I agree, however, that this is not ideal and look forward to the proper way of integrating this.

So, how do we go further? Do you want me to fix the quality issue and get it ready for merging till the backend tools framework is brought in or close the PR for now?

darthShadow:
Closing in order to handle this properly via #4180

Issue #4014, opened 2020-02-27
ncdu mode dies and fails when it doesn't have permission to access a remote subdirectory


ruphy:
#### What is the problem you are having with rclone?

when `rclone ncdu` can't access a subpath of a given remote, it fails and quits and provides no way to skip/ignore the directory in question, thus making it completely useless on the whole remote.

For example, suppose I have a shell where I log in as 'me', and where 'user2' has unique acess to her own subdirectory.

The filesystem looks like this:
```
$ ssh me@sftphost  ls -l
total 240
drwx------   10 19463  user2  10 Feb  3 10:08 user2
drwx------   3 19463  me   9 Feb 27 03:30 dir1
drwx------   3 19463  me   9 Feb 27 13:33 dir2
drwx------   3 19463  me   9 Feb 27 04:35 dir3
```

(sftphost is an sftp remote where i connect as user 'me')
```
$ rclone ncdu sftphost:
2020/02/27 18:25:57 Failed to ncdu with 4 errors: last error was: ncdu directory listing: ncdu listing failed: error listing "user2": sftp: "Permission denied" (SSH_FX_PERMISSION_DENIED)
$
```

#### What is your rclone version (output from `rclone version`)

```
rclone v1.49.1
- os/arch: linux/amd64
- go version: go1.12.9
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Ubuntu LTS

####  Which cloud storage system are you using? (eg Google Drive)

SFTP

ncw:
Can you try this with [the latest release](https://rclone.org/downloads/)? I think this might be fixed. It will give errors which will mess up the display but CTRL-L should refresh the display

ruphy:
same error with:

```
$ rclone --version      
rclone v1.51.0
- os/arch: linux/amd64
- go version: go1.13.7
```


I get the usual `sftp: "Permission denied" (SSH_FX_PERMISSION_DENIED)`
(by the way, I really think this is a bug, not an enhancement)

ruphy:
@ncw I took the liberty of modifying the title and the description, let me know if it's better now!

I also just tried to reproduce this error with the local filesystem:
```
mkdir dir1 dir1/a dir1/b dir1/c
sudo chown root dir1/a
sudo chmod 600 dir1/a
```
but in this case I get a working interface, and I receive simply the report of a debug message once I quit `rclone`:
[...]

Issue #3369, opened 2019-07-23
ncdu: show the folder empty and file empty indicator


jtagcat:
rclone v1.48.0
- os/arch: linux/amd64
- go version: go1.12.

ncdu shows empty files as:
@   0.0   B [          ]  emptyfile
..and empty folders as:
e   0.0   B [          ] /emptyfolder

Could rclone ncdu get the indicators as well?

ncw:
This would be relatively easy to add I think - we have the info already.

Is there a list of all the indicators that `ncdu` can produce?  I didn't see them in the man page

I don't see the `@` indicator when I try it with ncdu though?

![image](https://user-images.githubusercontent.com/536803/62130850-cab0bc80-b2d1-11e9-821b-fa7c3caab4aa.png)

Would you like to implement this?

jtagcat:
Unfortunately I don't do go. If I ever get in to go, I can look in to it. Currently I would suggest marking this as good first issue.

ncw:
Ok no probs.  WIll do :-)

yairm:
I think I found the legend in ncdu

ncw:
Ah ha!  We could certainly implement some of those...

albertony:
Here's a ncdu manpage for Ubuntu, showing the same flags with a few more words in the description, as well as one additional flag `^`: http://manpages.ubuntu.com/manpages/impish/man1/ncdu.1.html#file%20flags

jtagcat:
- [x] `!  An error occurred while reading this directory`
 - [x] `.  An error occurred while reading a subdirectory`
 - [ ] `<  File or directory is excluded from the statistics` (rclone: `--exclude`)
    - Excluded files are not listed at all.
 - [x] `e  Empty directory`
    - Currently, special files are treated as empty directories, in both
        - listings (`rclone ncdu /dev/cpu`)
        - root dirs (`rclone ncdu `/dev/full`)
 - [ ] ~~`>  Directory was on another filesystem`~~
 - [ ] ~~`@  This is not a file nor a dir (symlink, socket, ...)`~~
 - [ ] ~~`^  Excluded Linux pseudo-filesystem`~~
 - [ ] ~~`H  Same file was already counted (hard link)`~~
 - [ ] ~~`F Excluded firmlink`~~

I think most of these are not relevant, as they would pretty much only be a thing on local filesystems, where ncdu (and a variety of other clones) may be used. It is just too much out of scope.

jtagcat:
[...]

PR #4168, opened 2020-04-26
Make ncdu more robust in case of errors

rockdreamer:

* Make ncdu command handle errors reading directories.
* Highlight directories with errors in UI.

Yes, issue #4014 

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -281,7 +281,7 @@ func (u *UI) biggestEntry() (biggest int64) {
 		return
 	}
 	for i := range u.entries {
-		size, _, _, _ := u.d.AttrI(u.sortPerm[i])
+		size, _, _, _, _, _ := u.d.AttrI(u.sortPerm[i])
 		if size > biggest {
 			biggest = size
 		}
@@ -326,12 +326,18 @@ func (u *UI) Draw() error {
 			if y >= h-1 {
 				break
 			}
+			size, count, isDir, readable, entriesHaveErrors, err := u.d.AttrI(u.sortPerm[n])
 			fg := termbox.ColorWhite
+			if entriesHaveErrors {
+				fg = termbox.ColorYellow
+			}
+			if err != nil {
+				fg = termbox.ColorRed
+			}
 			bg := termbox.ColorBlack
 			if n == dirPos.entry {
 				fg, bg = bg, fg
 			}
-			size, count, isDir, readable := u.d.AttrI(u.sortPerm[n])
 			mark := ' '
 			if isDir {
 				mark = '/'
@@ -339,6 +345,12 @@ func (u *UI) Draw() error {
 			message := ""
 			if !readable {
 				message = " [not read yet]"
+			}
+			if entriesHaveErrors {
+				message = " [some subdirectories could not be read, size may be underestimated]"
+			}
+			if err != nil {
+				message = fmt.Sprintf(" [%s]", err)
 			}
 			extras := ""
 			if u.showCounts {
@@ -512,8 +524,8 @@ type ncduSort struct {
 // Less is part of sort.Interface.
 func (ds *ncduSort) Less(i, j int) bool {
 	var iAvgSize, jAvgSize float64
-	isize, icount, _, _ := ds.d.AttrI(ds.sortPerm[i])
-	jsize, jcount, _, _ := ds.d.AttrI(ds.sortPerm[j])
+	isize, icount, _, _, _, _ := ds.d.AttrI(ds.sortPerm[i])
+	jsize, jcount, _, _, _, _ := ds.d.AttrI(ds.sortPerm[j])
 	iname, jname := ds.entries[ds.sortPerm[i]].Remote(), ds.entries[ds.sortPerm[j]].Remote()
 	if icount > 0 {
 		iAvgSize = float64(isize / icount)
--- a/cmd/ncdu/scan/scan.go
+++ b/cmd/ncdu/scan/scan.go
@@ -13,13 +13,15 @@ import (
 
 // Dir represents a directory found in the remote
 type Dir struct {
-	parent  *Dir
-	path    string
-	mu      sync.Mutex
-	count   int64
-	size    int64
-	entries fs.DirEntries
-	dirs    map[string]*Dir
+	parent            *Dir
+	path              string
+	mu                sync.Mutex
+	count             int64
+	size              int64
+	entries           fs.DirEntries
+	dirs              map[string]*Dir
+	readError         error
+	entriesHaveErrors bool
 }
 
 // Parent returns the directory above this one
@@ -35,12 +37,13 @@ func (d *Dir) Path() string {
 }
 
 // make a new directory
-func newDir(parent *Dir, dirPath string, entries fs.DirEntries) *Dir {
+func newDir(parent *Dir, dirPath string, entries fs.DirEntries, err error) *Dir {
 	d := &Dir{
-		parent:  parent,
-		path:    dirPath,
-		entries: entries,
-		dirs:    make(map[string]*Dir),
+		parent:    parent,
+		path:      dirPath,
+		entries:   entries,
+		dirs:      make(map[string]*Dir),
+		readError: err,
 	}
 	// Count size in this dir
 	for _, entry := range entries {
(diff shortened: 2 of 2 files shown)


rockdreamer:
Output sample of a directory with unreadable subdirectories
Screenshot_20200426_140654
Output sample of an unreadable directory
Screenshot_20200426_140746

rockdreamer:
I'd like to add tests but the area has none I'm afraid. I'd appreciate a pointer to something I could use to fake a remote with an error.
One thing that I notice is that this now works for remotes, but keeps the previous suboptimal behaviour on local directories (errors fill the screen and are not calculated in the output)

ncw:
This is a very nice change - thank you :-)

> I'd like to add tests but the area has none I'm afraid. I'd appreciate a pointer to something I could use to fake a remote with an error.

Err, yes....

Some parts of `ncdu` should be easy enough to test. Rclone has a framework for running tests against real backends, and also against mock ones.

We don't currently have a backend which could return errors though....

> One thing that I notice is that this now works for remotes, but keeps the previous suboptimal behaviour on local directories (errors fill the screen and are not calculated in the output)

The local backend is a bit noisy about logging stuff it doesn't understand. What exactly are the logs you are seeing?

I thought your code looked good. The only thing that made me think was the extreme number of things returned from `AttrI`! Any ideas here?

rockdreamer:
Hi, thanks for reaching back!
This is the first time I've ever written a possibly useful piece of code in go. I've only read go code and played around with some examples up to now. I'm sure you're trying to get me to think what's the most idiomatic way and reasonable way of moving the data around and my biggest worry is that I'll approach this from a C++/Python perspective that's a poor fit for Go :)

With that said, it feels to me that AttrI is essentially returning most of cmd/ncdu/scan/Dir's attributes.
So I'm thinking that something like renaming Dir to NcduDirEntry (dropping the idea that it's a general 'Dir' class) and making it's contents more visible would make a bit more sense to me, but it also feels smelly to my C++ jarred nose that I'd be exposing such innards around between scan and ncdu.
I'd definitely accept some guidance on how you'd approach this and will work with that in mind :).

Regarding error messages, here's a sample. It's jumbled up exactly like it's on the bottom of the screen
`Total usage: 784.389M, Objects: 163 [listing in progress]2020/04/27 13:43:27 ERROR : ncdu-bad/d3: failed to open directory "ncdu-bad/d3": open /home/rdfm/code/ncdu-bad/d3: permission denied
              95 840            292020/04/27 13:43:27 NOTICE: puddletag/puddletag-docs/changelog: Can't follow symlink without -L/--copy-links
2020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python: Can't follow symlink without -L/--copy-links
               7 099            3442020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python3: Can't follow symlink without -L/--copy-links
2020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python3.8: Can't follow symlink without -L/--copy-links
                 807            4262020/04/27 13:43:27 NOTICE: puddletag/source/README: Can't follow symlink without -L/--copy-links
             4.335G, Objects: 60802020/04/27 13:43:27 NOTICE: chdkptpconversions/chdkptp-git/extlibs/built/lua52/lib/liblua52.a: Can't follow symlink without -L/--copy-links
             6 061            47632   `

What I'm missing is a high level picture of what would happen if those errors were bubbled up just like with other fs backends instead of relying in a workaround. What makes the local fs so special? That's missing from the workaround commit message :(
I'll probably have time to look at existing tests later in the week. Quite a bit of code to go through!

ncw:
> This is the first time I've ever written a possibly useful piece of code in go. I've only read go code and played around with some examples up to now. I'm sure you're trying to get me to think what's the most idiomatic way and reasonable way of moving the data around and my biggest worry is that I'll approach this from a C++/Python perspective that's a poor fit for Go :)

:-)  I think what you've done is great!
 
> With that said, it feels to me that AttrI is essentially returning most of cmd/ncdu/scan/Dir's attributes.

It is, but it is also doing locking which is important

> So I'm thinking that something like renaming Dir to NcduDirEntry (dropping the idea that it's a general 'Dir' class) and making it's contents more visible would make a bit more sense to me, but it also feels smelly to my C++ jarred nose that I'd be exposing such innards around between scan and ncdu.

I'm not big on the idea of renaming it, since it is already quite specialised! Making the innards more open would be nice, but the scan runs concurrently with the displaying (you can use rclone ncdu while it is scanning) so it needs that locking...

> I'd definitely accept some guidance on how you'd approach this and will work with that in mind :).

After thinking about it a bit, how about this

```go
// AttrI returns the size, count and flags for the i-th directory entry
//
// if the entry is a directory isDir will be set and if the sub directory
// has been read it will be returned in subDir.
func (d *Dir) AttrI(i int) (size int64, count int64, isDir bool, subDir *Dir) {
	d.mu.Lock()
	defer d.mu.Unlock()
	subDir, isDir = d.getDir(i)
	if !isDir {
		size = d.entries[i].Size()
	} else if subDir != nil {
		size, count = subDir.Attr()
	}
	return size, count, isDir, subDir
}
```

Then make individual accessor methods (with locking!) for entriesHaveErrors and readError.  Note that readable is just `subDir != nil`

> Regarding error messages, here's a sample. It's jumbled up exactly like it's on the bottom of the screen
> `Total usage: 784.389M, Objects: 163 [listing in progress]2020/04/27 13:43:27 ERROR : ncdu-bad/d3: failed to open directory "ncdu-bad/d3": open /home/rdfm/code/ncdu-bad/d3: permission denied 95 840 292020/04/27 13:43:27 NOTICE: puddletag/puddletag-docs/changelog: Can't follow symlink without -L/--copy-links 2020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python: Can't follow symlink without -L/--copy-links 7 099 3442020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python3: Can't follow symlink without -L/--copy-links 2020/04/27 13:43:27 NOTICE: conan/.tox/full/bin/python3.8: Can't follow symlink without -L/--copy-links 807 4262020/04/27 13:43:27 NOTICE: puddletag/source/README: Can't follow symlink without -L/--copy-links 4.335G, Objects: 60802020/04/27 13:43:27 NOTICE: chdkptpconversions/chdkptp-git/extlibs/built/lua52/lib/liblua52.a: Can't follow symlink without -L/--copy-links 6 061 47632 `
> 
> What I'm missing is a high level picture of what would happen if those errors were bubbled up just like with other fs backends instead of relying in a workaround. What makes the local fs so special? That's missing from the workaround commit message :(

The trouble is that there is no method for rclone to return a directory entry with a note that there was a problem reading it attached. So if let's say there is a permission denied error for reading a single file in a directory, rclone would ditch the entire directory and return an error.

Maybe rclone should change the List() convention so that it returns as many valid entries as it can, and it returns an error if there was an error reading some of it... That is probably quite a big change though...

The treatment in the local backend is a bodge that I'm not really happy with!

For the other backends, reading directories either works or it doesn't - they don't ever work half way...

> I'll probably have time to look at existing tests later in the week. Quite a bit of code to go through!

Given that we don't have a testing framework for ncdu at the moment, building one would be a lot of work... I'd be inclined to get this PR in without tests then we can have a think whether there are parts of ncdu which should be tested and if so what would the best way of doing it be.

Zirro:
As an alternative suggestion for a UI that matches the original ncdu, this could be a great opportunity to implement the first and second flags shown here: https://github.com/rclone/rclone/issues/3369#issuecomment-524086860

ivandeex:
> As an alternative suggestion for a UI that matches the original ncdu, this could be a great opportunity to implement the first and second flags shown here: [#3369 (comment)](https://github.com/rclone/rclone/issues/3369#issuecomment-524086860)

@Zirro feel free to send a PR

ivandeex:
@adamplansky could you please have a peer review of this pr?

ivandeex:
> Given that we don't have a testing framework for ncdu at the moment, building one would be a lot of work...

@rockdreamer 
Could you please post here your manual test procedure as a comment. That way we can all be on the same page. Future ncdu PR would have a place to link, too.

ivandeex:
Rebased on current master.
@rockdreamer Please save undone commits and pull to your workdir.

ivandeex:
> The treatment in the local backend is a bodge that I'm not really happy with!

@ncw
Is it in scope for this PR?
Maybe a new comment in code or new issue would be enough?

ncw:
I've just read through the code one more time, and I think it is sound.

What I'm gong to do is merge it as-is.

I think the ncdu could do with some refactoring - maybe on the lines I suggested above, but that can be a job for another day/PR.

Thank you for a very useful feature :-)

PR #4169, opened 2020-04-26
cmd: added --rmdirs flag to delete command - fixes #4055

### LINKED ISSUE #4055: Remove folders with rclone delete
```
rclone v1.51.0
- os/arch: linux/arm
- go version: go1.13.7
```

Delete all contents instead of just objects with `rclone delete`. 

As [discussed](https://forum.rclone.org/t/how-do-i-delete-all-the-contents-of-a-folder/14837) on the forums, a flag, `--rmdirs` along with `rclone delete` should delete all contents of the folder, but not the folder itself.

kushsharma:
If you supply `--rmdirs` flag with delete command, it will remove all empty directories along with it leaving the root intact.

Implementing remove folders with rclone delete #4055

https://github.com/rclone/rclone/issues/4055

#### Checklist

### DIFF
--- a/cmd/delete/delete.go
+++ b/cmd/delete/delete.go
@@ -4,12 +4,19 @@ import (
 	"context"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	rmdirs = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	flags.BoolVarP(cmdFlags, &rmdirs, "rmdirs", "", rmdirs, "rmdirs removes empty directories but leaves root intact")
 }
 
 var commandDefinition = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/delete/delete.go
```diff
@@ -41,7 +50,12 @@ delete all files bigger than 100MBytes.
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
 		cmd.Run(true, false, command, func() error {
-			return operations.Delete(context.Background(), fsrc)
+			err := operations.Delete(context.Background(), fsrc)
```
ncw, line 53: More idiomatic go here would be to take an early return so

```go
err := operations.Delete(context.Background(), fsrc)
if err != nil {
    return err
}
fdst := cmd.NewFsDir(args)
return operations.Rmdirs(context.Background(), fdst, "", true)
```

kushsharma: Sure, I will do an early return as suggested. Sorry but in your code you have also removed the rmdirs flag check, was that on purpose? How else should we check for that?
```
if err := operations.Delete(context.Background(), fsrc); err != nil {
	return err
}
if rmdirs {
	fdst := cmd.NewFsDir(args)
	return operations.Rmdirs(context.Background(), fdst, "", true)
}
return nil
```

ncw: Yes that looks fine now

PR #4172, opened 2020-04-27
s3: limit upload concurrency

ajankovic:
To allow concurrent FS operations, limiter is moved from function
to the file system object level so it can limit all concurrency for
the fs.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -941,18 +941,19 @@ type Options struct {
 
 // Fs represents a remote s3 server
 type Fs struct {
-	name          string           // the name of the remote
-	root          string           // root of the bucket - ignore all objects above this
-	opt           Options          // parsed options
-	features      *fs.Features     // optional features
-	c             *s3.S3           // the connection to the s3 server
-	ses           *session.Session // the s3 session
-	rootBucket    string           // bucket part of root (if any)
-	rootDirectory string           // directory part of root (if any)
-	cache         *bucket.Cache    // cache for bucket creation status
-	pacer         *fs.Pacer        // To pace the API calls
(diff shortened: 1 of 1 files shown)


ncw:
This is the way the b2 backend works, but uses `fs.Config.Transfers` as the limit.

The s3 backend uploader evolved from the s3manager in the SDK so allows `--transfers` * `--s3-concurrency` uploads at once. Your change will limit that to `--s3-concurrency` transfers....

Why do you want to reduce the concurrency?

ivandeex:
@ajankovic ping

ivandeex:
@ajankovic ?

ajankovic:
@ncw It is needed to limit concurrency for the single backend instance. When limiting concurrency of the operations usually you want more global limit that provides exact maximum that you don't want to go over. For example this is important for the buffer allocation since each new concurrent upload allocates it's own buffer. When you have `transfers * limit` calculation then each new transfer could potentially span new concurrent uploads which hog resources.

ivandeex:
For note: a.t.m. we have pending https://github.com/rclone/rclone/pull/4821 that seeks to adds more concurrency using `--checkers` and https://github.com/rclone/rclone/pull/4172 that limits concurrency stronger using `--s3-concurrency`.

ncw:
On reading the code again, I think that this will change the behaviour of the s3 backend too much.

By default we have

      --s3-upload-concurrency int       Concurrency for multipart uploads. (default 4)

This change makes that the total number of uploads possible, so setting `--transfers` bigger than 4 will have no effect.

I'm going to close this PR, but I would accept one which introduces an new config variable which defaulted to off, which limited the overall concurrency.

PR #4188, opened 2020-04-29
Backend Commands for Drive & Crypt

darthShadow:
drive: Added command to change service_account_file
crypt: Added decode/encode commands to replicate functionality of cryptdecode

Add command support for changing the service-account-file and encoding/decoding crypt file names.

forum.rclone.org/t/edit-service-account-while-mounted/15760

#### Checklist

### DIFF
--- a/backend/crypt/crypt.go
+++ b/backend/crypt/crypt.go
@@ -26,6 +26,7 @@ func init() {
 		Name:        "crypt",
 		Description: "Encrypt/Decrypt a remote",
 		NewFs:       NewFs,
+		CommandHelp: commandHelp,
 		Options: []fs.Option{{
 			Name:     "remote",
 			Help:     "Remote to encrypt/decrypt.\nNormally should contain a ':' and a path, eg \"myremote:path/to/dir\",\n\"myremote:bucket\" or maybe \"myremote:\" (not recommended).",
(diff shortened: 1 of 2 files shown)


darthShadow:
While testing this on a running mount, I discovered that only the first set of changes get applied to the mount. The subsequent changes don't. I am guessing this is probably because of the re-init of the fs with the config which doesn't match the running fs config and hence, doesn't update it.

@ncw Thoughts on how to fix this?

backend/drive/drive.go
```diff
@@ -2582,6 +2587,122 @@ func (f *Fs) Hashes() hash.Set {
...
+	}
+	return
+}
+
+func (f *Fs) changeServiceAccountFile(file string) (err error) {
+	fs.Debugf(nil, "Changing Service Account File from %s to %s", file, f.opt.ServiceAccountFile)
+	if file == f.opt.ServiceAccountFile {
+		return nil
+	}
+	oldCredentials := f.opt.ServiceAccountCredentials
```
ncw, line 2611: I don't like the duplication of the rather trick initialization here...

I think it would be better if you factored that code out of the initialization first then just called that blob of code.

You can do your `undo` by making a copy of fs `fsCopy := *f` then using that to revert f.. (that might not work though because Fs has a mutex in..)

backend/crypt/crypt.go
```diff
@@ -700,6 +701,50 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 	do(ctx, wrappedNotifyFunc, pollIntervalChan)
 }
 
+var commandHelp = []fs.CommandHelp{
+	{
+		Name:  "encode",
+		Short: "Encode the given filename(s)",
```
ncw, line 707: This could probably do with an example to be maximally user friendly as could "decode"

backend/drive/drive.go
```diff
@@ -542,6 +543,8 @@ type Fs struct {
 	exportExtensions []string           // preferred extensions to download docs
 	importMimeTypes  []string           // MIME types to convert to docs
 	isTeamDrive      bool               // true if this is a team drive
+
+	configMapper configmap.Mapper // TODO: Figure a way around this
```
ncw, line 547: My preference would be to call this just `m` like the rest of the code. Terse I know but conventions count!

I don't have a problem with you storing it in the config - seems like a reasonable thing to do (the oauthutil stuff does so that it can write back to the config file).

backend/drive/drive.go
```diff
@@ -2582,6 +2587,122 @@ func (f *Fs) Hashes() hash.Set {
...
+	}
+	chunkSize := fs.SizeSuffix(chunkSizeInt)
+	if chunkSize == f.opt.ChunkSize {
+		return nil
+	}
+	err = checkUploadChunkSize(chunkSize)
+	if err == nil {
+		f.opt.ChunkSize = chunkSize
+	}
+	return
```
ncw, line 2603: `return err` is preferred rather than a bare return

backend/drive/drive.go
darthShadow: Simplified the code a little bit. Let me know if that's fine.

darthShadow:
Adding `m.Set` hasn't solved the issue. On further investigation (with the pointer print statements), this is not deterministic in the sense that only the second change doesn't apply. It occasionally misses applying the first change too.

I am probably missing some lock or mutex which will allow me to modify the existing instance instead of creating a new one randomly.

ncw:
> Adding `m.Set` hasn't solved the issue. On further investigation (with the pointer print statements), this is not deterministic in the sense that only the second change doesn't apply. It occasionally misses applying the first change too.
> 
> I am probably missing some lock or mutex which will allow me to modify the existing instance instead of creating a new one randomly.

Ahhhh.....

We are relying on the [backend Cache](https://github.com/rclone/rclone/blob/master/fs/cache/cache.go) to find the backends

They only stay in the cache for 60 seconds by default so I'm betting if there is more than 60s between attempts you'll get a new backend.

This isn't trivial to fix, we need to somehow pin the backends in the cache if they are open. I'll need to have a think about that.

In the mean time can you verify the hypothesis?

darthShadow:
Yep, you are right. However, it's 5 minutes, not 1 minute, which lines up with the cache expiry duration.

ncw:
I've pushed a fix for this into master - can you rebase and have a go with it? I think it should work!

darthShadow:
This seems to be working fine now.

darthShadow:
Added docs and removed the debug statements. I think this is ready for a review now.

darthShadow:
I was trying to make this work for operations like copy, sync etc. (both via the command-line or via the respective rc operations), where people could change the service account on the fly for both the source and destination. Not sure whether I need to pin the cache in the respective operation or do you want it done on a global scope somewhere?

ncw:
> Added docs and removed the debug statements. I think this is ready for a review now.

That is looking good now - thank you :-)

I've split the commit back into two commits one for crypt and one for drive and I've fixed the conflict. I dropped the change to your email address - I need to figure out a way for it not to immediately be put back in the file when I run the update-authors.py script first!

I'll merge this now.

> I was trying to make this work for operations like copy, sync etc. (both via the command-line or via the respective rc operations), where people could change the service account on the fly for both the source and destination. Not sure whether I need to pin the cache in the respective operation or do you want it done on a global scope somewhere?

Probably the thing to do is to pin it here: https://github.com/rclone/rclone/blob/master/cmd/cmd.go

I'd probably make a `cache.GetAndPin` function and replace all the `cache.Get` commands in that file with `cache.GetAndPin`.

Do you want to send another PR with that?

ncw:
I've fixed your email address here dcf945ed5856359fa9e9f7c53b7c6a63b86b3814

darthShadow:
> Do you want to send another PR with that?

Will do soon.

> I've fixed your email address here dcf945e

Thanks.

Issue #2855, opened 2018-12-21
Improve listings of the "Directory listing" in serve http with more detail


calisro:
Currently the http listings look like this:

```
Directory listing of /
AudioBooks/
Pictures/
Software/
Videos/
apk/
eBooks/
```

I'd like to see the listings more verbose.  Perhaps a flag to increase the verbosity on the listings to look more like other http servers (nginx/apache):

Name  -  Last Modified  -  Size 

Along with sortable columns.  Should be fairly easy to implement in the html delivered to the client post rendering

Similar to this:
https://www.tecmint.com/wp-content/uploads/2017/01/Apache-Directory-Listing.png

If we want to get fancy, we could use a template to have the user create their own html for those pages pretty easily by just exposing those elements to the template {{name}} {{size}} {{last modified}} etc.

ncw:
Funnily enough we are in the process of doing this - this PR is the first step #2844

@poundifdef is looking at this.

calisro:
Closed in favor of that issue and PR.

darwis059:
how to add custom html template?

ncw:
> how to add custom html template?

That isn't possible at the moment, but it might be possible to add something... Do you want to have a go?

c64bob:
I would love to be able to format the output as an Apache Directory Listing. This way, Kodi could parse the output. Any progress using a custom html template, or providing a more verbose one?

ncw:
I'm hoping someone will step up with some code to implement external templates.  @c64bob fancy having a go?

c64bob:
@ncw Sure, I can try, although I'm not that experienced with go. I would add another parameter for the serve command, like`rclone serve http --custom-template /path/to/template.html` What do you think?

ncw:
> @ncw Sure, I can try, although I'm not that experienced with go. I would add another parameter for the serve command, like`rclone serve http --custom-template /path/to/template.html` What do you think?

Something like that sounds perfect!  It might want to be a directory though as you might want some other media in there, eg images.

poundifdef:
I had initially attempted to do exactly this! However, I never had time to complete the PR (new to golang, was planning a wedding, and so on.) But maybe this is a nice starting point?

https://github.com/rclone/rclone/pull/2874

ncw:
@poundifdef If you wanted to warm up that pull request, that would be great!

calisro:
[...]

PR #4200, opened 2020-05-04
added template default and modified tests to rc server #2855

calisro:

This pull will change the default index.html to a more fancy one.  This makes changes to rc-server and modifies the default.

https://github.com/rclone/rclone/issues/2855

#### Checklist


calisro:
Nick,would you take a look at this for me please?  I've made progress on changing the default index to a better one.  I've had to change that also in the rc-server.  Its not done but Im running into an issue with the remote names i'm still trying to figure out.  The colon in the remote names when serving the config root is causing issues.  I can use '%3A' instead of a colon and it works corrently.  I'm not sure if this is a bug yet in the golang templates or my code.  It is odd... 

Do you think what I had so far is going to be okay for changing the rc-server to also use the template?

fs/rc/rcserver/rcserver.go
```diff
@@ -325,8 +326,12 @@ func (s *Server) serveRemote(w http.ResponseWriter, r *http.Request, path string
 		directory := serve.NewDirectory(path, s.HTMLTemplate)
 		for _, entry := range entries {
 			_, isDir := entry.(fs.Directory)
-			directory.AddEntry(entry.Remote(), isDir)
+			directory.AddHTMLEntry(entry.Remote(), isDir, entry.Size(), entry.ModTime(r.Context()))
```
ncw, line 329: I'm concerned about this line of code..

entry.ModTime() can be expensive to calculate (eg in the S3 and Swift backends it will require a round trip to the server).

With the other changes that was based off the VFS which has a `--no-modtime` so the user can control this.

We don't have that option here although I guess the user could use `--use-server-modtime` which would have the same effect.

What do you think?

calisro: For my information, is there a reason VFS isn't in rc-server?  Or is it just that it hasn't been added yet?  

We could add the no-modtime to the rc-server which I could then pass the modtime as 'time.Time{}' which is a time of zero effectively disabling it. I've written a function in data.go and added to teh template that will allow me to hide the dates in the template with a —  dash like the other entries if the time is passed like that.  

Additionally if no-modtime is specified even in the http/webdav it would be good I think to pass the time{} to disable it since that is what the user asked for.  I think if no-modtime is passed now in http it seems to use the current start time of the remote.  What do you think about this?

Thoughts?

calisro: We could also just simply disable modtimes in rc-server since the default will use more transactions.  Easy enough.

calisro:
I've modified the tests now to use a basic template that i've stored in the golden directory per your suggestion.  I think this is pretty close to be being done now.

PR #4219, opened 2020-05-10
jottacloud: update docs regarding cleanup, removed remains from old…

albertony:
…auth, and added warning about special mountpoints.

Documentation only:
* Added information about the new cleanup command from #4119
* Removed remains from old auth, after introduction of new auth in #3711
@buengese: I removed the following section, my understanding is that this is no longer true:
> Device IDs
> Jottacloud requires each 'device' to be registered. Rclone brings such a registration to easily access your account but if you want to use Jottacloud together with rclone on multiple machines you NEED to create a seperate deviceID/deviceSecrect on each machine. You will asked during setting up the remote. Please be aware that this also means that copying the rclone config from one machine to another does NOT work with Jottacloud accounts. You have to create it on each machine.
* Extended the section about devices/mountpoints with a warning about the special mountpoints. (I tested and debugged these a bit, and I see that the Links mountpoint for instance can be used with non-recursing list commands, but limited usage, and probably does not work with much else).
* Some general cleanup

#### Checklist

### DIFF
--- a/docs/content/jottacloud.md
+++ b/docs/content/jottacloud.md
@@ -7,15 +7,25 @@ date: "2018-08-07"
 <i class="fa fa-cloud"></i> Jottacloud
 -----------------------------------------
 
+Jottacoud is a cloud storage service provider from a Norwegian company, using its own datacenters in Norway.
+
+In addition to the official service at [jottacloud.com](https://www.jottacloud.com/), there are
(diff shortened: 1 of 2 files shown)


buengese:
Ah yeah I kinda forgot about updating these with the most recent changes. You are indeed correct here.

Issue #3206, opened 2019-05-18
cache: server side copy not supported with --cache-tmp-upload-path


jimaan:
#### What is the problem you are having with rclone?
rclone does not seem to fallback to Copy&Delete if Move is unavailable for S3/DigitalOcean Storage Spaces. This seems supported based on: https://rclone.org/overview/#move

or maby the detection of Copy support/features fails for DigitalOcean Storage Spaces?
https://github.com/ncw/rclone/blob/cff85f0b9551881be535a4847b56742451e6b476/vfs/file.go#L112
features.Copy for DigitalOcean should exist.

When trying to move an existing file the following error is thrown:
ERROR : /dodata/data/movie.mkv.partial~: Dir.Rename error: Fs "Encrypted drive 'dodata:'" can't rename files (no server side Move or Copy)

But a copy and delete works just fine. ex.:
user@HOSTNAME:/dodata/data# touch a.txt
user@HOSTNAME:/dodata/data# mv a.txt b.txt
mv: cannot move 'a.txt' to 'b.txt': Input/output error
user@HOSTNAME:/dodata/data# cp a.txt b.txt
user@HOSTNAME:/dodata/data# rm -rf a.txt

/var/log/rclone.log contains:
ERROR : data/a.txt: Dir.Rename error: Fs "Encrypted drive 'dodata:'" can't rename files (no server side Move or Copy)

Some relevance to:
https://github.com/ncw/rclone/issues/2109 --> movie.mkv.partial~ is not uploaded yet and still exists only in local cache but will be uploaded if the timer of '--cache-tmp-wait-time' comes by.

https://github.com/ncw/rclone/issues/3184 --> As said by @ncw in this issue: 'Unfortunately B2 is the only remote which doesn't support server side move or server side copy. I've talked to the B2 team about this and they are planning it, but it isn't released yet.' which seems to indicate that S3/DigitalOcean should support the fallback if Move is not supported.

https://github.com/ncw/rclone/issues/2569 --> seems to be the original issue which fixed/supported this behavior. Commit: https://github.com/ncw/rclone/commit/a9fc86f5292005c7ce4b82e7720ea20a73ce8d9f

DigitalOcean seems to support the Bucket Object Copy operation: https://developers.digitalocean.com/documentation/spaces/#copy-object

#### What is your rclone version (output from `rclone version`)
rclone v1.47.0
- os/arch: linux/amd64
- go version: go1.12.4

[...]

PR #4225, opened 2020-05-12
Add Copy to Local Backend

DWSR:

This change implements the Copy feature on the `local` backend in order to fix an issue with the `tmp_upload_path` feature of the `cache` backend.

https://github.com/rclone/rclone/issues/3206

#### Checklist


ncw:
What is the problem in the cache backend that you are trying to fix?

I think there may be an easier way to fix it there?

DWSR:
> What is the problem in the cache backend that you are trying to fix?
> 
> 
> 
> I think there may be an easier way to fix it there?

The issue is the one linked in the PR body. Basically the cache backend needs to support a superset of all of the features supported by the wrapped backend when using `tmp_upload_path` and since that feature is implemented using the local backend and the local backend is missing Copy, server side copies are not supported when using the cache backend in that configuration.

There might be an easier way to do this but it looks like there are some assumptions made about the way that everything operates that requires the Copy feature to be implemented. That being said, I could be missing something. I’m certainly not a Go expert, nor well versed in rclone.

ncw:
I've replied with some ideas on the original issue

DWSR:
Based on comments from @ncw, choosing to solve this in another manner.

PR #4238, opened 2020-05-15
Sftp: Ssh add key pem to config

calisro:

I hope this doesn't fly in the face of best practice. I have some platforms for which I do not allow password authentication at all. Only keys. I'd like to be able to store the key in the config (or environment variable) with the rest of the parameters rather than carrying it around in an external file to deploy along with the rclone config.

So i'd allow for a config parameter to store the actual key rather than only read a file system file. This would also enable tools that abstract the filesystem from the app running rclone to work correctly here (RCX for android for example).

command:
--sftp-key-pem (actual key)

config:
key_pem = (actual key)

I was thinking since the config file doesn't allow multi-line that I would force a sane delimiter (\r\n and \n).

Thoughts?

no.  It was a simple change so figured i'd just do a PR for review.

https://forum.rclone.org/t/add-ssh-key-to-config-file-rather-than-filesystem-pointer/16404

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -908,6 +908,20 @@ changed and won't need copying then you shouldn't use `--no-traverse`.
 
 See [rclone copy](/commands/rclone_copy/) for an example of how to use it.
 
+### --no-unicode-normalization ###
+
+Don't normalize unicode characters in filenames during the sync routine.
+
+Sometimes, an operating system will store filenames containing unicode
+parts in their decomposed form (particularly macOS). Some cloud storage
+systems will then recompose the unicode, resulting in duplicate files if
+the data is ever copied back to a local filesystem.
+
+Using this flag will disable that functionality, treating each unicode
+character as unique. For example, by default é and é will be normalized
+into the same character. With `--no-unicode-normalization` they will be
+treated as unique characters.
+
 ### --no-update-modtime ###
 
 When using this flag, rclone won't update modification times of remote
--- a/fs/config.go
+++ b/fs/config.go
@@ -70,6 +70,7 @@ type ConfigInfo struct {
 	IgnoreCaseSync         bool
 	NoTraverse             bool
 	NoCheckDest            bool
+	NoUnicodeNormalization bool
 	NoUpdateModTime        bool
 	DataRateUnit           string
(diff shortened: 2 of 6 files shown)


calisro:
Perhaps the delimiter should be a ":" or other simple character though ? Since PEMs are base64 it doesn't really matter.   And I probably can remove the other Replace statements.  I'll adjust after feedback.

calisro:
Perhaps the delimiter could be a simple character though like a ":".  And do i don't think I actually need the replace statements besides the actual delimiter.   I can adjust after feedback.

ncw:
I think this is a perfectly reasonable thing to want to do!

I'd suggest a bit of experimentation first

1. I have a feeling that the config file can have multiple line input if the lines are indented with space, eg

```
value = line1
 line2
 line3
```

2. Did you try replacing all the cr/lf with spaces in the identity file - is that acceptable to ssh? It would be nice to be able to say to users, run `sed 's/\n/ /g < keyin > keyout` then paste keyout into your config. Whatever config scheme we come up with coming up with a little sed script will make peoples life easier!

calisro:
I did not experiment with indention for multiline.

It didn't like spaces though.  The only time it actually worked so far was when I replaced the delimiter with \n. It did need the carriage returns from my testing so I planned on just swapping them back in.  I'll give indention a try and also produce an easy command for people to load them if we keep on one line.

calisro:
1) multi-line didn't work with spaces.  Tried quotes and other things.  I took a look at the config source and their documentation (unknwon/goconfig) and nothing came out either.

Failed to load config file "/home/xxxx/.rclone.conf": could not parse line: xxxxxxxxxxxxxx

Assuming multi-line is out for now:
2) spaces in PEM doesn't work either.  Must be LF.

We could make an arbitrary delimiter and just use that.  In the code I replace it with real LF anyway.  

sed 's/\n/:/g <  ~/.ssh/id_rsa 

Thoughts?

calisro:
oops.  What did I just do.  How do I fix this?   I was trying to just remove the go.mod and go.sum files and apparently I removed them all.

ncw:
Good old git! I see you made a new PR :-)

PR #4240, opened 2020-05-16
Sftp: Ssh add key pem to config 2

calisro:
This was closed by mistake:

https://github.com/rclone/rclone/pull/4238

 I'd like to be able to store the key in the config (or environment variable) with the rest of the parameters rather than carrying it around in an external file to deploy along with the rclone config.

So i'd allow for a config parameter to store the actual key rather than only read a file system file. This would also enable tools that abstract the filesystem from the app running rclone to work correctly here (RCX for android for example).

command:
--sftp-key-pem (actual key)

config:
key_pem = (actual key)

I was thinking since the config file doesn't allow multi-line that I would force a sane delimiter (\r\n and \n).

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -70,6 +70,9 @@ func init() {
 			Help:       "SSH password, leave blank to use ssh-agent.",
 			IsPassword: true,
 		}, {
+			Name: "key_pem",
+			Help: "Raw PEM-encoded private key, If specified, will override key_file parameter.",
+		}, {
 			Name: "key_file",
 			Help: "Path to PEM-encoded private key file, leave blank or set key-use-agent to use ssh-agent.",
 		}, {
@@ -172,6 +175,7 @@ type Options struct {
 	User              string `config:"user"`
 	Port              string `config:"port"`
 	Pass              string `config:"pass"`
+	KeyPem            string `config:"key_pem"`
 	KeyFile           string `config:"key_file"`
 	KeyFilePass       string `config:"key_file_pass"`
 	KeyUseAgent       bool   `config:"key_use_agent"`
@@ -390,6 +394,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	}
 
 	keyFile := env.ShellExpand(opt.KeyFile)
+	//keyPem := env.ShellExpand(opt.KeyPem)
 	// Add ssh agent-auth if no password or file specified
 	if (opt.Pass == "" && keyFile == "" && !opt.AskPassword) || opt.KeyUseAgent {
 		sshAgentClient, _, err := sshagent.New()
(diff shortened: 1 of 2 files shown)


backend/sftp/sftp.go
```diff
@@ -427,10 +432,17 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
...
-			return nil, errors.Wrap(err, "failed to read private key file")
+	if keyFile != "" || keyPem != "" {
+		var key []byte
+		if keyPem == "" {
+			key, err = ioutil.ReadFile(keyFile)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to read private key file")
+			}
+		} else {
+			keyPem = strings.Replace(keyPem, `\n `, "\n", -1)
```
ncw, line 443: I note this doesn't handle `\r` but you specifically mention `\r` in the docs...

We could use [strconv.Unquote](https://golang.org/pkg/strconv/#Unquote) instead of this?

Also why does this replace `\n ` with `\n` why the trailing space?

docs/content/sftp.md
```diff
@@ -107,10 +107,20 @@ The SFTP remote supports three authentication methods:
 Key files should be PEM-encoded private key files. For instance `/home/$USER/.ssh/id_rsa`.
 Only unencrypted OpenSSH or PEM encrypted files are supported.
 
-If you don't specify `pass` or `key_file` then rclone will attempt to contact an ssh-agent.
+The key file can be specified in either an external file (key_file) or contained within the 
+rclone config file (key_pem).  If using key_pem in the config file, the entry should be on a
+single line with new line ('\n','\r\n') separating lines.  i.e.
```
ncw, line 112: Your awk doesn't replace `\r`... Is `\r` likely? I guess it might be on Windows

docs/content/sftp.md
```diff
@@ -107,10 +107,20 @@ The SFTP remote supports three authentication methods:
...
+rclone config file (key_pem).  If using key_pem in the config file, the entry should be on a
+single line with new line ('\n','\r\n') separating lines.  i.e. 
+
+key_pem = -----BEGIN RSA PRIVATE KEY-----\nMa...IXtE\n0gA....aSsd\n....\n-----END RSA PRIVATE KEY-----
+
+This will generate it correctly for key_pem for use in the config:  
+
+    awk '{printf "%s\\n", $0}' < ~/.ssh/id_rsa
+
+If you don't specify `pass`, `key_file`, `key_pem` then rclone will attempt to contact an ssh-agent.
```
ncw, line 120: There should probably be an `or` in that sentence after `keyfile`!

docs/content/sftp.md
calisro: Yes.  Thats better.   Thanks!

backend/sftp/sftp.go
calisro: Yes. Thats better. Thanks!

calisro: Since unquote expects actual quotes, i'm adding them manually to the string.  before calling unQuote but I think this is better than trying to replace strings manually.

backend/sftp/sftp.go
```diff
@@ -390,6 +394,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	}
 
 	keyFile := env.ShellExpand(opt.KeyFile)
+	keyPem := env.ShellExpand(opt.KeyPem)
```
ncw, line 397: Oh sorry I noticed this before but forgot to write a note

Do we need to `ShellExpand` the KeyPem? I think it is unlikely to have any environment variables we need expanding in so probably replace `keyPem` with `opt.KeyPem` below?

calisro: ah yeah.  :)  ty

calisro:
done.  Let me know if anything else.   The little things help me learn. :+1:

PR #4242, opened 2020-05-17
Cache - Add Support for Server Side Copy When Using Temp Upload

DWSR:

This change adds support for server side copying to the cache backend (when wrapping a backend that supports it) when the `tmp_upload_path` option is specified. Previously, this configuration would result in the following behaviour from e.g. `mv`:

```text
user@HOSTNAME:/data/data# mv a.txt b.txt
mv: cannot move 'a.txt' to 'b.txt': Input/output error
```

rclone itself would emit a message similar to the following:

```text
ERROR : data/a.txt: Dir.Rename error: Fs "Encrypted drive 'data:'" can't rename files (no server side Move or Copy)
```

This was caused by the `cache` backend "masking out" Copy support from the wrapped backend due to the `local` backend not supporting it.

https://github.com/rclone/rclone/issues/3206

#### Checklist

### DIFF
--- a/backend/cache/cache.go
+++ b/backend/cache/cache.go
@@ -520,9 +520,6 @@ func NewFs(name, rootPath string, m configmap.Mapper) (fs.Fs, error) {
 	// override only those features that use a temp fs and it doesn't support them
 	//f.features.ChangeNotify = f.ChangeNotify
 	if f.opt.TempWritePath != "" {
-		if f.tempFs.Features().Copy == nil {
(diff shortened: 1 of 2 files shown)


DWSR:
I'm not clear on whether or not this should have additional tests written for it. It's a bug that wasn't caught by the existing tests and normally I would say that that's something that warrants adding a test, but I'm fuzzy on exactly how the integration tests are run and it appears that the integration tests currently test this scenario.

PR #4265, opened 2020-05-21
sftp: adds server fingerprint check - fixes #4087

### LINKED ISSUE #4087: sftp: check fingerprint of remote server

rclone v1.51.0
- os/arch: windows/amd64
- go version: go1.13.7

As pointed out [here](https://forum.rclone.org/t/check-sha256-remote-sftp/13378), rclone does not check if the server fingerprint has changed, so it is possibile to easily perform man-in-the-middle attacks on an **sftp** remote.

Just saving the hash to the config file is ok :) as already noted in the thread linked above

FileGo:

The existing code avails itself of a very insecure ssh.InsecureIgnoreHostKey(), which ignores SFTP server fingerprint and could be a security issue with MITM attack or DNS spoofing. I've added a callback function, which checks for existing entries in known_hosts file and compares server key to local entry. If no entry exists, user is prompted to accept server fingerprint, which is then stored. In case of mismatch, user is prompted in a similar manner to openssh client.

#4087 

#### Checklist


FileGo:
@calisro Do you mind looking at why the tests keep failing? I'm probably missing something with go modules...

ncw:
The tests are failing because of this

```
?   	github.com/rclone/rclone/cmd/serve/servetest	[no test files]
2020/05/22 10:07:31 NOTICE: Config file "/notfound" not found - using defaults
2020/05/22 10:07:31 NOTICE: Loaded 0 authorized keys from "/home/runner/.ssh/authorized_keys"
2020/05/22 10:07:31 NOTICE: Generating 2048 bit key pair at "/home/runner/.cache/rclone/serve-sftp/id_rsa"
2020/05/22 10:07:31 NOTICE: SFTP server listening on 127.0.0.1:43233
panic: test timed out after 10m0s

goroutine 24 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.13.11/x64/src/testing/testing.go:1377 +0xdf
created by time.goFunc
	/opt/hostedtoolcache/go/1.13.11/x64/src/time/sleep.go:168 +0x44
```

So it looks like your changes are causing the `cmd/serve/sftp` tests to take too long. Note that these tests use `backend/sftp` so they are likely related to your code.

Try runing the `cmd/serve/sftp` tests locally and see what happens.

Issue #755, opened 2016-10-02
Feature Request: allow to specify DSCP/TOS of network connection


lsching17:
Please allow to specify DSCP/TOS of network connection so that they can be detected more easily for better QOS management

lsching17:
https://godoc.org/golang.org/x/net/ipv4#Conn.SetTOS
https://godoc.org/golang.org/x/net/ipv6#Conn.SetTrafficClass

ncw:
Great idea. It should be reasonably easy to implement in rclone's dial functions (fs/http_new.go and fs/http_old.go).

If you fancy giving it a go I'll write up exactly how?

ncw:
Here is what the SSH manual says

```
IPQoS   Specifies the IPv4 type-of-service or DSCP class for connections.  Accepted values are “af11”, “af12”, “af13”, “af21”, “af22”, “af23”, “af31”, “af32”, “af33”, “af41”,
             “af42”, “af43”, “cs0”, “cs1”, “cs2”, “cs3”, “cs4”, “cs5”, “cs6”, “cs7”, “ef”, “lowdelay”, “throughput”, “reliability”, or a numeric value.  This option may take one or two
             arguments, separated by whitespace.  If one argument is specified, it is used as the packet class unconditionally.  If two values are specified, the first is automatically
             selected for interactive sessions and the second for non-interactive sessions.  The default is “lowdelay” for interactive sessions and “throughput” for non-interactive ses‐
             sions
```

And they are defined thus in the source code

```c
/*
 * Definitions for IP type of service (ip_tos)
 */
#include <netinet/in_systm.h>
#include <netinet/ip.h>
#ifndef IPTOS_LOWDELAY
# define IPTOS_LOWDELAY          0x10
# define IPTOS_THROUGHPUT        0x08
# define IPTOS_RELIABILITY       0x04
# define IPTOS_LOWCOST           0x02
# define IPTOS_MINCOST           IPTOS_LOWCOST
#endif /* IPTOS_LOWDELAY */

/*
 * Definitions for DiffServ Codepoints as per RFC2474
 */
#ifndef IPTOS_DSCP_AF11
# define        IPTOS_DSCP_AF11         0x28
# define        IPTOS_DSCP_AF12         0x30
# define        IPTOS_DSCP_AF13         0x38
# define        IPTOS_DSCP_AF21         0x48
# define        IPTOS_DSCP_AF22         0x50
# define        IPTOS_DSCP_AF23         0x58
[...]

PR #4266, opened 2020-05-21
Add DSCP support

Max-Sum:

Allow specifying DSCP value for connections. Avoid dominating the bandwidth when doing long time operation.

#755

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -600,6 +600,21 @@ This flag can be useful for debugging and in exceptional circumstances
 (e.g. Google Drive limiting the total volume of Server Side Copies to
 100GB/day).
 
+### --dscp VALUE ###
+
+Specify a DSCP value or name to use in connections. This could help QoS
+system to identify traffic class. BE, EF, DF, LE, CSx and AFxx are allowed.
+
+See the description of [differentiated services](https://en.wikipedia.org/wiki/Differentiated_services) to get an idea of
+this field. Setting this to 1 (LE) to identify the flow to SCAVENGER class
+can avoid occupying too much bandwidth in a network with DiffServ support ([RFC 8622](https://tools.ietf.org/html/rfc8622)).
+
+For example, if you configured QoS on router to handle LE properly. Running:
+```
+rclone copy --dscp LE from:/from to:/to
+```
+would make the priority lower than usual internet flows.
+
 ### -n, --dry-run ###
 
 Do a trial run with no permanent changes.  Use this to see what rclone
--- a/fs/fshttp/dialer.go
+++ b/fs/fshttp/dialer.go
@@ -1,0 +1,114 @@
+package fshttp
+
+import (
+	"context"
+	"net"
+	"time"
+
+	"github.com/rclone/rclone/fs"
+	"golang.org/x/net/ipv4"
+	"golang.org/x/net/ipv6"
+)
+
+func dialContext(ctx context.Context, network, address string, ci *fs.ConfigInfo) (net.Conn, error) {
+	return NewDialer(ctx).DialContext(ctx, network, address)
+}
+
+// Dialer structure contains default dialer and timeout, tclass support
+type Dialer struct {
+	net.Dialer
+	timeout time.Duration
+	tclass  int
+}
+
+// NewDialer creates a Dialer structure with Timeout, Keepalive,
+// LocalAddr and DSCP set from rclone flags.
+func NewDialer(ctx context.Context) *Dialer {
+	ci := fs.GetConfig(ctx)
+	dialer := &Dialer{
+		Dialer: net.Dialer{
+			Timeout:   ci.ConnectTimeout,
+			KeepAlive: 30 * time.Second,
+		},
+		timeout: ci.Timeout,
+		tclass:  int(ci.TrafficClass),
+	}
+	if ci.BindAddr != nil {
+		dialer.Dialer.LocalAddr = &net.TCPAddr{IP: ci.BindAddr}
+	}
+	return dialer
+}
+
+// Dial connects to the address on the named network.
+func (d *Dialer) Dial(network, address string) (net.Conn, error) {
+	return d.DialContext(context.Background(), network, address)
+}
+
+// DialContext connects to the address on the named network using
+// the provided context.
+func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	c, err := d.Dialer.DialContext(ctx, network, address)
+	if err != nil {
+		return c, err
+	}
+	if d.tclass != 0 {
+		if addr, ok := c.RemoteAddr().(*net.IPAddr); ok {
+			if addr.IP.To16() != nil && addr.IP.To4() == nil {
+				err = ipv6.NewConn(c).SetTrafficClass(d.tclass)
+			} else {
+				err = ipv4.NewConn(c).SetTOS(d.tclass)
+			}
+			if err != nil {
+				return c, err
+			}
+		}
+	}
+	return newTimeoutConn(c, d.timeout)
+}
+
+// A net.Conn that sets a deadline for every Read or Write operation
+type timeoutConn struct {
+	net.Conn
+	timeout time.Duration
+}
+
+// create a timeoutConn using the timeout
+func newTimeoutConn(conn net.Conn, timeout time.Duration) (c *timeoutConn, err error) {
+	c = &timeoutConn{
+		Conn:    conn,
+		timeout: timeout,
+	}
+	err = c.nudgeDeadline()
+	return
+}
+
+// Nudge the deadline for an idle timeout on by c.timeout if non-zero
+func (c *timeoutConn) nudgeDeadline() (err error) {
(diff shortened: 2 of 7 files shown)


calisro:
I think you mean this.  

https://github.com/rclone/rclone/issues/755

Max-Sum:
> 
> 
> I think you mean this.
> 
> #755

My bad. Thank you.

ncw:
This needs another commit with the new dependencies in: https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#adding-a-dependency

Can you add the commit then rebase it so it comes before your existing commit?

Max-Sum:
Regards to the suggestions from @jgottula, will add RFC 4594 style named values (CS0, AF11, EF, etc).

fs/fshttp/dialer.go
```diff
@@ -0,0 +1,112 @@
...
+
+// DialContext connects to the address on the named network using
+// the provided context.
+func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
+	c, err := d.Dialer.DialContext(ctx, network, address)
+	if err != nil {
+		return c, err
+	}
+	if d.tclass != 0 {
+		ip := c.RemoteAddr().(*net.TCPAddr).IP
```
ncw, line 54: This should use a checked type assertion here

```
    tcp, ok: := c.RemoteAddr().(*net.TCPAddr)
    if ok {
```

As it is quite possible that the connection wasn't TCP

docs/content/docs.md
```diff
@@ -552,6 +552,15 @@ This flag can be useful for debugging and in exceptional circumstances
 (eg Google Drive limiting the total volume of Server Side Copies to
 100GB/day).
 
+### --dscp VALUE ###
+
+Specify a DSCP value or name to use in connections. This could help QoS
+system to identify traffic class. EF, DF, CSx and AFxx are allowed.
+
```
ncw, line 559: This should mention that this is only effective for HTTP based services as your patch doesn't affect the ftp/sftp backends.

Max-Sum: Should I also implement for ftp backends?

ncw: Nice one :-)

You appear to have a conflict though so you'll have to rebase on master.

Do you want to do backend/sftp too?

Max-Sum: I thought sftp already uses fshttp.NewDialer?

ncw: So it does! I'd forgotten that :-)

ncw:
I rebased this code (and fixed the conflicts), lets see if it passes the tests.

It needs some more docs.

What do you think @ivandeex - ready to merge?

ivandeex:
Probably 🤔 . I'm not yet aware what DSCP is. I just triaged the list of PRs and marked those waiting for review more than 4 months with "important" label or by "request review", and those where user did not reply since long by the respective label, those with old changes and no failed tests with "needs rebase". My intention was to estimate the technical debt. Sorry if I caused extra effort. 😬

Max-Sum:
Anything I can help?

docs/content/docs.md
```diff
@@ -571,6 +571,15 @@ This flag can be useful for debugging and in exceptional circumstances
...
+See the description of [differentiated services](https://en.wikipedia.org/wiki/Differentiated_services) to get an idea of
+this field. Setting this to 8 (CS1) to identify the flow to SCAVENGER class
+can avoid occupying too much bandwidth in a network with DiffServ support.
```
ivandeex, line 581: I agree with @ncw asking for "some more docs". @Max-Sum It would be nice if you added a short practical example, like
```
You have service1 with QoS xxx, then you run e.g.
rclone copy --dscp yyy from:/from to:/to
and in result you get...
```

ncw: I guess most rclone users copy bulk data about but want to lower the QoS on that so it doesn't interfere with their video conferencing/VOIP calls/Netflix viewing. What would be the best setting for that? Is that `CS1`?

I remember that @calebcase and I had a conversation about this because tardigrade uses this too - any thoughts?

calebcase: This is how we did it in storj/uplink: https://github.com/storj/common/tree/73bea234a20e88585742cfa6a794f10e7edfcaac/socket

We enabled it by default to help where we could with congestion issues.

Specifically: https://github.com/storj/common/blob/73bea234a20e88585742cfa6a794f10e7edfcaac/socket/background_linux.go#L19
Where `byte(dscpLE)<<2` (https://github.com/storj/common/blob/73bea234a20e88585742cfa6a794f10e7edfcaac/socket/dscp.go#L36) is `0b000001 << 2`
This was chosen based on the recommendations from https://tools.ietf.org/html/rfc8622#section-6 (which  recommends against `CS1`).

Max-Sum: How about I recommend people to set CS1 but not making it default? Because that might not be in people's favor.

ncw: Thanks @calebcase - that RFC recommendation seems solid.

What do you think about making the recommended setting `LE` @Max-Sum ? Or perhaps we should say that the modern recommendation is `LE` but for older systems `CS1` might be appropriate.

ivandeex:
@Max-Sum friendly ping

dtaht:
If it helps any, I am rather interested in this work, and in also being able to set/unset the ecn bits. I do not know enough about go to get anywhere....

docs/content/docs.md
Max-Sum: According to the RFC, CS1 might just not be used for this kind of flows. I think making LE the only recommanded setting would be appropriate.

Max-Sum:
> 
> 
> If it helps any, I am rather interested in this work, and in also being able to set/unset the ecn bits. I do not know enough about go to get anywhere....

ECN should be set in OS.
Linux: sysctl net.ipv4.tcp_ecn=1
Windows: netsh interface tcp set global ecncapability=enabled

dtaht:
No, the ecn bits are also set or reset via an aqm along the way. an app marks a flow ecn capable, an aqm sets CE (or in the case of L4S or SCE, moves the bit sideways. Or they get bleached, much like dscp does.

Max-Sum:
> 
> 
> No, the ecn bits are also set or reset via an aqm along the way. an app marks a flow ecn capable, an aqm sets CE (or in the case of L4S or SCE, moves the bit sideways. Or they get bleached, much like dscp does.

I believe that all these works are handled by AQM or OS. Applications don't need to explicitly set a bit or flag to enable ECN support.

ivandeex:
> No, the ecn bits are also set or reset via an aqm along the way. an app marks a flow ecn capable.....

@dtaht 
Can you point to a C code example?

docs/content/docs.md
ivandeex: So this thread is resolved.

docs/content/docs.md
ivandeex: So `LE` stays as the only recommended setting.

@Max-Sum @ncw @calebcase 
I think we can consider this thread resolved.

docs/content/docs.md
```diff
@@ -578,6 +578,21 @@ This flag can be useful for debugging and in exceptional circumstances
...
 
+### --dscp VALUE ###
+
+Specify a DSCP value or name to use in connections. This could help QoS
+system to identify traffic class. BE, EF, DF, LE, CSx and AFxx are allowed.
+
+See the description of [differentiated services](https://en.wikipedia.org/wiki/Differentiated_services) to get an idea of
+this field. Setting this to 1 (LE) to identify the flow to SCAVENGER class
+can avoid occupying too much bandwidth in a network with DiffServ support ([RFC 8622](https://tools.ietf.org/html/rfc8622)).
+
```
ivandeex, line 589: @Max-Sum 

Please add https://github.com/rclone/rclone/pull/4266#issuecomment-769554945
to the docs:

```
ECN should be set in OS.
Linux: sysctl net.ipv4.tcp_ecn=1
Windows: netsh interface tcp set global ecncapability=enabled
```

ivandeex:
@Max-Sum 
Please address my comment regarding the docs.
and request a review when you are done with this round of changes.

docs/content/docs.md
Max-Sum: ECN is a part of DSCP. I don't think this should be appended.

ivandeex: OK

Max-Sum:
@ncw This feature is ready for review.

ivandeex:
@Max-Sum rebased on master. please pull results from github, don't force-push now.

ivandeex:
@nwc PR is ready for your review.

ivandeex:
@ncw You can look at the resolved threads for discussion.

ncw (approved):
This change looks very good now - thank you @Max-Sum 

I will merge it now, nice and early in the dev cycle and we can see if anything breaks!

Thanks!

ncw:
Thank you @Max-Sum for implementing the feature and to @ivandeex for shepherding it in :-)

PR #4272, opened 2020-05-24
Rc/mount

negative0:
1. Add function to list the current mounts.
1. Add more information for current mount 

#### Checklist


cmd/mountlib/rc.go
```diff
@@ -4,27 +4,39 @@ import (
...
 	mountFns map[string]MountFn
-	// Map of mounted path => unmount function
-	unmountFns map[string]UnmountFn
+	// Map of mounted path => MountInfo
+	liveMounts map[string]MountInfo
 )
 
+func init() {
+
+	mountFns = make(map[string]MountFn)
```
ncw, line 31: If you want to initialise them to an empty `map` then change the var definition to

```go
    // Mount functions available
￼    mountFns = map[string]MountFn{}
    // Map of mounted path => MountInfo
￼    liveMounts = map[string]MountInfo{}
```

negative0: The mutex should be used only for modifying the map right?

cmd/mountlib/rc.go
```diff
@@ -80,7 +91,16 @@ func mountRc(_ context.Context, in rc.Params) (out rc.Params, err error) {
 	}
 
 	if mountFns[mountType] != nil {
```
ncw, line 93: the mount fns mutex needs to be locked here

cmd/mountlib/rc.go
```diff
@@ -4,29 +4,40 @@ import (
...
+	MountedOn  time.Time `json:"MountedOn"`
+	Fs         string    `json:"Fs"`
+}
+
 var (
 	// Mount functions available
-	mountFns map[string]MountFn
-	// Map of mounted path => unmount function
-	unmountFns map[string]UnmountFn
+	mountFns = map[string]MountFn{}
```
ncw, line 25: I would just have 1 mutex to protect both maps otherwise you'll get confused about which one you've got locked!

It is traditional in go code to call them `mu` or `somethingMu` so lets say it is called `mountsMu`.

It is also traditional to define them **before** the things they protect, so I think this block should look like

```go
    // mutex to protect all the variables in this block
    mountMu sync.Mutex
    // Mount functions available
    mountFns = map[string]MountFn{}
    // Map of mounted path => MountInfo
    liveMounts = map[string]MountInfo{}
```

cmd/mountlib/rc.go
```diff
@@ -54,7 +65,7 @@ Eg
 	})
 }
 
-// rcMount allows the mount command to be run from rc
+// mountRc allows the mount command to be run from rc
 func mountRc(_ context.Context, in rc.Params) (out rc.Params, err error) {
 	mountPoint, err := in.GetString("mountPoint")
```
ncw, line 70: for more complicated functions it is easiest just to write this for locking purposes

```go
    mountMu.Lock()
    defer mountMu.Unlock()
```

That will mean that the mutex is unlocked on function exit regardless. It is easy to audit and in go1.14 and later there is no performance penalty for using defer like this.

It is really super easy to mess locking up (I speak from lots of experience messing it up here!) so keeping it simple is best. I'd probably use this form on all your functions then you and I don't need to worry about it!

negative0: Ahh.. This looks easier :p

negative0: Shouldn't the mutex be released as soon as the reading, modification is finished?

negative0: Also, shouldn't read/read operation be allowed? As they are nonconflicting.

negative0: I'm curious

negative0:
Thank You :)

PR #4276, opened 2020-05-25
docs: fix markup of arguments

tbm:
This is probably incomplete but a good start.

I don't have a Go environment so this is not tested.

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -159,34 +159,36 @@ func NewMountCommand(commandName string, hidden bool, mount MountFn) *cobra.Comm
 		Use:    commandName + " remote:path /path/to/mountpoint",
 		Hidden: hidden,
 		Short:  `Mount the remote as file system on a mountpoint.`,
-		Long: `
-rclone ` + commandName + ` allows Linux, FreeBSD, macOS and Windows to
+		// Warning! "|" will be replaced by backticks below
+		//          "@" will be replaced by the command name
+		Long: strings.ReplaceAll(strings.ReplaceAll(`
+rclone @ allows Linux, FreeBSD, macOS and Windows to
 mount any of Rclone's cloud storage systems as a file system with
 FUSE.
 
-First set up your remote using ` + "`rclone config`" + `.  Check it works with ` + "`rclone ls`" + ` etc.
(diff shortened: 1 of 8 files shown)


tbm:
> I don't have a Go environment so this is not tested.

CI will complain if I introduced syntax errors. But I'm worried I deleted some whitesace and words will stick together in the output.  I *think* I've avoided such mistakes though.

cmd/mountlib/mount.go
```diff
@@ -199,7 +199,7 @@ normal drives to be fast and reliable, while cloud storage is far from
 that.  See also [Limitations](#limitations) section below for more
 info
 
-Add "--fuse-flag --VolumePrefix=\server\share" to your "mount"
+Add ` + "`--fuse-flag --VolumePrefix=\\server\\share`" + ` to your ` + "`mount`" + `
```
ncw, line 202: I don't think these \ need to be doubled up? I can see them displaying correctly here: https://tip.rclone.org/commands/rclone_mount/ (this is the docs website made from master).

tbm: Without it, Go was complaining about unknown escape sequences.

I don't know Go, but my understanding is that \`...\` is a literal string, so you can put anything in it without escaping.  But "..." is a normal string, so "\s" is seen as the escape sequence for "s", which doesn't exist.

ivandeex: resolved in the rebased tree

ivandeex:
@tbm
Your branch was rebased on current rclone master.
Please pull the result from github, **do not** force-push now.
Your original work had been saved at https://github.com/ivandeex/rclone/tree/tbm-markup-saved-2021-01-22

Also, in the second patch I propose a more readable solution to the problem.
Please note that `strings.ReplaceAll` was introduced in go 1.12 which is the minimum required version since https://github.com/rclone/rclone/commit/01adee75544dca750d929fdbf9cde0c9b70b5c4d#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721

@ncw
Please review :pray: :cat:

ncw (approved):
That looks good to me :-)

Please merge @ivandeex

PR #4280, opened 2020-05-26
s3: implement link sharing

Shareed2k:
 Add link sharing to S3
 
 #2178
 
 #### Checklist
 * [x]  I have read the [contribution guidelines](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
 * [ ]  I have added tests for all changes in this PR if appropriate.
 * [x]  I have added documentation for the changes if appropriate.
 * [x]  All commit messages are in [house style](https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#commit-messages).
 * [x]  I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1959,6 +1959,20 @@ func (f *Fs) getMemoryPool(size int64) *pool.Pool {
 	)
 }
 
+// PublicLink generates a public link to the remote path (usually readable by anyone)
+func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, unlink bool) (link string, err error) {
+	if _, err := f.NewObject(ctx, remote); err != nil {
+		return "", err
+	}
+	bucket, bucketPath := f.split(remote)
+	httpReq, _ := f.c.GetObjectRequest(&s3.GetObjectInput{
+		Bucket: &bucket,
+		Key:    &bucketPath,
+	})
+
+	return httpReq.Presign(time.Duration(expire))
+}
+
 // ------------------------------------------------------------
 
 // Fs returns the parent Fs
--- a/cmd/link/link.go
+++ b/cmd/link/link.go
@@ -3,35 +3,57 @@ package link
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	expire = fs.Duration(time.Hour * 24 * 365 * 100)
+	unlink = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	flags.FVarP(cmdFlags, &expire, "expire", "", "The amount of time that the link will be valid")
+	flags.BoolVarP(cmdFlags, &unlink, "unlink", "", unlink, "Remove existing public link to file/folder")
 }
 
 var commandDefinition = &cobra.Command{
 	Use:   "link remote:path",
 	Short: `Generate public link to file/folder.`,
-	Long: `
-rclone link will create or retrieve a public link to the given file or folder.
+	Long: `rclone link will create, retrieve or remove a public link to the given
+file or folder.
 
     rclone link remote:path/to/file
     rclone link remote:path/to/folder/
+    rclone link --unlink remote:path/to/folder/
+    rclone link --expire 1d remote:path/to/file
 
-If successful, the last line of the output will contain the link. Exact
-capabilities depend on the remote, but the link will always be created with
-the least constraints – e.g. no expiry, no password protection, accessible
-without account.
+If you supply the --expire flag, it will set the expiration time
+otherwise it will use the default (100 years). **Note** not all
+backends support the --expire flag - if the backend doesn't support it
+then the link returned won't expire.
+
+Use the --unlink flag to remove existing public links to the file or
+folder. **Note** not all backends support "--unlink" flag - those that
+don't will just ignore it.
+
+If successful, the last line of the output will contain the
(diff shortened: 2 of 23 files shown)


ncw:
I think this code looks fine! I will give a detailed review when I've finished getting the 1.52 release out so apologies for the delay...

backend/s3/s3.go
```diff
@@ -1959,6 +1967,24 @@ func (f *Fs) getMemoryPool(size int64) *pool.Pool {
 	)
 }
 
+// PublicLink generates a public link to the remote path (usually readable by anyone)
+func (f *Fs) PublicLink(ctx context.Context, remote string) (link string, err error) {
+	if ok, err := fs.FileExists(ctx, f, remote); !ok {
```
ncw, line 1972: The way to do this in a backend is to call `f.NewObject(ctx, remote)` and if you get any errors just return them.

backend/s3/s3.go
```diff
@@ -907,6 +912,8 @@ const (
 	maxUploadCutoff     = fs.SizeSuffix(5 * 1024 * 1024 * 1024)
 	minSleep            = 10 * time.Millisecond // In case of error, start at 10ms sleep.
 
+	defaultURLExpireTime = fs.Duration(15 * time.Minute)
```
ncw, line 915: From the `rclone link` docs

```
$ rclone link --help

rclone link will create or retrieve a public link to the given file or folder.

    rclone link remote:path/to/file
    rclone link remote:path/to/folder/

If successful, the last line of the output will contain the link. Exact
capabilities depend on the remote, but the link will always be created with
the least constraints – e.g. no expiry, no password protection, accessible
without account.
```

So this should probably be longer by default!

We should probably add an expiry parameter to the link command as quite a lot of the remotes support it... The other thing that some remotes support is deleting a public link.

Shareed2k: ok cool so I will add additional param

cmd/link/link.go
```diff
@@ -3,20 +3,32 @@ package link
...
+	"time"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	expire = time.Minute * 15
```
ncw, line 15: This should default to very large to be compatible with previous rclone versions - say 100 years!

Also using a type of `fs.Duration` here will give the user more options when configuring it including days, weeks absolute times etc! You'll need to cast it back to time.Duration for the call.

    expire = fs.Duration(time.Hour * 24 * 365 * 100)

cmd/link/link.go
```diff
@@ -3,20 +3,32 @@ package link
...
+	cmdFlags := commandDefinition.Flags()
+	flags.DurationVarP(cmdFlags, &expire, "expire", "", expire, "the amount of time that the URL is valid")
+	flags.BoolVarP(cmdFlags, &unlink, "unlink", "", unlink, "Remove existing public link to file/folder with link command rather than creating")
 }
 
 var commandDefinition = &cobra.Command{
 	Use:   "link remote:path",
 	Short: `Generate public link to file/folder.`,
 	Long: `
+If you supply the --expire flag, it will set expiration time else it will use default 15 minutes.
```
ncw, line 30: This needs a "Note: not all backends support this flag - if the backend doesn't support it then the link returned won't expire."

cmd/link/link.go
```diff
@@ -26,12 +38,15 @@ If successful, the last line of the output will contain the link. Exact
 capabilities depend on the remote, but the link will always be created with
 the least constraints – e.g. no expiry, no password protection, accessible
 without account.
+
+Use the --unlink flag to remove existing public link to file/folder with link command rather than creating,
+Default is false, meaning link command will create or retrieve public link
```
ncw, line 43: This needs a similar note: Not all backends support `--unlink` - those that don't will just ignore it.

fs/rc/cache.go
```diff
@@ -41,3 +44,37 @@ func GetFsAndRemoteNamed(in Params, fsName, remoteName string) (f fs.Fs, remote
 func GetFsAndRemote(in Params) (f fs.Fs, remote string, err error) {
 	return GetFsAndRemoteNamed(in, "fs", "remote")
 }
+
+// GetExpireAndUnlink get the `expire` and `unlink` parameters from in
+func GetExpireAndUnlink(in Params) (expire time.Duration, unlink bool, err error) {
+	unlink, err = in.GetBool("unlink")
```
ncw, line 50: I'd just call this in `rcPublicLink`

```go
unlink, _ = in.GetBool("unlink")
```

You don't want to give an error if the user doesn't pass the parameter you want it to default to `false`.

Likewise the `expire` parameter - missing should mean very large.

fs/rc/cache.go
```diff
@@ -41,3 +44,37 @@ func GetFsAndRemoteNamed(in Params, fsName, remoteName string) (f fs.Fs, remote
 func GetFsAndRemote(in Params) (f fs.Fs, remote string, err error) {
 	return GetFsAndRemoteNamed(in, "fs", "remote")
 }
+
+// GetExpireAndUnlink get the `expire` and `unlink` parameters from in
+func GetExpireAndUnlink(in Params) (expire time.Duration, unlink bool, err error) {
```
ncw, line 49: Can you move this to `fs/rc/params.go` and call it `GetDuration` and change it so it only reads a duration. This could do with a test too in `fs/rc/params_test.go`

Zirro:
> ...if the backend doesn't support it then the link returned won't expire.

Ignoring these options silently seems a bit risky. Wouldn't it be more appropriate to error out for backends without support (in addition to if a supporting backend responds with an error)?

ncw:
> Ignoring these options silently seems a bit risky. Wouldn't it be more appropriate to error out for backends without support (in addition to if a supporting backend responds with an error)?

I agree, but it is a lot more work to do that. It would require error checking in all the backends which implement `PublicLink`. Might be a good project if you fancied having a go @Zirro ?

PR #4288, opened 2020-05-28
ftp: add explicit tls support

### LINKED ISSUE #4100: Add support for explicit FTP over TLS
The FTP backend currently does not support explicit FTP over TLS for accessing existing servers.

Update:
Unfortunately, the used FTP client library [does not support](https://github.com/jlaffaye/ftp/issues/162) this feature at the moment.

heikobornholdt:
Add support for explicit FTP over TLS.

Fixes #4100

Add support for explicit FTP over TLS

#4100

#### Checklist

### DIFF
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -50,8 +50,19 @@ func init() {
 			IsPassword: true,
 			Required:   true,
 		}, {
-			Name:    "tls",
-			Help:    "Use FTP over TLS (Implicit)",
+			Name: "tls",
+			Help: `Use FTPS over TLS (Implicit)
+When using implicit FTP over TLS the client will connect using TLS
+right from the start, which in turn breaks the compatibility with
+non-TLS-aware servers. This is usually served over port 990 rather
+than port 21. Cannot be used in combination with explicit FTP.`,
+			Default: false,
+		}, {
+			Name: "explicit_tls",
+			Help: `Use FTP over TLS (Explicit)
+When using explicit FTP over TLS the client explicitly request
+security from the server in order to upgrade a plain text connection
+to an encrypted one. Cannot be used in combination with implicit FTP.`,
 			Default: false,
 		}, {
 			Name:     "concurrency",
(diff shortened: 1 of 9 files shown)


heikobornholdt:
I have not worked with go before.
The error looks to me like an old version of the ftp library is used?
I do not have this error on my local computer.

ncw:
I think to fix the build you need to update the ftp library dependency. So if you could make a separate vendor commit and move it before this change that would be perfect.

https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#updating-a-dependency

heikobornholdt:
Thanks. I have now properly updated the dependency.

backend/ftp/ftp.go
```diff
@@ -53,6 +53,10 @@ func init() {
 			Name:    "tls",
 			Help:    "Use FTP over TLS (Implicit)",
 			Default: false,
+		}, {
+			Name:    "explicit_tls",
+			Help:    "Use FTP over TLS (Explicit)",
```
ncw, line 58: Can you expand the help for `explicit_tls` and `tls` now we have both I think we need to write a bit more here.

Note the Help can stretch over multiple lines - the first line is used for the command line help the next is the long help which will go into the docs and is shown to the user at config time.

Eg something like this for explicit TLS.


```go
    Name:    "explicit_tls",
    Help:    `Use FTP over TLS (Explicit)

When using explicit FTP over TLS the client will connect using TLS
right from the start. This is usually served over port 990 rather than
port 21.`
```

backend/ftp/ftp.go
```diff
@@ -158,6 +163,12 @@ func (f *Fs) ftpConnection() (*ftp.ServerConn, error) {
 			InsecureSkipVerify: f.opt.SkipVerifyTLSCert,
 		}
 		ftpConfig = append(ftpConfig, ftp.DialWithTLS(tlsConfig))
+	} else if f.opt.ExplicitTLS {
```
ncw, line 166: Are explicit TLS and TLS mutually incompatible? If so we should flag that to the user by returning an error here.

heikobornholdt: I could imagine that it could work technically. But I could not find any information about it on the web, and I think there is no practical use case for it. An error is now returned when both methods are used together. A note has been added to the documentation.

backend/ftp/ftp.go
heikobornholdt: Thanks for the tip. I have extended the documentation of both parameters.

Issue #3930, opened 2020-02-03
Implement getDirectDownloadLink for Google Drive


Dibbyo456:
Drive API **v2** returns `downloadUrl` (not v3) which can be used to have direct download link.

If we do

```
curl \
  'https://www.googleapis.com/drive/v2/files/{drive_file_id}?supportsAllDrives=true&fields=downloadUrl' \
  --header 'Authorization: Bearer {access_token}' \
  --header 'Accept: application/json' \
  --compressed
```

It returns

```
{
 "downloadUrl": "https://doc-04-8g-docs.googleusercontent.com/docs/securesc/{redacted}?h={redacted}&e=download&gd=true"
}
```
But it won't work directly, to make it work we need to pass the access token to the url like this

```
https://doc-04-8g-docs.googleusercontent.com/docs/securesc/{redacted}?h={redacted}&e=download&gd=true&access_token={access_token}
```

A command something like `directlink' or `directdownloadlink'

So, if we do

```
rclone directdownloadlink remote:path/to/file
```
it would return the url.

The main benefit of this approach over the current `rclone link` is this one is direct link and can be used with any tools such curl, aria2 or idm.

PR #4294, opened 2020-05-31
feat: add onedrive make direct link feature

septs:

Export direct download link

No

https://github.com/rclone/rclone/issues/3930

#### Checklist

### DIFF
--- a/backend/onedrive/api/types.go
+++ b/backend/onedrive/api/types.go
@@ -160,22 +160,23 @@ type PackageFacet struct {
 
 // Item represents metadata for an item in OneDrive
 type Item struct {
-	ID                   string               `json:"id"`                   // The unique identifier of the item within the Drive. Read-only.
-	Name                 string               `json:"name"`                 // The name of the item (filename and extension). Read-write.
-	ETag                 string               `json:"eTag"`                 // eTag for the entire item (metadata + content). Read-only.
-	CTag                 string               `json:"cTag"`                 // An eTag for the content of the item. This eTag is not changed if only the metadata is changed. Read-only.
-	CreatedBy            IdentitySet          `json:"createdBy"`            // Identity of the user, device, and application which created the item. Read-only.
-	LastModifiedBy       IdentitySet          `json:"lastModifiedBy"`       // Identity of the user, device, and application which last modified the item. Read-only.
-	CreatedDateTime      Timestamp            `json:"createdDateTime"`      // Date and time of item creation. Read-only.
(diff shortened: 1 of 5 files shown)


septs:
@ncw

septs:
(Next a Pull request, implementation #3930)

Cnly:
First of all thanks for your work :) I have two comments on this PR:

First is that flags beginning with `--drive-` are usually Google-Drive-specific. If multiple backends support this operation we can consider a more general name.

Next, I'm not sure if `@microsoft.graph.downloadUrl` will be what most users expect, since that link is [only valid for an hour](https://docs.microsoft.com/en-us/onedrive/developer/rest-api/resources/driveitem?view=odsp-graph-online#instance-attributes). An alternative (but hacky) approach is the transformation exemplified below:

If you get the public link of an item in the form of `https://contoso-my.sharepoint.com/:p:/g/personal/you_contoso_org/verylongandopaquestring` and transform it to `https://contoso-my.sharepoint.com/personal/you_contoso_org/_layouts/15/download.aspx?share=verylongandopaquestring`, this new link is directly downloadable and doesn't have the one-hour limit.

I don't think this is something documented so I can't guarantee it'll work forever, but personally I think it has been fairly stable.

septs:
OK
1. `--drive-direct-link` to `--direct-link` (general)
2. Timeliness this is a design, We should follow the designed usage (use-case)

ncw:
This looks like exactly what should be returned from the `PublicLink` interface - not sure why we need a new `DownloadLink` interface. Can you explain your reasoning @septs ?

Thanks!

septs:
@ncw 
`PublicLink` method based, onedrive `createLink` method
the method is create a sharing link (official share page)
see https://docs.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_createlink

`DownloadLink` method based, onedrive `DriveItem` object - `@microsoft.graph.downloadUrl` field
generated direct download link without token
https://docs.microsoft.com/en-us/onedrive/developer/rest-api/resources/driveitem#instance-attributes

septs:
image

ncw:
I see!

The docs for PublicLink() say "PublicLink generates a public link to the remote path (usually readable by anyone)" so maybe we should change the implementation of PublicLink in the onedrive backend to be your new DownloadLink code?

What do you think?

septs:
@ncw This is definitely break semantics.
see Google Drive implementation
see #3930

PR #4296, opened 2020-06-01
docs: note the sort order

dandv:
Specify sort order for https://rclone.org/commands/rclone_ls/

### DIFF
--- a/cmd/ls/lshelp/lshelp.go
+++ b/cmd/ls/lshelp/lshelp.go
@@ -8,8 +8,8 @@ There are several related list commands
 
   * ` + "`ls`" + ` to list size and path of objects only
   * ` + "`lsl`" + ` to list modification time, size and path of objects only
-  * ` + "`lsd`" + ` to list directories only
-  * ` + "`lsf`" + ` to list objects and directories in easy to parse format
+  * ` + "`lsd`" + ` to list directories only, sorted ascending
+  * ` + "`lsf`" + ` to list objects and directories in easy to parse format, sorted ascending
   * ` + "`lsjson`" + ` to list objects and directories in JSON format
 
 ` + "`ls`,`lsl`,`lsd`" + ` are designed to be human readable.


cmd/ls/lshelp/lshelp.go
```diff
@@ -8,8 +8,8 @@ There are several related list commands
 
   * ` + "`ls`" + ` to list size and path of objects only
   * ` + "`lsl`" + ` to list modification time, size and path of objects only
-  * ` + "`lsd`" + ` to list directories only
-  * ` + "`lsf`" + ` to list objects and directories in easy to parse format
+  * ` + "`lsd`" + ` to list directories only, sorted ascending
```
ncw, line 11: This will only be sorted if the `--max-depth` command isn't used...

cmd/ls/lshelp/lshelp.go
```diff
@@ -8,8 +8,8 @@ There are several related list commands
...
+  * ` + "`lsf`" + ` to list objects and directories in easy to parse format, sorted ascending
```
ncw, line 12: And likewise this one won't be sorted unless `-R` is passed

dandv: In my tests with 1.52, neither command recursed, so I haven't used `-R` or `--max-depth` on either. Docs say "The other list commands lsd,lsf,lsjson do not recurse by default - use "-R" to make them recurse.". I ran `lsd` and `lsf` against the root dir, where I had the files I wanted sorted, and directories to recurse into as well (but they weren't recursed into). Beyond that, I don't know :)

ncw:
I think that having the directories sorted is not a guarantee I want to make at the moment.

Really the list commands should have a sort flag if you want them sorted.

So I'm not going to merge this, sorry.

Thank you for opening up the discussion of the issue.

dandv:
A way to guarantee sorting would be useful. My use case is keeping only the
N most recent backup files, named in ISO8601 format so I can sort by the
filename.

Should I pipe the lsf output to sort instead?

ncw:
> Should I pipe the lsf output to sort instead?

You don't need to if you aren't using `-R`. if you do use `-R` then you will need sort to get a consistent order.

dandv:
No recursion used or needed in any of my use cases above, but I'm confused by the apparent contradiction between these statements:

* "And likewise this one won't be sorted unless -R is passed" - [source](https://github.com/rclone/rclone/pull/4296#discussion_r433325417). "This one" refers to `lsf`, presumably? So `lsf` output won't be sorted unless -R is passed? To remove the double negation, **the output of `lsf` will be sorted only with `-R`**.
* "You don't need to [pipe the lsf output to sort] if you aren't using -R" [source](https://github.com/rclone/rclone/pull/4296#issuecomment-639458848). So **without `-R`, the output to `lsf` is already sorted**.

ncw:
Sorry, English is only my first language and I'm not very good at speaking it ;-)



> * So **without `-R`, the output to `lsf` is already sorted**.

This is correct. For the time being!

Issue #4315, opened 2020-06-07
putio: panic on (*Object).Open


schittler:
#### What is the problem you are having with rclone?

panics on `rclone move`

#### What is your rclone version (output from `rclone version`)

```
rclone v1.52.0
- os/arch: linux/amd64
- go version: go1.14.3
```

####  Which cloud storage system are you using? (eg Google Drive)

putio

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone move -P <remote name>:PATH/ .

`.`, to note, is a CIFS mount (using cifs-utils)

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0xe3ae60]

goroutine 20886 [running]:
github.com/rclone/rclone/backend/putio.(*Object).Open(0xc000185e40, 0x1d4eee0, 0xc000d7af00, 0xc0008fedf0, 0x1, 0x1, 0x0, 0x1, 0x80, 0x80)
	github.com/rclone/rclone/backend/putio/object.go:251 +0x220
github.com/rclone/rclone/fs/operations.(*ReOpen).open(0xc001209500, 0xc001209500, 0xc000a35ca0)
	github.com/rclone/rclone/fs/operations/reopen.go:96 +0x349
github.com/rclone/rclone/fs/operations.NewReOpen(0x1d4eee0, 0xc000d7af00, 0x1d64f00, 0xc000185e40, 0xa, 0xc0008fede0, 0x1, 0x1, 0x0, 0x0, ...)
	github.com/rclone/rclone/fs/operations/reopen.go:47 +0x10c
github.com/rclone/rclone/fs/operations.(*multiThreadCopyState).copyStream(0xc0005402d0, 0x1d4eee0, 0xc000d7af00, 0x0, 0x0, 0x0)
	github.com/rclone/rclone/fs/operations/multithread.go:74 +0x326
github.com/rclone/rclone/fs/operations.multiThreadCopy.func1(0xc0000ec700, 0x0)
	github.com/rclone/rclone/fs/operations/multithread.go:173 +0x45
github.com/rclone/rclone/vendor/golang.org/x/sync/errgroup.(*Group).Go.func1(0xc0007595c0, 0xc0007596e0)
[...]

PR #4317, opened 2020-06-08
putio: fix panic on Object.Open #4315

cenkalti:
#### Checklist


backend/putio/object.go
```diff
@@ -248,7 +248,10 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 		_ = resp.Body.Close()
 		return nil, fserrors.NoRetryError(err)
 	}
-	return resp.Body, err
```
ncw, line 251: I'm not a big fan of bare returns with unamed parameters...

How about this which would be a bit more traditional?

```go
if err != nil {
    return nil, err
}
return resp.Body, nil
```

cenkalti: I have updated the commit.

PR #4329, opened 2020-06-09
accounting: use rolling average for speed display - fixes #3909

### LINKED ISSUE #3909: Use rolling average to improve ETA accuracy

hexchain:
Maintain and display a rolling average speed.

Use two rings to store the last few measurements of `duration` and `bytes`. Use the difference between the most recent and the oldest measurement to calculate transfer speed. The speed then needs to be updated before displayed.

#3909 (sort of)

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -15,6 +15,11 @@ import (
 	"github.com/rclone/rclone/lib/terminal"
 )
 
+const (
+	averagePeriodLength = time.Second
+	averageStopAfter    = time.Minute
+)
+
 // MaxCompletedTransfers specifies maximum number of completed transfers in startedTransfers list
 var MaxCompletedTransfers = 100
 
@@ -48,6 +53,18 @@ type StatsInfo struct {
 	oldDuration       time.Duration // duration of transfers we have culled
 	group             string
 	startTime         time.Time // the moment these stats were initialized or reset
+	average           averageValues
+}
+
+type averageValues struct {
+	mu        sync.Mutex
+	lpBytes   int64
+	lpTime    time.Time
+	speed     float64
+	stop      chan bool
+	stopped   sync.WaitGroup
+	startOnce sync.Once
+	stopOnce  sync.Once
 }
 
 // NewStats creates an initialised StatsInfo
@@ -60,6 +77,7 @@ func NewStats(ctx context.Context) *StatsInfo {
 		transferring: newTransferMap(ci.Transfers, "transferring"),
 		inProgress:   newInProgress(ctx),
 		startTime:    time.Now(),
+		average:      averageValues{stop: make(chan bool)},
 	}
 }
 
@@ -109,7 +127,7 @@ func (s *StatsInfo) RemoteStats() (out rc.Params, err error) {
 	return out, nil
 }
 
-// Speed returns the average speed of the transfer in bytes/second
+// speed returns the average speed of the transfer in bytes/second
 //
 // Call with lock held
 func (s *StatsInfo) speed() float64 {
@@ -274,15 +292,59 @@ func (s *StatsInfo) calculateTransferStats() (ts transferStats) {
 	// note that s.bytes already includes transferringBytesDone so
 	// we take it off here to avoid double counting
 	ts.totalBytes = s.transferQueueSize + s.bytes + transferringBytesTotal - transferringBytesDone
-
-	dt := s.totalDuration()
-	ts.transferTime = dt.Seconds()
-	ts.speed = 0.0
-	if dt > 0 {
-		ts.speed = float64(s.bytes) / ts.transferTime
-	}
+	ts.speed = s.average.speed
 
 	return ts
+}
+
+func (s *StatsInfo) averageLoop() {
+	var period float64
+
+	ticker := time.NewTicker(averagePeriodLength)
+	defer ticker.Stop()
+
+	startTime := time.Now()
+	a := &s.average
+	defer a.stopped.Done()
+	for {
+		select {
+		case now := <-ticker.C:
+			a.mu.Lock()
+			var elapsed float64
+			if a.lpTime.IsZero() {
+				elapsed = now.Sub(startTime).Seconds()
+			} else {
+				elapsed = now.Sub(a.lpTime).Seconds()
+			}
+			avg := 0.0
+			if elapsed > 0 {
+				avg = float64(a.lpBytes) / elapsed
+			}
+			if period < averagePeriod {
+				period++
+			}
+			a.speed = (avg + a.speed*(period-1)) / period
+			a.lpBytes = 0
+			a.lpTime = now
+			a.mu.Unlock()
+		case <-a.stop:
+			return
+		}
+	}
+}
+
+func (s *StatsInfo) startAverageLoop() {
+	s.average.startOnce.Do(func() {
+		s.average.stopped.Add(1)
+		go s.averageLoop()
+	})
+}
+
+func (s *StatsInfo) stopAverageLoop() {
+	s.average.stopOnce.Do(func() {
+		close(s.average.stop)
+		s.average.stopped.Wait()
+	})
 }
 
 // String convert the StatsInfo to a string for printing
@@ -424,6 +486,10 @@ func (s *StatsInfo) Log() {
 
 // Bytes updates the stats for bytes bytes
 func (s *StatsInfo) Bytes(bytes int64) {
+	s.average.mu.Lock()
+	s.average.lpBytes += bytes
+	s.average.mu.Unlock()
+
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.bytes += bytes
@@ -549,6 +615,9 @@ func (s *StatsInfo) ResetCounters() {
 	s.renames = 0
 	s.startedTransfers = nil
 	s.oldDuration = 0
+
+	s.stopAverageLoop()
+	s.average = averageValues{stop: make(chan bool)}
 }
 
 // ResetErrors sets the errors count to 0 and resets lastError, fatalError and retryError
@@ -629,6 +698,7 @@ func (s *StatsInfo) GetTransfers() int64 {
 func (s *StatsInfo) NewTransfer(obj fs.Object) *Transfer {
 	tr := newTransfer(s, obj)
 	s.transferring.add(tr)
+	s.startAverageLoop()
 	return tr
 }
 
@@ -636,6 +706,7 @@ func (s *StatsInfo) NewTransfer(obj fs.Object) *Transfer {
 func (s *StatsInfo) NewTransferRemoteSize(remote string, size int64) *Transfer {
 	tr := newTransferRemoteSize(s, remote, size, false)
 	s.transferring.add(tr)
+	s.startAverageLoop()
 	return tr
 }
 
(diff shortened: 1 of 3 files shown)


hexchain:
The `TestParseDuration` does not seem related.

On a side note: it should fail when `$TZ` is not `UTC`.

fs/accounting/stats.go
```diff
@@ -2,6 +2,7 @@ package accounting
 
 import (
 	"bytes"
+	"container/ring"
```
ncw, line 5: I'm not a big fan of the container/ring package it uses an allocation per element and another allocation for the interface it wedges the element in to.

Using a slice of values is a lot more efficient...

fs/accounting/stats.go
```diff
@@ -13,8 +14,13 @@ import (
...
 )
 
-// MaxCompletedTransfers specifies maximum number of completed transfers in startedTransfers list
-var MaxCompletedTransfers = 100
+var (
+	// MaxCompletedTransfers specifies maximum number of completed transfers in startedTransfers list
+	MaxCompletedTransfers = 100
+
+	// SpeedRingSize specifies the size of duration and speed rings for computing the rolling average
+	SpeedRingSize = 16
```
ncw, line 22: The other speed transfer measures in rclone use [exponentially weighted averages](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average) which work very well and are a lot easier to calculate as you don't need to store the history.

https://github.com/rclone/rclone/blob/68afa28b27f922f1cdf5e872549a870ff2a84401/fs/accounting/accounting.go#L179-L184

fs/accounting/stats.go
hexchain: Sure, a slice is better in every way.

fs/accounting/stats.go
hexchain: Yes, using that seems to be a good idea. But according to my experience, it decays too slowly when there is a sudden drop. What I want to achieve is, when the network or the remote service gets stuck, the speed drops to zero in a few seconds.

hexchain: Hi, do you think something like this still worth pursuing? Or should I just (re)use the existing exponentially weighted average here?

ncw: Let's try the exponentially weighted average as it is a lot simpler - we can then consider whether we want to switch it out.

Sorry for the delay in responding!

hexchain:
Oops, reopened, and done.

ncw:
I've rebased this and in the process or reviewing it, I've realised there is a problem...

For the exponentially weighted moving average to work properly this means that `UpdateStats` should be called at regular intervals like this one here

https://github.com/rclone/rclone/blob/b2ae94de5b7cfeb9d9e7534a850042f13f9639b8/fs/accounting/accounting.go#L177

Your code calls UpdateStats regularly when using --progress and this will work very well, however when not using --progress then that is when things will start to go wrong. If we are just using -v then by default this will only get called once per minute.

So this really needs a similar loop to the averageLoop updating the values every second.

Do you want to have a go at that? Or I could merge this and fix that up myself?

Sorry this PR has taken such a long time to merge - my fault for getting distracted with other things!

hexchain:
> Your code calls UpdateStats regularly when using --progress and this will work very well, however when not using --progress then that is when things will start to go wrong. If we are just using -v then by default this will only get called once per minute.
> 
> So this really needs a similar loop to the averageLoop updating the values every second.
> 

But I don't think it will be a problem? When using only `-v`, before this PR it should output the average speed since the start of the transfer every minute, and now it should output the weighted average every minute. It should be more accurate than before anyway.

Also, `UpdateSpeed()` only updates `s.lastSpeed`, which is only used for displaying.

hexchain:
Friendly ping?

ncw:
I'd like to have these stats updated once per second so they are consistent for the user.

If we only call the stats updater once per minute then the exponential average will take approximately 15-30 minutes to converge on a value which I don't think is what the user will be expecting.

hexchain:
Fair enough, I'll change that.

hexchain:
Updated.

I've tried to use something similar to `averageLoop` by starting a goroutine in `startProgress` that has a ticker and calls `UpdateSpeed` itself, but once in every 4~5 times, it fails to update the speed before printing progress.

ivandeex:
I suspect this PR solves diferently the same problem as
#3929 (compute total running average speed and use that for ETA and display)

hexchain:
Rebased.

ivandeex:
Closing PR #3929 in favor of this PR.

ivandeex:
@ncw please review

ivandeex:
@ncw This PR waits for your attention. Please review it.

ivandeex:
Well, this PR is out of luck. I am aware of another one.
Maybe @ncw one day finds time to look into this.
I won't spam with repeated "please look here" anymore.

ivandeex:
from linked issue https://github.com/rclone/rclone/issues/3909#issuecomment-628788504

> It would be better to average the total otherwise the speed will be extremely bumpy when new transfers start I think.

So I think one of criteria for accepting this will be the smooth flow of estimated speed **at start** as well as in-fly.
Something to be checked by the author at least visually before @ncw finally finds time to look into this.

@hexchain 
I might find it useful looking into other points related to bandwidth smoothening in rclone: https://github.com/rclone/rclone/commit/463a18aa0713a55e12e4cb6d201144121fc60ba5 (and followup https://github.com/rclone/rclone/issues/4987)

ivandeex:
@ncw I think it's a good time to take a look at this PR.

ncw (commented):
Thank you for your patience @hexchain 

The calculation parts of this PR look great.

However the timing parts are wrong. This starts a timer in `startProgress` - however this is only called if `-P` is in use.

The stats need to work via the rc for example so the webgui has nice stats too.

For exponentially weighted averages to work properly they need to be called at the same frequency. So I think we need to start a go routine that ticks at once per second in `stats.go` in `NewStats`.

Now that isn't ideal since we'll have a goroutine for every stats group and we'll need to stop the goroutine in `func (sg *statsGroups) delete(group string)`

Ideally we'd start the go routine in the StatsInfo only when a transfer starts in that StatsInfo and stop it 1 minute (say) after the last transfer (zeroing the speed at that moment).

Do you want to have a go at that @hexchain ? If not I'm willing to take that bit over.

hexchain:
I'd like to have a try in a few days!

ncw:
> I'd like to have a try in a few days!

Great - looking forward to it :-)

ivandeex:
Unfortunately this PR was affected by the recent Stats patches (#5087) and needs a rebase.
Odds are that #5237 can bring more Stats conflicts if it gets merged soon.

ivandeex:
Adding the `waiting...` badge as a reminder for @ncw and @hexchain ....

hexchain:
I've implemented everything in the comment https://github.com/rclone/rclone/pull/4329#pullrequestreview-627518439.

The other commit is a simple fix for 35a86193b7c7ba190e9a70b62a5a07cc010bf9e1: after that commit, the "Running since" field on the web UI is always 0. Fixed by using `GlobalStats().startTime` to calculate uptime.

hexchain:
And sorry for the delay. My day job suddenly caught me in April, but things have settled down, fortunately.

ivandeex:
@hexchain 
You can locally run this
```
cd rclone
go test -count=1 -v ./fs/sync	
```
to find out why the test deadlocks after recent changes

hexchain:
Strangely, the race tests seem to have some problems on my system (Arch Linux, 5.12.11): the `cmd/mount` test sometimes hangs indefinitely, and kernel complains that a task is stalled (see below). I have to force unmount the FUSE mountpoint to even stop the process. This happens even with the master branch, but isn't seen on CI.

```
[  245.509663] INFO: task mount.test:5599 blocked for more than 122 seconds.
[  245.509675]       Not tainted 5.12.11-arch1-1 #1
[  245.509679] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[  245.509681] task:mount.test      state:D stack:    0 pid: 5599 ppid:  2626 flags:0x00000000
[  245.509689] Call Trace:
[  245.509695]  __schedule+0x2ff/0x8b0
[  245.509706]  schedule+0x5b/0xc0
[  245.509711]  schedule_preempt_disabled+0x11/0x20
[  245.509716]  __mutex_lock.constprop.0+0x367/0x550
[  245.509723]  do_epoll_wait+0x24b/0x760
[  245.509731]  do_compat_epoll_pwait.part.0+0xc/0x70
[  245.509735]  __x64_sys_epoll_pwait+0x75/0xa0
[  245.509739]  do_syscall_64+0x33/0x40
[  245.509745]  entry_SYSCALL_64_after_hwframe+0x44/0xae
[  245.509749] RIP: 0033:0x4aaf60
[  245.509753] RSP: 002b:00007fd843ffe6f0 EFLAGS: 00000246 ORIG_RAX: 0000000000000119
[  245.509758] RAX: ffffffffffffffda RBX: 0000000000000000 RCX: 00000000004aaf60
[  245.509761] RDX: 0000000000000080 RSI: 00007fd843ffe740 RDI: 0000000000000004
[  245.509764] RBP: 00007fd843ffed40 R08: 0000000000000000 R09: 0000000000000020
[  245.509766] R10: 0000000000000000 R11: 0000000000000246 R12: 00007fd843ffecf0
[  245.509768] R13: 000000c00007e800 R14: 000000c000702780 R15: 0000000000000000
```

`/proc/<pid>/stack`:
```
[<0>] request_wait_answer+0x17c/0x2a0 [fuse]
[<0>] fuse_simple_request+0x19b/0x310 [fuse]
[<0>] fuse_file_poll+0x199/0x210 [fuse]
[<0>] ep_item_poll.isra.0+0x2d/0x50
[<0>] do_epoll_wait+0x382/0x760
[<0>] do_compat_epoll_pwait.part.0+0xc/0x70
[<0>] __x64_sys_epoll_pwait+0x75/0xa0
[<0>] do_syscall_64+0x33/0x40
[<0>] entry_SYSCALL_64_after_hwframe+0x44/0xae
```

ivandeex:
cmd/mount is known to be unstable, on github runners it fails with ~10% probability
I think we are fine.
Let's wait for @ncw

ivandeex:
Thank you for your efforts!

ncw:
Thank you for doing this @hexchain and for your patience!

I'm going to merge this now for 1.56 and if it needs any fixes they can come in further PRs :-)

PR #4336, opened 2020-06-11
sftp: Fix SSH key PEM loading

z-0:

This PR fixes a bug in the SFTP Backend.
For SSH authentication, `key_pem` should both override `key_file`
and not require other SSH authentication methods to be set.

Prior to this fix, rclone would attempt to use an ssh-agent
when `key_pem` was the only authentication method set.

This fix has not been discussed.
[Original feature PR](https://github.com/rclone/rclone/pull/4240)
[Original feature discussion](https://forum.rclone.org/t/add-ssh-key-to-config-file-rather-than-filesystem-pointer/16404)

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -395,8 +395,8 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 
 	keyFile := env.ShellExpand(opt.KeyFile)
 	//keyPem := env.ShellExpand(opt.KeyPem)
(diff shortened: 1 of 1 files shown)


ncw:
That is great thank you 😃

I think this deserves to go in the next point release - what do you think?

z-0:
I would be happy to see it go in to the next release, especially as I want to use this feature 😄

ncw:
Ok I stuck that in the queue for 1.52.2!

PR #4361, opened 2020-06-16
Fix custom timezone in rclone docker image

NoLooseEnds:

Added the tzdata package to the dockerfile to fix custom timezone

https://forum.rclone.org/t/rclone-in-docker-uses-tz-as-utc-not-as-provided/17173


ncw:
That looks great thanks! Will merge now

PR #4362, opened 2020-06-16
dlna: Mark flags in docs as code

tipabu:

Mark the flags for `serve dlna` as code in the documentation. Otherwise, we get en dashes in the man page, making args more difficult to copy/paste to a command line. Previously, the man page would show flags like `–addr`; now it shows `--addr`.

No.

#### Checklist

### DIFF
--- a/cmd/serve/dlna/dlnaflags/dlnaflags.go
+++ b/cmd/serve/dlna/dlnaflags/dlnaflags.go
@@ -10,14 +10,14 @@ import (
 var Help = `
 ### Server options
 
-Use --addr to specify which IP address and port the server should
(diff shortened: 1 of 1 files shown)


ncw:
> Otherwise, we get en dashes in the man page, making args more difficult to copy/paste to a command line. Previously, the man page would show flags like `–addr`; now it shows `--addr`.

So it does. I never noticed this before... I'll fix the converter in just a moment.

I'll merge your fix - it makes the docs look nicer - thanks!

ncw:
I've fixed the underlying issue here fcc2db80939bb17619078877afd789148cafb4bd - thank you for pointing it out :-)

I had the same problem on the website some time ago - I didn't realise it would be in the man page (and also in MANUAL.html).

Issue #4326, opened 2020-06-09
Add cleanup support for Box


ambipur76:
#### What is your current rclone version (output from `rclone version`)?

1.51.0

#### What problem are you are trying to solve?

When performing an automated backup it is essential to be able to cleanup the Trash / Recycle Bin folder to free up the complete storage space.

#### How do you think rclone should be changed to solve that?

The rclone cleanup <remote> command could support this feature.

ncw:
When we looked at this here https://github.com/rclone/rclone/issues/575#issuecomment-330028744 we decided

> **box**: The Box API provides an endpoint for [listing trashed items](https://developer.box.com/reference#get-the-items-in-the-trash) and [permanently deleting trashed items](https://developer.box.com/reference#permanently-delete-a-trashed-file). However, there's no endpoint for permanently deleting trashed items and the Box API does not support batch API calls. Cleanup is not feasible since it involves listing all trashed items and making a delete request for each item. [Docs](https://developer.box.com/reference)

So I guess we could implement it but it might take a really long time.

Is this something you'd like to help with @ambipur76

PR #4364, opened 2020-06-17
Box cleanup

buengese:
As it was discussed in #4326. It's not great but reasonably fast. I've added some notes about this limitation to the docs.

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -1063,6 +1063,66 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 	return info.SharedLink.URL, err
 }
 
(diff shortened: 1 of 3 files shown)


ncw:
That looks great! Do the backend integration tests pass? If so then please merge 😃

buengese:
Of course they pass nothing here should break anything but I tend to run anyway for good measure. I don't think we even have an integration test for cleanup (either that or I missed it). That might be worth looking but it would probably boil to to simple require.NoError() because I'm not sure that all remotes that support emptying the trash also support listing it's contents.

As a side note: I generally leave these PRs open for at least 24h to give you a chance to have a quick look in case you have any notes.

PR #4399, opened 2020-06-30
gcs: add support for anonymous access

pothos:

Currently credentials are required to download a public bucket file
    which is not really necessary and makes automated usage more complex.
    Add a new option "anonymous" which when enabled configures the
    gcs backend to use an anonymous HTTP client. This of course only works
    for read access and trying to write will lead to errors like that:
    "googleapi: Error 401: Anonymous caller does not not have
    storage.objects.create access to the Google Cloud Storage object.",
    as expected. By default the anonymous access option is disabled so that
    the GCS Application Default Credentials are still used by default as
    before and an error is given if they can't be found.

no

#### Checklist

I didn't add tests - does this apply? I tested it manually but it could be tested with some hardcoded public bucket.

### DIFF
--- a/backend/googlecloudstorage/googlecloudstorage.go
+++ b/backend/googlecloudstorage/googlecloudstorage.go
@@ -79,7 +79,8 @@ func init() {
 		Config: func(name string, m configmap.Mapper) {
 			saFile, _ := m.Get("service_account_file")
 			saCreds, _ := m.Get("service_account_credentials")
-			if saFile != "" || saCreds != "" {
+			anonymous, _ := m.Get("anonymous")
+			if saFile != "" || saCreds != "" || anonymous == "true" {
 				return
 			}
 			err := oauthutil.Config("google cloud storage", name, m, storageConfig, nil)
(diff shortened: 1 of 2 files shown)


ncw:
This is a nice feature. I'm not sure about the configuration mode though.

When we did this for S3 we went anonymous when all the credentials were empty.

Another approach might be to explicitly have an anonymous flag.

What do you think?

pothos:
Thanks for the feedback. I added an explicit flag that takes precedence. This way anonymous access can be chosen even if default credentials are available. Fall back to anonymous access on error would be handy but also confusing because users with wrongly configured credentials would only notice this when they try to upload something but not on download.

ncw:
> Thanks for the feedback. I added an explicit flag that takes precedence.

Nice! What do you think about calling it `anonymous` rather than `anonymous_access`? That makes the command line flag `--gcs-anonymous` which seems unambiguous and easy to type?

> This way anonymous access can be chosen even if default credentials are available.

That will make it very useful as a command line flag.

> Fall back to anonymous access on error would be handy but also confusing because users with wrongly configured credentials would only notice this when they try to upload something but not on download.

I think guessing here is probably the wrong thing to do.

pothos:
Thanks, have updated the PR to use the name `anonymous` for the option.

ncw:

I'll merge it now.

pothos:
Thank you for the guidance

PR #4411, opened 2020-07-04
Updated install.md to reflect minimum go version

### LINKED ISSUE #3765: Install from source fails with import errors, later version required?
The installation from source instructions from https://rclone.org/install/ specify that Go 1.7 or later is sufficient for install, but trying to install using the given command: `go get -u -v github.com/rclone/rclone` results in:

```
> go get -u -v github.com/rclone/rclone
github.com/rclone/rclone (download)
package math/bits: unrecognized import path "math/bits" (import path does not begin with hostname)
package github.com/rclone/rclone/vendor/google.golang.org/api/transport/http/internal/propagation: no buildable Go source files in /home/eharris/prg/go/src/github.com/rclone/rclone/vendor/google.golang.org/api/transport/http/internal/propagation
package plugin: unrecognized import path "plugin" (import path does not begin with hostname)
```

This was generated using Go 1.7.4 on Debian Stretch 9.

eharris:
Fixes #3765.

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -174,7 +174,7 @@ kill %1
 
 ## Install from source ##
 
-Make sure you have at least [Go](https://golang.org/) 1.7
+Make sure you have at least [Go](https://golang.org/) 1.10
 installed.  [Download go](https://golang.org/dl/) if necessary.  The
 latest release is recommended. Then
 


eharris:
Updated.

eharris:
It appears that the docs at https://rclone.org/install/ still reflect the incorrect info.  Does something else need to be done to publish this change?

ncw:
It will go out when v1.53 is released or v1.52.3 whichever comes first!

Issue #4429, opened 2020-07-12
rclone sync complains of duplicate objects in destination, ignores and fails to remove them


eharris:
### What is the problem you are having with rclone?

`rclone sync <localpath> remote:path` sometimes gives errors about duplicate objects:

```
2020/07/12 02:54:56 NOTICE: 02 - Pixies - Tame.flac: Duplicate object found in destination - ignoring
2020/07/12 02:54:56 NOTICE: 03 - Pixies - Wave of Mutilation.flac: Duplicate object found in destination - ignoring
```

Successive runs of sync continue to emit these errors, and the duplicates are never cleaned up/dealt with.

Note that I'm concerned primarily with the fact that rclone isn't cleaning up the dupes when `sync` is specified.  So far I'm unclear about how the dupes occurred in the first place (probably from a run that received errors), so this ticket is only about not dealing with them once they exist.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.52.2-DEV
- os/arch: linux/amd64
- go version: go1.11.6
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Debian Buster amd64

####  Which cloud storage system are you using? (eg Google Drive)

Opendrive with crypt overlay (standard)

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`rclone sync 'audio/Archives_FLAC/Pixies - Doolittle' 'odc:audio/Archives_FLAC/Pixies - Doolittle'`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
> rclone sync -vv 'audio/Archives_FLAC/Pixies - Doolittle' 'odc:audio/Archives_FLAC/Pixies - Doolittle'
2020/07/12 02:54:50 DEBUG : rclone: Version "v1.52.2" starting with parameters ["rclone" "sync" "-vv" "audio/Archives_FLAC/Pixies - Doolittle" "odc:audio/Archives_FLAC/Pixies - Doolittle"]
2020/07/12 02:54:50 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2020/07/12 02:54:51 DEBUG : Starting OpenDrive session with ID: <hidden>
2020/07/12 02:54:53 DEBUG : Starting OpenDrive session with ID: <hidden>
2020/07/12 02:54:56 NOTICE: 02 - Pixies - Tame.flac: Duplicate object found in destination - ignoring
[...]

PR #4431, opened 2020-07-12
docs: Updated sync and dedupe command docs #4429

eharris:
Update the docs regarding sync and dedupe commands with info regarding handling duplicate objects.  See ticket #4429.

### DIFF
--- a/cmd/sync/sync.go
+++ b/cmd/sync/sync.go
@@ -27,7 +27,8 @@ var commandDefinition = &cobra.Command{
 Sync the source to the destination, changing the destination
 only.  Doesn't transfer unchanged files, testing by size and
 modification time or MD5SUM.  Destination is updated to match
-source, including deleting files if necessary.
+source, including deleting files if necessary (except duplicate
+objects, see below).
 
 **Important**: Since this can cause data loss, test first with the
 ` + "`--dry-run` or the `--interactive`/`-i`" + ` flag.
@@ -35,7 +36,8 @@ source, including deleting files if necessary.
     rclone sync -i SOURCE remote:DESTINATION
 
 Note that files in the destination won't be deleted if there were any
-errors at any point.
+errors at any point.  Duplicate objects (files with the same name, on
+those providers that support it) are also not yet handled.
 
 It is always the contents of the directory that is synced, not the
 directory so when source:path is a directory, it's the contents of
@@ -46,6 +48,9 @@ If dest:path doesn't exist, it is created and the source:path contents
 go there.
 
 **Note**: Use the ` + "`-P`" + `/` + "`--progress`" + ` flag to view real-time transfer statistics
+
+**Note**: Use the ` + "`rclone dedupe`" + ` command to deal with "Duplicate object/directory found in source/destination - ignoring" errors.
+See [this forum post](https://forum.rclone.org/t/sync-not-clearing-duplicates/14372) for more info.
 `,
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(2, 2, command, args)
--- a/cmd/dedupe/dedupe.go
+++ b/cmd/dedupe/dedupe.go
@@ -32,8 +32,8 @@ By default ` + "`dedupe`" + ` interactively finds files with duplicate
 names and offers to delete all but one or rename them to be
 different. This is known as deduping by name.
 
-Deduping by name is only useful with backends like Google Drive which
-can have duplicate file names. It can be run on wrapping backends
+Deduping by name is only useful with a small group of backends (e.g. Google Drive,
+Opendrive) that can have duplicate file names. It can be run on wrapping backends
 (e.g. crypt) if they wrap a backend which supports duplicate file
 names.
 


docs/content/commands/rclone_dedupe.md
```diff
@@ -13,8 +13,8 @@ Interactively find duplicate files and delete/rename them.
 
 
 By default `dedupe` interactively finds duplicate files and offers to
-delete all but one or rename them to be different. Only useful with
-Google Drive which can have duplicate file names.
+delete all but one or rename them to be different.  Only some backends
```
ncw, line 16: This is an autogenerated file - the file which needs editing is in `cmd/dedupe`

docs/content/commands/rclone_sync.md
```diff
@@ -15,13 +15,15 @@ Make source and dest identical, modifying destination only.
 Sync the source to the destination, changing the destination
 only.  Doesn't transfer unchanged files, testing by size and
 modification time or MD5SUM.  Destination is updated to match
-source, including deleting files if necessary.
+source, including deleting files if necessary (except duplicate
```
ncw, line 18: This is an autogenerated file - the file which needs editing is in `cmd/sync`

docs/content/commands/rclone_dedupe.md
```diff
@@ -13,8 +13,8 @@ Interactively find duplicate files and delete/rename them.
...
+(e.g. Google Drive, Opendrive) have the ability to store duplicate file names.
```
ncw, line 17: I'm not sure the Opendrive isn't a bug?

Google Drive and Mega definitely can and have rclone dedupe support. Opendrive doesn't have dedupe support yet so can't actually run dedupe.

docs/content/commands/rclone_sync.md
```diff
@@ -32,6 +34,8 @@ If dest:path doesn't exist, it is created and the source:path contents
 go there.
 
 **Note**: Use the `-P`/`--progress` flag to view real-time transfer statistics
+**Note**: Use the `rclone dedupe` command to deal with "Duplicate object found in destination" errors.
```
ncw, line 37: I think this needs a newline before it otherwise it will be rolled into a single paragraph

ncw, line 37: This should probably say ... to deal with "Duplicate object/directory found in source/destination - ignoring" errors as the error message can be any of those.

docs/content/commands/rclone_dedupe.md
eharris: I don't know for sure if Opendrive is a bug or unsupported feature, but dupes and `rclone dedupe` (of files at least) seems to work.

cmd/sync/sync.go
```diff
@@ -46,6 +48,9 @@ If dest:path doesn't exist, it is created and the source:path contents
 go there.
 
 **Note**: Use the ` + "`-P`" + `/` + "`--progress`" + ` flag to view real-time transfer statistics
+
+**Note**: Use the `rclone dedupe` command to deal with "Duplicate object/directory found in source/destination - ignoring" errors.
```
ncw, line 52: This needs the ` dealing with like the ones in the line above

docs/content/commands/rclone_dedupe.md
ivandeex: I removed "Opendrive" from this line

ivandeex:
@eharris If you set the check "Allow maintainers to edit this PR" on the PR page, please let me know. I will be able to rebase for you then.

docs/content/commands/rclone_dedupe.md
ivandeex: @eharris I propose to keep Google Drive and keep questionable Opendrive for future. It really isn't importand at all.

ivandeex: This small change hangs for too long. Let's have it in.

ivandeex:
The last @eharris reply is from July 2020. Marking as draft until this PR shows signs of life.

eharris:
I pushed an update to fix the merge conflict over a week ago. What else is needed?

ivandeex:
See inline comment thread

ivandeex:
@ncw 

We have a severe communication problem here.
Here is the "discussion" that started in June 2020 in the inline thread :-/

@ncw: `Opendrive doesn't have dedupe support yet so can't actually run dedupe.`
@eharris: `I don't know for sure if Opendrive is a bug or unsupported feature, but dupes and rclone dedupe (of files at least) seems to work.`

7 months later!!!!

@ivandeex: `I propose to keep Google Drive and keep questionable Opendrive for future. It really isn't importand at all.`
@eharris (out of thread): `I pushed an update to fix the merge conflict over a week ago. What else is needed?`

Obviously we have a communication problem.
I (@ivandeex) tried to push a patch that removes questionable word from docs but got permission denied.
Obviously we have a permission problem on top of communication problem.

I'm marking this thread with `regional` label which I normally use for threads with broken communication (usually due to bad English but probably for other reasons, e.g. see many threads with zhucаn where they continuously copy-paste the same sentence slightly edited instead of answering questions)

P.S. The whole patch here is less than 10 lines of code............

docs/content/commands/rclone_dedupe.md
eharris: Dupes are still a problem on Opendrive, and the `rclone dedupe` command does currently work (as I made clear in the inline thread) so I'm unclear as to why you want to remove Opendrive from being mentioned as a backend that has this issue.  What exactly are the changes you want me to make?

ivandeex:
@ncw
Please review this docs-only change.
(The author insists on having Opendrive)

eharris:
To be clear, I just think it's important to try to dispel the myth that dupes is only a problem for Google Drive, which is not the case.

ivandeex:
@animosity22 

In https://github.com/rclone/rclone/issues/4429#issuecomment-660263357 you wrote:

> > Whether the endpoints supports duplicates may not be known, but I would also argue that it doesn't matter since the behavior is the same. That is, even if the source and/or destination supports dupes, if the source doesn't contain any dupes for a particular file, then it should still use the methodology for case 2.
> >
> > [snip]
>
> That makes sense to me as well and a very good descriptions.
> This is only for OpenDrive and Google Drive I think as well or is there another one that would be impacted?

Now the author proposes this help string as part his patch (it's small):
```
Only some backends (e.g. Google Drive, Opendrive) have the ability to store duplicate file names.
```

Idea: review the patch and make a verdict

animosity22:
That looks good to me. I primarily use Google so not quite as familiar with the others, but adding more backends and listing them out seems proper to me.

ncw:
OK I will merge this now. Thanks for your patience @eharris :-)

PR #4432, opened 2020-07-13
docs: Updated crypt/cryptdecode docs with additional info

eharris:
Add additional links to crypt/cryptdecode docs.

### DIFF
--- a/cmd/cryptdecode/cryptdecode.go
+++ b/cmd/cryptdecode/cryptdecode.go
@@ -36,6 +36,9 @@ use it like this
 	rclone cryptdecode encryptedremote: encryptedfilename1 encryptedfilename2
 
 	rclone cryptdecode --reverse encryptedremote: filename1 filename2
+
+Another way to accomplish this is by using the ` + "`rclone backend encode` (or `decode`)" + `command.
+See the documentation on the ` + "`crypt`" +  ` overlay for more info.
 `,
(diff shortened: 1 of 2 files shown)


docs/content/commands/rclone_cryptdecode.md
```diff
@@ -40,4 +40,5 @@ See the [global flags page](/flags/) for global options not listed here.
 ## SEE ALSO
 
 * [rclone](/commands/rclone/)	 - Show help for rclone commands, flags and backends.
+* [crypt](/crypt/)	 - Info about the crypt overlay
```
ncw, line 43: This is an auto generated file from `rclone cryptdecode --help` so I don't think you can patch it like that alas.

You could put a link in the docs in `cmd/cryptdecode/cryptdecode.go`

eharris: The source code doesn't appear to contain the `## SEE ALSO` portion of the docs.  How do I add a customized item to that?

ncw: That is created by cobra I think. I'm not 100% sure how you do that!

I suggest just adding a note at the bottom of the main text.

ivandeex:
@ncw Please review

ncw (approved):
That looks great - thank you

I'll merge it now

(Sorry for the delay!)

PR #4434, opened 2020-07-14
docs: Updated mount command to reflect that it requires Go 1.13 or newer

eharris:
After being confused as to why the mount command disappeared from my rclone after updating, I discovered it was due to https://github.com/rclone/rclone/commit/17b4058ee944ab660b30593a858ff9d591b62a3b, so decided to add a note to the docs about it.

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -192,6 +192,9 @@ Stopping the mount manually:
     # OS X
     umount /path/to/local/mount
 
+**Note**: As of ` + "`rclone` 1.52.2, `rclone mount`" + ` now requires Go version 1.13
+or newer on some platforms depending on the underlying FUSE library in use.
+
 ### Installing on Windows
 
 To run rclone ` + commandName + ` on Windows, you will need to


ncw:
Sorry that caused you a problem.

That's an auto-generated file... The source is in cmd/mountlib/help.go I think.

This is a little complicated is that the cmount and mount2 commands have the same help but different go version constraints.

You could maybe add to the help in cmd/mount/ ?

eharris:
I couldn't find any help in `cmd/mount/`.  Also, since the `mount` command disappears from `rclone`, it disappears from the help (from the binary) as well.

cmd/mountlib/mount.go
```diff
@@ -149,6 +149,9 @@ Stopping the mount manually:
     # OS X
     umount /path/to/local/mount
 
+**Note**: As of `rclone` 1.52.2, `rclone mount` now requires Go version 1.13 or newer due to
```
ncw, line 152: Note that you can't use ` inside a ` string. You'll have to remove them or use " as appropriate.

eharris: Fixed.

ncw:
That is almost perfect. However we use the same help for `rclone cmount`  which is called `rclone mount` on Windows and that definitely does work on old go versions. Likewise `rclone mount2`.

This needs to be parameterized somehow which is annoying. Any ideas?

eharris:
Rather than parameterize it, I decided it was easier just to note it is a limitation only on certain platforms.

ncw:
That looks great - thanks!

PR #4437, opened 2020-07-14
core/command: Allow rc to execute rclone terminal commands.

negative0:
This will allow rclone to run the commands which are not available directly through rc, over rc.

This api will spawn a temporary instance of rclone to execute the command. Then it will return the raw command output.
This will also help in the development of a web-based rclone terminal. (

#### Checklist

### DIFF
--- a/fs/rc/internal.go
+++ b/fs/rc/internal.go
@@ -5,6 +5,7 @@ package rc
 import (
 	"context"
 	"os"
+	"os/exec"
 	"runtime"
 	"time"
 
@@ -334,4 +335,111 @@ func rcSetBlockProfileRate(ctx context.Context, in Params) (out Params, err erro
 	}
 	runtime.SetBlockProfileRate(int(rate))
 	return nil, nil
+}
+
+func init() {
+	Add(Call{
+		Path:         "core/command",
+		AuthRequired: true,
+		Fn:           rcRunCommand,
+		Title:        "Run a rclone terminal command over rc.",
+		Help: `This takes the following parameters
+
+- command - a string with the command name
+- arg - a list of arguments for the backend command
+- opt - a map of string to string of options
+
+Returns
+
+- result - result from the backend command
+- error	 - set if rclone exits with an error code
+
+For example
+
+    rclone rc core/command command=ls -a mydrive:/ -o max-depth=1
+	rclone rc core/command -a ls -a mydrive:/ -o max-depth=1
+
+Returns
+
+` + "```" + `
+{
+	"error": false,
+	"result": "<Raw command line output>"
+}
+
+OR 
+{
+	"error": true,
+	"result": "<Raw command line output>"
+}
+
+
+`,
+	})
+}
+
+// rcRunCommand runs an rclone command with the given args and flags
+func rcRunCommand(ctx context.Context, in Params) (out Params, err error) {
+
+	command, err := in.GetString("command")
+	if err != nil {
+		command = ""
+	}
+
+	var opt = map[string]string{}
+	err = in.GetStructMissingOK("opt", &opt)
+	if err != nil {
+		return nil, err
+	}
+
+	var arg = []string{}
+	err = in.GetStructMissingOK("arg", &arg)
+	if err != nil {
+		return nil, err
+	}
+
+	var allArgs = []string{}
+	if command != "" {
+		// Add the command eg: ls to the args
+		allArgs = append(allArgs, command)
+	}
+	// Add all from arg
+	for _, cur := range arg {
+		allArgs = append(allArgs, cur)
+	}
+
+	// Add flags to args for eg --max-depth 1 comes in as { max-depth 1 }.
+	// Convert it to [ max-depth, 1 ] and append to args list
+	for key, value := range opt {
+		if len(key) == 1 {
+			allArgs = append(allArgs, "-"+key)
+		} else {
+			allArgs = append(allArgs, "--"+key)
+		}
+		allArgs = append(allArgs, value)
+	}
+
+	// Get the path for the current executable which was used to run rclone.
+	ex, err := os.Executable()
+	if err != nil {
+		return nil, err
+	}
+
+	cmd := exec.CommandContext(ctx, ex, allArgs...)
+
+	// Run the command and get the output for error and stdout combined.
+	output, err := cmd.CombinedOutput()
+
+	if err != nil {
+		fs.Errorf(nil, "Command error %v", err)
+		return Params{
+			"result": string(output),
+			"error":  true,
+		}, nil
+	}
+
+	return Params{
+		"result": string(output),
+		"error":  false,
+	}, nil
 }
--- a/fs/rc/internal.go
+++ b/fs/rc/internal.go
@@ -4,6 +4,7 @@ package rc
 
 import (
 	"context"
+	"net/http"
 	"os"
 	"os/exec"
 	"runtime"
@@ -339,10 +340,12 @@ func rcSetBlockProfileRate(ctx context.Context, in Params) (out Params, err erro
 
 func init() {
 	Add(Call{
-		Path:         "core/command",
-		AuthRequired: true,
-		Fn:           rcRunCommand,
-		Title:        "Run a rclone terminal command over rc.",
+		Path:          "core/command",
+		AuthRequired:  true,
+		Fn:            rcRunCommand,
+		NeedsRequest:  true,
+		NeedsResponse: true,
+		Title:         "Run a rclone terminal command over rc.",
 		Help: `This takes the following parameters
 
 - command - a string with the command name
@@ -353,6 +356,7 @@ Returns
 
 - result - result from the backend command
 - error	 - set if rclone exits with an error code
+- returnType - one of ("COMBINED_OUTPUT", "STREAM", "STREAM_ONLY_STDOUT". "STREAM_ONLY_STDERR")
 
 For example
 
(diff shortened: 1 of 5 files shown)


fs/rc/internal.go
```diff
@@ -335,3 +336,110 @@ func rcSetBlockProfileRate(ctx context.Context, in Params) (out Params, err erro
...
+		return nil, err
+	}
+
+	cmd := exec.CommandContext(ctx, ex, allArgs...)
+
+	// Run the command and get the output for error and stdout combined.
+	output, err := cmd.CombinedOutput()
+
+	if err != nil {
+		fs.Errorf(nil, "Command error %v", err)
```
ncw, line 434: This should return `err.Error()` probably. It might be nice to return the return code of the command too (which is an integer 0 being OK)

fs/rc/internal_test.go
```diff
@@ -119,3 +119,21 @@ func TestCoreQuit(t *testing.T) {
...
+func TestCoreCommand(t *testing.T) {
+	call := Calls.Get("core/command")
+
+	in := Params{
+		"command": "version",
+		"opt":     map[string]string{},
+		"arg":     []string{},
+	}
+	got, err := call.Fn(context.Background(), in)
+	require.NoError(t, err)
```
ncw, line 133: You could check that the current version `fs.Version` is contained in the output returned from rclone here.

fs/rc/internal.go
negative0: I will check that once and update it. We could give out a parameter which says "exitCode". And handle non-zero exit codes in UI.

fs/rc/internal_test.go
negative0: Thanks for this :-)

negative0:
@ncw, I think the tests are failing because, in the test environment, the rclone executable might be a test executable. What do you think might be the case?

negative0:
Executable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable: /tmp/go-build332533305/b986/rc.testExecutable

This is the output when printing the value of the executable. It just keeps calling the test executable again and again instead of calling the rclone executable.

negative0:
@ncw Do you think we should go something like `go build` and then use that executable?

ncw:
I've attempted to fix up the tests here.

The magic is in TestMain which pretends to be `rclone version` if it gets a version parameter.

Crude but effective!

```go
func TestMain(m *testing.M) {
	// Pretend to be rclone version if we have a version string parameter
	if os.Args[len(os.Args)-1] == "version" {
		fmt.Printf("rclone %s\n", fs.Version)
		os.Exit(0)
	}
	os.Exit(m.Run())
}
```

negative0:
That's an interesting approach. Thanks for that :-)

fs/rc/rcserver/rcserver.go
```diff
@@ -265,6 +265,10 @@ func (s *Server) handlePost(w http.ResponseWriter, r *http.Request, path string)
 		in["_request"] = r
 	}
 
+	if call.NeedsResponse {
+		in["_response"] = &w
```
ncw, line 269: The `http.ResponseWriter` is an interface so is safe to pass by value here so just

    in["_response"] = w

The signature of `GetHTTPResponseWriter` will need to change to

    func (p Params) GetHTTPResponseWriter() (*http.ResponseWriter, error) {

An interface in Go is already contains a pointer. Effectively it is a very small structure with two pointers, one to the actual data and one to a description of the data.

fs/rc/internal.go
```diff
@@ -427,19 +442,30 @@ func rcRunCommand(ctx context.Context, in Params) (out Params, err error) {
...
+			}, nil
+		}
 		return Params{
-			"result": string(output),
-			"error":  true,
+			"result": string(out),
+			"error":  false,
 		}, nil
+	} else if returnType == "STREAM_ONLY_STDOUT" {
+		cmd.Stdout = *httpResponse
```
ncw, line 461: So this writes direct to the http output which is cool!

Do we need to change the Content-Type for this?

negative0: I think we should set it to text/plain

ncw:
This appears to be failing the tests?

```
--- FAIL: TestCoreCommand (0.00s)
    internal_test.go:149: 
        	Error Trace:	internal_test.go:149
        	            	response object is required
        	            	expecting http.ResponseWriter value for key "_response" (was *http.ResponseWriter)
        	            	github.com/rclone/rclone/fs/rc.rcRunCommand
        	            		/home/runner/work/rclone/rclone/fs/rc/internal.go:413
        	            	github.com/rclone/rclone/fs/rc.TestCoreCommand
        	            		/home/runner/work/rclone/rclone/fs/rc/internal_test.go:148
        	            	testing.tRunner
        	            		/opt/hostedtoolcache/go/1.14.6/x64/src/testing/testing.go:1039
        	            	runtime.goexit
        	            		/opt/hostedtoolcache/go/1.14.6/x64/src/runtime/asm_amd64.s:1373
        	Test:       	TestCoreCommand
FAIL
```

buengese:
What exactly is the use case for this? Every functionality required for the WebUI should be exposed directly via the rc interface. I don't really see why anyone would wanna use a web based rclone terminal over running rclone in the local terminal or via ssh.

For anything other than a web based terminal this looks like an ugly hack.

ncw:
> What exactly is the use case for this? Every functionality required for the WebUI should be exposed directly via the rc interface. 

Every thing that rclone provides should be provided over the API.

Some of it is quite hard to use at the moment and the docs and forum all show `rclone xyz` commands.

There is a problem at the moment with global options which we haven't fixed yet (all the options are global)

> I don't really see why anyone would wanna use a web based rclone terminal over running rclone in the local terminal or via ssh.

A reasonable point. It is really a demo of the new plugin system. It isn't much code and I don't think it opens any security vulnerabilities (which was my main concern).

> For anything other than a web based terminal this looks like an ugly hack.

Yes that is what it is for. And it is an ugly hack!

Do you think we should do this a different way?

buengese:
> There is a problem at the moment with global options which we haven't fixed yet (all the options are global)
I've noticed this too is this something you are planning to work on soon?

> Do you think we should do this a different way?
I agree that this is not a security issue and I guess it's good enough for now. In the long term I'd kinda prefer a web console to call the relevant rc commands directly (e.g. running rclone sync in the webui means the webui parses the command and directly calls rc/sync). Though this would be a lot of work and could potentially cause problems with feature parity with the normal command line so it's not ideal either.

ncw:
> > There is a problem at the moment with global options which we haven't fixed yet (all the options are global)
> 
> I've noticed this too is this something you are planning to work on soon?

I think what we need to do is attach an fs.Config to the context, then write a little shim `fs.GetConfig(ctx)` which attempts to read the config from the `ctx` and if it isn't found return the global ctx.

The `fs.Config` could be created in the rc by passing in an `_options` struct which could take the same format as returned by `options/get`. There is actually more than one global config struct we might want to set, eg the Filters.

I'd quite like to make a MarshalJSON method for each of the config flags so they could take strings too so you could set `_options: { "main": { "ConnectTimeout": "5m" } }` rather than using `30000000000` nano seconds.

However we have a way of serializing the options already - the command line - so if we could pass in a command line string of options and parse that into config that would probably be much easier to use, so pass in `_flags="--contimeout 5m"`. It would need a bit of rejiggling of internals to make that work, but maybe that is something worth a bit of effort - it would also enable the `core/command` API to work in a much cleaner way.

What do you think?
 
> > Do you think we should do this a different way?
> 
> I agree that this is not a security issue and I guess it's good enough for now. In the long term I'd kinda prefer a web console to call the relevant rc commands directly (e.g. running rclone sync in the webui means the webui parses the command and directly calls rc/sync). Though this would be a lot of work and could potentially cause problems with feature parity with the normal command line so it's not ideal either.

I think if we can use the actual command line parser then it would be a great feature (see above).

buengese:
Sounds good to me.
> I'd quite like to make a MarshalJSON method for each of the config flags so they could take strings too so you could set _options: { "main": { "ConnectTimeout": "5m" } } rather than using 30000000000 nano seconds.
This sounds like a bit of work but I quite like the idea given that it improves readability of the options quite a lot.

I'd also already thought of reusing thought of reusing the command line parser but hadn't looked into how that would be done. If that's something that can be done relatively easy than it's definitely the best option.

negative0:
> Sounds good to me.
> 
> > I'd quite like to make a MarshalJSON method for each of the config flags so they could take strings too so you could set _options: { "main": { "ConnectTimeout": "5m" } } rather than using 30000000000 nano seconds.
> > This sounds like a bit of work but I quite like the idea given that it improves readability of the options quite a lot.
> 
> I'd also already thought of reusing thought of reusing the command line parser but hadn't looked into how that would be done. If that's something that can be done relatively easy than it's definitely the best option.

That would be really nice

negative0:
Thank you!

Issue #4366, opened 2020-06-17
Bring back old elapsed time behavior


Dibbyo456:
In rclone **v1.50.2** (could be wrong) the elapsed time behavior has changed. According to nick on [#3860](https://github.com/rclone/rclone/issues/3860#issuecomment-572502504) 

> What `Elapsed time` means now is time that files are copying for and since you haven't copied any files yet that is why it is nearly zero. We made it like that so we could compute average transfer rates properly.

I asked about it on forum and [he wanted to bring back old behavior](https://forum.rclone.org/t/elapsed-time-stays-0-until-transfers-starts/15201/2?u=harry). According to him:

> The time displaying as elapsed time should really be called Transfer time - it is used to calculate the MB/s rates. I'd like to bring back elapsed time as it was!

Why not both Elapsed time & Transfer time! 😗

ncw:
Yes, let's do this!

So relabel `Elapsed time` to `Transfer time` and don't show if it is zero

Re-implement elapsed time.

Want to have a go at this @Dibbyo456 ?

Dibbyo456:
Fixed by: #4448

ncw:
Oops we forgot to link the change - thanks @Dibbyo456 for pointing that out.

I've merged this to master now which means it is in [the latest beta](https://beta.rclone.org) and released in v1.53 - thanks @negative0

PR #4448, opened 2020-07-19
fix elapsed time does not show actual time since beginning and show t…

negative0:
…ransfer time as time when jobs were running.

The elapsed time in core/stats does not show the actual elapsed time, instead, shows the job execution time. Show the job transfer times as transfer time and time since the start of rclone as elapsedTime.

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -251,9 +251,10 @@ func (s *StatsInfo) String() string {
 
 	s.mu.RLock()
 
-	dt := time.Since(startTime)
+	elapsedTime := time.Since(startTime)
+	elapsedTimeSecondsOnly := elapsedTime.Truncate(time.Second/10) % time.Minute
+	dt := s.totalDuration()
 	dtSeconds := dt.Seconds()
-	dtSecondsOnly := dt.Truncate(time.Second/10) % time.Minute
 	speed := 0.0
 	if dt > 0 {
 		speed = float64(s.bytes) / dtSeconds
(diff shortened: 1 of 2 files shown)


fs/accounting/stats.go
```diff
@@ -66,7 +68,8 @@ func (s *StatsInfo) RemoteStats() (out rc.Params, err error) {
 	out["transfers"] = s.transfers
 	out["deletes"] = s.deletes
 	out["renames"] = s.renames
-	out["elapsedTime"] = s.totalDuration().Seconds()
+	out["transferTime"] = s.totalDuration().Seconds()
+	out["elapsedTime"] = time.Now().Sub(startTime).Seconds()
```
ncw, line 72: This is usually written `time.Since(startTime)` instead of `time.Now().Sub(startTime)` - it means the same thing though!

negative0: That's pretty neat.

negative0:
> Nice fix! See inline for minor nit pick.
> 
> Can you fix this too? This is rclone's running stats display.
> 
> https://github.com/rclone/rclone/blob/d1617ce7ceb0525019075254ecff0658fe4ed403/fs/accounting/stats.go#L335

Done. please verify once.

fs/accounting/stats.go
```diff
@@ -248,7 +251,7 @@ func (s *StatsInfo) String() string {
 
 	s.mu.RLock()
 
-	dt := s.totalDuration()
+	dt := time.Since(startTime)
```
ncw, line 254: We need this dt to be `s.totalDuration()` as it is used for the calculation of the overall speed.

So we we probably need a `elapsedTime := time.Since(startTime)` for use in the stats logging.

negative0:
Thanks a lot. :t-rex:

Dibbyo456:
Does this fix #4366 ?

Dibbyo456:
@negative0 Awesome work. 👍 But the elapsed time updates in milliseconds, which is more frequent than its needs to be. The preferred would be update every second.



I can send a PR, if anyone could tell me what changes needs to be done.

ncw:
@Dibbyo456 has the format changed after this PR? I can't remember! I find the ms useful when doing performance testing.

PR #4460, opened 2020-07-25
Expose NewServer() in cmd/serve as http.RoundTripper so that rclone can be used as a library

jdeng:
To provide an option for developers that they can build restic + rclone as a library. The idea is to expose an http.RoundTripper interface via NewServer(), so that in restic the http request can be handled via a function call instead of spawning a rclone process, thus rclone code can run within the restic process. Further more restic can be compiled as a library too, so that developers can build apps to browse encrypted files stored in the cloud via restic + rclone.

The change is probably not elegant but it is minimal and I hope this can be upstreamed.

No.  A fork of restic that use this interface can be found here: https://github.com/jdeng/restic
A tag rclone needs to be specified when building it: `go run build.go -t rclone`.  To run it with `./restic -r inproc:rclone:gdrive:xxx snapshots` and you won't see a rclone process.

The change is pretty simple and mainly in this directory: https://github.com/jdeng/restic/tree/master/internal/backend/inproc

Basically it is reusing the rest backend but replace the transport with rclone's NewServer(). See below snippet:

```
	restBackend, err := rest.Open(restConfig, debug.RoundTripper(be.tr)) // be.tr is rclone's NewServer()
	if err != nil {
		return nil, err
	}

	be.Backend = restBackend
```

The commit history is a bit messy. Please just squash it.

#### Checklist


cmd/serve/restic/inproc.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/serve/httplib/httpflags"
+)
+
+// NewServer returns a rclone server object which talks the restic http protocol
+func NewServer(args []string) (http.RoundTripper, error) {
```
ncw, line 12: It would be best to return the specific type rather than the interface `http.RoundTripper` here - it leads to more flexibility later.

cmd/serve/restic/inproc.go
```diff
@@ -0,0 +1,21 @@
...
+	f := cmd.NewFsSrc(args)
+	return newServer(f, &httpflags.Opt), nil
+}
+
+func (s *server) RoundTrip(r *http.Request) (*http.Response, error) {
+	w := httptest.NewRecorder()
```
ncw, line 18: I'm not sure we should have any `httptest` stuff here?

cmd/serve/restic/inproc.go
```diff
@@ -0,0 +1,21 @@
...
+	"net/http/httptest"
+
+	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/serve/httplib/httpflags"
+)
+
+// NewServer returns a rclone server object which talks the restic http protocol
+func NewServer(args []string) (http.RoundTripper, error) {
+	f := cmd.NewFsSrc(args)
+	return newServer(f, &httpflags.Opt), nil
```
ncw, line 14: I think since `cmd.NewFsSrc` is public I'd probably just make `newServer` public and then you have all the bits you need.

cmd/serve/restic/inproc.go
jdeng: golint fails if returned objects is of unexported types. So this is a workaround unless the server type is exported which would require more changes and I am not sure whether it is worthwhile.

cmd/serve/restic/inproc.go
jdeng: Yes making newServer public would be helpful. But you may have to make server exported to keep golint quiet.

jdeng:
Just expose the server structure and the NewServer() function in the new change. Seems quite straightforward. Thanks!

jdeng:
Thanks!

ncw:
So is your plan to make a restic binary with rclone built in? Are you going to upstream this to restic?

I did have the opposite idea of making a restic sub command for rclone...

jdeng:
My use case is focused on reading using restic + rclone as a library so that I can access the data from mobile. I tried using the combined restic + rclone binary for backup and it seems the performance is not as good as simply uploading the local restic repository to cloud using rclone. And the files generated with the two approach seem different (but I haven't spent time to find out why). Reading specific files from a repo is less sensitive to performance so convenience is more important.

I tried to upstream my changes but they weren't accepted. For now I am maintaining my fork which isn't too difficult anyway. I do feel the asymmetric ways of using restic + rclone (backuping using two steps and retrieving using a combined binary) is a good use case.

PR #4466, opened 2020-07-29
cmd/obscure: Allow obscure command to accept password on STDIN

dtibarra:

`rclone obscure` currently only accepts a command line argument of `password` to generate
an obfuscated password. This is an issue since generating obfuscated passwords programatically
requires sending the plain text password as a shell argument, which can cause problems if the
password contains shell characters, or if the password is from an untrusted source.

This patch opens up STDIN which will allow developers to open the STDIN source and print a password
directly to `rclone obscure`, which can increase safety and convenince.

Unknown. A cursory glance did not turn up any forums posts.

#### Checklist

### DIFF
--- a/cmd/obscure/obscure.go
+++ b/cmd/obscure/obscure.go
@@ -3,6 +3,9 @@ package obscure
 import (
 	"fmt"
 
+	"io/ioutil"
+	"os"
+
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs/config/obscure"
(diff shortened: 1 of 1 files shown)


dtibarra:
I made the requested changes! I chose to make the combination of a hyphen as `arg[0]` AND the presence of data on `STDIN` be the trigger to accept input from `STDIN`. This way, we don't end up accidentally becoming unusable for someone who wants to obfuscate a literal hyphen. Let me know if this isn't acceptable and I can do it differently for you.

PR #4482, opened 2020-08-03
Vfs reset cache

leoluan1:

This patch detects out-of-space errors of the full cache mode and resets clean cache files (not changed) to make space.

https://forum.rclone.org/t/removing-recreating-cache-file-upon-enospc/17791

#### Checklist

### DIFF
--- a/vfs/read_write.go
+++ b/vfs/read_write.go
@@ -260,7 +260,9 @@ func (fh *RWFileHandle) _readAt(b []byte, off int64, release bool) (n int, err e
 		// Do the writing with fh.mu unlocked
 		fh.mu.Unlock()
 	}
+
 	n, err = fh.item.ReadAt(b, off)
+
 	if release {
 		fh.mu.Lock()
 	}
--- a/vfs/vfscache/cache.go
+++ b/vfs/vfscache/cache.go
@@ -17,6 +17,7 @@ import (
 	"github.com/rclone/rclone/fs"
 	fscache "github.com/rclone/rclone/fs/cache"
 	"github.com/rclone/rclone/fs/config"
+	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/rclone/rclone/lib/file"
@@ -47,9 +48,16 @@ type Cache struct {
 	writeback  *writeback.WriteBack // holds Items for writeback
 	avFn       AddVirtualFn         // if set, can be called to add dir entries
 
-	mu   sync.Mutex       // protects the following variables
-	item map[string]*Item // files/directories in the cache
-	used int64            // total size of files in the cache
+	mu            sync.Mutex       // protects the following variables
+	cond          *sync.Cond       // cond lock for synchronous cache cleaning
+	item          map[string]*Item // files/directories in the cache
+	errItems      map[string]error // items in error state
+	used          int64            // total size of files in the cache
+	outOfSpace    bool             // out of space
+	cleanerKicked bool             // some thread kicked the cleaner upon out of space
+	kickerMu      sync.Mutex       // mutex for clearnerKicked
+	kick          chan struct{}    // channel for kicking clear to start
+
 }
 
 // AddVirtualFn if registered by the WithAddVirtual method, can be
@@ -96,6 +104,7 @@ func New(ctx context.Context, fremote fs.Fs, opt *vfscommon.Options, avFn AddVir
 		root:       root,
 		metaRoot:   metaRoot,
 		item:       make(map[string]*Item),
+		errItems:   make(map[string]error),
 		hashType:   hashType,
 		hashOption: hashOption,
 		writeback:  writeback.New(ctx, opt),
@@ -116,6 +125,10 @@ func New(ctx context.Context, fremote fs.Fs, opt *vfscommon.Options, avFn AddVir
 
 	// Remove any empty directories
 	c.purgeEmptyDirs()
+
+	// Create a channel for cleaner to be kicked upon out of space con
+	c.kick = make(chan struct{}, 1)
+	c.cond = sync.NewCond(&c.mu)
 
 	go c.cleaner(ctx)
 
@@ -401,28 +414,125 @@ func (c *Cache) reload(ctx context.Context) error {
 	return nil
 }
 
-// purgeOld gets rid of any files that are over age
-func (c *Cache) purgeOld(maxAge time.Duration) {
-	c._purgeOld(maxAge, func(item *Item) {
-		item.remove("too old")
-	})
+// KickCleaner kicks cache cleaner upon out of space situation
+func (c *Cache) KickCleaner() {
+	/* Use a separate kicker mutex for the kick to go through without waiting for the
+	   cache mutex to avoid letting a thread kick again after the clearer just
+	   finished cleaning and unlock the cache mutex. */
+	fs.Debugf(nil, "vfs cache: at the beginning of KickCleaner")
+	c.kickerMu.Lock()
+	if !c.cleanerKicked {
+		c.cleanerKicked = true
+		fs.Debugf(nil, "vfs cache: in KickCleaner, ready to lock cache mutex")
+		c.mu.Lock()
+		c.outOfSpace = true
+		fs.Logf(nil, "vfs cache: in KickCleaner, ready to kick cleaner")
+		c.kick <- struct{}{}
+		c.mu.Unlock()
+	}
+	c.kickerMu.Unlock()
+
+	c.mu.Lock()
+	for c.outOfSpace == true {
+		fs.Debugf(nil, "vfs cache: in KickCleaner, looping on c.outOfSpace")
+		c.cond.Wait()
+	}
+	fs.Debugf(nil, "vfs cache: in KickCleaner, leaving c.outOfSpace loop")
+	c.mu.Unlock()
 }
 
-func (c *Cache) _purgeOld(maxAge time.Duration, remove func(item *Item)) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	cutoff := time.Now().Add(-maxAge)
-	for name, item := range c.item {
-		if !item.inUse() {
-			// If not locked and access time too long ago - delete the file
-			dt := item.getATime().Sub(cutoff)
-			// fs.Debugf(name, "atime=%v cutoff=%v, dt=%v", item.info.ATime, cutoff, dt)
-			if dt < 0 {
-				remove(item)
-				// Remove the entry
-				delete(c.item, name)
+// removeNotInUse removes items not in use with a possible maxAge cutoff
+// called with cache mutex locked and up-to-date c.used (as we update it directly here)
+func (c *Cache) removeNotInUse(item *Item, maxAge time.Duration, emptyOnly bool) {
+	removed, spaceFreed := item.RemoveNotInUse(maxAge, emptyOnly)
+	// The item space might be freed even if we get an error after the cache file is removed
+	// The item will not be removed or reset the cache data is dirty (DataDirty)
+	c.used -= spaceFreed
+	if removed {
+		fs.Infof(nil, "vfs cache RemoveNotInUse (maxAge=%d, emptyOnly=%v): item %s was removed, freed %d bytes", maxAge, emptyOnly, item.GetName(), spaceFreed)
+		// Remove the entry
+		delete(c.item, item.name)
+	} else {
+		fs.Infof(nil, "vfs cache RemoveNotInUse (maxAge=%d, emptyOnly=%v): item %s not removed, freed %d bytes", maxAge, emptyOnly, item.GetName(), spaceFreed)
+	}
+	return
+}
+
+// Retry failed resets during purgeClean()
+func (c *Cache) retryFailedResets() {
+	// Some items may have failed to reset becasue there was not enough space
+	// for saving the cache item's metadata.  Redo the Reset()'s here now that
+	// we may have some available space.
+	if len(c.errItems) != 0 {
+		fs.Debugf(nil, "vfs cache reset: before redoing reset errItems = %v", c.errItems)
+		for itemName := range c.errItems {
+			_, _, err := c.item[itemName].Reset()
+			if err == nil || !fserrors.IsErrNoSpace(err) {
+				// TODO: not trying to handle non-ENOSPC errors yet
+				delete(c.errItems, itemName)
 			}
 		}
+		fs.Debugf(nil, "vfs cache reset: after redoing reset errItems = %v", c.errItems)
+	}
+}
+
+func (c *Cache) purgeClean(quota int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var items Items
+
+	if quota <= 0 || c.used < quota {
+		return
+	}
+
+	// Make a slice of clean cache files
+	for _, item := range c.item {
+		if !item.IsDataDirty() {
+			items = append(items, item)
+		}
+	}
+
+	sort.Sort(items)
+
+	// Reset items until the quota is OK
+	for _, item := range items {
+		if c.used < quota {
+			break
+		}
+		resetResult, spaceFreed, err := item.Reset()
+		// The item space might be freed even if we get an error after the cache file is removed
+		// The item will not be removed or reset if the cache data is dirty (DataDirty)
+		c.used -= spaceFreed
+		fs.Infof(nil, "vfs cache purgeClean item.Reset %s: %s, freed %d bytes", item.GetName(), resetResult.String(), spaceFreed)
+		if resetResult == RemovedNotInUse {
+			delete(c.item, item.name)
+		}
+		if err != nil {
+			fs.Errorf(nil, "vfs cache purgeClean item.Reset %s reset failed, err = %v, freed %d bytes", item.GetName(), err, spaceFreed)
+			c.errItems[item.name] = err
+		}
+	}
+
+	// Resest outOfSpace without checking whether we have reduced cache space below the quota.
+	// This allows some files to reduce their pendingAccesses count to allow them to be reset
+	// in the next iteration of the purge cleaner loop.
+
+	c.outOfSpace = false
+	c.cond.Broadcast()
+}
+
+// purgeOld gets rid of any files that are over age
+func (c *Cache) purgeOld(maxAge time.Duration) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	// cutoff := time.Now().Add(-maxAge)
+	for _, item := range c.item {
+		c.removeNotInUse(item, maxAge, false)
+	}
+	if c.used < int64(c.opt.CacheMaxSize) {
+		c.outOfSpace = false
+		c.cond.Broadcast()
 	}
 }
 
@@ -439,16 +549,8 @@ func (c *Cache) purgeEmptyDirs() {
 	}
 }
 
-// Remove any files that are over quota starting from the
-// oldest first
-func (c *Cache) purgeOverQuota(quota int64) {
-	c._purgeOverQuota(quota, func(item *Item) {
-		item.remove("over quota")
-	})
-}
-
 // updateUsed updates c.used so it is accurate
-func (c *Cache) updateUsed() {
+func (c *Cache) updateUsed() (used int64) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
@@ -457,15 +559,19 @@ func (c *Cache) updateUsed() {
 		newUsed += item.getDiskSize()
 	}
 	c.used = newUsed
+	return newUsed
 }
 
-func (c *Cache) _purgeOverQuota(quota int64, remove func(item *Item)) {
+// Remove clean cache files that are not open until the total space
+// is reduced below quota starting from the oldest first
+func (c *Cache) purgeOverQuota(quota int64) {
 	c.updateUsed()
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
 	if quota <= 0 || c.used < quota {
+
 		return
 	}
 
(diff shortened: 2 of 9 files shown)


vfs/vfscache/cache.go
```diff
@@ -526,23 +598,27 @@ func (c *Cache) clean() {
 	fs.Infof(nil, "vfs cache: cleaned: objects %d (was %d) in use %d, to upload %d, uploading %d, total size %v (was %v)", newItems, oldItems, totalInUse, uploadsQueued, uploadsInProgress, newUsed, oldUsed)
 }
 
-// cleaner calls clean at regular intervals
+// cleaner calls clean at regular intervals and upon being kicked for out-of-space condition
 //
 // doesn't return until context is cancelled
 func (c *Cache) cleaner(ctx context.Context) {
+	c.kick = make(chan struct{}, 1)
```
ncw, line 605: I would initialise both of these in the main cache initialization `New` so we know they are valid immediately as the `go c.cleaner()` could conceivably run after something uses `c.kick` or `c.cond`

vfs/vfscache/downloaders/downloaders.go
```diff
@@ -386,6 +389,11 @@ func (dls *Downloaders) kickWaiters() (err error) {
 			fs.Errorf(dls.src, "vfs cache: restart download failed: %v", err)
 		}
 	}
+	if fserrors.IsErrNoSpace(dls.lastErr) {
+		fs.Errorf(dls.src, "vfs cache: cache is out of space %d/%d: last error: %v", dls.errorCount, maxErrorCount, dls.lastErr)
+		dls._closeWaiters(dls.lastErr)
```
ncw, line 394: This returns the error to the waiters - I thought you wanted to avoid that?

vfs/vfscache/downloaders/downloaders.go
```diff
@@ -576,9 +584,15 @@ func (dl *downloader) stopAndClose(inErr error) (err error) {
 func (dl *downloader) download() (n int64, err error) {
 	// defer log.Trace(dl.dls.src, "")("err=%v", &err)
 	n, err = dl.in.WriteTo(dl)
-	if err != nil && errors.Cause(err) != asyncreader.ErrorStreamAbandoned {
-		return n, errors.Wrap(err, "vfs reader: failed to write to cache file")
+	if err != nil {
+		if fserrors.IsErrNoSpace(err) {
```
ncw, line 588: I'm not sure you need this bit of code at all - `IsErrNoSpace()` should be able to unwrap the error just fine so I don't see the purpose?

vfs/vfscache/item.go
```diff
@@ -399,9 +402,37 @@ func (item *Item) IsDirty() bool {
...
+func (item *Item) IsDataDirty() bool {
+	item.mu.Lock()
+	defer item.mu.Unlock()
+	return item.info.Dirty
+}
+
+// Open the local file from the object passed in.  Wraps open()
+// to provide recovery from out of space error.
+func (item *Item) Open(o fs.Object) (err error) {
+	for {
```
ncw, line 415: This probably shouldn't retry forever...

vfs/vfscache/item.go
```diff
@@ -429,6 +460,7 @@ func (item *Item) Open(o fs.Object) (err error) {
 
 	fd, err := file.OpenFile(osPath, os.O_RDWR, 0600)
 	if err != nil {
+		item.opens--
```
ncw, line 463: None of the other error handling does `item.opens--` so this sticks out. Maybe they all should, I'm not sure...

vfs/vfscache/item.go
```diff
@@ -439,7 +471,14 @@ func (item *Item) Open(o fs.Object) (err error) {
 
 	err = item._save()
 	if err != nil {
-		return err
+		closeErr := item.fd.Close()
+		if closeErr != nil {
+			closeErr = fs.CountError(closeErr)
+			log.Fatal(closeErr)
```
ncw, line 477: You don't want `log.Fatal` this kills the program! Just `fs.Errorf` the error

vfs/vfscache/item.go
```diff
@@ -767,6 +806,165 @@ func (item *Item) remove(reason string) (wasWriting bool) {
...
+			fs.Errorf(item.o, "vfs cache: item reset failed: %v", e)
+			if err == nil {
+				err = e
+			}
+		}
+	}
+
+	if downloaders := item.downloaders; downloaders != nil {
+		item.downloaders = nil
+		// FIXME need to unlock to kill downloader - should we
```
ncw, line 866: Mmm, possibly.  The locking is super tricky here. It is really easy to cause a deadlock so if we want to do that then it should be a different commit.

vfs/vfscache/item.go
```diff
@@ -767,6 +806,165 @@ func (item *Item) remove(reason string) (wasWriting bool) {
...
+
+	// item.fd should be nil at this time
+	if item.fd != nil {
+		checkErr(errors.New("vfs cache item: internal error: didn't Close file"))
+		item.fd = nil
+		return
+	}
+	item.modified = false
+
+	fd, err := file.OpenFile(osPath, os.O_RDWR, 0600)
```
ncw, line 915: This code is duplicated from `Open` - we should probably factor it out of there rather than repeat it...

vfs/read_write.go
```diff
@@ -260,7 +261,21 @@ func (fh *RWFileHandle) _readAt(b []byte, off int64, release bool) (n int, err e
 		// Do the writing with fh.mu unlocked
 		fh.mu.Unlock()
 	}
-	n, err = fh.item.ReadAt(b, off)
+	for {
+		fh.item.ProtectCache()
```
ncw, line 265: I'm not a big fan of the `ProtectCache` calls - they are an implementation detail for the cache, so can we push them into the item.ReadAt call and any others which need it?

vfs/read_write.go
```diff
@@ -260,7 +261,21 @@ func (fh *RWFileHandle) _readAt(b []byte, off int64, release bool) (n int, err e
...
+		n, err = fh.item.ReadAt(b, off)
+		fh.item.UnprotectCache()
+		if err == nil {
+			break
+		}
+		log.Trace(fh.logPrefix(), "err=%v err.Error()=%v", &err, err.Error())
+		if !fserrors.IsErrNoSpace(err) && err.Error() != "no space left on device" {
```
ncw, line 272: Likewise this logic should be in Item I think, then you could get rid of `item.GetCache`

vfs/vfscache/item.go
```diff
@@ -811,6 +1009,10 @@ func (item *Item) _ensure(offset, size int64) (err error) {
 	}
 	r := ranges.Range{Pos: offset, Size: size}
 	present := item.info.Rs.Present(r)
+	/* This statement simulates a cache space error for test purpose */
```
ncw, line 1012: This could maybe be a private cache option that we can set during tests

vfs/vfscache/item_test.go
```diff
@@ -173,7 +173,7 @@ func TestItemTruncateExisting(t *testing.T) {
 	require.Error(t, item.Truncate(40))
 	checkObject(t, r, "existing", contents)
 
-	require.NoError(t, item.Open(obj))
```
ncw, line 176: I'm not sure the purpose of this change - I would have thought the test should work without it.

vfs/vfscache/cache.go
```diff
@@ -511,6 +571,18 @@ func (c *Cache) clean() {
 	// oldest first
 	c.purgeOverQuota(int64(c.opt.CacheMaxSize))
 
+	// removeCleanFiles indicates that we got ENOSPC error
```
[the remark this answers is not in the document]
leoluan1: In our use case, the cache size will be relatively small and it will be filled quickly, like in a few seconds.  So we probably want delete all clean files so that the cleaner does not get kicked even more quickly to repeat the cache reset process.  I think we can achieve this by setting a very low quota in our use case and allow other use cases to reset just enough clean cache files to make the total cache space usage to be below the quota.  I will add the quota check in purgeClean().

vfs/vfscache/cache.go
```diff
@@ -401,6 +407,60 @@ func (c *Cache) reload(ctx context.Context) error {
...
+	c.mu.Lock()
+	for c.outOfSpace == true {
+		fs.Debugf(nil, "vfs cache: in KickCleaner, looping on c.outOfSpace")
+		c.cond.Wait()
+	}
+	fs.Debugf(nil, "vfs cache: in KickCleaner, leaving c.outOfSpace loop")
+	c.mu.Unlock()
+}
+
+func (c *Cache) purgeClean() {
```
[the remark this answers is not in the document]
leoluan1: Can we add this in a subsequent pull request?  I hope to be able to catch the current release.  So minimizing the changes to an essential patch for this pull request might help?

vfs/vfscache/cache.go
leoluan1: Makes sense!  Will do.

vfs/vfscache/downloaders/downloaders.go
leoluan1: Don't we need the ENOSPC error to be passed back to the waiters so that the item._ensure() can get the error to start cache reset?

vfs/read_write.go
leoluan1: Agreed.  I also noticed that and will move them into item.ReadAt.

vfs/read_write.go
leoluan1: Agreed!

vfs/vfscache/item.go
leoluan1: Use a maximum retry count of 20 for this commit and potentially add a timeout in the future?

vfs/vfscache/downloaders/downloaders.go
leoluan1: The original code does not return non-nil error (or the ENOSPC error that we are considering currently) unless errors.Cause(err) != asyncreader.ErrorStreamAbandoned.
```
if err != nil && errors.Cause(err) != asyncreader.ErrorStreamAbandoned {
    return n, errors.Wrap(err, "vfs reader: failed to write to cache file")
}
return n, nil
```
Maybe that's bug?  Or I am missing something.
Should the line be 
```
return n, err
```

vfs/vfscache/item.go
leoluan1: How about factoring the following line out from item.open() and into a new item._newCacheFile() function and call it from both item.open() and item.Reset()?

```
	item.modified = false

	fd, err := file.OpenFile(osPath, os.O_RDWR, 0600)
	if err != nil {
		item.opens--
		return errors.Wrap(err, "vfs cache item: open failed")
	}
	err = file.SetSparse(fd)
	if err != nil {
		fs.Debugf(item.name, "vfs cache: failed to set as a sparse file: %v", err)
	}
	item.fd = fd

	err = item._save()
	if err != nil {
		closeErr := item.fd.Close()
		if closeErr != nil {
			closeErr = fs.CountError(closeErr)
			log.Fatal(closeErr)
		}
		item.fd = nil
		item.opens--
		return errors.Wrap(err, "vfs cache item: _save failed")
	}
```

vfs/vfscache/item.go
leoluan1: Sound a great idea! Could you help adding this cache option and the test triggers while I get the function implementation done to get it tested for the current beta?  I would like to do it myself, but it would probably take longer and affect the timeliness of the feature.

vfs/vfscache/item_test.go
leoluan1: As discussed in our PMs, this is due to some incorrect revert of some lines of an earlier commit.  Will fix.

vfs/vfscache/item.go
leoluan1: I will check all errors under item.Open, item.Close, and item.Reset to see if they adjust item.opens correctly and discuss with you upon questions.

vfs/vfscache/item.go
leoluan1: This was not the intention.  Will fix it accordingly.

vfs/vfscache/item.go
leoluan1: Are you referring to the comment in Line 866?  This is actually your comment that got copied from item.close since we are reusing a part of the item.close code.  I do not try to address this comment now. :)

vfs/vfscache/downloaders/downloaders.go
ncw: That makes sense - ignore the above then :-)

vfs/vfscache/downloaders/downloaders.go
ncw: The original code says in English, return an error unless it is ErrorStreamAbandoned in which case don't.

Your new code does exactly the same thing. However I think the original code would work just fine if `err == ENOSPC` then it would be returned in this line

    return n, errors.Wrap(err, "vfs reader: failed to write to cache file")

So I think the major difference between your code and the original is that the ENOSPC error doesn't get wrapped in the `errors.Wrap()` call. I don't think that matters since we are unwrapping the errors to find the ENOSPC anyway so one more layer of wrap doesn't make any difference.

So I think your change works fine, it just isn't needed.

leoluan1: You are right!  I miss-read the second part of the if condition as "==" (instead of !=).
if err != nil && errors.Cause(err) **==** asyncreader.ErrorStreamAbandoned {

vfs/vfscache/cache.go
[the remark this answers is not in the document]
ncw: A subsequent PR is fine

vfs/vfscache/item.go
ncw: Probably `fs.Config.LowLevelRetries` is the right number (which defaults to 10)

vfs/vfscache/item.go
ncw: Sounds good :-)

vfs/vfscache/item.go
ncw: OK carry on with the PR and at some point I'll merge into a branch we can both work on.

There is bound to be some stuff I'd like to add and that is quite a good way of working.

ncw:
Let me know when you are ready for me to take another look.

leoluan1:
> Let me know when you are ready for me to take another look.

I am mostly done with this round.  There is one area where I am still searching for an approach.   For the cache files that are still dirty after being closed (i.e., they are during the asynchronous write-back, I guess), what can we do about them?  Maybe we should skip the ones with dirty data (item.info.Dirty) and force removing the (item.metaDirty) ones with dirty cache info (item.info)?

vfs/vfscache/item.go
leoluan1: > OK carry on with the PR and at some point I'll merge into a branch we can both work on.
> 
> There is bound to be some stuff I'd like to add and that is quite a good way of working.

I think this is a good idea!

ncw:
> I am mostly done with this round. There is one area where I am still searching for an approach. For the cache files that are still dirty after being closed (i.e., they are during the asynchronous write-back, I guess), what can we do about them in the item.Reset(). Maybe we should skip the ones with dirty data (item.info.Dirty) and force removing the (item.metaDirty) ones with dirty cache info (item.info)?

Cache files that are still dirty we shouldn't be touching as if we delete them we'll cause data loss.

Dirty metadata isn't so much of a problem as it is usually quite small. It does take up disk space though and if the disk is full then we can't save it. If the metadata is dirty and we want to remove the file then that won't cause data loss so is no big deal.

I think you'll need to rebase this on master to get the build passing again as go1.15rc2 appears to have vanished...

leoluan1:
> > I am mostly done with this round. There is one area where I am still searching for an approach. For the cache files that are still dirty after being closed (i.e., they are during the asynchronous write-back, I guess), what can we do about them in the item.Reset(). Maybe we should skip the ones with dirty data (item.info.Dirty) and force removing the (item.metaDirty) ones with dirty cache info (item.info)?
> 
> Cache files that are still dirty we shouldn't be touching as if we delete them we'll cause data loss.

I am not touching the cache files with dirty data.

> Dirty metadata isn't so much of a problem as it is usually quite small. It does take up disk space though and if the disk is full then we can't save it. If the metadata is dirty and we want to remove the file then that won't cause data loss so is no big deal.

In my tests I did run into the cases where there is not space for creating a new metadata file after the original cache data and metadata files are removed.  Now I put them on a new errItems field as a part of the cache struct and retry resetting these items after more space if made available.  It's being tested now.  The hope is that we don't need to fail the IOs that already have the file open even after item.Reset fails due to the lack of the small space to save the item's metadata file.

> I think you'll need to rebase this on master to get the build passing again as go1.15rc2 appears to have vanished...

Will do.  Thanks for the heads up.

The refactoring changes in cache.go cause some cache_test.go cases to fail because they need to change as well.   I wonder whether we could work together on those after our branch merge (assuming it's still the plan)?

leoluan1:
> > I am mostly done with this round. There is one area where I am still searching for an approach. For the cache files that are still dirty after being closed (i.e., they are during the asynchronous write-back, I guess), what can we do about them in the item.Reset(). Maybe we should skip the ones with dirty data (item.info.Dirty) and force removing the (item.metaDirty) ones with dirty cache info (item.info)?
> 
> Cache files that are still dirty we shouldn't be touching as if we delete them we'll cause data loss.
> 
> Dirty metadata isn't so much of a problem as it is usually quite small. It does take up disk space though and if the disk is full then we can't save it. If the metadata is dirty and we want to remove the file then that won't cause data loss so is no big deal.
> 
> I think you'll need to rebase this on master to get the build passing again as go1.15rc2 appears to have vanished...

What git steps should I take here?   How does go1.15rc2 impact here?  I am lost here. :)

darthShadow:
> What git steps should I take here?

Update the master branch in your fork and rebase from it to your private branch via `git rebase -p master`. Make sure you have a diff of the changes backed up somewhere in case something goes wrong.

> How does go1.15rc2 impact here?

Currently, since your branch was from an earlier commit in the current master, it's using go1.15rc2 for the build pipeline which is no longer available since 1.15 has been released publicly. Hence the rebase, which will update the pipeline to use 1.15x (the latest 1.15 version available).

ncw:
The build is failing because of these errors

```
vfs/vfscache/cache_test.go:289:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
vfs/vfscache/cache_test.go:305:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
vfs/vfscache/cache_test.go:313:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
vfs/vfscache/cache_test.go:323:16: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
vfs/vfscache/cache_test.go:331:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
```

So the existing tests in `vfscache` need fixing.

leoluan1:
> > What git steps should I take here?
> 
> Update the master branch in your fork and rebase from it to your private branch via `git rebase -p master`. Make sure you have a diff of the changes backed up somewhere in case something goes wrong.
> 
> > How does go1.15rc2 impact here?
> 
> Currently, since your branch was from an earlier commit in the current master, it's using go1.15rc2 for the build pipeline which is no longer available since 1.15 has been released publicly. Hence the rebase, which will update the pipeline to use 1.15x (the latest 1.15 version available).

Thanks for the explanation.  I moved the fork's master to the base's master.  Things look correct and the branch is mergeable again now.

leoluan1:
> The build is failing because of these errors
> 
> ```
> vfs/vfscache/cache_test.go:289:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
> vfs/vfscache/cache_test.go:305:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
> vfs/vfscache/cache_test.go:313:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
> vfs/vfscache/cache_test.go:323:16: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
> vfs/vfscache/cache_test.go:331:17: cannot use removeFile (type func(*Item)) as type func(*Item, time.Duration) in argument to c._purgeOld
> ```
> 
> So the existing tests in `vfscache` need fixing.

I fixed the errors related to purge.  However item_test.go have some errors.  Is there any pending bug causing those errors?  I guess not, but just to make sure while I look into them.

ncw:
I think this code is looking very good now and ready to merge :-)

Can you squash the commits - it is probably use one commit isn't it?

Can you also rename the commit(s) so the first line starts with "vfs: " - the change log generator knows what to do with them - [see guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages)

I'd like to get this merged in the next couple of days for an end of the month 1.53 release.

Thank you

leoluan1:
> I think this code is looking very good now and ready to merge :-)
> 
> Can you squash the commits - it is probably use one commit isn't it?
> 
> Can you also rename the commit(s) so the first line starts with "vfs: " - the change log generator knows what to do with them - [see guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages)
> 
> I'd like to get this merged in the next couple of days for an end of the month 1.53 release.
> 
> Thank you

That's great!  Thanks!

BTW, my tests have been running 56 hours continuously over the weekend without encountering deadlock, indefinite hang, or internal errors for reads.   Internal errors for out-of-space writes do come up occasionally by design while all the reads continue happily.

I will squash the commits and rename the commit following the guideline.

leoluan1:
> I think this code is looking very good now and ready to merge :-)
> 
> Can you squash the commits - it is probably use one commit isn't it?
> 
> Can you also rename the commit(s) so the first line starts with "vfs: " - the change log generator knows what to do with them - [see guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages)
> 
> I'd like to get this merged in the next couple of days for an end of the month 1.53 release.
> 
> Thank you

Before squashing the commits, I pushed a small code cleanup fix and ran into a github test error on Mac.  It does not appear to be related to what I changed.  I could not reproduce it on my own Mac either.   Any insight?

The errors are like the following.
```--- FAIL: TestAddPlugin (0.58s)
    rc_test.go:42: 
        	Error Trace:	rc_test.go:42
        	            				rc_test.go:86
        	Error:      	Expected nil, but got: &errors.errorString{s:"could not find an asset in the release. check if asset was successfully added in github release assets"}
        	Test:       	TestAddPlugin
    rc_test.go:88: 
        	Error Trace:	rc_test.go:88
        	Error:      	Should be true
        	Test:       	TestAddPlugin
```

ncw:
I think it all looks ok!

I'll merge this now and hopefully we'll get a bit of testing in before the release.

Thank you very much for your contribution 😃

leoluan1:
That's great!  It's a pleasure to work with you and contribute to this important project!

PR #4490, opened 2020-08-06
Fix uploadfile only works for root of the fs

negative0:
This bug ignored the remote parameter while uploading a file, thus the file will appear only in the root of the fs. This PR fixes that and adds test for it

No

#### Checklist

### DIFF
--- a/fs/operations/rc.go
+++ b/fs/operations/rc.go
@@ -6,6 +6,7 @@ import (
 	"mime"
 	"mime/multipart"
 	"net/http"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -253,7 +254,7 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 					return nil, err
 				}
 				if p.FileName() != "" {
-					obj, err := Rcat(ctx, f, p.FileName(), p, time.Now())
+					obj, err := Rcat(ctx, f, filepath.Join(remote, p.FileName()), p, time.Now())
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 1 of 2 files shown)


fs/operations/rc.go
```diff
@@ -253,7 +254,7 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 					return nil, err
 				}
 				if p.FileName() != "" {
-					obj, err := Rcat(ctx, f, p.FileName(), p, time.Now())
+					obj, err := Rcat(ctx, f, filepath.Join(remote, p.FileName()), p, time.Now())
```
ncw, line 257: That should be `path.Join` not `filepath.Join`. `filepath` is for OS native strings so will have \ instead of / on Windows.

fs/operations/rc_test.go
```diff
@@ -450,11 +451,13 @@ func TestUploadFile(t *testing.T) {
 	testFileName := "test.txt"
 	testFileContent := "Hello World"
 	r.WriteFile(testFileName, testFileContent, t1)
+	testItem1 := fstest.NewItem(testFileName, testFileContent, t1)
+	testItem2 := fstest.NewItem(filepath.Join("subdir", testFileName), testFileContent, t1)
```
ncw, line 455: s/filepath/path/ here too

fs/operations/rc.go
negative0: Ahh I see. Will update that

negative0:
Should this be failing?

negative0:
Thanks!

Issue #4486, opened 2020-08-04
Docker - RC API Inaccessible


TaylorDale:
#### What is the problem you are having with rclone?
Unable to use the RC API, relatively following along with instructions from the docs.

#### What is your rclone version (output from `rclone version`)
v1.52.2

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
sudo docker run -d -p 5572:5572   rclone/rclone:latest rcd --rc-no-auth

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
2020/08/04 05:46:31 DEBUG : rclone: Version "v1.52.2" starting with parameters ["rclone" "rcd" "--rc-no-auth" "-vv"]
2020/08/04 05:46:31 NOTICE: Serving remote control on http://127.0.0.1:5572/

Getting the current output: 
```
 curl -v -X POST 'http://127.0.0.1:5572/rc/noop?potato=1&sausage=2'
*   Trying 127.0.0.1...
* TCP_NODELAY set
* connect to 127.0.0.1 port 5572 failed: Connection refused
* Failed to connect to 127.0.0.1 port 5572: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 127.0.0.1 port 5572: Connection refused
```

I'm not sure what I'm missing, I know there are no remotes or config in this setup but when I run docker ps the docker container is running fine, and I'm only passing test params so I'd expect something atleast other than a connection error.

Thanks

darthShadow:
If you are running rclone within a container, you need to make it listen on all interfaces via `--rc-addr=0.0.0.0:5572` in order to connect from outside the container.

Explanation about why this is necessary: https://pythonspeed.com/articles/docker-connection-refused/

TaylorDale:
> If you are running rclone within a container, you need to make it listen on all interfaces via `--rc-addr=0.0.0.0:5572` in order to connect from outside the container.
> 
> Explanation about why this is necessary: https://pythonspeed.com/articles/docker-connection-refused/

Ah great thanks for that & the explanation - I have not had to do this before to get a webserver inside a docker container working but glad there is a fix.

ncw:
> If you are running rclone within a container, you need to make it listen on all interfaces via `--rc-addr=0.0.0.0:5572` in order to connect from outside the container.

This should maybe go in the docs somewhere? Maybe in the docker install section?

What did you think @darthShadow ?

darthShadow:
[...]

PR #4491, opened 2020-08-08
Updated Docker Docs

darthShadow:

Updated Docker Docs regarding usage of the RC API from outside the container

#4486 

#### Checklist

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -136,6 +136,13 @@ from the rclone image.
   reside on the host with a non-root UID:GID, you need to pass these on the container
   start command line.
 
+- If you want to access the RC interface (either via the API or the Web UI), it is
+  required to set the `--rc-addr` to `:5572` in order to connect to it from outside
+  the container. An explanation about why this is necessary is present [here](https://web.archive.org/web/20200808071950/https://pythonspeed.com/articles/docker-connection-refused/).
+    * NOTE: Users running this container with the docker network set to `host` should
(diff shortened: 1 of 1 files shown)


docs/content/install.md
```diff
@@ -136,6 +136,11 @@ from the rclone image.
   reside on the host with a non-root UID:GID, you need to pass these on the container
   start command line.
 
+- If you want to access the RC interface (either via the API or the Web UI), it is
+  required to set the `--rc-addr` to `0.0.0.0:5572` in order to connect to it from outside
```
ncw, line 140: Using `:5572` instead of `0.0.0.0:5572` should work here - what do you think about changing it?

darthShadow: Will do.

While writing it, I thought of the security implications of keeping the port open for everyone if the docker network is set to host instead of a bridge. Should I also add a line warning against that or is that too much information for the install section?

ncw: I think a sentence here would be appropriate, yes.

robertbaker:
This isn't completely true. The other way is to use a reverse proxy like traefik
`traefik.http.services.rclone.loadbalancer.server.port=5572`

darthShadow:
I may be completely wrong about this, but you probably need the global bind even for traefik or any other proxy since they communicate over the bridge and not within the network namespace of the container. In that case, binding it to localhost would restrict the traffic to only within that namespace which would make it impossible for Traefik to proxy the traffic.

PR #4494, opened 2020-08-09
drive: Replace the Changes API with the Drive Activity API

darthShadow:

Updated the Drive Backend to use the Drive Activity API instead of the Changes API

https://forum.rclone.org/t/polling-mount-cache-issue-after-server-side-move/18163/

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -48,7 +48,9 @@ import (
 	"golang.org/x/oauth2/google"
 	drive_v2 "google.golang.org/api/drive/v2"
 	drive "google.golang.org/api/drive/v3"
+	"google.golang.org/api/driveactivity/v2"
 	"google.golang.org/api/googleapi"
+	"google.golang.org/api/option"
 )
 
 // Constants
@@ -143,6 +145,8 @@ func driveScopes(scopesString string) (scopes []string) {
 		scope = strings.TrimSpace(scope)
 		scopes = append(scopes, scopePrefix+scope)
 	}
+	// The minimal read-only scope for activities should be sufficient
+	scopes = append(scopes, scopePrefix+"drive.activity.readonly")
 	return scopes
 }
 
@@ -534,20 +538,21 @@ type Options struct {
 
 // Fs represents a remote drive server
 type Fs struct {
-	name             string             // name of this remote
-	root             string             // the path we are working on
-	opt              Options            // parsed options
-	features         *fs.Features       // optional features
-	svc              *drive.Service     // the connection to the drive server
-	v2Svc            *drive_v2.Service  // used to create download links for the v2 api
-	client           *http.Client       // authorized client
-	rootFolderID     string             // the id of the root folder
-	dirCache         *dircache.DirCache // Map of directory path to directory id
-	pacer            *fs.Pacer          // To pace the API calls
-	exportExtensions []string           // preferred extensions to download docs
-	importMimeTypes  []string           // MIME types to convert to docs
-	isTeamDrive      bool               // true if this is a team drive
-	fileFields       googleapi.Field    // fields to fetch file info with
+	name             string                 // name of this remote
+	root             string                 // the path we are working on
+	opt              Options                // parsed options
+	features         *fs.Features           // optional features
+	svc              *drive.Service         // the connection to the drive server
+	v2Svc            *drive_v2.Service      // used to create download links for the v2 api
+	activitySvc      *driveactivity.Service // used for fetching the drive activities
+	client           *http.Client           // authorized client
+	rootFolderID     string                 // the id of the root folder
+	dirCache         *dircache.DirCache     // Map of directory path to directory id
+	pacer            *fs.Pacer              // To pace the API calls
+	exportExtensions []string               // preferred extensions to download docs
+	importMimeTypes  []string               // MIME types to convert to docs
+	isTeamDrive      bool                   // true if this is a team drive
+	fileFields       googleapi.Field        // fields to fetch file info with
 	m                configmap.Mapper
 	grouping         int32               // number of IDs to search at once in ListR - read with atomic
 	listRmu          *sync.Mutex         // protects listRempties
@@ -922,7 +927,7 @@ func configTeamDrive(ctx context.Context, opt *Options, m configmap.Mapper, name
 	if !config.Confirm(false) {
 		return nil
 	}
-	f, err := newFs(name, "", m)
+	f, err := newFs(ctx, name, "", m)
 	if err != nil {
 		return errors.Wrap(err, "failed to make Fs to list teamdrives")
 	}
@@ -1034,7 +1039,7 @@ func (f *Fs) setUploadCutoff(cs fs.SizeSuffix) (old fs.SizeSuffix, err error) {
 //
 // It constructs a valid Fs but doesn't attempt to figure out whether
 // it is a file or a directory.
-func newFs(name, path string, m configmap.Mapper) (*Fs, error) {
+func newFs(ctx context.Context, name, path string, m configmap.Mapper) (*Fs, error) {
 	// Parse config into Options struct
 	opt := new(Options)
 	err := configstruct.Set(m, opt)
@@ -1082,13 +1087,19 @@ func newFs(name, path string, m configmap.Mapper) (*Fs, error) {
 
 	// Create a new authorized Drive client.
 	f.client = oAuthClient
-	f.svc, err = drive.New(f.client)
+
+	f.svc, err = drive.NewService(ctx, option.WithHTTPClient(f.client))
 	if err != nil {
 		return nil, errors.Wrap(err, "couldn't create Drive client")
 	}
 
+	f.activitySvc, err = driveactivity.NewService(ctx, option.WithHTTPClient(f.client))
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't create Drive Activity client")
+	}
+
 	if f.opt.V2DownloadMinSize >= 0 {
-		f.v2Svc, err = drive_v2.New(f.client)
+		f.v2Svc, err = drive_v2.NewService(ctx, option.WithHTTPClient(f.client))
 		if err != nil {
 			return nil, errors.Wrap(err, "couldn't create Drive v2 client")
 		}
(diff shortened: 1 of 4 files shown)


backend/drive/drive.go
```diff
@@ -144,6 +146,8 @@ func driveScopes(scopesString string) (scopes []string) {
 		scope = strings.TrimSpace(scope)
 		scopes = append(scopes, scopePrefix+scope)
 	}
+	// The minimal read-only scope for activities should be sufficient
+	scopes = append(scopes, scopePrefix+"drive.activity.readonly")
```
ncw, line 150: Shouldn't this go in `defaultScope`?

backend/drive/drive.go
```diff
@@ -1078,13 +1083,19 @@ func newFs(name, path string, m configmap.Mapper) (*Fs, error) {
...
-	f.svc, err = drive.New(f.client)
+
+	f.svc, err = drive.NewService(ctx, option.WithHTTPClient(f.client))
 	if err != nil {
 		return nil, errors.Wrap(err, "couldn't create Drive client")
 	}
 
+	f.activitySvc, err = driveactivity.NewService(ctx, option.WithHTTPClient(f.client))
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't create Drive Activity client")
```
ncw, line 1094: This is going to cause a backwards incompatibility problem as this will fail for anyone who doesn't have the correct scope in their oauth token.

So we need to fail soft here, maybe with a message to do `rclone config reconnect backend:`

It might be nice to delay opening this until the ChangeNotify service is used also as people who are just doing syncing don't care about changes.

backend/drive/drive.go
```diff
@@ -1078,13 +1083,19 @@ func newFs(name, path string, m configmap.Mapper) (*Fs, error) {
...
+	}
+
 	if f.opt.V2DownloadMinSize >= 0 {
-		f.v2Svc, err = drive_v2.New(f.client)
+		f.v2Svc, err = drive_v2.NewService(ctx, option.WithHTTPClient(f.client))
```
ncw, line 1098: What is this change about? Is it the New recommended API now?

backend/drive/drive.go
```diff
@@ -2578,125 +2588,18 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 				if pollInterval != 0 {
 					ticker = time.NewTicker(pollInterval)
 					tickerC = ticker.C
+					// TODO: Make the initial buffer configurable
```
ncw, line 2591: Is this looking for changes in the past? This seems to say 10 * the pollInterval which sounds a lot. I would have thought a few seconds is all we need?

backend/drive/drive_activity.go
```diff
@@ -0,0 +1,240 @@
...
+	"time"
+
+	"github.com/rclone/rclone/fs"
+	"google.golang.org/api/driveactivity/v2"
+)
+
+const (
+	initialBuffer = 10
+
+	MoveQuery   = "detail.action_detail_case:MOVE"
```
ncw, line 16: These probably shouldn't be globals with CapitalInitialLetters

backend/drive/drive_activity.go
```diff
@@ -0,0 +1,240 @@
...
+			return
+		}
+	}
+	return nil
+}
+
+func (f *Fs) getParent(target *driveactivity.DriveItem) *parent {
+	targetTitle := target.Title
+	q := createItemQueryRequest(parseItemID(target.Name), MoveQuery, "")
+	response, err := f.activitySvc.Activity.Query(q).Do()
```
ncw, line 89: This needs to be wrapped in a `pacer` retry loop.

backend/drive/drive_activity.go
```diff
@@ -0,0 +1,240 @@
...
+			invalidatedParent = f.getParent(target)
+			invalidatedParents = addParent(invalidatedParents, invalidatedParent)
+		}
+		*/
+		if action.Detail.Move != nil {
+			fs.Debugf(targetTitle, "Moved")
+			moveAction := action.Detail.Move
+			for _, addedParent := range moveAction.AddedParents {
+				if addedParent.DriveItem == nil {
+					fs.Errorf(targetTitle, "Invalid Parent: %v\n", addedParent)
```
ncw, line 155: This message should have a bit more context when we get reports about it on the forum! Eg "Invalid parent on Moved activity item"

It also needs to return or something here as it will crash if `addedParent` is `nil`.

backend/drive/drive_activity.go
```diff
@@ -0,0 +1,240 @@
...
+				}
+				parentItem := addedParent.DriveItem
+				invalidatedParent = &parent{
+					ID: parseItemID(parentItem.Name), Name: parentItem.Title}
+				invalidatedParents = addParent(invalidatedParents, invalidatedParent)
+			}
+			for _, removedParent := range moveAction.RemovedParents {
+				if removedParent.DriveItem == nil {
+					fs.Errorf(targetTitle, "Invalid Parent: %v\n", removedParent)
```
ncw, line 164: See above!

backend/drive/drive_activity.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+		fs.Infof(f, "Retrieved Activities : %d\n", len(activityResponse.Activities))
+
+		for _, activity := range activityResponse.Activities {
+			invalidatedParents = f.parseActivity(activity)
+
+			for parentID, parentName := range invalidatedParents {
+				parentName = f.opt.Enc.ToStandardName(parentName)
+				// translate the path of this dir
+				if parentPath, ok := f.dirCache.GetInv(parentID); ok {
```
ncw, line 218: Looking things up in the dirCache only works if we have listed that directory in the past.

I can't immediately think of a better solution, but I thought I'd point it out.

The old code has this problem too.

vfs/dir.go
```diff
@@ -188,13 +188,24 @@ func (d *Dir) invalidateDir(absPath string) {
 //
 // if entryType is a directory it invalidates the parent of the directory too.
 func (d *Dir) changeNotify(relativePath string, entryType fs.EntryType) {
-	defer log.Trace(d.path, "relativePath=%q, type=%v", relativePath, entryType)("")
 	d.mu.RLock()
 	absPath := path.Join(d.path, relativePath)
 	d.mu.RUnlock()
-	d.invalidateDir(vfscommon.FindParent(absPath))
+
+	defer log.Trace(d.path, "relativePath=%q, absPath=%q, type=%v", relativePath, absPath,
```
ncw, line 195: This is obviously a test, however if you want to change the VFS code then it is probably material for another PR!

vfs/dir.go
```diff
@@ -188,13 +188,24 @@ func (d *Dir) invalidateDir(absPath string) {
...
+		entryType)("")
+
+	/*
+		TODO: Check why this is necessary. Ideally, this should happen only if the
```
ncw, line 199: I think it is because if you remove a file or a directory then you need to flush the parent of that directory.

The changes mechanism isn't descriptive enough to say whether we are deleting a file or dir, only that stuff happened on it.

darthShadow: The Changes API is descriptive enough to not require flushing the parent since we calculate the parent anyway which in turn flushes the parent of the parent, which seems like a waste of resources.

How do you think this should be handled then to avoid issues with the other backends using the current implementation? Perhaps some argument to not invalidate the parent?

backend/drive/drive_activity.go
darthShadow: Is that actually a problem? I would have assumed that's perfectly fine since dirs which have not been listed don't need to be invalidated. It will anyway fetch the current listing from the remote when it goes for listing.

backend/drive/drive.go
darthShadow: Yep. https://godoc.org/google.golang.org/api/drive/v3#New

backend/drive/drive.go
darthShadow: I think that's what happens anyway. I don't think it does any API call till the first request. I assume the tests would have failed otherwise?

backend/drive/drive.go
darthShadow: Problem with that is that any scope override from the config causes the `defaultScope` to not be used.

backend/drive/drive.go
darthShadow: This is the initial buffer period for any changes that may get missed while doing directory listings before the change notify service has started, hence the larger initial buffer. I am not sure if that's even possible or how large of an interval that could be.

Let me know if you want me to remove that or make it shorter.

darthShadow:
@ncw Do you have some time to review the changes/comments? I would like to get this change into the beta as soon as possible for further testing before 1.53 gets released.

darthShadow:
@ncw I have been running this on a few systems for a few days now and there seem to be no other problems. So this is probably ready for another review now.

ncw:
In going to leave this until 1.53 is out as I think we have enough stuff in that. I'm planning on releasing 1.53 at the end of the month so in a few days. I'll merge this PR then so it can get more testing before being released.

Hope that's ok.

darthShadow:
That's fine, but would it be possible to get a review of the changes so I can test it out myself over the next few days in case anything else requires changes?

darthShadow:
Please hold off on merging this for a few more days. I am fixing some more issues that I have observed and will probably push those changes in the next few days.

I am bringing back the drive change notifications API and will use that supplemented by the Drive Activity API to get the actual parents of the objects.

ncw:
> I am bringing back the drive change notifications API and will use that supplemented by the Drive Activity API to get the actual parents of the objects.

Does that mean we need to make two API calls? That sounds like a disadvantage. I thought everything we needed was in the Drive Activity API or have you hit a snag?

darthShadow:
Yeah, unfortunately, currently what I have observed is that when the change is done in 1 location and if we poll in another region then it can take a few minutes (up to 10-15 minutes) or even longer sometimes for the changes to get propagated.

In that scenario, we would have already polled the Drive Activity API for the changes and invalidated the directories but they won't reflect the desired listing since the changes have not propagated yet.

This is probably because the Drive Change Notify API and Drive Activity API have different use cases, with the Activity API acting as more of an audit log where the changes are reflected immediately.

darthShadow:
I will close this for now and will send a new PR once the vfs virtual cache entries problem is fixed with the updated changes.

PR #4502, opened 2020-08-12
sftp: add options for subsystem and server_command - fixes #1801

### LINKED ISSUE #1801: sftp : support the -s option for specifying the remote subsystem (for use with sudo)
I connect to my servers with a user which is not root (ssh for root user is disabled) but this user has sudoers rights. 

To be able to use sftp on system files, I normally use sftp with the following option :  -s "sudo /usr/libexec/openssh/sftp-server" ; and that gives me system level access to the server.

It would be nice to be able to do the same with rclone.

aus:

The traditional OpenSSH `sftp` command supports specifying the SSH subsystem or the SFTP server via the `-s` option. This PR adds two options to the sftp backend: `subsystem` and `server_command`. 

The `subsystem` option is useful if the target server has multiple SFTP subsystems defined for users.

The `server_command` option can be useful if `sudo` is required to elevate the user sftp session, without the need to SSH as root user.

The new `newSftpClient()` function is essentially a just reimplementation of [sftp.NewClient](https://godoc.org/github.com/pkg/sftp#NewClient) to support custom subsystem or execution of a custom SFTP server on the remote host.

It was briefly discussed in 2017 in issue #1801.

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -164,6 +164,18 @@ Home directory can be found in a shared folder called "home"
 			Default:  false,
 			Help:     "Set to skip any symlinks and any other non regular files.",
 			Advanced: true,
+		}, {
+			Name:     "subsystem",
+			Default:  "sftp",
+			Help:     "Specifies the SSH2 subsystem on the remote host.",
+			Advanced: true,
+		}, {
+			Name:    "server_command",
+			Default: "",
+			Help: `Specifies the path or command to run a sftp server on the remote host.
+
+The subsystem option is ignored when server_command is defined.`,
+			Advanced: true,
 		}},
 	}
 	fs.Register(fsi)
@@ -187,6 +199,8 @@ type Options struct {
 	Md5sumCommand     string `config:"md5sum_command"`
 	Sha1sumCommand    string `config:"sha1sum_command"`
 	SkipLinks         bool   `config:"skip_links"`
+	Subsystem         string `config:"subsystem"`
+	ServerCommand     string `config:"server_command"`
 }
 
 // Fs stores the interface to the remote SFTP files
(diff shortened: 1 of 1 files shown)


backend/sftp/sftp.go
```diff
@@ -164,6 +164,16 @@ Home directory can be found in a shared folder called "home"
...
 			Advanced: true,
+		}, {
+			Name:     "subsystem",
+			Default:  "sftp",
+			Help:     "Specifies the SSH2 subsystem on the remote host.",
+			Advanced: true,
+		}, {
+			Name:     "sftp_server",
+			Default:  "",
+			Help:     "Specifies the path or command to run a sftp server on the remote host. The subsystem option is ignored when sftp_server is defined.",
```
ncw, line 175: Can you break this line into two parts - the first line makes the option help text for `--sftp-server` and its too long!

`Specifies the path or command to run a sftp server on the remote host.

The subsystem option is ignored when sftp_server is defined.`

backend/sftp/sftp.go
```diff
@@ -164,6 +164,16 @@ Home directory can be found in a shared folder called "home"
 			Default:  false,
 			Help:     "Set to skip any symlinks and any other non regular files.",
 			Advanced: true,
+		}, {
+			Name:     "subsystem",
+			Default:  "sftp",
+			Help:     "Specifies the SSH2 subsystem on the remote host.",
+			Advanced: true,
+		}, {
+			Name:     "sftp_server",
```
ncw, line 173: Can we call this something other than `sftp_server` otherwise the option will be called `--sftp-sftp-server`! Maybe `server` or `server_command`?

backend/sftp/sftp.go
```diff
@@ -187,6 +197,8 @@ type Options struct {
 	Md5sumCommand     string `config:"md5sum_command"`
 	Sha1sumCommand    string `config:"sha1sum_command"`
 	SkipLinks         bool   `config:"skip_links"`
+	Subsystem         string `config:"subsystem"`
+	SftpServer        string `config:"sftp_server"`
```
ncw, line 201: Don't forget to rename this too!

backend/sftp/sftp.go
```diff
@@ -299,6 +311,35 @@ func (f *Fs) sftpConnection() (c *conn, err error) {
 	return c, nil
 }
 
+// Creates a new SFTP client on conn, using specificed subsystem
```
ncw, line 314: s/specificed/the specified/

aus:
Great feedback. I've implemented your suggestions and squashed the commits. (Ignore the first two erroneous force-pushes 😅)

Also, just wanted to say that I've been using rclone for years, and it's awesome! Thanks for all your work.

ncw (approved):
This looks great now thanks. I'll merge it now.

ncw:
> Great feedback. I've implemented your suggestions and squashed the commits. (Ignore the first ~two~ three erroneous force-pushes

😃

> Also, just wanted to say that I've been using rclone for years, and it's awesome! Thanks for all your work.

Thank you, and thank you for your contribution 😃

smac89:
@aus 
This is not working the way it was originally stated in #1801.

I thought the original intent by raising that issue was to allow `stfp-server` to be run with sudo. However, that does not seem to be the case. I've tried all kinds of combinations and the only option that seems to work is to just specify `sftp` as the subsystem, not even a path like `/usr/lib/openssh/sftp-server`, works.

I was under the impression that with this PR, we can now supply something like `sudo /usr/lib/openssh/sftp-server` for the subsystem option (which btw is allowed when running sftp from the commandline)

smac89:
Ignore the above. I just read the PR again, and it looks like I need to be using `server_command` instead of `subsystem`

PR #4503, opened 2020-08-12
Remove --drive-alternate-export in favor of exportLinks

themylogin:

`--drive-alternate-export` is not necessary anymore, `exportLinks` file field provides links which can handle export of big files.

No
#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -17,7 +17,6 @@ import (
 	"log"
 	"mime"
 	"net/http"
-	"net/url"
 	"path"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


backend/drive/drive.go
```diff
@@ -346,20 +345,6 @@ date is used.`,
 			Default:  "",
 			Help:     `Impersonate this user when using a service account.`,
 			Advanced: true,
-		}, {
-			Name:    "alternate_export",
```
ncw, line 350: Removing this option completely will cause scripts with `--drive-alternate-export` in them to break.

What we should do here is leave it like this

```
                }, {
                        Name:     "alternate_export",
                        Default:  false,
                        Help:     "Deprecated: no longer needed",
                        Hide:     fs.OptionHideBoth,
                }, {
```

The last line hides it from the configurator and the command line flags, but it won't throw an error if used.

themylogin:
@ncw yes, I put a lot of images into doc / spreadsheet / presentation / drawing to force resulting files to be larger than 10 MB and then confirmed that all of them can be exported with this new approach.

ncw:
This looks great now - thank you!

I've dropped the auto generated files from the commit (MANUAL.md etc).

I'll merge it now :-)

Issue #3366, opened 2019-07-22
drive-shared-with-me for Dropbox


turkeyphant:
#### What is your current rclone version (output from `rclone version`)?
v1.47.0

#### What problem are you are trying to solve?
Access folders that are shared with you but not in your Dropbox.

#### How do you think rclone should be changed to solve that?
Add a `shared-with-me` option (analogous to `[drive-shared-with-me](https://rclone.org/drive/#drive-shared-with-me)`

c.f. https://forum.rclone.org/t/drive-shared-with-me-for-dropbox/10878/2

buengese:
This appears to be easily possible with
[https://www.dropbox.com/developers/documentation/http/documentation#sharing-list_folders](https://www.dropbox.com/developers/documentation/http/documentation#sharing-list_folders)

ncw:
If we did this we'd either want to provide a new root ID, so `--dropbox-root FOLDERID`  - that would be very easy to implement.  I don't know how hard gettings IDs of shared folders are though.

Or we'd want to implement a virtual layer with the shared folders in the root and their contents below.  That would be quite a bit harder.

Or maybe a command to list shared folder IDs... This could be done in the config wizard.

hiepsieunhan:
@ncw could I know the current status of this ticket? If this is not done, does it belong to the near future plan?

ncw:
@hiepsieunhan I don't think anyone is working on this at the moment. Do you want to have a go?

PR #4527, opened 2020-08-23
dropbox: add support for viewing shared files and folders

buengese:
Whoops that turned out to be bigger a change than I had remembered. It also works somewhat differently than `--drive-shared-with-me`. This actually adds 2 flags:
* ``--dropbox-shared-files``: instructs rclone to act on individual shared files (those that show up directly under shared). In this mode rclone's features are extremely limited - only listing shared files and downloading them is supported every other operation will be disabled.
* ``--dropbox-shared-folders``: instructs rclone to operate on shared folders. When this flag is used with an empty root only the List operation is supported and all shared folders will be listed. If you use ``--dropbox-shared-folders`` with a non empty root the first part of the path will be interpreted as the name of shared folder. rclone will then try to mount this shared to the root namespace. If this succeeds shared folder mode will be disabled and rclone proceeds normally. The shared folder is now pretty much a normal folder and all normal operations are supported. Note that we don't unmount the shared folder afterwards so the ``--dropbox-shared-folders`` can be omitted after the first use of a particular shared folder.

Some Extra notes: In order to work with the shared folder we have to mount it the api doesn't really allow working with shared folders without mounting them. I haven't implemented a way for unmounting folders as of right now because I couldn't think of good of doing it. I fully expect something to break If you have name collisions working around this wouldn't be easy.

 There are probably some edge cases I haven't thought of so testing is appreciated.

### DIFF
--- a/backend/dropbox/dropbox.go
+++ b/backend/dropbox/dropbox.go
@@ -143,6 +143,31 @@ memory.  It can be set smaller if you are tight on memory.`, maxChunkSize),
 			Default:  "",
 			Advanced: true,
 		}, {
+			Name: "shared_files",
+			Help: `Instructs rclone to work on individual shared files.
+
+In this mode rclone's features are extremely limited - only list (ls, lsl, etc.) 
+operations and read operations (e.g. downloading) are supported in this mode.
+All other operations will be disabled.`,
+			Default:  false,
+			Advanced: true,
+		}, {
+			Name: "shared_folders",
+			Help: `Instructs rclone to work on shared folders.
+			
+When this flag is used with no path only the List operation is supported and 
+all available shared folders will be listed. If you specify a path the first part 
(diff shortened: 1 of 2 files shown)


ncw:
This looks very nice! What do you think about posting it to the forum to get some testers? Also was there an issue about this? Might be worth digging that out to get some testers there.

buengese:
Well yeah this is related to #3366 but does quite a bit more and also works somewhat different than `--drive-shared-with-me`.

ncw:
I'll merge this for 1.54 (when you are happy), I'm trying to feature freeze 1.53 for release next week.

buengese:
@ncw this is done btw and could be merged if you're ok.

ncw:
I gave the code a skim - looks good to me!

Merge when ready!

Note that I'm adding batch mode support to dropbox at the moment which makes uploading small files much much quicker!

PR #4547, opened 2020-09-01
fs: add Resumer interface

### LINKED ISSUE #87: Incremental sync

mcalman:
Added an interface for resuming failed uploads. Later on this interface
can be implemented by any backend where resumes are possible.

#87

Adds an optional interface for resuming uploads that can later be implemented by any backend where resuming uploads is possible. Once implemented by a backend, this will allow uploads to continue after a failure, even when restarting Rclone operations.

https://github.com/rclone/rclone/issues/87
https://forum.rclone.org/t/drive-resume-resumable-upload-implementation/18289/4
https://forum.rclone.org/t/chunker-resume-upload-contribution/18571

#### Checklist

### DIFF
--- a/backend/chunker/chunker.go
+++ b/backend/chunker/chunker.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"crypto/md5"
 	"crypto/sha1"
+	"encoding"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -13,6 +15,7 @@ import (
 	gohash "hash"
 	"io"
 	"io/ioutil"
+	"log"
 	"math/rand"
 	"path"
 	"regexp"
@@ -379,6 +382,8 @@ type Fs struct {
 	features     *fs.Features   // optional features
 	dirSort      bool           // reserved for future, ignored
 	useNoRename  bool           // can be set with the transactions option
+	hashState    string         // set in resume(), used to restore hash state
+	resumeXactID string         // set in resume(), allows reuse of xactID upon resume
 }
 
 // configure sets up chunker for given name format, meta format and hash type.
@@ -1152,7 +1157,41 @@ func (f *Fs) put(
 
 	// Prepare to upload
 	c := f.newChunkingReader(src)
-	wrapIn := c.wrapStream(ctx, in, src)
+	// Prepare for resume if resumable
+	var resumeOpt *fs.OptionResume
+	// partialHashState will be used in wrapStream to restore hash state
+	var partialHashState []byte
+	for _, option := range options {
+		switch option.(type) {
+		case *fs.OptionResume:
+			resumeOpt = option.(*fs.OptionResume)
+			if resumeOpt.Pos != 0 {
+				numChunksOnRemote := resumeOpt.Pos / int64(f.opt.ChunkSize)
+				// Checks for existing chunks on the remote
+				for i := 0; i < int(numChunksOnRemote); i++ {
+					existingChunkName := f.makeChunkName(remote, i, "", f.resumeXactID)
+					existingChunk, err := f.base.NewObject(ctx, existingChunkName)
+					// If NewObject returns an error the chunk likely doesn't exist on the remote and we cannot resume
+					if err != nil {
+						resumeOpt.Pos = 0
+						c.chunks = nil
+						break
+					}
+					c.chunks = append(c.chunks, existingChunk)
+				}
+				fs.Debugf(f, "Resuming at chunk number: %d", numChunksOnRemote)
+				partialHashState, _ = base64.StdEncoding.DecodeString(f.hashState)
+				// Discard bytes that already exist on remote
+				written, err := io.CopyN(ioutil.Discard, in, resumeOpt.Pos)
+				if err != nil {
+					return nil, err
+				}
+				c.accountBytes(written)
+				c.sizeLeft = c.sizeTotal - c.readCount
+			}
+		}
+	}
+	wrapIn := c.wrapStream(ctx, in, src, partialHashState)
 
 	var metaObject fs.Object
 	defer func() {
@@ -1162,13 +1201,22 @@ func (f *Fs) put(
 	}()
 
 	baseRemote := remote
-	xactID, errXact := f.newXactID(ctx, baseRemote)
-	if errXact != nil {
-		return nil, errXact
+	var xactID string
+	if resumeOpt != nil && resumeOpt.Pos != 0 {
+		xactID = f.resumeXactID
+	} else {
+		xactID, err = f.newXactID(ctx, baseRemote)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// Transfer chunks data
 	for c.chunkNo = 0; !c.done; c.chunkNo++ {
+		// skip to chunk we can resume from if resumeOpt is set
+		if c.chunkNo == 0 && resumeOpt != nil && resumeOpt.Pos != 0 {
+			c.chunkNo = int(resumeOpt.Pos) / int(f.opt.ChunkSize)
+		}
 		if c.chunkNo > maxSafeChunkNumber {
 			return nil, ErrChunkOverflow
 		}
@@ -1230,6 +1278,41 @@ func (f *Fs) put(
 		c.chunkLimit = c.chunkSize
 
 		c.chunks = append(c.chunks, chunk)
+
+		// If an OptionResume was passed than we should call SetID so a resume can be attempted in event of a failure
+		// ID keeps track of the first chunk that should be uploaded if a resume is attempted
+		if resumeOpt != nil {
+			// Publish hash state to control chunk
+			marshaler, ok := c.hasher.(encoding.BinaryMarshaler)
+			if !ok {
+				return nil, fmt.Errorf("The hash type does not implement encoding.BinaryMarshaler")
+			}
+			state, err := marshaler.MarshalBinary()
+			if err != nil {
+				return nil, err
+			}
+			hashType := f.opt.HashType
+			data, err := marshalPartialHashJSON(ctx, hashType, base64.StdEncoding.EncodeToString(state))
+			if err != nil {
+				return nil, err
+			}
+			controlChunkName := f.makeChunkName(remote, -1, "phash", xactID)
+			controlInfo := f.wrapInfo(src, controlChunkName, int64(len(data)))
+			controlChunk, err := basePut(ctx, bytes.NewReader(data), controlInfo)
+			defer func() {
+				_ = controlChunk.Remove(ctx)
+			}()
+			if err != nil {
+				return nil, err
+			}
+			positionStr := strconv.Itoa(c.chunkNo + 1) // stores the number of chunks uploaded
+			chunkSizeStr := strconv.FormatInt(c.chunkSize, 10)
+			startFromStr := strconv.FormatInt(int64(f.opt.StartFrom), 10)
+			err = resumeOpt.SetID(ctx, chunkSizeStr+","+startFromStr+","+positionStr+","+xactID, f.opt.HashType, base64.StdEncoding.EncodeToString(state))
+			if err != nil {
+				return nil, err
+			}
+		}
 	}
 
 	// Validate uploaded size
@@ -1356,7 +1439,7 @@ func (f *Fs) newChunkingReader(src fs.ObjectInfo) *chunkingReader {
 	return c
 }
 
-func (c *chunkingReader) wrapStream(ctx context.Context, in io.Reader, src fs.ObjectInfo) io.Reader {
+func (c *chunkingReader) wrapStream(ctx context.Context, in io.Reader, src fs.ObjectInfo, partialHashState []byte) io.Reader {
 	baseIn, wrapBack := accounting.UnWrap(in)
 
 	switch {
@@ -1391,6 +1474,15 @@ func (c *chunkingReader) wrapStream(ctx context.Context, in io.Reader, src fs.Ob
 	}
 
 	if c.hasher != nil {
+		// Restores hash state during a resume
+		if partialHashState != nil {
+			unmarshaler, ok := c.hasher.(encoding.BinaryUnmarshaler)
+			if ok {
+				if err := unmarshaler.UnmarshalBinary(partialHashState); err != nil {
+					log.Fatal("unable to unmarshal hash:", err)
+				}
+			}
+		}
 		baseIn = io.TeeReader(baseIn, c.hasher)
 	}
 	c.baseReader = baseIn
@@ -2510,6 +2602,34 @@ func unmarshalSimpleJSON(ctx context.Context, metaObject fs.Object, data []byte)
 	return info, true, nil
 }
 
+// Format for partial hash control chunks
+type partialHashJSON struct {
+	HashType    string `json:"htype"`
+	PartialHash string `json:"phash"`
+}
+
+// marshalPartialHashJSON
+//
+// Creates a JSON containing the hashType being used and the partial hash state. This will be stored in
+// a control chunk and used for resume functionality.
+//
+func marshalPartialHashJSON(ctx context.Context, hashType, partialHash string) ([]byte, error) {
+	controlData := partialHashJSON{
+		HashType:    hashType,
+		PartialHash: partialHash,
+	}
+	data, err := json.Marshal(&controlData)
+	return data, err
+}
+
+// unmarshalPartialHashJSON parses partial hash control chunk.
+//
+func unmarshalPartialHashJSON(ctx context.Context, data []byte) (hashType, partialHashState string, err error) {
+	var partialHashData partialHashJSON
+	err = json.Unmarshal(data, &partialHashData)
+	return partialHashData.HashType, partialHashData.PartialHash, err
+}
+
 func silentlyRemove(ctx context.Context, o fs.Object) {
 	_ = o.Remove(ctx) // ignore error
 }
@@ -2544,9 +2664,58 @@ func (f *Fs) CanQuickRename() bool {
 	return f.base.Features().Move != nil
 }
 
+// Resume checks whether the (remote, ID) pair is valid and returns
+// the point the file should be resumed from or an error.
+func (f *Fs) Resume(ctx context.Context, remote, ID, hashName, hashState string) (Pos int64, err error) {
+	idSlice := strings.Split(ID, ",")
+	cachedChunkSize, err := strconv.ParseInt(idSlice[0], 10, 64)
+	cachedStartFrom, err := strconv.ParseInt(idSlice[1], 10, 64)
+	cachedChunkNo, err := strconv.ParseInt(idSlice[2], 10, 64)
+	cachedXactID := idSlice[3]
+	if err != nil {
+		return 0, err
+	}
+	if cachedChunkSize != int64(f.opt.ChunkSize) {
+		return 0, errors.New("ChunkSize doesn't match for file we are trying to resume")
+	}
+	if f.opt.StartFrom != int(cachedStartFrom) {
+		return 0, errors.New("StartFrom doesn't match for file we are trying to resume")
+	}
+	// Check partial hash control chunk
+	controlChunkName := f.makeChunkName(remote, -1, "phash", cachedXactID)
+	hashControlChunk, err := f.base.NewObject(ctx, controlChunkName)
+	if err != nil {
+		return 0, err
+	}
+	reader, err := hashControlChunk.Open(ctx)
+	data, err := ioutil.ReadAll(reader)
+	_ = reader.Close() // ensure file handle is freed on windows
+	if err != nil {
+		return 0, err
+	}
+	remoteHashType, remoteHashState, err := unmarshalPartialHashJSON(ctx, data)
+	if remoteHashType == hashName && remoteHashState == hashState {
+		if f.opt.HashType != remoteHashType {
+			fs.Debugf(f, "Resume skipped, mismatch hash types. prev: %s, curr: %s", remoteHashType, f.opt.HashType)
+			return 0, nil
+		}
+		pos := cachedChunkNo * cachedChunkSize
+		if err != nil {
+			return 0, err
+		}
+		f.hashState = hashState
+		f.resumeXactID = cachedXactID
+		return pos, nil
+	}
+
+	// No valid control chunks found, rewind from start
+	return 0, nil
+}
+
 // Check the interfaces are satisfied
 var (
 	_ fs.Fs              = (*Fs)(nil)
+	_ fs.Resumer         = (*Fs)(nil)
 	_ fs.Purger          = (*Fs)(nil)
 	_ fs.Copier          = (*Fs)(nil)
 	_ fs.Mover           = (*Fs)(nil)
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -24,6 +25,7 @@ import (
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/lib/atexit"
 	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/file"
 	"github.com/rclone/rclone/lib/readers"
@@ -230,6 +232,7 @@ type Fs struct {
 	precision   time.Duration       // precision of local filesystem
 	warnedMu    sync.Mutex          // used for locking access to 'warned'.
 	warned      map[string]struct{} // whether we have warned about this string
+	hashState   map[string]string   // set in resume(), used to restore hash state
 
 	// do os.Lstat or os.Stat
 	lstat        func(name string) (os.FileInfo, error)
@@ -267,11 +270,12 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	}
 
 	f := &Fs{
-		name:   name,
-		opt:    *opt,
-		warned: make(map[string]struct{}),
-		dev:    devUnset,
-		lstat:  os.Lstat,
+		name:      name,
+		opt:       *opt,
+		warned:    make(map[string]struct{}),
+		hashState: make(map[string]string),
+		dev:       devUnset,
+		lstat:     os.Lstat,
 	}
 	f.root = cleanRootPath(root, f.opt.NoUNC, f.opt.Enc)
 	f.features = (&fs.Features{
@@ -1115,6 +1119,7 @@ func (nwc nopWriterCloser) Close() error {
 func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) (err error) {
 	var out io.WriteCloser
 	var hasher *hash.MultiHasher
+	var resumeOpt *fs.OptionResume
 
 	for _, option := range options {
 		switch x := option.(type) {
@@ -1122,6 +1127,24 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 			if x.Hashes.Count() > 0 {
 				hasher, err = hash.NewMultiHasherTypes(x.Hashes)
 				if err != nil {
+					return err
+				}
+			}
+		case *fs.OptionResume:
+			resumeOpt = option.(*fs.OptionResume)
+			if resumeOpt.Pos != 0 {
+				fs.Logf(o, "Resuming at byte position: %d", resumeOpt.Pos)
+				// Discard bytes that already exist on backend
+				_, err := io.CopyN(ioutil.Discard, in, resumeOpt.Pos)
+				if err != nil {
+					return err
+				}
+				hashType := o.fs.Hashes().GetOne()
+				hasher, err = hash.NewMultiHasherTypes(hash.NewHashSet(hashType))
+				if err != nil {
+					return err
+				}
+				if err := hasher.RestoreHashState(hashType, o.fs.hashState[o.remote]); err != nil {
 					return err
 				}
 			}
@@ -1138,7 +1161,12 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 	// If it is a translated link, just read in the contents, and
 	// then create a symlink
 	if !o.translatedLink {
-		f, err := file.OpenFile(o.path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
+		var f *os.File
+		if resumeOpt != nil && resumeOpt.Pos != 0 {
+			f, err = file.OpenFile(o.path, os.O_WRONLY|os.O_APPEND, 0666)
+		} else {
+			f, err = file.OpenFile(o.path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
+		}
 		if err != nil {
 			if runtime.GOOS == "windows" && os.IsPermission(err) {
 				// If permission denied on Windows might be trying to update a
@@ -1152,7 +1180,7 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 				return err
 			}
 		}
-		if !o.fs.opt.NoPreAllocate {
+		if !o.fs.opt.NoPreAllocate && resumeOpt != nil && resumeOpt.Pos == 0 {
 			// Pre-allocate the file for performance reasons
 			err = file.PreAllocate(src.Size(), f)
 			if err != nil {
@@ -1173,7 +1201,40 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 		in = io.TeeReader(in, hasher)
 	}
 
-	_, err = io.Copy(out, in)
+	var cacheingWg sync.WaitGroup // Used to halt code execution while resume cache is written
+	var copyWg sync.WaitGroup     // Ensure that io.Copy has returned before writing resume data
+	copyWg.Add(1)
+	// Context for read so that we can handle io.copy being interrupted
+	ctxr, cancel := context.WithCancel(ctx)
+	// Create exit handler during Copy so that resume data can be written if interrupted
+	atexitHandle := atexit.Register(func() {
+		// If OptionResume was passed, call SetID to prepare for future resumes
+		// ID is the number of bytes written to the destination
+		if resumeOpt != nil {
+			// Stops the copy so cache is consistent with remote
+			cacheingWg.Add(1)
+			cancel()
+			copyWg.Wait()
+			fs.Infof(o, "Updating resume cache")
+			fileInfo, _ := o.fs.lstat(o.path)
+			writtenStr := strconv.FormatInt(fileInfo.Size(), 10)
+			hashType := o.fs.Hashes().GetOne()
+			hashState, err := hasher.GetHashState(hashType)
+			if err != nil {
+				return
+			}
+			_ = resumeOpt.SetID(ctx, writtenStr, hashType.String(), hashState)
+		}
+	})
+	cr := readers.NewContextReader(ctxr, in)
+	_, err = io.Copy(out, cr)
+	copyWg.Done()
+	atexit.Unregister(atexitHandle)
+	if errors.Is(err, context.Canceled) {
+		// If resume data is being written we want to wait here for the program to exit
+		cacheingWg.Wait()
+	}
+
 	closeErr := out.Close()
 	if err == nil {
 		err = closeErr
@@ -1338,9 +1399,44 @@ func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
 	return s
 }
 
+// Resume checks whether the (remote, ID) pair is valid and returns
+// the point the file should be resumed from or an error.
+func (f *Fs) Resume(ctx context.Context, remote, ID, hashName, hashState string) (Pos int64, err error) {
+	cachedPos, err := strconv.ParseInt(ID, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	// Compare hash of partial file on remote with partial hash in cache
+	remoteObject, err := f.NewObject(ctx, remote)
+	if err != nil {
+		return 0, err
+	}
+	if remoteObject.Size() != cachedPos {
+		return 0, errors.New("size on remote does not match resume cache")
+	}
+	hashType := hash.NameToType(hashName)
+	remoteHash, err := remoteObject.Hash(ctx, hashType)
+	if err != nil {
+		return 0, err
+	}
+	cachedHash, err := hash.SumPartialHash(hashName, hashState)
+	if err != nil {
+		return 0, err
+	}
+	// Hashes match, attempt resume
+	if cachedHash == remoteHash {
+		f.hashState[remote] = hashState
+		return cachedPos, nil
+	}
+	// No valid position found, restart from beginning
+	fs.Infof(remote, "Not resuming as cached hash state did not match hash state on remote")
+	return 0, nil
+}
+
 // Check the interfaces are satisfied
 var (
 	_ fs.Fs             = &Fs{}
+	_ fs.Resumer        = &Fs{}
 	_ fs.Purger         = &Fs{}
 	_ fs.PutStreamer    = &Fs{}
 	_ fs.Mover          = &Fs{}
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1140,6 +1140,14 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 					return err
 				}
 				hashType := o.fs.Hashes().GetOne()
+				if resumeOpt.Hash != "" {
+					if err = hashType.Set(resumeOpt.Hash); err != nil {
+						return err
+					}
+					if !o.fs.Hashes().Contains(hashType) {
+						return fmt.Errorf("unsupported resume hash: %q", resumeOpt.Hash)
+					}
+				}
 				hasher, err = hash.NewMultiHasherTypes(hash.NewHashSet(hashType))
 				if err != nil {
 					return err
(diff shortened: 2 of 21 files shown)


ncw:
This looks very nice :-)

I'm just in the middle of making the 1.53 release - I will review in detail after that is done.

fs/operations/operations.go
```diff
@@ -343,6 +344,112 @@ func CommonHash(fa, fb fs.Info) (hash.Type, *fs.HashesOption) {
...
+	resumeOpt := &fs.OptionResume{}
+	if doResume := f.Features().Resume; doResume != nil {
+		// Creates an OptionResume that will be passed to Put/Upload
+		resumeOpt = &fs.OptionResume{ID: "", Pos: 0, SetID:
+		// SetID will be called by backend's Put/Update function if the object's upload
+		// could be resumed upon failure
+		//
+		// SetID takes the passed resume ID and Fingerprint of the object and stores it in
+		// --cache-dir so that future Copy operations can resume the upload if it fails
+		func(ID string) (err error) {
```
ncw, line 356: This should be a separate function too I think, so the closure you pass as SetID: is only a line or two long.

fs/operations/operations.go
```diff
@@ -343,6 +344,112 @@ func CommonHash(fa, fb fs.Info) (hash.Type, *fs.HashesOption) {
...
+			rootCacheDir := filepath.Join(config.CacheDir, "resume")
+			if filepath.Ext(rootCacheDir) != "" {
```
ncw, line 358: I don't understand why you are looking at `filepath.Ext` here? Surely the last path item is "resume" isn't it?

fs/operations/operations.go
```diff
@@ -343,6 +344,112 @@ func CommonHash(fa, fb fs.Info) (hash.Type, *fs.HashesOption) {
...
+			}
+			cachePath := filepath.Join(dirPath, remote)
+			cacheFile, err := os.Create(cachePath)
+			if err != nil {
+				return errors.Wrapf(err, "failed to create cache file %v", cachePath)
+			}
+			defer func() {
+				_ = cacheFile.Close()
+			}()
+			// Check how many resume cache files already exist and remove oldest if more than limit
```
ncw, line 375: Do we want to do this on *every* file we upload?

fs/operations/operations.go
```diff
@@ -343,6 +344,112 @@ func CommonHash(fa, fb fs.Info) (hash.Type, *fs.HashesOption) {
...
+			}
+			// Write resume data to disk
+			_, errWrite := cacheFile.Write(data)
+			if err != nil {
+				return errors.Wrapf(errWrite, "failed to write JSON to file")
+			}
+			return nil
+		}}
+
+		// Check to see if a resume cache file exists for the file currrently being uploaded
```
ncw, line 426: Another candidate for a function?

fs/operations/operations.go
```diff
@@ -428,6 +535,10 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 							wrappedSrc = NewOverrideRemote(src, remote)
 						}
 						options := []fs.OpenOption{hashOption}
+						// Prepends OptionResume if it was set
+						if resumeOpt != (&fs.OptionResume{}) {
```
ncw, line 539: This could be `resumeOpt != nil` if you used the pointer above.

fs/operations/operations.go
```diff
@@ -428,6 +535,10 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 							wrappedSrc = NewOverrideRemote(src, remote)
 						}
 						options := []fs.OpenOption{hashOption}
+						// Prepends OptionResume if it was set
```
ncw, line 538: I don't think you need to prepend it, append will do like the other options.

fs/operations/operations.go
```diff
@@ -492,6 +609,33 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
...
+	resumedata := resumeJSON{
+		Fingerprint: fprint,
+		ID:          id,
+	}
+	data, err := json.Marshal(&resumedata)
+	return data, err
+}
+
+func unmarshalResumeJSON(ctx context.Context, data []byte) (fprint, id string, err error) {
+	if data == nil || len(data) < 2 || data[0] != '{' || data[len(data)-1] != '}' {
```
ncw, line 628: You don't need to do these tests `json.Unmarshal` will do them.

ncw (commented):
v1.53 is out the door! Apologies for the delay in reviewing.

I think this is a good start - thank you :-)

I wrote lots of comments inline.

Can you factor the ~100 extra lines in `operations.Copy` so we try to keep things readable.

We'll need to write tests too, so having functions we can test individually will be very useful :-)

fs/operations/operations.go
ivandeex: compare e.g. with _curl_ tool, which activates resume logic only when user adds _--continue-at_ on command line (https://curl.se/docs/manpage.html, google "curl continue")

ivandeex:
Related to https://github.com/rclone/rclone/issues/4798 (multi-thread upload interface)

ivandeex:
@mcalman 
Your PR has been rebased on current rclone master.
Please  **pull** results from github. **Don't** force-push now.
Please address remaining review comments too.
Thanks.

fs/operations/operations.go
mcalman: I made a slight tweak so that the cache is only cleaned once per file uploaded. Previously, if a backend had to repeatedly update the cache for a single file (e.g. Chunker updating the cache with each successfully uploaded chunk) the cache was being unnecessarily cleaned during each update.

@ncw Do you have any suggestions for how we might further mitigate how often the cache is cleaned?

@ivandeex This is certainly something I can look into adding. Would we want such an option to only control whether or not to attempt a resume on an interrupted upload, or would we also want to only conditionally write the necessary information to the cache to make a later resume possible? Either with an option or maybe only for files above a certain size where resumes are more useful?

ivandeex:
@mcalman 
Please address inline review comments and flag us in a comment when you are ready.

mcalman:
@ivandeex @ncw All inline comments have been addressed except for this one which I'd like to hear your feedback on. https://github.com/rclone/rclone/pull/4547#discussion_r570464650

backend/chunker/chunker.go
```diff
@@ -1352,6 +1352,20 @@ func (f *Fs) Purge(ctx context.Context, dir string) error {
 	return do(ctx, dir)
 }
 
+// Resume checks whether the (remote, ID) pair is valid and returns
+// the point the file should be resumed from or an error.
+//
+// This command will chain to `resume` from wrapped remote.
+// Can later be replaced by an implementaiton of `resume` for chunker
```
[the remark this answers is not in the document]
ivandeex: @mcalman
we can't just chain to the lower backend in a general case. if the file is chunked it's remote chains to a small metadata (and to nothing if metadata is disabled). if it's non-chunked, it can become chunked after resume, but we can't predict it given arguments here.

i suggest that you either return unsupported error or remove interface from chunker.

@ncw 
you requested to add resume support to all wrapping backends. i'm not sure how this can be done for chunker.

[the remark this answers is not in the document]
ivandeex: let's add the opt-out clause for chunker

ivandeex:
@ncw
the flag can be named `--resume-larger <SizeSuffix>` and it will attempt resuming only if already uploaded size exceeds the parameter
or it can mimic curl and be just `--continue`, meaning "resume anything with non-zero already upliaded size"
I'm in favor of parameterized flag.
What do you think??

ivandeex:
This pull d4sperately lacks tests.
Tests will need at least one backend suppirting resume
I can only think of the local fs in that role (using file Open in append mode)

ivandeex:
If (and only if!) you want to test resume with chunker, you'll have to implement resume in chunker.

Consequently you'll want to add a new per-transaction control chunk to save info about partial hash(es) and probably hashes of uploaded chunks. Then you'll want to add code that selects transaction to resume given partial hash and upliaded size so far. Maybe select the "best" partial transaction (when _rename_ is fast) or just pick a single partial txn ID (when it's slow).

Bottomline: this will create metadata v3 and combinatorically grow number of tested combinations. So if the answer is "yes I want", I propose to commit this and the other chunker pull on a branch (after succesful reviews). This will create a beta release for beta-testing. Later we can  merge both pulls from the branch on master together using a single metadata version number.

ivandeex (changes requested):

ivandeex:
specific notes about cache:
- who decides to copy source to upload cache, backend or generic code in operations.go?
- if a source is "stable", can we avoid copying to cache and keep only json?
- if a source is "dynamic", can we force copying to cache?

ivandeex:
a note why i care about resumable checksums (if supported by a backend).
consider this scenario:
- user uploads a large file
- network broken, upliad canceled
- source file is changed or another attempt changes partial upliad on target
- user asks to resume a file
- rclone resumes (here we could have checked validity of partial upload)
- after some hours rclone finds that fingerprint is wrong
- abort, retry

ivandeex:
Another note, which relates simulatenously to the upload cache size, rolling hash and specific implementation of Resumer in the chunker backend (it can be true for the Local backend true, not sure about others).

In case of chunker particularly, upload cache usage can be somewhat decreased because already uploaded chunks are isolated remotely and marked by a "txn id", we just need a way to re-check them, and this is where the rolling hash can help us.

Again, implementation of Resumer for chunker is absolutely OUT of scope for this pull request,
but providing a framework, mechanisms, unit tests - IN.

ivandeex:
@mcalman I will re-request review from from ncw after we clear current round.

ivandeex:
Related to https://forum.rclone.org/t/http-upload-contribution/13271/38
Related to #3842 and #3850

backend/chunker/chunker.go
```diff
@@ -1465,6 +1465,20 @@ func (f *Fs) Purge(ctx context.Context, dir string) error {
...
+//
+// This command will chain to `resume` from wrapped remote.
+// Can later be replaced by an implementation of `resume` for chunker
+//
+func (f *Fs) Resume(ctx context.Context, remote string, ID string) (Pos int64, err error) {
+	do := f.base.Features().Resume
+	if do == nil {
+		return 0, fs.ErrorCantResume
+	}
+	return do(ctx, remote, ID)
```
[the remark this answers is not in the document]
ivandeex: @mcalman
Please also add `UnimplementableFsMethods: []string{..., "Resume"},` to the `chunker_test.go`, and we can mark this bit as resolved

fs/operations/operations.go
ivandeex: `--resume-larger` satisfies the request.

> Would we want such an option to only control whether or not to attempt a resume on an interrupted upload, or would we also want to only conditionally write the necessary information to the cache to make a later resume possible?

I think this bit will become clear when you proceed to adding unit tests (resume normally or resume after corrupted source/destination) and making them pass.

backend/chunker/chunker.go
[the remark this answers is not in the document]
ivandeex: I think you can completely remove the `func Resume()` stub from `chunker.go`. Let's address it in another PR.
Testing resume interface with chunker is not a low hanging fruit, though it looks tempting.
I think `backend/local` is a better candidate for that.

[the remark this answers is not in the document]
ivandeex: @mcalman
I think you can completely remove the `func Resume()` stub from `chunker.go`.
Testing the resume interface with chunker is not as simple as it may seem, though tempting. Let's address it in another PR.
I think `backend/local` is a better candidate for that.
Please also add `UnimplementableFsMethods: []string{..., "Resume"},` to the `chunker_test.go` so we can mark this thread as resolved.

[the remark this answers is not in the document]
ivandeex: @mcalman
I think you can completely remove the `func Resume()` stub from `chunker.go`.
Testing the resume interface with chunker is not as simple as it may seem, though tempting. Let's address it elsewhere.
I think `backend/local` is the best candidate for unit tests here.
Please also add `UnimplementableFsMethods: []string{..., "Resume"},` to the `chunker_test.go` so we can mark this thread as resolved.

ivandeex:
@mcalman
Please squash commits and rebase on current master.

ivandeex (approved):

ivandeex:
@mcalman 
Please add tests, either in the form of synthetic tests against interface functions, or optionally using backend/local as the simplest possible implementation of Resumer.

ivandeex:
@ncw 
This PR will need some architecture coordination with TUS uploader #3850

ivandeex:
@ncw 
Please have a look at this.

Note that it additionally implements interface contracts stated in #5154 and in the discussion above (I hid all implemented comments, only the relevant ones are visible by default).
Also note that PR #5185 (implement Resumer in Chunker) depends on this one.

This PR will need some architecture coordination with #3850 (TUS uploader) and competing PR #5299 (ResumableUpload), which can be done only by you.

ivandeex:
@ncw 
Please have a look at this.

Note that in addition to your inline comments it implements interface contracts stated in #5154 and in the discussion above (I hid all outdated comments, only the relevant ones are visible by default).
Also note that PR #5185 (implement Resumer in Chunker) depends on this one.

This PR will need some architecture coordination with #3850 (TUS uploader) and competing PR #5299 (ResumableUpload), which can be done only by you.

fs/fs.go
```diff
@@ -932,6 +940,13 @@ func (ft *Features) WrapsFs(f Fs, w Fs) *Features {
 	return ft
 }
 
+// Resumer is an optional interface for Fs
+type Resumer interface {
+	// Resume checks whether the (remote, ID) pair is valid and returns
+	// the point the file should be resumed from or an error.
+	Resume(ctx context.Context, remote, ID, hashName, hashState string) (Pos int64, err error)
```
innovate-invent, line 947: I don't think I agree with the idea that a hash should be generally provided by backends. That means that all backends must be able to provide a hash.

fs/operations/resume.go
```diff
@@ -0,0 +1,178 @@
...
+			totalCacheSize -= pathsWithInfo[p].Size()
+			fs.Debugf(p, "Successfully removed oldest cache file")
+		}
+	}
+	return nil
+}
+
+// readResumeCache checks to see if a resume ID has been cached for the source object.
+// If it finds one it returns it along with true to signal a resume can be attempted
+func readResumeCache(ctx context.Context, f fs.Fs, src fs.Object, cacheName string) (resumeID, hashName, hashState string, attemptResume bool) {
```
innovate-invent, line 129: This implementation was written with the chunker in mind, but not all backends need to manage uploads as a set of files.

Assuming that there is a cache folder available forces backend resumable uploads into a specific method of managing the uploads.

For example, if I wanted to implement resumable uploads using the multipart upload functionality of the AWS S3 API, then there are no objects created in the backend until the upload is complete.

innovate-invent (commented):
I am a bit worried this implementation is very biased towards the chunker functionality and places restrictions on how a backend can support resumable uploads.

fs/fs.go
ivandeex: No. The proper supports that chunker feature but other backends are not required to support it.

from https://github.com/rclone/rclone/pull/4547#issuecomment-786889739
> if backend does not support it, its interface will return a nil accrued hashsum

fs/operations/resume.go
ivandeex: See above

ivandeex: The implementation is generic enough to handle every suppirted backend.

fs/fs.go
innovate-invent: but the hash is in the interface signature

fs/operations/resume.go
innovate-invent: The point isn't that some backends are not capable of managing a cache, it is that this forces backends to implement a cache when they might already have better support for resumable uploads via a different mechanism.

ivandeex:
The interface is generic enough to include features of known backends including S3 or chunker.
Please elaborate what restrictions you think are placed.

fs/fs.go
ivandeex: It's in signature but it can be nil.

fs/operations/resume.go
ivandeex: > this forces backends to implement a cache

No. They will just return empty hash.

innovate-invent:
> The interface is generic enough to include features of known backends including S3 or chunker.
> Please elaborate what restrictions you think are placed.

My reservation is that this is more of a framework than an interface. Decisions are being made on how the backend has to implement resumable uploads. In my implementation I was very careful to leave the interface abstract and keep any specific implementation in a separate location that a backend can then import. For example, my [fragmentUploader](https://github.com/innovate-invent/rclone/blob/resumable/fs/resumable/fragmentuploader.go) (which is very similar to your cache) is only one possible implementation available to backends.

fs/fs.go
ivandeex: > resumable uploads via a different mechanism.

This interface is generic enough but optional. Backends are free to ignore it. Please take a closer look at the source code.

ivandeex: Here is quick synopsis of the code.
There are three parts here: the backend, the resume proper which keeps local cache of partial uploads and generic interface between them.
The flow is:
- During upload the resume proper takes snapshot by requesting supported info from backend.
- If later user decides to attempt resuming, the proper performs cache lookup and queries backend providing available info and a callback.
- The backend picks up supported pieces from the provided info, matches it with upstream, chooses resume identifier and calks back to the proper
- The resume aborts or continues

Notes: Rolling hashsums are just part of matched info along with size or name. They are generally optional. They can be supported by local or chunker backends. They provide additional data safety against broken uploads. There must be a good reason to drop them.

innovate-invent: > resume proper which keeps local cache of partial uploads

Do you mean the cache is not stored on the remote? You can't assume that there is a local filesystem available to rclone.

> resume proper takes snapshot by requesting supported info from backend

rclone should make effort to support resumable uploads regardless of if the backend supports it or not. This can be done with a configurable fallback backend for intermediate files

> There must be a good reason to drop them.

I am not proposing that we drop the hashes, they just need to be a separate interface, complimentary to the resumer interface. I quite like the idea of the rolling hashsum, the implementation should just be composable rather than ignorable parameters.

fs/operations/resume.go
ivandeex: > resumable uploads via a different mechanism

If eg. S3 needs other data to resume an upload, that other data will be added to the signature. If the signature grows above 3-4 parameters, we can factor it and pass a structure.

fs/fs.go
ivandeex: > Do you mean the cache is not stored on the remote?

It's stored locally under ~/.cache/rclone/uploads or whatever

> You can't assume that there is a local filesystem available to rclone.

Indeed rclone keeps configs and vfs cache locally.

> rclone should make effort to support resumable uploads regardless of if the backend supports it or not. This can be done with a configurable fallback backend for intermediate files

Supporting "fallback upload backends" is out of scope for this PR. This can be discussed elsewhere.

fs/operations/resume.go
ivandeex: > the implementation should just be composable rather than ignorable parameters.

I'd prefer the latter. This is discussible.

fs/fs.go
ivandeex: > the implementation should just be composable rather than ignorable parameters.

I'd prefer the latter. This is discussible.

ivandeex:
> In my implementation I was very careful to leave the interface abstract and keep any specific implementation in a separate location that a backend can then import.

I just care that supported are chunker features that provide for upload safety and leverage chunker-specific on-backend store. It's up to resumer code (which will describe interface, implementation, framework and so on) to make it coherent with S3 or whatever features or choose appropriate level of abstraction. Above I called for @ncw to decide on that.

ivandeex:
> my [fragmentUploader](https://github.com/innovate-invent/rclone/blob/resumable/fs/resumable/fragmentuploader.go) (which is very similar to your cache) is `only one possible` implementation available to backends.

:smiley:

innovate-invent:
> > my [fragmentUploader](https://github.com/innovate-invent/rclone/blob/resumable/fs/resumable/fragmentuploader.go) (which is very similar to your cache) is `only one possible` implementation available to backends.
> 
> smiley

only one possible, with multiple provided. The backend is free to implement the Uploader interface any way it wants and is not required to use any of the provided implementations.

mcalman:
@ivandeex As you suggested I am working on an implementation of Resumer for the local backend so that tests can be written for the interface. Since the local backend supports using multiple hash types at once (with the `MultiHasher`), what should we store in `hashName` and `hashState`?

ivandeex:
@mcalman 
Cool! Unit test is the last major sumbling block, I guess.

I suggest using first found hash type (for localfs it will be `md5`) like [object fingerprinter ](https://github.com/rclone/rclone/blob/v1.55.1/fs/fingerprint.go#L45):

```
hashType := f.Hashes().GetOne()
```

For saving `hashState` (intermediate hashing state of an incomplete file) you'll want to extend the [rclone MultiHasher](https://github.com/rclone/rclone/blob/v1.55.1/fs/hash/hash.go#L198) with a function following the pattern in the golang [hash marshaling example ](https://golang.org/pkg/hash/#example__binaryMarshaler) to return a snapshot of partial hashing operation.  Here is a sketch (names are tentative, you can use `hex` instead of `base64` if you like):

```
import "encoding/base64"

func (m *MultiHasher) GetHashState(hashType Type) (string, error) {
	h, ok := m.h[hashType]
	if !ok {
		return "", ErrUnsupported
	}
	snapshotter, ok := h.(encoding.BinaryMarshaler)
	if !ok {
		return "", errors.New("snapshot not implemented")
	}
	data, err := snapshotter.MarshalBinary()
	if err != nil {
		return "", err
	}
	return base64.StdEncoding.EncodeToString(data), nil
}
```

You will also need a reverse function that resets MultiHasher (or one of its hashes) to a previous state,
so upload and hashing can resume.

ivandeex:
@mcalman
Rebased on new master. Please don't force-push now.

The pre-state of commits has been saved under [ivandeex/mcalman-add-resume-interface](https://github.com/ivandeex/rclone/tree/mcalman-add-resume-interface).

Also rebased #5185 on top of this PR.

mcalman:
@ivandeex @ncw As requested I implemented the Resumer interface for the local backend to allow for tests to be written. I also added an initial test that that simulates an upload being interrupted, restarted, and then checks that it was successfully resumed. Please let me know what other tests you think should be added or if you see any problems with the implementation.

ivandeex:
@mcalman 

> Please let me know what other tests you think should be added or if you see any problems with the implementation.

I don't have any more requirements.

New tests fail on Windows because you use Unix-only kill signal etc. The simplest and fastest way to mitigate this is to put them in a separate xxx_test.go file with build comment `// +build !windows`. The tests will not be run on Windows. Another way is to use core go APIs from `os/exec` instead of unix-only.

Thank you very much for patience.

@ncw It's your turn to review this patch.

lib/atexit/atexit.go
```diff
@@ -46,12 +46,12 @@ func Register(fn func()) FnHandle {
 			if sig == nil {
 				return
 			}
-			signal.Stop(exitChan)
+			//signal.Stop(exitChan)
```
ivandeex, line 49: This change will break a lot in rclone.

ivandeex (commented):
See inline comments

ivandeex:
Rebased on latest master to fix merge conflict.

fs/operations/resume.go
```diff
@@ -0,0 +1,200 @@
...
+// SetID will be called by backend's Put/Update function if the object's upload
+// could be resumed upon failure
+//
+// SetID takes the passed resume ID, hash state, hash name and Fingerprint of the object and stores it in
+// --cache-dir so that future Copy operations can resume the upload if it fails
+func createSetID(ctx context.Context, f fs.Fs, remote string, src fs.Object) func(ID, hashName, hashState string) (err error) {
+	ci := fs.GetConfig(ctx)
+	cacheCleaned := false
+	return func(ID, hashName, hashState string) (err error) {
+		rootCacheDir := filepath.Join(config.CacheDir, "resume")
```
ivandeex, line 56: @mcalman 
Sorry. I thought I've finished reviewing but just now I noticed that this place violates a Go convention.
The context pointer "ctx" must always be passed only with call stack, but here it's passed from the resume option setup code in `createSetID` via closure to the returned anonymus function that will be called later by backend, probably in a separate call stack.
I suggest to clarify this place as follows: get rid of the anonymous function with its work data passed via closure. Instead, declare it as a public method of the resume option structure:
```
func (opt *fs.OptionResume) SetID(ctx context.Context, ID, hashName,  hashState string) error {
 ...
}
```
Then call it in backend as `err := resumeOpt.SetID(ctx, id, ...)`.
Pass its state, e.g `cacheCleaned` as a private member of the resume option structure.
Don't pass context or anything derived from it (`ci`) via structure, it's a violation altogether.
Derive it righr from new context in `SetID`.

ivandeex (commented):
Rebased on latest master to fix merge conflict.
Please pull rebased tree from github.

Also fixed unit test because after commit 6b7d7d0 when rclone is signalled it will now exit with `signum` or `128+signum` code instead of `1` (or `0`).

fs/operations/resume_test.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return 0, err
+	}
+	go func() {
+		for {
+			if atexit.Signalled() {
+				wg.Done()
+				return
+			}
+		}
```
ivandeex, line 41: `Signalled` returns immediately so this loop is in fact a short-living 100% cpu hog, a spin-poll. I suggest to add a small delay like `time.Sleep(10*time.Millisecond)` after the `if`.

ivandeex (commented):
:(

fs/operations/resume_test.go
ivandeex: @mcalman
`Signalled` returns immediately so this loop is in fact a short-living spin-poll, 100% cpu hog.
Generally this means that you'd better revise your solution idea. I am leaving that decision to @ncw and yourself.
As a _quick fix_ I would add a small delay: (?)
```
go func() {
		for {
			if atexit.Signalled() {
				wg.Done()
				return
			}
			time.Sleep(10 * time.Millisecond)
		}
} 
```

ivandeex (approved):

ivandeex (approved):
approved

ivandeex (commented):
approved

ivandeex:
This comment is not a requirement but a side note.
I  skimmed through the subthread of #87spanning from https://github.com/rclone/rclone/issues/87#issuecomment-258645510 till https://github.com/rclone/rclone/issues/87#issuecomment-280484648 and compared with API developed by this PR. I have an impression that the API has got the key bits allowing to base yet another backend, the Amazon ACD resume feature upon it:

- ability to set the minimum size requirement for resume
- it's the backend who assigns breakpoint identifiers depending on upstream requirements
- resume id, being an opaque string, can be adapted to the ACD's LocalId/uuid case
- support for intermediate checksums

This does not guarantee that the API and the resume cache machinery will not need fine grinding. On the contrary, adding every other backend to the suit will result in more API refactorings. At least, now I can hope we are on the right way. Nick Craig-Wood for sure will add more comments and  requirements.

The key feature of this PR is that it brings in a passing unit test with it, the first working prototype and consequently a way to play with the feature in practice and watch the "loop closed" from command line to a backend. This will give developers who improve or refactor the interface (add new backends, clarify the logic or implement their favorite programming patterns) a firm ground and a way to check their ideas for sanity.

ivandeex:
This comment is not a requirement but a side note.

I  skimmed through the subthread of #87 spanning from https://github.com/rclone/rclone/issues/87#issuecomment-258645510 till https://github.com/rclone/rclone/issues/87#issuecomment-280484648 and compared with API developed by this PR. I have an impression that the API has got the key bits allowing to base yet another backend, the Amazon ACD resume feature upon it (not in this PR, of course, but in future):

- ability to set the minimum size requirement for resume
- it's the backend who assigns breakpoint identifiers depending on upstream requirements, by means of callback
- resume id, being an opaque string, can be adapted to the ACD's LocalId/uuid case
- support for intermediate checksums

This does not guarantee that the API and the resume cache machinery will not need fine grinding. On the contrary, adding every other backend to the suit will result in more API refactorings. I don't even try to predict what will be introduced by 30+ supported rclone backends. Maybe I'd do if I had a budget and 10+ engineers working on this full-time. Even in that case the YAGNI principle ("don't try to do too much ahead of time") might have stopped me. At least, now I can hope that we are on the right way. Nick Craig-Wood for sure will add more comments.

The key feature of this PR is that it brings in a passing unit test with it, the first working prototype and consequently a way to play with the feature in practice and watch the "loop closed" from command line to a backend. This will give developers who improve or refactor the interface (adding new backends, clarifying the logic or implementing their favorite programming patterns) the firm ground and a way to check their ideas for sanity.

ivandeex:
@mcalman 

1. I squash-rebased this PR on top of master. Please force-pull the tree from github. Now what we have here is just rclone v1.56.0 + single commit.
2. It's time to start beta-testing of the resulting localfs-based prototype and see whether CLI behavior feels consistent. So I published beta build https://github.com/ivandeex/rclone/releases/tag/v1.56.0-resume01
3. Currently your patch in #5185 lives on top of older version of this PR. I guess you might be interested in rebasing to look how they feel together now, e.g. run unit tests with chunker.

ivandeex (approved):

mcalman:
@ncw Will you have time to review this soon? I’ll have limited time to work on this PR after next week. If there are any major revisions that need to be made it would be great to be able to work on them soon.

ivandeex:
just rebased on master to resolve a conflict after recent master commits, please don't force-push.

ivandeex:
@ncw :pray: :pray: :pray:

backend/local/local.go
```diff
@@ -1338,9 +1399,44 @@ func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
...
+	if err != nil {
+		return 0, err
+	}
+	// Hashes match, attempt resume
+	if cachedHash == remoteHash {
+		f.hashState[remote] = hashState
+		return cachedPos, nil
+	}
+	// No valid position found, restart from beginning
+	fs.Debugf(f, "Cached hash state did not match hash state on remote")
```
ncw, line 1432: This should probably be an Info message

```
fs.Infof(remote, "Not resuming as cached hash state did not match hash state on remote")
```

fs/config/configflags/configflags.go
```diff
@@ -131,6 +131,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.DurationVarP(flagSet, &ci.FsCacheExpireDuration, "fs-cache-expire-duration", "", ci.FsCacheExpireDuration, "cache remotes for this long (0 to disable caching)")
 	flags.DurationVarP(flagSet, &ci.FsCacheExpireInterval, "fs-cache-expire-interval", "", ci.FsCacheExpireInterval, "interval to check for expired remotes")
 	flags.BoolVarP(flagSet, &ci.DisableHTTP2, "disable-http2", "", ci.DisableHTTP2, "Disable HTTP/2 in the global transport.")
+	flags.FVarP(flagSet, &ci.MaxResumeCacheSize, "max-resume-cache-size", "", "The maximum size of the cache used to store data necessary for resuming uploads. When the storage grows beyond this size, the oldest resume data will be deleted. (default 100k")
+	flags.FVarP(flagSet, &ci.ResumeLarger, "resume-larger", "", "If set, attempt to resume all partial uploads larger than this size. (default off)")
```
ncw, line 135: `resume-cutoff` would be more in keeping with the other rclone flags?

fs/config/configflags/configflags.go
```diff
@@ -131,6 +131,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.DurationVarP(flagSet, &ci.FsCacheExpireDuration, "fs-cache-expire-duration", "", ci.FsCacheExpireDuration, "cache remotes for this long (0 to disable caching)")
 	flags.DurationVarP(flagSet, &ci.FsCacheExpireInterval, "fs-cache-expire-interval", "", ci.FsCacheExpireInterval, "interval to check for expired remotes")
 	flags.BoolVarP(flagSet, &ci.DisableHTTP2, "disable-http2", "", ci.DisableHTTP2, "Disable HTTP/2 in the global transport.")
+	flags.FVarP(flagSet, &ci.MaxResumeCacheSize, "max-resume-cache-size", "", "The maximum size of the cache used to store data necessary for resuming uploads. When the storage grows beyond this size, the oldest resume data will be deleted. (default 100k")
```
ncw, line 134: This is `maximum number of items` rather than `maximum size` isn't it?

fs/hash/hash.go
```diff
@@ -248,6 +250,67 @@ func (m *MultiHasher) Size() int64 {
 	return m.size
 }
 
+// GetHashState returns the partial hash state for the given hash type encoded as a string
+func (m *MultiHasher) GetHashState(hashType Type) (string, error) {
+	h, ok := m.h[hashType]
+	if !ok {
+		return "", ErrUnsupported
+	}
+	marshaler, ok := h.(encoding.BinaryMarshaler)
```
ncw, line 259: How many of our hashes implement `encoding.BinaryMarshaler`? I don't think any I've written do (dropboxhash, onedrive hash).

fs/open_options.go
```diff
@@ -230,6 +236,150 @@ func (o *HashesOption) Mandatory() bool {
...
+	rootCacheDir := filepath.Join(o.CacheDir, "resume")
+	// Get the Fingerprint of the src object so that future Copy operations can ensure the
+	// object hasn't changed before resuming an upload
+	fingerprint := Fingerprint(ctx, o.Src, true)
+	data, err := marshalResumeJSON(ctx, fingerprint, ID, hashName, hashState)
+	if err != nil {
+		return errors.Wrapf(err, "failed to marshal data JSON")
+	}
+	if len(data) < int(ci.MaxResumeCacheSize) {
+		// Each remote will have its own directory for cached resume files
```
ncw, line 266: We should probably factor the code out of the VFS layer which makes a filename for a cache directory based on an Fs. Its needed fixing sooo many times so re-implementing it doesn't seem like a good idea!

fs/operations/interrupt.go
```diff
@@ -0,0 +1,22 @@
+//+build !windows
```
ncw, line 1: These should probably go in `interrupt_test.go` as they are only used in the tests aren't they?

fs/operations/operations.go
```diff
@@ -475,6 +484,12 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 						if err == nil {
 							newDst = dst
 							err = closeErr
+							// Remove resume cache file (if one was created) when Put/Upload is successful
+							cacheFile := filepath.Join(config.CacheDir, "resume", f.Name(), f.Root(), remote)
```
ncw, line 488: I think this is the incorrect path with the path mangling done when creating the cache dir?

fs/operations/resume.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// Creates an OptionResume that will be passed to Put/Upload
+func createResumeOpt(ctx context.Context, f fs.Fs, remote string, src fs.Object) (resumeOpt *fs.OptionResume) {
+	ci := fs.GetConfig(ctx)
+	resumeOpt = &fs.OptionResume{ID: "", Pos: 0, Src: src, F: f, Remote: remote, CacheCleaned: false, CacheDir: config.CacheDir}
+	if ci.ResumeLarger >= 0 {
+		root := f.Root()
+		if runtime.GOOS == "windows" {
+			if root[:4] == "//?/" {
```
ncw, line 23: Duplicated cache path mangling!

ncw (commented):
This looks very nice now :-)

I put a few comments inline for you to look at.

Let's try to get this merged for 1.57 :-)

fs/config/configflags/configflags.go
ivandeex: I recall I was asking to limit by size, not by items. AFAIC recall the author satisfied the request.

fs/hash/hash.go
ivandeex: stdlib hashes do like md5, sha1, sha256

ivandeex: I will submit a pr for mrhash

fs/operations/operations.go
ncw: Once per file seems OK thinking some more about it since we are only going to activate this for large files probably which take lots of time. I was thinking we'd be doing it for every file which might be 100/s but I think that is incorrect.

lib/file/preallocate_unix.go
```diff
@@ -47,7 +47,6 @@ func PreAllocate(size int64, out *os.File) (err error) {
 			// Try the next flags combination
 			index++
 			atomic.StoreInt32(&fallocFlagsIndex, index)
-			fs.Debugf(nil, "preAllocate: got error on fallocate, trying combination %d/%d: %v", index, len(fallocFlags), err)
```
mcalman, line 50: I removed this for now because there was an import cycle I couldn't seem to avoid with the `cacheroot`, `file`, and `fs` packages. If there is a way to avoid this or a better place to define the `CreateCacheRoot` function please let me know.

fs/hash/hash.go
ivandeex: ... and for dropbox/onedrive hashes after this pr gets merged

lib/file/preallocate_unix.go
ivandeex: `lib/file` is being modified concurrently by pr https://github.com/rclone/rclone/pull/5401#issuecomment-913845588
and needs attention of @albertony @ncw

ivandeex: probably we could introduce an interface for adhoc open option?

fs/hash/hash.go
ivandeex: Captured in https://github.com/rclone/rclone/issues/5624

mcalman:
@ncw @ivandeex Do we think we can still include this and #5185 in 1.57? Is there anything else that needs to be done?

ivandeex:
I just rebased this on the git tip, resolved merge conflicts and remade the cache root tefactoring after recent changes on the master.
It all depends on @ncw review now.

ivandeex:
I rebased this on the git tip, resolved merge conflicts, fixed order of CreateCacheRoot arguments in fs/operations.go, remade the cache root refactoring to follow os-safe cache dirs after recent changes on the master.
p
It all depends on @ncw review now.

ivandeex:
I rebased this on the git tip, resolved merge conflicts, fixed order of CreateCacheRoot arguments in fs/operations.go, remade the cache root refactoring to follow os-safe cache dirs after recent changes on the master.

please clean up commit messages and author emails.

all depends on @ncw review now.

ivandeex:
imho this is ready for 1.57

ivandeex:
@mcalman Yet another rebase on new master and resolved conflicts, don't force push..

@ncw PTAL!

ivandeex:
@ncw
For extra safety, we could use the release strategy similar to bisync:
merge this _right after_ 1.57,
let beta build users test it and submit patches,
and finally have it in 1.58 ofiicially.

ivandeex:
@ncw
For extra safety, we could use the release strategy similar to bisync:
- merge this _right after_ 1.57,
- let beta build users test it and submit patches (include mine above and the chunker-resume pr),
- and finally have it in 1.58 ofiicially.

innovate-invent:
My objections to this PR seem to have been collapsed under loads of commit messages.
@ncw please take them into consideration before merging
https://github.com/rclone/rclone/pull/4547#issuecomment-831718386

ivandeex:
This "loads of commits and messages" is called working on your PR.

ivandeex:
@mcalman

This PR is waiting for @ncw to review
- #5299

which is a spin off from
- #3850

which is stuck forever. 
Unfortunately that PR is pretty far from production and invasive on core parts of rclone.
Reviewing by reading the code it is not simple and not enough as it like this PR changes code in the rclone core.
The process will definitely include writing tests and ensuring that current functionality does not get broken.

As far as I can see @ncw actively discussed early design of that PR.
Currently waiting for @ncw to find time for review.
If it's not done by next January, I am going to untangle the knot myself.
Probably I will combine code from both proposals, add appropriate tests and submit as new patch.
In the meantime I have added the patch from
- #5185

on top of this one - at any rate I will consider all parts together.

I do not expect the solution to come before the 1.58 release so I expect the resume feature to appear in 1.59.


@innovate-invent This patch will not be merged on rclone master branch before that, which is at least 3 months from now.

However, the patch of this PR has already solved me hours of work time by resuming downloads ("uploads" onto local backend) failed due to network hiccups.
In the meantime I created a branch and a beta release to use this code in practice.
The beta releases will be available at https://beta.rclone.org/branch/resume/

@mcalman By the way, you can open pull requests against the beta branch https://github.com/rclone/rclone/commits/resume

cc @ncw

ivandeex:
@mcalman I am sorry about the situation

ivandeex:
The PR branch as well as https://beta.rclone.org/branch/resume/ is rebased on the current master.

maxcalman:
> If it's not done by next January, I am going to untangle the knot myself.

@ivandeex Was just wondering what your status is. Do you still think it is likely the resume feature will be included in 1.59?

ivandeex:
I'm hit heavily by anti-Rus sanctions, plans broken. Most of the short-term time I'll be recovering whatever I can w.r.t. work env, accounts and payments etc. Can't promise anything concrete a.t.m.

ncw:
Hi @mcalman, and thank you for this, and for the substantial design discussion that went into it.

I'm doing a tidy-up of long-inactive PRs, and this one has been quiet for around three years. The resumable-upload interface is still an interesting idea, but the code has drifted a long way from current `master` and the design thread never quite converged, so I'm going to close it to keep the queue manageable.

This really isn't a reflection on your effort, as there was a lot of good thinking here. If you (or anyone reading) would like to revive resumable uploads, a fresh PR or a [forum](https://forum.rclone.org/) discussion to settle the design would be very welcome. Thanks again for the work.

PR #4549, opened 2020-09-01
webui: Prompt user for updating webui if an update is available

negative0:

This change will check if there is an available update to the webui and notify the user about it.

No
#### Checklist

### DIFF
--- a/fs/rc/webgui/webgui.go
+++ b/fs/rc/webgui/webgui.go
@@ -59,17 +59,30 @@ func CheckAndDownloadWebGUIRelease(checkUpdate bool, forceUpdate bool, fetchURL 
 		return errors.New("Web GUI path exists, but is a file instead of folder. Please check the path " + extractPath)
 	}
 
+	// Get the latest release details
+	WebUIURL, tag, size, err := GetLatestReleaseURL(fetchURL)
+	if err != nil {
+		return errors.Wrap(err, "Error checking for web gui release update, skipping update")
+	}
+	dat, err := ioutil.ReadFile(tagPath)
+	tagsMatch := false
+	if err != nil {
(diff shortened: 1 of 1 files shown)


fs/rc/webgui/webgui.go
```diff
@@ -55,17 +55,25 @@ func CheckAndDownloadWebGUIRelease(checkUpdate bool, forceUpdate bool, fetchURL
...
 
+	// Get the latest release details
+	WebUIURL, tag, size, err := GetLatestReleaseURL(fetchURL)
+	if err != nil {
+		return errors.Wrap(err, "Error checking for web gui release update, skipping update")
+	}
+	dat, err := ioutil.ReadFile(tagPath)
+
+	// Check if current version is equal to latest release
+	tagsMatch := err == nil && string(dat) == tag
```
ncw, line 66: I think the error handling is wrong here...

If err != nil then you'll declare that the tags don't match and you'll print the error

fs/rc/webgui/webgui.go
```diff
@@ -55,17 +55,25 @@ func CheckAndDownloadWebGUIRelease(checkUpdate bool, forceUpdate bool, fetchURL
...
+
+	if !tagsMatch {
+		fs.Infof(nil, "A release (%s) for gui is present at %s. Use --rc-web-gui-update to update", tag, WebUIURL)
```
ncw, line 69: I think this should probably show your version tag and the new version tag so the user can gauge whether it is worth doing

ncw (commented):
See inline :-)

fs/rc/webgui/webgui.go
negative0: On it

ncw:
Looks great - thank you :-)

I'll merge it now.

Issue #2296, opened 2018-05-08
Bad Request (400) for file modified before Unix epoch (1970) [Box]


xaque:
I had a folder of files with modified dates ranged from 1930 to 1980. All of them before 1970 were failing and I was able to figure out it was related to the Unix epoch.

To reproduce:

`touch pre.epoch -t 196912310000`
`touch post.epoch -t 197001010000`

`rclone copy pre.epoch box-remote:`
`rclone copy post.epoch box-remote:`

The pre.epoch upload will fail with:
`ERROR : pre.epoch: Failed to copy: Error "bad_request" (400): Bad Request`
but the post.epoch upload will succeed.

This happens for me on my box remote (and box crypt remote). I tested it on Google Drive and there was no failure. I don't have any other remotes to test it on, so it may be more than just Box.

...

```
$ rclone --version
rclone v1.41
- os/arch: linux/amd64
- go version: go1.10.1
```
```
$ uname -a
Linux my-machine 4.16.6-1-ARCH #1 SMP PREEMPT Mon Apr 30 12:30:03 UTC 2018 x86_64 GNU/Linux
```

xaque:
After a little more research into it, it may be a box bug and unrelated to rclone. I assume that would mean that timestamps in a crypt remote are not encrypted/obfuscated since the issue happens in a box crypt remote as well.

ncw:
Yes I think this is a problem with box...  I had a look through their docs and I could find anything related.

Below is the result of the failing transaction with `-vv --dump bodies`.

Note what rclone sent `1944-01-01T01:01:00+01:00`

And the response from box `"Invalid value '-820540740'. Value must be greater than or equal to 0"` which implies that box converted that time into a `time_t` and then rejected it because it was before the epoch 1970-01-01.

I tried sending the file to drive and it was OK.  There maybe other remotes which do have a problem - I didn't do an exhaustive survey yet...

What do you think rclone should do here?  I could put a flag in `--local-truncate-old-times` say, which would make the local disk read timestamps before 1970 as 1970-01-01. The rest of rclone would have to treat that date specially and ignore it so it isn't completely trivial...

Or maybe I could get the box backend to put dates before 1970 as 1970-01-01 and make that date be ignored in the syncing routines so files with that date will get synced as size only...

[...]

PR #4552, opened 2020-09-02
opendrive: Do not retry 400 errors

eharris:
This type of error is unlikely to be an error that can be resolved by a retry, and is triggered in #2296 by files with a timestamp before the unix epoch.  This does not address the underlying timestamp issue that #2296 describes, only avoids that issue triggering useless retries.

OpenDrive previously retried 400 errors, but that should probably not be the case, and at least in the case of #2296 it is quite detrimental.

#2296 

#### Checklist

### DIFF
--- a/backend/opendrive/opendrive.go
+++ b/backend/opendrive/opendrive.go
@@ -646,7 +646,6 @@ func (f *Fs) Put(ctx context.Context, in io.Reader, src fs.ObjectInfo, options .
 
 // retryErrorCodes is a slice of error codes that we will retry
 var retryErrorCodes = []int{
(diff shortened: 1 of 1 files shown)


eharris:
Even with this change, there is a higher level of retry (3 times, in `cmd/cmd.go`) that probably should be avoided.  Not sure the best way to do that.

ncw:
> Even with this change, there is a higher level of retry (3 times, in `cmd/cmd.go`) that probably should be avoided. Not sure the best way to do that.

You can wrap the error in an `fserrors.NoRetryError` error. However I don't think current behaviour is too unreasonable.

I'll merge this now.

PR #4566, opened 2020-09-03
Experimental compression remote - third attempt

buengese:
Let's give this another try. In the last few days I've once again reworked the experimental compression remote started by id01 and continued by me a while ago. I've never been quite happy with the original - it just seemed to be to complicated and trying to do too many things as result this third attempt is now much smaller scale.
Differences from the previous attempts.

- only gzip compression. the entire compression interface is now gone though I might refactor this a bit further to at least retain the ability to add other compression backends
- my own custom gzip library sgzip forked from klauspost's excellent pgzip that allows for seeking in the compressed data and takes care of multi threading. all of this was previously done inside rclone but I think a dedicated library with it's own rather extensive testsuite is better suited to this task.
- the integration tests actually fully pass this time.

Like the previous attempts this is still in the work in progress stage. But I'm opening this now in case someone is willing to test it.


ncw:
> Let's give this another try. In the last few days I've once again reworked the experimental compression remote started by id01 and continued by me a while ago. I've never been quite happy with the original - it just seemed to be to complicated and trying to do too many things as result this third attempt is now much smaller scale.

That seems like a great idea to me!

> Differences from the previous attempts.
> 
> * only gzip compression. the entire compression interface is now gone though I might refactor this a bit further to at least retain the ability to add other compression backends

Is it 100% compatible gzip?

> * my own custom gzip library sgzip forked from klauspost's excellent pgzip that allows for seeking in the compressed data and takes care of multi threading. all of this was previously done inside rclone but I think a dedicated library with it's own rather extensive testsuite is better suited to this task.

Nice - yes that seems like a good solution.

You might want to upstream your changes to pgzip? Did you discuss with Klaus?

> * the integration tests actually fully pass this time.

Yipee!

> Like the previous attempts this is still in the work in progress stage. But I'm opening this now in case someone is willing to test it.

Promote it on the forum?

buengese:
> Is it 100% compatible gzip?

Yes! The compression creates 100% standard gzip files no special sauce. 

> Nice - yes that seems like a good solution. You might want to upstream your changes to pgzip? Did you discuss with Klaus?

Not yet. The changes are not yet completely final, I consider them beta right know. More importantly I don't believe these changes are upstreamable. Pgzip has a very clear goal - being a faster (multithreaded) drop in replacement for compress/gzip. While the changes technically retain this ability and the compression part barely changed at all it still introduces some overhead that I'd see contrary to the goal of pgzip.  
The decompression code is still designed to use an external file stroring block boundaries to achieve seeking (and more importantly reading from an offset). This design is very much tailored to rclone's use case. If you are not limited to an http stream and can freely jump around in the compressed file like it is the case with a file on disk there are potentially better ways to do this (and have done) that e.g. don't require an additional file. So even if one were to extend pgzip's scope to seekable gzip compression this isn't really a great fit.

> Promote it on the forum?
Yes! As soon as I've done a final cleanup pass on this.

One think I forgot to mention in the first post is that this implementation obviously still retains one very ugly limitation in that the uncompressed file sizes are stored in the names of the compressed files. I really, really hate this but it's only way to have accurate file sizes without requiring List to download metadata and parse metadata files.

ncw:
> One think I forgot to mention in the first post is that this implementation obviously still retains one very ugly limitation in that the uncompressed file sizes are stored in the names of the compressed files. I really, really hate this but it's only way to have accurate file sizes without requiring List to download metadata and parse metadata files.

Yes I see what you mean. It would be possible (but a long way from ideal!) to report the size as unknown `-1`.

Hmm, unless we implement a general metadata service for rclone... Which is certainly something I've been putting background cycles into thinking about. Just about all the backends can store metadata of some type (even if it is only a description field).

buengese:
I'm pretty much done with this for now. What do you should I make beta and post in the forum?

ncw:
> I'm pretty much done with this for now (of course i still consider this experimental). What do you should I make beta and post in the forum?

Yes let's get some real world testing - if you push it to a branch on the main repo then it will make a beta.

I was thinking of making all pull requests by rclone team members to make a beta automatically. If I can figure that out then that would be quite useful I think.  So if an rclone team member sends PR #4566 then this will make a beta with tag `pr-4566` - what do you think? I think that would help with the workflow of fixing an issue and getting the user to test it.

I'll give your code a look now.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+	objErr := o.Object.Remove(ctx)
+	if err != nil {
+		return err
+	}
+	return objErr
+}
+
+// ReadCloserWrapper combines a Reader and a Closer to a ReadCloser
+type ReadCloserWrapper struct {
+	dataSource io.Reader
```
ncw, line 1012: If you embed both of these

```go
type ReadCloserWrapper struct {
    io.Reader
    io.Closer
}
```

Then you don't need to implement the `Read` and `Close` methods.

You initialise with `ReadCloserWrapper{Reader: r, Closer: c}`

We implement this at least once elsewhere!

https://github.com/rclone/rclone/blob/d8239e01941b69fcdef2a62fad8791c393a058ec/fs/operations/operations.go#L1007-L1011

So maybe it should go into `lib/readers`, however it is a very small adaptor so I'm not sure it is worth it.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+}
+
+// Hash returns the selected checksum of the file
+// If no checksum is available it returns ""
+func (o *Object) Hash(ctx context.Context, ht hash.Type) (string, error) {
+	err := o.loadMetadataIfNotLoaded(ctx)
+	if err != nil {
+		return "", err
+	}
+	if ht&hash.MD5 == 0 {
```
ncw, line 1170: `ht hash.Type` isn't used as bitmap here so usually `ht == hash.MD5` is what we use.

You can put this first before the `loadMetadataIfNotLoaded` if you want to save loading metadata when we couldn't return a hash.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+	return o.src.ModTime(ctx)
+}
+
+// Hash returns the selected checksum of the file
+// If no checksum is available it returns ""
+func (o *ObjectInfo) Hash(ctx context.Context, ht hash.Type) (string, error) {
+	if ht != hash.MD5 {
+		return "", hash.ErrUnsupported
+	}
+	if o.Size() != o.src.Size() {
```
ncw, line 1308: I don't understand the logic here?

ncw (commented):
I've given that a first pass review - see comments inline.

I think it is looking nice!

Next thing is for me to play with it some more :-)

docs/content/press.md
```diff
@@ -0,0 +1,82 @@
...
+title: "Press"
+description: "Compression Remote"
+date: "2020-09-14"
+---
+
+Press (Experimental)
+-----------------------------------------
+
+The `press` remote adds compression to another remote. It is best used with remotes containing
+many large compressible files or on top of other remotes like crypt.
```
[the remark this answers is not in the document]
buengese: Definitely true.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+		WriteMimeType:           false,
+		GetTier:                 true,
+		SetTier:                 true,
+		BucketBased:             true,
+		CanHaveEmptyDirectories: true,
+	}).Fill(f).Mask(wrappedFs).WrapsFs(f, wrappedFs)
+	// We support reading MIME types no matter the wrapped fs
+	f.features.ReadMimeType = true
+	// We can only support putstream if we have serverside copy or move
+	if wrappedFs.Features().Move == nil && wrappedFs.Features().Copy == nil {
```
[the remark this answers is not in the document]
buengese: Didn't now of this yet. Definitely makes this cleaner.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+	if err != nil && err != io.EOF {
+		return nil, "", err
+	}
+	mime := mimetype.Detect(b.Bytes())
+	in = io.MultiReader(bytes.NewReader(b.Bytes()), in)
+	return wrap(in), mime.String(), nil
+}
+
+func isCompressible(mime string) bool {
+	return !unconpressibleRegexp.MatchString(mime)
```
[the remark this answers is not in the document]
buengese: This used to be what was implemented here initially. I'm not sure when and why I changed that I don't think there was ever a problem with that so I'll bring that back.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+	}
+	return o, newMetadata(o.Size(), Uncompressed, sgzip.GzipMetadata{}, sum, mimeType), nil
+}
+
+// This function will write a metadata struct to a metadata Object for an src. Returns a wrappable metadata object.
+func (f *Fs) putMetadata(ctx context.Context, meta *ObjectMetadata, src fs.ObjectInfo, options []fs.OpenOption, put putFn, verifyCompressedObject bool) (mo fs.Object, err error) {
+	// Generate the metadata contents
+	var b bytes.Buffer
+	gzipWriter := gzip.NewWriter(&b)
+	metadataEncoder := gob.NewEncoder(gzipWriter)
```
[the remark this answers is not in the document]
buengese: This goes back to id01 original implementation. I think that choice was made for performance reasons. These metadata files will contain an array with <= source file size / 1 MiB entries. As long as we aren't dealing with huge files JSON is probably fine.

backend/press/press.go
buengese: I think thats to small to but put into `lib/readers`

backend/press/press.go
buengese: No idea either. I don't think that was intentional.

backend/press/press.go
```diff
@@ -0,0 +1,1376 @@
...
+
+	// If verifyCompressedObject is on, find a hash the destination supports to compute a hash of
+	// the compressed data.
+	ht := f.Fs.Hashes().GetOne()
+	var hasher *hash.MultiHasher
+	var err error
+	if ht != hash.None && verifyCompressedObject {
+		// unwrap the accounting again
+		wrappedIn, wrap = accounting.UnWrap(wrappedIn)
+		hasher, err = hash.NewMultiHasherTypes(hash.NewHashSet(ht))
```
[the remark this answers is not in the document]
buengese: Yeah that would be nice.

buengese:
@ncw I've addressed all your comments now and also fixed a few more things I noticed while looking over the code again. 
- Now using the same 1MiB of data we use to determine mime time to check if the file is compressible. Currently I've hard coded 1.1 as the minimum compression ratio - not sure about that value yet. Maybe that should be config option.
- Switched to json for the metadata and also dropped the gzip compression of the metadata. Not entirely convinced but it seemed to perform pretty well.
- Also added one more experimental commit that changes the file name format to use base64 (url encoding without padding) for the file size. That should save a few more characters and I don't think that should cause any problems.

segator:
this feature is pretty cool!! and I would love to see the metadata service @ncw  mention. this meta service also will help for future chunking backends, allowing inline dedup/snapshot and other cool features.

docs/content/compress.md
```diff
@@ -0,0 +1,136 @@
...
+Currently only gzip compression is supported, it provides a decent balance between speed and strength and is well
+supported by other application. Compression strength can further be configured via an advanced setting where 0 is no
+compression and 9 is strongest compression.
+
+#### Filetype
+If you open a remote wrapped by press, you will see that there are many files with an extension corresponding to
+the compression algorithm you chose. These files are standard files that can be opened by various archive programs, 
+but they have some hidden metadata that allows them to be used by rclone.
+While you may download and decompress these files at will, do **not** upload any compressed files to a wrapped remote
+through any other means than rclone. This will upload files that do not contain metadata and **will** cause unexpected behavior.
```
ncw, line 77: What happens? :-)

ncw (commented):
I think this is ready to be merged - what do you think

Are you happy with the metadata format? Is it extensible enough for future needs?

If you are ready, can you squash it into 2 commits (one for hash one for press) and rebase?

I'm very shortly going to add more context.Context into the backend interface (into NewFs and ConfigFs) so depending on the order of things it might need an update.

docs/content/compress.md
buengese: If I didn't break anything they should in fact just be ignored. (I haven't tested this in a long time though)

buengese:
Hopefully final rebase done.

ncw:
That looks good to me - well done!

I'll leave it to you to press the merge button :-)

segator:
congratulations!

Ornias1993:
Interesting work, shame you guys have gone for gzip instead of the way more efficient zstd algorithm.

buengese:
We're always open to improvements I did consider using zstd but the performance of the go implementation was still somewhat lacking compared to the gain in compression ratio so I decided against it for now.

Issue #4529, opened 2020-08-24
TestParseDuration fails on Linux


negative0:
#### What is the problem you are having with rclone?

While running `make quicktest`, the test parse duration fails every time. @ncw has pointed out that the problem is with time zone.
```
TZ=Asia/Kolkata go test -v -run TestParseDuration
```
Fails while the following passes,
```
TZ=Europe/London go test -v -run TestParseDuration
```

#### What is your rclone version (output from `rclone version`)

rclone v1.52.2-269-g05703ba2-plugins2-beta
- os/arch: linux/amd64
- go version: go1.14.4

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Pop OS! 64 bits

####  Which cloud storage system are you using? (eg Google Drive)
NA

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
`make quicktest`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
--- FAIL: TestParseDuration (0.00s)
    parseduration_test.go:53: 
        	Error Trace:	parseduration_test.go:53
        	Error:      	Should be true
        	Test:       	TestParseDuration
        	Messages:   	2001-02-03
    parseduration_test.go:53: 
        	Error Trace:	parseduration_test.go:53
        	Error:      	Should be true
        	Test:       	TestParseDuration
        	Messages:   	2001-02-03 10:11:12
    parseduration_test.go:53: 
        	Error Trace:	parseduration_test.go:53
        	Error:      	Should be true
        	Test:       	TestParseDuration
        	Messages:   	2001-02-03T10:11:12
FAIL
```

PR #4573, opened 2020-09-07
Fix test failure in different local time zones

Zacho2:
To fix a broken unit test.

TestParseDuration relied on an elapsed time calculation which
would vary based on the system local time. Fix the test by not relying
on the system time location. Also make the test more deterministic
by injecting time in tests rather than using system time.

#4529

#### Checklist

### DIFF
--- a/fs/parseduration.go
+++ b/fs/parseduration.go
@@ -88,8 +88,9 @@ func parseDurationDates(age string, epoch time.Time) (t time.Duration, err error
 	return t, err
 }
 
-// ParseDuration parses a duration string. Accept ms|s|m|h|d|w|M|y suffixes. Defaults to second if not provided
-func ParseDuration(age string) (d time.Duration, err error) {
+// parseDurationFromNow parses a duration string. Allows ParseDuration to match the time
+// package and easier testing within the fs package.
+func parseDurationFromNow(age string, getNow func() time.Time) (d time.Duration, err error) {
 	if age == "off" {
 		return time.Duration(DurationOff), nil
 	}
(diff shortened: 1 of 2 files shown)


ncw (commented):
Thank you for doing this :-)

ParseDuration is supposed to mimic the `time.ParseDuration` interface so adding another parameter to it isn't ideal.

I guess we could have a `var parseDurationTimeNow = time.Now` and monkey patch that in the tests?

Or make a `parseDuration` which takes the extra parameter and test that.

What do you think?

ivandeex:
@Zacho2 Are you still with us?

> ParseDuration is supposed to mimic the `time.ParseDuration` interface so adding another parameter to it isn't ideal.
> I guess we could have a `var parseDurationTimeNow = time.Now` and monkey patch that in the tests?
> Or make a `parseDuration` which takes the extra parameter and test that.

Zacho2:
@ncw imo the `time` package doesn't export an interface for `time.ParseDuration` so we don't need to follow the function signature it to a T. Monkey`ing or a using a package global seems less than ideal to me. My patch minimizes `time.Now()` calls by only calling when the duration can't be parsed by the `time` package.

I could create a `parseDuration` function in package `fs`, but it seems like it would have to call `time.Now()` for all duration strings. Is this a concern?

ncw:
>  imo the `time` package doesn't export an interface for `time.ParseDuration` so we don't need to follow the function signature it to a T.

I would prefer not to add an extra parameter to `fs.ParseDuration` as it is supposed to be a drop in replacement for `time.ParseDuration`.

> Monkey patching or a using a package global seems less than ideal to me. My patch minimizes `time.Now()` calls by only calling when the duration can't be parsed by the `time` package.

Noted

> I could create a `parseDuration` function in package `fs`, but it seems like it would have to call `time.Now()` for all duration strings. Is this a concern?

How about creating a `ParseDurationFromNow` (say) which takes the `func() time.Time` pointer and we make a simple wrapper for `ParseDuration` so we don't have to change its interface. The tests can test `ParseDurationFromNow`. This keeps the efficiency of not calling time.Time where necessary and keeps the interface of `ParseDuration` the same.

Zacho2:
Thanks @ncw. I've implemented your suggestion.

ncw (approved):

I'll merge it now.

PR #4597, opened 2020-09-15
Fix cache reset.1.53.1

leoluan1:
#### Checklist

### DIFF
--- a/vfs/vfscache/item.go
+++ b/vfs/vfscache/item.go
@@ -1167,6 +1167,7 @@ func (item *Item) setModTime(modTime time.Time) {
 // ReadAt bytes from the file at off
 func (item *Item) ReadAt(b []byte, off int64) (n int, err error) {
 	n = 0
+	var expBackOff int
 	for retries := 0; retries < fs.Config.LowLevelRetries; retries++ {
 		item.preAccess()
 		n, err = item.readAt(b, off)
@@ -1180,6 +1181,12 @@ func (item *Item) ReadAt(b []byte, off int64) (n int, err error) {
 			break
 		}
 		item.c.KickCleaner()
+		expBackOff = 2 << uint(retries)
+		time.Sleep(time.Duration(expBackOff) * time.Millisecond) // Exponential back-off the retries
+	}
+
+	if fserrors.IsErrNoSpace(err) {
+		fs.Errorf(item.name, "vfs cache: failed to _ensure cache after retries %v", err)
 	}
 
 	return n, err
--- a/vfs/vfscache/item.go
+++ b/vfs/vfscache/item.go
@@ -43,6 +43,13 @@ import (
 // be taken before Item.mu. writeback may call into Item but Item may
 // **not** call writeback methods with Item.mu held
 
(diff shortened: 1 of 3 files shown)


vfs/vfscache/item.go
```diff
@@ -414,6 +428,12 @@ func (item *Item) _dirty() {
 
 // Dirty marks the item as changed and needing writeback
 func (item *Item) Dirty() {
+	item.preAccess()
```
ncw, line 431: Rather than splitting the function, I think you could be using

```go
item.preAccess()
defer item.postAccess()
```

Which means `item.postAccess()` will be executed regardless of exit point (and also if a panic happens)

This is a much more Go way of doing it I think.

You might want to do this for all the functions as it does have the extra (very minor in this case) advantage of being run during a panic.

Note that in go1.15 `defer` calls are almost free - the compiler does static analysis and effectively turns them into old C-style `goto out;` error handling behind the scene.

vfs/vfscache/item.go
```diff
@@ -1197,7 +1226,16 @@ func (item *Item) readAt(b []byte, off int64) (n int, err error) {
 }
 
 // WriteAt bytes to the file at off
+// Protected against Reset using preAccess/postAccess
 func (item *Item) WriteAt(b []byte, off int64) (n int, err error) {
+	item.preAccess()
```
ncw, line 1231: No need to split the function here either

ncw (commented):
I think the first two commits are perfect.

See comments on the third.

Can you also rename the commits slightly, prefixing them with `vfs: XXX` - this means that when I make the change log they will go into the correct section.

See: https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages

Thank you!

vfs/vfscache/item.go
leoluan1: Thanks for the comments.  They make sense.  I will rework this commit accordingly.

vfs/vfscache/item.go
leoluan1: Fixed

vfs/vfscache/item.go
leoluan1: This commit has been redone according to the suggestion.

leoluan1:
@ncw Can this PR be merged now so that they can be part of 1.53.2?  Since these small fixes address some critical functional issues, we hope they can be included sooner.

ncw:
Note to fix the go1.11 and go1.12 builds you need a cast to uint

    vfs/vfscache/item.go:1199:18: invalid operation: 2 << retries (shift count type int, must be unsigned integer)


> Can this PR be merged now so that they can be part of 1.53.2? Since these small fixes address some critical functional issues, we hope they can be included sooner.

If you fix up the above I'll merge to master and the stable branch :-)

Thank you!

leoluan1:
Fixed the go1.11 and go1.12 errors with the uint type cast.  Thanks!

ncw:
Thanks for that!

I'll merge this now. I'll also queue the patches for the stable release.

Thank you very much :-)

Issue #4600, opened 2020-09-15
Avoid creating web gui configuration file on startup


albertony:
#### What is the problem you are having with rclone?

Rclone creates web gui configuration file on startup, even with a simple `rclone --version` command.

It is created in a subfolder path below cache folder, given by global flag `--cache-dir`. Default on Windows: `%LocalAppData%\rclone\webgui\plugins\config\availablePlugins.json`.

To me this seems unnecessary, at least on a setup where web gui is never going to be used anyway. It is a bit "annoying" when rclone otherwise is set up to run "portable" with a local configuration file given by global flag `--config`. I could redirect it by setting `--cache-dir`, but it would it not be better if the webgui/plugin related directories/files were not created until actually needed?

I considered trying to change this in the code myself, but noticed there was a commit specifically introducing it in the original pull request from @negative0: [Create availablePlugins config file if it does not exist](https://github.com/rclone/rclone/pull/4309/commits/b2d3ff8067b70098fb9e20bf6f0285e472e8bb53), so there is probably some reason for it that must be considered...

#### What is your rclone version (output from `rclone version`)

rclone v1.53.1
- os/arch: windows/amd64
- go version: go1.15

darthShadow:
@negative0 Thoughts?

negative0:
@darthShadow Thanks for pointing this out. I think the commit should fix it. Let me know.

darthShadow:
Yes, that should fix it. Can you please create a PR so @albertony has a build to try that out?

negative0:
Yes. I was just doing that.

Issue #4592, opened 2020-09-14
Error reading available plugins when running as a user without a home directory


mmazerolle:
#### What is the problem you are having with rclone?

When running rclone as a user without a home directory, the first line reports an error trying to load plugins.
Example:
```
pi@quasar:~ $ sudo -u rclone rclone lsd --config /etc/rclone/rclone.conf Dropbox:
2020/09/14 12:33:40 ERROR : error reading available plugins: Error creating : /home/rclone/.cache/rclone/webgui/plugins/config
```

In this case, "rclone" is a user without password or home directory, who will issue rclone commands using the configuration file passed on the command line (in a cron job, for example).

#### What is your rclone version (output from `rclone version`)
rclone v1.53.1
- os/arch: linux/arm
- go version: go1.15

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Raspbian

####  Which cloud storage system are you using? (eg Google Drive)
Dropbox

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
`sudo -u rclone rclone lsd --config /etc/rclone/rclone.conf Dropbox:`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

2020/09/14 13:52:55 ERROR : error reading available plugins: Error creating : ```
/home/rclone/.cache/rclone/webgui/plugins/config
2020/09/14 13:52:55 DEBUG : rclone: Version "v1.53.1" starting with parameters ["rclone" "lsd" "-vv" "--config" "/etc/rclone/rclone.conf" "Dropbox:"]
2020/09/14 13:52:55 DEBUG : Using config file from "/etc/rclone/rclone.conf"
2020/09/14 13:52:55 DEBUG : Creating backend with remote "Dropbox:"
[correct listing of Dropbox]
2020/09/14 13:52:56 DEBUG : 3 go routines active
```

**Comment by mmazerolle** [comment]
Expectation is that rclone does not generate ERROR log entries for default behaviour when passed well-formed config files.

**Comment by ncw** [comment]
Yes that looks like a bug.

We shouldn't be reading anything about the plugins if we haven't initialized the webgui.

@negative0 ?

**Comment by negative0** [comment]
On it!

**Comment by minhphuc429** [comment]
had the same issue

**Comment by doob187** [comment]
here the same
```
2020/09/21 22:19:16 ERROR : error reading available plugins: Error creating : /root/.cache/rclone/webgui/plugins/config
```
[...]

PR #4606, opened 2020-09-17
plugins: Create plugins files only if webui is enabled.

negative0:

Create plugins related files only if webui is enabled.

#4600  #4592 

#### Checklist

### DIFF
--- a/fs/rc/webgui/plugins.go
+++ b/fs/rc/webgui/plugins.go
@@ -15,6 +15,8 @@ import (
 
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config"
+	"github.com/rclone/rclone/fs/rc/rcflags"
+	"github.com/rclone/rclone/lib/errors"
 )
 
 // PackageJSON is the structure of package.json of a plugin
@@ -62,6 +64,8 @@ var (
 	PluginsPath              string
 	pluginsConfigPath        string
 	availablePluginsJSONPath = "availablePlugins.json"
+	initSuccess              = false
+	initMutex                = &sync.Mutex{}
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


fs/rc/webgui/plugins.go
```diff
@@ -65,14 +67,16 @@ var (
 )
 
 func init() {
-	cachePath = filepath.Join(config.CacheDir, "webgui")
-	PluginsPath = filepath.Join(cachePath, "plugins")
-	pluginsConfigPath = filepath.Join(PluginsPath, "config")
+	if rcflags.Opt.WebUI {
```
ncw, line 70: I'm not sure this will work as I don't think `rcflags.Opt` will have been initialised when the `init()` method is called. Did you try it and see what happened?

ncw (commented):
Looking good!

See query inline - thanks :-)

fs/rc/webgui/plugins.go
negative0: Ahh, yes. After I removed the directories and tried again, it is not working. 
Any possible workaround for this?

negative0: Should we have a different flag for enabling plugins? As during testing, we don't want the webui to be instantiated, but the plugins should be tested.

ncw: I don't think you can read flags at that point... You can use a `sync.Once` to initialise it elsewhere if you want?

negative0:
@ncw. Do you think this will work? Or we can try something else

ncw:
I think that approach looks great - well done :-)

Can you fix up the tests? `--- FAIL: TestAddPlugin` is failing?

ncw:
@negative0 can you fix this up? We should merge it I think :-)

negative0:
I will do that ASAP. :)

negative0:
@ncw This should fix the original error, but some unrelated tests are failing now.

ncw:
I just rebased this and fixed goimports - hopefully it will go green now!

Issue #4584, opened 2020-09-11
Mail.ru Cloud upload http 403


ruatag:
#### What is the problem you are having with rclone?
Mail.ru Cloud upload operation (copy, sync) failed with 403
download from server (copy,sync), mkdir, ls works correct
First appears 10 Sep 2020 before it worked for a long time
Probably something changed from server side
#### What is your rclone version (output from `rclone version`)
```
rclone v1.53.0
- os/arch: linux/amd64
- go version: go1.15
```
latest beta - same problem
```
rclone v1.54.0-beta.4766.75de30cfa
- os/arch: linux/386
- go version: go1.15
```
#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
ArchLinux x64 5.8.7-arch1-1
####  Which cloud storage system are you using? (eg Google Drive)
Mail.ru Cloud
#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
```
rclone copy rclone-v1.53.0-linux-amd64.zip list:/backup/
```
#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
2020/09/11 10:54:34 DEBUG : rclone: Version "v1.53.0" starting with parameters ["rclone" "-vv" "copy" "rclone-v1.53.0-linux-amd64.zip" "list:/backup/"]
2020/09/11 10:54:34 DEBUG : Creating backend with remote "rclone-v1.53.0-linux-amd64.zip"
2020/09/11 10:54:34 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2020/09/11 10:54:34 DEBUG : fs cache: adding new entry for parent of "rclone-v1.53.0-linux-amd64.zip", "/home/user/Downloads"
2020/09/11 10:54:34 DEBUG : Creating backend with remote "list:/backup/"
2020/09/11 10:54:34 DEBUG : fs cache: renaming cache item "list:/backup/" to be canonical "list:backup"
2020/09/11 10:54:34 DEBUG : rclone-v1.53.0-linux-amd64.zip: Need to transfer - File not found at Destination
2020/09/11 10:54:34 DEBUG : [backup]: new meta server: https://cld-extapi3.datacloudmail.ru/meta/
2020/09/11 10:54:34 DEBUG : [backup]: new upload shard: https://cld-upload10.cloud.mail.ru/upload-web/
[...]

Issue #4376, opened 2020-06-20
mailru: backend is failing TestCat and other Ranging integration tests


ncw:
It looks like there is an issue with the mailru [integration tests](https://pub.rclone.org/integration-tests/2020-06-20-050017/index.html)

![image](https://user-images.githubusercontent.com/536803/85201023-f7a7fc80-b2f3-11ea-87a3-f00a1f5b5991.png)

The tests do do with Range requests seem to be failing - well not all of them the ones in the backend tests are fine.

If I try this test locally it works

```
$ cd fs/operations
$ go test -v -remote TestMailru -verbose -test.run '^TestCat$'
=== RUN   TestCat
    TestCat: run.go:176: Remote "Local file system at /home/ncw/go/src/github.com/rclone/rclone/fs/operations/TestMailru/rclone-test-dozemeh0jadofex0veriwuc6", Local "Local file system at /tmp/rclone439982983", Modify Window "1ns"
--- PASS: TestCat (0.00s)
PASS
2020/06/20 12:48:28 DEBUG : Local file system at /home/ncw/go/src/github.com/rclone/rclone/fs/operations/TestMailru/rclone-test-dozemeh0jadofex0veriwuc6: Purge remote
ok  	github.com/rclone/rclone/fs/operations	0.023s
$ 
```

So I think it must be something to do with the lock/unlock server you can see [in the log](https://pub.rclone.org/integration-tests/2020-06-20-050017/mailru-fs.operations-TestMailru-5.txt)

```
"./operations.test -test.v -test.timeout 1h0m0s -remote TestMailru: -verbose -test.run '^TestCat$'" - Starting (try 5/5)
=== RUN   TestCat
2020/06/20 05:39:42 DEBUG : [rclone-test-hutijeg5kewezuv6daqakov5]: new meta server: https://cld-extapi2.datacloudmail.ru/meta/
2020/06/20 05:39:43 DEBUG : [rclone-test-hutijeg5kewezuv6daqakov5]: Switch file server: locks 1, url https://cloclo21.datacloudmail.ru/oauth-get/, expiry "2020-06-20T05:42:43.130577394Z"
2020/06/20 05:39:43 DEBUG : [rclone-test-hutijeg5kewezuv6daqakov5]: Unlock file server: locks 0, url https://cloclo21.datacloudmail.ru/oauth-get/
[...]

PR #4607, opened 2020-09-17
mailru: fix uploads and range requests

ivandeex:

Recently mail.ru changed their protocol. First, now clients should book upload server via a request to https://dispatcher.cloud.mail.ru/u instead of shard info dispatcher requests to meta server. Next, starting from June 2020 download servers ignore range requests from clients and return response 200 instead of 206. This PR adds logic to detect this behavior. Since upload functionality is critical, I suggest publishing this fix with the next point release.

This PR fixes issues https://github.com/rclone/rclone/issues/4584, https://github.com/rclone/rclone/issues/4376 and reverts commit https://github.com/rclone/rclone/commit/9d3d397f50fe94df0f723a571507e2ce72428ab6

#### Checklist


ivandeex:
Integration tests passed.

ivandeex:
@ncw
How can we make VFS layer prefer full object downloads vs ranges?

ncw:
Thanks for fixing this :-)

> Recently mail.ru changed their protocol. First, now clients should book upload server via a request to https://dispatcher.cloud.mail.ru/u instead of shard info dispatcher requests to meta server. Next, starting from June 2020 download servers ignore range requests from clients and return response 200 instead of 206. This PR adds logic to detect this behavior.

Hmm, this is a bit unfortunate. So I understand this to mean that no partial requests are possible at all?

If so then this is the first backend where this hasn't been possible :-(

> Since upload functionality is critical, I suggest publishing this fix with the next point release.

So put commit "mailru: fix uploads after recent changes on server ￼…" into the stable release - sounds OK to me.

> Integration tests passed.

:-)

I gave the code a once over and it looks fine to me.

> How can we make VFS layer prefer full object downloads vs ranges?

I think we'd need 
- a backend flag to announce that it can't do range requests
- we'd use this flag to then force --vfs-cache-mode full and download the entire objects...

I guess we can still download from the start and close the connection when done.

Note that range requests are used with multi stream downloads so these should be disabled too. Range requests are also used for resumes when downloads fail, for `rclone cat --head --tail` and probably lots of other places I've forgotten.

I see your workaround in the code which is OK provided you've got bandwidth to burn...

Is there no chance of getting some sort of partial download working?

ivandeex:
> Hmm, this is a bit unfortunate. So I understand this to mean that no partial requests are possible at all?
> If so then this is the first backend where this hasn't been possible :-(
> I see your workaround in the code which is OK provided you've got bandwidth to burn...
> Is there no chance of getting some sort of partial download working?

I tried several methods I could imagine. Sent "Content-Range", "Accept-Ranges" and various "Content-Type" with requests to no avail. I will come up with new ideas when there is one.

I also played with another mature OSS project related to MailRu.Cloud,
https://github.com/yar229/WebDavMailRuCloud (WDMRC) which semi-transparently converts between MRC and WebDAV.
I created a dockerfile with WDMRC https://github.com/ivandeex/ivadocker/tree/master/wdmrc and ran it as a daemon.
I used it as a black-box since I'm not a .NET guru, so I created a test WebDAV remote [wdmrc] and ran integration tests against it.
TestCat failed on range requests in the same way.
This means WDMRC also suffers from disappeared support of partial downloads and has no cure currently.

We could discuss this with https://github.com/yar229 and try to find a solution together.
Alas, c'est la vie, and I've got unlimited home plan to burn.

ivandeex:
> > Since upload functionality is critical, I suggest publishing this fix with the next point release.
> So put commit "mailru: fix uploads after recent changes on server ￼…" into the stable release - sounds OK to me.

Added to the 1.54 milestone. Having it in 1.53.2 would be fine too.

ivandeex:
> > How can we make VFS layer prefer full object downloads vs ranges?
> 
> I think we'd need
>     * a backend flag to announce that it can't do range requests
>     * we'd use this flag to then force --vfs-cache-mode full and download the entire objects...
> I guess we can still download from the start and close the connection when done.
> 
> Note that range requests are used with multi stream downloads so these should be disabled too. Range requests are also used for resumes when downloads fail, for `rclone cat --head --tail` and probably lots of other places I've forgotten.

I suggest to open another issue or PR to implement that backend flag (fs.Features flag? option?)

ivandeex:
Do you agree to merge this PR? Squash or not to squash?

If we don't find a proper solution, I will submit a change to the mail.ru documentation with --vfs-cache-xxx notes
and stating this backend is bandwidth hungry.

ncw:
> I tried several methods I could imagine. Sent "Content-Range", "Accept-Ranges" and various "Content-Type" with requests to no avail. I will come up with new ideas when there is one.

:-)

> I also played with another mature OSS project related to MailRu.Cloud,
[snip]
> TestCat failed on range requests in the same way.
> This means WDMRC also suffers from disappeared support of partial downloads and has no cure currently.

OK it was worth a try!
 
> We could discuss this with https://github.com/yar229 and try to find a solution together.

Is there an official client? The web client? That probably doesn't do range requests unless it lets you do stuff like play videos.

> > I think we'd need
> > * a backend flag to announce that it can't do range requests
> > * we'd use this flag to then force --vfs-cache-mode full and download the entire objects...
> > I guess we can still download from the start and close the connection when done.
> > Note that range requests are used with multi stream downloads so these should be disabled too. Range requests are also used for resumes when downloads fail, for `rclone cat --head --tail` and probably lots of other places I've forgotten.
> 
> I suggest to open another issue or PR to implement that backend flag (fs.Features flag? option?)

Yes what we need is a Feature flag

https://github.com/rclone/rclone/blob/f3cf6fcdd7a7cd4952ed1c2043b97106d4750a2f/fs/fs.go#L504-L507

These are quite easy to create

Do you fancy opening an issue, then we can make notes on things which need fixing?

@creativeprojects - I just remembered that encrypted seafile backends don't support seek too - is that right? We should probably set the flag there too.

> Do you agree to merge this PR? Squash or not to squash?

Yes - LGTM to merge. I think I prefer the separate commits.

Any commits you want in the stable release can you `git cherry-pick` them into the v1.53-stable branch. Or send me the commit numbers and I'll do it.

I'm beginning to think we need a better workflow for the stable branch but I haven't figured it out yet - any suggestions welcome :-)

> If we don't find a proper solution by 1.54, I will submit a change to the mail.ru documentation with --vfs-cache-xxx notes
> and stating this backend is bandwidth hungry.

OK seems reasonable.

ivandeex:
Unsquashed merge (as 3 commits) is "Not enabled for this repository"

creativeprojects:
>  I just remembered that encrypted seafile backends don't support seek too - is that right? We should probably set the flag there too.
> 

This is right, we could add it there as well 👍

ivandeex:
> Is there an official client? The web client? That probably doesn't do range requests unless it lets you do stuff like play videos.
There are both, and web UI can play clips AFAIK. That's a good catch, I'll give it a try.

Issue #4596, opened 2020-09-15
Improve logging with systemd integration


hekmon:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.53.1
- os/arch: linux/amd64
- go version: go1.15

#### What problem are you are trying to solve?
As rclone can already be aware of systemd presence thru the notify socket (used by the mount command with systemd type notify) it would be nice to enhance a bit the systemd integration:

- **Cleaner logs**
The logger flags are not adjusted when running under systemd: it overloads the logs like this:
  ```
  Sep 15 05:36:00 server rclone[2838611]: 2020/09/15 05:36:00 INFO  : some/file: vfs cache: removed cache file as file deleted
  ```
  Date and time are present 2 times: one from journald and one from the rclone logger. Because rclone is already able to detect if it is launched by systemd it would be nice to change the rclone logger default flags when systemd launch is detected to remove the rlcone log date and time in order to have cleaner logs.

- **Prettier (and better integrated) logs**
journald handles [priority levels](https://wiki.archlinux.org/index.php/Systemd/Journal#Priority_level) just fine and can adapt output with text formating and colors for each level to facilitate log parsing by humans. It also allows an operator to filter any log/priority level using `journalctl` even if they are fired by the software and saved by journald for example.
There is [several ways](http://0pointer.de/blog/projects/journal-submit.html) to do that but in that case the easiest way is to prefix each line with a balise like `<0>`. A moddest logger implementation using this feature can be found [here](https://github.com/hekmon/hllogger). 
I am aware rclone does have its own more sophisticated logger: this would mean adding an auto prefixing feature if systemd launch is detected.

ncw:
This seems like a good idea . I'm sure we could adjust rclone's logger very easily.

Do you think the notify signal is enough? Potentially you can run of the services other than mount eg rclone serve off systemd too.

Do you want to have a go at this?

hekmon:
You are right. While having the notify socket annouced thru env means there is 99,9% of chances we are launched by systemd, the feature itself should be activatable manually if needed for other (futures ?) cases.

So this would mean:

- a way to activate systemd support (a flag ?)
- automatically turning this feature on if the [notify socket is detected](https://github.com/iguanesolutions/go-systemd/blob/v4.0.1/notify.go#L12) ?

I can definitly take a swig at it even if I not really familliar with the rlcone codebase (I won't be able to give an ETA :D)

ncw:
> So this would mean:
> 
[...]

PR #4609, opened 2020-09-18
Improve systemd integration (logging & status)

hekmon:
Following #4596

I hope the changes are in the right spots, don't hesitate to request additional changes even just to better match the rclone architecture. I have separated each step in a different commit for easier reviewing but in the end I think this PR should have its commits squashed.

preview:
```
✔ 18:16 ~/Golang/Workspaces/perso/src/github.com/rclone/rclone [ sysdjournald L | ✚ 1 ] $ ./rclone --log-level DEBUG version
2020/09/18 18:16:51 DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone" "--log-level" "DEBUG" "version"]
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.14.6
2020/09/18 18:16:51 DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone" "--log-level" "DEBUG" "version"]
✔ 18:16 ~/Golang/Workspaces/perso/src/github.com/rclone/rclone [ sysdjournald L | ✚ 1 ] $ ./rclone --log-level DEBUG --log-systemd version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone" "--log-level" "DEBUG" "--log-systemd" "version"]
<7>DEBUG : rclone: systemd logger support activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.14.6
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone" "--log-level" "DEBUG" "--log-systemd" "version"]
✔ 18:16 ~/Golang/Workspaces/perso/src/github.com/rclone/rclone [ sysdjournald L | ✚ 1 ] $ NOTIFY_SOCKET="yay" ./rclone --log-level DEBUG version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone" "--log-level" "DEBUG" "version"]
<7>DEBUG : rclone: 'NOTIFY_SOCKET' env var is set: systemd logger support automatically activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.14.6
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone" "--log-level" "DEBUG" "version"]
✔ 18:16 ~/Golang/Workspaces/perso/src/github.com/rclone/rclone [ sysdjournald L | ✚ 1 ] $
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	systemd "github.com/iguanesolutions/go-systemd/v5"
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
@@ -364,6 +365,12 @@ func StartStats() func() {
 
 // initConfig is run by cobra after initialising the flags
 func initConfig() {
+	// Activate logger systemd support if systemd invocation ID is detected
+	_, sysdLaunch := systemd.GetInvocationID()
+	if sysdLaunch {
+		fs.Config.LogSystemdSupport = true // used during fslog.InitLogging()
+	}
+
 	// Start the logger
 	fslog.InitLogging()
 
@@ -378,6 +385,13 @@ func initConfig() {
 
 	// Write the args for debug purposes
 	fs.Debugf("rclone", "Version %q starting with parameters %q", fs.Version, os.Args)
+
+	// Inform user about systemd log support now that we have a logger
+	if sysdLaunch {
+		fs.Debugf("rclone", "systemd logging support automatically activated")
+	} else if fs.Config.LogSystemdSupport {
+		fs.Debugf("rclone", "systemd logging support manually activated")
+	}
 
 	// Start the remote control server if configured
 	_, err = rcserver.Start(&rcflags.Opt)
--- a/fs/log.go
+++ b/fs/log.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	sysdjournald "github.com/iguanesolutions/go-systemd/v5/journald"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -70,7 +71,28 @@ func (l *LogLevel) Type() string {
 
 // LogPrint sends the text to the logger of level
 var LogPrint = func(level LogLevel, text string) {
-	text = fmt.Sprintf("%-6s: %s", level, text)
+	var prefix string
+	if Config.LogSystemdSupport {
+		switch level {
+		case LogLevelDebug:
+			prefix = sysdjournald.DebugPrefix
+		case LogLevelInfo:
+			prefix = sysdjournald.InfoPrefix
+		case LogLevelNotice:
+			prefix = sysdjournald.NoticePrefix
+		case LogLevelWarning:
+			prefix = sysdjournald.WarningPrefix
+		case LogLevelError:
+			prefix = sysdjournald.ErrPrefix
+		case LogLevelCritical:
+			prefix = sysdjournald.CritPrefix
+		case LogLevelAlert:
+			prefix = sysdjournald.AlertPrefix
+		case LogLevelEmergency:
+			prefix = sysdjournald.EmergPrefix
+		}
+	}
+	text = fmt.Sprintf("%s%-6s: %s", prefix, level, text)
 	_ = log.Output(4, text)
 }
 
--- a/fs/log/log.go
+++ b/fs/log/log.go
@@ -90,23 +90,25 @@ func Stack(o interface{}, info string) {
 func InitLogging() {
 	flagsStr := "," + Opt.Format + ","
 	var flags int
-	if strings.Contains(flagsStr, ",date,") {
-		flags |= log.Ldate
-	}
-	if strings.Contains(flagsStr, ",time,") {
-		flags |= log.Ltime
-	}
-	if strings.Contains(flagsStr, ",microseconds,") {
-		flags |= log.Lmicroseconds
+	if !fs.Config.LogSystemdSupport {
+		if strings.Contains(flagsStr, ",date,") {
+			flags |= log.Ldate
+		}
+		if strings.Contains(flagsStr, ",time,") {
+			flags |= log.Ltime
+		}
+		if strings.Contains(flagsStr, ",microseconds,") {
+			flags |= log.Lmicroseconds
+		}
+		if strings.Contains(flagsStr, ",UTC,") {
+			flags |= log.LUTC
+		}
 	}
 	if strings.Contains(flagsStr, ",longfile,") {
 		flags |= log.Llongfile
 	}
 	if strings.Contains(flagsStr, ",shortfile,") {
 		flags |= log.Lshortfile
-	}
-	if strings.Contains(flagsStr, ",UTC,") {
-		flags |= log.LUTC
 	}
 	log.SetFlags(flags)
 
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -11,7 +11,7 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/okzk/sdnotify"
+	sysdnotify "github.com/iguanesolutions/go-systemd/v5/notify"
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs"
@@ -448,13 +448,13 @@ func Mount(VFS *vfs.VFS, mountpoint string, mount MountFn, opt *Options) error {
 
 	// Unmount on exit
 	fnHandle := atexit.Register(func() {
+		_ = sysdnotify.Stopping()
 		_ = unmount()
-		_ = sdnotify.Stopping()
 	})
 	defer atexit.Unregister(fnHandle)
 
 	// Notify systemd
-	if err := sdnotify.Ready(); err != nil && err != sdnotify.ErrSdNotifyNoSocket {
+	if err := sysdnotify.Ready(); err != nil {
 		return errors.Wrap(err, "failed to notify systemd")
 	}
 
@@ -479,8 +479,8 @@ waitloop:
 		}
 	}
 
+	_ = sysdnotify.Stopping()
 	_ = unmount()
-	_ = sdnotify.Stopping()
 
 	if err != nil {
 		return errors.Wrap(err, "failed to umount FUSE fs")
--- a/fs/config.go
+++ b/fs/config.go
@@ -42,6 +42,7 @@ var (
 type ConfigInfo struct {
 	LogLevel               LogLevel
 	StatsLogLevel          LogLevel
+	LogSystemdSupport      bool
 	UseJSONLog             bool
 	DryRun                 bool
 	Interactive            bool
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -123,6 +123,7 @@ func AddFlags(flagSet *pflag.FlagSet) {
 	flags.StringArrayVarP(flagSet, &downloadHeaders, "header-download", "", nil, "Set HTTP header for download transactions")
 	flags.StringArrayVarP(flagSet, &headers, "header", "", nil, "Set HTTP header for all transactions")
 	flags.BoolVarP(flagSet, &fs.Config.RefreshTimes, "refresh-times", "", fs.Config.RefreshTimes, "Refresh the modtime of remote files.")
+	flags.BoolVarP(flagSet, &fs.Config.LogSystemdSupport, "log-systemd", "", fs.Config.LogSystemdSupport, "Activate systemd integration for the logger.")
 }
 
 // ParseHeaders converts the strings passed in via the header flags into HTTPOptions
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/google/go-querystring v1.0.0 // indirect
 	github.com/hanwen/go-fuse/v2 v2.0.3
+	github.com/iguanesolutions/go-systemd/v5 v5.0.0
 	github.com/jlaffaye/ftp v0.0.0-20200720194710-13949d38913e
 	github.com/jzelinskie/whirlpool v0.0.0-20170603002051-c19460b8caa6
 	github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 // indirect
@@ -35,7 +36,6 @@ require (
 	github.com/ncw/go-acd v0.0.0-20171120105400-887eb06ab6a2
 	github.com/ncw/swift v1.0.52
 	github.com/nsf/termbox-go v0.0.0-20200418040025-38ba6e5628f1
-	github.com/okzk/sdnotify v0.0.0-20180710141335-d9becc38acbd
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pkg/errors v0.9.1
 	github.com/pkg/sftp v1.11.0
--- a/go.sum
+++ b/go.sum
@@ -232,6 +232,8 @@ github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
 github.com/huandu/xstrings v1.0.0/go.mod h1:4qWG/gcEcfX4z/mBDHJ++3ReCw9ibxbsNJbcucJdbSo=
 github.com/ianlancetaylor/demangle v0.0.0-20181102032728-5e5cf60278f6/go.mod h1:aSSvb/t6k1mPoxDqO4vJh6VOCGPwU4O0C2/Eqndh1Sc=
+github.com/iguanesolutions/go-systemd/v5 v5.0.0 h1:E4OUiBdmlD1IsClS6cmRIdzWBW8T8UBitCqYem7A1KY=
+github.com/iguanesolutions/go-systemd/v5 v5.0.0/go.mod h1:VPlzL6z0rXd3HU7oLkMoEqTWBhHClInYX9rP2U/+giI=
 github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NHg9XEKhtSvM=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/jessevdk/go-flags v0.0.0-20141203071132-1679536dcc89/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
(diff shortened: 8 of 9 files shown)


hekmon:
I just added an extra feature related to systemd and the notify socket: status.

It allows to easily show a custom line when showing the status of a systemd unit. It was a perfect match for the the vfs cache stats. Following screenshot will show the custom status (5th line) and the different formating for each log level (actually 2 on this example).

I hope the result is to your liking :)

rclone_sysd_redacted

hekmon:
I have pushed the work a bit further in order to have a cleaner code:

- changed the lib used for systemd notify
- allows to skip saving the socket to conf
- cleaner code in rclone (testing is systemd notify socket is present simplified, auto noop notify calls when its not, etc...)

I did not integrated theses changes in this PR as it is not the initial scope but you can check the modifications it would bring [here](https://github.com/hekmon/rclone/compare/sysdjournald...hekmon:sysdig1lib). If it is ok for you, I will fast forward the PR branch to include theses changes.

Full disclosure: this is a systemd library we are using & maintaining at my company (it is MIT licensed).

fs/log.go
```diff
@@ -44,6 +44,25 @@ var logLevelToString = []string{
 	LogLevelDebug:     "DEBUG",
 }
 
+const (
+	// SystemdEmergPrefix is the string to prefix in Emergency for systemd-journald
```
ncw, line 48: Does your systemd module have constants for these BTW?

fs/log/log.go
```diff
@@ -90,10 +90,10 @@ func Stack(o interface{}, info string) {
 func InitLogging() {
 	flagsStr := "," + Opt.Format + ","
 	var flags int
-	if strings.Contains(flagsStr, ",date,") {
+	if strings.Contains(flagsStr, ",date,") && !fs.Config.LogSystemdSupport {
```
ncw, line 93: You've only address the default `--log-format` here.

I presume we don't want any time ever if `fs.Config.LogSystemdSupport` is set do we? In which case maybe `if` out all the time setting?

cmd/cmd.go
```diff
@@ -364,6 +364,11 @@ func StartStats() func() {
 
 // initConfig is run by cobra after initialising the flags
 func initConfig() {
+	// Activate logger systemd support if notify socket is detected
+	if socket := os.Getenv("NOTIFY_SOCKET"); socket != "" {
```
ncw: Are there any other ways of detecting if we are running under systemd?

For example if you were running `rclone serve http` under systemd?

Or maybe those should be supporting the notify socket too?

Potential commands you might run under systemd might be

- serve (http/ftp/sftp/webdav/dlna)
- rcd

ncw (commented):
This is very nice :-)

I've put a few comments inline.

I think what you've done here should probably be two commits - one for the systemd log support and another for the systemd status.

I'll reply to your other comments in just a sec!

ncw:
> I hope the changes are in the right spots, don't hesitate to request additional changes even just to better match the rclone architecture. I have separated each step in a different commit for easier reviewing but in the end I think this PR should have its commits squashed.

I think you've done a great job :-)

> I just added an extra feature related to systemd and the notify socket: status.
> 
> It allows to easily show a custom line when showing the status of a systemd unit. It was a perfect match for the the vfs cache stats. Following screenshot will show the custom status (5th line) and the different formating for each log level (actually 2 on this example).
> 
> I hope the result is to your liking :)

Nice! I didn't know you could do that with systemd.

I think the log you've chosen is pretty good! Note that you only get that log with `--vfs-cache-mode` > `off` but maybe that is OK.

> I have pushed the work a bit further in order to have a cleaner code:
> 
> * changed the lib used for systemd notify
> * allows to skip saving the socket to conf
> * cleaner code in rclone (testing if systemd notify socket is present simplified, auto noop notify calls when its not, etc...)
> 
> I did not integrated theses changes in this PR as it is not the initial scope but you can check the modifications it would bring [here](https://github.com/hekmon/rclone/compare/sysdjournald...hekmon:sysdig1lib). If it is ok for you, I will fast forward the PR branch to include theses changes.

I had a look and I think they look good - can you integrate them in this PR?

> Full disclosure: this is a systemd library we are using & maintaining at my company (it is MIT licensed).

It looks nice :-)

hekmon:
> I think what you've done here should probably be two commits - one for the systemd log support and another for the systemd status.

Good for me. I think I will remove the status support for now and once we are done reviewing the different log commits, I will squash them upstream then re-add the status on top of it.

> Nice! I didn't know you could do that with systemd.
> I think the log you've chosen is pretty good! Note that you only get that log with --vfs-cache-mode > off but maybe that is OK.

If you have an idea of what status could be on other cases, feel free to tell me :) But I have found myself scrolling the logs a lot to get the last vfs cache poll status (in use/uploading/to upload and size) and I think these information are really good candidates to show to the user when he requests a status update from systemd (but as the poll duration can be configured, we must add the time to the status line in order to give some context to these data).

> I had a look and I think they look good - can you integrate them in this PR?
Sure, thx !

Answering inline comments now ;)

fs/log.go
hekmon: It does not. But it should actually, thanks for the idea ! I will update the lib and bump the version asap.

fs/log/log.go
hekmon: yes no need to test it multiples times, especially after a strings contains as it is a bool (should be first test). I will fix that.

cmd/cmd.go
hekmon: Yes the trick here is not that we detect that we are running under systemd but that we are running by systemd in notify mode. Notify mode will allows startup and stop signals, status update etc... But this is independent from the logger configuration:
* if we detect the notify socket it means systemd is our parent so we automatically activate the logger systemd support, but:
* we could be launched by systemd in "simple" mode: no notify support, but user could still activate the logger to be compliant with systemd thru the flag. I don't know another way than the notify socket to detect a systemd launch unfortunatly, I will look it up.

So to answer your question (for now): if others commands want to be compatible the notify launch (which is different from just logger support) they will need, like mount, to send the `NotifyReady` signal otherwise systemd will think the start has timed out and kill the process after `ExecStartTimeout`.

ncw: OK. So the other server commands should probably support the NotifyReady signal...

That sounds like a job for a separate PR!

hekmon:
So here is the actual situation:

- [new lib version](https://github.com/iguanesolutions/go-systemd/tree/v5.0.0) to handle everything related to sysd containing your suggestions
- logger prefixes now in the lib
- "if out" time related flags when sysd log support is on
- removed status feature for now (will re-add it when logger reviewing is complete and commits squashed in order to have 2 clean commits in this PR)
- systemd startup detection improved: not relying on notify socket anymore but on the invocation ID set by sysd. This feature appeared in [v232](https://github.com/iguanesolutions/go-systemd/blob/v5.0.0/sysd.go#L7) so this means, for **logger support** :
  - Ubuntu [bionic](https://packages.ubuntu.com/fr/bionic/systemd) (18.04) and [focal](https://packages.ubuntu.com/fr/focal/systemd) (20.04) supported but not [xenial](https://packages.ubuntu.com/fr/xenial/systemd) (16.04)
  - Debian [stretch](https://packages.debian.org/stretch/systemd) and [buster](https://packages.debian.org/buster/systemd) supporter but not [jessie](https://packages.debian.org/jessie/systemd) (which is EOL anyway)
  - We could still test notify socket to enhance the detection of old versions of systemd if you like (if invocation id is present OR notify socket is present).
- notify actions are always tried without detecting if the socket is here because the lib handles it transparently (noop if no socket notify)

I will let you resolv previous inline comments if ok :)

cmd/cmd.go
hekmon: new commits allows to detect sysd launch even without notify support !
see [last comment](https://github.com/rclone/rclone/pull/4609#issuecomment-696155435)

hekmon:
@ncw still here ? :)

hekmon:
I have moved forward a bit more to prepare your review:
* rebased on current master
* squashed all the log related commit as we discussed
* create a separate commit for the status update on mount with cache

Current version give the following:
rclonedev

```
~$ ./rclone_dev --log-level DEBUG version
2020/10/05 12:35:35 DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
2020/10/05 12:35:35 DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]

~$ ./rclone_dev --log-level DEBUG --log-systemd version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "--log-systemd" "version"]
<7>DEBUG : rclone: systemd logging support manually activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "--log-systemd" "version"]

~$ INVOCATION_ID="fakeID" ./rclone_dev --log-level DEBUG version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
<7>DEBUG : rclone: systemd logging support automatically activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
hekmon@stationem:~$
```

As far as I am concerned, I think we are good to go ! Ready when you are :)

hekmon:
I have moved forward a bit more to prepare your review:
* rebased on current master
* squashed all the log related commit as we discussed
* create a separate commit for the status update on mount with cache

Current version give the following:
rclonedev

```
~$ ./rclone_dev --log-level DEBUG version
2020/10/05 12:35:35 DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
2020/10/05 12:35:35 DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]

~$ ./rclone_dev --log-level DEBUG --log-systemd version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "--log-systemd" "version"]
<7>DEBUG : rclone: systemd logging support manually activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "--log-systemd" "version"]

~$ INVOCATION_ID="fakeID" ./rclone_dev --log-level DEBUG version
<7>DEBUG : rclone: Version "v1.54.0-DEV" starting with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
<7>DEBUG : rclone: systemd logging support automatically activated
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2
<7>DEBUG : rclone: Version "v1.54.0-DEV" finishing with parameters ["./rclone_dev" "--log-level" "DEBUG" "version"]
```

As far as I am concerned, I think we are good to go ! Ready when you are :)

fs/log/log.go
```diff
@@ -90,24 +90,26 @@ func Stack(o interface{}, info string) {
...
-	if strings.Contains(flagsStr, ",date,") {
-		flags |= log.Ldate
-	}
-	if strings.Contains(flagsStr, ",time,") {
-		flags |= log.Ltime
-	}
-	if strings.Contains(flagsStr, ",microseconds,") {
-		flags |= log.Lmicroseconds
+	if !fs.Config.LogSystemdSupport {
+		if strings.Contains(flagsStr, ",date,") && !fs.Config.LogSystemdSupport {
```
ncw, line 94: I think you've got a spurious `&& !fs.Config.LogSystemdSupport` here?

fs/log/log.go
```diff
@@ -90,24 +90,26 @@ func Stack(o interface{}, info string) {
...
+			flags |= log.Ldate
+		}
+		if strings.Contains(flagsStr, ",time,") && !fs.Config.LogSystemdSupport {
```
ncw, line 97: ...and here

ncw (commented):
Apologies for the delay in reviewing - things very hectic here at the moment!

I think this looks almost ready to go.

I think activating the logging system with `systemd.GetInvocationID` looks fine - there is a workaround for old OSes (use the flag).

If you can fix up the very minor points inline then I'll merge :-)

Thank you

fs/log/log.go
hekmon: damn...

hekmon:
No matter how many times I have reread the changes, these slipped thru ! They are now fixed (rebase/edit/amend/force push to keep the 2 separate feature commits you requested)

I guess we are good to go :)

ncw:
> No matter how many times I have reread the changes, these slipped thru ! They are now fixed (rebase/edit/amend/force push to keep the 2 separate feature commits you requested)

Oh git! We love you really ;-)

I use `git commit --fixup ; git rebase -i` for this workflow.
 
> I guess we are good to go :)

It looks great now - thank you very much! I'll merge now.

This is a nice feature.

hekmon:
:tada: 

I will lookup --fixup ! I used rebase interactive with edit on the logger commit line then modified the file, commit amend and rebase continue :p

Thanks for accepting this PR ! (my eyes can't wait the next release ;)

darthShadow:
@hekmon Sorry for using a closed PR, but are there any flags available to add the timestamp in the systemd service if it's being logged to a file?

Currently, these are the logs printed from the systemd service to the file:
```
<5>NOTICE:
Elapsed time:       5m0.4s

<6>INFO  : vfs cache: cleaned: objects 4891 (was 4891) in use 0, to upload 0, uploading 0, total size 3.990T (was 3.990T)
<5>NOTICE:
Elapsed time:      5m30.4s

<5>NOTICE:
Elapsed time:       6m0.4s
```

And these were the earlier logs (before the PR):
```
2020/10/23 12:01:10 INFO  : vfs cache: cleaned: objects 4891 (was 4891) in use 31, to upload 0, uploading 0, total size 3.998T (was 3.998T)
2020/10/23 12:01:10 INFO  : Google drive root 'data': Retrieved Activities : 0
2020/10/23 12:01:40 NOTICE:
Transferred:      105.648T / 105.648 TBytes, 100%, 22.983 MBytes/s, ETA 0s
Transferred:      1908589 / 1908589, 100%
Elapsed time:  1535h0m30.4s
```

hekmon:
Hello @darthShadow ,

I am not sure to quite get your use case: your are using rclone as a systemd service but logging to a file ? (and if it is the case, why logging in 2 places at the same time and not use `journalctl` to access your logs ?)

darthShadow:
I know this may be a pretty weird case, but it's what I have been using for a few years now. There are reasons for going with this combination:
* Allow for auto process restarts with systemd
* Keep logging to file so I can easily use other unix tools like grep etc. and also allow managing the rotation of the files myself without needing to modify the global journal config.

The 2nd point is probably possible but I have not looked into it in detail.

I just wanted to know whether this is expected and if there is a way to make it work with my existing setup.

@ncw Do you think this needs a note in the docs somewhere since the observed output is changing slightly?

hekmon:
@darthShadow currently there is a flag to force the activation but not to force the deactivation, sorry :(

For your first point I am not sure to see why this would need a file based log.

About the second you definitly can use any unix tools you want as long as you use journalctl as a pipe source (ex: `journalctl -u yourunit.service | grep ...`). Journald can be configured to keep its logs persistent (`Storage=`) with advanced file rotation (too many options to list them here).

I strongly recommend you to take a look at [journald.conf](https://www.freedesktop.org/software/systemd/man/journald.conf.html) as it is way more powerfull than plain old flat text files but can still be used with any unix workflows thanks to stdout and piping. It is the de facto standard to handle logs when using systemd and its unit files.

hekmon:
@ncw could we consider printing a warning if both systemd output and a log file are active in the same time ? I think it is an edgy use case but systemd output is definitly meant for systemd... not a file :/

darthShadow:
The problem with using journal is that there is no service-specific config as an option, only global configs. This is a problem for me when I have different services, where for some I need to keep logs indefinitely and some need to be rotated automatically depending on how critical the service is. This is an open issue for systemd: https://github.com/systemd/systemd/issues/9519

darthShadow:
I would prefer to have the option to disable it too for the few edge scenarios like mine.

hekmon:
> The problem with using journal is that there is no service-specific config as an option, only global configs. This is a problem for me when I have different services, where for some I need to keep logs indefinitely and some need to be rotated automatically depending on how critical the service is. This is an open issue for systemd: [systemd/systemd#9519](https://github.com/systemd/systemd/issues/9519)

I see... Personnaly, if I would need to keep logs for a very long period of time I would export them elsewhere/to a central location. For example, with [Loki](https://grafana.com/oss/loki/), leaving the local system handling as much as logs as it can for a homogenous period of time. May be not a proper solution from your pov but I hope it helps :(

hekmon:
> I would prefer to have the option to disable it too for the few edge scenarios like mine.

This would means a 2nd flag to block systemd detection. Let's wait for @ncw input :)

ncw:
> could we consider printing a warning if both systemd output and a log file are active in the same time ? I think it is an edgy use case but systemd output is definitly meant for systemd... not a file :/

Perhaps we should disable the feature if `--log-file` is in use. Would that satisfy your requirements @darthShadow ?

We could also (with a bit of trickery) make `--log-systemd=false` disable the systemd logs if they have been activated.

darthShadow:
The first option would work for me and also be preferable since it maintains backwards compatibility.

hekmon:
using `log-file` auto detection/deactivation would mean adding an automated system over an other automated system: not the KISSiest way to go but it does enable backward compat :)

I personally prefer the `--log-systemd=false` leaving only one automated system which can be forced on (`--log-systemd`) or forced off (`--log-systemd=false`). Much cleaner !

ncw:
@hekmon wrote

> using `log-file` auto detection/deactivation would mean adding an automated system over an other automated system: not the KISSiest way to go but it does enable backward compat :)

:-)  I think that is my preferred option
 
> I personally prefer the `--log-systemd=false` leaving only one automated system which can be forced on (`--log-systemd`) or forced off (`--log-systemd=false`). Much cleaner !

Maybe we should do that too.

@hekmon - do you fancy doing a PR for disabling systemd integration if --log-file is set? Maybe do the --log-systemd=false fix too? (you'll need a bit of pflag trickery to see if the flag was actually set, eg

https://github.com/rclone/rclone/blob/e00bf3d723c636633474629968046342e599e15e/cmd/cmd.go#L229-L232

)

hekmon:
Sure thing. As always I can't promise any ETA but I will take a look asap :)

If I understand the trick correctly, using `--log-systemd=false` will set the flag at false but set the changed [property](https://godoc.org/github.com/spf13/pflag#Flag) to true (even if the default is false) ?

ncw:
@hekmon wrote
> Sure thing. As always I can't promise any ETA but I will take a look asap :)

Thank you :-)

> If I understand the trick correctly, using `--log-systemd=false` will set the flag at false but set the changed [property](https://godoc.org/github.com/spf13/pflag#Flag) to true (even if the default is false) ?

That is correct - the default is `false` but the `Changed` flag shows whether it was set on the command line or not. We use that trick in a few other places.

hekmon:
@ncw I had a very busy october, with very little free time... A lot going on with my personnal life and december will be even worse.

I don't think I will be able to take a swig at it before january, sorry :( If you want to do it yourself or ask someone else if this can not wait, please do.

aloababa:
I'm working on a PR for that i will make you a proposal in the next days.

Issue #3972, opened 2020-02-16
Chunker doesn't list all entries in directory


jfabaf:
#### What is the problem you are having with rclone?

I have this configuration:

Google Drive <- Crypt <- Chunker

When I list files of chunker configuration I get a lot of errors like this:

skip chunk "file1.ext.rcc001" without meta object

My chunker configuration is:

[gdrive3chunk]
type = chunker
remote = gdrive3crypt:/
chunk_size = 1G
hash_type = md5all
name_format = *.rcc###

#### What is your rclone version (output from `rclone version`)

rclone v1.51.0-DEV

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Ubuntu 18.04.4 LTS 

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive <- Crypter <- Chunker

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone ls -vv gdrive3chunk:/

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

rclone ls -vv gdrive3chunk:/
2020/02/16 19:58:21 DEBUG : rclone: Version "v1.51.0-DEV" starting with parameters ["./rclone" "ls" "-vv" "gdrive3chunk:/"]
2020/02/16 19:58:21 DEBUG : Using config file from "/home/jorge/.config/rclone/rclone.conf"
2020/02/16 19:58:21 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc001" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc002" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc011" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc016" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc019" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc026" without meta object
2020/02/16 19:58:26 DEBUG : Chunked 'gdrive3chunk:/': skip chunk "file1.ext.rcc030" without meta object
[...]

PR #4611, opened 2020-09-18
chunker: disable ListR to fix missing files on Drive

ivandeex:

Chunker needs to detect groups of files that form a compund object in the underlying directory.
Currently it uses a dead simple approach: it sorts directory listing by name and finds consecutive groups of file names with a common prefix that satisfies mask.
This approach requires that variable name part must be on the end of file name, for sorting to work.
This pre-condition is backed by the strict check at https://github.com/rclone/rclone/blob/v1.52-stable/backend/chunker/chunker.go#L413.

However, the other requirement has no check: the directory listing must be fed in a single piece.
But, if underlying remote (eg. `drive`) supports the `ListR` interface, the listing for a single directory could come in pieces
sorted one-by-one (see https://github.com/rclone/rclone/blob/v1.52-stable/backend/chunker/chunker.go#L660)
resulting in object parts randomly missing their compound group.

A complete solution will have `processEntries` routine keep directory scan state instead of sorting and let ListR feed partial listings in random order.
This PR is a workaround that simply marks chunker as incompatible with ListR.

Update: this issue was reported in https://github.com/rclone/rclone/issues/3972

The reporter confirmed that this workaround dismisses their issue (see https://github.com/rclone/rclone/issues/3972#issuecomment-632938201).
However, the ticket will stay open until finally solved by a later PR.

#### Checklist

- [ ] I have added documentation for the changes if appropriate (**no, it's a bugfix**).

### DIFF
--- a/backend/chunker/chunker.go
+++ b/backend/chunker/chunker.go
@@ -296,6 +296,8 @@ func NewFs(name, rpath string, m configmap.Mapper) (fs.Fs, error) {
 		ServerSideAcrossConfigs: true,
 	}).Fill(f).Mask(baseFs).WrapsFs(f, baseFs)
 
+	f.features.Disable("ListR") // Recursive listing may cause chunker skip files
+
 	return f, err
 }
 


ivandeex:
Mini-integration suite passed: https://github.com/ivandeex/rclone/runs/1138195210?check_suite_focus=true

ncw (approved):
Was there an issue? Or maybe you could put a link to the forum if it came up there in the commit message?

Code looks great.

One for stable branch?

ivandeex:
> Was there an issue?

This issue was reported in https://github.com/rclone/rclone/issues/3972

> Code looks great.

:) :) :) :)))))

> One for stable branch?

ok

PR #4625, opened 2020-09-24
sftp: allow user to optionally specify pubkey file to enable cert-based auth

sweharris:

Enable certificate based authentication, by allowing the user to optionally specify the public key to use.

We take the normal "signer" that was used for key auth, and if the user has specified the pubkey then we wrap that with a new signer that also holds the public key.

https://forum.rclone.org/t/ssh-certificate-based-authentication-does-not-work/

#### Checklist

- [*] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [*] I have added documentation for the changes if appropriate.
- [*] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -83,6 +83,11 @@ Only PEM encrypted key files (old OpenSSH format) are supported. Encrypted keys
 in the new OpenSSH format can't be used.`,
 			IsPassword: true,
 		}, {
+			Name: "pubkey_file",
+			Help: `Optional path to public key file.
+
+Set this if you have a signed certificate you want to use for authentication.` + env.ShellExpandHelp,
+		}, {
 			Name: "key_use_agent",
 			Help: `When set forces the usage of the ssh-agent.
 
@@ -190,6 +195,7 @@ type Options struct {
 	KeyPem            string `config:"key_pem"`
 	KeyFile           string `config:"key_file"`
 	KeyFilePass       string `config:"key_file_pass"`
+	PubKeyFile        string `config:"pubkey_file"`
 	KeyUseAgent       bool   `config:"key_use_agent"`
 	UseInsecureCipher bool   `config:"use_insecure_cipher"`
 	DisableHashCheck  bool   `config:"disable_hashcheck"`
@@ -438,6 +444,7 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 	}
 
 	keyFile := env.ShellExpand(opt.KeyFile)
+	pubkeyFile := env.ShellExpand(opt.PubKeyFile)
 	//keyPem := env.ShellExpand(opt.KeyPem)
 	// Add ssh agent-auth if no password or file or key PEM specified
 	if (opt.Pass == "" && keyFile == "" && !opt.AskPassword && opt.KeyPem == "") || opt.KeyUseAgent {
@@ -507,7 +514,38 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to parse private key file")
 		}
-		sshConfig.Auth = append(sshConfig.Auth, ssh.PublicKeys(signer))
+
+		// If a public key has been specified then use that
+		if pubkeyFile != "" {
+			certfile, err := ioutil.ReadFile(pubkeyFile)
+			if err != nil {
+				return nil, errors.Wrap(err, "unable to read cert file")
+			}
+
+			pk, _, _, _, err := ssh.ParseAuthorizedKey(certfile)
+			if err != nil {
+				return nil, errors.Wrap(err, "unable to parse cert file")
+			}
+
+			// And the signer for this, which includes the private key signer
+			// This is what we'll pass to the ssh client.
+			// Normally the ssh client will use the public key built
+			// into the private key, but we need to tell it to use the user
+			// specified public key cert.  This signer is specific to the
+			// cert and will include the private key signer.  Now ssh
+			// knows everything it needs.
+			cert, ok := pk.(*ssh.Certificate)
+			if !ok {
+				return nil, errors.New("public key file is not a certificate file: " + pubkeyFile)
+			}
+			pubsigner, err := ssh.NewCertSigner(cert, signer)
+			if err != nil {
+				return nil, errors.Wrap(err, "error generating cert signer")
+			}
+			sshConfig.Auth = append(sshConfig.Auth, ssh.PublicKeys(pubsigner))
+		} else {
+			sshConfig.Auth = append(sshConfig.Auth, ssh.PublicKeys(signer))
+		}
 	}
 
 	// Auth from password if specified
--- a/docs/content/sftp.md
+++ b/docs/content/sftp.md
@@ -102,7 +102,7 @@ excess files in the directory.
 The SFTP remote supports three authentication methods:
 
   * Password
-  * Key file
+  * Key file, including certificate signed keys
   * ssh-agent
 
 Key files should be PEM-encoded private key files. For instance `/home/$USER/.ssh/id_rsa`.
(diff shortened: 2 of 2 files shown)


sweharris:
Oh, I see some of the doc updates I made weren't necessary 'cos `make backenddocs` adds that bit.  Hopefully that won't be a blocker to this PR, though.  (I note that the `sftp.md` gets the current username added to the `--sftp-user` section, which probably isn't right!)

backend/sftp/sftp.go
```diff
@@ -507,7 +512,29 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
...
+			}
+
+			pk, _, _, _, err := ssh.ParseAuthorizedKey(certfile)
+			if err != nil {
+				return nil, errors.Wrap(err, "unable to parse cert file")
+			}
+
+			// And the signer for this, which includes the private key signer
+			// This is what we'll pass to the ssh client.
+			pubsigner, err := ssh.NewCertSigner(pk.(*ssh.Certificate), signer)
```
ncw, line 530: Can this type assertion `pk.(*ssh.Certificate)`  ever fail?

ncw (commented):
I think that looks really good - thank you :-)

See inline for a very minor query!

Can you stick a reference to the forum thread in the commit message

    See: https://forum.rclone.org/t/ssh-certificate-based-authentication-does-not-work/19222

Also I think the using the concatenated key/cert needs documenting.

ncw:
> Oh, I see some of the doc updates I made weren't necessary 'cos `make backenddocs` adds that bit. Hopefully that won't be a blocker to this PR, though. 

That is not a prob!

> (I note that the `sftp.md` gets the current username added to the `--sftp-user` section, which probably isn't right!)

Hmm!

The environment variables set here are supposed to fix that but I guess they aren't working.

```Makefile
backenddocs: rclone bin/make_backend_docs.py
        XDG_CACHE_HOME="" XDG_CONFIG_HOME="" HOME="\$$HOME" USER="\$$USER" ./bin/make_backend_docs.py
```

This isn't the only place either

```
github.com/rclone/rclone/docs/content$ git grep ncw
ftp.md:FTP username, leave blank for current username, ncw
sftp.md:SSH username, leave blank for current username, ncw
sftp.md:SSH username, leave blank for current username, ncw
```

Ideas?

backend/sftp/sftp.go
sweharris: I've been playing with this and I think I've managed to get it to fail; if a _standard_ public key is passed instead of a certificate then we get a `panic: interface conversion: ssh.PublicKey is *ssh.rsaPublicKey, not *ssh.Certificate`.  I'll wrap it in a `case` test so now it reads: `2020/09/25 07:57:42 Failed to create file system for "sftpcert:/tmp/XX": Public key file is not a certificate file: /tmp/TSTKEYS/id_rsa.pub`

sweharris:
Looking at `sftp.go`, you're calling `user.Current()` to get the user name.  If that fails you drop back to `$USER`.

Following the source chain, `user.Current()` in `os/user/lookup.go` calls `current()` to populate the cache.  On Unix this is defined in `os/user/cgo_lookup_unix.go` which calls `return lookupUnixUid(syscall.Getuid())` and so that path doesn't _use_ `$USER` but the current UID.

That'd explain why setting `$USER` doesn't work :-)

sweharris:
Hopefully I've updated and rebased properly (I hate git rebase; I don't do it often enough to feel confident!)

backend/sftp/sftp.go
ncw: I'd say more idiomatic for a type with a single case would be this, which also saves doing the type assertion more than once (the compiler may be clever enough to figure this out - I dont know!)

```go
cert, ok := pk.(*ssh.Certificate)
if !ok {
    return nil, errors.New("public key file is not a certificate file: " + pubkeyFile)
}
pubsigner, err := ssh.NewCertSigner(cert, signer)
```

backend/sftp/sftp.go
```diff
@@ -507,7 +512,39 @@ func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
...
+			// This is what we'll pass to the ssh client.
+			// Normally the ssh client will use the public key built
+			// into the private key, but we need to tell it to use the user
+			// specified public key cert.  This signer is specific to the
+			// cert and will include the private key signer.  Now ssh
+			// knows everything it needs.
+			switch pk.(type) {
+			case *ssh.Certificate: // This is the right type
+			default:
+				return nil, errors.New("Public key file is not a certificate file: " + pubkeyFile)
```
ncw, line 538: picky: the [Go style standard](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) states

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) 

This is so that the thing which does the final logging does `log.Printf("Failed doing thing: %v", err)` and we only get one capital.

docs/content/sftp.md
```diff
@@ -128,6 +128,17 @@ Using an ssh-agent is the only way to load encrypted OpenSSH keys at the moment.
...
+
+```
+[remote]
+type = sftp
+host = example.com
+user = sftpuser
+key_file = ~/id_rsa
+pubkey_file = ~/id_rsa-cert.pub
+````
+
```
ncw, line 141: Perhaps mention if you have a concatenated private key + cert, then mention it twice at `key_file` and `pubkey_file` - I think that is right isn't it?

backend/sftp/sftp.go
```diff
@@ -82,6 +82,9 @@ func init() {
 Only PEM encrypted key files (old OpenSSH format) are supported. Encrypted keys
 in the new OpenSSH format can't be used.`,
 			IsPassword: true,
+		}, {
+			Name: "pubkey_file",
+			Help: "Optional path to public key file; set this if you have a signed certificate you want to use for authentication." + env.ShellExpandHelp,
```
ncw, line 87: Can you split this over multiple lines? The first line is then used for the command line help - we don't want this to be too long, so something like

```go
Help: `Optional path to public key file.

Set this if you have a signed certificate you want to use for authentication.` + env.ShellExpandHelp,
```

Also, for discussion - should we set the `Advanced: true` flag here to push this into the advanced config in the config wizard?

ncw (commented):
> Hopefully I've updated and rebased properly (I hate git rebase; I don't do it often enough to feel confident!)

Looks good!

I looked through the code again and found some final polish things (nothing major) - see inline comments.

ncw:
> Looking at `sftp.go`, you're calling `user.Current()` to get the user name. If that fails you drop back to `$USER`.
> 
> Following the source chain, `user.Current()` in `os/user/lookup.go` calls `current()` to populate the cache. On Unix this is defined in `os/user/cgo_lookup_unix.go` which calls `return lookupUnixUid(syscall.Getuid())` and so that path doesn't _use_ `$USER` but the current UID.
> 
> That'd explain why setting `$USER` doesn't work :-)

Ah! Thanks for tracing that out. I think `currentUser()` should be factored out into `lib/env` and have a docs making workaround! I'll do that once we've landed this PR.

backend/sftp/sftp.go
sweharris: OK; I'm still a newbie at GoLang; I wasn't aware of that :-)  Changed!

docs/content/sftp.md
sweharris: Comment and example added.

backend/sftp/sftp.go
sweharris: I thought about this, but I don't think this should Advanced.  If a user has a cert then it's a simpler work flow to configure it straight off, and if they don't then they can just hit return to ignore.

ncw:
That looks great now - thank you very much :-)

I'll merge it now.

Issue #4645, opened 2020-10-01
Jottacloud authentication failing with 'cannot fetch token: 400 Bad Request'


gerph:
#### What is the problem you are having with rclone?

Using the backend Jottacloud, with the personal token for login, it works for about an hour or so and then starts failing with an error saying it cannot fetch token.

#### What is your rclone version (output from `rclone version`)

Installed from brew:

```
charles@laputa ~/backup (master) [1]> rclone version
rclone v1.53.1-DEV
- os/arch: darwin/amd64
- go version: go1.15.2
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

MacOS 10.14.6

####  Which cloud storage system are you using? (eg Google Drive)

Jottacloud, with cache and crypt.

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

`rclone copy file jottacloud:...`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

At the point at which it started happening I only have a log of it with `-v`:

2020/10/01 23:30:31 ERROR : Chris Wilkins/The Story Of The Commodore Amiga In Pixels (5175)/The Story Of The Commodore Amiga In Pixels - Chris Wilkins.pdf: put: error uploading: Post "https://api.jottacloud.com/files/v1/allocate": couldn't fetch token - maybe it has expired? - refresh with "rclone config reconnect jottacloud:": oauth2: cannot fetch token: 400 Bad Request

Repeating the command to get it to fail:

```
rclone -vv --checkers 10 --transfers 10 --bwlimit 1200 copy '/Users/charles/Calibre Library' jottacloud-cache-crypt:archives/2020-10/laputa/calibre 2>&1 | tee log
2020/10/01 23:44:37 DEBUG : rclone: Version "v1.53.1-DEV" starting with parameters ["rclone" "-vv" "--checkers" "10" "--transfers" "10" "--bwlimit" "1200" "copy" "/Users/charles/Calibre Library" "jottacloud-cache-crypt:archives/2020-10/laputa/calibre"]
2020/10/01 23:44:37 DEBUG : Creating backend with remote "/Users/charles/Calibre Library"
2020/10/01 23:44:37 DEBUG : Using config file from "/Users/charles/.config/rclone/rclone.conf"
2020/10/01 23:44:37 INFO  : Starting bandwidth limiter at 1.172MBytes/s
[...]

PR #4650, opened 2020-10-03
jottacloud: remove clientID/clientSecret from config when using v2 auth

buengese:
In #4645 we noticed jottacloud token refresh for v2 authentication failing due to oauth clientID and clientSecret being set in the config. Jottacloud v2 authentication uses the static value `jottacli` as clientID and no clientSecret at all and we set them as such but if different values are set config file oauthutil will override them. This happens when a user switched from v1 to v2 authentication because these values are left in the config file.
Given that there is probably good reason for oauthutil to have this behavior I've extended extended the configmap with Delete method so we can explicitly delete these from the config file. There would be other ways to fix this but I don't thing having a delete method for configmap hurts.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -373,6 +373,9 @@ func v2config(ctx context.Context, name string, m configmap.Mapper) {
 	fmt.Printf("Login Token> ")
 	loginToken := config.ReadLine()
 
+	m.Set(configClientID, "jottacli")
+	m.Set(configClientSecret, "")
+
 	token, err := doAuthV2(ctx, srv, loginToken, m)
 	if err != nil {
 		log.Fatalf("Failed to get oauth token: %s", err)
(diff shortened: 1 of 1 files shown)


backend/jottacloud/jottacloud.go
```diff
@@ -373,6 +373,11 @@ func v2config(ctx context.Context, name string, m configmap.Mapper) {
 	fmt.Printf("Login Token> ")
 	loginToken := config.ReadLine()
 
+	// We may have these left over from the legacy config. Token refresh will break if these are
+	// set so have to remove them from the config
+	m.Delete(configClientID)
```
ncw, line 378: It occurs to me that if just set them to empty string then most of this commit isn't needed.

In general the config system can't distinguish between empty string and key missing, so maybe we should just set them to empty and lose the rest of the commit?

buengese:
Yeah i realized that too after adding this code. I think setting this to empty is easier.

ncw:
Looks great :-) Please merge!

One for the stable branch?

PR #4652, opened 2020-10-04
sftp: Allow user to optionally check server hosts key to add security

sweharris:

Current behaviour is insecure.  If the user specifies this value then we
switch to validating the server hostkey and so can detect server changes
or MITM-type attacks.

Based on Issue 4087
  https://github.com/rclone/rclone/issues/4087

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -32,6 +32,7 @@ import (
 	"github.com/rclone/rclone/lib/readers"
 	sshagent "github.com/xanzy/ssh-agent"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/knownhosts"
 )
 
 const (
@@ -86,6 +87,16 @@ in the new OpenSSH format can't be used.`,
 			Help: `Optional path to public key file.
 
 Set this if you have a signed certificate you want to use for authentication.` + env.ShellExpandHelp,
+		}, {
+			Name: "known_hosts_file",
+			Help: `Optional path to known_hosts file.
+
+Set this value to enable server host key validation.` + env.ShellExpandHelp,
+			Advanced: true,
+			Examples: []fs.OptionExample{{
+				Value: "~/.ssh/known_hosts",
+				Help:  "Use OpenSSH's known_hosts file",
+			}},
 		}, {
 			Name: "key_use_agent",
 			Help: `When set forces the usage of the ssh-agent.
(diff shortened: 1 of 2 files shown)


backend/sftp/sftp.go
```diff
@@ -86,6 +87,12 @@ in the new OpenSSH format can't be used.`,
 			Help: `Optional path to public key file.
 
 Set this if you have a signed certificate you want to use for authentication.` + env.ShellExpandHelp,
+		}, {
+			Name: "knownhosts_file",
```
ncw, line 91: The file is called `~/.ssh/known_hosts` so we should probably call the config option `known_hosts_file`

backend/sftp/sftp.go
```diff
@@ -86,6 +87,12 @@ in the new OpenSSH format can't be used.`,
...
+			Help: `Optional path to known_hosts file.
+
+Set this value to enable server host key validation.` + env.ShellExpandHelp,
+			Advanced: true,
```
ncw, line 95: Maybe we should add an OptionExample here with

```
                       Examples: []fs.OptionExample{{
                                Value: "~/.ssh/known_hosts",
                                Help:  "Use OpenSSH's known_hosts file",
                        }},
```

?

ncw (commented):
I think that looks really good :-)

A couple of suggestions inline.

Also do you think this should be enabled automatically like it would be in OpenSSH? Pros and Cons?

backend/sftp/sftp.go
sweharris: But the OpenSSH configuration option is `UserKnownHostsFile` and `GlobalKnownHostsFile`, so it's all so confusing :-)   I can change the name if you think it's worth it.

backend/sftp/sftp.go
sweharris: I don't have any strong feelings, either way.  I think most people would work this out, anyway, and because of the subtle differences in implementation I'd really want them to read the docs, but I can add if you wish.

sweharris:
I deliberately didn't enable by default because
* backward compatibility with existing installs
* Problem with finding the default location (as mentioned in my comments; even worse for Termux or similar which may not put it in $HOME)
* Problem with multi-keys not parsing the same way

backend/sftp/sftp.go
sweharris: OK, change made.

backend/sftp/sftp.go
sweharris: Added

sweharris:
Hmm, that build failure looks like it might be a github infra issue 'cos it's in cmd/cmount which I'm not touching... and the same test passed on the other OS builds :-)

ncw:
> I deliberately didn't enable by default because
> 
> * backward compatibility with existing installs
> * Problem with finding the default location (as mentioned in my comments; even worse for Termux or similar which may not put it in $HOME)
> * Problem with multi-keys not parsing the same way

All good reasons. I thought it was worth enumerating them here so we can look at them later!

The `rclone config` wizard gives a good hint in the advanced settings now too on how to set it up if you are using openssh.

> Hmm, that build failure looks like it might be a github infra issue 'cos it's in cmd/cmount which I'm not touching... and the same test passed on the other OS builds :-)

Yes, that is one of the occasional fail tests which I haven't killed yet.

It all looks good to me now - thank you very much!

I'll merge it now.

PR #4653, opened 2020-10-04
Correction of repeated word

ameer1234567890:
Correction of repeated word


ncw:
Thank you!

Unfortunately the file you changed is an auto generated file - can you change this file instead?

https://github.com/rclone/rclone/blob/8e8ae1edc70d653a33db31f38dae561fb84c58e9/cmd/mountlib/mount.go#L165

Thanks

PR #4656, opened 2020-10-05
dropbox: speedup chunked uploads

SaveTheRbtz:
Dropbox chunked upload API consists of at least 3 requests: start, append, and finish.  It also allows uploading data along with the start and finish calls to minimize the roundtrip.  The problem lies in the fact that with a high concurrency (`--transfers=10` or more) there is a high chance that start/finish requests will fail with `429 too_many_requests` (This probability is way lower for append_v2.)

To avoid this problem I'm switching `uploadChunked` to not pass data on start and finish to minimize wasted bandwidth due to retries.  As a side effect this slightly increases the amount of API requests (though given that `uploadChunked` is only executed for large files this shouldn't be a bottleneck.)  And also reduces the amount of code, since chunking is limited to the append code.

PS. We'll also work on reducing chances of getting 429 on the server side and bring down the default 15s timeout on the `RetryAfter`.

cc: @rajatgoel


backend/dropbox/dropbox.go
```diff
@@ -1352,21 +1328,29 @@ func (o *Object) uploadChunked(in0 io.Reader, commitInfo *files.CommitInfo, size
...
+	currentChunk := 1
 	for {
-		if chunks > 0 && currentChunk >= chunks {
-			// if the size is known, only upload full chunks. Remaining bytes are uploaded with
-			// the UploadSessionFinish request.
-			break
-		} else if chunks == 0 && in.BytesRead()-cursor.Offset < uint64(chunkSize) {
-			// if the size is unknown, upload as long as we can read full chunks from the reader.
-			// The UploadSessionFinish request will not contain any payload.
+		if in.BytesRead() == cursor.Offset {
```
SaveTheRbtz, line 1342: @rajatgoel not sure about the exit condition here.

ncw:
I've found a much better solution to this problem, namely to use batches.

I haven't quite finished the code, but you can see the code in progress here: https://github.com/rclone/rclone/commits/fix-dropbox-batch

This completely removes the rate limiting problems.

Can you try that branch and tell me what you think?

SaveTheRbtz:
Using `FinishBatch` (and for that matter `CreateFolderBatch`) is indeed the best solution that we recommend to developers!  Thank you for doing this!!

I think we can combine both approaches: this diff will consolidate all the upload logic to single place (removing data from start and finish will make your diff simpler.)

ivandeex:
@SaveTheRbtz
Where are we with this PR?
It's not completely clear from the discussion :thinking:

SaveTheRbtz:
Sorry for terribly late reply.

The code in this diff looks like an improvement over the existing codebase: it is simpler and will work faster in case of server-side errors.  The only tricky part is the termination condition, so if anyone wants to review the diff -- it would be awesome.

ivandeex:
I've looked at your change (https://github.com/rclone/rclone/pull/4656/commits/c557a714e0718afbde4aba8078767a2cd1c5748c) and Nick's (https://github.com/rclone/rclone/commit/56e8d75cabe579030f6633b8ec82a1c8232d55d5, https://github.com/rclone/rclone/commit/c741c02fb67943d24423eb68f54c243d93ba8b05).

I see the problem in deciding whether two approaches combine or contradict. The answer needs some involvement so @SaveTheRbtz (as a Dropbox guy) or @ncw could answer that.

> I think we can combine both approaches: this diff will consolidate all the upload logic to single place (removing data from start and finish will make your diff simpler).

If the two approaches are "addable", a possible solution would be to rebase this PR on top of Nick's branch and do the test,
at least with Nick's CLI option turned off. The next problem is that both @ncw and @SaveTheRbtz are apparently rather busy at the moment.

cc @ncw

P.S. I'm volunteering to reduce some technical debt, e.g. janitor a number of stale PRs piled up here for the last year. In simpler cases I help to rebase or give advice. This PR needs more involvement than I can provide, alas. Sad to see good work rot.

ncw:
After 1.54 is released, I'm going to pick this PR up and integrate it with my other work on dropbox uploads.

I've got too much going on right at this minute but it is on the list.

rajatgoel:
Btw we have made some improvements to https://github.com/dropbox/dropbox-sdk-go-unofficial and updated it to latest spec. So, if you pick up those changes, you can also try using the new "concurrent" upload session functionality. Documentation is available in - https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-start.

In short, you can divide your file into multiple chunks of say 16MB and upload all these chunks in parallel. In our perf test we have seen 10x improvement in uploads (obviously a lot depends on client side factors like bandwidth, etc).

ivandeex:
@rajatgoel
I'm curious how do you in Dropbox solve the checksum issue with chunks uploaded in parallel?
Do you use something like Blake3 based on merkle tree or just leave the checksum blank?

rajatgoel:
@ivandeex Are you talking about `content_hash` field in FileMetadata response struct? The way that hash is defined - https://www.dropbox.com/developers/reference/content-hash allows us to compute it simply based on sha256 of 4MB chunks that make the file. To make this concurrent upload/hashing possible, we had to enforce the constraint that except for last chunk uploaded, all the other chunks need to be multiple of 4MB.

ivandeex:
Ah, you return an array of hashes. Thanks for info.

ncw:
@rajatgoel  wrote:
> Btw we have made some improvements to https://github.com/dropbox/dropbox-sdk-go-unofficial and updated it to latest spec. So, if you pick up those changes, you can also try using the new "concurrent" upload session functionality. Documentation is available in - https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-start.

This is new, I haven't seen that before! When did that hit the public API?

> In short, you can divide your file into multiple chunks of say 16MB and upload all these chunks in parallel. In our perf test we have seen 10x improvement in uploads (obviously a lot depends on client side factors like bandwidth, etc).

Oooh, that is exciting stuff! I'll put that on the list for the 1.54 revamp.

Do you have a recommended chunk size / number of concurrent uploads? What is the best performance?

Rclone will buffer the chunks in memory (in case of retries) so we'd normally choose relatively small chunks (the user will be able to control the size). If it turns out larger chunks are much better then we'd use those. Currently the chunks are 48MB but we only need to buffer 1 in memory at once.

With other backends we've found 3 or 4 concurrent uploads to be about right.

So maybe the defaults might be 3 concurrent uploads of 16MB chunks which would use 48 MB of memory in total per upload which would match the current memory use.

> Are you talking about `content_hash` field in FileMetadata response struct? The way that hash is defined - https://www.dropbox.com/developers/reference/content-hash allows us to compute it simply based on sha256 of 4MB chunks that make the file. To make this concurrent upload/hashing possible, we had to enforce the constraint that except for last chunk uploaded, all the other chunks need to be multiple of 4MB.

That has always struck me as a very clever design decision. Much better than S3's hash of hashes where the chunk size isn't defined.

ivandeex:
**[Offtopic]**

> That has always struck me as a very clever design decision. Much better than S3's hash of hashes where the chunk size isn't defined.

Actually "hash of hashes" is not bad generally (probably S3's is bad). The `merkle tree` thingy is in fact that:
- split your file in blocks of fixed small size (with last block probably smaller), calculate hashes of blocks
- concatenate sequential block hashes in groups of 2 or whatever (with last group probably smaller),
  get the 2-nd tier hashes
- proceed to next tier until there is only hash

The new [Blake3](https://github.com/BLAKE3-team/BLAKE3) hash is internally [constructed](https://github.com/oconnor663/bao/blob/master/docs/spec_0.9.1.md#tree-structure) this way with leaf block size of 1K (or 4K?).
Its main drawback is that it's very new and is not yet thoroughly analyzed or confirmed by worldwide crypto-gurus.

I think if one day we decide to implement parallel upload in chunker,
we can introduce `Blake3` support in chunker to have something to check
(or use the Dropbox approach with array of hashes, but what about a single file hash?).
I am biased towards the Blake3 approach because it provides a single 256-bit hash per file
and doesn't depend on predefined chunk size, just limits user chunk size to be modulo 1024.

The very new Linux [b3sum](https://aur.archlinux.org/packages/b3sum/) uitility can be used to verify Blak3
or we can add it as a new `rclone hashsum` option using the ready [module](https://github.com/lukechampine/blake3).

ivandeex:
**[Offtopic]**

> That has always struck me as a very clever design decision. Much better than S3's hash of hashes where the chunk size isn't defined.

@ncw
Actually "hash of hashes" is not bad generally (probably S3's is bad). The `merkle tree` thingy is in fact that:
- split your file in blocks of fixed small size (with last block probably smaller), calculate hashes of blocks
- concatenate sequential block hashes in groups of 2 or whatever (with last group probably smaller),
  get the 2-nd tier hashes
- proceed to next tier until there is only hash

The new [Blake3](https://github.com/BLAKE3-team/BLAKE3) hash is internally [constructed](https://github.com/oconnor663/bao/blob/master/docs/spec_0.9.1.md#tree-structure) this way with leaf block size of 1K (or 4K?).
Its main drawback is that it's very new and is not yet thoroughly analyzed or confirmed by worldwide crypto-gurus.

I think if one day we decide to implement parallel upload in chunker,
we can introduce `Blake3` support in chunker to have something to check
(or use the Dropbox approach with array of hashes, but what about a single file hash?).
I am biased towards the Blake3 approach because it provides a single 256-bit hash per file
and doesn't depend on predefined chunk size, just limits user chunk size to be modulo 1024.

The very new Linux [b3sum](https://aur.archlinux.org/packages/b3sum/) uitility can be used to verify Blak3
or we can add it as a new `rclone hashsum` option using the ready [module](https://github.com/lukechampine/blake3).

rajatgoel:
> This is new, I haven't seen that before! When did that hit the public API?

We added this support in Nov 2020 and it took some time for various SDKs to be updated. Few large customers are already using this feature with good success.

> Do you have a recommended chunk size / number of concurrent uploads? What is the best performance?

Your suggestion of 16MB chunk size and 3 to 4 parallel uploads should work pretty well. In our perf tests, we can upload 1GB file in ~6sec with 16MB chunk size and 64 uploads in parallel (but this perf test daemon is not disk, memory or network bounded. So, this is probably the best case you will see).

ncw:
@ivandeex wrote:
> **[Offtopic]**
> 
> > That has always struck me as a very clever design decision. Much better than S3's hash of hashes where the chunk size isn't defined.
> 
> @ncw
> Actually "hash of hashes" is not bad generally (probably S3's is bad).

The problem with S3's hash of hashes is that they aren't on fixed sized chunks so you can't verify the hash without knowing what size the chunks were uploaded with..

> The [merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) thingy is in fact that:
> 
> * split your file in blocks of fixed small size (with last block probably smaller), calculate hashes of blocks
> * concatenate sequential block hashes in groups of 2 or whatever (with last group probably smaller),
>   get the 2-nd tier hashes
> * proceed to next tier until there is only hash
> 
> The new [Blake3](https://github.com/BLAKE3-team/BLAKE3) hash is internally [constructed](https://github.com/oconnor663/bao/blob/master/docs/spec_0.9.1.md#tree-structure) this way with leaf block size of 1K (or 4K?).
> Its main drawback is that it's very new and is not yet thoroughly analyzed or confirmed by worldwide crypto-gurus.

I didn't know that about Blake3 - that does seem like a genuine advance in hashing. Wow it is fast!

If we are using hashes for integrity protection then the crypto strength of them is of secondary importance. For an example of this see the [Onedrive QuickXorHash](https://github.com/rclone/rclone/blob/master/backend/onedrive/quickxorhash/quickxorhash.go)

> I think if one day we decide to implement parallel upload in chunker,
> we can introduce `Blake3` support in chunker to have something to check
> (or use the Dropbox approach with array of hashes, but what about a single file hash?).
> I am biased towards the Blake3 approach because it provides a single 256-bit hash per file
> and doesn't depend on predefined chunk size, just limits user chunk size to be modulo 1024.

I like it :-)

ncw:
@rajatgoel wrote:
> > This is new, I haven't seen that before! When did that hit the public API?
> 
> We added this support in Nov 2020 and it took some time for various SDKs to be updated. Few large customers are already using this feature with good success.

Ah ha!

> > Do you have a recommended chunk size / number of concurrent uploads? What is the best performance?
> 
> Your suggestion of 16MB chunk size and 3 to 4 parallel uploads should work pretty well. In our perf tests, we can upload 1GB file in ~6sec with 16MB chunk size and 64 uploads in parallel (but this perf test daemon is not disk, memory or network bounded. So, this is probably the best case you will see).

In a week or two I will get testing. I know that dropbox uploads are a problem for rclone users right now and this will help a lot (there have been complaints about the pause between two blocks when uploading in the past).

Thanks

mauricev:
I just tried rclone with this commit on a Mac and it immediately had a fit and threw up its hands
```
uploaderfile5: corrupted on transfer: sizes differ 20687683584 vs 0
```

backend/dropbox/dropbox.go
SaveTheRbtz: Yeah, that was a wrong one!  Fixed =)

SaveTheRbtz:
@ivandeex @ncw Cleaned up, fixed, rebased.  Now ready to be reviewed.  This diff _slightly_ improves upload speed of medium-sized files (2-5 chunks (default 100-200M)) during lock contention (since it does not put data into start/finish requests.)

It also generally simplifies code (removes more code than it adds) and prepares code for the use of `UploadSessionType.concurrent` (it forbids sending data in start/finish.)

@mauricev can you please re-test?

mauricev:
Your changes fixed the bug I reported yesterday.

ncw:
Sorry for the long delay in addressing this PR

I have finally worked out the architecture for the batched upload for rclone. It has been quite tricky fitting fitting the batches and I ended up with two possible batch methods, sync and async.  sync waits for the batch response so the checksum can be checked. async doesn't.

This introduces 3 new flags which you can see documented in the commit 64427951b48bec361fa9fd50251c1192f758b9a0

[v1.55.0-beta.5331.64427951b.fix-dropbox-batch-sync](https://beta.rclone.org/branch/fix-dropbox-batch-sync/v1.55.0-beta.5331.64427951b.fix-dropbox-batch-sync/) on branch [fix-dropbox-batch-sync](https://github.com/rclone/rclone/tree/fix-dropbox-batch-sync) (uploaded in 15-30 mins)

So my question is, do we need this PR as well as batching?

My understanding is that this PR essentially works around the lack of batching in rclone trying to avoid the `429 too_many_requests` errors by doing smaller requests first.

So this PR will cause rclone to do more API roundtrips in the non-errored case.

So if, as my testing seems to show, that batching eliminates the `429 too_many_requests` errors do we need this PR?

I'd appreciate your thoughts on this.

I've made an issue #5156 to track the dropbox upload improvements.

rajatgoel:
This patch takes us one step closer to concurrent transfers. Concurrent transfers don't allow users to upload data with /start and /finish APIs and this patch makes that the default behavior. So, we don't strictly need this patch as-is but if we want to use concurrent uploads for large files (highly recommended), we will need to make some change along these lines.

ncw:
> This patch takes us one step closer to concurrent transfers. Concurrent transfers don't allow users to upload data with /start and /finish APIs and this patch makes that the default behavior. So, we don't strictly need this patch as-is but if we want to use concurrent uploads for large files (highly recommended), we will need to make some change along these lines.

I see - that is a good reason - thank you.

Do you think you can rebase it onto the [fix-dropbox-batch-sync](https://github.com/rclone/rclone/tree/fix-dropbox-batch-sync) branch? This has changed a lot of the same code so I'm afraid it will be more of a re-write rather than a rebase :-(

This will fit in well with the plan outlined in #5156 to implement batching and async uploads for v1.56

Thanks

SaveTheRbtz:
Ok. I'll work on it this weekend.

ncw:
> Ok. I'll work on it this weekend.

Thank you :-)

SaveTheRbtz:
Sorry, didn't have time to work it this weekend, let me try to get to it in the next couple of days.

SaveTheRbtz:
@ncw rebased on top of 69b091f34.

This may not be as efficient as original but it has a couple of advantages:
* code is smaller and arguably simpler: `31 insertions(+), 88 deletions(-)`
* no data uploads in start/finish, so it is now compatible with the `concurrent` upload mode.

This code can be simplified even further [if we know the size of the data before we send it](https://gist.github.com/SaveTheRbtz/a246b8d8c765038c248969c77477933b) but it will break bandwidth estimation.

backend/dropbox/dropbox.go
```diff
@@ -1597,84 +1597,37 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
-		res, err = o.fs.srv.UploadSessionStart(&arg, chunk)
+		res, err = o.fs.srv.UploadSessionStart(&files.UploadSessionStartArg{}, nil)
 		return shouldRetry(ctx, err)
 	})
 	if err != nil {
 		return nil, err
 	}
 
+	chunkSize := int64(o.fs.opt.ChunkSize)
+	chunks := int(math.Ceil(float64(size) / float64(chunkSize)))
```
ncw, line 1611: Floating point math? I'd have probably written this (modulo some casts!)

```go
chunks, remainder := size / chunkSize, size % chunkSize
if remainder > 0 {
    chunks++
}
```

The floating point arithmetic starts being inexact for size > 2**56 which is 64 PiB so it is probably academic!

Also the chunks could overflow an `int32` on 32 bit platforms which is also academic since we are going to die of heat death of the universe before we've uploaded 2**32 chunks.

backend/dropbox/dropbox.go
```diff
@@ -1597,84 +1597,37 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
+			fs.Debugf(o, "Uploading chunk %d/%d", currentChunk, chunks)
 		}
-		cursor.Offset = in.BytesRead()
-		fmtChunk(currentChunk, false)
-		chunk = readers.NewRepeatableLimitReaderBuffer(in, buf, chunkSize)
+
+		chunk := readers.NewRepeatableLimitReaderBuffer(in, buf, chunkSize)
 		err = o.fs.pacer.Call(func() (bool, error) {
 			// seek to the start in case this is a retry
 			if _, err = chunk.Seek(0, io.SeekStart); err != nil {
```
ncw, line 1633: Grr, GitHub reviews won't let me write a comment on the correct line but a bit below this you'll see

```go
			// after the first chunk is uploaded, we retry everything
			return err != nil, err
```

This is no longer correct since we aren't doing a "first" block.

It should be something like

```go
if currentChunk == 1 {
     		return shouldRetry(ctx, err)
}
// after the first chunk is uploaded, we retry everything
return err != nil, err
```

backend/dropbox/dropbox.go
```diff
@@ -1597,84 +1597,37 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 // unknown (i.e. -1) or smaller than uploadChunkSize, the method incurs an
```
ncw, line 1597: This comment is no longer correct :-)

ncw (commented):
I've been through the code and I think it looks great - thanks!

Much simpler and a step towards concurrent uploads which is great.

I put some minor nit-picks inline for you to look at.

Thank you

ncw:
> @ncw rebased on top of [69b091f](https://github.com/rclone/rclone/commit/69b091f34d765e2aff66f666f6323d6f207e773c). [Diff](https://gist.github.com/SaveTheRbtz/c12739b005ccdddbc90c6899112f1173).
> 
> This may not be as efficient as original but it has a couple of advantages:
> 
> * code is smaller and arguably simpler: `31 insertions(+), 88 deletions(-)`
> * no data uploads in start/finish, so it is now compatible with the `concurrent` upload mode.

I've done a review for that above.
 
> This code [can be simplified even further](https://gist.github.com/SaveTheRbtz/a246b8d8c765038c248969c77477933b) if we know the size of the data before we send it but chunk buffering will break short-term bandwidth estimation and likely increase memory consumption.

I think we can have the same effect and keep the bandwidth estimation by doing `in.BytesRead() < uint64(chunkSize)` as the equivalent of the EOF signal?

backend/dropbox/dropbox.go
SaveTheRbtz: I thought that `shouldRetry` was mostly needed because first chunk was the part of the `UploadSessionStart` call.  Basically, the assumption is that after the session is started everything is retriable.  If that's not correct, I can change it to the suggested code of only retrying the first chunk (though this kind of a special case looks a bit strange.)

SaveTheRbtz:
Sadly, the number of bytes read is only available after the read is done, but we need to know it before we call Append (to set `Close`).  I do use this method for unknown sizes (since there is no other way) but it does add an Append call with no data just for this `Close`.

backend/dropbox/dropbox.go
ncw: I see what you mean. Maybe just the comment is misleading now!

ncw (approved):
That is looking great now - thank you!

I'm going to pull this into the dropbox-batch-sync branch and close the commit.

Unfortunately there is no way reassign which branch a PR targets in GitHub so I'll close this PR when I've done that.

I'll rebase that branch and push it up again.

ncw:
This is merged to 

[v1.56.0-beta.5394.d16702566.fix-dropbox-batch-sync](https://beta.rclone.org/branch/fix-dropbox-batch-sync/v1.56.0-beta.5394.d16702566.fix-dropbox-batch-sync/) on branch [fix-dropbox-batch-sync](https://github.com/rclone/rclone/tree/fix-dropbox-batch-sync) (uploaded in 15-30 mins)


Thank you!

I hope to merge this to master shortly in time for 1.56

PR #4663, opened 2020-10-07
Correction of repeated word

ameer1234567890:
Correction of repeated word


ncw:
Super - thank you :-)

I'll merge this now.

Issue #4660, opened 2020-10-06
sftp should only ask for a password after the SSH connection has succeeded


sweharris:
With `known_hosts_file` changes, the user may be prompted for a password but then fail to connect because the host key is incorrect.  It's a better user experience to have the user prompted only if the SSH session succeeds.  This would also help if the remote server is down or unreachable for any reason.

It's also a minor security enhancement because it means the user doesn't need to enter their password more often than is necessary.

sweharris:
This can be achieved by changing the AskPassword conditional to call `ssh.PasswordCallback()` and have that prompt for the password. 

So now if the `known_hosts` entry is wrong the user never gets prompted.

What's interesting is that if the connection succeeds but the password is wrong then the user gets _reprompted_ to enter the password:

eg

```
$ ./rclone --config /tmp/testsftp/rclone.conf ls test2:/tmp/test
Enter SFTP password: 
Enter SFTP password: 
Enter SFTP password: 
        0 this_is_on_test2
```

This is a change in behaviour, where the user previously just got thrown out, but I think it's a worthwhile change and definitely enhances interactive behaviour.

ncw:
> This is a change in behaviour, where the user previously just got thrown out, but I think it's a worthwhile change and definitely enhances interactive behaviour.

Yes that looks much better.

ncw:
Thank you for doing this - these little improvements really add up to making rclone more usable. 

I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.54

zanna-37:
Something weird is happening after this. rclone keeps asking me the SFTP password after the previous time (so during operations). My guess is that when the connection gets closed for some reason it asks the password again.

See row 3
```
2020/10/08 14:08:25 DEBUG : /folder/folder: >Getattr: errc=0
2020/10/08 14:08:25 DEBUG : /folder/file.extension: Getattr: fh=0xFFFFFFFFFFFFFFFF
Enter SFTP password: 2020/10/08 14:08:25 DEBUG :/folder/file.extension: >Getattr: errc=-2
2020/10/08 14:08:25 DEBUG : /folder: Releasedir: fh=0x3
2020/10/08 14:08:25 DEBUG : /folder: >Releasedir: errc=0
```

sweharris:
Hmm, it will, yes.  

[...]

PR #4664, opened 2020-10-08
sftp: remember entered password in AskPass mode

sweharris:

After switching to a password callback function, if the ssh connection
aborts and needs to be reconnected then the user is-reprompted for their
password.  Instead we now remember the password they entered and just give
that back.  We do lose the ability for them to correct mistakes, but that's
the situation from before switching to callbacks.  We keep the benefits
of not asking for passwords until the SSH connection succeeds (right
known_hosts entry, for example).

  https://github.com/rclone/rclone/issues/4660#issuecomment-705502792

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -235,6 +235,7 @@ type Fs struct {
 	poolMu       sync.Mutex
 	pool         []*conn
 	pacer        *fs.Pacer // pacer for operations
+	savedpswd    string
 }
 
 // Object is a remote SFTP file that has been stat'd (so it exists, but is not necessarily open for reading)
@@ -413,6 +414,10 @@ func (f *Fs) putSftpConnection(pc **conn, err error) {
 // NewFs creates a new Fs object from the name and root. It connects to
 // the host specified in the config file.
 func NewFs(name, root string, m configmap.Mapper) (fs.Fs, error) {
+	// This will hold the Fs object.  We need to create it here
+	// so we can refer to it in the SSH callback, but it's populated
+	// in NewFsWithConnection
+	f := &Fs{}
 	ctx := context.Background()
 	// Parse config into Options struct
 	opt := new(Options)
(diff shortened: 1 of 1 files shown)


backend/sftp/sftp.go
```diff
@@ -44,6 +44,7 @@ const (
 
 var (
 	currentUser = env.CurrentUser()
+	savedpswd   = ""
```
ncw, line 47: This variable needs to be stored inside `Fs` in case there is more than one sftp backend active

ncw (commented):
Thanks for doing this :-)

See inline for something which needs fixing.

backend/sftp/sftp.go
sweharris: OK, I'm not sure how to do that.  

Since `getPass()` is called via an SSH callback, I don't know how to get access to the Fs object.  All the other variables stored there appear to be called within the scope of a `(f *Fs)` function and so can be accessed by `f.variable`, but we don't seem to have that option, here!

I guess I could re-use `config.Pass` since we know the user hadn't specified a password ?  Or is that also a bad idea?

ncw: I would make `getPass` into a method of `Fs`

```go
func (f *Fs) getPass() (string, error) {
```

Then you can use `f` within the function.

You can then pass the bound method `f.getPass` rather than `getPass` as the callback.

I hope that helps!

sweharris: Except `f` isn't defined in `NewFS` (where the `GetPass` method is referenced) because we don't _have_ an `Fs` object yet...

sweharris: I wonder if a simpler hack might be to define `f := &Fs{}` in `NewFS` so now I can refer to `f.getPass` in `NewFS`, and pass `f` as a parameter to a modified `func NewFsWithConnection(f *Fs,....)` and modify the initialiser to be `f.name =...` type stuff.  

+       return NewFsWithConnection(f, ctx, name, root, m, opt, sshConfig)

Let's test...

Hmm, that seems to work!

You're right it probably needs refactoring, but this'll may work around the problem.

ncw:
That looks like it works!

I'll merge that now - thank you.

> Except golint wanted ctx first, heh :-)

This is a Google style guide thing...

PR #4668, opened 2020-10-10
crypt: small simplification, no functionality change

dan-hipschman-od:

This is just a very minor simplification of some code in the `crypt` backend. There's no change to functionality or security, it just removes a few lines of unneeded code.

Basically, it looks like we're copying the nonce into a buffer and then immediately overwriting it. So we don't need to do that. Note, I'm not a Go programmer, so let me know if I'm missing something here.

No, I was just looking through the code and noticed it. The change is so small I didn't think it needed discussion.

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate. (I didn't add any new tests because there's no new functionality.)

Thanks for a great tool!

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -633,11 +633,8 @@ func (fh *encrypter) Read(p []byte) (n int, err error) {
 		}
 		// possibly err != nil here, but we will process the
 		// data and the next call to ReadFull will return 0, err
-		// Write nonce to start of block
-		copy(fh.buf, fh.nonce[:])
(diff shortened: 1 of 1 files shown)


ncw (approved):
Interesting spot - thank you :-)

I went through the code history and that has been there since the first commit of the crypt backend.

I conjecture that it looks like that as a remnant from a refactor during development of the backend.

I think your change looks good and will shave off a few uS per block so that is a win :-)

ncw:
Thank you :-)

PR #4669, opened 2020-10-10
genautocomplete: add support to output to stdout

lasombra:
## What is the purpose of this change?

Fix issue #3423 

Yes, https://github.com/rclone/rclone/issues/3423

#### Checklist

- [`x`] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [`x`] I have added tests for all changes in this PR if appropriate.
- [`x`] I have added documentation for the changes if appropriate.
- [`x`] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [`x`] I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/cmd/genautocomplete/genautocomplete_bash.go
+++ b/cmd/genautocomplete/genautocomplete_bash.go
@@ -2,6 +2,7 @@ package genautocomplete
 
 import (
 	"log"
+	"os"
 
 	"github.com/rclone/rclone/cmd"
 	"github.com/spf13/cobra"
@@ -29,11 +30,20 @@ them directly
 
 If you supply a command line argument the script will be written
 there.
+
+If output_file is "-", then the output will be written to stdout.
 `,
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(0, 1, command, args)
 		out := "/etc/bash_completion.d/rclone"
 		if len(args) > 0 {
+			if args[0] == "-" {
+				err := cmd.Root.GenBashCompletion(os.Stdout)
+				if err != nil {
+					log.Fatal(err)
+				}
+				return
+			}
 			out = args[0]
 		}
 		err := cmd.Root.GenBashCompletionFile(out)
--- a/cmd/genautocomplete/genautocomplete_test.go
+++ b/cmd/genautocomplete/genautocomplete_test.go
@@ -11,8 +11,10 @@ import (
 func TestCompletionBash(t *testing.T) {
 	tempFile, err := ioutil.TempFile("", "completion_bash")
 	assert.NoError(t, err)
-	defer func() { _ = tempFile.Close() }()
-	defer func() { _ = os.Remove(tempFile.Name()) }()
+	defer func() {
+		_ = tempFile.Close()
+		_ = os.Remove(tempFile.Name())
+	}()
 
 	bashCommandDefinition.Run(bashCommandDefinition, []string{tempFile.Name()})
 
@@ -21,11 +23,32 @@ func TestCompletionBash(t *testing.T) {
 	assert.NotEmpty(t, string(bs))
 }
 
+func TestCompletionBashStdout(t *testing.T) {
+	originalStdout := os.Stdout
+	tempFile, err := ioutil.TempFile("", "completion_zsh")
+	assert.NoError(t, err)
+	defer func() {
+		_ = tempFile.Close()
+		_ = os.Remove(tempFile.Name())
+	}()
+
+	os.Stdout = tempFile
+	defer func() { os.Stdout = originalStdout }()
+
+	bashCommandDefinition.Run(bashCommandDefinition, []string{"-"})
+
+	output, err := ioutil.ReadFile(tempFile.Name())
+	assert.NoError(t, err)
+	assert.NotEmpty(t, string(output))
+}
+
 func TestCompletionZsh(t *testing.T) {
 	tempFile, err := ioutil.TempFile("", "completion_zsh")
 	assert.NoError(t, err)
-	defer func() { _ = tempFile.Close() }()
-	defer func() { _ = os.Remove(tempFile.Name()) }()
+	defer func() {
+		_ = tempFile.Close()
+		_ = os.Remove(tempFile.Name())
+	}()
 
 	zshCommandDefinition.Run(zshCommandDefinition, []string{tempFile.Name()})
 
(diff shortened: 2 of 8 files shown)


lasombra:
Hi @ncw, I think the issue on Windows seems related to the test's stdout redirection, but I'm not sure why. Any ideas?

cmd/genautocomplete/genautocomplete_test.go
```diff
@@ -21,6 +21,23 @@ func TestCompletionBash(t *testing.T) {
...
+	os.Stdout = writeOut
+	bashCommandDefinition.Run(bashCommandDefinition, []string{"-"})
+
+	_ = writeOut.Close()
+	output, err := ioutil.ReadAll(readOut)
+
+	assert.NoError(t, err)
+	assert.NotEmpty(t, string(output))
+
+	os.Stdout = originalStdout
```
ncw, line 38: This should be in a `defer() { }()` just below `os.Stdout = writeOut`

cmd/genautocomplete/genautocomplete_test.go
```diff
@@ -21,6 +21,23 @@ func TestCompletionBash(t *testing.T) {
 	assert.NotEmpty(t, string(bs))
 }
 
+func TestCompletionBashStdout(t *testing.T) {
+	originalStdout := os.Stdout
+	readOut, writeOut, errorPipe := os.Pipe()
```
ncw, line 26: The test failure is caused by the Pipe blocking.

It only has a limited buffer (maybe none) so unless you read what is coming out, you can't add more.

If you run the `bashCommandDefinition.Run(bashCommandDefinition, []string{"-"})` in a go routine (with some synchronisation) then the read and write will proceed in parallel and the pipe will be OK.

cmd/genautocomplete/genautocomplete_bash.go
```diff
@@ -29,12 +30,23 @@ them directly
...
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(0, 1, command, args)
 		out := "/etc/bash_completion.d/rclone"
 		if len(args) > 0 {
-			out = args[0]
+			if args[0] == "-" {
+				err := cmd.Root.GenBashCompletion(os.Stdout)
+				if err != nil {
+					log.Fatal(err)
+				} else {
```
ncw, line 44: Can you lose the `} else {` here? It is much easier to see the code always returns without it.

So

```go
if err != nil {
	log.Fatal(err)
}
return
```

Same in the other 2 places.

ncw (commented):
See inline for a couple of notes and a hint as to why the test fails on Windows.

Thank you for doing this :-)

cmd/genautocomplete/genautocomplete_test.go
lasombra: An empty `defer`? What does this accomplish?

cmd/genautocomplete/genautocomplete_bash.go
lasombra: Sure.

cmd/genautocomplete/genautocomplete_test.go
lasombra: Looking at https://golang.org/pkg/os/#Pipe, there's no buffer. `os.Pipe()` returns connected `Files`, which I imagine are in-memory file references.

So, after writing to the "in-memory" files, the write file is closed and and read file is read here:
`output, err := ioutil.ReadAll(readOut)`.

I can't imagine why `os.Pipe()` would be blocking just on Windows. Can you?

cmd/genautocomplete/genautocomplete_test.go
lasombra: I feel so stupid sometimes. These 3 selected line or just the assignment?

cmd/genautocomplete/genautocomplete_test.go
ncw: os.Pipe calls syscall.Pipe2 so is using an OS specific pipe mechanism. According the the man page

>  Data  written  to  the write end of the pipe is buffered by the
       kernel until it is read from the read end of the pipe.  For further de‐
       tails, see pipe(7).

So there is some kind of buffer.

> I can't imagine why os.Pipe() would be blocking just on Windows. Can you?

I imagine the buffer is smaller or something like that.

You could redirect stdout to a file instead of using a pipe?

lasombra:
@ncw The test failure isn't related to my PR. Your suggestion worked like a charm and tests are passing now. :-)

cmd/genautocomplete/genautocomplete_test.go
```diff
@@ -21,6 +21,23 @@ func TestCompletionBash(t *testing.T) {
 	assert.NotEmpty(t, string(bs))
 }
 
+func TestCompletionBashStdout(t *testing.T) {
+	originalStdout := os.Stdout
+	tempFile, err := ioutil.TempFile("", "completion_zsh")
+	assert.NoError(t, err)
+	defer func() { _ = tempFile.Close() }()
```
ncw, line 28: These actually get run in the reverse order so the Remove will run before the Close...

This will work fine on Linux but will cause the file to fail to be removed on Windows.

I'd probably write this for the avoidance of doubt!

```go
defer func() {
    _ = tempFile.Close()
    _ = os.Remove(tempFile.Name())
}()
```

cmd/genautocomplete/genautocomplete_test.go
```diff
@@ -21,6 +21,23 @@ func TestCompletionBash(t *testing.T) {
...
+	defer func() { _ = os.Remove(tempFile.Name()) }()
+
+	os.Stdout = tempFile
+	bashCommandDefinition.Run(bashCommandDefinition, []string{"-"})
+
+	output, err := ioutil.ReadFile(tempFile.Name())
+	assert.NoError(t, err)
+	assert.NotEmpty(t, string(output))
+
+	defer func() { os.Stdout = originalStdout }()
```
ncw, line 38: This should go under the `os.Stdout` change - if any function panics then it will be undone which is useful when `os.Stdout` is concerned!  So


```go
	os.Stdout = tempFile
	defer func() { os.Stdout = originalStdout }()
```

ncw (commented):
Tests are looking good now - well done!

I put inline some comments about `defer` usage.

Don't worry about the failing Windows test - that is something unrelated.

cmd/genautocomplete/genautocomplete_test.go
lasombra: Ah, interesting stuff. I have to read more about `defer`. I was under the impression it would, in case of panic, execute all `defer` functions in order, no matter where in the code `defer` is.

cmd/genautocomplete/genautocomplete_test.go
lasombra: OK, makes sense. I corrected the other test methods as well.

lasombra:
@ncw many thanks for the patience and help getting this PR a 10/10 👍

lasombra:
@ncw Sorry for sounding anxious, but is there anything left that needs to be done?

PR #4672, opened 2020-10-11
new backend: zoho workdrive - fixes #4533

### LINKED ISSUE #4533: Zoho WorkDrive / Docs Support
rclone v1.52.3
- os/arch: windows/amd64
- go version: go1.14.7

Add Zoho WorkDrive to supported cloud providers.

Here is an API: https://writer.zohopublic.com/writer/published/ankqibdb934759c6446bca8899b2f426428fc

While Zoho have TrueSync Beta (Streaming) and WorkDrive Sync (Full downloads) it would be good if rclone can integrate it as a source/dest without having to mount the drive using the Zoho sync tools.

Thanks!

buengese:
Took longer than I thought. Zoho has some annoying quirks - see review comments. Other than that this pretty much done.

#### Checklist


backend/zoho/zoho.go
```diff
@@ -0,0 +1,1223 @@
...
+	if isSimpleName(leaf) {
+		info, err := f.upload(ctx, f.opt.Enc.FromStandardName(leaf), directoryID, size, in, options...)
+		if err != nil {
+			return nil, err
+		}
+		return f.newObjectWithInfo(ctx, remote, info)
+	}
+
+	tempName := "rcloneTemp" + random.String(8)
+	info, err := f.upload(ctx, tempName, directoryID, size, in, options...)
```
buengese, line 678: Zoho's upload endpoint has problems with special characters even though they are supported by every other API endpoint. For uploads, the filename is supplied both as a multipart param and an URL parameter. The multipart filename parameter cannot handle `/`, `\`, `"` and `'` and no way of escaping these characters seems to work. The URL parameter also has problems with an even larger set of special characters even if they are-URL encoded. 
This forced me to implement a workaround that uploads files with a temporary name and rename them afterward. 
The `isSimpleName` path is a shortcut without rename used for files that only contain alphanumeric characters and `.` which should be the majority of filenames anyway.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1223 @@
...
+	// Only set ID here because response is not complete Item struct
+	dstObject.id = item.ID
+
+	// Can't move and change name in one step so we have to check if we have
+	// the correct name after move
+	if f.opt.Enc.ToStandardName(item.Attributes.Name) != dstFilename {
+		if err = dstObject.rename(ctx, dstFilename); err != nil {
+			return nil, errors.Wrap(err, "move: couldn't rename moved file")
+		}
+	}
```
buengese, line 946: Zoho cannot move and rename an object or directory in a single API call. Renaming must always be done separately. I've decided to go with move first and rename afterward.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1223 @@
...
+// It attempts to read the objects mtime and if that isn't present the
+// LastModified returned in the http headers
+func (o *Object) ModTime(ctx context.Context) time.Time {
+	return o.modTime
+}
+
+// SetModTime sets the modification time of the local fs object
+func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
+	return fs.ErrorCantSetModTime
+}
```
buengese, line 1088: Zoho has no way at all to write modification times, neither during upload nor as a separate API call.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1223 @@
...
+		if start > 0 {
+			// We need to read and discard the beginning of the data...
+			_, err = io.CopyN(ioutil.Discard, resp.Body, start)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// ... and return a limited reader for the remaining of the data
+		return readers.NewLimitedReadCloser(resp.Body, end-start), nil
+	}
```
buengese, line 1154: Zoho ignores range requests, at least for all the common ways to do these, so I've added this workaround.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1223 @@
...
+}
+
+// Precision return the precision of this Fs
+func (f *Fs) Precision() time.Duration {
+	return fs.ModTimeNotSupported
+}
+
+// Hashes returns the supported hash sets.
+func (f *Fs) Hashes() hash.Set {
+	return 0
```
buengese, line 261: Zoho does not support any checksums.

backend/zoho/zoho.go
ivandeex: Zoho becomes a third candidate for a future fs.Feature flag codenamed "anti-ranger, bandwidth eater" along with mailru and seafile. See discussion here https://github.com/rclone/rclone/pull/4607#issuecomment-694810408

backend/zoho/zoho.go
ncw: Grrr, that is annoying! Your workaround seems like an efficient one.

backend/zoho/zoho.go
ncw: There is a little dance you need to do here. I see from the code that you are doing that already! Eg

https://github.com/rclone/rclone/blob/fc5b14b62046b240efd5963bb62a29ae47c0a747/backend/sharefile/sharefile.go#L921-L926

backend/zoho/zoho.go
ncw: :-( That is a shame... Is there some metadata we could use?

backend/zoho/zoho.go
ncw: > Zoho becomes a third candidate for a future fs.Feature flag codenamed "anti-ranger, bandwidth eater"

:-)

I think you are right.

backend/zoho/zoho.go
ncw: `return hash.Set(hash.None)` is what we normally write

ncw:
How are you doing with the integration tests?

From backend root

```
go install ./...
test_all -backends zoho
```

Are you ready for me to review the code in detail?

backend/zoho/zoho.go
ivandeex: @ncw
Are there other backends without modtime? Are they marked by a feature flag?
We will need modtime for a future chunker garbage collector.

backend/zoho/zoho.go
buengese: Ah true I should be using a different name during the move, didn't think of that.

backend/zoho/zoho.go
buengese: @ivandeex there are a few. They are listed in `docs/content/overview.md`. I don't think there is a feature flag right now.

backend/zoho/zoho.go
buengese: I can't seem to resist the urge to use simple int and binary operations here. :)

buengese:
> How are you doing with the integration tests?
> 
> From backend root
> 
> ```
> go install ./...
> test_all -backends zoho
> ```
They pass. I'm pretty familiar with them by now, this isn't my first backend after all.

backend/zoho/zoho.go
ncw: Really? That is strange. Maybe it it is supporting only a latin1 subset wrapped in UTF-8 or something like that?

backend/zoho/zoho.go
ncw: You can tell backends without modtime by looking at the `Precision()` returned from the Fs. Though a feature flag isn't a bad idea too.

Custom metadata is a two edged sword - it can be great, but then it is something potentially non-standard. Does the metadata get returned in the listings? If it does then there won't be a performance impact reading it. If you have to query each object to read the metadata then it might not be worth doing.

ncw:
> They pass. I'm pretty familiar with them by now, this isn't my first backend after all.

That is great news :-)

Do you want to finish of the todo list (docs etc) before I do a code review?

buengese:
Yeah and it's done now.

ivandeex:
@ncw I wonder if default `--multithread-streams=4` applies to backends with fake range requests like this, mailru or seafile (and quadruples bandwidth).

buengese:
@ncw you may review this now if you have the time.

ncw:
@ivandeex wrote:
> I wonder if default `--multithread-streams=4` applies to backends with fake range requests like this, mailru or seafile (and quadruples bandwidth).

It does and it shouldn't! You made a suggestion for a new Feature flag maybe `CantRange` or `InefficientRange`?

I think that is a great idea, and using it to disable multithread streams is a great idea.

ncw:
@buengese wrote:
> you may review this now if you have the time.

I'll take a look now :-)

ncw:
> @buengese wrote:
> 
> > you may review this now if you have the time.
> 
> I'll take a look now :-)

I was just giving it a go - I signed up for a trial Zoho account.

When do the oauth I get
- Invalid Client, Client ID passed does not exist - are the client_ids per account do you think?
- Do they do free accounts or am I going to have to beg for a free one for the integration test server?

buengese:
>Invalid Client, Client ID passed does not exist - are the client_ids per account do you think?
Huh. I thought they were global but they might actually be per organisation. 
>Do they do free accounts or am I going to have to beg for a free one for the integration test server?
I'm actually not sure about this. I was actually asked by a friend to implement this and was therefore provided with an account.

backend/zoho/api/types.go
```diff
@@ -0,0 +1,118 @@
...
+// Zoho uses milliseconds since unix epoch (Java currentTimeMillis)
+type Time time.Time
+
+// UnmarshalJSON turns JSON into a Time
+func (t *Time) UnmarshalJSON(data []byte) error {
+	millis, err := strconv.ParseInt(string(data), 10, 64)
+	if err != nil {
+		return err
+	}
+	*t = Time(time.Unix(0, millis*int64(time.Millisecond)))
```
ncw, line 18: time.Millisecond = 1000000

So this will overflow at 2^63 / 1000000 = 9223372036854 milliseconds from the epoch which turns out to be 2262 so we are OK here for 200 years!

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/rclone/rclone/lib/encoder"
+	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/random"
+	"github.com/rclone/rclone/lib/readers"
+
```
ncw, line 23: There is probably an extra space here which has upset the sorting of stuff!

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+		NewFs:       NewFs,
+		Config: func(name string, m configmap.Mapper) {
+			opt := oauthutil.Options{
+				// No refresh token unless ApprovalForce is set
+				OAuth2Opts: []oauth2.AuthCodeOption{oauth2.ApprovalForce},
+			}
+			if err := oauthutil.Config("zoho", name, m, oauthConfig, &opt); err != nil {
+				log.Fatalf("Failed to configure token: %v", err)
+			}
+			// We need to rewrite the token type to because Zoho wants
```
ncw, line 84: Ah, the wonderful thing about standards! Nearly every backend that uses oauth has some oauth related bodge.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+
+// CreateDir makes a directory with pathID as parent and name leaf
+func (f *Fs) CreateDir(ctx context.Context, pathID, leaf string) (newID string, err error) {
+	//fs.Debugf(f, "CreateDir(%q, %q)\n", pathID, leaf)
+	var resp *http.Response
+	var info *api.ItemInfo
+	opts := rest.Opts{
+		Method:       "POST",
+		Path:         "/files",
+		ExtraHeaders: map[string]string{"Accept": "application/vnd.api+json"},
```
ncw, line 515: If you have this header for all calls then you can add it into the rest.Client just once with `f.srv.AddHeader(...)`

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+	directory, leaf = dircache.SplitPath(remote)
+	if f.root != "" {
+		// Adds the root folder to the path to get a full path
+		directory = path.Join(f.root, directory)
+	}
+	return
+}
+
+// readMetaDataForPath reads the metadata from the path
+func (f *Fs) readMetaDataForPath(ctx context.Context, path string) (info *api.Item, err error) {
```
ncw, line 287: Is there a way in the API to turn a dir/subdir/file path into an ID? If so then here would be the place to implement it otherwise rclone will have to traverse the directories to find the path.

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+	opts := rest.Opts{
+		Method:       "PATCH",
+		Path:         "/files",
+		ExtraHeaders: map[string]string{"Accept": "application/vnd.api+json"},
+	}
+	delete := api.WriteMultiMetadataRequest{
+		Meta: []api.WriteMetadata{
+			{
+				Attributes: api.WriteAttributes{
+					Status: "51",
```
ncw, line 710: 51?

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+	}
+	var start, end int64 = 0, o.size
+	partialContent := false
+	for _, option := range options {
+		switch x := option.(type) {
+		case *fs.SeekOption:
+			start = x.Offset
+			partialContent = true
+		case *fs.RangeOption:
+			if x.Start >= 0 {
```
ncw, line 1130: I think this is the same logic as [fs.FixRangeOption](https://github.com/rclone/rclone/blob/e00bf3d723c636633474629968046342e599e15e/fs/options.go#L141) so I suggest calling that

backend/zoho/zoho.go
```diff
@@ -0,0 +1,1240 @@
...
+	err = o.fs.pacer.Call(func() (bool, error) {
+		resp, err = o.fs.srv.Call(ctx, &opts)
+		return shouldRetry(resp, err)
+	})
+	if err != nil {
+		return nil, err
+	}
+	if partialContent && resp.StatusCode == 200 {
+		if start > 0 {
+			// We need to read and discard the beginning of the data...
```
ncw, line 1161: Does the server do this a lot? Ie return the whole thing instead of the range we asked for?

ncw (commented):
That code looks really nice - thank you :-)

I put some discussion points inline.

Next job for me - get the backend running here and prepare the integration test server.

backend/zoho/zoho.go
buengese: It's weirder than that cyrillic and chinese are supported.

backend/zoho/api/types.go
buengese: That's plenty time.

backend/zoho/zoho.go
buengese: It's not used for upload and download (and these won't work with that header) so I opted to only add it for the calls that really need it.

backend/zoho/zoho.go
buengese: There isn't and it really sucks.

backend/zoho/zoho.go
buengese: Status 51 is deleted. I guess i could make this a named constant but's it's the only status used and it's only used in this one place so I didn't bother.

backend/zoho/zoho.go
buengese: Well, it seems it does so 100% of the time. The docs don't say anything about this so I'm kinda hoping this is a bug/side effect of the API not being completely final yet. Right now the work around is used for every single range request. This is why we definetly want an CantRange/InefficientRange flag for this backend.

backend/zoho/zoho.go
ncw: OK that is fine.

backend/zoho/zoho.go
ncw: It is a common problem - just thought I'd check.

backend/zoho/zoho.go
ncw: It just seemed such a random number! Make a const if you want but if it is just used here then a comment would be fine.

backend/zoho/zoho.go
ncw: I made an issue about this: https://github.com/rclone/rclone/issues/4733

backend/zoho/zoho.go
ncw: :-(

ncw:
I had another go at making this backend work...

I made my own client ID on

```
Storage> zoho
** See help for zoho backend at: https://rclone.org/zoho/ **

OAuth Client Id
Leave blank normally.
client_id> 1000.N3XXX
OAuth Client Secret
Leave blank normally.
client_secret> 1a3fd1XXX
Edit advanced config? (y/n)
y) Yes
n) No (default)
y/n> n
Remote config
Make sure your Redirect URL is set to "https://oauth.rclone.org/" in your custom config.
Use auto config?
 * Say Y if not sure
 * Say N if you are working on a remote or headless machine
y) Yes (default)
n) No
y/n> y
If your browser doesn't open automatically go to the following link: http://127.0.0.1:53682/auth?state=8JYOFLxVgQbKJGJr_uMxow
Log in and authorize rclone for access
Waiting for code...
Got code
2020/11/03 16:22:26 Failed to configure token: failed to get token: oauth2: server response missing access_token
```

If I use `rclone config -vv --dump bodies` I can see the response is

```
2020/11/03 16:26:16 DEBUG : HTTP RESPONSE (req 0xc0005a6000)
2020/11/03 16:26:16 DEBUG : HTTP/1.1 200 
Content-Length: 26
Connection: keep-alive
Content-Language: en-US
Content-Type: application/json;charset=UTF-8
Date: Tue, 03 Nov 2020 16:26:16 GMT
Server: ZGS
Set-Cookie: b266a5bf57=dcb92d0f99dd7421201f8dc746d54606; Path=/
Set-Cookie: iamcsr=8b62800d-bb2e-4bc3-8409-7400e9572d88;path=/;SameSite=None;Secure;priority=high
Set-Cookie: _zcsr_tmp=8b62800d-bb2e-4bc3-8409-7400e9572d88;path=/;SameSite=Strict;Secure;priority=high
Strict-Transport-Security: max-age=63072000
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1

{"error":"invalid_client"}
```

:-( Any ideas?

Also this is using `https://oauth.rclone.org/` as a redirect URL. Does the `http://localhost` or `http://127.0.0.1` one not work? The `oauth.rclone.org` one is undesirable as it redirects the credentials through an rclone server which I'd prefer to avoid if possible.

ivandeex (approved):
My rant wrt body close has been solved.

buengese:
@ncw That's annoying!. I've also tried a trial account now. Using the local redirect works. Looks like we have to use different endpoints for different regions (currently I hardcoded it .com (global)). I thought that would work for eu accounts to but it doesn't. 
 I had to register rclone as a server application for oauth to work.

buengese:
I added a config option for the region. @ncw you can now try this again.

ncw:
Sorry I dropped the ball on this @buengese 

I'd like to merge this for 1.54 if possible.

Other than me getting it to work on the integration test server, I think it is all done - is that what you think too?

buengese:
As far as my memory goes yeah.

ncw:
I've

- rebased on master and fixed the conflicts
- fixed up the bitrot (`ctx` was added to `NewFs` etc)
- added a few missing docs (eg menu entry)
- tested the backend again - it passed :-)
- squashed the commits

I'll push it up now and if it all goes green then I'll merge :-)

ncw:
Looking good - merging now :-)

Thank you very much for this!

Issue #4673, opened 2020-10-12
PROTOCOL_ERROR on S3


armsby:
#### What is the problem you are having with rclone?
Then using rclone mount I have been seeing errors like this: Dir.Stat error: SerializationError: failed to decode REST XML response
status code: 200, request id: 163C63E6DEB02A2C
caused by: stream error: stream ID 19137; PROTOCOL_ERROR

this goes away only with a restart of the rclone mount,
after a small talk with darthShadow, it was suspected that it was http2 that is causing the issue, 
after running some test where http2 was disabled in rclone the issue has not appeared.

a flag to disable http2 for s3 would be nice

#### What is your rclone version (output from `rclone version`)
rclone v1.53.1

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Ubuntu

####  Which cloud storage system are you using? (eg Google Drive)
S3

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
stress an S3 mount and you will see it

ncw:
Are you using this with AWS S3?

> stress an S3 mount and you will see it

Do you have a recipe for that?

This is almost certainly a bug in one of
- go http2 library
- aws s3 sdk
- s3 servers

It would be nice to get to the bottom of which...

Can you dump the response? You'd do that with `-vv --dump responses`. I relalise that may be tricky depending on how you are stressing the backend.

Which go version is your rclone compiled with? `rclone version` will tell you?

@darthShadow did you manage to reproduce this yourself?

darthShadow:
This is with minio proxied via Cloudflare. I have heard reports of it happening directly with minio too, however.

I am guessing this is another bug with the golang http/2 library since it looks very similar to the one we are facing with google drive.

I haven't gotten a chance to reproduce it myself. I will be doing some testing for the same the next month so I will try to debug it further then.

armsby:
that patch that darthShadow provided worked, I will try to run a rclone without it, both over cloudflare and without, and I hope that I can catch it again

ncw:
I've merged @darthShadow 's fix for this this to master now which means it will be in [the latest beta](https://beta.rclone.org) in 15-30 mins and released in v1.54

PR #4674, opened 2020-10-12
s3: Added `--s3-disable-http2` to disable http/2

darthShadow:

Added flag to disable HTTP/2 on S3 backends

#4673 

#### Checklist


backend/s3/s3.go
```diff
@@ -1390,7 +1415,19 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 		SessionToken:    opt.SessionToken,
 	}
 
-	lowTimeoutClient := &http.Client{Timeout: 1 * time.Second} // low timeout to ec2 metadata service
+	defaultClient := getClient(opt)
+	defaultClient.Timeout = 1 * time.Second
+	lowTimeoutClient := defaultClient
+
+	//defaultTransport := new(http.Transport)
```
darthShadow, line 1422: This was the earlier default i.e create a new client instead of using the one given by fshttp. 

Do you want me to revert to that behaviour and uncomment the following code or the new code is fine and I can delete the commented part?

backend/s3/s3.go
```diff
@@ -1381,6 +1393,19 @@ func (o *Object) split() (bucket, bucketPath string) {
 	return o.fs.split(o.remote)
 }
 
+// getClient makes an http client according to the options
+func getClient(opt *Options) *http.Client {
+	// TODO: Do we need cookies too?
```
darthShadow, line 1398: Cookies?

backend/s3/s3.go
```diff
@@ -1381,6 +1393,19 @@ func (o *Object) split() (bucket, bucketPath string) {
...
+	t := fshttp.NewTransportCustom(fs.Config, func(t *http.Transport) {
+		if opt.DisableHTTP2 {
```
ncw, line 1400: Aside: It occurs to me that this would be easy to make a global option... That doesn't disable it per backend but it might be useful for debugging.

backend/s3/s3.go
```diff
@@ -1390,7 +1415,19 @@ func s3Connection(opt *Options) (*s3.S3, *session.Session, error) {
 		SessionToken:    opt.SessionToken,
 	}
 
-	lowTimeoutClient := &http.Client{Timeout: 1 * time.Second} // low timeout to ec2 metadata service
+	defaultClient := getClient(opt)
+	defaultClient.Timeout = 1 * time.Second
```
ncw, line 1419: I think you've got the timeouts the wrong way round for `lowTimeoutClient` and `defaultClient`

backend/s3/s3.go
ncw: I don't think you need to touch the lowTimeoutClient - it is only used for communicating with the local host server to get credentials - it won't use HTTP/2 for that so we don't need to disable it.

ncw (commented):
Thanks for doing this - see inline comments :-)

backend/s3/s3.go
darthShadow: @ncw Do we need the cookies too btw? I noticed that even for Google drive the cookie jar is not set so am not sure whether we need that or not?

backend/s3/s3.go
darthShadow: Yep, had the same thought. I also wanted to add experimental support for quic (using quic-go) in the fshttp client so will probably do both of them together when I get around to it.

backend/s3/s3.go
ncw: I don't think we need cookies.

backend/s3/s3.go
ncw: Great! That would be very useful.

ncw (approved):
This looks great now! I'll merge it :-)

Thank you

Issue #4298, opened 2020-06-02
rclone deleted everything on my computer


islandcmd:
#### What is the problem you are having with rclone?

rclone deleted everything on my computer

#### What is your rclone version (output from `rclone version`)

the latest version

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

ubuntu

####  Which cloud storage system are you using? (eg Google Drive)

google drive

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone sync GoogleDrive:drive ~

calisro:
Sorry to hear that.  I hope you have backups of your data and can restore.

The sync command tells rclone to make the destination (~) look like the source (GoogleDrive:drive).  On Linux the '~' means the home directory of the user.

https://rclone.org/commands/rclone_sync/#synopsis

animosity22:
If we can do anything to help out, please post on the forums and we are happy to try to assist as this really isn't an issue.

https://forum.rclone.org

ncw:
I'm sorry to hear that :-(

I did have an idea for improving this

"sync" is the dangerous operation.  We can tell if rclone is being run at the command line so maybe we could use those two bits of info somehow...

If "sync" and "command line"
- if source is empty, stop with an error
- if deleted more than X files stop

Any other ideas?

animosity22:
I'd recommend to add any suggestions or comments to the running issue we have going:

https://github.com/rclone/rclone/issues/1574

Issue #1574, opened 2017-08-02
Highly destructive actions require confirmation


lucasgonze:
#1573 See https://forum.rclone.org/t/unexpected-deletions-in-gdrive/3421 for user-level discussion. This issue is about rclone itself.

A new user just getting to know rclone may naturally explore the sync command. A typical such user will try out the arguments, using trial and error until they are just right.

For example, they may do this:
rclone sync to from

And of course that won't work. The order of the arguments is:
rclone sync from to

However, there is no second chance. The first command will have wiped out the "from" directory. 

- The deletion is permanent. There is no undo or trash can. 
- The deletion is recursive. If you are looking to sync from the root of your remote directory, everything is gone.
- The deletion is massive. Thousands or tens of thousands of files can disappear.
- The deletion includes git history, so no restore will be possible.

Suggested fix: 

1. A prompt should ask the user to confirm they want to delete the target permanently 
2. A -f flag should suppress the prompt
3. Consider an undo feature

lucasgonze:
Let me also add a personal note.  Are you actually not aware that extreme data destruction requires at least a trivial amount of caution? How is it physically possible to be that out of it? 

Does rclone create any kind of log, at least? Is there a record of the remote URLs nuked?

9182bfe2d3832daf3774c86a85e5a305:
I can't tell if it's you that had the issue, but https://support.google.com/drive/contact/missing_items might be of help, if you're on gsuite you can even roll back deletions yourself if they're less than 30 days old.

lucasgonze:
To roll back deletions you have to be an enterprise customer. I started up a trial subscription. However the restore feature has not worked. It is possible that if I can get a URL to one of the deleted items it would help. That's why ask I ask about a log.

9182bfe2d3832daf3774c86a85e5a305:
You can't get url logs unless you turn on --dump-headers I believe, rclone doesn't log anything without that I know of being explicitly told to. Are you sure it  wouldn't be in your browser history though?

ncw:
@lucasgonze I'm very sorry to hear about your problems with rclone.

The sync command is dangerous - note the bold **important** [in the docs](https://rclone.org/commands/rclone_sync/)

> **Important:** Since this can cause data loss, test first with the --dry-run flag to see exactly what would be copied and deleted.

`rsync` which `rclone` is modeled on has the same behavior and `rsync` is the reason rclone works like it does.

> Consider an undo feature

[...]

PR #4675, opened 2020-10-12
chunker: allow for no-rename checkpoints on backends with slow rename 

mcalman:

Some storage providers, such as **S3** (edit) or IBM COS, don't have an efficient rename operation. When chunker finishes an upload, the server-side copy and delete operations that are done to rename temporary chunks to their final chunk names takes a very long time (in some cases nearly as long as the upload took).

~~This change replaces temporary chunk names with a "status" control chunk that is uploaded at the beginning of a chunked upload. All operations that require ignoring temporary chunks will check for the existence of this "status" control chunk. When an upload succeeds, the "status" control chunk is removed to signify that it is a complete upload and the chunks are no longer temporary. This removes the need for temporary chunk names ~~entirely~~, and therefore avoids the rename operation that is normally done at the end of a chunked upload.~~

**EDIT**
This PR extends metadata with a random "transaction identifier" striving to remove the need for the rename operation on such backends.

#### Checklist


ncw:
@ivandeex have you got time to review this?

ivandeex:
Thanks for contribution. I have a few comments.

I keep in mind that any remote filesystem can be accessed from several rclone instances, possibly modifying the same file in parallel, possibly from different rclone versions, and rclone must prevent data corruption in all known cases.

1. Every change that introduces or removes metadata fields or introduces new control chunks or changes their semantics/behavior, must increase metadata version number. As soon a file is modified by a recent rclone (chunker), any attempt to work on it from an older rclone can corrupt data and should be rejected due to version mismatch. Currently the error message explicitly asks user to upgrade rclone (in case they upgraded one installation and forgot another). Also, chunker documentation must explicitly state what previous rclone versions are affected.

2. With your patch, when user uploads a new version of existing file, it disappears from directory listing until upload finishes. This looks slightly unnatural. If another copy of rclone tried to work with the same file in parallel, what would give?

3. Did you consider making parallel operations wait on files marked by *status* instead of hiding them?

4. Looking at line https://github.com/rclone/rclone/pull/4675/files#diff-0cc14ded5dbad90b4cd5673dc7716447fde13995a5d4fa504cb9c33701f5d7e2R964 I suspect that removal of old chunks would remove just uploaded chunks with same chunk numbers since they don't have temporary suffix anymore. I may be wrong. Did you test such scenario?

5. I admit that chunker test suite lacks testing of parallel operations, e.g. *start a few long multichunk Update()s or Update()+Move(), then cancel one of them*, etc. Before we have those (I will submit an issue and start working on PR later), could you try something like that manually with your patch? I suspect there be dragons.

6. AFAIK most supported rclone backends have fast/atomic server-side move or fast/atomic server-side copy (at least AWS S3 behaves as if it had copy-on-write and consequently fast copy). If we decide to go on with your approach to backends that don't have it, I think we'd better have a `fs.Feature` flag like `SlowServerSideMove` and activate your codepath conditionally, so we could keep on using parallel safe operations on existing backends.

7. Instead of removing transaction ID suffix from chunk names, we could go the other way and instead of renaming chunks just update the ID in the metadata upon operation completion. This approach avoids renaming but changes notion of what temporary chunk is (permanent if suffix is registered with metadata, temporary or garbage otherwise) and needs more thinking.

ivandeex:
Also, worker pools could be used to speed up parallelize-able bits like chunk renames/deletes:
- https://gobyexample.com/worker-pools
- https://github.com/rclone/rclone/blob/master/backend/union/union.go#L868

mcalman:
@ivandeex I just pushed a new commit trying out your last suggestion of permanently keeping the transaction ID suffix in chunk names. It seems like this should address most of your other concerns as well, but let me know what you think.

leoluan1:
@ivandeex @ncw The impact of the performance penalty and server-side overhead of using the copy-and-delete rename operation is a problem broader than IBM COS.  I tried Amazon S3 backend it also incurs copy and delete and it takes a few seconds for each one GB chunk object.  So we should use the “Remote: S3” label instead of "IBM COS"?  Chunker is a great feature for large files, but the S3 server-side copy for large files (e.g., terabytes) can add hours of delay before a Rclone copy operation completes.  Is it possible to allow this PR to be merged for 1.54?   I had a discussion with Max on addressing backward compatibility by adding a simple extension to his fix to allow the new Rclone to read the files created by existing Rclone versions without the fix.  We will also test the current Rclone version's behavior when seeing files created by the new version.  Max said that the code already passes the integration tests.   Please let us know what else we might be able to do now to ready the PR for merging.  Thanks!

backend/chunker/chunker.go
```diff
@@ -2139,9 +2179,10 @@ func (o *Object) ID() string {
 // Meta format `simplejson`
 type metaSimpleJSON struct {
 	// required core fields
-	Version  *int   `json:"ver"`
-	Size     *int64 `json:"size"`    // total size of data chunks
-	ChunkNum *int   `json:"nchunks"` // number of data chunks
+	Version  *int    `json:"ver"`
+	Size     *int64  `json:"size"`    // total size of data chunks
+	ChunkNum *int    `json:"nchunks"` // number of data chunks
+	XactID   *string `json:"xactID"`  // transaction ID for a completed upload
```
ivandeex, line 2185: Please mark as `omitempty` and put after the _optional_ comment.
This will help to keep backwards compatibility.

ivandeex (changes requested):

backend/chunker/chunker.go
```diff
@@ -2154,13 +2195,14 @@ type metaSimpleJSON struct {
 // - if file contents can be mistaken as meta object
 // - if consistent hashing is On but wrapped remote can't provide given hash
 //
-func marshalSimpleJSON(ctx context.Context, size int64, nChunks int, md5, sha1 string) ([]byte, error) {
+func marshalSimpleJSON(ctx context.Context, size int64, nChunks int, md5, sha1, xactID string) ([]byte, error) {
 	version := metadataVersion
```
ivandeex, line 2199: As we know, the difference between version 1 and 2 is the transaction id field.
If xactId=="" here, let's write version=1 and make sure the go marshaler omits it.
Please add unit test that checks ability to read and write metadata ver.1.

backend/chunker/chunker.go
```diff
@@ -1719,6 +1757,7 @@ type Object struct {
 	chunks []fs.Object // active data chunks if file is composite, or wrapped file as a single chunk if meta format is 'none'
 	size   int64       // cached total size of chunks in a composite file or -1 for non-chunked files
 	isFull bool        // true if metadata has been read
+	xactID string      //
```
ivandeex, line 1760: Leftover

backend/chunker/chunker.go
```diff
@@ -728,6 +754,12 @@ func (f *Fs) processEntries(ctx context.Context, origEntries fs.DirEntries, dirP
 				break
 			}
 			object := f.newObject("", entry, nil)
+
+			currentXactID, err = readXactID(ctx, object)
```
ivandeex, line 758: Before this change a directory listing could be finished in one network request on average on most storage providers.
This adds one extra remote file reading (and extra network roundtrip) per a composite file in a directory.
For exanple, given a subdirectory with ~100 composites and network latency ~70ms the directory listing jumps fron 0,1s to 7s.
Dilemma: we improve upload of terabate composites on S3 but slow down directory list on eg GDrive
I don't see a single solution that fits all.

backend/chunker/chunker.go
```diff
@@ -1050,16 +1089,6 @@ func (f *Fs) put(ctx context.Context, in io.Reader, src fs.ObjectInfo, remote st
 	// If previous object was chunked, remove its chunks
 	f.removeOldChunks(ctx, baseRemote)
```
ivandeex, line 1090: You have to pass old transaction id here.
If you don't, this will only remove v1 chunks but gather v2 chunk garbage.
Integration tests won't detect this but users won't thank you.

ivandeex:
@mcalman @leoluan1 

rclone supports 30+ storage backends:
- some with slow uploads
- some with slow server-side copy (because they are dumb and copy data when told to)
- some with instant server-side copy (because they just bind same data to another parh or due to hash-based deduplication under the cover eg _Mailru_)
- some with slow server-side moves (because they select disk based on full name hash, e.g. _Ceph_ in S3 mode)
- some with `Move` (slow or fast) directly in the API
- some without `Move` in API, so rclone will repkace it with `Copy` (slow or fast) + `Delete` (sometimes slow eg. on _Hubic_)
- more exotic cases. since you work with bigdata you migh make up some.

Please run a simple test: create a 10G local file, upload to your storage and compare `rclone move` vs `rclone copy` time for uploaded and copied file.
```
dd if=/dev/zero of=./test1 bs=1m count=10240
rclone copy ./test1 remote:
time rclone copy remote:test1 remote:test2
time rclone move remote:test1 remote:test3
time rclone move remote:test2 remote:test4
```
Please compare results for _IBM COS_ and _S3_ as `remote` and post here. For better accuracy you might repeat your test 2-3 times and take average. It's gonna be quick as you have access already.

Let's make sure we all (including me and @ncw) know what we do and how it affects others.

backend/chunker/chunker.go
ivandeex: I believe this is the small fix you mentioned.

ivandeex:
By the way, this PR does not do avoid renames completely. Currently it only deals with `put` (uploads) but leaves `move` as is. With measurements requested above we can decide what to do about it.

backend/chunker/chunker.go
ivandeex: The reason why I ask will become clear later, when we discuss `List` and `Move`.

backend/chunker/chunker.go
```diff
@@ -1461,6 +1490,7 @@ func (f *Fs) copyOrMove(ctx context.Context, o *Object, remote string, do copyMo
 			break
 		}
 		chunkSuffix := chunkRemote[len(mainRemote):]
+		fs.Debugf(o, "CHUNK SUFFIX: %s", chunkSuffix)
```
ivandeex, line 1493: Debugging leftover. Let's keep it for now but clean em up before merging to master.

backend/chunker/chunker.go
```diff
@@ -1077,7 +1106,7 @@ func (f *Fs) put(ctx context.Context, in io.Reader, src fs.ObjectInfo, remote st
 	switch f.opt.MetaFormat {
 	case "simplejson":
 		c.updateHashes()
-		metadata, err = marshalSimpleJSON(ctx, sizeTotal, len(c.chunks), c.md5, c.sha1)
+		metadata, err = marshalSimpleJSON(ctx, sizeTotal, len(c.chunks), c.md5, c.sha1, xactID)
```
ivandeex, line 1109: `marshalJson` is the commit point of this operation when external actors and parallel operations start seeing changes (here it's the uploaded chunks).
With meta v1 we should have removed old permanent chunks and renamed new temporary chunks by far.
With meta v2 (when transaction id is non-empty string):
- there is no need to rename new chunks (hurray)
- old chunks become garbage automatically when meta is updated, so we can postpone their removal until return and let others see changes earlier

The above is not true if `useMeta==false` or for non-composite objects (aka wrappers).

ivandeex:
> Is it possible to allow this PR to be merged for 1.54?

@ncw has not yet attached a term to that milestone, so it depends. Fairly possible as we have three pair of eyes reviewing this stuff now (with Nick to join later). We can also publish a beta and use it before a release. I used to install releases from my own fork to start eating own dogfood before it hit the release. If an important patch misses 1.5x there always is 1.5x.1 to land. Missing a release is not good but introducing regressions or breaking peoples data is much worse.

backend/chunker/chunker.go
ivandeex: Garbage collection becomes harder (for a human) with new commit method and calls for changes in `Cleanup` and for custom chunker's `About` but it's out of scope for this PR.

leoluan1:
@ivandeex Good idea to get some performance numbers recorded here as a reference!  https://github.com/rclone/rclone/pull/4675#issuecomment-727936051
Here are some quick measurements that I got.  
Using an Amazon S3 bucket, I got the following numbers.
```
Leos-MacBook-Pro:tmp leoluan$ dd if=/dev/zero of=./test1 bs=1m count=10240
10240+0 records in
10240+0 records out
10737418240 bytes transferred in 9.070568 secs (1183764693 bytes/sec)
bash-3.2# rclone copy ./test1 aws:/firebird-leo/

bash-3.2# time rclone copy aws:/firebird-leo/test1 aws:/firebird-leo/test2

real	3m55.478s
user	0m0.129s
sys	0m0.075s
bash-3.2# time rclone move aws:/firebird-leo/test1 aws:/firebird-leo/test3

real	3m15.304s
user	0m0.134s
sys	0m0.066s

bash-3.2# time rclone move aws:/firebird-leo/test2 aws:/firebird-leo/test4

real	1m56.005s
user	0m0.113s
sys	0m0.054s
bash-3.2# 
```
Using an IBM COS bucket, I got the following numbers.
```
root@tapsrv01:/tmp# dd if=/dev/zero of=./test1 bs=1M count=10240
10240+0 records in
10240+0 records out
10737418240 bytes (11 GB, 10 GiB) copied, 8.05723 s, 1.3 GB/s
root@tapsrv01:/tmp# rclone --no-check-certificate copy ./test1 tucson:leo/
root@tapsrv01:/tmp# time rclone --no-check-certificate copy tucson:leo/test1 tucson:leo/test2

real    1m31.070s
user    0m0.608s
sys     0m0.164s
root@tapsrv01:/tmp# time rclone --no-check-certificate move tucson:leo/test1 tucson:leo/test3

real    1m31.824s
user    0m0.597s
sys     0m0.132s
root@tapsrv01:/tmp# time rclone --no-check-certificate move tucson:leo/test2 tucson:leo/test4

real    1m32.685s
user    0m0.651s
sys     0m0.084s
```
The AWS S3 move operation had a large fluctuation between the first and second moves.  But it's consistent that they all took significant amount of time to complete.

ivandeex:
Looks like both _S3_ and _IBM COS_ are `[NOAPIMOVE] [MOVEISCOPY] [DUMBCOPY] [FASTDELETE]`
(see legend in updated https://github.com/rclone/rclone/pull/4675#issuecomment-728376880).
Time variations can be attributed to the provider ops scheduler selecting different disk groups for each other operation.
Conclusions:
1. You are right, chunker should avoid `move` completely.
2. Current implementation of `Move` in chunker is transaction-unsafe on such providers (see [copyOrMove](https://github.com/rclone/rclone/blob/v1.53.2/backend/chunker/chunker.go#L1464)). Moving chunks is on average as slow as copy and by the time it "commits" (ie moves metadata) the concurrent operations will hit the composite in an inconsistent state with a good probability. The solution is to
  rewrite the sequence `move chunks, move metraobject`
  as `get new xactID, copy chunks, commit metaobject, delete old chunks`.
  For "normal" providers (slow copy, quasi-instant move) the _former_ sequence can stay as is (_quasi_-safe but reasonably fast).
  I'm not sure though if this change belongs with this PR or can be postponed to another PR. Let @ncw decide.

backend/chunker/chunker.go
leoluan1: This is a good point.   It does seem that we should mitigate the problem by keeping the rename mechanism for the backends with fast rename and use this PR's approach for the other backends (Amazon S3, IBM COS, Hubic).  

The other thinking is that, when chunker is used, it is more likely that the directory will contain a relatively smaller number of large files.  So the impact of the additional object reads may be smaller.  Hopefully, some reminder/warning on the tradeoff made between slow upload and slow large-directory-listing in the S3 backend documentation can help.

In the longer term, maybe a per-directory meta object can be considered?

backend/chunker/chunker.go
ivandeex: More precisely, the very commit point is here: https://github.com/rclone/rclone/blob/v1.53.2/backend/chunker/chunker.go#L1084

backend/chunker/chunker.go
ivandeex: Currently `removeOldChunks` doesn't accept transaction id because it rereads old metadata (if any) and rescans directory for old chunks.

backend/chunker/chunker.go
ivandeex: > we can postpone their removal until return and let others see changes earlier

Correcting myself.
For this optimization the `removeOldChunks()` will have to know the old transaction id (pre commit point).

backend/chunker/chunker.go
ivandeex: In the long term chunker is only one of metadata consumers in rclone. Others are: new _compress_ layer (to seek over transparently compressed files), _crypt_ layer (to support checksums) and many backends (to simulate lacking modification time, permissions etc). Refer to https://github.com/rclone/rclone/issues?q=label%3Ametadata for use cases and https://github.com/rclone/rclone/issues?q=author%3Ayonjah+is%3Aopen for requirements. Implementing `metadir` will help only one use case but add another format to convert and another set of test cases to write and maintain. I would rather wait for an experimental Metadata interface in rclone. Probably it will be a per-remote key-value store with quasi-atomic access, locking and caching.

backend/chunker/chunker.go
ivandeex: @mcalman @leoluan1 Feel free to skip the `removeOldChunks` optimization, it's optional for this PR.

mcalman:
@ivandeex The latest commit should address most of the points you brought up. Mainly, I made changes to ensure that chunker is backwards compatible, to the extent possible, and made it so the new codepath is only used conditionally (to avoid the inefficiencies you pointed out with reading the metadata file for backends that won’t benefit from the changes).

From my testing, the new version of chunker can correctly identify and deal with files that were written using the original version of chunker. The only scenario I see any problems is when the original version of chunker is interacting with version 2 files (files that were written by the new version of chunker using the new style). In this scenario, the permanent chunks are ignored and the metadata file gets treated like a small, unchunked file.

Currently, I am just checking if the wrapped file system has a `Move` operation in order to decide whether or not to use the new codepath. I wanted to hear your input on what the most robust solution might be. I know you mentioned previously the idea of adding an `fs.Feature` flag for `SlowServerSideMove`. If we are worried about older versions of rclone incorrectly handling new chunked files, then another thought would be to create a new backend flag and only use the new codepath when it is set.

ivandeex:
> I made changes to ensure that chunker is backwards compatible, to the extent possible, and made it so the new codepath is only used conditionally

Great. We are getting there.

> From my testing, the new version of chunker can correctly identify and deal with files that were written using the original version of chunker.

I can't find a unit test for that in your patch. I prefer extensive unit tests activated on every merge as they guarantee against regressions and make changes more safe.

> The only scenario I see any problems is when the original version of chunker is interacting with version 2 files

That' expected

> In this scenario, the permanent chunks are ignored

rclone before 1.54 should treat them as temporary or garbage chunks, that's fine.

> and the metadata file gets treated like a small, unchunked file.

Hmm. I would expect that "List" will list it (as it doesn't read metadata details for performance sake) but any operation that needs NewObject, like `cat`, `hash`, `copy` or `move` _must_ fail with "please upgrade rclone". I'd start to worry if it doesn't fail.

> Currently, I am just checking if the wrapped file system has a `Move` operation in order to decide whether or not to use the new codepath.

I don't see a better way a.t.m.

> I know you mentioned previously the idea of adding an `fs.Feature` flag for `SlowServerSideMove`.

@ncw has a big picture of backends. Before we merge we'll ask him about a better way to check it than `Features.Move != nil`

> another thought would be to create a new backend flag and only use the new codepath when it is set

Greatly appreciated. Let's add a parameter `transactions` (CLI option `--chunker-transactions`, or maybe name it `commit_method` or `checkpointing`) with three possible values: `rename` (old-style checkpoints), `norename` (avoid renames), `auto` (apple-styled "do me good" option). Such a multi-select option `f.opt.Transactions string` allows for more flexibility in future.

However the internal code logic is currently binary: either use rename or avoid it, thus internally we'd better use something like `f.PreferRenames bool`. The `auto` value will set it to true/false depending on `baseFs.Features().Move`. Non-auto values let advanced users decide.

I'm not sure what would be the better default. `auto` is more bold. `rename` is more backwards compatible. I'd prefer `rename` for 1.54, let beta-testers set it to `auto` or `norename` and enjoy faster uploads. Then in 1.55 after we test _norename_ uploads on beta testers and implement _norename_ checkpoints for `Move` too, we can safely change default to `auto`
What do you think?

backend/chunker/chunker.go
```diff
@@ -672,6 +672,34 @@ func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (
 	})
 }
 
+// readXactID returns the xactID stored in the passed metadata object
```
ivandeex, line 675: "returns the transaction ID" (comments are for humans)

backend/chunker/chunker.go
```diff
@@ -672,6 +672,34 @@ func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (
...
+func readXactID(ctx context.Context, metaObject *Object) (xactID string, err error) {
+	// Avoid reading metadata for backends that don't use xactID to identify permanent chunks
+	if metaObject.f.CanQuickRename() {
+		return "", err
```
ivandeex, line 679: Why get filesystem flags implicitly from `metaObject`, which is b.t.w. an object of the _underlying_ backend?
Let's make `readXactID` a method on chunker filesystem (it's known in advance in all places where it's called):
`func (f *Fs) readXactID(...) ...` and invoke it like `f.ReadXactID(...)`

backend/chunker/chunker.go
```diff
@@ -2143,8 +2201,9 @@ type metaSimpleJSON struct {
 	Size     *int64 `json:"size"`    // total size of data chunks
 	ChunkNum *int   `json:"nchunks"` // number of data chunks
 	// optional extra fields
-	MD5  string `json:"md5,omitempty"`
-	SHA1 string `json:"sha1,omitempty"`
+	MD5    string `json:"md5,omitempty"`
+	SHA1   string `json:"sha1,omitempty"`
+	XactID string `json:"xactID,omitempty"` // transaction ID for a completed upload
```
ivandeex, line 2206: Please consider using a short json name like "txn" to save 3 bytes and keep metaobject small.

mcalman:
> Hmm. I would expect that "List" will list it (as it doesn't read metadata details for performance sake) but any operation that needs NewObject, like `cat`, `hash`, `copy`, `move` or _anything_ that modifies data _must_ fail with "please upgrade rclone". I'd start to worry if it doesn't fail. Please correct me. If I'm wrong I will develop a bugfix and backport it to 1.53 and 1.52.

I tested `ls`, `cat`, `hashsum`, `copy`, `move`, and `delete`. They are all not failing and have the same issue I described. It looks like `readMetadata` is not called for all of these commands. Even when it is, `readMetadata` is identifying the objects as not being composite files so it returns without actually checking the metadata file contents.

backend/chunker/chunker.go
ivandeex: Note: I think that the _xactID_ contraction I once used for "transaction ID" is rather unusual (might be confused with "exact") and shall be replaced e.g. by _txnID_ by a future code refactoring.

mcalman:
> Much appreciated. A "level of indirection" will help here, I think. Let's add a parameter `transactions` (CLI option `--chunker-transactions`), or maybe name it `commit_method` or `checkpointing`. It will have three possible values: `rename` (old-style checkpoints), `norename` (avoid renames), `auto` (Apple-style "do me good" option). Such a multi-select string option `f.opt.Transactions` will allow for more flexibility in future.

I added a `transactions` flag that works as you suggested. One thing to note is that problems can arise if the user is inconsistent with the value they pass (e.g. if you upload a file with `norename`, but then try to interact with the file using `rename` the data chunks are ignored). One way to mitigate the issue would be to only set `transactions` in the config file. Is there a way to mandate this or would it be sufficient to document this caveat?

mcalman:
> The new flag can also be used to write an internal test like what I suggest above: set flag to true, upload object, set to false, read object, and vice versa.

I added one unit test that sets `f.useRename` to true, uploads an object, sets `f.useRename` to false, and then looks for the file. This should be passing. However, if you try to do the same test in the opposite order it will fail. Since we want to avoid reading the metadata file when using the old codepath there is no way for it to know that the data chunks uploaded with `norename` aren't temporary chunks. This is the same problem that occurs if users are inconsistent when setting the `transactions` flag.

If there is anything else I should add to this unit test, or additional tests I should add, please let me know.

mcalman:
> I'm not sure what would be the better default. `auto` is more bold, `rename` is more backwards compatible. I'd prefer `rename` for 1.54 and let beta-test volunteers set it to `auto`/`norename` to enjoy faster uploads. Then in 1.55 after we beta-test _norename_ uploads and implement _norename_ checkpoints for `Move`, we can safely change default to `auto`
> 
> What do you think?

I think that makes sense. I've set the default to `rename` for now.

ivandeex:
> additional tests I should add, please let me know.

Here is pseudocode for unit test "server-side move"

1. c1 := chunker.NewFs("x:",baseFS,rename)
2. c2 := chunker.NewFs("y:",same_baseFS,norename)
3. c1.chunk_size = c2.chunk_size = 3 bytes
3. testRun.makeTempFile(c1,"x:file1",content="abcdef")
4. operations.Move("x:file1","y:file1")
5. verify that source and destination have equal size,data,etc

My hunch is it's gonna fail because `okForServerSide` will not impose chunk transformation.

backend/chunker/chunker.go
```diff
@@ -405,6 +427,20 @@ func (f *Fs) setHashType(hashType string) error {
...
+	case "rename":
+		f.useRename = true
+	case "norename":
+		f.useRename = false
+	case "auto":
+		f.useRename = f.CanQuickRename()
+	default:
+		return fmt.Errorf("unsupported transaction mode '%s'", transactionMode)
+	}
+	return nil
```
ivandeex, line 441: Tracking transaction-marked chunks is obviously impossible without metadata.
setTransactionMode should refuse `useRename==true` when `metaFormat==none`.
returning `errors.New("incompatible transaction options")` or alike

backend/chunker/chunker.go
```diff
@@ -218,6 +218,23 @@ It has the following fields: ver, size, nchunks, md5, sha1.`,
...
 			},
+		}, {
+			Name:     "transactions",
+			Advanced: true,
+			Default:  "rename",
+			Help:     `Choose how chunker should handle temporary files during transactions.`,
+			Examples: []fs.OptionExample{
+				{
+					Value: "rename",
+					Help:  "Rename temporary files after a successful transaction.",
```
ivandeex, line 229: Tracking transaction-marked chunks is obviously impossible without metadata.
Let's give user a hint.
Also, we are not going to implement `norename` for for Copy/Move in this PR and have unresolved compatibility problems with previous versions of rclone. I suggest to add a warning like `this method is EXPERIMENTAL, don't use on production systems`.

ivandeex: It's better than nothing.
Though manytimes users will ignore warnings and loose data. Enjoy https://github.com/rclone/rclone/issues/1574#issuecomment-624623587 and #4298.

ivandeex:
Please add a `norename` remote in the rclone integration test https://github.com/rclone/rclone/blob/v1.53.3/fstest/test_all/config.yaml#L85
Name it to your taste.
Capture the relevant _rclone.conf_ snippet in a comment to this PR. Nick will add it to the CI box after merging.

backend/chunker/chunker.go
ivandeex: The rclone integration test includes `nometa` case https://github.com/rclone/rclone/blob/v1.53.3/fstest/test_all/config.yaml#L42,  so you'll have to tweak (`t.Skip("...")` conditionally) the new test case https://github.com/rclone/rclone/pull/4675#issuecomment-748709047.

backend/chunker/chunker.go
ivandeex: Please hide this option from command line:
[visibility value](https://github.com/rclone/rclone/blob/v1.53.3/fs/fs.go#L131)
[how to use visibility values](https://github.com/rclone/rclone/blob/v1.53.3/backend/googlecloudstorage/googlecloudstorage.go#L98)
(I should have done this to the _metadata_format_)

ivandeex:
> I added a `transactions` flag that works as you suggested. One thing to note is that problems can arise if the user is inconsistent with the value they pass (e.g. if you upload a file with `norename`, but then try to interact with the file using `rename` the data chunks are ignored). One way to mitigate the issue would be to only set `transactions` in the config file. Is there a way to mandate this or would it be sufficient to document this caveat?

Please mark this option with `Hide: fs.OptionHideCommandLine`.

Things to mention in user visible documentation, at least:
- "changing `transactions` is dangerous and requires explicit migration": - https://github.com/rclone/rclone/blob/master/docs/content/chunker.md#migrations, https://github.com/rclone/rclone/blob/master/docs/content/chunker.md#caveats-and-limitations
- "`txn` identifies version of the file" - https://github.com/rclone/rclone/blob/master/docs/content/chunker.md#simple-json-metadata-format
- "besides sequential number, the chunk name might have a unique file version suffix" (I don't think we should load users with atomic transactions or rocket sciense) - https://github.com/rclone/rclone/blob/master/docs/content/chunker.md#chunk-names

Things to note in the internal documentation, at least:
- extend definition of permanent and temporary chunks given new mode of operation - https://github.com/rclone/rclone/blob/v1.53.3/backend/chunker/chunker.go#L40

ivandeex:
@mcalman 
I am going to rebase your branch on current rclone master.
Please stash your unpushed commits and pull the rebased branch from github.
Then you can re-apply stashed commits and proceed with work.

backend/chunker/chunker_internal_test.go
```diff
@@ -686,6 +734,9 @@ func (f *Fs) InternalTest(t *testing.T) {
 	t.Run("MetadataInput", func(t *testing.T) {
 		testMetadataInput(t, f)
 	})
+	t.Run("testBackwardsCompatibility", func(t *testing.T) {
```
ivandeex, line 737: nitpick: `t.Run("testBackwardsCompatibility", func)` --> `t.Run("BackwardsCompatibility", func)`
(the `test` prefix stays with function name but goes from visible subtest name)

ivandeex:
@ncw 
Do we have in rclone a log for critical changes e.g. breaking compatibility with previous versions, similar to https://github.com/restic/restic/blob/master/CHANGELOG.md ?

ivandeex:
Thank you @mcalman. My preliminary list of todos is empty now. If you have anything else, please commit now. I suspect 1.54 will happen before February.

I hope integration tests pass for the norename+tinychunk remote.
If yes, I'm going to squash commits, fix a couple small moments I noticed (like unused `m` field), fix the "_branch cannot be rebased_" (github doesn't like merging a moving master into pr, you'd better rebase) and wait for #4902 to rebase upon.

We also need a summary of postponed features like norename-style Copy/Move and a verdict from @ncw.

mcalman:
> unused `m` field

I added the `m` field to the chunker `Fs` struct to preserve the `configmap.Mapper` that is passed in to `NewFs`. I saw that some other backends did this for their `Fs` struct ([example](https://github.com/rclone/rclone/blob/50cf97fc729ee305cb89260f2c7ed387cf12bac0/backend/sugarsync/sugarsync.go#L194)) and it was the only way I could figure out how to access the `configmap.Mapper` from within the server-side move unit test in order to pass it to `NewFs`. If there is a better way to do this please let me know.

mcalman:
> Capture the relevant _rclone.conf_ snippet in a comment to this PR. Nick will add it to the CI box after merging.

```
[TestChunkerNoRenameLocal]
type = chunker
remote = /tmp
chunk_size = 3b
transactions = norename
```

ivandeex:
@mcalman The server-side test could have passed if norenaming fs accessed a chunked file created by renaming fs. I suspect that it would fail in reverse order.

backend/chunker/chunker_internal_test.go
```diff
@@ -663,6 +663,93 @@ func testMetadataInput(t *testing.T, f *Fs) {
...
+	ctx := context.Background()
+
+	f1, err := NewFs(ctx, "x:", f.base.String(), f.m)
+	c1 := f1.(*Fs)
+	c1.useRename = true
+	c1.opt.ChunkSize = fs.SizeSuffix(3)
+	assert.NoError(t, err)
+
+	f2, err := NewFs(ctx, "y:", f.base.String(), f.m)
+	c2 := f2.(*Fs)
```
ivandeex, line 730: > I added the `m` field to the chunker `Fs` struct to preserve the `configmap.Mapper` that is passed in to `NewFs`. I saw that some other backends did this for their `Fs` struct ([example](https://github.com/rclone/rclone/blob/50cf97fc729ee305cb89260f2c7ed387cf12bac0/backend/sugarsync/sugarsync.go#L194)) and it was the only way I could figure out how to access the `configmap.Mapper` from within

@ncw 
We need two ad-hoc remotes for a server-side test.
I recall you recently introduced a context attached config.
Can we use that instead of an otherwise unused link to config mapper inside the Fs struct.

mcalman:
> @mcalman The server-side test could have passed if norenaming fs accessed a chunked file created by renaming fs. I suspect that it would fail in reverse order.

That was my thought as well. However, switching the `useRename` values so that the test is done in the reverse order does not cause the test to fail for me. This makes me suspicious that the test may not be performing as intended although I am not sure why.

backend/chunker/chunker.go
ivandeex: @mcalman
"Don't kill him, he's mine"
(c) random villain

ivandeex:
I am resolved the cheesy "merge-to-PR" and rebased the branch on master. Please pull the result from github, don't force-push now.
Your work before the force-push is saved here https://github.com/ivandeex/rclone/tree/mcalman-20210108-top--chunker-avoid-renames

I can see 4 subtests of rclone integration test failing for the "norename" remote. We shouln't request @ncw review a.t.m. I'm going to fix those.

mcalman:
> We also need a summary of postponed features

- Norename-style Copy/Move
- Switch default `transactions` value to `auto` (after sufficient beta-testing is done)
- Enable rename-style transactions to recognize norename-style chunks

ivandeex:
1. I rebased the PR again on the new master top, reran _integration_ tests and fixed most of failures. Details follow.
2. _ServerSideMove_ test failed on S3 backend with error "specified bucket is not valid". As I don't care about S3 specifically, I skipped it for S3. Testing this on local-based chunker remotes is enough
3. _PreventCorruption_ and _MetadataInput_ tests needed transaction ID to construct test specific chunks. I found that you reserved the field in the chunker _Object_ but **didn't fill** it. Fixed.
4. _FutureProof_ test found that _norename_ mode never checks metadata validity at all. Reason: you mask errors by returning uninitialized `err` instead of real error code. I fixed masking, this fixed the future-proof test but revealed many more tests in fact failing.
5. I added more checks in `readXactID`, this fixed some integration tests.
6. I noticed that many tests fail for `NoMetadata` remotes. This means that this PR inadvertently affects old codepaths. I noticed that `useRename` flag being false can mean two unconnected cases: either `norename` or `no-metadata`. I reversed the flag and named `useNoRename`. This allowed me to perfectly isolate new codepaths and fixed _NoMetadata_ test cases.
7. I removed the line in user docs that allowed to migrate to `norename`, until beta-testing shows it's _really_ safe. You can return to that when you make `auto` the default (you can't make that if it's unsafe).
8. There are still (less) tests failing: see https://github.com/ivandeex/rclone/runs/1718027312?check_suite_focus=true, open the `integration test` section, scroll to bottom and behold). I'm out of available time to address them. **Please fix remaining integration tests.** (The page https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#testing describes the procedure)
9. Please compare `readXactID` against `readMetadata`. I placed the former next to the latter one for better visual diff. They are close in their goals but the former is lacking consistency checks, is susceptible to DDoS (it reads a file afull in memory without checking its size), etc. It causes repeated rereading of metadata, while the latter one caches results in the Object structure saving on API calls and provider charges. I think you can improve by reusing readMetadata with some tweaks (readXactID is called in the very early object stage).
10. I'm leaving for @ncw to decide on the fate of postponed action items (https://github.com/rclone/rclone/pull/4675#issuecomment-761604023)

Thanks

backend/chunker/chunker_internal_test.go
ivandeex: @ncw ?

ivandeex: @ncw here is comment by author --> https://github.com/rclone/rclone/pull/4675#issuecomment-756367287

mcalman:
@ivandeex I believe I've addressed the remaining items you outlined. The integration tests are now passing for me against a local norename backend I created. If there is anything else that needs to be done before we are ready to merge please let me know.

mcalman:
@ncw Are there any further changes that you think should be included in this PR? Can this PR be included in a minor release or will we need to wait for 1.55?

Also, although there are mentions of it throughout this thread, I thought it may be helpful to summarize the backwards compatibility of this PR in one spot. First, the default `transactions` option is to use the original `rename` style. When this is left as the default, the code executes identically to how it did before this PR. This ensures that users who may be unaware of this new option are unaffected. There is also a unit test that ensures that new `no-rename` style transactions can still correctly interpret files that were uploaded using the old `rename` style transactions. The only compatibility issue is if a user wishes to switch from `no-rename` to `rename` transactions they must explicitly migrate their files. This note is included in the chunker documentation and the `transactions` option is hidden from the command line for this reason.

ivandeex:
@mcalman

> I believe I've addressed the remaining items you outlined. The integration tests are now passing for me against a local norename backend I created.

Integration tests _fail_ for me against the _norename backend_.

For your reference, here is:
- the branch (recent upstream https://github.com/rclone/rclone/commit/b2b5b7598c5bd91731233c74118a2c1ad7532e1c from 2021-02-04 + patch to run build on github runners + 18 commits cherry-picked from your branch): https://github.com/ivandeex/rclone/tree/ci-chunker-norename
- test command: https://github.com/ivandeex/rclone/blob/ci-chunker-norename/.github/workflows/build.yml#L105
- test_all/config.yaml: https://github.com/ivandeex/rclone/blob/ci-chunker-norename/fstest/test_all/config.yaml#L30
- rclone.conf with secrets redacted: https://gist.github.com/ivandeex/0e15193b48b12300e428cae21a4fb9a5
- test log (summary is at the bottom): https://github.com/ivandeex/rclone/runs/1883492001?check_suite_focus=true

Summary of errors:
```
2021/02/12 00:46:58 SUMMARY
2021/02/12 00:46:58 FAIL: 2 tests failed in 29m44.41897674s
2021/02/12 00:46:58   * ./sync.test -test.v -test.timeout 50m0s -remote TestChunkerChunk3bNoRenameLocal: -verbose -size-limit 6144
2021/02/12 00:46:58     * Failed tests: [TestSyncBackupDirSuffixOnly TestSyncSuffix TestSyncSuffixKeepExtension TestSyncUTFNorm TestSyncImmutable TestSyncIgnoreCase TestMaxTransfer/Hard TestMaxTransfer/Soft TestMaxTransfer/Cautious]
2021/02/12 00:46:58   * ./sync.test -test.v -test.timeout 50m0s -remote TestChunkerChunk3bNoRenameLocal: -verbose -fast-list -size-limit 6144
2021/02/12 00:46:58     * Failed tests: [TestSyncBackupDirSuffixOnly TestSyncSuffix TestSyncSuffixKeepExtension TestSyncUTFNorm TestSyncImmutable TestSyncIgnoreCase TestMaxTransfer/Hard TestMaxTransfer/Soft TestMaxTransfer/Cautious]
```

Confirmed by running the test on a local linux box with **no-rename remote only**: `$ test_all -verbose -timeout 50m -backends chunker -remotes TestChunkerChunk3bNoRenameLocal:`, test log is here: https://gist.github.com/ivandeex/c6804f66e3282c1670f9d841a214f778

Note: all tests pass for _other flavors_ of chunker.

Please have it fixed.
Thanks.

cc @ncw

backend/chunker/chunker.go
```diff
@@ -709,8 +771,14 @@ func (f *Fs) processEntries(ctx context.Context, origEntries fs.DirEntries, dirP
 			}
 			// this is some kind of chunk
 			// metobject should have been created above if present
-			isSpecial := xactID != "" || ctrlType != ""
 			mainObject := byRemote[mainRemote]
+			if f.useNoRename {
+				currentXactID, err = mainObject.readXactID(ctx)
+				if err != nil {
```
ivandeex, line 777: This will re-read metadata object for every item in the directory!
It will make integration tests pass, but the price is too high.

backend/chunker/chunker_internal_test.go
ivandeex: I think, if we use `fs.NewFs` instead of `chunker.NewFs`, we can avoid the need for the extra mapper reference.

backend/chunker/chunker.go
mcalman: Since `readXactID` caches the `xactID` after it is read (and immediately returns it if it was previously cached), it should still be reading only once per `mainObject`. I believe that is what we want no?

ivandeex: Agreed. I was misguided by the use of single variable `currentXactID` in `processEntries`.
I will replace it by a hash now, so it looks cleaner.

ivandeex: Here is the commit: https://github.com/rclone/rclone/pull/4675/commits/1d219095fb5fd9903504f78b6a80e79173d3bb8b
The commit name is outdated after your note.

backend/chunker/chunker_internal_test.go
ivandeex: My commit https://github.com/rclone/rclone/pull/4675/commits/b3b97aa8725576c79c06ddab703e7ea6c8f5775d closes this thread.

ivandeex (approved):
All inline comments addressed

ivandeex:
All commits are squashed in preparation for the next review. The original chain of commits is saved here: https://github.com/ivandeex/rclone/tree/save-chunker-avoid-renames-20210221

ivandeex:
I can confirm that integration suite passes for all tested flavors of chunker now.

ivandeex:
@ncw
Please review this change.
Below are a few comments for your consideration.

`1.` Currently [fs.Features](https://github.com/rclone/rclone/blob/v1.54.0/fs/fs.go#L506) does not have a flag like `SlowServerSideMove`.
This PR detects such backends using the `Features.Move == nil` guess.
Can you propose a better method?

`2.` A note about directory listings:
After this PR the metadata for _every_ composite object in a directory will be read during the listing operation (by the `readXactID` function, once per object). _Additionlly_, it will be re-read again by `readMetadata` when a concrete operation is requested (Open, Move, Hash, etc). Probably, this can be partially optimized during implementation of https://github.com/rclone/rclone/issues/3972#issuecomment-781563777.

`3.` Summary of features yet unimplemented by this PR:
- Avoid server-side moves in `chunker.Move` for the new "norename" transactions
- Use transaction identifiers in `chunker.Copy` for the new "norename" transactions
- Switch default transactions value to `auto` . This is postponed to after sufficient beta-testing of the new feature
- Enable "rename"-style transactions to recognize norename-style chunks

`4.` Note about integration tests:
rclone.conf on the integration test box will need a new test remote after this PR:
```
[TestChunkerChunk3bNoRenameLocal]
type = chunker
remote = /tmp
chunk_size = 3b
transactions = norename
```

ncw (approved):
> Please review this change.

Sorry for the delay in the review.

> I can confirm that integration suite passes for all tested flavors of chunker now.

:-)

> `1.` Currently [fs.Features](https://github.com/rclone/rclone/blob/v1.54.0/fs/fs.go#L506) does not have a flag like `SlowServerSideMove`.
> This PR detects such backends using the `Features.Move == nil` guess (see [chunker.CanQuickRename](https://github.com/rclone/rclone/pull/4675/files#diff-0cc14ded5dbad90b4cd5673dc7716447fde13995a5d4fa504cb9c33701f5d7e2R2533)).
> Can you propose a better method?

Normally we'd use [operations.CanServerSideMove()](https://github.com/rclone/rclone/blob/b029fb591fb97957c00086a4ebe207a416d050c1/fs/operations/operations.go#L626-L635) but if you are trying to specifically avoid that Copy+Delete then `Features.Move == nil` is pretty close.

> `2.` A note about directory listings:
> After this PR the metadata for _every_ composite object in a directory will be read during the listing operation (by the `readXactID` function, once per object). _Additionlly_, it will be re-read again by `readMetadata` when a concrete operation is requested (Open, Move, Hash, etc). Probably, this can be partially optimized during implementation of [#3972 (comment)](https://github.com/rclone/rclone/issues/3972#issuecomment-781563777).

Composite objects are usually large so there hopefully won't be too many of them in a directory so this seems like an acceptable tradeoff.

> UPD: Just noticed that [UpdateNotify](https://github.com/rclone/rclone/pull/4675/commits/6066f71b5e90ac915d38716d17fc7bedeb9b3f80#diff-0cc14ded5dbad90b4cd5673dc7716447fde13995a5d4fa504cb9c33701f5d7e2R1931) will also reread metadata for every change.
> 
> `3.` Summary of features yet unimplemented by this PR:
> 
> * Avoid server-side moves in `chunker.Move` for the new "norename" transactions
> * Use transaction identifiers in `chunker.Copy` for the new "norename" transactions
> * Enable "rename"-style transactions to recognize norename-style chunks
> * Switch default transactions value to `auto` . This is postponed until the above bullets are finished and sufficient _beta-testing_ of the new feature is done (i.e. while it's released but the default is set to `rename`)
> 
> If @ncw decides to merge this PR as is, these requirements should be captured in feature request(s).

OK

> `4.` Note about integration tests:
> rclone.conf on the integration test box will need a new test remote after this PR:
> 
> ```
> [TestChunkerChunk3bNoRenameLocal]
> type = chunker
> remote = /tmp
> chunk_size = 3b
> transactions = norename
> ```

I've put that in in preparation.

What is the backwards compatibility story with this commit?

I've had a look through the code and I'm happy for you to merge this if you are :-)

Thank you

ivandeex:
@ncw 
Thank you for review.
Before it I was in doubt whether this can be merged as is.

> What is the backwards compatibility story with this commit?

After this patch chunker in "rename mode" will not be able to detect "metadata in norename mode" as a chunker object. That's partially why I requested to enable new mode manually only. Nor do I have time now to fix open items left after author.

Generally the more chunker features we merge -> the more format flavors we have -> the more combinations we test.
I have an idea to merge chunker changes on a dedicated branch and flush on master with metaformat versions "squashed" to tangle the complexity.
See https://github.com/rclone/rclone/pull/4547#issuecomment-786593956

@mcalman 
I was in doubt whether this can be merged as is.

I can merge it as soon as you submit tickets for open items.

Issue #4633, opened 2020-09-27
issues uploading files on minio using a policy with depth


armsby:
#### What is the problem you are having with rclone?
when trying to upload a file to a sub folder im getting a 403 when using rclone
the storage system is minio,
it works if have access to write in the root, but if the policy only gives you access to write in a sub folder I get a 403
it works if im using the minio client

#### What is your rclone version (output from `rclone version`)
rclone v1.53.1
- os/arch: linux/amd64
- go version: go1.15

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
debian

####  Which cloud storage system are you using? (eg Google Drive)
own running minio server

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
rclone copy mc test:bucket/folder/folder/test.bin

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
rclone log:
rclone copy -vv mc test:bucket/folder/folder/test.bin
2020/09/27 20:57:13 DEBUG : rclone: Version "v1.53.1" starting with parameters ["rclone" "copy" "-vv" "mc" "test:bucket/folder/folder/test.bin"]
2020/09/27 20:57:13 DEBUG : Creating backend with remote "mc"
2020/09/27 20:57:13 DEBUG : Using config file from "/opt/.config/rclone/rclone.conf"
2020/09/27 20:57:13 DEBUG : fs cache: adding new entry for parent of "mc", "/root"
2020/09/27 20:57:13 DEBUG : Creating backend with remote "test:bucket/folder/folder/test.bin"
2020/09/27 20:57:14 DEBUG : mc: Need to transfer - File not found at Destination
2020/09/27 20:57:14 ERROR : mc: Failed to copy: AccessDenied: Access Denied.
        status code: 403, request id: 1638B8F5F8648B78, host id:
2020/09/27 20:57:14 ERROR : Attempt 1/3 failed with 1 errors and: AccessDenied: Access Denied.
        status code: 403, request id: 1638B8F5F8648B78, host id:
2020/09/27 20:57:14 DEBUG : mc: Need to transfer - File not found at Destination
2020/09/27 20:57:15 ERROR : mc: Failed to copy: AccessDenied: Access Denied.
        status code: 403, request id: 1638B8F62B5C12EC, host id:
2020/09/27 20:57:15 ERROR : Attempt 2/3 failed with 1 errors and: AccessDenied: Access Denied.
[...]

Issue #4589, opened 2020-09-14
fails to create files in AWS S3 in version >=1.53.0


jti989:
#### What is the problem you are having with rclone?
Rclone gives access denied when creating a file via touch or sync.

#### What is your rclone version (output from `rclone version`)
tested with 

rclone v1.53.0
rclone v1.53.1
rclone v1.54.0-beta.4770.a73c78545

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Ubuntu 18.04

####  Which cloud storage system are you using? (eg Google Drive)
Amazon S3

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
./rclone touch ops2:xsuite-ops2/touch.2 -vv

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
2020/09/14 14:13:40 DEBUG : rclone: Version "v1.54.0-beta.4770.a73c78545" starting with parameters ["./rclone" "touch" "ops2:xsuite-ops2/touch.2" "-vv"]
2020/09/14 14:13:40 DEBUG : Creating backend with remote "ops2:xsuite-ops2/"
2020/09/14 14:13:40 DEBUG : Using config file from "/home/xxx/.config/rclone/rclone.conf"
2020/09/14 14:13:40 DEBUG : fs cache: renaming cache item "ops2:xsuite-ops2/" to be canonical "ops2:xsuite-ops2"
2020/09/14 14:13:40 ERROR : Attempt 1/3 failed with 1 errors and: AccessDenied: Access Denied
        status code: 403, request id: C2163B1XXXXXX9AB, host id: mAI8lg7RB/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxAcUIP3VKQc6FJClT4Lu4txJLSsPKgWUoAcg=
2020/09/14 14:13:40 ERROR : Attempt 2/3 failed with 1 errors and: AccessDenied: Access Denied
        status code: 403, request id: DN6W6XXXXXXYFXEP, host id: 864dfdPB4Xexxxxxxxxxxxxxxxxxxxxxxxxxxxxxx3MbJI2cOyyQXIboxkNkgpTrbd8UENqvZlM=
2020/09/14 14:13:40 ERROR : Attempt 3/3 failed with 1 errors and: AccessDenied: Access Denied
[...]

PR #4680, opened 2020-10-14
mkdir: ignore --s3-no-check-bucket - fixes #4665

### LINKED ISSUE #4665: mkdir is a no-op with --s3-no-check-bucket

The new `--s3-no-check-bucket` option from https://github.com/rclone/rclone/issues/4449 **causes `mkdir` to be a no-op**.
Is that intended behavior? 

I know it doesn't make sense to use that option with mkdir but I'm trying to workaround new behavior from rclone, where `rclone copy hello.txt caringo:mybucket` does a `PUT /mybucket` even if the bucket already exists. Yes that should be okay, AWS S3 returns a 200 OK for a PUT of a bucket that already exists. But the S3-compatible system I'm using is not quite compatible here and returns a 409 error :-(. 

Anyway I'll file a bug with the S3-compatible server about this incompatibility and change the test script I'm using to allow specifying different options for the `mkdir` and `copy` operations. 
**But it seems `mkdir` should create the bucket regardless of `--s3-no-check-bucket`.**

```
bash-4.2# rclone-current --version
rclone v1.53.1
- os/arch: linux/amd64
- go version: go1.15
```

centOS 7.7

S3-compatible

`rclone --s3-no-check-bucket mkdir caringo:s3test-rclone-current-https-r721y1gn`
 

```
$ rclone-current -vv --dump headers --s3-no-check-bucket mkdir caringo:s3test-rclone-current-https-r721y1gn
2020/10/09 00:20:55 DEBUG : rclone: Version "v1.53.1" starting with parameters ["rclone-current" "-vv" "--dump" "headers" "--s3-no-check-bucket" "mkdir" "caringo:s3test-rclone-current-https-r721y1gn"]
2020/10/09 00:20:55 DEBUG : Creating backend with remote "caringo:s3test-rclone-current-https-r721y1gn"
2020/10/09 00:20:55 DEBUG : Using config file from "/root/.rclone.conf"
2020/10/09 00:20:55 DEBUG : You have specified to dump information. Please be noted that the Accept-Encoding as shown may not be correct in the request and the response may not show Content-Encoding if the go standard libraries auto gzip encoding was in effect. In this case the body of the request will be gunzipped before showing it.
2020/10/09 00:20:55 DEBUG : S3 bucket s3test-rclone-current-https-r721y1gn: Making directory
2020/10/09 00:20:55 DEBUG : 2 go routines active
```

jamshid:

ignore --s3-no-check-bucket on mkdir

https://github.com/rclone/rclone/issues/4665#issuecomment-706289251

#### Checklist

- [X ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [N/A?] I have added tests for all changes in this PR if appropriate.
- [N/A?] I have added documentation for the changes if appropriate.

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2069,6 +2069,7 @@ func (f *Fs) bucketExists(ctx context.Context, bucket string) (bool, error) {
 // Mkdir creates the bucket if it doesn't exist
 func (f *Fs) Mkdir(ctx context.Context, dir string) error {
 	bucket, _ := f.split(dir)
+	f.opt.NoCheckBucket = false
 	return f.makeBucket(ctx, bucket)
 }
 


darthShadow:
@ncw How is this supposed to behave when using a user with limited permissions like #4589 & #4633? This change seems to revert the exact thing that solved those issues.

jamshid:
Ah I didn't realize mkdir() was called for other commands. Agreed it's good to rely only on minimal permissions. Let me know if you have any ideas but I'll see if I can rework a fix.

ncw:
> Ah I didn't realize mkdir() was called for other commands. Agreed it's good to rely only on minimal permissions. Let me know if you have any ideas but I'll see if I can rework a fix.

We could possibly put a marker in the context to say this is an intentional mkdir rather than one we want to ignore...

I'm going to close this PR as it is going in the wrong direction, but hopefully we can come up with a new direction :-)

PR #4709, opened 2020-10-26
ncdu: add sort by average size in directory

### LINKED ISSUE #4699: ncdu: add sort by average size in a directory
I love rclone ncdu, but you know what would be a cool new sort mode? size divided by count.

That's basically it. That would then let you sort by average filesize within a folder. I keep finding myself wanting to find where I can spend the least amount of my human time looking at something to gain the largest amount of disk space. I used to think sorting by size was good enough, then while I was sorted by size I found a folder with 2000 files in it, next to a folder with 10 files in it. Both of them essentially right next to each other on any du scan when sorted by size. Obviously it's a lot easier to consider freeing up 100gb if it's 10 files than if it's 2000 files.

adamplansky:

Add keyboard shortcut 'A' for ncdu which sort by average size in directory.

Fixes #4699 

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -533,10 +533,14 @@ func (ds *ncduSort) Less(i, j int) bool {
 		if iAvgSize != jAvgSize {
 			return iAvgSize < jAvgSize
 		}
+		// if avgSize is equal, sort by size
+		return isize < jsize
 	case ds.u.sortByAverageSize > 0:
 		if iAvgSize != jAvgSize {
 			return iAvgSize > jAvgSize
 		}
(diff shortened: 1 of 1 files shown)


ncw (commented):
That is looking nice :-)

I think it is sorting files alphabetically though as it is treating all the sizes as 0 isn't it? They should be sorted as per their size I think.

Playing with this makes me think that we could probably do with an average size column too - but that can be a different PR if we want to do that.

Thanks

adamplansky:
I added sort by size if avgSize is equal (or 0).

ncw (approved):
That looks great now - I'll merge it - thank you :-)

PR #4710, opened 2020-10-26
mount: docs: add note that allow-root and allow-other is not relevant on windows

albertony:

Inform about mount command options not in use on Windows.

According to the following comment from the WinFsp author the `--allow-other` is useless on Windows:

> The allow_other option is actually ignored on WinFsp, because the concept of root and other users makes little sense on Windows.

(https://github.com/billziss-gh/winfsp/issues/138#issuecomment-407212400)

It was also discussed and tested with rclone in this forum thread:

https://forum.rclone.org/t/what-does-allow-other-do/12366

As indicated above the situation is same with `--allow-root`, although could not find any 100% definitive statements confirming this..

#### Checklist


ncw:
That looks great thanks - I'll merge that now.

Issue #4658, opened 2020-10-06
change the text of a log entry when using --suffux


asdffdsa6131:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.53.1
- os/arch: windows/amd64
- go version: go1.15

#### What problem are you are trying to solve?
as per https://forum.rclone.org/t/suffix-not-working-on-folder-upload-via-ssh-sftp/19526/9
change the text of a log entry
#### How do you think rclone should be changed to solve that?
change log entry from 
`INFO  : hidden.txt: Moved (server side)`
to
`INFO  :hidden.txt: Moved (server side) to "path/whatever"`
and as quoted from ncw

"I'd probably do

hidden.txt: Moved (server side) to "path/whatever"

where the to ... but was only added if the "remote" was different which it won't be in most cases.

I could do the same thing to Copy too."

PR #4712, opened 2020-10-26
operations: move verbosly log name of the destination object

adamplansky:

If the object is moved rclone prints name of the destination object into the info log in verbose mode .

fixes: #4658 
#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -501,8 +501,11 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 			return newDst, err
 		}
 	}
-
-	fs.Infof(src, actionTaken)
+	if src.String() != newDst.String() {
+		fs.Infof(src, "%s to: %s", actionTaken, newDst.String())
+	} else {
+		fs.Infof(src, actionTaken)
+	}
 	return newDst, err
 }
 
@@ -556,7 +559,12 @@ func Move(ctx context.Context, fdst fs.Fs, dst fs.Object, remote string, src fs.
 		newDst, err = doMove(ctx, src, remote)
 		switch err {
 		case nil:
-			fs.Infof(src, "Moved (server-side)")
+			if src.String() != newDst.String() {
+				fs.Infof(src, "Moved (server-side) to: %s", newDst.String())
+			} else {
+				fs.Infof(src, "Moved (server-side)")
+			}
+
 			return newDst, nil
 		case fs.ErrorCantMove:
 			fs.Debugf(src, "Can't move, switching to copy")


darthShadow:
Can you please add the same improvement to the copy log too?

adamplansky:
Yes, no problem. Do you want to use same pattern as I did in move?

Current behaviour:
```
2020/10/27 10:50:05 INFO  : test.txt: Copied (new)
```
New behaviour
```
2020/10/27 10:50:05 INFO  : test.txt: Copied (new) to test.txt
```

If I undestand it correctly if file is copied it has always same name, so what is purpose to have this feature implemented?

darthShadow:
Yes, that should be fine.

> If I undestand it correctly if file is copied it has always same name, so what is purpose to have this feature implemented?

That's not necessary. There is no restriction on copy in that it should have the same name on the destination too.

adamplansky:
So I added also destination name into info log when copy is invoked.

fs/operations/operations.go
```diff
@@ -502,7 +502,7 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 		}
 	}
 
-	fs.Infof(src, actionTaken)
```
ncw, line 505: I think this should only have the extra info if the file name was different.

fs/operations/operations.go
```diff
@@ -502,7 +502,7 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
...
+	fs.Infof(src, fmt.Sprintf("%s %%s", actionTaken), newDst.String())
```
ncw, line 505: You can write this

    fs.Infof(src, "%s to: %s", actionTaken, newDst)

I prefer the leading `to: `

fs/operations/operations.go
```diff
@@ -556,7 +556,7 @@ func Move(ctx context.Context, fdst fs.Fs, dst fs.Object, remote string, src fs.
 		newDst, err = doMove(ctx, src, remote)
 		switch err {
 		case nil:
-			fs.Infof(src, "Moved (server side)")
+			fs.Infof(src, "Moved (server side) to %s", newDst.String())
```
ncw, line 559: This should also only print if the destination string is different as most of the time it will be the same.

ncw (commented):
Thanks for doing this - I left some comments inline :-)

adamplansky:
Thanks for guidence, now the fs.Infof is much more readable.

ncw:
That looks great now - thank you - will merge now.

Users are going to love that change :+1: 

I fixed the conflict.

PR #4723, opened 2020-10-30
docs ftp.md put limitations together

edwardxml:

Changes are:

The topic is mostly about limitations so all of these are grouped together with a section hyperlink near the top of the page. Intention is to avoid potential duplication and make it more straightforward (there is a place and it is essentially just a list so wording doesn't need to be elegant) to add notes about limitations in future.

Minor wording alterations that do not intend to change meaning

Issues:

Would it be better to build wikimedia like page content navigation boxes automatically for all documentation pages? I am mostly thinking of a benefit from a link table in lang pages containing many flag descriptions generated from code. 

I have tinkered in a limited way with adding in page links within the Implicit TLS section. This may or may not be a good idea and is a distinct deviation from established style

My changes break any existing bookmarks linking to the the config file free example because I changed the heading name slightly.

Not on this occasion. Change doesn't intend to change the meaning of the documentation

#### Checklist

- [x ] I'm done, this Pull Request is ready for review :-)


ncw:
> The topic is mostly about limitations so all of these are grouped together with a section hyperlink near the top of the page. Intention is to avoid potential duplication and make it more straightforward (there is a place and it is essentially just a list so wording doesn't need to be elegant) to add notes about limitations in future.

That looks very nice - thank you :-)

> Would it be better to build wikimedia like page content navigation boxes automatically for all documentation pages? I am mostly thinking of a benefit from a link table in lang pages containing many flag descriptions generated from code.

Yes I have thought about this. You can do this automatically with [hugo's TOC support](https://gohugo.io/content-management/toc/) - this looks relatively straight forward to integrate and would really help on the bigger pages.

Do you think this is worth doing?
 
> I have tinkered in a limited way with adding in page links within the Implicit TLS section. This may or may not be a good idea and is a distinct deviation from established style

Links are good! There are some links in the docs, but not many. In general the docs need to be readable without links because not all docs formats support them (eg man page).

> My changes break any existing bookmarks linking to the the config file free example because I changed the heading name slightly.

I think that is OK. You can put a permanent link anchor in with `{ anchor }` - it might be a good idea to use more of these, but I don't think the ones you've changed deserve it.

I'll merge your change now - thank you :-)

edwardxml:
I just had a look at the TOC.

It works but there is a bit of inconsistency in heading tag usage between pages. startLevel: 2 and endLevel: 3 works for most of the command pages reasonably but the backend pages are all formatted to suit 3-4. For instance, the backend specific flags are not shown (and that would be the big win) as they are presently H4 and excluded whilst the backend logos are H2 and their alt text messes up the TOC.

I found some documentation for the Hugo TOC at https://gohugo.io/content-management/toc/. That left me with the impression it would all be OK onee doabl day if the H tags were eventually harmonised. I'm presuming hugo is v 0.60.0 or above because goldmark is referenced in docs/config.json.

The mobile rendering is good, but at the bottom of the page. My personal preference would be for text wrap in the narrow desktop link boxes but things never look great with long headings. I didn't go through the font awesome css definition for a suggestion - would have lost the will to live.

Is 'Contents' necessary in docs/layouts/chrome/menu.html? Or 'Share and Enjoy' or 'Links' come to that. The contents of the cards look pretty self explanatory to me.

Maybe I should raise an issue to harmonise the H tags?

ncw:
> I just had a look at the TOC.

Thanks
 
> It works but there is a bit of inconsistency in heading tag usage between pages. startLevel: 2 and endLevel: 3 works for most of the command pages reasonably but the backend pages are all formatted to suit 3-4. For instance, the backend specific flags are not shown (and that would be the big win) as they are presently H4 and excluded whilst the backend logos are H2 and their alt text messes up the TOC.

Yes the backend logos are a mess!

> I found some documentation for the Hugo TOC at https://gohugo.io/content-management/toc/. That left me with the impression it would all be OK onee doabl day if the H tags were eventually harmonised.

Harmonising the headings would probably be the way to go. 

It looks like the backend docs headings need to go up a level.

> I'm presuming hugo is v 0.60.0 or above because goldmark is referenced in docs/config.json.

Yes, we track a recent version of hugo for this - 0.73 at the moment, but I upgrade it quite often.

> The mobile rendering is good, but at the bottom of the page.

Do you think it should go at the top? The trouble with that is the search engines index the text from the top so the snippet for the page becomes the contents menu which maybe isn't as useful as the text. I could probably use some semantic markup here too like hugo suggests, eg the `aside` tag

> My personal preference would be for text wrap in the narrow desktop link boxes but things never look great with long headings.

I played with it quite a bit, but I couldn't get them to wrap and to keep a clear distinction between the levels of menu. That is probably because of my limited CSS skills though!

> I didn't go through the font awesome css definition for a suggestion - would have lost the will to live.

:-)

> Is 'Contents' necessary in docs/layouts/chrome/menu.html? Or 'Share and Enjoy' or 'Links' come to that. The contents of the cards look pretty self explanatory to me.

I wonder whether on doc pages we should just have the contents on the right hand side and lose the other boxes?

(I don't suppose anyone gets my [Share and Enjoy](https://www.urbandictionary.com/define.php?term=share%20and%20enjoy) joke, but it amuses me :-)

I did try to get the contents in the menu which would probably be possible, but my hugo-fu wasn't strong enough when I tried. Worth trying harder do you think?

> Maybe I should raise an issue to harmonise the H tags?

I think we are going to need that whatever we decide on the TOC!

edwardxml:
```
The trouble with that is the search engines index the text from the top
```
My starting point is that the mobile layout looks fine and I am fairly OK with the position at the bottom particularly because the same css is used for the Share and Enjoy and Links card and they definitely belong there. The extent to which search engines index menus and navigation cards is a complex one...

```
I wonder whether on doc pages we should just have the contents on the right hand side and lose the other boxes?
```
Maybe a separate discussion but I'm quite comfortable with them really.

```
Share and Enjoy 
```
OK, yes I missed that! Whenever I see X and Y slogans my mind wanders to the 'Protect and Survive' Civil Defence campaigns (before my time). Now all you need to do is find H2G references instead of 'Contents' and 'Links'!

```
contents in the menu 
```
No, I think you were right with the card - Drop down context specific menus are so Windows 2.0 (not before my time). It is the list of option links displaying on the face of the web page that would be the big win.

When I get back will raise an H1 to H4 harmonisation proposal issue. This morning I did finally get around to wrestling with filter documentation. Now I realise how little I understand rclone filters and the extent I have been avoiding them.

PR #4740, opened 2020-11-04
check Hashes by Etags if chunk size is equal for s3 to s3 sync

kmlebedev:
#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1217,6 +1217,11 @@ disabled here.  When the issue is solved this flag will be removed.
 See: https://github.com/rclone/rclone/issues/4673, https://github.com/rclone/rclone/issues/3631
 
 `,
+		}, {
+			Name:     "use_multipart_etag_as_md5",
+			Default:  true,
+			Advanced: true,
+			Help:     `if set try checks multipart Etags as Md5`,
 		},
 		}})
 }
@@ -1275,6 +1280,7 @@ type Options struct {
 	MemoryPoolFlushTime   fs.Duration          `config:"memory_pool_flush_time"`
 	MemoryPoolUseMmap     bool                 `config:"memory_pool_use_mmap"`
 	DisableHTTP2          bool                 `config:"disable_http2"`
+	useMultipartEtagAsMd5 bool                 `config:"use_multipart_etag_as_md5"`
 }
 
 // Fs represents a remote s3 server
@@ -2661,7 +2667,14 @@ func (o *Object) Remote() string {
 	return o.remote
 }
 
-var matchMd5 = regexp.MustCompile(`^[0-9a-f]{32}(-[0-9]+)?$`)
+// Remote returns the match md5 regexp
+func (o *Object) matchMd5() *regexp.Regexp {
+	matchMd5Str := "^[0-9a-f]{32}$"
+	if o.fs.opt.useMultipartEtagAsMd5 {
+		matchMd5Str = "^[0-9a-f]{32}(-[0-9]+)?$"
+	}
+	return regexp.MustCompile(matchMd5Str)
+}
 
 // Hash returns the Md5sum of an object returning a lowercase hex string
 func (o *Object) Hash(ctx context.Context, t hash.Type) (string, error) {
@@ -2670,7 +2683,7 @@ func (o *Object) Hash(ctx context.Context, t hash.Type) (string, error) {
 	}
 	hash := strings.Trim(strings.ToLower(o.etag), `"`)
 	// Check the etag is a valid md5sum
-	if !matchMd5.MatchString(hash) {
+	if !o.matchMd5().MatchString(hash) {
 		err := o.readMetaData(ctx)
 		if err != nil {
 			return "", err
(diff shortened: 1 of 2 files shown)


ncw:
I see what you are trying to achieve here.

I don't think I can merge that code as it is as it is very specific, but maybe we can come up with something!

What if we had a new flag (say) `--s3-use-multipart-etag-as-md5` (could do with a better name!)

This could then return the hashes as you have in your patch to the `s3` code.

Without the patch to the `operations` code this would throw errors - that part would need to be controlled by a flag too probably.

kmlebedev:
ok

kmlebedev:
@ncw  I added ```option --s3-use-multipart-etag-as-md5```

backend/s3/s3.go
```diff
@@ -1217,6 +1217,11 @@ disabled here.  When the issue is solved this flag will be removed.
 See: https://github.com/rclone/rclone/issues/4673, https://github.com/rclone/rclone/issues/3631
 
 `,
+		}, {
+			Name:     "use_multipart_etag_as_md5",
+			Default:  true,
+			Advanced: true,
+			Help:     `if set try checks multipart Etags as Md5`,
```
ivandeex, line 1224: Please rephrase as something more clear e.g. `check multipart uploads with MD5 taken from server ETags`

backend/s3/s3.go
```diff
@@ -2661,7 +2667,14 @@ func (o *Object) Remote() string {
 	return o.remote
 }
 
-var matchMd5 = regexp.MustCompile(`^[0-9a-f]{32}$`)
+// Remote returns the match md5 regexp
+func (o *Object) matchMd5() *regexp.Regexp {
+	matchMd5Str := "^[0-9a-f]{32}$"
+	if o.fs.opt.UseMultipartEtagAsMd5 {
+		matchMd5Str = "^[0-9a-f]{32}(-[0-9]+)?$"
+	}
```
ivandeex, line 2675: Please make both regexp's statically compiled outside the function.
Name them e.g. `matchNormalMd5` and `matchMultipartMd5`.
The will return either one depending on the flag without compiling it every time.
Don't needelessly load golang regexp compiler/cache.

fs/operations/operations.go
```diff
@@ -91,6 +92,26 @@ func checkHashes(ctx context.Context, src fs.ObjectInfo, dst fs.Object, ht hash.
 	if dstHash == "" {
 		return true, hash.None, srcHash, dstHash, nil
 	}
+	srcHashes := strings.Split(srcHash, "-")
+	dstHashes := strings.Split(dstHash, "-")
+	if len(srcHashes) == 2 && len(dstHashes) == 2 && srcHashes[1] != dstHashes[1] {
+		fs.Debugf(src, "chunk size differ")
+		return true, hash.None, "", "", nil
```
ivandeex, line 99: This chunk is located in a generic function working with backends of all types and hashsums of all types.
Hashsum here is an **opaque blob of bytes** that can be equal or not.
You can't treat it as having any internal structure at this level.

ivandeex (changes requested):

fs/operations/operations.go
ivandeex: This chunk should be factored into a separate function and activated by a non-backend runtime flag.

ivandeex: This chunk should be factored into a separate function. You will call it from this place if enabled by a new non-backend flag.

kmlebedev: ok

ivandeex: Another solution that frees user from the need to add **two** flags:
Create a generic `Fs` interface, e.g. call it `MultipartHashComparer` with a single function `CompareMultipartHashes` and add it solely to `Object` returned by `S3`. At this point in _operations.go_ you will do something like
```
if mhc, ok := src.(MultipartHashComparer); ok {
  result, err := src.(MultipartHashComparer).CompareMultipartHashes(dst)
 .........
}
```

ivandeex: @kmlebedev @ncw 
Do you think it's worth an effort?

ivandeex: Another solution that frees user from the need to add **two** flags:
Create a generic `Fs` interface, e.g. call it `MultipartHashComparer` with a single function `CompareMultipartHashes` and add it solely to `Object` returned by `S3`. At this point in _operations.go_ you will do something like
```
if mhc, ok := src.(MultipartHashComparer); ok {
  result, err := src.(MultipartHashComparer).CompareMultipartHashes(dst)
 .........
}
```

@ncw @kmlebedev
Do you think it's worth the effort?

ivandeex: Actually I see why it can be **worth** the effort.
You will have to add a test for your new feature anyway!
The test procedure for special-case provider requirements is usually `Take such'n'such remote with such'n'such config and run integration test again it`. Well... An S3 flag fits fine in this scenario, but a non-backend flag doesn't. For generic flags we usually write a new test case in `fs/operatations_test.go`. If such a test requires a specific type of provider, we add a skip condition. What can be required in context of this change??
I hope I'm clear enough.

ivandeex:
@kmlebedev Please address review comments

ivandeex:
@kmlebedev Did you forget about this ticket?

kmlebedev:
> @kmlebedev Did you forget about this ticket?
No, it's just not relevant at the moment,
I've been waiting for a new sync moment

ivandeex:
This pull request was last updated 5 months ago.
Please be aware that after 6 months of no code changes pull requests become eligible for closing.

ivandeex:
closing stale requests

PR #4746, opened 2020-11-06
mount: more user friendly mounting as network drive on windows

### LINKED ISSUE #2678: rclone mount Windows: use next available drive letter
rclone v1.43.1
- os/arch: windows/amd64
- go version: go1.11

The built-in Windows "net use" command can take a wildcard instead of a specific drive letter when mapping a network drive. Windows will pick the next unused drive letter (after C:) for the mount point.

This request is to have the "rclone mount" command have the same functionality.

It could be invoked with an argument: "rclone mount --use-next-drive cfg:"

albertony:

On Windows one can mount in one of two modes: fixed disk drive or network drive. Currently the default is disk drive, and to mount as network drive you must pass argument --VolumePrefix into libfuse/WinFsp via the --fuse-flag option in rclone mount:

```
rclone mount remote: X: --fuse-flag --VolumePrefix=\server\share
```

This syntax is a bit convoluted, especially if the network mode is more or less the preferred method (too little user experience myself, but sounds it from what I read it). My intention is to make it a bit more elegant and user-friendly - make the choice between fixed vs network drive mounting more "native" to rclone, and possibly also making the network mode the default.

See also discussion in #4714.

**Part 1: Network mode option**

The existing `--volname` option, currently used to set custom volume label when mounting using disk method into a drive letter, could be re-used for setting the "share" name, and then just add a default prefix `\server\` to mount as network path `\\server\share`. Now we need a new method for selecting disk or network type mounting: We can introduce a new dedicated rclone option to select mode:
```
rclone mount remote: X: --network-mode
rclone mount remote: X: --network-mode --volname myshare
rclone mount remote: X: --network-mode=false
```

Update a:

Added support for specifying the full volume prefix in the `--volname` option. Using regular Windows UNC path style, with double leading backslash, instead of the WinFsp single backslash style. Also found that a subfolder path is legal (added note below). If specified value is UNC style then network mode is implied, regardless of --network-mode. Examples:

```
rclone mount remote: X: --network-mode --volname \\myserver\myshare
rclone mount remote: X: --network-mode --volname \\myserver\myshare\some\folder
```

**Part 2: UNC path as mountpoint**

An even more elegant approach would be to make rclone accept a Windows UNC style path as the mountpoint argument, such as `rclone mount remote: \\server\share`, and then use that as the "hint" to select network type mounting. One issue then is that from rclone it is not possible to mount as network drive without mapping to drive letter.

From discussion in #4714:
*This is something which is supported by WinFsp, but not by current implementation of rclone and/or cgofuse. You can actually supply an empty string as mountpoint argument, rclone mount remote: `"" --fuse-flag --VolumePrefix=\server\share`, and it works... But it will still map to a drive letter - first free letter counting backwards from Z:, logic from WinFsp method FspFileSystemSetMountPoint (which seems to be always called, while WinFsp sample utilities airfs/memfs skips calling this method when no mountpoint is specified, and then the network share is mounted but not mapped to drive letter). Rclone will show an error, because it waits for the mountpoint to become ready polling os.Stat, and since it is empty it times out and reports `ERROR : mountpoint "" didn't became available on mount - continuing anyway`, but it works as an undocumented trick for now.*

What I did do (for now) is, when UNC path is specified, then automatically find a free drive letter and mount that. As noted above this is just what will be done if an empty mountpoint is sent to cgofuse, but by implementing this in rclone it will know exactly what drive letter is picked.

**Part 3: Fixes #2678**

Implemented syntax for automatically picking drive letter (without having to use the unc path method from above, so auto drive letter and auto volume prefix / share path), based on tip on feature request #2678 "Use next available drive letter on Windows" in comments. Mountpoint "*" or "" (empty string) will be used to trigger this feature.

**Part 4: Cleanup handling of os specific mount options**

Were wondering if os-specific flags... should be just that, 
```
	flags.BoolVarP(flagSet, &Opt.NetworkMode, "network-mode", "", Opt.NetworkMode, "Mount as drive type remote network drive, instead of fixed disk drive.")
}
```
or instead documented as such:
```
flags.BoolVarP(flagSet, &Opt.NetworkMode, "network-mode", "", Opt.NetworkMode, "Mount as drive type remote network drive, instead of fixed disk drive (Windows only).")
```

Found later [this](https://forum.rclone.org/t/cant-delete-empty-directory-because-of-ds-store/13003/2), which indicates it should be available on all os and just documented:
>Confusingly these aren't in the docs as they are macOS only and the docs are built on a linux machine. This is probably an oversight!

Done:
- Moved `--noappledouble` and `--noapplexattr` from Darwin specific section and documented as OSX specific, according to comment [here](https://forum.rclone.org/t/cant-delete-empty-directory-because-of-ds-store/13003/2).
- Ignoring `--daemon-timeout`, `--allow-non-empty`, `--allow-other` and `--allow-root` entirely on windows, since the mount "backend" does not support them anyway, and documented this. Also documented that `--deamon`, `--async-read` and `--write-back-cache` are not supported on windows (these were already ignored).
- Ignoring  `--noappledouble` and `--noapplexattr` on "cmd\mount" since it no longer supports OSX (see [this](https://forum.rclone.org/t/rclone-mount-and-macos-testers-needed/17527) and [this](https://github.com/bazil/fuse/blob/88ebede33592c7147c85956887ac4ece38c241b9/options.go#L80-L88)).
- Documented `--volname` as Windows and OSX specific.

**Examples:**

`rclone mount remote: *`
- Mount using default mode, default volume name, automatic assigning a free drive letter.

`rclone mount remote: \\server\share`
- Mount using specific volume prefix (network share path), implicit --network-mode=true, automatic assigning a free drive letter.

`rclone mount remote: \\server\share --network-mode=false`
- Same as above, option "--network-mode=false" silently ignored due to UNC path (or should we warn user?)

`rclone mount remote: X: --volname share`
- Mounting as drive letter X: in default mode (which should be network, implied `--network-mode=true`?). If network mode the volume name "share" is used after an implicit prefix "\\server\" as the volume prefix (resulting in mounted network share unc path \\server\share).

`rclone mount remote: X: --network-mode --volname \\myserver\myshare`
- Mounting as drive letter X: in network mode, with full volume prefix "\\myserver\myshare", which will be shown as the mounted network share unc path.

`rclone mount remote: X: --network-mode --volname \\myserver\myshare\some\folder`
- Same as above, but mounting as a subfolder path into the network share, which seems perfectly valid.

`rclone mount remote: X: --network-mode=false --volname \\myserver\myshare`
- Mounting as drive letter X: in forced network mode, --network-mode=false is ignored, with full volume prefix "\\myserver\myshare", which will be shown as the mounted network share unc path.

`rclone mount remote: X: --volname share --network-mode=true`
- Explicit network mode (on).

`rclone mount remote: X: --volname share --network-mode=false`
- Explicit disk mode (network mode off).

`rclone mount remote: X: --fuse-flag --VolumePrefix=\server\share`
- Works just as in previous versions

`rclone mount remote: X: --fuse-flag --VolumePrefix=\server\share`
- Works just as in previous versions

`rclone mount jfs_dev_crypt: X: --network-mode --volname \\myserver\myshare --fuse-flag --VolumePrefix=\server\share --fuse-flag --VolumePrefix=\server2\share2`
- The volume prefix is specified three times here, last oe wins (nothing rclone does at the moment, WinFsp handles it this way), so will be mounted as `\\server2\share2`.

**Some additional notes:**

* When mounting with directory path the path must not exist, but the parent must! Added a check for this, so that rclone reports an error (instead of relaying error from WinFsp).
* When mounting with drive letter it must be specified in the form "X:", trying "X:\" (with backslash) leads to error:  `Cannot create WinFsp-FUSE file system: invalid mount point.`. This will now be caught by the parent path error check above.
* When mounting as network drive it is not possible to mount into directory path (just drive letter). Added check for this.
* When mouting as network drive using a share name that already exists fails with error messages "Cannot create WinFsp-FUSE file system: invalid mount point.`. No additional check implemented for this, so error will come from WinFsp.
* When mounting as network drive, the volume prefix must use single backslash path `\server\testing` instead of the standard windows unc format `\\server\share` when sent to cgofuse/winfsp: `--VolumePrefix=\server\testing`. When specifying unc path as mountpoint this is handled automatically, when using --fuse-flag the error will come from WinFsp.
* When mounting as network drive, the volume prefix must at least have the `\server\share` part (which identifies the volume), but can also include subfolder path such as `\server\share\some\folder`.

**Questions and TODOs:**

- Needs manual testing, possible corner cases etc. Perhaps a beta release with volunteer testers. Are there relevant unit tests that should be considered?

- Make mounting as network drive the new default, or keep existing disk mode as default?
  - I do not know specifically which positive and negative effects either approach have, but from what I read I get a feeling that network drive should normally be the preferred mode, at least assuming mounting of a remote/cloud backend which does not offer you "near-disk" speed.
  - One limitation with the network mode is that you cannot mount into a directory, you must mount it to a drive letter.
  - One approach could be to always use disk mode when mounting into directory path, and default to network mode when mounting to drive letter (and all other cases).
  - Another approach, since the UNC mountpoint syntax above (using implicit --network-mode=true) makes network mode fell more "native" to rclone, would be to keep disk mode (`--network-mode=false`) as default when mounting into directory path and drive letter, same as in previous versions, and when mounting to a drive letter one would have to use the flag `--network-mode` or set an UNC path as `--volname` to enable network mode. If one does not need to set specific drive letter using the unc path as "mountpoint" in the command syntax is quite nice.

- Should it be checked for, and warned or removed, explicit --VolumePrefix option, in the highly likely case that user is still including the arguments needed in the previous versions? It seems that the last one will be used, and all other silently ignored; when mount code passes `--VolumePrefix=\\server\\share --VolumePrefix=\\server\\share2` into cgofuse it will mount as "\\server\share2". Any fuse options/flags user specifies on the command line are added last, after any set from rclone code, so user specified ones will "win".

- Continuing from previous point, but more generally: Currently the "-o uid" and "-o gid" options are checked for, and default set only if not set by user, but for "--FileSystemName", "--VolumePrefix" and "-o volname" a default value are set without checking if already set by user. Seems the checking for uid/gid were introduced after a [forum discussion](https://forum.rclone.org/t/issue-with-rclone-mount-and-resilio-sync/14730/28), but there are no conclusion there that this made any difference. From my testing it works perfectly fine to set these multiple times, last one wins. This includes "-o uid" and "-o gid"! I can add to rclone mount `-o gid=11 -o gid=-1` and the resulting mount has "Domain Users" (-1) on it. If i use `-o gid=11 -o gid=-1 -o gid=11` it gets "Authenticated Users" (11) instead. Setting `--FileSystemName=xxx` means it gets filesystem "FUSE-xxx" and not "FUSE-rclone" even though rclone always sets "--FileSystemName=rclone". Edit: Confirmed by https://github.com/billziss-gh/winfsp/issues/204#issuecomment-441845116. So, for consistency, should we just remove the checking of uid/gid from code and rely on the fact that cfgofuse/WinFsp is handling this well, and in a way that lets rclone users override anything set by rclone on their command line? If not, the check function had a tiny miss: You can specify multiple options such as both uid and gid in the same comma separated string, so both `-o uid=-1 -o gid=11` and `-o uid=-1,gid=11` have same result, but the check function did only consider the first. Added a commit to fix this (just splitting on comma, which could still have an issue if individual values are allowed to contain quoted strings with commas or something like that)... just to be followed by a commit that removes the check entirely..!

- And another "spin-off" from the previous point: You can set username and groupnames now, and don't have to find the POSIX uid int numbers. For example group name can be set as `-o GroupName="Authenticated Users"` or `--fuse-flag --GroupName="Authenticated Users"` (both works). Same with "UserName". See https://github.com/rclone/rclone/issues/3216#issuecomment-509412571 (tip from comments). Not sure if this deserves documentation? Maybe not, can't document everyfuse/WinFsp option there is, unless they are especially relevant.
Edit: Found these new options has already been mentioned in forum: https://forum.rclone.org/t/the-winfsp-thread-how-new-features-affect-rclone/12488/3

- Error handling in rclone for invalid paths etc, or rely on winfsp/cgofuse to report understandable error message? This merge request introduces a somewhat fair amount of error handling in rclone part, but there are still many cases where winfsp fails and its error is reported back to user.

~- Update main documentation of the mount command. Note: Consider changes in #4714! Need to merge those first, and then update further.~

Not directly. But related merge request #4714 and issue #2678.

#### Checklist

- [ ] I have added tests for all changes in this PR if appropriate. *TODO: Are there any existing unit tests that should be extended?*

### DIFF
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/billziss-gh/cgofuse/fuse"
@@ -46,28 +45,9 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 		options = append(options, "-o", "debug")
 	}
 
-	// Determine if ExtraOptions already has an opt in
-	hasExtraOption := func(optionName string) bool {
-		optionName += "="
-		for _, option := range opt.ExtraOptions {
-			suboptions := strings.Split(option, ",")
-			for _, suboption := range suboptions {
-				if strings.HasPrefix(suboption, optionName) {
-					return true
-				}
-			}
-		}
-		return false
-	}
-
 	if runtime.GOOS == "windows" {
-		// Setting uid and gid to -1 by default, which mean current user in WinFsp
-		if !hasExtraOption("uid") {
-			options = append(options, "-o", "uid=-1")
-		}
-		if !hasExtraOption("gid") {
-			options = append(options, "-o", "gid=-1")
-		}
+		options = append(options, "-o", "uid=-1")
+		options = append(options, "-o", "gid=-1")
 		options = append(options, "--FileSystemName=rclone")
 		if opt.VolumeName != "" {
 			if opt.NetworkMode {
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -50,8 +50,11 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 	hasExtraOption := func(optionName string) bool {
 		optionName += "="
 		for _, option := range opt.ExtraOptions {
-			if strings.HasPrefix(option, optionName) {
-				return true
+			suboptions := strings.Split(option, ",")
+			for _, suboption := range suboptions {
+				if strings.HasPrefix(suboption, optionName) {
+					return true
+				}
 			}
 		}
 		return false
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -40,32 +40,14 @@ func init() {
 func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.Options) (options []string) {
 	// Options
 	options = []string{
-		"-o", "fsname=" + device,
-		"-o", "subtype=rclone",
-		"-o", fmt.Sprintf("max_readahead=%d", opt.MaxReadAhead),
 		"-o", fmt.Sprintf("attr_timeout=%g", opt.AttrTimeout.Seconds()),
-		// This causes FUSE to supply O_TRUNC with the Open
-		// call which is more efficient for cmount.  However
-		// it does not work with cgofuse on Windows with
-		// WinFSP so cmount must work with or without it.
-		"-o", "atomic_o_trunc",
 	}
 	if opt.DebugFUSE {
 		options = append(options, "-o", "debug")
 	}
 
-	// OSX options
-	if runtime.GOOS == "darwin" {
-		if opt.NoAppleDouble {
-			options = append(options, "-o", "noappledouble")
-		}
-		if opt.NoAppleXattr {
-			options = append(options, "-o", "noapplexattr")
-		}
-	}
-
-	// determine if ExtraOptions already has an opt in
-	hasOption := func(optionName string) bool {
+	// Determine if ExtraOptions already has an opt in
+	hasExtraOption := func(optionName string) bool {
 		optionName += "="
 		for _, option := range opt.ExtraOptions {
 			if strings.HasPrefix(option, optionName) {
@@ -75,13 +57,12 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 		return false
 	}
 
-	// Windows options
 	if runtime.GOOS == "windows" {
-		// These cause WinFsp to mean the current user
-		if !hasOption("uid") {
+		// Setting uid and gid to -1 by default, which mean current user in WinFsp
+		if !hasExtraOption("uid") {
 			options = append(options, "-o", "uid=-1")
 		}
-		if !hasOption("gid") {
+		if !hasExtraOption("gid") {
 			options = append(options, "-o", "gid=-1")
 		}
 		options = append(options, "--FileSystemName=rclone")
@@ -92,31 +73,47 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 				options = append(options, "-o", "volname="+opt.VolumeName)
 			}
 		}
-	} else if runtime.GOOS == "darwin" {
-		if opt.VolumeName != "" {
-			options = append(options, "-o", "volname="+opt.VolumeName)
+	} else {
+		options = append(options, "-o", "fsname="+device)
+		options = append(options, "-o", "subtype=rclone")
+		options = append(options, "-o", fmt.Sprintf("max_readahead=%d", opt.MaxReadAhead))
+		// This causes FUSE to supply O_TRUNC with the Open
+		// call which is more efficient for cmount.  However
+		// it does not work with cgofuse on Windows with
+		// WinFSP so cmount must work with or without it.
+		options = append(options, "-o", "atomic_o_trunc")
+		if opt.DaemonTimeout != 0 {
+			options = append(options, "-o", fmt.Sprintf("daemon_timeout=%d", int(opt.DaemonTimeout.Seconds())))
 		}
-	}
-	if opt.AllowNonEmpty {
-		options = append(options, "-o", "nonempty")
-	}
-	if opt.AllowOther {
-		options = append(options, "-o", "allow_other")
-	}
-	if opt.AllowRoot {
-		options = append(options, "-o", "allow_root")
-	}
-	if opt.DefaultPermissions {
-		options = append(options, "-o", "default_permissions")
-	}
-	if VFS.Opt.ReadOnly {
-		options = append(options, "-o", "ro")
-	}
-	if opt.WritebackCache {
-		// FIXME? options = append(options, "-o", WritebackCache())
-	}
-	if opt.DaemonTimeout != 0 {
-		options = append(options, "-o", fmt.Sprintf("daemon_timeout=%d", int(opt.DaemonTimeout.Seconds())))
+		if opt.AllowNonEmpty {
+			options = append(options, "-o", "nonempty")
+		}
+		if opt.AllowOther {
+			options = append(options, "-o", "allow_other")
+		}
+		if opt.AllowRoot {
+			options = append(options, "-o", "allow_root")
+		}
+		if opt.DefaultPermissions {
+			options = append(options, "-o", "default_permissions")
+		}
+		if VFS.Opt.ReadOnly {
+			options = append(options, "-o", "ro")
+		}
+		if opt.WritebackCache {
+			// FIXME? options = append(options, "-o", WritebackCache())
+		}
+		if runtime.GOOS == "darwin" {
+			if opt.VolumeName != "" {
+				options = append(options, "-o", "volname="+opt.VolumeName)
+			}
+			if opt.NoAppleDouble {
+				options = append(options, "-o", "noappledouble")
+			}
+			if opt.NoAppleXattr {
+				options = append(options, "-o", "noapplexattr")
+			}
+		}
 	}
 	for _, option := range opt.ExtraOptions {
 		options = append(options, "-o", option)
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -85,9 +85,14 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 			options = append(options, "-o", "gid=-1")
 		}
 		options = append(options, "--FileSystemName=rclone")
-	}
-
-	if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+		if opt.VolumeName != "" {
+			if opt.NetworkMode {
+				options = append(options, "--VolumePrefix="+opt.VolumeName)
+			} else {
+				options = append(options, "-o", "volname="+opt.VolumeName)
+			}
+		}
+	} else if runtime.GOOS == "darwin" {
 		if opt.VolumeName != "" {
 			options = append(options, "-o", "volname="+opt.VolumeName)
 		}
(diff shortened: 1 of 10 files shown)


ivandeex:
There is a related feature request https://github.com/rclone/rclone/issues/2678 "Use next available drive letter on Windows"

albertony:
> There is a related feature request #2678 "Use next available drive letter on Windows"

Yes, in the current version of this mr this will be handled by the unc path mountpoint case:
```
rclone mount remote: \\server\share
```

You will not get both auto-generated volume prefix (network share unc path) *and* drive letter, so will have to choose.

cmd/cmount/mount.go
```diff
@@ -135,6 +141,22 @@ func waitFor(fn func() bool) (ok bool) {
 	return false
 }
 
+// findUnusedDriveLetter searches mounted drive list on the system
```
ncw, line 144: We have some code which does this already here:

https://github.com/rclone/rclone/blob/30c8b1b84f548f08740148c6e9a086146af1f60d/vfs/vfstest/fs.go#L157-L165

Maybe we should factor your code out into `lib/file` (then we can have a non-windows version which always returns "") and call it from both places?

cmd/cmount/mount.go
```diff
@@ -143,17 +165,90 @@ func waitFor(fn func() bool) (ok bool) {
 // report an error when fusermount is called.
 func mount(VFS *vfs.VFS, mountpoint string, opt *mountlib.Options) (<-chan error, func() error, error) {
 	f := VFS.Fs()
-	fs.Debugf(f, "Mounting on %q", mountpoint)
 
-	// Check the mountpoint - in Windows the mountpoint mustn't exist before the mount
-	if runtime.GOOS != "windows" {
+	// Check the mountpoint
+	if runtime.GOOS == "windows" {
```
ncw, line 170: Now that this part has got quite long, it might be worth splitting it into a function in its own file.

We can then have a windows and a non-windows version which will be easier to read and save bloating the non-windows versions with code they don't need.

The logic in here is quite complicated now too, so it might be worth splitting the function up a bit to try to make it easier to follow.

ncw (commented):
I think this is looking nice!

See inline comments.

It is missing documentation though - that goes in `cmd/mountlib/help.go`

albertony:
Thanks. I will look into the suggested refactoring.

> It is missing documentation though - that goes in `cmd/mountlib/help.go`

I think it is just in mountlib\mount.go? I was delaying the documentation because I wanted to get some confirmation that this has a chance to get merged, and is not "too much".. Also I need to figure out what to do about changes in #4714, as there will be conflict. Should I wait for that MR to be merged into master, before syncing up this branch? Or should I merge that branch directly into this one? And if the latter, should that MR then just be skipped and included in this one or can they both later be merged without issue?

And one related thing, regarding the os specific flags I had a question about: I just found the [following](https://forum.rclone.org/t/cant-delete-empty-directory-because-of-ds-store/13003/2) quote from you which seem to indicate the list of flags should be same on all os and just documented if os-specific:

> Confusingly these aren't in the docs as they are macOS only and the docs are built on a linux machine. This is probably an oversight!

ncw:
> Thanks. I will look into the suggested refactoring.
> 
> > It is missing documentation though - that goes in `cmd/mountlib/help.go`
> 
> I think it is just in mountlib\mount.go? I was delaying the documentation because I wanted to get some confirmation that this has a chance to get merged, and is not "too much"..

I think it looks helpful for users :-)

> Also I need to figure out what to do about changes in #4714, as there will be conflict. Should I wait for that MR to be merged into master, before syncing up this branch? Or should I merge that branch directly into this one? And if the latter, should that MR then just be skipped and included in this one or can they both later be merged without issue?

I merged that now which should help things along!

> And one related thing, regarding the os specific flags I had a question about: I just found the [following](https://forum.rclone.org/t/cant-delete-empty-directory-because-of-ds-store/13003/2) quote from you which seem to indicate the list of flags should be same on all os and just documented if os-specific:

Yes that is probably the best thing to do as it keeps the docs cross platform.

> > Confusingly these aren't in the docs as they are macOS only and the docs are built on a linux machine. This is probably an oversight!
> 
> **Edit:** Have pushed a refactored version. Will probably see through the code a bit more, reconsidering names etc, but main parts I think is done. Feedback are, of course, welcome, and I will gladly improve code even more. The main thing left is the documentation, but before I can start I must decide what to do about the conflicting changes #4714. Any suggestions? Also if the default mode should be changed to network will affect the documentation. Feedback are welcome.

I suggest you rebase on master to pick up the changes from #4714 then edit as appropriate.

When you've done the docs, I'll give it another review - ping me here when you are ready.

Thank you :-)

ivandeex:
Related to https://github.com/rclone/rclone/issues/3216#issuecomment-503880038

albertony:
> This PR is related to:
> 
> * WinFSP internals: [#3216 (comment)](https://github.com/rclone/rclone/issues/3216#issuecomment-503880038)
> * WinFSP mount credentials: [#3216 (comment)](https://github.com/rclone/rclone/issues/3216#issuecomment-509412571)

Thank you for the tip. Relevant, yes, and interesting discussions. From first read, I'm not sure there are any specific TODOs to take from it, at this point..(?)

albertony:
> When you've done the docs, I'll give it another review - ping me here when you are ready.

@ncw: I'm ready with a refactored version, including documentation.

cmd/cmount/mount.go
```diff
@@ -53,28 +52,9 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 		options = append(options, "-o", "debug")
 	}
 
-	// Determine if ExtraOptions already has an opt in
```
ncw: Not sure why you've removed this code - the `hasExtraOption` works out if that option has been specified so it doesn't override the one supplied by the user.

ncw (commented):
I think this is looking great!

See one inline comment.

Note also the commit patching the vfs help should be `vfs: ....`

I don't know whether you want to squash any of these commits? It seems like you've made a logical progression of commits so I'm not sure they need squashing as each one is a different improvement, rather than working on some overall idea.

cmd/cmount/mount.go
albertony: If user supplies any options, those are added at the end. From my testing the last one wins. I can even add the same option multiple times as, e.g. `-o gid=11 -o gid=-1` and the resulting mount will use the last -1 value. The same is indicated here: https://github.com/billziss-gh/winfsp/issues/204#issuecomment-441845116.

I noticed the function were introduced after a [discussion](https://forum.rclone.org/t/issue-with-rclone-mount-and-resilio-sync/14730/28), but did not see any "proof" that it actually made a difference.

The reason for not just keeping it was that I found that you can specify multiple options, such as both uid and gid, in the same comma separated string. So both `-o uid=-1 -o gid=11` and `-o uid=-1,gid=11` have the same result. But the existing hasExtraOptions check did not detect this second case. I added a commit to fix this, but then realized there were still weaknesses. And then, what about the other options that rclone sets by default? "--FileSystemName", "--VolumePrefix" and "-o volname"? There is no check if these are set by user..

So I concluded it is better to rely on WinFsp handling this correctly, instead of trying to catch all corner cases ourselves.

Tried to explain in the section "Continuing from previous point, but more generally", but totally understand it is easy to miss in that all too long description! (darn it, this comment ended up quite long as well..)

albertony:
> It seems like you've made a logical progression of commits

Yes, I tried to do that.

> I don't know whether you want to squash any of these commits? 

I think there are one or two that are more or less intermediate commits, and should be squashed. One of them is related to your inline comment, and what to do with it depends on the conclusion there. When working on it, I can also see through the others as well to see if some of them can be combined, and perhaps reduce the number of commits a bit, at least.

ncw (approved):
This is a fantastic set of changes - thank you very much.

I've been through the changes one more time and I think they are ready to go now so I will merge them :-)

PR #4761, opened 2020-11-12
onedrive: Support addressing site by server-relative URL

kice:

Allow for addressing site by SPO URL, e.g. `https://contoso.sharepoint.com/sites/mysite` or `mysite`
1. By URL is more friendly than finding the site-id
2. For too many SPO sites, like more than 200, using URL is the only way

This method is using [server-relative URL](https://docs.microsoft.com/en-us/graph/api/site-get?view=graph-rest-1.0&tabs=http#access-a-site-by-server-relative-url), thus also adding a "low-level" support for it.

ref: https://docs.microsoft.com/en-us/graph/api/site-get?view=graph-rest-1.0&tabs=http#access-a-site-by-server-relative-url
ref2: https://docs.microsoft.com/en-us/graph/api/resources/sharepoint?view=graph-rest-1.0#sharepoint-api-root-resources

### DIFF
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"net/http"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -119,9 +120,15 @@ func init() {
 			var opts rest.Opts
 			var finalDriveID string
 			var siteID string
+			var relativePath string
 			switch config.Choose("Your choice",
-				[]string{"onedrive", "sharepoint", "driveid", "siteid", "search"},
-				[]string{"OneDrive Personal or Business", "Root Sharepoint site", "Type in driveID", "Type in SiteID", "Search a Sharepoint site"},
+				[]string{"onedrive", "sharepoint", "url", "driveid", "siteid", "path", "search"},
+				[]string{
+					"OneDrive Personal or Business",
+					"Root Sharepoint site", "Sharepoint site URL",
+					"Type in driveID", "Type in SiteID",
+					"Sharepoint site by server-relative URL", "Search a Sharepoint site",
+				},
 				false) {
 
 			case "onedrive":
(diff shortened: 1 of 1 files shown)


backend/onedrive/onedrive.go
```diff
@@ -142,6 +149,20 @@ func init() {
...
+			case "url":
+				fmt.Println("Example: \"https://contoso.sharepoint.com/sites/mysite\" or \"mysite\"")
+				fmt.Printf("Paste your Site URL here> ")
+				siteURL := config.ReadLine()
+				re := regexp.MustCompile(`https://.*\.sharepoint.com/sites/(.*)`)
+				match := re.FindStringSubmatch(siteURL)
+				if len(match) == 2 {
+					relativePath = "/sites/" + match[1]
+				} else {
+					relativePath = "/sites/" + siteURL
```
ncw, line 161: This looks strange adding `https://whatever.com` onto `/sites/` - is that correct?

backend/onedrive/onedrive.go
```diff
@@ -142,6 +149,20 @@ func init() {
...
+				}
+			case "path":
+				fmt.Printf("Enter server-relative URL here> ")
```
ncw, line 164: Is this something users would know what it is? Maybe we need an example here?

ncw (commented):
I think the change looks good.

I'm wondering whether it needs more docs though.  There isn't a section on how to configure onedrive explaining what all the options mean in [onedrive.md](https://github.com/rclone/rclone/blob/master/docs/content/onedrive.md)

Maybe it is self explanatory...

@Cnly what do you think of this PR - happy?

backend/onedrive/onedrive.go
kice: In short:
+ If user enter something like `https://contoso.sharepoint.com/sites/mysite`, then we find out the `mysite`
+ If not, then assuming the user already find out the `mysite` part, just use it as it is

Since something user could easily remember the `mysite` from `https://contoso.sharepoint.com/sites/mysite`, no need for entering the whole URL.

backend/onedrive/onedrive.go
kice: If you are looking for "what is server-relative URL", I don't think use would ever use it but will be very useful when we need it, like adding a Sharepoint group/team site. More or less this is the siteid equivalent but much more easy to find.

This option is the by-product of addressing the site by URL.

Here is an exmaple form MS Graph API Doc and some other cases
Assuming the user root site is `contoso.sharepoint.com`

| Path        | Description |
| ------------- |:-------------:|
| /teams/hr | The site associated with https://contoso.sharepoint.com/teams/hr |
| /teams/hr:/drive | Access the default **drive** for this site |
| /sites/mysite | The site associated with https://contoso.sharepoint.com/sites/mysite |

However, since we assume the path is pointing to a site, the second one, **pointing to a drive is not vaild**.

ref:
https://docs.microsoft.com/en-us/graph/api/site-getbypath?view=graph-rest-1.0
https://docs.microsoft.com/en-us/graph/api/resources/sharepoint?view=graph-rest-1.0#sharepoint-api-root-resources
https://stackoverflow.com/questions/45406451/how-can-i-get-the-siteid-of-the-current-site-with-microsoft-graph-api

Cnly (approved):
Looks good to me in general :) Thanks! I think we can reduce possible confusion by adding some more explanation and moving commonly used options to the front - see my comments.

ncw:
Please merge this when you are happy @Cnly :-)

Cnly:
Okay, let me apply those minor changes and merge this. Thanks again @kice!

PR #4765, opened 2020-11-13
docs: update documentation of commands delete,purge,rmdir,rmdirs

albertony:

Update documentation of commands delete,purge,rmdir,rmdirs. Mostly some "cross-referencing" between them because the distinction between these closely related commands can be a bit confusing, and it is easy to miss for example the "rmdirs" variant since it is not in the main "Commands" dropdown on rclone.org.

Not directly, but several forum posts discusses the usage of these commands.

#### Checklist

### DIFF
--- a/cmd/delete/delete.go
+++ b/cmd/delete/delete.go
@@ -21,25 +21,26 @@ func init() {
 
 var commandDefinition = &cobra.Command{
 	Use:   "delete remote:path",
-	Short: `Remove the contents of path.`,
+	Short: `Remove the files in path.`,
 	Long: `
 Remove the files in path.  Unlike ` + "`" + `purge` + "`" + ` it obeys include/exclude
 filters so can be used to selectively delete files.
 
-` + "`" + `rclone delete` + "`" + ` only deletes objects but leaves the directory structure
+` + "`rclone delete`" + ` only deletes files but leaves the directory structure
 alone. If you want to delete a directory and all of its contents use
-` + "`" + `rclone purge` + "`" + `
+the ` + "`purge`" + ` command.
 
 If you supply the --rmdirs flag, it will remove all empty directories along with it.
+You can also use the separate command ` + "`rmdir`" + ` or ` + "`rmdirs`" + ` to
+delete empty directories only.
 
-Eg delete all files bigger than 100MBytes
-
-Check what would be deleted first (use either)
+For example, to delete all files bigger than 100MBytes, you may first want to check what
+would be deleted (use either):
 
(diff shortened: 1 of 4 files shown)


edwardxml:
I'm glad you are having a go at this. The point you address has confused me in the past.

Your line 29 - Consider changing `objects` to `files`. I appreciate this is a pre-existing issue, not of your making but I have been trying to change some references from objects to files in the documentation except where it is clearly just relevant to an object store backend. 

Your line 37 - `Eg` is pretty widely used in the rclone documentation rather than `Example:` It might be easier to go with that equally satisfactory form rather than change the others? I wrestled with this issue in https://github.com/rclone/rclone/pull/4767

Mixed references to `directories` and `folders` are pretty prelevant elsewhere. I am trying to stick to dircetories as you have done here.

Ed.

albertony:
I will look into it, appreciate the feedback!
PS: Just saw your pull request regarding filtering documentation, great work.

edwardxml:
My original view was to prefer not to go with Eg without punctuation but I'm not strongly against it and mostly it is very prevalent in the existing docs.

albertony:
> My original view was to prefer not to go with Eg without punctuation but I'm not strongly against it and mostly it is very prevalent in the existing docs.

I just noticed this #4682 huge doc cleanup, which among other things replaced *a lot* of "eg" with "e.g.". I think I will continue that path..

edwardxml:
I'm not #4682 is so huge compared to what is out there in the wild but

edwardxml:
Off topic but I was looking at https://forum.rclone.org/t/encryption-at-rest-untrusted-destination-server/20539/2 I did have a minor go at the crypt backend page a while back and wasn't especially happy with the result. It sort of lacks the basic mechanism of how the remote and encrypt remotes work together and the way you encrypt stuff by uploading unencrypted files to the encrypted remote. Your initial comment there did seem to put it quite well. Maybe you could have a look at the crypt page and improve my efforts? Many of the forum explanations are much better than the docs pages. I suppose they answer a real question rather than second guessing what the question should be. Also people write much more freely in the forum.

albertony:
> I'm not sure #4682 is so huge compared to what is out there in the wild but I do prefer `e.g.` to `eg`. I'm happy to do the same...

I see your point.
Cleaning up everything might be too big a task to do in one go, so I think it is better to move in the right direction and just accept the inconsistency for now..

> Maybe you could have a look at the crypt page and improve my efforts.

I'm not so sure I will be able to improve it that much, but can take a look at it when I have time.

ncw:

I will merge it.

PR #4769, opened 2020-11-14
vfs: fix invalid cache path on windows when using :backend: as remote

### LINKED ISSUE #4544: `rclone mount` configured by env variables uses invalid VFS cache dir on Windows

Using `rclone mount` with a non-persistent configuration (all settings given by environment variables) & VFS caching enabled attempts to use a VFS cache dir with illegal characters (`:`) on Windows.

```
rclone v1.52.3
- os/arch: windows/amd64
- go version: go1.14.7
```

Windows Server Core 2019 LTSC (64-bit) Docker Container (but I suspect the problem appears on all Windows versions)

Seafile (But I suspect the problem appears with all backends)

```powershell
$Env:RCLONE_SEAFILE_URL="https://example.com"
$Env:RCLONE_SEAFILE_USER="me@example.com"
$Env:RCLONE_SEAFILE_PASS="mypassword"
$Env:RCLONE_SEAFILE_LIBRARY="mylibrary"
rclone mount :seafile: R: --vfs-cache-mode writes
```

#### A log from the command

As soon as attempting to write a file to the mounted drive:

```
2020/08/30 23:33:45 ERROR : IO error: open RW handle failed to make cache directory: make cache directory failed: mkdir C:\Users\ContainerAdministrator\AppData\Local\rclone\vfs\:seafile: The filename, directory name, or volume label syntax is incorrect.
```

#### Workaround

Create a temporary config.

```powershell
rclone config create seafile seafile
rclone mount seafile: R: --vfs-cache-mode writes
```

Using `--cache-dir` does not help, since rclone always appends `vfs\:seafile` to the custom directory path.

albertony:

When using the [:backend:](https://rclone.org/docs/#backend-path-to-dir) syntax for creating remote on the fly, the cache will try to create a directory with name ":backend", which contains illegal character in Windows.

Example:

```
rclone mount :webdav: Z: --fuse-flag --VolumePrefix=\server\home --vfs-cache-mode writes --webdav-url https://live.sysinternals.com
```

This fix will trim off the leading ":".

There is a workaround:
Set environment variable `RCLONE_CONFIG_name_TYPE=backend`, which will create an actual remote with specified name, and then use "name:" as the remote.

Example:

```
set RCLONE_CONFIG_MYREMOTE_TYPE=webdav&&rclone mount myremote: Z: --fuse-flag --VolumePrefix=\server\home --vfs-cache-mode writes --webdav-url https://live.sysinternals.com
```

**Note/question:**

After this fix, using the example above, if you create another mount using ":webdav:", but a completely different server (--webdav-url), it will end up using the same cache directory. This could potentially lead to issues, I think? If, for some reason, the first mount were shut down - but did not manage to upload all changes. When you create the second, different, mount it will start uploading from the previous cache, will it not? Should this be warned about? Either as an message (Error? Notice?) from the place in code where I did a change, and/or in the documentation?

Actually, you have potentially the same risk with regular remotes as well: If you create a remote with name "mycloud", then same as before: Shut it down in a way that leaves pending uploads. Then delete the remote from configuration and create a different one using the same name. Next time you mount the new remote it will pick up the existing cache, and start uploading.

The second case is probably not *that* likely to happen, but I think it deserves a little warning in the documentation?
Using the workaround will be the same case, if you dont use the same name for different remotes you are fine, but perhaps a bit more likely that you will use the same name now, since the remote name are just temporary.
The first case is even more likely to happen, as all you have to do is to use the "on the fly" remote functionality with several of the same backend types (assuming the case of pending uploads after shutdown is something that may occur now and then).

**Edit: Thinking about possible solutions..**
- Could there be a config flag of some kind, that are set first time a remote is used - or at least when it creates its cache for the first time? So that when this flag is not set on startup, rclone would exit with fatal if cache directory already exists. On later startups it would accept that it does not exist.
- I was also thinking about making cache location configurable per remote, independent of each other. Maybe this could be combined: If cache location is configured then any existing location on first startup is used as is, user's responsibility. If cache location is not configured on first startup then rclone builds a default path - and now checks it does not exist, and then stores this in config. Downside is that you cannot easily relocate the entire cache by just setting --cache-dir, this would then just change the default for any future remotes, since existing remotes have their cache path stored... That could be a big downside, so would probably need to do something more/else.. 

**Edit: From the issue #4544:**

> Though it might be better to substitute the : with a character that can't appear in remote names but can appear in Windows file names rather than removing it, say =

I have not done this, just removed the ":" prefix (which by the way, should only be possible with this special "on the fly" remotes, and then there should always be exactly one occurence - at index 0, since `Remote names may only contain 0-9, A-Z ,a-z ,_ , - and space`, quote from https://rclone.org/docs/#valid-remote-names). By just using the backend name, as I have done, there is an additional risk of name clash if user configures a remote with the same name as the backend type (for instance configures a remote of type "webdav" and gives it name "webdav" also). Could add a character as suggested to avoid this (but perhaps append it to the end instead, to get more natural sort order?), but will wait for a decision regarding my additional concerns above, in case this leads to a more thorough fix that handles all cases.
 
https://forum.rclone.org/t/rclone-one-liner-to-connect-a-webdav-mount-without-rclone-conf/20444

Edit: Got a tip that there was actually an issue for this as well - this will fix #4544

#### Checklist

### DIFF
--- a/vfs/vfscache/cache.go
+++ b/vfs/vfscache/cache.go
@@ -74,21 +74,26 @@ type AddVirtualFn func(remote string, size int64, isDir bool) error
 // This starts background goroutines which can be cancelled with the
 // context passed in.
 func New(ctx context.Context, fremote fs.Fs, opt *vfscommon.Options, avFn AddVirtualFn) (*Cache, error) {
+	fName := fremote.Name()
 	fRoot := filepath.FromSlash(fremote.Root())
 	if runtime.GOOS == "windows" {
 		if strings.HasPrefix(fRoot, `\\?`) {
 			fRoot = fRoot[3:]
 		}
 		fRoot = strings.Replace(fRoot, ":", "", -1)
+		// Replace leading ':' if remote was created on the fly as ":backend:/path" as it is illegal in Windows
+		if fName[0] == ':' {
+			fName = "^" + fName[1:]
+		}
 	}
 	cacheDir := config.CacheDir
 	cacheDir, err := filepath.Abs(cacheDir)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to make --cache-dir absolute")
 	}
-	root := file.UNCPath(filepath.Join(cacheDir, "vfs", fremote.Name(), fRoot))
+	root := file.UNCPath(filepath.Join(cacheDir, "vfs", fName, fRoot))
 	fs.Debugf(nil, "vfs cache: root is %q", root)
-	metaRoot := file.UNCPath(filepath.Join(cacheDir, "vfsMeta", fremote.Name(), fRoot))
+	metaRoot := file.UNCPath(filepath.Join(cacheDir, "vfsMeta", fName, fRoot))
 	fs.Debugf(nil, "vfs cache: metadata root is %q", root)
 
 	fcache, err := fscache.Get(ctx, root)


vfs/vfscache/cache.go
```diff
@@ -74,16 +74,21 @@ type AddVirtualFn func(remote string, size int64, isDir bool) error
...
+	fName := fremote.Name()
 	fRoot := filepath.FromSlash(fremote.Root())
 	if runtime.GOOS == "windows" {
 		if strings.HasPrefix(fRoot, `\\?`) {
 			fRoot = fRoot[3:]
 		}
 		fRoot = strings.Replace(fRoot, ":", "", -1)
+		// Remove leading ':' if remote was created on the fly as ":backend:/path"
+		if fName[0] == ':' {
+			fName = fName[1:]
```
ncw, line 86: I think we need to replace the ':' with another character, preferably one that doesn't occur in backend names otherwise there can be collisions.

The only non-alpha characters allowed in backends is `_ -` so a non alpha character which is allowed in windows file names not from that list - how about `^` - that doesn't appear on anyone's list of illegal filename characters.

ncw (commented):
Apologies for the very long time it has taken to review this.

I'm catching up with the PRs slowly!

> After this fix, using the example above, if you create another mount using ":webdav:", but a completely different server (--webdav-url), it will end up using the same cache directory. This could potentially lead to issues, I think? If, for some reason, the first mount were shut down - but did not manage to upload all changes. When you create the second, different, mount it will start uploading from the previous cache, will it not? Should this be warned about? Either as an message (Error? Notice?) from the place in code where I did a change, and/or in the documentation?

Yes this can cause issues definitely.

I'm planning on addressing this after the 1.54 release where I'm going to implement the concept of the "config string" which will need a unique caching key.

> Actually, you have potentially the same risk with regular remotes as well: If you create a remote with name "mycloud", then same as before: Shut it down in a way that leaves pending uploads. Then delete the remote from configuration and create a different one using the same name. Next time you mount the new remote it will pick up the existing cache, and start uploading.

Hmm, that is harder to address...

> The second case is probably not _that_ likely to happen, but I think it deserves a little warning in the documentation?

Possibly, yes.

> Using the workaround will be the same case, if you dont use the same name for different remotes you are fine, but perhaps a bit more likely that you will use the same name now, since the remote name are just temporary.
> The first case is even more likely to happen, as all you have to do is to use the "on the fly" remote functionality with several of the same backend types (assuming the case of pending uploads after shutdown is something that may occur now and then).
> 
> **Edit: Thinking about possible solutions..**
> 
> * Could there be a config flag of some kind, that are set first time a remote is used - or at least when it creates its cache for the first time? So that when this flag is not set on startup, rclone would exit with fatal if cache directory already exists. On later startups it would accept that it does not exist.

I had the idea that the directory used should be a hash of the backend config, but only the backend config which could actually change which files get looked at. So --drive-use-trash wouldn't be in the hash but --drive-root-id would be.

> * I was also thinking about making cache location configurable per remote, independent of each other. Maybe this could be combined: If cache location is configured then any existing location on first startup is used as is, user's responsibility. If cache location is not configured on first startup then rclone builds a default path - and now checks it does not exist, and then stores this in config. Downside is that you cannot easily relocate the entire cache by just setting --cache-dir, this would then just change the default for any future remotes, since existing remotes have their cache path stored... That could be a big downside, so would probably need to do something more/else..

Interesting idea but it will complicate the vfs cache directory....

> **Edit: From the issue #4544:**
> 
> > Though it might be better to substitute the : with a character that can't appear in remote names but can appear in Windows file names rather than removing it, say =
> 
> I have not done this, just removed the ":" prefix (which by the way, should only be possible with this special "on the fly" remotes, and then there should always be exactly one occurence - at index 0, since `Remote names may only contain 0-9, A-Z ,a-z ,_ , - and space`, quote from https://rclone.org/docs/#valid-remote-names). By just using the backend name, as I have done, there is an additional risk of name clash if user configures a remote with the same name as the backend type (for instance configures a remote of type "webdav" and gives it name "webdav" also). Could add a character as suggested to avoid this (but perhaps append it to the end instead, to get more natural sort order?), but will wait for a decision regarding my additional concerns above, in case this leads to a more thorough fix that handles all cases.

Yes I think this probably could be a problem in practice. In the comments I've suggested using `^` as a safe character.

Can you make the commit message say `Fixes #4544` in it too please :-)

albertony:
> Apologies for the very long time it has taken to review this.

No worries.

> Yes I think this probably could be a problem in practice. In the comments I've suggested using ^ as a safe character.

> Can you make the commit message say Fixes #4544 in it too please :-)

Done. Also did a rebase master.

ncw (approved):
Very nice :-)

I'll merge this now - thank you very much.

PR #4782, opened 2020-11-17
cmount: Add optional `brew` tag

### LINKED ISSUE #4775: Homebrew depreciation
Homebrew installation is being depreciated

rclone v1.53.2-DEV
- os/arch: darwin/amd64
- go version: go1.15.3

None of this other stuff applies to this issue but I read them.
I took a look for reports of this issue already but didn't see one. This seems to be a pretty new change too. The problem is related to the fact fuse is required. See https://github.com/Homebrew/homebrew-core/commit/4f0402fb80551779c603add4c8b67a51bd324f09 for more details.

darthShadow:
cmount: Add optional `brew` tag to throw an error when using mount in the binaries installed via Homebrew - Fixes #4775

Add optional `brew` tag to throw an error when using mount in the binaries installed via Homebrew

#4775 

#### Checklist

### DIFF
--- a/cmd/cmount/mount_brew.go
+++ b/cmd/cmount/mount_brew.go
@@ -1,0 +1,32 @@
+// Build for macos with the brew tag to handle the absence
+// of fuse and print an appropriate error message
+
+// +build brew
+// +build darwin
+
+package cmount
+
+import (
+	"github.com/pkg/errors"
+	"github.com/rclone/rclone/cmd/mountlib"
+	"github.com/rclone/rclone/vfs"
+)
+
+func init() {
+	name := "mount"
+	cmd := mountlib.NewMountCommand(name, false, mount)
+	cmd.Aliases = append(cmd.Aliases, "cmount")
+	mountlib.AddRc("cmount", mount)
+}
+
+// mount the file system
+//
+// The mount point will be ready when this returns.
+//
+// returns an error, and an error channel for the serve process to
+// report an error when fusermount is called.
+func mount(_ *vfs.VFS, _ string, _ *mountlib.Options) (<-chan error, func() error, error) {
+	return nil, nil, errors.New("mount is not supported on MacOS when installed via Homebrew. " +
+		"Please install the binaries available at https://rclone." +
+		"org/downloads/ instead if you want to use the mount command")
+}
--- a/cmd/cmount/mount_unsupported.go
+++ b/cmd/cmount/mount_unsupported.go
@@ -1,6 +1,6 @@
 // Build for cmount for unsupported platforms to stop go complaining
 // about "no buildable Go source files "
 
-// +build !linux,!darwin,!freebsd,!windows !cgo !cmount
+// +build !linux,!darwin,!freebsd,!windows !brew !cgo !cmount
 
 package cmount
--- a/cmd/mount/mount_unsupported.go
+++ b/cmd/mount/mount_unsupported.go
@@ -1,7 +1,7 @@
 // Build for mount for unsupported platforms to stop go complaining
 // about "no buildable Go source files "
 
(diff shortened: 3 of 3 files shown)


darthShadow:
This is the error message thrown when using mount or cmount with the brew build tag:

```
2020/11/17 22:34:05 Fatal error: failed to mount FUSE fs: mount is not supported on MacOS when installed via Homebrew. Please install the binaries available at https://rclone.org/downloads/ instead if you want to use the mount command
```

ivandeex:
@ncw @darthShadow 

https://github.com/rclone/rclone/issues/4775#issuecomment-728197928 suggests a `macos caveat` note in https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone_mount.md#limitations

> [...] the right thing to do is [...] add something in the `caveats` stanza that mentions that mount won't work without fuse and leave it up to the user to decide how they want to handle.

darthShadow:
Yep, I intend to add this caveat to the homebrew formula and not to the rclone docs.

darthShadow:
Just to clarify on this, my thought process has always been to reflect the limitations etc. of the actual binaries available on the website on the respective pages. The package-specific or OS-specific limitations should always be present in those docs, especially since Homebrew provides an easy way to add caveats in the formula.

xaocon:
I may not understand something. Why not just fail gracefully if fuse isn't available in all builds rather than take functionality out for some builds?

cmd/cmount/mount_brew.go
```diff
@@ -0,0 +1,32 @@
+// Build for macos with the brew tag to handle the absence
+// of fuse and print an appropriate error message
+
+// +build brew
```
ncw, line 4: Does this need to mention the cmount tag? Or do we do this regardless of the cmount tag?

If we can expect `cmount` and `brew` to be set then we need to adjust the constraints for `cmount/{fs,mount}.go` something like:

```
// +build cmount
// +build cgo
// +build linux darwin,!brew freebsd windows
```

ncw (commented):
I think this is OK.

Query about whether we expect brew + cmount inline

cmd/cmount/mount_brew.go
darthShadow: I will just be replacing the `cmount` tag with the `brew` tag in Homebrew so this case probably won't happen there but I can add it anyway to be safe if you want?

ncw: No I think that is fine. Adding too many build constraints makes my head explode anyway as there are too many combinations to think through.

ncw:
As far as I can see we now have two options for building with brew

1. build with `brew` build tag - `rclone mount` gives a nice error message
2. build with `cmount` build tag. Since the rclone binary doesn't depend on fuse being installed then it doesn't need to have a FUSE dependency in brew.

Option 2) would be my preference if that is OK with the brew developers, but it might not be possible as they will need FUSE installed on the dev machine to build rclone with the `cmount` tag.

I note that removing OSXFUSE from brew will break rclone's build

https://github.com/rclone/rclone/blob/211b08f771472cbe6f0a60824010b84303d9b12e/.github/workflows/build.yml#L130

So we'll have to fix that too at some point!

darthShadow:
> I note that removing OSXFUSE from brew will break rclone's build

I don't think that's the end result. Based on my understanding of the discussions at https://github.com/Homebrew/homebrew-core/pull/64491, it is just that they are removing any formula that depends on a Cask and not the cask itself. Anyway, I will confirm this understanding once.

darthShadow:
This change has been accepted in the homebrew repository too so this PR can be merged now.

ncw:
Thank you @darthShadow - I'll merge this now

Issue #3923, opened 2020-02-02
--download option for md5sum


aicioara:
#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.50.2
- os/arch: linux/amd64
- go version: go1.13.4
```

#### What problem are you are trying to solve?

The md5sum that rclone stores with the object is not guaranteed to be the md5sum of the actual object for multipart files (see https://github.com/rclone/rclone/issues/523). This means that the output of the following command is not guaranteed to be correct

```
rclone md5sum remote://path/to/file
```

This then cascades to `rclone check` as well.

#### How do you think rclone should be changed to solve that?

Rclone should add a `--download` flag for the md5sum command (possibly for hashsum as well) which downloads the file and returns the md5sum.

`--download` is a suggestion based on the `rclone check` command, but other options should also work.

ncw:
That isn't a bad suggestion.. Do you want to work on it?

You could use `rclone cat remote:path/to/file | md5sum -` as a work-around.

ivandeex:
this feature was merged by #4803 and released with 1.54

PR #4803, opened 2020-11-24
operations: Add a download flag to hashsum and related commands to force rclone to download and hash files locally

lostheli:
Add a download flag to hashsum and related commands to force rclone to download and hash files locally

This commit modifies the operations.hashSum function by adding an alternate code path.  This code path is triggered by passing downloadFlag = True.  When activated, rclone will download files from the remote and hash them locally.  downloadFlag = False preserves the existing behavior of using the remote to retrieve the hash.

This commit modifies HashLister to support the new hashSum method as well as consolidating the roles of HashLister, HashListerBase64, Md5sum, and Sha1sum.

The unit test operations_test.TestHashSums is modified to support this change and test the download functionality.

The command functions hashsum, md5sum, sha1sum, and dbhashsum are modified to support this change.

https://forum.rclone.org/t/force-download-hash/20333

#### Checklist

This pull request requires functionality and project best practices review before finalization.

### DIFF
--- a/cmd/dbhashsum/dbhashsum.go
+++ b/cmd/dbhashsum/dbhashsum.go
@@ -2,10 +2,10 @@ package dbhashsum
 
 import (
 	"context"
-	"os"
 
 	"github.com/rclone/rclone/backend/dropbox"
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/hashsum"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
@@ -13,6 +13,8 @@ import (
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	hashsum.AddHashFlags(cmdFlags)
 }
 
 var commandDefinition = &cobra.Command{
@@ -23,6 +25,11 @@ Produces a Dropbox hash file for all the objects in the path.  The
 hashes are calculated according to [Dropbox content hash
 rules](https://www.dropbox.com/developers/reference/content-hash).
 The output is in the same format as md5sum and sha1sum.
+
+By default, the hash is requested from the remote.  If Dropbox hash is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling Dropbox hash for any remote.
 `,
 	Hidden: true,
 	Run: func(command *cobra.Command, args []string) {
@@ -30,7 +37,15 @@ The output is in the same format as md5sum and sha1sum.
 		fsrc := cmd.NewFsSrc(args)
 		fs.Logf(nil, `"rclone dbhashsum" is deprecated, use "rclone hashsum %v %s" instead`, dropbox.DbHashType, args[0])
 		cmd.Run(false, false, command, func() error {
-			return operations.HashLister(context.Background(), dropbox.DbHashType, fsrc, os.Stdout)
+			if hashsum.HashsumOutfile == "" {
+				return operations.HashLister(context.Background(), dropbox.DbHashType, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, nil)
+			}
+			output, close, err := hashsum.GetHashsumOutput(hashsum.HashsumOutfile)
+			if err != nil {
+				return err
+			}
+			defer close()
+			return operations.HashLister(context.Background(), dropbox.DbHashType, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, output)
 		})
 	},
 }
--- a/cmd/hashsum/hashsum.go
+++ b/cmd/hashsum/hashsum.go
@@ -2,25 +2,55 @@ package hashsum
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 
+	"github.com/pkg/errors"
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
+// Global hashsum flags for reuse in md5sum, sha1sum, and dbhashsum
 var (
-	outputBase64 = false
+	OutputBase64   = false
+	DownloadFlag   = false
+	HashsumOutfile = ""
 )
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
 	cmdFlags := commandDefinition.Flags()
-	flags.BoolVarP(cmdFlags, &outputBase64, "base64", "", outputBase64, "Output base64 encoded hashsum")
+	AddHashFlags(cmdFlags)
+}
+
+// AddHashFlags is a convenience function to add the command flags OutputBase64 and DownloadFlag to hashsum, md5sum, sha1sum, and dbhashsum
+func AddHashFlags(cmdFlags *pflag.FlagSet) {
+	flags.BoolVarP(cmdFlags, &OutputBase64, "base64", "", OutputBase64, "Output base64 encoded hashsum")
+	flags.StringVarP(cmdFlags, &HashsumOutfile, "output-file", "", HashsumOutfile, "Output hashsums to a file rather than the terminal")
+	flags.BoolVarP(cmdFlags, &DownloadFlag, "download", "", DownloadFlag, "Download the file and hash it locally; if this flag is not specified, the hash is requested from the remote")
+}
+
+// GetHashsumOutput opens and closes the output file when using the output-file flag
+func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
+	out, err = os.Create(filename)
+	if err != nil {
+		err = errors.Wrapf(err, "Failed to open output file %v", filename)
+		return nil, nil, err
+	}
+
+	close = func() {
+		err := out.Close()
+		if err != nil {
+			fs.Errorf(nil, "Failed to close output file %v: %v", filename, err)
+		}
+	}
+
+	return out, close, nil
 }
 
 var commandDefinition = &cobra.Command{
@@ -31,7 +61,12 @@ Produces a hash file for all the objects in the path using the hash
 named.  The output is in the same format as the standard
 md5sum/sha1sum tool.
 
-Run without a hash to see the list of supported hashes, e.g.
+By default, the hash is requested from the remote.  If the hash is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling any hash for any remote.
+
+Run without a hash to see the list of all supported hashes, e.g.
 
     $ rclone hashsum
     Supported hashes are:
@@ -61,11 +96,17 @@ Then
 			return err
 		}
 		fsrc := cmd.NewFsSrc(args[1:])
+
 		cmd.Run(false, false, command, func() error {
-			if outputBase64 {
-				return operations.HashListerBase64(context.Background(), ht, fsrc, os.Stdout)
+			if HashsumOutfile == "" {
+				return operations.HashLister(context.Background(), ht, OutputBase64, DownloadFlag, fsrc, nil)
 			}
-			return operations.HashLister(context.Background(), ht, fsrc, os.Stdout)
+			output, close, err := GetHashsumOutput(HashsumOutfile)
+			if err != nil {
+				return err
+			}
+			defer close()
+			return operations.HashLister(context.Background(), ht, OutputBase64, DownloadFlag, fsrc, output)
 		})
 		return nil
 	},
--- a/cmd/md5sum/md5sum.go
+++ b/cmd/md5sum/md5sum.go
@@ -2,15 +2,18 @@ package md5sum
 
 import (
 	"context"
-	"os"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/hashsum"
+	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	hashsum.AddHashFlags(cmdFlags)
 }
 
 var commandDefinition = &cobra.Command{
@@ -19,12 +22,25 @@ var commandDefinition = &cobra.Command{
 	Long: `
 Produces an md5sum file for all the objects in the path.  This
 is in the same format as the standard md5sum tool produces.
+
+By default, the hash is requested from the remote.  If MD5 is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling MD5 for any remote.
 `,
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
 		cmd.Run(false, false, command, func() error {
-			return operations.Md5sum(context.Background(), fsrc, os.Stdout)
+			if hashsum.HashsumOutfile == "" {
+				return operations.HashLister(context.Background(), hash.MD5, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, nil)
+			}
+			output, close, err := hashsum.GetHashsumOutput(hashsum.HashsumOutfile)
+			if err != nil {
+				return err
+			}
+			defer close()
+			return operations.HashLister(context.Background(), hash.MD5, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, output)
 		})
 	},
 }
--- a/cmd/sha1sum/sha1sum.go
+++ b/cmd/sha1sum/sha1sum.go
@@ -2,15 +2,18 @@ package sha1sum
 
 import (
 	"context"
-	"os"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/hashsum"
+	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	hashsum.AddHashFlags(cmdFlags)
 }
 
 var commandDefinition = &cobra.Command{
@@ -19,12 +22,25 @@ var commandDefinition = &cobra.Command{
 	Long: `
 Produces an sha1sum file for all the objects in the path.  This
 is in the same format as the standard sha1sum tool produces.
+
+By default, the hash is requested from the remote.  If SHA-1 is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling SHA-1 for any remote.
 `,
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
 		cmd.Run(false, false, command, func() error {
-			return operations.Sha1sum(context.Background(), fsrc, os.Stdout)
+			if hashsum.HashsumOutfile == "" {
+				return operations.HashLister(context.Background(), hash.SHA1, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, nil)
+			}
+			output, close, err := hashsum.GetHashsumOutput(hashsum.HashsumOutfile)
+			if err != nil {
+				return err
+			}
+			defer close()
+			return operations.HashLister(context.Background(), hash.SHA1, hashsum.OutputBase64, hashsum.DownloadFlag, fsrc, output)
 		})
 	},
 }
--- a/docs/content/commands/rclone_hashsum.md
+++ b/docs/content/commands/rclone_hashsum.md
@@ -16,7 +16,12 @@ Produces a hash file for all the objects in the path using the hash
 named.  The output is in the same format as the standard
 md5sum/sha1sum tool.
 
-Run without a hash to see the list of supported hashes, eg
+By default, the hash is requested from the remote.  If the hash is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling any hash for any remote.
+
+Run without a hash to see the list of all supported hashes, e.g.
 
     $ rclone hashsum
     Supported hashes are:
@@ -37,8 +42,10 @@ rclone hashsum <hash> remote:path [flags]
 ## Options
 
 ```
-      --base64   Output base64 encoded hashsum
-  -h, --help     help for hashsum
+      --base64               Output base64 encoded hashsum
+      --download             Download the file and hash it locally; if this flag is not specified, the hash is requested from the remote
+  -h, --help                 help for hashsum
+      --output-file string   Output hashsums to a file rather than the terminal
 ```
 
 See the [global flags page](/flags/) for global options not listed here.
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -797,13 +797,27 @@ func ListFn(ctx context.Context, f fs.Fs, fn func(fs.Object)) error {
 // mutex for synchronized output
 var outMutex sync.Mutex
 
+// SyncPrintf is a global var holding the Printf function used in syncFprintf so that it can be overridden
+// Note, despite name, does not provide sync and should not be called directly
+// Call syncFprintf, which provides sync
+var SyncPrintf = func(format string, a ...interface{}) {
+	fmt.Printf(format, a...)
+}
+
 // Synchronized fmt.Fprintf
 //
 // Ignores errors from Fprintf
+//
+// Updated to print to terminal if no writer is defined
+// This special behavior is used to allow easier replacement of the print to terminal code by progress
 func syncFprintf(w io.Writer, format string, a ...interface{}) {
 	outMutex.Lock()
 	defer outMutex.Unlock()
-	_, _ = fmt.Fprintf(w, format, a...)
+	if w == nil {
+		SyncPrintf(format, a...)
+	} else {
+		_, _ = fmt.Fprintf(w, format, a...)
+	}
 }
 
 // List the Fs to the supplied writer
@@ -833,63 +847,103 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
 	})
 }
 
-// Md5sum list the Fs to the supplied writer
-//
-// Produces the same output as the md5sum command - obeys includes and
-// excludes
-//
-// Lists in parallel which may get them out of order
-func Md5sum(ctx context.Context, f fs.Fs, w io.Writer) error {
-	return HashLister(ctx, hash.MD5, f, w)
-}
-
-// Sha1sum list the Fs to the supplied writer
-//
-// Obeys includes and excludes
-//
-// Lists in parallel which may get them out of order
-func Sha1sum(ctx context.Context, f fs.Fs, w io.Writer) error {
-	return HashLister(ctx, hash.SHA1, f, w)
-}
-
 // hashSum returns the human readable hash for ht passed in.  This may
 // be UNSUPPORTED or ERROR. If it isn't returning a valid hash it will
 // return an error.
-func hashSum(ctx context.Context, ht hash.Type, o fs.Object) (string, error) {
+func hashSum(ctx context.Context, ht hash.Type, downloadFlag bool, o fs.Object) (string, error) {
+	var sum string
 	var err error
-	tr := accounting.Stats(ctx).NewCheckingTransfer(o)
-	defer func() {
-		tr.Done(ctx, err)
-	}()
-	sum, err := o.Hash(ctx, ht)
-	if err == hash.ErrUnsupported {
-		sum = "UNSUPPORTED"
-	} else if err != nil {
-		fs.Debugf(o, "Failed to read %v: %v", ht, err)
-		sum = "ERROR"
+
+	// If downloadFlag is true, download and hash the file.
+	// If downloadFlag is false, call o.Hash asking the remote for the hash
+	if downloadFlag {
+		// Setup: Define accounting, open the file with NewReOpen to provide restarts, account for the transfer, and setup a multi-hasher with the appropriate type
+		// Execution: io.Copy file to hasher, get hash and encode in hex
+
+		tr := accounting.Stats(ctx).NewTransfer(o)
+		defer func() {
+			tr.Done(ctx, err)
+		}()
+
+		// Open with NewReOpen to provide restarts
+		var options []fs.OpenOption
+		for _, option := range fs.GetConfig(ctx).DownloadHeaders {
+			options = append(options, option)
+		}
+		in, err := NewReOpen(ctx, o, fs.GetConfig(ctx).LowLevelRetries, options...)
+		if err != nil {
+			return "ERROR", errors.Wrapf(err, "Failed to open file %v", o)
+		}
+
+		// Account and buffer the transfer
+		in = tr.Account(ctx, in).WithBuffer()
+
+		// Setup hasher
+		hasher, err := hash.NewMultiHasherTypes(hash.NewHashSet(ht))
+		if err != nil {
+			return "UNSUPPORTED", errors.Wrap(err, "Hash unsupported")
+		}
+
+		// Copy to hasher, downloading the file and passing directly to hash
+		_, err = io.Copy(hasher, in)
+		if err != nil {
+			return "ERROR", errors.Wrap(err, "Failed to copy file to hasher")
+		}
+
+		// Get hash and encode as hex
+		byteSum, err := hasher.Sum(ht)
+		if err != nil {
+			return "ERROR", errors.Wrap(err, "Hasher returned an error")
+		}
+		sum = hex.EncodeToString(byteSum)
+	} else {
+		tr := accounting.Stats(ctx).NewCheckingTransfer(o)
+		defer func() {
+			tr.Done(ctx, err)
+		}()
+
+		sum, err = o.Hash(ctx, ht)
+		if err == hash.ErrUnsupported {
+			return "UNSUPPORTED", errors.Wrap(err, "Hash unsupported")
+		} else if err != nil {
+			return "ERROR", errors.Wrapf(err, "Failed to get hash %v from backed: %v", ht, err)
+		}
 	}
-	return sum, err
+
+	return sum, nil
 }
 
 // HashLister does an md5sum equivalent for the hash type passed in
-func HashLister(ctx context.Context, ht hash.Type, f fs.Fs, w io.Writer) error {
-	return ListFn(ctx, f, func(o fs.Object) {
-		sum, _ := hashSum(ctx, ht, o)
-		syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
+// Updated to handle both standard hex encoding and base64
+// Updated to perform multiple hashes concurrently
+func HashLister(ctx context.Context, ht hash.Type, outputBase64 bool, downloadFlag bool, f fs.Fs, w io.Writer) error {
+	concurrencyControl := make(chan struct{}, fs.GetConfig(ctx).Transfers)
+	var wg sync.WaitGroup
+	err := ListFn(ctx, f, func(o fs.Object) {
+		wg.Add(1)
+		concurrencyControl <- struct{}{}
+		go func() {
+			defer func() {
+				<-concurrencyControl
+				wg.Done()
+			}()
+			sum, err := hashSum(ctx, ht, downloadFlag, o)
+			if outputBase64 && err == nil {
+				hexBytes, _ := hex.DecodeString(sum)
+				sum = base64.URLEncoding.EncodeToString(hexBytes)
+				width := base64.URLEncoding.EncodedLen(hash.Width(ht) / 2)
+				syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
+			} else {
+				syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
+			}
+			if err != nil {
+				err = fs.CountError(err)
+				fs.Errorf(o, "%v", err)
+			}
+		}()
 	})
-}
-
-// HashListerBase64 does an md5sum equivalent for the hash type passed in with base64 encoded
-func HashListerBase64(ctx context.Context, ht hash.Type, f fs.Fs, w io.Writer) error {
-	return ListFn(ctx, f, func(o fs.Object) {
-		sum, err := hashSum(ctx, ht, o)
-		if err == nil {
-			hexBytes, _ := hex.DecodeString(sum)
-			sum = base64.URLEncoding.EncodeToString(hexBytes)
-		}
-		width := base64.URLEncoding.EncodedLen(hash.Width(ht) / 2)
-		syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
-	})
+	wg.Wait()
+	return err
 }
 
 // Count counts the objects and their sizes in the Fs
--- a/cmd/progress.go
+++ b/cmd/progress.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
 	"github.com/rclone/rclone/fs/log"
+	"github.com/rclone/rclone/fs/operations"
 	"github.com/rclone/rclone/lib/terminal"
 )
 
@@ -28,6 +29,8 @@ const (
 func startProgress() func() {
 	stopStats := make(chan struct{})
 	oldLogPrint := fs.LogPrint
+	oldSyncPrint := operations.SyncPrintf
+
 	if !log.Redirected() {
 		// Intercept the log calls if not logging to file or syslog
 		fs.LogPrint = func(level fs.LogLevel, text string) {
@@ -35,6 +38,12 @@ func startProgress() func() {
 
 		}
 	}
+
+	// Intercept output from functions such as HashLister to stdout
+	operations.SyncPrintf = func(format string, a ...interface{}) {
+		printProgress(fmt.Sprintf(format, a...))
+	}
+
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
@@ -52,6 +61,7 @@ func startProgress() func() {
 				ticker.Stop()
 				printProgress("")
 				fs.LogPrint = oldLogPrint
+				operations.SyncPrintf = oldSyncPrint
 				fmt.Println("")
 				return
 			}
--- a/docs/content/commands/rclone_md5sum.md
+++ b/docs/content/commands/rclone_md5sum.md
@@ -15,6 +15,11 @@ Produces an md5sum file for all the objects in the path.
 Produces an md5sum file for all the objects in the path.  This
 is in the same format as the standard md5sum tool produces.
 
+By default, the hash is requested from the remote.  If MD5 is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling MD5 for any remote.
+
 
 ```
 rclone md5sum remote:path [flags]
@@ -23,7 +28,10 @@ rclone md5sum remote:path [flags]
 ## Options
 
 ```
-  -h, --help   help for md5sum
+      --base64               Output base64 encoded hashsum
+      --download             Download the file and hash it locally; if this flag is not specified, the hash is requested from the remote
+  -h, --help                 help for md5sum
+      --output-file string   Output hashsums to a file rather than the terminal
 ```
 
 See the [global flags page](/flags/) for global options not listed here.
--- a/docs/content/commands/rclone_sha1sum.md
+++ b/docs/content/commands/rclone_sha1sum.md
@@ -15,6 +15,11 @@ Produces an sha1sum file for all the objects in the path.
 Produces an sha1sum file for all the objects in the path.  This
 is in the same format as the standard sha1sum tool produces.
 
+By default, the hash is requested from the remote.  If SHA-1 is
+not supported by the remote, no hash will be returned.  With the
+download flag, the file will be downloaded from the remote and
+hashed locally enabling SHA-1 for any remote.
+
 
 ```
 rclone sha1sum remote:path [flags]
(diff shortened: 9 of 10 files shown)


ivandeex:
Fixes https://github.com/rclone/rclone/issues/3923

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
-		sum = "UNSUPPORTED"
-	} else if err != nil {
-		fs.Debugf(o, "Failed to read %v: %v", ht, err)
-		sum = "ERROR"
+
+	// If downloadFlag is true, download and hash the file.
+	// If downloadFlag is false, call o.Hash asking the backend for the hash
+	if downloadFlag {
+		
+		// Setup accounting as NewTransfer.  Same approach used in Cat
```
lostheli, line 857: Flagging questions as I realized this has the rough comments I wrote when coding it and not the slightly refined versions I posted to the forum.

Here: confirm account usage is correct

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+		tr := accounting.Stats(ctx).NewTransfer(o)
+		defer func() {
+			tr.Done(ctx, err)
+		}()
+
+		// Setup download options.  Similar approach to Cat.
+		// We do not have a start / end, so we only use fs.Config.DownloadHeaders.
+		// Unsure if copying to a new OpenOption object is required.
+		// Copy also has a very similar setup but includes hashOption
```
lostheli, line 866: Confirm what DownloadHeaders are required

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+		var options []fs.OpenOption
+		for _, option := range fs.Config.DownloadHeaders {
+			options = append(options, option)
+		}
+
+		// Using NewReOpen.  Error handling seems to be done different everwhere.  This version is from Cat
```
lostheli, line 872: Confirm usage of NewReOpen and how error handling should be done for this

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+			err = fs.CountError(err)
+			fs.Errorf(o, "Failed to open: %v", err)
+			return "ERROR", err
+		}
+		in = tr.Account(ctx, in).WithBuffer() // account and buffer the transfer
+
+		// Unclear if we need to close in.
+		// Copy, which uses NewReOpen does.
+		// Cat, rcat, and maybe some other methods use their own readCloser but don't appear to close
+		// If the error does not occur until close, we log the error but return the hash as okay
```
lostheli, line 884: Confirm if the input stream should be closed here

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+			err = in.Close()
+			if err != nil {
+				err = fs.CountError(err)
+				fs.Errorf(o, "Failed to close: %v", err)
+			}
+		}()
+
+		// Transfer is now setup.  Create a multihasher to receive it
+		// Should we instead use hash.StreamTypes here instead of creating the hasher and copying?
+		// We don't fs.CountError or fs.Errorf here because other functions that creat MultiHashers do not.  Should we?
```
lostheli, line 895: Confirm direct use of multihasher or use of hash.StreamTypes.
Confirm error handling

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+		if err != nil {
+			err = fs.CountError(err)
+			fs.Errorf(o, "Failed to copy file to hasher: %v", err)
+			return "ERROR", err
+		}
+
+		// hasher.Sum returns an error, but I don't believe it's possible to error here.
+		// It appears the only error is unsupported.
+		// If NewMultiHasherTypes succeeds and ht has not changed, we should not receive this error.
+		// Sum returns a byte array, and we want to return a hex string, so encode
```
lostheli, line 913: Confirm if code style is to always catch returned errors even for unreachable states.  Alternatively this should be byteSum, _ :=

lostheli:
Ivan, thank you for flagging the related open issues.  When I was originally trying to accomplish this with rclone, I looked at the forum and didn't anything specific.  I'm both surprised and not surprised this has come up multiple times before!

All, I've added comments where review is most needed.  I realized the pull request version doesn't have the cleaned up comments I posted to the forum.  Comments will be revised when finalizing the code anyway.

ivandeex:
Re build issue: `file not goimports-ed` is just a bad wording for `go source code formatting needs improvement`.
Run
```
gofmt -w ./cmd/dbhashsum/dbhashsum.go
gofmt -w ./fs/operations/operations.go
```
and commit the result. It will change some indenting or sort imports by name or smth like that.

> fs/operations/operations.go:856: File is not `goimports`-ed (goimports)
> cmd/dbhashsum/dbhashsum.go:9: File is not `goimports`-ed (goimports)
> "github.com/rclone/rclone/fs/config/flags"
Makefile:86: recipe for target 'check' failed
make: *** [check] Error 1

ivandeex:
An example of how this very flag [was](https://github.com/rclone/rclone/commit/541929258b81a078dbad56843e27119315eb7ca0#diff-278907099aff13e9697bfbe47a7a8bad3fcdddeebb518e10caaa9877e1840a70R839) added to `check` in 2017 and how it [looks](https://github.com/rclone/rclone/blob/v1.53.3/fs/operations/check.go#L303) now.

cmd/dbhashsum/dbhashsum.go
```diff
@@ -6,13 +6,22 @@ import (
...
 )
 
+var (
+	outputBase64 = false
+	downloadFlag = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
```
ncw, line 22: Gah! We depracated dbhashsum some time ago - perhaps it is time to kill it now, rather than updating it!

cmd/sha1sum/sha1sum.go
```diff
@@ -5,12 +5,22 @@ import (
...
 
+var (
+	outputBase64 = false
+	downloadFlag = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	flags.BoolVarP(cmdFlags, &outputBase64, "base64", "", outputBase64, "Output base64 encoded hashsum")
```
ncw, line 22: It would be nice to factor the flags...

I'd move all the flags to `hashsum` and have a public `AddFlags` function which md5sum/sha1sum etc call.

You can either make the flags global variables and refer to them in md5sum/sha1sum like that or you can make a wrapper for this `operations.HashLister(context.Background(), hash.SHA1, outputBase64, downloadFlag, fsrc, os.Stdout)`

fs/operations/operations.go
ncw: Yes `NewTransfer` is correct

fs/operations/operations.go
ncw: Adding `fs.Config.DownloadHeaders` is correct.

It looks a bit untidy though - maybe something needs factoring...

You don't need to copy the DownloadHeaders - you can just use them directly in NewReOpen

fs/operations/operations.go
ncw: `tr.Done` closes the file and accounts errors, so no it doesn't need to be closed here

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+			syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
+		} else {
+			syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())	
 		}
-		width := base64.URLEncoding.EncodedLen(hash.Width(ht) / 2)
-		syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
 	})
 }
 
+// HashListerBase64 does an md5sum equivalent for the hash type passed in with base64 encoded
```
ncw, line 958: Can you remove the dead code please?

fs/operations/operations.go
ncw: When you get an error you should run fs.CountError() on it and log it at some level, probably ERROR

Normally rclone wouldn't do this if you are returning the error, but the callers of hashSum don't log the error or return it to their callers so it is necessary.

I think the error handling here needs a re-write, I would not call `fs.CountError()` or `fs.Errorf` but I would return the error wrapped `return "ERROR", errors.Wrap(err, "failed to open")`

Then the caller of `hashLister` should log these errors with `fs.CountError()` and `fs.Errorf` and return the first one as its own ERROR.

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
+		// We do not have a start / end, so we only use fs.Config.DownloadHeaders.
+		// Unsure if copying to a new OpenOption object is required.
+		// Copy also has a very similar setup but includes hashOption
+		var options []fs.OpenOption
+		for _, option := range fs.Config.DownloadHeaders {
+			options = append(options, option)
+		}
+
+		// Using NewReOpen.  Error handling seems to be done different everwhere.  This version is from Cat
+		in, err := NewReOpen(ctx, o, fs.Config.LowLevelRetries, options...)
```
ncw, line 873: Note that if you rebase on master you'll need to write `fs.Config.LowLevelRetries` as `fs.GetConfig(ctx).LowLevelRetries`

fs/operations/operations.go
ncw: What you've written looks fine

fs/operations/operations.go
ncw: I usually write what you've written.

ncw (commented):
Looking good :-)

Please see inline for a first pass review!

Hopefully I've answered your questions.

Can you remove the dead code and questions from the code to make it easier to read.

You'll have a few small fix-ups if you rebase on master - I suggest you do that now.

Thank you

lostheli:
Thanks for the review!  I have a few follow up questions, and there are a few still pending.

As I move to the parallel part of this, I wanted to better understand the suggestion to implement by making the anonymous function in HashLister parallel.  If I understand correctly,
- ListFn will wrap this with a .ForObject call and then pass to walk.ListR
- In the simple case, walk.ListR will pass to to walk.listRwalk
- Walk will pass to one of several functions, but let's assume walkListDirSorted
- walkListDirSorted passes to walk
- walk finally calls the function, but does so within a mutex

Somewhere fn gets wrapped to take a remote and error variable instead of just the fs.DirEntries it takes in ListFn.

This level of function passing is tough for me to follow - a bit mind numbing actually.  As far as I can tell, the parallel version of HashLister could be called but only once at a time.  Inside that we'd be able to process a number of files in parallel based on the number of transfers available.  However, the whole process seems very dependent on how all of these layers of walk construct the list of files to act on.

For example, what would happen if we called this on directory with 4 sub-directories and each sub-directory contained a single file.  Would walk flatten the 4 files into a single fs.DirEntries and then hand that to our new HashLister?  If not, why does the second file (in a separate invocation of HashLister) not get blocked by the mutex in walk?

And finally, a broader question.  Are you just using the channel + WaitGroup as a semaphore?  Are you avoiding semaphore because go put it in a weird non-standard package?

ncw:
> Thanks for the review! I have a few follow up questions, and there are a few still pending.

Pending that I haven't answered?

> As I move to the parallel part of this, I wanted to better understand the suggestion to implement by making the anonymous function in HashLister parallel. If I understand correctly,
> 
> * ListFn will wrap this with a .ForObject call and then pass to walk.ListR
> * In the simple case, walk.ListR will pass to to walk.listRwalk
> * Walk will pass to one of several functions, but let's assume walkListDirSorted
> * walkListDirSorted passes to walk
> * walk finally calls the function, but does so within a mutex
> 
> Somewhere fn gets wrapped to take a remote and error variable instead of just the fs.DirEntries it takes in ListFn.
> 
> This level of function passing is tough for me to follow - a bit mind numbing actually. As far as I can tell, the parallel version of HashLister could be called but only once at a time. Inside that we'd be able to process a number of files in parallel based on the number of transfers available. However, the whole process seems very dependent on how all of these layers of walk construct the list of files to act on.
> 
> For example, what would happen if we called this on directory with 4 sub-directories and each sub-directory contained a single file. Would walk flatten the 4 files into a single fs.DirEntries and then hand that to our new HashLister? If not, why does the second file (in a separate invocation of HashLister) not get blocked by the mutex in walk?

It depends is the answer!

If you aren't using ListR then you'll get --checkers directory threads reading directories and they will call hashlister concurrently.

However you don't need to worry about that as you are going to start a go routine to do the processing in the background and return immediately. So you'll be making your own parallelism. It therefore doesn't matter whether the listing comes in sequentially or in parallel as long as it comes in faster than you can download files (which it will).

> And finally, a broader question. Are you just using the channel + WaitGroup as a semaphore? Are you avoiding semaphore because go put it in a weird non-standard package?

Ah you mean: https://godoc.org/golang.org/x/sync/semaphore

To tell you the truth I didn't know about that package! The golang.org/x packages are the Go team's place to put packages that might be considered for the standard library. Or maybe now-a-days they can be considered as standard library part 2. I quite often use golang.org/x/sync/errgroup

I've been writing Go for a long time now (since before Go 1.0) so I guess I just got used to the using the primitives Go provides out of which you can build anything at all.

Here was my suggestion from before

```go
func HashLister(ctx context.Context, ht hash.Type, f fs.Fs, w io.Writer) error {
	concurrencyControl := make(chan struct{}, fs.Config.Transfers)
	var wg sync.WaitGroup
	err := ListFn(ctx, f, func(o fs.Object) {
		concurrencyControl <- struct{}{} // get a concurrency slot
		wg.Add(1)
		go func() {
			defer func() {
				wg.Done()
				<-concurrencyControl // return the concurrency slot
			}()
			sum, _ := hashSum(ctx, ht, o)
			syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
		}()
	})
	wg.Wait()
	return err
}
```

Lets see what it looks like with semaphore

```go
func HashLister2(ctx context.Context, ht hash.Type, f fs.Fs, w io.Writer) error {
	maxWorkers := int64(fs.Config.Transfers)
	concurrencyControl := semaphore.NewWeighted(maxWorkers)
	err := ListFn(ctx, f, func(o fs.Object) {
		if err := sem.Acquire(ctx, 1); err != nil {
			fs.CountError(err)
			fs.Errorf(o, "Failed to acquire semaphore: %v", err)
			return
		}
		go func() {
			defer concurrencyControl.Release(1)
			sum, _ := hashSum(ctx, ht, o)
			syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
		}()
	})
	if semErr := sem.Acquire(ctx, maxWorkers); semErr != nil {
		fs.CountError(semErr)
		fs.Errorf(o, "Failed to acquire semaphore: %v", semErr)
	}
	return err
}
```

3 lines longer and some rather ugly error handling! It does do one
thing that the original code does though and that is exit if the
context is cancelled and that might be valuable. Though I think if you
did cancel the context you'll get one error per ListFn item which
might be quite a lot before the context cancel comes through there...

Looking at the semaphore code and docs a bit more the only way Acquire it can fail is if the context is cancelled, so it could be simplified to this

```go
func HashLister3(ctx context.Context, ht hash.Type, f fs.Fs, w io.Writer) error {
	maxWorkers := int64(fs.Config.Transfers)
	concurrencyControl := semaphore.NewWeighted(maxWorkers)
	err := ListFn(ctx, f, func(o fs.Object) {
		if sem.Acquire(ctx, 1) == nil {
			go func() {
				defer concurrencyControl.Release(1)
				sum, _ := hashSum(ctx, ht, o)
				syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
			}()
		}
	})
	_ = sem.Acquire(ctx, maxWorkers)
	return err
}
```

If you want do so it like that, then that is fine with me :-)

(All code untested!)

lostheli:
> Pending that I haven't answered?

Yes, though that's my fault.  I did not click the overall response submit and my in-line comments were stuck in edit mode.

> > For example, what would happen if we called this on directory with 4 sub-directories and each sub-directory contained a single file. Would walk flatten the 4 files into a single fs.DirEntries and then hand that to our new HashLister? If not, why does the second file (in a separate invocation of HashLister) not get blocked by the mutex in walk?
> 
> It depends is the answer!
> 
> If you aren't using ListR then you'll get --checkers directory threads reading directories and they will call hashlister concurrently.
>
> However you don't need to worry about that as you are going to start a go routine to do the processing in the background and return immediately. So you'll be making your own parallelism. It therefore doesn't matter whether the listing comes in sequentially or in parallel as long as it comes in faster than you can download files (which it will).

Okay, I believe in the process of trying to rationalize and explain the behavior here I understood a bit of how it works.  I**s this correct?**
- For the sake of writing this out, I'm going to refer to the anonymous function written in HashLister and handed to ListFn as hashSumCaller.  We'll assume a version of HashLister like the above where hashSumCaller starts a go routine with its own inner function that calls hashSum.
- ListFn wraps hashSumCaller in .ForObject making it sequential across all of the objects in a single entries list (more on those later).  We  'escape' this serial nature by having hashSumCaller simply kick off the go routine instead of calling hashSum directly. This causes hashSumCaller to return immediately once the go routine **starts** instead of waiting for hashSum.
- Assuming  transfers = 4 and an entries list of 5 or more, the first 4 invocations of hashSumCaller by .ForObject will kick off their go routine and return immediately. .ForObject will **block** on each object starting with number 5 until that call to hashSumCaller gets a slot, starts its go routine to call hashSum, and returns.  This continues until .ForObject calls hashSumCaller for the last object in entries and hashSumCaller returns.  At this point, .ForObject has **started** the hashing process for all objectives in entities but hashing is **ongoing** for roughly the last 4 objects in entities.
- The anonymous go routine in walk calls this .ForObject loop of hashSumCaller invocations on the contents of each directory.  While there are multiple 'checker' threads constructing the directory lists and calling the function, only **one** checker thread can actually call the function at a time because of the mutex.  Checker #1 will block checker #2 until .ForObject returns, which happens when the last hash for checker #1's directory is **started**.  This has minimal impact on us as the checker thread is freed as soon as the last hash **starts**.  While Checker #1 is blocked by .ForObject and .ForObject by hashSumCaller and hashSumCaller by the available slots, it doesn't matter where Checker #2 is blocked.

What was weird to me, is wg.Wait() is actually **5 or so levels removed** on the call stack from the hashSumCaller code on the lines immediately above it.  As a result of all the function passing, hashSumCaller gets nested many levels down.  wg.Wait() by contrast is sitting just one level under the command.  wg.Wait() does not get **called** until ListFn returns, which happens after the walk stack returns, which happens (skipping a few levels of function calls in walk) after the checker threads finish, which happens after they have completed the .ForObject loop on the last batch of entities, which happens after the last hashes are **started**.  wg.Wait() then simply blocks while those final 4 hashes finish.

> > And finally, a broader question. Are you just using the channel + WaitGroup as a semaphore? Are you avoiding semaphore because go put it in a weird non-standard package?
> 
> Ah you mean: https://godoc.org/golang.org/x/sync/semaphore
> 
> To tell you the truth I didn't know about that package! The golang.org/x packages are the Go team's place to put packages that might be considered for the standard library. Or maybe now-a-days they can be considered as standard library part 2. I quite often use golang.org/x/sync/errgroup
> 
> I've been writing Go for a long time now (since before Go 1.0) so I guess I just got used to the using the primitives Go provides out of which you can build anything at all.

I'm coming at this from the opposite side!  I want a semaphore, so I search and see what they called semaphores in Go.

ivandeex:
The `worker pool` pattern (https://gobyexample.com/worker-pools) is one of the first language examples I did when I learned go. Check it out online using the link above. It's used widely in rclone, e.g. see https://github.com/rclone/rclone/blob/master/backend/union/union.go#L868
Semaphores are rare mostly for historic reasons, though I doubt you'll see any performance fluctuations.

lostheli:
> The `worker pool` pattern (https://gobyexample.com/worker-pools) is one of the first language examples I did when I learned go. Check it out online using the link above. It's used widely in rclone, e.g. see https://github.com/rclone/rclone/blob/master/backend/union/union.go#L868
> Semaphores are rare mostly for historic reasons, though I doubt you'll see any performance fluctuations.

Indeed, and walk follows that pattern.  It sets up a pool of workers, hands them work through a channel, and then exits when all of the work is done.  However, the total construct proposed here does not follow that pattern - at least not if I now understand it correctly.  We **do not** set up a pool of hashSum workers and feed them work through a channel.  The channel has empty concurrency tokens.

hashSum is instead 'fed' work by calling it repeatedly after handing the function (or the anonymous go routine that calls it) down the call stack.  However, the concurrency control for this repeated calling has to be provided at the top of the call stack (by the proposed concurrencyControl channel or semaphore, which I left out in my last post).  The entire worker pool pattern of walk takes place **between** the level of concurrencyControl and the level of hashSumCaller!

Coming from outside go and this project, that's what got me.  Not using a worker pool / channel pattern to give work to hashSum.  Rather nesting the work pool / channel pattern of walk between the layers of another semaphore-like pattern of repeated function calls to hashSum.  Even if I roughly understand how this works, I'm still not to the point where I can really reason about it's implications.

But this all rests on - is my understanding of how this would work correct?

ivandeex:
1. There obviously will be two working pools here.
One is already written, it walks lists.
Another will be written by you, it will download and hash data, if --download is supplied. The former will feed jobs to latter e.g. via channel.
2. What do you call _nesting_? The sketch below is not nesting. Thanks to the `go` word it creates a completely parallel thread of execution and continues immediately.
```
if flag {
  go func() {
    actions
  }()
}
```
3. Forget about semaphores. They give no advantage vs `wg`, only distract from fulfilling the goal. If you wish, use interactive go console to play with them in a toy program.
4. Please fix `goimportsed` :)

Ivan

lostheli:
>     1. There obviously will be two working pools here.
>        One is already written, it walks lists.
>        Another will be written by you, it will download and hash data, if --download is supplied. The former will feed jobs to latter e.g. via channel.

Okay, I'll be picky on language for a minute, and I apologize.  E.g. as in 'for example this is one way it could work?'   Or that is how it **would** work if I implemented it the way ncw suggested?  Because that might be where I am stuck.

I agree we could hand work between the directory walking worker pool and a hashing worker pool with a channel.  However, if we implement it the way that's suggested, I do not see where we setup a pool of hashing workers or what channel is used to hand work to them.  'Setup a pool of workers' is a little vague but the channel bit is pretty clear cut.  The object (file) to be hashed is never placed into a channel to relay it to hashSum unless, potentially, go implicitly uses a channel to pass arguments to new go routines.  It is instead handed to hashSum through a function call.  It's a little harder to illustrate the lack of a worker pool, but maybe I can say this - there does not appear to be any reuse of hashSum workers.  A new go routine is created for each object subject to a cap on how many can concurrently exist.  If we were creating a worker pool and then feeding it work, I would expect to see those workers reused (as they are in walk).

>     2. What do you call _nesting_? The sketch below is not nesting. Thanks to the `go` word it creates a completely parallel thread of execution and continues immediately.
> 
> 
> ```
> if flag {
>   go func() {
>     actions
>   }()
> }
> ```

My use of call stack here is almost certainly not technically correct.  In your example I would say the creation of the go routine is nested between the action and whatever called this function with if flag.  I don't actually know how go handles stacks for go routines.  

In our case:
1. Command calls HashLister
2. HashLister calls ListFn; in the proposal it also provides concurrency control for hashSum
3. HashLister calls walk.ListR
4. ListR calls listRwalk (as one option depending on what the remote supports)
5. listRwalk calls Walk
6. Walk calls walkListDirSorted
7. walkListDirSorted calls walk
8. walk creates a worker pool of go routines to walk the directory structure and sets them running
9. the go routines put directories into the job queue but also call a function defined in ListFn with a list of entities - and this is where files get processed
10. that function calls entities.ForObject with the function defined in HashLister (which I called hashSumCaller in the post above)
11. hashSumCaller starts another go routine when the concurrency control in number 2 allows
12. this go routine calls hashSum

So, we probably don't have a single stack that contains all of these calls because of the go routines, but this is the order the functions are called in.  The worker pool + channel concurrency pattern of walk exists in steps 8 and 9.  The concurrency pattern used to hash four files at a time is defined by steps 2, 11, and 12.  That's what I mean by nested - steps 8 and 9 are between 2, 11, and 12.

>     3. Forget about semaphores. They give no advantage vs `wg`, only distract from fulfilling the goal. If you wish, use interactive go console to play with them in a toy program.

Sorry, by semaphore-like I just mean filling the role of a semaphore.  The pattern of using a channel with tokens and a wait group to limit the number of concurrent function calls appears to implement a semaphore.

>     4. Please fix `goimportsed` :)

Of course


Thank you for your patience!

cmd/dbhashsum/dbhashsum.go
ivandeex: We'd better drop `dbhashsum` in a dedicated patch for transparency.
Right now let's add the line `dbhashsum will be removed from rclone` in the docs.

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
 // excludes
 //
 // Lists in parallel which may get them out of order
-func Md5sum(ctx context.Context, f fs.Fs, w io.Writer) error {
-	return HashLister(ctx, hash.MD5, f, w)
-}
+// func Md5sum(ctx context.Context, f fs.Fs, w io.Writer) error {
+// 	return HashLister(ctx, hash.MD5, false, false, f, w)
+// }
```
ivandeex, line 835: Please remove completely, don't comment out.

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
 //
 // Obeys includes and excludes
 //
 // Lists in parallel which may get them out of order
-func Sha1sum(ctx context.Context, f fs.Fs, w io.Writer) error {
-	return HashLister(ctx, hash.SHA1, f, w)
-}
+// func Sha1sum(ctx context.Context, f fs.Fs, w io.Writer) error {
+// 	return HashLister(ctx, hash.SHA1, false, false, f, w)
+// }
```
ivandeex, line 844: Also. Remove completely.

cmd/sha1sum/sha1sum.go
ivandeex: Normally if code passes Nick's review, it works.

fs/operations/operations.go
ivandeex: Not removed yet.

fs/operations/operations.go
ivandeex: Yes. It's 1-2 days ago. You'll have to rebase again.

cmd/hashsum/hashsum.go
```diff
@@ -15,12 +15,14 @@ import (
...
 var (
 	outputBase64 = false
+	downloadFlag = false
 )
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
 	cmdFlags := commandDefinition.Flags()
 	flags.BoolVarP(cmdFlags, &outputBase64, "base64", "", outputBase64, "Output base64 encoded hashsum")
+	flags.BoolVarP(cmdFlags, &downloadFlag, "download", "", downloadFlag, "Download files from remote and hash locally instead of using hash from remote")
```
ivandeex, line 25: Please factor 4 lines inside init() into new `AddCommand()` and call it from `init()`.

cmd/md5sum/md5sum.go
```diff
@@ -5,12 +5,22 @@ import (
...
 
+var (
+	outputBase64 = false
+	downloadFlag = false
+)
+
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
+	cmdFlags := commandDefinition.Flags()
+	flags.BoolVarP(cmdFlags, &outputBase64, "base64", "", outputBase64, "Output base64 encoded hashsum")
```
ivandeex, line 22: Please replace by
```
init() {
  hashsum.AddCommand()
}
```

cmd/md5sum/md5sum.go
```diff
@@ -5,12 +5,22 @@ import (
...
+	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	outputBase64 = false
+	downloadFlag = false
+)
```
ivandeex, line 17: Don't add new flags. Let's reuse flags from `hashsum`.

fs/operations/operations.go
ivandeex: Yes, please use directly and remove "unsure" comments.
Also please replace `fs.Config` by `fs.GetConfig(ctx)`
as you are already doing elsewhere.

ivandeex (changes requested):

fs/operations/operations.go
ivandeex: > For example, I see copy doing CountError and Errorf in a similar use case.

Because it hasn't been tidyed. It might be if we return to that code later. You touch your part, you look at it, you decide tidying is good. You tidy. You don't touch `Copy`. It stays as is. Step by step.

> If the caller of hashLister needs to log the errors, does this mean we should log them all the way back in the commands?

No. Why log repeatedly?

fs/operations/operations.go
ivandeex: Also, please remove "i don't beleave... it appears..." comments

ivandeex:
> Or that is how it **would** work if I implemented it the way ncw suggested? Because that might be where I stuck.

Concurrency is tricky. Let's forget about working pools, semaphore patterns, channels, locking and all that black locking magic (BLM!) for a moment. It's all optimizations. It's driving FAST. We'll get back to ir later.

Right now please start from addressing current review comments. Make your code compile.

You've added `downloadFlag` argument to `hashSum`, right? Now make the new `hashSum` pass tests.  Let's drive slow at stay alive.

After (only after) that we can return to BLM.

fs/operations/operations.go
```diff
@@ -830,59 +830,144 @@ func ListLong(ctx context.Context, f fs.Fs, w io.Writer) error {
...
-		sum, err := hashSum(ctx, ht, o)
-		if err == nil {
+		sum, err := hashSum(ctx, ht, downloadFlag, o)
+		if outputBase64 && err == nil {
 			hexBytes, _ := hex.DecodeString(sum)
 			sum = base64.URLEncoding.EncodeToString(hexBytes)
+			width := base64.URLEncoding.EncodedLen(hash.Width(ht) / 2)
+			syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
+		} else {
+			syncFprintf(w, "%*s  %s\n", hash.Width(ht), sum, o.Remote())
```
ivandeex, line 953: "git show" shows trailing spaces on this line, please clean

ivandeex (changes requested):

ivandeex:
Rebased on master. Please save your changes and pull to your workdir.

ivandeex:
@lostheli 
At this stage please finish refactoring and close review comments after pushing commits that resolve them.
Postpone thread pools to the next round of reviews.
You may even not implement the real download now. It's enough for the moment that `hashSum` simply logs `Errorf("Pretending to download")` when it sees `downloadFlag`. We'll look into it later.
Thanks

lostheli (commented):
I've committed an updated version of the code.

Apologies if there was some misunderstanding here Ivan - you reviewed the original code thus no changes.  I marked questions and comments as resolved where the next steps were clear.  The updates I've made are a mix of the different suggestions and there are a few outstanding questions.  Notably, I already implemented the multi-threaded version while trying to understand the suggestion, so that's what's here.

When you say I may not implement non-threaded downloads and suggest logging an error for downloadFlag, I do not follow at all.  To be clear, the code works with some caveats I'm aware of and of course probably some I'm not.

Questions:
- My question on errors is - where should the error be logged with fs.CountError and fs.Errorf.  Nick said in the caller of HashLister, which is the command.  That seemed an odd place for it, partially because of what I see elsewhere, so wanted to confirm if I understood correctly.  If Copy hasn't been cleaned up recently and doesn't match the current coding style preference, I don't know that.  I based my code as much as possible on these existing functions.
- There are a couple of additional questions on the code I either forgot or couldn't figure out that I've flagged this time.
- I'm unsure how to update the documentation.  CONTRIBUTING.md suggests manual edits to md files in docs/content.  The files in the sub-folder docs/content/commands that we'd update here say they are autogenerated and to run make commanddocs.  When I run make commanddocs, I get the same docs as before and they don't contain the new flags (even the doc for hashsum where those flags are added directly).  I wanted to test this because I'm concerned how the new flags will reflect in the documentation for the other commands (e.g. sha1sum).  For now, I've edited them manually hoping these changes carry over when regenerating.

Caveats I know of:
- When using this with --progress, the output with the hash was suppressed.  Once per file, the line 'Transferred: xG / y GBytes, z%, a MBytes/s, ETA 1s' is reprinted rather than updated in place.  Suspect the way progress resets this line is not playing nice with having other output.  It is overwriting the hashsum output rather than the Transferred line.  All I thought of to resolve was move the output to the info logging system, which appears designed to work with progress but is not idea, or implement output to a file, which requires more work and is maybe a future task.  If there's an easy solution that would be great.
- The restart functionality for failed transfers may not be working correctly.  In running this unattended, I had 4 files fail with an IO timeout and then the next 100 or so files fail with a TCP error 'no such host.'  I suspect the problem was caused by a network hiccup not rclone itself, and I don't know how long any network interruption was.  In rerunning on the same files, the issue was not reproduced.  Now, rclone did produce errors on a lot of files before the remote API started working again (maybe insufficient backoff) and did not 'go back' for those files (would a different error implementation would signal rclone to do that?).  Because it's not repeatable, it's difficult to verify the retry / restart functionality.  Are there test cases designed to test the behavior when transfers fail?
- It seems to ignore --bwlimit.  I've not looked into this one at all but noticed it when thinking of things to help avoid IO failures.

cmd/sha1sum/sha1sum.go
lostheli: As I understood it, there were 2 possible approaches.  Option 1, public flags.  Option 2, a wrapper.  I do not understand how to implement option 2.  I have implemented Option 1 in a manner largely consistent with check / cryptcheck.

cmd/dbhashsum/dbhashsum.go
lostheli: dbhashsum is retained for now.

docs/content/commands/rclone_hashsum.md
```diff
@@ -37,13 +39,16 @@ rclone hashsum <hash> remote:path [flags]
...
+      --download	Download the file and hash it locally; if this flag is not specified, the hash is requested from the remote
+  -h, --help		Help for hashsum
 ```
 
 See the [global flags page](/flags/) for global options not listed here.
 
 ## SEE ALSO
 
 * [rclone](/commands/rclone/)	 - Show help for rclone commands, flags and backends.
+* [rclone md5sum](/commands/rclone_md5sum/)	 - Shortcut equivalent to 'hashsum MD5'
```
ncw, line 52: Did you edit this by hand? It is an auto generated file, generated by `make commanddocs`.

Or maybe you figured out some clever way of getting see also entries in with pflag - I don't see anything like that though.

docs/content/commands/rclone_hashsum.md
```diff
@@ -16,6 +16,8 @@ Produces a hash file for all the objects in the path using the hash
 named.  The output is in the same format as the standard
 md5sum/sha1sum tool.
 
+By default, the hash is requested from the remote.  If the hash is not supported, no hash will be returned.
```
ncw, line 19: This text needs to go in the help for `rclone hashsum`

I run `make backenddocs` as part of the release, so I prefer commits without patches to the auto generated docs.

Anything in docs/contents/commands/ is auto generated

ncw (commented):
> I've committed an updated version of the code.

Thank you. I've put a few comments inline.

I think the code is looking very good.

The multithreaded code is very in-style for rclone so is easy for me to read.
 
> Questions:
> 
> * My question on errors is - where should the error be logged with fs.CountError and fs.Errorf.  Nick said in the caller of HashLister, which is the command.

If you call `fs.CountError` then rclone will return a non zero error code, even if the error didn't get bubbled up.

Calling `fs.Errorf` shows the error to the user.

Generally the convention is in rclone is to **either** return the error, **or* call `fs.Errorf` and `fs.CountError`. If you do both then you run the risk of double logging of errors.

So that is the reason for the convention and why I suggested that we remove the logging there since the error is returned and make sure we make sure it is done in the caller.

>  That seemed an odd place for it, partially because of what I see elsewhere, so wanted to confirm if I understood correctly.  If Copy hasn't been cleaned up recently and doesn't match the current coding style preference, I don't know that.  I based my code as much as possible on these existing functions.

`Copy` is the oldest rclone function and also the most worked upon function so I would be surprised if there were inconsistencies.

> * There are a couple of additional questions on the code I either forgot or couldn't figure out that I've flagged this time.
> * I'm unsure how to update the documentation.  CONTRIBUTING.md suggests manual edits to md files in docs/content.  The files in the sub-folder docs/content/commands that we'd update here say they are autogenerated and to run make commanddocs.  When I run make commanddocs, I get the same docs as before and they don't contain the new flags (even the doc for hashsum where those flags are added directly).  I wanted to test this because I'm concerned how the new flags will reflect in the documentation for the other commands (e.g. sha1sum).  For now, I've edited them manually hoping these changes carry over when regenerating.

Hmm, `make commanddocs` should work. It might be that you have to run `make` first. Changes to `docs/content/commands/*` will get overwritten, alas.

> Caveats I know of:
> 
> * When using this with --progress, the output with the hash was suppressed.  Once per file, the line 'Transferred: xG / y GBytes, z%, a MBytes/s, ETA 1s' is reprinted rather than updated in place.  Suspect the way progress resets this line is not playing nice with having other output.  It is overwriting the hashsum output rather than the Transferred line.  All I thought of to resolve was move the output to the info logging system, which appears designed to work with progress but is not idea, or implement output to a file, which requires more work and is maybe a future task.  If there's an easy solution that would be great.

I don't think this is easily fixable. the hashsum etc write to stdout whereas progress writes to stderr. If users want to use hashsum with progress then they should redirect stdout to a file.

> * The restart functionality for failed transfers may not be working correctly.  In running this unattended, I had 4 files fail with an IO timeout and then the next 100 or so files fail with a TCP error 'no such host.'  I suspect the problem was caused by a network hiccup not rclone itself, and I don't know how long any network interruption was.  In rerunning on the same files, the issue was not reproduced.  Now, rclone did produce errors on a lot of files before the remote API started working again (maybe insufficient backoff) and did not 'go back' for those files (would a different error implementation would signal rclone to do that?).  Because it's not repeatable, it's difficult to verify the retry / restart functionality.  Are there test cases designed to test the behavior when transfers fail?

There are lots of test of the NewReOpen functionality. If you run with `-vv` then you will see the retries. However if a non-retriable error happened then rclone won't retry. `no such host` may be such an error, I'm not sure.

> * It seems to ignore --bwlimit.  I've not looked into this one at all but noticed it when thinking of things to help avoid IO failures.

Hmm we need to fix that! This line

    tr := accounting.Stats(ctx).NewTransfer(o)

and this one

    in = tr.Account(ctx, in).WithBuffer()

Should be making sure we are limiting the transfer rates. You know the parameter to --bwlimit is in kBytes/s or MByte/s?

lostheli:
> If you call `fs.CountError` then rclone will return a non zero error code, even if the error didn't get bubbled up.
> 
> Calling `fs.Errorf` shows the error to the user.
> 
> Generally the convention is in rclone is to **either** return the error, *_or_ call `fs.Errorf` and `fs.CountError`. If you do both then you run the risk of double logging of errors.
> 
> So that is the reason for the convention and why I suggested that we remove the logging there since the error is returned and make sure we make sure it is done in the caller.

I was looking without luck for commands that do error logging in the command with the operation result  - for example the existing version of hashsum only handles errors with inputs.  Eventually I realized there is some general error handling in cmd.Run for returned errors.  Is that the right path to go down then?  Wrap and return the errors then cmd.Run will take care of the logging?  How does that play out timing wise?  If I am running a 12hr job, I may want to see when it starts having issues instead of waiting for the end.

> Hmm, `make commanddocs` should work. It might be that you have to run `make` first. Changes to `docs/content/commands/*` will get overwritten, alas.

I'll take another look at make commanddocs to see if I can determine why it was not working.  I did try running make and make rclone.  Maybe I didn't get the order right.  Once working, I can move the documentation into the function.

> I don't think this is easily fixable. the hashsum etc write to stdout whereas progress writes to stderr. If users want to use hashsum with progress then they should redirect stdout to a file.

What you propose would work for my use case and likely others.  Unfortunately it doesn't seem to work that way on my system; progress appears to be on stdout not stderr.  Is there a flag or setting to get it on stderr?  If so, we can suggest this to users in the documentation.

> Hmm we need to fix that! This line
> 
> ```
> tr := accounting.Stats(ctx).NewTransfer(o)
> ```
> 
> and this one
> 
> ```
> in = tr.Account(ctx, in).WithBuffer()
> ```
> 
> Should be making sure we are limiting the transfer rates. You know the parameter to --bwlimit is in kBytes/s or MByte/s?

Looked into this one further.  Running it locally, the bwlimit has an impact.  Since I wasn't using progress, I was just looking at total process usage - it's possible the difference was all overhead.  Where is bwlimit applied?  Is it final file io or incoming packets or somewhere in between?

lostheli:
> > Hmm, `make commanddocs` should work. It might be that you have to run `make` first. Changes to `docs/content/commands/*` will get overwritten, alas.
> 
> I'll take another look at make commanddocs to see if I can determine why it was not working. I did try running make and make rclone. Maybe I didn't get the order right. Once working, I can move the documentation into the function.

I solved this mystery.  make commanddocs runs 'rclone gendocs docs/content/'.  For me, that's a call to the stable install in /usr/bin/rclone not a call to the development version it just built.  Calling gendocs manually on the new build has the expected result.  Automatically generating the documents will reflect in the next version.

Other questions remain.

ncw:
> > If you call `fs.CountError` then rclone will return a non zero error code, even if the error didn't get bubbled up.
> > Calling `fs.Errorf` shows the error to the user.
> > Generally the convention is in rclone is to **either** return the error, *_or_ call `fs.Errorf` and `fs.CountError`. If you do both then you run the risk of double logging of errors.
> > So that is the reason for the convention and why I suggested that we remove the logging there since the error is returned and make sure we make sure it is done in the caller.
> 
> I was looking without luck for commands that do error logging in the command with the operation result - for example the existing version of hashsum only handles errors with inputs. Eventually I realized there is some general error handling in cmd.Run for returned errors. Is that the right path to go down then? Wrap and return the errors then cmd.Run will take care of the logging? How does that play out timing wise? If I am running a 12hr job, I may want to see when it starts having issues instead of waiting for the end.

I think the place to do the `fs.Errorf` and the `fs.CountError` is in `HashLister` if we want that. Note that errors from HashLister currently aren't returned to the parent, however the hashlist line will have an ERROR in it which is an indication to the user that things have gone wrong.

> > I don't think this is easily fixable. the hashsum etc write to stdout whereas progress writes to stderr. If users want to use hashsum with progress then they should redirect stdout to a file.
> 
> What you propose would work for my use case and likely others. Unfortunately it doesn't seem to work that way on my system; progress appears to be on stdout not stderr. Is there a flag or setting to get it on stderr? If so, we can suggest this to users in the documentation.

Ah, I was wrong about that :-(

To get uncorrupted output it would be necessary to print each line with this function - that is how rclone gets logging and progress to play nice.

https://github.com/rclone/rclone/blob/50cf97fc729ee305cb89260f2c7ed387cf12bac0/cmd/progress.go#L73

I don't know whether it is worth doing this though, but it is possible with a bit of indirection!

> Looked into this one further. Running it locally, the bwlimit has an impact. Since I wasn't using progress, I was just looking at total process usage - it's possible the difference was all overhead. Where is bwlimit applied? Is it final file io or incoming packets or somewhere in between?

bwlimit is applied in the core so when reading stuff from input and writing it to output. There can often be buffering in the input and output which makes it not 100% accurate

> I solved this mystery. make commanddocs runs 'rclone gendocs docs/content/'. For me, that's a call to the stable install in /usr/bin/rclone not a call to the development version it just built. Calling gendocs manually on the new build has the expected result. Automatically generating the documents will reflect in the next version.

Great

Normally people put `~/go/src/bin` on the PATH so you can run `go install` and the binary just works.

lostheli:
New commit with changes to the error handling and output.  First try failed code quality for unnecessary use of else; hopefully second attempt will pass all tests!

> I think the place to do the `fs.Errorf` and the `fs.CountError` is in `HashLister` if we want that. Note that errors from HashLister currently aren't returned to the parent, however the hashlist line will have an ERROR in it which is an indication to the user that things have gone wrong.

Changed the error approach to wrap the errors in hashSum and then use fs.CountError and fs.Errorf in HashLister.  I believe it's important to have some errors as we go along in case the user has redirected stdout.

Note, I put CountError and Errorf in the anonymous function defined in HashLister not in the HashLister body itself (i.e. in the function that's run when walk's go routines call entities.ForObject).  Not sure if that was your intent.  Functionally, it doesn't seem to change much from doing it in hashSum, so I'm unsure if this avoids the double counting that raised concerns.  As you said, the anonymous function does not return the error any further, so it would require significant changes handle errors elsewhere.  The other goal in returning errors to the anonymous function appears to be avoiding the base64 encoding for the fake hashes 'ERROR' and 'UNSUPPORTED'.  Can you confirm if my changes are what you had in mind?

> Ah, I was wrong about that :-(
> 
> To get uncorrupted output it would be necessary to print each line with this function - that is how rclone gets logging and progress to play nice.
> 
> https://github.com/rclone/rclone/blob/50cf97fc729ee305cb89260f2c7ed387cf12bac0/cmd/progress.go#L73
> 
> I don't know whether it is worth doing this though, but it is possible with a bit of indirection!

I have made several changes to address the progress issue.  I changed syncFprintf to use a new wrapper function on fmt.Printf when printing to the terminal, made that function global, and then overrode that function from cmd.startProgress when the progress flag is used.  The override uses printProgress as suggested in place of fmt.Printf.  This is similar to how the override for the logging function is implemented.  It enables hash output to terminal and the progress lines to play nicely together.  While doing this I noticed that syncFprintf is well **F** print, and we hand a 'file' from the command to control the output location.  That file was always os.Stdout.  I have implemented an output-file flag for all of the hash commands to enable the user to optionally send the hashes to a text file.  This enables saving the hashes to a file without redirecting stdout and thus can be used with progress.  Users should see the progress information along with any actual logging (e.g. errors, no config file, etc.) in the terminal.

lostheli:
golint is very particular about how comments are written on global variables and functions!  The comments were slightly re-written to make it happy.

There is now a new test failure... build/mac says 'Error: No available tap local/openssl' and said it in the 'install libraries' section.  That almost sounds like an issue with how the test server is setup to manage capacity rather than an issue with this code.  Is there any way to simply force it to rerun the test on the same code?

ncw:
> golint is very particular about how comments are written on global variables and functions! The comments were slightly re-written to make it happy.

I think this is looking great now - thank you.

I shall merge it now :-)
 
> There is now a new test failure... build/mac says 'Error: No available tap local/openssl' and said it in the 'install libraries' section. That almost sounds like an issue with how the test server is setup to manage capacity rather than an issue with this code. Is there any way to simply force it to rerun the test on the same code?

I've fixed the master build now - the macOS builder broke. Once this is rebased on master after the merge it should be all good!

PR #4806, opened 2020-11-25
crypt: docs: extended description

albertony:

Extend main description of crypt remote.

There have been several questions in the forum about the basics of the crypt remote, so I have extended the top section of the doc page to try to explain some of the most relevant consepts in one place. Much of the text is taken from various forum posts and other documentation. Perhaps it became a bit too long, and it repeats several of the things that are explained further down the page.. Let me know what you think.

https://github.com/rclone/rclone/pull/4765#issuecomment-730681973
https://forum.rclone.org/t/encryption-at-rest-untrusted-destination-server/20539
https://forum.rclone.org/t/is-was-a-security-auditing-procedure-in-place-for-crypt/450
https://forum.rclone.org/t/how-to-correctly-setup-the-salt-for-the-crypt-remote/4273

#### Checklist

- [X] ~I have added documentation for the changes if appropriate.~

### DIFF
--- a/docs/content/crypt.md
+++ b/docs/content/crypt.md
@@ -8,59 +8,126 @@ description: "Encryption overlay remote"
 
 Rclone `crypt` remotes encrypt and decrypt other remotes.
 
-To use `crypt`, first set up the underlying remote. Follow the `rclone
-config` instructions for that remote.
+A remote of type `crypt` does not access a [storage system](https://rclone.org/overview/)
+directly, but instead wraps another remote, which in turn accesses
+the storage system. This is similar to how [alias](https://rclone.org/alias/),
+[union](https://rclone.org/union/), [chunker](https://rclone.org/chunker/)
+and a few others work. It makes the usage very flexible, as you can
+add a layer, in this case an encryption layer, on top of any other
+backend, even in multiple layers. Rclone's functionality
+can be used as with any other remote, for example you can
+[mount](https://rclone.org/commands/rclone_mount/) a crypt remote.
 
-`crypt` applied to a local pathname instead of a remote will
-encrypt and decrypt that directory, and can be used to encrypt USB
-removable drives.
+Accessing a storage system through a crypt remote realizes client-side
+encryption, which makes it safe to keep your data in a location you do
+not trust will not get compromised.
+When working against the `crypt` remote, rclone will automatically
+encrypt (before uploading) and decrypt (after downloading) on your local
+system as needed on the fly, leaving the data encrypted at rest in the
+wrapped remote. If you access the storage system using an application
+other than rclone, or access the wrapped remote directly using rclone,
+there will not be any encryption/decryption: Downloading existing content
+will just give you the encrypted (scrambled) format, and anything you
+upload will *not* become encrypted.
+
+The encryption is a secret-key encryption (also called symmetric key encryption)
+algorithm, where a password (or pass phrase) is used to generate real encryption key.
+The password can be supplied by user, or you may chose to let rclone
+generate one. It will be stored in the configuration file, in a lightly obscured form.
+If you are in an environment where you are not able to keep your configuration
+secured, you should add
+[configuration encryption](https://rclone.org/docs/#configuration-encryption)
+as protection. As long as you have this configuration file, you will be able to
+decrypt your data. Without the configuration file, as long as you remember
+the password (or keep it in a safe place), you can re-create the configuration
+and gain access to the existing data. You may also configure a corresponding
+remote in a different installation to access the same data.
+See below for guidance to [changing password](#changing-password).
+
+Encryption uses [cryptographic salt](https://en.wikipedia.org/wiki/Salt_(cryptography)),
+to permute the encryption key so that the same string may be encrypted in
+different ways. When configuring the crypt remote it is optional to enter a salt,
+or to let rclone generate a unique salt. If omitted, rclone uses a built-in unique string.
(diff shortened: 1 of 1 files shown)


edwardxml:
I am glad you tackled this thorny one. Will have a look in the morning from a wording not tech perspective. Ed.

albertony:
Not sure I have tackled it, yet, but its a suggestion and a start at least.. :)

I'm was wondering a bit about the terms "remote" and "backend", in my head "remote" is what you configure in rclone config. Backend is the underlying system behind a remote, usually a cloud service such as dropbox, but also crypt is a backend.

The example section configures a remote with name "remote". I think that can lead to some confusion. It makes it difficult to discuss remote as a general term and this specific configured remote, as in: "the underlying remote is called remote:path". I considered changing this, but it seems this is sort of a "standard" throughout the documentation.

edwardxml:
> start at least.. :)

It is another good bit of progress!

> "remote" and "backend"

I spend quite some time deciding which I should be referring to. Backends seem to be items off the list of backends and remotes are users' instances of them - as you say, as defined in rclone.conf.

> The example section configures a remote with name "remote".

As far as I can tell, that was just done for convenience. Crypt is a bit different because your creation explanation has to cover two remotes. I don't think it needs to be an issue.

A few generl points that struck me:

1. Is it good practice to make the crypt remote a subdirectory of the underlying remote? If so we maybe need to say that.

2. Should the subdirectory being used as crypt empt before being defined as a crypt remote? Do we say that anytwhere. It goes to the heart of us saying that to encrypt files they have to be loaded into a crypt remote. They are not encrypted by just defining a directory containing the files as a crypt.

3. Which leads to changing passwords. Do we need a two para subsection for that? Essemtially the need is to remove the prior encrypted files, change the password etc and then re-upload (or create a fresh crypt directory maybe with use of alias wrap remote). again the point is that changing the password on a remote doesn't change the encryption of pre existing files in a crypt - which is the concept people in the rclone forum seem to have struggled generally with. A subsection might be an opportunity to draw attention to the 1.53.3 disjoint - not often you get the change to do a highlight box!. It isn''t often rclone asks users to break backwards compatibility on their data. I don't know how @ncw has managed it really.

4. You might have removed the section on encrypting local folders / usb removeable drives. I struggled with where that belongs in the page. Maybe again it could have its own subsection. It doesn't particularly fit anywhere else but it does raise an interesting idea that can get the reader thinking about the capabilities and potential uses of rclone, albeit it isn't a mainstream use.

I will try to comment on the actual words in the PR as well today.

Cheers, Ed.

albertony:
Very good points there, thank you. I will look into it in more depth later.
Regarding the local folders / usb, I did not remove the text if that was what you meant? I basically moved it from being a separate section to be a part of the first section, following the text about the flexibility of the feature.

edwardxml:
> I did not remove the text if that was what you meant

Yes, just got to it. Personally I just don't think it belongs in the intro - based on the fact I left it there myself last time! It is a useful oddball point that was made and deserves to be on the page

I was returning to edit/correct my earlier post but you beat me to it :-)

Ed.

edwardxml:
> which is a very strong and highly regarded encryption method

This is something of an opinion and one I have little argument but I don't think it is necessary in reference documentation. This isn't front page sales documentation and to be fair to the average reader of this site they, once presented with the facts as you set them out, are competent to make a good judgement themselves. Those that are not competent to do so will have their own arbitrary beliefs and disregard whatever you say anyway.

> names is optional, and is also weaker

To make this sort of thing shorter you don't need the second 'is' and the 'also' is a bit superfluous given that it follows an 'and'. So 'names is optional but weaker' has the same meaning. Mostly I am trying to strip out excess words and shorten sentences but don't get too exercised about it because that can always be done afterwards - getting the meaning in to start with is what matters.

> content will be encrypted

Mostly I find it clearer to put most things in the present tense and reserve indications something is conditional or set in the future for those few occasions in reference documentation where it matters. So in this, and many other cases, 'content is encrypted' is simpler and has a satisfactorily equivalent meaning.

I think I saw a repeated 'the the' somewhere but you will already have got that!

There is some good new stuff in there.

Ed.

docs/content/crypt.md
```diff
@@ -119,20 +191,79 @@ Rclone does not encrypt
...
+effectively decrypting everything on the fly using the old password and
+re-encrypting using the new password. When done, delete the original crypt
+remote directory and finally the rclone crypt configuration with the old password.
+All data will be streamed from the storage system and back, so you will
+get half the bandwith and be charged twice if you have upload and download quota
+on the storage system.
+
+**Note**: Passwords generated by rclone config in version 1.49.0 (released 2019-08-26)
+to 1.53.2 (released 2020-10-26) are not considered secure and should be changed.
+then it will have been selected from a limited set of passwords and should be changed.
```
ncw, line 260: then it will => They will

ncw (commented):
I think you've done a brilliant job with that, thank you :-)

You've answered all the questions that come up frequently on the forum (like forgetting the `:` in your remote).

I noticed one minor typo - see inline.

If you can fix that then I'll merge it :-)

albertony:
I gave it a wash and a rebase. If there are any more typos or other comments let me know and I will look into it, but if not I'm fine with it being merged as is.

ncw (approved):
Looking very good and a great improvement - thank you :-)

I will merge this now.

ncw:
I fixed the conflict.

Will merge now.

PR #4807, opened 2020-11-25
About command docs

edwardxml:
Purpose:
Create a full loop of documentation for rclone about, backends overview and individual backend pages.

Discussion:
#4774 relates

Previously pull was requested, in part, under ref #4801

Notes:
Introduce a tentative draft see-link format at the end of sections to try rather than lots of in-para links.

Update about.go incl link to list of backends not supporting about.

In list of backends not supporting about, include link to about command reference.

I appreciate there may be decisions to make going forward about whether command links should be code formatted, and using proper pretty url links, but I have fudged that for now.

Update backend pages that do not support about with wording used previously for ftp - it is in passive voice but I can live with it. (my
own wording and fault). The note is applied to a limitations section. If one does not already exist it is created (even if there are other
limitations with their own sections)


albertony:
Looks good to me!
A bit unsure about the link text "List of backends that do not support rclone about". If the current backend does not support it, why would I want to see which other backends are also not supporting it? :) But I do not have a better text at the moment, and agree there should be a link, so don't know if it was right of me to mention it at all... :/

And I learned something new: That the file system size (total, used, free) of a mount is taken from the "about" feature (I use a remote with unlimited storage, so haven't noticed). I added a note about this in the mount documentation (as part of my open PR https://github.com/rclone/rclone/pull/4746). Thank you.

edwardxml:
> List of backends that do not support rclone about"

Yes, it is a bit strange! Maybe it is a list of backends that do support it! Let's get a link in for now. My main point was to try a format for links following paras or sections. At present there are few inline links and most likely sites are already code formatted. I don't really want to see lots of inline links superimposed if we can do it like this.

> if it was right of me to mention it

Definitely!

> And I learned something new

I hope I got it right then!

Ed.

ncw (approved):
This looks great now - I'll merge it

Thank you :-)

Issue #3213, opened 2019-05-21
Add Auth-msi to Azure blob Storage backend


MasterKuat:
#### What problem are you are trying to solve?
We need to rotate access key of our azure blob storage, so we need to change configuration in rclone as often as key's change.

#### How do you think rclone should be changed to solve that?
Azure provide AAD managed identity in order to retrieve Access key from storage account so we do not need storing key in rclone config.
More information on MSI : https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-msi

ncw:
This seems like a good idea.  This is the library we use: https://github.com/Azure/azure-storage-blob-go and here are [the docs](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob).

Any ideas how we would configure the SDK to do AAD/MSI?

@sandeepkru ?

sandeepkru:
Sorry for late response and also got disappeared for last few months, this definitely sounds like a good idea, I'm not sure if this supported in `go sdk` yet. I'll dig more into this and update with details.

sandeepkru:
Based on initial investigation, it doesn't look like supported to me. I've opened up an issue with client SDK team, will update more details when we hear from them.

symgryph:
I was wondering if this actually got done?

backerman:
The current SDK supports OAuth but doesn't actually get the token. On the gripping hand, azcopy does support MSI authentication and is MIT-licensed, so the [relevant bits](https://github.com/Azure/azure-storage-azcopy/blob/master/common/oauthTokenManager.go) can be lovingly copied into rclone.

ncw:
@backerman is this something you want to have a go at?

backerman:
@ncw Taking a look.

backerman:
@ncw It actually looks easier than I thought. I need to add some tests but I've got an [initial implementation](https://github.com/backerman/rclone/tree/personal/backerman/msi) in my fork.

ncw:
@backerman I had a look at your fork and I think you are doing very well :-)

I suggest you write lots of help under the `use_msi` definition as that is what gets built into the docs. The first line is the config flag parameter `--azureblob-use-msi` help string.

> It's currently using a dedicated boolean to enable this option, but defaulting to a system-assigned identity if there are no explicit credentials added could also work.

[...]

PR #4820, opened 2020-11-30
azureblob: add managed identity support

backerman:

Adds a configuration option (`azureblob_use_msi`) to use a managed service identity when authenticating to Azure from an Azure virtual machine, which eliminates the need to manage explicit credentials. This method acquires a bearer token (valid for 24 hours) from the [Azure Instance Metadata Service](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http), and if a temporary error is returned will retry using the exponential backoff method specified in the linked documentation.

The token is not presently persisted to disk; if users are expected to invoke the `rclone` binary a large number of times in a short period that functionality would be desirable. The present implementation also does not check token validity before calling the Azure Storage blob API, so if a `rclone` invocation lasts long enough that it calls the API more than 24 hours after process start that call will fail.

Issue #3213.

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -99,6 +99,31 @@ for more details.
 			Name: "sas_url",
 			Help: "SAS URL for container level access only\n(leave blank if using account/key or Emulator)",
 		}, {
+			Name: "use_msi",
+			Help: `Use a managed service identity to authenticate (only works in Azure)
+
+When true, use a [managed service identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/)
+to authenticate to Azure Storage instead of a SAS token or account key.
+
+If the VM(SS) on which this program is running has a system-assigned identity, it will
+be used by default. If the resource has no system-assigned but exactly one user-assigned identity,
+the user-assigned identity will be used by default. If the resource has multiple user-assigned
+identities, the identity to use must be explicitly specified using exactly one of the msi_object_id,
+msi_client_id, or msi_mi_res_id parameters.`,
+			Default: false,
+		}, {
+			Name:     "msi_object_id",
+			Help:     "Object ID of the user-assigned MSI to use, if any. Leave blank if msi_client_id or msi_mi_res_id specified.",
+			Advanced: true,
+		}, {
+			Name:     "msi_client_id",
+			Help:     "Object ID of the user-assigned MSI to use, if any. Leave blank if msi_object_id or msi_mi_res_id specified.",
+			Advanced: true,
+		}, {
+			Name:     "msi_mi_res_id",
+			Help:     "Azure resource ID of the user-assigned MSI to use, if any. Leave blank if msi_client_id or msi_object_id specified.",
+			Advanced: true,
+		}, {
 			Name:    "use_emulator",
 			Help:    "Uses local storage emulator if provided as 'true' (leave blank if using real azure storage endpoint)",
 			Default: false,
@@ -188,6 +213,10 @@ type Options struct {
 	Account              string               `config:"account"`
 	ServicePrincipalFile string               `config:"service_principal_file"`
 	Key                  string               `config:"key"`
+	UseMSI               bool                 `config:"use_msi"`
+	MSIObjectID          string               `config:"msi_object_id"`
+	MSIClientID          string               `config:"msi_client_id"`
+	MSIResourceID        string               `config:"msi_mi_res_id"`
 	Endpoint             string               `config:"endpoint"`
 	SASURL               string               `config:"sas_url"`
 	UploadCutoff         fs.SizeSuffix        `config:"upload_cutoff"`
@@ -217,6 +246,7 @@ type Fs struct {
 	isLimited     bool                            // if limited to one container
 	cache         *bucket.Cache                   // cache for container creation status
 	pacer         *fs.Pacer                       // To pace and retry the API calls
+	imdsPacer     *fs.Pacer                       // Same but for IMDS
 	uploadToken   *pacer.TokenDispenser           // control concurrency
 	pool          *pool.Pool                      // memory pool
 }
@@ -319,6 +349,8 @@ func (f *Fs) shouldRetry(err error) (bool, error) {
 				return true, err
 			}
 		}
+	} else if httpErr, ok := err.(httpError); ok {
+		return fserrors.ShouldRetryHTTP(httpErr.Response, retryErrorCodes), err
 	}
 	return fserrors.ShouldRetry(err), err
 }
@@ -479,6 +511,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		opt:         *opt,
 		ci:          ci,
 		pacer:       fs.NewPacer(ctx, pacer.NewS3(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
+		imdsPacer:   fs.NewPacer(ctx, pacer.NewAzureIMDS()),
 		uploadToken: pacer.NewTokenDispenser(ci.Transfers),
 		client:      fshttp.NewClient(ctx),
 		cache:       bucket.NewCache(),
@@ -490,6 +523,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			opt.MemoryPoolUseMmap,
 		),
 	}
+	f.imdsPacer.SetRetries(5) // per IMDS documentation
 	f.setRoot(root)
 	f.features = (&fs.Features{
 		ReadMimeType:      true,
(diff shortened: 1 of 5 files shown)


backerman:
Assuming the checks don't uncover any new issues that need fixing, I'll be able to check the "ready for review" box once I've added some tests to validate that the expected errors are returned when the IMDS response is 400/500-series.

backend/azureblob/imds.go
```diff
@@ -0,0 +1,141 @@
...
+	if isTest {
+		req.URL.Host = fmt.Sprintf("localhost:%d", testPort)
+		req.Host = req.URL.Host
+	}
+
+	// Send request
+	httpClient := fshttp.NewClient(ctx)
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		fs.Errorf(nil, "MSI is not enabled on this VM: %v", err)
```
ncw, line 98: That would probably be better returned as `errors.Wrap(err, "MSI is not enabled on this VM")` and the log isn't necessary in that case

backend/azureblob/imds.go
```diff
@@ -0,0 +1,141 @@
...
+	default:
+		body, _ := ioutil.ReadAll(resp.Body)
+		fs.Errorf(nil, "Couldn't obtain OAuth token from IMDS; server returned status code %d and body: %v", resp.StatusCode, string(body))
+		return result, httpError{Response: resp}
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		fs.Errorf(nil, "Couldn't read IMDS response: %v", err)
+		return result, err
```
ncw, line 125: This should be `errors.Wrap()` as well without the log.

backend/azureblob/imds.go
```diff
@@ -0,0 +1,141 @@
...
+	// Remove BOM, if any. azcopy does this so I'm following along.
+	b = bytes.TrimPrefix(b, []byte("\xef\xbb\xbf"))
+
+	// This would be a good place to persist the token if a large number of rclone
+	// invocations are being made in a short amount of time. If the token is
+	// persisted, the azureblob code will need to check for expiry before every
+	// storage API call.
+	err = json.Unmarshal(b, &result)
+	if err != nil {
+		fs.Errorf(nil, "Couldn't unmarshal IMDS response: %v", err)
```
ncw, line 136: ...and here

ncw (commented):
This is very nice code - thank you :-)

> Adds a configuration option (`azureblob_use_msi`) to use a managed service identity when authenticating to Azure from an Azure virtual machine, which eliminates the need to manage explicit credentials. This method acquires a bearer token (valid for 24 hours) from the [Azure Instance Metadata Service](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http), and if a temporary error is returned will retry using the exponential backoff method specified in the linked documentation.

Great

> The token is not presently persisted to disk; if users are expected to invoke the `rclone` binary a large number of times in a short period that functionality would be desirable.

Noted. If rclone users want to do this then they would be better off using the rc API.

> The present implementation also does not check token validity before calling the Azure Storage blob API, so if a `rclone` invocation lasts long enough that it calls the API more than 24 hours after process start that call will fail.

Hmm, I can see this being a potential problem. Users often kick off rclone jobs to copy TBs of data.

How difficult would it be to renew the token every few hours?

We'd then need locking around a lot more stuff though which would be rather annoying. If we want to do it then a different PR would probably be needed.

See inline for some more comments.

backend/azureblob/azureblob.go
```diff
@@ -453,16 +488,63 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		}
 		pipeline := f.newPipeline(credential, azblob.PipelineOptions{Retry: azblob.RetryOptions{TryTimeout: maxTryTimeout}})
 		serviceURL = azblob.NewServiceURL(*u, pipeline)
-	case opt.Account != "" && opt.Key != "":
-		credential, err := azblob.NewSharedKeyCredential(opt.Account, opt.Key)
+	case opt.UseMSI:
+		var token adal.Token
+		trimmedClientID := strings.TrimSpace(opt.MSIClientID)
```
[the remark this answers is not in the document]
backerman: Fixed.

backend/azureblob/imds.go
backerman: Changed these ones as suggested.

backend/azureblob/imds.go
```diff
@@ -0,0 +1,141 @@
...
+	return fmt.Sprintf("HTTP error %v (%v)", e.Response.StatusCode, e.Response.Status)
+}
+
+// GetMSIToken attempts to obtain an MSI token from the Azure Instance
+// Metadata Service.
+func GetMSIToken(ctx context.Context, identity *userMSI) (adal.Token, error) {
+	// Attempt to get an MSI token; silently continue if unsuccessful.
+	// Future work: Allow user to assert that an MSI token must be
+	// available.
+	// This code has been lovingly stolen from azcopy's OAuthTokenManager.
```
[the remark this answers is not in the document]
backerman: I'll look around again, but I don't think so - the sample Go for getting the token does it with `net/http` as well.

backerman:
> How difficult would it be to renew the token every few hours?

Looks like it's just including a [`TokenRefresher`](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#TokenRefresher) function where I've currently got a `nil`, so I'll do that. Shouldn't be too difficult and the pipeline package is already dealing with the concurrency issues there.

backend/azureblob/azureblob.go
```diff
@@ -453,16 +488,60 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
-			return nil, errors.Wrapf(err, "Failed to parse credentials")
+			return nil, errors.Wrapf(err, "Failed to acquire MSI token")
 		}
 
 		u, err = url.Parse(fmt.Sprintf("https://%s.%s", opt.Account, opt.Endpoint))
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to make azure storage url from account and endpoint")
 		}
+
+		credential := azblob.NewTokenCredential(token.AccessToken, nil)
```
backerman, line 537: This is where the `TokenRefresher` goes.

backerman:
Okay, added token refresh support, but I'm not sure how to fix this conflict warning; there doesn't even appear to be a conflicta.

backend/azureblob/imds.go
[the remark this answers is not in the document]
backerman: Looks like there is in fact `auth.NewAuthorizerFromEnvironment` which is the droid we're looking for.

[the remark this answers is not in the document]
backerman: And that's actually only for the main Azure Go SDK; the blob storage SDK doesn't use a compatible interface.

ncw:
Sorry for the delay responding. 

This looks ready to merge now - would you agree?

Thanks

Nick

backerman:
> This looks ready to merge now - would you agree?

I just rebased to master and modified the message generated when no authentication type is available (getting a bit long to list them individually). Looks good to me once the checks pass.

ncw:
That looks great now - thank you for the great feature :-)

I'll merge this now.

ncw:
Hi @backerman 

I just fixed a small oopsie in this code that the integration tests threw up cb97c2b0d39b38e1ae4f424a62503717495fa36c - that isn't a problem - these things happen!

What I was wondering about while looking at that was this in the new code you added. `opt.Account` and `opt.Endpoint` probably won't be set here will they? So maybe we don't need to make `u` at all.

https://github.com/rclone/rclone/blob/cb97c2b0d39b38e1ae4f424a62503717495fa36c/backend/azureblob/azureblob.go#L617-L620

We use it here - maybe we can pass in empty string or nil?

https://github.com/rclone/rclone/blob/cb97c2b0d39b38e1ae4f424a62503717495fa36c/backend/azureblob/azureblob.go#L645

What do you think?

backerman:
> Hi @backerman
> 
> I just fixed a small oopsie in this code that the integration tests threw up [cb97c2b](https://github.com/rclone/rclone/commit/cb97c2b0d39b38e1ae4f424a62503717495fa36c) - that isn't a problem - these things happen!
> 
> What I was wondering about while looking at that was this in the new code you added. `opt.Account` and `opt.Endpoint` probably won't be set here will they? So maybe we don't need to make `u` at all.

They should be set (at least `Account`) in the configuration for any authentication method; it would probably be a good idea to extract them from the switch statement, though.

ncw:
> > I just fixed a small oopsie in this code that the integration tests threw up [cb97c2b](https://github.com/rclone/rclone/commit/cb97c2b0d39b38e1ae4f424a62503717495fa36c) - that isn't a problem - these things happen!
> > What I was wondering about while looking at that was this in the new code you added. `opt.Account` and `opt.Endpoint` probably won't be set here will they? So maybe we don't need to make `u` at all.
> 
> They should be set (at least `Account`) in the configuration for any authentication method; it would probably be a good idea to extract them from the switch statement, though.

Ah OK!

I'm happy to leave it how it is then. Factoring it would be nice but not all the if branches need it so it probably isn't worth it.

PR #4822, opened 2020-11-30
fs: fix canonical name construction for local backends

ajankovic:

Name of the local fs was hard coded as a requirement for returning
naked root path. This is in conflict with Features().IsLocal because
fs can be of local type but with a different name.

This is making "local" to be an option not a condition for returning
naked root path.

No

#### Checklist

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1370,7 +1370,7 @@ func NewFs(ctx context.Context, path string) (Fs, error) {
 func ConfigString(f Fs) string {
 	name := f.Name()
 	root := f.Root()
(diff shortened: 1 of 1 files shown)


ncw (commented):
It is possible to have type=local backends with different configurations which will get mixed up with your change.

Do you have multiple local backends with different names?

ivandeex:
@ajankovic 
Could you please answer the question above?

ajankovic:
Closing after realization of how `Features().IsLocal` is supposed to be used.

PR #4836, opened 2020-12-03
gcs: Storage class object header support - fixes #3043

### LINKED ISSUE #3043: Support setting StorageClass on individual GCS objects
As an rclone GCS user, I would like to be able to set storage class for each object individually, separate from the bucket's current default class. 
As it is currently, during non server side copies, the StorageClass does not attempt to be set per object.
So for example, I could have a bucket with a default of `StorageClass = REGIONAL`, but I want to upload a series of files into the bucket with the `StorageClass = COLDLINE` defined per object, there is no way to do this.

GCS supports the option of setting an individual object storage class, different from the bucket class default. So I could have half my objects be Regional, and half Nearline in the same bucket for example.
Or my buckets default could be set to Regional, but I decide to upload any new objects as Coldline.
This is described here:
https://cloud.google.com/storage/docs/per-object-storage-class
https://cloud.google.com/storage/docs/storage-classes#bucket_vs_object_storage_class
The reason for this is generally so people can do tiering within the bucket.

The metadata field on the Object for that setting is:
```
    // StorageClass: Storage class of the object.
    StorageClass string `json:"storageClass,omitempty"`
```

From a previous thread:
> > I wonder if there should be another option added called object_storage_class to distinguish it from the bucket storage class?

> That would be easy enough to do, then you could set --gcs-object-storage-class when you wanted to do something different to what is set on the bucket.

>> Either that, or just go about setting the objects storage class individually based on the existing storage_class configuration setting.

> That would be easy enough too.
> I think the first idea is probably the most flexible, but I wonder whether it is overkill?

I think you are correct that adding a separate config seems like overkill. In my opinion it all comes down to how the user expects the application will function. 
Based on current documentation `The storage class to use when storing objects in Google Cloud Storage.` this seems misleading, given this issue. I can upload objects today, and they will be stored with a different class than what is defined in the config.

I took a very brief look at the AWS documentation (Im much less familiar with it) and it seems they ALSO support individual object StorageClasses. 
https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-class-intro.html#sc-howtoset

So its possible this change could apply to other systems beyond GCS?

iiLaurens:

Allow storage class to be set individually on objects, overriding the default storage class setting at the bucket level.

There is a valid reason to want this kind of feature; it allows for cost optimization. See also [this comment on stackoverflow](https://stackoverflow.com/a/62105146). With the proposed changes to the code, it would now be possible to set a default storage class that is cost-effective in operations cost, but still have individual objects stored using the cheaper COLDLINE or ARCHIVE storage classes.

See issue #3043

#### Checklist

### DIFF
--- a/backend/googlecloudstorage/googlecloudstorage.go
+++ b/backend/googlecloudstorage/googlecloudstorage.go
@@ -1091,6 +1091,8 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 			object.ContentLanguage = value
 		case "content-type":
 			object.ContentType = value
+		case "x-goog-storage-class":
+			object.StorageClass = value
 		default:
 			const googMetaPrefix = "x-goog-meta-"
 			if strings.HasPrefix(lowerKey, googMetaPrefix) {
--- a/docs/content/googlecloudstorage.md
+++ b/docs/content/googlecloudstorage.md
@@ -232,6 +232,7 @@ flag. Google Cloud Storage supports the headers as described in the
 - Content-Encoding
 - Content-Language
 - Content-Type
+- X-Goog-Storage-Class
 - X-Goog-Meta-
 
 Eg `--header-upload "Content-Type text/potato"`


iiLaurens:
Note that I did not explicit tests because the change is very small. However the google cloud storage documentation is pretty clear in its implementation in the Go language, [as can be seen here](https://cloud.google.com/storage/docs/viewing-editing-metadata#storage-set-object-metadata-go) and [also here](https://cloud.google.com/storage/docs/xml-api/put-object). Also if any anecdotal evidence helps, I already applied this patch on my NAS and it works like a charm.

ncw (approved):
That looks really great - thank you.

I'll merge it now :-)

PR #4848, opened 2020-12-06
azure: add support for service principals - fixes #3230

### LINKED ISSUE #3230: Azure Blob using service principal and oauth2

rpm -q rclone 
rclone-1.47.0-1

accessing azure blob using oauth2 and service principal

Access azureblob using oauth2 to get an access token owning client_id and client_key:

```
POST https://login.microsoftonline.com/{{tenant_id}}/oauth2/token

body:
client_id: xxxxx (id_service principal)
client_secret: XXXXXXXXXXXXXXXXXXXXXXX
resource: https://storage.azure.com/
grant_type: client_credentials
```

below a simple shell script client:

```sh
#!/bin/bash

client_id="xxx"
client_secret="yyy"
app_id="zzz"
storage_account="myaccount"
container_name="backup"

# url encoded resource
resource="https%3A%2F%2Fstorage.azure.com%2F"

curl_opts='-H "Content-Type: application/x-www-form-urlencoded"'
mthd="https"

# Access Token
json=$(curl -s -X POST -d "grant_type=client_credentials&resource=${resource}&client_id=${client_id}&client_secret=${client_secret}" "${mthd}://login.microsoftonline.com/${app_id}/oauth2/token?api-version=1.0")

if [[ $? -gt 0 || ${#json} -lt 80 ]]; then
echo "error getting token"
exit
fi

access_token=$(awk '/access_token/ {if (match($0,/"access_token":"([^"]+)/,at)) print at[1];}' <<< $json)
token_type=$(awk '/token_type/ {if (match($0,/"token_type":"([^"]+)/,tt)) print tt[1];}' <<< $json)

body=$(curl -s \
-H "Authorization: $token_type $access_token" \
-H 'x-ms-blob-type: BlockBlob' \
-H 'x-ms-version: 2018-03-28' \
"${mthd}://${storage_account}.blob.core.windows.net/${container_name}?restype=container&comp=list")
```

jimjh:
Hi! 👋  first-timer here.

This PR uses `adal` and creates its own token refresher instead of using `oauthutil`, because I could not figure out how to integrate Azure Storage's pipeline with a custom oauth client. It was relatively straightforward, however, to use `newPipeline` with a new service principal credential type.

Ref:
- https://github.com/rclone/rclone/issues/3230
- https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-app?tabs=dotnet#well-known-values-for-authentication-with-azure-ad
- https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#available-authentication-types-and-methods
- https://gist.github.com/ItalyPaleAle/ec6498bfa81a96f9ca27a2da6f60a770

azure: add support for service principals

Before: users can only connect to Azure blob containers using the access keys
from the storage account.

After: users can additionally choose connect to Azure blob containers
using service principals. This uses OAuth2 under the hood to exchange
a client ID and client secret for a short-lived access token.

https://github.com/rclone/rclone/issues/3230

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -10,8 +10,10 @@ import (
 	"crypto/md5"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"path"
@@ -21,6 +23,7 @@ import (
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
@@ -33,6 +36,7 @@ import (
 	"github.com/rclone/rclone/fs/walk"
 	"github.com/rclone/rclone/lib/bucket"
 	"github.com/rclone/rclone/lib/encoder"
+	"github.com/rclone/rclone/lib/env"
 	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/pool"
 	"github.com/rclone/rclone/lib/readers"
@@ -74,6 +78,20 @@ func init() {
 		Options: []fs.Option{{
 			Name: "account",
 			Help: "Storage Account Name (leave blank to use SAS URL or Emulator)",
+		}, {
+			Name: "service_principal_file",
+			Help: `Path to file containing credentials for use with a service principal.
+
+Leave blank normally. Needed only if you want to use a service principal instead of interactive login.
+
+    $ az sp create-for-rbac --name "<name>" \
+      --role "Storage Blob Data Owner" \
+      --scopes "/subscriptions/<subscription>/resourceGroups/<resource-group>/providers/Microsoft.Storage/storageAccounts/<storage-account>/blobServices/default/containers/<container>" \
+      > azure-principal.json
+
+See [Use Azure CLI to assign an Azure role for access to blob and queue data](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-rbac-cli)
+for more details.
+`,
 		}, {
 			Name: "key",
 			Help: "Storage Account Key (leave blank to use SAS URL or Emulator)",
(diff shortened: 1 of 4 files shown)


backend/azureblob/azureblob.go
```diff
@@ -74,6 +79,24 @@ func init() {
...
+
+Leave blank normally. Needed only if you want to use a service principal instead of interactive login.
+
+First, follow the instructions at [Microsoft AD](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals)
+to create a service principal by registering a new application. Grant this application permissions to use your storage
+container.
+
+Then, create a new client secret, and copy the application ID, client secret, and tenant ID into a JSON file, like so:
+
+	{
```
ncw, line 94: Is this a format that the user can download? I suspect it isn't, in which case we'd normally store these in the config file. That will mean the user can copy the config file about and rclone will carry on working.

So I'd suggest 3 entries in the config file client_id/client_secret/tenant_id

I think this will simplify your logic below

backend/azureblob/azureblob.go
```diff
@@ -10,17 +10,22 @@ import (
...
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"path"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/rclone/rclone/lib/env"
+
```
ncw, line 25: Can you remove this blank line and put the added import in sorted order (likely your editor will do this)

backend/azureblob/azureblob_test.go
```diff
@@ -5,8 +5,11 @@
 package azureblob
 
 import (
+	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
```
ncw, line 12: Can you remove this blank line and put the added import in sorted order (likely your editor will do this)

ncw (commented):
I've given this a quick once over.

Great feature - thank you!

You say this is like oauth, but does it include a challenge the user has to do in the browser? When does that happen if so?

See inline for comments

Thank you!

backend/azureblob/azureblob.go
jimjh: Thanks for the suggestion. I think there is a way I can simplify the instructions for creating a service principal, and as part of that auto-generate a file that users can use as a service principal file. I am going to try it out, and let me know what you think. If the UX remains awkward, I am happy to take your suggestion and create 3 additional entries in the config file.

jimjh:
@ncw Thanks for the review! This specific OAuth flow does not prompt the user, and is designed to work with daemons and other non-interactive applications. It matches "4. Client Credentials Flow" on https://darutk.medium.com/diagrams-and-movies-of-all-the-oauth-2-0-flows-194f3c3ade85.

backend/azureblob/azureblob.go
```diff
@@ -74,6 +78,20 @@ func init() {
...
+		}, {
+			Name: "service_principal_file",
+			Help: `Path to file containing credentials for use with a service principal.
+
+Leave blank normally. Needed only if you want to use a service principal instead of interactive login.
+
+    $ az sp create-for-rbac --name "<name>" \
+      --role "Storage Blob Data Owner" \
+      --scopes "/subscriptions/<subscription>/resourceGroups/<resource-group>/providers/Microsoft.Storage/storageAccounts/<storage-account>/blobServices/default/containers/<container>" \
+      > azure-principal.json
```
jimjh, line 90: @ncw example of how one could automatically generate credentials using the Azure CLI.

ncw: Nice :-)

ncw (approved):
I think that looks great now.

A command to generate the JSON file is perfect. If there is user complaints we can always add 3 separate config variables as well ;-)

I'll merge this now - thank you very much for an excellent first contribution :-)

jonathansp:
@ncw @jimjh could you please provide an example of what this _azure-principal.json_ file should look like? 

I think it is more common having this info already created in another place like an akv, for instance:
```python
"azure.account.auth.type": "OAuth",
"azure.account.oauth.provider.type": "org.apache.hadoop.fs.azurebfs.oauth2.ClientCredsTokenProvider",
"azure.account.oauth2.client.id": <SERVICE_PRINCIPAL_ID>,
"azure.account.oauth2.client.secret":< SERVICE_PRINCIPAL_SECRET>,
"azure.account.oauth2.client.endpoint": "https://login.microsoftonline.com/<MS_DIRECTORY_ID>/oauth2/token",
```
so I wonder how I can achieve this by using a file.

PR #4852, opened 2020-12-07
docs: fix references to upstreams in union example

sntran:

Updates documentation for Union remote to reflect the correct paths and upstreams

No.

#### Checklist

### DIFF
--- a/docs/content/union.md
+++ b/docs/content/union.md
@@ -19,11 +19,11 @@ e.g. `remote:directory/subdirectory:ro` or `remote:directory/subdirectory:nc`.
 
 Subfolders can be used in upstream remotes. Assume a union remote named `backup`
 with the remotes `mydrive:private/backup`. Invoking `rclone mkdir backup:desktop`
-is exactly the same as invoking `rclone mkdir mydrive2:/backup/desktop`.
+is exactly the same as invoking `rclone mkdir mydrive:private/backup/desktop`.
 
 There will be no special handling of paths containing `..` segments.
 Invoking `rclone mkdir backup:../desktop` is exactly the same as invoking
-`rclone mkdir mydrive2:/backup/../desktop`.
+`rclone mkdir mydrive:private/backup/../desktop`.
 
 ### Behavior / Policies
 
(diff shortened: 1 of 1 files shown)


docs/content/union.md
```diff
@@ -119,7 +119,7 @@ Storage> union
 List of space separated upstreams.
 Can be 'upstreama:test/dir upstreamb:', '\"upstreama:test/space:ro dir\" upstreamb:', etc.
 Enter a string value. Press Enter for the default ("").
-upstreams>
+upstreams> C:\dir1 C:\dir2 C:\dir3
```
ncw, line 122: I don't like this example on the local disk - can we use `mydrive:private/backup` like the example above?

ncw (commented):
Great spotting on the text! See inline for a comment on the `upstreams>` help

docs/content/union.md
sntran: I based this on the following examples when the usage of the remote are showed.

image

If you want to change this, we will need to change those examples too.

ncw: Ah! Yes we probably should as `C:\dir1` is particularly confusing as it is a Windows path, but looks kind of like an rclone remote.

Maybe make 3 remotes for examples and stick to those?  s3:dir1 gdrive:dir2 dropbox:dir3 or something like that? (to pick 3  popular backends!)

ivandeex:
@sntran Are you still with us?

> Maybe make 3 remotes for examples and stick to those? s3:dir1 gdrive:dir2 dropbox:dir3 or something like that? (to pick 3 popular backends!)

sntran:
Hi, yes.

I was thinking about the @ncw's comment, but I was not sure whether to proceed as he suggested, and I lost track of time to other projects.

I think the reason the examples used local remotes was to simplify the config, and I think it's the correct approach. We want to show example of union remote, not how to create different types of remote. Switching to use other types of remote would make the example config long.

docs/content/union.md
ncw: I've fixed this up now

ncw:
I've fixed this up so we can merge this change - thank you very much @sntran 

Do send a followup if you think it can be improved further.

PR #4853, opened 2020-12-08
operations + accounting: add stats to --dry-run

### LINKED ISSUE #4624: RClone Dry run for sync and copy should provide summary at the end
1.53

RClone Dry run for sync and copy should provide a summary section at the end without -v flag. The summary should include the following on separate lines:

1. total amount of data to be transferred,
2. Total files that will be added
3. Total files that will be modified 
4. Total files that will be deleted
5. Size of the smallest file and 
6. Size of the largest file.

Additionally, display the size of each file (currently only the file names are displayed in the list format when running with -v)

This should work for both regular log formats as well as JSON log formats. It should also work if using cli or in remote control mode

Currently, the dry run does not display any of this information. Change the code to display the summary block as outlined above at the end of the current dry run output. Also, display the file size next to each file displayed currently.

lasombra:
Fixes #4624

#### Checklist

### DIFF
--- a/fs/accounting/accounting.go
+++ b/fs/accounting/accounting.go
@@ -280,6 +280,12 @@ func (acc *Account) ServerSideCopyEnd(n int64) {
 	acc.stats.Bytes(n)
 }
 
+// DryRun accounts for statistics without running the operation
+func (acc *Account) DryRun(n int64) {
+	acc.ServerSideCopyStart()
+	acc.ServerSideCopyEnd(n)
+}
+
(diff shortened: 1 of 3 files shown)


lasombra:
I don't think those are related to the changes from this PR.

ncw:
I've removed the extraneous bits from this commit. I've also added another one which turns the stats logs on if --dry-run/--interactive is set.

I've pushed it up and if it goes green I'll merge!

Issue #4793, opened 2020-11-23
sync: add option to always skip hashsum check


zhucan:
https://github.com/rclone/rclone/blob/211b08f771472cbe6f0a60824010b84303d9b12e/fs/operations/operations.go#L216

zhucan:
@ncw  I have other question, why not check 'mtime' first, check 'size' second  and check 'hash' third?

ncw:
Rclone will only get to that point if the sizes are the same but the modtimes are different which usually means a file needs transferring. However rclone does a check with checksum just to make sure and if it is really the same it just updates the modtime.



> I have other question, why not check 'mtime' first, check 'size' second and check 'hash' third?

Size is the quickest thing to check so we check that first.

We only check the checksum if the modtime if different, but the size is the same (without `--checksum`)

zhucan:
> Rclone will only get to that point if the sizes are the same but the modtimes are different which usually means a file needs transferring. However rclone does a check with checksum just to make sure and if it is really the same it just updates the modtime.
> 
> > I have other question, why not check 'mtime' first, check 'size' second and check 'hash' third?
> 
> Size is the quickest thing to check so we check that first.
> 
> We only check the checksum if the modtime if different, but the size is the same (without `--checksum`)

If sizes are the same but the modtimes are different which usually means files needs transferring,  it's needed to check the checksum? Because of check the checksum that there is a lot of pressure on storage.

zhucan:
@ncw I have other question, Can not copy acl for object from a bucket to other bucket with rclone？

ncw:
> If sizes are the same but the modtimes are different which usually means files needs transferring, it's needed to check the checksum? Because of check the checksum that there is a lot of pressure on storage.

What would you like rclone to do here? Just upload the file?

If so then you need the flag

      --no-update-modtime    Don't update destination mod-time if files identical.

> I have other question, Can not copy acl for object from a bucket to other bucket with rclone？

Which backend is this?

And are you doing a server side copy or not?

I think for a server side copy rclone will copy the ACL for S3.

zhucan:
> I think for a server side copy rclone will copy the ACL for S3.

Backend is ceph rgw, maybe not support server side copy.

zhucan:
> What would you like rclone to do here? Just upload the file?

Yes, Just upload, But we want to check hash, only want to check  modtime & size.

ncw:
> Yes, Just upload, But we want to check hash, only want to check modtime & size.

[...]

PR #4856, opened 2020-12-10
sync: add option to always skip hashsum check

zhucan:
Signed-off-by: zhuc <zhucan.k8s@gmail.com>

add option to always skip hashsum check

https://github.com/rclone/rclone/issues/4793

#### Checklist

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -124,6 +124,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.StringArrayVarP(flagSet, &downloadHeaders, "header-download", "", nil, "Set HTTP header for download transactions")
 	flags.StringArrayVarP(flagSet, &headers, "header", "", nil, "Set HTTP header for all transactions")
 	flags.BoolVarP(flagSet, &ci.RefreshTimes, "refresh-times", "", ci.RefreshTimes, "Refresh the modtime of remote files.")
+	flags.BoolVarP(flagSet, &ci.LogSystemdSupport, "log-systemd", "", ci.LogSystemdSupport, "Activate systemd integration for the logger.")
+	flags.BoolVarP(flagSet, &ci.NoCheckSum, "--no-checksum", "", ci.NoCheckSum, "If set assume size match and mtime differ, but not to check checksum")
 }
 
 // ParseHeaders converts the strings passed in via the header flags into HTTPOptions
--- a/fs/config.go
+++ b/fs/config.go
@@ -121,6 +121,7 @@ type ConfigInfo struct {
 	DownloadHeaders        []*HTTPOption
 	Headers                []*HTTPOption
 	RefreshTimes           bool
+	NoCheckSum             bool
 }
 
 // NewConfig creates a new config with everything set to the default
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -140,6 +140,7 @@ type equalOpt struct {
 	checkSum          bool // if set check checksum+size instead of modtime+size
 	updateModTime     bool // if set update the modtime if hashes identical and checking with modtime+size
 	forceModTimeMatch bool // if set assume modtimes match
+	noCheckSum        bool // if set assume size match and mtime differ, not to check checksum
 }
 
 // default set of options for equal()
@@ -150,6 +151,7 @@ func defaultEqualOpt(ctx context.Context) equalOpt {
 		checkSum:          ci.CheckSum,
 		updateModTime:     !ci.NoUpdateModTime,
 		forceModTimeMatch: false,
+		noCheckSum:        ci.NoCheckSum,
 	}
 }
 
@@ -215,15 +217,17 @@ func equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object, opt equalOpt) 
 		fs.Debugf(src, "Modification times differ by %s: %v, %v", dt, srcModTime, dstModTime)
 	}
 
-	// Check if the hashes are the same
-	same, ht, _ := CheckHashes(ctx, src, dst)
-	if !same {
-		fs.Debugf(src, "%v differ", ht)
-		return false
-	}
-	if ht == hash.None && !ci.RefreshTimes {
-		// if couldn't check hash, return that they differ
-		return false
+	if !ci.NoCheckSum {
+		// Check if the hashes are the same
+		same, ht, _ := CheckHashes(ctx, src, dst)
+		if !same {
+			fs.Debugf(src, "%v differ", ht)
+			return false
+		}
+		if ht == hash.None && !ci.RefreshTimes {
+			// if couldn't check hash, return that they differ
+			return false
+		}
 	}
 
 	// mod time differs but hash is the same to reset mod time if required


ncw (commented):
The code looks OK, however docs are missing from `docs/content/docs.md`

What this flag does is

While syncing, if the size matches and the modtime is different then assume the files are equal **without** doing a checksum test. This then sets the modification time of the destination file unless `--no-update-modtime`

I have a problem with the name `--no-checksum` - it isn't descriptive enough for a flag which probably should never be used by the average user

I also have a problem working out why you would ever want this flag. If you set the flag then you are effectively matching sizes on the source and dest and if they are equal then you are setting the modification time on the dest. This is a really unreliable sync! So it has the worst parts of the --size-only sync.

I need a really good use case for this flag to go in the docs if it is to be merged.

zhucan:
> I also have a problem working out why you would ever want this flag. If you set the flag then you are effectively matching sizes on the source and dest and if they are equal then you are setting the modification time on the dest. This is a really unreliable sync! So it has the worst parts of the --size-only sync.

I use the flag on rclone copy.  I know if not check sum, it's really unreliable. But user know the source and dest are the same, only mtime is not same, It's no need to check sum. Because of more check sum operation will cause more read pressure on backend storage .

ivandeex:
Remaining open items:

- docs in `docs/content/docs.md` with technical details of `what this flag does`
- name `--no-checksum` isn't descriptive enough, choose better name
- add a `real life` use case in the docs (billions of files? petabytes of storage? the way you use rclone in `your project`? etc)

zhucan:
@ivandeex Is there better suggestion name for the flags? I thinks we can determine the name first and add docs to it.

fs/config/configflags/configflags.go
```diff
@@ -124,6 +124,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.StringArrayVarP(flagSet, &downloadHeaders, "header-download", "", nil, "Set HTTP header for download transactions")
 	flags.StringArrayVarP(flagSet, &headers, "header", "", nil, "Set HTTP header for all transactions")
 	flags.BoolVarP(flagSet, &ci.RefreshTimes, "refresh-times", "", ci.RefreshTimes, "Refresh the modtime of remote files.")
+	flags.BoolVarP(flagSet, &ci.LogSystemdSupport, "log-systemd", "", ci.LogSystemdSupport, "Activate systemd integration for the logger.")
```
ncw, line 127: There seems to be a spurious addition here?

ncw (commented):
Let's think about what to call this...

Firstly exactly what does it do

> While syncing, if the size matches and the modtime is different then assume the files are equal without doing a checksum test. This then sets the modification time of the destination file unless --no-update-modtime

This flag reminds me of [--refresh-times](https://rclone.org/docs/#refresh-times) which has the help

> The --refresh-times flag can be used to update modification times of existing files when they are out of sync on backends which don't support hashes.

So maybe this flag should be called --refresh-times-ignore-checksum

Which would make its help something like

> The --refresh-times-ignore-checksum flag can be used to update modification times of existing files when they are out of sync. It works like --refresh-times except it doesn't attempt to check checksums. Instead if the size **only** of the source and destination files match then it will update the modification time in the destination.

We could also possibly change the `--refresh-times` flag to do what this flag does. It emphasises in the help that `--refresh-times` should only be used on backends which don't support checksums, so all existing uses should have  `ht == hash.None` always and `--refresh-times` is equivalent to `--refresh-times-ignore-checksum`.

```go
        if !ci.NoCheckSum {
		// Check if the hashes are the same
		same, ht, _ := CheckHashes(ctx, src, dst)
		if !same {
			fs.Debugf(src, "%v differ", ht)
			return false
		}
		if ht == hash.None && !ci.RefreshTimes {
			// if couldn't check hash, return that they differ
			return false
		}
	}
```

So another possibility would be to move the check of ci.RefreshTimes into the place we hae ci.NoChecksum now.

This would mean then we only have just the one crazy `--refresh-times` flag.

Thoughts? @ivandeex ?

zhucan:
> The --refresh-times-ignore-checksum flag can be used to update modification times of existing files when they are out of sync. It works like --refresh-times except it doesn't attempt to check checksums. Instead if the size **only** of the source and destination files match then it will update the modification time in the destination.

It's ok. Any other thoughts or suggestions?

zhucan:
The use cases: There is only "incremental write" in source backend,  only the mtime of the file in source backend maybe modified,  so we only need to check size or mtime, no need to checksum, if there are many files on the source backend, checksum will cause many pressure on source backend；This is customer's real case. @ncw

ivandeex:
@ncw marking "regional" as I expect this thread to take long

ivandeex:
> I think we can determine the name first and add docs to it.

The problem here is that we don't understand what problem you are trying to solve and why you use the sequence of _if_s you use. Other people who use rclone will be happy to understand what this feature can do for them, too. Otherwise it's going to be your private flag, which is useless for others.

> Is there better suggestion name for the flags?

As long as I don't have any insight, I will internally call this flag `--zhucan-quirk3` (it's not the first, I expect there can be more)
Or just describe the sequence of _if_s `--check-only-maybe-mtime-not-checksum`

> There is only "incremental write" in source backend, only the mtime of the file in source backend maybe modified, so we only need to check size or mtime, no need to checksum, if there are many files on the source backend, checksum will cause many pressure on source backend

How do you measure "pressure"?
How much gain in your units of pressure does it give you?
What is amount of data? Several petabyte object or millions of megabyte-size objects or what?
Did you consider removing time check too?
Maybe you should use network cards with direct access to remove Linux kernel (if you use Linux) from access path?
Did you consider other tools for your task?
Maybe writing a custom high-performance data cracker in pure C can solve your particular task?

Without answering at least some of these question I don't understand how we can test this change or measure its success.
Waiting for your docs.

Last but not least. Does the integration test pass with this flag on?

ivandeex:
> I think we can determine the name first and add docs to it.

The problem here is that we don't understand what problem you are trying to solve and why you use the sequence of _if_s you use. Other people who use rclone will be happy to understand what this feature can do for them, too. Otherwise it's going to be your private flag, which is useless for others.

> Is there better suggestion name for the flags?

As long as I don't have any insight, I will internally call this flag `--zhucan-quirk3` (it's not the first, I expect there can be more)
Or just describe the sequence of _if_s `--check-only-maybe-mtime-not-checksum`

Or `--refresh-times-ignore-checksum` proposed by @ncw above

> There is only "incremental write" in source backend, only the mtime of the file in source backend maybe modified, so we only need to check size or mtime, no need to checksum, if there are many files on the source backend, checksum will cause many pressure on source backend

This is actually a description of the sequence of confitional statements in the patch, not a use case. I'll try to shed some light again. Generally I can see that you do something large and exotic, bigdata or something like that.

"many" - What is amount of data? Several petabyte-size objects or millions of megabyte-size objects or what?
How do you measure "pressure"? Time to wait for I/O? Bandwidth? Else?
How much gain in your units of pressure does this patch give you?
Did you consider removing time check and update too?
You use "Ceph", I recall. Did you consider removing gateway?
If you use ceph over network, maybe you should use network cards with direct access to remove Linux kernel (if you use Linux) from access path?
Maybe rclone is wrong tool for your very specific project. Did you try other tools for your task? Maybe some very fast "Ceph Rados" management client?
Maybe writing a custom high-performance data cracker in pure C can solve your particular task?

Without answering at least some of these questions I don't understand how we can test this change or measure its success.
Waiting for your docs.

Last but not least - testing. You didn't write any unit test for your feature. Does the current integration test pass with this flag on, at least?

ivandeex:
Copy-pasting here another attempt (from linked issue) by @ncw to understand what's going on in your testbed
(which wasn't answered either....)

> >   Because of there is hundreds of millions of small files will cause great read pressure on storage, if we check hash. so we want to only check modtime and size under the condition that the customer can guarantee the consistency of md5.

> Why do you have 100s of millions of files which has
>  -  size the same
>  -  modtime different
> That isn't a usual situation. Are you trying to get the modtimes set on all the files that didn't have them set before?

> What you want is a flag which says "if mod times are the same then don't check the checksum, transfer anyway".
> I looked to see if rsync has a flag which does this - it doesn't!

(See my question above about other tools you've tried to resolve your unspoken task.....)

That thread had more proposals from @ncw for the flag name

> So we could make a flag called --no-checksum-if-same-age

> Another idea would be to only do the checksum calculation if we are not using --no-update-modtime but that would mean that if you had --no-update-modtime=true then rclone would transfer identical files which is undesirable.

Bottomline:

> It's ok. Any other thoughts or suggestions?

Yes, plenty of thoughts, request for docs and four suggestions for flag name.

ivandeex:
@zhucan ping

ivandeex:
@zhucan
Probably you could write an automatic unit test or describe below a manual test that presents your use case and demonstrates performance improvement introduced by your patch. It does not have to involve bulk or production data. If it shows e.g. 1sec improvement on average with a test dataset, that would give us input for extrapolation.
Thank you

zhucan:
> @zhucan
> Probably you could write an automatic unit test or describe below a manual test that presents your use case and demonstrates performance improvement introduced by your patch. It does not have to involve bulk or production data. If it shows e.g. 1sec (or N-better "pressure") improvement on average with a test dataset, that would give us input for extrapolation.
> Thank you

I'm sorry, now, I have no time to do this, I will do it out of day

ivandeex:
This pull request is nearly abandoned.
Open questions are waiting for resolution since 2021-02-17, ~4 months.
It can become eligible for closing after ~6 months of radio silence!
cc @ncw @zhucan

ivandeex:
This PR is abandoned by author and waits for volunteers.
cc @ncw @zhucan

ivandeex:
Closing as abandoned

Issue #1077, opened 2017-01-30
[Feature request] Encrypt filenames only.


AggelosM:
I have encrypted all my files on ACD but not the filenames, and I think it would be nice to have the option to encrypt just the filenames without having to decrypt an then re-encrypt the whole files.

I don't know if such a think is possible though.

ajkis:
Encrypting files names would be just rename and yea it would be great to have some option like encrypt file names and decrypt file names.

alex-phillips:
+1 - would be a great feature to be able to swap between "standard" and "off" filename encrypting without having to redown/up everything. Simply execute rename commands on each file / folder based on which direction you're going.

Vesnyx:
@ncw 

I would like to being implementing this sort of feature, but I'm not sure where to get started.
They way I would like to see it implemented is as an option available to all storage options. If the user created the remote with a flag like "--encrypt-names", then it would decrypt the filenames on mount. Is this possible to implement this way and which file would I begin with?

ncw:
> @ncw
> 
> I would like to being implementing this sort of feature, but I'm not sure where to get started.
> They way I would like to see it implemented is as an option available to all storage options. If the user created the remote with a flag like "--encrypt-names", then it would decrypt the filenames on mount. Is this possible to implement this way and which file would I begin with?

THis is going to be a one off process isn't it?  Run through the remote decrypting the names, renaming the files then changing the crypt mode.

It would be possible to script this using the output from `rclone lsjson --encrypted`.  Probably mounting the underlying remote and doing renames on the files would be most effective.

Vesnyx:
The usecase for the original issue can be solved like that.

I interpreted the issue's name slightly differently. My intention was to add a new feature to crypt that only encrypts filenames (copying a new file results in only name encryption not the data). For this, I'm not sure if it's best to implement in the crypt bucket or as a general flag. 

Essentially, creating a crypt overlay on a bucket with this option will only encrypt and decrypt filenames, not contents.

ncw:
> The usecase for the original issue can be solved like that.
> 
> I interpreted the issue's name slightly differently. My intention was to add a new feature to crypt that only encrypts filenames (copying a new file results in only name encryption not the data). For this, I'm not sure if it's best to implement in the crypt bucket or as a general flag.
> 
> Essentially, creating a crypt overlay on a bucket with this option will only encrypt and decrypt filenames, not contents.

[...]

PR #4866, opened 2020-12-13
crypt: add --crypt-filename-compression option to compress unicode filename before encryption (WIP)

Milly:
**WIP** now. Please advice to this feature.
I will combine the commits into one later.

Filename encryption inflates its length.
Some remote backends has a limitation of the filename length that counted by *characters* (not bytes).
So, if it contains multi-bytes characters in UTF-8, the filename length was not problematic in the remote can be a problem in the crypt backend.

If remote backend has 255 characters limit.  "standard" mode has limits below.

* ASCII (7bits, 1byte/char) ~143 characters **limited**
* Alphabetic (Latin, Cyrillic, etc., 2bytes/char) ~71 characters **big limited**
* East-Asian (CJK, 3bytes/char) ~47 characters  **extremely limited**
                                                                         
If `--crypt-filename-compression` option is enabled, the filename is compressed by [SCSU (Standard Compression Scheme for Unicode)](https://en.wikipedia.org/wiki/Standard_Compression_Scheme_for_Unicode) before encryption.
This feature allows non-ASCII file names to be stored up to the same length as ASCII file names.

The problem of filename length has been discussed a lot.
However, it seems that the problem extremely limited in East Asian characters has not been discussed.

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -26,15 +26,17 @@ import (
 
 // Constants
 const (
-	nameCipherBlockSize = aes.BlockSize
-	fileMagic           = "RCLONE\x00\x00"
-	fileMagicSize       = len(fileMagic)
-	fileNonceSize       = 24
-	fileHeaderSize      = fileMagicSize + fileNonceSize
-	blockHeaderSize     = secretbox.Overhead
-	blockDataSize       = 64 * 1024
-	blockSize           = blockHeaderSize + blockDataSize
-	encryptedSuffix     = ".bin" // when file name encryption is off we add this suffix to make sure the cloud provider doesn't process the file
+	nameCipherBlockSize    = aes.BlockSize
+	fileMagic              = "RCLONE\x00\x00"
+	fileMagicSize          = len(fileMagic)
+	fileNonceSize          = 24
+	fileHeaderSize         = fileMagicSize + fileNonceSize
+	blockHeaderSize        = secretbox.Overhead
+	blockDataSize          = 64 * 1024
+	blockSize              = blockHeaderSize + blockDataSize
+	encryptedSuffix        = ".bin" // when file name encryption is off we add this suffix to make sure the cloud provider doesn't process the file
+	compressNameHeader     = "\x00"
+	compressNameHeaderSize = len(compressNameHeader)
 )
 
 // Errors returned by cipher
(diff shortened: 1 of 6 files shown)


ivandeex:
Related to _crypt: add option to not encrypt data_ #1077 #2981

Milly:
Currently CI failed in go1.12 and earlier, because the [SCSU library](https://github.com/dop251/scsu) uses `errors.Is`.
I'm PRed to it.

Milly:
https://forum.rclone.org/t/filename-too-long/12753

ivandeex:
@ncw
Does the crypt format allow for future extensions?
Is it possible to add a _data format_ marker to a crypt-encrypted value? Like a byte or variadic integer meaning e.g. 1="encrypted bytes", 2="SCSU-compressed encrypted bytes", 3-...="reserved"

If a meta-format field is not reserved a.t.m., is there a _currently impossible_ (but valid) combination of `crypt`-stream bits, which can be used to "hack" the field into existing wire format?

If SCSU appears succesful we could default to it in future and decode **automatically** while maintaining backward compatibility. Or "SCSU+" or "encrypted bytes with modtime" or whatever...

ivandeex:
Marking as `breaking change` because older versions of rclone cannot parse new _wire format_ correctly.

ncw:
This is very interesting - thank you.

We've been working on a file name compression scheme based on pre-defined huffman tables using @klauspost 's excellent library.

Currently the scheme uses the first byte of the name to select the huffman table to use - the idea being that we collect distributions of file name characters from users to build some typical huffman tables. The compression is quick enough that we can try all the tables to find the minimum match.

Using SCSU as well sounds ike an interesting idea. Maybe before the huffman coding or on its own.

Do you have a corpus of file names that we could experiment on? It would be interesting to experiment with this vs huffman encoding.

Milly:
The problematic point in that case.
The appropriate corpus depends on the language and users.

klauspost:
This seems to mostly achieve compression by shifting down character codes.

This means it retains codes 32 -> 127 ( plus CR, LF, TAB) as is, but places another set of characters in the "upper" 128->255 part, so they don't take up several bytes. The remaining are used for selecting what should be in the upper window.

It could be interesting to combine this.

klauspost:
@Milly See #4901 - this adds scsu as a compression type that can be selected if it is optimal.  This extends the filename compression I prepared for @ncw

ivandeex:
Marking WIP PRs as draft.

Milly:
Features that use SCSU were merged in #4901.
So close this.
Thank you @klauspost !!

PR #4868, opened 2020-12-14
systemd log integration

aloababa:

The thing was to not make auto systemd detection if `--log-file` was set.

So this proposal only auto detect systemd integration if `--log-file` is not set.
And if both `--log-file` and `--log-systemd` are set then systemd integration is disabled since logs will not go into stdout/err.

I also moved `--log-systemd` from config flags to log flags to keep log flags together.

There is also the case when `--syslog` and `--log-systemd` are both set. In this PR since the `startSysLog` function is called after `enableSystemdLog` function and they both override the fs.LogPrint variable, the systemd integration will always be disabled.

https://github.com/rclone/rclone/pull/4609

#### Checklist

### DIFF
--- a/fs/log/log.go
+++ b/fs/log/log.go
@@ -3,7 +3,6 @@ package log
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"log"
 	"os"
@@ -12,7 +11,6 @@ import (
 	"strings"
 
 	systemd "github.com/iguanesolutions/go-systemd/v5"
-	sysdjournald "github.com/iguanesolutions/go-systemd/v5/journald"
 	"github.com/rclone/rclone/fs"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


hekmon (approved):

ncw:
I'm really sorry, but I fixed this independently in c98dd8755cb99f1e05531d7bf692140058d790f9 and I didn't see your pull request first :-(

Your patch has some stuff mine doesn't in (though they are remarkably similar!) so if you wanted to change this PR to add those extra bits that would be great (like not compiling in systemd except on linux).

aloababa:
I have updated the PR :)

ncw:
Thank you :-) That looks great!

Can you rebase it on master and force push just to check the build works OK.

Also can you update the commit message with a note that it doesn't compile systemd support on non UNIX platforms.

Thanks

(I'd do this for you but I think you've disabled pushing to your branch - at least it isn't working for me!)

aloababa:
Seems that the checks have passed ;)

ncw:
That looks great now - thank you very much

I'll merge it now.

Issue #4889, opened 2020-12-28
Rclone not delete orphan segments. 


nguyenhuuluan434:
#### What is the problem you are having with rclone?
When i umount the mount point with the command 'fusermount -u /65d4e907c8c248b099d2f63a070a41b9' during the rclone sync/upload data to remote (in this case is Openstack Swift).The error occurred when the object/file is large object (the object what has size greater than 'chunk_size'). The upload process is not clean the segments what was uploaded.

Example:
I has a file with size 100M and set up chunk_size = 20M and assume segment 1 and segment 2 upload successful and then upload segment 3 but the command umount is executed. The rclone stop process and segment 1 and segment 2 still on remote backend. In the next mount, the rclone check the file is dirty and upload process is started again.

Could we trigger delete the segments what was upload successful but it is not use any where ?
Another proposal is waiting finish the upload of the file.

#### What is your rclone version (output from `rclone version`)
rclone v1.54.0-DEV
- os/arch: linux/amd64
- go version: go1.15.2

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Linux LuanNH 4.15.0-128-generic #131-Ubuntu SMP Wed Dec 9 06:57:35 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

####  Which cloud storage system are you using? (e.g. Google Drive)
Openstack Swift

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
rclone mount test1-swift:container /65d4e907c8c248b099d2f63a070a41b9 --vfs-cache-mode full --vfs-cache-max-size=10000000B --debug-fuse --allow-other --file-perms 0644 --config=/home/luannh/.config/rclone/test1 --vfs-read-chunk-size=200M --umask=002 --cache-writes --vfs-cache-max-age=10s --log-file=/tmp/log --log-level=DEBUG --dir-cache-time=10s --vfs-cache-poll-interval=5s

ncw:
I just merged a potential fix for this in 6342499c472752c3837e557f76d40a5d7f6620cf so can you try [the latest beta](https://beta.rclone.org) please?

ivandeex:
Related to #4075

nguyenhuuluan434:
Hi @ncw @ivandeex ,
After reading source implement for s3, I see the package atexit what we can apply to handle this case. I am rewriting for swift and commit soon. @ncw please help me review.

Thanks.

[...]

PR #4893, opened 2020-12-29
Modifying code handle swift backend

nguyenhuuluan434:
#### What is the purpose of this change? ####

  

The PR is create with some purpose:
  - Add '_' char to white list when encode url.
  - Unify the util check and get manifest for DLO or SLO to common function.
  - Using delete bulk during delete segments.
  - Using the atexit package to handle unexpected error during sync cache to remote backend

[link request https://github.com/rclone/rclone/issues/4889](https://github.com/rclone/rclone/issues/4889)
#### Checklist

### DIFF
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -283,7 +283,6 @@ var retryErrorCodes = []int{
 	401, // Unauthorized (e.g. "Token has expired")
 	408, // Request Timeout
 	409, // Conflict - various states that could be resolved on a retry
-	413, // Upload exceeds quota
 	429, // Rate exceeded.
 	500, // Get occasional 500 Internal Server Error
 	503, // Service Unavailable/Slow Down - "Reduce your request rate"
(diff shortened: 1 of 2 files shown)


ivandeex:
This change must include a unit test that simulates acting on a single segment from two rclone copies and checks that segment is neither left dangling nor prematurely deleted afterwards.

nguyenhuuluan434:
Hi @ivandeex ,
I had added unit test for large object but this is happy case upload successful. Could you recommend any suggestion what i can cancel during run unit test to trigger the atexit.OnError to delete the uploaded segments ? 

Thanks,
Luan

ncw:
> I had added unit test for large object but this is happy case upload successful.

Well done - looks nice :-)

> Could you recommend any suggestion what i can cancel during run unit test to trigger the atexit.OnError to delete the uploaded segments ?

What we usually do here is create a reader which gives an error after reading so many bytes. You can create one by combining the bytes reader and an error reader something like this

```go
	contents := random.String(contentSize)
￼	buf := bytes.NewBufferString(contents[:errposition])
        er := &readers.ErrorReader{Err: errors.New("potato")} // from lib/readers
        in := ioutil.NopCloser(io.MultiReader(buf, err))
￼```

**Comment by nguyenhuuluan434** [comment]
Hi @ncw ,
I had added the unit test for the case what need clean up segments.
Please help me review.
Thanks,
Luan

**Comment by nguyenhuuluan434** [comment]
Hi @ncw ,
Just remind. :-)

File: backend/swift/swift.go [hunk 1]
```diff
@@ -1127,48 +1135,77 @@ func min(x, y int64) int64 {
...
 		}
-		return shouldRetry(err)
-	})
-	if err == nil {
-		fs.Debugf(o, "Removed empty container %q", segmentsContainer)
 	}
 	return nil
 }
 
+//func removeObjects(f *Fs, container string, prefix string) (err error) {
```
ncw, line 1171: Can you remove the dead code please.

backend/swift/swift.go
```diff
@@ -1194,7 +1231,7 @@ func urlEncode(str string) string {
 	var buf bytes.Buffer
 	for i := 0; i < len(str); i++ {
 		c := str[i]
-		if (c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '/' || c == '.' {
+		if (c == '_' || c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '/' || c == '.' {
```
ncw, line 1234: What is this change about?

It looks wrong though - I think it should be:

```go
if (c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '/' || c == '.'  || c == '_' {
```

backend/swift/swift.go
```diff
@@ -276,6 +283,7 @@ var retryErrorCodes = []int{
 	401, // Unauthorized (e.g. "Token has expired")
 	408, // Request Timeout
 	409, // Conflict - various states that could be resolved on a retry
+	413, // Upload exceeds quota
```
ncw, line 286: Is this something we want to retry? If we retry it then surely the upload will still exceed the quota?

ncw (commented):
See inline for a few notes.

Note that when I run the integration tests with `go test -v` (you'll need docker installed) these don't pass, in particular the `TestIntegration/FsMkdir/FsPutChunked` tests fail indicating there is a problem with the new code :-(

I didn't figure out what - can you see the problem?

backend/swift/swift.go
nguyenhuuluan434: my mistake :(

nguyenhuuluan434:
Hi @ncw ,
The integration tests fail because the Swift accept dash char ("-") but in urlEncode function will encode this to %2D. I has added dash to white list char during encode.
Please help me review again.
Thanks,
Luan.

nguyenhuuluan434:
Hi @ncw ,
Help me review, please.
Thanks,
Luân Nguyễn Hữu.

ivandeex:
@nguyenhuuluan434 
Please mark resolved converssations

ncw (approved):
This is looking great now, thank you @nguyenhuuluan434 

I'm going to squash and merge now :-)

PR #4895, opened 2020-12-30
restic: rest server metrics

aloababa:

I wanted to have metrics like the restic rest server has -> https://github.com/restic/rest-server.
So i've implemented a /metrics endpoint in the restic serve command with a `--prometheus` flag to enable it.
But i was wondering if the right place for those metrics is the rc server.
Like it will be a private field of `accounting.StatsInfo` and will be a part of core/stats if they are available.
And the metrics will be served by the prometheus registry already present in the accounting package.

### DIFF
--- a/cmd/serve/restic/prometheus.go
+++ b/cmd/serve/restic/prometheus.go
@@ -15,7 +15,7 @@ var (
 	promGatherer   prometheus.Gatherer   = promRegistry
 )
 
-var metricLabelList = []string{"repo", "type"}
+var metricLabelList = []string{"user", "repo", "type"}
 
 var (
 	metricBlobWriteTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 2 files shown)


ncw:
> I wanted to have metrics like the restic rest server has -> https://github.com/restic/rest-server.
> So i've implemented a /metrics endpoint in the restic serve command with a `--prometheus` flag to enable it.

Are those metrics compatible with rest-server?

> But i was wondering if the right place for those metrics is the rc server.
> Like it will be a private field of `accounting.StatsInfo` and will be a part of core/stats if they are available.
> And the metrics will be served by the prometheus registry already present in the accounting package.

The `rc` server can already serve prometheus metrics if you enable the --rc and supply --rc-enable-metrics.

So I guess your question is, should the rc serve the restic serve metrics? I think some of them will be represented in there already, but some will be missing. So should we add the missing ones to the `rc` metrics?

aloababa:
> Are those metrics compatible with rest-server?

No, since prometheus labels for those metrics are repository, user and type (where type is the type of the blob like locks or snapshots or data...) are restic specific.

> So I guess your question is, should the rc serve the restic serve metrics? I think some of them will be represented in there already, but some will be missing. So should we add the missing ones to the rc metrics?

Yes that was my question if those metrics are restic specific should we add them in the rc server or leave them into the restic server.

ivandeex:
@ncw @aloababa
The latest comment here dates back to 2021-01-04
Please do not abandon this ticket!
Thank you

ivandeex:
@ncw Please review

ncw:
@aloababa are you interested in continuing with the PR? If so can you rebase it to see if anything has broken since quite a bit has changed in the http handling since it was started.

Looking at the stats, they don't seem very restic specific, so maybe they should be in the HTTP server?

ivandeex:
closing stale requests

PR #4901, opened 2021-01-04
Add SCSU as tables

klauspost:

Instead of only adding SCSU to #4866, add it as an existing table.

Allow direct SCSU and add a reasonable table as well.

Add byte interfaces that doesn't base64 encode the URL as well with `EncodeBytes` and `DecodeBytes`.

Fuzz tested and decode tests added.

Only in issue #4866

File used for generating the table: [sample.txt.gz](https://github.com/rclone/rclone/files/5765636/sample.txt.gz) using `go run gentable.go -index=sample.txt -scsu -all`

#### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/calebcase/tmpfile v1.0.2 // indirect
 	github.com/colinmarc/hdfs/v2 v2.2.0
 	github.com/coreos/go-semver v0.3.0
+	github.com/dop251/scsu v0.0.0-20200422003335-8fadfb689669
 	github.com/dropbox/dropbox-sdk-go-unofficial v5.6.0+incompatible
 	github.com/gabriel-vasile/mimetype v1.1.2
 	github.com/gogo/protobuf v1.3.2 // indirect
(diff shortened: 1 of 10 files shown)


ncw:
Sorry for the long delay looking at this!

I think the idea is great and we should merge :-)

The tests indicate we probably need to exclude go1.12 from the build though (we've dropped go1.11 support just recently).

klauspost:
@ncw No problem. I don't feel like forking the upstream just for this, so removing 1.12 (which AFAICT has know security issues) would probably be fine.

ncw:
Do you want to put some build constraints in to exclude go1.12? That would be fine with me.

The mac build failure is a flakey test

The xgo failure is because the go.mod is out of date WRT the code and go is trying to update it on a read only fs I think.

klauspost:
@ncw Updated+master merged.

klauspost:
The remaining failures appears to be unrelated now.

ncw:
I just added a commit to try to fix the build/other_os - I think what is happening is that the go.mod file isn't sorted so go1.13 is sorting it and can't write it back to the read only docker image...

ncw:
If that works, I'll rebase the branch and merge :-)

ncw:
OK all built!  I'll merge now - thank you @klauspost 

Next crypt integration and a bit of experimentation.

klauspost:
@ncw Great. For crypt input you can use the byte interfaces which of course is more compact than the base64 string versions.

Issue #4924, opened 2021-01-13
Telia Cloud support


patriknordlen:
#### What is your current rclone version (output from `rclone version`)?

v1.54.0-DEV

#### What problem are you are trying to solve?

rclone currently doesn't support Telia Cloud. Telia Cloud is a whitelabel version of Jottacloud. Unlike other whitelabels (Elgiganten, Com Hem, Telia Sky) that are supported through legacy authentication, Telia Cloud uses a slightly different authentication flow that among other things keep the username internally and uses BankID for authentication instead of a password.

#### How do you think rclone should be changed to solve that?

Add support for Telia Cloud in the jottacloud backend.

PR #4930, opened 2021-01-14
jottacloud: Add support for Telia Cloud

patriknordlen:

This PR adds support for Telia Cloud in the jottacloud backend. It adds another authentication option during setup, so where users previously chose between standard and legacy authentication, a third option to use Telia Cloud authentication has been added. The rest is handled via a standard oauth2 flow.

#4924

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -276,7 +276,6 @@ func teliaCloudConfig(ctx context.Context, name string, m configmap.Mapper) {
 
 	m.Set("configVersion", strconv.Itoa(configVersion))
 	m.Set(configClientID, teliaCloudClientID)
-	m.Set(configClientSecret, "")
 	m.Set(configTokenURL, teliaCloudTokenURL)
 }
 
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -63,6 +63,10 @@ const (
 	v1ClientID              = "nibfk8biu12ju7hpqomr8b1e40"
 	v1EncryptedClientSecret = "Vp8eAv7eVElMnQwN-kgU9cbhgApNDaMqWdlDi5qFydlQoji4JBxrGMF2"
 	v1configVersion         = 0
+
+	teliaCloudTokenURL = "https://cloud-auth.telia.se/auth/realms/telia_se/protocol/openid-connect/token"
+	teliaCloudAuthURL  = "https://cloud-auth.telia.se/auth/realms/telia_se/protocol/openid-connect/auth"
+	teliaCloudClientID = "desktop"
 )
 
 var (
@@ -105,11 +109,18 @@ func init() {
 				}
 			}
 
-			fmt.Printf("Use legacy authentication?.\nThis is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.\n")
-			if config.Confirm(false) {
-				v1config(ctx, name, m)
-			} else {
+			fmt.Printf("Choose authentication type:\n" +
+				"1: Standard authentication - use this if you're a normal Jottacloud user.\n" +
+				"2: Legacy authentication - this is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.\n" +
+				"3: Telia Cloud authentication - use this if you are using Telia Cloud.\n")
+
+			switch config.ChooseNumber("Your choice", 1, 3) {
+			case 1:
 				v2config(ctx, name, m)
+			case 2:
+				v1config(ctx, name, m)
+			case 3:
+				teliaCloudConfig(ctx, name, m)
 			}
 		},
 		Options: []fs.Option{{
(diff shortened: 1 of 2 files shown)


patriknordlen:
@buengese

backend/jottacloud/jottacloud.go
```diff
@@ -228,6 +239,47 @@ func shouldRetry(resp *http.Response, err error) (bool, error) {
...
+			log.Fatalf("Failed to setup mountpoint: %s", err)
+		}
+		m.Set(configDevice, device)
+		m.Set(configMountpoint, mountpoint)
+	}
+
+	m.Set("configVersion", strconv.Itoa(configVersion))
+	m.Set(configClientID, teliaCloudClientID)
+	m.Set(configClientSecret, "")
+	m.Set(configTokenURL, teliaCloudTokenURL)
```
buengese, line 280: You are setting clientID and clientSecret here but using configVersion 1 (which is v2 config) which means those aren't actually loaded. Does refresh actually work?

The config Version stuff is turning out kinda messy I might have to rethink that a bit at some point.

patriknordlen: Yeah it threw me off at first that v1configVersion is actually 0 and configVersion is 1 :)

Refresh works with non-legacy (i.e. configVersion = 1). I confirmed it works when the client_id is specified in the config but doesn't work if it's missing. I didn't go through the whole flow to verify if this is just due to a lucky side effect of unintentional behaviour though. FWIW it's worked for me without problems for a month.

buengese: Thinking about it clientID and clientSecret aren't even part of the normal OAuth refresh flow. I know it's just a nitpick but setting them here gives the suggestion of them being used when in reality they aren't (unless I'm grossly mistaken or missed an obvious bug) so I'd suggest removing them here and checking if it still works.

Besides this minor nitpick this looks good.

I'm kinda

patriknordlen: I can confirm that it doesn't work without `client_id` but works with it. It's trying to send the wrong token as refresh token when no `client_id` is specified. Will investigate further why it does that and circle back.

patriknordlen: Ok, so correcting myself, the token is the same but the `client_id` is passed in the refresh request, first in an `Authorization: Basic` header, and if that fails it tries to issue another request where `client_id` is specified in the POST data instead. So, when `client_id` isn't set the default value `jottacli` will be used. This client is not valid for the `telia_se` realm used by Telia Cloud.

You're right that `client_secret` isn't needed though, so it can be omitted. Will push a commit for that.

TheCake21: Oh yeah I forgot about the auth header. Sorry

buengese: Oh yeah sorry forgot about the Auth header.

buengese:
I've merged this now. Thank you.

larsvinc:
I am trying to Telia Cloud, but when I choose "Telia Cloud authentication" a Swedish Telia login opens up. It will not accept my credentials, and the "Bank-ID" is different from the Norwegian one. I wonder if there is something domestic to the login here. I see some _se references in the code. If you agree, I would file this as an issue of not supporting Norwegian Telia Cloud?

patriknordlen:
@larsvinc Telia Cloud only exists as an offering in Sweden as far as I know. If there is a Norwegian version, it might be possible to handle it via the legacy authentication. Do you have a URL to the Norwegian service?

larsvinc:
@patriknordlen I have Telia as an ISP. It used to be called "get", and get had a whitelabel of yottacloud called "getsky". It is now relabeled "Telia Sky". It is still redirecting from sky.telia.no to the login to login.get.no.

I have tried the legacy authentication, but it does not work.

This is the old config entry that does not work any longer:
[getsky]
type = jottacloud
tokenURL = https://www.getsky.no/auth/realms/get/protocol/openid-connect/token
token = {"access_token":"SOMEVERYLONGTOKEN","expiry":"2020-09-15T17:48:43.451515165+02:00"}
configVersion = 1
md5_memory_limit = 1G

There is also an option to generate a token after logging in.

patriknordlen:
@larsvinc go ahead and create an issue and we can continue there. I'm curious about when you say that it doesn't work anymore - i.e. is this something that used to work in earlier versions of rclone and no longer doesn't, or did it stop working for reasons more likely on their end?

larsvinc:
@patriknordlen It seems some token expired. I was running version 1.51, and it used to work. Now I can no-longer re-authenticate. I tried to revert back to version 1.51, and it fails with the following message:
2021/02/10 20:28:30 Failed to create file system for "getsky:/": couldn't get customer info: Get https://api.jottacloud.com/account/v1/customer: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description":"Offline user session not found"}

I will also post this as an separate issue.

Issue #42, opened 2015-03-24
HDFS support


mlanner:
Hi,

Any thoughts or plans around HDFS support? It could be very nice to have a way to, for example, bring a given data set out of Swift into Hadoop to run jobs on.

ncw:
Interesting idea!  I don't know anything about HDFS so I've done a small amount of reading...

rclone would work fine with HDFS if it was mounted in the file system (eg using NFS Proxy or Fuse).  rclone is very undemanding on the local filesystem it scans directories, opens and reads or writes files in sequence which I think should all be supported by HDFS.

Alternatively it could use the API.  There has been come work using go with hadoop
  * https://github.com/hortonworks/gohadoop/ (HDFS not supported yet though)

Or use [LibHDFS](http://wiki.apache.org/hadoop/LibHDFS)

I don't have access to a hadoop cluster so I can't work on this, but would be grateful for assistance or patches!

mlanner:
Hi Nick,

Thanks for the quick response and links. I started looking at those ... and some others. I will probably spin up a test Hadoop cluster to see what I can do. I'll update this when/if I have any updates.

seanorama:
Anyone played more with making this happen?

ncw:
> Anyone played more with making this happen?

Not as far as I know.  Do you want to have a go?

volth:
Using HDFS API it possible to get hashes of file chunks without reading file content.
So using FUSE is not optimal here.

ei-grad:
Here is a native hdfs client library: https://github.com/colinmarc/hdfs

urykhy:
I try to make HDFS support. Backend tests passed, but hashes not implemented.

ncw:
> I try to make HDFS support. Backend tests passed, but hashes not implemented.

Great work!

- Is there a docker image we can test against?
- What have you been testing against?
- Are there hashes we could use?
- Are you going to submit a PR?

urykhy:
> Is there a docker image we can test against?

i fail to find simple docker image, maybe [this](https://jcristharif.com/hadoop-test-cluster/) can help.

> What have you been testing against?

i run tests with [this](https://bitbucket.org/urykhy/docker/src/master/hadoop/) image.

> Are there hashes we could use?

i have read about hadoop checksums, and it's complicated:
[...]

PR #4931, opened 2021-01-16
hdfs: support kerberos authentication #42

urykhy:

access HDFS cluster with KERBEROS authentication

[yes](https://github.com/rclone/rclone/issues/42#issuecomment-759630846)

#### Checklist


backend/hdfs/fs.go
```diff
@@ -35,11 +83,26 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
 
-	client, err := hdfs.NewClient(hdfs.ClientOptions{
+	options := hdfs.ClientOptions{
 		Addresses:           []string{opt.Namenode},
-		User:                opt.Username,
 		UseDatanodeHostname: false,
-	})
+	}
+
+	if len(opt.ServicePrincipalName) > 0 {
```
ncw, line 91: `if opt.ServicePrincipalName != "" {` would probably be more idiomatic. It compiles to the same code anyway!

backend/hdfs/fs.go
```diff
@@ -35,11 +83,26 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
+		options.KerberosClient, err = getKerberosClient()
+		if err != nil {
+			return nil, fmt.Errorf("Problem with kerberos authentication: %s", err)
+		}
+		options.KerberosServicePrincipleName = opt.ServicePrincipalName
+
+		if len(opt.DataTransferProtection) > 0 {
```
ncw, line 98: ...same here

backend/hdfs/hdfs.go
```diff
@@ -32,6 +32,25 @@ func init() {
 				Value: "root",
 				Help:  "Connect to hdfs as root",
 			}},
+		}, {
+			Name:     "service_principal_name",
+			Help:     "Enables KERBEROS authentication. Specifies the Service Principle Name (<SERVICE>/<FQDN>) for the namenode.",
```
ncw, line 37: Note that we try to wrap the help at 80 characters. The first line becomes the command line help for the flag.

You can use the backquote character for writing strings over multiple lines, so I'd write


```go
Help: `Kerberos service principal name for the namenode.

Enables KERBEROS authentication. Specifies the Service Principal Name
(<SERVICE>/<FQDN>) for the namenode.
```

backend/hdfs/hdfs.go
```diff
@@ -32,6 +32,25 @@ func init() {
...
+			Required: false,
+			Examples: []fs.OptionExample{{
+				Value: "hdfs/namenode.hadoop.docker",
+				Help:  "Namenode running as service 'hdfs' with FQDN 'namenode.hadoop.docker'.",
+			}},
+			Advanced: true,
+		}, {
+			Name: "data_transfer_protection",
+			Help: "Specifies whether or not authentication, data signature integrity checks, and wire encryption is required when communicating the the datanodes." +
```
ncw, line 46: Likewise here

```go
Help: `Kerberos data transfer protection: authentication|integrity|privacy

Specifies whether or not authentication, data signature integrity
checks, and wire encryption is required when communicating the the
datanodes. Possible values are 'authentication', 'integrity' and
'privacy'. Used only with KERBEROS enabled.
`
```

backend/hdfs/fs.go
```diff
@@ -35,11 +83,26 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
 		UseDatanodeHostname: false,
-	})
+	}
+
+	if len(opt.ServicePrincipalName) > 0 {
+		options.KerberosClient, err = getKerberosClient()
+		if err != nil {
+			return nil, fmt.Errorf("Problem with kerberos authentication: %s", err)
+		}
+		options.KerberosServicePrincipleName = opt.ServicePrincipalName
```
ncw, line 96: Principle or Principal? We should stick to one spelling!

I think `principal` is correct, which means that `hdfs.ClientOptions` is spelled wrong!

ncw (commented):
I think this basically looks fine :-)

See inline for some minor spelling/layout nits.

Can we add this to the integration tests easily?

PS Forgot to say - thank you for doing this work - it is in demand by users :-)

backend/hdfs/fs.go
urykhy: yes.

should we replicate tipo from hdfs library ?

urykhy:
> Can we add this to the integration tests easily?

i've add option to use kerberos in tests.

backend/hdfs/hdfs.go
```diff
@@ -33,8 +33,11 @@ func init() {
 				Help:  "Connect to hdfs as root",
 			}},
 		}, {
-			Name:     "service_principal_name",
-			Help:     "Enables KERBEROS authentication. Specifies the Service Principle Name (<SERVICE>/<FQDN>) for the namenode.",
+			Name: "service_principal_name",
+			Help: `Kerberos service principal name for the namenode
+
+Enables KERBEROS authentication. Specifies the Service Principle Name
```
ncw, line 39: Principal here!

backend/hdfs/fs.go
ncw: That is in the exported struct from the hdfs library isn't it? We'll have to live with that one!

docs/content/hdfs.md
```diff
@@ -187,7 +187,10 @@ Here are the advanced options specific to hdfs (Hadoop distributed file system).
 
 #### --hdfs-service-principal-name
 
-Enables KERBEROS authentication. Specifies the Service Principle Name (<SERVICE>/<FQDN>) for the namenode.
+Kerberos service principal name for the namenode
+
+Enables KERBEROS authentication. Specifies the Service Principle Name
```
ncw, line 192: Principal here

ncw (commented):
I spotted some more `Principles` inline!

Well done making a test - that looks great.

Do we need to test with and without Kerberos? If you think testing with Kerberos only is sufficient I'm happy to go with that.

docs/content/hdfs.md
urykhy: yes.
`KerberosServicePrincipleName` comes from library.

urykhy:
> Do we need to test with and without Kerberos? If you think testing with Kerberos only is sufficient I'm happy to go with that.

Kerberos tests is a bit more complicated to run (local krb5.conf required).
So... i'm in doubt.

ncw:
> > Do we need to test with and without Kerberos? If you think testing with Kerberos only is sufficient I'm happy to go with that.
> 
> Kerberos tests is a bit more complicated to run (local krb5.conf required).
> So... i'm in doubt.

What I'll do is merge it as it is and we can see how the integration tests go.

I'll pull and push the updated image.

Thank you very much for doing this @urykhy

Issue #4928, opened 2021-01-13
failure to upload small text file to S3+compress


ivandeex:
#### What is the problem you are having with rclone?

I cannot copy a small local file (content: `123`) to S3+compress remote
It fails with error
```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>BadDigest</Code><Message>The Content-MD5 you specified did not match what we received.</Message><ExpectedDigest>d577273ff885c3f84dadb8578bb41399</ExpectedDigest><CalculatedDigest>Cr3pWSEOQp0EYXU/Aj4L5g==</CalculatedDigest><RequestId>B60273C6B66E4A1A</RequestId><HostId>waSWEItiupTc3weCov8Hw8ORuyOOzj9kPpznvb3aENVcsxfIIPk4a7hy97crYGk8DA2nwrwk9vA=</HostId></Error>
```
Copying to the underlying S3 remote works as expected.

#### What is your rclone version (output from `rclone version`)

https://beta.rclone.org/v1.54.0-beta.5062.db2c38b21/rclone-v1.54.0-beta.5062.db2c38b21-linux-amd64.zip

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Ubuntu 18.04 
Linux 4.15.0-128-generic #131-Ubuntu SMP Wed Dec 9 06:57:35 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

####  Which cloud storage system are you using? (e.g. Google Drive)

S3 + Compress

```
[TestS3]
type = s3
provider = AWS
env_auth = false
access_key_id = xxxx
secret_access_key = xxxx
region = us-east-1
edndpoint =
acl = private
[TestCompress]
type = compress
remote = TestS3:
```

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
echo 123 > test.txt
./rclone-v1.54.0-beta.5062.db2c38b21-linux-amd64/rclone copy -vvv --dump headers,bodies --log-file test.log test.txt TestCompress:ivatest20210111
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

<details>
<summary>test.log</summary>

```
deex@hardy:~$ cat test.log
[...]

PR #4932, opened 2021-01-17
compress: fix remotes without PutStream & ObjectInfo wrapping.

buengese:
Quick postmortem here: #4928 was caused by the wrapped ObjectInfo actually returning the checksum & size of the uncompressed object which obviously differs from the compressed object actually upload (duh!). Since we cannot know the checksum & size of the compressed object beforehand there is really no useful information that can be returned here.

While wondering how I didn't catch that beforehand I found a second bug.  Prior to https://github.com/rclone/rclone/commit/886b3abac17404910d89192fae74cdec47515702 I was using operations.Rcat() instead of put when uploading compressed files which had the side effect of masking this bug since it didn't make use of the ObjectInfo wrapping. Well, turns out there was a reason I was using operations.Rcat() there since we cannot know the size of the compressed file beforehand (second duh!) this was required to be able to support underlying remotes without put stream. I've now implemented a Jottacloud-MD5 style workaround for this that spools the file to ram or disk to determine it's size and uploads using normal Put.

I'd also strongly suggest adding additional integration tests for this remote similar chunker. At least one non-local file-based & one bucket based remote would probably be a good idea here @ncw.


ncw:
> Quick postmortem here: #4928 was caused by the wrapped ObjectInfo actually returning the checksum & size of the uncompressed object which obviously differs from the compressed object actually upload (duh!). Since we cannot know the checksum & size of the compressed object beforehand there is really no useful information that can be returned here.

OK
 
> While wondering how I didn't catch that beforehand I found a second bug. Prior to [886b3ab](https://github.com/rclone/rclone/commit/886b3abac17404910d89192fae74cdec47515702) I was using operations.Rcat() instead of put when uploading compressed files which had the side effect of masking this bug since it didn't make use of the ObjectInfo wrapping. Well, turns out there was a reason I was using operations.Rcat() there since we cannot know the size of the compressed file beforehand (second duh!) this was required to be able to support underlying remotes without put stream. I've now implemented a Jottacloud-MD5 style workaround for this that spools the file to ram or disk to determine it's size and uploads using normal Put.

That seems like a reasonable work around :-)

I wonder whether we should factor the streaming to disk part of

- jottacloud backend
- rcat
- this new code

into `lib/readers`?
 
> I'd also strongly suggest adding additional integration tests for this remote similar chunker. At least one non-local file-based & one bucket based remote would probably be a good idea here @ncw.

Can you add the suggested remotes to `fstest/test_all/config.yaml` and paste the definitions here and I'll make sure they get into the integration test machine.

ncw (commented):
Code looks great! Please merge when you are ready.

buengese:
These are my compress test remotes. All with different underlying remote.

```
[TestCompress]
type = compress
remote = /tmp/rclone-compress-test
mode = gzip

[TestCompressSwift]
type = compress
remote = TestSwift:rclone-compress-test
mode = gzip

[TestCompressDrive]
type = compress
remote = TestDrive:rclone-compress-test
mode = gzip

[TestCompressS3]
type = compress
remote = TestS3:rclone-compress-test
mode = gzip
```

ncw:
> These are my compress test remotes. All with different underlying remote

Thanks. I've stuck those in the integration tester now.

PR #4953, opened 2021-01-24
fs: add support for flag --no-console on windows to hide the console …

albertony:

Add option to hide rclone's console window, with a new flag `--no-console`. Mostly relevant in combination with remote control, mount, web gui etc commands, but also long running sync operations etc, and especially when starting rclone automatically from task scheduler or windows service (nssm).

The implementation is based on Syncthing's:
https://github.com/syncthing/syncthing/commit/80dca96ee822d2f1a23cadb2a6868e1322a8d90b

Note: When starting `rclone` with `--no-console` from "run" dialog, desktop shortcut etc, the console window "flashes" briefly on startup, before it gets set to hidden. If you start up `cmd.exe` and then execute `rclone` with `--no-console` from there, it actually hides the existing cmd.exe window, and when rclone process exits the cmd process is left running hidden. This can perhaps be a bit surprising for users, but I do not know a way around it (this is how Syncthing does it as well).

PS: Should it be called --no-console or --hide-console, or something else? Syncthing used no-console, and rclone already have some --no-... flags so I kept that.

PS 2: It would be nice to include this in a doc page for the typical use case of auto-starting rclone for mount etc. The following from Syncthing is really detailed and well written and could serve as an inspiration: https://docs.syncthing.net/users/autostart.html

The actual option has not been discussed, but tips and workarounds on how to hide the console window has been discussed numerous times.

https://forum.rclone.org/t/how-to-mount-a-personal-jottacloud-as-a-drive-on-win10/21601/37
https://forum.rclone.org/t/autostart-mount-on-windows-without-cmd-popping-up/20896

#### Checklist

- [X] ~I have added documentation for the changes if appropriate.~

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -124,6 +124,7 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.StringArrayVarP(flagSet, &downloadHeaders, "header-download", "", nil, "Set HTTP header for download transactions")
 	flags.StringArrayVarP(flagSet, &headers, "header", "", nil, "Set HTTP header for all transactions")
 	flags.BoolVarP(flagSet, &ci.RefreshTimes, "refresh-times", "", ci.RefreshTimes, "Refresh the modtime of remote files.")
+	flags.BoolVarP(flagSet, &ci.NoConsole, "no-console", "", ci.NoConsole, "Hide console window. Supported on Windows only.")
 }
 
 // ParseHeaders converts the strings passed in via the header flags into HTTPOptions
--- a/lib/terminal/hidden_windows.go
+++ b/lib/terminal/hidden_windows.go
@@ -1,0 +1,19 @@
+// +build windows
+
+package terminal
+
+import (
+	"syscall"
+)
+
+// HideConsole hides the console window and activates another window
+func HideConsole() {
+	getConsoleWindow := syscall.NewLazyDLL("kernel32.dll").NewProc("GetConsoleWindow")
+	showWindow := syscall.NewLazyDLL("user32.dll").NewProc("ShowWindow")
+	if getConsoleWindow.Find() == nil && showWindow.Find() == nil {
+		hwnd, _, _ := getConsoleWindow.Call()
+		if hwnd != 0 {
+			showWindow.Call(hwnd, 0)
+		}
+	}
+}
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -382,6 +382,11 @@ func initConfig() {
 	// Finish parsing any command line flags
 	configflags.SetFlags(ci)
 
+	// Hide console window
+	if ci.NoConsole {
+		terminal.HideConsole()
+	}
+
 	// Load filters
 	err := filterflags.Reload(ctx)
 	if err != nil {
--- a/fs/config.go
+++ b/fs/config.go
@@ -121,6 +121,7 @@ type ConfigInfo struct {
 	DownloadHeaders        []*HTTPOption
 	Headers                []*HTTPOption
 	RefreshTimes           bool
+	NoConsole              bool
 }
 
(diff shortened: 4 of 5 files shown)


fs/config/configflags/configflags.go
```diff
@@ -124,6 +124,7 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.StringArrayVarP(flagSet, &downloadHeaders, "header-download", "", nil, "Set HTTP header for download transactions")
 	flags.StringArrayVarP(flagSet, &headers, "header", "", nil, "Set HTTP header for all transactions")
 	flags.BoolVarP(flagSet, &ci.RefreshTimes, "refresh-times", "", ci.RefreshTimes, "Refresh the modtime of remote files.")
+	flags.BoolVarP(flagSet, &ci.HideConsole, "no-console", "", ci.HideConsole, "Hide console window. Supported on Windows only.")
```
ncw, line 127: I try (I don't always succeed!) to make the internal names `HideConsole` agree with the flag names `--no-console` - so let's pick one or the other! Note that the internal names show in the `rc`.

lib/terminal/hidden_windows.go
```diff
@@ -0,0 +1,19 @@
+// +build windows
+
+package terminal
+
+import (
+	"syscall"
+)
+
+// HideConsole hides the console window and activates another window
```
ncw, line 9: Which windows gets activated? A random other window?

ncw (commented):
This is a nice feature :-)

See inline for a could of comments.

I thought you had to do this with linker flags...

> PS: Should it be called --no-console or --hide-console, or something else? Syncthing used no-console, and rclone already have some --no-... flags so I kept that.

Happy to go with prior art of syncthing

> PS 2: It would be nice to include this in a doc page for the typical use case of auto-starting rclone for mount etc. The following from Syncthing is really detailed and well written and could serve as an inspiration: https://docs.syncthing.net/users/autostart.html

Yes autostart could be a new page, or maybe a section in the `install.md` - nice idea. We could put systemd for mount and cron for syncing in there.

lib/terminal/hidden_windows.go
albertony: Who (*maybe* Microsoft) knows.. Its taken from documentation at https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow. When I test it no window is active after rclone's consoles has been hidden.

fs/config/configflags/configflags.go
albertony: Makes sense. Function is called `HideConsole`, and flag is called `no-console`, so internal name matched the function name instead of the flag name. Switched that. Could have renamed the function also to `NoConsole` but think it is more correct to call it `HideConsole` since that is what it *does*.

albertony:
Force-pushed an update according to your comments. Merge if you think its ok now, if not just let me know and I will be happy to do another iteration.

> This is a nice feature :-)

It will be a nice addition, and easier for new users, ref various forum threads on the subject. But its not at all perfect, it annoys me that the console window flashes on startup, but seemed like Syncthing had a couple of iterations on it without finding a better method, and I did not want to invest much more time myself.

> I thought you had to do this with linker flags...

I wanted to debug this issue related to mount, but since build failed I had to find something else to do that I could build without `-tags cmount`.. :)

ncw:
This looks great now - thank you!

I will merge it now.

Note that here: https://stackoverflow.com/a/23250506/164234 shows how to link the executable so the console window never appears. I don't think we want to do this but I thought you might like to see it if you haven't already.

ncw:
> I wanted to debug this issue related to mount, but since build failed I had to find something else to do that I could build without `-tags cmount`.. :)

What is broken? I think you posted on the forum about this but my memory is failing me here ;-)

albertony:
Yes, it was the mingw gcc error "export ordinal too large". Solved by adding "-buildmode exe". Thank you for that tip! :)

Issue #3340, opened 2019-07-11
Feature Request: Support for WinFsp Launcher


billziss-gh:
The question of how to setup rclone for use with the WinFsp Launcher often comes up in this forum and elsewhere. I am opening this issue primarily to address this question, but also to request that rclone natively supports the creation of the necessary registry entries to be used by the WinFsp Launcher.

#### What is your current rclone version (output from `rclone version`)?

(Via chocolatey)
```
rclone v1.48.0
- os/arch: windows/amd64
- go version: go1.12.3
```

## What problem are you are trying to solve?

Users often ask how to setup `rclone mount` so that they can "map" it as a network drive (via the Explorer or the `net use` command) or how to setup `rclone mount` to run under the SYSTEM account. The WinFsp Launcher can be used to support such scenarios; however setting up the Launcher for `rclone mount` is not straightforward, both because it requires users to make direct registry changes, but also because `rclone mount` does not understand the default command lines that the Launcher uses to launch file systems.

## How do you think rclone should be changed to solve that?

The most recent Beta release of WinFsp (2019.3 B2 / v1.5B2) has added the Launcher ability to rewrite paths passed to the file system during launch. For example this allows a UNC path of `\rclone\REMOTE\PATH\TO\FILES` to be rewritten to `REMOTE:PATH/TO/FILES` before being passed to `rclone mount`. This eliminates the need for `rclone mount` to support the native UNC syntax in its command line. I discuss this in **Path Rewriting** below.

Furthermore rclone could be enhanced to **have the ability to add the necessary Launcher registry entries during `rclone config`**. I discuss this in **Registry Entries** further below.

### Path Rewriting

For reference in the following, I am assuming that the user issed the command: `net use y: \\rclone\REMOTE\PATH\TO\FILES`. Assuming that the Launcher is properly configured to handle `\rclone` prefixes, it will pass the following arguments to the file system:

- `%1`: UNC path used in the `net use` command **with a single backslash**: `\rclone\REMOTE\PATH\TO\FILES`
- `%2`: drive used in the `net use` command: `y:`
- `%U`: user name of the user that issued the `net use` command in `DOMAIN\USERNAME` format: e.g. `WINDOWS\billziss`

The latest Launcher (2019.3 B2 / v1.5B2 or later) includes the ability to rewrite these paths in a format that `rclone mount` accepts:

[...]

PR #4972, opened 2021-01-27
Document autorun

albertony:

Documenting how to set up autostart of rclone, e.g. for mount or serve commands.

Mostly focus on Windows.

TODO:
- Currently appended to the install page. Don't know if it should be on a separate page.
- Windows:
  - Uncertainties around user accounts and mounts, see discussions in this PR comments!
  - Anyone with experience with running rclone as service or scheduled task? What works good/bad?
- Linux:
  - Did included a section for Linux, but could need some extension if anyone with experience wants to fill in...
- macOS
  - Nothing about mac included for now...

Sources:
- https://docs.syncthing.net/users/autostart.html#autostart-windows-startup

Related to this PR: https://github.com/rclone/rclone/pull/4953

#### Checklist

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -350,11 +350,9 @@ Command Prompt and then try to access the same drive from Explorer
 new drive.
 
 The easiest way around this is to start the drive from a normal
-command prompt. It is also possible to start a drive from the SYSTEM
-account (using [the WinFsp.Launcher
-infrastructure](https://github.com/billziss-gh/winfsp/wiki/WinFsp-Service-Architecture))
-which creates drives accessible for everyone on the system or
-alternatively using [the nssm service manager](https://nssm.cc/usage).
+command prompt. It is also possible to start a drive as the SYSTEM
+account, which creates drives accessible for everyone on the system,
+read more in the [install documentation](https://rclone.org/install/).
 
 ### Limitations
 
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -342,17 +342,38 @@ by specifying |-o FileSecurity="D:P(A;;FA;;;OW)"|, for file all access (FA) to t
 
 #### Windows caveats
 
-Note that drives created as Administrator are not visible by other
-accounts (including the account that was elevated as
-Administrator). So if you start a Windows drive from an Administrative
-Command Prompt and then try to access the same drive from Explorer
-(which does not run as Administrator), you will not be able to see the
-new drive.
+Drives created as Administrator are not visible to other accounts,
+not even an account that was elevated to Administrator with the
+User Account Control (UAC) feature. A result of this is that if you mount
+to a drive letter from a Command Prompt run as Administrator, and then try
+to access the same drive from Windows Explorer (which does not run as
+Administrator), you will not be able to see the mounted drive.
 
-The easiest way around this is to start the drive from a normal
-command prompt. It is also possible to start a drive as the SYSTEM
-account, which creates drives accessible for everyone on the system,
-read more in the [install documentation](https://rclone.org/install/).
+If you don't need to access the drive from applications running with
+administrative privileges, the easiest way around this is to always
+create the mount from a non-elevated command prompt.
+
+To make mapped drives available to the user account that created them
+regardless if elevated or not, there is a special Windows setting called
+[linked connections](https://docs.microsoft.com/en-us/troubleshoot/windows-client/networking/mapped-drives-not-available-from-elevated-command#detail-to-configure-the-enablelinkedconnections-registry-entry)
+that can be enabled.
+
+It is also possible to make a drive mount available to everyone on the system,
+by running the process creating it as the built-in SYSTEM account.
+There are several ways to do this: One is to use the command-line
+utility [PsExec](https://docs.microsoft.com/en-us/sysinternals/downloads/psexec),
+from Microsoft's Sysinternals suite, which has option |-s| to start
+processes as the SYSTEM account. Another alternative is to run the mount
+command from a Windows Scheduled Task, or a Windows Service, configured
+to run as the SYSTEM account. A third alternative is to use the
+[WinFsp.Launcher infrastructure](https://github.com/billziss-gh/winfsp/wiki/WinFsp-Service-Architecture)).
+Note that when running rclone as another user, it will not use
+the configuration file from your profile unless you tell it to
+with the [|--config|](https://rclone.org/docs/#config-config-file) option.
+Read more in the [install documentation](https://rclone.org/install/).
+
+Note that mapping to a directory path, instead of a drive letter,
+does not suffer from the same limitations.
 
 ### Limitations
 
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -12,6 +12,7 @@ Rclone is a Go program and comes as a single binary file.
   * [Download](/downloads/) the relevant binary.
   * Extract the `rclone` or `rclone.exe` binary from the archive
   * Run `rclone config` to setup. See [rclone config docs](/docs/) for more details.
+  * Optionally configure [automatic execution](#autostart).
 
 See below for some expanded Linux / macOS instructions.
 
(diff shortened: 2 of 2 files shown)


ncw:
I noted a stray `Syncthing` in the docs :-)

The Linux docs look a bit thin? Maybe we should ask @animosity22 to contribute a setting up rclone with systemd section?

Advice on how to avoid multiple rclone's running with cron is useful too.

ivandeex:
I encountered that rclone mounts on Windows tend to have rather long command line which doee not fit in the slim field of nssm dialog, but environment dialog is wide enough, so I prefer
```
RCLONE_OPT1=val1
...
```
to `rclone --opt1=val1 --opt2=val2 ........... remote:`

albertony:
Good tip, @ivandeex. I can include a note about this. I don't know if we should include even more details on how to configure it with nssm? You seem to have experience with rclone mount and nssm, are you running it as SYSTEM user then? When installing as "self-contained" service with powershell `new-service`, I were not able to make it create mounts that I could access without running the service as system user.

Same with task scheduler, except with the "Run only when user is logged on" mode where it runs in my interactive session. With the "Run whether user is logged on or not" it will always run elevated regardless of the "Run with highest privileges" option, and I can not access the mount from my user - not even from an elevated command prompt.

albertony:
> I noted a stray `Syncthing` in the docs :-)

No, impossible! ... after a quick amend and force-push :)

It was in the linux description, which I did not spend much time on. The Synchting guide had what seemed to be good descriptions for Linux and macOS as well, but I was kind of hoping someone else were willing to verify/adapt them to rclone (or write from scratch).. Syncthing has only the "syncing forever in background" mode, while rclone typically will do periodic syncing on schedule, and then also have mounting and all the other possible modes, so the description may not fit exactly.

ivandeex:
@albertony 

> are you running it as SYSTEM user then?

No, I'm running it with my interactive Windows user.
My nssm configuration normally looks like this:
- Application
  - Path: `C:\path\to\rclone.exe`
  - Startup directory: `C:\temp`
  - Arguments: `mount --volname Volume remote: X:`
- Details
  - Display name: `Mount-Name`
  - Startup: `Automatic (delayed start)`
- Logon
  - This user: `.\ivandeex`
  - Password: `super secret`
- I/O
  - Output: `C:\path\to\this-mount.log`
  - Error: `C:\path\to\this-mount.log` (same file)
- File rotation
  - Overwrite, Rotate, Rotate while service is running: `Yes`
  - Restrict rotation to file bigger than `1000000` bytes
- Environment
```
RCLONE_CONFIG=C:\path\to\rclone.conf
RCLONE_VERBOSE=1
RCLONE_VFS_CACHE_MODE=full
RCLONE_CACHE_DIR=C:\path\to\cache
... vfs and backend options ...
```
Currently I configure manually on Windows.
Later I will automate it with an Ansible role.

If I need to restart a mount, I use `Task Manager` (the `Services` tab).

> When installing as "self-contained" service with powershell `new-service`, I were not able to make it create mounts that I could access without running the service as system user.

It's because issue #3216 is still unresolved.
Someone will have to make rclone implement https://github.com/rclone/rclone/issues/3216#issuecomment-503880038
I'll pick it sometime in 2021 H2 unless someone else fixes it.

For now you should run Windows mounts as a normal user.

albertony:
> No, I'm running it with my interactive Windows user.

OK. It will run in session 0 as normal windows services I guess, so hidden, but does the rclone process run elevated with this setup?

> My nssm configuration normally looks like this:

Thanks for the details!

> It's because issue #3216 is still unresolved.

I'm not sure, but I think this is a somewhat different case. My case was that I could never find the mounts at all. Rclone would always run elevated, don't know if that is part of the reason.

> Someone will have to make rclone implement [#3216 (comment)](https://github.com/rclone/rclone/issues/3216#issuecomment-503880038)

Either that, or https://github.com/rclone/rclone/issues/3340, are not these two different ways to solving the same thing? And with https://github.com/rclone/rclone/issues/3340 it is actually possible without changes in rclone - but it takes quite some advanced registry "hacking"?

And isn't that primarily to be administrated as part of the WinFsp.Launcher, with its network provider allowing you to use "net use" instead of "rclone mount" commands? From my experience, even when installing as stand-alone service with `New-Service`, it will use the WinFsp.Launcher service to perform the mount (the DefineDosDevice API call), because WinFsp has code that checks `if running as service AND not LocalSystem THEN call WinFsp.Launcher via pipe api`. Or does also being fully integrated with WinFsp.Launcher have additional handling of user/permissions? (the issue mentions "I think I should add an -o username=USERNAME", and that has been done now, so that and https://github.com/rclone/rclone/issues/3340 makes me a bit confused what the current state of it is).

ivandeex:
> > No, I'm running it with my interactive Windows user.
> 
> OK. It will run in session 0 as normal windows services I guess, so hidden, but does the rclone process run elevated with this setup?

No, it's not elevated.

> My case was that I could never find the mounts at all. Rclone would always run elevated, don't know if that is part of the reason.

If I change account to SYSTEM or run rclone as elevated user, I hit the same issue.
So yes, it is the reason IMO

> > Someone will have to make rclone implement [#3216 (comment)](https://github.com/rclone/rclone/issues/3216#issuecomment-503880038)
> 
> Either that, or #3340, are not these two different ways to solving the same thing? And with #3340 it is actually possible without changes in rclone - but it takes quite some advanced registry "hacking"?

I see it this way (I may be **wrong**):
WinFSP (and its launcher) is a bridge between rclone and Windows "mount subsystem", which is full of legacy I'd prefer to not dig in if possible :disappointed_relieved: 
WinFSP provides a few solutions (_different ways_) to help with mounts in various scenarios:
- rclone runs in background as a "normal" non-elevated user **(this is my nssm case)**
- rclone runs in background (as a local System or elevated user) and feeds WinFSP with --UserName/--GroupName programmatically https://github.com/rclone/rclone/issues/3216#issuecomment-509412571
- rclone does not run in background but user can run "net use..." which trigger WinFSP _launcher_ which looks into registry and runs rclone with templated command line arguments https://github.com/rclone/rclone/issues/3340, and rclone can pass UserName/GroupName from those arguments back to WinFSP _proper_

albertony:
I just did a test with NSSM: It starts rclone elevated. And surprisingly (to me) WinFsp still detects it as running as a service, and uses the WinFsp.Launcher as I've described. After some reading it seems WinFsp is looking for process token WinServiceSid to detect if running as service, and rclone process has this token even when not being started directly as service but started as subprocess of nssm.exe. Starting from task scheduler it does not get this token (but instead WinBatchSid), and neither from command prompt (but instead WinInteractiveSid).

This means it uses WinFsp.Launcher, even when started from NSSM, which means the mounting (DefineDosDevice api call) is performed as SYSTEM user. So I can see the mounted drive as my user. 

Starting identical rclone mount command from task scheduler with "Run whether user is logged on or not", it also is started elevated and process seems identical to when started as service - but it will not use WinFsp.Launcher, and therefore perform the mount as the configured task user - elevated - and I can not see the mounted drive. I think that is the explanation of the difference, which I found extremely strange and annoying when comparing the two methods.

> No, it's not elevated.

Are you sure? If you stop the WinFsp.Launcher prosess, does your rclone mount from nssm still work?

> If I change account to SYSTEM or run rclone as elevated user, I hit the same issue.

When you tested rclone as elevated user, was that interactive then? I can't see anything to control this from nssm..

ivandeex:
> Are you sure? If you stop the WinFsp.Launcher prosess, does your rclone mount from nssm still work?

It works with WinFsp.Launcher service running or stopped.
Tried a minute ago: stop WinFsp.Lanucher, restart rclone nssm service, check mount and logs,
then start WinFsp.Launcher, restart rclone, check mount
Windows 8.1 Pro x64
`HELM\SOFTWARE\Wow6432Node\WinFsp` exists but is empty

> When you tested rclone as elevated user, was that interactive then? I can't see anything to control this from nssm..

SYSTEM user - on the nssm `Log on` tab switched to `Local System account` (I remember clearly)
Elevated user - ran rclone from elevated powershell console (it was more than a year ago, I forgot details)

albertony:
> It works with WinFsp.Launcher service running or stopped.

Strange, it is definitely opposite in my setup! Are you also on WinFsp version 2020.2?

> HKLM\SOFTWARE\Wow6432Node\WinFsp exists but is empty

In my case too, or actually it contains subkeys for built-in memfs, but nothing rclone related. But if I am correct, then an entry here is (only) needed when you want the service to be controlled by WinFsp, and then also get the "net use" integration etc. But the pipe api from WinFsp library (rclone) to the WinFsp.Launcher service to request it to perform the mounting (DefineDosDevice) is independent of the WinFsp registry configuration. It only checks that: 1) Process has the service token on it. 2) Is not running as SYSTEM user. And in my case both are true. The rclone.exe gets the service token, also when started by nssm.

ivandeex:
> > It works with WinFsp.Launcher service running or stopped.
> 
> Strange, it is definitely opposite in my setup! Are you also on WinFsp version 2020.2?

Only 2020.1, upgraded last year.

> > HKLM\SOFTWARE\Wow6432Node\WinFsp exists but is empty
> 
> In my case too, or actually it contains subkeys for built-in memfs, but nothing rclone related. But if I am correct, then an entry here is (only) needed when you want the service to be controlled by WinFsp _[...snip...]_

Probably it was created by WinFSP installer. I never edited it, only customized nssm services.
Maybe you have different Windows? Mine is 8.1 Pro.

ivandeex:
> It only checks that: 1) Process has the service token on it. 2) Is not running as SYSTEM user. And in my case both are true. The rclone.exe gets the service token, also when started by nssm.

I didn't dig this deep.

albertony:
> Only 2020.1

No changes between 2020.1 and 2020.2 relevant for this case, as far as I can see. I'm considering installing 2020.1 to verify, though..

> Maybe you have different Windows? Mine is 8.1 Pro.

Yes, I have Win 10 (1909). If WinFsp version is not the cause, then this should mean that in Win8.1 rclone does not get the service token, as shown by the following screenshot mix from SysInternals Process Explorer:

rclone

ivandeex:
@albertony 
I would like to check the "token", "elevation" and "session" thingies on my side. I'm not good at Windows. What would you like me to look at (besides process explorer)?

albertony:
> @albertony
> I would like to check the "token", "elevation" and "session" thingies on my side. I'm not good at Windows. What would you like me to look at (besides process explorer)?

Great! I was hoping to write a guide that describes exactly how to get it to work, but if there is so much differences between versions etc it is going to be hard - must either write much shorter/vaguer, or separate guides for different versions etc... More work than I was planning to invest..

These process/security internals are hard to grasp, so I am by no means an expert myself. Process Explorer is a great tool to get some insight into Windows! I don't know if you meant you had checked it or not: But if you are willing to check: Find nssm, and the rclone beneath it. On both of them: Double click to show properties. Go to Security tab. Find an entry for group name starting with "Mandatory Label\" and flag "Integrity". The last part of the name should be Low, Medium or High, where High is an indication of it running with full administrative privileges ("elevated"). Then see if there is an entry for group "NT AUTHORITY\SERVICE" and flag "Mandatory". This is the "token" given to processes running as services. Applications you start yourself will get the interactive token (NT AUTHORITY\SERVICE), and task scheduler uses batch token (NT AUTHORITY\BATCH). I am wondering if rclone is elevated or not, and if it has the service token or not. It would also be interesting to know the same for the nssm process.

From what I understand WinFsp checks the presence of "NT AUTHORITY\SERVICE" in the process, in our case rclone, and when this token is present it assumes it is running as service, and does not separate between it being running as a stand-alone service (like when I just installed it with New-Service command in PowerShell) or as part of the WinFsp.Launcher system.

ivandeex:
> I don't know if you meant you had checked it or not

Checked just now.

Process explorer

procexp-rclone

nssm process

nssm-props

rclone process

rclone-props

Windows version

win8-ver

Notes:
- "Обязательная метка" (rus) == "Mandatory label" (eng)
- "Высокий обязательный уровень" (cryptic) == "High mandatory level" (plain)
- "СЛУЖБА" == "SERVICE"

ivandeex:
Ah...
WinFSP version:

winfsp-ver

```
rclone version
rclone v1.54-iva01
- os/arch: windows/amd64
- go version: go1.13.4
```
(which is in fact vanilla rclone 1.54.0)

and this stuff works _for me_.

albertony:
Thank you for that info, I really appreciate it! Glad you included the english translations at the end! ;)

So, I conclude that your nssm and rclone processes both:
- Are elevated (High Integrity Level)
- Has service token (NT AUTHORITY\SERVICE)
- Run as your personal user (not the built-in SYSTEM / Local System account)

That matches what I see in my test!

But in my test it works *only* when WinFsp.Launcher service is running. When I stop WinFsp.Launcher the mount no longer works: The nssm-based service I created for rclone mount is running, but mount drive does not appear, rclone log file contains "Mount failed", and I see in process explorer that rclone.exe is being restarted by nssm.

I have WinFsp 2020.2. I think I tried to see through the changes between .1 and .2 without finding anything that stood out as relevant. I may have to uninstall and install 2020.1 to try it out, I can't think of anything else that could explain the difference..

albertony:
> I may have to uninstall and install 2020.1 to try it out, I can't think of anything else that could explain the difference..

Just did. No difference between WinFsp 2020.1 and 2020.2.

ncw:
Is this OK to merge now? Merge window for 1.55 approaching fast ;-)

albertony:
> Is this OK to merge now? Merge window for 1.55 approaching fast ;-)

No! 😞 

I'm not sure it is correct as is, and the observed differences in behavior without finding a good reason makes me rethink this entire doc, e.g. if it must be rewritten with much less detail, I'm not sure...

albertony:
> > Is this OK to merge now? Merge window for 1.55 approaching fast ;-)
> 
> No! 😞

Changed my mind!

**But:**

I decided to change direction a bit, and deleted the detailed step-by-step guide for scheduled task, and some other details, and most importantly any sections affected by the uncertainty around user account, elevation, service token discussed above!

My thinking is that I will *not* be able to complete the detail level as it was, and to clear out all uncertainties, within reasonable amount of time! And then there would be a lot of, hopefully, useful documentation that would also just rot in that big draft...

So, if we can make this PR into something that we want to merge on short term, then a new PR can be created to with intention of getting the uncertainties cleared out and fill in more details as needed...

@ncw, @ivandeex : What do you think? I have pushed a version now that should be ready for review. I will read through it a couple of times more myself, so might push some small updates still. Also, it is currently spread over several commits that I will squash later.

ncw:
> So, if we can make this PR into something that we want to merge on short term, then a new PR can be created to with intention of getting the uncertainties cleared out and fill in more details as needed...

Yes that sounds like a good plan. I'd rather have something now and fill in gaps later.

Will take a look now

ncw (approved):
I think those docs look very useful.

The Linux ones need filling out but I'm happy to merge this now.

Can you squash the commits before merging?

albertony:
Great!



> Can you squash the commits before merging?

Done!

ncw:
Merge when ready :-)

ivandeex:
@albertony 
There is a trick with github.
If you squash to one commit, it adds PR number to the commit message.
This allows to easily find full discussion thread from the commit.
But if you don't squash, github retains original messages and the link gets lost (unless you manually amend in advance).

albertony:
Ah, ok, sorry about that - and thanks for the tip.

Actually, I kind of knew about it... except I thought the commit messages should rather be "clean", for changelogs etc, so I have even removed the PR number when doing squash merges 🤦  

If I click on the commit, there is a link to the PR in small font just below the commit message, but that is not enough I take it?

PR #5001, opened 2021-02-08
mount: allow mounting to root directory on windows

albertony:

Fixed bug in mount on Windows introduced in 1.54.0, where mounting on root directory is no longer possible.

Edit: Added a second commit that also fixes a less critical issue regarding relative paths when mounting to directory. More details below. Let me know if you want me to move this to a separate MR or something..

##### Done

- Removed incorrect error return for mountpath that represents a directry at root level on existing drive.
- Also removed the distinction between `X:` and `X:\`: The latter will be converted to `X:` and used for drive letter mount, previously it would be reported as error "parent of mountpoint directory does not exist". Any other paths logically representing the same, such as `X:\some\folder\..\..` will still be reported as error: "parent of mountpoint directory does not exist".
- Edit (second commit): Calling filepath.Abs on the supplied mount path, which fixes handling of relative paths (see below). It is also better as the debug log message "Mounting on .." will show the actual path used, e.g. when mounting to `C:some\..\crazy\..\..\path` (which actually works)
##### Examples:

If C: exists and X: does not:

`rclone mount remote: C:` fails (mountpoint path already exists)
`rclone mount remote: C:\` fails (mountpoint path already exists)
`rclone mount remote: C:\MyMount` works (directory mounting)
`rclone mount remote: C:\test\abc\..\..\MyMount` works (same result as with `C:\MyMount`)

`rclone mount remote: X:` works (drive mounting)
`rclone mount remote: X:\` works (drive mounting)
`rclone mount remote: X:\MyMount` fails (parent of mountpoint directory does not exist)

##### ~Possible future~ Additional improvements (relative paths):

In the released version some relative paths work, but others that should/could does not:

`rclone mount remote: .` fails, but that is correct (mountpoint path already exists)
`rclone mount remote: MyMount` fails, which it should not (trying to find parent path fails)
`rclone mount remote: .\MyMount` fails, same as above
`rclone mount remote: ..\MyMount` works

Fixed this with the second commit, so any ©️ ®️ kind of relative path work as expected.

##### Bad excuses

Part of the issue were due to the following consideration: When user tries to mount to path D:\\", which is illegal (in WinFsp), should it be reported as an error to inform the user that this is illegal, or should rclone just assume the intention was to mount as a drive letter and fix it to "D:"? And then the other part was not understanding correctly how the filepath.Clean function, implicitely called from filepath.Join, manipulates the path (which is, by the way, the reason for using internal regex methods to classify the path) There *could* still be some remaining corner cases, depending on how os.Stat vs filepath.Clean etc handles different crazy variants...

https://forum.rclone.org/t/fatal-error-failed-to-mount-fuse/22119

#### Checklist

### DIFF
--- a/cmd/cmount/mountpoint_windows.go
+++ b/cmd/cmount/mountpoint_windows.go
@@ -114,11 +114,12 @@ func handleLocalMountpath(mountpath string, opt *mountlib.Options) (string, erro
 			fs.Errorf(nil, "Ignoring --network-mode as it is not supported with directory mountpoint")
 			opt.NetworkMode = false
 		}
-		parent := filepath.Join(mountpath, "..")
-		if parent == "" || parent == "." {
-			return "", errors.New("mountpoint parent path is not valid: " + parent)
+		var err error
+		if mountpath, err = filepath.Abs(mountpath); err != nil { // Ensures parent is found but also more informative log messages
+			return "", errors.Wrap(err, "mountpoint path is not valid: "+mountpath)
 		}
-		if _, err := os.Stat(parent); err != nil {
+		parent := filepath.Join(mountpath, "..")
+		if _, err = os.Stat(parent); err != nil {
 			if os.IsNotExist(err) {
 				return "", errors.New("parent of mountpoint directory does not exist: " + parent)
 			}
--- a/cmd/cmount/mountpoint_windows.go
+++ b/cmd/cmount/mountpoint_windows.go
@@ -103,8 +103,9 @@ func handleLocalMountpath(mountpath string, opt *mountlib.Options) (string, erro
 	} else if !os.IsNotExist(err) {
 		return "", errors.Wrap(err, "failed to retrieve mountpoint path information")
 	}
-	//if isDriveRootPath(mountpath) { // Assume intention with "X:\" was "X:"
-	//	mountpoint = mountpath[:len(mountpath)-1] // WinFsp needs drive mountpoints without trailing path separator
+	if isDriveRootPath(mountpath) { // Assume intention with "X:\" was "X:"
+		mountpath = mountpath[:len(mountpath)-1] // WinFsp needs drive mountpoints without trailing path separator
+	}
 	if !isDrive(mountpath) {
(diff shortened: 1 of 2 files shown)


ncw:
OK that looks good to me - I will merge it now - thank you :-)

> Bad excuses

No excuses necessary! I appreciate your help :-)

In fact you've made loads of great contributions to rclone - do you want to join the project? If so I'll send you an invite.

albertony:
> do you want to join the project? If so I'll send you an invite.

It would be an honour... I think - I don't really know what that means, if there are suddenly super high expectations and a ton of responsibilities? :)

ncw:
> > do you want to join the project? If so I'll send you an invite.
> 
> It would be an honour... I think - I don't really know what that means, if there are suddenly super high expectations and a ton of responsibilities? :)

It is mainly so you can be part of the rclone team. You also get the shiny rclone badge on your github profile!

I made a list of things maintainers might do in the [MAINTAINERS](https://github.com/rclone/rclone/blob/master/MAINTAINERS.md) doc.

But it is an open source project so it is completely up to you how much work you put in - no expectations at all :-)

I'll send you a link for the rclone organisation which is the first step if you want to carry on. That will give you membership of the rclone org and rights to do stuff to tickets and code on rclone/rclone. I'd expect you to send PRs for code changes for the moment. It will probably barrage you with notifications too, which I expect you'll want to disable!

fxzxmic:
Why did I fail in this situation: `rclone mount remote: C:\MyMount`
```
rclone mount remote: "D:/Storage"
Fatal error: failed to mount FUSE fs: mountpoint directory is at root: D:\
```

albertony:
I don't think this has been released yet. It will be in 1.55 due in a few days: https://github.com/rclone/rclone/issues/5162.

fxzxmic:
> I don't think this has been released yet. It will be in 1.55 due in a few days: #5162.

Sorry, I just found that it hasn't been released yet. Now I use beta version. Everything is normal.

albertony:
Great. Sorry for the inconvenience.

PR #5014, opened 2021-02-10
operations: Made copy and sync operations obey a RetryAfterError

ankur0493:
For the copy operation, if `Put()` or `Update()` functions of any backend are called, the backends use `CallNoRetry` for the actual upload operation since the retry is supposed to be done using the low level retry, not using the pacer so that the file may be re-opened. If a `pacer.retryAfterError` is returned by the backend, Rclone will not retry the error. This MR fixes that.

https://forum.rclone.org/t/copy-low-level-retry-does-not-respect-pacer-retryaftererror/22183

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -32,6 +32,7 @@ import (
 	"github.com/rclone/rclone/fs/object"
 	"github.com/rclone/rclone/fs/walk"
 	"github.com/rclone/rclone/lib/atexit"
+	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/random"
 	"github.com/rclone/rclone/lib/readers"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


fs/operations/operations.go
```diff
@@ -483,7 +484,14 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 			break
 		}
 		// Retry if err returned a retry error
+		var retry bool
 		if fserrors.IsRetryError(err) || fserrors.ShouldRetry(err) {
+			retry = true
+		} else if t, ok := pacer.IsRetryAfter(err); ok {
+			time.Sleep(t)
```
ncw, line 491: A debug message is probably a good idea here before the sleep as retry afters can be very long (5 minutes) and users get upset when things grind to a halt!

Something like this

```go
    fs.Debugf(src, "Sleeping for %v to obey Retry-After error: %v", t, err)
    time.Sleep(t)
```

ncw (commented):
This looks great - thanks!

I put a suggestion for a debug inline.

fs/operations/operations.go
ankur0493: Updated.

ncw (approved):
Perfect! I'll merge now - thanks :-)

Issue #3864, opened 2020-01-09
Windows: Can't create top level directories with a uuid UNC path


iseahound:
Hey Nick,
Windows has a lot of specific quirks, and this is another one of them 🙁.
This one is really minor (and only affects top level directories).

Using a uuid instead of a drive letter introduces an error only when creating top level directories. 

    2020/01/09 02:20:41 ERROR : string.txt: Failed to copy: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting files as there were IO errors
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting directories as there were IO errors
    2020/01/09 02:20:41 ERROR : Attempt 1/3 failed with 1 errors and: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
    2020/01/09 02:20:41 ERROR : string.txt: Failed to copy: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting files as there were IO errors
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting directories as there were IO errors
    2020/01/09 02:20:41 ERROR : Attempt 2/3 failed with 1 errors and: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
    2020/01/09 02:20:41 ERROR : string.txt: Failed to copy: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting files as there were IO errors
    2020/01/09 02:20:41 ERROR : Local file system at //?/Volume{a224c74f-da87-c295-3b90-cde0f64a29cc}/folder: not deleting directories as there were IO errors
    2020/01/09 02:20:41 ERROR : Attempt 3/3 failed with 1 errors and: mkdir \\?: The filename, directory name, or volume label syntax is incorrect.
[...]

PR #5018, opened 2021-02-11
mount: docs: add note about volume path syntax on windows

albertony:

Document that path syntax `\\?\Volume{b75e2c83-0000-0000-0000-602f00000000}\some\path` is not supported.

https://github.com/rclone/rclone/issues/3864

#### Checklist

### DIFF
--- a/docs/content/local.md
+++ b/docs/content/local.md
@@ -10,15 +10,17 @@ Local paths are specified as normal filesystem paths, e.g. `/path/to/wherever`, 
 
     rclone sync -i /home/source /tmp/destination
 
-Will sync `/home/source` to `/tmp/destination`
+Will sync `/home/source` to `/tmp/destination`.
 
-These can be configured into the config file for consistencies sake,
-but it is probably easier not to.
+For consistencies sake one can also configure a remote of type
+`local` in the config file, and access the local filesystem using
+rclone remote paths, e.g. `remote:path/to/wherever`, but it is probably
+easier not to.
 
 ### Modified time ###
 
 Rclone reads and writes the modified time using an accuracy determined by
-the OS.  Typically this is 1ns on Linux, 10 ns on Windows and 1 Second
+the OS. Typically this is 1ns on Linux, 10 ns on Windows and 1 Second
 on OS X.
 
 ### Filenames ###
@@ -109,15 +111,35 @@ These only get replaced if they are the last character in the name:
 Invalid UTF-8 bytes will also be [replaced](/overview/#invalid-utf8),
 as they can't be converted to UTF-16.
 
-### Long paths on Windows ###
+### Paths on Windows ###
 
-Rclone handles long paths automatically, by converting all paths to long
-[UNC paths](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#maxpath)
-which allows paths up to 32,767 characters.
+On Windows there are many ways of specifying a path to a file system resource.
+Both absolute paths like `C:\path\to\wherever`, and relative paths like
+`..\wherever` can be used, and path separator can be either
+`\` (as in `C:\path\to\wherever`) or `/` (as in `C:/path/to/wherever`).
+Length of these paths are limited to 259 characters for files and 247
+characters for directories, but there is an alternative extended-length
+path format increasing the limit to (approximately) 32,767 characters.
+This format requires absolute paths and the use of prefix `\\?\`,
+e.g. `\\?\D:\some\very\long\path`. For convenience rclone will automatically
+convert regular paths into the corresponding extended-length paths,
+so in most cases you do not have to worry about this (read more [below](#long-paths)).
 
-This is why you will see that your paths, for instance `c:\files` is
-converted to the UNC path `\\?\c:\files` in the output,
-and `\\server\share` is converted to `\\?\UNC\server\share`.
+Note that Windows supports using the same prefix `\\?\` to
+specify path to volumes identified by their GUID, e.g.
+`\\?\Volume{b75e2c83-0000-0000-0000-602f00000000}\some\path`.
+This is *not* supported in rclone, due to an [issue](https://github.com/golang/go/issues/39785)
+in go.
+
+#### Long paths ####
+
+Rclone handles long paths automatically, by converting all paths to
+[extended-length path format](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation), which allows paths up to 32,767 characters.
+
+This conversion will ensure paths are absolute and prefix them with
+the `\\?\`. This is why you will see that your paths, for instance
+`.\files` is shown as path `\\?\C:\files` in the output, and `\\server\share`
+as `\\?\UNC\server\share`.
 
 However, in rare cases this may cause problems with buggy file
 system drivers like [EncFS](https://github.com/rclone/rclone/issues/261).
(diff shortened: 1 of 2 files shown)


albertony:
@ivandeex, @iseahound, @ncw: Added a section with some more generic info about path format on Windows, where I included the note that volume syntax is not supported. But after I was done, I started wondering: This is in the mount docs, but the example in the issue was with sync, I think, so this path format "limitation" is in rclone overall? Should it be moved somewhere else, e.g. it could be relevant here https://rclone.org/docs/#path-to-dir, but that section is very short today so it would be sort of "hijacking" it. Could make it a much shorter notice, of course. Do you have an opinion?

ncw:
I think this is a limitation of the go runtime rather than rclone.

I forgot I made an issue about this already - you could link to it if you want: https://github.com/golang/go/issues/39785

> This is in the mount docs, but the example in the issue was with sync, I think, so this path format "limitation" is in rclone overall? Should it be moved somewhere else, e.g. it could be relevant here https://rclone.org/docs/#path-to-dir, but that section is very short today so it would be sort of "hijacking" it. Could make it a much shorter notice, of course. Do you have an opinion?

There is a section on this in the local backend: https://rclone.org/local/#long-paths-on-windows which could be expanded.

That might be the place for more explanation and we could link https://rclone.org/docs/#path-to-dir and the mount docs there maybe?

ncw:
Actually reading that again makes me think your doc is needed in the mount docs because mount doesn't convert paths to UNC paths unlike the local backend. It could, but that might confuse people even more.

albertony:
I will look into it. Thanks for the links!

PS: I think the use of term "UNC" is a bit confusing, both in code and documentation - but I guess also in other sources, including Microsoft's documentation! (sorry if I've mentioned this before, can't remember if I have just thought about it or also discussed it.. :)). Is `\\?\C:\Files` really an UNC path? "Close enough" in informal communication, but is it formally true and should be used in docs etc? I'm thinking no. Obviously `\\server\share` is an UNC path and `C:\Files` is not. And then you the alternative form `\\?\UNC\server\share`, which obviously is also UNC. I would (and did) call  `\\?\C:\Files` for "extended-length path format", or possible "long path", but not UNC. The `\\?\` prefix is a Windows API concept, to specify "raw" or "uninterpreted" paths into API functions (and there are other prefixes, such as `\\.\` and `\\??\`). UNC is a more of a standard naming scheme for network resources.

> To specify an extended-length path, use the "\\\\?\\" prefix. For example, "\\\\?\\D:\very long path".
> ...
> The "\\\\?\\" prefix can also be used with paths constructed according to the universal naming convention (UNC). To specify such a path using UNC, use the "\\\\?\\UNC\\" prefix.

(https://docs.microsoft.com/en-gb/windows/win32/fileio/maximum-file-path-limitation)

> [UNC] (...) specifies a common syntax to describe the location of a network resource, such as a shared file, directory, or printer.

(https://en.wikipedia.org/wiki/Path_(computing))

albertony:
> Actually reading that again makes me think your doc is needed in the mount docs because mount doesn't convert paths to UNC paths unlike the local backend.

What's the effect of this: That you cannot mount to a path longer than 260 characters? Or cannot access files/dirs with path longer than 260 characters in a mount?

ivandeex:
Looks like core devs have no time to dig this deeper right now. Marking as draft for the moment.

albertony:
I made an attempt at completing the text, just being a bit vague on the path length limit for mount that I was wondering about. Did "spread" the text between https://rclone.org/local, https://rclone.org/docs and https://rclone.org/commands/rclone_mount, and cross-linking them. Good enough @ncw and @ivandeex ? :satisfied:

docs/content/local.md
```diff
@@ -109,15 +111,31 @@ These only get replaced if they are the last character in the name:
...
+`\` (as in `C:\path\to\wherever`) or `/` (as in `C:/path/to/wherever`).
+Length of these paths are limited to 259 characters, but there is an
+alternative extended-length path format extending the limit to 32,767 characters.
+This format requires absolute paths and the use of prefix `\\?\`,
+e.g. `\\?\D:\some\very\long\path`. For convenience rclone will automatically
+convert regular paths into the corresponding extended-length paths,
+see below. Note that Windows supports using the same prefix to
+specify path to volumes identified by their GUID, e.g.
+`\\?\Volume{b75e2c83-0000-0000-0000-602f00000000}\some\path`,
+but this is *not* supported in rclone (due to [this] issue in go).
```
ncw, line 128: Is there a link missing at `[this]`?

cmd/mountlib/mount.go
```diff
@@ -356,6 +356,12 @@ infrastructure](https://github.com/billziss-gh/winfsp/wiki/WinFsp-Service-Archit
 which creates drives accessible for everyone on the system or
 alternatively using [the nssm service manager](https://nssm.cc/usage).
 
+Note also that filesystem paths in mounts are not automatically converted
```
ncw, line 359: I'm not quite sure what I meant by saying that filesystem paths in mount are not automatically converted...

We do since 28255f1bacb2b4c46a0fd002b7477a92e0144c5b use ~~UNC~~ extended path format files in the VFS cache.

As to whether the external interface to mount as in when you are using `Z:` on Windows works with extended path format - I expect it does but I don't know whether WInFSP supports it.

Can you check? If you mount a google drive - can you make file names with longer than 259 characters and access then from Windows. I suspect the answer will be Yes.

So I think we should proabably drop this paragraph until we know what we are talking about, as I certainly don't!

ncw (commented):
> I would (and did) call `\\?\C:\Files` for "extended-length path format", or possible "long path", but not UNC. The `\\?\` prefix is a Windows API concept, to specify "raw" or "uninterpreted" paths into API functions (and there are other prefixes, such as `\\.\` and `\\??\`). UNC is a more of a standard naming scheme for network resources.

Thanks for clearing that up. I'm going to blame Klaus Post for my confusion who wrote the original path mangling for Windows and called it UNC!

> > Actually reading that again makes me think your doc is needed in the mount docs because mount doesn't convert paths to UNC paths unlike the local backend.
> 
> What's the effect of this: That you cannot mount to a path longer than 260 characters? Or cannot access files/dirs with path longer than 260 characters in a mount?

I think neither of these are true any more even with the VFS cache. Some testing would be a good idea, but I suggest we drop that para.

> I made an attempt at completing the text, just being a bit vague on the path length limit for mount that I was wondering about. Did "spread" the text between https://rclone.org/local, https://rclone.org/docs and https://rclone.org/commands/rclone_mount, and cross-linking them. Good enough @ncw and @ivandeex ?

I think it is looking good.

I think we should remove the mount changes pending further testing

See inline for more!

docs/content/local.md
albertony: Oops, disappeared in a refactoring/rebase I think.. Fixed it.

cmd/mountlib/mount.go
albertony: I did some quick testing, and seems mount handles long paths without any apparant limitations. Including writing to paths longer than 259 characters within a mounted drive letter, and even even crazy things like  `rclone.exe mount [C:\total path of 299 characters] X:`. Has not done a very thorough test, so there could still be issues, but I don't think I will invest the time cover everything...

albertony:
> I think we should remove the mount changes pending further testing.

I agree. I did some quick tests, and long paths seems to work fine with mount (and cache). There may be corner cases, I did not try very hard to find any, and just dropped the mount docs instead, as suggested. 

Did som small tweaks in the text, and force pushed an update.

ncw (approved):
That looks great now - thank you!

I'll merge it now.

PR #5023, opened 2021-02-13
drive: optimize performance for --min-age, --max-age (recovered)

ivandeex:

The Google Drive API allows to place a clause like "modifiedTime > '2012-06-04T12:00:00'" in the query param, so the filterflags --max-age and --min-age can be applied directly at the directory listing phase rather than in a filter. This is extremely helpful when we want to do an incremental backup of a remote drive with a large number of files but the number of recently changed file is small.

This PR is recovered from dead PR #3806. See discussion there.

> Apologies for the delay in reviewing this.
> The approach looks very ingenious
> I'm a bit concerned about the muddling of the layers - however that is the way of most optimizations!
> I think there might be a better way of getting the info from the sync layer to the backend layer but I haven't figured it out yet.
> I don't think this optimization will work with `rclone sync --delete-excluded`

> We now have the machinery to do this patch properly. Namely we are passing the filter flags in the context. So the google drive backend could read the min age and max age directly from there.
> So I think it would be worth rescuing and adjusting to the new context use (calling `filter.GetConfig(ctx)`)

#### Checklist


ivandeex:
Includes go 1.16 build patch from #5036 to see real PR test result

@ncw 
Please review this one and a number of other PRs waiting for you.
Thank you

ivandeex:
Integration tests pass for me.

ivandeex:
This PR is recovered and reworked from another PR that was abondoned by author after waiting for review for several months. Two weeks later, the recovered PR is ignored again. Something is wrong here.

ivandeex:
waiting for review

fs/sync/sync.go
```diff
@@ -152,6 +152,8 @@ func newSyncCopyMove(ctx context.Context, fdst, fsrc fs.Fs, deleteMode fs.Delete
 	} else {
 		s.ctx, s.cancel = context.WithCancel(ctx)
 	}
+	// Record source name in context for aware backends (see list() in Google Drive)
+	fi.SrcFsName = fsrc.Name()
```
ncw, line 156: What you are trying to do is make the filters apply to just the backend in question.

However I don't like this approach - there could easily be multiple syncs running at once using the same `fi` block.

Perhaps what we need is an explicit marker to say to all backends - use the filters. This could be passed in the context.

This needs to pass down the value of the `includeAll` flag as used in `fs/walk` and `fs/march` - applying the flag to the context in those places will make it work for all backends.

So if we made a context variable `UseFilters` and applied it in those places in `fs/walk` and `fs/march` where we actually call `Fs.List` or `Fs.ListR` then this would fix it up.

ncw (commented):
Apologies for the delay in reviewing - every time I've looked at this PR something has seemed wrong, but I haven't been able to work out why!

I finally figured it out - see the inline comment.

I like the idea of this - it is a great idea and it is worth making work.

ncw:
I just had a thought about this - the same technique of passing the filters + a flag down the ctx chain could be used to avoid the local backend listing things which are excluded which causes annoyance quite often. For example if you make a dangling symlink, you can't exclude the error from it

```
$ ls -l
total 0
lrwxrwxrwx 1 ncw ncw 7 Apr 19 16:24 dangling-link -> nowhere

$ rclone ls -L --exclude dangling-link .
2021/04/19 16:24:49 ERROR : dangling-link: Listing error: symlink: stat /tmp/test/dangling-link: no such file or directory
2021/04/19 16:24:49 Failed to ls: symlink: stat /tmp/test/dangling-link: no such file or directory
```

ivandeex:
I last updated this pull request 3 months ago.
After 6 months of no code changes PRs become eligible for closing.

fs/sync/sync.go
ivandeex: @ncw 
Done. PTAL!

ivandeex: @ncw  Done! PTAL...

fs/march/march.go
```diff
@@ -76,13 +77,16 @@ type listDirFn func(dir string) (entries fs.DirEntries, err error)
 
 // makeListDir makes constructs a listing function for the given fs
 // and includeAll flags for marching through the file system.
-func (m *March) makeListDir(ctx context.Context, f fs.Fs, includeAll bool) listDirFn {
+// Note: this will optionally flag filter-aware backends!
+func (m *March) makeListDir(ctx context.Context, f fs.Fs, includeAll, useFilter bool) listDirFn {
```
ncw, line 81: I think `useFilter` should always be `!includeAll`...

Does that seem correct to you?

backend/drive/drive.go
```diff
@@ -829,11 +831,27 @@ func (f *Fs) list(ctx context.Context, dirIDs []string, title string, directorie
 	if filesOnly {
 		query = append(query, fmt.Sprintf("mimeType!='%s'", driveFolderType))
 	}
+
+	// Constrain query using filter if this remote is a sync/copy/walk source.
+	if fi, use := filter.GetConfig(ctx), filter.GetUseFilter(ctx); fi != nil && use {
+		if !fi.ModTimeFrom.IsZero() {
+			query = append(query, fmt.Sprintf("(modifiedTime >= '%s' or mimeType = '%s')",
+				fi.ModTimeFrom.Format("2006-01-02T15:04:05"), driveFolderType))
```
ncw, line 839: Is there no timezone in this time? Or is it guaranteed to be UTC? EIther way this doesn't seem quite right.

ncw (commented):
Nearly there - looking nice :-)

I put a couple of comments for thought inline.

fs/march/march.go
ivandeex: I'm afraid not. `includeAll` is common between march sides and should be passed downstairs unchanged. `useFilter` is `true` only on the source side. Do I miss something?

backend/drive/drive.go
ivandeex: Ouch! My bad!

fs/march/march.go
ivandeex: You, the call stack in `march` is so complex that I decided to treat such notions as "march side" and "include all" separately.
The "march side" (aka `useFilter` which is true on the source side)  is passed down the stack in context.
The `includeAll` term is passed as an argument like before. This has an advantage that I don't need to look in every place in the code where it's used.
Probably this can be rewritten differently. See my last commit

ivandeex: Note:

I didn't introduce the new context flag in every public march/walk function. Instead, I added them selectively in the functions called from `sync.Copy/Paste`, from `operations.List` and in `Walk`  which is called from some of them. I didn't touch e.g. `list.NewDirList` but modified march initialization instead.

I tagged involved functions by a comment line containing the word `filter-aware` so to simplify analysis a little.

ivandeex: To me the call stack in `march` was so complex that I decided to treat such notions as "march side" and "include all" separately:
- passed the "march side" (aka `useFilter` which is true on the source side)  down the stack in context.
- passed `includeAll` term is passed as an argument like before. This has an advantage that I don't need to look in every place in the code where it's used.

Now rewritten to your review. After that I had to modify the unit test (I made it in a sidecar commit) because now query constraining can happen even if drive remote appears on the right side (it's now controlled purely by the march's src/dst includeAll flags).

PTAL!

backend/drive/drive.go
ivandeex: PTAL

PR #5040, opened 2021-02-18
Docs: directory filter rules

edwardxml:

An attempt to make directory filter rules marginally less confusing inspired by reading 
https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone_hashsum.md

I think the filter documentation is in step with that commit. The prior position when, with `ListR`, --exclude /foo/ did not have an equivalent effect to to --exclude /foo/** was not properly explained in the rewrite I did.

The three bullet points identifying when directory recursion optimisation is applied to the listing trouble me. The two prior attempts at putting the code logic into words were as bad. Maybe something closer go code logic should be incorporated?

#### Was the change discussed in an issue or in the forum before.

No.

#### Checklist

### DIFF
--- a/docs/content/filtering.md
+++ b/docs/content/filtering.md
@@ -113,7 +113,7 @@ With `--ignore-case`
 
 ## How filter rules are applied to files
 
-Rclone path / file name filters are made up of one or more of the following flags:
+Rclone path/file name filters are made up of one or more of the following flags:
 
   * `--include`
   * `--include-from`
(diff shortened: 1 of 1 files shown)


albertony:
I think it looks good!

> The three bullet points identifying when directory recursion optimisation is applied to the listing trouble me. 

I can understand that, but this version is definitely an improvement.

> Maybe something closer go code logic should be incorporated?

Not sure I understand what you meant by this.. and if its something you wanted to discuss before merging, or if it was more of a comment? :flushed:

edwardxml:
> Not sure I understand what you meant by this.. and if its something you wanted to discuss before merging, or if it was more of a comment? 😳

Just thinking out loud whether there is a better way to present it. The logic has been written different ways round in the past but always ends up rather complex. Ultimately the logic has to be included because it can have a significant effect on transaction fees. I'm happy to see this merged as is.

albertony:
I see. I find it understandable as you made it now. Have read through it several times, and can't think of a better way to describe it.

Issue #3375, opened 2019-07-24
filter: exclude=d/ should behave alike what exclude=d/** does


pjw91:
As discussed in [the forum](https://forum.rclone.org/t/fast-list-changes-behavior-of-exclude/10946), `--exclude d/` and `--exclude 'd/**'` works differently and this difference causes `sync`/`copy` mistakenly delete some files on the remote when `--fast-list` is ON.

Making `--exclude=d/` works alike `--exclude=d/**` would be OK.

ncw:
I've made an implementation of this here

[v1.54.0-beta.4847.51e810828.fix-3375-filter-dir](https://beta.rclone.org/branch/fix-3375-filter-dir/v1.54.0-beta.4847.51e810828.fix-3375-filter-dir/) on branch [fix-3375-filter-dir](https://github.com/rclone/rclone/tree/fix-3375-filter-dir) (uploaded in 15-30 mins)

ivandeex:
Closed after 4 months of inactivity
@ncw appears that branch [fix-3375-filter-dir](https://github.com/rclone/rclone/tree/fix-3375-filter-dir) can be purged

ncw:
Thanks @ivandeex  - I've now merged the above and deleted the branch!

PR #5043, opened 2021-02-18
Docs: / versus /* or /**

edwardxml:

Attempt to put prominently in docs the rsync - rclone filter difference. I was drawn to this by https://github.com/rclone/rclone/issues/3375 because I'm not sure if @ncw 's Commit there changes the behaviour.

There have been several forum posts but the key prompt for me was 
https://github.com/rclone/rclone/issues/3375
#### Checklist

### DIFF
--- a/docs/content/filtering.md
+++ b/docs/content/filtering.md
@@ -14,6 +14,13 @@ lists; file age and size, or presence of a file in a directory. Bucket
 based remotes without the concept of directory apply filters to object
 key, age and size in an analogous way.
 
+* Rclone filter patterns differ from rsync. Whereas
+`/foo/` matches a directory foo and its
+contents in rsync, in rclone that it is a directory filter.
+Rclone directory filters can optimise listing of a remote.
+In rclone, to refer to a directory and its contents at a
+single level use `/foo/*` or `/foo/**` to recurse. *
+
 Rclone `purge` does not obey filters.
 
 To test filters without risk of damage to data, apply them to `rclone


edwardxml:
Not especially bright of me using a bullet point in the markup but hopefully the means of highlighting the breakout point can be dealt eith after the wording.

albertony:
I can't say if the meaning is correct or not, but if it is then its a good paragraph to have there.
Bullet point "list" with one entry, I don't know... Maybe instead of bullet point, just prefix it with "Note ..", or even "**Note**:"? That seems to be used a few places around the docs. Also "**Important**:" is used (https://rclone.org/commands/rclone_sync/ uses both), but I guess this is more of just a note than something "important"..

PS: I think there is a stray `*` at the end. Or did you actually intend surrounding with `*` to have everything in italics?

ncw:
The change I made in 8897377a541428b4a3f3be4e61d4d97b90fbfdaa is a half way house to `dir/` being exactly the same as `dir/**`

The whole `**` thing causes a lot of confusion in general, and with the benefit of hindsight I think `dir/` should mean exactly `dir/**`

From the commit

> Rclone uses directory exclusions to cut down the listing it has to do,
> so before this fix `--exclude dir/` would make sure nothing in `dir/`
> was scanned, **except** if --fast-list was used, in which case only
> the directory was excluded and everything within it was included.
> 
> This is rather unexpected, so this patch makes `--exclude dir/` be
> equivalent to `--exclude dir/**`, meaning that excluding a directory
> excludes it and its contents.
> 
> We can't do the same for --include without changing the semantics of
> filtering slightly.
> 

However making `--include dir/` be the same as `include dir/**` is a backwards incompatible change.

Now maybe I should make that change - I wasn't brave enough in that commit, but maybe I should do it and write it up in the change log.

```
$ tree .
.
├── dir1
│   ├── a
│   ├── b
│   └── c
├── dir2
│   ├── a
│   ├── b
│   └── c
└── dir3
    ├── a
    ├── b
    └── c

3 directories, 9 files
```

Demo that `--include dir/` and `--include dir/**` are different now

```
$ rclone lsf -R . --include "dir1/"
dir1/
```

```
$ rclone lsf -R . --include "dir1/**"
dir1/
dir1/a
dir1/b
dir1/c
$ 
$ 
```

Demo that `--exclude dir/` and `--exclude dir/**` are not different

```
$ rclone-v1.54.0 lsf -R . --exclude "dir1/"
dir2/
dir3/
dir3/a
dir3/b
dir3/c
dir2/a
dir2/b
dir2/c
```

```
$ rclone-v1.54.0 lsf -R . --exclude "dir1/**"
dir2/
dir3/
dir3/a
dir3/b
dir3/c
dir2/a
dir2/b
dir2/c
```

```
$ rclone lsf -R . --exclude "dir1/"
dir2/
dir3/
dir3/a
dir3/b
dir3/c
dir2/a
dir2/b
dir2/c
```

```
$ rclone lsf -R . --exclude "dir1/**"
dir2/
dir3/
dir3/a
dir3/b
dir3/c
dir2/a
dir2/b
dir2/c
```

Anyway to answer your question (finally!), I think your text is fine, but I'm not sure it should all be surrounded by `*`

edwardxml:
Thanks @albertony and @ncw I will look at this over the weekend.

The difference between rsync and rclone is confusing but I can see exactly why you chose to keep the `/foo/` syntax for listing optimisation which rsync doesn't need to bother with. I probably prefer the whole thing as it was but then I already went through the pain of figuring out using both rsync and rclone.

My only real concern is that the documentation exactly follows what the code does.

Yes, my whole * thing was a mess!!

edwardxml:
@ncw I am presuming `--exclude... /foo/` being interchangeable with `--exclude... /foo/**` except that the former can potentially optimise listing will apply from 1.54.1?

Should I expect the corresponding change for `--include` will be baked in from the same build?

Indeed would @ivandeev 's #5023 optimize of `--min-age` and `--max-age` listig on Drive be expected to land at the same time?

ncw:
> The difference between rsync and rclone is confusing but I can see exactly why you chose to keep the `/foo/` syntax for listing optimisation which rsync doesn't need to bother with. I probably prefer the whole thing as it was but then I already went through the pain of figuring out using both rsync and rclone.

:-)

I think if I was doing it again I would make `dir/` and `dir/**` be identical for include and exclude. Having to add `**` after directories is a major source of confusion for new users.

> The prospect of `--exclude` and `--include` behaving differently freaked me almost as much as a process breaking change.

I think this particular change will be almost unnoticeable, thankfully, otherwise I wouldn't have put it in.

> That said, at the end of the day my only real concern is that the documentation exactly follows what the code does.

:-)

> I am presuming `--exclude... /foo/` being interchangeable with `--exclude... /foo/**` (except the former can also potentially optimise listing) will apply from 1.54.1?

No from v1.55 (I don't put new features in point releases) and note that `--exclude foo/` and `--exclude foo/**` will now behave identically including optimising directory listings.

> Should I expect the corresponding change for `--include` will be baked in from the same build?

No, I haven't decided whether to make the change for `--include` - feedback needed!  The change for `--exclude` is safe and almost unoticable. To do the same change for `--include` would be noticeable and might break stuff.

> Indeed would @ivandeex 's #5023 optimize of `--min-age` and `--max-age` listing on Drive be expected to land at the same time?

That would be a v1.55 feature.

Note that the docs wouldn't go live until v1.55 either.

ncw:
Are you happy with the text @edwardxml ? If so I'll merge what we've got so far!

ivandeex:
@edwardxml ping

ivandeex:
This pull request is **abandoned for 3 months** since 2021-02-23

ivandeex:
@edwardxml
Friendly ping. Did you forget about your PR?

edwardxml:
> @edwardxml
> Friendly ping. Did you forget about your PR?

Hi @ivandeex I'm just absolutely bogged down with real life.

Not likely to return to it for a while but would then need to update myself with the issues again.

I could have no objection to anybody picking it up or closing the PR to clean up clutter.

All the very best.

ivandeex:
This PR is waiting for volunteers.
cc @ncw @albertony

ivandeex:
Closing abandoned pull requestd.

PR #5048, opened 2021-02-20
azureblob: Add container public access level support. Fixes #5045

### LINKED ISSUE #5045: Azure and mkdir command failed to set container public access level to blob or container
I think instead of keeping the access level to None there are two more options that it should support while using `mkdir` command.
In place of hardcode `PublicAccessNone` it should get the options from CLI and set default if missing. Other options are `PublicAccessBlob` &  `PublicAccessContainer`.

https://github.com/rclone/rclone/blob/7db68b72f12bfa71bd03f399c7e9302b2fc4fec7/backend/azureblob/azureblob.go#L1084

krmanish:
Add option to set public access level of a container with a new flag `azureblob-public-access`. Mostly useful when to create container using command `mkdir` however it should be useful for `sync` or `copy` command. Flag must require value and should be either `blob` or `container` or `''`(empty).
Flag is useful when the container is being created as part of the command. 

Discussed the actual option required for this change along with other important points.

https://forum.rclone.org/t/azure-container-and-set-public-access-to-read-only/22363
https://github.com/rclone/rclone/issues/5045

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -59,7 +59,6 @@ const (
 	emulatorBlobEndpoint = "http://127.0.0.1:10000/devstoreaccount1"
 	memoryPoolFlushTime  = fs.Duration(time.Minute) // flush the cached buffers after this long
 	memoryPoolUseMmap    = false
-	defaultPublicAccess  = azblob.PublicAccessNone
 )
 
 var (
@@ -219,15 +218,18 @@ This option controls how often unused buffers will be removed from the pool.`,
 				encoder.EncodeBackSlash |
 				encoder.EncodeRightPeriod),
 		}, {
-			Name: "public_access",
-			Help: `Public access level of a container: blob, container or private(default).
-
-This sets the public access level of a container. Default is none that means only authorized user can get access to the container.
-Other options are "blob" and "container" where
-container: Specifies full public read access for container and blob data. Clients can enumerate blobs within the container via anonymous request,
-but cannot enumerate containers within the storage account.
-blob: Specifies public read access for blobs. Blob data within this container can be read via anonymous request, but container data is not available.
-Clients cannot enumerate blobs within the container via anonymous request.`,
+			Name:    "public_access",
+			Help:    "Public access level of a container: blob, container.",
+			Default: string(azblob.PublicAccessNone),
+			Examples: []fs.OptionExample{
+				{
+					Value: "blob",
+					Help:  "Blob data within this container can be read via anonymous request.",
+				}, {
+					Value: "container",
+					Help:  "Allow full public read access for container and blob data.",
+				},
+			},
 			Advanced: true,
 		}},
 	})
@@ -526,9 +528,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			string(azblob.AccessTierHot), string(azblob.AccessTierCool), string(azblob.AccessTierArchive))
 	}
 
-	if opt.PublicAccess == "" {
-		opt.PublicAccess = string(defaultPublicAccess)
-	} else if !validatePublicAccess((opt.PublicAccess)) {
+	if !validatePublicAccess((opt.PublicAccess)) {
 		return nil, errors.Errorf("Azure Blob: Supported public access level are %s and %s",
 			string(azblob.PublicAccessBlob), string(azblob.PublicAccessContainer))
 	}
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -59,6 +59,7 @@ const (
 	emulatorBlobEndpoint = "http://127.0.0.1:10000/devstoreaccount1"
 	memoryPoolFlushTime  = fs.Duration(time.Minute) // flush the cached buffers after this long
 	memoryPoolUseMmap    = false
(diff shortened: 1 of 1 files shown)


backend/azureblob/azureblob.go
```diff
@@ -217,6 +218,17 @@ This option controls how often unused buffers will be removed from the pool.`,
 				encoder.EncodeDel |
 				encoder.EncodeBackSlash |
 				encoder.EncodeRightPeriod),
+		}, {
+			Name: "public_access",
+			Help: `Public access level of a container: blob, container or private(default).
+
```
ncw, line 224: Can you reformat this a bit please?

These docs should be hard wrapped at 80 characters so they view ok in terminals.

They are also markdown docs displayed in the config pages.

backend/azureblob/azureblob.go
```diff
@@ -217,6 +218,17 @@ This option controls how often unused buffers will be removed from the pool.`,
...
+This sets the public access level of a container. Default is none that means only authorized user can get access to the container.
+Other options are "blob" and "container" where
+container: Specifies full public read access for container and blob data. Clients can enumerate blobs within the container via anonymous request,
```
ncw, line 227: The docs for `container` etc might be better done as an examples list - this will let the user choose in the configurator and they will be shown in the docs, eg

https://github.com/rclone/rclone/blob/cc08f66dc178776b44faabdb90d1846a0c525fa7/backend/crypt/crypt.go#L39-L49

backend/azureblob/azureblob.go
```diff
@@ -217,6 +218,17 @@ This option controls how often unused buffers will be removed from the pool.`,
 				encoder.EncodeDel |
 				encoder.EncodeBackSlash |
 				encoder.EncodeRightPeriod),
+		}, {
+			Name: "public_access",
+			Help: `Public access level of a container: blob, container or private(default).
```
ncw, line 223: You don't need to write the default here - rclone will add it if it isn't `""`

backend/azureblob/azureblob.go
```diff
@@ -217,6 +218,17 @@ This option controls how often unused buffers will be removed from the pool.`,
...
+
+This sets the public access level of a container. Default is none that means only authorized user can get access to the container.
+Other options are "blob" and "container" where
+container: Specifies full public read access for container and blob data. Clients can enumerate blobs within the container via anonymous request,
+but cannot enumerate containers within the storage account.
+blob: Specifies public read access for blobs. Blob data within this container can be read via anonymous request, but container data is not available.
+Clients cannot enumerate blobs within the container via anonymous request.`,
+			Advanced: true,
```
ncw, line 231: Maybe add `Default: azblob.PublicAccessNone` - might need to cast that to a string - not sure what type `azblob.PublicAccessNone` is...

backend/azureblob/azureblob.go
```diff
@@ -499,6 +526,13 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			string(azblob.AccessTierHot), string(azblob.AccessTierCool), string(azblob.AccessTierArchive))
 	}
 
+	if opt.PublicAccess == "" {
```
ncw, line 529: If you add the `Default:` in the options above, then you don't need to test for `opt.PublicAccess == ""` as it will always be set.

I'm not sure of the actual value of `azblob.PublicAccessNone` - is it `""`

ncw (commented):
Thank you for doing this :-)

I put some notes inline, mostly about the docs!

backend/azureblob/azureblob.go
krmanish: Yes `azblob.PublicAccessNone` does resolve to `''`.

krmanish:
> Thank you for doing this :-)
> 
> I put some notes inline, mostly about the docs!

Thanks @ncw for the review comments. Those have been addressed.
Please re-review.

krmanish:
waiting for review

backend/azureblob/azureblob.go
```diff
@@ -217,6 +217,20 @@ This option controls how often unused buffers will be removed from the pool.`,
 				encoder.EncodeDel |
 				encoder.EncodeBackSlash |
 				encoder.EncodeRightPeriod),
+		}, {
+			Name:    "public_access",
+			Help:    "Public access level of a container: blob, container.",
+			Default: string(azblob.PublicAccessNone),
+			Examples: []fs.OptionExample{
+				{
```
ncw, line 225: This needs a "none" branch for no public access which should be the first option and the default.

ncw (commented):
I think this is almost done! I have a concern that the chooser is missing a "none" option - I think it should have one.

Apart from that perfect!

backend/azureblob/azureblob.go
krmanish: Since the default value is `''` ("None" branch resolves to empty string i.e `''`) hence wonder if this does sound okay? Please advise.

```
{
       Value: "",
	Help:  "The container and its blobs can be accessed only with an authorized request. This is the default value.",
}

```

ncw: I think using the `string(azblob.PublicAccessNone)` names is fine.

ncw (approved):

I'll merge it now.

PR #5049, opened 2021-02-21
dropbox: add polling support - fixes #2949

### LINKED ISSUE #2949: [Feature Request] Dropbox polling support
It would be great if polling could be added for dropbox as well. 
API polling support: https://www.dropbox.com/developers/documentation/http/documentation#files-list_folder-longpoll

wolveix:

This implements polling support for the Dropbox backend. The Dropbox SDK dependency had to be updated due to an auth issue, which was fixed on Jan 12 2021. A secondary internal Dropbox service was created to handle unauthorized SDK requests, as is necessary when using the ListFolderLongpoll function/endpoint. The config variable was renamed to cfg to avoid potential conflicts with the imported config package.

https://github.com/rclone/rclone/issues/2949

#### Checklist

### DIFF
--- a/backend/dropbox/dropbox.go
+++ b/backend/dropbox/dropbox.go
@@ -244,6 +244,7 @@ type Fs struct {
 	opt            Options        // parsed options
 	features       *fs.Features   // optional features
 	srv            files.Client   // the connection to the dropbox server
+	svc            files.Client   // the connection to the dropbox server (unauthorized)
 	sharing        sharing.Client // as above, but for generating sharing links
 	users          users.Client   // as above, but for accessing user information
 	team           team.Client    // for the Teams API
@@ -372,17 +373,21 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		opt:   *opt,
 		pacer: fs.NewPacer(ctx, pacer.NewDefault(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
 	}
-	config := dropbox.Config{
+	cfg := dropbox.Config{
 		LogLevel:        dropbox.LogOff, // logging in the SDK: LogOff, LogDebug, LogInfo
 		Client:          oAuthClient,    // maybe???
 		HeaderGenerator: f.headerGenerator,
 	}
 
+	// unauthorized config for endpoints that fail with auth
+	ucfg := dropbox.Config{
+		LogLevel: dropbox.LogOff, // logging in the SDK: LogOff, LogDebug, LogInfo
+	}
+
 	// NOTE: needs to be created pre-impersonation so we can look up the impersonated user
-	f.team = team.New(config)
+	f.team = team.New(cfg)
 
 	if opt.Impersonate != "" {
-
 		user := team.UserSelectorArg{
 			Email: opt.Impersonate,
 		}
@@ -397,12 +402,13 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			return nil, errors.Wrapf(err, "invalid dropbox team member: %q", opt.Impersonate)
 		}
 
-		config.AsMemberID = memberIds[0].MemberInfo.Profile.MemberProfile.TeamMemberId
+		cfg.AsMemberID = memberIds[0].MemberInfo.Profile.MemberProfile.TeamMemberId
 	}
 
-	f.srv = files.New(config)
-	f.sharing = sharing.New(config)
-	f.users = users.New(config)
+	f.srv = files.New(cfg)
+	f.svc = files.New(ucfg)
+	f.sharing = sharing.New(cfg)
+	f.users = users.New(cfg)
 	f.features = (&fs.Features{
 		CaseInsensitive:         true,
 		ReadMimeType:            false,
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/colinmarc/hdfs/v2 v2.2.0
 	github.com/coreos/go-semver v0.3.0
 	github.com/dop251/scsu v0.0.0-20200422003335-8fadfb689669
-	github.com/dropbox/dropbox-sdk-go-unofficial v5.6.0+incompatible
+	github.com/dropbox/dropbox-sdk-go-unofficial v1.0.1-0.20210114204226-41fdcdae8a53
```
ncw, line 25: This go.mod tag looks wrong - the latest version is 5.6 - it has the right commit hash though. Maybe it is just `go mod` doing mad stuff!

https://github.com/dropbox/dropbox-sdk-go-unofficial/tags

backend/dropbox/dropbox.go
```diff
@@ -1191,6 +1197,139 @@ func (f *Fs) About(ctx context.Context) (usage *fs.Usage, err error) {
...
+				if err != nil {
+					fs.Infof(f, "Change notify listener failure: %s", err)
+				}
+			}
+		}
+	}()
+}
+
+func (f *Fs) changeNotifyCursor() (cursor string, err error) {
+	fs.Debugf("", "changeNotifyCursor")
```
ncw, line 1251: Stray debug? Or make it proper debug `fs.Debugf(f, "Getting change notify cursor")`

backend/dropbox/dropbox.go
```diff
@@ -1191,6 +1197,139 @@ func (f *Fs) About(ctx context.Context) (usage *fs.Usage, err error) {
...
+}
+
+func (f *Fs) changeNotifyRunner(ctx context.Context, notifyFunc func(string, fs.EntryType), startCursor string) (newCursor string, err error) {
+	cursor := startCursor
+	var res *files.ListFolderLongpollResult
+
+	err = f.pacer.Call(func() (bool, error) {
+		args := files.ListFolderLongpollArg{
+			Cursor:  cursor,
+			Timeout: 60,
```
ncw, line 1271: Does this 60 need to be configurable? What happens at the timeout? You could read the rclone global timeout setting.

backend/dropbox/dropbox.go
```diff
@@ -1191,6 +1197,139 @@ func (f *Fs) About(ctx context.Context) (usage *fs.Usage, err error) {
...
+			return "", errors.Wrap(err, "list continue")
+		}
+		cursor = changeList.Cursor
+		var entryType fs.EntryType
+		for _, entry := range changeList.Entries {
+			entryPath := ""
+			switch info := entry.(type) {
+			case *files.FolderMetadata:
+				entryType = fs.EntryDirectory
+				entryPath = strings.TrimLeft(info.PathDisplay, f.slashRootSlash)
```
ncw, line 1310: I think this has the potential problem that `info.PathDisplay` doesn't reflect the actual case of the object - see notes here

https://github.com/rclone/rclone/blob/cc08f66dc178776b44faabdb90d1846a0c525fa7/backend/dropbox/dropbox.go#L7-L21

I'm pretty sure that dropbox hasn't created an files like this for > 5 years so I propose we don't worry about it here.

Possibly we should implement `ListR` for dropbox now too...

ncw (commented):
This is very nice code for a much requested feature - thank you :-)

I put some notes and queries inline.

go.mod
wolveix: Yep, blame Go Mod for that! I'll fix it up shortly.

backend/dropbox/dropbox.go
wolveix: Hah, yep, forgot to remove it!

backend/dropbox/dropbox.go
wolveix: Yes, this should really be configurable. I was going to await your feedback to ask whether a dedicated flag would be necessary or not, but the global flag should work just fine.

backend/dropbox/dropbox.go
wolveix: I based this off of your comment made here: https://github.com/rclone/rclone/issues/2949#issuecomment-752210127

I'll look into `ListR` while I make the other requested changes :)

backend/dropbox/dropbox.go
wolveix: @ncw could you point me in the right direction of how I could access the global timeout setting? I can't seem to find it.

go.mod
wolveix: @ncw I believe this may be due to the fact that the last GitHub release predates the commit needed, so `go mod` is using the latest commit which will drop the version to v1 for the repo rather than base it off of the release version (if that makes sense). I can't seem to get it to work any other way.

ncw: Ok! Well it is the right commit version anyway.

I can see the v5.6.0 tag on the `master` branch so it looks like `go mod`.

I wonder if it is worth reporting upstream?

backend/dropbox/dropbox.go
ncw: First make a `ci` variable in the `Fs` struct

https://github.com/rclone/rclone/blob/cc08f66dc178776b44faabdb90d1846a0c525fa7/backend/sftp/sftp.go#L257

Then populate it

https://github.com/rclone/rclone/blob/cc08f66dc178776b44faabdb90d1846a0c525fa7/backend/sftp/sftp.go#L480

Then access it with `f.ci.Timeout`

backend/dropbox/dropbox.go
ncw: I think what you've done is fine. And `ListR` for dropbox would probably be a good idea (in a new PR) :-)

backend/dropbox/dropbox.go
wolveix: Awesome, thanks! It's a Duration type rather than uint64, so I've had to divide it by `time.Second` and convert it to `uint64`. Additionally, Dropbox sets a hard limit of 480, so I've added a check that sets it to 480 if the global timeout is higher than that.

backend/dropbox/dropbox.go
wolveix: I may do that then :D

ncw (approved):

I'll merge it now

Issue #1763, opened 2017-10-21
Feature Request: Option to display sizes in base 10 as in MacOS Finder


dia3olik:
Hey Nick!

It would be great to have an rclone option to make it default to base 10 sizing instead just base 2,
either in the config file and/or as flag in the command line.

What do you think?

Cheers!!!!

IEC: KiB (1024 bytes), MiB (1024^2 bytes), GiB (1024^3 bytes), TiB (1024^4 bytes), ...
SI: kB (1000 bytes), MB (1000^2 bytes), GB (1000^3 bytes), TB (1000^4 bytes), ...

ncw:
Maybe an `--si` option?  `du` has such an option so there is precedent.

If I implemented that then I suppose I should change the units to show MiB and MB.  I'm not a big fan of the notation,but it is becoming a standard I think!

dia3olik:
no other supporter of this option? ;-)

ivandeex:
Related to https://github.com/rclone/rclone/issues/1890 (human readable sizes with `--human`)

PR #5053, opened 2021-02-21
docs: uppercase for MBytes

pvalls:
M**B**ytes and M**b**ytes are used interchangeably at docs.md.
Lowercase is preferably used  for bits(`Mbits`).

* Use uppercase consistently across all docs.md. Intances of `Mbytes` replaced to `MBytes`

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -363,7 +363,7 @@ This option controls the bandwidth limit. For example
 
     --bwlimit 10M
     
-would mean limit the upload and download bandwidth to 10 Mbyte/s.
+would mean limit the upload and download bandwidth to 10 MByte/s.
 **NB** this is **bytes** per second not **bits** per second. To use a
 single limit, specify the desired bandwidth in kBytes/s, or use a
 suffix b|k|M|G. The default is `0` which means to not limit bandwidth.
@@ -373,7 +373,7 @@ The upload and download bandwidth can be specified seperately, as
 
     --bwlimit 10M:100k
 
-would mean limit the upload bandwidth to 10 Mbyte/s and the download
+would mean limit the upload bandwidth to 10 MByte/s and the download
 bandwidth to 100 kByte/s. Either limit can be "off" meaning no limit, so
 to just limit the upload bandwidth you would use
 
@@ -402,7 +402,7 @@ working hours could be:
 `--bwlimit "08:00,512k 12:00,10M 13:00,512k 18:00,30M 23:00,off"`
 
 In this example, the transfer bandwidth will be set to 512kBytes/sec
-at 8am every day. At noon, it will rise to 10Mbytes/s, and drop back
+at 8am every day. At noon, it will rise to 10MBytes/s, and drop back
 to 512kBytes/sec at 1pm. At 6pm, the bandwidth limit will be set to
 30MBytes/s, and at 11pm it will be completely disabled (full speed).
 Anything between 11pm and 8am will remain unlimited.
@@ -412,8 +412,8 @@ An example of timetable with `WEEKDAY` could be:
 `--bwlimit "Mon-00:00,512 Fri-23:59,10M Sat-10:00,1M Sun-20:00,off"`
 
 It means that, the transfer bandwidth will be set to 512kBytes/sec on
-Monday. It will rise to 10Mbytes/s before the end of Friday. At 10:00
-on Saturday it will be set to 1Mbyte/s. From 20:00 on Sunday it will
+Monday. It will rise to 10MBytes/s before the end of Friday. At 10:00
+on Saturday it will be set to 1MByte/s. From 20:00 on Sunday it will
 be unlimited.
 
 Timeslots without `WEEKDAY` are extended to the whole week. So this
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -402,9 +402,9 @@ working hours could be:
 `--bwlimit "08:00,512k 12:00,10M 13:00,512k 18:00,30M 23:00,off"`
 
 In this example, the transfer bandwidth will be set to 512kBytes/sec
-at 8am every day. At noon, it will rise to 10MBytes/s, and drop back
+at 8am every day. At noon, it will rise to 10MByte/s, and drop back
 to 512kBytes/sec at 1pm. At 6pm, the bandwidth limit will be set to
-30MBytes/s, and at 11pm it will be completely disabled (full speed).
+30MByte/s, and at 11pm it will be completely disabled (full speed).
 Anything between 11pm and 8am will remain unlimited.
 
 An example of timetable with `WEEKDAY` could be:
@@ -412,7 +412,7 @@ An example of timetable with `WEEKDAY` could be:
 `--bwlimit "Mon-00:00,512 Fri-23:59,10M Sat-10:00,1M Sun-20:00,off"`
 
 It means that, the transfer bandwidth will be set to 512kBytes/sec on
-Monday. It will rise to 10MBytes/s before the end of Friday. At 10:00
+Monday. It will rise to 10MByte/s before the end of Friday. At 10:00
 on Saturday it will be set to 1MByte/s. From 20:00 on Sunday it will
 be unlimited.
 


edwardxml:
That could be a useful one to be consistent across all of the documentation.

albertony:
Should we then also think about clearifying which of Megabyte=MB=1000000 bytes vs Mebibyte=MiB=1048576 bytes is meant? ... or is it too :cop: ?

There is a note about [options which use SIZE](https://rclone.org/docs/#options), but when just browsing "random" docpages its easily missed:
> These are the binary units, e.g. 1, 2\*\*10, 2\*\*20, 2\*\*30 respectively.

I get the impression that "MiB" is getting some traction "out there" compared to "MB" for this, and that "MByte" is a lesser used form.

From wikipedia it seems the following is closest to standard:
MB=1000000 bytes
MiB=1048576 bytes
Mbit = 1000000 bits
Mibit = 1048576 bits

albertony:
There is also a mix of singular 10MByte/s and plural 10Mbytes/s, even within that same doc page. I'm not sure what is most correct?

pvalls:
Nice findings @albertony. As you mention, the docs note [options which use SIZE](https://rclone.org/docs/#options) states that units are in Base 2:
> "These are the binary units, e.g. 1, 2\**10, 2\**20, 2\**30 respectively."

Therefore, according to [Wikipedia's Data-rate-units](https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table) and [IEC 60027-2](http://webstore.iec.ch/webstore/webstore.nsf/mysearchajax?Openform&key=60027-2&sorting=&start=1&onglet=1/) "Letter symbols to be used in electrical technology standard", notation shall be in mebibyte per second (MiB/s). What do you think, do we change this?

pvalls:
> There is also a mix of singular 10MByte/s and plural 10Mbytes/s, even within that same doc page. I'm not sure what is most correct?

About this singular 10MByte/s and plural 10Mbytes/s duplicity, seems like the most correct or frequently used expression is singular:
* [Wikipedia's Data-rate-units](https://en.wikipedia.org/wiki/Data-rate_units#Megabyte_per_second) uses singular.
* Google's results count is 454,000 for ["MByte/s"](https://www.google.com/search?q=%22MByte/s%22) and 219,000 for ["MBytes/s"](https://www.google.com/search?q=%22MBytes/s%22).
* Most frequent usage in the docs.md is also in singular. (6 vs. 3 instances for each).

I'm resolving this duplicity to singular 👍

albertony:
> I'm resolving this duplicity to singular

Great, thanks!

> notation shall be in mebibyte per second (MiB/s). What do you think, do we change this?

My vote: :heavy_check_mark: :+1: :clap:
...but this is a probably a bit bigger decision, as I expect there to be a lot of occurrences all over the docs. What do you think, @edwardxml and @ncw ?

ncw:
> M**B**ytes and M**b**ytes are used interchangeably at docs.md.
> Lowercase is preferably used for bits(`Mbits`).
> 
> * Use uppercase consistently across all docs.md. Intances of `Mbytes` replaced to `MBytes`

I think you are correct and wikipedia agrees: https://simple.wikipedia.org/wiki/Byte

So bytes are `B` and `Byte` and bits are `b` and `bit`


> There is also a mix of singular 10MByte/s and plural 10Mbytes/s, even within that same doc page. I'm not sure what is most correct?

I think singular make most sense to me as a native English speaker.

I think that is the PR as-is so I will merge it now and comment on the other things in this thread next!

ncw:
> Should we then also think about clearifying which of Megabyte=MB=1000000 bytes vs Mebibyte=MiB=1048576 bytes is meant? ... or is it too  ?

Now there is a can of worms ;-)
 
> There is a note about [options which use SIZE](https://rclone.org/docs/#options), but when just browsing "random" docpages its easily missed:
> 
> > These are the binary units, e.g. 1, 2**10, 2**20, 2**30 respectively.

Yes, all SI prefixes in rclone are their binary versions so `k` == `1024`, `M` = 1048576 etc. Rclone is 100% consistent here.
 
> I get the impression that "MiB" is getting some traction "out there" compared to "MB" for this, and that "MByte" is a lesser used form.
> 
> From wikipedia it seems the following is closest to standard:
> MB=1000000 bytes
> MiB=1048576 bytes
> Mbit = 1000000 bits
> Mibit = 1048576 bits

Yes there is some traction to use `Mi` for 2<sup>20</sup> rather than 10<sup>6</sup>

I've spent most of my working career working with computers and telecom and between them they are very inconsistent! Hard disks are measured in GB whereas RAM is measured in GiB, but if you were to order a 2M telecom line you'll get a line with 2048000 bit/s of data capacity (a 2kikbit/s line?), but 10M ethernet is 1000000 bit/s.

I personally find the MiB units a little ugly, but it does appear to be an established standard now

> Therefore, according to [Wikipedia's Data-rate-units](https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table) and [IEC 60027-2](http://webstore.iec.ch/webstore/webstore.nsf/mysearchajax?Openform&key=60027-2&sorting=&start=1&onglet=1/) "Letter symbols to be used in electrical technology standard", notation shall be in mebibyte per second (MiB/s). What do you think, do we change this?

If we do change this, then my question would be - are we changing only the docs or the code also?

Rclone shows bytes and bytes/s with `k`, `M`, `G`, `T` suffixes. However it doesn't usually show units, so if you are transferring a file

For example stats display after transferring a file says this..

```
Transferred:   	        1G / 1 GBytes, 100%, 345.632 MBytes/s, ETA 0s
Elapsed time:         2.9s
```

This could be changed to

```
Transferred:   	        1G / 1 GiByte, 100%, 345.632 MiByte/s, ETA 0s
Elapsed time:         2.9s
```

I'll just note that changing rclone's output stands the risk of breaking peoples apps...

When configuring things the unit isn't normally used, so rclone will say `--max-size 4M` - I'm not sure I want to change this to `--max-size 4Mi` - I think `4M` here is an abbreviation for `4MiByte` or `4MiByte/s` and that is fine.

Here is a quick grep as to things which might need to change

```
$ git grep '[kMGTP]Byte' | grep -v MANUAL | grep -v rclone.1
cmd/delete/delete.go:For example, to delete all files bigger than 100MBytes, you may first want to check what
cmd/delete/delete.go:delete all files bigger than 100MBytes.
docs/content/commands/rclone_delete.md:For example, to delete all files bigger than 100MBytes, you may first want to check what
docs/content/commands/rclone_delete.md:delete all files bigger than 100MBytes.
docs/content/docs.md:Options which use SIZE use kByte by default.  However, a suffix of `b`
docs/content/docs.md:for bytes, `k` for kBytes, `M` for MBytes, `G` for GBytes, `T` for
docs/content/docs.md:TBytes and `P` for PBytes may be used.  These are the binary units, e.g.
docs/content/docs.md:would mean limit the upload and download bandwidth to 10 MByte/s.
docs/content/docs.md:single limit, specify the desired bandwidth in kBytes/s, or use a
docs/content/docs.md:would mean limit the upload bandwidth to 10 MByte/s and the download
docs/content/docs.md:bandwidth to 100 kByte/s. Either limit can be "off" meaning no limit, so
docs/content/docs.md:this would limit the upload bandwidth to 10MByte/s but the download
docs/content/docs.md:In this example, the transfer bandwidth will be set to 512kBytes/sec
docs/content/docs.md:at 8am every day. At noon, it will rise to 10MByte/s, and drop back
docs/content/docs.md:to 512kBytes/sec at 1pm. At 6pm, the bandwidth limit will be set to
docs/content/docs.md:30MByte/s, and at 11pm it will be completely disabled (full speed).
docs/content/docs.md:It means that, the transfer bandwidth will be set to 512kBytes/sec on
docs/content/docs.md:Monday. It will rise to 10MByte/s before the end of Friday. At 10:00
docs/content/docs.md:on Saturday it will be set to 1MByte/s. From 20:00 on Sunday it will
docs/content/docs.md:to use half of it - 5 Mbit/s. This is 5/8 = 0.625MByte/s so you would
docs/content/docs.md:For example use this to allow no transfers to be faster than 1MByte/s
docs/content/drive.md:files may be transferred much faster at 100s of MBytes/s but lots of
docs/content/filtering.md:Default units are `kBytes` but abbreviations `k`, `M`, or `G` are valid.
docs/content/filtering.md:E.g. `rclone ls remote: --min-size 50k` lists files on `remote:` of 50kByte
docs/content/filtering.md:Default units are `kBytes` but abbreviations `k`, `M`, or `G` are valid.
docs/content/filtering.md:E.g. `rclone ls remote: --max-size 1G` lists files on `remote:` of 1GByte
docs/content/filtering.md:All files on `B:` which are less than 50 kBytes are deleted
docs/content/flags.md:      --bwlimit BwTimetable                  Bandwidth limit in kBytes/s, or use suffix b|k|M|G or a full timetable.
docs/content/flags.md:      --bwlimit-file BwTimetable             Bandwidth limit per file in kBytes/s, or use suffix b|k|M|G or a full timetable.
fs/config/configflags/configflags.go:	flags.FVarP(flagSet, &ci.BwLimit, "bwlimit", "", "Bandwidth limit in kBytes/s, or use suffix b|k|M|G or a full timetable.")
fs/config/configflags/configflags.go:	flags.FVarP(flagSet, &ci.BwLimitFile, "bwlimit-file", "", "Bandwidth limit per file in kBytes/s, or use suffix b|k|M|G or a full timetable.")
fs/sizesuffix_test.go:		{1024, "1 kBytes"},
fs/sizesuffix_test.go:		{1024 * 1024, "1 MBytes"},
fs/sizesuffix_test.go:		{1024 * 1024 * 1024, "1 GBytes"},
fs/sizesuffix_test.go:		{10 * 1024 * 1024 * 1024, "10 GBytes"},
fs/sizesuffix_test.go:		{10.1 * 1024 * 1024 * 1024, "10.100 GBytes"},
fs/sizesuffix_test.go:		{10 * 1024 * 1024 * 1024 * 1024, "10 TBytes"},
fs/sizesuffix_test.go:		{10 * 1024 * 1024 * 1024 * 1024 * 1024, "10 PBytes"},
fs/sizesuffix_test.go:		{1 * 1024 * 1024 * 1024 * 1024 * 1024 * 1024, "1024 PBytes"},
```

Note that the code here would need to change

```
$ git grep '\.Unit'
cmd/size/size.go:                       fmt.Printf("Total size: %s (%d Bytes)\n", fs.SizeSuffix(results.Bytes).Unit("Bytes"), results.Bytes)
cmd/tree/tree.go:       flags.BoolVarP(cmdFlags, &opts.UnitSize, "human", "", false, "Print the size in a more human readable way.")
fs/accounting/stats.go:         fs.SizeSuffix(totalSize).Unit("Bytes"),
fs/accounting/stats.go:         fs.SizeSuffix(displaySpeed).Unit(strings.Title(s.ci.DataRateUnit)+"/s"),
fs/sizesuffix_test.go:          got := ss.Unit("Bytes")
```

And if we want to get "bits" vs "Bytes" right in the code then there is some work to do here

```
$ git grep DataRateUnit
cmd/cmd.go:             ci.DataRateUnit = "bytes"
cmd/cmd.go:             ci.DataRateUnit = *dataRateUnit
fs/accounting/accounting.go:    if acc.ci.DataRateUnit == "bits" {
fs/accounting/stats.go: if s.ci.DataRateUnit == "bits" {
fs/accounting/stats.go:         fs.SizeSuffix(displaySpeed).Unit(strings.Title(s.ci.DataRateUnit)+"/s"),
fs/config.go:   DataRateUnit           string
fs/config.go:   c.DataRateUnit = "bytes"
```

@albertony and @edwardxml is it time for a style guide for rclone docs now? This could start off as a section in the CONTRIBUTING.md guide. In particular I'm thinking of `e.g.` vs `eg` and things like that.

albertony:
Great "write-up", and I totally agree:

> Now there is a can of worms ;-)

> I personally find the MiB units a little ugly, but it does appear to be an established standard now

> When configuring things the unit isn't normally used, so rclone will say --max-size 4M - I'm not sure I want to change this to --max-size 4Mi - I think 4M here is an abbreviation for 4MiByte or 4MiByte/s and that is fine.

I noticed for instance that pacman on Arch has done the change, and from annoyed user on [forum](https://bbs.archlinux.org/viewtopic.php?id=133786) (which I will not be surprised if there will be with rclone as well) it seems it was done already in 2012:
> pacman -Syu
:: Synchronizing package databases...
 testing                21.8 KiB  7.09 MiB/s 00:00 [######################] 100%
 core                  133.8 KiB   257 KiB/s 00:01 [######################] 100%
 extra                1629.5 KiB  1045 KiB/s 00:02 [######################] 100%
 
I see wget still reports MB/s, so I'm sure there is a good mix, and perhaps even a minority using the MiB format still, but more and more..

> I'll just note that changing rclone's output stands the risk of breaking peoples apps...

That is an important point to consider. I would not expect it to cause a lot of problems, but that is just my guess...

I still personally think it is close enough to being an established standard, to stick with my vote for change! :)

ncw:
@albertony I agree with your points - I think we should change probably (I'm pretty sure there is an issue about it)

There is https://github.com/rclone/rclone/issues/1763 which proposes an `--si` unit to switch from MiB to MB. We need to be extremely careful with that as the k/M/G are mixed up with the config and you really don't want `chunk_size 16M` in a config file to suddenly be a different size. So I don't think I'm in favor of that in general.

albertony:
@ncw I am not able to find any existing issue specifically for this, although there are a few touching in on the subject, so I decided to create a new: https://github.com/rclone/rclone/issues/5068.

Issue #5029, opened 2021-02-16
Zoho - Invalid Client


c4tom:
First of all, thanks for adding this zoho option.

I tried to configure a zoho remote drive, with this new version 1.5.4 and I can test 2 types of zoho licenses, Zoho One and Zoho Workplace, and both, I got the message

```
Invalid Client

The transmitted client ID does not exist
```

> Storage> 39 
> ** See help for zoho backend at: https://rclone.org/zoho/ **
> 
> Zoho region to connect to. You'll have to use the region you organization is registered in.
> Enter a string value. Press Enter for the default ("").
> Choose a number from below, or type in your own value
>  1 / United states / Global
>    \ "com"
>  2 / Europe
>    \ "eu"
>  3 / India
>    \ "in"
>  4 / Australia
>    \ "com.au"
> region> 1
> Edit advanced config? (y/n)
> y) Yes
> n) No (default)
> y/n> 
> Remote config
> Use auto config?
>  * Say Y if not sure
>  * Say N if you are working on a remote or headless machine
> y) Yes (default)
> n) No
> y/n> 
> If your browser doesn't open automatically go to the following link: http://127.0.0.1:53682/auth?state=wF0U_y_ZIGLZycBnv2zqxA
> Log in and authorize rclone for access
> Waiting for code...

ncw:
I wonder if the client_id's are regional? We've been testing with the EU region.

What do you think @buengese

c4tom:
> I wonder if the client_id's are regional? We've been testing with the EU region.
> 
> What do you think @buengese

i tried with EU, but, not work

https://accounts.zoho.com/oauth/v2/auth?access_type=offline&client_id=**1000.OZNFWW075EKDSIE1R42HI9I2SUPC9A**&prompt=consent&redirect_uri=http%3A%2F%2Flocalhost%3A53682%2F&response_type=code&scope=aaaserver.profile.read+WorkDrive.team.READ+WorkDrive.workspace.READ+WorkDrive.files.ALL&state=FqjKOuka0obkE8tS3h2KKg

Note: is client_id=**1000.OZNFWW075EKDSIE1R42HI9I2SUPC9A** generated?


Note 2: I followed the example of the page https://rclone.org/zoho/, but there is no place in this version, which asks

client_id>
client_secret>

buengese:
[...]

PR #5059, opened 2021-02-22
zoho: fix custom client id's

buengese:
Fix setting custom client id's for zoho as discussed in #5029


ncw:
That lookks great thanks.

I'll put in in the point release also.

PR #5079, opened 2021-03-01
build: Update GitHub Actions to 'ubuntu-20.04'

InspectorWilliamHenderson:
#### What Is The Purpose Of This Change?

GitHub Actions currently runs on 'ubuntu-latest' (18.04/20.04), whereas the Latest One is Available by Specifically Stating the runs on to 'ubuntu-20.04'.

The GitHub Actions 'ubuntu-latest' is Currently in Migration Phase, and might take Several Weeks to Completely Update to Ubuntu 20.04 but Explicitly Stating Ubuntu Version to 20.04, assures Latest Ubuntu Version Every Time GitHub Actions in Initiated.

This PR changes that in Workflow Files.

#### Was The Change Discussed In An Issue or In The Forum Before?

No

#### Checklist

- [x] I have read the [Contribution Guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [ ] I have Added Tests for All Changes in This PR, If Appropriate.
- [ ] I have Added Documentation for The Changes If Appropriate.
- [x] All Commit Messages are In [House Style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [x] I'm Done, This Pull Request is Ready For Review :-)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
 
         include:
           - job_name: linux
-            os: ubuntu-latest
+            os: ubuntu-20.04
             go: '1.16.x'
             gotags: cmount
             build_flags: '-include "^linux/"'
@@ -70,25 +70,25 @@ jobs:
             deploy: true
 
           - job_name: other_os
-            os: ubuntu-latest
+            os: ubuntu-20.04
             go: '1.16.x'
             build_flags: '-exclude "^(windows/|darwin/|linux/)"'
             compile_all: true
             deploy: true
 
           - job_name: go1.13
-            os: ubuntu-latest
+            os: ubuntu-20.04
             go: '1.13.x'
             quicktest: true
 
           - job_name: go1.14
-            os: ubuntu-latest
+            os: ubuntu-20.04
             go: '1.14.x'
             quicktest: true
             racequicktest: true
 
           - job_name: go1.15
-            os: ubuntu-latest
+            os: ubuntu-20.04
             go: '1.15.x'
             quicktest: true
             racequicktest: true
@@ -125,7 +125,7 @@ jobs:
           sudo chmod 666 /dev/fuse
           sudo chown root:$USER /etc/fuse.conf
           sudo apt-get install fuse libfuse-dev rpm pkg-config
-        if: matrix.os == 'ubuntu-latest'
+        if: matrix.os == 'ubuntu-20.04'
 
       - name: Install Libraries on macOS
         shell: bash
@@ -204,7 +204,7 @@ jobs:
       - name: Deploy built binaries
         shell: bash
         run: |
-          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then make release_dep_linux ; fi
+          if [[ "${{ matrix.os }}" == "ubuntu-20.04" ]]; then make release_dep_linux ; fi
           if [[ "${{ matrix.os }}" == "windows-latest" ]]; then make release_dep_windows ; fi
           make ci_beta
         env:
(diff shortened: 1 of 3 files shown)


darthShadow:
What is the benefit of doing this ahead of time if migration is already on-going?

It only adds to the maintenance burden by needing to change the tag manually for every ubuntu version as they are released instead of just using `latest` and letting it upgrade automatically.

InspectorWilliamHenderson:
Well, They have been updating since October 2020, and At That Time, They said It will be done in Several Weeks.

It has been about 20 Weeks since then , and I am pretty sure It's still in Migration Process, because In My Repository GitHub Actions, I get Ubuntu 18.04 Every Single Time, even though They said You will get a Random One out of Ubuntu 18.04 / 20.04 if you are using Latest Tag.

And, About the Improvement Part, It's always beneficial to use the latest Ubuntu Version and About the Burden, It will have to be changed almost once in Every 2 Years so not much considering we update other packages too in the Workflow Build File, It can be easily updated alongside those.

InspectorWilliamHenderson:
Sorry, Accidentally Clicked Close with Comment Button.

darthShadow:
> Well, They have been updating since October 2020, and At That Time, They said It will be done in Several Weeks.
> 
> It has been about 20 Weeks since then , and I am pretty sure It's still in Migration Process

Per the latest comments on https://github.com/actions/virtual-environments/issues/1816, the original ETA was several weeks from November 30th which has slipped due to capacity issues to the 20s this month, which is not very far from now.

> , because In My Repository GitHub Actions, I get Ubuntu 18.04 Every Single Time, even though They said You will get a Random One out of Ubuntu 18.04 / 20.04 if you are using Latest Tag.

Checking the actions runs for this PR, all the ones I have checked seem to be 20.04 which lends credence to the fact that they are close to finishing the migration.

> About the Improvement Part, It's always beneficial to use the latest Ubuntu Version

We also don't depend much on any specific OS features (other than requiring it to be relatively new) so the benefits (if any) should be negligible.

> It will have to be changed almost once in Every 2 Years

Every small thing like this adds up to become a larger effort which is especially problematic due to the small dev team and the backlog of much more important issues.

---

I still believe we should stick to `latest` but I will leave the final decision to @ivandeex / @ncw.

ivandeex:
> I still believe we should stick to latest but I will leave the final decision to ivandeex / ncw.

I don't have an opinion on this, will trade my ballot to ncw :)

ncw:
I was thinking about this recently actually!

From a developer point of view `-latest` is great and that is why I made the file like that originally.

However from a reproducability point of view it isn't. In particular I was thinking about building the stable branch some time in the future - if we assume that the 20.04 update broke something then that is one more thing we'd have to backport.

However GItHub actions isn't written in stone and we quite often have to back port fixes to GitHub actions to the stable branch - examples from v1.54-stable - 98dbbc78abb0660ef63cef9a2ca247023adf062f cfc5d76fcabae8bdc92a0523ccde99c8d2b5f009 0af493f693a27f09942b472df334fd85f5825667

I think that Go uses so little from the environment (rclone's builds are reproducible on any OS) that which OS you build on is almost irrelevent as long as it can run the Go compiler.

The fact that all the actions built with this commit proves the point - rclone will build fine on 20.04 and it will build an identical binary to that build on 18.04.

So I think, on balance, the fact that we have to fix up stable branches anyway with patches for GitHub actions changes means that a potentially breaking update to `-latest` is just another thing we'll have to patch up.

So I think we probably shouldn't merge this.

However thank you @InspectorWilliamHenderson for encouraging a discussion about this.

Issue #5076, opened 2021-02-28
cmd/self-update: add --stable flag


ivandeex:
#### What is your current rclone version (output from `rclone version`)?

This request blocks on the `selfupdate` command (#548) due after 1.55.0

It is based on the thread https://forum.rclone.org/t/previously-working-rclone-backup-now-giving-error-unknown-command-rcat-for-rclone/22507/21

#### What problem are you are trying to solve?

Many Debian repositories and Ubuntu PPAs provide two distribution flavors: `main` and `stable` (they usually have the relevant word in the URL).
`main`  installs the project provided binary with latest features.
`stable` provides a **pre-latest** rolling release which usually includes stability patches back-ported from the top

#### How do you think rclone should be changed to solve that?

I propose to add a flag which could be named `--stable` (we should choose a proper name!) that will mean a similar thing.

For example, if we ran the command right now,
`self-update` without flags will install `v1.54` from https://download.rclone.org
`self-update --beta` will install the latest **beta** from https://beta.rclone.org
`self-update --stable` would install `v1.53.4` (this is a new behavior)

Normally `self-update` will get the number from https://download.rclone.org/version.txt which reflects latest release tag from the `master` git branch.

The new feature will need a change on the CI/CD pipeline side.
I propose new `stable-version.txt` which should contain latest release from the _pre-latest_ `v1.x-stable` branch.
What a "pre-latest release" means will be decided by CI/CD, the `self-update` will just follow it without any fragile version heuristics.

As @ncw noted in his [forum comment](https://forum.rclone.org/t/previously-working-rclone-backup-now-giving-error-unknown-command-rcat-for-rclone/22507/22) cited below, the term `--stable` does not align with the rclone CI/CD process:

> Right now the current v1.54-stable version has commits which haven't made it into v1.54.1 as that hasn't been released yet. That has a version number v1.54.1-beta.5153.da9dd543e.v1.54-stable right now. But it doesn't have any release tags on it at the moment.
> When it does get a release tag the main version would become 1.54.1 - at least that is the way it works at the moment.
> You can find these in https://beta.rclone.org/branch/v1.54-stable/ - there is no version.txt file in there at the moment, but there could be with a little bit of work.
> I would argue that https://downloads.rclone.org/version.txt is always the stable version but only sometimes the main version - is that distinction important?
> I think that if --stable meant install the latest beta from the stable branch that would be useful.

I think that `1.54.1` can be released only after `1.55` sees the daylight.
[...]

Issue #548, opened 2016-06-27
rclone self-update command


nagualcode:
Please, add a option to get the last binary version from command line.
Rclone is very useful to embed hardware, such as the Pi, and others. 
It would be very handy to have a cron job to search for the last version and be safe that rclone will be always up to date across a network of embeds,

ncw:
Nice idea.  Might be a bit of work.  I wonder if there are any Go frameworks about to help?

ncw:
A bit of research comes up with
  * https://github.com/sanbornm/go-selfupdate
  * https://github.com/inconshreveable/go-update

fredericofs:
I wish I could help you in here

ivandeex:
Duplicate of https://github.com/rclone/rclone/issues/3269

ivandeex:
Duplicate of https://github.com/rclone/rclone/issues/2867

ivandeex:
This ticket duplicates #3269 so merging in.

ivandeex:
Sorry all. I was wrong when I closed this in favor of #3269. I can see now that that ticket actually requests automatic update feature. Sorry for noise. Reopening.

ivandeex:
I'd like to quote https://github.com/rclone/rclone/issues/3269#issuecomment-502997308 which better fits in this ticket, I think

> Great idea! I'd like to emulate `restic self-update` which seems to work really well.

ivandeex:
A quote of the idea of https://github.com/rclone/rclone/issues/2867#issuecomment-450499371 with minor edits:

`rclone self-update  --check` will just tell you if rclone needs updating!

ivandeex:
I think the algorithm can look like:
- check whether self executable is writable by process user id (there may be windows specific tricks)
- if not writable, ask to rerun as _root_ (Administrator)
- parse https://downloads.rclone.org/version.txt
- parse [https://downloads.rclone.org/[VERSION]/SHA1SUMS](https://downloads.rclone.org/v1.54.0/SHA1SUMS)
- download architecture-compatible zip into /tmp from [https://downloads.rclone.org/[VERSION]/](https://downloads.rclone.org/v1.54.0)
- validate hash
- extract executable into self binary directory with temporary suffix (not to `/tmp`, to ensure atomicity of name swap  below)
- atomically swap name (similar to `mv -f` on Unix, I expect trickery on Windows)

What will not be implemented by this ticket:
[...]

Issue #5074, opened 2021-02-27
Instruct users to upgrade rclone


ivandeex:
#### What is your current rclone version (output from `rclone version`)?

1.54.0

This feature idea additionally _blocks_ on availability of the `rclone self-update` command which will be implemented by #548 

#### What problem are you are trying to solve?

https://forum.rclone.org/t/previously-working-rclone-backup-now-giving-error-unknown-command-rcat-for-rclone/22507/14

Many support requests on rclone forum or github involve an outdated / unsupported / ancient version of rclone. The Support team repeatedly ask users to upgrade, users rerun their tests, thread continues only after these spare cycles. Rinse, repeat.

We can try to save people's time here.

#### How do you think rclone should be changed to solve that?

Let's add simple upgrade instructions in three places:

1. In the rclone command line parser. Whenever it enounters unknown command it should print:

`Error: unknown command "xxx" for "rclone". You could use "rclone self-update" to get latest features.`

2. In the forum post template. It should say:

```
#### What is your rclone version (output from `rclone version`)
<!-- **STOP and  READ**: Are you on the latest version of rclone?
     Many newbie problems are solved with a quick update, just run `rclone self-update --check`.
 -->
```

3. In the github bug/feature template - ???

darthShadow:
We should probably also update the Bug template to be similar to the "Help & Support" template on the forum.

---

<details><summary>Current Bug Template on Github:</summary>

<p>

```
<!--

Welcome :-) We understand you are having a problem with rclone; we want to help you with that!

If you've just got a question or aren't sure if you've found a bug then please use the rclone forum:

    https://forum.rclone.org/

instead of filing an issue for a quick response.

If you think you might have found a bug, please can you try to replicate it with the latest beta?

    https://beta.rclone.org/
    
If you can still replicate it with the latest beta, then please fill in the info below which makes our lives much easier.  A log with -vv will make our day :-)

Thank you

The Rclone Developers

-->
```

#### What is the problem you are having with rclone?



#### What is your rclone version (output from `rclone version`)



#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)



####  Which cloud storage system are you using? (e.g. Google Drive)



#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)



#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)


</p>

</details>

---

<details><summary>Current "Help & Support" Template on the forum:</summary>

[...]

Issue #3545, opened 2019-09-17
rclone drives not showing up in Finder sidebar and Desktop


v57:
When i connect my drives i want them to show up in my finder sidebar so i can easily access them and see if they are connected or not.

I tried this command with flag `local` (which works in `sshfs`)
```
rclone mount --allow-non-empty --fuse-flag local yandex:/ /Volumes/Yandex
```
But it still didn't shows in my sidebar.

ncw:
I think you are using this option: https://github.com/osxfuse/osxfuse/wiki/Mount-options#user-content-local

Unfortunately the standard `rclone mount` doesn't support `--fuse-flag` - you should have seen an ERROR in the log.

It would be relatively straight forward to add a new flag, or parse the `--fuse-flag` looking for that option...

https://github.com/rclone/rclone/blob/c49a71f4385b371563a81bfae8959eb29823e94a/cmd/mount/mount.go#L73-L75

Because the fuse library does support it

https://github.com/bazil/fuse/blob/65cc252bf6691cb3c7014bcb2c8dc29de91e3a7e/options.go#L71-L72

aatreya:
Why is it important to parse these flags rather than just passing them through to FUSE as-is?

ncw:
> Why is it important to parse these flags rather than just passing them through to FUSE as-is?

That is because the fuse library rclone uses doesn't actually use libfuse - it talks straight to the kernel.

Does anyone want to have a go at this?

We could either parse the input to `--fuse-flag` - probably not my favoured option as these options are for passing to fuse, or add a new option `--local` say which will only work on macOS and will set the fuse option.

andresmatasuarez:
so, in the meantime, is there a fix for this? how can I achieve passing fuse flags down via rclone?

ncw:
The macOS build now support `rclone cmount` which will become `rclone mount` in v1.54

So you can use `rclone cmount` now and pass fuse options with the `-o local` flag as this does use libfuse.

ivandeex:
@andresmatasuarez @v57 
Please try `rclone cmount` with latest rclone 1.53.2
Does the problem reproduce?

ivandeex:
Closed after 3 months of inactivity

andresmatasuarez:
@ncw @ivandeex sorry for the ultra late response. I've just got time to give this another try and sadly it didn't work.

[...]

PR #5080, opened 2021-03-01
cmd: rclone selfupdate

ivandeex:

Implements `rclone self-update` command (ticket #548)

Flags and features:
- verifies sha256 hash of release archive
- verifies access to replaced file and directory, optionally asks user to run with elevated privileges (in Scotland - elavated)
- mimics permission bits of original file, if any
- `--output` - downloads new executable to a custom location
- `--beta` - self-explanatory
- `--check` - just an alias of `rclone version --check`

Yet unimplemented, left for future patches:
- `--stable` (`--backports`)
- `--version`

**UPDATE 2021-03-08**:
Will be implemented:
- `--stable` (mutually exclusive with `--beta`)
- `--version VER` (micro-release autocompletion if stable)
- `--package deb` (linux only)
- `--package rpm` (linux only)
- PGP signature validation (pending public key)

Yes:
- https://github.com/rclone/rclone/issues/548
- https://forum.rclone.org/t/previously-working-rclone-backup-now-giving-error-unknown-command-rcat-for-rclone/22507/6
- https://github.com/rclone/rclone/issues/5076

#### Checklist

### DIFF
--- a/cmd/selfupdate/help.go
+++ b/cmd/selfupdate/help.go
@@ -1,0 +1,18 @@
+package selfupdate
+
+// Note: "|" will be replaced by backticks in the help string below
+var selfUpdateHelp string = `
+This command downloads the latest release of rclone and replaces
+the currently running binary. The download is verified with a hashsum.
+
+The |--version VER| flag, if given, will update to a concrete version
+instead of the latest one. If you omit micro version from |VER| (for
+example |1.53|), the latest matching micro version will be used.
+
+If you previously installed rclone via a package manager, the package may
+include local documentation or configure services. This command will update
+only the rclone executable so the local manual may become inaccurate after it.
+
+Note: Windows forbids deletion of a currently running executable so this
+command will rename the old executable to 'rclone.old.exe' upon success.
+`
--- a/cmd/selfupdate/help.go
+++ b/cmd/selfupdate/help.go
@@ -3,7 +3,8 @@ package selfupdate
 // Note: "|" will be replaced by backticks in the help string below
 var selfUpdateHelp string = `
 This command downloads the latest release of rclone and replaces
-the currently running binary. The download is verified with a hashsum.
+the currently running binary. The download is verified with a hashsum
+and cryptographically signed signature.
 
 The |--version VER| flag, if given, will update to a concrete version
 instead of the latest one. If you omit micro version from |VER| (for
--- a/cmd/selfupdate/help.go
+++ b/cmd/selfupdate/help.go
@@ -11,8 +11,11 @@ instead of the latest one. If you omit micro version from |VER| (for
 example |1.53|), the latest matching micro version will be used.
 
 If you previously installed rclone via a package manager, the package may
-include local documentation or configure services. This command will update
-only the rclone executable so the local manual may become inaccurate after it.
+include local documentation or configure services. You may wish to update
+with the flag |--package deb| or |--package rpm| (whichever is correct for
+your OS) to update these too. This command with the default |--package zip|
+will update only the rclone executable so the local manual may become
+inaccurate after it.
 
 Note: Windows forbids deletion of a currently running executable so this
 command will rename the old executable to 'rclone.old.exe' upon success.
--- a/cmd/selfupdate/selfupdate.go
+++ b/cmd/selfupdate/selfupdate.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
@@ -12,7 +13,9 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -20,6 +23,8 @@ import (
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/fs/fshttp"
+	"github.com/rclone/rclone/lib/random"
 	"github.com/spf13/cobra"
 
 	versionCmd "github.com/rclone/rclone/cmd/version"
@@ -32,6 +37,7 @@ type Options struct {
 	Beta    bool
 	Stable  bool
 	Version string
+	Format  string
 }
 
 // Opt is options set via command line
@@ -45,59 +51,99 @@ func init() {
 	flags.BoolVarP(cmdFlags, &Opt.Stable, "stable", "", Opt.Stable, "Install latest stable release (this is the default)")
 	flags.BoolVarP(cmdFlags, &Opt.Beta, "beta", "", Opt.Beta, "Install latest beta release.")
 	flags.StringVarP(cmdFlags, &Opt.Version, "version", "", Opt.Version, "Install the given rclone path (default: auto-detect)")
+	flags.StringVarP(cmdFlags, &Opt.Format, "package", "", Opt.Format, "Package format: zip|deb|rpm (default: zip)")
 }
 
 var cmdSelfUpdate = &cobra.Command{
 	Use:     "selfupdate",
 	Aliases: []string{"self-update"},
 	Short:   `Update the rclone binary.`,
-	Long: `
-This command downloads the latest release of rclone and replaces
-the currently running binary. The download is verified with a hashsum.
-
-If you previously installed rclone via a package manager, the package may
-include local documentation or configure services. This command will update
-only rclone executable so the local manual may become inaccurate after it.
-
-Note: Windows forbids deletion of a currently running executable so this
-command will rename the old executable to 'rclone.exe.old' upon success.
-`,
+	Long:    strings.ReplaceAll(selfUpdateHelp, "|", "`"),
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(0, 0, command, args)
+		if Opt.Format == "" {
+			Opt.Format = "zip"
+		}
 		if Opt.Check {
-			if Opt.Stable || Opt.Beta || Opt.Output != "" || Opt.Version != "" {
-				fmt.Println("Warning: --stable, --beta, --version and --output are ignored with --check")
+			if Opt.Stable || Opt.Beta || Opt.Output != "" || Opt.Version != "" || Opt.Format != "zip" {
+				fmt.Println("Warning: --stable, --beta, --version, --format and --output are ignored with --check")
 			}
 			versionCmd.CheckVersion()
 			return
 		}
-		if err := InstallUpdate(&Opt); err != nil {
+		if Opt.Format != "zip" {
+			if Opt.Format != "deb" && Opt.Format != "rpm" {
+				log.Fatalf("--package should be one of zip|deb|rpm")
+			}
+			if runtime.GOOS != "linux" {
+				log.Fatalf(".deb and .rpm packages are supported only on Linux")
+			} else if os.Geteuid() != 0 {
+				log.Fatalf(".deb and .rpm must be installed by root")
+			}
+			if Opt.Output != "" {
+				fmt.Println("Warning: --output is ignored with --package .deb|.rpm")
+			}
+		}
+		if err := InstallUpdate(context.Background(), &Opt); err != nil {
 			log.Fatalf("Error: %v", err)
 		}
 	},
 }
 
-// GetVersion is a wrapper for versionCmd.GetVersion with extra outputs
-func GetVersion(beta bool, version string) (newVersion, siteURL string, err error) {
+// GetVersion can get the latest release number from the download site
+// or massage a stable release number - prepend semantic "v" prefix
+// or find the latest micro release for a given major.minor release.
+// Note: this will not be applied to beta releases.
+func GetVersion(ctx context.Context, beta bool, version string) (newVersion, siteURL string, err error) {
 	siteURL = "https://downloads.rclone.org"
 	if beta {
 		siteURL = "https://beta.rclone.org"
 	}
-	newVersion = version
-	if newVersion == "" {
+
+	if version == "" {
+		// Request the latest release number from the download site
 		_, newVersion, _, err = versionCmd.GetVersion(siteURL + "/version.txt")
+		return
+	}
+
+	newVersion = version
+	if version[0] != 'v' {
+		newVersion = "v" + version
+	}
+	if beta {
+		return
+	}
+
+	if valid, _ := regexp.MatchString(`^v\d+\.\d+(\.\d+)?$`, newVersion); !valid {
+		return "", siteURL, errors.New("invalid semantic version")
+	}
+
+	// Find the latest stable micro release
+	if strings.Count(newVersion, ".") == 1 {
+		html, err := downloadFile(ctx, siteURL)
+		if err != nil {
+			return "", siteURL, errors.Wrap(err, "failed to get list of releases")
+		}
+		reSubver := fmt.Sprintf(`href="\./%s\.\d+/"`, regexp.QuoteMeta(newVersion))
+		allSubvers := regexp.MustCompile(reSubver).FindAllString(string(html), -1)
+		if allSubvers == nil {
+			return "", siteURL, errors.New("could not find the minor release")
+		}
+		// Use the fact that releases in the index are sorted by date
+		lastSubver := allSubvers[len(allSubvers)-1]
+		newVersion = lastSubver[8 : len(lastSubver)-2]
 	}
 	return
 }
 
 // InstallUpdate performs rclone self-update
-func InstallUpdate(opt *Options) error {
+func InstallUpdate(ctx context.Context, opt *Options) error {
 	// Find the latest release number
 	if opt.Stable && opt.Beta {
 		return errors.New("--stable and --beta are mutually exclusive")
 	}
 
-	newVersion, siteURL, err := GetVersion(opt.Beta, opt.Version)
+	newVersion, siteURL, err := GetVersion(ctx, opt.Beta, opt.Version)
 	if err != nil {
 		return errors.Wrap(err, "unable to detect new version")
 	}
@@ -115,7 +161,16 @@ func InstallUpdate(opt *Options) error {
 		return nil
 	}
 
-	// Find the executable path
+	// Install .deb/.rpm package if requested by user
+	if opt.Format == "deb" || opt.Format == "rpm" {
+		err := installPackage(ctx, opt.Beta, newVersion, siteURL, opt.Format)
+		if err == nil {
+			fmt.Printf("Successfully updated rclone package to version %s\n", newVersion)
+		}
+		return err
+	}
+
+	// Get the current executable path
 	executable, err := os.Executable()
 	if err != nil {
 		return errors.Wrap(err, "unable to find executable")
@@ -126,11 +181,18 @@ func InstallUpdate(opt *Options) error {
 		targetFile = executable
 	}
 
-	// Check for possible access errors in advance
-	newFile := targetFile + ".new"
+	// Make temporary file names and check for possible access errors in advance
+	var newFile string
+	if newFile, err = makeRandomExeName(targetFile, "new"); err != nil {
+		return err
+	}
 	savedFile := ""
 	if runtime.GOOS == "windows" {
-		savedFile = targetFile + ".old"
+		savedFile = targetFile
+		if strings.HasSuffix(savedFile, ".exe") {
+			savedFile = savedFile[:len(savedFile)-4]
+		}
+		savedFile += ".old.exe"
 	}
 
 	if savedFile == executable || newFile == executable {
@@ -142,25 +204,64 @@ func InstallUpdate(opt *Options) error {
 	}
 
 	// Download the update as a temporary file
-	if err := downloadUpdate(opt.Beta, newVersion, siteURL, newFile); err != nil {
+	err = downloadUpdate(ctx, opt.Beta, newVersion, siteURL, newFile, "zip")
+	if err != nil {
 		return errors.Wrap(err, "failed to update rclone")
 	}
 
-	// Copy permission bits from the old executable
-	fileMode := os.FileMode(0755)
-	if fileInfo, err := os.Lstat(targetFile); err == nil {
-		fileMode = fileInfo.Mode()
+	err = replaceExecutable(targetFile, newFile, savedFile)
+	if err == nil {
+		fmt.Printf("Successfully updated rclone to version %s\n", newVersion)
 	}
-	if err := os.Chmod(newFile, fileMode); err != nil {
-		return errors.Wrap(err, "failed to set permissions")
+	return err
+}
+
+func installPackage(ctx context.Context, beta bool, version, siteURL, format string) error {
+	tempFile, err := ioutil.TempFile("", "rclone.*."+format)
+	if err != nil {
+		return errors.Wrap(err, "unable to write temporary package")
+	}
+	packageFile := tempFile.Name()
+	_ = tempFile.Close()
+	defer func() {
+		if rmErr := os.Remove(packageFile); rmErr != nil {
+			fs.Errorf(nil, "%s: could not remove temporary package: %v", packageFile, rmErr)
+		}
+	}()
+	if err := downloadUpdate(ctx, beta, version, siteURL, packageFile, format); err != nil {
+		return err
 	}
 
-	// Replace current executable by the new file
+	packager := "dpkg"
+	if format == "rpm" {
+		packager = "rpm"
+	}
+	cmd := exec.Command(packager, "-i", packageFile)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to run %s: %v", packager, err)
+	}
+	return nil
+}
+
+func replaceExecutable(targetFile, newFile, savedFile string) error {
+	// Copy permission bits from the old executable
+	// (it was extracted with mode 0755)
+	fileInfo, err := os.Lstat(targetFile)
+	if err == nil {
+		if err = os.Chmod(newFile, fileInfo.Mode()); err != nil {
+			return errors.Wrap(err, "failed to set permission")
+		}
+	}
+
 	if err = os.Remove(targetFile); os.IsNotExist(err) {
 		err = nil
 	}
+
 	if err != nil && savedFile != "" {
-		// Windows forbids removal of a running executable so we rename it
+		// Windows forbids removal of a running executable so we rename it.
+		// For starters, rename download as the original file with ".old.exe" appended.
 		var saveErr error
 		if saveErr = os.Remove(savedFile); os.IsNotExist(saveErr) {
 			saveErr = nil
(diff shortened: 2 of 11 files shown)


ivandeex:
> Should we also add a note in the Docs that the manpages will no longer be accurate after the binary is updated using this?

I will add this warning in the success message

ivandeex:
As I expected there is a Windows quirk. Windows never allows to rewrite currently running executable, even if user has write permission. The command will fail with exit code 1 and print the following message
```
Failed to replace old file, new version 1.54.xxx saved as C:\bin\rclone.exe.new
```
The workaround for scheduled update tasks is to use the `--output` flag:
```
C:\bin\rclone.exe self-update --output C:\bin\rclone.new.exe
del C:\bin\rclone.exe
rename C:\bin\rclone.new.exe C:\bin\rclone.exe
```

darthShadow:
This seems like a reasonable alternate way of accomplishing it: https://stackoverflow.com/a/55247459/3476387

albertony:
The rename trick works great, and it is also possible to tell Windows to automatically delete it on next startup, simply by adding the path into registry value HKLM\System\CurrentControlSet\Control\Session Manager\PendingFileRenameOperations. Not that it is very important in this case, but just an idea..

ivandeex:
I would like to have the first released version of the command only the required features, have it out as fast as possible and leave optional features for later enhancements.
I think we can postpone registry tricks or leave it for chocolatey installer. If someone installed a choco package, I'd recommend to update it using choco too, `cinst -y rclone` without mixing package with self-update strategy.
@albertony

ivandeex:
I have just answered a question https://github.com/rclone/rclone/issues/3545#issuecomment-789415244 and noticed that rclone for linux has two distinct flavors.
The one distributed in `zip` is a static binary that doesn't support `cmount`, the other in `deb` or `rpm` does have it. By the way, I think the `version` command should be extended to print current flavor.

Current draft of `selfupdate` supports only `zip`, consequently it will replace `libfuse`-aware binary and probably break tooling.

This problem manifests iself only on Linux. Windows and OSX distributions are always count-aware. Plan9 and other exotics never have it.

I will now add a detection code that will _refuse to update_ if rclone runs on Linux has libfuse/cmount support and updates its own running executable. I admit, it's a limitation, but we can solve it later.

Another solution is to extract single dynamic binary from distribution `.deb`. I believe that `.rpm` contains the same single dynamic binary, no `.so` files. AFAIK deb packages contain a plain old `ar` combined with _gnu-tar_ and _gzip_. This can be handled with golang.

Other solutions will take time, for example: make CI/CD produce a separate zip compiled with `cmount` tag (this needs a decision from @ncw), detect linux distribution (it's fragile and needs a lit of testing), run apt/rpm/yum/whatever as root, and so long.

@ncw Sorry for bothering, I would like to have your opinion

albertony:
> I think we can postpone registry tricks

Sure. BTW: It's exposed as an Windows API function, [MoveFileEx](https://docs.microsoft.com/en-gb/windows/win32/api/winbase/nf-winbase-movefileexa) with flag `MOVEFILE_DELAY_UNTIL_REBOOT`, so no direct registry access necessary - if it is easy to call the api from golang (or there is a library wrapping that function, could not find any with a quick search).

> I would like to have the first released version of the command have only the required features, release it as fast as possible and leave optional features for later enhancements.

I agree. Have not really decided myself what I would want (delete rclone.exe.old on boot or not), but just wanted to note it in case it could turn up useful..

Great work btw!

ivandeex:
> Have not really decided myself what I would want (delete rclone.exe.old on boot or not), but just wanted to note...

A feature request ticket is something you can close when unsure or spam with `@youlazybastard love to see this` when...

ivandeex:
Based on the discussion in https://forum.rclone.org/t/rclone-cmount-on-linux/22602 the official build of rclone for Linux does not support `cmount` and runtime detection of such features will need tweaks in the build process, so it's out of scope for this PR. I will address it in a later patch.

ivandeex:
### Implementation details:
- The command was named `selfupdate` (no dash) to be in line with other rclone commands, but added the `self-update` alias (with a dash) to aid restic users
- Included all features from https://github.com/rclone/rclone/pull/5080#issue-581944497: hash validation, proactive permission checks, `--output` flag for custom executable locations, `--check` flag as an alias for `rclone version --check`
- Included `--stable` flag (it is implied so in fact does nothing) from https://github.com/rclone/rclone/issues/5076#issuecomment-787441976
- Included `--beta` flag (mutually exclusive with `--stable`) from https://github.com/rclone/rclone/issues/5076#issuecomment-787441976
- Implemented `--version` flag proposed by @darthShadow  in https://github.com/rclone/rclone/issues/5076#issuecomment-787441976
- Implemented Windows trick that keeps old executable renamed with `.old` suffix (recommended by @darthShadow in https://github.com/rclone/rclone/pull/5080#issuecomment-788016526)
- Excluded _pending delete after boot_ feature proposed by @albertony in https://github.com/rclone/rclone/pull/5080#issuecomment-788902774 and https://github.com/rclone/rclone/pull/5080#issuecomment-789514033
- Added reasonable unit tests for Unix and Windows behavior (yet disabled it for `go test -short` to reduce load on rclone.org somewhat)
- Postponed the `dynamic` vs `static` check until later as explained in https://github.com/rclone/rclone/pull/5080#issuecomment-789592959
- Added the message `You could use 'rclone self-update' to get latest features.` in case of unknown subcommand (proposed by https://github.com/rclone/rclone/issues/5074 and [on the forum](https://forum.rclone.org/t/previously-working-rclone-backup-now-giving-error-unknown-command-rcat-for-rclone/22507/14))
- Postponed updating github issue templates until the next release. Unlike the item above they would become effective right after merge into master
- Included quick fix from https://github.com/rclone/rclone/pull/5093 as otherwise PR tests fail

All unit tests pass for me

@ncw 
Please review

@albertony @darthShadow 
Please take a look and give it a try

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+		if saveErr == nil {
+			saveErr = os.Rename(targetFile, savedFile)
+		}
+		if saveErr == nil {
+			fmt.Printf("The old executable was saved as %s\n", savedFile)
+			err = nil
+		} else {
+			// The rename trick didn't work out, proceed like on Unix
+			_ = os.Remove(savedFile)
```
albertony, line 176: Is this necessary? Either it will be a second attempt at deleting rclone.xe.old file when it already failed at line 165, or it were successfully deleted but renaming rclone.exe to .old failed in line 169 and there is nothing to delete.

albertony:
`.\rclone.exe selfupdate --stable --version 1.53`
Fails:
> Error: failed to update rclone: failed with 404 Not Found downloading https://downloads.rclone.org/1.53/rclone-1.53-windows-amd64.zip

`.\rclone.exe selfupdate --stable --version 1.53.0`
Fails:
>  Error: failed to update rclone: failed with 404 Not Found downloading https://downloads.rclone.org/1.53.0/rclone-1.53.0-windows-amd64.zip

`.\rclone.exe selfupdate --stable --version v1.53`
Fails:
> Error: failed to update rclone: failed with 404 Not Found downloading https://downloads.rclone.org/v1.53/rclone-v1.53-windows-amd64.zip

`.\rclone.exe selfupdate --stable --version v1.53.0`
Works!

albertony:
Finished my testing for now. OS: windows/amd64. Self-compiled from your code, with go1.16.
Normal usage works fine! :clap:
Seems the `--version` option needs a fix as commented above, and I have one comment/question regarding the code for removal of rclone.exe.old.

ivandeex:
> Seems the `--version` option needs a fix as commented above
Do you mean automatically prepend `v` as a convenience?
Ok, will do 👍 
I thought that `--version` should be literal and explicit by design.

albertony:
> I thought that `--version` should be literal and explicit by design.

Ah, ok. I assumed it was suppose to work as in the following examples: https://github.com/rclone/rclone/issues/5076#issuecomment-787441976

Still think it would be nice to be a bit "lenient" on the version strings, although main point of this feature is for people to update to latest version so the `--version` option kind of defeats the purpose anyway.. :)

> Will prepend v to 1.53.0 if not found at start.

Good.

> But will not append .0 to e.g. 1.53 as beta versions have different structure.

That's ok.

ivandeex:
@darthShadow 
Implementing proposed conversions from your table
`--stable --version 1.53` --> `v1.53.4`
`--stable --version 1.54` --> `v1.54.0`
requires some `index.html` parsing
(or trying micro versions sequentially creating extra load on download site)
It's more convenient for user but more fragile.
Do you think we should go for it?
I can do it but need a confirmation as I'm in doubt

albertony:
> --stable --version 1.53 --> v1.53.4
> --stable --version 1.54 --> v1.54.0

An easy approach would be just appending ".0" when fewer than 3 part version numbers. That is not something that will surprise users, I think. Official releases seems always use 3 version numbers, and beta versions adds a suffix, so that should be safe. Resulting:

`--stable --version 1.53 --> v1.53.0`
`--stable --version 1.54 --> v1.54.0`

albertony:
Just wanted to add, that instead of index.html parsing one could use the GitHub API to get the release information as json: https://api.github.com/repos/rclone/rclone/releases. Not saying its something that must be considered now.

ivandeex:
> instead of index.html parsing one could use the GitHub API

json parsing vs html parsing - no problem (unless it requires access tokens)
accessing two sites (github.com) to achieve single download - less robust. 
i'd stick with index.html

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		_ = os.Remove(newFile)
+		return err
+	}
+	fmt.Printf("Successfully updated rclone to version %s\n", newVersion)
+	return nil
+}
+
+func downloadUpdate(beta bool, version, siteURL, newFile string) error {
+	archiveFilename := fmt.Sprintf("rclone-%s-%s-%s.zip", version, runtime.GOOS, runtime.GOARCH)
```
ncw, line 191: Note that for GOOS="darwin" we rename that to "osx" since no mac user knows their machines run darwin.

(This should probably be `macOS` now-a-days in the age of macOS 11)

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+	_, err = io.Copy(writer, reader)
+	_ = writer.Close()
+	if err != nil {
+		_ = os.Remove(newFile)
+	}
+	return err
+}
+
+func downloadFile(url string) ([]byte, error) {
+	resp, err := http.Get(url)
```
ncw, line 318: If you were to use an `http.Client` from `fshttp` here then stuff like `--bwlimit` and `--dump headers` will start working which is probably useful.

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+
+	_, err = io.Copy(writer, reader)
+	_ = writer.Close()
+	if err != nil {
+		_ = os.Remove(newFile)
+	}
+	return err
+}
+
+func downloadFile(url string) ([]byte, error) {
```
ncw, line 317: Downloading the release in memory seems OK provided the release doesn't get too big!

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+		return err
+	}
+	fs.Debugf(nil, "downloaded release archive: %s", archiveURL)
+	gotHash := sha256.Sum256(archiveBuf)
+	fs.Debugf(nil, "archive hashsum: %s", hex.EncodeToString(gotHash[:]))
+
+	// CI/CD does not provide hashsums for beta releases
+	if !beta {
+		hashsumsURL := fmt.Sprintf("%s/%s/SHA256SUMS", siteURL, version)
+		hashsumsBuf, err := downloadFile(hashsumsURL)
```
ncw, line 204: In a perfect world we'd check the signature on SHA256SUMS - it is gpg signed with my pgp key. We'd need to embed the PGP signature in rclone I think to make this work securely.

Let's leave this out for the moment.

cmd/selfupdate/selfupdate_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"os/exec"
+	"regexp"
+	"runtime"
+	"testing"
+
+	"github.com/rclone/rclone/fs"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInstallOnLinux(t *testing.T) {
```
ncw, line 15: These tests are downloading real files aren't they?

Which makes them more integration tests as if the download server is down the tests will fail.

We could run a test server to serve the files and have an Opt to configure the URLs?

Or we could only run these tests on the integration test server and skip them on the CI with fstest/testy.SkipUnreliable ?

ncw (commented):
I think this is looking very nice @ivandeex - a great implementation of a nice feature!

I put some notes inline.

I note also that if people have installed via a package, maybe we should update the package? Something for the next revision?

cmd/selfupdate/selfupdate.go
ivandeex: Temporary file? Doable.
But I didn't see rclone releases more than 50-60M
It's not big even for embedded boxes

cmd/selfupdate/selfupdate.go
ivandeex: Embedding gpg key will require tweaks in CI/CD (github secret -> workflow variable -> golang linker key -> executable)
I will submit it as a separate PR.

cmd/selfupdate/selfupdate_test.go
ivandeex: Testing against a dummy rclone.org doesn't look like a convincing weighty test to me.
In the same time, yeah, the test will fail if downloads.rclone.org or beta.rclone.org breaks down.
I'd constrain it also for the DDoS reason as every force-push on every PR on every rclone fork will send few more requests to rclone.org
I tried to tackle it with the `!testing.Short` check. I'll use `testy.SkipUnreliable` after your advice.

cmd/selfupdate/selfupdate.go
ivandeex: What threshold do you think is reasonable?

cmd/selfupdate/selfupdate.go
```diff
@@ -0,0 +1,327 @@
...
+	Use:     "selfupdate",
+	Aliases: []string{"self-update"},
+	Short:   `Update the rclone binary.`,
+	Long: `
+This command downloads the latest release of rclone and replaces
+the currently running binary. The download is verified with a hashsum.
+
+If you previously installed rclone via a package manager, the package may
+include local documentation or configure services. This command will update
+only rclone executable so the local manual may become inaccurate after it.
```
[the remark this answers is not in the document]
ivandeex: > I note also that if people have installed via a package, maybe we should update the package? Something for the next revision?

This is not the first or second time I see this idea.
I thought about installing .deb/.rpm
TL;DR I wouldn't plan it even for next or next after next revision. :-)

Cons:
1. detection of whether **current binary** was installed directly, via .deb/.rpm package or via install.sh 
     even if one of binaries on the host was installed by a package sometime ago
     any such detection will be fragile
2. to detect .deb/.rpm, we must at least know official rclone path in .deb/.rpm, and keep it in sync with the check in this function
3. to decide whether we want .deb or .rpm for current system, we have to detect the system
     (look for example at the ansible code that does the detection, you'll see a lot of surprises)
     tl;dr it is fragile
4. unlike binary install, package install will **always** require root
5. it brings in yet another external dependency

Pros:
1. it will install manual page

What else is in .deb/.rpm, besides executable?
Custom rclone systemd services? No
Package contains executable, manual page, and MIT license
Who reads manual page when there is rclone.org and rclone --help?
Well, we address rpm/deb. What about .apk, flatpak, snap? We fallback to executable-only update.

ivandeex:
I finally see that `index.html` from https://downloads.rclone.org can be parsed by a really simple regex. Will do what drShadow ordered.

cmd/selfupdate/selfupdate.go
albertony: Yes, that is probably the way to do it. Although, probably not super important to handle.

@ivandeex: My comment here were originally regarding the specific code lines, where this else:
```
// The rename trick didn't work out, proceed like on Unix
_ = os.Remove(savedFile)
```
only seemed to perform a second attempt at deleting `rclone.exe.old` when it has already failed once, or attempting to deleting a non-existant `rclone.exe.old` after renaming `rclone.exe` to `rclone.exe.old` has just failed. So the else seems unnecessary... or so I wondered, but perhaps I read the code wrong..

ivandeex:
@ncw
> The solution here would be to make the save file `rclone.RANDOMNUMBER.old.exe`.
> This might be a good idea for the `.new` file also - this will help with races in the very unlikely case that the user runs `self-update` twice.
> You can use `random.String()` to make random suffixes.

I'd like to enforce some `output results stability` if possible. It will greatly simplify shell/cron wrappers around `selfupdate`.
I will be default use random suffix with `.new` files because they are never retained, neither on success, nor on failure.
I will try `.old` without random suffix first, and add random suffix **only** if I detect race.

ivandeex:
@ncw 
Note: as soon as your PGP key gets released in the wild embedded in the next rclone release, it's not secret anymore.

cmd/selfupdate/selfupdate.go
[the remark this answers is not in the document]
ncw: I was just thinking that we supply .deb and .rpm packages so if the user has installed from a .deb or .rpm we should offer to install our version of those instead.

The question - is this binary installed by a package isn't trivial to answer and I think we should side-step it. If users want to install a deb package then they should do `rclone self-update --deb` which is much simpler to code. This should be limited to downloading the correct `.deb` from the rclone page and running `dpkg -i package.deb` (possibly prefixed by sudo) on it.

I think updating the executable in place like you've done is exactly the way `restic self-update` works and will help enormously with support. Its also the correct way for macOS and Windows - the `deb`, `rpm` thing is of secondary importance I think.

cmd/selfupdate/selfupdate.go
ncw: I think we should leave your code as-is and not worry about it for the moment.

```
$ rclone --fast-list lsl memstore:downloads-rclone-org --include *.zip | sort -n | tail
 19355621 2019-10-05 12:17:21.161062944 v1.49.5/rclone-v1.49.5-osx-amd64.zip
 19624754 2019-11-19 16:21:16.063618980 v1.50.2/rclone-v1.50.2-osx-386.zip
 19625809 2019-11-02 15:03:25.089316344 v1.50.1/rclone-v1.50.1-osx-386.zip
 19633882 2019-10-26 11:55:15.169383338 v1.50.0/rclone-v1.50.0-osx-386.zip
 19914331 2020-02-01 10:59:48.391080948 v1.51.0/rclone-v1.51.0-osx-386.zip
 22953431 2020-06-24 14:08:44.602862189 v1.52.2/rclone-v1.52.2-osx-386.zip
 22956842 2020-08-08 09:44:20.717505397 rclone-current-osx-386.zip
 22956842 2020-08-08 09:44:20.717505397 v1.52.3/rclone-v1.52.3-osx-386.zip
 22970878 2020-05-27 17:53:06.774147983 v1.52.0/rclone-v1.52.0-osx-386.zip
 22981766 2020-06-10 16:27:10.754567269 v1.52.1/rclone-v1.52.1-osx-386.zip
```

The biggest official release is 23M which as you said is peanuts now-a-days.

cmd/selfupdate/selfupdate.go
ncw: It would only need the public key embedded, not the private key so that could just go in the code as a text blob.

However this should really be done with an `rclone release key` rather than my private key.

There is probably a much better way of doing it other than using GPG too - something much lighter weight.

ncw:
> I finally see that `index.html` from https://downloads.rclone.org can be parsed by a really simple regex. Will do what drShadow ordered.

It can actually be parsed by the http backend too...

> I'd like to enforce some `output results stability` if possible. It will greatly simplify shell/cron wrappers around `selfupdate`.
> I will by default use random suffix with `.new` files because they are never retained, neither on success, nor on failure.
> I will try `.old` without random suffix first, and add random suffix **only** if I detect race.

OK sounds perfect

> Note: as soon as your PGP key gets released in the wild embedded in the next rclone release, it's not secret anymore.

Only the public part needs to be in the binary, eg [restic verify.go](https://github.com/restic/restic/blob/master/internal/selfupdate/verify.go)

cmd/selfupdate/selfupdate.go
ivandeex: I don't know what your public key is
I tried this one https://www.craig-wood.com/nick/pub/pgp-key.txt
But golang cannot read it. I think PGP-2 isn't supported.
We'll need a modern public key.

```
package main

import "fmt"
import "os"
import "golang.org/x/crypto/openpgp"

func main() {
        in, err := os.Open("pgp-key.txt")
        _, err = openpgp.ReadArmoredKeyRing(in)
        fmt.Println(err)
}
```

cmd/selfupdate/selfupdate.go
ivandeex: @albertony You are right, it is indeed unnecessary. It removed it for the next iteration.

cmd/selfupdate/selfupdate.go
ivandeex: leaving as is

cmd/selfupdate/selfupdate.go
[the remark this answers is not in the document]
ivandeex: done

ivandeex:
### Summary of new changes after 1st review:

- using `fshttp.Client` now [ncw]
- replace GOOS `darwin` with `osx` in archive names [ncw]
- use `fstest/testy.SkipUnreliable` to run in integration suite only [ncw]
- removed unneeded `os.Remove(savedFile)` [albertony]
- prepend `v` to incomplete semantic `--version` [darthShadow]
- extract last micro release from `index.html` to complete version number [darthShadow]
- random suffix for `.new` files
- `.old` file by default is `rclone.old.exe`
- random old suffix `rclone.xxxx.old.exe` if a race was detected
- install `.deb`/`.rpm` by flag `--package deb|rpm` (default: zip) [ncw]
- preliminary pgp check (still waiting for a modern public pgp key) [ncw]

@ncw please have it reviewed

cmd/selfupdate/selfupdate.go
ivandeex: @ncw
we could make this world perfect 🐇 🌈 by finding such a public release key that passes the test above ✨

ivandeex:
TODO will create an RFC issue for the [MoveFileEx](https://docs.microsoft.com/en-gb/windows/win32/api/winbase/nf-winbase-movefileexa) aka _MOVEFILE_DELAY_UNTIL_REBOOT_ feature requested in https://github.com/rclone/rclone/pull/5080#issuecomment-789514033 when we merge this pull.

ncw:
> ### Summary of new changes after 1st review:
[snip]

Great changes :-) I'll run through a code review now.

> Could you provide a valid ([#5080 (comment)](https://github.com/rclone/rclone/pull/5080#discussion_r589033247)) public key ([#5080 (comment)](https://github.com/rclone/rclone/pull/5080#discussion_r589226236)) ?

I think this should do it

```
$ gpg --armor --export-options export-minimal --export nick@craig-wood.com
```

Which produced

```
-----BEGIN PGP PUBLIC KEY BLOCK-----

mQGiBDuy3V0RBADVQOAF5aFiCxD3t2h6iAF2WMiaMlgZ6kX2i/u7addNkzX71VU9
7NpI0SnsP5YWt+gEedST6OmFbtLfZWCR4KWn5XnNdjCMNhxaH6WccVqNm4ALPIqT
59uVjkgf8RISmmoNJ1d+2wMWjQTUfwOEmoIgH6n+2MYNUKuctBrwAACflwCg1I1Q
O/prv/5hczdpQCs+fL87DxsD/Rt7pIXvsIOZyQWbIhSvNpGalJuMkW5Jx92UjsE9
1Ipo3Xr6SGRPgW9+NxAZAsiZfCX/19knAyNrN9blwL0rcPDnkhdGwK69kfjF+wq+
QbogRGodbKhqY4v+cMNkKiemBuTQiWPkpKjifwNsD1fNjNKfDP3pJ64Yz7a4fuzV
X1YwBACpKVuEen34lmcX6ziY4jq8rKibKBs4JjQCRO24kYoHDULVe+RS9krQWY5b
e0foDhru4dsKccefK099G+WEzKVCKxupstWkTT/iJwajR8mIqd4AhD0wO9W3MCfV
Ov8ykMDZ7qBWk1DHc87Ep3W1o8t8wq74ifV+HjhhWg8QAylXg7QlTmljayBDcmFp
Zy1Xb29kIDxuaWNrQGNyYWlnLXdvb2QuY29tPohxBBMRCAAxBQsHCgMEAxUDAgMW
AgECF4AWIQT79zfs6firGGBL0qyTk14C/ztU+gUCXjg2UgIZAQAKCRCTk14C/ztU
+lmmAJ4jH5FyULzStjisuTvHLTVz6G44eQCfaR5QGZFPseenE5ic2WeQcBcmtoG5
Ag0EO7LdgRAIAI6QdFBg3/xa1gFKPYy1ihV9eSdGqwWZGJvokWsfCvHy5180tj/v
UNOLAJrdqglMSvevNTXe8bT65D6423AAsLhch9wq/aNqrHolTYABzxRigjcS1//T
yln5naGUzlVQXDVfrDk3Md/NrkdOFj7r/YyMF0+iWwpFz2qAjL95i5wfVZ1kWGrT
2AmivE1wD1sWT/Ja3FDI0NRkU0Nbz/a0TKe4ml8iLVtZXpTRbxxCCPdkHXXgSyu1
eZ4NrF/wTJuvwGn12TJ1EF95aVkHxAUw0+KmLGdcyBG+IKuHamrsjWIAXGXV///K
AxPgUthccQ03HMjltFsrdmen5Q034YM3eOsAAwUH/jAKiIAA8LpZmZPnt9GZ4+Ol
Zp22VAfyfDOFl4Ol+cWjkLAgjAFsm5gnOKcRSE/9XPxnQqkhw7+ZygYuUMgTDJ99
/5IM1UQL3ooS+oFrDaE99S8bLeOe17skcdXcA/K83VqD9m93rQRnbtD+75zqKkZn
9WNFyKCXg5P6PFPdNYRtlQKOcwFR9mHRLUmapQSAM8Y2pCgALZ7GViKQca8/TT1T
gZk9fJMZYGez+IlOPxTJxjn80+vywk4/wdIWSiQj+8u5RzT9sjmm77wbMVNGRqYd
W/EemW9Zz9vi0CIvJGgbPMqcuxw8e/5lnuQ6Mi3uDR0P2RNIAhFrdZpVSME8xQaI
RgQYEQIABgUCO7LdgQAKCRCTk14C/ztU+mLBAKC2cdFy7eLaQAvyzcE2VK6HVIjn
JACguA00bxLQuJ4+RCJrLFZP8ZlN2sc=
=TtR5
-----END PGP PUBLIC KEY BLOCK-----
```

I've attached a zip with the ascii version and the binary version too just in case you want the binary version.

[ncw-pgp-keys.zip](https://github.com/rclone/rclone/files/6108210/ncw-pgp-keys.zip)

cmd/selfupdate/help.go
```diff
@@ -0,0 +1,18 @@
+package selfupdate
+
+// Note: "|" will be replaced by backticks in the help string below
+var selfUpdateHelp string = `
+This command downloads the latest release of rclone and replaces
+the currently running binary. The download is verified with a hashsum.
```
ncw, line 6: And cryptographically signed signature (if we get that working :-)

cmd/selfupdate/help.go
```diff
@@ -0,0 +1,18 @@
...
+
+The |--version VER| flag, if given, will update to a concrete version
+instead of the latest one. If you omit micro version from |VER| (for
+example |1.53|), the latest matching micro version will be used.
+
+If you previously installed rclone via a package manager, the package may
+include local documentation or configure services. This command will update
+only the rclone executable so the local manual may become inaccurate after it.
```
ncw, line 14: Note here `--format deb`?

ncw (commented):
That is looking really nice :-)

I put a couple of docs comments inline.

Well done!

cmd/selfupdate/help.go
ivandeex: Thinking about name for this option I considered `--format` and selected `--package`
What do you think?
@ncw @albertony

cmd/selfupdate/help.go
ivandeex: And we got it working

ivandeex:
PGP signature validation implemented. Waiting for replies to https://github.com/rclone/rclone/pull/5080#discussion_r590341930
Proceeding to `rclone bisync`.

cmd/selfupdate/selfupdate_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// incomplete stable version should have micro release added
+	resultVer, _, err = GetVersion(ctx, false, "1.52")
+	assert.NoError(t, err)
+	assert.Equal(t, "v1.52.3", resultVer)
+}
+
+func makeTestDir() (testDir string, err error) {
+	testDirBase := filepath.Join(os.TempDir(), "rclone-test-selfupdate.")
+	for attempt := 0; attempt < 5; attempt++ {
```
ivandeex, line 48: const maxAttempts = 5

ivandeex:
PGP signature validation - implemented.

The only _minor_ thing left is a flag name: --package vs --format and a help message
See https://github.com/rclone/rclone/pull/5080#discussion_r590277016 

Will now proceed on to `rclone bisync` and `sum` file support.

cmd/selfupdate/help.go
albertony: No strong opinion here. As zip is the default, I'm fine with `--package`.
In general I find that "package" sort of implies something handled by a package manager, but then "zip package" is also sometimes used as a synonym for "zip archive", so `--package zip` is not that strange, perhaps.. Maybe "distribution" is a generic term for different variants of the same release, but it is also a word with strong associations to linux, so..

ncw: I think I mildly prefer `zip|deb|rpm package` over `zip|deb|rpm format`. I think `package` reads slightly better.

Not sure why I said `--format deb` above - I think the option was called `format` then, or maybe I just got it wrong :-)

I'm happy with `--package`.

My note was do indicate that this para should probably talk about `--package`

> If you previously installed rclone via a package manager, the package may
> include local documentation or configure services. This command will update
> only the rclone executable so the local manual may become inaccurate after it.
> 

Should maybe be

> If you previously installed rclone via a package manager, the package may
> include local documentation or configuration services. You may wish to update with the flag `--package deb` or `--package rpm` (whichever is correct for your OS) to update these too. This command with the default `--package zip` will update
> only the rclone executable so the local manual may become inaccurate after it.
>

ncw:
> PGP signature validation - implemented.
> 
> The only _minor_ thing left is name for the flag: _--package_ vs _--format_
> and wording in the help message
> See [#5080 (comment)](https://github.com/rclone/rclone/pull/5080#discussion_r590277016)

`package` is great!

I put some suggested text in the comment above
 
> Will now proceed on to `rclone bisync` and `sum` file support.

:-)

ivandeex:
> Testing the `--version` parameter, I found it useful to include support for `--dry-run` and `--interactive`. Then I can run with `--version 1.50 --dry-run` and verify that I would get `v1.50.2` and also get a confirmation of the full path to the rclone binary.....

I think `--check` in fact already means the _dry run_ mode (in context of `selfupdate`).
The flag combination `selfupdate --check` (without other flags) is an alias for `rclone version --check`, it prints the latest stable/beta versions.

Let's add more flag combinations to cover your use cases with an output like below:
```
$ rclone selfupdate --check [--version VER] [--stable] [--beta] [--output PATH] [--package PKG]
Warning: --package flag is ignored in --check mode
If used without --check, this would install rclone version RESOLVED_VER at RESOLVED_PATH
```

> ```
> if operations.SkipDestructive(ctx, targetFile, "update to version " + newVersion) {
>     return nil
> }
> ```

I think that `SkipDestructive` as well as  `interactive` mode is relevant for operations over data while `self-update` is a maintenance command.
It's a bit of overkill here.

cmd/selfupdate/help.go
ivandeex: Done

ivandeex:
All requests addressed.
Unit tests pass.
LGTM.
Merging.

PR #5085, opened 2021-03-02
rcd: Added systemd notification during the 'rclone rcd' command call.

### LINKED ISSUE #5073: systemd notify in rcd

Running latest stable and a couple of betas on various machines.

I've got a systemd unit that starts `rclone rcd`.  
I've got other systemd units that depend on the first one, which use `rclone rc mount/mount` in order to mount various remotes.  

Right now I have to add `sleep` commands before all the `rclone rc` commands because they run before the RCD is able to accept them.  

`rclone rcd` should send systemd service notifications when it's ready and when it's quitting, just like `rclone mount` does.

Navdevl:
rcd: Added systemd notification during the 'rclone rcd' command call. This also fixes #5073 

Signed-off-by: Naveen Honest Raj <naveendurai19@gmail.com>

When running `rclone rcd` command, there is no systemd notification sent. There are cases where the end user depends on such systemd notification and this enhancement I worked on in this PR will come in handy.

Yes, it is discussed here. (https://github.com/rclone/rclone/issues/5073#issuecomment-788024498)

#### Checklist

### DIFF
--- a/cmd/rcd/rcd.go
+++ b/cmd/rcd/rcd.go
@@ -3,10 +3,13 @@ package rcd
 import (
 	"context"
 	"log"
+	"sync"
 
+	sysdnotify "github.com/iguanesolutions/go-systemd/v5/notify"
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs/rc/rcflags"
 	"github.com/rclone/rclone/fs/rc/rcserver"
+	"github.com/rclone/rclone/lib/atexit"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


Navdevl:
@darthShadow Should I also be adding tests for asserting the `systemd` notification is sent?

darthShadow:
I don't think that's necessary since we don't have it for the mountlib too.

Going through this, I wonder if we should move the code to the server wait function so it can be used for all serve commands too instead of only rcd but that's beyond the scope of the issue.

The changes look good otherwise, thanks.

@ncw will have to check if anything else is required or if this is good to go as-is.

ivandeex:
Sorry. Converted to draft by mistake.

ncw:
> I don't think that's necessary since we don't have it for the mountlib too.
> 
> Going through this, I wonder if we should move the code to the server wait function so it can be used for all serve commands too instead of only rcd but that's beyond the scope of the issue.

I think that would be a good idea.
 
> The changes look good otherwise, thanks.

I will merge them as-is so we get them in.

Would you like to refactor it @darthShadow or  @Navdevl  along the above lines?

Navdevl:
@ncw I'd love to work on this. I am very new to the language as well as to the rcloud codebase. If I could get some little directions, I can definitely pick it up on the refactoring.

darthShadow:
It won't be as straight-forward as this simply because there are multiple files to touch so please create a new issue so that we can get a list of the files that need modifying. Once we have identified the files, the change in them should be straight-forward.

ncw:
I've made a new issue here @Navdevl https://github.com/rclone/rclone/issues/5117 if you want to have a go at it. I've assigned you to it @darthShadow to be project member in charge :-)

Navdevl:
@ncw Thank you for raising an issue. I'll follow up on that and work on the part.

Issue #3881, opened 2020-01-15
Confusing "about" space units


elig0n:
Space is shown in G/T/M units notation appearing to be measured in GB/TB/MB/etc but is actually GiB/TiB/MiB/etc See https://en.wikipedia.org/wiki/Binary_prefix#Consumer_confusion
Please use the appropriate notation to prevent confusion

animosity22:
Can you give an example of what you are referring to?

elig0n:
@animosity22 I specifically means this command and its output:

    felix@gemini:~$ rclone about GD:
    Used:    106.295T
    Trashed: 0
    Other:   159.438M

One who would see the 'T' and 'M' designations would naturally think they mean "Terabyte' and 'Megabyte' respectively, but in reality they are 'Tebibyte' and 'Mebibyte'.

ncw:
The units are indeed in "binary SI" multiples of 1024

This is pretty common for disk/memory sizes.

Rclone is 100% consistent in always using the "binary SI" units whether you are inputting `--buffer-size`, or looking at disk stats or setting bandwidth limits.

```
$ rclone about .
Total:   466.771G
Used:    405.587G
Free:    37.407G
```

vs

```
$ df -h .
Filesystem                    Size  Used Avail Use% Mounted on
/dev/mapper/xubuntu--vg-root  467G  406G   38G  92% /
```

It would be fairly easy to make rclone put an `i` after every unit so you'd get

```
$ rclone about .
Total:   466.771Gi
Used:    405.587Gi
Free:    37.407Gi
```

However I suspect that would result in people creating even more issues!

Rclone could change to using 1000 multiples but that would confuse people when then use memory units `--buffer-size`, `--chunk-size` etc.

I could implement an `--si` flag to control that, however rclone stores the units in config files etc so changing the meaning of the units would be a bad idea...

If you want to avoid confusion you can always use the `--json` flag to about!

```
$ rclone about --json .
{
	"total": 501192073216,
	"used": 435492089856,
	"free": 40169467904
}
```

I don't think there is a good solution to this other than maybe adding an `--si` flag to `rclone about` only.

animosity22:
Just re-reading this as I think the final agreement was not to make a change to rclone based on @ncw comments so I was going to close this out.

Issue #1890, opened 2017-12-07
Human-readable sizes for ls


cowwoc:
rclone 1.38

As a follow-up to https://github.com/ncw/rclone/issues/1303, please add a `--human` option to `ls` similar to the `tree` command so we can view file sizes in human-readable format.

ncw:
I'd like to re-implement the whole ls command, but I don't want to break everyone's scripts!  Nice idea though.

And checkout `rclone ncdu` for the TUI version!

albertony:
Available in latest beta, please check it out @cowwoc. 🙈

timotheecour:
the flag that was added is --human-readable, not `--human`, eg this works: `rclone ls --human-readable path`

Issue #428, opened 2016-04-09
Display in MB's instead of kByte


sbcrumb:
Is there a way I can simply display results in -v and then it is done copying in MB's instead of kBytes?
Something little more human readable?
Transferred:   193651128918 Bytes (19959.84 kByte/s)
Errors:                 1
Checks:               232
Transferred:          114
Elapsed time:  2h37m54.6s

pmwoodward3:
+1 I would love to have this too.

ncw:
This should be easy to do!

HenrikBengtsson:
Note the difference between IEC and SI unit standards:

IEC: KiB (1024 bytes), MiB (1024^2 bytes), GiB (1024^3 bytes), TiB (1024^4 bytes), ...
SI: kB (1000 bytes), MB (1000^2 bytes), GB (1000^3 bytes), TB (1000^4 bytes), ...

There are lots of home-cooked alternatives out there (I'm guilty too), but the above two are the two main standards and adopted by most OSes etc.

From the example output it looks like your kByte/s is KiB/s.  If you go with KB/s, please change to base 10.

My $0.02

ncw:
I've done this.  You can check it out in: http://pub.rclone.org/v1.29-48-ge2788aa%CE%B2/

The units are the binary units as used in the rest of rclone.

PR #5087, opened 2021-03-02
Use binary/IEC prefix for base-2 and decimal/SI prefix for base-10 numbers

### LINKED ISSUE #5068: Use IEC standard unit notation (KiB, MiB, GiB, TiB)
Terminology:
- Binary prefixes:
  - According to the [ISO/IEC 80000-13](https://en.wikipedia.org/wiki/IEC_80000-13) standard
  - Power of 2 scale.
  - Kibibyte (KiB), Mebibyte (MiB) etc.
  - Also known as: IEC units
- Decimal prefixes:
  - Used in the [SI](https://en.wikipedia.org/wiki/International_System_of_Units) unit system (a.k.a. the metric system).
  - Power of 10 scale.
  - Kilobyte (kB), Megabyte (MB), etc
  - Also known as: SI units

Suggestions (feature request):
- Always use binary/IEC prefixes for base-2 units (KiB, MiB, GiB, TiB).
- Use decimal/SI prefixes only when specifically referring to base-10 units (kB, MB, GB, TB).
   - Kilo (1000), and lower, have lowercase prefix (kB), higher have uppercase prefix (MB).
- Use uppercase B for Byte and lowercase b for bit
- Use singular form for transfer rate, such as 10 MiByte/s not 10 MiBytes/s (if allowing MiByte notation at all, see below).
   - Plural form when writing out complete words like "bytes per second" and "bits per second".
- Must be clearly documented, and perhaps this is a good incentive to start a style guide document for rclone.

Current status:
- Rclone is today consistently using the SI standard notation in documentation, log output etc, but has [documented](https://rclone.org/docs/#options) that it represents binary units:

    > Options which use SIZE use kByte by default. However, a suffix of b for bytes, k for kBytes, M for MBytes, G for GBytes, T for TBytes and P for PBytes may be used. These are the binary units, e.g. 1, 2\*\*10, 2\*\*20, 2\*\*30 respectively.

Additional notes, questions, decision points:
- Single letter versions: k/K, M, G etc. Used throughout rclone, and must probably still be used - and then e.g. M should mean base 2 MiB and not base 10 MB? In case of K, use upper- or lowercase? If short for KiB then uppercase K makes sense.
  - An alternative could be as [ccache](https://ccache.dev/manual/4.2.html#_common_options) uses: Single letter prefix as decimal units, and specific prefixes Ki, Mi, Gi, Ti for binary units.
- Currently a notation such as "GBytes" and "MByte/s" are used, e.g. in stats display after transferring a file.
    - This seems not to be part of standard. For bits there is Mbit/s and Mibit/s, but not similar for byte.
    - I would say, ideally, not to use this notation, stick with MiB/s and Mibit/s. But perhaps it is well-established enough, at least in existing rclone, that its acceptable? Compared to the standard short notation (MiB/MB) it has the advantage of avoiding any bit/byte confusion (although consistent use of Mibit for bits and MiB for bytes would should with time avoid confusion as well).
    - If allowed, should then use MiByte for base 2 units and MByte for base 10 units.
- With the above two points we would get 4 formats for IEC and 4 for SI:
    - Single (unstandard): M
    - Short: MiB, MB
    - Intermediate (unstandard): MByte, MiByte
    - Long: Mebibyte, Megabyte
- Could be made configurable, e.g. an `--si` unit to switch from MiB to MB, but will probably [not do this](https://github.com/rclone/rclone/pull/5053#issuecomment-785346251).
- Could result in breaking change for people with wrappers relying on rclone's output.

Discussed here, and there are a lot of relevant input in the comments:
https://github.com/rclone/rclone/pull/5053

See also related issues:
https://github.com/rclone/rclone/issues/3881
https://github.com/rclone/rclone/issues/1763
https://github.com/rclone/rclone/issues/1890
https://github.com/rclone/rclone/issues/428

External references:
- https://en.wikipedia.org/wiki/International_System_of_Units
- https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table
- https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
- https://wiki.ubuntu.com/UnitsPolicy
- https://docs.microsoft.com/en-US/style-guide/a-z-word-list-term-collections/term-collections/bits-bytes-terms

Work notes:
- Tree command uses third party module (github.com/a8m/tree) which shows single character prefixes when option for human readable units are specified. It uses base 2 values, with uppercase K for KiB (see https://github.com/a8m/tree/pull/10). Example: `rclone.exe tree --size --human .`
- Keep config values for DataRateUnit and --stats-unit as is: plural and lowercase "bits"/"bytes".
- Change output of datarate to "bit/s" and "Byte/s" (singular, lowercase b for bit and uppercase b for byte). Output of size still Bytes (plural) as before. Test --stats with --stats-unit=bits|bytes.

albertony:

Rclone is already consistently using binary units for sizes, as [documented](https://rclone.org/docs/#options):
> Options which use SIZE use kByte by default. However, a suffix of b for bytes, k for kBytes, M for MBytes, G for GBytes, T for TBytes and P for PBytes may be used. These are the binary units, e.g. 1, 2<sup>10</sup>, 2<sup>20</sup>, 2<sup>30</sup> respectively.

But: It uses the SI standard notation. This is, strictly speaking, describing decimal (base-10) units. Although traditionally used for binary (base-2, or rather base-1024) units as well, it may lead to uncertainty about whether values are actually in decimal or binary units.

This PR changes to consistent use of IEC standard notation when referring to size and rate, and any other values in binary/base-2 scale, and SI notation for counts and other values in decimal/base-10 units. This should make the definition explicit, and avoid any confusion. There are mostly changes in documentation, command output and logs, 

**EDIT:** Updated several times after comments (discussions) below, so those may be "outdated" now.

**EDIT:** Split out `--human-readable` option into separate PR: https://github.com/rclone/rclone/pull/5237

##### Work in progress
- Things are looking up...
  - Still a few decisions to make, but current version is coming close to a suggestion that can be considered, at least!
  - **Beta build**: If you want to have a quick test without building it, I've released a beta here:
    https://github.com/albertony/rclone/releases/tag/v1.56.0-units03
- Biggest decision points (see examples below):
  - Use `MiByte`, `MiB` or `Mi` as "primary" variant for "non-tabular" output?
     - Uses the `Mi` format in tabular output, where space is limited,such as results from `ls` with `--human-readable`, but what about more "non-tabular" output, from `size`, `about` etc? EDIT: Moved to separate PR!
     - Currently more or less consistently using the `Mi` format everywhere for command output, and using `MiByte` in some log messages, most importantly the stats output - where previously `MByte` was used.
     - Should we switch to `MiB`? Instead of `MiByte`? Also instead of `Mi` for non-tabular output? It seems more "standardized" than `MiByte`, but the latter avoids confusion regarding bit vs byte.

- Remaining work:
  - Document decisions in a new style guide document for rclone.
    - ~~There is a commit adding a codestyle chapter in existing contributing doc, but it is just a rough sketch for now, so *that* part is definitely not done!~~
    - Have written down some details in more of a "documentation-style" later on in this pr's descriptions, with an intention of reusing it for documentation later.

##### Main changes
- Consistent use of binary/IEC prefix, whenever referring to base-2 units.
  - Size values (that rclone has control over) are always in base-2 units. There are exceptions where external dependencies etc use base-10 - if any of this ends up in rclone input/output then decimal/SI prefix must be used to clearly identify this.
  - Input:
    - Accepts single letter suffix such as `M` as before (backwards compatible), and interpret them as binary units as before, but now also accepts `Mi` and `MiB` forms, without space after value
    - Examples: `--chunk-size 20M`, `--chunk-size 20Mi`, `--chunk-size 20MiB`
  - Output:
    - Previous uses of single-letter SI prefix symbol, e.g. `M`, in rclone with an implicit meaning of binary scale, are now shown as corresponding IEC prefix symbol, e.g. `Mi` - for values that are base-2.
    - Using different forms depending on situation, as before, but now with explicit binary prefix symbol: Short form `Mi`, medium form `MiB` or long form `MiBytes`.
- Consistent use of decimal/SI prefix whenever referring to base-10 units.
  - Count values always in base-10 units.
  - Will use the existing single-letter suffixes, e.g. `M`, since they are the SI standard symbols for decimal units.

##### Additional changes
- General spelling and cleanup of size and rate units (continuation of #5053)
  - Use uppercase B for Byte and lowercase b for bit
  - Singular form of suffix, such as "10 MiByte", and for transfer rate, such as 10 MiByte/s (not 10 MiBytes/s).
    - Plural form when writing out complete words like "bytes per second" and "bits per second".
  - Stats output changed to "bit/s" and "Byte/s" (singular, lowercase b for bit and uppercase b for byte), not "bits/s" and "bytes/s".
    - Kept config values for DataRateUnit and --stats-unit as is: plural and lowercase "bits"/"bytes".
- Use decimal suffix for counts
  - Fixes issue #5126, describing how `ncdu` previously used binary suffix.
  - Also found one similar case in `about` command (described below).

##### Breaking changes
- Potentially breaking changes for people with wrappers relying on rclone's output related to sizes (`ls`, `about`, `size`, logs, stats etc).

##### Not including
- Command `tree` shows single character prefixes when option for human readable units are specified. It uses base 2 values, with uppercase K for KiB. Since this is handled within the third party module code (github.com/a8m/tree), it will not be changed to showing `Ki` etc consistent with other commands.
- Optional decimal units. #1763 proposes an option `--si` to be able to switch rclone from using binary to use decimal units for sizes. This would be in line with linux tools such as df and ls, which also has such an option. The current pull request, making the binary unit notation more explicit, e.g. "MiB" vs "MB", but still there is a consern regarding input and the single letter variants: "you really don't want chunk_size 16M in a config file to suddenly be a different size".
- Fix for #5169: Size options can not be set to value less than 1K.

##### Examples
- Stats output:
  - Before:
    ```
    Transferred:      189.522M / 189.522 MBytes, 100%, 100.146 MBytes/s, ETA 0s
    ```
  - Now:
    ```
    Transferred:      189.522Mi / 189.522 MiByte, 100%, 384.928 MiByte/s, ETA 0s
    ```
  - Possible alternative:
    ```
    Transferred:      189.522 MiB / 189.522 MiB, 100%, 384.928 MiB/s, ETA 0s
    ```
- Stats output with `--stats-unit bits`:
  - Before:
    ```
    Transferred:      189.522M / 189.522 MBytes, 100%, 3.525 GBits/s, ETA 0s
    ```
  - Now:
    ```
    Transferred:      189.522Mi / 189.522 MiByte, 100%, 2.460 Gibit/s, ETA 0s
    ```
  - Possible alternative:
    ```
    Transferred:      189.522 MiB / 189.522 MiB, 100%, 2.460 Gibit/s, ETA 0s
    ```
- Log output:
  - Before:
    ```
    NOTICE: npp/x64/plugins/NppConverter/NppConverter.dll: Skipped copy as --dry-run is set (size 152.914k)
    ```
  - Now:
    ```
    NOTICE: npp/x64/plugins/NppConverter/NppConverter.dll: Skipped copy as --dry-run is set (size 152.914Ki)
    ```
  - Possible alternative:
    ```
    NOTICE: npp/x64/plugins/NppConverter/NppConverter.dll: Skipped copy as --dry-run is set (size 152.914 KiB)
    ```

##### Details (documentation)

Terminology:
- Suffix
  - Used with meaning "unit suffix of value": Symbol or label following a value to indicating the unit and/or scaling.
  - Examples: "100 MB", where "MB" is the suffix.
- Prefix:
  - Used with meaning "prefix of unit": A specifier prepended to a base unit to indicate multiples (or fractions) of it.
  - Also represented by a symbol, e.g. 'k' for kilo, 'M' for mega.
  - Examples: "kilobyte", "kilobit", "kilogram", where "kilo" is the prefix, or "MB" where "M" is the prefix.
  - For multiples of bytes (and bits), what wikipedia calls [multiple-byte units](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units), there are two main groups of prefixes for multipliers powers of 10 and for powers of 2.
- Binary prefix
  - Also known as: IEC prefix.
  - Unit prefixes for power of 2 scale.
  - According to the [ISO/IEC 80000-13](https://en.wikipedia.org/wiki/IEC_80000-13) standard.
  - Examples: Kibibyte (KiB), Mebibyte (MiB) etc.
      - 1 KiB is 1024 B, 1 MiB is 1024 KiB, etc.
      - Note: IEC uses uppercase 'K', in contrast to SI where 'k' is lowercase (but all larger are uppercase).
- Decimal prefix:
  - Also known as: SI prefix.
  - Unit prefixes for power of 10 scale, but (historically) also used for power of 2 scale creating ambiguity.
  - Used in the [SI](https://en.wikipedia.org/wiki/International_System_of_Units) unit system (a.k.a. the metric system).
  - Examples: Kilobyte (kB), Megabyte (MB), etc
    - 1 kB is 1000 B, 1 MB is 1000 kB, etc.
    - Note: SI uses lowercase 'k', uppercase 'M' and larger.

Binary unit suffix variants in rclone:
  - Base: `B`
    - To indicate that a number is bytes, and without any scaling (no prefix).
    - Input options may interpret values without suffix as `KiB`, then suffix `B` must be specified to set a value less than `1KiB`.
  - Single: `K`, `M`, `G`, `T`, `P`, `Ki`, `Mi`, `Gi`, `Ti`, `Pi`
    - Examples: "100M", "100Mi"
    - Note: Immediately following a numeric value (no space between).
    - Used when a single word is convenient, such as option input values, and when space is limited such a in output of tabular data.
    - The single-letter variants without trailing "i" are actually SI decimal prefixes, but in rclone they will be interpreted as binary units depending on context, for convenience - and backwards compatibility.
    - In input the trailing i is allowed but optional, and does not change the meaning (`M` == `Mi` == `MiB`): Sizes are always assumed to be base-2. Casing should not matter either, lowercase suffix "k" still means KiB, as in 1,024 bytes, and not kB, as in 1,000 bytes.
        - Currently no input for base-10 values accept suffix, only raw numbers (e.g. `--drive-list-chunk` takes int value 100-1000), but if there would be any then `M` suffix would have to be treated as base-10 value for these and base-2 value for sizes, so interpretation of *single*-letter suffix depends on the context.
  - Short: `KiB`, `MiB`, `GiB`, `TiB`, `PiB`
    - Example: "100 MiB"
    - Note: Space between the value and the unit.
    - Limited or no use in rclone?
  - Medium: `Byte`, `KiByte`, `MiByte`, `GiByte`, `TiByte`, `PiByte`
    - Example: "100 MiByte"
    - Always singular form, "10 MiByte" not "10 MiBytes".
    - Primary variant for more descriptive output, log messages etc, where space is not that limited.
    - Currently a notation such as "GBytes" and "MByte/s" are used, e.g. in stats display after transferring a file.
    - This seems not to be part of standard. For bits there is Mbit/s and Mibit/s, but not similar for byte.
    - I would say, ideally, not to use this notation, stick with MiB/s and Mibit/s. But perhaps it is well-established enough, at least in existing rclone, that its acceptable? Compared to the standard short notation (MiB/MB) it has the advantage of avoiding any bit/byte confusion (although consistent use of Mibit for bits and MiB for bytes would should with time avoid confusion as well).
  - Long: `Byte`, `Kebibyte`, `Mebibyte`, `Gibibyte`, `Tebibyte`, `Pebibyte`
    - Example: "100 mebibytes"
    - Not used in rclone (except Byte).
    - Plural form as suited, e.g. "10 mebibytes per second".

##### Other examples for comparison

- rsync, ls, df, dh

  All print raw byte count by default, have argument `--human-readable` which shows single letter suffix as proper binary unit (uppercase K for 1024, at least `du`), with `--si` to show single letter suffix as decimal units (lowercase k for 1000, at least `du`).

- git
  
  Short-form binary suffix (`MiB`):

  > Writing objects: 100% (82/82), 8.27 KiB | 58.00 KiB/s, done.

- pacman
  
  Short-form binary suffix (`MiB`)

  > pacman -Syu
  > :: Synchronizing package databases...
  > testing 21.8 KiB 7.09 MiB/s 00:00 [######################] 100%
  > core 133.8 KiB 257 KiB/s 00:01 [######################] 100%
  > extra 1629.5 KiB 1045 KiB/s 00:02 [######################] 100%

- wget

   Single letter suffix for size, base-2 interpretation, short-form SI-style suffix for speed (`MB/s`) (also base-2?)

  > rclone-v1.55.0-windows-amd64.zip  100 
  > [============================================================>]  13.58M  6.97MB/s    in 2.0s

- curl

  Single-letter SI-style suffix (lowercase `k`) with base-2 interpretation

  >   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current Dload  Upload   Total   Spent    Left  Speed
  > 100 13.5M  100 13.5M    0     0  6892k      0  0:00:02  0:00:02 --:--:-- 6896k

- ccache

  Size options accept suffix k, M, G, T and interpret it as decimal units, and Ki, Mi, Gi, Ti interpreted as binary units.
  Output is always shown in decimal units with corresponding single-letter suffix.
  
##### External references
- https://en.wikipedia.org/wiki/International_System_of_Units
- https://en.wikipedia.org/wiki/Data-rate_units#Conversion_table
- https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
- https://wiki.ubuntu.com/UnitsPolicy
- https://docs.microsoft.com/en-US/style-guide/a-z-word-list-term-collections/term-collections/bits-bytes-terms

Fixes #5068

See also:
- #5053
- #1763
- #5126
- #1890
- #3881
- #428

#### Checklist


backend/b2/b2.go
```diff
@@ -116,7 +116,7 @@ in the [b2 integrations checklist](https://www.backblaze.com/b2/docs/integration
 
 Files above this size will be uploaded in chunks of "--b2-chunk-size".
 
-This value should be set no larger than 4.657GiB (== 5GB).`,
+This value should be set no larger than 4.657 GiB (== 5 GiB).`,
```
ncw, line 119: Note that all the units in the B2 docs are in MB not MiB! So this is incorrect.

backend/b2/upload.go
```diff
@@ -231,7 +231,7 @@ func (up *largeUpload) transferChunk(ctx context.Context, part int64, body []byt
 		// The number of bytes in the file being uploaded. Note that
 		// this header is required; you cannot leave it out and just
 		// use chunked encoding.  The minimum size of every part but
-		// the last one is 100MB.
+		// the last one is 100 MiB.
```
ncw, line 234: This should also be 100 MB

backend/s3/s3.go
```diff
@@ -1271,7 +1271,7 @@ See: https://github.com/rclone/rclone/issues/4673, https://github.com/rclone/rcl
 const (
 	metaMtime   = "Mtime"     // the meta key to store mtime in - e.g. X-Amz-Meta-Mtime
 	metaMD5Hash = "Md5chksum" // the meta key to store md5hash in
-	// The maximum size of object we can COPY - this should be 5GiB but is < 5GB for b2 compatibility
+	// The maximum size of object we can COPY - this should be 5 GiB but is < 5 GiB for b2 compatibility
```
ncw, line 1274: This is incorrect - the original line was correct (see notes on b2 above!)

cmd/delete/delete.go
```diff
@@ -36,8 +36,8 @@ If you supply the |--rmdirs| flag, it will remove all empty directories along wi
 You can also use the separate command |rmdir| or |rmdirs| to
 delete empty directories only.
 
-For example, to delete all files bigger than 100MBytes, you may first want to check what
-would be deleted (use either):
+For example, to delete all files bigger than 100 mebibytes, you may first want to
```
ncw, line 39: Not a fan of `mebibytes`! I'd prefer 100 MiB or 100 MiByte I think.

What do you think?

docs/content/b2.md
```diff
@@ -406,7 +406,7 @@ Cutoff for switching to chunked upload.
 
 Files above this size will be uploaded in chunks of "--b2-chunk-size".
 
-This value should be set no larger than 4.657GiB (== 5GB).
+This value should be set no larger than 4.657 GiB (== 5 GiB).
```
ncw, line 409: This is an autogenerated repro of the mistake above!

docs/content/filtering.md
```diff
@@ -564,17 +564,17 @@ remote or flag value. The fix then is to quote values containing spaces.
 ### `--min-size` - Don't transfer any file smaller than this
 
 Controls the minimum size file within the scope of an rclone command.
-Default units are `kBytes` but abbreviations `k`, `M`, or `G` are valid.
+Default units are `KiByte` but abbreviations `K`, `M`, `G` or `P` are valid.
```
ncw, line 567: No T?

docs/content/filtering.md
```diff
@@ -564,17 +564,17 @@ remote or flag value. The fix then is to quote values containing spaces.
...
 
-E.g. `rclone ls remote: --min-size 50k` lists files on `remote:` of 50kByte
+E.g. `rclone ls remote: --min-size 50k` lists files on `remote:` of 50 KiByte
 size or larger.
 
 ### `--max-size` - Don't transfer any file larger than this
 
 Controls the maximum size file within the scope of an rclone command.
-Default units are `kBytes` but abbreviations `k`, `M`, or `G` are valid.
+Default units are `KiByte` but abbreviations `K`, `M`, `G` or `P` are valid.
```
ncw, line 575: No T?

ncw (commented):
I've given this the quick once over.

There are a couple of errors to do with B2 measuring everything in SI units but otherwise looks great - well done :-)

cmd/delete/delete.go
albertony: I agree. Failed attempt at writing a bit more more "natural language".. :) Like MiB most, as that seems to be *the* standard unit prefix.

albertony:
> I've given this the quick once over.

Great, thank you!

> There are a couple of errors to do with B2 measuring everything in SI units

Ah, that was a gotcha when doing brute force search/replace! Was planning on going through the details of all changes later, I like to think I would have found them then - but probably just wishful thinking... So I appreciate your feedback. Just hope I remember to not repeat the same errors later on, when I just want to perform some quick final touches or something like that... :laughing:

Have fixed according to all your comments, and will get the checks green, and then have a thorough "review" myself (and also finish the CONTRIBUTING.md text, which is just a quick draft as of now). I might ping you later on (probably in days), when closer to something "mergeable"...

ncw:
Let me know when you are ready for a review :-)

albertony:
Added commit "Use decimal prefixes for counts" to make ncdu file counts reported with decimal/SI unit kilo (1000) and not binary/IEC kilo (1024). Added it as a separate type SizeSuffixDecimal to avoid cluttering existing SizeSuffix code. Quick question, @ncw, do you agree or should I reconsider this?

Note: Rest of this PR not completed, so not ready for a full review yet..

PS:

The ncdu command will now show "1M" as size and mean 1024\*1024 bytes, and "1M" as count and mean "1000\*1000" files.. I've seen uses of "Mi" as suffix instead of "M" for binary units, I have not decided what I prefer, but I know you said you are not fond of this:

> When configuring things the unit isn't normally used, so rclone will say --max-size 4M - I'm not sure I want to change this to --max-size 4Mi - I think 4M here is an abbreviation for 4MiByte or 4MiByte/s and that is fine.

ncw:
> Added commit "Use decimal prefixes for counts" to make ncdu file counts reported with decimal/SI unit kilo (1000) and not binary/IEC kilo (1024). Added it as a separate type SizeSuffixDecimal to avoid cluttering existing SizeSuffix code. Quick question, @ncw, do you agree or should I reconsider this?

This looks OK. I note you ported the setting part of SizeSuffixDecimal which I don't think we need strictly speaking. With the Setting part then it becomes elegible to be a config parameter (backend or global) which may/may not be useful!

> Note: Rest of this PR not completed, so not ready for a full review yet..
> 
> Edit: One more thing:
> 
> The ncdu command would then show "1M" as size and mean 1024*1024 bytes, and "1M" as count and mean "1000*1000" files.. We have the possibility of using "Mi" as suffix instead of "M" for binary units, which is what the IEC standard specifies, and although still unsure, I'm leaning towards preferring that to make it explicit in cases like with ncdu.. You said:

I think that sounds find with `ncdu`

I just had a play with actual `ncdu`. It reports `KiB` by default but if you use the `--si` flag then it reports `KB`.

> > When configuring things the unit isn't normally used, so rclone will say --max-size 4M - I'm not sure I want to change this to --max-size 4Mi - I think 4M here is an abbreviation for 4MiByte or 4MiByte/s and that is fine.
> 
> I don't know if you are against Mi in general, or just differentiating between M and Mi in input?

My main purpose is not to break existing config files! These are full of things like `chunk_size = 64M` - this is parsed by `SizeSuffix`.

Having two distinct types for `SizeSuffix` and `SizeSuffixDecimal` is a good idea so we know what is what and where. I don't think we could have an `--si` flag for rclone proper if it meant it changes the meanings of all the config file settings!

> What about showing "Mi" etc in output, but allow "M" as input with implied binary meaning for size (and decimal for count, if there are any such options)?  ...
> 
> `Transferred: 140.943Mi / 140.943 MiByte, 100%, 3.516 GiByte/s, ETA 0s`

I think that will probably be OK just as long as we don't break the parsing of it without the `i`!

albertony:
> just had a play with actual `ncdu`. It reports `KiB` by default

albertony:
@ncw If you have time to take a new look, that would be great. I have been adding some stuff, and done several rounds of refactoring and rebasing, over the last few days... I'm having big trouble declaring this to be *done* from my part, but I fear I will never be able to do that 100%... See for example the different alternatives in the example paragraph (decisions, decisions, decisions....) :thinking: But I *think* the current state is pretty good, so a fresh pair of 👀 would be perfect!

docs/content/cache.md
```diff
@@ -364,11 +364,11 @@ will need to be cleared or unexpected EOF errors will occur.
 - Default:     5M
 - Examples:
     - "1m"
```
zanna-37, line 366: minor: should probably be `1M`

albertony: Probably.. Although it is not *wrong*: "1m", "1M", "1Mi", "1MI", "1mi", "1MiB", "1mib", "1mIB"...etc.. all are accepted and with identical results. All other examples use the "rclone standard" single letter uppercase, though, so I will update this also.
Thank you for reporting, really great that you are willing to have a look!

zanna-37: I noticed you corrected it in `backend/cache/cache.go` line `101` so I thought it was worth pointing out, even if very minor.
By the way thank _you_ for your amazing work!

albertony: > I noticed you corrected it in backend/cache/cache.go line 101 so I thought it was worth pointing out, even if very minor.

Oh, right, I agree!

ncw:
I think what we should do is merge the first 4 or so commits of this and put the human readable stuff in a new PR. I'm less convinced of the human readable stuff and I think it needs a bit more discussion, but I would like to merge the binary/IEC changes this PR was originally about now.

So if you could make a copy of this branch, reset it to the appropriate place, I'll give it a last once over then we can merge it. Then we can discuss the human readable stuff in a new PR - how does that sound?

albertony:
> I think what we should do is merge the first 4 or so commits of this and put the human readable stuff in a new PR. I'm less convinced of the human readable stuff and I think it needs a bit more discussion, but I would like to merge the binary/IEC changes this PR was originally about now.
> 
> So if you could make a copy of this branch, reset it to the appropriate place, I'll give it a last once over then we can merge it. Then we can discuss the human readable stuff in a new PR - how does that sound?

DONE!
The human readable moved to PR: https://github.com/rclone/rclone/pull/5237

ncw (approved):
This looks great now. Apologies for the delay in approval - I should have done that a couple of weeks ago but it slipped off my radar.

Please merge.

PR #5088, opened 2021-03-02
WIP / RFC: s3: Use ListObjectsV2 for much faster listings

felixbuenemann:
This PR is WIP, since it was only manually tested against real AWS S3 endpoints.

Additional research is required to check if a fallback from `ListObjectsV2` to `ListObjects` is required for S3-compatible storage, either automatic, through a cli switch or default per s3 backend type.

It would also be great to get feedback from @ncw on what is the preferred option.

Using `ListObjectsV2` with a continuation token is about **5-6x faster** than `ListObjects` with a marker.

No, but there were several posts related to slow listings, which usually resolved around using `--no-modtime` or `--use-server-modtime`, but using `ListObjectsV2` API massively speeds up the response time from the AWS S3 API.

Example:

```shell
$ time sudo -iu rclone /usr/bin/rclone ls s3:my-large-bucket --fast-list --use-server-modtime | wc -l
71689

real	0m56.609s
user	0m2.363s
sys	0m0.407s

$ time sudo -iu rclone /tmp/rclone ls s3:my-large-bucket  --fast-list --use-server-modtime | wc -l
71689

real	0m9.502s
user	0m2.355s
sys	0m0.298s
```

Test was done on an aws t4g.nano (linux-arm64) in eu-west-1 against a bucket in eu-central-1, but I've seen similar results on an aws c5.xlarge (linux-amd64)  in eu-central-1 against the same bucket in eu-central-1.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -64,6 +64,7 @@ func init() {
 		Options: []fs.Option{{
 			Name: fs.ConfigProvider,
 			Help: "Choose your S3 provider.",
+			// NB if you add a new provider here, then add it in the setQuirks function
 			Examples: []fs.OptionExample{{
 				Value: "AWS",
 				Help:  "Amazon Web Services (AWS) S3",
@@ -1204,6 +1205,22 @@ In Ceph, this can be increased with the "rgw list buckets max chunk" option.
 			Default:  1000,
 			Advanced: true,
 		}, {
+			Name: "list_version",
+			Help: `Version of ListObjects to use: 1,2 or 0 for auto.
+
+When S3 originally launched it only provided the ListObjects call to
+enumerate objects in a bucket.
+
+However in May 2016 the ListObjectsV2 call was introduced. This is
+much higher performance and should be used if at all possible.
+
+If set to the default, 0, rclone will guess according to the provider
+set which list objects method to call. If it guesses wrong, then it
+may be set manually here.
+`,
+			Default:  0,
+			Advanced: true,
+		}, {
 			Name: "no_check_bucket",
 			Help: `If set, don't attempt to check the bucket exists or create it.
 
@@ -1357,6 +1374,7 @@ type Options struct {
 	UseAccelerateEndpoint bool                 `config:"use_accelerate_endpoint"`
 	LeavePartsOnError     bool                 `config:"leave_parts_on_error"`
 	ListChunk             int64                `config:"list_chunk"`
+	ListVersion           int                  `config:"list_version"`
 	NoCheckBucket         bool                 `config:"no_check_bucket"`
 	NoHead                bool                 `config:"no_head"`
 	NoHeadObject          bool                 `config:"no_head_object"`
@@ -1579,6 +1597,7 @@ func s3Connection(ctx context.Context, opt *Options, client *http.Client) (*s3.S
 	if opt.Provider == "Scaleway" && opt.MaxUploadParts > 1000 {
 		opt.MaxUploadParts = 1000
 	}
+	setQuirks(opt)
 	awsConfig := aws.NewConfig().
 		WithMaxRetries(ci.LowLevelRetries).
 		WithCredentials(cred).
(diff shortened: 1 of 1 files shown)


felixbuenemann:
@ncw Do you have insights on wether any of the s3 compatible backends do not support `ListObjectsV2` (`list-type=2`)?

felixbuenemann:
@ncw It looks like some S3 compatible implementations like NetApp Storage Grid and Ceph RGW do not support ListObjectsV2 in older versions, so some kind of fallback is needed.

What is your preference for implementing a fallback?
1. Backend config option like `list_type = 1` (default 2)
2. Command Line option like `--s3-list-objects-type 1`
3. Automatic fallback on errors like `XNotImplemented`

I'm not sure how streamlined the error responses are, so my preference would be the backend option, which could also have a different default for certain s3 compatible flavours when configured through the wizard.

ncw:
> Using `ListObjectsV2` with a continuation token is about **5-6x faster** than `ListObjects` with a marker.

That is great news!

According to https://docs.aws.amazon.com/AmazonS3/latest/API/WhatsNew.html the ListObjectsV2 API was introduced in May 4, 2016 which is long enough ago that nearly everything should support it, but not long enough ago that everything will :-(

So I think support for this will need to be a config parameter

> Do you have insights on whether any of the s3 compatible backends do not support `ListObjectsV2` (`list-type=2`)?

Not off the top of my head, but we do have 5 providers in the daily integration tests which are easy to test, and I have quite a few more providers that I have configs for.

> @ncw It looks like some S3 compatible implementations like NetApp Storage Grid and Ceph RGW do not support ListObjectsV2 in older versions, so some kind of fallback is needed.

Ceph is popular with rclone users - it looks like v2 support wasn't merged until Mid 2019 so this will be a problem.
 
> What is your preference for implementing a fallback?
> 
> 1. Backend config option like `list_type = 1` (default 2)
> 2. Command Line option like `--s3-list-objects-type 1`

Note that a backend config automatically makes a command line option. So a `list_type` option becomes `--s3-list-type`.

> 3. Automatic fallback on errors like `XNotImplemented`
> 
> I'm not sure how streamlined the error responses are, so my preference would be the backend option, which could also have a different default for certain s3 compatible flavours when configured through the wizard.

Hmm, from this bug report on CEPH https://tracker.ceph.com/issues/37801 it looks like the error there is just returning the wrong kind of listing as in CEPH just ignores the `list-type=2` parameter.

Hmm, does this mean that we need a bucket with more than 1000 things in it to test properly? That is a really annoying failure mode!

We could potentially detect this when we get to here `return errors.New("s3 protocol error: received listing with IsTruncated set, no NextContinuationToken found")` and use that as a signal to switch to v1 listings.

The safe thing would probably be to have `list_type 0` mean `auto` in which case rclone uses its knowledge about Providers to set `2` for `AWS` and other supported providers and set `1` for everything else. CEPH will be the annoying one as there will be CEPHs out there which do support v2 and ones which don't.

felixbuenemann:
> > 3. Automatic fallback on errors like `XNotImplemented`
> > 
> > I'm not sure how streamlined the error responses are, so my preference would be the backend option, which could also have a different default for certain s3 compatible flavours when configured through the wizard.
> 
> Hmm, from this bug report on CEPH https://tracker.ceph.com/issues/37801 it looks like the error there is just returning the wrong kind of listing as in CEPH just ignores the `list-type=2` parameter.
> 
> Hmm, does this mean that we need a bucket with more than 1000 things in it to test properly? That is a really annoying failure mode!
> 
> We could potentially detect this when we get to here `return errors.New("s3 protocol error: received listing with IsTruncated set, no NextContinuationToken found")` and use that as a signal to switch to v1 listings.
> 
> The safe thing would probably be to have `list_type 0` mean `auto` in which case rclone uses its knowledge about Providers to set `2` for `AWS` and other supported providers and set `1` for everything else. CEPH will be the annoying one as there will be CEPHs out there which do support v2 and ones which don't.

It would be great if we could do a fallback with zero overhead (like repeated requests), but I'm not sure the AWS Go SDK will allow us to do that.

If we do the ListObjectsV2 call, we receive an `s3.ListObjectsV2Output` instead of `s3.ListObjectsOutput`, so we are lacking the `NextMarker` struct field required for the V1 happy path.

At this point we would have to retry the API call with V1, which will slow things down.

So unless there is a clever way around that I think a backend option with sane defaults for the different providers is the better choice and makes the behaviour more predictable.

ncw:
> So unless there is a clever way around that I think a backend option with sane defaults for the different providers is the better choice and makes the behaviour more predictable.

OK, let's do that! We can make the default `0` which we can use for auto detection later if we want.

darthShadow:
Do we want to try introducing the v2 API only for listing (if that's even easy/possible) since it has native pagination support:

https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/

https://aws.github.io/aws-sdk-go-v2/docs/migrating/#paginators

ncw:
> Do we want to try introducing the v2 API only for listing (if that's even easy/possible) since it has native pagination support:
> 
> https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/
> 
> https://aws.github.io/aws-sdk-go-v2/docs/migrating/#paginators

I don't really want to include two AWS SDKs and I don't think this issue should be about switching SDKs so I think we should do it with the existing SDK and revist when we migrate to the new SDK.

ivandeex:
Friendly ping after 6 months of NOTHING

ivandeex:
Please note that after 6 months of no activity the PR becomes eligible for REMOVAL

ncw:
I've added a commit on top of this which falls back to ListObjectsV1 on unsupported providers.

If the tests pass, I'll merge this so it can get testing in the integration tests and exposure during the 1.58 cycle.

ncw:
@felixbuenemann 
> Thanks for taking the time to finish this up.

No probs! I was motivated to do this because I found a bug in Blackblaze's S3 interface which is fixed in the V2 list objects!

> Nice use of `structs.SetFrom`, I didn't know about that.

Thats an rclone special. A little bit of reflection really helps with the verbose S3 API structs!

https://github.com/rclone/rclone/blob/25366268fe921608ee68c6dc33b744b80d860abe/lib/structs/structs.go#L8-L19

Issue #5094, opened 2021-03-03
get filename with command "copyurl"


gandolfi974:
hello,

i open an issue for a option in rclone.

- i use rclone copyurl to download file from the web with -a flag (redirect url). I need to get the filename of the downloaded file in order to insert in a database.
rclone copyurl -P "https://video.com/stream/?vid=vmDZtFsMQ" gdrive:/film -a --transfers 10 --drive-chunk-size 128M

- I can't retrieve the filename. is it possible to add this option ?

thanks

albertony:
Suggested a fix in https://github.com/rclone/rclone/pull/5095

Note: Combining with option `-P, --progress` as in your example will not give good result, as that will clutter the output (and as noted in the forum thread `--transfers` is not relevant here).

PS: I'm still wondering what problem you had without the `-a` option, mentioned in the [forum](https://forum.rclone.org/t/get-filename-from-the-command-urlcopy-a/22613). Does not the following work?
```
rclone copyurl -P "https://video.com/stream/?vid=vmDZtFsMQ" gdrive:/film/testing123.mp4 --drive-chunk-size 128M
```

gandolfi974:
without -a or without indicate the final filename i can't copy. "command fail - "gdrive:/film/" is a directory".

it's ok with the command line `rclone copyurl -P "https://video.com/stream/?vid=vmDZtFsMQ" gdrive:/film/testing123.mp4 --drive-chunk-size 128M`

i have to create a name and don't retrieve the name

PR #5095, opened 2021-03-03
copyurl: add option to print resulting auto-filename

albertony:

Added new option:

> -p, --print-filename   Print the resulting name from --auto-filename

Example usage (PowerShell):
```
$x = .\rclone.exe copyurl https://www.youtube.com/watch?v=SzuZnh91U8E C:\Temp -a -p
PS > Write-Host "Downloaded file: ${x}"
Downloaded file: watch
```

https://github.com/rclone/rclone/issues/5094

#### Checklist

### DIFF
--- a/cmd/copyurl/copyurl.go
+++ b/cmd/copyurl/copyurl.go
@@ -3,6 +3,7 @@ package copyurl
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/rclone/rclone/cmd"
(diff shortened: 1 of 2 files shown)


cmd/copyurl/copyurl.go
```diff
@@ -13,15 +14,17 @@ import (
...
+	printFilename = false
+	stdout        = false
+	noClobber     = false
 )
 
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
 	cmdFlags := commandDefinition.Flags()
 	flags.BoolVarP(cmdFlags, &autoFilename, "auto-filename", "a", autoFilename, "Get the file name from the URL and use it for destination file path")
+	flags.BoolVarP(cmdFlags, &printFilename, "print-filename", "p", printFilename, "Print the resulting name from --auto-filename")
```
ncw, line 27: Just a note for the future - we have to be very careful with short flags in sub commands - this means we can't use `-p` as a global flag... In `rsync` it is

        -p, --perms                 preserve permissions

So probably not much use!

We also use `-p` in `rclone tree` also.

I think this use is OK though.

cmd/copyurl/copyurl.go
```diff
@@ -61,10 +65,14 @@ output to be written to standard output.
...
 		}
 		cmd.Run(true, true, command, func() error {
+			var dst fs.Object
 			if stdout {
 				err = operations.CopyURLToWriter(context.Background(), args[0], os.Stdout)
 			} else {
-				_, err = operations.CopyURL(context.Background(), fsdst, dstFileName, args[0], autoFilename, noClobber)
+				dst, err = operations.CopyURL(context.Background(), fsdst, dstFileName, args[0], autoFilename, noClobber)
+				if printFilename && err == nil && dst != nil {
+					fmt.Printf("%s", dst.Remote())
```
ncw, line 74: I think you want to print a newline here too - that is the unix way. So something like

    fmt.Println(dst.Remote())

ncw (commented):
Looks good! One small query inline.

cmd/copyurl/copyurl.go
albertony: Sure. I had it, amended it away, and now amended it back in! :)

cmd/copyurl/copyurl.go
albertony: Oh, I did not think of that, thanks for the reminder. I can change it if you think that is best, but I were unable to come up with better alternative so no change in latest update... 👎

PR #5113, opened 2021-03-11
docs: cleanup optional feature table

buengese:
The features table had quite a few mistakes this fixes them. We should really autogenerate this table.


ncw:
Thank you for doing this - looks much better :-)

I'll merge it now.

ncw (approved):

PR #5128, opened 2021-03-14
local: make nounc advanced option except on windows

albertony:

Avoid the following irrelevant question when configuring local remote on linux:
```
Disable UNC (long path names) conversion on Windows
 1 / Disables long file names
   \ "true"
nounc> 
```

Simple quick-fix: Just hiding it under the advanced config section when not on windows. Could have removed it from the list instead, but was thinking about generated docs etc..

No

#### Checklist


ncw (approved):
That looks great - thanks :-)

Issue #572, opened 2016-07-08
documentation is missing typical rclone.conf examples for each provider


zoobab:
Hi,

I am finding that the documentation talks about this interactive `rclone config` mode, but the most common way of configuring things is via a config file. It would be useful to provide typical config files for each provider, right now I am using rackspace with swift.

ncw:
The reason why I haven't documented the config file is that the majority of the remotes can't be configured that way as they use oauth2 which requires rclone interacting with a web server.

`rclone config` for swift has the rackspace endpoints built in which should make configuring rackspace swift really easy.

The config file for rackspace should look like this

```
[rackspace]
type = swift
user = USER
key = KEY
auth = https://identity.api.rackspacecloud.com/v2.0
```

edwardxml:
I will try to review this in the next few weeks. Ed.

edwardxml:
I had originally thought of going through creating example rclone.conf sections for a range of backends and pasting them into backend hep topics. That is just something else to maintain in step when in reality any rclone user wanting an example rclone.conf section will just create one with rclone config.  Backend specific flags and advanced ones are already documented directly from backend code.

I suggest this issue is closed.

ivandeex:
@albertony 

Do you think we need rclone.conf examples in backend docs?

albertony:
Hm... I agree with @edwardxml. But perhaps we could add a short description of the general syntax? I did not find that anywhere.. I mean something describing: Text file, where remote names are withing [], followed by "key = value" lines, where you at least must have key "type" with value according to the backend id as returned by `rclone help backends`, and optionally other options as described for each backend, etc...

ivandeex:
Or a link to https://en.wikipedia.org/wiki/INI_file#Format ?

albertony:
Yes, sure, that is fine. So just describe how "sections" are "remote names" etc..

jessuppi:
For future Googlers like myself, it seems the challenge is that Rclone supports simultaneous backup services (remotes) so the config file reflects that ability to have multiple sections in the INI format per-service.

Technical users who want to review everything first and don't really like using setup wizards may feel overwhelmed.

My project SlickStack is aiming to integrate Rclone using standardized config variables e.g.:

```
RCLONE_SERVICE="backblaze"
RCLONE_MODE="copy"
RCLONE_CLIENT_ID=""
RCLONE_CLIENT_SECRET=""
RCLONE_USER=""
RCLONE_PASSWORD=""
[...]

PR #5142, opened 2021-03-18
docs: add short description of configuration file format

albertony:

Add some documentation of the configuration file format, without going too much into details.

Seems rclone currently uses https://github.com/unknwon/goconfig library for handling the configuration file, but not mentioning this in the doc, as I guess it may change over time - and that one is actually documented as being sort of deprecated ("under bug fix only mode, which means no more features will be added").

I am linking to https://rclone.org/overview/ for list of backend "types", but it does not show the internal "codename" for the backends that must be put as value of the `type` key in config file. Command `rclone help backends` seems to show these type values. Perhaps the overview page should have listed them as well?

Added an example, used mega because I found that supported plain username/password auth, which makes it a simpler example than tokens etc..

#572

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -8,8 +8,8 @@ Configure
 
 First, you'll need to configure rclone.  As the object storage systems
 have quite complicated authentication these are kept in a config file.
-(See the `--config` entry for how to find the config file and choose
-its location.)
+(See the [`--config`](#config-config-file) entry for how to find the config
+file and choose its location.)
 
 The easiest way to make the config is to run rclone with the config
 option:
@@ -639,7 +639,7 @@ See `--copy-dest` and `--backup-dir`.
 
 ### --config=CONFIG_FILE ###
 
-Specify the location of the rclone config file.
+Specify the location of the rclone configuration file.
 
 Normally the config file is in your home directory as a file called
 `.config/rclone/rclone.conf` (or `.rclone.conf` if created with an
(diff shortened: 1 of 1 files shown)


ncw (approved):
This description looks great!

It is probably only missing how to make comments (with `#`)

At some point I would like to ditch `goconfig` - I recently refactored the config system to make it pluggable.

`goconfig` was the only ini style config system that I could find that could write the config file, preserving commands. There may be others now...

If you want to merge as-is that is fine, or add a section on comments if you want. Though not having comments in the config file would make re-writing goconfig a lot easier ;-)

albertony:
> It is probably only missing how to make comments (with `#`)

Yes, I did consider it... :innocent:

Both `;` and `#` can be used. Must be on separate lines, at least when I entered `type = mega # this is a comment` rclone reported error:

> Failed to find remote: didn't find backend called "mega # this is a comment"

When testing I see that comments at end of file, after last remote, are not preserved.

I'm a bit hesitant if it is worth mentioning, and if so if just small note or describe where it is working and not.. :thinking:

Could also add a note about file permissions, but there are considerations with encrypted vs unencrypted config, and there are differences between Windows and others that probably would make this more than just a short sentence as well, see https://github.com/rclone/rclone/issues/5168.

> add short description of configuration file format

Also, the commit message must be changed if adding these points! :smiley:

ncw:
> > It is probably only missing how to make comments (with `#`)
> 
> Yes, I did consider it... 
> 
> Both `;` and `#` can be used. Must be on separate lines, at least when I entered `type = mega # this is a comment` rclone reported error:

I didn't know that you could us `;` and I also didn't know that you can't commend at the end of a line, but it figures as it would mean you couldn't put `;` or `#` in a value.

> I'm a bit hesitant if it is worth mentioning, and if so if just small note or describe where it is working and not.. 

I do comment stuff out in my config file for testing.

Maybe a quick mention?

> Could also add a note about file permissions, but there are considerations with encrypted vs unencrypted config, and there are differences between Windows and others that probably would make this more than just a short sentence as well, see #5168.

He he!

Maybe "rclone tries to save config file so that only the current user can read it" or something similar which shows that rclone does something but is deliberately non-specific to leave us room to change the implementation.

> > add short description of configuration file format
> 
> Also, the commit message must be changed if adding these points!

;-)

albertony:
@ncw: Pushed a quick update where the two mentioned points are included. If you have time to give it a quick read-through, and just let me know if there any nitpicks, or else I will probably sneak-merge it at the end of the day...

PR #5151, opened 2021-03-21
Use build tags in the `version` and `selfupdate` commands

### LINKED ISSUE #5091: cmd/version: add info about build

1.54.0

(from https://forum.rclone.org/t/rclone-cmount-on-linux/22602/4)

Knowing whether it is a CGO enabled build or not (so whether it is static or dynamic) would be really useful for debugging DNS problems.

Knowing build tags would be useful too. For example, only `cmount` builds support passing custom `libfuse` flags. (See #3545 for an example why it matters).

`rclone version` output should look like below:
```
rclone v1.54.0
- os/arch: linux/amd64
- go version: go1.15.7
- executable: static (dynamic)
- tags: cmount (none)
```

Additionally, `rclone selfupdate` should refuse update if the running executable supports `cmount` but the update would remove it (and consequently break tooling wrappers).

Implementation details:
We'd have to pass the build tags into a variable in the Makefile (and probably a _built by CGO_ flag)

ivandeex:
Use build tags in the `version` and `selfupdate` commands

Knowing build tags would be useful.
For example, only `cmount` builds support passing custom libfuse flags.
(See #3545 for an example why it matters).

As stated in https://github.com/golang/go/issues/7007#issuecomment-66089610, golang does not and will not provide any runtime API for finding build tags on the fly:
> Build tags are for building. I do not want them to be exposed in the runtime.
> if you need to know ... you can write different files with different build tags and different initialized ... variables

This PR uses this very technique to initialize an internal array of build tags.
The technique has a limitation that every detectable tag must be known in advance.
Currently only the `cmount` and `cgo` tags are supported.

Knowing whether it is a CGO enabled build or not (so whether it is static or dynamic) would be really useful for debugging DNS problems.
As described in the go documentation (https://golang.org/cmd/cgo), when cgo builds a dynamic executable, it marks itself by the `cgo` build tag.

This PR makes `rclone version` show a list of predefined build tags or `none`.
The `cgo` tag will not be included in the list but used to determine whether the rclone binary is static or dynamic.

Additionally this PR makes `rclone selfupdate` aware of the `cmount` build tag to detect whether the FUSE support is in.
An update will be refused if the running executable supports FUSE semantics but the update would discard it and consequently break tooling wrappers.

Proposed in https://forum.rclone.org/t/rclone-cmount-on-linux/22602/4

Fixes #5091

#### Checklist

#### Testing

It is hard to test build capabilities automatically as CI/CD builds heavily depend on the build environment.
Instead, I will run the following manual test:

```
$ CGO_ENABLED=0 go build -o rclone-static . ; ./rclone-static version
rclone v1.55.0-DEV
- os/arch: linux/amd64
- go version: go1.16
- executable: static
- tags: none

$ CGO_ENABLED=1 go build -o rclone-dynamic . ; ./rclone-dynamic version
rclone v1.55.0-DEV
- os/arch: linux/amd64
- go version: go1.16
- executable: dynamic
- tags: none

$ xgo -image=billziss/xgo-cgofuse -targets=linux/amd64 -tags=cmount . ; ./rclone-linux-amd64 version
rclone v1.55.0-DEV
- os/arch: linux/amd64
- go version: go1.13.4
- executable: dynamic
- tags: cmount
```

#### Merging

I think this PR can be merged as two commits, where one patches the `version` command and the other patches `selfupdate`.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -36,6 +36,7 @@ import (
 	"github.com/rclone/rclone/fs/rc/rcflags"
 	"github.com/rclone/rclone/fs/rc/rcserver"
 	"github.com/rclone/rclone/lib/atexit"
+	"github.com/rclone/rclone/lib/buildinfo"
 	"github.com/rclone/rclone/lib/random"
 	"github.com/rclone/rclone/lib/terminal"
 	"github.com/spf13/cobra"
@@ -74,9 +75,13 @@ const (
 
 // ShowVersion prints the version to stdout
 func ShowVersion() {
+	linking, tagString := buildinfo.GetLinkingAndTags()
 	fmt.Printf("rclone %s\n", fs.Version)
-	fmt.Printf("- os/arch: %s/%s\n", runtime.GOOS, runtime.GOARCH)
-	fmt.Printf("- go version: %s\n", runtime.Version())
+	fmt.Printf("- os/type: %s\n", runtime.GOOS)
+	fmt.Printf("- os/arch: %s\n", runtime.GOARCH)
+	fmt.Printf("- go/version: %s\n", runtime.Version())
+	fmt.Printf("- go/linking: %s\n", linking)
+	fmt.Printf("- go/tags: %s\n", tagString)
 }
 
 // NewFsFile creates an Fs from a name but may point to a file.
--- a/cmd/cmount/arch.go
+++ b/cmd/cmount/arch.go
@@ -1,0 +1,7 @@
+package cmount
+
+// ProvidedBy returns true if the rclone build for the given OS
+// provides support for lib/cgo-fuse
+func ProvidedBy(osName string) bool {
+	return osName == "windows" || osName == "darwin"
+}
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -26,7 +26,7 @@ import (
 
 func init() {
 	name := "cmount"
-	cmountOnly := runtime.GOOS == "windows" || runtime.GOOS == "darwin"
+	cmountOnly := ProvidedBy(runtime.GOOS)
 	if cmountOnly {
 		name = "mount"
 	}
--- a/cmd/selfupdate/help.go
+++ b/cmd/selfupdate/help.go
@@ -37,6 +37,10 @@ your OS) to update these too. This command with the default |--package zip|
 will update only the rclone executable so the local manual may become
 inaccurate after it.
 
+The |rclone mount| command (https://rclone.org/commands/rclone_mount/) may
+or may not support extended FUSE options depending on the build and OS.
+|selfupdate| will refuse to update if the capability would be discarded.
+
 Note: Windows forbids deletion of a currently running executable so this
 command will rename the old executable to 'rclone.old.exe' upon success.
 
--- a/cmd/selfupdate/selfupdate.go
+++ b/cmd/selfupdate/selfupdate.go
@@ -21,9 +21,11 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/cmount"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/fshttp"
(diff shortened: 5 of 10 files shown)


albertony:
LGTM.

> An update will be refused if the running executable supports FUSE semantics but the update would discard it

Could have an option, e.g. `--force`, to override this. Just a thought, I have no idea how relevant it would be to use it...

ivandeex:
> Could have an option, e.g. `--force`, to override this. Just a thought, I have no idea how relevant it would be to use it...

I feel reluctant to add flags, as usual. Let's see what Nick says during review. Opposed by two, I will instantly yield :-)

cmd/cmount/arch.go
```diff
@@ -0,0 +1,7 @@
+package cmount
+
+// ProvidedBy returns true if the rclone build for the given
+// architecture provides support for lib/cgo-fuse
+func ProvidedBy(arch string) bool {
```
darthShadow, line 5: I would prefer this to be moved to the cmount file in buildinfo so that one command doesn't depend on a function from another (selfupdated depending on this function from cmount).

darthShadow:
I am also of the opinion to skip the `--force` flag for now. We can add it later if required. Best to keep the code as minimal/simple as possible.

Should we consider separating out the version info based on its type (`os` or `go`)? Something like this:

```
rclone v1.55.0-DEV
- os/arch: linux/amd64
- go/version: go1.13.4
- go/linking: dynamic
- go/tags: cmount
```

This does have the possibility of breaking any scripts that parse this output, however.

cmd/cmd.go
```diff
@@ -74,9 +76,25 @@ const (
 
 // ShowVersion prints the version to stdout
 func ShowVersion() {
+	exeType := "static"
```
darthShadow, line 79: We should also apply these changes to the `core/version` command from rc.

cmd/cmount/arch.go
ivandeex: The function `ProvidedBy` is logically closer to `cmount`. Writing `cmount.ProvidedBy(arch)` feels more natural than `buildinfo.ProvidedBy(arch)`.
Though it lives in the `cmount` package, it is independent from the `cmount` command. The command may not be compiled in due to build conditions but the function will still be there.
Probably we could move it to `cmd/mountlib` and call as `mountlib.CmountProvidedBy`, but I am slightly biased to `cmount.ProvidedBy`.

ivandeex:
> ```
> - go/version: go1.13.4
> ```
> 
> This does have the possibility of breaking any scripts that parse this output, however.

I feel like it's not going to wreak havoc. The `go version` line is rather internal to the rclone support team.

cmd/cmount/arch.go
ivandeex: Moreover, the `lib/buildinfo/cmount.go` is not always present.

ivandeex:
Added a commit for `rc:core/version`. It can be squashed with the `version:` commit while the `selfupdate:` one can stay independent.

ivandeex:
Added a commit for rc:core/version. It can be squashed with the version: commit while the selfupdate: one can stay independent.

ivandeex:
Added a commit for `rc:core/version`. It can be squashed with the `version:` commit while the `selfupdate:` one can stay independent.

cmd/cmount/arch.go
darthShadow: Missed that. The current implementation is fine then.

ivandeex:
Added a commit for `rc:core/version`. It can be squashed with the `version:` commit while the `selfupdate:` one can stay independent.

darthShadow:
One minor consideration: Do we want to split `os/arch` further to `os/type` & `os/arch` for the values of `linux` & `amd64` respectively?

ivandeex:
> One minor consideration: Do we want to split `os/arch` further to `os/type` & `os/arch` for the values of `linux` & `amd64` respectively?

What would Nick say?

albertony:
> One minor consideration: Do we want to split `os/arch` further to `os/type` & `os/arch` for the values of `linux` & `amd64` respectively?

I'm thinking the same..

ncw (approved):
Both those commits look great - thanks.

Please merge when ready.

Issue #5071, opened 2021-02-25
cmd/hash: add simple aliases for rclone hash names


ivandeex:
#### What is your current rclone version (output from `rclone version`)?

1.54.0

#### What problem are you are trying to solve?

I cannot easily remember rclone hash names. Compare: `MD5`, `SHA-1`, `MailruHash`, etc. I would like to write them simply as `md5`, `sha1`, `mailru` and so on. For example `rclone hash sha1 remote:file`.

#### How do you think rclone should be changed to solve that?

Introduce canonical hash name in rclone and a function that modifies a provided name:
- remove dashes
- convert to lowercase
- remove trailing `hash`, if any

Canonicalization should happen when
- a hash algorithm is internally registered
- one of hash commands is parsed
- in `rc` protocol handler

Notes:
- what name should appear in documentation? needs thinking
- there are official hash names out there that contain a dash like `sha256-something`, but none of them are used in rclone a.t.m.

ivandeex:
related to #4841

PR #5164, opened 2021-03-23
rclone bisync

### LINKED ISSUE #118: Two-way (bidirectional) synchronization
I'm sorry if this is answered elsewhere but I couldn't find it in that case.

I want to replace my current `Owncloud+Owncloud client (Linux)+FolderSync(Android)` setup with `Drive+Rclone+FolderSync`. But there is one thing I can't figure out how to do with rclone — smart two-way deletion synchronization. Which means: if a file was present on both server (Drive) and local machine, and then was deleted on either of them, the file will be eventually removed on both regardless of which direction you run `sync` first. Same, if a file was added on either server or client, it will be uploaded to the other one.

Can rclone do that, and if doesn't is there a chance of such functionality in future?

#### How to use GitHub

ivandeex:

Implement the `rclone bisync` command (port of [rclonesync](https://github.com/cjnaz/rclonesync-V2))

Fixes #118

#### Checklist

- [x] I'm done, this Pull Request is ready for review :-Q

### Action Items

- [x] Core functionality, first sync and dry run modes
- [x] Access health check
- [x] Filters file MD5 validation
- [x] Test engine for original test scenarios
- [x] Honor `--max-delete` flag
- [x] API endpoint `sync/bisync`
- [x] Basic test (`test_basic`) passes
- [x] Draft documentation
- [x] Really compare console logs
- [x] Make first-sync and dry-run tests pass
- [x] Implement ABORT/CRITICAL logic
- [x] Compatibility with `--legacy` messages and tests
- [x] Port README.md and TESTING.md
- [x] All tests pass on Local FS
- [x] All tests pass with Google Drive
- [x] All tests pass with Dropbox
- [x] Tests run and pass on Windows
- [x] Fix tests for Mailru
- [x] Fix tests for FTP (depends on #5283)
- [x] Fix tests for SFTP
- [x] Fix tests for AWS S3
- [x] Fix tests for Chunker/LocalFs (depends on #5297)
- [x] Put verbosity of bisync messages under control
- [x] Develop legacy free syntax for test scenarios
- [x] Keep hash in listings, if available (don't use for sync yet). See [rclonesync#5](https://github.com/cjnaz/rclonesync-V2/issues/5)
- [x] Parallelize all list/copy/delete operations for performance
- [x] Drop legacy support as soon as we are fully decoupled from the original project
- [x] Gracefully handle invalid UTF-8 in filenames. ([rcsync#62](https://github.com/cjnaz/rclonesync-V2/issues/62))
- [x] Handle files that change during a bisync run - _documented_ ([rcsync#24](https://github.com/cjnaz/rclonesync-V2/issues/24))
- [x] Fix remaining FIXMEs
- [x] Add original license, author and attributions
- [x] More docs: FILTERING.md, USAGE.md, TROUBLESHOOTING.md
- [x] Inventory bug reports from the original project


ivandeex:
## Optional items (will be postponed to later PRs)

- [ ] Use _trie_ in-memory format for listings to reduce footprint. See [rclonesync#59](https://github.com/cjnaz/rclonesync-V2/issues/59)
- [ ] Support gzipped listings to reduce workdir size
- [ ] Parallelize move/rename operations
- [ ] Add option: number of listings to keep
- [ ] Honor `--interactive` and `--auto-confirm`
- [ ] Handle deletion conflicts. See https://github.com/rclone/rclone/issues/118#issuecomment-812205745
- [ ] Honor `--checksum` or `--ignore-checksum` to control hashsum use
- [ ] Handle hashsums, respect `--track-renames`
- [ ] Detect moved/renamed directories
- [ ] Honor special `.rcignore` files
- [ ] Delete into `--backup-dir`
- [ ] Revisit bisync progress display
- [ ] Handle case-insensitive filenames. Respect `--ignore-case` and `--ignore-case-sync`. Also see https://github.com/rclone/rclone/issues/4854
- [ ] Synchronize empty dirs if supported by backend
- [ ] Handle Google documents

ncw:
Is this ready for review?

Looks very exciting :-)

ivandeex:
After 3 months we have a working bisync beta that is feature-wise on par with [rclonesync](https://github.com/cjnaz/rclonesync-V2). It still should go through the merge review (up to 1-2 months).
If you are impatient, you are free to try [current beta](https://github.com/ivandeex/rclone/releases/tag/v1.55.1-bisync01) at your own risk.
Note that a number of [important features](https://github.com/rclone/rclone/pull/5164#issuecomment-839764078) has not been included due to time constraints. Please refrain from reporting them here or in #118. I will work on adding them as well as on extending backend support, after 1.57.

ivandeex:
> Is this ready for review?

Yes please :)

cmd/bisync/cmd.go
```diff
@@ -0,0 +1,224 @@
...
+		isDropbox1 := strings.HasPrefix(fs1.String(), "Dropbox")
+		isDropbox2 := strings.HasPrefix(fs2.String(), "Dropbox")
+		if commonHashes == hash.Set(0) && (isDropbox1 || isDropbox2) {
+			ci := fs.GetConfig(ctx)
+			if !ci.DryRun && !ci.RefreshTimes {
+				fs.Debugf(nil, "Using flag --refresh-times is recommended")
+			}
+		}
+
+		// FIXME Why runtime errors make rclone print usage?
```
ivandeex, line 143: @ncw 
Why runtime errors make cobra print usage?
How can I disable that?

cmd/bisync/cmd.go
```diff
@@ -0,0 +1,224 @@
...
+			os.Exit(2)
+		}
+		return err
+	},
+}
+
+func (opt *Options) applyContext(ctx context.Context) {
+	maxDelete := DefaultMaxDelete
+	ci := fs.GetConfig(ctx)
+	if ci.MaxDelete >= 0 {
```
ivandeex, line 157: @ncw 
I made bisync use the generic `--max-delete` flag but as a percentage in range `0..100` rather than absolute number of deleted objects.
Maybe it's better to rename it as `--max-delete-percent` or `--bisync-max-delete`?

cmd/bisync/rc.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if dryRun, err := in.GetBool("dryRun"); err == nil {
+		ci.DryRun = dryRun
+	} else if rc.NotErrParamNotFound(err) {
+		return nil, err
+	}
+
+	if maxDelete, err := in.GetInt64("maxDelete"); err == nil {
+		if maxDelete < 0 || maxDelete > 100 {
+			// TODO Cannot return rc.ErrParamInvalid as its field is private
```
ivandeex, line 35: @ncw 
I would like to return a `ErrParamInvalid` error here but it won't let me set a required field.
May we change its definition and make the field public?

cmd/bisync/bisync_test.go
```diff
@@ -0,0 +1,1249 @@
...
+// Command line flags for bisync test
+var (
+	argTestCase  = flag.String("case", "", "Bisync test case to run")
+	argRemote2   = flag.String("remote2", "", "Path2 for bisync tests")
+	argNoCompare = flag.Bool("no-compare", false, "Do not compare test results with golden")
+	argNoCleanup = flag.Bool("no-cleanup", false, "Keep test files")
+	argGolden    = flag.Bool("golden", false, "Store results as golden")
+	argDebug     = flag.Bool("debug", false, "Print debug messages")
+	argStopAt    = flag.Int("stop-at", 0, "Stop after given test step")
+	// Flag -refresh-times helps with Dropbox tests failing with message
```
ivandeex, line 124: @ncw 

Dropbox needs `RefreshTimes` for tests to pass. Probably other backends will need more specific flags.
I don't know how to solve it cleanly.

cmd/bisync/bisync_test.go
```diff
@@ -0,0 +1,1249 @@
...
+			log.Printf("  File found in Results but not in Golden:  %s", file)
+		}
+	}
+
+	for _, file := range goldenFiles {
+		if !resultSet.Has(file) {
+			continue
+		}
+
+		goldenText := b.mangleResult(b.goldenDir, file, false)
```
ivandeex, line 898: @ncw 

Bisync tests are slightly unusual in relation with rclone test suite.
Instead of creating a test situation programmatically and checking selected result properties
the are based on a simplistic but flexible test scenarios run by a test engine and compare text results.
Therefore, time stamps of test files are predefined. The suite will reject backends without mod-time support.
Regular expressions are used to further make results comparable.
However, I like the principle and thinks it's worth adopting.

cmd/bisync/bilib/hash.go
```diff
@@ -0,0 +1,57 @@
+// Package bilib provides common stuff for bisync and bisync_test
+// Functions here provide simple aliases for rclone hash names.
+// TODO will be later implemented in fs/hash.go
+// SEE https://github.com/rclone/rclone/issues/5071
```
ivandeex, line 4: @ncw 

This whole kludge will have gone after I submit a PR for #5071

cmd/bisync/help.go
```diff
@@ -0,0 +1,198 @@
...
+
+Most of these events come up due to a error status from an internal call.
+On such a critical error the |{...}.path1.lst| and |{...}.path2.lst|
+listing files are renamed to extension |.lst-err|, which blocks any future
+bisync runs (since the normal |.lst| files are not found).
+
+Some errors are considered temporary and re-running the bisync is not blocked.
+The _critical return_ blocks further bisync runs.
+
+### Command line syntax
```
ivandeex, line 97: @ncw 

`make commanddocs` will eat up one leading pound from every heading here so I had to shell out some. Could you reimburse me? :)

cmd/bisync/bisync_test.go
```diff
@@ -0,0 +1,1249 @@
+// TestBisync is a test engine for bisync test cases.
+// See TESTING.md for documentation.
```
ivandeex, line 2: This should be https://rclone.org/bisync/#testing

Jwink3101:
This looks really impressive. And I have a few comments but it is worth noting that I am *not* impartial since I wrote three (though only one that really matters, [syncrclone](https://github.com/Jwink3101/syncrclone)) tools this "sherlocks" and those tools were not the basis for this. But I have given these things a ton of thought both in theory and in practice (though in Python, not golang)

And this is clearly the result of lots of very hard work and lots of thought. No intention to downplay that.

Also, I only read the proposed documentation and didn't dive into the code.

Finally, a lot of my comments are based on **my perception** of the goals and ideals of rclone. They may diverge from @ncw's true view.

**TL/DR**: This is impressive work but I really think it would have been better to *not* base this approach on an existing tool (mine or otherwise) and instead design the algorithm to be "rclonic" (I coined a term similar to "pythonic")

**TL/DR 2**: There is a lot of philosophical and opinionated questions here. While this is super impressive, I *personally* do not agree with them for rclone. But I 100% recognize that it isn't up to me.

## Limited remote support (i.e. needs ModTime)

One of the many features of rclone is that it can handle nearly all  remotes the same with minor details. The fact that this only supports a small subset of them is disappointing. Especially with some popular ones not supporting mod time. And others, like S3, are way more performant if you disable it.

This is an issue with the overall algorithm of some sync tools but there is no reason you *need* modtime to decide what to sync. You may need it for *conflicts* but if you have past state, then you can tell if a changed file-size is a modification on one side only.

## Compare Methods and ModTime Precision.

In the docs, it says:

> Note that on some cloud storage systems it is not possible to have file
> timestamps that match _precisely_ between the local and other filesystems.
> Bisync's approach to this problem is by tracking the Path1-to-Path1 and
> Path2-to-Path2 deltas on each side _separately_, and then applying
> the resulting changes on the other side.

Again, I didn't read all of the code but stepping back a bit, there are two fundamental ways to do bi-sync:

1. Compare each side independently to past state to develop a changelog. Propagate that change log
2. Compare current state on each side using past state *only* for resolution (i.e. is this new on one or deleted on the other? Is it modified on one? Both?)
    - Additional use for tracking renames though this isn't needed if you don't mind rclone's current approach requiring shared hashes or ModTIme

Based on your comment, you are doing the first. This has the advantage of being robust to time differences but also breaks with rclone's current behavior of `--modify-window` and the more general behavior of looking at current state.

With approach 2, you can handle ModTime's not agreeing **and/or size and/or hash** like regular rclone. 

The sub-bullet on the second is also interesting as it *does* allow you to track renames even if hashes aren't compatible. But you could also track with rclone's current behavior of looking at the extra files. Then use the past state to decide which side did the moving.

Approach 1 also requires the whole idea of `--resync` while it is implicit if you don't have past state in approach 2.

## Lock Files

rclone right now also has the potential for long running jobs and there isn't a facility for lock files. I don't see why we need one for this *as part of rclone*. [My tool](https://github.com/Jwink3101/syncrclone), for example, has the option but it is a separate entity from rclone.

Same idea of "something evil". Regular rclone doesn't have this either.

Personally, I'd *much* rather see `--backup-dir` support out of the gate with that being the check. And again, it matches current rclone behavior.

## State Storage

Unlike mirroring or one-way sync (a la `sync`), true bidirectional sync cannot be stateless. However, I think where to store the state warrants more discussion. One thing often discussed on the forum and rclone subreddit is that you can pull your config to a VPS and run it there. If you store the state of past syncs on the syncing machine, you now need to keep track of that.

This is actually, in my opinion, one of the fundamental issues with rclone and bi-directional sync. You need a robust way to handle this and you can't just go into a deeper dir. (for example `rclone sync A: B:` works as does `rclone sync A:sub B:sub`). This is the primary reason I chose to for my tool to not be a pure CLI one and instead of a sync config file that dictated these things. 

## Filters

Why are the filters different? The only one of concern is `--exclude-if-present` which can cause issues with bi-directional sync. The following is the note I include in my readme about them:

> Filters work well since they are applied to both sides. This means that changing a filter will make the files look like it was deleted on *both* sides and nothing will happen.
> 
> However, the [`--exclude-if-present`][eifp] filter is **very dangerous** if the excluded file is added on one side only *after* files in the directory are in sync. It will cause rclone to skip the directory on one side and appear like it was deleted. 
> 
> There *are* safe ways to use `--exclude-if-present`. For example, you can place exclude file in place on *both* sides before syncing. (Or placing it without adding the filter, syncing, and then adding the filter). Or, if the files have *never* been synced before (i.e. it's a new directory with the exclude file) then nothing bad will happen.


This isn't a new problem in bisync from regular rclone though I do see it amplified for it. Again this goes back to the philosophical idea of a sync-config vs CLI flags. The reason I chose not to use the CLI flags was because I wanted it very clear that a "sync job" is not a one-off.

## "In the face of uncertainty..."

Just a wild thought, ncw has said before that when uncertain how to handle a situation, look at rsync. Is it worth the same idea with Unison? Not that I use Unison either and is the reason I wrote my [first sync tool](https://github.com/Jwink3101/PBrsync) followed by [my second](https://github.com/Jwink3101/PyFiSync) (and then followed by [my third](https://github.com/Jwink3101/syncrclone) to better work with rclone)

---

I hope this was helpful and please (**please**!) do not take it as undervaluing or me downplaying the hard work! 

I just think it is worth a larger discussion (and/or maybe, to borrow from the Python community, a BDFL decision by @ncw) on the approach forward. There is something to be said for the fact that while I clearly have an opinion on this, I didn't do the legwork to add it to rclone! (especially since I don't know golang). So take my comments with that huge grain of salt!

Thanks!

ivandeex:
@Jwink3101 

Your post calls for a bigger discussion which needs more time. I don't have answers now for all the topics raised. Let me just add a few quick thoughts.

>  I really think it would have been better to _not_ base this approach on an existing tool (mine or otherwise) and instead design the algorithm to be [...]

My decision to adopt _rclonesync_ as a basis for rework was dictated by good coverage of its test suite. Having tests passing let me continuously refactor the code and direct it wherever wanted with confidence (now or in future) - e.g. port to another language, add hashsum support, optimize bottlenecks or change the change detection algorithm.

> ## Limited remote support (i.e. needs ModTime)

I have left hashsum support for later patches. However I reserved fields in state for it to ensure state format compatibility.

> This is an issue with the overall algorithm of some sync tools but there is no reason you _need_ modtime to decide what to sync.

Theoretically no. The bisync change detection is all about [bit masks](https://github.com/rclone/rclone/pull/5164/files#diff-25f7d45d3740899fdbbb93ff87a682943b7085b9c2f7f67b9cc58756c90e3630R31)

But it needs more testing.

> Again, I didn't read all of the code but stepping back a bit, there are two fundamental ways to do bi-sync:

> With approach 2, you can handle ModTime's not agreeing **and/or size and/or hash** like regular rclone.

Interesting. I'm not ready to discuss aporoaches. It needs time to get literate in that. Feels like switching to 2nd aporoach would mean a complete overhaul. I'm not sure.

> rclone right now also has the potential for long running jobs and there isn't a facility for lock files. I don't see why we need one for this _as part of rclone_. 

That's discussible. I will just note that bisync locks are pretty fine grained (per src/dst pair).

> Same idea of "something evil". Regular rclone doesn't have this either.

Saving from disaster is not a bad idea. Bisync has --force for unattended operations.

> However, I think where to store the state warrants more discussion.

Yes

> This is the primary reason I chose  for my tool to not be a pure CLI one and instead of a sync config file that dictated these things.

I thought in different direction - to snapshot the current state of flags into bisync session (extending the idea of hashsumming the filters) and request a resync or at least warn if it changes. I'll add this on my book.

> I just think it is worth a larger discussion and a BDFL decision by @ncw on the approach forward. 
> Thanks!

Thank you for comments and ideas.

cmd/bisync/deltas.go
```diff
@@ -0,0 +1,310 @@
...
+
+	for _, file := range old.list {
+		d := deltaZero
+		if !now.has(file) {
+			b.indent(msg, file, "File was deleted")
+			ds.deleted++
+			d |= deltaDeleted
+		} else {
+			if old.getTime(file) != now.getTime(file) {
+				if old.beforeOther(now, file) {
```
ivandeex, line 132: @ncw

This approach is simplistic based on fact that we always compare a filesystem with itself. Compare this e.g. with [equal()](https://github.com/rclone/rclone/blob/v1.55.1/fs/operations/operations.go#L167) or even with proposal #4810.

ivandeex: However, @Jwink3101 is right. (In)equality and before/after relations in this piece should be rethought with backend's time window in mind.

cmd/bisync/bisync_test.go
ivandeex: Probably, mod-time discrepancies with golden results can be ignored under cirtcumstances, like it's done for hashsums, but it needs more testing.

cmd/bisync/deltas.go
ivandeex: Something like [fstest.CheckTimeEqualWithPrecision] (https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L116) for equaliy but extended to the `>` and `<` relations. I'd put it in separate functions and add unit tests that ensure it's [anticommutative](https://en.wikipedia.org/wiki/Anticommutative_property) etc.

cmd/bisync/bisync_test.go
ivandeex: `CompareResults` is a very far relative of [fstest.CheckListingWithRoot](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L265) and it bases on `mangleListing` down the code.

cmd/bisync/bisync_test.go
```diff
@@ -0,0 +1,1249 @@
...
+func mangleListing(text string, golden bool) string {
+	lines := strings.Split(text, eol)
+
+	hasHeader := len(lines) > 0 && strings.HasPrefix(lines[0], bisync.ListingHeader)
+	if hasHeader {
+		lines = lines[1:]
+	}
+
+	// Split line in 4 groups:    (flag, size)(hash.)( .id., .......modtime....... )(name).
+	regex := regexp.MustCompile(`^([^ ] +\d+ )([^ ]+)( [^ ]+ [\d-]+T[\d:.]+[\d+-]+ )(".+")$`)
```
ivandeex, line 1099: @ncw 

`mangleListing` is a very far relative of [fstest.CompareItems](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L356). It already has ability to ignore hashsums on demand. It still needs ability to compare times with per-backend tolerance or even ignore times if unsupported. Having that would let us run tests against backend without time support and probably lift the corresponding ban.

ivandeex: Tentatively this can be implemented as:
1. parse out the time filed and convert to `time.Time`
2. round time down to _doubled_ tolerance
3. put time field back into string
4. feed the result to `diff`.
I hope it does not sound scary, but it would be part of the proposed approach to testing.

ivandeex: Why not just parse and compare with tolerance?
Because `mangleResult` is two-fold. Besides comparing it can prepare a test state snapshot for storing in golden directory, which later can provide a bigger picture of what's going on.

cmd/bisync/testdata/test_all_changed/golden/_testdir_path1.._testdir_path2.copy1to2.que
```diff
@@ -0,0 +1,8 @@
+"RCLONE_TEST"
```
ivandeex, line 1: @ncw 

Double quotes are not actually needed for most listings.
Probably we could use the more discriminating `bisync.escapePath()` here.

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+means that they should match when run with a different cloud service.
+
+Your normal workflow might be as follows:
+1. Git-clone the rclone sources locally
+2. Modify bisync source and check that it builds
+3. Run the whole test suite `go test ./cmd/bisync -remote local`
+4. If some tests show log difference, recheck them individually, e.g.:
+   `go test ./cmd/bisync -remote local -case basic`
+5. If you are convinced with the difference, goldenize all tests at once:
+   `go test ./cmd/bisync -remote local -golden`
```
ivandeex, line 961: @ncw @cjnaz

The `rclonetest -golden` was **the thing** that bought me.
It polished it a little with _anti-hoppers_ (a far relative of [sorting in fstest.CompareItems](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L391)) and [replacers](https://github.com/rclone/rclone/pull/5164/files#diff-ee6acf898897fdf3c0a0b890fc8feffdcac69b41323b24da034da40af46df01bR1145) and gave me a mechanism to snapshot a **bigger picture**.
Git's colorful word diff then gives a tool to lead the **continuous refactoring** with all noise filtered away.
I think this approach can have its chance make into the rclone test suite alongside with [fstest framework](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go) ~~and live happily together ever after~~.

ivandeex:
@Jwink3101 

Today I've got more references to further address concerns raised by you (reordered).

## 1. ModTime

> Limited remote support (i.e. needs ModTime) [...] The fact that this only supports a small subset of them is disappointing. Especially with some popular ones not supporting mod time. And others, like S3, are way more performant if you disable it.
> This is an issue with the overall algorithm of some sync tools but there is no reason you _need_ modtime to decide what to sync. You may need it for _conflicts_ but if you have past state, then you can tell if a changed file-size is a modification on one side only.

With a feature _must_ come a method of testing it.

Rclone has a comprehensive `fstest` framework dedicated to peculiarities of the plethora of supported cloud providers. Planning to work on this patch I had the heart to keep the project test engine's nature (while translating it from python to golang and refactoring it to work natively and unattended on Windows and with a wider set of backends) and propose it as a new addition to the set of rclone approaches (see my inline comment above), instead of rewriting it from the ground up for `fstest`.

The bisync test suite **lacks** the ability to compare imprecise/absent timestamps as of the time of this writing (yes I said it, _absent_ timestamps should be "comparable" too). It's still correct now only because it compares logs from a particular filesystem **with itself**. Unless this ward is fixed, we can't run tests against backends without modtime and consequently can't declare them supported. I have an idea on how to fix that (see another inline comment), but the effort will need time.

## 2. --exclude-if-present

> Why are the filters different? The only one of concern is `--exclude-if-present` which can cause issues with bi-directional sync.
> > Filters work well since they are applied to both sides. This means that changing a filter will make the files look like it was deleted on _both_ sides and nothing will happen.
> > However, the [`--exclude-if-present`] filter is **very dangerous** if the excluded file is added on one side only _after_ files in the directory are in sync. It will cause rclone to skip the directory on one side and appear like it was deleted.

Nice catch! We have open bug reports (#4523) about this. We have requests (#4753, #4793) to add more _non-commutive_ flags.  We try (#4011, #4856) to restrain it reasonably. I'd probably make bisync ban this flag or issue a warning or put it under the `--force` conditional (?).

## 3. Not so rclonic

> I really think it would have been better to [...] design the algorithm to be "rclonic"

The heart of rclone's omnivory change detection runtime lies around [operations.equal](https://github.com/rclone/rclone/blob/v1.55.1/fs/operations/operations.go#L167), [operations.checkHashes](https://github.com/rclone/rclone/blob/v1.55.1/fs/operations/operations.go#L67), et cetera. It's kept afloat by the astute test framework, in particular [fstest.CheckListing](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L264), [fstest.CheckHashes](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L137), [fstest.CheckTime](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L116) and (drums....) [retrying](https://github.com/rclone/rclone/blob/v1.55.1/fstest/test_all/run.go#L400).

This code is not big in terms of LOC but its [real complexity](https://github.com/rclone/rclone/issues/4793#issuecomment-733647201) is high, while users continue to request for more [change detection](https://github.com/rclone/rclone/issues?q=is%3Aopen+is%3Aissue+label%3A%22change+detection%22) and [filtering](https://github.com/rclone/rclone/issues?q=is%3Aopen+is%3Aissue+label%3Afiltering) features. My first approach to bisync about three months ago came across problems I can't clearly remember now so I tried it differently.

There is an ongoing proposal (#4810) that seeks to put it under control. Ultimately we need to inventory the state of affairs if we want to add more features safely. I think we could continue this discussion in #4810. Maybe your experience in the synchronization field could help here? 

## 4. Standing _on_ the shoulders of giants watching their heights from the _ground_

<p align="right">
(<i>Probably rephrased from Mark Twain?</i>)
</div>

> when uncertain how to handle a situation, look at rsync. Is it worth the same idea with Unison?

Unfortunately rclone has already diverged (#1228, #3375, #5040, #5043 [docs](https://rclone.org/filtering/#how-filter-rules-are-applied-to-directories)) from rsync :((
We can't easily fix that due to API stability concerns so there exists a waiting proposal (#4776) of `rclone rsync` besides `rclone sync`.

Mimicking Unison is a sound idea. However, if it would result in complete algorithm overhaul (not a critical problem) or if unison algorithms don't extend easily to the cloud world (that's worse), then I'd **maybe** think of developing an independent `rclone unison` (similar to `rclone rsync` besides `sync` above).

> I hope this was helpful

ABSOLUTELY. THANK YOU.

jtagcat:
I'd make a recommendation, following https://github.com/rclone/rclone/issues/118#issuecomment-842263888, that a beta label, a warning, or similar be added to shorthelp for the time being.

Jwink3101:
@ivandeex, Thanks for the further discussions.

I really do not know the internals of rclone very well. I only know the behavior based on testing and documentation. So I am not sure what is possible. But it seems to me like an added technical debt to have bisync do its own comparison routines as opposed to being automatically up to date.

One thing I toyed with when I first tried to make a two-way sync around rsync was to do two `--dry-run` calls `A-->B` and then `B-->A`. I then processed the actions and made my own decisions based on past state.

Could you do something like that but intrusive as opposed to a wrapper? The major downside of doing it from outside is two file listings but assuming the rclone structure allows, you can repeat it with the same listing. Then the initial compare implicitly follows the rclone logic as it stands today and in the future. And you end up with a lot fewer objects inside of your process since you've removed the ones that are the same.

This, BTW, also then aligns with what I called "method 2" for determining changes. (that is, being based *first* on current state rather than propagating changes. Both work but I still believe to be way more robust in the real world).

Thanks again for your hard work on this. I wish I knew golang and had the time to help. I will read #4810 and see if I can be of help

cjnaz:
A few comments on alternative algorithms (assuming I'm fully following the discussion)

1) Detecting on size change only leaves out the case where a file is changed but is the same size.  This could happen when a process writes a time string to a file, for example.  Other than this case, in practice changed-but-same-size is relatively rare; however, I'd be submitting a bug report if such a file was missed in the sync.

2) (Perhaps the rsync mode is a different algorithm, but assuming it applies to bisync...)   Regarding `
do two --dry-run calls A-->B and then B-->A`, I think the change detection must be based on a prior state/history file/db/etc.  This is needed in order to detect deletes.  For rclonesync I considered making a single LSL file that represents both paths.  This would have required generisizing this listing file format a bit and supporting mod-time tolerance.  The value is limited - cuts memory usage, and designs out the possibility of divergence since both paths are compared to the same master history list.  In rclonesync V3.2 I added a final compare of the final LSL files since the [divergence bug](https://github.com/cjnaz/rclonesync-V2/issues/70) did come up - actually triggered by an rclone bug in V1.52.  Memory footprint can be a big deal.  One rclonesync user had abut 2M files and ran out of memory.  I optimized rclonesync to get it down to two in-memory file listings at any time.  I understand that rclone deals with the file system recursively by directory rather than the whole tree?

3) Hashing the user's filter file in rclonesync addressed a bug where many files were deleted because they did not show up in the current state list and were in the prior history list, so they needed to get deleted on the other side.  The rclonesync algorithm evolved, so the specific fail case may not be there now.  In any case, damage control must be a prime design requirement.  `--backup-dir` sounds appropriate.  Are CLI filter options accepted by bisync?

4) One user [offered up](https://github.com/cjnaz/rclonesync-V2/issues/68) an inotify change notification technique, along with a Dropbox change notification solution for triggering an rclonesync.  If we could get change notifications from all the remotes then perhaps bisync itself is not needed.  :>

I'm very pleased to hear that the test framework is of real value.  :>

cmd/bisync/cmd.go
```diff
@@ -0,0 +1,224 @@
...
+	Workdir         string
+	DryRun          bool
+	NoCleanup       bool
+	SaveQueues      bool // save extra debugging files (test only flag)
+}
+
+// Default values
+const (
+	DefaultMaxDelete     int    = 50
+	DefaultCheckFilename string = "RCLONE_TEST"
```
ivandeex, line 47: @ncw 

Probably `.rclonetest` would be a better name to be in line with `.rclonelink` (https://rclone.org/flags)

Jwink3101:
@cjnaz 

> 1. Detecting on size change only leaves out the case where a file is changed but is the same size.  This could happen when a process writes a time string to a file, for example.  Other than this case, in practice changed-but-same-size is relatively rare; however, I'd be submitting a bug report if such a file was missed in the sync.

Yep. I agree. It is a bad method but it is the same one, to my knowledge, used my rclone already for remotes that do not provide ModTime or a hash

> I think the change detection must be based on a prior state/history file/db/etc.

Yes and no. I am not saying you don't need deletes. You do (as you note). But you do not need to detect changes and then propagate. That is what I referred to as method 1. The alternative (and I think much better and more robust) is to FIRST compare the **current** state and then use the previous state to detect whether it was a delete or a new file

> Hashing the user's filter file in rclonesync addressed a bug where many files were deleted because they did not show up in the current state list and were in the prior history list, so they needed to get deleted on the other side

Again, I am not disagreeing that it is an okay design for a *wrapper tool*. It is just that it **diverges from rclone's current behavior**.  Consider:

    $ rclone sync remote1:interesting-dir remote2:

and then

    $ rclone sync remote1:empty-dir remote2:

poof. There goes your files.

I am all for guardrails. But lets also not diverge from the way things work now.

> inotify

I agree. That is not why rclone (or your tool or my tool) is designed to do. But a user could write something that uses inotify to call it. (though it is worth noting that tools like Dropbox are more efficient than either of these methods since they do not call for the full listing but require a server process too)

> [re changes]

I agree with these issues of sorts. This isn't new in rclone either. If you change the filesystem while it is running, things can break. For my tool, I [enumated what would happen](https://github.com/Jwink3101/syncrclone/blob/master/docs/misc.md#interruptions) if it is interrupted. Something similar happens if a file changes.

But again, this is the beauty of the not-using-history-to-make-a-changelog approach (\#2). You only need history to figure out what to do!. So if a file was  added between the lsl, it will still copy.

Jwink3101:
Just to add 

> Memory footprint can be a big deal.
> ...
> I understand that rclone deals with the file system recursively by directory rather than the whole tree?

I don't know but this is a good point. I still think my "dry-run-first" approach makes a ton of sense since you implicitly rely on the current logic but it means it needs to be *more* intrusive than i first though guessed since you would need to work inside the loop.

As a side note, I ran into the same problem. I keep them in memory in a pretty inefficient object (one I created: [dicttable](https://github.com/Jwink3101/syncrclone/blob/master/syncrclone/dicttable.py)) but I have been thinking about how to improve that if needed. The best option at the moment is to write to SQLite and move some compare and query logic to SQL. (I would lose O(1) for O(logn) on queries with B-tree tables but that is acceptable).

ivandeex:
@cjnaz

> 3. Hashing the user's filter file in rclonesync addressed a bug where many files were deleted because they did not show up in the current state list and were in the prior history list, so they needed to get deleted on the other side.  The rclonesync algorithm evolved, so the specific fail case may not be there now.  In any case, damage control must be a prime design requirement.

YES

[postponed item](https://github.com/rclone/rclone/pull/5164#issuecomment-839764078) "Track critical filtering and change detection flags" will snapshot selected settings (backend options, command-line flags, environment) and keep them in the _per-session_ file together with filters file hash (instead of the per-filter file). On the next runs it will detect critical changes and maybe request a resync.

> `--backup-dir` sounds appropriate.

> postponed item "Delete into --backup-dir"

> Are CLI filter options accepted by bisync?

Yes. Not tested yet.

> 4. One user [offered up](https://github.com/cjnaz/rclonesync-V2/issues/68) an inotify change notification technique, along with a Dropbox change notification solution for triggering an rclonesync.  If we could get change notifications from all the remotes then perhaps bisync itself is not needed.

Covered by #4152 (LocalFS notify, open), #2949 (Dropbox notify, done recently), #249 (Run sync on notify, open).
Not in scope for this PR.

> 5. rclonesync is not robust for the case of a file no longer there after the current LSL snapshot was taken.  Also, if a file is being written to / blocked when rclonesync wants to copy it.  In such cases rclonesync throws up its hands and forces user interaction via a --first-sync.

Pretty close to #840 (recheck source during operation, open)
Related to #3770 (requests _.partial_ file support in LocalFS / FTP / SFTP, which can be _filtered out_ by bisync)

ivandeex:
> One rclonesync user had abut 2M files and ran out of memory.  I optimized rclonesync to get it down to two in-memory file listings at any time.  I understand that rclone deals with the file system recursively by directory rather than the whole tree?

Listings keep a lot of self-recursive path strings:
`[/movies/]alpha`, `[movies/]bravo`, `[movies/][zeta/]hello`, `[movies/][zeta/]world`.
This gives a good possibility for compression using _trie_ (using by-character or by-path segmentation).

In short, I want to make something like a modifed [dghubble/trie](https://github.com/dghubble/trie) with 3 fast primitives: _add path_, _map path to int32_, _map int32 to path_. I'd fill it up when a prior listing is parsed or new one generated. Delta engine and listing functions will pass the trie as a shared per-session object and use int32 instead of file names.

This is another postponed item. I'd rather start from thousands, then proceed to zillions.

ivandeex:
@cjnaz

> 
> > rclonesync is not robust for the case of a file no longer there after the current LSL snapshot was taken.

Such files should be just **skipped** (by default or under a command-line flag) similar to https://github.com/rclone/rclone/pull/5297#pullrequestreview-649944326

> 
> > Also, if a file is being written to / blocked when rclonesync wants to copy it.  In such cases rclonesync throws up its hands and forces user interaction via a --first-sync.
>
> Pretty close to #840 (recheck source during operation, open)
> Related to #3770 (_.partial_ file support in LocalFS / FTP / SFTP, which can be _filtered out_ by bisync)

I think rclonesync approach is right. If bisync detects a volatile file, it should abort operation or just skip the file (depending on a flag e.g. `--force`), then request a resync **and recommend adding this file to a filter.**

cmd/bisync/deltas.go
ivandeex: I have added a related _optional item_

ivandeex:
@Jwink3101 

> 
> > Hashing the user's filter file in rclonesync addressed a bug where many files were deleted because they did not show up in the current state list and were in the prior history list, so they needed to get deleted on the other side
> 
> Again, I am not disagreeing that it is an okay design for a _wrapper tool_. It is just that it **diverges from rclone's current behavior**. Consider:
> ```
> $ rclone sync remote1:interesting-dir remote2:
> ```
> and then
> ```
> $ rclone sync remote1:empty-dir remote2:
> ```
> poof. There goes your files.

Yesterday I added the new item "Track critical filtering and change detection flags in sessions," to the [TODO list](https://github.com/rclone/rclone/pull/5164#issuecomment-839764078).
Naturally, we could add another flag `--saved-sessions-only` to abort if combination `remote1:empty-dir remote2:`
has not been used with bisync before and prevent **poof**.

User data safety also belongs with item "Honor --interactive and --auto-confirm flags" (these flags have long been supported by `sync` and other destructive commands).

> I am all for guardrails. But lets also not diverge from the way things work now.

Diverging from an "rclone way" belongs with overall project health. I'll wait for @ncw to take decision.

cjnaz:
One critical design feature / flaw / weakness of rclonesync is that at the
end of the operation the two paths, and thus the two LSL files must match.
If not, then the sync has failed and a --first-sync is forced.  So if an
individual file can't be copied then there is no recovery other than a
--first-sync.  In the future this will be something to be fixed - so that
transient out-of-sync can be recovered from.  Perhaps track mismatches and
retry in the next run?

cmd/bisync/cmd.go
ivandeex: I guess that the cobra framework in rclone expects me to use only `cmdDefinition.Run(myFuncWithoutErrorReturn)`,
never return errors and use the `md.Run(myFailingFunc)` high-level retries inside.
But what if I want `cmdDefinition.RunE(myFuncWithError)` so Bisync can just fail fast?
I'm still confused why **any** failure is considered a CLI syntax error calling for help?
Please advise.

ivandeex:
@ncw

Please review

cjnaz:
@ivandeex, I installed your latest...
```
$ ./v1.55.1-iva18/rclone -V
rclone v1.55.1-iva18
- os/version: centos 7.9.2009 (64 bit)
- os/kernel: 3.10.0-1160.25.1.el7.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.16.4
- go/linking: static
- go/tags: none
```
and started playing with a small pair of local directories, and I'm impressed.  Nice work.  :>

Minor issues

### 1 - Attempting to --resync to a non-existing directory gives an unclear error message

Dir junkp2 did not exist on this initial run:
```
$ ./v1.55.1-iva18/rclone bisync junkp1 junkp2 -1
2021/06/09 20:51:45 ERROR : : error listing: directory not found
2021/06/09 20:51:45 NOTICE: Bisync aborted. Please try again.
Error: directory not found
2021/06/09 20:51:45 Fatal error: directory not found
```
Should the first ERROR line have listed `junkp2`?  Perhaps drop lines 3 & 4?

### 2 - --check-access matches files _starting with_ the check filename

After a clean --resync, I renamed the `junkp2\RCLONE_TEST` to `junkp2\RCLONE_TESTX`, resulting in:
```
$ ./v1.55.1-iva18/rclone bisync junkp1 junkp2 --check-access
2021/06/09 20:53:35 ERROR : -          Access test failed: Path1 file not found in Path2 - RCLONE_TEST
2021/06/09 20:53:35 ERROR : -          Access test failed: Path2 file not found in Path1 - RCLONE_TESTX
2021/06/09 20:53:35 ERROR : Bisync critical error: check file check failed
2021/06/09 20:53:35 ERROR : Bisync aborted. Must run --resync to recover.
```
I don't think the Path2 (junkp2) RCLONE_TESTX file should have been found and used for a crosscheck to Path1.  I confirmed that `--check-filename file1` matches all files that start with `file1...` in Path1/Path2.

### 3 - .lst files contain md5 hashes of the local files

The remotes are nice in that they seem to calculate hashes as part of their background operations, so requesting a hash is cheap.  However, for local (and sftp?) I assume that the hashes must be calculated in-line and for every run.  This can lead to long run times if there are lots of files and big files on the local filesystem.  And since I don't think you are yet using hashes within bisync, perhaps drop local hash generation for now, and make it a switch option for future support?

cjnaz:
Data point...
The TiBU directory has 19GB of files.  Path2 is a local dir.  2+ minutes for the `Path2 checking for diffs` step.

I tried the --ignore-checksum switch to no avail.

```
2021/06/10 14:36:26 INFO  : Synching Path1 "owncloud:TiBU/" with Path2 "/.../TiBU/"
2021/06/10 14:36:26 INFO  : Path1 checking for diffs
2021/06/10 14:36:26 INFO  : Path2 checking for diffs
2021/06/10 14:38:37 INFO  : No changes found
```

ivandeex:
> I tried the --ignore-checksum switch to no avail.

It's on official todo list: https://github.com/rclone/rclone/pull/5164#issuecomment-839764078

cmd/bisync/cmd.go
ivandeex: @ncw 
https://github.com/rclone/rclone/pull/5164/commits/d94b77b3dd4d2e47be2eaee3ff8cf4b558eba2a6
Does this fix look right?

ivandeex: @ncw 
Does the fix https://github.com/rclone/rclone/pull/5164/commits/d94b77b3dd4d2e47be2eaee3ff8cf4b558eba2a6 look right for you?

cmd/bisync/bilib/hash.go
ivandeex: Simple hash aliases were merged in core by #5339.
So I removed this kludge by https://github.com/rclone/rclone/pull/5164/commits/fcb9cbe1dbdfb7e4ac892f1860e5e2b1d155d1b1.

ivandeex:
@ncw @jtagcat @cjnaz
I relased [bisync beta 2](https://github.com/ivandeex/rclone/releases/tag/v1.55.1-bisync02) for your evaluation. Please try it.

@jtagcat
> I'd make a recommendation ... that a beta warning, or similar be added...

Done. Enjoy.

@cjnaz

> ### 1 - Attempting to --resync to a non-existing directory gives an unclear error message
> 
> ```
> $ ./v1.55.1-iva18/rclone bisync junkp1 junkp2 -1
> 2021/06/09 20:51:45 ERROR : : error listing: directory not found
> 2021/06/09 20:51:45 NOTICE: Bisync aborted. Please try again.
> Error: directory not found
> 2021/06/09 20:51:45 Fatal error: directory not found
> Should the first ERROR line have listed `junkp2`? Perhaps drop lines 3 & 4?
> ```

I made lines 3&4 squash into one, but fixing empty directory in the `ERROR : : error listing: directory not found` message is not so easy as the trick is not in my code but in the commonly used [rclone directory walker](https://github.com/rclone/rclone/blob/v1.55.1/fs/walk/walk.go#L163).
Let's bear with it for now.

> ### 2 - --check-access matches files _starting with_ the check filename
> I don't think the Path2 (junkp2) RCLONE_TESTX file should have been found and used

Bug fixed [here](https://github.com/ivandeex/rclone/commit/2379864d13544340a20e8952cee8f3fdb3190eeb#diff-25f7d45d3740899fdbbb93ff87a682943b7085b9c2f7f67b9cc58756c90e3630L164)

> ### 3 - .lst files contain md5 hashes of the local files
> perhaps drop local hash generation for now, and make it a switch option for future support?
> I tried the --ignore-checksum switch to no avail.

The beta contains a *quick fix*: with --ignore-checksum hashes will never be calculated (local fs, remote fs, whatever). This is not a real solution but a kludge for your test.
For a real solution we have to develop a number of _rclonic_ rules for all case combinations: what to do if listings contain hash but user flagged ignore, how future name tracker will interact with this flag, how --checksum and --ignore-checksum interact, etc, etc... It's a separate discussion and I'll drop this quick fix after all questions get solved.
For now you can try `--ignore-checksum` and evaluate the speed.

@ncw 
This thread has grown beyond control. I hope you won't lose all inline threads squashed by github that need your attention!

Ivan

cjnaz:
RCLONE_TEST matches fixed.  --ignore-checksum brought the 19GB no-changes run down from 2 minutes to less than 1 second.  Thanks for the interim hack.

I tried --no-cleanup with the goal of looking at the generated copy/delete list files.  Not finding them in the ~/.cache/rclone directory I assume that you manage these queue lists in memory (sounds appropriate).  What piqued my curiosity was seeing all the file "Excluded" debug messages while executing the queued copies.  Does --no-cleanup leave anything behind other than the .lst.new files?  Probably no issues here.

ivandeex:
> Does --no-cleanup leave anything behind other than the .lst.new files?

Only `.lst.new`. Queues are saved only by internal flag set by `go test`.

ivandeex:
@ncw
I think we found a problem with `walk.Walk*` [here](https://github.com/rclone/rclone/blob/v1.55.1/fs/walk/walk.go#L163). In case of a local path the function will receive the root of the path passed in `fs.Root`. The remainder in the `path` argument will be empty. This makes the `fs.Errorf(path,"<msg>: %v", err)` produce strange messages like `ERROR : : error listing: directory not found`. The quick fix would be `fs.Errorf(f, "<msg> failed at path %s: %v", path, err)` which IMHO will look least strange when path is empty. The **right** fix is to delegate error printing to caller. This deserves a separate ticket. Thoughts?

ivandeex:
@ncw
I have an idea how to structure bisync work for near future. Let's create an umbrella issue ticket and put the digest of open items there. This is convenient for prioritizing and higher level look. Items deemed urgent (e.g. "extend test harness to backends without modtime" or whatever) can be discussed in separate tickets. While this pull request is open, new patches will go here for code review. When you decide that initial bisync command deserves merging into master, this pull request gets merged closing some issues. Remaining issues will then have their dedicated pull requests. The umbrella issue keeps tracking open items, open issues and pull request(s).

ivandeex:
This comment is an attempt to triage open items from https://github.com/rclone/rclone/pull/5164#issuecomment-839764078 and select priority items required for this PR to complete, leaving the remainder to future PRs.

- [ ] Make test suite work with imprecise modtime and backends without modtime
- [ ] Make bisync delta engine aware of backend modtime tolerance (depends on the test suite readiness, above)
- [ ] Add size checks to the delta engine and relevant test cases
- [ ] Save critical filtering and change detection flags in sessions, Add flag `--use-saved-options` or **by default restore such flags from the session** json, allowing user to change such flags only when `--resync`.
- [ ] Honor `--checksum` / `--ignore-checksum` to control hashsum use

Waiting for @ncw to discuss/approve this plan as well as provide 1st code review.

ivandeex:
@ncw 
Will you find time for at least one peer review before this September?

ncw:
> Will you find time for at least one peer review before this September?

I hope so! I've been on vacation and been busy with some other stuff, but both those things are drawing to a close!

cmd/bisync/cmd.go
```diff
@@ -140,14 +140,15 @@ var commandDefinition = &cobra.Command{
...
-		// FIXME Why runtime errors make rclone print usage?
-		command.SilenceUsage = true
-
-		err := Bisync(ctx, fs1, fs2, &opt)
-		if err == ErrBisyncAborted {
-			os.Exit(2)
-		}
-		return err
+		fs.Logf(nil, "bisync is EXPERIMENTAL. Don't use in production!")
+		cmd.Run(false, true, command, func() error {
```
ncw, line 144: This looks like the rclone way to me :-)

ncw (approved):
I've been through the code (at a medium level) and docs and left comments.

The code and docs look great!

I think docs will be really important as part of the user experience.

Once question which I have...

Will the user be able to use something like `--backup-dir` or `--suffix`? I can see users wanting that to prevent data loss. Maybe `--suffix` works already and you could filter the backups out using the filtering? Perhaps this should be automatic.

I'm going to approve this as I'd like you to merge it when you are ready so we can get it into v1.57.

Thank you for a truly amazing piece of work :-)

cmd/bisync/help.go
```diff
@@ -17,10 +17,182 @@ func makeHelp(help string) string {
...
+
+If you or your application should change the content of a file
+without changing the modification time then bisync will _not_
+notice the change, and thus will not copy it to the other side.
+
+Note that on some cloud storage systems it is not possible to have file
+timestamps that match _precisely_ between the local and other filesystems.
+Bisync's approach to this problem is by tracking the Path1-to-Path1 and
+Path2-to-Path2 deltas on each side _separately_, and then applying
+the resulting changes on the other side.
```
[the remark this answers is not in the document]
ivandeex: Right. 
Will rephrase this ~~RUnglish~~.

cmd/bisync/help.go
```diff
@@ -17,10 +17,182 @@ func makeHelp(help string) string {
...
+status and history of the Path1 and Path2 filesystems cannot be trusted,
+so it is safer to block any further changes until someone checks things out.
+The recovery is to do a |--resync| again.
+
+It is recommended to use |--resync --dry-run --verbose| initially and
+_carefully_ review what changes will be made before running the |--resync|
+without |--dry-run|.
+
+Most of these events come up due to a error status from an internal call.
+On such a critical error the |{...}.path1.lst| and |{...}.path2.lst|
```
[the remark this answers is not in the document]
ivandeex: All work artifacts are stored under `$RCLONE_CACHE_DIR/bisync` (usually `~/.cache/rclone/bisync`).
Will mention.

cmd/bisync/help.go
```diff
@@ -17,10 +17,182 @@ func makeHelp(help string) string {
...
+to individual directories within the tree may be set up, e.g.:
+|path_to_local_tree..dropbox_subdir.lst|.
+
+Arbitrary rclone flags may be specified on the bisync command line.
+The most important bisync flags are listed below.
+
+#### --resync
+This will effectively make both Path1 and Path2 filesystems contain a
+matching superset of all files. Path2 files that do not exist in Path1 will
+be copied to Path1, and the process will then sync the Path1 tree to Path2.
```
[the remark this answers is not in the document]
ivandeex: Yes.
It's implemented differently a.t.m. but the result is equivalent to your sequence.
Probably can be internally implemented right so, if it improves performance and keeps tests passing.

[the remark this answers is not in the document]
ivandeex: ~~Yes. It's implemented differently a.t.m. but the result is equivalent to your sequence. Probably can be internally implemented right so, if it improves performance and keeps tests passing.~~
Sorry. I'll reply later.

cmd/bisync/help.go
```diff
@@ -17,10 +17,182 @@ func makeHelp(help string) string {
...
+Path1 and Path2 are treated equally, in that neither has priority for
+file changes, and access efficiency does not change whether a remote
+is on Path1 or Path2.
+
+The listings in bisync working directory (default: |{WORKDIR}|)
+are named based on the Path1 and Path2 arguments so that separate syncs
+to individual directories within the tree may be set up, e.g.:
+|path_to_local_tree..dropbox_subdir.lst|.
+
+Arbitrary rclone flags may be specified on the bisync command line.
```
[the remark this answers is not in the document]
ivandeex: I will implement `--backup-dir` before merging this PR.

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+  or `C:\Users\MyLogin\AppData\Local\rclone\bisync` on Windows.
+  Make sure that this location is writable.
+- Run bisync with the `--resync` flag, specifying the paths
+  to the local and remote sync directory roots.
+- For successive sync runs, leave off the `--resync` flag.
+- Consider using a [filters file](#filtering)
+  for excluding unnecessary files and directories from the sync.
+- Consider setting up the [--check-access](#check-access-option)
+  feature for safety.
+- On Linux, consider setting up a [crontab entry](#cron).
```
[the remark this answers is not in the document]
ivandeex: Yes, there is a lock file, described elsewhere in the docs.
Will refer to it in this place in text.

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+### Renamed directories
+
+Renaming a folder on the Path1 side results is deleting all files on
+the Path2 side and then copying all files again from Path1 to Path2.
+Bisync sees this as all files in the old directory name as deleted and all
+files in the new directory name as new. Similarly, renaming a directory on
+both sides to the same name will result in creating `..path1` and `..path2`
+files on both sides.
+Currently the most effective and efficient method of renaming a directory
+is to rename it on both sides, then do a `--resync`.
```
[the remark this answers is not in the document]
ivandeex: I won't be able to implement this by 1.57
Probably in 1.58

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+vs. case dependent Linux
+
+## Filtering {#filtering}
+
+See [filtering documentation](/filtering/)
+for how filter rules are written and interpreted.
+
+Bisync's [`--filters-file`](#filters-file) flag slightly extends the rclone's
+[--filter-from](/filtering/#filter-from-read-filtering-patterns-from-a-file)
+filtering mechanism.
```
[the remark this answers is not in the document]
ivandeex: Bisync will track changes in the filters file set by `--filters-file` and enforce `--resync` if any.
By the merge time I'm going to extend this idea to other sync-related flags, so we can ensure even more safety.

Probably bisync should set hook on every `--filter-from` file making this special flag moot.
I've got to think this over.

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+2021/05/12 00:49:40 ERROR : Attempt 3/3 failed with 3 errors and: not deleting files as there were IO errors
+2021/05/12 00:49:40 ERROR : Failed to sync: not deleting files as there were IO errors
+2021/05/12 00:49:40 NOTICE: WARNING  rclone sync try 3 failed.           - /path/to/local/tree/
+2021/05/12 00:49:40 ERROR : Bisync aborted. Must run --resync to recover.
+```
+
+### Illegal filenames
+
+Some cloud services allow characters in filenames that are not legal on the
+local filesystem, such as `/`, `:`.
```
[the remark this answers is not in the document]
ivandeex: Yes, this paragraph is outdated and should be rewritten.

docs/content/bisync.md
```diff
@@ -0,0 +1,1090 @@
...
+that by default cannot be copied from Google Drive to the local filesystem.
+If you are having problems, run with `--verbose` to see specifically which
+files are generating complaints. If the error is
+`This file has been identified as malware or spam and cannot be downloaded`,
+consider using the flag
+[--drive-acknowledge-abuse](/drive/#drive-acknowledge-abuse).
+
+### Google Doc files
+
+Google docs exist as virtual files on Google Drive. While it is possible to
```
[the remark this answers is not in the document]
ivandeex: Yes, Will do.

ivandeex:
Rebased on current master, created dedicated branch https://github.com/rclone/rclone/tree/pr-bisync to produce beta builds on https://beta.rclone.org

ivandeex:
**Note:**
When #5401 gets merged, I should move `cmd/bisync/bilib` stuff to new `lib/osutil` - https://github.com/rclone/rclone/pull/5401#issuecomment-859599396

cjnaz:
Hi Ivan,
For the bisync docs References section, please add an attribution note something to the effect of...

- rclone's bisync implementation was derived from the [rclonesync-V2](https://github.com/cjnaz/rclonesync-V2) project, including documentation and test mechanisms, with @cjnaz 's full support and encouragement.

Thank you, and nice work!

ivandeex:
@ncw @cjnaz 
So my plan is to:

1. Add to this PR's branch only the below fixes:
- attribution notes requested at https://github.com/rclone/rclone/pull/5164#issuecomment-921954290
- a fix for `rmdirs` test which started to fail after recent rebase on master
- minor docs cleanups requested on online threads here

2. I will merge _right after_ 1.57 is out so the result will _not_ be in 1.57 but will be available for download as a _1.58-dev_ official beta.

3. Work on items from https://github.com/rclone/rclone/pull/5164#issuecomment-881324582 and submit them as pull requests, hopefully before 1.58, the first _public release_ to include bisync.

4. Triage remaining comments and proposals from ~~kitchen sink~~ this ticket and linked issue into organized set of tickets under umbrella of the bisync project.

ivandeex:
@ncw @cjnaz 
So my plan is to:

1. Add to this PR's branch only the below fixes:
  - attribution notes requested at https://github.com/rclone/rclone/pull/5164#issuecomment-921954290
  - a fix for `rmdirs` test which started to fail after recent rebase on master
  - minor documentation cleanups requested in the inline threads above

2. I will merge into master _right after_ 1.57 published so the result will _not_ be in 1.57 but will be available for download as a _1.58-dev_ official beta. This idea resembles Nginx with unstable odd vs stable even releases... Note that many users in rclone community actively use the beta builds. They will be able to try the feature right away and submit reports.

3. Work on items from https://github.com/rclone/rclone/pull/5164#issuecomment-881324582 and submit them as pull requests, hopefully before 1.58, the first _public release_ to include bisync.

4. Triage remaining comments and proposals from ~~kitchen sink~~ this ticket and linked issue into organized set of tickets under umbrella of the bisync project.

ivandeex:
Like I planned, I am merging bisync right after 1.57.0


@ncw 
You might want to remove the `pr-bisync` branch snd clean up download server under https://beta.rclone.org/branch/pr-bisync/

ivandeex:
@ncw 
Could you ensure that merged documentation gets published on https://tip.rclone.org ?

Asmageddon:
<3 So much love for all this work.

PR #5198, opened 2021-04-03
sync: don't warn about --no-traverse when --files-from is set

nickgaya:

Suppress spurious error message when performing `rclone sync` with `--files-from` and `--no-traverse` flags.

The `rclone sync` command logs an error if the `--no-traverse` flag is specified, as traversing destination directories is necessary to compare them with the source for syncing.

However, when the `--files-from` flag is present, the `--no-traverse` flag modifies the file listing behavior to query each specified path individually rather than traversing the directory listing (see the [docs](https://rclone.org/filtering/#files-from-read-list-of-source-file-names)).  This functionality works as expected with the sync operation, so this PR adds a check to suppress the error message in this case.

No

#### Checklist

- [ ] ~~I have added tests for all changes in this PR if appropriate.~~ (n/a)
- [ ] ~~I have added documentation for the changes if appropriate.~~ (n/a)


ncw (approved):
That looks great - thanks.

I'll merge it now.

PR #5205, opened 2021-04-03
crypt: support timestamped filenames from --b2-versions (attempt 2)

### LINKED ISSUE #1627: --b2-versions doesn't work for crypt backed by B2

domyd:

Continues #4024.

Fixes #1627 

I refactored the versioning code into its own package `lib/version`, including the filename versioning tests in `backend/b2`.

While `rclone --b2-versions ls TestCryptB2:` would now correctly list all file versions in plaintext, `rclone --b2-versions copy TestCryptB2:testfile-v(...).txt tmp/` wouldn't work. That turned out to be due to `encryptFileName` not dealing with the versioned filenames, yet `decryptFileName` _does_, so I've also fixed that.

I'd also like to add some integration tests to check the interaction between `crypt` and `b2` - particularly for the `copy` issue I ran across - but I'm not sure how to go about it.

Yes, in the previous PR.

#### Checklist

* As noted above, an integration test between `crypt` and `b2` with `--b2-versions` is probably something we want.

### DIFF
--- a/backend/crypt/cipher_test.go
+++ b/backend/crypt/cipher_test.go
@@ -160,22 +160,29 @@ func TestEncryptFileName(t *testing.T) {
 	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s", c.EncryptFileName("1"))
 	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s/l42g6771hnv3an9cgc8cr2n1ng", c.EncryptFileName("1/12"))
 	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s/l42g6771hnv3an9cgc8cr2n1ng/qgm4avr35m5loi1th53ato71v0", c.EncryptFileName("1/12/123"))
+	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s-v2001-02-03-040506-123", c.EncryptFileName("1-v2001-02-03-040506-123"))
+	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s/l42g6771hnv3an9cgc8cr2n1ng-v2001-02-03-040506-123", c.EncryptFileName("1/12-v2001-02-03-040506-123"))
 	// Standard mode with directory name encryption off
 	c, _ = newCipher(NameEncryptionStandard, "", "", false)
 	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s", c.EncryptFileName("1"))
 	assert.Equal(t, "1/l42g6771hnv3an9cgc8cr2n1ng", c.EncryptFileName("1/12"))
 	assert.Equal(t, "1/12/qgm4avr35m5loi1th53ato71v0", c.EncryptFileName("1/12/123"))
+	assert.Equal(t, "p0e52nreeaj0a5ea7s64m4j72s-v2001-02-03-040506-123", c.EncryptFileName("1-v2001-02-03-040506-123"))
+	assert.Equal(t, "1/l42g6771hnv3an9cgc8cr2n1ng-v2001-02-03-040506-123", c.EncryptFileName("1/12-v2001-02-03-040506-123"))
 	// Now off mode
 	c, _ = newCipher(NameEncryptionOff, "", "", true)
 	assert.Equal(t, "1/12/123.bin", c.EncryptFileName("1/12/123"))
 	// Obfuscation mode
 	c, _ = newCipher(NameEncryptionObfuscated, "", "", true)
 	assert.Equal(t, "49.6/99.23/150.890/53.!!lipps", c.EncryptFileName("1/12/123/!hello"))
+	assert.Equal(t, "49.6/99.23/150.890/53-v2001-02-03-040506-123.!!lipps", c.EncryptFileName("1/12/123/!hello-v2001-02-03-040506-123"))
+	assert.Equal(t, "49.6/99.23/150.890/162.uryyB-v2001-02-03-040506-123.GKG", c.EncryptFileName("1/12/123/hello-v2001-02-03-040506-123.txt"))
 	assert.Equal(t, "161.\u00e4", c.EncryptFileName("\u00a1"))
 	assert.Equal(t, "160.\u03c2", c.EncryptFileName("\u03a0"))
 	// Obfuscation mode with directory name encryption off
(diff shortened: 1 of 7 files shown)


domyd:
I'm not sure what that Race test failure is about, seems like a flaky test? It passed the first time around.

ivandeex:
Most prominent flaky tests: cmd/mount cmd/cmount TestMaxTransfer/Cautious
Rerunning test suite: click "Rerun all jobs" from the "Check" tab. If you don't have permissions for that, the `git commit --amend --no-tedit; git push --force`

ivandeex:
@ncw can give good suggestions about adding integration tests

ncw (commented):
I'm finding this rather difficult to review...

Can you swap the order of the commits please? So the first commit is `b2: factor version handling into lib/version` and the second is `crypt: support timestamped filenames from --b2-versions`

Let me know if you need help with this. (I would `git reset HEAD~2` then commit the b2 + lib/versions changes then the crypt changes in two separate commits).

Thanks

domyd:
You're right, that makes much more sense. I was mostly concerned with preserving @wlritchi's commit, before I realized that there's a co-authoring feature. This should now be easier to review and still preserve authorship.

docs/content/b2.md
```diff
@@ -172,11 +172,6 @@ the file instead of hiding it.
 Old versions of files, where available, are visible using the 
 `--b2-versions` flag.
 
-**NB** Note that `--b2-versions` does not work with crypt at the
```
ncw, line 175: This feature should probably get a mention in the docs but not here!

backend/crypt/cipher_test.go
```diff
@@ -194,20 +197,37 @@ func TestDecryptFileName(t *testing.T) {
 		{NameEncryptionStandard, true, "p0e52nreeaj0a5ea7s64m4j72s/l42g6771hnv3an9cgc8cr2n1ng/qgm4avr35m5loi1th53ato71v0", "1/12/123", nil},
 		{NameEncryptionStandard, true, "p0e52nreeaj0a5ea7s64m4j72s/l42g6771hnv3an9cgc8cr2n1/qgm4avr35m5loi1th53ato71v0", "", ErrorNotAMultipleOfBlocksize},
 		{NameEncryptionStandard, false, "1/12/qgm4avr35m5loi1th53ato71v0", "1/12/123", nil},
+		{NameEncryptionStandard, true, "p0e52nreeaj0a5ea7s64m4j72s-v9999-99-99-000000-000", "", base32.CorruptInputError(26)},
```
ncw, line 200: I think I'd rather see some tests here with valid date suffixes too than your extra loop adding the suffixes programatically and checking them I think.

That would mean that if date suffix adding code changes we'll get a warning about it in the tests and we can see explicitly in the tests what we have.

lib/version/version.go
```diff
@@ -0,0 +1,52 @@
...
+package version
+
+import (
+	"path"
+	"regexp"
+	"strings"
+	"time"
+)
+
+const versionFormat = "-v2006-01-02-150405.000"
```
ncw, line 12: The exact number of significant figures here the `.000` depends on the backend. For b2 it is 3, if we implement this for AWS s3 it would probably be 9.

However 3 sig figs is probably enough and we shouldn't worry about it.

backend/crypt/cipher_test.go
```diff
@@ -194,20 +197,37 @@ func TestDecryptFileName(t *testing.T) {
...
+		timestampedExpected := version.Add(testSpec.expected, timestamp)
+		if testSpec.expectedErr != nil {
+			timestampedExpected = ""
+		}
+		for _, test := range []struct {
+			in          string
+			expected    string
+			expectedErr error
+		}{
+			{testSpec.in, testSpec.expected, testSpec.expectedErr},
```
ncw, line 222: I think this code is just a bit too clever - see comment above :-)

ncw (commented):
I think that is looking very nice.

Thanks for swizzling the commits and well done for figuring out the co-author feature.

I put a very small number of comments inline for you to look at.

I don't think this feature can ever clash with standard name encryption (as - isn't valid base32) or no name encryption. However it could clash with obfuscation...

Not sure what to do about that. Nothing probably as I think it is very unlikely.

docs/content/b2.md
domyd: I'm not quite sure what you mean. Since this is merely a bugfix for B2 users, the removal of this notice seems appropriate? Where else might I change the docs?

domyd:
I've made the requested changes to the tests, and also added a few more test cases.

Something I noticed is that the version adding is _really_ awkward on obfuscated file names (e.g. `hello-v2021-04-06-183434-000` turns into `20-v2021-04-06-183434-000.AxEEH`), due to how it tries to be clever with file extensions. I tried having `version.Add` put the version string at the end of the filename when using obfuscation, but that ended up not working since the b2 backend doesn't know when obfuscation is being used. Fixing that would require b2 - and any other backend using the version strings - to know about obfuscated file names somehow, not sure if it's worth? You'd only notice if you were to access the raw `crypt` backend with `--b2-versions` and obfuscation enabled, so it's probably really niche.

ncw:
> I've made the requested changes to the tests, and also added a few more test cases.

Thank you - this looks great now, I will merge it in a moment.
 
> Something I noticed is that the version adding is _really_ awkward on obfuscated file names (e.g. `hello-v2021-04-06-183434-000` turns into `20-v2021-04-06-183434-000.AxEEH`), due to how it tries to be clever with file extensions. I tried having `version.Add` put the version string at the end of the filename when using obfuscation, but that ended up not working since the b2 backend doesn't know when obfuscation is being used. Fixing that would require b2 - and any other backend using the version strings - to know about obfuscated file names somehow, not sure if it's worth? You'd only notice if you were to access the raw `crypt` backend with `--b2-versions` and obfuscation enabled, so it's probably really niche.

Yes you are right - the namespace for obfuscated names and versions overlaps which is annoying but I don't think we can do anything sensible about it.

Thank you for taking this up and finishing it.

enginefeeder101:
@domyd Thank you for making the effort and finishing the work started by @wlritchi
Hopefully this will hit stable soon, I am not fond of using the master branch on my main backup remote to retrieve old versions.

PR #5215, opened 2021-04-06
Add --size flag for rclone rcat - closes #4403

### LINKED ISSUE #4403: Add a "hint-file-size" option to rcat to allow "streaming" transfers to cloud hostings which doesn't allow it directly
rclone v1.52.2
- os/arch: linux/amd64
- go version: go1.14.4

I need to transfer a HUGE... Nope, not a file. A block device image. 2 TB in size - to a remote storage which *does not support* streaming uploads.

The obvious solution is using someting like `cat /dev/sdb | rclone rcat`, but...

The problem is that it does attempt create a "spool file", except that this will end up in a failure because, well, fitting a 2 TB file on a filesystem which can never be that big is impossible, and I can't add another 4 TB drive just for the sake of backing it up. And anyway how will I then back up this 4 TB drive?...

Add to "rcat" command an option "hint-file-size", which would allow to preallocate the file in advance at the remote end and then just actually stream it via standard chunked transfer mechanics (with maybe using a limited-sized cache file, but definitely not copying the whole file in the cache before actually transferring anything).

nazar554:
This allows preallocating space at remote end with `RcatSize`.

To allow streaming uploads even when remote backend doesn't support them.
`RcatSize` should fall back to regular `Rcat` if `--size` flag is missing.

See #4403

#### Checklist

### DIFF
--- a/cmd/rcat/rcat.go
+++ b/cmd/rcat/rcat.go
@@ -47,6 +47,10 @@ setting this cutoff too high will decrease your performance.
 Use the |--size| flag to preallocate the file in advance at the remote end
 and actually stream it, even if remote backend doesn't support streaming.
 
+|--size| should be the exact size of the input stream in bytes. If the
+size of the stream is different in length to the |--size| passed in
+then the transfer will likely fail.
+
 Note that the upload can also not be retried because the data is
 not kept around until the upload succeeds. If you need to transfer
 a lot of data, you're better off caching locally and then
--- a/cmd/rcat/rcat.go
+++ b/cmd/rcat/rcat.go
@@ -7,12 +7,19 @@ import (
 	"time"
 
 	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
 
+var (
+	size = int64(-1)
+)
+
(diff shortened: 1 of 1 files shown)


cmd/rcat/rcat.go
```diff
@@ -37,6 +44,9 @@ must fit into RAM. The cutoff needs to be small enough to adhere
 the limits of your remote, please see there. Generally speaking,
 setting this cutoff too high will decrease your performance.
 
+Use the |--size| flag to preallocate the file in advance at the remote end
+and actually stream it, even if remote backend doesn't support streaming.
+
```
ncw, line 49: I think this needs a bit more explanation - how about this as well?

```
|--size| should be the exact size of the input stream in bytes. If the
size of the stream is different in length to the |--size| passed in
then the transfer will likely fail.
```

ncw (commented):
I think this looks good.

Suggestion for extra docs inline - what do you think?

Thanks

Nick

cmd/rcat/rcat.go
nazar554: Yep, I agree. I was actually testing this locally with `zfs send` and I had to use `split --filter` and `head -c /dev/zero` to pad the stream to correct length.

ncw (approved):
Looks perfect - thanks.

Will merge now

PR #5223, opened 2021-04-08
config: fix issues with memory-only config file paths

### LINKED ISSUE #5222: Memory-only config file issues

https://forum.rclone.org/t/rclone-1-55-with-config-dev-null/23355

Documentation at https://rclone.org/docs/#config-config-file:

> If the location is set to empty string "" or the special value /notfound, or the os null device represented by value NUL on Windows and /dev/null on Unix systems, then rclone will keep the config file in memory only.

##### Case 1. `--config /dev/null` on Linux

```
rclone \
    --config /dev/null \
    sync /local/path/ :sftp:remote_path/ \
    --sftp-host hidden --sftp-port hidden --sftp-user hidden --sftp-pass hidden \
    -v --bwlimit 80k:600k --max-delete 2 --transfers 1 --checkers 4 --size-only

2021/04/07 21:55:12 ERROR : Can't save config "md5sum_command" = "md5sum" for on the fly backend ":sftp"
2021/04/07 21:55:18 ERROR : Failed to save config after 10 tries: Failed to create temp file for new config: open /dev/null804515289: permission denied
2021/04/07 21:55:18 ERROR : Can't save config "sha1sum_command" = "sha1sum" for on the fly backend ":sftp"
2021/04/07 21:55:24 ERROR : Failed to save config after 10 tries: Failed to create temp file for new config: open /dev/null062934686: permission denied
```

##### Case 2. `--config NUL` on Windows

```
2021/04/08 08:47:22 ERROR : Failed to save config after 10 tries: Failed to create temp file for new config: open \NUL324278613: Access is denied.
```

##### Case 3. `--config ""`
```
rclone \
    --config "" \
    sync /local/path/ :sftp:remote_path/ \
    --sftp-host hidden --sftp-port hidden --sftp-user hidden --sftp-pass hidden \
    -v --bwlimit 80k:600k --max-delete 2 --transfers 1 --checkers 4 --size-only

2021/04/07 22:01:06 Failed to load config file "/tmp": read /tmp: is a directory
```
/tmp was the curent directory

##### Case 4. `--config /notfound`
```
rclone \
    --config /notfound \
    sync /local/path/ :sftp:remote_path/ \
    --sftp-host hidden --sftp-port hidden --sftp-user hidden --sftp-pass hidden \
    -v --bwlimit 80k:600k --max-delete 2 --transfers 1 --checkers 4 --size-only

2021/04/07 22:02:23 NOTICE: Config file "/notfound" not found - using defaults
2021/04/07 22:02:23 ERROR : Can't save config "md5sum_command" = "md5sum" for on the fly backend ":sftp"
2021/04/07 22:02:23 ERROR : Can't save config "sha1sum_command" = "sha1sum" for on the fly backend ":sftp"
```

rclone v1.55.0

os/type: linux
os/arch: amd64
go/version: go1.16.2
go/linking: static
go/tags: cmount

albertony:

Solving issues and making it more user friendly with memory-only config.

Ref the following cases described in the issue:

Case 1. --config /dev/null on Linux
Case 2. --config NUL on Windows
Case 3. --config ""
Case 4. --config /notfound

All should be handled by this PR, ~~but I have only tested on Windows~~ tested on Windows 10 and arch linux in WSL.

Fixes #5222

#### Checklist

### DIFF
--- a/cmd/version/version_test.go
+++ b/cmd/version/version_test.go
@@ -26,12 +26,12 @@ func TestVersionWorksWithoutAccessibleConfigFile(t *testing.T) {
 	}
 	// re-wire
 	oldOsStdout := os.Stdout
-	oldConfigPath := config.ConfigPath
-	config.ConfigPath = path
+	oldConfigPath := config.GetConfigPath()
+	assert.NoError(t, config.SetConfigPath(path))
 	os.Stdout = nil
 	defer func() {
 		os.Stdout = oldOsStdout
-		config.ConfigPath = oldConfigPath
+		assert.NoError(t, config.SetConfigPath(oldConfigPath))
 	}()
 
 	cmd.Root.SetArgs([]string{"version"})
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -28,6 +28,7 @@ import (
 const (
 	configFileName       = "rclone.conf"
 	hiddenConfigFileName = "." + configFileName
+	noConfigFile         = "/notfound"
 
 	// ConfigToken is the key used to store the token under
 	ConfigToken = "token"
@@ -107,17 +108,21 @@ var (
 	// and any parents.
 	CacheDir = makeCacheDir()
 
-	// ConfigPath points to the config file
-	ConfigPath = makeConfigPath()
-
 	// Password can be used to configure the random password generator
 	Password = random.Password
+)
+
+var (
+	configPath   string
+	noConfigPath string
 )
 
 func init() {
 	// Set the function pointers up in fs
 	fs.ConfigFileGet = FileGetFlag
 	fs.ConfigFileSet = SetValueAndSave
+	noConfigPath, _ = filepath.Abs(noConfigFile)
+	configPath = makeConfigPath()
 }
 
 // Return the path to the configuration file
@@ -212,18 +217,44 @@ func makeConfigPath() string {
 	return hiddenConfigFileName
 }
 
+// GetConfigPath returns the current config file path
+func GetConfigPath() string {
+	return configPath
+}
+
+// SetConfigPath sets new config file path
+//
+// Checks for empty string, os null device, or special path, all of which indicates in-memory config.
+func SetConfigPath(path string) (err error) {
+	if path == "" || path == os.DevNull {
+		configPath = ""
+	} else {
+		if configPath, err = filepath.Abs(path); err != nil {
+			return err
+		}
+		if configPath == noConfigPath {
+			configPath = ""
+		}
+	}
+	return nil
+}
+
 // LoadConfig loads the config file
 func LoadConfig(ctx context.Context) {
 	// Set RCLONE_CONFIG_DIR for backend config and subprocesses
-	_ = os.Setenv("RCLONE_CONFIG_DIR", filepath.Dir(ConfigPath))
-
-	// Load configuration file.
+	// If empty configPath (in-memory only) the value will be "."
+	_ = os.Setenv("RCLONE_CONFIG_DIR", filepath.Dir(configPath))
+	// Load configuration from file (or initialize sensible default if no file or error)
 	if err := Data.Load(); err == ErrorConfigFileNotFound {
-		fs.Logf(nil, "Config file %q not found - using defaults", ConfigPath)
+		if configPath == "" {
+			fs.Debugf(nil, "Config is memory-only - using defaults")
+		} else {
+			fs.Logf(nil, "Config file %q not found - using defaults", configPath)
+		}
 	} else if err != nil {
-		log.Fatalf("Failed to load config file %q: %v", ConfigPath, err)
+		log.Fatalf("Failed to load config file %q: %v", configPath, err)
 	} else {
-		fs.Debugf(nil, "Using config file from %q", ConfigPath)
+		fs.Debugf(nil, "Using config file from %q", configPath)
 	}
 }
 
@@ -233,6 +264,10 @@ var ErrorConfigFileNotFound = errors.New("config file not found")
 // SaveConfig calling function which saves configuration file.
 // if SaveConfig returns error trying again after sleep.
 func SaveConfig() {
+	if configPath == "" {
+		fs.Debugf(nil, "Skipping save for memory-only config")
+		return
+	}
 	ctx := context.Background()
 	ci := fs.GetConfig(ctx)
 	var err error
@@ -244,7 +279,6 @@ func SaveConfig() {
 		time.Sleep(time.Duration(waitingTimeMs) * time.Millisecond)
 	}
 	fs.Errorf(nil, "Failed to save config after %d tries: %v", ci.LowLevelRetries, err)
-	return
 }
 
 // SetValueAndSave sets the key to the value and saves just that
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rclone/rclone/fs/config/obscure"
 	"github.com/rclone/rclone/fs/fspath"
 	"github.com/rclone/rclone/fs/rc"
+	"github.com/rclone/rclone/lib/file"
 	"github.com/rclone/rclone/lib/random"
 )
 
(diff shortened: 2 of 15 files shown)


albertony:
Ended up with a little more restructuring than perhaps strictly necessary.. Moved the main memory-only config handling from fs.config.configfile up to fs.config, this to be able to treat it more like a "feature" than an "error", and e.g. `rclone config file` command will report more informative that an in memory config is in use etc..

Part of the solution had to do with configflags doing filepath.Abs on the --config value, so value from `--config=""` and `--config="/notfound"` etc would be made absolute before the logic that checked for these values in the configfile code vere able to check for them.. The way I did this was to make the global ConfigPath into a setter and getter, so that the setter could check for special values and perform abs etc, to avoid having to do this many times - e.g. with a "IsMemoryOnlyPath" kind of solution. I don't know if this is a proper go (or rclone) way of doing it..

albertony:
A couple of small additions:

1: Checking for the reserved names on windows to avoid trouble, because even a path like this `--config=C:\Temp\NUL.txt` will be treated as the NULL device - any paths with reserved name as basename without extension! 😲 

2: Lowering message about "using defaults" from notice to debug when using memory config.

Originally did this:
```
rclone.exe --config=NUL lsd .
2021/04/08 21:30:46 NOTICE: Config is memory-only - using defaults
          -1 2021-04-08 21:28:33        -1 .git
          -1 2021-03-17 16:40:53        -1 .github
          -1 2021-04-08 21:26:59        -1 .vscode
          -1 2021-01-29 23:22:07        -1 backend
```
To match this:
```
rclone.exe --config=xxx lsd .
2021/04/08 21:30:55 NOTICE: Config file "C:\\CurrentPath\\xxx" not found - using defaults
          -1 2021-04-08 21:28:33        -1 .git
          -1 2021-03-17 16:40:53        -1 .github
          -1 2021-04-08 21:26:59        -1 .vscode
          -1 2021-01-29 23:22:07        -1 backend
```

But found it to be a bit "anoying", and figured that when you are using --config=NUL or any of the other special values then you don't need a reminder every time that you are using it..?

cmd/version/version_test.go
```diff
@@ -26,12 +26,12 @@ func TestVersionWorksWithoutAccessibleConfigFile(t *testing.T) {
 	}
 	// re-wire
 	oldOsStdout := os.Stdout
-	oldConfigPath := config.ConfigPath
-	config.ConfigPath = path
+	oldConfigPath := config.GetConfigPath()
+	_ = config.SetConfigPath(path)
```
ncw, line 30: Anywhere you see this in a test

    _ = config.SetConfigPath(path)

It should be

    require.NoError(t, config.SetConfigPath(path))

There are rather a lot of these in this, and I'll just comment on this one!

fs/config/config.go
```diff
@@ -229,6 +229,14 @@ func SetConfigPath(path string) (err error) {
 	if path == "" || path == os.DevNull {
 		configPath = ""
 	} else {
+		if runtime.GOOS == "windows" {
+			// On Windows prevent referring to legacy device, which will happen if basename without extension
+			// is one of the reserved names (https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file)
+			if reserved, _ := regexp.MatchString(`^(?i:con|prn|aux|nul|com[1-9]|lpt[1-9])(?:\.|$)`, filepath.Base(path)); reserved {
```
ncw, line 235: Ewww!  You could put this into `lib/file` as `IsReserved` or something like that?

I note there is an implementation of this in the standard library but it isn't exported :-(

```go
/ reservedNames lists reserved Windows names. Search for PRN in
// https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file
// for details.
var reservedNames = []string{
        "CON", "PRN", "AUX", "NUL",
        "COM1", "COM2", "COM3", "COM4", "COM5", "COM6", "COM7", "COM8", "COM9",
        "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9",
}

// isReservedName returns true, if path is Windows reserved name.
// See reservedNames for the full list.
func isReservedName(path string) bool {
        if len(path) == 0 {
                return false
        }
        for _, reserved := range reservedNames {
                if strings.EqualFold(path, reserved) {
                        return true
                }
        }
        return false
}
```

It also seems to be wrong as it doesn't consider `aux.c` to be a reserved name, whereas Windows certainly does (I can tell you this from bitter personal experience porting a large C++ code base to Windows!).

ncw (commented):
I think this looks OK :-)

The idea of moving the concern of "is this a special config file name" out of configfile and into config seems like a good idea.

I don't like the coupling between configfile and config - reading the config file location from config seems the wrong way round to me - config should be telling configfile about the location when it changes.

Perhaps in another PR!

The one disadvantage with this is that the diff looks too big to put into a point release.

I put a few comments inline for you to look at.

albertony:
Thanks, I will update according to your input!

> I don't like the coupling between configfile and config - reading the config file location from config seems the wrong way round to me - config should be telling configfile about the location when it changes.

Yes, that was my first though as well, but didn't dig more into it..

> The one disadvantage with this is that the diff looks too big to put into a point release.

I wouldn't mind writing a smaller patch to commit into a point release branch, if that is something we would want?

fs/config/config.go
albertony: > Ewww! You could put this into lib/file as IsReserved or something like that?

Yes, that will be better.

> I can tell you this from bitter personal experience porting a large C++ code base to Windows!

My condolences!

cmd/version/version_test.go
```diff
@@ -27,11 +27,11 @@ func TestVersionWorksWithoutAccessibleConfigFile(t *testing.T) {
 	// re-wire
 	oldOsStdout := os.Stdout
 	oldConfigPath := config.GetConfigPath()
-	_ = config.SetConfigPath(path)
+	assert.NoError(t, config.SetConfigPath(path))
```
ncw, line 30: These changes should have gone in the previous commit unless you are planning to squash them?

ncw (commented):
I think that is looking very good!

There has been a slight mixup with the commits - see inline for details.

Just in case you are not a git master, I'd fix it like this

`git reset HEAD~` to pop the last set of changes

`git add cmd/version/version_test.go fs/config/config_test.go fs/config/configfile/configfile_test.go fs/config/crypt_test.go fs/config/ui_test.go`

`git commit --amend` to add the changes to the previous commit

`git commit -a -c ORIG_HEAD` to commit the remains of the changes with the original commit message.

cmd/version/version_test.go
albertony: Ai... Probably was a bit too quick on the trigger with "git commit --amend --no-edit". Will fix!

albertony:
> There has been a slight mixup with the commits - see inline for details.

Fixed!

> Just in case you are not a git master, I'd fix it like this

Not quite there yet, so your example definitely saved me some time - thanks!

ncw (approved):
LGTM - please merge :-)

Thank you !

albertony:
@ncw: Just did... But there is a build error I initially interpreted as "random" issue, but now starting to wonder..

It runs:
> RCLONE_CONFIG="/notfound" go test -tags "cmount" ./...

And what I now noticed is this:
> 2021/04/11 11:43:30 NOTICE: Config file "C:\\Program Files\\Git\\notfound" not found - using defaults

So its interpreting /notfound as a config file path, and even relative??, and not as a memory-config..

It fails in the windows-386, but same test runs fine in windows_amd64..

I don't understand, do you?

ncw:
Good debugging! Yes the string "/notfound" was just designed so that there was never a config file there.

We could change the use of "/notfound" to "" within rclone which is probably more sensible and would hopefully work with git-bash.

```
$ git grep /notfound # edited
Makefile:       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) ./...
Makefile:       RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -cpu=2 -race ./...
fstest/fstests/fstests.go:      if strings.HasSuffix(os.Getenv("RCLONE_CONFIG"), "/notfound") && *fstest.RemoteName == "" {
```

albertony:
Yes, made a go at that here: https://github.com/rclone/rclone/pull/5235
Only changed the Makefile first, so I thought I was wrong (again), but then just recently found that check in fstest and have updated that as well, so perhaps it works better.
(To be continued in the PR)

albertony:
Discovered after this PR was merged that there is one "special" issue still, resulting from a combinations of limitations:
- /notfound does not work from Git/MSYS2 bash because it convertes the "/" into its internal filesystem root being the Git/ MSYS2 installation directory, and therefore rclone gets an absolute path which is not at the root and does not catch this as the memory-only identifier.
- Null device name is different on Windows and Linux.
- Environment variable cannot be set to empty string on Windows, there is no difference between unset variable and empty string variable.

The result of all this is that one cannot enable memory only config by setting environment variable RCLONE_CONFIG from a command that are executed on both Linux and on Windows via Git/MSYS2 bash... Which is exactly what I tried to do in the above mentioned PR, changing rclone's makefile for running the quick tests into using memory only config.

PR #5226, opened 2021-04-09
Change default location of config file to %AppData% on Windows

### LINKED ISSUE #4667: Move the config location on Windows to a proper directory that follows windows standards
When looking for most Windows application config files they are normally in the C:\Users\{User}\AppData\Local\  or \Roaming folder. rclone chose to use the C:\Users\{User}\.config\ folder. It's not a huge deal but it does not really help windows user that are used to this format to find configs for an app. So i was just curious if this is for a good reason why it is here and does not follow the normal windows convention? Maybe in the future we can think about moving it there?

albertony:

Use %AppData% as primary default for configuration file on Windows, which is more in line with Windows standards, while existing default of using home directory is more Unix standards - though that made rclone more consistent accross different OS.

Main changes from existing (something for release notes?)
- Any existing configuration file should be picked up in exactly the same way as before, so no changes for **existing** configs.
  - Well.. with one possible exception: If there happens to be a file `%APPDATA%/rclone/rclone.conf`, it would not have been picked up automatically before, but could be used by explicitely referring to it with `$RCLONE_CONFIG` or `--config`, but will now be picked up automatically. That *could* lead to surprises on upgrade, but not very likely.
- Windows:
  - Default path when creating new configuration file is now `%APPDATA%/rclone/rclone.conf`.
  - Previously, configuration file would typically be created in `%USERPROFILE%\.config\rclone\rclone.conf`
    - One could set environment variable `XDG_CONFIG_HOME` and it would instead be created in `%XDG_CONFIG_HOME%/rclone/rclone.conf`, this is no longer the case (see next point).
  - `%USERPROFILE%\.config\rclone\rclone.conf` will still be used in the unlikely event of `%AppData%` not being defined, but in any case a variable `%XDG_CONFIG_HOME%` will no longer be considered when rclone is creating new config (but will still be considered if an existing configuration file is found there, to avoid breaking existing setups).

Main rules (see also changed documentation):
- If custom location configured with `$RCLONE_CONFIG` or `--config`, it will be used regardless.
- Else, search for existing file in prioritized order:
  1. `<rclone_exe_dir>/rclone.conf`
  2. `%APPDATA%/rclone/rclone.conf` (Windows only)
  3. `$XDG_CONFIG_HOME/rclone/rclone.conf`
  4. `~/.config/rclone/rclone.conf`
  5. `~/.rclone.conf`
- Else, create new in default location:
  - Windows: Path 2 above.
  - Unix/non-windows: If XDG_CONFIG_HOME defined path 3, else path 4.

Note: `XDG_CONFIG_HOME` is kind of a "bastard" on Windows: This is not a common variable to set in Windows, but previous versions of rclone did consider it on all platforms, preferring it over `~/.config`. It will still be considered after this PR, for backwards compatible reasons, but only for reading existing configuration file found there.

#### Testing and documenting behavior

Tests performed with command `rclone config file`, no existing configuration file, verify what is chosen as default config location.
Last performed with beta version: https://github.com/albertony/rclone/releases/tag/v1.56.0-appdata05

Note regarding permission testing:
- Rclone will first look for a configuration file, and if not found pick a configuration directory to use for a new file.
- When picking a new configuration directory rclone will check permissions: It checks that it can create the config directory, e.g. `~/.config/rclone`, but does not verify permission to actually create a config file in it. This means when HOME is defined, but rclone has no permission to write to it, then rclone will find it cannot create directory `~/.config/rclone` and fallback to `~/.rclone.conf`, and "happily" assume that is a valid path - even though it will not actually have permission to create such a file! Attempting to create it at a later point in time (`rclone config touch`) will fail with:
  > ERROR : Failed to save config after 10 tries: Failed to create temp file for new config: open /tmp/xxx_home/.rclone.conf693553371: permission denied
- If the configuration directory already exists, e.g. `~/.config/rclone`, then permissions will not be checked. So if rclone does not (no longer) have write access here, it will still pick this one instead of using fallback - unlike what it would do if `~/.config/rclone` did not exist and rclone failed to create it.
- After the configuration file path has been found (either an existing, or path to where to store a new one), rclone will try to read it. This is also done in a simple `rclone config file` test.
  - If the file exists, a debug message is printed:
    > Using config file from %q
  - If user has permission to directory but file is not there, rclone will report notice:
    > Config file %q not found - using defaults
  - If user does not have traverse permissions to the directory, rclone will fail with fatal:
    > Failed to load config file %q: %v

Linux (arch WSL):
- Normal user (with a home)
  - XDG_CONFIG_HOME defined:
    - Permission: `$XDG_CONFIG_HOME/rclone/.rclone.conf`
    - No write permission: `~/.rclone.conf`
       - Note: Does not fallback to  ~/.config/rclone/rclone.conf!
  - XDG_CONFIG_HOME **not** defined:
    - Permission to ~/.config: `~/.config/rclone/rclone.conf`
    - No write permission to ~/.config: `~/.rclone.conf`
  - No write or traverse permission to \~: `Failed to load config file "~/.rclone.conf": permission denied`
     - Note: The rclone config file command just returns this error message. It found that it could not create the config directory and that ~/.rclone.conf is the path to use, but then attempting to load it fails with "permission denied" instead of "not found".
     - Note: Does not fallback to .rclone.conf
  - HOME variable modified: `$HOME/.config/rclone/rclone.conf`
- User with no home (empty in passwd file)
  - No home-related environment: `.rclone.conf`
      ERROR : Couldn't find configuration directory nor home directory
      ERROR : Defaulting to storing config in current directory.
      ERROR : Use --config flag to workaround.
  - HOME only defined:
    - Permission: `$HOME/rclone/.rclone.conf`
    - No write permission: `$HOME/.rclone.conf`
  - HOME and XDG_CONFIG_HOME defined:
    - Permission to XDG_CONFIG_HOME: `$XDG_CONFIG_HOME/rclone/.rclone.conf`
    - No write permission to XDG_CONFIG_HOME: `$HOME/.rclone.conf`
  - XDG_CONFIG_HOME only defined:
    - Permission: `$XDG_CONFIG_HOME/rclone/.rclone.conf`
    - No write permission: `.rclone.conf`
      ERROR : Couldn't find home directory nor create configuration directory: mkdir $XDG_CONFIG_HOME/rclone: permission denied
      ERROR : Defaulting to storing config in current directory.
      ERROR : Use --config flag to workaround.

Windows 10:
- Normal user: `%AppData%\rclone\rclone.conf`
- No APPDATA variable, but HOME: `%HOME%\rclone\rclone.conf`
- No APPDATA, nor HOME: `%USERPROFILE%\.config\rclone\rclone.conf`
- No APPDATA, nor HOME, nor USERPROFILE: `%HOMEDRIVE%%HOMEPATH%\.config\rclone\rclone.conf`
- No APPDATA, nor HOME, nor USERPROFILE, nor HOMEDRIVE: `.rclone.conf`
  ERROR : Couldn't find configuration directory nor home directory: HOMEDRIVE, HOMEPATH, or USERPROFILE are blank
  ERROR : Defaulting to storing config in current directory.
  ERROR : Use --config flag to workaround.
- XDG_CONFIG_HOME defined, no APPDATA, HOME, USERPROFILE, nor HOMEDRIVE: `.rclone.conf`
  ERROR : Couldn't find configuration directory nor home directory: HOMEDRIVE, HOMEPATH, or USERPROFILE are blank
  ERROR : Defaulting to storing config in current directory.
  ERROR : Use --config flag to workaround.
- APPDATA set to path with no access, no HOME, USERPROFILE, nor HOMEDRIVE: `.rclone.conf`
  ERROR : Couldn't find home directory nor create configuration directory: mkdir C:\NoAccess\rclone: Access is denied.
  ERROR : Defaulting to storing config in current directory.
  ERROR : Use --config flag to workaround.
- APPDATA set to path with no access, HOME set to path with access: `%HOME%\.rclone.conf`

Fixes #4667

Related:
#2515
#3245
#5223

#### Checklist

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -126,59 +126,127 @@ func init() {
 	configPath = makeConfigPath()
 }
 
+// Join directory with filename, and check if exists
+func findFile(dir string, name string) string {
+	path := filepath.Join(dir, name)
+	if _, err := os.Stat(path); err != nil {
+		return ""
+	}
+	return path
+}
+
+// Find current user's home directory
+func findHomeDir() (string, error) {
+	path, err := homedir.Dir()
+	if err != nil {
+		fs.Debugf(nil, "Home directory lookup failed and cannot be used as configuration location: %v", err)
+	} else if path == "" {
+		// On Unix homedir return success but empty string for user with empty home configured in passwd file
+		fs.Debugf(nil, "Home directory not defined and cannot be used as configuration location")
+	}
+	return path, err
+}
+
+// Find rclone executable directory and look for existing rclone.conf there
+// (<rclone_exe_dir>/rclone.conf)
+func findLocalConfig() (configDir string, configFile string) {
+	if exePath, err := os.Executable(); err == nil {
+		configDir = filepath.Dir(exePath)
+		configFile = findFile(configDir, configFileName)
+	}
+	return
+}
+
+// Get path to Windows AppData config subdirectory for rclone and look for existing rclone.conf there
+// ($AppData/rclone/rclone.conf)
+func findAppDataConfig() (configDir string, configFile string) {
+	if appDataDir := os.Getenv("APPDATA"); appDataDir != "" {
+		configDir = filepath.Join(appDataDir, "rclone")
+		configFile = findFile(configDir, configFileName)
+	} else {
+		fs.Debugf(nil, "Environment variable APPDATA is not defined and cannot be used as configuration location")
+	}
+	return
+}
+
+// Get path to XDG config subdirectory for rclone and look for existing rclone.conf there
+// (see XDG Base Directory specification: https://specifications.freedesktop.org/basedir-spec/latest/).
+// ($XDG_CONFIG_HOME\rclone\rclone.conf)
+func findXDGConfig() (configDir string, configFile string) {
+	if xdgConfigDir := os.Getenv("XDG_CONFIG_HOME"); xdgConfigDir != "" {
+		configDir = filepath.Join(xdgConfigDir, "rclone")
+		configFile = findFile(configDir, configFileName)
+	}
+	return
+}
+
+// Get path to .config subdirectory for rclone and look for existing rclone.conf there
+// (~/.config/rclone/rclone.conf)
+func findDotConfigConfig(home string) (configDir string, configFile string) {
+	if home != "" {
+		configDir = filepath.Join(home, ".config", "rclone")
+		configFile = findFile(configDir, configFileName)
+	}
+	return
+}
+
+// Look for existing .rclone.conf (legacy hidden filename) in root of user's home directory
+// (~/.rclone.conf)
+func findOldHomeConfig(home string) (configDir string, configFile string) {
+	if home != "" {
+		configDir = home
+		configFile = findFile(home, hiddenConfigFileName)
+	}
+	return
+}
+
 // Return the path to the configuration file
 func makeConfigPath() string {
-	// Use rclone.conf from rclone executable directory if already existing
-	exe, err := os.Executable()
-	if err == nil {
-		exedir := filepath.Dir(exe)
-		cfgpath := filepath.Join(exedir, configFileName)
-		_, err := os.Stat(cfgpath)
-		if err == nil {
-			return cfgpath
+	// Look for existing rclone.conf in prioritized list of known locations
+	// Also get configuration directory to use for new config file when no existing is found.
+	var (
+		configFile        string
+		configDir         string
+		primaryConfigDir  string
+		fallbackConfigDir string
+	)
+	// <rclone_exe_dir>/rclone.conf
+	if _, configFile = findLocalConfig(); configFile != "" {
+		return configFile
+	}
+	// Windows: $AppData/rclone/rclone.conf
+	// This is also the default location for new config when no existing is found
+	if runtime.GOOS == "windows" {
+		if primaryConfigDir, configFile = findAppDataConfig(); configFile != "" {
+			return configFile
 		}
 	}
-
-	// Find user's home directory
-	homeDir, err := homedir.Dir()
-
-	// Find user's configuration directory.
-	// Prefer XDG config path, with fallback to $HOME/.config.
-	// See XDG Base Directory specification
-	// https://specifications.freedesktop.org/basedir-spec/latest/),
-	xdgdir := os.Getenv("XDG_CONFIG_HOME")
-	var cfgdir string
-	if xdgdir != "" {
-		// User's configuration directory for rclone is $XDG_CONFIG_HOME/rclone
-		cfgdir = filepath.Join(xdgdir, "rclone")
-	} else if homeDir != "" {
-		// User's configuration directory for rclone is $HOME/.config/rclone
-		cfgdir = filepath.Join(homeDir, ".config", "rclone")
+	// $XDG_CONFIG_HOME/rclone/rclone.conf
+	// Also looking for this on Windows, for backwards compatibility reasons.
+	if configDir, configFile = findXDGConfig(); configFile != "" {
+		return configFile
+	}
+	if runtime.GOOS != "windows" {
+		// On Unix this is also the default location for new config when no existing is found
+		primaryConfigDir = configDir
+	}
+	// ~/.config/rclone/rclone.conf
+	// This is also the fallback location for new config
+	// (when $AppData on Windows and $XDG_CONFIG_HOME on Unix is not defined)
+	homeDir, homeDirErr := findHomeDir()
+	if fallbackConfigDir, configFile = findDotConfigConfig(homeDir); configFile != "" {
+		return configFile
+	}
+	// ~/.rclone.conf
+	if _, configFile = findOldHomeConfig(homeDir); configFile != "" {
+		return configFile
 	}
 
-	// Use rclone.conf from user's configuration directory if already existing
-	var cfgpath string
-	if cfgdir != "" {
-		cfgpath = filepath.Join(cfgdir, configFileName)
-		_, err := os.Stat(cfgpath)
-		if err == nil {
-			return cfgpath
-		}
-	}
-
-	// Use .rclone.conf from user's home directory if already existing
-	var homeconf string
-	if homeDir != "" {
-		homeconf = filepath.Join(homeDir, hiddenConfigFileName)
-		_, err := os.Stat(homeconf)
-		if err == nil {
-			return homeconf
-		}
-	}
-
-	// Check to see if user supplied a --config variable or environment
-	// variable.  We can't use pflag for this because it isn't initialised
-	// yet so we search the command line manually.
+	// No existing config file found, prepare proper default for a new one.
+	// But first check if if user supplied a --config variable or environment
+	// variable, since then we skip actually trying to create the default
+	// and report any errors related to it (we can't use pflag for this because
+	// it isn't initialised yet so we search the command line manually).
 	_, configSupplied := os.LookupEnv("RCLONE_CONFIG")
 	if !configSupplied {
 		for _, item := range os.Args {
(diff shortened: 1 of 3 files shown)


albertony:
@MrCaspan, @ivandeex: If you fancy some testing, that would be great. I'm quite confident it just adds %AppData% as new default on Windows, without changing rest of the existing behavior, but I really want to avoid regression on this part.

There is a beta built from my pr branch here: https://github.com/albertony/rclone/releases/tag/v1.56.0-appdata02

If you want to have a read-through of the changes in docs, that would also be awesome: https://github.com/rclone/rclone/blob/5718fed527f291bbbb3b12a2aadc73009fc2b771/docs/content/docs.md#--configconfig_file

Any feedback are welcome! 🐑 🐺

albertony:
Force pushed an amend. Only some small improvements to error message.

New beta: https://github.com/albertony/rclone/releases/tag/v1.56.0-appdata03

ncw (commented):
Well done for attacking a difficult problem :-)

I think we should re-organise the code according to the docs...

I like your 5 numbered points `1. rclone.conf`...

Can we factor each of those points into a new function something like this?

```go

var config string
if config = findLocalRcloneConf(); config != "" {
    return config
} else if config = findAppdataConf(); config != "" {
    return config
} else if config = findXDGConf():; config != "" {
    return config
} else if config = findDotConfigConf():; config != "" {
    return config
} else if config = findOldHomeDirConf():; config != "" {
    return config
} else {
    // ...
}
```

They may need to take and return more things but you get the idea. We can then read a) the fact that it matches the docs and b) each function will be small enough to understand.

fs/config/config.go
```diff
@@ -122,57 +122,97 @@ func init() {
...
+			if cfgPath, exists := fileExists(cfgDir, configFileName); exists {
+				return cfgPath
+			}
+			// Prefer this as the default location for new config when no existing is found
+			newCfgDir = filepath.Join(appDataDir, "rclone")
+		} else {
+			fs.Debugf(nil, "Environment variable APPDATA is not defined and cannot be used as configuration location")
+		}
+
+		// On Windows: Look for existing rclone.conf in subdirectory rclone of XDG config home,
```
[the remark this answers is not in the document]
albertony: Definitely. It was intentional, saving one string comparison. Premature optimization, you say?? :)

albertony:
> Can we factor each of those points into a new function something like this?

Yes we can! I did consider it ©️ ®️, but then was scared that I would introduce regression.. But, as mentioned, the changes that I did do ended up being somewhat substantial, and then went through all that testing and documentation, so I think its a good idea to do it after all. My version is not as pretty as your if-elseif example, because I have the logic of picking the default dir to use for new config file outside of the helper functions, and also wanted the findHome function outside to avoid doing it more than once (since it may end up checking environment, checking passwd file, shell execute cd and all that). But as always; I appreciate any feedback, good or bad, big or small.. :)

Sneaked in another commit: A new command `rclone config touch`. It simply calls `config.SaveConfig()`. If you think its superfluous, I'll just remove it, and will not go into deep depression because of it! :) I found it useful during testing of these configuration location things. It will ensure the configuration file exists, and it will also test that it is writable, report error if not. It will also ease a portable setup by just download rclone.exe, followed by a quick `rclone config touch --config=rclone.conf` from same location, and then I can trust that further rclone commands will not touch(!) %AppData%. On windows there is no native touch command the same way as on unix (of course easy enough to do, but maybe not for less technical users), but even so the advantage of `rclone config touch` is that I can also test that the default location is writable without thinking what that might be. Again, don't know if it is just 1 person in the world which will have *some* use of it...?

ivandeex:
> If you fancy some testing, that would be great. I'm quite confident it just adds %AppData% as new default on Windows, without changing rest of the existing behavior, but I really want to avoid regression on this part.

I have an idea to try Powershell for the test. Will post here in 1-2 days.

ivandeex:
> On windows there is no native touch command the same way as on unix (of course easy enough to do, but maybe not for less technical users)

rclone touch C:\temp\touch.me

albertony:
> rclone touch C:\temp\touch.me

Ah, of course! Forgot about that.

`rclone config touch` still has the additional convenience (it's primary intention, really) of "implicit" using the current config path..

fs/config/config.go
```diff
@@ -120,59 +120,132 @@ func init() {
...
+	}
+	// Windows: $AppData/rclone/rclone.conf
+	if runtime.GOOS == "windows" {
+		if configDir, configFile = findAppDataConfig(); configFile != "" {
+			return configFile
 		}
+		// On Windows this is the preferred default location for new config when no existing is found
+		defaultConfigDir = configDir
+	}
+	// $XDG_CONFIG_HOME\rclone\rclone.conf
```
ncw, line 221: `\` in path should probably be `/`

fs/config/config.go
```diff
@@ -120,59 +120,132 @@ func init() {
...
+	}
+	// ~/.config/rclone/rclone.conf
+	homeDir, homeDirErr := findHomeDir()
+	if configDir, configFile = findDotConfigConfig(homeDir); configFile != "" {
+		return configFile
+	}
+	if defaultConfigDir == "" {
+		// When $AppData on Windows and $XDG_CONFIG_HOME on Unix does not exist,
+		// fallback to this as the default location when no existing file is found
+		defaultConfigDir = configDir
```
ncw, line 237: What exactly is `configDir` set to here? I think it will have been set to `""` by the call to `findDotConfigConfig`?

So I don't think this is right, but I'm not sure what it should be! The comment needs to say what it should be. `homeDir` maybe?

fs/config/config.go
```diff
@@ -182,32 +255,47 @@ func makeConfigPath() string {
...
-	// and assume rclone.conf can be written there. If user supplied config
-	// then skip creating the directory since it will not be used.
-	if cfgpath != "" {
-		// cfgpath != "" implies cfgdir != ""
+	// If we found a configuration directory to be used for new config during search
+	// above, then create it to be ready for rclone.conf file to be written into it
+	// later, and also as a test of permissions to use fallback if not even able to
+	// create the directory.
+	if defaultConfigDir != "" {
+		configFile := filepath.Join(defaultConfigDir, configFileName)
```
ncw, line 263: That should be just `configFile = filepath...` so you carry on using the same global var.

ncw (commented):
That looks much easier to understand - well done!

I put some small things inline for you to look at.

I think we are almost there :-)

fs/config/config.go
albertony: If homeDir is "" then configDir will be "", and then if defaultConfigDir=="" it will just be set to "" once more.
Otherwise it will be "~/.config/rclone" regardless if rclone.conf was found there or not.
So I think this actually works as intended.

Regardless: I'm trying a little adjustment: Keeping a separate variable primaryConfigDir with the APPDATA path on Windows and XDG path on Unix, and then keeping the ~/.config/rclone path in another variable fallbackConfigDir, and then waiting to choose between them until they are actually needed at the end.

albertony:
Thanks again!

Pushed an update. Should become beta 5 soon: https://github.com/albertony/rclone/releases/tag/v1.56.0-appdata05

Only did a very quick test so far, planning to go through all the testcases documented above to see if they are still the same after these last changes..

ncw (approved):
Looks good to me, now let's merge this :-)

Hopefully testers will find any problems before we release 1.56!

ncw:
> EDIT:
> 
> Retested on Linux, everything ok and same as before.
> 
> EDIT 2:
> 
> Retested on Windows, everything ok and same as before.

Can we automate these tests?

As in make them run via the rclone unit tests on the CI infrastructure?

ivandeex:
> I have an idea to try Powershell for the test. Will post here in 1-2 days.

> Can we automate these tests?
> As in make them run via the rclone unit tests on the CI infrastructure?

Sorry, I was busy elsewhere, did not start yet. If you can wait, I will write the test in golang.

albertony:
> Sorry, I was busy elsewhere, did not start yet. If you can wait, I will write the test in golang.

Sure, that would be great!

ncw:
Do you want to merge this now and put tests into a new PR?

albertony:
Perhaps we should just do it that way, @ivandeex ?

ivandeex:
> Do you want to merge this now and put tests into a new PR?

I'll merge this now to let the next config PR rebase on top. Will submit test as separate PR. Let's integrate by parts :)

ivandeex:
I think merging as 2 commits is fine: 1st one updates config search, 2nd one adds a subcommand.

PR #5230, opened 2021-04-11
Silence the progress output with curl requests

ashokgelal:
Having a progress seems to break some automated scripts and there isn't a way to pass some flags to these curl requests to disable them. I don't know if showing the progress is any helpful to be honest.


ncw (approved):
That looks sensible to me - thank you :-)

I'll merge it now.

PR #5235, opened 2021-04-12
Platform independent memory-only config option to be used in build

albertony:

Two things:
1. Run tests with memory-only configuration on all platforms (dependent on the next).
2. Continuation of  https://github.com/rclone/rclone/pull/5223:

    Case 1. --config /dev/null on Linux
    Case 2. --config NUL on Windows
    Case 3. --config ""
    Case 4. --config notfound

   Extending case 4 to any path where `notfound` is the file/base name.

##### Background for part 1

Previously used RCLONE_CONFIG="/notfound", assuming it was path to a non-existant file. Then at some point in time (quite recently) we started detecting this, as well as empty string and os null device, as an indicator of memory-only config. On most platforms the tests therefore automatically started to run with memory-only config. But on Windows the tests are run from Git (MSYS2) bash, which convertes the "/" into its internal filesystem root being the Git (MSYS2) installation directory. This means on Windows, the tests are now run with config path `C:/Program Files/Git/notfound`. No big deal, but its a discrepancy between tests run on Windows and Unix platforms, and also it is a bit confusing since the configured "reserved name" /notfound now is easily mistaken for a memory-only config.

##### Background for part 2

While doing part 1 another problem came up, due to the facts:

- /notfound does not work from Git/MSYS2 bash because it convertes the "/" into its internal filesystem root being the Git/ MSYS2 installation directory, and therefore rclone gets an absolute path which is not at the root and does not catch this as the memory-only identifier.
- Null device name is different on Windows and Linux.
- Environment variable cannot be set to empty string on Windows, there is no difference between unset variable and empty string variable.

The result of all this is that one cannot enable memory only config by setting environment variable RCLONE_CONFIG from a command that are executed on both Linux and on Windows via Git/MSYS2 bash... Which is exactly what I tried to do in the above mentioned PR, changing rclone's makefile for running the quick tests into using memory only config.

Fixed by checking any path where `notfound` is the file/base name, not just absolute path from root.

https://github.com/rclone/rclone/pull/5223#issuecomment-817974404

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -686,8 +686,8 @@ to run in "portable" mode by downloading rclone executable to a
 writable directory and then create an empty file `rclone.conf` in the
 same directory.
 
-If the location is set to empty string `""` or the special value
-`/notfound`, or the os null device represented by value `NUL` on
+If the location is set to empty string `""` or path to a file
+with name `notfound`, or the os null device represented by value `NUL` on
 Windows and `/dev/null` on Unix systems, then rclone will keep the
 config file in memory only.
 
@@ -1920,11 +1920,12 @@ Nevertheless, rclone will read any configuration file found
 according to the rules described [above](https://rclone.org/docs/#config-config-file).
 If an encrypted configuration file is found, this means you will be prompted for
 password (unless using `--password-command`). To avoid this, you can bypass
-the loading of the configuration file by overriding the location with an empty
-string `""` or the special value `/notfound`, or the os null device represented
-by value `NUL` on Windows and `/dev/null` on Unix systems (before rclone
-version 1.55 only this null device alternative was supported).
-E.g. `rclone --config="" genautocomplete bash`.
+the loading of the default configuration file by overriding the location,
+e.g. with one of the documented special values for memory-only configuration:
+
+```
+rclone genautocomplete bash --config=""
+```
 
 Developer options
 -----------------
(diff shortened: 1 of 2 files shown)


ncw:
Not sure I like the "0" special case very much.

How about if we detected the config string if it ends with `/notfound` or maybe `\notfound` also on Windows?

This should work around the git windows thing.

We appear to be doing this already in

    fstest/fstests/fstests.go:      if strings.HasSuffix(os.Getenv("RCLONE_CONFIG"), "/notfound") && *fstest.RemoteName == "" {

albertony:
> Not sure I like the "0" special case very much.

I like it better than /\notfound, but ok! 😃 

> How about if we detected the config string if it ends with `/notfound` or maybe `\notfound` also on Windows?

I can just do that, sure. And yes, I think we should treat both variants `/notfound` and `\notfound` as equivalent on Windows.

ncw (approved):
This looks fine now - please merge :-)

The whole `notfound` thing is a historical accident we should probably remove at some point but I think your fix looks good - thank.

albertony:
Great, thanks!

I must admit I like the `0` value a *tiny* bit less after thinking some more.. 😊 But I'm not saying I like it less than `notfound` yet... 😆 

> is a historical accident

I get it! 👍 

> please merge

Will do!

PR #5237, opened 2021-04-12
Introduce global flag --human-readable

albertony:
*UPDATE 24 Aug 2021, from discussions (see below):*
- *Refactored into a bit more conservative approach, documented in [this comment](https://github.com/rclone/rclone/pull/5237#issuecomment-834413850).*
- *Short flag `-H` removed, postponed since it was raised concern that means transfer hardlinks in rsync*
- *Unit format changed from `MiByte` to `MiB` (which is more of an continuation of [#5087](https://github.com/rclone/rclone/pull/5087))*

Introduce global flag `--human-readable`/`-H` to switch output between raw values and values with suffix.

NOTE:
- Was depending on PR https://github.com/rclone/rclone/pull/5087 (this PR was split out from that), it has now been merged and this PR rebased accordingly.

DONE:
- Added option `--human-readable`/`-H` for `ls` and friends.
   - Prints sizes with suffix, e.g. `9.762Ki`, instead of raw byte count, `9996`.
   - Syntax for linux commands rsync, ls, df, dh etc all have `-h, --human-readable`, so I also chose `--human-readable`. The short form `-h` is already taken by `--help` in rclone, so picked `-H`. `df` command takes `-h` for `--human-readable` and `-H` as short form for `--si` (mean human-readable with decimal prefix), so this will be inconsistent with rclone, but other commands have no shortform for `--si` and `--human-readable` is so long to type so it is best to have some short form - so ended up with `-H`..
   - Uses decimal suffix for human-readable counts in `lsd` (1k means 1000 files), same as `ncdu`.
   - Fixes issue #1890
- Also consider option `--human-readable` for `ncdu` command.
  - Previously always human-readable format.
  - Now supports raw format, switching between them interactively with key `u`.
  - Note: To consider global flag `--human-readable` the default had to be changed, so it will now default to raw numbers, and must either explicitely start it with option `--human-readable` or use enter key `u` after startup to get back old to what was default before.
    - TODO: Stick with human readable format as the default as before? And just accept that ncdu does not consider `--human-readable` flag, but keep interactive key `u` as the (only) way of switching?
- Also consider option `--human-readable` for `size` command.
  - Previously would output string with both raw number and human-readable, now either or.
  - Uses decimal suffix for human-readable counts, same as `lsd` and `ncdu`.
- Also consider option `--human-readable` for `about` command.
  - Previously did this by default, and had option `--full` to output raw numbers.
  - Now the opposite: Raw numbers by default, unless `--human-readable`.
  - Note: The `--full` option was removed!
  - Also fixed that count `Objects: Total number of objects in the storage.` uses decimal suffix, used binary suffix before.
- Also consider option `--human-readable` for `tree` command.
  - It did already have a custom option `--human`. Removed this in favor of the global `--human-readable`/`-H` option for consistency.
- Copy, sync, stats etc always use human-readable sizes, like before, without considering the option!

TODO:
   - Uses the `Mi` format in tabular output, where space is limited,such as results from `ls` with `--human-readable`, but what about more "non-tabular" output, from `size`, `about` etc?
  - Show human-readable counts with suffix (not just sizes)?
    - When using single-letter suffix for size and count it looks consistent:
      ```
      Total objects: 2.171k
      Total bytes:   176.584Mi
      ```
      But what if deciding (previous decision) to use longer suffix for size, should we still use the single-letter suffix for count, as there is no unit `kObject`, or does that look inconsistent?
      ```
      Total objects: 2.171k
      Total bytes:   176.584 MiByte
      ```
      See this and more examples below.

##### Breaking changes
- The `about` command:
   - Option `--full` no longer exist.
   - Opposite default: Raw numbers by default, must now use global option `--human-readable`/`-H` to get output similar to previous default.
- The `tree` command:
   - Option `--human` no longer exist, replaced by global option  `--human-readable`/`-H`.
- The `ncdu` command:
   - Opposite default: Raw numbers by default, must now use global option `--human-readable`/`-H` or interactive key `u` to get output similar to previous default.
 
##### Examples

- `ls` with `--human-readable` (new option)
  ```
  336.125Ki langs.model.xml
   34.827Ki license.txt
    3.439Mi notepad++.exe
        955 shortcuts.xml
  ```
- `size`:
  - Before:
    ```
    Total objects: 2171
    Total size: 176.584 MBytes (185161745 Bytes)
    ```
  - Now
    - Default:
      ```
      Total objects: 2171
      Total bytes:   185161745
      ```
    - With `--human-readable`
      ```
      Total objects: 2.171k
      Total bytes:   176.584Mi
      ```
  - Possible alternatives
    - Longer suffix with `--human-readable`:
      ```
      Total objects: 2.171k
      Total bytes:   176.584 MiB
      ```
    - Even longer suffix with `--human-readable` for size:
      ```
      Total objects: 2.171k
      Total bytes:   176.584 MiByte
      ```
    - Longer suffix with `--human-readable` but not considered for counts:
      ```
      Total objects: 2171
      Total bytes:   176.584 MiB
      ```
    - Even longer suffix with `--human-readable` for size:
      ```
      Total objects: 2171
      Total bytes:   176.584 MiByte
      ```
- `about`:
  - Before
    - Default:
      ```
      Total:   952.131G
      Used:    438.671G
      Free:    513.460G
      ```
    - With `--full`:
      ```
      Total:   1022342836224
      Used:    471019343872
      Free:    551323492352
      ```
  - Now
    - Default:
      ```
      Total:   1022342836224
      Used:    471019180032
      Free:    551323656192
      ```
    - With `--human-readable`:
      ```
      Total:   952.131Gi
      Used:    438.671Gi
      Free:    513.460Gi
      ```
  - Possible alternatives:
    - Longer suffix with `--human-readable`:
      ```
      Total:   952.131 GiB
      Used:    438.671 GiB
      Free:    513.460 GiB
      ```
    - Even longer suffix With `--human-readable`:
      ```
      Total:   952.131 GiByte
      Used:    438.671 GiByte
      Free:    513.460 GiByte
      ```
    
Fixes https://github.com/rclone/rclone/issues/1890
Highly dependendent on https://github.com/rclone/rclone/pull/5087

#### Checklist


buengese:
>Note: To consider global flag --human-readable the default had to be changed, so it will now default to raw numbers, and must either explicitely start it with option --human-readable or use enter key u after startup to get back old to what was default before.

I'm not entirely sure about this. I can't really think of a situation where I wouldn't want human-readable. Not just with ncdu but in general. The only exception here would be when I'm running rclone as part of script, but then I'd be using `--rc` or at least json output.

albertony:
> I'm not entirely sure about this. I can't really think of a situation where I wouldn't want human-readable. Not just with ncdu but in general.

One point of this is consistency: Currently `ls` will show raw values only, `size` a mix, `about` defaults to human-readable but takes option `--full` to show raw, `tree` defaults to raw but takes option ` --human`, and then `ncdu` only human-readable. Its a "good" mix.

Regarding ncdu, although the default with this are switched, I feel starting it with a simple `-U` argument or pressing the `u` key after launch is not *that* inconvenient even if you want the human readable every time.. (or you can set `RCLONE_HUMAN_READABLE=true` permanently in your user profile.... 💣 😆 )

Sure, I understand it is a bit controversial.

ncw:
> * Dependent on PR #5087, and currently includes duplicated commits (first 4) from that PR (this PR was split out from that)!

You'll want to rebase this one before review...

> * Added option `--human-readable`/`-H` for `ls` and friends.
>   
>   * Prints sizes with suffix, e.g. `9.762Ki`, instead of raw byte count, `9996`.
>   * Syntax for linux commands rsync, ls, df, dh etc all have `-h, --human-readable`, so I also chose `--human-readable`. The short form `-h` is already taken by `--help` in rclone, so picked `-H`. `df` command takes `-h` for `--human-readable` and `-H` as short form for `--si` (mean human-readable with decimal prefix), so this will be inconsistent with rclone, but other commands have no shortform for `--si` and `--human-readable` is so long to type so it is best to have some short form - so ended up with `-H`..

Is this a global flag? If so we need to be very careful with adding them to make sure no sub commands have a `-H` flag. Note also that `-H` means transfer hardlinks in `rsync` which we are probably unlikely to every use, but just noting using `-H` closes the door to that bit of rsync compatibility.

>   * Uses decimal suffix for human-readable counts in `lsd` (1k means 1000 files), same as `ncdu`.

OK

>   * Fixes issue #1890
> * Also consider option `--human-readable` for `ncdu` command.
>   
>   * Previously always human-readable format.
>   * Now supports raw format, switching between them interactively with key `u`.
>   * Note: To consider global flag `--human-readable` the default had to be changed, so it will now default to raw numbers, and must either explicitely start it with option `--human-readable` or use enter key `u` after startup to get back old to what was default before.

I'd prefer for ncdu usage to remain human readable by default.

>     
>     * TODO: Stick with human readable format as the default as before? And just accept that ncdu does not consider `--human-readable` flag, but keep interactive key `u` as the (only) way of switching?

That seems reasonable to me.

> * Also consider option `--human-readable` for `size` command.

It is already isn't it?

```
$ rclone size ~/go/src/github.com/rclone/
Total objects: 4167
Total size: 1.062 GBytes (1140069956 Bytes)
```
  
>   * Previously would output string with both raw number and human-readable, now either or.

I think that is probably a backwards step - showing both is convenient for the user.

>   * Uses decimal suffix for human-readable counts, same as `lsd` and `ncdu`.
> * Also consider option `--human-readable` for `about` command.
>   
>   * Previously did this by default, and had option `--full` to output raw numbers.
>   * Now the opposite: Raw numbers by default, unless `--human-readable`.
>   * Note: The `--full` option was removed!

I don't like the backwards incompatibility here - I think we should probably leave `about` alone.

>   * Also fixed that count `Objects: Total number of objects in the storage.` uses decimal suffix, used binary suffix before.
> * Also consider option `--human-readable` for `tree` command.
>   
>   * It did already have a custom option `--human`. Removed this in favor of the global `--human-readable`/`-H` option for consistency.

Backwards incompat but probably not a big deal!

> * Copy, sync, stats etc always use human-readable sizes, like before, without considering the option!

:-)
 
> TODO:
> 
> * Uses the `Mi` format in tabular output, where space is limited,such as results from `ls` with `--human-readable`, but what about more "non-tabular" output, from `size`, `about` etc?
> * Show human-readable counts with suffix (not just sizes)?
>   
>   * When using single-letter suffix for size and count it looks consistent:
>     ```
>     Total objects: 2.171k
>     Total bytes:   176.584Mi
>     ```
>     
>     
>     But what if deciding (previous decision) to use longer suffix for size, should we still use the single-letter suffix for count, as there is no unit `kObject`, or does that look inconsistent?
>     ```
>     Total objects: 2.171k
>     Total bytes:   176.584 MiByte
>     ```
>     
>     
>     See this and more examples below.

I think the second example looks nice.

> ##### Breaking changes
> * The `about` command:
>   
>   * Option `--full` no longer exist.
>   * Opposite default: Raw numbers by default, must now use global option `--human-readable`/`-H` to get output similar to previous default.

Not fond of this - see above!

> * The `tree` command:
>   
>   * Option `--human` no longer exist, replaced by global option  `--human-readable`/`-H`.

Seems OK.

> * The `ncdu` command:
>   
>   * Opposite default: Raw numbers by default, must now use global option `--human-readable`/`-H` or interactive key `u` to get output similar to previous default.

I don't like that change. ncdu is really useful because it gives you the human readable stuff by default.

> ##### Examples
> * `ls` with `--human-readable` (new option)
>   ```
>   336.125Ki langs.model.xml
>    34.827Ki license.txt
>     3.439Mi notepad++.exe
>         955 shortcuts.xml
>   ```

:+1:

> * `size`:
>   
>   * Before:
>     ```
>     Total objects: 2171
>     Total size: 176.584 MBytes (185161745 Bytes)
>     ```
>   * Now
>     
>     * Default:
>       ```
>       Total objects: 2171
>       Total bytes:   185161745
>       ```
>     * With `--human-readable`
>       ```
>       Total objects: 2.171k
>       Total bytes:   176.584Mi
>       ```

I think this is a step back in usability - we should just show both like we do now.

> * `about`:
>   
>   * Before
>     
>     * Default:
>       ```
>       Total:   952.131G
>       Used:    438.671G
>       Free:    513.460G
>       ```
>     * With `--full`:
>       ```
>       Total:   1022342836224
>       Used:    471019343872
>       Free:    551323492352
>       ```
>   * Now
>     
>     * Default:
>       ```
>       Total:   1022342836224
>       Used:    471019180032
>       Free:    551323656192
>       ```
>     * With `--human-readable`:
>       ```
>       Total:   952.131Gi
>       Used:    438.671Gi
>       Free:    513.460Gi
>       ```
>   * Possible alternatives:
>     
>     * Longer suffix with `--human-readable`:
>       ```
>       Total:   952.131 GiB
>       Used:    438.671 GiB
>       Free:    513.460 GiB
>       ```
>     * Even longer suffix With `--human-readable`:
>       ```
>       Total:   952.131 GiByte
>       Used:    438.671 GiByte
>       Free:    513.460 GiByte
>       ```

I don't like this change - about is useful because it shows the human readable stuff by default. And we definitely can't take away the `--raw` flag as likely people are scripting that.

In summary - positive on the general idea, negative on some specifics!

PS Do I hear a `--si` flag coming? ;-)

zanna-37:
Another idea would be to have both version in every output, like this:
```
Total size: 1.062 GBytes (1140069956 Bytes)
```
and agree with the flag where space constrained. 

-----

For the backword compatibility discussion, I just want to add my 2 cents here. I generally agree with @buengese regarding human-readable by default. Output parsing should be discouraged and json output, or similar, should be the way to go for scripts.

That said, I think that backword compatibility is very nice to have, but also consistency and least-astonishment are. After all we are not breaking a protocol or a file format, and script users should be able to adapt quite easily.

If we agree that it is too important to break, I suggest at least to track all the changes in a new milestone associated with a new major version number (v2.0.0?) so that it is clear that breaking changes might occur. We may then merge all these "uncomfortable" changes in a single shot.

albertony:
I have now removed the commits "inherited" from the original PR, and refactored into a more conservative change according to comments.

DONE:
- ls: New global option `--human-readable`.
   - Prints sizes with suffix, e.g. `9.762Ki`, instead of raw byte count, `9996`.
   - Uses decimal suffix for human-readable counts in `lsd` (1k means 1000 files), same as `ncdu`.
   - Fixes issue #1890
   - Example, `ls` with `--human-readable` (new option)
     ```
      336.125Ki langs.model.xml
       34.827Ki license.txt
        3.439Mi notepad++.exe
            955 shortcuts.xml
     ```
- size: Include human-readable count.
  - As before: Outputs string with both raw number and human-readable for size.
  - Now also includes human-readable value for count, using decimal suffix.
  - Example:
    ```
    Total objects: 1.931k (1931)
    Total size: 261.527 MiByte (274230930 Byte)
    ```
  - Example, before this change:
    ```
    Total objects: 1931
    Total size: 261.527 MBytes (274230930 Bytes)
    ```
- ncdu: Add toggle key for human-readable vs raw.
  - As before: Always human-readable by default.
  - Introduced support for raw format, switching between raw and human interactively with key `u`.
  - Also included some small tweaks on column widths. E.g. increased raw size width to 12 from 9, since
    9 limits the reserved space to sizes <10G, which is not *that* uncommon I think... The human-readable,
    default, is still the same width of 9.
  - Example:
    ```
      1.538Gi  47.196k  34.180Ki [##########] /go
    641.324Mi  15.222k  43.142Ki [####      ] /.cache
    261.530Mi   1.935k 138.400Ki [##        ] /projects
    ```
- about: Make human-readable output more consistent with other commands
  - As before: Outputs human-readable by default, option `--full` to output raw numbers.
  - Changed the suffix to proper binary/IEC suffix, and same variant as `size`.
  - Changed the count `Objects: Total number of objects in the storage.` to use decimal suffix instead of binary suffix.
  - Example:
    ```
    Total:   952.131 GiByte
    Used:    438.671 GiByte
    Free:    513.460 GiByte
    ```
  - Example, before this change:
    ```
    Total:   952.131G
    Used:    438.671G
    Free:    513.460G
    ```
- tree: Consider global option `--human-readable`.
  - Did already have a custom option `--human`, so the only change is that this was removed in favor of the global `--human-readable`, for consistency.
  - Note: Will not show the binary/IEC standard suffix as the other commands, e.g. `[9.3K]  write.go` instead of `[9.3Ki]  write.go`. This comes from the external library, so needs a change there for this to be consistent.
- test: Consider global option `--human-readable`.
  - Now outputs sizes and counts with suffix instead of raw numbers when `--human-readable`.
  - Also fixed an issue with unsigned integer overflow.

TODO:
- Did `I think we should probably leave about alone.` mean it should be left 100% untouched?
- Should `--human-readable` be global flag, or command-specific for the `ls*`, `tree` and `test`
  commands where it is considered? It is also a bit long to type, so would be nice to find a suitable
  shortform (which is easier when it is not global)
- Which variants to choose for sizes in output from `size`, `about` etc: `Mi`, `MiB`, `MiByte`?
  For counts we only have the short variant, e.g. `M`, since there is no unit (`MObjects`)...
- (Implementation) Reconsider functions SizeString/SizeStringField and corresponding for Count introduced in fs.operations.
  Less relevant when `--human-readable` option did not end up being considered "everywhere", but still
  useful with this logic centralized. Thinking if I can make their usage more self-explanatory, and also
  if there is a better place to put them.

ncw:
> I have now removed the commits "inherited" from the original PR, and refactored into a more conservative change according to comments.

Great :-)

> * ls: New global option `--human-readable`.

Great

> * size: Include human-readable count.

Looks good. There is a --json flag for this so no excuse for scripters!

> * ncdu: Add toggle key for human-readable vs raw.

OK sounds good

> * about: Make human-readable output more consistent with other commands

Looks fine

> * tree: Consider global option `--human-readable`.
>   
>   * Did already have a custom option `--human`, so the only change is that this was removed in favor of the global `--human-readable`, for consistency.

That is a backwards incompatible change. Not a huge deal as it is a minor flag on a minor command.

>   * Note: Will not show the binary/IEC standard suffix as the other commands, e.g. `[9.3K]  write.go` instead of `[9.3Ki]  write.go`. This comes from the external library, so needs a change there for this to be consistent.

OK.

> * test: Consider global option `--human-readable`.
>   
>   * Now outputs sizes and counts with suffix instead of raw numbers when `--human-readable`.
>   * Also fixed an issue with unsigned integer overflow.

Nice!

> TODO:
> 
> * Did `I think we should probably leave about alone.` mean it should be left 100% untouched?

Just don't break backwards compatibility - judging by your before and after above it looks fine.

> * Should `--human-readable` be global flag, or command-specific for the `ls*`, `tree` and `test`
>   commands where it is considered? It is also a bit long to type, so would be nice to find a suitable
>   shortform (which is easier when it is not global)

It is the kind of command you do type a lot...

I'm kind of leaning towards making it a global variable maybe with `-H` as you suggested earlier.

> * Which variants to choose for sizes in output from `size`, `about` etc: `Mi`, `MiB`, `MiByte`?

I'm not fond of `Mi` on its own, but either of the other two is fine.

>   For counts we only have the short variant, e.g. `M`, since there is no unit (`MObjects`)...

OK

> * (Implementation) Reconsider functions SizeString/SizeStringField and corresponding for Count introduced in fs.operations.
>   Less relevant when `--human-readable` option did not end up being considered "everywhere", but still
>   useful with this logic centralized. Thinking if I can make their usage more self-explanatory, and also
>   if there is a better place to put them.

A global flag and having these in operations sounds like a good idea if I'm understanding you correctly.

> 
> And:
> 
> > PS Do I hear a --si flag coming? ;-)
> 
> If this is something that we would want (my impression was that it was not), I could definitely look into it. Biggest question that comes to mind is what to do with input values? Should `--some-size-option 1K` still be considered binary 1024 or should it suddenly mean 1000 if also option `--si`?

It seems the logical conclusion. However it **mustn't** affect parsing of stuff in the config file. It should probably be output only. It maybe is too difficult because of that.

ivandeex:
I merged this patch for my staging branch and I ❤️ it.

@ncw @albertony
when ?...

albertony:
> I merged this patch for my staging branch and I ❤️ it.

Really? Glad to hear!

I need to refresh my memory a bit and look through it, will try to do that later today.

After a quick recap I see there was some confusion regarding short form `-H`. Seems I ended up removing it, leaving only the `--human-readable` variant, due to this comment:

> Is this a global flag? If so we need to be very careful with adding them to make sure no sub commands have a -H flag. Note also that -H means transfer hardlinks in rsync which we are probably unlikely to every use, but just noting using -H closes the door to that bit of rsync compatibility.

But then there was this comment:

> I'm kind of leaning towards making it a global variable maybe with -H as you suggested earlier.

So I should probably add it again, or what do you think @ivandeex ?

ivandeex:
The short form made the PR stuck. Postpone it to another patch.

albertony:
Sure. I'm fine with it being merged as is. But also, if @ncw is leaning even more since last time, perhaps tipped over, I can add the shortform back immediately as well, lets hear what he has to say..

ivandeex:
RFC from me:

1. Postpone short flag `-H` to a later PR
2. Instead of `--human-readable` use name `--human` for long flag (make it not so long)
3. Drop `XiByte` in favor of `XiB` everywhere.  Replace `Xi` by `XiB` in the standard pacer for uniformity.
With current state of PR the standard pacer looks for me like:
```
Transferred:        1.148Gi / 1.758 GiByte, 65%, 4.528 MiByte/s, ETA 2m17s
```
If item 3 gets accepted then it will become this:
```
Transferred:        1.148 GiB / 1.758 GiB, 65%, 4.528 MiB/s, ETA 2m17s
```

albertony:
> Drop XiByte in favor of XiB everywhere

💣 .... I have thought about that one a lot, more with https://github.com/rclone/rclone/pull/5087 but also with this one. In principle I too like `MiB` better than `MiByte`. Its more of a standard form. Pacman, git etc uses it too, can't remember if I've seen `MiByte` (or `MByte`) used in any of the popular tools (except rclone). The upside of `MiByte` is that it avoids *any* confusion if its bit or byte, and rclone used it (`MByte`) before so I landed on keeping that. 🤔

ivandeex:
> The upside of `MiByte` is that it avoids *any* confusion if its bit or byte

Bits (Kbits, Mbits) will be never used for **sizing** things, only in context of **speed**. The convention is to write it as `Kbps` or `Mbps` with `s` for seconds glued in the end without any slash.

In contrast, **bytes** use upper `B`, a multiplier (`KiB` `MiB` `GiB` for Si, or `KB` `MB` `GB` for 2^10 powers) for sizes. Byte speeds always use `/s` with slash for "per-second": `KB/s` `MB/s` `GB/s` or `KiB/s` `MiB/s` `GiB/s`.

I think this slash/no-slash notation can be considered a good visual distinction between bits and bytes.
- bits: `bps` `Kbps` `Mbps` `Gbps`
- 2^10 Bytes: `B/s` `KB/s` `MB/s` `GB/s`
- Si Bytes: `KiB/s` `MiB/s` `GiB/s`

I believe I see it more frequently written this way on the net...

albertony:
Good explanation! And I *do* agree, don't get me wrong. What I meant was that with MiByte(/s) the chance of someone wondering (and asking in forum) if this is bits or bytes should be 0%. With MiB I think it is >0%. And then since this was used by rclone before, I ended up (after a lot of discussions - with myself) keeping it in the already merged https://github.com/rclone/rclone/pull/5087. To quote myself:

> I would say, ideally, not to use this notation, stick with MiB/s. But perhaps it is well-established enough, at least in existing rclone, that its acceptable?

So: I'm fine with changing to `MiB` form, drop `MiByte` form entirely (both for sizes and speeds).

Would be nice to know @ncw 's opinion before cranking up the chain saw, though...

ncw:
@ivandeex wrote:

> RFC from me:
> 
> 1. Postpone short flag `-H` to a later PR

Yes, merging this PR sooner rather than later would be my preference. I don't like PRs that keep getting bits added on - it makes reviewing them even harder!

> 2. Instead of `--human-readable` use name `--human` for long flag (make it not so long)

SGTM

> 3. Drop `XiByte` in favor of `XiB` everywhere.  Replace `Xi` by `XiB` in the standard pacer for uniformity.
>    With current state of PR the standard pacer looks for me like:
> 
> ```
> Transferred:        1.148Gi / 1.758 GiByte, 65%, 4.528 MiByte/s, ETA 2m17s
> ```
> 
> If item 3 gets accepted then it will become this:
> 
> ```
> Transferred:        1.148 GiB / 1.758 GiB, 65%, 4.528 MiB/s, ETA 2m17s
> ```
> > The upside of `MiByte` is that it avoids _any_ confusion if its bit or byte
> 
> Bits (Kbits, Mbits) will be never used for **sizing** things, only in context of **speed**. The convention is to write it as `Kbps` or `Mbps` with `s` for seconds glued in the end without any slash.
> 
> In contrast, **bytes** use upper `B`, and multiplier (`KiB` `MiB` `GiB` for Si, or `KB` `MB` `GB` for 2^10 powers) for sizes. Byte speeds always use `/s` with slash for "per-second": `KB/s` `MB/s` `GB/s` or `KiB/s` `MiB/s` `GiB/s`.
> 
> I think this slash/no-slash notation can be considered a good visual distinction between bits and bytes.
> 
> * bits: `bps` `Kbps` `Mbps` `Gbps`
> * 2^10 Bytes: `B/s` `KB/s` `MB/s` `GB/s`
> * Si Bytes: `KiB/s` `MiB/s` `GiB/s`
[ these last two are the wrong way round! ]
> 
> I believe I see it more frequently written this way on the net...

Hmm, if `Bytes` were an SI unit and `B` was the abbreviation then this would be 100% correct. According to my physics training you'd write it `Bs⁻¹` which would confuse just about everyone!

I guess it is a question of what people find more/less confusing.

@albertony  wrote
> Good explanation! And I _do_ agree, don't get me wrong. What I meant was that with MiByte(/s) the chance of someone wondering (and asking in forum) if this is bits or bytes should be 0%.

This is a good point and why it was originally written as `Byte` rather than `B`.

> With MiB I think it is >0%. And then since this was used by rclone before, I ended up (after a lot of discussions - with myself) keeping it in the already merged #5087. To quote myself:
> 
> > I would say, ideally, not to use this notation, stick with MiB/s. But perhaps it is well-established enough, at least in existing rclone, that its acceptable?
> 
> So: I'm fine with changing to `MiB` form, drop `MiByte` form entirely (both for sizes and speeds).
> 
> Would be nice to know @ncw 's opinion before cranking up the chain saw, though...

Non scientific survey

- `MiB/s` has 360,000 hits on Google
- `MiByte/s` has just 750
- `MB/s` has 38,000,000
- `MByte/s` has 450,000

So according to Google at least the preferred form would be `MiB/s`

I'm happy to make the change. We can always change it back if users are confused.

ivandeex:
> * `MiB/s` has 360,000 hits on Google
> * `MiByte/s` has just 750
> * `MB/s` has 38,000,000
> * `MByte/s` has 450,000
> 
> So according to Google at least the preferred form would be `MiB/s`

Looking at the counters I'd say that...
- `MiB/s` beats `MiByte/s`
- `MB/s` beats `MByte/s`

by orders of magnitude

albertony:
Sounds like we have a plan on the units issue.

I just pushed the change - as a new commit. Should be ready, but will do some more manual testing. Perhaps @ivandeex  can verify if it matches your expectations?

Regarding:

> The short form made the PR stuck.

That was not my impression.

> I don't like PRs that keep getting bits added on - it makes reviewing them even harder!

Of that I am guilty! And haven't learned yet (ref mentioned push)..

> Postpone short flag -H to a later PR

Because we are worried that `-H` closes the door to the transfer hardlinks rsync compatibility? Because if not, then:
- Changing MiByte to MiB etc is a bigger change, and more warrants a separate PR (it is more related to the preceding https://github.com/rclone/rclone/pull/5087 than this PR).
- It was included in the originally PR, my suggestion was - and still is - to use it.
- It is absolutely minimal effort to do the change - same amount as renaming  `--human-readable` that was suggested.
- I personally like `--human-readable` better than `--human`. In addition: rsync, as well as ls, df, dh etc all have `--human-readable`, so it is consistent - almost standardized! Unfortunately they use `-h` as shortform, which is taken in rclone (for help), hence suggestion is `-H`. And if we are going to have  a single character short form anyway (at some point), then it does not matter if it is a bit long.

There! Now I made the PR stuck!! 💪 👹

ivandeex:
> rsync, as well as ls, df, dh etc all have --human-readable, so it is consistent - almost standardized

Convincing. Makes me vote for `--human-readable`

ivandeex:
> Perhaps @ivandeex can verify if it matches your expectations?

```
deex@hardy:~/rclone$ rclone.dev -H -P md5sum keen:test --download
Transferred:       37.984 MiB / 6.100 GiB, 1%, 2.691 MiB/s, ETA 38m27s
Transferred:            0 / 4, 0%
Elapsed time:        15.8s
Transferring:
 *                                            5g:  2% /1Gi, 1.923Mi/s, 8m39s
 *                                          1800:  0% /1.758Gi, 278.912Ki/s, 1h49m57s
 *                                           27g:  0% /3.245Gi, 278.845Ki/s, 3h23m7s
 *                                           100:  4% /100Mi, 465.058Ki/s, 3m29s^C
deex@hardy:~/rclone$ rclone.dev -H ls keen:test
    100Mi 100
  1.758Gi 1800
  3.273Gi 27g
      1Gi 5g
    700Mi 700
        8 hello.txt
deex@hardy:~/rclone$
```

Yes it does.
Upvoting the merge.

ivandeex:
> Regarding:
> 
> > The short form made the PR stuck.
> 
> That was not my impression.
> 
> > Postpone short flag -H to a later PR
> 
> Because we are worried that `-H` closes the door to the transfer hardlinks rsync compatibility? Because if not, then:
> 

I abstain w.r.t `-H`
Let you and @ncw decide
Just wish it merged earlier

albertony:
Sure. If there are still real conserns about taking `-H`, I'm fine with merging without it (couldn't find any good alternatives either). Up to @ncw.

ivandeex:
> couldn't find any good alternatives

With Unicode, we are not limited by 52 short flags.

`rclone ls -本 remote:path`

ivandeex:
> couldn't find any good alternatives

With Unicode, we are not limited by 52 [short](https://ja.wikipedia.org/wiki/%E6%97%A5%E6%9C%AC%E4%BA%BA) flags.

`rclone ls -出 remote:path` (reads [dash-Jin])

ivandeex:
> couldn't find any good alternatives

With Unicode, we are not limited by 52 [short](https://ja.wikipedia.org/wiki/%E6%97%A5%E6%9C%AC%E4%BA%BA) flags.

```
rclone ls -出 remote:path
```

albertony:
> With Unicode, we are not limited by 52 short flags.

Thats a touch one on my keyboard... 😅 

If the short variant could be more than one (ascii!!) character it would be easier.. We could probably quite easily "fake" variants by creating a separat option that just sets the same boolean variable, e.g. to have `-hr` as shortcut for `--human-readable`. Don't know if there are any traps going down that route, and it would sort of introduce a new standard (in rclone), but is it an idea to try out?

ivandeex:
> If the short variant could be more than one (ascii!!) character it would be easier...
> ... but is it an idea to try out?

No, because `-hr` is in fact super-short form of `-h -r` in GNU world and [Cobra](https://github.com/spf13/cobra).
Thou go's [flag](https://pkg.go.dev/flag) broke the tradition and treats `-hr` as equivalent to `--hr`, cobra is primary CLI parser in rclone.

ncw:
> > rsync, as well as ls, df, dh etc all have --human-readable, so it is consistent - almost standardized
> 
> Convincing. Makes me vote for `--human-readable`

Let's go for that!



> > couldn't find any good alternatives
> 
> With Unicode, we are not limited by 52 [short](https://ja.wikipedia.org/wiki/%E6%97%A5%E6%9C%AC%E4%BA%BA) flags.
> 
> ```
> rclone ls -出 remote:path
> ```

He he  - @ivandeex you are missing your career in comedy :-)

Let's leave out the shortform for the moment, get this PR merged, then open another issue where we can talk about it !

albertony:
> Let's leave out the shortform for the moment, get this PR merged

Agreed. I guess next step is a formal review from @ncw then.

ncw (approved):
This looks great to me - thank you @albertony :-)

Please merge now and we'll pick up any fallout before 1.57 is released.

albertony:
Thanks. Merged!
Hopefully the beta will get some users, and they will find the feature... which made me wonder: How well is this documented? Not great, probably. 🤒 Probably deserves a line or two in `flags.md` and `docs.md`, I will look into that.

albertony:
Oh wait, it will be included in flags.md - after running the gendocs command. 😌 

Then perhaps that is enough, for now.

I will consider adding a section in docs.md later. It could be noted that only some commands consider `--human-readable`, some does not but outputs readable by default with another option for raw numbers (`about`), some output both human and raw numbers (`size`), and `ncdu` has its own interactive key, etc..

And docs for ls:

> Lists the objects in the source path to standard output in a human
> readable format with size and path. Recurses by default.

Should be updated with:

> Use --human-readable to make it super-human-readable! 

🤦

Issue #4139, opened 2020-04-16
onedrive hangs on token issue without any notice


Jwink3101:
#### What is the problem you are having with rclone?

When I have a token issue on OneDrive, it just hangs without anything helpful. 

**Note**: I am not asking how to fix the issue (I haven't tried yet but I suspect I just need to reauthenticate). I am reporting the lack of clarity from rclone

#### What is your rclone version (output from `rclone version`)

```
rclone v1.51.0
- os/arch: darwin/amd64
- go version: go1.13.7
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

macOS 10.14.6

####  Which cloud storage system are you using? (eg Google Drive)

OneDrive personal

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

This is *intentionally* without `-vv`

    $ rclone ls --max-depth 1 --config onedrive.cfg  onedrive:

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

Now the same thing as above but *with* `-vv`

```
$ rclone ls --max-depth 1 --config onedrive.cfg  -vv onedrive:
2020/04/16 17:57:57 DEBUG : rclone: Version "v1.51.0" starting with parameters ["rclone" "ls" "--max-depth" "1" "--config" "onedrive.cfg" "-vv" "onedrive:"]
2020/04/16 17:57:57 DEBUG : Using config file from "<<REDACTED>>/onedrive.cfg"
2020/04/16 17:57:57 DEBUG : One drive root '': Token expired but no uploads in progress - doing nothing
2020/04/16 17:57:57 DEBUG : onedrive: Loaded invalid token from config file - ignoring
```

The issue is that it is hanging but not exiting as one would expect.

ncw:
Does your token have a `refresh_token` in it? That is probably why it isn't working...

Rclone should be printing some kind of error here!

Jwink3101:
I do not have the old token that was causing this issue. As I noted (and it seems you understood) was *not* that it wasn't working but rather than it hangs on the error.

I reauthorized the next day and it works fine again!

ncw:
> I do not have the old token that was causing this issue. As I noted (and it seems you understood) was _not_ that it wasn't working but rather than it hangs on the error.

Sure, I was just trying to work out where it was hanging!

If it was because of an empty refresh token that is something we could fix easily I think.

> I reauthorized the next day and it works fine again!

:-)

[...]

PR #5252, opened 2021-04-17
oauthutil: raise fatal error if token expired without refresh token

Cnly:
This should improve the situations reported in #4139.

Normally we can automatically renew the OAuth access token with the refresh token, but in some cases we don't have the refresh token and there's no way we can recover from it I guess.


ncw (approved):

Please merge when you are ready.

PR #5258, opened 2021-04-19
Delay load config file

### LINKED ISSUE #5228: Regression: --version requires configuration password
In v1.50.2, running `rclone --version` with an encrypted config file would output the version.

In v1.55.0, the same command asks for the configuration password first. This should not be necessary, as none of the information output by `--version` seems related to the configuration.

```
$ rclone --version
Enter configuration password:
password:
rclone v1.55.0
- os/type: linux
- os/arch: amd64
- go/version: go1.16.2
- go/linking: static
- go/tags: cmount
```

(Also, the password prompt is doubled, but that's a separate issue.)

### LINKED ISSUE #5236: rclone mount fails with --daemon flag when asking for config password
Originally posted on the forum [here](https://forum.rclone.org/t/1-55-rclone-mount-with-daemon-looks-like-it-works-but-actually-fails-instantly/23456/8)

rclone mount fails with an encrypted config file when the --daemon flag is included. Works fine without --daemon.

albertony:

Restructuring of config code in v1.55 resulted in config file being loaded early at process startup. If configuration file is encrypted this means user will need to supply the password, even when running commands that does not use config. This also lead to an issue where mount with --deamon failed to decrypt the config file when it had to prompt user for passord.

This PR removes the forced config file loading at startup, and instead waits until first actual config access.

Fixes #5236
Fixes #5228

#### Checklist

### DIFF
--- a/backend/alias/alias_internal_test.go
+++ b/backend/alias/alias_internal_test.go
@@ -20,7 +20,7 @@ var (
 )
 
 func prepare(t *testing.T, root string) {
-	configfile.LoadConfig(context.Background())
+	configfile.Install()
 
 	// Configure the remote
 	config.FileSet(remoteName, "type", "alias")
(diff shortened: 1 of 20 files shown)


ncw (approved):
Apologies for the delay in reviewing.

That code looks great now - please merge :-)

Issue #5225, opened 2021-04-09
Add support for GOIndex/GDIndex?


13steinj:
#### The associated forum post URL from `https://forum.rclone.org`

N/A, but I think the rclone docs told me to directly make an issue for such a thing on the http backend (sorry if I should have made a post on the forums first):

> This has been tested with common webservers such as Apache/Nginx/Caddy and will likely work with file listings from most web servers. (If it doesn't then please file an issue, or send a pull request!)

#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.55.0
- os/type: linux
- os/arch: amd64
- go/version: go1.16.2
- go/linking: static
- go/tags: cmount
```

#### What problem are you are trying to solve?

rclone does not support GOIndex/GDIndex, but it can!

Similar to #3913, however it is *not* entirely JS based!

JS is only used to hide google app credentials from the public, all files/folders are fetchable via http request (POST to get a list of files, you get a JSON response listing the mime-type of file (/something custom if it's a folder / google doc), the google ID (irrelevant), the name of the file, and the modTime, GET with url params to either get a file / get a page that can view the file in browser / get a page that lists the relevant folder in-browser by sending a post request to the same URL and then parsing the response.

For context, a JS listing response is as follows:

```json
{
    "files": [
        {
            "id":"[redacted]",
            "name":"Music",
            "mimeType":"application/vnd.google-apps.folder",
            "modifiedTime":"2021-03-09T13:03:52.009Z"
        },
        {
            "id":"[redacted]",
            "name":"Video",
            "mimeType":"application/vnd.google-apps.folder",
            "modifiedTime":"2021-03-09T13:03:59.865Z"
        },
        {
            "id":"[redacted]",
            "name":"0x.goindex.gif",
            "mimeType":"image/gif",
            "modifiedTime":"2021-03-09T14:01:58.788Z",
            "size":"458131"
        },
        ...
    ]
}
```

No idea if I had to redact that stuff or not, but just in case.

People use this + some server / serverless setup (usually cheap/free cloudflare workers) to get around Google Drive sharing limits (because this way it's technically not necessarily sharing, it's accessing Google Drive in some way and then piping the file request from a custom server to the user being shared with). People also use this because no email sharing necessary, multiple drive support in some cases, easier / better preview support, custom statistics, etc.

#### How do you think rclone should be changed to solve that?

I'd love to make a pull request, but I wanted to make the issue first because

I wanted to make sure such a PR would be welcome, and if yes:
[...]

Issue #3913, opened 2020-01-29
Unable to fetch file list from index


hjyoung1:
#### What is the problem you are having with rclone?

I'm trying to add http index using http connection. But for some reason rclone is not getting any results from index.
Index I'm adding is running GDIndex (https://github.com/maple3142/GDIndex/) (example https://gdindex-demo.maple3142.workers.dev) 

#### What is your rclone version (output from `rclone version`)

rclone v1.50.2
- os/arch: android/arm64
- go version: go1.13.4

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Using termux to run rclone. Device arm64. 

####  Which cloud storage system are you using? (eg Google Drive)

http Connection

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

rclone -vv ls dindex:

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

![Screenshot_20200129-161606](https://user-images.githubusercontent.com/45882061/73350980-17db0e80-42b4-11ea-82df-b81fd7dce8ce.jpg)

ncw:
Alas that index is 100% javascript - rclone needs a static index to read.

```html
<!DOCTYPE html>
<html lang=en>
  <head>
    <meta charset=utf-8>
    <meta http-equiv=X-UA-Compatible content="IE=edge">
    <meta name=viewport content="width=device-width,initial-scale=1">
    <title>GDIndex</title>
    <link href="/~_~_gdindex/resources/css/app.css" rel=stylesheet>
  </head>
  <body>
    <script>window.props = { title: 'GDIndex', default_root_id: 'root', api: location.protocol + '//' + location.host, upload: false }</script>
    <div id=app>
    </div>
    <script src="/~_~_gdindex/resources/js/app.js">
    </script>
  </body>
</html>
```

hjyoung1:
Oh  :(

PR #5280, opened 2021-04-27
[WIP] GoIndex/GdIndex support

13steinj:

Add [initial] support for GoIndex/GdIndex remotes.

TODO: Support for donwa-style GoIndexes (uncommitted, testing) and GdIndex (incomplete), 

#5225, #3913 (mistakenly thought not possible)

#### Checklist

### DIFF
--- a/backend/http/http.go
+++ b/backend/http/http.go
@@ -262,6 +262,9 @@ func (f *Fs) Features() *fs.Features {
 
 // Precision is the remote http file system's modtime precision, which we have no way of knowing. We estimate at 1s
 func (f *Fs) Precision() time.Duration {
+	if f.opt.GoIndexDriveIndex >= 0 { // uses precision of milliseconds just like Google Drive
+		return time.Millisecond
+	}
 	return time.Second
 }
 
--- a/backend/http/http.go
+++ b/backend/http/http.go
@@ -98,6 +98,11 @@ If you set this option, rclone will not do the HEAD request.  This will mean
 `,
 			Default:  false,
 			Advanced: true,
+		}, {
+			Name:     "goindex_drive_idx",
+			Help:     "Index number of the drive on a GoIndex http backend. Only to be used with GoIndex sites",
+			Default:  -1,
+			Advanced: true,
 		}},
 	}
 	fs.Register(fsi)
@@ -105,10 +110,11 @@ If you set this option, rclone will not do the HEAD request.  This will mean
 
 // Options defines the configuration for this backend
 type Options struct {
-	Endpoint string          `config:"url"`
-	NoSlash  bool            `config:"no_slash"`
-	NoHead   bool            `config:"no_head"`
-	Headers  fs.CommaSepList `config:"headers"`
+	Endpoint          string          `config:"url"`
+	NoSlash           bool            `config:"no_slash"`
+	NoHead            bool            `config:"no_head"`
+	GoIndexDriveIndex int             `config:"goindex_drive_idx"`
+	Headers           fs.CommaSepList `config:"headers"`
 }
 
 // Fs stores the interface to the remote HTTP files
@@ -160,6 +166,9 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 
 	if !strings.HasSuffix(opt.Endpoint, "/") {
 		opt.Endpoint += "/"
+	}
+	if opt.GoIndexDriveIndex >= 0 {
+		opt.Endpoint += strconv.Itoa(opt.GoIndexDriveIndex) + ":/"
 	}
 
 	// Parse the endpoint and stick the root onto it
--- a/backend/http/http.go
+++ b/backend/http/http.go
@@ -519,7 +519,11 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 					add(file)
 				case fs.ErrorNotAFile:
 					// ...found a directory not a file
-					add(fs.NewDir(remote, timeUnset))
+					if f.opt.GoIndexDriveIndex >= 0 { // since goindexes are based on google drive, this is accurate
+						add(fs.NewDir(remote, file.modTime))
+					} else {
+						add(fs.NewDir(remote, timeUnset))
+					}
 				default:
 					fs.Debugf(remote, "skipping because of error: %v", err)
 				}
@@ -595,14 +599,18 @@ func (o *Object) url() string {
 
 // stat updates the info field in the Object
 func (o *Object) stat(ctx context.Context) error {
-	if o.fs.opt.NoHead {
+	if o.fs.opt.NoHead && o.fs.opt.GoIndexDriveIndex < 0 {
 		o.size = -1
 		o.modTime = timeUnset
 		o.contentType = fs.MimeType(ctx, o)
 		return nil
 	}
 	url := o.url()
-	req, err := http.NewRequestWithContext(ctx, "HEAD", url, nil)
+	reqMethod := "HEAD"
+	if o.fs.opt.GoIndexDriveIndex >= 0 {
+		reqMethod = "POST"
+	}
+	req, err := http.NewRequestWithContext(ctx, reqMethod, url, nil)
 	if err != nil {
 		return errors.Wrap(err, "stat failed")
 	}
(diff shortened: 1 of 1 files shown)


backend/http/http.go
```diff
@@ -161,6 +168,9 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if !strings.HasSuffix(opt.Endpoint, "/") {
 		opt.Endpoint += "/"
 	}
+	if opt.GoIndexDriveIndex >= 0 {
+		opt.Endpoint += strconv.Itoa(opt.GoIndexDriveIndex) + ":/"
```
13steinj, line 172: Technically a drive index does not need to be specified, and if so should be determined by a test request to see redirects.

backend/http/http.go
```diff
@@ -365,6 +378,36 @@ func parse(base *url.URL, in io.Reader) (names []string, err error) {
...
 }
 
+type GoIndexFile struct {
+	Id           string `json:"id"`
+	MimeType     string `json:"mimeType"`
+	ModifiedTime string `json:"modifiedTime"`
+	Name         string `json:"name"`
+	// don't know why, but this is a string instead of an int64 (against gdrive doc)
+	Size          string `json:"size"`
+	ThumbnailLink string `json:"thumbnailLink"`
```
13steinj, line 388: AFAIK ThumbnailLink is unneeded and can be discarded, but left in the chance that rclone can actually use it (if not I'll remove)

backend/http/http.go
```diff
@@ -365,6 +378,36 @@ func parse(base *url.URL, in io.Reader) (names []string, err error) {
 	return names, nil
 }
 
+type GoIndexFile struct {
+	Id           string `json:"id"`
```
13steinj, line 382: Id might be useful for cache purposes

backend/http/http.go
```diff
@@ -365,6 +378,36 @@ func parse(base *url.URL, in io.Reader) (names []string, err error) {
...
+	MimeType     string `json:"mimeType"`
+	ModifiedTime string `json:"modifiedTime"`
+	Name         string `json:"name"`
+	// don't know why, but this is a string instead of an int64 (against gdrive doc)
+	Size          string `json:"size"`
+	ThumbnailLink string `json:"thumbnailLink"`
+}
+
+type GoIndexResults struct {
+	Data struct {
```
13steinj, line 392: Should add curPageIndex and nextPageToken fields, for larger go-indexes which paginate file listings.

backend/http/http.go
```diff
@@ -365,6 +378,36 @@ func parse(base *url.URL, in io.Reader) (names []string, err error) {
...
+	} `json:"data"`
+}
+
+// parser for GoIndex remotes
+func parseGoIndex(u *url.URL, in io.Reader) (names []string, err error) {
+	var results GoIndexResults
+	dec := json.NewDecoder(in)
+	err = dec.Decode(&results)
+	if err != nil {
+		return nil, err
```
13steinj, line 403: I'll wrap this error in a soon-to-be commit.

backend/http/http.go
```diff
@@ -390,7 +433,13 @@ func (f *Fs) readDir(ctx context.Context, dir string) (names []string, err error
 		return nil, errors.Errorf("internal error: readDir URL %q didn't end in /", URL)
 	}
 	// Do the request
-	req, err := http.NewRequestWithContext(ctx, "GET", URL, nil)
+	reqMethod := "GET"
+	var body io.Reader = nil
+	if f.opt.GoIndexDriveIndex >= 0 {
+		reqMethod = "POST"
+		body = strings.NewReader(`{"q":""}`)
```
13steinj, line 440: Needed for Acirour-style indexes.

backend/http/http.go
```diff
@@ -414,6 +463,14 @@ func (f *Fs) readDir(ctx context.Context, dir string) (names []string, err error
 		if err != nil {
 			return nil, errors.Wrap(err, "readDir")
 		}
+	case "text/plain":
```
13steinj, line 466: Go figure, for json responses GoIndexes use text/plain.

backend/http/http.go
```diff
@@ -538,14 +599,18 @@ func (o *Object) url() string {
 
 // stat updates the info field in the Object
 func (o *Object) stat(ctx context.Context) error {
-	if o.fs.opt.NoHead {
+	if o.fs.opt.NoHead && o.fs.opt.GoIndexDriveIndex < 0 {
```
13steinj, line 602: If this stays in the http backend, perhaps "NoHead" should be changed, because one would disable individual-file / folder POST requests instead.

backend/http/http.go
```diff
@@ -558,20 +623,37 @@ func (o *Object) stat(ctx context.Context) error {
...
+	var (
+		fSize int64
+		fTime time.Time
+		fType string
+	)
+	if o.fs.opt.GoIndexDriveIndex >= 0 {
+		var fInfo GoIndexFile
+		dec := json.NewDecoder(res.Body)
+		dec.Decode(&fInfo)
+		fSize = parseInt64(fInfo.Size, -1)
```
13steinj, line 635: Yeah I can't figure out why I need to manually parse this. GoIndex directly passes the data from the v3 Google Drive API, but those docs say it's supposed to be a long (unless that's mistaken / longs are strings in that context, which is possible, I've seen more weird things from official Google API docs).

backend/http/http.go
```diff
@@ -558,20 +623,37 @@ func (o *Object) stat(ctx context.Context) error {
...
 	// If NoSlash is set then check ContentType to see if it is a directory
-	if o.fs.opt.NoSlash {
+	if o.fs.opt.NoSlash || o.fs.opt.GoIndexDriveIndex >= 0 {
 		mediaType, _, err := mime.ParseMediaType(o.contentType)
 		if err != nil {
 			return errors.Wrapf(err, "failed to parse Content-Type: %q", o.contentType)
 		}
-		if mediaType == "text/html" {
+		if o.fs.opt.NoSlash && mediaType == "text/html" ||
+			o.fs.opt.GoIndexDriveIndex >= 0 && mediaType == "application/vnd.google-apps.folder" {
```
13steinj, line 656: I haven't encountered a GoIndex with online-google files (ex, a google doc) yet, so I still need to test that out, but as far as I know GoIndex doesn't support those types of files so it shouldn't be encountered / need to be handled.

backend/http/http.go
13steinj: (and doesn't do any harm for non-acirou style indexes)

backend/http/http.go
```diff
@@ -365,6 +378,36 @@ func parse(base *url.URL, in io.Reader) (names []string, err error) {
...
+	var results GoIndexResults
+	dec := json.NewDecoder(in)
+	err = dec.Decode(&results)
+	if err != nil {
+		return nil, err
+	}
+	for _, goIndexFile := range results.Data.Files {
+		names = append(names, goIndexFile.Name)
+	}
+	return names, nil
```
13steinj, line 408: @ncw So this is where we get into "should this be its own backend territory", because

* I can modify the http backend to return `Object`s here and in the normal `parse`, adding a name field, and modify the rest of the `List` routine to handle that. Don't know if there's significantly more cost to passing around `Object`s with empty fields bar "Name" (in the case of normal http remotes). But doing this would provide a significant speed boost to GoIndex remotes, because instead of doing a POST for each file, stats can be determined from the directory listing (at least initially)
*  Copy+paste a lot of the http backend code into a new, "goindex" backend, so violating DRY, but avoiding and GoIndex specifics in the http backend.
* While still unimplemented by me, the GdIndex flavor supports uploads (still figuring that out)
* While still unimplemented by me, GoIndexes support directory-level (afaik piercing, but not general) passwords *on top of* HTTP basic authentication. Ex, without the password, if someone knows a file in the directory they still can *read* it as well as *stat* it, but the user can't list the directory (ex, on *nix, mark the directory with `chmod u-r+x`

I'm fine with either keeping this in http or making it separate, just that I'd want this to be Nick's / maintainers' decision, not mine. I would imagine a different backend would be preferable, because otherwise one could just keep adding tiny hack upon tiny hack for each specific custom http remote to the http backend until it becomes a gigantic mess.

ncw (commented):
I've had a look at the code so far...

I think this probably should be its own backend.

If it ever grows write support it would definitely want to be.

Another advantage of being its own backend is I can get you to maintain it!

It needs tests though - thoughts on that?

13steinj:
> I've had a look at the code so far...

> I think this probably should be its own backend.

> If it ever grows write support it would definitely want to be.

Sounds good, I thought that this was going to be the end-scenario given option specialization.

> Another advantage of being its own backend is I can get you to maintain it!

Haha mere fact that I have use of such a system makes it favorable (personally) that I maintain it, and I know that others use it as well so I'd be glad to. As far as I know, not many changes have been made to the codebases either in a long time.

> It needs tests though - thoughts on that?

I'll keep writing code (disregarding tests) for now, and hopefully I won't have to change much after adding tests.

There's a couple of options regarding tests here:

1. hit the example goindexes that are currently up as examples from that repo I linked in a [comment on the previous GH issue](https://github.com/rclone/rclone/issues/5225#issuecomment-819093666):
  > Links to code above. Here's some links direct to some active indices: [Acrou theme in dark mode (I think only a single account)](https://goindex.teamsdrives.workers.dev/), [Acrou's light mode theme is currently broken, but the code itself works, so probably old credentials](https://oss.achirou.workers.dev/), [gdindex dark mode](https://gdindex-dark.teamsdrives.workers.dev/), [gdindex normal/light mode](https://gdindex-demo.maple3142.workers.dev/), [donwa's classic theme](https://goindex-classic.teamsdrives.workers.dev/), [donwa material theme](https://goindex-classic.teamsdrives.workers.dev/).
  
  Caveat is that these could technically go down for whatever reason. Also we'd want to record the http responses, because if these hit their daily limit, they'd error out.

2. set up a couple of google accounts and make our own goindexes for those, with at least 1001 empty text files and a few example files (ex, like in the other goindexes). The goindex url would have to be kept secret via an env variable on the test suite in some form.

3. similar to option 2, but use [cloudworker + docker](https://github.com/dollarshaveclub/cloudworker), but I'd have to update cloudworker to any changes that cloudflare made between feb 9 and now (though I somehow doubt they've actually made changes)

ivandeex:
Hmm. 5 months since last update :O
Do you plan to finish this work @13steinj ?

13steinj:
@ivandeex apologies, I do plan to complete this PR sometime in December. This is more complex than I originally anticipated (since the backend I'm referencing varies a bit in terms of API implementation; there's roughly 4-6 different dominating forks of the project each with their own slightly different backend API, but they all behave to a user in the same way). With researching these as I find them, school, work, and illness, I've been slow in getting my changes working (and slow in pushing them to this branch).

ncw:
I'm going to close this since I haven't heard from you in a while.

If you want to re-open or resubmit that would be fine.

PR #5292, opened 2021-04-29
build: add gomobile android build

x0b:

Add `gomobile bind` to build.yml to know when it breaks. I added it directly to the existing android build process since an independent failure is deemed unlikely.

https://github.com/x0b/rcx/issues/123#issuecomment-828410059

#### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -267,6 +267,15 @@ jobs:
          run: |
            make
 
+       - name: install gomobile
+         run: |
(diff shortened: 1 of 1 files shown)


ncw:
That looks brilliant thanks! It only add 1:20 to the build which makes the android build not the slowest still :-)

I'll merge that now thank you.

PS I'm working on the new state based config system - it is a huge change alas but should be good when done! I'll Update the issue when I'm ready to share.

PR #5297, opened 2021-05-02
local: skip files removed concurrently with `List`

ivandeex:

**UPDATED on 2021-06-12**

This change fixes the bug described below:
if a file is removed while the local backend List() runs, the call will flag an accounting error.
The bug manifests itself if local backend is the Sync target due to intrinsic concurrency.
The odds to hit this bug depend on --checkers and --transfers.
Chunker over local backend is affected even more because updating a composite object
with a smaller size content translates into removing chunks on the underlying file system
and involves a number of List() calls.

This patch will make `List` **skip the file**.

Unit test:
```
go test ./fs/sync -remote TestChunkerChunk3bLocal: -run TestSyncConcurrent
```
or run the full rclone integration test with `-backends local,chunker`.

No

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -467,6 +467,10 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 				for _, name := range names {
 					namepath := filepath.Join(fsDirPath, name)
 					fi, fierr := os.Lstat(namepath)
+					if os.IsNotExist(fierr) {
+						// skip entry removed by a concurrent goroutine
+						continue
+					}
 					if fierr != nil {
 						err = errors.Wrapf(err, "failed to read directory %q", namepath)
 						fs.Errorf(dir, "%v", fierr)
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -624,9 +624,7 @@ func (s *syncCopyMove) srcParentDirCheck(entry fs.DirEntry) {
 	if parentDir == "." {
 		parentDir = ""
 	}
-	if _, ok := s.srcEmptyDirs[parentDir]; ok {
-		delete(s.srcEmptyDirs, parentDir)
-	}
+	delete(s.srcEmptyDirs, parentDir)
 }
 
 // parseTrackRenamesStrategy turns a config string into a trackRenamesStrategy
--- a/fs/sync/sync_test.go
+++ b/fs/sync/sync_test.go
@@ -1963,3 +1963,54 @@ func TestMaxTransfer(t *testing.T) {
 	t.Run("Soft", func(t *testing.T) { test(t, fs.CutoffModeSoft) })
 	t.Run("Cautious", func(t *testing.T) { test(t, fs.CutoffModeCautious) })
 }
+
+func testSyncConcurrent(t *testing.T, subtest string) {
+	const (
+		NFILES     = 20
+		NCHECKERS  = 4
+		NTRANSFERS = 4
+	)
+
+	ctx, ci := fs.AddConfig(context.Background())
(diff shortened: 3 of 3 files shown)


ncw (commented):
The window for this to happen is quite small isn't it?

Do we need to recheck? Maybe just checking for `os.IsNotExist` on the `os.Lstat`  is enough for us to skip the file. The listing said the file was there - the stat says its missing which means

1. the file got  deleted, or
2. the filesystem is broken

The first is the likely case, the second is very unlikely - I've never seen it. So I'd propose just a log, skip the file and carry on.

Effectivelly your code stats the missing files twice doesn't it? So if you feel uneasy about that you could stat the file twice, but I'm not sure it is necessary.

ivandeex:
I'm easy about it. The digging wasn't quite as easy.
tl;dr Chunker falled victim of its heavy use of `List`.
<gory-details>
I started from the bisync torture test, which gave me unexpectedly high failure rate if one of sides would be chunker. Failed mostly when a chunked file would be overwritten by a smaller contents. All returned error in the call chain happen to be nil, but the sync (bisync calls `Sync` in a couple places) will abort due to the lstat error flagged by someone in the context structure shared between a number of goroutines and bubbling up in a random sibling thread.
</gory-details>

ncw:
> I'm easy about it. 

Simpler fix until we need to make it more complicated seems good to me.

> The [digging](https://github.com/ivandeex/rclone/commits/chunker-torture) wasn't quite as easy. :)

No-ones favorite sort of bug - concurrency problems!

> tl;dr Chunker fell victim of its heavy use of `List`. By fixing localFs I'm rescuing the victim.

:-)

> > The window for this to happen is quite small
> 
> The interval is large enough for 4+ checkers and 4+ remover gorutines to step on each others toes within a single `Sync` call. Context switches happen due to IO kernel calls, and go densely because SSD+cache works fast.

Ah yes. The interval is small in terms of code, but there is IO in there and syscalls which likely the go runtime is scheduling on so what should never happen happens all the time!

ncw (approved):
This looks OK to me if you want to merge it?

(It is still set to draft)

ivandeex:
> This looks OK to me if you want to merge it?
> 
> (It is still set to draft)

It's missing unit tests:

1. Add N "even" files to local path A, N "even" + N "odd"  to B, run `sync` A -> B with 4+ transfers, 4+ checkers. sync runners will delete and check (list) in parallel. Failure probability is >0 without, =0 with patch.

2. Similar but with `TestChunkerChunk3bLocal:` remote from the integration suite. Mix of 9-byte (3 chunks + metadata) and 0-byte (only metadata) chunked files on side A, only 9-byte guys on B, then `sync` A -> B. Internally reduces to removing underlying files and a lot of listings.

ncw:
This is going to be very hard to make a test which fails reliably I think, though I'd be delighted to be proved wrong :-)

ivandeex:
> This is going to be very hard to make a test which fails reliably I think, though I'd be delighted to be proved wrong :-)

Flaky test is "better" (as in "gives you food for thought") than no test.
A note in PR to run the test manually as `go test ./backend/bbb/... -remote rrr -count 1000`
will make fail "reliably" :)

ivandeex:
@ncw

I have added the unit test which is very quick and fails with ~99% probability :)
Check it out - comment out the single `continue` statement added by this patch and run
```
go test ./fs/sync -remote TestChunkerChunk3bLocal: -run TestSyncConcurrent
```
The `TestChunkerChunk3bLocal` remote is part of our standard integration tests.

Description updated. Ready to merge. PTAL :pray:

ivandeex:
@ncw ready for 1.56

Issue #4685, opened 2020-10-15
Deglobalise rclone config


DigitalMarc:
Hi, 
I'm using rclone as a library but so far I had to prevent concurrent access (rclone config updates).
I was wondering if there is a plan to make rclone usable with various configs in parallel.

Thanks!

ncw:
# Deglobalise rclone config

At the moment having global config is limiting rclone's ability to run different processes concurrently.

This is not noticeable in the command line, but when using rclone via the API or as a library it is very noticeable.

## Globals which need fixing

Rclone global config is registered by `rc.AddOption` and these can be seen using the rclone API with

    rclone rc --loopback options/get

The top level general config blocks are these - these are used for all rclone commands more or less.

- filter - filters in use for listing
- main - configures the general non-backend flags
- log - configures --log-file/format/syslog
- rc - configures remote control server (and `rclone rcd`)

These config blocks are used only by the relevant commands

- dlna - configures `rclone serve dlna`
- ftp - configures `rclone serve ftp`
- http - configures `rclone serve http`
- mount - configures `rclone mount`
- sftp - configures `rclone serve sftp`
- vfs - configures vfs flags for rclone mount/serve

## High level plan

Create a system to attach the current configuration to a `context.Context`.

Wherever global config is used, fetch it from the context, so instead of reading config like this

    fs.Config.DryRun

Read it with

    fs.GetConfig(ctx).DryRun

Where `fs.GetConfig` will read the ctx looking for a key for the main config. If one isn't found then it will use the (now private) default config.

This will need functions to create the new config contexts - maybe with a merge non-default values into the new config. This will enable use of rclone as a library with non-global config. So something like

    myConfig := fs.NewConfig()
    myConfig.DryRun = true // set custom config here
    newCtx := fs.SetConfig(ctx)

This will need a similar Set/Get function for each config block, however only the global config is widely spread througout the rclone code so there may be a neater way of doing it. This needs a bit of thought.

As part of this plan the now global config variables should be made package private so they can't be read from different packages.

## Potential difficulties

### Context everwhere

Rclone was started (at the time of go 1.0!) well before the `context.Context` system was invented. There has been some effort threading context through rclone, but there are likely more places that it needs to go. So for this change to be successful it will be necessary to thread context through more of rclone.

This is by and large a mechanical process and with the right editor support it doesn't take too long.

It will generate a lot of changes though!

See: https://github.com/rclone/rclone/issues/3257 for background.

### Config everywhere

[...]

PR #5300, opened 2021-05-03
Replace defaultConfig with a thread-safe in-memory implementation.

macklin-10x:

Replaces the "every method panics" defaultStorage implementation with a thread-safe in-memory implementation backed by a map, protected by a RWMutex.

I opted to replace the default implementation rather than adding this in parallel, though I can of course rename this type and add it to the codebase without clobbering the existing default (opt-in rather than opt-out of in-memory config).

Discussed in the context of #4685

#### Checklist

### DIFF
--- a/fs/config/default_storage.go
+++ b/fs/config/default_storage.go
@@ -1,61 +1,114 @@
 package config
 
-// Default config.Storage which panics with a useful error when used
-type defaultStorage struct{}
+import (
+	"encoding/json"
+	"sync"
+)
 
-var noConfigStorage = "internal error: no config file system found. Did you call configfile.Install()?"
-
-// GetSectionList returns a slice of strings with names for all the
-// sections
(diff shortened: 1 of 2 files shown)


macklin-10x:
Hmm looks like ci is failing on some generic build issue.

ncw:
Took me a moment to find it (that output isn't very helpful though if you look at https://github.com/rclone/rclone/pull/5300/files it shows you the problem too)

```
# github.com/rclone/rclone/fs/config
fs/config/default_storage.go:17:43: syntax error: unexpected newline, expecting comma or }
```

So I think that is your new code

macklin-10x:
Ah oops; also we did some repo shuffling and the repo that I created this PR from no longer exists.  I'll need to close it and open a fresh version.

macklin-10x:
https://github.com/rclone/rclone/pull/5306

PR #5305, opened 2021-05-04
Docs: Add notes about leading path separators and unc paths

albertony:

1. Mention that network/unc paths are supported in local filesystem on windows
2. Mention that additional leading path separators are ignored
   - This is the case from the linked forum thread.
   - Are we sure that this is how it is working? From my testing and code inspection it seems to be true.
   - Are we confident this is how we want it to be, so that it is correct to document it - which avoids uncertainty as in the forum question? Or is it more of a coincidence that should be left undocumented?
3. Fixes a wrong link

https://forum.rclone.org/t/leading-slash-in-source-dest-paths/24005

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -163,7 +163,7 @@ The syntax of the paths passed to the rclone command are as follows.
 This refers to the local file system.
 
 On Windows `\` may be used instead of `/` in local paths **only**,
-non local paths must use `/`. See [local filesystem](https://rclone.org/local/#windows-paths)
+non local paths must use `/`. See [local filesystem](https://rclone.org/local/#paths-on-windows)
 documentation for more about Windows-specific paths.
 
 These paths needn't start with a leading `/` - if they don't then they
--- a/docs/content/local.md
+++ b/docs/content/local.md
@@ -114,9 +114,9 @@ as they can't be converted to UTF-16.
 ### Paths on Windows ###
 
 On Windows there are many ways of specifying a path to a file system resource.
-Both absolute paths like `C:\path\to\wherever`, and relative paths like
-`..\wherever` can be used, and path separator can be either
-`\` (as in `C:\path\to\wherever`) or `/` (as in `C:/path/to/wherever`).
+Local paths can be absolute, like `C:\path\to\wherever`, or relative,
+like `..\wherever`. Network paths in UNC format, `\\server\share`, are also supported.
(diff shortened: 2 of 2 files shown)


docs/content/docs.md
```diff
@@ -183,6 +183,10 @@ refer to a different directory.  On these, paths without a leading `/`
 will refer to your "home" directory and paths with a leading `/` will
 refer to the root.
 
+Additional leading path separators will be ignored, which means a path
```
ncw, line 186: I'm not sure I want to encourage people to add `//`

At some point (for backends which support it like s3) that might actually be a different result.

ncw (commented):
Sorry missed this PR!

All find except for the // - see comment inline!

docs/content/docs.md
albertony: This was primary reason for the PR in the first place... 🤦  But was really just a question/suggestion, so I totally accept your explanation. 😄

albertony:
No worries. I will remove the // part, and then just merge the remaining. 👍

albertony:
@ncw: We should probably give a final answer to the questions in the linked forum thread:

> Q1. Is rclone then ignoring these or is this remote specific behaviour?

Q1 answer is that rclone in many cases trim off excess slashes, but it is depeding on the remote types.

> Q2. In context of s3 and and google cloud storage remote types can one rely that gs://seldon-models/sklearn/iris or s3://seldon-models/sklearn/iris will keep working when defined with leading slashes and it is just not side effect of current implementation?

Q2 answer is that OP should **not** rely on this for S3 and GCS?

ncw:
> No worries. I will remove the // part, and then just merge the remaining.

Perfect - go for it :-)

> We should probably give a final answer to the questions in the linked forum thread:
> 
> > Q1. Is rclone then ignoring these or is this remote specific behaviour?
> 
> Q1 answer is that rclone in many cases trim off excess slashes, but it is depeding on the remote types.

:+1:
 
> > Q2. In context of s3 and and google cloud storage remote types can one rely that gs://seldon-models/sklearn/iris or s3://seldon-models/sklearn/iris will keep working when defined with leading slashes and it is just not side effect of current implementation?
> 
> Q2 answer is that OP should **not** rely on this for S3 and GCS?

I think that is correct too.

Current implementation allows it but it may change in the future.

PR #5317, opened 2021-05-08
vfs: avoid unnecessary subdir in cache path

### LINKED ISSUE #5316: Extra directories are created in cache folder!
Hello, 
OS Windows
rclone ver. >rclone-v1.54.1-windows-amd64
Question: 
Why did you add {} to the name of the goal in the cache folder, now extra folders are created?

for example: 
rclone.conf
[Book]
type = local

[BookY]
type = crypt
remote = Book:B:\b
filename_encryption = obfuscate
directory_name_encryption = false
password = ***

Cache folder: D:\rc\
was: D:\rc\vfs\BookY\
became: D:\rc\vfs\BookY{Db\Y9}\
directory structure:
D:\rc\vfs\BookY{Db\
                                       Y9}\
That is, extra directories are created in cache folder!

albertony:

Avoid sometimes creating additional subdirectory in cache folder. It may happen as a result of "detected overridden config - adding ... suffix to name", where the suffix is base64 encoded MD5, when the encoded suffix happens to include the `/` character - which is in the base64 "alphabet".

Edit: Is backwards compatibility an issue here? That there may be an existing cache directory `cache`\\`path`\\`{Db`\\`Y9}` with content, and after this change rclone will not find the existing cache and begin with a new cache directory `cache`\\`path`\\`{Db_Y9}`? And then, any changes still in the old cache that has not been written back to remote will be lost? 🤔 

Fixes #5316

#### Checklist

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1419,7 +1419,7 @@ func NewFs(ctx context.Context, path string) (Fs, error) {
 		extraConfig := overridden.String()
 		//Debugf(nil, "detected overriden config %q", extraConfig)
 		md5sumBinary := md5.Sum([]byte(extraConfig))
-		suffix := base64.RawStdEncoding.EncodeToString(md5sumBinary[:])
+		suffix := base64.RawURLEncoding.EncodeToString(md5sumBinary[:])
 		// 5 characters length is 5*6 = 30 bits of base64
 		const maxLength = 5
 		if len(suffix) > maxLength {


fs/fs.go
```diff
@@ -1419,7 +1419,7 @@ func NewFs(ctx context.Context, path string) (Fs, error) {
 		extraConfig := overridden.String()
 		//Debugf(nil, "detected overriden config %q", extraConfig)
 		md5sumBinary := md5.Sum([]byte(extraConfig))
-		suffix := base64.RawStdEncoding.EncodeToString(md5sumBinary[:])
+		suffix := strings.Replace(base64.RawStdEncoding.EncodeToString(md5sumBinary[:]), "/", "_", -1)
```
ncw, line 1422: I think using `base64.RawURLEncoding` is a better fix here - it uses `-` and `_` as its extra chars so works in filenames.

ncw (approved):
Thank you - that looks great now :-)

Merge when ready!

Issue #3453, opened 2019-08-18
Bucket Based Remotes: empty directories


yparitcher:
#### Progress
1. error on empty directory
- [ ] azureblob
- [ ] b2
- [x] gcs
- [ ] qingstor
- [x] s3
- [ ] swift

#### Current Situation:
Rclone does not officially support empty directories on BBR, however as many parts of the rclone api use file based systems there are many edge cases where rclone acts as if empty directories exist on BBR. 
this causes non existent paths to be validated as rclone cant tell the difference between them and empty directories. this can cause inconsistencies with `mount` etc.
this has been discussed before, see #3427 #3415

#### Proposed solutions:

**1.  enforce non support of empty directories on BBR:**
- make empty directories be an invalid path
- must fix list commands
- this will need work to make consistent with `mount` and similar commands which expect a filesystem (empty directories)

**2. support empty directories on BBR via a hidden empty file:**

- this is already done by some clients on other remotes (B2 web client, some S3 clients)
- need standardization to ensure that rclone uses the same files as others
- require refactoring of many BBR to include this logic
- cause some confusion about these files, but make filesystem commands (`mount`) work better
- requires more api calls - more expensive

**3. 1 & 2 via a flag:**

- this give the most options but will be slightly more confusing to use and require duplicate codepaths

---
if i missed anything please comment and i will add them in.

this will be a significant decision. rclone should have an policy which will help unified implementations for different backends.   

i do not mind helping will this but it will definitely require help from many people.
thanks to all those that have made these remotes work until now
@ncw Please review

ncw:
I think we should probably do 1. first - no flag needed.

As for 2. rclone mostly just contents itself with ignoring the 0 length directory files other programs make.  Making rclone actively put them in would be quite a lot of work so if we want to do it, should probably be in phase 2.  I think it needs a backend flag so you can set it on a per backend basis.

Unifying code between BBR backends will be important!

There are 6 BBR b2, s3, gcs, swift, qingstor, azureblob.  It is possible to get free accounts for all of those except for gcs and azureblob. (you can test s3 against minio or alibabacloud).  gcs has a free tier and azureblob may too.  Signing up for free services will probably mean putting your credit card in though!

I'd love to have your help with this :-)

Hermain:
https://privacy.com/ can be used to generate a prepaid-card for free and just put a minimal amount of money on there. This way you make sure you'll never be overcharged since its a prepaid card...

yparitcher:
[...]

PR #5323, opened 2021-05-13
Emtpy directory markers for S3 and GCS backends

Jancis:

Creates directory markers for Bucket Based Remotes to allow having empty folders on remote.

Usage: Adding `--s3-directory-markers` or `--gcs-directory-markers` flag will create empty file `.dirfile` in new folders. The directory marker file (`.dirfile`) is removed from file listing.
 
Related issue: https://github.com/rclone/rclone/issues/3453

#### Checklist


ncw:
This isn't the accepted way of doing this.

The normal way is to create a zero sized object which ends in a /

These are already filtered out of directory listings.

Can you change the pr to do that? Thank you

Jancis:
Is there an example backend or other product where that method actually works? I noticed something similar in the codebase, but either i don't know how to enable it or it just does not work.

ncw:
> Is there an example backend or other product where that method actually works? I noticed something similar in the codebase, but either i don't know how to enable it or it just does not work.

I know Cyberduck makes them so it would be great to make rclone compatible with that.

ivandeex:
Please run `gofmt -w ./backend/s3/s3.go` and commit its formatting changes. This will fix the unit test.

Jancis:
Thank You for checking the tests, @ivandeex, i'll fix them. 

@ncw, I noticed that the file listing for directory objects is already implemented in rclone, but the part that creates those files is not. I replaced the directory marker filename with `/` and it works as expected now, but the directory removal part is broken since the trailing slash is trimmed off the path when `Remove()` is called. I'll fix the removal part and get back.

Jancis:
This is up for reviewing again. Seems to work as expected on S3/Minio and GCS

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -248,6 +248,18 @@ Docs: https://cloud.google.com/storage/docs/bucket-policy-only
...
+			Name:     "directory_markers",
+			Default:  false,
+			Advanced: true,
+			Help: `Upload empty object with a trailing slash in name when new directory is created
+
+Empty folders are unsupported for bucket based remotes, so we create an empty 
+object named "/", to persist this folder. The directory marker object 
+is filtered from file listing.
+
+See: https://github.com/rclone/rclone/issues/3453
```
ncw, line 261: I don't think we need a ref to the issue here

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -798,6 +845,26 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) (err error) {
 // Returns an error if it isn't empty: Error 409: The bucket you tried
 // to delete was not empty.
 func (f *Fs) Rmdir(ctx context.Context, dir string) (err error) {
+	// Add trailing slash to folder path
+	dirFilePath := fmt.Sprintf("%s/", dir)
+	if len(f.rootDirectory) != 0 {
+		dirFilePath = fmt.Sprintf("%s/%s", f.rootDirectory, dirFilePath)
+	}
+	// Query directory marker object before attempting to remove it.
```
ncw, line 853: Why not just remove it to save an API call. It might return not found in which case you can ignore the error.

backend/s3/s3.go
```diff
@@ -1266,6 +1266,18 @@ disabled here.  When the issue is solved this flag will be removed.
...
+			Name:     "directory_markers",
+			Default:  false,
+			Advanced: true,
+			Help: `Upload empty object with a trailing slash in name when new directory is created
+
+Empty folders are unsupported for bucket based remotes, so we create an empty 
+object named "/", to persist this folder. The directory marker object 
+is filtered from file listing.
+
+See: https://github.com/rclone/rclone/issues/3453
```
ncw, line 1280: see above

backend/s3/s3.go
```diff
@@ -2137,7 +2150,50 @@ func (f *Fs) bucketExists(ctx context.Context, bucket string) (bool, error) {
...
+		return f.uploadDirFile(ctx, bucket, dir)
+	}
+	return nil
+}
+
+// Creates empty file as directory marker
+func (f *Fs) uploadDirFile(ctx context.Context, bucket string, dir string) (err error) {
+	dirFilePath := fmt.Sprintf("%s/", dir)
+	fs.Debugf(nil, "Creating directory object %v", dirFilePath)
+	req := s3.PutObjectInput{
```
ncw, line 2167: Unfortunate duplication here...

Can you not just call Object.Update with a constructed Object?

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -743,7 +756,41 @@ func (f *Fs) PutStream(ctx context.Context, in io.Reader, src fs.ObjectInfo, opt
...
+		return e
+	}
+	if f.opt.DirectoryMarkers {
+		return f.uploadDirFile(ctx, bucket, dir)
+	}
+	return nil
+}
+
+// Creates empty file as directory marker
+func (f *Fs) uploadDirFile(ctx context.Context, bucket string, dir string) (err error) {
```
ncw, line 770: Can you call Object.Update here instead of duplicating the upload code?

backend/s3/s3.go
```diff
@@ -2177,6 +2233,39 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) error {
...
+		Key:    &dirFilePath,
+	})
+	err := f.pacer.Call(func() (bool, error) {
+		var err error
+		httpReq.HTTPRequest = httpReq.HTTPRequest.WithContext(ctx)
+		err = httpReq.Send()
+		return f.shouldRetry(ctx, err)
+	})
+	if err == nil {
+		// Using API call due to trailing slash being removed before Remove() can get full path.
```
ncw, line 2253: It would be better to call Object.Remove here rather than duplicating the code if possible.

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -798,6 +845,26 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) (err error) {
...
+	if len(f.rootDirectory) != 0 {
+		dirFilePath = fmt.Sprintf("%s/%s", f.rootDirectory, dirFilePath)
+	}
+	// Query directory marker object before attempting to remove it.
+	e := f.pacer.Call(func() (bool, error) {
+		_, e := f.svc.Objects.Get(f.rootBucket, dirFilePath).Context(ctx).Do()
+		return shouldRetry(ctx, e)
+	})
+	if e == nil {
+		// Using API directly due to trailing slash being removed before Remove() can get full path
```
ncw, line 859: I'd prefer to call Object.Remove here or something factored out of Object.Remove rather than duplicate this.

ncw (commented):
I think this is looking good :-)

I put some comments inline - mostly about trying to use the existing Upload file/Delete file interfaces (factor them if necessary) rather than making your own.

This also needs an integration test - add a `TestIntegration2` in the *_test.go files, see [backend/sftp/sftp_test.go](https://github.com/rclone/rclone/blob/master/backend/sftp/sftp_test.go) for an example and opt.ExtraConfig for setting the extra config item.

Then see if it passes [the integration tests](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend)!

backend/googlecloudstorage/googlecloudstorage.go
Jancis: Both Upload and Remove are expanding bucket and bucketPath from o.split() that looses the directory marker filename (`/`) after it's run trough `path.Join(f.root, rootRelativePath)`. Basically, it trims the trailing slash that would be the "filename" of that marker.
See this playground example: https://play.golang.org/p/m8Y1OyvKw1f
Due to this, custom functions are used. I could try replacing `bucket, bucketPath := o.split()`, runing tests and checking if they still pass. The other option is still keeping the duplicate functions (that I'd love to avoid too).

Jancis: This is the best i could do for split() 
```
// path.Join might stripped trailing slash that can be used as an object name for bucket based remotes.
if strings.HasSuffix(rootRelativePath, "/") {
	bucketPath += "/"
}
```

backend/googlecloudstorage/googlecloudstorage.go
Jancis: Changed; Done

backend/googlecloudstorage/googlecloudstorage.go
Jancis: Removed, return code ignored; Done

backend/s3/s3.go
Jancis: Changed; Done

backend/s3/s3.go
Jancis: This required o.split change to reattach trailing slash and an extra metadata header; Done

backend/googlecloudstorage/googlecloudstorage.go
Jancis: Also done, required o.split change

backend/s3/s3.go
Jancis: Done, indeed looks better. This one missed `DirectoryMarkers` conditional that caused them always to be removed.

backend/googlecloudstorage/googlecloudstorage.go
Jancis: The snippet above did it, This works now.

Jancis:
Suggested fixes done, custom API calls removed in favor of `Put` and `Remove` methods. The fix was adjusting `split` method to reattach the trailing slash if it was there before passing trough `path.Join(f.root, rootRelativePath)`.
Took a while to get back to this, figure out tests and make relevant backend configurations for local tests, but it was well worth it as it caught one or two mistakes (i.e. s3 Rmdir didn't have `DirectoryMarker` conditional for the marker file removal).
All tests seem to pass in local now, Minio tests complain about `TiersToTest`, but AWS/S3 works as expected.
I even tried enabling `CanHaveEmptyDirectories` feature for both backends, but then some errors popped up.
There is one thing I noticed, though - I might be doing something wrong, but tests don't seem to obey `ExtraConfig`, so i had to explicitly set option defaults value to true in code while running tests.
This does not seem to do anything -
```
ExtraConfig: []fstests.ExtraConfigItem{
	{Name: name, Key: "directory_markers", Value: "true"},
},
```

Jancis:
Forgot to run the coding standards check `gofmt -w`, fix commited.

backend/s3/s3.go
```diff
@@ -2178,6 +2213,18 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) error {
...
+	if f.opt.DirectoryMarkers && bucket != "" && directory != "" {
+		markerFilePath := fmt.Sprintf("%s/", directory)
+		if f.rootDirectory != "" {
+			markerFilePath = fmt.Sprintf("%s/%s", f.rootDirectory, markerFilePath)
+		}
+		markerFileObject := &Object{
+			fs:     f,
+			remote: markerFilePath,
+		}
+		markerFileObject.Remove(ctx)
```
ivandeex, line 2226: [linter wants you to](https://github.com/rclone/rclone/runs/3263796986?check_suite_focus=true#step:14:23)
```
_ = markerFileObject.Remove(ctx)
```

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -799,6 +831,18 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) (err error) {
...
+	if f.opt.DirectoryMarkers && bucket != "" && directory != "" {
+		markerFilePath := fmt.Sprintf("%s/", directory)
+		if f.rootDirectory != "" {
+			markerFilePath = fmt.Sprintf("%s/%s", f.rootDirectory, markerFilePath)
+		}
+		markerFileObject := &Object{
+			fs:     f,
+			remote: markerFilePath,
+		}
+		markerFileObject.Remove(ctx)
```
ivandeex, line 844: [linter wants you](https://github.com/rclone/rclone/runs/3263796986?check_suite_focus=true#step:14:23) to:
```
_ = markerFileObject.Remove(ctx)
```

backend/googlecloudstorage/googlecloudstorage_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/rclone/rclone/backend/googlecloudstorage"
+	"github.com/rclone/rclone/fstest"
```
ivandeex, line 9: [linter wants you](https://github.com/rclone/rclone/runs/3263796986?check_suite_focus=true#step:14:23) to:
```
gofmt -w  ./backend/googlecloudstorage/googlecloudstorage_test.go
git commit -a -m 'gratify golang linter'
git push
```

ivandeex (commented):
go linter wants you to do something...

Jancis:
Of course I forgot to lint the test files themselves.. 
Is weird that it did not complain about `markerFileObject.Remove(ctx)` though.

ivandeex:
please rebase on current master

Jancis:
:checkered_flag: Branch rebased on master, linter test rerun.

ivandeex:
@Jancis thank you

@ncw your turn

ncw (commented):
Apologies for the delay in reviewing

I think this is almost ready to go - looking really good :-)

> Suggested fixes done, custom API calls removed in favor of `Put` and `Remove` methods. The fix was adjusting `split` method to reattach the trailing slash if it was there before passing trough `path.Join(f.root, rootRelativePath)`.
> 
> Took a while to get back to this, figure out tests and make relevant backend configurations for local tests, but it was well worth it as it caught one or two mistakes (i.e. s3 Rmdir didn't have `DirectoryMarker` conditional for the marker file removal).

Great
 
> All tests seem to pass in local now, Minio tests complain about `TiersToTest`, but AWS/S3 works as expected.
> I even tried enabling `CanHaveEmptyDirectories` feature for both backends, but then some errors popped up.

I think your patch should enable `CanHaveEmptyDirectories` otherwise you get this error

```
$ rclone mkdir s3,directory_markers:rclone/empty_directory
2021/09/04 12:07:42 NOTICE: S3 bucket rclone path empty_directory: Warning: running mkdir on a remote which can't have empty directories does nothing
```

> There is one thing I noticed, though - I might be doing something wrong, but tests don't seem to obey `ExtraConfig`, so i had to explicitly set option defaults value to true in code while running tests.
> This does not seem to do anything -
> 
> ```
> ExtraConfig: []fstests.ExtraConfigItem{
> 	{Name: name, Key: "directory_markers", Value: "true"},
> },
> ```

The `name` needs to be the same as the remote name. So a patch like this would fix S3, similar for GCS

```patch
--- a/backend/s3/s3_test.go
+++ b/backend/s3/s3_test.go
@@ -25,9 +25,9 @@ func TestIntegration2(t *testing.T) {
 	if *fstest.RemoteName != "" {
 		t.Skip("skipping as -remote is set")
 	}
-	name := "TestDirectoryMarkers"
+	name := "TestS3"
 	fstests.Run(t, &fstests.Opt{
-		RemoteName:  "TestS3:",
+		RemoteName:  name + ":",
 		NilObject:   (*Object)(nil),
 		TiersToTest: []string{"STANDARD", "STANDARD_IA"},
 		ChunkedUpload: fstests.ChunkedUploadConfig{
```

So in summary I think this needs:
- Setting of `CanHaveEmptyDirectories` Feature flag is `directory_markers` is set
- Fixing of integration test parameters so they work

Are you happy with this being squashed into 1 commit for merge to master or would you like more than one?

Jancis:
I'm struggling with `CanHaveEmptyDirectories`, can't pass the `FsMkdirRmdirSubdir` test.
`go test -run=TestIntegration2/FsMkdir/FsMkdirRmdirSubdir -v` fails with 
```
expected: []string{"dir", "dir/subdir"}
actual  : []string{}
```
I tried to debug and see what other BBR's are doing about it, but i start to suspect they have similar issue (did not check each and every one of them, but at least swift backend fails exactly the same test when i enable that feature in it's code.

I'm stuck with this one, can we have it off for now?

Current summary:
- _Setting of CanHaveEmptyDirectories Feature flag if directory_markers is set._ Failed
- _Fixing of integration test parameters so they work._ Done
- _Are you happy with this being squashed into 1 commit for merge to master or would you like more than one?_ I'm absolutely fine with that, no need to keep the history.

Jancis:
Rebased on latest master to fix conflicts with `DownloadURL` changes.

aymanbagabas:
just tested this, would love to see it merged <3

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -681,9 +697,15 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 			}
 			// is this a directory marker?
 			if isDirectory {
-				continue // skip directory marker
+				// Don't insert the root directory
+				if remote == directory {
+					continue
+				}
+				// process directory markers as directories
+				remote = strings.TrimRight(remote, "/")
```
joec4i, line 705: @ncw  This is causing `rclone lsd -R` to throw out the following errors when encountering directory markers:
```
2023/05/26 22:02:24 ERROR : foo: Entry doesn't belong in directory "foo" (too short) - ignoring
           0 2023-05-26 22:02:24        -1 foo/bar
           0 2023-05-26 22:02:24        -1 foo/baz
2023/05/26 22:02:24 ERROR : httpdocs: Entry doesn't belong in directory "httpdocs" (too short) - ignoring
           0 2023-05-26 22:02:24        -1 httpdocs/admin
           0 2023-05-26 22:02:24        -1 httpdocs/bar
           0 2023-05-26 22:02:24        -1 httpdocs/cache
```

In the above example, `foo/` and `httpdocs/` are both GCS directory markers. They fall into https://github.com/rclone/rclone/blob/e43b5ce/fs/list/list.go#L79 : `prefix` is set to `foo/` when the `remote` is `foo`. 

Also, if the root directory is a directory marker, it falls into https://github.com/rclone/rclone/blob/e43b5ce/fs/list/list.go#L82 ,  triggering an error like this:
```
2023/05/26 22:08:30 ERROR : : Entry doesn't belong in directory "" (same as directory) - ignoring
```

I think we need to handle directory markers properly in `filterAndSortDir()`. 

A possible solution is to prepend `/` to every entry, including directory markers, e.g.
```
/httpdocs (directory-marker)
/httpdocs/admin (non directory-marker subdir)
/httpdocs/cache (non directory-marker subdir)
```
When running `rclone lsd -r httpdocs`, the output would be:
```
/ (directory-marker)
/admin (non directory-marker subdir)
/cache (non directory-marker subdir)
```

Alternatively, we could show the entries as follows:
```
httpdocs/
httpdocs/admin/
httpdocs/cache/
```
and
```
/
admin/
cache/
```

Thoughts or suggestions? 🙏 

Note: I added this comment to the original commit but I realised that it's probably not visible. So moving it here. Hope it's okay.

Issue #3353, opened 2019-07-16
1fichier: Support downloading password-protected files in shared folders


treyturner:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.48.0-067-g0a1169e6-beta
- os/arch: linux/amd64
- go version: go1.12.7

#### What problem are you are trying to solve?
I'm trying to copy files out of a 1fichier shared folder that is not my own, where each file is protected with the same password.

#### How do you think rclone should be changed to solve that?
The 1fichier API seems to allow for the password to be sent along with the request, but the rclone backend for 1fichier does not yet seem to support this, which isn't surprising since 1fichier support is very new (thanks for this!)

From https://1fichier.com/api.html?lg=en:

```
{
  "url" : "https://1fichier.com/?example",
  "inline" : 0/1,
  "cdn" : 0/1,
  "restrict_ip" : 0/1/2,
  "single" : 0/1,
  "pass" : "mot_de_pass",
  "no_ssl" : 0/1,
  "folder_id" : 1234,
  "filename" : "FileName",
  "sharing_user" : "user@example.com"
}
```

```
pass : (optional) Access password to the file, if restricted by password.
```

My suggestion is that a password could be supplied via a `--fichier-shared-password` flag, similar to `--fichier-shared-folder`.

Thanks for your consideration!

ncw:
What do you think @lhaus fancy having a go at this?

A `--fichier-shared-password` flag seems perfect!  This could be an advanced option since I don't think most users will want to set it.

Pfuenzle:
Are there any news regardin this topic?
Im trying to download files from a 1fichier shared folder, but the files now require a password.
When I try to copy, I get this error:
`(error HTTP error 403 (403 Forbidden) returned body: "{\"status\":\"KO\",\"message\":\"Password not provided. Resource not allowed #749\"}")`
Rclone then retries the connection 7 times, after that following errer gets returned:
`(error HTTP error 403 (403 Forbidden) returned body: "{\"message\":\"IP Locked #78\",\"status\":\"KO\"}")`

Kcchouette:
I think this ticket can be close as the PR #5333 has been merged @ivandeex

buengese:
Looks like I did a manual squash and merge there. It's this commit 92b3518c78848d454a7352707fa7c849fd93403a

ivandeex:
@buengese Got you. Thanks
@Kcchouette You are right, it was merged. Closing.

PR #5333, opened 2021-05-16
Support downloading password protected files with 1Fichier backend

hkr:

Add a new advanced option `shared_password` to support accessing files in shared folders that are protected by a password.

https://github.com/rclone/rclone/issues/3353

#### Checklist

### DIFF
--- a/backend/fichier/api.go
+++ b/backend/fichier/api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"net/http"
+	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
@@ -123,6 +124,12 @@ func (f *Fs) listSharedFiles(ctx context.Context, id string) (entries fs.DirEntr
 		RootURL:    "https://1fichier.com/dir/",
 		Path:       id,
 		Parameters: map[string][]string{"json": {"1"}},
+	}
+	if f.opt.SharedPassword != "" {
+		opts.Method = "POST"
+		opts.Body = strings.NewReader("json=1&pass=" + url.QueryEscape(f.opt.SharedPassword))
+		opts.ContentType = "application/x-www-form-urlencoded"
+		opts.Parameters = nil
 	}
 
 	var sharedFiles SharedFolderResponse
--- a/backend/fichier/api.go
+++ b/backend/fichier/api.go
@@ -90,6 +90,7 @@ func (f *Fs) getDownloadToken(ctx context.Context, url string) (*GetTokenRespons
 	request := DownloadRequest{
 		URL:    url,
 		Single: 1,
+		Pass: f.opt.SharedPassword,
 	}
 	opts := rest.Opts{
 		Method: "POST",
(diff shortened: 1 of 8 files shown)


backend/fichier/fichier.go
```diff
@@ -44,6 +44,11 @@ func init() {
 			Name:     "shared_folder",
 			Required: false,
 			Advanced: true,
+		}, {
+			Help:     "If you want to download a shared file that is password protected, add this parameter",
+			Name:     "shared_password",
+			Required: false,
+			Advanced: true,
```
ncw, line 51: This should probably say `IsPassword: true,` here - we are entering a password aren't we. This will mean it is stored obfuscated in the config and will need to be revealed.

backend/fichier/api.go
```diff
@@ -90,6 +90,7 @@ func (f *Fs) getDownloadToken(ctx context.Context, url string) (*GetTokenRespons
 	request := DownloadRequest{
 		URL:    url,
 		Single: 1,
+		Pass: f.opt.SharedPassword,
```
ncw, line 93: This will be passing empty string when SharedPassword is not set - is that OK?

ncw (commented):
This looks useful thanks :-)

Can you drop the doc changes from the commit please? So rclone.1 MANUAL.* - these get made as part of the release.

ncw:
@buengese any thoughts on this?

backend/fichier/api.go
hkr: I didn't have any problems with that.
However I made another change and added `omitempty`.
I have not tried to set a password using the API but in the web UI an empty string removes the password.

hkr:
I noticed that 1Fichier also has folder passwords. Currently 'list' does not work on those.
Any opinion if I should use the same `shared_password` option also for folder password or is it better to add a new option?

buengese:
This looks mostly good now.
Regarding folder passwords: In theory file and folder password are independent, so this should be a different option. In reality, I don't think I've ever actually seen a folder password, and I'm not even sure where you would set them.

hkr:
https://github.com/hkr/rclone/commit/30facdac89a1f3b8c4809cb0d606b7d4fd4fea34 worked for me for folder password. It's not part of this PR

ivandeex:
Code quality test failed. Please run `gofmt -w ./backend/fichier/*.go` and commit the result.

hkr:
> This looks mostly good now.
> Regarding folder passwords: In theory file and folder password are independent, so this should be a different option. In reality, I don't think I've ever actually seen a folder password, and I'm not even sure where you would set them.

You can set them in the Web UI similar to file password. Click a folder and then click the lock icon.

buengese:
>You can set them in the Web UI similar to file password. Click a folder and then click the lock icon.
Thanks. That's what I meant. I had so far missed that option. Since it's possible to have files with a different password in a password-protected folder it' would make sense to have separate parameters for file and folder passwords.

hkr:
I renamed `shared_password` to `file_password`, merged the changes to list files of a password protected folder and added a new option `folder_password`

buengese:
I've merged this now sorry for the delay.

ivandeex:
Merged by @buengese in https://github.com/rclone/rclone/commit/92b3518c78848d454a7352707fa7c849fd93403a

PR #5341, opened 2021-05-20
config: fixed issues with flags and options set by environment variables

olefrost:

Fixing issues related to the use of environment variables.

Adding debug logging of flags and options set by environment variables.

Detailing the documentation of config option precedence.

The changes are explained in the following sections. 

#### Issue1: Some env vars didn’t behave like the command line flags.

_Example and impact:_

    RCLONE_STATS='10s'
    rclone check remote: remote: --progress
    # Expected behaviour: rclone check remote: remote: --progress --stats=10s
    # Actual behaviour: rclone check remote: remote: --progress

The affected flags were: --stats, --log-level, --separator, --multi-thread-streams, --rc-addr, --rc-user and --rc-pass.

_Root cause and fix:_

This line
https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config/flags/flags.go#L24
sets the flag value directly, without updating the entire flag status such as flag.Changed. This caused unexpected behavior in code checking flag.Changed.

It is fixed by replacing “flag.Value.Set(newValue)” with “flags.Set(name, newValue)”.

#### Issue2: Backend env vars overruled remote specific env vars.

_Example and impact:_

    RCLONE_DRIVE_USE_TRASH='false'
    RCLONE_CONFIG_MYDRIVE_USE_TRASH='true'
    rclone deletefile myDrive:my-test-file
    # Expected result: my-test-file is recoverable in the trash folder
    # Actual result: my-test-file is permanently deleted (not recoverable)

The issue is reproduced in [this forum post](https://forum.rclone.org/t/example-to-hit-func-oev-optionenvvars-get-key-string/24150/7).

It affects all [backend options](https://rclone.org/flags/#backend-flags).

_Root cause and fix:_

The line 
https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/cmd/cmd.go#L542
was intended to create a backend flag with a possible default from an env var, but it actually sets both the value and the default from the env var. This gives the value precedence like command line values. This causes the env var to override other more specific env vars.

This happens because [flags.VarPF(…)](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config/flags/flags.go#L161-L166) calls [setDefaultFromEnv(…)](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config/flags/flags.go#L14-L31) that sets both [the value](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config/flags/flags.go#L24) and [the default](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config/flags/flags.go#L29) from an env var if possible.

The issue was introduced in v1.54 by this commit: https://github.com/rclone/rclone/commit/72083c65adff1d563a0f9ba92e15c2a837c420c8

The issue has been fixed by 

- Calling pflag.CommandLine.VarPF(…) directly without the (command line) wrapper 
- Creating a new func SetDefaultFromEnv(…) – only to be used when creating backend flags
- Renaming func setDefaultFromEnv(…) to setValueFromEnv(…) – to avoid future mistakes.
- Deleting func flags.VarPF(…)  - to avoid confusion with the wrappers for command line flags

#### Issue3: Backend generic env vars overruled backend specific env vars

_Example and impact:_

    RCLONE_SKIP_LINKS='true'
    RCLONE_LOCAL_SKIP_LINKS='false'
    rclone lsd local:
    # Expected result: Warnings when symlinks are skipped
    # Actual result: No warnings when symlinks are skipped (RCLONE_SKIP_LINKS takes precedence)

The issue is reproduced in [this forum post](https://forum.rclone.org/t/example-to-hit-func-oev-optionenvvars-get-key-string/24150/3).

_Root cause and fix:_

This code 
https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/fs.go#L1306-L1313
checks the generic option before the specific option. This causes the generic option to take precedence.

Fixed by checking the specific (=prefixed) option first.

#### Enhance1: Debug logging when options are set by environment variables.

Discussed in multiple forum posts, e.g. [this ](https://forum.rclone.org/t/how-can-i-check-if-environment-variables-are-being-used/17490)and [this](https://forum.rclone.org/t/gdrive-settings-environment-variables-not-mapped/19603).

Fixed by
- Scanning command line and environment in [func NewConfig()](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/config.go#L137-L141) to set LogLevelDebug preliminarily before flags are being initialised and possibly read from the environment.
- Adding and streamlining debug logging whenever environment variables are being read.
- Disabling a very frequent [`rc` debug message](https://github.com/rclone/rclone/blob/07f2f3a62e5c0461858cbd76cc04e59d49fab02e/fs/rc/registry.go#L48) during initialisation.
- Adding tips about debug logging of environment variables to the documentation.

#### Enhance2: Improved the documentation of config option precedence.

Inspired by [this forum post](https://forum.rclone.org/t/example-to-hit-func-oev-optionenvvars-get-key-string/24150/9).

#### Checklist


ncw:
I've given github actions another poke...

It would be really nice to have a test to test the config priorities - can you make one?

How many existing configs will this break do you think?

olefrost:
> I've given github actions another poke...

Thanks, all green :)
 
> It would be really nice to have a test to test the config priorities - can you make one?

Good idea, I will take a look and report back. Let me know if you have a good example to start from.

> How many existing configs will this break do you think?

None, there are few support questions/issues on env vars, so don't think it is that much used and I doubt many if any have a setup where they depend on the precedence between the different levels of env vars. If so, they would have complained about the bugs already and insisted on having more debug info to find out was was happening. 

I have a good understanding of the changes I made in the backend functions, and have debugged/tested the entire precedence from top to bottom.

To me the biggest risk lies in an unforeseen consequence of the fix to issue1, something that worked before and now breaks due to something I didn't understand about the pflag package or it's usage in rclone. It would therefore great if you scrutiny that change and possibly do some extra tests in the areas that you consider most used.

A quick test of the affected rc options (using env vars) may also be a good idea. I have no experience with rc. The affected options are --rc-addr, --rc-user and --rc-pass (based on code searching).

olefrost:
> It would be really nice to have a test to test the config priorities - can you make one?

I have looked at the tests and is unable to find a good example to start from. 

I miss something that will show me how to convert a launch configuration like this

	{
		"name": "rclone test priority2 - Expect: Symlink warnings",
		"type": "go",
		"request": "launch",
		"mode": "auto",
		"program": "${workspaceFolder:rclone}/rclone.go",
		"args": ["lsd","myLocal,skip_links=false:/folder/with/symlinks","-vv","--skip-links"],
		"env": {
			"rclone_config_mylocal_skip_links":"true",
			"rclone_local_skip_links":"true",
			"rclone_skip_links":"true"
		}
	},

into a go/rclone test with a mockFolder of type Local named myLocal and an assert on the expected output (symlink warnings).

ncw:
> > It would be really nice to have a test to test the config priorities - can you make one?
> 
> I have looked at the tests and I am unable to find a good example to start from.

Hmm, yes, now I think about it this is quite hard!

To do a full end to end test would require quite a bit of cunning - it would need a test which execs itself with different command line flags and environment variables and checks they get set as expected. [This article](https://www.joeshaw.org/testing-with-os-exec-and-testmain/) contains a great example. In this case we'd make the test print the value of a variable, then we could exec the test code with different combinations of environment variables, command line flags, etc and check the variable was as we expected.

Not sure exactly where this test code should go - perhaps in `fs/fs_test.go`

Does that make sense?

olefrost:
> Does that make sense?

Yes, it does, thanks!

I went all the way and made an end-to-end test framework where it is possible to test against the normal output and/or the output from DEBUG logging. It was then relatively easy to make a general test of environment variables, both non-backend and backend - including precedence etc.

The end-to-end test framework is placed in cmdtest. The name is inspired by fstest (unit/backend tests) and chosen to signal test of cmd’s (incl. flags and environment). The separate folder and package were necessary to avoid circular imports. Technically, the ideal placement would be in the root folder together with rclone main, but I prefer the root being as clean as possible. The current placement also goes well with the actual main code being in cmd.go. (On a side note: You may want to (re)move notes.txt in the root)

My tests revealed a twist/consequence of connection strings (incl. remote specific environment vars) when using multilevel remotes like alias, crypt etc. I guess others would be somewhat surprised like me and therefore added a note/clarification to the documentation.

Known GitHub integration test issues to be discussed:
* There seems to be a bug in rclone on linux. It doesn’t print the date when using RCLONE_LOG_FORMAT=”date,shortfile”
*  Linux and Mac throw a permission denied when trying to create a test file. Can you quickly spot the issue around line 119 of cmd_test.go? (I am Windows)

I hope you like it.

ivandeex:
@olefrost
Rebased on new master.
Please don't force-push now!
The pre-state of commits has been saved under [ivandeex/olefrost-fix-environment-flags](https://github.com/ivandeex/rclone/commits/olefrost-fix-environment-flags).

olefrost:
@ivandeex 
Thanks, it is stable waiting for review/merge :)

ncw:
@olefrost wrote:
> I went all the way and made an end-to-end test framework where it is possible to test against the normal output and/or the output from DEBUG logging. It was then relatively easy to make a general test of environment variables, both non-backend and backend - including precedence etc.

A very nice piece of work - well done :-)

> My tests revealed a twist/consequence of connection strings (incl. remote specific environment vars) when using multilevel remotes like alias, crypt etc. I guess others would be somewhat surprised like me and therefore added a note/clarification to the documentation.

Worth while documenting definitely!

@ivandeex - thanks for fixing up after the fs.go refactor.

I'll merge this now in time for 1.56 - thank you!

PR #5342, opened 2021-05-20
Add selection to ncdu command

CrossR:

Add the ability to select multiple items in the ncdu interface to delete them.

Either press `v` to select the current item, or `V` to highlight items as you move over them (slightly awkward behaviour that I'll tidy up here if you swap direction, the final item isn't toggled back). Then I've added a `D` command that will delete all the currently selected items.

I've not discussed this elsewhere, it was just something I wanted!

I haven't fully fleshed this out yet (fixing the highlight on move feel, keybind changes/squash commits/docs/look for ncdu tests if they exist), since I wanted to check if this was something you'd be open to adding/if it fits as part of the `ncdu` command.

Assuming it is something you'd be fine to add, I'll fix up the things I mentioned above, as well as do a pass over the code. I've not written a line of go before this, so I expect I've done some things in odd, non-standard ways (whatever the equivalent of "Pythonic" is for Go!) so may need some refactoring.

Also just need to do a pass to make sure I'm not missing any edge cases (i.e. leaving a folder, making sure works as expected if a single selected file is deleted (it won't right now I think), fix cursor pos for multiple moves, etcetc).

This was just a "check a minimal case" pass, to make sure you were open to the idea first!

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -74,6 +74,9 @@ func helpText() (tr []string) {
 		" a toggle average size in directory",
 		" n,s,C,A sort by name,size,count,average size",
 		" d delete file/directory",
+		" v select file/directory",
+		" V enter visual select mode",
+		" D delete selected files/directories",
 	}
 	if !clipboard.Unsupported {
 		tr = append(tr, " y copy current path to clipboard")
@@ -107,11 +110,13 @@ type UI struct {
 	showGraph          bool          // toggle showing graph
 	showCounts         bool          // toggle showing counts
 	showDirAverageSize bool          // toggle average size
+	visualSelectMode   bool          // toggle visual selection mode
 	sortByName         int8          // +1 for normal, 0 for off, -1 for reverse
 	sortBySize         int8
 	sortByCount        int8
 	sortByAverageSize  int8
 	dirPosMap          map[string]dirPos // store for directory positions
+	selectedEntries    map[string]dirPos // selected entries of current directory
 }
 
 // Where we have got to in the directory listing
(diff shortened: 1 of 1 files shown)


ncw:
> Add the ability to select multiple items in the ncdu interface to delete them.

Sounds useful!

Is this something ncdu does?
 
> Either press `v` to select the current item, or `V` to highlight items as you move over them (slightly awkward behaviour that I'll tidy up here if you swap direction, the final item isn't toggled back). Then I've added a `D` command that will delete all the currently selected items.

I like `D` for delete marked items. `v` to mark them seems odd - what is the rational there? Can we use `m` for `Mark`? Shift arrow keys might work for multiple selections?

Is there prior art we can ~~steal~~ imitate so make it work like some other terminal app people are familiar with?

CrossR:
>Is this something ncdu does?

Its not (at least not in the versions I use), which is why I wanted to check it was something you'd be open to adding. I know some features, even when useful, don't necessarily make sense. (Perhaps this moves the `ncdu` command closer to being a basic file explorer, which isn't necessarily desirable).

> I like D for delete marked items. v to mark them seems odd - what is the rational there? Can we use m for Mark? Shift arrow keys might work for multiple selections?
>
>Is there prior art we can steal imitate so make it work like some other terminal app people are familiar with?

I added `V` first, then realised it would be useful to add a single select too, so added it as the non-capitalised version..... so no real reasoning, just that it was the lower case of the thing I just added.

Something like ranger (https://github.com/ranger/ranger) is where I usually go to _borrow_ keybinding ideas from, so in that case it would be `Space` to single select, and `V` for visual select (though it is vim inspired so maybe its just reinforcing the binds I already know from there, and there is a nicer more general thing to look at). Either space or m sound reasonable to me, and I think adding shift-arrow keys makes a lot of sense too.

Zirro:
I've found rclone's ncdu to be pretty handy as a makeshift file explorer in situations where the `mount`-command has not been available. It'd be unfortunate to limit its potential just for the sake of strictly mirroring the original.

One possibility would be to implement more features as an optional layer on top of the standard ncdu functionality, which could be toggled through a flag to the `ncdu` command depending on whether the user wants the extra features or prefers the "pure" ncdu experience.

Later on, assuming that additional non-ncdu features get implemented over time, these could potentially be forked into two separate commands - one which sticks to the original ncdu-features and one which fully embraces the role of file explorer.

ivandeex:
> I like `D` for delete marked items. `v` to mark them seems odd - what is the rational there? Can we use `m` for `Mark`? Shift arrow keys might work for multiple selections?

- [x] `v` looks like a check glyph so it's better than `m` IMO

SonaliBendre:
Hi, is it possible to add copy,sync and move besides delete? First select all files needed to copy then shortcut key to add them to task, go to destination directory, shortcut key for paste Done. a multi window feature in ncdu may also be helpful for file explore and copy/move tasks.

PR #5347, opened 2021-05-21
🕰️ GCS: Compatible with gsutil's mtime metadata

### LINKED ISSUE #5331: Google Cloud Storage: Compatibility with gsutil's modification time metadata

```
rclone v1.55.1
- os/type: windows
- os/arch: amd64
- go/version: go1.16.3
- go/linking: dynamic
- go/tags: cmount
```

The [gsutil rsync](https://cloud.google.com/storage/docs/gsutil/commands/rsync) command writes modification time as a separate metadata entry `goog-reserved-file-mtime`. Attempting to use `rclone sync` to sync files uploaded by `gsutil rsync` will result in metadata updates for every single file.

If `goog-reserved-file-mtime` exists in a file's metadata, use it as the modification time instead. This behavior can be opted out with an option.

#### How to use GitHub

database64128:

- GCS: Write `goog-reserved-file-mtime` in addition to `mtime`.
- GCS: Fallback to `goog-reserved-file-mtime` if `mtime` doesn't exist.
- Docs: Mention the new modification time behavior and the modify window issue.
- Docs: Unify markdown format.

Fixes #5331.

#### Checklist

### DIFF
--- a/backend/googlecloudstorage/googlecloudstorage.go
+++ b/backend/googlecloudstorage/googlecloudstorage.go
@@ -51,8 +51,7 @@ import (
 const (
 	rcloneClientID              = "202264815644.apps.googleusercontent.com"
 	rcloneEncryptedClientSecret = "Uj7C9jGfb9gmeaV70Lh058cNkWvepr-Es9sBm0zdgil7JaOWF1VySw"
-	timeFormatIn                = time.RFC3339
-	timeFormatOut               = time.RFC3339Nano
+	timeFormat                  = time.RFC3339Nano
 	metaMtime                   = "mtime"                    // key to store mtime in metadata
 	metaMtimeGsutil             = "goog-reserved-file-mtime" // key used by GSUtil to store mtime in metadata
 	listChunks                  = 1000                       // chunk size to read directory listings
@@ -922,7 +921,7 @@ func (o *Object) setMetaData(info *storage.Object) {
 	// read mtime out of metadata if available
 	mtimeString, ok := info.Metadata[metaMtime]
 	if ok {
-		modTime, err := time.Parse(timeFormatIn, mtimeString)
+		modTime, err := time.Parse(timeFormat, mtimeString)
 		if err == nil {
 			o.modTime = modTime
 			return
@@ -942,7 +941,7 @@ func (o *Object) setMetaData(info *storage.Object) {
 	}
 
 	// Fallback to the Updated time
-	modTime, err := time.Parse(timeFormatIn, info.Updated)
+	modTime, err := time.Parse(timeFormat, info.Updated)
 	if err != nil {
 		fs.Logf(o, "Bad time decode: %v", err)
 	} else {
@@ -999,7 +998,7 @@ func (o *Object) ModTime(ctx context.Context) time.Time {
 // Returns metadata for an object
 func metadataFromModTime(modTime time.Time) map[string]string {
 	metadata := make(map[string]string, 1)
-	metadata[metaMtime] = modTime.Format(timeFormatOut)
+	metadata[metaMtime] = modTime.Format(timeFormat)
 	metadata[metaMtimeGsutil] = strconv.FormatInt(modTime.Unix(), 10)
 	return metadata
 }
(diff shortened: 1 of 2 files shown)


ncw (commented):
I think that code looks fine - well done :-)

Did you run the integration tests against GCS to make sure they still pass?

database64128:
> Did you run the integration tests against GCS to make sure they still pass?

I think the tests are already broken before I made these changes. I ran the integration tests with:

```bash
$ ./test_all.exe -backends googlecloudstorage -remotes gcs:
```

The operations and sync tests are failing with weird errors like `Modification time difference too big |89ns| > 1ns (want 2011-12-25 12:59:59.123456789 +0000 UTC vs got 2011-12-25 20:59:59.1234567 +0800 CST) (precision 1ns)`. For some reason it lost 2 digits of precision.

database64128:
I checked the test files' metadata in my GCS bucket. Test files with UTC timestamps seem to have the correct timestamp precision, while those with timestamps of local time seem to have 2 digits of precision lost. Changing `modTime.Format()` to `modTime.UTC().Format()` won't help. Let me give it a try on Linux to see if it has something to do with the host OS.

database64128:
```
2021/05/21 13:21:45 PASS: All tests finished OK in 2m17.569398191s
```

All tests passed on Linux.

database64128:
> In an NTFS file system, timestamps are stored as 8-byte file time values which represent the number of 100-nanosecond intervals that have elapsed since 12:00 A.M. January 1, 1601 (MSDN Article on File Times). Consequently, NTFS timestamps have 100 nanosecond (0.1 microsecond) precision.

This explains the 2-digit precision loss observed on Windows.

ncw:
Ah yes, sorry that caused you a problem. The tests should probably be fixed, however they are usually run from a linux box so we tend not to notice problems like that.

Rclone sync itself adjusts the modify window for Windows - no reason why the tests shouldn't either.

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -51,9 +52,10 @@ const (
 	rcloneClientID              = "202264815644.apps.googleusercontent.com"
 	rcloneEncryptedClientSecret = "Uj7C9jGfb9gmeaV70Lh058cNkWvepr-Es9sBm0zdgil7JaOWF1VySw"
 	timeFormatIn                = time.RFC3339
-	timeFormatOut               = "2006-01-02T15:04:05.000000000Z07:00"
-	metaMtime                   = "mtime" // key to store mtime under in metadata
-	listChunks                  = 1000    // chunk size to read directory listings
+	timeFormatOut               = time.RFC3339Nano
```
ncw: `RFC3339Nano` removes trailing 0s which aren't used whereas the current format doesn't. Why do you wish to make this change?

ncw (commented):
Looking good apart from a couple of worries about time formats - see inline :-)

Thank you!

backend/googlecloudstorage/googlecloudstorage.go
```diff
@@ -51,8 +51,7 @@ import (
 const (
 	rcloneClientID              = "202264815644.apps.googleusercontent.com"
 	rcloneEncryptedClientSecret = "Uj7C9jGfb9gmeaV70Lh058cNkWvepr-Es9sBm0zdgil7JaOWF1VySw"
-	timeFormatIn                = time.RFC3339
-	timeFormatOut               = time.RFC3339Nano
```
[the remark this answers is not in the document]
database64128: > This will lose the precision from the timestamp won't it?

I'm not sure I understand what you mean. `timeFormatIn` was only used in `time.Parse(timeFormatIn, t)`, which gives the same result if you replace it with `time.RFC3339Nano`. `timeFormatOut` was assigned the same value as `time.RFC3339Nano`. And it was only used in `time.Format(timeFormatOut)`.

backend/googlecloudstorage/googlecloudstorage.go
database64128: I didn't realize the current format doesn't remove trailing zeros. But removing them shouldn't cause any problems. If you're concerned, I can revert this change.

backend/googlecloudstorage/googlecloudstorage.go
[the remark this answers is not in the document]
ncw: Sorry I got confused and I thought you were replacing the format with `time.RFC3339` not `time.RFC3339Nano`.

I think this change is OK!

backend/googlecloudstorage/googlecloudstorage.go
ncw: I don't think it will cause problems either - lets leave your change alone.

ncw (approved):
I think this looks fine now I've got over my confusion!

I'll merge it now - thank you very much :-)

Issue #2749, opened 2018-11-19
Print status of file-hashing


MrCheeze96:
This issue can be read about in the following thread of the rclone forum: https://forum.rclone.org/t/hash-status-progress-bar/7582/9

ivandeex:
Implementation of this request is somewhat related to #2740 as it activates pacer inside the hash/download/localfs stack.

ivandeex:
Let me copy here a comment by @ncw from https://github.com/rclone/rclone/pull/5352#discussion_r641623469

> In an ideal world we'd have local checkers counting transfers like downloads do - there have been loads of requests for that.
> 
> This could use a similar mechanism to this
> 
> https://github.com/rclone/rclone/blob/df60e6323c07a20de08c73839c67aa5f44e141e4/fs/accounting/accounting.go#L275-L295
> 
> If checksums for the local backend always used this mechanism then it would solve the UI problem were users are wondering why their B2/S3 uploads don't start because the local backend is making the checksum in advance...
> 
> > I still don't like current pacer. In case of many files it walks up next --checkers files, calculates ETA, then adds next checker, updates ETA and so on. This way ETA is pretty small all the time seeing only partial work, then steadily decreases close to zero, then jumps back to another small duration, then again. This is useless in practice.
> 
> To achieve this we'd want to move some of the logic from fs/sync into fs/march or possibly run `rclone check` from the `rclone sync` framework rather than using fs/march.
> 
> The logic missing is queuing the `checks` so we have an idea of how many we need to do which is only in `fs/sync`.
> 
> Anyway I think any of these suggestions should probably go in a new PR

PR #5352, opened 2021-05-22
cmd: add hashSUM file support

### LINKED ISSUE #157: md5sum file support
Would it be an improvement if local filesystem would read ".md5" files, so it could get checksums without reading the file?

If you aren't against it, I wouldn't mind implementing it. Of course it should check the file modified times.

I would try to make it generic, so it could also be used by other filesystems without native checksum support.

### LINKED ISSUE #1005: rclone md5sum: implement an Unix md5sum-like "-c" option
As commented here: https://github.com/ncw/rclone/issues/999#issuecomment-271163504.

ivandeex:

Currently `rclone check` supports matching two file trees by sizes and hashes.
However, rclone does not support so called hash-SUM files.
These are produced by the de-facto standard family of GNU core utilities like [sha1sum](https://en.wikipedia.org/wiki/Sha1sum).

### Use Cases

SUM files provide end-to-end checksums, an invaluable tool where storage sums don't exist.
They serve the following use cases:
- User-side "cache" of remote hashsums.
- As a measure to detect [bit-rot](https://en.m.wikipedia.org/wiki/Data_corruption#Silent) timely.
- Tracking integrity across multiple data migrations.
- Migration of ancient data corpus to incompatible provider.
- And more...

### File Naming

There exist two widely used naming schemes for SUM files:
1. `MD5SUM`, `SHA1SUM`, etc - uppercase name without extension carrying the hash digest type right in the name,
   usually located right beside the files to check.
2. `something.md5`, `somethingelse.sha1`, etc - having hash type in extension.

### File Format

The SUM file format is [very simple](https://en.wikipedia.org/wiki/Sha1sum#Hash_file_trees), has been stable for years and supported by a wide range of software. However, there is no official standard. GNU man pages ([md5sum](https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html#md5sum-invocation) et al) is the _de-facto_ standard.

SUM files are line oriented. They **don't record** the hash digest type - it should be known in advance.
Each line contains:
1. Hash digest
2. Space character
3. Another space or `*` (asterisk, see below)
4. File name, up to the line end.

The asterisk modifier is a legacy artifact of _GNU coreutils_ builds for DOS/Windows where it used to denote that the file is _binary_ (with Unix line endings). It means nothing on Unix. The proposed implementation of SUM file support simply ignores asterisks and treats all files as _binary_ (no DOS line feed translation), following the [SUM file parser](https://github.com/ericlagergren/go-coreutils/blob/master/md5sum/checksum_common/check_sum.go#L85) in the [Go coreutils port](https://github.com/ericlagergren/go-coreutils/).

The "distance" between digest and file name **must** always be 2 characters, be it `" *"` for binary files or `"  "` (two spaces) for **ex** text files. The 3rd space, if any, will belong to the file name.

Note: I could not find any real world use of the `" |"` modifier mentioned by @klauspost in https://github.com/rclone/rclone/issues/157#issuecomment-144439399.

### Unimplemented Features

Besides _ignoring asterisks_, this implementation will not:
- perform _backslash unescaping_ (ancient hack to tackle extended unicode and whitespace in file names)
- support DOS-related flags _--binary_ and _--text_ (we treat files as _binary_, use them as is and never translate EOLs)
- handle logic of rarely used flags _--strict_ and _--zero_ from the [GNU manual](https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html#md5sum-invocation)

These features could be implemented by future PRs, but I guess nobody will ever need them.

Note that integration with rclone `fs.Hash` API (e.g. using SUM files as a transparent hashsum source or cache) is also out of scope for this PR.

### Making Sum Files

rclone can produce valid SUM files with command
`rclone hashsum HASH REMOTE:PATH --output-file FILE [--download]`
(one minor feature that I miss sometimes is sorting output by file name - one day I'm gonna submit a PR)

Standard coreutils normally print SUM data to _STDOUT_ so the file can be **produced** by redirection `sha1sum /files > SHA1SUM`. To **check** file tree against a SUM file, use the `-c` flag like `sha1sum -c SHA1SUM /files`. The hash digest used is defined by the name of the executable but **not recorded anywhere** in the sum file. This pull request handles only the **check** case above.

### CLI: Constraints

This patch tries to enroll the `-c` GNU convention into the rclone CLI flag family respecting the following constraints:
1. The `-c` short flag is already taken, so we use upper-case `-C` to feel as close as possible.
2. The `--checksum` long flag is also taken, so we use `--checkfile` (no dash since it's a noun).
    Probably `--sum-file`  would be closer to the file format name, but the first letter `c` better correlates with the short flag. See a note below.

### CLI: Variants

The patch introduces a single internal function `operations.CheckSum` and the following command-line interfaces for it:

1. The `hashsum` command gains new flag `--checkfile SUMFILE`, ex.:
   `rclone hashsum CRC32 -C sumfile.crc32 remote:path/to/files`
2. Commands `sha1sum` and `md5sum` also gain this flag, ex.:
   `rclone sha1sum --checkfile SHA1SUM remote:path/tofiles`
3. New command `checksum HASHNAME SUMFILE REMOTE:PATH` is added, ex.:
   `rclone checksum md5 remote:path/to/MD5SUM remote:path/to --exclude MD5SUM`
4. The `check` command gains a new form with **3** arguments, ex.:
   `rclone check HASHNAME SUMFILE REMOTE:PATH`
5. The `check` command in the current 2-argument form
   gains the new flag `--checkfile HASHNAME`.

As the last case feels slightly out of sync, let me bring an example:
```
rclone check REMOTE1:PATH1 REMOTE2:PATH2     # match two file trees by hash/size
rclone check -C sha1 SUMFILE REMOTE:PATH     # check file tree against a SUM file by hash
```

All the variants are placed in separate commits within this PR so **reviewer can decide** which are good for final squash/merge and which are doomed.

### Notes

1. In all cases above the SUM file can point to a remote object using the `REMOTE:PATH` notation as well as be a local file
   (SUM files are frequently distributed within the same tree as the data).
2. The new `rclone checksum` command supports all extended reporting features from `rclone check`,
   like `--combined`, `--missing-...` and so on.
3. The `md5sum`, `sha1sum` and `hashsum` commands _do not_ support extended check reporting features.
   Instead, they work as if `--combined -` was supplied (i.e. _all sigils_ are reported to **stdout**, unless `--progress` is enabled).
   This behavior is compatible with `-c` in GNU utilities.
4. All proposed commands support filtering, which can be useful if a sum file is located beside the files to be checked.
5. All variants support the `--download` flag and respect `--checkers`.
6. The `--checkfile` option might be named as `--check-from` or `--check-by` (with short name `-C`),
   or just `--check` if not taken. I'm leaving the decision up to reviewer.

### Was the change discussed in an issue or in the forum before?

Approved by https://github.com/rclone/rclone/issues/157#issuecomment-532762729
Fixes #1005

### Checklist

- [x] I'm done, this Pull Request is ready for review :-Q


albertony:
Good stuff!
How does the various command variants handle non-existent and added files? If checksum file contains information for a file that no longer exists in the remote, I'm guessing this will be reported by all variants? But what about the opposite, if remote contains a file that is not represented in checksum file, is there a difference between how, say hashsum and check commands, handle this?

ivandeex:
> How does the various command variants handle non-existent and added files? If checksum file contains information for a file that no longer exists in the remote, I'm guessing this will be reported by all variants? But what about the opposite, if remote contains a file that is not represented in checksum file, is there a difference between how, say hashsum and check commands, handle this?

It follows `check` command convention, where a file _absent in source_ tree is marked by `-`, a file _absent on destination_ is matked by `+`. It just declares the sum file as a source and the file tree under check as destination. That's all.

ivandeex:
Beta build is here https://github.com/ivandeex/rclone/releases/tag/v1.55.1-checksum01

ivandeex:
Find beta build here  https://github.com/ivandeex/rclone/releases/tag/v1.55.1-checksum01

ivandeex:
Find beta build here  https://github.com/ivandeex/rclone/releases/tag/v1.55.1-checksum03

ivandeex:
Find beta build here  https://github.com/ivandeex/rclone/releases/tag/v1.55.1-checksum04

ivandeex:
Find beta build here  https://github.com/ivandeex/rclone/releases/tag/v1.55.1-checksum05

ivandeex:
@ncw Please review!

fs/operations/check.go
```diff
@@ -358,3 +378,215 @@ func CheckDownload(ctx context.Context, opt *CheckOpt) error {
...
+			err = err2 // pass to matchSum
+			return
+		}
+		objHash = hashVals[hashType]
+	}()
+}
+
+// matchSum sums up the results of hashsum matching for an object
+func (c *checkMarch) matchSum(ctx context.Context, sumHash, objHash string, obj fs.Object, err error, hashType hash.Type) {
+	switch {
```
ivandeex, line 515: @ncw 
The checking transfer accounting used to start and end here, in `matchSum`.
In the last commit (I marked it as _experimental_) I made it start earlier.

fs/operations/check.go
```diff
@@ -358,3 +378,215 @@ func CheckDownload(ctx context.Context, opt *CheckOpt) error {
...
+	sumHash, sumFound := hashes[remote]
+	hashes[remote] = "" // mark sum as consumed
+	c.ioMu.Unlock()
+
+	if !sumFound && c.opt.OneWay {
+		return
+	}
+
+	var err error
+	tr := accounting.Stats(ctx).NewCheckingTransfer(obj)
```
ivandeex, line 463: @ncw 

After the last _experimental_ commit the accounting for checking transfer starts **here**,
before we call `object.Hash` or spawn a go-routine to hash object contents.
I am not sure what behavior is more correct, both variants feel fine.
**Please advise!**

Sample real-task screenshot with early checker accounting:
```
rclone --exclude sha1.txt sha1sum -C hdd:dumps/sha1.txt hdd:dumps -P --checkers 2 --transfers 2 --download
2021-05-24 18:58:17 NOTICE: = e5-02.bin
2021-05-24 18:58:18 NOTICE: = e5-01.bin
2021-05-24 19:00:31 NOTICE: = e5-03.bin
2021-05-24 19:00:31 NOTICE: = e5-04.bin
Transferred:       19.976Gi / 20.027 GiByte, 100%, 40.205 MiByte/s, ETA 1s
Checks:                 8 / 9, 89%
Transferred:            6 / 8, 75%
Elapsed time:      8m29.5s
Checking:
 *                      e5-07.bin: checking
Transferring:
 *                      e5-05.bin: 99% /2.502Gi, 22.286Mi/s, 0s
 *                      e5-06.bin: 98% /2.504Gi, 22.098Mi/s, 1s
```

fs/operations/check.go
```diff
@@ -358,3 +378,215 @@ func CheckDownload(ctx context.Context, opt *CheckOpt) error {
...
+	if !download {
+		var objHash string
+		objHash, err = obj.Hash(ctx, hashType)
+		c.matchSum(ctx, sumHash, objHash, obj, err, hashType)
+		return
+	}
+
+	c.wg.Add(1)
+	c.tokens <- struct{}{} // put a token to limit concurrency
+	go func() {
```
ivandeex, line 485: @ncw 
Do you think it would be more natural to limit the `!download` branch by `checkers` and go-routines by `transfers`?

fs/operations/check.go
ivandeex: I still don't like current pacer. In case of many files it _walks_ next --checkers files, calculates ETA, then adds next checker, updates ETA and so on. This way ETA is all way small, decreases to almost zero, then jumps to another small duration, then again. This is useless. I would like it to behave like `rclone sync` - count number (or sum sizes in case of --download) of as many files as possible during _walk_. I don't know how to modify existing code to get there. Please help, @ncw.

ncw:
This is a very nice bit of work :-)

I'll comment on your initial doc - I've deleted everything I agree with which is most of it!

I'm very fond of storing MD5SUM files at the root of my archives so this is great for that!

> SUM files are line oriented. They **don't record** the hash digest type - it should be known in advance.

I'll just note there is another more sophisticated format invented by the BSDs which you enable with the `--tag` command line switch. This does have the hash type in. However it is quite uncommon and I much prefer the simpler format you've described.

> rclone can produce valid SUM files with command
> `rclone hashsum HASH REMOTE:PATH --output-file FILE [--download]`
> (one minor feature that I miss sometimes is sorting output by file name - one day I'm gonna submit a PR)

A PR to list everything in some defined order would be great! It will slow down listings I think but it could apply to anything which calls `fs.ListFn`

> This patch tries to enroll the `-c` GNU convention into the rclone CLI flag family respecting the following constraints:
> 
> 1. The `-c` short flag is already taken, so we use upper-case `-C` to feel as close as possible.

Seems OK

> 2. The `--checksum` long flag is also taken, so we use `--checkfile` (no dash since it's a noun).
>    Probably `--sum-file`  would be closer to the file format name, but the first letter `c` better correlates with the short flag. See a note below.

`md5sum` calls this `--check` if we want to engage prior art

    -c, --check          read MD5 sums from the FILEs and check them

Though `--checkfile` is probably more general purpose. It could even be `--checksumfile`

> The patch introduces a single internal function `operations.CheckSum` and the following command-line interfaces for it:
> 
> 1. The `hashsum` command gains new flag `--checkfile SUMFILE`, ex.:
>    `rclone hashsum CRC32 -C sumfile.crc32 remote:path/to/files`

Great

> 2. Commands `sha1sum` and `md5sum` also gain this flag, ex.:
>    `rclone sha1sum --checkfile SHA1SUM remote:path/tofiles`

Super

> 3. New command `checksum HASHNAME SUMFILE REMOTE:PATH` is added, ex.:
>    `rclone checksum md5 remote:path/to/MD5SUM remote:path/to --exclude MD5SUM`

Is there anything this command does that can't be done with 1-3,5? Reading below, it is the `rclone check` outputs that make it unique.

I wonder whether this command should be able to generate the checksum files too? Effectively obsoleting the badly named `rclone hashsum`?

> 4. The `check` command gains a new form with **3** arguments, ex.:
>    `rclone check HASHNAME SUMFILE REMOTE:PATH`

I think I prefer `--checkfile` over this as in 5, then it leaves the
door open for `rclone check` to take multiple arguments.

> 5. The `check` command in the current 2-argument form
>    gains the new flag `--checkfile HASHNAME`.

Great

> All the variants are placed in separate commits within this PR so **reviewer can decide** which are good for final squash/merge and which are doomed.

Decisions, decisions!
 
> 1. In all cases above the SUM file can point to a remote object using the `REMOTE:PATH` notation as well as be a local file
>    (SUM files are frequently distributed within the same tree as the data).

Nice.

> 2. The new `rclone checksum` command supports all extended reporting features from `rclone check`,
>    like `--combined`, `--missing-...` and so on.

I see...

> 3. The `md5sum`, `sha1sum` and `hashsum` commands _do not_ support extended check reporting features.
>    Instead, they work as if `--combined -` was supplied (i.e. _all sigils_ are reported to **stdout**, unless `--progress` is enabled).
>    This behavior is compatible with `-c` in GNU utilities.

OK

Except for how do they report with --progress - to the log?

> 6. The `--checkfile` option might be named as `--check-from` or `--check-by` (with short name `-C`),
>    or just `--check` if not taken. I'm leaving the decision up to reviewer.

I think `--checkfile` is fine. Or possibly `--checksumfile` but maybe that is too long?

fs/operations/check.go
ncw: In an ideal world we'd have local checkers counting transfers like downloads do - there have been loads of requests for that.

This could use a similar mechanism to this

https://github.com/rclone/rclone/blob/df60e6323c07a20de08c73839c67aa5f44e141e4/fs/accounting/accounting.go#L275-L295

If checksums for the local backend always used this mechanism then it would solve the UI problem were users are wondering why their B2/S3 uploads don't start because the local backend is making the checksum in advance...

> I still don't like current pacer. In case of many files it walks up next --checkers files, calculates ETA, then adds next checker, updates ETA and so on. This way ETA is pretty small all the time seeing only partial work, then steadily decreases close to zero, then jumps back to another small duration, then again. This is useless in practice.

To achieve this we'd want to move some of the logic from fs/sync into fs/march or possibly run `rclone check` from the `rclone sync` framework rather than using fs/march.

The logic missing is queuing the `checks` so we have an idea of how many we need to do which is only in `fs/sync`.

Anyway I think any of these suggestions should probably go in a new PR

ncw (commented):
Fantastic work :-)

I put some comments inline but I think this is nearly ready to merge :-)

fs/operations/check.go
```diff
@@ -358,3 +375,205 @@ func CheckDownload(ctx context.Context, opt *CheckOpt) error {
...
+
+	return c.reportResults(ctx, listErr)
+}
+
+// checkSum checks single object against golden hashes
+func (c *checkMarch) checkSum(ctx context.Context, obj fs.Object, download bool, hashes HashSums, hashType hash.Type) {
+	remote := obj.Remote()
+	c.ioMu.Lock()
+	sumHash, sumFound := hashes[remote]
+	hashes[remote] = "" // mark sum as consumed
```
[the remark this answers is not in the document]
ivandeex: This is deliberate!
Deleting an entry will make `hash, found := hashes[remote]` return `found==false` as if remote was not present in the checkfile. Blanking it results in `found==true && hash==""` which marks remote as present in checkfile and handled by `walk`.
Given that checkfiles cannot have empty remotes, this allows to detect the "missing on filesystem" condition!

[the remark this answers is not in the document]
ivandeex: Note: in future I plan to introduce _trie_ in-memory structure as a way to keep file trees more efficiently.
See https://github.com/rclone/rclone/pull/5164#issuecomment-843481228

cmd/check/check.go
```diff
@@ -151,13 +151,17 @@ If you supply the |--checkfile HASH| flag with a valid hash name,
...
 	RunE: func(command *cobra.Command, args []string) error {
-		cmd.CheckArgs(2, 2, command, args)
+		cmd.CheckArgs(2, 3, command, args)
 		var (
 			fsrc, fdst fs.Fs
 			hashType   hash.Type
 			fsum       fs.Fs
 			sumFile    string
 		)
+		if len(args) == 3 {
```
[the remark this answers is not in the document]
ivandeex: Ok. Consider that commit dropped.

fs/operations/check.go
```diff
@@ -56,6 +56,7 @@ type checkMarch struct {
 	dstFilesMissing int32
 	matches         int32
 	opt             CheckOpt
+	progress        bool
```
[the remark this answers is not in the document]
ivandeex: > Maybe we should put this code in `operations.syncPrintf` (and make it public and rename the current SyncPrintf to something more sensible)?

Then `SyncPrintf` will have to check its file descriptor. If it's `os.Stdout`, it will use the progress hook (if any). If not, it will output to the file like before. Right?

fs/operations/check.go
ivandeex: > I think any of these suggestions should probably go in a new PR.

Totally agree. This is too tricky!

ivandeex: Please let me return to this problem when get to self-assigned #2749

cmd/check/check.go
```diff
@@ -25,12 +26,14 @@ var (
 	match        = ""
 	differ       = ""
 	errFile      = ""
+	checkSum     = ""
```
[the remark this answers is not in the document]
ivandeex: There is a trick. The option is called `--checkfile` but in the realm of `check` command it carries the `hash name` (and changes the role of the 1st arg).

Please confirm.

fs/operations/check.go
ivandeex: I saved your suggestions at https://github.com/rclone/rclone/issues/2749#issuecomment-850789836

fs/operations/check.go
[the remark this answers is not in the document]
ivandeex: Implemented by https://github.com/rclone/rclone/pull/5352/commits/45196536ba46a246a84bf56ba85094f5b068a7de
PTAL :pray:

ivandeex:
@ncw
Summary of introduced command forms:
```
rclone check sums.sha1 remote:path --checkfile sha1        # note: option is named "--checkfile" (-C) but carries hash name
rclone checksum sha1 sums.sha1 remote:path                   # checksum by default checks, hashsum by default produces sums
rclone hashsum sha1 remote:path --checkfile sums.sha1
rclone sha1sum --checkfile sums.sha1 remote:path
rclone md5sum remote:path --checkfile sums.md5
```

ivandeex:
> `md5sum` calls this `--check` if we want to engage prior art
> ```
> -c, --check          read MD5 sums from the FILEs and check them
> ```

This is [not present](https://github.com/rclone/rclone/blob/v1.55.1/cmd/md5sum/md5sum.go) on the master branch.
Probably `make commanddocs` has not been run for long. This PR reruns it for affected commands so this legacy will disappear from docs.

> I think `--checkfile` is fine. Or possibly `--checksumfile` but maybe that is too long?

Feels too long IMO.

ivandeex:
@ncw
Summary of introduced command forms after 1st review:
```
rclone check sums.sha1 remote:path --checkfile sha1    # note: flag is named "--checkfile" but carries hash name!
rclone checksum sha1 sums.sha1 remote:path             # checksum by default checks, hashsum by default produces sums
rclone hashsum sha1 remote:path --checkfile sums.sha1
rclone sha1sum remote:path --checkfile sums.sha1
rclone md5sum remote:path --checkfile sums.md5
```

PTAL :pray:

ivandeex:
@ncw
I addressed your review comments.

Summary of introduced command forms after 1st review:
```
rclone check sums.sha1 remote:path --checkfile sha1    # note: flag is named "--checkfile" but carries hash name!
rclone checksum sha1 sums.sha1 remote:path             # checksum by default checks, hashsum by default produces sums
rclone hashsum sha1 remote:path --checkfile sums.sha1
rclone sha1sum remote:path --checkfile sums.sha1
rclone md5sum remote:path --checkfile sums.md5
```

PTAL :pray:

fs/operations/check.go
ivandeex: Let me return to check accounting later in #2749

ivandeex: Let me return to the check accounting problem later in #2749

ivandeex: Let me return to the check accounting problem later in #2749 :pray:

cmd/check/check.go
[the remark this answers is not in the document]
ivandeex: There is a trick. The option is called `--checkfile` but in the realm of `check` command it carries the `hash name`
(but also _changes the role_ of the 1st arg - it becomes "checkfile").

Do you still want to rename? Please confirm.

ivandeex:
@ncw
I addressed your review comments.

Summary of introduced command forms after 1st review:
```
rclone check sums.sha1 remote:path --checkfile sha1    # note: flag is named "--checkfile" but carries hash name!
rclone checksum sha1 sums.sha1 remote:path             # checksum by default checks, hashsum by default produces sums
rclone hashsum sha1 remote:path --checkfile sums.sha1
rclone sha1sum remote:path --checkfile sums.sha1
rclone md5sum remote:path --checkfile sums.md5
```

PTAL :pray:

ivandeex:
> `md5sum` calls this `--check` if we want to engage prior art
> ```
> -c, --check          read MD5 sums from the FILEs and check them
> ```

This is [not present](https://github.com/rclone/rclone/blob/v1.55.1/cmd/md5sum/md5sum.go) on the master branch. Probably `make commanddocs` has not been run for long. This PR reruns it for affected commands so this legacy will disappear from docs.

> I think `--checkfile` is fine. Or possibly `--checksumfile` but maybe that is too long?

Feels too long IMO.

ivandeex:
@ncw
I addressed your review comments.

Summary of introduced command forms after 1st review:
```
rclone check sums.sha1 remote:path --checkfile sha1    # note: flag is named "--checkfile" but carries hash name!
rclone checksum sha1 sums.sha1 remote:path             # checksum by default checks, hashsum by default produces sums
rclone hashsum sha1 remote:path --checkfile sums.sha1
rclone sha1sum remote:path --checkfile sums.sha1
rclone md5sum remote:path --checkfile sums.md5
```

PTAL :pray:

cmd/check/check.go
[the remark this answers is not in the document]
ivandeex: There is a trick.
The option is called `--checkfile` but in the realm of `check` command it carries the `hash name`
(but also _changes the role_ of the 1st arg - it becomes "checkfile").

Rename the var to `checkFileHashType`? (commit https://github.com/rclone/rclone/pull/5352/commits/7d86a77c94dc5f8b19b4ff1768986e38108c4e22)
Please confirm.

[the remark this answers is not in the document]
ivandeex: There is a trick.
The option is called `--checkfile` but in the realm of `check` command it actually carries the `hash type`,
but also _changes the role_ of the 1st arg - it becomes "checkfile".

Rename the var to `checkFileHashType`? (commit https://github.com/rclone/rclone/pull/5352/commits/7d86a77c94dc5f8b19b4ff1768986e38108c4e22)
Please confirm.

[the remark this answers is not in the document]
ivandeex: There is a trick.
The flag is called `--checkfile`, but in context of the `check` command it actually carries the `hash type`,
though also _changes the role_ of the 1st arg - it becomes "checkfile".

Rename the var to `checkFileHashType`? (commit https://github.com/rclone/rclone/pull/5352/commits/7d86a77c94dc5f8b19b4ff1768986e38108c4e22)
Please confirm.

ivandeex:
@ncw PTAL

ivandeex:
@ncw IMHO this can be merged for 1.56

ncw (approved):
> IMHO this can be merged for 1.56

Great!

I think there are some commits that need removing from the commit log - these are noted in our discussion, but other than those this looks great!

So please remove those commits and merge :-)

Thanks

Nick

ivandeex:
sqiashing to single commit. reverted commits will annihilate.

tb582:
am I reading this right that its not possible to supply rclone your own already existing file of SUM's ?

ncw:
@tb582 you can now, either with `rclone md5sum -C` or `rclone checksum` - check out [the latest beta](https://beta.rclone.org/). You can't yet supply checksums for use in a sync, but we are thinking about that.

PR #5362, opened 2021-05-27
librclone: add FreeString function

wengh:
to free the string pointer returned by RcloneRPC function; it can be useful for languages that lack built-in ways to free malloc'ed memory (such as .NET).

Some languages (e.g. C#) cannot appropriately free the memory allocated by `RcloneRPC` function using built-in API.
To allow these languages to consume librclone we can add a function that exposes `C.free` which [is used to free](https://stackoverflow.com/a/47192786/5491216) the pointer created by `C.CString` and returned to the caller from `RcloneRPC`.

No

#### Checklist

### DIFF
--- a/librclone/librclone.go
+++ b/librclone/librclone.go
@@ -20,12 +20,15 @@
 package main
 
 /*
+#include <stdlib.h>
+
 struct RcloneRPCResult {
 	char*	Output;
 	int	Status;
 };
 */
 import "C"
+import "unsafe"
 
 import (
 	"github.com/rclone/rclone/librclone/librclone"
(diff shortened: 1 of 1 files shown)


librclone/librclone.go
```diff
@@ -79,5 +82,11 @@ func RcloneRPC(method *C.char, input *C.char) (result C.struct_RcloneRPCResult)
 	return result
 }
 
+// Use this to free the string returned by RcloneRPC
+//export FreeString
+func FreeString(str *C.char) {
```
ncw, line 87: Can we call this `RcloneFreeString` - I've been namespacing the C symbols starting with `Rclone`.

librclone/librclone.go
```diff
@@ -79,5 +82,11 @@ func RcloneRPC(method *C.char, input *C.char) (result C.struct_RcloneRPCResult)
 	return result
 }
 
+// Use this to free the string returned by RcloneRPC
```
ncw, line 85: This comment will need to start `// RcloneFreeString may be used to free the string returned by RcloneRPC`

Perhaps command here that you can equally well use `free` from the shared C library if you have access to it.

ncw (commented):
Looks useful - thank you.

I put some minor comments inline!

PR #5363, opened 2021-05-27
vfs: ensure names used in cache path are legal on current os

### LINKED ISSUE #5360: Rclone fails to make cache dir if file/folder name has illegal characters on Windows

With a mounted remote if said remote has an illegal character in any of it's filenames rclone will fail in creating the cache dir. In my case the parent directories (parent for the mount) have the character `:` in their names which in Windows is illegal. This should be easy to reproduce, just name a folder with one of those illegal characters, mount it on Windows and try to open a file.

1.55.1

Windows 10, x64

Google Drive

`rclone mount Crypt_Remote:REDACTED U: --read-only --volname Folder --vfs-cache-mode full --network-mode`

```
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: vfs cache: failed to remove cache file as failed to stat cache file: CreateFile \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.: remove \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: vfs cache: failed to remove metadata from cache as failed to stat cache file: CreateFile \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.: remove \\?\C:\Users\1024mb\AppData\Local\rclone\vfsMeta\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: vfs cache: failed to remove cache file as failed to load metadata: vfs cache item: failed to read metadata: open \\?\C:\Users\1024mb\AppData\Local\rclone\vfsMeta\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.: remove \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: vfs cache: failed to remove metadata from cache as failed to load metadata: vfs cache item: failed to read metadata: open \\?\C:\Users\1024mb\AppData\Local\rclone\vfsMeta\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.: remove \\?\C:\Users\1024mb\AppData\Local\rclone\vfsMeta\Crypt_Remote\REDACTED\HDD (Y:)\quarentine\file.jpg: El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos.
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: vfs cache: failed to open item: vfs cache item: open mkdir failed: make cache REDACTED failed: mkdir \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:): El nombre del directorio no es válido.
2021/05/25 23:47:35 ERROR : HDD (Y:)/quarentine/file.jpg: Non-out-of-space error encountered during open
2021/05/25 23:47:35 ERROR : IO error: open RW handle failed to open cache file: vfs cache item: open mkdir failed: make cache REDACTED failed: mkdir \\?\C:\Users\1024mb\AppData\Local\rclone\vfs\Crypt_Remote\REDACTED\HDD (Y:): El nombre del directorio no es válido.
```

#### How to use GitHub

albertony:

As described in the issue: You can have a dir with `:` in the name in a remote. When mounting this with cache on Windows, the current cache code did not handle such characters that are invalid in paths, and it tried (without luck) to create a dir within the cache location containing the `:`.

With this change: When cache builds the path to be used within the local file cache for a given remote path, it now relies on `FromStandardPath`, with an encoder that properly handle characters not valid in paths on current os.

This is similar to what the local backend does: So refactored the default encoder from the local backend into lib/encoder as `OS` encoder, to re-use it.

An early version was confirmed to fix the linked issue, but I'm still a bit uncertain if this is the correct/best way to handle it...

Fixes #5360

#### Checklist


vfs/vfscache/cache.go
```diff
@@ -75,17 +81,21 @@ type AddVirtualFn func(remote string, size int64, isDir bool) error
...
 	fName := fremote.Name()
-	fRoot := filepath.FromSlash(fremote.Root())
+	fRoot := fremote.Root()
 	if runtime.GOOS == "windows" {
-		if strings.HasPrefix(fRoot, `\\?`) {
-			fRoot = fRoot[3:]
-		}
-		fRoot = strings.Replace(fRoot, ":", "", -1)
-		// Replace leading ':' if remote was created on the fly as ":backend:/path" as it is illegal in Windows
+		// Make remote name valid as dir name below cache dir.
```
ncw, line 86: Why don't we just use the encoder.OS encoding here?

vfs/vfscache/cache.go
```diff
@@ -75,17 +81,21 @@ type AddVirtualFn func(remote string, size int64, isDir bool) error
...
+		// If remote was created on the fly as ":backend:/path" we must remove the leading
+		// ':', illegal character in Windows. Replacing it with something unlikely to cause
+		// collision, and something more convenient than the "fullwidth colon" encoder.OS uses.
 		if fName[0] == ':' {
 			fName = "^" + fName[1:]
 		}
+		// Make remote root valid as relative dir path below remote name dir.
+		// If it is a local path: If prefix "//?/" then remove it, and if drive
+		// letter then remove the ':' following the letter.
+		fRoot = winCachePathRegex.ReplaceAllString(fRoot, "$1$2")
```
ncw, line 96: I don't really understand what this whole `if runtime.GOOS == "windows"` branch is about...

Cant we just do this on all OSes now?

```
fRoot := filepath.FromSlash(fremote.Root())
fRoot = toOSPath(fRoot)
```

ncw (commented):
I think this is a really nice idea.

Can you do the factoring the OS encoding out of the local backend in a separate commit?

A few comments inline!

vfs/vfscache/cache.go
albertony: Could definitely do that. I *was* thinking about it, but decided not to touch that part..
All "on the fly" remotes would then be in cache folders starting with the `fullwidth colon`, which looks a bit "alien", but does probably not matter since it is just a cache folder for "internal use".

vfs/vfscache/cache.go
albertony: For a local path we need to chop off the `\\?\C:\` prefix before appending it to root path of the cache.

E.g. for command `rclone mount "C:\Temp\TEMP" X:`
The local path `C:\Temp\TEMP`, first gets translated to `//?/C:/Temp/TEMP` by rclone. Then this windows-specific block in cache makes that into `C/Temp/TEMP`, and appends that to the cache root, converting UNC, slash etc. into final result, e.g.  `\\?\C:\Rclone\data\vfs\local\C\Temp\TEMP`

So I think this is something that must be done..

albertony: Oh, now i see: You probably meant we could just rely on the os encoder to transform `C:` into "C" followed by the fullwidth colon? That should work. Must test to see what happens, and how the `\\?\` prefix will be "encoded"..

ncw: Yes that was what I meant - just use the path. It might look a little odd so probably worth a check!

ncw:
> @ncw: Is this a bug:
> 
> https://github.com/rclone/rclone/blob/0574ebf44a4c22999e38fa5e5c72d2b8985c301a/vfs/vfscache/cache.go#L103
> 
> Should it have used `metaRoot` instead of `root`?

Yes that does look like a bug... Luckily we only use it in one place

    err = operations.Rmdirs(ctx, c.fcacheMeta, dir, leaveRoot)

So the bug will have been causing empty directories to build up in the meta path tree.

albertony:
Refactored and fixed some issues with original attempt.

Now relying more on the OS encoding to fix illegal characters. Example of cache paths (with unicode fullwidth variants of `?` and `:`):

```
C:\Rclone\data\vfs\local\？\C：\Path
C:\Rclone\data\vfs\local\？\UNC\Server\Share\Path
C:\Rclone\data\vfs\：local{HsCM1}\C：\Dev
```

The `runtime.GOOS == "windows"` section is reduced, but not completely gone...

There are some concerns that makes it hard to find a very clean path through this that I'm happy with:
- If the remote root has prefix `\\?\`, which is normally the case for local paths, then must remove the double path separator or else there will be an error down the road: `failed to load cache: failed to walk cache ...: CreateFile ...: The filename, directory name, or volume label syntax is incorrect.`
- Cannot simply build the path once as standard path, and send it through `FromStandardPath` (via `toOSPath`), because that would encode the `:` after the drive letter to `unicode fullwidth colon` resulting in an incorrect path.
- Uses the "OS Path" variant for most of the cache implementation, but there will be created two local remotes for internal use (purgeEmptyDirs), and these must be created using "standard paths" because they will encode the given paths to the local (=OS) encoding, which means if we send in an OS path with already encoded `？` then it will be escaped to `‛？`, leading to errors..

Some of this could be eased with more use of functionality from golang's `filepath` etc (although it has its ups and downs..). The second point can be achieved with the approach that the local backend does in the `cleanRootPath` function (split drive-letter/volume-name from rest of the path, then encode the path, and prepend the drive-letter back on). I have started on several different attempts at this, but end up aborting with conclusion it does not gain all that much while leading to a significant larger change requiring more testing etc.. There is quite a few corner cases here, when digging into it.. I already have some windows path handling refactoring on my backlog from the mount and UNC path related commits, so I'm considering just adding to that...

So...
Should we go for such a change to cache paths? A bit more alien-looking paths when there are special characters involved, which is the case for all pure local and all on-the fly backends.. But at least it is consistent, also with uses of the local backend, and it should handle any characters the OS would not accept.
And if so, is the current implementation acceptable and readable *enough*?

albertony:
Hold on a bit.. got an(other) idea of refactoring that I want to try out.. 😆

albertony:
Done! 😅

ncw:
Sorry I've dropped the ball on this PR @albertony 

Is this ready to merge?

The encoder commit and bug fix commit look straight forward enough.

The actual fix commit looks more complicated than I expected?

albertony:
> Is this ready to merge?

Yes, should be.

> The actual fix commit looks more complicated than I expected?

True. Initial implementation got a bit messy with managing the two paths (data and meta) each in two forms (OS and internal), so improved that with some refactoring, ending up with some more changes than a minimal fix. I think its worth it, but take a look and if you dont then I can revisit it - but probably not until late august (on vacation).

ncw (approved):
I read through this one more time!

Let's merge it now!

I'll let you press the merge button - I suggest a rebase merge to preserve the 3 commits.

We can fix any fallout before the next release :-)

Thank you

PR #5367, opened 2021-05-29
touch: add support for touching files in directory, with options for recursive, filtering and dry-run/interactive

### LINKED ISSUE #5301: Add `--recursive` to rclone touch.
Currently rclone touch only works on single files, this is fine for that task.  However when needing to bulk touch files in a directory this is only possible by first fetching all the files contained with the directory and any subdirectories and then touch each file one at a time, this is very time consuming and as a result `rclone touch` would benefit of being able to recursively touch all files within a directory.

When using rclone touch with the new `--recursive` flag it should only touch already existing files, and should not create new files by default.

albertony:
Better touch:
- Touch on existing file:
  - Updates timestamp (same as before)
  - Note: Specifying option --recursive (described below) have no effect, meaning it still touches the single file. This is up for debate...
- Touch on existing directory:
  - Default is to touch any files in that directory non-recursively
  - If option --recursive then recursively touching all files in subdirectories as well
  - Filtering options can be used
- Touch on non-existent path:
  - Default is to create it as an empty file (same as before)
  - Does nothing if --no-create (notice, not an error) (same as before)
  - Does nothing if --recursive (notice, not an error), for consistency --recursive never creates *new* files.
- Supports --interactive and --dry-run
- Added debug log

Fixes #5301
Supersedes #5330
Discussed https://forum.rclone.org/t/using-rclone-touch-recursively/23975

#### Checklist


ivandeex:
LGTM

PR #5378, opened 2021-06-02
Making a minor change to support directory-lister for the http remote

Luca-spopo:

The HTTP endpoint was not able to parse html from a product known as "directory lister"
https://www.directorylister.com/

It has slightly non-standard paths compared to apache and nginx, so some tweaking is required in the html parser

Yes, it was requested as a feature earlier by someone unrelated. I also faced a similar use case.
https://forum.rclone.org/t/http-not-listing-from-directory-lister/7803

#### Checklist

- [X ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [X ] I have added tests for all changes in this PR if appropriate.
- [X ] I have added documentation for the changes if appropriate.
- [X ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [X ] I'm done, this Pull Request is ready for review :-)


ivandeex:
similar to #5280 ?

13steinj:
Ah! That was that site I found a while ago.

Somewhat similar to my #5280 however they are completely different products internally.

You'll want to add tests with a live directory-lister server, because given a cursory look it's complex enough that this change may not be enough (and as such, might want to be its' own backend how I was told to do mine).

Luca-spopo:
I have tested this with the directory-lister server I wanted to use it with, it has nested folders and it works fine.
One problematic feature I see is that directory-lister allows one to have custom readme.md style document rendering in addition the the files, which may have links itself which could be mistaken for directories/files. The HTML parser here is not equipped to handle those complex cases, but those cases are rare enough that I can't justify the effort required to add it as a new backend.

ivandeex:
please fix code quality test as follows:
```
go mod tidy -w ./backend/http/http.go
git add ./backend/http/http.go
git commit -m "gratify source quality linter"
git push
```

ivandeex:
Marked as draft until we hear more from submitter.

ncw:
I'm going to close this as it is stale now.

Sorry we never merged it.

PR #5398, opened 2021-06-10
Create direct share link for "koofr" backend

KAction:
#### Checklist

### DIFF
--- a/backend/koofr/koofr.go
+++ b/backend/koofr/koofr.go
@@ -608,5 +608,25 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 	if err != nil {
 		return "", translateErrorsDir(err)
 	}
-	return linkData.ShortURL, nil
+
+	// URL returned by API looks like following:
+	//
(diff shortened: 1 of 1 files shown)


ncw (commented):
I gave this a quick test with `curl` - it works fine there.

However `wget` doesn't and if you open the link in the browser you'll see the reason why

```
Hotlinking not allowed.

Visit this page to see this link
```

So I guess curl is putting a `Referer:` from the same site in.

Other than that it seems to work.

So code-wise it all looks fine.

However this is circumventing Koofr's hotlink protection by the look of it which they could change to be a more secure method very easily.

Given that, do you think I should merge this?

KAction:
I do not promise to maintain this patch forever, but I think it is making life better for now (for curl-based applications), and when (should?) they break it, it should not be hard to revert this patch.

KAction:
Maybe it worth adding warning that URL generated may not work in all applications, e.g
```
logging.warning("backend service actively fighting direct links, so this link may break any time. See #5398")
```

KAction:
@ncw So, what is your take on this?

ncw:
If it is working for you as-is then I think we should merge it.

If you are happy with the code as it is, then I'll do exactly that!

PR #5410, opened 2021-06-14
fs/operations: Don't update timestamps of files in --compare-dest

nickgaya:

The [`--compare-dest`](https://rclone.org/docs/#compare-dest-dir) flag specifies a remote directory to compare files to when syncing.

Currently, if a file in the compare-dest is identical to a source file (same size and checksum), but its timestamp differs, the timestamp of the compare-dest file will be updated.

The documentation does not mention this behavior and it is likely an unintended side effect of how the checking is implemented.  I believe the compare-dest should be treated as an immutable reference point and should not be modified by the operation.

No, as far as I know

#### Checklist

- [x] I have added documentation for the changes if appropriate. (N/A)

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -1408,7 +1408,9 @@ func compareDest(ctx context.Context, dst, src fs.Object, CompareDest fs.Fs) (No
 	default:
 		return false, err
 	}
-	if Equal(ctx, src, CompareDestFile) {
+	opt := defaultEqualOpt(ctx)
+	opt.updateModTime = false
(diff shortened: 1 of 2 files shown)


ncw:
I think you are correct, this is an unintended consequence and therefore a bug!

Thank you for fixing - I'll merge it now :-)

ncw:
This broke quite a few integegration tests - see

https://pub.rclone.org/integration-tests/2021-07-07-050020/index.html

Would you mind taking a look?

ncw:
Don't worry I figured it out - it was to do with those backends not supporting timestamps. I'll add a fix shortly...

Issue #4750, opened 2020-11-08
rclone as docker volume plugin


ivandeex:
Here is a feature idea from https://github.com/sapk/docker-volume-rclone/issues/6.
It would be fairly possible to add a `rclone serve docker` command exposing a docker volume api if allowed by rclone maintainers.

The implentation is in fact a combination of `mount` and http server with a limited number of API commands (create, remove, mount, unmount).
Currently implemented by https://github.com/sapk/docker-volume-rclone mostly in https://github.com/sapk/docker-volume-rclone/blob/master/rclone/driver/driver.go.
As can be seen, the codebase is pretty small.

ivandeex:
blocks on #2968

ncw:
This looks equivalent in complexity to `rclone serve restic` so not too complex.

There seem to be a lot of boilerplate around creating images to run as a docker plugin but maybe we should put the core `rclone serve docker` into rclone itself. What do you think @sapk ?

sapk:
I can work on it. I had already work on the `serve ftp` command so that why I suggest it.

There is two way of running docker plugin:
- Running a daemon on host, listening on a socket for docker requests.
- Package everything in a specific container that docker manage itself.

For the alter one the process is working fine but still a bit hacky to put together.

ivandeex:
Let's start from the first bullet, which implements the `rclone serve docker` listener.
Let's think of adding the plugin container packaging into rclone makefiles later. For now this can live in the sapk/docker-volume-rclone makefiles.

sapk:
I started working on that issue here: https://github.com/sapk-fork/rclone/tree/serve-docker

Issue #5414, opened 2021-06-16
cannot use AWS roles to list/copy data from S3 buckets Using a k8s container


jonjesse:
#### What is the problem you are having with rclone?

Using a container running in AWS EKS environment, I am getting an access denied when trying to list files from a bucket. It is not a permissions issue since I can access bucket using the aws s3 commands

#### What is your rclone version (output from `rclone version`)

```
 rclone version
rclone v1.55.1
- os/type: linux
- os/arch: amd64
- go/version: go1.16.3
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Linux container

#### Which cloud storage system are you using? (e.g. Google Drive)

S3

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
rclone -vvvv ls s3://<private>/
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
 rclone -vvvv ls s3://<private>
2021/06/16 18:08:33 DEBUG : Using config file from "/root/.config/rclone/rclone.conf"
2021/06/16 18:08:33 DEBUG : rclone: Version "v1.55.1" starting with parameters ["rclone" "-vvvv" "ls" "s3://<private>"]
2021/06/16 18:08:33 DEBUG : Creating backend with remote "s3://<private>"
2021/06/16 18:08:38 DEBUG : fs cache: renaming cache item "s3://<private>" to be canonical "s3:<private>"
2021/06/16 18:08:38 DEBUG : 4 go routines active
2021/06/16 18:08:38 Failed to ls: AccessDenied: Access Denied
	status code: 403, request id: 6EE309SFXR03MW4S, host id: iKFzmE/c5K2DEABT9vOtH2vmwAllEwB7pqa/NNWLj5/CmHqpUv3I+CWK7L0orPKEvYAhoNYIVGU=
```

Please note that this is not an S3 permissions issue.
Additional information:
container uses a role attached to it:
```
aws sts get-caller-identity
{
    "UserId": "AROASI6AIV3SMVUZ62UUB:<private>",
    "Account": "<private>",
    "Arn": "arn:aws:sts::<private>:assumed-role/<private>"
}
```
Using this role I can list/read/write files to the S3 bucket using aws s3 command

below are contents of the rclone.conf file:

```
cat /root/.config/rclone/rclone.conf 
[s3]
type = s3
provider = AWS
[...]

PR #5402, opened 2021-06-11
cmd/version: add support for openbsd/amd64

ivandeex:
#### What is the purpose of this change?

#5204 added the `rclone version` command based on the [gopsutil](https://github.com/shirou/gopsutil) project, but we excluded it from the OpenBSD build due to build failures.

The root of the problem appeared to lie in the [CGO dependency left in process_openbsd.go](https://github.com/shirou/gopsutil/blob/v3.21.5/v3/process/process_openbsd.go#L83).

I noticed that https://github.com/shirou/gopsutil/pull/965 claiming to fix the CGO issue is stalled since October 2020. I [improved its patch](https://github.com/shirou/gopsutil/pull/1087/files) so it applies successfully to the current gopsutil master on gopsutil/process and on `gopsutil/v3`.
I then built rclone with patched gopsutil and successfully ran `rclone version` on OpenBSD 6.7 x86_64 ([Virtualbox](https://app.vagrantup.com/fnichol/boxes/openbsd-6.7)):

```
rclone v1.56.0-DEV
- os/version: openbsd 6.7 (64 bit)
- os/kernel: 6.7 (amd64)
- os/type: openbsd
- os/arch: amd64
- go/version: go1.16.4
- go/linking: static
- go/tags: none
```

I submitted the changes as upstream PR https://github.com/shirou/gopsutil/pull/1087 which is currently awaiting approval by @shirou.

#### Was the change discussed in an issue or in the forum before?

Explained here https://github.com/rclone/rclone/pull/5204#discussion_r606765569

Waiting for https://github.com/shirou/gopsutil/pull/1087 merged or for https://github.com/shirou/gopsutil/pull/965 extended to gopsutil/v3 and merged.

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [ ] I have added tests for all changes in this PR if appropriate.
- [ ] I have added documentation for the changes if appropriate.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ ] I'm done, this Pull Request is ready for review :-)

PR #5415, opened 2021-06-19
rclone as docker volume plugin

ivandeex:

This change:
- adds the [Docker Volume](https://docs.docker.com/engine/extend/plugins_volume/)  [**API**](https://docs.docker.com/engine/extend/plugin_api/) support to rclone in the form of `rclone serve docker` command
- slightly refactors `cmd/mountlib` as the implementation heavily uses VFS/mount
- improves Makefile and the CI/CD to build the rclone docker plugin container

Inspired by [sapk/docker-volume-rclone](https://github.com/sapk/docker-volume-rclone) and based on the prior work by @sapk in [sapk-fork/serve-docker](https://github.com/sapk-fork/rclone/tree/serve-docker)

- https://github.com/rclone/rclone/issues/4750
- https://github.com/sapk/docker-volume-rclone/issues/6

#### Checklist

### MountLib

I continued the refactoring started by @sapk on the [sapk-fork/serve-docker](https://github.com/sapk-fork/rclone/commits/serve-docker/cmd/serve/docker) branch. Not only do I factor `RunMount` from `cmd/mountlib/Mount` but also introduce the `MountPoint` structure with dedicated methods and separate `MountInfo` structure for marshalling mount state by the rclone `rcd` API. All refactoring commits as well as original commits from @sapk are **separate** for easier review.

### Docker Plugin SDK

I tried to continue using the [docker/go-plugins-helpers](https://github.com/docker/go-plugins-helpers) SDK but encountered a number of problems:
- We actually use only the Volume API which is less than 20% of SDK
- SDK spams system logs with trace messages and gives no way to disable it
- Systemd socket activation is declared but does not actually work
- SDK unconditionally tries to modify root-only files in `/run/docker/plugins` and `/etc/docker/plugins` and provides *no way* to change this behavior so testing API by non-root was impossible
- SDK depends on external package [docker/go-connections](https://github.com/docker/go-connections/sockets) which attempts to `chown` files adding another root-only dependency
- SDK adds more external dependencies in particular on Azure packages, this growing rclone executable too much

I created the [SDK fork](https://github.com/ivandeex/go-plugins-helpers), developed [a number of patches](https://github.com/ivandeex/go-plugins-helpers/commits/devel) and started submitting pull requests, in particular https://github.com/docker/go-plugins-helpers/pull/130. However, the SDK repository looks *a little bit abandoned*, still has its CI/CD based on the obsolete travis.org so even my first patch got stuck like [a number of other PRs](https://github.com/docker/go-plugins-helpers/pulls). Some other patches depend on unreleased features in [external docker packages](https://github.com/docker/go-connections) so I can't predict whether they can get through in reasonable time.

All in all, I decided to drop dependency on [docker/go-plugins-helpers](https://github.com/docker/go-plugins-helpers) altogether. The use of *go-chi http router* let the resulting code be very small (take a look at [api.go](https://github.com/ivandeex/rclone/blob/ci-docker/cmd/serve/docker/api.go) and [serve.go](https://github.com/ivandeex/rclone/blob/ci-docker/cmd/serve/docker/serve.go)) and get rid of all above problems. Now we can have a complete unit test for the new command.

### httplib and go-chi

I also tried to use the brand new [rclone/lib/http](https://github.com/rclone/rclone/tree/e489a101f631db31ec7c6d97ca4d8034c573031a/lib/http) but found it problematic. I love that it introduces *go-chi*, but it does that in a strange way. I needed to create a server listening on provided socket but the structure returned by [NewServer](https://github.com/rclone/rclone/blob/e489a101f631db31ec7c6d97ca4d8034c573031a/lib/http/http.go#L120) does not have a *Serve method*, at all. I tried to understand how the new `cmd server http` [uses the new library](https://github.com/rclone/rclone/blob/e489a101f631db31ec7c6d97ca4d8034c573031a/cmd/serve/http/http.go#L66) but I failed to wrap my head around who serves what. I guess that the real listening server is hidden as a singleton inside the library, internally attached to the `httplib.Router` and can be configured only via command line. :cry: Maybe I miss something... By the way, what was the reasoning behind renaming `httplib` to `http` and then adding ugly import aliases in a dozen places inside rclone to decipher it from `net/http`? We could just have `rclone/lib/httplib` and simply use it directly. I abandoned further attempts to use this strange code and stick with good old `net/http`  dressed by the seasoned `go-chi`.

### Options

Every mount needs three groups of options: Backend, VFS and Mount options. The base fork has left this problem [for future](https://github.com/sapk-fork/rclone/blob/serve-docker/cmd/serve/docker/driver.go#L41). The generic Volume API will pass options as a `[string]string` map when the volume is created (following mount requests have no way to customize it, so I attach options to the volume not to its mounts). The docker command line provides for `docker volume create ... -o option1=value1 -o ...`, i.e. all options have a *flat scheme* with string names and values. Docker Swarm has a YAML format where options are represented by a yaml map (note that YAML accepts dashes in key names only if the key is quoted, but underscores can be used as is). The least common divisor is the flag string-to-string dictionary.

Rclone provides a number of ways to pass options:
1. as command-line flags with lower-case dash-separated option names with backend name as a prefix,
2. as environment variables with upper-case underscore-separated names,
3. as `name = value` tokens in `rclone.conf` with lower-case underscore-separated names without backend prefix,
4. as connection string tokens similar to item 3 above (no backend prefix here, so only the immediate backend can be configured this way, not the underlying backend in case of alias or crypt),
5. as json dictionaries in the rclone daemon API, with nested dictionaries for mount and VFS options.

The flat option scheme used by the docker plugin mimics the first case above. Every volume option looks like a twin command line flag with obvious exception of the `--` CLI flag prefix. Underscores can be used instead of dashes in option names for cleaner use with YAML. Additionally there are 5 special options:
- `remote` (aliased as `fs`) determines existing remote from config file with a path or on-the-fly remote using the `-o remote=remote_name[,connect_string]:path/subdir` syntax.
- using `-o type=backend` is equivalent to the `:backend:` syntax
- `path` provides explicit on-remote path for *type*, for example `-o type=storj -o path=dir/subdir` is equivalent to `-o remote=:storj:dir/subdir`
- `mount-type` can be "mount", "cmount" or "mount2", and defaults to the first found mount method.
- `persist` is reserved for future to create remotes persisted in rclone.conf similar to rcd, for example `-o mount-type=mount2`

@ncw @sapk I would like to hear from you. Are the `fs type path` options excessive or we can keep them?

Also you can pass a complete `rclone.conf` to the plugin if you put it at `/var/lib/docker/plugins/rclone/config/rclone.conf`. This should allow user to pass storage provider tokens obtained externally similar to headless rclone setup.

### Mount Registry

The next design decision was how to maintain a volume/mount lifecycle. The plugin can run **(a)** in a special `runc` container beside normal docker containers (so called managed mode) or **(b)** as a process independent from docker daemon (under systemd or just invoked from command line). When it runs separately, it can be restarted independently from docker.

The plugin serves mounts to recipients via the **kernel FUSE module** and serves API on a Unix domain or TCP socket. Docker daemon can request volume create or removal requests to the plugin as well as volume mount/unmount requests. Docker passes mounts created by the plugin (on the host or in its container) to docker managed containers via bind mounting mechanism. Every volume can be mounted by several containers independently.

Containers can be killed or crash otherwise, docker daemon can be restarted on crash otherwise, the host can be rebooted, volumes can fail to created due to network problems. Docker swarm can activate a volume on multiple swarm nodes. What should give if the plugin got restarted after crash or reboot? Will docker re-send volume creation requests? Mount requests? How will docker daemon know what requests to resend? It will take care of most situations, e.g. it will restart the plugin and restart (I mean not just *start* but *stop-kill-restart*) the containers. If a container crashed, docker will restart it and pass *new* mount request to plugin with new *mounter ID*. Looks clean, almost... Unfortunately guys from Docker, Inc. who designed the API, have forgotten about the HEALTHCHECK protocol message so their implementation is not 100% clean.

What should the volume plugin do when it's started after crash or host reboot? The correct behavior is to recreate all volumes that existed before the crash. The same does not hold for volume mounts - the restarted plugin cannot restore mounts due to broken connection with storage providers. If a docker container will try to access a zombie mount, the process in it will get `Transport point not connected` from the kernel and that's correct. So I looked into existing volume plugin implementations and found that the prominent ones ([sshfs](https://github.com/vieux/docker-volume-sshfs/blob/v1.4/main.go#L50), [local-persist](https://github.com/MatchbookLab/local-persist/blob/master/driver.go#L231) [netshare](https://github.com/ContainX/docker-volume-netshare/blob/v0.36/netshare/netshare.go#L287), etc) do use a kind of volume registry persisted locally. Moreover, I encountered the ticket [moby#35120](https://github.com/moby/moby/issues/35120) where @cpuguy83 (a docker mainter) says in clear English:

> Yes, this is a flaw in the v1 plugin design itself which propagates itself through to the volume plugin API.
> Again, willing to take a patch to alleviate some of this (and we must understand the trade-offs), but do not generally want to modify the existing volume plugin API itself except for major issues as we are working on bringing CSI on and making modifications to the current API is risky. [...]
> However the plugin should still essentially **treat the caller as an idiot and keep track of it's on it's own** regardless of what capabilities the caller has.

Note that he mentions the industry standard [CSI API](https://github.com/container-storage-interface/spec/blob/master/spec.md), which is currently used by such container orchestrators as [Kubernetes](https://kubernetes-csi.github.io/docs/) or [Nomad](https://www.nomadproject.io/docs/internals/plugins/csi#csi-plugins) and is [going to be](https://github.com/moby/moby/issues/39624) implemented by Docker Swarm in a [future release](https://github.com/moby/moby/pull/41982). I am going to submit `rclone serve csi` when I have resources for it. Hopefully its design is cleaner and CSI plugins can trust their plugin host a little more.

So... now we have the registry persisted in `docker-plugin.state` under rclone cache directory and unit tests for it. In case something gets stuck upon startup, we have the `--forget-state` flag that can be passed to the command and a way to pass it to managed plugin. The base fork [sapk-fork/server-docker](https://github.com/sapk-fork/rclone/tree/serve-docker) proposed [preserving the whole rclone.conf section](https://github.com/sapk-fork/rclone/blob/serve-docker/cmd/serve/docker/driver.go#L62) but I see problems with this approach, in particular how to pass dynamic provider tokens in and out of the managed plugin container. I think that state file is a better solution, waiting for your comments. Some drivers have the ability to [restore state from docker daemon](https://github.com/MatchbookLab/local-persist/blob/master/driver.go#L190) in addition to state file, but I think we a good enough for now and this can be left for future PRs.

### Documentation

I decided to split the short (put aside the VFS/mount options) command help from the long docker plugin documentation. Please take a look at [command help](https://github.com/ivandeex/rclone/blob/ci-docker/docs/content/commands/rclone_serve_docker.md#rclone-serve-docker) and [docker.md](https://github.com/ivandeex/rclone/blob/ci-docker/docs/content/docker.md#docker-volume-plugin). Please take a look and verify against [my plugin build](https://hub.docker.com/r/ivandeex/docker-volume-rclone/tags?page=1&ordering=last_updated), which can be installed using `docker plugin install ivandeex/docker-volume-rclone`. I am using it on a test rig for a month now.

## Conclusion

@ncw @sapk
Looking forward to your reviews :pray:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,3 +256,36 @@ startstable:
 
 winzip:
 	zip -9 rclone-$(TAG).zip rclone.exe
+
+# docker volume plugin
+PLUGIN_IMAGE_USER ?= rclone
+PLUGIN_IMAGE_TAG ?= latest
+PLUGIN_IMAGE_NAME ?= docker-volume-rclone
+PLUGIN_IMAGE ?= $(PLUGIN_IMAGE_USER)/$(PLUGIN_IMAGE_NAME):$(PLUGIN_IMAGE_TAG)
+
+PLUGIN_BASE_IMAGE := rclone/rclone:latest
+PLUGIN_BUILD_DIR := ./build/docker-plugin
+PLUGIN_CONTRIB_DIR := ./cmd/serve/docker/contrib/plugin
+PLUGIN_CONFIG := $(PLUGIN_CONTRIB_DIR)/config.json
+PLUGIN_DOCKERFILE := $(PLUGIN_CONTRIB_DIR)/Dockerfile
+PLUGIN_CONTAINER := docker-volume-rclone-dev-$(shell date +'%Y%m%d-%H%M%S')
+
+docker-plugin: docker-plugin-rootfs docker-plugin-create
+
+docker-plugin-image: rclone
+	docker build --no-cache --pull --build-arg BASE_IMAGE=${PLUGIN_BASE_IMAGE} -t ${PLUGIN_IMAGE} -f ${PLUGIN_DOCKERFILE} .
+
+docker-plugin-rootfs: docker-plugin-image
+	mkdir -p ${PLUGIN_BUILD_DIR}/rootfs
+	docker create --name ${PLUGIN_CONTAINER} ${PLUGIN_IMAGE}
+	docker export ${PLUGIN_CONTAINER} | tar -x -C ${PLUGIN_BUILD_DIR}/rootfs
+	docker rm -vf ${PLUGIN_CONTAINER}
+	cp ${PLUGIN_CONFIG} ${PLUGIN_BUILD_DIR}/config.json
+
+docker-plugin-create:
+	docker plugin rm -f ${PLUGIN_IMAGE} 2>/dev/null || true
+	docker plugin create ${PLUGIN_IMAGE} ${PLUGIN_BUILD_DIR}
+
+docker-plugin-push: docker-plugin-create
+	docker plugin push ${PLUGIN_IMAGE}
+	docker plugin rm ${PLUGIN_IMAGE}
--- a/cmd/serve/docker/docker.go
+++ b/cmd/serve/docker/docker.go
@@ -1,0 +1,72 @@
+// Package docker serves a remote suitable for use with docker volume api
+package docker
+
+import (
+	"context"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/spf13/cobra"
+
+	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/cmd/mountlib"
+	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/vfs"
+	"github.com/rclone/rclone/vfs/vfsflags"
+)
+
+var (
+	pluginName  = "rclone"
+	pluginScope = "local"
+	baseDir     = "/var/lib/docker-volumes/rclone"
+	sockDir     = "/run/docker/plugins"
+	defSpecDir  = "/etc/docker/plugins"
+	stateFile   = "docker-plugin.state"
+	socketAddr  = "" // TCP listening address or empty string for Unix socket
+	socketGid   = syscall.Getgid()
+	canPersist  = false // allows writing to config file
+	forgetState = false
+	noSpec      = false
+)
+
+func init() {
+	cmdFlags := Command.Flags()
+	// Add command specific flags
+	flags.StringVarP(cmdFlags, &baseDir, "base-dir", "", baseDir, "base directory for volumes")
+	flags.StringVarP(cmdFlags, &socketAddr, "socket-addr", "", socketAddr, "<host:port> or absolute path (default: /run/docker/plugins/rclone.sock)")
+	flags.IntVarP(cmdFlags, &socketGid, "socket-gid", "", socketGid, "GID for unix socket (default: current process GID)")
+	flags.BoolVarP(cmdFlags, &forgetState, "forget-state", "", forgetState, "skip restoring previous state")
+	flags.BoolVarP(cmdFlags, &noSpec, "no-spec", "", noSpec, "do not write spec file")
+	// Add common mount/vfs flags
+	mountlib.AddFlags(cmdFlags)
+	vfsflags.AddFlags(cmdFlags)
+}
+
+// Command definition for cobra
+var Command = &cobra.Command{
+	Use:   "docker",
+	Short: `Serve any remote on docker's volume plugin API.`,
+	Long:  strings.ReplaceAll(longHelp, "|", "`") + vfs.Help,
+
+	Run: func(command *cobra.Command, args []string) {
+		cmd.CheckArgs(0, 0, command, args)
+		cmd.Run(false, false, command, func() error {
+			ctx := context.Background()
+			drv, err := NewDriver(ctx, baseDir, nil, nil, false, forgetState)
+			if err != nil {
+				return err
+			}
+			srv := NewServer(drv)
+			if socketAddr == "" {
+				// Listen on unix socket at /run/docker/plugins/<pluginName>.sock
+				return srv.ServeUnix(pluginName, socketGid)
+			}
+			if filepath.IsAbs(socketAddr) {
+				// Listen on unix socket at given path
+				return srv.ServeUnix(socketAddr, socketGid)
+			}
+			return srv.ServeTCP(socketAddr, "", nil, noSpec)
+		})
+	},
+}
--- a/docs/content/docker.md
+++ b/docs/content/docker.md
@@ -1,0 +1,526 @@
+---
+title: "Docker Volume Plugin"
+description: "Docker Volume Plugin"
+---
+
+# Docker Volume Plugin
+
+## Introduction
+
+Docker 1.9 has added support for creating
+[named volumes](https://docs.docker.com/storage/volumes/) via
+[command-line interface](https://docs.docker.com/engine/reference/commandline/volume_create/)
+and mounting them in containers as a way to share data between them.
+Since Docker 1.10 you can create named volumes with
+[Docker Compose](https://docs.docker.com/compose/) by descriptions in
+[docker-compose.yml](https://docs.docker.com/compose/compose-file/compose-file-v2/#volume-configuration-reference)
+files for use by container groups on a single host.
+As of Docker 1.12 volumes are supported by
+[Docker Swarm](https://docs.docker.com/engine/swarm/key-concepts/)
+included with Docker Engine and created from descriptions in
+[swarm compose v3](https://docs.docker.com/compose/compose-file/compose-file-v3/#volume-configuration-reference)
+files for use with _swarm stacks_ across multiple cluster nodes.
+
+[Docker Volume Plugins](https://docs.docker.com/engine/extend/plugins_volume/)
+augment the default `local` volume driver included in Docker with stateful
+volumes shared across containers and hosts. Unlike local volumes, your
+data will _not_ be deleted when such volume is removed. Plugins can run
+managed by the docker daemon, as a native system service
+(under systemd, _sysv_ or _upstart_) or as a standalone executable.
+Rclone can run as docker volume plugin in all these modes.
+It interacts with the local docker daemon
+via [plugin API](https://docs.docker.com/engine/extend/plugin_api/) and
+handles mounting of remote file systems into docker containers so it must
+run on the same host as the docker daemon or on every Swarm node.
+
+## Getting started
+
+In the first example we will use the [SFTP](/sftp/)
+rclone volume with Docker engine on a standalone Ubuntu machine.
+
+Start from [installing Docker](https://docs.docker.com/engine/install/)
+on the host.
+
+The _FUSE_ driver is a prerequisite for rclone mounting and should be
+installed on host:
+```
+sudo apt-get -y install fuse
+```
+
+Create two directories required by rclone docker plugin:
+```
+sudo mkdir -p /var/lib/docker-plugins/rclone/config
+sudo mkdir -p /var/lib/docker-plugins/rclone/cache
+```
+
+Install the managed rclone docker plugin:
+```
+docker plugin install rclone/docker-volume-rclone args="-v" --alias rclone --grant-all-permissions
+docker plugin list
+```
+
+Create your [SFTP volume](/sftp/#standard-options):
+```
+docker volume create firstvolume -d rclone -o type=sftp -o sftp-host=_hostname_ -o sftp-user=_username_ -o sftp-pass=_password_ -o allow-other=true
+```
+
+Note that since all options are static, you don't even have to run
+`rclone config` or create the `rclone.conf` file (but the `config` directory
+should still be present). In the simplest case you can use `localhost`
+as _hostname_ and your SSH credentials as _username_ and _password_.
+You can also change the remote path to your home directory on the host,
+for example `-o path=/home/username`.
+
+
+Time to create a test container and mount the volume into it:
+```
+docker run --rm -it -v firstvolume:/mnt --workdir /mnt ubuntu:latest bash
+```
+
+If all goes well, you will enter the new container and change right to
+the mounted SFTP remote. You can type `ls` to list the mounted directory
+or otherwise play with it. Type `exit` when you are done.
+The container will stop but the volume will stay, ready to be reused.
+When it's not needed anymore, remove it:
+```
+docker volume list
+docker volume remove firstvolume
+```
+
+Now let us try **something more elaborate**:
+[Google Drive](/drive/) volume on multi-node Docker Swarm.
+
+You should start from installing Docker and FUSE, creating plugin
+directories and installing rclone plugin on _every_ swarm node.
+Then [setup the Swarm](https://docs.docker.com/engine/swarm/swarm-mode/).
+
+Google Drive volumes need an access token which can be setup via web
+browser and will be periodically renewed by rclone. The managed
+plugin cannot run a browser so we will use a technique similar to the
+[rclone setup on a headless box](/remote_setup/).
+
+Run [rclone config](/commands/rclone_config_create/)
+on _another_ machine equipped with _web browser_ and graphical user interface.
+Create the [Google Drive remote](/drive/#standard-options).
+When done, transfer the resulting `rclone.conf` to the Swarm cluster
+and save as `/var/lib/docker-plugins/rclone/config/rclone.conf`
+on _every_ node. By default this location is accessible only to the
+root user so you will need appropriate privileges. The resulting config
+will look like this:
+```
+[gdrive]
+type = drive
+scope = drive
+drive_id = 1234567...
+root_folder_id = 0Abcd...
+token = {"access_token":...}
+```
+
+Now create the file named `example.yml` with a swarm stack description
+like this:
+```
+version: '3'
+services:
+  heimdall:
+    image: linuxserver/heimdall:latest
+    ports: [8080:80]
+    volumes: [configdata:/config]
+volumes:
+  configdata:
+    driver: rclone
+    driver_opts:
+      remote: 'gdrive:heimdall'
+      allow_other: 'true'
+      vfs_cache_mode: full
+      poll_interval: 0
+```
+
+and run the stack:
+```
+docker stack deploy example -c ./example.yml
+```
+
+After a few seconds docker will spread the parsed stack description
+over cluster, create the `example_heimdall` service on port _8080_,
+run service containers on one or more cluster nodes and request
+the `example_configdata` volume from rclone plugins on the node hosts.
+You can use the following commands to confirm results:
+```
+docker service ls
+docker service ps example_heimdall
+docker volume ls
+```
+
+Point your browser to `http://cluster.host.address:8080` and play with
+the service. Stop it with `docker stack remove example` when you are done.
+Note that the `example_configdata` volume(s) created on demand at the
+cluster nodes will not be automatically removed together with the stack
+but stay for future reuse. You can remove them manually by invoking
+the `docker volume remove example_configdata` command on every node.
+
+## Creating Volumes via CLI
+
+Volumes can be created with [docker volume create](https://docs.docker.com/engine/reference/commandline/volume_create/).
+Here are a few examples:
+```
+docker volume create vol1 -d rclone -o remote=storj: -o vfs-cache-mode=full
+docker volume create vol2 -d rclone -o remote=:tardigrade,access_grant=xxx:heimdall
+docker volume create vol3 -d rclone -o type=tardigrade -o path=heimdall -o tardigrade-access-grant=xxx -o poll-interval=0
+```
+
+Note the `-d rclone` flag that tells docker to request volume from the
+rclone driver. This works even if you installed managed driver by its full
+name `rclone/docker-volume-rclone` because you provided the `--alias rclone`
+option.
+
+Volumes can be inspected as follows:
+```
+docker volume list
+docker volume inspect vol1
+```
+
+## Volume Configuration
+
+Rclone flags and volume options are set via the `-o` flag to the
+`docker volume create` command. They include backend-specific parameters
+as well as mount and _VFS_ options. Also there are a few
+special `-o` options:
+`remote`, `fs`, `type`, `path`, `mount-type` and `persist`.
+
+`remote` determines an existing remote name from the config file, with
+trailing colon and optionally with a remote path. See the full syntax in
+the [rclone documentation](/docs/#syntax-of-remote-paths).
+This option can be aliased as `fs` to prevent confusion with the
+_remote_ parameter of such backends as _crypt_ or _alias_.
+
+The `remote=:backend:dir/subdir` syntax can be used to create
+[on-the-fly (config-less) remotes](/docs/#backend-path-to-dir),
+while the `type` and `path` options provide a simpler alternative for this.
+Using two split options
+```
+-o type=backend -o path=dir/subdir
+```
+is equivalent to the combined syntax
+```
+-o remote=:backend:dir/subdir
+```
+but is arguably easier to parameterize in scripts.
+The `path` part is optional.
+
+[Mount and VFS options](/commands/rclone_serve_docker/#options)
+as well as [backend parameters](/flags/#backend-flags) are named
+like their twin command-line flags without the `--` CLI prefix.
+Optionally you can use underscores instead of dashes in option names.
+For example, `--vfs-cache-mode full` becomes
+`-o vfs-cache-mode=full` or `-o vfs_cache_mode=full`.
+Boolean CLI flags without value will gain the `true` value, e.g.
+`--allow-other` becomes `-o allow-other=true` or `-o allow_other=true`.
+
+Please note that you can provide parameters only for the backend immediately
+referenced by the backend type of mounted `remote`.
+If this is a wrapping backend like _alias, chunker or crypt_, you cannot
+provide options for the referred to remote or backend. This limitation is
+imposed by the rclone connection string parser. The only workaround is to
+feed plugin with `rclone.conf` or configure plugin arguments (see below).
+
+## Special Volume Options
+
+`mount-type` determines the mount method and in general can be one of:
+`mount`, `cmount`, or `mount2`. This can be aliased as `mount_type`.
+It should be noted that the managed rclone docker plugin currently does
+not support the `cmount` method and `mount2` is rarely needed.
+This option defaults to the first found method, which is usually `mount`
+so you generally won't need it.
+
+`persist` is a reserved boolean (true/false) option.
+In future it will allow to persist on-the-fly remotes in the plugin
+`rclone.conf` file.
+
+## Connection Strings
+
+The `remote` value can be extended
+with [connection strings](/docs/#connection-strings)
+as an alternative way to supply backend parameters. This is equivalent
+to the `-o` backend options with one _syntactic difference_.
+Inside connection string the backend prefix must be dropped from parameter
+names but in the `-o param=value` array it must be present.
+For instance, compare the following option array
+```
+-o remote=:sftp:/home -o sftp-host=localhost
+```
+with equivalent connection string:
+```
+-o remote=:sftp,host=localhost:/home
+```
+This difference exists because flag options `-o key=val` include not only
+backend parameters but also mount/VFS flags and possibly other settings.
+Also it allows to discriminate the `remote` option from the `crypt-remote`
+(or similarly named backend parameters) and arguably simplifies scripting
+due to clearer value substitution.
+
+## Using with Swarm or Compose
+
+Both _Docker Swarm_ and _Docker Compose_ use
+[YAML](http://yaml.org/spec/1.2/spec.html)-formatted text files to describe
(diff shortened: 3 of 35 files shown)


ivandeex:
## Implementation details

### MountLib

I continued the refactoring started by @sapk on the [sapk-fork/serve-docker](https://github.com/sapk-fork/rclone/commits/serve-docker/cmd/serve/docker) branch. Not only do I factor `RunMount` from `cmd/mountlib/Mount` but also introduce the `MountPoint` structure with dedicated methods and separate `MountInfo` structure for marshalling mount state by the rclone `rcd` API. All refactoring commits as well as original commits from @sapk are **separate** for easier review.

### Docker Plugin SDK

I tried to continue using the [docker/go-plugins-helpers](https://github.com/docker/go-plugins-helpers) SDK but encountered a number of problems:
- We actually use only the Volume API which is less than 20% of SDK
- SDK spams system logs with trace messages and gives no way to disable it
- Systemd socket activation is declared but does not actually work
- SDK unconditionally tries to modify root-only files in `/run/docker/plugins` and `/etc/docker/plugins` and provides *no way* to change this behavior so testing API by non-root was impossible
- SDK depends on external package [docker/go-connections](https://github.com/docker/go-connections/sockets) which attempts to `chown` files adding another root-only dependency
- SDK adds more external dependencies in particular on Azure packages, this growing rclone executable too much

I created the [SDK fork](https://github.com/ivandeex/go-plugins-helpers), developed [a number of patches](https://github.com/ivandeex/go-plugins-helpers/commits/devel) and started submitting pull requests, in particular https://github.com/docker/go-plugins-helpers/pull/130. However, the SDK repository looks *a little bit abandoned*, still has its CI/CD based on the obsolete travis.org so even my first patch got stuck like [a number of other PRs](https://github.com/docker/go-plugins-helpers/pulls). Some other patches depend on unreleased features in [external docker packages](https://github.com/docker/go-connections) so I can't predict whether they can get through in reasonable time.

All in all, I decided to drop dependency on [docker/go-plugins-helpers](https://github.com/docker/go-plugins-helpers) altogether. The use of *go-chi http router* let the resulting code be very small (take a look at [api.go](https://github.com/ivandeex/rclone/blob/ci-docker/cmd/serve/docker/api.go) and [serve.go](https://github.com/ivandeex/rclone/blob/ci-docker/cmd/serve/docker/serve.go)) and get rid of all above problems. Now we can have a complete unit test for the new command.

### httplib and go-chi

I also tried to use the brand new [rclone/lib/http](https://github.com/rclone/rclone/tree/e489a101f631db31ec7c6d97ca4d8034c573031a/lib/http) but found it problematic. I love that it introduces *go-chi*, but it does that in a strange way. I needed to create a server listening on provided socket but the structure returned by [NewServer](https://github.com/rclone/rclone/blob/e489a101f631db31ec7c6d97ca4d8034c573031a/lib/http/http.go#L120) does not have a *Serve method*, at all. I tried to understand how the new `cmd server http` [uses the new library](https://github.com/rclone/rclone/blob/e489a101f631db31ec7c6d97ca4d8034c573031a/cmd/serve/http/http.go#L66) but I failed to wrap my head around who serves what. I guess that the real listening server is hidden as a singleton inside the library, internally attached to the `httplib.Router` and can be configured only via command line. :cry: By the way, what was the reasoning behind renaming `httplib` to `http` and then adding ugly import aliases in a dozen places inside rclone to decipher it from `net/http`? We could just have `rclone/lib/httplib` and simply use it directly. I abandoned further attempts to use this strange code and stick with good old `net/http`  dressed by the seasoned `go-chi`.

### Options

Every mount needs three groups of options: Backend, VFS and Mount options. The base fork has left this problem [for future](https://github.com/sapk-fork/rclone/blob/serve-docker/cmd/serve/docker/driver.go#L41). The generic Volume API will pass options as a `[string]string` map when the volume is created (following mount requests have no way to customize it, so I attach options to the volume not to its mounts). The docker command line provides for `docker volume create ... -o option1=value1 -o ...`, i.e. all options have a *flat scheme* with string names and values. Docker Swarm has a YAML format where options are represented by a yaml map. The least common divisor is the flag string-to-string dictionary.

Rclone provides a number of ways to pass options:
1. as command-line flags with lower-case dash-separated option names with backend name as a prefix,
2. as environment variables with upper-case underscore-separated names,
3. as `name = value` tokens in `rclone.conf` with lower-case underscore-separated names without backend prefix,
4. as connection string tokens similar to item 3 above (no backend prefix, only for the immediate connection string backend, impossible to pass underlying backend options in case of alias or crypt)

[TO BE CONTINUED]

### Mount Registry

The next design decision was how to maintain a volume/mount lifecycle. The plugin can run **(a)** in a special `runc` container beside normal docker containers (so called managed mode) or **(b)** as a process independent from docker daemon (under systemd or just invoked from command line). When it runs separately, it can be restarted independently from docker.

The plugin serves mounts to recipients via the **kernel FUSE module** and serves API on a Unix domain or TCP socket. Docker daemon can request volume create or removal requests to the plugin as well as volume mount/unmount requests. Docker passes mounts created by the plugin (on the host or in its container) to docker managed containers via bind mounting mechanism. Every volume can be mounted by several containers independently.

Containers can be killed or crash otherwise, docker daemon can be restarted on crash otherwise, the host can be rebooted, volumes can fail to created due to network problems. Docker swarm can activate a volume on multiple swarm nodes. What should give if the plugin got restarted after crash or reboot? Will docker re-send volume creation requests? Mount requests? How will docker daemon know what requests to resend? It will take care of most situations, e.g. it will restart the plugin and restart (I mean not just *start* but *stop-kill-restart*) the containers. If a container crashed, docker will restart it and pass *new* mount request to plugin with new *mounter ID*. Looks clean, almost... Unfortunately guys from Docker, Inc. who designed the API, have forgotten about the HEALTHCHECK protocol message so their implementation is not 100% clean.

What should the volume plugin do when it's started after crash or host reboot? The correct behavior is to recreate all volumes that existed before the crash. The same does not hold for volume mounts - the restarted plugin cannot restore mounts due to broken connection with storage providers. If a docker container will try to access a zombie mount, the process in it will get `Transport point not connected` from the kernel and that's correct. So I looked into existing volume plugin implementations and found that the prominent ones ([sshfs](https://github.com/vieux/docker-volume-sshfs/blob/v1.4/main.go#L50), [local-persist](https://github.com/MatchbookLab/local-persist/blob/master/driver.go#L231) [netshare](https://github.com/ContainX/docker-volume-netshare/blob/v0.36/netshare/netshare.go#L287), etc) do use a kind of volume registry persisted locally. Moreover, I encountered the ticket [moby#35120](https://github.com/moby/moby/issues/35120) where @cpuguy83 (a docker mainter) says in clear English:

> Yes, this is a flaw in the v1 plugin design itself which propagates itself through to the volume plugin API.
> Again, willing to take a patch to alleviate some of this (and we must understand the trade-offs), but do not generally want to modify the existing volume plugin API itself except for major issues as we are working on bringing CSI on and making modifications to the current API is risky. [...]
> However the plugin should still essentially **treat the caller as an idiot and keep track of it's on it's own** regardless of what capabilities the caller has.

Note that he mentions the industry standard **CSI** (container storage interface) API, that is currently used by such container orchestrators as Kubernetes or Nomad and is going to be implemented by Docker Swarm in future. I am going to submit `rclone serve csi` when I have resources for it. Hopefully its design is cleaner and CSI plugins can trust their plugin host a little more.

So... now we have the registry persisted in `docker-plugin.state` under rclone cache directory and unit tests for it. In case something gets stuck upon startup, we have the `--forget-state` flag that can be passed to the command and a way to pass it to managed plugin. The base fork [sapk-fork/server-docker](https://github.com/sapk-fork/rclone/tree/serve-docker) proposed [preserving the whole rclone.conf section](https://github.com/sapk-fork/rclone/blob/serve-docker/cmd/serve/docker/driver.go#L62) but I see problems with this approach, in particular how to pass dynamic provider tokens in and out of the managed plugin container. I think that state file is a better solution, waiting for your comments. Some drivers have the ability to [restore state from docker daemon](https://github.com/MatchbookLab/local-persist/blob/master/driver.go#L190) in addition to state file, but I think we a good enough for now and this can be left for future PRs.

### Documentation

I decided to split the short (put aside the VFS/mount options) command help from the long docker plugin documentation. Please take a look at [command help](https://github.com/ivandeex/rclone/blob/ci-docker/docs/content/commands/rclone_serve_docker.md#rclone-serve-docker) and [docker.md](https://github.com/ivandeex/rclone/blob/ci-docker/docs/content/docker.md#docker-volume-plugin).

@ncw @sapk
I am using this plugin on test rig for a month now.
Looking forward to your reviews :pray:

ncw:
This looks like a great bit of work @ivandeex :-)

I don't think the http lib refactoring is complete which is why it looks a bit odd at the moment. It might need some tidying up - can you add stuff to it to make it do what you want?

Do you want this for the 1.56 release? I'm assuming not - if that is correct I'll concentrate on getting the release out then have a look at your PR backlog.

ivandeex:
> I don't think the http lib refactoring is complete which is why it looks a bit odd at the moment. It might need some tidying up - can you add stuff to it to make it do what you want?

I looked again at the http code in `cmd/serve/docker/serve.go` - it's small and clean IMHO. Adding dependency on incomplete `lib/http` is a worthless extra work.

> Do you want this for the 1.56 release?

It will need a code review by you for which you don't have time. OTOH I have deployment harness which installs rclone from current master plys this patch on a staging rig (docker swarm, 3 ubuntu 18.04 master-workers + 2 pure workers) with a dozen docker stacks and volumes (backends: sftp, storj, amazon s3). It works fine for 2 months now, so I'm sure in the quality of this code.

> then have a look at your PR backlog.

#5352 is **ready** for 1.56 and has been reviewed once by you. it needs a 2nd review which is gonna be rather quick.

#5297 was already accepted by you. I added a representative unit test and committed it myself.

my other pulls are *not* ready:
- #5023 - postponed draft
- #4514, #5402 - waiting for upstream pr
- #5164, #5414 - both are big and require your time to learn code and review

Makefile
```diff
@@ -256,3 +256,40 @@ startstable:
 
 winzip:
 	zip -9 rclone-$(TAG).zip rclone.exe
+
+# docker volume plugin
+PLUGIN_IMAGE_USER := ivandeex
```
sapk, line 261:
```suggestion
PLUGIN_IMAGE_USER := rclone
```

sapk:
This seems pretty good and complete with even managed plugin format. I will try to test this.

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+The only way to deal with it is to pass `rclone.conf` to the plugin or set
+plugin args. This limitation is imposed by the connection string parser,
+"serve docker" just follows the suit.
+There is however a syntactic difference. Inside connection string the backend
+prefix must be dropped but in the --opt=x array it must be present. This is so
+because --opt=x can include mount/vfs flags now and possibly other parameters
+in future. Also it helps to discriminate between "remote" filesystem path and
+"crypt-remote" backend setting.
+
+## Using with Docker Compose
```
sapk, line 121: This part seems to be a copy-paste not updated. 
Maybe this is not needed since the version 3 example can provide enough information.

ivandeex: I looked again at `docker.md`. It has yet more confusing copy-paste places and needs even more cleanup. Sorry :(

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+
+Now install and enable the plugin:
+```
+docker plugin remove rclone
+docker plugin install rclone/docker-volume-rclone --grant-all-permissions --disable --alias rclone
+docker plugin set rclone RCLONE_VERBOSE=2 config=/etc/rclone args="--vfs-cache-mode=writes --allow-other"
+docker plugin enable rclone
+```
+
+The `-v /run/docker/plugins/:/run/docker/plugins/` part will make sure the
```
ivandeex, line 55: Managed plugin does not need "-v" for unix socket. This para must be removed.

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+`sock` file gets created at the right place. You also need to add one or more
+volumes to places you want to mount your volumes later at.
+
+**Note:** Plugin will fail if you make args empty, use e.g. "args=-v",
+due to the [docker bug](github/moby/plugin/v2/plugin.go:195).
+
+Lastly, the `-v /path/to/store/json/for/restart/:/var/lib/docker/plugin-data/`
```
ivandeex, line 62: Again, managed plugin does not need "-v" for state and cache. They will be saved by default in `/var/lib/docker/plugins/rclone/cache` (however you can customize location via `docker plugin set rclone cache=/path/to/custom/plugin/cache`). I must rewrite this paragraph.

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+part is so that the plugin can create a `json` file to know what volumes
+existed in case of a system restart, etc.
+
+When the container is destroyed, etc, it will look at a file it created in
```
ivandeex, line 66: "container destroyed" --> "plugin restarted or reinstalled"

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+`/var/lib/docker/plugins/rclone/docker-plugin.state` to recreate any volumes
```
ivandeex, line 67: correction: `/var/lib/docker/plugins/default/cache/docker-plugin.state` (or custom cache location)

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+
+Docker daemon can install plugins from an image registry and run them managed.
+We maintain the
+[docker-volume-rclone](https://hub.docker.com/r/rclone/docker-volume-rclone/)
+plugin image on the [Docker hub](https://hub.docker.com).
+
+Before installing the plugin, let's create a couple required directories as
+plugin will *not* create them automatically:
+```
+mkdir -p /var/lib/docker/plugins/rclone/config
```
ivandeex, line 43: I should mention that `rclone.conf` should be put here on every host where plugin is used.

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+docker volume create vol1 -d rclone -o remote=storj: -o vfs-cache-mode=full
+docker volume create vol2 -d rclone -o remote=:tardigrade,access_grant=xxx:heimdall
+docker volume create vol3 -d rclone -o type=tardigrade -o path=heimdall -o tardigrade-access-grant=xxx -o poll-interval=0
+docker volume inspect vol1
+```
+
+When you create a container, you can connect it to the Volume:
+```
+docker run --rm -it -v vol1:/mnt --workdir /mnt ubuntu:bionic
+```
```
ivandeex, line 84: I should mention that after the containter using rclone volume is destroyed and/or recreated by swarm manager on another worker host, the volume will be unmounted but the volume definition will not be removed by docker from the host. If you absolutely want to get rid of the volume, you must `docker volume rm myvol` it manually.

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+## Using with Swarm
+
+For this example you will need `rclone.conf` with the following section:
+```
+[storage_box]
+type = sftp
+host = u123456.your-storagebox.de
+user = u123456
+pass = ObscuredPasswordForSFTP
+```
```
ivandeex, line 150: I should mention that the section must be added to `/var/lib/docker/plugins/rclone/config/rclone.conf` on every swarm worker host.

ivandeex (commented):
docs need more work.

Makefile
```diff
@@ -256,3 +256,40 @@ startstable:
...
+PLUGIN_IMAGE_USER := rclone
+PLUGIN_IMAGE_TAG := latest
+PLUGIN_IMAGE_NAME := docker-volume-rclone
+PLUGIN_IMAGE := $(PLUGIN_IMAGE_USER)/$(PLUGIN_IMAGE_NAME):$(PLUGIN_IMAGE_TAG)
+PLUGIN_BASE_IMAGE := rclone/rclone:latest
+PLUGIN_BUILD_DIR := ./build/docker-plugin
+PLUGIN_CONTRIB_DIR := ./cmd/serve/docker/contrib/plugin
+PLUGIN_CONFIG := $(PLUGIN_CONTRIB_DIR)/config.json
+PLUGIN_DOCKERFILE := $(PLUGIN_CONTRIB_DIR)/Dockerfile
+PLUGIN_CONTAINER := docker-volue-rclone-dev-$(shell date +'%Y%m%d-%H%M%S')
```
ncw, line 270: There is a typo here `volue` should be `volume` I'm guessing?

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+mkdir -p /var/lib/docker/plugins/rclone/config
+mkdir -p /var/lib/docker/plugins/rclone/cache
+```
+
+Now install and enable the plugin:
+```
+docker plugin remove rclone
+docker plugin install rclone/docker-volume-rclone --grant-all-permissions --disable --alias rclone
+docker plugin set rclone RCLONE_VERBOSE=2 config=/etc/rclone args="--vfs-cache-mode=writes --allow-other"
```
ivandeex, line 51: Plugin verbosity can be tweaked via `docker plugin set args="... -v..."` or `docker plugin set RCLONE_VERBOSE=...`. Arguments are more generic, they support tweaking not only verbosity but mount/vfs defaults too. Also note that only `RCLONE_VERBOSE` environment variable can be passed into plugin.

ncw: One for the docs?

docs/content/docker.md
ncw: One for the docs?

ncw (approved):
> I looked again at the http code in `cmd/serve/docker/serve.go` - it's small and clean IMHO. Adding dependency on incomplete `lib/http` is a worthless extra work.

OK

> > Do you want this for the 1.56 release?
> 
> It will need a code review by you for which you don't have time.

It depends what level of code review!

I checked out the code and had an investigate. The majority of the new code is in `cmd/serve/docker` and that is fine and very low risk.

The only part with risk of breaking stuff is the `mountlib` refactor so that is what I will concentrate on in my review and I'll just skim the other stuff.

> OTOH I have deployment harness which installs rclone from current master plys this patch on a staging rig (docker swarm, 3 ubuntu 18.04 master-workers + 2 pure workers) with a dozen docker stacks and volumes (backends: sftp, storj, amazon s3). It works fine for 2 months now, so I'm sure in the quality of this code.

That sounds like most excellent testing :-)
 
> > then have a look at your PR backlog.
> 
> #5352 is **ready** for 1.56 and has been reviewed once by you. it needs a 2nd review which is gonna be rather quick.

Done

> #5297 was already accepted by you. I added a representative unit test and committed it myself.

Great
 
> my other pulls are _not_ ready:
> 
> * [drive: optimize performance for --min-age, --max-age (recovered) #5023](https://github.com/rclone/rclone/pull/5023) - postponed draft
> * [sia: setup unit tests #4514](https://github.com/rclone/rclone/pull/4514), [cmd/version: fix openbsd build (wip) #5402](https://github.com/rclone/rclone/pull/5402) - waiting for upstream pr
> * [rclone bisync #5164](https://github.com/rclone/rclone/pull/5164), [rclone as docker volume plugin #5415](https://github.com/rclone/rclone/pull/5415) - both are big and require your time to learn code and review

Thank you for the summary.

I'll give the mountlib refactor a look now and skim the rest of the code so we can merge this for 1.56.

I'd like to get your other PRs merged soon after we release 1.56 so we have time to fix them up and get testing before the 1.57 release.

Comments on this PR
- mountlib refactor looks OK!
- some of the commits need more appropriate titles
- some more docs?
- I added you to the docker hub team since you might need access to that

I'll approve this now so we can get it into 1.56 - merge when happy :-)

docs/content/docker.md
ivandeex: Sorry I don't understand what "one for the docs" means?

ivandeex:
>   * some more docs 

I will rewrite `docker.md` and fix typo `voume -> volume` in Makefile.

When you release 1.56, the new github job will build the docker plugin and push to docker hub. Hopefully fine. If not, we'll have to coordinate, e.g. push plugin image manually.

>   * some of the commits need more appropriate titles

@ncw Probably squash all commits into two:

commit 1 "refactor mountlib before adding serve docker command" (authored by sapk, co-authored by ivandeex) includes:
- mountlib: factor help string
- mountlib: refactor for docker server

commit 2 "implement serve docker command and build docker plugin" (authored by sapk, co-authored by ivandeex) includes:
- sapk: initial docker volume api
- serve/docker: implement docker volume
- serve/docker: vendorize docker helpers
- build: docker plugin
- docs: serve/docker
- Update PLUGIN_IMAGE_USER
- New commit fixing typo "voume"
- New commit with docs

docs/content/docker.md
```diff
@@ -0,0 +1,280 @@
...
+## Running Under Systemd
+
+**Note:** MacOS and Windows do not support native Docker plugins. Please use
+managed mode on these systems. Proceed further if you are on Linux.
+
+First, [install rclone](https://rclone.org/install).
+You can just run it (type `rclone serve docker` and hit enter) to test.
+
+Download two systemd configuration files:
+[docker-volume-rclone.service](https://raw.githubusercontent.com/ivandeex/rclone/ci-docker/cmd/serve/docker/contrib/systemd/docker-volume-rclone.service)
```
ivandeex, line 243: Correct URL: https://raw.githubusercontent.com/rclone/rclone/...

docs/content/docker.md
ivandeex: The docker-compose doc section is similar to the swarm section because docker guys architected swarm config format from the compose one. However, the two modes are different by nature. Compose runs on a single node creating standalone volumes. Swarm volumes are cloned across nodes under control of swarm manager. Their descriptions after initial parsing are stored in swarm log shared and negotiated by swarm masters. I will think a little now, then remove copy-paste artifacts and keep only relevant stuff.

sapk:
> docs need more work.
> 
> @sapk
> I hope with my inline comments your testing will be smooth.

Don't worry I should find my way and docs are not easy furthermore with the confusion introduced by the various ways and version of docker volume plugin system.
My advice on that side would be to keep it simple first and let people that need corner-cases and fine tuning find details in the parts under. I don't know the requirement on the docs side but my feeling is that it can always evolve to add details later if needed.

cmd/serve/docker/docker.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/rclone/rclone/cmd/mountlib"
+	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/vfs"
+	"github.com/rclone/rclone/vfs/vfsflags"
+)
+
+var (
+	pluginName  = "rclone"
+	pluginScope = "local"
+	baseDir     = "/var/lib/docker/plugins/rclone/volumes"
```
sapk, line 22: I am not a fan on writing in /var/lib/docker for this reason (explain in the note): https://docs.docker.com/engine/extend/plugins_volume/#volume-plugin-protocol
I think we should stick we default value for volume of https://github.com/docker/go-plugins-helpers/blob/master/volume/api.go#L12 
and find somewhere else to store cache and config.

ivandeex: /var/lib/docker-plugins/rclone/config
/var/lib/docker-plugins/rclone/cache
?

ncw:
> When you release 1.56, the new github job will build the docker plugin and push to docker hub. Hopefully fine. If not, we'll have to coordinate, e.g. push plugin image manually.

Cross fingers! I find it very hard to retry the docker hub stuff manually - you'll see various notes in RELEASE.md about it!
 
> > * some of the commits need more appropriate titles
> 
> Probably squash all commits into four?

Looks nice :-)

> **UPDATE**: done

:-)

ivandeex:
> > When you release 1.56, the new github job will build the docker plugin and push to docker hub. Hopefully fine.
> 
> Cross fingers!

@ncw
The plugin release job was tested on [ci branch](https://github.com/ivandeex/rclone/commits/ci-docker) **with success**. See [build log](https://github.com/ivandeex/rclone/runs/3019344969?check_suite_focus=true) and [results on docker hub](https://hub.docker.com/r/ivandeex/docker-volume-rclone/tags?page=1&ordering=last_updated).

@sapk
Note that latest managed plugin by default will look for config on host at `/var/lib/docker-plugins/rclone/config/rclone.conf` and maintains state at `/var/lib/docker-plugins/rclone/cache/docker-plugin.state`. Unmanaged plugin (systemd or standalone) will by default mount volumes under `/var/lib/docker-volumes/rclone`.

I'm finishing the docs.

ivandeex:
I am done with documentation.
It will soon appear on https://tip.rclone.org/docker
Merging now.

Issue #5381, opened 2021-06-05
Creating direct link with "rclone link"


KAction:
#### What is your current rclone version (output from `rclone version`)?

````
>>= rclone --version
rclone v1.55.1-DEV
- os/type: linux
- os/arch: amd64
- go/version: go1.16.4
- go/linking: dynamic
- go/tags: cmount
````

#### What problem are you are trying to solve?

I am trying to use rclone to automate backing up public, but endanged files. Once I had my build system breaking since file
disappeared from CTAN, so I'd rather mirror files I care about on multiple free hostings.

#### How do you think rclone should be changed to solve that?

rclone has "link" subcommand, but links generated with this subcommand are links to web-interface, not to file itself (at least I tried Microsoft OneDrive and  Koofr), so they can't be downloaded automatically with, e.g curl. 

I worked a lot with Dropbox and know that their public links end with "?dl=0", and if you replace it with "?dl=1", resulting link becomes direct. I think something similar should exist for other services, and this knowledge -- how to convert link to web-interface into direct link (I doubt usefulness of web-interface links in general, but this is another story) -- should be embedded into rclone code.

So I propose to introduce new option "--direct" to "link" subcommand, that should return direct link if it supported by service, or error-out if it is not supported or it is impossible to get it automatically.

If you agree with this idea, I would probably be able to provide patch for dropbox and koofr (I think I got their link logic).

albertony:
Haven't had the need for it myself, yet, but sounds like a good idea..

**Jottacloud:**
After some quick snooping around the web ui it seem to use direct download urls of form `https://www.jottacloud.com/opin/io/downloadPublic/<user_id>/<uuid>`
- where user_id is the internal id and not the name or login email of the user
- where uuid is share id except first three digits(?)
- where downloadPublic can be replaced with streamPublic or publicThumb for other features

ncw:
> rclone has "link" subcommand, but links generated with this subcommand are links to web-interface, not to file itself (at least I tried Microsoft OneDrive and Koofr), so they can't be downloaded automatically with, e.g curl.

`rclone link` should provide a link to the file directly (so usable with `curl`) if at all possible. It isn't always possible so it returns a web interface link otherwise.

So if you can figure out how to make any of the backends use a direct link instead of the web interface link then please send PRs to change them - no need for a `--direct` flag. :-)

KAction:
Okay. Let's start with #5398.

albertony:
[...]

PR #5417, opened 2021-06-20
onedrive: Make link return direct download link

Levi-Hope:

Make `rclone link onedrive` return direct download link.
Share link of file will be converted to direct link, but folder won't, since Onedrive manages folder download in a different and intransparent way.

https://github.com/rclone/rclone/issues/5381

#### Checklist

### DIFF
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -1507,43 +1507,80 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 	url := result.Link.WebURL
 
 	// Convert share link to direct download link if target is not a folder
+	cnvFailMsg := "Don't know how to convert share link to direct link - returning the link as is"
+	if info.Folder != nil {
+		fs.Logf(f, cnvFailMsg)
+		return url, nil
+	}
+
 	segments := strings.Split(url, "/")
-	cnvFailMsg := "Conversion method is outdated. Cannot convert to downloadable URL"
-	switch {
-	case f.driveType == driveTypePersonal:
-		// https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
+	switch f.driveType {
+	case driveTypePersonal:
+		/*
+			Method: https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
+		*/
 		if len(segments) != 5 {
 			fs.Logf(f, cnvFailMsg)
-		} else if segments[3] != "u" {
-			enc := base64.StdEncoding.EncodeToString([]byte(url))
-			url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
+			return url, nil
 		}
-	case f.driveType == driveTypeBusiness:
-		// https://docs.microsoft.com/en-us/sharepoint/dev/spfx/shorter-share-link-format
+
+		enc := base64.StdEncoding.EncodeToString([]byte(url))
+		strings.ReplaceAll(enc, "/", "_")
+		strings.ReplaceAll(enc, "+", "-")
+		url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
+
+	case driveTypeBusiness:
+		/*
+			Method: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/shorter-share-link-format
+			Example:
+				https://{tenant}-my.sharepoint.com/:t:/g/personal/{user_email}/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/personal/{user_email}/_layouts/15/download.aspx?share={Opaque_String}
+		*/
 		if len(segments) != 8 {
 			fs.Logf(f, cnvFailMsg)
-		} else if segments[3] != ":f:" {
-			url = strings.Join(segments[0:3], "/") + "/" + segments[5] + "/" + segments[6] +
-				"/_layouts/15/download.aspx?share=" + segments[7]
+			return url, nil
 		}
-	case f.driveType == driveTypeSharepoint:
+
+		url = strings.Join(segments[:3], "/") + "/" + segments[5] + "/" + segments[6] +
+			"/_layouts/15/download.aspx?share=" + segments[7]
+
+	case driveTypeSharepoint:
+		/*
+			Method: Similar to driveTypeBusiness
+			Example:
+				https://{tenant}-my.sharepoint.com/:t:/s/{site_name}/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/sites/{site_name}/_layouts/15/download.aspx?share={Opaque_String}
+
+				https://{tenant}-my.sharepoint.com/:t:/t/{team_name}/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/teams/{team_name}/_layouts/15/download.aspx?share={Opaque_String}
+
+				https://{tenant}-my.sharepoint.com/:t:/g/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/_layouts/15/download.aspx?share={Opaque_String}
+		*/
 		if len(segments) < 6 {
 			fs.Logf(f, cnvFailMsg)
-		} else if segments[3] != ":f:" {
-			url = strings.Join(segments[0:3], "/")
-			switch segments[4] {
-			case "s":
-				url += "/sites/" + segments[5]
-			case "t":
-				url += "/teams/" + segments[5]
-			case "g": // Root site
-			default:
-				fs.Logf(f, cnvFailMsg)
-			}
-			url += "/_layouts/15/download.aspx?share=" + segments[len(segments)-1]
+			return url, nil
 		}
+
+		url = strings.Join(segments[:3], "/")
+		switch segments[4] {
+		case "s": // Site
+			url += "/sites/" + segments[5]
+		case "t": // Team
+			url += "/teams/" + segments[5]
+		case "g": // Root site
+		default:
+			fs.Logf(f, cnvFailMsg)
+		}
+		url += "/_layouts/15/download.aspx?share=" + segments[len(segments)-1]
+
 	default:
-		fs.Logf(f, "Unsupported drive type. Cannot convert to downloadable URL")
+		fs.Logf(f, cnvFailMsg)
+		return url, nil
 	}
 
 	return url, nil
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -1523,12 +1523,10 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 			fs.Logf(f, cnvFailMsg)
 			return url, nil
 		}
-
 		enc := base64.StdEncoding.EncodeToString([]byte(url))
 		strings.ReplaceAll(enc, "/", "_")
 		strings.ReplaceAll(enc, "+", "-")
 		url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
-
 	case driveTypeBusiness:
 		/*
 			Method: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/shorter-share-link-format
@@ -1541,10 +1539,8 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 			fs.Logf(f, cnvFailMsg)
 			return url, nil
 		}
-
 		url = strings.Join(segments[:3], "/") + "/" + segments[5] + "/" + segments[6] +
 			"/_layouts/15/download.aspx?share=" + segments[7]
-
 	case driveTypeSharepoint:
 		/*
 			Method: Similar to driveTypeBusiness
@@ -1565,7 +1561,6 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 			fs.Logf(f, cnvFailMsg)
 			return url, nil
 		}
-
 		url = strings.Join(segments[:3], "/")
 		switch segments[4] {
 		case "s": // Site
@@ -1577,10 +1572,6 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 			fs.Logf(f, cnvFailMsg)
 		}
 		url += "/_layouts/15/download.aspx?share=" + segments[len(segments)-1]
-
-	default:
-		fs.Logf(f, cnvFailMsg)
-		return url, nil
 	}
 
 	return url, nil
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -1528,13 +1528,11 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration, 
 		enc = strings.ReplaceAll(enc, "=", "")
 		directURL = "https://api.onedrive.com/v1.0/shares/u!" + enc + "/root/content"
 	case driveTypeBusiness:
-
 		// Method: https://docs.microsoft.com/en-us/sharepoint/dev/spfx/shorter-share-link-format
 		// Example:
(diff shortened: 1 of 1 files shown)


backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,33 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
 		fmt.Println(err)
 		return "", err
 	}
-	return result.Link.WebURL, nil
+
+	url := result.Link.WebURL
+
+	// Convert share link to direct download link if target is not a folder
+	chunks := strings.Split(url, "/")
+	if f.driveType == driveTypePersonal && chunks[3] != "u" {
```
ivandeex, line 1511: You cannot expect that url path will always have 3+ segments.
Sooner or later Microsoft will change something and rclone will panic.
Next, the term "chunk" in rclone is usually related to data bytes, let's use something else.
Also log something loud if all your conditions failed.
Please rewrite this snippet like this:
```
segments := strings.Split(url, "/")
linkType := ""
if len(segments) >= 3 {
  linkType = segments[3]
}
switch {
case f.driveType == driveA && linkType == urlB:
  url = ...
case ...:
  ...
default:
  fs.Logf(..., "Unsupported link type. Cannot convert to downloadable URL")
}
```

Levi-Hope: Thanks for your review. Suggestions of terms and logs are great. But I'm a little bit confused about the panic handling. I have two concerns:
1. If Microsoft did change its format, this function will return something differently. Instead of waiting for people to notice that something is wrong, wouldn't it be nice to just let it panic so the problem can be noticed and fixed immediately?
2. If we do want to handle some changes, how about rewriting the snippet like this to handle it more properly:
```go
       	url := result.Link.WebURL
        segments := strings.Split(url, "/")
	switch {
	case f.driveType == driveTypePersonal:
		// https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
		if len(segments) != 5 {
			fs.Logf(f, "Microsoft might change URL format. Cannot convert to downloadable URL")
		} else if segments[3] != "u" {
			enc := base64.StdEncoding.EncodeToString([]byte(url))
			url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
		}
	case ...:
		...
	default:
		fs.Logf(...)
```
In the case of format changes, the original URL will be used.

ivandeex: > 1. ... let it panic so the problem can be noticed...

panic == crash, i.e. wrong program
instead if crashing we must log error

> 2. If we do want to handle some format changes, how about rewriting the snippet like this to handle it more properly:
> Since some case will access up to the 7th segment.
> ```go
> url := result.Link.WebURL
> segments := strings.Split(url, "/")
> switch {
> case f.driveType == driveTypeBusiness:
>     if len(segments) != 8 {
>         fs.Logf(f, "Conversion method is outdated. Cannot convert to downloadable URL")
>     } else if segments[3] != ":f:" {
>         url = strings.Join(segments[0:3], "/") + "/" + segments[5] + "/" + segments[6] +
>             "/_layouts/15/download.aspx?share=" + segments[7]
>     }

before accessing segment #3 this fragment checks that no. of segments is 8, so OK

> case ...:
>     ...
> default:
>     fs.Logf(...)
> ```
> In the case of format changes, the original URL will be used.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+		if len(segments) != 5 {
+			fs.Logf(f, cnvFailMsg)
+		} else if segments[3] != "u" {
+			enc := base64.StdEncoding.EncodeToString([]byte(url))
+			url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
+		}
+	case f.driveType == driveTypeBusiness:
+		// https://docs.microsoft.com/en-us/sharepoint/dev/spfx/shorter-share-link-format
+		if len(segments) != 8 {
+			fs.Logf(f, cnvFailMsg)
```
Cnly, line 1524: Generally I think [early returning](https://danp.net/posts/reducing-go-nesting/) is preferred when there's something failing and we don't have anything else to do afterwards. It helps with the readability. (The same applies to others)

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+		} else if segments[3] != ":f:" {
+			url = strings.Join(segments[0:3], "/") + "/" + segments[5] + "/" + segments[6] +
```
Cnly, line 1526:
```suggestion
			url = strings.Join(segments[:3], "/") + "/" + segments[5] + "/" + segments[6] +
```

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+				"/_layouts/15/download.aspx?share=" + segments[7]
+		}
+	case f.driveType == driveTypeSharepoint:
+		if len(segments) < 6 {
+			fs.Logf(f, cnvFailMsg)
+		} else if segments[3] != ":f:" {
+			url = strings.Join(segments[0:3], "/")
```
Cnly, line 1533:
```suggestion
			url = strings.Join(segments[:3], "/")
```

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+				url += "/sites/" + segments[5]
+			case "t":
+				url += "/teams/" + segments[5]
+			case "g": // Root site
+			default:
+				fs.Logf(f, cnvFailMsg)
+			}
+			url += "/_layouts/15/download.aspx?share=" + segments[len(segments)-1]
+		}
+	default:
```
Cnly, line 1545: We shouldn't need the default clause here as we've exhausted the possibilities for `f.driveType`.

Cnly (changes requested):
Thank you for your contribution @Levi-Hope! I have the following comments regarding this PR, some of which may be more important but others are not really so :)

PS: If possible, please don't force push as it makes reviews, discussions, etc. harder to follow.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
 	}
-	return result.Link.WebURL, nil
+
+	url := result.Link.WebURL
+
+	// Convert share link to direct download link if target is not a folder
+	segments := strings.Split(url, "/")
+	cnvFailMsg := "Conversion method is outdated. Cannot convert to downloadable URL"
+	switch {
+	case f.driveType == driveTypePersonal:
```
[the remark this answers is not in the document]
ivandeex: Agree.
It's perfectly fine to write
```
switch {
case a==123 && b==456:
...
}
```

But
```
switch {
case a==123:
...
case a==456:
...
}
```

should be simplified into
```
switch a {
case 123:
...
case 456:
...
}
```

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+		// https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
+		if len(segments) != 5 {
+			fs.Logf(f, cnvFailMsg)
```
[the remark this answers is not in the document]
Cnly: @ivandeex I think what you've said is reasonable. In that case Notice is fine for me. Also Notice is a lower level than Warn and can be suppressed with `-q` so I prefer it over Warn which can't be silenced.

I took a closer look and found that the docs of `Logf` says it uses the Info level (which has an `Infof` equivalent) but in fact it uses Notice which has no corresponding level-specific function. That said, the original `fs.Logf(f, cnvFailMsg)` call already looks perfect and needs no changes. Sorry for that!

I'll fix the docs in a while.

backend/onedrive/onedrive.go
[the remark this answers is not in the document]
Levi-Hope: Sure, this does get better.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
 		fmt.Println(err)
 		return "", err
 	}
-	return result.Link.WebURL, nil
+
+	url := result.Link.WebURL
+
+	// Convert share link to direct download link if target is not a folder
+	segments := strings.Split(url, "/")
+	cnvFailMsg := "Conversion method is outdated. Cannot convert to downloadable URL"
```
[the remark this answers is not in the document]
Levi-Hope: OK. I'm fine with it.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,50 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+	switch {
+	case f.driveType == driveTypePersonal:
+		// https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
+		if len(segments) != 5 {
+			fs.Logf(f, cnvFailMsg)
+		} else if segments[3] != "u" {
+			enc := base64.StdEncoding.EncodeToString([]byte(url))
+			url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
```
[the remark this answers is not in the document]
Levi-Hope: Yep, I will go check with this.

backend/onedrive/onedrive.go
Levi-Hope: Sounds like a good idea.

Cnly:
@Levi-Hope thank you for taking care of the comments. If you've resolved them, could you please also click on "Resolve conversation"?

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,78 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+	cnvFailMsg := "Don't know how to convert share link to direct link - returning the link as is"
+	if info.Folder != nil {
+		fs.Logf(f, cnvFailMsg)
+		return url, nil
+	}
+
+	segments := strings.Split(url, "/")
+	switch f.driveType {
+	case driveTypePersonal:
+		/*
```
Cnly, line 1519:
```suggestion
		// Method: https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
```

In Go, `//` line comments are [the norm](https://golang.org/doc/effective_go#commentary). And this can also be a one-line comment. (also applies to others)

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,78 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+			Method: https://stackoverflow.com/questions/37951114/direct-download-link-to-onedrive-file
+		*/
+		if len(segments) != 5 {
+			fs.Logf(f, cnvFailMsg)
+			return url, nil
+		}
+		enc := base64.StdEncoding.EncodeToString([]byte(url))
+		strings.ReplaceAll(enc, "/", "_")
+		strings.ReplaceAll(enc, "+", "-")
+		url = "https://api.onedrive.com/v1.0/shares/u!" + enc[:len(enc)-1] + "/root/content"
```
Cnly, line 1529: If I understand this correctly, you're just stripping off the last character from the encoded string unconditionally? Does this work when the last character is not a padding character, or if there are two padding characters at the end? (Please feel free to correct me.)

Levi-Hope:
> @Levi-Hope thank you for taking care of the comments. If you've resolved them, could you please also click on "Resolve conversation"?

Done.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,78 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+
+				https://{tenant}-my.sharepoint.com/:t:/t/{team_name}/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/teams/{team_name}/_layouts/15/download.aspx?share={Opaque_String}
+
+				https://{tenant}-my.sharepoint.com/:t:/g/{Opaque_String}
+				--convert to->
+				https://{tenant}-my.sharepoint.com/_layouts/15/download.aspx?share={Opaque_String}
+		*/
+		if len(segments) < 6 {
```
Cnly, line 1560: Maybe it'll be better to also put an upper bound here, just to be safe.

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,78 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+		}
+		url = strings.Join(segments[:3], "/")
+		switch segments[4] {
+		case "s": // Site
+			url += "/sites/" + segments[5]
+		case "t": // Team
+			url += "/teams/" + segments[5]
+		case "g": // Root site
+		default:
+			fs.Logf(f, cnvFailMsg)
```
Cnly, line 1572: If this path is taken, is the returned url usable? If not, we should probably just return the original one.

Cnly (changes requested):
Thanks :) We've got just a few things to discuss and do before we can get this merged.

backend/onedrive/onedrive.go
Levi-Hope: KK

backend/onedrive/onedrive.go
Levi-Hope: Well, actually the Microsoft API works even with the padding character included. But sure, let's strictly follow the format.

backend/onedrive/onedrive.go
Levi-Hope: KK

backend/onedrive/onedrive.go
Levi-Hope: KK

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,75 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
+		if len(segments) != 8 {
+			fs.Logf(f, cnvFailMsg)
+			return shareURL, nil
+		}
+		directURL = strings.Join(segments[:3], "/") + "/" + segments[5] + "/" + segments[6] +
+			"/_layouts/15/download.aspx?share=" + segments[7]
+	case driveTypeSharepoint:
+		// Method: Similar to driveTypeBusiness
+		// Example:
+		//   https://{tenant}-my.sharepoint.com/:t:/s/{site_name}/{Opaque_String}
```
Cnly, line 1545:
```suggestion
		//   https://{tenant}.sharepoint.com/:t:/s/{site_name}/{Opaque_String}
```

URLs for Sharepoint don't contain `-my`. (Same applies to others)

backend/onedrive/onedrive.go
```diff
@@ -1503,7 +1503,75 @@ func (f *Fs) PublicLink(ctx context.Context, remote string, expire fs.Duration,
...
 		return "", err
 	}
-	return result.Link.WebURL, nil
+
+	shareURL := result.Link.WebURL
+
+	// Convert share link to direct download link if target is not a folder
+	cnvFailMsg := "Don't know how to convert share link to direct link - returning the link as is"
+	if info.Folder != nil {
+		fs.Logf(f, cnvFailMsg)
```
Cnly, line 1512:
```suggestion
		fs.Debugf(nil, "Can't convert share link for folder to direct link - returning the link as is")
```

IMO we're not expecting share links for folders to be convertible so we don't need to disturb the user

Cnly (approved):
LGTM in general! I'm still running a few tests on my side now, and we'll be good to go once they're done! :D Meanwhile I just have some remaining nitpicking comments...

Cnly:
It just occurred to me that we should probably be conservative and refrain from blindly attempting the conversion for regional versions. I'll merge once the checks succeed. Thank you again! :)

Levi-Hope:
Great thanks to you guys, nice reviews.

PR #5423, opened 2021-06-24
docs: added tip to reduce SharePoint throttling - fixes #5404

### LINKED ISSUE #5404: Add tip to OneDrive documentation to avoid SharePoint throttling/blocking

https://forum.rclone.org/t/onedrive-errors/24677/13
https://forum.rclone.org/t/onedrvie-business-mount-problem/22561/6

v1.55.1

Facilitate a great tip to users experiencing extensive throttling or blockage by SharePoint (OneDrive Business).

Add a user agent tip from Microsoft to the [OneDrive Troubleshooting guide](https://rclone.org/onedrive/#troubleshooting) 

The wording could be something like:

> ##### Excessive throttling or blocked on SharePoint
> If you experience excessive throttling or is being blocked on SharePoint then it may help to set the user agent explicitly with a flag like this: `--user-agent "ISV|rclone.org|rclone/v1.55.1"`  
> 
> The specific details can be found in the Microsoft document: [Avoid getting throttled or blocked in SharePoint Online](https://docs.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-decorate-your-http-traffic-to-avoid-throttling)
> 

**Thanks to @jcastill for finding and testing the tip.**

I can do the Pull Request if approved and don't mind if jcastill or somebody else wants to take the change from here. 

#### How to use GitHub

olefrost:

To facilitate a great tip to users experiencing extensive throttling or blockage by SharePoint (OneDrive Business).

Yes, see issue #5404. 

#### Checklist

- [x] ~~I have added tests for all changes in this PR if appropriate.~~
- [x] ~~I have added documentation for the changes if appropriate.~~

### DIFF
--- a/docs/content/onedrive.md
+++ b/docs/content/onedrive.md
@@ -517,6 +517,12 @@ is a great way to see what it would do.
 
 ### Troubleshooting ###
 
+#### Excessive throttling or blocked on SharePoint
+
+If you experience excessive throttling or is being blocked on SharePoint then it may help to set the user agent explicitly with a flag like this: `--user-agent "ISV|rclone.org|rclone/v1.55.1"`  
+ 
+The specific details can be found in the Microsoft document: [Avoid getting throttled or blocked in SharePoint Online](https://docs.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-decorate-your-http-traffic-to-avoid-throttling)
+
 #### Unexpected file size/hash differences on Sharepoint ####
 
 It is a


jcastilloalonso:
Might modify the file to use rclone version and not hardcode the version in the documentation. This will create issues down the road when customers upgrade and keep using old version as identification. Even worse if there's an issue with certain version and Microsoft decides to block any version older than. Otherwise it looks great!

olefrost:
Thanks @jcastilloalonso! 

Good point. I did consider this but haven’t found a way to refer to the current version in the docs. I therefore intentionally used the phrase “like this” to convey that some attention/thinking was required when setting the user-agent. Some might also copy the user-agent directly into script and a slightly lacking version number may actually highlight the risks with this. Do you know a better phrase/formulation to trigger adequate attention/thinking? (English is my second language)

Otherwise, I think we are in line with the rest of the documentation. It generally assumes a reader with a solid understanding of IT and skills to identify the consequences of additional parameters/tweaking. As an example, consider the consequences of following [this OneDrive tip](https://rclone.org/onedrive/#unexpected-file-size-hash-differences-on-sharepoint) without taking the necessary precautions - the tip comes with implicit assumptions about the use case and protective measures in place - you can easily lose data if they aren’t met/implemented.

ncw:
Do you think rclone should be doing this automatically would be my only question?

olefrost:
@ncw I thought this was the preferred solution by @jcastilloalonso, but now that I am rereading the posts I see that I probably misunderstood. I apologise for the confusion.

I did briefly consider a code change and then decided to go for the documentation tip. First, because I do not have a OneDrive Business account for development and test. Second, because the Microsoft guidance doesn’t seem to follow the international standard for user agents (RFC2616 section 3.8). I am not an HTTP expert, but I guess it could cause other issues if the traffic passes a sniffing router, firewall, proxy etc. 

Perhaps it would be better to use “rclone/v1.55.1 ISV|rclone.org”. This seems better in line with the standard and also seems within Microsoft’s note in [the guide](https://docs.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-decorate-your-http-traffic-to-avoid-throttling): “Format of the user agent string is expected to follow RFC2616, so please follow up on the above guidance on the right separators. its also fine to append existing user agent string with the requested information.” 

I will be happy to update the doc tip. I am hesitating to do the code change due to my current tools, knowledge and time.

@jcastilloalonso: What do you think?

jcastilloalonso:
@ncw If we do a code change I might recommend to implement as a flag (at least initially) as lot of people currently use SharePoint without it. Based on the same documentation seems Microsoft checks previous usage as reference and really hates sporadic exponential use of the service and rather have users keep a predicable/repetitive use of the service. Even the documentation says traffic will still be throttled, just given priority over people not using the expected user agent.

I agree with @olefrost, documentation looks kinda confusing and ideally we would need to reach out to Microsoft for clarification and expectations. This also supports a flag first implementation so we can further test and fix before pushing a code that might get rclone flagged and banned by Microsoft for not following their expected behavior.

Would be interesting to see if other open source applications are doing this and how is their implementation or interpretation.

ncw:
OK, let's stick with the docs for the moment and we can think about a flag later.

@olefrost ping me when you are happy with the docs!

olefrost:
I am happy with the docs as there are.
 
The current non-standard tip (ISV|rclone.org|rclone/v1.55.1) has a proven positive effect in some situations. Should somebody experience negative effects then they will just fall back to the (standard compliant) default. That is a good help to the affected users with minimum risk to the rest.

@jcastilloalonso  How about you? Anything you would like to have reworded or changed before I ping ncw?

jcastilloalonso:
I'm good with the docs as well.

olefrost:
@ncw We are happy; PR ready for merge

ncw:
Thank you @olefrost and @jcastilloalonso - I'll merge now :-)

PR #5431, opened 2021-06-29
Drive: Support parsing id from URL Drive 

arlessweschler:

for improved use for 1 remote with all access to teamdrives joined or shared folders by id
~and do not configure multiple remotes for copy / size / lsd by path only.~
usage: rclone lsd/move/copy/sync/size/ remote:{ID} 

https://forum.rclone.org/t/copy-files-using-drive-file-id/18881
https://forum.rclone.org/t/copy-gdrive-folder-by-id-to-another-gdrive-folder/17574
https://forum.rclone.org/t/can-i-copy-file-folder-using-their-id-to-my-gdrive/17358

#### Checklist

-- [ ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

### DIFF
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -18,7 +18,7 @@ FUSE.
 
 First set up your remote using `rclone config`.  Check it works with `rclone ls` etc.
 
-On Linux and OSX, you can either run mount in foreground mode or background (daemon) mode.
+On Linux and macOS, you can either run mount in foreground mode or background (daemon) mode.
 Mount runs in foreground mode by default, use the `--daemon` flag to specify background mode.
 You can only run mount in foreground mode on Windows.
 


ncw:
I would have thought you could do this now using the [connection string](https://rclone.org/docs/#connection-strings) syntax?

Eg assuming you have a remote called `drive` already

    rclone lsf drive,root_folder_id=XXXXX:

arlessweschler:
> I would have thought you could do this now using the [connection string](https://rclone.org/docs/#connection-strings) syntax?
> 
> Eg assuming you have a remote called `drive` already
> 
> ```
> rclone lsf drive,root_folder_id=XXXXX:
> ```

IMG_20210706_204829_460

I have compiled and tested ID support for Google Drive only. available for copy, sync, lsd, move commands 

IMG_20210706_204828_155

here 1 remote already configured as teamdrive, called arless. 
1 is my remote specified, I used size to calculate size. 
2 I calculated the size  by ID of a [public link](https://drive.google.com/drive/u/0/mobile/folders/1PDYZo2zMIJjERU3HqW-noSQACSqHuYPR?usp=sharing)
For copy , sync and move, simple usage

`rclone copy arless:{ID SOURCE} arless:{ID_DEST}`
Also it can specify a custom folder name  in the destination by slash and, between quotation if the name contains spaces.
`rclone copy arless:{ID SOURCE} "arless:{ID_DEST}/courses reader/"`
it is also very useful and necessary to simplify the transfer to other backends, for example, Mega.

ncw:
Why doesn't this work for you?

    rclone copy arless,root_folder_id=ID_SOURCE: arless,root_folder_id=ID_DEST:

I don't want to add more syntax to the rclone connection strings if we don't have to.

Issue #3673, opened 2019-10-27
Implement an ioctl to get a file's hash


eatnumber1:
I'm building a tool to verify files against a database of known good hashes. The files are stored in cloud storage, so it would be useful if I could leverage the cloud backend's native method of getting a file's hash without having to download the entire file.

It seems like this could be solved in a useful way by implementing some new ioctls, e.g. `RCLONE_IOC_MD5`, `RCLONE_IOC_SHA1`, etc. that my application could call on a mounted filesystem.

Does that sound reasonable / feasible? I saw issue #3170 and bazil/fuse#93, but sadly Bazil is starting to seem like an abandoned project, with its last commit more than a year ago.

ncw:
Unless bazil/fuse supports it then will be impossible!

The other thing you could do is run your rclone mount with `--rc` and use the [rclone API to read hashes:](https://rclone.org/rc/#operations/list).

animosity22:
This looks to be answered. If you have any other questions, please ask on https://forums.rclone.org

Thanks!

Issue #1710, opened 2017-09-30
Add extended attributes


vesper8:
I'm trying very hard to restore my macOS system from b2 after copying it with Rclone.

I'm now realizing that Rclone didn't save extended attributes and that means that all my macOS Applications won't open at all since they're missing those attributes.

I know it's been brought up a few times before but if support for xattrs could be added at some point that would be very useful

otherwise.. thankfully I still have the original files with the attributes on them. I was hoping to easily substitute rclone by rsync in my script.. but it isn't working out because rsync doesn't support filter-from and their include/exclude/include-from/exclude-from works differently from rclone's. Rclone is much more intuitive and I'm finding it impossible to use the same filters =/

However rsync has a --files-from which says "rsync all those files in a file and nothing else" This would work perfectly if I could get Rclone to produce a list of files that match Rclone's filters.. is there a way to do that? I realize Rclone's dry-run comes close but then a bunch of text would need to be grepped out so rsync would accept it as a list of filepaths

any help would be.. really appreciated!

vesper8:
@ncw would it be possible and maybe not too difficult to keep extended attributes when doing local to local rclone at least?

btw.. I managed to create a list of files that match the rclone filter in a format that is usable by rsync's files-from

it's ugly.. because my regex is quite poor.. but it does work. Basically I did a --dry-run from source -> empty folder with rclone and then cleaned up the extra noise with sed:

```
logFile=${1}

sed -E -i.bak 's#^[0-9]{1,4}/[0-9]{1,2}/[0-9]{1,2} [0-9]{1,2}:[0-9]{1,2}:[0-9]{1,2} [a-zA-Z ]{1,6}: ##' $logFile
sed -E -i.bak '/^Local file system at/d' $logFile
sed -E -i.bak 's/: Not copying as --dry-run//g' $logFile

sed -E -i.bak '/^Transferred:/d' $logFile
sed -E -i.bak '/^Errors:/d' $logFile
sed -E -i.bak '/^Checks:/d' $logFile
sed -E -i.bak '/^Transferred:/d' $logFile
sed -E -i.bak '/^Elapsed time:/d' $logFile
```

ncw:
> I'm trying very hard to restore my macOS system from b2 after copying it with Rclone.
> 
> I'm now realizing that Rclone didn't save extended attributes and that means that all my macOS Applications won't open at all since they're missing those attributes.

[...]

Issue #1190, opened 2017-03-01
[Feature Request] Store checksums for plain and encrypted files in xattrs


gordan-bobic:
It would be really handy if there was a feature to label local files with md5 checksums both for plain files as they are locally and encrypted versions (if they are uploaded with encryption enabled).

Many cloud back-end services support checksums to be checked without retrieving the entire file. This could then be checked against the checksum precomputed locally, including against the encrypted version of the file.

Extended attributes (if available) seem like an obvious place to store the checksums in question.

As a side-effect, this would allow for error checking the local store against bit rot as well.

ncw:
You can use the new cryptcheck command in the latest beta for checking checksums on crypted filesystems.

----
rclone cryptcheck checks a remote against a crypted remote.  This is
the equivalent of running rclone check, but able to check the
checksums of the crypted remote.

For it to work the underlying remote of the cryptedremote must support
some kind of checksum.

It works by reading the nonce from each file on the cryptedremote: and
using that to encrypt each file on the remote:.  It then checks the
checksum of the underlying file on the cryptedremote: against the
checksum of the file it has just encrypted.

Use it like this

    rclone cryptcheck /path/to/files encryptedremote:path

You can use it like this also, but that will involve downloading all
the files in remote:path.

    rclone cryptcheck remote:path encryptedremote:path

After it has run it will log the status of the encryptedremote:.

AzureBlaze:
@ncw I'm not really sure where to find info on these newer commands, does this mean --checksum will also work on encrypted remotes?

ncw:
@azureblaze the info from the docs is above.  I'm planning on a full release this weekend which will update the docs on the website.

It doesn't mean --checksum will work, no, it is a specialised mode for checking only.

gordan-bobic:
Not quite the same thing, what I was asking for is local checksum "caching" in xattrs. When a file is uploaded, compute the encrypted checksum, and store it in an xattr. Then when a check is requested, compare that checksum with the checksum on the object storage side. This would allow a very fast check that doesn't require reading all the files locally.

The local health check of the files can then be entirely separate (e.g. zfs scrub).

DurvalMenezes:
Hello @gordan-bobic,

> Not quite the same thing, what I was asking for is local checksum "caching" in xattrs. 

I understand @ncw is planning to do this as part of #637.

Cheers, 
   Durval.

PR #5438, opened 2021-07-04
Expose file hashes as extended attributes.

eatnumber1:
Hey there, in spite of the checklist below, I'd like some feedback on this direction before I put effort into making tests and implementing it for all the fuse backends. Can you take an early look?

This change adds extended attributes found at `system.hash.<hash_type>`.
By reading these extended attributes, rclone will query and return the
hash from the storage backend without needing to download the file.

The following is an example of using these attributes when reading a
file stored in Google Drive using the Mac OSX `xattr` tool.

```
$ xattr image.jpg
system.hash.md5
$ xattr -p system.hash.md5 image.jpg
920e3d7899c8eed81b50d6f26fe0f64b
```

Not directly, although there's been discussion about exposing hashes through the fs layer, and discussion about extended attributes in the past. The topics I've found are: #1710, #1190, and #3673.

#### Checklist


ncw:
I had a look through the code and it looks sensible to me :-)

What other options are there for extended attributes we could expose?

Are there any standards docs on extended attributes?

I'll get the CI to run on this change and see if it has any feedback.

eatnumber1:
> Are there any standards docs on extended attributes?

Not that I can tell. From Linux's [xattr(7)](https://man7.org/linux/man-pages/man7/xattr.7.html):

```
CONFORMING TO
       Extended attributes are not specified in POSIX.1, but some other
       systems (e.g., the BSDs and Solaris) provide a similar feature.
```

With respect to documentation though, each platform has their own documentation. Here's the ones for [Linux](https://man7.org/linux/man-pages/man7/xattr.7.html), [OSX](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getxattr.2.html), and [FreeBSD](https://www.freebsd.org/cgi/man.cgi?extattr). Supposedly there's also [equivalent functionality](https://en.wikipedia.org/wiki/Extended_file_attributes#Windows_NT) in Windows.

None of the platforms have completely compatible implementations, but the FUSE layer provides some amount of portability for FUSE filesystems. In the case of my code, I've tested it on OSX, and I'm fairly confident that it will work on Linux (in part because I'm conforming to Linux's namespacing rules), but I'll need to do a bit more research into how to get it working on BSD and especially Windows. If I'm lucky, it'll just work.

Another thing to consider here is my choice of exactly what key to use. I'm currently using `system.hash.<hash_type>`, where on Linux the `system` prefix is documented as follows:

```
  System extended attributes
      System extended attributes are used by the kernel to store system
      objects such as Access Control Lists.  Read and write access
      permissions to system attributes depend on the policy implemented
      for each system attribute implemented by filesystems in the
      kernel.
```

True we're not the kernel, but we're sort of a stand-in for the kernel here. I also don't think there's any actual enforcement that prevents us from using the system namespace at the FUSE layer, so it's somewhat of only a theoretical problem.

More practically though, there's other tools that store checksums as xattrs, so it might be useful to figure out if there's a de-facto standard for checksums-in-xattrs emerging, and reuse that. I did a bit of searching for this, and am not finding any consensus.

- [rmlint](https://rmlint.readthedocs.io/en/latest/) uses `user.rmlint.$USER.cksum`
- glusterfs [seems to use](https://bugzilla.redhat.com/show_bug.cgi?id=1208124) `trusted.glusterfs.bit-rot.signature`
- cshatag [uses](https://github.com/rfjakob/cshatag/blob/fe9aae114aa05987e8e54617dfeee088fb7f9bb7/check.go#L18) `user.shatag.sha256`
- various documentation about doing this manually seems to use [`user.checksum`](https://wiki.archlinux.org/title/File_permissions_and_attributes#User_extended_attributes) and [`user.checksum.md5`](https://www.linuxtoday.com/blog/extended-file-attributes-rock.html)

The Linux xattr documentation does give some examples, which includes an md5 hash:

```
  Extended attribute namespaces
      Attribute names are null-terminated strings.  The attribute name
      is always specified in the fully qualified namespace.attribute
      form, for example, user.mime_type, trusted.md5sum,
      system.posix_acl_access, or security.selinux.
```

So yeah, no clear consensus on what to use. Most tools seem to put their name in the key to prevent collisions, so there doesn't seem to be much attempt at reuse. Happy to do that for rclone too, or change it to `trusted.md5sum` to follow the example from the Linux docs (which I didn't notice until just now).

> What other options are there for extended attributes we could expose?

Hm, good question. For me my current goal is to get some ROM [management tooling](https://github.com/SmokeMonsterPacks/EverDrive-Packs-Lists-Database) to support organizing ROMs without downloading them (it manages them using the file's hashes), so I hadn't put much thought into other things we could export via xattrs.

Looking around a bit, I found https://www.freedesktop.org/wiki/CommonExtendedAttributes/ which lists both some useful design practices for naming xattrs, and some common ones in use today along with some proposed ones. Ones that seem useful for us include:

- `user.mime_type`
- `user.charset` (do we actually know the charset of files when applicable?)
- `user.xdg.origin.url` - perhaps put the URL from the backend here?

Looking at the Google Drive API [docs](https://developers.google.com/drive/api/v3/reference/files), there's a few more things that could be exported (although I don't know if it would actually be useful to do so), including

 - description
 - parents (since there can be more than one)
 - properties (this would be the read side of supporting arbitrary user specified xattrs)
 - version
 - thumbnailLink (I did some searching and it looks like thumbnails in most apps don't use xattrs, so proper thumbnail support would be a separate effort)
 - and a bunch more...

Then of course there would be the (mostly separate) work to allow user-defined xattrs.

> I'll get the CI to run on this change and see if it has any feedback.

Thanks! I'll finish up the other fuse backends, and write some tests and documentation in a bit.

cmd/cmount/fs.go
```diff
@@ -550,7 +550,22 @@ func (fsys *FS) Setxattr(path string, name string, value []byte, flags int) (err
 
 // Getxattr gets extended attributes.
 func (fsys *FS) Getxattr(path string, name string) (errc int, value []byte) {
-	return -fuse.ENOSYS, nil
+	defer log.Trace(path, "name=%q", name)("errc=%d", &errc)
+
+	node, handle, errc := fsys.getNode(path, fhUnset)
```
eatnumber1, line 555: Do you think it's worth grabbing the handle here considering we're passing `fhUnset` into `getNode`? From what I can tell, `getNode` will always return a `nil` handle in this case, so adding the code to cover the case where it doesn't would only be future-proofing if that behavior ever changes.

eatnumber1:
Another thought: if we do eventually add support for user-defined xattrs, we could have special ones which set specific properties on the backend. For example, `user.gdrive.description` could set `description`, or `user.gdrive.starred` could set `starred`.

cmd/cmount/fs.go
ncw: If we don't have an `fh` then probably best to explicitly call this instead.

    node, errc := fsys.lookupNode(path)

ncw (commented):
Can you check out the build failures in the CI - looks like there is more stuff to add!

ncw:
> > Are there any standards docs on extended attributes?
> 
> Not that I can tell. From Linux's [xattr(7)](https://man7.org/linux/man-pages/man7/xattr.7.html):
> 
> ```
> CONFORMING TO
>        Extended attributes are not specified in POSIX.1, but some other
>        systems (e.g., the BSDs and Solaris) provide a similar feature.
> ```
> 
> With respect to documentation though, each platform has their own documentation. Here's the ones for [Linux](https://man7.org/linux/man-pages/man7/xattr.7.html), [OSX](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getxattr.2.html), and [FreeBSD](https://www.freebsd.org/cgi/man.cgi?extattr). Supposedly there's also [equivalent functionality](https://en.wikipedia.org/wiki/Extended_file_attributes#Windows_NT) in Windows.
> 
> None of the platforms have completely compatible implementations, but the FUSE layer provides some amount of portability for FUSE filesystems. In the case of my code, I've tested it on OSX, and I'm fairly confident that it will work on Linux (in part because I'm conforming to Linux's namespacing rules), but I'll need to do a bit more research into how to get it working on BSD and especially Windows. If I'm lucky, it'll just work.

Cross fingers! I'm not sure how WinFSP deals with attributes.

> Another thing to consider here is my choice of exactly what key to use. I'm currently using `system.hash.<hash_type>`, where on Linux the `system` prefix is documented as follows:
> 
> ```
>   System extended attributes
>       System extended attributes are used by the kernel to store system
>       objects such as Access Control Lists.  Read and write access
>       permissions to system attributes depend on the policy implemented
>       for each system attribute implemented by filesystems in the
>       kernel.
> ```
> True we're not the kernel, but we're sort of a stand-in for the kernel here. I also don't think there's any actual enforcement that prevents us from using the system namespace at the FUSE layer, so it's somewhat of only a theoretical problem.

Yes I would say `system` is appropriate here. As a FUSE fs we are standing in for the kernel. `user` prefix is for random stuff the user wants to store.

> More practically though, there's other tools that store checksums as xattrs, so it might be useful to figure out if there's a de-facto standard for checksums-in-xattrs emerging, and reuse that.

Good idea

> I did a bit of searching for this, and am not finding any consensus.
> 
> * [rmlint](https://rmlint.readthedocs.io/en/latest/) uses `user.rmlint.$USER.cksum`
> * glusterfs [seems to use](https://bugzilla.redhat.com/show_bug.cgi?id=1208124) `trusted.glusterfs.bit-rot.signature`
> * cshatag [uses](https://github.com/rfjakob/cshatag/blob/fe9aae114aa05987e8e54617dfeee088fb7f9bb7/check.go#L18) `user.shatag.sha256`
> * various documentation about doing this manually seems to use [`user.checksum`](https://wiki.archlinux.org/title/File_permissions_and_attributes#User_extended_attributes) and [`user.checksum.md5`](https://www.linuxtoday.com/blog/extended-file-attributes-rock.html)
> 
> The Linux xattr documentation does give some examples, which includes an md5 hash:
> 
> ```
>   Extended attribute namespaces
>       Attribute names are null-terminated strings.  The attribute name
>       is always specified in the fully qualified namespace.attribute
>       form, for example, user.mime_type, trusted.md5sum,
>       system.posix_acl_access, or security.selinux.
> ```
> 
> So yeah, no clear consensus on what to use. Most tools seem to put their name in the key to prevent collisions, so there doesn't seem to be much attempt at reuse. Happy to do that for rclone too, or change it to `trusted.md5sum` to follow the example from the Linux docs (which I didn't notice until just now).

I guess `rclone.md5sum` is absolutely explicit and won't step on anyone's toes? I'd rather use a standard name if we can figure out the best one though.

> > What other options are there for extended attributes we could expose?
> 
> Hm, good question. For me my current goal is to get some ROM [management tooling](https://github.com/SmokeMonsterPacks/EverDrive-Packs-Lists-Database) to support organizing ROMs without downloading them (it manages them using the file's hashes), so I hadn't put much thought into other things we could export via xattrs.
> 
> Looking around a bit, I found https://www.freedesktop.org/wiki/CommonExtendedAttributes/ which lists both some useful design practices for naming xattrs, and some common ones in use today along with some proposed ones. Ones that seem useful for us include:
> 
> * `user.mime_type`

This would be easy to add. You can find the mime type of any `Object` with the `fs.MimeType()` call.

> * `user.charset` (do we actually know the charset of files when applicable?)

No.

> * `user.xdg.origin.url` - perhaps put the URL from the backend here?

This could potentially be the output of `rclone link` (the internal method is called PublicLink).

I think maybe we should add `user.mime_type` as well 1) because it is useful, and 2) to generalise the very special case system we have at the moment.

> Looking at the Google Drive API [docs](https://developers.google.com/drive/api/v3/reference/files), there's a few more things that could be exported (although I don't know if it would actually be useful to do so), including
> 
> * description
> * parents (since there can be more than one)
> * properties (this would be the read side of supporting arbitrary user specified xattrs)
> * version
> * thumbnailLink (I did some searching and it looks like thumbnails in most apps don't use xattrs, so proper thumbnail support would be a separate effort)
> * and a bunch more...

Do be useful they would need to exposed via the Object layer which they aren't right now.

The format specifiers for `rclone lsf` give an idea of things which would be easy to export.

```
p - path
s - size
t - modification time
h - hash
i - ID of object
o - Original ID of underlying object
m - MimeType of object if known
e - encrypted name
T - tier of storage if known, e.g. "Hot" or "Cool"
```

> Then of course there would be the (mostly separate) work to allow user-defined xattrs.

This would need support from the backend to store the xattrs which is part of the general metadata for each backend project.

> > I'll get the CI to run on this change and see if it has any feedback.

It did!

> Another thought: if we do eventually add support for user-defined xattrs, we could have special ones which set specific properties on the backend. For example, `user.gdrive.description` could set `description`, or `user.gdrive.starred` could set `starred`.

That is possible certainly...

What I envision eventually is that each backend exposes a set of metadata, some of which will be backend specific (eg `starred` and some of which may be set by the user). This would then be exposed as xattrs and when copying from backend to backend "compatible" metadata would be copied, with the rest being stored as user keys.

eatnumber1:
Okay I've fleshed this out a bit. Support is in place for both the cmount and mount2 backends, and some (very simplistic) test are added to the vfs layer. In addition, I added the user.mime_type attribute.

Please take a look at the commit messages, as I made a few changes.

As for what attribute to put the hashes under, I ended up settling for `system.org.rclone.hash` to follow the freedesktop.org recommendations. Happy to change it if you'd prefer something else though.

I've actually also implemented support in the mount backend, but I wasn't able to test it (I wasn't even able to compile it) as I'm testing on OSX. For that reason, the changes to support the mount backend aren't part of this PR but can be seen [here](https://github.com/eatnumber1/rclone/commit/d68f1dd5c98e8c3f5c98c579d6d04eca3fcbcf7c). What, if anything, would you like me to do with it?

> Can you check out the build failures in the CI - looks like there is more stuff to add!

I moved the xattr logic out of the Handle (since we don't actually need an open file to read xattrs) which fixed it. I tried replicating the CI a bit with the following:

```
$ RCLONE_CONFIG="/notfound" go test -tags "cmount" ./...
```

... and that passed on my mac.

So if the CI runs again and is happy with my changes, I think the only remaining thing is to add documentation. Do you have a suggestion on the best place to put docs on this feature?

Also to be clear, I've tested this with both cmount and mount2 on my mac, but I haven't tested this on Linux, BSD, or Windows.

cmd/cmount/fs.go
eatnumber1: Done!

cmd/mount2/node.go
```diff
@@ -404,3 +404,83 @@ func (n *Node) Rename(ctx context.Context, oldName string, newParent fusefs.Inod
...
+		errno = syscall.E2BIG
+		return
+	}
+
+	if datalen > len(dest) {
+		errno = syscall.ERANGE
+		return
+	}
+
+	copy(dest, data)
```
eatnumber1, line 431: If we change the API of n.node.Getxattr to take a non-reallocateable slice, it moves this complexity into the vfs layer, but is more performant (we can skip this copy). Do you have a preference on simplicity vs. performance in the vfs layer?

vfs/file.go
```diff
@@ -784,3 +787,46 @@ func (f *File) Truncate(size int64) (err error) {
...
 	return nil
 }
+
+// Note: If we end up adding many different xattrs, rewrite this function into
+// something more extensible.
+func (f *File) Getxattr(name string) (value []byte, err error) {
+	obj := f.getObject()
+
+	if name == "user.mime_type" {
+		return []byte(fs.MimeType(context.TODO(), obj)), nil
```
eatnumber1, line 797: In the mount2 backend it does give us Context objects that could be passed up. I see however that none of these other methods accept a Context, so I didn't add it to this one.

Do you think it's worth adding a Context argument for Getattr/Listattr?

eatnumber1:
Ah, based on the CI failures I'm reminded that there's disagreement between platforms on whether ENOATTR exists. I'll fix that shortly.

eatnumber1:
I've:

1. Fixed the ENOATTR issue
2. Plumbed the context through from the mount layer.
3. Tested on Linux
4. Implemented the mount backend (and tested it on Linux)

As mentioned all three mount backends work correctly on Linux, although it looks like `getfattr` hides the system namespace by default, so you need to pass `-m '.*'` to get the hash to appear.

Here's my tests, with identical output for all three mount backends:

```
$ getfattr -d -m '.*' image.jpg
getfattr: Removing leading '/' from absolute path names
# file: image.jpg
system.org.rclone.hash.md5="920e3d7899c8eed81b50d6f26fe0f64b"
user.mime_type="image/jpeg"
$ getfattr -n user.noexist image.jpg
image.jpg: user.noexist: No such attribute
```

vfs/file.go
eatnumber1: Opted to pass the context object through.

eatnumber1:
Fixed a couple of style issues, which should fix the CI for Linux.

For FreeBSD, I had to make a few fixes to the way I was handling ENODATA. That's done now, but I don't have a FreeBSD machine to test, so may need a few rounds with the CI to get it right.

eatnumber1:
I just had a thought: it would be nice if file hashes get preserved on files even if you copy the file out of rclone. Putting the hash in the system namespace won't allow that to happen (on Linux).

I'm gonna change it to `user.checksum.<hashtype>` so the hashes can be preserved.

eatnumber1:
Fixed missing syscall import on Linux and FreeBSD, and manually tested on FreeBSD (which I hadn't done before).

```
$ lsextattr user image.jpg
image.jpg	checksum.md5	mime_type
$ lsextattr user image.jpg
image.jpg	checksum.md5	mime_type
$ getextattr user checksum.md5 image.jpg
image.jpg	920e3d7899c8eed81b50d6f26fe0f64b
$ getextattr user mime_type image.jpg
image.jpg	image/jpeg
```

eatnumber1:
Is the test failure on Linux a flaky test? It seems to pass on my Linux machine:

```
go test -count=10 -tags=cmount,mount2,mount github.com/rclone/rclone/cmd/mount2
ok      github.com/rclone/rclone/cmd/mount2     58.844s
```

ncw:
> Is the test failure on Linux a flaky test? It seems to pass on my Linux machine:
> 
> ```
> go test -count=10 -tags=cmount,mount2,mount github.com/rclone/rclone/cmd/mount2
> ok      github.com/rclone/rclone/cmd/mount2     58.844s
> ```

Hmm, I think it is probably a flakey test.

I'll kick the CI to have another go.

ncw:
Can you squash the noise commits (reformatting, fix this etc) into the relevant commit so I have a nice clean set of commits and then I'll review them one more time :-)

Thanks

eatnumber1:
Cleaned up most of them except for c61af39b454ab84478bb297ef0e85cc4753d3fdc, which will require some merge deconflicting to eliminate, so I left it. Let me know if you'd prefer I remove it.

I opted to leave the commit history of the evolution from `system.hash` through `system.org.rclone.hash` to `user.checksum` just in case it's useful for others to see why we decided on `user.checksum` .

eatnumber1:
According to the CI, this is the errors:

```
cmd/cmount/errors.go:1: File is not `goimports`-ed (goimports)
// +build cmount
cmd/mount/file.go:135: File is not `goimports`-ed (goimports)
		return true  // keep going
```

I've fixed them both.

Running golangci-lint manually, I get similar errors from across rclone though. It looks like the `// +build` syntax [has changed](https://pkg.go.dev/go/build#hdr-Build_Constraints) to `//go:build`, which is generating these errors. I suppose the CI isn't complaining about files I didn't change?

ivandeex:
you can either
- install go 1.17 and run `gofmt -w ./file1.go ...; git commit ...`
- or analyze the latest build finished after this very comment

eatnumber1:
```
Error: cmd/all/all.go:15:2: //go:build comment without // +build comment
```

Fixed.

I'll see about installing go 1.17 (I'm on 1.13) so I don't keep needing you to rerun the CI. Sorry for the trouble.

eatnumber1:
I installed go 1.17 and indeed, gofmt now reports warnings for most of the files in rclone (all those with `// +build` comments).

```
$ gofmt -l . | wc -l
177
```

I've now updated the build comment for all of the files I touched.

```
gofmt -w $(git log fb1b45086593fecd6e765dbbec8802580649f4e6..HEAD | grep commit | cut -d\  -f2 | xargs -n1 git diff-tree --name-only --no-commit-id -r | sort -u | grep '.*\.go$')
```

eatnumber1:
Friendly ping here, looks like because this has been sitting for a while it needs some rebasing to merge successfully now. I'm happy to do that if you'd prefer, but would like to get this submitted.

eatnumber1:
Is there anything left for me to do here, or are we just waiting for @ncw to review?

eatnumber1:
@ncw friendly ping, are you still willing to pull in this change?

ivandeex:
We are waiting for me. I'm busy elsewhere now. Will review outstanding PRs assigned on me in a month.
Sorry for delay.

eatnumber1:
Hey there, if I do the necessary rebasing, is this still a PR that can get merged?

ncw:
Firstly, many apologies for the long delay on this. So many pull requests - so little time!

There have been some developments in the rclone code base since you did this, namely [the metadata framework](https://rclone.org/docs/#metadata) and the [local backend's support of metadata](https://rclone.org/local/#metadata) so what I'd suggest would be the best way forward would be for this PR to expose the metadata on objects in the same way the local backend does, so on a `user.*` prefix.

There is more work to be done on the metadata framework (its only supported by s3 and local at the moment) but doing it like this gives proper support for all backends.

Backends don't currently expose hashes as metadata - maybe they should.

What do you think?

ncw:
@eatnumber1 do you want to continue with this PR? I'm sorry we didn't get it merged.

I like the basic idea but it needs to be combined with the metadata framework I think.

We read xattrs from the local backend into metadata. It would be nice if we could serve that metadata via xattrs on the 3 rclone mount options `mount`, `mount2`, `cmount`.

The hashes could then be part of that metadata.

eatnumber1:
I don't think so. I've since moved off of cloud storage altogether, largely obsoleting my need for rclone. I'm still interested in the idea of checksums-as-xattrs in general, but probably don't have the time for a change that I won't be using (nor, perhaps, should you want a feature with zero users).

Issue #5435, opened 2021-07-02
Google Drive API Client ID Authentication Error


jcguu95:
#### What is the problem you are having with rclone?

(This is a report of a potential issue of the official guide on how to setup google drive client ID.)

Setting up with Google Drive using client_ID. Setup one following the [official guide](https://rclone.org/drive/#making-your-own-client-id) without issue. Got the client ID and the secret. 

While configuring `rclone` using the ID and the secret, the browser asked me to logged into my google drive account. Clicked on it, and it showed the following error

```
Authorization Error
Error 403: access_denied
The developer hasn’t given you access to this app. It’s currently being tested and it hasn’t been verified by Google. If you think you should have access, contact the developer (<HIDDEN>@gmail.com).
Learn more
Request Details
access_type=offline
response_type=code
redirect_uri=http://127.0.0.1:53682/
state=GVfOTKroPpO9dof4P8nDFg
client_id=<HIDDEN>.apps.googleusercontent.com
scope=https://www.googleapis.com/auth/drive
```

#### What is your rclone version (output from `rclone version`)

```
- os/type: linux
- os/arch: amd64
- go/version: go1.16.4
- go/linking: dynamic
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Archlinux, 64 bit.

#### Which cloud storage system are you using? (e.g. Google Drive)

Google Drive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

`rclone config`

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

jcguu95:
Resolved by following this [thread](https://forum.rclone.org/t/after-completing-the-setup-wizard-and-launching-the-rclone-on-windows-10-the-message-error-403-access-denied-appears/24814/9). The documentation needs to be updated indeed.

ncw:
@animosity22 you were going to update the docs for that? I don't see a PR?

jcguu95:
It's fine. I have PR'd [here](https://github.com/rclone/rclone/pull/5442).

Ariadne70:
[...]

PR #5442, opened 2021-07-05
Add an extra step in doc:google-drive to proceed correctly.

jcguu95:
* This is to fix this issue: https://github.com/rclone/rclone/issues/5435#issuecomment-873985284

#### Checklist

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -58,7 +58,7 @@ Choose a number from below, or type in your own value
    \ "drive.metadata.readonly"
 scope> 1
 ID of the root folder - leave blank normally.  Fill in to access "Computers" folders. (see docs).
-root_folder_id> 
+root_folder_id>
 Service Account Credentials JSON file path - needed only if you want use SA instead of interactive login.
 service_account_file>
 Remote config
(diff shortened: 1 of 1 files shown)


animosity22:
There are two paths for getting your own client ID.

Breaking down the two paths:

Non GSuite User:
- User type must be external as you cannot pick internal
- If you are using testing, you must add the users to the test group:
image
- If you hit publish and push to production, most likely that will work as well as there aren't any other changes that most users will make, it just works so it feels like this  would work generally

If you are a GSuite user:
- You can pick external following the same flow above
- If you pick internal (only users on the same gsuite domain can use it), you don't need to publish (that's what I use)

@jcguu95 - let me know if you want to touch up that to cover that as that's how it seems like the process works now. Let me know if you have a different flow.

jcguu95:
Ah. Thanks for providing those information.

I'm no GSuite user, and dunno if I can say that's right. I'm just adding the extra step that worked for me (and another dude on the forum). Perhaps someone else should make the PR?

ncw:
If between you, you can come up with some wording, I can mash it into the PR :-)

ivandeex:
@animosity22 @jcguu95 
Do you think you can come to agreement by **this** christmas?

ivandeex:
Closed due to total inactivity

PR #5443, opened 2021-07-05
jottacloud: add no versions option

buengese:
This is another one of those annoying workaround. Currently, Jottacloud doesn't provide a way to delete old versions of files via the API and no such option exists in the web ui either. This means the only way to remove old versions from the file is to completely delete. Since old versions obviously count against your quota, they can get rather annoying. This pr adds an option to avoid any server side versioning by deleting old files before uploading a new version.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -100,6 +100,11 @@ func init() {
 			Default:  fs.SizeSuffix(10 * 1024 * 1024),
 			Advanced: true,
 		}, {
+			Name:     "no_versions",
+			Help:     "Avoid server side versioning by deleting files and recreating files instead of overwriting them.",
+			Default:  false,
+			Advanced: true,
+		}, {
 			Name:     config.ConfigEncoding,
 			Help:     config.ConfigEncodingHelp,
 			Advanced: true,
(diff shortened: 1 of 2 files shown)


albertony:
Nice.
Is there a risk worthy a doc warning or something, that the file may be "lost" from the destination if something bad happens? I mean after deleting existing file but before copying the new file, e.g. if it bails out on "failed to calculate MD5"? I don't know, just a thought..

buengese:
I think that's pretty minor risk, since it's not like rclone is going to ever delete the source files unless the transfer succeeded. Though better safe than sorry. I'll add a short warning in the docs.

ncw:
> I think that's pretty minor risk, since it's not like rclone is going to ever delete the source files unless the transfer succeeded. Though better safe than sorry. I'll add a short warning in the docs.

I think that is probably a good idea.

If you were feeling very paranoid you could rename the old file to .TEMP and delete it or rename it back if successful/unsuccesfull.

I don't think that is necessary though provided we put the warning in the docs.

albertony:
> If you were feeling very paranoid you could rename the old file to .TEMP and delete it or rename it back if successful/unsuccesfull.

Another positive side of this approach is that it would take advantage of deduplication. With the existing approach, permanent delete followed by upload, it is more likely resulting in entire content having to be uploaded.

What I have found is that a simple copy, i.e. server-side copy, will not include the revisions. One could therefore instead upload to .TEMP, then permanent delete target, then rename .TEMP to target. Result will be the same - no revisions on resulting target. But I guess no big advantages from former approach either: 1 rename + 1 delete + 1 upload (when everything succeeds).

> I don't think that is necessary.

No, probably not. The downside of these two approaches is complexity (e.g. with upload/rename to .TEMP - what if .TEMP already exists?), and also if you are near your quota then replacing a huge file could (depending on deduplication) count twice for the short time before original is deleted.

So... Sorry for the "bump". Primarly just wanted to "record" the finding of copy not including revisions. 😊

PR #5445, opened 2021-07-05
Terminate with non-zero status after receiving signal

### LINKED ISSUE #5437: rclone exits with status 0 after receiving SIGINT/SIGTERM

When rclone receives a SIGINT (Ctrl+C) or SIGTERM signal while an atexit function is registered it always exits with status code 0. Unix convention is to exit with a non-zero status code. Often it's `128 + int(signum)`, but at least not zero. Wikipedia writes ([source](https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts)):

> When a command is terminated by a signal whose number is N, a shell sets the variable $? to a value greater than 128. Most shells use 128+N, while ksh93 uses 256+N.

The non-zero exit status enables scripts to handle commands aborted by signals, e.g. by the user. The responsible code is in `atexit.go:Run` where the exit status is hardcoded: https://github.com/rclone/rclone/blob/e635f4c0be24cdf2272b2c6018939e8469e4c6d5/lib/atexit/atexit.go#L54

I'll be happy to submit a pull request. Would the rclone developers prefer to always use status 2 ("error not otherwise categorised") or the 128+signum convention?

```
$ rclone --version
rclone v1.55.1
- os/type: linux
- os/arch: amd64
- go/version: go1.16.3
- go/linking: static
- go/tags: none
```

#### How to use GitHub

hansmi:
With this change fatal signals handled by the `atexit` package causea non-zero exit code. On Unix systems it's `128 + int(signum)` while
on other systems, such as Windows, it's always 2 ("error not otherwise
categorised").

Resolves #5437.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -37,6 +37,7 @@ import (
 	"github.com/rclone/rclone/fs/rc/rcserver"
 	"github.com/rclone/rclone/lib/atexit"
 	"github.com/rclone/rclone/lib/buildinfo"
+	"github.com/rclone/rclone/lib/exitcode"
 	"github.com/rclone/rclone/lib/random"
 	"github.com/rclone/rclone/lib/terminal"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


ncw (approved):
This looks very nice and thorough code.

I'll merge it now - thank you :-)

ncw:
...actually better run the CI first - If that passes then I'll merge it :-)

ncw:
CI passed (apart from one heisenbug). I'll merge this now - thank you :-)

PR #5449, opened 2021-07-08
operations: fix goroutine leak in case of copy retry

ankur0493:

Whenever transfer.Account() is called, a new goroutine acc.averageLoop()
is started. This goroutine exits only when the channel acc.exit is closed.
acc.exit is closed when acc.Done() is called, which happens during tr.Done().

However, if tr.Reset is called during a copy low level retry, it replaces
the tr.acc, without calling acc.Done(), which results in the goroutine
mentioned above never exiting.

This commit calls acc.Done() during a tr.Reset()

Also cleaned the authors list to remove multiple emails for this author.

No

#### Checklist

### DIFF
--- a/fs/accounting/transfer.go
+++ b/fs/accounting/transfer.go
@@ -132,6 +132,7 @@ func (tr *Transfer) Reset(ctx context.Context) {
 	ci := fs.GetConfig(ctx)
 
 	if acc != nil {
+		acc.Done()
 		if err := acc.Close(); err != nil {
 			fs.LogLevelPrintf(ci.StatsLogLevel, nil, "can't close account: %+v\n", err)
 		}


docs/content/authors.md
```diff
@@ -314,7 +314,7 @@ put them back in again.` >}}
   * Barry Muldrey <barry@muldrey.net>
   * Sebastian Brandt <sebastian.brandt@friday.de>
   * Marco Molteni <marco.molteni@mailbox.org>
-  * Ankur Gupta <ankur0493@gmail.com> <7876747+ankur0493@users.noreply.github.com>
+  * Ankur Gupta <7876747+ankur0493@users.noreply.github.com>
```
ncw, line 317: Note this at the top of authors.md

```
{{< rem `email addresses removed from here need to be addeed to
bin/.ignore-emails to make sure update-authors.py doesn't immediately
put them back in again.` >}}
```

So you need to put the emails you want ignored into `bin/.ignore-emails` as well.

Can you do the change to authors.md and .ignore-emails in a separate commit please? I'm happy to have them in this PR or a new one whichever is easiest for you.

ncw (commented):
I've seen this leak in other peoples traces but I've never managed to track it down - well done :-)

I think it indicates that the interface isn't very well defined... Why doesn't account.Close just stop the stats goroutine? I don't know!

I think what I'd like to do is merge this fix then think about revamping the interface.

docs/content/authors.md
ankur0493: I removed the change from this PR - will create a separate one for the authors.md change.

ncw (approved):

Apologies for not merging it sooner!

PR #5456, opened 2021-07-12
Fix single request upload for files smaller than `--drive-upload-cutoff`

YenForYang:
I discovered that `rclone` always upload in chunks of 16MiB whenever uploading a file smaller than `--drive-upload-cutoff`. This is undesirable since the purpose of the flag `--drive-upload-cutoff` is to *prevent* chunking below a certain file size.

I realized that it wasn't `rclone` forcing the 16MiB chunks. The `google-api-go-client` forces a chunk size default of [`googleapi.DefaultUploadChunkSize`](https://github.com/googleapis/google-api-go-client/blob/32bf29c2e17105d5f285adac4531846c57847f11/googleapi/googleapi.go#L55-L57) bytes for resumable type uploads. This means that all requests that use `*drive.Service` directly for upload without specifying a `googleapi.ChunkSize` will be forced to use a *`resumable`* `uploadType` (rather than `multipart`) for files less than `googleapi.DefaultUploadChunkSize`. This is also noted directly in the Drive API client documentation [here](https://pkg.go.dev/google.golang.org/api/drive/v3@v0.44.0#FilesUpdateCall.Media).

We can fix this problem by passing `googleapi.ChunkSize(0)` to `Media()` method calls, which is the only way to disable chunking completely.  This is mentioned in the API docs [here](https://pkg.go.dev/google.golang.org/api/googleapi@v0.44.0#ChunkSize). The other alternative would be to pass `googleapi.ChunkSize(f.opt.ChunkSize)` -- however, I'm *strongly* in favor of *not* doing this for performance reasons. By not explicitly passing a `googleapi.ChunkSize(0)`, we effectively allow [`PrepareUpload()`](https://pkg.go.dev/google.golang.org/api/internal/gensupport@v0.44.0#PrepareUpload) to create a [`NewMediaBuffer`](https://pkg.go.dev/google.golang.org/api/internal/gensupport@v0.44.0#NewMediaBuffer) that copies the original `io.Reader` passed to `Media()` in order to check that its size is less than `ChunkSize`, which will unnecessarily consume time and memory.  

While on the topic of chunk sizes, I suggest changing `const defaultChunkSize = 8 * fs.Mebi` to `const defaultChunkSize = googleapi.DefaultUploadChunkSize` to keep things in the `drive` backend consistent and more maintainable with the API client it uses. However, this may require other changes so I'll leave the matter for future discussion(s).

`minChunkSize` should definitely be changed to be `googleapi.MinUploadChunkSize`, as it is something specified we have no control over.

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -67,7 +67,7 @@ const (
 	defaultScope                = "drive"
 	// chunkSize is the size of the chunks created during a resumable upload and should be a power of two.
 	// 1<<18 is the minimum size supported by the Google uploader, and there is no maximum.
-	minChunkSize     = 256 * fs.Kibi
+	minChunkSize     = googleapi.MinUploadChunkSize
 	defaultChunkSize = 8 * fs.Mebi
 	partialFields    = "id,name,size,md5Checksum,trashed,explicitlyTrashed,modifiedTime,createdTime,mimeType,parents,webViewLink,shortcutDetails,exportLinks"
 	listRGrouping    = 50   // number of IDs to search at once when using ListR
@@ -2116,7 +2116,7 @@ func (f *Fs) PutUnchecked(ctx context.Context, in io.Reader, src fs.ObjectInfo, 
 		// Don't retry, return a retry error instead
 		err = f.pacer.CallNoRetry(func() (bool, error) {
 			info, err = f.svc.Files.Create(createInfo).
-				Media(in, googleapi.ContentType(srcMimeType)).
+				Media(in, googleapi.ContentType(srcMimeType), googleapi.ChunkSize(0)).
 				Fields(partialFields).
 				SupportsAllDrives(true).
 				KeepRevisionForever(f.opt.KeepRevisionForever).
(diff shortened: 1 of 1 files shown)


YenForYang:
Hopefully someone can squash that extra commit for me. Also, I'm not sure what could be causing the remaining 2 checks to fail.

ncw:
> I discovered that `rclone` always upload in chunks of 16MiB whenever uploading a file smaller than `--drive-upload-cutoff`. This is undesirable since the purpose of the flag `--drive-upload-cutoff` is to _prevent_ chunking below a certain file size.

That is a great fix - thank you

The remaining failure is this - just `go fmt` the file and it should fix it.

> backend/drive/drive.go:70: File is not `goimports`-ed (goimports)

I'll squash the commits when we are done with the PR.

I'll try to merge this for 1.56 if possible.

> While on the topic of chunk sizes, I suggest changing `const defaultChunkSize = 8 * fs.Mebi` to `const defaultChunkSize = googleapi.DefaultUploadChunkSize` to keep things in the `drive` backend consistent and more maintainable with the API client it uses. However, this may require other changes so I'll leave the matter for future discussion(s).

This will use more memory I believe since rclone buffers the chunks in memory in case it has to retry them so is something which deserves a bit of discussion.

I think it would probably be OK

PR #5478, opened 2021-07-22
cryptcheck: agnostify src dst order

### LINKED ISSUE #5425: make cryptcheck order agnostic
At present, when one runs cryptcheck the command must be structured as follows:
`rclone cryptcheck [unencrypted local] [crypted remote]`

It would be useful if rclone was able to run this command regardless of the order of remotes. A rationale would be that after I run `rclone copy crypt_a: b:` I can simply scroll back and change `copy` to `cryptcheck` and the command will execute properly. Presently, the command must be re-written with the order reversed in order to run this check.

Levi-Hope:

Make `rclone cryptcheck` being agnostic about src dst order.

So that both
`rclone cryptcheck [unencrypted local] [crypted remote]`
and
`rclone cryptcheck [crypted remote] [unencrypted local]`
work.

close #5425

#### Checklist

### DIFF
--- a/cmd/cryptcheck/cryptcheck.go
+++ b/cmd/cryptcheck/cryptcheck.go
@@ -57,11 +57,16 @@ After it has run it will log the status of the encryptedremote:.
 
 // cryptCheck checks the integrity of a crypted remote
 func cryptCheck(ctx context.Context, fdst, fsrc fs.Fs) error {
-	// Check to see fcrypt is a crypt
+	// Check to see fdst is crypted. If not, check fsrc.
 	fcrypt, ok := fdst.(*crypt.Fs)
 	if !ok {
-		return errors.Errorf("%s:%s is not a crypt remote", fdst.Name(), fdst.Root())
+		fcrypt, ok = fsrc.(*crypt.Fs)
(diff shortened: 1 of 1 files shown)


Levi-Hope:
I just found that implicitly switching `src` and `dst` may surprise users when they used flags like `--missing-on-dst`,  `--missing-on-src`, `--combined` and `--one-way`.
Should we increase the log level to Logf? Or handle the flags in some way.

ncw:
> I just found that implicitly switching `src` and `dst` may surprise users when they used flags like `--missing-on-dst`, `--missing-on-src`, `--combined` and `--one-way`. Should we increase the log level to `Logf` so users can be warned? Or should we handle the flags in some way?

I think switching `src` and `dst` is probably a bad idea and will lead to user confusion!

How about a bit more generalisation so either the source or the dest or both can be crypt remotes?

ivandeex:
@Levi-Hope 
I'm afraid this PR has been refused in its current form.

If you are still interested in the idea, we can accept a PR if it
**allows the cryptcheck source / dest / both be crypt remotes**

Anyway, thanks for your effort

PR #5485, opened 2021-07-26
copyurl: Update copyurl to honor HTTP filename directive

jptreen:

Adding `--auto-filename-header`

To set preferred download filenames for HTTP requests, RFC 6226
specifies a 'filename' directive, available within 'Content-Disposition'
header, and suggests both token and quoted string syntaxes should be allowable
for the filename directive. i.e.: either `filename=name.txt` or
`filename="name.txt"` should be acceptable.

See below for details:
https://httpwg.org/specs/rfc6266.html#disposition.parameter.filename
https://httpwg.org/specs/rfc6266.html#advice.generating

`--auto-filename-header` is designed to work in conjunction with `--auto-filename`, honoring the filename directive when retrieving files from a server.

Directives contained within the `Content-Disposition` header are semicolon delimited, and can vary in length and order. Consideration had to be given to parsing entries like `inline; filename="file.txt";` as well as the more straightforward examples above.

I raised it on the forum, got no response, and thought it was an clear enough issue to tackle myself without altering existing functionality. Maybe that was good enough. If not, maybe a rogue PR is enough to get a response. 😉 

https://forum.rclone.org/t/copyurl-auto-filename-could-be-cleverer-and-respect-the-http-content-disposition-optional-directive-filename/25553

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"os"
 	"path"
@@ -1743,7 +1744,7 @@ func RcatSize(ctx context.Context, fdst fs.Fs, dstFileName string, in io.ReadClo
 type copyURLFunc func(ctx context.Context, dstFileName string, in io.ReadCloser, size int64, modTime time.Time) (err error)
 
 // copyURLFn copies the data from the url to the function supplied
-func copyURLFn(ctx context.Context, dstFileName string, url string, dstFileNameFromURL bool, fn copyURLFunc) (err error) {
+func copyURLFn(ctx context.Context, dstFileName string, url string, autoFilename, dstFileNameFromHeader bool, fn copyURLFunc) (err error) {
 	client := fshttp.NewClient(ctx)
 	resp, err := client.Get(url)
 	if err != nil {
@@ -1757,7 +1758,17 @@ func copyURLFn(ctx context.Context, dstFileName string, url string, dstFileNameF
 	if err != nil {
 		modTime = time.Now()
 	}
-	if dstFileNameFromURL {
+	if autoFilename {
+		if dstFileNameFromHeader {
+			_, params, err := mime.ParseMediaType(resp.Header.Get("Content-Disposition"))
+			headerFilename := path.Base(strings.Replace(params["filename"], "\\", "/", -1))
+			if err != nil || headerFilename == "" {
+				return fmt.Errorf("copyurl failed: filename not found in the Content-Dispoition header")
+			}
+			fs.Debugf(headerFilename, "filename found in Content-Disposition header.")
+			return fn(ctx, headerFilename, resp.Body, resp.ContentLength, modTime)
+		}
+
 		dstFileName = path.Base(resp.Request.URL.Path)
 		if dstFileName == "." || dstFileName == "/" {
 			return fmt.Errorf("CopyURL failed: file name wasn't found in url")
@@ -1768,9 +1779,9 @@ func copyURLFn(ctx context.Context, dstFileName string, url string, dstFileNameF
 }
 
 // CopyURL copies the data from the url to (fdst, dstFileName)
-func CopyURL(ctx context.Context, fdst fs.Fs, dstFileName string, url string, dstFileNameFromURL bool, noClobber bool) (dst fs.Object, err error) {
+func CopyURL(ctx context.Context, fdst fs.Fs, dstFileName string, url string, autoFilename, dstFileNameFromHeader bool, noClobber bool) (dst fs.Object, err error) {
 
-	err = copyURLFn(ctx, dstFileName, url, dstFileNameFromURL, func(ctx context.Context, dstFileName string, in io.ReadCloser, size int64, modTime time.Time) (err error) {
+	err = copyURLFn(ctx, dstFileName, url, autoFilename, dstFileNameFromHeader, func(ctx context.Context, dstFileName string, in io.ReadCloser, size int64, modTime time.Time) (err error) {
 		if noClobber {
 			_, err = fdst.NewObject(ctx, dstFileName)
 			if err == nil {
@@ -1785,7 +1796,7 @@ func CopyURL(ctx context.Context, fdst fs.Fs, dstFileName string, url string, ds
 
 // CopyURLToWriter copies the data from the url to the io.Writer supplied
 func CopyURLToWriter(ctx context.Context, url string, out io.Writer) (err error) {
-	return copyURLFn(ctx, "", url, false, func(ctx context.Context, dstFileName string, in io.ReadCloser, size int64, modTime time.Time) (err error) {
+	return copyURLFn(ctx, "", url, false, false, func(ctx context.Context, dstFileName string, in io.ReadCloser, size int64, modTime time.Time) (err error) {
 		_, err = io.Copy(out, in)
 		return err
 	})
--- a/fs/operations/operations_test.go
+++ b/fs/operations/operations_test.go
@@ -729,9 +729,14 @@ func TestCopyURL(t *testing.T) {
 
 	// check when reading from regular HTTP server
 	status := 0
+	nameHeader := false
+	headerFilename := "headerfilename.txt"
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if status != 0 {
 			http.Error(w, "an error ocurred", status)
+		}
+		if nameHeader {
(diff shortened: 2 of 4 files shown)


jptreen:
Fixed linux workflow failure and squashed the commits.

jptreen:
Fixed some linting issues.

ncw:
> `--auto-filename-header` is designed to work in conjunction with `--auto-filename`, honoring the filename directive when retrieving files from a server.

Is a new flag necessary? Maybe it should always be parsing `Content-Disposition` if `--auto-filename` is set? That is what browsers would do, after all.

What do you think?

> I raised it on the forum, got no response, and thought it was an clear enough issue to tackle myself without altering existing functionality. Maybe that was good enough. If not, maybe a rogue PR is enough to get a response. 

:-) It looks like a sensible idea to me.

I'd be interested to hear what actual use case you have for it.

I'll review the code now.

fs/operations/operations.go
```diff
@@ -1614,11 +1614,27 @@ func RcatSize(ctx context.Context, fdst fs.Fs, dstFileName string, in io.ReadClo
 	return obj, nil
 }
 
+// CopyURLParseHeader returns a valid filename from a Content-Disposition HTTP string
```
ncw, line 1617: In Go a capitalised first letter makes a function externally accessible to the package.

I don't think this one needs to be.

So I'd probably rename it `parseContentDisposition` which seems more accurate.

fs/operations/operations.go
```diff
@@ -1614,11 +1614,27 @@ func RcatSize(ctx context.Context, fdst fs.Fs, dstFileName string, in io.ReadClo
...
+func CopyURLParseHeader(contentDisposition string) (dstFileName string) {
+	headerDirectives := strings.Split(contentDisposition, ";")
```
ncw, line 1619: I think it would be better to leverage Go's MIME parsing code here.

You can use [mime.ParseMediaType](https://pkg.go.dev/mime#ParseMediaType) to parse the `Content-Disposition` header. There is a very nice [stack overflow answer](https://stackoverflow.com/a/28845255/164234) about this.

fs/operations/operations.go
```diff
@@ -1633,6 +1649,14 @@ func copyURLFn(ctx context.Context, dstFileName string, url string, dstFileNameF
 		modTime = time.Now()
 	}
 	if dstFileNameFromURL {
+		if dstFileNameFromContentDisposition {
+			headerFileName := CopyURLParseHeader(resp.Header.Get("Content-Disposition"))
+			if len(headerFileName) > 0 {
```
ncw, line 1654: This is usually written `if headerFileName != ""`

fs/operations/rc.go
```diff
@@ -224,9 +224,10 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 			return nil, err
 		}
 		autoFilename, _ := in.GetBool("autoFilename")
+		autoFilenameHeader, _ := in.GetBool("autoFilenameHeader")
```
ncw, line 227: If we decide to keep this flag, then it needs documenting above in the `copyurl` docs.

ncw (commented):
See inline for code comments :-)

If we decide to drop the flag it will simplify the code considerably!

jptreen:
Thanks for the review:

> Is a new flag necessary? Maybe it should always be parsing Content-Disposition if --auto-filename is set? That is what browsers would do, after all.

Through an abundance of caution, I didn't want to tread on any existing behavior, even if that behavior was wrong. I'm happy, however, to aim for browser functionality.

> I'd be interested to hear what actual use case you have for it.

I have to work with a list of temporary web available files, which look like they've been secured with a salted hash, and the filename just isn't in the path, the hash is. If I don't want to download the files as something like `0894635ac5a1128086581d149e963c2dd9a4ea90`, but instead `name_i_would_like.jpg`, I need to be able to parse the headers. This job was done by somebody in a browser.

> In Go a capitalised first letter makes a function externally accessible to the package.

Many thanks for the review. Apologies for my appauling Go code, its *not exactly* a language I'm best familiar with. The reason I exported this function is because I wanted to test it on its own, and I was getting complaints when trying to run tests directly calling a non-exported function.

I'll take a look at your notes, remove the new flag, check out MIME parsing code, and rework this PR when I get the chance.

ncw:
> Thanks for the review:
> 
> > Is a new flag necessary? Maybe it should always be parsing Content-Disposition if --auto-filename is set? That is what browsers would do, after all.
> 
> Through an abundance of caution, I didn't want to tread on any existing behavior, even if that behavior was wrong. I'm happy, however, to aim to emulate a generic browser functionality.

I checked the docs for `curl` and `wget` and they both require a flag (`--remote-header-name` and `--content-disposition` respectively).

Reading the curl docs

>        -J, --remote-header-name
>               (HTTP) This option tells the -O, --remote-name option to use the
>               server-specified  Content-Disposition  filename  instead  of ex‐
>               tracting a filename from the URL.
>               [snip]
>               WARNING:  Exercise  judicious  use of this option, especially on
>               Windows. A rogue server could send you the  name  of  a  DLL  or
>               other  file  that could possibly be loaded automatically by Win‐
>               dows or some third party software.

Made me think that we should probably excercise a little care with the filename

The [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) say (emphasis mine)

> filename
> 
> Is followed by a string containing the original name of the file transmitted. The filename is always optional and **must not be used blindly by the application: path information should be stripped**, and conversion to the server file system rules should be done. This parameter provides mostly indicative information. When used in combination with Content-Disposition: attachment, it is used as the default filename for an eventual "Save As" dialog presented to the user.

So I think your flag is a good idea.

I think we should also sanitise the filename...

I'd do this by converting `\` to `/` (with `strings.Replace`) then running path.Base() on the result to discard the path information as suggested above. This string could still contain characters which can't be represented on the destination filesystem, however the backend should then convert it into something which is safe.

> > I'd be interested to hear what actual use case you have for it.
> 
> I have to work with a list of temporary web available files, which look like they've been secured with a salted hash, and the filename just isn't in the path, the hash is the final segment. If I don't want to download the files as something like `0894635ac5a1128086581d149e963c2dd9a4ea90`, but instead `name_i_would_like.jpg`, I need to be able to parse the headers. This job was previously done by a person with a browser, so it just never came up as an issue.

Thanks for the explanation.

> > In Go a capitalised first letter makes a function externally accessible to the package.
> 
> Many thanks for the review. Apologies for my appauling Go code, its _not exactly_ a language I'm best familiar with. 

That is fine - very happy to help newcomers to Go with the syntax!


> The reason I exported this function is because I wanted to test it on its own, and I was getting complaints when trying to run tests directly calling a non-exported function.

What you want to do is stick the test in `fs/operations/operations_internal_test.go` which is the internal tests - the `operations_test.go` are the external (black box) tests which don't have access to the internal symbols.  (The `package` header is different between those two files, one is `operations` and the other is `operations_test` which is why they are different).
 
> I'll take a look at your notes, remove the new flag, check out MIME parsing code, and rework this PR when I get the chance.

Great! Give me a ping when you are read for another review.

ivandeex:
@jptreen 
> we should probably exercise a little care with the filename
> I think we should also sanitize the filename...
> I'd do this by converting `\` to `/` then running path.Base() on the result to discard the path information as suggested above
> you want to stick the test in `fs/operations/operations_internal_test.go`
> ping when you are read for another review.

ivandeex:
Unfortunately rclone is short on resources to finish all PRs unfinished by authors.

buengese:
So cleaned up and switched to [mime.ParseMediaType](https://pkg.go.dev/mime#ParseMediaType). I've kept a separate flag for using the content-disposition header that has to be set in addition to `--auto-filename` but renamed it `--header-filename`.

fs/operations/operations_test.go
```diff
@@ -776,7 +776,7 @@ func TestCopyURL(t *testing.T) {
 	tss := httptest.NewTLSServer(handler)
 	defer tss.Close()
 
-	o, err = operations.CopyURL(ctx, r.Fremote, "file2", tss.URL, false, false)
+	o, err = operations.CopyURL(ctx, r.Fremote, "file2", tss.URL, false, false, false)
```
ncw, line 779: This could do with a new test for the new flag somewhere round here!

ncw (commented):
Code looks good :-)

Needs a test - luckily there is a framework for them already so it should be easy!

buengese:
I've added some test now.

ncw (approved):
Super job :-)

And thank you for rescuing this and adding tests @buengese 

I'll merge this now.

PR #5303, opened 2021-05-04
Docs: Contributing eased for beginners in Go, Git and GitHub

olefrost:
#### What is the purpose of this change?

To make it easier to contribute without prior knowledge of Git, GitHub and Go.

The change includes improved/added steps to:
 * Install Git with basic setup
 * Use both SSH and HTTPS for the git origin
 * Install Go and verify the GOPATH
 * Update the forked master
 * Find a popular editor for Go

#### Was the change discussed in an issue or in the forum before?

The addition of an HTTPS link for the git origin was discussed and approved in this forum thread:
https://forum.rclone.org/t/contributing-guide-git-github-com-https-github-com/23855

The rest is based on my own experiences while preparing [my first contribution](https://github.com/rclone/rclone/pull/5273) on a Windows computer. 

I have done my best to research the changes and make the guide applicable for all platforms, protocols and preferences.

I am however still a beginner, so a careful review by somebody with native skills in Go, Git, GitHub and non-Windows platforms is needed. 

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [x] I have tested the added/modified links.
- [x] I have tested the modified parts of the installation flow on Windows.
- [x] I have tested the document layout in Preview mode on GitHub and Visual Studio Code.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [x] I'm done, this Pull Request is ready for review :-)

PR #5487, opened 2021-07-27
docs: prefer make over go build; mention cmount tag for macos

Cnly:
This PR instructs users to build with `make` instead of `go build`, which should give better version strings and help with debugging. Also the `cmount` build tag is mentioned for macOS users who want the `mount` function.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ Now [install Go](https://golang.org/doc/install) and verify your installation:
 
 Great, you can now compile and execute your own version of rclone:
 
-    go build                              
+    make
     ./rclone version
 
 Finally make a branch to add your new feature
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -63,6 +63,11 @@ Run `rclone config` to setup. See [rclone config docs](/docs/) for more details.
 
     brew install rclone
 
+NOTE: This version of rclone will not support `mount` any more (see
+[#5373](https://github.com/rclone/rclone/issues/5373)). If mounting is wanted
+on macOS, either install a precompiled binary or enable the relevant option
+when [installing from source](#install-from-source).
+
 ## macOS installation from precompiled binary, using curl ##
 
 To avoid problems with macOS gatekeeper enforcing the binary to be signed and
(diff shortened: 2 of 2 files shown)


albertony:
Is it not true that the original description only assumed you have git and go tools installed? While this change requires you to also have make, which on Windows, in particular, often may not be the case for an existing, somewhat minimal, dev env?

Cnly:
I've also considered that before. But at least CONTRIBUTING.md also have other references to `make` commands. Maybe we should mention both ways of building there, but recommend the use of `make`?

albertony:
Yes I saw that too - that the testing section mentions make. Mentioning both is a good idea. I was just concerned that newcomers on Windows would hit the wall on the make command with infamous error:

> 'make' is not recognized as an internal or external command,
> operable program or batch file.

Docs lists steps "Then install Git.." and "Now install Go and verify your installation", and then "Great, you can now compile and execute your own version of rclone" - but if only showing make command here then there is a "missing link" as the previous steps have not described how you would get it.

olefrost:
@albertony is absolutely right.

I tried to make the first section as simple and cross platform friendly as possible in #5303 (Contributing eased for beginners in Go, Git and GitHub). The level of detail and complexity then gradually increases down the document.

@Cnly How about adding your make tip in a small section about mount testing, just before the section on backend testing?

ncw:
I don't think we should take away the go standard way of building. This makes life much easier for Windows users and also for users that know the Go workflow.

I think that building with `make` is worth a mention. We know it works on all supported platforms in GitHub actions at least, but I don't know how much extra stuff you need to install on all the various OSes to make it work, wheras it is very well defined what you need to install to make `go install` etc work.

I note that quite a few of the bigger go projects don't use make ([restic](https://github.com/restic/restic) and [perkeep](https://github.com/perkeep/perkeep/blob/master/BUILDING) come to mine), but instead a go "script" which gets run as `go run make.go`. We could potentially do that. Getting rid of the make file wouldn't be too hard and it would make this sort of thing (as an example) much easier:

https://github.com/rclone/rclone/blob/bb6b44d199b9341cda0610eb5dc14203b009b36e/Makefile#L10-L12

But even if we did do that, I still want rclone to build as `go build`!

Cnly:
Thanks for all the comments! I've reverted some of the changes and now we still default to `go build`.

ncw (approved):
LGTM - thank you :-)

Squash and merge when ready!

Issue #5499, opened 2021-07-30
Documentation: two questions/issues re: Google Drive


gregsadetsky:
In the GDrive help document, i.e. [docs/content/drive.md](docs/content/drive.md):

1. At the `3. Configure rclone, assuming a new install` section, the line

```
y/n>             # Auto config, y
```

should probably read

```
y/n>             # Auto config, n
```

instead, as the question that rclone asks is "Edit advanced config? (y/n)". i.e. answering "y" goes into advanced config. The "auto config" is achieved by answering "n"

2. Right underneath,  under "4. Verify that it's working", it's said:

```
use rclone without specifying the --drive-impersonate option, like this: rclone -v foo@example.com lsf gdrive:backup
```

but that command (`rclone -v foo@example.com lsf gdrive:backup`) does not work. I am able to run the command (with the correct email address, obviously) with the `--drive-impersonate` option. Without the `--drive-impersonate` option, I get:

```
2021/07/30 13:30:59 Fatal error: unknown command "<valid email address>" for "rclone"
```

Thanks!

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
1. That needs an update for sure!
2. I don't see that here: https://github.com/rclone/rclone/blob/master/docs/content/drive.md#4-verify-that-its-working ?

Maybe I'm blind!

If you would like to send a fix for the docs, you can do this entirely in the web browser (click the pencil icon at the top right of the above page).

gregsadetsky:
1. Thanks, will update
2. It's the very last sentence of the "4. Verify that it's working" section. It's just above "Shared drives (team drives)".

Go to the [file](https://github.com/rclone/rclone/blob/master/docs/content/drive.md) and search for "use rclone without specifying the"

Actually, it's [right here](https://github.com/rclone/rclone/blob/master/docs/content/drive.md?plain=1#L277)

gregsadetsky:
Re: issue 2, I have a feeling that the `foo@example.com` email address should simply be removed from the command. i.e., rclone should simply be run like so: `rclone -v lsf gdrive:backup`

But I do want to double check first! :-)

ncw:
[...]

PR #5500, opened 2021-07-30
docs: fix instructions for Google Drive auto config

gregsadetsky:

Documentation fix

Yes, issue #5499


ncw:
Thank you :-)

PR #5502, opened 2021-07-31
-P, --progress by default, --stats-per-transfer

### LINKED ISSUE #5497: Add flag for print stats without Transfering info
[Original topic on forum](https://forum.rclone.org/t/progress-without-transfers-info-only-stats/25607)

The problem of **-P** flag is that it prints Transfering info. When i set -transfers 100, the terminal always scroll, down-up, down-up.

For my opinion, current behavior of -P flag, but without Transfering info make as default (so -P flag is no more needed).
In current version, user, that uses program first time want to know, what is happening, but need to read doc on how to enable progress info. Bad UX.

van9petryk:
Progress was turned on by default for better UX.
-P, --progress flag was removed
Was added --no-progress flag for turn off progress.
Transferring and Checking block is now disabled by default,
because when --transfers 100 or --checking 100
the progress block is not readable, as console window
constantly scrolls.
Was added --stats-per-transfer flag for enabling them
in case user need them.

Fixes #5497

#### Checklist


ncw:
> Progress was turned on by default for better UX.
> -P, --progress flag was removed
> Was added --no-progress flag for turn off progress.
> Transferring and Checking block is now disabled by default,

That is a backwards incompatible change which will undoubtedly break 1000s of scripts so I don't think we can do that.

In general we try not to make backwards incompatible changes. Rclone has many thousands of users and not breaking their scripts is important to us!

Note that if you always want the progress flag enabled then you can set an environment variable `RCLONE_PROGRESS=true`.

> because when --transfers 100 or --checking 100
> the progress block is not readable, as console window
> constantly scrolls.

This is bad. Maybe there is a way to solve this - if we read the height of the terminal?

> Was added --stats-per-transfer flag for enabling them
> in case user need them.

PR #5504, opened 2021-08-01
doc: fix instructions regarding gdrive

gregsadetsky:

fix documentation re: gdrive

yes, #5499


albertony:
Good catch, thank you. I will merge this now.

PR #5510, opened 2021-08-04
fix URL for test_proxy.py

joschl:
Just fixed the URL for the test_proxy.py script in the docs.

### DIFF
--- a/docs/content/commands/rclone_serve_sftp.md
+++ b/docs/content/commands/rclone_serve_sftp.md
@@ -324,7 +324,7 @@ together, if `--auth-proxy` is set the authorized keys option will be
 ignored.
 
 There is an example program
-[bin/test_proxy.py](https://github.com/rclone/rclone/blob/master/test_proxy.py)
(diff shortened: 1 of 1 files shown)


ncw:
Thanks for fixing this.

Unfortunately that file is an auto-generated file - the fix needs to go here so can you update your PR?

https://github.com/rclone/rclone/blob/3626f10f263b54b86dc5ed40a8f3215cb091978a/cmd/serve/proxy/proxy.go#L38

Thanks

ivandeex:
@joschl 
> Unfortunately that file is an auto-generated file - the fix needs to go here so can you update your PR?
> https://github.com/rclone/rclone/blob/3626f10f263b54b86dc5ed40a8f3215cb091978a/cmd/serve/proxy/proxy.go#L38

ivandeex:
Unfortunately rclone is short on resources to finish all PRs unfinished by authors.

albertony:
Closing, as this was fixed in:
- #6008

PR #5515, opened 2021-08-05
doc: fix typo in s3 documentation

gregsadetsky:

fix a typo in the documentation

no

### DIFF
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -251,7 +251,7 @@ y/e/d>
 ### Modified time ###
 
 The modified time is stored as metadata on the object as
-`X-Amz-Meta-Mtime` as floating point since the epoch accurate to 1 ns.
+`X-Amz-Meta-Mtime` as floating point since the epoch is accurate to 1 ns.
 
 If the modification time needs to be updated rclone will attempt to perform a server
 side copy to update the modification if the object can be copied in a single part.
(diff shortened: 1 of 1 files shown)


albertony:
The changed sentence:

> The modified time is stored as metadata on the object as `X-Amz-Meta-Mtime` as floating point since the epoch is accurate to 1 ns.

makes it sound like the 1ns accuracy is the primary reason for using a floating point, which I think is not the case.

In the original:

> The modified time is stored as metadata on the object as `X-Amz-Meta-Mtime` as floating point since the epoch accurate to 1 ns.

the "accurate to 1 ns" is just part of the description of how the values are interpreteted. The main point is that the value is a number of nanoseconds since the epoch.

That is my opinion anyway. I do agree the original sentence is a bit hard to read, with the three "as" etc, so I appreciate you trying to improve it!

gregsadetsky:
Hey, you're right of course, I re-re-read and I'm seeing now what I wasn't seeing before.

Would it better if the sentence became:

> The modified time is stored as metadata on the object as X-Amz-Meta-Mtime as floating point since the epoch, accurate to 1 ns.

i.e. adding a comma before the word "accurate"?

albertony:
> i.e. adding a comma before the word "accurate"?

Yes, I think that is good.

albertony:
Looks good to me.
If you are happy with the results, then I think we can just merge it?
Thank you 👍

gregsadetsky:
That's great, please go ahead! Thank you!

PR #5529, opened 2021-08-13
rc: fix speed does not update in core/stats

negative0:

The `speed` in core/stats does not change the speed which is returned by the global stats using sum.

No. 

#### Checklist

### DIFF
--- a/fs/accounting/stats_groups.go
+++ b/fs/accounting/stats_groups.go
@@ -375,6 +375,9 @@ func (sg *statsGroups) sum(ctx context.Context) *StatsInfo {
 			sum.startedTransfers = append(sum.startedTransfers, stats.startedTransfers...)
 			sum.oldDuration += stats.oldDuration
 			sum.oldTimeRanges = append(sum.oldTimeRanges, stats.oldTimeRanges...)
+			stats.average.mu.Lock()
+			sum.average.speed += stats.average.speed
+			stats.average.mu.Unlock()
 		}
 		stats.mu.RUnlock()
 	}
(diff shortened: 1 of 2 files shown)


fs/accounting/stats_groups.go
```diff
@@ -375,6 +375,9 @@ func (sg *statsGroups) sum(ctx context.Context) *StatsInfo {
 			sum.startedTransfers = append(sum.startedTransfers, stats.startedTransfers...)
 			sum.oldDuration += stats.oldDuration
 			sum.oldTimeRanges = append(sum.oldTimeRanges, stats.oldTimeRanges...)
+			stats.average.mu.Lock()
+			defer stats.average.mu.Unlock()
```
ncw, line 379: Can you do the locking without `defer` as there may be lots of `sg.m` and we don't need to keep everything locked until the end of the function, so

```
stats.average.mu.Lock()
sum.average.speed += stats.average.speed
stats.average.mu.Unlock()
```

I'm not happy about the stats stuff in general - it is all a bit of a mess, but I don't expect you to fix that!

ncw (commented):
Thanks for fixing this :-)

See comment inline.

fs/accounting/stats_groups.go
negative0: Hi @ncw Thanks for bringing that up, I think we should discuss about how stats should be provided. Currently, its very difficult to show everything in the UI. 

I think we should give everything we have about current transfers in the core/stats itself,
- Current upload speed
- Current download speed
- Currently transferring
- Checking/ Renaming/ Deleting
- Transferred

ncw (approved):
Looks great - thanks!

fs/accounting/stats_groups.go
ncw: Do you want to make an issue with what we have at the moment and what you'd like it? Then we can discuss - thank you :-)

negative0: Yes, I will do that. Thanks

negative0: https://github.com/rclone/rclone/issues/5534

PR #5535, opened 2021-08-15
googlephotos: Use encoder for album names

timevortex:

This fixes the problem of album names containing \n which caused problems as local directories could not be created during sync.

Yes: https://forum.rclone.org/t/google-photos-sync-failed-for-multi-lined-album-names/25589

#### Checklist

#### Manual Testing
Please note the diff is intentionally redacted - it show cases the two primary differences in album names due to this change.
```
$ rclone_new lsf -Fp --csv gphotos:shared-album &> rclone_new_2021-08-15_shared-album_lsf.txt
$ rclone lsf -Fp --csv gphotos:shared-album &> rclone_2021-08-15_shared-album_lsf.txt
$ diff rclone_2021-08-15_shared-album_lsf.txt rclone_new_2021-08-15_shared-album_lsf.txt
88c88
< "2018 US Trip - San Francisco, Seattle/"
---
> "2018 US Trip - San Francisco, Seattle／Kirkland, New York/"
164,166c164
< "California Citrus State Historic Park
< 
<  29 Dec 2019/"
---
> California Citrus State Historic Park␊␊ 29 Dec 2019/
$
```

### DIFF
--- a/backend/googlephotos/googlephotos.go
+++ b/backend/googlephotos/googlephotos.go
@@ -20,6 +20,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/backend/googlephotos/api"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/config"
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/config/obscure"
@@ -28,6 +29,7 @@ import (
 	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/log"
+	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/oauthutil"
 	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/rest"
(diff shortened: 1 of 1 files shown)


ncw:
I think this will break existing users as rclone uses the `/` to make a heirarchical directory structure - this may not be what **you** want but you'll see it is [documented here](https://rclone.org/googlephotos/)

```
88c88
< "2018 US Trip - San Francisco, Seattle/"
---
> "2018 US Trip - San Francisco, Seattle／Kirkland, New York/"
```

> Directories within the album directory are also writeable and you may create new directories (albums) under album. If you copy files with a directory hierarchy in there then rclone will create albums with the / character in them. For example if you do

This is easy to fix though - just use `anAlbum.Title = f.opt.Enc.FromStandardPath(anAlbum.Title)` instead of `anAlbum.Title = f.opt.Enc.FromStandardName(anAlbum.Title)`

Other thoughts

- Is this going to need a  `ToStandardPath` for the albums too? I guess maybe not since that I think is the only place rclone finds albums
- What about the photo names? Do you think those need encoding too?

timevortex:
Thanks for taking a look. I've updated the branch to use `f.opt.Enc.FromStandardPath` instead.

As expected the / diffs found initially are gone:
```
$ diff rclone_2021-08-15_shared-album_lsf.txt rclone_new_2021-08-15_shared-album_lsf.txt
164,166c164
< "California Citrus State Historic Park
< 
<  29 Dec 2019/"
---
> California Citrus State Historic Park␊␊ 29 Dec 2019/
366,368c364
$
```

My thoughts to your questions:
> Is this going to need a ToStandardPath for the albums too? I guess maybe not since that I think is the only place rclone finds albums

From what I can tell both `/albums` and `/shared-album` use this function `listAlbums` where I have made this change so I think this should cover both use cases. I don't have any way to test or check with `/albums` though.

> What about the photo names? Do you think those need encoding too?

I haven't seen any issues yet and I think it makes sense as photo names are least likely to have human input - automatically named by camera or phone using the most widely accepted set of characters. Also even manually named files imported into Google Photos will have to have come from an existing filesystem with character limitations. The problem with Album names is that Google Photos hasn't done any sanity checking but with photo names this seems to be automatically taken care of?

I had a quick look and I *think* we would add this after [this](https://github.com/rclone/rclone/blob/9e2fbe0f1a9314e1330b4adc9d69de137be12ee6/backend/googlephotos/googlephotos.go#L612) line:
```
o.remote = f.opt.Enc.FromStandardName(o.remote)
```

I have no way to test it though so I wouldn't feel comfortable making this change :/

timevortex:
I had a quick look at the failing test - doesn't seem to be related to this change?

ncw (approved):
I think that looks great now - thank you :-)

Let's ignore files for the moment - I think you are right about them being sanitised already.

I'll merge it now.

timevortex:
Awesome, thanks!

PR #5538, opened 2021-08-16
docs: Cleanup punctuation

albertony:

Punctuation (and spacing) nitpick in documentation and help output.

The following output from `rclone help flags encoding` triggered my ocd:

```
Backend Flags:
      --acd-encoding MultiEncoder            This sets the encoding for the backend. (default Slash,InvalidUtf8,Dot)
      --azureblob-encoding MultiEncoder      This sets the encoding for the backend. (default Slash,BackSlash,Del,Ctl,RightPeriod,InvalidUtf8)
```

So this changes it to:
```
Backend Flags:
      --acd-encoding MultiEncoder            This sets the encoding for the backend (default Slash,InvalidUtf8,Dot)
      --azureblob-encoding MultiEncoder      This sets the encoding for the backend (default Slash,BackSlash,Del,Ctl,RightPeriod,InvalidUtf8)
```

What is done:
- Help text on backend options:
   - Rules:
      - First line must be a single sentence, only `period`/`full stop` at the end (not within the line).
      - If more description is needed, then include it after an empty line (two linefeeds).
   - Reasoning:
      - When showing command-line help, the first line of the help text is retrieved and sent as "usage" string to the flags library that rclone relies on. It will automatically append information about default value in parenthesis at the end.
      - When showing config ui and generating md/web docs the entire help text will be used, and then the empty line looks nice - separating the "title line" from the "details".
- Usage text on command flags:
    - Rules:
       - Only a single sentence, **no** `period`/`full stop` at the end!
    - Reasoning:
       - This text is used only for showing command-line help, and then the entire text is sent as the "usage" string to the flags library.

**TODO:**

- Should I run `make backenddocs` and commit generated .md files?

#### Checklist


CONTRIBUTING.md
```diff
@@ -273,9 +273,13 @@ If you add a new general flag (not for a backend), then document it in
 alphabetical order.
 
 If you add a new backend option/flag, then it should be documented in
-the source file in the `Help:` field.  The first line of this is used
-for the flag help, the remainder is shown to the user in `rclone
-config` and is added to the docs with `make backenddocs`.
+the source file in the `Help:` field. The first line must be a single
```
ncw, line 276: Should probably mention that it shouldn't be too long - ideally <80 characters. If it is too long the text wraps too many times in a small (or large!) terminal.

backend/crypt/crypt.go
```diff
@@ -45,7 +45,7 @@ func init() {
 					Help:  "Very simple filename obfuscation.",
 				}, {
 					Value: "off",
-					Help:  "Don't encrypt the file names.  Adds a \".bin\" extension only.",
+					Help:  "Don't encrypt the file names. Adds a \".bin\" extension only.",
```
ncw, line 48: This should probably have a `\n\n` in

backend/drive/drive.go
```diff
@@ -360,12 +360,12 @@ commands (copy, sync, etc.), and with all other commands too.`,
 		}, {
 			Name:     "allow_import_name_change",
 			Default:  false,
-			Help:     "Allow the filetype to change when uploading Google docs (e.g. file.doc to file.docx). This will confuse sync and reupload every time.",
+			Help:     "Allow the filetype to change when uploading Google docs (e.g. file.doc to file.docx).\n\nThis will confuse sync and reupload every time.",
```
ncw, line 363: Shorten text before `\n\n` ideally!

docs/content/tardigrade.md
```diff
@@ -1,4 +1,4 @@
----
+s---
```
ncw, line 1: This looks like a mistake?

ncw (approved):
Thank you for your mighty work :-)

I placed a couple of comments inline, but basically looks fine!

ncw:
> Should I run make backenddocs and commit generated .md files?

Auto commit? No, I think I'd rather these were committed by hand.

`make backenddocs` could produce warnings about strings which aren't in style. Though it only ever gets run by a few people so having a --lint flag for it where it could generate problems and run it as part of `make check` would be idea. A job for a new PR anyway!

backend/crypt/crypt.go
albertony: Well, this is in the example part - the help/description of an the example value. I don't think these are being processed like the option help strings, i.e. extract first line, and are never shown on command line, so I tried to leave these as is - just removed an additional space on this one.

docs/content/tardigrade.md
albertony: Darn it. Yes, its probably me attempting to do a quick Ctrl+s - without pressing the Ctrl properly. Happens now and then.. Good that you spotted it!

albertony:
> Should I run make backenddocs and commit generated .md files?

> > Auto commit? No, I think I'd rather these were committed by hand.

What I mean is, I'm a bit confused whether to run `make backenddocs` or not when doing changes like this.

The contributing guide mentions it for writing documentation:

> is added to the docs with make backenddocs.

But its not saying it should be run - so I guess not.

The section for writing a new backend says it should be done, though:

> update them with make backenddocs - revert any changes in other backends

albertony:
> Thank you for your mighty work :-)

It ended up being a bit much for such an "unimportant" thing, so it crossed my mind midway that I would be alone thinking its even worth merging.. But sounds like we are at two, then! :) It leaves a bit better first impression, I think.

ncw:
> > Should I run make backenddocs and commit generated .md files?
> 
> > > Auto commit? No, I think I'd rather these were committed by hand.
> 
> What I mean is, I'm a bit confused whether to run `make backenddocs` or not when doing changes like this.

It is worth running `make backenddocs` so you can see the results with `make serve`.

However whether you commit the changes is a different matter! I quite often run `make backenddocs` then `git checkout` the changes.

I think the answer is you don't have to run `make backenddocs` and commit the changes, but if you do, that is OK!

> The contributing guide mentions it for writing documentation:
> 
> > is added to the docs with make backenddocs.
> 
> But its not saying it should be run - so I guess not.

I usually run `make backenddocs` as part of making a release. However since you are doing a big doc re-org it would be appropriate here too. We don't need drive-by contributors to run `make backendocs`, that's all.

> The section for writing a new backend says it should be done, though:
> 
> > update them with make backenddocs - revert any changes in other backends

So that's is you are making a new backend - you want to see if the docs look OK.

What we definitely don't want is people patching the generated docs which has happened several times before...

> > Thank you for your mighty work :-)
> 
> It ended up being a bit much for such an "unimportant" thing, so it crossed my mind midway that I would be alone thinking its even worth merging.. But sounds like we are at two, then! :) It leaves a bit better first impression, I think.

Consistency in docs is a good thing :-)

albertony:
I see, thank you for that explanation!

albertony:
Force pushed update:
- Changes described in added "config ui" chapter in original description.
- Updated contributing docs with more details about writing docs (after I understand more of the process myself), and made paragraph into list.
- Fixed some small issues (e.g. bool options are created with example help strings "Yes" and "No" and this is detected by config ui to create an improved y/n prompt, which broke when using "Yes." and "No.").

ncw:
Great! Let me know when you are ready for a final once over!

ivandeex:
> PS: I am generating backenddocs and serving docs locally for testing, but have not checked it in.

BTW, here is my recipe to continuously serve docs from the source dir checked out on my development box (after @olefrost advise on the forum). I `docker swarm deploy` the [hugo.yml](https://github.com/ivandeex/ivadocker/blob/master/swarm/hugo.yml) compose script and tell nginx to proxy the service port (1313 in this case) to https://hugo.deex.fun/

albertony:
@ncw: PTAL

albertony:
Force-pushed (sorry), after rebasing on master to fix conflicts.

PR #5539, opened 2021-08-16
docs/vfs: Merge duplicate chunked reading documentation from mount docs.

albertony:

Merging duplicate documentation of VFS chunked reading. It was previously described as part of the common VFS documentation, in the VFS Performance chapter, and in addition there was a separate chapter "chunked reading" in the mount specific documentation. Tried to merge them into a separate chapter in the common VFS documentation: VFS Chunked Reading.

Not sure about the meaning behind the following sentence previously in VFS Performance, if there is a detail relevant to keep:

> This is advantageous because some cloud providers account for reads being all the data requested, not all the data delivered.

I left that out, in favor of the following sentence from the chunked reading doc:

>  This can reduce the used download quota for some remotes by requesting only chunks from the remote that are actually read at the cost of an increased number of requests.

But maybe there is some nuance in the former that should be included?

https://forum.rclone.org/t/documenation-issue-with-rclone-mount/26000

#### Checklist

### DIFF
--- a/cmd/mountlib/help.go
+++ b/cmd/mountlib/help.go
@@ -283,20 +283,4 @@ to use Type=notify. In this case the service will enter the started state
 after the mountpoint has been successfully set up.
 Units having the rclone @ service specified as a requirement
 will see all files and folders immediately in this mode.
-
-### chunked reading
-
-|--vfs-read-chunk-size| will enable reading the source objects in parts.
-This can reduce the used download quota for some remotes by requesting only chunks
-from the remote that are actually read at the cost of an increased number of requests.
-
-When |--vfs-read-chunk-size-limit| is also specified and greater than
-|--vfs-read-chunk-size|, the chunk size for each open file will get doubled
-for each chunk read, until the specified value is reached. A value of |-1| will disable
-the limit and the chunk size will grow indefinitely.
-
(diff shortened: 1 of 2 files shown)


ncw:
> Merging duplicate documentation of VFS chunked reading. It was previously described as part of the common VFS documentation, in the VFS Performance chapter, and in addition there was a separate chapter "chunked reading" in the mount specific documentation. Tried to merge them into a separate chapter in the common VFS documentation: VFS Chunked Reading.

Great idea.
 
> Not sure about the meaning behind the following sentence previously in VFS Performance, if there is a detail relevant to keep:
> 
> > This is advantageous because some cloud providers account for reads being all the data requested, not all the data delivered.

What that means is that if you request 1GB of data from your cloud provider using a Range: request, you may get billed for 1GB data transfer even though you only downloaded 100MB.

> I left that out, in favor of the following sentence from the chunked reading doc:
> 
> > This can reduce the used download quota for some remotes by requesting only chunks from the remote that are actually read at the cost of an increased number of requests.
> 
> But maybe there is some nuance in the former that should be included?

I think you summed it up pretty well!

ncw (approved):
Looks good to me :-)

Merge when ready.

Thanks

Issue #5541, opened 2021-08-18
Added statistic for not transferred files (same hash)


co42:
#### The associated forum post URL from `https://forum.rclone.org`

None

#### What is your current rclone version (output from `rclone version`)?

rclone v1.56.0

#### What problem are you are trying to solve?

When I copy files I want to know how many are not transferred because they don't need two (same hash).

#### How do you think rclone should be changed to solve that?

A new statistics "skips" may be added

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

co42:
To add more context, right now it's not possible to know how many files have been handled by rclone.
If you copy 50 files all already in remote but with 10 changed the stats will be `transfers: 10, checks: 50`, from that I don't know if rclone handled 50 or 60 files (50 the same and 10 new files)

ncw:
I see what you are saying - it is impossible to reconstruct the total number of files either transferred or considered for transfer isn't it?

I'm worried about overwhelming the user with stats that are very similar - `checks` and `skips` will likely be the same number for a lot of transfers.

Perhaps rclone should count `skips` instead of `checks` (perhaps named `checks` to stay backwards compatible!).

This could be counting a check and take it away again if we transfer that file.

Or maybe we should show the total number of files transferred or considered for transfer at the end.

co42:
> Perhaps rclone should count skips instead of checks (perhaps named checks to stay backwards compatible!).
>
> This could be counting a check and take it away again if we transfer that file.

Looks good to me, this stat would have more value I think.

> Or maybe we should show the total number of files transferred or considered for transfer at the end.

It's a stat that the user can easily compute from `transfers + checks` if we implement the first solution.


As soon as a decision is taken I can modify my PR.

co42:
@ncw Is https://github.com/rclone/rclone/issues/5541#issuecomment-903508465 OK for you ?

PR #5542, opened 2021-08-18
Add a new statistic "skips" counting skipped transfers (already sync)

co42:

On copy, move, ... New statistic "skips" counting how many files are not transferred because they don't need to (same hash).

https://github.com/rclone/rclone/issues/5541

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -47,6 +47,7 @@ type StatsInfo struct {
 	renameQueueSize   int64
 	deletes           int64
 	deletedDirs       int64
+	skips             int64
 	inProgress        *inProgress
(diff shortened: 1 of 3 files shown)


ncw:
Looking at the stats we have at the moment, I'm just trying to work out if we need skips.

We have `transfers` and `checks` so I think `skips = checks - transfers` - would you agree?

co42:
Say we 10 files with the same content locally and on the remote
If we **modify** 2 of those files on the next copy we get `checks=10 transfers=2`
If we **add** 2 files we will also have `checks=10 transfers=2`

So we don't know if we copied 10 or 12 fields and if 8 or 10 are skipped.

PR #5551, opened 2021-08-23
config: fix handling of default, exclusive and required properties of multiple-choice options

### LINKED ISSUE #5549: Multiple choice config option default handling

For exclusive multiple choice options, handling of default value is not working according to description:

```
Authentication type
Enter a string value. Press Enter for the default ("standard").
Choose a number from below, or type in an existing value
 1 / Standard authentication - use this if you're a normal Jottacloud user.
   \ "standard"
 2 / Legacy authentication - this is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.
   \ "legacy"
 3 / Telia Cloud authentication - use this if you are using Telia Cloud.
   \ "telia"
config_type>
config_type>
config_type>
config_type>
```

In contrast to what is indicated, pressing enter will not give the default - you are required to enter one of the three listed values.

#### How to use GitHub

### LINKED ISSUE #5562: OneDrive configuration doesn't properly set default config_type

n/a

During OneDrive configuration, the `config_type` prompt says to press enter for the default value (`onedrive`), but pressing enter at the prompt doesn't continue, instead prompting for `config_type` again. Filling in `onedrive` as the value works fine.

`rclone 1.56.0`

NixOS, 64 bit

OneDrive

`rclone config`

```
Type of storage to configure.
 1 / 1Fichier
   \ "fichier"
 2 / Alias for an existing remote
   \ "alias"
 3 / Amazon Drive
   \ "amazon cloud drive"
 4 / Amazon S3 Compliant Storage Providers including AWS, Alibaba, Ceph, Digital Ocean, Dreamhost, IBM COS, Minio, SeaweedFS, and Tencent COS
   \ "s3"
 5 / Backblaze B2
   \ "b2"
 6 / Box
   \ "box"
 7 / Cache a remote
   \ "cache"
 8 / Citrix Sharefile
   \ "sharefile"
 9 / Compress a remote
   \ "compress"
10 / Dropbox
   \ "dropbox"
11 / Encrypt/Decrypt a remote
   \ "crypt"
12 / Enterprise File Fabric
   \ "filefabric"
13 / FTP Connection
   \ "ftp"
14 / Google Cloud Storage (this is not Google Drive)
   \ "google cloud storage"
15 / Google Drive
   \ "drive"
16 / Google Photos
   \ "google photos"
17 / Hadoop distributed file system
   \ "hdfs"
18 / Hubic
   \ "hubic"
19 / In memory object storage system.
   \ "memory"
20 / Jottacloud
   \ "jottacloud"
21 / Koofr
   \ "koofr"
22 / Local Disk
   \ "local"
23 / Mail.ru Cloud
   \ "mailru"
24 / Mega
   \ "mega"
25 / Microsoft Azure Blob Storage
   \ "azureblob"
26 / Microsoft OneDrive
   \ "onedrive"
27 / OpenDrive
   \ "opendrive"
28 / OpenStack Swift (Rackspace Cloud Files, Memset Memstore, OVH)
   \ "swift"
29 / Pcloud
   \ "pcloud"
30 / Put.io
   \ "putio"
31 / QingCloud Object Storage
   \ "qingstor"
32 / SSH/SFTP Connection
   \ "sftp"
33 / Sugarsync
   \ "sugarsync"
34 / Tardigrade Decentralized Cloud Storage
   \ "tardigrade"
35 / Transparently chunk/split large files
   \ "chunker"
36 / Union merges the contents of several upstream fs
   \ "union"
37 / Uptobox
   \ "uptobox"
38 / Webdav
   \ "webdav"
39 / Yandex Disk
   \ "yandex"
40 / Zoho
   \ "zoho"
41 / http Connection
   \ "http"
42 / premiumize.me
   \ "premiumizeme"
43 / seafile
   \ "seafile"
Storage> onedrive
OAuth Client Id
Leave blank normally.
client_id>
OAuth Client Secret
Leave blank normally.
client_secret>
Choose national cloud region for OneDrive.
Enter a string value. Press Enter for the default ("global").
 1 / Microsoft Cloud Global
   \ "global"
 2 / Microsoft Cloud for US Government
   \ "us"
 3 / Microsoft Cloud Germany
   \ "de"
 4 / Azure and Office 365 operated by 21Vianet in China
   \ "cn"
region>
Edit advanced config?
y) Yes
n) No (default)
y/n>
Use auto config?
 * Say Y if not sure
 * Say N if you are working on a remote or headless machine

y) Yes (default)
n) No
y/n> n
For this to work, you will need rclone available on a machine that has
a web browser available.

For more help and alternate methods see: https://rclone.org/remote_setup/

Execute the following on the machine with the web browser (same rclone
version recommended):

	rclone authorize "onedrive"

Then paste the result.

config_token> [REDACTED]
Type of connection
Enter a string value. Press Enter for the default ("onedrive").
Choose a number from below, or type in an existing value
 1 / OneDrive Personal or Business
   \ "onedrive"
 2 / Root Sharepoint site
   \ "sharepoint"
 3 / Sharepoint site name or URL (e.g. mysite or https://contoso.sharepoint.com/sites/mysite)
   \ "url"
 4 / Search for a Sharepoint site
   \ "search"
 5 / Type in driveID (advanced)
   \ "driveid"
 6 / Type in SiteID (advanced)
   \ "siteid"
 7 / Sharepoint server-relative path (advanced, e.g. /teams/hr)
   \ "path"
config_type>
config_type>
```

#### How to use GitHub

albertony:

Main purpose of this PR is to fix/improve the handling of `Default`, `Exclusive` and `Required` properties of multiple-choice options. Does also include some additional improvements around the config ui, fixes and nitpicks discovered during development, such as default value prompting (not only for multiple-choice), required password editing, several prompting/spelling tweaks etc.

**Fix handling of default, exclusive and required properties**

Previously an empty input (just pressing enter) was only allowed for multiple-choice options that did not have the Exclusive property set. Here is an example session with an exclusive multiple-choice option:
```
Option config_type.
Authentication type
Enter a string value. Press Enter for the default ("standard").
Choose a number from below, or type in an existing value.
   / Standard authentication.
 1 | Use this if you're a normal Jottacloud user.
   \ "standard"
   / Legacy authentication.
 2 | This is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.
   \ "legacy"
   / Telia Cloud authentication.
 3 | Use this if you are using Telia Cloud.
   \ "telia"
config_type>
config_type>
config_type>
```

With this change the existing `Required` property is introduced into the multiple choice handling, so that one can have `Exclusive` and `Required` options where only a value from the list is allowed, and one can have `Exclusive` but not `Required` options where an empty value is accepted but any non-empty value must still be matching an item from the list. And, as before, a multiple choice that is neither `Exclusive` nor `Required`, where any input is accepted.

Some notes about `Required`:
- The `Required` property means the option must have a value, but it can be explicitely set or it can be an implicit default.
- This means that when there is a default value, that is not an empty string, then setting `Required: true` makes no difference.
- When there is no default value, setting `Required: true` means user have to supply a value, cannot simply press Enter.
- This interpretation did not change with this PR, that was how it was done for free-text input already, but the same logic was introduced for multiple-choice inputs as well.

Note that there are several combinations of properties that will achieve the same: `Exclusive` and `Required` boolean properties set true or false, `Default` value set or not set (nil, treated as empty string), and in addition one can have empty string (`""`) as an entry in the multiple-choice list. Not something that is directly intentional, but more of a consequence, backwards compat etc. The way to achieve the different behavior can be summarized:
- If only values from the specified list should be allowed: `Exclusive=true && Required=true && Default=""`
   - "Exclusive and Required, with no Default"
- If empty string and values from specified list should be allowed: `Exclusive=true && (Required=false || Default!="")`
   - "Exclusive, and not Required or a Default"
- If anything but empty string should be allowed: `Exclusive=false && Required=true && Default==""`
   - "Required, but not Exclusive and no Default"
- If anything should be allowed: `Exclusive=false && Required=false`
   - "Not Required and not Exclusive"

In other words, the recommended approach is: For multiple-choice option do specify `Examples`, and then:
- The default behavior is to allow any value, either from specified examples, or any other value. To restrict exclusively to the specified examples, set `Exclusive=true`.
- If empty string should not be allowed then set `Required=true`, and do not set `Default` (leave it at `nil`).

**Press Enter for the default**

A related issue is the misleading description `Press Enter for the default ("")`, as can be seen from example above. This is also an issue for regular free-text (non-multiple-choice) input, e.g.:
```
FTP host to connect to
 1 / Connect to ftp.example.com
   \ "ftp.example.com"
host>
This value is required and it has no default.
 1 / Connect to ftp.example.com
   \ "ftp.example.com"
host>
```
This was also fixed, now the message `Press Enter for the default ("standard").` will only be shown when this is actually allowed.

**Quoting of values in prompt**

Previously the default and multiple choice values were always shown in quotes. For strictly numeric values, e.g. port number, this instinctively seemed wrong (at least with the `FTP port, leave blank to use default (21)` case, described later). So I tried a quick change in the code code to only quote values that are string types. E.g. instead of:
```
Enter a string value. Press Enter for the default ("standard").
Enter a signed integer. Press Enter for the default ("21").
```
Then:
```
Enter a string value. Press Enter for the default ("standard").
Enter a signed integer. Press Enter for the default (21).
```
But then there are other types of values which are not strings, but not simple value types either:
```
Enter a size with suffix K,M,G,T. Press Enter for the default (10Mi).
Enter a boolean value (true or false). Press Enter for the default (false).
Enter a encoder.MultiEncoder value. Press Enter for the default (Slash,LtGt,DoubleQuote,Colon,Question,Asterisk,Pipe,Del,Ctl,InvalidUtf8,Dot).
```

And that made me think: The input that user enters in the config ui is not programming language / go syntax. When entering string the user should not surround it by quotes. So just as a number input is `21` a string input is also just `this is a string`.

The code that is printing default value is converting it to a string and then using the format verb `%q`, which prints it *safely escaped with Go syntax*. This means surrounded by quotes, but also that `\` is escaped to `\\`, `"` to `\"` etc. This is different from the actual input the user should enter.

Example where rclone is printing default value with `%q`, an option where default is current username with domain prefix:
- When default value is printed with `%q` it is escaped like this:
  ```
  Enter a string value. Press Enter for the default ("domain\\user").
  ```
- But if a user then were to enter his own value in the same format:
  ```
  user>domain\\otheruser
  ```
  Then it would be stored in the config file as `domain\\otheruser`, and when read back this will be escaped into string `"domain\\\\otheruser"`. This can be observed if editing the config, then the default will be the value from config,
  again formatted with `%q`:
  ```
  Press Enter for the default ("domain\\\\otheruser").
  ```

The same thing is true for multiple-choice options, where rclone is printing the "example" values with `%q`:
  ```
  Username
  Choose a number from below, or type in your own string value.
  Press Enter for the default ("").
   1 / Current user
     \ "domain\\user"
   2 / This is "me", yes it is
     | More text here
     \ "mini \"me\" is me"
  ```

An improvement could be to print the string values with `%s` or `%v` instead of `%q`:
  ```
  Enter a string value. Press Enter for the default (some standard value).
  ```
Another variant is to manually add quotes around, `fmt.Sprintf("\"%s\"", v)` to get
quotes without escaping:
  ```
  Enter a string value. Press Enter for the default ("some standard value").
  ```
But then if the original value contains escaped quotes, although not very likely, this will look wrong:
  ```
  Enter a string value. Press Enter for the default ("some "quoted" value").
  ```
So I'm thinking it is best just to print the value without quotes or anything: Everything between `(` and `)` is exactly as the user would have to write it! ...Although same thing here if original value contains parens as well, but parens within parens maybe does not look *that* wrong as with quotes within quotes? Regardless, the main purpose with no quoting and just parens, is it looks a bit cleaner, to avoid confusion with regards to if values containing spaces must be quoted or not, less targeted at programmers and more at end-users.. I don't know. Thats the idea, at least.. :)

But then for non-required options without a default value, the current version prompts:
```
Press Enter for the default ("").
```
With the change above this would look worse:
```
Press Enter for the default ().
```
But can easily consider this in code, so that this case is instead shown as:
```
Press Enter to leave empty.
```

(not sure if it should say "to leave empty", or "to skip" or "not set" or something like that instead?)

But then when it comes to how multiple-choice are prompted, the option value for each choice are in a section together with the help text today, and not separated in the same way as the default values described above. Currently:
```
   2 / This is "me", yes it is
     | More text here
     \ "this \"me\" is not really me"
```
But perhaps we can just do the same here - plain string (content), and add parenthesis to separate from help text:
```
   2 / This is "me", yes it is
     | More text here
     \ (this "me" is not really me)
```

That is what I have done, so the "look" will be a bit different - notice the values in parenthesis:

```
Type of storage to configure.
Choose a number from below, or type in your own string value.
 1 / 1Fichier
   \ (fichier)
 2 / Alias for an existing remote
   \ (alias)
 3 / Amazon Drive
   \ (amazon cloud drive)
Storage>

Host to connect to
Choose a number from below, or type in your own string value.
 1 / Connect to ftp.example.com
   \ (ftp.example.com)
host>

Username
Press Enter for the default (domain\user).
user>

Port
Press Enter for the default (21).
port>

Username2
Choose a number from below, or type in your own string value.
Press Enter to leave empty.
 1 / Current user
   \ (domain\user)
   / This is "me", yes it is
 2 | Here we go with more text
   \ (this "me" is not really me)
 3 / This is an empty choice
   \ ()
user2>
```

(The choice 3 in the last example still shows an empty `()`. This is a very special case where the multiple choice list have an explicit entry with value empty string. Rarely used, and is a bit confusing compared to the Default, Required etc discussed above, but I have seen it. Left it just printing `()` for now, have not come up with a better option yet).

**Fixed edit of required password**

Another related issue is in the process of editing existing remotes, where there is a password. The editing process was changed in v1.56 from a separate edit workflow:
```
Value "port" = "2222"
Edit? (y/n)>
y) Yes
n) No (default)
```
..into making edit being more like the initial create - just with existing values as default. This lead to a regression for required password inputs, as reported here: https://forum.rclone.org/t/v1-56-beta-i-have-to-re-enter-the-password-when-editing-the-remote/26437. Fixed this by making password input handle the `Default` property of options, in addition to `Required`, the same way as other option types. When editing the existing value will be the default value, and user gets the additional choice `n` to keep existing/default in addition to `y` and `g`:
```
FTP password
y) Yes, type in my own password
g) Generate random password
n) No, keep existing (default)
```
PS: The changes in the edit process introduced in v1.56 came with the following commits:
   https://github.com/rclone/rclone/commit/296ceadda61b82fe47e22e71ca85d01b00dc7c67#diff-d43e96d17fe826ebacb1c626935c1ced77f6c508a47cd7fa19832eb724c473c2
   https://github.com/rclone/rclone/commit/77cda6773ca6d46d95e7e63f0d6b2be93ce81af3#diff-d43e96d17fe826ebacb1c626935c1ced77f6c508a47cd7fa19832eb724c473c2

**Additional improvements to config ui prompts:**

- Moved the code for input of the three types, free-text, password and multiple-choice, into separate functions, and made their prompting a bit mot tailor-made.
- Improved "title" for multiple-choice options, especially the ones without default:
    ```
    Enter a string value.
    Choose a number from below, or type in an existing value
    ```
    Now:
    ```
    Choose a number from below, or type in an existing string value.
    ```
- Improved feedback on wrong input: Print message such as `This value is required and it has no default.` and then a new "prompt" such as `my_option>`, also for multiple-choice options (consistent for all option types):
    ```
    Choose a number from below, or type in your own string value.
     1 / Rclone should act on behalf of a user.
       \ "user"
     2 / Rclone should act on behalf of a service account.
       \ "enterprise"
    my_option>
    This value is required and it has no default.
    my_option>
    This value is required and it has no default.
    my_option>
    ```
    Previously it would, depending on error, whether free-text input or multiple-choice etc, either just repeat the prompt as in the initial example at the top or repeat the entire "question" with description (and choices) as the second example with ftp host input.
- Made ftp and ssh port and username option default values explicit, using `Default` property.
  - This avoids a prompt that would be repetative and confusing.
  - When creating a ftp remote you get this question:
  ```
  FTP port, leave blank to use default (21)
  port>
  ```
  If explicitely entering a port number, e.g. 2222, then when later editing the remote you get this question:
  ```
  FTP port, leave blank to use default (21)
  Enter a string value. Press Enter for the default ("2222").
  port>
  ```
  (as mentioned above I've removed the quoting around the "2222" value so with this change it would show `Press Enter for the 
  default (2222).`)

  So default is 21, "", 2222, or what??

  The explanation is that the help text of the ftp port option is defined as:
  ```
  Help: "FTP port, leave blank to use default (21)",
  ```
  So the "leave blank to use default (21)" was actually "hard-coded" in the help text. The ftp backend implementation will connect to port 21 when nothing else is specified - this is an *implicit* default. The option definition did not actually have a `Default` parameter, meaning it is `nil`. When the config ui prompts for this option during creation it will therefore add the description `Press Enter for the default ("").` When editing an existing remote, where the ftp option has a value explicitely set (e.g. 2222), then the existing value will be the option's `Default` parameter and therefore the prompt changes to `Press Enter for the default ("2222").`.
  
  Fixed by setting the `Default` property of the option, and removing the description of default from help text.
  
  Similar for username, where it would show:
   ```
  FTP username, leave blank for current username, DOMAIN\user
  user>
  ```
  But now instead:
  ```
  FTP username
  Press Enter for the default (DOMAIN\user).
  user>
  ```
- Includes some initial commits with minor cleanup done to make the main refactoring easier, but these did not end up being important for the solution (I can remove them if this is preferred):
  - Removed explicit setting of defaults `Required: false` and `Default: ""`
  - Removed setting `Required: true` when there is a `Default` value set.

**NOTE: Some issues and imperfections noticed but not fixed**

- The datatype of options are decided by the type of the `Default` property, but if there is no default then the type is always referred to as string (`Enter a string value`), even if it naturally is an integer (like a port number).
   - We could have introduced a new property `Type` that defines the value type expected, e.g. "bool" or "string". If this is not  specified we could check `Default` as before (backwards compat.).
- When editing an existing remote that has set an explicit value for an option that has an implicit default, it is not possible to "clear" the value and get back to the implicit default. E.g. for ftp port number: Ftp backend will use port 21 if none is specified. Say an existing remote has this set to 2122. When editing this remote, to be able to use the default value 21 one have to explicit set it:
  ```
  FTP port, leave blank to use default (21)
  Enter a signed integer. Press Enter for the default (2122).
  port>
  ```  
  Empty input here (just pressing enter) will keep existing value 2122.
- The question to edit advanced config is repeated when done with all advanced options, but could be intentional?
  ```
  Edit advanced config?
  y) Yes
  n) No (default)
  y/n>
  ```
- Multiple choice options cannot take an integer as value! Any input entered by user that is convertible to integer will be treated as a number referencing the choice number. So if there are two choices then entering `2` will pick the second choice. Entering `999` will not pick any, but will be treated as a invalid choice because it there are no choices with that number. So if one had a multiple-choice option for configuring a port number, e.g. predefined values `1) 80`, `2) 8080`, `3) 8088`, but the option not being `Exclusive`, then entering 1, 2 or 3 would pick the corresponding predefined value, but entering 8 or 80 or 8888, would not be allowed, although that is supposed to be a way for the user to enter a custom port number.

**TODO: Remaining**

Update references to the interactive config prompts in documentation, to match the changes. Especially the example config session shown in most backend md docs.

Fixes #5549
Fixes #5562

Also fixes an issue reported in forum: https://forum.rclone.org/t/v1-56-beta-i-have-to-re-enter-the-password-when-editing-the-remote/26437

**Also related to #5538, should probably merge that first (if accepted) and then fix up a few conflicts before merging this one.**

#### Checklist

### DIFF
--- a/backend/fichier/fichier.go
+++ b/backend/fichier/fichier.go
@@ -42,18 +42,15 @@ func init() {
 		}, {
 			Help:     "If you want to download a shared folder, add this parameter.",
 			Name:     "shared_folder",
-			Required: false,
 			Advanced: true,
 		}, {
 			Help:       "If you want to download a shared file that is password protected, add this parameter.",
 			Name:       "file_password",
-			Required:   false,
 			Advanced:   true,
 			IsPassword: true,
 		}, {
 			Help:       "If you want to list the files in a shared folder that is password protected, add this parameter.",
 			Name:       "folder_password",
-			Required:   false,
 			Advanced:   true,
 			IsPassword: true,
 		}, {
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -51,11 +51,13 @@ func init() {
 			Help:     "FTP host to connect to.\n\nE.g. \"ftp.example.com\".",
 			Required: true,
 		}, {
-			Name: "user",
-			Help: "FTP username, leave blank for current username, " + currentUser + ".",
+			Name:    "user",
+			Help:    "FTP username.",
+			Default: currentUser,
 		}, {
-			Name: "port",
-			Help: "FTP port, leave blank to use default (21).",
+			Name:    "port",
+			Help:    "FTP port number.",
+			Default: 21,
 		}, {
 			Name:       "pass",
 			Help:       "FTP password.",
--- a/backend/hdfs/hdfs.go
+++ b/backend/hdfs/hdfs.go
@@ -22,9 +22,8 @@ func init() {
 			Help:     "Hadoop name node and port.\n\nE.g. \"namenode:8020\" to connect to host namenode at port 8020.",
 			Required: true,
 		}, {
-			Name:     "username",
-			Help:     "Hadoop user name.",
-			Required: false,
+			Name: "username",
+			Help: "Hadoop user name.",
 			Examples: []fs.OptionExample{{
 				Value: "root",
 				Help:  "Connect to hdfs as root.",
@@ -36,7 +35,6 @@ func init() {
 Enables KERBEROS authentication. Specifies the Service Principal Name
 (SERVICE/FQDN) for the namenode. E.g. \"hdfs/namenode.hadoop.docker\"
 for namenode running as service 'hdfs' with FQDN 'namenode.hadoop.docker'.`,
-			Required: false,
 			Advanced: true,
 		}, {
 			Name: "data_transfer_protection",
@@ -46,7 +44,6 @@ Specifies whether or not authentication, data signature integrity
 checks, and wire encryption is required when communicating the the
 datanodes. Possible values are 'authentication', 'integrity' and
 'privacy'. Used only with KERBEROS enabled.`,
-			Required: false,
 			Examples: []fs.OptionExample{{
 				Value: "privacy",
 				Help:  "Ensure authentication, integrity and encryption enabled.",
--- a/backend/koofr/koofr.go
+++ b/backend/koofr/koofr.go
@@ -39,7 +39,6 @@ func init() {
 		}, {
 			Name:     "mountid",
 			Help:     "Mount ID of the mount to use.\n\nIf omitted, the primary mount is used.",
-			Required: false,
 			Default:  "",
 			Advanced: true,
 		}, {
--- a/backend/koofr/koofr.go
+++ b/backend/koofr/koofr.go
@@ -34,18 +34,15 @@ func init() {
 			Name:     "endpoint",
 			Help:     "The Koofr API endpoint to use.",
 			Default:  "https://app.koofr.net",
-			Required: true,
 			Advanced: true,
 		}, {
 			Name:     "mountid",
 			Help:     "Mount ID of the mount to use.\n\nIf omitted, the primary mount is used.",
-			Default:  "",
 			Advanced: true,
 		}, {
 			Name:     "setmtime",
 			Help:     "Does the backend support setting modification time.\n\nSet this to false if you use a mount ID that points to a Dropbox or Amazon Drive backend.",
 			Default:  true,
-			Required: true,
 			Advanced: true,
 		}, {
 			Name:     "user",
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -59,11 +59,13 @@ func init() {
 			Help:     "SSH host to connect to.\n\nE.g. \"example.com\".",
 			Required: true,
 		}, {
-			Name: "user",
-			Help: "SSH username, leave blank for current username, " + currentUser + ".",
+			Name:    "user",
+			Help:    "SSH username.",
+			Default: currentUser,
 		}, {
-			Name: "port",
-			Help: "SSH port, leave blank to use default (22).",
+			Name:    "port",
+			Help:    "SSH port number.",
+			Default: 22,
 		}, {
 			Name:       "pass",
 			Help:       "SSH password, leave blank to use ssh-agent.",
--- a/backend/tardigrade/fs.go
+++ b/backend/tardigrade/fs.go
@@ -99,13 +99,11 @@ func init() {
 			{
 				Name:     "access_grant",
 				Help:     "Access grant.",
-				Required: false,
 				Provider: "existing",
 			},
 			{
 				Name:     "satellite_address",
 				Help:     "Satellite address.\n\nCustom satellite address should match the format: `<nodeid>@<address>:<port>`.",
-				Required: false,
 				Provider: newProvider,
 				Default:  "us-central-1.tardigrade.io",
 				Examples: []fs.OptionExample{{
@@ -123,13 +121,11 @@ func init() {
 			{
 				Name:     "api_key",
 				Help:     "API key.",
-				Required: false,
 				Provider: newProvider,
 			},
 			{
 				Name:     "passphrase",
 				Help:     "Encryption passphrase.\n\nTo access existing objects enter passphrase used for uploading.",
-				Required: false,
 				Provider: newProvider,
 			},
 		},
--- a/backend/tardigrade/fs.go
+++ b/backend/tardigrade/fs.go
@@ -84,10 +84,9 @@ func init() {
 		},
 		Options: []fs.Option{
 			{
-				Name:     fs.ConfigProvider,
-				Help:     "Choose an authentication method.",
-				Required: true,
-				Default:  existingProvider,
+				Name:    fs.ConfigProvider,
+				Help:    "Choose an authentication method.",
+				Default: existingProvider,
 				Examples: []fs.OptionExample{{
 					Value: "existing",
 					Help:  "Use an existing access grant.",
--- a/backend/union/union.go
+++ b/backend/union/union.go
@@ -33,25 +33,21 @@ func init() {
 			Help:     "List of space separated upstreams.\n\nCan be 'upstreama:test/dir upstreamb:', '\"upstreama:test/space:ro dir\" upstreamb:', etc.",
 			Required: true,
 		}, {
-			Name:     "action_policy",
-			Help:     "Policy to choose upstream on ACTION category.",
-			Required: true,
-			Default:  "epall",
+			Name:    "action_policy",
+			Help:    "Policy to choose upstream on ACTION category.",
+			Default: "epall",
 		}, {
-			Name:     "create_policy",
-			Help:     "Policy to choose upstream on CREATE category.",
-			Required: true,
-			Default:  "epmfs",
+			Name:    "create_policy",
+			Help:    "Policy to choose upstream on CREATE category.",
+			Default: "epmfs",
 		}, {
-			Name:     "search_policy",
-			Help:     "Policy to choose upstream on SEARCH category.",
-			Required: true,
-			Default:  "ff",
+			Name:    "search_policy",
+			Help:    "Policy to choose upstream on SEARCH category.",
+			Default: "ff",
 		}, {
-			Name:     "cache_time",
-			Help:     "Cache time of usage and free space (in seconds).\n\nThis option is only useful when a path preserving policy is used.",
-			Required: true,
-			Default:  120,
+			Name:    "cache_time",
+			Help:    "Cache time of usage and free space (in seconds).\n\nThis option is only useful when a path preserving policy is used.",
+			Default: 120,
 		}},
 	}
 	fs.Register(fsi)
--- a/docs/content/alias.md
+++ b/docs/content/alias.md
@@ -33,7 +33,7 @@ First run:
 This will guide you through an interactive setup process:
 
 ```
-No remotes found - make a new one
+No remotes found, make a new one?
 n) New remote
 s) Set configuration password
 q) Quit config
--- a/docs/content/amazonclouddrive.md
+++ b/docs/content/amazonclouddrive.md
@@ -53,7 +53,7 @@ Here is an example of how to make a remote called `remote`.  First run:
 This will guide you through an interactive setup process:
 
 ```
-No remotes found - make a new one
+No remotes found, make a new one?
 n) New remote
 r) Rename remote
 c) Copy remote
--- a/docs/content/azureblob.md
+++ b/docs/content/azureblob.md
@@ -19,7 +19,7 @@ configuration.  For a remote called `remote`.  First run:
 This will guide you through an interactive setup process:
 
 ```
-No remotes found - make a new one
+No remotes found, make a new one?
 n) New remote
 s) Set configuration password
 q) Quit config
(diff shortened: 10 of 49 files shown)


albertony:
Thanks for the pointers! I was also thinking there should be unit tests covering the different combinations, but was worried the interactive nature of the code would make it difficult. Looking more thorough into the existing ui_tests, it seems it should not be that hard, though. Will definitely give it a try. I also need to go through the code again, do more manual testing etc, so still a work in progress that may take some time.

ncw:
Let me know when you want another review :-)

albertony:
@ncw: Please take a look.

As mentioned there will be conflict with #5538, so if/when one of them is merged the other needs a new iteration.

albertony:
@ncw: Should I just merge this, to get it in the beta testing?

albertony:
I'll take 🤐 as a yes! 😆

PR #5555, opened 2021-08-26
docs/config: remove use of backticks around words within a larger code block

albertony:

Minor improvement in docs for rclone config subcommands, removing backticks that are intended for markdown code elements in strings that are inserted into larger code blocks.

Result is, instead of:

```
rclone config delete `name` [flags]
```

Then this:
```
rclone config delete name [flags]
```

Also removed `name` code element in short string of rclone config delete, as it makes it stick out in the docs at https://rclone.org/commands/, none of the other have this.

#### Checklist

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -196,7 +196,7 @@ Note that |bin/config.py| in the rclone source implements this protocol
 as a readable demonstration.
 `, "|", "`")
 var configCreateCommand = &cobra.Command{
-	Use:   "create `name` `type` [`key` `value`]*",
+	Use:   "create name type [key value]*",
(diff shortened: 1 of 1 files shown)


ncw (approved):
I think this looks find and is consistent with the other docs.

I'll just note that this was changed in 74b8cbfb841c1b5491d5f840b2c541e91177bd6c where these parameters were marked as

    <name>

which is CLI standard, but that causes them to be omitted from the HTML rendering of the docs! These were changed to

    `name`

To make the fact that they are quoted names apparent.

I think it is obvious enough without it though and consistent with the rest of the docs.

PR #5556, opened 2021-08-26
[DISCUSSION ONLY issue #5553] KF_DIRECT_STAT: file read seems to work. ls does not at all.

kforner:

This is for discussion, about issue https://github.com/rclone/rclone/issues/5553
I've implemented a proof of concept of direct file access on s3 mount that does not populate directories in Lookup Requests.
That seems to work for accessing files:
- now it takes less than 1 second instead of ~25s for reading a file (md5sum) in a large directory, or under a large directory.
- ls does not work anymore though

https://github.com/rclone/rclone/issues/5553

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ fuzz-build.zip
 *.orig
 *.rej
 Thumbs.db
+.vscode/launch.json
+.vscode/settings.json
+KARL_DEBUG.md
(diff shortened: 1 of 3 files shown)


ncw:
Comments on this PR https://github.com/rclone/rclone/issues/5553#issuecomment-907760854

ncw:
I'm going to close this PR since it is very old and needs work.

Thank you for submitting and sorry we never merged.

Issue #4810, opened 2020-11-26
feature: implement advanced change detection


ivandeex:
**(Q)** Currently rclone provides `copy` and `sync` commands spinning around three basic traits (_hashsum_, _modtime_, _size_) to detect changes using internal [equal()](https://github.com/rclone/rclone/blob/v1.53.3/fs/operations/operations.go#L136) function and its helpers.
The latter used to grow _organically_ in response to the stream of [change detection](https://github.com/rclone/rclone/issues?q=label%3A%22change+detection%22) feature requests, often serving contradictory user requirements such as _consistency_ (this PR https://github.com/rclone/rclone/pull/4011#issuecomment-592186479), _cost savings_ (see https://github.com/rclone/rclone/issues/2889#issuecomment-508683225) and many others (sometimes [unexplained ;)](https://github.com/rclone/rclone/issues/4793)), consequently blurring the line between "bug" ang "feature".

This in turn has led to a number of change-related CLI flags (see https://rclone.org/flags), some of them having negative (_no-check_), conditional (_x-if-y_) or even double-negative (_noignore_) names, which interact in exotic ways with 30+ supported backends (https://rclone.org/overview#features) and their (in)ability to provide required traits.

On top of that we have [filtering](https://github.com/rclone/rclone/issues?q=label%3A%22filtering%22) flags and thin difference of `sync` vs `copy`. I feel that many users have problems telling it, and documentation doesn't stress enough.

In the end we might expect that _**any** behavior can be requested_, even that that feels esoteric or illogical.
¯\\\_(ツ)\_/¯

**(A)** This feature request is up to:
- inventory existing flags related to change detection
- assign ordering priorities (weights?) to flags and conditions
- describe which flag(s) win if they are used in combination
- probably develop a syntax similar (in idea) to Unix PAM (see https://linux.die.net/man/5/pam.conf or https://docs.ansible.com/ansible/latest/collections/community/general/pamd_module.html#examples) or NSS directives (https://linux.die.net/man/5/nsswitch.conf)
- (I started thinking about CSS rule weights but stopped myself ;))
- see whether existing flags can be described by the new syntax
- make it as simple as possible (think okkam's razor)
- revise test coverage, make sure existing shell wrappers around rclone don't break
- add a simple enough automata in _equal_ to handle the logic
- add a combined flag like `--change-criteria +size,-hash,!modtime,?whatever` with a syntax reach enough to cover existing and esoteric cases
- improve docs: add examples
[...]

PR #5576, opened 2021-09-06
vfs: check if equal before uploading to remote

wloot:
#### Checklist

### DIFF
--- a/vfs/vfscache/item.go
+++ b/vfs/vfscache/item.go
@@ -581,13 +581,18 @@ func (item *Item) _store(ctx context.Context, storeFn StoreFn) (err error) {
 	if cacheObj != nil {
 		o, name := item.o, item.name
 		item.mu.Unlock()
-		o, err := operations.Copy(ctx, item.c.fremote, o, name, cacheObj)
-		item.mu.Lock()
-		if err != nil {
-			return errors.Wrap(err, "vfs cache: failed to transfer file from cache to remote")
+		if o == nil || !operations.Equal(ctx, cacheObj, o) {
+			o, err := operations.Copy(ctx, item.c.fremote, o, name, cacheObj)
+			item.mu.Lock()
+			if err != nil {
+				return errors.Wrap(err, "vfs cache: failed to transfer file from cache to remote")
+			}
+			item.o = o
+			item._updateFingerprint()
+		} else {
+			item.mu.Lock()
+			fs.Debugf(item.name, "vfs cache: unchanged skipping")
 		}
-		item.o = o
-		item._updateFingerprint()
 	}
 
 	item.info.Dirty = false


ncw (commented):
This patch looks like a sensible idea - thank you :-)

I'm slightly concerned about changing the behaviour when using backends which don't support modtimes or checksums.

I'd like a bit more background if possible.

How did you notice that this was a problem?

Can you give me a recipe for making a duplicate upload? So I can try it here.

If so then we could make a test for it too.

Thank you

wloot:
> I'm slightly concerned about changing the behaviour when using backends which don't support modtimes or checksums.

The Equal() is already being used elsewhere(iirc such as rclone copy), so it shouldn't be a problem.

> I'd like a bit more background if possible.
>
> How did you notice that this was a problem?

I use rclone mount with a google drive backend. When I write the same content to a file as itself, a new version of the file will be generated on drive.google.com if without the patch.

```
# mount fs
rclone mount gd:/ /gd
# create a new file
echo "content" > /gd/test
# without the patch, a new version of /gd/test generated
# with the patch, only modtime is updated
echo "content" > /gd/test
```

ncw:
> > I'm slightly concerned about changing the behaviour when using backends which don't support modtimes or checksums.
> 
> The Equal() is already being used elsewhere(iirc such as rclone copy), so it shouldn't be a problem.

Can you try this test with your patch?

Run

```
mkdir /tmp/ftp-files
rclone serve ftp /tmp/ftp-files --user test --pass test
```

and

```
rclone mount --vfs-cache-mode full -vv  ":ftp,host=localhost,port=2121,user=test,pass=6V5Bl1rGjBXdBzAbnHP2FBMOH08:" /mnt/data/
```

And then try overwriting a file on the FTP server with a file that is the same size

```
$ echo one > /mnt/data/file.txt
$ sleep 6
$ cat /tmp/ftp-files/file.txt 
one
$ echo two > /mnt/data/file.txt
$ sleep 6
$ cat /tmp/ftp-files/file.txt 
two
$ 
```

If you do this with your patch the file won't get uploaded. This is because FTP remotes don't support modification times.

> I use rclone mount with a google drive backend. When I write the same content to a file as itself, a new revision of the file will be generated on drive.google.com if without the patch.
> 
> ```
> # mount fs
> rclone mount gd:/ /gd
> # create a new file
> echo "content" > /gd/test
> # without the patch, a new revision of /gd/test generated
> # with the patch, only modtime is updated
> echo "content" > /gd/test
> ```

I see.

So what is happening is that Equal is saying that the modtimes are different, then doing remote and local checksums and seeing they are identical and not uploading.

Perhaps what we should be doing is calling

https://github.com/rclone/rclone/blob/2bc2546d5c93d8a34361e546279e991986090c3f/fs/operations/operations.go#L55

And if `equal && ht != hash.None` then skip the transfer and set the ModTime of the destination instead.

This is slightly annoying duplication of the logic in `Equal` - however `Equal` is designed for syncing which is less demanding than mounting in a lot of ways.

Whatever logic we decide on it needs a test :-)

Another thought crosses my mind - how common is it to alter a file and leave it the same? This patch makes us checksum every file we are about to update which potentially slows everything down.

wloot:
> Another thought crosses my mind - how common is it to alter a file and leave it the same? This patch makes us checksum every file we are about to update which potentially slows everything down.

Prehaps we only should check hash if `--checksum` is set, and otherwise use modtime-size check.
Do you think it is a good idea to modify Equal() itself's logic.

wloot:
> Perhaps what we should be doing is calling
> 
> https://github.com/rclone/rclone/blob/2bc2546d5c93d8a34361e546279e991986090c3f/fs/operations/operations.go#L55
> 
> And if `equal && ht != hash.None` then skip the transfer and set the ModTime of the destination instead.

https://github.com/rclone/rclone/blob/f1f923a986ca193f13ec55dfef31c92f66dd6369/fs/operations/operations.go#L226

It seems the case that remote does not support check sum is checked in equal().

> 
> This is slightly annoying duplication of the logic in `Equal` - however `Equal` is designed for syncing which is less demanding than mounting in a lot of ways.
> 
> Whatever logic we decide on it needs a test :-)
> 
> Another thought crosses my mind - how common is it to alter a file and leave it the same? This patch makes us checksum every file we are about to update which potentially slows everything down.

Perhaps check modtime & size is enough and only do hash check when `--checksum` is set. And user who do not like this can set `--ignore-times` so that we skip all check.

wloot:
> Another thought crosses my mind - how common is it to alter a file and leave it the same? This patch makes us checksum every file we are about to update which potentially slows everything down.

This is actually common in my use case, emby writes same content to metadata file within almost every check cycle, in the meantime there are thousands of metadata files.

ivandeex:
This patch fixes a corner case and has potential to break stuff for a lot of existing cases/
I think we'd better reject this proposal for safety sake and postpone solution to
- https://github.com/rclone/rclone/issues/4810

ivandeex:
If this ever gets accepted,
this must use a dedicated vfs flag like `--vfs-no-upload-if-same-data`
and a have a clear explanation of a warranting user case in the vfs docs

ivandeex:
Captured the requested behavior as a new user case in https://github.com/rclone/rclone/issues/4810#issuecomment-972987493

ivandeex:
In its current form, this PR cannot be accepted.
I wrote the requirements above.

P.S. Generally, the change detection logic in rclone has come to the verge of unsupportable. It requires deeper overall discussion rather than the stream of small per-user-case optimizations.

PR #5578, opened 2021-09-06
docs/jottacloud: update description of whitelabel services

albertony:

Fix mentions of obsolete whitelabel services in docs:
- `Com Hem Cloud` is now `Tele2 Cloud` (the company Com Hem was discontinued as a brand as of April 2021, after being merged into Tele2 in 2018–2019).
- `Get Sky` is now `Telia Sky` (my previous Get Sky account was rebranded to Telia Sky mid 2020, Telia company bought the Get company some time before that so if not all accounts are rebranded I guess they will be soon).

Added list of known whitelabel services to Jottacloud docs:
- Not 100% decided if we should do this, maintain such an unpublished list in rclone docs.
- The reason to include it is that many (most?) users of such services may not know that it is a Jottacloud based service, and that they can use rclone with Jottacloud backend to access it.
- So, I guess my suggestion is to include it. What do you think?

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,6 @@ Rclone *("rsync for cloud storage")* is a command line program to sync files and
   * Dropbox [:page_facing_up:](https://rclone.org/dropbox/)
   * Enterprise File Fabric [:page_facing_up:](https://rclone.org/filefabric/)
   * FTP [:page_facing_up:](https://rclone.org/ftp/)
-  * GetSky [:page_facing_up:](https://rclone.org/jottacloud/)
   * Google Cloud Storage [:page_facing_up:](https://rclone.org/googlecloudstorage/)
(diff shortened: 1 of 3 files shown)


ivandeex:
If they split, spawn and merge so actively.. Probably we might have a dedicated wiki page and link it from main docs?

albertony:
That is another option, sure.

My impression is that it is more of a coincidence with multiple changes recently. Also, some (Telia) are even referenced in code/config with its dedicated auth, so would probably still be mentioned by name in docs. Hmm.. 🤔

ncw (approved):
It is very helpful for users to name the providers in the page and will also make them appear in search engines so I think this is a great change.

In general providers don't change too quickly so I don't think keeping them here will be too much of a burden.

I'll merge this now.

PR #5585, opened 2021-09-07
Add precision debug line to local fs

hyperknot:

Add a single debug line to help debugging local fs precision issues.

The final line looks like this:
```
2021/09/07 20:11:13 DEBUG : Local file system at /Volumes/Blue/temp: precision detected: 1ns
```

https://forum.rclone.org/t/bug-without-modify-window-1s-sync-to-always-copies-all-files/26394/22

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -601,6 +601,9 @@ func (f *Fs) Precision() (precision time.Duration) {
 	f.precisionOk.Do(func() {
 		f.precision = f.readPrecision()
 	})
+
+	fs.Debugf(f, "precision detected: %v", f.precision)
+
 	return f.precision
 }
 


ivandeex:
`rclone backend features /path/to/local` will print the precision so this debugging is not needed

hyperknot:
> `rclone backend features /path/to/local` will print the precision so this debugging is not needed

In that case a similar line in fs.go / GetModifyWindow would be very helpful, to understand what modifyWindows is calculated and why.

hyperknot:
Also a +1 case for adding it to debug log is that it seems that the local fs have been wrongly calculating precision for 9 years.
I mean if it's in the debug log than it'd be obviously visible to anyone ever sharing or reading a debug log.

ncw:
Perhaps the best solution might be to make sure it is in the DEBUG messages for file transfers. At the moment we have

    2021/09/09 08:26:09 DEBUG : 1: Modification times differ by -33.95628693s: 2021-09-09 08:26:08.072165846 +0100 BST, 2021-09-09 08:25:34.115878916 +0100 BST
    2021/09/09 08:26:09 DEBUG : 2: Size and modification time the same (differ by 0s, within tolerance 1ns)

That `tolerance` is the combined modify-window. This should probably be in the first log also.

Maybe we should change the word `tolerance` to `modify window` which is the user facing words we use (internally we call this precision). So these might become

    2021/09/09 08:26:09 DEBUG : 1: Modification times differ by -33.95628693s, outside modify window 1ns: 2021-09-09 08:26:08.072165846 +0100 BST, 2021-09-09 08:25:34.115878916 +0100 BST
    2021/09/09 08:26:09 DEBUG : 2: Size and modification time the same (differ by 0s, within modify window 1ns)

Would that have helped debug the issue @hyperknot ?

hyperknot:
@ncw, interestingly on my log I don't see the tolerance displayed at all. Seeing the 1ns tolerance displayed would have helped a lot!

non-crypt log:
https://forum.rclone.org/t/bug-without-modify-window-1s-sync-to-always-copies-all-files/26394/11?u=hyperknot

crypt log:
https://forum.rclone.org/t/bug-without-modify-window-1s-sync-to-always-copies-all-files/26394/6?u=hyperknot

ncw:
@hyperknot wrote:
> interestingly on my log I don't see the tolerance displayed at all. Seeing the 1ns tolerance displayed would have helped a lot!

I think that is because you didn't have any matches.

Do you want to adjust your PR to make this change with the wording I've suggested above?

I think only these lines need changing:

https://github.com/rclone/rclone/blob/2bc2546d5c93d8a34361e546279e991986090c3f/fs/operations/operations.go#L213

https://github.com/rclone/rclone/blob/2bc2546d5c93d8a34361e546279e991986090c3f/fs/operations/operations.go#L217

hyperknot:
I'm happy to change the wordings but that wouldn't answer the key problem of me not seeing any of the values. I think that'd be the most important.

```
rclone copy --config='' /etc/hosts /Volumes/microsd/ -vv
2021/09/09 19:39:17 DEBUG : rclone: Version "v1.56.0" starting with parameters ["rclone" "copy" "--config=" "/etc/hosts" "/Volumes/microsd/" "-vv"]
2021/09/09 19:39:17 DEBUG : Creating backend with remote "/etc/hosts"
2021/09/09 19:39:17 DEBUG : Using config file from ""
2021/09/09 19:39:17 DEBUG : fs cache: adding new entry for parent of "/etc/hosts", "/etc"
2021/09/09 19:39:17 DEBUG : Creating backend with remote "/Volumes/microsd/"
2021/09/09 19:39:17 DEBUG : hosts: Modification times differ by -956.462099ms: 2021-09-07 16:31:26.956462099 +0200 CEST, 2021-09-07 16:31:26 +0200 CEST
2021/09/09 19:39:17 DEBUG : hosts: md5 = e37634b906b4608c696b01434061bfd3 OK
2021/09/09 19:39:17 INFO  : hosts: Updated modification time in destination
2021/09/09 19:39:17 DEBUG : hosts: Unchanged skipping
2021/09/09 19:39:17 INFO  :
Transferred:   	          0 / 0 Byte, -, 0 Byte/s, ETA -
Elapsed time:         0.0s

2021/09/09 19:39:17 DEBUG : 4 go routines active
```

ncw:
> 2021/09/09 08:26:09 DEBUG : 1: Modification times differ by -33.95628693s, outside modify window 1ns: 2021-09-09 08:26:08.072165846 +0100 BST, 2021-09-09 08:25:34.115878916 +0100 BST
> 2021/09/09 08:26:09 DEBUG : 2: Size and modification time the same (differ by 0s, within modify window 1ns)

Both of these logs now contain the modify window which should be what you need for debugging?

hyperknot:
> now contain

By "now" what do you refer to? Did you make a PR for this or updated master since my logs?

ivandeex:

PR #5586, opened 2021-09-08
Tweak the email

abhi18av:
Hello `rclone` team,

This is a tiny PR updating the email.


ivandeex:
Merging

ncw:
Note that email addresses in git log will get added back into the changelog - see the top of that file

https://github.com/rclone/rclone/blob/master/docs/content/authors.md#contributors

> email addresses removed from here need to be addeed to bin/.ignore-emails to make sure update-authors.py doesn't immediately put them back in again.

So you can either add the bad email to `bin/.ignore-emails` or add them both after your name.

PR #5587, opened 2021-09-09
Add hasher backend

### LINKED ISSUE #157: md5sum file support

### LINKED ISSUE #626: Cache MD5s
For large syncs, calculating the MD5s of all the files can take many hours, even though only a few GB has changed. A feature to cache the MD5 based on size+modified time would be very helpful.

### LINKED ISSUE #949: fs/local: implement hashsum cache
I think a nice feature would be the ability to generate a cache of checksums for local files, which could be used to accelerate subsequent calls to rclone. This would make the following sequence of operations much faster, by avoiding redundant checksum computations on the local side:

> **rclone check --checksum cloud:/path /local/path**
> [user thinks about why the files might be different, and decides what to do]
> **rclone copy some files**
> [repeat as necessary]
> **rclone check --checksum cloud:/path /local/path**
> [see where we stand now]

I'm thinking that this might be accomplished with a single command-line argument..something like --local-file-checksum-cache=path/to/database/file. If the file is missing, it's created, and in either case is written out when rclone terminates. The checksum DB would cache the file's size and mod-time along with the checksum, and when the checksum is needed for a local file, the cached one is used if the size and mod-time match.

Note that the cache file would potentially need to cache both MD5 and SHA1 checksums, and there should also be a way to disable reading the cache (e.g., if I believe my filesystem may have been damaged, and want to re-download pristine versions of any corrupted files, then I don't want to trust the local checksum cache).

See also related issues #725, #626 and #157.

#### How to use GitHub

ivandeex:
## What is the purpose of this change?

This PR adds new `hasher` backend after design proposal in https://github.com/rclone/rclone/issues/949#issuecomment-849122505
The actual design is described below

### Goals

1. fix slow hashing of large files (localfs, ftp, sftp)
2. emulate hash types unimplemented by backends.
3. warm up cache from external SUM files to accelerate even more.

### Configuration

The backend would take the following parameters:

- `remote` (like `alias`, required)
- `hash_names` - comma/blank separated list, by default `md5, sha1`.
- `min-size` - hashes will be stored only for objects above this size (default: 1K)
- `max-age` - maximum time to keep a given checksum, e.g. `1h30m` or `30d`. `0` will disable caching, `off` will cache forever. (default: off)
- `cache-dir` (default: `~/.cache/rclone/hasher/CleanedRemote`)

### Object.Hash

Algorithm of the new backends's `Object.Hash` is like this:

1. if requested hash is supported by lower level, just pass it.
2. if object size is below `min-size` then download object and calculate _requested_ hashes on the fly.
3. if unsupported and the size is big enough, build object `fingerprint` including size, modtime (if supported), _other_ first-found hash if any. Note: fingerprint API does not support _inode number_.
4. if the strict match is found in cache for the requested remote, return the stored hash.
5. if remote found but fingerprint mismatched, then: purge the entry and proceed to step 6.
5. if remote not found or had no requested hash type or after step 5: download object, calculate all _supported_ hashes on the fly and store in cache; return requested hash.

### Fs API

It handles:

- copy/move within backend controlled remote (update cache entry key)
- object.Remove (remove cache entry)
- whenever object is uploaded or downloaded in full, capture the stream to calculate all supported hashes on the fly and update workdir
- periodically prune workdir entries (if `max-age` is not `off`) - this should help against thrashing workdir if a file was removed externally or via lower level backend

### Storage

Cached checksums are stored as `bolt` database files, one per base backend, under `~/.cache/rclone/hasher/`.
You can print or drop database:

```
rclone backend dump Hasher:dir/subdir
rclone backend drop Hasher:
```

### Warming up

Hasher will update cached file checksums when the file is overwritten or file is fully read like
```
rclone cat hasher:path/to/file > /dev/null
```

So the best way to refresh cached caches on a subtree of files is to re-download all file in the subtree, for example:
```
rclone md5sum Hasher:path/to/subtree --download >/dev/null
```

### Warming up with a SUM file

```
rclone backend import hasher:path/to/data sha1 remote:/path/to/sum.sha1
```

Instead of SHA1 it can be any hash supported by the remote. The last argument can point to either a local or an `other-remote:path` text file in SUM format.

The `import` command will parse SUM file (parser is developed by #5352), then walk down the path given by the 1st argument, snapshot current fingerprints and fill in the cache entries correspondingly. This will not check that supplied values are correct. Note that this is a one-time action. The SUM files will not get "attached" to the remote. Cache entries can still be overwritten later, should the object's fingerprint change.

```
rclone backend stickyimport hasher:path/to/data sha1 remote:/path/to/sum.sha1
```

`stickyimport` is similar to `import` but works much faster because it skips initial walk. Instead of cache entries marked by fingerprint it creates "sticky" entries that are bound to a file name only ignoring size or modification time. Such hash entries can be replaced only by `backend drop` or by full re-read of relevant files.

Fixes #949
Fixes #157
Fixes #626

#### Checklist


ivandeex:
The failed code quality test is irrelevant

ivandeex:
1. Documentation not done yet
2. Generic backend integration test passes
3. Unit tests for hasher specific features not done yet
4. The failed code quality test is irrelevant

cc @ncw

ncw:
This is a great implementation @ivandeex. I was imagining that we'd patch all the backends but this is much better.

I wonder whether we should support easy integration of this as nested remotes are a bit of a pain to set up... I had the idea that wrapping backends should support doing `:hasher:remote:mypath`. This is easy to do, if `remote` is empty in `NewFs` then use `root` instead. I made a zip and a tar backend using this and it worked quite well. What do you think?

> 1. Documentation's not in the patch yet.
>    After a code review I'll slightly edit the top post here and add to the commit.

I'll give it a quick once over now.

> 2. Generic backend integration test passed successfully.
> 3. Unit tests for hasher specific features not done yet. Probably in the next PR.
> 4. Code quality test is addressed by [build: apply gofmt from golang 1.17 #5588](https://github.com/rclone/rclone/pull/5588)
>    (will rebase when that's merged)

Reviewed :-)

> 5. Betas to play with are published as `iva` releases at https://github.com/ivandeex/rclone/releases

I'd like to get all devs on the same page with releasing test binaries to users. At the moment if you push to a branch on the main repo you'll get betas built etc. What I'd like to do is make it so that if a team member makes a PR then this also makes a beta - what do you think of that idea? Ideally we'd make some actions bot which posted a link in the PR once it was built.

ivandeex:
> I wonder whether we should support easy integration of this as nested remotes are a bit of a pain to set up...

:+1:  :+1:  :+1: 

> I had the idea that wrapping backends should support doing `:hasher:remote:mypath`. This is easy to do, if `remote` is empty in `NewFs` then use `root` instead.

:+1:  An API like `fs.NewFsChain(":wrapper,wrap_opt1=val1:remote,inner_opt1=val2:mypath")` for
- building an FS chain from extended remote path and
- returning the top-most Fs from the chain

seems reasonable and straightforward in the case when wrapping remote has empty root relative to the inner remote (we could even stuff the chain builder right into good old `fs.NewFs` ???)

:+1: FS chains with empty wrapper root should be straightforward on the command line too.

:-1:  I can't imagine INI syntax for FS chains even with empty wrapper root. Probably needs TOML, not INI ???

:-1:  I can't imagine what would extended remote path be for wrappers with non-empty root like `:wrapper,opt1=val1[[[wrapper-root-goes-somewhere-here]]]:remote,opt2=val2:/path/to/remote/root`

> I'll give it a quick once over now.
> [5min]
> Reviewed :-)

I will add docs tomorrow. I'm not lightning fast like you :)

> I'd like to get all devs on the same page with releasing test binaries to users. At the moment if you push to a branch on the main repo you'll get betas built etc.

Ohhh! I'd be happy to make a beta branch for bisync!

> What I'd like to do is make it so that if a team member makes a PR then this also makes a beta - what do you think of that idea? Ideally we'd make some actions bot which posted a link in the PR once it was built.

I don't know. I just can test out whatever you make.

ivandeex:
@ncw Please verify whether this patch fixes #157

backend/hasher/hasher.go
```diff
@@ -0,0 +1,550 @@
...
+			Name:     "remote",
+			Required: true,
+			Help:     "Remote to cache hashes for, like myremote:mypath",
+		}, {
+			Name:     "hash_names",
+			Default:  "md5",
+			Advanced: false,
+			Help:     "Space/comma separated list of checksum types to support.",
+		}, {
+			Name:     "auto_size",
```
ncw, line 44: Does this mean that we download files < 1k automatically if they have no checksum?

I think the default should probably be no downloading.

backend/hasher/hasher.go
```diff
@@ -0,0 +1,550 @@
...
+		}, {
+			Name:     "max_age",
+			Advanced: false,
+			Default:  fs.DurationOff,
+			Help:     "Maximum time to keep checksums in cache. 0 = no-cache, off = forever",
+		}, {
+			Name:     "lock_time",
+			Advanced: true,
+			Default:  fs.Duration(time.Second),
+			Help:     "Maximum time interval to keep cache database locked by process",
```
ncw, line 57: I see - this is to help with multi-process access. I wonder how well it works in practice?

backend/hasher/hasher.go
```diff
@@ -0,0 +1,550 @@
...
+		Description: "Hasher will cache hashes",
+		NewFs:       NewFs,
+		CommandHelp: commandHelp,
+		Options: []fs.Option{{
+			Name:     "remote",
+			Required: true,
+			Help:     "Remote to cache hashes for, like myremote:mypath",
+		}, {
+			Name:     "hash_names",
+			Default:  "md5",
```
ncw, line 40: You could use fs.CommaSepList or fs.SpaceSepList here if you wanted...

backend/hasher/hasher_test.go
```diff
@@ -0,0 +1,39 @@
...
+// TestIntegration runs integration tests against the remote
+func TestIntegration(t *testing.T) {
+	opt := fstests.Opt{
+		RemoteName: *fstest.RemoteName,
+		NilObject:  (*hasher.Object)(nil),
+		UnimplementableFsMethods: []string{
+			"OpenWriterAt",
+		},
+		UnimplementableObjectMethods: []string{
+			"MimeType",
```
ncw, line 27: ? I would have thought you could just pass this through?

backend/hasher/object.go
```diff
@@ -0,0 +1,234 @@
...
+func (r *hashingReader) Close() error {
+	if rc, ok := r.rd.(io.ReadCloser); ok {
+		return rc.Close()
+	}
+	return nil
+}
+
+// Return object fingerprint or empty string in case of errors
+func (o *Object) fingerprint(ctx context.Context) string {
+	size := o.Object.Size()
```
ncw, line 217: Just wondering why you aren't using `fs.Fingerprint` with `fast=true`?

docs/content/hasher.md
```diff
@@ -0,0 +1,188 @@
+---
+title: "Hasher"
+description: "Handle checksums for other remotes"
+---
+
+# {{< icon "fa fa-check-double" >}}Hasher (EXPERIMENTAL)
+
+The `hasher` overlay caches hashsums for underlying remote.
```
ncw, line 8: Should probably give an example here

> The `hasher` overlay caches hashsums (such as MD5, SHA-1) for underlying remotes.

docs/content/hasher.md
```diff
@@ -0,0 +1,188 @@
...
+
+To use it, first set up the underlying remote following the configuration
+instructions for that remote. You can also use a local pathname instead of
+a remote.
```
ncw, line 12: Could do with a paragraph on why you would use hasher.

docs/content/hasher.md
```diff
@@ -0,0 +1,188 @@
...
+
+Many storage backends support MD5 and SHA1 hash types, so does chunker.
+With chunker you can choose one or another but not both.
+MD5 is set by default as the most supported type.
+Since chunker keeps hashes for composite files and falls back to the
+wrapped remote hash for non-chunked ones, we advise you to choose the same
+hash type as supported by wrapped remote so that your file listings
+look coherent.
+
+If your storage backend does not support MD5 or SHA1 but you need consistent
```
ncw, line 79: Is this bit of the doc out of date?

ncw (commented):
I've given this a quick once over - looking great!

I'd like to have a discussion about the use of bolt? I'd like to only have one KV database in rclone (once we have disabled the cache backend). Is bolt the right one?

backend/hasher/hasher.go
ivandeex: ok, will change default to 0

docs/content/hasher.md
ivandeex: i haven't yet worked on docs.

backend/hasher/object.go
ivandeex: I needed an ad-hoc fingerprint because instead of `baseFs.Hashes().GetOne()` I want `baseFs.HashesNotInHasherCache().GetOne()`

ncw:
@ivandeex wrote:

> > I wonder whether we should support easy integration of this as nested remotes are a bit of a pain to set up...
> 
>   
> 
> > I had the idea that wrapping backends should support doing `:hasher:remote:mypath`. This is easy to do, if `remote` is empty in `NewFs` then use `root` instead.
> 
>  An API like `fs.NewFsChain(":wrapper,wrap_opt1=val1:remote,inner_opt1=val2:mypath")` for
> 
> * building an FS chain from _extended remote path_ and
> * returning the top-most Fs from the new chain
> 
> seems reasonable and straightforward in the case when wrapping remote has empty root relative to the inner remote (we could even stuff the chain builder right into good old `fs.NewFs` ???)

I was thinking that if you called `wrapper.NewFs` with `remote=""` it would say, ah no remote, lets use `root` instead.

This would enable this syntax to just work...

    :wrapper,wrap_opt1=val1:remote,inner_opt1=val2:mypath

`wrapper.NewFs` would be called with `opt={wrap_opt1=val1}`, root=`remote,inner_opt1=val2:mypath`

It would then decide that since its config `remote` was empty to set `remote = remote,inner_opt1=val2:mypath` and `root=""` and then call `remote.NewFs` with `inner_opt` and root=`mypath`.

So I don't think a new API is needed as long as we are happy with that syntax.
 
>  FS chains with empty wrapper root should be straightforward on the command line too.
> 
>  I can't imagine INI syntax for FS chains even with empty wrapper root. Probably needs TOML, not INI ???
> 
>  I can't imagine what would extended remote path be for wrappers with _non-empty root_ like `:wrapper,opt1=val1[[[wrapper-root-goes-somewhere-here]]]:remote,opt2=val2:/path/to/remote/root`

Yes that is a potential problem. So we are saying that if we had a `hasher` set up on `/backup` how do we access a path `/backup/current`?

What would hasher do if it you set up a `hasher` for `/backup` and then tried to use `/backup/current`. Is that effectively a new hasher? The VFS cache is designed so that you can do that, but it is quite a heavy load to impose on all backends.

Here are some ideas

    :wrapper,opt1=val1,remote='remote,opt2=val2:/backup':current  # this would work now
    :wrapper,opt1=val1,root=current:remote,opt2=val2:/backup         # use root as a parameter as well?
    :wrapper,opt1=val1:remote,opt2=val2:/backup//current                # parse off wrapper root from end?

> > I'd like to get all devs on the same page with releasing test binaries to users. At the moment if you push to a branch on the main repo you'll get betas built etc.
> 
> I'd be happy to make a beta branch for bisync!
> Upd: https://github.com/rclone/rclone/tree/pr-bisync

Great :-)
 
> > What I'd like to do is make it so that if a team member makes a PR then this also makes a beta - what do you think of that idea? Ideally we'd make some actions bot which posted a link in the PR once it was built.
> 
> I don't know. I just can test out whatever you make.

I figured out how to do everything except work out if the PR is from a team member, but I'll carry on!

backend/hasher/hasher.go
ivandeex: During tests I ran 2-3 rclone instances doing `rclone sha1sum /some/local/dir123` with `-vv`, touched random files under test dir  and watched them reopen bolt periodically due to changed fingerprint and following cache updates.

As a stress test for db locking I ran dump (I decided that this maintenance operation can lock db as a whole) in parallel with above operations and watched low-level retries.

I still wonder how I can convert it into backend unit tests.

In practice I usually have at most 1 concurrent rclone instance doing checksum ops and multiple rclone mounts that don't need checksums.

backend/hasher/hasher.go
ivandeex: For user convenience I wanted to split "comma optionally surrounded spaces or just by a space" which is easy on my eyes IMHO. Strictly speaking the regexp for such delimiter looks `\s*[,\s]\s*`. I simplified it to `[,\s]+` so splitting is a one-liner. This way a string like `a,b,,,c` or `a, b   ,c` will be split into slice `[a b c]`. I'm fine with such relaxed syntax.

If you wish, I can use your splitters resulting in more stringent syntax (where delimiter is exactly one comma or 1+ spaces but no comma). I don't care, just tell me

docs/content/hasher.md
ivandeex: ok

docs/content/hasher.md
ivandeex: ok.

docs/content/hasher.md
ivandeex: I noted above that docs are not yet ready. What you see is a stub (I guess I copy-pasted it from chunker). I'll let you know when I do the docs.

backend/hasher/object.go
ivandeex: I needed an ad-hoc fingerprint because instead of `baseFs.Hashes().GetOne()` I want `baseFs.HashesNotInHasherCache().GetOne()` (and of course none if base hashes are slow).
fs.Fingerprint can't do that.

backend/hasher/hasher_test.go
ivandeex: I hit import loop when I tried to pass it.

backend/hasher/bolt.go
```diff
@@ -0,0 +1,371 @@
...
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/config"
+	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/fs/operations"
+	"github.com/rclone/rclone/lib/encoder"
+	"go.etcd.io/bbolt"
```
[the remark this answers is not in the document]
ivandeex: >  I was worried about the project being maintained.

Bolt is heavily used by etcd which is in turn used by k8s which is a mature industry standard. I don't think they can deprecate it easily.

> can't be open from multiple processes at once which can be quite annoying

Bolt can be shared if everyone just want to read.
If someone wants to write, others should wait.
So each instance should handle locking and switch access mode.

Shared SQL databases (postgres, oracle) solve locking problem by using client-server pattern.
I solved it by using a serving goroutine. It handles locking and switches access mode to guarantee other instances a time slot. Other parts of rclone access database via API provided by the goroutine (I made an exception for the dump operation). I reduced core API to one function `do` and one `Op` interface. Everything database-related extends these two.

I considered a few alternatives
- use external KV server (e.g. redis is widespread and has very small go client)
- write a copy of VFS layer (fine grained per-file access but too much work)
- factor parts of VFS code (but it's a moving target)
- write my own KV store with locking (too much work)
- write my own KV store without locking and extend RCD REST protocol (or implement minimal redis server) so one rclone can serve KV to others (but I don't like revolutions and paradigm shifts)
- use bolt (it has transactions and locking, KV access needs some kind of locking anyway, transactions good though optional, so I chose bolt)

I don't see a single best choice. I chose something satisfactory.
In the end, I don't worry much. It's a cache. We can break and completely replace its implementation. In the worst case users will not loose data, just experience a delay.

A year ago I thought about metadata store for rclone. It's a bigger problem. It's not a cache, it can't loose data. It can't be recreated on every client machine. It needs either (a) a central store e.g. redis for all rclone instances or (b) use backends as golden sources of their metadata. (a) is simpler to impelement. I tried to tackle (b) i.e. sketch a KV storage engine that can use backends without modtime or locking for a consistent storage. Please take a look at [my design sketch](https://github.com/orgs/rclone/teams/rclone/discussions/1). It's a lot of work.

ivandeex:
> I'd like to only have one KV database in rclone (once we have disabled the cache backend).

Currently only hasher plans to use it.
We could however make it a common API:
- factor `hasher/bolt.go` to `lib/kv/bolt.go`
- rename interface `hashOp` to `dbOp`
- make `--hasher-lock-time` a global option `--kv-lock-time`
- change database location from `~/.cache/rclone/hasher/REMOTE` to `~/.cache/rclone/kv/REMOTE`
- run a central pool of database goroutines per rclone instance (one goroutine per database)
- etc

> Is bolt the right one?

bolt is a satisfactory one.

ivandeex:
> I'd like to only have one KV database in rclone (once we have disabled the cache backend).

Currently only hasher plans to use it.
Probably the resumer framework can pretend to have a kind of resumption store.

We could however make it a common API:
- factor `hasher/bolt.go` to `lib/kv/bolt.go`
- rename interface `hashOp` to `kvOp`
- make `--hasher-lock-time` a global option `--kv-lock-time`
- change database location from `~/.cache/rclone/hasher/REMOTE` to `~/.cache/rclone/kv/REMOTE`
- run a central pool of database goroutines per rclone instance (one goroutine per database)
- slowly develop new kinds of `kvOp` to satisfy more internal consumers
- etc

> Is bolt the right one?

bolt is a satisfactory one.

ivandeex:
> What would hasher do if it you set up a `hasher` for `/backup` and then tried to use `/backup/current`.

Hasher will jump in if you setup `Hasher{remote=/backup}` and use the path `Hasher:current`.

> Is that effectively a new hasher?

No. `/backup/current` is a short for `local:/backup/current` handled by the `local` backend.
Hasher does not install hooks in the local backend. It reacts to its own paths.

> The VFS cache is designed so that you can do that, but it is quite a heavy load to impose on all backends.

VFS jumps in when you `mount` VFS controlled stuff somewhere and access the mount point.
In fact you install a hook in the kernel and let it trigger.
rclone mounts do not use file checksums so far so Hasher is irrelevant here (maybe _in future_ mounts will return checksums as extended file attributes).

ivandeex:
Currently wrapping rclone backends only provide means for wrapping other remotes. You are talking about **"injecting"** (or mixing in) behaviors of transient backends into other remotes. I didn't set such a goal. rclone lacks internal APIs for such mixins. It's out of scope for this PR.

ivandeex:
> Here are some ideas

```
:wrapper,opt1=val1,remote='remote,opt2=val2:/backup':current  # this would work now
:wrapper,opt1=val1,root=current:remote,opt2=val2:/backup         # use root as a parameter as well?
:wrapper,opt1=val1:remote,opt2=val2:/backup//current                # parse off wrapper root from end?
```

I planned to use September for publishing a number of prepared patches: ftp, autofs mounts, probably bisync.
Let me experiment with your sketch for a month. Maybe I can deliver a PR with new _rpath parser_ in 2-3 months.

ivandeex:
> Here are some ideas

```
:wrapper,opt1=val1,remote='remote,opt2=val2:/backup':current  # this would work now
:wrapper,opt1=val1,root=current:remote,opt2=val2:/backup         # use root as a parameter as well?
:wrapper,opt1=val1:remote,opt2=val2:/backup//current                # parse off wrapper root from end?
```

@ncw
I planned to use September for publishing a number of prepared patches: ftp, autofs mounts, probably bisync.
Let me experiment with your sketch for a month. Maybe I can deliver a PR with new _rpath parser_ in 2-3 months.

backend/hasher/bolt.go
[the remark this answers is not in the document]
ivandeex: P.S. A year ago I thought about _metadata store_ for rclone. It's a bigger problem. It's not cache - it can't lose data. It can't be recreated on every client machine.
It needs either (a) a central store e.g. redis for all rclone instances or (b) use backends as golden sources of their metadata.
(a) is simpler to impelement. I tried to tackle (b), sketch a KV storage engine that can use backends _without modtime or locking for consistent_ storage. Please take a look at [my design sketch](https://github.com/orgs/rclone/teams/rclone/discussions/1).

ivandeex:
> I was imagining that we'd patch all the backends but this is much better.

Probably adding optional `hash_names` parameter to every backend and wrapping `backend.Object.Hash` is a viable alternative......

backend/hasher/hasher.go
```diff
@@ -0,0 +1,550 @@
...
+			return err
+		}
+		return callback(hashEntries)
+	})
+}
+
+// Purge a directory
+func (f *Fs) Purge(ctx context.Context, dir string) error {
+	if do := f.Fs.Features().Purge; do != nil {
+		return do(ctx, dir)
```
ivandeex, line 259: I forgot to purge cache entries here!

ivandeex (commented):
:-1:

ncw:
> > I'd like to only have one KV database in rclone (once we have disabled the cache backend).
> 
> Currently only hasher plans to use it.
> We could however make it a common API:
> 
> * factor `hasher/bolt.go` to `lib/kv/bolt.go`
> * rename interface `hashOp` to `kvOp`
> * make `--hasher-lock-time` a global option `--kv-lock-time`
> * change database location from `~/.cache/rclone/hasher/REMOTE` to `~/.cache/rclone/kv/REMOTE`

This should probably have a separate name for each user  `~/.cache/rclone/kv/hasher/REMOTE` `~/.cache/rclone/kv/VFS/REMOTE` (or something!)

> * run a central pool of database goroutines per rclone instance (one goroutine per database)
> * etc

Bolt is really the right solution for the VFS metadata - when there gets to be a lot of it, scanning lots of little JSON files is very slow. However I didn't want to commit to Bolt at the time and was wary of the multiple process limitations.

Another place I'd like a KV database is doing syncs larger than memory - eg sync two 20 million entry directories - without using 20 GB of RAM.

I think your idea to stick it in `lib` is a good one and probably worth doing now.

> > Is bolt the right one?
> 
> bolt is a satisfactory one.

I was looking also at [badger](https://github.com/dgraph-io/badger) however my main concern is longevity and stability rather than absolute performance.

> Currently wrapping(transient) rclone backends only provide means for wrapping other remotes. You are talking about **"injecting"** (or mixing in) behaviors of transient backends into other remotes. I didn't set such a goal. rclone lacks internal APIs for such mixins. It's out of scope for this PR.

Yes lets discuss this later. I just wanted to share my ideas.

> > Here are some ideas
> 
> ```
> :wrapper,opt1=val1,remote='remote,opt2=val2:/backup':current  # this would work now
> :wrapper,opt1=val1,root=current:remote,opt2=val2:/backup         # use root as a parameter as well?
> :wrapper,opt1=val1:remote,opt2=val2:/backup//current                # parse off wrapper root from end?
> ```
> 
> I planned to use September for publishing a number of prepared patches: ftp, autofs mounts, probably bisync.
> Let me experiment with your sketch for a month. Maybe I can deliver a PR with new _rpath parser_ in 2-3 months.

It would be a good idea to finish some of the things we've started. I've got a list of things too sitting in branches here.

ivandeex:
> This should probably have a separate name for each user `~/.cache/rclone/kv/hasher/REMOTE` `~/.cache/rclone/kv/VFS/REMOTE` (or something!)

Alternatively we could use multiple bolt buckets per database, but the more files the better due to fine grained locking.

The "user" approach also calls for:
- extended [API](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L173) like `kv.Do(ctx context.Context, user string, f fs.Fs, op kvOp) error`
- something like a `map[user_string][fs.Fs]func (...)` (pool) of [goroutines](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L182)
- a common function to [clean](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/hasher.go#L175) and [encode](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L86) remote names like `kv.CleanFsName(fs.Fs)`

> Another place I'd like a KV database is doing syncs larger than memory - eg sync two 20 million entry directories - without using 20 GB of RAM.

I also had an idea to use a `struct Trie` (details don't matter now) for _large recursive_ listings to replace long paths by 32-bit integers around the code. Could save some RAM especially in sync operations.

> I was looking also at [badger](https://github.com/dgraph-io/badger) however my main concern is longevity and stability rather than absolute performance.

If we make it an API then "procedural users" will accept anything.
If we decide to switch to another database engine later, the "human users" will notice just a one-time delay due to database regeneration.

> Yes lets discuss this later. I just wanted to share my ideas.

I'll create a dedicated ticket and copy the discussion there.

> It would be a good idea to finish some of the things we've started.

Holy true.

ncw:
> > This should probably have a separate name for each user `~/.cache/rclone/kv/hasher/REMOTE` `~/.cache/rclone/kv/VFS/REMOTE` (or something!)
> 
> Alternatively we could use multiple bolt buckets per database, but the more files the better due to fine grained locking.

I like the idea of one file per "purpose" if you see what I mean.

> The "user" approach also calls for:
> 
> * extended [API](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L173) like `kv.Do(ctx context.Context, user string, f fs.Fs, op kvOp) error`
> * something like a `map[user_string][fs.Fs]func (...)` (pool) of [goroutines](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L182)
> * a common function to [clean](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/hasher.go#L175) and [encode](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L86) remote names like `kv.CleanFsName(fs.Fs)`

There is the stuff in the VFS cache which has had many patches and we should probably start from! This could be factored out into `fs` maybe?

https://github.com/rclone/rclone/blob/b35db61a80a90cb5e2d59966f5e711807d4048fc/vfs/vfscache/cache.go#L78-L100

> > Another place I'd like a KV database is doing syncs larger than memory - eg sync two 20 million entry directories - without using 20 GB of RAM.
> 
> I also had an idea to use a `struct Trie` ([details](https://en.wikipedia.org/wiki/Trie) don't matter a.t.m.) for _large recursive_ listings to replace long paths by 32-bit integers around the code. Could save some RAM especially in sync operations.

Interesting idea.

> > I was looking also at [badger](https://github.com/dgraph-io/badger) however my main concern is longevity and stability rather than absolute performance.
> 
> If we make it an API then "procedural users" will accept anything.
> If we decide to switch to another database engine later, the "human users" will notice just a one-time delay due to database regeneration.

Agreed
 
> > Yes lets discuss this later. I just wanted to share my ideas.
> 
> I'll create a dedicated ticket and copy the discussion there.

:+1: 

> > It would be a good idea to finish some of the things we've started.
> 
> Holy true.

:-)

ivandeex:
> This should probably have a separate name for each user ~/.cache/rclone/kv/hasher/REMOTE ~/.cache/rclone/kv/VFS/REMOTE (or something!)

We can consider various db file naming schemes below ~/.cache/rclone/kv:
- SUBSYS/REMOTE - db files grouped in dirs by subsystem (vfs, hasher, etc)
- REMOTE/SUBSYS - grouped by remote (easier to cleanup when a remote got deprecated)
- REMOTE~SUBSYS - flat scheme with some delimiter, for cleanup using shell globs `*~vfs` or `sftp~*`
   (`~` or wide `:` or wide `/` or whatever prohibited in remote names)
- REMOTE~SUBSYS.bolt - variation of above with explicit db format
   to facilitate experiments with budger or schema migrations
- etc

ivandeex:
> This should probably have a separate name for each user ~/.cache/rclone/kv/hasher/REMOTE ~/.cache/rclone/kv/VFS/REMOTE (or something!)

We can consider various db file naming schemes below ~/.cache/rclone/kv:
- SUBSYS/REMOTE - db files grouped in dirs by subsystem (vfs, hasher, etc)
- REMOTE/SUBSYS - grouped by remote (easier to cleanup when remotes get deprecated)
- REMOTE~SUBSYS - flat scheme with some delimiter, for cleanup using shell globs `*~vfs` or `sftp~*`
   (delimiter can be `~` or wide `:` or wide `/` or whatever prohibited in remote names)
- REMOTE~SUBSYS.bolt - variation of above with explicit db format
   to facilitate experiments with budger or schema migrations
- etc

backend/hasher/hasher.go
ivandeex: or maybe I can add new `fs.CommaSpaceSepList` ?

ivandeex:
> > * a common function to [clean](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/hasher.go#L175) and [encode](https://github.com/rclone/rclone/blob/4709c9c2f2053e5eac8e2f99cbbaf1dc475553d8/backend/hasher/bolt.go#L86) remote names like `kv.CleanFsName(fs.Fs)`
> 
> There is the stuff in the VFS cache which has had many patches and we should probably start from! This could be factored out into `fs` maybe?
> 
> https://github.com/rclone/rclone/blob/b35db61a80a90cb5e2d59966f5e711807d4048fc/vfs/vfscache/cache.go#L78-L100

**Note to myself.** Look what was factored in reply to
https://github.com/rclone/rclone/pull/4547#discussion_r703007850

ivandeex:

iceworks:
Hello. Sorry for interruption. Maybe it would it be possible to add a feature for this hasher backend which would let you always cache local backend files checksum values (for e.g. with --hasher-always-cache-local flag) with single sync/copy operation?

I want that "local->crypt->hasher", "local->crypt->chunker->hasher" or similar backend chain would always make checksum cache (to "./rclone/hasher/local" hasher cache database file) for newly created files in local backend. Also for some reason hasher from "local->crypt->hashercrypt" chain doesn't create hash values of original files for its crypt backend bellow [using v1.56.2-iva18, Windows 10 64-bit]. I must run "rclone hashsum MD5 hashercrypt: --download" to get files checksum values calculated and cached (to "./rclone/hasher/hashercrypt" database). They are supposed to be calculated with copy/sync operations, aren't they? I can do the same thing with "rclone hashsum MD5 hasherlocal:" for crypt encrypted local backend files (local->hasherlocal). That is very inefficient hence I need that new feature so everything could be done with one copy/sync operation in hashercrypt:. And the reason I do that is that I need to upload encrypted files to remote which limits upload speed and needs pre-calculated file checksums. That thing can have very huge effect for upload speed when many upload transfer threads are used. I even could use SD card for that (my SSD is low on space). With HDD you can't have many upload threads (10+) due to its mode of operation. I hope my feature suggestion was understandable.

ivandeex:
@iceworks 
If you read the PR description:
- Hasher will transparently update cache with new checksums when a file is fully read `or overwritten`

you'll see that it works as you want. Or install latest `iva` build from https://github.com/ivandeex/rclone/releases and try beta hasher live.

iceworks:
@ivandeex 
Yes, it always works by reading the files [**_rclone hashsum MD5 ^hasher^: --download_**] however I need that checksums (_for hasher-local: (**most important**) and hasher-localcrypt:_) would be calculated and cached during file creation using "local:->localcrypt:->hasher-localcrypt:" or other similar backend chain and copy operation. So all checksums are calculated and cached into DB file in one operation. I don't want to do additional file re-reading. No checksums are saved this way although it seems it should (hasher-localcrypt: copy). I also tried with empty files pre-created (overwrites these files when copying). Same results. It only updates (makes some very minor changes) cache database file without any new entries. If I do source to hasher-local: copy, the local checksums are calculated and saved correctly to local checksum cache DB. But I need my files encrypted so that doesn't help me.

Now I do hasher-localcrypt: copy and then "_**rclone hashsum MD5 hasher-local: --download"**_ to update local cache DB entries by re-reading all those newly created encrypted files. That is not very fast from SD card. Some files seems to be cached in RAM so they are processed much faster. That is is why I need "--hasher-always-cache-local" flag feature. So local backend files are cached too and instantly ready for upload to that restrictive remote (upload speed limit per file, file should have checksum before upload) and I would be getting significantly higher upload speed.

```
**My exemplary config entries:**
[local]
type = local
nounc = false

[localcrypt]
type = crypt
remote = local:J:\CRYPT
filename_encryption = standard
directory_name_encryption = true
password = *password*
password2 = *password2*

[hasher-local]
type = hasher
remote = J:\CRYPT
hash_names = md5,sha1
max_age = off

[hasher-localcrypt]
type = hasher
remote = localcrypt:
hash_names = md5,sha1
max_age = off
```

ivandeex:
I will retest and change code to ensure that caches are updated automatically for both
`copy hasher-local: hasher-localcrypt` and `copy hasher-localcrypt: hasher-local`
without special flags.

iceworks:
@ivandeex 
> I will retest and change code to ensure that caches are updated automatically for both `copy hasher-local: hasher-localcrypt:` and `copy hasher-localcrypt: hasher-local:` cases without special flags.

Will resulting encrypted local: files checksum values be saved if copy destination is hasher-localcrypt:?
I imagine that if I run `copy hasher-local:D:\BACKUP hasher-localcrypt:BACKUP` then I would get "local" and "localcrypt" checksum databases. "local" database would contain files checksums from "D:\BACKUP" and "J:\CRYPT\~BACKUP~" (resulting encrypted files). And "localcrypt" database would contain original files checksums for files which were copied from "D:\BACKUP". If I would run `copy local:D:\BACKUP hasher-localcrypt:~BACKUP~` then it would be the same thing, except that "local" database won't have "D:\BACKUP" files checksums.

ivandeex:
@iceworks 
Please open a separate ticket where we can discuss your particular patterns of using this yet to be released feature.

ivandeex:
@ncw
I refactored key-value operations into new API in `lib/kv` API, basically:

- `kv.Start(context, facility string, filesystem fs.Fs) (db *kv.DB, err error)` - start a database loop
   facility can be `hasher`, `vfs` and so on
- `DB.Stop()` - dereference a database, completely stop if there are no more references
- `DB.Do(isWriting bool, operation kv.Op) error` - enqueues an operation, returns results when ready
- `kv.Op interface{ func Do(context, bucket kv.Bucket) error }` - interface for key-value operations
- `kv.Bucket interface{ Get, Put, Delete, Seek, ... }` - decouples operations from bolt details and imports
- `kv.Exit()` - reserved for future use with `atexit`

Key-value databases are stored as `~/.cache/rclone/kv/CLEANED_REMOTE~facility.bolt`.
Maximum continuous lock interval per process is configured by global flag `--kv-lock-time` (default: 1s)

Discussion of planned connect string enhancements was outsourced to ticket #5712

@ncw
You requested to implement `Object.MimeType`  by delegating (https://github.com/rclone/rclone/pull/5587#discussion_r706040685).
I did so but the `fstests` test does not like delegation. I don't have enough info to make it pass without your input.
OTOH it's optional and out of scope for this PR. I skip it for now (see dedicated commit).
Please decide on it.

@albertony @ncw
Backend docs are ready at https://github.com/rclone/rclone/blob/pr-hasher/docs/content/hasher.md
Please take a look.

@iceworks 
I think I managed to address your request.
I verified that checksums get created in cache for each copy scenario: `local-crypt -> hasher-local`, `local -> hasher-crypt`, `local -> hasher-local`, `local-crypt -> hasher-crypt`.
If I misunderstood you, please submit a **separate ticket providing config, DUMP of commands, and verbose log**.

Beta builds for evaluation are available at https://beta.rclone.org/branch/pr-hasher/
Windows builds are currently lagging behind due depleted Windows runners on Github side. They'll fix it soon.

ivandeex:
> I didn't want to force push your branch - it needs squashing and re-ordering my commit should go first if you think it is OK.

I think you know it better. Squashing and reording...

ivandeex:
A quick note about last two commits...
After testing concurrent calls of `kv.Start` and `db.Stop` I had to restrict more parts of these under mutex to make results **deterministic** without `Sleep`-s in the test body.
It's more safe but has potential to lock `Start` for **up to 1-2 seconds**. Not fond of it, will seek for ways to improve performance in future.

Sarke:
Hi, first off: great work!

I have one question, not sure if it's an implementation detail or a bug.  Let me know if I should open a new issue or not, since this is still marked experimental.

I am in the process of moving 2M files from a Swift remote to a B2 remote.  The former uses MD5 and the latter uses SHA1, and I got this notice:

```
2021-11-02 06:35:21 NOTICE: hasher::B2UPLOADS:/525: --checksum is in use but the source and destination have no hashes in common; falling back to --size-only
```

This was for the second run to verify, after already doing the transfer.  I assumed, perhaps wrongfully, that because I had specified the two different hashes for the remotes and had done the transfer, that the SHA1 of the target would have been calculated and stored in the BoltDB.  In fact, I am almost certain that those are in there.

So my question is: can I use these two combined checksum as a bridge to verify the transfer?

ivandeex:
@Sarke
This ticket is implemented and closed.
Please post your question as a new ticket.

Issue #5584, opened 2021-09-07
local filesystem's precision it not tested correctly


hyperknot:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/bug-when-using-the-automatic-modify-window-1s-is-not-applied-on-macos-filesystems/26394

#### What is the problem you are having with rclone?
local.go / readPrecision() is testing the wrong filesystem. It always tests the host's tmp folder, which is totally unrelated to the filesystems being used for the operations.

For example on recent macOS, it always returns 1ns as the root fs is always APFS these days. This means that sync operations to external drives will break if those drives are non-APFS, as seen in the forum where crypt always copies everything.

The problem is very clear in the source, the test file is always created in the system's temp folder.

https://github.com/rclone/rclone/blob/2bc2546d5c93d8a34361e546279e991986090c3f/backend/local/local.go#L608

#### What is your rclone version (output from `rclone version`)
master

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
macOS Big Sur

#### Which cloud storage system are you using? (e.g. Google Drive)
local, external USB drive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
Detailed commands in forum.

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
Detailed logs in forum.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

hyperknot:
Added a PR to help debugging this issue:
https://github.com/rclone/rclone/pull/5585

mrx23dot:
Syncing to FAT FS pendrive always checks md5, so mod.time window is not correctly detected by readPrecision()'s temp file.

I have a feeling that writing the temp file and reading it back fast gives the correct mod.time, but reading it back later gives +-1sec difference.

```
2022/02/10 23:40:29 DEBUG : 1265674333_68.jpg: Modification times differ by 1.3928903s: 2012-09-05 13:06:46.6071097 +0100 BST, 2012-09-05 13:06:48 +0100 BST
[...]

PR #5591, opened 2021-09-09
Config paths

albertony:

**1) Add `rclone config paths` command**

New config sub-command `paths` to show the different paths on the running system that rclone might use. Convenient for investigations when there is doubt, and I think it is something that is nice to be open with users about. Other utilities do the same, e.g. my favorite golang application to borrow from: [Syncthing](https://docs.syncthing.net/users/syncthing.html#cmdoption-paths).

```
.\rclone.exe config paths
Config file: C:\Users\me\AppData\Roaming\rclone\rclone.conf
Cache dir:   C:\Users\me\AppData\Local\rclone
Temp dir:    C:\Users\me\AppData\Local\Temp\1
```

**2) Configurable TempDir**

Instead of always using the OS configured temporary directory, allow user to override this (but with same default). Primary use case is running in portable mode. To be considered completely portable, aka "stealth", all paths should be configurable (if not default set to local application directory). E.g. VSCode has portable mode with optional [portable TMP directory](https://code.visualstudio.com/docs/editor/portable#_tmp-directory).
Also: Issue #5584 discusses a case where rclone "secretly" uses the os temp dir without possibility to override.
Although: It is actually possible to override temp dir already, by modifying built-in environment variable TMPDIR/TMP/TEMP, but this is rather dirty, I think, e.g. setting it in a shell session makes any other commands executed in the same session consider it as well.

Decisions:
- Create configured `--temp-dir` if not exists? Verify access to configured `--temp-dir` if exists?
	- Currently 👍  and 👍 .
	- When creating temporary files and directories rclone uses `ioutil.TempDir` and `ioutil.TempFile`, which is same as `os.MkdirTemp` and `os.CreateTemp` in go 1.17, and these fail if parent dir does not exist. Therefore making sure any configured temp-dir exists - on rclone startup.
	- NOTE: Added as a commit on top of initial version which did not do this, in case the decision changes...
- Create/verify OS default the same way if not `--temp-dir`?
	- Currently 👎 .
	- Docs says *the directory is neither guaranteed to exist nor have accessible permissions* so in principle it should be just as relevant as with custom dir, but in reality I think it is expected to always exist? And existing version did assume it existed, so this way the default behavior when no `--temp-dir` is unchanged compared to existing version.

PS:
- Path configuration of TempDir compared to existing CacheDir: CacheDir is not created or verified access on startup like TempDir, but it will be done when it is actually needed. Main reason TempDir is not implemented like this is to avoid having to handle this everywhere `ioutil.TempDir`/`ioutil.TempFile`/`os.MkdirTemp`/`os.CreateTemp` is used. Would not be too hard to fix, of course, with a bit of refactoring.
- Rclone uses `ioutil.TempDir` and `ioutil.TempFile`, but `ioutil` seems sort of deprecated: *As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.* And `ioutil.TempDir` can be replaced with `os.MkdirTemp` (*As of Go 1.17, this function simply calls os.MkdirTemp.*) and `ioutil.TempFile` with `os.CreateTemp` (*As of Go 1.17, this function simply calls os.CreateTemp.*).
	- But as rclone is built with go 1.14, 1.15 og 1.16, in addition to 1.17, we'll probably have to wait...

**3) Document CacheDir**

Noticed that the existing `--cache-dir` flag was not documented, except listed in the flags table. Should have been a separate PR, but I mention the new flag `--temp-dir` because cache uses TempDir as fallback, so it was convenient to just include it here.

#### Checklist


backend/cache/cache_internal_test.go
```diff
@@ -841,7 +841,7 @@ func newRun() *run {
 	}
 
 	if uploadDir == "" {
-		r.tmpUploadDir, err = ioutil.TempDir("", "rclonecache-tmp")
+		r.tmpUploadDir, err = ioutil.TempDir(testy.GetTempDir(), "rclonecache-tmp")
```
ncw, line 844: I think this commit is bigger than it needs to be.

I think I would have just set the correct environment variable `TMPDIR` or `TMP` depending on OS. This means it will work also in libraries that we call, and will be inherited by any processes we fork, both of which are desirable.

cmd/serve/docker/driver.go
```diff
@@ -40,7 +40,7 @@ type Driver struct {
 // NewDriver makes a new docker driver
 func NewDriver(ctx context.Context, root string, mntOpt *mountlib.Options, vfsOpt *vfscommon.Options, dummy, forgetState bool) (*Driver, error) {
 	// setup directories
-	cacheDir, err := filepath.Abs(config.CacheDir)
+	cacheDir, err := filepath.Abs(config.GetCacheDir())
```
ncw, line 43: This no longer needs a `filepath.Abs` does it?

ncw (commented):
This looks like a nice set of changes. `config paths` is a great idea :-)

See inline for some comments.

cmd/serve/docker/driver.go
albertony: Correct. Good catch.

backend/cache/cache_internal_test.go
albertony: > I think I would have just set the correct environment variable TMPDIR or TMP (in the config routines) depending on OS.

Not very fond of this alternative. 😠 

> This means it will work also in libraries that we call, and will be inherited by any processes we fork, both of which are desirable.

But now you've convinced me! 😆

albertony:
Thank you. 

I did push a new version now, where:
- SetTempDir set value of environment variable TMPDIR on unix and TMP+TEMP on Windows.
- GetTempDir is removed, all references reverted to os.TempDir.

albertony:
> I think this looks great now and good for merge - thank you :-)

Great! I can merge it later, probably this evening.

> Unfortunately the last SFTP docs commit I did introduced a conflict, but that should be easy to fix.

Yes, I noticed. Will fix. I like your version best🥇

Issue #3246, opened 2019-06-05
Mounting using AutoFS doesn't work


ScuttleSE:
#### Output of `rclone version`
rclone v1.47.0
- os/arch: linux/amd64
- go version: go1.12.3

#### Describe the issue
I have been trying to add a few rclone-mounts to my autofs configuration without success

Mounting the drive using the exact commandline that autofs uses works though, so something is weird...

```handle_packet: type = 3
handle_packet_missing_indirect: token 20, name TeamDrive-Archive, request pid 3291
attempting to mount entry /media/GoogleTest/TeamDrive-Archive
lookup_mount: lookup(file): looking up TeamDrive-Archive
lookup_mount: lookup(file): TeamDrive-Archive -> -fstype=fuse,config=/home/scuttle/.config/rclone/rclone.conf	:rclonefs\#dom-td01\:
parse_mount: parse(sun): expanded entry: -fstype=fuse,config=/home/scuttle/.config/rclone/rclone.conf	:rclonefs\#dom-td01\:
parse_mount: parse(sun): gathered options: fstype=fuse,config=/home/scuttle/.config/rclone/rclone.conf
parse_mount: parse(sun): dequote(":rclonefs\#dom-td01\:") -> :rclonefs#dom-td01:
parse_mount: parse(sun): core of entry: options=fstype=fuse,config=/home/scuttle/.config/rclone/rclone.conf, loc=:rclonefs#dom-td01:
sun_mount: parse(sun): mounting root /media/GoogleTest, mountpoint TeamDrive-Archive, what rclonefs#dom-td01:, fstype fuse, options config=/home/scuttle/.config/rclone/rclone.conf
do_mount: rclonefs#dom-td01: /media/GoogleTest/TeamDrive-Archive type fuse options config=/home/scuttle/.config/rclone/rclone.conf using module generic
mount_mount: mount(generic): calling mkdir_path /media/GoogleTest/TeamDrive-Archive
mount_mount: mount(generic): calling mount -t fuse -o config=/home/scuttle/.config/rclone/rclone.conf rclonefs#dom-td01: /media/GoogleTest/TeamDrive-Archive
```

Just running "mount -t fuse -o config=/home/scuttle/.config/rclone/rclone.conf rclonefs#dom-td01: /media/GoogleTest/TeamDrive-Archive" mounts the folder just fine, but when automount tries it, it just hangs on that last line

ScuttleSE:
[...]

Issue #992, opened 2017-01-06
Feature Request: mount.rclone helper


gordan-bobic:
It would be really handy if there was a mount.rclone wrapper that can be used in /etc/fstab. It would need to accept all the usual paremeters using slightly different syntax, e.g.

rclone --max-read-ahead 4M --read-only --config /path/to/config mount foo:/bar baz

would need to be expressed to the mount helper as:

mount.rclone -o max-read-ahead=4M,ro,config=/path/to/config foo:/bar baz

And, of course, mount.rclone would have to wait until the fs has completed mounting and then go into the background.

Perhaps this could be done by making rclone check whether the binary name it was invoked as is mount.rclone, and if so, alter it's parameter format expectations?

ncw:
Nice idea.

I think this could be achieved by detecting the name of the binary as you say then doing textual manipulation of the `-o parameters` to make the as per what rclone normally requires.

note putting rclone in the background is #723

gordan-bobic:
This should work for mount.rclone until the functionality is folded into the rclone binary:

```
#!/bin/sh

src=$1
dst=$2

shift 2

# Process -o parameters
while getopts :o: opts; do
	case $opts in
		o)
			parms=`echo $OPTARG | sed -e 's/,/ /g'`
			for parm in $parms; do
				if [ $parm == "rw"   ]; then continue; fi
				if [ $parm == "dev"  ]; then continue; fi
				if [ $parm == "suid" ]; then continue; fi
				trans="$trans --$parm"
			done
			;;
		\?)
			echo "Invalid option: -$OPTARG"
			;;
	esac
done

trans="$trans $src $dst"

rclone mount $trans

```

Then in fstab:

`mount.rclone#crypt:/media       /mnt/tmp        fuse    config=/home/media/.rclone.conf,allow-other,default-permissions,read-only,max-read-ahead=16M 0 0`

It produces the command line, but for some reason I cannot actually get it to work because $PATH seems to disappear half way through the mount process, so mount can find neither mount.rclone nor fusermount. And for once it doesn't look like it's SELinux getting in the way...

gordan-bobic:
Now on the wiki: https://github.com/ncw/rclone/wiki/rclone-fstab-mount-helper-script

keithchilders:
+1 for the main rclone binary to function as an fstab helper. Using shell script until then.

ncw:
I'll try to do this for the 1.37 release...

Or unless anyone else wants to have a go?

[...]

PR #5593, opened 2021-09-10
mountlib: fix --daemon for automount and systemd mount

### LINKED ISSUE #2968: daemonize after mount

```
rclone v1.46
- os/arch: linux/amd64
- go version: go1.11.5
```

using mount + --daemon, rclone forks to early. Rclone should daemonize after the mount action in order to avoid sleeps after that.

1) mount the mount point
2) Daemonize on success

Of course: Thanks for this impressive tools.

ivandeex:

`rclone mount` works fine under Windows, Linux or BSD as a standalone process that can be cancelled by **^C** when not needed anymore.
`rclone mount --daemon` somewhat works but is incompatible with automountd or systemd mount due to reasons explained below.
Users used to mitigate incompatibilities by hacks in the rclonefs wrapper (see https://github.com/rclone/rclone/issues/3246#issuecomment-888375594 or [wiki/rclone-mount-helper-script](https://github.com/rclone/rclone/wiki/rclone-mount-helper-script)).

This pull request removes the need in hacks.
However the wrapper is still needed to translate `mount -o opt=val`-style options into rclone flags (but see #5594).

Disclaimer: this PR does not change `rclone mount` behavior on Windows. We simply don't need to as `--daemon` is no-op on this OS and automount is not present, it's a Unix only notion. And it's a Unix only PR.

#### New Mountpoint Checks

Current way of checking whether mount point is already covered by a mount is badly broken. It lists the directory to see if it contains files. First, an empty mount will be wrongly treated as no mount. Next, if this mount point is watched by Automountd or systemd.autofs then depending on the rclone PGID such access can trigger autofs kernel hooks. If rclone itself runs under autofs control, this can result in a _race_. The 2nd commit of this PR is dedicated to fixing this.

https://github.com/rclone/rclone/issues/992#issuecomment-612318533 noted possible fixes, in particular watching [kernel mount uevents](https://stackoverflow.com/a/1113371), a _stat trick_ and `/proc/mounts` (procfs). The _uevent_ method can only find state changes but cannot tell if a directory was mounted before rclone started. The _stat_ method suffers from the same _autofs race_. The _procfs_ method uses one of major Linux kernel APIs (they are syscalls, procfs/sysfs, device ioctl, net sockets) and requires manual polling. It allows to query what software is responsible for a mount: autofs, kernel, rclone, fuse and so on. It's free from the race as it doesn't touch the mountpoint. Unfortunately it's only applicable on Linux.

This PR adopts _procfs_ method on Linux for checking whether mountpoint is already mounted by anyone or not (the `CheckMountEmpty` method). Additionally it polls _procfs_ to check whether mountpoint was mounted by rclone (in the `CheckMountReady` method).

I'm not a BSD or macOS guru so I left them as is. That's out of scope for this PR. The `CheckMountEmpty` method on **BSD/macOS will still list directory**. The `CheckMountReady` will just do nothing as I don't know how to identify mount source without procfs. Consequently I don't recomment to use rclone mount with Automountd on BSD unless you know what you are doing.

#### Backgrounding

Golang does not natively support Unix `fork` due to its problems with goroutines. Currently rclone uses the `sevlyar/go-daemon` package which mitigates this by invoking the same executable file with the same arguments as a second rclone process. The new process will go through the same initialization steps until it reaches the daemonization point, thus mimics a fork. Nevertheless, the package solves a slightly different task. First, it provides options to redirect the background process output to a log which duplicates rclone's `--log-file` approach (this patch suppresses background output by redirecting to `/dev/null`, so you should use `--log-file` to see output). Next, it allows to save identifier of background process in a pid file. This option was useful for classic SysV Unix daemons but not for rclone. This patch removes the `go-daemon` dependency and reimplements backgrounding directly.

Most importantly, `go-daemon` unconditionally performs `setsid` during backgrounding which appeared to be incompatible with Automount. The [setsid](https://linux.die.net/man/2/setsid) call in the child process resets its process group id (`PGID`) to its PID thus detaching it from parent. While it was important for `fork`-based SysV Unix daemons, it makes autofs fail because Automount needs to tell the process serving the mount from processes accessing it, and they do it by the main process's `PGID` (see https://blog.janestreet.com/how-does-automount-work-anyway, https://www.kernel.org/doc/Documentation/filesystems/automount-support.rst, https://www.kernel.org/doc/Documentation/filesystems/autofs4.txt). If `PGID` is changed, Automount will not treat the background rclone process as a mount provider. If the background process attempted to access the mountpoint directory (eg. when it's listed on macOS/BSD to ensure it's empty), the automount daemon (classic or systemd) will spawn another rclone mount resulting in a race. Note that the official `rclonefs` wrapper avoids the `setsid` problem -- it does not use the `--daemon` argument at all but sends rclone to background using the `&` shell operation. This patch **removes** `setsid` altogether.

The parent process and the background process must know their roles. This patch just sets a special environment variable when it spawns a background process in the same way as `sevlyar/go-daemon` and checks for it early during initialization.

The Automount daemon requires that mount is fully setup by the time the parent process exits. Currently `rclone --daemon` parent exits immediately without waiting for setup to complete. This results in Automount failing with `mount not ready`. The `rclonefs` wrapper used to use `sleep` as a workaround. This patch adds a new flag named `--daemon-wait` which defines how long should the parent wait. On Linux we poll `/proc` to detect the ready state so this flag defines a **maximum time to wait** (see the `WaitMountReady` function). On macOS/BSD the detection is not yet implemented so this flag defines a **constant time to wait** equivalent to a simple `sleep`. I'd rather call this flag `--daemon-timeout` but this name has been already taken.

If mount fails in background or if user hits `^C`, the parent will cleanup and kill the background process to prevent zombies.

#### Miscellaneous

The function `IsDaemon` is put in the `fs` package rather than in `lib/daemonize` to prevent import loops because #5594 will call it very early, in the `fs` init block.

Systemd runs mount units _without any environment_ (PATH, HOME or whatever) so this patch provides a fallback `PATH` to let `go-fuse` find the `fusermount` binary.

Additionally this patch moves a few utility functions unchanged from `mount.go` to `utils.go`: `CheckOverlap`, `CheckAllowings`, `SetVolumeName`.

https://forum.rclone.org/t/ls-on-autofs-triggered-rclone-mount-hangs-the-first-time/26781
Fixes #2968
Fixes https://github.com/rclone/rclone/issues/3246#issuecomment-888404144

#### Checklist


ivandeex:
Manual unit test:
- create mount helper for mount arguments translation `/sbin/mount.rclone`
```
#!/bin/bash
remote=$1
mountpoint=$2
shift 2
args=""
while getopts :o: opts; do
    case $opts in
        o)
            for p in ${OPTARG//,/ }; do
                case "$p" in
                    rw|ro|suid|nosuid|dev|nodev|exec|noexec|noauto|x-systemd*)
                        continue ;;
                    *)
                        args="$args --$p" ;;
                esac
            done
            ;;
        *)
            echo "Invalid option: -$OPTARG" ;;
    esac
done
# note that no hacks are needed, just replace bash by rclone and move on
exec rclone mount --daemon $args "$remote" "$mountpoint" </dev/null >/dev/null 2>&1
```
- create a sample rclone config to your likes
- run mount like
```
sudo mount sftp1: /mnt -t rclone -o verbose=2,log-file=/tmp/rclone.log
```
- verify that it works
- optionally configure automountd or systemd automount (see examples in #5594)

ivandeex:

fs/log/log.go
```diff
@@ -113,6 +113,8 @@ func InitLogging() {
 	}
 	log.SetFlags(flags)
 
+	fs.LogPrintPid = strings.Contains(flagsStr, ",pid,")
```
ncw, line 116: This needs to go in the docs

https://github.com/rclone/rclone/blob/bb0c4ad2d8af8af1ff9abea54dbdf97eadc9018f/docs/content/docs.md#L1026

ncw (approved):
I put an inline note about docs for `pid`.

We probably need some docs about automount but they should go in #5594 probably which fixes it properly.

If you can add the small doc I'm happy for you to merge - it should be a great improvement.

kforner:
Hey, @ncw just pointed me at your fix. 
I have this problem (among many) : https://forum.rclone.org/t/ls-on-autofs-triggered-rclone-mount-hangs-the-first-time/26781
and I'm delighted to see you're working on this. 
I'm going to test it right now.

kforner:
this fixes my issue !

**auto.master.debug**
```
/- 	/etc/auto.rclone.rebex --timeout=30
```
**auto.rclone.rebex**
```
/tmp/rebex2   -fstype=rclone,allow-non-empty,allow-other,log-file=/tmp/rclonefs.log,verbose   "::ftp,host=test.rebex.net,user=demo,pass=7l3xYbNT_gPzlsphlwKo11VRU4JUTAuE:"
```
**running:**
```
sudo automount -fFvd  auto.master.debug
```

accessing the mountpoint:
```
%mount  | grep rebex2
/etc/auto.rclone.rebex on /tmp/rebex2 type autofs (rw,relatime,fd=5,pgrp=20174,timeout=30,minproto=5,maxproto=5,direct,pipe_ino=196900)

 %ls -l /tmp/rebex2/pub/example/
total 404
drwxr-xr-x 1 root root     0 Feb 10  2021 ./
drwxr-xr-x 1 root root     0 Oct 19  2020 ../
-rw-r--r-- 1 root root 19156 Feb 16  2007 imap-console-client.png
-rw-r--r-- 1 root root 36672 Mar 19  2007 KeyGenerator.png

%mount  | grep rebex2
/etc/auto.rclone.rebex on /tmp/rebex2 type autofs (rw,relatime,fd=5,pgrp=20174,timeout=30,minproto=5,maxproto=5,direct,pipe_ino=196900)
:ftp{Q8nY_}: on /tmp/rebex2 type fuse.rclone (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other)

### after the timeout
%mount  | grep rebex2
/etc/auto.rclone.rebex on /tmp/rebex2 type autofs (rw,relatime,fd=5,pgrp=20174,timeout=30,minproto=5,maxproto=5,direct,pipe_ino=196900)
```
rclone log:
```
2021/10/01 17:45:13 NOTICE: Config file "/.rclone.conf" not found - using defaults
2021/10/01 17:45:13 NOTICE: Config file "/root/.config/rclone/rclone.conf" not found - using defaults
2021/10/01 17:45:13 INFO  : ftp://test.rebex.net:21: poll-interval is not supported by this remote
```
Note that we don't see the `INFO  : Exiting...` in the log, that may be the only glitch so far.

ivandeex:
> And funny thing your processing args loop looks exactly like mine in rclonefsdev

With #5594 you can forget rclonefsdev

kforner:
> > And funny thing your processing args loop looks exactly like mine in rclonefsdev
> 
> With #5594 you will forget rclonefsdev

Interesting...
But I'm also trying to run the autofs rclone mount with my current user (not root) with expected benefit of:
- using the user config automatically
- having more control on the rclone executable
- security...

I was trying to hack your rclone.mount for that but with no success so far.
Any idea ?

ivandeex:
Possible but out of scope for this ticket

ivandeex:
Possible but
- niche (suggested way is to run mount as root and specify some of `-o uid=x,gid=y,allow_other`)
- out of scope for this ticket

kforner:
Thanks. So what should I do ? Create a new ticket ?

ivandeex:
Looks more suitable for support ticket on the forum

fs/log/log.go
ivandeex: done

ivandeex:
Merged!

Issue #2805, opened 2018-11-30
Automatic unmount after idle time


tobiisdesired:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.43.1
- os/arch: linux/amd64
- go version: go1.11

#### What problem are you are trying to solve?

An encrypted mount should automatically be unmounted after some time without access to it. So your data stays secure even if you forget to unmount it.

#### How do you think rclone should be changed to solve that?

`rclone mount` could provide an idle timeout parameter. Other tools like `encfs` have this feature.
Here is an excerpt from the `encfs` manual pages:

>   -i, --idle=MINUTES
>            Enable automatic unmount of the filesystem after a period of inactivity.  The period is specified in minutes, so the shortest timeout period that can be requested is one minute.  EncFS
>            will not automatically unmount if there are files open within the filesystem, even if they are open in read-only mode.  However simply having files open does not count as activity.

ncw:
This is an interesting idea.

The FS can't umount if it has files open - that would break stuff alas.

You could check out autofs which unmounts file systems when not in use.  I use it for NFS mounts.  I haven't tried it with rclone mount but it should work...

ncw:
I think running rclone on automount is the best solution here - it lets the OS do the tricky job of working out whether we have any open files.

Since we have patches in the pipeline to make that really smooth I'm going to close this one.

PR #5594, opened 2021-09-10
Run rclone as a mount helper program

### LINKED ISSUE #992: Feature Request: mount.rclone helper

ivandeex:

This PR will make rclonefs wrappers moot!

The PRedecessor #5593 solve(d) most problems that prevented rclone from running the mount in background i.e. working _like a generic unix mount_ program (`sshfs`, `mount.glusterfs` and others can do it out of the box), namely:
- have the main process prepare for mounting,
- spawn a background process to handle the mount,
- wait a little until the mount success or timeout (kill the background process if mount failed)
- don't block, exit with status code

The last incompatibility left is the way mount arguments are passed.
The core `/bin/mount` program will (in the basic scenario) get the `-t FSTYPE` argument then run the `/sbin/mount.FSTYPE` helper program passing it mount options. Automountd and systemd generally follow the same logic...
But rclone expects GNU-styled flags like `--key val` while mount options are passed as `-o key=val,...` (or `--opt=...`)!
This patch fills the last gap.

### Usage

Please symlink the rclone binary to `/sbin/mount.rclone` (and optionally `/usr/bin/rclonefs`) and **just**
- run something like
```
mount sftp1:subdir /mnt/data -t rclone -o vfs_cache_mode=writes,sftp_key_file=/path/to/pem
```
- or create systemd mount units (provide explicit `config=...,cache-dir=...` as mounts run without `HOME` !)
```
# /etc/systemd/system/mnt-data.mount
[Unit]
After=network-online.target
[Mount]
Type=rclone
What=sftp1:subdir
Where=/mnt/data
Options=rw,allow_other,args2env,vfs-cache-mode=writes,config=/etc/rclone.conf,cache-dir=/var/rclone
```
- optionally augmented by systemd automount unit
```
# /etc/systemd/system/mnt-data.automount
[Unit]
After=network-online.target
Before=remote-fs.target
[Automount]
Where=/mnt/data
TimeoutIdleSec=600
[Install]
WantedBy=multi-user.target
```
- or add in `/etc/fstab` a line like
```
sftp1:subdir /mnt/data rclone rw,noauto,nofail,_netdev,x-systemd.automount,args2env,vfs_cache_mode=writes,config=/etc/rclone.conf,cache_dir=/var/cache/rclone 0 0
```
- or use classic `Automountd` !

P.S. There was a ticket #2805 requesting the new `rclone mount --idle=X` flag but now we can do it in a standard way making it mostly moot (except on windows)...

### Implementation

~~This patch introduces new slice `fs.Args` which is by default a copy of `os.Args` but can be rewritten if rclone executable was invoked as `mount.rclone` or `rclonefs`.~~
**UPDATE**
This patch rewrites `os.Args` inplace in an early **init()** block if rclone executable was invoked as `mount.rclone` or `rclonefs`.

We use `os.Args[0]` rather than `os.Executable` to detect the name since the latter might resolve symlink while we want it unresolved.

Mount helpers usually lack the `HOME` or `RCLONE_CONFIG` environment variables, so config setup code in `fs/config` will require `fs.Args` to find a correct (probably converted) `--config FILE` running **very early** (most probably called from init blocks). Consequently the command line transformer also runs **super early** in the `init` block of the core `fs` package and avoids using other packages to **prevent import loops**. Utility functions `IsDaemon` and `IsMountHelper` were put directly in `fs` for this very reason.

Basically the conversion splits `-o` argument(s) by comma, replaces `_` by `-` and prepends `--` to transform `-o` options into Cobra flags. Options containing commas or spaces can be wrapped in single or double quotes (any quotes inside outer quotes should be doubled). The parser algorithm is based on the _connection string parser_ with a few command-line specific modifications.

The option syntax additionally includes a few pseudo mount options treated specially:
- `env.NAME=VALUE` will set an environment variable for this rclone instance and its children.
   This helps with Automountd and Systemd.mount which don't let set custom environment for mount helpers.
   Typically you will use `env.HTTPS_PROXY=proxy.host:3128` or `env.HOME=/root`
- `command=cmount` can be used to run any other command rather than default `mount` in the helper mode 
- `args2env` will pass mount options to the daemonized mount helper via `RCLONE_X` environment variables instead of CLI flags. This allows to hide secrets from `ps` or `pgrep`.
- `-o vv...` will be transformed into `--verbose=N`, just for convenience
- `x-systemd.automount`, `_netdev`, `nosuid` and alike are intended for automountd or systemd and ignored by rclone

Based on #5593 (this PR includes preliminary patch of #5593 expected to annihilate at merge time)
Fixes #992

#### Checklist

### DIFF
--- a/fs/config.go
+++ b/fs/config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -178,6 +179,11 @@ func NewConfig() *ConfigInfo {
 		}
 		if arg == "--log-level=DEBUG" || (arg == "--log-level" && len(os.Args) > argIndex+1 && os.Args[argIndex+1] == "DEBUG") {
 			c.LogLevel = LogLevelDebug
+		}
+		if strings.HasPrefix(arg, "--verbose=") {
+			if level, err := strconv.Atoi(arg[10:]); err == nil && level >= 2 {
+				c.LogLevel = LogLevelDebug
+			}
 		}
 	}
 	envValue, found := os.LookupEnv("RCLONE_LOG_LEVEL")
--- a/cmd/mountlib/help.go
+++ b/cmd/mountlib/help.go
@@ -12,13 +12,22 @@ On Linux and OSX, you can either run mount in foreground mode or background (dae
 Mount runs in foreground mode by default, use the |--daemon| flag to specify background mode.
 You can only run mount in foreground mode on Windows.
 
+In background mode rclone acts as a generic Unix mount program: the main program
+starts, spawns a background rclone process to setup and maintain the mount, waits
+until success or timeout, kills the child process if mount fails, and immediately
+exits with appropriate return code.
+
 On Linux/macOS/FreeBSD start the mount like this, where |/path/to/local/mount|
 is an **empty** **existing** directory:
 
     rclone @ remote:path/to/files /path/to/local/mount
 
 On Windows you can start a mount in different ways. See [below](#mounting-modes-on-windows)
-for details. The following examples will mount to an automatically assigned drive,
+for details. If foreground mount is used interactively from a console window,
+rclone will serve the mount and occupy the console so another window should be
+used to work with the mount until rclone is interrupted e.g. by pressing Ctrl-C.
+
+The following examples will mount to an automatically assigned drive,
 to specific drive letter |X:|, to path |C:\path\parent\mount|
 (where parent directory or drive must exist, and mount must **not** exist,
 and is not supported when [mounting as a network drive](#mounting-modes-on-windows)), and
@@ -225,6 +234,13 @@ Hubic) do not support the concept of empty directories, so empty
 directories will have a tendency to disappear once they fall out of
 the directory cache.
 
+When mount is invoked on Unix with |--daemon|, the main rclone program
+will wait until the background mount is ready until timeout specified by
+the |--daemon-wait| flag. On Linux rclone will poll ProcFS to check status
+so the flag sets the **maximum time to wait**. On macOS/BSD the time to wait
+is constant and the check is performed only at the end of sleep so don't
+set it too high...
+
 Only supported on Linux, FreeBSD, OS X and Windows at the moment.
 
 ### rclone @ vs rclone sync/copy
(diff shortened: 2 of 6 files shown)


ivandeex:
@ncw
The latest `iva` build from https://github.com/ivandeex/rclone/releases already includes this patch so you can  play right now. I've run the patch on a staging box for a month or two before submitting.
Note the `-o args2env` mount flag that hides mount options from `pgrep -af rclone`.
I just need to write docs...

cmd/cmd.go
```diff
@@ -273,7 +273,7 @@ func Run(Retry bool, showStats bool, cmd *cobra.Command, f func() error) {
 			break
 		}
 		if retryAfter := accounting.GlobalStats().RetryAfter(); !retryAfter.IsZero() {
-			d := retryAfter.Sub(time.Now())
+			d := time.Until(retryAfter)
```
ivandeex, line 276: Just to satisfy a warning in the VScode linter...

cmd/cmd.go
```diff
@@ -458,7 +458,7 @@ func initConfig() {
 		})
 	}
 
-	if m, _ := regexp.MatchString("^(bits|bytes)$", *dataRateUnit); m == false {
+	if m, _ := regexp.MatchString("^(bits|bytes)$", *dataRateUnit); !m {
```
ivandeex, line 461: Just to satisfy a warning in the VScode linter...

fs/config/config.go
```diff
@@ -246,9 +246,11 @@ func makeConfigPath() string {
 	// variable, since then we skip actually trying to create the default
 	// and report any errors related to it (we can't use pflag for this because
 	// it isn't initialised yet so we search the command line manually).
+	// Use processed command line from fs.Args so we can pick `-o config=...`
+	// if rclone was invoked as a mount helper.
 	_, configSupplied := os.LookupEnv("RCLONE_CONFIG")
 	if !configSupplied {
-		for _, item := range os.Args {
+		for _, item := range fs.Args {
 			if item == "--config" || strings.HasPrefix(item, "--config=") {
```
ivandeex, line 254: This `fs/config` code can be used very early so we run conversion super early, in the `init` block of `fs` package.

ivandeex (commented):
see notes in the code...

ivandeex:
@ncw
The latest `iva` build from https://github.com/ivandeex/rclone/releases already includes this patch so you can  play right now. I've run the patch on a staging box for a month or two before submitting.
Note the `-o args2env` mount flag that hides mount options from `pgrep -af rclone`.

I don't yet call for review as I still need to put user notes from this PR's top post in the mount docs.

ncw:
I had a quick skim of the code and it seems to do the right things :-)

I don't know if you saw my prototype for this? 78ffe3707792730354249524a8cfd8fb36bb8a57

Not sure there is anything you can learn from that, but I thought I'd put it out there.

It directly modifies `os.Args` rather than using an `fs.Args` which is less code to change.

ivandeex:
> It directly modifies `os.Args` rather than using an `fs.Args` which is less code to change.

I wanted to keep a backup of original arguments just for debugging purposes.
If you think it's an overkill I can drop `fs.Args` and modify `os.Args` inplace.

ivandeex:
> I don't know if you saw my prototype for this? [78ffe37](https://github.com/rclone/rclone/commit/78ffe3707792730354249524a8cfd8fb36bb8a57)

Not yet. Noted...

ncw:
> > It directly modifies `os.Args` rather than using an `fs.Args` which is less code to change.
> 
> I wanted to keep a backup of original arguments just for debugging purposes.
> If you think it's an overkill I can drop `fs.Args` and modify `os.Args` inplace.

I'd say let's modify os.Args directly - we can always put a debug in with the old and the new values or something.

ivandeex:

ivandeex:
> let's modify os.Args directly - we can always put a debug in with the old and the new values or something.

done

ncw (commented):
I think this code looks great - thank you.

However it could do with some docs - you wrote some nice ones in the PR - maybe cut and paste into `cmd/mountlib/help.go`? (Though I think that file is getting too long and we should probably break it out into its own docs page say `docs/content/mount.md` at some point...)

fs/config.go
```diff
@@ -179,6 +180,11 @@ func NewConfig() *ConfigInfo {
 		if arg == "--log-level=DEBUG" || (arg == "--log-level" && len(os.Args) > argIndex+1 && os.Args[argIndex+1] == "DEBUG") {
 			c.LogLevel = LogLevelDebug
 		}
+		if strings.HasPrefix(arg, "--verbose=") {
```
ivandeex, line 183: @ncw 
Thank you very much for review.

Do you mind if I put mountlib documentation in a separate commit?

Also I'd like to note that this early `--verbose=x` support in fs/config.go isn't directly connected with mount helper cli translator? I could split it in yet another commit. Do you mind?

Thank you

ivandeex (commented):
two small questions...

fs/config.go
ivandeex: @ncw 
Thank you very much for review.

Do you mind if I put mountlib documentation in a separate commit?

Also I'd like to note that this early `--verbose=x` parser in `fs/config.go` isn't directly connected with mount helper cli translator and may be split it in yet another commit. Do you agree? Or better keep it squashed?

Thank you

ivandeex:
@ncw @albertony 

Please be aware that I documented rclone mount helper in a dedicated commit https://github.com/rclone/rclone/pull/5594/commits/bd82c19dc0084094b86d2919d4f58283601e2c62 partially including the `make commanddocs` output for your reference.

Also updated the related wiki article https://github.com/rclone/rclone/wiki/rclone-mount-helper-script

ivandeex:
@ncw @albertony 

Please be aware that I documented rclone mount helper in a dedicated commit https://github.com/rclone/rclone/pull/5594/commits/bd82c19dc0084094b86d2919d4f58283601e2c62 (https://github.com/rclone/rclone/commit/a065fb23e5238f5252002799905d455265e94998 on master) partially including the `make commanddocs` output for your reference.

Also updated the related wiki article https://github.com/rclone/rclone/wiki/rclone-mount-helper-script

ncw:
Thank looks great, thank you.

ivandeex:
Cool.
I'm moving on with submitting FTP stuff accumulated this summer.

PR #5590, opened 2021-09-09
ftp: fix deadlock after failed update when concurrency=1

ivandeex:
#### What is the purpose of this change?

Our users complained many times that rclone FTP client manifests random problems with completing a file upload.

As my investigations have shown, the bug triggers because [jlaffaye/ftp does not properly nudge](https://github.com/jlaffaye/ftp/blob/5d41901190670f202a4aafb39c81d72e2e42bd8d/ftp.go#L721) the [fshttp anti-hanging timer](https://github.com/rclone/rclone/blob/v1.56.0/fs/fshttp/dialer.go#L85) on command channel after a file upload. I have a [working fix](https://github.com/ivandeex/rclone/releases) for [jlaffaye/ftp](https://github.com/ivandeex/ftp/commit/08390f8cbaa60e98bb1687989752e31feaf8b416) and [backend/ftp](https://github.com/ivandeex/rclone/commit/f70e98d1595feadf6d52ade6e9565aec3ed77aa3) which I will submit **later**.

This pull request solves a second-stage bug triggered by the above bug which can be fixed **solely** in rclone. Frequently when users reported the above bug we recommended them to simplify their FTP client settings, in particular set `--ftp-concurrency=1`. When the FTP backend's [upload routine](https://github.com/rclone/rclone/blob/v1.56.0/backend/ftp/ftp.go#L1063) encounters a error, it will remove the failed file. The removal needs another connection but due to low concurrency setting it will wait for [connection token](https://github.com/rclone/rclone/blob/v1.56.0/backend/ftp/ftp.go#L317) forever.

This patch modifies the upload routine so that after error it will _first_ release the failed connection and _then_ remove the failed file and thus avoid the deadlock.
Due to presence of another bug I will include a [relevant unit test](https://github.com/ivandeex/rclone/commit/f70e98d1595feadf6d52ade6e9565aec3ed77aa3#diff-67452ac6849fa36379c4aa76d2d528fd2dbe411ac03493522a63e59ea03a8c40R84) with a dedicated pull request #5596, not here. I just hope that the change looks self-evident. @ncw ??

#### Was the change discussed in an issue or in the forum before?

- https://forum.rclone.org/t/rclone-copy-to-ftp-rclone-sends-a-quit-and-never-ends-itself/24046/10
- #5596 

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
[...]

Issue #5328, opened 2021-05-14
FTP upload and download for big files (over 10GB) fails


jcn50:
#### Output of `rclone version`
 rclone: Version "v1.55.1"

#### Describe the issue
This happens only for big files (over 10GB): the upload goes to 100% then rclone disconnects, reconnects to check the file on destination but declares that the upload has failed and restarts it! Adding `--no-check-dest --low-level-retries 1` does not resolve the issue.

**It works when uploading manually with the `ftp` command! So the issue is NOT with the FTP server!**

Typical rclone log/verbose going into a loop:
```
2021-05-13 18:34:22 DEBUG : big.zip: Need to transfer - File not found at Destination
2021-05-13 18:44:53 DEBUG : ftp://server.com:21/: Connecting to FTP server
2021-05-13 18:44:54 DEBUG : big.zip: Removed after failed upload: read tcp IP1:37706->IP2:21: i/o timeout
2021-05-13 18:44:54 DEBUG : big.zip: Received error: update stor: read tcp IP1:37706->IP2:21: i/o timeout - low level retry 1/10
```
[I suspect the FTP library to be behind this bug](https://sourcegraph.com/github.com/artpar/rclone/-/blob/backend/ftp/ftp.go#L1044:8): there is no timeout issue, I can see the file going to 100% on the FTP destination before it gets deleted by rclone. TLS is not used.

**My suggestions/questions to fix**:
a) Why does rclone disconnects + reconnects to check the file instead of using the same FTP session?
b) Can't rclone sleeps for 2-3 seconds before doing the check of the uploaded file?

=> see log with `--dump bodies` attached!
[rclone_issue_5328.txt](https://github.com/rclone/rclone/files/6476155/rclone_issue_5328.txt)

ncw:
> a) Why does rclone disconnects + reconnects to check the file instead of using the same FTP session?

Good idea...

In the code you linked, rclone could open a new connection and check the size of the file. It if was as expected then it could leave it alone rather than deleting it.

Fancy sending a pull request to fix?

> b) Can't rclone sleeps for 2-3 seconds before doing the check of the uploaded file?

Why do you think that would help?

jcn50:
> Fancy sending a pull request to fix?

Sorry this is above my abilities 😞 

> Why do you think that would help?

[...]

Issue #4622, opened 2020-09-24
Cannot finish downloading files from FTP


ben108472:
#### What is the problem you are having with rclone?
Cannot finish downloading files from FTP.
The file is downloaded repeatedly but never finish downloading.
The transferred size becomes bigger and bigger.

#### What is your rclone version (output from `rclone version`)
rclone v1.53.1-DEV

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
CentOS 6.5

####  Which cloud storage system are you using? (eg Google Drive)
FTP

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
rclone copy

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
2020/09/24 09:16:24 DEBUG : Using config file from "/home/.config/rclone/rclone.conf"
2020/09/24 09:16:24 DEBUG : Creating backend with remote "URL/filename"
2020/09/24 09:16:24 DEBUG : URL/filename: Connecting to FTP server
2020/09/24 09:16:28 DEBUG : fs cache: adding new entry for parent of "URL/filename", ""
2020/09/24 09:16:28 DEBUG : Creating backend with remote ""
2020-09-24 09:16:29 DEBUG : filename: Need to transfer - File not found at Destination
2020-09-24 09:16:29 DEBUG : preAllocate: got error on fallocate, trying combination 1/2: operation not supported
2020-09-24 09:16:29 DEBUG : preAllocate: got error on fallocate, trying combination 2/2: operation not supported
2020-09-24 09:16:29 DEBUG : filename: Starting multi-thread copy with 4 parts of size 1.197G
2020-09-24 09:16:29 DEBUG : filename: multi-thread copy: stream 4/4 (3857055744-5142538872) size 1.197G starting
2020-09-24 09:16:29 DEBUG : filename: multi-thread copy: stream 2/4 (1285685248-2571370496) size 1.197G starting
2020-09-24 09:16:29 DEBUG : URL/filename: Connecting to FTP server
2020-09-24 09:16:29 DEBUG : filename: multi-thread copy: stream 3/4 (2571370496-3857055744) size 1.197G starting
2020-09-24 09:16:29 DEBUG : URL/filename: Connecting to FTP server
2020-09-24 09:16:29 DEBUG : filename: multi-thread copy: stream 1/4 (0-1285685248) size 1.197G starting
[...]

PR #5596, opened 2021-09-11
ftp: fix I/O timeout after long uploads

ivandeex:

Uploading of large files via this library and rclone frequently ends up with `i/o timeout`.

My investigation shows the following:
- `jlaffaye/ftp` makes control/data streams from a connection factory provided by rclone. This factory wraps standard go tcp streams in the [code](https://github.com/rclone/rclone/blob/v1.56.0/fs/fshttp/dialer.go#L85) which safeguards against silently dead network connections (lost without explicit TCP shutdown), namely cancels streams after timeout but resets timeout timer after each read/write.
- A typical FTP upload looks like: client talks with ftp server on control stream, sets upload parameters, then create one or more data streams and do the upload (this can be long), then return to control stream and ask ftp server about upload results  (here be dragons)
- While client was busy uploading on data stream, the control stream timeout happened and client's attempt to read status returned **I/O timeout**

This patch makes `rclone ftp client` nudge the control connection right after a (probably long) upload has finished. It depends on the https://github.com/jlaffaye/ftp/pull/231

Please note that the `fs/fshttp` commit is not strictly required, I just wanted to make the code look a little bit cleaner on my eyes...

The `DON`T MERGE` commit can be removed after https://github.com/jlaffaye/ftp/pull/231 gets merged. It just allows to verify changes in advance...

Derivative manifestations of this problem were reported a number of times on rclone forum and tracker:
- https://forum.rclone.org/t/rclone-copy-to-ftp-rclone-sends-a-quit-and-never-ends-itself/24046
- https://github.com/rclone/rclone/issues/5328
- https://github.com/rclone/rclone/issues/4622
- https://github.com/rclone/rclone/pull/5590 (fixes a derivative bug)

#### Checklist


jcn50:
Thanks a lot for that work! Looking forward to v1.57!

jcn50:
_I think curl should be used as much as possible (for FTP for example) because it is the fastest & most reliable library~_

ivandeex:
> Looking forward to v1.57!

Whether we have it in 1.57 depends primarily on how long it will take to accept the upstream patch. You should see the 1.57 label as a preliminary possibility.

> You know this is also affecting the download.

The upstream patch introduces the deadline nudging in a few places though the investigation started from upload case.

> I think curl should be used as much as possible (for FTP for example) because it is the fastest & most reliable library

curl is a C application / library. we support C bindings only on windows (probably on macOS). rclone is primarily (on unix) distributed as a static golang application so can't use it. moreover rclone uses custom tcp connections written in golang for features like bandwidth accounting / limiting or dead peer elimination. if you ask me, I would prefer to get rid of ftp dependencies and parse ftp protocol messages directly in the ftp backend.

ivandeex:
> Looking forward to v1.57!

Whether we have it in 1.57 depends primarily on how long it will take to accept the upstream patch. You should see the 1.57 label as a preliminary possibility.

> You know this is also affecting the download.

The upstream patch introduces the deadline nudging in a few places though the investigation started from an upload case.

> I think curl should be used as much as possible (for FTP for example) because it is the fastest & most reliable library

You know `curl` is a C application / library. We support C bindings only on windows and macOS. On Unix rclone is distributed as a static golang application, so can't use it.

Moreover rclone uses custom TCP connections written in golang for features like bandwidth accounting / limiting or dead server elimination.

By the way the latter feature while useful is the cause of I/O timeout in the _jlaffaye_ client addressed by this patch. In other words we keep the feature but improve its compatibility with the protocol parser.

In future I plan to add even more features to the custom TCP class: socks proxy support and DNS resolver hooks.

If you ask me, I'd prefer to get rid of any ftp dependencies and parse ftp messages directly in the ftp backend.

cc @ncw

ivandeex:
> Looking forward to v1.57!

Whether we have it in 1.57 depends primarily on how long it will take to accept the upstream patch. You should see the 1.57 label as a preliminary possibility.

> You know this is also affecting the download.

The upstream patch introduces deadline nudging in a few places though the investigation started from an upload case.

> I think curl should be used as much as possible (for FTP for example) because it is the fastest & most reliable library

You know `curl` is a C application / library. We support C bindings only on windows and macOS. On Unix rclone is distributed as a static golang application, so can't use it.

Moreover rclone uses custom TCP connections written in golang for features like bandwidth accounting / limiting or dead server elimination. By the way the latter feature is the cause of I/O timeout in the `jlaffaye/ftp` client. Here we keep it as it's useful, but polish.

And finally, my opinion. In future I'd prefer to get rid of any ftp dependencies and parse protocol messages directly in the backend, I think it can generally simplify code but I don't want to overhaul and retest now.

ivandeex:

CarlosHartmann:
Noob-alert: does this mean:

1. That the fix is done?
2. That it will be implemented in v. 1.58? Or possible earlier?

In any case, thanks a ton! I was getting desperate over this issue.

ivandeex:
@CarlosHartmann 
I am waiting for @jlaffaye to review https://github.com/jlaffaye/ftp/pull/231. This can take long depending on how busy they are. Hopefully by 1.58.
If you need an immediate solution, use the latest `iva` release from https://github.com/ivandeex/rclone/releases. It's roughly equivalent to current rclone 1.57-beta plus my mount and FTP patches..

ivandeex:
This pull request can be long waiting for [upstream](https://github.com/jlaffaye/ftp/pull/231). I created branch [iva-ftp](https://github.com/rclone/rclone/commits/iva-ftp) to have ready [beta](https://beta.rclone.org/branch/iva-ftp/) in advance.

CarlosHartmann:
I just uploaded 20 files all between 12GB and 35GB without one single error. At least for me, this is working perfectly! Let me know if there's a way to tip you some money. But in any case, you have my deepest gratitude for fixing the issue!

ivandeex:
Glad it works for you :)
You can consider donating the project - https://rclone.org/donate/
Thanks you

ivandeex:
@ncw
Also note that I repoint the build to `github.com/rclone/ftp` via the `replace` go.mod directive.
This will let us quickly switch back when needed.
Do you agree?

PR #5597, opened 2021-09-12
Update the ignored email as per #5586

abhi18av:
This PR follows up from the comment here https://github.com/rclone/rclone/pull/5586#issuecomment-917408535


ncw:
Thank you :-)

Issue #2110, opened 2018-03-01
Support for digest authenticaton


zsnmwy:
Hi:

rclone doesn’t support digest authenticaton yet.
You can see this [link](https://forum.rclone.org/t/head-not-found-with-webdav/4935) in detail.
How about support digest authenticaton?
English is not my native language; please excuse typing errors.

buq96:
I need it。

ncw:
I'd like to support this. Does anyone want to have a go at it?

buq96:
> I'd like to support this. Does anyone want to have a go at it?

I want to try.

ncw:
@buq96 wrote:
> > I'd like to support this. Does anyone want to have a go at it?
> 
> I want to try.

That would be great!

We already use https://github.com/abbot/go-http-auth which I think can support digest too - we just need to enable it somehow.

x4455:
please excuse typing errors.
I don't know how to use Golang. But I found some similar code.
This may help with the issues.

https://github.com/studio-b12/gowebdav/blob/master/requests.go
https://github.com/studio-b12/gowebdav/blob/master/digestAuth.go

Drakonas:
Is there any progress on this request being pulled? It seems the build is now producing errors for the PR.

mikegerber:
My Apache DAV service also requires digest authentication.

laurento:
May I ask what's the current status of this issue? I noticed that a couple or PRs didn't make it to upstream over the years. Any direction, idea from the devs? I'm willing to help

pekhota:
@ncw, I'm trying to resolve the issue. Would you mind to check my PR please https://github.com/rclone/rclone/pull/9875

PR #5601, opened 2021-09-15
backend(http): support for digest-authentication

equals03:
The http backend did not support digest authentication.
This update allows for an additional setting:
`use_digest_auth = true`

This will transform the current url, in the usual basic auth format:
`https://user:pass@example.com`
... and use those details via digest authentication if required.

The user information is removed from the final url.

Not directly. All others surround WebDAV. WebDAV was not in my  
current use-case though, while HTTP was so I've started here.

- https://forum.rclone.org/t/rclone-to-webdav-docker-fails/8089
- #2110 

 
While I haven't checked,hopefully, my minor patches here will be
easily transferable.

#### Checklist


ncw:
That is really useful thanks.

A couple of thoughts
- do we need an option? Can't this be auto detected?
- the http and webdav infrastructure are quite tightly intertwined - I think it might "just work" if webdav had the option (or we removed it).

equals03:
no problems 👍 

> do we need an option? Can't this be auto detected?
I had the same thoughts but I elected to go with the option for a few reasons:  

1. I didn't want to send the username and password over the wire (even though its probably over `https`) for the initial request (before the server challenges) if we already know that the server uses digest authentication and I didn't want to introduce additional discrete usernames/passwords as options.
2. the implementation would probably be significantly more complicated than the simple patch here due to changes to the core  implementation, and finally
3. digest auth is so infrequently used (in my experience anyways) that making it opt-in reduces the likelihood of any potential issues with the current implementation down the line as this will be the exception rather than the rule.

> the http and webdav infrastructure are quite tightly intertwined - I think it might "just work" if webdav had the option (or we removed it).   

You are right they are quite intertwined... I took a very quick look at it and I can see where to implement it but the way I do it is to wrap the original `fshttp.NewTransport` round-tripper with the `digest.NewTransport` [here](https://github.com/rclone/rclone/commit/d81179ca64f4909c036db417d8d89f5bcd3cd293#diff-d1784a901d48036c8946afb2386edcc796b1b48bf097f545019a99835b2a4e2fR140-R142)... making it a non-trivial change to that system. 

`NewClient` vs `NewDigestClient`.
I kept it like that so that I could strip the user:pass out of the request before it was sent for [the first time](https://github.com/rclone/rclone/commit/d81179ca64f4909c036db417d8d89f5bcd3cd293#diff-e157f24df76cfd77613bdccd5acb39f6e62351c8b1f96c4a84dd2c7f1f123b9fR183-R190).

Either way, it would probably be a small change to implement something similar there but I didn't want to entangle each with this change. I was tempted but decided to keep this PR focused.

If there is still appetite for it, and this implementation is acceptable, I will definitely look moving the code over.

Nsbx:
Any update about this PR ?
Sorry for the ping, but digest auth option would be really great

ncw:
Hi @equals03, and thank you for this contribution.

Issue #3506, opened 2019-09-03
remote name with dots is always interpreted as local filesystem directory


diegosainz:
#### Output of `rclone version`
```
rclone v1.48.0
- os/arch: linux/amd64
- go version: go1.12.6
```

#### Describe the issue

Given the following two remotes (in this case using backblaze backend):

```
[my.remote]
type = b2
account = xyz
key = xyz

[my-remote]
type = b2
account = xyz
key = xyz
```

When using the remote with dots in its name, it is always interpreted as a local filesystem:

```
$ rclone ls my.remote:mybucket
2019/09/03 14:01:58 ERROR : : error listing: directory not found
2019/09/03 14:01:58 Failed to ls with 2 errors: last error was: directory not found
```

Without dots it works as expected:

```
$ rclone ls my-remote:mybucket
... files....
```

BTW - Thank you for this wonderful software

ncw:
I replicated this with the latest beta...

I'm amazed no-one has noticed this before!

ncw:
This is because the remote names are defined like this

https://github.com/rclone/rclone/blob/b3db38ae3150aad2764a50728e9fddaa3a17ce31/fs/fspath/path.go#L14

So that is `[0-9A-Za-z_ -]` - this does not include `.`

The rclone config checks the name using the same regexp

https://github.com/rclone/rclone/blob/b3db38ae3150aad2764a50728e9fddaa3a17ce31/fs/config/config.go#L1044

```
e) Edit existing remote
n) New remote
d) Delete remote
[snip]
e/n/d/r/c/s/q> n
name> asfsdf.afsf
Can't use "asfsdf.afsf" as it has invalid characters in it.
```

So I guess the question is how did you make these remotes?

I don't think it would cause any backwards compatibility problems to add `.` to the allowed names for a remote.

diegosainz:
Hi Nick, thank you for your quick reply and diagnose of the issue. It surprised me as well to find it, 

My use case is within a configuration management context using Ansible, where I'm using restic to configure backups for many servers and restic is using rclone as backend. Ansible is the one editing directly the `.ini` configuration file (gathering the rclone remote name, type and secrets from other sources), so it didn't go through the rclone config  process.

An alternative, fail-safe approach for this and future cases, may be:

1. If the `:` character is detected then compare the "remote" part directly with the name of configured remotes in the configuration file. 
[...]

PR #5606, opened 2021-09-16
config: allow dot in remote names

albertony:

Add `.` as legal character in remote names. :bomb:

Beta build here:
- https://beta.rclone.org/branch/remote-name-dot/

Note to self:
<sub><sup>- PR from branch on fork</sup></sub>
<sub><sup>  (https://github.com/albertony/rclone/tree/remote_name_dot)</sup></sub>
<sub><sup>- Pushed the branch also to upstream to get beta build, renamed from remote_name_dot to remote-name-dot to be valid in semver</sup></sub>
<sub><sup>  (https://github.com/rclone/rclone/tree/remote-name-dot)</sup></sub>

https://forum.rclone.org/t/problem-with-parsing-of-environment-variables/26528/10

#### Checklist

### DIFF
--- a/fs/fspath/path_test.go
+++ b/fs/fspath/path_test.go
@@ -35,6 +35,9 @@ func TestCheckConfigName(t *testing.T) {
 		{"-remote", errCantStartWithDash},
 		{"r-emote-", nil},
 		{"_rem_ote_", nil},
+		{".", nil},
+		{"..", nil},
+		{".r.e.m.o.t.e.", nil},
 	} {
 		got := CheckConfigName(test.in)
 		assert.Equal(t, test.want, got, test.in)
@@ -49,6 +52,9 @@ func TestCheckRemoteName(t *testing.T) {
 		{":remote:", nil},
 		{":s3:", nil},
 		{"remote:", nil},
+		{".:", nil},
+		{"..:", nil},
+		{".r.e.m.o.t.e.:", nil},
 		{"", errInvalidCharacters},
 		{"rem:ote", errInvalidCharacters},
 		{"rem:ote:", errInvalidCharacters},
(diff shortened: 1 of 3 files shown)


fs/fspath/path_test.go
```diff
@@ -177,8 +179,12 @@ func TestParse(t *testing.T) {
 				Path:         "/path/to/file",
 			},
 		}, {
-			in:      "rem.ote:/path/to/file",
-			wantErr: errInvalidCharacters,
+			in: "rem.ote:/path/to/file",
```
ncw, line 182: You should change this test so it has a different invalid character - same with the ones below and maybe insert a `.` or two in currently passing tests.

ncw (commented):
Last time this came up we decided not to allow dots in remote names, let me see if I can find the discussion...

https://github.com/rclone/rclone/issues/3506

I don't think the argument in that issue holds any more since the parser is a lot more robust now.

I think there was some other discussion on the forum too I can't find.

Can you think of any reasons why this would be a bad idea for backwards compatibility or cause confusion?

The only one I can think of is the confusion with host names. But that is maybe a good thing being allowed to have a remote which looks like a host name.

fs/fspath/path_test.go
albertony: Yes, of course. Will do.

albertony:
Thanks.

I was sort of expecting this PR to be immeditaly shot down and burned... 😆 But its of course because I also see there are different things to consider here..

I will extend the tests, do more manual testing (the issue reporter from the forum thread did report back after testing the beta, with no issues found), and think about if there can be any gotchas..

Will push to see the checks going through on all os etc, but will ping you when its ready for another review!

albertony:
I have extended the unit test. The parsing seem to play very well with whatever "dots" I throw at it..
Haven't found any issues when doing manual testing either.

PR #5613, opened 2021-09-20
backend/tardigrade: backend renamed to storj (tardigrade still works)

elek:

Tardigrade was [renamed](https://storj.io/blog/introducing-storj-dcs-decentralized-cloud-storage-for-developers) to Storj (or Storj DCS) recently. This change applies the renaming to the rclone backend.

Old `tardigrade` backend is kept, but the existing code (which is moved to `backend/storj`) is reused, the existing `tardigrade` becomes to a shortcut for the code. 

I tested it with both an existing `tardigrade` backend and a new `storj` backend. 

The only code change is making the configurations/options public in `storj` package to make it possible to reuse from `tardigrade` 

Some discussion has been started here:

https://forum.rclone.org/t/backward-compatibility-rules/26569

#### Checklist


ncw:
@elek - I was hoping for an issue to discuss this on rather than a pull request!

@calebcase - ignoring the backwards compatibility issue for the moment - do you think this is a good idea?

olefrost:
@elek - I was hoping to see some analysis, investigation, and feedback on the proposed 4th option before seeing a pull request!

I was also hoping for an answer to [my forum questions and](https://forum.rclone.org/t/backward-compatibility-rules/26569/5) and some convincing performance data to support your choice.

elek:
> I was hoping to see some analysis, investigation, and feedback on the proposed 4th option before seeing a pull request!

Sure, let's continue the discussion on the forum. I changed the PR to draft to make this state clear.

I was interested if it's possible to organize the code in a way to maintain only in one place but -- for compatibility reasons -- keep the tardigrade backend, too. And when I realized that it's possible, I just uploaded it.

But let's call it as an illustration of one of the potential option.

calebcase:
> @elek - I was hoping for an issue to discuss this on rather than a pull request!
> 
> @calebcase - ignoring the backwards compatibility issue for the moment - do you think this is a good idea?

:+1: This is a great idea

ivandeex:
Great idea and perfect code reuse!

elek:
Closing the PR as @ncw had a better approach (with directly supporting aliases for each backend).

Issue #5598, opened 2021-09-13
Request to have manual configuration of download URL for S3


waranlogesh:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/sync-objects-from-s3-to-wasabi-using-cloudfront-url/26465

#### What is your current rclone version (output from `rclone version`)?
v1.56.0

#### What problem are you are trying to solve?
Egress from AWS S3 is costly. But if the same is accessed via CloudFront URL the cost is cheaper. So if we are able to provide a custom URL for download like in #2961 it would be really helpful to reduce the S3 egress cost when using rclone.

#### How do you think rclone should be changed to solve that?
I am not exactly sure on how to solve that but I think it can be done like #2961. Am ready to add the change if someone could point me in the right direction.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
Seems like a reasonable idea!

Do you want to make a patch @waranlogesh ?

waranlogesh:
> Seems like a reasonable idea!
> 
> Do you want to make a patch @waranlogesh ?

Yes, I would love to. Any pointers on where to start?

ncw:
I just wanted to check this will actually be cheaper....

From https://www.cloudforecast.io/blog/aws-cloudfront-pricing-and-cost-guide/ 

> It’s important to realize that CloudFront has a special status in the AWS universe. Normally, you’re charged for data transferred between availability zones. However, data transferred from AWS services to CloudFront servers doesn’t cost a cent.

So that is a validation of the idea.

> Yes, I would love to. Any pointers on where to start?

You'd need to add the option like this

https://github.com/rclone/rclone/blob/18f39291863e3d57dad84d306771716fae0cf125/backend/b2/b2.go#L152-L162

Then stick it in `Options`

https://github.com/rclone/rclone/blob/18f39291863e3d57dad84d306771716fae0cf125/backend/b2/b2.go#L209

Then in here

https://github.com/rclone/rclone/blob/18f39291863e3d57dad84d306771716fae0cf125/backend/s3/s3.go#L2961-L2962

You want to check if `DownloadURL` is set and jump off to a new method on Object.

This should be implemented something like

[...]

PR #5619, opened 2021-09-24
s3: add support to use CDN URL to download the file

waranlogesh:

The egress charges while using a CloudFront CDN url is cheaper when
compared to accessing the file directly from S3. So added a download
URL advanced option, which when set downloads the file using it.

The issue was discussed in #5598

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1296,6 +1296,12 @@ disabled here.  When the issue is solved this flag will be removed.
 See: https://github.com/rclone/rclone/issues/4673, https://github.com/rclone/rclone/issues/3631
 
 `,
+		}, {
+			Name: "download_url",
+			Help: `Custom endpoint for downloads.
+This is usually set to a CloudFront CDN URL as AWS S3 offers
+cheaper egress for data downloaded through the CloudFront network.`,
+			Advanced: true,
 		},
 		}})
 }
@@ -1357,6 +1363,7 @@ type Options struct {
 	MemoryPoolFlushTime   fs.Duration          `config:"memory_pool_flush_time"`
 	MemoryPoolUseMmap     bool                 `config:"memory_pool_use_mmap"`
 	DisableHTTP2          bool                 `config:"disable_http2"`
+	DownloadURL           string               `config:"download_url"`
 }
 
 // Fs represents a remote s3 server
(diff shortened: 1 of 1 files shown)


waranlogesh:
@ncw Please review

backend/s3/s3.go
```diff
@@ -2958,9 +2967,32 @@ func (o *Object) Storable() bool {
...
+		Options: options,
+	}
+	err = o.fs.pacer.Call(func() (bool, error) {
+		resp, err = o.fs.srvRest.Call(ctx, &opts)
+		return o.fs.shouldRetry(ctx, err)
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Body, err
```
ncw, line 2985: This should be setting the metadata as is done here if possible.

https://github.com/rclone/rclone/blob/bb0c4ad2d8af8af1ff9abea54dbdf97eadc9018f/backend/s3/s3.go#L3013-L3029

I don't know whether the metadata is returned when objects are fetched via the CDN - I need your advice there!

ncw (commented):
This looks good thank you.

I have a query about the metadata inline.

Thanks

backend/s3/s3.go
waranlogesh: Thank you for the review.

I just checked out the response headers from CDN. It does contain the metadata. Will update the PR.

```
HTTP/1.1 200 OK
Content-Type: image/jpeg
Content-Length: 3968662
Connection: keep-alive
Date: Sat, 02 Oct 2021 10:57:28 GMT
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, PUT, POST, DELETE, HEAD
Access-Control-Expose-Headers: ETag
Last-Modified: Sat, 18 Apr 2020 16:41:17 GMT
ETag: "08bfadc133h323f6663f434ecc9743"
x-amz-storage-class: STANDARD_IA
x-amz-version-id: null
Accept-Ranges: bytes
Server: AmazonS3
X-Edge-Origin-Shield-Skipped: 0
Vary: Access-Control-Request-Headers,Access-Control-Request-Method
X-Cache: Miss from cloudfront
Via: 1.1 dsfefefewefwfw.cloudfront.net (CloudFront)
X-Amz-Cf-Pop: MAA50-C1
X-Amz-Cf-Id: fdfdfsd_wWfsfsfksksks==
```

rtsainear:
@waranlogesh I am trying to use rclone with my cloutfront + s3. Do you mind giving me an example on how to use the cloudfront url?

I set my rclone.conf like following:
[far_cf]
type = s3
provider = AWS
download_url = https://xxxxxx.cloudfront.net/backups/

rclone copy far_cf:/file/jar/latest ./

but I get an error like its trying to download from s3 bucket

waranlogesh:
@rtsainear Ideally your command should work. One assumption for this to work is that your cloudfront distribution should have access to the s3 bucket without any signing needed.

You could also enable verbose logging with option `rclone -vv ` and add your logs so I could better understand the issue.

rtsainear:
@waranlogesh nm I figured it out. Thanks!

Issue #4101, opened 2020-04-01
anacrolix/dms license violation


anacrolix:
I recently received https://github.com/anacrolix/dms/issues/73#issuecomment-606762705 to anacrolix/dms, and in following the link to https://github.com/rclone/rclone/blob/66e08e0cc8972000d84a61f262286f5644d9fdee/cmd/serve/dlna/cds.go found the code to be remarkably familiar. I see that this was copied in https://github.com/rclone/rclone/pull/2648. I believe the BSD 3-Clause license allows modifications, but requires a copyright notice whether modifications are made or not, and there does not appear to be any such license notice in rclone. The changes since made look very positive, and it would be great if an attempt was made to upstream those, as there is significant overlap with issues currently reported against anacrolix/dms, and the internals which are currently copied and modified, and still linked into are intended to be used more broadly such as they are in rclone. Thanks!

ncw:
Thanks for pointing out the licence oops! Will fixup ASAP!

@nicolov I think this was imported by you originally wasn't it?

@sretlawd I think you've made improvements to this - what do you think about upstreaming them?

Thanks

nicolov:
Sorry for the license oversight, and thanks a lot for your efforts.

Yes, I did import the code originally to get streaming to work, but haven't made any of the improvements. Can I help out somehow?

sretlawd:
Happy to send up a patch.  I don't currently have access to any Smart TVs, and no idea when I'll be going home, so I can't really do much at the moment..

anacrolix:
@sretlawd I don't really have access either, I expect it should be pretty straightforward to upstream many of your changes, the code hasn't changed much, if at all in many areas and the layout remains the same.

anacrolix:
@ncw could you link to any license/copyright that has been included as a result of this issue?

The anacrolix/dms project has recently received a lot of interest and user uptake. It's also now packaged by numerous Linux distros. It would still be great if many of the changes made here were upstreamed so that both projects can benefit.

ivandeex:
@anacrolix @nicolov @sretlawd 
Any updates?

anacrolix:
This is not waiting for upstream. I am waiting for this project to review the licensing, and/or upstream or contribute some changes back.

ivandeex:
As PR authors do not seem much interested in responding, I made a look into rclone git history.

- Our use of anacrolix/dms started from PR #2648, which was merged into master by on 2019-01-09 and released with rclone v1.46
[...]

PR #5622, opened 2021-09-24
serve: do not override mime types from os

albertony:

Fixes issue with `serve http` and audio files of `.m4a` extension. If selecting "save as" in browser it would default to `.mp3` extension.

Problem is the change introduced here: https://github.com/rclone/rclone/commit/67fae720d715a792dd8b5aa0f02332d2b21d393b. This overrides the mime type for the matching extensions in golang's mimelist, which has been initialized from os - on Windows the registry. So in the example above, I have in my Windows registry mime type `audio/mp4` for file extension `.m4a`, and on startup golangs' mime list is initialized with this, but this is later replaced with `audio/mpeg` when it comes to the dlna packages init.

https://forum.rclone.org/t/rclone-serve-http-save-as/26672

#### Checklist

- [ ] ~I have added documentation for the changes if appropriate.~


ivandeex:
Relates to https://github.com/rclone/rclone/issues/4101#issuecomment-817103806

ncw (commented):
The code looks great thanks.

The commit message the title should be `serve dlna: do not....` - this doesn't affect all the servers.

Is there an issue or a forum post to link to in the commit message

albertony:
> The commit message the title should be serve dlna: do not.... - this doesn't affect all the servers.

Well it did at least affect http, which was the original issue! And I thought that the cmd/serve imports all serve sub-packages, and therefore will run their init, which means the cds.go init will always be run on any `rclone serve`, but I could definitely be wrong.. :)

albertony:
> > The commit message the title should be serve dlna: do not.... - this doesn't affect all the servers.

> Well it did at least affect http, which was the original issue! And I thought that the cmd/serve imports all serve sub-packages, and therefore will run their init, which means the cds.go init will always be run on any rclone serve, but I could definitely be wrong.. :)

.. But I am not! 😋 Although not all serve commands use mime types for anything, only dlna and http do, I think. But it is even *worse*: Its not just `serve` that is affected, *all* commands are.

Tested with `rclone lsl`, on folder containing .m4a file.
- Existing version (before this PR): lists it with:
    `"MimeType":"audio/mpeg"`
- This PR version lists it with:
    `"MimeType":"audio/mp4"`

So, I'm wondering even more:  Should we move this mime initialization code, from cds.go's init? It feels like a relatively "big" change to hide away deep into this package's init.. If you see what I mean. Don't know if there are any good candiates, central inits somewhere? Alternatively could make it "delay-load" to only affect the serve dlna and http features?

albertony:
> Is there an issue or a forum post to link to in the commit message

There is the forum post in the original PR comment, that is directly related - and the originating "issue". I did not realize we wanted forum links in comments, but I can definitely do that.

ncw:
> > > The commit message the title should be serve dlna: do not.... - this doesn't affect all the servers.
> 
> > Well it did at least affect http, which was the original issue! And I thought that the cmd/serve imports all serve sub-packages, and therefore will run their init, which means the cds.go init will always be run on any rclone serve, but I could definitely be wrong.. :)

Ah, I completely missed the fact it was done in the `init()` and thus affecting the global mime types registry!
 
> .. But I am not! yum Although not all serve commands use mime types for anything, only dlna and http do, I think. But it is even _worse_: Its not just `serve` that is affected, _all_ commands are.

Yes they are, anything which calls `fs.MimeType` which is nearly everything!
 
> So, I'm wondering even more: Should we move this mime initialization code, from cds.go's init? It feels like a relatively "big" change to hide away deep into this package's init.. If you see what I mean. Don't know if there are any good candiates, central inits somewhere? Alternatively could make it "delay-load" to only affect the serve dlna and http features?

I think if it is a good idea for the http server then it is a good idea for everything.

I'd suggest moving it into `fs/mimetype.go` - we have `fs.MimeType` there already with some related files so putting it there will make it easy to find.

albertony:
> I think if it is a good idea for the http server then it is a good idea for everything.

Ok, yes, I can agree.

> I'd suggest moving it into fs/mimetype.go - we have fs.MimeType there already with some related files so putting it there will make it easy to find.

Sounds like the perfect place!

Have force pushed the move.
Considering adding a unit test, but the fact that the changes are in init func makes it difficult.. Not the most risky code change, so not crucial I guess..

PR #5629, opened 2021-09-25
docs: cleanup header levels

albertony:

Some cleanup in the usage, install and download docs. Mainly related to headers.

Ensure each page starts with a level 1 header (`# header`), and then only use level 2 or higher for sections below. E.g. existing install doc uses header level 1 for "Autostart", above [Autostart on Windows](https://rclone.org/install/#autostart-on-windows), but then this does not get a anchor on hover-over and also is not listed in the contents box to the right. And existing downloads doc have several level 1 headers but only one level 2 header, and then only the level 2 header "Script download and install" is shown in the contents box to the right.

The [usage](https://rclone.org/docs/) docs started with level 2 header called "Configure". So for consistency I introduced a level 1 header with the page name, "Usage", and *tried* to come up with something short to write there to avoid having an empty level 1 header followed only by a level 2 header.
Edit: I see that existing Changelog and Privacy Policy pages does have an empty level 1 header followed by a level 2 header. It does not look too bad, so I can just do the same in Usage. One thing is that the usage site is at the /docs url, so feels like som sort of main/entry point for documentation. So I tried to write a short, generic intro to rclone doc. But open for suggestions to do this differently.

#### Checklist

### DIFF
--- a/docs/content/authors.md
+++ b/docs/content/authors.md
@@ -3,6 +3,8 @@ title: "Authors"
 description: "Rclone Authors and Contributors"
 ---
 
+# Authors and contributors
+
 Authors
 -------
 
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -3,6 +3,15 @@ title: "Documentation"
 description: "Rclone Usage"
 ---
 
+# Usage
+
+Rclone is a command line program to manage files on cloud storage.
+After [download](/downloads/) and [install](/install), continue
+here to learn how to use it: Initial [configuration](#configure),
+what the [basic syntax](#basic-syntax) looks like, describes the
+various [subcommands](#subcommands), the various [options](#options),
+and more.
+
 Configure
 ---------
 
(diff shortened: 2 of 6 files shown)


ncw:
... I presume you've noticed that level 2 and 3 headers go in the contents box on the RHS of the web page?

albertony:
> ... I presume you've noticed that level 2 and 3 headers go in the contents box on the RHS of the web page?

Yes, exactly. Take existing https://rclone.org/downloads/ page for example: It contains several headers in the page, but only one entry in the RHS box: `Script download and install`. Because the headers are level 1. I changed that to have only one level 1 header at the top, and made the rest level 2++.

ncw (approved):
This looks great - thank you - please merge :-)

I imagine there is more of this to come!

albertony:
> I imagine there is more of this to come!

By the way: Do you @ncw have an idea why there are no contents box at all on this page: https://rclone.org/bugs/

Issue #1234, opened 2017-03-12
[Mount] can't rename non empty directory


Powerbless:
Hello.

Using newest Rclone Beta under Linux Alpine 64bit

Command:
rclone mount --allow-non-empty --allow-other --checkers 10 encrypt: /med
ia/EACD/ &

when i try to RENAME a folder then linux print this::

`2017/03/12 17:28:36 _Kodi/SD Filme/Der Prinz hinter den sieben Meeren (1982).(tt0084535).SDp..mp2.2: Dir.Rename can't rename non empty directory
2017/03/12 17:28:36 _Kodi/SD Filme/Der Prinz hinter den sieben Meeren (1982).(tt0084535).SDp..mp2.2: Dir.Rename can't rename non empty directory`

i think there is a problem with the Folder name? because when i try to rename Folder "Film.1990" to "Film.1990a" works fine. EDIT: but after a refresh, the folder is renamed to the old name....

ncw:
Can you check you are using the latest beta please - that error text doesn't exist in the code any more.

Powerbless:
I have again downloaded the latest BETA.
the renaming and deleting now WORK!

my steps: (for other users who have the problem)
after rclone mount command run this:
chmod 777 /media/eacd/

and in smb.conf i added:
[global]
dos charset = cp866
unix charset = UTF8
display charset = UTF8

now, renaming, deleting and all the other works

ncw:
Excellent :-)

Issue #5627, opened 2021-09-25
Jottacoud documentation of modtime support


albertony:
On https://rclone.org/overview/ Jottacloud is listed with `ModTime: Yes`.

Below, `ModTime` is defined:

> The cloud storage system supports setting modification times on objects. If it does then this enables a using the modification times as part of the sync. If not then only the size will be checked by default, though the MD5SUM can be checked with the --checksum flag.

> All cloud storage systems support some kind of date on the object and these will be set when transferring from the cloud storage system.

Now from the code we can see that the backend does not support the operation of setting the modtime-only, but it does store modtimes. So modtimes can be read, and used to compare files during sync, but to update them the file must be re-uploaded.

https://github.com/rclone/rclone/blob/1409b89f6c07d80bdfa880ad018a4a7771c049f1/backend/jottacloud/jottacloud.go#L1394-L1410

Now, does this mean the `ModTime: Yes` entry wrong, should it be `ModTime: No` in docs? Please enlighten me! :)

(I've investigated Jottacloud in this case, but I see similar situation for Dropbox, pCloud and Tardigrade).

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

albertony:
This is expected to be clearified with: https://github.com/rclone/rclone/issues/1066

Issue #1066, opened 2017-01-25
Documentation: Expand storage systems capabilities tables


eharris:
Please expand the capabilities tables as listed on http://rclone.org/overview/ to:

 * Add a column for DirModTime support
 * Add a column for DirCopy support (DirMove is already present)
 * Add a column for indicate support of some form of arbitrary metadata support

Thanks.

ncw:
> Add a column for DirModTime support

Do you mean remotes where rclone will read the dir mod time if it is available (like drive, acd but not s3, swift)?

> Add a column for DirCopy support (DirMove is already present)

I don't think anything supports `DirCopy` - at least rclone doesn't support calling it.  There is a `DirMove` column.

> Add a column for Folder support (see #100)

What do you mean by Folder support?  Remotes which actually support the creation of folders like acd, google drive but not s3, swift?  This is probably the same list as DirModTime if I'm understanding you correctly.

> Add a column for indicate support of some form of arbitrary metadata support

Rclone doesn't support any sort of arbitrary metadata at the moment.  Are you wondering what the remotes could support in the future?

eharris:
For DirModTime, I mean exactly the same as the already present ModTime column, but for directories rather than files; i.e. that the storage method understands folders and supports setting of modtime for them.

OpenDrive (#1026) supports DirCopy, hence the request to add the column.  I realize that OpenDrive support isn't available yet, but I'm hoping it will happen soonish.  I didn't realize none of the others supported it, but perhaps those need some investigation to verify.

By folder support, yes, I mean to mark which remotes natively understand the concept of folders (and operations on them like mkdir/rmdir/set modtime, as discussed in the linked ticket).

Yes, I realize that rclone doesn't support arbitrary metadata yet, but I believe at least ACD supports it (according to [this comment](https://github.com/ncw/rclone/issues/371#issuecomment-188724500) on #371) and there's already another ticket asking for support of it (#111), and I hope that support for using backend-specific metadata capabilities within rclone will be available at some point in the not too distant future.  Having those capabilities listed would be helpful for future implementation plans.

ncw:
OK I'll see what I can do.  Any help much appreciated so any answers you can find - please place in the ticket.

The metadata bit might be complicated - there are restrictions, etc.  But the total length of the metadata supported would be a useful thing to know - and if that would fit an encrypted name+hash then that solves  a lot of the crypt problems.

eharris:
[...]

PR #5634, opened 2021-09-26
jottacloud: implement SetModTime to support modtime-only changes

albertony:

Implementing SetModTime function of Jottacloud backend, to add support for `rclone touch` on existing file, modtime-only changes of files in mounts etc.

Note: With the existing backend implementation's use of the allocate API, with its deduplication/resume features, whenever rclone wants to upload an identical file with changed timestamp, then it will be reduced to a "SetModTime" operation anyway: The allocate request updates the timestamp and returns that file is already COMPLETED, no need to upload data with a following update request. So in that sense, adding support for SetModTime does not change anything for regular sync/copy operations. What it does do is add support for modtime-only operations that does not resort to "re-upload".

Implementation note: Updates both Created and Modified timestamps, which also the existing Update function does, because Jottacloud seems to always update both anyway. If not including a Created value in the allocate request, Jottacloud will automatically set one (current time when server processes the request?).

The following issues touches in on it, at least:

https://github.com/rclone/rclone/issues/5627
https://github.com/rclone/rclone/issues/1066

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -807,15 +807,8 @@ func (f *Fs) newObjectWithInfo(ctx context.Context, remote string, info *api.Jot
 	}
 	var err error
 	if info != nil {
-		if info.State != "COMPLETED" {
-			return nil, fs.ErrorObjectNotFound // File is incomplete or corrupt
-		}
-		if o.fs.opt.TrashedOnly {
-			if !info.Deleted {
-				return nil, fs.ErrorObjectNotFound // Regular file but TrashedOnly
-			}
-		} else if bool(info.Deleted) {
-			return nil, fs.ErrorObjectNotFound // Deleted file and not TrashedOnly
+		if !f.validFile(info) {
+			return nil, fs.ErrorObjectNotFound
 		}
 		err = o.setMetaData(info) // sets the info
 	} else {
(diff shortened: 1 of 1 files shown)


backend/jottacloud/jottacloud.go
```diff
@@ -1406,7 +1406,46 @@ func (o *Object) ModTime(ctx context.Context) time.Time {
...
+		Bytes:    o.size,
+		Created:  fileDate,
+		Modified: fileDate,
+		Md5:      o.md5,
+		Path:     path.Join(o.fs.opt.Mountpoint, o.fs.opt.Enc.FromStandardPath(path.Join(o.fs.root, o.remote))),
+	}
+
+	// send it
+	var response api.AllocateFileResponse
+	err = o.fs.pacer.CallNoRetry(func() (bool, error) {
```
ncw, line 1438: Why is this `NoRetry`? We generally use that for uploads which read from an `io.Reader` which can't be rewound.

ncw (commented):
Very nice!

Can you reference the relevant issues in your commit message please? If it completes them then `Fixes #1234` otherwise `#1234`.

Now for the important question...

Do the integration tests pass with your change?

backend/jottacloud/jottacloud.go
albertony: No idea! I copied it from the existing upload implementation without thinking about this. I guess I should just do `Call`. Will fix.

albertony: Done. Also removed the `response api.AllocateFileResponse` object, as it was not checked anyway, sending nil as argument saves a json parsing I guess.

albertony:
> Do the integration tests pass with your change?

Yes. Ran through some fs tests as well, operations and sync. Get some errors (TestCopyFile, TestCopyFileCopyDest, TestSyncCopyDest), but identical results on master.

I think this should be good to go now.

backend/jottacloud/jottacloud.go
albertony: I think I will re-add the response, and instead actually check it. I managed to provoke an error that went unnoticed, although not very likely in real life cases, its probably better safe than sorry.

ncw (approved):
I think this looks good to go now.

Please merge when ready - thank you :-)

backend/jottacloud/jottacloud.go
```diff
@@ -1406,7 +1406,45 @@ func (o *Object) ModTime(ctx context.Context) time.Time {
...
-	return fs.ErrorCantSetModTime
+	// make sure metadata is available, we need its current size and md5
+	err := o.readMetaData(ctx, false)
+	if err != nil {
+		fs.Logf(o, "Failed to read metadata: %v", err)
+		return err
+	}
+
+	// prepare allocate request with existing metadata but changed timestamps
+	o.fs.tokenRenewer.Start()
```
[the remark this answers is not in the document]
albertony: Thanks for the info. I decided to remove it.

PR #5635, opened 2021-09-26
docs: add faq section explaining why rclone changes fullwidth characters in file names

albertony:

Title says it all.

https://forum.rclone.org/t/cannot-correctly-handle-filenames-with-u-ff1a-full-width-colon/26698

#### Checklist

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -211,38 +211,20 @@ directory with thousands or millions of files in.  Rclone has to load
 this entirely into memory as rclone objects.  Each rclone object takes
 0.5k-1k of memory.
 
-### Rclone changes fullwidth characters in file names
+### Rclone changes fullwidth Unicode punctuation marks in file names
 
-On a Windows system, you have a file with name `C：Backup.txt`,
-where `：` is the Unicode fullwidth colon symbol, and use rclone to
-copy this to your Google Drive with rclone. Then you will notice that
-it in Google Drive it has been renamed to `C:Backup.txt`, where `:` is
-the regular (halfwidth) colon.
+For example: On a Windows system, you have a file with name `Test：1.jpg`,
+where `：` is the Unicode fullwidth colon symbol. When using rclone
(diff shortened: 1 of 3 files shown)


docs/content/faq.md
```diff
@@ -210,3 +210,17 @@ The most common cause of rclone using lots of memory is a single
 directory with thousands or millions of files in.  Rclone has to load
 this entirely into memory as rclone objects.  Each rclone object takes
 0.5k-1k of memory.
+
+### Rclone changes fullwidth characters in file names
+
+To handle varying sets of [restricted filenames](/overview/#restricted-filenames)
+on different backends, rclone transparently converts between restricted
+characters and similar looking Unicode characters. If you have file names
```
ncw, line 218: I think we should have an example here with `:` and its unicode equivalent.

ncw (commented):
Very nice docs - thank you :-)

I think we need an example in the FAQ, but otherwise perfect!

albertony:
New attempt! But I'm getting quite 😵 , trying to write explanation in one place (e.g. the faq), then improve some another place (e.g. in the overview), just to end up writing almost the same (and too long) section both places😫 ! But this encoding stuff can be quite confusing to users - and me, so I feel if I try to keep it short and concise then there are nuances that get lost that may well be exactly the thing that made it blow up for some user..

Well... Have a read, and let me know if I should rewrite, restructure or fix any nitpicks, and I will take another round 👨‍🌾

albertony:
Given the new information, I "promoted" the detailed explanation originally written in faq into a new section `Caveats` in the overview page. I also moved and merged large sections from the `Encoding example: Windows`. I think this is better - less duplication. But perhaps the "negatively charged" `Caveats` section now gets *too* much attention? What do you think, @ncw?

ncw:
...I think I'm going to merge this for 1.57 as it is a great improvement. We can do further followups to tweak stuff :-)

Issue #5616, opened 2021-09-22
Rename tardigrade->storj and clarify s3/native backend use cases


elek:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/backward-compatibility-rules/26569/12

#### What is your current rclone version (output from `rclone version`)?

```
clone v1.57.0-beta.5661.1dd360a01
- os/version: arch (64 bit)
- os/kernel: 5.14.2-arch1-2 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17
- go/linking: dynamic
- go/tags: none
```

#### What problem are you are trying to solve?

Tardigrade was renamed to Storj (or Storj DCS) [recently](https://storj.io/blog/introducing-storj-dcs-decentralized-cloud-storage-for-developers).

To avoid confusion it seems to be better to do the rename on rclone side too.

During the forum discussion, it was suggested to explain better the advantages and disadvantages of using s3/native storj backend as storj also support s3 interface via `gateway-mt`.

S3 gateway can be useful for edge clients as less CPU/bandwidth is required, however, these CPU/bandwidth are currently provided by Storj Labs Inc (for free as of today). 

Native tardigrade/storj backend uses slightly more CPU and bandwidth (for upload) but it uses the decentralized cloud of storage nodes and it's more secure (as the secret credential is not shared with central service).

@olefrost suggest the following action points in the forums:

 * [x] Make a GitHub issue to discuss the overall approach and plan
 * [x] Confirm full Storj S3 compatibility by running the automated S3 backend test against an S3 TestDrive having Storj as an endpoint.
 * [ ] Make a pull request to Add Storj to the list of S3 providers on [this page](https://github.com/rclone/rclone/blob/master/docs/content/s3.md)
 * [x] Make a performance comparison of the S3 and tardigrade backends to collect information on the expected resource usage and speed on representative use cases using identical equipment.
 * [ ] Make a pull request to update the tardigrade (and S3) backend documentation with this information.
  * [ ] Make a pull request to rename the tardigrade backend to Storj - or to depreciate it in favor of Storj’s own client (https://www.storj.io/integrations/uplink-cli). The approach is to be based on the results and agreements obtained in the previous steps.
 
#### How do you think rclone should be changed to solve that?

The rename can be done with a backward-compatible way (POC is in #5613) with keeping the tardigrade backend as a pointer to the storj.

Documentation should be added to clarify s3 compatibility...

[...]

PR #5646, opened 2021-09-30
s3: add specific provider for Storj Shared gateways

elek:

Make it easier (and documented) to use Storj distributed cloud storage via the s3 interface.

  1. Both on the Tardigrade and S3 backend it's clarified what are the pros and cons of using them.
  2. Storj added as a special s3 provider (with the predefined URL) 

The discussion started here: https://forum.rclone.org/t/backward-compatibility-rules/26569/12
And later moved to #5616

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -103,6 +103,9 @@ func init() {
 				Value: "StackPath",
 				Help:  "StackPath Object Storage",
 			}, {
+				Value: "Storj",
+				Help:  "Storj (S3 Compatible Gateway)",
+			}, {
 				Value: "TencentCOS",
 				Help:  "Tencent Cloud Object Storage (COS)",
 			}, {
@@ -288,7 +291,7 @@ func init() {
 		}, {
 			Name:     "region",
 			Help:     "Region to connect to.\n\nLeave blank if you are using an S3 clone and you don't have a region.",
-			Provider: "!AWS,Alibaba,RackCorp,Scaleway,TencentCOS",
+			Provider: "!AWS,Alibaba,RackCorp,Scaleway,Storj,TencentCOS",
 			Examples: []fs.OptionExample{{
 				Value: "",
 				Help:  "Use this if unsure.\nWill use v4 signatures and an empty region.",
@@ -598,6 +601,20 @@ func init() {
 				Help:  "EU Endpoint",
 			}},
 		}, {
+			Name:     "endpoint",
+			Help:     "Endpoint of the Shared Gateway.",
+			Provider: "Storj",
+			Examples: []fs.OptionExample{{
+				Value: "gateway.eu1.storjshare.io",
+				Help:  "EU1 Shared Gateway",
+			}, {
+				Value: "gateway.us1.storjshare.io",
+				Help:  "US1 Shared Gateway",
+			}, {
+				Value: "gateway.ap1.storjshare.io",
+				Help:  "Asia-Pacific Shared Gateway",
+			}},
+		}, {
 			// cos endpoints: https://intl.cloud.tencent.com/document/product/436/6224
 			Name:     "endpoint",
 			Help:     "Endpoint for Tencent COS API.",
@@ -726,7 +743,7 @@ func init() {
 		}, {
 			Name:     "endpoint",
 			Help:     "Endpoint for S3 API.\n\nRequired when using an S3 clone.",
-			Provider: "!AWS,IBMCOS,TencentCOS,Alibaba,Scaleway,StackPath,RackCorp",
+			Provider: "!AWS,IBMCOS,TencentCOS,Alibaba,Scaleway,StackPath,Storj,RackCorp",
 			Examples: []fs.OptionExample{{
 				Value:    "objects-us-east-1.dream.io",
 				Help:     "Dream Objects endpoint",
@@ -1014,7 +1031,7 @@ func init() {
 		}, {
 			Name:     "location_constraint",
 			Help:     "Location constraint - must be set to match the Region.\n\nLeave blank if not sure. Used when creating buckets only.",
-			Provider: "!AWS,IBMCOS,Alibaba,RackCorp,Scaleway,StackPath,TencentCOS",
+			Provider: "!AWS,IBMCOS,Alibaba,RackCorp,Scaleway,StackPath,Storj,TencentCOS",
 		}, {
 			Name: "acl",
 			Help: `Canned ACL used when creating buckets and storing or copying objects.
@@ -1025,6 +1042,7 @@ For more info visit https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview
 
 Note that this ACL is applied when server-side copying objects as S3
 doesn't copy the ACL from the source but rather writes a fresh one.`,
+			Provider: "!Storj",
 			Examples: []fs.OptionExample{{
 				Value:    "default",
 				Help:     "Owner gets Full_CONTROL.\nNo one else has access rights (default).",
@@ -1936,6 +1954,11 @@ func setQuirks(opt *Options) {
 		listObjectsV2 = false // untested
 		virtualHostStyle = false
 		urlEncodeListings = false
+	case "Storj":
+		// Force chunk size to >= 64 MiB
+		if opt.ChunkSize < 64*fs.Mebi {
+			opt.ChunkSize = 64 * fs.Mebi
+		}
 	case "TencentCOS":
 		listObjectsV2 = false // untested
 	case "Wasabi":
@@ -2069,6 +2092,11 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		}
 		// return an error with an fs which points to the parent
 		return f, fs.ErrorIsFile
+	}
+	if opt.Provider == "Storj" {
+		f.features.Copy = nil
+		f.features.SetTier = false
+		f.features.GetTier = false
 	}
 	// f.listMultipartUploads()
 	return f, nil
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -19,6 +19,7 @@ The S3 backend can be used with a number of different providers:
 {{< provider name="Scaleway" home="https://www.scaleway.com/en/object-storage/" config="/s3/#scaleway" >}}
 {{< provider name="SeaweedFS" home="https://github.com/chrislusf/seaweedfs/" config="/s3/#seaweedfs" >}}
 {{< provider name="StackPath" home="https://www.stackpath.com/products/object-storage/" config="/s3/#stackpath" >}}
+{{< provider name="Storj" home="https://storj.io/" config="/s3/#storj" >}}
 {{< provider name="Tencent Cloud Object Storage (COS)" home="https://intl.cloud.tencent.com/product/cos" config="/s3/#tencent-cos" >}}
 {{< provider name="Wasabi" home="https://wasabi.com/" config="/s3/#wasabi" end="true" >}}
 {{< /provider_list >}}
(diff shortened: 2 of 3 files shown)


ncw (commented):
This is a great idea thanks.

I don't think the integration tests pass with this though do they?

Maybe we need to mask off the `Copy` feature flag and any other features which aren't implemented by Storj S3?

I'm not concerned about the file names which can't be uploaded.

elek:
> I don't think the integration tests pass with this though do they?

Good question. Let's record here the current state of integration test runs:

```
cd bacend/s3
go test -v -remote storjs3local: -fast-list -json | tee test2.json
jq '. | select(.Action=="fail") | .Test' test.json 

"TestIntegration/FsMkdir/FsEncoding/control_chars"
"TestIntegration/FsMkdir/FsEncoding/punctuation"
"TestIntegration/FsMkdir/FsEncoding/leading_LF"
"TestIntegration/FsMkdir/FsEncoding/leading_VT"
"TestIntegration/FsMkdir/FsEncoding/trailing_LF"
"TestIntegration/FsMkdir/FsEncoding/trailing_VT"
"TestIntegration/FsMkdir/FsEncoding"
"TestIntegration/FsMkdir/FsPutFiles/ObjectSetModTime"
"TestIntegration/FsMkdir/FsPutFiles/SetTier"
"TestIntegration/FsMkdir/FsPutFiles/FsPutStream/0"
"TestIntegration/FsMkdir/FsPutFiles/FsPutStream/100"
"TestIntegration/FsMkdir/FsPutFiles/FsPutStream"
"TestIntegration/FsMkdir/FsPutFiles"
"TestIntegration/FsMkdir/FsPutChunked/5Mi/5242880"
"TestIntegration/FsMkdir/FsPutChunked/5Mi/10485761"
"TestIntegration/FsMkdir/FsPutChunked/5Mi"
"TestIntegration/FsMkdir/FsPutChunked/5.054Mi/5300000"
"TestIntegration/FsMkdir/FsPutChunked/5.054Mi/10600001"
"TestIntegration/FsMkdir/FsPutChunked/5.054Mi"
"TestIntegration/FsMkdir/FsPutChunked/5.055Mi/5300053"
"TestIntegration/FsMkdir/FsPutChunked/5.055Mi/10600107"
"TestIntegration/FsMkdir/FsPutChunked/5.055Mi"
"TestIntegration/FsMkdir/FsPutChunked/6.667Mi/6990506"
"TestIntegration/FsMkdir/FsPutChunked/6.667Mi/13981013"
"TestIntegration/FsMkdir/FsPutChunked/6.667Mi"
"TestIntegration/FsMkdir/FsPutChunked/8Mi/8388608"
"TestIntegration/FsMkdir/FsPutChunked/8Mi/16777217"
"TestIntegration/FsMkdir/FsPutChunked/8Mi"
"TestIntegration/FsMkdir/FsPutChunked/10Mi/10485760"
"TestIntegration/FsMkdir/FsPutChunked/10Mi/20971521"
"TestIntegration/FsMkdir/FsPutChunked/10Mi"
"TestIntegration/FsMkdir/FsPutChunked"
"TestIntegration/FsMkdir"
"TestIntegration"
```

These are actually two problems:

 * Storj naming scheme is slightly more string (`FsEncoding`)
 * object metadata modifications are not supported in multipart upload. This is missing from the native protocol of Storj and tracked by storj/gateway-mt#39 Not an easy fix and this is responsible for all the FsPutChunked/FsPutFiles errors
 * ObjectSetModTime: It's the same for MPU will check if it requires a different fix for simple upload or not
 
 `SetTier` is not supported, this is the only one which can be turned off with features. I did it in 4daefaf294fc6c73176f7d003546eadcd3c7f9ad (thanks for the suggestion)

I don't see any copy test failed, but (just for the safety) I turned it off as well...

elek:
FTR: three encoding problems were fixed by eb3d8ff. `urlEncodeListings` is enabled for `StorJ` just like for minio.

Still some characters are not handled, though...

olefrost:
I also like the approach and addition of Storj to the list of S3 backends. 
 
I strongly recommend that the list of differences is expanded to give more details and facts. The current "use … to …" statements can easily be misunderstood and thereby lead to disappointed and frustrated rclone/Storj users.
 
The main issue is the implicit/hidden assumption, that Storj users have high-end (=pro) equipment with sufficiently free resources and internet bandwidth to honor the multipliers of the Stoj protocol without costs or delays. This makes the "use" statements wrong and misleading if the user has average equipment or network connection, as we found in the performance tests.
 
I therefore recommend you make the "use" statements clearer by explicitly stating the requirements and suggest you base them on [the conclusions from the performance tests](https://github.com/rclone/rclone/issues/5616#issuecomment-929553323): (with better language and formatting)

> … the S3 protocol is the best and simplest way to interface to StorJ for the ordinary rclone user (the default so to speak). 
>  
> The Tardigrade/StorJ protocol is interface for pro users having the equipment, internet connection (above 300/300 Mbit/s) and technical skills to configure and troubleshoot such an environment, e.g. increase the maximum number of open files/tcp connections on the system (requiring root/admin access and skills) - and of cause anybody with an interest in the distributed security model of Storj.
> 
 
I also recommend that the download multiplier is mentioned to complete the picture and help users find their own sweet spot.
 
I have noted that the multipliers you are using deviates from the [end user experiences shared in the Storj forum](https://forum.storj.io/t/amazon-s3-vs-tardigrade/4083). Do you know why? A guess: Storj strarts upload/download to 50% more servers than needed by the erasure multiplier and stops all when enough have completed.  It would be nice to have this clarified and adjusted in the docs, if needed.
 
I also recommend that any functional differences between the S3 and tardigrade interface are mentioned. For example, the difference in hash verification of uploads/downloads (it is apparently missing in the tardigrade backend [according to the docs](https://rclone.org/overview/#features))

I also recommend that any limitations in the Storj support of the S3 interface are mentioned, e.g. the currently missing support of the copy feature, discussed above.
 
I finally recommend that the section briefly mentions that client-side encryption also is possible with the S3 backend if combined the [rclone crypt backend](https://rclone.org/crypt/).

elek:
Thanks, @olefrost, for very good suggestions. 

I am not sure how is it possible in this favor of Markdown but I am thinking about a side-by-side comparison with a table where all of the mentioned differences can be summarized. And after that we can explain the details in text...

Will update the PR with it.

> increase the maximum number of open files/tcp connections on the system (requiring root/admin access and skills) -

I am not particularly convinced about this small detail. I think most linux environments have enough tcp/file limits to use the native connector. But I do agree to emphasize it as a difference 

> TCP connections to be used:
>     S3: a few (depends on parallelism)
>    multiple dozens to hundreds (separated connections to each data host)

olefrost:
> ... I am thinking about a side-by-side comparison with a table where all of the mentioned differences can be summarized. And after that we can explain the details in text...

Good idea!

> I think most linux environments have enough tcp/file limits to use the native connector.

Well I am not expert in linux or Storj, but the clone docs say this under [tardigrade known issues](https://rclone.org/tardigrade/#known-issues): 

> … If you do many transfers you eventually open a connection to most storage nodes (thousands of nodes).

My Ubuntu 20.04 LTS (on WSL2) has a process limit of 1024 and requires root access to change it. My Synology NAS has the same limit. My Windows pc has a limit of 3977 (from googling). I don’t know if my boxes are representative, but I have noted that several users in the forum are using surprisingly small equipment like pi4 and even home routers. That is why I found it important to highlight.

elek:
3d8470e7eac96ea11b1107064e114279a3d19f3b: more information about the differences (and suggestion about choosing). I decided to use a list format as I explained it in more detail, not just with +/- signs.

> I also recommend that any functional differences between the S3 and tardigrade interface are mentioned. For example, the difference in hash verification of uploads/downloads (it is apparently missing in the tardigrade backend according to the docs)

This is the only one, that I couldn't address. I don't know any special difference, after all both s3 gateway and tardigrade backend use the same uplink lib / protocols.

olefrost:
Great, that’s a very good overview!

I only have a few comments to improve layout and spelling:

Your table mark down doesn’t render correctly at my end (it actually doesn’t make a table at all :) The markdown syntax shown in this link however seems to work: [https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables)

Spelling: RCP -> TCP (in the overview), with combining -> when combining (in security)

>... functional differences between the S3 and tardigrade interface  ... This is the only one, that I couldn't address. I don't know any special difference, after all both s3 gateway and tardigrade backend use the same uplink lib / protocols.

I guess I was a bit to brief in my comment, let me elaborate a bit.

When I look in the Hash column in this backend feature table: [https://rclone.org/overview/#features](https://rclone.org/overview/#features) then I can see that the S3 backend supports MD5 hashes, so upload and downloads are verified when performed from/to backends supporting MD5 - and data integrity can be quickly verified with `rclone check`.

When I look at the tardigrade backend then it doesn’t support any hash, not even a custom hash like Dropbox. This means that uploads cannot be verified by rclone and I cannot use `rclone check` to quickly verify that the 10GB file on my computer is the same as the one in Storj. I will have to do a complete download and comparison using `rclone check --download` or similar.

This an important difference between the two backends and something I think should go into the table too.

elek:
> Your table markdown doesn’t render correctly at my end (it actually doesn’t make a table at all :)

Because it's not a table ;-) 

As I wrote I decided to make a list instead. The table is good if it only contains +/-, but I thought that table with a lot of text is less useful. But I can reformat, if you think it's better. And it's harder to maintain in markdown.

> When I look in the Hash column...

Thanks to explain it. I was not aware of this difference. I added an additional point with "limitations" where I mention it.

Typos are fixed, too (thanks to report)

olefrost:
> Because it's not a table ;-)

Ahh, sorry!

I played a little with the formatting and have found a layout that I find easier to read - and more in line with the rest of the documentation. Below is an example with the first two sections. I also did some rephrasing ;-) Take it as inspiration and  use what you like:

* Characteristics:
  * *Tardigrade backend*: Uses native RPC protocol, connects directly to the storage nodes which hosts the data. Requires more CPU resource of encoding/decoding and has network amplification (especially during the upload), uses lots of TCP connections 
  * *S3 backend*: Uses S3 compatible HTTP Rest API via the shared gateways. There is no network amplification, but performance depends on the shared gateways and the secret encryption key should be shared with the gateway. 
* Typical usage:
  * *Tardigrade backend*: Server environments and desktops with enough resources, internet speed and connectivity - and applications where tardigrades client-side encryption is required.
  * *S3 backend*: Desktops and similar with limited  resources, internet speed or connectivity.

here is the raw text (without linebreaks):

```
* Characteristics:
  * *Tardigrade backend*: Uses native RPC protocol, connects directly to the storage nodes which hosts the data. Requires more CPU resource of encoding/decoding and has network amplification (especially during the upload), uses lots of TCP connections 
  * *S3 backend*: Uses S3 compatible HTTP Rest API via the shared gateways. There is no network amplification, but performance depends on the shared gateways and the secret encryption key should be shared with the gateway. 
* Typical usage:
  * *Tardigrade backend*: Server environments and desktops with enough resources, internet speed and connectivity - and applications where tardigrades client-side encryption is required.
  * *S3 backend*: Desktops and similar with limited resources, internet speed or connectivity.
```

Typo: Tadigrade -> Tardigrade (several places)

elek:
Thanks for the suggestions.

I agree with your phrasing, I just started to use it (almost) as is ;-) 

I did only one micro modification: using `end-to-end` encryption instead of `client-side` encryption. (I think it's more expressive...)

Also fixed the typos and started to use multi-level list as formatting:

image

olefrost:
Thanks, looks good to me!

> I did only one micro modification: using end-to-end encryption instead of client-side encryption. (I think it's more expressive...)

That is OK with me, I can see that it is often used by Storj e.g. in this blog post:
https://www.storj.io/blog/what-is-end-to-end-encryption

I do however find the term end-to-end encryption weaker than client-side encryption when it comes to cloud storage.

Example: I have end-to-end encryption to my GoogleDrive (because the REST API uses HTTPS/TLS), but this does not keep my data safe from preying eyes at Google. If I use the crypt backend, Cryptomator or something similar on top of my Google Drive, then I have client-side encryption and is protected from preying eyes at Google.

This is in line with the definitions found on Wikipedia:
https://en.wikipedia.org/wiki/End-to-end_encryption
https://en.wikipedia.org/wiki/Client-side_encryption

elek:
Ok, I am convinced (3888f7a).

elek:
Hi, thanks for all the feedback.

Please let me know if need to improve anything here.

>  All commit messages are in house style.

Do I need to squash commits to one or it's done during the merge?

elek:
I would suggest separating the question of 64 MB chunks from this generic doc update + providers.

As far as I know, there is no open question here, but please let me know if I need to update it in any way...

Thanks a lot...

ncw:
Yesterday I merged a patch which deals with the differences between the providers in a more sane way.

Can you rebase this on master, then add a `Storj` case to `setQuirks` and work out what quirks (if any) Storj needs? You can test with the integration tests (use `go test -v -remote rackcorp:` in the backend/s3 directory).

https://github.com/rclone/rclone/blob/25366268fe921608ee68c6dc33b744b80d860abe/backend/s3/s3.go#L1692-L1698

The way to change the min chunk size would be to adjust it in setQuirks to 64MB as a minimum.

Thank you :-)

elek:
Thanks for the hint @ncw, I created the `setQuirck` case for `Storj` and set the chunk size (64MB). Other quirks are not necessary (fine with `urlEncodeListings = true`).

Rebased the branch and tested it from CLI with creating bucket / uploading data to Storj gatewayt...

elek:
@ncw are you fine with this approach? Would be happy to further improve this PR if you have any suggestions...

elek:
Hi, can anybody help this PR to be merged, please? If you need any more modifications, please let me know...

calebcase (approved):
LGTM Thanks!

calebcase:
@ncw Anything else we should adjust?

elek:
Hi, please let me know if I need to improve this patch in any way...

olefrost:
Hi @elek, I haven’t seen any @ncw online activity since December 9th; I hope he is OK and returns.

Tip: The [rclone commit log](https://github.com/rclone/rclone/commits/master) can be used to monitor approval/merge activity.

ncw:
@elek I've squashed, rebased and reworded a few docs so I think this is ready to merge if you are OK?

elek:
> @elek I've squashed, rebased and reworded a few docs so I think this is ready to merge if you are OK?

Sure, I am fine with that. Thank you very much.

ncw:
I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.58

Thank you for taking a look @elek

PR #5654, opened 2021-10-02
ftp: add hook for about command

ivandeex:

FTP servers do not have a standard way to report used or free space. The rclone FTP client consequently does not implement the About feature. This locks it from being used in Union with space-related policies.

However many clouds provide FTP and SFTP together. While SFTP on average has lower transfer speed than FTP, it's more widespread and usually **SFTP is able to report used/free space**. Here comes the idea to fill another gap in the FTP backend on the way to finally become a first class citizen in the rclone ecosystem.

This patch adds new `--ftp-about-command "shell command"` flag and related backend option allowing users to transparently submit output of `rclone about --json twin_sftp_remote:` as a hint for `rclone about ftp_remote:`
In fact, any shell command can be used if its output is compatible with rclone's **about report (json)**.

@ncw Please review. Do you think it looks too niche?

No

#### Checklist

### DIFF
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -4,15 +4,18 @@ package ftp
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"io"
 	"net"
 	"net/textproto"
+	"os/exec"
 	"path"
 	"runtime"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/google/shlex"
 	"github.com/jlaffaye/ftp"
 	"github.com/pkg/errors"
 	"github.com/rclone/rclone/fs"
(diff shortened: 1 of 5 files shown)


ncw:
I just had a look over this and it occurred to me that there is nothing specifically about FTP in this patch and there are lots of other backends which don't have a working About command (eg the sftp backend if the ssh channel isn't working) but might have a side channel (or the user might want to set some fixed values of their own).

So can we think of a way of making this more general?

The `about_command` could be general purpose thing potentially on all backends and we could make an `fs.About` wrapper which gets called instead of the `Features().About` which implements the extra logic.

There is the wrapping backend option too, but that seems like a sledgehammer to crack a nut!

Or alternatively we could just factor all the bits and include them in backends which might need it. That is probably the least invasive thing to do.

ivandeex:
I have more important stuff to work in short term.
Closing.

PR #5667, opened 2021-10-06
Compress backend docs: minor improvements.

pataquets:


docs/content/compress.md
```diff
@@ -117,11 +117,11 @@ Here are the advanced options specific to compress (Compress a remote).
 GZIP compression level (-2 to 9).
 
 			Generally -1 (default, equivalent to 5) is recommended.
-			Levels 1 to 9 increase compressiong at the cost of speed.. Going past 6 
+			Levels 1 to 9 increase compressiong at the cost of speed. Going past 6
```
albertony, line 120: This needs to be changed in compress.go instead. See marker on line 85:

```
{{< rem autogenerated options start" - DO NOT EDIT - instead edit fs.RegInfo in backend/compress/compress.go then run make backenddocs" >}}
```

Same with rest of the changes below this point.

docs/content/compress.md
```diff
@@ -66,16 +66,16 @@ y/e/d> y
...
 compression and 9 is strongest compression.
 
 #### Filetype
-If you open a remote wrapped by press, you will see that there are many files with an extension corresponding to
+If you open a remote wrapped by compress, you will see that there are many files with an extension corresponding to
 the compression algorithm you chose. These files are standard files that can be opened by various archive programs, 
 but they have some hidden metadata that allows them to be used by rclone.
 While you may download and decompress these files at will, do **not** manually delete or rename files. Files without
-correct metadata files will not be recognized by rclone.
+correct metadata will not be recognized by rclone.
```
albertony, line 78: It is actually metadata *files* isn't it? Have not used the Compressed backend myself, but it seems to store metadata in companion .json files? This is not explained anywhere in docs, though. I agree that *metadata* looks better here, in the current state of the docs, but would you be willing to add a sentence about how the compress backend uses metadata files?

albertony (changes requested):
Thank you, nice improvements! I have some comments in-line.

PR #5698, opened 2021-10-09
docs: cleanup header levels in backend docs

albertony:

Cleanup header levels in backend docs, create better grouping and more hierarchical structure in table of contents.

The following screenshots from Google Drive Table of Contents pretty much sums up the changes:

Before:
Before

Now:
After

Full changelog:
- For all backend docs, creating a header level 2 `Configuration` in the beginning, just after the intro, before the config example.
  - Previously there were normally no level 2 headers at all, just level 1 (the backend name) and then the intro text, the example section and possibly more descriptions. Then the next header would be level 3.
  - Some backend docs, e.g. Crypt, did have level 2 headers far down the docs. All the headers above were level 3, by putting a level 2 header above them as well the TOC gets a more natural hierarchy.
  - Autogenerated level 3 headers `Standard Options` and `Advanced Options` will be children of this.
- Changes `Backend commands` header from level 3 to level 2.
  - This is autogenerated, so change is in `cmd/backend/backend.go`.
  - This is inserted after the level 3 Standard/Advanced options headers, at the end of level 2 `Configuration`, and will now be a sibling of `Configuration` and not a child.
  - The individual commands will now get level 3 headers, instead of level 4, and this means they will be included in the TOC now.
  - Not many backends have it, example of one that does is Google Drive.
- Many backend docs have a `Limitations` header at the end. Changed that from level 3 to level 2, i.e. sibling of `Configuration` and `Backend command`.
- Changes header `Restricted filename characters` from level 4 to level 3.
  - Don't think it was correct as it was with level 4, then it became a child of the previous level 3 header - typically `Modified time and hashes`. Now it will also be included in TOC.
- Capitalization:
  - The following autogenerated used "title case", uppercase letter at beginning of each word:
      - `Standard Options`
      - `Advanced Options`
  - Other autogenerated did not:
      - `Backend commands`
  - Most custom headers does not.
  - Changed the two options headers to lowercase, for consistency:
      - `Standard options`
      - `Advanced options`
  - This change is in code, `cmd/help.go`.
- Some adjustments to headers describing the alternative authentication methods in Jottacloud docs.
- Removed `#` markers at the end of headers, only using it as prefix. This makes it easier to adjust them, and more consistent with autogenerated parts etc.

#### Checklist


albertony:
You're too late to the party! 😆No.. I do not disagree. Current merge was sort of a conservative approach, and partly about consistency. Could definitely do more. For sure, the wizard is a bit boilerplate, and I've noticed it gets outdated as well. Good suggestions, I will need some time to think it through, though..

ncw:
@ivandeex wrote:
> 1. Let's add level-2 header `## Configuration reference` before two auto-generated level-3 sections for standard and advanced options.
> 2. Being an advanced user I find wizard-based intro in many (not always!) backend docs rather cumbersome boilerplate.
>    In the newly added docs (hasher and more or less sia) I first describe the goals and idea/behaviors of backend, maybe usage scenarios or quickstart.
>    Having got the idea behind it the user will frequently find its options/knobs self-evident just from their names.
>    With this structure the wizard becomes less important so we can augment it by (manual config) rclone.conf examples presenting typical option combinations by scenario.
>    The wizard-based (interactive) and manual (rclone.conf) config can become level-3 sections under level-2 "Configuration" section.

There is a lot of boiler plate and redundant examples in most of the backend docs that is for sure!

I had the idea that we should auto make the wizard transcripts.

I agree that these aren't that useful for advanced users - I don't know if new users find them comforting or not?

@albertony wrote:
> You're too late to the party! laughingNo.. I do not disagree. Current merge was sort of a conservative approach, and partly about consistency. Could definitely do more.

Once step at a time and we'll get there :-)

> For sure, the wizard is a bit boilerplate, and I've noticed it gets outdated as well. Good suggestions, I will need some time to think it through, though..

Perhaps we should try to make the one true perfect backend doc then migrate the others into that format.

If we have words to repeat in each backend then we could be auto inserting them also.

PR #5700, opened 2021-10-09
filter: add `{{ regexp }}` syntax to pattern matches - fixes #4074

### LINKED ISSUE #4074: Enhance Filters : for example Match excluding substrings
What is your current rclone version (output from `rclone version`)?
rclone v1.51.0

What problem are you are trying to solve?
I'm trying to match a path/filename not having some substrings in order to discard them on filters.
Now you can match substrings but cannot match strings not having certain substrings.

Rclone is using standard golang regex lib, but it doesn't support features like lookahead assertion (?!) that will allow it.

How do you think rclone should be changed to solve that?
Using, optionaly, another regex lib, like https://github.com/dlclark/regexp2 for example.

Regards

ncw:
There has been a desire from more advanced rclone users to have regexp filtering as well as the glob filtering.

This patch adds regexp filtering using this syntax `{{ regexp }}` which is currently a syntax error, so is backwards compatibile.

This means regexps can be used everywhere globs can be used, and that they also can be mixed with globs in the same pattern, eg `*.{{jpe?g}}`

See: #4074 for background

Fixes #4074 

@ivandeex and @albertony  would appreciate your feedback on this - is this a good idea or not?

### DIFF
--- a/docs/content/filtering.md
+++ b/docs/content/filtering.md
@@ -33,6 +33,9 @@ you expect. Instead use a `--filter...` flag.
 
 ### Pattern syntax
 
+Here is a formal definition of the pattern syntax,
+[examples](#examples) are below.
+
 Rclone matching rules follow a glob style:
 
     *         matches any sequence of non-separator (/) characters
@@ -42,8 +45,10 @@ Rclone matching rules follow a glob style:
               character class (must be non-empty)
     { pattern-list }
               pattern alternatives
+    {{ regexp }}
+              regular expression to match
     c         matches character c (c != *, **, ?, \, [, {, })
-    \c        matches reserved character c (c = *, **, ?, \, [, {, })
+    \c        matches reserved character c (c = *, **, ?, \, [, {, }) or character class
 
 character-range:
 
@@ -61,6 +66,10 @@ character classes (see [Go regular expression reference](https://golang.org/pkg/
     Named character classes (e.g. [\d], [^\d], [\D], [^\D])
     Perl character classes (e.g. \s, \S, \w, \W)
     ASCII character classes (e.g. [[:alnum:]], [[:alpha:]], [[:punct:]], [[:xdigit:]])
+
+regexp for advanced users to insert a regular expression - see [below](#regexp) for more info:
+
+    Any re2 regular expression not containing `}}`
 
 If the filter pattern starts with a `/` then it only matches
 at the top level of the directory tree,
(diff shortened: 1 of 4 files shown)


albertony:
My first impression is that this is a very usefull addition! 🥳 

Can't think of any big caveats with the `{{ }}` alternative, but have to think it more through - and I hope to get time to try it out within a few days.

PS: Will the https://filterdemo.rclone.org/ site handle this syntax out of the box, when this is merged?

ivandeex:
I don't know if `{{ }}` is safe with Powershell but we have a guru here.

ivandeex:
Last note: we'd better leave it for 1.58 to let idea settle.

albertony:
Quick test, matching `notepad++.exe`:

Works in PowerShell:
`--include "*{{\+{2}\.exe}}"`
`--include '*{{\+{2}\.exe}}'`

Works in Command Prompt:
`--include *{{\+{2}\.exe}}`
`--include "*{{\+{2}\.exe}}"`

albertony:
Confirmed. The following quick-and-dirty patch fixes it:

```
-                       if c == '}' && last == '}' {
+                       if !inBraces && c == '}' && last == '}' {
                                inRegexp = false
                                // Unwrite last }
                                re.Truncate(re.Len() - 1)
                                _, _ = re.WriteRune(')')
                        } else {
+                               if c == '{' {
+                                       inBraces = true
+                                       next = 0
+                               } else if inBraces && c == '}' {
+                                       inBraces = false
+                                       next = 0
+                               }
```

albertony:
> Unfortunately @albertony your patch doesn't address `{{\}}}`.

😿 

> I came up with an alternative solution which I've pushed. Not a beautiful solution, but oh-well!

😹 

Passes the few test cases that I tried on previous iteration, so this is the best version ever... so far! :)

ncw:
> just downloaded win 64 bit, it works. thanks
> 
> given the constraint that regex needs to be compatible with glob, your solution is as good as can be expected.
> 
> regex is confusing, glob+regex is confusing squared.

Aye!

> for me, i am lost without websites like https://pythex.org/ and https://regoio.herokuapp.com/ so the only tester will be https://filterdemo.rclone.org/ would it be possible to:
> 
> * enable `--dump=filters` output

Good plan

> * add the documentation

Another great idea. We could maybe load the filter page in an iframe.

> > Any ideas for improving the UI on that site? I got stuck on the UI so never quite completed it!
> 
> Could be fun to play with it some time, but not really a web/gui developer so could easily end up in a rabbit hole of tweaking css, finding color codes or whatever... Is the source somewhere public?

I just chucked it on https://github.com/rclone/filterdemo

I'm not so much worried about css/colours (I can tweak those) it is more how should the website work. So I guess the UX rather than the UI. Any thoughts much appreciated!

ncw:
So now we've got the release out of the way - shall we

- m) merge
- f) fix-up
- c) cancel

this PR?

asdffdsazqqq:
not sure what `fix-up` means?

ncw:
> not sure what `fix-up` means?

Change something in the proposal.

asdffdsazqqq:
@ivandeex wrote
"The only thing lacking is post-modifers (like {{NeEdLe}}i for ignored case) used commonly with regexp."

ivandeex:
> lacking is post-modifers (like {{NeEdLe}}i for ignored case

A unit test with `{{?i:poTAto}}` or `{{(?i)poTAto}}` and an explaining line/example in the docs would be great.

ncw:
> > lacking is post-modifers (like {{NeEdLe}}i for ignored case
> 
> A unit test with `{{?i:poTAto}}` or `{{(?i)poTAto}}` and an explaining line/example in the docs would be great.

I did a unit test with `(?i)` and an example in the docs.

If no-one has any violent objections I'll merge this when it builds green :-)

PR #5703, opened 2021-10-10
librclone: add RcloneFreeString function

albertony:

Export new function `RcloneFreeString` from librclone, to `free` strings returned by `RcloneRPC`.

This basically revives PR #5362, which was closed without being merged (just because it was "abandoned"?).

In the original PR the reasoning was that it would help in a C# client that did not have a convenient way of freeing the string. In my case it is a C++ client, on Windows, using Visual C++ compiler: Building librclone with go, cgo and mingw gcc produces a dll depending on the older C runtime library `MSVCRT.dll`. Building a client application with latest Visual Studio produces an exe depending on `VCRuntime140.dll` and `UCRTBase.dll`. I'm able to get this to work, except attempting to call free on the strings from `RcloneRPC` triggers HeapFailure. Calling the new RcloneFreeString avoids this, because it ensures that `malloc` and `free` are performed from within the same library - using the same C runtime library.

Ideally I should build a `librclone.dll` using the same VCRuntime as my client application, but I don't know yet how to get this through the go-cgo-mingw-gcc "pipeline". But in any case, I think it is good practice (if not a requirement - on Windows) for a plain C API to pair memory allocation and deallocation within the same library boundary.

(I haven't tried librclone before, but now in the early stages of setting up a test bench with C++ and CMake, primarly Windows but possibly cross platform. May result in docs/example PR later on).

#5362

#### Checklist

### DIFF
--- a/librclone/README.md
+++ b/librclone/README.md
@@ -31,6 +31,7 @@ For documentation see the Go documentation for:
 - [RcloneInitialize](https://pkg.go.dev/github.com/rclone/rclone/librclone#RcloneInitialize)
 - [RcloneFinalize](https://pkg.go.dev/github.com/rclone/rclone/librclone#RcloneFinalize)
 - [RcloneRPC](https://pkg.go.dev/github.com/rclone/rclone/librclone#RcloneRPC)
+- [RcloneFreeString](https://pkg.go.dev/github.com/rclone/rclone/librclone#RcloneFreeString)
 
 ### C Example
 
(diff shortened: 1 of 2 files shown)


ncw:
> > Thinking about this, I think that the python lib isn't freeing any of the memory being returned which it really should be!
> 
> That is true. I guess it needs to use this new `RcloneFreeString` as well?

I think it does indeed. Though that might be easier said than done...

PR #5705, opened 2021-10-11
chunker: md5all must create metadata if base hash is slow

### LINKED ISSUE #5508: Chunker: local to local chunker transfers ignores "hash_type = md5all" config setting and acts as "md5"
None.

When I am trying to copy from 'local' to 'local->chunker' and 'hash_type = md5all' config setting or '--chunker-hash-type md5all' flag is set, the files which are smaller than 'chunk_size' (e.g.: 1P) don't have their metadata file created and "forcefully chunked" original file renamed with additional 'name_format' file extension for chunked files. Basically I get identical copy (with 'chunk_size = 1P' or similar). I want to use that 'md5all' thing, so my files I want to upload have md5 checksums pre-calculated before upload. That way the upload would be much faster (uploading to Jottacloud requires md5 checksum in advance).

```rclone v1.56.0```

Windows 10, 64 bit

Local

```rclone copy local:D:\Data local-chunk:D:\Data-chunked --chunker-hash-type md5all -P --fast-list --transfers=1```
That is my pseudo command.

From rclone config file (pseudo example):
```
[local]
type = local
nounc = false

[local-chunk]
type = chunker
remote = local:
chunk_size = 1P
name_format = *.#
```

I think there is no need for the log as the problem is very obvious, clear and reproduceable (I think so).

I hope there is enough information from my pseudo examples. The best thing would be if rclone could use md5 checksum files for that. NTFS file system don't store file checksums. It seems that similar feature is currently worked on and should be released with the next rclone version: https://github.com/rclone/rclone/issues/949. I am really happy about that.

#### How to use GitHub

ivandeex:

Before this patch the md5all option would skip creating metadata with
hashsum if base filesystem provided md5, in hope to pass it through.
However, if base hash is slow (for example on local fs), chunker passed
slow md5 but never reported this fact in features.

This patch makes chunker snapshot base hashsum in metadata when md5all is
set and base hashsum is slow since chunker was intended to provide only
instant hashsums from the start.

Fixes #5508 

#### Checklist

### DIFF
--- a/backend/chunker/chunker.go
+++ b/backend/chunker/chunker.go
@@ -432,10 +432,10 @@ func (f *Fs) setHashType(hashType string) error {
 		f.hashFallback = true
 	case "md5all":
 		f.useMD5 = true
-		f.hashAll = !f.base.Hashes().Contains(hash.MD5)
+		f.hashAll = !f.base.Hashes().Contains(hash.MD5) || f.base.Features().SlowHash
 	case "sha1all":
 		f.useSHA1 = true
-		f.hashAll = !f.base.Hashes().Contains(hash.SHA1)
+		f.hashAll = !f.base.Hashes().Contains(hash.SHA1) || f.base.Features().SlowHash
 	default:
 		return fmt.Errorf("unsupported hash type '%s'", hashType)
 	}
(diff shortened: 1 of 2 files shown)


ivandeex:
@ncw this is ready for your review

PR #5710, opened 2021-10-11
docs: note minimum supported docker engine

ivandeex:

Note minimum supported Docker Engine

https://forum.rclone.org/t/can-not-install-docker-volume-plugin-layers-from-manifest-dont-match-image-configuration/26784/9

#### Checklist

### DIFF
--- a/docs/content/docker.md
+++ b/docs/content/docker.md
@@ -306,6 +306,8 @@ We maintain the
 [docker-volume-rclone](https://hub.docker.com/p/rclone/docker-volume-rclone/)
 plugin image on [Docker Hub](https://hub.docker.com).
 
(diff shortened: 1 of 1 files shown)


ivandeex:
> There are a lot of documentation before this section. Could it be that many users will "jump the gun" and start hacking, before reaching this part with the important requirement? Open question, I zero experience with docker volume plugins..

Based on forum and tracker, we have only Karl Forner who was affected.
He used to use a standard `Ubuntu 16.04` image on AWS.
Even now Docker 19.03 is rather old.
As time goes and docker installations get upgraded I expect the probability to approach zero.

ivandeex:
I'd rather keep this requirement away from "Quickstart/Getting started".
Users who don't have a problem will not care.
Affected users are expected to get desperate and read more.

PR #5714, opened 2021-10-12
fix suffix keep extension flag for multiple extensions

salvawye:
#### Checklist

- [ x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [x ] I have added tests for all changes in this PR if appropriate.

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -648,7 +648,7 @@ func SuffixName(ctx context.Context, remote string) string {
 		ext := path.Ext(remote)
 		base := remote[:len(remote)-len(ext)]
 		fullExt := ext
-		while ext != "" {
+		for ext != "" {
 			ext = path.Ext(base)
 			base = base[:len(base)-len(ext)]
 			fullExt = ext + fullExt


salvawye:
Yes unfortunately you are right and this change is not very backwards compatible. 
However i think that when adding suffixes to files with multiple extensions it is usually logical to do this before all the extensions so instead of having file.tar.20200101.gz you would have file.20200101.tar.gz as file.20200101 is the real name for the file with the rest just conveying metadata about its format. once unziped and untarred you would expect this to remain as the final name.

This is a challenge to achieve however due to the possibility of people adding . into the file name in certain contets as you both mentioned. Maybe it would be more appropriate to have options on this flag or some way of controlling this.

ivandeex:
this patch is attempting to be "logical" but then it hits real life and just opens a can of worms

it opens new questions ("is this dot an extension or part of name?") and adds extra complexity, tricky flags, in exchange for... what?

salvawye:
yes unfortunately i do agree with most of that. There is no good standard to rely on here and a change would be backward incompatible.

i think that the way it handles these multiple extensions is poor but maybe its not worth the added complexity to address this edge case.

PR #5731, opened 2021-10-18
Change S3 default ACL back to no ACL

### LINKED ISSUE #5730: Rclone defaults to private ACL for S3 backend.  Should default to no ACL
ACL is optional. Some S3 backends do not support ACLs.   e.g. GCS with uniform access control does not support ACLs.  
With the current implementation there is no way to pass no ACL since it overrides to passing "private".
Rclone should not enforce passing an ACL since they are optional.   This is new behavior 
 since this commit and I believe this is unintentional since it was not mentioned in this commit. https://github.com/rclone/rclone/commit/39f5059d48a5e58efb2ca48f68ca02e7fc5e3d46

In AWS S3 if you don't specify an ACL it defaults to private ACL, so there is no need to send private as an default ACL to AWS S3.

https://forum.rclone.org/t/multipart-uploads-to-gcs-using-s3-protocol-fail-with-400-error/26198

philip-harvey:
Change S3 default ACL back to no ACL as it was prior to https://github.com/rclone/rclone/commit/39f5059d48a5e58efb2ca48f68ca02e7fc5e3d46

ACL is optional for S3, currently it is enforced in rclone.  This breaks some S3 backends that do not accept an ACL. e.g. GCS XML API with uniform bucket access.

https://forum.rclone.org/t/multipart-uploads-to-gcs-using-s3-protocol-fail-with-400-error/26198/15
fixes #5730 

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1421,7 +1421,11 @@ This ACL is used for creating objects and if bucket_acl isn't set, for creating 
 For more info visit https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
 
 Note that this ACL is applied when server-side copying objects as S3
-doesn't copy the ACL from the source but rather writes a fresh one.`,
+doesn't copy the ACL from the source but rather writes a fresh one.
+
+If the acl is an empty string then no X-Amz-Acl: header is added and
+the default (private) will be used.
+`,
 			Provider: "!Storj,Cloudflare",
 			Examples: []fs.OptionExample{{
 				Value:    "default",
@@ -1475,7 +1479,11 @@ doesn't copy the ACL from the source but rather writes a fresh one.`,
 For more info visit https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
 
 Note that this ACL is applied when only when creating buckets.  If it
-isn't set then "acl" is used instead.`,
+isn't set then "acl" is used instead.
+
+If the "acl" and "bucket_acl" are empty strings then no X-Amz-Acl:
+header is added and the default (private) will be used.
+`,
 			Advanced: true,
 			Examples: []fs.OptionExample{{
 				Value: "private",
@@ -2619,6 +2627,14 @@ func (f *Fs) setRoot(root string) {
 	f.rootBucket, f.rootDirectory = bucket.Split(f.root)
 }
 
+// return a pointer to the string if non empty or nil if it is empty
+func stringPointerOrNil(s string) *string {
+	if s == "" {
+		return nil
+	}
+	return &s
+}
+
 // NewFs constructs an Fs from the path, bucket:path
 func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, error) {
 	// Parse config into Options struct
@@ -3502,7 +3518,7 @@ func (f *Fs) makeBucket(ctx context.Context, bucket string) error {
 	return f.cache.Create(bucket, func() error {
 		req := s3.CreateBucketInput{
 			Bucket: &bucket,
-			ACL:    &f.opt.BucketACL,
+			ACL:    stringPointerOrNil(f.opt.BucketACL),
 		}
 		if f.opt.LocationConstraint != "" {
 			req.CreateBucketConfiguration = &s3.CreateBucketConfiguration{
@@ -3567,7 +3583,7 @@ func pathEscape(s string) string {
 // method
 func (f *Fs) copy(ctx context.Context, req *s3.CopyObjectInput, dstBucket, dstPath, srcBucket, srcPath string, src *Object) error {
 	req.Bucket = &dstBucket
-	req.ACL = &f.opt.ACL
+	req.ACL = stringPointerOrNil(f.opt.ACL)
 	req.Key = &dstPath
 	source := pathEscape(path.Join(srcBucket, srcPath))
 	if src.versionID != nil {
@@ -5037,7 +5053,7 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 
 	req := s3.PutObjectInput{
 		Bucket: &bucket,
-		ACL:    &o.fs.opt.ACL,
+		ACL:    stringPointerOrNil(o.fs.opt.ACL),
 		Key:    &bucketPath,
 	}
 
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2638,9 +2638,6 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if opt.Versions && opt.VersionAt.IsSet() {
 		return nil, errors.New("s3: cant use --s3-versions and --s3-version-at at the same time")
 	}
-	if opt.ACL == "" {
-		opt.ACL = "private"
-	}
 	if opt.BucketACL == "" {
 		opt.BucketACL = opt.ACL
 	}


philip-harvey:
@ncw any thoughts on this?  It's not correct to always send an ACL to a S3 backend, hence either this default needs to be changed or a switch added to stop this behavior.  e.g. --s3-no-ACL  or something like that
Thanks!

ncw:
I'm sure I made a patch for this already so you could set the value to `unset`...

Found it eventually! (Should have looked back through the thread - I see I posted it already!)

[v1.58.0-beta.5880.2e99232a0.fix-b2-acl](https://beta.rclone.org/branch/fix-b2-acl/v1.58.0-beta.5880.2e99232a0.fix-b2-acl/) on branch [fix-b2-acl](https://github.com/rclone/rclone/tree/fix-b2-acl) (uploaded in 15-30 mins)

> Note that is says you CAN use headers to grant ACL based permissions. Currently rclone is enforcing passing ACLs, which is going to become a problem down the road since ACLs are deprecated and more systems will stop handling them.

Can you link some docs on ACLs being deprecated? And what do you mean by "more systems will stop handling them"?

I like the idea of not sending ACLs if it helps in the future, but I really don't want to break stuff!

philip-harvey:
Hi @ncw 
I guess ACL are not deprecated, apologies, but AWS and GCP both recommend that ACLs not be used.  
from this page:  
https://aws.amazon.com/blogs/security/iam-policies-and-bucket-policies-and-acls-oh-my-controlling-access-to-s3-resources/

> "As a general rule, AWS recommends using S3 bucket policies or IAM policies for access control. S3 ACLs is a legacy access control mechanism that predates IAM"

GCS is a good example of a system that allows, and recommends, that you turn off ACLs, and if you turn off ACLs then rclone can't talk to it using S3.  
Given that both AWS and GCP recommend ACLs are not used I expect some  S3 systems in future will either not implement ACLs or will allow disabling of ACLs.  GCS already allows ACLs to be disabled.  It also seems that backblaze is an example of a system that does not support ACLs at all. I assume the patch backblaze did on their side was to simply discard all ACLs that are sent to them rather than error.

from:
https://cloud.google.com/storage/docs/uniform-bucket-level-access
> Should you use uniform bucket-level access?
> Generally, using uniform bucket-level access is recommended, because it unifies and simplifies how you grant access to your Cloud Storage resources."

and

> Behavior when enabled
> You can enable uniform bucket-level access either when you create a new bucket, or when you explicitly enable uniform bucket-level access on an existing bucket.
> 
> Once enabled, a bucket has the following behavior:
> 
> Requests to set, read, or modify bucket and object ACLs fail with 400 Bad Request errors.
> 
> This includes JSON API requests using any BucketAccessControls, DefaultObjectAccessControls, or ObjectAccessControls methods.
> 

I'm not 100% sure, but I think the the code in the new branch introduces a new issue.  e.g. what happens when you set ACL = unset and don't set a bucketACL, I think the bucketACL will be sent as unset, which most likely isn't a valid ACL.

backend/s3/s3.go
```diff
@@ -1684,11 +1684,8 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if err != nil {
 		return nil, errors.Wrap(err, "s3: upload cutoff")
 	}
-	if opt.ACL == "" {
-		opt.ACL = "private"
-	}
 	if opt.BucketACL == "" {
-		opt.BucketACL = opt.ACL
+		opt.BucketACL = "private"
```
ncw, line 1688: This isn't quite the same logic as before...

Before if ACL == "something" and "BucketACL" == "" then BucketACL becomes "something"

After if ACL == "something" and "BucketACL" == "" then BucketACL becomes "private"

Now whether that is the correct logic or not I don't know!

ncw (commented):
OK I'm convinced by your arguments :-)

Should we do the same thing to BucketACL?

This should probably have some more docs here

https://github.com/rclone/rclone/blob/f9e54f96c3880246e0ce0988b1534ab3423354a4/backend/s3/s3.go#L1015-L1023

See inline for comment.

backend/s3/s3.go
philip-harvey: I'm not sure what would be a good default ACL to apply to a new bucket if a bucket ACL isn't supplied, however I guess this has the same issue, i.e. you can't create an ACL on a bucket in a system that doesn't support ACLs or has ACLs disabled, so we should allow not setting an ACL.

rpattcorner:
@ncw @philip-harvey The branch looks great, and is solving a severe operational problem for us, but it isn't in the current beta that I can tell.  Any hints on the likelihood that the patch will be in something official soon?

philip-harvey:
I've lost track of what needs to be done to get this merged.  I'm still using my own build of rclone with GCS since the public builds are not GCS compatible with S3 protocol on GCS buckets with uniform access control enabled.  I'm using S3 protocol since it's > 2X faster for large file uploads on a 10GB pipe vs GCS protocol.

rpattcorner:
@philip-harvey well, darn!  We're only experiencing the problem in cross-account copies, which may mask the issue for many users, but it's a killer in cross-account.  I'll try filing a bug report and see if that unsticks anything.  Reference https://github.com/rclone/rclone/issues/6413

ncw:
Sorry for the long delay on this...

I think that changing the ACL for objects is decided.

- Not supplying an ACL with PUT defaults to private, and ACLs are generally deprecated

I don't think we've sorted out changing the ACL for buckets? I've been unable to find out if the default access for buckets is `private` if the ACL isn't supplied - can you see this in the docs somewhere?

 Assuming it is, then I think the patch should look something like

```patch
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2638,9 +2638,6 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if opt.Versions && opt.VersionAt.IsSet() {
 		return nil, errors.New("s3: cant use --s3-versions and --s3-version-at at the same time")
 	}
-	if opt.ACL == "" {
-		opt.ACL = "private"
-	}
 	if opt.BucketACL == "" {
 		opt.BucketACL = opt.ACL
 	}
```

If you agree @philip-harvey then please update your patch to the above and I'll merge it.

rpattcorner:
@ncw Hi Nick.  I'd just like to be sure that @philip-harvey 's addition of the `unset` option remains in the PR, as that's the critical functionality that allows rclone operation with the new ACL scheme.  I think that's what you meant by `changing the ACL for objects is decided.` but am not sure.

In terms of doc, at least on the object side, I'm looking here and seeing
```
ACLs disabled

Bucket owner enforced (recommended) – ACLs are disabled, and the bucket owner automatically owns and has full control over every object in the bucket. ACLs no longer affect permissions to data in the S3 bucket. The bucket uses policies to define access control.
```
Which I think means that there is no default ACL at all (not even private), which is why the inability to specify `no acl` and asserting `private` instead is failing for us

philip-harvey:
I actually think that if no ACL is passed that no ACL should be set.  
The default ACL on a new bucket in AWS is private anyway, so no need to pass this. 

ACL: private |  Applies to: Bucket and object | ACL: Owner gets FULL_CONTROL. No one else has access rights (default).

AWS has started making it possible to enforce no ACL being set on buckets in an org, and passing an ACL in this case will cause it to fail to create the bucket.
"Requests to set ACLs or update ACLs fail and return the AccessControlListNotSupported error code. Requests to read ACLs are still supported." 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ensure-object-ownership.html

philip-harvey:
I'm not sure if any other code needs updated to not pass the empty string for the Bucket ACL, or if passing the empty string is fine.

rpattcorner:
I’m reasonably certain that passing an empty ACL string to a no-ACL bucket will fail, but not passing the header at all will succeed

rpattcorner:
Let me distinguish a couple things.  

1. Based on _current code_, specifying an empty ACL using the ` --s3-acl "" ` switch **fails** because the current code passes an `x-amz-acl` header of `private` when specifying an empty value to the switch

2. A test using the _fork_ allows sending a request with **no** `x-amz-acl` header by means of the switch `--s3-acl unset`, which **succeeds**

I know of no way to test passing an `x-amz-acl` header with an empty value (different than case 1 where the switch had an empty value but the header did not, and different than case 2 where there was no `x-amz-acl` header in the call).  It looks like neither the current nor the forked code allows that case, so it should probably be tested.

backend/s3/s3.go
```diff
@@ -2638,12 +2638,6 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if opt.Versions && opt.VersionAt.IsSet() {
 		return nil, errors.New("s3: cant use --s3-versions and --s3-version-at at the same time")
 	}
-	if opt.ACL == "" {
-		opt.ACL = "private"
-	}
-	if opt.BucketACL == "" {
-		opt.BucketACL = opt.ACL
```
ncw, line 2645: I think we still need these 3 lines otherwise we'll cause a backwards compatibility problem where users have set `acl` and are expecting `bucket_acl` to default to that.

```go
if opt.BucketACL == "" {
		opt.BucketACL = opt.ACL
}
```

ncw (commented):
Please see inline query

Thanks

backend/s3/s3.go
ncw: Rather than using the special case `unset` I've decided that empty string should mean not to supply the x-amz-acl header.

I'm going to write a bit more about this below (not in a review comment)!

ncw:
Rather than using the special case `unset` I've decided that empty string should mean not to supply the x-amz-acl header.

To this end I've modifed @philip-harvey 's commit to add back the defaulting of the bucket acl to the normal acl (as per the docs).

I've also made an unset (empty string) acl or bucket acl mean - do not add the  x-amz-acl header at all.

I think these changes are
- backwards compatible
- secure by default
- fix the issue

I'd appreciate testing!

I've pushed these changes here for review and also here which will make binaries for testing @rpattcorner 

[v1.60.0-beta.6443.bf46a86e3.pr-5731-s3-acl](https://beta.rclone.org/branch/pr-5731-s3-acl/v1.60.0-beta.6443.bf46a86e3.pr-5731-s3-acl/) on branch [pr-5731-s3-acl](https://github.com/rclone/rclone/tree/pr-5731-s3-acl) (uploaded in 15-30 mins)

teachers-uk-net:
Hi

I was experiencing the issue reported for pushing to an S3 bucket.

I cannot say that I have fully tested the branch for backwards compatibility or security but it does not look like it has broken anything else and looks secure.  It also definitely allowed me to push to an S3 bucket that has No ACL enforced.

Thanks!

ncw:
Thanks for testing.

I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.61

ncw:
I merged this in 028832ce73099e387e81533f21007b95e9e99b33 and c7c9356af5f9b3be84c9c9f220aa0181c176ccd7

Issue #5734, opened 2021-10-18
v1.57 release


ncw:
Dear @rclone/rclone 

The v1.57 release is overdue! I've pushed it back until the end of the month (12 days from now) to give us a little breathing space.

In those days I'd like to

- merge any outstanding PRs we should have in this release
- fix [the integration tests](https://pub.rclone.org/integration-tests/current/)

## PRs

I'd like a list of PRs that you'd like merged for the release. In particular from @ivandeex and @albertony I have lots of outstanding PRs which need reveiwing so which of those do you want for v1.57. And any other candidates!

## Integration tests

As far as the integration tests go there are 35 failing tests which is more than I'm happy with (I've beaten this down from 126 already!), so I'd like some help with those if possible.

- @buengese have you got time to look at the `zoho` and `compress` ones?
- @ivandeex can you have a look at the `chunker` ones?
- There are quite a few `TestSyncConcurrentTruncate` failures - since you introduced this test @ivandeex would you mind looking at those?
- I'll look at the `azureblob` ones and the `filefabric` ones
- There is also `TestCopyFileMaxTransfer` which should fix `TestDrive` and `TestCryptDrive` - any takers for that? @albertony ?

That should bring the list down to a more manageable level. I'll just note that the integration tests for `mega` are often broken as are the ones for `qingcloud`. The Onedrive china tests are broken because the access token needs refreshing - I'll try to sort that out!

ivandeex:
#5735 will fix integration tests for chunker

buengese:
I've been rather absent from rclone related things for a while, so there are quite a few things I want to catch up on. I'll look into the failing zoho and compress tests first. If I have the time I'm also going to look into the failing encoding tests for 1fichier.

ncw:
@buengese I had a quick go at fixing 1fichier in 9742648fce294315da8afe347e923dd2f64bdb81 - my local tests indicated more list retries will fix the problem - we'll see tomorrow!

ivandeex:
https://github.com/rclone/rclone/pull/5462 has got open questions...

ncw:
After a lot of bashing on bugs, the integration test scores are as follows...

| backend | remote | test | failure | comments |
|----|----|----|----|----|
| compress | TestCompressDrive: | backend/compress | FsChangeNotify | @buengese  - not that important |
| fichier | TestFichier: | backend/fichier | Lots | API rate limiter is killing the tests |
[...]

PR #5741, opened 2021-10-21
touch: fix issue with touch on dir and integration test failure when no modtime support

albertony:

Fix TestTouchDir integration test

https://github.com/rclone/rclone/issues/5734#issuecomment-948665933

#### Checklist

### DIFF
--- a/fs/operations/operations_test.go
+++ b/fs/operations/operations_test.go
@@ -34,6 +34,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
 	_ "github.com/rclone/rclone/backend/all" // import all backends
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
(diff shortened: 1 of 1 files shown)


albertony:
@ncw

>I noticed that TestTouchDir that was merged as part of 41876dd broke 6 integration tests this morning :-(
>
>I think you need to not run this test on backends which don't support modtime (check the features flag and t.Skip - there are lots of examples of this in the tests) - these tests will always pass because the acceptable mod time range is infinite so there is no point running them.
>
>And also t.Skip if the operations.TouchDir returns either of these errors - these are backends which only support modtime when you upload a file and can't change it afterwards which is fine for sync purposes but terrible for touch!

Don't think there is a feature flag for modtime, is there?

But I think checking for the specific errors is enough. Though - the errors are not currently returned by TouchDir, they are just "fs.CountError'ed".  I'm still looking into what to do with this..

albertony:
Now:
- Fixed touch of directory due to mentioned bug
- Fixed the integration test (skipping when no modtime support, not failing if ErrorCantSetModTime or ErrorCantSetModTimeWithoutDelete)
- Added some more unit tests of touch command on directory
- Tiny improvement in error message with recursive touch failures (wrap original error)

PR #5752, opened 2021-10-24
librclone: allow empty string or null input instead of empty json object

albertony:

Currently rpc methods must always specify a valid json string as "input", even for methods which does not take any "arguments" - such as listremotes. This means one must send the string "{}". Trying to send empty string, or null, results in status code 400 and error message: "failed to read input JSON: EOF".

With this change I suggest treating empty string and null same as "{}".

#### Checklist

### DIFF
--- a/librclone/librclone/librclone.go
+++ b/librclone/librclone/librclone.go
@@ -88,9 +88,11 @@ func RPC(method string, input string) (output string, status int) {
 	}()
 
 	// create a buffer to capture the output
-	err := json.NewDecoder(strings.NewReader(input)).Decode(&in)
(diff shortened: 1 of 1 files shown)


ncw (approved):
I think this looks fine. I'll merge it now - apologies for the delay!

We could potentially allow the C interface to allow NULL strings also (that would be very C-style).

albertony:
> We could potentially allow the C interface to allow NULL strings also (that would be very C-style).

It does already, tested by passing `nullptr` from C++ (assuming that was what you meant). I guess `C.GoString` handles that for us.

PR #5753, opened 2021-10-24
docs/librclone: add -ldflags -s to suggested build command

albertony:

Suggest using build option `-ldflags -s` to create optimized build (omit symbol table and debug information).

With my test, building c-shared on Windows there is a huge different - half the size:
- No optimization: 79 012 KiB
- Optimized (-ldflags -s): 36 917 KiB (53% reduction)

Edit:
- Tested same on Linux, Arch in WSL: Reduction from 74084 KiB to 56319 KiB (24% reduction)
- Adding -w as well, `-ldflags -s -w`, changes nothing.

https://github.com/rclone/rclone/pull/5729#issuecomment-950344364

#### Checklist

### DIFF
--- a/librclone/README.md
+++ b/librclone/README.md
@@ -33,6 +33,10 @@ The library will depend on `libdl` and `libpthread` on Linux/macOS, unless
 linking with a C standard library where their functionality is integrated,
 which is the case for glibc version 2.34 and newer.
 
+You may add arguments `-ldflags -s` to make the library file smaller. This will
+omit symbol table and debug information, reducing size by about 25% on Linux and
+50% on Windows.
+
 ### Documentation
 
 For documentation see the Go documentation for:


ncw (commented):
I'm not sure the consequences of removing debug symbols for debugabillity with C programs...

Perhaps we should just have another paragraph to say add `-libwhatever` to remove the debug symbols to halve the binary size.?

albertony:
> I'm not sure the consequences of removing debug symbols for debugabillity with C programs...

True, for debugging it makes sense to keep it, at least on Linux.

> Perhaps we should just have another paragraph to say add `-libwhatever` to remove the debug symbols to halve the binary size.?

I'm not sure I understood exactly what you meant here, but force-pushed a suggestion. Just let me know if you were thinking something different.

Issue #5749, opened 2021-10-24
Core/stats shows wrong `totalBytes` during sync


mion00:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/core-stats-during-sync-shows-wrong-totalbytes/27113

#### What is the problem you are having with rclone?
I report here the information from the forum post:
Statistics from the HTTP API _core/stats_ show wrong value in the field **totalBytes** while a _sync_ is in progress.

Note: I noticed that after each single file transfer is done, the field **totalBytes** increases, and its value is the sum of the already transferred files + the file currently being transferred.

I am not familiar at all with the codebase, but after a quick search I stumbled upon [this line](https://github.com/rclone/rclone/blob/4dfcf6899dbf5234cf65ff01b6c7df5362d2f841/fs/accounting/stats.go#L294).
Maybe there is something going on with `transferQueueSize`, such that it is growing only with the file that is being transferred, instead of counting also the one that must still be picked up to be sent?

#### What is your rclone version (output from `rclone version`)
```
rclone v1.56.2
- os/version: linuxmint 20.1 (64 bit)
- os/kernel: 5.11.0-25-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.16.8
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Linux Mint, 64bit

#### Which cloud storage system are you using? (e.g. Google Drive)
Backblaze B2, local

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
In the first terminal start a sync:
```
rclone --verbose --bwlimit=5k --transfers=1 --stats 10s --rc sync /var/log/ /tmp/test-logs/
```

In the second terminal ask for stats using `rclone rc`:
```
rclone -vv rc core/stats
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
Output from `rclone sync`:
```
2021/10/22 21:59:42 INFO  : Starting bandwidth limiter at 5KiByte/s
2021/10/22 21:59:42 NOTICE: Serving remote control on http://localhost:5572/
2021/10/22 21:59:42 ERROR : boot.log: Failed to copy: failed to open source object: open /var/log/boot.log: permission denied
2021/10/22 21:59:42 ERROR : private: failed to open directory "private": open /var/log/private: permission denied
[...]

PR #5754, opened 2021-10-24
stats: fix missing computation of transferQueueSize

mion00:
This fixes the value of `transferQueueSize`, that was not summed up when calculating the sum of the statistics group, resulting in the bug described in #5749 .
Since I don't have familiarity with the code base, I don't know if this change may introduce any side effect.

#5749 

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -24,6 +24,8 @@ const (
 var MaxCompletedTransfers = 100
 
 // StatsInfo accounts all transfers
+// N.B.: if this struct is modified, please remember to also update sum() function in stats_groups
+// to correctly count the updated fields
 type StatsInfo struct {
 	mu                sync.RWMutex
 	ctx               context.Context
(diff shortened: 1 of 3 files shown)


ncw (commented):
I think this looks like the correct fix to me.

Its easy to miss changing that function when we add new stats values which is super annoying.

Probably we could do with a comment here

https://github.com/rclone/rclone/blob/cf2c2792e62beda08821578f041bab3bfe81d7a8/fs/accounting/stats.go#L26-L27

To say that the `sum` function needs adjusting.

I can also see that other values are missing from sum() if you compare with the `StatsInfo` definition...

Ideally the operations in the sum() function should be in the same order as in the `StatsInfo` definition so they are easy to check.

So do you want to extend this to add the missing values and the comment? Or I can merge as-is?

ncw (approved):
Thank you - that looks great now - I'll merge those fixes :-)

PR #5763, opened 2021-10-27
sftp: improved support for Windows OpenSSH

### LINKED ISSUE #5758: Windows sha1sum via SFTP

I am using the official OpenSSH that is now part of windows (link below). This works well for file transfer but hash checks do not seem to work. 
https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse

I have tested using both sha1sum and md5sums. 

https://www.adaic.org/resources/add_content/standards/sha1sum.exe
--sftp-sha1sum-command "sha1sum"

http://www.pc-tools.net/files/win32/freeware/md5sums-1.2.zip
--sftp-md5sum-command "md5sums -u"

The first issue seems to be the path being different to sftp and i got around this using --sftp-path-override "C:\test folder". 

Looking at the console output the issue seems to be the way rclone handles the escape characters and backslashes. 

I think the best way forward would most likely just be an OS detection on SSH connect or a optional flag for Windows.

v1.57.0-beta.5801.bb11803f1

Client - Ubuntu 21.10
Server - Windows Server 2019

SFTP

rclone sha1sum "windows:/C:/test folder/example1/" --sftp-sha1sum-command "sha1sum" -vv

```
2021/10/26 18:08:39 DEBUG : rclone: Version "v1.57.0-beta.5801.bb11803f1" starting with parameters ["rclone" "sha1sum" "windows:/C:/test folder/example1/" "--sftp-sha1sum-command" "sha1sum" "-vv"]
2021/10/26 18:08:39 DEBUG : Creating backend with remote "windows:/C:/test folder/example1/"
2021/10/26 18:08:39 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2021/10/26 18:08:39 DEBUG : windows: detected overridden config - adding "{h5EBD}" suffix to name
2021/10/26 18:08:40 DEBUG : sftp://*****@*****//C:/test folder/example1/: New connection *****:39518->*****:22 to "SSH-2.0-OpenSSH_for_Windows_8.1"
2021/10/26 18:08:40 DEBUG : fs cache: renaming cache item "windows:/C:/test folder/example1/" to be canonical "windows{h5EBD}:/C:/test folder/example1/"
2021/10/26 18:08:40 DEBUG : sftp cmd = /C:/test\ folder/example1/file1.txt
2021/10/26 18:08:40 DEBUG : file1.txt: Failed to calculate sha1 hash: Process exited with status 1 (can't open `/C:/test\': Invalid argument)
                                          file1.txt
2021/10/26 18:08:40 DEBUG : 11 go routines active

```

albertony:

Windows 10, Server 2019, and later can run a SSH server that is a port of OpenSSH. See official [installation guide](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse).

Existing SFTP backend in rclone assumes there is a Unix shell, if any, on the server when executing remote commands. Windows SSH server *can* run with bash shell (Cygwin? WSL?), but defaults to CMD, and even more natural is probably to configure it to use PowerShell. Since both CMD and PowerShell uses different quoting/escaping rules from bash, and also different from each other, the existing remote command execution does not (necessarily) work. If staying away from checksums and the about command, which are the two places remote commands are used currently, the existing version should work. But this PR intends to fix also the missing pieces, such that Windows OpenSSH server can be said to be fully supported.

What it does:
- Introduces new config option `shell_type` with possible values `none|cmd|powershell|unix`
- Auto-detects shell when option `shell_type` is not explicitely set (config file, command-line, env-var)
   - On first access it will execute a simple shell command and try to figure out if there is a "powershell" or a "cmd" behind the remote, and if neither of them are detected then assumes "unix".
       - If a remote ssh session could not be created it sets value "none", which disables shell command execution.
       - If remote ssh session creation succeeded, but the execution of the detection command failed (nonzero exit code, or any other errors), then assumes "unix".
   - The result is stored back permanently to the config file, to avoid having to run the same detection command next time.
       - The existing checksum command options was already doing this sort of thing.
   - If explicitely specifying remote shell option, e.g. command line option `--sftp-shell-type=cmd`, then this will be used; the auto-detect logic will be skipped if this is first run, or else any previously stored config will be ignored.
       - This option will *not* be saved back to the config file (which is also similar to existing handling of checksum command options).
- When shell is powershell, this PR also added default commands for checksum and about by using built-in PowerShell functionality, so these features will be supported by default without having to configure external commands.
    - It works, but is quite verbose and looks a bit weird in the config file:
      `md5sum_command = &{param($Path);Get-FileHash -Algorithm MD5 -LiteralPath $Path -ErrorAction Stop|Select-Object -First 1 -ExpandProperty Hash|ForEach-Object{"$($_.ToLower())  ${Path}"}}`
      `sha1sum_command = &{param($Path);Get-FileHash -Algorithm SHA1 -LiteralPath $Path -ErrorAction Stop|Select-Object -First 1 -ExpandProperty Hash|ForEach-Object{"$($_.ToLower())  ${Path}"}}`

In addition:
- Added use of vendor-specific VFS statistics extension for about if available. Seems OpenSSH implements this on every platform, at least in the Windows port which I am working on here. This means the usage data can be retrieved directly with api function instead of running shell commands at all. Still falling back to existing shell command if necessary.
- ~Fixes an issue with session handling in hash calculation function. It would release the connection to connection pool premature, before using it to execute the hash command.~ (removed after same change was added to master in 099eff8891d1d7bd081bb1c9218ff5c0863921b2)

Beta builds:
- https://beta.rclone.org/branch/sftp-windows-server/

Notes:
- The autodetection logic is not bullet proof, but the idea is that it works in most cases and makes it convenient for users, and one can always override it from config in the few cases it fails.

Implementation details:
- Shell autodetection:
  - It executes `echo ${ShellId}%ComSpec%`, which seems to be a single command to cover most cases.
  - The expected output is:
      - CMD: `${ShellId}C:\WINDOWS\system32\cmd.exe`
      - PowerShell: `Microsoft.PowerShell%ComSpec%`
      - Bash: `%ComSpec%`
  - Based on this we can deduce:
      - If output has prefix Microsoft.PowerShell I'm quite confident its PowerShell.
      - If output does not have suffix %ComSpec% I'm pretty sure its cmd. Only cmd has the %variable% syntax, as far as I know.
          - Could probably in theory have a cmd without ComSpec environment variable set, and then it will not be detected and fallback to Unix, but I think this is rare - and then user can always force a specific shell type in config file or command-line argument.
      - No explicit check for Unix shell, this is the fallback if the cmd or powershell is not detected.
  - If connecting to an "rclone serve sftp" instance it will always return `${ShellId}%ComSpec%`, and fall-back to unix shell, but this is correct as rclone supports "unix shell api" in this mode (it responds to df, md5sum, etc, and unescapes unix shell escaping performed by sftp remote).
  - If configuring OpenSSH on Windows to use `ssh-shellhost.exe`, the autodetection logic fails, and user must force "cmd".
     - The echo command returns with exit code 255, and on stdout: "ssh-shellhost cannot run 'echo ${ShellId}%ComSpec%', error: 2"
- Shell path conversion:
   - If remote shell is powershell or cmd, then server is probably Windows. The sftp package converts everything to POSIX paths: Forward slashes, and absolute paths starts with a slash. An absolute path on a Windows server will then look like this "/C:/Windows/System32". We must remove the "/" prefix to make this a valid for shell commands. In case of PowerShell there is a possibility that it is a Unix server, with PowerShell Core shell, but assuming root folders with names such as "C:" are rare, we just take this risk,  option path_override can always be used to work around corner cases.
- When implementing shell path conversion, a possible issue with the about command and override_path option was also fixed.
   - Hashsum commands joined override_path with path to the file object relative to fs root.
      - E.g. with `rclone sha1sum remote:/Path/To/File.txt --sftp-path-override=/a/b/c` the executed command is :
         ```
         sha1sum /a/b/c/File.txt
         ```
   - About command joined override_path with root.
       - E.g. with `rclone about remote:/Path/To/File.txt --sftp-path-override=/a/b/c` the executed command would be wrong:
          ```
          df -k /a/b/c/Path/To/File.txt
          ```
       - Normally about command is run without path, e.g. `rclone about remote:`, and in these cases the result would be correct:
          ```
          df -k /a/b/c
          ```

TODO:
- Name of the option: `shell`, `shell_type`, `remote_shell`, `remote_shell_type`,...? Initially used `remote_shell`, currently `shell_type`... I can't make up my mind which to use, so suggestions are welcome.
- If OpenSSH is configured to use ssh-shellhost.exe the shell autodetection fails.
  - This is an included OpenSSH ShellHost, seems to be a VT100 adapter over Windows Console used for compatibility before support for this was added to the built-in console in Windows 10.
  - When I tested this on a Win10, the autodetect logic failed so it ended up being set to "unix", but for paths to be escaped correctly with an md5sum_command etc it must be forced to "cmd". Will see if there is something to do, but if not I think/hope we can just leave this as a "known imperfection" (as my project manager likes to call it), as this seems to be a legacy shell setup anyway.
- Test with bash shell on Windows? E.g. Cygwin, MinGW or WSL..
- Evaluate logic with saving back autodetected value to config file.
   - With the built-in powershell sum commands, there is a dependency between `shell_type` and `md5sum_command`/`sha1sum_command`. If running with `--sftp-shell-type=powershell` first time, then `md5sum_command` and `sha1sum_command` will get automatic values which are powershell commands and be saved to config file. But `shell_type` is not saved to config file. So if next time running with `--sftp-shell-type=cmd`, or without this option and it is autodetected to something different than powershell, then the `md5sum_command` and `sha1sum_command` that will be picked up from config file is not valid.
       - But perhaps such cases can be considered user error?

OLD:
- Is "built-in" support for about-command with cmd shell possible?
    - There is `fsutil volume diskfree c:`, but it requires admin privileges...?
    - Could re-use the same as powershell, just execute it explicitely via powershell.exe.
    - Less important after support for VFS statistics extension feature was introuced, this works regardless of shell - for servers that has this extension.
- String encoding
    - UTF-8
       - Everything should be UTF-8, Windows OpenSSH server seems to do conversion as needed.
       - When executing remote shell command for checksum and path contains non-ascii characters, will it always be correct then?
      - Have done *some* testing with special characters, haven't seen any issues so far - but there may be a ton of variables in play here..
    - The sftp backend does not seem to use rclone's internal encoding, with Unicode replacements, and functions FromStandardName/ToStandardName...?
    
Fixes #5758

#### Checklist


albertony:
New version with a fix to handle relative paths:
- In previous version the following did not work:
  ```
  rclone sha1sum -vv remote:some/path/to/file.txt
  ```
- Reason was the path was made absolute, but then as a POSIX path with leading slash, e.g. `/C:/Users/aaa/some/path/to/file.txt`, which the sha1sum utility does not handle.
- Fix is to detect and remove leading slash from such absolute paths - but only when shell type is cmd or powershell.

albertony:
Added another version, fixing a corner case from the previos fix: When referring directly to the top level `/`.

TLDR;
When executing `rclone sha1sum local_sftp:/` the root path in the fs is `/`. My original fix looked for `/C:` kind of root paths, so it would not match. But when traversing, it will go from `/` to `/C:` and further down to directories, and then it ended up with the bad prefix in command line again: `sha1sum "/C:/test.txt"`. Most other paths would work though, such as `rclone sha1sum local_sftp:/Temp`, as it would travererse from `/Temp/` (successfully, I guess `/Temp` is an alias for `/C:/Temp`).

ivandeex (approved):

ivandeex:
Manual test passed on Windows-8.1+CMD - see linked ticket.

albertony:
@ncw, @ivandeex: I have one question to the existing code. Not directly to this PR but concerns executing shell commands in general - which is part of the solution in this PR:

What is the sequence of: getSftpConnection, putSftpConnection, sshClient.NewSession() and session.Run

 - When executing about command, and testing possible hash commands, the sequence is:
    - getSftpConnection
    - NewSession
    - Run
    - putSftpConnection
https://github.com/rclone/rclone/blob/a8059b8a90ec980b9c2112d571a9eb3e3a638526/backend/sftp/sftp.go#L1095-L1113
- But the Hash function does this:
  - getSftpConnection
  - NewSession
  - putSftpConnection
  - Run
https://github.com/rclone/rclone/blob/a8059b8a90ec980b9c2112d571a9eb3e3a638526/backend/sftp/sftp.go#L1258-L1275

I'm wondering if it is correct to call putSftpConnection before using the session with session.Run?

albertony:
Finally had some time to revisit this one, with some final adjustments to make it ready for review. Still some TODOs (as always), but I think the current functionality is "close enough", and that it is better to instead (if necessary) continue on these later. Its been a while since I started this, so its about time to get it done.

backend/sftp/sftp.go
```diff
@@ -1285,33 +1463,73 @@ func (o *Object) Hash(ctx context.Context, r hash.Type) (string, error) {
...
+// when sending it to a Unix shell.
+func unixShellEscape(str string) string {
+	safe := unixShellEscapeRegex.ReplaceAllString(str, `\$0`)
 	return strings.Replace(safe, "\n", "'\n'", -1)
 }
 
+// quoteOrEscapeShellPath makes path a valid string argument in configured shell
+func (f *Fs) quoteOrEscapeShellPath(shellPath string) string {
+	if f.shellType == "powershell" {
+		return "'" + strings.Replace(shellPath, "'", "''", -1) + "'"
```
ncw, line 1491: This is a potential security problem if we get this wrong BTW so we need to make sure it is 100% correct!

backend/sftp/sftp.go
```diff
@@ -1285,33 +1463,73 @@ func (o *Object) Hash(ctx context.Context, r hash.Type) (string, error) {
...
+	}
+	if f.shellType == "cmd" {
+		// By assuming str is a windows filepath we can simply wrap in double quotes,
+		// no characters that would need escaping are legal in paths.
+		return "\"" + shellPath + "\""
```
ncw, line 1496: ...as is this

backend/sftp/sftp_internal_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestShellEscape(t *testing.T) {
+func TestUnixShellEscape(t *testing.T) {
```
ncw, line 13: Add a test for the new shell escape things since they are security critical?

backend/sftp/sftp.go
```diff
@@ -788,37 +816,84 @@ func NewFsWithConnection(ctx context.Context, f *Fs, name string, root string, m
...
-	if err != nil {
-		fs.Debugf(f, "Failed to read current directory - using relative paths: %v", err)
-	} else if !path.IsAbs(f.root) {
-		f.absRoot = path.Join(cwd, f.root)
-		fs.Debugf(f, "Using absolute root directory %q", f.absRoot)
+	// Check remote shell type, try to auto-detect if not configured and save to config for later
+	if f.opt.ShellType != "" {
+		f.shellType = f.opt.ShellType
+		fs.Debugf(f, "Shell type %q from config", f.shellType)
+	} else {
```
ncw, line 823: This means that all sftp backends will make an additional ssh command on startup. Often sftp is run without a config file so it won't be able to save the config.

I'm not sure this is an actual problem just thought I'd mention it!

ncw (commented):
This is looking great - It would be good to get it merged.

I think the security critical shell escaping needs tests and pointers to the docs on PowerShell and CMD escaping, other than that I think it is fine :-)

backend/sftp/sftp.go
albertony: That is a fair point!

To avoid that I guess we would have to change it to "opt-in". Either default no shell access, but that would be a breaking change, so for backwards compat reasons maybe would have to assume unix shell by default. And then user could explicitely specify other shell types. And then this auto-detect logic could either be removed or moved to a separate option such as `--auto-detect-shell-or-something-like-that`.... Problem is when running the hash command, which rely on shell access, it would maybe fail with strange error messages before user realises he needs to configure different shell type.

Another alternative could be to try some sort of "delay-loading" this auto-detection part. That is, not run it on "startup", but first time it is needed - probably from the Hash command.

Not sure how these would work out, needs more thinking and experimenting...

backend/sftp/sftp.go
albertony: Ah... My assumption, I guess, was that at this point it is a valid path, so any crazy things like that would not get so far. But that is not necessarily correct!

I actually managed to start notepad on my server - hidden in session 0, since sshd is running as service user:

`rclone about "local_sftp:\"^&\"notepad"`

Though, I had to cheat a bit in order to get my "scary news headline" right..:
- Commented out StatVFS extension. It is supported on windows openssh, so then it would not execute shell command on about.
- Commented out a check that aborts with error when shell type is "cmd" instead of trying to run the "df -k" shell command.

So not really a *proof* of concept, but probably risky enough to warrant a change...

albertony: This could end up being a bit of a challenge.. Previous comments relates to cmd, not powershell. But user can change the shell_type config, and by that also the quoting/escaping rules applied. So even if making "perfect" escaping/quoting of cmd strings when server is set up with cmd shell, if user changes the shell_type into something else (e.g. powershell) then suddenly different escaping/quoting will be applied by rclone - but server will still execute them on cmd, so there may be "loop holes"..

But, to take one step back: What is the security risk here? If rclone is able to execute a harmfull command through this, then user could do the same using a plain ssh shell instead, couldn't he? So isn't the primary concern to avoid file paths and other valid arguments to legitimate shell commands **accidentally** being treated as something else, breaking the command, and in worst case have side-effects? E.g. the example above relies on insertion of double quote into the argument, but this is illegal in filenames on Windows, and then since the argument to about command should be a file path this is not something we absolutely must prevent?

backend/sftp/sftp.go
albertony: Oh, and the third alternative (or alternative zero), the current implementation, already have an "opt-out" switch, in that user can explicit set "shell_type", e.g. when running without config file as you say. Then the extra shell command will not be executed on startup. With an extra note in the docs, this could be considered an acceptable solution.

ncw: I think a few docs to address this is fine.

backend/sftp/sftp.go
ncw: > So isn't the primary concern to avoid file paths and other valid arguments to legitimate shell commands accidentally being treated as something else, breaking the command, and in worst case have side-effects?

Yes that is the primary concern.

However experience has taught me that this kind of problem can be escalated into a security problem...

> E.g. the example above relies on insertion of double quote into the argument, but this is illegal in filenames on Windows,

In a perfect world we never try to write a non Windows filename to a windows disk. However there are a lot of code paths in rclone and there might be one where we do, leaving the door open for an exploit.

Given that this is a security critical bit of code we should have a test to make sure the escaping is working as we expect for all 3 different shell types.

The shell escape routines should check their assumptions, so if we are expecting no quotes in the string before escaping then we should check for it and return an error first.

A note in the doc that configuring the wrong shell type might lead to security problems would probably be sensible.

albertony: > However experience has taught me that this kind of problem can be escalated into a security problem...

Ah, yes, that I can believe! Such claims can be real, while some are not - though adding the security issue label to ones opinion is hard to argue against without it backfiring.

> The shell escape routines should check their assumptions, so if we are expecting no quotes in the string before escaping then we should check for it and return an error first.

> A note in the doc that configuring the wrong shell type might lead to security problems would probably be sensible.

I've given this more consideration, and I'm pretty sure that checking for double quotes and wrapping in double quotes should be enough on cmd, and wrapping in single quotes doubling any single-quotes within the string should be good on  powershell. But hard to be 100% sure. E.g. for cmd shell; we have sshd.exe executing cmd.exe with command line arguments, which then runs one of the arguments as a shell command, which then runs a command-line utility expecting the last part being a filepath argument...
- C:\Program Files\OpenSSH\sshd.exe
    - "c:\windows\system32\cmd.exe" /c "md5sum "C:/Some/path""
        - md5sum "C:/Some/path"

Added some simple tests, as well as tried to document it clearly (Made the section about Windows OpenSSH server into a more generic section about Shell access, since it was mostly that anyway, and linked to a subsection with the important points from the backend intro text.)

What do you think?

ncw: I think that is looking very good now - nice tests and good to have a think about potential exploits.

PR #5781, opened 2021-11-03
Hashsum stdin

albertony:

Add support for hashing from stdin from rclone hashum/md5sum/sha1sum commands, like standard linux tools:

> With no FILE, or when FILE is -, read standard input.

Output format:

```
2f8f3230bbc42e379a1554ca3419d46d  -
```

Then it can be used as a drop-in replacement for md5sum in existing sftp backend code.

#### Checklist

### DIFF
--- a/cmd/hashsum/hashsum.go
+++ b/cmd/hashsum/hashsum.go
@@ -2,7 +2,6 @@ package hashsum
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 
@@ -26,11 +25,11 @@ var (
 func init() {
 	cmd.Root.AddCommand(commandDefinition)
 	cmdFlags := commandDefinition.Flags()
-	AddHashFlags(cmdFlags)
+	AddHashsumFlags(cmdFlags)
 }
 
-// AddHashFlags is a convenience function to add the command flags OutputBase64 and DownloadFlag to hashsum, md5sum, sha1sum
-func AddHashFlags(cmdFlags *pflag.FlagSet) {
+// AddHashsumFlags is a convenience function to add the command flags OutputBase64 and DownloadFlag to hashsum, md5sum, sha1sum
+func AddHashsumFlags(cmdFlags *pflag.FlagSet) {
 	flags.BoolVarP(cmdFlags, &OutputBase64, "base64", "", OutputBase64, "Output base64 encoded hashsum")
 	flags.StringVarP(cmdFlags, &HashsumOutfile, "output-file", "", HashsumOutfile, "Output hashsums to a file rather than the terminal")
 	flags.StringVarP(cmdFlags, &ChecksumFile, "checkfile", "C", ChecksumFile, "Validate hashes against a given SUM file instead of printing them")
@@ -41,7 +40,7 @@ func AddHashFlags(cmdFlags *pflag.FlagSet) {
 func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
 	out, err = os.Create(filename)
 	if err != nil {
-		err = fmt.Errorf("Failed to open output file %v: %w", filename, err)
+		err = fmt.Errorf("failed to open output file %v: %w", filename, err)
 		return nil, nil, err
 	}
 
@@ -53,6 +52,32 @@ func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
 	}
 
 	return out, close, nil
+}
+
+// CreateFromStdinArg checks args and produces hashsum from standard input if it is requested
+func CreateFromStdinArg(ht hash.Type, args []string, startArg int) (bool, error) {
+	var stdinArg bool
+	if len(args) == startArg {
+		// Missing arg: Always read from stdin
+		stdinArg = true
+	} else if len(args) > startArg && args[startArg] == "-" {
+		// Special arg: Read from stdin only if there is data available
+		if fi, _ := os.Stdin.Stat(); fi.Mode()&os.ModeCharDevice == 0 {
+			stdinArg = true
+		}
+	}
+	if !stdinArg {
+		return false, nil
+	}
+	if HashsumOutfile == "" {
+		return true, operations.HashSumStream(ht, OutputBase64, os.Stdin, nil)
+	}
+	output, close, err := GetHashsumOutput(HashsumOutfile)
+	if err != nil {
+		return true, err
+	}
+	defer close()
+	return true, operations.HashSumStream(ht, OutputBase64, os.Stdin, output)
 }
 
 var commandDefinition = &cobra.Command{
@@ -67,6 +92,11 @@ By default, the hash is requested from the remote.  If the hash is
 not supported by the remote, no hash will be returned.  With the
 download flag, the file will be downloaded from the remote and
 hashed locally enabling any hash for any remote.
+
+This command can also hash data received on standard input (stdin),
+by not passing a remote:path, or by passing a hyphen as remote:path
+when there is data to read (if not, the hypen will be treated literaly,
+as a relative path).
 
 Run without a hash to see the list of all supported hashes, e.g.
 
@@ -83,8 +113,6 @@ Note that hash names are case insensitive and values are output in lower case.
 		if len(args) == 0 {
 			fmt.Print(hash.HelpString(0))
 			return nil
-		} else if len(args) == 1 {
-			return errors.New("need hash type and remote")
 		}
 		var ht hash.Type
 		err := ht.Set(args[0])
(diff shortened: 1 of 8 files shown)


cmd/md5sum/md5sum.go
```diff
@@ -27,9 +28,22 @@ By default, the hash is requested from the remote.  If MD5 is
...
+This command can also hash data received on standard input (stdin),
+by not passing a remote:path, or by passing a hyphen as remote:path
+when there is data to read (if not, the hypen will be treated literaly,
+as a relative path).
 `,
-	Run: func(command *cobra.Command, args []string) {
-		cmd.CheckArgs(1, 1, command, args)
+	RunE: func(command *cobra.Command, args []string) error {
+		cmd.CheckArgs(0, 1, command, args)
+		if len(args) == 0 {
```
ncw, line 39: You could factor these 8 lines into hashsum as they are repeated 3 times so it is probably worth it. (Pass in `startArg` which will be 0 here and 1 in hashsum).

Can we give it a good enough name is the questions?

cmd/hashsum/hashsum.go
```diff
@@ -55,6 +59,32 @@ func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
 	return out, close, nil
 }
 
+// HashStream produces hash from a stream
+func HashStream(ht hash.Type, outputBase64 bool, in io.ReadCloser) error {
+	hasher, err := hash.NewMultiHasherTypes(hash.NewHashSet(ht))
+	if err != nil {
+		return errors.Wrap(err, "Hash unsupported")
```
ncw, line 66: According to Go style rules, error messages should start with a small letter.

This is because they get chained "error a: error b: error c" then eventually logged "Something went wrong: error a: error b: error c"

cmd/hashsum/hashsum.go
```diff
@@ -55,6 +59,32 @@ func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
...
+	written, err := io.Copy(hasher, in)
+	fs.Debugf(nil, "Creating %s hash of %d bytes read from input stream", ht, written)
+	if err != nil {
+		return errors.Wrap(err, "Failed to copy input to hasher")
+	}
+	byteSum, err := hasher.Sum(ht)
+	if err != nil {
+		return errors.Wrap(err, "Hasher returned an error")
+	}
+	sum := hex.EncodeToString(byteSum)
```
ncw, line 77: This printing logic is duplicated from operations.go isn't it?

https://github.com/rclone/rclone/blob/b323bf34e2313f692dd013921bf87079bea73bf8/fs/operations/operations.go#L1031-L1040

I wonder whether we should factor that logic into a function and call it here.

The logic is probably complicated enough and using syncPrintf is probably a good idea as the other printing in this routine goes through there.

cmd/hashsum/hashsum.go
```diff
@@ -55,6 +59,32 @@ func GetHashsumOutput(filename string) (out *os.File, close func(), err error) {
 	return out, close, nil
 }
 
+// HashStream produces hash from a stream
```
ncw, line 62: This function should go in `operations/operations.go`

We can then write a unit test for it in there where we have lots of useful framework.

In general we try to do the absolute minimum in the cmd/ files and do everything in operations where it is easier to test.

cmd/md5sum/md5sum.go
```diff
@@ -27,9 +28,22 @@ By default, the hash is requested from the remote.  If MD5 is
...
+			return hashsum.HashStream(hash.MD5, hashsum.OutputBase64, os.Stdin)
+		}
+		if args[0] == "-" {
+			if fi, _ := os.Stdin.Stat(); fi.Mode()&os.ModeCharDevice == 0 {
```
ncw, line 43: What is the purpose of the `fi.Mode()&os.ModeCharDevice == 0` check? If the user passes `-` don't we use stdin regardless?

ncw (commented):
This is a nice idea :-)

I put lots of comments inline!

cmd/hashsum/hashsum.go
albertony: I can change that. I was aware of it, but the code I copied from (existing hash functionality) had these exact messages, and also other errors within the same file used uppercase, so I chose to keep it. In general, this is something I've been thinking about several times, as there is a good mix throughout the code base. But sure, at least new code can do the right thing! :)

cmd/md5sum/md5sum.go
albertony: It checks if there is data on stdin. Same thing was done by obscure command. It means you can have a file named `-` and refer to it with relative path: `rclone md5sum "-"`. If you send it data it will be hashed instead of the file: `echo "something" | rclone md5sum -`

cmd/hashsum/hashsum.go
albertony: My reasoning for not using syncPrintf was that since this is a "standalone" operation, not part of listing, no stats, or anything (like the original hashing), it was not necessary with syncPrintf.

albertony:
> This is a nice idea :-)
> 
> I put lots of comments inline!

Thanks! I agree with your suggestions for refactoring, I will spend some time on that - now that I know it has a chance to survive! :)

albertony:
@ncw: Suggestion: Move conversion from byte hash to string, and then both hex and base64 encoded, into MultiHasher object?

It did already have `Width` returning hex width. Extended that with option to return base64 width. And then added new related function `SumString` to produce the actual strings (completes the existing function `Sum` which returns raw byte hash, chose to not just add "optional" arguments to that).

PS: Pushed a version where this is done, but not 100% completed, so don't need a full re-review yet.

cmd/md5sum/md5sum.go
ncw: I guess this would stop you running `rclone md5sum - < /dev/ttyUSB1` but that is probably pretty unlikely!

We'll see if we get any bug reports (I doubt we will!)

ncw:
> @ncw: Suggestion: Move conversion from byte hash to string, and then both hex and base64 encoded, into MultiHasher object?
> 
> It did already have `Width` returning hex width. Extended that with option to return base64 width. And then added new related function `SumString` to produce the actual strings (completes the existing function `Sum` which returns raw byte hash, chose to not just add "optional" arguments to that).
> 
> PS: Pushed a version where this is done, but not 100% completed, so don't need a full re-review yet.

Without looking at the code, I'd say that idea makes sense :-)

cmd/md5sum/md5sum.go
albertony: Oh, that's not something I'm too familiar with.. But you could drop the hypen to achieve this, just use: `rclone md5sum < /dev/ttyUSB1`, as the hypen is optional (same as with the gnu tools). If I change to always treat `-` as stdin, we have the equally unlikely case that someone will have a relative path reference to something named "-"? 50-50 call? If you are 51% on changing it I will do it right away, if not I'll keep it as is  🤷‍♂️

ncw: Let's keep it as is.

ncw:
Is this ready for another review?

albertony:
> Is this ready for another review?

Yes, it is now.

fs/operations/operations.go
```diff
@@ -1028,22 +1028,38 @@ func HashLister(ctx context.Context, ht hash.Type, outputBase64 bool, downloadFl
...
-				sum = base64.URLEncoding.EncodeToString(hexBytes)
-			}
 			syncFprintf(w, "%*s  %s\n", width, sum, o.Remote())
 		}()
 	})
 	wg.Wait()
 	return err
 }
 
+// HashStream produces hash from a stream
```
ncw, line 1043: I don't think HashStream is a good name for this function... I'd expect a function of that name to take a stream and return a hash, not print a hashsum compatible line to `w` :-)

It is a public function so should have a bit more docs. Something like

```go

// HashStream outputs a line compatible with md5sum to w based on the
// input stream in and the hash type ht passed in. If outputBase64 is
// set then the hash will be base64 instead of hexadecimal.

```

I'm struggling to think of a better name though... `HashSumStream` maybe?

ncw (commented):
Code looks good :-)

Inline discussion of naming functions (or painting the bikeshed ;-)

fs/operations/operations.go
albertony: I see your point! But impossible to find a better name.. 

One reason was to use a name similar to the existing function in operations called `HashLister`, which also writes hashsum compatible lines.

There seem to be some mixing of `Hashsum` vs `HashSum`, same with `Checksum` vs `CheckSum`, unless there is a distinction there that I'm not aware of? I was considering your suggestion spelled as `HashsumStream`.. but ended up with your `HashSumStream`.

albertony:
Rebased on master to fix conflicts, and changed naming.

ncw (approved):
LGTM :-)

Please merge.

PR #5786, opened 2021-11-05
lib/http: Add Compress middleware to default http server

### LINKED ISSUE #5777: Activate Gzip compression on the fly with the “Serve” command
https://forum.rclone.org/t/activate-gzip-compression-on-the-fly-with-the-serve-command/27209

1.56.2

Optimize transfer rate via Rclone serve

using on-the-fly gzip compression.

#### How to use GitHub

innovate-invent:
Requires that all services set w.Header().Set("Content-Type", http.DetectContentType(content)) to enable compression

Waiting on #5294 and other services to be refactored. Webdav is half done but I am waiting to see what happens with #5294 

Compresses data downloaded from http server when client sets `Accept-Encoding` header in request.

Resolves #5777 

#### Checklist


lib/http/http.go
```diff
@@ -353,6 +354,7 @@ func start() error {
 	}
 	defaultServer = s.(*server)
 	defaultServer.Serve()
+	defaultServer.Router().Use(middleware.Compress(5, CompressibleTypes...))
```
klauspost, line 357: Use better-than-stdlib compressors:

```suggestion
	comp := middleware.Compress(5, CompressibleTypes...)
	comp.SetEncoder("gzip", func(w io.Writer, level int) io.Writer {
		gw, err := gzip.NewWriterLevel(w, level)
		if err != nil {
			return nil
		}
		return gw
	})
	comp.SetEncoder("deflate", func(w io.Writer, level int) io.Writer {
		dw, err := flate.NewWriter(w, level)
		if err != nil {
			return nil
		}
		return dw
	})
	defaultServer.Router().Use(comp)
```

Import
```
	"github.com/klauspost/compress/flate"
	"github.com/klauspost/compress/gzip"
```

Deflate is typically not used, but it seems like this middleware uses it, so maybe it is ok.

innovate-invent: I am not sure what this accomplishes over the default implementation:

https://github.com/go-chi/chi/blob/c9e87efe9691a63d6a89de8bbd16b04fe4d6640e/middleware/compress.go#L94-L123

Can you clarify a bit? The encoder used is specified by the client, so long as the server supports it.

klauspost: @innovate-invent You replace the compressors with the other imports. They are faster ,typically 1.8x the throughput on web data.

innovate-invent: Understood

innovate-invent:
Do we want to add a global flag to set the compression level?

ncw:
@innovate-invent - we've merged all the lib/http bits and moved over it completely. It changed a bit in the process so this no longer applies. Do you want to bring it up to date?

acme:
Can we close this now?

PR #5791, opened 2021-11-06
Set proper exit code for errors that are not low-level retried (e.g. size/timestamp changing)

### LINKED ISSUE #5785: Incorrect error code being returned when file sizes are changing
The error code returned by rclone is 1 when a file is unable to be sync'ed because the file size is changing.  For example:

> dd if=/dev/zero of=file bs=1 count=10240000 &
> rclone sync file file-copy --use-json-log -vv --log-file /tmp/blah.log
> echo $?

It seems like this is the incorrect error code number, perhaps 6 would make more sense.

rclone v1.55.1-DEV
- os/type: linux
- os/arch: amd64
- go/version: go1.15.5
- go/linking: dynamic
- go/tags: none

For more details on system, versions, etc., see: https://forum.rclone.org/t/non-jsoned-output-when-using-use-json-log/27299/9

albertony:

Setting the following exit code when error is NoLowLevelRetryError, same as for NoRetryError:

> 6 - Less serious errors (like 461 errors from dropbox) (NoRetry errors)

Only case where this error is returned, is currently in local backend, when size or timestamp are changing during sync.

Fixes #5785

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -491,7 +491,7 @@ func resolveExitCode(err error) {
 		os.Exit(exitcode.TransferExceeded)
 	case fserrors.ShouldRetry(err):
 		os.Exit(exitcode.RetryError)
-	case fserrors.IsNoRetryError(err):
+	case fserrors.IsNoRetryError(err), fserrors.IsNoLowLevelRetryError(err):
 		os.Exit(exitcode.NoRetryError)
 	case fserrors.IsFatalError(err):
 		os.Exit(exitcode.FatalError)


cmd/cmd.go
```diff
@@ -491,7 +491,7 @@ func resolveExitCode(err error) {
 		os.Exit(exitcode.TransferExceeded)
 	case fserrors.ShouldRetry(err):
 		os.Exit(exitcode.RetryError)
-	case fserrors.IsNoRetryError(err):
+	case fserrors.IsNoRetryError(err), fserrors.IsNoLowLevelRetryError(err):
```
ncw, line 494: Shouldn't that say 

    fserrors.IsNoRetryError(err) || fserrors.IsNoLowLevelRetryError(err)

Or is this some case syntax I'm not aware of...

Turns out it is syntax I'm not aware of: https://go.dev/play/p/i6tDdKufO3L !

ncw (approved):
This looks great - I'll merge now

PR #5792, opened 2021-11-06
backend: Fix union eplus policy returned nil

megatron0000:

Union generated a `panic` when it had to select the least-used-space backend, because the `lusEntries` function returned `nil`.

To solve it: I initialized the `minUsedSpace` variable with `math.MaxInt64` so that the first backend would trigger the condition `if space < minUsedSpace`, thus `entry` would never be `nil` after the loop.

Not that I could find.

#### Checklist

### DIFF
--- a/backend/union/policy/eplus.go
+++ b/backend/union/policy/eplus.go
@@ -39,7 +39,7 @@ func (p *EpLus) lus(upstreams []*upstream.Fs) (*upstream.Fs, error) {
 }
 
 func (p *EpLus) lusEntries(entries []upstream.Entry) (upstream.Entry, error) {
-	var minUsedSpace int64
+	var minUsedSpace int64 = math.MaxInt64
 	var lusEntry upstream.Entry
 	for _, e := range entries {
 		space, err := e.UpstreamFs().GetFreeSpace()


ncw:
This looks like it never could have worked... Would you agree?

megatron0000:
I would. In my case, the panic did not happen when first syncing a file to the union remote, but it did happen when the file was locally modified and had to be update in the remote. 

My guess from partially reading the source code is: first, the union defines two versions of action/create/search. These two versions are `action` and `actionEntries`, `create` and `createEntries`,  `search` and `searchEntries`. For example below, from `union.go`:

https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/backend/union/union.go#L723-L729

The first version calls `policy.Action`, while the second version calls `policy.ActionEntries`. In the case of eplus, these implementations are as below (snippet from `eplus.go`):

https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/backend/union/policy/eplus.go#L58-L76

The first implementation calls `p.lus` while the second calls `p.lusEntries`. Then we finally take a look at these two function implementations (below):

https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/backend/union/policy/eplus.go#L21-L56

`lus` was correctly implemented. `lusEntries` was flawed. So we only observe the panic when someone calls the union's `actionEntries`/`createEntries`/`searchEntries` (because they end up invoking the flawed `lusEntries`, as seen).

But I didn't dig deeper to discover why two versions exist in the first place... (why  `X` and `XEntries`)

ncw (approved):
Thanks for that fix, and also for the analysis.

I think we need to do some more work in the union integration tests - this really should have been caught earlier.

I'll merge it now - thank you.

PR #5796, opened 2021-11-08
cmd/ncdu: display correct path in delete confirmation dialog

r-ricci:
If the remote on the command line is "remote:subdir", when
deleting "filename", the confirmation message shows the path
"remote:subdirfilename". Using path.Join() fixes this.
A superfluous '/' will be added after the ':' even if the
remote is just "remote:", but that's still a valid syntax.

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -19,6 +19,7 @@ import (
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/cmd/ncdu/scan"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/fspath"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
 )
@@ -512,7 +513,7 @@ func (u *UI) delete() {
 		}
 		u.popupBox([]string{
 			"Delete this file?",
-			u.fsName + dirEntry.String()})
+			fspath.JoinRootPath(u.fsName, dirEntry.String())})
 	} else {
 		u.boxMenuHandler = func(f fs.Fs, p string, o int) (string, error) {
 			if o != 1 {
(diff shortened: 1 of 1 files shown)


albertony:
> A superfluous '/' will be added after the ':' even if the
> remote is just "remote:", but that's still a valid syntax.

With [some](https://rclone.org/docs/#remote-path-to-dir-1) exceptions. But this is just a string shown in GUI right? So it may not be a catastrophe if its not 100% correct..

ncw:
I think this is probably the correct joining function

https://github.com/rclone/rclone/blob/b16f603c51fe451a4752add654ac5a621df67076/fs/fspath/path.go#L308-L316

4ricci:
> I think this is probably the correct joining function
> 
> https://github.com/rclone/rclone/blob/b16f603c51fe451a4752add654ac5a621df67076/fs/fspath/path.go#L308-L316

Yes, this is exactly what I was looking for.

I think that using a join function which is aware of `rclone`'s syntax is better than using `path.Join()` or `+`, even when the result is the same (for consistency). Should `fspath.JoinRootPath()` be used on the following lines too?
https://github.com/rclone/rclone/blob/7366e97dfc80b2ee8c0482d8d5cc7e67b8cde770/cmd/ncdu/ncdu.go#L645
https://github.com/rclone/rclone/blob/7366e97dfc80b2ee8c0482d8d5cc7e67b8cde770/cmd/ncdu/ncdu.go#L728

ncw:
> I think that using a join function which is aware of `rclone`'s syntax is better than using `path.Join()` or `+`, even when the result is the same (for consistency).

Yes it should always do the right thing.

> Should `fspath.JoinRootPath()` be used on the following lines too?
> 
> https://github.com/rclone/rclone/blob/7366e97dfc80b2ee8c0482d8d5cc7e67b8cde770/cmd/ncdu/ncdu.go#L645

Yes here

> https://github.com/rclone/rclone/blob/7366e97dfc80b2ee8c0482d8d5cc7e67b8cde770/cmd/ncdu/ncdu.go#L728

Not here, but that line should really be

```go
fsName: fs.ConfigString(f)
```

buengese:
@ncw can we merge this?

ncw (approved):
LGTM :-)

I'll merge now

Issue #5801, opened 2021-11-10
Base32768 file name encoding for crypt backend


tinytangent:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/base32768-to-compress-filename-length/13202
https://forum.rclone.org/t/whats-going-on-with-the-proposal-about-base32768-file-name-encoding/27341

#### What is your current rclone version (output from `rclone version`)?
v1.56.2

#### What problem are you are trying to solve?
The OneDrive backend has a very strict file name length limitation. When used together with the crypt backend, the file name encryption feature will generate a even longer name with the default base32 encoding, making some files with longer names failed to be uploaded.

However there is a possible workaround for this problem. The OneDrive backend stores characters in UTF-16 encoding, therefore the default base32 encoding waste a lot of spaces. The base32768 encoding would instead encode the file name more efficiently.

#### How do you think rclone should be changed to solve that?
Support base32768 file name encoding for the crypt backend.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

PR #5802, opened 2021-11-10
Support base64 and base32768 file name encoding for crypt.

tinytangent:

This PR add support for base64 and base32768 file name encoding for crypt. This resolves https://github.com/rclone/rclone/issues/5801 and relieve the problem of file name length being too long when encrypted.

It is based on the commit 78cab2ba2071dd4c58a34afa22486ad1dec306fc by @Max-Sum . For other changes in https://github.com/Max-Sum/rclone/commits/compress_crypted_filename, I' m worried that some of them might break backward compatibility or involving security issue, and I' m not very familiar in cryptography.

https://forum.rclone.org/t/whats-going-on-with-the-proposal-about-base32768-file-name-encoding/27341
https://forum.rclone.org/t/base32768-to-compress-filename-length/13202

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -7,6 +7,7 @@ import (
 	gocipher "crypto/cipher"
 	"crypto/rand"
 	"encoding/base32"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -16,6 +17,7 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"github.com/Max-Sum/base32768"
 	"github.com/rclone/rclone/backend/crypt/pkcs7"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
@@ -114,6 +116,57 @@ func (mode NameEncryptionMode) String() (out string) {
 	return out
 }
 
+// fileNameEncoding are the encoding methods dealing with encrypted file names
+type fileNameEncoding interface {
+	EncodeToString(src []byte) string
+	DecodeString(s string) ([]byte, error)
+}
+
+// caseInsensitiveBase32Encoding defines a file name encoding
+// using a modified version of standard base32 as described in
+// RFC4648
+//
+// The standard encoding is modified in two ways
+//  * it becomes lower case (no-one likes upper case filenames!)
+//  * we strip the padding character `=`
+type caseInsensitiveBase32Encoding struct{}
+
+// EncodeToString encodes a strign using the modified version of
+// base32 encoding.
+func (caseInsensitiveBase32Encoding) EncodeToString(src []byte) string {
+	encoded := base32.HexEncoding.EncodeToString(src)
+	encoded = strings.TrimRight(encoded, "=")
+	return strings.ToLower(encoded)
+}
+
+// DecodeString decodes a string as encoded by EncodeToString
+func (caseInsensitiveBase32Encoding) DecodeString(s string) ([]byte, error) {
+	if strings.HasSuffix(s, "=") {
+		return nil, ErrorBadBase32Encoding
+	}
+	// First figure out how many padding characters to add
+	roundUpToMultipleOf8 := (len(s) + 7) &^ 7
+	equals := roundUpToMultipleOf8 - len(s)
+	s = strings.ToUpper(s) + "========"[:equals]
+	return base32.HexEncoding.DecodeString(s)
+}
+
+// NewNameEncoding creates a NameEncoding from a string
+func NewNameEncoding(s string) (enc fileNameEncoding, err error) {
+	s = strings.ToLower(s)
+	switch s {
+	case "base32":
+		enc = caseInsensitiveBase32Encoding{}
+	case "base64":
+		enc = base64.RawURLEncoding
+	case "base32768":
+		enc = base32768.SafeEncoding
+	default:
+		err = fmt.Errorf("Unknown file name encoding mode %q", s)
+	}
+	return enc, err
+}
+
 // Cipher defines an encoding and decoding cipher for the crypt backend
 type Cipher struct {
 	dataKey        [32]byte                  // Key for secretbox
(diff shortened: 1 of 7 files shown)


backend/crypt/cipher.go
```diff
@@ -114,22 +117,65 @@ func (mode NameEncryptionMode) String() (out string) {
 	return out
 }
 
-// Cipher defines an encoding and decoding cipher for the crypt backend
```
ncw, line 117: This comment didn't want to get deleted!

backend/crypt/cipher.go
```diff
@@ -114,22 +117,65 @@ func (mode NameEncryptionMode) String() (out string) {
...
+// StrEncoding is encoding methods dealing with strings
+type StrEncoding interface {
+	EncodeToString(src []byte) string
+	DecodeString(s string) ([]byte, error)
+}
+
+type CaseInsensitiveBase32Encoding struct {
```
ncw, line 126: This seems over complicated to me....

I'd suggest you define this (note lower case so we don't make a public symbol)

```go
type caseInsensitiveBase32Encoding struct {}
```

Then define the two methods

```
func (_ caseInsensitiveBase32Encoding) EncodeToString(src []byte) string {
// contents the same as encodeFileName in original source
}
func (_ caseInsensitiveBase32Encoding) DecodeString(s string) ([]byte, error) {
// contents the same as decodeFileName in original source
}
```

And finally check it satisfies the interface

```
var _ StrEncoding = caseInsensitiveBase32Encoding{}
```

backend/crypt/cipher.go
```diff
@@ -114,22 +117,65 @@ func (mode NameEncryptionMode) String() (out string) {
 	return out
 }
 
-// Cipher defines an encoding and decoding cipher for the crypt backend
+// StrEncoding is encoding methods dealing with strings
+type StrEncoding interface {
```
ncw, line 121: This should be called `filenameEncoding` - initial lower so it isn't exported and more specific name

backend/crypt/cipher.go
```diff
@@ -114,22 +117,65 @@ func (mode NameEncryptionMode) String() (out string) {
...
+func (enc CaseInsensitiveBase32Encoding) DecodeString(s string) ([]byte, error) {
+    return enc.enc.DecodeString(strings.ToLower(s))
+}
+
+// NewNameEncoding creates a NameEncoding from a string
+func NewNameEncoding(s string) (enc StrEncoding, err error) {
+	s = strings.ToLower(s)
+	switch s {
+	case "base32":
+		base32hex := base32.NewEncoding(encodeHexLower)
```
ncw, line 149: ...this then becomes

    enc = caseInsensitiveBase32Encoding{}

ncw (commented):
I put some initial comments inline.

This should probably also have some tests in `backend/crypt/crypt_test.go` to run the backend tests against the new file name encoding.

I'm going to allow the CI to run now - expect some linting problems you'll need to fix up!

backend/crypt/cipher.go
tinytangent: Sorry about this. Fixed in 79817be

backend/crypt/cipher.go
tinytangent: Indeed this change is modifying codes that already work. I reverted the base32 implementation in 79817be.

tinytangent:
I also added some new fstests in `backend/crypt/crypt_test.go` for the newly added base64 and base32768 encoding in 046656d.

tinytangent:
Linting problems should now be fixed in fb0fb43.

ncw (commented):

Can you rebase it on master, fix the conflicts and squash it down to two commits - one for Max-Sum's work and one for your work.

(can you make sure Max-Sum commit compiles on its own - if it doesn't fix it!)

I'll merge it then - thank you.

PS If any of this git stuff is too much for you then I can do it easily enough.

tinytangent:
Thank you~

I have reorganized the commit history to make it cleaner. So now d2e180f contains all the code changes including go.sum and go.mod, and it compiles on it own  (but will fail most of test cases due to internal interfaces changes).

1bd9007 updates test cases and docs.

tinytangent:
There is another thing to note: Currently there is a failing test case in crypt, but it is unlikely introduced by my changes because it is also happening on the master branch, it looks like this:

```
--- FAIL: TestNoDataObfuscate (0.53s)
    fstests.go:418: Using remote "TestCrypt4:"
    --- FAIL: TestNoDataObfuscate/FsMkdir (0.49s)
        --- FAIL: TestNoDataObfuscate/FsMkdir/FsPutFiles (0.39s)
            --- FAIL: TestNoDataObfuscate/FsMkdir/FsPutFiles/Internal (0.00s)
                --- FAIL: TestNoDataObfuscate/FsMkdir/FsPutFiles/Internal/ObjectInfo (0.00s)
                    crypt_internal_test.go:94: 
                                Error Trace:    crypt_internal_test.go:94
                                                                        crypt_internal_test.go:140
                                Error:          Not equal: 
                                                expected: 148
                                                actual  : 100
                                Test:           TestNoDataObfuscate/FsMkdir/FsPutFiles/Internal/ObjectInfo
                --- FAIL: TestNoDataObfuscate/FsMkdir/FsPutFiles/Internal/ObjectInfoWrap (0.00s)
                    crypt_internal_test.go:94: 
                                Error Trace:    crypt_internal_test.go:94
                                                                        crypt_internal_test.go:141
                                Error:          Not equal: 
                                                expected: 148
                                                actual  : 100
                                Test:           TestNoDataObfuscate/FsMkdir/FsPutFiles/Internal/ObjectInfoWrap
```

I' m not very certain if such a failure is expected or something went wrong. Maybe I should open another issue about this?

ncw (approved):
That looks great now - thank you - I will merge now.

ncw:
...actually I will wait for the CI to complete before merging!

ncw:
> There is another thing to note: Currently there is a failing test case in `backend/crypt`, but it is unlikely introduced by my changes because it is also happening on the `master` branch, it looks like this:
[snip]
> I' m not very certain if such a failure is expected or something went wrong. Maybe I should open another issue about this?

That is very interesting - thank you.

What that shows up is that test isn't being run on the CI, nor in the integration tests which is a bit of an oversight, so thank you for bringing it to my attention.

I'll fix the test not being run and the test itself shortly :-)

Max-Sum:
Onedrive used to have a bug on processing special characters. Microsft ignores this issue and therefor I didn't continue on developing this. Sorry for that.
You could test if it works on Onedrive now.

PR #5815, opened 2021-11-15
yandex: add permanent deletion support

deinferno:

This PR add support for optional permament deletion for Yandex backend

This change will be useful for sync that involves a lot of deletions where without this option you need to delete file twice, first by moving it to trash and then emptying it.

What it does:

Adds option --yandex-hard-delete that will delete files permanently instead of moving them to trash

I don't know.

#### Checklist


ncw:
Thank you :-)

PR #5818, opened 2021-11-16
docs: Remove broken dependency from automount unit in `rclone mount` documentation

duxovni:

The purpose of this change is to fix the example systemd automount unit file in the `rclone mount` documentation, so that it no longer gives users a broken configuration.

As noted in [the systemd.automount manpage](https://www.freedesktop.org/software/systemd/man/systemd.automount.html), giving an automount unit for a remote filesystem a dependency on `network-online.target` is unnecessary, since what actually requires the network is the mount itself.  It can also create ordering cycles, for instance on my server:

```
local-fs.target: Found ordering cycle on mnt-rclone.automount/start
local-fs.target: Found dependency on network-online.target/start
local-fs.target: Found dependency on network.target/start
local-fs.target: Found dependency on networking.service/start
local-fs.target: Found dependency on apparmor.service/start
local-fs.target: Found dependency on local-fs.target/start
local-fs.target: Job mnt-rclone.automount/start deleted to break ordering cycle starting with local-fs.target/start
```

Not that I'm aware of.

#### Checklist

### DIFF
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -354,7 +354,6 @@ optionally accompanied by systemd automount unit
 ```
 # /etc/systemd/system/mnt-data.automount
 [Unit]
-After=network-online.target
 Before=remote-fs.target
 [Automount]
 Where=/mnt/data


albertony:
Thank you for your contribution, but identical PR #5911 was recently merged, so will close this PR.

PR #5822, opened 2021-11-16
 new backend: STRATO HiDrive - fixes #1069

### LINKED ISSUE #1069: HiDrive Strato cloud storage support
Hello,

Is there any chance that this provider can be supported as well? It has a well documented api available here: https://dev.strato.com/hidrive/documentation

It's homepage is: https://dev.strato.com/hidrive/documentation
And a free account is available at: https://www.free-hidrive.com/

Grazie mille!

Impelon:

This adds support for the cloud storage technology HiDrive as a new backend called `hidrive`.
It directly uses the [HiDrive REST API](https://developer.hidrive.com/http-api-reference/), offering access to more features and better usability, as opposed to connecting rclone to HiDrive via webdav, etc.
As the API is the same for [HiDrive by STRATO](https://www.strato.de/cloud-speicher/) and [HiDrive by IONOS](https://www.ionos.de/office-loesungen/hidrive-cloud-speicher), this new backend will work for both.

**The code is written by myself with support from my team here at [STRATO AG](https://www.strato.de/).**

Yes, see #1069.

#### Checklist

-------

We are mostly ready to start the review process for this change,
but there are some last quirks we would like to discuss with you before.

I will also add a few notes to the code in its current state, highlighting some design choices that you may disagree with and thus need to change.

Back to the discussion:

<details>
<summary>This has been resolved.</summary>

#### TestIntegration/FsMkdir/FsPutError

When running `test_all -backends hidrive` there are currently two tests that are failing:
- `TestIntegration/FsMkdir/FsPutError`
- `TestCopyFileMaxTransfer`

Both fail for the same reason, that is highlighted by the comment above `TestFsPutError`:
https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/fstest/fstests/fstests.go#L698-L701
This is not a condition that the API can satisfy for files of arbitrary size.

**TL;DR**: The question is how to proceed?
If it suffices that the tests pass for small files, that change can be made without too much hassle.
If the condition should hold true for files of any size, a system resembling the `Chunker`-backend would be a possible solution.

Specifically files up to 2 GiB can be uploaded using a single request; files larger than this need to be uploaded in multiple parts.
There is no transactional mechanism to upload a file in multiple parts.

Currently the `hidrive`-backend always creates an empty file as a first step, even if the file could be uploaded using a single request.
This is done to simplify the code and program flow, but can easily be changed, which would suffice for the tests to pass successfully. 
But the condition `TestFsPutError` is supposed to test, will still not hold for files of any size, just for smaller ones.

----------------

</details>

#### Symbolic Links

_Note: The following section concerns symbolic links that can be natively stored with HiDrive, not the `.rclonelink`-files that rclone uses to encode and store symbolic links. The `hidrive`-backend will handle `.rclonelink`-files as regular files._

HiDrive can natively store symbolic links (symlinks) and has a special type for them, but they can not be up- or downloaded directly using the API.
In general the operations the API can execute on symlinks directly are limited to:
- delete
- move (+ rename)
- copy

This means that symlinks can not be handled by rclone as regular files.

At the moment the `hidrive`-backend simply ignores any native symlinks on `List(...)`, meaning that users can not interact with existing native symlinks in HiDrive-Accounts using rclone. 
As a consequence, deleting using `rclone delete --rmdirs` will *not* be able to remove directories containing symlinks; they can only be deleted via `rclone purge`.

As noted, the `hidrive`-backend is not able to create native symlinks eitherway, but it still may be confusing for users if they can not interact with existing native symlinks.

The preferred solution would be for backends to be able to support native symlinks directly.

Otherwise, a solution that comes to mind is to implement a special `fs.Object`-type in the `hidrive`-backend, that only supports the 3 API-operations mentioned above.

### DIFF
--- a/backend/hidrive/helpers.go
+++ b/backend/hidrive/helpers.go
@@ -1,0 +1,888 @@
+package hidrive
+
+// This file is for helper-functions which may provide more general and
+// specialized functionality than the generic interfaces.
+// There are two sections:
+// 1. methods bound to Fs
+// 2. other functions independent from Fs used throughout the package
+
+// NOTE: Functions accessing paths expect any relative paths
+// to be resolved prior to execution with resolvePath(...).
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"path"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/rclone/rclone/backend/hidrive/api"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/accounting"
+	"github.com/rclone/rclone/fs/fserrors"
+	"github.com/rclone/rclone/lib/ranges"
+	"github.com/rclone/rclone/lib/readers"
+	"github.com/rclone/rclone/lib/rest"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+)
+
+const (
+	// MaximumUploadBytes represents the maximum amount of bytes
+	// a single upload-operation will support.
+	MaximumUploadBytes = 2147483647 // = 2GiB - 1
+	// iterationChunkSize represents the chunk size used to iterate directory contents.
+	iterationChunkSize = 5000
+)
+
+var (
+	// retryErrorCodes is a slice of error codes that we will always retry.
+	retryErrorCodes = []int{
+		429, // Too Many Requests
+		500, // Internal Server Error
+		502, // Bad Gateway
+		503, // Service Unavailable
+		504, // Gateway Timeout
+		509, // Bandwidth Limit Exceeded
+	}
+	// ErrorFileExists is returned when a query tries to create a file
+	// that already exists.
+	ErrorFileExists = errors.New("destination file already exists")
+)
+
+// MemberType represents the possible types of entries a directory can contain.
+type MemberType string
+
+// possible values for MemberType
+const (
+	AllMembers       MemberType = "all"
+	NoMembers        MemberType = "none"
+	DirectoryMembers MemberType = api.HiDriveObjectTypeDirectory
+	FileMembers      MemberType = api.HiDriveObjectTypeFile
+	SymlinkMembers   MemberType = api.HiDriveObjectTypeSymlink
+)
+
+// SortByField represents possible fields to sort entries of a directory by.
+type SortByField string
+
+// possible values for SortByField
+const (
+	descendingSort             string      = "-"
+	SortByName                 SortByField = "name"
+	SortByModTime              SortByField = "mtime"
+	SortByObjectType           SortByField = "type"
+	SortBySize                 SortByField = "size"
+	SortByNameDescending       SortByField = SortByField(descendingSort) + SortByName
+	SortByModTimeDescending    SortByField = SortByField(descendingSort) + SortByModTime
+	SortByObjectTypeDescending SortByField = SortByField(descendingSort) + SortByObjectType
+	SortBySizeDescending       SortByField = SortByField(descendingSort) + SortBySize
+)
+
+var (
+	// Unsorted disables sorting and can therefore not be combined with other values.
+	Unsorted = []SortByField{"none"}
+	// DefaultSorted does not specify how to sort and
+	// therefore implies the default sort order.
+	DefaultSorted = []SortByField{}
+)
+
+// CopyOrMoveOperationType represents the possible types of copy- and move-operations.
+type CopyOrMoveOperationType int
+
+// possible values for CopyOrMoveOperationType
+const (
+	MoveOriginal CopyOrMoveOperationType = iota
+	CopyOriginal
+	CopyOriginalPreserveModTime
+)
+
+// OnExistAction represents possible actions the API should take,
+// when a request tries to create a path that already exists.
+type OnExistAction string
+
+// possible values for OnExistAction
+const (
+	// IgnoreOnExist instructs the API not to execute
+	// the request in case of a conflict, but to return an error.
+	IgnoreOnExist OnExistAction = "ignore"
+	// AutoNameOnExist instructs the API to automatically rename
+	// any conflicting request-objects.
+	AutoNameOnExist OnExistAction = "autoname"
+	// OverwriteOnExist instructs the API to overwrite any conflicting files.
+	// This can only be used, if the request operates on files directly.
+	// (For example when moving/copying a file.)
+	// For most requests this action will simply be ignored.
+	OverwriteOnExist OnExistAction = "overwrite"
+)
+
+// shouldRetry returns a boolean as to whether this resp and err deserve to be retried.
+// It tries to expire/invalidate the token, if necessary.
+// It returns the err as a convenience.
+func (f *Fs) shouldRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	if fserrors.ContextError(ctx, &err) {
+		return false, err
+	}
+	if resp != nil && (resp.StatusCode == 401 || isHTTPError(err, 401)) && len(resp.Header["Www-Authenticate"]) > 0 {
+		fs.Debugf(f, "Token might be invalid: %v", err)
+		if f.tokenRenewer != nil {
+			iErr := f.tokenRenewer.Expire()
+			if iErr == nil {
+				return true, err
+			}
+		}
+	}
+	return fserrors.ShouldRetry(err) || fserrors.ShouldRetryHTTP(resp, retryErrorCodes), err
+}
+
+// resolvePath resolves the given (relative) path and
+// returns a path suitable for API-calls.
+// This will consider the root-path of the fs and any needed prefixes.
+//
+// Any relative paths passed to functions that access these paths should
+// be resolved with this first!
+func (f *Fs) resolvePath(objectPath string) string {
+	resolved := path.Join(f.opt.RootPrefix, f.root, f.opt.Enc.FromStandardPath(objectPath))
+	return resolved
+}
+
+// iterateOverDirectory calls the given function callback
+// on each item found in a given directory.
+//
+// If callback ever returns true then this exits early with found = true.
+func (f *Fs) iterateOverDirectory(ctx context.Context, directory string, searchOnly MemberType, callback func(*api.HiDriveObject) bool, fields []string, sortBy []SortByField) (found bool, err error) {
+	parameters := api.NewQueryParameters()
+	parameters.SetPath(directory)
+	parameters.AddFields("members.", fields...)
+	parameters.AddFields("", api.DirectoryContentFields...)
+	parameters.Set("members", string(searchOnly))
+	for _, v := range sortBy {
+		// The explicit conversion is necessary for each element.
+		parameters.AddList("sort", ",", string(v))
+	}
+
+	opts := rest.Opts{
+		Method:     "GET",
+		Path:       "/dir",
+		Parameters: parameters.Values,
+	}
+
+	iterateContent := func(result *api.DirectoryContent, err error) (bool, error) {
+		if err != nil {
+			return false, err
+		}
+		for _, item := range result.Entries {
+			item.Name = f.opt.Enc.ToStandardName(item.Name)
+			if callback(&item) {
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+	return f.paginateDirectoryAccess(ctx, &opts, iterationChunkSize, 0, iterateContent)
+}
+
+// paginateDirectoryAccess executes requests specified via ctx and opts
+// which should produce api.DirectoryContent.
+// This will paginate the requests using limit starting at the given offset.
+//
+// The given function callback is called on each api.DirectoryContent found
+// along with any errors that occurred.
+// If callback ever returns true then this exits early with found = true.
+// If callback ever returns an error then this exits early with that error.
+func (f *Fs) paginateDirectoryAccess(ctx context.Context, opts *rest.Opts, limit int64, offset int64, callback func(*api.DirectoryContent, error) (bool, error)) (found bool, err error) {
+	for {
+		opts.Parameters.Set("limit", strconv.FormatInt(offset, 10)+","+strconv.FormatInt(limit, 10))
+
+		var result api.DirectoryContent
+		var resp *http.Response
+		err = f.pacer.Call(func() (bool, error) {
+			resp, err = f.srv.CallJSON(ctx, opts, nil, &result)
+			return f.shouldRetry(ctx, resp, err)
+		})
+
+		found, err = callback(&result, err)
+		if found || err != nil {
+			return found, err
+		}
+
+		offset += int64(len(result.Entries))
+		if offset >= result.TotalCount || limit > int64(len(result.Entries)) {
+			break
+		}
+	}
+	return false, nil
+}
+
+// fetchMetadataForPath reads the metadata from the path.
+func (f *Fs) fetchMetadataForPath(ctx context.Context, path string, fields []string) (*api.HiDriveObject, error) {
+	parameters := api.NewQueryParameters()
+	parameters.SetPath(path)
+	parameters.AddFields("", fields...)
+
+	opts := rest.Opts{
+		Method:     "GET",
+		Path:       "/meta",
+		Parameters: parameters.Values,
+	}
+
+	var result api.HiDriveObject
+	var resp *http.Response
+	var err error
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.CallJSON(ctx, &opts, nil, &result)
+		return f.shouldRetry(ctx, resp, err)
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
+// copyOrMove copies or moves a directory or file
+// from the source-path to the destination-path.
+//
+// The operation will only be successful
+// if the parent-directory of the destination-path exists.
+//
+// NOTE: Use the explicit methods instead of directly invoking this method.
+// (Those are: copyDirectory, moveDirectory, copyFile, moveFile.)
+func (f *Fs) copyOrMove(ctx context.Context, isDirectory bool, operationType CopyOrMoveOperationType, source string, destination string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	parameters := api.NewQueryParameters()
+	parameters.Set("src", source)
+	parameters.Set("dst", destination)
+	if onExist == AutoNameOnExist ||
+		(onExist == OverwriteOnExist && !isDirectory) {
+		parameters.Set("on_exist", string(onExist))
+	}
+
+	endpoint := "/"
+	if isDirectory {
+		endpoint += "dir"
+	} else {
+		endpoint += "file"
+	}
+	switch operationType {
+	case MoveOriginal:
+		endpoint += "/move"
+	case CopyOriginalPreserveModTime:
+		parameters.Set("preserve_mtime", strconv.FormatBool(true))
+		fallthrough
+	case CopyOriginal:
+		endpoint += "/copy"
+	}
+
+	opts := rest.Opts{
+		Method:     "POST",
+		Path:       endpoint,
+		Parameters: parameters.Values,
+	}
+
+	var result api.HiDriveObject
+	var resp *http.Response
+	var err error
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.CallJSON(ctx, &opts, nil, &result)
+		return f.shouldRetry(ctx, resp, err)
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
+// copyDirectory moves the directory at the source-path to the destination-path and
+// returns the resulting api-object if successful.
+//
+// The operation will only be successful
+// if the parent-directory of the destination-path exists.
+func (f *Fs) copyDirectory(ctx context.Context, source string, destination string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	return f.copyOrMove(ctx, true, CopyOriginalPreserveModTime, source, destination, onExist)
+}
+
+// moveDirectory moves the directory at the source-path to the destination-path and
+// returns the resulting api-object if successful.
+//
+// The operation will only be successful
+// if the parent-directory of the destination-path exists.
+func (f *Fs) moveDirectory(ctx context.Context, source string, destination string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	return f.copyOrMove(ctx, true, MoveOriginal, source, destination, onExist)
+}
+
+// copyFile copies the file at the source-path to the destination-path and
+// returns the resulting api-object if successful.
+//
+// The operation will only be successful
+// if the parent-directory of the destination-path exists.
+//
+// NOTE: This operation will expand sparse areas in the content of the source-file
+// to blocks of 0-bytes in the destination-file.
+func (f *Fs) copyFile(ctx context.Context, source string, destination string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	return f.copyOrMove(ctx, false, CopyOriginalPreserveModTime, source, destination, onExist)
+}
+
+// moveFile moves the file at the source-path to the destination-path and
+// returns the resulting api-object if successful.
+//
+// The operation will only be successful
+// if the parent-directory of the destination-path exists.
+//
+// NOTE: This operation may expand sparse areas in the content of the source-file
+// to blocks of 0-bytes in the destination-file.
+func (f *Fs) moveFile(ctx context.Context, source string, destination string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	return f.copyOrMove(ctx, false, MoveOriginal, source, destination, onExist)
+}
+
+// createDirectory creates the directory at the given path and
+// returns the resulting api-object if successful.
+//
+// The directory will only be created if its parent-directory exists.
+// This returns fs.ErrorDirNotFound if the parent-directory is not found.
+// This returns fs.ErrorDirExists if the directory already exists.
+func (f *Fs) createDirectory(ctx context.Context, directory string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	parameters := api.NewQueryParameters()
+	parameters.SetPath(directory)
+	if onExist == AutoNameOnExist {
+		parameters.Set("on_exist", string(onExist))
+	}
+
+	opts := rest.Opts{
+		Method:     "POST",
+		Path:       "/dir",
+		Parameters: parameters.Values,
+	}
+
+	var result api.HiDriveObject
+	var resp *http.Response
+	var err error
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.CallJSON(ctx, &opts, nil, &result)
+		return f.shouldRetry(ctx, resp, err)
+	})
+
+	switch {
+	case err == nil:
+		return &result, nil
+	case isHTTPError(err, 404):
+		return nil, fs.ErrorDirNotFound
+	case isHTTPError(err, 409):
+		return nil, fs.ErrorDirExists
+	}
+	return nil, err
+}
+
+// createDirectories creates the directory at the given path
+// along with any missing parent directories and
+// returns the resulting api-object (of the created directory) if successful.
+//
+// This returns fs.ErrorDirExists if the directory already exists.
+//
+// If an error occurs while the parent directories are being created,
+// any directories already created will NOT be deleted again.
+func (f *Fs) createDirectories(ctx context.Context, directory string, onExist OnExistAction) (*api.HiDriveObject, error) {
+	result, err := f.createDirectory(ctx, directory, onExist)
+	if err == nil {
+		return result, nil
+	}
+	if err != fs.ErrorDirNotFound {
+		return nil, err
+	}
+	parentDirectory := path.Dir(directory)
+	_, err = f.createDirectories(ctx, parentDirectory, onExist)
+	if err != nil && err != fs.ErrorDirExists {
+		return nil, err
+	}
+	// NOTE: Ignoring fs.ErrorDirExists does no harm,
+	// since it does not mean the child directory cannot be created.
+	return f.createDirectory(ctx, directory, onExist)
+}
+
+// deleteDirectory deletes the directory at the given path.
+//
+// If recursive is false, the directory will only be deleted if it is empty.
+// If recursive is true, the directory will be deleted regardless of its content.
+// This returns fs.ErrorDirNotFound if the directory is not found.
+// This returns fs.ErrorDirectoryNotEmpty if the directory is not empty and
+// recursive is false.
+func (f *Fs) deleteDirectory(ctx context.Context, directory string, recursive bool) error {
+	parameters := api.NewQueryParameters()
+	parameters.SetPath(directory)
+	parameters.Set("recursive", strconv.FormatBool(recursive))
+
+	opts := rest.Opts{
+		Method:     "DELETE",
+		Path:       "/dir",
+		Parameters: parameters.Values,
+		NoResponse: true,
+	}
+
+	var resp *http.Response
+	var err error
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.Call(ctx, &opts)
+		return f.shouldRetry(ctx, resp, err)
+	})
+
+	switch {
+	case isHTTPError(err, 404):
+		return fs.ErrorDirNotFound
+	case isHTTPError(err, 409):
+		return fs.ErrorDirectoryNotEmpty
+	}
+	return err
+}
+
+// deleteObject deletes the object/file at the given path.
+//
+// This returns fs.ErrorObjectNotFound if the object is not found.
+func (f *Fs) deleteObject(ctx context.Context, path string) error {
+	parameters := api.NewQueryParameters()
+	parameters.SetPath(path)
+
+	opts := rest.Opts{
+		Method:     "DELETE",
+		Path:       "/file",
+		Parameters: parameters.Values,
+		NoResponse: true,
+	}
+
+	var resp *http.Response
+	var err error
+	err = f.pacer.Call(func() (bool, error) {
+		resp, err = f.srv.Call(ctx, &opts)
+		return f.shouldRetry(ctx, resp, err)
+	})
+
+	if isHTTPError(err, 404) {
+		return fs.ErrorObjectNotFound
+	}
+	return err
+}
+
+// createFile creates a file at the given path
+// with the content of the io.ReadSeeker.
+// This guarantees that existing files will not be overwritten.
+// The maximum size of the content is limited by MaximumUploadBytes.
+// The io.ReadSeeker should be resettable by seeking to its start.
+// If modTime is not the zero time instant,
+// it will be set as the file's modification time after the operation.
+//
+// This returns fs.ErrorDirNotFound
+// if the parent directory of the file is not found.
+// This returns ErrorFileExists if a file already exists at the specified path.
+func (f *Fs) createFile(ctx context.Context, path string, content io.ReadSeeker, modTime time.Time, onExist OnExistAction) (*api.HiDriveObject, error) {
+	parameters := api.NewQueryParameters()
+	parameters.SetFileInDirectory(path)
+	if onExist == AutoNameOnExist {
+		parameters.Set("on_exist", string(onExist))
+	}
+
+	var err error
+	if !modTime.IsZero() {
+		err = parameters.SetTime("mtime", modTime)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	opts := rest.Opts{
+		Method:      "POST",
+		Path:        "/file",
+		Body:        content,
+		ContentType: "application/octet-stream",
+		Parameters:  parameters.Values,
+	}
+
+	var result api.HiDriveObject
+	var resp *http.Response
+	err = f.pacer.Call(func() (bool, error) {
+		// Reset the reading index (in case this is a retry).
+		if _, err = content.Seek(0, io.SeekStart); err != nil {
+			return false, err
+		}
+		resp, err = f.srv.CallJSON(ctx, &opts, nil, &result)
+		return f.shouldRetry(ctx, resp, err)
+	})
+
+	switch {
+	case err == nil:
+		return &result, nil
+	case isHTTPError(err, 404):
+		return nil, fs.ErrorDirNotFound
+	case isHTTPError(err, 409):
+		return nil, ErrorFileExists
+	}
+	return nil, err
+}
+
+// overwriteFile updates the content of the file at the given path
+// with the content of the io.ReadSeeker.
+// If the file does not exist it will be created.
+// The maximum size of the content is limited by MaximumUploadBytes.
+// The io.ReadSeeker should be resettable by seeking to its start.
+// If modTime is not the zero time instant,
+// it will be set as the file's modification time after the operation.
+//
+// This returns fs.ErrorDirNotFound
+// if the parent directory of the file is not found.
+func (f *Fs) overwriteFile(ctx context.Context, path string, content io.ReadSeeker, modTime time.Time) (*api.HiDriveObject, error) {
+	parameters := api.NewQueryParameters()
+	parameters.SetFileInDirectory(path)
+
+	var err error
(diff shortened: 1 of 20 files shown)


lib/oauthutil/oauthutil.go
```diff
@@ -269,6 +269,23 @@ func (ts *TokenSource) Invalidate() {
 	ts.mu.Unlock()
 }
 
+// Expire marks the token as expired
+//
+// This also marks the token in the config file as expired, if it is the same one
+func (ts *TokenSource) Expire() error {
```
Impelon, line 275: Currently there is no way for backends to detect invalid access tokens and request a new one by invalidating the old access token (but keeping the refresh token).

Of course this change is not strictly needed, as invalid access tokens can not be generated by valid refresh tokens, thus such a situation can not be caused by rclone itself, at least regarding HiDrive.
This is more or less a safeguard against accidental changes to the access-token when editing the configuration.

Impelon: This may be useful in situations where an access token can be revoked early.

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,962 @@
...
+	fs          *Fs       // what this object is part of
+	remote      string    // The remote path
+	hasMetadata bool      // whether info below has been set
+	size        int64     // size of the object
+	modTime     time.Time // modification time of the object
+	id          string    // ID of the object
+	hash        string    // content-hash of the object
+}
+
+// ZeroDelayCalculator is a pacer.Calculator that never delays.
```
Impelon, line 249: The `hidrive`-backend uses the `fs.pacer`-mechanism to retry some operations once.
(That is the `retrier`-pacer of the `Fs`-implementation.)

For retrying those operations no delays are needed; the same could be accomplished using a for-Loop or some duplicate code.
As no delay is needed, the `ZeroDelayCalculator` is introduced for this pacer.

If considered useful in general, the `ZeroDelayCalculator` could also be moved to `lib/pacer/parcers.go`.

ivandeex:
> TestIntegration/FsMkdir/FsPutError
> TestCopyFileMaxTransfer
> Both fail for the same reason, that is highlighted by the comment
> // It makes sure that aborting a file half way through does not create a file on the remote. 
> This is not a condition that the API can satisfy for files of arbitrary size.

Try returning both error and **broken half-object** from `fs.Put`. This step is warranted by the comment
https://github.com/rclone/rclone/blob/169990e270b2977c39bd6ecd8a2921cf30a6d2b7/fstest/fstests/fstests.go#L719-L721

If the test still fails, make `fs.Put` remove the stray file. We have other providers not removing half baked objects, so their backend code fills the API gap to ensure behavior standardized by test suite. Copy code from examples:

1. commit https://github.com/rclone/rclone/commit/badefdb060825a592c5be2208b2cbefbd4808a1e

2. https://github.com/rclone/rclone/blob/169990e270b2977c39bd6ecd8a2921cf30a6d2b7/backend/sia/sia.go#L338-L352

lib/oauthutil/oauthutil.go
ivandeex: We don't know details of your API so we defer to your authority.
rclone has been widely used in automated workflows so the more stable is the code, the better.

backend/hidrive/hidrive.go
ivandeex: @ncw Do you want `ZeroDelayCalculator` in `lib/pacer` ?

ivandeex:
Our CI requires lint-clear source code.
The best way to fix linter errors is to install https://golangci-lint.run/usage/install/#linux-and-windows,
run
```
cd rclone-source
golangci-lint run ./backend/hidrive/...
```

and change whatever requested by linter.

Impelon:
Thank you for the quick and helpful reply!

`TestFsPutError` fails at line 724-725, when it tries to verify that no Object has been created, I should have been more clear about that.
https://github.com/rclone/rclone/blob/169990e270b2977c39bd6ecd8a2921cf30a6d2b7/fstest/fstests/fstests.go#L723-L725

> I suggest you still add the workaround in the `fs.Put` code and test it manually.

Will do so! Thank you for the pointers towards existing workarounds.

> Our CI requires lint-clear source code.

I have now integrated `golangci-lint` in my setup.
The next time I push an update to the `hidrive`-backend, I will have made changes according to the linter-suggestions :+1:

Impelon:
Documentation has been added, linter-suggestions implemented and some workarounds have been added to avoid files being created if the upload fails.

In other words: This PR ready for review!

stkr22:
Any chance someone can review and approve this? Using this provider a lot and would really appreciate the added backend to rclone. :+1:

ncw:
I will review this after 1.58 is released. Sorry a bit busy getting the release ready at the moment and I don't think we've time to squeeze this in.

stkr22:
Ah too bad, but really appreciate the quick feedback and the effort @ncw . 👍

ncw:
I have some time to review and merge this now - sorry for the delay.

As a first step can you rebase it on master and fix the conflict - it shouldn't be hard.


> HiDrive can natively store symbolic links (symlinks) and has a special type for them, but they can not be up- or downloaded directly using the API. In general the operations the API can execute on symlinks directly are limited to:
> 
> * delete
> * move (+ rename)
> * copy
> 
> This means that symlinks can not be handled by rclone as regular files.
> 
> At the moment the `hidrive`-backend simply ignores any native symlinks on `List(...)`, meaning that users can not interact with existing native symlinks in HiDrive-Accounts using rclone. As a consequence, deleting using `rclone delete --rmdirs` will _not_ be able to remove directories containing symlinks; they can only be deleted via `rclone purge`.

This is not too far from other backends. For example some backends you can't remove directories unless you remove all the deleted files from them which don't show in a directory listing and you'd need to use `rclone purge` for that.

> As noted, the `hidrive`-backend is not able to create native symlinks eitherway, but it still may be confusing for users if they can not interact with existing native symlinks.
> 
> The preferred solution would be for backends to be able to support native symlinks directly.

Can you describe the semantics of the symlinks? Are they exactly the same as unix symlinks? Those little differences in semantics have made it hard to think of a general purpose way of representing symlinks.

> Otherwise, a solution that comes to mind is to implement a special `fs.Object`-type in the `hidrive`-backend, that only supports the 3 API-operations mentioned above.

Or you could use a flag in the existing Object. That doesn't sound too bad as a solution.

You could also implement some backend specific commands to make the links if you wanted (see https://rclone.org/drive/#shortcut for example).

...

How well does the backend pass the integration tests mentioned in this section of the docs https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend ?

I need all the integration tests to pass if at all possible!

I'm also going to need an account to run the daily integration tests to make sure the backend stays up to date. Are free accounts available?

Impelon:
> I have some time to review and merge this now - sorry for the delay.
No problem! :) Thanks for the suggestions regarding symlinks.

Unfortunately I'm now on vacation myself and will not be able to make code changes until after the 15th of May.
Afterwards I'm happy to rebase the code.

> How well does the backend pass the integration tests mentioned in this section of the docs https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend ?

It should pass all integrations tests; at least it did so in December.
Although there is currently a known inconsistency with the HiDrive Platform, which causes `TestFsPutChunked` to fail non-deterministically for certain chunk-sizes.
But we decided that this implementation is right in its behavior, and instead the HiDrive API should be able to handle these cases gracefully.
As such I've avoided to specify `CeilChunkSize` in `backend/hidrive/hidrive_test.go`.

Hopefully we can solve this inconsistency of the API before merging the pull-request;
We already had a solution queued up, so I'll ask my colleges to move it back up on the priority list.

> I'm also going to need an account to run the daily integration tests to make sure the backend stays up to date. Are free accounts available?

Yes, I talked to my colleges back in December to organize a test account for this purpose.
I'll arrange for the credentials to be send to you via mail, so you can test things out if you like to. [nick@craig-wood.com](mailto:nick@craig-wood.com) is correct, right?

As a heads-up: The root-directory of a HiDrive-Remote is not writable, so the automatic integration tests will fail there.
My preferred solution for this was to specify the `root_prefix` option in the config for the test remote `TestHiDrive:` to a writeable location, e.g. the user-directory `/users/<user-name>`.
As this path naturally is different for every user, it cannot simply be hardcoded in `fstest/test_all/config.yaml`.

Impelon:
> Can you describe the semantics of the symlinks? Are they exactly the same as unix symlinks?

Yes, these are basically unix symlinks. If the user includes a unix symlink in a zip-archive for example, and instructs the HiDrive API to unpack that archive on the remote side (f.e. via the Web-Interface), the symlink will be created on that remote.
Though, these symlinks are not really interactive via the Web-Interface. One cannot directly create or download them in the Web-Interface, only by using archives.
But it allows users to sync their local symlinks to their HiDrive storage and recover them anytime.

ncw:
> Unfortunately I'm now on vacation myself and will not be able to make code changes until after the 15th of May. Afterwards I'm happy to rebase the code.

No rush!

> > How well does the backend pass the integration tests mentioned in this section of the docs https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend ?
> 
> It should pass all integrations tests; at least it did so in December. Although there is currently a known inconsistency with the HiDrive Platform, which causes `TestFsPutChunked` to fail non-deterministically for certain chunk-sizes. But we decided that this implementation is right in its behavior, and instead the HiDrive API should be able to handle these cases gracefully. As such I've avoided to specify `CeilChunkSize` in `backend/hidrive/hidrive_test.go`.

That all sounds great :-)

> Hopefully we can solve this inconsistency of the API before merging the pull-request; We already had a solution queued up, so I'll ask my colleges to move it back up on the priority list.

If integration tests fail sporadically then they get retried, so this isn't a barrier to merging.

> > I'm also going to need an account to run the daily integration tests to make sure the backend stays up to date. Are free accounts available?
> 
> Yes, I talked to my colleges back in December to organize a test account for this purpose. I'll arrange for the credentials to be send to you via mail, so you can test things out if you'd like to. [nick@craig-wood.com](mailto:nick@craig-wood.com) is correct, right?

Perfect, thank you.

> As a heads-up: The root-directory of a HiDrive-Remote is not writable, so the automatic integration tests will fail there. My preferred solution for this was to specify the `root_prefix` option in the config for the test remote `TestHiDrive:` to a writeable location, e.g. the user-directory `/users/<user-name>`. As this path naturally is different for every user, it cannot simply be hardcoded in `fstest/test_all/config.yaml`.

Gotcha.
> > Can you describe the semantics of the symlinks? Are they exactly the same as unix symlinks?
> 
> Yes, these are basically unix symlinks. If the user includes a unix symlink in a zip-archive for example, and instructs the HiDrive API to unpack that archive on the remote side (f.e. via the HiDrive web-interface), the symlink will be created on that remote. Though, these symlinks are not really interactive via the web-interface. One cannot directly create or download them in the web-interface, only by using archives or rsync. But it allows users to sync their local symlinks to their HiDrive storage and recover them anytime.

OK. I'll have a think about generic symlink support, but in the mean time you could have the backend follow the symlink itself (this is what the `drive` backend does) and present it to the user like that?

Impelon:
I've now had the time to come back to this and rebased the code to the current master.
From a quick glance through the new commits, the implementation should be mostly unaffected, if I'm not mistaken.
I did update the documentation with the automatic `make backenddocs`, but I may have missed some of the manual style changes @albertony authored in the last months.

> I'll arrange for the credentials to be send to you via mail

Hopefully you have received them by now, if not, let me know!

> > How well does the backend pass the integration tests [...]?
> It should pass all integrations tests; at least it did so in December. 

Checked it now; they still do!

> OK. I'll have a think about generic symlink support, but in the mean time you could have the backend follow the symlink itself (this is what the `drive` backend does) and present it to the user like that?

Reading the link itself (and then following it) would require an addition to HiDrive's current API. At the moment, the API can only delete, copy, or move symlinks, and detect them as such. It cannot interact with them in any other way (like figuring out the location the link points to). 
We've discussed this in our team and we think it would be easiest for now, if we do not extend the API's behaviour.
Especially since native symlinks are rather an edge-case in a HiDrive.

What *could* be implemented without an extension of the API, is the movement, deletion etc. of symlinks, I was not sure initially how to best map this to the 'rclone world'. However you mentioned:
> [...] [S]ome backends [...] can't remove directories unless you remove all the deleted files from them which don't show in a directory listing and you'd need to use `rclone purge` for that.

If you are of the opinion that using `rclone purge` is not that big of a deal, I think it would be easiest to leave the implementation as is and postpone this extension to when and if there exists a demand for it. :) 
Symlinks are pretty niche and the implementation of this will require some more substantial changes in the hidrive-Backend for rclone.

Impelon:
Accidentally force-pushed the current master into Impelon:hidrive and git closed this automatically. Not what I intented.

ncw:
@Impelon thank you for the rebase. I've had a brief look through the code - it looks really good :-) I'm going to try the integration tests now.

With any luck we can merge this for v1.59 and if there are any little fixups they can go in a point release. I'll report back in a bit...

ncw:
Integration tests 100% pass :star: well done :-) They were very quick too, you have a speedy backend.

In the setup I noticed this

> Option disable_fetching_member_count.
>
> Do not fetch number of objects in directories unless it is absolutely necessary.
> Requests may be faster if the number of objects in subdirectories is not fetched.
> Enter a boolean value (true or false). Press Enter for the default (false).

The count of numbers of files is returned in the directory object, however this is only really used in one place in rclone, when you do an `rclone lsd`. It also appears in `rclone lsjson` I think. So users might use this, but if it slows things down a lot then setting the default to not fetching the member count would be fine.

I'll leave the choice to you.

I'm going to go through the code now.

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+// See note about sparse files in patchFile.
+// If any of the uploads fail, the process will be aborted and
+// the first error that occurred will be returned.
+// This is not an atomic operation,
+// therefore if the upload fails the file may be partially modified.
+//
+// This returns fs.ErrorObjectNotFound if the object is not found.
+func (f *Fs) updateFileChunked(ctx context.Context, path string, content io.Reader, offset uint64, chunkSize int, transferLimit int64) (amount uint64, err error) {
+	amount = uint64(0)
+	mutex := new(sync.Mutex)
```
ncw, line 632: Can you document exactly what the mutex is protecting here?

In the rclone code we usually write this (no need for `new(sync.Mutex)` these can live on the stack well enough)

```go
var (
    amountMu sync.Mutex // protects the variables below
    amount = uint64(0)
)
```

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+		waitgroup.Go(func() error {
+			// After this upload is done,
+			// signal that another transfer can be started.
+			defer transferSemaphore.Release(1)
+			uploadErr := f.patchFile(wgCtx, path, cachedReader(chunkReader), chunkOffset, zeroTime)
+			// Safely update lower-bound
+			// for amount of bytes written before failures.
+			// The errgroup-context wgCtx is canceled implicitly
+			// as soon as the first routine returns an error.
+			if uploadErr == nil {
```
ncw, line 667: This if statement looks a bit untidy...

There is duplicated code and the mutex is being held for longer than necessary.

I dont think the wgCtx checks are needed here

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+				mutex.Lock()
+				select {
+				case <-wgCtx.Done():
+					// Another upload failed already.
+					// Nothing more to do.
+				default:
+					contentPosition := contentOffset + uint64(bytesRead)
+					fs.Debugf(f, "Done uploading chunk of size %v at offset %v.", bytesRead, chunkOffset)
+					if amount < contentPosition {
```
ncw, line 676: I'm not sure this calculation on `amount` is correct if it is intended to be the last written block position.

What if block 1 was OK, block 2 failed and block 3 was OK. This could easily complete in the order block 1 OK, block 3 OK, block 2 FAIL. Won't amount be set wrong in this case? As in set to end of block 3 where there is a gap for block 2?

Or is it just the high water mark for correctly written blocks?

Or is it the high water mark for any blocks written, correctly or not.

Either way I think you can tidy this if statement up by dealing with `amount` first if uploadErr == nil.

You might want to check out the s3 multipart uploader for ideas - this code has been through many iterations to get the concurrency right!

https://github.com/rclone/rclone/blob/5c6a958ad88db85471fc8d1975a5378f67e16fc1/backend/s3/s3.go#L4085

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+
+	if err == nil {
+		info, err = f.resizeFile(ctx, path, amount, modTime)
+	}
+	return amount, info, err
+}
+
+// updateFileChunked updates the content of the existing file at the given path
+// starting at the given offset.
+// Returns the amount of successfully written bytes.
```
ncw, line 606: I wasn't quite sure what `amount` is

- Is this just a counter of number of successfully written bytes? If so then using `sync/atomic.AddUint64` will simplify this code a lot.
- If it is supposed to be the last successfully written block then I think it is probably being calculated incorrectly - see below.
- Or is is the end position of the last success write?

I suspect from looking at the code it is supposed to be the last, since its only use seems to be passed into truncate. Maybe it would be best renamed `lastPosition` or something like that if so?

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,1021 @@
...
```
ncw, line 249: For backends which don't care how hard you hit them (usually ones where people are paying per transaction) I'd recommend using the S3 pacer which has no delays, unless there are errors in which case backing off is a good idea.

https://github.com/rclone/rclone/blob/5c6a958ad88db85471fc8d1975a5378f67e16fc1/lib/pacer/pacers.go#L270-L277

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,1021 @@
...
+		return nil, fmt.Errorf("failed to configure HiDrive: %w", err)
+	}
+
+	f := &Fs{
+		name:    name,
+		root:    root,
+		opt:     *opt,
+		srv:     rest.NewClient(client).SetRoot(opt.EndpointAPI),
+		pacer:   fs.NewPacer(ctx, pacer.NewDefault(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
+		retrier: pacer.New(pacer.RetriesOption(2), pacer.MaxConnectionsOption(-1), pacer.CalculatorOption(&ZeroDelayCalculator{})),
```
ncw, line 335: Number of retries is traditionally controlled by `--low-level-retries` unless we have a strong reason not to, so you don't need to override this.

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,1021 @@
...
+	}
+	return f.newObject(remote, metaFiller)
+}
+
+// NewObject finds the Object at remote.
+//
+// If remote points to a directory then it returns fs.ErrorIsDir.
+// If it can not be found it returns the error fs.ErrorObjectNotFound.
+func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
+	fs.Debugf(f, "executing NewObject(%s).", remote)
```
ncw, line 423: For release we'll probably want to comment out these sort of debugs as they get too noisy. I usually leave them in the code but commented out.

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,1021 @@
...
+		return nil, err
+	}
+
+	o, err := f.newObjectFromHiDriveObject(remote, info)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if bytesRead < int(f.opt.UploadCutoff) {
```
ncw, line 575: Integer overflow? uploadcutoff could be bigger than 2GB which on a 32bit machine will cause integer overflow.

backend/hidrive/hidrive.go
```diff
@@ -0,0 +1,1021 @@
...
+		if seekoption, ok := option.(*fs.SeekOption); ok {
+			offset = uint64(seekoption.Offset)
+			break
+		}
+	}
+
+	var info *api.HiDriveObject
+	var err, metaErr error
+	if offset > 0 || src.Size() == -1 || src.Size() >= int64(o.fs.opt.UploadCutoff) {
+		fs.Infof(o.fs, "Uploading with chunks of size %v and %v transfers in parallel at path '%s'.", int(o.fs.opt.UploadChunkSize), o.fs.opt.UploadConcurrency, resolvedPath)
```
ncw, line 968: This should be a debug - Info logs are for rclone to announce transfers completed.

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+	var readErr error
+	startMoreTransfers := true
+	zeroTime := time.Time{}
+	for chunk := uint64(0); startMoreTransfers; chunk++ {
+		// Acquire semaphore to limit number of transfers in parallel.
+		readErr = transferSemaphore.Acquire(wgCtx, 1)
+		if readErr != nil {
+			break
+		}
+
```
ncw, line 645: The check for `<-wgCtx.Done()` should probably be done before the transfer starts otherwise this will upload transferLimit blocks it didn't need to. Its probably best written as

```go
                // Fail fast, in case an errgroup managed function returns an error
		// gCtx is cancelled. There is no point in uploading all the other parts.
		if gCtx.Err() != nil {
			free()
			break
		}
```

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+// If any of the uploads fail, the process will be aborted and
+// the first error that occurred will be returned.
+// This is not an atomic operation,
+// therefore if the upload fails the file may be partially modified.
+//
+// This returns fs.ErrorObjectNotFound if the object is not found.
+func (f *Fs) updateFileChunked(ctx context.Context, path string, content io.Reader, offset uint64, chunkSize int, transferLimit int64) (amount uint64, err error) {
+	amount = uint64(0)
+	mutex := new(sync.Mutex)
+	waitgroup, wgCtx := errgroup.WithContext(ctx)
```
ncw, line 633: It will make my life easier if you call these variables `g, gCtx` as that is what we use elsewhere in rclone!

ncw (commented):
That is really nice code - well done :-)

The only bit I had an argument with was the multipart upload code which I think needs another revisit.

If you could comment out the tracing debug calls - that is what we usually do for a release. Leave the non tracing debugs in - that is fine.

There is a few other little things inline.

Great work - I don't see a problem merging this for v1.59 given a few little tweaks!

backend/hidrive/hidrive.go
Impelon: This `retrier` is how I "abuse" rclone's pacer mechanism for something it was not originally designed for:
To repeat a fragment of code at most once, in case a necessary precondition is not fulfilled and the code needs to be repeated after that condition is fulfilled.
The `retrier` should never actually be used to repeat API-calls; that's what the above `pacer` object is for.

This is also what I briefly mentioned in [the discussion a while back](https://github.com/rclone/rclone/pull/5822#discussion_r750397613). As I mention there, the same effect could be accomplished through a for-loop or duplicated code.

I'll try to illustrate the motivation behind this using the `Copy` implementation:
We could first check if the parent-folders of the destination already exists, and create them accordingly before issuing the copy operation.
However we can expect that the parent-folders already exists most of the time, so this would be a bit inefficient.
Instead the design is as follows:
1. Try to copy the file.
2. If it fails because the parents are missing, create the parents and repeat step 1 *once*.

<details>
<summary>This could be implemented in the following ways (simplified):</summary>

1. duplicated code:
```golang
err = copyFile(...)
if err {
    if sourceIsMissing(...) {
      // fail. Abort copy-operation.
    }
    err = createParents(...)
    if err == nil {
        err = copyFile(...)
    }
}
```
2. for loop:
```golang
for attempt := 0; attempt < 2; attempt++ {
    err = copyFile(...)
    if err == nil {
        break
    } else {
        if sourceIsMissing(...) {
          // fail. Abort copy-operation.
        }
        if attempt == 0 {
            createParents(...)
            if err == nil {
                continue
            }
        }
    } 
}
```
3. `retrier`:
```golang
f.retrier.Call(func() (bool, error) {
    err = copyFile(...)
    if err {
        if sourceIsMissing(...) {
            return false, err
        }
        err = createParents(...)
        if err {
            return false, err
        }
        return true, err
    }
    return false, err
}
```
</details>

I decided to use the `retrier`-mechanism as it seemed more readable than the for loop I used before. Though, looking back at it now, it may have just been easiest to use duplicated code...

If you dislike this "abuse" of the pacer-mechanism I'll rewrite the 4 affected methods to use duplicated code instead.
Otherwise, maybe `retrier` needs a better name like `retryOnce` and a longer comment-block explaining the idea behind it?

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+)
+
+// shouldRetry returns a boolean as to whether this resp and err deserve to be retried.
+// It tries to expire/invalidate the token, if necessary.
+// It returns the err as a convenience.
+func (f *Fs) shouldRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	if fserrors.ContextError(ctx, &err) {
+		return false, err
+	}
+	if resp != nil && (resp.StatusCode == 401 || isHTTPError(err, 401)) && len(resp.Header["Www-Authenticate"]) > 0 {
```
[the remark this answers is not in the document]
Impelon: Hm, I'm not exactly sure what `this` is referring to. **TL;DR**: `TokenSource.Expire` was introduced to handle corruption of the access token in the config, or the revocation of tokens.

I'll explain how this code sprang into existence:
To test whether the backend fetches a fresh access-token automatically, I decided to manually insert an invalid token into the configuration by deleting one of the characters.
Unfortunately, this led to the token never being updated, because the `expiry` time of the token in the config still indicated it was valid. The error was not particularly helpful either:
```
2022/07/05 20:33:54 Failed to create file system for "TestHiDrive:": could not access root-prefix: Error "401": Unauthorized
```
I did not like this behavior, because it meant that if a user accidentally corrupted their access token, they would not be able to use rclone as long as the token is supposedly valid, and the error message is too unspecific to understand what went wrong.

So I thought I could use `TokenSource.Invalidate` to manually tell rclone to fetch a new token, if the backend decides the current one is definitely invalid. But this does not seem to be the use-case for `Invalidate`: It just sets the token to an empty string and does not update the configuration, so when issuing the next retry, `TokenSource.Token` simply reads the invalid token from the config again.
This is `TokenSource.Expire` was introduced, so backends could decide via custom logic when a token is definitely invalid and needs to be refreshed. We decided not to change the existing behavior of `oauthutil` and instead introduced this as a new method.

Besides tolerating corruption of the token in the config, this is also useful in situations where an access token can be revoked by the OAuth-sever, as this would lead to the same situation: rclone believes the token from the config to be valid, because the `expiry`-time was not reached yet, but the server denies the request.
With HiDrive specifically, access tokens are only revoked when their corresponding refresh token has been revoked, so a new token cannot be fetched anyways. But even in this case, at least with `TokenSource.Expire` the user is informed about the invalid refresh token as soon as the access tokens are revoked via a helpful error message, not just after the current access token was supposed to expire.

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+
+// resolvePath resolves the given (relative) path and
+// returns a path suitable for API-calls.
+// This will consider the root-path of the fs and any needed prefixes.
+//
+// Any relative paths passed to functions that access these paths should
+// be resolved with this first!
+func (f *Fs) resolvePath(objectPath string) string {
+	resolved := path.Join(f.opt.RootPrefix, f.root, f.opt.Enc.FromStandardPath(objectPath))
+	if !f.opt.UnicodeNormalizationDisabled {
```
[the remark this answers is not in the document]
Impelon: I had to dig around in some team-discussions to remember the purpose of this one. :)
Apparently I originally planned to discuss this when opening the PR, but I forgot.

The unicode-normalization is required so that `fs/sync/sync_text.go:TestSyncUTFNorm` passes as intended.
It expects the two filenames to be treated equally, but HiDrive does not: We store filenames as raw-byte sequences and do not care about the encoding. Without the code above performing normalization on remote accesses, `rclone sync` will therefore fail to find the desired file, *because* it normalizes paths.

We decided that there are valid use cases for both accesses with and without normalization, so I made normalization optional.

backend/hidrive/helpers.go
Impelon: It is supposed to be the position of the last successful write, so that all the content before this position is guaranteed to be correct. Using your previous example
> What if block 1 was OK, block 2 failed and block 3 was OK.

`amount` should be the end position of block 1.

Impelon: > amount of successfully written bytes

is a bit misleading in that regard, as one could understand it to include the bytes from block 3. I'll try to clarify it more.

`Sucessful` in this case means that there are no failures up until that position; everything after the first failed write is declared unsuccessful.

Impelon: I'm not sure which of your options fits my description best (it's not 1), but I don't think `lastPosition` makes it clearer. But there may be another more fitting name?

backend/hidrive/hidrive.go
Impelon: The `ZeroDelayCalculator` is only used for the `retrier`-mechanism below.

The default pacer does wonderful things for API-calls, I would not want to replace it. :)

backend/hidrive/helpers.go
Impelon: > block 1 OK, block 3 OK, block 2 FAIL

Hm, I don't see anything wrong in this scenario. The desired result would be that `amount` points to the end of block 1.
I think this is exactly what the code is doing:
0. `amount = 0`
1. block 1 OK -> no upload error & no other upload failed -> `amount < end_block1` -> `amount = end_block1`
2. block 3 OK -> no upload error & no other upload failed -> `amount < end_block3` -> `amount = end_block3`
3. block 2 FAIL -> upload error & no other upload failed -> `amount = start_block2 (== end_block1)`

Impelon: > Either way I think you can tidy this if statement up by dealing with amount first if uploadErr == nil.

Not sure what you mean by this, could you elaborate? :)

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,902 @@
...
+// the first error that occurred will be returned.
+// This is not an atomic operation,
+// therefore if the upload fails the file may be partially modified.
+//
+// This returns fs.ErrorObjectNotFound if the object is not found.
+func (f *Fs) updateFileChunked(ctx context.Context, path string, content io.Reader, offset uint64, chunkSize int, transferLimit int64) (amount uint64, err error) {
+	amount = uint64(0)
+	mutex := new(sync.Mutex)
+	waitgroup, wgCtx := errgroup.WithContext(ctx)
+	transferSemaphore := semaphore.NewWeighted(transferLimit)
```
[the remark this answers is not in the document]
Impelon: Hm, interesting, I did not notice it before.

Although it will not cancel if the errgroup-context failed already, aka. the code would still be starting more routines after a failed upload. But maybe this would go hand-in-hand with your other suggestion to fail fast, before starting the upload.

backend/hidrive/helpers.go
Impelon: I think it may not start unnecessary uploads, because the semaphore takes the cancelled context into account, but your suggestion may also work well.

Impelon:
Alright, please excuse the long-winded responses.

I'll take the other comments into consideration and tick them of as `resolved` one by one as I fix them locally.

I'm very glad you are happy with the quality of the code! :)

backend/hidrive/hidrive.go
Impelon: Hopefully it never is, because as the help indicates, 2GiB is the maximum HiDrive can handle.
But overflows are ugly regardless, will fix this.

backend/hidrive/helpers.go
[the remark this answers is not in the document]
ncw: Yes, that is a very good point - the cancel on context failure is very good and is a good reason to use the `semaphore` so I suggest you leave it as it is.

backend/hidrive/hidrive.go
ncw: I see! Sorry careless reviewing on my part.

Yes I like the idea of renaming it `retryOnce` with a little comment - sounds perfect. You'll also get debug messages for retries which could well be useful.

We could also hoist `ZeroDelayCalculator` into `lib/pacers` - it looks like it could be useful elsewhere. Though it is only a few lines of code!

ncw:
> Alright, please excuse the long-winded responses.

Likewise!
 
> I'll take the other comments into consideration and tick them of as `resolved` one by one as I fix them locally.

Great. I've put some more thoughts inline and I think I've come to the bottom of the unicode normalization issue and found a better solution for the multipart upload end detection.

> I'm very glad you are happy with the quality of the code! :)

:-)

backend/hidrive/hidrive.go
Impelon: Sounds good, I'll create a separate commit for `ZeroDelayCalculator` and move it to `lib/pacers` then!

backend/hidrive/helpers.go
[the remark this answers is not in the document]
Impelon: > In Update you are supposed to be updating an existing object, not creating a new one.

You are right, `overwriteFile` can create a new file, if it does not exist yet, as it's documentation indicates.
Maybe a direct call to `patchFile` with offset 0 would be more appropriate; it will fail if the file does not exist yet.

I'll need to think a bit about the rest of your findings.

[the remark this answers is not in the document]
Impelon: You're right, actually _none_ of the other remotes uses `src.Remote()` in the `Update` method. My bad!

So from what I understand there are 2 things here:
1. `Update` uses the wrong path, if it is supposed to update an existing object.
2. `Update` is able to create a new object, which it is not supposed to do.

Your simple suggestion from above completely fixes 1. which is great, but `Update` is still technically able to create new objects. This patch would solve the 2. problem:
```diff
--- a/backend/hidrive/hidrive.go
+++ b/backend/hidrive/hidrive.go
@@@ -984,7 -984,7 +984,13 @@@ func (o *Object) Update(ctx context.Con
                // Metadata should be updated even if the upload fails.
                info, metaErr = o.fs.fetchMetadataForPath(ctx, resolvedPath, api.HiDriveObjectWithMetadataFields)
        } else {
--              info, err = o.fs.overwriteFile(ctx, resolvedPath, cachedReader(in), modTime)
++              // Single-part upload with patchFile;
++              // it guarantees not to create a new object,
++              // but to overwrite the existing one.
++              err = o.fs.patchFile(ctx, resolvedPath, cachedReader(in), 0, time.Time{})
++              if err == nil {
++                      info, err = o.fs.resizeFile(ctx, resolvedPath, uint64(src.Size()), modTime)
++              }
                metaErr = err
        }
```
But it is a bit harder to understand from a first glance. Without your patch this would then lead to `TestSyncUTFNorm` not being able to access the file, instead of creating a new one.

Of course your patch makes this 2. problem more or less obsolete, so I am wondering if you like my above patch or would prefer to keep it how it was previously?
The only situation I see where this makes a difference is, if someone deletes the remote file before rclone can `Update` its contents: With this additional patch the operation will fail then, while without it, a new file with the same name and location as the previous one would be silently created. I guess that would then be a question of who has priority: rclone or anyone else using the remote.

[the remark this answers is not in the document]
Impelon: Also your patch makes my reasoning for introducing unicode normalization in the backend obsolete, since `TestSyncUTFNorm` passes now without it.

Though, from what I understand you would still like to keep it around (but to copy the existing flag from the local backend)?
Don't misunderstand me, I'm not opposed to keeping normalization around; I'm just wondering what your reasoning is.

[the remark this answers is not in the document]
ncw: `Update` should try to update an existing file. The real purpose of this is for backends which can have duplicate file names (like Google Drive). It also helps with backends that keep an object history (eg revisions). Rclone will try to call `Update` if possible before calling `Put`.

Since your backend can't have duplicate file names, just using `o.remote` is probably enough. There may be reasons why you'd prefer to use patchfile - you'd be the expert on that.

So I'd say unless patchfile preserves some interesting metadata you'd lose otherwise (eg file revisions or something like that) then you should probably just upload it with the new name.

backend/hidrive/hidrive.go
ncw: Great - thanks.

backend/hidrive/helpers.go
Impelon: I'm resolving this as I'm now reworking the multi-part upload as discussed previously and I'm not sure this will apply anymore.
If you dislike anything with the reworked code please open a new comment when I push the new code. :)

Impelon:
Yes, just a second, was about to push. :)

PS:
By the way, apparently errgroup got a new method [`SetLimit(concurrency)`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit) on June 1st 2022, which essentially mirrors the way `pacer.NewTokenDispenser(concurrency)` works, but the token is grabbed implicitly at each invocation of `Go` and released through a defer. Although it does not seem to take the context into account... Just FYI :)

Impelon:
@ncw Alright, as you correctly suspected: I just pushed an updated version!

Some things to note:
- The [master branch of my fork](https://github.com/Impelon/rclone/tree/b2420758e5c7e07dafde463594b8534db192e63d) contains the new commits separately. The commits are different, but the changes are exactly the same (I double-checked with `git diff`). You might find it easier to see the changes I have made since the last push there. To keep the 3 commits that do not need further squashing separately, I squashed and reordered these smaller commits into the ones included in this PR.
- I introduced some unsafe casts from `uint64` to `int64` in the new multipart-upload code. As long as nobody uploads 8 exbibytes of data this should not be a problem!
- I removed the unicode normalization from this backend.

backend/hidrive/helpers.go
```diff
@@ -0,0 +1,888 @@
...
+	if err == nil {
+		info, err = f.resizeFile(ctx, path, okSize, modTime)
+	}
+	return okSize, info, err
+}
+
+// updateFileChunked updates the content of the existing file at the given path
+// starting at the given offset.
+// Returns the position of the last successfully written byte, stopping before the first failed write.
+// If nothing was written this will be 0.
```
Impelon, line 604:
```suggestion
// If nothing was written this will be equal to the given offset.
```

Actually, this is not true anymore for `updateFileChunked` specifically. `uploadFileChunked` is unaffected.

ncw:
> @ncw Alright, as you correctly suspected: I just pushed an updated version!
> 
> Some things to note:
> 
> * The [master branch of my fork](https://github.com/Impelon/rclone/tree/b2420758e5c7e07dafde463594b8534db192e63d) contains the new commits separately. Those commits are different, but the changes are exactly the same (I double-checked with `git diff`).

Thanks. Normally github is pretty good at keeping the conversation markers in the diff so you can see what has changed. It isn't as good as Gerrit at this sort of thing though!

>   You might find it easier to see the changes I have made since the last push there. (To keep the 3 commits in this PR, I squashed and reordered those smaller commits into the ones included in this PR.)
> * I introduced some unsafe casts from `uint64` to `int64` in the new multipart-upload code. As long as nobody uploads 8 exbibytes of data this should not be a problem!

Go uses signed ints for sizes of things. This seems like an odd choice, but actually it is very pragmatic as it allows you to signal other things with negative sizes. I think your users are unlikely to upload 8 EiB of files which according to my maths would take 140,737,488,355,328 x 16 TiB hard disks to store ;-)

>   This could be avoided if `ranges.Range` used `uint64` instead of `int64`, but there is probably little reason to introduce a change there.
> * I removed the unicode normalization from this backend.

Super.

I've been through the code and it looks great!

The CI is good, the integration tests pass, so I will merge now - thank you very much :-)

I'm planning to release this in v1.59 tomorrow. If you find anything you need fixing before then, then send another PR.

Impelon:
Great, I'm glad we got this into the release just in time! :tada:

Thank you for your in-depth review and helpful comments! It was a pleasure to work with you :)

stkr22:
Thanks @Impelon and @ncw for the awesome work! Really looking forward to start using the new backend in 1.59 🥳

ncw:
That's all released now :-)

PR #5823, opened 2021-11-17
backend:jottacloud change api used by ListR ( --fast-list )

Kimbsen:

To update the jottacloud backend to use a better performing api for the ListR() call. The current api `mode=list` has poor time to first byte and poor support for very large trees. The `mode=liststream` is better in both of those. 

No.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"encoding/xml"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


albertony:
Looking back at my notes - though 3 years old 😪... I considered `mode=liststream`, but found it did only work for requests at root of mountpoint, i.e. could not be used to list from any given folder. I guess this has been fixed now?

Kimbsen:
@albertony It has been fixed yes.

albertony:
Great!

Had a very quick look at the code (I'll leave the full review up to others):
- Seems it no longer takes the paths through `f.opt.Enc.ToStandardName`, if I'm not mistaken, you need that to correctly handle filenames with special characters (https://rclone.org/jottacloud/#restricted-filename-characters).
- Also, I notice this version does not get the mime type information from backend, and just hard codes it to `application/octet-stream`. That may be a problem in some cases, but perhaps acceptable as a documented limitation, I don't know.

Kimbsen:
The mimetype cannot be provided by this api no. But i'll look into using `f.opt.Enc.ToStandardName`.

ncw:
Do the integration tests (which test recursive listing) work OK?

    go install ./...
    test_all -backends jottacloud

You also need to run gofmt on the files I think according to the linter.

buengese:
If I recall this correctly, we do not guarantee that `Object.MimeType` returns a mime type even for backends that support this feature, so you should probably leave mime type empty instead of hard coding it.

As a side note: your test case for `TestParseListRStream` feels unnecessary large to me. It's not big issue but I'd like to avoid clutter

albertony:
> If I recall this correctly, we do not guarantee that `Object.MimeType` returns a mime type even for backends that support this feature, so you should probably leave mime type empty instead of hard coding it.

True:

https://github.com/rclone/rclone/blob/38dc3e93eef9cd38f0668ce7efca73a4ca188cbe/fs/types.go#L149-L151

Kimbsen:
> 
> 
> Do the integration tests (which test recursive listing) work OK?
> 
> ```
> go install ./...
> test_all -backends jottacloud
> ```
> 
> You also need to run gofmt on the files I think according to the linter.

The tests were failing. They are passing now though after https://github.com/rclone/rclone/pull/5823/commits/317f09aba203e3696e7fad98d4635da1f22f265d. Thanks for pointing those out to me. Kind of embarrased to not have run them before submitting the initial pr :) Diving into why they were failing was very informative :)


> 
> 
> If I recall this correctly, we do not guarantee that `Object.MimeType` returns a mime type even for backends that support this feature, so you should probably leave mime type empty instead of hard coding it.
> 
> As a side note: your test case for `TestParseListRStream` feels unnecessary large to me. It's not a big issue but I'd like to avoid clutter

I've now left the mimetype blank and i removed the test. The integration tests tests for the same thing so it was redundant.

The commit history is abit messy now though. Should i squash it?

ncw:
> > Do the integration tests (which test recursive listing) work OK?
> The tests were failing. They are passing now though after [317f09a]

Great!

> I've now left the mimetype blank and i removed the test. The integration tests tests for the same thing so it was redundant.

Sound good.

> The commit history is abit messy now though. Should i squash it?

Yes, please rebase on master, squash and force push.

@buengese do you want to give this a review when that is done?

Kimbsen:
> Yes, please rebase on master, squash and force push.

Done :)

backend/jottacloud/jottacloud.go
```diff
@@ -988,12 +1061,29 @@ func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (
...
+		resp, err = f.srv.Call(ctx, &opts)
+		if err != nil {
+			return shouldRetry(ctx, resp, err)
+		}
+
+		// raw path is /username/device/mountpoint/path
+		// liststream paths are /mountpoint/path
+		trimPrefix := f.filePathRaw("")
+		trimPrefix = trimPrefix[strings.Index(trimPrefix, "/")+1:]
+		trimPrefix = trimPrefix[strings.Index(trimPrefix, "/"):]
```
buengese, line 1078: Instead of building the root path using filePathRaw and trimming stuff back of it would probably be cleaner to just build the path you need here.

backend/jottacloud/jottacloud.go
```diff
@@ -931,49 +932,121 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
...
+			md5:         f.Checksum,
+			modTime:     time.Time(f.Modified),
+		})
+	}
+
+	trimPathPrefix := func(p string) string {
+		p = strings.TrimPrefix(p, path.Join(trimPrefix))
+		p = strings.TrimPrefix(p, "/")
+		return p
+	}
```
buengese, line 994: Path handling is still a bit messy and hard to follow, with you once again trimming off more stuff here. Also using `path.Join` to trim off a `/` suffix is far from clear.

backend/jottacloud/jottacloud.go
```diff
@@ -931,49 +932,121 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
...
-func (f *Fs) listFileDir(ctx context.Context, remoteStartPath string, startFolder *api.JottaFolder, fn listFileDirFn) error {
-	pathPrefix := "/" + f.filePathRaw("") // Non-escaped prefix of API paths to be cut off, to be left with the remote path including the remoteStartPath
-	pathPrefixLength := len(pathPrefix)
-	startPath := path.Join(pathPrefix, remoteStartPath) // Non-escaped API path up to and including remoteStartPath, to decide if it should be created as a new dir object
-	startPathLength := len(startPath)
-	for i := range startFolder.Folders {
-		folder := &startFolder.Folders[i]
-		if !f.validFolder(folder) {
-			return nil
+type listStreamTime time.Time
```
buengese, line 935: Do we actually have a third timeformat for Jottacloud now?  😨

buengese (commented):
A bit of nitpick on the path handling. Besides that it looks fine.

backend/jottacloud/jottacloud.go
Kimbsen: I agree. Will do that.

backend/jottacloud/jottacloud.go
Kimbsen: Yeah, that `path.Join()` should not have been there. I forgot it. It should just have been `p = strings.TrimPrefix(p, trimPrefix)`. Will fix that.

Kimbsen:
I pushed those changes @buengese . Let me know if i should rebase of master and squash again.

buengese:
That would be nice.

Kimbsen:
Done.

albertony:
There should probably be a note in this section that mime is not supported when using --fast-list:

https://rclone.org/jottacloud/#fast-list

buengese:
Oh, that's very true. I always forget about doc chances

Issue #4289, opened 2020-05-28
Rclone size reports incorrect size if --fast-list is used ( default behaviour )


xd003:
#### What is the problem you are having with rclone?

```rclone size remote:``` reports incorrect size for non team drive remotes if --fast-list is used . If i use ```rclone size remote: --disable ListR``` , i am getting accurate results

#### What is your rclone version (output from `rclone version`)

rclone v1.52.0

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

- os/arch: windows/386
- go version: go1.14.3

####  Which cloud storage system are you using? (eg Google Drive)

Google Drive

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
**Actual size - 2.853 gb**
**Test with rclone size remote:**
```
PS C:\rclone> rclone size test: -vv
2020/05/28 22:57:48 DEBUG : rclone: Version "v1.52.0" starting with parameters ["C:\\rclone\\rclone.exe" "size" "test:" "-vv"]
2020/05/28 22:57:48 DEBUG : Using config file from "C:\\Users\\Sanjeev\\.config\\rclone\\rclone.conf"
Total objects: 63
Total size: 292.184 MBytes (306377599 Bytes)
2020/05/28 22:57:52 DEBUG : 18 go routines active
PS C:\rclone> rclone size test: -vv
2020/05/28 22:58:01 DEBUG : rclone: Version "v1.52.0" starting with parameters ["C:\\rclone\\rclone.exe" "size" "test:" "-vv"]
2020/05/28 22:58:01 DEBUG : Using config file from "C:\\Users\\Sanjeev\\.config\\rclone\\rclone.conf"
Total objects: 74
Total size: 334.623 MBytes (350877842 Bytes)
2020/05/28 22:58:03 DEBUG : 18 go routines active
```
As you can see its incorrect and even reports inconsistent size randomly everytime

**Test with rclone size remote: --disable ListR**
```
PS C:\rclone> rclone size test: --disable ListR -vv
2020/05/28 23:05:18 DEBUG : rclone: Version "v1.52.0" starting with parameters ["C:\\rclone\\rclone.exe" "size" "test:" "--disable" "ListR" "-vv"]
2020/05/28 23:05:18 DEBUG : Using config file from "C:\\Users\\Sanjeev\\.config\\rclone\\rclone.conf"
2020/05/28 23:05:18 DEBUG : Reset feature "ListR"
[...]

PR #5832, opened 2021-11-20
drive: prevent retries when query filter returned no entries

ivandeex:

This patch fixes retries caused by optimization introduced by
- #5023

The `driveFs.ListR` result handler `listRRunner` treated empty directory listings returned by Google Drive in response to query constrained by `--min-age` and/or `--max-age` as an indication of the Google bug and retried the listing request with `--fast-list` turned off resulting in overall degraded performance.

There are two possible solutions for that:
1. disable query filter optimization in presence of `--fast-list` i.e. call ` SetUseFilter(ctx,false)` in `listRRunner()` to reset the optimization flag.
2. treat empty directory listings as an acceptable result if the query optimization was applied (ignoring possible server bugs)

This PR applies the latter solution, which bears risk of hitting the drive empty result bug
but it can provide better overall performance.

https://forum.rclone.org/t/performance-degradation-between-v1-56-2-and-v1-57-0-when-copying-to-google-drive-using-max-age-min-age-and-fast-list/27580/2

#### Checklist

    **IDK what a proper unit test should look like. Leaving unit test until review time.**
   **It's a bug fix, it needs no docs.**

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -1693,6 +1693,11 @@ func (f *Fs) listRRunner(ctx context.Context, wg *sync.WaitGroup, in chan listRE
 	var paths []string
 	var grouping int32
 
+	usingQueryFilter := false
+	if fi, use := filter.GetConfig(ctx), filter.GetUseFilter(ctx); fi != nil && use {
+		usingQueryFilter = true
+	}
+
 	for dir := range in {
 		dirs = append(dirs[:0], dir.id)
 		paths = append(paths[:0], dir.path)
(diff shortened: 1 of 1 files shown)


backend/drive/drive.go
```diff
@@ -1693,6 +1693,11 @@ func (f *Fs) listRRunner(ctx context.Context, wg *sync.WaitGroup, in chan listRE
 	var paths []string
 	var grouping int32
 
+	usingQueryFilter := false
```
ncw, line 1696: I don't think this test is working - when I tried it it always seemed to set `usingQueryFilter = true`

I was trying the example here:

https://github.com/rclone/rclone/issues/4289#issuecomment-636016383

Which you can probably try too since the folder is public.

ncw (commented):
I'm not sure this is the right fix... Imagine that `--min-age 0s` was set to select all files. This would then activate the bug.

I checked the bug is still present using the example here: https://github.com/rclone/rclone/issues/4289#issuecomment-636016383

I think this needs to not use ListR at all if the query filter is in use. 

We don't have an easy way of doing this other than by calling walk.ListR like the union backend does here

https://github.com/rclone/rclone/blob/a351484997b84682af72412f08e49afd029376cb/backend/union/union.go#L669

ncw:
I'm going to close this as it is no longer applies and we were uncertain of the approach.

Issue #4885, opened 2020-12-25
Unable to mount Webdav remote root URL using MEGAcmd as webdav server


AllanVan:
#### What is the problem you are having with rclone?
There is an issue with MEGA authentication that requires a workaround, suggested in #3393. This will be implemented in the future, it seems, according to #3660 .

Anyway, I thought I would open this because it might affect the development of the feature. Wasn't sure if I should post in any of the above issues, so sorry if I was supposed to.

The issue is related to the use of the whitespace character when MEGAcmd creates the path to the root mount "/". 
So if you run "webdav /" in MEGAcmd, it returns
http://127.0.0.1:4443/XXXXXX/Cloud%20Drive

If I use this path when creating the webdav remote in rclone, it throws the error in the section below, when trying to run a command. It seems to be an issue with %20.
What I tried to do:
- encasing the URL in quotes - doesn't work, another error in the section below
- using whitespace itself
- ending URL with /
- changing %20 to another character from https://rclone.org/local/ - tried null, space, both Value and Representation items. Also tried &, and.
- checked MEGAcmd issues for a workaround, but didn't find any, probably because it works in some places (Web browser opens the link, for example)

Related ,but not an issue of rclone: Windows' network drive mapping also fails because of the same reason. I can map it just fine, but when trying to access through Explorer, it says it's inaccessible.

Only workaround I found is to use a folder that contains all folders. Then you can reference it when creating the webdav in MEGAcmd, and as long as there is no whitespace or %20, rclone access it just fine (Windows Explorer too).

Is there any better workaround on rclone's end? 
I'll suggest being able to change "Cloud%20Drive" to root or something in the MEGAcmd github, but if someone uses a folder with all folders inside as the Webdav mount, and it has whitespace in the name, the issue remains in rclone's end.

#### What is your rclone version (output from `rclone version`)
v1.53.2
Also tested on latest beta version.
I run rclone on cmd, MEGAcmd directly from executable.

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Windows 7, 64bit

####  Which cloud storage system are you using? (e.g. Google Drive)
MEGA through Webdav backend (MEGAcmd)

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

rclone size MegaWebDav:

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

[...]

Issue #4695, opened 2020-10-19
Mega.nz Session Cache


metabox-cloud:
#### What is your current rclone version (output from `rclone version`)?
rclone v1.53.1

#### What problem are you are trying to solve?

` 2020/10/19 02:35:08 ERROR : *FILENAME*.rar: Failed to copy: upload file failed to create session: Object (typically, node or user) not found`

#### How do you think rclone should be changed to solve that?
I have contacted Mega.nz support, which is trying to get me to go to megacmd, although all my stuff is rclone based.. so don't really want to, and they have informed me that rclone doesn't cache the session on mega. which causes multiple logins (one for every file) and in part, gets limited in the end.. especially if running multiple threads over multiple servers..

![ZTYX5FUhLL](https://user-images.githubusercontent.com/72478288/96390345-ba00a500-11fb-11eb-90a3-9f5e093ca53b.png)

ImportTaste:
Fuck, no wonder. So this is what's been going wrong.

ncw:
Rclone logs in each time it is run, so if you running rclone lots then you'll get lots of logins.

However rclone does work with the webdav server in megacmd quite well. It isn't quite as fast as using rclone directly but it works pretty well.

ImportTaste:
My current workaround, on Windows, is mounting my MEGA with a batch script...

```batch
start "rcd" cmd /c rclone rcd --rc-no-auth
rclone rc mount/mount fs=MEGA: mountPoint=C:/mount/MEGA "vfsOpt={\"CacheMode\": 2, \"CaseInsensitive\": false}"
```

And then performing operations on the mount instead, using `rclone rc vfs/forget` to refresh the mounted file system cache as needed.

ncw:
Yes, running a mount (or `rclone serve webdav` etc) will also make rclone persistent.

ImportTaste:
> Rclone logs in each time it is run, so if you running rclone lots then you'll get lots of logins.
> 
> However rclone does work with the webdav server in megacmd quite well. It isn't quite as fast as using rclone directly but it works pretty well.

I tried doing what you suggested here, using megacmd's webdav server, but apparently this is prone to IO errors like these:
```
2020/11/16 12:48:44 ERROR : IO error: couldn't list files: XML syntax error on line 42: invalid character entity &percnt;
2020/11/16 12:50:00 ERROR : IO error: couldn't list files: XML syntax error on line 253: invalid character entity &add;
[...]

Issue #3345, opened 2019-07-13
upstream XML decoding bug.. command rclone info has broken s3 remote


SocialCrypt-Verify:
#### Output of `rclone version`
rclone v1.48.0
- os/arch: linux/amd64
- go version: go1.12.6

Ubuntu 18.04.2 LTS
#### Describe the issue
My setup is S3-Compatable (Scaleway) remote >> Crypt.

After unmounting the crypt remote "encloud" using "fusermount -u", I then manually cleared the vfs cache folder, and remounted with --cache-db-purge option. But then I was unable to list the directory or perform any other action on that remote, even when mounting with bypassing the crypt remote to see the encrypted files. Object storage crashed. The Scaleway support had to apply a hotfix for this bug at their end and had coprehensively explained the issue. 

**Scaleway team replicated the same situation on Amazon S3 and got the same error from rclone.**

Mounting string:
`rclone mount encloud: /mnt/s3 --allow-other --contimeout 10m --max-read-ahead 64M --log-file "/var/www/rclone.log" --log-level NOTICE --fast-list --transfers 16 --buffer-size 64M --multi-thread-cutoff 64M --multi-thread-streams 16 --poll-interval 0 --vfs-cache-mode full --dir-cache-time 2160h --vfs-cache-max-age 720h --vfs-cache-max-size 10G --vfs-read-chunk-size 64M --vfs-read-chunk-size-limit 128M --retries 200 --retries-sleep 1s --no-gzip-encoding --stats 24h --stats-one-line --gid 33 --uid 33 --umask 007 --daemon`

Here is the rclone config:
`[cloud]
type = s3
provider = Other
access_key_id = ***
secret_access_key = ***
region = nl-ams
endpoint = s3.nl-ams.scw.cloud
acl = private
bucket_acl = private
upload_cutoff = 256M
chunk_size = 256M
upload_concurrency = 8
force_path_style = false

[encloud]
type = crypt
remote = cloud:/album/cloud
filename_encryption = standard
directory_name_encryption = true
password = ***
password2 = ***`

Scaleway Team response to he issue:

> Hello,
> 
> First of all, thank you for your report, we did have an encoding bug on our
> end, which is now fixed. The problem you are now having is on the rclone's
> side, and there is nothing we can do about it.
> 
> Rclone did upload object with control characters in the names. For example:
> 
>     data/-position-left-05
> 
> Which is:
> 
[...]

PR #5848, opened 2021-12-01
lib/rest: process HTML entities within XML

CodingKoopa:

The MEGA SDK - and, by extension, MEGAcmd - [escapes many HTML entities](https://github.com/meganz/sdk/blob/f1f05bc7d7ec67cb6d03d9e327ec3ee64e82562d/src/utils.cpp#L2125-L2398) in the XML it serves as the WebDAV server. This is not conformant XML, and I have already [submitted a PR](https://github.com/meganz/sdk/pull/2600) to fix this. That having been said, I am submitting a fix for rclone too for a couple of reasons:
- MEGAcmd development seems to be rather slow, and it could be some time until it's updated with the latest SDK - that is, _if_ a fix is merged into the SDK. Moreover, the `mega` backend currently has an issue (#4695) that makes it difficult to use, making this WebDAV functionality useful.
- The MEGA SDK might not be the only WebDAV application exhibiting this behavior. In these cases, it should be preferable for rclone to process the entities rather than throwing an error.

For reference, this is what currently happens when using a problematic filename with MEGAcmd's WebDAV server, and rclone:
```
2020/11/16 12:50:00 ERROR : IO error: couldn't list files: XML syntax error on line 253: invalid character entity &add;
```

Most directly, this issue is reported [here](https://github.com/rclone/rclone/issues/4695#issuecomment-728256438) and in #4885. The "rather strict" XML parser is also discussed in [this](https://github.com/rclone/rclone/issues/3345#issuecomment-532625962) comment.

#### Checklist

Thanks!

### DIFF
--- a/lib/rest/rest.go
+++ b/lib/rest/rest.go
@@ -167,6 +167,10 @@ func DecodeJSON(resp *http.Response, result interface{}) (err error) {
 func DecodeXML(resp *http.Response, result interface{}) (err error) {
 	defer fs.CheckClose(resp.Body, &err)
 	decoder := xml.NewDecoder(resp.Body)
+	// MEGAcmd has included escaped HTML entities in its XML output, so we have to be able to
+	// decode them.
+	decoder.Strict = false
+	decoder.Entity = xml.HTMLEntity
 	return decoder.Decode(result)
 }
 


ncw:
What do you think the backwards compatibility problems with this patch might be given that the jottacloud and sugarsync backends use the XML parser in lib/rest as well as the webdav backend. Does this need to be an option?

CodingKoopa:
Great question! I did notice that those backends use this function too, and indeed was concerned about breaking them. I didn't address it initially because I deemed the act of processing HTML entities to be a "non-destructive" action, particularly as the alternative is the parser throwing an error. If you think it would be wiser to restrict this to webdav, I could make it an option.

ncw:
Here are the docs: https://pkg.go.dev/encoding/xml#Decoder

```
	// Strict defaults to true, enforcing the requirements
	// of the XML specification.
	// If set to false, the parser allows input containing common
	// mistakes:
	//	* If an element is missing an end tag, the parser invents
	//	  end tags as necessary to keep the return values from Token
	//	  properly balanced.
	//	* In attribute values and character data, unknown or malformed
	//	  character entities (sequences beginning with &) are left alone.
```

The first fix seems OK and unlikely to cause a problem. The second fix - means that the parse would have thrown an errror rather than leave the undecodable entities according to the source code, so that is OK too.

The second part of your patch is `decoder.Entity = xml.HTMLEntity` and after a look at the source code of the XML library, I don't think this will cause a problem either since those entities would have thrown an error before.

So I think, on further reflection, this patch is safe and won't break anything.

We'll see soon enough in the integration tests and if it does cause a problem we'll need to make it conditional.

So I'll merge this now - thank you :-)

ncw:
PS since you are interested in using mega with webdav, would you like to write some docs in the webdav backend for it? Or maybe we could make a webdav provider for it? Possibly rclone could even start/stop the mega process.

CodingKoopa:
Sure, I can think about what I could do ^^

Issue #5612, opened 2021-09-20
Specific error message when attempting a transfer exceeding the file path limit on Windows


CarlosHartmann:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/display-warnings-for-paths-that-exceed-windows-path-length-limit/26608

#### What is your current rclone version (output from `rclone version`)?

rclone v1.55.1
- os/type: windows
- os/arch: amd64
- go/version: go1.16.3
- go/linking: dynamic
- go/tags: cmount

#### What problem are you are trying to solve?

When nounc is true in the conf ([explanation](https://rclone.org/local/)) and a file transfer is attempted that would exceed the Windows path length limit, the error message reads "The system could not find the path". The problem is that this error message is ambiguous. 

#### How do you think rclone should be changed to solve that?

An error message specifically for this case should appear. That could be solved by counting the path length before transferring on Windows machines. As I don't know how rclone works on the inside, I have no idea if this is realistic.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

abhishek-dalbanjan:
Can i take this up as my first issue? @CarlosHartmann Can you assign it to me?

PR #5862, opened 2021-12-06
lib/file: check max path length for windows

herlon214:

Show a better error message when reach max file path for Windows.

https://forum.rclone.org/t/display-warnings-for-paths-that-exceed-windows-path-length-limit/26608
https://github.com/rclone/rclone/issues/5612

#### Checklist

### DIFF
--- a/lib/file/file_windows.go
+++ b/lib/file/file_windows.go
@@ -28,6 +28,12 @@ func OpenFile(path string, mode int, perm os.FileMode) (*os.File, error) {
 	if len(path) == 0 {
 		return nil, syscall.ERROR_FILE_NOT_FOUND
 	}
+	// For windows the max path length is 260 characters
+	// https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+	if len(path) >= 260 {
+		return nil, &os.PathError{Path: path, Op: "open", Err: errors.New("path length higher than 260")}
+	}
+
 	pathp, err := syscall.UTF16PtrFromString(path)
 	if err != nil {
 		return nil, err
--- a/lib/file/file_windows_test.go
+++ b/lib/file/file_windows_test.go
@@ -1,0 +1,16 @@
+//go:build windows
+// +build windows
+
+package file
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFileOpenMaxPathLength(t *testing.T) {
+	path := `C:\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path`
+	_, err := OpenFile(path, 0644, 0644)
+	assert.Error(t, err, "mkdir: Max path length can't exceed 260")
+}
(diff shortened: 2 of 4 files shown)


Kintar (changes requested):
The 260 character max length is not necessarily true. In Windows 10 v1607 and later, you can opt-in to longer paths by setting a registry value. See https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd

Maybe add a check for that registry value using the registry library from golang.org/x/sys/windows/registry?

albertony:
Does not this change assume `nounc = true` is always set?

herlon214:
@Kintar I added the check there, but I'm not sure how can we mock that to test.
For now I just skipped the test if the `LongPaths` are enabled.

herlon214:
@albertony Sorry, I don't have a lot of experience in this lib. I'm just trying to help with the good first issue.
Could you explain a little bit better?

albertony:
> I'm just trying to help with the good first issue.

Sure, and that is highly appreciated. I didn't mean to be negative.

> Could you explain a little bit better?

With extended-length path format, the existing functions should work also with paths of 260 characters and longer. E.g., the example from your test case, if it has prefix `\\?\` then it cannot return the error `path length higher than 260` (by the way: strictly speaking "path length 260 or higher" would be more correct):

```
\\?\C:\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path\my\path
```

Rclone uses this syntax by default, automatically converts "regular" paths into extended-length paths. Which means you are *not* limited to <260 characters in rclone - **unless** you set the config option `nounc` to disable this conversion. This is documented here: https://rclone.org/local/#long-paths

Unless I've misunderstood something, the original issue/forum thread came from a user who had `nounc` set.

herlon214:
@albertony got it.

I added the check, can you take a look?

albertony:
Great, that should cover it.
Though, to be honest, I am a bit worried that such a test as this may incorrectly trigger on cases that would in reality work if we instead just left it up to Go/OS functions to "decide", but perhaps I'm just 😱 ... So I'll 🙈 and leave the final review up to others. 😆 In any case: Thanks for your contribution!

lib/file/file_windows.go
```diff
@@ -101,3 +112,24 @@ func IsReserved(path string) error {
 	}
 	return nil
 }
+
+// IsLongPathsEnabled checks if the current have the LongPathsEnabled enabled
+func IsLongPathsEnabled() bool {
```
ncw, line 117: I conjecture that reading stuff from the registry might be time consuming. Rclone creates a lot of directories so should we cache this?

I'd suggest a couple of package level variables a [sync.Once](https://pkg.go.dev/sync#Once) and a bool to hold the cached value.

lib/file/file_windows.go
```diff
@@ -101,3 +112,24 @@ func IsReserved(path string) error {
...
+		return false
+	}
+	defer k.Close()
+
+	s, _, err := k.GetIntegerValue("LongPathsEnabled")
+	if err != nil {
+		return false
+	}
+
+	if s == 0x00000001 {
```
ncw, line 130: I'd probably write this as

    return s == 0x00000001

Or maybe

    return s != 0

ncw (commented):
This looks like a useful piece of work - thank you :-)

I've put some comments in line for you to take a look at.

ncw:
Since I haven't heard from you for a while I'm going to close this.

Please feel free to re-open or resubmit with an updated version.

Thank you

PR #5875, opened 2021-12-19
vfs: Fix bug caused by vfs cache special character ("rune") escape ("quoting") inconsistency

cosine0:

When `vfs-cache` is enabled, rclone fails to upload every file with a name that contains special characters (those which are called "runes" in the rclone's codebase). But, if the runes are only in the directory names, it succeeds.
From what I have inspected, it is because rclone copies original files to cache with a fully escaped path (with a quote character in front of each rune), but when it flushes the cache to the actual remote, it looks for the cached file in a partially escaped path (with a quote character only in front of each rune in the directory part of the path).
The correct operation would be fully escaping the cache paths every time they are referred to.
So, I submit the patch in this pull request.

No

#### Checklist

### DIFF
--- a/vfs/vfscache/cache.go
+++ b/vfs/vfscache/cache.go
@@ -196,7 +196,6 @@ func createRootDirs(parentOSPath string, relativeDirOSPath string) (dataOSPath s
 // Returns an os path for the data cache file.
 func (c *Cache) createItemDir(name string) (string, error) {
 	parent := vfscommon.FindParent(name)
-	leaf := filepath.Base(name)
 	parentPath := c.toOSPath(parent)
 	err := createDir(parentPath)
 	if err != nil {
@@ -207,7 +206,7 @@ func (c *Cache) createItemDir(name string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to create metadata cache item directory: %w", err)
 	}
-	return filepath.Join(parentPath, leaf), nil
+	return c.toOSPath(name), nil
 }
 
 // getBackend gets a backend for a cache root dir


albertony:
Yes, that seem like a clear bug.. Good catch!

I think your fix looks great. The only question is whether to take advantage of the fact that `parentPath` has already been converted to OS Path, e.g.:

```
leaf := c.toOSPath(filepath.Base(name))
return filepath.Join(parentPath, leaf), nil
```

But I think your version looks safe and just as good - maybe even better.

Let's see if someone else wants a quick review before merging it.

Thanks!

albertony (approved):

cosine0:
@albertony
> I think your fix looks great. The only question is whether to take advantage of the fact that `parentPath` has already been converted to OS Path, e.g.:
> 
> ```
> leaf := c.toOSPath(filepath.Base(name))
> return filepath.Join(parentPath, leaf), nil
> ```
> 
> But I think your version looks safe and just as good - maybe even better.

Right. I found it more appealing to use my way in a readability/semantic manner; because at my first glance, `c.toOSPath` somewhat "magically" prepended proper absolute root path, and I thought separately applying `c.toOSPath` or similar conversion to the path components and joining might not work. And I assumed confusion about those concepts made this bug.
However, as long as it fixes this bug (this bug has bothered my google drive-to-onedrive migration from long ago...), any way convenient to you community would be great.

Thank you for your hard work!

albertony:
Thank you. I'm gonna merge this now (I think it's best to get this bugfix into beta now, I've reviewed it thoroughly and am confident it is safe and sound, *and* I have some conflicting changes coming up...).

PR #5878, opened 2021-12-20
cmd: fix broken example link in help.go

vsoch:
This link appears to be broken, so here is another reference to (I think) the same file that provides a good example of coba. We could also do the current commit https://github.com/moby/moby/blob/8312004f41e9500824fa16ae991eeee0083f4771/cli/cobra.go although it might be better to maintain an up to date example.

The link to the example for how to format usage strings appears to be broken  (404).

Nope.

#### Checklist

- [ ] I have added tests for all changes in this PR if appropriate (not appropriate).
- [ ] I have added documentation for the changes if appropriate. (not appropriate)

### DIFF
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -165,7 +165,7 @@ func runRoot(cmd *cobra.Command, args []string) {
 // setupRootCommand sets default usage, help, and error handling for
 // the root command.
 //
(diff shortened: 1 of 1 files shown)


ncw (approved):
This looks great thank you :-)

PR #5879, opened 2021-12-21
Add curl to the rclone docker image

onedr0p:

Add curl to the Docker image, this is useful if you use the container to perform a sync and want to send an HTTP request to a service like https://healthchecks.io/ when the sync has completed. Take for example:

```yaml
---
version: "3.8"

services:
  backup:
    image: docker.io/rclone/rclone:1.57.0
    container_name: backup
    entrypoint: /bin/sh
    command:
      - "rclone"
      - "--config"
      - "/config/rclone.conf"
      - "--progress"
      - "sync"
      - "/mnt/test"
      - "backblaze:rcUFz3wc/"
      - "&&"
      - "curl"
      - "-fsS"
      - "-m" 
      - "10"
      - "--retry"
      - "5"
      - "-o"
      - "/dev/null https://healthchecks.io/xyz"
```

This way I will be only notified if the command fails and my sync didn't happen.

There's many other benefits to having curl in the image as well.

No

#### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN ./rclone version
 # Begin final image
 FROM alpine:latest
 
-RUN apk --no-cache add ca-certificates fuse tzdata && \
+RUN apk --no-cache add ca-certificates curl fuse tzdata && \
   echo "user_allow_other" >> /etc/fuse.conf
 
 COPY --from=builder /go/src/github.com/rclone/rclone/rclone /usr/local/bin/


ncw:
Why not use rclone to do the http request instead?

`rclone copyurl` is a poor man's version of curl

onedr0p:
I did not know about that feature. The nice thing about curl is the retries and the timeout and other flags it supports.

ncw:
I don't really want to add curl to the very minimal rclone image - its very easy to build your own image based on the rclone image if that is what you want without having to bulk up the rclone image.

I just thought I'd point out that rclone has some curl functionality already.  It had retries and timeouts too but in no way has all the features of curl, though it does have a surprising number of them!

Issue #1770, opened 2017-10-24
Can't connect to SharePoint Online team sites such as https://orgname.sharepoint.com/sites/Site-Name


hi2u:
I’ve been able to successfully connect to the default https://orgname-my.sharepoint.com/ personal SharePoint Site...

```
$ rclone lsd sp3:
-1 2017-01-04 22:16:34         0 Attachments
-1 2015-01-23 11:13:10         0 Shared with Everyone
```

But I’m having difficultly figuring out how to connect to team sites on URLs such a: https://orgname.sharepoint.com/sites/Site-Name etc.

The “rclone config” guided process doesn’t let you set the resource_url when setting it up. So I’ve tried editing ~/.config/rclone.conf using a few different methods, changing the resource_url and then reauthorizing,  I've tried a number of different addresses like...

For the main/default team site:
```
https://orgname.sharepoint.com/ 
https://orgname.sharepoint.com/Shared Documents
```

For separate team sites, or what Microsoft call "site collections":

```
https://orgname.sharepoint.com/sites/Site-Name
https://orgname.sharepoint.com/sites/Site-Name/
https://orgname.sharepoint.com/sites/Site-Name/Shared Documents
https://orgname.sharepoint.com/sites/Site-Name/Shared Documents/
https://orgname.sharepoint.com/sites/Site-Name/Shared%20Documents
https://orgname.sharepoint.com/sites/Site-Name/Shared%20Documents/
```

I'm not sure which address format I'm meant to use? (for either the main team site, or all the other ones under /sites/)

I always get the error:

```
$ rclone -vv lsd sp3:
2017/10/25 03:17:18 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2017/10/25 03:17:18 DEBUG : rclone: Version "v1.38" starting with parameters ["rclone" "-vv" "lsd" "sp3:"]
2017/10/25 03:17:19 Failed to create file system for "sp3:": failed to get root: 401 Unauthorized: 
```
(there's nothing after that last colon)

Does anyone know how I access team SharePoint sites?

My rclone version is:
```
rclone v1.38
- os/arch: linux/amd64
- go version: go1.9
```
...on Manjaro 64bit, installed from the distro's repos.

I'm choosing the "business" option when asked in `rclone config`.

ncw:
@olihey any ideas how we could make this work?

olihey:
Changing the resource url in the rclone.conf will not work as the access token is only valid for ONE resource url.

[...]

PR #5883, opened 2021-12-25
onedrive: about the Sites.Read.All permission request

cqjjjzr:

In fix of #1770 , a new permission request was added in the OneDrive backend support: `Sites.Read.All`. This is intended to solve the error when using a custom SharePoint site. However, such changes are undocumented. 

On a domain with an excessively strict authorization policy (e.g. my university), the domain disallows users to consent permissions on their own and the `Sites.Read.All` permission is undocumented thus not configured, the extra `Sites.Read.All` scope when requesting an access token will trigger "Application needs permission to access resources in your organization that only an admin can grant. Please ask an admin to grant permission to this app before you can use it." message.

Therefore, this PR involves a new advanced config entry `disable_site_permission` (default to false). When turned on, this config will prevent the OAuth client to request `Sites.Read.All` permission. The "Search for a SharePoint site" may be unusable because of this, but for those under a strict policy who don't need custom site configuration, this should be enough.

For the documentation, since I'm unfamiliar with the documentation structure, I didn't attempt to include all documentation for this. Instructions on writing docs will be much appreciated

See #1770

#### Checklist

- [ ] I have added tests for all changes in this PR if appropriate. (not applicable)
- [ ] I have added documentation for the changes if appropriate. (See above)

### DIFF
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -144,9 +144,9 @@ Note that the chunks will be buffered into memory.`,
 			Name: "disable_site_permission",
 			Help: `Disable the request for Sites.Read.All permission.
 
-If enabled, you will no longer be able to search for a SharePoint site
-when configuring drive ID.
-Enable if your organization didn't assign Sites.Read.All permission to the
+If set to true, you will no longer be able to search for a SharePoint site when
+configuring drive ID, because rclone will not request Sites.Read.All permission.
+Set it to true if your organization didn't assign Sites.Read.All permission to the
 application, and your organization disallows users to consent app permission
 request on their own.`,
 			Default:  false,
@@ -388,8 +388,8 @@ func Config(ctx context.Context, name string, m configmap.Mapper, config fs.Conf
 	region, graphURL := getRegionURL(m)
 
 	if config.State == "" {
-		disableSitePermission, ok := m.Get("disable_site_permission")
-		if disableSitePermission == "true" && ok {
+		disableSitePermission, _ := m.Get("disable_site_permission")
+		if disableSitePermission == "true" {
 			oauthConfig.Scopes = scopesWithoutSitePermission
 		} else {
 			oauthConfig.Scopes = scopesWithSitePermission
--- a/backend/onedrive/onedrive.go
+++ b/backend/onedrive/onedrive.go
@@ -65,9 +65,12 @@ var (
 	authPath  = "/common/oauth2/v2.0/authorize"
 	tokenPath = "/common/oauth2/v2.0/token"
 
+	scopesWithSitePermission    = []string{"Files.Read", "Files.ReadWrite", "Files.Read.All", "Files.ReadWrite.All", "offline_access", "Sites.Read.All"}
+	scopesWithoutSitePermission = []string{"Files.Read", "Files.ReadWrite", "Files.Read.All", "Files.ReadWrite.All", "offline_access"}
+
 	// Description of how to auth for this app for a business account
 	oauthConfig = &oauth2.Config{
-		Scopes:       []string{"Files.Read", "Files.ReadWrite", "Files.Read.All", "Files.ReadWrite.All", "offline_access", "Sites.Read.All"},
+		Scopes:       scopesWithSitePermission,
 		ClientID:     rcloneClientID,
 		ClientSecret: obscure.MustReveal(rcloneEncryptedClientSecret),
 		RedirectURL:  oauthutil.RedirectLocalhostURL,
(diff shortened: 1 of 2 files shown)


cqjjjzr:
I added (partial) docs about the Sites.Read.All permission.

backend/onedrive/onedrive.go
```diff
@@ -137,6 +140,17 @@ Note that the chunks will be buffered into memory.`,
...
 			Advanced: true,
+		}, {
+			Name: "disable_site_permission",
+			Help: `Disable the request for Sites.Read.All permission.
+
+If enabled, you will no longer be able to search for a SharePoint site
+when configuring drive ID.
+Enable if your organization didn't assign Sites.Read.All permission to the
+application, and your organization disallows users to consent app permission
+request on their own.`,
```
Cnly, line 151: The name of the option has the word "disable" and the documentation uses "enable." This can create confusion. I'd suggest changing the wording in the docs to something like this: "If set to true, ..." and "Set to true if ..."

backend/onedrive/onedrive.go
```diff
@@ -374,6 +388,12 @@ func Config(ctx context.Context, name string, m configmap.Mapper, config fs.Conf
 	region, graphURL := getRegionURL(m)
 
 	if config.State == "" {
+		disableSitePermission, ok := m.Get("disable_site_permission")
+		if disableSitePermission == "true" && ok {
```
Cnly, line 392:
```suggestion
		if ok && disableSitePermission == "true" {
```

Usually we put `ok` before the value predicate. Or we can just remove it here.

backend/onedrive/onedrive.go
```diff
@@ -374,6 +388,12 @@ func Config(ctx context.Context, name string, m configmap.Mapper, config fs.Conf
...
+			oauthConfig.Scopes = scopesWithoutSitePermission
+		} else {
```
Cnly, line 394: We should be able to get rid of this else branch. (The same for the other if-else block)

docs/content/onedrive.md
```diff
@@ -132,7 +132,7 @@ Client ID and Key by following the steps below:
...
 3. Under `manage` select `Certificates & secrets`, click `New client secret`. Enter a description (can be anything) and set `Expires` to 24 months. Copy and keep that secret _Value_ for later use (you _won't_ be able to see this value afterwards).
 4. Under `manage` select `API permissions`, click `Add a permission` and select `Microsoft Graph` then select `delegated permissions`.
-5. Search and select the following permissions: `Files.Read`, `Files.ReadWrite`, `Files.Read.All`, `Files.ReadWrite.All`, `offline_access`, `User.Read`. Once selected click `Add permissions` at the bottom.
+5. Search and select the following permissions: `Files.Read`, `Files.ReadWrite`, `Files.Read.All`, `Files.ReadWrite.All`, `offline_access`, `User.Read`, `Sites.Read.All` (optional). Once selected click `Add permissions` at the bottom.
```
Cnly, line 135:
```suggestion
5. Search and select the following permissions: `Files.Read`, `Files.ReadWrite`, `Files.Read.All`, `Files.ReadWrite.All`, `offline_access`, `User.Read`, and optionally `Sites.Read.All` (see below). Once selected click `Add permissions` at the bottom.
```

Perhaps this and an extra sentence below explaining the option & linking to the issue #1770 would be better.

Cnly (changes requested):
Thanks for the PR @cqjjjzr! Please see my comments below :)

backend/onedrive/onedrive.go
cqjjjzr: How? Should I lift the `oauthConfig.Scopes = scopesWithSitePermission` up before the `if` statement? The `scopesWithSitePermission` is configured as the default, so here it's just for belt and braces..

cqjjjzr:
Should I push multiple commits or squash them using rebase and do a force push?

backend/onedrive/onedrive.go
Cnly: I initially thought we could remove it altogether, but now I see that the object can get reused, so I think it's good as it is, no changes needed.

backend/onedrive/onedrive.go
Cnly: After some more thinking I think we should just remove the check on `ok` to allow possible changes to the default value. If at some later point we decided `true` will be better as the default, the `ok` check here will break that as it assumes `false` should be the default.

Cnly:
You don't need to force push :) Force pushing will probably mess up with the review history. Just push them normally and everything will be squashed before merging into master.

Cnly (approved):
Many thanks! I'll merge this now.

PR #5884, opened 2021-12-27
onedrive: minor opt of quickxorhash

isaacl:
Avoid copying the hash array.

#### Checklist

### DIFF
--- a/backend/onedrive/quickxorhash/quickxorhash.go
+++ b/backend/onedrive/quickxorhash/quickxorhash.go
@@ -136,7 +136,8 @@ func (q *quickXorHash) Write(p []byte) (n int, err error) {
 func (q *quickXorHash) checkSum() (h [Size]byte) {
 	// Output the data as little endian bytes
 	ph := 0
-	for _, d := range q.data[:len(q.data)-1] {
+	for i := 0; i < len(q.data)-1; i++ {
+		d := q.data[i]
 		_ = h[ph+7] // bounds check
 		h[ph+0] = byte(d >> (8 * 0))
 		h[ph+1] = byte(d >> (8 * 1))


backend/onedrive/quickxorhash/quickxorhash.go
```diff
@@ -136,7 +136,8 @@ func (q *quickXorHash) Write(p []byte) (n int, err error) {
 func (q *quickXorHash) checkSum() (h [Size]byte) {
 	// Output the data as little endian bytes
 	ph := 0
-	for _, d := range q.data[:len(q.data)-1] {
```
ncw, line 139: This doesn't copy the `q.data` array, it slices it, so the backing store remains in the same place. Thus its quite a cheap thing to do as all it is is filling in 3 fields in the slice header.

If you look at a disassembly you'll see that your code saves about 5 instructions of setup.

https://godbolt.org/z/jh1rM7coY

So it is probably worth doing but the difference it makes will be very small.

ncw (commented):
I don't think this makes as much performance impact as you might think - see comments inline! It is probably worth merging though.

Thoughts?

ncw:
> ok! I had read on go documentation that slices copy the backing array. Plus of course, the main expense of this hashing algorithm is in the hashing part, not the digest. I’m not sure if there’s any way to make this hasher faster, it’s a complicated af algorithm

I haven't made any attempt to optimize this so I expect there is low hanging fruit!

If you want to have a go I'd make a benchmark, maybe something like this one

https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/crypto/md5/md5_test.go;l=218

then try to eliminate all array bounds checks from the loop here (article https://go101.org/article/bounds-check-elimination.html )

https://github.com/rclone/rclone/blob/9d4eab32d82c304666d7d2331ab3c4f6ca0ed3eb/backend/onedrive/quickxorhash/quickxorhash.go#L72

I'd be surprised if you couldn't get it to run 10 or 20% faster with only a few small changes.

I'll merge what you've done as it makes a difference - thank you.

isaacl:
How do you feel about the using pointers (`package unsafe`)? (btw I can't join slack due to #5887)

ncw:
> How do you feel about the using pointers (`package unsafe`)?

I'd prefer not to use unsafe. I don't think it will make stuff go a lot faster... though I could be proved wrong by benchmarks!

PR #5894, opened 2022-01-02
docs/jottacloud: add note about upload only being supported on jotta device

albertony:

Some minor documentation improvements:
- Add note that if configuring any other devices than the default, and Jottacloud built-in, "Jotta" device then upload is not supported - it will throw error `Failed to copy: upload not supported for devices other than Jotta`
- Link to newly created [--temp-dir](https://rclone.org/docs/#temp-dir-dir) documentation, which correctly describes defaults for all operating systems and also mentions the new `rclone config paths` command. Previously only mentioned `TMPDIR`, which is not correct on Windows.

#### Checklist

### DIFF
--- a/docs/content/jottacloud.md
+++ b/docs/content/jottacloud.md
@@ -148,9 +148,11 @@ To copy a local directory to an Jottacloud directory called backup
 The official Jottacloud client registers a device for each computer you install it on,
 and then creates a mountpoint for each folder you select for Backup.
 The web interface uses a special device called Jotta for the Archive and Sync mountpoints.
-In most cases you'll want to use the Jotta/Archive device/mountpoint, however if you want to access
-files uploaded by any of the official clients rclone provides the option to select other devices
-and mountpoints during config.
+
+With rclone you'll want to use the Jotta/Archive device/mountpoint in most cases, however if you
+want to access files uploaded by any of the official clients rclone provides the option to select
+other devices and mountpoints during config. Note that uploading files is currently not supported
+to other devices than Jotta.
 
 The built-in Jotta device may also contain several other mountpoints, such as: Latest, Links, Shared and Trash.
 These are special mountpoints with a different internal representation than the "regular" mountpoints.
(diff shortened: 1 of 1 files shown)


docs/content/jottacloud.md
```diff
@@ -148,9 +148,11 @@ To copy a local directory to an Jottacloud directory called backup
...
 and then creates a mountpoint for each folder you select for Backup.
 The web interface uses a special device called Jotta for the Archive and Sync mountpoints.
-In most cases you'll want to use the Jotta/Archive device/mountpoint, however if you want to access
-files uploaded by any of the official clients rclone provides the option to select other devices
-and mountpoints during config.
+
+With rclone you'll want to use the Jotta/Archive device/mountpoint in most cases, however if you
+want to access files uploaded by any of the official clients rclone provides the option to select
+other devices and mountpoints during config. Note that uploading files is currently not supported
+to other devices than Jotta.
```
ncw, line 155: What happens if you upload to another device other than Jotta?

ncw (approved):
Happy for you to merge this when you are done with it :-)

docs/content/jottacloud.md
albertony: I was actually a bit surprised that it was blocked in rclone, because when I did other experimenting with Jottacloud API years ago there was no difference. But my guess is that the newer "allocate API" for doing the uploading, introduced here https://github.com/rclone/rclone/pull/2498, did not support it (or at least that is known).

ncw: You could see if the integration tests pass on a non standard device. If they do they we are good!

albertony: I can test it out.. Unless @buengese manages to stop me with a good reason not to... :)

albertony: Found out how we can use the allocate API with other devices than "Jotta", and that seems to work very well: https://github.com/rclone/rclone/pull/5926 (Work in progress)

PR #5911, opened 2022-01-12
Remove "After" in automount example

maxxer:
According to [systemd.automount](https://www.freedesktop.org/software/systemd/man/systemd.automount.html) manual

> Note that automount units are separate from the mount itself, so you should 
> not set After= or Requires= for mount dependencies here. 
> For example, you should not set After=network-online.target or 
> similar on network filesystems. Doing so may result in an ordering cycle.

Small change in mount documentation 

#### Checklist

### DIFF
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -354,7 +354,6 @@ optionally accompanied by systemd automount unit
 ```
 # /etc/systemd/system/mnt-data.automount
 [Unit]
-After=network-online.target
 Before=remote-fs.target
 [Automount]
 Where=/mnt/data


albertony:
Closes #6459

ncw (approved):
Thank you for finding and fixing this - much appreciated :-)

ncw:
@albertony can you merge this once we've had the OK back from @wilcooo 's tests - thanks!

wilcooo (changes requested):
I suspect the `After` shoulld be removed from the `mnt-data.mount` file as well, as I kept getting dependency cycles even after removing the line from `mnt-data.automount`. See my issue #6459 for more details.

yuraih:
This pull request is a duplicate of https://github.com/rclone/rclone/pull/5818 that was reported a year ago.

PR #5921, opened 2022-01-16
docs: improve guide for installing from source

### LINKED ISSUE #5809: Doc fix: install instructions specify go 1.14 but modules depend on at least 1.15
From the install instructions at https://rclone.org/install/
Go version 1.14 is required to build from source.  However dependency requirements require at least 1.17:

```
# github.com/rclone/rclone/fs
fs/versioncheck.go:8:15: undefined: Go_version_1_15_required_for_compilation
note: module requires Go 1.17
# github.com/golang-jwt/jwt/v4
../../../../pkg/mod/github.com/golang-jwt/jwt/v4@v4.1.0/ecdsa.go:135:4: r.FillBytes undefined (type *big.Int has no field or method FillBytes)
../../../../pkg/mod/github.com/golang-jwt/jwt/v4@v4.1.0/ecdsa.go:136:4: s.FillBytes undefined (type *big.Int has no field or method FillBytes)
note: module requires Go 1.15
# golang.org/x/net/http2
../../../../pkg/mod/golang.org/x/net@v0.0.0-20211109214657-ef0fda0de508/http2/transport.go:417:45: undefined: os.ErrDeadlineExceeded
note: module requires Go 1.17
```

albertony:

Main purpose is to updated according to Go version 1.17: Use `go install` instead of `go get`.

Included additional information regarding cmount buildtag, required winfsp installation on windows, and "optimized builds" with `-ldflags -s`.

Maybe the text should be split into subsections now? E.g. the "git + go build" method vs the "go install/go get" method? (Not sure what to name the headers then)

Fixes #5809

#### Checklist

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -3,11 +3,11 @@ title: "Install"
 description: "Rclone Installation"
 ---
 
-# Install #
+# Install
 
 Rclone is a Go program and comes as a single binary file.
 
-## Quickstart ##
+## Quickstart
 
   * [Download](/downloads/) the relevant binary.
   * Extract the `rclone` executable, `rclone.exe` on Windows, from the archive.
@@ -22,7 +22,7 @@ run `rclone -h`.
 Already installed rclone can be easily updated to the latest version
 using the [rclone selfupdate](/commands/rclone_selfupdate/) command.
 
-## Script installation ##
+## Script installation
 
 To install rclone on Linux/macOS/BSD systems, run:
 
@@ -35,7 +35,7 @@ For beta installation, run:
 Note that this script checks the version of rclone installed first and
 won't re-download if not needed.
 
-## Linux installation from precompiled binary ##
+## Linux installation from precompiled binary
 
 Fetch and unpack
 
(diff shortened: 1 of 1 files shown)


ncw (commented):
Sorry seem to have lost this PR!

Can you rebase it and update it (go1.18 is now current and go1.16 is the minimum) and I will merge - thank you, and sorry for the delay.

ncw:
> Sure. And also; I think it should get some updates due to your recent cgofuse/cmount changes. E.g. the following, about building with "-tags cmount":
> 
> > This assumes you have a GCC compatible C compiler (GCC or Clang) in your PATH. On Windows you can use the MinGW port of GCC,
> 
> and:
> 
> > and set environment variable CPATH pointing to the fuse include directory within the WinFsp installation.
> 
> Should probably be updated to mention that on Windows one can instead build with `SET CGO_ENABLED=0`, and then avoid GCC, MinGW, CPATH etc, and this is how the official release version is built - starting with the upcoming 1.59 release(?)

Yes that is definitely worth mentioning.

I should probably make `cmount` build by default on Windows without the `-tags cmount` just like normal `cmd/mount` does. That would hurt my brain with the build constraints though!

albertony:
Rebased and updated with information about Windows build with cmount but without cgo. Also added a note about `-trimpath`, resulting in the following as the "close to official" build command:
```
go build -trimpath -ldflags -s -tags cmount
```

ncw (approved):
LGTM :-) I'll merge now.

PR #5925, opened 2022-01-18
Make bash completion respect --config.

coolwanglu:
The bash completion script may add remotes to the candidates, by running
`rclone listremotes`. However it does not make sense if the current
command is `rclone --config MY_CONFIG_FILE`.

With this commit, the bash completion script will try to extract the
config file, then run `rclone --cofnig MY_CONFIG_FILE listremotes`
instead.

See above (commit message)

No

#### Checklist

### DIFF
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -48,11 +48,22 @@ __rclone_custom_func() {
         else
             __rclone_init_completion -n : || return
         fi
-	local rclone=(command rclone --ask-password=false)
+        local rclone=(command rclone --ask-password=false)
         if [[ $cur != *:* ]]; then
+            local config_file=""
+            for idx in ${!words[@]}; do
+              if [[ "${words[idx]}" == "--config" && "${#words[@]}" -gt $((idx+1)) ]]; then
+                config_file="${words[idx+1]}"
+              fi
+            done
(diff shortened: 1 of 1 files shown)


coolwanglu:
To test this PR:

0. Generate and install the bash completoin files, as described in [the doc](https://rclone.org/commands/rclone_genautocomplete_bash/).
1. Set up two config files, say CONFIG_A and CONFIG_B. One of them may be the default one in ~/.config/rclone/. Make sure that the contents are different
2. Type in bash `rclone --config CONFIG_A copy <TAB><TAB>`, omit `--config CONFIG_A` if it's the default path. You shoud be able to see remotes in CONFIG_A as candidates
3. DO NOT PRESS CTRL-C (will explain later), press `backspace` to remove the characters, then type `rclone --config CONFIG_B copy <TAB><TAB>`, then you should see remotes in CONFIG_B as candidates.

About ctrl-c in step #3, it is to make sure that script always check the current command/flags, instead of caching something.

ncw:
The bash completion has been reworked since this PR so the PR is now quite out of date.

I'm going to close this.

Thank you for submitting it and sorry we never merged it.

Issue #3625, opened 2019-10-13
Add gdrive file id support


ShixAJ:
Hello,
Adding gdrive file id support to rclone would be great.
At the moment there is only support for gdrive folders id's using "root-folder-id".
I guess a similar feature but for files would be great.

ncw:
What I was imagining was that you'd supply `--drive-file-id` and then the remote created would be a remote which just points to that one file.

This would mean you could copy it or whatever you wanted.

Does that sound OK?

ShixAJ:
Yeah I guess. So rclone copy —drive-file-id xxxxxxxxxxxxxxxx destination: ?

ShixAJ:
Also now thinking about it. Have some other ideas. At the moment I have a python script that extracts folder-id's from links in a text file, changes the remote folder id in the config(making it do that because don't know how to use the --drive-root-folder-id flag hehe) and does Rclone copy using that remote as source.
So thinking maybe we could let Rclone take a txt file that has Gdrive ids as input (would need a flag) and copy them all to the destination remote. Not essential to implement since I could just do a small script for it. 
But knowing that I want to see if its possible for Rclone to distinguish between folder-ids and file-id's from the id itself and maybe make a third flag for that. Some gdrive links do say "folder" or "file" inside the URL which I could make the script change the flag depending on them, but some Gdrive links are like "/open?id=xxxxxxxxxxxxx" which will make it impossible for a script that just reads links from a text file to now which flag to use.

fionera:
I just started to work on this. But as I see it we have to make the drive readonly when a file id is given, right?

ncw:
@fionera 
> I just started to work on this. 

Great :-)

> But as I see it we have to make the drive readonly when a file id is given, right?

Do you think so?  I'm assuming that we return ErrIsFile and return the parent directory.

ncw:
> Also now thinking about it. Have some other ideas. At the moment I have a python script that extracts folder-id's from links in a text file, changes the remote folder id in the config(making it do that because don't know how to use the `--drive-root-folder-id` flag hehe) and does Rclone copy using that remote as source.
> So thinking maybe we could let Rclone take a txt file that has Gdrive ids as input (would need a flag) and copy them all to the destination remote. Not essential to implement since I could just do a small script for it.

It is getting a bit specialized but maybe :-)

[...]

PR #5940, opened 2022-01-23
rc: Allow user to disable authentication for web gui

negative0:

As per https://github.com/rclone/rclone-webui-react/issues/139, if the user has secured their server using some alternate mechanism, we should allow them to bypass the authentication in rclone by using an existing flag `--rc-no-auth`.

No

https://github.com/rclone/rclone-webui-react/issues/139

#### Checklist

### DIFF
--- a/fs/rc/rcserver/rcserver.go
+++ b/fs/rc/rcserver/rcserver.go
@@ -95,20 +95,20 @@ func newServer(ctx context.Context, opt *rc.Options, mux *http.ServeMux) *Server
 			fs.Errorf(nil, "Error while fetching the latest release of Web GUI: %v", err)
 		}
 		if opt.NoAuth {
-			opt.NoAuth = false
-			fs.Infof(nil, "Cannot run Web GUI without authentication, using default auth")
-		}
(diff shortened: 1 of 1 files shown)


albertony (approved):

ncw (approved):
This looks sensible to me.

The user has explicitly allowed it with `--rc-no-auth` so has acknowledged that this is a security disaster!

skyler14:
What parts of rclone can you use without authentication? Does this for example allowing pulling public google drive files by gid as mentioned in [in this Issue](https://github.com/rclone/rclone/issues/3625), or folders in a bucket with publicly accessibly folders? Want to use this in rclone-py with user credentials as an optional feature since already in the gdrive case alone it would be super helpful. 

But if it extends to public buckets or like a bucket link that has built in access controls/timers that would be _very handy._

ncw:
@skyler14 talking specifically about the rc - all the methods which require authentication are marked with `Authentication is required for this call.` in [the docs](https://rclone.org/rc/) so you can do anything else without authentication.

PR #5944, opened 2022-01-24
docs: add label before options spec list

albertony:

See this section: https://rclone.org/http/#http-no-head

The help description for this flag ends with a list, and when the generic help template appends default list items (Config, Env Var, Type etc) after this help text it will be automatically joined into the same list, which is not the intention. I've rewritten that particular text in another pr, but in general I think it is better to extend the generic help template to include a small "intro" before the list items, to make sure it is separated from the rest of the help text.

~I could not come up with a better name than "Specification"... Can you??~

EDIT:
- Added header "Properties:" as per  [comment](#issuecomment-1021375128").
- Added Provider as an additional entry, if set. 
   - s3 docs have a lot of variants of same options with different provider filters, to set different multiple choice lists for the providers. The Provider filter was not included in the docs, which lead to a lot of seemingly duplicate options.
- Added Required as an additional entry
    - Only showing it when Default is not set (empty), and then not showing Default. I.e. only showing one of them: Default if set (non-empty), else Required (true/false). The reality is that Required means option must have a value, but if there is a default then a value does not have to be explicitely set and therefore Required makes no real difference.
- Some examples showing the above:
    ```
    Properties:

    - Config:      requester_pays
    - Env Var:     RCLONE_S3_REQUESTER_PAYS
    - Provider:    AWS
    - Type:        bool
    - Default:     false
    ```

    ```
    Properties:

    - Config:      shared_credentials_file
    - Env Var:     RCLONE_S3_SHARED_CREDENTIALS_FILE
    - Type:        string
    - Required:    false
    ```

    ```
    Properties:

    - Config:      location_constraint
    - Env Var:     RCLONE_S3_LOCATION_CONSTRAINT
    - Provider:    IBMCOS
    - Type:        string
    - Required:    false
    - Examples:
        - "us-standard"
            - US Cross Region Standard
        - "us-vault"
            - US Cross Region Vault
        - "us-cold"
    ```
 - Renamed "Examples" to "Choices" when the Exclusive property is set, since the list is then the possible values and not what I would think of as "examples". Considered adding more alternative wordings here (like "Must be one of", "Must be one of, if set" or something like that) depending on the Required and Default properties, which also play a part related to decide if empty value is allowed or not, but stuck with the two.
     - Required=true && Default=="" means empty is not allowed, so if Exclusive=True it means the value must be strictly from the list:
         ```
         - Required:     true
         - Choices:
             - "must"
             - "choose"
         ```
        Else, if Exclusive=False it means it must be set to something - i.e. not empty, but not necessarily from the list:
         ```
         - Required:     true
         - Examples:
             - "may"
             - "choose"
             - "or write something else"
         ```
     - (Required=false || Default!="") means empty value is allowed, but if Exclusive=True and a non-empty value is specified it must still be strictly from the list:
         ```
         - Default:     "some"
         - Choices:
             - "may"
             - "choose"
             - "or stick with default"
         ```
        Else, if Exclusive=False it can be anything - i.e. empty input allowed, as it will mean default:
         ```
         - Default:     "some"
         - Examples:
             - "may"
             - "choose"
             - "or write something else"
             - "or stick with default"
         ```
- Ref: https://github.com/rclone/rclone/pull/5551

#### Checklist

### DIFF
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -329,12 +329,29 @@ func showBackend(name string) {
 			if opt.IsPassword {
 				fmt.Printf("**NB** Input to this must be obscured - see [rclone obscure](/commands/rclone_obscure/).\n\n")
 			}
+			fmt.Printf("Properties:\n\n")
 			fmt.Printf("- Config:      %s\n", opt.Name)
 			fmt.Printf("- Env Var:     %s\n", opt.EnvVarName(backend.Prefix))
+			if opt.Provider != "" {
+				fmt.Printf("- Provider:    %s\n", opt.Provider)
+			}
 			fmt.Printf("- Type:        %s\n", opt.Type())
-			fmt.Printf("- Default:     %s\n", quoteString(opt.GetValue()))
+			defaultValue := opt.GetValue()
+			// Default value and Required are related: Required means option must
+			// have a value, but if there is a default then a value does not have
+			// to be explicitely set and then Required makes no difference.
+			if defaultValue != "" {
+				fmt.Printf("- Default:     %s\n", quoteString(defaultValue))
+			} else {
+				fmt.Printf("- Required:    %v\n", opt.Required)
+			}
+			// List examples / possible choices
 			if len(opt.Examples) > 0 {
-				fmt.Printf("- Examples:\n")
+				if opt.Exclusive {
+					fmt.Printf("- Choices:\n")
+				} else {
+					fmt.Printf("- Examples:\n")
+				}
 				for _, ex := range opt.Examples {
 					fmt.Printf("    - %s\n", quoteString(ex.Value))
 					for _, line := range strings.Split(ex.Help, "\n") {


ncw:
Well spotted.

I think "Properties" is probably slightly clearer than "Specification".

The markdown is this

```
If you set this option, rclone will not do the HEAD request.  This will mean

- directory listings are much quicker
- rclone won't have the times or sizes of any files
- some files that don't exist may be in the listing


- Config:      no_head
- Env Var:     RCLONE_HTTP_NO_HEAD
- Type:        bool
- Default:     false

```

Which produces HTML which looks like this with weird double spacing for some reason...

image

What do you think about making the info a table instead?

```
| Property    | Value       |
| :---------- | :---------- |
| Config      | no_head |
| Env Var     | RCLONE_HTTP_NO_HEAD |
| Type        | bool |
| Default     | false |
```

| Property   | Value       |
| :---------- | :---------- |
| Config      | no_head |
| Env Var     | RCLONE_HTTP_NO_HEAD |
| Type        | bool |
| Default     | false |

It looks nice in the HTML but those things should look good in the terminal if possible so a bit of care formatting the spaces might be needed if we decide to do that.

Possibly just a paragraph might be best as it is a lot of quite unrelated things in the list/table and that gives a chance to put a little bit of explanation which goes a long way!

```
This should be set using the `no_head` key in the config file or using `RCLONE_HTTP_NO_HEAD` if setting with an environment variable. It's type is `bool` with a default of `false`.
```

This should be set using the `no_head` key in the config file or using `RCLONE_HTTP_NO_HEAD` if setting with an environment variable. It's type is `bool` with a default of `false`.

What do you think?

albertony:
> I think "Properties" is probably slightly clearer than "Specification".

Yes - much better!

> What do you think about making the info a table instead?

> Possibly just a paragraph..

Using a table could look really nice - in html, at least! But there are some that may be awkward to fit, in table but also as a paragraph-only...

```
#### --ftp-encoding

The encoding for the backend.

See the [encoding section in the overview](/overview/#encoding) for more info.

- Config:      encoding
- Env Var:     RCLONE_FTP_ENCODING
- Type:        MultiEncoder
- Default:     Slash,Del,Ctl,RightSpace,Dot
- Examples:
    - "Asterisk,Ctl,Dot,Slash"
        - ProFTPd can't handle '*' in file names
    - "BackSlash,Ctl,Del,Dot,RightSpace,Slash,SquareBracket"
        - PureFTPd can't handle '[]' or '*' in file names
    - "Ctl,LeftPeriod,Slash"
        - VsFTPd can't handle file names starting with dot
```

```
#### --s3-bucket-acl

Canned ACL used when creating buckets.

For more info visit https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl

Note that this ACL is applied when only when creating buckets.  If it
isn't set then "acl" is used instead.

- Config:      bucket_acl
- Env Var:     RCLONE_S3_BUCKET_ACL
- Type:        string
- Default:     ""
- Examples:
    - "private"
        - Owner gets FULL_CONTROL.
        - No one else has access rights (default).
    - "public-read"
        - Owner gets FULL_CONTROL.
        - The AllUsers group gets READ access.
    - "public-read-write"
        - Owner gets FULL_CONTROL.
        - The AllUsers group gets READ and WRITE access.
        - Granting this on a bucket is generally not recommended.
    - "authenticated-read"
        - Owner gets FULL_CONTROL.
        - The AuthenticatedUsers group gets READ access.
```

Though, the problem here is the examples, which could remain a lists - and that would be easier to combine with paragraph than table.

Speaking of the "Examples": It can be misleading. Ref my previous PR in config: Depending on Exclusive, Required, Default it could be considered to be "examples" because you are allowed to enter anything else as well, or (for pure exclusive) it is more correct to callit "choices" or "options" than "examples" because you have to pick one of them. Maybe that will get a bit long for a generated paragraph?

I'm really not sure, but maybe I should give it an effort to continue on the existing approach? Rename to "Properties", and add some more/better information about the "examples", and we can see how that works?

albertony:
Pushed a late night update, with a quick (but long) edit of the initial pr description. Will take a fresh look at it tomorrow.

ncw (approved):
I think this is a nice change and should make the docs clearer.

Thank you :-)

PR #5953, opened 2022-01-26
touch: fix issue where directory is created instead of file

### LINKED ISSUE #5952: Rclone “touch” command fails since 1.57.0
https://forum.rclone.org/t/rclone-touch-command-fails-since-1-57-0/28890

Something changed with rclone 1.57.0 so that the rclone "touch" command is not usable for me anymore.
It all worked fine, and still does when I run rclone with 1.56.2, but it fails with newer versions.
I see that the --recursive option has been added to the "touch" command with 1.57.0, so maybe it is related to this addition?

Background: I use the "touch" command to test if the remote server is set up correctly. So I "touch" a new file that does not exist on the remote server and delete the file after success.

The errors I get with type FTP:
```
2022/01/26 12:47:06 ERROR : Attempt 1/1 failed with 1 errors and: failed to touch (create): update stor: 550 /files/ftp: Not a regular file
2022/01/26 12:47:06 Failed to touch: failed to touch (create): update stor: 550 /files/ftp: Not a regular file
```

The errors I get with type SFTP:
```
2022/01/26 13:01:26 ERROR : Attempt 1/1 failed with 1 errors and: failed to touch (create): Update Create failed: sftp: "Failure" (SSH_FX_FAILURE)
2022/01/26 13:01:26 Failed to touch: failed to touch (create): Update Create failed: sftp: "Failure" (SSH_FX_FAILURE)
```

```
rclone v1.57.0
- os/kernel: 5.4.0-94-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17.2
- go/linking: static
- go/tags: none
```

I have also tested the latest Beta, but without success:

```
rclone v1.58.0-beta.5960.a667e03fc
- os/kernel: 5.4.0-94-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17.6
- go/linking: static
- go/tags: none
```

Ubuntu 20.04 - 64 bit

I have tested FTP, SFTP, Dropbox -> All report the same: "failed to touch"

With FTP:
```
rclone --config '/tmp/rclone-ftp.config' --retries 1 touch 'rclone-storage:/files/ftp'
```
-> Result: Nothing happens on the remote side. An rclone error message is shown (see above).

With SFTP:
```
rclone --config '/tmp/rclone-sftp.config' --retries 1 touch 'rclone-storage:/files/sftp'
```
-> Result: On the remote side, a new directory "sftp" is created in "/files/" (??? Why a directory ???). An rclone error message is shown (see above).

#### The rclone config contents with secrets removed.

```
[rclone-storage]
type = ftp
host = dev23.keymachine.de
user = test
pass = <secret-pass>
explicit_tls = true
```

```
[rclone-storage]
type = sftp
host = dev23.keymachine.de
user = test
pass = <secret-pass>
md5sum_command = none
sha1sum_command = none
```

#### A log from the command with the `-vv` flag

With FTP:
```
2022/01/26 12:47:28 DEBUG : rclone: Version "v1.57.0" starting with parameters ["rclone" "--config" "/tmp/rclone-ftp.config" "-vv" "--retries" "1" "touch" "rclone-storage:/files/ftp"]
2022/01/26 12:47:28 DEBUG : Creating backend with remote "rclone-storage:/files/ftp"
2022/01/26 12:47:28 DEBUG : Using config file from "/tmp/rclone-ftp.config"
2022/01/26 12:47:28 DEBUG : ftp://dev23.keymachine.de:21/files/ftp: Connecting to FTP server
2022/01/26 12:47:28 DEBUG : Touch time 2022-01-26 12:47:28.42150678 +0100 CET m=+0.361783329
2022/01/26 12:47:28 DEBUG : ftp://dev23.keymachine.de:21/files/ftp: Touching (creating)
2022/01/26 12:47:29 DEBUG : ftp://dev23.keymachine.de:21/files/ftp: Connecting to FTP server
2022/01/26 12:47:29 DEBUG : : Removed after failed upload: 550 /files/ftp: Not a regular file
2022/01/26 12:47:29 ERROR : Attempt 1/1 failed with 1 errors and: failed to touch (create): update stor: 550 /files/ftp: Not a regular file
2022/01/26 12:47:29 DEBUG : 2 go routines active
2022/01/26 12:47:29 Failed to touch: failed to touch (create): update stor: 550 /files/ftp: Not a regular file
```

With SFTP:
```
2022/01/26 13:04:07 DEBUG : rclone: Version "v1.57.0" starting with parameters ["rclone" "--config" "/tmp/rclone-sftp.config" "-vv" "--retries" "1" "touch" "rclone-storage:/files/sftp"]
2022/01/26 13:04:07 DEBUG : Creating backend with remote "rclone-storage:/files/sftp"
2022/01/26 13:04:07 DEBUG : Using config file from "/tmp/rclone-sftp.config"
2022/01/26 13:04:07 DEBUG : sftp://test@dev23.keymachine.de:22//files/sftp: New connection 62.141.56.104:51682->62.141.56.45:22 to "SSH-2.0-OpenSSH_8.4p1 Debian-5"
2022/01/26 13:04:07 DEBUG : Touch time 2022-01-26 13:04:07.852656441 +0100 CET m=+0.612075729
2022/01/26 13:04:07 DEBUG : sftp://test@dev23.keymachine.de:22//files/sftp: Touching (creating)
2022/01/26 13:04:07 ERROR : Attempt 1/1 failed with 1 errors and: failed to touch (create): Update Create failed: sftp: "Failure" (SSH_FX_FAILURE)
2022/01/26 13:04:07 DEBUG : 11 go routines active
2022/01/26 13:04:07 Failed to touch: failed to touch (create): Update Create failed: sftp: "Failure" (SSH_FX_FAILURE)
```

#### How to use GitHub

Thank you very much for your help!
Alex

albertony:

Fixes #5952

#### Checklist

### DIFF
--- a/cmd/touch/touch.go
+++ b/cmd/touch/touch.go
@@ -5,11 +5,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"time"
 
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/fs/fspath"
 	"github.com/rclone/rclone/fs/object"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/spf13/cobra"
@@ -63,11 +65,31 @@ then add the ` + "`--localtime`" + ` flag.
 `,
 	Run: func(command *cobra.Command, args []string) {
 		cmd.CheckArgs(1, 1, command, args)
-		f, fileName := cmd.NewFsFile(args[0])
+		f, remote := newFsDst(args)
 		cmd.Run(true, false, command, func() error {
-			return Touch(context.Background(), f, fileName)
+			return Touch(context.Background(), f, remote)
 		})
 	},
+}
+
+// newFsDst creates a new dst fs from the arguments.
+//
+// The returned fs will never point to a file. It will
+// point to the parent directory of specified path,
+// is returned together with the name of file or directory,
+// except if argument is only a remote name.
+// Similar to cmd.NewFsDstFile, but without raising fatal
+// when name of file or directory is empty.
+func newFsDst(args []string) (f fs.Fs, remote string) {
+	root, remote, err := fspath.Split(args[0])
+	if err != nil {
+		log.Fatalf("Parsing %q failed: %v", args[0], err)
+	}
+	if root == "" {
+		root = "."
+	}
+	f = cmd.NewFsDir([]string{root})
+	return f, remote
 }
 
 // parseTimeArgument parses a timestamp string according to specific layouts
@@ -107,47 +129,51 @@ func createEmptyObject(ctx context.Context, remote string, modTime time.Time, f 
 }
 
 // Touch create new file or change file modification time.
-func Touch(ctx context.Context, f fs.Fs, fileName string) error {
+func Touch(ctx context.Context, f fs.Fs, remote string) error {
 	t, err := timeOfTouch()
 	if err != nil {
 		return err
 	}
 	fs.Debugf(nil, "Touch time %v", t)
-	file, err := f.NewObject(ctx, fileName)
+	file, err := f.NewObject(ctx, remote)
 	if err != nil {
 		if errors.Is(err, fs.ErrorObjectNotFound) {
-			// Touch single non-existent file
+			// Touching non-existant path, possibly creating it as new file
+			if remote == "" {
+				fs.Logf(f, "Not touching empty directory")
+				return nil
+			}
 			if notCreateNewFile {
 				fs.Logf(f, "Not touching non-existent file due to --no-create")
 				return nil
 			}
 			if recursive {
+				// For consistency, --recursive never creates new files.
 				fs.Logf(f, "Not touching non-existent file due to --recursive")
 				return nil
 			}
 			if operations.SkipDestructive(ctx, f, "touch (create)") {
 				return nil
 			}
-			fs.Debugf(f, "Touching (creating)")
-			if err = createEmptyObject(ctx, fileName, t, f); err != nil {
+			fs.Debugf(f, "Touching (creating) %q", remote)
+			if err = createEmptyObject(ctx, remote, t, f); err != nil {
 				return fmt.Errorf("failed to touch (create): %w", err)
 			}
 		}
 		if errors.Is(err, fs.ErrorIsDir) {
+			// Touching existing directory
 			if recursive {
-				// Touch existing directory, recursive
-				fs.Debugf(nil, "Touching files in directory recursively")
-				return operations.TouchDir(ctx, f, t, true)
+				fs.Debugf(f, "Touching recursively files in directory %q", remote)
+				return operations.TouchDir(ctx, f, remote, t, true)
 			}
-			// Touch existing directory without recursing
-			fs.Debugf(nil, "Touching files in directory non-recursively")
-			return operations.TouchDir(ctx, f, t, false)
+			fs.Debugf(f, "Touching non-recursively files in directory %q", remote)
+			return operations.TouchDir(ctx, f, remote, t, false)
 		}
 		return err
 	}
 	// Touch single existing file
-	if !operations.SkipDestructive(ctx, fileName, "touch") {
-		fs.Debugf(f, "Touching %q", fileName)
+	if !operations.SkipDestructive(ctx, remote, "touch") {
+		fs.Debugf(f, "Touching %q", remote)
 		err = file.SetModTime(ctx, t)
 		if err != nil {
 			return fmt.Errorf("failed to touch: %w", err)
--- a/cmd/touch/touch_test.go
+++ b/cmd/touch/touch_test.go
@@ -113,6 +113,15 @@ func TestTouchCreateMultipleDirAndFile(t *testing.T) {
 	fstest.CheckListingWithPrecision(t, r.Fremote, []fstest.Item{file1}, []string{"a", "a/b"}, fs.ModTimeNotSupported)
 }
 
+func TestTouchEmptyName(t *testing.T) {
+	r := fstest.NewRun(t)
+	defer r.Finalise()
+
+	err := Touch(context.Background(), r.Fremote, "")
+	require.NoError(t, err)
+	fstest.CheckListingWithPrecision(t, r.Fremote, []fstest.Item{}, []string{}, fs.ModTimeNotSupported)
+}
+
 func TestTouchEmptyDir(t *testing.T) {
 	r := fstest.NewRun(t)
 	defer r.Finalise()
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -1930,8 +1930,8 @@ func SetTier(ctx context.Context, fsrc fs.Fs, tier string) error {
 }
 
 // TouchDir touches every file in f with time t
-func TouchDir(ctx context.Context, f fs.Fs, t time.Time, recursive bool) error {
-	return walk.ListR(ctx, f, "", false, ConfigMaxDepth(ctx, recursive), walk.ListObjects, func(entries fs.DirEntries) error {
+func TouchDir(ctx context.Context, f fs.Fs, remote string, t time.Time, recursive bool) error {
+	return walk.ListR(ctx, f, remote, false, ConfigMaxDepth(ctx, recursive), walk.ListObjects, func(entries fs.DirEntries) error {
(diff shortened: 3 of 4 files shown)


backend/ftp/ftp.go
```diff
@@ -868,11 +868,15 @@ func (f *Fs) mkdir(ctx context.Context, abspath string) error {
...
 }
 
-// mkParentDir makes the parent of remote if necessary and any
-// directories above that
-func (f *Fs) mkParentDir(ctx context.Context, remote string) error {
-	parent := path.Dir(remote)
-	return f.mkdir(ctx, path.Join(f.root, parent))
+// mkParentDir makes the parent of the native path dirPath if
+// necessary and any directories above that
+func (f *Fs) mkParentDir(ctx context.Context, dirPath string) (err error) {
```
ncw, line 873: Can we fix this without making `mkParentDir` take an absolute path. In my experience its best to keep the functions which take absolute paths down to a minimum.

backend/sftp/sftp.go
```diff
@@ -972,11 +972,15 @@ func (f *Fs) PutStream(ctx context.Context, in io.Reader, src fs.ObjectInfo, opt
...
 }
 
-// mkParentDir makes the parent of remote if necessary and any
-// directories above that
-func (f *Fs) mkParentDir(ctx context.Context, remote string) error {
-	parent := path.Dir(remote)
-	return f.mkdir(ctx, path.Join(f.absRoot, parent))
+// mkParentDir makes the parent of the native path dirPath if
+// necessary and any directories above that
+func (f *Fs) mkParentDir(ctx context.Context, dirPath string) (err error) {
```
ncw, line 977: Same here...

ncw (commented):
Can you write a test in [fstests.go](https://github.com/rclone/rclone/blob/master/fstest/fstests/fstests.go) which fails for the ftp/sftp/dropbox backends? I think the underlying problem is that the `Put` method is failing to write a single file to the root - is that right? That should be easy enough to test for.

Then can you re-organize these patches so they are one per backend  to fix the test. Then this will never happen again as we'll test it in the integration tests :-) We'll probably find other backends with the same problem would be my guess.

backend/ftp/ftp.go
albertony: Do this internally in `Put` instead, you mean? I did something like that first, but then found that a `mkParentDirs` function exists in a few other backends already (webdav, jottacloud, mailru, yandex) - all of which had absolute path as argument.

albertony:
I guess you meant more like this (pushed an update)? So the mkParentDir takes just the remote as before, but now joins it with root before finding and creating parents?

ncw (commented):
This code is a lot neater :-) I'm not sure it is correct though!

After some time thinking about what exactly is going on I came up with this

## Before

```go
// remote = ""
// f.root = "file.txt" 
parent := path.Dir(remote)
// parent = "."
return f.mkdir(ctx, path.Join(f.root, parent))
// path.Join(f.root, parent) = "file.txt"
// => makes the "file.txt" directory
```

## After

```go
// remote = ""
// f.root = "file.txt" 
parent := path.Dir(path.Join(f.root, remote))
// path.Join(f.root, remote) = file.txt
// parent = "."
if parent == "." {
    parent = ""
}
// parent = ""
return f.mkdir(ctx, parent)
```

## Discussion

The problem stems from the behaviour of opening a backend pointing to a non-existent file let's say `remote:file.txt`

When the file doesn't exist, rclone assumes it is a directory and returns the backend pointing to that directory (`remote:file.txt`), with the root directory as yet uncreated.

When you write a file, the root directory gets created. All good so far.

However what `touch` seems to be doing is writing a file with an empty name using `f.Put("")` to the remote called (`remote:file.txt`).

This seems a bit odd to me and I'm not suprised it doesn't work on some backends! There certainly aren't integration tests for it. If we want it to work they we definitely need an integration test for it.

Wouldn't it be better for touch to parse the incoming path itself when --recursive isn't set using `fspath.Split`? Then it could be just creating the backend `remote:` and creating the file `file.txt` in it which doesn't seem problematic!

What do you think?

albertony:
> However what `touch` seems to be doing is writing a file with an empty name using `f.Put("")` to the remote called (`remote:file.txt`).

Correct. And I agree it seems a bit strange, when you put it that way...

> Wouldn't it be better for touch to parse the incoming path itself when --recursive isn't set using `fspath.Split`? Then it could be just creating the backend `remote:` and creating the file `file.txt` in it which doesn't seem problematic!

Yes, probably! I will look into it. Thanks.

albertony:
Yes, the NewFsDstFile is what was used before the recursive touch implementation.

But, this is only testing if --recursive is present or not, right? So, if I were to `rclone touch remote:temp.txt --recursive` for some strange reason, then on ftp and sftp it would still hit the same issue - strange (for users) error message and side-effect of creating a directory with name `temp.txt`..?

ncw:
> Yes, the NewFsDstFile is what was used before the recursive touch implementation.
> 
> But, this is only testing if --recursive is present or not, right? So, if I were to `rclone touch remote:temp.txt --recursive` for some strange reason, then on ftp and sftp it would still hit the same issue - strange (for users) error message and side-effect of creating a directory with name `temp.txt`..?

Hmm, that is a tricky one.

I think the best solution is that we document that `--recursive` needs a directory.

I've written about 10 different answers to this and deleted them, but I decided to try it and the answer looks sensible and doesn't create unexpected directories.

```
$ rclone touch -vv --recursive --retries 1 :sftp,host=localhost,user=sftptest:doesnotexist11.txt
2022/01/27 17:17:49 NOTICE: sftp://sftptest@localhost:22/doesnotexist11.txt: Not touching non-existent file due to --recursive
```

```
$ rclone touch -vv --recursive --retries 1 :sftp,host=localhost,user=sftptest:normal-file.txt
2022/01/27 17:20:05 DEBUG : Touch time 2022-01-27 17:20:05.025510659 +0000 GMT m=+0.446746561
2022/01/27 17:20:05 DEBUG : sftp://sftptest@localhost:22/normal-file.txt: Touching "normal-file.txt"
```

Which looks sensible

albertony:
> Not touching non-existent file due to --recursive

I see, it gets caught by that check! Thats good, then.

I'm testing your code change myself now, on top of clean master...

Found a different issue: When touching an existing directory without --recursive it touches the files in the parent. E.g. with:

```
.\rclone touch local_sftp:parent/existing_folder -vv
```

It logs:
> DEBUG : Touching files in directory non-recursively

And then "Touching" on all files in `parent`.

If appending a trailing slash:

```
.\rclone touch local_sftp:parent/existing_folder/ -vv
```

It does almost nothing, just exists with an (error??) message:

> 2022/01/27 18:59:42 DEBUG : rclone: Version "v1.58.0-DEV" starting with parameters [".\\rclone" "touch" "local_sftp:parent/existing_directory/" "-vv"]
> 2022/01/27 18:59:42 "local_sftp:parent/existing_directory/" is a directory

albertony:
Pushed a new version - with your fix and the additional change I suggested above.

Unit test runs ok, integration test runs ok on sftp remote. Manual testing ok so far, but will test some more...

I removed (moved into a different branch) the original ones, they are not needed for touch - and I guess we do not want to include them then?

albertony:
> Edit: Well, there is the mentioned is a directory case - looking into that, at least.

Thats a fatal error from the NewFsDstFile when there is no (potential) filename. E.g. path with ending slash. (I'm sure you know, but I didn't).

Maybe we have to split the path within touch after all? Pushed a version doing this. Need to "review" it myself, and test more - after a break.

albertony:
@ncw: PTAL

The last version seems to work. It handles corner cases like:
- Root does not exist: `rclone touch remote:`
	- Skipped: Not touching empty directory
- Path does not exist and ends with slash: `rclone touch remote:notexists/`
	- Skipped: Not touching empty directory
- Path exist and ends with slash: `rclone touch remote:exists/`
	- Touching all files in the directory.
- Path does not exist: `rclone touch remote:touch.me`
	- if !recursive: Creates as file touch.me. If remote: points to a directory that does not exists it will be created too.
	- if recursive: Skipped: Not touching non-existent file due to --recursive

ncw (approved):
I think this looks good now - well done :-)

It was a bit of a subtle problem wasn't it!

albertony:
> I think this looks good now - well done :-)

Great, thanks!

> It was a bit of a subtle problem wasn't it!

Yes, it was! I guess part of the reason was all the file vs folder + existing vs nonexisting + recursive vs nonrecursive combinations that was hard to fit perfectly in existing fs functions etc.

albertony:
PS: This was merged into master before it was closed.

PR #5958, opened 2022-01-27
docs: include all commands in online help top menu drop-down

albertony:

*This is suggestion, up for discussion (see below).*

This makes it easier for new users to know what commands are available. There have been cases where users did not know about a command because they only looked at the dropdown showing the few most common ones. They would have to go into the "Overview" site to find the rest.

I kept the existing entries as is, added a divider, and then appended all the missing ones. Did not include subcommands (like "config create") that are linked to from parent ("config") docs.

A possible confusion is that the top ones (the existing ones) are ordered according to relevance, and then the ones below the divider are in alphabetic order. This means "copyto" and "copyurl" are next to each other in the lower section, but "rclone copy" is in a different position in the top section.

Up for discussion:

❓ An alternative could be to included all in alphabetic order in the lower section, duplicated the ones in the top section? Perhaps then see if we can reduce the top section, e.g. remove cat, rcat, ncdu, lsd etc..

❓ Another alternative could be to drop the top section with the main commands ordered by relevance, and just show all in alphabetic order? ☑️ I think I'm voting for this one!

PS: I personally have also found it a bit inconsistent that the "Storage Systems" drop down show all, but not the "Commands". Sometimes when I am going to look for a backend doc I mix up the two in my head; I am so used to go into "Overview" of the "Commands" so I also click "Overview" of "Storage Systems" - but there are no links to the backend docs, they are only in the drop down. My point is that some improved consistency between them will be good, although this is in reality a second issue regarding the Overview pages, while what I have covered above is the drop-downs. I considered extended the "Storage Systems > Overview" to include links, but the downside of this is that the markdown, which is already not very minimalistic, gets quite ugly and tables get wide columns...

#### Checklist

### DIFF
--- a/docs/layouts/chrome/navbar.html
+++ b/docs/layouts/chrome/navbar.html
@@ -34,24 +34,14 @@
         </a>
         <div class="dropdown-menu pre-scrollable" aria-labelledby="navbarDropdown">
           <a class="dropdown-item" href="/commands/"><i class="fas fa-map"></i> Overview</a>
-          <div class="dropdown-divider"></div>          
-          <a class="dropdown-item" href="/commands/rclone_config/"><i class="fa fa-book"></i> rclone config</a>
-          <a class="dropdown-item" href="/commands/rclone_copy/"><i class="fa fa-book"></i> rclone copy</a>
-          <a class="dropdown-item" href="/commands/rclone_sync/"><i class="fa fa-book"></i> rclone sync</a>
-          <a class="dropdown-item" href="/commands/rclone_move/"><i class="fa fa-book"></i> rclone move</a>
-          <a class="dropdown-item" href="/commands/rclone_purge/"><i class="fa fa-book"></i> rclone purge</a>
-          <a class="dropdown-item" href="/commands/rclone_mkdir/"><i class="fa fa-book"></i> rclone mkdir</a>
-          <a class="dropdown-item" href="/commands/rclone_rmdir/"><i class="fa fa-book"></i> rclone rmdir</a>
-          <a class="dropdown-item" href="/commands/rclone_check/"><i class="fa fa-book"></i> rclone check</a>
-          <a class="dropdown-item" href="/commands/rclone_ls/"><i class="fa fa-book"></i> rclone ls</a>
-          <a class="dropdown-item" href="/commands/rclone_lsd/"><i class="fa fa-book"></i> rclone lsd</a>
-          <a class="dropdown-item" href="/commands/rclone_delete/"><i class="fa fa-book"></i> rclone delete</a>
-          <a class="dropdown-item" href="/commands/rclone_size/"><i class="fa fa-book"></i> rclone size</a>
-          <a class="dropdown-item" href="/commands/rclone_mount/"><i class="fa fa-book"></i> rclone mount</a>
-          <a class="dropdown-item" href="/commands/rclone_ncdu/"><i class="fa fa-book"></i> rclone ncdu</a>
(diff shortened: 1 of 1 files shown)


ncw:
All the commands sounds like a good idea.

They should be autogenerated though (one less thing to forget to update!)

Check out the stuff here for ideas on how to do that (this generates the commands page)

https://github.com/rclone/rclone/blob/164774d7e1d12aed0222fa55ba22f71766c5fa2d/docs/layouts/section/commands.html#L15-L20

You'll need to find the Content section to apply `.Data.Pages` and you'll need to ignore the sub commands but it shouldn't be too tricky.

albertony:
> hey should be autogenerated though (one less thing to forget to update!)
> 
> Check out the stuff here for ideas on how to do that (this generates the commands page)

I looked into that for half a second when doing the original code, but decided I didn't want to dig (drown) into that - unless necessary. I am thinking we should exclude subcommands, like the numerous config subcommands, from the list. Or else the list gets so long. And these are usually linked from the main command anyway. Do you agree? Then the autogen functionality would have to be able to filter that, but I'm guessing that is doable..

ncw:
> I looked into that for half a second when doing the original code, but decided I didn't want to dig (drown) into that - unless necessary.

:-)

I'm quite up to speed at Hugo at the moment, so if you get stuck I can help.

> I am thinking we should exclude subcommands, like the numerous config subcommands, from the list.

Yes

> Or else the list gets so long. And these are usually linked from the main command anyway. Do you agree? Then the autogen functionality would have to be able to filter that, but I'm guessing that is doable..

Its quite easy to filter stuff, if necessary we could do it with a regexp

Perhaps more sanely we could do it in `make commanddocs` and add another bit of data in the header

https://github.com/rclone/rclone/blob/164774d7e1d12aed0222fa55ba22f71766c5fa2d/docs/content/commands/rclone_about.md?plain=1#L2-L5

Say

```yaml
toplevel: true
```

Those things are very easy to filter with in hugo.

albertony:
Hacked together a first version of autogenerated commands drop-down.

Using dumb filtering using wordcount on title, anything with more than 2 words are skipped. Seems to match all the ones we want, and exclude the ones we don't want. If not, I'll have to create a new version using regex or add a property as you mentioned, but wanted to just try with the simplest version first.

Included the basic `rclone` command for now. It's mostly redundant with the "Overview" page, as it more or less just lists all subcommands in "SEE ALSO", so the only reason to include is that it in principle is a separate command:

> Show help for rclone commands, flags and backends.

>   -h, --help   help for rclone

But I'm not sure...

albertony:
Do we want to mark the most relevant ones (some or all of those that were originally in the short-list) to make them stand a bit out from the crowd? Bold font, `*` suffix, or maybe 🔥.. :) I'm afraid it just gets noisy, and I hope that this menu is used more for "reference" and that new users start out from https://rclone.org/docs, where there is a [subcommands](https://rclone.org/docs/#subcommands) section listing:

> The main rclone commands with most used first
> 
> rclone config - Enter an interactive configuration session.
> rclone copy - Copy files from source to dest, skipping already copied.
> rclone sync - Make source and dest identical, modifying destination only.
> rclone move - Move files from source to dest.
> ...

But it could make it quicker to get to the ones that you want to reach most of the times, like mount which is deep down in the list... I'm really not sure what I think is best (where are the ux usability test team when we need it!?), I may try out some variants a late night when I'm bored, but we could also just go with the plain version first and then consider improving later.. Thoughts?

ncw:
I think your change looks good. Its simple and will always be up to date which is good.

I'd like to think new users start from /docs with the curated list of commands, but when a bit more experienced having them all in alphabetical order on the menu is great for visibility.

So I think we should probably go with this :-)

PR #5966, opened 2022-01-30
operations: improve error message when source and destination hashes differ after copying - fixes #5268

### LINKED ISSUE #5268: Improve corrupted on transfer: MD5 hash differ message

v1.55.0

Help user with the error.

```
2021/04/22 22:27:01 ERROR : Google Photos/2016/10/VID_20161015_234919.mp4: corrupted on transfer: MD5 hash differ "a9f23639b60607a0571312a999e6ac84" vs "910b8923f1d418425ccc5d52dbac7cea"
```

It's much better if it said:

```
MD5 corrupted on transfer: expected a9f23639b60607a0571312a999e6ac84, got 910b8923f1d418425ccc5d52dbac7cea.
```

Or 

```
MD5 corrupted on transfer: expected 910b8923f1d418425ccc5d52dbac7cea, got a9f23639b60607a0571312a999e6ac84.
```

as you can see, this makes it distinguishable, which one is the wrong one.

LewisHook:

Improves the error message to include the names of the source and destination filesystems, when the source and destination file hashes differ after copying.

For example, if we are copying a file from the local filesystem to Amazon S3 and the file hashes are different after copying, then currently, we'll get the following error message:

> corrupted on transfer: md5 hash differ "eb260e9ae827821beceeed4104f0ad89" vs "59d0d19fc45ca69230d858f60a5557f8"

After this improvement, the error message will be:

> corrupted on transfer: md5 hash differ. source(local): "eb260e9ae827821beceeed4104f0ad89", destination(s3): "59d0d19fc45ca69230d858f60a5557f8"

I couldn't see an existing test for this error, let me know if you want me to try and add one.

I also noticed that the following error in the same function could benefit from a similar change, would you like me to make that change in a different PR?

https://github.com/rclone/rclone/blob/051685baa1e5c58d16e6dcc1d11ff136cfec8dae/fs/operations/operations.go#L513-L519

https://github.com/rclone/rclone/issues/5268

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -1763,14 +1763,14 @@ func (file *openFile) Close() (err error) {
 
 	// Check to see we read the correct number of bytes
 	if file.o.Size() != file.bytes {
-		return fmt.Errorf("object corrupted on transfer - length mismatch (want %d got %d)", file.o.Size(), file.bytes)
+		return fmt.Errorf("corrupted on transfer: lengths differ want %d vs got %d", file.o.Size(), file.bytes)
 	}
 
 	// Check the SHA1
 	receivedSHA1 := file.o.sha1
 	calculatedSHA1 := fmt.Sprintf("%x", file.hash.Sum(nil))
 	if receivedSHA1 != "" && receivedSHA1 != calculatedSHA1 {
-		return fmt.Errorf("object corrupted on transfer - SHA1 mismatch (want %q got %q)", receivedSHA1, calculatedSHA1)
+		return fmt.Errorf("corrupted on transfer: SHA1 hashes differ want %q vs got %q", receivedSHA1, calculatedSHA1)
 	}
 
 	return nil
--- a/backend/compress/compress.go
+++ b/backend/compress/compress.go
@@ -455,7 +455,7 @@ func (f *Fs) verifyObjectHash(ctx context.Context, o fs.Object, hasher *hash.Mul
 		if err != nil {
 			fs.Errorf(o, "Failed to remove corrupted object: %v", err)
 		}
-		return fmt.Errorf("corrupted on transfer: %v compressed hashes differ %q vs %q", ht, srcHash, dstHash)
+		return fmt.Errorf("corrupted on transfer: %v compressed hashes differ src(%s) %q vs dst(%s) %q", ht, f.Fs, srcHash, o.Fs(), dstHash)
 	}
 	return nil
 }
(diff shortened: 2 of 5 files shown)


ncw:
There are actually quite a few instance of this kind of error in rclone

```
$ git grep -i "corrupted on transfer" # with irrelevant results removed
backend/b2/b2.go:               return fmt.Errorf("object corrupted on transfer - length mismatch (want %d got %d)", file.o.Size(), file.bytes)
backend/b2/b2.go:               return fmt.Errorf("object corrupted on transfer - SHA1 mismatch (want %q got %q)", receivedSHA1, calculatedSHA1)
backend/compress/compress.go:           return fmt.Errorf("corrupted on transfer: %v compressed hashes differ %q vs %q", ht, srcHash, dstHash)
backend/crypt/crypt.go:                         return nil, fmt.Errorf("corrupted on transfer: %v crypted hash differ src %q vs dst %q", ht, srcHash, dstHash)
fs/operations/operations.go:            err = fmt.Errorf("corrupted on transfer: sizes differ %d vs %d", src.Size(), dst.Size())
fs/operations/operations.go:                    err = fmt.Errorf("corrupted on transfer: %v hash differ %q vs %q", hashType, srcSum, dstSum)
vfs/read.go:                    return fmt.Errorf("corrupted on transfer: %v hash differ %q vs %q", hashType, dstSum, srcSum)
```

So it would be a good idea to be consistent everywhere and fix them all in one commit.

In `crypt.go` we put in the `src` and `dst` words - do you think that is enough? That keeps the messages from being too long.

(The b2 ones are probably OK as they are.)

If we want to name the backends then we usually just use the `Fs` directory as its string representation is a sensible identifier for the backend. These are a bit longer than just the `Name()`, but the Name can easily be the same, eg if doing a transfer between two SFTP backends.

So I'd suggest either

    "corrupted on transfer: %v crypted hash differ src %q vs dst %q", ht, srcHash, dstHash

Or

    "corrupted on transfer: %v crypted hash differ src(%s) %q vs dst(%s) %q", ht, srcFs, srcHash, dstFs, dstHash

LewisHook:
I took a look around and most of the errors had access the `src` and `dst` string representations, there were a few that didn't so I left them out of those.

ncw (commented):
This looks very nice thank you :-)

Can you change the `f.Fs.String()` to `f.Fs`? The `.String()` method isn't necessary with the `%s` operator which will call that automatically. See inline for an example.

LewisHook:
Thanks, I've updated the messages to remove the `.String()`.

PR #5977, opened 2022-02-03
drive: added --drive-copy-shortcut-content, fixes #4604

### LINKED ISSUE #4604: Copy the contents of the shortcut when server side copying (google drive)
rclone v1.53.1
- os/arch: windows/amd64
- go version: go1.15

Copy the contents of the shortcut when server side copying (google drive)

Adding a flag to change this default action of copy to copy the contents of the shortcut.
And for the circular things, maybe we can combine such flag with a number to limit the tracking depth of shortcuts or the default depth of 1?
See more discussions on https://forum.rclone.org/t/copy-the-contents-of-the-shortcut-when-server-side-copying-google-drive/19094 .

Thanks!

abhiraj2:

Added --drive-copy-shortcut-content flag, that enables the user to copy the content of shortcuts when performing server side transfer.

#4604 

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -300,9 +300,15 @@ a non root folder as its starting point.
 			Help:     "Send files to the trash instead of deleting permanently.\n\nDefaults to true, namely sending files to the trash.\nUse `--drive-use-trash=false` to delete files permanently instead.",
 			Advanced: true,
 		}, {
-			Name:     "copy_shortcut_content",
-			Default:  false,
-			Help:     "Copy the files instead of just shortcuts when performing a server-side transfer.\nUse `--drive-copy-shortcut-content=true` to enable it. ",
+			Name:    "copy_shortcut_content",
+			Default: false,
+			Help: `Server side copy contents of shortcuts instead of the shortcut.
+
+When doing server side copies, normally rclone will copy shortcuts as
+shortcuts.
+
+If this flag is used then rclone will copy the contents of shortcuts
+rather than shortcuts themselves when doing server side copies.`,
 			Advanced: true,
 		}, {
 			Name:     "skip_gdocs",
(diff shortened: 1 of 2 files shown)


backend/drive/drive.go
```diff
@@ -299,6 +299,11 @@ a non root folder as its starting point.
 			Default:  true,
 			Help:     "Send files to the trash instead of deleting permanently.\n\nDefaults to true, namely sending files to the trash.\nUse `--drive-use-trash=false` to delete files permanently instead.",
 			Advanced: true,
+		}, {
+			Name:     "copy_shortcut_content",
+			Default:  false,
+			Help:     "Copy the files instead of just shortcuts when performing a server-side transfer.\nUse `--drive-copy-shortcut-content=true` to enable it. ",
```
ncw, line 305: The first line of this should be what you see when you run `rclone help flags` or `rclone help flags copy-shortcut` so it shouldn't be too long.

There should then be a `\n\n` and the help to go in the help text.

Its easier to write the help if you use multiple lines - see "root_folder_id" for an example.

Note also that `--drive-copy-shortcut-content=true` is equivalent to `--drive-copy-shortcut-content` and you probably don't need to mention it.

So I'd suggest something like

```go
Help: `Server side copy contents of shortcuts instead of the shortcut.

When doing server side copies, normally rclone will copy shortcuts as
shortcuts.

If this flag is used then rclone will copy the contents of shortcuts
rather than shortcuts themselves when doing server side copies.
```

backend/drive/drive.go
```diff
@@ -2375,8 +2381,13 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 	}
 
 	// get the ID of the thing to copy - this is the shortcut if available
```
ncw, line 2383: This comment needs updating

ncw (commented):
This looks good :-) Did it work?

This line needs updating in the docs

https://github.com/rclone/rclone/blob/c26dc69e1b42e2b4ca1f845cb9682bc432366bf4/docs/content/drive.md?plain=1#L413

Maybe add

```
(unless `--drive-copy-shortcut-content` is in use in which case the contents of the shortcut gets copied).
```

I put some comments inline also.

Thank you :-)

abhiraj2:
> This looks good :-) Did it work?

Yep it did work.
Let me make the changes as you've commented.

abhiraj2:
check it out now @ncw

ncw (approved):
This looks great now - thanks :-)

I'll just run the CI on it and if it passes I'll merge.

Thank you.

PR #5979, opened 2022-02-04
Configurable version suffix independent of version number

albertony:

Introduces a way to quickly set a custom pre-release label / version suffix on my own builds of rclone. 

As of now master version is "v1.58.0-DEV". I can override this easily with something like `-ldflags "-X github.com/rclone/rclone/fs.Version=v9.9.9-CUSTOM"`, but then I loose the version number it was based on. It would be better if I could keep the *number*, and only set my own suffix. That could be achived, of course, with a little `sed` magic or similar. But with this PR I can simply just set fs.VersionSuffix instead. This makes the build command very easy on any platform, and easy configurable in IDEs like VSCode etc.

Nice to have? Definitely not a must have. And perhaps there are better ways to do this already? Perhaps this file needs to be single variable-only to not break some build pipeline I'm not aware of?

PS: I would like to leave out the `-` separator from the VersionSuffix, which makes it more "semver-safe", but then it would no longer be variables-only file, so I started with the simplest "proof of concept" suggestion...

```
var Version = "v" + VersionNumber
if VersionSuffix != "" {
	Version += "-" + VersionSuffix
}
```

PS2: For reference, .NET uses the name `VersionPrefix` for the number part ("1.58.0") and `VersionSuffix` for the pre-release label ("DEV"), and `Version` for the combination of the two separated with `-`. And it also has a `InformationalVersion` which typically is the combination of `Version` and a commit hash separated with `+` ("1.58.0-DEV+27d5f545fa0e2f09629d0bcbb0e571fc79341f2b").

PS3: Speaking of semver-safe... in `cmd/version/version.go` and `bin/cross-compile.go` we parse the string into a semver struct (after first trimming off the "v" prefix which is not strict semver) and use the version number components:
https://github.com/rclone/rclone/blob/aa2d7f00c2376f2dc44f50afb3276402ddfdb44d/bin/cross-compile.go#L202-L227

#### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,18 +245,18 @@ retag:
 startdev:
 	@echo "Version is $(VERSION)"
 	@echo "Next version is $(NEXT_VERSION)"
-	echo -e "package fs\n\n// Version of rclone\nvar Version = \"$(NEXT_VERSION)-DEV\"\n" | gofmt > fs/version.go
+	echo -e "package fs\n\n// VersionTag of rclone\nvar VersionTag = \"$(NEXT_VERSION)\"\n" | gofmt > fs/versiontag.go
 	echo -n "$(NEXT_VERSION)" > docs/layouts/partials/version.html
 	echo "$(NEXT_VERSION)" > VERSION
-	git commit -m "Start $(NEXT_VERSION)-DEV development" fs/version.go VERSION docs/layouts/partials/version.html
+	git commit -m "Start $(NEXT_VERSION)-DEV development" fs/versiontag.go VERSION docs/layouts/partials/version.html
 
 startstable:
 	@echo "Version is $(VERSION)"
 	@echo "Next stable version is $(NEXT_PATCH_VERSION)"
-	echo -e "package fs\n\n// Version of rclone\nvar Version = \"$(NEXT_PATCH_VERSION)-DEV\"\n" | gofmt > fs/version.go
+	echo -e "package fs\n\n// VersionTag of rclone\nvar VersionTag = \"$(NEXT_PATCH_VERSION)\"\n" | gofmt > fs/versiontag.go
 	echo -n "$(NEXT_PATCH_VERSION)" > docs/layouts/partials/version.html
 	echo "$(NEXT_PATCH_VERSION)" > VERSION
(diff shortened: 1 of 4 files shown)


ncw:
> Nice to have? Definitely not a must have. And perhaps there are better ways to do this already? Perhaps this file needs to be single variable-only to not break some build pipeline I'm not aware of?

What it breaks is this...

https://github.com/rclone/rclone/blob/aa2d7f00c2376f2dc44f50afb3276402ddfdb44d/Makefile#L245-L259

albertony:
Ah yes, that is; The makefile would revert this pr's changes next time it's run.

Pushed an alternative implementation, what do you think?

Now any of the following will work, depending on what you want to change, assuming default is "v1.58.0-DEV":
- `-ldflags "-X github.com/rclone/rclone/fs.VersionTag=v9.9.9"` --> "v9.9.9-DEV"
- `-ldflags "-X github.com/rclone/rclone/fs.VersionSuffix=CUSTOM"` --> "v1.58.0-CUSTOM"
- `-ldflags "-X github.com/rclone/rclone/fs.Version=v9.9.9-CUSTOM"`  --> "v9.9.9-CUSTOM"

Script editing the file, like the makefile does, is easy for VersionTag and VersionSuffix individually, but not the complete Version.

ncw (commented):
I'm happy to merge this now, but it needs updating since we did a release, so the version number needs changing.

I'm sure its going to break something, but I can't think what!

If you rebase and update I'll merge

Thank you

albertony:
> I'm happy to merge this now, but it needs updating since we did a release, so the version number needs changing.

Done: Rebased, and updated version number according to master.

> I'm sure its going to break something, but I can't think what!

Sure, I may have overlooked something (but then at least you get to yell "I told you" or "I knew it"... 🙇‍♂️ )

ncw:
> > I'm happy to merge this now, but it needs updating since we did a release, so the version number needs changing.
> 
> Done: Rebased, and updated version number according to master.

Thank you.
 
> > I'm sure its going to break something, but I can't think what!
> 
> Sure, I may have overlooked something (but then at least you get to yell "I told you" or "I knew it"... bowing_man )

I predict that this will come to bite at release time or point release time - we shall see :-)

And if it goes wrong, it will be all on me as I did the review, so no yelling!

PR #6018, opened 2022-03-01
digistorage: a new backend for Digi Storage cloud storage service.

jkmcnk:

Adds a new backend for Digi Storage, a Koofr API compatible cloud storage service.

No.

#### Checklist


ncw:
This is looking good :-)

How many koofr compatible cloud storage providers are there?

I'd probably prefer it if we make a `Provider` setting for the `Koofr` backend like we do with [s3](https://rclone.org/s3/)

digistorage can still have its own menu entries and links - that is helpful for users but I don't want to proliferate the backends since I think the only difference is the `endpoint` isn't it and we have the `Provider` mechanism already to cope with that.

jkmcnk:
a few.  :)

do not really plan on adding unless the client asks, like digistorage did. they will also be getting a spanish site in the near future.

will look at the provider thing and adapt accordingly, noticed the s3-compatible backends having their own menu entries, but didn't bother to investigate further.

Issue #4391, opened 2020-06-26
Display names / description field for remotes.


jtagcat:
#### What problem are you are trying to solve?
 - Have many, many remotes, with easy to type, cryptic names
 - You don't exactly remember, what all the remotes do
 - So you name everything with 30 various characters
 - but I think that's too much to type, and copy pasta isn't the fastest

#### How do you think rclone should be changed to solve that?
What about 'display names'? They would only appear in rclone config and listremotes (with optional flag, bc lsremotes could be used by scripts). Other alternative for display names could be a description field.

ncw:
This is a nice idea - I like the name `description`

To implement we probably need to ask the user for the description at the point they make the remote.

Do you want to have a go at this?

jtagcat:
> This is a nice idea - I like the name `description`
> 
> To implement we probably need to ask the user for the description at the point they make the remote.

I think it should be shoved under advanced (the last thing before confirming). Thoughs/do you agree?

> Do you want to have a go at this?

Suuureee.. I'll assign myself, but as with the previous time, can't promise anything.

jtagcat:
@ncw:
> I think it should be shoved under advanced (the last thing before confirming). Thoughs/do you agree?

_digging through the backlog_ I might throw this under hacktoberfest.

Talking about backlog, I did some ~stonks~ stats. Tab count (backlog, net total compared to the previous hour online) (around ¼-⅓ of the actual measure):

![image](https://user-images.githubusercontent.com/38327267/95528968-352ab580-09c9-11eb-9dc4-569585029c6d.png)

ncw:
@jtagcat wrote
>  I might throw this under hacktoberfest.
Go for it!

> Talking about backlog, I did some stats. Tab count (backlog, net total compared to the previous hour online) (around ¼-⅓ of the actual measure):

Nice stats! I don't know what they are measuring - can you explain more?

jtagcat:
I made a wiki page on STG explaining it, but they disabled the wiki... I use SimpleTabGroups for my tabs, the numbers indicate tabs. I guess this could go in more detail:
![Screenshot from 2020-10-10 09-45-46](https://user-images.githubusercontent.com/38327267/95648101-4d332f80-0ac4-11eb-8967-9547f497b910.png)
![Screenshot from 2020-10-10 09-45-58](https://user-images.githubusercontent.com/38327267/95648098-4c020280-0ac4-11eb-9d92-dc6ccf6c6d91.png)

[...]

PR #6019, opened 2022-03-02
backend: adding description option to all backend filesystems

andrewwillette:
#### Checklist

Attempting to add the feature as discussed in [this issue](https://github.com/rclone/rclone/issues/4391).

A piece I could be missing, but I'm unsure of how it's used, is adding documentation to each of the markdown files for the backends for `--description` or `--s3-description` [here for s3](https://github.com/rclone/rclone/blame/5ce3019fc1b64fccfce3aef7e25be97ee453fca2/docs/content/s3.md#L1823), and this would apply to each of the backends.

If I'm missing tests, please let me know. It wasn't clear to me that the backends test that each of the options exist.

Thank you!

### DIFF
--- a/backend/alias/alias.go
+++ b/backend/alias/alias.go
@@ -3,6 +3,7 @@ package alias
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/rclone/rclone/fs"
@@ -18,6 +19,7 @@ func init() {
 		Name:        "alias",
 		Description: "Alias for an existing remote",
 		NewFs:       NewFs,
+		Config:      riConfig,
 		Options: []fs.Option{{
 			Name:     "remote",
 			Help:     "Remote or path to alias.\n\nCan be \"myremote:path/to/dir\", \"myremote:bucket\", \"myremote:\" or \"/local/path\".",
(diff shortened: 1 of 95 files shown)


albertony:
Nice!

> A piece I could be missing, but I'm unsure of how it's used, is adding documentation to each of the markdown files for the backends for --description or --s3-description [here for s3](https://github.com/rclone/rclone/blame/5ce3019fc1b64fccfce3aef7e25be97ee453fca2/docs/content/s3.md#L1823), and this would apply to each of the backends.

Documentation of the option will be added to markdown files automatically, in time for release. So as long as you write a `Help:` string you are covered.

PS: Some linting errors, it expects whitespace alignment like this:
```
Name:     config.ConfigDescription,
Help:     config.ConfigDescriptionHelp,
Advanced: true,
```

andrewwillette:
@albertony, thanks for the feedback, I really appreciate it. I believe I've fixed the linting issues. (not sure how to trigger build, guessing maybe you do it [manually](https://github.com/rclone/rclone/actions/runs/1919738523/workflow#L18)?) I don't know.

Please let me know if anything else is needed/desired, I'll gladly do it.

ncw:
I wonder whether the description definition is in the right place.

To be useful, we probably need this in the layer above the backends - none of the backends use the description in the config, but we want it there so we can display it to the user (eg in `rclone listremotes`) or the user can see it in the config file.

Maybe we should be asking for the desciption as part of the remote creation / update rather than injecting it into each backend?

We have one parameter like that already `type` which isn't handled by the backend.

What do you think?

andrewwillette:
@ncw and @albertony, I've force pushed updates to set the "description" option via the RegisterInfo#Config callback.

If you have time, I'd love to get feedback/+1s.

Thanks so much,
Andrew

andrewwillette:
Closing due to staleness.

PR #6025, opened 2022-03-06
sync: overlap check is now filter-sensitive

Schlossgeist:

Previously, the overlap check was based on simple prefix checks of the source and destination paths. Now it actually checks whether the destination is excluded via any filter rule or a "--exclude-if-present"-file.

[Forum Post](https://forum.rclone.org/t/excluding-overlapping-destination/29591)

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -806,13 +806,13 @@ func OverlappingFilterCheck(ctx context.Context, fdst fs.Fs, fsrc fs.Fs) bool {
 	fsrcRoot := fixRoot(fsrc)
 	if strings.HasPrefix(fdstRoot, fsrcRoot) {
 		fdstRelative := fdstRoot[len(fsrcRoot):]
-		return FilterCheckR(ctx, fdstRelative, 0, fsrc)
+		return filterCheckR(ctx, fdstRelative, 0, fsrc)
 	}
 	return strings.HasPrefix(fsrcRoot, fdstRoot)
 }
 
-// FilterCheckR checks if fdst would be included in the sync
-func FilterCheckR(ctx context.Context, fdstRelative string, pos int, fsrc fs.Fs) bool {
+// filterCheckR checks if fdst would be included in the sync
+func filterCheckR(ctx context.Context, fdstRelative string, pos int, fsrc fs.Fs) bool {
 	include := true
 	fi := filter.GetConfig(ctx)
 	includeDirectory := fi.IncludeDirectory(ctx, fsrc)
@@ -831,7 +831,7 @@ func FilterCheckR(ctx context.Context, fdstRelative string, pos int, fsrc fs.Fs)
 				return true
 			}
 			pos++
-			include = FilterCheckR(ctx, fdstRelative, pos, fsrc)
+			include = filterCheckR(ctx, fdstRelative, pos, fsrc)
 		}
 	}
 	return include
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -796,6 +796,53 @@ func Overlapping(fdst, fsrc fs.Info) bool {
 	return strings.HasPrefix(fdstRoot, fsrcRoot) || strings.HasPrefix(fsrcRoot, fdstRoot)
 }
 
+// OverlappingFilterCheck returns true if fdst and fsrc point to the same
+// underlying Fs and they overlap without fdst being excluded by any filter rule.
+func OverlappingFilterCheck(ctx context.Context, fdst fs.Fs, fsrc fs.Fs) bool {
+	if !SameConfig(fdst, fsrc) {
+		return false
+	}
+	fdstRoot := fixRoot(fdst)
+	fsrcRoot := fixRoot(fsrc)
+	if strings.HasPrefix(fdstRoot, fsrcRoot) {
+		fdstRelative := strings.ReplaceAll(fdstRoot, fsrcRoot, "")
+		return FilterCheckR(ctx, fdstRelative, "", fsrc)
+	}
+	return strings.HasPrefix(fsrcRoot, fdstRoot)
+}
+
+// FilterCheckR checks if fdst would be included in the sync
+func FilterCheckR(ctx context.Context, fdstRelative string, inputPath string, fsrc fs.Fs) bool {
+	include := true
+	fi := filter.GetConfig(ctx)
+	includeDirectory := fi.IncludeDirectory(ctx, fsrc)
+	entries, _ := fsrc.List(ctx, inputPath)
+	for _, entry := range entries {
+		switch x := entry.(type) {
+		case fs.Directory:
+			currentPath := x.Remote()
+			if !strings.HasSuffix(currentPath, "/") {
+				currentPath += "/"
+			}
+			if strings.HasPrefix(fdstRelative, currentPath) {
+				include, _ = includeDirectory(currentPath)
+				if include {
+					if currentPath == fdstRelative {
+						return true
+					}
+					include = FilterCheckR(ctx, fdstRelative, currentPath, fsrc)
+				}
+			}
+		case fs.Object:
+			basename := path.Base(x.Remote())
+			if basename == fi.Opt.ExcludeFile {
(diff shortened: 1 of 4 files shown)


Schlossgeist:
I've noticed some initially unwanted (but maybe reasonable) behavior. Because of the way I implemented the recursive checks, the sync command is only allowed to be executed when the destination directory already exists (in contrast to sync commands under normal circumstances where the destination directory is created automatically).
Although I think that the ability to sync to subdirectories which are exculded from the sync is inherently useful, I also think that this is kind of an edge-case usecase which should required some additional action from the user (i. e. creating the destination directory). It also gives beginner users the impression that it is generally not ok to try and sync or move directories in this way.

What are your thoughts on this? @ncw

ncw:
> I've noticed some initially unwanted (but maybe reasonable) behavior. Because of the way I implemented the recursive checks, the sync command is only allowed to be executed when the destination directory already exists

This is because you are listing the directory I think.

It should be possible to do these checks without actually calling `List` shouldn't it? I'm concerned about calling `List` since it is potentially an expensive operation (think a bucket with 1 million entries) so we don't want to do it twice.

> (in contrast to sync commands under normal circumstances where the destination directory is created automatically). Although I think that the ability to sync to subdirectories which are exculded from the sync is inherently useful, I also think that this is kind of an edge-case usecase which should required some additional action from the user (i. e. creating the destination directory). It also gives beginner users the impression that it is generally not ok to try and sync or move directories in this way.

I'm not sure I understood the problem there! Can you give some examples please?

I will give the code a quick first review now.

fs/operations/operations.go
```diff
@@ -796,6 +796,53 @@ func Overlapping(fdst, fsrc fs.Info) bool {
 	return strings.HasPrefix(fdstRoot, fsrcRoot) || strings.HasPrefix(fsrcRoot, fdstRoot)
 }
 
+// OverlappingFilterCheck returns true if fdst and fsrc point to the same
+// underlying Fs and they overlap without fdst being excluded by any filter rule.
+func OverlappingFilterCheck(ctx context.Context, fdst fs.Fs, fsrc fs.Fs) bool {
```
ncw, line 801: This needs testing in `operations_test.go`

fs/operations/operations.go
```diff
@@ -796,6 +796,53 @@ func Overlapping(fdst, fsrc fs.Info) bool {
...
+	fsrcRoot := fixRoot(fsrc)
+	if strings.HasPrefix(fdstRoot, fsrcRoot) {
+		fdstRelative := strings.ReplaceAll(fdstRoot, fsrcRoot, "")
+		return FilterCheckR(ctx, fdstRelative, "", fsrc)
+	}
+	return strings.HasPrefix(fsrcRoot, fdstRoot)
+}
+
+// FilterCheckR checks if fdst would be included in the sync
+func FilterCheckR(ctx context.Context, fdstRelative string, inputPath string, fsrc fs.Fs) bool {
```
ncw, line 815: This needs testing in `operations_test.go` also

fs/operations/operations.go
```diff
@@ -796,6 +796,53 @@ func Overlapping(fdst, fsrc fs.Info) bool {
...
+	include := true
+	fi := filter.GetConfig(ctx)
+	includeDirectory := fi.IncludeDirectory(ctx, fsrc)
+	entries, _ := fsrc.List(ctx, inputPath)
```
ncw, line 819: Really don't like calling `List` here. We should be able to do this just by examining the filter and paths shouldn't we?

fs/sync/sync_test.go
```diff
@@ -1442,6 +1442,38 @@ func TestSyncOverlap(t *testing.T) {
...
+	defer r.Finalise()
+
+	fi, err := filter.NewFilter(nil)
+	require.NoError(t, err)
+	require.NoError(t, fi.Add(false, "/rclone-sync-test/"))
+	ctx = filter.ReplaceConfig(ctx, fi)
+
+	subRemoteName := r.FremoteName + "/rclone-sync-test"
+	FremoteSync, err := fs.NewFs(ctx, subRemoteName)
+	FremoteSync.Mkdir(ctx, "")
```
ncw, line 1458: The error for this needs to be checked. 

    require.NoError(t, FremoteSync.Mkdir(ctx, ""))

is fine

ncw (commented):
A quick initial review :-)

Thank you

fs/operations/operations.go
```diff
@@ -796,6 +796,57 @@ func Overlapping(fdst, fsrc fs.Info) bool {
...
+// OverlappingFilterCheck returns true if fdst and fsrc point to the same
+// underlying Fs and they overlap without fdst being excluded by any filter rule.
+func OverlappingFilterCheck(ctx context.Context, fdst fs.Fs, fsrc fs.Fs) bool {
+	if !SameConfig(fdst, fsrc) {
+		return false
+	}
+	fdstRoot := fixRoot(fdst)
+	fsrcRoot := fixRoot(fsrc)
+	if strings.HasPrefix(fdstRoot, fsrcRoot) {
+		fdstRelative := strings.ReplaceAll(fdstRoot, fsrcRoot, "")
```
ncw, line 808: Probably better written something like this (`ReplaceAll` could replace stuff it isn't supposed to).

    fdstRelative := fdstRoot[len(fsrcRoot):]

fs/operations/operations.go
```diff
@@ -796,6 +796,57 @@ func Overlapping(fdst, fsrc fs.Info) bool {
...
+			if currentPath == fdstRelative {
+				return true
+			}
+			pos++
+			include = FilterCheckR(ctx, fdstRelative, pos, fsrc)
+		}
+	}
+
+	if !include {
+		err := fsrc.Mkdir(ctx, fdstRelative)
```
ncw, line 841: It shouldn't be necessary to run `Mkdir` here - the sync will do it for you.

fs/operations/operations.go
```diff
@@ -796,6 +796,57 @@ func Overlapping(fdst, fsrc fs.Info) bool {
...
+func FilterCheckR(ctx context.Context, fdstRelative string, pos int, fsrc fs.Fs) bool {
+	include := true
+	fi := filter.GetConfig(ctx)
+	includeDirectory := fi.IncludeDirectory(ctx, fsrc)
+	dirs := strings.SplitAfterN(fdstRelative, "/", pos+2)
+	newPath := ""
+	for i := 0; i <= pos; i++ {
+		newPath += dirs[i]
+	}
+	dir := fs.NewDir(newPath, time.Now())
```
ncw, line 824: I don't think this `dir` gets used anywhere does it? How about `currentPath := newPath`?

ncw (commented):
I think this looks much better now.

I don't think you need to call `Mkdir` either.

Can you write a table driven test for `FilterCheckR` testing all the corner cases - it is quite complicated that function and I'd like to be sure it is correct!

A test for `OverlappingFilterCheck` would be nice too.

Thank you :-)

Schlossgeist:
I used the existing `TestOverlapping` for _inspiration_. I only wrote a test for `OverlappingFilterCheck` though. Can you explain why there should be a separate test for `FilterCheckR`? This function is only called from `OverlappingFilterCheck` with almost the same arguments every time. It feels a bit redundant to test the same function in `sync_test.go` and twice in `operations_test.go`.

ncw:
> I used the existing `TestOverlapping` for _inspiration_. I only wrote a test for `OverlappingFilterCheck` though. Can you explain why there should be a separate test for `FilterCheckR`? This function is only called from `OverlappingFilterCheck` with almost the same arguments every time. It feels a bit redundant to test the same function in `sync_test.go` and twice in `operations_test.go`.

It might be that `FilterCheckR` shouldn't be an exported function, if so rename it `filterCheckR` and then you don't need to make tests for it :-)

Schlossgeist:
@ncw Is this PR ready for merging or should I change anything else?

Schlossgeist:
No problem. Thank you for investing your time and providing this great piece of software for free.

Issue #2126, opened 2018-03-09
Ability to restore a snapshot of the filesystem at a specific timestamp


cowwoc:
#### What is the problem you are having with rclone?
When using versioned remotes, like B2, I would like to be able to restore the contents of my system at a given date. Although the information is technically available, it is extremely difficult from a usability perspective to figure out which files needs to be downloaded, and renaming them as necessary so they match the original filename (e.g. b2 appends the store timestamp to the filename. We'd need to reverse this process).

#### What is your rclone version (eg output from `rclone -V`)
rclone v1.39
- os/arch: windows/amd64
- go version: go1.9.2

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Windows 10, 64bit

####  Which cloud storage system are you using? (eg Google Drive)
B2

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
rclone copy remote:blah blah

Recapping what I wrote on https://forum.rclone.org/t/how-to-restore-view-of-system-at-a-specific-timestamp/5004/5:

> https://rclone.org/b2/ explains how to list file versions and restore a specific version of a file, one at a time.
> 
> I am trying to protect myself from the possibility where the latest version of my filesystem has become corrupt by a virus. Is there a mechanism that will allow me to restore the view of the filesystem at a particular date/time?
> 
> Specifically, I expect to be able to to COPY files from the remote to the local filename, providing a single date/time. I expect B2 to find the latest version of a file on or before this date/time and COPY it to the local filesystem without the date/time suffix.
> 
> Meaning:
> 
> * I would invoke: rclone copy --latest-before=2018-03-08-205431-000 remote:* local:*
> * rclone would find files test.txt-v2018-03-08-205524-000, test2.txt-v2017-03-08-205524-000 as the latest file before the date/time
> * rclone would restore target filenames “test.txt” and “test2.txt” (suffix removed)

ncw:
This would probably be a b2 specific command, say `--b2-latest-before datespec`.  It wouldn't be too tricky to implement.  If anyone would like to go I'd happily spec out what needs to be done.

wolfv6:
This issue might be related: [incremental strategy](https://github.com/ncw/rclone/issues/18)

wolfv6:
Maybe I am missing something, but I don't see how rclone can be used to restore a snapshot at a specific timestamp reliably.
For example, this daily backup:

[...]

Issue #1532, opened 2017-07-15
Generalize --b2-versions for other backends


jrk:
Dropbox, Google Drive, and others support file versions in their APIs. They are less rich than b2's versions (I believe Dropbox's versions are read-only, for example), but accessing them would still be very useful.

jrk:
FWIW, I would personally like to see the interface be a bit more orthogonal than the `--b2-versions` support seems to be, so relevant commands (e.g., copy/sync/...) would generically apply to a specified source version, using normal file names, rather than embedding the version information in magic file names.

My current use case: I'd like to record local snapshots of my Dropbox at specific points in time in the past. So: `rclone sync --version some-date-or-version-number-string dropbox:/ /local/dropbox/mirror`, then snapshot `/local/dropbox/mirror`, and repeat with the next desired version.

ncw:
What would you see as an interface?  Something like `--versions-before 2017-02-03 00:00:00` which would show file versions before that date?

The advantage of the `--b2-versions` flag is that you can easily discover all the versions you do have.

> My current use case: I'd like to record local snapshots of my Dropbox at specific points in time in the past. So: rclone sync --version some-date-or-version-number-string dropbox:/ /local/dropbox/mirror, then snapshot /local/dropbox/mirror, and repeat with the next desired version.

You can do this manually at the moment using [--backup-dir](https://rclone.org/docs/#backup-dir-dir) which works with most providers

jrk:
To be clear, I want to record and save snapshots of past versions which Dropbox has already recorded (I'm trying to effectively export my long Dropbox version history to somewhere I control), not just avoid deleting old versions on new Dropbox updates, so `--backup-dir` isn't actually what I want.

For my use case, and I think perhaps in many cases in general, I think the ideal interface would be something like `--version-at-date 2017-02-03 00:00:00` which would copy/sync the whole requested tree as it existed at that point in time. This is not very isomorphic to the underlying API, though, and I believe would require pre-indexing the version history of everything at the source before reconstructing that point in time view to know what revisions to pull of each file. This would be an extremely useful feature of the tool, though, specifically because it is a straightforward human-level thing to want to do which requires a fair amount of legwork to build on top of the raw APIs.

Seeing all versions is also useful, but I'd expect that to be a function of the `ls` command family.

#### Footnote:
[...]

PR #6035, opened 2022-03-15
b2: Add b2-version-at flag to show file versions at time

swazrgb:

The goal is to be able to browse the b2 bucket state as it was at a specified point in time, using b2's file versions. This is especially useful in combination with other backup tools, such as restic, which may use rclone as a backend.

I have not added any tests, since I did not find any for `--b2-versions` which is similar functionality but I might be looking in the wrong place.

For example, this allows the following:

```
rclone mount --b2-version-at=2022-03-13T00:00:00Z b2:bucket /mnt/rclone
```

Or when using a backup tool such as restic:

```
restic -o 'rclone.args=serve restic --stdio --b2-version-at=2022-03-13T00:00:00Z' --no-lock mount /mnt/restic
```

To access pruned backup files whose data is still kept by b2 lifecycle rules.

Yes, but not by me: https://github.com/rclone/rclone/issues/2126

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -64,7 +64,8 @@ const (
 
 // Globals
 var (
-	errNotWithVersions = errors.New("can't modify or delete files in --b2-versions mode")
+	errNotWithVersions  = errors.New("can't modify or delete files in --b2-versions mode")
+	errNotWithVersionAt = errors.New("can't modify or delete files in --b2-version-at mode")
 )
 
 // Register with Fs
@@ -105,6 +106,11 @@ in the [b2 integrations checklist](https://www.backblaze.com/b2/docs/integration
 			Name:     "versions",
 			Help:     "Include old versions in directory listings.\n\nNote that when using this no file write operations are permitted,\nso you can't upload files or delete them.",
 			Default:  false,
+			Advanced: true,
+		}, {
+			Name:     "version_at",
+			Help:     "Show file versions as they were at the specified time.\n\nNote that when using this no file write operations are permitted,\nso you can't upload files or delete them.",
+			Default:  fs.Time{},
 			Advanced: true,
 		}, {
 			Name:    "hard_delete",
@@ -211,6 +217,7 @@ type Options struct {
 	Endpoint                      string               `config:"endpoint"`
 	TestMode                      string               `config:"test_mode"`
 	Versions                      bool                 `config:"versions"`
+	VersionAt                     fs.Time              `config:"version_at"`
 	HardDelete                    bool                 `config:"hard_delete"`
 	UploadCutoff                  fs.SizeSuffix        `config:"upload_cutoff"`
 	CopyCutoff                    fs.SizeSuffix        `config:"copy_cutoff"`
@@ -696,9 +703,12 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 		Method: "POST",
 		Path:   "/b2_list_file_names",
 	}
-	if hidden {
+	if hidden || f.opt.VersionAt.IsSet() {
 		opts.Path = "/b2_list_file_versions"
 	}
+
+	lastFileName := ""
+
 	for {
 		var response api.ListFileNamesResponse
 		err := f.pacer.Call(func() (bool, error) {
@@ -728,7 +738,21 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 			if addBucket {
 				remote = path.Join(bucket, remote)
 			}
+
+			if f.opt.VersionAt.IsSet() {
+				if time.Time(file.UploadTimestamp).After(time.Time(f.opt.VersionAt)) {
+					// Ignore versions that were created after the specified time
+					continue
+				}
+
+				if file.Name == lastFileName {
+					// Ignore versions before the already returned version
+					continue
+				}
+			}
+
 			// Send object
+			lastFileName = file.Name
 			err = fn(remote, file, isDirectory)
 			if err != nil {
 				if err == errEndList {
@@ -1828,6 +1852,9 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 	if o.fs.opt.Versions {
 		return errNotWithVersions
 	}
+	if o.fs.opt.VersionAt.IsSet() {
+		return errNotWithVersionAt
+	}
 	size := src.Size()
 
 	bucket, bucketPath := o.split()
@@ -1982,6 +2009,9 @@ func (o *Object) Remove(ctx context.Context) error {
 	bucket, bucketPath := o.split()
 	if o.fs.opt.Versions {
 		return errNotWithVersions
+	}
+	if o.fs.opt.VersionAt.IsSet() {
+		return errNotWithVersionAt
 	}
 	if o.fs.opt.HardDelete {
 		return o.fs.deleteByID(ctx, o.id, bucketPath)
--- a/fs/parsetime.go
+++ b/fs/parsetime.go
@@ -1,0 +1,91 @@
+package fs
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// Time is a time.Time with some more parsing options
+type Time time.Time
+
+// For overriding in unittests.
+var (
+	timeNowFunc = time.Now
+)
+
+// Turn Time into a string
+func (t Time) String() string {
+	if !t.IsSet() {
+		return "off"
+	}
+	return time.Time(t).Format(time.RFC3339Nano)
+}
+
+// IsSet returns if the time is not zero
+func (t Time) IsSet() bool {
+	return !time.Time(t).IsZero()
+}
+
+// ParseTime parses a time or duration string as a Time.
+func ParseTime(date string) (t time.Time, err error) {
+	if date == "off" {
+		return time.Time{}, nil
+	}
+
+	now := timeNowFunc()
+
+	// Attempt to parse as a text time
+	t, err = parseTimeDates(date)
+	if err == nil {
+		return t, nil
+	}
+
+	// Attempt to parse as a time.Duration offset from now
+	d, err := time.ParseDuration(date)
+	if err == nil {
+		return now.Add(-d), nil
+	}
+
+	d, err = parseDurationSuffixes(date)
+	if err == nil {
+		return now.Add(-d), nil
+	}
+
+	return t, err
+}
+
+// Set a Time
+func (t *Time) Set(s string) error {
+	parsedTime, err := ParseTime(s)
+	if err != nil {
+		return err
+	}
+	*t = Time(parsedTime)
+	return nil
+}
+
+// Type of the value
+func (t Time) Type() string {
+	return "Time"
+}
+
+// UnmarshalJSON makes sure the value can be parsed as a string in JSON
+func (t *Time) UnmarshalJSON(in []byte) error {
+	var s string
+	err := json.Unmarshal(in, &s)
+	if err != nil {
+		return err
+	}
+
+	return t.Set(s)
+}
+
+// Scan implements the fmt.Scanner interface
+func (t *Time) Scan(s fmt.ScanState, ch rune) error {
+	token, err := s.Token(true, nil)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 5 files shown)


backend/b2/b2.go
```diff
@@ -696,9 +703,20 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 		Method: "POST",
 		Path:   "/b2_list_file_names",
 	}
-	if hidden {
+	if hidden || versionAt != "" {
 		opts.Path = "/b2_list_file_versions"
 	}
+
+	var versionAtTime time.Time
```
ncw, line 710: This should be parsed into a time.Time in NewFs so we only do it once at the start.

We can then use `f.versionAtTime.IsZero()` to see if the feature is active.

backend/b2/b2.go
```diff
@@ -662,7 +669,7 @@ var errEndList = errors.New("end list")
 // If hidden is set then it will list the hidden (deleted) files too.
 //
 // if findFile is set it will look for files called (bucket, directory)
-func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBucket bool, recurse bool, limit int, hidden bool, findFile bool, fn listFn) error {
+func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBucket bool, recurse bool, limit int, hidden bool, versionAt string, findFile bool, fn listFn) error {
```
ncw, line 672: I'd say don't change the arguments for the function since the desired item is in `f` anyway.

backend/b2/b2.go
```diff
@@ -1828,6 +1860,9 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 	if o.fs.opt.Versions {
 		return errNotWithVersions
 	}
+	if o.fs.opt.VersionAt != "" {
```
ncw, line 1863: I would set `opt.Versions` in `NewFs` if `VersionAt` is specified. Then we can do away with all these extra checks.

backend/b2/b2.go
```diff
@@ -106,6 +107,11 @@ in the [b2 integrations checklist](https://www.backblaze.com/b2/docs/integration
 			Help:     "Include old versions in directory listings.\n\nNote that when using this no file write operations are permitted,\nso you can't upload files or delete them.",
 			Default:  false,
 			Advanced: true,
+		}, {
+			Name:     "version_at",
+			Help:     "Show file versions as they were at the specified time.\n\nNote that when using this no file write operations are permitted,\nso you can't upload files or delete them.",
+			Default:  "",
```
ncw, line 113: I think using a `fs.Duration` for the type would be best here. You'd then be able to say

```
--version-at 1w # for 1 week ago
--version-at 2006-01-01 # for 1st jan 2006
```

Or any of the formats here

```
        time.RFC3339,
        "2006-01-02T15:04:05",
        "2006-01-02 15:04:05",
        "2006-01-02",
```

This wants to be converted into a `time.Time` in `NewFs` to make it absolute.

backend/b2/b2.go
```diff
@@ -211,6 +217,7 @@ type Options struct {
 	Endpoint                      string               `config:"endpoint"`
 	TestMode                      string               `config:"test_mode"`
 	Versions                      bool                 `config:"versions"`
+	VersionAt                     string               `config:"version_at"`
```
ncw, line 220: `fs.Duration` here

ncw (commented):
Nice feature - thank you :-)

See inline for comments

backend/b2/b2.go
swazrgb: I'm worried that might give incorrect behavior in certain obscure situations. When `opt.Versions` is set, various methods change their behavior because they expect filenames to have the version suffixes, which is not the case in version-at mode.

`getMetaDataListing` will try to strip the version suffix if `opt.Versions` is set, so if a regular file with such a version suffix happens to be present, unexpected behavior will occur. Instead it should call list as normal, which will return the correct result.

backend/b2/b2.go
swazrgb: That looks way nicer. What is the best way to convert this duration to an absolute time? I'm thinking:

```
	var versionAtTime time.Time
	if opt.VersionAt.IsSet() {
		versionAtTime = time.Now().Add(-time.Duration(opt.VersionAt))
	}
	f := &Fs{
		...,
		versionAtTime: versionAtTime,
	}
```

But there is no guarantee that `time.Now()` is the same value here as it was when the duration was parsed, which matters if the user specifies an absolute time using `--b2-version-at=2022-03-13T00:00:00Z`. I'd expect a `time.Time` of `2022-03-15 04:47:56.0`, not something like `2022-03-15 04:47:56.10338005`.

The `ParseDuration` appears to offset from `time.Now()`, instead of some accessible constant. So, it's close, but not exactly the same as the user input. Probably not a huge deal, especially since the b2 api counts time by the second, but I am concerned this may not be correct.

Is this fine, or do you have a better suggestion?

backend/b2/b2.go
ncw: OK fair enough!

backend/b2/b2.go
swazrgb: Thanks for the suggestion. I pushed 2 new commits, one implementing the new `fs.Time`, and one to use it in b2. Now the parsing step in `newFs` is no longer necessary either.

Could you please let me know if this looks ok?

fs/parsetime.go
```diff
@@ -0,0 +1,106 @@
...
+	"strconv"
+	"time"
+)
+
+// Time is a time.Time with an offset, with some more parsing options
+type Time time.Time
+
+// For overriding in unittests.
+var (
+	timeNowFunc = func() time.Time { return time.Now() }
```
swazrgb, line 14: I had to add this mock so the `TestParseTimeUnmarshalJSON` & `TestTimeScan` could test duration offsets as well

Awbmilne:
Oh my god. You're an absolute hero.!
This is so awesome!
BackBlaze B2 Backups, HERE I COME!

fs/parsetime.go
ncw: This should just be

`timeNowFunc = time.Now`

fs/parsetime.go
```diff
@@ -0,0 +1,106 @@
...
+// IsSet returns if the time is not zero
+func (t Time) IsSet() bool {
+	return !time.Time(t).IsZero()
+}
+
+// parse the date as time in various date formats
+func parseTimeDates(date string) (t time.Time, err error) {
+	var instant time.Time
+	for _, timeFormat := range timeFormats {
+		instant, err = time.Parse(timeFormat, date)
```
ncw, line 34: This should be `time.ParseInLocation(timeFormat, date, time.Local)`

See the very recently fixed 887cccb2c17fa5f9c70ce32bc4c949e5906897a7

In fact you you'll want to update your tests based on that too.

I think you should probably get `parseDurationDates` to call this function so we don't duplicate the logic! I'd move this function to `parseduration.go` to be with the list of times also.

fs/parsetime.go
```diff
@@ -0,0 +1,106 @@
...
+		return time.Time{}, nil
+	}
+
+	now := timeNowFunc()
+
+	// Attempt to parse as unix timestamp
+	unixMillis, err := strconv.ParseInt(date, 10, 64)
+	if err == nil && unixMillis >= 0 {
+		return time.UnixMilli(unixMillis).UTC(), nil
+	}
```
ncw, line 54: Do we really need this? It isn't a good date format!

ncw (commented):
I think this is looking great!

I put a couple of small things to fixup inline.

Can you rebase on master, squash into two commits

1. addition of fs.Time
2. Changes to b2

Then force push please?

Hopefully that will be it :-)

fs/parsetime.go
swazrgb: The main reason I added this is because of the JSON unmarshalling being there, and whenever I encounter timestamps in JSON they are unix millis (`Date.now()` in javascript). But I'm not sure why the json unmarshalling is there, but I copied it from `fs.Duration`. Either way I'd be fine removing it, I don't see a real need for it otherwise.

How should I implement `UnmarshalJSONFlag`'s `setInt` method if not a unix milli? Just raise an error?

swazrgb:
@ncw Done! I did leave the unix milli parsing in there for now, but please let me know what you think about https://github.com/rclone/rclone/pull/6035#discussion_r840980866

swazrgb:
@ncw Could you take another look? Thanks!

fs/parsetime.go
ncw: Ah I see. The JSON parsing of fs.Duration as an integer is because it is returned as an integer number of Nanoseconds so needs to be parsed that way too. I think we can do without the integer parsing here, so I think we should remove this.

ncw (commented):
Sorry for the delay.

I think the CI shows a broken test and I left one comment inline about the `int` parsing.

If you can fix those up, rebase on master then I'll merge - thank you :-)

fs/parsetime.go
swazrgb: Alright, I've removed UnixMilli parsing in https://github.com/rclone/rclone/pull/6035/commits/e57cebdacb86ab2866e81811cad9458690bd2ac4

swazrgb:
@ncw No problem, thanks for the review! I've removed the `int` parsing in https://github.com/rclone/rclone/pull/6035/commits/e57cebdacb86ab2866e81811cad9458690bd2ac4

The test failure was dependent on the timezone of the machine running the tests. My machine is running `Europe/Amsterdam`, and CI is using `UTC` presumably. This affects the result of `time.ParseInLocation` when `time.Local` is used as you indicated in https://github.com/rclone/rclone/pull/6035#discussion_r840501255

On my machine with `TZ=Europe/Amsterdam` this returns `time.FixedZone("", 0)`, but when ran with `TZ=UTC` this returns `time.Local`. To make the test timezone independent I've switched to `time.Equal` instead in https://github.com/rclone/rclone/pull/6035/commits/bfa33026a3fd4ea4c7935fd6a6f6bf54982f4c48

It should pass now :crossed_fingers:

swazrgb:
Thanks for running the tests. Looks like I missed removing one UnixMillis test :sweat_smile: 

Removed that test & ran the tests locally on 1.16 as well now.

ncw (commented):

Can you squash it down into two commits, one for implementing `fs.Time` - that should come first, and one for the b2 changes

[fs: Implement fs.Time](https://github.com/rclone/rclone/pull/6035/commits/0051969e496de9d25d30513439ebfe7560167158) 
- [fs: Compare times using time.Equal instead of assert.Equal](https://github.com/rclone/rclone/pull/6035/commits/bfa33026a3fd4ea4c7935fd6a6f6bf54982f4c48) 

[b2: Add b2-version-at flag to show file versions at time](https://github.com/rclone/rclone/pull/6035/commits/de479c63ef7eebcce2d527e613609f938b697c75) 
- [b2: Use fs.Duration and only parse once in newFs](https://github.com/rclone/rclone/pull/6035/commits/f65ac77c17a714cc836db497bf768db493b99bd4)
- [b2: Use fs.Time for version-at](https://github.com/rclone/rclone/pull/6035/commits/7cae03bcf413d334fcede5b723f66a6891c2436d)
- [fs: Remove time.UnixMilli parsing](https://github.com/rclone/rclone/pull/6035/commits/d4a897b2957ef644bb0b1c3bb29545f4537e84af) 

Then I'll merge it.

I can do the above squashing if you would prefer (you can't do it though the github interface alas, so I'd pull it locally).

swazrgb:
@ncw Thanks, done!

swazrgb:
Nice, thanks! Unfortunately I don't really use s3 myself nor am I familiar with its API. 

While reading that issue I did come across #1532 which I'll mention here, since this PR covers a few of the use cases mentioned there (albeit just for B2), so the issues are linked together.

Awbmilne:
Would this work through a crypt layer?  
If I had `Crypt > B2`, Could I mount my Crypt volume using the --b2-version-at flag?  
There is support for using --b2-versions through Crypt in #5205.  
Im unfamiliar with Rclone's internals, so this may be intrinsic to how remote layering works.

ncw:
@Awbmilne I think it should - give it a go!

Issue #5429, opened 2021-06-27
Add Drive ID to rclone size output


chscott:
#### The associated forum post URL from `https://forum.rclone.org`

N/A

#### What is your current rclone version (output from `rclone version`)?

1.55.1

#### What problem are you are trying to solve?

I'm working on a tool that uses `rclone size` in generate a report of drive sizes. This is hard to do because the output doesn't have information that correlates the data to the drive. The end result is a log file of count:bytes pairs that don't indicate which drive they belong to.

#### How do you think rclone should be changed to solve that?

Modify rclone size to include the Drive ID. For example:

```
Drive ID: ABC
Total objects: 13629
Total size: 116.126 GBytes (124688889989 Bytes)
```

Or, with the --json flag:

```
{"id": "XYZ","count":1,"bytes":0}
```

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
Presumably at the time you call `rclone size` you know the ID of the drive? Can't you echo it into the log at that point?

```
echo "ABC" >> size.log
rclone size ABC: >> size.log
```

?

chscott:
I'm doing that already, but I need the information to be on the same line. Then it's all part of a single JSON object and can be parsed and handled accordingly.

ncw:
You could do it something like this

$ rclone size --json . | jq --arg id YOURID '. + {id: $id}'
{
  "count": 4936,
  "bytes": 417099838,
  "id": "YOURID"
}

Or if you want the output all on one line

```
$ rclone size --json . | jq -c --arg id YOURID '. + {id: $id}'
{"count":4936,"bytes":417099838,"id":"YOURID"}
```

chscott:
That's a good solution for Linux, but I'm using PowerShell functions and invoking Rclone via Start-Process, so I don't have a way of intermingling output in that manner.

ncw:
I'm sure you can read the JSON object from rclone into powershell, add things to it then write it out again.

I don't think this is a very useful feature to be adding to rclone as it is something that is easy for the caller to do.

chscott:
[...]

PR #6047, opened 2022-03-19
added DriveID and Path to size output and struct.

makinje16:
This change adds in more verbosity to `rclone size`. Specifically, we add in `DriveId` and `Path` to the result struct from information in the fs src.

This was asked for in #5429 and I saw it labeled as a good first issue.

#### Checklist


cmd/size/size.go
```diff
@@ -30,9 +30,13 @@ var commandDefinition = &cobra.Command{
 		cmd.Run(false, false, command, func() error {
 			var err error
 			var results struct {
-				Count int64 `json:"count"`
-				Bytes int64 `json:"bytes"`
+				Count   int64  `json:"count"`
+				Bytes   int64  `json:"bytes"`
+				DriveID string `json:"drive_id"`
```
albertony, line 35: Should we pick a different name for it?

cmd/size/size.go
```diff
@@ -42,6 +46,8 @@ var commandDefinition = &cobra.Command{
 			if jsonOutput {
 				return json.NewEncoder(os.Stdout).Encode(results)
 			}
+			fmt.Printf("DriveID: %s\n", results.DriveID)
+			fmt.Printf("Path: %s\n", results.Path)
```
albertony, line 50: I think outputting this information here is superfluous, json-only is enough.

albertony (commented):
Thanks!

This needs rebasing.

Also, as I understand the conclusion in the issue discussion:

> I'd suggest adding an output to the --json output called `fs` which is what was passed in on the command line.

This should only be part of the json output, and not the normal "human-readable" console output. And also the suggestion was to use property name "fs", I find that a bit less ambiguous than "DriveId", which I don't think is an "rclone term"? Sounds like a Google thing, if you ask me.. What do you think?

ncw:
I'm going to close this PR as it is stale - sorry we didn't get round to merging it.

If you want to resubmit with the extra info only in a --json output that would be great.

PR #6049, opened 2022-03-20
backend/http: parse get responses when no_head is set

Arnie97:

When the `no_head` flag is enabled for a http remote, the `ModTime` was always set to the 1970 UNIX epoch.
This patch support copying a file from the http remote with the correct `ModTime` and `MimeType` without the `HEAD` request.

Not yet, AFAIK.

#### Checklist

### DIFF
--- a/backend/http/http.go
+++ b/backend/http/http.go
@@ -601,23 +601,21 @@ func (o *Object) head(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to stat: %w", err)
 	}
-	return o.stat(ctx, res, true)
+	return o.stat(ctx, res)
 }
 
 // stat updates info fields in the Object according to HTTP response headers
-func (o *Object) stat(ctx context.Context, res *http.Response, isRangeRequest bool) error {
+func (o *Object) stat(ctx context.Context, res *http.Response) error {
 	t, err := http.ParseTime(res.Header.Get("Last-Modified"))
-	if err != nil {
-		t = timeUnset
+	if err == nil {
+		o.modTime = t
 	}
-	o.modTime = t
 
-	// TODO: parse Content-Range for total size
-	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
-	if !isRangeRequest {
-		o.size = parseInt64(res.Header.Get("Content-Length"), -1)
-		o.contentType = res.Header.Get("Content-Type")
+	size := rest.ParseSizeFromHeaders(res.Header)
+	if size >= 0 {
+		o.size = size
 	}
+	o.contentType = res.Header.Get("Content-Type")
 
 	// If NoSlash is set then check ContentType to see if it is a directory
 	if o.fs.opt.NoSlash {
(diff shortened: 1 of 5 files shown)


backend/http/http.go
```diff
@@ -601,13 +591,22 @@ func (o *Object) stat(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to stat: %w", err)
 	}
+	return o.stat(ctx, res)
+}
+
+// stat updates info fields in the Object according to HTTP response headers
+func (o *Object) stat(ctx context.Context, res *http.Response) error {
```
ncw, line 598: In other backends we call this function `decodeMetadata` which might be a better description of what it does, or maybe `decodeResponse`. I don't like `stat` because that implies to me that it is fetching the info like the `os.Stat` call.

backend/http/http.go
```diff
@@ -601,13 +591,22 @@ func (o *Object) stat(ctx context.Context) error {
...
 	t, err := http.ParseTime(res.Header.Get("Last-Modified"))
-	if err != nil {
-		t = timeUnset
+	if err == nil {
+		o.modTime = t
+	}
+
+	size := rest.ParseSizeFromHeaders(res.Header)
```
ncw, line 604: Note that this wants to set the size to `-1` if it isn't known to make rclone's unknown object size handling kick in.

backend/http/http.go
```diff
@@ -601,13 +591,22 @@ func (o *Object) stat(ctx context.Context) error {
...
 	}
+	return o.stat(ctx, res)
+}
+
+// stat updates info fields in the Object according to HTTP response headers
+func (o *Object) stat(ctx context.Context, res *http.Response) error {
 	t, err := http.ParseTime(res.Header.Get("Last-Modified"))
-	if err != nil {
-		t = timeUnset
+	if err == nil {
```
ncw, line 600: This needs to set the modtime to the default value if it doesn't parse - we don't want the old value left by accident.

backend/http/http.go
```diff
@@ -653,6 +652,12 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 	if err != nil {
 		return nil, fmt.Errorf("Open failed: %w", err)
 	}
+
+	if o.fs.opt.NoHead {
```
ncw, line 656: We should probably do this regardless of `o.fs.opt.NoHead` - that is what the other backends do - the response  fields should be accurate after a `GET` request.

lib/rest/headers.go
```diff
@@ -0,0 +1,37 @@
+package rest
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func ParseSizeFromHeaders(headers http.Header) (size int64) {
```
ncw, line 9: The lint needs fixing here

ncw (commented):
This looks very useful - thank you :-)

I put some suggestions inline. There is a lint error to fix also.

If you rebase on master you'll get a fix for the azureblob lint error.

backend/http/http.go
Arnie97: Make sense.  I can't figure out any scenario where `HEAD` returns the correct modTime while `GET` do not.

Arnie97:
I've rebased on top of the latest main branch, and improved the code according to the code view comments.  Please review it again.

Arnie97:
I've rebased on top of the latest main branch, and improved the code according to the code review comments. Please review it again, thank you.

Arnie97:
I've rebased on top of the latest master branch, and improved the code according to the code review comments. Please review it again, thank you.

ncw (approved):
Looks very nice - thank you.

I'll just run the CI and if that passes I'll merge - thank you :-)

PR #6061, opened 2022-03-25
Expanded docker volume plugin documentation

remygrandin:
This proposal expande the current docker volume plugin troubleshooting possible steps to inclue a state cleanup command and a reminder that a un/reinstall don't clean up those cache files.

This proposal expande the current docker volume plugin troubleshooting possible steps to inclue a state cleanup command and a reminder that a un/reinstall don't clean up those cache files.

Change triggered by an issue : https://forum.rclone.org/t/docker-plugin-fail-to-start-reinstall/29871/7

#### Checklist

### DIFF
--- a/docs/content/docker.md
+++ b/docs/content/docker.md
@@ -536,7 +536,7 @@ sudo curl -H Content-Type:application/json -XPOST -d {} --unix-socket /run/docke
 ```
 though this is rarely needed.
 
-If the plugin fail to work properly, and only In last resort after you tried diagnosing with the above methods, you can try clearing the state of the plugin. **Note that all exiting rclone docker volume will probably have to be recreated** (This might be needed as a reinstall don't cleanup existing state files to allow for easy restoration, as stated above.)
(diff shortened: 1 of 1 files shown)


docs/content/docker.md
```diff
@@ -536,6 +536,13 @@ sudo curl -H Content-Type:application/json -XPOST -d {} --unix-socket /run/docke
 ```
 though this is rarely needed.
 
+If the plugin fail to work properly, and only In last resort after you tried diagnosing with the above methods, you can try clearing the state of the plugin. **Note that all exiting rclone docker volume will probably have to be recreated** (This might be needed as a reinstall don't cleanup existing state files to allow for easy restoration, as stated above.)
```
albertony, line 539: `only In last resort` : Change to lowercase `i` - but even better, I think, it is to write `as a last resort`.
`Note that all exiting rclone docker volume` : `exiting` should be `existing`, I assume? And make plural `volumes`.
`to be recreated**`: Add ending punctuation, `to be recreated.**`

albertony (commented):
Some nit-picking on the spelling, otherwise text looks good to me. I cannot comment on the docker snippet, though...

Issue #4834, opened 2020-12-03
s3: copy ACL for an object from bucket to bucket"


zhucan:

zhucan:
@ncw @ivandeex

ncw:
So when doing a server side copy, you want to copy the ACL of the original object rather than set the ACL to that defined in the config - is that correct?

zhucan:
@ncw yes

ncw:
I think this would be relatively easy to implement - want to have a go?

Do you think it should be controlled by a flag, or do you think rclone should always do this?

zhucan:
> I think this would be relatively easy to implement - want to have a go?
Yes, I will try it.

> Do you think it should be controlled by a flag, or do you think rclone should always do this?
I think it should be controlled by a flag. After we uploaded the object, we can get the acl and then set the acl.

zhucan:
There are some question in my mind:
1.  Is it needed to compare acl, if acl is not same, it is needed to cover the dest object acl by src object acl ?
2. Is there one request to acquire metadata and acl, not to do more request.
3. If size and md5 are same, but mtime and acl are not same, It is needed to update mtime and acl.
4. we can set acl to object only when the owner of the bucket is same?

ncw:
> There are some question in my mind:
> 
> 1. Is it needed to compare acl, if acl is not same, it is needed to cover the dest object acl by src object acl ?

I think rclone should just copy the ACL or not. I don't think it is rclone's job to compare ACLs.

> 2. Is there one request to acquire metadata and acl, not to do more request.

It might be that skipping this line will cause the ACL to be copied

https://github.com/rclone/rclone/blob/6ab6c8eefa083202525d3969aec88279c031ae48/backend/s3/s3.go#L2168

> 3. If size and md5 are same, but mtime and acl are not same, It is needed to update mtime and acl.

Rclone will only update ACL when it copies an object. So whether that happens depends on the sync rules active.

> 4. we can set acl to object only when the owner of the bucket is same?

I don't understand this question.

zhucan:
> It might be that skipping this line will cause the ACL to be copied

I want copy the object with acl, but not assign the acl by user; for example, not like this: "rclone copy --acl ****" @ncw

ncw:
> > It might be that skipping this line will cause the ACL to be copied
> 
> I want copy the object with acl automatically, but not assign the acl by user; for example, not like this: "rclone copy --acl ****" 

Great.

Try removing this line


[...]

PR #6074, opened 2022-03-31
s3: add ACL comparison and synchronization for S3 storages by flag #4834

mrhakimov:
Add support for checking and synchronizing ACLs by flag (`--compare-acl`) for S3 storages. So, if flag is set, then `rclone sync` will copy ACLs from source objects to destination ones, and the same works for `rclone check` command

Yes, https://github.com/rclone/rclone/issues/4834

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -8,6 +8,7 @@ import (
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/hex"
+	"encoding/json"
 	"encoding/xml"
 	"errors"
 	"fmt"
@@ -32,7 +33,9 @@ import (
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/cloudfront"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/google/go-cmp/cmp"
 	"github.com/ncw/swift/v2"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config"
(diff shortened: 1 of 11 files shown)


ncw:
I see ACLs as part of a broader picture of metadata in general. For example S3 objects have this metadata (at least)

- `cache-control`
- `content-disposition`
- `content-encoding`
- `content-language`
- `content-type`
- `x-amz-tagging`
- `x-amz-meta-*`

Should these be included as part of a metadata inclusive sync?

What about metadata in other storage systems? acls/xattrs in local disk? Comments in google drive? It would be nice to be able to sync an object to s3 from your local disk and when you sync it back again it has the correct permissions.

As implemented your patch only syncs from s3 -> s3 at the moment doesn't it? Do the s3 ACLs map well enough to be implemented by GCS for example?

My vision (as yet unimplemented because it is too complicated probably!) is that each `fs.Object` has a `Metadata` method which returns a `map[string]string` of metadata.

On upload this would be read from the source object and applied to the destination object.

This metadata map would be in a standardised format so attributes the destination understood it would implement (eg ACL) and attributes it didn't it would just store as metadata.

There could be a sync mode where we fix up metadata, but not all metadata is writable without re-uploading the object.

I like your idea about having an `rclone check` mode which can check metadata and an `rclone sync` mode which can update it.

I'd like the definition of metadata to be more general though as as you've implemented it is very S3 specific.

Would you like to morph this into more general metadata support?

See https://github.com/rclone/rclone/issues/111 for more info.

mrhakimov:
> As implemented your patch only syncs from s3 -> s3 at the moment doesn't it?

Yes, it does. Well, this PR resolves an issues we faced when migrating data from one S3 storage to another, more specifically - it copies ACLs and allows users to access their data in a destination storage after migrating :)

> My vision (as yet unimplemented because it is too complicated probably!) is that each fs.Object has a Metadata method which returns a map[string]string of metadata.

I like this idea and it is worthy of implementation, but I am thinking of merging the current patch and then thinking of something more general. What do you think about it? Can we merge the current PR, maybe with some fixes (if needed)?

shanness:
Just putting in my 2c.  I only need S3 -> S3, and ACL's as said are only part of the metadata that is needed (and I wasn't aware it wasn't already supported).  It would be great if all metadata was synced (particularly ACL, content-type and x-amz-meta-*).  This would solve me waiting for #111 or having to move to s3cmd.

ncw:
We have a metadata framework now and there is a specific issue about preserving ACLs in it #8905

So sadly, I think, this PR has been overtaken by events and I will close it.

PR #6089, opened 2022-04-06
size: warn about inaccurate results when objects with unknown size

albertony:

Some backends do not provide object size and simply return -1, e.g. Google Docs documents in Google Drive backend. The reported size will ignore these, i.e. assume size 0. This PR adds information when this is the case.

This PR adds reporting of a notice log message:

```
2022/04/06 14:09:32 NOTICE: Google drive root '': Size may be underestimated due to 2 objects with unknown size
Total objects: 13 (13)
Total size: 27.125 MiB (28442166 Byte)
```

An alternative could be to include it in the output string:

```
Total objects: 13 (13)
Total size: 27.125 MiB (28442166 Byte) [Size may be underestimated due to 2 objects with unknown size]
```

I've also included the relevant information in the json and rc output, that is: The number of objects with unknown size. Don't know if this is something we want? I named the property `sizeless`, which is probably not the best name for it...!

By using log message alternative above this will be shown always, regardless if json output or not. If switching to the alternative 2 above (including the information in output string only), then called with option for json output this information would have to be included in the json (as it currently is) for user to get the information.

A third alternative could be to do it all: Log it as a notice log message, include it in the output string, and possibly also include in the json/rc output.

Related to #6085, which handles similar case for `rclone ncdu` command.

#### Checklist

### DIFF
--- a/cmd/size/size.go
+++ b/cmd/size/size.go
@@ -30,20 +30,26 @@ var commandDefinition = &cobra.Command{
 		cmd.Run(false, false, command, func() error {
 			var err error
 			var results struct {
-				Count int64 `json:"count"`
-				Bytes int64 `json:"bytes"`
+				Count    int64 `json:"count"`
+				Bytes    int64 `json:"bytes"`
+				Sizeless int64 `json:"sizeless"`
 			}
 
-			results.Count, results.Bytes, err = operations.Count(context.Background(), fsrc)
+			results.Count, results.Bytes, results.Sizeless, err = operations.Count(context.Background(), fsrc)
 			if err != nil {
 				return err
 			}
-
+			if results.Sizeless > 0 {
+				fs.Logf(fsrc, "Size may be underestimated due to %d objects with unknown size", results.Sizeless)
+			}
 			if jsonOutput {
 				return json.NewEncoder(os.Stdout).Encode(results)
 			}
 			fmt.Printf("Total objects: %s (%d)\n", fs.CountSuffix(results.Count), results.Count)
 			fmt.Printf("Total size: %s (%d Byte)\n", fs.SizeSuffix(results.Bytes).ByteUnit(), results.Bytes)
+			if results.Sizeless > 0 {
+				fmt.Printf("Total objects with unknown size: %s (%d)\n", fs.CountSuffix(results.Sizeless), results.Sizeless)
+			}
 			return nil
 		})
 	},
--- a/cmd/size/size.go
+++ b/cmd/size/size.go
@@ -24,6 +24,25 @@ func init() {
 var commandDefinition = &cobra.Command{
 	Use:   "size remote:path",
 	Short: `Prints the total size and number of objects in remote:path.`,
+	Long: `
+Counts objects in the path and calculates the total size. Prints the
+result to standard output.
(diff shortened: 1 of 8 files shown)


albertony:
Thank you for your feedback!

> Regarding the property name, maybe objsWithUnknownSize? Too long?

Maybe, not sure.. In json the existing properties are simply called "Count" and "Bytes", so wanted a concise name also for this - but maybe it is not possible. And also, I'm not sure if we even need or want this in the json. Your suggestion has been noted, but I think I'll wait and see if Nick has an opinion on the above first.

> Maybe we can include both the log and the output message. I think the log is harder to spot.

Sure. We have three places to include it:
- Log
- String output
- Json and rc output

Log sort of ensures it is always informed in some way, at least it should be possible for user to get the information if looking into logs one way or another. But the default size command returns human-readable string by default, so it does make sense to inform about it as part of the command output as well. 🤔

ncw (approved):
I think that looks nice - please merge, thank you :-)

zanna-37:
Are we adding the string in the output as well? (in addition to the log) I think it would be beneficial.

albertony:
Hard to find the perfect solution (as always).. I find that suggested alternative a bit too "messy":

```
Total size: 27.125 MiB (28442166 Byte) [Size may be underestimated due to 2 objects with unknown size]
```

Another alternative could be to included as a separate result property (when value is not 0), like it is with json/rc output:

```
Total objects: 13 (13)
Total size: 27.125 MiB (28442166 Byte)
Total objects with unknown size: 2
```

That does not explicitely tell user that size is underestimated, but that is implied? And also, could still report the notice log message.

I don't know if @ncw's approval meant he prefers the current implementation, or did not consider the alternatives, or have no strong opinion...?

albertony:
Also, @zanna-37: I guess this is a edge case limited to Google Drive/Photos, pehaps some very few other cases? For Google Photos it is documented here: https://rclone.org/googlephotos/#size. But in Google Drive backend docs it is not mentioned, pehaps it should. Do you want to do that? Or if not; do you know in which cases Google Drive does and does not return file sizes?

ncw:
> Hard to find the perfect solution (as always).. I find that suggested alternative a bit too "messy":
> 
> ```
> Total size: 27.125 MiB (28442166 Byte) [Size may be underestimated due to 2 objects with unknown size]
> ```

I think if you wanted to do this I'd just write

    Total size: 27.125 MiB (28442166 Byte) [2 objects with unknown size]

> Another alternative could be to include it as a separate result property (when value is not 0), like it is with json/rc output:
> 
> ```
> Total objects: 13 (13)
> Total size: 27.125 MiB (28442166 Byte)
> Total objects with unknown size: 2
> ```

I think this is the best solution and is matches up with the JSON.

> That does not explicitely tell user that size is underestimated, but that is implied? And also, could still report the notice log message.
> 
> I don't know if @ncw's approval meant he prefers the current implementation, or did not consider the alternatives, or have no strong opinion...?

My approval meant happy for you to merge what you have got already, though if you want to carry on refining it then that is fine too :-)

ncw:
This looks ready to merge if you want to merge it?

PR #6093, opened 2022-04-07
walk: cover edge case where a dir is also an object (e.g. via embedding)

miku:

Fix an edge case regarding [`ListType.Filter`](https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/fs/walk/walk.go#L98-L119).

Background: A prototypical backend has a relatively uniform abstraction which allows
to have a situation where a directory embeds an object.

    type Object struct ...

    type Directory struct {
        Object
    }

    ...

However, now a directory is also an object and would not be filtered
correctly, when using [`ListType.Filter`](https://github.com/rclone/rclone/blob/e43b5ce5e59b5717a9819ff81805dd431f710c10/fs/walk/walk.go#L98-L119).

This fix covers this edge case.

No.

#### Checklist

### DIFF
--- a/fs/walk/walk.go
+++ b/fs/walk/walk.go
@@ -100,18 +100,16 @@ func (l ListType) Filter(in *fs.DirEntries) {
 	if l == ListAll {
 		return
 	}
-	out := (*in)[:0]
+	var (
+		out          = (*in)[:0]
+		isObj, isDir bool
+	)
 	for _, entry := range *in {
-		switch entry.(type) {
-		case fs.Object:
-			if l.Objects() {
-				out = append(out, entry)
-			}
-		case fs.Directory:
-			if l.Dirs() {
-				out = append(out, entry)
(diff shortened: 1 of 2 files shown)


ncw:
I'm a bit concerned about this change...

```
$ git grep 'case.*fs.Object' | wc -l
37
```

Which means that there are approximately 37 places in the code where this might be a problem. Rclone very definitely assumes that something is either an `fs.Object` or an `fs.Directory`. Having something be both potentially breaks a lot of stuff!

Can you explain more why you think this might be necessary?

miku:
Thanks (also for rclone)! 

I'm working on a backend, where it seemed to be simpler to go with a single type for both objects and directories, but it not a big deal to separate the types - I wasn't aware of the various places this distinction plays a role. Sorry for the noise.

PR #6103, opened 2022-04-13
httplib: Allow CORS when serving

### LINKED ISSUE #5078: Allow CORS when serving (WebDAV)

rclone v1.54.0
- os/arch: linux/amd64
- go version: go1.15.7

I also use it through Termux on Android

rclone v1.54.0-DEV
- os/arch: android/arm64
- go version: go1.15.6

or through Arch installation in Termux.

Working with remote Org note files over WebDAV.

Long version:

To expand a little bit - I am slowly migrating from native mobile apps to PWAs and, in parallel, my info collection (notes, tasks, etc.) to Org mode. While Orgzly, a native Android app, is great, I believe that PWAs offer better opportunities, both to the developers and to the users. Therefore, I would like to use a PWA (https://organice.200ok.ch/) for the purpose of working with .org files.
This app supports working over WebDAV (since local file access in PWAs is still clunky). Unfortunately, my storage provider, while supporting WebDAV access, does not support CORS. Therefore, it is impossible to access the storage via WebDAV in any JavaScript-based browser application.
I was hoping rclone would come to the rescue with it's `rclone serve webdav` but unfortunately that provides exactly the same result - it doesn't support CORS. Or at least I can't find one in the docs and help texts.

Include CORS header `Access-Control-Allow-Origin: *` when serving files over WebDAV (and potentially other protocols).

Similar to #2575, I would really appreciate if CORS headers were supported (through an option parameter) when serving files, either via HTTP or WebDAV (`rclone server webdav`).

ralismark:

This fixes #5078 by adding basic support for [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) requests when serving http-based protocols. This means that webapps are able to send requests to this server even if they are on a different domain. The main usecase of this is to support webapps that don't manage their own storage but instead delegate to a WebDAV server, e.g. https://organice.200ok.ch/ from the linked issue.

- #5078 

#### Checklist

  - not sure how to test this so I've left this


ralismark:
Just noticed there's already some CORS code in rclone in `fs/rc/rcserver/rcserver.go`; I'll probably have a look at what that does and rework this PR.

ralismark:
Closing since https://github.com/rclone/rclone/commit/6c8148ef39ea51b1d4092128c598ac12c6fc1523 does the same thing.

PR #6106, opened 2022-04-15
.github: Add dependabot config for automatic dependency/action updates

dbast:

This enables the Github integrated dependabot service to automatically propose update PRs for go dependencies and github actions.

The update PRs not only propose the version bumps, but also provide the release notes / changelog / commits of the dependency as PR comment for an easy read to check what changed etc.

Comments within the PRs allow to instruct dependabot to ignore a dependency version or e.g. rebase a PR, like `@dependabot ignore this major version`.

The created PRs are similar to those created here by this PR https://github.com/dbast/rclone/pulls

#### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


ncw:
I usually bump all the dependencies once per release cycle using `make updatedirect`

https://github.com/rclone/rclone/blob/82e24f521f3dddb9909bb5f4b52dd69d96a04889/RELEASE.md?plain=1#L37-L54

and

https://github.com/rclone/rclone/blob/82e24f521f3dddb9909bb5f4b52dd69d96a04889/Makefile#L107-L115

I'm not sure I want the dependencies continuously updated as it makes stabilizing the release harder. Each dependency update stands a chance of breaking something so I'd rather fix all that breakage all at once.

So I'm leaning towards rejecting this pull request, but I'd like to hear what you have to say first :-)

dbast:
I would argue a bit different: Doing one by one dependency updates via dependabot has potentially advantages (at least for me in other projects):
* One by one updates allow for easier bisects in case something turns out being broken to identify the exact package update which cause the breakage. (If there are multiple updates per commit then bisect would only identify the entire big commit and trilling down to the single update is then an extra step).
* The one by one update PRs as proposed by dependabot contain a lot of extra information, like changelog and release notes from upstream, which both allow an easy read to keep up with potential important changes announced by upstream, without the need to go anywhere else except the PR to read that information.... (with a lot updates at once, finding and reading all the information tends to be often to much to read here)
* If I find new projects with dependabot already enabled, that gives me some confidence in the projects as one can assume that the projects gets alerted in time if there important security relevant dependency updates, which in the end help to keep the entire supply chain secure.

This is of course a personal preference, so feel free to just close the PR :)

PR #6108, opened 2022-04-16
webdav: add support for spec compliant modification time update

jpalus:

Add support for spec compliant modification time update to WebDav backend.
rfc4918 section 15.7 states that getlastmodified SHOULD be protected
(read-only) but does not mandate it. implementations are free to choose
whether it is possible to update modification time. one such
implementation operating in the wild that allows it is FastMail WebDav
access. since specification recommends property to be protected make it
disabled by default.

No

#### Checklist

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -125,9 +125,12 @@ You can set multiple headers, e.g. '"Cookie","name=value","Authorization","xxx"'
 			Default:  fs.CommaSepList{},
 			Advanced: true,
 		}, {
-			Name:     "update_modtime",
-			Help:     "Adjust modification time on servers which allow DAV:getlastmodified property update",
-			Default:  false,
+			Name: "update_modtime",
+			Help: `Adjust modification time on servers which allow DAV:getlastmodified property update.
+
+Use provider's default if unset.
+`,
+			Default:  fs.Tristate{},
 			Advanced: true,
 		}},
 	})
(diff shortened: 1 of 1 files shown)


ncw (commented):
I think this code looks great - thank you - and sorry for the delay in reviewing.

I wonder whether this would be better as a provider quirk, and adding a provider for FastMail?

How many other webdav providers support this do you think?

If you make the type of the option a `fs.Tristate{}` that will make overriding provider defaults easier (assuming we want to keep an option).

jpalus:
> I think this code looks great - thank you - and sorry for the delay in reviewing.

Thanks for review!

> I wonder whether this would be better as a provider quirk, and adding a provider for FastMail?

I was thinking about it as well but ultimately went for no quirk since that's essentially part of specification. I'm not insisting though, let me know if you prefer quirk.

> How many other webdav providers support this do you think?

Oh I can't really tell, it was hard enough to figure for FastMail already which is the only webdav provider that I use. Actually it's their support that made me aware of that (after they decomissioned FTP without prior notice).

> If you make the type of the option a `fs.Tristate{}` that will make overriding provider defaults easier (assuming we want to keep an option).

Done. I was considering `Tristate` myself but for different purpose:
* `true` means update modtime but fail if PROPPATCH fails
* `false` means don't try to update modtime
* `unset` means an `auto` mode: try to update modtime the first time and if PROPPATCH fails, swallow the error and continue as if it was `false`, otherwise as if it was `true`
But I guess it's not possible now or at least I couldn't figure what `Fs.precision` should become.

ncw (commented):
I think this is looking great now.

Do you want to (either as a separate commit in this PR or as a separate PR) make a provider for fastMail which sets the quirk automatically?

jpalus:
`fastmail` provider added in this PR which enables modtime update by default.

jpalus:
Pitty spec compliant modtime didn't land after all but instead vendor specific props did. No need to keep it open, if needed will maintain it for myself.

ncw:
@jpalus no reason why we should have both spec compliant and vendor specific modtime?

Issue #6115, opened 2022-04-19
Feature request: Autmatically use Larger chunk size for large files to Azure blob (stay below 50000 blocks)


lrvl:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/rclone-sync-copy-no-progress-on-one-large-641gb-file-to-azure-blob/30086/13

#### What is your current rclone version (output from `rclone version`)?

rclone v1.58.0

os/version: freebsd 12.0-release-p4 (64 bit)
os/kernel: 12.0-release-p4 (amd64)
os/type: freebsd
os/arch: amd64
go/version: go1.17.8
go/linking: static
go/tags: none

#### What problem are you are trying to solve?

Gain the ability to transfer large files to Azure without manual tweaks. The increase of the --azureblob-chunk-size to for example 100Mi should be set automatically to stay below the Microsoft provided Maximum number of blocks in a block blob or append blob.

#### How do you think rclone should be changed to solve that?

Re-use the S3 method

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

lrvl:
go test -v of backend/azureblob 

```
PASS
ok  	github.com/rclone/rclone/backend/azureblob	0.014s
```
Created pull request #6116

kaburkett:
I'm having grave difficulty on large (200 or 300 gb) Azure Blob file transfers when using mounted storage in windows. Is this autochunking feature already released? I'm not familiar with rclone I only use it for one small job and it works well for all the files under 100 gb but is really not handling the two largest files well.

ncw:
@kaburkett yes it should be working. Make sure you are using the latest version as there were lots of changes in azure blob recently. If you still can't get it to work then post details on forum.rclone.org thanks

PR #6116, opened 2022-04-19
azureblob: Calculate Chunksize/blocksize to stay below maxUploadParts

lrvl:

Adopt the automatic sizing of the ChunkSize/blocksize for large files to backend type Azureblob

https://github.com/rclone/rclone/issues/6115

#### Checklist


backend/azureblob/azureblob.go
```diff
@@ -1689,8 +1700,25 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 		}
 	}
 
+	// calculate size of parts/blocks
+	partSize := int(o.fs.opt.ChunkSize)
+
+	uploadParts := o.fs.opt.MaxUploadParts
+        if uploadParts < 1 {
+                uploadParts = 1
+        } else if uploadParts > maxUploadParts {
```
ncw, line 1709: We've made max_upload_parts a config item, but here we override it with `maxUploadParts` we have as a constant.

We should probably either not have the config item, or don't override its maximum value, since the point of having a config value is for microsoft to say - the limit is now 100,000 and people can update their configs to match.

Thoughts?

ncw (commented):
I think this looks very good - thank you.

I put a question inline about limits.

backend/azureblob/azureblob.go
lrvl: I lean towards that we should not introduce tuning of what has been provided as fixed scaling number.
Proceed by updating this branch to just to use the constant?

ncw: Remove the configurable sounds fine to me - that would be in line with the S3 backend for example.

ncw (commented):
It looks like it the file needs `go fmt` run on it.

If you can do that then I'll merge - thank you :-)

lrvl:
Did this make it into the release notes?

PR #6120, opened 2022-04-22
Jottacloud: Fix issue with server-side copy when destination is in trash

### LINKED ISSUE #6112: Jottacloud server-side copy fails when destination is a file in trash

While testing Jottacloud's API from personal projects, I found a strange case with the copy operation, i.e. post query parameter "cp" in Jottacloud API. If the destination file is in trash, the copy operation succeeds but the destination will still be in trash!

Knowing this, I investigated how it affects rclone. The Jottacloud backend has server-side copy using the same post query with "cp" parameter. Ran some commands with rclone to provoke the situation, and the result was that a `copyto` command where the destination path matches a file in trash fails with "Failed to copyto: object not found". This error message comes from the fact that rclone uses the response from the, apparently successful, copy request to `newObjectWithInfo`, and there it is checked if it is a `validFile`, which in turn checks `!info.Deleted` which obviously is false.

I would say this is a bug in Jottacloud's API. Being an unofficial API, the question is how should we handle it, work around it, in rclone.

rclone v1.59.0-DEV
- os/version: Microsoft Windows 10 Enterprise 21H2 (64 bit)
- os/kernel: 10.0.19044.1586 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.18.1
- go/linking: dynamic
- go/tags: cmount

Jottacloud

The following command sequence illustrates the case. Initially the copy succeeds. Then the destination is deleted by moving into trash. New copy fails. Emptying the trash and performing a new copy succeeds. Deleting the destination again, but this time without using trash, and then performing a new copy succeeds.

```
> rclone copyto remote:dir/file.txt remote:dir/copy/file.txt
> rclone deletefile remote:dir/copy/file.txt
> rclone copyto remote:dir/file.txt remote:dir/copy/file.txt
2022/04/18 22:58:41 ERROR : test.txt: Failed to copy: object not found
2022/04/18 22:58:41 ERROR : Attempt 1/3 failed with 1 errors and: object not found
2022/04/18 22:58:41 ERROR : test.txt: Failed to copy: object not found
2022/04/18 22:58:41 ERROR : Attempt 2/3 failed with 1 errors and: object not found
2022/04/18 22:58:41 ERROR : test.txt: Failed to copy: object not found
2022/04/18 22:58:41 ERROR : Attempt 3/3 failed with 1 errors and: object not found
2022/04/18 22:58:41 Failed to copyto: object not found
> rclone cleanup remote:
> rclone copyto remote:dir/file.txt remote:dir/copy/file.txt
> rclone deletefile remote:dir/copy/file.txt --jottacloud-hard-delete
> rclone copyto remote:dir/file.txt remote:dir/copy/file.txt
```

#### How to use GitHub

albertony:

From commit messages:

1. A strange feature (probably bug) in the api used by the server-side copy implementation in Jottacloud backend is that if the destination file is in trash, the copy request succeeds but the destination will still be in trash! When this situation occurs in rclone, the copy command will fail with "Failed to copy: object not found" because rclone verifies that the file info in the response from the copy request is valid, and since it is marked as deleted it is treated as invalid. This PR works around this problem by looking for this situation in the response from the copy operation, and send an additional request to a built-in deduplication endpoint that will restore the file from trash.

2. Now using the utility function for deduplication that was newly implemented to fix an issue with server-side copy. This function uses the original, and generic, "jfs" api (and its "cphash" feature), instead of the newer "allocate" api dedicated
for uploads. Both apis support similar deduplication functionaly that we rely on for the SetModTime operation. One advantage of using the jfs variant is that the allocate api is specialized for uploads, an initial request performs modtime-only changes and deduplication if possible but if not possible it creates an incomplete file revision and returns a special url to be used with a following request to upload missing content. In the SetModTime function we only sent the first request, using metadata from existing remote file but different timestamps, which lead to a modtime-only change. If, for some reason, this should fail it would leave the incomplete revision behind. Probably not a problem, but the jfs implementation used with this commit is simpler and a more "standalone" request which either succeeds or fails without expecting additional requests.

Fixes #6112

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -1599,38 +1599,17 @@ func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
 		return err
 	}
 
-	// prepare allocate request with existing metadata but changed timestamps
-	var resp *http.Response
-	var options []fs.OpenOption
-	opts := rest.Opts{
-		Method:       "POST",
-		Path:         "files/v1/allocate",
-		Options:      options,
-		ExtraHeaders: make(map[string]string),
-	}
-	fileDate := api.Time(modTime).APIString()
-	var request = api.AllocateFileRequest{
-		Bytes:    o.size,
-		Created:  fileDate,
-		Modified: fileDate,
-		Md5:      o.md5,
-		Path:     path.Join(o.fs.opt.Mountpoint, o.fs.opt.Enc.FromStandardPath(path.Join(o.fs.root, o.remote))),
-	}
-
-	// send it
-	var response api.AllocateFileResponse
-	err = o.fs.pacer.Call(func() (bool, error) {
-		resp, err = o.fs.apiSrv.CallJSON(ctx, &opts, &request, &response)
-		return shouldRetry(ctx, resp, err)
-	})
+	// request check/update with existing metadata and new modtime
+	// (note that if size/md5 does not match, the file content will
+	// also be modified if deduplication is possible, i.e. it is
+	// important to use correct/latest values)
+	_, err = o.fs.createOrUpdate(ctx, o.remote, modTime, o.size, o.md5)
 	if err != nil {
+		if err == fs.ErrorObjectNotFound {
+			// file was modified (size/md5 changed) between readMetaData and createOrUpdate?
+			return errors.New("metadata did not match")
+		}
 		return err
-	}
-
-	// check response
-	if response.State != "COMPLETED" {
-		// could be the file was modified (size/md5 changed) between readMetaData and the allocate request
-		return errors.New("metadata did not match")
 	}
 
 	// update local metadata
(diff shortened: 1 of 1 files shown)


backend/jottacloud/jottacloud.go
```diff
@@ -1253,6 +1292,12 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 	}
 	info, err := f.copyOrMove(ctx, "cp", srcObj.filePath(), remote)
 
+	// if destination was a trashed file then after a successfull copy the copied file is still in trash (bug in api?)
+	if err == nil && bool(info.Deleted) && !f.opt.TrashedOnly && info.State == "COMPLETED" {
```
ncw, line 1296: I wouldn't have thought you'd need the `bool()` around `info.Deleted` here since its underlying type is a `bool`.

ncw (approved):
Looks good to me.

Happy for you to merge provided integration tests are still passing!

backend/jottacloud/jottacloud.go
albertony: Without it: `invalid operation: err == nil && info.Deleted && !f.opt.TrashedOnly (mismatched types api.Flag and bool)`

ncw: Well I never!

PR #6128, opened 2022-04-26
go: tidy go.mod and go.sum files

aerfio:

Updating `go.mod` and `go.sum` files by running `go mod tidy` command.

See above

Didnt see one an issue, havent checked forum

#### Checklist

Feel free to update the name of PR if it doesnt suit any existing category, I looked through merged commits and PRs but couldnt find fitting category


ncw:
I think you'll find this will break the go1.16 build... I'll press the CI button and we'll find out!

aerfio:
Shouldnt we then switch to go@v1.16 in go.mod? I switched to it in go.mod in latest commit, even if you dont want to merge it I'd love if you run tests, just to satisfy my curiosity :D

ncw:
> Shouldnt we then switch to go@v1.16 in go.mod? I switched to it in go.mod in latest commit, even if you dont want to merge it I'd love if you run tests, just to satisfy my curiosity :D

I've given that a go.

It may be that is a better thing to do - I spent quite some time finding a go.mod that was acceptable to go1.18, go1.17 and go1.16 so if changing the version works then that is probably a good idea!

I can't remember why I wanted to change the version in the go.mod anyway - can you think of a reason why having go1.18 in the go.mod is advantageous over go1.16? Here are the docs: https://go.dev/ref/mod#go-mod-file

Perhaps the right thing to do is keep it on the lowest version of go rclone supports building with, in this case go1.16 and update it when we drop support for go1.16 (when go1.19 comes out most likely).

ncw:
I think you will need to manually add the `go mod download golang.org/x/mobile` package. Not sure why this doesn't happen automatically - any ideas?

ncw:
I have figured out, after quite a lot of messing around, how to add the golang.org/x/mobile as a dependency so `go mod tidy` doesn't remove it.

I think your idea of putting `go1.16` as the highest supported version in go.mod is a good one and with that change everything seems to be fixed.

So I'll run the CI and if that works I'll merge your patch, on top of mine. (I squashed and rebased your patch).

ncw:
Whee! Worked. Thank you for your help with this - lots of corner cases with go modules!

aerfio:
Happy to help! One more reason to switch to go 1.16 (at least in this case) that if someone used some feature from go1.17 or go1.18 like generics in PR, it would not compile with those older versions, so you put there the lowest version you want to support 

...or at least that's what I generally see everywhere - if someone knows better, please correct me 🙇🏻

ncw:
Yes I think it is the right thing for me to do to bump the go version in go.mod when deprecating go versions so it is always on the minimum go version rclone is willing to be compiled with. We used to support 4 versions, but I squeezed it to 3 in anticipation of rclone's dependencies using lots of go1.18 features.

I made a mistake last time though in thinking I wanted this to be the highest go version supported which caused quite a lot of trouble! Lesson learned :-)

Issue #4643, opened 2020-09-30
B2 backend should be able to automatically size chunks to fit in to 10,000 chunk limit


Adam-BZ:
#### What is your current rclone version (output from `rclone version`)?
1.53.1

#### What problem are you are trying to solve?
Just like Amazon S3, Backblaze B2 has a 10,000 chunk limit per file. Any user trying to upload a file larger than 937.5G (the default chunk size of 96M * 10000 chunks) has to manually configure a chunk size.

rclone's S3 backend supports automatically increasing chunk size to fit a file within that 10000 chunk limit, however rclone's B2 backend does not currently have that feature.

#### How do you think rclone should be changed to solve that?

Add the same automatic chunk size feature to the B2 backend, so instead of failing on files that are larger than chunk_size * 10000, it would increase chunk_size to accommodate them.

ncw mentioned there might be some concerns with memory usage since it may unpredictably increase depending on what file size rclone encounters, so perhaps this could be a non-default flag, or disabled with a flag if it is default. I think since the behavior between S3 and B2 is identical, any defaults or flags should be the same between the two backends if possible, just to keep a consistent user experience.

https://forum.rclone.org/t/b2-backend-dynamically-increasing-chunk-size/19288

ncw:
Here is the relevant bit of code in the s3 backend

https://github.com/rclone/rclone/blob/20b00db3909906d782c3ee4f3c0aa5cfb4a92f23/backend/s3/s3.go#L2794-L2808

And here is the equivalent code in the b2 backend

https://github.com/rclone/rclone/blob/20b00db3909906d782c3ee4f3c0aa5cfb4a92f23/backend/b2/upload.go#L97-L108

Fancy having a go at this @Adam-BZ ?

PR #6138, opened 2022-05-01
B2: Dynamically adjust --b2-chunk-size when necessary

Slugger:

Add support for dynamic modification of `--b2-chunk-size` when necessary as discussed in #4643.

Created a new package, `chunksize` that implements a chunk size calculator, which is then called by B2 to dynamically adjust the chunk size when needed.  This package was created with the intent to eventually leverage this calculator in all backends that need to do dynamic chunk size adjustments (S3 at a minimum).

Yes.  #4643
#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -26,6 +26,7 @@ import (
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/chunksize"
 	"github.com/rclone/rclone/fs/config"
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
@@ -1690,25 +1691,17 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 		}
 	}
 
-	// calculate size of parts/blocks
-	partSize := int(o.fs.opt.ChunkSize)
-
 	uploadParts := int64(maxUploadParts)
 	if uploadParts < 1 {
 		uploadParts = 1
 	} else if uploadParts > maxUploadParts {
 		uploadParts = maxUploadParts
 	}
-
-	// Adjust partSize until the number of parts/blocks is small enough.
-	if o.size/int64(partSize) >= uploadParts {
-		// Calculate partition size rounded up to the nearest MiB
-		partSize = int((((o.size / uploadParts) >> 20) + 1) << 20)
-		fs.Debugf(o, "Adjust partSize to %q", partSize)
-	}
+	// calculate size of parts/blocks
+	partSize := chunksize.Calculator(o, int(uploadParts), o.fs.opt.ChunkSize)
 
 	putBlobOptions := azblob.UploadStreamToBlockBlobOptions{
-		BufferSize:      partSize,
+		BufferSize:      int(partSize),
 		MaxBuffers:      o.fs.opt.UploadConcurrency,
 		Metadata:        o.meta,
 		BlobHTTPHeaders: httpHeaders,
--- a/backend/b2/upload.go
+++ b/backend/b2/upload.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rclone/rclone/backend/b2/api"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
+	"github.com/rclone/rclone/fs/chunksize"
 	"github.com/rclone/rclone/fs/hash"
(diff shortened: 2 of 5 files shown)


fs/dynchunksize/dynchunksize.go
```diff
@@ -0,0 +1,37 @@
+// Dynamically calculate suitable chunk size for large uploads
+package dynchunksize
```
ncw, line 2: I'm not a fan of this name!

Perhaps call the package `chunksize` and the function `Calculator`?

fs/dynchunksize/dynchunksize.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/rclone/rclone/fs"
+)
+
+const oneMiB = int64(1024 * 1024)
```
ncw, line 8: There is a constant for this already in `fs.Mebi` which has the correct type.

fs/dynchunksize/dynchunksize.go
```diff
@@ -0,0 +1,37 @@
...
+ possible chunk size that's going to allow the upload to proceed. Rounding up to the nearest MiB on the assumption
+ that some backends may only allow integer type parameters when specifying the chunk size.
+
+ Returns the default chunk size if it is sufficiently large enough to support the given file size otherwise returns the
+ smallest chunk size necessary to allow the upload to proceed.
+*/
+func ComputeChunkSize(fileSize fs.SizeSuffix, maxParts int, defaultChunkSize fs.SizeSuffix) fs.SizeSuffix {
+	requiredChunks := fileSize / defaultChunkSize
+	if requiredChunks < fs.SizeSuffix(maxParts) || (requiredChunks == fs.SizeSuffix(maxParts) && fileSize%defaultChunkSize == 0) {
+		fs.Debugf(nil, "size: %v, parts: %v, default: %v; returned default value", fileSize, maxParts, defaultChunkSize)
```
ncw, line 23: We should probably comment out this debug otherwise it will appear for every chunked upload.

fs/dynchunksize/dynchunksize.go
```diff
@@ -0,0 +1,37 @@
...
+		return defaultChunkSize
+	}
+
+	minChunk := fileSize / fs.SizeSuffix(maxParts)
+	remainder := minChunk % fs.SizeSuffix(oneMiB)
+	if remainder != 0 {
+		minChunk += fs.SizeSuffix(oneMiB) - remainder
+	} else if int64(fileSize/minChunk) == int64(maxParts) && int64(fileSize)%int64(maxParts) != 0 { // when right on the boundary, we need to add a MiB
```
ncw, line 31: I think you can probably get rid of most of the `int64()` casts in this line, as everything is `fs.SizeSuffix`.

I'm not convinced the `else if` is correct here... I think it should be (and would be safe if it was)

```go
    if remainder != 0 {
        minChunk += fs.SizeSuffix(oneMiB) - remainder
    }
    if int64(fileSize/minChunk) == int64(maxParts) && ...{
    }
```

fs/dynchunksize/dynchunksize.go
```diff
@@ -0,0 +1,37 @@
...
+		minChunk += fs.SizeSuffix(oneMiB)
+	}
+
+	fs.Debugf(nil, "size: %v, parts: %v, default: %v, new: %v; default chunk size insufficient, returned new chunk size", fileSize, maxParts, defaultChunkSize, minChunk)
```
ncw, line 35: Not having the object to log here will make these logs not very useful. So either pass the object to log in, or comment out the Debug.

ncw (commented):
Looks like a nice change :-)

I put some comments inline.

In order to accept the new package I'd like you to add a commit which uses it for `s3` and another ideally for `azureblob` - I don't like abstractions which don't get used more than once :-)

I'll run the CI on this too so you can see if it passes or not.

Slugger:
Thanks for the feedback. I've addressed all the comments except the implementation in additional backends.  I'll do that next time I can put some time in -- some evening hopefully later this week.

Just dipping my toes into the project here with this relatively simple feature request I found in the backlog. Got to say, so far so good.  It's always nice to get a super quick, constructive response when starting out in a new project. Looking forward to contributing (I have a big idea, which is what brought me here, but learning to crawl before I try to climb Everest! 😄 ).

Slugger:
Refactored `s3` and `azureblob` to leverage the new `chunksize.Calculator()`

ncw (commented):
Changes look great - thank you :-)

I'll run the CI now.

Can you squash into 4 commits, one for the chunksize lib, then one for each backend then I'll run the CI one more time and merge - thank you :-)

Slugger:
Linter issues fixed, commits squashed.  Everything should be ready to go.

ncw (approved):
Thank you - looks perfect. I'll run the CI one last time then merge :-)

ncw:
Thank you :-)

PR #6147, opened 2022-05-04
Pre verify sudo authorization "-v" before calling curl.

MCT-MIT-RCP:
The way this was, the curl interferes with the
	sudo and bash invocation. Here it will authenticate first.

Documentation "correction" showing the sequence of "verify sudo" then "use sudo" before any other commands would interfere with the input/output

No.

#### Checklist

- [ N ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ N/A ] I have added tests for all changes in this PR if appropriate.
- [ N/A ] I have added documentation for the changes if appropriate.
- [ ?" ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ N ] I'm done, this Pull Request is ready for review :-)


docs/content/install.md
```diff
@@ -26,11 +26,11 @@ using the [rclone selfupdate](/commands/rclone_selfupdate/) command.
 
 To install rclone on Linux/macOS/BSD systems, run:
 
-    curl https://rclone.org/install.sh | sudo bash
+    sudo -v ; curl https://rclone.org/install.sh | sudo bash
```
ncw, line 29: Maybe this should be

    sudo -v && curl ...

So we don't run the curl etc if the sudo fails?

ncw: I'm happy to defer to your experience here - I'm not a shell scripting expert.

So you recommend we merge as-is with the `;` and experiment after?

MCT-MIT-RCP: Yes, exactly. I also suspect that the behavior of `sudo &&` across platforms may not be consistent enough to rely on.

Random thought if there is a "Would be nice" task list:
"Create a test script that looks for some of the corner cases of sudo in this shell 'phrase'"

ncw: OK I'll merge what we;ve got so far - thank you.

Doing cross platform experiments with shell scripts is tricky indeed since I don't have access to all the platforms we support!

PR #6157, opened 2022-05-06
s3: Digital Ocean quirk: virtualHostStyle=false Fixes #6110

### LINKED ISSUE #6110: bug: copying files to DigitalOcean Spaces using rclone causes timeout

`Copying files to DigitalOcean Spaces causes timeout`

`rclone v1.58.0`

`ubuntu 20.4`

DigitalOcean

```
rclone copy /root/foo.txt --s3-access-key-id=<ACCESS_KEY> --s3-secret-access-key=<ACCESS_SECRET> --s3-provider=DigitalOcean --s3-endpoint=https://bucket.nyc3.digitaloceanspaces.com :s3:foo --error-on-no-transfer
```

```
2022/04/17 06:35:54 ERROR : Attempt 1/3 failed with 1 errors and: RequestError: send request failed
caused by: Head "https://foo.bucket.nyc3.digitaloceanspaces.com/foo.sh": x509: certificate is valid for *.nyc3.digitaloceanspaces.com, nyc3.digitaloceanspaces.com, not foo.bucket.nyc3.digitaloceanspaces.com
```

Now, if I run the following command the file gets copied without any issues. 

```
rclone copy /root/foo.txt --s3-access-key-id=<ACCESS_KEY> --s3-secret-access-key=<ACCESS_SECRET> --s3-provider=DigitalOcean --s3-endpoint=https://bucket.nyc3.digitaloceanspaces.com :s3:foo_bar --error-on-no-transfer
```

The only difference here is in the bucket name. The first one with issue uses a single word as bucket name which times out. The second one uses `_` in the bucket name and it makes it work.

Any help is much appreciated.

Slugger:

Fixes #6110 

Digital Ocean does not support virtual host style bucket access.  Their wild card ssl cert only goes one level deep.

#6110 

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2212,6 +2212,7 @@ func setQuirks(opt *Options) {
 		urlEncodeListings = false
 	case "DigitalOcean":
 		urlEncodeListings = false
+		virtualHostStyle = false
 	case "Dreamhost":
 		urlEncodeListings = false
 	case "IBMCOS":


ncw:
This presumably is only a problem with buckets with a dot in?

How does aws work around this problem? Does it really have a wildcard for each dot suffix?

Do digital ocean recommend path style or non path style? In aws there is a strong preference for non path style.

Slugger:
You're right.  This fix just masks the underlying problem and isn't correct.  The problem is when the folder name is also a valid hostname, rclone (on linux only, I could not reproduce this on BSD for example) submits a request to a hostname of `<folder>.<bucket>.nyc3.digitaloceanspaces.com`.  This is incorrect (even for AWS proper).  Virtual host style only goes upto the bucket name.  And it also explains why the underscore doesn't fail because underscores aren't valid in FQDNs so the AWS sdk isn't including it in the hostname for the request.

A deeper dive suggests this is _probably_ an upstream bug in the s3 sdk:

```
	setQuirks(opt)
	awsConfig := aws.NewConfig().
		WithMaxRetries(ci.LowLevelRetries).
		WithCredentials(cred).
		WithHTTPClient(client).
		WithS3ForcePathStyle(opt.ForcePathStyle). // Looks like the s3 sdk is generating the request headers, not rclone
		WithS3UseAccelerate(opt.UseAccelerateEndpoint).
		WithS3UsEast1RegionalEndpoint(endpoints.RegionalS3UsEast1Endpoint)
```

So looks like a bug with the s3 sdk since rclone isn't generating the requests itself.  Manually specifying `--s3-force-path-style` on your rclone command line would be a workaround until s3 fixes the sdk upstream.  But overriding the value in the quirks is incorrect.  What's interesting is that this seems to only be a problem on Linux.  I was originally trying to reproduce this on BSD and had no luck.  Then when I went to an Ubuntu 20.04 vm, easily reproducible.  So just a bug in the Linux AWS sdk?  Going to close this PR since this fix isn't correct.

PR #6184, opened 2022-05-19
Mct install set modes linux

MCT-MIT-RCP:

I tried to keep the changes specific for 'linux' installation target.

- Added target specific variables used for locations.
- Changed to not assume anything (if possible).
- Target directory creation is attempted. If successful, modes and ownership are set.
- Files are copied and ownership and modes are deterministically set.

***N.B.*** A minor change was implemented here.
 The original used the paths of `/usr/bin` and `/usr/local/share/man` as the targets when I believe it should be *either* `/usr/` or `/usr/local/` for both and not mixing in the use of the `local` part of the path without good reason.
I create the change using a simple variable pair which means it can be altered easily should it be needed.

No

#### Checklist

- [ n] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ n/a] I have added tests for all changes in this PR if appropriate.
- [ y] I have added documentation for the changes if appropriate.
- [ ?] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ y] I'm done, this Pull Request is ready for review :-)

***A 'caveat' about this change.***  
Since 'linux' encompasses so many flavors, there is a good chance of a quirk appearing. I attempted to remain as agnostic as possible but some commands may fail. I did test this on a Kubuntu and a CentOS79 system but cannot be sure of cases that will fail on other choices.

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -156,13 +156,13 @@ case "$OS" in
   'linux')
     # - - - - - - - - - - - - - - - - - - - -
     #binary
-	# First verify our sudo ability. If we pass that then execute the steps
-	# requested but make sure we're doing it as root and not as any other user.
-	# (If sudo was run before this, the user may be set to something *other*
-	# than 'root', this makes sure.
+    # First verify our sudo ability. If we pass that then execute the steps
+    # requested but make sure we're doing it as root and not as any other user.
+    # (If sudo was run before this, the user may be set to something *other*
+    # than 'root', this makes sure.
     sudo -v
     sudo -u root -- bash -c "\
-		mkdir -m 0755 ${binTgtDir} && \
+        mkdir -m 0755 ${binTgtDir} && \
         cp rclone ${binTgtDir}/rclone.new && \
         chmod 0755 ${binTgtDir}/rclone.new && \
         chown root:root ${binTgtDir}/rclone.new && \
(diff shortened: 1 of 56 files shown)


docs/content/install.sh
```diff
@@ -147,18 +150,32 @@ cd $unzip_dir/*
 
 case "$OS" in
   'linux')
+    # - - - - - - - - - - - - - - - - - - - -
     #binary
-    cp rclone /usr/bin/rclone.new
-    chmod 755 /usr/bin/rclone.new
-    chown root:root /usr/bin/rclone.new
-    mv /usr/bin/rclone.new /usr/bin/rclone
+    # First verify our sudo ability. If we pass that then execute the steps
```
ncw, line 155: Do we need to sudo in here?

The instructions say

    sudo -v ; curl https://rclone.org/install.sh | sudo bash

So we should be root?

ncw (commented):
See query about sudo inline

I get this from the github UI which I've never seen before! So I think it needs you to rebase.

> This branch cannot be rebased safely
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge.

docs/content/install.sh
MCT-MIT-RCP: Sorry, I obviously didn't do a good enough job putting the comments in.
The sudo is actually there. The critical parts that require root priv's occurs just after the snippet you showed it reads like this:
```
# First verify our sudo ability. If we pass that then execute the steps
# requested but make sure we're doing it as root and not as any other user.
# (If sudo was run before this, the user may be set to something *other*
# than 'root', this makes sure.
sudo -v
sudo -u root -- bash -c
```

The last two lines there are the enablers. First, verify (and authenticate) the ability to become root. then second line shown above says "Make sure you become root and then execute the following steps in the bash shell."

Should I go back and modify the comments to explain better?

MCT-MIT-RCP:
The error regarding needing to rebase, I'm not advanced enough to reliably do that without breaking something.

What would be preferred? I could see if I can fudge a solution?
(It might be smarter for me to make a new untouched branch and patch these changes into it and resubmit it.)

If anyone wants to offer pointers for me to understand and then correct it, by all means let me know.

docs/content/install.sh
ncw: Sorry, I didn't make my point very clearly. My point was  that if you follow the instructions then the script will already be running as root, so there is no need for `sudo` at all is there?

ncw:
I'm happy to fix the branch - it would require a git of expert git surgery which I've done lots of times before.

> (It might be smarter for me to make a new untouched branch and patch these changes into it and resubmit it.)

That would work too - I've done that before! You could use `git diff` and the `patch` utility.

ncw:
I'm going to close this since it is stale.

Sorry we never got round to merging it.

PR #6188, opened 2022-05-21
Add scripts to generate repos

### LINKED ISSUE #136: linux packages
Perhaps at some point you add can packages for, say, debian and rpm?

mwt:

This adds two scripts. One builds the apt repo and the other builds the yum repo. The scripts can be run using `make repo_apt` and `make repo_yum`.

Once done, this should close #136.

#### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,6 +190,12 @@ upload:
 upload_github:
 	./bin/upload-github $(TAG)
 
+repo_apt:
+	./bin/make_repo_apt.sh
+
+repo_yum:
+	./bin/make_repo_yum.sh
+
 cross:	doc
 	go run bin/cross-compile.go -release current $(BUILD_FLAGS) $(BUILDTAGS) $(BUILD_ARGS) $(TAG)
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -31,6 +31,8 @@ This file describes how to make the various kinds of releases
   * make upload
   * make upload_website
   * make upload_github
+  * make apt_repo
+  * make yum_repo
   * make startdev # make startstable for stable branch
   * # announce with forum post, twitter post, patreon post
 
--- a/bin/make_repo_apt.sh
+++ b/bin/make_repo_apt.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Upload a release
+#
+# Requires reprepro from https://github.com/ionos-cloud/reprepro
+
+set -e
+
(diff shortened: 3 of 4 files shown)


ncw:
This is looking very nice :-)

I think I'd like to install the dependencies separately rather than download them into the mount.

What is the best way to install https://github.com/ionos-cloud/reprepro ? I built a deb from its source because I couldn't find a repo, but I'm guessing there is one somewhere?

I used

```
apt install createrepo-c rpm
```

To get the yum deps.

The apt repo worked fine, but the yum repo has thrown up a bug in `rclone mount` - using `--vfs-cache-mode writes` I get this, whereas it works find with `--vfs-cache-mode full`.

```
gpg: can't open '/mnt/yum.rclone.org/i386/repodata/repomd.xml': Input/output error
gpg: signing failed: Input/output error
```

Which was caused by

```
2022/05/23 12:18:07 DEBUG : i386/.repodata/repomd.xml: >Open: fd=<nil *ReadFileHandle>, err=can't open file - writer failed
2022/05/23 12:18:07 DEBUG : i386/.repodata/repomd.xml: >Open: fh=<nil>, err=can't open file - writer failed
```

Which in turn was caused by the file being in the cache and not uploaded yet, which is something rclone should deal with.

Anyway, that's not relevant to this issue!

I've got the two repos hosted now

- https://apt.rclone.org/
- https://yum.rclone.org/

How do they look to you?

I've got lots of older versions I can pump in - do you think that is a good idea?

mwt:
@ncw The repo is mostly generating fine. However, your key is only dsa1024 which isn't accepted by current versions of Debian/Ubuntu  or the RHEL version I tried.

It might be a good idea to use a new key for this repo without changing your default key in gpg. You can use dsa4096 which will probably be accepted for a while.

GPG is also using SHA1 to sign both repos which I think is also not allowed. Do you have something like `personal-digest-preferences SHA1` in your config (`~/.gnupg/gpg.conf`)? You could probably fix this by changing it to `personal-digest-preferences SHA256`.

mwt:
Hi, I'm still available to help with this if you have an additional dsa4096 key that I can set the script to use. Additionally,
https://mirror.mwt.me/rclone/
is still up with 17 versions.

ncw:
Hi @mwt, and thank you for this contribution.

We're doing a tidy-up of the PR backlog. This one has had no activity for over three years, and rclone's code and docs have moved on quite a bit since it was opened, so I'm going to close it to keep the queue manageable.

Issue #6207, opened 2022-05-31
Verbose Environment Variable doesn't work?


19wolf:
#### What is the problem you are having with rclone?

Using the environment variable `RCLONE_VERBOSE=true` fails

#### What is your rclone version (output from `rclone version`)

```
# rclone version
2022/05/31 09:22:48 Invalid value when setting --verbose from environment variable RCLONE_VERBOSE="true": invalid argument "true" for "-v, --verbose" flag: strconv.ParseInt: parsing "true": invalid syntax
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Linux NepheleTower 5.15.43-Unraid #1 SMP Wed May 25 11:54:05 PDT 2022 x86_64 AMD Ryzen 5 2600 Six-Core Processor AuthenticAMD GNU/Linux

#### Which cloud storage system are you using? (e.g. Google Drive)

Google Drive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

rclone copy /mnt/user/Media GoogleDrive:Media/

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
# rclone copy -vv /mnt/user/Media GoogleDrive:Media
2022/05/31 09:24:28 Invalid value when setting --verbose from environment variable RCLONE_VERBOSE="true": invalid argument "true" for "-v, --verbose" flag: strconv.ParseInt: parsing "true": invalid syntax
````

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
Feel free to submit a PR to fix the docs.

PR #6208, opened 2022-05-31
Verbosity environment variable information

19wolf:
RCLONE_VERBOSE does not use a true/false setting, instead using a 0,1,2 setting. This updates the documents to reflect this information.

Update documentation for environment variable RCLONE_VERBOSE

[6207](https://github.com/rclone/rclone/issues/6207)

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -2234,7 +2234,7 @@ override the environment variable setting.
 Or to always use the trash in drive `--drive-use-trash`, set
 `RCLONE_DRIVE_USE_TRASH=true`.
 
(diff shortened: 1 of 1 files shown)


albertony:
That is a nice clearification in the docs, thank you! If you don't mind a couple of tiny nitpicks; the second change would be even better with a `.` at the end, and perhaps you could break the lines similar to how the rest of the file is?

19wolf:
> That is a nice clearification in the docs, thank you! If you don't mind a couple of tiny nitpicks; the second change would be even better with a `.` at the end, and perhaps you could break the lines similar to how the rest of the file is?

@albertony, I've fixed that up, and broke up a couple other long lines along the way.

Issue #6187, opened 2022-05-20
readability


C0deGyver:
#### Output of `rclone version`
rclone v1.58.1
- os/version: Microsoft Windows 10 Pro 21H2 (64 bit)
- os/kernel: 10.0.19044.1706 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.17.9
- go/linking: dynamic
- go/tags: cmount

#### Describe the issue
When running rclone config it is difficult to tell where one setting starts and another begins. It would be more readable if one or more blank lines were between "sections" (see example below). I know not everyone may feel this way... I am dyslexic so having no spacing around sections of text makes it very difficult to read and tell where I am.

I LOVE this utility and really appreciate the hard work that has gone into creating it!

example:
```
>rclone.exe config
2022/05/20 09:19:45 NOTICE: Config file "C:\\Users\\<user>\\AppData\\Roaming\\rclone\\rclone.conf" not found - using defaults
No remotes found, make a new one?
n) New remote
s) Set configuration password
q) Quit config
n/s/q> n
name> test
    
Option Storage.
Type of storage to configure.
Choose a number from below, or type in your own value.
 1 / 1Fichier
   \ (fichier)
 2 / Akamai NetStorage
   \ (netstorage)
 3 / Alias for an existing remote
   \ (alias)
```

ncw:
So would you prefer it to look like this?

```
 1 / 1Fichier
   \ (fichier)

 2 / Akamai NetStorage
   \ (netstorage)

 3 / Alias for an existing remote
   \ (alias)
```

I made the different section different colours to make them easier to see a while back.

Maybe

```
 1 / 1Fichier
   \ (fichier)
-----------------------------------
 2 / Akamai NetStorage
   \ (netstorage)
-----------------------------------
 3 / Alias for an existing remote
   \ (alias)
```

I did a prototype where you pressed up and down to choose the one you want from a scrolling menu which was cute, but I'm not sure it is the way to go either.

albertony:
With "section", do you mean the individual numbered list entries of multiple choice options, or the information related to individual configuration parameters, or both?

Some versions ago I made the config output a bit more condensed than before. I found it confusing when there was empty lines within the description of a parameter, it was easy to misread which parameter the information lines related to.

E.g. in older versions:

```
Files bigger than this can be resumed if the upload fail's.
Enter a size with suffix k,M,G,T. Press Enter for the default ("10M").
upload_resume_limit>
[...]

PR #6211, opened 2022-06-01
Config UI readability

albertony:

Improve readability in the config ui.

Primary intention was to add spacing (i.e. empty lines) between individual sections, such as individual options, or questions. But; improving one thing, makes another thing stand out, so I added some more changes to improve consistency and overall readability.

The changes are:
- Empty line between sections, i.e. before each individual option is presented, and before individual questions like "Edit advanced config?"
    - Not a very generic or elegant solution, that would require too much refactoring I think. But to make it obvious where newlines are inserted as part of this I created a separate helper function `newSection`. This should make it easy to read, avoid accidental removal of "duplicate" newlines, and make a good starting point for any future improved implementation.
- More consistent (and less "noisy") listing of configuration options at end of config session, as part of the "Yes this is OK" confirmation. Also at beginning when doing an "Edit existing remote" session.
- Added/extended some header labels to make the different type of sections (individual options, and questions like "Select remote", "Configuration complete" etc) consistent.

Note: The related issue also discusses readability of "multiple choice" listings, but I have not considered that in this PR. I think that is a separate issue, and in my opinion more debatable.

https://github.com/rclone/rclone/issues/6187

#### Checklist

    - There are two commits that I will squash before merge. They are doing similar things, but are separate because the first starts with a conservative approach and the next builds on that, in case we decide to revert to the more conservative approach.

### DIFF
--- a/fs/config/ui.go
+++ b/fs/config/ui.go
@@ -338,6 +338,11 @@ func OkRemote(name string) bool {
 	return false
 }
 
+// newSection prints an empty line to separate sections
+func newSection() {
+	fmt.Println()
+}
+
 // backendConfig configures the backend starting from the state passed in
 //
 // The is the user interface loop that drives the post configuration backend config.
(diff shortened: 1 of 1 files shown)


ncw (approved):
I like it :-)

PR #6220, opened 2022-06-07
Add huawei obs to s3 provider list 

mpb159753:

Add Huawei OBS to s3 provider list 

no

#### Checklist

- [y ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [n ] I have added tests for all changes in this PR if appropriate.
- [ y] I have added documentation for the changes if appropriate.
- [y ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ y] I'm done, this Pull Request is ready for review :-)


ncw (commented):
Code changes look great - thank you :-)

Did you run the integration tests?

Assuming you've got a remote called `huawei:` then you'd run `go test -v -remote huawei:` while in the `backend/s3` directory.

This will enable you to see if you have the quirks set right.

I'll give the CI a run now.

ncw:
The failing tests are

1. `FAIL: TestIntegration/FsMkdir/FsEncoding/dot_dot (7.60s)`
2. `FAIL: TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR `

The first is most likely a bug very minor bug in huawei obs - rclone can't create a path called `../..` - this is a legal key name in S3 but trips up many implementations as it looks like an HTTP path.

The second, I'm not sure about

```
=== RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR
    fstests.go:1562: 
                Error Trace:    fstests.go:1562
                Error:          Received unexpected error:
                                directory not found
                Test:           TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR
    fstests.go:1565: 
                Error Trace:    fstests.go:1565
                Error:          Should be true
                Test:           TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR
                Messages:       file1Root not found
    fstests.go:1566:
                Error Trace:    fstests.go:1566
                Error:          Should be true
                Test:           TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR
                Messages:       file2Root not found
```

This is testing all the files are present if you do `rclone lsf -R obs:` - what happens when you do that by hand? It might be that you have buckets in your account that can't be listed by rclone, in which case it probably isn't anything to worry about either.

mpb159753:
I have run several times，the first test case passed, and the ListR is still fail.

Maybe it's a common case for s3? I tested for Alibaba, also failed this test case

ncw:
Can you run

    rclone lsf -R obs: -vv

That should tell you what is happening. I expect it will be a permissions error.

mpb159753:
> Can you run
> 
> ```
> rclone lsf -R obs: -vv
> ```
> 
> That should tell you what is happening. I expect it will be a permissions error.

All test case are passed now, it may be because my network is unstable

mpb159753:
Find another bug, in Huawei OBS, they have two type bucket, normal bucket like other providers, and Parallel File System(PFS) which support POSIX protocol ([About Parallel File System](https://support.huaweicloud.com/intl/en-us/pfsfg-obs/obs_13_0007.html))

But PFS not support listObjectsV2, and when I set `urlEncodeListing` to `true`, the `marker` parameter will be  encoded twice.

For example, when I turn on `urlEncodingListing`, the raw next marker from server is  `figure%2Fcustomer%2F` which means I need to use `figure/customer/` as the marker for next listing. 

But when rclone send request, the marker change to `figure%252Fcustomer%252F`, and now it's stuck here, keep listing the same page forever.

I have two option now:
* Set `urlEncodeListing`  to false in quirks, and user may encounter problems when their objects have control chars; 
* Just leave it alone and add a hint to document, tell user to set `urlEncodeListing` to `fasle` when they are using PFS.

Which do you think is best for user?

mpb159753:
There is actually a third option——not put HuaweiOBS under S3 provider and use their own SDK write the full backend, but I think it's too much for this case.

ncw:
> I have two option now:
> 
> * Set `urlEncodeListing`  to false in quirks, and user may encounter problems when their objects have control chars;
> * Just leave it alone and add a notice to document, tell user to set `urlEncodeListing` to `false` when they are using PFS.
> 
> Which do you think is best for user?

I think I would go for "Set `urlEncodeListing`  to false in quirks" and document that if the user is **not** using PFS **and** they have control characters in their filenames (unlikely) then then should use the flag `list_url_encode = true`.

> There is actually a third option——not put HuaweiOBS under S3 provider and use their own SDK write the full backend, but I think it's too much for this case

Writing and maintaining a backend is a lot of work - if we can get it to work with S3 then I think that is best.

mpb159753:
> > I have two option now:
> > 
> > * Set `urlEncodeListing`  to false in quirks, and user may encounter problems when their objects have control chars;
> > * Just leave it alone and add a notice to document, tell user to set `urlEncodeListing` to `false` when they are using PFS.
> > 
> > Which do you think is best for user?
> 
> I think I would go for "Set `urlEncodeListing` to false in quirks" and document that if the user is **not** using PFS **and** they have control characters in their filenames (unlikely) then then should use the flag `list_url_encode = true`.
> 
> > There is actually a third option——not put HuaweiOBS under S3 provider and use their own SDK write the full backend, but I think it's too much for this case
> 
> Writing and maintaining a backend is a lot of work - if we can get it to work with S3 then I think that is best.

I agree with you, and just set in quirks, also add some notice in doc, now it's ready for merge.

ncw (approved):
This is looking great - thank you. I'll merge it when the CI passes (hopefully in a few minutes!)

Issue #6141, opened 2022-05-02
Is a RealDebrid integration possible?


itsToggle:
#### What problem are you are trying to solve?

I would like to rclone my RealDebrid Downloads. (So that I can use Plex to serve the files to family members)

#### How do you think rclone should be changed to solve that?

The RealDebrid api only allows for files to be downloaded, not uploaded and is very rudementary overall.
Would it be possible to make a read-only backend, that simply serves the downloaded files?
Ive had a look into the backend of premiumize, which is a pretty similar service. 
If its too niche of a problem, could you guys maybe give me some pointers so that I can try to implement the backend myself?
Thanks!

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
Very doubtful as it's not a cloud backend but a download tool.

https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend

Can tag as a new backend, but unless you plan to implement it, I'll bet it just sits since there are so many requests for new backends without much demand/anyone to actually implement them.

itsToggle:
I see.. well thank you, i'll have a go at writing it myself then :)

animosity22:
ncw is very good at asnwering/helping as well.

Your request is the 56th new backend that's open right now to give an idea :)

itsToggle:
Might be faster for me to learn Go then haha. If you guys do find the time:
Ive rewritten some of the "premiumizeme" backend to instead call realdebrid's api. After a bit of tinkering I can now `ls` and `lsd` the files :) 
Since I dont have the "mount" method to play with (seems to be omitted from self build versions), Ive instead gone with
`serve webdav remote:` to mount the remote. 
The explorer now correctly shows the root directory. When opening a 30B .txt file, the backends `Open()` function is correctly called and the file is opened. When trying to open a large .mkv or .rar file though, the `Open()` function is not called. Instead, `readMetaData()` is called a bunch of times (without giving an error) and program that was supposed to open the file gives an error. Any idea what could be going wrong?

itsToggle:
[...]

PR #6224, opened 2022-06-08
remote for real-debrid.com

itsToggle:

This change adds a remote for real-debrid.com, a 'cloud-storage' service much like premiumizeme (premiumize.me).

Its based on the premiumizeme remote, since the services are similar. Realdebrid is, however, entirely read-only (except for file-deletion). 

Eventhough the realdebrid remote is based on the premiumize remote, there are a number of changes to the 'types.go' and 'remote.go' files - some of which are a bit 'hacky', since realdebrid is not designed to be a filesystem:

- The 'types.go' file has been adapted to fit responses from realdebrid:
    - the 'Item' struct has been adapted to unmarshall realdebrid responses correctly
- The 'remote.go' file has been adapted to work with realdebrid:
    - the init function now asks for an api key, as realdebrid does not support oauth (they say they do, but there is no way to register a new app)
    - the listall function has been adapted to work with realdebrid:
        - user files on realdebrid are located on the realdebrid /torrents page.
        - these files can only be accessed, if the files are 'unrestricted' first. Unrestricting a file produces a direct link to the file, which is then stored on the realdebrid /downloads page. These links are only valid for 7 days.
        - realdebrid provides user files, aswell as these direct links, in pages of 100 items per page. To reduce the amount of calls made to the api, two global variables have been added that store both the list of files and the list of direct links. The lists are only updated entirely when changes in their total length are detected.
        - When listing a users files, the remote checks the list of direct links for a corresponding link. If a file does not have a corresponding link, the remote automatically re-'unrestricts' the file, allowing it to be accessed by the file system.
    - some functions that require write priviliges have been removed
    - the remove function/s have been adapted to remove both the file from the /torrents page, aswell as the corresponding link from the /downloads page

As discussed in #6141 I have not included a "remote_test.go" file.

The integration of realdebrid has been discussed in issue #6141 as stated above.

#### Checklist


buengese (changes requested):
I'm really not sure if this suited for inclusion in rclone. Its extremely limited and doesn't really match rclone's api. 

There is also a ton of stuff from the premiumize backend in here which you should probably remove.

itsToggle:
Yeah I thought so.. Its not really a traditional cloud storage service. It is extremely usefull though, since realdebrid has unlimited traffic and unlimited (torrent-) storage for next to no money.

Im not sure how to remove some of the leftover functions from premiumize. Can I limit the remotes capabilities somewhere, and then delete the write functions entirely? If not, the write-functions should return some 'not capable' error which I don't know how to implement either. I havent found another remote thats read-only in the way this one would be.

itsToggle:
Realdebrid just announced that they will implement webdav support soon, so there will not be any need for a separate rclone remote. Thanks again for the help, it's been fun learning go :)

PR #6226, opened 2022-06-08
Hashsum: Changed output format when listing algorithms

albertony:

Suggestion for a supermicro change.... switching the list output format of `rclone hashsum` command from `*` prefixed (with indention) to `-` prefixed (without indention).

Only reason to do this is consistency, or perhaps also readability? (Or maybe actually less readable?) Somewhat related to https://github.com/rclone/rclone/pull/6211. In docs (markdown) we ususally use `-` prefixed list entries, and therefore this is also what is presented in help command output and config ui. The `hashsum` command  (without arguments) output information (list of supported algorithms) in "human-readable markdown-like list format", similar to help and config commands, and therefore I feel it is more consistent to use the same `-` prefixed list format also for this. Not sure what other commands have similar output? There are commands that output list-like information, such as `listremotes`, but it outputs it as plain newline-separated strings. The difference is that this is the "primary and intended output" of the command. The `hashsum` command is a bit different, perhaps, as the output is more of a help text? (Or else it could also just list without the header text and without list item prefixes.)

Current (before this PR):
```
$ rclone.exe hashsum
Supported hashes are:
  * md5
  * sha1
  * whirlpool
  * crc32
  * sha256
  * dropbox
  * mailru
  * quickxor
```

New (after this PR):
```
$ rclone.exe hashsum
Supported hashes are:
- md5
- sha1
- whirlpool
- crc32
- sha256
- dropbox
- mailru
- quickxor
```

#### Checklist

### DIFF
--- a/fs/hash/hash.go
+++ b/fs/hash/hash.go
@@ -410,7 +410,7 @@ func HelpString(indent int) string {
 	help.WriteString(padding)
 	help.WriteString("Supported hashes are:\n")
 	for _, h := range supported {
-		fmt.Fprintf(&help, "%s  * %v\n", padding, h.String())
+		fmt.Fprintf(&help, "%s- %v\n", padding, h.String())
 	}
 	return help.String()
 }


ncw (approved):
The `-` will cause it to render as a markdown list properly, so get my vote!

I wonder if anyone is parsing the output of this?

PR #6234, opened 2022-06-10
Changed handling of volume name (Windows and OSX)

albertony:

Fixes an issue on Windows where mounting the local filesystem in network mode failed when not using option --volname. Reason was that the volume name in network mode is a network share path in the basic UNC format, and characters that are invalid in regular file and directory names (as documented [here](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)) are also invalid in such a path. And the default volume name would typically include a '?', which is invalid, from the unc path of the local, e.g. "C:\Temp" is converted to "//?/C:/Temp" in rclone, and default volname would be created as "\\server\\? C  Temp".

The fix is to use an encoder to encode invalid characters such as '?' with the unicode equivalent, similar to how rclone encodes filesystem paths in normal operations, when mounting in network mode. (But not the same ToStandardPath/FromStandardPath since we here are not really dealing with regular paths the same way). Also performs some automatic cleanup of path separators, but in general, tries to be conservative on restrictions, and instead rely on user changing their --volname to something realistic. These changes are added to the windows specific part of the cmount specific implementation.

The previous strategy to replace the two characters ':' and '/' with space, regardless of mount implementation and mounting mode, was removed. On Windows/cmount in network mode the new approach handles these in a better way. Also the existing method did not apply at all when using the implicit network mode where volume names are taken from mountpath instead of volname option (`rclone mount remote:path/to/files \\cloud\remote`). For non-network mode they were not needed. Also I think it is better to move the os- and mount-implementation specific handling into the individual mount implementations, instead of here in common mountlib as it was. 

Default volume names, when not specified by user, will be different with this change. In general the remotes and paths is shown  more similar to how they are used in rclone.

Some quick examples, showing the resulting labels in Windows ("Prev" is previous versions of rclone, "New" is after this PR):
```
rclone.exe mount C:\Temp X:
	--> Prev: ? C  Temp (X:)
	--> New: //?/C:/Temp (X:)
rclone.exe mount C:\Temp X: --network-mode
	--> Prev: ? C  Temp (\\server)
		- Fails!
	--> New: Temp (\\server\？\C：)
rclone.exe mount C:\Temp \\my\server:2\path!
	--> Prev: path! (\\my\server:2)
		- Fails!
	--> New: path! (\\my\server：2)
rclone.exe mount C:\Temp X: --volname "abcdefghijklmnopqrstuvwxyzzzz123456789xxxyyyzzz111222333"
	--> Prev and New: abcdefghijklmnopqrstuvwxyzzzz123 (X:)
rclone.exe mount C:\Temp X: --volname "abcdefghijklmnopqrstuvwxyzzzz123456789xxxyyyzzz111222333" --network-mode	
	--> Prev: abcdefghijklmnopqrstuvwxyzzzz (\\server)
	--> New: abcdefghijklmnopqrstuvwxyzzzz123456789xxxyyyzzz111222333 (\\server)
rclone.exe mount C:\Temp X: --volname "  hello  " --network-mode
	--> New: hello   (\\server)
	(Prev: Same but with spaces were removed)
rclone.exe mount C:\Temp X: --volname "  hello  "
	--> New: hello  (X:)
	(Prev: Same but with spaces were removed)
rclone.exe mount remote:Temp X:
	--> Prev: remote Temp
	--> New: remote:Temp
rclone.exe mount jfs:Temp X: --network-mode
	--> Prev: remote Temp (\\server)
	--> New: remote：Temp (\\server)
```

https://forum.rclone.org/t/cannot-mount-remote-of-type-local-in-network-mode/

#### Checklist


cmd/mountlib/utils.go
```diff
@@ -89,18 +89,7 @@ func (m *MountPoint) SetVolumeName(vol string) {
 	if vol == "" {
 		vol = fs.ConfigString(m.Fs)
 	}
-	m.MountOpt.SetVolumeName(vol)
-}
-
-// SetVolumeName removes special characters from volume name if necessary
-func (o *Options) SetVolumeName(vol string) {
```
ncw, line 96: This patch seems to have removed all the checking for the other OSes?

If we just want a valid name, then I would have thought using [encoder.OS](https://github.com/rclone/rclone/blob/master/lib/encoder/os_windows.go#L26) here and `.FromStandardName()` (not `Path`) would be enough wouldn't it?

I don't actually know what the volume names can be in macOS/FreeBSD though.

ncw (commented):
The first commit seems very complicated to fix what seems to be a simple problem?

I haven't reviewed the second commit.

The third looks OK though

albertony:
Using encoder.OS.FromStandardName seem to work, yes. I don't remember if I just missed it or wasn't happy with the results. It will do a bit extra unnecessary encoding, e.g. encode slash into unicode slash, and ending periods (including names `.` and `..`) into unicode periods, so I think I wanted to be specific about the required encoding for network share path. E.g. given a case where fs.ConfigString returns `//?/C:/Temp/Temp`, the default volume name would be:
- In current master: `\\server\? C  Temp Temp` (illegal)
- This PR: `\\server\？\C：\Temp`
- With encoder.OS.FromStandardName: `\\server\／／？／C：／Temp／Temp`

Also in my commit, something adding to the complexity, is that the encoding is deliberately done only for windows network mode. Volume name for windows network mode is used as a network share path, which is more or less a file system path, and needs cleanup/encoding. While volume names for drives I think are more of a generic string label, and on Windows basically anything goes. That's why I didn't want to put it into the SetVolumeName function, which would lead to encoding everything.

If you prefer using the OS encoder, I can look into that.  But I think using it for network mode only, is still the best approach. This also means it will still not be simple one-liner fix, even if there will be less changes than with the current commit. Also; some of the changed lines in the commit are added error checks, comments, and I see some of the lines was a due to a small refactoring (reducing complexity) of the logic around assignment to opt.VolumeName, which I found made it easier to implement the encoding when in network mode only etc.

Replacing `:` and `/` with space for non-windows OS-es was probably a mistake, and can be put back. You know for a fact that `:` and `/` is illegal in linux?

Do you want me to split the PR into 3, one for each commit?

ncw:
Sorry we stalled on this commit!

> Using encoder.OS.FromStandardName seem to work, yes. I don't remember if I just missed it or wasn't happy with the results. It will do a bit extra unnecessary encoding, e.g. encode slash into unicode slash, and ending periods (including names `.` and `..`) into unicode periods, so I think I wanted to be specific about the required encoding for network share path. E.g. given a case where fs.ConfigString returns `//?/C:/Temp/Temp`, the default volume name would be:
> 
> * In current master: `\\server\? C  Temp Temp` (illegal)
> * This PR: `\\server\？\C：\Temp`
> * With encoder.OS.FromStandardName: `\\server\／／？／C：／Temp／Temp`

I think if you want to use the specialised encoder that is fine since the rules are obviously different here which I didn't realise :-)

> Also in my commit, something adding to the complexity, is that the encoding is deliberately done only for windows network mode. Volume name for windows network mode is used as a network share path, which is more or less a file system path, and needs cleanup/encoding. While volume names for drives I think are more of a generic string label, and on Windows basically anything goes. That's why I didn't want to put it into the SetVolumeName function, which would lead to encoding everything.

:+1:

> If you prefer using the OS encoder, I can look into that. But I think using it for network mode only, is still the best approach. This also means it will still not be simple one-liner fix, even if there will be less changes than with the current commit. Also; some of the changed lines in the commit are added error checks, comments, and I see some of the lines was a due to a small refactoring (reducing complexity) of the logic around assignment to opt.VolumeName, which I found made it easier to implement the encoding when in network mode only etc.
> 
> Replacing `:` and `/` with space for non-windows OS-es was probably a mistake, and can be put back. You know for a fact that `:` and `/` is illegal in linux?

Linux doesn't use volume names, but macOS does and I'm pretty sure `:` and `/` are illegal there. I can't find any docs about this though!

> Do you want me to split the PR into 3, one for each commit?

I think that will make them easier to review so its probably a good idea.

albertony:
> Sorry we stalled on this commit!

That was my fault!

> > Do you want me to split the PR into 3, one for each commit?
> 
> I think that will make them easier to review so its probably a good idea.

Sure. But then I'll put one of them (the most important fix) into https://github.com/rclone/rclone/pull/6712, assume that was what you meant in your comment there. And then the third commit is just a small improvement you've already approved, so maybe I'll just push that and reduce the admin work a bit. That leaves one change, the one related to directory mount checks, which could stay in this PR (or maybe I'll close it and create a new referring to it, I'll decide when I start the work).

albertony:
Even though I just said I wouldn't bother... I have now split the third commit (regarding UNC paths) into separate PR: https://github.com/rclone/rclone/pull/6713. Will see that the build is green, and then just merge it.

ncw:
Grand! Give me a ping when you want me to review again :-)

albertony:
Split the other two commits into separate PRs as well (or moved the one commit into another existing PR as discussed above), so now there is:
 - https://github.com/rclone/rclone/pull/6712
 - https://github.com/rclone/rclone/pull/6713
 - https://github.com/rclone/rclone/pull/6715

This PR is therefore obsolete, but contains detailed description that I just refer to from the others. I will keep the PR and branch for a while, also as a backup, but they are safe to delete (not to self).

PR #6236, opened 2022-06-12
Use variable type int instead of time.Duration for keeping number of seconds

albertony:

This started out as an issue reported by the staticcheck linting tool: "Poorly chosen name for variable of type time.Duration" (ST1011). It refers to the `serverPool.expirySec` variable. I fixed this by renaming it from `expirySec` to just `expiry`. But this revealed another problem: It is initialized with value of constant `serverExpirySec`, integer 180, but despite the name of this constant its value in this assignment will be treated as nanoseconds! I fixed this by multiplying with `time.Second`, which makes the code/naming consistent, and I assume this is according to the intention ❓ But it changes the actual logic since the initial value is now 180 seconds while before was 180 nanoseconds ❗

#### Checklist

### DIFF
--- a/backend/mailru/mailru.go
+++ b/backend/mailru/mailru.go
@@ -2277,7 +2277,7 @@ type serverPool struct {
 	pool      pendingServerMap
 	mu        sync.Mutex
 	path      string
-	expirySec time.Duration
+	expirySec int
 	fs        *Fs
 }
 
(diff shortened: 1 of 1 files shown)


albertony:
> The original author was carfeful to name any times in seconds with the suffix Sec and you can see if you run though the code that these are always multiplied by time.Second before use.

Yes, I did notice that. Made the code easy to understand. Except that one `expirySec time.Duration`...

> I think the code is actually correct.

Ah, that is true, I see it now after re-reading it - a couple of times! It confused me; storing the number of seconds as time.Duration, which I read as nonseconds, but since it is scaling it back to seconds the one place where it is used it got correct after all.

> If that makes the linter happy

It does. And me too; less confusing! :)

> then we should do that change at is the minimal change needed.

Done.

ncw (approved):
LTGM :-)

I'll merge now.

PR #6243, opened 2022-06-13
Add selection to ncdu command

buengese:

Add the ability to select multiple items in the ncdu interface to delete them.

Either press `v` to select the current item, or `V` to highlight items as you move over them (slightly awkward behaviour that I'll tidy up here if you swap direction, the final item isn't toggled back). Then I've added a `D` command that will delete all the currently selected items.

Rebase and slight cleanup of this feature. One small additional change I might suggest is using the same eybind for delete and delete selection and just changing it's behavior depending on if items are selected.

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -92,6 +92,9 @@ func helpText() (tr []string) {
 		" u toggle human-readable format",
 		" n,s,C,A sort by name,size,count,average size",
 		" d delete file/directory",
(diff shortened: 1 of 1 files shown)


ncw:
> Rebase and slight cleanup of this feature.

Great!

Unfortunately it now has a conflict!

> One small additional change I might suggest is using the same keybind for delete and delete selection and just changing it's behavior depending on if items are selected.

That seems OK to me and it is one less thing to have in the help.

Issue #6124, opened 2022-04-24
Make Use of IPv4 also in Dual Stack Network


ghost:
#### The associated forum post URL

N/A

#### What is your current rclone version ?

v1.58.0

#### What problem are you are trying to solve?

Some ISPs allocate full bandwidth on both IPv4 and IPv6 Network. So, If I am on 100 Mbps Plan, I will get 100 Mbps on both IPv4 and IPv6 seperately and I can use this to get total output of 200 Mbps if software (in this case rclone) is able to make use of it. I was able to test it by running two rclone instances with one binded to IPv4 and another to IPv6. But, This is a hectic process and can be troublesome to use when trying to upload or download a folder through single instance. Would be cool thing if rclone either by default make use of it or could add a flag to enable this feature.

#### How do you think rclone should be changed to solve that?

Try and Use IPv4 and IPv6 alternatively while downloading files from services which support Dual Stack Networks like Google Drive etc.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
That's feels like a really edge use case.

Happy to make it an enhancement but doubtful it'll move.

ncw:
I think this would need a new http transport. It could maybe be done by modifying rclone's existing transport. You'd need to modify the name resolution in the Dial command so it only returned IPv4 or IPv6 addresses.

I think it is possible, but not easy!

@SmogrHzzgn do you want to work on this?

SmogrHzzgn:
> @SmogrHzzgn do you want to work on this?

Seems too much complicated for me as of right now.

SmogrHzzgn:
> @SmogrHzzgn do you want to work on this?

@ncw, This seems too much complicated for me as of right now as I don't have much experience in Go.

ncw:
I wonder if you can get an http proxy which does this? If so then you could use it via rclone.

buengese:
We already customize the Dialer for fshttp.Client and we might achieve this with relatively minimal changes to `DialContext`. I'll look into it.

ncw:
I've merged a fix for this this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.64 

[...]

PR #6244, opened 2022-06-13
Implement a way to force rclone to use ipv4/ipv6

buengese:

Implement a way force rclone to use IPv4/IPv6 on a dual-stack network as requested in #6124 . This was the simplest way I could think of that should have desired effect. 

Limitations include this forcing the dialer to make tcp connections only. This shouldn't be problem for `net/http` since HTTP/1.1 and HTTP/2 are tcp only .
I might have also missed other problems with this approach.

Alternative ways of achieving this is modifying address resolution probably by providing a custom resolver to the dialer.

#### Checklist

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -137,6 +137,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.BoolVarP(flagSet, &ci.DisableHTTP2, "disable-http2", "", ci.DisableHTTP2, "Disable HTTP/2 in the global transport")
 	flags.BoolVarP(flagSet, &ci.HumanReadable, "human-readable", "", ci.HumanReadable, "Print numbers in a human-readable format, sizes with suffix Ki|Mi|Gi|Ti|Pi")
 	flags.DurationVarP(flagSet, &ci.KvLockTime, "kv-lock-time", "", ci.KvLockTime, "Maximum time to keep key-value database locked by process")
+	flags.BoolVarP(flagSet, &ci.PreferIPv4, "prefer-ipv4", "", ci.PreferIPv4, "Force rclone to prefer IPv4 connections for all backends")
+	flags.BoolVarP(flagSet, &ci.PreferIPv6, "prefer-ipv6", "", ci.PreferIPv6, "Force rclone to prefer IPv6 connections for all backends")
 }
 
 // ParseHeaders converts the strings passed in via the header flags into HTTPOptions
@@ -178,6 +180,9 @@ func SetFlags(ci *fs.ConfigInfo) {
 	}
 	if (ci.DryRun || ci.Interactive) && ci.StatsLogLevel > fs.LogLevelNotice {
 		ci.StatsLogLevel = fs.LogLevelNotice
+	}
+	if ci.PreferIPv4 && ci.PreferIPv6 {
+		log.Fatalf("Cant force IPv4 and IPv6 at the same time")
 	}
 	if quiet {
 		if verbose > 0 {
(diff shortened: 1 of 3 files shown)


fs/fshttp/dialer.go
```diff
@@ -52,7 +52,16 @@ var warnDSCPFail, warnDSCPWindows sync.Once
 
 // DialContext connects to the network address using the provided context.
 func (d *Dialer) DialContext(ctx context.Context, network, address string) (net.Conn, error) {
-	c, err := d.Dialer.DialContext(ctx, network, address)
+	var c net.Conn
+	var err error
+	ci := fs.GetConfig(ctx)
+	if ci.PreferIPv4 {
```
ncw, line 58: What are the possible values of `network` here? Maybe these if statements should be wrapped in an `if network == "tcp"`?

fs/config/configflags/configflags.go
```diff
@@ -137,6 +137,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.BoolVarP(flagSet, &ci.DisableHTTP2, "disable-http2", "", ci.DisableHTTP2, "Disable HTTP/2 in the global transport")
 	flags.BoolVarP(flagSet, &ci.HumanReadable, "human-readable", "", ci.HumanReadable, "Print numbers in a human-readable format, sizes with suffix Ki|Mi|Gi|Ti|Pi")
 	flags.DurationVarP(flagSet, &ci.KvLockTime, "kv-lock-time", "", ci.KvLockTime, "Maximum time to keep key-value database locked by process")
+	flags.BoolVarP(flagSet, &ci.PreferIPv4, "prefer-ipv4", "", ci.PreferIPv4, "Force rclone to prefer IPv4 connections for all backends")
```
ncw, line 140: This should probably be `Force rclone to prefer IPv4 connections` - it isn't just for backends.

This would traditionally have the short flag `-4` - which might be worth adding?

fs/config/configflags/configflags.go
```diff
@@ -137,6 +137,8 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
...
+	flags.BoolVarP(flagSet, &ci.PreferIPv6, "prefer-ipv6", "", ci.PreferIPv6, "Force rclone to prefer IPv6 connections for all backends")
```
ncw, line 141: Same as above but s/4/6/g!

ncw (commented):
This is a useful feature, but it doesn't actually implement what the user wanted in #6124 which was to use **both** IPv4 and IPv6, so (maybe) alternate betwen them or choose one at random.

Either way that should be on a different flag in a different commit!

buengese:
Hm true automatically switching between them could probably be implemented in the same place if something like round robin is enough. Everything more fancy would be rather complicated.

fs/fshttp/dialer.go
buengese: The http transport will only use `tcp` which lets the dialer decide between IPv4 and IPv6. Other possible values would be `udp, udp4, udp6, ip, ip4, ip6`.

ncw:
> Hm true automatically switching between them could probably be implemented in the same place if something like round robin is enough. Everything more fancy would be rather complicated.
> 
> Connection reuse might interfere tough.
> 
> I kinda need this feature too, so I'll think about it.

If you want to do it, I'd start with round robin and see if that is good enough.  (or even `rand.Intn(2) == 0`)

ncw:
I've merged a variant of this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.64 

You can use `--bind 0.0.0.0` to force rclone to use IPv4 addresses and `--bind ::0` to force rclone to use IPv6 addresses.

PR #6258, opened 2022-06-17
ncdu: replace termbox with tcell's termbox wrapper

eNV25:
https://github.com/gdamore/tcell

There are some issues with this implementation.

Using Clear with fg bg ColorDefault resulted in a white background for
some reason. Clear with fg ColorWhite bg ColorBlack was used instead.

tcell's termbox wrapper doesn't support ColorLightYellow.
ColorYellow + 8 was used instead.

ncdu: replace termbox with tcell's termbox wrapper

n/a

#### Checklist


eNV25:
A notable benefit is that this implementation should work on solaris.

buengese:
With tembox-go being somewhat unsupported this probably a good idea.

ncw (commented):
This code looks good and having read up on the tcell library and the deprecation of termbox it seems like a good idea to move over to it.

How well does it work on Windows? @albertony could you test it on all the different windows terminals?

albertony:
I did a very quick test earlier today, and that looked good. But can test some more systematic, sure.

I was/am curious about the lack of `ColorDefault`, worked around with `termbox.Clear(termbox.ColorWhite, termbox.ColorBlack)`, if there is a difference when the terminal have another default color, e.g. like Powershell that previously had blue background color. But maybe ncdu is setting specific fg/bg colors so much elsewhere that it does not make much difference..

eNV25:
@albertony 

> I was/am curious about the lack of `ColorDefault`, worked around with `termbox.Clear(termbox.ColorWhite, termbox.ColorBlack)`

It's probably a bug in tcell's termbox wrapper. It doesn't happen when tcell is used directly.

I could look into using tcell directly, it could produce a better result. In particular instead of using `fg White, bg Black` and `fg Black, bg White`, we could use `Default` and `Default.Reverse()`. This way the terminal's color scheme is respected better.

> if there is a difference when the terminal have another default color, e.g. like Powershell that previously had blue background color. But maybe ncdu is setting specific fg/bg colors so much elsewhere that it does not make much difference..

ncdu already sets colors explicitly. So I think this is more consistent anyway. Unless we change to using tcell directly as I explained above.

eNV25:
I think porting ncdu to use tcell directly will be tedious but not too difficult.

The biggest changes would be that `UI` will embed `tcell.Screen` i.e. `type Ui struct{ tcell.Screen; ... }`, functions like `Print` and `Line` will become methods of `UI` and will use `u.SetContent` instead of `termbox.SetCell`.

Color handling would need to be rethought a little bit. tcell uses a single Style instead of a pair of fg bg Attributes.

albertony:
> ncdu already sets colors explicitly. So I think this is more consistent anyway

Agreed.

Just tested with a command prompt with custom colors, and it behaves exactly as previous version - basically ignores the terminal color scheme and uses the black/white scheme.

So far so good.

eNV25:
> Edit: Tried to run it from from MSYS2 bash on Windows, that does not work, though didn't really expect it to either, and old version is no different.

Are compiling using native Windows Go or Cygwin/MSYS2 Go? There might be a difference.

I don't use Windows anymore so I can't test it out.

albertony:
> Are compiling using native Windows Go or Cygwin/MSYS2 Go? There might be a difference.

Good point. I was testing native Windows Go compiled version. Never tried building rclone from msys, that would be interesting, maybe I will try it out, but not sure I will find the time to do so in near future (weeks).

ncw:
> I could look into using tcell directly, it could produce a better result. In particular instead of using `fg White, bg Black` and `fg Black, bg White`, we could use `Default` and `Default.Reverse()`. This way the terminal's color scheme is respected better.

I couldn't figure out how to respect the terminals default colours when I originally wrote ncdu - hence the white on black scheme!

Using tcell directly would be a good idea if we can fix this.

However I suggest we get this pr merged first as I want to make a release soon.

I think this probably is ok to merge - do you think it has had enough windows testing @albertony  ?

cmd/ncdu/ncdu.go
```diff
@@ -1,7 +1,7 @@
 // Package ncdu implements a text based user interface for exploring a remote
 
-//go:build !plan9 && !solaris && !js
-// +build !plan9,!solaris,!js
+//go:build !plan9 && !js
```
eNV25, line 3: Do any docs need to be updated?

eNV25:
I'm done here so I think this can be merged.

cmd/ncdu/ncdu.go
ncw: We don't usually state which subcommands work on which OSes, the OS just won't have that subcommand.

However making it run on solaris is a nice bonus :-)

albertony:
> However I suggest we get this pr merged first as I want to make a release soon.

Agreed.

> I think this probably is ok to merge - do you think it has had enough windows testing @albertony ?

I have only done some quick testing so far. It might take a couple of days to get the time to more, but if you can think of any particular things that typically could be problematic I run a quick test on that. Though, I have smoke-tested it on Command Prompt and PowerShell, stand-alone and via Windows Terminal, and haven't seen any issues so far. So I would say it is ok to merge and get a beta build with it, at least.

ncw:
Thanks for testing @albertony - you've hit the major points users, so I'll merge this now so it can hit the beta.

ncw:
@eNV25 if you would like to prepare a PR to move ncdu over to the tcell native implementation that would be fantastic!

I think `rclone ncdu` should respect the users terminal settings if possible rather than go white text on black background so if we can do that with tcell it would be marvellous :-)

eNV25:
@ncw Unfortunately I don't have time for that, at the moment.

If you want to work on it yourself, here is a small patch that ports some functions to tcell for example.

```
diff --git a/cmd/ncdu/ncdu.go b/cmd/ncdu/ncdu.go
index aea2501eb..b2d029dec 100644
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/atotto/clipboard"
+	"github.com/gdamore/tcell/v2"
 	"github.com/gdamore/tcell/v2/termbox"
 	runewidth "github.com/mattn/go-runewidth"
@@ -110,6 +111,7 @@ func helpText() (tr []string) {
 
 // UI contains the state of the user interface
 type UI struct {
+	tcell.Screen
 	f                  fs.Fs     // fs being displayed
 	fsName             string    // human name of Fs
 	root               *scan.Dir // root directory
@@ -145,17 +147,17 @@ type dirPos struct {
 }
 
 // Print a string
-func Print(x, y int, fg, bg termbox.Attribute, msg string) {
+func (u *UI) Print(x, y int, style tcell.Style, msg string) {
 	for _, c := range msg {
-		termbox.SetCell(x, y, c, fg, bg)
+		u.SetContent(x, y, c, nil, style)
 		x++
 	}
 }
 
 // Printf a string
-func Printf(x, y int, fg, bg termbox.Attribute, format string, args ...interface{}) {
+func (u *UI) Printf(x, y int, style tcell.Style, format string, args ...interface{}) {
 	s := fmt.Sprintf(format, args...)
-	Print(x, y, fg, bg, s)
+	u.Print(x, y, style, s)
 }
 
 // Line prints a string to given xmax, with given space

```

albertony:
> Thanks for testing @albertony - you've hit the major points users, so I'll merge this now so it can hit the beta.

I had some more time to test after all, and still found no issues, so I'm considering this task `done` ©️ ™️

albertony:
Could be a fun task. I will keep it on my list, but not sure when I will ~have~ take the time (weeks/months 🌞) so until then its up for grabs..

PR #6259, opened 2022-06-17
ncdu: capture log while running and print at exit

eNV25:

ncdu: capture log while running and print at exit

n/a

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -8,6 +8,7 @@ package ncdu
 import (
 	"context"
 	"fmt"
+	"log"
 	"path"
 	"reflect"
 	"sort"
@@ -21,6 +22,7 @@ import (
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/fspath"
 	"github.com/rclone/rclone/fs/operations"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -74,6 +76,14 @@ UI won't respond in the meantime since the deletion is done synchronously.
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
 		cmd.Run(false, false, command, func() error {
+			var buf strings.Builder
+			logSave := log.Writer()
+			defer log.SetOutput(logSave)
+			logrusSave := logrus.StandardLogger().Writer()
+			defer logrus.SetOutput(logrusSave)
+			log.SetOutput(&buf)
+			logrus.SetOutput(&buf)
+			defer fmt.Fprint(logSave, &buf)
 			return NewUI(fsrc).Show()
 		})
 	},


eNV25:
This prevent screen corruption when something is logged while rclone ncdu is running.

eNV25:
What's up with the failed check?

albertony:
Seemed unrelated. I triggered a re-run, lets see if it helps.

ncw (commented):
This should probably check to see if stderr is being redirected as you can currently do `rclone ncdu ... 2>file.log` or even `rclone ncdu --log-file file.log ...` and we need to keep both of those working.

Just wondering why you redirected the logrus logger too? I think we only use that for --use-json-log don't we?

I would also be tempted to put this into a `log.Pause` in `fs/log.go` - this should return a function to be run to clean up and show the log, so you'd use it like

```go
    cleanup := log.Pause()
    defer cleanup()
```

Or even like this if you were feeling fancy

```
    defer log.Pause()()
```

eNV25:
> This should probably check to see if stderr is being redirected as you can currently do `rclone ncdu ... 2>file.log` or even `rclone ncdu --log-file file.log ...` and we need to keep both of those working.

This seems to work perfectly with `2>log`.

But when using `--log-file log` my drive dangling shortcut warnings are duplicated. I will investigate.

> Just wondering why you redirected the logrus logger too? I think we only use that for --use-json-log don't we?

I now really use how logging works in rclone. So I copied `cmd/bisync/bilib.CaptureOutput`. I am not sure if that was the current way to do it.

eNV25:
> But when using `--log-file log` my drive dangling shortcut warnings are duplicated. I will investigate.

Never mind, false alarm. I just didn't delete the file.

albertony:
> This seems to work perfectly with `2>log` and `--log-file log`.

I guess the difference is that the log file will not be written to until everything is dumped when you exit ncdu?

eNV25:
> > This seems to work perfectly with `2>log` and `--log-file log`.
> 
> I guess the difference is that the log file will not be written to until everything is dumped when you exit ncdu?

Yes

albertony:
> This prevent screen corruption when something is logged while rclone ncdu is running.

Just found out there is already a shortcut to fix it manually; Ctrl+L. But maybe you knew that? I did not! 😊 (While testing some other stuff I though it was supposed to reload the file list, but it is in fact a "redraw" of the screen - which gets rid of all error message "corruption" issues). Don't know if it changes anything regarding this PR, just wanted to mention it.

eNV25:
> Just found out there is already a shortcut to fix it manually; Ctrl+L.

That's true. It's conventional for terminal UIs like this to redraw screen completely with Ctrl+L.

ncw:
I'm not sure this is going about this the right way.

I agree, it would be nice if logs don't corrupt the screen with ncdu.

I'm not sure storing them all until the end is the right thing to do though. I have a concern about memory use and also about rclone pumping out thousands of logs when you exit ncdu isn't going to be a good user experience.

Maybe we should be showing the logs to the user in an overlay text window while ncdu is running? That would be more useful to the user.

If we want to do this then intercepting the logs the way `--progress` does is the right thing to do

https://github.com/rclone/rclone/blob/7a909ebfb0d100187fab7949455bf74b8ce07042/cmd/progress.go#L29-L40

We can then capture the last 40 logs in a buffer and show them in an overlay window like the help. This could maybe pop up on a new log and be dismissed with ESC.

What do you think?

ncw:
> > I'm not sure storing them all until the end is the right thing to do though. I have a concern about memory use and also about rclone pumping out thousands of logs when you exit ncdu isn't going to be a good user experience.
> 
> I agree this is probably not the best use experience. In this change logs are only visible once the program ends and would have no idea logs are even printed until then. It would be more useful if the user could view logs while running ncdu.
> 
> But I don't think this is completely a bad idea in terms of user experience. With this change (at least in my computer) it looks as if the logs were always there in the background but the ncdu screen was hiding it until then.

Thats good.

> > Maybe we should be showing the logs to the user in an overlay text window while ncdu is running? That would be more useful to the user.
> > If we want to do this then intercepting the logs the way `--progress` does is the right thing to do
> 
> One idea would be to show the latest log line in the existing status bar. This would alert the user that something has happened. Once the user exits ncdu they can see full logs like in this patch.

Great plan. So have one log on display in the status bar and on exit show the last N logs - I do think there should be a limit, but it could be large - 1000 say. We could at some point make a key to open the log window and scroll about it.

> On another note, since we are discussing ideas for how to fix the issue should we close this PR and open an issue instead? It might be more appropriate.

That's probably a good plan - I made a new issue here: https://github.com/rclone/rclone/issues/6271

PR #6268, opened 2022-06-19
Doc fixes

albertony:

Some small doc fixes here and there. Commit messages describes the individual changes.

Partly related to #6267, where I accidentally merged a change in autogenerated section. There is one commit here that fixes that.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ These backends adapt or modify other storage providers
   * Optional encryption ([Crypt](https://rclone.org/crypt/))
   * Optional FUSE mount ([rclone mount](https://rclone.org/commands/rclone_mount/))
   * Multi-threaded downloads to local disk
-  * Can [serve](https://rclone.org/commands/rclone_serve/) local or remote files over HTTP/WebDav/FTP/SFTP/dlna
+  * Can [serve](https://rclone.org/commands/rclone_serve/) local or remote files over HTTP/WebDAV/FTP/SFTP/DLNA
 
 ## Installation & documentation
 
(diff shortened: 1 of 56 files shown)


docs/content/flags.md
```diff
@@ -616,7 +616,7 @@ and may be set in the config file.
       --webdav-pass string                           Password (obscured)
       --webdav-url string                            URL of http host to connect to
       --webdav-user string                           User name
-      --webdav-vendor string                         Name of the Webdav site/service/software you are using
+      --webdav-vendor string                         Name of the WebDAV site/service/software you are using
```
ncw, line 619: This is an auto generated file I think

ncw (approved):
Very nice changes - thank you :-)

The cross referencing is a really nice touch.

Please merge when happy!

docs/content/flags.md
albertony: Oh really, I didn't know that (or perhaps just completely forgot it). Thanks.

albertony:
Thanks. I added some more similar fixes, but should be done now. Have looked over the changes several times, so I'm pretty sure it is correct, but I want to verify the live docs in browser also before merge.

ncw:
> Thanks. I added some more similar fixes, but should be done now.

Great. LGTM :-)

> Have looked over the changes several times, so I'm pretty sure it is correct, but I want to verify the live docs in browser also before merge.

I'm sure you know how to do this, but just in case you don't you do with `make serve` - requires [hugo](https://gohugo.io/) to be installed and on the path.

PR #6270, opened 2022-06-20
Jottacloud: Always store username in config and use it to avoid initial api request

albertony:

Existing version did save username in config, but only when entering the non-standard device/mountpoint sequence during config. Regardless of that, it did always request the username from the api at startup (NewFS). So the first one is a bug, but then the second superfluous api request masks the bug. (Alternatively one could say the api request is correct, but the first bug is just that the username should never be stored in config).

This commit improves these things:
- The username will now always be stored in config.
    -> Consistency.
- When using standard authentication the username is picked from the login token, instead of requesting it separately from the remote api during the config process.
    -> Saves 1 api request during config.
- The fs constructor picks the username from config instead of always requesting it from remote api, but for backwards compatibility it checks if not present in config and resorts to api request as before. Added a info message informing if the superfluous request is being made (i.e. shown with `-v`, don't think it is necessary to make it always show as notice, and it is more important than just debug shown with `-vv`?)
   -> Saves 1 api request at startup.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -152,7 +152,7 @@ func Config(ctx context.Context, name string, m configmap.Mapper, config fs.Conf
 		m.Set(configClientSecret, "")
 
 		srv := rest.NewClient(fshttp.NewClient(ctx))
-		token, tokenEndpoint, err := doTokenAuth(ctx, srv, loginToken)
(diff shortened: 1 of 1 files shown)


buengese (approved):
This looks fine. The auth code has been changed/refactored and adjusted so many that these things just sneak in.

Issue #6273, opened 2022-06-22
Add staticcheck to build pipeline


albertony:
#### What problem are you are trying to solve?

Add [staticcheck](<https://staticcheck.io>) based linting to the build pipeline. Maybe as an advisory to start with.

Staticcheck is the default linter in VSCode-based golang dev environment, and from my experience it gives many relevant suggestions that the current linter does not consider.

First we need to fix as many as possible of the reported "problems", and if necessary add ignores of any false positives or suggestions we for other reasons don't agree with (and then decide on the best strategy for managing such ignores, see https://staticcheck.io/docs/configuration/#ignoring-problems). I've started going through them myself, fixing in different PRs:
- https://github.com/rclone/rclone/pull/6228
- https://github.com/rclone/rclone/pull/6239
- https://github.com/rclone/rclone/pull/6238
- https://github.com/rclone/rclone/pull/6236
- https://github.com/rclone/rclone/pull/6237
- https://github.com/rclone/rclone/pull/6275
- https://github.com/rclone/rclone/pull/6276
- https://github.com/rclone/rclone/pull/6277
- https://github.com/rclone/rclone/pull/6389 (New: Deprecations starting with go 1.16, detected after minimum was raised to go 1.17)
- https://github.com/rclone/rclone/pull/6387 (New)

When I started it reported about 300 "problems", if/when all these are merged we are down to below 30 (with staticcheck v0.3.2 and go v1.18.3). Remaining includes:
- func is unused (13 occurrences)
- httplib is deprecated and has been replaced with lib/http (8 occurrences)
- infinite recursive call, here: https://github.com/rclone/rclone/blob/73e3bb09d71c31814e00ebcad2eec9be3b74ccf0/fs/pacer.go#L76
- argument err is overwritten before first use, here: https://github.com/rclone/rclone/blob/73e3bb09d71c31814e00ebcad2eec9be3b74ccf0/backend/onedrive/onedrive.go#L1670
- argument should be pointer-like to avoid allocations, here: https://github.com/rclone/rclone/blob/73e3bb09d71c31814e00ebcad2eec9be3b74ccf0/backend/crypt/cipher.go#L242

#### How do you think rclone should be changed to solve that?

See the mentioned PRs in general, and this: https://github.com/rclone/rclone/pull/6228#issuecomment-1154869610

[...]

PR #6275, opened 2022-06-24
Fix staticcheck linting problems

albertony:

Fix more "problems" reported by the staticcheck lint tool (v0.3.2).

Continuing from https://github.com/rclone/rclone/pull/6228
See also https://github.com/rclone/rclone/issues/6273

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -4288,7 +4288,7 @@ func (o *Object) uploadMultipart(ctx context.Context, req *s3.PutObjectInput, si
 func unWrapAwsError(err error) (found bool, outErr error) {
 	if awsErr, ok := err.(awserr.Error); ok {
 		var origErrs []error
-		if batchErr, ok := awsErr.(awserr.BatchError); ok {
+		if batchErr, ok := awsErr.(awserr.BatchedErrors); ok {
 			origErrs = batchErr.OrigErrs()
 		} else {
 			origErrs = []error{awsErr.OrigErr()}
--- a/backend/yandex/yandex.go
+++ b/backend/yandex/yandex.go
@@ -782,11 +782,11 @@ func (f *Fs) DirMove(ctx context.Context, src fs.Fs, srcRemote, dstRemote string
 	}
 
 	_, err = f.readMetaDataForPath(ctx, dstPath, &api.ResourceInfoRequestOptions{})
-	if apiErr, ok := err.(*api.ErrorResponse); ok {
+	if _, ok := err.(*api.ErrorResponse); ok {
 		// does not exist
-		if apiErr.ErrorName == "DiskNotFoundError" {
-			// OK
-		}
+		//if apiErr.ErrorName == "DiskNotFoundError" {
+		//	// OK
+		//}
 	} else if err != nil {
 		return err
 	} else {
--- a/backend/yandex/yandex.go
+++ b/backend/yandex/yandex.go
@@ -782,11 +782,10 @@ func (f *Fs) DirMove(ctx context.Context, src fs.Fs, srcRemote, dstRemote string
 	}
 
 	_, err = f.readMetaDataForPath(ctx, dstPath, &api.ResourceInfoRequestOptions{})
-	if _, ok := err.(*api.ErrorResponse); ok {
-		// does not exist
-		//if apiErr.ErrorName == "DiskNotFoundError" {
-		//	// OK
-		//}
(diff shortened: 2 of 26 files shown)


backend/s3/s3.go
```diff
@@ -3837,7 +3839,8 @@ func (o *Object) downloadFromURL(ctx context.Context, bucketPath string, options
 	for key, value := range resp.Header {
 		if strings.HasPrefix(key, "x-amz-meta") {
 			metaKey := strings.TrimPrefix(key, "x-amz-meta-")
-			metaData[strings.Title(metaKey)] = &value[0]
+			metaKey = cases.Title(language.Und, cases.NoLower).String(metaKey)
```
ncw, line 3842: Gah, this must be wrong! Your change is fine, but the code is wrong before. I suggest you leave this one, and I will fix it properly in my metadata branch!

cmd/help.go
```diff
@@ -315,7 +317,8 @@ func showBackend(name string) {
 			optionsType = "advanced"
 			continue
 		}
-		fmt.Printf("### %s options\n\n", strings.Title(optionsType))
+		optionsType = cases.Title(language.Und, cases.NoLower).String(optionsType)
```
ncw, line 320: This is really ugly! the strings are always ASCII strings for which `strings.Title` works properly.

fs/operations/operations_test.go
```diff
@@ -270,7 +272,7 @@ func TestHashSums(t *testing.T) {
 		if !hashes.Contains(test.ht) {
 			continue
 		}
-		name := strings.Title(test.ht.String())
+		name := cases.Title(language.Und, cases.NoLower).String(test.ht.String())
```
ncw, line 275: Likewise these are always ASCII strings...

backend/yandex/yandex.go
```diff
@@ -782,11 +782,11 @@ func (f *Fs) DirMove(ctx context.Context, src fs.Fs, srcRemote, dstRemote string
...
 
 	_, err = f.readMetaDataForPath(ctx, dstPath, &api.ResourceInfoRequestOptions{})
-	if apiErr, ok := err.(*api.ErrorResponse); ok {
+	if _, ok := err.(*api.ErrorResponse); ok {
 		// does not exist
-		if apiErr.ErrorName == "DiskNotFoundError" {
-			// OK
-		}
+		//if apiErr.ErrorName == "DiskNotFoundError" {
+		//	// OK
```
ncw, line 788: This should probably return the error if `apiErr.ErrorName != "DiskNotFoundError"`...

cmd/selfupdate/verify.go
```diff
@@ -10,9 +10,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/ProtonMail/go-crypto/openpgp"
```
ncw, line 13: How did you choose this package?

ncw (commented):
I've been though these. See comments inline.

Thank you :-)

cmd/help.go
albertony: It's not "neat", I can agree with that. Still, ascii or not, `strings.Title` is listed as deprecated on https://pkg.go.dev/strings#Title, so I would think it is better not to use it.

cmd/selfupdate/verify.go
albertony: There wasn't a whole lot of good alternatives to pick from. This was one which had recent commits, and came from a somewhat known company, with privacy and encryption related focus, though I cannot vouch for them any more then others..

The deprecation notice in https://pkg.go.dev/golang.org/x/crypto/openpgp and the linked issue there tries to steer users away from OpenPGP in general. The ProtonMail variant is mentioned at the of the issue description:

> Every major player that actually relies on OpenPGP and Go maintains a fork (https://github.com/keybase/go-crypto, https://github.com/ProtonMail/go-crypto, ...). Developers would be better served by finding those, and by centralizing efforts onto one, without blocking on the resource limits of the Go team...

If this is something that needs more consideration, maybe I should split it out in a separate PR? I'll probably be away a lot the coming weeks, so then I could try to get the rest merged..

cmd/help.go
ncw: OK. Let's go with the flow!

albertony:
Updated according to comments. I assume this can be merged now...?

ncw (approved):
LGTM - please merge

PR #6276, opened 2022-06-24
Fix missing response when using custom auth handler in http serve

albertony:

I was running the staticcheck linter and investigated the following reported issues:

> this value of r is never used (SA4006)
> WithContext is a pure function but its return value is ignored (SA4017)

On the following line:

https://github.com/rclone/rclone/blob/fdd2f8e6d21d1cf95b33fd2bb09f9c1caef0069d/lib/http/auth/basic.go#L114

I'm not sure I understand this line and the `CustomAuth` function it is in, but my impression is that staticcheck is in fact correct here, and that this line does nothing as the implementation currently is.

```
// WithContext returns a shallow copy of r with its context changed
// to ctx. The provided ctx must be non-nil.
//
// For outgoing client request, the context controls the entire
// lifetime of a request and its response: obtaining a connection,
// sending the request, and reading the response headers and body.
//
// To create a new request with a context, use NewRequestWithContext.
// To change the context of a request, such as an incoming request you
// want to modify before sending back out, use Request.Clone. Between
// those two uses, it's rare to need WithContext.
```

If this is true then is it not the case that the CustomAuth function does nothing at all, on success, and that an http serve would not work at all if "custom auth" is enabled (I have no idea how to do that)? Wild guess: It should have called `next.ServeHTTP(w, r)`, using the copy of `r` returned from `WithContext`, just like the handler implementation of the other auth implementations (HtPasswdAuth and SingleAuth).

https://github.com/rclone/rclone/blob/fdd2f8e6d21d1cf95b33fd2bb09f9c1caef0069d/lib/http/auth/basic.go#L73

#### Checklist

### DIFF
--- a/lib/http/auth/basic.go
+++ b/lib/http/auth/basic.go
@@ -113,6 +113,7 @@ func CustomAuth(fn CustomAuthFn, realm string) httplib.Middleware {
 				if value != nil {
 					r = r.WithContext(context.WithValue(r.Context(), ContextAuthKey, value))
 				}
+				next.ServeHTTP(w, r)
 			}
 		})
 	}


lib/http/auth/basic.go
```diff
@@ -112,6 +112,7 @@ func CustomAuth(fn CustomAuthFn, realm string) httplib.Middleware {
 				}
 				if value != nil {
 					r = r.WithContext(context.WithValue(r.Context(), ContextAuthKey, value))
+					next.ServeHTTP(w, r)
```
ncw, line 115: I think this needs to go after the `}` so it always gets called, but it looks like the right fix.

ncw (commented):
This is part of the unfinished http refactor and I think your fix is more or less correct but needs to go down one line!

I'm still tempted to revert the http refactor and start again introducing the `chi` router but leaving the rest of the code alone as the refactor has introduced a lot of bugs.

The custom Auth is used by the auth proxy when using `rclone serve http` or `rclone serve webdav`.

Tracing the code, I think the reason why this isn't causing a problem is that that code installs its own Context

https://github.com/rclone/rclone/blob/fdd2f8e6d21d1cf95b33fd2bb09f9c1caef0069d/cmd/serve/httplib/httplib.go#L251-L260

So maybe that bit of code is never called?

What do you think about the whole http refactor? Its a really mess at the moment, there are two ways of doing everything.

Perhaps I will revert it and start again after the next rclone release...

albertony:
> This is part of the unfinished http refactor and I think your fix is more or less correct but needs to go down one line!

Yes, that was a blunder. I pushed an update to fix this just for now, and will convert this PR to draft until we can conclude on how to proceed. It will probably be some weeks until I have time to pick up on this again.

PR #6277, opened 2022-06-25
Use jwt-go (golang-jwt) instead of deprecated jws (x/oauth2/jws)

albertony:

golang.org/x/oauth2/jws is deprecated

> this package is not intended for public use and might be removed in the future. It exists for internal use only. Please switch to another JWS package or copy this package into your own source tree.

github.com/golang-jwt/jwt/v4 seems to be a good alternative, and was already an implicit dependency.

Box backend is affected. Code change in common lib/jwtutil as well as backend/box, but only box implementation uses jwtutil I think.

❗ Untested ❗ 

#### Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,11 +32,6 @@ issues:
       - staticcheck
       text: 'SA1019: "github.com/rclone/rclone/cmd/serve/httplib" is deprecated'
 
-    # TODO: Remove if/when this is fixed by merging PR #6277.
-    - linters:
-      - staticcheck
-      text: 'SA1019: "golang.org/x/oauth2/jws" is deprecated'
-
 run:
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -27,6 +27,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/golang-jwt/jwt/v4"
 	"github.com/rclone/rclone/backend/box/api"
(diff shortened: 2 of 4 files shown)


ncw (commented):
That's quite a big change not to have tested!

@dbramwell can you help us check this is working properly? Box was kind enough to give me an enterprise account so I can probably test this myself, but I'm struggling to work out how to create the right things!

albertony:
> That's quite a big change not to have tested!

Definitely. So we need to ☎️ some testers before thinking about merge..

ncw:
@dbramwell no problems! I just need to work out how to make the right kind of file...

Any chance you could tell us what goes in the config file? I've been unable to make it work so far.

dbramwell:
I think there was an option in the box gui to download one. Possibly this https://developer.box.com/guides/sso-identities-and-app-users/connect-okta-to-app-users/configure-box/#download-required-data ? Sorry my memory is a bit fuzzy, not used it for a few years

ncw:
@albertony I see you updated this branch. What do you think we should do with it? Merge it?

albertony:
Yeah, I just rebased it on master. I think there was a merge conflict.

Don't know what to do with it... Reason I did this change was probably that, because of the deprecation warning, ci/check tools in other PRs would complain about it. You think there is a chance it will be tested on the beta, if merged? Or maybe there are not many users of this feature, maybe none at all...

ncw (approved):
I think we should merge this. It looks correct.

If it needs fixing then we can fix it in a point release as usual!

albertony:
Ok, lets do it. 🤞

PR #6278, opened 2022-06-25
lib/cache: if cached value supports it, run Shutdown on removal

miku:

Some backends implement the optional fs.Shutdowner interface. Until now,
Shutdown is only checked and called, when a backend is wrapped (e.g.
crypt, compress, ...).

To have a common way to perform operations at the end of the backend
lifecycle with proper error handling, we can call Shutdown at cache
clear time.

Since Shutdown returns an error, we add two new methods: cache.Err() and
cache.ClearErrors(); cache.Err() returns the error or errors that
occured. ClearErrors() resets any errors. All errors are recorded and
kept around until they are cleared.

The cache.Err() approach is inspired from similar implementations, e.g.
in bufio.Scanner. This way we can leave method signatures in lib/cache
unchanged and we also do not emphasize the error reporting (but allow
for checks).

Yes, previous discussion: https://forum.rclone.org/t/31336

#### Checklist


lib/cache/cache.go
```diff
@@ -16,6 +20,7 @@ type Cache struct {
 	expireRunning  bool
 	expireDuration time.Duration // expire the cache entry when it is older than this
 	expireInterval time.Duration // interval to run the cache expire
+	err            error         // any error, e.g. while dropping backends which can shutdown
```
ncw, line 23: Instead of storing the error here, I'd suggest adding a value

```go
finalize func(value interface{})
```

and a method

```go
func (c *Cache) SetFinalizer(finalize func(value interface{}))
```

Which can be used to have a callback for finalizing objects.

That will look a lot like `tryShutdown` except it will use `fs.Errorf` to log each error, and it will call `fs.CountError` to set the global error flag.

func (c *Cache) AddFinalizer(func()

lib/cache/cache.go
```diff
@@ -154,7 +159,12 @@ func (c *Cache) GetMaybe(key string) (value interface{}, found bool) {
 // Returns true if the entry was found
 func (c *Cache) Delete(key string) bool {
 	c.mu.Lock()
-	_, found := c.cache[key]
+	entry, found := c.cache[key]
+	if found {
+		if err := tryShutdown(entry.value); err != nil {
```
ncw, line 164: This then becomes

```go
if c.finalize != nil {
    c.finalize(entry.value)
}
```

I'd probably make sure c.finalize is never nil so points to a no-op function so you can do without the `if c.finalize { }` which will make the code a lot neater.

ncw (commented):
I think the basic idea looks sound :-)

I've put some suggestions for rework which makes the `lib/cache` free from `fs` and makes the error handling a bit easier.

Thank you :-)

miku:
Thanks a lot for your review, I'm just getting into the rclone codebase; the finalizer approach is much nicer. I addressed your feedback and hope the code looks a bit better now.

cmd/cmd.go
```diff
@@ -321,6 +321,14 @@ func Run(Retry bool, showStats bool, cmd *cobra.Command, f func() error) {
 		}
 	}
 
+	// add hook to wind down backends
```
ncw, line 324: I think this should go here

https://github.com/rclone/rclone/blob/326c43ab3f5c7c8570bf87b4656b41846ce00877/fs/cache/cache.go#L22-L27

fs/cache/cache.go
```diff
@@ -167,3 +167,9 @@ func Entries() int {
 	createOnFirstUse()
 	return c.Entries()
 }
+
```
ncw, line 170: ... then we don't need this method

ncw (commented):
That is looking really good now :-)

I think we should move the call to SetFinalizer and I've put a couple of notes inline about that.

Thank you :-)

miku:
Thank you! - This will greatly improve error handling options for our backend implementation (still in development).

ncw:
And thank you for doing this useful, but unglamorous bit of work :-)

Issue #6280, opened 2022-06-26
Reduce default parallelism to improve stability on slow network backends; Document significance of --checkers and/or split them into finer controls


mirekphd:
#### What is your current rclone version (output from `rclone version`)?
```
/ $ rclone version
rclone v1.58.1-DEV
- os/version: alpine 3.15.4 (64 bit)
- os/kernel: 3.10.0-957.27.2.el7.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.18.2
- go/linking: dynamic
- go/tags: none
```

#### What problem are you are trying to solve?
Trying to verify whether any file in a large remote repo (a few hundreds of files) needs any updates. The remote server is a S3-compatible MinIO (containerized, `mirekphd/ml-minio22:latest`), with distributed network storage on an NFS server (slow). The container with minio has 1000 CPU millicores request (i.e. a soft. burstable, not a hard limit) (not exhaused - this server has many CPU threads available) and 4 to 40 GiB RAM limit (also not exhausted - most of the limit is free). 

The default `rclone` multi-threading settings - 8 threads (as controlled by the `--checkers` argument) is giving us consistent crashes (crash loop back-offs) of the MinIO servers during `rclone`'s "checking" phase in multiple namespaces and nodes. Error messages in the MinIO 2022 server log indicate an excessive number of threads spawned by the server when servicing rclone requests (approaching a "fork bomb"), e.g. as follows:
```
[..]
runtime: failed to create new OS thread (have 4097 already; errno=11)
runtime: may need to increase max user processes (ulimit -u)
fatal error: newosproc
runtime: failed to create new OS thread (have 4098 already; errno=11)
runtime: may need to increase max user processes (ulimit -u)
fatal error: newosproc
[..]
```

#### How do you think rclone should be changed to solve that?
The default value of the `--checkers ` argument should be **reduced at least to 4 threads**, on par with the default number of threads used for transferring files in `rclone copy`. Such a reduced setting (4 instead of default 8) has been tested to prevent any such errors for MinIO on NFS in our problematic use cases - namespaces with hundreds of thousands of files (in multiple subfolders and totalling up to 1.2 TB of data in each namespace), but of  course there's no guarantee that it will also hold for millions of files and petabytes of data - maybe a single thread would be most prudent, like in `restic`? 

[...]

PR #6282, opened 2022-06-27
 Wider impact of --checkers=N on parallelism (see #6280).

mirekphd:


ncw (approved):
That looks great - thank you - I'll merge it now.

PR #6284, opened 2022-06-28
operations: fix possible unintended --size-only skip with multithread…

### LINKED ISSUE #6206: Make rclone more robust to false-negatives with local copies and --size-only

https://forum.rclone.org/t/does-size-only-account-for-local-sparse-files-properly/31022/5

```
rclone v1.58.1
- os/version: darwin 10.15.7 (64 bit)
- os/kernel: 19.6.0 (x86_64)
- os/type: darwin
- os/arch: amd64
- go/version: go1.17.9
- go/linking: dynamic
- go/tags: cmount
```

(theoretical I think but...) if doing a copy from a remote to local with multiple streams and `--size-only` (either explicit or implied by the remote, e.g., WebDAV), it is conceivable for an interrupted transfer to not get restarted since it's "size" is correct even if the content is not

Either add an extra null byte that gets truncated or withhold the very last byte until the rest of the transfer is complete. This way, the size of the file always mismatches until completed

#### How to use GitHub

Evan-sp:
… copy

Fix #6206 by withholding the very last byte until the rest of the transfer is complete

#6206

#### Checklist

### DIFF
--- a/fs/operations/multithread.go
+++ b/fs/operations/multithread.go
@@ -69,8 +69,9 @@ func (mc *multiThreadCopyState) copyStream(ctx context.Context, stream int) (err
 		return nil
 	}
 	end := start + mc.partSize
-	if end > mc.size {
-		end = mc.size
+	if end >= mc.size {
+		// Withhold last byte until all streams finish to prevent --size-only skip
+		end = mc.size - 1
 	}
 
 	fs.Debugf(mc.src, "multi-thread copy: stream %d/%d (%d-%d) size %v starting", stream+1, mc.streams, start, end, fs.SizeSuffix(end-start))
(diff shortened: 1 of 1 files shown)


Jwink3101:
Is there an update on this? Plan to merge?

fs/operations/multithread.go
```diff
@@ -178,10 +179,25 @@ func multiThreadCopy(ctx context.Context, f fs.Fs, remote string, src fs.Object,
 		})
 	}
 	err = g.Wait()
-	closeErr := mc.wc.Close()
 	if err != nil {
 		return nil, err
 	}
+
+	// Copy last withheld byte
```
ncw, line 186: This Open will cause another HTTP transaction to the backend and hence quite a lot of extra work / delay for each file.

Ideally what we'd do is fetch the last byte above where we have the connection open, stash it away, and store it here.

ncw (commented):
I think this is a great idea.

The implementation could be made more efficient though - please see inline.

Jwink3101:
I mentioned in the original forum post that an alternative would be to preallocate an extra byte and truncate. Since this is _only_ for local destination, can all local remotes support truncate? That would solve the book-keeping and be simple.

ncw:
Yes that seems like a good idea.

You should be able to truncate the file with OpenWriterAt by opening it again I think.

We could always add truncate to that interface if necessary.

Jwink3101:
Just checking if there has been any progress on this? It is a conceivable scenario and it worries me that we aren't protected. (e.g., downloading WebDAV to local).

@ncw? @Evan-sp

ncw:
I'd like to get the efficiency thing resolved, either by keeping the last byte, or using the preallocate trick.

@Evan-sp do you want to have a go at this? If not how about you @Jwink3101 ?

ncw:
In 5c594fea90429ccc35328a3c387b9d326167cfa9 we implemented transfers to a temporary file name on the local, ftp and sftp backends unless the `--inplace` flag is not used.

I think this therefore fixes the issue that this PR is intended to fix as the file won't be renamed to its proper name unless the download completes properly.

I'm therefore going to close this PR.

Thank you @Evan-sp for doing it and sorry we didn't get it merged.

Issue #5993, opened 2022-02-15
use ETag to verify S3 file transfers


asdffdsazqqq:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/how-does-rclone-verify-file-transfer-s3-to-s3/29261

#### What is your current rclone version (output from `rclone version`)?
```
rclone v1.57.0
- os/version: ubuntu 20.04 (64 bit)
- os/kernel: 5.10.60.1-microsoft-standard-WSL2 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17.2
- go/linking: static
- go/tags: none
```

#### What problem are you are trying to solve?
after file transfer, rclone cannot verify the dest file matches the source file.

currently, rclone relies on
--- the https protocol
--- Content-Md5: for each chunk transferred

local -> s3
1. calculate MD5 hash
2. save `X-Amz-Meta-Md5chksum` and copy source to dest
3. rclone prints `DEBUG : file.txt: md5 = fc873c046f6fdf1384d2cba806e69d6c OK`

s3 -> s3
1. if exist, get `X-Amz-Meta-Md5chksum` from source.
2. save `X-Amz-Meta-Md5chksum` for dest file metadata and copy source to dest.
3. rclone prints `DEBUG : file.txt: md5 = fc873c046f6fdf1384d2cba806e69d6c OK`

s3 -> local
have not tested in detail, can do as needed.

#### How do you think rclone should be changed to solve that?
--- ETag is calculated by s3 provider after file transfer completes. a reliable value to verify file transfers
--- have rclone verify file transfers using ETag.
--- have `rclone check` use ETag, fallback to MD5, if needed.
--- `rclone lsf` - add `--format=E` for ETag
--- `rclone hashsum` - add `ETag`

for local -> s3
1. rclone already has to calculate the MD5 hash before the file transfers starts.
in addition, calculate ETag
2. copy the file from source to dest
3. compare the dest ETag to the source ETag.

for s3 -> s3
1. if the source file had been uploaded by rclone, read the source's `X-Amz-Meta-Md5chksum` and copy that to the dest.
3. after file transfer completes, compare the dest ETag to source ETag.

thanks much,

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
[...]

Issue #523, opened 2016-06-15
s3: Multipart files not checked or syncronized based on Checksum (ETag, additional headers)


robbat2:
Multipart uploads to S3 do not have working checksum-based synchronization or validations. This can be implemented in two different ways:
1. Set additional headers of the known checksums of the file (as a bonus, you can set multiple different checksums on a single file
2. If the chunk size is consistent, the local ETag can be calculated for comparison. The chunk size can also be stored as a header, to ensure later calculations can be repeated, This is possible as documented here: http://www.perlmonks.org/?node_id=1140988

> What is your rclone version (eg output from `rclone -V`)
> rclone v1.29
> 
> Which OS you are using and how many bits (eg Windows 7, 64 bit)
> Gentoo Linux, 64-bit
> 
> Which cloud storage system are you using? (eg Google Drive)
> S3 (Ceph Hammer)
> 
> The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
> rclone  sync .  $REMOTE_NAME:$BUCKET_NAME/  --checksum
> 
> A log from the command with the `-v` flag (eg output from `rclone -v copy /tmp remote:tmp`)
> 
> ```
> rclone  sync .  $REMOTE_NAME:$BUCKET_NAME/      --checksum --dump-headers --dump-bodies
> 2016/06/14 17:09:27 >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> 2016/06/14 17:09:27 HTTP REQUEST
> 2016/06/14 17:09:27 HEAD /$BUCKET_NAME HTTP/1.1
> Host: objects-us-west-1.dream.io
> User-Agent: rclone/v1.29
> Authorization: AWS XXXXXX:Ke4mVHLXl4WUSPoyfVYZ6LoPFiM=
> Date: Wed, 15 Jun 2016 00:09:27 UTC
> 
> ```

2016/06/14 17:09:27 >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2016/06/14 17:09:27 <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
2016/06/14 17:09:27 HTTP RESPONSE
2016/06/14 17:09:27 HTTP/1.1 200 OK
[...]

PR #6303, opened 2022-07-06
Supporting Etag for `check` command for files uploaded to S3 storage using a multipart uploader

bzick:

S3 backend now supports Etag for objects that are chunk-uploaded. To enable Etag, simply enable the `UseMultipartEtag` option, and then specify a `ChunkSize` bytes greater than 0 .

After that, S3 will register a hash function `s3hash<ChunkSize>` (e.g. for `ChunkSize=8Mi` it will be named `s3hash8mi`),
 which returns an ordinary MD5 for all objects smaller than `ChunkSize`, and computes its own checksum otherwise:

```
checkSum = md5bin(md5bin(part1) + md5bin(part2) + md5bin(part3) + ...)
```

In this case, object `Etag` that is returned by S3 will look like this:

```
Etag = hex(checkSum) + "-" + itoa(partsCount)
// 60b725f10c9c85c70d97880dfe8191b3-33
```

For all object checks with enabled `UseMultipartEtag` only hash part of Etag is used.

Don't forget, this functionality **only** works with enabled `UseMultipartEtag` **and** `ChunkSize` set to a value greater than 0.

Now `rclone check` can validate hashes of objects that are chunk-uploaded

TODO: replace old chunk-hash in `s3.uploadMultipart()` method to new `s3hash` 

- https://github.com/rclone/rclone/issues/5993
- https://github.com/rclone/rclone/issues/523 

#### Checklist

- [ ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages). Well, i forgot about that :-(


backend/s3/s3hash/s3hash.go
```diff
@@ -0,0 +1,140 @@
...
+//      // result be like — 60b725f10c9c85c70d97880dfe8191b3-33
+package s3hash
+
+import (
+	"crypto/md5"
+	"encoding"
+	"hash"
+)
+
+type S3Hash struct {
```
rvadim, line 26: What do you think? Here we need a lock in case of use with goroutines?

rvadim (approved):

backend/s3/s3hash/s3hash.go
ncw: No locks needed - hashes aren't expected to be thread-safe

ncw:
I ran the CI.

I think this is a very interesting idea, and we've had attempts at this in the past. The idea of using a differently named hash for each chunk size is an excellent one.

However, it has a flaw in that it needs to know the chunk size of the uploaded files, and I don't think you can read this from the objects themselves. Each chunk needn't even be the same size!

We can assume, that the chunks are uploaded with size --chunk-size, but even then the size of the actual chunks can be different from this to get the files to below 10,000 chunks

https://github.com/rclone/rclone/blob/b5efffee9d53eb6e7801011846cb248feb9c83a7/backend/s3/s3.go#L4114

So for large files the backend might have to register many different hashes and I think that might just be a fatal flaw in this plan.

My concern is there are just too many corner cases for the hash calculated not to match the hash uploaded

- file not uploaded with rclone
- wrong `--chunk-size` used
- large file with expanded chunk size => multiple hash algorithms in play => rclone check only uses one hash
- rclone changes the file size to chunk size algorithm

Maybe this should be using the `-partsCount` info - we could probably reverse this function to give a chunk size used given a default chunk size and partsCount, but registering multiple hashes is unlikely to play well with `rclone check`.

https://github.com/rclone/rclone/blob/b5efffee9d53eb6e7801011846cb248feb9c83a7/fs/chunksize/chunksize.go#L18

I think, if we merge this, it will need to be gated on a flag which is off by default as the potential for trouble is large.

I would be interested to hear your thoughts.

bzick:
Sorry to keep you waiting.
Indeed, using existing parameters is not the best choice. I suggest a few changes:
- Create a new `MultipartEtagChinkSize` parameter. and if it is specified `MultipartEtagChinkSize` and `UseMultipartEtag` then calculate Multipart Etag.
- Make creation of new hash function only at first requirement (in `fs.Hashes()`). Will avoid creating multiple hash functions when declaring multiple s3 backends.
- If number of chunks in Etag does not match - return error, as it happens now.

backend/s3/s3.go
```diff
@@ -1968,9 +1970,20 @@ This should be true, false or left unset to use the default for the provider.
 `,
 			Default:  fs.Tristate{},
 			Advanced: true,
-		}, {
-			Name: "use_presigned_request",
-			Help: `Whether to use a presigned request or PutObject for single part uploads
+		},
```
bzick, line 1973: We have to reindent lines in this file. But i didn't because this merge has another purpose.

bzick:
Fixed notices from linter

asdffdsazqqq:
this is very interesting, 

if you need a beta tester, let me know.

and having a hard time finding more about s3hash.
perhaps you have a weblink from aws website.

thanks much,
jojo

PR #6308, opened 2022-07-10
jobs: add ability to stop group

### LINKED ISSUE #5561: Add functionality to stop an operation using the _group parameter
https://forum.rclone.org/t/no-way-to-stop-a-group/26210

v1.56.0

RClone does not have a way to stop a group. You can run the "job/stop" command can stop a specific job but RClone is lacking similar functionality to stop a group. By "group", I am referring to https://rclone.org/rc/#assigning-operations-to-groups-with-group-value. This could be useful for long running operations that need to be halted and returned to at a later point. 

Add functionality to stop a group. The method would look something like "core/stop" and accept a parameter for the "_group" identifier. It might look similar to https://github.com/rclone/rclone/blob/master/fs/rc/jobs/job.go#L393-L408.

#### How to use GitHub

0evan:
Adds new rc call to stop all running jobs in a group. Fixes #5561

Adds new rc call `job/stopgroup` to stop jobs by _group

#5561

#### Checklist

### DIFF
--- a/fs/rc/jobs/job.go
+++ b/fs/rc/jobs/job.go
@@ -406,3 +406,34 @@ func rcJobStop(ctx context.Context, in rc.Params) (out rc.Params, err error) {
 	job.Stop()
 	return out, nil
 }
+
+func init() {
+	rc.Add(rc.Call{
+		Path:  "job/stopgroup",
+		Fn:    rcGroupStop,
+		Title: "Stop all running jobs in a group",
+		Help: `Parameters:
+
+- group - name of the group (string).
+`,
+	})
+}
+
+// Stops all running jobs in a group
+func rcGroupStop(ctx context.Context, in rc.Params) (out rc.Params, err error) {
+	group, err := in.GetString("group")
+	if err != nil {
+		return nil, err
+	}
+	running.mu.RLock()
+	defer running.mu.RUnlock()
+	for _, job := range running.jobs {
+		if job.Group == group {
+			job.mu.Lock()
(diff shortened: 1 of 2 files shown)


fs/rc/jobs/job.go
```diff
@@ -406,3 +406,34 @@ func rcJobStop(ctx context.Context, in rc.Params) (out rc.Params, err error) {
...
+func rcGroupStop(ctx context.Context, in rc.Params) (out rc.Params, err error) {
+	group, err := in.GetString("group")
+	if err != nil {
+		return nil, err
+	}
+	running.mu.RLock()
+	defer running.mu.RUnlock()
+	for _, job := range running.jobs {
+		if job.Group == group {
+			job.mu.Lock()
```
ncw, line 432: Is taking this mutex necessary? We don't take it in rcJobStop?

ncw (commented):
I think that code looks nice - thank you?

A question about one mutex inline!

fs/rc/jobs/job.go
0evan: Thank you for the review. I believe we take it here in rcJobStop:
https://github.com/rclone/rclone/blob/ea2e951f5869e1a0eb6046e897b437f2e5b40126/fs/rc/jobs/job.go#L399
https://github.com/rclone/rclone/blob/ea2e951f5869e1a0eb6046e897b437f2e5b40126/fs/rc/jobs/job.go#L171-L176

Issue #6197, opened 2022-05-24
onedrive backend - support client credentials grant


gursharan001:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/onedrive-backend-client-credentials-grant/30885

#### What is your current rclone version (output from `rclone version`)?
rclone v1.58.0
- os/version: Microsoft Windows 10 Pro 21H2 (64 bit)
- os/kernel: 10.0.19044.1645 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.17.8
- go/linking: dynamic
- go/tags: cmount

#### What problem are you are trying to solve?
Automatically include onedrive/sharepoint document libraries into rclone sync without requiring human intervention to obtain access tokens

#### How do you think rclone should be changed to solve that?
[onedrive backend](https://github.com/rclone/rclone/blob/master/backend/onedrive/onedrive.go) can be modified to support client credential grant - https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
Do you want to work on this @gursharan001 ?

I think this needs to be part of the config flow when running `rclone config`.

I think https://github.com/rclone/rclone/blob/master/lib/oauthutil/oauthutil.go could be modified fairly easily to do what you need.

gursharan001:
Thank you for the pointer. I will attempt it this weekend.

gursharan001:
clientcredentials has a separate package which defines it's own config type - https://pkg.go.dev/golang.org/x/oauth2/clientcredentials#Config 

We might have to introduce a flag in rclone config to indicate this. what do you think?

ncw:
> clientcredentials has a separate package which defines it's own config type - https://pkg.go.dev/golang.org/x/oauth2/clientcredentials#Config

Looks annoyingly similar, but different :-(

> We might have to introduce a flag in rclone config to indicate this.

> We will also have to change rest of the code in [oauthutil.go](https://github.com/rclone/rclone/blob/master/lib/oauthutil/oauthutil.go) to create and use the right config

[...]

PR #6343, opened 2022-07-26
Implement one drive client credentials

martin-redmaple:
This change implements OAuth Client Credential flow for the OneDrive backend.

OAuth Client Credential flow will allow rclone to use permissions directly associated with the Azure AD Enterprise application, rather that adopting the context of an Azure AD user account.

As a quick fix, this has been implemented within the onedrive backend code. A more elegant solution would be to implement this within the oauthutil module.

This has been discussed in #6197.

https://github.com/rclone/rclone/issues/6197

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -46,7 +46,6 @@ import (
 	"github.com/rclone/rclone/lib/random"
 	"github.com/rclone/rclone/lib/rest"
 	"github.com/youmark/pkcs8"
-	"golang.org/x/oauth2"
 )
 
 const (
@@ -65,12 +64,10 @@ const (
 // Globals
 var (
 	// Description of how to auth for this app
-	oauthConfig = &oauth2.Config{
-		Scopes: nil,
-		Endpoint: oauth2.Endpoint{
-			AuthURL:  "https://app.box.com/api/oauth2/authorize",
-			TokenURL: "https://app.box.com/api/oauth2/token",
-		},
+	oauthConfig = &oauthutil.Config{
+		Scopes:       nil,
+		AuthURL:      "https://app.box.com/api/oauth2/authorize",
+		TokenURL:     "https://app.box.com/api/oauth2/token",
 		ClientID:     rcloneClientID,
 		ClientSecret: obscure.MustReveal(rcloneEncryptedClientSecret),
 		RedirectURL:  oauthutil.RedirectURL,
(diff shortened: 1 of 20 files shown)


ncw:
> As a quick fix, this has been implemented within the onedrive backend code. A more elegant solution would be to implement this within the oauthutil module

I'd rather put this in the oauth module if it is useful elsewhere.

Do you think it will be useful for other backends? Azureblob? others?

martin-redmaple:
Yes - I suspect this could be useful elsewhere, although I have only ever used for accessing OneDrive/Sharepoint back ends. I would imagine that Azureblob access would be able to use this auth flow.

This flow is covered in section [4.4 of the OAUTH2 RFC](https://www.rfc-editor.org/rfc/rfc6749#section-4.4:~:text=example_value%22%0A%20%20%20%20%20%7D%0A%0A4.4.-,Client%20Credentials%20Grant,-The%20client%20can) so it's use is possibly not just limited to MS backends.

ncw:
Do you want to have a go at putting int into oauthutil @martin-redmaple ?

You can add config options to oauth using backends here: 

https://github.com/rclone/rclone/blob/2c78f56d48fa876a8a6367462cb14df896a36b62/lib/oauthutil/oauthutil.go#L78

User configurable options would go here

https://github.com/rclone/rclone/blob/2c78f56d48fa876a8a6367462cb14df896a36b62/lib/oauthutil/oauthutil.go#L414

tooptoop4:
can this be merged?

martin-redmaple:
sorry - no. This is still a work in progress.

ncw:
@martin-redmaple sorry this PR got lost.

I've rebased it and reworked it to run with current rclone and hopefully I'll commit it soon.

I had a question though... Your code sets scopes to `.default` on the inital oauth when using client credentials. However when I tried that it didn't work - I got an error `invalid_scopes` back from the server. I'm just wondering under what conditions the `.default` scopes work? I found a potential error in your code also which might mean that the scope `.default` was never set.

Any insight much appreciated.

Thank you.

ncw:
@martin-redmaple I have fixed up this PR and tested it is working properly. I worked out what was wrong with the scopes so it all looks good now! I'll merge when the CI goes green.

ncw:
Thank you @martin-redmaple

martin-redmaple:
@ncw - sorry for the slow response. Thanks for picking this is. I'll hopefully be able to test the latest `master` on my setup.

PR #6359, opened 2022-08-04
Jottacloud: Implement reading and writing of metadata (file created timestamp)

albertony:

Introduce support for the new Metadata feature in rclone.

There is not much support in the Jottacloud API, but I have added the few that I found that could be useful... 

The most useful addition is the `created` timestamp, as especially on Windows this is a property often presented to user, and by having this in the metadata feature users can copy files from local to Jottacloud and back, and preserve timestamps - not only the `modified` timestamp rclone already preserves, but also the `created` timestamp.

Public links (share urls) have also been added. The existing rclone link command enables share and returns the share link, but there is no easy way to list shared files or check if a file is currently shared or not. By including this as read-only meta data it is now possible to extract this information from file lists.

The full set of supported metadata items:
- `name` (read-only) shows the full path to file on backend, within the user account, i.e. unambiguous path regardless of Jottacloud user/account name and regardless of rclone remote/path setting.
- `btime` is a standard LastCreated timestamp.
- `mtime` is a standard LastModified timestamp, when read it returns the same as existing ModTime.
- `utime` (read-only) is a Jottacloud-specific timestamp which is called "updated" in the api response. It seems to be set at current time whenever the api creates a new revision of a file, i.e. whenever there is an upload or another operation changing the state or content of a file. So similar to mtime but entirely backend-controlled.
- `content-type` (read-only) is MIME type information.
- `public-link` is the full direct download url for a shared file, as returned by rclone link command.
- `public-link-web` is the alternative web download url for a shared file, as previously returned by rclone link command but now only shown in debug log.

Notes:
- Is it necessary with both `public-link` and `public-link-web`? And should it be the complete url, or only the identifier returned by the API?
- ListR implementation does not have the `updated` timestamp available, but does have `created` as well as `modified`. The `updated` is the least important one, and is read-only, so don't think it should matter... Also, ListR does not include mime information already, and this is documented in the backend docs, so I'm just extending that description to also indicate that not all metadata items are present when using `--fast-list`.
- The path manipulation and utility functions are getting a bit messy. I'm working on a refactoring commit on these, but moved it out to a separate PR following this one.

#### Checklist


albertony:
One question, probably to @ncw: Looking at existing implementations, i.e. local, s3 and internetarchive, I'm unsure if I should include mtime as system metadata for the backend or not, duplicating the existing ModTime function in rclone... Seems they do that, so I've done the same in Jottacloud. But what is the benefit?

ncw:
> One question, probably to @ncw: Looking at existing implementations, i.e. local, s3 and internetarchive, I'm unsure if I should include mtime as system metadata for the backend or not, duplicating the existing ModTime function in rclone... Seems they do that, so I've done the same in Jottacloud. But what is the benefit?

Yes you should include it as system metadata.

I know it is duplicated info, but it helps make the metadata more consistent between backends.

`created` is very useful for windows users - they often ask for created time to be synced as you can see it easily in the Windows UI I think. In the unix world, created time is a recent invention and not supported everywhere!

albertony:
Thanks, that is noted!
Would the same apply for mime? I.e. should I set it as metadata item "content-type" (even though in Jottacloud's case its read from a property "mime" in response xml and not content-type response header, and also Jottacloud only supports reading it, not writing it, afaik).

ncw:
> Would the same apply for mime, ie. should I set it as metadata item "content-type"? In Jottacloud's case its read from a property "mime" in response xml and not actual content-type response header, but I don't think that matters. But Jottacloud only supports reading mtime, not writing it (afaik), that reduces the relevancy significantly. But still, I wonder if in general mime should also be included as a metadata key whenever the backend has some sort of mime support?

Yes it should be included if possible. I think the integration tests test for that too, but maybe only if you can set content types.

albertony:
I'm wondering a bit about performance implications, also in light of the recent issue https://github.com/rclone/rclone/pull/6393 - though I don't think the same problem handled there is present in Jottacloud backend.

The Update function was changed to now always perform an extra get request at the end. The upload response itself contains main metadata properties (size, md5 and modTime) which was previously set back to the object, but it does not contain all the metadata information - specifically the createTime and updateTime properties. Therefore I now changed to an extra read. But can we avoid this? E.g. can we detect if we are in a "metadata-aware operation" where this is needed, and if not just update the object as before without reading the additional metadata?

albertony:
I think the result object from the Update backend function is returned here:

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/fs/operations/operations.go#L507

And then here, where it is just ignored:

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/fs/operations/operations.go#L1934

But does this mean it is safe to return an incomplete/outdated object from backend? Or is the returned object used somewhere else? Or should we by definition always return a valid updated and complete object?

ncw:
> I'm wondering a bit about performance implications, also in light of the recent issue #6393 - though I don't think the same problem handled there is present in Jottacloud backend.

Its OK for metadata to use a bit more memory. It won't be much. The problem with that issue was that it was pinning the massive XML response from S3 in memory by using a small part of it.

> The Update function was changed to now always perform an extra get request at the end. The upload response itself contains main metadata properties (size, md5 and modTime) which was previously set back to the object, but it does not contain all the metadata information - specifically the createTime and updateTime properties. Therefore I now changed to an extra read. But can we avoid this? E.g. can we detect if we are in a "metadata-aware operation" where this is needed, and if not just update the object as before without reading the additional metadata?

I think that is fine. Update is only called when uploading objects and if that takes a fraction of a second longer that is not a problem.


> I think the result object from the Update backend function is returned here:
> 
> https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/fs/operations/operations.go#L507
> 
> And then here, where it is just ignored:
> 
> https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/fs/operations/operations.go#L1934
> 
> But does this mean it is safe to return an incomplete/outdated object from backend? Or is the returned object used somewhere else? Or should we by definition always return a valid updated and complete object?

Update is expected to update the metadata of the object it is working on. Rclone will ask for hashes with the updated object so it is important it is up to date after an upload.

There are other parts of rclone (like the VFS layer) which will stash the object for later use.

Where are we at with this PR? Shall we try to merge it for 1.63?

ncw:
Give me a nudge when you want me to look at this :-) I'm using github notifications exclusively to manage stuff now-a-days so I need a post for it to reappear on my notifications!

albertony:
@ncw I've revived this PR now, and decided to remove the `name`, `public-link` and `public-link-web` entries to reduce number of decision points (which was partly why the PR stalled back in the day), but also I figured it's best to just include the minimum set of metadata entries which we know can actually be useful - and its easy to add the other ones later if there should be a need. I've left the changes as additional commits for now, but will squash them at the end, if that's ok with you (i.e. it isn't too noisy when doing the review or something). So; when it suits you - please review. 😅

docs/content/jottacloud.md
```diff
@@ -233,16 +233,17 @@ them. Generally you should avoid these, unless you know what you are doing.
...
 
 Note that the implementation in Jottacloud always uses only a single
 API request to get the entire list, so for large folders this could
 lead to long wait time before the first results are shown.
 
-Note also that with rclone version 1.58 and newer information about
-[MIME types](/overview/#mime-type) are not available when using `--fast-list`.
+Note also that with rclone version 1.58 and newer, information about
+[MIME types](/overview/#mime-type) and metadata item [utime](#metadata)
+are not available when using `--fast-list`.
```
ncw, line 246: So if you need metadata you can't use `ListR` - is that correct? That is quite a limitation. We should probably say here use `--disable ListR` to work around or something like that?

backend/jottacloud/jottacloud.go
```diff
@@ -92,6 +92,33 @@ func init() {
...
+					Help:    "Time of file birth (creation), read from rclone metadata",
+					Type:    "RFC 3339",
+					Example: "2006-01-02T15:04:05.999999999Z07:00",
+				},
+				"mtime": {
+					Help:    "Time of last modification, read from rclone metadata",
+					Type:    "RFC 3339",
+					Example: "2006-01-02T15:04:05.999999999Z07:00",
+				},
+				"utime": {
```
ncw, line 108: We should probably put `utime` in the standard table?

https://rclone.org/docs/#standard-system-metadata

ncw (commented):
This is looking good :-)

I put a couple of minor questions inline.

Also

I think this is missing the metadata flags in Fs.Features()?

This will enable more integration tests so you'll need to run them again!

backend/jottacloud/jottacloud.go
```diff
@@ -972,6 +1001,9 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		CanHaveEmptyDirectories: true,
 		ReadMimeType:            true,
 		WriteMimeType:           false,
+		ReadMetadata:            true,
+		WriteMetadata:           true,
+		UserMetadata:            false,
```
albertony, line 1006: > I think this is missing the metadata flags in Fs.Features()?

@ncw did you miss these, or did I miss something? 😄

docs/content/jottacloud.md
albertony: > So if you need metadata you can't use ListR - is that correct?

Not quite. You'll get the `btime` (created) and `mtime` (modified) timestamps, just not `utime` (updated) which I don't see as that big of a limitation. It does not have `content-type` (MIME) either, but that's already a documented limitation with ListR so it just extends to the metadata section.

backend/jottacloud/jottacloud.go
albertony: Sure.
Jottacloud uses term "updated", which I think is confusing to understand what represents compared to the last modified timestamp, so I suggest we make the "u" stand for "uploaded" and document it as "Time of file upload"? Transfer to a local system which happens to support it would for a user be more of a "download", but not a huge problem I think 😅 , although I'm open for other suggestions.

backend/jottacloud/jottacloud.go
ncw: I missed them sorry - I blame my old, tired eyeballs ;-)

backend/jottacloud/jottacloud.go
ncw: Yes `utime` == uploaded time sounds perfect to me.

> Transfer to a local system which happens to support it would for a user be more of a "download", but not a huge problem I think 😅 , although I'm open for other suggestions.

He he! Happy with `utime`!

ncw (approved):
Looks perfect now :-)

Do you want to squash all the commits? Or squash the jotta commits and leave the docs one for two commits?

albertony:
> squash the jotta commits and leave the docs one for two commits?

I have done this now. Will merge as soon as the checks are all green.

PR #6366, opened 2022-08-08
dlna: specify SSDP interface names from command line

simonbos:

Enabling a user to specify the interfaces used for DLNA (SSDP). This is especially useful when serving DLNA on a machine with many virtual network interfaces (for example when also running Kubernetes) for which you don't wish to advertise DLNA.

No.

#### Checklist

### DIFF
--- a/cmd/serve/dlna/dlna.go
+++ b/cmd/serve/dlna/dlna.go
@@ -51,7 +51,10 @@ files that they are not able to play back correctly.
 		f := cmd.NewFsSrc(args)
 
 		cmd.Run(false, false, command, func() error {
-			s := newServer(f, &dlnaflags.Opt)
+			s, err := newServer(f, &dlnaflags.Opt)
+			if err != nil {
+				return err
+			}
 			if err := s.Serve(); err != nil {
 				return err
 			}
@@ -92,17 +95,32 @@ type server struct {
 	vfs *vfs.VFS
 }
 
-func newServer(f fs.Fs, opt *dlnaflags.Options) *server {
+func newServer(f fs.Fs, opt *dlnaflags.Options) (*server, error) {
 	friendlyName := opt.FriendlyName
 	if friendlyName == "" {
 		friendlyName = makeDefaultFriendlyName()
+	}
+	interfaces := make([]net.Interface, 0, len(opt.InterfaceNames))
+	for _, interfaceName := range opt.InterfaceNames {
+		var err error
+		intf, err := net.InterfaceByName(interfaceName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve interface name '%s': %w", interfaceName, err)
+		}
+		if !isAppropriatelyConfigured(*intf) {
+			return nil, fmt.Errorf("interface '%s' is not appropriately configured (it should be UP, MULTICAST and MTU > 0)", interfaceName)
+		}
+		interfaces = append(interfaces, *intf)
+	}
+	if len(interfaces) == 0 {
+		interfaces = listInterfaces()
 	}
 
 	s := &server{
 		AnnounceInterval: 10 * time.Second,
 		FriendlyName:     friendlyName,
 		RootDeviceUUID:   makeDeviceUUID(friendlyName),
-		Interfaces:       listInterfaces(),
+		Interfaces:       interfaces,
 
 		httpListenAddr: opt.ListenAddr,
 
@@ -138,7 +156,7 @@ func newServer(f fs.Fs, opt *dlnaflags.Options) *server {
 			http.FileServer(data.Assets))))
 	s.handler = logging(withHeader("Server", serverField, r))
 
-	return s
+	return s, nil
 }
 
 // UPnPService is the interface for the SOAP service.
--- a/cmd/serve/dlna/dlnaflags/dlnaflags.go
+++ b/cmd/serve/dlna/dlnaflags/dlnaflags.go
@@ -23,16 +23,18 @@ logging of all UPNP traffic.
 
 // Options is the type for DLNA serving options.
 type Options struct {
-	ListenAddr   string
-	FriendlyName string
-	LogTrace     bool
+	ListenAddr     string
+	FriendlyName   string
+	LogTrace       bool
+	InterfaceNames []string
 }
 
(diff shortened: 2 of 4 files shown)


cmd/serve/dlna/dlna.go
```diff
@@ -97,12 +97,25 @@ func newServer(f fs.Fs, opt *dlnaflags.Options) *server {
...
+		var err error
+		tmp, err := net.InterfaceByName(interfaceName)
+		if err != nil {
+			fs.Errorf(nil, "Failed to resolve interface name: %v", err)
+			continue
+		}
+		interfaces = append(interfaces, *tmp)
+	}
+	if len(interfaces) == 0 {
+		interfaces = listInterfaces()
```
simonbos, line 111: `listInterfaces` also contains some logic for checking whether the interfaces are actually UP, multicast etc. Should we also perform this validation for the provided interfaces or trust the user?

ncw: What happens if the interface is DOWN for example? Does it give an error that is sensible? Or ignore it until the interface is UP? Or ignore it forever?

cmd/serve/dlna/dlnaflags/dlnaflags.go
```diff
@@ -45,6 +47,7 @@ func addFlagsPrefix(flagSet *pflag.FlagSet, prefix string, Opt *Options) {
 	flags.StringVarP(flagSet, &Opt.ListenAddr, prefix+"addr", "", Opt.ListenAddr, "The ip:port or :port to bind the DLNA http server to")
 	flags.StringVarP(flagSet, &Opt.FriendlyName, prefix+"name", "", Opt.FriendlyName, "Name of DLNA server")
 	flags.BoolVarP(flagSet, &Opt.LogTrace, prefix+"log-trace", "", Opt.LogTrace, "Enable trace logging of SOAP traffic")
+	flags.StringArrayVarP(flagSet, &Opt.InterfaceNames, prefix+"ifnames", "", Opt.InterfaceNames, "The interfaces to use for SSDP")
```
ncw, line 50: A `StringArray` is so that you can repeat the flag, so `--interface eth0` `--interface eth1`

So `--ifnames` is the wrong name for this flag, it should singular.

I'd suggest `--interface` as being unambigous, so this needs to be something like

    flags.StringArrayVarP(flagSet, &Opt.InterfaceNames, prefix+"interface", "", Opt.InterfaceNames, "The interface to use for SSDP (repeat as necessary)")

ncw (commented):
Looks like a nice change - thank you :-)

See comments inline.

cmd/serve/dlna/dlna.go
simonbos: Good question. In my environment the interface seems to be ignored until the interface is UP; giving the log message `<6>INFO  : DLNA server on :7879: Started SSDP on enp0s31f6` (and only starting to work when starting the interface). However, I don't know if this is the intended behavior of `anacrolix/dms` ... In my opinion checking the same flags as in `listInterfaces` would be best.

cmd/serve/dlna/dlnaflags/dlnaflags.go
simonbos: Ok, good to know will adapt this!

cmd/serve/dlna/dlna.go
```diff
@@ -97,12 +97,25 @@ func newServer(f fs.Fs, opt *dlnaflags.Options) *server {
 	if friendlyName == "" {
 		friendlyName = makeDefaultFriendlyName()
 	}
+	interfaces := make([]net.Interface, 0, len(opt.InterfaceNames))
+	for _, interfaceName := range opt.InterfaceNames {
+		var err error
+		tmp, err := net.InterfaceByName(interfaceName)
+		if err != nil {
+			fs.Errorf(nil, "Failed to resolve interface name: %v", err)
+			continue
```
simonbos, line 106: Should we return here instead, i.e. if the user provided configuration is wrong? (Also if the user-provided interface(s) do(es) not contain the correct flags.)

cmd/serve/dlna/dlna.go
ncw: If it starts working when the interface comes up that can be very useful if you are starting the service at boot when the interface might not be up yet. Not a real problem nowadays though with the advent of things like systemd so if you want to do the same checks then that sounds ok to me.

simonbos: Agreed! I added the checks now.

cmd/serve/dlna/dlna.go
```diff
@@ -97,12 +100,29 @@ func newServer(f fs.Fs, opt *dlnaflags.Options) *server {
...
+	for _, interfaceName := range opt.InterfaceNames {
+		var err error
+		intf, err := net.InterfaceByName(interfaceName)
+		if err != nil {
+			fs.Errorf(nil, "Failed to resolve interface name: %v", err)
+			return nil
+		}
+		if !isAppropriatelyConfigured(*intf) {
+			fs.Errorf(nil, "Interface '%s' is not appropriately configured (it should be UP, MULTICAST and MTU > 0)", interfaceName)
+			return nil
```
simonbos, line 113: So I changed the `continue` to `return nil` with a nilcheck above; OK?

ncw: I'd probably change the function to return a `(*server, error)` and return the errors properly, so

```go
		if err != nil {
			return nil, fmt.Errorf("failed to resolve interface name: %w", err)
		}
```

That would be more go-like I think as checking `newServer` for `nil` looks a bit odd.

simonbos:
Ok, I don't have any more questions; thanks for all the usefull comments :smile: Merge?

ncw (commented):
Looks perfect now - thank you.

I'll run the CI and if it passes I'll merge - thank you :-)

ncw:
Looks like the DLNA tests need updating

simonbos:
Yes, I forgot to run the tests locally after changing the `newServer` output signature. With the latest change, `go test -v` (in the `cmd/serve/dlna` directory) runs succesfully.

ncw (approved):
LGTM - thank you :-)

Will merge now.

Issue #6381, opened 2022-08-15
Print a reasonable ETA


zzjjbb:
#### What is your current rclone version (output from `rclone version`)?

rclone v1.59.1
- os/version: centos 7.9.2009 (64 bit)
- os/kernel: 3.10.0-1160.71.1.el7.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.18.5
- go/linking: static
- go/tags: none

#### What problem are you are trying to solve?

Print a reasonable ETA, like >1yr when it is very slow rather than a negative number
![image](https://user-images.githubusercontent.com/31069326/184716857-5143bc87-f4a3-4e5a-a88a-7dee225d0b2a.png)

#### How do you think rclone should be changed to solve that?

When the speed is too low, just print a >1yr or similar in the ETA. 

Also, only showing 2 units seems better to read. For example, when the ETA is longer than 1d, only show `d` and `h`; when longer than 1h, only show `h` and `m`; etc.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

PR #6385, opened 2022-08-19
Avoid negative ETA values for very slow speeds

albertony:

Part 1: Integer overflow would lead to ETA such as "-255y7w4h11m22s966ms", as reported in #6381. Now the value will be clipped at the maximum "292y24w3d23h47m16s", and it will be shown as "-".
- I would like to show the Unicode symbol for infinity, "∞", for this situation, but I'm worried it will cause issues in some consoles (especially Windows, with code page nightmare etc). Don't know if there are any other comparable uses of Unicode symbols in output from rclone? Tree command probably, but maybe that is a somewhat unconventional command with limited use, compared to the stats output, so less important with compatibility. Maybe "**" could serve as an ascii version of the infinity symbol, or perhaps it will look a bit "noisy". I landed on just showing "-", which is the same as already used for any unknown/invalid/completed ETA.

Part 2: As suggested in the issue, it does not make much sense to report hours, minutes, and even seconds when the duration is  several years. I've changed to show a shorter ETA string, at most 3 significant parts: "292y24w3d", "24w3d23h", "3d23h47m", "23h47m16s". This means when ETA is days or more, then precision will be sacrificed, but anything within 24 hours will be full precision (hours, minute, second - eta is only second precision).

#6381

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -229,6 +229,11 @@ func (s *StatsInfo) totalDuration() time.Duration {
 	return s.oldDuration + timeRanges.total()
 }
 
+const (
+	etaMaxSeconds = (1<<63 - 1) / int64(time.Second)           // Largest possible ETA as number of seconds
+	etaMax        = time.Duration(etaMaxSeconds) * time.Second // Largest possible ETA, which is in second precision, representing "292y24w3d23h47m16s"
+)
+
 // eta returns the ETA of the current operation,
 // rounded to full seconds.
 // If the ETA cannot be determined 'ok' returns false.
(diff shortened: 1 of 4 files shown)


ncw:
> #### What is the purpose of this change?
> Part 1: Integer overflow would lead to ETA such as "-255y7w4h11m22s966ms", as reported in #6381. Now the value will be clipped at the maximum "292y24w3d23h47m16s", and it will be shown as "-".

Nice!

> * I would like to show the Unicode symbol for infinity, "∞", for this situation, but I'm worried it will cause issues in some consoles (especially Windows, with code page nightmare etc). Don't know if there are any other comparable uses of Unicode symbols in output from rclone?

There are. We use the elipsis symbol `…` when truncating names in the stats - I remember a few reports from years ago of these not working, but I think we should really assume that people are using UTF-8 capable terminals - we certainly assume all file names can be UTF-8.

> Part 2: As suggested in the issue, it does not make much sense to report hours, minutes, and even seconds when the duration is several years. I've changed to show a shorter ETA string, at most 3 significant parts: "292y24w3d", "24w3d23h", "3d23h47m", "23h47m16s". This means when ETA is days or more, then precision will be sacrificed, but anything within 24 hours will be full precision (hours, minute, second - eta is only second precision).

Sounds perfectly reasonable.

Will take a look at code now :-)

ncw (approved):
Code looks great - thank you :-) Please merge.

Another pull which fell through the gaps sorry :-(

albertony:
> Another pull which fell through the gaps sorry :-(

:zero: problem. I'm normally quite patient, and I think I can understand some of your situation.. I ping you, like I just did, when I think something deserves a little bit of attention. 😄  Thanks.

PR #6387, opened 2022-08-20
Add staticcheck linters to build pipeline

albertony:

Extend linting in build pipeline to include checks from the Staticcheck tool, see https://staticcheck.io.

##### About staticcheck and golangci-lint

The [Staticcheck](<https://staticcheck.io>) tool is the default golang linter in Visual Studio Code dev environment with the Go extension, and from my experience it gives many relevant suggestions that the existing linting performed by our build pipeline does not report. The existing ci linting in rclone is based on [golangci-lint](https://golangci-lint.run), which supports running many different linters - and staticcheck is one of them. This means we can enable staticcheck in the existing golangci-lint configuration, instead of running it as a separate build step or something. This will probably be much easier to implement, and no additional build steps/scripts/configuration files etc.

But, the staticcheck feature of golangci-lint is not the same as the original/stand-alone staticcheck:

> It's a set of rules from staticcheck. It's not the same thing as the staticcheck binary. The author of staticcheck doesn't support or approve the use of staticcheck as a library inside golangci-lint.

Staticcheck is in fact a binary that somewhat similar to golangci-lint incorporates a set of different linter libraries used to perform the actual checks. The full overview of checks it performs is here: https://staticcheck.io/docs/checks. The linter that golangci-lint calls "staticcheck" is only a subset of these, but the rest are also available as separate configurable [golangci-lint linters](https://golangci-lint.run/usage/linters/);
- Checks with [SA-prefix](https://staticcheck.io/docs/checks#SA) are labelled `staticcheck`, supported in golangci-lint as linter "staticcheck".
- Checks with [ST-prefix](https://staticcheck.io/docs/checks#ST) are labelled `stylecheck`, supported in golangci-lint as linter "stylecheck".
    - Note that some of these are disabled by default in staticcheck binary, enabling "stylecheck" in golangci-lint will by default enable all so to match staticcheck one must configure the [checks option](https://staticcheck.io/docs/configuration/options/#checks).
    - One that is enabled is the following, which is very relevant to have in rclone: `ST1005: error strings should not be capitalized`.
- Checks with [S-prefix](https://staticcheck.io/docs/checks#S) are labelled `simple`, supported in golangci-lint as linter "gosimple".
- Check U1000 is implemented in `unused`, supported in golangci-lint as linter "unused".
   - It is not documented in the staticcheck site, but in golangci-lint docs it is described as reporting unused constants, variables, functions and types.

I have enabled the first three: `staticcheck`, `gosimple` and `stylecheck`. Configured `stylecheck` to only consider the same limited set of checks as the staticcheck tool does, as mentioned above. Did not enable `unused`, because there are currently some 15 occurrences of internal functions that are never called in the rclone codebase.

##### Excludes

Since staticcheck reports many issues not reported by the existing set of linters we are running, we need to:
1. Fix as many of the reported issues as we can.
2. Add excludes for any false positives, or issues that we otherwise don't want to fix, or don't want reported.

Regarding 1:

Issue https://github.com/rclone/rclone/issues/6273 contains information of work done, with links to individual issues fixing specific staticcheck-reported problems. This PR also includes a couple of additional minor fixes.

Regarding 2:

There will typically always be some issues that must be ignored, i.e. actual false positives. One example is code passing a `nil` context, which staticcheck reports should instead be `context.TODO`, but when this is in a unit test specifically testing `nil` contexts we obviously can't do that. Such issues must be permanently excluded from the lint results.

Other issues may be relevant to fix, but not important enough to "red flag" a PR, for example, they may be "known issues" that are not easy to fix but should not be completely forgotten either, or it may not be obvious if they are false positives or not and we need developers/time to find out. These issues could be relevant to show in the lint result, but as information, or warning, not as a regular lint error (though some may argue that, like with unit tests, either it should be an error or not and not build up a backlog of "unknowns"). If staticcheck proves to be too "noisy", we could also consider the possibility to make it purely advisory. Golang-ci does actually support [configuring severity](https://golangci-lint.run/usage/configuration/#severity-configuration), but unfortunately it does not affect the exit code - i.e. as long as there are any issues found, regardless if none are reported as errors, but only warning or notice, the golangci-lint execution will still fail with "Error: issues found". There are several issues/PRs at [golangci-lint](https://github.com/golangci/golangci-lint) discussing this, but for now it seems we either have to ignore issues entirely or let them fail (and fix them asap).

To ignore issues there are two main strategies:
- Configure excludes in `.golangci.yml`.
   - In [linter-specific settings](https://golangci-lint.run/usage/linters/#staticcheck) configuring specific staticcheck [checks](https://staticcheck.io/docs/configuration/options/#checks), or in issue settings [exclude/exclude-rules](https://golangci-lint.run/usage/configuration/#issues-configuration).
   - There are many possible ways to achieve similar things, where one can choose to consider path, text (issue message), source code, regex-matching, specific linters, etc.
- Write special ignore comments in the source code.
   - Note that golangci-lint does not (currently) consider the same format as the original staticcheck (as discussed in several [golangci-lint](https://github.com/golangci/golangci-lint) issues):
      - Stand-alone staticcheck supports ignoring specific checks using format `//lint:ignore SA1012 explain reason here` ([ref](https://staticcheck.io/docs/configuration/#ignoring-problems)).
      - Golangci-lint staticcheck does not support the above, but can disable all checks from staticcheck for a given line using format `//nolint:staticcheck // Explain reason here` ([ref](https://golangci-lint.run/usage/false-positives/#nolint-directive)). Note that this disables all staticcheck issues on the line, while the lint:ignore format supported by staticcheck itself allows ignoring specific lint checks.
   - To support both tools, i.e. both golangci-lint linting on GitHub and staticcheck linting done in dev environment from VSCode etc, both must be specified. Note that my experience is that they must be on separate lines, in a comment section before the line containing the issue to be ignored, i.e. cannot be combined at the end of the code line despite I've seen examples of this claiming to work. Working example:
      ```
      //lint:ignore SA1012 false positive when running staticcheck, we want to test passing a nil Context and therefore ignore lint suggestion to use context.TODO
      //nolint:staticcheck // Don't include staticcheck when running golangci-lint to avoid SA1012
      config := GetConfig(nil)
      assert.Equal(t, globalConfig, config)
      ```
- See golangci-lint/issues/741 for more info and other alternatives.

TODO:
- Look into reported issues that are not obvious - fix code, or exclude - but only if we can conclude it is really a false positive.
   - This mainly boils down to the following:
      - argument should be pointer-like to avoid allocations (SA6002)
   https://github.com/rclone/rclone/blob/58b327a9f6382f1c104c7ed8e1355ad14806d20d/backend/crypt/cipher.go#L242
      - infinite recursive call (SA5007)
   https://github.com/rclone/rclone/blob/58b327a9f6382f1c104c7ed8e1355ad14806d20d/fs/pacer.go#L76
      - argument err is overwritten before first use (SA4009), backend/onedrive/onedrive.go:1671:3: assignment to err
   https://github.com/rclone/rclone/blob/58b327a9f6382f1c104c7ed8e1355ad14806d20d/backend/onedrive/onedrive.go#L1670
   - 🆘 I really need help on this one.
- So, how do we choose to define excludes?
   - Some arguments I can think of:
      - Do we want to maintain excludes spread around the source code, or do we want to bloat the .golangci.yml with exclude rules?
      - In comment it pinpoints the exact line, while a .golangci.yml exclude-rule is more or less global, or at least one entire file.
      - In comment it must be repeated for each occurrence, though a special comment can be added at top of a file to cover the entire file.
      - In comment it is easier to lose overview of what we have actually excluded, but (see next point)...
      - With comment-based excludes, staticcheck has dedicated checks for unnecessary excludes and will report them as errors, so it is easy to maintain them and keep them relevant! - But not sure if the golangci-lint variant supports the same?
      - Exclude-rules in .golangci.yml only affects golangci-lint, so separate runs with staticcheck, such as from VSCode, will not consider these.. Which may be seen as positive or negative... Negative if the excludes are false positives, which will just be noise in VSCode etc. Positive if not definitively classified as false positives, because in a VSCode dev env you get the complete list of possible problems and can look into them. 
   - For "temporary" excludes, e.g. some of the ones reported when I now first enable these checks which I can't figure out myself, I leaning towards adding them as excludes in the .golangci.yml. This makes them more visible, we have the full overview in the yml file, they will all still be reported in VSCode dev env, i.e. more likely to be picked up and actually considered and not just forgotten.
   - For "permanent" excludes, i.e. reported issues that have been investigated and found to be false positives, I think exclude by code comments is good. Make a comment that covers both stand-alone staticcheck as well as golangci-lint, and we can all stop thinking about this one.
- Consider enable the `unused` linter.
   - Must then first consider the about 15 occurrences of unused functions in the codebase.

#6273

#### Checklist

- [ ] ~I have added documentation for the changes if appropriate.~
   - Some of these refactor, build, ci etc are hard to decide on what prefix to use, e.g. "build:", "test:" or what.. Let me know if I should change some of them.. 


albertony:
@ncw: Do you have any input to the staticcheck reported issues in cipher, pacer, and ondrive, as described above (and also in the original issue)? Can we get these out of the way immediately - i.e. are you able to see if these are false positives or if there is a code change that will improve them and silence the linter?

albertony:
> Did not enable `unused`, because there are currently some 15 occurrences of internal functions that are never called in the rclone codebase.

The three linters `deadcode`, `structcheck` and `varcheck` we have been using are now marked as deprecated, and replaced by `unused`. From latest lint run:

```
level=warning msg="[runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
level=warning msg="[runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
level=warning msg="[runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
```

So this makes me want to reconsider the decision to not enable `unused`. Then the question is what to do with the 15 reported unused functions in the codebase?

albertony:
@ncw Do you have a suggestion what to do with the reported "unused" issues? And the three cipher, pacer and onedrive issues mentioned before? If we could conclude on these, then I think we should be able to get this merged quite soon, as we've discussed in the linked issue..

albertony:
Rebased on master and force pushed.

Then pushed a new commit where I've fixed the onedrive issue.

ncw:
> Do you have a suggestion what to do with the reported "unused" issues?

Some of them look like they need deleting. The VFS ones are for debugging so we should leave those.

> And the three cipher, pacer and onedrive issues mentioned before?

Those are very odd! I don't understand the cipher one. The pacer one looks correct, but that means that code doesn't work. Did it ever get called? The OneDrive one probably needs to explicitly ignore the incoming err (_ = err) but it would be worth checking the walk contract.

> If we could conclude on these, then I think we should be able to get this merged quite soon, as we've discussed in the linked issue..

Great!

albertony:
> The OneDrive one probably needs to explicitly ignore the incoming err (_ = err) but it would be worth checking the walk contract.

Yes, based on other uses of Walk, I suggested in my recent commit the following fix:

```
	err := walk.Walk(ctx, f, "", true, -1, func(path string, entries fs.DirEntries, err error) error {
		if err != nil {
			fs.Errorf(f, "Failed to list %q: %v", path, err)
			return nil
		}
```

> The pacer one looks correct, but that means that code doesn't work. Did it ever get called?

Searching the code base I can't find any calls to `ModifyCalculator`, so I can't imagine its in use.

> I don't understand the cipher one.

Yes, I'm definitely on thin ice here, but some googling seem to agree that it is better to actually use pointer than a slice when getting/putting on a sync.Pool: "A slice is not a pointer. A slice is a struct consisting of three fields: length, capacity, and a pointer to data."

https://github.com/dominikh/go-tools/issues/302

https://deepsource.io/gh/beego/beego/issue/SCC-SA6002/occurrences

albertony:
I went for it and hammered through a complete set of suggested changes! Then at least we have actual code lines to pin further discussions on.. :)

The most questionable change is the one in cipher! It seems to me that use of slice with sync.Pool is definitely considered suboptimal, and that one should use pointers. The normal suggestion is to use pointer to bytes.Buffer, but since we here are working with strictly fixed sized blocks then I was thinking we might as well use pointer to array? 

@ncw Let me know what you think. Feel free to do a full review.

ncw:
> > The OneDrive one probably needs to explicitly ignore the incoming err (_ = err) but it would be worth checking the walk contract.
> 
> Yes, based on other uses of Walk, I suggested in my recent commit the following fix:
> 
> ```
> 	err := walk.Walk(ctx, f, "", true, -1, func(path string, entries fs.DirEntries, err error) error {
> 		if err != nil {
> 			fs.Errorf(f, "Failed to list %q: %v", path, err)
> 			return nil
> 		}
> ```

That looks OK. That means it will continue the cleanup but print an error.

> > The pacer one looks correct, but that means that code doesn't work. Did it ever get called?
> 
> Searching the code base I can't find any calls to `ModifyCalculator`, so I can't imagine its in use.

Ha!

I think this is supposed to say

```
@@ -73,7 +73,7 @@ func (p *Pacer) SetCalculator(c pacer.Calculator) {
 // ModifyCalculator calls the given function with the currently configured
 // Calculator and the Pacer lock held.
 func (p *Pacer) ModifyCalculator(f func(pacer.Calculator)) {
-	p.ModifyCalculator(func(c pacer.Calculator) {
+	p.Pacer.ModifyCalculator(func(c pacer.Calculator) {
 		switch _c := c.(type) {
 		case *logCalculator:
 			f(_c.Calculator)
```

> > I don't understand the cipher one.
> 
> Yes, I'm definitely on thin ice here, but some googling seem to agree that it is better to actually use pointer than a slice when getting/putting on a sync.Pool: "A slice is not a pointer. A slice is a struct consisting of three fields: length, capacity, and a pointer to data." ([dominikh/go-tools#302](https://github.com/dominikh/go-tools/issues/302)). See:
> 
> * https://staticcheck.io/docs/checks#SA6002
> * https://deepsource.io/gh/beego/beego/issue/SCC-SA6002/occurrences

Ah I see. I think this is probably marginal as the slice struct is quite small (3 words) vs a pointer (1 word) but I get their point.

I'll see how the code turned out!

fs/pacer.go
```diff
@@ -70,20 +70,6 @@ func (p *Pacer) SetCalculator(c pacer.Calculator) {
 	p.Pacer.SetCalculator(c)
 }
 
-// ModifyCalculator calls the given function with the currently configured
-// Calculator and the Pacer lock held.
-func (p *Pacer) ModifyCalculator(f func(pacer.Calculator)) {
-	p.ModifyCalculator(func(c pacer.Calculator) {
```
ncw, line 76: We could fix this according to my idea if you want

backend/crypt/cipher.go
```diff
@@ -191,7 +191,7 @@ func newCipher(mode NameEncryptionMode, password, salt string, dirNameEncrypt bo
 		dirNameEncrypt: dirNameEncrypt,
 	}
 	c.buffers.New = func() interface{} {
-		return make([]byte, blockSize)
```
ncw, line 194: ingenious using an array rather than a slice. This definitely uses less allocations :+1:

ncw (approved):
These all look fine  - well done :-)

We could fix up the pacer thing if you want that is the only thing I might change but you could merge it like this if you want!

fs/pacer.go
albertony: Yes, I will do that!

albertony:
Great! I've fixed the pacer according to your idea, I think that looks like a very good solution.

I'll wait for the checks to finish, and then probably just merge it now. Then I'll also close the related issue (https://github.com/rclone/rclone/issues/6273), even if it lists the jwt/jws PR that is not concluded yet.

PR #6389, opened 2022-08-20
Replace deprecated ioutil

albertony:

As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code.

TLDR;

Most changes are simple search-replace of ioutil function call to identical io or os function call. The only exception is ReadDir:
- ioutil had ReadDir returning FileInfo
- os has ReadDir returning DirEntry

The only place where this lead to some extra thought was regarding the Oser interface in rclone. It has a ReadDir function which for the realOs implementation was one-liner mapping to ioutil.ReadDir. I landed on the smallest possible change: Changing implementation of the realOs.ReadDir to use os.ReadDir and convert from DirEntry to FileInfo, as suggested in the osutil.ReadDir documentation. I.e. as little change to rclone code as possible, I think. But there is some mismatch with the standard package the Oser interface is emulating(?) (though there was before as well, since it used ioutil, and not os).

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -12,9 +12,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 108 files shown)


ncw (approved):
> The only place where this lead to some extra thought was regarding the Oser interface in rclone. It has a ReadDir function which for the realOs implementation was one-liner mapping to ioutil.ReadDir. I landed on the smallest possible change: Changing implementation of the realOs.ReadDir tThe only place where this lead to some extra thought was regarding the Oser interface in rclone. It has a ReadDir function which for the realOs implementation was one-liner mapping to ioutil.ReadDir. I landed on the smallest possible change: Changing implementation of the realOs.ReadDir to use os.ReadDir and convert from DirEntry to FileInfo, as suggested in the osutil.ReadDir documentation. I.e. as little change to rclone code as possible, I think.

That looks great to me

> But there is some mismatch with the standard package the Oser interface is emulating(?) (though there was before as well, since it used ioutil, and not os).

This might need fixing at some point, but note that that interface is designed to slot into the webdav server code for example.

Rclone should really be providing an io/fs.FS interface, probably two, one for plain rclone/fs and one for rclone/VFS so we can fit into the go ecosystem better. This is a step towards that :-)

I'm going to merge this now before it bitrots!

Thank you

PR #6396, opened 2022-08-25
Doc update: Install with docker: interactive use

mitthu:
Interactive use from `docker` requires the `-it` flag. Otherwise the command files.

See: https://docs.docker.com/engine/reference/commandline/cli/#option-types

No

#### Checklist

- [n/a] I have added tests for all changes in this PR if appropriate.

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -176,7 +176,6 @@ docker run --rm -it \
 # make sure the config is ok by listing the remotes
 docker run --rm \
     --volume ~/.config/rclone:/config/rclone \
-    --volume ~/data:/data:shared \
     --user $(id -u):$(id -g) \
     rclone/rclone \
     listremotes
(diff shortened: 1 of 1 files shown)


albertony:
Maybe the original author assumed there was already a rclone.conf prepared on the host rclone config dir? But your addition makes sense to me! I'm not a heavy docker user though... and speaking of... is the `--volume ~/data:/data:shared \` really needed in your command, as well as the existing "listremotes" command below? If not, maybe you can delete them from the examples to make it a tiny bit cleaner...

mitthu:
Sure. I removed `--volume ~/data:/data:shared \` since it's not really needed during "config".

I personally would keep the "listremotes" as a way to verify that the remote was indeed added.

albertony:
> I personally would keep the "listremotes" as a way to verify that the remote was indeed added.

Actually I meant just removing the `--volume ~/data:/data:shared \` from that command as well, if not needed? Sorry for being unclear. And it's unrelated to your changes, my suggestion was to sneak in an additional improvement while you're at it. 🙈

mitthu:
Ah okay. Updated it.

albertony (approved):

albertony:
Great, thank you very much. This now looks good to me. Lets wait some time to see if anyone with more experience with the docker functionality have something to say, if not I'll just merge it in a couple of days.

PR #5401, opened 2021-06-10
Improve MkdirAll on Windows

albertony:
#### What is the purpose of this change?

Improve os.MkdirAll on Windows, to avoid it trying to create `\\?` as a directory for a path where the drive letter does not exist. E.g. `\\?\A:\Temp`. Calling os.MkdirAll on such a path results in error message `mkdir \\?: The filename, directory name, or volume label syntax is incorrect.` The improved version will instead report it as ~`mkdir \\?\A:\Temp: The system cannot find the path specified.`~ `mkdir \\?\A:\: The system cannot find the path specified.`

~Edit:~
~Added additional commit with refactoring: Renaming package file (lib/file) to osutil (lib/osutil). Reasons for doing this:~
~- It already did not only contain *file* related functionality, but e.g. functions for driveletter and uncpath handling were also stuffed in there. And now the new MkdirAll does not necessarily have anything to do with *files* either (but directories).~
~- All of the functionality in the old file package seems to be either replacing (fixing) corresponding functions in built-in package `os`, or closely related.~
  ~- Existing functions `file.Open` and `file.Create` had quite intuitive names, but they replaces built-in `os.Open` and `os.Create`, so I think renaming to `osutil.Open` and `osutil.Create` can be defended..~
~- Many names that would be natural to use, `path` and similar, can easily be ambiguous with rclone's main path handling, working with backend/remote paths etc. While functionality in this package is related to paths on the host, so i think the "os" name fits well.~
    
~TODO:~
- ~Naming things...~
    - ~The current draft has simply added MkdirAll function to lib/file, but `file.MkdirAll` is probably not the most natural name (I mean: file..dir..).~
    - ~Though lib/file already contains functions for driveletter, which is also not related to file, and uncpath handling - which may or may not be file-related.~
    - ~So either: a) just swallow the 💊 and leave it as is b) rename lib/file to something else (lib/osutil? not to get in conflict with built-in package os) c) move mkdirall, driveletter, and possibly unc out from lib/file to some other lib package (lib/:thinking:)~

#### Was the change discussed in an issue or in the forum before?

https://github.com/rclone/rclone/issues/5400

#### Checklist

- [X] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [X] I have added tests for all changes in this PR if appropriate.
- [ ] ~I have added documentation for the changes if appropriate.~
[...]

PR #6420, opened 2022-09-02
Improve MkdirAll on Windows for non-existent network host

albertony:

When copying to network path `\\Server\Share`, and the Server does not exist, then instead of reporting the error as

`mkdir \\?\UNC: The filename, directory name, or volume label syntax is incorrect.`

... which is not so informative, and a little confusing. Instead, report it as:

`mkdir \\?\UNC\Server: The specified path is invalid.`

This builds on the previously  custom variant of os.MkdirAll, introduced with #5401. Rclone transforms `\\Server\Share` into extended-length format `\\?\UNC\Server\Share`. When Server does not exist, then with standard os.MkdirAll it would fail with `mkdir \\?: The filename, directory name, or volume label syntax is incorrect.`. The mentioned PR changed this to `mkdir \\?\UNC: The filename, directory name, or volume label syntax is incorrect.`, which is still not very informative (reason is the code relies on `filename.VolumeName`, and it is not aware of extended-length network paths and therefore returns `\\?\UNC` - which is *not* a volume name!). This PR adds a small extra to improve this. 

I'm not so much worried about the error description, "The specified path is invalid" or "The filename, directory name, or volume label syntax is incorrect", I'm more interested in getting the path reported being something the user can relate to. And then the error description more or less is what it is, without introducing too much additional logic compared to os.MkdirAll.

https://forum.rclone.org/t/use-rclone-to-copy-file-to-list-of-remote-computers-locations/31998

#### Checklist

### DIFF
--- a/lib/file/mkdir_windows.go
+++ b/lib/file/mkdir_windows.go
@@ -11,8 +11,9 @@ import (
 
 // MkdirAll creates a directory named path, along with any necessary parents.
 //
-// Improves os.MkdirAll by avoiding trying to create a folder \\? when the
-// volume of a given extended length path does not exist.
+// Improves os.MkdirAll by avoiding trying to create a folder `\\?` when the
+// volume of a given extended length path does not exist, and `\\?\UNC` when
+// a network host name does not exist.
 //
 // Based on source code from golang's os.MkdirAll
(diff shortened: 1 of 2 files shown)


ncw (approved):
This looks like a nice fix - that error message is most confusing for users!

LGTM :-)

PR #5889, opened 2021-12-30
docs: align menu items when icons have different sizes

albertony:
#### What is the purpose of this change?

The "Links" menu looks a little bit "messy", due to icons having different sizes. This adds a custom css to align the icons and the following text labels. Image below shows original first, then the new version:

Capture

#### Was the change discussed in an issue or in the forum before?

Similar to #5701, which aligned the "Storage systems" drop down. I chose to use width 18px here for the menu, instead of 20px as with dropdown. Concluded that it looked a tiny bit better with the smaller font size, but anything from 16px and up looks fine to me, and since texts are short saving of horizontal space is not something that needs to be considered, so exact value is not important (to me).

#### Checklist

- [X] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ ] I have added tests for all changes in this PR if appropriate.
- [X] I have added documentation for the changes if appropriate.
- [X] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [X] I'm done, this Pull Request is ready for review :-)

PR #6421, opened 2022-09-04
Docs: Better alignment of icons

albertony:

Better alignment of icons of different sizes

This reverts the changes from PR https://github.com/rclone/rclone/pull/5889 and https://github.com/rclone/rclone/pull/5701 and implements an alternative fix which is more "native" Font Awesome solution:

> Font Awesome icons are designed on grid and share a consistent height. But they vary in
width depending on how wide or narrow each symbol is. If you prefer to work with icons
that have a consistent width, adding fa-fw will render each icon using the same width.

(https://fontawesome.com/docs/web/style/fixed-width)

The original solution had wrong css, as a result of misunderstanding the precedence of `element,element` and `element element` selectors, leading to an issue with icons in backend doc headers.

Before:

memory-before

After:

memory-after

As a bonus, it also gives slightly better results for the menu/dropdown icons which was the focus of the original PRs.

Before:

before

After:

after

#### Checklist

### DIFF
--- a/docs/layouts/chrome/menu.html
+++ b/docs/layouts/chrome/menu.html
@@ -17,9 +17,9 @@
   <div class="card-body">
     <p class="menu">
       <!-- Non tracking sharing links from: https://sharingbuttons.io/ -->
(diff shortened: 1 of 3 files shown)


ncw:
On first glance this seems like a lot of repetition instead of a few lines of css?

If only you could include other css classes into a css class...

albertony:
I see what you mean, but... :) I think there are already repetition with all the `fa fa-book` etc. And I think it is just as clean with the `fa-fw`, letting it be handled by Font Awesome css, than the css width, alignment setting - which although more centralized, feels a bit like a hack (although I'm not a css guru, so it all feels a bit hacky at times... :)

ncw (approved):
OK I agree!

Let's merge :-)

PR #6427, opened 2022-09-08
Update install.md

opexxx:
cholateley package manager

#### Checklist

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -188,6 +188,15 @@ ls ~/data/mount
 kill %1
 ```
 
+## Install on Windows via Chocolateley Packet Manager
+
+Make sure you have [Choco](https://chocolatey.org/) installed
+```
+choco search rclone
+choco install rclone
+```
+This will install rclone on your windows machine
+
 ## Install from source
 
 Make sure you have git and [Go](https://golang.org/) installed.


albertony:
I guess the question is if we want to endorse a third party source like that. The risk is that it will not be updated, or very slowly. There are many cases in the forum of users running old versions installed from linux repos. Also there is a wiki page for Chocolatey install (using `choco upgrade rclone` by the way): https://github.com/rclone/rclone/wiki/Windows-Installation-with-Chocolatey. 🤔

ncw:
> I guess the question is if we want to endorse a third party source like that.

Here is the chocolatey page for rclone

https://community.chocolatey.org/packages/rclone

You can see that it isn't any rclone project members maintaining the package.

Looking at the release history https://community.chocolatey.org/packages/rclone#versionhistory and comparing it to the changelog the releases are either on the day of the rclone release or the day after, so the maintainers are doing a very good job.

In general we haven't pointed to 3rd party installers, but we do point to brew for macOS. `brew` always seems to have the latest rclone.

Maybe we should allow this, since `choco install` is a godsend on Windows (we use it in the Windows build machine to install WinFSP for example) but put a note against all the 3rd party installers that they aren't maintained by the rclone developers and may be out of date?

Actually we could put a chocolatey and a brew badge in very easily in the instructions which would remove all doubt.

[Chocolatey package](https://repology.org/project/rclone/versions)

and

[Homebrew package](https://repology.org/project/rclone/versions)

> The risk is that it will not be updated, or very slowly. There are many cases in the forum of users running old versions installed from linux repos.

Linux repos are the worst (except for Arch which is always bang up to date)!

I did investigate a tool to show the repo status from https://repology.org/project/rclone/badges eg this (which is probably too big but can be tweaked)

[Packaging status](https://repology.org/project/rclone/versions)

albertony:
> Maybe we should allow this

Yes, maybe we should. But ensure we mark this, and the brew, as third party in the docs, with a notice to check that the version is up to date, or something. They are up to date now, but might not be in future.

> I did investigate a tool to show the repo status from https://repology.org/project/rclone/badges eg this (which is probably too big but can be tweaked)

That was cool - and it even includes the Chocolatey and Homebrew, not just classic linux distros/repos! A bit big and flashy, for sure, and glaring number of reds, but I think we should find a place for it!

ncw:
OK I'll merge this as it is, and then send an update to install.md with some extra bits in.

Thank you @opexxx for your contribution and for kicking off this discussion.

PR #6442, opened 2022-09-14
sftp: check for directory existence if mkdir fails

Lesmiscore:

This checks directory existence if mkdir operation fails.
Race condition happens very often when an union remote points to the same (physical) directory, and directory creation is needed.
As servers find the directory that others are created, you get "Failure" errors and unfortunate files will be skipped.

Although it looks it's patched for the FTP backend, but SFTP is not.

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -1171,6 +1171,11 @@ func (f *Fs) mkdir(ctx context.Context, dirPath string) error {
 	err = c.sftpClient.Mkdir(dirPath)
 	f.putSftpConnection(&c, err)
 	if err != nil {
+		ok, _ := f.dirExists(ctx, dirPath)
+		if ok {
+			fs.LogPrintf(fs.LogLevelWarning, f, "directory %q exists after Mkdir is attempted", dirPath)
+			return nil
+		}
 		return fmt.Errorf("mkdir %q failed: %w", dirPath, err)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


backend/sftp/sftp.go
```diff
@@ -1171,6 +1171,11 @@ func (f *Fs) mkdir(ctx context.Context, dirPath string) error {
 	err = c.sftpClient.Mkdir(dirPath)
 	f.putSftpConnection(&c, err)
 	if err != nil {
+		ok, _ := f.dirExists(ctx, dirPath)
+		if ok {
+			fs.LogPrintf(fs.LogLevelWarning, f, "directory %q exists after Mkdir is attempted", dirPath)
```
ncw, line 1176: Can you use one of the established log levels here fs.Debugf/Infof/Logf/Errorf - we only document use of those log levels.

I think `fs.Debugf` is fine here - we've created the directory and it exists.

backend/sftp/sftp.go
```diff
@@ -1171,6 +1171,11 @@ func (f *Fs) mkdir(ctx context.Context, dirPath string) error {
 	err = c.sftpClient.Mkdir(dirPath)
 	f.putSftpConnection(&c, err)
 	if err != nil {
+		ok, _ := f.dirExists(ctx, dirPath)
```
ncw, line 1174: In theory you should be able to use `os.IsExist(err)` and avoid calling `dirExists` here - I don't know how well that works in practice though.

ncw (commented):
Nice fix - thank you.

See inline for some comments.

backend/sftp/sftp.go
Lesmiscore: I didn't test but `dirExists` suggests that it should - I can test in the actual transfer after current batch finishes though

ncw (approved):
Looks great - thank you.

I'll merge now.

PR #6445, opened 2022-09-15
docs: update install docs to make more consistent

ncw:
This also adds repology badges where appropriate to show versions in external repositories.

Can you take a look @albertony and see what you think?

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -14,7 +14,7 @@ Rclone is a Go program and comes as a single binary file.
   * Run `rclone config` to setup. See [rclone config docs](/docs/) for more details.
   * Optionally configure [automatic execution](#autostart).
 
-See below for some expanded Linux / macOS instructions.
+See below for some expanded Linux / macOS / Windows instructions.
 
 See the [usage](/docs/) docs for how to use rclone, or
 run `rclone -h`.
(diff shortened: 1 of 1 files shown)


albertony (approved):
I think this is good. Very nice improvement to this part of the docs. 🚢 it.

In general I try to avoid empty headers, or what to call it. E.g. I try to add some body text below a L1 header before first L2 header. But its not always easy to come up with something that fits in. Also, here we have L1 headers "Windows installation" and "Linux installation" immediately followed by a single L2 header "Precompiled binary", but not additional L2 headers, then it instinctively feels like the L2 header is unnecessary. But I see the common structure you've created, and I like that, and perhaps there will be more details/variants added later that makes the structure even more relevant. Don't know if my ramblings made any sense to you, I blame it on this cold I've got.. 🤧 Anyway, lets go with this now. We can think about further iterations later (including connecting https://rclone.org/downloads/ and https://rclone.org/install/ pages better).

ncw:
> I think this is good. Very nice improvement to this part of the docs. ship it.

Thanks!
 
> In general I try to avoid empty headers, or what to call it. E.g. I try to add some body text below a L1 header before first L2 header. But its not always easy to come up with something that fits in. Also, here we have L1 headers "Windows installation" and "Linux installation" immediately followed by a single L2 header "Precompiled binary", but not additional L2 headers, then it instinctively feels like the L2 header is unnecessary. But I see the common structure you've created, and I like that, and perhaps there will be more details/variants added later that makes the structure even more relevant. Don't know if my ramblings made any sense to you, I blame it on this cold I've got.. sneezing_face

I do see what you mean. 

> Anyway, lets go with this now. We can think about further iterations later (including connecting https://rclone.org/downloads/ and https://rclone.org/install/ pages better).

I'll merge this now and we can do another iteration!

PR #6446, opened 2022-09-16
Feature/sse c base64

### LINKED ISSUE #6400: You cannot use a full binary value for sse_customer_key

https://forum.rclone.org/t/cant-use-sse-customer-key-with-non-printable-charactersin-config-file/32667

tried with v1.56.2 and v1.50.2

There does not seem to be any way to put a non-printable sse_customer_key in the config file -- also, that key should probably be obscured anyway.

The easiest method that occurs to me is to just allow it to be obscured; since you can obscure a non-printable key and it results in printable characters that should do it. The other option I see would be to detect in some way if it is a base64 encoded value, e.g. "this is longer than 32 bytes and it's an AES256 key, so it must be b64 encoded" or supporting `base64:<b64 encoded customer key>`

Seems like it should be a simple enough change.

#### How to use GitHub

taxilian:

Fixes #6400 -- make it possible to provide a binary sse_customer_key for s3 SSE-C usage

Yes, see #6400 and https://forum.rclone.org/t/cant-use-sse-customer-key-with-non-printable-charactersin-config-file/32667

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1536,7 +1536,16 @@ isn't set then "acl" is used instead.`,
 			}},
 		}, {
 			Name:     "sse_customer_key",
-			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data.",
+			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data. Alternatively you can provide sse_customer_key_base64.",
+			Provider: "AWS,Ceph,ChinaMobile,Minio",
+			Advanced: true,
+			Examples: []fs.OptionExample{{
+				Value: "",
+				Help:  "None",
+			}},
+		}, {
+			Name:     "sse_customer_key_base64",
+			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data in base64 format. Alternatively you can provide sse_customer_key.",
 			Provider: "AWS,Ceph,ChinaMobile,Minio",
 			Advanced: true,
 			Examples: []fs.OptionExample{{
@@ -2111,6 +2120,7 @@ type Options struct {
 	SSEKMSKeyID           string               `config:"sse_kms_key_id"`
 	SSECustomerAlgorithm  string               `config:"sse_customer_algorithm"`
 	SSECustomerKey        string               `config:"sse_customer_key"`
+	SSECustomerKeyB64     string               `config:"sse_customer_key_base64"`
 	SSECustomerKeyMD5     string               `config:"sse_customer_key_md5"`
 	StorageClass          string               `config:"storage_class"`
 	UploadCutoff          fs.SizeSuffix        `config:"upload_cutoff"`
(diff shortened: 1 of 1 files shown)


backend/s3/s3.go
```diff
@@ -2644,6 +2654,17 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if opt.BucketACL == "" {
 		opt.BucketACL = opt.ACL
 	}
+	if opt.SSECustomerKeyB64 != "" && opt.SSECustomerKey != "" {
+		return nil, errors.New("s3: can't use sse_customer_key and sse_customer_key_base64 at the same time")
+	}
+	if opt.SSECustomerKeyB64 != "" && opt.SSECustomerKey == "" {
```
ncw, line 2660: The `&& opt.SSECustomerKey == ""` is redundant due to the if above I think.

backend/s3/s3.go
```diff
@@ -1536,7 +1536,16 @@ isn't set then "acl" is used instead.`,
 			}},
 		}, {
 			Name:     "sse_customer_key",
-			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data.",
+			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data. Alternatively you can provide sse_customer_key_base64.",
```
ncw, line 1539: The `Help:` is used to make to make the docs and also the help for the flag.

Only the first line is used for the flag help text (try `rclone help flags sse`)

The other lines should be separated by a blank line and wrapped to 80 chars.

So I think this needs to be something like this. Note that we usually
refer to options by their flag name in the docs.

```go
			Help:     `Secret encryption key for SSE-C

If using SSE-C you must provide the secret encryption key used to
encrypt/decrypt your data. Alternatively you can provide
--sse-customer-key-base64.`,

```

backend/s3/s3.go
```diff
@@ -1536,7 +1536,16 @@ isn't set then "acl" is used instead.`,
...
+			Provider: "AWS,Ceph,ChinaMobile,Minio",
+			Advanced: true,
+			Examples: []fs.OptionExample{{
+				Value: "",
+				Help:  "None",
+			}},
+		}, {
+			Name:     "sse_customer_key_base64",
+			Help:     "If using SSE-C you must provide the secret encryption key used to encrypt/decrypt your data in base64 format. Alternatively you can provide sse_customer_key.",
```
ncw, line 1548: This help needs to change too.

backend/s3/s3.go
```diff
@@ -2111,6 +2120,7 @@ type Options struct {
 	SSEKMSKeyID           string               `config:"sse_kms_key_id"`
 	SSECustomerAlgorithm  string               `config:"sse_customer_algorithm"`
 	SSECustomerKey        string               `config:"sse_customer_key"`
+	SSECustomerKeyB64     string               `config:"sse_customer_key_base64"`
```
ncw, line 2123: I prefer these to match so `SSECustomerKeyBase64` even if it is a bit of a mouthful!

ncw (commented):
Looking very good :-)

I put some comments about help text (which was wrong before your change!) and naming inline.

Thank you :-)

backend/s3/s3.go
taxilian: Changed -- I also made it an `else if`, which is technically irrelevant, but I feel like it makes the code more clear that those two are related and the following one is not (or not directly, anyway -- it's about making sure we have the md5, not about the key field itsself)

backend/s3/s3.go
taxilian: done

taxilian:
Rewrote the commit messages to match your requested format (I think, I have a short attention span sometimes) =]

ncw (approved):
That looks great, thanks.

I'll merge that now :-)

taxilian:
Not trying to jog your elbow, just wondering -- when do you expect to have a release with this in it? It looks like it just missed the last one?

ncw:
This week is the plan!

PR #6449, opened 2022-09-17
Stricter check for remote names with leading/trailing space

albertony:

This started as a documentation task; to add a statement that remote names may not *end* with space. This triggered more investigations, to verify that this is really always the case, and then one thing lead to another... First fixed/simplified the remote/config name regex by removing superfluous elements, unnecessary capture groups etc. Then extended it to explicitly match only when no leading/trailing space, so that an error is reported in the same way as when illegal characters are used.

Normally the remote names from config are trimmed (`strings.TrimSpace`) early on, and will never contain leading/trailing space when matched against the changed regex. This is the case when using interactive command line (config ui) and when the remote names are read/parsed from the config file. I.e. there will e no change to these uses. One place I found it will make a difference is rc, as illustrated below - which is using rclone without this PR:

```
rclone rcd --rc-no-auth
rclone rc config/create "name=   space   " type=local parameters="{}"
	{}
rclone rc config/listremotes
	{
		"remotes": [
			"   space   "
		]
	}
rclone rc config/get "name=   space   "
	{
		"type": "local"
	}
rclone listremotes
	space:
rclone lsd space:
	-1 2022-09-16 16:59:13        -1 .git
	-1 2022-01-26 14:50:08        -1 .github
	-1 2022-08-17 07:39:15        -1 backend
	-1 2022-09-14 13:29:43        -1 bin
	-1 2022-08-30 13:16:45        -1 cmd

```

Maybe not a very realistic example? And one could say that rcd(?) instead could also do `strings.TrimSpace` on the remote name. But that was not really my point. My point with this PR was primarily to improve the code and documentation according to the fact that rclone does not/should not/have never (?) supported remote names with trailing (or leading) space.

Note:
- The changes to a single line of regex (more or less) are currently spread across several commits, to carefully document each individual change leading to the end result.

Not really. But the following discussion is what made me look into the handling of space characters in remote names:

https://forum.rclone.org/t/sync-to-onedrive-personal-lands-file-in-localfilesystem-but-not-in-onedrive/32956

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -3431,7 +3431,7 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
 		if err != nil {
 			return nil, err
 		}
-		re := regexp.MustCompile(`[^\w_. -]+`)
+		re := regexp.MustCompile(`[^\w\p{L}\p{N}. -]+`)
 		if _, ok := opt["config"]; ok {
 			lines := []string{}
 			upstreams := []string{}
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -338,9 +338,17 @@ Will get their own names
 ### Valid remote names
 
 Remote names are case sensitive, and must adhere to the following rules:
- - May only contain `0`-`9`, `A`-`Z`, `a`-`z`, `_`, `-`, `.` and space.
+ - May contain number, letter, `_`, `-`, `.` and space.
  - May not start with `-` or space.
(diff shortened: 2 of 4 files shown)


ncw (commented):
These changes look very nice - I like the methodical approach here.

Do you want to fix #6618 in this series by adding `\p{L}\p{N}` alongside the `\w` (yes there may be some overlap but at least that way we know it is 100% backwards compatible) and add some tests?

PR #6455, opened 2022-09-19
vfs: fix directory cache serving stale data

### LINKED ISSUE #6335: VFS doesn't update mtime for directories
I'm using `serve webdav` to serve a local directory over WebDAV. This works fine except that the VFS layer doesn't update the mtime (ModTime) of directories even after `dir-cache-time` has expired. Debugging revealed that the problem is in `dir.go:635` where rclone reuses `node` (which has the old modTime), `item` (the new value with the correct modTime) is just discarded.

I'm not familiar with the design principles of the rclone VFS layer so I'm not sure what the right fix here is.

Self-built from revision 2f461f13e3efe603bdd8fb4057f3e8242a59956f

Linux x86_64

lorenz:
The VFS directory cache layer didn't update directory properties like the last modified time at all, as the new objects were not used. The same problem is already solved for files, implement the same solution for directories.

Fixes #6335

#### Checklist

### DIFF
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -760,6 +760,7 @@ func (d *Dir) _readDirFromEntries(entries fs.DirEntries, dirTree dirtree.DirTree
 			dir := node.(*Dir)
 			dir.mu.Lock()
 			dir.modTime = item.ModTime(context.TODO())
+			dir.entry = item
 			if dirTree != nil {
 				err = dir._readDirFromDirTree(dirTree, when)
 				if err != nil {
--- a/vfs/dir_test.go
+++ b/vfs/dir_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"runtime"
 	"sort"
 	"testing"
 	"time"
@@ -654,4 +655,35 @@ func TestDirFileOpen(t *testing.T) {
 	fi, err = vfs.Stat("dir/sub/file2")
 	require.NoError(t, err)
 	assert.Equal(t, int64(12), fi.Size())
+}
+
+func TestDirEntryModTimeInvalidation(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("dirent modtime is unreliable on Windows filesystems")
+	}
+	r, vfs := newTestVFS(t)
+
+	// Needs to be less than 2x the wait time below, othewrwise the entry
+	// gets cleared out before it had a chance to be updated.
+	vfs.Opt.DirCacheTime = fs.Duration(50 * time.Millisecond)
+
+	r.WriteObject(context.Background(), "dir/file1", "file1 contents", t1)
+
+	node, err := vfs.Stat("dir")
+	require.NoError(t, err)
+	modTime1 := node.(*Dir).DirEntry().ModTime(context.Background())
+
+	// Wait some time, then write another file which must update ModTime of
+	// the directory.
+	time.Sleep(75 * time.Millisecond)
+	r.WriteObject(context.Background(), "dir/file2", "file2 contents", t2)
+
+	node2, err := vfs.Stat("dir")
+	require.NoError(t, err)
+	modTime2 := node2.(*Dir).DirEntry().ModTime(context.Background())
+
+	// ModTime of directory must be different after second file was written.
+	if modTime1.Equal(modTime2) {
+		t.Error("ModTime not invalidated")
+	}
 }


ncw (commented):
That looks like a nice fix :-)

How did you come across the problem?

Can you make a test for it?

lorenz:
I had WebDAV client software which cached directory contents if the modified time of a directory wasn't updated. This caused new data to never show up. Re Testing: I have pretty much no experience with the codebase. I'll happily write a test for it with a bit of guidance. Manual testing was done by serving a local directory with WebDAV, setting dir-cache-time=10s and touching a new file in the directory, then waiting for >10s and refreshing to see if the modified time changed.

lorenz:
@ncw Anything I can do to get this merged? It's been sitting for 2.5 years and people are still running into the issue.

vfs/dir.go
```diff
@@ -916,6 +918,17 @@ func (d *Dir) cachedNode(relativePath string) Node {
 	return node
 }
 
+// Update the object but don't update the directory cache - for use by
```
ncw, line 921: I think this comment is wrong. It is certainly confusing me!

I think you mean something like this?

```go
// Update the directory in the directory cache with a new fs.Directory.
// For use by the directory cache.
// Does not update the VFS cache.
```

ncw (commented):
@lorenz sorry for the loooong delay!

To merge first it needs rebasing on master.

Then ideally we'd have a test.

> Manual testing was done by serving a local directory with WebDAV, setting dir-cache-time=10s and touching a new file in the directory, then waiting for >10s and refreshing to see if the modified time changed

So we need to automate that. We don't need to serve it with WebDAV though - the test can just look at the VFS directly.

If you take a look in https://github.com/rclone/rclone/blob/master/vfs/dir_test.go you'll see lots of test which you could copy and adapt.

I think you can probably call `ForgetAll()` rather than waiting for the vfs cache to expire also.

Let me know if you need more help.

lorenz (commented):
The original approach turned out to perform far too much work. The only problem is the `entry` field which doesn't get updated. Also this is now harder to reproduce as these stale entries get cleaned out after twice the invalidation period. But stale data can still persist indefinitely if it is frequently accessed, so this patch is still necessary.

Fixed all of this and added a test which fails without the fix.

ncw:
Looks like the tests are failing under Windows.

The windows timers have lower granularity than other OSes which often causes stuff like this.

Can you use `ForgetAll()` rather than waiting for the vfs cache entry to time out?

ncw:
> ForgetAll drops the direntry entirely. On next access it gets recreated, thus the bug doesn't happen. The behavior this is relying on (that the parent directory modtime gets updated when files are created/deleted) might also not be applicable to Windows.

I think this should happen.

Chat GPT came up with this table for whether a parent directory modtime gets updated or not. I veriefied the Windows line with my Windows VM

OS / FS | File Created | File Deleted | File Modified
-- | -- | -- | --
Linux (Ext4, XFS, Btrfs, etc.) | ✅ | ✅ | ❌
macOS (APFS, HFS+) | ✅ | ✅ | ❌
Windows (NTFS, FAT32, exFAT) | ✅ | ✅ | ❌
BSD (UFS, ZFS) | ✅ | ✅ | ❌
NFS / SMB | ⚠️ | ⚠️ | ❌

- ✅ = Updates modtime
- ❌ = Does not update modtime
-⚠️ = May depend on caching or implementation

Maybe you could repeat the failing `t.Error("ModTime not invalidated")` test until it succeeds or 10 times?

Something like
	

```go
        var ok bool
	for tries := 1; tries <= 10; tries++ {
            // ModTime of directory must be different after second file was written.
            ok = !modTime1.Equal(modTime2)
            if ok {
                break
            }
            time.Sleep(100*time.Millisecond)
	}
	if  !ok {
		t.Error("ModTime not invalidated")
	}
```

lorenz:
Spun up a Windows 26001 (24H2) VM, this cannot be made reliable on Windows. The filesystem API (directly using os.Stat on the backing files) behaves erratically. Even waiting multiple seconds between writes sometimes ends up with no update to the last modified timestamp on the directory. Either we force the timestamp to be "correct" via a Windows-specific API or skip the entire test on Windows.

ncw:
Ok let's skip the test on windows. I've noticed before that the VM filesystem is a bit erratic!

lorenz:
Test is now getting skipped on Windows

ncw (approved):
Great - thank you :-)

I'll merge now.

ncw:
Unfortunately this PR has generated a lot of failures in the [daily integration tests](https://pub.rclone.org/integration-tests/current/)

You'll see if you look at the link above that the new TestDirEntryModTimeInvalidation is failing.

Have you got time to take a look at this @lorenz ?

Thanks

darthShadow:
I think this is failing because it's not skipping the test for remotes that don't support the `DirModTimeUpdatesOnWrite` feature. All the failing backends in the test suite seem to be those which are missing that feature.

ncw:
I have attempted to fix the integration test failures in 0010090d0517976248894bbd48a9bb1ac5bc0182 :crossed_fingers:

Issue #3382, opened 2019-07-29
serve s3: let rclone act as an S3 compatible server serving any backend


jkaberg:
### What is your current rclone version (output from `rclone version`)?
```
rclone v1.47.0
- os/arch: linux/amd64
- go version: go1.12.4
```

#### What problem are you are trying to solve?
Serve S3 protocol for backends that don't support it

#### How do you think rclone should be changed to solve that?
Let rclone via the serve parameter act as an "s3 proxy". Currently rclone supports many backends which lacks this, while many other tools commonly implements S3 as a backend. Something like https://github.com/minio/minio coupled with rclone backend's would be very cool. Yes I could do it with fuse+minio, however that would add another tool and complexity.

ncw:
I really like this idea.  The obvious thing would be to take the internals of minio and use them as a library, however last time I looked at its code it wasn't designed for that.

It might be worth another look...

jkaberg:
Thanks for considering @ncw, afaik there aren't many golang compatible implementations to the [official API](https://docs.aws.amazon.com/AmazonS3/latest/API/Welcome.html) that I know

- https://github.com/minio/minio (100% compatible?)
- https://github.com/johannesboyne/gofakes3 (missing features)

There seems to be an minio issue tracking an more library/package approach over [here](https://github.com/minio/minio/issues/4812)

ncw:
Gofakes looks quite promising, eg the [backend definiton](https://godoc.org/github.com/johannesboyne/gofakes3#Backend).

It would be possible with a bit of work to make an rclone backend to that spec.

Fancy having a go?

Rick7C2:
I would really like to see this feature! Unfortunately I don't know go very well so will be unable to help but I hope someone who can add this.

I would even be willing to start a bounty if enough people will add to the pool just to get this feature added to rclone.

gary-kim:
It looks like the [referenced issue regarding minio taking a more library approach](https://github.com/minio/minio/issues/4812) seems to be closed saying that enough sub-systems have been moved to their own packages.

pmjohann:
> Gofakes looks quite promising, eg the [backend definiton](https://godoc.org/github.com/johannesboyne/gofakes3#Backend).
> 
> It would be possible with a bit of work to make an rclone backend to that spec.
> 
> Fancy having a go?

+1 on this, as S3 being a (or THE) popular object storage protocol, I think this would benefit many.
[...]

PR #6461, opened 2022-09-22
serve: add s3 server

Mikubill:
Implements s3 server for rclone - https://github.com/rclone/rclone/issues/3382

Built with gofakes3 and serve/httplib. Some integration tests have passed so far, but still more work to do:
- [x] auth support - requires auth middleware
- [x] hash - might not work
- [x] pager - needs fix
- [x] more features - any suggestions?
- [x] docs and examples

#### Problems

* To comply with s3 (no empty folder), empty directories will be removed while deleting objects, which may cause unexpected/unwanted operations.
* `Listobjects` needs to list all files to apply the prefix filter. In the implementation I used `walk.ListR(ctx,fs)`, but this seems to be an expensive operation, especially when there are thousands of files in the directory
* The failures in integration tests are mainly caused by metadata. It seems that different backends implement metadata r/w func separately... Is there any generic way to handle it and work with vfs?

#### Checklist


Mikubill:
```
    s3_test.go:109: 
        ----------
        === RUN   TestVersionLess
        --- PASS: TestVersionLess (0.00s)
        === RUN   TestMergeDeleteMarkers
        --- PASS: TestMergeDeleteMarkers (0.00s)
        === RUN   TestIntegration
            fstests.go:454: Using remote "s3test:"
        === RUN   TestIntegration/FsCheckWrap
            fstests.go:495: Not a wrapping Fs
        === RUN   TestIntegration/FsCommand
        === RUN   TestIntegration/FsRmdirNotFound
        === RUN   TestIntegration/FsString
        === RUN   TestIntegration/FsName
        === RUN   TestIntegration/FsRoot
        === RUN   TestIntegration/FsRmdirEmpty
        === RUN   TestIntegration/FsMkdir/FsMkdirRmdirSubdir
            fstest.go:251: Filtering empty directory "dir"
            fstest.go:251: Filtering empty directory "dir/subdir"
            fstest.go:251: Filtering empty directory "dir"
        === RUN   TestIntegration/FsMkdir/FsListEmpty
        === RUN   TestIntegration/FsMkdir/FsListDirEmpty
        === RUN   TestIntegration/FsMkdir/FsListRDirEmpty
        === RUN   TestIntegration/FsMkdir/FsListDirNotFound
        2022/09/22 05:15:36 ERROR : does not exist: error listing: directory not found
        === RUN   TestIntegration/FsMkdir/FsListRDirNotFound
        2022/09/22 05:15:36 ERROR : does not exist: error listing: directory not found
        === RUN   TestIntegration/FsMkdir/FsEncoding
        === RUN   TestIntegration/FsMkdir/FsEncoding/control_chars
            fstests.go:710: testing "␀␁␂␃␄␅␆␇␈␉␊␋␌␍␎␏␐␑␒␓␔␕␖␗␘␙␚␛␜␝␞␟␡"
        === RUN   TestIntegration/FsMkdir/FsEncoding/dot
            fstests.go:710: testing "．"
        === RUN   TestIntegration/FsMkdir/FsEncoding/dot_dot
            fstests.go:710: testing "．．"
        === RUN   TestIntegration/FsMkdir/FsEncoding/punctuation
            fstests.go:710: testing "!\"#$%&'()*+,-.／:;<=>?@[\\]^_`{|}~"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_space
            fstests.go:710: testing " leading space"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_tilde
            fstests.go:710: testing "~leading tilde"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_CR
            fstests.go:710: testing "␍leading CR"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_LF
            fstests.go:710: testing "␊leading LF"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_HT
            fstests.go:710: testing "␉leading HT"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_VT
            fstests.go:710: testing "␋leading VT"
        === RUN   TestIntegration/FsMkdir/FsEncoding/leading_dot
            fstests.go:710: testing ".leading dot"
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_space
            fstests.go:710: testing "trailing space "
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_CR
            fstests.go:710: testing "trailing CR␍"
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_LF
            fstests.go:710: testing "trailing LF␊"
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_HT
            fstests.go:710: testing "trailing HT␉"
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_VT
            fstests.go:710: testing "trailing VT␋"
        === RUN   TestIntegration/FsMkdir/FsEncoding/trailing_dot
            fstests.go:710: testing "trailing dot."
        === RUN   TestIntegration/FsMkdir/FsEncoding/invalid_UTF-8
            fstests.go:710: testing "invalid utf-8\xfe"
        === RUN   TestIntegration/FsMkdir/FsEncoding/URL_encoding
            fstests.go:710: testing "test%46.txt"
        === RUN   TestIntegration/FsMkdir/FsNewObjectNotFound
        === RUN   TestIntegration/FsMkdir/FsPutError
        === RUN   TestIntegration/FsMkdir/FsPutZeroLength
        === RUN   TestIntegration/FsMkdir/FsOpenWriterAt
            fstests.go:783: FS has no OpenWriterAt interface
        === RUN   TestIntegration/FsMkdir/FsChangeNotify
            fstests.go:819: FS has no ChangeNotify interface
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListDirFile2
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListRDirFile2
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListR
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListRSubdir
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListDirRoot
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListRDirRoot
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListSubdir
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListRSubdir#01
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListLevel2
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListRLevel2
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListFile1
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsNewObject
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsNewObjectCaseInsensitive
            fstests.go:1072: Not Case Insensitive
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsListFile1and2
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsNewObjectDir
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsPurge
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsCopy
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsMove
            fstests.go:1175: FS has no Mover interface
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsDirMove
            fstests.go:1246: FS has no DirMover interface
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsRmdirFull
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsPrecision
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectString
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectFs
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectRemote
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectHashes
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectModTime
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType
            fstests.go:1385: 
                        Error Trace:    /workspaces/rclone-2/backend/s3/fstests.go:1385
                        Error:          Not equal: 
                                        expected: "text/csv"
                                        actual  : "text/plain; charset=utf-8"
                                    
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -1 +1 @@
                                        -text/csv
                                        +text/plain; charset=utf-8
                        Test:           TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType
                        Messages:       can read and write mime types but failed
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectMetadata
            fstests.go:1432: 
                        Error Trace:    /workspaces/rclone-2/backend/s3/fstests.go:1432
                        Error:          Not equal: 
                                        expected: "potato"
                                        actual  : ""
                                    
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -1 +1 @@
                                        -potato
                                        +
                        Test:           TestIntegration/FsMkdir/FsPutFiles/ObjectMetadata
                        Messages:       can read and write metadata but failed on key "rclone-test"
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectMetadata/mtime
            fstests.go:1462: 
                        Error Trace:    /workspaces/rclone-2/backend/s3/fstests.go:1462
                        Error:          Not equal: 
                                        expected: "application/zip"
                                        actual  : "application/octet-stream"
                                    
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -1 +1 @@
                                        -application/zip
                                        +application/octet-stream
                        Test:           TestIntegration/FsMkdir/FsPutFiles/ObjectMetadata/mtime
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectSetModTime
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectSize
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectOpen
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectOpenSeek
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectOpenRange
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectPartialRead
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectUpdate
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectStorable
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsIsFile
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsIsFileNotFound
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot
            fstests.go:1612: Opening root remote "s3test:" path "rclone-test-hodayag6toceheh4ruwibec5" from "s3test:rclone-test-hodayag6toceheh4ruwibec5"
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/List
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListEntries
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/ListR
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/Put
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FromRoot/Put/Remove
        === RUN   TestIntegration/FsMkdir/FsPutFiles/PublicLink
            fstests.go:1753: skipping directory tests as not supported on this backend
        === RUN   TestIntegration/FsMkdir/FsPutFiles/SetTier
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectCheckWrap
            fstests.go:1808: Not a wrapping Fs
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectRemove
        === RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectAbout
            fstests.go:1842: FS does not support About
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsPutStream
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsPutStream/0
        2022/09/22 05:15:36 NOTICE: S3 bucket rclone-test-hodayag6toceheh4ruwibec5: Streaming uploads using chunk size 5Mi will have maximum file size of 48.828Gi
        === RUN   TestIntegration/FsMkdir/FsPutFiles/FsPutStream/100
        === RUN   TestIntegration/FsMkdir/FsPutFiles/Internal
        === RUN   TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata
            s3_internal_test.go:76: 
                        Error Trace:    /workspaces/rclone-2/backend/s3/s3_internal_test.go:76
                        Error:          Not equal: 
                                        expected: "en-US"
                                        actual  : ""
                                    
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -1 +1 @@
                                        -en-US
                                        +
                        Test:           TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata
                        Messages:       content-language
            s3_internal_test.go:76: 
                        Error Trace:    /workspaces/rclone-2/backend/s3/s3_internal_test.go:76
                        Error:          Not equal: 
                                        expected: "text/plain"
                                        actual  : "application/octet-stream"
                                    
                                        Diff:
                                        --- Expected
                                        +++ Actual
                                        @@ -1 +1 @@
                                        -text/plain
                                        +application/octet-stream
                        Test:           TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata
                        Messages:       content-type
        2022/09/22 05:15:36 Failed to parse time "": parsing time "" as "2006-01-02T15:04:05.999999999Z07:00": cannot parse "" as "2006"
        exit status 1
        FAIL    github.com/rclone/rclone/backend/s3     0.259s
        ----------
    s3_test.go:111: 
                Error Trace:    /workspaces/rclone-2/backend/s3/s3_test.go:111
                                                        /workspaces/rclone-2/backend/s3/s3_test.go:56
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestS3
                Messages:       Running s3 integration tests
--- FAIL: TestS3 (0.73s)
```

ncw:
This is great work :-)

> Built with gofakes3 and serve/httplib. Some integration tests have passed so far, but still more work to do:
> 
> * [ ]  auth support - requires auth middleware
> * [ ]  hash - might not work
> * [ ]  pager - needs fix
> * [ ]  more features - any suggestions?
> * [ ]  docs and examples

If you can get the s3 integration tests more or less passing then this will be a useful feature.

I would have thought you could do auth with the existing httplib auth?

> * To comply with s3 (no empty folder), empty directories will be removed while deleting objects, which may cause unexpected/unwanted operations.

Sounds OK to me. Note you only need to do this on backends with the Feature `features.CanHaveEmptyDirectories` which will save useless work if serving a bucket based backend.

> * `Listobjects` needs to list all files to apply the prefix filter. In the implementation I used `walk.ListR(ctx,fs)`, but this seems to be an expensive operation, especially when there are thousands of files in the directory

Hmm...

Assuming for a moment that we only support listing with a prefix which is a directory boundary (which is all rclone uses) then you can use the max depth controls on listing.

`ListR` will be listing all the things so if you are doing it for prefix searches then it will be expensive.

It would probably be more efficient to set up listing filters and just use `List` to emulate it properly. I think that would give a proper implementation at not too much computational cost. And I think if you assume that prefixes are always directory boundaries then you won't even need the filters.

> * The failures in integration tests are mainly caused by metadata. It seems that different backends implement metadata r/w func separately... Is there any generic way to handle it and work with vfs?

Metadata is an optional feature. If the backend doesn't support it that is fine.

You can probably disable the `Metadata` feature for the duration of the tests - add "Metadata" to this `[]string{}`

https://github.com/rclone/rclone/blob/7e547822d667e644576d250655fa3013a7159076/fs/config.go#L101

If this is serving a bucket based backend then it could potentially be more efficient but we don't expose the prefix listings that they have - maybe we should?

Mikubill:
>  s3 integration tests

Passed with the latest commit (on my end). Note that I used a small cache to store metadata, which is volatile and won't be transferred to any backend. Maybe change later to adapt with fs/vfs.

> features.CanHaveEmptyDirectories

added

> auth with the existing httplib

serve/httplib doesn't support per-request auth like AWS-Signature-V4 etc, so make a separate one for s3 serve. 

> only support listing with a prefix which is a directory boundary

I've modified the prefixParser to assume that slash is the default delimiter when not specified. That should make major requests goes to directory-based handler rather than fuzzy search. But I still wonder if we always assume the prefix is a directory boundary, will that cause recognition problems for some s3 clients...?

ncw:
This is looking good - well done :-)

Do you want me to review the code yet?

I should probably give it a go and try to break it!

Mikubill:
Almost done, I think you can start reviewing the code.

cmd/serve/s3/auth.go
```diff
@@ -0,0 +1,25 @@
+package s3
+
+import (
+	"net/http"
+
+	"github.com/rclone/rclone/cmd/serve/s3/signature"
+)
+
+func (p *Server) authMiddleware(handler http.Handler) http.Handler {
```
ncw, line 9: Would it be possible to make the auth proxy work with this server?

https://github.com/rclone/rclone/blob/master/cmd/serve/proxy/proxy.go

cmd/serve/s3/backend.go
```diff
@@ -0,0 +1,577 @@
...
+)
+
+type s3Backend struct {
+	opt  *Options
+	lock sync.Mutex
+	fs   *vfs.VFS
+}
+
+// newBackend creates a new SimpleBucketBackend.
+func newBackend(fs *vfs.VFS, opt *Options) gofakes3.Backend {
```
ncw, line 37: It should be possible to implement this without using the VFS, using the lower level fs commands only.

However the VFS provides caching etc which could be useful. It will introduce inefficiencies though of course... For example you can run ListR directly on bucket backed backends, but via the VFS it will list each directory one by one.

cmd/serve/s3/help.go
```diff
@@ -0,0 +1,45 @@
...
+
+S3 server supports Signature Version 4 authentication. Just
+ use ` + `--s3-auth accessKey1-secretKey1,accessKey2-secretKey2` + ` and
+ set  Authorization Header correctly in the request. (See 
+https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html)
+
+Please note that some clients may require HTTPS endpoints. 
+See [#SSL](#ssl-tls) for SSL configuration.
+
+Use ` + `--host-bucket` + ` if you want to use bucket name as a part of
```
ncw, line 16: Is this what we call path style vs virtual host style in the s3 backend? If so we should probably name it similarly

      --s3-force-path-style   If true use path style access if false use virtual hosted style (default true)

cmd/serve/s3/s3.go
```diff
@@ -0,0 +1,68 @@
...
+// Opt is options set by command line flags
+var Opt = DefaultOpt
+
+func init() {
+	flagSet := Command.Flags()
+	httpflags.AddFlags(flagSet)
+	vfsflags.AddFlags(flagSet)
+	flags.BoolVarP(flagSet, &Opt.hostBucketMode, "host-bucket", "", Opt.hostBucketMode, "Whether to use bucket name in hostname (such as mybucket.local)")
+	flags.StringVarP(flagSet, &Opt.hashName, "etag-hash", "", Opt.hashName, "Which hash to use for the ETag, or auto or blank for off")
+	flags.StringVarP(flagSet, &Opt.authPair, "s3-auth", "", Opt.authPair, "Set key pairs for authorization, split by comma. example: ak-sk,ak2-sk2")
```
ncw, line 35: It might be best to use `flags.StringArrayVarP` then you can repeat the flag

cmd/serve/s3/signature/signature-errors.go
```diff
@@ -0,0 +1,146 @@
+package signature
```
ncw, line 1: There seems to be a lot of boilerplate that should be supplied by gofakes in here?

ncw:
Do you want to merge this for 1.60 which I want to release in the next week or two? We could merge it with **beta** in the docs.

cmd/serve/s3/s3_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/config/configmap"
+	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/fstest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	endpoint = "localhost:8080"
```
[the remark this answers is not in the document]
Mikubill: Changed. Currently using new `lib/http` with `localhost:0`.

cmd/serve/s3/backend.go
```diff
@@ -0,0 +1,577 @@
...
+	"github.com/Mikubill/gofakes3"
+	"github.com/ncw/swift/v2"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/walk"
+	"github.com/rclone/rclone/lib/encoder"
+	"github.com/rclone/rclone/vfs"
+)
+
+var (
+	charEncoder    = encoder.EncodeCtl | encoder.EncodeLeftCrLfHtVt | encoder.EncodeRightCrLfHtVt | encoder.EncodeDel
```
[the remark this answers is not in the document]
Mikubill: Changed. Removed `charEncoder`.

cmd/serve/s3/auth.go
Mikubill: Changed. But I'm not sure if this will work with the s3 test.

cmd/serve/s3/backend.go
Mikubill: Considering the high request rate of S3, it should be better to use VFS on some backends with strict rate limits. I'd added some logic for the BucketBased backend to execute direct `fs.ListR`, hope this will work.

cmd/serve/s3/help.go
Mikubill: Changed. Currently using `--force-path-style`.

cmd/serve/s3/s3.go
Mikubill: Changed.

cmd/serve/s3/signature/signature-errors.go
Mikubill: Changed. Moved code to gofakes3

Mikubill:
Seems that some race issues occurred with `lib/http` (see https://github.com/Mikubill/rclone/actions/runs/3339975457/jobs/5529550702)

cmd/serve/s3/backend.go
stevefan1999-personal: Is it possible to also have a VFS passthrough mode or disable it entirely? I may need it because I will load balance the S3 server using multiple nodes (they restrict it by IP traffic) so having a VFS may cause cache invalidation problem.

blaineam:
> Seems that some race issues occurred with `lib/http` (see https://github.com/Mikubill/rclone/actions/runs/3339975457/jobs/5529550702)

Im not sure what caused that to error there, I forked your repo and ran the action myself and it originally errored out somewhere completely different and I tried to just update the repo with the 100+ commits on the main upstream branch and running the tests again in GitHub seems promising. 

The Github Actions don't seem to fail on the same test every time which makes me think it has nothing to do with the code changes but has everything to do with the Github Action job runners maybe deprioritizing these long running tasks for free GitHub accounts like my own. 

Also running make racequicktest locally your serve tests pass just fine on my MacBook and it just complains about osxfuse not being installed(I never installed it so it makes sense for the mount tests in macOS)

I would say this should be fine even ignoring the random failing GitHub action tests that don't even pertain to the changes in this pr.

In any case I am super excited to have this feature and plan to test it with some of my hobby projects when it makes it to beta or release. I will be able to test it with a few different cloud backends and can use it for large "buckets" with 4mil "objects" easily whenever this pr lands.

blaineam:
I can confirm that re-running the failed job with zero code changes it eventually passed (3rd try) https://github.com/blaineam/rclone/actions/runs/3589303971/jobs/6041966413
I think it is Github Actions being suspicious.

ncw:
@Mikubill do you want me to merge this for rclone 1.61? I think it is done isn't it +/- a few test failures.

blaineam:
I setup a docker container to test it with one of my backends and I could not get it to work with my mountain duck client at least not fully. It loaded the list of "buckets" but all the spaces were replaced with + signs and I couldn't open any of the buckets. I think it may have been a misconfiguration of my mountain duck because the s3 tests seem to be passing but I thought I should report this concern.

blaineam:
ok so I found some good news. the serve s3 is compatible with the web app file stash it just doesn't play well with transmit 5 or mountain duck so no clue why they have issues but it seems functional with certain clients which is super cool.

oniumy:
I just tested this branch with both a WebDAV and an SFTP backend. While requesting already existing files works perfectly, uploading new files with minio client produced the following issue for me:

All files stored in the backends get prepended with something like `10000;chunk-signature=d97affd26ae03e44fd492c9b2ba801e6a96c0f63c39d54e660b6d8de49be5c3e`, changing the actual content and therefore breaking the file. I checked the same files with minio client uploading directly to an s3 bucket, without running into this issue, so I would assume that it is an issue of my setup or this implementation.

I haven't changed any default settings and uploading the file with `rclone move file.txt sftp-remote:bucket/` works without this issue.

unixfox:
Any progress here? @Mikubill  Do you want to merge this PR into rclone?

Rick7C2:
> Any progress here? @Mikubill Do you want to merge this PR into rclone?

I'm eager for this feature as well but what about the issue here https://github.com/rclone/rclone/pull/6461#issuecomment-1347397215

blaineam:
so I decided to test writes on my own today using my docker container with a build of this pr inside of it. I was able to connect and see the list of buckets using cyberduck but it did not work when trying to access the contents of any bucket using normal s3 profiles. 
I decided to reconnect via the file stash web app and that still remained operational for reads. I then proceeded to try writing to confirm the bug listed above about it arbitrarily prepending a metadata string to the beginning of the file.
I uploaded a pdf via filestash to my docker container running rclone serve s3 with rather a default options not setting anything other than the s3 auth settings.
I downloaded the file both from s3 and the backing storage pcloud to compare how the file begins with a terminal head call and both of those downloads match the source file exactly for a pdf. I'm curious why minio-client was prepending that to the files.

I still would like to know why I cannot connect to the rclone serve s3 with Cyberduck and only the web app filestash seems to support it. I have tried the deprecated path style s3 profile and it says invalid signature. I have also tried using force-path-style false and it doesn't seem to fix anything even with proper DNS mappings and SSL configuration.

Although I can open and navigate most folders and files using filestash I am unable to open folders with a space in their key path which makes me think that whatever is telling the s3 client to replace spaces with '+' is causing issues with opening these folders.

I am able to open the list of buckets with cyberduck but cannot open the contents of any of the buckets and it errors out with various unknown errors such as missing '</link>' while listing directories or invalid signatures between server and client. I have switched to using the gofakes3 default region eu-central-1 and that seems to have removed those errors but it still will not support read or writes in cyberduck so filestash must be doing some magic that allows this to almost work.

What I find odd is the fact that the s3 rclone tests appear to be passing for this pr which makes me wonder if they are actually testing the right things.

overall it is not stable and has major bugs but file mutations are not one of them in my testing.

oniumy:
Thank you for testing writes @blaineam 

I did some more testing locally and figured out that it only affects uploads with v4 authorization. After tracing the issue through the code, I ended up in gofakes3, which had missing support for chunked multipart uploads as well as a bug in the chunked reader implementation. I created a pull request https://github.com/Mikubill/gofakes3/pull/1 with the required fixes.

With these fixes applied, everything works now in my limited testing with minio client (with and without v4 authorization).

blaineam:
> Thank you for testing writes @blaineam
> 
> I did some more testing locally and figured out that it only affects uploads with v4 authorization. After tracing the issue through the code, I ended up in gofakes3, which had missing support for chunked multipart uploads as well as a bug in the chunked reader implementation. I created a pull request [Mikubill/gofakes3#1](https://github.com/Mikubill/gofakes3/pull/1) with the required fixes.
> 
> With these fixes applied, everything works now in my limited testing with minio client (with and without v4 authorization).

that explains why my 1.4MB pdf doesnt have any issues because it likely fits within the size of a single chunk. Nice working patching the gofakes3 fork!

individual-it:
This looks amazing!
We are currently researching the possibility to have an S3 to WebDAV proxy and this does most of what we need. rclone can access a WebDAV storage and with this PR serve a S3 front-end. The only thing that is missing is the option to access different WebDAV accounts. Would it be somehow possible to use the S3 AccessKey and forward it to the WebDAV side as Bearer Token?

oniumy:
I would love to help getting this PR merged, but I'm not sure what issues need to get fixed to get there. All the issues I discovered in my testings are fixed and I'm successfully running this branch on a staging environment for a few weeks now without any issues.

dalbani:
Github says that there are merge conflicts to begin with.

dan3o3:
> I would love to help getting this PR merged, but I'm not sure what issues need to get fixed to get there. All the issues I discovered in my testings are fixed and I'm successfully running this branch on a staging environment for a few weeks now without any issues.

You probably can share your parameters? I compiled the PR successfully and started the s3 server but were not able to add buckets etc.

footlooseboss:
Perhaps try JuiceFS. Supports webdav/sftp storage (rclone serve’d), POSIX compliant, and can run an S3 gateway based on minio

individual-it:
The issue I've faced was the one described and fixed by @oniumy , it just never ended up in this PR. So a `go get github.com/Mikubill/gofakes3@723f1854b42fb1d141eb2893d21d8ed74169d1d4` fixed it

individual-it:
I've created a PR that solves the conflicts:  #7062
Should we close this and keep on working in the new one?

tomyl:
Hi, I gave this a try and it seems to work but the process gets OOM killed after a while. How to reproduce:
1. Create a large file e.g. ` dd if=/dev/zero of=/tmp/largefile bs=1M count=5000`
2. Add add an rclone backend like
```
[local]
type = local
```
3. Run the s3 server: `./rclone serve s3 local:/tmp/rclone --addr localhost:8081 --s3-authkey test,test --no-modtime --nochecksum
4. Upload the large file a few times e.g. with `mc`.
```bash
$ cat ~/.mc/config.json
....
        "rclone": {
            "url": "http://localhost:8081",
            "accessKey": "test",
            "secretKey": "test",
            "api": "S3v4",
            "path": "auto"
        }
...
$ mc mb rclone/foo
$ mc cp /tmp/largefile rclone/foo/largefile                            
$ mc cp /tmp/largefile rclone/foo/largefile           
...    
```
It looks like each upload consumes about 2x the file size in memory and the memory is not released (fast enough?) after upload and eventually you get
```
[12127.265358] Out of memory: Killed process 131176 (rclone) total-vm:32901932kB, anon-rss:22842736kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:45196kB oom_score_adj:0
```

The branch from https://github.com/rclone/rclone/pull/7062 behaves the same.

individual-it:
@tomyl hi could you try to set `GOMEMLIMIT` and see if it fixes the problem? https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

tomyl:
@individual-it Thanks, that indeed fixes the OOM panic for me. I tried `GOMEMLIMIT=1GiB` and the `rclone` process still consumes 2x file size during the upload (RSS goes up to 10G in my example above) but once the upload is done, RSS drops down to 1.7G each time.

ncw:
Hmm, this really needs fixing so the file gets streamed and not held in memory.

cmd/serve/s3/backend.go
```diff
@@ -0,0 +1,497 @@
...
+		return db.TouchObject(fp, meta)
+	}
+
+	f, err := db.fs.Create(fp)
+	if err != nil {
+		return result, err
+	}
+
+	hasher := md5.New()
+	w := io.MultiWriter(f, hasher)
```
sychan, line 310: This md5 hash object gets initialized, and then it gets a copy of all the bytes from the input. However hasher.Sum(nil) is never called, and the md5 results seem to be never used at all. There is a report that PUT operations use up memory equal to twice the input filesize, and never releases it, eventually resulting in an OOM condition. Could the md5 hasher be holding on to the memory and never releasing it because .Sum() is never called?

And if nothing is ever done with the md5sum, why not just remove it entirely?

ncw:
I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.65

Thank you everyone for contributing - it is a great feature.

@Mikubill would you like to move your gofakes3 repo to the rclone org and become a member of the rclone project?

I think we need to do major surgery on it to stop it caching multipart uploads in memory so it would be good if it was in the rclone project at that point.

Mikubill:
Great to hear and wow - thank you everyone for making this feature come true. Also i am happy to move gofakes3 to rclone and help strengthening codebase in the future.

ncw:
@Mikubill great to hear that :-) Drop me an email to nick@craig-wood.com and we can discuss mechanics.

PR #6471, opened 2022-09-28
Easier generation of version information and icon resources on Windows

albertony:

Refactor windows .exe resource (version information and icon) handling, to make it easier (and more flexible) to use with local builds. Will also try to make it usable for building librclone.dll.

This succeeds #6467

#### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -216,7 +216,6 @@ jobs:
         shell: bash
         run: |
           if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then make release_dep_linux ; fi
-          if [[ "${{ matrix.os }}" == "windows-latest" ]]; then make release_dep_windows ; fi
           make ci_beta
         env:
           RCLONE_CONFIG_PASS: ${{ secrets.RCLONE_CONFIG_PASS }}
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ Thumbs.db
 __pycache__
 .DS_Store
 /docs/static/img/logos/
+resource_windows_*.syso
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ ifdef RELEASE_TAG
 	TAG := $(RELEASE_TAG)
 endif
 GO_VERSION := $(shell go version)
+GO_OS := $(shell go env GOOS)
 ifdef BETA_SUBDIR
 	BETA_SUBDIR := /$(BETA_SUBDIR)
 endif
@@ -46,7 +47,13 @@ endif
 .PHONY: rclone test_all vars version
 
 rclone:
+ifeq ($(GO_OS),windows)
+	go run bin/resource_windows.go -version $(TAG) -syso resource_windows_`go env GOARCH`.syso
+endif
 	go build -v --ldflags "-s -X github.com/rclone/rclone/fs.Version=$(TAG)" $(BUILDTAGS) $(BUILD_ARGS)
+ifeq ($(GO_OS),windows)
+	rm resource_windows_`go env GOARCH`.syso
+endif
 	mkdir -p `go env GOPATH`/bin/
 	cp -av rclone`go env GOEXE` `go env GOPATH`/bin/rclone`go env GOEXE`.new
 	mv -v `go env GOPATH`/bin/rclone`go env GOEXE`.new `go env GOPATH`/bin/rclone`go env GOEXE`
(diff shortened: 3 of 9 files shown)


albertony:
@ncw: If you have time for a quick early review, what do you think about the latest strategy (there are multiple commits that will have to be squashed, so just look at the overall/latest)?

ncw:
> @ncw: If you have time for a quick early review, what do you think about the latest strategy (there are multiple commits that will have to be squashed, so just look at the overall/latest)?

I gave it a quick review. It is quite a big change but the idea appears to be to factor the windows version building stuff out of the cross compiler into its own go file.

I think this is probably a good idea :-)

One thing that occurs to me is that is stale build info going to be left behind so you'll be forever using the build info of the last time you ran the build command? That might be a reason to run it when you type `make`?

albertony:
> the idea appears to be to factor the windows version building stuff out of the cross compiler into its own go file.

Exactly. To make it possible/easy to run outside of cross-compile.go, e.g. with my own "go builds", and then as an additional benefit easy to also do the same on librclone.dll.

Initially I just factored out the existing go functions, basically, but then in the final iteration I ended up more of a rewrite - from depending on a "go installed" goversioninfo.exe that are executed as subprocess (with additional moving parts, as the cross-compile.go would execute the "go run" helper script, which in turn wrote a temporary json file, and then it executed goversioninfo.exe with that json as input) , to adding goversioninfo to go.mod and use it as a lib from the "go run" helper script (which does not write temporary json file, nor executes subprocess).

I take it you do not oppose this strategy, and will continue with cleanup, testing, and will also look into your question/suggestion regarding make.

albertony:
> One thing that occurs to me is that is stale build info going to be left behind so you'll be forever using the build info of the last time you ran the build command? That might be a reason to run it when you type `make`?

Hm.. When running cross-compile it will cleanup the .syso file it generates. `make compile_all` and `make ci_beta` is using this, as you know, and these are used from GitHub CI. Should be good. Plain `make` (I assume that is the "rclone" target, performing "go build" and copy to GOPATH) will not generate the .syso, but should not find any stale build info either - if not an error has occurred with a previous CI run or something. So I think we are good. Though, I guess we could make the rclone target also generate the syso if we wanted to, and/or create a separate target to do it, that would be simple. Depends what we want..

ncw:
> Plain make (I assume that is the "rclone" target, performing "go build" and copy to GOPATH) 

It is

>  but should not find any stale build info either

OK sounds good :-)

> Though, I guess we could make the rclone target also generate the syso if we wanted to, and/or create a separate target to do it, that would be simple.

It would be nice if `make` made the fully decorated binary, but there is a lot of clever stuff in the cross compile script we shouldn't go breaking and it isn't that important :-)

Do you want o merge this for 1.62 (imminent) or wait for the 1.63?

albertony:
This can wait.

ncw:
1.62 is out the door :-) Let me know when you are ready with this - thanks.

ncw:
Still time to slip this into 1.63 at the last minute if you want?

albertony:
Maybe, but not sure if I will make it - don't wait up :)

ncw (approved):
Looks very nice - well done :-)

albertony:
Thanks. I just pushed to upstream to be able to check beta uploads.

I'm a bit worried the android-all build failure is related, but don't really understand it:

> write /tmp/gomobile-work-3[9](https://github.com/rclone/rclone/actions/runs/6693557057/job/18184942275#step:8:10)43130021/src-android-arm/go.mod
> go: error reading go.mod: missing module declaration. To specify the module path:
> go mod edit -module=example.com/mod
> gomobile: go mod tidy -v failed: exit status 1

ncw:
Is there a duplicate go.mod somewhere? That's what it looks like. One without a module line.

albertony:
> Is there a duplicate go.mod somewhere? That's what it looks like. One without a module line.

Right. There is only one in the git repo, but it seems to me the gomobile build generates one and then this is the one with that issue, for some reason:

> write /tmp/gomobile-work-2655124098/src-android-arm/go.mod

ncw:
OK that makes sense, but why do your changes provoke the issue?

I think this is probably something to do with the go mod.

A wild stab in the dark - try running `go mod tidy` check the changes in and retry the push.

albertony:
> A wild stab in the dark - try running `go mod tidy` check the changes in and retry the push.

Bingo! 😮

ncw:
Hooray! Squash merge when ready!

PR #6475, opened 2022-10-01
dlna listen and announce address filtering, announcement interval option

YanceyChiew:

When rclone_dlna is specified to listen to `0.0.0.0`, it will also listen to `[::]`, which may be accessed by devices outside the local area network. On the other hand, the UPnP protocol itself is also based on ipv4. This change ensures the convenience of using `0.0.0.0` without facing additional threats.

At the same time, the SSDP server used all available addresses on the interface when it announced the dlna service. Now it can only announce the addresses that are actually listening. 

High announcing frequency can also be a problem. Usually, a device only announces once when it joins a multicast, and sets a longer cache validity period. Now users can control it via flag `--announce-interval` to get the experience that suits them.

The above changes have been tested in my daily use for more than four months, and everything is normal.

---

In addition, I simply added a few suffix matches to the original function of loading external subtitles, and they work stably on supported media controllers.

And I'm wondering why the dlna server has a new icon, but the code that generates it is not compiled, and the rclone I installed through apt on debian11 is also an old version of the icon, so that I almost wanted to make my own icon before.

Adding more subtitle support and generating new icons doesn't seem like a good place here, but they're minor changes, so I'll try to submit them together.

no

#### Checklist


cmd/serve/dlna/dlna.go
```diff
@@ -279,7 +279,14 @@ func (s *server) resourceHandler(w http.ResponseWriter, r *http.Request) {
 // use s.Wait() to block on the listener indefinitely.
 func (s *server) Serve() (err error) {
 	if s.HTTPConn == nil {
-		s.HTTPConn, err = net.Listen("tcp", s.httpListenAddr)
+		// Specifying network as tcp in go will cause -
+		// the bind address 0.0.0.0 to listen to [::] at the same time.
```
ncw, line 283: Binding to 0.0.0.0 is also a global address. Its IPv4 only, whereas [::] is IPv6 only.

Home networks are likely to have a public IPv6 but a private IPv4 so in this case it might improve security, but it doesn't in the general case. Also I'd like to think that even home networks have a firewall for IPv6.

I see your comment later that ssdp only listens on the IPv4 multicast address which I think is a better reason - maybe the comment should be updated to that?

cmd/serve/dlna/dlna.go
```diff
@@ -336,6 +343,32 @@ func (s *server) startSSDP() {
...
 func (s *server) ssdpInterface(intf net.Interface) {
+	// Figure out whether should an ip be announced
+	ipfilterFn := func(ip net.IP) bool {
+		 listenaddr := s.HTTPConn.Addr().String()
+		 listenip := listenaddr[:strings.LastIndex(listenaddr, ":")]
+		 switch listenip {
+			  case "0.0.0.0":
+				   if strings.Contains(ip.String(), ":") {
+					    // Any ipv6 address should not be announced
+					    // because ssdp only listen on ipv4 multicast address
```
ncw, line 354: That's a good argument for not listening on IPv6

cmd/serve/dlna/dlnaflags/dlnaflags.go
```diff
@@ -36,6 +39,7 @@ var DefaultOpt = Options{
 	FriendlyName:   "",
 	LogTrace:       false,
 	InterfaceNames: []string{},
+	AnnounceInterval: 10 * time.Second,
```
ncw, line 42: Feel free to increase the default if you think it is wrong.

ncw (commented):
I think this looks OK!

I put some discussion inline.

Can you `go fmt` your files please and fix the lint problems

Thank you :-)

cmd/serve/dlna/dlna.go
YanceyChiew:
```patch
index 604e91f9b..579d695d8 100644
--- a/cmd/serve/dlna/dlna.go
+++ b/cmd/serve/dlna/dlna.go
@@ -279,11 +279,11 @@ func (s *server) resourceHandler(w http.ResponseWriter, r *http.Request) {
 // use s.Wait() to block on the listener indefinitely.
 func (s *server) Serve() (err error) {
        if s.HTTPConn == nil {
-               // Specifying network as tcp in go will cause -
-               // the bind address 0.0.0.0 to listen to [::] at the same time.
-               // And the latter may be a global address, so there is a security risk
+               // Currently, the SSDP server only listens on an IPv4 multicast address.
+               // Differentiate between two INADDR_ANY addresses,
+               // so that 0.0.0.0 can only listen on IPv4 addresses.
                network := "tcp4"
```
Is it ok to comment like this?

cmd/serve/dlna/dlnaflags/dlnaflags.go
YanceyChiew: I think, increasing the default value to 12 minutes, so according to the upstream SSDP [code](https://github.com/anacrolix/dms/blob/cf5ba6b1180099f2fe8dbec232ad10f072fb5d13/ssdp/ssdp.go#L174), the cache timeout is 1800 seconds, which may be more appropriate.

YanceyChiew:
> I put some discussion inline.

Thanks for the callouts and tips!

> Can you `go fmt` your files please and fix the lint problems

I'll do these and submit with some typos corrections, sorry for not writing the code in the correct format, I'm new to Go.

YanceyChiew:
That's strange, the lint check still fails.

After I made the new modifications, running `golangci-lint run` locally passed and only got three deprecation warnings.

What I did next was just merge the previous commit with the revised commit without touching the source code, but now I do have two more errors, but they appear in a different place than which in the output of workflow.

Although there is no specific information in the output of the workflow, I believe that the problem I faced before was a directly return should not be wrapped in curly brackets within the else branch.

After fixing it, my lint no longer reported errors, and now running it again, I got this:

```
 D:\..\..\rclone   dlna_addr_ssdp_op  golangci-lint run
level=warning msg="[runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
level=warning msg="[runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
level=warning msg="[runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused."
level=error msg="[linters context] typechecking error: D:\\Documents\\GitHub\\rclone\\librclone\\librclone.go:30:8: could not import C (cgo preprocessing failed)"
level=error msg="[linters context] typechecking error: C:\\Program Files\\Go\\src\\runtime\\cgo\\cgo.go:31:8: could not import C (cgo preprocessing failed)"
level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package : could not load export data: no export data for \"golang.org/x/mobile/event/key\""
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: inspect: failed to load package : could not load export data: no export data for \"golang.org/x/mobile/event/key\"\n\n"
```

I'm not sure what the problem is and if lint in the workflow throws the same error.

ncw:
The problem is that `cmd/serve/dlna/dlnaflags/dlnaflags.go` needs `go fmt` - that's all.

If you run `gofmt -d .` in the source root it will show you diffs for all the files which need fixing. Interestingly the auto generated file appears to need fixing but that isn't picked up by golangci-lint.

cmd/serve/dlna/dlnaflags/dlnaflags.go
ncw: Will this increase the time for clients to find the `rclone dlna` server?

cmd/serve/dlna/dlna.go
ncw: Looks great

cmd/serve/dlna/dlnaflags/dlnaflags.go
ncw: Thanks for explaining. This change sounds good.

ncw (approved):
This looks great now - thank you. I'll merge it now.

PR #6476, opened 2022-10-01
docs/content: fix Scaleway C14 broken link

marco-m:
It is now called Glacier and documentation is at https://www.scaleway.com/en/glacier-cold-storage/

On the other hand, it is unclear to me the role of "name": should it stay "C14" or should it become "Glacier" or "Scaleway Glacier" ?


ncw:
Given that the link links to the Scaleway section of the s3 docs and we have a link to there already under Scaleway, maybe we should just remove this line?

What do you think?

marco-m:
Following your observation, I read https://rclone.org/s3/#scaleway.
Indeed, it makes it clear that Scaleway S3 also offers Glacier and how to set the storage class.
So I agree, it makes sense to simply remove this line.
Will update the PR now.

marco-m:
@ncw ready to be reviewed again

ncw (approved):
Looks great thanks

Issue #6290, opened 2022-07-01
AzureBlob: use emulator on a remote machine (override default emulator endpoint)


maiorfi:
#### What problem are you are trying to solve?

Config a azureblob remote pointing to emulator running on a different machine (not on localhost)

#### How do you think rclone should be changed to solve that?

Allow emulator cabled endpoint url override (with endpoint you can already specify on remote advanced config)

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

PR #6477, opened 2022-10-03
azureblob: allow emulator account/key override

roelarents:

In #3285 support for `azureblob` was added. In #6290 the option was added to override the endpoint. This adds the option to override the account name and key also.

Note: this could be regarded backwards incompatible. E.g. for people who keep their production credentials in the config but only switch `use_emulator` on and off. In which case an attempt would be made to authorize with the emulator with wrong credentials.

I couldn't find any mentions of it.

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -595,7 +595,15 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	)
 	switch {
 	case opt.UseEmulator:
-		credential, err := azblob.NewSharedKeyCredential(emulatorAccount, emulatorAccountKey)
+		var actualEmulatorAccount = emulatorAccount
+		if opt.Account != "" {
+			actualEmulatorAccount = opt.Account
+		}
+		var actualEmulatorKey = emulatorAccountKey
+		if opt.Key != "" {
+			actualEmulatorKey = opt.Key
+		}
+		credential, err := azblob.NewSharedKeyCredential(actualEmulatorAccount, actualEmulatorKey)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ncw (commented):
This change looks good to me.

> Note: this could be regarded backwards incompatible. E.g. for people who keep their production credentials in the config but only switch use_emulator on and off. In which case an attempt would be made to authorize with the emulator with wrong credentials.

How likely do you think that is? I guess maybe a note in the docs saying prior to v1.60 that rclone ignored the account and key would at least document it?

ncw:
Agreed! Do you want to add the note in this PR?

roelarents:
Where should I put it exactly? I've added it in the azureblob docs now, in the emulator paragraph at the bottom.

roelarents:
By the way: instead of the additional `use_emulator` I would actually prefer that the full `endpoint` could be overridden. That saves an extra parameter in e.g. a production/test situation where the connection parameters are switched with config.

It would be backwards incompatible, because the `endpoint` param is [now interpreted as](https://github.com/rclone/rclone/blob/master/backend/azureblob/azureblob.go#L688) (part of) the hostname for the full endpoint url, i.e. `blob.core.windows.net`. Unless testing the endpoint parameter for starting with `http` (new style) or not (backwards compat)?

roelarents:
What can I do to get this in the next release?

roelarents:
Good one :). Thanks, no problem, I can see rclone keeps you very busy. (Which makes me happy, I used it to abstract away from 3 backends already.) The 1.60 mention has been updated. I will put the endpoint thing in another PR. We can see then if it's desirable at all.

ncw (approved):
Great - thank you :-)

Will run through the CI then merge if it goes green.

ncw:
Thank you :-)

PR #6482, opened 2022-10-09
--config-command-in and --config-command-out Functionality - Resolves #6447

### LINKED ISSUE #6447: --config-command="..." or --config-key-command="..." option
Version: rclone v1.59.1

It would be nice to have a `--config-command` as analogues to the `--password-command` so we can source our own encryption calls for the config file, or provide our own master encryption key (**not** password) to supplement in:

https://github.com/rclone/rclone/blob/f0e439de0df77b6ceee197d62f51b47455fe8a3a/fs/config/config.go#L80

https://github.com/rclone/rclone/blob/f0e439de0df77b6ceee197d62f51b47455fe8a3a/fs/config/config.go#L275

This is particularly useful when we have devices which have their own hardware-backed keystore, such as smartphones. I am [currently](https://github.com/termux/termux-api/issues/550) implementing in Termux the ability to encrypt and decrypt using AES in Android's keystore. The problem with the current method is that I would still need to include the encrypted master password somewhere (such as in a file or sharedpreferences), but it would be preferable if I could source the master key directly from the keystore, or call a command to provide the decrypted config, therefore making sure the password isn't stored anywhere feasibly accessible on the device (even in encrypted form), nor in memory/cache.

Android's keystore supports biometric and pattern authentication and can be unlocked for a certain duration, only when the device is unlocked, for x times, et cetera, so the security of the config file would be greater and never expose the key (encrypted or unencrypted) outside the call.

It would be something like [connection strings](https://rclone.org/docs/#connection-strings), having no persistent configuration data, but maintaining the convenience of not having to manually input the configuration details

EduardDurech:
Adds functionality for `--config-command-in` and `--config-command-out` as analogous to `--password-command`, allowing command functionality to generate config files (also allowing the config file to be stored for example in a secure hardware strongbox or remotely)

It serves as an interface to the existing `Save()` and `Load()`; therefore, functionality of flags such as `--config` (will specify output if `--[...]-in` supplied or input if `--[...]-out` supplied) and `--password-command` still work

Resolves #6447 

#### Checklist


EduardDurech:
@ncw these will auto-create `RCLONE_CONFIG_COMMAND_IN` and `RCLONE_CONFIG_COMMAND_OUT` EnvVars, correct?

EduardDurech:
I cannot "test" my tests, so hopefully somebody can double-check them

ncw:
Hi @EduardDurech, and thank you for this contribution.

We're doing a tidy-up of the PR backlog. This one has had no activity for over three years, and rclone's code and docs have moved on quite a bit since it was opened, so I'm going to close it to keep the queue manageable.

Issue #4553, opened 2020-09-02
build: ARMv6 builds for 1.53.0 result in "Illegal instruction" error


darren12345:
#### What is the problem you are having with rclone?

All commands (including running rclone with no arguments) result in the error "Illegal instruction".

#### What is your rclone version (output from `rclone version`)

1.53.0

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Linux Debian, 10.5, Arm 32 bit:

```
root@serverb3:~# uname -m
armv5tel
root@serverb3:~# uname -r
4.19.117-1

root@serverb3:~# cat /proc/cpu
cpu/     cpuinfo
root@serverb3:~# cat /proc/cpuinfo
processor       : 0
model name      : Feroceon 88FR131 rev 1 (v5l)
BogoMIPS        : 111.11
Features        : swp half thumb fastmult edsp
CPU implementer : 0x56
CPU architecture: 5TE
CPU variant     : 0x2
CPU part        : 0x131
CPU revision    : 1

Hardware        : Marvell Kirkwood (Flattened Device Tree)
Revision        : 0000
Serial          : 0000000000000000
```

####  Which cloud storage system are you using? (eg Google Drive)

pCloud, not that it matters.

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)

Any.

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

Not possible.

Downgraded to 1.52 and all works OK.

creativeprojects:
I can't seem to find proper documentation about it but I think I remember by default Go was compiling for ARMv6 if not specified otherwise. And it looks like you have an ARMv5 architecture.

What is surprising is that it was working in 1.52 though.

Would you be able to try to compile it for ARMv5? (works from a Linux/Windows/Mac machine)

```
$ git clone https://github.com/rclone/rclone.git
$ cd rclone
$ GOARM=5 GOARCH=arm go build
```

ncw:
I just check the build script

https://github.com/rclone/rclone/blob/7e2488af10b316115217dc8246024d3de692d283/bin/cross-compile.go#L77

We don't set GOARM explicitly unless building the ARMv7 build

And as far as I can tell, the default is to build for ARMv6 and always has been.

I looked through the Go changelog and I don't think the default has changed.

So what I think must have happened is that the startup instructions in go1.15 now include an instruction which ARMv5 can't execute.

[...]

PR #6505, opened 2022-10-21
install.sh: fix arm-v7 download

olefrost:

To download the correct executable om armv7 (E.g. RPi 2 Model B)

The observation:
https://forum.rclone.org/t/out-of-memory-after-moving-to-rpi-8gb-and-ssd/33521/8

This initially proposed solution:
https://forum.rclone.org/t/out-of-memory-after-moving-to-rpi-8gb-and-ssd/33521/19

The decision to match everything starting with armv7
https://forum.rclone.org/t/out-of-memory-after-moving-to-rpi-8gb-and-ssd/33521/27

#### Tests

The modification was tested on Ubuntu with different hardcoded settings of OS_type.

#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -106,6 +106,9 @@ case "$OS_type" in
   aarch64|arm64)
     OS_type='arm64'
     ;;
+  armv7*)
+    OS_type='arm-v7'
+    ;;
   arm*)
     OS_type='arm'
     ;;


albertony:
I agree with this change, just adding a note a bit on the side - if it is of any interest... 🤓 

I have experimented on an old router with ARM Cortex A9, 32-bit ARMv7-A architecture without hardware floating point support. Running `uname -m` lists it as "armv7l", but it will not work with the official "ARMv7 - 32 Bit" build of rclone as that requires hardfloat support. It will not work with "ARMv6 - 32 Bit" either, for same reason. What it would need is an "ARMv5 - 32 Bit" build, i.e. with `GOARM=5` (according to [this](https://github.com/golang/go/wiki/GoArm)). This is how [Entware](https://github.com/Entware/entware-go/blob/master/rclone/Makefile) builds the "rclone_nohf" (softfloat) variant for armv7.

We would then have the following Linux ARM builds:
- ARM - 64 Bit: `GOARCH=arm64`
- ARMv7 - 32 Bit: `GOARCH=arm GOARM=7`
- ARMv6 - 32 Bit: `GOARCH=arm` (implicit default `GOARM=6` according to [this](https://go.dev/doc/install/source#environment))
- ARMv5 - 32 Bit: `GOARCH=arm GOARM=5` (my little feature request 😈)

Now, when it comes to the install.sh script: This router would be detected as "armv7l", but we cannot deduce from that that it requires the ARMv5 version. The "l" suffix only means "little-endian", I think. Some processors with hardfloat identifies as "armv7hl", with an extra "h" to indicate hardfloat, but this is not consistent so lack of this "h" does not necessarily mean softfloat. Tough luck, I guess we just have to assume "ARMv7 - 32 Bit" for any "armv7*" like you have suggested here! Or, ideally, we could have made the check more advanced; find some way to deduce hardfloat/softfloat/armv5/armv6/armv7 compatibility, maybe via `cat /proc/cpuinfo`, don't know. In any case way out of scope for this PR. Oh well, TL;DR, I know.. 😩

albertony (approved):

ncw:
> I have experimented on an old router with ARM Cortex A9, 32-bit ARMv7-A architecture without hardware floating point support.

Urgh! How old is this router? I would have thought its kernel or userspace should be emulating the floating point - that is what the 

There was a time that the arm compile was v5 IIRC but it changed at some point...

See also #4553 #4553

Anyway I'll merge this fix - thank you @olefrost

olefrost:
@albertony  DR;TL (Did Read; Truly Liked) 🤓

It very much echoes the research and thoughts I had, so I very much appreciate your comments.

You are right the "l" suffix means "little-endian" and there also seams to be a (less used) armv7b (big-endian), which probably cannot use the arm-v7 download - nor the arm download. The more you dig into it, the more it seems to be a nightmare, at some point I was desperately considering this change instead 🤣

```
  arm*)
    echo 'Sorry, arm is just too complicated, go figure out yourself!'
    exit 2
    ;;
```

@ncw Thanks!

albertony:
> The more you dig into it, the more it seems to be a nightmare

Definitely, my impression as well!

> at some point I was desperately considering this change instead

I like it🥇 😆

ncw:
> Or, ideally, we could have made the check more advanced; find some way to deduce hardfloat/softfloat/armv5/armv6/armv7 compatibility, maybe via `cat /proc/cpuinfo`,

I think if we can figure out a reliable way of detecting that the arm core doesn't not have hardfloat support in install.sh then I'd consider an ARMv5 architecture! Though the fact that we haven't had any complaints about it since v5 support was dropped means that there probably aren't many users!

PR #6507, opened 2022-10-21
Clean absolute paths in local backend

### LINKED ISSUE #6493: rclone sync creates a directory named ".." when destination path contains /../
**What is the problem you are having with rclone?**

If you launch the command;
`rclone sync <a local path> <some dir>/parentDir/../<some dir>
`
so a synchronisation on a local machine from a local directory to another local directory, it will create a directory named ".." in parentDir instead of going one directory up.

**What is your rclone version (output from rclone version)**

All versions >= 1.50 (it works in 1.49)

**Which OS you are using and how many bits (e.g. Windows 7, 64 bit)**

Windows 11 Pro 21H2  (arch: amd64) and macOS 12.6 (arch: arm64)

**Which cloud storage system are you using? (e.g. Google Drive)**

Local filesystem

**The command you were trying to run (e.g. rclone copy /tmp remote:tmp)**

`rclone sync XXX XXX/../XXX
`

albertony:

In existing version, relative paths are made absolute with `filepath.Abs`, which implicitely performs a `filepath.Clean` which eliminates path elements such as `..`. But paths that are already absolute are not cleaned, which means a path element `..` will be kept and encoded according to backend encoding.

This PR adds a call to `filepath.Clean` for absolute paths, making it more consistent.

Windows paths with prefix `\\` are excluded from this, they were already excluded from `filepath.Abs` in existing version, as UNC paths cannot be relative (this is documented) nor contain "relative" path elements (undocumented, but at least the `dir` command such as `dir \\?\C:\Some\Path\..\OtherPath` errors with "The filename, directory name, or volume label syntax is incorrect", and also, in any case, this means using `\\?` prefix can be used as workaround in case the Clean/Abs leads to problems).

Fixes #6493

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1400,29 +1400,26 @@ func (o *Object) writeMetadata(metadata fs.Metadata) (err error) {
 }
 
 func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
-	if runtime.GOOS == "windows" {
-		if !filepath.IsAbs(s) && !strings.HasPrefix(s, "\\") {
+	if runtime.GOOS != "windows" || !strings.HasPrefix(s, "\\") {
+		if !filepath.IsAbs(s) {
 			s2, err := filepath.Abs(s)
 			if err == nil {
 				s = s2
 			}
+		} else {
+			s = filepath.Clean(s)
 		}
+	}
+	if runtime.GOOS == "windows" {
 		s = filepath.ToSlash(s)
 		vol := filepath.VolumeName(s)
 		s = vol + enc.FromStandardPath(s[len(vol):])
 		s = filepath.FromSlash(s)
-
 		if !noUNC {
 			// Convert to UNC
 			s = file.UNCPath(s)
 		}
 		return s
-	}
-	if !filepath.IsAbs(s) {
-		s2, err := filepath.Abs(s)
-		if err == nil {
-			s = s2
-		}
 	}
 	s = enc.FromStandardPath(s)
 	return s


backend/local/local.go
```diff
@@ -1423,6 +1427,8 @@ func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
 		if err == nil {
 			s = s2
 		}
+	} else {
```
ncw, line 1430: We probably don't need to so this on Windows now do we? Either we've made the UNC already which is by definition absolute or the path is Abs and clean.

backend/local/local.go
```diff
@@ -1401,10 +1401,14 @@ func (o *Object) writeMetadata(metadata fs.Metadata) (err error) {
 
 func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
 	if runtime.GOOS == "windows" {
-		if !filepath.IsAbs(s) && !strings.HasPrefix(s, "\\") {
-			s2, err := filepath.Abs(s)
-			if err == nil {
-				s = s2
+		if !strings.HasPrefix(s, "\\") {
```
ncw, line 1404: Read comment below first...

Maybe we should move this if statement above `if runtime.GOOS == "windows"` and make it work for all OSes

```go
if rutime.GOOS != "windows" || !strings.HasPrefix(s, "\\") {
```

Then we can avoid duplicating the following 8 lines of code

Note that `runtime.GOOS == "windows"` is evaluated entirely at compile time so effectively becomes a constant `true` or `false` so duplicating that is free!

ncw (commented):
Good catch.

See inline comments - do you agree with that logic?

albertony:
Although a bit confused, I *think* I understood what you meant... Pushed a new version, so you tell me! :)

> Note that runtime.GOOS == "windows" is evaluated entirely at compile time so effectively becomes a constant true or false so duplicating that is free!

Good to know, thanks.

ncw (approved):
Perfect, that's exactly what I meant 🙂

Please merge when ready

PR #6510, opened 2022-10-23
wasm: fix walltime link error

joaohenrifranco:

Fixes the following error while running [rclone as wasm](https://github.com/rclone/rclone/tree/master/fs/rc/js): ```LinkError: import object field 'runtime.walltime' is not a Function```.

[`walltime1` function was renamed to `walltime`](https://go-review.googlesource.com/c/go/+/314770/4), therefore code on wasm_exec.js was updated accordingly.

No.

#### Checklist

### DIFF
--- a/fs/rc/js/wasm_exec.js
+++ b/fs/rc/js/wasm_exec.js
@@ -2,587 +2,658 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+"use strict";
+
 (() => {
-	// Map multiple JavaScript environments to a single common API,
-	// preferring web standards over Node.js API.
-	//
-	// Environments considered:
-	// - Browsers
-	// - Node.js
-	// - Electron
-	// - Parcel
+  const enosys = () => {
+    const err = new Error("not implemented");
+    err.code = "ENOSYS";
+    return err;
+  };
 
-	if (typeof global !== "undefined") {
-		// global already exists
-	} else if (typeof window !== "undefined") {
-		window.global = window;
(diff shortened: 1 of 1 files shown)


ncw:
Well spotted.

Did you include the whole new wasm_exec.js ?

How can we stop this happening again?

The CI tests compile wasm they don't actually run it though.

Maybe they should somehow? Any ideas?

Maybe the tests should check the wasm_exec.js matches the current version?

joaohenrifranco:
> Did you include the whole new wasm_exec.js ?

I didn't initially, I trying only solving the error blocking execution. But thanks for pointing that out. Just added the [last version](https://github.com/golang/go/blob/master/misc/wasm/go_js_wasm_exec) and checked it is working fine.

> How can we stop this happening again?

I'm inclined to think that any sort of automation is premature at this point, as the wasm version is still "experimental". Also, using always the most updated version might break things unexpectedly, maybe it should continue to be updated on demand?

>The CI tests compile wasm they don't actually run it though.
Maybe they should somehow? Any ideas?

Tried some ideas here such as starting the server and checking if rc module was successfully initialized by placing some HTML in the body and asserting it is contained in body but it was't a good solution. For the moment I don't have any suggestions :/

> Maybe the tests should check the wasm_exec.js matches the current version?

Wouldn't that envolve some request to the golang repo to assert if file there matches our version? Making external request doesn't seem suitable inside automated tests. Any other ideas?

ncw (approved):
Thank you for doing this. I'm going to merge this now and write a little dev script to update wasm_exec.js for the future.

PR #6512, opened 2022-10-23
Add vulnerability testing using govulncheck to build pipeline

albertony:

Running govulncheck as part of the build workflow, so that a "check" in pull requests etc will fail if there are any vulnerabilities detected. See https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck for background.

Note:
- Govulncheck must be built with Go version 1.18 or later. Rclone currently supports 1.17, and builds on 1.17, 1.18 and 1.19.
- Govulncheck only reports vulnerabilities that apply to the current Go version. For example, a standard library vulnerability that only applies for Go 1.18 will not be reported if the current Go version is 1.19. See https://go.dev/issue/54841 for updates to this limitation.
- According to https://go.dev/issue/54841, it will report vulnerabilities for any platform regardless which platform it runs on (if I understand it correctly).
- The above means that:
   - We don't need to run it for each of the individual platforms in the build matrix - one OS is enough (e.g. ubuntu-latest).
   - We should probably run it for all supported go versions > 1.17, i.e. currently 1.18 and 1.19 (or we *could* decide that running on latest is enough?).
   - And the above means that (sorry for the multi-levels...):
      - The govulncheck is not part of the build job definition in build.yml
      - The govulncheck is not part of the lint job definition in build.yml
      - The govulncheck is a separate job with matrix running for each go version.
         - Takes about 1m 10s each.

~TODO~Done:
- ~Update according to [#6513](https://github.com/rclone/rclone/pull/6513) if/when merged.~ (done)
- ~Found 1 known vulnerability that we need to fix:~ (done)
```
Vulnerability #1: GO-2022-0969
  HTTP/2 server connections can hang forever waiting for a clean
  shutdown that was preempted by a fatal error. This condition can
  be exploited by a malicious client to cause a denial of service.

  Call stacks in your code:
Error:       cmd/serve/restic/restic.go:150:22: github.com/rclone/rclone/cmd/serve/restic.init$1$1 calls golang.org/x/net/http2.Server.ServeConn

  Found in: golang.org/x/net/http2@v0.0.0-20220[8](https://github.com/rclone/rclone/actions/runs/3307150451/jobs/5458560848#step:6:9)05013720-a33c5aa5df48
  Fixed in: golang.org/x/net/http2@v0.0.0-20220[9](https://github.com/rclone/rclone/actions/runs/3307150451/jobs/5458560848#step:6:10)06165[14](https://github.com/rclone/rclone/actions/runs/3307150451/jobs/5458560848#step:6:15)6-f3363e06e74c
  More info: https://pkg.go.dev/vuln/GO-2022-0969
```

https://forum.rclone.org/t/go-vulncheck-results-suggestion/33574/10

#### Checklist


ncw (commented):
This is nice :-)

I don't think we need to run it on anything except the latest go version - that is the go version we build our binaries with.

I wonder about putting it in with the `lint` action to save resources? Or do you think it merits its own builder?

albertony:
@ncw Pushed changes:
- Moved the new govulncheck feature into existing lint action, running on a single (current/latest) version of go only.
- Updated x/net to fix detected vulnerability (`go get golang.org/x/net`)

ncw (approved):
Perfect, thank you :-)

Please merge

Issue #5066, opened 2021-02-25
s3: seems to not figure out bucket zone "automagically"


yarikoptic:
#### What is the problem you are having with rclone?

it works so good that I want to (ab)use it more, but it is resisting ;)

summary: having a generic s3 configuration without zone specification seems to work only for some public buckets (in `us-east-1`?) zone but not the others.

<details>
<summary>
Not sure how, but e.g. python's good old boto 2.49.0 seems to connect to sample buckets anonymously just fine without any zone specification:
</summary> 

```shell
$> python3 -c 'import boto; print(next(iter(boto.connect_s3(anon=True).get_bucket("fcp-indi").list_versions())))'
<boto.s3.deletemarker.DeleteMarker object at 0x7f81e2e679d0>

$> python3 -c 'import boto; print(next(iter(boto.connect_s3(anon=True).get_bucket("openfmri").list_versions())))'  
<Key: openfmri,corral-repl/utexas/poldracklab/openfmri/inprocess/ds113c/ds113c-gitannex/tarballs/ds113c_R1.0.0_metadata.tar.gz>

$> python3 -c 'import boto; print(next(iter(boto.connect_s3(anon=True).get_bucket("dandiarchive").list_versions())))'
<Key: dandiarchive,dandiarchive/dandiarchive/2019-10-02T04-00Z/manifest.checksum>
```
</details>

whenever rclone would work for first two but fail for the last:

```shell
$> tail -3 ~/.config/rclone/rclone.conf
[s3]
type = s3
provider = AWS

$> rclone ls s3:dandiarchive       
2021/02/24 19:16:53 ERROR : S3 bucket dandiarchive: Failed to update region for bucket: reading bucket location failed: AccessDenied: Access Denied
	status code: 403, request id: 1C82D50F1B54F94D, host id: P2abi9H9rNSseBsDWVHlgw+rRalYTtAyakieS269zt2ltlPaFzUH1WhmLS3dMhKzZZ47KRxf8fM=
2021/02/24 19:16:53 Failed to ls: BucketRegionError: incorrect region, the bucket is not in 'us-east-1' region at endpoint ''
	status code: 301, request id: , host id: 
```

For this bucket I could even query for the zone:

```shell
$> python3 -c 'import boto; print(boto.connect_s3(anon=True).get_bucket("dandiarchive").get_location())' 
us-east-2
```

And rclone does work if I use rclone config where I did specify zone:

```
[s3_us_east_2]
type = s3
provider = AWS
[...]

PR #6525, opened 2022-10-29
s3: use different strategy to resolve s3 region

panthony:

Improve how rclone figure out where an S3 region is located.

Today it relies on `GetBucketLocation` endpoint but it requires top level permission.

Another method is to simply perform an HEAD request to S3 with the bucket as host and it will return its location
in the HTTP response headers.

You can see by using:

```
curl -v -X HEAD --header "Host: {bucket}.s3.amazonaws.com" s3.amazonaws.com
```

And looking for `x-amz-bucket-region` header.

See https://github.com/aws/aws-cli/issues/2431#issuecomment-278725415

See https://github.com/rclone/rclone/issues/5066

#### Notes

At first I tried to use AWS SDK directly but the method `HeadBucketRequest` does not perform this exact query, if there is a better way to achieve this let me know!

Regarding tests I'm not quite sure how to do in this context.

#### Checklist


backend/s3/s3.go
```diff
@@ -2992,19 +2993,12 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
...
-		Bucket: &bucket,
-	}
-	var resp *s3.GetBucketLocationOutput
-	var err error
-	err = f.pacer.Call(func() (bool, error) {
-		resp, err = f.c.GetBucketLocation(&req)
-		return f.shouldRetry(ctx, err)
-	})
+	sess := session.Must(session.NewSession())
+	region, err := s3manager.GetBucketRegion(ctx, sess, bucket, "us-east-1")
```
ncw, line 2997: Reading the docs for s3manager.GetBucketRegion we should probably be forcing PathStyle = false, or perhaps better setting it to what it is set to in the config `f.opt.ForcePathStyle`

```
        region, err := s3manager.GetBucketRegion(ctx, sess, "bucketname", "us-west-2", func(r *request.Request) {
            r.S3ForcePathStyle = aws.Bool(f.opt.ForcePathStyle)
        })
```

Actually thinking on this a bit more, we should probably use use `f.ses` as the session as it will have all the right settings in already. It will do an authenticated request rather than unauthenticated, but I don't suppose that will be a problem.

Do you think the region hint should be the users current region in the config `f.opt.Region`? This will default to `us-east-1` if not set.

I don't really understand what the region hint is for!

ncw (commented):
This is a great fix - thank you :-)

I put some comments about some of the details inline.

backend/s3/s3.go
panthony: > Do you think the region hint should be the users current region in the config f.opt.Region?

Sure! I wasn't aware it was accessible in the context.

> I don't really understand what the region hint is for!

Me neither actually because using `curl` I can target `s3.amazonaws.com` directly and it works. I do not need to target a specific region's endpoint (ex: `s3.us-east-1.amazonaws.com`).

The SDK also clearly check for `regionHint` being empty:

https://github.com/aws/aws-sdk-go/blob/36e99e2d8b7f9efaa27e9665760788b3169e24bb/service/s3/s3manager/bucket_region.go#L65

But the comment also says:

```
// If the regionHint parameter is an
// empty string GetBucketRegion will fallback to the ConfigProvider's region
// config. If the regionHint is empty, and the ConfigProvider does not have a
// region value, an error will be returned..
```

If I use `f.ses` I might not need to pass the region here if I assume it is already configured with it.

panthony: For `f.opt.ForcePathStyle` I'll do the change, I did not even see that flag already existed in rclone 🤦🏻

backend/s3/s3.go
```diff
@@ -2992,19 +2993,12 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
...
-	req := s3.GetBucketLocationInput{
-		Bucket: &bucket,
-	}
-	var resp *s3.GetBucketLocationOutput
-	var err error
-	err = f.pacer.Call(func() (bool, error) {
-		resp, err = f.c.GetBucketLocation(&req)
-		return f.shouldRetry(ctx, err)
-	})
+	sess := session.Must(session.NewSession())
```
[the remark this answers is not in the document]
panthony: I was able to use `f.ses` instead which comes from pre-configured region thus I do not even need to pass `regionHint` anymore!

Much cleaner :)

backend/s3/s3.go
panthony: PR updated with both changes for `f.ses` & `f.opt.ForcePathStyle` which is much cleaner 👍🏻

7qqq:
Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

backend/s3/s3.go
```diff
@@ -2992,19 +2993,13 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
...
-	req := s3.GetBucketLocationInput{
-		Bucket: &bucket,
-	}
-	var resp *s3.GetBucketLocationOutput
-	var err error
-	err = f.pacer.Call(func() (bool, error) {
-		resp, err = f.c.GetBucketLocation(&req)
-		return f.shouldRetry(ctx, err)
+	region, err := s3manager.GetBucketRegion(ctx, f.ses, bucket, "", func(r *request.Request) {
+		r.Config.S3ForcePathStyle = aws.Bool(f.opt.ForcePathStyle)
```
ncw, line 2997: Is this necessary if we use f.ses which will already be configured with path style?

panthony: It is necessary because `GetBucketRegion` forces it to true:

https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/bucket_region.go#L108

Just to be safe I added a log in the callback:

```go
	region, err := s3manager.GetBucketRegion(ctx, f.ses, bucket, "", func(r *request.Request) {
		log.Printf("Live S3ForcePathStyle=%v", *r.Config.S3ForcePathStyle)
		log.Printf("Configured S3ForcePathStyle=%v", f.opt.ForcePathStyle)
	})
```

Got:
```
2022/10/31 07:24:03 Live S3ForcePathStyle=true
2022/10/31 07:24:03 Configured S3ForcePathStyle=false
```

ncw: Nice - thanks for checking

ncw (approved):
This looks great now - I'll merge now

Thank you for a very useful contribution :-)

panthony:
Updated the PR after running `golangci-lint run --fix` to correct lint error:

> File is not `goimports`-ed (goimports)

Since the import was added automatically by the IDE & `go fmt` was changing nothing I assumed it was ok, apparently not.

ncw:
Thank you :-)

PR #6529, opened 2022-10-29
docs: improve google drive usage information

fzn0x:
Just leaving `remote` is confusing. Modify it to something user must aware of.

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -99,17 +99,21 @@ firewall, or use manual mode.
 
 You can then use it like this,
 
+To see the name of your remote name
+
+    rclone config
+
 List directories in top level of your drive
 
-    rclone lsd remote:
+    rclone lsd <remote_name>:
 
 List all the files in your drive
 
-    rclone ls remote:
+    rclone ls <remote_name>:
 
 To copy a local directory to a drive directory called backup
 
-    rclone copy /home/source remote:backup
+    rclone copy /home/source <remote_name>:backup
 
 ### Scopes
 
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -99,21 +99,19 @@ firewall, or use manual mode.
 
 You can then use it like this,
 
-To see the name of your remote
-
-    rclone config
-
 List directories in top level of your drive
 
-    rclone lsd <remote_name>:
+    rclone lsd remote:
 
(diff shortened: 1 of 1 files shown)


albertony:
I understand this can be confusing, and perhaps we should come up with better naming in the examples. But, in the interactive sequence shown just above your changes the name `remote` is given as input to the `rclone config` command. So therefore, running `rclone lsd remote:` follows naturally from that.

```
No remotes found, make a new one?
n) New remote
r) Rename remote
c) Copy remote
s) Set configuration password
q) Quit config
n/r/c/s/q> n
name> remote
```

Also, this part of the documentation is very similar in many of the backends, so if we are going to change it I think we'll have to do it for all of them.

Perhaps you could create an issue for this instead, where possible improvements can be discussed and agreed upon, before we consider doing changes that affect many of the backend doc pages?

fzn0x:
> I understand this can be confusing, and perhaps we should come up with better naming in the examples. But, in the interactive sequence shown just above your changes the name `remote` is given as input to the `rclone config` command. So therefore, running `rclone lsd remote:` follows naturally from that.
> 
> ```
> No remotes found, make a new one?
> n) New remote
> r) Rename remote
> c) Copy remote
> s) Set configuration password
> q) Quit config
> n/r/c/s/q> n
> name> remote
> ```
> 
> Also, this part of the documentation is very similar in many of the backends, so if we are going to change it I think we'll have to do it for all of them.
> 
> Perhaps you could create an issue for this instead, where possible improvements can be discussed and agreed upon, before we consider doing changes that affect many of the backend doc pages?

We can discuss it here I guess

fzn0x:
I changed it to warning message, someone may ignore the configuration steps after runs `rclone config`.

albertony:
I'm still not sure about this one.

I do agree there can be some confusion around this for newcomers, but my personal opinion is that it is a bit too "noisy" with a warning like the latest change. And all backends use the same "template" for this docs, so an added description like this would have to be in all of them, not only drive.

The section does already start by this short explanation:

> Here is an example of how to make a remote called `remote`.

And also we have the usage docs which users should definitely take a look at first, e.g. the https://rclone.org/docs/#syntax-of-remote-paths section is highly relevant, so its a question how much of this kind of "generic help" we want in the backend docs.

PR #6539, opened 2022-11-04
Ncdu improvements

### LINKED ISSUE #6537: ncdu: Ability to see and sort on date/time

~~~
rclone v1.60.0
- os/version: darwin 10.14.6 (64 bit)
- os/kernel: 18.7.0 (x86_64)
- os/type: darwin
- os/arch: amd64
- go/version: go1.19.2
- go/linking: dynamic
- go/tags: cmount
~~~

Browse files in a useful manner :)

I have an Azure Blob Storage container in which there are a lot of files, all named with a SHA256 hash. The Azure portal shows columns for the name of course, and also the modification time.

I would like to sort the list on modification time, such that I can see the most recent files without having to scroll a ton of pages down while at the same time waiting for it to load more lines over and over again.

Unfortunately, one cannot sort the list in the Azure portal, it just lists all of the blobs/objects by their name, which when they're all hashes is totally useless to see the most recent ones.

I tried using `rclone ncdu` to view the list in a better way, but it seems that it 1) does not show any date/time info for the blobs/objects, and 2) does not allow to sort the list on a date/time.

The feature request is simple: Make ncdu able to show date/time information (at least modification time, but also others if there are more), and also enable the user to sort the list on these dates/times.

#### How to use GitHub

albertony:

Some smaller improvements to ncdu. The biggest being added support for showing modification times.

Modification times can now be togled with `m` and sorted with `M`, like the [original ncdu](https://dev.yorhel.nl/ncdu) tool. Time format is the same as when testing ncdu 1.17 (on Fedora and Alpine): "2006-01-02 15:04:05 -0700". One downside of this format is that it contains spaces, while other fields like count and size in the output are separated by space.

Example output, with everything (size, count, average and modified time) shown:

```
     153        1       153 2021-04-15 18:11:40 +0200 [          ] /i18n
21.252Ki       23       946 2021-04-15 18:11:40 +0200 [          ] /layouts
 3.019Mi       28 110.402Ki 2021-04-15 18:11:40 +0200 [##########] /static
     757                    2021-08-31 09:56:44 +0200 [          ]  config.json
 3.677Ki                    2022-08-15 07:36:44 +0200 [          ]  README.md
 1.657Mi      158  10.737Ki 2022-11-04 11:30:06 +0100 [#####     ] /content   
```
        
(Originally also adjusted how empty directories are displayed, making them shown with count and average size as value 0 instead of blank, so that blank in these columns are only for files, but discovered that the original ncdu version tested above shows them like rclone ncdu did before, so dropped these changes for consistency - assuming this ncdu version is representative).

Fixes #6537

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -89,11 +89,12 @@ func helpText() (tr []string) {
 		" ↑,↓ or k,j to Move",
 		" →,l to enter",
 		" ←,h to return",
-		" c toggle counts",
 		" g toggle graph",
+		" c toggle counts",
 		" a toggle average size in directory",
+		" m toggle modified time",
 		" u toggle human-readable format",
-		" n,s,C,A sort by name,size,count,average size",
+		" n,s,C,A,M sort by name,size,count,asize,mtime",
 		" d delete file/directory",
 		" v select file/directory",
 		" V enter visual select mode",
@@ -131,12 +132,14 @@ type UI struct {
 	showGraph          bool          // toggle showing graph
 	showCounts         bool          // toggle showing counts
 	showDirAverageSize bool          // toggle average size
+	showModTime        bool          // toggle showing timestamps
 	humanReadable      bool          // toggle human-readable format
 	visualSelectMode   bool          // toggle visual selection mode
-	sortByName         int8          // +1 for normal, 0 for off, -1 for reverse
-	sortBySize         int8
+	sortByName         int8          // +1 for normal (lexical), 0 for off, -1 for reverse
+	sortBySize         int8          // +1 for normal (largest first), 0 for off, -1 for reverse (smallest first)
 	sortByCount        int8
 	sortByAverageSize  int8
+	sortByModTime      int8              // +1 for normal (newest first), 0 for off, -1 for reverse (oldest first)
 	dirPosMap          map[string]dirPos // store for directory positions
 	selectedEntries    map[string]dirPos // selected entries of current directory
 }
@@ -332,6 +335,7 @@ func (u *UI) hasEmptyDir() bool {
 
 // Draw the current screen
 func (u *UI) Draw() error {
+	ctx := context.Background()
 	w, h := termbox.Size()
 	u.dirListHeight = h - 3
 
@@ -365,7 +369,13 @@ func (u *UI) Draw() error {
 			if y >= h-1 {
 				break
 			}
-			attrs, err := u.d.AttrI(u.sortPerm[n])
+			var attrs scan.Attrs
+			var err error
+			if u.showModTime {
+				attrs, err = u.d.AttrWithModTimeI(ctx, u.sortPerm[n])
+			} else {
+				attrs, err = u.d.AttrI(u.sortPerm[n])
+			}
 			_, isSelected := u.selectedEntries[entry.String()]
 			fg := termbox.ColorWhite
 			if attrs.EntriesHaveErrors {
@@ -420,6 +430,9 @@ func (u *UI) Draw() error {
 				} else {
 					extras += strings.Repeat(" ", len(ss))
 				}
+			}
+			if u.showModTime {
+				extras += attrs.ModTime.Local().Format("2006-01-02 15:04:05") + " "
 			}
 			if showEmptyDir {
 				if attrs.IsDir && attrs.Count == 0 && fileFlag == ' ' {
@@ -656,8 +669,15 @@ type ncduSort struct {
 // Less is part of sort.Interface.
 func (ds *ncduSort) Less(i, j int) bool {
 	var iAvgSize, jAvgSize float64
-	iattrs, _ := ds.d.AttrI(ds.sortPerm[i])
-	jattrs, _ := ds.d.AttrI(ds.sortPerm[j])
+	var iattrs, jattrs scan.Attrs
+	if ds.u.sortByModTime != 0 {
+		ctx := context.Background()
+		iattrs, _ = ds.d.AttrWithModTimeI(ctx, ds.sortPerm[i])
+		jattrs, _ = ds.d.AttrWithModTimeI(ctx, ds.sortPerm[j])
+	} else {
+		iattrs, _ = ds.d.AttrI(ds.sortPerm[i])
+		jattrs, _ = ds.d.AttrI(ds.sortPerm[j])
+	}
 	iname, jname := ds.entries[ds.sortPerm[i]].Remote(), ds.entries[ds.sortPerm[j]].Remote()
 	if iattrs.Count > 0 {
 		iAvgSize = iattrs.AverageSize()
@@ -678,6 +698,14 @@ func (ds *ncduSort) Less(i, j int) bool {
 	case ds.u.sortBySize > 0:
 		if iattrs.Size != jattrs.Size {
 			return iattrs.Size > jattrs.Size
+		}
+	case ds.u.sortByModTime < 0:
+		if iattrs.ModTime != jattrs.ModTime {
+			return iattrs.ModTime.Before(jattrs.ModTime)
+		}
+	case ds.u.sortByModTime > 0:
+		if iattrs.ModTime != jattrs.ModTime {
+			return iattrs.ModTime.After(jattrs.ModTime)
 		}
 	case ds.u.sortByCount < 0:
 		if iattrs.Count != jattrs.Count {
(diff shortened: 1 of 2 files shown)


rawtaz:
@albertony Really nice quick work there! I have built rclone with your branch myself and tested it manually, including toggling the mtime column(s) as well as sorting back and forth. From what I can tell it works.

There is one weird thing though. I don't know if it's my inexperience with Azure Blob Storage in particular or something else. I'm looking at a [restic](https://restic.net) repository here. If I start at the root of the storage account, then all "folders" in the path `/test1/backup/snapshots/` are shown with timezone +0100 which is CET and is correct as that is what I have on my system here. But when I enter into the `snapshots/` "folder", all the actual blobs/objects in there are shown with +0000 (UTC) timezone.

Example when viewing the "folder" `snapshots/`:
~~~
-- mystorageaccountremote:test1/backup ----------------------
  220.647Mi 2022-11-04 15:24:45 +0100 [##########] /data                                                                                                                                                                                                                      
    1.185Mi 2022-11-04 15:24:45 +0100 [          ] /index                                                                                                                                                                                                                     
   24.650Ki 2022-11-04 15:24:45 +0100 [          ] /snapshots        
~~~

Example when viewing the blobs in `snapshots/`:
~~~
-- mystorageaccountremote:test1/backup/snapshots ------------
        238 2022-11-04 01:59:06 +0000 [##########]  29802e0d7646216442445babab1de70942fedda45e95b79aafa7f39aae48ee37                                                                                                                                                          
        238 2022-11-04 01:33:22 +0000 [##########]  3fb3b69bef91f21aefe72967fe08abbf90087a976fe839b189f77675daced744                                                                                                                                                          
        238 2022-11-04 01:58:36 +0000 [##########]  4095683e7d2c4a1bb78244bfb4bfcc2cf9167d98d18f8c8d682331647af6352f                                                                                                                                                          
~~~

I think it would be good from a UX perspective to see everything (paths and blobs/objects) with the same timezone. I'm not sure if it should be UTC or the local timezone (in my case CET). Personally I would like to see it primarily with my local timezone, as otherwise I have to do math, and we all know how that might go.

What do you think about a way to toggle the display of timezones either between UTC and local timezone, or even between UTC, local timezone and "original" timezone as stored in Azure (i.e. what I'm seeing here, where the timezone varies depending on where I'm looking)?

albertony:
Thank you for testing!
Does the timestamp and zone information from other commands, e.g. `rclone lsjson`, match what you see in `rclone ncdu`?

rawtaz:
> Does the timestamp and zone information from other commands, e.g. `rclone lsjson`, match what you see in `rclone ncdu`?

It does. And to be more specific, it seems that *blobs/objects* show up with UTC/Zulu while folders/paths show up with CET, see this example that shows the `config` file restic created last night:

~~~
/rclone lsjson mystorageaccountremote:test1/backup/
[
{"Path":"config","Name":"config","Size":155,"MimeType":"application/octet-stream","ModTime":"2022-11-04T01:02:37.000000000Z","IsDir":false,"Tier":"Hot"},
{"Path":"data","Name":"data","Size":-1,"MimeType":"inode/directory","ModTime":"2022-11-04T16:04:05.695614000+01:00","IsDir":true},
{"Path":"index","Name":"index","Size":-1,"MimeType":"inode/directory","ModTime":"2022-11-04T16:04:05.695663000+01:00","IsDir":true},
{"Path":"keys","Name":"keys","Size":-1,"MimeType":"inode/directory","ModTime":"2022-11-04T16:04:05.695674000+01:00","IsDir":true},
{"Path":"locks","Name":"locks","Size":-1,"MimeType":"inode/directory","ModTime":"2022-11-04T16:04:05.695734000+01:00","IsDir":true},
{"Path":"snapshots","Name":"snapshots","Size":-1,"MimeType":"inode/directory","ModTime":"2022-11-04T16:04:05.695747000+01:00","IsDir":true}
]
~~~

albertony:
Thanks again. I will look a bit more into it, but I haven't thought much about the time zone handling in rclone before so I'm hoping @ncw stops by soon and can enlighten us on the subject... 🙏  😃

albertony:
Could it be that the backend does not supply any timestamps for the "directory" entries at all? And what you see on these are simply the current time?

Rclone is, to a large extent, only concerned about "files". Directory time stamps, for instance, is a "best effort" thing - with fallback to current date/time - I *think*...

rawtaz:
> Could it be that the backend does not supply any timestamps for the "directory" entries at all? And what you see on these are simply the current time?

Yes, that seems to be the case. I didn't look at the actual times earlier, but they indeed seem to be the current time of viewing. And yes, the paths are just metadata basically, so not something that has their own timestamp probably. Makes sense.

Your latest patches work fine in that they make the time for blobs/objects appear in the local timezone.

I'm wondering though, wouldn't it make more sense to just print empty space/nothing for non-blobs/-objects? It seems pointless to show the current time, better to not show anything. Also less clutter :)

albertony:
> Your latest patches work fine in that they make the time for blobs/objects appear in the local timezone.

Great, 👍 

> I'm wondering though, wouldn't it make more sense to just print empty space/nothing for non-blobs/-objects? It seems pointless to show the current time, better to not show anything. Also less clutter :)

I see what you mean. Though, the ncdu code does not know this, it just fetches the modtime from a common library object, without knowing the origin of the value there, so it may be a dummy "Now" timestamp or it may be a timestamp from the backend. Ncdu could easily be changed to only show timestamps for files, though, i.e. blank for directories, but I'm not sure that this is what we want in general.. The final decision is up to @ncw 🙈

rawtaz:
But if we can make it a dummy, is it technically possible to make it an empty string, or just a space (" ") or at the very least "-"?

albertony:
Well, its not that easy since the fall-back time.Now value is set deep within shared code, so it would affect all of rclone - not only the ncdu subcommand.

ncw:
> It does. And to be more specific, it seems that _blobs/objects_ show up with UTC/Zulu while folders/paths show up with CET

This is because the folders don't actually exist - rclone syntesises them.

Maybe showing all the times in local time would be the thing to do - that is what rclone lsl does I think? We could then lose the timezone.

albertony:
> that is what rclone lsl does I think

It is. Local (DST-aware) zone. Without specifying the time zone. And with full precision:

```
  3765 2022-08-15 07:36:44.253948486 README.md
   757 2021-08-31 09:56:44.883992925 config.json
```

While this PR as of now shows them identical to upstream ncdu: Same local zone and similar format as lsl - but only down to second precision and including the time zone offset spec:
```
    3.677Ki 2022-08-15 07:36:44 +0200 [          ]  README.md
        757 2021-08-31 09:56:44 +0200 [          ]  config.json  
```

albertony:
Not having the zone specifier as part of a local time string may give ambiguity in results, as I'm sure you're aware of:
```
$ rclone touch f0 -t "2022-10-30T00:00:00.000"
$ rclone touch f1 -t "2022-10-30T01:00:00.000"
$ rclone lsl .
        0 2022-10-30 02:00:00.000000000 f0
        0 2022-10-30 02:00:00.000000000 f1
```
But its the same with native ls (unless maybe if you customizing the time format):
```
$ ls -l
-rw-r--r--  1 aon aon    0 Oct 30 02:00 f0
-rw-r--r--  1 aon aon    0 Oct 30 02:00 f1
```
Keeping the zone specifier, as in this PR code still, removes the ambiguity:
```
$ rclone ncdu .
     0 2022-10-30 02:00:00 +0200 [          ]  f0
     0 2022-10-30 02:00:00 +0100 [          ]  f1
```
But I guess most people prefer aesthetics over precision... 🔥 😆

ncw:
Note that some systems store the time zone and time of objects, so you could get objects with different timezones. And some backends store the timezone as the timezone of the server, so it's all rather inconsistent.

That's why rclone lsl converts everything to local time. In general you want to show local time to users I think.

albertony:
I see, makes sense. Suspected something like that. Explicit conversion to local time before output is a good idea. Note that I had already added such conversion in the ncdu output before my last comment, so my example above was all with local time - but the time zone +01:00 and +0200 indicates if it is in DST or not, and therefore showing it removes that transition-hour ambiguity. That was my main point there.

Anyway, I've removed the zone part from the ncdu format, deviating from the upstream ncdu format a bit, but making it more like the lsl output - except reduced to second precision to save screen width.

```
$ rclone ncdu .
     0 2022-10-30 02:00:00 [          ]  f0
     0 2022-10-30 02:00:00 [          ]  f1
```

ncw (approved):
This is very nicely implemented - thank you.

I appreciate that you don't call ModTime unless it is needed - that is great for performance on s3/swift.

The times look nice now.

PR #6553, opened 2022-11-08
SFTP: Auto-detect shell type for fish

### LINKED ISSUE #6552: Fish shell misdetection
EDIT: This issue seems to come from misdetecting the fish-shell as cmd.

Original issue:
## SFTP remote not working properly with files that have a dollar sign in their filename

When copying a file with a dollar sign in its name to (or from) an SFTP remote, no file gets copied and the md5sum check fails with an empty md5sum. I assume this is because it tries to expand the dollar as a shell variable (the md5sum command in the log replaces "foo $bar" with "foo ").

I am using up-to-date Arch Linux both locally and the remote. This occurs with current rclone `v1.60.0` as well as on `v1.61.0-DEV`.
Originally I assumed this might be due to running fish shell on the server, but it happens with bash as well.

A simple repro:
```
cd ~/rclone-repro
echo "test" >> "abc \$def ghj"
rclone copy abc\ \$def\ ghj thekka:rclone-repro -vv --log-file bug.log
```
Outputs: [bug.log](https://github.com/rclone/rclone/files/9961323/bug.log)

Note that copying via the sftp command directly works without issue in both directions (uploading and downloading), so it's not an issue of the target file system not liking the name.
Downloading the file with rclone after copying via regular sftp results in basically the same error.

albertony:

Fish is different from POSIX-based Unix shells such as bash, and a bracketed variable references like we use for the auto-detection echo command is not supported. The command will return with zero exit code but produce no output on stdout. There is a message on stderr, but we don't log it due to the zero exit code:

> fish: Variables cannot be bracketed. In fish, please use {$ShellId}.

Fixes #6552

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -975,20 +975,24 @@ func NewFsWithConnection(ctx context.Context, f *Fs, name string, root string, m
 			fs.Debugf(f, "Running shell type detection remote command: %s", shellCmd)
 			err = session.Run(shellCmd)
 			_ = session.Close()
+			f.shellType = defaultShellType
 			if err != nil {
-				f.shellType = defaultShellType
 				fs.Debugf(f, "Remote command failed: %v (stdout=%v) (stderr=%v)", err, bytes.TrimSpace(stdout.Bytes()), bytes.TrimSpace(stderr.Bytes()))
 			} else {
 				outBytes := stdout.Bytes()
 				fs.Debugf(f, "Remote command result: %s", outBytes)
 				outString := string(bytes.TrimSpace(stdout.Bytes()))
-				if strings.HasPrefix(outString, "Microsoft.PowerShell") { // If PowerShell: "Microsoft.PowerShell%ComSpec%"
-					f.shellType = "powershell"
-				} else if !strings.HasSuffix(outString, "%ComSpec%") { // If Command Prompt: "${ShellId}C:\WINDOWS\system32\cmd.exe"
-					f.shellType = "cmd"
-				} else { // If Unix: "%ComSpec%"
-					f.shellType = "unix"
-				}
+				if outString != "" {
+					if strings.HasPrefix(outString, "Microsoft.PowerShell") { // PowerShell: "Microsoft.PowerShell%ComSpec%"
+						f.shellType = "powershell"
+					} else if !strings.HasSuffix(outString, "%ComSpec%") { // Command Prompt: "${ShellId}C:\WINDOWS\system32\cmd.exe"
+						// Additional positive test, to avoid misdetection on unpredicted Unix shell variants
+						s := strings.ToLower(outString)
+						if strings.Contains(s, ".exe") || strings.Contains(s, ".com") {
+							f.shellType = "cmd"
+						}
+					} // POSIX-based Unix shell: "%ComSpec%"
+				} // fish Unix shell: ""
 			}
 		}
 		// Save permanently in config to avoid the extra work next time


backend/sftp/sftp.go
```diff
@@ -984,10 +984,10 @@ func NewFsWithConnection(ctx context.Context, f *Fs, name string, root string, m
 				outString := string(bytes.TrimSpace(stdout.Bytes()))
 				if strings.HasPrefix(outString, "Microsoft.PowerShell") { // If PowerShell: "Microsoft.PowerShell%ComSpec%"
 					f.shellType = "powershell"
-				} else if !strings.HasSuffix(outString, "%ComSpec%") { // If Command Prompt: "${ShellId}C:\WINDOWS\system32\cmd.exe"
+				} else if outString != "" && !strings.HasSuffix(outString, "%ComSpec%") { // If Command Prompt: "${ShellId}C:\WINDOWS\system32\cmd.exe"
```
ncw, line 987: It would be better to detect this on a positive feature if at all possible, otherwise the next strange unix shell is going to be detected as a Windows cmd prompt.

How about containing case insensitive ".exe" or ".com" ?

backend/sftp/sftp.go
```diff
@@ -984,10 +984,10 @@ func NewFsWithConnection(ctx context.Context, f *Fs, name string, root string, m
...
 					f.shellType = "cmd"
-				} else { // If Unix: "%ComSpec%"
-					f.shellType = "unix"
+				} else { // If POSIX-based shell: "%ComSpec%", if fish shell: ""
```
ncw, line 989: Do we actually need the `${ShellId}` at all? We don't seem to be using it.

ncw (commented):
Thoughts inline!

backend/sftp/sftp.go
albertony: It is used for this rule:
```go
if strings.HasPrefix(outString, "Microsoft.PowerShell") { // If PowerShell: "Microsoft.PowerShell%ComSpec%"
```

${ShellId} expands to "Microsoft.PowerShell" in PowerShell.

ncw: Ah good! So both parts of the input `${ShellId}` and `%COMSPEC%` are to detect windows shells.

albertony: Correct.

backend/sftp/sftp.go
albertony: > It would be better to detect this on a positive feature if at all possible, otherwise the next strange unix shell is going to be detected as a Windows cmd prompt.

I see what you mean.

> How about containing case insensitive ".exe" or ".com" ?

Maybe, not sure, need to think about that.

ncw: According to the Wikipedia article https://en.wikipedia.org/wiki/COMSPEC it should point to `COMMAND.COM`, `CMD.EXE` or `cmd.exe` - that was the basis of my suggestion. I don't think we'll ever get `.exe` or `.com` from anything other than Windows so it seemed like a nice positive ID to case insensitive search the string for `.exe` or `.com`.

albertony: Sure, I agree it probably is the better solution after all. :)

Not very common, I think, but it can also point to tcc.exe, which is a compatible [alternative](https://jpsoft.com/products/tcc-le.html), and maybe there are others. I found an issue in the VSCode project requesting them to respect the ComSpec variable because the user used this TCC tool. And I also *think* it is possible to include arguments, like `cmd.exe /k`. And then I was thinking perhaps it is possible point ComSpec to some wrapper tool or script like `.vbs`, which in turn may launch cmd or tcc or something "behind the scenes", and then would we have to consider all of the extensions in PATHEXT (.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL)... 🤣  On the other hand the ComSpec variable is not very much used for anything in Windows itself anymore, I think, as there are other settings, e.g. in registry, to configure what is normally interpreted as the default shell, and ComSpec is now more like a de facto standard method for other programs to find the correct path to the shell, so I was starting to think it may not even be set at all... Aaaaaanyway, as you point out, it is probably better to get some strange cmd shell config to be auto-detected as Unix shell than the opposite - considering the amount sftp servers running the different shells, and also the amount of different shells which must be considered Unix shell with its escape rules. So... 😩 Will fix!

albertony: Pushed a change. Did not think very much about performance, as it should normally run only once - except for ad-hoc remotes.. If you have further comments I'm happy to do another iteration before merging.

ncw (approved):
Looks good to me :-)

I'll merge it now.

ncw:
Thank you :-)

PR #6554, opened 2022-11-09
s3: Fix unchecked err in S3 Listv2

Skylion007:
* Fix a bug caught by static analysis where a nil value might accidentally be accessed.

An unchecked error value could cause resp to be nil which would cause issues when it is accessed on the next line. This adds a check around which doesn't reference the resp variable unless the err value is nil

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -3159,8 +3159,11 @@ func (f *Fs) newV2List(req *s3.ListObjectsV2Input) bucketLister {
 // Do a V2 listing
 func (ls *v2List) List(ctx context.Context) (resp *s3.ListObjectsV2Output, versionIDs []*string, err error) {
 	resp, err = ls.f.c.ListObjectsV2WithContext(ctx, &ls.req)
+	if err != nil {
+		return nil, nil, err
+	}
 	ls.req.ContinuationToken = resp.NextContinuationToken
-	return resp, nil, err
+	return resp, nil, nil
 }
 
 // URL Encode the listings


backend/s3/s3.go
```diff
@@ -3159,7 +3159,9 @@ func (f *Fs) newV2List(req *s3.ListObjectsV2Input) bucketLister {
 // Do a V2 listing
 func (ls *v2List) List(ctx context.Context) (resp *s3.ListObjectsV2Output, versionIDs []*string, err error) {
 	resp, err = ls.f.c.ListObjectsV2WithContext(ctx, &ls.req)
-	ls.req.ContinuationToken = resp.NextContinuationToken
+	if err == nil {
```
ncw, line 3162: Can you re-write as an early return to make it look like the other listing routines - something like

```go
if err != nil {
    return nil, nil, err
}
ls.req.ContinuationToken = resp.NextContinuationToken
return resp, nil, nil
```

ncw (commented):
Excellent find - thank you :-)

What piece of static analysis magic did you use?

I made a suggest inline for a reword to make it bring it into the same style as the other listing routines.

Skylion007:
> Excellent find - thank you :-)
> 
> What piece of static analysis magic did you use?
> 
> I made a suggest inline for a reword to make it bring it into the same style as the other listing routines.

I used LGTM: https://lgtm.com/projects/g/rclone/rclone/ which is migrating to Github Actions after being bought by Github: https://github.com/marketplace/lgtm

Skylion007:
@ncw Actually, looks like most of the features have been rolled into Github code scanning:  https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository

ncw (approved):
This code looks great now - thanks.

I'll merge it now.

ncw:
> I used LGTM: https://lgtm.com/projects/g/rclone/rclone/ which is migrating to Github Actions after being bought by Github: https://github.com/marketplace/lgtm

Thanks.

> Actually, looks like most of the features have been rolled into Github code scanning: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository

Ah.

> Tried setting up over here on this PR: #6555

I'll comment on the PR - thank you.

PR #6555, opened 2022-11-09
ci: Add codeql to rclone

Skylion007:
#### Checklist


ncw:
Did this find the error you found with LGTM - I can't see if it if did!

> I used LGTM: https://lgtm.com/projects/g/rclone/rclone/ which is migrating to Github Actions after being bought by Github: https://github.com/marketplace/lgtm

The vulnerability scanning seems to have found an awful lot of false positives!

How do you control it?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ go ]
```
Skylion007, line 25: You can also enable more languages like Python, JS etc... too but that results in a lot of warnings.

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,20 @@
+name: "Rclone config"
+
+query-filters:
+  - exclude:
```
Skylion007, line 4: @ncw You can either globally surpress warnings here or locally supress them with a special comment.

Skylion007:
@ncw I finally got around to fixing this static analysis CI and it did find some new problems in the codebase. There are still a few errors left, but I am unsure about the best way to fix them and would appreciate your guidance on that. If they are false positives, I can just add another exclude rule.

ncw:
Can we make the CodeQL result just a warning for the moment? Its going to take a bit of time to fix the errors it shows and I don't want to break the build.

What do you think?

Skylion007:
> Can we make the CodeQL result just a warning for the moment? Its going to take a bit of time to fix the errors it shows and I don't want to break the build.

@ncw I think only someone with access to the repo settings like yourself can do that: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#defining-the-severities-causing-pull-request-check-failure

You can also configure it so only NEW issues become errors.

ncw:
Now 1.62 is out of the way I'm in a position to merge this.

Can you rebase it and squash it please?

Is there anything else we need to do to it?

Is it going to make a lot of warnings that we need to fix? :-)

Skylion007:
> https://docs.github.com/en/code-security/code-scanning/troubleshooting-sarif-uploads/default-setup-enabled
@ncw Seems like I would need admin access to the repo to customize / dismiss codeql alerts now. :(

albertony:
Answering https://github.com/rclone/rclone/pull/8009#issuecomment-2292269148:

Hm, we *do* run CodeQL already - using GitHub's "Default setup". Scanning on push to master, presenting results in the "Security" page, under "Code scanning" (the LGTM service has been shut down).

I guess the question is if we want it to be part of the PR process, i.e. in addition to the linters and govulncheck we already run? Then we must probably disable the current "Default setup" and add it through GitHub Actions as in this PR.

Could be useful, I guess. However, in my experience it is a bit slow, and looking at the history of the current runs of the default setup it takes from 5 to 15 minutes. And there are a lot of false positives, many of the "won't fix" kind - less important ones that are potentially time consuming to fix. We need to have a strategy for these.

> Can you rebase it and squash it please?

That would be a first/next step. :) And make sure it is using latest version of GitHub actions etc, and configured according to latest recommendations.

Skylion007:
>And there are a lot of false positives, many of the "won't fix" kind - less important ones that are potentially time consuming to fix. We need to have a strategy for these

We can just add suppressions to the false positives.

Skylion007:
>presenting results in the "Security" page, under "Code scanning"

The security fixes are the default of LGTM, but there are a lot of other good code quality lints in the non-default setup.

PR #6561, opened 2022-11-11
Added cache cleaner for directories

darthShadow:

Added a cache cleaner for the VFS directories that runs in a goroutine and exectues after `2 * DirCacheTime`.

N/A

#### Checklist

### DIFF
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -22,21 +22,19 @@ import (
 
 // Dir represents a directory entry
 type Dir struct {
-	vfs   *VFS   // read only
-	inode uint64 // read only: inode number
-	f     fs.Fs  // read only
-
-	mu      sync.RWMutex // protects the following
-	parent  *Dir         // parent, nil for root
-	path    string
-	entry   fs.Directory
-	read    time.Time         // time directory entry last read
-	items   map[string]Node   // directory entries - can be empty but not nil
-	virtual map[string]vState // virtual directory entries - may be nil
-	sys     atomic.Value      // user defined info to be attached here
-
-	modTimeMu sync.Mutex // protects the following
 	modTime   time.Time
+	read      time.Time // time directory entry last read
+	entry     fs.Directory
+	f         fs.Fs             // read only
+	sys       atomic.Value      // user defined info to be attached here
+	virtual   map[string]vState // virtual directory entries - may be nil
+	parent    *Dir              // parent, nil for root
+	items     map[string]Node   // directory entries - can be empty but not nil
+	vfs       *VFS              // read only
+	path      string
+	inode     uint64       // inode number: read only
+	mu        sync.RWMutex // protects the directory attributes
+	modTimeMu sync.Mutex   // protects the modTime
 }
 
 //go:generate stringer -type=vState
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -22,19 +22,21 @@ import (
 
 // Dir represents a directory entry
 type Dir struct {
-	modTime   time.Time
-	read      time.Time // time directory entry last read
-	entry     fs.Directory
-	f         fs.Fs             // read only
-	sys       atomic.Value      // user defined info to be attached here
-	virtual   map[string]vState // virtual directory entries - may be nil
-	parent    *Dir              // parent, nil for root
-	items     map[string]Node   // directory entries - can be empty but not nil
-	vfs       *VFS              // read only
-	path      string
-	inode     uint64       // inode number: read only
-	mu        sync.RWMutex // protects the directory attributes
-	modTimeMu sync.Mutex   // protects the modTime
+	modTime           time.Time
+	read              time.Time // time directory entry last read
+	entry             fs.Directory
+	sys               atomic.Value      // user defined info to be attached here
+	hasVirtual        atomic.Value      // atomic boolean to indicate if the directory has virtual entries
+	f                 fs.Fs             // read only
+	parent            *Dir              // parent, nil for root
+	virtual           map[string]vState // virtual directory entries - may be nil
+	items             map[string]Node   // directory entries - can be empty but not nil
+	vfs               *VFS              // read only
+	cacheCleanupTimer *time.Timer       // timer to call cacheCleanup
+	path              string
+	mu                sync.RWMutex // protects the directory attributes
+	modTimeMu         sync.Mutex   // protects the modTime
+	inode             uint64       // inode number: read only
 }
 
 //go:generate stringer -type=vState
@@ -50,7 +52,7 @@ const (
 )
 
 func newDir(vfs *VFS, f fs.Fs, parent *Dir, fsDir fs.Directory) *Dir {
-	return &Dir{
+	d := &Dir{
 		vfs:     vfs,
 		f:       f,
 		parent:  parent,
@@ -60,15 +62,16 @@ func newDir(vfs *VFS, f fs.Fs, parent *Dir, fsDir fs.Directory) *Dir {
 		inode:   newInode(),
 		items:   make(map[string]Node),
 	}
+	d.cacheCleanupTimer = time.AfterFunc(vfs.Opt.DirCacheTime*2, d.cacheCleanup)
+	d.SetHasVirtual(false)
+	return d
 }
 
 func (d *Dir) cacheCleanup() {
 	defer func() {
+		// We should never panic here
 		_ = recover()
-		return
 	}()
-
-	<-time.After(d.vfs.Opt.DirCacheTime * 2)
 
 	when := time.Now()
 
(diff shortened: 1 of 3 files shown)


darthShadow:
@ncw How does this look? This is an effort to address the high memory usage till we get the metadata DB.

Attached is a screenshot of a server with 8 GB of memory before & after deploying the change:
Screenshot 2022-11-11 at 12 46 44 PM

vfs/dir.go
```diff
@@ -176,8 +197,8 @@ func (d *Dir) Node() Node {
 // so could not be forgotten. Children which didn't have virtual entries and
 // children with virtual entries will be forgotten even if true is returned.
 func (d *Dir) ForgetAll() (hasVirtual bool) {
-	d.mu.Lock()
-	defer d.mu.Unlock()
+	d.mu.RLock()
```
darthShadow, line 200: Does this seem fine? Taking a Write lock was causing lockups for multiple minutes at once.

ncw: Splitting the lock like that potentially makes the `hasVirtual` flag out of date.

Splitting this lock is a good idea though, I often see it in deadlock traces!

Maybe it would be better to keep the hasVirtual up to date continuously when we add and remove virtual entries so we don't have to deal with it here.

So when adding the first virtual entry you'd call hasVirtual(true) on the parent and it in turn would call its parents etc. Maybe the virtual flag could be read by sync/atomic to avoid the locking.

vfs/dir.go
```diff
@@ -463,7 +493,10 @@ func (d *Dir) _readDir() error {
 		return err
 	}
 
+	go d.cacheCleanup()
```
ncw, line 496: I'm not a big fan of running a go routine per directory read - these are going to mount up!

What you should probably do is use 

time.AfterFunc(d Duration, f func()) *Timer

And 
- keep the *Timer in the Dir structure
- kick the timer on every time the Dir is read with Timer.Reset()

ncw (commented):
I think this is a nice idea but the implementation needs a bit of work.

See inline for some suggestions.

The locking changes in ForgetAll if you want to attempt it should probably be in a different commit.

vfs/dir.go
darthShadow: I like the `time.AfterFunc` method but want to verify my understanding once.

With the current implementation, wouldn't the goroutine only be created when the cache is actually stale? Otherwise, it returns from the `_read` function before it can create the goroutine.

darthShadow:
I have done the requested changes, please take a look again.

I did make the `hasVirtual` an atomic.Value but I am not sure I understood this part of your comment:

```
So when adding the first virtual entry you'd call hasVirtual(true) on the parent and it in turn would call its parents etc.
```

Why do we need to call the parent on adding any virtual entries?

vfs/dir.go
darthShadow: Done, thanks.

ncw:
Ach, this patch has fallen off my radar and it is a good one saving lots of memory.

I'll put it on the 1.64 milestone.

In the mean time can you rebase it @darthShadow ?

darthShadow:
@ncw Rebased with the latest changes.

ncw:
I've just been looking at the middle commit `Change field alignment for reducing the struct size from 184 to 136 for dir.go & 192 to 120 for file.go`

By my eyeball your changes shouldn't make any difference (at least on 64 bit architectures). I added this in

```patch
diff --git a/vfs/vfs.go b/vfs/vfs.go
index dc15f2586..7f9a538fd 100644
--- a/vfs/vfs.go
+++ b/vfs/vfs.go
@@ -31,6 +31,7 @@ import (
        "sync"
        "sync/atomic"
        "time"
+       "unsafe"
 
        "github.com/rclone/rclone/fs"
        "github.com/rclone/rclone/fs/cache"
@@ -235,6 +236,9 @@ func New(f fs.Fs, opt *vfscommon.Options) *VFS {
        // This can take some time so do it after the Pin
        vfs.SetCacheMode(vfs.Opt.CacheMode)
 
+       // Debug sizes
+       fs.Infof(nil, "sizes: Dir = %d, File=%d", unsafe.Sizeof(Dir{}), unsafe.Sizeof(File{}))
+
        return vfs
 }
```

To see and I get (on amd64)

```
 after: INFO  : sizes: Dir = 184, File=192
before: INFO  : sizes: Dir = 184, File=192
```

So on my 64 bit machine it doesn't seem to do anything...

On 32 bit i386 it doesn't either...

```
before: sizes: Dir = 128, File=132
after: sizes: Dir = 128, File=132
```

What tool where you using to do this?

----

I removed the middle commit then did a bit of tidying.

I also added another commit on top (which can be removed easily) which does the keeping track of whether a directory has virtuals at the time they are created, incrementing atomically a number in the parent.

I pushed this to the pr-6561-vfs branch if you want to take a look. You can remove the top commit to get back to just your two commits (tidied up a bit by me).

I'm not sure which approach is best - the updating the parents approach could cause deadlocks as it needs to take the dir Rlock briefly to look up the parent which isn't a problem your original approach has. However your original approach still has a point after releasing the Rlock and gaining the Lock when things can get out of sync. Its a small window and its consequences are probably minor...

Anyway your thoughts/testing appreciated!

darthShadow:
> What tool where you using to do this?

https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment

It also considers the optimizations for GC.

Further info:

https://github.com/1pkg/gopium/issues/24#issuecomment-950188558

darthShadow:
I like the new approach but the new virtual object creation time is going to become dependent on the tree depth, which doesn't seem ideal. Perhaps we could update only the current & parent object virtuals synchronously and do the rest of the tree walk/update in a goroutine?

darthShadow:
I have also been thinking about the memory patterns after this change and it doesn't seem ideal to have it be a sawtooth pattern rather than relatively constant (for a relative stable mount) pattern from before the change.

Kopia uses a sync.Pool based implementation for the file & directory objects, so perhaps we should do something similar?
References:
* https://github.com/kopia/kopia/blob/d2ddda7114bd4e3ff26ddb9c52fa2c778b806514/fs/localfs/local_fs_pool.go
* https://github.com/kopia/kopia/blob/d2ddda7114bd4e3ff26ddb9c52fa2c778b806514/internal/freepool/freepool.go

ncw:
> I like the new approach but the new virtual object creation time is going to become dependent on the tree depth, which doesn't seem ideal.

On further reflection I'm not so keep on the new approach! Having to take the parent lock gives loads of opportunity for deadlocks, and as you say it puts the time for this up front which will be noticeable.

> Perhaps we could update only the current & parent object virtuals synchronously and do the rest of the tree walk/update in a goroutine (but that introduces the possibility for a race again)?

I think the probability of a race in your code is very low and the consequences are we miscount the virtual directories which won't be catastrophic so I think we should just use that until we think of something better!

> I have also been thinking about the memory patterns after this change and it doesn't seem ideal to have it be a sawtooth pattern rather than relatively constant (for a relative stable mount) pattern from before the change.
> 
> Kopia uses a sync.Pool based implementation for the file & directory objects, so perhaps we should do something similar? References:
> 
> * https://github.com/kopia/kopia/blob/d2ddda7114bd4e3ff26ddb9c52fa2c778b806514/fs/localfs/local_fs_pool.go
> * https://github.com/kopia/kopia/blob/d2ddda7114bd4e3ff26ddb9c52fa2c778b806514/internal/freepool/freepool.go

We could use a sync pool for the directory and file objects quite easily. That is only part of the problem though - we are also allocating fs.Objects and fs.Directories which I suspect might be the bigger problem. They in turn allocate loads of memory for strings etc.

It would probably be instructive to do a go memory profile on vfs with lots of objects (eg load the whole directory tree) and see how many pools we might need and what we could gain.

The freepool abstraction looks nice. Now that the lowers supported go can use generics we can make a nice generic one easily. Rclone has a pool in in `lib/pool` which doesn't not use `sync.Pool` as it can sometimes behave rather pathologically (as s3 users found!). It can also use `mmap` for allocations which makes it only really useful for buffers as you need to be super careful with pointers to mmap memory.

darthShadow:
I have created a separate issue for adding an object pool to reduce memory usage: https://github.com/rclone/rclone/issues/7170

It should probably come along with this change but I am not sure of the complexity involved.

ncw:
I've merged the first and third commit from your patch series. I don't want to merge the changing struct layout patch until we have evidence that it actually does stuff!

Can you make a new PR with your final patch so we can discuss that. Its difficult to merge PRs if you keep adding stuff to them!

I rebased your last patch for you - here is the diff
 
[vfs.patch.txt](https://github.com/rclone/rclone/files/12158672/vfs.patch.txt)

I'll close this one - please make a new one with the final part - thanks!

PR #6562, opened 2022-11-11
ncdu: use bitmask for some options

eNV25:

The previous version of ncdu, used int8 or bool values to represent
some options. The struct is larger, as a result.

This version replaces those options with a bitmask, which takes less
space.

Some implementation details:

- Sorting options
  The previous version used an int8 to represent 3 possible values -1 (reversed on), 0 (off), 1 (normal on). Each sortBy options used a separate int8.
  This version uses one bit for reverse, and a bit for each sortBy options.

No

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -114,35 +114,48 @@ func helpText() (tr []string) {
 
 // UI contains the state of the user interface
 type UI struct {
-	f                  fs.Fs     // fs being displayed
-	fsName             string    // human name of Fs
-	root               *scan.Dir // root directory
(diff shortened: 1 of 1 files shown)


ncw:
> The previous version of ncdu, used int8 or bool values to represent some options. The struct is larger, as a result.

How much extra memory could be saved?

There is only ever one of these structs in memory at once, so I can't believe it is a noticeable amount.

> This version replaces those options with a bitmask, which takes less space.

I'm not sure the more complicated code justifies the minute space savings.

What do you think?

eNV25:
That's fair as well. If you're not interested, I'll just close this and update the other request.

Issue #6272, opened 2022-06-22
rclone ncdu: Move from Tcell's termbox compatibility layer to native Tcell


albertony:
#### What problem are you are trying to solve?

PR #6258 changed ncdu's implementation from using `termbox` (<https://github.com/nsf/termbox-go>) to `Tcell` (<https://github.com/gdamore/tcell>). It used tcell's compatibility layer for termbox, which kept the required code changes and risk on a minimum. This issue is for the natural next step: Changing to tcell's "native" api. This will prepare the ground for making use of more advanced features of Tcell.

#### How do you think rclone should be changed to solve that?

An example of a small patch that ports some functions to tcell is shown here: https://github.com/rclone/rclone/pull/6258#issuecomment-1159684288

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

eNV25:
This issue can be closed.

albertony:
Ah yes, you are right, thanks. Consider it done! ;)

PR #6564, opened 2022-11-12
ncdu: move to native tcell

eNV25:

Based the commit on #6562.

Use tcell directly instead of the termbox wrapper

Following up on 36add0af, which switched from `termbox`
to `tcell`'s wrapper.

Here are some differences from the previous code.

##### Graphemes are taken into account

The original `termbox` library only supports using one rune (code-point) for a cell. But many Unicode characters are made up of multiple runes (code-points), called grapheme clusters.

With the `tcell` library, you can also set additional runes, which seems to work for graphemes. This code iterates through graphemes through the `uniseg` library, which seems to be an indirect dependency already.

Previous: #6258

Issue:  #6272

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -13,13 +13,14 @@ import (
 	"strings"
 
 	"github.com/atotto/clipboard"
-	"github.com/gdamore/tcell/v2/termbox"
+	"github.com/gdamore/tcell/v2"
 	runewidth "github.com/mattn/go-runewidth"
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/cmd/ncdu/scan"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/fspath"
 	"github.com/rclone/rclone/fs/operations"
+	"github.com/rivo/uniseg"
 	"github.com/spf13/cobra"
 )
 
@@ -114,6 +115,7 @@ func helpText() (tr []string) {
 
 // UI contains the state of the user interface
 type UI struct {
+	s                  tcell.Screen
 	f                  fs.Fs     // fs being displayed
 	fsName             string    // human name of Fs
 	root               *scan.Dir // root directory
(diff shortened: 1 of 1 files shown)


cmd/ncdu/ncdu.go
```diff
@@ -150,77 +152,91 @@ type dirPos struct {
 	offset int
 }
 
+// graphemeWidth returns the number of cells in rs.
+//
+// The original [runewidth.StringWidth] iterates through graphemes
+// and uses this same logic. To avoid iterating through graphemes
+// repeatedly, we separate that out into its own function.
+func graphemeWidth(rs []rune) (wd int) {
+	// copied/adapted from [runewidth.StringWidth]
```
eNV25, line 161: This is something I'm not sure about. Do I properly attribute this code fragment?

The original code is MIT licensed.

eNV25: @ncw

cmd/ncdu/ncdu.go
```diff
@@ -926,39 +933,50 @@ outer:
...
+				if k := ev.Key(); k == tcell.KeyRune {
+					c = ev.Rune()
+				} else {
+					c = key(k)
+				}
+				switch c {
+				case key(tcell.KeyEsc), key(tcell.KeyCtrlC), 'q':
 					if u.showBox {
-						u.showBox = false
+						u.showBox = !u.showBox
```
ncw, line 955: This appears to be a change in operation?

cmd/ncdu/ncdu.go
ncw: I think this is a small enough bit of code that no-one is going to be bothered

ncw (commented):
Thanks for doing this work :-)

How important is the grapheme support? Are there any terminals which actually support graphemes? The ones I tested didn't.

I put a couple of comments inline also.

cmd/ncdu/ncdu.go
eNV25: True. I had fixed it, but I forgot to push.

eNV25:
> How important is the grapheme support? Are there any terminals which actually support graphemes? The ones I tested didn't.

The previous version doesn't work for accented characters made up of two code points. e.g. `A + ◌́ => Á`. This work correctly in this PR.

```
$ touch 'Á'
$ go run . ncdu .
```

But this PR isn't perfect either (take for example this emoji `🐕‍🦺`), but I think it works well enough. I would have to ask tcell's author on how to best handle this.

cmd/ncdu/ncdu.go
```diff
@@ -1001,15 +1022,18 @@ outer:
...
 		}
 		// listen to key presses, etc.
 	}
 	return nil
 }
+
+// key returns a rune representing the key k. It is larger than the maximum Unicode code-point.
+func key(k tcell.Key) rune {
+	// This is the maximum possible Unicode code point. Anything greater fails to compile as a Go quoted rune literal.
+	return '\U0010FFFF' + rune(k)
```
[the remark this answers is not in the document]
ncw: I think it is OK as an innovative solution :-)

ncw (approved):
I think this is ready to merge now, so I will do so now - thank you :-)

PR #6568, opened 2022-11-14
azureblob: support service principals using the EnvironmentCredential scheme

nwf-msr:

Add support for inferring the Azure Service Principal identity and secret from environment variables as per [EnvironmentCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet).   I find myself using `rclone` in a context where `EnvironmentCredential`s are already available for other reasons and it seems silly to go write a JSON file just to use the values therein.

Are there additional tests you'd like to see for this feature?

No, sorry.

#### Checklist

I believe the ticked boxes are correct, but please be gentle if I've missed something.

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -89,6 +89,15 @@ Leave blank normally. Needed only if you want to use a service principal instead
 See ["Create an Azure service principal"](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli) and ["Assign an Azure role for access to blob data"](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-rbac-cli) pages for more details.
 `,
 		}, {
+			Name: "env_auth",
+			Help: `Read credentials from runtime (environment variables).
+
+Pull credentials from AZURE_TENANT_ID and AZURE_CLIENT_{ID,SECRET} environment vars.
+See EnvironmentCredential in the Azure docs for more info.
+
+Other authentication methods will, if specified, override this flag.`,
+			Default: false,
+		}, {
 			Name: "key",
 			Help: "Storage Account Key.\n\nLeave blank to use SAS URL or Emulator.",
 		}, {
@@ -270,6 +279,7 @@ type Options struct {
 	Account              string               `config:"account"`
 	ServicePrincipalFile string               `config:"service_principal_file"`
 	Key                  string               `config:"key"`
+	EnvAuth              bool                 `config:"env_auth"`
 	UseMSI               bool                 `config:"use_msi"`
 	MSIObjectID          string               `config:"msi_object_id"`
 	MSIClientID          string               `config:"msi_client_id"`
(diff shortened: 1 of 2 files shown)


backend/azureblob/azureblob.go
```diff
@@ -88,6 +88,10 @@ Leave blank normally. Needed only if you want to use a service principal instead
 
 See ["Create an Azure service principal"](https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli) and ["Assign an Azure role for access to blob data"](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-rbac-cli) pages for more details.
 `,
+		}, {
+			Name: "use_env_creds",
```
ncw, line 92: Assuming this is the same thing, we call this `env_auth` elsewhere so can you rename it for consistency?

Eg https://rclone.org/s3/#s3-env-auth

backend/azureblob/azureblob.go
```diff
@@ -88,6 +88,10 @@ Leave blank normally. Needed only if you want to use a service principal instead
...
+			Help: `Pull credentials from AZURE_TENANT_ID and AZURE_CLIENT_{ID,SECRET} env vars a la EnvironmentCredential.`,
```
ncw, line 93: What does "env vars a la EnvironmentCredential" mean? I'm not sure what `EnvironmentCredential` is - is that a bit of Azure jargon? I  found the reference - I think we should try to make that a bit clearer.

Note that the first line of this `Help` is used for the help text of the command line option, so I suggest you shorten the first line, then put more explanation in subsequent lines.

Maybe something like this (adapted from the s3 env_auth)

```go
Help: `Read credentials from runtime (environment variables).

Only applies if service_principal_file is blank. ???Maybe we should be checking the other auth too???

Pull credentials from AZURE_TENANT_ID and AZURE_CLIENT_{ID,SECRET} environment vars.

See EnvironmentCredential in the Azure docs for more info.
```

backend/azureblob/azureblob.go
```diff
@@ -468,11 +473,7 @@ const azureActiveDirectoryEndpoint = "https://login.microsoftonline.com/"
 const azureStorageEndpoint = "https://storage.azure.com/"
 
 // newServicePrincipalTokenRefresher takes the client ID and secret, and returns a refresh-able access token.
```
ncw, line 475: Can you update this comment as I think it is out of date now?

backend/azureblob/azureblob.go
```diff
@@ -718,19 +719,32 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
+			loadedCreds, err := os.ReadFile(env.ShellExpand(opt.ServicePrincipalFile))
+			if err != nil {
+				return nil, fmt.Errorf("error opening service principal credentials file: %w", err)
+			}
+
+			if err := json.Unmarshal(loadedCreds, &spCredentials); err != nil {
+				return nil, fmt.Errorf("error parsing credentials from JSON file: %w", err)
+			}
+		} else {
+			spCredentials = servicePrincipalCredentials {
```
ncw, line 740: What happens if any of these env vars are missing? Should we check a minimum set is present?

ncw (commented):
Thanks for doing this - it looks really useful :-)

Questions and comments inline.

backend/azureblob/azureblob.go
nwf-msr: All three must be present.  Right now, the behavior in the presence of a missing environment variable would be identical to having a JSON file that omits that value.  The test suite gives an example:
https://github.com/rclone/rclone/blob/591fc3609aea47441fbd4bc01235291efea9bb4e/backend/azureblob/azureblob_test.go#L62

That may not be the best error message for this case (nor, perhaps, for the case under test), and so I am open to suggestions.

backend/azureblob/azureblob.go
nwf-msr: Thanks for the suggestion; I've tweaked it a bit but hopefully in agreeable ways. :)

backend/azureblob/azureblob.go
ncw: I think that is OK then.

ncw (approved):

I'm considering swapping to the new Azure SDK which does this automatically along with lots of other things so I may change it before 1.61 is released.

ncw:
Can you fixup the go-lint problem before I merge please? It looks like you need to gofmt the file.

ncw:
Thank you :-)

PR #6577, opened 2022-11-18
SMB: Fix issue where spurious dot directory is created

albertony:

Given existing code line in the ensureDirectory function:

```
cn.smbShare.MkdirAll(f.toSambaPath(path.Dir(_path)), 0o755)
```

When uploading a file to the root of the share, _path here is empty string, then path.Dir will return "." (from its implicit call to path.Clean), toSambaPath will encode it to the Unicode wide equivalent, and the result is as described in the forum thread: A directory `．` is being created, in addition to the uploaded file.

I fixed this by checking for the "." and return from ensureDirectory before ever calling MkdirAll - assuming it is not necessary to call with empty string "".

(I've been thinking about the use of path.Clean/Join/Dir etc on remote paths in the backend implementations in general; if it is strictly speaking not the correct thing to do, i.e. the resolving of `.`, `..` and removal of `/`... but that's a topic for later)

https://forum.rclone.org/t/smb-remote-creating-a-dir-named/34188

#### Checklist

### DIFF
--- a/backend/smb/smb.go
+++ b/backend/smb/smb.go
@@ -478,11 +478,15 @@ func (f *Fs) makeEntryRelative(share, _path, relative string, stat os.FileInfo) 
 }
 
 func (f *Fs) ensureDirectory(ctx context.Context, share, _path string) error {
+	dir := path.Dir(_path)
+	if dir == "." {
+		return nil
+	}
 	cn, err := f.getConnection(ctx, share)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


ncw:
I think this looks good.

@Lesmiscore what do you think?

Lesmiscore:
Looks good too

ncw (approved):
Looks good to me and has @Lesmiscore 's blessing so I'll merge this now - thank you :-)

PR #6578, opened 2022-11-19
Improve documentation of backend experimental status

albertony:
Cleanup and making headers in backend docs more consistent.

Apart from borderline OCD whitespace changes, I changed this:

#### Iteration 1

> Chunker (BETA)
> Compress (Experimental)
> Hasher (EXPERIMENTAL)

Into:

> Chunker (EXPERIMENTAL)
> Compress (EXPERIMENTAL)
> Hasher (EXPERIMENTAL)

Assuming there was no intentional distinction between a beta status of Chunker vs experimental status of Compress and Hasher?

#### Iteration 2

Thinking more about it, considering the newly introduced version introduced floating label (discussed here: https://forum.rclone.org/t/smb-cifs-missing-from-list-of-protocols-in-rclone-config/34132/, see https://tip.rclone.org/chunker/ for example), we could also make the experimental status a similar concept:

image

I think it looks better, but perhaps more risk of users not noticing it - at at least on a wide screen where the floating box will be far right? Maybe some red colouring or something makes it stand out more handling both concerns? 

image

Or even more noticeable:

image

That last example is what I have implemented.

#### Checklist


ncw:
Nice idea :-)

Perhaps rather than having

`experimental: true`

we should have

`status: Experimental`

Then we could keep the `Beta` vs `Experimental` distinction. We could also add `Deprecated` to the Cache and Amazon Cloud Files backends.

What do you think?

ncw:
If we can think of a way of defining the tags in a pcloud.Command definition - maybe as the last line of the help - then we can make these come out on the command pages too.

albertony:
> Then we could keep the `Beta` vs `Experimental` distinction.

So, I take it there is a distinction?

> We could also add `Deprecated` to the Cache and Amazon Cloud Files backends.

Yes, sounds like a good idea. Downside is we get the same potential for spelling/casing inconsistency as before, some use "experimental" others "EXPERIMENTAL" etc, but probably we can use bootstrap utility `text-capitalize`, `text-uppercase` or `text-lowercase` or something to avoid that.

Another thing is the tooltip text: It would have to be different for, say, experimental and deprecated statuses. We could have a switch/if-then-else test in the baseof.html code, but maybe it warrants an extra variable, e.g. `statusMessage` - and then maybe rename the first one `statusCode`, or something along those lines..?

```
---
title: "Cache"
description: "Rclone docs for cache remote"
versionIntroduced: "v1.39"
statusCode: "Deprecated"
statusMessage: "The cache backend is due to be phased out in favour of the VFS caching layer eventually which is more tightly integrated into rclone."
---
```

This:

> If we can think of a way of defining the tags in a pcloud.Command definition - maybe as the last line of the help - then we can make these come out on the command pages too.

.. I think I partly understand what you are referring to, but would like to get *some* more explanation.. :)

ncw:
Pehaps using a hugo dict https://gohugo.io/functions/dict/ to define the status lines for the statuses would be best.

Let's try that instead of writing each one in each header.



> > If we can think of a way of defining the tags in a pcloud.Command definition - maybe as the last line of the help - then we can make these come out on the command pages too.
> 
> .. I think I partly understand what you are referring to, but would like to get _some_ more explanation.. :)

Sorry that was a bit cryptic!

My point is is that we auto generate the command docs using `make commanddocs`  which runs `rclone gendocs` and these end up in docs/content/commands/

We could change `rclone gendocs` to add the tags to the markdown.

But where do we get the data from? It will have to live in a [cobra.Command](https://pkg.go.dev/github.com/spf13/cobra#Command) definition.

It looks like they've thought of that already with this

```go
        // Annotations are key/value pairs that can be used by applications to identify or
	// group commands.
	Annotations map[[string](https://pkg.go.dev/builtin#string)][string](https://pkg.go.dev/builtin#string)
```

So perhaps as simple as `rclone gendocs` mapping the `Annotations` directly into headers in the markdown file.

So we'd define

```go
Annotations: map[string]string{
    "status": "Deprecated",
    "minimumVersion": "v1.33",
}
```

And `rclone gendocs` would write those verbatim into the markdown header.

To implement this we'd just have to add a bit to this template I think


https://github.com/rclone/rclone/blob/46b080c0922fc7e811fa53547b2d8d0b2e71383e/cmd/gendocs/gendocs.go#L36

We could then label up commands with minimum versions and status. For instance we deprecated the `rclone dropboxsum` command at one point.

albertony:
Thank you for the walkthrough, very instructive.

I have pushed a new version with the suggested improvements/additions. All pieces are there, I think, but still it is a bit work in progress. I will take a fresh look at it some time soon, but any suggestions before that are also welcome, of course.

- Dictionary for status badge tooltips
   - The dict is currenty only a local variable in the base layout (baseof.html), maybe there is a more correct hugo-way of putting this into a "partials" file or something like that?
   - This means the text will be tailored to the different statuses (deprecated, experimental, beta etc).
   - But it also means it will be identical for all badges showing the same status. E.g. they will be the same for backends and for commands (see below), and there will not be a tooltip explaining more specifically why for instance the cache backend is deprecated, such as "The cache backend is due to be phased out in favour of the VFS caching layer eventually which is more tightly integrated into rclone.". Would be easy enough to add another property to support an optional detailed status description that could be used instead of, or in addition to, the default text from the dict, but perhaps unnecessary complex/code-bloat, and we can leave such details being described in the actual documentation instead...
- Version and status badge support also for commands.
  - Both are optional, none of the backends currently have them set.
  - To add one or both of them, include something like this in the cobra command definition:
    ```go
    Annotations: map[string]string{
      "versionIntroduced": "v1.33",
      "status":            "Deprecated",
    },
    ```
- Made the status text all lowercase instead of capitalized as in my initial version with example shown in original comment. Not 100% decided what I prefer, so don't be surprised if I change it back in a day or two... 🤒  
image

albertony:
Pushed a commit adding "Annotations" with "versionIntroduced" to most commands.

TLDR;
Found the version numbers to set partly from modifying the backend-versions.sh script - to work for the cmd directory and to only print the results without modifying files directly (not so easy to update directly with the command docs as with the backend docs). Then I manually went through the changelog and also manually digged into the git history. In v1.33 there was a restructuring that moved commands into separate files, which makes the digging a bit challenging, and e.g. the "rclone version" command seems to have silently appeared during this refactoring while before it was simply an argument -v, --version etc. For the most original commands such as ls and mkdir I've just set v0.90, which is the "big bang" time for rclone, it seems. One issue was the "config file" and "config show" commands which according to changelog were added in v1.38 but then the parent command "config" were added in next version, v1.39, and file/show moved as subcommands and completed with other subcommands, so the versioning of command and subcommands are a bit of a mix. There are probably more such cases, so impossible to get it 100% accurate, I guess. Anyway, I think I have managed to get it "close enough"..?

cmd/gendocs/gendocs.go
```diff
@@ -25,19 +25,27 @@ func init() {
...
+	VersionIntroduced string
+	Status            string
 }
 
 var frontmatterTemplate = template.Must(template.New("frontmatter").Parse(`---
 title: "{{ .Title }}"
 description: "{{ .Description }}"
 slug: {{ .Slug }}
 url: {{ .URL }}
+{{- with .VersionIntroduced }}
```
ncw, line 43: I was thinking you could pass the Annotations map in here directly and just iterate it

Something like

```
{{ range $key, $value := . Annotations }}
{{ $key }}: {{  $value }}
{{ end }}
```

ncw:
> I have pushed a new version with the suggested improvements/additions. All pieces are there, I think, but still it is a bit work in progress. I will take a fresh look at it some time soon, but any suggestions before that are also welcome, of course.

That looks great!
 
> * Dictionary for status badge tooltips
>   
>   * The dict is currenty only a local variable in the base layout (baseof.html), maybe there is a more correct hugo-way of putting this into a "partials" file or something like that?

Its near where it is used so I think it is OK! There might be a better Hugo way putting it into a data blob, but I suggest we merge it like this until we decide it needs to change.

>   * This means the text will be tailored to the different statuses (deprecated, experimental, beta etc).

:+1:

>   * But it also means it will be identical for all badges showing the same status. E.g. they will be the same for backends and for commands (see below), and there will not be a tooltip explaining more specifically why for instance the cache backend is deprecated, such as "The cache backend is due to be phased out in favour of the VFS caching layer eventually which is more tightly integrated into rclone.". Would be easy enough to add another property to support an optional detailed status description that could be used instead of, or in addition to, the default text from the dict, but perhaps unnecessary complex/code-bloat, and we can leave such details being described in the actual documentation instead...

On a bit of reflection, I think describing this in the docs is the best solution as it will get indexed by google etc, so I'm happy with the generic messages.

> * Version and status badge support also for commands.

Nice!

> * Made the status text all lowercase instead of capitalized as in my initial version with example shown in original comment. Not 100% decided what I prefer, so don't be surprised if I change it back in a day or two... 

:-) Happy to go with whatever you think.

> Pushed a commit adding "Annotations" with "versionIntroduced" to most commands.

Very nice - well done :-)
 
> TLDR; Found the version numbers to set partly from modifying the backend-versions.sh script - to work for the cmd directory and to only print the results without modifying files directly (not so easy to update directly with the command docs as with the backend docs). Then I manually went through the changelog and also manually digged into the git history.

Good work.

> In v1.33 there was a restructuring that moved commands into separate files, which makes the digging a bit challenging, and e.g. the "rclone version" command seems to have silently appeared during this refactoring while before it was simply an argument -v, --version etc.

Ha!

Yes at some point rclone adopted cobra so before that the history will definitely be muddled. And there have been lots of code re-orgs, but I think the changelog should probably be the final decider.

>For the most original commands such as ls and mkdir I've just set v0.90, which is the "big bang" time for rclone, it seems.

I was thinking about this with the backend versions. Though its fun to see those v0.9x versions, maybe we should just remove all versions before v1.00 (backends and commands) as rclone was still in beta then and the dating of most of them is a bit suspect. What do you think? Sensible, or losing too much history?

> One issue was the "config file" and "config show" commands which according to changelog were added in v1.38 but then the parent command "config" were added in next version, v1.39, and file/show moved as subcommands and completed with other subcommands, so the versioning of command and subcommands are a bit of a mix. There are probably more such cases, so impossible to get it 100% accurate, I guess. Anyway, I think I have managed to get it "close enough"..?

Close enough is fine I think!

ncw (commented):
See comment inline.

cmd/gendocs/gendocs.go
albertony: Ah yes, that is much more flexible. Pushed a change to this strategy.

albertony:
> I was thinking about this with the backend versions. Though its fun to see those v0.9x versions, maybe we should just remove all versions before v1.00 (backends and commands) as rclone was still in beta then and the dating of most of them is a bit suspect. What do you think? Sensible, or losing too much history?

I was thinking something similar as well. And I guess even anything older than, say, v1.40 (2018-03-19) is not strictly necessary to label as there are (hopefully) no users on them anyway? But I personally find the badge unobtrusive enough that I think it's kinda nice to just have it there. At least for backends, where it is more of a limited set of doc pages and it appears nice and complete if all have the badge. Commands are a bit more of a mix, and it may not make that much sense to label each and every command.. So maybe keep version labels on all backends but remove the "v0.9x" ones from commands? But no strong opinion, I can definitely remove them also for backends if you think that is more sensible...

The ones currently labelled v0.9x are:
- Backends: Drive, Local, S3 and Swift.
- Commands: check, copy, ls, lsd, mkdir, purge, rmdir, sync and rclone (root/help command) 

> See comment inline.

Force-pushed an amend.

ncw:
> > I was thinking about this with the backend versions. Though its fun to see those v0.9x versions, maybe we should just remove all versions before v1.00 (backends and commands) as rclone was still in beta then and the dating of most of them is a bit suspect. What do you think? Sensible, or losing too much history?
> 
> I was thinking something similar as well. And I guess even anything older than, say, v1.40 (2018-03-19) is not strictly necessary to label as there are (hopefully) no users on them anyway? But I personally find the badge unobtrusive enough that I think it's kinda nice to just have it there. At least for backends, where it is more of a limited set of doc pages and it appears nice and complete if all have the badge. Commands are a bit more of a mix, and it may not make that much sense to label each and every command.. So maybe keep version labels on all backends but remove the "v0.9x" ones from commands?

Yes that sounds like a nice compromise. The command history is definitely a bit more mixed up as early rclone went through quite a few revisions of commands!

> The ones currently labelled v0.9x are:
> 
> * Backends: Drive, Local, S3 and Swift.
> * Commands: check, copy, ls, lsd, mkdir, purge, rmdir, sync and rclone (root/help command)

I'd forgotten check was one of the first batch of commands! I remember move coming along later as one of those aha moments when I realised sync/copy and move were all basically the same thing.

ncw (approved):
Nice - LGTM :-)

Please merge when you are happy!

albertony:
@ncw I don't see any of the changes in tip.rclone.org yet, isn't it updated daily or something like that?

ncw:
@albertony Turns out that the version of hugo on the integration test server was very old and didn't understand the new stuff!

I upgraded it to

hugo v0.107.0-2221b5b30a285d01220a26a82305906ad3291880 linux/amd64 BuildDate=2022-11-24T13:59:45Z VendorInfo=gohugoio

If you email your public ssh key to nick@craig-wood.com I'll give you access to the integration test server so you can see how it all works!

PR #6584, opened 2022-11-26
cmd/ncdu: use negative values for key runes

eNV25:

The previous version used values after the maximum Unicode code-point
to encode a key. This could lead to an overflow since a key is a int16,
a rune is int32 and the maximum Unicode code-point is larger than int16.

A better solution is to simply use negative runes for keys.

PR #6564

#### Checklist


ncw (approved):
LGTM - thank you :-)

PR #6591, opened 2022-11-30
Add global flag '--color'

kverstae:

This PR adds a new global flag `--terminal-color-mode` which allows you to change the way ANSI codes are handled in output. This is similar to the `-color` flag for many UNIX commands (grep, ls...).

Currently, rclone only adds ANSI codes to output if target file descriptor is an actual terminal. With this new flag, the behavior can be changed by using one of the following modes:

- `AUTO` (default) = current behaviour => only add ANSI codes when the output medium is an actual terminal
- `ALWAYS` => always add ANSI codes, even to normal files
- `NEVER` => never add ANSI codes

I am still not sure if this is the best name for the flag. Maybe just `--terminal-color` or `--color` is better?

This has been discussed by someone else on the forum 2 years ago:
https://forum.rclone.org/t/specifying-preservation-of-ansi-codes-in-rclone-output/17305

#### Checklist

**Remark**: no tests for the terminal part itself, since I am not sure how to test this...

### DIFF
--- a/cmd/tree/tree.go
+++ b/cmd/tree/tree.go
@@ -61,7 +61,6 @@ func init() {
 	flags.StringVarP(cmdFlags, &sort, "sort", "", "", "Select sort: name,version,size,mtime,ctime")
 	// Graphics
 	flags.BoolVarP(cmdFlags, &opts.NoIndent, "noindent", "", false, "Don't print indentation lines")
-	flags.BoolVarP(cmdFlags, &opts.Colorize, "color", "C", false, "Turn colorization on always")
 }
 
 var commandDefinition = &cobra.Command{
@@ -113,6 +112,7 @@ For a more interactive navigation of the remote see the
 		opts.SizeSort = sort == "size"
 		ci := fs.GetConfig(context.Background())
 		opts.UnitSize = ci.HumanReadable
+		opts.Colorize = ci.TerminalColorMode != fs.TerminalColorModeNever
 		if opts.DeepLevel == 0 {
 			opts.DeepLevel = ci.MaxDepth
 		}
--- a/docs/content/commands/rclone_tree.md
+++ b/docs/content/commands/rclone_tree.md
@@ -48,7 +48,6 @@ rclone tree remote:path [flags]
 
 ```
   -a, --all             All files are listed (list . files too)
-  -C, --color           Turn colorization on always
   -d, --dirs-only       List directories only
       --dirsfirst       List directories before files (-U disables)
       --full-path       Print the full path prefix for each file
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -823,6 +823,16 @@ quicker than without the `--checksum` flag.
 When using this flag, rclone won't update mtimes of remote files if
 they are incorrect as it would normally.
 
+### --color WHEN ###
+
+Specifiy when colors (and other ANSI codes) should be added to the output.
+
+`AUTO` (default) only allows ANSI codes when the output is a terminal
+
+`NEVER` never allow ANSI codes
+
+`ALWAYS` always add ANSI codes, regardless of the output format (terminal or file)
+
 ### --compare-dest=DIR ###
 
 When using `sync`, `copy` or `move` DIR is checked in addition to the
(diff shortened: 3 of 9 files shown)


ncw:
This looks like a great piece of work :-)

> I am still not sure if this is the best name for the flag. Maybe just `--terminal-color` or `--color` is better?

In general rclone tries to follow usage of other command line tools.

Can you do a quick survey and see what the most popular usage is?

For example this is `grep`

```
       --color[=WHEN], --colour[=WHEN]
              Surround   the  matched  (non-empty)  strings,  matching  lines,
              context lines, file  names,  line  numbers,  byte  offsets,  and
              separators  (for fields and groups of context lines) with escape
              sequences to display them in color on the terminal.  The  colors
              are  defined  by  the  environment  variable  GREP_COLORS.   The
              deprecated environment variable GREP_COLOR is  still  supported,
              but  its setting does not have priority.  WHEN is never, always,
              or auto.
```

Maybe we should be implementing the env var `RCLONE_COLORS` - though a command flag `--color` will automatically give us `RCLONE_COLOR`.

?

kverstae:
Great idea on the survey! What would be the best platform to reach the rclone userbase?

> Maybe we should be implementing the env var RCLONE_COLORS - though a command flag --color will automatically give us RCLONE_COLOR

If you want rclone to be more in line with other cli tools, i think `RCLONE_COLORS` makes the most sense, but this is then not in line with all other global flags in rclone if we would use `--color`. Or we could allow both `RCLONE_COLOR` and `RCLONE_COLORS`, giving priority to the plural form like grep, ls... do?
Since this new flag not only affects colors, but basically every ANSI code (so also cursor movements, which are used in the progress bar). Another option is to call this new flag `--ansi`...

I have also been thinking of adding more fine grained control to this flag. What I mean by this is giving users the option to keep cursor movement, but disable color. Or no cursor movement but still colors... I don't know if this goes too far and it will probably be a bit harder to implement...

ncw:
> Great idea on the survey! What would be the best platform to reach the rclone userbase?

I was thinking more of surveying the existing command line tools which have a `--color` flag or similar rather than the users.
 
> > Maybe we should be implementing the env var RCLONE_COLORS - though a command flag --color will automatically give us RCLONE_COLOR
> 
> If you want rclone to be more in line with other cli tools, i think `RCLONE_COLORS` makes the most sense, but this is then not in line with all other global flags in rclone if we would use `--color`. Or we could allow both `RCLONE_COLOR` and `RCLONE_COLORS`, giving priority to the plural form like grep, ls... do? Since this new flag not only affects colors, but basically every ANSI code (so also cursor movements, which are used in the progress bar). Another option is to call this new flag `--ansi`...

I expect it would be easy to get rclone to respond to `RCLONE_COLOR` and `RCLONE_COLORS` then that at least keeps it consistent with the other flags **and** external usage!

> I have also been thinking of adding more fine grained control to this flag. What I mean by this is giving users the option to keep cursor movement, but disable color. Or no cursor movement but still colors... I don't know if this goes too far and it will probably be a bit harder to implement...

As long as you can make the options backwards compatible then this can be added later.

Are more options needed? What do other programs do?

ncw:
Thank you for your research.

> I think `--color` makes the most sense, with the options auto, never and always (these are already implemented). This will give use the env variable `RCLONE_COLOR`. I wouldn't add `RCLONE_COLORS` after finding out what it actually does for the unix commands. Maybe later if we want to give users control over colors...

OK lets do that. If you want to update this PR to implement that, I'll merge it - thank you :-)

> I can't really think of many other tools that have colors and also cursor movement through ANSI....
> I just feel that with the 3 options (auto, never, always), you are somewhat limited. Disabling colors, would in the current implementation also break (or at least stop displaying it nicely) the progress bar.
> I am happy to keep it as is and see in the future if we want/can extend the control over ANSI.

Let's get this PR merged, then you can send a followup PR implementing some more options if you think what we have is too limiting.

kverstae:
I just discovered that `rclone tree` already has a `--color` flag to enable coloring in the tree.

Would you mind if I removed this flag and made it work with the new global `--color` flag?
If so, what should the default behaviour be? Currently it only shows colors when the flag `--color` is passed. This is not in line with the unix `tree` command, which always shows colors if the output destination is a terminal.

ncw:
> I just discovered that `rclone tree` already has a `--color` flag to enable coloring in the tree.

Yes that is fine.

It is a slight backwards incompatibility.

> Would you mind if I removed this flag and made it work with the new global `--color` flag? If so, what should the default behaviour be? Currently it only shows colors when the flag `--color` is passed. This is not in line with the unix `tree` command, which always shows colors if the output destination is a terminal.

I'd say it should obey `--colors` precisely

```
AUTO (default) = current behaviour => only add ANSI codes when the output medium is an actual terminal
ALWAYS => always add ANSI codes, even to normal files
NEVER => never add ANSI codes
```

This will mean that `rclone tree` starts making colors which is fine.

kverstae:
This has already been added in 0d4f5b1c35fc586fbecc583255e853639dcdc607 and indeed makes `rclone tree` a bit nicer with colors by default.

ncw (approved):
This looks ready to merge now :-) So I will do just that now - thank you!

Please open a new PR for any followups.

Thank you.

PR #6597, opened 2022-12-03
Clarify the docs of remote control - operations/copyfile

yuudi:

The document need to be clarified. I have been struggling for this and find an answer from [this thread](https://forum.rclone.org/t/rclone-rc-operations-copyfile-copy-local-file-to-remote-fail/15608/3). This pull request is to help others avoid the same problem.

It is discussed in the forum 2 years ago <https://forum.rclone.org/t/rclone-rc-operations-copyfile-copy-local-file-to-remote-fail/15608/3>

#### Checklist

### DIFF
--- a/fs/operations/rc.go
+++ b/fs/operations/rc.go
@@ -170,9 +170,9 @@ func init() {
 			Title: name + " a file from source remote to destination remote",
 			Help: `This takes the following parameters:
 
-- srcFs - a remote name string e.g. "drive:" for the source
+- srcFs - a remote name string e.g. "drive:" for the source, "/" for local filesystem
 - srcRemote - a path within that remote e.g. "file.txt" for the source
-- dstFs - a remote name string e.g. "drive2:" for the destination
+- dstFs - a remote name string e.g. "drive2:" for the destination, "/" for local filesystem
 - dstRemote - a path within that remote e.g. "file2.txt" for the destination
 `,
 		})


ncw:
This particular phrase is repeated many many times throughout the docs.

I'd prefer to only write it once - maybe a section near the top explaining what `srcFs` etc mean might be best? What do you think?

yuudi:
> This particular phrase is repeated many many times throughout the docs.
> 
> I'd prefer to only write it once - maybe a section near the top explaining what `srcFs` etc mean might be best? What do you think?

It is not good idea to repeat I will going to modify this. Since people may not notice the section on somewhere else, I am thinking to put annotation subscript to every `srcFs` to help people find the section of explanation. I want to know is it OK for the unity of style of the document?

I am going to modify the pull request soon.

yuudi:
> This particular phrase is repeated many many times throughout the docs.
> 
> I'd prefer to only write it once - maybe a section near the top explaining what `srcFs` etc mean might be best? What do you think?

Thank you for reviewing.

I have noticed that this document is auto-generated from the help string in the source code, and only this `moveOrCopyFile` need to be clarify. Other functions with `srcFs` does not need to be clarify because that includes the whole path. Only `moveOrCopyFile` use separated `srcFs` and `srcRemote`.

I have edited that file and updated the pull request.

ncw (approved):
That looks great now - thanks.

I should have spotted that it was in auto generated docs.

I'll merge now :-)

ncw:
Whoops, forgot to merge this!

PR #6610, opened 2022-12-07
docs: workaround for onedrive shared with me files

vanplus:


ncw:
Nice - thank you. This only works for onedrive personal is that right? Should we mention that?

ncw:
I'll merge this now. It mentions go to live.onedrive.com which hopefully should be enough of a hint.

Thank you.

PR #6621, opened 2022-12-11
vfs: fix reload: failed to add virtual dir entry: file does not exist…

moduller:
\


ncw:
No need to send this as a Pull request. Its is in a branch already. It needs some tests before merging which I haven't had time to write yet but hope to for 1.61.

Issue #4261, opened 2020-05-20
Disallow remotes beginning with dash in a backwards compatible way


jtagcat:
#### What is the problem you are having with rclone?
Can't use remotes beginning with the character `-`.
 
#### What is your rclone version (output from `rclone version`)
```
rclone v1.51.0
- os/arch: linux/amd64
- go version: go1.13.7
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Fedora 32

####  Which cloud storage system are you using? (eg Google Drive)
Any

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
`rclone lsf -my-remote-:` or `rclone lsf "-my-remote-:"`

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)
```
Error: unknown shorthand flag: 'm' in -my-remote-:
Usage:
  rclone lsf remote:path [flags]

Flags:
      --absolute           Put a leading / in front of path names.
      --csv                Output in CSV format.
  -d, --dir-slash          Append a slash to directory names. (default true)
      --dirs-only          Only list directories.
      --files-only         Only list files.
  -F, --format string      Output format - see  help for details (default "p")
      --hash h             Use this hash when h is used in the format MD5|SHA-1|DropboxHash (default "MD5")
  -h, --help               help for lsf
  -R, --recursive          Recurse into the listing.
  -s, --separator string   Separator for the items in the format. (default ";")

Use "rclone [command] --help" for more information about a command.
Use "rclone help flags" for to see the global flags.
Use "rclone help backends" for a list of supported services.

2020/05/20 22:46:47 Fatal error: unknown shorthand flag: 'm' in -my-remote-:
```

animosity22:
Why do you want to start a remote name with a "-" out of curiousity?

animosity22:
Also sorry as the work around is using -- at the end to denote you are done passing options:

```
felix@gemini:/opt/rclone$ rclone lsd -- -gcrypt:
          -1 2018-06-17 10:24:02        -1 Movies
          -1 2017-04-18 16:14:26        -1 TV
felix@gemini:/opt/rclone$
```

as an example.

jtagcat:
> Why do you want to start a remote name with a "-" out of curiosity?

For sorting a long config:
[...]

PR #6623, opened 2022-12-12
Remote name fix

### LINKED ISSUE #6618: Can't add UTF-8 characters to alias names
Hello, 
Happy we have 
`rclone backend drives -o config gdrive:`
for list all shared drive for config file, but my shared drive have utf-8 char, and with large folder with char "_" very hard to read.
To much char "_" to very confuse, 

I have > 1000 shared drive, so it very confuse, hope a people help do that
folder rclone support utf-8 very great

image

Thank you very much

albertony:

Main purpose is to ensure `rclone backend drives` command produces valid remote names. To fix #6618 we recently opened for Unicode characters, but there was one remaining issue when shared drive had space at end.

To fix this I refactored the regex that replaces character sequences invalid in remote names - from the drive backend code, and into the fspath code - where it is together with the main regex for validating config/remote names. Less risk of getting out of sync.

In addition, I refactored the existing check making remote names with leading `-` invalid, which was done as part of #4261. The existing check was done in `CheckConfigName` only, it was not considered by `checkRemoteName`. I'm pretty sure this has to be a small mistake? I fixed this by moving the check for leading `-` into the regex instead of being a separate check.

#6618

#### Checklist


fs/fspath/path_test.go
```diff
@@ -19,34 +19,38 @@ var (
 
 func TestCheckConfigName(t *testing.T) {
```
ncw, line 20: You appear to have overwritten this test instead of making a new one?

albertony: Yes, I extended the `TestCheckConfigName` test by simply adding a second "result" value and using the same input to test both `CheckConfigName` and `MakeConfigName` (should probably have renamed the test). This to avoid another "currently-identical-soon-just-almost-identical" copy of all the input strings - like there to some degree already is in the separate `TestCheckRemoteName`. But I could split it into a separate test, duplicating the same input, if you think that is better, that is completely fine as well..?

ncw (approved):
I think this looks great, please merge when you are ready.

PR #6642, opened 2022-12-20
docs: show only significant parts of version number in version introduced label

albertony:

With current version of official docs, after the 1.61 release, there is a bit of a mix in the version number formats used in the newly introduced version badges. E.g. the Box backend has "v1.38" (two parts) while 1Fichier has "v1.49.0" (three parts). This PR makes them consistent by removing the trailing ".0" parts. My thinking is that it looks a bit better this way, that the looks of the full "v1.49.0" label is a tiny bit more on the technical side, if you see what I mean...? But if you have a strong opinion the other way, that's fine too, my primary goal is consistency here. (Sorry for not noticing before the release, really should have seen this in tip.rclone.org). 

#### Checklist

### DIFF
--- a/docs/content/bisync.md
+++ b/docs/content/bisync.md
@@ -1,7 +1,7 @@
 ---
 title: "Bisync"
 description: "Bidirectional cloud sync solution in rclone"
-versionIntroduced: "v1.58.0"
+versionIntroduced: "v1.58"
 ---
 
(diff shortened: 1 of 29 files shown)


ncw (approved):
I think that is a good idea. We don't add new features in point releases so they should always say `.0`

PR #6663, opened 2023-01-02
Port a fast C-implemented quickXorHash

LXY1226:
speed up `onedrive Bussiness` specified hash algo by port the algo from a fast C-implementation from https://github.com/namazso/QuickXorHash
It was slower than a single HDD in my nas(N5105) so i done this

```
# goos: darwin
# goarch: arm64
# model: Apple M1 Pro 10-core
# Bench for every 1<<20 (1MB)
name             old time/op   new time/op    delta
QuickXorHash-10   2.01ms ± 1%    0.34ms ± 1%   -83.07%  (p=0.000 n=10+9)

name             old speed     new speed      delta
QuickXorHash-10  521MB/s ± 1%  3077MB/s ± 1%  +490.83%  (p=0.000 n=10+9)
```

Thanks for @namazso

Nope. 

#### Checklist


LXY1226:
Added another enhancement at [f440be0](https://github.com/rclone/rclone/pull/6663/commits/f440be03d235ca68b83a0702976698b495c4c207) and all tests passed but a bit of `unsafe` used 
```
goos: linux
goarch: amd64
pkg: od_down/quickxorhash
cpu: Intel(R) Celeron(R) N5105 @ 2.00GHz
name            old time/op   new time/op    delta
QuickXorHash-4   2.39ms ± 3%    0.13ms ± 1%    -94.61%  (p=0.000 n=9+9)

name            old speed     new speed      delta
QuickXorHash-4  438MB/s ± 3%  8125MB/s ± 1%  +1754.62%  (p=0.000 n=9+9)
```

namazso:
No idea how Go works, but my original C version relies on the compiler to vectorize and maybe unroll the block xors. Is the Go compiler bad at that? Maybe there's some hinting macros?
The most compatible fastest way on x86 would be with SSE, doing 128 bits at a time. This is also what clang does: https://godbolt.org/z/5GePfhzG5

LXY1226:
> reply https://github.com/rclone/rclone/pull/6663#issuecomment-1369460462

hmm yes, gc (go compiler) is just make that loop xor in bytes, so i [manual adjust](https://github.com/rclone/rclone/pull/6663/commits/f440be03d235ca68b83a0702976698b495c4c207) it to 8-bytes for a balance between speed and portability. most host supports 64-bit operations and gc would split it to 32-bit if necessary
and the SIMD version of xoring could be implemented by assembly only but losing portability between hosts... 
and that is fast enough! thanks for codes again @namazso

backend/onedrive/quickxorhash/quickxorhash.go
```diff
@@ -199,5 +161,6 @@ func (q *quickXorHash) BlockSize() int {
 func Sum(data []byte) [Size]byte {
 	var d quickXorHash
 	_, _ = d.Write(data)
-	return d.checkSum()
+	sum := d.checkSum()
+	return *(*[Size]byte)(unsafe.Pointer(&sum))
```
ncw, line 165: Can you remove the use of unsafe? I've tried to avoid it in rclone code.

I don't think it will affect performance since it is only called to read the output.

backend/onedrive/quickxorhash/quickxorhash.go
```diff
@@ -7,34 +7,15 @@
...
-//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-//  THE SOFTWARE.
-// ------------------------------------------------------------------------------
+// This code was ported from a fast C-implementation from
+// https://github.com/namazso/QuickXorHash
+// which has licenced as BSD Zero Clause License
```
ncw, line 12: Can you include the full LICENSE text here please

ncw (commented):
This looks great - thank you.

I put a couple of comments inline.

Did you write some benchmarks for this? Maybe you could include those too?

LXY1226:
@ncw Thanks for review this!
Benchmark and some Fuzz test added.
all suggestions are accepted.

could you help me review again

ncw (approved):

I'll merge when the CI goes green.

ncw:
Looks like this needs a bit of fixing up (some imports in the test code I think).

LXY1226:
> Looks like this needs a bit of fixing up (some imports in the test code I think).

fixed (i think)  
im writing another onedrive operating tool so that tests is pasted here.. sorry about that

backend/onedrive/quickxorhash/subtle/xor.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 The Go Authors. All rights reserved.
```
ncw, line 1: I'd prefer not to import this source into rclone.

My preference would be to use build tags to make it compile before and after go1.20, so when go1.20 is released, this will get a speed boost.

LXY1226: okay, i ll remove that :)  
once go 1.20 release, i’d like to re-open this pr to track it.

backend/onedrive/quickxorhash/quickxorhash.go
```diff
@@ -7,58 +7,88 @@
...
+	size uint64
 }
 
 // New returns a new hash.Hash computing the quickXorHash checksum.
 func New() hash.Hash {
 	return &quickXorHash{}
 }
 
+func bytes2U64(v []byte) (lv []uint64) {
+	bh := (*reflect.SliceHeader)(unsafe.Pointer(&v))
```
ncw, line 56: Can you remove these `unsafe` uses?

ncw (commented):
Yes, let's merge this then if you want to send another PR with go1.20 build tags then that would be great as go1.20 is due to be released very soon.

Rclone policy is to support current go, plus 2 older releases, so go1.19, go1.18 and go1.17 now but when go1.20 is released go1.20, go1.19, go1.18

So I'd be happy to take the go1.20 tags immediately.

In fact go1.20rc1 has been released so in the next day or so I'll make a go1.20 build target.

backend/onedrive/quickxorhash/quickxorhash.go
LXY1226: unfortunately, that is the smallest solution to prevent usage of unsafe
conversion of []byte to []uint64 in place is required, otherwise compiler would calculate a byte per cycle or unnecessary memory copy in performance critical area

TheSnowfield (approved):
LGTM

backend/onedrive/quickxorhash/quickxorhash.go
ncw: In your original commit https://github.com/rclone/rclone/pull/6663/commits/f1897c2070f130d98f9fe3893f9f0b4f5501f099 you didn't have this use of unsafe. Can you put it back like that?

namazso: According to the comments it'd reduce the speed from 8125 MB/s back to 3077 MB/s. I get that unsafe is not nice but is it really worth throwing away a 264% speed increase?

LXY1226: There is another solution that uses package `binary`, that might take effect on removing `unsafe` and keeping acceptable performance (will still get worse), but the code will be more chaotic due to different solutions in different archs.  
The decline of 264% is really painful, so I agree with namazso 👍

LXY1226:
Benchmark results: 
```
goos: linux
goarch: amd64
cpu: Intel(R) Celeron(R) N5105 @ 2.00GHz
QuickXorHash-Before  2.49ms   422MB/s ±11%  100.00%
QuickXorHash-Bytes    796µs  1317MB/s ± 1%  212.50% △212%
QuickXorHash-uint64   129µs  8129MB/s ± 1% 1828.46% △617%
QuickXorHash-Subtle  87.9µs 11932MB/s ± 5% 2730.83% △146%
```

backend/onedrive/quickxorhash/quickxorhash.go
ncw: Could this use of unsafe disappear with go1.20?

LXY1226: yes, it will move into runtime after go1.22 is released (rclone's go version policy)

ncw: So could we make this with 3GB/s without `unsafe` and make a version with go1.20 which runs at 8GB/s also without `unsafe` in rclone code. I don't care if there is a use of `unsafe` in the crypto library as Google have better tools and more manpower than me to validate such uses.

Cnly:
The performance boost looks impressive indeed :) But I also kinda think we may not need to bring the `unsafe` into the code at this point. For one thing it might need extra effort to validate maintain, as @ncw said. For another, go 1.20 is around the corner and, if I understand correctly, we'll get the even larger boost without `unsafe`, which the majority of rclone users will also get if they install the official releases compiled with go 1.20. Even if we achieve better performance with `unsafe` now, we may quickly want to remove it again at that time.

LXY1226:
okay, i'll try my best to achieve more without unsafe and push back later.

ncw:
go1.20 is about to be released and rclone v1.61 will very likely be released using it.

ncw:
I have merged to master now go1.20 to be the default compiler for rclone.

So if you want to adjust this PR so it uses the go1.20 goodness for go >= 1.20 (need build tag `go1.20`) that would be great :-)

The update to go1.20 took a lot longer than I thought it was going to as there were two changes in the go stdlib which caused regressions. All fixed now though.

LXY1226:
Sorry for the lack of response recently 😢   
I have removed the `unsafe` part and it is ready to merge now. But I don't have a clue about the test failure mentioned above, can you help me trigger the pipeline again? 🤔

backend/onedrive/quickxorhash/xor_1.20.go
```diff
@@ -0,0 +1,9 @@
+//go:build 1.20 || 1.21 || 1.22
```
ncw, line 1: This build constraint should say just

    //go:build go1.20

I think. That means go 1.20 or later.

backend/onedrive/quickxorhash/xor.go
```diff
@@ -0,0 +1,20 @@
+//go:build !1.20 || !1.21 || !1.22
```
ncw, line 1: This build constraint should say just

    //go:build !go1.20

I think.

That means go version < 1.20

ncw (commented):
This is looking very good - thank you.

I put a note about the build constraints - I don't think they are quite right.

backend/onedrive/quickxorhash/xor.go
LXY1226: Thanks for suggestions! I don't use go's build constraints very often so I'm not familiar with them. Applied in last amend-commit. thanks again 🍻

backend/onedrive/quickxorhash/xor_1.20.go
```diff
@@ -0,0 +1,9 @@
+//go:build 1.20
```
ncw, line 1: Likewise

    //go:build go1.20

ncw:
I've merged that now - thank you very much, and sorry it has taken so long!

PR #6667, opened 2023-01-04
Just a quick typo fix on the documentation

juanbretti:

Small typo on the documentation.

No.

#### Checklist

### DIFF
--- a/docs/content/union.md
+++ b/docs/content/union.md
@@ -252,7 +252,7 @@ Properties:
 - Config:      min_free_space
 - Env Var:     RCLONE_UNION_MIN_FREE_SPACE
 - Type:        SizeSuffix
-- Default:     1Gi
+- Default:     1GiB
 
 ### Metadata
 


ncw:
This is an auto generated file so you can't just edit it.

Maybe the default should be changed, but I don't see an easy way of doing it. @albertony ?

albertony:
@juanbretti: In short it is not a typo. The value is a number with size suffix but no unit. And to emphasize IEC/binary interpretation instead of SI/decimal the additional `i` is included -  `1Gi` instead of `1G` to emphasize it is `1024M` and not `1000M`. https://rclone.org/docs/#human-readable explains part of it. Maybe doc could be improved (right answer: yes - it always can.. :)

@ncw: Maybe it should change, but it would lead to some refactoring, I think, as it is currently handled as just a size value, without unit. I would have to set aside some time to dig more into it, if this is something we would really like to consider changing.

ncw:
@albertony We could make fs.SizeSuffix print the `B` for us. It seems to parse it already. That won't be correct for `Byte/s` units though so we'd need a fs.SizeSuffixPerSecond unit or something like that.

I'm not sure its worth fixing as it doesn't seem to cause confusion.

albertony:
Yes, there is already a simple method `ByteUnit()` on `fs.SizeSuffix` which does this, and there is `ByteRateUnit()` which gives `B/s`. But the challenge is the logic to pick the right one in each case, and perhaps a separate type as you indicate would be good, but hopefully we can leave it as is for now..

ncw:
OK, let's leave this as-is for the moment.

Thank you @juanbretti for your contribution - we can't merge this one but we appreciate the effort you put into it.

Issue #6670, opened 2023-01-06
 SignatureDoesNotMatch: release 1.60.0 broke the copy using S3 backend with Other (GCS) as provider


panthony:
#### The associated forum post URL from `https://forum.rclone.org`

N/A

#### What is the problem you are having with rclone?

The copy of a file using S3 backend with `Other` provider (Google Cloud Storage) started to fail with the 1.60.0 release.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.60.0
- os/version: darwin 12.5.1 (64 bit)
- os/kernel: 21.6.0 (x86_64)
- os/type: darwin
- os/arch: amd64
- go/version: go1.19.2
- go/linking: dynamic
- go/tags: cmount
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

macOS 12.5.1, x86_64

#### Which cloud storage system are you using? (e.g. Google Drive)

S3 with Other as provider.

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
./rclone --config gcs_over_s3.conf copy source:/bucket/file .
```

With a config like:
```
[source]
type = s3
access_key_id = XXX
provider = Other
secret_access_key = XXX
endpoint = https://storage.googleapis.com
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
Failed to copy: failed to open source object: SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.
```

Looking at the result of `--dump headers`:

- HEAD request still works
- GET request fails

When I compare the output between 1.60.0 and 1.59.2 for the GET request the only difference I can see (despite dates,...) is the order of request headers..

FAIL
```
Host: storage.googleapis.com
User-Agent: rclone/v1.60.0
Accept-Encoding: gzip
Authorization: XXXX
X-Amz-Content-Sha256: XXXX
X-Amz-Date: 20230106T081010Z
```

WORK
```
Host: storage.googleapis.com
User-Agent: rclone/v1.59.2
Authorization: XXXX
X-Amz-Content-Sha256: XXXXXX
X-Amz-Date: 20230106T080942Z
Accept-Encoding: gzip
```

The `X-Amz-Content-Sha256` is the same value for both request.

The headers of the HEAD request (works) of v1.60.0:

```
Host: storage.googleapis.com
User-Agent: rclone/v1.60.0
[...]

PR #6677, opened 2023-01-09
s3: add GCS to provider list

panthony:

Add GCS as a provider to workaround a specific issue with that storage.

https://github.com/rclone/rclone/issues/6670

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -66,7 +66,7 @@ import (
 func init() {
 	fs.Register(&fs.RegInfo{
 		Name:        "s3",
-		Description: "Amazon S3 Compliant Storage Providers including AWS, Alibaba, Ceph, China Mobile, Cloudflare, ArvanCloud, DigitalOcean, Dreamhost, Huawei OBS, IBM COS, IDrive e2, IONOS Cloud, Liara, Lyve Cloud, Minio, Netease, RackCorp, Scaleway, SeaweedFS, StackPath, Storj, Tencent COS, Qiniu and Wasabi",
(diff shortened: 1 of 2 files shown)


backend/s3/s3.go
```diff
@@ -2393,6 +2404,7 @@ type Options struct {
 	VersionAt             fs.Time              `config:"version_at"`
 	Decompress            bool                 `config:"decompress"`
 	MightGzip             fs.Tristate          `config:"might_gzip"`
+	ForceGzipHeader       bool                 `config:"force_gzip_header"`
```
ncw, line 2407: This should be an `fs.Tristate`.

The default is `auto` which should use the provider default. The user can set it explicitly.

----

This also needs to appear in the Options above which is where the help docs go.

Note that not having this header breaks downloading files with `Content-Encoding: gzip` so it might be worth experimenting with that before writing the help.

----

I'm not 100% happy with the name `ForceGzipHeader`, how about  calling it `UseAcceptEncodingGzip`?

backend/s3/s3.go
```diff
@@ -2882,6 +2899,9 @@ func setQuirks(opt *Options) {
 		opt.MightGzip.Valid = true
 		opt.MightGzip.Value = mightGzip
 	}
+
+	// set ForceGzipHeader depending on provider
+	opt.ForceGzipHeader = forceGzipHeader
```
ncw, line 2904: This should then say

```
if !opt.ForceGzipHeader.Valid {
    opt.ForceGzipHeader = forceGzipHeader
}
```

ncw (commented):
Very good :-)

I put some comments inline about the new flag.

Thank you.

backend/s3/s3.go
panthony: Fine by me! PR updated 👍🏻

ncw (approved):
This looks great now - apologies for the delay merging. This will make it into the 1.63 release.

panthony:
No worries! Thanks for the merge 👍🏻

PR #6678, opened 2023-01-09
lib/oauthutil: don't retry for probably fatal errors

Cnly:
According to the OAuth2 RFC (https://www.rfc-editor.org/rfc/rfc6749#section-5.2), the server should return a 400 (or maybe sometimes 401) for a list of unrecoverable errors, such as expired refresh tokens. Rclone shouldn't still make 5 attempts to get a token in those cases. @ncw do you see any potential problem with this?

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -219,6 +219,15 @@ func (ts *TokenSource) Token() (*oauth2.Token, error) {
 	)
 	const maxTries = 5
 
+	shouldRetry := func(err error) bool {
+		if rErr, ok := err.(*oauth2.RetrieveError); ok {
+			if rErr.Response.StatusCode == 400 || rErr.Response.StatusCode == 401 {
+				return false
+			}
+		}
+		return true
+	}
+
 	// Try getting the token a few times
 	for i := 1; i <= maxTries; i++ {
 		// Try reading the token from the config file in case it has
@@ -239,7 +248,7 @@ func (ts *TokenSource) Token() (*oauth2.Token, error) {
 		}
 
 		token, err = ts.tokenSource.Token()
-		if err == nil {
+		if err == nil || !shouldRetry(err) {
 			break
 		}
 		fs.Debugf(ts.name, "Token refresh failed try %d/%d: %v", i, maxTries, err)
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -204,6 +204,37 @@ func (ts *TokenSource) reReadToken() (changed bool) {
 	return changed
 }
 
+type retrieveErrResponse struct {
+	Error string `json:"error"`
+}
+
+// If err is nil or an error other than fatal OAuth errors, returns err itself.
+// Otherwise returns a more user-friendly error.
+func maybeWrapOAuthError(err error, remoteName string) (newErr error) {
+	newErr = err
+	if rErr, ok := err.(*oauth2.RetrieveError); ok {
+		if rErr.Response.StatusCode == 400 || rErr.Response.StatusCode == 401 {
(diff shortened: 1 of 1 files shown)


ncw:
I think the idea looks like a good one. We shouldn't retry 400 errors. 401 too probably.

The way we do this in most backends is to call

https://github.com/rclone/rclone/blob/98fa93f6d17909e9c6b9778a4dc9e67c58151f6a/fs/fserrors/error.go#L440

And then have a list of HTTP error codes we will retry, eg

https://github.com/rclone/rclone/blob/98fa93f6d17909e9c6b9778a4dc9e67c58151f6a/backend/box/box.go#L318-L326

Is it worth using this method? That will retry fewer errors, but it should be a lot more targeted.

Cnly:
Actually there's another thing I thought about that didn't go into this PR. Right now when the user has an expired refresh token the error looks like

```
2023/01/09 14:18:27 Failed to create file system for "xxx:": failed to get root: Get "https://xxx/root": couldn't fetch token - maybe it has expired? - refresh with "rclone config reconnect xxx:": oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description":"AADSTS70000: The user could not be authenticated as the grant is expired. The user must sign in again. <...>
```

which is not exactly user friendly. Do you think we should make the new `shouldRetry` function give a better message and hide the original response in debug log? And with that we won't need to think about the existing function anymore.

ncw:
> Actually there's another thing I thought about that didn't go into this PR. Right now when the user has an expired refresh token the error looks like
> 
> ```
> 2023/01/09 14:18:27 Failed to create file system for "xxx:": failed to get root: Get "https://xxx/root": couldn't fetch token - maybe it has expired? - refresh with "rclone config reconnect xxx:": oauth2: cannot fetch token: 400 Bad Request
> Response: {"error":"invalid_grant","error_description":"AADSTS70000: The user could not be authenticated as the grant is expired. The user must sign in again. <...>
> ```

Its rather wordy but correct that error. Not that users are very good at reading errors!

As for users, this is the important bit

> couldn't fetch token - maybe it has expired? - refresh with "rclone config reconnect xxx:"

The rest of the message could conceivably show a different error I guess.

Were you thinking that we should be parsing the Response - it appears to have useful stuff in, though I don't know how well standardised it is in the oauth protocol.

This is clear, but in an rclone context "must sign in again" isn't obvious.

> which is not exactly user friendly. Do you think we should make the new `shouldRetry` function give a better message and hide the original response in debug log? And with that we won't need to think about the existing function anymore.

Better messages are always good! If it is standardised, parsing the `"error":"invalid_grant"` sounds like a great idea to return a targeted message and in other cases we can carry on showing the full message.

Cnly:
> Its rather wordy but correct that error. Not that users are very good at reading errors!

With the new commit I pushed today we should have 2 kinds of messages based on the returned `error` field (which is a required enum field in the RFC), and if the user runs with `-vv` the original response will be included.

```
Failed to create file system for "xxx:": failed to get root: Get "https://xxx/root": couldn't fetch token: invalid_grant: maybe token expired? - try refreshing with "rclone config reconnect xxx:"

Failed to create file system for "xxx:": failed to get root: Get "https://xxx/root": couldn't fetch token: invalid_client: if you're using your own client id/secret, make sure they're properly set up following the docs

DEBUG : xxx: got fatal oauth error: oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description": <...>
```

So here my only concern is that some services may choose to return errors other than `invalid_grant` to indicate expired refresh tokens (I hope not).

ncw (approved):
Love the new error messages :-)

Looks good to merge.

Cnly:
Alright, let me merge it and we'll see how things turn out.

Issue #5319, opened 2021-05-10
Provide alternate image registry to dockerhub


onedr0p:
#### What problem are you are trying to solve?

Hello,

In order to combat dockerhub's imposed rate limits it would be great to host these images on ghcr or quay.io. Are there on any thoughts from the community on this initiation?

ncw:
There shouldn't be any rate limits for rclone on docker hub - I've registered rclone with them as an open source project.

Are you seeing rate limits?

onedr0p:
Interesting, that's good to know. Too bad there's no way to see this on https://hub.docker.com/u/rclone to see you are a supported Dockerhub "open-source project".

I guess I will leave this issue open to request the image also being pushed to GHCR.

ncw:
> Interesting, that's good to know. Too bad there's no way to see this on https://hub.docker.com/u/rclone to see you are a supported Dockerhub "open-source project".

I think docker are missing a trick there...
 
> I guess I will leave this issue open to request the image also being pushed to GHCR.

GHCR == GitHub container registry.

Do they charge for open source projects?

A PR to adjust the build scripts to push to GHCR would probably be quite easy

onedr0p:
> GHCR == GitHub container registry.

Yes

> Do they charge for open source projects?

No

ferferga:
@onedr0p I believe this issue can be closed as solved?

PR #5219, opened 2021-04-07
backend/ftp: fix implicit TLS after PR #4266

ivandeex:
#### What is the purpose of this change?

PR #4266 modified ftpConnection to make ftp library into using
a custom dial function which is QoS aware and takes care of TLS.
([edfe183ba,ftp.go:222](https://github.com/rclone/rclone/commit/edfe183ba2f95120062596c98f23fe91e615b6f7#diff-cf7209b00e4d37b7cba2a369884faf762795e7860dbb6d6a417635e8b3b778e7R222))

However the ServerConn.Login function from the ftp library also needs TLS config passed explicitly
as a trigger for sending PSBZ and PROT options to FTP server.
([jlaffaye/ftp,ftp.go:311](https://github.com/jlaffaye/ftp/blob/master/ftp.go#L311))

This was not taken care of resulting in failure to connect via FTP with implicit TLS.

This PR fixes that and slightly refactors custom dialer.

#### Was the change discussed in an issue or in the forum before?

Fixes #5210

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-pull-request).
- [ ] I have added tests for all changes in this PR if appropriate.
- [ ] I have added documentation for the changes if appropriate.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [x] I'm done, this Pull Request is ready for review :-)

#### Testing

Successfully tested against FileZilla server with _Require TLS session resumption_ unticked as recommended by https://github.com/rclone/rclone/issues/5210#issuecomment-813554177
1. FileZilla, plain FTP on port 21
2. FileZilla, implicit TLS on port 21
3. FileZilla, explicit TLS on port 990
4. Quick integration suite passed: https://github.com/ivandeex/rclone/runs/2288274298?check_suite_focus=true
5. Got user confirmation at https://github.com/rclone/rclone/issues/5210#issuecomment-815158620

@ncw Please review

PR #6698, opened 2023-01-15
#5319 add ghcr option for docker images, and use the upstream build actions

RichardoC:
#### Checklist


RichardoC:
I'm only changing the beta workflow with this PR so that we can confirm everything works as expected. If it does then we can also change the release workflow too.

ncw:
I'd like a bit more discussion of this please?

I'm not sure I understand the reason for the change.

Will the images still end up in docker hub as they do currently?

Thanks

RichardoC:
Of course!

This will push images to both Dockerhub and GHCR, the section starting on line 53 is how this configured. If a registry is defined, it'll push to there and if there isn't (eg rclone/rclone:beta) it'll be pushed to dockerhub.


If you'd prefer this can we re-scoped to only change the workflow to the upstream actions and then discussions about dockerhub (or not) can take place in #5219

ncw:
I see!

It seems like a good idea to upload packages in both places.

Using the github supported actions is a good idea also I think.

I don't think that is the right issue number though? I think you meant #5319

I'll run the CI on this and see what happens!

RichardoC:
Thanks, I've just updated the issue reference. Would you prefer this as 2 PRs? One for the github actions, and one for the ghcr one?
I'm then planning to follow up with the equivalents but for the release steps

RichardoC:
Is that windows failure a known issue?

albertony:
> Is that windows failure a known issue?

I triggered a rebuild, and its gone.

RichardoC:
Yeah, I don't think you'll be able to run this from a test branch but I'll change this to be manually triggered after merging for testing if that suits?

ncw:
How do you manually trigger it? Does it have to be on master first, or can I do it somehow from this PR?

RichardoC:
Unfortunately I don't think we can run the workflow on this PR since it's a branch outside this project. 

I had hoped you'd have the option to manually trigger it somehow when I added the `on: workflow_dispatch` section but I don't know where that would show as an option. If you can't find that I guess we'll have to merge it?
If that's the case I could add this as a separate workflow, and you could keep both old and new version until you're confident and then we remove the old one?
Sorry I can't give an easy answer on this!

ncw:
Hi @RichardoC - apologies for the delay on this.

Now I have 1.62 out of the door I can experiment with this. Do you agree merging it is the right thing to do and we can see if it works properly?

Do I need to set up GHCR for use with rclone?

Thanks

RichardoC:
Yeah, I think this is still worth trying as it gives you users multiple options for obtaining the container images. I believe this PR should be all that is needed. 
On one of my personal [projects](https://github.com/RichardoC/kube-audit-rest) this was all that was needed to make the images available via github

ncw (approved):
That is great - thank you :-)

I'll merge this now and we can see how well it works. If it works OK then can you do a PR for the release build too?

I'll probably have a 1.63.3 to do in a week or two so it can get tested for real then.

RichardoC:
Sounds good, happy to raise that when we're happy with this on the beta image!

ncw:
I put the correct credentials in, and this one got a bit further

https://github.com/rclone/rclone/actions/runs/4448657379/jobs/7811754107

```
#2 [internal] load build definition from Dockerfile-alpine
#2 transferring dockerfile: 2B done
#2 DONE 0.0s
ERROR: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount3670878127/Dockerfile-alpine: no such file or directory
Error: buildx failed with: ERROR: failed to solve: failed to read dockerfile: open /tmp/buildkit-mount3670878127/Dockerfile-alpine: no such file or directory
```

I'll leave it broken over the weekend - that isn't a problem.

PR #6705, opened 2023-01-17
cmd/ncdu: refactor redraw handling

eNV25:

It refactors the handling of our `Draw` function, the tcell `Screen.Show` and `Screen.Sync` functions, the tcell resize event and the tcell `Ctrl-L` keyboard event.

This change should make it smoother and changes more instantaneous.

This also renames the `ncdu.Show` function to `ncdu.Run`, to reduce confusion with `Screen.Show`.

n/a

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -81,7 +81,7 @@ the remote you can also use the [size](/commands/rclone_size/) command.
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
 		cmd.Run(false, false, command, func() error {
-			return NewUI(fsrc).Show()
+			return NewUI(fsrc).Run()
 		})
 	},
 }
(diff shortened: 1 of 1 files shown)


cmd/ncdu/ncdu.go
```diff
@@ -1026,11 +1017,16 @@ outer:
...
+					goto redraw
 				}
 			}
 		}
-		// listen to key presses, etc.
+
+		u.Draw()
+		u.s.Show()
+	redraw:
+		u.Draw()
```
ncw, line 1028: Why does it `Draw `it twice?

ncw (commented):
This seems like a good idea :-)

Query inline!

cmd/ncdu/ncdu.go
eNV25: Oops. I forgot to do a `continue` before the label.

Maybe I shouldn't use a `goto`? I thought it would be more clear this way.

ncw: I am an old C programmer so `goto exit` is very familiar.

However if it is easy to remove the goto then we probably should!

eNV25:
@ncw

PR #6707, opened 2023-01-17
cmd/ncdu: fix screen corruption

### LINKED ISSUE #6271: rclone ncdu: Fix screen corruption when logs happen

This was discussed in https://github.com/rclone/rclone/pull/6259 which was an attempt to fix this problem.

Currently the `rclone ncdu` screen gets corrupted if rclone logs anything. This can be fixed with CTRL-L but it would be nice if it didn't happen at all.

After a bit of discussion the best approach is probably

- capture the logging in the same way `--progress` does
    - https://github.com/rclone/rclone/blob/7a909ebfb0d100187fab7949455bf74b8ce07042/cmd/progress.go#L29-L40
- store the logs as they come in up to a maximum of 1000 (say)
- show the last log in the status bar
- show the logs in the buffer when exiting ncdu

At some point we could also make a log viewing window which would allow you to view / scroll the logs so far.

#### How to use GitHub

eNV25:

Fixes #6271

This PR implements two different approaches to fixing the issue, mentioned in the issue.

Issue #6271

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -19,6 +19,7 @@ import (
 	"github.com/rclone/rclone/cmd/ncdu/scan"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/fspath"
+	"github.com/rclone/rclone/fs/log"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/rivo/uniseg"
 	"github.com/spf13/cobra"
@@ -911,6 +912,28 @@ func (u *UI) Show() error {
 	if err != nil {
 		return fmt.Errorf("screen init: %w", err)
 	}
+
+	// Hijack fs.LogPrint so that it doesn't corrupt the screen.
+	if logPrint := fs.LogPrint; !log.Redirected() {
+		type log struct {
+			text  string
+			level fs.LogLevel
+		}
+		var logs []log
+		fs.LogPrint = func(level fs.LogLevel, text string) {
+			if len(logs) > 100 {
+				logs = logs[len(logs)-100:]
+			}
+			logs = append(logs, log{level: level, text: text})
+		}
+		defer func() {
+			fs.LogPrint = logPrint
+			for i := range logs {
+				logPrint(logs[i].level, logs[i].text)
+			}
+		}()
+	}
+
 	defer u.s.Fini()
 
 	// scan the disk in the background


eNV25:
TIP: To create a warning on purpose, you can use a Google Drive backend with a dangling shortcut.

ncw (approved):
Sorry this has taken me a while to review.

I think this code looks good and I think it will work correctly on non terminfo systems, so I will merge it now for 1.63 - thank you :-)

ncw:
Actually, maybe I won't merge it since you marked it as EXPERIMENTAL in the commit message :-)

I'm happy with the approach - the alternate screen is neat on terminfo systems and the printing the log messages at the end is OK on the others.

Anything else you want to do to the commit before merge?

eNV25:
I don't know if you have noticed already, but the terminfo approach causes some screen flickering. I personally don't think it's too bad, but if you have many logs it pretty noticeable.

To mitigate it, I tried adding [synchronized update escape codes](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036), but it didn't seem to help in kitty. I did NOT add this in the latest change, but if you want that anyway I can add it.

If you think this is not an issue, you can merge as-is.

The last change if only a small cleanup.

eNV25:
> I gave it a bit more testing.
> 
> I noticed that it deadlocks after I press `q` when running
> 
> ```
> rclone ncdu -vv . --exclude "*"
> ```
> 

I see it too. I don't why that is, but I'll check it out with dlv.

> Using ubuntu 22.04 and xfce4-terminal
> 
> Which makes me think the double buffer term trick isn't ready for the big time.
> 
> I guess the alternative would be to do all platforms the same way by storing the log then printing it at the end so maybe we should merge that bit of code now and leave the double buffer trick?
> 
> This maybe has the advantage that we could display the last log on the screen
> 

I agree, the array option seems like a safer choice.

> > To mitigate it, I tried adding [synchronized update escape codes](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036), but it didn't seem to help in kitty. I did NOT add this in the latest change, but if you want that anyway I can add it.
> 
> How many terminals support those codes?

Not many terminal support this. And there are two competing sets of escape code. But tmux has a terminfo extension for that, see the `Sync` option in `man tmux`. tcell's terminfo library doesn't support that, so it requires upstream changes.

cmd/ncdu/ncdu.go
```diff
@@ -911,6 +912,27 @@ func (u *UI) Show() error {
...
+		}
+		var logs []log
+		fs.LogPrint = func(level fs.LogLevel, text string) {
+			if len(logs) > 100 {
+				logs = logs[len(logs)-100:]
+			}
+			logs = append(logs, log{level: level, text: text})
+		}
+		defer func() {
+			for i := range logs {
```
ncw, line 930: This function should also replace `fs.LogPrint = logPrint` I think?

ncw (commented):
Looks good - thank you :-)

See query inline.

Let's get this merged first then we can do a followup with fancy term stuff if we think we can get it to work reliably.

cmd/ncdu/ncdu.go
eNV25: True. I thought I pushed that, but it turns out I missed it.

ncw (approved):
Perfect - thank you :-)

PR #6712, opened 2023-01-19
Mount in network mode by default on windows

albertony:

NB: Beta here: https://beta.rclone.org/branch/mount-windows-default-network-mode/

Part 1: Mount in network mode by default on windows

Part 2:  Fix default volume name on windows being invalid for network mode.
- When mounting a local filesystem on windows the default assigned volume name is illegal in network mode (illegal as a network share path), which means when this PR changes the network mode to being default true this leads to mounting of local paths no longer working without either explicitly turning off network mode or explicitly setting a valid value with --volname. 
- This part was split from https://github.com/rclone/rclone/pull/6234, see that for details.

https://forum.rclone.org/t/unexpected-results-when-mounting-azure-blob-on-windows/35535/17

#### Checklist

### DIFF
--- a/cmd/cmount/mountpoint_other.go
+++ b/cmd/cmount/mountpoint_other.go
@@ -26,5 +26,6 @@ func getMountpoint(f fs.Fs, mountPath string, opt *mountlib.Options) (string, er
 	if err = mountlib.CheckAllowNonEmpty(mountPath, opt); err != nil {
 		return "", err
 	}
+	opt.VolumeName = mountlib.MakeVolumeNameValidOnUnix(opt.VolumeName)
 	return mountPath, nil
 }
--- a/cmd/cmount/mountpoint_windows.go
+++ b/cmd/cmount/mountpoint_windows.go
@@ -9,9 +9,11 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/rclone/rclone/cmd/mountlib"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/file"
 )
 
@@ -19,10 +21,13 @@ var isDriveRegex = regexp.MustCompile(`^[a-zA-Z]\:$`)
 var isDriveRootPathRegex = regexp.MustCompile(`^[a-zA-Z]\:\\$`)
 var isDriveOrRootPathRegex = regexp.MustCompile(`^[a-zA-Z]\:\\?$`)
 var isNetworkSharePathRegex = regexp.MustCompile(`^\\\\[^\\\?]+\\[^\\]`)
+var isAnyPathSeparatorRegex = regexp.MustCompile(`[/\\]+`) // Matches any path separators, slash or backslash, or sequences of them
 
-// isNetworkSharePath returns true if the given string is a valid network share path,
-// in the basic UNC format "\\Server\Share\Path", where the first two path components
-// are required ("\\Server\Share", which represents the volume).
+// isNetworkSharePath returns true if the given string is a network share path,
+// in the basic UNC format "\\Server\Share\Path". The first two path components
+// are required ("\\Server\Share"), and represents the volume. The rest of the
+// string can be anything, i.e. can be a nested path ("\\Server\Share\Path\Path\Path").
+// Actual validity of the path, e.g. if it contains invalid characters, is not considered.
 // Extended-length UNC format "\\?\UNC\Server\Share\Path" is not considered, as it is
 // not supported by cgofuse/winfsp, so returns false for any paths with prefix "\\?\".
 // Note: There is a UNCPath function in lib/file, but it refers to any extended-length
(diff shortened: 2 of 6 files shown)


ncw:
Do you want to put the volume name fixes as another commit on this PR so we merge them together fixing the problem you describe above in the process?

Masamune3210:
Just a heads up, at least on my machine, mounting one drive works fine in network mode, but trying to mount two at once causes the rclone service to fail to start

ncw:
@Masamune3210 with two separate rclone instances? Or with one instance and using the rc?

Masamune3210:
Separate

ncw:
@albertony can you confirm this?

albertony:
Not sure what the case is; but if it is mounting the same "source" twice, without customizing volume name, then yes:

```
rclone mount C:\Temp X:
```

```
rclone mount C:\Temp Y:
```

The second one (in time) will fail, because the network share path (volume name) default is the same (`\\server\？\C：\Temp`).

Was this what you were tryging @Masamune3210?

albertony:
(FYI: Did a quick rebase on master to resolve conflicts etc, but intend to go through the code more thorough later)

ncw:
Is this ready for review?

albertony:
No, not really. I wanted to go through it again. Will ping you.

albertony:
Maybe you could take a look now @ncw?
I seem to keep delaying the decision to invest the necessary time to finish this one, so I'll try to get it done now. I did a quick read-through and test on Windows, and it seems to me it works well as it is, but you indicated earlier you didn't necessarily agree with the strategy so let me know what you think now.

cmd/mount/mount.go
```diff
@@ -79,6 +79,7 @@ func mount(VFS *vfs.VFS, mountpoint string, opt *mountlib.Options) (<-chan error
 	if err := mountlib.CheckAllowNonEmpty(mountpoint, opt); err != nil {
 		return nil, nil, err
 	}
+	opt.VolumeName = mountlib.MakeVolumeNameValidOnUnix(opt.VolumeName)
```
ncw, line 82: This should be done in mountlib shouldn't it?

cmd/mount2/mount.go
```diff
@@ -151,6 +151,7 @@ func mount(VFS *vfs.VFS, mountpoint string, opt *mountlib.Options) (<-chan error
 	if err := mountlib.CheckAllowNonEmpty(mountpoint, opt); err != nil {
 		return nil, nil, err
 	}
+	opt.VolumeName = mountlib.MakeVolumeNameValidOnUnix(opt.VolumeName)
```
ncw, line 154: and here?

cmd/cmount/mountpoint_other.go
```diff
@@ -26,5 +26,6 @@ func getMountpoint(f fs.Fs, mountPath string, opt *mountlib.Options) (string, er
 	if err = mountlib.CheckAllowNonEmpty(mountPath, opt); err != nil {
 		return "", err
 	}
+	opt.VolumeName = mountlib.MakeVolumeNameValidOnUnix(opt.VolumeName)
```
ncw, line 29: This should be done in mountlib I think

cmd/cmount/mountpoint_windows.go
```diff
@@ -132,30 +137,47 @@ func handleLocalMountpath(f fs.Fs, mountpath string, opt *mountlib.Options) (str
 	return mountpath, nil
 }
 
+// networkSharePathEncoder is an encoder used to make strings valid as (part of) Windows network share UNC paths
+const networkSharePathEncoder = (encoder.EncodeZero | // NUL(0x00)
```
ncw, line 141: I would have thought you could put the encoder and `encodeNetworkSharePath` into mountlib and make this patch a lot neater. I think then it would only be in mountlib.

ncw (commented):
I was thinking that maybe you should be doing all this validation / volume manipulation in mountlib. That would make this patch much simpler I think.

The first commit looks straight forward though.

albertony:
> I was thinking that maybe you should be doing all this validation / volume manipulation in mountlib. That would make this patch much simpler I think.

> I would have thought you could put the encoder and encodeNetworkSharePath into mountlib and make this patch a lot neater. I think then it would only be in mountlib.

Some of the challenge here is that the windows cmount implementation has the distinction between "disk mode" and "network mode" mounting, each of them with different rules regarding special characters. And also that it supports an implicit network mode specification where volume names are taken from mountpath instead of the volname option (e.g. `rclone mount remote:path/to/files \\cloud\remote`), as well as auto drive letter (e.g. `rclone mount remote:path/to/files *`). My thinking was that these things are mount-implementation specific, and therefore best fit in the cmount package. Maybe that is not so much the case? It is very much OS-specific, at least. Currently we only have cmount on Windows, so it is more of a theoretical/design question, I guess.

The volume handling could be moved into the common mountlib, but due to the mentioned challenges I think that would actually be a bigger PR. Most of mountpoint_windows.go probably would have to be moved. In short because the current mountlib.Mount, when mounting on windows/cmount, does not (necessarily) know the actual volume name and mount path. Don't you agree?

What do you think?

ncw:
> > I was thinking that maybe you should be doing all this validation / volume manipulation in mountlib. That would make this patch much simpler I think.
> 
> > I would have thought you could put the encoder and encodeNetworkSharePath into mountlib and make this patch a lot neater. I think then it would only be in mountlib.
> 
> Some of the challenge here is that the windows cmount implementation has the distinction between "disk mode" and "network mode" mounting, each of them with different rules regarding special characters.

Urg, so the layers are mixed up here...

> And also that it supports an implicit network mode specification where volume names are taken from mountpath instead of the volname option (e.g. `rclone mount remote:path/to/files \\cloud\remote`), as well as auto drive letter (e.g. `rclone mount remote:path/to/files *`). My thinking was that these things are mount-implementation specific, and therefore best fit in the cmount package. Maybe that is not so much the case? It is very much OS-specific, at least. Currently we only have cmount on Windows, so it is more of a theoretical/design question, I guess.

If we were talking about Linux volume name handling then we'd put it in mountlib once and use it 3 times. So the Windows handling should really go in the mountlib layer and that is what I was aiming at with the `SetVolumeName` calls.

However we do only have the one Windows implementation of mount. In theory we could make another and plug into the plan 9 interface in WSL2 but I don't see that happening any time soon! That would probably have a whole different set of rules about volume names.

> The volume handling could be moved into the common mountlib, but due to the mentioned challenges I think that would actually be a bigger PR. Most of mountpoint_windows.go probably would have to be moved. In short because the current mountlib.Mount, when mounting on windows/cmount, does not (necessarily) know the actual volume name and mount path. Don't you agree?
> 
> What do you think?

Hmm, let's try not to break it too much all at once :-) We can merge this and refactor later I think.

We need to merge both these commits at once don't we. Do you want to merge what we have for 1.62 or punt to 1.63?

albertony:
> We need to merge both these commits at once don't we.

I guess so. If just merging the change to network mode by default, then mounting a local filesystem on windows would not work without explicitly setting a valid volume name. I guess mounting local isn't the most relevant use case, but my guess is its used enough that it would look bad...(?)

albertony:
> We can merge this and refactor later I think.

Sounds good.

> Do you want to merge what we have for 1.62 or punt to 1.63?

If 1.63 is just hours or few days away, maybe its best to wait for it to "settle" in betas for some time...?

ncw:
Hi @albertony, tidying up the PR backlog.

Mounting in network mode by default on Windows is now in master (`opt.NetworkMode = true` in `cmd/cmount/mountpoint_windows.go`), so this PR's main change is effectively done. Closing as superseded, but please reopen if any of the parts here are still outstanding.

albertony:
It is not really the default still, when using the most basic command `rclone mount C:\Temp *`. It is default if using the alternative syntax `rclone mount C:\Temp \\server\temp`. But if you want a specific drive letter you still need to enable it explicitly  `rclone mount C:\Temp X: --network-mode`.

Also the other part of this PR is still an issue: `rclone mount C:\Temp * --network-mode -v` fails because it defaults to a volume name not valid for network mount, while a working alternative is `rclone mount C:\Temp * --network-mode --volname temp`.

PR #6719, opened 2023-01-22
build: fix condition for manual workflow run

albertony:

This fixes the the input property `manual` in the `inputs` context of our GitHub workflow. It was introduced in #5275, see that for background.

The first change is only a small improvement: Added a `description` property, which shows a more human-readable label in the "Run workflow" dialog, and added a `type` property with value "boolean" which makes GitHub present it as a checkbox, instead of a regular text box with the text "true" set by default.

Furthermore I discovered that the value is not really a boolean, even after I added the "type: boolean" property. It is in reality always a string - either the string 'true' or the string 'false'. But this means a condition `if: ${{ github.event.inputs.manual }}` will always be met, because any non-empty string is treated as boolean true, even the string 'false'!

#### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,12 +15,14 @@ on:
   workflow_dispatch:
     inputs:
       manual:
+        description: Manual run (bypass default conditions)
+        type: boolean
         required: true
         default: true
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


ncw (approved):
LGTM - thank you :-)

PR #6738, opened 2023-02-01
Avoid running workflow twice for pull requests with branch on main repo

albertony:

Currently, when pushing a branch to upstream rclone/rclone and making a pull request from that (instead of from a branch pushed to my own fork, which has the advantage of publishing beta binaries) the same github workflow runs twice - once for the branch push and once for the pull request.

Using the trick from here:

https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/

The condition is getting a bit long, unfortunately, but to break down the change: The existing condition
```
github.repository == 'rclone/rclone'
```
is extended
```
github.repository == 'rclone/rclone' && 
```
with a condition that it is not a pull request (i.e. it is branch push):
```
(github.event_name != 'pull_request'
```
or, if it is a pull request, then it is not from a branch on the same repo:
```

|| github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name))
```

I tested it by first pushing this change to a branch on my fork, and then creating a PR from that: #6737. It ran through 10 checks as before (the PR now shows duplicate checks because I messed it up by not closing/deleting before the next step). Then I pushed the branch to upstream and created a new PR from that, the current one. It also ran through 10 checks, instead of 20 (all checks twice) as before. The checks are run as "on: push". There are 3 additional checks shown for "on: pull_request" with status "skipped" - which means the label on the "Check" tab of the PR shows the number 13 instead of 10, but only the 10 checks are active - no duplicate runs anymore.

#### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ on:
 
 jobs:
   build:
-    if: ${{ github.repository == 'rclone/rclone' || github.event.inputs.manual == 'true' }}
+    if: ${{ github.event.inputs.manual == 'true' || (github.repository == 'rclone/rclone' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name)) }}
     timeout-minutes: 60
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


ncw (approved):
Brilliant - well done :-)

Please merge!

Issue #6741, opened 2023-02-02
 Customize HTML page used in responses from `rclone authorize` 


hwittenborn:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/customize-html-page-used-in-responses-from-rclone-authorize/35128/1

#### What is your current rclone version (output from `rclone version`)?
v1.53.3-DEV

#### What problem are you are trying to solve?
Customizing the HTML shown on `rclone authorize` output so that it fits my application's branding

#### How do you think rclone should be changed to solve that?
Add a flag to the `rclone authorize` command to read a Go template that will be used for output (i.e. `rclone authorize --template my-template.html`)
.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
I believe you are working on that @hwittenborn is that right?

hwittenborn:
Yeah that's correct @ncw. I'm just looking around the codebase to figure out how it's structured, but I'll send a message if I have any questions :)

hwittenborn:
I'm looking into this a little bit but I'm getting a bit stumped: I'm needing to pass a template into the `newAuthServer` like shown at https://github.com/rclone/rclone/blob/master/lib/oauthutil/oauthutil.go#L758, but when I attempt to add a `template string` argument I get the following error in my terminal:

```
lib/oauthutil/oauthutil.go:823:22: cannot use s.handleAuth (value of type func(w http.ResponseWriter, req *http.Request, template string)) as type func(http.ResponseWriter, *http.Request) in argument to mux.HandleFunc
```

From what it's looking like it looks like some library Rclone uses is requiring a certain function signature, but I'm not exactly sure how I would get around that. Should I just make a global variable that can be accessed outside that function and then just read that inside of it?

ncw:
> From what it's looking like it looks like some library Rclone uses is requiring a certain function signature, but I'm not exactly sure how I would get around that. Should I just make a global variable that can be accessed outside that function and then just read that inside of it?

What you want to do is get the template into the auth server struct

[...]

PR #6745, opened 2023-02-04
Add support for custom templates in 'rclone authorize'

hwittenborn:

Adds the ability to provide a custom template to `rclone authorize` calls. Previously a hardcoded template was used, but a custom one may want to be used, such as for branding purposes when integrating Rclone into another app (such is my personal use case).

See #6741 and https://forum.rclone.org/t/customize-html-page-used-in-responses-from-rclone-authorize/35128/1.

#### Checklist


hwittenborn:
@ncw: I'm just thinking about this now, but there's probably gonna need to be some stable API for this template, right? I'd also like to add some variables to it if you'd be fine with it, the main one I'm wanting at the moment is just the name of the remote type being authenticated. Any thoughts?

hwittenborn:
On second thought I probably don't need nothing stabilized too much, I'll probably end up bundling rclone with my application so preserving the version shouldn't be an issue. But just having some extra data that I could use would be nice, I don't care too much about all the effort into making a stable interface at this moment though.

lib/oauthutil/oauthutil.go
```diff
@@ -24,6 +25,11 @@ import (
 	"golang.org/x/oauth2"
 )
 
+var (
+	TemplatePath string
```
albertony, line 29: You should add a comment to this exported variable, which starts with the word "TemplatePath". The linter will complain, when I enable the workflow to run..

lib/oauthutil/oauthutil.go
```diff
@@ -587,6 +593,19 @@ version recommended):
 		}
 		return fs.ConfigGoto(newState("*oauth-done"))
 	case "*oauth-do":
+		// Make sure we can read the HTML template file if it was specified.
+		if len(TemplatePath) > 0 {
+			dat, err := os.ReadFile(TemplatePath)
+
+			if err != nil {
+				return nil, fmt.Errorf("Failed to read template file: %w", err)
```
albertony, line 601: Error messages should start with lowercase letter (lint issue).

ncw:
@hwittenborn we can document some template variables that will be available - that would be a great idea, and feel free to add more.

I'll allow the CI to run now so you can see the lint errors!

I'm not 100% happy with the way the template name gets passed in, I think there must be a better way, but I need to think about it :-)

hwittenborn:
> we can document some template variables that will be available - that would be a great idea, and feel free to add more.

On that note the `AuthResult` struct is capitalized so it's part of rclone's public API, correct? I'm just asking because I'd like to rename some of the fields in the struct (like `Name` to `ErrMsg`) so they better fit their purpose.

Also - it looked like some of the fields in the struct (`HelpURL`, and `Err`) are never added to the struct in the `oathutil.go` file. Is that intentional, or are they just there from a past time when they were used or something? If they're never gonna be used, I'd rather just remove them as part of the public API being created.

> I'm not 100% happy with the way the template name gets passed in

Is it the way it's passed on the CLI, how it's passed between functions, or what's it you don't like about it?

cmd/authorize/authorize.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/rclone/rclone/fs/config"
 	"github.com/rclone/rclone/fs/config/flags"
 	"github.com/spf13/cobra"
+	"github.com/rclone/rclone/lib/oauthutil"
```
albertony, line 11: The lint error is because the imports should be in alphabetical order, I'm using VSCode with go extension which does this automatically on save, but I think you can use the go fmt tool to do it file by file as well:

`gofmt -w cmd/authorize/authorize.go`

lib/oauthutil/oauthutil.go
```diff
@@ -13,7 +13,8 @@ import (
 	"strings"
 	"sync"
 	"time"
-
+	"os"
```
albertony, line 16: Same here.

ncw:
> > I'm not 100% happy with the way the template name gets passed in
> 
> Is it the way it's passed on the CLI, how it's passed between functions, or what's it you don't like about it?

I think the name of the template file needs to be passed as a config parameter. We should also allow this to be passed in its entirety also which will make usage via the API easier.

So add some more definitions like this

```diff
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -58,6 +58,12 @@ const (
 
 	// ConfigAuthNoBrowser indicates that we do not want to open browser
 	ConfigAuthNoBrowser = "config_auth_no_browser"
+
+	// ConfigTemplate template contents to be used in the config webserver
+	ConfigTemplate = "config_template"
+
+	// ConfigTemplateFile path to a template file to use for the config webserver
+	ConfigTemplateFile = "config_template_file"
 )
 
 // Storage defines an interface for loading and saving config to
```

And read them out of `m` in `lib/oauthutil`.

Passing the config via a command line parameter will mean it **only** works for `rclone authorize` and not for `rclone config` or use via the API.

You'd then do something like this in `rclone authorize`

```diff
--- a/cmd/authorize/authorize.go
+++ b/cmd/authorize/authorize.go
@@ -39,6 +39,6 @@ Use --template to generate HTML output via a custom Go template. If a blank stri
 	},
 	RunE: func(command *cobra.Command, args []string) error {
 		cmd.CheckArgs(1, 3, command, args)
-		return config.Authorize(context.Background(), args, noAutoBrowser)
+		return config.Authorize(context.Background(), args, noAutoBrowser, template)
 	},
 }
```

And something like this in `config.Authorize`

```diff
--- a/fs/config/authorize.go
+++ b/fs/config/authorize.go
@@ -15,7 +15,7 @@ import (
 //	rclone authorize "fs name"
 //	rclone authorize "fs name" "base64 encoded JSON blob"
 //	rclone authorize "fs name" "client id" "client secret"
-func Authorize(ctx context.Context, args []string, noAutoBrowser bool) error {
+func Authorize(ctx context.Context, args []string, noAutoBrowser bool, template string) error {
 	ctx = suppressConfirm(ctx)
 	ctx = fs.ConfigOAuthOnly(ctx)
 	switch len(args) {
@@ -40,6 +40,9 @@ func Authorize(ctx context.Context, args []string, noAutoBrowser bool) error {
 	if noAutoBrowser {
 		inM[ConfigAuthNoBrowser] = "true"
 	}
+	if template != "" {
+		inM[ConfigTemplateFile] = template
+	}
 
 	// Add extra parameters if supplied
 	if len(args) == 2 {
```

Yes it is rather convoluted but the rclone config system is very flexible!


> On that note the `AuthResult` struct is capitalized so it's part of rclone's public API, correct? I'm just asking because I'd like to rename some of the fields in the struct (like `Name` to `ErrMsg`) so they better fit their purpose.

Yes `Name` does look wrong now! I'd call it `Error`.

Rclone doesn't guarantee a stable public API so you can rename things if you want.
 
> Also - it looked like some of the fields in the struct (`HelpURL`, and `Err`) are never added to the struct in the `oathutil.go` file. Is that intentional, or are they just there from a past time when they were used or something? If they're never gonna be used, I'd rather just remove them as part of the public API being documented.

I think we can lose `Err` - we have a textual version in `Error`. I think I'd probably leave HelpURL just in case we do decide to add some URLs (would be nice!)

cmd/authorize/authorize.go
hwittenborn: I ran `go fmt` from the root of the repository, I'm not sure why it didn't catch it. I'll get that fixed though.

hwittenborn:
I just pushed some changes, but I won't be able to test them until I'm home in about two hours. It compiled and all that just fine though, I just got to make sure it's all working fine.

ncw:
@hwittenborn ping me here when you want me to give it another review.

hwittenborn:
Sorry for the wait again @ncw, I got a bit sidetracked on my application that I'm using rclone with, but I just confirmed everything is working locally, so it should be ready for another review.

hwittenborn:
@ncw: Any update on this?

ncw (approved):
I think this looks perfect now - thank you :-)

I will merge ~~now~~ after the CI passes.

hwittenborn:
Sorry for the delay @ncw, this should be fixed now though (my googling showed it just being some `gofmt` errors, but I guess I'll find out once the CI runs :p).

ncw:
Thank you for doing this and sorry it has taken a while to merge :-)

hwittenborn:
All good @ncw :) Any idea when it'll land in a release though? I'd like to start including it with my application, but ideally I can pull rclone from a release first to do that.

ncw:
1.62 is due in 10 days or so - it will land then :-) In the mean time it is in the [the latest beta](https://beta.rclone.org/).

Issue #6668, opened 2023-01-05
Colorizing the tree output as default causing issue on Windows Shell or PowerShell stdout


luddinho:
#### The associated forum post URL from `https://forum.rclone.org`
https://github.com/rclone/rclone/pull/6591

#### What is the problem you are having with rclone?
Since the revision 1.61.0 the colorization of the tree command has become default.
In the previous versions you must use the -C or the --color option explicit.

The issue is present on Windows Shell or PowerShell.

#### What is your rclone version (output from `rclone version`)
rclone v1.61.1
- os/version: Microsoft Windows 10 Pro 22H2 (64 bit)
- os/kernel: 10.0.19045.2364 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.19.4
- go/linking: static
- go/tags: cmount

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Microsoft Windows 10 Pro 22H2 (64 bit)

#### Which cloud storage system are you using? (e.g. Google Drive)

- Onedrive
- koofr
- sftp

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
Versrions before 1.61.0: `rclone tree Onedrive: -C` to colorize the output.

Since version 1.61.0 the -C option can be ignored `rclone tree Onedrive:`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
The default colorization works fine on unix systems, but not on Windows Shell or PowerShell.

The output looks like this:
```
rclone tree koofr:temp
←[1;34m/←[0m
└── ←[1;34msubdir←[0m
    ├── file_1
    ├── file_2
    ├── file_3
    └── file_4
```

On the unix terminal it looks as expected including colorization:
```
rclone tree koofr:temp
/
└── subdir
    ├── file_1
    ├── file_2
    ├── file_3
    └── file_4
```

Windows Shell does not support the unix colorization parameters.

This is the result when using the v1.59.2 binary where the colorization option is available.
Comparision between lsiting the tree as default (no colorization) and with the extra option "-C".

```
PS C:\Users\luddinho> C:\Tools\rclone-windows-amd64\rclone_1-59-2.exe tree koofr:temp
/
└── subdir
    ├── file_1
    ├── file_2
    ├── file_3
    └── file_4

1 directories, 4 files

[...]

PR #6760, opened 2023-02-10
Enable console virtual terminal sequences processing (ANSI/VT100 colors)

albertony:
On Windows this makes sure console mode `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is enabled on the rclone process, by using `GetConsoleMode`/`SetConsoleMode` functions from Windows API. This mode is default in many cases, e.g. when using the Windows Terminal application, but in other cases it relies on a registry setting:
```
[HKEY_CURRENT_USER\Console]
"VirtualTerminalLevel"=dword:00000001
```

(I'm still confused as to exactly how this works; when it is default enabled and not, as can be seen from the discussions in the linked PR and forum thread, but seems certain that it is not always enabled by default and therefore it is better to set it explicitly on the process).

Since rclone version 1.61.0 the tree command uses ANSI color sequences in output by default, and this lead to issues in some terminals that were not handling these (see links below).

Background:
- https://learn.microsoft.com/windows/console/console-virtual-terminal-sequences
- https://ss64.com/nt/syntax-ansi.html

#6668
#6591
https://forum.rclone.org/t/error-with-build-v1-61-1-tree-command-panic-runtime-error-invalid-memory-address-or-nil-pointer-dereference/35922

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -401,9 +401,15 @@ func initConfig() {
 	// Start accounting
 	accounting.Start(ctx)
 
-	// Hide console window
+	// Configure console
 	if ci.NoConsole {
+		// Hide the console window
 		terminal.HideConsole()
+	} else {
+		// Enable color support on stdout if possible.
+		// This enables virtual terminal processing on Windows 10,
+		// adding native support for ANSI/VT100 escape sequences.
+		terminal.EnableColorsStdout()
 	}
 
 	// Load filters
--- a/cmd/tree/tree.go
+++ b/cmd/tree/tree.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rclone/rclone/fs/dirtree"
 	"github.com/rclone/rclone/fs/log"
 	"github.com/rclone/rclone/fs/walk"
+	"github.com/rclone/rclone/lib/terminal"
 	"github.com/spf13/cobra"
 )
 
@@ -100,22 +101,26 @@ For a more interactive navigation of the remote see the
 	RunE: func(command *cobra.Command, args []string) error {
 		cmd.CheckArgs(1, 1, command, args)
 		fsrc := cmd.NewFsSrc(args)
-		outFile := os.Stdout
+		ci := fs.GetConfig(context.Background())
+		var outFile io.Writer
 		if outFileName != "" {
 			var err error
 			outFile, err = os.Create(outFileName)
 			if err != nil {
 				return fmt.Errorf("failed to create output file: %w", err)
 			}
+			opts.Colorize = false
+		} else {
+			terminal.Start()
+			outFile = terminal.Out
+			opts.Colorize = true
 		}
 		opts.VerSort = opts.VerSort || sort == "version"
 		opts.ModSort = opts.ModSort || sort == "mtime"
 		opts.CTimeSort = opts.CTimeSort || sort == "ctime"
 		opts.NameSort = sort == "name"
 		opts.SizeSort = sort == "size"
-		ci := fs.GetConfig(context.Background())
 		opts.UnitSize = ci.HumanReadable
-		opts.Colorize = ci.TerminalColorMode != fs.TerminalColorModeNever
 		if opts.DeepLevel == 0 {
 			opts.DeepLevel = ci.MaxDepth
 		}
--- a/lib/terminal/terminal.go
+++ b/lib/terminal/terminal.go
@@ -111,3 +111,14 @@ func Write(out []byte) {
 	Start()
 	_, _ = Out.Write(out)
 }
+
+// EnableColorsStdout enable colors if possible.
+// This enables virtual terminal processing on Windows 10 console,
+// adding native support for VT100 escape codes. When this terminal
+// package is used for output, the result is that the colorable library
+// don't have to decode the escapes and explicitely write text with color
+// formatting to the console using Windows API functions, but can simply
+// relay everything to stdout.
+func EnableColorsStdout() {
+	_ = colorable.EnableColorsStdout(nil)
+}


albertony:
Not sure if this is implemented correctly with regards to the other code in the terminal package, with the TerminalColorMode variable etc, introduced in https://github.com/rclone/rclone/pull/6591. Is that related at all, or is it only for the stats/progress/config doing more direct terminal manipulations? I didn't see the terminal.Start function being called on a simple tree command.

ncw:
> Not sure if this is implemented correctly with regards to the other code in the terminal package, with the TerminalColorMode config variable etc, introduced in #6591. I'm confused; is that related at all, or is it only for the stats/progress/config doing more direct terminal manipulations?

Hmm, not sure.

If we want to use colours on windows then we need to be calling this. Is that right?

> I didn't see the terminal.Start function being called on a simple tree command.

`Start` is called automatically if `terminal.Write()` is used.

> And when running rclone tree and piping to file `> tree.txt` in command prompt, the resulting text file does include the ansi escape codes, wasn't the intention to prevent this?

I think the problem is that tree is not using `terminal.Out` but using `os.Stdout`

I suspect if we called `terminal.Start` and used `terminal.Out` in place of `os.Stdout` in `cmd/tree/tree.go` then it would work better and maybe this patch wouldn't be needed.

Do you fancy giving it a go?

albertony:
Hm..

> I suspect if we called `terminal.Start` and used `terminal.Out` in place of `os.Stdout` in `cmd/tree/tree.go` then it would work better

I guess that is the correct fix for this case:

> when running rclone tree and piping to file > tree.txt in command prompt, the resulting text file does include the ansi escape codes, wasn't the intention to prevent this?

I can take a look at that.

Regardless, when running tree in terminal, without piping to file, this would not make a difference for users seeing the "strange" output `←[1;34m` today, would it? I think setting the console mode explicitly, like this PR does, is something we would want to do to avoid this. So we need both fixes, for the two cases.

> If we want to use colours on windows then we need to be calling this. Is that right?

Well... To be *sure* it works we need this. In many cases it is not needed because it is already default, as controlled with the `VirtualTerminalLevel` registry value. I originally thought it was always the default in recently new versions of Windows, but it seems not. This part confuses me, but haven't had the time to dig deeper (busy times...). When I googled this before, it seemed that explicitly setting console mode was not entirely uncommon for console-oriented applications, even in recently updated ones, some of them written in go.

ncw:
Ok!

Do you want to merge this as-is or put some more stuff on top?

albertony:
I think I've understood the terminal package now, and added a commit which makes the tree command utilize its functionality to properly handle coloring.

TLDR;

Rclone's lib/terminal heavily relies on third party github.com/mattn/go-colorable. When colors are enabled in the terminal library, on Windows, go-colorable uses the Windows Console API. It first checks if the console mode ENABLE_VIRTUAL_TERMINAL_PROCESSING is set, and if it is then this means the console handles VT100 escape sequences natively, and go-colorable will not do anything except just relay text written to it into os.Stdout. But if the console does not have the ENABLE_VIRTUAL_TERMINAL_PROCESSING mode, then go-colorable handles the writing of text directly to the console using Console API functionality (GetConsoleScreenBufferInfo, SetConsoleTextAttribute, etc). It decodes any escape sequences found in the text, and explicitly writes the text to console instead of sending it to stdout, and are then able to set colors and other things using the API.

When colors are disabled in the terminal library, or when os.Stdout is not associated with a terminal (e.g. redirected to file), then go-colorable will strip off any escape sequences in the text, before sending the remaining text directly to the os.Stdout.

This means: Using lib/terminal for output, either through terminal.Write, which automatically calls terminal.Start(), or explicitely calling terminal.Start() and then sending text to terminal.Out intead of os.StdOut, then you should always get colored output to your Windows console - regardless of the console mode setting!

To enable native VT100 escape handling in the Windows console, the go-colorable provides utility function EnableColorsStdout which uses the Console API to set console mode ENABLE_VIRTUAL_TERMINAL_PROCESSING. Further use of lib/terminal and go-colorable will then detect this and simply send text unchanged with escape codes to os.Stdout, without adding anything extra, but writing through lib/terminal still provides the feature that colors can easily be disabled with an option that makes it strip off escape codes automatically.

Rclone now has global option `--color`, which can be set to AUTO, NEVER or ALWAYS, and this configures how lib/terminal, and then go-colorable, should treat color escape sequences as described above.

The tree command in rclone, using the third party github.com/a8m/tree, has its own option which controls if it should write color escape sequences in the first place. When using the lib/terminal this can always be enabled, since we can best let it (and go-colorable) decide what to do with the escape sequences.

albertony:
Now my original commit, with `EnableColorsStdout` is not really needed - as you @ncw predicted:

> I suspect if we called terminal.Start and used terminal.Out in place of os.Stdout in cmd/tree/tree.go then it would work better and maybe this patch wouldn't be needed.

Still, I think perhaps it would be better. To get native Windows handling of coloring, instead of relying on the go-colorable doing a lot of stuff with the Console API. Probably faster (if it matters), and less risk of bugs/inconsistencies. Any thoughts? I will think more about it myself, a couple of days, if there are any implications etc, so if we are getting close to a release and wants the other (new) commit included I can split and merge that - just let me know.

ncw (approved):
LGTM if you want to merge - thank you :-)

albertony:
Great. I've polished it a bit, with comments etc, but I think I'll just merge it as it is now.

We *could* have done the call to `terminal.EnableColorsStdout()` in the `terminal.Start` instead, to only affect "terminal-aware commands" like `tree`, instead of in the common init for all commands where it is now. But I think it is probably fine to just leave it where it is, which ensures *any* escapes emitted to stdout are being processed - on windows like it would on linux. Maybe... 😆

ncw:
Sounds perfect :-)

PR #6762, opened 2023-02-12
Update s3.go

MuhammadAbrar1984:

Setting default value of useMultipartEtag  false for idrive. It seems like Md5 calculated by Idrive e2 for the encrypted bucket is different from the one calculated by rclone. Probably it's considering encrypted data only to calculate Md5.

https://forum.rclone.org/t/issues-with-idrive-e2-corrupted-transfers/36085

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -2791,6 +2791,7 @@ func setQuirks(opt *Options) {
 		useMultipartEtag = false // untested
 	case "IDrive":
 		virtualHostStyle = false
+		useMultipartEtag = false //Etag calculated by Idrive e2 for the encrypted bucket is different from the one calculated by rclone.
 	case "IONOS":
 		// listObjectsV2 supported - https://api.ionos.com/docs/s3/#Basic-Operations-get-Bucket-list-type-2
 		virtualHostStyle = false


ncw:
I think

    server_side_encryption = aws:kms

Is the correct fix for this in the config.

Encrypted files done have correct md5sums.

What do you think?

MuhammadAbrar1984:
> I think
> 
> ```
> server_side_encryption = aws:kms
> ```
> 
> Is the correct fix for this in the config.
> 
> Encrypted files done have correct md5sums.
> 
> What do you think?

you are right @ncw  but this adds unnecessary confusion. Moreover, it's not the right way to check the integrity of the file uploaded. Rather checksums can be sent in CompleteMultiparUpload request header and the remote side should validate that on the received object to check the integrity of the data. https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

Users would still be able to enable this option if absolutely required.

ncw:
> you are right @ncw but this adds unnecessary confusion.

You need to set `server_side_encryption = aws:kms` when using server side encryption with rclone otherwise anything to do with hashes doesn't work. Uploading files smaller than the upload cutoff for chunked files will fail I think, as will `rclone hash`, `rclone check` etc.

> Moreover, it's not the right way to check the integrity of the file uploaded. Rather checksums can be sent in CompleteMultiparUpload request header and the remote side should validate that on the received object to check the integrity of the data. https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

Rclone does this already. Double checking the ETag is a little bit of extra security for the user.

> Users would still be able to enable this option if absolutely required.

Setting aws:kms in the config is better I think.

ncw:
I'll close this now since the IDrive support is correct:

```
$ rclone copy -vv --s3-upload-cutoff 0 10M idrivee2:rclone/
2023/03/24 11:37:55 DEBUG : rclone: Version "v1.63.0-beta.6867.abbf80afb.fix-webdav-keepalive" starting with parameters ["rclone" "copy" "-vv" "--s3-upload-cutoff" "0" "10M" "idrivee2:rclone/"]
2023/03/24 11:37:55 DEBUG : Creating backend with remote "10M"
2023/03/24 11:37:55 DEBUG : Using config file from "/home/ncw/.rclone.conf"
2023/03/24 11:37:55 DEBUG : fs cache: adding new entry for parent of "10M", "/tmp"
2023/03/24 11:37:55 DEBUG : Creating backend with remote "idrivee2:rclone/"
2023/03/24 11:37:55 DEBUG : idrivee2: detected overridden config - adding "{Wgtlt}" suffix to name
2023/03/24 11:37:55 DEBUG : Resolving service "s3" region "us-east-1"
2023/03/24 11:37:55 DEBUG : fs cache: renaming cache item "idrivee2:rclone/" to be canonical "idrivee2{Wgtlt}:rclone"
2023/03/24 11:37:55 DEBUG : 10M: Need to transfer - File not found at Destination
2023/03/24 11:37:56 DEBUG : 10M: multipart upload starting chunk 1 size 5Mi offset 0/10Mi
2023/03/24 11:37:56 DEBUG : 10M: multipart upload starting chunk 2 size 5Mi offset 5Mi/10Mi
2023/03/24 11:38:00 DEBUG : 10M: Multipart upload Etag: a7d414b9133d6483d9a1c4e04e856e3b-2 OK
2023/03/24 11:38:00 DEBUG : 10M: md5 = f1c9645dbc14efddc7d8a322685f26eb OK
2023/03/24 11:38:00 INFO  : 10M: Copied (new)
2023/03/24 11:38:00 INFO  : 
```

L1Cafe:
Sorry but this is not acceptable.

I went through the configuration of rclone expecting this to work out of the box, and I stumbled upon this to find out I have to edit the configuration manually anyway.

rclone should ask whether the buckets / remote added have "default encryption" enabled in the IDrive configuration wizard. A boolean with yes/no would quickly fix this, as I would have said "yes", and I would have never faced such problem.

Instead, I had to stumble upon a relatively obscure thread instructing me to manually edit the settings: https://forum.rclone.org/t/issues-with-idrive-e2-corrupted-transfers/36085/9

If possible I will make a merge request later to add this step to the wizard, but I have zero experience with rclone, so we'll see.

PR #6765, opened 2023-02-14
pikpak: add support for PikPak

### LINKED ISSUE #6429: add support for pikpak
rclone v1.59.1
- os/version: Microsoft Windows 10 Pro 21H1 (64 bit)
- os/kernel: 10.0.19043.1889 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.18.5
- go/linking: static
- go/tags: cmount

I also use the Linux version: 
rclone v1.59.1
- os/kernel: 5.4.0-84-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.18.5
- go/linking: static
- go/tags: none

requesting for a new Storage support called pikpak. The post on forum wasn't from me, but I also want to suggest it.
Their website is: https://mypikpak.com/

Add a new storage support. They seems like have no api for it yet, but [the project mentioned on the forum](https://github.com/alist-org/alist) might be help. It said they [support pikpak](https://alist.nn.ci/guide/) file listing.
The pikpak website is: https://mypikpak.com/

#### How to use GitHub

wiserain:

Add support for a new backend `pikpak`

Closes #6429 
https://forum.rclone.org/t/advice-in-writing-a-new-backend/36022

#### Reference

API is not publicly available. Implementation for `uploadByResumable` is mostly from https://github.com/alist-org/alist/tree/main/drivers/pikpak

#### Test results

Due to the daily limit for creating objects, `test_all` is carried out with `maxtries=1` and disect sub-tests one bye one.

```bash
~/go/bin/test_all -backends pikpak -maxtries 1
```

All tests get passed after manual retries except

`TestCheckSum`, `TestHashSums/Md5`, `TestSyncWithTrackRenames`, and `TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType`

I believe the first three are due to the empty md5 sums for uploaded files. The last is because it somehow returns wrong value as follows:

```
=== RUN   TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType
    fstests.go:1357: 
        	Error Trace:	/home/me/wiserain/rclone/backend/pikpak/fstests.go:1357
        	Error:      	Not equal: 
        	            	expected: "text/plain"
        	            	actual  : "application/octet-stream"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-text/plain
        	            	+application/octet-stream
        	Test:       	TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType
```

I have checked files with a proper extension like '.txt' return mime as expected.

For more detail log, please find attached below:
[2023-02-14-193931.zip](https://github.com/rclone/rclone/files/10732425/2023-02-14-193931.zip)
[2023-02-14-disects.zip](https://github.com/rclone/rclone/files/10732481/2023-02-14-disects.zip)

#### Checklist


ncw:
> API is not publicly available.

:-(

Are you willing to commit to maintaining this backend indefinitely?

> Due to the daily limit for creating objects, `test_all` is carried out with `maxtries=1` and disect sub-tests one bye one.

Sounds painful. Is this likely to work in the integration tester? This essentially runs `test_all` once a day.

See https://pub.rclone.org/integration-tests/current/

> ```shell
> ~/go/bin/test_all -backends pikpak -maxtries 1
> ```
> 
> All tests get passed after manual retries except
> 
> `TestCheckSum`, `TestHashSums/Md5`, `TestSyncWithTrackRenames`, and `TestIntegration/FsMkdir/FsPutFiles/ObjectMimeType`
> 
> I believe the first three are due to the empty md5 sums for uploaded files.

Why are the MD5 sums empty? Will they be filled in later, or are they empty sometimes?

Empty MD5 sums can be OK in some circumstances.

> The last is because it somehow returns wrong value as follows:

What you can do is stop declaring `WriteMimeType`. Do exactly what you are doing at the moment as a best effort. This will stop the tests moaning.

ncw:
> > Are you willing to commit to maintaining this backend indefinitely?
> 
> OK

Great. Once we've merged the PR I'll put you down as the maintainer and add you to the project if you want.

> > Sounds painful. Is this likely to work in the integration tester? This essentially runs `test_all` once a day.
> 
> Found that `-maxtries 1` and manual retries were necessary in the early stage of development. Major errors are sorted out and it's okay without it now. Please find the latest result of executing `test_all -backends pikpak` below.
> 
> image

Very good - well done :-)

> > Why are the MD5 sums empty? Will they be filled in later, or are they empty sometimes?
> > Empty MD5 sums can be OK in some circumstances.
> 
> Unfortunately no idea with this. Files uploaded by users are missing md5 at all but some of remotely added files are with it. Note that pikpak supports a remote download like premiumize.me and users usually add files to their own drives using this. Pikpak dev team doesn't really seem to care about this either. I guess this will not be changed in near future.
> 
> If empty md5 hash is critical and not qualified(or early) to be integrated to rclone right now, that's okay. Just let me know.

Missing MD5 sums won't hurt anything other than `rclone check`. Rclone will use the MD5s if available and it won't if not so I think you are good here.

> > What you can do is stop declaring `WriteMimeType`. Do exactly what you are doing at the moment as a best effort. This will stop the tests moaning.
> 
> `WriteMimeType` is not declared in the code. Do I need to explicitly set it to `false` like you did it for `yandex.go`? I though it will be `false` by default if not specified.

I just looked through the code of that test. Sorry `WriteMimeType` is not for that test! The file that is uploaded is a `.txt` file which really should have mime type `text/plain`

The fact that it doesn't is possibly a bug... If you upload `.txt` files with the web interface do they get the correct mime type?

I don't think there are any show stopping problems here so it looks like we can merge eventually :-)

Things to do

1. I need an account for the integration tester. Can I test with a free account? Or can you help me get an account?
2. I need to do a review of the code. I will do this now - you'll see review comments when I'm done.

backend/pikpak/api/types.go
```diff
@@ -0,0 +1,530 @@
+// Package api has type definitions for pikpak
+//
+// Converted from the API docs with help from https://mholt.github.io/json-to-go/
```
ncw, line 3: Is this comment true?

backend/pikpak/helper.go
```diff
@@ -0,0 +1,235 @@
...
+		_ = os.Remove(tempFile.Name()) // Delete the file - may not work on Windows
+
+		// clean up the file after we are done downloading
+		cleanup = func() {
+			// the file should normally already be close, but just to make sure
+			_ = tempFile.Close()
+			_ = os.Remove(tempFile.Name()) // delete the cache file after we are done - may be deleted already
+		}
+
+		// copy the ENTIRE file to disc and calculate the MD5 in the process
```
ncw, line 212: SHA1?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+	"github.com/rclone/rclone/lib/oauthutil"
+	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/random"
+	"github.com/rclone/rclone/lib/rest"
+	"golang.org/x/oauth2"
+)
+
+// Constants
+const (
+	rcloneClientID              = "YcrttD06T9PIkqAY"
```
ncw, line 68: Where did these IDs come from?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+			opt.Advanced = true
+		}
+		opts = append(opts, opt)
+	}
+	return opts
+}
+
+// pikpakAutorize retrieves OAuth token using user/pass and save it to rclone.conf
+func pikpakAuthorize(ctx context.Context, opt *Options, name string, m configmap.Mapper) error {
+	pass := obscure.MustReveal(opt.Password)
```
ncw, line 108: You want to use `obscure.Reveal` here `MustReveal` will panic if the password doesn't decrypt which is user unfriendly!

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		}, {
+			Name:     "hash_memory_limit",
+			Help:     "Files bigger than this will be cached on disk to calculate hash if required.",
+			Default:  fs.SizeSuffix(10 * 1024 * 1024),
+			Advanced: true,
+		}, {
+			Name:     config.ConfigEncoding,
+			Help:     config.ConfigEncodingHelp,
+			Advanced: true,
+			Default: (encoder.EncodeCtl |
```
ncw, line 187: How sure are you of these encodings?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		return true, err
+	}
+	authRetry := false
+
+	if resp != nil && resp.StatusCode == 401 && len(resp.Header["Www-Authenticate"]) == 1 && strings.Contains(resp.Header["Www-Authenticate"][0], "expired_token") {
+		authRetry = true
+		fs.Debugf(nil, "Should retry: %v", err)
+	}
+
+	// traceback to possible api.Error wrapped in err, and re-authorize if necessary
```
ncw, line 349: Is this block of code needed? The oauthutil code should do this for you?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		}
+		return nil, err
+	}
+
+	// checking whether fileObj with name of leaf exists in dirID
+	trashed := "false"
+	if f.opt.TrashedOnly {
+		trashed = "true"
+	}
+	found, err := f.listAll(ctx, dirID, api.KindOfFile, trashed, func(item *api.File) bool {
```
ncw, line 517: If there is a quicker way of finding a given file than listing the parent, then this `readMetaDataForPath` should implement it. It makes quite a difference to performance if such a call is available.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+func (f *Fs) Mkdir(ctx context.Context, dir string) error {
+	_, err := f.dirCache.FindDir(ctx, dir, true)
+	return err
+}
+
+// About gets quota information
+func (f *Fs) About(ctx context.Context) (usage *fs.Usage, err error) {
+	info, err := f.getAbout(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get drive#quota: %w", err)
```
ncw, line 741: `drive#quota`? Typo?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		o, err := f.NewObject(ctx, remote)
+		if err != nil {
+			return "", err
+		}
+		id = o.(*Object).id
+	}
+	req := api.RequestShare{
+		FileIds:        []string{id},
+		ShareTo:        "publiclink",
+		ExpirationDays: -1,
```
ncw, line 772: You can calculate the `ExpirationDays` from the `expire` duration passed in I think.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+
+	if srcParentID != dstParentID {
+		// Do the move
+		err = f.moveObjects(ctx, []string{srcID}, dstParentID)
+		if err != nil {
+			return fmt.Errorf("couldn't dir move: %w", err)
+		}
+	}
+
+	// Can't copy and change name in one step so we have to check if we have
```
ncw, line 937: This can cause data loss if there was an existing object with the old name at the new position.

Is that true?

I thought the unit tests tested for this...

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+}
+
+// copy objects
+func (f *Fs) copyObjects(ctx context.Context, IDs []string, dirID string) (err error) {
+	if len(IDs) == 0 {
+		return nil
+	}
+	req := api.RequestBatch{
+		Ids: IDs,
+		To:  &map[string]string{"parent_id": parentIDForRequest(dirID)},
```
ncw, line 1034: We don't usually use pointers to `map` as they are a reference type and can be `nil`?

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("copy: couldn't locate copied file: %w", err)
+		}
+	}
+	return dstObj, nil
+}
+
+// MergeDirs merges the contents of all the directories passed
+// in into the first one and rmdirs the other directories.
+func (f *Fs) MergeDirs(ctx context.Context, dirs []fs.Directory) (err error) {
```
ncw, line 1106: Can pikpak have duplicate file names? So the same file name duplicated in the same directory?

If so you need to declare `DuplicateFiles` in Features. If not you don't need this method and you can remove it.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+
+func (f *Fs) upload(ctx context.Context, in io.Reader, leaf, dirID string, size int64, options ...fs.OpenOption) (*api.File, error) {
+	// Calculate sha1sum; grabbed from package jottacloud
+	//
+	// unwrap the accounting from the input, we use wrap to put it
+	// back on after the buffering
+	var wrap accounting.WrapFn
+	in, wrap = accounting.UnWrap(in)
+	var cleanup func()
+	sha1Str, in, cleanup, err := readSHA1(in, size, int64(f.opt.HashMemoryThreshold))
```
ncw, line 1227: Note that some source objects (eg files on the local disk) will have an SHA1 available so you don't need to read it.

You should try to read it first like the jottacloud code does

https://github.com/rclone/rclone/blob/9baa4d1c3c5b592510af75495c718faadeaf9cd0/backend/jottacloud/jottacloud.go#L1853-L1867

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+	defer cleanup()
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate SHA1: %w", err)
+	}
+	// Wrap the accounting back onto the stream
+	in = wrap(in)
+
+	// determine upload type
+	uploadType := api.UploadTypeResumable
+	if size >= 0 && size < int64(5*fs.Mebi) {
```
ncw, line 1237: This limit is traditionally called `upload_cutoff` if you need to make it configurable.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+}
+
+// PutUnchecked the object into the container
+//
+// This will produce an error if the object already exists.
+//
+// Copy the reader in to the new object which is returned.
+//
+// The new object may have been created if an error is returned
+func (f *Fs) PutUnchecked(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) (fs.Object, error) {
```
ncw, line 1287: If duplicates aren't possible, you don't need this method at all.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		fs.Errorf(f, "%v", err)
+	}
+	if r.Errors != 0 {
+		return r, r
+	}
+	return r, nil
+}
+
+var commandHelp = []fs.CommandHelp{{
+	Name:  "getid",
```
ncw, line 1426: I wouldn't have thought you would need this - you can do this with `rclone lsf` or `rclone lsjson` which can show the IDs of files and directories.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1749 @@
...
+		return fmt.Errorf("failed to rename temp object: %w", err)
+	}
+	return o.setMetaData(info)
+}
+
+// Check the interfaces are satisfied
+var (
+	// _ fs.ListRer         = (*Fs)(nil)
+	// _ fs.ChangeNotifier  = (*Fs)(nil)
+	// _ fs.PutStreamer     = (*Fs)(nil)
```
ncw, line 1731: Not having PutStream is sad - this means that `rclone mount` will need `--vfs-cache-mode writes` or `full` in order to upload files amongst other things.

fstest/test_all/config.yaml
```diff
@@ -302,6 +302,18 @@ backends:
  - backend:  "pcloud"
    remote:   "TestPcloud:"
    fastlist: true
+ - backend:  "pikpak"
+   remote:   "TestPikPak:"
+   fastlist: false
+   oneonly:  true
+   ignore:
+    #  fs/operations
```
ncw, line 310: These exclusions seem reasonable to me given the limitations of the platform.

ncw (commented):
I've given the code a one over - see comments inline.

It looks like you've done an extremely thorough job - well done :-)

backend/pikpak/pikpak.go
ncw: Hmm, not ideal. I guess there is no way of registering an rclone app ID yet?

backend/pikpak/pikpak.go
ncw: The `.csv` is the easiest thing to look at here. Look at the `Get L/M/R` and `List L/M/R` colums. This is Get a file with the character on the Left/Middle/Right or List it. If any of these fails it needs some kind of escaping. Some things can be escaped for only use on Left/Right.

Looks like you need to encode
- control chars
- " * : < > ? \ |
- InvalidUtf8
- LeftSpace, RightSpace, RightPeriod (LeftPeriod is OK I think)

I think you can lose a few things from your encoding list as `EncodeWin` covers some of the other ones. I recommend you don't use `EncodeWin` as this isn't a windows platform I don't think.

Note that EncodeLeftCrLfHtVt and EncodeRightCrLfHtVt should be covered by EncodeCtl

`~` seemed to be OK so you don't need EncodeLeftTilde

backend/pikpak/pikpak.go
ncw: Oh, OK.

backend/pikpak/pikpak.go
ncw: That is fine if you don't want to make it configurable. I just thought I'd mention it :-)

backend/pikpak/pikpak.go
ncw: Yes `rclone rcat` is a good way of testing. Note that it will stream larger files to disk unless you advertise `PutStream`.

backend/pikpak/pikpak.go
ncw: Check the drive backend for a backend which can have duplicate file names.

If you call PutUnchecked on drive twice you will get two files with the same name, whereas if you call Put twice you will get one file and the second file updates the first file. This is what users expect!

Are duplicates possible?

ncw:
> > The fact that it doesn't is possibly a bug... If you upload `.txt` files with the web interface do they get the correct mime type?
> 
> Yes.
> 
> Regarding an account for the integration tester, it is required a paid subscription. Let me know how I can help you for this.

Do you know anyone at PikPak who would give rclone a free account for integration testing?

ncw:
> > Do you know anyone at PikPak who would give rclone a free account for integration testing?
> 
> No I think I have to make one for test and give you auth info (user and pass). How can I send you the info?

You can email me nick@craig-wood.com - put a link to the issue in the body so I have some context - thanks

backend/pikpak/pikpak.go
ncw: If duplicates aren't possible then you can remove this method (or merge it into Put/Putstream like it is in the other backends).

You can also remove MergeDirs

backend/pikpak/pikpak.go
ncw: OK great!

ncw:
Got the email thanks. Will test the code and review tomorrow :-)

ncw:
> Hm... I have no idea on one extra file things.

I'm not too worried about that. 1 file in 1000 is no big deal!

> Btw, I have investigated the `can't download - no link to download` whether the retry may resolve the issue using the pacer like below,
> 
> ```go
> 	err = f.pacer.Call(func() (bool, error) {
> 		resp, err = f.srv.CallJSON(ctx, &opts, nil, &info)
> 		if err == nil {
> 			if info.Phase != api.PhaseTypeComplete {
> 				// could be pending right after file is created/uploaded.
> 				return true, nil
> 			} else if info.Size != 0 && info.WebContentLink == "" {
> 				fs.Debugf(nil, ">>> Retrying as no downLink")
> 				// ensure download link except for zero-byte objects
> 				return true, nil
> 			}
> 		}
> 		return f.shouldRetry(ctx, resp, err)
> 	})
> ```
> 
> This is actually helpful to the case that the link is provided eventually but there are other cases it's not so we have to wait until low-level retries are exhausted. I have also confirmed that unnecessary retries can possibly cause timeout to the integration test.
> 
> Are there any fancy ways to limit the number of low-level retries in this special case only? Maximum three retries are enough according to the test. Or any other tips?

If you want to do fewer low level retries then return from the function with `return false, nil /*or err*/` here `false` means do not retry.

You could also put a `time.Sleep()` in if you think that would be helpful to slow down the retry time because you need to wait.

ncw:
Let me know when you are ready for me to re-test - I think we are nearly there :-)

ncw:
502 bad gateway errors is most likely a problem at their end. You get this error from intermediary proxies. So maybe the storage layer is timing out for some reason.

Though it is super weird having the correct info stuck on the end.

I'd really expect a 502 HTTP status also but I suspect you are not getting those.

If this was a provider with a published API I'd report a bug. I don't know whether you have sufficient contacts within PikPak to report a bug?

Does this only happen intermittently? Maybe we can just ignore it and hope they fix the problem?

By the look of the failing tests this is due to the `Range` header.  I see you are calling `FixRangeOption` which is good - some providers don't like indefinite Range requests. See if you can figure out what about the Range header it doesn't like.

I still suspect a provider bug through rather than anything we are doing. Possibly there is a workaround!

ncw:
> Most of the 50x gateway errors have been resolved. 

Great.

> However, PikPak has recently changed its policy to refuse concurrent requests for a file. As a result, multi-threaded copying always fails even if we get a new link for each thread.

:-(
 
> Fortunately, there is another download link we can take advantage of. Pikpak prepares extra links for online media playback that definitely allow multiple concurrent requests for a single file. Let’s call it the media link. The problem is that the media link is not always available, so we have to enable/disable multi-thread copying adaptively based on the availability of the media link for individual objects.

What sort of objects have a media link? These are probably the sort of objects users will want to multithread download (if they are large).

This will break `--vfs-cache-mode full` also which depends on  opening a new connection for each reader of the file and each seek can potentially open them too.

> Is this possible? The current logic seems to call `doMultiThreadCopy()` before `Fs.Open()` to get links. It doesn’t look feasible to me, but I’m asking for your opinion just in case. Alternatively, we can simply disable it completely by setting `ci.MultiThreadStreams=1`.

Setting the global config might work but there can be more than one config in use and it doesn't cover the `--vfs-cache-mode` problem.

We could introduce a new Feature flag in the backends say `NoMultithread` but that is either on or off.

What we could do is introduce a new object Method say `CanMultithread` and assert the existense of it in the multipart streaming code.

Not sure what to do about the VFS cache though.

ncw:
I've run out of time to review today (vacation backlog!) but I'll try to do it tomorrow.

backend/pikpak/pikpak.go
```diff
@@ -0,0 +1,1707 @@
...
+	opt := new(Options)
+	if err := configstruct.Set(m, opt); err != nil {
+		return nil, err
+	}
+
+	root := parsePath(path)
+
+	// overrides global `--multi-thread-streams` by local one
+	ci := fs.GetConfig(ctx)
+	ci.MultiThreadStreams = opt.MultiThreadStreams
```
ncw, line 442: I don't like this very much! I'd rather it was a Feature we advertise and get the multithread streaming code to adapt.

However its OK for the initial merge - we can fixup later.

ncw (approved):
I think this is looking good now and is ready to merge.

I'd like to fix up the multipart streams thing, but we can do that later.

What do you think - ready to merge?

dhtpdud:
I really appreciate your efforts.  
Guys... you don't know how long I've been waiting for this.  
  
thanks you again...

dhtpdud:
oh... sry... its not merge yet...

ncw:
I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.63 - thank you :-)

Check out the integration tests tomorrow to see how things are working: https://pub.rclone.org/integration-tests/current/

I have sent you @wiserain an invite to be an rclone team member on GitHub.

Can you add yourself to MAINTAINERS.md as maintainer of the PikPak backend please?

I made a new issue about the multithreading here: https://github.com/rclone/rclone/issues/6915 - can you have a go at that?

Thank you

djniche:
When I copy from google drive to pikpak in bulk, there may be some files with inconsistent hash values.
The pikpak file system supports fast uploads. The server has a unique hash value recorded, but some of the files will have one hash value when they are first uploaded, and then after about 10-60 minutes, they will suddenly have a different (correct) hash value.
If I delete the file after getting the hash first, the hash value will be permanently invalid, as if the upload did not generate the correct value, but if I upload it via webdav alist:/pikpak, there is no error.

Sorry, I'm using the translator, the error only occurs in some of the files I just uploaded.

ncw:
@djniche if you have found a bug, please open a [new issue](https://github.com/rclone/rclone/issues/new/choose) with full details - thanks. If you want to discuss with other users first please [post on the rclone forum](https://forum.rclone.org/) about this - thank you.

djniche:
> @djniche I've just noticed hash-related problems and making some fixes to resolve it. As ncw recommended, can you please open a new issue?

Okay, I'll create a new issue later.

PR #6782, opened 2023-02-20
all: Apply codeql fixes

Skylion007:

I split off the bugfixes from #6555 to try to get them into a rclone a bit faster. This fixes two issues, one is parseInt trying to store a 64 width int into a 32 width one and some MustCompile regex strings for URLs that have some bugs that could be exploited.
#### Checklist

### DIFF
--- a/backend/uptobox/uptobox.go
+++ b/backend/uptobox/uptobox.go
@@ -214,7 +214,7 @@ func NewFs(ctx context.Context, name string, root string, config configmap.Mappe
 
 	client := fshttp.NewClient(ctx)
 	f.srv = rest.NewClient(client).SetRoot(apiBaseURL)
-	f.IDRegexp = regexp.MustCompile("https://uptobox.com/([a-zA-Z0-9]+)")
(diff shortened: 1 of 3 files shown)


backend/uptobox/uptobox.go
```diff
@@ -214,7 +214,7 @@ func NewFs(ctx context.Context, name string, root string, config configmap.Mappe
 
 	client := fshttp.NewClient(ctx)
 	f.srv = rest.NewClient(client).SetRoot(apiBaseURL)
-	f.IDRegexp = regexp.MustCompile("https://uptobox.com/([a-zA-Z0-9]+)")
+	f.IDRegexp = regexp.MustCompile("^https://uptobox\\.com/([a-zA-Z0-9]+)")
```
ncw, line 217: This isn't quite the same regexp as it was - the original didn't have a `^` at the start?

This would look much nicer in backquotes now it has backslashes in!

ncw (commented):
Thanks for doing this.

See inline comment :-)

backend/uptobox/uptobox.go
Skylion007: @ncw Done

ncw (approved):

I'll merge now.

PR #6787, opened 2023-02-23
rcd: fix regular expression error when folder name contain square brackets

LSOIVNE:
Fixed object not found error when folder/file name contain square brackets.
Screenshot_20230224_125538

No

#### Checklist

### DIFF
--- a/fs/rc/rcserver/rcserver.go
+++ b/fs/rc/rcserver/rcserver.go
@@ -384,7 +384,7 @@ func (s *Server) serveRemote(w http.ResponseWriter, r *http.Request, path string
 }
 
 // Match URLS of the form [fs]/remote
-var fsMatch = regexp.MustCompile(`^\[(.*?)\](.*)$`)
+var fsMatch = regexp.MustCompile(`^\[(.*)](\/.*)`)
 
 func (s *Server) handleGet(w http.ResponseWriter, r *http.Request, path string) {
 	// Look to see if this has an fs in the path


fs/rc/rcserver/rcserver.go
```diff
@@ -384,7 +384,7 @@ func (s *Server) serveRemote(w http.ResponseWriter, r *http.Request, path string
 }
 
 // Match URLS of the form [fs]/remote
-var fsMatch = regexp.MustCompile(`^\[(.*?)\](.*)$`)
+var fsMatch = regexp.MustCompile(`^\[(.*)](\/.*)`)
```
ncw, line 387: I don't think this regexp is correct - it breaks the tests.

Can you explain what you were trying to do and we can see if we can fix it up.

ncw (commented):
See inline - thanks

albertony:
Haven't used the rc/serve and Web GUI functionality much myself, but here are the steps I used to reproduce:
- Create directory with name `[test]`, and a file named `file.txt` in it.
- Serve Web GUI with command: `rclone rcd --rc-no-auth --rc-web-gui`
- Navigate Web GUI: Explorer tab > your remote > directory `[test]`
   - I use local disk for testing, and have a remote configured called "local" which I select here.
- Click the file `file.txt` to start download.
- Server console has error: `ERROR : rc: "]/file.txt": error: failed to find object: object not found`

fs/rc/rcserver/rcserver.go
albertony: One test that fails is using an url such as `[remote:path]file.txt`, i.e. no initial slash. This new regex assumes there is always an initial slash and that this therefore can be used to split `[remote:path]/path/to/object` while accepting `]` in path.

ncw:
Yes to fix this properly the client should URL encode the [ ] and things should detect that. This might work already .

Alternatively we could parse the [ ] pairs and ignore inner pairs. This wouldn't fix all the edge cases though.

This is a problem with the web gui by the look of it so we could implement an improved protocol for it.

albertony:
Maybe we should just close this PR and create an issue for it, since there is no response and it turns out to be a different beast anyway?

ncw:
Sounds like a good plan 🙂

Issue #6803, opened 2023-03-03
crypt base32768 file names encoding on case insensitive remotes  (e.g. OneDrive) is not 100% safe


kapitainsky:
For various reasons I was trying to understand base32768 in details and I have suspicion that base32768 encoding rclone is using is not 100% safe on case insensitive remotes like OneDrive.

The reason is that so called safe alphabet used contains both upper and lower case characters.

I have raised the question with @Max-Sum as his go encoding procedure is used in rclone.

https://github.com/Max-Sum/base32768/issues/1

Chances of upper/lower case collisions are extremely low (but not zero) given that 99% of alphabet used does not have  upper and lower cases (asian characters) and it is only used for file names. We would need by chance two different files resulting with base32768 encoding results like:

䏠Ҡ䈠䉀䉠䊀䊠䋀䋠䌀䌠䍀䍠䎀䎠䏀
and
䏠ҡ䈠䉀䉠䊀䊠䋀䋠䌀䌠䍀䍠䎀䎠䏀

In this example Ҡ and ҡ are problematic and as I understand it they are both part of alphabet used by rclone.

Maybe I am wrong but I think it would be helpful for somebody more knowledgeable to double check it.

ncw:
I guess we should work out what the probability of a collision is.

I wrote a little program to figure out how many characters are case folding

https://go.dev/play/p/IEOcELc1yrH

```
Case folded ƀ into Ƀ
Case folded ɀ into Ɀ
Case folded ɠ into Ɠ
Case folded ʀ into Ʀ
Case folded ҡ into Ҡ
Case folded ԁ into Ԁ
Case folded ⴀ into Ⴀ
Case folded ꭰ into Ꭰ
Case folded ꮐ into Ꮐ
Case folded ⰰ into Ⰰ
Case folded ⲁ into Ⲁ
Case folded ⲡ into Ⲡ
Case folded ⳁ into Ⳁ
Case folded ⴀ into Ⴀ
Case folded ꙁ into Ꙁ
Case folded ꝁ into Ꝁ
Case folded ꞁ into Ꞁ
Found 17 case folding characters
```

Note that 16 of these characters only contain either the upper case or lower case character which means that when rclone is generating file names, they will never contain something which can be confused by case folding.

The only problem character is `ⴀ` and `Ⴀ` which does contain both an upper case and lower case version.

[...]

PR #6806, opened 2023-03-04
Add case insensitive alphabet option to base32768 encoding.

kapitainsky:
Fix for https://github.com/rclone/rclone/issues/6803

Add case insensitive alphabet option (base32768ci) to base32768 encoding.

https://github.com/rclone/rclone/issues/6803

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -161,6 +161,8 @@ func NewNameEncoding(s string) (enc fileNameEncoding, err error) {
 		enc = base64.RawURLEncoding
 	case "base32768":
 		enc = base32768.SafeEncoding
+	case "base32768ci":
+		enc = base32768.SafeEncodingCI
 	default:
 		err = fmt.Errorf("unknown file name encoding mode %q", s)
 	}
--- a/backend/crypt/crypt.go
+++ b/backend/crypt/crypt.go
@@ -138,7 +138,11 @@ length and if it's case sensitive.`,
 				},
 				{
 					Value: "base32768",
-					Help:  "Encode using base32768. Suitable if your remote counts UTF-16 or\nUnicode codepoint instead of UTF-8 byte length. (Eg. Onedrive)",
+					Help:  "Encode using base32768. Suitable if your remote counts UTF-16 or\nUnicode codepoint instead of UTF-8 byte length.",
+				},
+				{
+					Value: "base32768ci",
+					Help:  "Encode using base32768 with case insensitive alphabet. Suitable eg. for Onedrive.",
 				},
 			},
 			Advanced: true,
--- a/backend/crypt/crypt_test.go
+++ b/backend/crypt/crypt_test.go
@@ -95,6 +95,28 @@ func TestStandardBase32768(t *testing.T) {
 	})
 }
 
+func TestStandardBase32768Ci(t *testing.T) {
+	if *fstest.RemoteName != "" {
+		t.Skip("Skipping as -remote set")
+	}
+	tempdir := filepath.Join(os.TempDir(), "rclone-crypt-test-standard")
+	name := "TestCrypt"
+	fstests.Run(t, &fstests.Opt{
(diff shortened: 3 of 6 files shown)


Max-Sum:
I think we could obsolete the legacy base32768 option for the CI one, only keeping the old one for compatibility?

kapitainsky:
> I think we could obsolete the legacy base32768 option for the CI one, only keeping the old one for compatibility?

base32768 works fine and nothing is broken with it. Its constraint is case sensitive output hence it is not good choice for e.g. OneDrive. I think making it obsolete would confuse many people...

kapitainsky:
Also FYI I have tested one of my archive  ~200k files on Onedrive stored with base32768 names encryption. I server side copied them to the new folder with base32768ci encryption and then compared both with sync. All works perfectly.

Max-Sum:
> > I think we could obsolete the legacy base32768 option for the CI one, only keeping the old one for compatibility?
> 
> 
> 
> base32768 works fine and nothing is broken with it. Its constraint is case sensitive output hence it is not good choice for e.g. OneDrive. I think making it obsolete would confuse many people...
> 
> 

I think providing too many choices also confuses people. Since base32768CI works on both case sensitive and case intensive environments with no drawbacks compared to base32768, why not just recommend all new users to use CI instead?

kapitainsky:
> > > I think we could obsolete the legacy base32768 option for the CI one, only keeping the old one for compatibility?
> > 
> > 
> > base32768 works fine and nothing is broken with it. Its constraint is case sensitive output hence it is not good choice for e.g. OneDrive. I think making it obsolete would confuse many people...
> 
> I think providing too many choices also confuses people. Since base32768CI works on both case sensitive and case intensive environments with no drawbacks compared to base32768, why not just recommend all new users to use CI instead?

Valid point as well. Let's see what @ncw thinks

kapitainsky:
I wonder why tests fail on macOS Intel... This is exactly system I am using and no issues.

kapitainsky:
ok I tested and I know - macOS 10.11 - Big Sur (used for automated tests) does not like newer UTF characters. New from its perspective. 

It works fine on macOS 12 and 13 where all characters are accepted and code points used instead when there is no glyph. Not all characters can be displayed but OS does not care - the same as Windows or Linux.

macOS 10.11 was released in 2020 and will be deprecated about September this year when macOS 14 is released -  usual macOS cycle

So now question is what to do?

1. wait until macOS 10.11 is history? then all tests will work
2. release it now and make a note in documentation that due to Unicode v14 (released Sep/2021) some older systems might not work

kapitainsky:
OK I have another idea but would need @Max-Sum help.

At the moment we are looking to generates 1024 plus 4 contiguous, aligned blocks of 32 code points. For case insensitive characters they can only be found for Unicode 14. 

but we can change it to 2056 blocks of  16 code points... 

now I can find such blocks for Unicode v9 (2143 suitable blocks), and still even better because excluding all lower and upper case letters I can have 2120 blocks

@Max-Sum what do you think? I am not good at go and I bet you understand algorithm better. base32768 is not ISO standard - we can be creative here without doing anything crazy.

We would make is a bit slower but much better in terms of compatibility.

Max-Sum:
I appreciate your finding but I am afraid that I couldn't implement this by the end of April since I am quite busy now.
Are you willing to implement this?

kapitainsky:
> I appreciate your finding but I am afraid that I couldn't implement this by the end of April since I am quite busy now. Are you willing to implement this?

No worries. Good excuse for me to learn golang. I think I understand algorithm used well enough to do it.

kapitainsky:
Existing base32768 encoding algorithm can only be case insensitive for Unicode v14 or newer.

This algorithm can be modified and according to my research can be made case insensitive for Unicode v6 (released in 2010 ) or newer. It would provide much safer option for older systems.

I will work on this and in the meantime I am closing this PR.

PR #6839, opened 2023-03-13
webdav: add "fastmail" provider for Fastmail Files

### LINKED ISSUE #6837: webdav: support FastMail Files

```
rclone v1.61.1-DEV
- os/version: opensuse-tumbleweed (64 bit)
- os/kernel: 6.2.1-1-default (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.19.4
- go/linking: dynamic
- go/tags: none
```

[FastMail Files](https://www.fastmail.help/hc/en-us/articles/1500000280121-About-Fastmail-file-storage) is a WebDAV implementation that supports SHA1 hashes, but in a different way than the current OwnCloud / NextCloud implementation.

Example PROPFIND response:

```xml
<?xml version="1.0" encoding="utf-8"?>
<D:multistatus xmlns:D="DAV:">
  <D:response>
    <D:href>/example.org/filename</D:href>
    <D:propstat>
      <D:prop>
        <sha1hex xmlns="ME:">86d6c70cfbef2b84394a1c59b23915d40888ebcc</sha1hex>
        <D:getcontentlength>999936</D:getcontentlength>
        <D:quota-available-bytes>7639552</D:quota-available-bytes>
        <D:displayname>filename</D:displayname>
        <D:lockdiscovery/>
        <D:supportedlock>
          <D:lockentry>
            <D:lockscope>
              <D:shared/>
            </D:lockscope>
            <D:locktype>
              <D:write/>
            </D:locktype>
          </D:lockentry>
          <D:lockentry>
            <D:lockscope>
              <D:exclusive/>
            </D:lockscope>
            <D:locktype>
              <D:write/>
            </D:locktype>
          </D:lockentry>
        </D:supportedlock>
        <D:getlastmodified>Sun, 12 Mar 2023 08:11:54 GMT</D:getlastmodified>
        <D:creationdate>2023-03-12T08:11:54Z</D:creationdate>
        <D:getcontenttype>application/octet-stream</D:getcontenttype>
        <D:getetag>a7cc87894e5f0929006f8caf7b18c6f81a5102bd</D:getetag>
        <D:quota-used-bytes>9992360448</D:quota-used-bytes>
        <D:resourcetype/>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
  </D:response>
```

Note the `ME:sha1hex` prop. The value is indeed the `sha1sum` of the file.

It would be nice to implement a new `vendor = fastmail` with support for this kind of SHA1 hash, so that syncing can use the SHA1 instead of relying on mtime. I cannot find any documentation about this property, so I assume it's specific to FastMail's server ("ME" might be "Messaging Engine", another name that FastMail uses for its domains etc). So `vendor = fastmail` should be appropriate.

FastMail also supports the `x-oc-mtime` header and uses it to set the mtime. The `PUT` response contains a `x-oc-mtime: accepted` header if the request set it. So the `useOCMtime` quirk should be set to `true` for it.

FastMail does *not* support the `oc-checksum` header in either of the two ways that [this comment](https://github.com/rclone/rclone/blob/e405ca7733b19b9430afc344324eee3b608c1466/backend/webdav/webdav.go#L1323-L1325) talks about. Specifically, the response contains the `ME:sha1hex` prop regardless of whether the client sends the header or not (ie it doesn't have the NextCloud behavior of using the client-provided value), and the server accepts the `PUT` request even if the client sends an incorrect hash (ie it doesn't have the OwnCloud behavior of using the client-provided value to validate the request body).

#### How to use GitHub

Arnavion:
This provider:

- supports the `X-OC-Mtime` header to set the mtime

- calculates SHA1 checksum server side and returns it as a `ME:sha1hex` prop

To differentiate the new hasMESHA1 quirk, the existing hasMD5 and hasSHA1
quirks for Owncloud have been renamed to hasOCMD5 and hasOCSHA1.

Fixes #6837

---

Add a WebDAV provider for Fastmail Files that understands the way it provides SHA1 hashes, so that it can be used with `--checksum`.

#6837

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Rclone *("rsync for cloud storage")* is a command-line program to sync files and
   * Dreamhost [:page_facing_up:](https://rclone.org/s3/#dreamhost)
   * Dropbox [:page_facing_up:](https://rclone.org/dropbox/)
   * Enterprise File Fabric [:page_facing_up:](https://rclone.org/filefabric/)
+  * Fastmail Files [:page_facing_up:](https://rclone.org/webdav/#fastmail-files)
   * FTP [:page_facing_up:](https://rclone.org/ftp/)
   * Google Cloud Storage [:page_facing_up:](https://rclone.org/googlecloudstorage/)
   * Google Drive [:page_facing_up:](https://rclone.org/drive/)
--- a/backend/webdav/api/types.go
+++ b/backend/webdav/api/types.go
@@ -75,6 +75,7 @@ type Prop struct {
 	Size         int64     `xml:"DAV: prop>getcontentlength,omitempty"`
 	Modified     Time      `xml:"DAV: prop>getlastmodified,omitempty"`
 	Checksums    []string  `xml:"prop>checksums>checksum,omitempty"`
+	MESha1Hex    *string   `xml:"ME: prop>sha1hex,omitempty"` // Fastmail-specific sha1 checksum
 }
 
 // Parse a status of the form "HTTP/1.1 200 OK" or "HTTP/1.1 200"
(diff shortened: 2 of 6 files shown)


ncw:
This looks very useful thanks.

I'll run the ci now 

I'll do a proper review after 1.62 has been released tomorrow, but in the mean time could you drop the auto generated docs from the commit - we generally only update those on release. Thanks

MANUAL.md
ncw, line 1: Can you drop this file from the commit please as it is auto generated.

ncw (commented):
This looks great - thank you.

There is one more auto generated file to drop.

I've run the CI as well which might tell you some more stuff which needs fixing.

MANUAL.md
Arnavion: Done.

For future reference, it would help if you said what exactly are the files I did need to change. You kept saying "drop these files" which is not correct because I would lose the changes I made to them. I had to figure out which files I needed to replicate the changes to.

ncw: Sorry I wasn't clearer. There is a section in the contribution doc about this - I should have linked to that.

Arnavion: Ah yes, I see it now. Sorry about that.

ncw:
That is great - thank you :-) All merged for release in v1.63

PR #6840, opened 2023-03-13
vfs: add refresh flag

### LINKED ISSUE #6830: mount flag to run vfs/refresh recursive=true
n/a

1.6.1

vfs/refresh w/o needing --rc

add a flag to be used on mounts that would warm up the vfs cache.
I don't want to enable --rc because I don't need to except to do this at service startup.

A flag could do this.

#### How to use GitHub

beyondmeat:

Add vfs/refesh recursive=true flag so it can run w/o needing rc on mount.

closes #6830 

#### Checklist

### DIFF
--- a/vfs/vfs.go
+++ b/vfs/vfs.go
@@ -232,10 +232,24 @@ func New(f fs.Fs, opt *vfscommon.Options) *VFS {
 	// removed when the vfs is finalized
 	cache.PinUntilFinalized(f, vfs)
 
+	// Refresh the dircache if required
+	if vfs.Opt.Refresh {
+		go vfs.refresh()
+	}
+
 	// This can take some time so do it after the Pin
 	vfs.SetCacheMode(vfs.Opt.CacheMode)
 
 	return vfs
+}
+
+// refresh the directory cache for all directories
+func (vfs *VFS) refresh() {
+	fs.Debugf(vfs.f, "Refreshing VFS directory cache")
+	err := vfs.root.readDirTree()
+	if err != nil {
+		fs.Errorf(vfs.f, "Error refreshing VFS directory cache: %v", err)
+	}
 }
 
 // Stats returns info about the VFS
--- a/vfs/vfscommon/options.go
+++ b/vfs/vfscommon/options.go
@@ -15,6 +15,7 @@ type Options struct {
 	ReadOnly           bool          // if set VFS is read only
 	NoModTime          bool          // don't read mod times for files
 	DirCacheTime       time.Duration // how long to consider directory listing cache valid
+	Refresh            bool          // refreshes the directory listing recursively on start
 	PollInterval       time.Duration
 	Umask              int
 	UID                uint32
@@ -44,6 +45,7 @@ var DefaultOpt = Options{
 	NoChecksum:         false,
 	NoSeek:             false,
 	DirCacheTime:       5 * 60 * time.Second,
+	Refresh:            false,
 	PollInterval:       time.Minute,
 	ReadOnly:           false,
 	Umask:              0,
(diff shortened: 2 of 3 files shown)


ncw (approved):
Looks great - thank you :-)

I'll merge this now.

ncw:
@beyondmeat can you commit the missing bit of this PR? I think you forgot to commit the `vfs/vfscommon/options.go` file.

beyondmeat:
@ncw will get to this by end of weekend. I didn't finish that part yet.

beyondmeat:
```
--vfs-refresh=true           Refreshes the directory listing recursively
```

I made the default value false. Should be ready for review @ncw

darthShadow:
Can we add an option to do it async too? So perhaps instead of `true`/`false`, `off` (default)/`sync`/`async` ?

beyondmeat:
Is there a reason why anyone would ever want the refresh to be blocking/sync?

darthShadow:
The current behaviour is sync. I would prefer async always but not sure about others so may as well add compatibility with the rc way of doing it which is both sync and sync.

beyondmeat:
I'm unsure if sync would ever be used. It would effectively block the mount until that process is done, which I doubt anyone would ever need.

The use-case for that implies the options recursive=true and _async=true

I would also need another flag to control if it's recursive or not if you just want it to match the --rc functionality, which I don't think it's neccessary b/c this only happens at mount and why wouldn't it be recursive?  The whole point of this flag is to pre-warm the dir cache at mount so you don't need --rc otherwise. Having it sync and non-recursive, I fail to think of a use-case where that would be needed, if it is, they might need --rc anyway.

I'm aware my current implementation is sync, I'm exploring how to make it async, any example code I could reference to help with that. I tried code searching for async but not sure how _async=true works in the code.

darthShadow:
It runs the async jobs in a goroutine: https://github.com/rclone/rclone/blob/9a9ef040e3be057227b14d43c0be1549bf5b1d67/fs/rc/jobs/job.go#L281

I guess you could do the same here by calling the `readDirTree` function in a goroutine.

ncw:
This has a build failure which needs fixing

```
Error: vfs/vfsflags/vfsflags.go:33:59: not enough arguments in call to flags.BoolVarP
	have (*pflag.FlagSet, *bool, string, string, string)
	want (*pflag.FlagSet, *bool, string, string, bool, string)
```

I'd be fine if this was always async so just run the call to read tree in a go routine to do that.

beyondmeat:
@ncw is this good now?

vfs/vfs.go
```diff
@@ -231,6 +231,11 @@ func New(f fs.Fs, opt *vfscommon.Options) *VFS {
 	// with the same remote string we get this one. The Pin is
 	// removed when the vfs is finalized
 	cache.PinUntilFinalized(f, vfs)
+	
+	if vfs.Opt.Refresh {
+		fs.Debugf(f, "Refreshing VFS directory cache")
+		vfs.root.readDirTree()
```
darthShadow, line 237: This is still synchronous. Run it in a goroutine by adding `go` in front of the statement to make it async.

darthShadow:
I have also approved the CI run so let's see if anything else needs to be changed.

beyondmeat:
@darthShadow ok fixed that lint error, please run the checks again. Not sure why it's still complaining about an import.

beyondmeat:
@ncw @darthShadow please check

ncw:
@beyondmeat sorry this dropped off the radar! I rebased this - I'll wait for the CI to go green then merge :-)

ncw:
Thank you :-)

PR #6859, opened 2023-03-16
sftp: also try reading pubkey from keyfile when using ssh-agent

### LINKED ISSUE #6791: sftp: using key_use_agent and key_file together requires private key outside of SSH agent

I have multiple SSH keys in my agent, and want rclone to use one specific key to connect to the host rather than try all of them in sequence. The latter behavior is problematic since hosts can disconnect clients after too many key attempts, and even with regular openssh it's common practice to specify `IdentityKeyFile /path/to/public/key` to limit `ssh` to just that one key.

For example, an openssh config might look like:

```
Host somehost
	IdentityFile ~/identities/arnavion
	IdentitiesOnly yes
```

... where `~/identities/arnavion` is the *public key* file (despite being used as the `IdentityFile`, ie the one that contains `ssh-ed25519 AAAA...`.

On the surface, it looks like rclone supports that behavior with this config:

```
[somehost]
type = sftp
host = somehost
key_file = ~/identities/arnavion
key_use_agent = true
```

Problem 1: [The code](https://github.com/rclone/rclone/blob/56b582cdb9e68c9941e90bae47f6144c45e82a8a/backend/sftp/sftp.go#L784-L788) looks for `~/identities/arnavion.pub`, which in my case does not exist. I assume the code was written with the assumption that it would be used with paths like `~/.ssh/id_ed25519` so that it would use the matching pubkey.

While I would prefer to not have to change my workflow just for rclone's sake, I decided to work around it temporarily with:

```sh
ln -s ~/identities/arnavion ~/.ssh/id_ed25519.pub
```

... and changed the rclone config to:

```
key_file = ~/.ssh/id_ed25519
```

... but this led to...

Problem 2: [The code](https://github.com/rclone/rclone/blob/56b582cdb9e68c9941e90bae47f6144c45e82a8a/backend/sftp/sftp.go#L811-L817) still fails because it still requires `~/.ssh/id_ed25519` to exist, even though it should have no reason to access that file because it ought to use the agent.

So ideally:

1. rclone would use `key_file` as-is instead of appending `.pub` to it. Or maybe repurpose `pubkey_file` for this for the sake of backward-compatibility with `key_file`'s behavior.

2. If (1) is not done, ie rclone continues to append `.pub` to the `key_file` value, then it should not care about whether the `key_file` itself exists or not.

```
rclone v1.61.1-DEV
- os/version: opensuse-tumbleweed (64 bit)
- os/kernel: 6.1.10-1-default (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.19.4
- go/linking: dynamic
- go/tags: none
```

OpenSUSE Tumbleweed x86_64

Personal SSH-accessible server

`rclone lsd somehost:/`

Problem 1:

```
<7>DEBUG : rclone: Version "v1.61.1-DEV" starting with parameters ["rclone" "-vv" "lsd" "somehost:/"]
<7>DEBUG : rclone: systemd logging support activated
<7>DEBUG : Creating backend with remote "somehost:/"
<7>DEBUG : Using config file from "/home/arnavion/.config/rclone/rclone.conf"
Failed to create file system for "somehost:/": failed to read public key file: open /home/arnavion/identities/arnavion.pub: no such file or directory
```

Problem 2:

```
<7>DEBUG : rclone: Version "v1.61.1-DEV" starting with parameters ["rclone" "-vv" "lsd" "somehost:/"]
<7>DEBUG : rclone: systemd logging support activated
<7>DEBUG : Creating backend with remote "somehost:/"
<7>DEBUG : Using config file from "/home/arnavion/.config/rclone/rclone.conf"
Failed to create file system for "somehost:/": failed to read private key file: open /home/arnavion/.ssh/id_ed25519: no such file or directory
```

#### How to use GitHub

Arnavion:
When using ssh-agent to hold multiple keys, it is common practice to configure openssh to use a specific key by setting the corresponding public key as the `IdentityFile`. This change makes a similar behavior possible in rclone by having it parse the `key_file` config as the public key when `key_use_agent` is `true`.

rclone already attempted this behavior before this change, but it assumed that `key_file` is the private key and that the public key is specified in `${key_file}.pub`. So for parity with the openssh behavior, this change makes rclone first attempt to read the public key from `${key_file}.pub` as before (for the sake of backward compatibility), then fall back to reading it from `key_file`.

Fixes #6791

As above.

#6791

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -800,11 +800,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			pubBytes, err := os.ReadFile(keyFile + ".pub")
 			if err != nil {
 				if errors.Is(err, iofs.ErrNotExist) && opt.KeyUseAgent {
-					pubBytes_, err := os.ReadFile(keyFile)
+					pubBytes, err = os.ReadFile(keyFile)
 					if err != nil {
 						return nil, fmt.Errorf("failed to read public key file: %w", err)
 					}
-					pubBytes = pubBytes_
 				} else {
 					return nil, fmt.Errorf("failed to read public key file: %w", err)
 				}
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	iofs "io/fs"
 	"os"
 	"path"
 	"regexp"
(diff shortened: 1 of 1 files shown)


Arnavion:
If you don't care about backward compatibility, we can simplify the code to directly parse the pubkey from `keyFile` in case `opt.KeyUseAgent` is true, ie something like:

```go
var pubKeyFile string
if opt.KeyUseAgent {
	pubKeyFile = keyFile
} else {
	pubKeyFile = keyFile + ".pub"
}
pubBytes_, err := os.ReadFile(keyFile)
if err != nil {
	return nil, fmt.Errorf("failed to read public key file: %w", err)
}

```

ncw (approved):
This looks very nice - thank you

I'll merge it now.

Issue #6857, opened 2023-03-16
Jottacloud: vfs writeback stuck in a failed upload loop with file versioning disabled


cycneuramus:
#### What is the problem you are having with rclone?

VFS writeback on a mounted Jottacloud remote, with file versioning _disabled_, gets stuck in a failed upload loop if there is a problem copying the file to the remote.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.62.2
- os/version: arch (64 bit)
- os/kernel: 6.2.6-arch1-1 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.20.2
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Arch Linux

#### Which cloud storage system are you using? (e.g. Google Drive)

Jottacloud

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

Mount command:

```
rclone mount \
	--allow-other \
	--gid 1000 \
	--uid 1000 \
	--cache-dir $HOME/.cache/rclone/jotta \
	--config $HOME/rclone.conf \
	--log-file $HOME/rclone.log \
	--vfs-cache-mode full \
	--vfs-cache-max-size 5G \
	--ignore-checksum \
	--ignore-size \
	-vv \
	jotta: $HOME/.local/mnt/jotta
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

`echo "test" > .local/mnt/jotta/test.txt`

Succeeds:

```
2023/03/16 16:39:06 DEBUG : /: Lookup: name="test.txt"
2023/03/16 16:39:06 DEBUG : /: >Lookup: node=<nil>, err=no such file or directory
2023/03/16 16:39:06 DEBUG : /: Create: name="test.txt"
2023/03/16 16:39:06 DEBUG : test.txt: Open: flags=O_WRONLY|O_CREATE|O_TRUNC
2023/03/16 16:39:06 DEBUG : test.txt: newRWFileHandle:
2023/03/16 16:39:06 DEBUG : test.txt(0xc00103a840): openPending:
2023/03/16 16:39:06 DEBUG : test.txt: vfs cache: truncate to size=0 (not needed as size correct)
2023/03/16 16:39:06 DEBUG : : Added virtual directory entry vAddFile: "test.txt"
[...]

PR #6860, opened 2023-03-16
Jottacloud: Avoid returning error when option no_versions and remove fail

albertony:

When using option `no_versions` with the Jottacloud backend, uploading a file will first delete any existing - using the permanent ("hard") delete option. In contrast to regular delete, the permanent delete api request will fail with 404 if path is not found. If, for whatever reason, this should occurr, it would lead to upload error:

> Failed to copy: failed to remove old object: error 404: no.jotta.backup.errors.NoSuchFileException: File /xxxxxx/Jotta/Archive/test.txt does not exist (Not Found)

I think this will solve https://github.com/rclone/rclone/issues/6857. I didn't reproduce a real case myself, but provoked it in debugger by deleting file while paused on break point etc. I'm pretty sure the same can happen in real life though, and I think it is similar to the mentioned isseu.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -1838,12 +1838,12 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 		if err == nil {
 			// if the object exists delete it
 			err = o.remove(ctx, true)
-			if err != nil {
+			if err != nil && err != fs.ErrorObjectNotFound {
+				// if delete failed then report that, unless it was because the file did not exist after all
 				return fmt.Errorf("failed to remove old object: %w", err)
 			}
-		}
-		// if the object does not exist we can just continue but if the error is something different we should report that
-		if err != fs.ErrorObjectNotFound {
+		} else if err != fs.ErrorObjectNotFound {
+			// if the object does not exist we can just continue but if the error is something different we should report that
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


albertony:
@ncw : I think this is quite safe to merge!?

I am thinking a bit about a larger case, though; Jottacloud and other backends uses internal caching of object metadata:

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/backend/jottacloud/jottacloud.go#L454

Could it happen that this is out of sync and it leads to more problems similar to the one I'm fixing here? E.g. the remove function in Jottacloud does not set `hasMetaData=false`, so if we somewhere calls `readMetaData` on the same object without `force=true` it would believe the file still exists:

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/backend/jottacloud/jottacloud.go#L1691-L1694

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/backend/jottacloud/jottacloud.go#L1835-L1837

https://github.com/rclone/rclone/blob/0361acbde40815534ab11b4027f8083151ff8640/backend/jottacloud/jottacloud.go#L1709-L1711

Or is this object so short-lived that it shouldn't matter?

ncw (approved):
This code looks great - thank you :-)

Can you re-write the commit message so the first line is from the point of the user as that is what goes in the changelog. So something like

```
jottacloud: fix vfs writeback stuck in a failed upload loop with file versioning disabled

Avoid returning error when option no_versions and remove fail 

Fixes #6857
```

Thanks!

This looks like it is a candidate for a point release also.

ncw:
> @ncw : I think this is quite safe to merge!?

Yes :+1:

> I am thinking a bit about a larger case, though; Jottacloud and other backends uses internal caching of object metadata:

Stuff can always be out of date - that is the nature of the beast. We need to be prepared to deal with it gracefully if it happens.

Returning an error to the user is fine. Getting stuck in a loop is not!

albertony:
> Can you re-write the commit message

Of course. I wasn't sure if it really fixes that issue - but I guess it is safe to assume that.

albertony:
Done.

> This looks like it is a candidate for a point release also.

I think it is safe to include, yes. Not a new bug, though, its probably been around since july 2021 when the no_versions option was implemented.

ncw:
Thank you :-)

I'll pull it into the stable branch also.

Issue #3165, opened 2019-05-03
add 2fa support for mega


bingoxo:
#### What is your current rclone version (output from `rclone version`)?
1.47.0

#### What problem are you are trying to solve?

logging in mega with 2fa

#### How do you think rclone should be changed to solve that?

use same implementation as Megasync : https://github.com/meganz/MEGAsync/blob/master/build/MEGAsync/MEGAsync/megasync.changes
https://github.com/meganz/MEGAsync/issues/233

ncw:
This probably needs work doing in the upstream repository: https://github.com/t3rm1n4l/go-mega

Do you want to implement this?

bingoxo:
@polmr @wizzard @javiserrano , can you guys help with this please ?

pixel8383:
Any progress on this?

ncw:
I'm looking for volunteers for this :-)

alexanderadam:
Just to be sure: there's no upstream issue, right? :thinking:

tddschn:
other mega cli clients don't support 2fa either

MrSAHaseeb:
There is Android App on Play Store called "Autosync for MEGA - MegaSync" that support Mega 2FA.

here is a link: https://play.google.com/store/apps/details?id=com.ttxapps.megasync&hl=en_IN
Contact app developer for help if you may - megasync@metactrl.com

krazyjakee:
@alexanderadam there was no upstream issue so I [created one](https://github.com/t3rm1n4l/go-mega/issues/25).

krazyjakee:
MegaCMD has an undocumented [auth-code](https://github.com/meganz/MEGAcmd/blob/bdf70803e6bfc3b4caadec9c82d9a2786b0094fe/src/megacmdexecuter.cpp#L8028) flag. This looks to be the only library I have found that has that functionality.

It looks like the go-mega library uses some rest api which I can't find any documentation for. So, if that REST api doesn't support 2fa, the only other solution I can see is that either they wrap megaCMD or rclone wraps megaCMD.

I have posted this information on the [go-mega](https://github.com/t3rm1n4l/go-mega/issues/25) issue as well and will report back here with updates.

ncw:
Thanks, I'll take a look at that.

EinarasGar:
Any updates on this issue?

AverageHelper:
[...]

PR #6868, opened 2023-03-19
Added initial support for Mega MFA

thegabriele97:

Adding support for Mega Multi Factor Auth (MFA).

Yes, Issue https://github.com/rclone/rclone/issues/3165

#### Checklist

### DIFF
--- a/backend/mega/mega.go
+++ b/backend/mega/mega.go
@@ -67,6 +67,10 @@ func init() {
 			Required:   true,
 			IsPassword: true,
 		}, {
+			Name:     "mfac",
+			Help:     "Multi-factor authentication code.",
+			Required: false,
+		}, {
 			Name: "debug",
 			Help: `Output more debug from Mega.
 
@@ -109,6 +113,7 @@ Enabling it will increase CPU usage and add network overhead.`,
 type Options struct {
 	User       string               `config:"user"`
 	Pass       string               `config:"pass"`
+	MFAC       string               `config:"mfac"`
 	Debug      bool                 `config:"debug"`
 	HardDelete bool                 `config:"hard_delete"`
 	UseHTTPS   bool                 `config:"use_https"`
@@ -226,7 +231,14 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			})
 		}
 
-		err := srv.Login(opt.User, opt.Pass)
+		// Login to mega
+		if opt.MFAC != "" {
+			// Login with MFA if set in the config by the user
+			err = srv.MultiFactorLogin(opt.User, opt.Pass, opt.MFAC)
+		} else {
+			err = srv.Login(opt.User, opt.Pass)
+		}
+
 		if err != nil {
 			return nil, fmt.Errorf("couldn't login: %w", err)
 		}


ncw (approved):
This looks great - thanks.

I'll merge it now and then we can discuss options.

ncw:
Can you rebase this on master and force push please - hopefully this will clear the lint errors.

Xelbayria:
Have the support implemented into the master yet? I see nothing have happened and it looks like this have been forgotten?

harshsyd:
> Can you rebase this on master and force push please - hopefully this will clear the lint errors.

This seems to be have forgotten? Can nthis be merged please, rclone still fails to support 2FA for MEGA and security wise it's important!

See other threads regarding this https://github.com/rclone/rclone/issues/3165

harshsyd:
Actually thing does not work. I've tried compiling it myself. It logs in, but as soon as the 30 second TOTP from mega changes, we get the error "2023/09/18 15:37:36 Failed to create file system for "harsh:/": couldn't login: The upload target URL you are trying to access has expired. Please request a fresh one".

Maybe it is not saving the given token/session or either not attaching it on every call.

bingoxo:
can you re-take a look into this please ?

harshsyd:
I've decided to troubleshoot this myself but I've never used Go in my life (C++, Java, Python yes). So trying to back track and reverse the code, only 1 login request works. So if I login and do "rclone lsd". It works but only once. Next ones all fail with error "**The upload target URL you are trying to access has expired. Please request a fresh one**".

So looking back at the code for the "go-mega" library (not rclone code), after we login with MultiFactorLogin function. We call the internal function m.login in the go-mega library. That internal function calls MEGA and as much as I can see, calls "m.api_request(req)" where "req" holds the JSON with our MFA and all login data.

So far so good and the result of this api_request function is stored and I guess passed to json.Unmarshal. This is stored in a variable called "res". This is then base64 decoded for "res[0].Key" object perhaps and we decrypt the session id and store the return in "m.sid".

But it seems that since the subsequent calls to any API of Mega fails with error "request a fresh one", does this mean our session ID is not getting stored or perhaps not decrypted properly? I'm not sure as I didn't code that go-mega library, if anyone has an idea?

FranciscoPombal:
I have started looking into this.
There are at least 2 issues:

1. Even if `go-mega` library did not have any issues, this PR would not work in its current form.
The library expects us to supply a 6-digit TOTP code.
However, these codes are only valid for 30 seconds (MEGA uses TOTP for 2FA, with the default TOTP parameters - 6 digits, 30 seconds duration, SHA-1 hash).
Saving one of these values into the config as the 2FA parameter will cause logins to work **only** for the next 30 seconds at most, after which all subsequent logins will fail.

    Instead, we need to either:
    - save the 2FA secret itself to the config and generate the 2FA codes from it just-in-time (with [this library](https://github.com/pquerna/otp/), for example, but it's also easy to do it from scratch).
    - make it possible to have users make 2FA codes available just-in-time from the environment with a mechanism analogous to `export RCLONE_PASSWORD_COMMAND="pass rclone_config"` but for 2FA secrets.

     Personally I prefer the the first alternative.
     It would also be interesting to know what other backends, if any, support 2FA - how do they deal with this problem?
     I believe the MEGA backend should be as consistent as possible with the way things are done elsewhere.

2. The `go-mega` library has initial 2FA support at this point but has issues dealing with 2FA accounts.
For example, as a starting point to tackling this issue in `rclone`, I have submitted a PR attempting to implement tests for the newly added 2FA login ability in the library, but I've run into several problems with it (e.g. the `The upload target URL you are trying to access has expired. Please request a fresh one` error that the comment above mentions): https://github.com/t3rm1n4l/go-mega/pull/51.
Any help with it would be appreciated from someone more knowledgeable about the MEGA API.

ncw:
> It would also be interesting to know what other backends, if any, support 2FA - how do they deal with this problem?

When you do oauth with any of the other providers that support 2FA, as part of the oauth flow you put the 2FA code into their website before control is returned to rclone. This is a much more sensible way of doing it IMHO and rclone doesn't need to deal with the 2FA at all.

I'll comment on your mega PR now too.

FranciscoPombal:
> When you do oauth with any of the other providers that support 2FA, as part of the oauth flow you put the 2FA code into their website before control is returned to rclone. This is a much more sensible way of doing it IMHO and rclone doesn't need to deal with the 2FA at all.

Thanks for the reply.

It doesn't look like Mega supports OAuth though, unfortunately: https://github.com/meganz/sdk/issues/2575.

hangukhiphop:
> Regarding Rclone:
Please note that these tools are insecure and do not fully implement our cryptographic protocols.
We do not recommend any third-party tools. Our onsite download is already using multichannel downloading.
Also, these foreign tools cannot detect your quota and could limit your bandwidth.

Oof. This HR sort of response alone is making me reconsider using MEGA, altogether.

ncw:
This has now been done with --mega-2fa-string so I'm going to close this.

Thank you for your submission and sorry we never merge it.

Issue #5318, opened 2021-05-09
Google Drive Public Link Issue when Parent Folder is Publically Shared


InspectorWilliamHenderson:
#### The Associated Forum Post URL

`https://forum.rclone.org/t/rclone-link-google-drive-permission-error/20240`

Someone has Mentioned about the Problem which I am facing in rClone in Forum but Didn't Specify the Correct Reason why that Problem is Happening. And, That Thread has been Marked as Closed, so Creating an Issue on  GitHub with a Possible Workaround for an Issue.

#### Problem You are Having with rClone

When the Folder is Shared in Google Drive (Through Official Apps or rClone), Then All the SubFolders and Files within that Folder becames Publicly Available (The Way It is Intended to Be). Now, When I try to Get Link of  Any SubFolders or Files Inside that Main Folder (Which are Already Publically Shared) through `rClone Link`, then rClone gives an `insufficientFilePermissions` probably because rClone is trying to Set/Change the Permission which isn't Allowed as long as Parent Folder is Publically Shared. Now, I was hoping If it would be possible to Check Current SubFolder or File Permissions whenever There is an `insufficientFilePermissions`, and If it is Publically Shared (Checked Through File Permissions), Output the Generic Google Drive Link using Folder/File ID.

#### rClone Version
```
rclone v1.56.0-beta.5482.732bc08ce
- os/version: ubuntu 18.04 (64 bit)
- os/kernel: 4.19.112+ (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.16.3
- go/linking: static
- go/tags: non
```
#### Which OS You are Using ?
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic
Bits:		x86_64
```
#### Which Cloud Storage System are You Using ? 

[Google Drive](https://drive.google.com)

#### rClone Command You were Trying to Run
```
rclone lsf InspectorWilliamHenderson:A

	B/
	C.mkv

rclone link InspectorWilliamHenderson:A/B

	https://drive.google.com/open?id=1-yFDDtM0UFULOIYXpFVAo8I_9xxxxxxx

rclone link InspectorWilliamHenderson:A/C.mkv

	https://drive.google.com/open?id=1KaodHXY7bPFvBUrPh8Uw7zXGkxxxxxxx

rclone link InspectorWilliamHenderson:A

[...]

PR #6874, opened 2023-03-21
#5319 Migrate release docker image to new workflow

RichardoC:

* To use a supported workflow step for creating the docker images
* To push to GHCR as well as dockerhub

#6698 and issue #5318

#### Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ x ] I have added tests for all changes in this PR if appropriate.
- [ x ] I have added documentation for the changes if appropriate.
- [ x ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ x ] I'm done, this Pull Request is ready for review :-)


RichardoC:
One for a future PR
* hopefully we can remove `provenance: false` now that you've fixed the build step
* Can we replace "Get actual patch version" etc with something less magic?

RichardoC:
Please don't merge this just yet, I'm wary of the `build_docker_volume_plugin` part, I'm not sure what it does so am unclear how to make it work with the new steps/registry

ncw:
> One for a future PR
> 
> * hopefully we can remove `provenance: false` now that you've fixed the build step

Yes that is fine. Don't really know what it does anyway ;-)

> * Can we replace "Get actual patch version" etc with something less magic?

Doesn't everybody love a bit of `$(echo $GITHUB_REF | cut -d / -f 3 | sed 's/v//g' | cut -d "." -f 1,2)` ;-)

Do you have a suggestion? I made the above monstrosity!

> Please don't merge this just yet, I'm wary of the `build_docker_volume_plugin` part, I'm not sure what it does so am unclear how to make it work with the new steps/registry

Yes I think the build docker volume part is wrong.

I think you need to adjust the Makefile to add the new registry. It uses buildx already so should be good to go.

Probably just changing this bit

https://github.com/rclone/rclone/blob/23579e3b99d909c760795b798928c1545aac6298/Makefile#L291-L293

To push to both container registries will work (provided the auth is set).

RichardoC:
I've made the changes your requested except to the makefile, what way would you preferred that changed? A duplicate step just for Github, or that the registry is parameterised?

ncw:
> I've made the changes your requested except to the makefile, what way would you preferred that changed? A duplicate step just for Github, or that the registry is parameterised?

I'd just duplicate the push line and push to both repos one after the other. That should work.

RichardoC:
> > I've made the changes your requested except to the makefile, what way would you preferred that changed? A duplicate step just for Github, or that the registry is parameterised?
> 
> I'd just duplicate the push line and push to both repos one after the other. That should work.

I've currently misused the PLUGIN_USER variable so we can push to both, how's that as an approach?

ncw:
Thats a good idea. Your code seems to push all the architectures to ghcr but only amd64 to docker hub - is that right?

.github/workflows/build_publish_beta_docker_image.yml
```diff
@@ -48,14 +48,13 @@ jobs:
...
-                push: true # push the image to ghcr
+                push: true
                 tags: |
                   ghcr.io/rclone/rclone:beta
                   rclone/rclone:beta
                 labels: ${{ steps.meta.outputs.labels }}
                 platforms: linux/amd64,linux/386,linux/arm64,linux/arm/v7,linux/arm/v6
                 cache-from: type=gha
                 cache-to: type=gha,mode=max
-                provenance: false
```
qdm12, line 59: Actually using [`provenance: true`](https://docs.docker.com/build/attestations/slsa-provenance/) is quite nice security-wise, maybe in another PR. But yes it's unneeded to set it to `false` since it looks disabled by default anyway

.github/workflows/build_publish_release_docker_image.yml
```diff
@@ -23,27 +23,65 @@ jobs:
...
+                images: ghcr.io/${{ github.repository }}
+            - name: Set up QEMU
+              uses: docker/setup-qemu-action@v2
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v2
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v2
+              with:
+                registry: ghcr.io
+                # This is the user that triggered the Workflow. In this case, it will
```
qdm12, line 44: nit it would be nice to have this after the docker hub login to group logins together 🤔

.github/workflows/build_publish_release_docker_image.yml
```diff
@@ -23,27 +23,65 @@ jobs:
...
+                # This is used to authenticate against GitHub Container Registry.
+                # See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret
+                # for more detailed information.
+                password: ${{ secrets.GITHUB_TOKEN }}
 
+            - name: Build and publish image
+              uses: docker/build-push-action@v4
+              with:
+                file: Dockerfile
+                context: .
```
qdm12, line 59: nit unnecessary, these are the defaults

```suggestion
```

.github/workflows/build_publish_release_docker_image.yml
```diff
@@ -23,27 +23,65 @@ jobs:
...
+                cache-to: type=gha,mode=max
     build_docker_volume_plugin:
         if: github.repository == 'rclone/rclone'
         needs: build
         runs-on: ubuntu-latest
         name: Build docker plugin job
         steps:
             - name: Checkout master
-              uses: actions/checkout@v3
+              uses: actions/checkout@v2
```
qdm12, line 81: Why the change to v2?

qdm12 (commented):
Nice changes 👍 Although the duplication of the last step for each registry is not so pretty, but there might be no alternative unless we refactor the Makefile quite a bit.

On a side note, if any of you is familiar with Docker buildx and cross building, feel free to review related PR https://github.com/rclone/rclone/pull/7224 I opened recently, thanks!

PR #6897, opened 2023-03-26
webdav: make pacer minSleep configurable

9001:

This adds the config argument `--webdav-pacer-min-sleep` which specifies the http-request rate limit, keeping the previous default of 10ms. Reducing this value can be very beneficial when synchronizing small files.

Yes; https://forum.rclone.org/t/webdav-with-persistent-connections/37024/10

#### Checklist

the autogenerated documentation seems to cover this change appropriattely, and I believe the change is trivial enough that it doesn't need any additional tests :>

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	minSleep      = 10 * time.Millisecond
+	minSleep      = fs.Duration(10 * time.Millisecond)
 	maxSleep      = 2 * time.Second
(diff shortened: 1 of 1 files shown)


backend/webdav/webdav.go
```diff
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	minSleep      = 10 * time.Millisecond
+	minSleep      = fs.Duration(10 * time.Millisecond)
```
9001, line 46: This change ensures the setting is given the correct data type, as otherwise it wouldn't permit specifying the value using an `ms` suffix for instance.

PR #6901, opened 2023-03-28
backend: allow specifying SSH host key algorithms to use for sftp

joelnb:

Add an option `host_key_algorithms` to the SFTP backend such that you can specify `--sftp-host-key-algorithms` as a list of SSH host key algorithms in preference order. For example:

```
--sftp-host-key-algorithms ssh-ed25519,ssh-rsa
```

This can be useful when working with hosts that have a non-standard SSH implementation or when you have a particular preference for a type of SSH host keys.

Yes but not by me.

I found this [forum thread](https://forum.rclone.org/t/simple-sftp-connection-fails-with-eof/35065/5) and I have the same issue with the SFTP backend against a Remarkable running dropbear SSH. While I agree with comments by `ncw` in that thread that this is potentially an issue on the server side I would much prefer a client side fix given that the server side is a proprietary solution. I think that this change is a reasonable one to allow us (`Silver7845` from the forum and myself) to work around the issue without being too specific to the problem we have - there is no change in behaviour unless specifying the new option and there are other reasons why a user may wish to specify a specific list of SSH host key algorithms.

With this change (and using the example option from above) I can connect to the Remarkable device just fine.

#### Checklist

##### Tests

I did not add any tests because there does not seem to be anywhere in this folder to fit this in. If I have got that wrong please let me know and I will do my best to get something added.

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -370,6 +370,20 @@ Example:
     umac-64-etm@openssh.com umac-128-etm@openssh.com hmac-sha2-256-etm@openssh.com
 `,
 			Advanced: true,
+		}, {
+			Name:    "host_key_algorithms",
+			Default: fs.SpaceSepList{},
+			Help: `Space separated list of host key algorithms, ordered by preference.
+
+At least one must match with server configuration. This can be checked for example using ssh -Q HostKeyAlgorithms.
+
+Note: This can affect the outcome of key negotiation with the server even if server host key validation is not enabled.
+
+Example:
+
+    ssh-ed25519 ssh-rsa ssh-dss
+`,
+			Advanced: true,
 		}},
 	}
 	fs.Register(fsi)
@@ -408,6 +422,7 @@ type Options struct {
 	Ciphers                 fs.SpaceSepList `config:"ciphers"`
 	KeyExchange             fs.SpaceSepList `config:"key_exchange"`
 	MACs                    fs.SpaceSepList `config:"macs"`
+	HostKeyAlgorithms       fs.SpaceSepList `config:"host_key_algorithms"`
 }
 
 // Fs stores the interface to the remote SFTP files
@@ -738,6 +753,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 		Timeout:         f.ci.ConnectTimeout,
 		ClientVersion:   "SSH-2.0-" + f.ci.UserAgent,
+	}
+
+	if len(opt.HostKeyAlgorithms) != 0 {
+		sshConfig.HostKeyAlgorithms = []string(opt.HostKeyAlgorithms)
 	}
 
 	if opt.KnownHostsFile != "" {


joelnb:
## Validation

Possibly should have attempted to do some validation that the strings passed here are actually valid SSH host key types but as `golang.org/x/crypto/ssh` only seems to expose them as individual `const`s rather than any kind of list it didn't seem like it would be very nice to implement. Happy to have a go if it is required for merge but if so I would appreciate any input on how best to go about this to avoid over complicating things.

backend/sftp/sftp.go
```diff
@@ -740,6 +752,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		ClientVersion:   "SSH-2.0-" + f.ci.UserAgent,
 	}
 
+	if opt.HostKeyAlgorithms.String() != "" {
```
ncw, line 755:
```go
if len(opt.HostKeyAlgorithms) != 0 {
```

would be idiomatic here

backend/sftp/sftp.go
```diff
@@ -370,6 +370,17 @@ Example:
...
+		}, {
+			Name:    "host_key_algorithms",
+			Default: fs.CommaSepList{},
+			Help: `Comma separated list of host key algorithms, ordered by preference.
+
+At least one must match with server configuration.
+
+Example:
+
+    ssh-ed25519,ssh-rsa
```
ncw, line 382: I'd probably put all the acceptable values here in the example just to save the users time looking stuff up.

ncw (commented):
This looks very nice - thank you.

The error from the ssh library looks fine if the user passes a bad value so no need for validation.

I put a couple of very minor comments inline.

I'll run the CI now and see if that turns anything up.

docs/content/sftp.md
```diff
@@ -578,6 +578,20 @@ Properties:
     - "~/.ssh/known_hosts"
         - Use OpenSSH's known_hosts file.
 
+#### --sftp-host-key-algorithms
```
albertony, line 581: This part will be autogenerated from the definition of the option in the .go file - and overwritten. You currently have different description here and in there, so you should make sure the description you want to use are in the .go file (Help property), and then you can remove the change here in the .md.

ncw:
The CI looks good (the linux test failure is not your fault!)

backend/sftp/sftp.go
joelnb: Appreciate the feedback. Was completely forgetting I could just use standard list operations here but you are correct that is much better!

backend/sftp/sftp.go
albertony: First of all, sorry for all the nitpicking (from my part), I really think you made a great first-time contribution! 😃 

In my opinion, after thinking about it, `SpaceSepList` makes more sense here, since other very similar options (?) in this backend such as `key_exchange` and `macs` uses that. Though I do agree that matching the openssh syntax with comma separated does also make sense, but I assume that would be the same with the other options as well? Elsewhere in rclone we seem to use both, e.g. http backend uses comma for headers, hasher backend uses it for hash algorithms.

It seems to me `ssh -Q HostKeyAlgorithms` would given an even more complete set of legal values, including rsa-sha2-256, see: https://security.stackexchange.com/questions/248406/ssh-q-key-not-listing-all-key-types

On the other hand, the resulting list gets long, I agree with that. Maybe we could just make a shorter list which is a realistic example? Seems this is what is done for macs and key_exchange. And these even mention running a corresponding ssh command to see the full list, so could do the same here?

Oh, and doesn't this qualify for `Advanced: true` as well, like the other mentioned related options? This way it will not confuse newbies and will be "hidden" in the rclone config sequence behind:

```
Edit advanced config?
y) Yes
n) No (default)
y/n>
```

joelnb: Not a problem at all - I really appreciate the advice about how to make this as good as it can be.

I agree with the suggestions you have made so I have made those changes and all seems to work well in my testing.

albertony: Perfect. I like this version. But I'll let @ncw have the final word, I think I sort of hijacked his review...

ncw (approved):
Looks great - thank you :-)

Will merge now.

joelnb:
@ncw @albertony thanks both for your review!

I updated the text of this MR so that the example shown matches the current usage (after switching to `SpaceSepList`). Hopefully this will help anyone reading in future to use the correct syntax for the new option.

PR #6912, opened 2023-04-04
Config file save improvements

albertony:

See commit messages and issues linked from those.

The following example should also illustrate the changes.

Given symbolic link:
```
tempconfig.txt -> tempconfig/temp/tempconfig.txt
```
Then:
```
./rclone -vv config create yyy local --config=tempconfig.txt
2023/04/04 16:42:36 DEBUG : rclone: Version "v1.63.0-DEV" starting with parameters ["./rclone" "-vv" "config" "create" "yyy" "local" "--config=tempconfig.txt"]
2023/04/04 16:42:36 DEBUG : Using config file from "/home/user/projects/rclone/tempconfig.txt"
2023/04/04 16:42:36 DEBUG : yyy: config in: state="", result=""
2023/04/04 16:42:36 DEBUG : yyy: config out: out=<nil>, err=<nil>
2023/04/04 16:42:36 DEBUG : Saving config file: /home/user/projects/rclone/tempconfig.txt
2023/04/04 16:42:36 DEBUG : Config path resolved to: tempconfig/temp/tempconfig.txt
2023/04/04 16:42:36 DEBUG : Writing new config to temporary file: tempconfig/temp/tempconfig.txt740963405
2023/04/04 16:42:36 DEBUG : Moving existing config to temporary file: tempconfig/temp/tempconfig.txt.old1388108941
2023/04/04 16:42:36 DEBUG : Moving temporary file with new config to real config path: tempconfig/temp/tempconfig.txt
[yyy]
type = local
2023/04/04 16:42:36 DEBUG : rclone: Version "v1.63.0-DEV" finishing with parameters ["./rclone" "-vv" "config" "create" "yyy" "local" "--config=tempconfig.txt"]
```

#### Checklist

### DIFF
--- a/fs/config/configfile/configfile.go
+++ b/fs/config/configfile/configfile.go
@@ -106,12 +106,30 @@ func (s *Storage) Save() error {
 	if configPath == "" {
 		return fmt.Errorf("failed to save config file, path is empty")
 	}
-	dir, name := filepath.Split(configPath)
-	err := file.MkdirAll(dir, os.ModePerm)
+	configDir, configName := filepath.Split(configPath)
+
+	info, err := os.Lstat(configPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to resolve config file path: %w", err)
+		}
+	} else {
+		if info.Mode()&os.ModeSymlink != 0 {
+			configPath, err = os.Readlink(configPath)
+			if err != nil {
+				return fmt.Errorf("failed to resolve config file symbolic link: %w", err)
+			}
+			if !filepath.IsAbs(configPath) {
+				configPath = filepath.Join(configDir, configPath)
+			}
+			configDir = filepath.Dir(configPath)
+		}
+	}
+	err = file.MkdirAll(configDir, os.ModePerm)
 	if err != nil {
 		return fmt.Errorf("failed to create config directory: %w", err)
 	}
-	f, err := os.CreateTemp(dir, name)
+	f, err := os.CreateTemp(configDir, configName)
 	if err != nil {
 		return fmt.Errorf("failed to create temp file for new config: %w", err)
 	}
@@ -138,7 +156,7 @@ func (s *Storage) Save() error {
 	}
 
 	var fileMode os.FileMode = 0600
-	info, err := os.Stat(configPath)
+	info, err = os.Stat(configPath)
 	if err != nil {
 		fs.Debugf(nil, "Using default permissions for config file: %v", fileMode)
 	} else if info.Mode() != fileMode {
@@ -153,7 +171,7 @@ func (s *Storage) Save() error {
 		fs.Errorf(nil, "Failed to set permissions on config file: %v", err)
 	}
 
-	fbackup, err := os.CreateTemp(dir, name+".old")
+	fbackup, err := os.CreateTemp(configDir, configName+".old")
 	if err != nil {
 		return fmt.Errorf("failed to create temp file for old config backup: %w", err)
 	}
--- a/fs/config/configfile/configfile.go
+++ b/fs/config/configfile/configfile.go
@@ -106,7 +106,6 @@ func (s *Storage) Save() error {
 	if configPath == "" {
 		return fmt.Errorf("failed to save config file, path is empty")
 	}
-
 	dir, name := filepath.Split(configPath)
 	err := file.MkdirAll(dir, os.ModePerm)
 	if err != nil {
@@ -119,7 +118,7 @@ func (s *Storage) Save() error {
 	defer func() {
 		_ = f.Close()
 		if err := os.Remove(f.Name()); err != nil && !os.IsNotExist(err) {
-			fs.Errorf(nil, "failed to remove temp config file: %v", err)
+			fs.Errorf(nil, "Failed to remove temp file for new config: %v", err)
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


fs/config/configfile/configfile.go
```diff
@@ -107,8 +107,27 @@ func (s *Storage) Save() error {
 		return fmt.Errorf("failed to save config file, path is empty")
 	}
 	fs.Debugf(nil, "Saving config file: %v", configPath)
-	dir, name := filepath.Split(configPath)
-	err := file.MkdirAll(dir, os.ModePerm)
+
+	configPathReal := configPath
+	info, err := os.Lstat(configPath)
```
ncw, line 112: You'll probably be better off using [filepath.EvalSymlinks](https://pkg.go.dev/path/filepath?utm_source=godoc#EvalSymlinks)

Note that the argument it it should be absolute otherwise it won't resolve relative symlinks properly.

I think you can call it even if the argument isn't a symlink which will simplify the code a bit.

fs/config/configfile/configfile.go
```diff
@@ -107,8 +107,27 @@ func (s *Storage) Save() error {
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to resolve config file path: %w", err)
+		}
+	} else {
+		if info.Mode()&os.ModeSymlink != 0 {
```
ncw, line 118: This is only going to work for absolute symlinks, for relative symlinks it will assume they are relative to rclone's current directory.

fs/config/configfile/configfile.go
```diff
@@ -140,15 +159,15 @@ func (s *Storage) Save() error {
 	}
 
 	var fileMode os.FileMode = 0600
-	info, err := os.Stat(configPath)
+	info, err = os.Stat(configPathReal)
```
ncw, line 162: I think I'd try not to rename the `configPath` variable to minimize the patch.

ncw (commented):
I think you are right - those extra Debugs are going to be very noisy :-)

I put some notes inline about resolving symlinks (another thing which is surprisingly hard!).

fs/config/configfile/configfile.go
albertony: I started out with EvalSymlinks. Problem was that if the target did not exist, it just returned an error. While with Lstat + Readlink I was able to get the target path even if it does not exist.

albertony: And also the golang issue 40180 scared me... :) (deliberately not linking to it, I'm find the github cross-linking of all mentions a bit too much sometimes...)

fs/config/configfile/configfile.go
albertony: The reason for this was to keep the following at the end, with the s.fi pointing to FileInfo retrieved from the original path. To reduce risk of introducing issue due to some nuances I hadn't considered. But I guess Stat follows the symlink anyway and we can trust the result to be the same directly on the resolved path..?

```
// Update s.fi with the newly written file
s.fi, _ = os.Stat(configPath)
```

albertony: I guess the s.fi is only to record the modtime and size:

https://github.com/rclone/rclone/blob/243bcc9d0787276ccaccbbd914d70ede6a0305e6/fs/config/configfile/configfile.go#L40

So that should be safe - will fix.

albertony:
Thanks.

> I think you are right - those extra Debugs are going to be very noisy :-)

Definitely. Useful for testing at this stage, so I'll keep them for now and (hopefully ;)) remember to remove them in time...

> I put some notes inline about resolving symlinks (another thing which is surprisingly hard!).

Yes, maybe I stepped right into a classic "how hard can it be" trap... and opened a can of worms! If so, perhaps the first commit (overwriting/removing .old files) should be a separate PR after all...

fs/config/configfile/configfile.go
albertony: Right... gotcha! :)

ncw:
> Pushed a couple of smaller improvements according to your comments, but not so important until the main question regarding how "deep" to handle symlink (EvalSymlinks) is decided..

I think just the last one is fine! Hence your original code with a small mod - see above.

fs/config/configfile/configfile.go
albertony: Sure, very similar to what I had pushed in an update just before your comment (I assume you hadn't looked at that - or else I'm really confused...). I think this will be a good improvement and as far as I can tell would solve most cases where user's have symlinked the config file, without diving too deep into the rabbit hole with potential for introducing new problems (especially on windows and maybe in combination with network paths, judging from the EvalSymlinks discussions).

albertony:
I think I have it fixed according to our discussions, unless I've misunderstood you? And cleaned up (removed the debug output). And documented, but wasn't too sure how to explain the symlink part so feel free to suggest an alternative paragraph.

albertony:
Argh... On testing with symlink on windows I got exactly what I was fearing: 11 backup (.old) config files remaining in my directory, as a result of "Failed to save config after 10 tries".

Test case:
```
mklink configlink.txt temp\config.txt
rclone --config=configlink.txt -vv config create zzz local
```
When directory temp does not exist the empty file created by os.CreateTemp was left behind, after error exit on "failed to move newly written config from ... to final location" before backup file cleanup. Such an empty backup file is of no use, so I've changed the code to take this into account and delete the backup even on write error of the new file if the backup was simply an empty file.

Now what if one comes into a similar situation, with "Failed to save config after 10 tries", but there was an existing config file? It should then leave behind the renamed backup of the existing file on the first failed save, and on the following failed saves it will not find an existing config file (since it has been renamed to .old) and the result should be as above. End result is one backup file left behind containing the actual original config file.

albertony:
Back from some Easter holiday (hope you're enjoying some time off too!); fixed what was really an underlying bug in the case from my previous comment (introduced after I refactored away the configPathReal used initially). It will now actually create the directory `temp` and the config create command will succeed.

fs/config/configfile/configfile.go
```diff
@@ -106,20 +106,37 @@ func (s *Storage) Save() error {
...
+	} else {
+		if info.Mode()&os.ModeSymlink != 0 {
+			configPath, err = os.Readlink(configPath)
+			if err != nil {
+				return fmt.Errorf("failed to resolve config file symbolic link: %w", err)
+			}
+			if !filepath.IsAbs(configPath) {
+				configPath = filepath.Join(configDir, configPath)
+			}
+			configDir, _ = filepath.Split(configPath)
```
ncw, line 125: ?

    configDir = filepath.Dir(configPath)

ncw (commented):
I've been though that logic carefully. I think it is OK!

I put one suggestion inline.

Maybe we should try to test all the failure branches in the test code - that will be hard though!

albertony:
Great, thanks. I've pushed an update according to your last comment.

> Maybe we should try to test all the failure branches in the test code - that will be hard though!

Sounds like a good idea, but would definitely not be easy - at least to cover all of them! I could give it a try, but will probably be several days or even weeks before I get enough time to get anything out of it.

albertony:
Actually I got some extra time today, so I wrote a few tests around the configfile save. Covering symlinks (but not on Windows, as it seem to require admin), creation of directories in the path, and a few FileMode/read-only related cases. Some failure branches are covered with this, but not nearly all of them.

ncw (approved):
Very nice tests :-)

This looks good now - please merge when you are ready. I'll test it some more once you've merged it.

PR #6914, opened 2023-04-04
azureblob: send nil tier if empty string

roelarents:

Azurite [complains](https://github.com/Azure/Azurite/blob/main/src/blob/handlers/BlockBlobHandler.ts#L127) about an invalid `ms-access-tier` header when uploading a blob when no tier is [configured](https://rclone.org/azureblob/#azureblob-access-tier) (and [defaults to empty/nothing](https://github.com/rclone/rclone/blob/master/backend/azureblob/azureblob.go#L65)).

The purpose of this change is to be compatible with Azurite.

The API spec states [specifying the access tier is optional](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/storage/data-plane/Microsoft.BlobStorage/preview/2021-08-06/blob.json#L4145). [As do the docs](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad#request-headers-all-blob-types). However I think Azurite is more strict than the Azure production env and assumes an empty header/string as supplied but invalid.

Couldn't find a mentioning of it.

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -1471,9 +1471,8 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 	srcBlobSVC := srcObj.getBlobSVC()
 	srcURL := srcBlobSVC.URL()
 
-	tier := blob.AccessTier(f.opt.AccessTier)
 	options := blob.StartCopyFromURLOptions{
-		Tier: &tier,
+		Tier: parseTier(f.opt.AccessTier),
 	}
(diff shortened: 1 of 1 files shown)


ncw (commented):
This seems like a sensible idea.

There are some other places that it needs fixing though - search for `Tier:` and you'll see what I mean.

Can you rebase on master please to fix the conflict - thanks!

roelarents:
Sorry that was a bit sloppy. Rebased and mended the other `Tier:`s. Thx for looking at it.

ncw (approved):
Looks great - thank you :-)

Will merge now.

PR #6931, opened 2023-04-09
serve: trigger systemd notify for restic

shyim:

I would like to react for systemd notify hooks to savely build my backup scripts.

I have an Restic Service + Restic Timer and they start Rclone before they run. I want to make it safe that the Restic API is ready before the Restic process itself starts. Therefore I want to use `Type=notify` on my Rclone.

Nope

#### Checklist

### DIFF
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	sysdnotify "github.com/iguanesolutions/go-systemd/v5/notify"
+
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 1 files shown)


ncw:
This looks reasonable to me.

Can you think of any reasons why we might need to control this with a flag, or do you think it is safe to leave on all the time?

shyim:
In the Docker serve it's also without flag https://github.com/rclone/rclone/blob/52fbb10b47718220ca987e80f691e776fadbf5eb/cmd/serve/docker/driver.go#L90.

So I would leave it so :)

ncw (approved):
Looks good - thank you :-)

Will merge now.

PR #6954, opened 2023-04-17
Backend/cryptomator

### LINKED ISSUE #1647: Cryptomator support
Cryptomator support for rclone would be amazing. It could be a new option for encrypted remotes that could live alongside existing crypt remotes. The big advantage of Cryptomator is that its vaults can be mounted as a drive (i.e. filesystem) directly from the cloud on mobile (using their 1st-party app) & desktop (using Mountain Duck, Expandrive, etc).

https://cryptomator.org

Encryption architecture: https://cryptomator.org/architecture/

Source code: https://github.com/cryptomator

There’s also a page listing the other apps that have integrated Cryptomator, and inviting people to get in touch about integrating it into a new app: https://cryptomator.org/coop/

fhilgers:

This PR implements the backend for cryptomator. It should be noted that it is far from being completed as some parts are still missing but the integration tests pass.

This should close #1647.

#### Checklist

As this is only a draft, most of the checklist is not completed yet...


ncw:
This looks like an excellent start :-)

You'll need to put comments in for all the exported methods as the linter will complain about them all :-(

When you say the integration tests pass - which ones? The backend one or `test_all -remotes cryptomator`?

This is missing a lot of optional methods like `Move`, `Copy` etc - these make the file system a lot more usable.

Is this going to need any more config parameters to create a vault?

Are there any actions users are going to want to perform on vaults which they might like a separate command line interface for? That is what the `rclone backend` functions are for.

fhilgers:
The backend test as well as `test_all -remotes cryptomator` passes now.

I will try to implement the optional methods that I suppose are possible either today or tomorrow.

I do not think cryptomator needs additional config paramaters, as the real configuration for the `Vault` is contained in a configuration file at the root of it. This is necessary to make it compatible with the existing cryptomator tools.

Regarding the `rclone backend` functions I will implement a command to create a new vault, if there is none.

I am going to work on the Documentation and Proper commit messages over the next few days.

ncw:
> The backend test as well as `test_all -remotes cryptomator` passes now.

Super job - well done :-)

> I will try to implement the optional methods that I suppose are possible either today or tomorrow.

Probably the most important one is `Move` which is used to rename a file and the next most important is `MoveDir` to rename a directory.

Note that since this wraps another backend it should declare itself a wrapping backend and implement all the optional methods. You do this with

https://github.com/rclone/rclone/blob/589b7b4873f8ba1d77f982ce14d4ac9a8adb18f6/backend/crypt/crypt.go#L636-L639

Then tell the test harness the names of the methods you can't implement

https://github.com/rclone/rclone/blob/589b7b4873f8ba1d77f982ce14d4ac9a8adb18f6/backend/crypt/crypt_test.go#L27-L28

A lot of the optional methods you can pass straight on to the backend being wrapped (after translating the path of course), eg

https://github.com/rclone/rclone/blob/589b7b4873f8ba1d77f982ce14d4ac9a8adb18f6/backend/crypt/crypt.go#L627-L634

or

https://github.com/rclone/rclone/blob/589b7b4873f8ba1d77f982ce14d4ac9a8adb18f6/backend/crypt/crypt.go#L1145-L1152

> I do not think cryptomator needs additional config paramaters, as the real configuration for the `Vault` is contained in a configuration file at the root of it. This is necessary to make it compatible with the existing cryptomator tools.

Perfect.

If it doesn't need any config we could use a shortcut so people could access a vault as `:cryptomator:myremote:path/to/vault`. I'll describe how to do this later!

> Regarding the `rclone backend` functions I will implement a command to create a new vault, if there is none.

That is great - that makes rclone usable on its own.

> I am going to work on the Documentation and Proper commit messages over the next few days.

:+1:

fhilgers:
Every Filesystem feature apart from Purge, ListR, ChangeNotify, MergeDirs and OpenWriterAt can and is implemented now.

Due to the fact that the directory structure is flattened, we cannot make use of Purge or ListR. 
Also as we cannot get the unencrypted remote path from an encrypted remote path ChangeNotify is also not possible.

Currently the performance of reading and writing files is not different from the `crypt` remote, but directory traversal is much slower as many requests have to be made to resolve the directory ids. The same is true for Put requests, when the directory was not yet created.

To create a new directory we have to create three directories and write to files which is a lot of overhead.

Is there any simple way to batch those requests?
When Mkdir is called we currently try to find it first and if found add it into our directory id cache, if not we create the three necessary directories and the two files.
Would it be possible to just create a new id, add it into our cache and into a queue and then batch all those requests which just put 36 bytes for the id into one?

Situation Currently:

```
Put dir/file1.txt
Put dir/file2.txt
```

With multiple transfers those two requests can be done, as far as I am aware concurrently. But as `dir` does not yet exist, we have to lock, so that we do not generate two random id's for the same directory and both Put requests have to wait for the lock.

ncw:
> Every Filesystem feature apart from Purge, ListR, ChangeNotify, MergeDirs and OpenWriterAt can and is implemented now.

Great work :-)
 
> Due to the fact that the directory structure is flattened, we cannot make use of Purge or ListR. Also as we cannot get the unencrypted remote path from an encrypted remote path ChangeNotify is also not possible.

:+1:

> Currently the performance of reading and writing files is not different from the `crypt` remote, but directory traversal is much slower as many requests have to be made to resolve the directory ids. The same is true for Put requests, when the directory was not yet created.
> 
> To create a new directory we have to create three directories and write to files which is a lot of overhead.
> 
> Is there any simple way to batch those requests? When Mkdir is called we currently try to find it first and if found add it into our directory id cache, if not we create the three necessary directories and the two files. Would it be possible to just create a new id, add it into our cache and into a queue and then batch all those requests which just put 36 bytes for the id into one?
> 
> Situation Currently:
> 
> ```
> Put dir/file1.txt
> Put dir/file2.txt
> ```
> 
> With multiple transfers those two requests can be done, as far as I am aware, concurrently. But as `dir` does not yet exist, we have to lock, so that we do not generate two random id's for the same directory and both Put requests have to wait for the lock.

Go is good at doing things concurrently and in the background. You can use `x/sync/errgroup` (there are lots of examples in the code) for an easy way of doing a few things concurrently. You could then do the directory creations / fil puts simultaneously. Note that `Put` will create directories for you so you don't need to explicitly create them.

Would that help?

fhilgers:
Sorry, I had a lot to do in the last few weeks.

I will fix the windows issue in the next few days and look into how I might fix the locking issue!

ncw:
> Sorry, I had a lot to do in the last few weeks.

No worries!
 
> I will fix the windows issue in the next few days and look into how I might fix the locking issue!

Great.

The windows issue looks like some encoding problem, or maybe a bug?

```
            --- FAIL: TestStandard/FsMkdir/FsEncoding/punctuation (14.05s)
                fstests.go:679: testing "!\"#$%&'()*+,-.／:;<=>?@[\\]^_`{|}~"
                fstest.go:298: Sleeping for 1s for list eventual consistency: 1/3
                fstest.go:298: Sleeping for 2s for list eventual consistency: 2/3
                fstest.go:298: Sleeping for 4s for list eventual consistency: 3/3
                fstest.go:327: 
                    	Error Trace:	D:/a/rclone/rclone/fstest/fstest.go:327
                    	            				D:/a/rclone/rclone/fstest/fstest.go:338
                    	            				D:/a/rclone/rclone/fstest/fstests/fstests.go:686
                    	Error:      	Not equal: 
                    	            	expected: []string{"!\"#$%&'()*+,-.／:;<=>?@[\\]^_`{|}~"}
                    	            	actual  : []string{"!\"#$%&'()*+,-.／:;<=>?@[\\]^_`{|}~",   <--- looks OK
                                                           "!\"#$%&'()*+,-.／:;<=>?@[\\]^_`{|}~\\!\"#$%&'()*+,-.／:;<=>?@["}  <-- Unexpected
```

There seems to be an extra directory with some junk on the end which rclone wasn't expecting.
 
> About the other_os error, I am not really sure how I could fix this. As far as I am aware, the aes-siv library I used is probably the only tested one in go and the cryptomator spec heavily depends on that algorithm.

Use a build constraint to exclude the backend from being built on troublesome architectures.

Take a look at the sftp backend and grep for `plan9` and you'll see what you need to do. You'll need an `xxx_unsupported.go` file also so the build doesn't fail on unsupported architectures.

fhilgers:
I have hopefully fixed the windows issue and also excluded the js/wasm build target as it is not supported by the crypto library.
As I currently have no windows machine, it would be great if you could run the ci jobs again because I am not 100% sure about the windows fix.
Thanks :)

ncw:
Oops this fell off the radar. 

I think we are nearly ready to merge this?

fhilgers:
Oops, I must have made a mistake merging master in the Github UI. I will fix that tomorrow.

After the windows bug is fixed correctly, I think everything should be done.

ncw:
I think if you rebase this on master to pick up the current go.mod and go.sum (fix the conflicts) then run `go mod tidy` I think it should build OK.

ncw:
Can you rebase on master and fix the conflicts please?

fhilgers:
Sorry, I thought merging was fine as well. Now it is rebased on the current master branch.

Spaceghost:
@fhilgers This is really cool! Besides another rebase, is there anything that I might want to do to build this?
@ncw Your reviews on this PR are so helpful! I learned a lot just from your comments and @fhilgers' code. 

~👻

devnoname120:
@ncw What is the status on this pull request? Is anything missing before it can be merged?

fhilgers:
I merged master and added some of the methods added on filesystems since I have last looked at this to the `UnImplementable{}Methods` for testing, because I currently did not have the time to implement them. 

Before this can be merged there have to be some users who try out this backend first. As I am not affiliated in any way with the cryptomator team and my go library to access the cryptomator vaults is not yet integration tested against the official cryptomator app I would not use this in production.

I also do not feel confident in adding this into rclone yet as I do not have the time currently to maintain it.

If anyone is interested in helping, you are very welcome, if you have questions just write me a mail.

Issue #6932, opened 2023-04-10
Custom encrypted suffix


youtehub:
#### What is your current rclone version (output from `rclone version`)?
[rclone v1.62.2](https://github.com/rclone/rclone/releases/tag/v1.62.2)

#### What problem are you are trying to solve?

The current encryption suffix is *.bin, I hope there is an option to customize the encrypted file suffix, such as changing to *.jpg, *.mp4, which is to customize the suffix to disguise the encrypted file.
In this way, rclone encrypted files can be placed in different places according to user needs

#### How do you think rclone should be changed to solve that?

In the encrypted crypt configuration file of rclone, an optional configuration item **custom_suffix** is provided to allow users to customize, and the default is still in the form of **bin**

Custom encrypted suffixes are not yet supported, the following example is a sample configuration file template given by myself
```
[folderCrypt]
type = crypt
remote = D:\folders
directory_name_encryption = false
password = password1
password2 = password2 
filename_encoding = base32768
filename_encryption = standard
custom_suffix = mp4
```

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

albertony:
In my mind that makes sense, and don't think it would be that much work to make the following into an option:

https://github.com/rclone/rclone/blob/c502e00c87854781a939b385dda0e9d38ecbd81f/backend/crypt/cipher.go#L41

ncw:
Yes it could be an option if someone wants to have a go at it.

albertony:
Oh, and then allowing none / empty string should probably be allowed, in cases where path length is critical..

ncw:
I think this is all done now in 281a007b1a7867db5904ee09131e5809c0606f3d ghank you @jladbrook 

Currently in [the latest beta](https://beta.rclone.org/) and to be released in v1.63

88elie88:
Feature request: preserve suffix on encrypted names (filename_encryption = standard)
What problem are you trying to solve?
I use multiple crypt remotes on the same underlying storage, all sharing the
same password, to fit two different workflows:
Local device: filename encryption off (I need to recognize files by
name day-to-day), but file content encrypted.
Cloud backup: filename encryption standard (full anonymity, since
[...]

Issue #6392, opened 2022-08-22
Force sync "clean up"


DarrenPIngram:
#### What is your current rclone version (output from `rclone version`)?

1.59.1

#### What problem are you are trying to solve?

Would be nice to have a "force resync clean" feature, meaning that the source is compared to 100% on destination (remote) and anything not present on the remote is added and anything present on the remote that should not be is deleted.

I appreciate this is what sync should be, but in some sporadic cases I've found where this is not (at least "old" stuff left in remote (never have tracked down the "why" as it is noticed long after the event). Thus an occasional "clean", even of very large datastores would be nice.

#### How do you think rclone should be changed to solve that?

I presume a slight extension to the sync behaviour, possible an extra flag or something, but I would not suggest how it may be achieved. Especially if regular sync should be picking this up.

If there is such a problem, whether a compare could be easily made (compare remote to source) and delete/flag to delete could be one way.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
A sync would mirror A to B and clean up anything needed.

What isn't happening in that scenario that you are asking for? i can't figure out why another flag is needed.

animosity22:
It already does that so not sure why another command/flag would be needed.

If it's not, you'd have to share a log/etc to answer the question of why as you'd have found a bug perhaps.

animosity22:
That already exists as you can do --dry-run to see what will happen.

If you re-run a sync, it does the whole remote every time as that's how sync works.

A is mirrored to B.

animosity22:
Sync is one direction. A-> B

If you an issue with something not doing what you'd expect, you'd want to share all the details for a bug report with a log file of the issue.

If you want to go two directions, check out -> https://rclone.org/bisync/

animosity22:
The internet is always a great place as I'm volunteering my time and trying to figure out your question, help you out and you want to go with insulting me for trying to do that. 

Please raise your issue on the forums as someone else might help you out:

https://forum.rclone.org/

Masamune3210:
You don't, go to the forum like you were told and ask for help there. People arent exactly the most accepting when they are being insulted.....

PR #6958, opened 2023-04-19
crypt: adding an option to set a custom_suffix #6392

jladbrook:

Adds an additional option when encrypting files allowing for a custom suffix.

#6932 

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -38,7 +38,6 @@ const (
 	blockHeaderSize     = secretbox.Overhead
 	blockDataSize       = 64 * 1024
 	blockSize           = blockHeaderSize + blockDataSize
-	encryptedSuffix     = ".bin" // when file name encryption is off we add this suffix to make sure the cloud provider doesn't process the file
 )
 
 // Errors returned by cipher
@@ -54,7 +53,7 @@ var (
 	ErrorEncryptedBadBlock       = errors.New("failed to authenticate decrypted block - bad password?")
 	ErrorBadBase32Encoding       = errors.New("bad base32 filename encoding")
 	ErrorFileClosed              = errors.New("file already closed")
-	ErrorNotAnEncryptedFile      = errors.New("not an encrypted file - no \"" + encryptedSuffix + "\" suffix")
+	ErrorNotAnEncryptedFile      = errors.New("not an encrypted file - does not match suffix")
 	ErrorBadSeek                 = errors.New("Seek beyond end of file")
 	defaultSalt                  = []byte{0xA8, 0x0D, 0xF4, 0x3A, 0x8F, 0xBD, 0x03, 0x08, 0xA7, 0xCA, 0xB8, 0x3E, 0x58, 0x1F, 0x86, 0xB1}
 	obfuscQuoteRune              = '!'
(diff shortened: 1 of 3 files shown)


backend/crypt/crypt.go
```diff
@@ -48,7 +48,7 @@ func init() {
 					Help:  "Very simple filename obfuscation.",
 				}, {
 					Value: "off",
-					Help:  "Don't encrypt the file names.\nAdds a \".bin\" extension only.",
+					Help:  "Don't encrypt the file names.\nAdds a \".bin\", or \"custom_extension\" extension only.",
```
ncw, line 51: The parameter is called `custom_suffix` not `custom_extension`?

backend/crypt/crypt.go
```diff
@@ -151,6 +151,14 @@ length and if it's case sensitive.`,
 				},
 			},
 			Advanced: true,
+		}, {
+			Name: "custom_suffix",
+			Help: `If this is set it will override the default suffix of ".bin".
+
+If you leave blank, or do not set to a valid extension, e.g. "jpg" instead of ".jpg", 
+then the suffix will be empty. This may be useful when the path length is critical.`,
```
ncw, line 159: I'm not sure leaving it blank will work. I think if the user leaves it blank then they will get the default won't they?

I guess there might be a difference between not providing the `custom_suffix` argument and providing `custom_suffix =`.

The point that no suffix is useful when the path length is critical is a very good one.

backend/crypt/crypt.go
```diff
@@ -151,6 +151,14 @@ length and if it's case sensitive.`,
 				},
 			},
 			Advanced: true,
+		}, {
+			Name: "custom_suffix",
```
ncw, line 155: I think we should probably just call this option `suffix`. Then the config will be

```
suffix = .bin
```

Which I think reads nicer than `custom_suffix` as all the config is custom isn't it?

backend/crypt/cipher.go
```diff
@@ -200,6 +201,15 @@ func newCipher(mode NameEncryptionMode, password, salt string, dirNameEncrypt bo
 	return c, nil
 }
 
+// setEncryptedSuffix set suffix, or an empty string
+func (c *Cipher) setEncryptedSuffix(suffix string) {
+	if !strings.Contains(suffix, ".") {
+		c.encryptedSuffix = ""
```
ncw, line 207: I don't like setting it blank if there isn't a `.` - I think we should probably be adding a `.` and writing an `fs.Errorf` level warning for the user.

Maybe we should have a suffix of `none` to mean don't add a suffix rather than leaving it blank?

ncw (commented):
I think this is looking very nice.

I put a few suggestions inline.

backend/crypt/crypt.go
jladbrook: agreed, this has been updated.

backend/crypt/crypt.go
jladbrook: have updated as per your suggestion, e.g. ```none``` now sets the suffix to blank.

backend/crypt/cipher.go
jladbrook: have updated the "blank" detection and added the error as suggested

ncw (approved):

I'll just run the CI one last time and if it passes I'll merge

ncw:
Thank you :-)

PR #6969, opened 2023-04-22
cat: adds --separator option to cat command

### LINKED ISSUE #6968: feat: Add option to `rclone cat` that injects a separator between files

https://forum.rclone.org/t/print-new-line-between-files-when-using-rclone-cat/37732/8

```
❯ rclone version
rclone v1.62.2
- os/kernel: 5.15.90.1-microsoft-standard-WSL2 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.20.2
- go/linking: static
- go/tags: none
```

I am trying to output several remote files from an S3 bucket to stdout. Each file contains a single line. I'd like each file to output to a separate line also. However, rclone cat joins the files without any separator, which makes the output unusable/unreadable. Is there any way to accomplish this?

The simplest idea I had is to add an option to `rclone cat` that will print a "separator" string after outputting each file. E.g.

```
rclone cat s3:bucket/path --include */file.txt --separator '\n'
file1
file2
file3
...
```

#### How to use GitHub

lorengordon:

When using `rclone cat` to print the contents of several files, the
user may want to inject some separator between the files, such as a
comma or a newline. This patch adds a `--separator` option to the `cat`
command to make that possible. The default value remains an empty
string, `""`, maintaining the prior behavior of `rclone cat`.

Closes #6968

* https://forum.rclone.org/t/print-new-line-between-files-when-using-rclone-cat/37732/14
* https://github.com/rclone/rclone/issues/6968

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -1259,7 +1259,7 @@ type readCloser struct {
 //
 // if count < 0 then it will be ignored
 // if count >= 0 then only that many characters will be output
-func Cat(ctx context.Context, f fs.Fs, w io.Writer, offset, count int64) error {
+func Cat(ctx context.Context, f fs.Fs, w io.Writer, offset, count int64, sep []byte) error {
 	var mu sync.Mutex
 	ci := fs.GetConfig(ctx)
 	return ListFn(ctx, f, func(o fs.Object) {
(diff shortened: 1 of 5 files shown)


fs/operations/operations.go
```diff
@@ -1301,6 +1301,14 @@ func Cat(ctx context.Context, f fs.Fs, w io.Writer, offset, count int64) error {
 			err = fs.CountError(err)
 			fs.Errorf(o, "Failed to send to output: %v", err)
 		}
+		if len(sep) >= 0 {
+			sepReader := bytes.NewReader(sep)
```
ncw, line 1305: This will do to write the sep to the stream, no need to make a reader.

   _ , err =  w.Write(sep)

ncw (commented):
Looking good :-) I put a comment inline and I've given the tests another whirl.

fs/operations/operations.go
lorengordon: Ok, I was wondering about something like that, but wasn't sure if there was a reason to use `io.Copy()`. Updated, ran the test, and confirmed the behavior for my use case also!

ncw (approved):
This looks perfect - thank you :-)

I'll merge it now.

Issue #6841, opened 2023-03-13
Bisync fails when files are added during bisync operation


laurensknoll:
#### What is the problem you are having with rclone?

Bisync fails when files are added during bisync operation. Typically, files are added to the Path1 source (GCS), causing Listing1 to include files not actually copied. 

As a result, the final Path1 Listing does not match the Path2 Listing. Hence, the bisync is marked as "failed" and the succeeding runs fail as well.

**Error flow**
* [Optimistic copy at end of bisync-operation](https://github.com/rclone/rclone/blob/ce3b65e6dc0866cac041b82aabbf790ebd9778b6/cmd/bisync/operations.go#L263)
* [Error triggered by checkSync at end of bisync-operation](https://github.com/rclone/rclone/blob/ce3b65e6dc0866cac041b82aabbf790ebd9778b6/cmd/bisync/operations.go#L295)

**Solution area**
As we update files based on deltas, it makes sense to also update the Listing1 and Listing2 fileLists based on these deltas, instead of scanning the source and destination folders.

To do so, the copy operation needs to return the actual updates as well. Currently, `fastCopy` only returns the error details. By including the actual updates, we have the information to update the listing. Sadly, this information needs to propagate quite some layers and modules, from actual background transfers up to the bisync module.

* [Queued delta execution reference](https://github.com/rclone/rclone/blob/ce3b65e6dc0866cac041b82aabbf790ebd9778b6/cmd/bisync/deltas.go#L253)
* [fastCopy reference](https://github.com/rclone/rclone/blob/98fd00a655cada0bda53b7be3d54dd656a8ab21c/cmd/bisync/queue.go#L14)
* [copyDir reference](https://github.com/rclone/rclone/blob/98fd00a655cada0bda53b7be3d54dd656a8ab21c/fs/sync/sync.go#L1126)
* [run copy reference](https://github.com/rclone/rclone/blob/98fd00a655cada0bda53b7be3d54dd656a8ab21c/fs/sync/sync.go#L844)

#### What is your rclone version (output from `rclone version`)

1.61.1

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Windows Server 2022, 64 bit

#### Which cloud storage system are you using? (e.g. Google Drive)

Google Cloud Storage and Local File Storage (Windows)

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

`bisync gcs:/ C:\clone\`

[...]

Issue #5683, opened 2021-10-08
bisync: honor `--checksum` and `--ignore-checksum`


ivandeex:
## Synopsis

Delta engine in the _bisync beta_ can only use `modtime` to detect changes.
We should take hashsums into consideration too, respect the `--checksum` flag (and probably `--ignore-checksum`, but this has to be confirmed).

## Prior discussions

See https://github.com/cjnaz/rclonesync-V2/issues/5 and...

### https://github.com/rclone/rclone/pull/5164#issuecomment-859114705 (cjnaz)

Data point...
The TiBU directory has 19GB of files.  Path2 is a local dir.  2+ minutes for the `Path2 checking for diffs` step.

I tried the --ignore-checksum switch to no avail.

```
2021/06/10 14:36:26 INFO  : Synching Path1 "owncloud:TiBU/" with Path2 "/.../TiBU/"
2021/06/10 14:36:26 INFO  : Path1 checking for diffs
2021/06/10 14:36:26 INFO  : Path2 checking for diffs
2021/06/10 14:38:37 INFO  : No changes found
```

### https://github.com/rclone/rclone/pull/5164#issuecomment-859691994 (ivandeex) -> (ncw)

Current beta contains a [quick fix](https://github.com/ivandeex/rclone/commit/66b175d4dcd7ed56e9d79c25cd6cf5364686074b): in presence of --ignore-checksum the hashes will never be calculated (be it local fs, remote fs, or whatever). This is not a real solution but a kludge for your test.
For a real solution we have to develop a number of _rclonic_ rules for all case combinations: what to do if listings contain hash but user flagged ignore, how future name tracker will interact with this flag, should we treat backends with fast and slow hashsums differently, how --checksum and --ignore-checksum interact, etc, etc... It's a separate discussion and I'll drop this quick fix after all questions get solved.
For now you can try `--ignore-checksum` and evaluate the speed.

### https://github.com/rclone/rclone/pull/5164#issuecomment-859744308 (cjnaz)

--ignore-checksum brought the 19GB no-changes run down from 2 minutes to less than 1 second. Thanks for the interim hack.

## How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

[...]

Issue #5681, opened 2021-10-08
bisync: optimize --resync


ivandeex:
## Synposis

Clearly explain in documentation whether `--resync` is asymmetrical.
(@cjnaz could you comment?)

Investigate whether `--resync` can be re-implemented by the equivalent of:

```
rclone copy --update Path1 Path2
rclone copy --update Path2 Path1
```

### https://github.com/cjnaz/rclonesync-V2/issues/66 (eric-void) -> (cjnaz)

> looking at the code there is a lot of work to copy Path2 files not in Path1, and then do an "rclone sync" from Path1 to Path2. Why don't you just do and "rclone copy" from Path1 to Path2, and then an "rclone copy" from Path2 to Path1?
> This is much simpler, there is no need to read 2 times LSL files and to loop for missing Path2 files, and the result should be the same. Am i missing something?

The first-sync optimization is deferred for now since its not broke and should be rarely run.

### https://github.com/rclone/rclone/pull/5164#discussion_r688913159 (ncw)

So `--resync` is asymetrical?

Perhaps it would be better implemented by the equivalent of

```
rclone copy --update Path1 Path2
rclone copy --update Path2 Path1
```

## How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ivandeex:
Unit tests and documentation will be updated respectively.

nielash:
I had the same thought recently:

https://github.com/rclone/rclone/blob/156c372cd760460f419708c75506434c48decfc8/cmd/bisync/operations.go#L426-L433

I tend to think it would be better to change it, as in addition to being simpler and more symmetrical, it would also make `--resync` substantially faster for users of `--create-empty-src-dirs`. However, it would be a breaking change, and I'm not sure how concerned we should be about that. It's possible that some users have come to depend on the asymmetry in some way. For instance, imagine a scenario where you're bisyncing one central server with multiple devices, and the server is the source-of-truth. The asymmetry allows you to declare that the server should take priority in the event of a `--resync`, even if the corresponding file on the other path is newer.

Perhaps the best compromise is to make the behavior configurable with a flag?

crocinsocks:
[...]

Issue #5676, opened 2021-10-08
bisync: handle volatile changes consistently


ivandeex:
## Synopsis

This problem can be also title `recover after transient failures`.

@cjnaz once wrote in https://github.com/rclone/rclone/pull/5164#issuecomment-843406665

One perpetual problem I ran into was with a dynamically changing file system.  rclonesync is not robust for the case of a file no longer there after the current LSL snapshot was taken.  Also, if a file is being written to / blocked when rclonesync wants to copy it.  In such cases rclonesync throws up its hands and forces user interaction via a `--resync`.

## Prior discussions

Related to https://github.com/cjnaz/rclonesync-V2/issues/56
Related to #1316

### https://github.com/rclone/rclone/pull/5164#issuecomment-843426747 (JWink3101)

I agree with these issues of sorts. This isn't new in rclone either. If you change the filesystem while it is running, things can break. For my tool, I [enumated what would happen](https://github.com/Jwink3101/syncrclone/blob/master/docs/misc.md#interruptions) if it is interrupted. Something similar happens if a file changes.

But again, this is the beauty of the not-using-history-to-make-a-changelog approach (\#2). You only need history to figure out what to do!. So if a file was  added between the lsl, it will still copy.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843460561 (ivandeex)

Pretty close to #840 (recheck source during operation, open)
Related to #3770 (_.partial_ file support in LocalFS / FTP / SFTP, which can be _filtered out_ by bisync)

Covered by postponed item "Use temporary files with FTP, SFTP"
I guess it will change nothing in bisync but improve FTP/SFTP and only add related tests to bisync.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843851607 (ivandeex)

@cjnaz

> 
> > rclonesync is not robust for the case of a file no longer there after the current LSL snapshot was taken.

Such files should be just **skipped** (by default or under a command-line flag) similar to https://github.com/rclone/rclone/pull/5297#pullrequestreview-649944326

> 
> > Also, if a file is being written to / blocked when rclonesync wants to copy it.  In such cases rclonesync throws up its hands and forces user interaction via a --first-sync.
>
> Pretty close to #840 (recheck source during operation, open)
> Related to #3770 (_.partial_ file support in LocalFS / FTP / SFTP, which can be _filtered out_ by bisync)

[...]

PR #6971, opened 2023-04-23
bisync: bug fixes and new features including --create-empty-src-dirs and equality check

### LINKED ISSUE #6109: bisync flag "--create-empty-src-dirs" missing
The Flag "--create-empty-src-dirs", that is implemented in copy, sync, etc is missing for bisync.

nielash:

Several `bisync` bug fixes and new features including `--create-empty-src-dirs` and equality check.
Described in (excruciating) detail [here](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636). And further articulated in the [proposed documentation changes](https://github.com/nielash/rclone/blob/256892d85c0f13934988bd64731e19e119836be5/docs/content/bisync.md).

In summary:
* Fixed an [issue](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=1.%20Dry%20runs%20are%20not%20completely%20dry) causing dry runs to inadvertently commit filter changes
* Fixed an [issue](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=2.%20%2D%2Dresync%20deletes%20data%2C%20contrary%20to%20docs) causing `--resync` to erroneously delete empty folders and duplicate files unique to Path2
* `--check-access` is now enforced during `--resync`, preventing data loss in [certain user error scenarios](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=%2D%2Dcheck%2Daccess%20doesn%27t%20always%20fail%20when%20it%20should)
* Fixed an [issue](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=5.%20Bisync%20reads%20files%20in%20excluded%20directories%20during%20delete%20operations) causing bisync to consider more files than necessary due to overbroad filters during delete operations
* [Improved detection of false positive change conflicts](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=1.%20Identical%20files%20should%20be%20left%20alone%2C%20even%20if%20new/newer/changed%20on%20both%20sides) (identical files are now left alone instead of renamed)
* Added [support for `--create-empty-src-dirs`](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=3.%20Bisync%20should%20create/delete%20empty%20directories%20as%20sync%20does%2C%20when%20%2D%2Dcreate%2Dempty%2Dsrc%2Ddirs%20is%20passed)
* Added experimental `--resilient` mode to allow [recovery from self-correctable errors](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=2.%20Bisync%20should%20be%20more%20resilient%20to%20self%2Dcorrectable%20errors)
* Added [new `--ignore-listing-checksum` flag](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=6.%20%2D%2Dignore%2Dchecksum%20should%20be%20split%20into%20two%20flags%20for%20separate%20purposes) to distinguish from `--ignore-checksum`
* [Performance improvements](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=6.%20Deletes%20take%20several%20times%20longer%20than%20copies) for large remotes
* Documentation and testing improvements

Yes, mainly in this forum post:
https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636

It also fixes #6109 and addresses several other open issues: #6841 #5683 #5681 #5676 #5675 #5674

#### Checklist


ncw:
Would it be possible to break this down into a commit per fix/feature?

If the commit messages can link to either the forum or an rclone issue as well that would be perfect.

Reviewing a lot of changes all at once is very difficult :-(

I don't mind receiving this all as one pull request with lots of commits, but I'd suggest you start with a PR with one or a few of the simple fixes so we can get into the swing of things.

nielash:
Sure thing, sorry about that 😃

Btw, it looks like a lot more than it is, because the vast majority is docs and test file changes.

Also the tests all pass locally...looks like they are failing on GH because the test.log files got gitignored out of the golden directory. Will fix and resubmit. 😃

nielash:
Ok I broke this down into a commit per fix/feature and confirmed the [GitHub tests are passing now](https://github.com/nielash/rclone/actions/runs/4788341304).

Note that I left all the test changes together in one commit: https://github.com/rclone/rclone/commit/e1e5593bedd950a8ca750128c1d8d88924877a1c I don't think it's really possible to separate them due to the way the Bisync test suite is designed (it diffs an actual log with a golden log, so [just about any unrelated change you make will cause false positives](https://rclone.org/bisync/#updating-golden-results). I could make them each pass separately, but then they might not all pass together.)

If you want to start with just a few, I think these are the most important:

- https://github.com/rclone/rclone/pull/6971/commits/748d14bb9e9c378e0e01f35a00ed123ad0b8fb0a
- https://github.com/rclone/rclone/pull/6971/commits/eb3ee5593355171cccca91314b5eecc2a81788e4
- https://github.com/rclone/rclone/pull/6971/commits/d41cc7bae211784ebf7c1001cbaa7404eff18f97

But these are the simplest:

- https://github.com/rclone/rclone/pull/6971/commits/718d9d63ceea35573beab8a81932989250992329
- https://github.com/rclone/rclone/pull/6971/commits/bd1f31d4c70f51d25245da9beee50b381e19b771

cmd/bisync/deltas.go
```diff
@@ -290,6 +316,24 @@ func (b *bisyncRun) applyDeltas(ctx context.Context, ds1, ds2 *deltaSet) (change
 		if err != nil {
 			return
 		}
+
+		//propagate deletions of empty dirs from path2 to path1 (if --create-empty-src-dirs)
+		if b.opt.CreateEmptySrcDirs && !b.opt.Resync {
+			for _, s := range delete1.ToList() {
+				_, objerr := b.fs1.NewObject(ctx, s)
```
ncw, line 323: same comments as above!

cmd/bisync/deltas.go
```diff
@@ -299,6 +343,24 @@ func (b *bisyncRun) applyDeltas(ctx context.Context, ds1, ds2 *deltaSet) (change
 		if err != nil {
 			return
 		}
+
+		//propagate deletions of empty dirs from path1 to path2 (if --create-empty-src-dirs)
+		if b.opt.CreateEmptySrcDirs && !b.opt.Resync {
+			for _, s := range delete2.ToList() {
+				_, objerr := b.fs2.NewObject(ctx, s)
```
ncw, line 350: Duplicated code?

nielash:
Thanks so much for the excellent feedback! I'm happy to hear that you're considering this for the next release.

The `NewObject` thing is something I wondered about too, and wanted to ask your advice about. What would you recommend instead? Also, is there a simpler way of checking whether it's a file or a dir, without relying on `ErrorIsDir`? (As an aside, I'm presuming that bucket-based backends will not be using `--create-empty-src-dirs` in the first place, as the feature would be meaningless... but it does concern me that `NewObject` only returns `ErrorIsDir` "if possible without doing any extra work"...)

Regarding the tests -- I can do what you're requesting and I don't mind the extra work, but I do fear that the result you're looking for is not truly possible due to the unique way the bisync test suite is designed. I'll explain what I mean below, and if you're ok with this caveat I am happy to do it. (And apologies in advance if this is all stuff you already know -- just making sure we're on the same page.)

The bisync tests compare not only the resulting files but **also the debug log** to the respective golden versions. (This is why I had to remove `*.log` from `.gitignore`, as you noted.) [As documented here](https://rclone.org/bisync/#updating-golden-results), this means that just about anything you change will cause false positives, whether or not it's related to what's being tested. For this reason, the recommended workflow is to goldenize, check the difference carefully, and then commit if you're convinced that all the diffs are due either to intentional changes or false positives. (In my opinion, this is kind of a design flaw -- it has the upside of being an extremely sensitive test, but the downside of relying on humans to do the checking, and thus being more prone to human error. It is probably a little too easy, currently, for a future developer to accidentally turn a true positive into a false negative among all the noise.)

As an example, my "equality check" commit adds a line that prints `"Files are equal! Skipping."` to the log in certain scenarios. Let's assume for simplicity's sake that this is all it does. Every subsequent commit will then expect that line to be there, and fail if it is not. So, let's say that my next commit deals with something totally unrelated, and adds a new test for whatever it is. If you were to then revert the first commit, it would cause tests to fail for the second commit -- even though both commits passed all tests individually, and even though nothing is actually "broken". Removing `"Files are equal! Skipping."` from all the 'golden' logs might be enough to fix this, but git isn't going to know that, at least for the new test log that didn't exist prior to the second commit. (As another example: even merging the same commits in a different order would likely cause failures.)

So, I can give you a sequence of commits that each pass all their tests, but they will be dependent on their sequence, and therefore not necessarily cherry-pickable. If you were to later revert a commit from the middle of the pack, as you describe, **you'd likely have to also revert every commit that came after it** (or alternatively push a new commit to re-goldenize.) Does that make sense, and are you ok with that?

Assuming yes, what's your preferred way of receiving these revisions? Should I revert these commits and add new ones, or is it cleanest to just create a new branch?

All of your other comments made total sense and shouldn't be a problem.

Thanks!

nielash:
> Wonder if these extended changes might have an effect on #6572 too...

(I replied in the #6572 thread... TL;DR: I think it's fixed in the latest beta, and not specific to bisync.)

Also, @ncw just tagging you here in case you didn't see my previous post with some questions for you (no rush at all.)

callegar:
With the beta I am seeing this all the time:
```
ERROR : 01_EXAMINED/XXX/commenti_progetto.md.rclonelink: Failed to read link size: readlink /home/user/OneDrive/here/01_EXAMINED/XXX/commenti_progetto.md: invalid argument
```

ncw:
Sorry have been rather busy!

Will answer questions inline.

> Thanks so much for the excellent feedback! I'm happy to hear that you're considering this for the next release.

I think we might have to push until next release - sorry!

> The `NewObject` thing is something I wondered about too, and wanted to ask your advice about. What would you recommend instead? Also, is there a simpler way of checking whether it's a file or a dir, without relying on `ErrorIsDir`? (As an aside, I'm presuming that bucket-based backends will not be using `--create-empty-src-dirs` in the first place, as the feature would be meaningless... but it does concern me that `NewObject` only returns `ErrorIsDir` "if possible without doing any extra work"...)

Ideally you'd keep a the objects from the directory listing so you don't have to re-instantate them. This may not be possible of course, but in general `NewObject` should be avoided in favour of saving objects from listings. For most backends 1 `NewObject` takes about the same amount of time as 1 directory listing, and on some it takes longer (eg box).

> Regarding the tests -- I can do what you're requesting and I don't mind the extra work, but I do fear that the result you're looking for is not truly possible due to the unique way the bisync test suite is designed. I'll explain what I mean below, and if you're ok with this caveat I am happy to do it. (And apologies in advance if this is all stuff you already know -- just making sure we're on the same page.)
> 
> The bisync tests compare not only the resulting files but **also the debug log** to the respective golden versions. (This is why I had to remove `*.log` from `.gitignore`, as you noted.) [As documented here](https://rclone.org/bisync/#updating-golden-results), this means that just about anything you change will cause false positives, whether or not it's related to what's being tested. For this reason, the recommended workflow is to goldenize, check the difference carefully, and then commit if you're convinced that all the diffs are due either to intentional changes or false positives. (In my opinion, this is kind of a design flaw -- it has the upside of being an extremely sensitive test, but the downside of relying on humans to do the checking, and thus being more prone to human error. It is probably a little too easy, currently, for a future developer to accidentally turn a true positive into a false negative among all the noise.)

These are really what I'd call regression tests then...

> As an example, my "equality check" commit adds a line that prints `"Files are equal! Skipping."` to the log in certain scenarios. Let's assume for simplicity's sake that this is all it does. Every subsequent commit will then expect that line to be there, and fail if it is not. So, let's say that my next commit deals with something totally unrelated, and adds a new test for whatever it is. If you were to then revert the first commit, it would cause tests to fail for the second commit -- even though both commits passed all tests individually, and even though nothing is actually "broken". Removing `"Files are equal! Skipping."` from all the 'golden' logs might be enough to fix this, but git isn't going to know that, at least for the new test log that didn't exist prior to the second commit. (As another example: even merging the same commits in a different order would likely cause failures.)
> 
> So, I can give you a sequence of commits that each pass all their tests, but they will be dependent on their sequence, and therefore not necessarily cherry-pickable. If you were to later revert a commit from the middle of the pack, as you describe, **you'd likely have to also revert every commit that came after it** (or alternatively push a new commit to re-goldenize.) Does that make sense, and are you ok with that?

Yes that makes sense. I think the intellectual exercise of checking the diffs match what we expect after each change is probably worth doing. We can hopefully spot any problems with that commit.

Noted that it effectively makes the commits un cherry-pickable.

> Assuming yes, what's your preferred way of receiving these revisions? Should I revert these commits and add new ones, or is it cleanest to just create a new branch?

I'm happy for you to force push this branch with new commits. That is what I'd normally do. But it doesn't matter much to me so if you want to create a new branch which probably means a new PR that is fine. What I'd probably do is clone the branch as `bisync-fixes-and-features-old` to make a backup. Then I'd use revert all the code and use `git add -p` to add the parts of each bit, `git stash` then make the golden tests, check the diff then add to the commit. Then use git stash pop to get the remaining changes back and repeat for each feature. Then when done, force push here.

> All of your other comments made total sense and shouldn't be a problem.

:-)

nielash:
Very helpful, thank you! I think I've found a good solution that eliminates the need to use `NewObject`. It will take me some time to do all the testing, but I'll get it to you as soon as I can. Thanks!

docs/content/bisync.md
```diff
@@ -398,38 +498,97 @@ This will be solved in a future release, there is no workaround at the moment.
...
+and the default (no `--create-empty-src-dirs`) without running `--resync`. 
+This is because it may appear as though all directories (not just the empty ones) were created/deleted,
+when actually you've just toggled between making them visible/invisible to bisync. 
+It looks scarier than it is, but it's still probably best to stick to one or the other, 
+and use `--resync` when you need to switch.
 
 ### Renamed directories
 
-Renaming a folder on the Path1 side results is deleting all files on
+Renaming a folder on the Path1 side results in deleting all files on
```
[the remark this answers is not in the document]
nielash: Worth noting: #5690

nielash:
@ncw Ok, I think this is ready for you to take a look, whenever you have a chance. I think it addresses all of your comments. Each commit should now include all relevant documentation and testing changes, and each commit should compile and pass all tests.

The equality check and `--create-empty-src-dirs` commits are significantly revised to be much more efficient and avoid having to use `NewObject`. Which of the objects are dirs is now noted during the initial directory listing so we don't have to re-list for each object later. The equality check is now performed once, outside the loop, using the same underlying function as `check` (so it should be just as efficient.) We do the check only if we need to, and filter it for only the few files we're interested in.

Since this branch has gotten quite noisy with all the reversions, I also made a [new "clean" branch](https://github.com/nielash/rclone/tree/bisync-fixes-and-features-CLEAN) with just the "real" commits cherry-picked. Feel free to use whichever you prefer.

Thanks!

nielash:
(I [added a fix](https://github.com/rclone/rclone/pull/6971/commits/c00356b2f81dc4c17946ebb51a10ef1fce308cd7) for a small bug I found after a few weeks of real-world use.)

ncw:
Thanks @nielash for your hard work :-)

Can you rebase this on master and fix the conflict, please. (Its probably the fact I added a new parameter to the flags, you can just use `""` for the flag group).

If you can force push your clean branch over this one (with your new fix integrated into the correct commit + the rebase) - that would be my preferred way of working.

I'd like to merge this for v1.64 :crossed_fingers:

nielash:
I think it should be all set now 👍

ncw (approved):
Thank you @nielash that is a very nice set of patches :-)

I will merge them now.

Any chance you'd like to be the maintainer of rclone's `bisync` code from now on? Drop me an email to nick@craig-wood.com if you are interested :-)

nielash:
> Thank you @nielash that is a very nice set of patches :-)

Thanks! They've certainly improved my workflow and I hope others will find them helpful as well.

> Any chance you'd like to be the maintainer of rclone's bisync code from now on? Drop me an email to [nick@craig-wood.com](mailto:nick@craig-wood.com) if you are interested :-)

I'm definitely interested, depending on the time commitment you're looking for. I'll drop you an email 😃

PR #6976, opened 2023-04-25
webdav set modtime using propset for owncloud

WeidiDeng:

Rclone will set the modified time of the object instead of reuploading for owncloud.

Rclone forum [post](https://forum.rclone.org/t/rclone-sync-between-two-webdav-issues/14061/5).

This change is based on [owncloud issue](https://github.com/owncloud/core/issues/2542#issuecomment-15385897).

Tested against [owncloud quick evaluation](https://doc.owncloud.com/server/next/admin_manual/installation/docker/#quick-evaluation).

#### Checklist

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -1321,7 +1321,7 @@ func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
 			Method:     "PROPPATCH",
 			Path:       o.filePath(),
 			NoRedirect: true,
-			Body:       bytes.NewBuffer(fmt.Appendf(nil, owncloudPropset, modTime.Unix())),
+			Body:       strings.NewReader(fmt.Sprintf(owncloudPropset, modTime.Unix())),
 		}
 		var result api.Multistatus
 		var resp *http.Response
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -1342,6 +1342,8 @@ func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
 		}
 		// FIXME check if response is valid
 		if len(result.Responses) == 1 && result.Responses[0].Props.StatusOK() {
+			// update cached modtime
+			o.modTime = modTime
 			return nil
 		}
 		// fallback
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -175,6 +175,7 @@ type Fs struct {
 	precision          time.Duration // mod time precision
 	canStream          bool          // set if can stream
 	useOCMtime         bool          // set if can use X-OC-Mtime
+	propsetMtime       bool          // set if can use propset
 	retryWithZeroDepth bool          // some vendors (sharepoint) won't list files when Depth is 1 (our default)
 	checkBeforePurge   bool          // enables extra check that directory to purge really exists
 	hasOCMD5           bool          // set if can use owncloud style checksums for MD5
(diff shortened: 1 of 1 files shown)


backend/webdav/webdav.go
```diff
@@ -1299,8 +1301,51 @@ func (o *Object) ModTime(ctx context.Context) time.Time {
...
+`
+
 // SetModTime sets the modification time of the local fs object
 func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
+	if o.fs.propsetMtime {
+		opts := rest.Opts{
+			Method:     "PROPPATCH",
+			Path:       o.filePath(),
+			NoRedirect: true,
+			Body:       bytes.NewBuffer(fmt.Appendf(nil, owncloudPropset, modTime.Unix())),
```
ncw, line 1324: We can't use `fmt.Appendf` as we are still supporting go1.18

ncw:
This looks nice :-)

It also looks quite generic? Does it work with nextcloud or any of the other providers?

Have you tried the integration tests?

WeidiDeng:
I haven't tested with nextcloud or any other special flavors of webdav servers. Maybe it will work for nextcloud server? The two repos share the same commits [owncloud](https://github.com/owncloud/core/commit/2387fbd9ddfcbb992850ae2b61a2ccb26fd52c06) and [nextcloud](https://github.com/nextcloud/server/commit/d16dd1ad2aaeb322cd3b1f40dcf367de968cf701).

How do I write integration test and test them? I find a test file with those tests in it.

ncw:
> I haven't tested with nextcloud or any other special flavors of webdav servers. Maybe it will work for nextcloud server? The two repos share the same commits [owncloud](https://github.com/owncloud/core/commit/2387fbd9ddfcbb992850ae2b61a2ccb26fd52c06) and [nextcloud](https://github.com/nextcloud/server/commit/d16dd1ad2aaeb322cd3b1f40dcf367de968cf701).

Interesting. That would be good.

> How do I write integration test and test them? I find a test file with those tests in it.

In here https://github.com/rclone/rclone/blob/master/backend/webdav/webdav_test.go you'll see test blocks for OwnCloud and NextCloud

so if you run

    go test -v

in backend/webdav then it should run the tests against nextcloud and owncloud docker images.

For a fuller test, from the root of the source

    go install ./...
    test_all -backends webdav

This will run all the tests for the webdav backend and after a while open an HTML page with the results in.

I run these tests on Linux with docker installed. They will probably work on macOS, not sure about Windows.

These tests should test the setting/reading of modification times sufficiently already so you shouldn't need to write any more.

WeidiDeng:
I tested both nextcloud and owncloud.

Both failed set modtime test with **want 2011-12-13 14:15:16.999999999 +0000 UTC vs got 2001-02-03 04:05:06 +0000 GMT**. The message is the same. I'm wondering if this is a problem with oc/nc themselves.

WeidiDeng:
image

image

I see rclone only log this error and marking it as passed.

WeidiDeng:
I think it's because webdav caches modified time. When running the test, first rclone [creates](https://github.com/rclone/rclone/blob/v1.62-stable/fstest/fstests/fstests.go#L1439) a new file object. Webdav [caches](https://github.com/rclone/rclone/blob/v1.62-stable/backend/webdav/webdav.go#L621) metadata information (including modified time) by default.

When checking new updated modifed time, the file object [returns the old one](https://github.com/rclone/rclone/blob/v1.62-stable/backend/webdav/webdav.go#L1255). That's the problem.

There are several other tests that failed. But that's out of scope for this patch.

ncw (approved):

I will merge it now.

PR #6977, opened 2023-04-25
docs: clarify backends that use --no-update-modtime flag

joshtrichards:

The --no-update-modtime flag is supported by a only a couple of backends, even though many of the backends support modtime writing. Since the flag gets silently ignored, which ones support it should be made clear to users.

No

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -1576,6 +1576,11 @@ files if they are incorrect as it would normally.
 This can be used if the remote is being synced with another tool also
 (e.g. the Google Drive client).
 
+Currently only implemented for Openstack Swift, Oracle Object Storage,
+Amazon S3, and Microsoft Azure Blob Storage. If specified with other
+backends - even those that [support ModTime](/overview/) -
+the flag is silently ignored.
+
 ### --order-by string ###
 
 The `--order-by` flag controls the order in which files in the backlog


joshtrichards:
I went to use it during some WebDAV testing and then got to poking around.

Unless I'm mistaken, `UseServerModTime` is *only* checked for in the four backends mentioned in this PR in functions like `ModTime()` and `setMetadata()` (mostly the former).

Specifically 

- s3: https://github.com/rclone/rclone/blob/c4a9e480c9b41aa004a33471d2eb1112f8f6a846/backend/s3/s3.go#L4868
- azure: https://github.com/rclone/rclone/blob/c4a9e480c9b41aa004a33471d2eb1112f8f6a846/backend/azureblob/azureblob.go#L1566
- https://github.com/rclone/rclone/blob/c4a9e480c9b41aa004a33471d2eb1112f8f6a846/backend/swift/swift.go#L1226
- https://github.com/rclone/rclone/blob/c4a9e480c9b41aa004a33471d2eb1112f8f6a846/backend/oracleobjectstorage/object.go#L275

Other backends don't even look for the parameter and it's not handled anywhere outside the backends as far as I can tell. :-)

ncw:
You've patched the docs for `--no-update-modtime` as per the title of this PR, but I think you meant to patch the docs for `--use-server-modtime`?

ncw:
I patched the correct bit of docs here with updated info: 9f9569ad6fb8efec0787f179ceea2637b14f7e47

Thank you

PR #6988, opened 2023-05-04
sia: switch to renterd

redsolver:

The current Sia integration in rclone uses the **siad** API (https://github.com/SiaFoundation/siad).

The Sia Foundation is working on a new next-generation Sia renter called **renterd** (https://github.com/SiaFoundation/renterd). This PR changes the Sia backend to support the new renterd API. 

Changing the implementation would break existing rclone setups still using siad, so it might make sense to instead create a new backend or add a migration process. I'm open for suggestions.

No.

#### Checklist

### DIFF
--- a/backend/sia/api/types.go
+++ b/backend/sia/api/types.go
@@ -1,23 +1,13 @@
 // Package api provides types used by the Sia API.
 package api
 
-import (
-	"strings"
-	"time"
-)
+import "strings"
 
-// DirectoriesResponse is the response for https://sia.tech/docs/#renter-dir-siapath-get
-type DirectoriesResponse struct {
-	Directories []DirectoryInfo `json:"directories"`
-	Files       []FileInfo      `json:"files"`
+type FileInfo struct {
+	Name string `json:"name"`
+	Size uint64 `json:"size"`
 }
 
-// FilesResponse is the response for https://sia.tech/docs/#renter-files-get
-type FilesResponse struct {
-	Files []FileInfo `json:"files"`
-}
-
-// FileResponse is the response for https://sia.tech/docs/#renter-file-siapath-get
 type ObjectResponse struct {
 	Object ObjectInfo `json:"object"`
 }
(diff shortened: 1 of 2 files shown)


ncw:
> The current Sia integration in rclone uses the **siad** API (https://github.com/SiaFoundation/siad).
> 
> The Sia Foundation is working on a new next-generation Sia renter called **renterd** (https://github.com/SiaFoundation/renterd). This PR changes the Sia backend to support the new renterd API.
> 
> Changing the implementation would break existing rclone setups still using siad, so it might make sense to instead create a new backend or add a migration process. I'm open for suggestions.

What is the status of siad vs renterd? If rclone switched from one to the other what would users need to do?

At the very minimum we'd want to have a migration process documented.

I read on the renterd page

> All of the key low-level APIs have been implemented and are ready for use. However, renterd currently does not ship with a UI or autopilot functionality, and it lacks backwards-compatibility with siad renter metadata. This means that, while renterd is already capable of serving as the backbone for new Sia applications, most users should continue to use siad.

That doesn't inspire confidence! Also I didn't really understand the downsides - can you elaborate?

Creating a new backend might be a good idea - not sure - it depends on the answers to the above! If we did create a new backend how long would we need to keep the old one around for?

redsolver:
It will still take some months until `renterd` reaches a stable release and some users are still using the old `siad` with rclone. But all new projects are building on top of `renterd` and many in the Sia community would like to use rclone with the new `renterd` now. After the first stable `renterd` release, I expect almost all users to migrate away from `siad` because it brings major improvement in almost all areas.

So maybe the best solution would be to introduce this PR as a new `renterd` storage backend and change the name to `sia` later (including removing the old implementation) in a year or so when we're sure nobody is using `siad` with rclone anymore.

multikatt:
Just wanted to mention I've been using your fork for around a month now, @redsolver, and it's working really great. The only real hassle I had was with renaming/moving files and not being able to remove folders. I've got a fork of your fork here [multikatt/rclone/tree/sia ](https://github.com/multikatt/rclone/tree/sia) where i've tried to solve that using the new rename endpoint.

redsolver:
@multikatt cool, could you create a PR with your fixes on my forked repo?

spartanz51:
I think we will always have this middle ground in the medium term, from experience. Since we need to remain compatible with the current Sia daemon, not everyone will migrate. 
Additionally, we need to be compatible with the new daemon and the new projects built on top of it, like you said. 

Therefore, the only solution is to create a new entry for Sia with the client 'renterd,' so that we have two entries: 'Siad' and 'Sia' (renterd)

So, I have committed 'Siad' under the 'siad' module, in your fork, and the default 'Sia' should be 'renterd.'
https://github.com/redsolver/rclone/pull/2

@redsolver Can you please review and merge my pull request into this one, along with the modifications made by @multikatt ? https://github.com/redsolver/rclone/pull/1

Thanks!

ChrisSchinnerl:
FYI we have rolled out S3 support in `renterd` a few days ago. It doesn't have support for authentication through v4 sigs yet but that's just around the corner. 
Meaning we don't really need to maintain this integration here and could instead update the docs with instructions on how to use rclone's S3 integration with Sia through `renterd`.
Of course we could also support the native integration if you think that is preferable.

ncw:
As I said above, I'm happy to merge this if the sia experts agree it is a good idea. I know nothing about sia so I'm not in a fit state to judge.

However if it isn't backwards compatible we need a documented migration path and the backend should produce an ERROR when run with a link to the migration procedure in the docs.

I'd probably prefer that, than having two backends to maintain!

redsolver:
thanks for the PRs @spartanz51 , I merged the first one. for the second one, I think it makes more sense to keep the old implementation in the "sia" namespace (so existing configurations continue working) and rename the new one to be 
"renterd"

n8maninger:
A much-needed update. Sia now supports S3 natively. Going forward, that's our recommended method for using `rclone`. We'll likely want to mark the Sia gateway as deprecated. We'll add migration info to the docs as well.

Issue #3147, opened 2019-04-29
Extend rclone serve webdav to support the owncloud extensions for modtime and checksums


jcasale:
#### What is your current rclone version (output from `rclone version`)?
1.47

#### What problem are you are trying to solve?
The native webdav implementation does not support modtime and checksums when synchronizing. As a result, the reliability of the transfer is compromised. The issue is highly visible when syncing git repos, after the initial transfer, all subsequent syncs are invalid.

#### How do you think rclone should be changed to solve that?
Extend `rclone serve webdav` to support the owncloud extensions for modtime and checksums.

ncw:
This should be relatively straight forward to implement the owncloud extensions.

The first is the modtime for the uploaded file is delivered in a header like this

```
	if f.useOCMtime {
		opts.ExtraHeaders["X-OC-Mtime"] = fmt.Sprintf("%f", float64(src.ModTime().UnixNano())/1E9)
	}
```

And the second is to return the checksum of the file in some extra properties the `oc:checksums` here.  This should be possible too - the webdav server library is capable of adding custom properties.

```xml
<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns">
 <d:prop>
  <d:displayname />
  <d:getlastmodified />
  <d:getcontentlength />
  <d:resourcetype />
  <d:getcontenttype />
  <oc:checksums />
 </d:prop>
</d:propfind>
```

gary-kim:
I'm working on the FTPS support right now but if noone's taken this by the time that gets merged, I'll also have a go at this.

ncw:
> I'm working on the FTPS support right now but if noone's taken this by the time that gets merged, I'll also have a go at this.

That would be super :-)  Let me know if you need any help.  I'm not 100% sure how easy or not it will be.

gary-kim:
Also, I noticed while looking at this, is go1.8 support being officially dropped? If so, we can change the `"context"` imports.

ncw:
> Also, I noticed while looking at this, is go1.8 support being officially dropped? If so, we can change the `"context"` imports.

Looks like I missed some things when deprecating go1.8! I'll patch these shortly as there are quite a few more than just `serve/webdav`!

ncw:
This has been implemented already so I'll close this issue.

ncw:
[...]

PR #6990, opened 2023-05-05
implement owncloud checksum and modtime extensions for webdav server

WeidiDeng:

Implement owncloud extensions for webdav server

[3147](https://github.com/rclone/rclone/issues/3147)

#### Checklist

### DIFF
--- a/cmd/serve/webdav/webdav_test.go
+++ b/cmd/serve/webdav/webdav_test.go
@@ -65,7 +65,7 @@ func TestWebDav(t *testing.T) {
 		// Config for the backend we'll use to connect to the server
 		config := configmap.Simple{
 			"type":   "webdav",
-			"vendor": "other",
+			"vendor": "owncloud",
 			"url":    w.Server.URLs()[0],
 			"user":   testUser,
 			"pass":   obscure.MustObscure(testPass),


ncw (commented):
This code looks great!

My only questions are  about the testing.

This is tested by looping back the webdav backend and running the webdav backend tests on the webdav server.

However the backend created in the tests won't be expecting modtimes or hashes to work so it won't test them.

So I think we need to adjust the backend so it knows an rclone webdav server has hashes and modtimes. So maybe we need an `rclone` provider there which sets the correct quirks to enable hashes and modtimes?

This will be backwards compatible since we don't have an Rclone provider already.

Thoughts?

WeidiDeng:
When I worked on the fix for the vfs cache, I think that test is using local file system. I tested this patch locally using local as backend. I'll see what can be done.

WeidiDeng:
How do I write tests for this? I can't find how to test checksum and modtime. `TestWebDav` only contains how to create a backend webdav server, not specifying the features that should be tested.

ncw:
You'll need to add an rclone provider to the WebDAV backend which sets the correct quirks to use modtime and checksums.

Then change the test to add the provider string.

The tests will then test the modtime and checksums.

WeidiDeng:
image

Isn't this creating a webdav backend? Where should I put the test code?

ncw (approved):
Code looks good, test looks good - let's merge :-)

Thank you

Jwink3101:
I hope it is okay to say this here, but I am really grateful for this feature! When communicating rclone-to-rclone, SFTP is just way too slow so I have to choose between WebDAV speed or ModTime support. Now, once 1.63 is released, this will greatly improve my options!

PR #7005, opened 2023-05-15
Update parent dir modtime on vfs actions

Limero:

I noticed that when doing actions on files/directories in a rclone mounted directory that the modtime on the parent directory was never updated. [lf](https://github.com/gokcehan/lf) relies on this information to know when to refresh a directory, but as it was never updated, any changes to files would not show up. [ranger](https://github.com/ranger/ranger) seems to have the same problem.

This PR replicates the normal unix behavior (same as sshfs) which is to update the modtime of the parent directory when creating/renaming/removing files/directories.

#### Checklist

### DIFF
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -923,6 +923,10 @@ func (d *Dir) Create(name string, flags int) (*File, error) {
 	if d.vfs.Opt.ReadOnly {
 		return nil, EROFS
 	}
+	if err = d.SetModTime(time.Now()); err != nil {
+		fs.Errorf(d, "Dir.Create failed to set modtime on parent dir: %v", err)
+		return nil, err
+	}
 	// This gets added to the directory when the file is opened for write
 	return newFile(d, d.Path(), nil, name), nil
 }
@@ -957,6 +961,10 @@ func (d *Dir) Mkdir(name string) (*Dir, error) {
 	fsDir := fs.NewDir(path, time.Now())
 	dir := newDir(d.vfs, d.f, d, fsDir)
 	d.addObject(dir)
+	if err = d.SetModTime(time.Now()); err != nil {
+		fs.Errorf(d, "Dir.Mkdir failed to set modtime on parent dir: %v", err)
+		return nil, err
+	}
 	// fs.Debugf(path, "Dir.Mkdir OK")
 	return dir, nil
 }
@@ -1026,6 +1034,10 @@ func (d *Dir) RemoveName(name string) error {
 	node, err := d.stat(name)
 	if err != nil {
 		fs.Errorf(d, "Dir.Remove error: %v", err)
+		return err
+	}
+	if err = d.SetModTime(time.Now()); err != nil {
+		fs.Errorf(d, "Dir.Remove failed to set modtime on parent dir: %v", err)
 		return err
 	}
 	return node.Remove()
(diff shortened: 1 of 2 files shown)


ncw (commented):
This code looks sensible to me.

Do you think that it is a problem that the modtimes are not being synced back to the directories on the cloud storage?

Not all cloud storage systems support directory modtimes but rclone will read those that do. Rclone does not have any support for setting directory modtimes yet.

The consequence of this will be if you stop rclone and restart it then the directory modification times will reset to what they were before.

Also if the directories fall out of the directory cache then the modtimes will reset when they are loaded back in.

This patch will help with compatibility but I wonder if it is giving a false sense of security?

What do you think?

Limero:
> Do you think that it is a problem that the modtimes are not being synced back to the directories on the cloud storage?

I thought this was the initial problem, that modtimes were not updated on the server. But after mounting a SFTP directory and checking with "stat" in a separate shell over SSH, the modtimes are in fact being updated. So what this fix will do is give a better representation on the actual modtimes of the server.

So at least for SFTP, stopping rclone and restarting it will reset modtimes to the new times.

Even if it's not 100%, I'd say this fixes most issues with modtimes, but there might be edge cases like you mention.


Btw, I see the tests are failing on Windows. Probably because they handle modtimes differently. For making the tests pass, do you have a better solution than just doing?
```
assert.True(t, dir.ModTime().After(origModTime) || runtime.GOOS == "windows")
```

ncw:
> > Do you think that it is a problem that the modtimes are not being synced back to the directories on the cloud storage?
> 
> I thought this was the initial problem, that modtimes were not updated on the server. But after mounting a SFTP directory and checking with "stat" in a separate shell over SSH, the modtimes are in fact being updated. So what this fix will do is give a better representation on the actual modtimes of the server.

That isn't rclone updating the modtimes of the directories though, it is the normal unix modtime thing for directories going on. So likely a remount the directory times will change..

> So at least for SFTP, stopping rclone and restarting it will reset modtimes to the new times.

The times will be about right, but they won't be exactly the same is my guess.

Can you check with `ls -l --full-time` and see whether the fractional digits of the times change in the mount when you unmount and remount?

> Even if it's not 100%, I'd say this fixes most issues with modtimes, but there might be edge cases like you mention.

It is probably an improvement, I would agree, I just wanted us to think about the consequences.

> Btw, I see the tests are failing on Windows. Probably because they handle modtimes differently. For making the tests pass, do you have a better solution than just doing?

I don't see why these tests are failing... Possibly because the time resolution is lower on Windows. Maybe it needs to check if the time is >= rather than just > ? 

```go
assert.True(t, !dir.ModTime().Before(origModTime))
```

Limero:
You're right that the modtimes differ very slightly. It also seems that once remounting, the decimals are wiped (on everything mounted through rclone, not specific to this scenario).

Checking rclone mounted dir:
2023-05-21 19:13:10.644471653

Checking directly over SSH:
2023-05-21 19:13:10.802628070

Remounting and checking:
2023-05-21 19:13:10.000000000

Regarding the tests, checking if ">=" instead of ">" would give false positives on cases where modtimes are not updated at all (like with the previous code).

ncw:
I've rebased, squashed and tweaked the windows tests on this. Hopefully the CI will pass and I'll merge!

ncw:
Thank you :-)

PR #7021, opened 2023-05-22
swift: add dirctory marker support #3453

joec4i:

Support --swift-directory-markers .

I also included a change to `TestSwiftAIO` to allow retries. That fixes occasional errors because the `swift-aio` container wasn't cleaned up properly between the first and the second integration test runs.

Please let me know if I miss anything. Thanks!

#3453 

#### Checklist

### DIFF
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -216,6 +216,11 @@ It should be set to true for resuming uploads across different sessions.`,
 			Default:  false,
 			Advanced: true,
 		}, {
+			Name:     "directory_markers",
+			Default:  false,
+			Advanced: true,
+			Help:     `If true, create directory markers with the content type application/directory for Mkdir requests.`,
+		}, {
 			Name: "storage_policy",
 			Help: `The storage policy to use when creating a new container.
 
@@ -263,6 +268,7 @@ type Options struct {
 	NoChunk                     bool                 `config:"no_chunk"`
 	NoLargeObjects              bool                 `config:"no_large_objects"`
 	Enc                         encoder.MultiEncoder `config:"encoding"`
+	DirectoryMarkers            bool                 `config:"directory_markers"`
 }
 
 // Fs represents a remote swift server
@@ -384,7 +390,7 @@ func parsePath(path string) (root string) {
 // split returns container and containerPath from the rootRelativePath
 // relative to f.root
 func (f *Fs) split(rootRelativePath string) (container, containerPath string) {
-	container, containerPath = bucket.Split(path.Join(f.root, rootRelativePath))
+	container, containerPath = bucket.Split(bucket.Join(f.root, rootRelativePath))
 	return f.opt.Enc.FromStandardName(container), f.opt.Enc.FromStandardPath(containerPath)
 }
 
@@ -506,6 +512,9 @@ func NewFsWithConnection(ctx context.Context, opt *Options, name, root string, c
 		BucketBasedRootOK: true,
 		SlowModTime:       true,
 	}).Fill(ctx, f)
+	if opt.DirectoryMarkers {
+		f.features.CanHaveEmptyDirectories = true
+	}
 	if f.rootContainer != "" && f.rootDirectory != "" {
 		// Check to see if the object exists - ignoring directory markers
 		var info swift.Object
@@ -624,13 +633,13 @@ func (f *Fs) listContainerRoot(ctx context.Context, container, directory, prefix
 			objects, err = f.c.Objects(ctx, container, opts)
 			return shouldRetry(ctx, err)
 		})
+
+		dirs := make(map[string]bool)
+
 		if err == nil {
 			for i := range objects {
 				object := &objects[i]
-				isDirectory := false
-				if !recurse {
-					isDirectory = strings.HasSuffix(object.Name, "/")
-				}
+				isDirectory := strings.HasSuffix(object.Name, "/") || object.ContentType == directoryMarkerContentType
 				remote := f.opt.Enc.ToStandardPath(object.Name)
 				if !strings.HasPrefix(remote, prefix) {
 					fs.Logf(f, "Odd name received %q", remote)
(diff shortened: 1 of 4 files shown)


backend/swift/swift.go
```diff
@@ -201,6 +201,11 @@ func init() {
 It should be set to true for resuming uploads across different sessions.`,
 			Default:  false,
 			Advanced: true,
+		}, {
+			Name:     "directory_markers",
+			Default:  false,
+			Advanced: true,
+			Help:     `If true, create directory markers with the content type application/directory for Mkdir requests.`,
```
ncw, line 208: Is this the swift way? Should the markers end with trailing / ? Be zero sized?

backend/swift/swift.go
```diff
@@ -646,6 +652,8 @@ type addEntryFn func(fs.DirEntry) error
 
 // list the objects into the function supplied
 func (f *Fs) list(ctx context.Context, container, directory, prefix string, addContainer bool, recurse bool, includeDirMarkers bool, fn addEntryFn) error {
+	// Mark true if anything is found under given directory
+	itemsFound := false
```
ncw, line 656: This is called foundItems in my patch

ncw (commented):
Sorry I missed this for a while!

Thanks for doing this.

Can you try to make it follow the patches I did for gcs **exactly** please?

Attached is a patch file with the accumulated patches in.

[gcs.patch.gz](https://github.com/rclone/rclone/files/11712708/gcs.patch.gz)

Please can you make your patch as close as possible to that patch. This makes maintenance of the bucket based backends much easier which have a lot of duplicated logic. I know, not ideal but the backends are just different enough to make factoring things into a common library very hard.

Can you also drop the SwiftAIO patches from this. They should probably be in a separate PR and I'd like to discuss why they are needed because I haven't seen a problem it.

If you then force push an update I'll review.

Thank you

backend/swift/swift.go
joec4i: Yes, it is. `application/directory` is the preferred content type and the trailing slash is not required - https://github.com/openstack/python-swiftclient/blob/e65070964c7b1e04119c87e5f344d39358780d18/swiftclient/service.py#L213 .

It's also worth mentioning that Swift treats `b`, `b/`, `b//`, and `b///`  as different objects, so I think it's better to just remove all trailing slashes to prevent conflicts. 

This simple go program creates `a/b`, `a/b/` and `a/b//` under a container `foo` https://gist.github.com/joec4i/446e2cc4deb9bc276dc8fdb37559f7a9 . 

The swift CLI shows them as 3 different objects.
```
$ swift -A http://127.0.0.1:8080/auth/v1.0 -U test:tester -K testing list foo
a/b
a/b/
a/b//
```

backend/swift/swift.go
joec4i: That's updated now.

fstest/testserver/init.d/TestSwiftAIO
```diff
@@ -7,9 +7,15 @@ NAME=swift-aio
 . $(dirname "$0")/docker.bash
 
 start() {
+    # When running the second set of integration tests against SwiftAIO, we need to wait for the first run to stop.
+    if docker inspect ${NAME} >/dev/null 2>&1; then
+        echo "Waiting for the previous run to stop..."
+        sleep 3
+    fi
+
```
joec4i, line 15: This is needed between now we are running two integration test suites in a row. In my Linux vm it fails because docker is not able to remove the container from the first run in time:
```
    --- SKIP: TestIntegration/FsShutdown (0.00s)
=== RUN   TestIntegrationWithDirectoryMarkers
    fstests.go:423: Using remote "TestSwiftAIO:"
    fstests.go:433:
        	Error Trace:	/Users/joe.cai/git/rclone/fstest/fstests/fstests.go:433
        	            				/Users/joe.cai/git/rclone/backend/swift/swift_test.go:33
        	            	failed to run /Users/joe.cai/git/rclone/fstest/testserver/init.d/TestSwiftAIO start
        	            	docker: Error response from daemon: Conflict. The container name "/swift-aio" is already in use by container "d5b6cdaae65058d4159bf1f79013239d49ab910a5102df513762fab230dfef9e". You have to remove (or rename) that container to be able to reuse that name.
        	            	See 'docker run --help'.
        	            	: exit status 125
        	Test:       	TestIntegrationWithDirectoryMarkers
```

backend/swift/swift.go
```diff
@@ -632,6 +641,16 @@ func (f *Fs) listContainerRoot(ctx context.Context, container, directory, prefix
...
 				}
+
+				if isDirectory {
+					// Swift returns duplicate directories when a directory marker exists, hence dedupe is required
+					remoteTrimmed := strings.TrimRight(remote, "/")
+					if _, found := dirs[remoteTrimmed]; found {
+						continue
+					}
+					dirs[remoteTrimmed] = true
+				}
```
joec4i, line 652: This can probably be avoided if the Swift library doesn't return duplicated objects. If I have a directory marker `empty-dir`, Swift will return the following payload:

```
[
  {
    "bytes": 0,
    "hash": "d41d8cd98f00b204e9800998ecf8427e",
    "name": "empty-dir",
    "content_type": "application/directory",
    "last_modified": "2023-06-14T11:49:19.990560"
  },
  {
    "subdir": "empty-dir/"
  }
]
```
Please find the request log [here](https://gist.github.com/joec4i/b79afd610d96457d6c62eb68f315b6ef). 

At the moment, the go Swift library will translate the response into two `swift.Object`s with the same name. So we need the deduplication step here to avoid outputting duplicated entries.

[The Swift client library](https://github.com/ncw/swift) is probably a better place to dedupe though.

joec4i:
Hi @ncw ! I have updated the patch. Please let me know if I miss anything. Thanks!

joec4i:
Hey @ncw ! I noticed that you just closed #3453 , but I'm wondering if we can get this in for swift support as well. Please let me know if I should have this patch updated just so it's in sync with implementations for other backends. 🙏

ncw:
Hi @joec4i, and thank you for this contribution.

PR #7028, opened 2023-05-26
http: add client certificate user auth middleware

pdf:

This populates the authenticated user from the client certificate
common name.

Also added tests for the existing client certificate functionality.

This was briefly mentioned in #3249 and tabled for a later update.

#### Checklist


pdf:
I couldn't work out from the contributing docs whether I need to go and update all of the markdown documents for the various subcommands, or if just updating the help text within the modified package was enough - when I ran `make doc` it produced a significant diff that included formatting changes etc in the HTML output.

lib/http/auth.go
```diff
@@ -31,6 +32,10 @@ To create an htpasswd file:
 
 The password file can be updated while rclone is running.
 
+Use ` + "`--{{ .Prefix  }}mtls`" + ` to enable retrieving the username from the common name field
```
ncw, line 35: Presumably this flag also needs the --client-ca flag? This should certainly be mentioned here.

We should probably try to straighten out the terminology as rclone refers to these as client certificates not mTLS (which is the server + the client certificate as far as I understand it).

So what you are saying is the --mtls flag will enable parsing of the common name field from the client certificate as the user, assuming I'm understanding correctly.

So this flag is more like

  --client-ca-contains-username

Maybe we should be doing this by default if `--client-ca` is supplied? Since it is overridden by basic auth.

lib/http/server.go
```diff
@@ -293,12 +293,19 @@ func NewServer(ctx context.Context, options ...Option) (*Server, error) {
...
 func (s *Server) initAuth() {
+	s.usingAuth = false
+
 	if s.auth.CustomAuthFn != nil {
 		s.usingAuth = true
 		s.mux.Use(MiddlewareAuthCustom(s.auth.CustomAuthFn, s.auth.Realm))
 		return
 	}
 
+	if s.auth.MTLS {
```
ncw, line 304: This depends on client CA being provided too

ncw (commented):
I think this looks like a great feature - thank you :-)

You did the right thing not checking in the results of `make docs` - I do that just before a release normally.

See inline for queries.

lib/http/server.go
pdf: Enabling this without enabling client CA will result in unauthorized requests, but should we refuse to enable the option here in that case? And if so, do we warn about it? Perhaps we should error and exit earlier if cert usernames are enabled but client-ca is not?

If we decide that certs as usernames should be the default when client-ca is provided then this logic will need to change in any case.

lib/http/auth.go
pdf: The CLI flag docs mention that it depends on `--client-ca` being enabled, but makes sense to mention it in the general docs too.

I see no problem with changing the flag to something more descriptive or in-line with existing flags.

I can send changes for the above.

Enabling this functionality by default when `--client-ca` is enabled *may* result in a behaviour change for users that are currently using `--client-ca`, but not basic auth. It's hard for me to gauge whether this is likely or not though, since I'm not super-familiar with all the use-cases for basic auth (primarily I want this for restic, so that we don't have to rely on htpasswd for clients).

If we were to enable it by default, would there be a benefit in providing a flag to disable it?

ncw: I just did a little trace through the code to see where we use the user from the auth to see what the impact of enabling it by default would be.

We use it in `serve restic` which is what you are familiar with I think.

We also use it in the auth proxy.

However the auth proxy is reading the user direct from the basic auth in the request. The auth proxy installs a custom auth function which is called with the `username` and `password`

https://github.com/rclone/rclone/blob/51a468b2bae4ca8e21760435211623a8199a9167/lib/http/middleware.go#L100

My point being that it doesn't read the `ctxKeyUser` and so won't be using the user set in the common name of the client certificate.

It probably should do that if there is no basic auth and call the auth proxy with the username and a blank password.

So should we enable it by default? Any user that has gone to the trouble of making client certificates is highly technical and likely to have read the documentation so adding a flag to enable it is not a problem. I guess this could break existing setups if they are using client certs and basic auth. Maybe we could disable the reading from the Common name if any of the basic auth flags are set? I think that would be backwards compatible.

Sorry long post... In summary
1. Need to fix up the custom auth function
2. Disable the Common Name reading if any basic auth options are set should be enough

What do you think?

lib/http/server.go
ncw: If we decide to have a flag then we should error if `--client-ca` is not provided.

lib/http/auth.go
pdf: Sorry for the delay, other priorities.

I think enabling user from cert by default is fine, and relatively safe if disabled when basic auth is enabled.

Change incoming.

lib/http/server.go
pdf: Resolved by removing the flag and enabling by default.

pdf:
Ready for review

ncw (approved):

Apologies for the delay , will merge now :-)

Issue #7037, opened 2023-06-02
rclone will deadlock after all checkers exit by error


douchen:
#### What is the problem you are having with rclone?
rclone will deadlock and the program will block at `remotes <- remote` when:

1. all checkers exit
2. remotes are not fully consumed

https://github.com/rclone/rclone/blob/51a468b2bae4ca8e21760435211623a8199a9167/fs/filter/filter.go#L426-L463

#### What is your rclone version (output from `rclone version`)
```
rclone v1.62.2-DEV
- os/version: darwin 13.3.1 (64 bit)
- os/kernel: 22.4.0 (arm64)
- os/type: darwin
- os/arch: arm64 (ARMv8 compatible)
- go/version: go1.20.4
- go/linking: dynamic
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

macOS 13.3.1, arm64

#### Which cloud storage system are you using? (e.g. Google Drive)

no

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
`go test -timeout 30s -run ^TestNewFilterMakeListR$ github.com/rclone/rclone/fs/filter -count=1 -v`

https://github.com/rclone/rclone/blob/51a468b2bae4ca8e21760435211623a8199a9167/fs/filter/filter_test.go#L359-L363

add code at TestNewFilterMakeListR function end:
```go
	// The checker will exit by the error above
	ci := fs.GetConfig(context.Background())
	ci.Checkers = 1

	// Now check an error is returned from NewObject
	require.NoError(t, f.AddFile("error"))
	err = listR(context.Background(), "", listRcallback)
	require.EqualError(t, err, assert.AnError.Error())
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

no

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.63 - thank you :-)

PR #7039, opened 2023-06-02
fs: fix deadlock error when all checkers exit

douchen:

fix deadlock

#7037

#### Checklist

### DIFF
--- a/fs/filter/filter.go
+++ b/fs/filter/filter.go
@@ -433,7 +433,7 @@ func (f *Filter) MakeListR(ctx context.Context, NewObject func(ctx context.Conte
 		var (
 			checkers = ci.Checkers
 			remotes  = make(chan string, checkers)
-			g        errgroup.Group
+			g, gCtx  = errgroup.WithContext(ctx)
 		)
 		for i := 0; i < checkers; i++ {
 			g.Go(func() (err error) {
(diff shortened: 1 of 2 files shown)


fs/filter/filter.go
```diff
@@ -433,7 +433,7 @@ func (f *Filter) MakeListR(ctx context.Context, NewObject func(ctx context.Conte
 		var (
 			checkers = ci.Checkers
 			remotes  = make(chan string, checkers)
-			g        errgroup.Group
+			g, gCtx  = errgroup.WithContext(ctx)
```
ncw, line 436: If we are making a gCtx then we should use it instead of ctx in call to `NewObject` below

fs/filter/filter.go
```diff
@@ -454,10 +454,17 @@ func (f *Filter) MakeListR(ctx context.Context, NewObject func(ctx context.Conte
 				return nil
 			})
 		}
-		for remote := range f.files {
-			remotes <- remote
-		}
-		close(remotes)
+		go func() {
```
ncw, line 457: I think this can be written without the extra go routine which I think makes the logic a bit simpler

```go
	outer:
		for remote := range f.files {
			select {
			case remotes <- remote:
			case <-gCtx.Done():
				break outer
			}
		}
		close(remotes)
```

ncw (commented):
Great fix thank you :-)

I've put some thoughts and suggestions inline.

fs/filter/filter.go
douchen: This has been updated

fs/filter/filter.go
douchen: Thanks for the suggestion, it was very helpful

ncw (approved):
This looks great now - thank you

PR #7045, opened 2023-06-06
`rclone backend restore s3:bucket --no-traverse` storageClass should not be checked

vdm:
`rclone backend restore s3:bucket --no-traverse --files-from` object keys are being skipped, with Status: "Not GLACIER or DEEP_ARCHIVE storage class".

This changes the storageClass nil check to not cause the object key to be skipped.

It also adds INTELLIGENT_TIERING as a storage class to check, because RestoreObject can be used with this storage class.

No.

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -4761,9 +4761,19 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
 				st.Status = "Not an S3 object"
 				return
 			}
-			if o.storageClass == nil || (*o.storageClass != "GLACIER" && *o.storageClass != "DEEP_ARCHIVE") {
-				st.Status = "Not GLACIER or DEEP_ARCHIVE storage class"
-				return
+			if !o.fs.ci.NoTraverse {
+				if o.storageClass == nil {
+					st.Status = "storageClass == nil, skipped"
+					return
+				}
+				if *o.storageClass == "STANDARD" {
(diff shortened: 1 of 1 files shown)


backend/s3/s3.go
```diff
@@ -4517,8 +4517,8 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
 				st.Status = "Not an S3 object"
 				return
 			}
-			if o.storageClass == nil || (*o.storageClass != "GLACIER" && *o.storageClass != "DEEP_ARCHIVE") {
-				st.Status = "Not GLACIER or DEEP_ARCHIVE storage class"
+			if o.storageClass != nil && (*o.storageClass != "GLACIER" && *o.storageClass != "DEEP_ARCHIVE" && *o.storageClass != "INTELLIGENT_TIERING") {
```
ncw, line 4520: This changes the meaning slightly if the storage class isn't set.

Before an unset storage class was considered to be not restorable, but with this change it will be restored.

I'm not sure how likely an unset storage class is - maybe this is academic - what do you think?

ncw (commented):
Looks very useful thank you.

I put a query inline.

backend/s3/s3.go
vdm: Not academic; this would have changed semantics. I added a --no-traverse check.

ncw:
Hi @vdm, and thank you for this contribution.

Issue #6540, opened 2022-11-05
Remotes defined with env vars are not listed using config/listremotes via rcd


elthariel:
#### The associated forum post URL from `https://forum.rclone.org`

I have not found any mention of that issue on the forum, or on existing issues

#### What is the problem you are having with rclone?

I'm using `rclone rcd` on a kubernetes cluster to move my backups around and out of the cluster.

The rclone rcd have the following env vars:
```
env | grep RCLONE_CONFIG
RCLONE_CONFIG_INTERNAL_PROVIDER=Minio
RCLONE_CONFIG_INTERNAL_TYPE=s3
RCLONE_CONFIG_INTERNAL_ACCESS_KEY_ID=bH0t8Cb4b8UrPpXV
RCLONE_CONFIG_INTERNAL_SECRET_ACCESS_KEY=GMBGlXPVsLzCzzQnohBL86Il
RCLONE_CONFIG_INTERNAL_ENDPOINT=https://minio.foundation
```
> Those are temporary credentials, this is not a leak :)

Despite the variables being defined correctly running `rclone rc config/listremotes` returns an empty array.

If I use `--rc-serve`, the remote is correctly listed and working.
If I try to access the remote using `operations/list` it works correctly.

I've found this extremely confusing and spent a few hours trying to debug something that was actually working.

It seems the function `config.FileSections()` is not called when the `listremotes` operation is called via rc.

So essentially my question is: is it a bug or a feature ?If this is expected behavior, I'd like to see something added in the doc to save the next fellow the few hours I've lost. If this is a bug, let's try to fix it ?

#### What is your rclone version (output from `rclone version`)

```
rclone v1.60.0
- os/version: alpine 3.16.2 (64 bit)
- os/kernel: 5.4.0-122-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.19.2
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

The official rclone docker image

#### Which cloud storage system are you using? (e.g. Google Drive)

S3, but probably irrelevant here

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

[...]

PR #7057, opened 2023-06-14
rc: config/listremotes include from env vars

kapitainsky:
Fixes: 
#6540

Discussed:
https://forum.rclone.org/t/environment-variable-config-not-used-for-remote-control/39014


ncw (commented):
Code looks nice :-)

Can you write a test for it here

https://github.com/rclone/rclone/blob/3faa84b47c6e84af288c96d92c19c6b872b0f053/fs/config/rc_test.go#L70

The help for config/listremotes should probably say it includes ones defined by env var also.

I think you need to `go fmt` this file also.

kapitainsky:
Test added

kapitainsky:
Done and ready for review.

fs/config/rc_test.go
```diff
@@ -68,6 +69,7 @@ func TestRc(t *testing.T) {
 	})
 
 	t.Run("ListRemotes", func(t *testing.T) {
+		_ = os.Setenv("RCLONE_CONFIG_MY-LOCAL_TYPE", "local")
```
ncw, line 72: I would tidy this env var up at the end of the function otherwise it might interfere with other tests.

I'd also check the error from os.Setenv which is very easy to do in the test functions.

Something like this (untested)


```go
assert.NoError(t, os.Setenv("RCLONE_CONFIG_MY-LOCAL_TYPE", "local"))
defer func() {
    assert.NoError(t, os.Unsetenv("RCLONE_CONFIG_MY-LOCAL_TYPE"))
}()
```

ncw (commented):
This is looking very nearly perfect - thank you :-)

See inline for tidying up the env var.

fs/config/rc_test.go
kapitainsky: Yes this is good practice - sorry for not thinking about it. Done.

fs/config/rc.go
```diff
@@ -70,9 +73,18 @@ See the [listremotes](/commands/rclone_listremotes/) command for more informatio
 	})
 }
 
+var rcMatchEnv = regexp.MustCompile(`^RCLONE_CONFIG_(.*?)_TYPE=.*$`)
+
 // Return the a list of remotes in the config file
+// including any defined by environment variables.
 func rcListRemotes(ctx context.Context, in rc.Params) (out rc.Params, err error) {
 	remotes := LoadedData().GetSectionList()
```
ncw, line 81: I was looking to at your other PR #7058 and I realised we already have a function to do this which is called by `rclone listremotes`

https://github.com/rclone/rclone/blob/998d1d1727e5a76b0af2b61bc29a2bd8fa882b9a/cmd/listremotes/listremotes.go#L38

https://github.com/rclone/rclone/blob/998d1d1727e5a76b0af2b61bc29a2bd8fa882b9a/fs/config/config.go#L642-L644

So what I think we should do is call that function here and lose the looking in the environment because that function does it already.

Hopefully your tests won't need any changes

kapitainsky: It not only fixes env problem in rc but simplifies code indeed. I have made changes following your suggestion.

I do not think that tests need any changes.

ncw: Thank you

ncw (approved):
Perfect - thank you :-)

Will merge now.

PR #7058, opened 2023-06-14
docs: listremotes also includes remotes from env vars

kapitainsky:
"rclone listremotes" lists all the remotes in the config file and defined in environment variables. This is was clear from documentation.


ncw (approved):
Thank you :-)

PR #7060, opened 2023-06-15
fix(fs/accounting): bwlimit signal handler should always start

slai:

The SIGUSR2 signal handler for bandwidth limits currently only starts _if rclone is started at a time when a bandwidth limit applies_. This means that if rclone starts _outside_ such a time, i.e. when there's no bandwidth limits, _then_ enters a time when bandwidth limits do apply, it will not be possible to use SIGUSR2 to toggle it.

This PR fixes that by always starting the signal handler, but only toggling the limiter if there is a bandwidth limit configured.

It also removes the `tb.currLimitMu` mutex because it doesn't seem to serve a purpose that `tb.mu` doesn't already do, at least from what I can see. Happy to add it back in if needed though!

No.

#### Checklist

### DIFF
--- a/fs/accounting/token_bucket.go
+++ b/fs/accounting/token_bucket.go
@@ -105,11 +105,11 @@ func (tb *tokenBucket) StartTokenBucket(ctx context.Context) {
 	if tb.currLimit.Bandwidth.IsSet() {
 		tb.curr = newTokenBucket(tb.currLimit.Bandwidth)
 		fs.Infof(nil, "Starting bandwidth limiter at %v Byte/s", &tb.currLimit.Bandwidth)
-
-		// Start the SIGUSR2 signal handler to toggle bandwidth.
-		// This function does nothing in windows systems.
-		tb.startSignalHandler()
 	}
+
+	// Start the SIGUSR2 signal handler to toggle bandwidth.
+	// This function does nothing in windows systems.
+	tb.startSignalHandler()
 }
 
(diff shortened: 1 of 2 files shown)


fs/accounting/token_bucket.go
```diff
@@ -126,7 +125,6 @@ func (tb *tokenBucket) StartTokenTicker(ctx context.Context) {
 	go func() {
 		for range ticker.C {
 			limitNow := ci.BwLimit.LimitAt(time.Now())
-			tb.currLimitMu.Lock()
 
 			if tb.currLimit.Bandwidth != limitNow.Bandwidth {
 				tb.mu.Lock()
```
ncw, line 130: If we are using the tb.mu for protection of all the tb variables then we need to lock it before reading `tb.currLimit` so the tb.mu.Lock() needs to move to where the `tb.currLimitMu.Lock()` was. Same with the unlock.

ncw (commented):
I think this looks fine :-)

I think there is a small locking error which needs fixing (see inline) but your idea of removing the currLimitMu seems like a good one to me.

Can you split this into two commits (they can come in the same PR), one for the locking change and one for the signal handler change - thanks!

slai:
Thanks for the review!

You're right, there was a locking error which is now fixed. I've also split the change into two commits as requested.

ncw (approved):

I'll merge now.

Issue #7056, opened 2023-06-14
allow parallel/multithreaded/multipart transfers between remotes


jorjao81:
#### What is your current rclone version (output from `rclone version`)?
v1.63.0

#### What problem are you are trying to solve?
Specifically, transfers from S3 to S3 are much slower than they have to be. In our case, the buckets need different credentials, so remote copying is not an option (however, in our experience remote copy is actually *slower* than local copy, and it is also not done in parallel). This happens sequentially, resulting for us in speeds of about 70MiB/s. With multithreading we can download at speeds limited only by disk and network speed. Same thing for upload using multipart.
So, actually, download to local and re-uploading is faster than a direct copy, but still half the speed we should have; AND requires more fumbling with infrastructure as the disk needs to be big enough for holding the files (> TiB).

I would like to make the in-memory copy limited only by network (and maybe cpu/mem). Our main problem is S3, but I think the solution should be at least somewhat generic.

See related forum post: https://forum.rclone.org/t/s3-s3-large-file-copy/18574/4

#### How do you think rclone should be changed to solve that?

I would actually be happy to receive some opinions on how to go about it ;-)

My initial though would be to change (or create a new) the multithreaded downloading to make it more similiar to how multipart uploads work. Instead of simply selecting threads and then using `filesize/threads` to get ranges AND threads,  we would us `chunk-size` for ranges and `download-concurrency` for number of threads.

Then we could plug this into the multi-part uploading by selecting the same sizes.

But I'm fuzzy on the details, and on how to make it more generic.

Is this a feasible idea? Any other approaches that might be better than the one I suggested?

ncw:
Now that we've merged the first part of this - thank you @vitorog and @jorjao81 I've had a go at the next part.

This is making the multipart upload generic.

I'd really appreciate testing on this.

[v1.64.0-beta.7247.5bd9e6527.multithread](https://beta.rclone.org/branch/multithread/v1.64.0-beta.7247.5bd9e6527.multithread/) on branch [multithread](https://github.com/rclone/rclone/tree/multithread) (uploaded in 15-30 mins)

The main changes I've made are:

- pool the memory use - this is really important for not exploding memory use
- factor a multipart uploader to lib/multipart from backend/s3
- tidy and polish the ChunkWriter code
- don't do buffering when writing to local disk

This takes all the memory management out of the s3 backend which is what I wanted to achieve - that will also fix #7141

I'd appreciate it if you would test and/or review this branch.

[...]

PR #7061, opened 2023-06-15
parallel transfers between remotes

jorjao81:

Make transfers between remotes *fast*. 
At the moment, it's an ugly PoC, but on a m6in.8xlarge instance (with theoretically 50Gbps) using it allowed me to increase S3 to S3 transfers from ~70MiB/s to 5.2GiB/s.

At the moment, it replaces the multithreaded downloader with a "chunked" downloader, that works like chunked uploads to s3. I then hacked an S3 Multipart upload at the end of it (not using the existing S3 multipart code).

It is ugly, introduces aws sdk 2 dependency, breaks multithreaded downloads, surely breaks a ton of other stuff. But it's *fast*. 

I was thinking of concocting some sort of "ChuckedUploader" interface/Feature that could be added to providers to plug into it and make it generic, any thoughts?

And in general, what else should I keep in mind to get this mergeable? Separate feature while leaving multithreaded alone?

Thanks!

https://github.com/rclone/rclone/issues/7056

#### Checklist


ncw:
> At the moment, it's an ugly PoC, but on a m6in.8xlarge instance (with theoretically 50Gbps) using it allowed me to increase S3 to S3 transfers from ~70MiB/s to 5.2GiB/s.

Wow!

> At the moment, it replaces the multithreaded downloader with a "chunked" downloader, that works like chunked uploads to s3. I then hacked an S3 Multipart upload at the end of it (not using the existing S3 multipart code).

So this is adds a more limited interface to the S3 backend `OpenChunkWriter` rather than the too general `OpenWriterAt` and then rewrites the multi thread copy to use it instead.

This is a great idea! It also means that the multipart upload code which is essentially duplicated in the b2/s3/azureblob backends could be moved into operations.
 
> It is ugly, introduces aws sdk 2 dependency, breaks multithreaded downloads, surely breaks a ton of other stuff. But it's _fast_.

I don't think there is anything in this which needs the aws sdk 2 is there? Your new interface is essentially a factored version of the multipart upload code which we have already.

> I was thinking of concocting some sort of "ChuckedUploader" interface/Feature that could be added to providers to plug into it and make it generic, any thoughts?

Yes that would be great!

Note that we've discussed this interface in the context of resuming uploads also.

> And in general, what else should I keep in mind to get this mergeable? Separate feature while leaving multithreaded alone?

I would be inclined to add a `OpenChunkwriter` interface to the local backend and re-write the multithread uploader to use it instead.

Here is what you've got so far - I'll comment inline

```go
type OpenChunkWriter interface {
	// OpenChunkWriter returns a ChunkWriter
	//
	// Pass in the remote desired and the chunk size
	//
// ncw: it would make more sense for the backend to choose the chunk size probably as backends have flags for configuring the chunk size already. Some backends have odd limits on chunk sizes.
// ncw: note that if we don't know the size of an incoming file then we might want to pass fileSize as -1 here and it is up to the backend to deal with that or not. If it can't deal with it it should return some standardized error that we can check and take a backup path.
// ncw: (idea) if the backend would prefer you to use a non chunked copy the could potentially return the special error too. In that way we could remove the multipart upload code from all the backends and just use the new multithreaded copy.
//ncw: note that some backends (eg google drive) can do chunked uploads but they have to be in sequence. We might want to implement this...
//ncw: this will have to take the `src` parameters that Update takes so we can deal with stuff like metadata. This has the `size` available with `src.Size()` and the remote with `src.Remote()`
//ncw: it should also take OpenOptions - we could use these to hint at a chunk size if we wanted.

	// OpenChunkWriter(ctx context.Context, remote string, chunkSize int64, fileSize int64) (ChunkWriter, error)
	OpenChunkWriter(ctx context.Context, src ObjectInfo , options ...OpenOption) (chunkSize int64, ChunkWriter, error)

}

type ChunkWriter interface {
	// some backends require us to explicitly start a chunked upload, if it doesn't this can just be a NOP
// ncw: Shouldn't OpenChunkWriter do this and we can remove this method?
	StartChunkWrite() error // ncw: I'd just name this Start if we need it

	// the reader should correspond to the chunk number
// ncw: I'll just note that we almost always have the chunk in memory at this point so we can do retries. I guess if this is running from the multi-thread uploader then potentially we could re-open the stream at the right place if we need to do a retry.
// ncw: Not having to buffer the streams in memory would be nice, but the interfaces may not work out - the s3 SDK may need something different! So this might need to take a []byte rather than an `io.Reader`.
	WriteChunk(chunkNumber int, reader io.Reader) error

	// complete chunked writer
	CompleteChunkWrite() error  // ncw: This is Close() in go speak I think

	// abort chunk write
	AbortChunkWrite() error  // ncw: This should just be Abort()
}
```

So I think a rough plan with one commit for each might be

- Define new interfaces in `fs`
- Implement `OpenChunkWriter` and `ChunkWriter`  for `local` backend.
- Add tests of `OpenChunkWriter` and `ChunkWriter` to the integration tests `fstest/fstests/fstests.go` 
- Refactor multipart upload in `s3` backend to create `OpenChunkWriter` and `ChunkWriter` and Leave multipart upload in s3 backend for the moment just using the new interface.
- Rework multi-thread copy to use new interface
- Refactor multipart upload in `s3` to use version in operations

I'd like to implement a few backends to make sure the interface is general enough.

Great idea! I like your upload chunk interface a lot - we've been missing that. Let's discuss a bit more. I suspect we should make an issue with the results of our discussions and start from that.

jorjao81:
I had already created a ticket before I went ahead and started to code anyway :P https://github.com/rclone/rclone/issues/7056

As for the aws sdk: the reason I picked 2 was that it requires an `io.Reader`, whereas 1 requires a `ReadSeeker`. I think we usually won't have a `ReadSeeker`. I also don't really want to put the whole Chunk in memory (which would give me an easy ReadSeeker) as this can use quite a bit of memory for TB files with large chunks and high parallelism. I guess it would also make it a bit slower.

ncw:
I bet if you trace the SDK it asserts the io.Reader into an io.ReadSeeker or reads the chunk into memory. That's what the old SDK does. The protocol needs the md5sum or sha256 of the chunk in advance of sending it so will need to seek to calculate that. I guess it it possible they've implemented http trailers but that certainly didn't use to work.

ncw:
> I bet if you trace the SDK it asserts the io.Reader into an io.ReadSeeker or reads the chunk into memory. That's what the old SDK does. The protocol needs the md5sum or sha256 of the chunk in advance of sending it so will need to seek to calculate that. I guess it it possible they've implemented http trailers but that certainly didn't use to work.

I did exactly that and yes the SDK appears try to rewind the stream if
- it needs to calculate an sha256 for the chunk
- or on a retry

Eg

https://github.com/aws/aws-sdk-go-v2/blob/ea71314ea4cf35a4ad53570483bb2fce29a7573d/aws/signer/v4/middleware.go#L182-L196

I suspect that it will buffer a non-rewindable stream in memory but I didn't find that bit of code yet!

I did try to have a go with the code myself but I couldn't get the awsSDKv2 to auth. What does your rclone config look like to make it work?

jorjao81:
My config looks like this

      [s3]
      type = s3
      provider = AWS
      env_auth = true
      region = eu-west-1
      
      [local]
      type = local

ncw:
I'm going to close this once since we've landed #7154

However there is more work to do which will be discussed in #7056

Thank you :-)

Issue #6459, opened 2022-09-21
Doc issue: Setting up systemd (auto)mount units like in the docs creates a dependency cycle.


wilcooo:
Creating Systemd units, exactly like described in [this section](https://rclone.org/commands/rclone_mount/#rclone-as-unix-mount-helper) of the documentation, creates a dependency cycle.

The automount is set to be `After=network-online.target` (which is fair), but also has an implicit `Before=local-fs.target` set automatically by systemd.

From `man systemd.automount`:

> Automount units automatically gain an After= dependency on
           local-fs-pre.target, and a Before= dependency on local-fs.target.

This causes one element in the cycle to randomly fail. Sometimes, it's the automount itself (causing the onedrive folder to be empty). Other times, it's the NetworkManger (another unit in the created cycle). The latter one is quite annoying, as you can imagine!

(one of) the cycles: `mnt-onedrive.automount` -> `local-fs.target` -> `ldconfig.service` -> `sysinit.target` -> `NetworkManager-wait-online.service` -> `network-online.target` -> `mnt-onedrive.automount`

I don't know much about systemd, most of it I only just learned by investigating this issue. I guess `DefaultDependencies=no` should solve this, but someone with better knowledge might need to confirm this.

#### Technical details

Clean install of Fedora 36 workstation. rclone installed via dnf.

    rclone v1.57.0-DEV
    - os/version: fedora 36 (64 bit)
    - os/kernel: 5.19.9-200.fc36.x86_64 (x86_64)
    - os/type: linux
    - os/arch: amd64
    - go/version: go1.18beta1
    - go/linking: dynamic
    - go/tags: none
 
I've setup a onedrive business mount. When an unimportant unit of the cycle happens to be failing, all works fine.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
I'm not sure what the issue as I've used that service file for years now without any issue.

I'm not following what problem happens as I use:

```
Wants=network-online.target
After=network-online.target
```

In my network related services.

albertony:
Related: https://github.com/rclone/rclone/pull/5911

animosity22:
[...]

PR #7075, opened 2023-06-22
docs: remove "After" in systemd mount example again

URenko:
It used to be discussed in #6459 and fixed in #5911 and 7fbc928a19431f566b87c8c33d98fbb5099dad5f, but was overwritten by auto-generation in 01dbbff62e2470603d0c3abf4adf5c96e60cdbb5.

#### Checklist

### DIFF
--- a/cmd/mountlib/help.go
+++ b/cmd/mountlib/help.go
@@ -407,8 +407,6 @@ mount sftp1:subdir /mnt/data -t rclone -o vfs_cache_mode=writes,sftp_key_file=/p
 or create systemd mount units:
 |||
 # /etc/systemd/system/mnt-data.mount
-[Unit]
-After=network-online.target
 [Mount]
 Type=rclone
 What=sftp1:subdir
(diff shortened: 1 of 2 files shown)


ncw (approved):
Thanks for fixing this.

The system of auto generated files is maybe too complicated. We should probably have a linter which checks that auto generated files don't get edited.

I'll merge this now - thanks.

darthShadow:
Ideally, you just need to specify `_netdev`, so that systemd treats it as a remote fs rather than a local fs, in the options rather than removing the dependency on `network-online.target`.

Reference: https://www.freedesktop.org/software/systemd/man/systemd.mount.html#_netdev

```
Normally the file system type is used to determine if a mount is a "network mount", i.e. if it should only be started after the network is available. Using this option overrides this detection and specifies that the mount requires network.

Network mount units are ordered between remote-fs-pre.target and remote-fs.target, instead of local-fs-pre.target and local-fs.target. They also pull in network-online.target and are ordered after it and network.target.
```

ncw:
@darthShadow that must be mirroring the mount option `_netdev`. Do you want to send a PR to add that? Thank you

PR #7077, opened 2023-06-23
Update s3.md (documentation) - restoring single objects from glacier requires --include argument

danielkrajnik:

example in the [documentation](https://rclone.org/s3/#restore) doesn't seem to work:
`rclone backend restore s3:bucket/path/to/object [-o priority=PRIORITY] [-o lifetime=DAYS]`
should be:
`rclone backend restore s3:bucket/path/to --include object [-o priority=PRIORITY] [-o lifetime=DAYS]
`

https://forum.rclone.org/t/cant-restore-files-from-aws-glacier-deep-only-directories/39258/3

#### Checklist

### DIFF
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -3029,7 +3029,7 @@ to normal storage.
 
 Usage Examples:
 
-    rclone backend restore s3:bucket/path/to/object [-o priority=PRIORITY] [-o lifetime=DAYS]
+    rclone backend restore s3:bucket/path/to/ --include /object [-o priority=PRIORITY] [-o lifetime=DAYS]
     rclone backend restore s3:bucket/path/to/directory [-o priority=PRIORITY] [-o lifetime=DAYS]
     rclone backend restore s3:bucket [-o priority=PRIORITY] [-o lifetime=DAYS]
 


ncw:
I think this behaviour is a bug.

I've posted a binary for you to try in [the forum](https://forum.rclone.org/t/cant-restore-files-from-aws-glacier-deep-only-directories/39258/9)

danielkrajnik:
thank you, I've replied back on the forum as well

kapitainsky:
I would suggest to make it more clear that this is the way to restore file and use file not object:

`rclone backend restore s3:bucket/path/to --include file [-o priority=PRIORITY] [-o lifetime=DAYS]`

docs/content/s3.md
```diff
@@ -3029,7 +3029,7 @@ to normal storage.
 
 Usage Examples:
 
-    rclone backend restore s3:bucket/path/to/object [-o priority=PRIORITY] [-o lifetime=DAYS]
+    rclone backend restore s3:bucket/path/to --include object [-o priority=PRIORITY] [-o lifetime=DAYS]
```
ncw, line 3032: I think this should be `--include /object` otherwise you'll restore any objects under the file heirachy called `object`.

I'd also leave the `/` on the end of `s3:bucket/path/to/` which signals to rclone that this is definitely a directory.

ncw (commented):
Scratch the idea of this being a bug - it would break too much stuff if I changed it!

I'd like to merge this but I think it needs a small tweak first.

Thank you

docs/content/s3.md
danielkrajnik: I think that's done now, sorry, new to using github, but this should be updated now?

kapitainsky: > I think that's done now, sorry, I'm new to using github, but there should be a new "tweaked path slashes" commit with these changes now?

Yes there are two - now you should squash them into one - edit commit message - and force push. It will make it one again.

kapitainsky: git cmd can be challenging - myself I use GitKraken (free for public repos) and recommend it to everybody:)

danielkrajnik: yeah, still trying to crack it. Almost there...

kapitainsky: PR should contain one commit only - so squashing and forced pushing is daily bread for PR authors:)

danielkrajnik: I've squashed them, but when trying to `git push`  them back it kept rejecting them `(non-fast-forward)`
I hope that it won't cause any issue that I "forced that push": `git push -f`
It seems to be okay now.

kapitainsky: yeap - they have to forced push - and all is now how it should:)

ncw: Its easy for me to squash commits into one when merging - there is a nice button on GitHub which does that. So two commits would have been fine.

However one is even better :-)

ncw (approved):

PR #7093, opened 2023-06-28
backend: Add Proton Drive backend

henrybear327:

Add support for a new backend [Proton Drive](https://proton.me/drive)

 #6072 

#### Checklist

#### Tests performed (passed all tests for now)
- At `root`: `make && test_all -backends protondrive`
- At `backend/protondrive`: `go test -race -v ./...`
- At `root`: `make && rclone purge -v TestProtonDrive:rclone-info; rclone test info --all -vv --write-json remote.json TestProtonDrive:rclone-info && go run cmd/test/info/internal/build_csv/main.go -o remote.csv remote.json`
- Linter
  - At `backend/protondrive`, run `golangci-lint run`

#### Notes
- Willing to keep on maintaining and improving this backend
- Wrote the [Proton-API-Bridge](https://github.com/henrybear327/Proton-API-Bridge) library to free the rclone codebase from dealing with the encryption and decryption details
- All the rclone unit and integration tests are run and passed (as specified above)

#### Known issue
- After adding the `Proton-API-Bridge`, as we need this for handling the specifics of the encryption, it introduces a lot of new entries in the `go.mod` and `go.sum`
- AppString, which is currently handled by the bridging library, is hardcoded
- Access token isn’t cached, so if the account has 2FA enabled, the config needs to be changed every time as the 2FA code is read from the config file
- Not working
    - `rclone purge -v TestProtonDrive:`: not implemented as we can't delete root folder and recreate one 
    - ~~`rclone ls TestProtonDrive:`: error msg `Failed to ls with 2 errors: last error was: gopenpgp: error in reading message: openpgp: incorrect key`~~

#### The improvement work for the future 
I foresee [the work](https://github.com/henrybear327/Proton-API-Bridge#v2) to be done mostly from the bridging library, as there won't be big changes in the rclone backend code, and would be mostly centered around performance, edge cases, and some error-handling improvements in the bridging library, e.g.:
- Caching the `proton.Link` object, should be able to reduce quite some API calls when doing List()
- Downloading and uploading large files: currently, the entire file will be loaded into the memory and then the encryption/decryption operation for upload/download will then be performed
- Optimize the MimeType detection
- Check the client code and completes better signature verification and edge case handling, e.g. failed upload will leave the draft revision and blocks behind, validation/limitations seen in the open source clients
- Thumbnail support


henrybear327:
Screenshot 2023-06-29 at 12 37 50 AM

[remote.csv](https://github.com/rclone/rclone/files/11900444/remote.csv)

henrybear327:
Will fix 2 things in the meantime in the Bridging library 
- (Edge case) When the upload got canceled midway, a draft will be left behind, causing restarting the uploading operation to fail
- Set the flag for Size() to report the original size to default to true, instead of false, which is reporting the size after encryption, returned by the server. (Currently, it defaults to false and this will cause the upload verification to fail. It passes all the unit and integration tests because I turned the flag on during testing, but I didn't realize that during normal upload operation, the original content size will be checked, too)

ncw:
This looks like excellent work with all the integration tests passing too - well done :-)

I'm just in the middle of making the 1.63 release so I can't review this right now but I'll get on to it next week :crossed_fingers: 

> #### Notes
> * Willing to keep on maintaining and improving this backend

Great. Given that this is a reverse engineered protocol I'd like to recruit you as the maintainer for it once we've merged it.

> #### Known issue
> * After adding the `Proton-API-Bridge`, as we need this for handling the specifics of the encryption, it introduces a lot of new entries in the `go.mod` and `go.sum`

Not a problem unless it bloats the rclone binary excessively.

> * AppString, which is currently handled by the bridging library, is hardcoded

Is this the equivalent of the `client_id` in oauth speak? If so having a hardcoded one is fine, but we should probably let the user change it too.

> * Access token isn’t cached, so if the account has 2FA enabled, the config needs to be changed every time as the 2FA code is read from the config file

This can be cached in the config file fairly easily which is what the oauth backends do.

> * Not working
>   
>   * `rclone purge -v TestProtonDrive:`: not implemented as we can't delete root folder and recreate one

Not many of the providers can delete the root folder - that is fine. Some of them just return an error if you try to purge or rmdir the root.

>   * `rclone ls TestProtonDrive:`: error msg `Failed to ls with 2 errors: last error was: gopenpgp: error in reading message: openpgp: incorrect key`

I guess this is one for you :-)
 
> #### The improvement work for the future
> I foresee [the work](https://github.com/henrybear327/Proton-API-Bridge#v2) to be done mostly from the bridging library, as there won't be big changes in the rclone backend code, and would be mostly centered around performance, edge cases, and some error-handling improvements in the bridging library,

I update all of the deps once or twice per release cycle but if you wanted it quicker than this send a PR.

All in all - looking very good!

Please send questions if you have any.

I haven't looked at the code yet - don't want to get distracted from finalising the release (again!).

ncw:
> > > * Access token isn’t cached, so if the account has 2FA enabled, the config needs to be changed every time as the 2FA code is read from the config file
> > 
> > 
> > This can be cached in the config file fairly easily which is what the oauth backends do.
> 
> Sorry, maybe I missed this when tracing the codebase. Would you my giving me some pointers? Thank you! ^_^

When your `NewFs` is called you are given a parameter called `m configmap.Mapper`. Normally we just unpack this in the the `Options` struct and that is the end of it.

However you can write config back to `m` with `m.Set(key, value string)`. This  writes the value back to the config file so that the next time `NewFs` is called you get the new value. If you grep the backends for `m.Set` you'll find quite a few examples.

henrybear327:
Small update: optimization on uploading (progress display fix and mime type detection) and caching/refreshing the access credentials are underway. Should be able to push the final reviewable version by mid-next week (quite some underlying bridging library changes)!

One question: should the access token, refresh token, etc. somehow be encrypted? Or we just let the mapper do its job and let the user takes care of securing their tokens :) 

Thank you!

> > > > * Access token isn’t cached, so if the account has 2FA enabled, the config needs to be changed every time as the 2FA code is read from the config file
> > > 
> > > 
> > > This can be cached in the config file fairly easily which is what the oauth backends do.
> > 
> > 
> > Sorry, maybe I missed this when tracing the codebase. Would you my giving me some pointers? Thank you! ^_^
> 
> When your `NewFs` is called you are given a parameter called `m configmap.Mapper`. Normally we just unpack this in the the `Options` struct and that is the end of it.
> 
> However you can write config back to `m` with `m.Set(key, value string)`. This writes the value back to the config file so that the next time `NewFs` is called you get the new value. If you grep the backends for `m.Set` you'll find quite a few examples.

ncw:
> Small update: optimization on uploading (progress display fix and mime type detection) and caching/refreshing the access credentials are underway. Should be able to push the final reviewable version by mid-next week (quite some underlying bridging library changes)!

Great :-)
 
> One question: should the access token, refresh token, etc. somehow be encrypted? Or we just let the mapper do its job and let the user takes care of securing their tokens :)

We don't normally encrypt them. Users can use config file encryption if they want.

We do obfuscate passwords to prevent "eyedropping" as passwords are often easy to remember unlike tokens.

henrybear327:
Hello @ncw, 

I will squash the commits and remove logging in the next push of the PR, but the current code is now ready to be reviewed as I have tested and passed all the tests listed here:
```
At root: make && test_all -backends protondrive
At backend/protondrive: go test -race -v ./...
At root: make && rclone purge -v TestProtonDrive:rclone-info; rclone test info --all -vv --write-json remote.json TestProtonDrive:rclone-info && go run cmd/test/info/internal/build_csv/main.go -o remote.csv remote.json
```

The progress display is now fixed and credentials are now cached!

Thank you for the time and sorry for the wait! Looking forward to the comments and I will address them ASAP.

Henry

henrybear327:
Hello @ncw, 

The current code is now ready to be reviewed as I have tested and passed all the tests listed here:
```
At root: make && test_all -backends protondrive
At backend/protondrive: go test -race -v ./...
At root: make && rclone purge -v TestProtonDrive:rclone-info; rclone test info --all -vv --write-json remote.json TestProtonDrive:rclone-info && go run cmd/test/info/internal/build_csv/main.go -o remote.csv remote.json
```

I will squash the commits after the review if this ok for you :) 

The progress display is now fixed and credentials are now cached!

Thank you for the time and sorry for the wait! Looking forward to the comments and I will address them ASAP.

Henry 

Sidenotes:
- Upload will now be reading in data on-the-fly so memory usage is fine now! But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory. 
- The main future improvements are mostly on the performance regarding file/folder object caching on the bridging library side. For now, we can use `--transfers 10` to improve upload performance.

ncw:
That is all looking great - well done :-)

> But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory.

This is unfortunate. Is there no way to stream the file or part of the file?

I see the Ranged requests download the entire file too. This will make `rclone mount` with `--vfs-cache` very inefficient as it will download the whole file multiple times :-(

If this can't be fixed then we should at least write the file to disk if it is above a certain size. We'll need some big warnings in the docs too.

The fs should set the `NoMultiThreading` feature flag also to disable multi-thread downloads otherwise the file will just get downloaded multiple times.

So I think there are two issues here
- excessive RAM use when downloading large files
- ranged requests downloading the whole file

Both of which are quite important for a smooth backend!

henrybear327:
> That is all looking great - well done :-)
> 
> > But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory.
> 
> This is unfortunate. Is there no way to stream the file or part of the file?

We can definitely download parts of the file, but it will rely on making the following assumption:
- All blocks except the last block, will have the same size. So we can infer the block size from the first block, and then we can compute the offset, and then the ranged request will work!

I am not certain if there exists a safer way to do this. The reason is the following: the raw file is split into equally-sized (currently 4MB, but it might change in the future) blocks, except the last one which might be smaller than 4MB. Each block is encrypted separately, and size and sha1 after the encryption is added to the metadata of the block. 

So technically, we can implement this feature, but with the aforementioned catch!

What do you think? (I Hope I explained it clearly)
 
> 
> I see the Ranged requests download the entire file too. This will make `rclone mount` with `--vfs-cache` very inefficient as it will download the whole file multiple times :-(
> 
> If this can't be fixed then we should at least write the file to disk if it is above a certain size. We'll need some big warnings in the docs too.

This is also another option (which I think I would prefer), as we don't need to deal with offsets by inferring and making assumptions. Is there a code pointer for this so I can just turn my code into writing into the file buffer as the decryption goes? Blocks can be downloaded in parallel, just that we piece them together sequentially :) 

> 
> The fs should set the `NoMultiThreading` feature flag also to disable multi-thread downloads otherwise the file will just get downloaded multiple times.
> 
> So I think there are two issues here
> 
> * excessive RAM use when downloading large files
> * ranged requests downloading the whole file
> 
> Both of which are quite important for a smooth backend!

henrybear327:
> That is all looking great - well done :-)
> 
> > But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory.
> 
> This is unfortunate. Is there no way to stream the file or part of the file?
> 
> I see the Ranged requests download the entire file too. This will make `rclone mount` with `--vfs-cache` very inefficient as it will download the whole file multiple times :-(
> 
> If this can't be fixed then we should at least write the file to disk if it is above a certain size. We'll need some big warnings in the docs too.
> 
> The fs should set the `NoMultiThreading` feature flag also to disable multi-thread downloads otherwise the file will just get downloaded multiple times.
> 
> So I think there are two issues here
> 
> * excessive RAM use when downloading large files
> * ranged requests downloading the whole file
> 
> Both of which are quite important for a smooth backend!

With this quick patch, I set `NoMultiThreading` to true in the feature flag, and implemented a download mechanism that is fetching blocks batch-by-batch, and passes data using an `io.ReadCloser`, so it's now memory-efficient!

Chunker support will not be available with `NoMultiThreading` set to true I believe.

Integration tests are currently in progress!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+		default:
+			if option.Mandatory() {
+				fs.Logf(o, "Unsupported mandatory option: %v", option)
+			}
+		}
+	}
+
+	retReader := io.NopCloser(reader) // the NewLimitedReadCloser will deal with the limit
+
+	// deal with offset
```
henrybear327, line 855: I believe that with the `noMultiThreading`, the file will only be downloaded once. But I am not certain what are the use cases for `Seek`. 

Due to the way the file is encrypted, as explained and discussed [here](https://github.com/rclone/rclone/pull/7093#issuecomment-1636737655), I don't have a better way to support seeking in this backend. 

But never the less, the code now is very efficient, as the download is done in a way that
- we decrypt and buffer data of 8 blocks 
- the reader can perform read
- when the buffer runs out of data, we decrypt and buffer data up to next 8 blocks 
- repeat until EOF 

For `Seek`, it's just reading but discarding the result, it's a bit wasteful, but at least it's memory efficient and the file is downloaded once only!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+	if _, err := io.CopyN(io.Discard, retReader, offset); err != nil {
```
henrybear327, line 856: I believe that with the `noMultiThreading`, the file will only be downloaded once. But I am not certain what are the use cases for `Seek`. 

Due to the way the file is encrypted, as explained and discussed [here](https://github.com/rclone/rclone/pull/7093#issuecomment-1636737655), I don't have a better way to support seeking in this backend. 

But never the less, the code now is very efficient, as the download is done in a way that
- we decrypt and buffer data of 8 blocks 
- the reader can perform read
- when the buffer runs out of data, we decrypt and buffer data up to next 8 blocks 
- repeat until EOF 

For `Seek`, it's just reading but discarding the result, it's a bit wasteful, but at least it's memory efficient and the file is downloaded once only!

ncw:
> May I ask if there are other concerns, besides the ones mentioned below :)

No I don't think so!

> > > But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory.
> > 
> > This is unfortunate. Is there no way to stream the file or part of the file?
> 
> We can definitely download parts of the file, but it will rely on making the following assumption:
> 
> * All blocks except the last block, will have the same size. So we can infer the block size from the first block, and then we can compute the offset to which blocks we should be downloading, and then the ranged request will work!
> 
> I am not certain if there exists a safer way to do this. The reason is the following: the raw file is split into equally-sized (currently 4MB, but it might change in the future, say to 8MB, 16MB, etc.) blocks, except the last one which might be smaller than 4MB. Each block is encrypted separately, and size and sha1 after the encryption is added to the metadata of the block, but not the original block size and sha1.
> 
> So technically, we can implement this feature, but with the aforementioned catch!

This is exactly the way the `crypt` backend works. The `crypt` backend uses smaller blocks of 64k but otherwise it is very similar. The metadata is added to the block itself so the block size is 64k+a bit.

> What do you think? (I Hope I explained it clearly)

I think that will work very well. The major assumption you are making is that all the blocks are the same size. I think if you write your code to return an error if it ever detects otherwise it will be fine.

> This is also another option (which I think I would prefer), as we don't need to deal with offsets by inferring and making assumptions. Blocks can be downloaded in parallel, just that we piece them together sequentially :) And then all the integrity checks can be performed normally.

This still downloads the file multiple times on a seek.

> I think I got an idea to return the downloaded file as a reader, so we can operate with small memory footprint!

That is great :-) A reader is the right interface. That ticks off one of the problems.

> With this quick patch, I set `NoMultiThreading` to true in the feature flag, and implemented a download mechanism that is fetching blocks batch-by-batch, and passes data using an `io.ReadCloser`, so it's now memory-efficient!

:+1:
 
> Is there a way to disable `SeekOption`? Or do we have to support this in any case :)
> I believe that with the `noMultiThreading`, the file will only be downloaded once. But I am not certain what are the use cases for `Seek`.

Seek is used in quite a few places.

- When using anything which uses the VFS (eg `rclone mount`, `rclone serve`) backends are expected to be able to seek. If they can't seek then they stand the risk of downloading the file multiple times. For example a typical video player when playing a fille will open the file, read some bytes, seek to the end, read a few more bytes before starting to play anything. With your backend as it is at the moment, this will end up downloading the file completely for the seek to the end and then throwing all that data away and downloading it again to play it.
- The vfs cache also reads files in chunks normally using ranged requests even when the user isn't seeking. This is to signal to the backend that it doesn't have to queue the whole file, just a part of it. Some backends reduce your download quota by the size of the request you make even if you don't actually download those bytes, so it is best not to ask for more data than you want.
- The other really important part is on retries. If there is a temporary networking glitch on files rclone is reading then it will seek back to where it got to and retry.
- There are some more minor uses of seeking (eg the --head/--tail flags in `rclone cat`).

> Due to the way the file is encrypted, as explained and discussed [here](https://github.com/rclone/rclone/pull/7093#issuecomment-1636737655), I don't have a better way to support seeking in this backend.
> 
> But never the less, the code now is very efficient, as the download is done in a way that
> 
> * we decrypt and buffer data of 8 blocks
> * the reader can perform read
> * when the buffer runs out of data, we decrypt and buffer data up to next 8 blocks
> * repeat until EOF

8 blocks = 64MB may be more than you need. Rclone has a read ahead buffer so it will be reading data ahead of where the consumer wants it, so you don't have to implement this in the backend. I'd probably just do one block as they are quite big at 8MB - if the user has to wait for 64 MB of data to be downloaded then that is quite a lot of latency.

> For `Seek`, it's just reading but discarding the result, it's a bit wasteful, but at least it's memory efficient and the file is downloaded once only!

Hopefully we can fix this too :-)

henrybear327:
> > May I ask if there are other concerns, besides the ones mentioned below :)
> 
> No I don't think so!

Good to know! Thank you for the reviews!

> 
> > > > But downloading will be a problem for big files still as the entire file will be downloaded and decrypted in memory.
> > > 
> > > 
> > > This is unfortunate. Is there no way to stream the file or part of the file?
> > 
> > 
> > We can definitely download parts of the file, but it will rely on making the following assumption:
> > 
> > * All blocks except the last block, will have the same size. So we can infer the block size from the first block, and then we can compute the offset to which blocks we should be downloading, and then the ranged request will work!
> > 
> > I am not certain if there exists a safer way to do this. The reason is the following: the raw file is split into equally-sized (currently 4MB, but it might change in the future, say to 8MB, 16MB, etc.) blocks, except the last one which might be smaller than 4MB. Each block is encrypted separately, and size and sha1 after the encryption is added to the metadata of the block, but not the original block size and sha1.
> > So technically, we can implement this feature, but with the aforementioned catch!
> 
> This is exactly the way the `crypt` backend works. The `crypt` backend uses smaller blocks of 64k but otherwise it is very similar. The metadata is added to the block itself so the block size is 64k+a bit.
> 
> > What do you think? (I Hope I explained it clearly)
> 
> I think that will work very well. The major assumption you are making is that all the blocks are the same size. I think if you write your code to return an error if it ever detects otherwise it will be fine.

OK! I will proceed with this approach and update a PR!

> 
> > This is also another option (which I think I would prefer), as we don't need to deal with offsets by inferring and making assumptions. Blocks can be downloaded in parallel, just that we piece them together sequentially :) And then all the integrity checks can be performed normally.
> 
> This still downloads the file multiple times on a seek.
> 
> > I think I got an idea to return the downloaded file as a reader, so we can operate with small memory footprint!
> 
> That is great :-) A reader is the right interface. That ticks off one of the problems.

:) 

> 
> > With this quick patch, I set `NoMultiThreading` to true in the feature flag, and implemented a download mechanism that is fetching blocks batch-by-batch, and passes data using an `io.ReadCloser`, so it's now memory-efficient!
> 
> 👍
> 
> > Is there a way to disable `SeekOption`? Or do we have to support this in any case :)
> > I believe that with the `noMultiThreading`, the file will only be downloaded once. But I am not certain what are the use cases for `Seek`.
> 
> Seek is used in quite a few places.

Thanks for this awesome reply. Making a lot of sense now!!!
> 
> * When using anything which uses the VFS (eg `rclone mount`, `rclone serve`) backends are expected to be able to seek. If they can't seek then they stand the risk of downloading the file multiple times. For example a typical video player when playing a fille will open the file, read some bytes, seek to the end, read a few more bytes before starting to play anything. With your backend as it is at the moment, this will end up downloading the file completely for the seek to the end and then throwing all that data away and downloading it again to play it.
> * The vfs cache also reads files in chunks normally using ranged requests even when the user isn't seeking. This is to signal to the backend that it doesn't have to queue the whole file, just a part of it. Some backends reduce your download quota by the size of the request you make even if you don't actually download those bytes, so it is best not to ask for more data than you want.
> * The other really important part is on retries. If there is a temporary networking glitch on files rclone is reading then it will seek back to where it got to and retry.
> * There are some more minor uses of seeking (eg the --head/--tail flags in `rclone cat`).
> 
> > Due to the way the file is encrypted, as explained and discussed [here](https://github.com/rclone/rclone/pull/7093#issuecomment-1636737655), I don't have a better way to support seeking in this backend.
> > But never the less, the code now is very efficient, as the download is done in a way that
> > 
> > * we decrypt and buffer data of 8 blocks
> > * the reader can perform read
> > * when the buffer runs out of data, we decrypt and buffer data up to next 8 blocks
> > * repeat until EOF
> 
> 8 blocks = 64MB may be more than you need. Rclone has a read ahead buffer so it will be reading data ahead of where the consumer wants it, so you don't have to implement this in the backend. I'd probably just do one block as they are quite big at 8MB - if the user has to wait for 64 MB of data to be downloaded then that is quite a lot of latency.
> 

Oh ok! Good to know. So I will quickly change the code into reading one block at a time, and let rclone does the caching, offloading this task from the bridging library :) 

> > For `Seek`, it's just reading but discarding the result, it's a bit wasteful, but at least it's memory efficient and the file is downloaded once only!
> 
> Hopefully we can fix this too :-)

To sum up, I will try to put up the last modification with the following changes:
- Implement the ability to seek, with the assumption that we currently have!
- For downloading, we fetch one block at a time, and let rclone does the heavy lifting regarding buffering the data! (glad that I coded this with a config file, so it's a one-line change 👍 )

Please let me know if there is anything else on your mind! Have a nice weekend and thank you again for the thorough reply! See you on the next PR in a bit I hope!

ncw:
> To sum up, I will try to put up the last modification with the following changes:
> 
> * Implement the ability to seek, with the assumption that we currently have!
> * For downloading, we fetch one block at a time, and let rclone does the heavy lifting regarding buffering the data! (glad that I coded this with a config file, so it's a one-line change +1 )
> 
> Please let me know if there is anything else on your mind! Have a nice weekend and thank you again for the thorough reply! See you on the next PR in a bit I hope!

That sounds great - thank you :-)

I will go through the code in detail now and see if anything pops up! The main thing for backends though is that the integration tests pass :-)

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+
+// Put in to the remote path with the modTime given of the given size
+//
+// May create the object even if it returns an error - if so
+// will return the object and the error, otherwise will return
+// nil and the error
+//
+// May create duplicates or return errors if src already
+// exists.
+func (f *Fs) PutUnchecked(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) (fs.Object, error) {
```
ncw, line 674: If proton can't have duplicates you don't have to implement this. 

I'm not sure whether it can or not or maybe only sometimes!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+func (f *Fs) Disconnect(ctx context.Context) error {
+	return f.pacer.Call(func() (bool, error) {
+		err := f.protonDrive.Logout(ctx)
+		return shouldRetry(ctx, err)
+	})
+}
+
+// Move src to this remote using server-side move operations.
+//
+// # This is stored with the remote path given
```
ncw, line 954: You need to put a full stop on the end of the sentence to stop `go fmt` insterting these strange `#` symbols!

docs/content/protondrive.md
```diff
@@ -0,0 +1,107 @@
...
+---
+
+# {{< icon "fa fa-atom-simple" >}} Proton Drive
+
+[Proton Drive](https://proton.me/drive) is an end-to-end encrypted Swiss vault for your files that protects your data.
+
+This is a rclone backend for Proton Drive which supports the file transfer
+features of Proton Drive using the same client-side encryption.
+
+Due to the fact that Proton Drive doesn't publish its API documentation, this backend is implemented with the best effort by reading the open-sourced client source code and observing the Proton Drive traffic on the browser.
```
ncw, line 14: We normally hard wrap the docs to 80ish chars too.

docs/content/protondrive.md
```diff
@@ -0,0 +1,107 @@
...
+
+    rclone copy /home/source remote:backup
+
+### Modified time and hashes
+
+Proton Drive does not support modification times or hashes yet.
+
+### Restricted filename characters
+
+Invalid UTF-8 bytes will be [replaced](/overview/#invalid-utf8).
```
ncw, line 93: LeftSpace and RightSpace are missing from here I think.

docs/content/protondrive.md
```diff
@@ -0,0 +1,107 @@
...
+
+    rclone ls remote:
+
+To copy a local directory to an Proton Drive directory called backup
+
+    rclone copy /home/source remote:backup
+
+### Modified time and hashes
+
+Proton Drive does not support modification times or hashes yet.
```
ncw, line 89: Proton drive has hashes on each block doesn't it? If you could read those hashes quickly (as in read all the hashes at once) we could make a hashing scheme like the dropbox hashing scheme. The Dropbox hashing scheme is a hash of all the hashes on each 4 MB block.

This I'd view as a future enhancement!

https://github.com/rclone/rclone/blob/master/backend/dropbox/dbhash/dbhash.go

ncw, line 89: Proton drive has hashes on each block doesn't it? If you could read those hashes quickly (as in read all the hashes at once) we could make a hashing scheme like the dropbox hashing scheme. The Dropbox hashing scheme is a hash of all the hashes on each 4 MB block.

This I'd view as a future enhancement!

https://github.com/rclone/rclone/blob/master/backend/dropbox/dbhash/dbhash.go

ncw (commented):
I have been through all the code carefully now - it looks great :-)

I put some comments in but these are mostly about rclone naming conventions and doc formatting!

Thank you

ncw:
Sorry for duplicates above - I'm on a train with spotty WiFi.

henrybear327:
> Sorry for duplicates above - I'm on a train with spotty WiFi.

Thank you for the quick and thorough response! I will fix those now!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Help:     config.ConfigEncodingHelp,
+			Advanced: true,
+			// Encode invalid UTF-8 bytes as json doesn't handle them properly.
+			Default: (encoder.Base |
+				encoder.EncodeInvalidUtf8 |
+				encoder.EncodeLeftSpace |
+				encoder.EncodeRightSpace),
+		}, {
+			Name:     "returnoriginalfilesize",
+			Help:     "The size of the encrypted file will be different from (bigger than) the original file size. Unless there is a reason to return the size after encryption, otherwise, set this option to true, or features such as Open() which will need to be supplied with original content size, will fail to operate properly",
```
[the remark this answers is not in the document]
henrybear327: Re-written all the options!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Name:     config.ConfigEncoding,
+			Help:     config.ConfigEncodingHelp,
+			Advanced: true,
+			// Encode invalid UTF-8 bytes as json doesn't handle them properly.
+			Default: (encoder.Base |
+				encoder.EncodeInvalidUtf8 |
+				encoder.EncodeLeftSpace |
+				encoder.EncodeRightSpace),
+		}, {
+			Name:     "returnoriginalfilesize",
```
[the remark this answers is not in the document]
henrybear327: Done as requested. Same done for `replace_existing_draft` and `enable_caching`

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Help:     "2FA code (if the account requires one)",
+			Required: false,
+		}, {
+			Name:     config.ConfigEncoding,
+			Help:     config.ConfigEncodingHelp,
+			Advanced: true,
+			// Encode invalid UTF-8 bytes as json doesn't handle them properly.
+			Default: (encoder.Base |
+				encoder.EncodeInvalidUtf8 |
+				encoder.EncodeLeftSpace |
```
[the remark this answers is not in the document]
henrybear327: [code reference](https://github.com/ProtonMail/WebClients/blob/b4eba99d241af4fdae06ff7138bd651a40ef5d3c/applications/drive/src/app/store/_links/validation.ts#L51)

I read the web client source and saw their validation, thus, adding these rules. 

Is this not the right place to perform encoding / validation? :)

Thank you!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+				encoder.EncodeRightSpace),
+		}, {
+			Name:     "returnoriginalfilesize",
+			Help:     "The size of the encrypted file will be different from (bigger than) the original file size. Unless there is a reason to return the size after encryption, otherwise, set this option to true, or features such as Open() which will need to be supplied with original content size, will fail to operate properly",
+			Advanced: true,
+			Default:  true,
+		}, {
+			Name:     "appversion",
```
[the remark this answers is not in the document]
henrybear327: Done as requested

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Help:     "The app version string that will be sent with every API request. Looks like using [OS]-drive (e.g. windows-drive) prefix will spare us from getting aggressive human verification blocking",
```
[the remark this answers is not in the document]
henrybear327: Done as requested

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Advanced: true,
+			Default:  getAPIOS() + "-drive@5.0.14.2",
+		}, {
+			Name:     "useragent",
+			Help:     "The user agent string that will be sent with every API request",
+			Advanced: true,
+			Default:  "",
+		}, {
+			Name: "replaceexistingdraft",
```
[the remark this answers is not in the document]
henrybear327: Done as requested!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Help: `When an upload is cancelled or failed before completion, a draft will be created and the subsequent upload of the same file to the same location will be reported as a conflict.
+				If the option is set to true, the draft will be replaced and then the upload operation will restart. If there are other clients also uploading at the same file location at the same time, the behavior is currently unknown. Need to set to true for integration tests.
+				If the option is set to false, an error "a draft exist - usually this means a file is being uploaded at another client, or, there was a failed upload attempt" will be returned, and no upload will happen.
+			`,
+			Advanced: true,
+			Default:  false,
+		}, {
+			Name:     "enableCaching",
```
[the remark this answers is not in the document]
henrybear327: Done as requested!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+	link *proton.Link // link data on proton server
+}
+
+// shouldRetry returns a boolean as to whether this err deserves to be
+// retried.  It returns the err as a convenience
+func shouldRetry(ctx context.Context, err error) (bool, error) {
+	if fserrors.ContextError(ctx, &err) {
+		return false, err
+	}
+	// Let the mega library handle the low level retries
```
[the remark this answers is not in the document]
henrybear327: Sorry that I was reading and borrowing code from Mega, Box, and Dropbox....
Fixed!

backend/protondrive/protondrive.go
henrybear327: Ahh this solves my long-term question of what's going on with the `go fmt`! Thanks for pointing this out!
Fixed as requested.

docs/content/protondrive.md
henrybear327: Updated as requested!

docs/content/protondrive.md
henrybear327: Proton drive has hashes on each block *after* encryption only. 
It also has a manifest hash which is all the block's hash values concatenated together in order.

I agree that modtime and hashes are for future improvements! Let's rollout the basic features stably first! :)

henrybear327: Discarded as it's a duplicate.

backend/protondrive/protondrive.go
henrybear327: Proton's backend doesn't allow duplicated file names. So I will have to remove this function entirely I believe, right?

Sorry for missing the case, as I was following other backends and didn't realize this is specific to this "duplicated filename case"!

backend/protondrive/protondrive.go
henrybear327: Explained and resolved in other threads.

docs/content/protondrive.md
henrybear327: Got it! Fixed as requested.

henrybear327:
> To sum up, I will try to put up the last modification with the following changes:
> 
> * Implement the ability to seek, with the assumption that we currently have!
> * For downloading, we fetch one block at a time, and let rclone does the heavy lifting regarding buffering the data! (glad that I coded this with a config file, so it's a one-line change 👍 )

With the ability to obtain the block size array and sha1 original file content from`xattr`, we can now perform SHA1 hashing and seeking normally, provided that `xattr` is present. 

Questions 
- If `xattr` is missing, should I fallback to bad performance seek?

backend/protondrive/protondrive.go
henrybear327: Removed as requested!

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1040 @@
...
+			Help:     "The size of the encrypted file will be different from (bigger than) the original file size. Unless there is a reason to return the size after encryption, otherwise, set this option to true, or features such as Open() which will need to be supplied with original content size, will fail to operate properly",
+			Advanced: true,
+			Default:  true,
+		}, {
+			Name:     "appversion",
+			Help:     "The app version string that will be sent with every API request. Looks like using [OS]-drive (e.g. windows-drive) prefix will spare us from getting aggressive human verification blocking",
+			Advanced: true,
+			Default:  getAPIOS() + "-drive@5.0.14.2",
+		}, {
+			Name:     "useragent",
```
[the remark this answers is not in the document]
henrybear327: Removed the option and used this as the default now!

ncw:
> With the ability to obtain the block size array and sha1 original file content from`xattr`, we can now perform SHA1 hashing and seeking normally, provided that `xattr` is present.

Great :-)

> Questions
> 
> * If `xattr` is missing, should I fallback to bad performance seek?

How likely is it to be missing? If it is expected then yes you should probably fall back to bad performance seek with a warning to the user. If it is unlikely then you can just generate an error with some instructions for the user.

henrybear327:
> > With the ability to obtain the block size array and sha1 original file content from`xattr`, we can now perform SHA1 hashing and seeking normally, provided that `xattr` is present.
> 
> Great :-)
> 
> > Questions
> > 
> > * If `xattr` is missing, should I fallback to bad performance seek?
> 
> How likely is it to be missing? If it is expected then yes you should probably fall back to bad performance seek with a warning to the user. If it is unlikely then you can just generate an error with some instructions for the user.

This is the part I have no idea on. I would think it's probably not very common since this is the only way to quickly check if 2 files are the same content or not as they are also rolling out the desktop client. But I can't be sure about this.

henrybear327:
A quick question: how long does the sync test usually takes to run on other crypto-based backends? 

Currently, all `test_all` integration tests have passed, but for sync test, it's pretty slow, e.g. `TestSyncConcurrentTruncate` takes `--- PASS: TestSyncConcurrentTruncate (1104.78s)` to run.

backend/protondrive/protondrive.go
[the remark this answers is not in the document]
henrybear327: Mentioned in the documentation!

fstest/fstest.go
```diff
@@ -283,6 +283,7 @@ func CheckListingWithRoot(t *testing.T, f fs.Fs, dir string, items []Item, expec
 		}
 
 		gotListing = makeListingFromObjects(objs)
+
```
ncw, line 286: This looks like a stray change!

ncw (commented):
This is looking great now!

There is one stray change I spotted.

Is this ready for merge now do you think?

henrybear327:
> This is looking great now!
> 

Thank you so much for the guidance and help! It's been quite a journey! 

> There is one stray change I spotted.

Sorry, I have rolled it back...

> 
> Is this ready for merge now do you think?

I think this is ready to be merged! I have updated the library for the one last time for now to fix a typo. 

I will be optimizing the bridging library in the coming weeks, but I don't foresee any more major changes to the rclone codebase besides updating the `go.mod` part here. As the main improvements are thumbnail support and caching improvement, or more integrity checks on file download, etc., which are all contained in the bridging library from what I can think of now!

Will be happy to be the maintainer of this backend! Please let me know how we can proceed with this :))

Have a nice day!

Henry

backend/protondrive/protondrive.go
```diff
@@ -0,0 +1,1069 @@
...
+	// sha1 not cached
+	log.Println("sha1 not cached")
+	// we fetch and try to obtain the sha1 of the link
+	fileSystemAttrs, err := o.fs.protonDrive.GetActiveRevisionAttrsByID(ctx, o.ID())
+	if err != nil {
+		return "", err
+	}
+
+	if fileSystemAttrs == nil || fileSystemAttrs.Digests == "" {
+		return "", ErrFileSHA1Missing
```
henrybear327, line 806: @ncw one last quick question, if the sha1 is somehow not present in the file attribute from the API, should I probably throw `hash.ErrUnsupported` instead of my custom error so that operations like Copy() can handle it better?

Erik1000:
Thanks @henrybear327 and @ncw! Your work looks promising :heart: Can't wait to use this in the next release.

backend/protondrive/protondrive.go
ncw: If a given file doesn't have an sha1 then return empty string with no error - that's what the other backends do.

I'd probably write a debug also if you don't think it will be common.

ncw:
> > This is looking great now!
> 
> Thank you so much for the guidance and help! It's been quite a journey!

:-)  Its not easy writing an rclone backend and you've written a nice library as well :-)

> > There is one stray change I spotted.
> 
> Sorry, I have rolled it back...

no probs
 
> > Is this ready for merge now do you think?
> 
> I think this is ready to be merged! I have updated the library for the one last time for now to fix a typo, and also, squashed all the commits into 1 and uses the in-house commit message.

Great.

I'll merge it now and you can fixup the hash thing later.
 
> I will be optimizing the bridging library in the coming weeks, but I don't foresee any more major changes to the rclone codebase besides updating the `go.mod` part here. As the main improvements are thumbnail support and caching improvement, or more integrity checks on file download, etc., which are all contained in the bridging library from what I can think of now!

Great.
 
> Will be happy to be the maintainer of this backend! Please let me know how we can proceed with this :))

After I've merged this I'll send you an invite via GitHub. Your first task will be to add yourself to the MAINTAINERS file!

You might want to see if you can round up some testers too @Erik1000 - maybe you'd like to help? You could also ask on [on the rclone forum](https://forum.rclone.org/) about this.

Erik1000:
> > > This is looking great now!
> 
> > 
> 
> > Thank you so much for the guidance and help! It's been quite a journey!
> 
> 
> 
> :-)  Its not easy writing an rclone backend and you've written a nice library as well :-)
> 
> 
> 
> > > There is one stray change I spotted.
> 
> > 
> 
> > Sorry, I have rolled it back...
> 
> 
> 
> no probs
> 
>  
> 
> > > Is this ready for merge now do you think?
> 
> > 
> 
> > I think this is ready to be merged! I have updated the library for the one last time for now to fix a typo, and also, squashed all the commits into 1 and uses the in-house commit message.
> 
> 
> 
> Great.
> 
> 
> 
> I'll merge it now and you can fixup the hash thing later.
> 
>  
> 
> > I will be optimizing the bridging library in the coming weeks, but I don't foresee any more major changes to the rclone codebase besides updating the `go.mod` part here. As the main improvements are thumbnail support and caching improvement, or more integrity checks on file download, etc., which are all contained in the bridging library from what I can think of now!
> 
> 
> 
> Great.
> 
>  
> 
> > Will be happy to be the maintainer of this backend! Please let me know how we can proceed with this :))
> 
> 
> 
> After I've merged this I'll send you an invite via GitHub. Your first task will be to add yourself to the MAINTAINERS file!
> 
> 
> 
> You might want to see if you can round up some testers too @Erik1000 - maybe you'd like to help? You could also ask on [on the rclone forum](https://forum.rclone.org/) about this.
> 
> 
> 
> 

Gonna test this in my setup with restic for backups ASAP ❤️

backend/protondrive/protondrive.go
henrybear327: Thanks @ncw, I will proceed with the change and run the integration tests for the one last time!

henrybear327:
> > > This is looking great now!
> > 
> > 
> > Thank you so much for the guidance and help! It's been quite a journey!
> 
> :-) Its not easy writing an rclone backend and you've written a nice library as well :-)

Thank you for all the guidance and help!

> 
> > > There is one stray change I spotted.
> > 
> > 
> > Sorry, I have rolled it back...
> 
> no probs
> 
> > > Is this ready for merge now do you think?
> > 
> > 
> > I think this is ready to be merged! I have updated the library for the one last time for now to fix a typo, and also, squashed all the commits into 1 and uses the in-house commit message.
> 
> Great.
> 
> I'll merge it now and you can fixup the hash thing later.
> 
> > I will be optimizing the bridging library in the coming weeks, but I don't foresee any more major changes to the rclone codebase besides updating the `go.mod` part here. As the main improvements are thumbnail support and caching improvement, or more integrity checks on file download, etc., which are all contained in the bridging library from what I can think of now!
> 
> Great.
> 
> > Will be happy to be the maintainer of this backend! Please let me know how we can proceed with this :))
> 
> After I've merged this I'll send you an invite via GitHub. Your first task will be to add yourself to the MAINTAINERS file!

I have received the email and will join after I got home today! Thank you!
For the MAINTAINERS file, I will update it later, too!
> 
> You might want to see if you can round up some testers too @Erik1000 - maybe you'd like to help? You could also ask on [on the rclone forum](https://forum.rclone.org/) about this.

Is there anything I can do to help on this :)

ncw:
> Is there anything I can do to help on this :)

Make a forum account and post a message saying you are looking for testers!

backend/protondrive/protondrive.go
henrybear327: > If a given file doesn't have an sha1 then return empty string with no error - that's what the other backends do.

Fix in #7163

PR #7100, opened 2023-06-30
build: update to go1.21rc2 and make go1.19 the minimum required version

darthShadow:

Update to 1.21-rc.2 for builds.

N/A

#### Checklist


darthShadow:
I think we should also add a pgo file obtained from running the rclone testsuite multiple times for however little optimization it gives over the default but that should probably come from @ncw.

backend/cache/cache_upload_test.go
```diff
@@ -160,11 +160,11 @@ func TestInternalUploadQueueMoreFiles(t *testing.T) {
 	minSize := 5242880
 	maxSize := 10485760
 	totalFiles := 10
-	rand.Seed(time.Now().Unix())
+	randInstance := rand.New(rand.NewSource(time.Now().Unix()))
```
ncw, line 163: Why did we need these changed? Did something complain?

ncw (commented):
Thanks for doing this :-)

I think you meant make go1.19 the minimum!

When doing this I also do `git grep go1.18` and `git grep go1.19` see if I can remove any constraints.

If go1.19 is the minimum we don't need to go1.18 part of these constraints. (In fact we didn't need it for go1.18)

```
backend/azureblob/azureblob.go://go:build !plan9 && !solaris && !js && go1.18
backend/azureblob/azureblob.go:// +build !plan9,!solaris,!js,go1.18
backend/azureblob/azureblob_internal_test.go://go:build !plan9 && !solaris && !js && go1.18
backend/azureblob/azureblob_internal_test.go:// +build !plan9,!solaris,!js,go1.18
backend/azureblob/azureblob_test.go://go:build !plan9 && !solaris && !js && go1.18
backend/azureblob/azureblob_test.go:// +build !plan9,!solaris,!js,go1.18
backend/azureblob/azureblob_unsupported.go://go:build plan9 || solaris || js || !go1.18
backend/azureblob/azureblob_unsupported.go:// +build plan9 solaris js !go1.18
```

backend/cache/cache_upload_test.go
darthShadow: This came about as part of the linting tests for go1.20. There are a few more similar complaints, which I haven't gotten around to fixing, so I can separate these out and put them all as part of a separate PR if you want.

darthShadow:
> make go1.19 the minimum

That's what I thought too but the earlier bump from you (1680c5af8f2408a66f2e1aea11ab7b15b9db70af) had go 1.17 as the mention so I thought I was missing something :)

---

Will update the constraints and the commit message accordingly.

---

Not sure about the windows failures though... 😕

ncw:
> > make go1.19 the minimum
> 
> That's what I thought too but the earlier bump from you ([1680c5a](https://github.com/rclone/rclone/commit/1680c5af8f2408a66f2e1aea11ab7b15b9db70af)) had go 1.17 as the mention so I thought I was missing something :)

:-) Just another typo from me!

> Will update the constraints and the commit message accordingly.

:+1:

> Not sure about the windows failures though... 😕

Looks like one of the occasionally failing windows tests - I gave it a retry. It could of course be a regression in which case we'll have to investigate further.

backend/cache/cache_upload_test.go
ncw: Which lint did you run? The golangci-lint we use as part of the build is passing so it thinks everything is OK now!

Did it fail before those changes?

darthShadow: Yeah, govet as part of golangci-lint, but I ran it with go.mod temporarily modified to 1.20 rather than 1.19.

It's only a problem with Go 1.20 where the above function was deprecated: https://pkg.go.dev/math/rand#Seed (https://tip.golang.org/doc/go1.20#math/rand)

ncw:
I re-ran the Windows build.

The macos build is failing because of https://github.com/Homebrew/homebrew-core/issues/135315 which should be fixed but still seems to be failing intermittently.

ncw:
The windows error is definitely a regression - I managed to replicate it on my Windows VM.

Will attempt to bisect the go source now to find the root cause.

darthShadow:
Great, thanks.

I was trying to go through the release notes for any possible windows-specific changes that may have caused this and the only thing that even seems related in anyway is the TempDir change but I don't think that's the cause.

> On Windows calling [TempDir](https://tip.golang.org/pkg/os/#TempDir) now uses GetTempPath2W when available, instead of GetTempPathW. The new behavior is a security hardening measure that prevents temporary files created by processes running as SYSTEM to be accessed by non-SYSTEM processes.

ncw:
I had a go at bisecting it but the bisect kept getting stuck. Go built properly but it could no longer build the test giving this error

```
~/go/src/github.com/rclone/rclone/cmd/cmount$ go version && GOOS=windows go test -v -c -tags cmount
go version devel go1.21-b15297fcd2 Fri Jan 27 21:59:45 2023 +0000 linux/amd64
# github.com/rclone/rclone/cmd/cmount.test
golang.org/x/sys/cpu.runtime_getAuxv·f: relocation target runtime.getAuxv not defined
```

That narrowed it down to about 400 commits and the one you mentioned was part of it.

I reverted it (with a bit of conflict fixing) and discovered that was not the problem.

I looked at the commits with Windows in the title and tried reverting them...

```
34e701ece2 path/filepath: document that Clean does not change Windows volume names
95eb5abd63 path/filepath: do not Clean("a/../c:/b") into c:\b on Windows
79e01ab7e6 cmd/dist: remove windows-amd64-2008 code path
38d1a6665f syscall: regenerate zsyscall_windows.go
4952090739 internal/syscall/windows: regenerate zsyscall_windows.go
932d0ae83e internal/poll: don't queue port completion entry for udp on windows
007d8f4db1 time: update windows zoneinfo_abbrs
26fd1fba12 os: use GetTempPath2 on Windows if available - TRIED
2c7856087a runtime: delete unused cbctxts variable on windows assembly
1d3088effd os: make Lstat for symlinks on Windows consistent with POSIX - TRIED
cc82867f6b runtime/race: update race_windows_amd64.syso
083d94f69c runtime: use explicit NOFRAME on windows/amd64
1951857ec0 os: use handle based APIs to read directories on windows - TRIED - SUCCESS
7b5a34418c runtime: remove unused badsignal2 on windows
cf9263dee1 runtime: factor out windows sigtramp
030ca34f57 path/filepath: remove extra Clean call in EvalSymlinks on Windows
28f8dbd7b9 runtime,cmd/internal/obj/x86: use TEB TLS slots on windows/i386
e587a7691b cmd/link: fix incorrect DOS header on Windows binaries
```

Eventually this one turned out to be the culprit

    1951857ec0 os: use handle based APIs to read directories on windows

It isn't particularly straightforward to revert - I had to fix a quite a few conflicts.

I searched on github and there are no open issues about this so I reported an issue here https://github.com/golang/go/issues/61159

darthShadow:
Awesome work, thanks.

I have also fixed the MacOS failures since we don't actually require the taps in the newer homebrew versions which default to API installs instead.

ncw:
rc3 is out now and should have the windows fix if you want to update this 😀

darthShadow:
Pushed. 🤞

darthShadow:
Looks like all checks passed so this is good to merge now.

ncw (approved):

I'll merge now.

darthShadow:
Just a reminder that a profile for PGO still needs to be added.

ncw:
Just read the docs https://go.dev/doc/pgo on this - apparently a pgo of the unit tests is not much good!

What we probably want is a pgo of rclone
- doing some syncs to and from representative backends (say local, s3, dropbox, drive would cover the top ones)
- running `rclone mount` and exercising the VFS - probably a mount of local would be fine for this.

Do you know if you can combine profiles? Otherwise it is going to be difficult to generate the right profiles. I guess it could be scripted via the rc.

I think we should probably have another issue about this - can you make one? Thank you :-)

darthShadow:
https://github.com/rclone/rclone/issues/7142

Issue #7103, opened 2023-07-01
[rclone v1.63.0] remote Nextcloud not working after upgrade from 1.62.2 to 1.63.0


mpsd:
Thank you for developing rclone!

#### What is the problem you are having with rclone?

After upgrading from version 1.62.2 to 1.63.0 the use of Nextcloud remote dumps with
`2023/07/01 12:43:26 Failed to create file system for "***my-remotename***": chunked upload with nextcloud must use /dav/files/USER endpoint not /webdav`

I use already that hinted endpoint (/dav/files/USER), have tried to delete and add the remote via `rclone config` again and from scratch and got the same error.

My `rclone config` entry is
Editing existing "***my-remotename***" remote with options:
- type: webdav
- url: https://***my webservice url***/remote.php/dav/files/***my-username***/Directory
- vendor: nextcloud
- pass: *** ENCRYPTED ***
- user: ***my-username***

#### What is your rclone version (output from `rclone version`)
rclone v1.63.0
- os/version: debian 11.7 (64 bit)
- os/kernel: 6.1.0-0.deb11.7-amd64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.20.5
- go/linking: static
- go/tags: none

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Debian Linux 11.7

#### Which cloud storage system are you using? (e.g. Google Drive)
Nextcloud v27 (worked with v1.62.2 / doesn't work with 1.63.0) 

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
All rclone commands like `ls` or `copy`etc.

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
`rclone -vv ls "***my-remotename***":`
`2023/07/01 12:56:23 DEBUG : rclone: Version "v1.63.0" starting with parameters ["rclone" "-vv" "ls" "***my-remotename***:"]`
`2023/07/01 12:56:23 DEBUG : Creating backend with remote "***my-remotename***:"`
`2023/07/01 12:56:23 DEBUG : Using config file from "/root/.config/rclone/rclone.conf"`
`2023/07/01 12:56:23 DEBUG : found headers:`
`2023/07/01 12:56:23 Failed to create file system for "***my-remotename***:": chunked upload with nextcloud must use /dav/files/USER endpoint not /webdav`

#### How to use GitHub

[...]

PR #7108, opened 2023-07-02
webdav: nextcloud: fix must use /dav/files/USER endpoint not /webdav error

devnoname120:
Fix https://github.com/rclone/rclone/issues/7103

Before this change the RegExp validating the endpoint URL was a bit
too strict allowing only /dav/files/USER due to chunking limitations.

This patch adds back support for /dav/files/USER/dir/subdir etc.

Co-authored-by: Nick Craig-Wood <nick@craig-wood.com>

### DIFF
--- a/backend/webdav/chunking.go
+++ b/backend/webdav/chunking.go
@@ -14,7 +14,6 @@ import (
 	"io"
 	"net/http"
 	"path"
-	"strings"
 
 	"github.com/rclone/rclone/fs"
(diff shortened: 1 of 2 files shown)


devnoname120:
@ncw Review requested

ncw (approved):
This looks good to me - thank you :-)

Let's get it some testing and if all goes well we can put it into the v1.63.1 patch release.

PR #7115, opened 2023-07-04
Box: Fix reconnect failing with HTTP 400 Bad Request

albertony:
The error is:

> Error: failed to configure token with jwt authentication: jwtutil: failed making auth request: 400 Bad Request

With the following additional debug information:

> jwtutil: Response Body: {"error":"invalid_grant","error_description":"Please check the 'aud' claim. Should be a string"}

Problem is that in jwt-go the RegisteredClaims type has Audience field (aud claim) that is a list, while box apparantly expects it to be a singular string. In jwt-go v4 we currently use there is an alternative type StandardClaims which matches what box wants. Unfortunately StandardClaims is marked as deprecated, and is removed in the newer v5 version, so we this is a short term fix only. Edit: Suggested long term fix compatible with v5 here: https://github.com/rclone/rclone/pull/7116.

Fixes #7114

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ ] I have added tests for all changes in this PR if appropriate.
- [ ] I have added documentation for the changes if appropriate.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [x] I'm done, this Pull Request is ready for review :-)

Issue #7114, opened 2023-07-04
v.1.63.0 regression on command "rclone config reconnect <theRemoteName>:"


minnieshi:
#### What is the problem you are having with rclone?
after 30th June, the release v1.63.0 is out. our rclone command stops working and below is the comparision for the command.
error message
```
Error: failed to configure token with jwt authentication: jwtutil: failed making auth request: 400 Bad Request
```

#### What is your rclone version (output from `rclone version`)
v1.63.0

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
linux

#### Which cloud storage system are you using? (e.g. Google Drive)
Box

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
```
rclone -vv config reconnect source_box:
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
**Below is the older version example, run on Mac, but it worked for linux too**

```
2023/07/03 17:43:20 DEBUG : Setting default for box-box-config-file="/Users/blahblah/.config/rclone/box_apps_config.json" from environment variable RCLONE_BOX_BOX_CONFIG_FILE
2023/07/03 17:43:20 DEBUG : rclone: Version "v1.61.1" starting with parameters ["rclone" "-vv" "config" "reconnect" "source_box:"]
2023/07/03 17:43:20 DEBUG : Using config file from "/Users/blahblah/.config/rclone/rclone.conf"
2023/07/03 17:43:20 DEBUG : source_box: config in: state="", result=""
2023/07/03 17:43:20 DEBUG : Setting box_box_config_file="/Users/blahblah/.config/rclone/box_apps_config.json" from environment variable RCLONE_BOX_BOX_CONFIG_FILE
2023/07/03 17:43:20 DEBUG : jwtutil: Response Body: {"access_token":"<blah blah blah blah purposely removed by me>CNGA4jCvXDtt0pvQZDpzNOA3lsn0WSvC5MSqnnFW0uqjbvsEqjxnQOCYHvVNd6Rg44wxw0OU","expires_in":3948,"restricted_to":[],"token_type":"bearer"}
2023/07/03 17:43:20 DEBUG : Saving config "token" in section "source_box" of the config file
2023/07/03 17:43:20 DEBUG : source_box: Saved new token in config file
2023/07/03 17:43:20 DEBUG : source_box: config out: out=<nil>, err=<nil>
[...]

PR #7116, opened 2023-07-04
Compatibility with jwt-go (golang-jwt) v5

albertony:
Includes the StandardClaims implementation from jwt-go v4, where it was marked as deprecated before removed in v5. The box backend needs this.

See #7115

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -27,7 +27,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/golang-jwt/jwt/v4"
 	"github.com/rclone/rclone/backend/box/api"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config"
@@ -77,7 +76,7 @@ var (
 )
 
 type boxCustomClaims struct {
-	jwt.RegisteredClaims
+	jwtutil.StandardClaims
 	BoxSubType string `json:"box_sub_type,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


ncw:
Thanks for having a go at this @albertony 

> Including the StandardClaims implementation from jwt-go v4 directly into rclone's jwtutil. In jwt-go v4 it was marked as deprecated, before removed in v5. The box backend needs this, due to the singular string audience field, see: #7114.
> 
> This is a possible long-term solution for the jwt-go v4 fix in #7115. Not the best solution, but not the worst either...

It has its charms :-) It is explicit which is nice.

We should include the [COPYRIGHT](https://github.com/golang-jwt/jwt/blob/main/LICENSE) at the top of the borrowed file.

> Maybe we should rename our variant from StandardClaims, which is the original name that was used in jwt-go, to something like LegacyClaims - which avoids confusing contributors into thinking this is one to choose by default?

How about `LegacyStandardClaims`?

> Maybe we could just go ahead and update jwt-go to v5 as well?

I think that would be a great idea to do as part of this, though I think the `LegacyStandardClaims` will have to change to fit the new interface.

Do you have a way of testing this with box?

ncw:
I'm pretty sure I can test this. However the JWT stuff is broken for me at the moment - see #7214 - that is in the process of being fixed though.

Do you want to update it to jwt 5 and we can merge and fix if necessary?

albertony:
Great. I'm a bit on thin ice here, and you are right the implementation needs to change to satisfy the new interface (which was probably a part that I missed in the original PR). But if I'm not mistaken, the implementation is now quite simple.

What do you think?

(I was tricked by a change in my local linter setup, and needed a couple of extra pushes to get rid of a lint issue. I think it is ok now, but the current lint run on github failed with a http 500 response. I will re-trigger it as soon that is possible.)

PR #7137, opened 2023-07-14
--max-duration flag -  add new exit code (10)

kapitainsky:
The purpose of this PR is to add dedicated exit code (10) for `--max-duration` flag.

Rclone will exit with exit code 10 if the duration limit is reached.

It will behave in similar fashion as `--max-transfer` and exit code 8.

discussed on the forum:

https://forum.rclone.org/t/max-duration-option-is-triggering-exit-with-error/39917/6

### DIFF
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -20,6 +20,14 @@ import (
 	"github.com/rclone/rclone/fs/operations"
 )
 
+// ErrorMaxDurationReached defines error when transfer duration is reached
+// Used for checking on exit and matching to correct exit code.
+var ErrorMaxDurationReached = errors.New("max transfer duration reached as set by --max-duration")
+
+// ErrorMaxDurationReachedFatal is returned from when the max
+// duration limit is reached.
+var ErrorMaxDurationReachedFatal = fserrors.FatalError(ErrorMaxDurationReached)
+
 type syncCopyMove struct {
 	// parameters
 	fdst               fs.Fs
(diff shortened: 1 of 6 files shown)


kapitainsky:
I have tried to follow the same logic as was used with `max-transfer` error code but can not get it fully work yet.

It fails in cmg.go:

`case errors.Is(err, sync.ErrorMaxDurationReached):`

I checked what err unwrap to:

```
_, unwrapped := fserrors.Cause(err)
fmt.Printf("err = %#v\n", unwrapped)

err = &errors.errorString{s:"max transfer duration reached as set by --max-duration"}
```

so error is here:) but `sync.ErrorMaxDurationReached` seems unreferenced

fs/sync/sync.go
```diff
@@ -20,6 +20,14 @@ import (
 	"github.com/rclone/rclone/fs/operations"
 )
 
+// errorMaxDurationReached defines error when transfer duration is reached
+// Used for checking on exit and matching to correct exit code.
+var errorMaxDurationReached = errors.New("max transfer duration reached as set by --max-duration")
```
ncw, line 25: Make the name `ErrorMaxDurationReached` and it will be exported. You'll need to change the comment above too.

kapitainsky: sorry for that - it was capitalized. but I manually copied changes from one branch to another and used wrong clip book.

Anyway - now it is corrected.

darthShadow: You also need to import `fs/sync` so a import like this:

`fssync "github.com/rclone/rclone/fs/sync"` (fssync so that it doesn't collide with the golang sync package).

kapitainsky: "fssync" was what I was missing - I could not figure out how to import it.

Thank you very much for help. It is already working and throwing exit code 10:)

Now a bit of tests and last touches and should be ready for review.

kapitainsky:
This PR is ready for review.

ncw (approved):

I'll merge now.

Issue #3558, opened 2019-09-20
SOCKS5 proxy support for FTP, SFTP


fdefilippo:
#### What is your current rclone version (output from `rclone version`)?

1.49.3

#### What problem are you are trying to solve?

By supporting the SOCKS5 type proxy rclone can connect from within enterprise dmz using any protocol.

#### How do you think rclone should be changed to solve that?

Implementing the RFC-1928 using the golang package: 

https://godoc.org/golang.org/x/net/proxy#SOCKS5

J0s3f:
Rclone has socks5 proxy support: https://rclone.org/faq/#can-i-use-rclone-with-an-http-proxy

just start e.g. with:

	export http_proxy=socks5://proxyserver:1080
	export https_proxy=$http_proxy
	export HTTP_PROXY=$http_proxy
	export HTTPS_PROXY=$http_proxy
	rclone ...

fdefilippo:
> Rclone has socks5 proxy support: https://rclone.org/faq/#can-i-use-rclone-with-an-http-proxy
> 
> just start e.g. with:
> 
> ```
> export http_proxy=socks5://proxyserver:1080
> export https_proxy=$http_proxy
> export HTTP_PROXY=$http_proxy
> export HTTPS_PROXY=$http_proxy
> rclone ...
> ```
> 
> Documentation is missing as described in #1861

No, unfortunately you are confusing with the support of the socks for the http/s protocol. with Socks Type proxy you can incapsulate any protocols: sftp and so on.

ivandeex:
The FTP client library used by rclone supports custom Dial func option: https://pkg.go.dev/github.com/jlaffaye/ftp#DialWithDialFunc
and we can feed it with https://pkg.go.dev/golang.org/x/net/proxy#Dialer

ivandeex:
I'm yet curious how we can tweak rclone test suite to add a mock of socks5 proxy.

fdefilippo:
but is it not enough to create a dialer that uses socks if set? something like:

```
// socks user
        auth := proxy.Auth{
                User:     "user",
                Password: "password",
        }
// create a socks5 dialer
        dialer, err := proxy.SOCKS5("tcp", SOCKS + ":" + PORT, &auth, proxy.Direct)
        if err != nil {
                fmt.Fprintln(os.Stderr, "can't connect to the proxy:", err)
                os.Exit(1)
        }
[...]

PR #7158, opened 2023-07-20
Add socks_proxy support for ftp/sftp/ssh backend

snowzach:
Add support for socks 5 proxies to the SFTP/SSH backend

https://github.com/rclone/rclone/issues/3558

#### Checklist

### DIFF
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -178,7 +178,7 @@ If this is set and no password is supplied then rclone will ask for a password
 		}, {
 			Name:    "socks_proxy",
 			Default: "",
-			Help: `Socks 5 proxy host for both SSH and SFTP connections.
+			Help: `Socks 5 proxy host.
 		
 		Supports the format user:pass@host:port, user@host:port, host:port.
 		
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -28,8 +28,8 @@ import (
 	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/env"
 	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/proxy"
 	"github.com/rclone/rclone/lib/readers"
-	"golang.org/x/net/proxy"
 )
 
 var (
@@ -375,30 +375,7 @@ func (f *Fs) ftpConnection(ctx context.Context) (c *ftp.ServerConn, err error) {
 		}()
 		baseDialer := fshttp.NewDialer(ctx)
 		if f.opt.SocksProxy != "" {
-			var (
-				proxyAddress string
-				proxyAuth    *proxy.Auth
-				proxyDialer  proxy.Dialer
-			)
-			if credsAndHost := strings.SplitN(f.opt.SocksProxy, "@", 2); len(credsAndHost) == 2 {
-				proxyCreds := strings.SplitN(credsAndHost[0], ":", 2)
-				proxyAuth = &proxy.Auth{
-					User: proxyCreds[0],
-				}
-				if len(proxyCreds) == 2 {
-					proxyAuth.Password = proxyCreds[1]
-				}
-				proxyAddress = credsAndHost[1]
-			} else {
-				proxyAddress = credsAndHost[0]
-			}
-			// Build the proxy dialer
-			proxyDialer, err = proxy.SOCKS5("tcp", proxyAddress, proxyAuth, baseDialer)
-			if err != nil {
-				return nil, fmt.Errorf("failed to create proxy dialer: %w", err)
-			}
-			conn, err = proxyDialer.Dial(network, address)
-
+			conn, err = proxy.SOCKS5Dial(network, address, f.opt.SocksProxy, baseDialer)
 		} else {
 			conn, err = baseDialer.Dial(network, address)
 		}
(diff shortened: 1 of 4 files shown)


backend/ftp/ftp.go
```diff
@@ -174,6 +175,18 @@ Enabled by default. Use 0 to disable.`,
 If this is set and no password is supplied then rclone will ask for a password
 `,
 			Advanced: true,
+		}, {
+			Name:    "socks_proxy",
+			Default: "",
+			Help: `Socks 5 proxy host for both SSH and SFTP connections.
```
ncw, line 181: This should just say `Socks 5 proxy host.` - it is for the FTP backend so the ssh part isn't relevant.

backend/sftp/sftp.go
```diff
@@ -388,6 +390,18 @@ Example:
     ssh-ed25519 ssh-rsa ssh-dss
 `,
 			Advanced: true,
+		}, {
+			Name:    "socks_proxy",
+			Default: "",
+			Help: `Socks 5 proxy host for both SSH and SFTP connections.
```
ncw, line 396: Again just `Socks 5 proxy host.`

backend/sftp/sftp.go
```diff
@@ -467,8 +482,36 @@ type Object struct {
...
+			}
+			if len(proxyCreds) == 2 {
+				proxyAuth.Password = proxyCreds[1]
+			}
+			proxyAddress = credsAndHost[1]
+		} else {
+			proxyAddress = credsAndHost[0]
+		}
+		// Build the proxy dialer
+		proxyDialer, _ := proxy.SOCKS5("tcp", proxyAddress, proxyAuth, baseDialer)
```
ncw, line 508: Don't ignore the error please!

backend/ftp/ftp.go
```diff
@@ -359,7 +373,31 @@ func (f *Fs) ftpConnection(ctx context.Context) (c *ftp.ServerConn, err error) {
...
+				}
+				if len(proxyCreds) == 2 {
+					proxyAuth.Password = proxyCreds[1]
+				}
+				proxyAddress = credsAndHost[1]
+			} else {
+				proxyAddress = credsAndHost[0]
+			}
+			// Build the proxy dialer
+			proxyDialer, _ := proxy.SOCKS5("tcp", proxyAddress, proxyAuth, baseDialer)
```
ncw, line 395: Don't ignore error.

ncw (commented):
Thanks for doing this - this has been an often requested feature.

See comments inline.

Perhaps rclone should be picking up the SOCKS5 proxy from environment vars like we do for the http services?

https://rclone.org/faq/#can-i-use-rclone-with-an-http-proxy

Are there standard env vars we could/should use instead of the backend configs?

Also the code duplicated between the sftp and ftp backends is slightly annoying! Maybe it could be factored into lib/proxy or something like that? I suspect the code might grow and its just on the edge of too much duplication for my tastes!

Thank you

backend/ftp/ftp.go
snowzach: Updated

backend/sftp/sftp.go
snowzach: Updated

backend/sftp/sftp.go
snowzach: I am not sure why the function returns an error. It doesn't actually trigger one.. but it looks bad so I handled it..

backend/ftp/ftp.go
snowzach: Same as above

snowzach:
Yeah, makes sense to make proxy dialing it's own package. I moved it to `lib/proxy` in case you want to do more with it someday.

ncw (approved):
Looks very nice now 😀

What did you think of the idea of picking up environment variables?

I can merge this and we can discuss further or you could update this , whichever you like.

snowzach:
I think this one probably stands pretty good on it's own. I can see cases where you want to configure a socks proxy on it's own for one connection and not everything (which is actually what I need for my use case). The other option being implementing something globally on your `fshttp.NewDialer` function. That's gonna effect everything that uses it though. 

I would be willing to work on that if you think it's needed. The only danger being if you configure both... I am sure that could be guarded against though. 

I'd be happy to look into that for you if you've received enough requests for it. I'll back out these changes if it ends up superseding them. (Perhaps an option for NewDialer with the socks proxy endpoint. If not populated defaults to global config which may or may not be for a proxy) I actually started that way but it was a bit messy because there's a lot of Dial/DialContext overlap/usage. For whatever reason the proxy dialer doesn't implement DialContext so you have to create a wrapper for it.

ncw:
OK I will merge this now and we can wait and see if anyone requests environment variables. I'm not sure there is a standard for them - here is the best doc I found

https://gist.github.com/yougg/5d2b3353fc5e197a0917aae0b3287d64#set-ssh-proxy-environment-variables

> The other option being implementing something globally on your `fshttp.NewDialer` function. That's gonna effect everything that uses it though.

I think we already support that for SOCKS5 proxies. I may be wrong though! That is HTTP only backends though.

Thank you for the contribution :-)

SaladinoGW:
Thanks guys, I'm already using this new feature and I imagine many people are too

BunnySakura:
以防有人不知道，使用`rclone config`配置sftp后端，在高级设置找到`socks_proxy`，配置为代理地址即可。配置完成会在配置文件增加一行

```ini
socks_proxy = <IP地址>:<端口>
````

**注意**：低版本rclone不支持此设置

---

Just in case anyone doesn't know, use `rclone config` to configure the sftp backend, find `socks_proxy` in the advanced settings and configure it as a proxy address. When you're done, you'll add a line to the config file

```ini
socks_proxy = <IP>:<port>
````

**Note**: Lower versions of rclone do not support this setting.

PR #7177, opened 2023-07-27
docs: b2 versions names caveat

kapitainsky:
When using --b2-versions flag rclone is relying on the file name
to work out whether the objects are versions or not. Versions' names
are created by inserting timestamp between file name and its extension.

        9 file.txt
        8 file-v2023-07-17-161032-000.txt
       16 file-v2023-06-15-141003-000.txt

If there are real files present with the same names as versions, then
behaviour of --b2-versions can be unpredictable.


ncw (approved):
Thank you :-)

PR #7179, opened 2023-07-28
chore(deps): tidy go modules

qdm12:

See title

No

#### Checklist


ncw:
I've just updated the dependencies so I don't think this is necessary any more.

`go mod tidy` gets run as part of the release process.

I'll close this now, but if it is necessary I can re-open.

Thank you.

qdm12:
FYI It looks like there are still go.sum dependencies that can be removed on the current master branch

ncw:
> FYI It looks like there are still go.sum dependencies that can be removed on the current master branch

My understanding of stuff in `go.sum` is that it is only really used for verification, so having some extra lines in `go.sum` is very low cost. Nothing extra will be downloaded there are just a few lines in the file that aren't used.

> Maybe worth having a CI check run go mod tidy on every PR?

That could work. Maybe in the lint run. How would you do it?

I'm a bit concerned it would make too much noise for Go beginners (and there are plenty of those who contribute to rclone) for too little benefit.

PR #7181, opened 2023-07-29
sftp: support dynamic --sftp-path-override

nielash:

Before this change, rclone always expected `--sftp-path-override` to be the absolute SSH path to `remote:path/subpath` which effectively made it unusable for wrapped remotes (for example, when used with a `crypt` remote, the user would need to provide the full decrypted path.)

After this change, the old behavior remains the default, but dynamic paths are now also supported, if the user adds `@` as the first character of `--sftp-path-override`. Rclone will ignore the `@` and treat the rest of the string as the path to the SFTP remote's root. Rclone will then add any relative subpaths automatically (including unwrapping/decrypting remotes as necessary).

In other words, the `path_override` config parameter can now be used to specify the _difference_ between the SSH and SFTP paths. Once specified in the config, it is no longer necessary to re-specify for each command.

Yes: https://forum.rclone.org/t/sftp-path-override-breaks-on-wrapped-remotes/40025

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -168,7 +168,19 @@ E.g. if shared folders can be found in directories representing volumes:
 
 E.g. if home directory can be found in a shared folder called "home":
 
-    rclone sync /home/local/directory remote:/home/directory --sftp-path-override /volume1/homes/USER/directory`,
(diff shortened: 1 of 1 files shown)


ncw (approved):
Looks great - thank you. Nice docs too :-)  Will merge now.

PR #7185, opened 2023-07-30
docs: box client_id creation

kapitainsky:
based on the forum discussion:

https://forum.rclone.org/t/box-getting-403-errors-when-using-chunker-working-fine-without-it/40388/22


kapitainsky:
it has failed built on mac_arm64 but I guess it has nothing to do with PR itself... but rather with some build scripts... If I am wrong let me know and I investigate further.

ncw (approved):
That is great - thank you :-)

I'll merge it now.

Issue #6632, opened 2022-12-16
Faster Directory Listings in VFS


darthShadow:
#### The associated forum post URL from `https://forum.rclone.org`

N/A

#### What is your current rclone version (output from `rclone version`)?

N/A (but `v1.61.0-beta.6609.ca23c6435`)

#### What problem are you are trying to solve?

Slow directory listings in VFS when backed by a Ceph mount.

#### How do you think rclone should be changed to solve that?

Generally, the Ceph mount is fast enough that these issues are negligible but due to unavoidable issues, the Ceph mount has become slower temporarily which has highlighted these issues.

There are 2 approaches that I can see:

* For the Local Ceph FS: Do the listing of the directory in parallel?
* For the VFS: When the directory is expired, perhaps it can be loaded lazily i.e. return the stale entries till the remote has finished listing the directory and then replace the entries? 

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

darthShadow:
I am mostly looking for thoughts on the above approaches or any alternatives so I can experiment with those and come back with proper numbers and maybe even a PR.

darthShadow:
> Which CEPH interface are you using? S3 or Swift?

This is a CephFS kernel mount so effectively a local filesystem for all intents and purposes. Perhaps we could parallelize at least the `os.Stat` calls?

> but it effectively breaks the directory caching scheme.

Yeah, I have been unable to think of a good way to do this.

The problem is sometimes the MDS servers or remotes can take time (especially with larger directory trees) and any further listing commands on that will get stuck waiting on the locks to be released.

darthShadow:
> Do the directories have many entries?

Yeah, some directories have upto 40k entries.

ncw:
> > Which CEPH interface are you using? S3 or Swift?
> 
> This is a CephFS kernel mount so effectively a local filesystem for all intents and purposes. Perhaps we could parallelize at least the `os.Stat` calls?

That is very do-able and a great idea.

It might also be worth revisiting the listing interface for the local backend. At the moment unix does a file listing then stats each file, whereas windows gets the listing with stats straight from the kernel.

If there was an interface like that for linux it would speed stuff up enormously.

[...]

PR #7194, opened 2023-07-31
feat: local: do stat in parallel controlled by the --checkers option

darthShadow:

Do parallel stat-ing in the local FS for unix platforms

https://github.com/rclone/rclone/issues/6632

#### Checklist


darthShadow:
@ncw This is an initial approach for https://github.com/rclone/rclone/issues/6632

backend/local/local.go
```diff
@@ -488,41 +498,20 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
...
+			fis, err = fd.Readdir(dirListingBatchSize)
 			if err == io.EOF && len(fis) == 0 {
 				break
 			}
 		} else {
 			// For other OSes we read the names only (which shouldn't fail) then stat the
-			// individual ourselves so we can log errors but not fail the directory read.
+			// individual ourselves, so we can log errors but not fail the directory read.
 			var names []string
-			names, err = fd.Readdirnames(1024)
```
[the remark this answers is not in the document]
ncw: I think the 1024 names will provide enough parallelism, unless for some reason, one of them takes a very long time.

ncw:
My preferred way of doing this sort of paralleism is with https://pkg.go.dev/golang.org/x/sync/errgroup

You'll see lots of examples in the rclone code. It also has [SetLimit](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit) which lets you easily limit the concurrency.

I would have thought you should be able to write it entirely inline with `errgroup` with each `errgroup.Go` routine returning its result directly into a pre-allocated `fis` slice.

Here is a sketch of what I mean

```patch
diff --git a/backend/local/local.go b/backend/local/local.go
index bb9f739eb..559f2def0 100644
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -501,35 +501,43 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 				break
 			}
 			if err == nil {
-				for _, name := range names {
+				fis = make([]os.FileInfo, len(names))
+				// FIXME set up errgroup
+				for i, name := range names {
+					i, name := i, name // make sure we add these to the closure
+					errgroup.Go(func() error {
 						namepath := filepath.Join(fsDirPath, name)
 						fi, fierr := os.Lstat(namepath)
 						if os.IsNotExist(fierr) {
 							// skip entry removed by a concurrent goroutine
-							continue
+							return nil
 						}
 						if fierr != nil {
 							// Don't report errors on any file names that are excluded
 							if useFilter {
 								newRemote := f.cleanRemote(dir, name)
 								if !filter.IncludeRemote(newRemote) {
-									continue
+									return nil
 								}
 							}
 							fierr = fmt.Errorf("failed to get info about directory entry %q: %w", namepath, fierr)
 							fs.Errorf(dir, "%v", fierr)
 							_ = accounting.Stats(ctx).Error(fserrors.NoRetryError(fierr)) // fail the sync
-							continue
+							return nil
 						}
-						fis = append(fis, fi)
+						fis[i] = fi
+					})
 				}
 			}
+			err = errgroup.Wait()
 		}
 		if err != nil {
 			return nil, fmt.Errorf("failed to read directory entry: %w", err)
 		}
 
 		for _, fi := range fis {
+			if fi == nil {
+				continue
+			}
 			name := fi.Name()
 			mode := fi.Mode()
 			newRemote := f.cleanRemote(dir, name)

```

darthShadow:
I have also parallelized the fsObject creation now.

I used the ants library since I wanted to keep the goroutines persistent rather than creating and destroying them repeatedly but I can switch to errgroup if you want.

darthShadow:
All changes are done from my end and the tests are passing so this PR is ready for review now. I will add a few before & after results shortly.

darthShadow:
On further testing, I think it may be a good idea to run the for loop of the 1024 dirnames in a parallel thread too since I haven't been able to see too much of a benefit with the current approach (~15-30s on average for a ~5m total listing, comparing the before/after) on both Ceph (via Kernel) & NFS.

I believe that's because both of these fetch the file/dir stat along with the listing and doing a stat separately seems to only test the kernel syscall which isn't a lot.

backend/local/local.go
```diff
@@ -494,92 +496,145 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
...
+					// Check whether this link should be translated
+					if f.opt.TranslateSymlinks && fi.Mode()&os.ModeSymlink != 0 {
+						newRemote += linkSuffix
+					}
+					// Don't include non directory if not included
+					// we leave directory filtering to the layer above
+					if useFilter && !filter.IncludeRemote(newRemote) {
+						return nil
+					}
+					fso, err := f.newObjectWithInfo(newRemote, fi)
```
darthShadow, line 613: There also seems to be lock contention here for the lock on `fs.objectMetaMu`, but I haven't been able to understand why it's a global fs-level lock rather than an object-specific lock. All it's uses seem to be for object-specific metadata.

darthShadow:
I will do some further testing over the weekend by creating a large directory with a million+ files to see if we see any benefits there since that seems the ideal scenario which will be benefited by this change.

backend/local/local.go
ncw: Its a global because I didn't want to make the Object struct any bigger. However in needs to be a local then I suggest you add another commit to this PR making it local to Object. We can then see if that helps.

ncw:
> I have also parallelized the fsObject creation now.
> 
> I used the ants library since I wanted to keep the goroutines persistent rather than creating and destroying them repeatedly but I can switch to errgroup if you want.

I see you've done the errgroup change - looks very neat :-)

> All changes are done from my end and the tests are passing so this PR is ready for review now. I will add a few before & after results shortly.

OK. I gave the code a quick run through and I think it looks OK, but I need to give it a careful review when I am more awake!

> On further testing, I think it may be a good idea to run the for loop of the 1024 dirnames in a parallel thread too since I haven't been able to see too much of a benefit with the current approach (~15-30s on average for a ~5m total listing, comparing the before/after) on both Ceph (via Kernel) & NFS.

Can we run the dirnames in parallel? `os.Readdirnames` needs to be called in sequence but I guess we could be doing the stat calls in parallel with the `os.Readdirnames` which I think is what you mean.

This would need some quite careful coding as `append` will reallocate the array so this needs to work in the chunks returned by `os.Readdirnames`.

> I believe that's because both of these fetch the file/dir stat along with the listing and doing a stat separately seems to only test the kernel syscall which isn't a lot.
> 
> I was able to validate the above behaviour with doing a `ls -1` followed by a `ls -l` which took around the same time (difference in 10s of ms).

Interesting. It might be worth timing the total `os.Readdirnames` execution time for a large directory because that will be the fastest we can go.

I wrote a little program to run `os.Readdirnames` and call stat to see the timings.

https://play.golang.org/p/babvjaBRQUL

If I run this on a local directory I get this

```
Time to read directory with 19971 entries: 11.876484ms
Time to stat 19971 directory entries: 46.399772ms
```

So this can never run faster than 11ms since that is how long the read directory takes and that can't be parallelized.

What do you get with that program on your CEPH & NFS tests?

darthShadow:
> doing the stat calls in parallel with the os.Readdirnames

I have been doing further reading on this and I think we can actually skip the stat call altogether until it's actually needed (for creating the fs.object) by using the newer `ReadDir` call.

Reference: https://pkg.go.dev/os#File.ReadDir

It returns `[]DirEntry` (https://pkg.go.dev/io/fs#DirEntry) which has the info about the name, type & mode, which is all we need for further processing until the time of the actual object creation.

ncw:
> > doing the stat calls in parallel with the os.Readdirnames
> 
> I have been doing further reading on this and I think we can actually skip the stat call altogether until it's actually needed (for creating the fs.object) by using the newer `ReadDir` call.

This is an abstraction over the fact that Windows returns the stat info which makes the code nicer. Combining the stat with the object generation seems like it might be a win.

However does it actually improve the performance?

Please do performance tests for each iteration otherwise we are just stabbing in the dark! Maybe build a benchmark?

Also did you try my test program above? That will tell us what we stand to gain on your production systems and whether this is worth doing at all.

darthShadow:
Yeah, sorry for the repeated changes without any kinds of benchmarks.

I think I have made the final set of changes (in different commits for easier review) that address the following:

* Use the newer ReadDir for a cleaner implementation and no OS-level segregation.
* Move the mutex lock from the fs to the object.

I will be comparing the results of the 4 binaries (mainline, parallel-stat but fs lock, parallel-stat and object lock, plain golang code that was shared above) on the following types of directory structures over the weekend:

* CephFS kernel mount
* NFS mount
* Single directory with 1M+ files
* Single directory with 1M+ sub-directories and 1-4 files in each
* Single directory with 1M+ sub-directories and no files in each

That should give us definite results and whether there are any improvements (and any potential for further improvements).

ncw:
Sounds like an excellent plan :-) Look forward to the results!

ncw:
You should probably add a local SSD &or HDD in the mix too.

ncw:
Is this ready for review / merge?

darthShadow:
Sorry, I have been busy the last 2 weeks and will be busy for another 2-3 weeks too, so haven't gotten a chance to perform the tests listed above and I don't want to merge the changes without noticeable improvements.

PR #7202, opened 2023-08-02
protondrive: update the information regarding the advance setting enable_caching

henrybear327:

Inform VFS users of a potential issue with the caching mechanism that is currently in place.

#### Checklist

### DIFF
--- a/backend/protondrive/protondrive.go
+++ b/backend/protondrive/protondrive.go
@@ -123,6 +123,10 @@ will be returned, and no upload will happen.`,
 			Name: "enable_caching",
 			Help: `Caches the files and folders metadata to reduce API calls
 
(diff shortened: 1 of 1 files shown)


henrybear327:
@ncw please check if the updated documentation makes sense :) 

The idea is that the current caching mechanism is done based on the idea that when rclone is running sync/copy/etc., it will be the only client doing operations on the protondrive. So basically, it just caches the file/folder nodes and keyrings, unless we are moving or deleting it during the sync operation. 

The event system, which is the polling update mechanism from Proton API, isn't implemented yet, so the cache will likely be holding stale data if protondrive is mounted as a VFS for hours or days.

ncw (approved):
Looks good to me :-)

PR #7216, opened 2023-08-08
lib/rest: remove unnecessary nil check

Juneezee:

Just a pedantic change.

From the Go docs:

> "A `nil` map is equivalent to an empty map. https://go.dev/ref/spec#Map_types

Therefore, an additional nil check for `opts.ExtraHeaders` before the loop is unnecessary because `opts.ExtraHeaders` is a `map`. Example: https://go.dev/play/p/EnnDEF7svOd

No.

#### Checklist


ncw (approved):
Looks fine to me :-)

Thank you

PR #7224, opened 2023-08-12
docker: faster cross building and re-building

qdm12:

Changes only to the Dockerfile for faster buildx cross building and re-building

- Build layer caching for Go module dependencies
- Builder runs on native build platform for faster cross building
- GOARCH and GOARM are set by translating TARGETPLATFORM using [xcputranslate](https://github.com/qdm12/xcputranslate)
- Build instructions shuffled for faster rebuilds and better caching (example `COPY` the binary as the last instruction, not the first in the final image)
- Pin Alpine and Go versions to 3.18 and 1.21 (same Go version as current `latest` Go image) - versions should be pinned to avoid rebuild surprises.
- Go builder is based on Alpine instead of Debian (smaller/faster to download)

You can verify with:

```
docker build --platform linux/arm64 -t rclone/rclone:test .
docker run --rm --platform linux/arm64 --entrypoint /bin/sh rclone/rclone:test -c "apk add file && file /usr/local/bin/rclone"
```

Which prints

```
/usr/local/bin/rclone: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=AQKvJ4exsexfJ3s9jPVd/4X3clScN0bmQYPht3mp0/puBR_Zf40DCVGm3ADfhw/or1CzOk6s1Bk5ARGnJiv, stripped
```

No

#### Checklist


Dockerfile
```diff
@@ -1,24 +1,28 @@
...
-
-# Begin final image
-FROM alpine:latest
-
+ARG ALPINE_VERSION=latest
+ARG GO_VERSION=latest
+ARG XCPUTRANSLATE_VERSION=v0.6.0
+ARG BUILDPLATFORM=linux/amd64
+
+FROM --platform=${BUILDPLATFORM} qmcgaw/xcputranslate:${XCPUTRANSLATE_VERSION} AS xcputranslate
```
qdm12, line 6: FYI this is a small Go program to convert for example `linux/arm/v7` to `GOARCH=arm` + `GOARM=7`

Dockerfile
```diff
@@ -1,24 +1,28 @@
...
+
+FROM --platform=${BUILDPLATFORM} golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
+COPY --from=xcputranslate /xcputranslate /usr/local/bin/xcputranslate
+RUN apk --update add git make bash
+ENV CGO_ENABLED=0
+WORKDIR /tmp/gobuild
```
qdm12, line 12: FYI changed the workdir to `/tmp/gobuild` to highlight it's not required to build from `/go/src/github.com/rclone/rclone/` since we use go modules. I can revert or change to another directory if you prefer.

ncw: That is fine. Building in `/go/src/github.com/rclone/rclone/` is a leftover from pre-modules days.

ncw:
Thanks for doing this :-)

I'm not very experienced with docker builds so I'm going to ask some (probably silly) questions here!

> Changes only to the Dockerfile for faster buildx cross building and re-building
> 
> * Build layer caching for Go module dependencies

This caches the modules in another local docker image - is that right?

> * Builder runs on native build platform for faster cross building

So doesn't require qemu?

> * GOARCH and GOARM are set by translating TARGETPLATFORM using [xcputranslate](https://github.com/qdm12/xcputranslate)
> * Build instructions shuffled for faster rebuilds and better caching (example `COPY` the binary as the last instruction, not the first in the final image)

Good optimization.

> * Pin Alpine and Go versions to 3.18 and 1.19 (go.mod version) - versions should be pinned to avoid rebuild surprises.
> * Go builder is based on Alpine instead of Debian (smaller/faster to download)

OK and OK! Pinning the Go version gives me something to update when we update go versions.

How does this change affect our docker build scripts

- https://github.com/rclone/rclone/blob/master/.github/workflows/build_publish_release_docker_image.yml
- https://github.com/rclone/rclone/blob/master/.github/workflows/build_publish_beta_docker_image.yml

I note I still haven't merged https://github.com/rclone/rclone/pull/6874 which I probably should - what are your opinions on that?

qdm12:
Thanks and happy to knowledge-share!

> This caches the modules in another local docker image - is that right?

It's a layer cached (I believe the same as an untagged image); every instruction (env, run, entrypoint etc.) in the Dockerfile creates a new layer, so there are always multiple layers when building an image. Now the advantages of `COPY go.mod go.sum .` + `RUN go mod download` are:
- modifying the code without changing go.mod or go.sum does not re-trigger a dependencies download on a docker re-build, since go.mod & go.sum are the same
- for cross building, it downloads dependencies once for all platforms, instead of once per target platform. This is because of an implicit yet making-sense detail: from the `ARG TARGETPLAFORM` line, the build parallelizes per target platform. So `make rclone` will run n times, but `go mod download` once for all target platforms. And then `go build` inside the Makefile won't run implicitely modules download since it's already done.

> So doesn't require qemu?

It does require qemu+buildx for the final image build (apk add etc.) since this has to be for the target platform. The Go binary build doesn't need any emulation/qemu.

> Pinning the Go version gives me something to update when we update go versions.

Yes that is an annoyance, but it would prevent some unexpected issues. For example building a Go 1.17 go.mod versioned codebase with go 1.20 would make sha1 not working anymore, and that would break the program without a code change. These are rare breaking changes in the Go language but they do happen from time to time.

An additional step to ensure this is updated is to have a job running a docker build on PRs which we don't have yet, I'll do another PR for this (running on the host machine without cross build should be enough).

> How does this change affect our docker build scripts

I checked those, and surprisingly I don't believe anything is needed. You already used buildx to cross build and inject different target platforms, so nothing to change :+1:

> I note I still haven't merged https://github.com/rclone/rclone/pull/6874 which I probably should - what are your opinions on that?

Good to know, I wanted to do a similar PR after 😄 I'll review it shortly (on my phone right now). But I would say it's independent to this PR. When reviewing it, I'll ask for a review from participants there for this PR here, to double check everything should work as expected 😉

Dockerfile
```diff
@@ -1,24 +1,28 @@
...
-RUN \
-  CGO_ENABLED=0 \
-  make
-RUN ./rclone version
-
-# Begin final image
-FROM alpine:latest
-
+ARG ALPINE_VERSION=3.18
+ARG GO_VERSION=1.19
```
qdm12, line 2: Changing this to 1.21 (as it is on the master branch), to keep from having any funny surprise. It will however need to be bumped once we start using 1.22 language features or bump go.mod to 1.22 at a later time.

```suggestion
ARG GO_VERSION=1.21
```

PR #7240, opened 2023-08-20
oracleobjectstorage: Use ChunkWriter Interfaces

msays2000:

Use the new ChunkWriter interfaces for multi part uploads introduced in https://github.com/rclone/rclone/pull/7154

Yes, in https://github.com/rclone/rclone/pull/7154 and in https://github.com/rclone/rclone/issues/7056

#### Checklist

### DIFF
--- a/backend/oracleobjectstorage/multipart.go
+++ b/backend/oracleobjectstorage/multipart.go
@@ -4,43 +4,390 @@
 package oracleobjectstorage
 
 import (
-	"bytes"
 	"context"
 	"crypto/md5"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"io"
-	"sort"
-	"strconv"
+	"strings"
 	"sync"
+	"time"
+
+	"github.com/ncw/swift/v2"
+	"github.com/rclone/rclone/lib/multipart"
+	"golang.org/x/net/http/httpguts"
 
 	"github.com/oracle/oci-go-sdk/v65/common"
 	"github.com/oracle/oci-go-sdk/v65/objectstorage"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/chunksize"
-	"github.com/rclone/rclone/lib/atexit"
-	"github.com/rclone/rclone/lib/pacer"
-	"github.com/rclone/rclone/lib/readers"
-	"golang.org/x/sync/errgroup"
+	"github.com/rclone/rclone/fs/hash"
 )
 
 var warnStreamUpload sync.Once
 
-func (o *Object) uploadMultipart(
+// Info needed for an upload
+type uploadInfo struct {
+	req       *objectstorage.PutObjectRequest
+	md5sumHex string
+}
+
+type objectChunkWriter struct {
+	chunkSize       int64
+	size            int64
+	f               *Fs
+	bucket          *string
+	key             *string
+	uploadId        *string
+	partsToCommit   []objectstorage.CommitMultipartUploadPartDetails
+	partsToCommitMu sync.Mutex
+	existingParts   map[int]objectstorage.MultipartUploadPartSummary
+	eTag            string
+	md5sMu          sync.Mutex
+	md5s            []byte
+	ui              uploadInfo
+	o               *Object
+}
+
+func (o *Object) uploadMultipart(ctx context.Context, src fs.ObjectInfo, in io.Reader, options ...fs.OpenOption) error {
+	_, err := multipart.UploadMultipart(ctx, src, in, multipart.UploadMultipartOptions{
+		Open:              o.fs,
+		Concurrency:       o.fs.opt.UploadConcurrency,
+		LeavePartsOnError: o.fs.opt.LeavePartsOnError,
+		OpenOptions:       options,
+	})
+	return err
+}
+
+// OpenChunkWriter returns the chunk size and a ChunkWriter
+//
+// Pass in the remote and the src object
+// You can also use options to hint at the desired chunk size
+func (f *Fs) OpenChunkWriter(
 	ctx context.Context,
-	putReq *objectstorage.PutObjectRequest,
-	in io.Reader,
-	src fs.ObjectInfo) (err error) {
-	uploadID, uploadedParts, err := o.createMultipartUpload(ctx, putReq)
+	remote string,
+	src fs.ObjectInfo,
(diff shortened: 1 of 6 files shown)


msays2000:
@ncw I have used the new chunk writer interfaces as you have requested via https://github.com/rclone/rclone/pull/7154.
please review.

backend/oracleobjectstorage/options.go
```diff
@@ -225,16 +221,18 @@ The minimum is 0 and the maximum is 5 GiB.`,
...
-		Default:  memoryPoolFlushTime,
+		Default:  fs.Duration(time.Minute),
+		Hide:     fs.OptionHideBoth,
 		Advanced: true,
 		Help: `How often internal memory buffer pools will be flushed.
 
 Uploads which requires additional buffers (f.e multipart) will use memory pool for allocations.
 This option controls how often unused buffers will be removed from the pool.`,
 	}, {
 		Name:     "memory_pool_use_mmap",
```
Skylion007, line 232: Why doesn't this  rclone's flag `--use-mmap` which already exists?

ncw: It will after this is merged :-)

backend/oracleobjectstorage/options.go
```diff
@@ -225,16 +221,18 @@ The minimum is 0 and the maximum is 5 GiB.`,
 		Advanced: true,
 	}, {
 		Name:     "memory_pool_flush_time",
-		Default:  memoryPoolFlushTime,
+		Default:  fs.Duration(time.Minute),
+		Hide:     fs.OptionHideBoth,
 		Advanced: true,
 		Help: `How often internal memory buffer pools will be flushed.
```
ncw, line 227: In the s3 implementation I deleted the help text here except for the first line and added ` (no longer used)` on the end. Same with the use-mmap flag below

ncw (commented):
Just a couple of quick comments inline.

Can you rebase on the `multithread2` branch please?

I had to change the `ChunkWriter` interface - it needed a ctx parameter on WriteChunk otherwise we can't cancel pending uploads properly.

I fixed the bug in `multithread` too but in a different way so you won't need the `fs/operations/multithread.go` fix.

Once you've rebased I'll give it a more thorough review.

Thank you.

msays2000:
@ncw rebased with multithread2 branch. please review.

backend/oracleobjectstorage/multipart.go
```diff
@@ -104,231 +86,351 @@ func (o *Object) uploadParts(
...
+			return true
+		}
+	}
+	return false
+}
+
+func (w *objectChunkWriter) Abort(_ context.Context) error {
+	fs.Debugf(w.o, "Cancelling multipart upload")
+	err := w.o.fs.abortMultiPartUpload(
+		context.Background(),
```
ncw, line 257: You should be passing the `ctx` passed into `Abort(ctx)` here I think?

backend/oracleobjectstorage/options.go
```diff
@@ -223,19 +219,6 @@ copied in chunks of this size.
 The minimum is 0 and the maximum is 5 GiB.`,
 		Default:  fs.SizeSuffix(maxSizeForCopy),
 		Advanced: true,
-	}, {
```
ncw, line 226: In the s3 changes I left  `memory_pool_flush_time` and `memory_pool_use_mmap` but with `HideBoth` so we don't break existing scripts / config and just ignore these quietly.

ncw (commented):
This looks very nice :-)

I put 2 minor points inline.

Do the integration tests all pass still?

Thank you :-)

backend/oracleobjectstorage/multipart.go
msays2000: yes i thought so too, but if passed in ctx is cancelled, the abort should proceed or not, I guess it should stop as well, so passing in the ctx as you suggested.

backend/oracleobjectstorage/options.go
msays2000: I introduced these flags recently in my previous PR that hasn't been released yet, so we are good wrt oracle backend that no one is using them.

msays2000:
@ncw many tests are failing with `hash type not supported` . 

```
// Hash returns the requested hash of the contents
func (i *StaticObjectInfo) Hash(ctx context.Context, h hash.Type) (string, error) {
	if len(i.hashes) == 0 {
		return "", hash.ErrUnsupported // this is the failure that is being returned. 
	}
	if hash, ok := i.hashes[h]; ok {
		return hash, nil
	}
	return "", hash.ErrUnsupported
}
```

```
=== RUN   TestRcat/withChecksum=true,ignoreChecksum=true
    run.go:180: Remote "oos:bucket rclone-test-jiyujuj8hujeqov0qiwixav0", Local "Local file system at /var/folders/c2/fxthjc2x34q0ztcr8h557h900000gn/T/rclone3100234066", Modify Window "1ms"
    operations_test.go:1740: 
        	Error Trace:	/Users/mkghosh/Code/rs/rclone/fs/operations/operations_test.go:1740
        	            				/Users/mkghosh/Code/rs/rclone/fs/operations/operations_test.go:1751
        	            	multipart upload failed to initialise: failed to prepare upload: hash type not supported
        	Test:       	TestRcat/withChecksum=true,ignoreChecksum=true
```

backend/oracleobjectstorage/options.go
ncw: Great - I didn't spot that - thanks :-)

ncw:
Let me know when you want me to take another look :-)

msays2000:
@ncw please review again. all tests have passed. I have ignored a test case of uploading a stream of 0 length (i guess we are unable to determine its size prior to reading the stream) that results in multipart upload which doesn't accept empty list of parts to commit. A fix will be to fall back to PutObject for such empty stream later. Ok for now as I have seen it fail in TestS3 compatible way as well. Some other s3 providers have it in exception too. 

Screen Shot 2023-08-22 at 9 43 58 AM

ncw (approved):
Thank you this looks great now.

Don't worry about the 0 byte putstream this will never happen in practice as rclone buffers streams less than `--streaming-upload-cutoff` in memory and uploads then normally.

I'll merge this onto the multithread2 branch now and I'll merge that branch to master very soon.

Thank you :-)

ncw:
I've merged this to master now - thank you :-)

PR #7243, opened 2023-08-22
quatrix: add support for Quatrix

oks-maytech:

Add new backend Quatrix.

This change was not discussed before.

#### Checklist


justakit:
@ncw Hello)
`test_all` test suite works perfect 🙂 
image

Also, we have done manual testing of Quatrix backend.

Do we need to take a look at some other points? We followed the instructions of new backend creation in contribution guide, did we miss some checklist there?

vkit-maytech:
@ncw Hello)
`test_all` test suite works perfect 🙂 
image

Also, we have done manual testing of Quatrix backend.

Do we need to take a look at some other points? We followed the instructions of new backend creation in contribution guide, did we miss some checklist there?

vkit-maytech:
@ncw Hello)
`test_all` test suite works perfect 🙂 
image

Also, we have done manual testing of Quatrix backend.

Do we need to take a look at some other points? We followed the instructions of new backend creation in contribution guide, did we miss some checklist there?

ncw:
> Hello) `test_all` test suite works perfect 🙂

Excellent work - well done :-) Not many backend contributions manage that on their first try!

> Also, we have done manual testing of Quatrix backend.
> 
> Do we need to take a look at some other points? We followed the instructions of new backend creation in contribution guide, did we miss some checklist there?

I just wanted to check you'd been following the guide but it looks like you did!

I need an account for integration testing - will a free one work OK?

I'll take a look through the code now.

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+	err := configstruct.Set(m, opt)
+	if err != nil {
+		return nil, err
+	}
+
+	// http client
+	client := fshttp.NewClient(ctx)
+
+	// since transport is a global variable that is initialized only once (due to sync.Once)
+	// we need to reset it to have correct transport per each client (with proper values extracted from rclone config)
```
ncw, line 168: Why do you need one per client?

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+	o = &Object{
+		fs:     f,
+		remote: remote,
+	}
+	return o, leaf, directoryID, nil
+}
+
+// Put the object into the container
+//
+// # Copy the reader in to the new object which is returned
```
ncw, line 423: To avoid `go fmt` putting this `#` in, make sure the previous sentence has a full stop! Or is it this one? Anyway try adding full stops and you can remove the `#`.

vkit-maytech:
Thank you for your feedback 🙂 

> I need an account for integration testing - will a free one work OK?

Our team will create you an account, we just need to pick a Quatrix Plan for you. Should it be a temporary trial for testing or a permanent account? And should we use email from your profile for Quatrix account?

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+
+func (f *Fs) setMTime(ctx context.Context, id string, t time.Time) (result *api.File, err error) {
+	opts := rest.Opts{
+		Method: "POST",
+		Path:   "file/metadata",
+	}
+
+	params := &api.SetMTimeParams{
+		ID:    id,
+		MTime: (api.JSONTime)(t),
```
ncw, line 575: I don't think you need the `()` around `api.JSONTime`?

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+// Optional interface: Only implement this if you have a way of
+// deleting all the files quicker than just running Remove() on the
+// result of List()
+func (f *Fs) Purge(ctx context.Context, dir string) error {
+	return f.purgeCheck(ctx, dir, false)
+}
+
+// Copy src to this remote using server-side copy operations.
+//
+// # This is stored with the remote path given
```
ncw, line 673: See `#` comment above!

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return dstObj, nil
+}
+
+// Move src to this remote using server-side move operations.
+//
+// # This is stored with the remote path given
```
ncw, line 752: `#`

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+	}
+
+	uploadSession, err := o.uploadSession(ctx, directoryID, leaf)
+	if err != nil {
+		return fmt.Errorf("object update: %w", err)
+	}
+
+	o.id = uploadSession.FileID
+
+	defer func() {
```
ncw, line 1068: I'm wondering why you need this?

Does this backend show partially uploaded files? 

If so you might be better setting the `PartialUploads` feature flag.

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+		speed      float64
+		localChunk int64
+	)
+
+	defer o.fs.uploadMemoryManager.Return(o.id)
+
+	for offset := int64(0); offset < size; offset += localChunk {
+		localChunk = o.fs.uploadMemoryManager.Consume(o.id, size-offset, speed)
+
+		buf := make([]byte, localChunk)
```
ncw, line 1096: Rather than making `buf` you'd be better off using `lib/multipart.NewRW()` which you can use as a memory buffer but rclone will recycle the memory and/or use mmap to allocate it.

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+
+		_, err := io.ReadFull(in, buf)
+		if err != nil {
+			deleteErr := o.fs.deleteObject(ctx, o.id)
```
ncw, line 1100: Why delete object?

backend/quatrix/quatrix.go
```diff
@@ -0,0 +1,1255 @@
...
+	}
+
+	return o.fs.uploadLink(ctx, parentID, encName)
+}
+
+func (o *Object) upload(ctx context.Context, uploadKey string, chunk []byte, fullSize int64, offset int64, chunkSize int64, options ...fs.OpenOption) (err error) {
+	opts := rest.Opts{
+		Method:       "POST",
+		RootURL:      fmt.Sprintf(uploadURL, o.fs.opt.Host) + uploadKey,
+		Body:         bytes.NewReader(chunk),
```
ncw, line 1202: `lib/multipart.NewRW()` here too?

docs/content/quatrix.md
```diff
@@ -0,0 +1,238 @@
...
+name> remote
+Type of storage to configure.
+Choose a number from below, or type in your own value
+[snip]
+XX / Quatrix by Maytech
+   \ "quatrix"
+[snip]
+Storage> quatrix
+API key for accessing Quatrix account.
+api_key> jS2pbTDjp9VNlzq3PsSVgyxK2BgE423vcPV.N3CCmqyHiyCM
```
ncw, line 43: I hope this isn't a valid API key?!

I'm guessing it isn't but it might be best to change the string to `your_api_key` or something like that.

docs/content/quatrix.md
```diff
@@ -0,0 +1,238 @@
...
+d) Delete this remote
+y/e/d> y
+```
+
+### Modified time and hashes
+
+Quatrix allows modification times to be set on objects accurate to 1 microsecond.
+These will be used to detect whether objects need syncing or not.
+
+Quatrix does not support SHA1 type hashes, so you cannot use the `--checksum` flag.
```
ncw, line 129: This should say "...does not support hashes..." as there are many more than just SHA1 hashes that rclone supports!

docs/content/quatrix.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Restricted filename characters
+
+File names in Quatrix are case sensitive and have limitations like the maximum length of a filename is 255, and the minimum length is 1. A file name cannot be equal to `.` or `..` nor contain `/` or `\`.
```
ncw, line 133: You are using the standard encoding which encodes stuff like control characters. Is this intentional? Might be worth a mention here.

docs/content/quatrix.md
```diff
@@ -0,0 +1,238 @@
...
+The size of the uploaded chunk will dynamically change depending on the upload speed.
+The total memory use equals the number of transfers multiplied by the minimal chunk size.
+In case there's free memory allocated for the upload (which equals the difference of `maximal_summary_chunk_size` and `minimal_chunk_size` * `transfers`),
+the chunk size may increase in case of high upload speed. As well as it can decrease in case of upload speed problems.
+If no free memory is available, all chunks will equal `minimal_chunk_size`.
+
+### Deleting files
+
+Files you delete with rclone will end up in Trash and be stored there for 30 days.
+Quatrix also provides an API to permanently delete files and an API to empty the Trash so that you can remove files permanently from your account.
```
ncw, line 149: Backends typically provide access to this with a `hard_delete` flag do delete stuff and not put it into the trash. Eg see `b2` backend. Rclone should default to `false` so make versions by default.

I don't think you need to implement this necessarily, but if it is easy (eg add a parameter to the DELETE call) then it would be a nice addition.

ncw (commented):
This looks very good indeed - well done :-)

I put a few comments / queries inline - nothing serious.

Would you like to merge this for v1.64 in a few days if possible?

ncw:
> > I need an account for integration testing - will a free one work OK?
> 
> Our team will create you an account, we just need to pick a Quatrix Plan for you. Should it be a temporary trial for testing or a permanent account? And should we use email from your profile for Quatrix account?

That would be fantastic! We run the integration tests daily so a permanent account would be best. It will only need a tiny quota though. nick@craig-wood.com would be the preferred email.

BTW see the daily integration tests here: https://pub.rclone.org/integration-tests/current/

vkit-maytech:
> Would you like to merge this for v1.64 in a few days if possible?

It would be great! Hopefully we can fix everything quickly

backend/quatrix/quatrix.go
vkit-maytech: Quatrix shows partially uploaded files. We can't set `PartialUploads` for now, because of behavior of Move. It does not replace a file (like a local FS), but returns the error `File exists`. I guess we will change it in a while, as separate improvement.

backend/quatrix/quatrix.go
vkit-maytech: To remove corrupted file. But I've already noticed that it will be deleted in [that](https://github.com/rclone/rclone/pull/7243#discussion_r1315172232) `defer`

backend/quatrix/quatrix.go
vkit-maytech: We use different `fs.ConfigInfo.Headers` for each call to `fs.NewFs` and before this change got the same for each 🙂 
Should it be an option to Quatrix backend?

ncw: I don't see you setting `.Headers` in this code. Is this something you are doing with command line parameters? Or maybe rc calls? It makes sense if you are setting them via the rc and in which case maybe shouldn't be caching the Transport and re-using it, or perhaps only re-using the cached transport if there isn't a custom config in the `ctx`.

If you want to set headers for this backend only you should use `f.srv.SetHeader`

backend/quatrix/quatrix.go
ncw: Ah, OK. Leave it how it is for the moment then.

Do you have a `force` flag or something on the Move API? Rclone generally expects `Move` to overwrite existing files.

docs/content/quatrix.md
ncw: I see you've done this :-)

ncw (commented):
I'm happy to merge this now if you want?

You can send any tweaks/fixes on when they are ready.

backend/quatrix/quatrix.go
vkit-maytech: Not yet, unfortunately. But we plan to add `overwrite` flag to API, like we have for `Copy`

ncw:
I've stuck quatrix on the integration tester - check out the results here tomorrow

https://pub.rclone.org/integration-tests/current/

Would one or other of you like to volunteer to maintain the quatrix backend?

If so please send a PR to adjust https://github.com/rclone/rclone/blob/master/MAINTAINERS.md :-)

You could also be members of the rclone project if you wanted - let me know either here or by email nick@craig-wood.com 

Thank you for a very nicely done backend :-)

PR #7252, opened 2023-08-27
"may not" -> "might not" to remove ambiguity

crantok:
"may not" can be interpreted as "is not allowed".
Replaced with "might not" in both cases to remove this ambiguity.


ncw (approved):
Thank you

PR #7254, opened 2023-08-28
Mounting functionality for latest versions of macOS using NFS

salehqt:

To provide the `mount` command on macOS in cases where FUSE kernel extensions are not available (e.g. enterprise machines)

- [Mount without FUSE on macOS](https://forum.rclone.org/t/mount-without-fuse-on-macos/28826)
- [FR: FUSE-T Support for mounting volumes on Mac](https://forum.rclone.org/t/fr-fuse-t-support-for-mounting-volumes-on-mac/33110/19)

#### Checklist

    - [ ] Need to document under [mount command](https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone_mount.md)
    - [ ] Need to document under [serve command](https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone_serve.md)

Please provide guidance on testing this feature. I am not very familiar with testing in rclone. I manually tested the changes and documented the testing in each commit.


cmd/nfsmount/nfsmount.go
```diff
@@ -0,0 +1,50 @@
...
+	"os/exec"
+	"runtime"
+
+	"github.com/rclone/rclone/cmd/mountlib"
+	"github.com/rclone/rclone/cmd/serve/nfs"
+	"github.com/rclone/rclone/vfs"
+)
+
+func init() {
+	mountlib.NewMountCommand("mount", false, mount)
```
ncw, line 21: We should make this available as `nfsmount` when it isn't being aliased to mount. Check out `cmd/cmount` for how it does it.

cmd/nfsmount/nfsmount.go
```diff
@@ -0,0 +1,50 @@
...
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- s.Serve()
+	}()
+	_, port, err := net.SplitHostPort(s.Addr().String())
+	if err != nil {
+		err = fmt.Errorf("cannot find port number in %s", s.Addr().String())
+		return
+	}
+	err = exec.Command("mount", fmt.Sprintf("-oport=%s,mountport=%s", port, port), "localhost:", mountpoint).Run()
```
ncw, line 39: Does this stand a chance of working on non macOS systems?

I guess in non macOS world then you could `sudo` it.

cmd/nfsmount/nfsmount.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/rclone/rclone/vfs"
+)
+
+func init() {
+	mountlib.NewMountCommand("mount", false, mount)
+	mountlib.AddRc("mount", mount)
+}
+
+func mount(VFS *vfs.VFS, mountpoint string, opt *mountlib.Options) (asyncerrors <-chan error, unmount func() error, err error) {
+	s, err := nfs.NewServer(context.Background(), VFS, &nfs.Options{})
```
ncw, line 26: Does this choose a random port? If not it probably should do.

ncw (commented):
This is a great submission - thank you!

I put some initial comments inline.

It needs some more testing though. If you copy (adjust as needed) `cmd/mount/mount_test.go` into `cmd/nfsmount` can you get `go test -v` to work?

Ideally we'd run some tests on `rclone serve nfs` directly too. I think they would end up effectively being the same as the `cmd/nfsmount` tests unless we implement an `nfs` backend for rclone.

Can you rebase this on master please as I just fixed the linter!

I'll run the CI for this so we can see what happens.

cmd/serve/nfs/filesystem.go
```diff
@@ -0,0 +1,104 @@
...
+	"path"
+	"strings"
+	"time"
+
+	billy "github.com/go-git/go-billy/v5"
+	"github.com/rclone/rclone/vfs"
+)
+
+type ROFS struct {
+	vfs *vfs.VFS
```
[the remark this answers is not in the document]
salehqt: I tried this one. The methods are somewhat compatible, but interface casts are required so the inheritance didn't work correctly.

cmd/serve/nfs/filesystem.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (f *ROFS) Root() string {
+	return f.vfs.Fs().Root()
+}
+
+// Capabilities exports the filesystem as readonly
+func (ROFS) Capabilities() billy.Capability {
+	return billy.WriteCapability | billy.ReadCapability |
+		billy.ReadAndWriteCapability | billy.SeekCapability | billy.TruncateCapability
+}
```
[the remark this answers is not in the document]
salehqt: Added the interface check

cmd/serve/nfs/filesystem.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	billy "github.com/go-git/go-billy/v5"
+	"github.com/rclone/rclone/vfs"
+)
+
+type ROFS struct {
```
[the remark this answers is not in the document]
salehqt: Renamed it to FS

cmd/serve/nfs/filesystem.go
```diff
@@ -0,0 +1,104 @@
...
+func (f *ROFS) Remove(filename string) error {
+	return f.vfs.Remove(filename)
+}
+func (f *ROFS) Join(elem ...string) string {
+	return path.Join(elem...)
+}
+func (f *ROFS) TempFile(dir, prefix string) (billy.File, error) {
+	return nil, os.ErrInvalid
+}
+func (f *ROFS) MkdirAll(filename string, perm os.FileMode) error {
```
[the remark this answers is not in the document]
salehqt: Redirected to MkdirAll instead.

cmd/serve/nfs/server.go
```diff
@@ -0,0 +1,50 @@
...
+	listener net.Listener
+}
+
+func NewServer(ctx context.Context, vfs *vfs.VFS, opt *Options) (s *server, err error) {
+	// Our NFS server doesn't have any authentication, we run it on localhost by default
+	if opt.ListenAddr == "" {
+		opt.ListenAddr = "localhost:"
+	}
+	// NFS doesn't work without CacheModeFull
+	vfs.SetCacheMode(vfscommon.CacheModeFull)
```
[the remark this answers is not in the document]
salehqt: I will remove this. It turned out the problem was with the backend I was testing againt.

cmd/serve/nfs/server.go
```diff
@@ -0,0 +1,50 @@
...
+// server contains everything to run the server
+type server struct {
+	opt      Options
+	handler  nfs.Handler
+	ctx      context.Context // for global config
+	listener net.Listener
+}
+
+func NewServer(ctx context.Context, vfs *vfs.VFS, opt *Options) (s *server, err error) {
+	// Our NFS server doesn't have any authentication, we run it on localhost by default
```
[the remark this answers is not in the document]
salehqt: As far as I know, [NFS before v4 didn't have any reasonable authentication](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/storage_administration_guide/s1-nfs-security#s3-nfs-security-hosts-nfsv4). NFSv4 introduces Kerberos but it is way too complicated and requires enterprise infrastructure. 

go-nfs only implements NFSv3

cmd/nfsmount/nfsmount.go
salehqt: I didn't test it. I would consider that a rarely usable feature since on other Unix platforms FUSE is working pretty well.

cmd/nfsmount/nfsmount.go
salehqt: If the port is not explicitly specified in options, then a random port is chosen automatically by TCP/IP library. in the subsequent lines. the actual port is extracted from server object and is passed to the mount command.

salehqt:
I tried adding a test, but it failed because there is no way to stop the NFS server.

Here is the log:
https://gist.github.com/salehqt/d1dc7fb04094d514548c433062cddcfe

Here is the forever loop that has no escape:
https://github.com/willscott/go-nfs/blob/master/server.go#L61C1-L81C3

cmd/serve/nfs/server.go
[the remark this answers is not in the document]
salehqt: I tested it manually. somehow if the vfs cache is not set to full it doesn't work at all.

ncw:
It looks like go-nfs doesn't compile on these architectures so these need excluding with build tags

-  js/wasm
-  plan9/386
-  plan9/amd64

There are some lint-y things to fix mostly to do with Exported methods needing comments. You can add a comment or unexport the method by making it small initial letter.

    exported: exported method RWFileHandle.Unlock should have comment or be unexported (revive)

These ones should be easy to fix too

      Warning: receiver-naming: receiver name fh should be consistent with previous receiver name f for RWFileHandle (revive)

salehqt:
It looks like there is still a lot of work left on this PR. Aside from fixes for the CI, I managed to run the tests and none of them passed. I tried the actions in the test manually and I can see that they all error out. This requires debugging every single test case and figure out why NFS doesn't behave properly.

In my usage, I only used NFS with CacheModeFull and mostly read operations. I knew at the time that it couldn't handle the entire matrix of operations but it was good enough for my use. I could navigate the remote in Finder and quickly look at all the files.

I have made my code public. So if someone wants to take this and put it over the finish line, they are more than welcome. Otherwise I will slowly work on this PR in the next couple of weeks.

ncw:
> It looks like there is still a lot of work left on this PR. Aside from fixes for the CI, I managed to run the tests and none of them passed.

:-(

It might be that we need to force CacheModeFull for this and I don't think the tests deal with that at the moment.

If you comment this can you get more of the tests to pass?

```patch
--- a/vfs/vfstest/fs.go
+++ b/vfs/vfstest/fs.go
@@ -52,11 +52,11 @@ func RunTests(t *testing.T, useVFS bool, mountFn mountlib.MountFn) {
 		cacheMode vfscommon.CacheMode
 		writeBack time.Duration
 	}{
-		{cacheMode: vfscommon.CacheModeOff},
-		{cacheMode: vfscommon.CacheModeMinimal},
-		{cacheMode: vfscommon.CacheModeWrites},
+		// {cacheMode: vfscommon.CacheModeOff},
+		// {cacheMode: vfscommon.CacheModeMinimal},
+		// {cacheMode: vfscommon.CacheModeWrites},
 		{cacheMode: vfscommon.CacheModeFull},
-		{cacheMode: vfscommon.CacheModeFull, writeBack: 100 * time.Millisecond},
+		// {cacheMode: vfscommon.CacheModeFull, writeBack: 100 * time.Millisecond},
 	}
 	for _, test := range tests {
 		vfsOpt := vfsflags.Opt
```

> I have made my code public. So if someone wants to take this and put it over the finish line, they are more than welcome. Otherwise I will slowly work on this PR in the next couple of weeks.

No hurry! I'll put this on the v1.65 milestone which is in about 2 months probably which will give lots of time for fixing.

ncw:
I don't know why GitHub isn't trying to run the CI on this... Maybe rebase on master, fix the conflicts in go.mod/go.sum and force push to try again?

cmd/nfsmount/nfsmount.go
ncw: Great

vfs/vfstest/fs.go
```diff
@@ -310,7 +313,7 @@ func writeFile(filename string, data []byte, perm os.FileMode) error {
 
 func (r *Run) createFile(t *testing.T, filepath string, contents string) {
 	filepath = r.path(filepath)
-	err := writeFile(filepath, []byte(contents), 0600)
```
ncw, line 313: Why do you need this change? Is it something to do with the NFS server?

vfs/vfstest/fs.go
```diff
@@ -324,7 +327,7 @@ func (r *Run) readFile(t *testing.T, filepath string) string {
 
 func (r *Run) mkdir(t *testing.T, filepath string) {
 	filepath = r.path(filepath)
-	err := r.os.Mkdir(filepath, 0700)
+	err := r.os.Mkdir(filepath, 0755)
```
ncw, line 330: This one too :-) ?

ncw (commented):
This is looking very good now :-)

How do the tests run? Do they run clean now?

At some point I'd like you to clean up the history...

maybe into the following commits

- commit for adding new parameter to vfs.RunTests
- Lock additions to vfs
- commit for adding `nfs serve`
- commit for adding `nfsmount`

salehqt:
I have been debugging at this and pushing fixes. But I didn't get the chance to comment here. I have made all the tests pass except for **TestDirCacheFlush**.

There is a lot to unpack here:

1. I had to fix at least one proper bug in VFS module, with that fixed, most of the tests pass
2. VFS is quite inconsistent when it comes to Unix modes on files. It accepts any mode on Create/Mkdir, but completely ignores it and always returns a fixed mode on stat. While create/mkdir takes any mode, chmod always fails. This is a problem since NFS requires a certain level of consistency. In order to achieve this consintency in the tests, I had to modify the modes in the tests to match what VFS returns for created files and directories (e.g. `0644` and `0755`)
3. Many write operations in VFS require a cache. [see write.go](https://github.com/search?q=repo%3Arclone%2Frclone+%22without+--vfs-cache-mode%22&type=code). Hence I added the minimal required cache option and made NFS server read-only without a cache.
4. Directory caching has multiple problems and I have only managed to make the test pass with a hack:
      1. There is caching at the NFS client side, which can be disabled for the test
      2. NFS relies on timestamps for directories to figure out if it needs to re-read the directory. Not all backends support consistent timestamp on directories. I couldn't even make it pass on **local** backend, which should have consistent timestamps.

I couldn't figure out how attach a debugger to the test because it spawns a separate go process for the server. I even tried printing the PID and using attach, but got other issues. I had to resort to printf debugging which is really slow. 

For debugging TestDirCacheFlush, I have attached the log here: [TestDirCacheFlush test log](https://gist.github.com/salehqt/67b8b049241017855107d4fd0c553ae8). This has all my debug prints to trace what is happening. Basically, NFS doesn't re-read the directory because its timestamp doesn't change when it is invalidated. I might spend more time on this part later. Any help would be appreciated.

Once I have made all the tests pass, I am planning to clean-up history. With a commit specifically for fixing VFS bugs.

salehqt:
This hack makes the test pass:
Screenshot 2023-09-20 at 2 01 59 PM

vfs/vfstest/fs.go
salehqt: When the file is read again, it is always 0644, regardless of the mode that is passed in. NFS tries to do a chmod if the mode is not correct and it would fail.

vfs/vfstest/fs.go
salehqt: Directories are always created as 0755 regardless of the mode that is passed in.

vfs/vfstest/fs.go
```diff
@@ -78,7 +81,7 @@ func RunTests(t *testing.T, useVFS bool, mountFn mountlib.MountFn) {
 			t.Run("TestDirRenameEmptyDir", TestDirRenameEmptyDir)
 			t.Run("TestDirRenameFullDir", TestDirRenameFullDir)
 			t.Run("TestDirModTime", TestDirModTime)
-			t.Run("TestDirCacheFlush", TestDirCacheFlush)
+			// t.Run("TestDirCacheFlush", TestDirCacheFlush)
```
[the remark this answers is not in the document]
salehqt: Done in next commit.

vfs/vfstest/fs.go
```diff
@@ -52,13 +52,16 @@ func RunTests(t *testing.T, useVFS bool, mountFn mountlib.MountFn) {
 		cacheMode vfscommon.CacheMode
 		writeBack time.Duration
 	}{
-		{cacheMode: vfscommon.CacheModeOff},
```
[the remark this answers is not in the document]
salehqt: Done!

ncw:
It sounds like you've been doing great work :-)

> I have been debugging at this and pushing fixes. But I didn't get the chance to comment here. I have made all the tests pass except for **TestDirCacheFlush**.

Well done :-)

> 1. I had to fix at least one proper bug in VFS module, with that fixed, most of the tests pass [7654473](https://github.com/rclone/rclone/commit/7654473f71695255adc6b19f71981dbd319fc700)

Ah. You should probably make a test for that if you can.

> 2. VFS is quite inconsistent when it comes to Unix modes on files. It accepts any mode on Create/Mkdir, but completely ignores it and always returns a fixed mode on stat. While create/mkdir takes any mode, chmod always fails. This is a problem since NFS requires a certain level of consistency. In order to achieve this consintency in the tests, I had to modify the modes in the tests to match what VFS returns for created files and directories (e.g. `0644` and `0755`)

Yes VFS is very sloppy with permissions!

> 3. Many write operations in VFS require a cache. [see write.go](https://github.com/search?q=repo%3Arclone%2Frclone+%22without+--vfs-cache-mode%22&type=code). Hence I added the minimal required cache option and made NFS server read-only without a cache.

OK

> 4. Directory caching has multiple problems and I have only managed to make the test pass with a hack:
>    
>    1. There is caching at the NFS client side, which can be disabled for the test
>    2. NFS relies on timestamps for directories to figure out if it needs to re-read the directory. Not all backends support consistent timestamp on directories. I couldn't even make it pass on **local** backend, which should have consistent timestamps.

This might be a VFS bug I guess. Note the recent change 071c3f28e56c8288aea27f4ea7f4508fa1f82f26 (which I don't know if you have in your tree) that might help too.
 
> I couldn't figure out how attach a debugger to the test because it spawns a separate go process for the server. I even tried printing the PID and using attach, but got other issues. I had to resort to printf debugging which is really slow.

Yes we have to run the mount in a separate process otherwise the Go runtime deadlocks. It took me a very long time to work that out!

> For debugging TestDirCacheFlush, I have attached the log here: [TestDirCacheFlush test log](https://gist.github.com/salehqt/67b8b049241017855107d4fd0c553ae8). This has all my debug prints to trace what is happening. Basically, NFS doesn't re-read the directory because its timestamp doesn't change when it is invalidated. I might spend more time on this part later. Any help would be appreciated.

Is this fixed with your hack? It looks reasonable.

> Once I have made all the tests pass, I am planning to clean-up history. With a commit specifically for fixing VFS bugs.

Great!

Give me a ping when you are ready and I'll give it another review.

salehqt:
Update on **TestDirCacheFlush**: it is not fixable.

NFS has multiple levels of caching and this structure is not compatible with the expectations of the test. NFS uses ctime of directories to figure out if it needs to list them again. But the way that VFS handles ctime is very loose. and in general most backends don't report correct ctime anyways.

I fixed a ctime bug in VFS/dir.go in the latest commit, with the assumption that the backend has consistent ctimes (in this case it does). But it is still not compatible with expectation of the test. ctime of directory get updated when the parent is invalidated. This means that the `dir` is re-read when `otherdir` is invalidated. breaking the test.

My proposal is to remove this test. Because in general, filesystem implementations have their own cache and this test doesn't send cache invalidation commands to the filesystem. Otherwise we can exclude this test for NFS.

ncw:
> Update on **TestDirCacheFlush**: it is not fixable.

Great effort in trying to get it to work :-)

I suggest we detect the NFS file system in the test and skip it. You'll probably have to do that by name or something untidy like that!

```
	if run.fremoteName == "nfswhatever" {
		t.Skip("Skipping test on NFS")
	}
```

cmd/serve/nfs/handler.go
```diff
@@ -0,0 +1,70 @@
...
+
+// BackendAuthHandler returns a NFS backing that exposes a given file system in response to all mount requests.
+type BackendAuthHandler struct {
+	vfs *vfs.VFS
+}
+
+// Mount backs Mount RPC Requests, allowing for access control policies.
+func (h *BackendAuthHandler) Mount(ctx context.Context, conn net.Conn, req nfs.MountRequest) (status nfs.MountStatus, hndl billy.Filesystem, auths []nfs.AuthFlavor) {
+	status = nfs.MountStatusOk
+	hndl = &FS{vfs: h.vfs}
```
lucasew, line 29: Why not use the full name (handle) instead of the abbreviation?

cmd/serve/nfs/handler.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// FromHandle handled by CachingHandler
+func (h *BackendAuthHandler) FromHandle([]byte) (billy.Filesystem, []string, error) {
+	return nil, []string{}, nil
+}
+
+// HandleLimit handled by cachingHandler
+func (h *BackendAuthHandler) HandleLimit() int {
+	return -1
```
lucasew, line 63: What this -1 means?

cmd/serve/nfs/filesystem.go
```diff
@@ -0,0 +1,159 @@
+//go:build unix
```
lucasew, line 1: Is this necessary? Is it possible to run a NFS server on Windows, for example?

cmd/serve/nfs/handler.go
```diff
@@ -0,0 +1,70 @@
...
+func (h *BackendAuthHandler) ToHandle(f billy.Filesystem, s []string) []byte {
+	return []byte{}
+}
+
+// FromHandle handled by CachingHandler
+func (h *BackendAuthHandler) FromHandle([]byte) (billy.Filesystem, []string, error) {
+	return nil, []string{}, nil
+}
+
+// HandleLimit handled by cachingHandler
```
lucasew, line 61: I guess it's a small typo

cachingHandler -> CachingHandler

cmd/serve/nfs/filesystem.go
lucasew: Just cloned `willscott/go-nfs` to find out if there is any reference around it and these two are the only references to the word "build":

image

cmd/serve/nfs/server.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Shutdown stops the server
+func (s *Server) Shutdown() error {
+	return s.listener.Close()
+}
+
+// Serve starts the server
+func (s *Server) Serve() (err error) {
+	fs.Logf(nil, "NFS Server running at %s\n", s.listener.Addr())
+	fmt.Printf("%v", nfs.Serve(s.listener, s.handler))
```
lucasew, line 61: `nfs.Serve` returns an error.

Shouldn't that error be returned in this function?

lucasew:
I'm really hyped for rclone to be able to serve NFS as it may be a little faster than FUSE because less kernel back and forth on Linux. The last time I was really hyped about a rclone feature was when I dropped UMS and minidlnad because rclone got a native, simple, straightforward and overpowered DLNA server.

I am not an expert in the codebase at all so far but I hope I am helping in some way.

ncw:
@salehqt how are you doing with this? Let me know when you want another review. I'd like to merge this for v1.65. (This needs a rebase too).

Can you tidy the history also? Squash into a small number of commits?

Thank you

salehqt:
@ncw, I am now working on the documentation for both serve nfs and nfsmount. Once I have done that. I will rebase and clean-up history.

salehqt:
@lucasew thanks for taking the time to review. There are a couple of features that are beyond this PR and you could help with:
 1. Build and test `serve nfs` on Windows
 2. Add `nfs` as backend using NFS client module in willscot/go-nfs package. This would be very useful for testing `serve nfs`.
 3. Create unit tests for `serve nfs` by using an NFS client library.

salehqt:
@ncw , this PR is ready for review now. I did my best in the documentation, please let me know if you want the wording changed.

I also cleaned up the history into:
 - three diffs for modifications to VFS
 - two diffs for adding `serve nfs` and `nfsmount`
 - one diff for modifying the documentation of mount feature

salehqt:
I ran the workflow on my own repo and it is passing: https://github.com/salehqt/rclone/actions/runs/6410303444

cmd/serve/nfs/filesystem.go
salehqt: I remember I only added the `+build unix` after I saw failures in the CI. This needs to be tested on Windows, but that is beyond this PR.

cmd/serve/nfs/handler.go
salehqt: I have no idea. This was copy-pasted from an example.

cmd/serve/nfs/handler.go
salehqt: I copy pasted this code.

cmd/serve/nfs/server.go
salehqt: Fixed this one in latest push.

cmd/serve/nfs/filesystem.go
ncw: It would be a nice feature for this to work on Windows, but that can be worked on afterwards if you want to, or maybe @lucasew would like to.

ncw (approved):
This looks very good.

Thank you for making a nice set of organised commits - lovely :-)

I will merge this now - thank you!

If you think of anything else then please send a follow up PR.

Do you want to announce this in [on the rclone forum](https://forum.rclone.org/)? It might be good to have a bit more testing before we release it.

PR #7273, opened 2023-09-04
cmd: refactor and use sysdnotify in more commands

### LINKED ISSUE #5117: Refactor systemd notifications for servers
As a follow up from https://github.com/rclone/rclone/pull/5085#issuecomment-789907801 we would like to add systemd notification for all the `rclone serve` commands.

We currently do this for `rclone mount` and `rclone rcd` but it would be great to do this for `rclone serve http` etc.

This is currently done in rcd here

https://github.com/rclone/rclone/blob/d260e3824e15702040db2cecaa49ced852d07d6d/cmd/rcd/rcd.go#L54-L70

I'd suggest we factor that into `lib/systemd` as something like

```go
// Notify systemd that the service is starting. This returns a
// function which should be called to notify that the service is
// stopping. This function will be called on exit if the service exits
// on a signal.
func Notify() func() {
	var (
		finaliseOnce sync.Once
		fnHandle = atexit.Register(finalise)
	)
	
	finalise := func() {
		atexit.Unregister(fnHandle)		
		finaliseOnce.Do(func() {
			_ = sysdnotify.Stopping()
		})
	}

	// Notify ready to systemd
	if err := sysdnotify.Ready(); err != nil {
		log.Printf("failed to notify ready to systemd: %v", err)
	}
	
	return finalise
}
```

Then this can be called just before `s.Wait` in 

- cmd/rcd/rcd.go
- cmd/serve/dlna/dlna.go
- cmd/serve/http/http.go
- cmd/serve/restic/restic.go
- cmd/serve/sftp/sftp.go
- cmd/serve/webdav/webdav.go

Like this

```go
	defer systemd.Notify()()
	s.Wait
```

Or maybe this if you prefer

```go
	finalise := systemd.Notify()
        defer finalise()
	s.Wait
```

It might be that we could use this in `mount` too, but I haven't investigated that.

eNV25:

cmd: refactor and use sysdnotify in more commands

Fixes #5117

#### Checklist

### DIFF
--- a/cmd/rcd/rcd.go
+++ b/cmd/rcd/rcd.go
@@ -4,14 +4,12 @@ package rcd
 import (
 	"context"
 	"log"
-	"sync"
 
-	sysdnotify "github.com/iguanesolutions/go-systemd/v5/notify"
 	"github.com/rclone/rclone/cmd"
 	"github.com/rclone/rclone/fs/rc/rcflags"
 	"github.com/rclone/rclone/fs/rc/rcserver"
-	"github.com/rclone/rclone/lib/atexit"
 	libhttp "github.com/rclone/rclone/lib/http"
+	"github.com/rclone/rclone/lib/systemd"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


ncw (approved):

I'll merge this now and we can have a think about mount and friends.

eNV25:
I forgot to mention I tested with: `systemd-run --user --service-type=notify go run . serve dlna drive:`.

PR #7277, opened 2023-09-05
rc: Add operations/settier to API

drewstinnett:

This change adds `operations/settier` to the API. It behaves in a way similar to the `rclone settier` CLI option.

A couple things I'm not sure how to best handle, would love any pointers/examples on how to make this better:

* Test case currently looks for the SetTier feature, which doesn't appear to be in the test framework, so while there is a test case, I don't think it's being properly evaluated
* In my local testing, even when the remote is set to a file, the settier is being done on the entire remote directory. I'm unsure why this is, or what the best fix is

https://forum.rclone.org/t/using-settier-with-librclone/41461/2

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -2091,6 +2091,20 @@ func SetTier(ctx context.Context, fsrc fs.Fs, tier string) error {
 	})
 }
 
+// SetTierFile changes tier of a single file in remote
+func SetTierFile(ctx context.Context, o fs.Object, tier string) error {
+	do, ok := o.(fs.SetTierer)
+	if !ok {
+		return errors.New("remote object does not implement SetTier")
+	}
+	err := do.SetTier(tier)
+	if err != nil {
+		fs.Errorf(o, "Failed to do SetTier, %v", err)
+		return err
+	}
+	return nil
+}
+
 // TouchDir touches every file in directory with time t
 func TouchDir(ctx context.Context, f fs.Fs, remote string, t time.Time, recursive bool) error {
 	return walk.ListR(ctx, f, remote, false, ConfigMaxDepth(ctx, recursive), walk.ListObjects, func(entries fs.DirEntries) error {
--- a/fs/operations/rc.go
+++ b/fs/operations/rc.go
@@ -209,6 +209,8 @@ func init() {
 		{name: "copyurl", title: "Copy the URL to the object", help: "- url - string, URL to read from\n - autoFilename - boolean, set to true to retrieve destination file name from url\n"},
 		{name: "uploadfile", title: "Upload file using multiform/form-data", help: "- each part in body represents a file to be uploaded\n", needsRequest: true},
 		{name: "cleanup", title: "Remove trashed files in the remote or path", noRemote: true},
+		{name: "settier", title: "Changes storage tier or class on all files in the path", noRemote: true},
+		{name: "settierfile", title: "Changes storage tier or class on the single file pointed to"},
 	} {
 		op := op
 		remote := "- remote - a path within that remote e.g. \"dir\"\n"
@@ -316,6 +318,28 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
 		return nil, nil
 	case "cleanup":
 		return nil, CleanUp(ctx, f)
+	case "settier":
+		if !f.Features().SetTier {
+			return nil, fmt.Errorf("remote %s does not support settier", f.Name())
+		}
+		tier, err := in.GetString("tier")
+		if err != nil {
+			return nil, err
+		}
+		return nil, SetTier(ctx, f, tier)
+	case "settierfile":
+		if !f.Features().SetTier {
+			return nil, fmt.Errorf("remote %s does not support settier", f.Name())
+		}
+		tier, err := in.GetString("tier")
+		if err != nil {
+			return nil, err
+		}
+		o, err := f.NewObject(ctx, remote)
+		if err != nil {
+			return nil, err
+		}
+		return nil, SetTierFile(ctx, o, tier)
 	}
 	panic("unknown rcSingleCommand type")
 }
--- a/fs/operations/rc_test.go
+++ b/fs/operations/rc_test.go
@@ -306,6 +306,61 @@ func TestRcStat(t *testing.T) {
 	})
 }
 
+// operations/settier: Set the storage tier of a fs
+func TestRcSetTier(t *testing.T) {
+	ctx := context.Background()
+	r, call := rcNewRun(t, "operations/settier")
+	if !r.Fremote.Features().SetTier {
+		t.Skip("settier not supported")
+	}
+	file1 := r.WriteObject(context.Background(), "file1", "file1 contents", t1)
+	r.CheckRemoteItems(t, file1)
+
+	// Because we don't know what the current tier options here are, let's
+	// just get the current tier, and re-use that
+	o, err := r.Fremote.NewObject(ctx, file1.Path)
+	require.NoError(t, err)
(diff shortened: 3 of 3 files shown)


drewstinnett:
Testing a bit more, if I copy the `cmd.NewFsSrc` function from the `cmd` package, along with the accompanying functions, then use that function instead of `fs.NewFs` to create the file source, the tier is properly set on a single file, but not on a full directory.

Attempted to try calling `cmd.NewFsSrc` from the operations package generates an import loop error

fs/operations/rc_test.go
```diff
@@ -306,6 +306,26 @@ func TestRcStat(t *testing.T) {
...
 }
 
+// operations/settier: Set the storage tier of a file
+func TestRcSettier(t *testing.T) {
+	// ctx := context.Background()
+	r, call := rcNewRun(t, "operations/settier")
+	file1 := r.WriteObject(context.Background(), "file1", "file1 contents", t1)
+	r.CheckRemoteItems(t, file1)
+
+	if r.Fremote.Features().SetTier {
```
ncw, line 316: This should say something like
```
if !r.Fremote.Features().SetTier {
    t.Skip("settier not supported")
}
```

fs/operations/rc_test.go
```diff
@@ -306,6 +306,26 @@ func TestRcStat(t *testing.T) {
...
+		in := rc.Params{
```
ncw, line 317: Getting this test right is tricky because we don't know the valid values for tiers.

I think I would get an object with `o, err = f.NewObject(ctx, file1.Path)` read the tier from that by using a type assertion `o.(fs.GetTierer)` and then calling `GetTier()` to read the current tier.

I'd then set the object to its current tier.

ncw (commented):
I've put some tips inline.

Note also that this shouldn't include changes to MANUAL* and it doesn't need to include the changes in `docs/content/rc.md` as these will be auto generated as part of the release.

If you want a `settierfile` which sets the tier on a single file rather than a directory of them then this should probably be a different call (see `delete` vs `deletefile`).

fs/operations/rc.go
```diff
@@ -209,6 +209,7 @@ func init() {
 		{name: "copyurl", title: "Copy the URL to the object", help: "- url - string, URL to read from\n - autoFilename - boolean, set to true to retrieve destination file name from url\n"},
 		{name: "uploadfile", title: "Upload file using multiform/form-data", help: "- each part in body represents a file to be uploaded\n", needsRequest: true},
 		{name: "cleanup", title: "Remove trashed files in the remote or path", noRemote: true},
+		{name: "settier", title: "Changes storage tier or class at remote"},
```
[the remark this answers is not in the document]
drewstinnett: Got it, ty!

fs/operations/rc.go
```diff
@@ -316,6 +317,19 @@ func rcSingleCommand(ctx context.Context, in rc.Params, name string, noRemote bo
...
 		return nil, CleanUp(ctx, f)
+	case "settier":
+		if !f.Features().SetTier {
+			return nil, fmt.Errorf("remote %s does not support settier", f.Name())
+		}
+		tier, err := in.GetString("tier")
+		if err != nil {
+			return nil, err
+		}
+		fsrc, err := fs.NewFs(ctx, path.Join(f.Name()+":", remote))
```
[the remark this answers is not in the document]
drewstinnett: Fixed, ty!

fs/operations/rc_test.go
drewstinnett: Fixed, ty!

fs/operations/rc_test.go
drewstinnett: I believe this is doing the correct thing now, but would appreciate a second look if you could 😁

drewstinnett:
Thank you for all the feedback! I think I've put all of these in place now, but any additional feedback is very welcome!

One thing of note, because SetTier takes an fs.Fs object, I needed to make a new SetTierFile function that can take an fs.Object for the single files, similar to the way DeleteFile does. If there's a better way than a completely new function, I'd be happy to implement that, but not sure how to get it in.

Thank you again!

fs/operations/operations.go
```diff
@@ -2091,6 +2091,21 @@ func SetTier(ctx context.Context, fsrc fs.Fs, tier string) error {
...
+func SetTierFile(ctx context.Context, o fs.Object, tier string) error {
+	objImpl, ok := o.(fs.SetTierer)
+	if !ok {
+		fs.Errorf(o, "Remote object does not implement SetTier")
+		return errors.New("remote object does not implement SetTier")
+	}
+	err := objImpl.SetTier(tier)
+	if err != nil {
+		fs.Errorf(o, "Failed to do SetTier, %v", err)
+		return err
```
ncw, line 2104: You don't need to log an Error and report it. Do one or the other :-)

fs/operations/operations.go
```diff
@@ -2091,6 +2091,21 @@ func SetTier(ctx context.Context, fsrc fs.Fs, tier string) error {
 	})
 }
 
+// SetTierFile changes tier of a single file in remote
+func SetTierFile(ctx context.Context, o fs.Object, tier string) error {
+	objImpl, ok := o.(fs.SetTierer)
```
ncw, line 2096: I normally call `objImpl` just `do` then to call it you say `do.SetTier()` which looks neat!

fs/operations/rc_test.go
```diff
@@ -306,6 +306,61 @@ func TestRcStat(t *testing.T) {
...
 }
 
+// operations/settier: Set the storage tier of a fs
+func TestRcSetTier(t *testing.T) {
+	ctx := context.Background()
+	r, call := rcNewRun(t, "operations/settier")
+	file1 := r.WriteObject(context.Background(), "file1", "file1 contents", t1)
+	r.CheckRemoteItems(t, file1)
+
+	if !r.Fremote.Features().SetTier {
```
ncw, line 316: If you put the Skip above the WriteObject then the tests will do a bit less work on unsupported backends.

fs/operations/rc_test.go
```diff
@@ -306,6 +306,61 @@ func TestRcStat(t *testing.T) {
...
+}
+
+// operations/settier: Set the storage tier of a file
+func TestRcSetTierFile(t *testing.T) {
+	ctx := context.Background()
+	r, call := rcNewRun(t, "operations/settierfile")
+	file1 := r.WriteObject(context.Background(), "file1", "file1 contents", t1)
+	r.CheckRemoteItems(t, file1)
+
+	if !r.Fremote.Features().SetTier {
```
ncw, line 343: see above

ncw (commented):
This is looking great now. I think your tests will work just fine. `settierfile` looks good.

I put some minor things inline.

fs/operations/operations.go
drewstinnett: Oops, got it!

fs/operations/operations.go
drewstinnett: Good call, updating

fs/operations/rc_test.go
drewstinnett: Updating this now

fs/operations/rc_test.go
drewstinnett: Updating this as well

drewstinnett:
Made the changes recommended above, thank you for the awesome feedback and product!

ncw (approved):
This looks perfect now :-)

I'll just run the CI and if it passes I'll merge!

ncw:
Thank you :-)

drewstinnett:
Thank YOU!

PR #7281, opened 2023-09-06
local: check if the path is a dir before calling os.Remove()

zjx20:

In the previous implementation, `Rmdir()` removes empty directories by calling `os.Remove()`. However, `os.Remove()` can remove files as well as empty directories (see https://pkg.go.dev/os#Remove). So if the dir argument points to a file, that file will be removed. This behavior is inconsistent with the Linux `rmdir` command, and probably inconsistent with the intended behavior of the `Rmdir()` method.

This PR changes the behavior of `Rmdir()` to determine if dir is a directory before calling `os.Remove`.

#### Checklist


backend/local/local.go
```diff
@@ -642,7 +642,13 @@ func (f *Fs) Mkdir(ctx context.Context, dir string) error {
 //
 // If it isn't empty it will return an error
 func (f *Fs) Rmdir(ctx context.Context, dir string) error {
-	return os.Remove(f.localPath(dir))
+	localPath := f.localPath(dir)
+	if fi, err := os.Stat(localPath); err != nil {
+		return err
+	} else if !fi.IsDir() {
+		return errors.New("is not a dir")
```
ncw, line 649: I think this should be returning `fs.ErrorIsFile`

ncw (commented):
Looks like a good idea - thank you.

There is an error for this already which it would be useful to return - see inline.

backend/local/local.go
zjx20: Updated, good to know.

ncw (approved):
Looks great - thank you :-)

Issue #6310, opened 2022-07-11
lib/http/serve： slice bounds out of range [-90402:]


secsys-go:
#### The associated forum post URL from `https://forum.rclone.org`

None

#### What is the problem you are having with rclone?

I test some API in rclone with [go-fuzz](https://github.com/dvyukov/go-fuzz). 
It shows that some API  is not robust enough

#### What is your rclone version (output from `rclone version`)

I use rclone with commit f065a267f6ca66df6814520f406155087b28bbcc

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Ubuntu 18.04 64bit

#### Which cloud storage system are you using? (e.g. Google Drive)

None 

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
The  poc is here:
```go
package main

import (
	"net/http/httptest"

	"github.com/rclone/rclone/fstest/mockobject"
	"github.com/rclone/rclone/lib/http/serve"
)

func main() {
	w := httptest.NewRecorder()
	r := httptest.NewRequest("GET", "http://example.com/aFile", nil)

	r.Header.Add("Range", "bytes=-90407")
	o := mockobject.New("aFile").WithContent([]byte("hello"), mockobject.SeekModeNone)
	serve.Object(w, r, o)

}
```

and it crashed as:

```
panic: runtime error: slice bounds out of range [-90402:]

goroutine 1 [running]:
github.com/rclone/rclone/fstest/mockobject.(*ContentMockObject).Open(0xc0004223c0, {0xc00041e720, 0x8736e7}, {0xc000404250, 0x1, 0xc00041e7e0})
        /home/zjx/workspace/gowork/src/go-fdg-exmaples/rclone/fstest/mockobject/mockobject.go:153 +0x3e5
github.com/rclone/rclone/lib/http/serve.Object({0x907600, 0xc000456140}, 0xc00046e000, {0x910d98, 0xc0004223c0})
        /home/zjx/workspace/gowork/src/go-fdg-exmaples/rclone/lib/http/serve/serve.go:72 +0xac3
main.main()
        /home/zjx/workspace/gowork/src/go-fdg-exmaples/rclone/lib/http/serve/pocFuzzTestObjectBadRange/main.go:16 +0x1fe

```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

None

#### How to use GitHub

[...]

PR #7283, opened 2023-09-07
Add Fuzz Driver for serve.Object API (Issue  #6310)

secsys-go:
## Overview

This pull request introduces a fuzz driver that addresses issue #6310. The purpose of this fuzz driver is to rigorously test the serve.Object API by injecting mutated data into its arguments. By doing so, we aim to enhance our continuous testing efforts and ensure the robustness of this API.

## Details

The fuzz driver has been designed to thoroughly exercise the serve.Object API.
It accomplishes this by systematically injecting various forms of mutated data into the API's arguments.
Through this approach, we can identify potential vulnerabilities and edge cases that might otherwise go unnoticed.
## Additional Information

We want to highlight that we have already developed numerous fuzz drivers, each serving as a valuable addition to our testing suite. If this PR is well-received, we are eager to contribute more fuzz drivers to further bolster our testing capabilities.

Thus, we are eager for your feedback and guidance on the process of adding fuzz drivers. Specifically, we would appreciate insights on the following:

- Deployment: How should we handle the deployment of these fuzz drivers within the project's testing infrastructure?
- Testing Metrics: Are there specific testing metrics or criteria that you consider essential for evaluating the effectiveness of these fuzz drivers?
- Any Other Concerns: Please let us know if you have any other concerns or preferences related to the inclusion of fuzz drivers in the project.

Thank you for considering this contribution, and we look forward to your feedback.

### DIFF
--- a/lib/http/serve/serve_test_FuzzTestObjectBadRange.go
+++ b/lib/http/serve/serve_test_FuzzTestObjectBadRange.go
@@ -1,0 +1,33 @@
+package serve
+
+import (
+	"net/http/httptest"
+	"os"
+	"secsys/gout-transformation/pkg/transstruct"
+	"testing"
+
+	"github.com/rclone/rclone/fstest/mockobject"
+)
+
+func FuzzTestObjectBadRange(XVl []byte) int {
+	t := &testing.T{}
+	_ = t
+	var skippingTableDriven bool
+	_, skippingTableDriven = os.LookupEnv("SKIPPING_TABLE_DRIVEN")
+	_ = skippingTableDriven
+	transstruct.SetFuzzData(XVl)
+	FDG_FuzzGlobal()
+
+	w := httptest.NewRecorder()
+	r := httptest.NewRequest("GET", "http://example.com/aFile", nil)
+	r.Header.Add("Range", "xxxbytes=3-5")
+	o := mockobject.New(transstruct.GetString("aFile")).WithContent([]byte(transstruct.GetString("0123456789")), mockobject.SeekModeNone)
+	Object(w, r, o)
+	_ = w.Result()
+
+	return 1
+}
+
+func FDG_FuzzGlobal() {
+
+}


lib/http/serve/serve_test_FuzzTestObjectBadRange.go
```diff
@@ -0,0 +1,33 @@
+package serve
+
+import (
+	"net/http/httptest"
+	"os"
+	"secsys/gout-transformation/pkg/transstruct"
```
ncw, line 6: Where does this package come from?

ncw (commented):
Fuzz testing sounds great.

We could integrate it as a github actions run. I'm not sure we'd want to run it every commit - not sure how long we'd run it for, but it could be a scheduled thing maybe?

Your code seems to use a private library though - what is the transstruct library and were does it come from?

lib/http/serve/serve_test_FuzzTestObjectBadRange.go
thaJeztah: I'm curious as well (arrived here, after we received a similar pull request in our repository https://github.com/moby/moby/pull/46421). I'm not sure if these PRs are "legit", also looking at the author's profile activity

Screenshot 2023-09-07 at 16 08 09

ncw: Hmm, curious. The PR itself looks like code but I have no idea how it might work!

The `{insert repo here}` PRs are odd too.

I'm hoping we might hear from the author to explain.

thaJeztah: Yup! So "Fuzzing" code (to what I've seen) usually looks a bit odd (on purpose).

I'm a big "+1" on having more fuzzing on projects; just the PRs from this contributor looked (at first glance) "dodgy", so I thought let me at least comment!

If you're interested in Fuzzing, @AdamKorcz has been helping various projects to set up (and can be trusted). We're still not doing as much fuzzing in Moby as would be desired, but that's really "us" to blame (time and all that)

thaJeztah: @ncw we discussed the contribution in our maintainers call, and after looking, we considered that these are _NOT_ legit contributions;

- golang package names MUST have a `.` (domain name) in the first element, so the `secsys/gout-transformation/pkg/transstruct` is invalid
- The fuzzing functions do _not_ match the correct signature
- The fuzzing functions must be in a `_test.go` file, and both contributions do _not_ (and would break production code)

I'm reporting the account for abuse, but you could decide to already ban them from your org.

ncw:
I'm going to close this as it really doesn't seem like it could ever work and we haven't heard back from the submitter.

secsys-go:
We are sorry to submit this unformed PR. The PR's background is that we are now developing a new technique to automatically transform official unit tests into fuzz drivers. We are so sorry that we didn't review carefully in this PR so that there are some errors in code. We have fixed these errors, and this version #7315 can work well in OSS-Fuzz. We are appreciated if anyone could point out possible errors we have not found in this version #7315. Suggestions are welcome too. Thanks.

PR #7286, opened 2023-09-10
Protondrive: introduce kill-switch 

henrybear327:

Introduce a kill-switch for the minimum allowed version for the proton bridge library. 

No. 

#### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,8 +28,6 @@ require (
 	github.com/go-chi/chi/v5 v5.0.10
 	github.com/google/uuid v1.3.1
 	github.com/hanwen/go-fuse/v2 v2.3.0
-	github.com/henrybear327/Proton-API-Bridge v0.0.0-20230908065933-5bfa15b567db
-	github.com/henrybear327/go-proton-api v0.0.0-20230907193451-e563407504ce
 	github.com/hirochachacha/go-smb2 v1.1.0
 	github.com/iguanesolutions/go-systemd/v5 v5.1.1
 	github.com/jcmturner/gokrb5/v8 v8.4.4
@@ -115,6 +113,8 @@ require (
 	github.com/hashicorp/errwrap v1.0.0 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
+	github.com/henrybear327/Proton-API-Bridge v0.0.0-20230910175313-f6671447d051 // indirect
+	github.com/henrybear327/go-proton-api v0.0.0-20230907193451-e563407504ce // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jcmturner/aescts/v2 v2.0.0 // indirect
 	github.com/jcmturner/dnsutils/v2 v2.0.0 // indirect
--- a/backend/protondrive/protondrive.go
+++ b/backend/protondrive/protondrive.go
@@ -3,13 +3,16 @@ package protondrive
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"path"
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	protonDriveAPI "github.com/henrybear327/Proton-API-Bridge"
 	"github.com/henrybear327/go-proton-api"
 
@@ -18,6 +21,7 @@ import (
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/config/obscure"
+	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/lib/dircache"
 	"github.com/rclone/rclone/lib/encoder"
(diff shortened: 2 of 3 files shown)


henrybear327:
Sorry that I messed up the pull request title. 

The PR will completed in full later today. It’s put up for tracking as the release is happening soon.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rclone/rclone
 
-go 1.19
```
ncw, line 3: Please don't change the go version! It will break a huge amount of stuff.

henrybear327: Won't do! I will make sure this is removed on the final version!

henrybear327:
Hello @ncw, 

So the code will work like this:
- During ProtonDrive init, it will fetch from 
- It will use the semver package to parse the bridging library version and the minimum allowed version

If the bridge version is less than the allowed version, it will return something like
```
2023/09/10 19:44:44 Failed to create file system for "TestProtonDrive:": the current rclone version is using a disallowed version of the Proton Bridge API library. Please upgrade your rclone version
```

If other type of error occurs, such as Github Static Page being down (since I am hosting the API response up there, which should be reliable enough), we would be safe than sorry, so the protondrive will also not work in this case.

Otherwise, the code will work as-is.

---

I am adding this feature because we are reverse engineering the code, and, in the worst case, if the encryption is done wrongly, or some other types of decryption has having serious enough flaw, I believe having this mechanism is better than just letting the code running in the wild.

Proton API also have a similar mechanism, but it's based on the appversion string, which we are "borrowing" from another client (macos for now), so I decided to implement a simple one for our own control.

go.mod
henrybear327: I am having issues trying to just update the dependencies by running `go get -u github.com/henrybear327/Proton-API-Bridge@master`. 

This part will be automatically added to the `go.mod` file
```
go 1.21

toolchain go1.21.1
```. 

Trying to resolve this issue now :(

henrybear327: Resolved by 
- `brew install go@1.19`
- manually removing the dependencies `Proton-API-Bridge` and `proton-go-api` from both `go.mod` and `go.sum` files
- run `go get github.com/henrybear327/Proton-API-Bridge@master` to reinstall the package

henrybear327:
@ncw please review when you have time :) 

I would really like to have this into the next release, since it's a feature that enables us to perform damage control in the end.

go.mod
ncw: Don't use `-u` with `go get` - it is almost always the wrong thing to do! It is a leftover from the early days of go.

henrybear327: Thanks! Will put it on my notes :(

ncw:
> So the code will work like this:
> 
> * During ProtonDrive init, it will fetch from https://github.com/henrybear327/Proton-API-Bridge-Config/blob/main/config.json by accessing https://henrybear327.github.io/Proton-API-Bridge-Config/test_config.json
> * It will use the semver package to parse the bridging library version and the minimum allowed version
> 
> If the bridge version is less than the allowed version, it will return something like
> 
> ```
> 2023/09/10 19:58:42 Failed to create file system for "TestProtonDrive:": the current rclone version is using a disallowed version of the Proton Bridge API library. Please upgrade your rclone version
> ```
> 
> If other type of error occurs, such as Github Static Page being down (since I am hosting the API response up there, which should be reliable enough), we would be safe than sorry, so the protondrive will also not work in this case. The error will be something like this:
> 
> ```
> 2023/09/10 19:57:48 Failed to create file system for "TestProtonDrive:": allowed Proton API Bridge version check error: 404 Not Found
> ```
> 
> Otherwise, the code will work as-is.

I'm not keen on this implementation for several reasons.

1. Adding another HTTP request into the backend initialisation shouldn't be done lightly. Some people run strings of rclone command lines one after another and this adds another HTTP call into each one.
2. Calling an HTTP API on a third-party site could be seen as a privacy violation which would allow logging of all the protondrive users (maybe not by us but GitHub could certainly do it).
3. Failing when GitHub isn't available is bad also. Users may not be able to access GitHub for some reason (think firewalls, networking problems, administrative decisions) so making the backend dependent on that is a bad idea.
 
> I am adding this feature because we are reverse engineering the code, and, in the worst case, if the encryption is done wrongly, or some other types of decryption has having serious enough flaw, I believe having this mechanism is better than just letting the code running in the wild.

Rclone checks sha1s of files it downloads. Is that enough protection?

> Proton API also have a similar mechanism, but it's based on the appversion string, which we are "borrowing" from another client (macos for now), so I decided to implement a simple one for our own control.

Can we use a proton API here instead? That would get round the problems above?

henrybear327:
> > So the code will work like this:
> > 
> > * During ProtonDrive init, it will fetch from https://github.com/henrybear327/Proton-API-Bridge-Config/blob/main/config.json by accessing https://henrybear327.github.io/Proton-API-Bridge-Config/test_config.json
> > * It will use the semver package to parse the bridging library version and the minimum allowed version
> > 
> > If the bridge version is less than the allowed version, it will return something like
> > ```
> > 2023/09/10 19:58:42 Failed to create file system for "TestProtonDrive:": the current rclone version is using a disallowed version of the Proton Bridge API library. Please upgrade your rclone version
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If other type of error occurs, such as Github Static Page being down (since I am hosting the API response up there, which should be reliable enough), we would be safe than sorry, so the protondrive will also not work in this case. The error will be something like this:
> > ```
> > 2023/09/10 19:57:48 Failed to create file system for "TestProtonDrive:": allowed Proton API Bridge version check error: 404 Not Found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise, the code will work as-is.
> 
> I'm not keen on this implementation for several reasons.
> 
> 1. Adding another HTTP request into the backend initialisation shouldn't be done lightly. Some people run strings of rclone command lines one after another and this adds another HTTP call into each one.
> 2. Calling an HTTP API on a third-party site could be seen as a privacy violation which would allow logging of all the protondrive users (maybe not by us but GitHub could certainly do it).
> 3. Failing when GitHub isn't available is bad also. Users may not be able to access GitHub for some reason (think firewalls, networking problems, administrative decisions) so making the backend dependent on that is a bad idea.
> 
> > I am adding this feature because we are reverse engineering the code, and, in the worst case, if the encryption is done wrongly, or some other types of decryption has having serious enough flaw, I believe having this mechanism is better than just letting the code running in the wild.
> 
> Rclone checks sha1s of files it downloads. Is that enough protection?
> 
> > Proton API also have a similar mechanism, but it's based on the appversion string, which we are "borrowing" from another client (macos for now), so I decided to implement a simple one for our own control.
> 
> Can we use a proton API here instead? That would get round the problems above?

I don’t think it’s possible to use proton api solution since it’s something that needs to be coordinated with them. 

I agree that this is a suboptimal approach, but I can’t think of a better way to implement the kill-switch. 

Let’s proceed with not merging then!

ncw:
Ok great. If we find a really important problem we can issue a point release fix. We've also marked the backend as beta for the moment.

henrybear327:
> Ok great. If we find a really important problem we can issue a point release fix. We've also marked the backend as beta for the moment.

Thanks @ncw!

Issue #4110, opened 2020-04-05
VFS Cache File-Level Size Options


trumblejoe:
#### What is your current rclone version (output from `rclone version`)?
v1.51

#### What problem are you are trying to solve?
File-level encryption w/ file-name encryption creates small, must-access files on a per-directory basis. These files will often be rate-limited by the underlying providers resulting in errors and temporarily, and even potentially, service bans.

#### How do you think rclone should be changed to solve that?
By implementing options for controlling the --vfs-cache-file-max-size and vfs-cache-file-min-size, you would be able to create a range of files which would be unaffected by vfs-cache-mode full.  These unaffected files should default back to vfs cache-mode writes keeping in-line with the hierarchy structure.

Example File-Sizes :
25k, 1g
5g, 100g

File-max-size 26k.
Caches : 25k
Un-cached : 1g, 5g, 100g

File-min-size 2.5g
Caches : 5g, 100g
Un-cached : 25k, 1g

File-max-size 26k
File-min-size 6g
Caches : 25k, 100g
Un-cached : 1g,5g

Thoughts? Is the get file size function fast-enough to support these operations on-the-fly?

ncw:
I think this could certainly be done.

I'm unsure how it would help in your scenario though? For `--vfs-cache-mode full` the small files are downloaded to disk. If you exclude them then the small files will be downloaded on the fly. Can you explain further?

trumblejoe:
In my case, I'm not interested in excluding the small files actually, only in excluding larger files from monopolizing disk space. I actually don't run vfs-cache-mode full over writes because with my use-case my local storage fills up far too quickly with cache-mode full. I see this request as a way to make cache-mode full more useful than what it already is.

darthShadow:
@ncw You did add this in https://forum.rclone.org/t/rclone-vfs-cache-maximum-size-per-file/30037/6 under the https://github.com/rclone/rclone/tree/fix-vfscache-strategy branch.

Would it be possible to add this to the mainline too? Any help that you require that's preventing it's merge?

ncw:
I had forgotten about that code!

I rebased it here if anyone wants to try it

[...]

PR #7290, opened 2023-09-12
vfs: add --vfs-cache-strategy to configure cache eviction strategy

elldw:

Adds the --vfs-cache-strategy flag, tests, and documentation to support the lru-sp cache strategy

https://github.com/rclone/rclone/issues/4110
https://forum.rclone.org/t/rclone-vfs-cache-maximum-size-per-file/30037

#### Checklist


ncw:
I see you've taken the ball and run with it - fantastic :-)

What extra do you think needs to be done to this?

Note that I made a generic Enum config type for rclone recently which you could use for the cache-mode strategy

https://github.com/rclone/rclone/blob/3ab9077820420d481463d828ae2d44ce3e3077ce/vfs/vfscommon/cachemode.go#L8-L33

I am considering that we should probably keep the score of everything up to date continuously so we can evict things from the cache in a more timely fashion. That is probably a job for another day though!

elldw:
If you don't think any other strategies are needed right now, then I suppose the main thing left is just updating the documentation. I'll go ahead and update cachestrategy to use the new Enum type as well.

ncw:
Can you squash this into one commit and unset the draft flag when you are ready and I'll give it a final review.

If you make that final commit by you but put a `Co-authored-by: Nick Craig-Wood <nick@craig-wood.com>` at the end to acknowledge my contribution that would be perfect!

Thank you :-)

Issue #4687, opened 2020-10-15
Copying file from s3 to s3 results in empty md5 checksum if source has empty checksum


ThiefMaster:
#### What is the problem you are having with rclone?

When copying a file from a s3 bucket to another s3 bucket and the source file has no (or an empty, not sure) checksum, then the target file will have no checksum either. If I manually copy the file to my local system first and then copy it to the target bucket, the checksum is present.

I would expect the checksum to be set in the first case as well. FWIW, I do not know why the checksum is missing on my source bucket. In case it matters, the file was copied there from a normal file system (AFS) with rclone as well not too long ago.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.53.0-DEV
- os/arch: linux/amd64
- go version: go1.14.7
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)

Gentoo, 64-bit

####  Which cloud storage system are you using? (eg Google Drive)

Ceph S3

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
[adrian@blackhole:~/dev/indico/src:master $]> rclone hashsum MD5 s3-backup-prevessin:adrian-test-2/event/926804/event/3550126-3306548-2020_VS_PPFS_Ye.mp4
                                  3550126-3306548-2020_VS_PPFS_Ye.mp4

[adrian@blackhole:~/dev/indico/src:master $]> rclone copyto -vvP s3-backup-prevessin:adrian-test-2/event/926804/event/3550126-3306548-2020_VS_PPFS_Ye.mp4 s3-backup-prevessin:adrian-test-manual/md5test-direct.mp4
2020/10/15 23:08:14 DEBUG : rclone: Version "v1.53.0-DEV" starting with parameters ["rclone" "copyto" "-vvP" "s3-backup-prevessin:adrian-test-2/event/926804/event/3550126-3306548-2020_VS_PPFS_Ye.mp4" "s3-backup-prevessin:adrian-test-manual/md5test-direct.mp4"]
2020/10/15 23:08:14 DEBUG : Using config file from "/home/adrian/.config/rclone/rclone.conf"
2020/10/15 23:08:14 DEBUG : Creating backend with remote "s3-backup-prevessin:adrian-test-2/event/926804/event/3550126-3306548-2020_VS_PPFS_Ye.mp4"
[...]

PR #7292, opened 2023-09-12
s3: fix missing etag on multipart objects in s3

hokie228:
In S3, multipart objects return an ETag value with the number of parts as a suffix, "somehash-10".  The regex to validate the ETag does not take this into account, causing the ETag value to be empty when listing objects from an S3 bucket.  This fix updates the regex to optionally validate the part count suffix.

I was using the clone code embedded in a prototype application to list files from S3, and noticed that the ETag value was not being returned.  It appears that the regex used to validate the md5 hash does not take into account the part count suffix that S3 uses for objects uploaded using the multipart API.  This fix just updates the regex to validate standard and multipart md5 / ETag values returned by S3.

I have not found this exact issue, but I believe it relates to issue #4687 

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -5037,7 +5037,7 @@ func (o *Object) Remote() string {
 	return o.remote
 }
 
-var matchMd5 = regexp.MustCompile(`^[0-9a-f]{32}$`)
+var matchMd5 = regexp.MustCompile(`^[0-9a-f]{32}(-[0-9]+)?$`)
 
 // Set the MD5 from the etag
 func (o *Object) setMD5FromEtag(etag string) {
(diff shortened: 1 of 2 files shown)


ncw:
I'm pretty sure that will break the md5sums for S3 objects.

Can you run the backend integration tests with your patch?

    cd backend/s3
    go test -v -remote yours3:

hokie228:
You are correct, it does indeed fail the integration tests.  I see that in the contribution notes now, I apologize for missing that the first time.

Given that the code is deliberately ignoring the AWS ETag value for multipart files, is it safe to assume that this is the desired / expected behavior then?

ncw:
Rclone only uses the `ETag` if it represents a valid MD5SUM - it isn't interested in any of the other forms of Etag (eg from multipart uploads, or SSE).

So, yes this is intended behaviour as Etags of the form XXXX-N aren't the MD5 sum of the object.

Rclone does validate these when uploading objects but it is very hard to validate them afterwards.

hokie228:
Fair enough...thanks for your time.

PR #7307, opened 2023-09-14
box: polling tweaks - add more logging and remember more EventIDs

dsze:

Add more logging to the polling code (originally PR #7248), in an attempt to better understand the occasional integration test failures. It's likely that either there are some events being incorrectly filtered in the code, or the Box Event API is slow and doesn't send all events before the test ends. Either way, this additional logging should help us to understand the issue.

Also expand the scope of the EventID filter so that IDs up to 1 minute old will be remembered. This reduces the number of duplicate notifications, since the Box Event API seems to repeat the same events up to 10s later.

No

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -1211,6 +1211,10 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 			fs.Infof(f, "Failed to get StreamPosition: %s", err)
 		}
 
+		// box can send duplicate Event IDs. Use this map to track and filter
+		// the ones we've alread processed.
+		processedEventIDs := make(map[string]time.Time)
+
 		var ticker *time.Ticker
 		var tickerC <-chan time.Time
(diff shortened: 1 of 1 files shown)


ncw (approved):
Look great - thank you.

Will merge now.

brimur:
Hi guys, I only just discovered polling was added for Box but I cannot get it working. I opened a topic here... https://forum.rclone.org/t/can-i-add-directory-updates-to-the-rclone-log/42818/6 

I dont think it is working or I am using it wrong. I added --poll-interval 60s and --dir-cache-time 9999h on both my servers and rebooted. After a reboot I started a tail -f rclone.log. On one server I went to the root directory of my mount and created a directory called myDir and a file called myFile.txt. Then two more files. On the other server I am repeatedly doing an 'ls' command but the new files are not showing up.

I also see no mention of polling in the DEBUG log such as "Checking for changes on remote"  or changeNotify

PR #7319, opened 2023-09-18
vfs: add exclusions for vfs write back - fixes #2109

### LINKED ISSUE #2109: Stop rclone uploading specific directories and files
Issue 1: rclone uploads ~partial files (which are not ready yet) and I want rclone mount to "ignore" these files and NOT uploading them to the cloud
Issue 2: I want rclone to ignore whole directories to NOT upload to it's mount. ie, .grab directories that Plex DVR creates.

Thanks!

utkarshdalal:
#### Checklist

### DIFF
--- a/vfs/vfs.go
+++ b/vfs/vfs.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"reflect"
 	"sort"
 	"strings"
 	"sync"
@@ -200,7 +201,7 @@ func New(f fs.Fs, opt *vfscommon.Options) *VFS {
 	defer activeMu.Unlock()
 	configName := fs.ConfigString(f)
 	for _, activeVFS := range active[configName] {
-		if vfs.Opt == activeVFS.Opt {
+		if reflect.DeepEqual(vfs.Opt, activeVFS.Opt) {
 			fs.Debugf(f, "Re-using VFS from active cache")
 			activeVFS.inUse.Add(1)
 			return activeVFS
--- a/vfs/vfscache/item.go
+++ b/vfs/vfscache/item.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"sync"
 	"time"
 
@@ -629,6 +630,15 @@ func (item *Item) store(ctx context.Context, storeFn StoreFn) (err error) {
 	return item._store(ctx, storeFn)
 }
 
+func matches_exclusion_patterns(filename string, vfsExcludeRegex []*regexp.Regexp) bool {
+	for _, regexPattern := range vfsExcludeRegex {
+		if regexPattern.MatchString(filename) {
+			return true
+		}
+	}
+	return false
+}
+
 // Close the cache file
 func (item *Item) Close(storeFn StoreFn) (err error) {
 	// defer log.Trace(item.o, "Item.Close")("err=%v", &err)
@@ -715,6 +725,10 @@ func (item *Item) Close(storeFn StoreFn) (err error) {
 
 	// upload the file to backing store if changed
 	if item.info.Dirty {
+		if matches_exclusion_patterns(item.name, item.c.opt.VfsExcludeRegex) {
+			fs.Infof(item.name, "vfs cache: skipping writeback due to exclusion pattern match")
+			return
+		}
 		fs.Infof(item.name, "vfs cache: queuing for upload in %v", item.c.opt.WriteBack)
 		if syncWriteBack {
 			// do synchronous writeback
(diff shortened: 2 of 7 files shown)


vfs/vfs.go
```diff
@@ -200,7 +201,7 @@ func New(f fs.Fs, opt *vfscommon.Options) *VFS {
 	defer activeMu.Unlock()
 	configName := fs.ConfigString(f)
 	for _, activeVFS := range active[configName] {
-		if vfs.Opt == activeVFS.Opt {
+		if reflect.DeepEqual(vfs.Opt, activeVFS.Opt) {
```
utkarshdalal, line 204: I had to do this because it couldn't do a shallow compare of the struct with a string array as a variable (we've added `VfsUploadExclude` as a new string array to `Options`)

vfs/vfscache/item.go
```diff
@@ -629,6 +630,15 @@ func (item *Item) store(ctx context.Context, storeFn StoreFn) (err error) {
 	return item._store(ctx, storeFn)
 }
 
+func matches_exclusion_patterns(filename string, vfsExcludeRegex []*regexp.Regexp) bool {
```
ncw, line 633: Go naming scheme is `matchesExclusionPatterns` for non exported file

ncw (commented):
Thanks for doing this - looks very useful :-)

Can you drop the auto generated docs from this commit? Eg MANUAL* rclone.1 and rclone_mount.md

Though this maybe deserves a note in the vfs help as to what it does with a link to the filtering syntax.

This needs tests also.

I'll run the CI now.

utkarshdalal:
Thanks. Could you give me a pointer or a reference for the tests? This is
actually my first time working with Go and Rclone and I have no idea where
to start.

Also, I'll add documentation to the vfs.md file and drop it from the auto
generated ones.

utkarshdalal:
@ncw I've made the changes you requested. If it's looking ok I can squash the commits? Or you could also do a squash and merge.

utkarshdalal:
@ncw found a potential issue here with the use case I have in mind. If I write a file to a --vfs-upload-exclude path (`downloading/**`), and then move/rename that same file to another folder, it now never gets uploaded as well. Any idea if this is a bug, and if so how to fix it?

utkarshdalal:
pushed a fix for this, it uses `item.reload()`. It's a little ugly since the if condition is repeated. Maybe on the item cache we can add a flag for whether it's been excluded, if so, update `writeback.Rename` to force writeback. Thoughts? Would love your guidance on this.

utkarshdalal:
I tested it out, and while it correctly writes the files to the backend, something is up (perhaps the cache is not cleared) and I can't run Windows applications that depend on the files "natively". When I then close and re-open the mount, it works fine.

vfs/vfscache/item.go
```diff
@@ -1426,5 +1440,9 @@ func (item *Item) rename(name string, newName string, newObj fs.Object) (err err
 		_ = downloaders.Close(nil)
 	}
 	item.c.writeback.Rename(id, newName)
+	if matchesExclusionPatterns(name, item.c.opt.VfsExcludeRegex) {
+		fs.Infof(item.name, "vfs cache: performing writeback due to file rename")
+		item.reload(context.Background())
```
ncw, line 1445: Please check this error

vfs/vfscache/item_test.go
```diff
@@ -118,8 +131,56 @@ func TestItemDirty(t *testing.T) {
 	checkObject(t, r, "potato", "hello")
 }
 
+func TestItemExcludedDirty(t *testing.T) {
+	customOptions := &vfscommon.Options{
+		VfsUploadExclude: []string{"exclude_folder/**", "**.json"}, // Replace with your actual patterns
```
ncw, line 136: `*.json` should work here instead of `**.json`

ncw (commented):
I think calling `reload()` is just about OK. It does some things we don't need though and maybe that is the problem. The `NewObject` call can be expensive. However I suggest we leave it as-is for the moment.

Can you fix the lint errors please?

Please check all errors. In the tests use `require.NoError(t, err)`

If you fix up what you've done so far, then I'll pull it locally and try to debug what is going on.

Can you describe exactly what is the problem in a way that I can replicate it please? Preferably using the command line or python. I'll have a go at fixing it then.

The VFS is very complicated and it has a lot of moving parts!

utkarshdalal:
@ncw just fixed the linting errors. Sorry, I just realised I can run the actions on my fork as well.

I'm not sure exactly how to describe it, because I don't think anything is "wrong" in the execution. But for some background, I'm using rclone to install Steam games (mounting a directory, so the games are directly written to S3). After my changes, the files get saved and uploaded correctly, but when I try to run the game it says files are missing.

When I restart the mount, it works fine.

utkarshdalal:
@ncw think this is ok to merge now, please take a quick look when you can :)

ncw:
Looks like I took my eye off the ball for this useful PR - many apologies :-(

This PR needs rebasing (which is a non-trivial amount of work as the config system has changed and there have been many changes to the VFS) - do you want to do that @utkarshdalal or @feixm1 could do it?

I guess I could do it too if I can squeeze it in somewhere!

PR #7333, opened 2023-09-23
Update documentation for --fast-list adding info about ListR

albertony:

The existing documentation for `--fast-list` has no mention of `ListR` feature. Also its a bit confusing that rclone will use the `ListR` feature by default when doing list commands, while the `--fast-list` is to enable it for other commands. This is if I have understood it correctly - because I'm a bit confused myself...!?

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -2121,34 +2121,50 @@ there were IO errors`.
 ### --fast-list ###
 
 When doing anything which involves a directory listing (e.g. `sync`,
-`copy`, `ls` - in fact nearly every command), rclone normally lists a
-directory and processes it before using more directory lists to
-process any subdirectories.  This can be parallelised and works very
-quickly using the least amount of memory.
+`copy`, `ls` - in fact nearly every command), rclone has different
+strategies to choose from.
 
-However, some remotes have a way of listing all files beneath a
-directory in one (or a small number) of transactions.  These tend to
-be the bucket-based remotes (e.g. S3, B2, GCS, Swift).
+The basic strategy is to list one directory and processes it before using
+more directory lists to process any subdirectories. This is a mandatory
+backend feature, called `List`, which means it is supported by all backends.
+This strategy uses small amount of memory, and because it can be parallelised
+it is fast for operations involving processing of the list results.
 
-If you use the `--fast-list` flag then rclone will use this method for
-listing directories.  This will have the following consequences for
-the listing:
+Some backends provide the support for an alternative strategy, where all
+files beneath a directory can be listed in one (or a small number) of
+transactions. Rclone supports this alternative strategy through an optional
+backend feature called [`ListR`](/overview/#listr). You can see in the storage
+system overview documentation's [optional features](/overview/#optional-features)
+section which backends it is enabled for (these tend to be the bucket-based
+ones, e.g. S3, B2, GCS, Swift). This strategy requires fewer transactions
+for highly recursive operations, which is important on backends where this
+is charged or heavily rate limited. It may be faster (due to fewer transactions)
+or slower (because it can't be parallelized) depending on different parameters,
+and may require more memory if rclone has to keep the whole listing in memory.
 
-  * It **will** use fewer transactions (important if you pay for them)
-  * It **will** use more memory.  Rclone has to load the whole listing into memory.
-  * It *may* be faster because it uses fewer transactions
-  * It *may* be slower because it can't be parallelized
+Which listing strategy rclone picks for a given operation is complicated, but
+in general it tries to choose the best possible. It will prefer `ListR` in
+situations where it doesn't need to store the listed files in memory, e.g.
+for unlimited recursive `ls` command variants. In other situations it will
+prefer `List`, e.g. for `sync` and `copy`, where it needs to keep the listed
+files in memory, and is performing operations on them where parallelization
+may be a huge advantage.
 
-rclone should always give identical results with and without
-`--fast-list`.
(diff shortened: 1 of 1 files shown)


docs/content/docs.md
```diff
@@ -2110,33 +2110,43 @@ there were IO errors`.
...
-If you use the `--fast-list` flag then rclone will use this method for
-listing directories.  This will have the following consequences for
-the listing:
+However, some backends provide the support for an alternative strategy,
+where all files beneath a directory are listed in one (or a small number)
+of transactions. This will have the following consequences for the listing:
 
   * It **will** use fewer transactions (important if you pay for them)
-  * It **will** use more memory.  Rclone has to load the whole listing into memory.
+  * It **will** use more memory because rclone has to load the whole listing into memory
```
ncw, line 2123: `ListR` when used by `rclone ls` doesn't use more memory.

You need to explicitly provide the `--fast-list` flag to allow rclone to use more memory.

ncw (commented):
I salute your attempt to explain this. It is complicated!

Here is something for your understanding.

There are four main high level listing primitives in rclone: `walk.Walk`, `walk.ListR`, `march.March`, `walk.NewDirTree`.

- `walk.Walk` works exactly as described going through each directory one by one, unless `--fast-list` is set.
    - is used for `ncdu`, `rmdirs`
- `walk.ListR` will use the `ListR` primitive by default (unless the recursion depth is limited etc, etc)
    -  used by `ls*`, `*sum`, `size`, `delete`, `cat`
    - will fall back to `walk.Walk` if the command isn't recursive like `lsf` but if you put `lsf -R` then it will use `ListR`
- `march.March` works very much like `walk.Walk` and needs `--fast-list` to use `ListR`.
    - used by `sync`, `copy`, `move`, `check*`
- `walk.NewDirTree` - loads the directory listing into memory like `--fast-list`
    - used by `tree`
    - Can use `ListR`

So its a bit difficult for the user to tell whether the command will use ListR by default or whether they need to add `--fast-list` or `--disable ListR`.

How should we document these? We could do it here. Or in the individual commands, or both.

We could make a new command line flag group `ListR` and only apply that to the commands where `--fast-list` would make a difference so the `--fast-list` flag would only show then.

Hmm, this is complicated though!

- `rclone ls` - will use `ListR` - `--fast-list` would not work here.
- `rclone ls --max-depth 3` - will not use `ListR` but `--fast-list` would work here.

Perhaps we need a set of principles rather than trying to bombard the the user with the details. Something like this

- In general rclone tries to choose the best listing mechanism possible preferring `ListR` if it doesn't need to store the listed files in memory and `List` otherwise.
- You can stop rclone using `ListR` by using `--disable ListR`
- To allow rclone to use `ListR` in a way that would use more memory you add `--fast-list`.

Clear?

docs/content/docs.md
albertony: > ListR when used by rclone ls doesn't use more memory.

Right, because it is creating the resulting file objects for the entire list regardless, summing up to the same amount of memory? However, when submitting a request for the entire tree the response xml will surely be bigger, i.e. using more memory, than if requesting one directory or whatever at a time, but I guess that is small in comparison.

ncw: If you use `--fast-list` in `rclone sync` then rclone will load in every object in the listing tree into memory which might be millions of objects which uses GB of memory.

However just using `ListR` on its own doesn't use significantly more memory as it will be getting the objects in chunks of 1000 (s3 for example) and dealing with them before asking for the next chunk.

`rclone ls` will call `ListR` and output the objects to stdout as they come in, without storing them.

albertony:
> It is complicated!

Oh dear, it sure is! I appreciate your effort in explaining this to me, very instructive - thank you!

> Perhaps we need a set of principles rather than trying to bombard the the user with the details. 

I really think we should go this way. Trying to explain all the gory details seems unnecessary, difficult, probably just ending up confusing for the majority of users - and more work to keep up-to-date. I have tried to merge your last bullet points, which was very clear, into the existing text. So this doc section still tries to explain the background around ListR etc in some detail, without going too deep (?) into the specifics. I'm not done with it, but pushed an update so that if someone thinks I'm still heading the wrong way they can stop me... 😅

docs/content/docs.md
```diff
@@ -2111,33 +2111,48 @@ there were IO errors`.
...
+This strategy uses small amount of memory, and because it can be parallelised
+it is fast for operations involving processing of list results.
+
+Some backends provide the support for an alternative strategy, where all
+files beneath a directory can be listed in one (or a small number) of
+transactions. Rclone supports this alternative strategy through an optional
+backend feature called "[ListR](/overview/#listr)". You can see in the storage
+system overview documentation's [optional features](/overview/#optional-features)
+section which backends it is enabled for (these tend to be the bucket-based
+ones, e.g. S3, B2, GCS, Swift). This strategy require fewer transactions
```
ncw, line 2129: requires

ncw (commented):
Very very nice :-) :star:

I think that is very clear and also accurate without bombarding the user with details.

Shall we always put `List` and `ListR` in backquotes when talking about the internal operations?

I put some language suggestions inline.

ncw (approved):
Very nice - let's merge :-)

albertony:
Great. Sorry for the delay, I've been on a short vacation. Had a fresh read-through and did a small improvement(?), starting from `Rclone is not able to take all relevant parameters into account` and down. If you want to have another look then please do, or else I think I will just risk merging it it a while.. 🤞

ncw (approved):
LGTM :-)

PR #7352, opened 2023-10-06
operations: implement --partial-upload-extension option

vkit-maytech:
Added `--partial-upload-extension` to allow setting temporary file name extension 

Default is `partial`

Some systems may already have defined filter for temporary files (e.g. `*.tmp`). So it would be great to have an option to configure how `rclone` names it.

No, it wasn't. This is our need. And it can be useful for others.

#### Checklist

### DIFF
--- a/fs/config/configflags/configflags.go
+++ b/fs/config/configflags/configflags.go
@@ -148,6 +148,7 @@ func AddFlags(ci *fs.ConfigInfo, flagSet *pflag.FlagSet) {
 	flags.FVarP(flagSet, &ci.TerminalColorMode, "color", "", "When to show colors (and other ANSI codes) AUTO|NEVER|ALWAYS", "Config")
 	flags.FVarP(flagSet, &ci.DefaultTime, "default-time", "", "Time to show if modtime is unknown for files and directories", "Config,Listing")
 	flags.BoolVarP(flagSet, &ci.Inplace, "inplace", "", ci.Inplace, "Download directly to destination file instead of atomic download to temp/rename", "Copy")
+	flags.StringVarP(flagSet, &ci.PartialUploadExtension, "partial-upload-extension", "", ci.PartialUploadExtension, "Add partial-upload-extension to temporary file name when --inplace is not used", "Copy")
 }
 
 // ParseHeaders converts the strings passed in via the header flags into HTTPOptions
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -361,7 +361,7 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 	if !ci.Inplace && f.Features().Move != nil && f.Features().PartialUploads && !strings.HasSuffix(remote, ".rclonelink") {
 		// Avoid making the leaf name longer if it's already lengthy to avoid
 		// trouble with file name length limits.
-		suffix := "." + random.String(8) + ".partial"
+		suffix := "." + random.String(8) + "." + ci.PartialUploadExtension
 		base := path.Base(remotePartial)
 		if len(base) > 100 {
 			remotePartial = remotePartial[:len(remotePartial)-len(suffix)] + suffix
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -1278,8 +1278,9 @@ flag set) such as:
 - sftp
 
 Without `--inplace` (the default) rclone will first upload to a
-temporary file with an extension like this where `XXXXXX` represents a
-random string.
+temporary file with an extension like this, where `XXXXXX` represents a
+random string and `partial` is [--partial-upload-extension](#partial-upload-extension) value
+(`partial` by default).
 
     original-file-name.XXXXXX.partial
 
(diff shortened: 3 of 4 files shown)


vkit-maytech:
We know about the problem with the length of `partial-upload-extension`. It can be greater than 100, and it causes panic 🙁 
Need advice, how to validate it @ncw

ncw:
The panic will be here

```
		if len(base) > 100 {
			remotePartial = remotePartial[:len(remotePartial)-len(suffix)] + suffix
		} else {
			remotePartial += suffix
		}
```

I think we should probably make sure the suffix isn't too long and return an error to the user.

You can do that here

https://github.com/rclone/rclone/blob/a752563842b3a603c7d73b00e71c6ff1fd120382/fs/config/configflags/configflags.go#L171

fs/operations/operations.go
```diff
@@ -361,10 +361,21 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
...
+			ext = fs.GetConfig(context.Background()).PartialUploadExtension
+		}
+
+		suffix := "." + random.String(8) + "." + ext
 		base := path.Base(remotePartial)
 		if len(base) > 100 {
-			remotePartial = remotePartial[:len(remotePartial)-len(suffix)] + suffix
+			limit := len(base) - len(suffix)
+
+			if len(suffix) > len(base) {
```
ncw, line 374: This can never be true since len(base) > 100 and we know len(suffix) = len("." + random.String(8) + "." + ext) <= 25

fs/operations/operations.go
```diff
@@ -361,10 +361,21 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 	if !ci.Inplace && f.Features().Move != nil && f.Features().PartialUploads && !strings.HasSuffix(remote, ".rclonelink") {
 		// Avoid making the leaf name longer if it's already lengthy to avoid
 		// trouble with file name length limits.
-		suffix := "." + random.String(8) + ".partial"
+		ext := strings.TrimSpace(ci.PartialUploadExtension)
+		if len(ext) == 0 {
```
ncw, line 365: I don't think you need this `if len(ext) == 0 { ... }` clause do you as the context.Background() Config wasn't what the user asked for here.

fs/config/configflags/configflags.go
```diff
@@ -322,6 +324,12 @@ func SetFlags(ci *fs.ConfigInfo) {
 	multiThreadStreamsFlag := pflag.Lookup("multi-thread-streams")
 	ci.MultiThreadSet = multiThreadStreamsFlag != nil && multiThreadStreamsFlag.Changed
 
+	partialUploadExtension = strings.TrimSpace(partialUploadExtension)
+	if len(partialUploadExtension) < 1 || len(partialUploadExtension) > 15 {
```
ncw, line 328: Just looking at rclone's other options, it would be more consistent to call this `--partial-suffix`

Also upload is inaccurate since these are used for downloads to local, so let's leave it out.

So the default should be `.partial` and an empty string is acceptable.

fs/operations/operations.go
```diff
@@ -361,10 +361,21 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 	if !ci.Inplace && f.Features().Move != nil && f.Features().PartialUploads && !strings.HasSuffix(remote, ".rclonelink") {
 		// Avoid making the leaf name longer if it's already lengthy to avoid
 		// trouble with file name length limits.
-		suffix := "." + random.String(8) + ".partial"
+		ext := strings.TrimSpace(ci.PartialUploadExtension)
```
ncw, line 364: No need for `TrimSpace` - if the user wants spaces, they shall have spaces!

fs/operations/operations.go
```diff
@@ -361,10 +361,21 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
...
+		if len(ext) == 0 {
+			ext = fs.GetConfig(context.Background()).PartialUploadExtension
+		}
+
+		suffix := "." + random.String(8) + "." + ext
 		base := path.Base(remotePartial)
 		if len(base) > 100 {
-			remotePartial = remotePartial[:len(remotePartial)-len(suffix)] + suffix
+			limit := len(base) - len(suffix)
```
ncw, line 372: I think since we've limited the size of the extension, we can just use the original 

    remotePartial = remotePartial[:len(remotePartial)-len(suffix)] + suffix

Here, which will give a max total length of 125 which is fine.

fs/operations/operations.go
justakit: I've added all that checks for those cases, when `rclone` is used as library, and not as `cmd`. So the user won't be able to break execution even if they really want 🙂
I am still worried about that case. Should the user be responsible for value they use?

fs/operations/operations.go
justakit: Yeah. If it will be `suffix` and not `extension` than it definitely can be zero-length
Previous logic with file ends with `*.RNDSTRNG.` just had no sense

fs/operations/operations.go
ncw: If you want to check the length is <= 15 in the function itself then that is fine - do that and return an error.

ncw (approved):
This looks great now - thank you! I will merge now.

PR #7358, opened 2023-10-10
operations: fix OpenOptions ignored in copy if operation was a multiT…

vitorog:

Similar to https://github.com/rclone/rclone/pull/7354, but now for multithreadCopy.
When doing a multi-threaded copy operation, the OpenOption array was ignored.

No

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -422,8 +422,17 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 					removeFailedPartialCopy(ctx, f, remotePartial)
 				})
 			}
+
+			uploadOptions := []fs.OpenOption{hashOption}
+			for _, option := range ci.UploadHeaders {
+				uploadOptions = append(uploadOptions, option)
+			}
+			if ci.MetadataSet != nil {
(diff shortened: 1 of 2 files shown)


fs/operations/operations.go
```diff
@@ -423,7 +423,14 @@ func Copy(ctx context.Context, f fs.Fs, dst fs.Object, remote string, src fs.Obj
 				})
 			}
 			if doMultiThreadCopy(ctx, f, src) {
-				dst, err = multiThreadCopy(ctx, f, remotePartial, src, ci.MultiThreadStreams, tr)
+				options := []fs.OpenOption{hashOption}
```
ncw, line 426: I think we should rename this `uploadOptions` and and move it and the following 6 lines up before the `if`, then we can re-use `uploadOptions` a bit further down when calling `Update` and `Put`.

I have on my TODO list to refactor `operations.Copy` - it is getting out of hand!

ncw (commented):
Another great spot - thank you!

See comment inline to reduce duplication.

fs/operations/operations.go
vitorog: hi, @ncw , thanks for the review! I updated the PR

ncw (approved):
That looks perfect - thank you :-)

PR #7385, opened 2023-10-23
[fs/http] Reload certificates when they change

salehqt:

In corporate environments, client certificates have short life times for added security, and they get renewed automatically. This means that client certificate can expire in the middle of long running command such as `mount`.

This commit adds a simple timestamp check to automatically reload the certificate if it has been renewed on disk.

No

#### Checklist

- [x] I have added documentation for the changes if appropriate. **No documentation needed**

### DIFF
--- a/fs/fshttp/http.go
+++ b/fs/fshttp/http.go
@@ -69,6 +69,13 @@ func NewTransportCustom(ctx context.Context, customize func(*http.Transport)) ht
 		if err != nil {
 			log.Fatalf("Failed to load --client-cert/--client-key pair: %v", err)
 		}
+		if cert.Leaf == nil {
+			// Leaf is always the first certificate
+			cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
+			if err != nil {
+				log.Fatalf("Failed to parse the certificate")
+			}
+		}
 		t.TLSClientConfig.Certificates = []tls.Certificate{cert}
 	}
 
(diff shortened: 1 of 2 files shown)


ncw (commented):
I think this is a great idea :-)

I'd like to avoid doing two Stat's on every call to RoundTrip. There can be 1000s of those per second.

Note also that RoundTrip can be called concurrently and the code you've written isn't safe for concurrent use.

It would be easy to add a time limit so it only checks for the new cert once per 1s/10s/60s or so.

I think adding a mutex will be enough to fix most of the concurrency problems.

A final thought - does anything need reloading when these certs are updated?

Thank you

salehqt:
> I think this is a great idea :-)
> 
> I'd like to avoid doing two Stat's on every call to RoundTrip. There can be 1000s of those per second.
I have a better way of avoiding the Stats: checking the expiration date of the certificate. I will code it up when I have the chance. In practice, overhead of Stat is negligible compared to HTTP requests.
> 
> Note also that RoundTrip can be called concurrently and the code you've written isn't safe for concurrent use.
> 
I agree, the code doesn't have any protection around writes, but it is idempotent. In practice doesn't cause any problems. I will add mutex just to improve reliability.

> It would be easy to add a time limit so it only checks for the new cert once per 1s/10s/60s or so.
> 
> I think adding a mutex will be enough to fix most of the concurrency problems.
> 
> A final thought - does anything need reloading when these certs are updated?
This code have been working for a while. I am only publishing it now as a part of upstreaming our fork of rclone.
> 
> Thank you

Any code examples on how to test this feature would be appreciated.

salehqt:
@ncw , I made the changes you requested:
 - Replaced Stat calls with parsing certificate and checking NotAfter time
 - Added mutex for concurrent reloading of certificates
 - Added tests for this feature

ncw:
I have rebased this and I'll merge when the CI goes green. Thanks for doing this @salehqt

Issue #7029, opened 2023-05-27
Data direction when using --dry-run is unclear


glibg10b:
```
$ rclone bisync . remote:university --exclude='*.{log,out,aux,toc}' --dry-run
2023/05/27 09:21:59 NOTICE: bisync is EXPERIMENTAL. Don't use in production!
2023/05/27 09:22:02 NOTICE: wtw-158/notes.rnote: Skipped copy as --dry-run is set (size 313)
2023/05/27 09:22:02 NOTICE: wtw-158/examples.rnote: Skipped copy as --dry-run is set (size 2.873Mi)
2023/05/27 09:22:03 NOTICE: ebn-111/notes.pdf: Skipped copy as --dry-run is set (size 70.623Ki)
2023/05/27 09:22:03 NOTICE: ebn-111/notes.tex: Skipped copy as --dry-run is set (size 14.479Ki)
2023/05/27 09:22:03 NOTICE: has-110/notes.pdf: Skipped copy as --dry-run is set (size 45.267Ki)
2023/05/27 09:22:03 NOTICE: has-110/notes.tex: Skipped copy as --dry-run is set (size 11.756Ki)
2023/05/27 09:22:05 NOTICE: 
Transferred:   	   3.012 MiB / 3.012 MiB, 100%, 1.503 MiB/s, ETA 0s
Checks:                 6 / 6, 100%
Transferred:            6 / 6, 100%
Elapsed time:         7.8s
```

How do I know which way the files would be copied?

```
$ rclone --version
rclone v1.61.1-DEV
- os/version: gentoo 2.13 (64 bit)
- os/kernel: 6.1.24-gentoo-dist (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.20.3
- go/linking: dynamic
- go/tags: none
```

nielash:
If you add `-v` you should see this:

```
2023/07/12 08:09:36 INFO  : - Path1    Queue copy to Path2                 - /path/to/your/file.txt
2023/07/12 08:09:36 INFO  : - Path1    Do queued copies to                 - Path2
```

That said, I agree this could be improved (and it's an easy fix.)

SatyajitR:
DEV test results
`> rclone bisync ..\hello_world\ remotemc:uni --exclude='*.{log,out,aux,toc}' --resync --dry-run
2023/10/25 08:31:07 NOTICE: bisync is EXPERIMENTAL. Don't use in production!
[...]

PR #7389, opened 2023-10-25
#7029 Data direction when using --dry-run 

SatyajitR:
#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -2507,7 +2507,7 @@ func SkipDestructive(ctx context.Context, subject interface{}, action string) (s
 			size = do.Size()
 		}
 		if size >= 0 {
-			fs.Logf(subject, "Skipped %s to %s as %s is set (size %v)", fs.LogValue("skipped", action), flag, subject.(fs.Object).Remote(), fs.LogValue("size", fs.SizeSuffix(size)))
+			fs.Logf(subject, "Skipped %s to %s as %s is set (size %v)", fs.LogValue("skipped", action), subject.(fs.Object).Remote(), flag, fs.LogValue("size", fs.SizeSuffix(size)))
 		} else {
 			fs.Logf(subject, "Skipped %s as %s is set", fs.LogValue("skipped", action), flag)
 		}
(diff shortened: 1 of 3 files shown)


fs/operations/operations.go
nielash, line 1: A few thoughts:

- `subject.(fs.Object).Remote()` is a relative path from the Fs's root, so it does not actually tell us if this is Path1 or Path2. It is also already printed here because of the first `subject` parameter. I think maybe you were looking for something more like `subject.(fs.Object).Fs().Root()` (or a `filepath.Join()` of this with `Remote()`.
- The message string makes sense for copies, but may not make sense for other types of actions that use `operations.SkipDestructive()`. For example, a delete would read: 
```
NOTICE: potato: Skipped delete to potato as --dry-run is set (size 21)
```
- Since so many things outside of Bisync use this function, I'm not sure if this is the best place to address this, as it's not a clarification that would really be needed outside of Bisync (for example, in Sync, the destination root would be obvious.) I think a better place might be Bisync's [`indent` function](https://github.com/rclone/rclone/blob/53aea1a71df59c0b1285f67cff8c9c42e76446a5/cmd/bisync/log.go#L17) (perhaps using NOTICE instead of INFO if `--dry-run` is set).

nielash (commented):
Thank you for taking a stab at this! I made a few comments inline. Also, the bisync tests are failing here because they need to be "goldenized" after log changes. See [this guide](https://rclone.org/bisync/#updating-golden-results).

KDreynolds:
I can try and pick up this issue if it has stalled.

nielash:
> I can try and pick up this issue if it has stalled.

If you're still interested, that would be great 🙂 Here's an example of where that `indent` function is called:

https://github.com/rclone/rclone/blob/810644e8735330db1cbd70745802284e8624b423/cmd/bisync/deltas.go#L489-L494

I think the info is already there, so it should just be a matter of making sure it gets seen at the `NOTICE` log level when `b.opt.DryRun` is true.

KDreynolds:
Alright sounds good. I'm spending some time working on a separate unrelated issue but I can begin spending some time on this now.

nielash:
Closing as duplicate of #7647

PR #7397, opened 2023-10-31
fix: dlan server init waitChan

wuxingzhong:


ncw:
This looks like a great fix.

How did you notice it? I think Go should have paniced if we ever called `close(s.waitChan)` but no-one has reported that bug!

ncw:
Thank you :-)

Issue #7391, opened 2023-10-26
Crash when setting --s3-copy-cutoff=0


gramakri:
#### What is the problem you are having with rclone?

When setting `--s3-copy-cutoff=0` there is a divide by 0 crash.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.64.2
- os/version: ubuntu 22.04 (64 bit)
- os/kernel: 5.15.0-86-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.3
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Linux Ubuntu 22.04 64-bit

#### Which cloud storage system are you using? (e.g. Google Drive)

Vultr Object Storage

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

`./rclone copyto -vvvv vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz.2 --s3-copy-cutoff=0`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
./rclone copyto -vvvv vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz.2 --s3-copy-cutoff=0
2023/10/26 07:41:57 DEBUG : rclone: Version "v1.64.2" starting with parameters ["./rclone" "copyto" "-vvvv" "vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz" "vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz.2" "--s3-copy-cutoff=0"]
2023/10/26 07:41:57 DEBUG : Creating backend with remote "vultr://cloudron-backups2/snapshot/app_410ea704-713a-4cf0-bb9d-e59aa14ab56b.tar.gz"
2023/10/26 07:41:57 DEBUG : Using config file from "/root/.config/rclone/rclone.conf"
[...]

PR #7400, opened 2023-11-02
s3: validate CopyCutoff size before copy

hoyho:
I can can reproduce this with other compatible S3 backend. Try to fix it. 

fix panic when using `--s3-copy-cutoff=0` 

When use CopyCutoff=0, it just similar to set chunk-size as original object and do server side copy

https://github.com/rclone/rclone/issues/7391

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -3022,6 +3022,14 @@ func (f *Fs) setUploadChunkSize(cs fs.SizeSuffix) (old fs.SizeSuffix, err error)
 	return
 }
 
+func checkCopyCutoff(cs fs.SizeSuffix) error {
+	minCopySize := fs.SizeSuffixBase
+	if cs < minCopySize {
+		return fmt.Errorf("value is too small (%v is less than %v)", cs, minCopySize)
+	}
+	return nil
+}
+
 func checkUploadCutoff(cs fs.SizeSuffix) error {
 	if cs > maxUploadCutoff {
 		return fmt.Errorf("%s is greater than %s", cs, maxUploadCutoff)
@@ -3321,6 +3329,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	err = checkUploadCutoff(opt.UploadCutoff)
 	if err != nil {
 		return nil, fmt.Errorf("s3: upload cutoff: %w", err)
+	}
+	err = checkCopyCutoff(opt.CopyCutoff)
+	if err != nil {
+		return nil, fmt.Errorf("s3: --s3-copy-cutoff: %w", err)
 	}
 	if opt.Versions && opt.VersionAt.IsSet() {
 		return nil, errors.New("s3: can't use --s3-versions and --s3-version-at at the same time")


backend/s3/s3.go
```diff
@@ -4419,6 +4419,10 @@ func (f *Fs) copyMultipart(ctx context.Context, copyReq *s3.CopyObjectInput, dst
 
 	srcSize := src.bytes
 	partSize := int64(f.opt.CopyCutoff)
+	if f.opt.CopyCutoff <= 0 { //making sure we don't divide by zero
```
ncw, line 4422: I don't think this bit of code can be called any more can it? So we don't need a check here.

backend/s3/s3.go
```diff
@@ -4346,7 +4346,7 @@ func (f *Fs) copy(ctx context.Context, req *s3.CopyObjectInput, dstBucket, dstPa
 		req.StorageClass = &f.opt.StorageClass
 	}
 
-	if src.bytes >= int64(f.opt.CopyCutoff) {
+	if src.bytes >= int64(f.opt.CopyCutoff) && f.opt.CopyCutoff > 0 {
```
ncw, line 4349: This works fine unless the size of the object is > 5 GB then the single part copy will fail.

I think we should probably insert a check here instead and return an error to the user when initialising the Fs

https://github.com/rclone/rclone/blob/52d444f4a91cc7afd1552574dced484fd3b5138b/backend/s3/s3.go#L3252

ncw (commented):
Thank you for doing this :-)

I think we should probably return an error earlier in the process - see inline comments.

Thank you

backend/s3/s3.go
hoyho: yes, I was debugging into this line and make a change so it was included in this commit. Actually, we don't need it anymore

backend/s3/s3.go
hoyho: Thanks for reminding that. I didn't aware of single part copy before..
Now I refactor it to verify when initialing the backend.
please take a look

backend/s3/s3.go
```diff
@@ -3022,6 +3022,14 @@ func (f *Fs) setUploadChunkSize(cs fs.SizeSuffix) (old fs.SizeSuffix, err error)
 	return
 }
 
+func checkCopyCutoff(cs fs.SizeSuffix) error {
+	minCopySize := fs.SizeSuffixBase
+	if cs < minCopySize {
+		return fmt.Errorf("%s is less than %s", cs, minCopySize)
```
ncw, line 3028: Can you say what is out of range in the error? Also use %d/%v is preferred, so something like

    return fmt.Errorf("--s3-copy-cutoff is too small (%d is less than %d) ", cs, minCopySize)

ncw (commented):
This is looking better now - thank you.

I put a note inline about re-wording the error message to be more explicit.

Thank you.

backend/s3/s3.go
hoyho: Actually, there's additional information in outer layer.  
```go
	err = checkCopyCutoff(opt.CopyCutoff)
	if err != nil {
		return nil, fmt.Errorf("s3: copy cutoff: %w", err)
	}
```
Though it doesn't specific which parameter is wrong. I would like to include `--s3-copy-cutoff` here.  
So the output will be:
```
2024/03/14 17:20:47 Failed to create file system for "minipc-s3:/buildscripts/6G": s3: --s3-copy-cutoff: value is too small (0 is less than 1)
```
Is the output clear now?

ncw: Looks good :-)

ncw (approved):
Looks great - thank you :-)

PR #7404, opened 2023-11-03
Listremotes filtering

albertony:

The main purpose is to add support for filtering the output from `listremotes` command, as discussed in the forum. Some additional, somewhat related, features were also added, namely ordering and json output. During extended testing of this command, some issues and inconsistencies within the fs/filter and fs/config libs, which the listremotes command depends on, were discovered and improved.

Now to the details: First the improvements to listremotes, as these are the most user-facing part, then later some details regarding the changes in fs libs - although these commits comes first in the PR as some of they create a foundation for some of the listremote improvements.

Improvements to listremotes:
- Including source of the remote definition in the `--long` output format, i.e. if it was defined in environment variable or in config file. The complete remote configuration may contain options from both environment and config, but this shows where it is "defined" - where the name and type comes from. This follows the documented [preference](https://rclone.org/docs/#precedence), so if it is defined in both places its the one from environment that will be shown.
  ```
  $ rclone listremotes --long
  temp_box: dropbox    environment
  temp_jfs: jottacloud environment
  box:      dropbox    file
  box_1:    dropbox    file
  box_2:    dropbox    file
  jfs:      jottacloud file
  gd:       drive      file
  box_1b:   dropbox    file
  ```
- Optional flag `--type` to list only remotes of a specific type:
  ```
  $ rclone listremotes --long --type dropbox
  temp_box: dropbox environment
  box:      dropbox file
  box_1:    dropbox file
  box_2:    dropbox file
  box_1b:   dropbox file
  ```
- Optional flag `--source` to list only remotes from a specific source:
  ```
  $ rclone listremotes --type dropbox --source environment
  temp_box: dropbox environment
  ```
- Search/filter on name. A query string can be specified as an optional (unnamed) argument. I did considered implementing it as a simple string "binary" matching, or full blown regular expression, but I find the glob syntax quite convenient, flexible while easy to use - easier to understand than regex for less technical users at least. I ended up re-using rclone's [filtering syntax](https://rclone.org/filtering/): This is convenient in that it uses a quite easy-to-use glob pattern by default, while allowing for some more advanced features and even regular expressions for those who want it:
  ```
  $ rclone listremotes box
  
  $ rclone listremotes box*
  box:
  box_1:
  box_2:
  box_1b:
  $ rclone listremotes *box
  temp_box:
  $ rclone listremotes box{{(_\d+)?}} --type dropbox
  box:
  box_1:
  box_2:
  ```
- Sorting. Option flag `--order-by` can be used to sort on any of the fields (`name`, `type` and `source`) in any direction (`ascending` or `asc`, and `descending` or `desc`), and they can be combined with the syntax: `field[=order],field[=order],...` (order is optional).
  ```
  $ rclone listremotes --long --order-by type,source=desc,name
  gd:       drive      file
  box:      dropbox    file
  box_1:    dropbox    file
  box_1b:   dropbox    file
  box_2:    dropbox    file
  temp_box: dropbox    environment
  jfs:      jottacloud file
  temp_jfs: jottacloud environment
  ```
- Machine-readable output with option `--json`, also for completeness with other commands such as `size` and `about`. It produces a valid array json output in format similar to `lsjson`, and can be combined with filtering and other options (except `--long`):
  ```
  $ rclone listremotes --json --type jottacloud
  [
  {"name":"temp_jfs","type":"jottacloud","source":"environment"},
  {"name":"jfs","type":"jottacloud","source":"file"}
  ]
  ```

Improvements in fs/filter:
- To implement the glob-filtering I had to tweak the `fs.GlobToRegexp` function a little, so that we can tell it to not treat `/` as path separator, and to not make `*` stop consuming them, and not support `**` as additional symbol that will consume them, etc.. As can be seen from the examples above, the `rclone listremotes box` does not give any result as it would with the simplest binary matching alternative, so this is sort of a tradeoff between the absolute simplest syntax and flexibility.
- For consistency I updated the `help flags` command to use the same logic. It used plain regex before, and my reasoning was that we should try to avoid using different filtering syntax for different command, as that will not be easy to remember for users. The main difference is that a glob pattern is implicitly anchored. E.g. previously you could write `rclone help flags dropbox` and it would return any flags with this word in its name or description. But now you will have to write `rclone help flags *dropbox*` to get the same result. That is a less convenient, so a bit unfortunate actually. Maybe in an attempt at compensating - I added some more flexibility to the flags filtering:
  - You can now add flag `--name` to make the filter match the *name* of the flag only. E.g. `rclone help flags *password* --name` will only match flags with "password" in their name, it will not include `--sftp-pass` even if it mentions "password" in its description.
  - You can now specify name of a flag group with `--group`, e.g. `rclone help flags --group sync` or  `rclone help flags --group backend`. This uses (case insensitive) string matching by the way, requires complete match on the group name. Contradicting myself a bit here, I guess. Could have used the same pattern matching also here, but that complicated flags and parameter passing code, and there are only a few short group names so I figured there is not much to loose from having to spell out the entire name. The same is true for the new `--type` and `--source` flags in `listremote` by the way.
- **Note:** This part I'm least certain about. Is the existing filter syntax for help flags so well established, and easy to use, that it is bad to change it? If uncertain, I could split this out to a separate PR. In any case, do you agree with me that the glob syntax is clever, and useful enough that we want to keep it for the listremotes command (and possibly use it elsewhere as well)?

Improvements in fs/config:
- Issues:
  - It is possible to create nameless remotes with environment variable, like this:
    ```
    $ set RCLONE_CONFIG__TYPE=sftp
    $ rclone listremotes --long
    : sftp
    ```
  - If you define a remote with same name in environment as in file, it is listed twice:
    ```
    $ type rclone.conf
    [box]
    type = dropbox
    
    [box_1]
    type = dropbox
    
    [box_1b]
    type = dropbox
    
    $ set RCLONE_CONFIG_BOX_1B_TYPE=sftp
    $ rclone listremotes --long
    box:    dropbox
    box_1:  dropbox
    box_1b: dropbox
    box_1b: dropbox
    ```
    And even worse; notice that the duplicate remote is listed with same type - the type from config file. So it does not actually reflect the truth very well. And another point to this is that it chooses the remote type from config file over the environment, which is opposite of what the documentation says - and what rclone uses in "real" operations:
    ```
    $ rclone lsd box_1b: -vvv
    DEBUG : Setting type="sftp" for "box_1b" from environment variable RCLONE_CONFIG_WORK_BOX_1B_TYPE
    ```
- Explanations:
  - While looking into the issues described above, I found the some parts of the fs/config a bit confusing, misleading and wrong (in my opinion): Function `config.FileSections` did not only return the sections, i.e. remotes, from the config file - but also included any configured as environment variables. Same with `config.FileGet`, which retrieves value of options. Misleading function names, at least. But also; while collecting remote names from environment it appends them to the list even if there is one with same name in configuration file, which could be considered a feature to inform users - except as we see above `listremotes` does not present it with its correct type so its a bug there regardless what we want it to show. Which brings me to next point; when collecting remotes from environment, remember these are defined by the `_TYPE` environment variable specifying the type of remote, the type information is ignored and when using `listremotes` command it uses `config.FileGet` to get the value of the type property. But this function reads the configuration file first, and only if not found there considers environment variable - which is the wrong way around. Together this leads to the issue above.
- Improvements:
  - Make fs/config functions with prefix `File` consider the configuration file only, consistent with its name. Then make new "convenience" functions `GetRemotes` and `GetValues` which also considers environment (there were already a couple of similar functions).
  - The new `GetRemotes` returns not only the remote names, but also the type and the "source" (if it was found from file or environment). This ensures we make use of the type information from the `_TYPE` environment variable when that is used to define a remote (name) as well, avoiding the extra call but more importantly avoiding inconsistency (although the next point should also fix this).
  - When considering options set in environment variables, make sure these take precedence over configuration file.

https://forum.rclone.org/t/suggestion-separate-crypt-and-non-crypt-remotes/42581

#### Checklist

### DIFF
--- a/backend/alias/alias_internal_test.go
+++ b/backend/alias/alias_internal_test.go
@@ -23,8 +23,8 @@ func prepare(t *testing.T, root string) {
 	configfile.Install()
 
 	// Configure the remote
-	config.FileSet(remoteName, "type", "alias")
-	config.FileSet(remoteName, "remote", root)
+	config.FileSetValue(remoteName, "type", "alias")
+	config.FileSetValue(remoteName, "remote", root)
 }
 
 func TestNewFS(t *testing.T) {
--- a/backend/cache/cache_internal_test.go
+++ b/backend/cache/cache_internal_test.go
@@ -875,12 +875,12 @@ func (r *run) newCacheFs(t *testing.T, remote, id string, needRemote, purge bool
 	cacheRemote := remote
 	if !remoteExists {
 		localRemote := remote + "-local"
-		config.FileSet(localRemote, "type", "local")
-		config.FileSet(localRemote, "nounc", "true")
+		config.FileSetValue(localRemote, "type", "local")
+		config.FileSetValue(localRemote, "nounc", "true")
 		m.Set("type", "cache")
 		m.Set("remote", localRemote+":"+filepath.Join(os.TempDir(), localRemote))
 	} else {
-		remoteType := config.FileGet(remote, "type")
+		remoteType := config.GetValue(remote, "type")
 		if remoteType == "" {
 			t.Skipf("skipped due to invalid remote type for %v", remote)
 			return nil, nil
@@ -891,14 +891,14 @@ func (r *run) newCacheFs(t *testing.T, remote, id string, needRemote, purge bool
 				m.Set("password", cryptPassword1)
 				m.Set("password2", cryptPassword2)
 			}
-			remoteRemote := config.FileGet(remote, "remote")
+			remoteRemote := config.GetValue(remote, "remote")
 			if remoteRemote == "" {
 				t.Skipf("skipped due to invalid remote wrapper for %v", remote)
 				return nil, nil
 			}
 			remoteRemoteParts := strings.Split(remoteRemote, ":")
 			remoteWrapping := remoteRemoteParts[0]
-			remoteType := config.FileGet(remoteWrapping, "type")
+			remoteType := config.GetValue(remoteWrapping, "type")
 			if remoteType != "cache" {
 				t.Skipf("skipped due to invalid remote type for %v: '%v'", remoteWrapping, remoteType)
 				return nil, nil
--- a/backend/cache/cache_internal_test.go
+++ b/backend/cache/cache_internal_test.go
@@ -850,8 +850,8 @@ func (r *run) encryptRemoteIfNeeded(t *testing.T, remote string) string {
 func (r *run) newCacheFs(t *testing.T, remote, id string, needRemote, purge bool, flags map[string]string) (fs.Fs, *cache.Persistent) {
 	fstest.Initialise()
 	remoteExists := false
-	for _, s := range config.FileSections() {
-		if s == remote {
+	for _, s := range config.GetRemotes() {
+		if s.Name == remote {
 			remoteExists = true
 		}
 	}
--- a/cmd/bisync/bisync_test.go
+++ b/cmd/bisync/bisync_test.go
@@ -1304,7 +1304,7 @@ func touchFiles(ctx context.Context, dateStr string, f fs.Fs, dir, glob string) 
 		return files, fmt.Errorf("invalid date %q: %w", dateStr, err)
 	}
 
-	matcher, firstErr := filter.GlobToRegexp(glob, false)
+	matcher, firstErr := filter.GlobPathToRegexp(glob, false)
 	if firstErr != nil {
 		return files, fmt.Errorf("invalid glob %q", glob)
 	}
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/configflags"
 	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/fs/filter"
 	"github.com/rclone/rclone/fs/filter/filterflags"
 	"github.com/rclone/rclone/fs/log/logflags"
 	"github.com/rclone/rclone/fs/rc/rcflags"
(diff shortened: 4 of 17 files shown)


ncw:
Hi @albertony - I don't know why I didn't merge this - it looks great!

Can you rebase it please and I'll try to get it in v1.68 - thank you :-)

albertony:
✅ Rebased

ncw (approved):
Lovely - thank you :-)

ncw:
@albertony 

I've just noticed that the `rclone help flags` seems to be broken

```
$ rclone help flags "multi"
Usage:
  rclone help flags [<filter>] [flags]

Flags:
      --group string   Only include flags from specific group
  -h, --help           help for flags
      --name           Apply filter only on flag names



Additional help topics:

```

albertony:
It would now have to be: 
```
$ rclone help flags "*multi*"
```

See section "Improvements in fs/filter":

> For consistency I updated the help flags command to use the same logic...

> Note: This part I'm least certain about...

ncw:
Ah. Maybe should change that then because I use that command several times a day and having to type `rclone help flags "*string*"` is going to annoy me!

albertony:
I see. I think reverting a single small commit is enough - done in PR https://github.com/rclone/rclone/pull/7994

albertony:
Well, these entire config example sections are notoriously outdated, in my opinion. Didn't we discuss somewhere what to do about these in general - if we should remove them, or if not how to keep them updated.. not sure. You are probably correct about that commit, coming from PR https://github.com/rclone/rclone/pull/6211, where there also were other changes - e.g. the empty line between individual sections where before (and still in docs) they were no such empty lines:

```
Option nounc.
Disable UNC (long path names) conversion on Windows.
Choose a number from below, or type in your own boolean value (true or false).
Press Enter for the default (false).
 1 / Disables long file names.
   \ (true)
nounc>

Edit advanced config?
y) Yes
n) No (default)
y/n>
```

But there are definitely other commits/PRs as well, that have changes making this part of the docs outdated.

My vote goes to 3) for now, create an issue for 2), either just the configuration complete "summary" part, or updating them more general. I think I'm able to update the summary sections without too much effort, with some multi-line search + regex + manual use of vscode editor features, so if that part is the biggest concern I can jump into it within a couple of days.

ncw:
> Well, these entire config example sections are notoriously outdated, in my opinion.

Yes some of them are!

Some users really like them though.

> Didn't we discuss somewhere what to do about these in general - if we should remove them, or if not how to keep them updated.. not sure. 

In an ideal world we'd have a script which regenerated them automatically. Sounds like a lot of work though!

> You are probably correct about that commit, coming from PR #6211, where there also were other changes - e.g. the empty line between individual sections where before (and still in docs) they were no such empty lines:

Ah, I hadn't noticed that.

Probably less of a problem for user confusion.

> But there are definitely other commits/PRs as well, that have changes making this part of the docs outdated.
> 
> My vote goes to 3) for now, create an issue for 2), either just the configuration complete "summary" part, or updating them more general.

OK sounds good to me :-)

> I think I'm able to update the summary sections without too much effort, with some multi-line search + regex + manual use of vscode editor features, so if that part is the biggest concern I can jump into it within a couple of days.

It is the thing that stuck out but whether it will actually confuse users, I don't know - probably not!

albertony:
> 2\. update the 50 docs (note there are several in the s3 and other setups).

Done here: https://github.com/rclone/rclone/pull/8015

Issue #5685, opened 2021-10-08
bisync: respect --backup-dir


ivandeex:
## Synopsis

bisync should support the `--backup-dir` flag similar to `sync`.

### https://github.com/rclone/rclone/pull/5164#pullrequestreview-729822158 (ncw)

Will users be able to use something like `--backup-dir` or `--suffix`? I can see users wanting that to prevent data loss. Maybe `--suffix` works already and you could filter the backups out using the filtering? Perhaps this should be automatic.

### https://github.com/rclone/rclone/pull/5164#issuecomment-842407758 (JWink3101)

Personally, I'd **much** rather see `--backup-dir` support out of the gate with that being the check. And again, it matches current rclone behavior.

## How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

acegene:
Behavior with rclone v1.64.2 seems to silently ignore the flag. Should it be changed to give a warning/error?

Bumping as this feature is especially useful given the beta nature of bisync.

nielash:
@acegene I have great news for you -- this will be fixed by #7410 (specifically https://github.com/rclone/rclone/pull/7410/commits/6109b14aa34c60649c92741c85164a1f2f3fe9ac.) Any chance you'd like to beta test it? 😃 

The underlying issue was the same one as `--track-renames` -- deletes being handled in separate operations (not using sync). Technically, `--backup-dir` did work before if you were overwriting an existing file, but not if deleting (or renaming) a file. After #7410 it will work for everything, as one would expect.

acegene:
> Technically, --backup-dir did work before if you were overwriting an existing file, but not if deleting (or renaming) a file. After https://github.com/rclone/rclone/pull/7410 it will work for everything, as one would expect.

Thanks for clarifying this.

> I have great news for you -- this will be fixed by https://github.com/rclone/rclone/pull/7410 (specifically https://github.com/rclone/rclone/commit/6109b14aa34c60649c92741c85164a1f2f3fe9ac.) Any chance you'd like to beta test it? 😃

[...]

PR #7410, opened 2023-11-08
bisync: logger, march, and several big bug fixes

### LINKED ISSUE #4854: New `--fix-case` flag to workaround case insensitive filesystem (macOS/Windows)
When syncing to a case sensitive file system, it would be nice if rclone could rename files/directories to match the case on the source.

Currently this isn't possible - rclone views `HELLO.txt` and `hello.txt` on a case insensitive file system as being the same file and skips over them.

This could be controlled with a new flag `--fix-case` which would only be supported on backends which implemented the `Move` and maybe `MoveDir` optional methods.

`--fix-case` would likely rename the source to the same name as the destination if it was different. This would mean it would make the unicode normalisation the same also.

See the forum for more info: https://forum.rclone.org/t/force-rclone-to-respect-case/20794/5

### LINKED ISSUE #5676: bisync: handle volatile changes consistently

### LINKED ISSUE #5690: bisync: implement `--track-renames`
## What problem are you are trying to solve?

bisync does not honor `--track-renames` !

## How do you think rclone should be changed to solve that?

bisync should honor `--track-renames` ?

Also see https://github.com/cjnaz/rclonesync-V2/issues/71

## How to use GitHub

### LINKED ISSUE #5696: bisync: fix handling of Google Docs
### Snippet from https://github.com/cjnaz/rclonesync-V2/issues/1

The problem comes up if we attempt to upload a modified exported document, with the same name,
back to Google Drive: "Failed to copy: can't update a google document". Similarly, deleting the
local copy and then rclone syncing local to Drive: results in "Couldn't delete: can't delete a google
document".

Proposed behavior:

A new --export-google-docs switch will export (copy to Local) every found Google document as
_export.. The final rclone sync will put a copy of the exported document on Google
Drive as well. The date-stamp on the file will match the original Google document.

Later, with --export-google-docs, If the native Google document timestamp is newer, and the
existing exported file is unchanged (relative to the prior sync), then the document will
be exported again, replacing the current _export file.

If the exported file is changed (newer) then it will be synchronized as usual/normal, but the
native Google document cannot be updated. Effectively, the file version and the native version
are out of sync.

If the exported file had previously been changed, and now the native Google document is changed,
then --export-google-docs will blast/replace the file version. User beware. There is no reliable
way to detect that the file version and the native Google document version have both been
modified. The user is advised to rename the modified file version if he/she plans to subsequently
edit the native Google document via the web interface.

If --export-google-docs switch is not set then all Google document files (as indicated by
file size -1) will be ignored. Previously exported documents (which are now regular files)
will be synchronized as usual.

### Other comments

As `@ncw` noted at https://github.com/rclone/rclone/pull/5164#discussion_r688914855
we should probably consider other things with `size < 0` (eg Google Photos) which will have that too.

As @cjnaz noted in his [commit review](https://github.com/rclone/rclone/commit/935559fbe9cf83b9b755036916b267c4c152378a#commitcomment-57724387), the log message `Ignoring incorrect line: "- -1 - - ... GoogleDoc.docx` is rather vague and should be improved. The commented dump from [bisync.md](https://github.com/rclone/rclone/blob/f9ada52400307c813c1aa2be222e75c7a087e0d1/docs/content/bisync.md#google-doc-files) was:
```
2021/05/11 08:23:15 INFO  : Synching Path1 "/path/to/local/tree/base/" with Path2 "GDrive:"
2021/05/11 08:23:15 INFO  : ...path2.lst-new: Ignoring incorrect line: "- -1 - - 2018-07-29T08:49:30.136000000+0000 GoogleDoc.docx"
2021/05/11 08:23:15 INFO  : Bisync successful
```

### References

- https://github.com/cjnaz/rclonesync-V2/issues/1
- comment by @cjnaz w.r.t google docs in https://github.com/rclone/rclone/files/7314114/bisync.md
- https://github.com/rclone/rclone/pull/2479
- https://github.com/rclone/rclone/pull/2411
- https://github.com/rclone/rclone/issues/2893
- https://github.com/rclone/rclone/issues/3986

### How to use GitHub

### LINKED ISSUE #7270: bisync: handle unicode normalization consistently
Bisync sometimes normalizes NFD to NFC and sometimes does not, causing errors in some scenarios (particularly for users of macOS). See https://forum.rclone.org/t/bisync-question/41434/12

Here is a bare-minimum case to reproduce:
```
% rclone touch NFD/ö 
% rclone touch NFC/ö
% rclone bisync NFD NFC --resync
2023/09/02 18:23:08 NOTICE: bisync is EXPERIMENTAL. Don't use in production!
% rclone bisync NFD NFC         
2023/09/02 18:23:17 NOTICE: bisync is EXPERIMENTAL. Don't use in production!
2023/09/02 18:23:17 ERROR : -          Path1 file not found in Path2       - ö
2023/09/02 18:23:17 ERROR : -          Path2 file not found in Path1       - ö
2023/09/02 18:23:17 ERROR : Bisync critical error: path1 and path2 are out of sync, run --resync to recover
2023/09/02 18:23:17 ERROR : Bisync aborted. Must run --resync to recover.
% 
```

As background context: most of the world uses NFC, but macOS prefers NFD for some reason, and will sometimes ([but not always](https://eclecticlight.co/2017/04/06/apfs-is-currently-unusable-with-most-non-english-languages/)) convert NFC filenames to NFD invisibly. (See also: #7072)

Rclone's `copy` and `sync` commands (including the ones used by `bisync` under-the-hood) automatically normalize to NFC, unless `--no-unicode-normalization` is specified. This is why you can usually `rclone sync` between mac and non-mac without issue -- rclone will recognize the NFC and NFD versions as the same filename.

The trouble is that there are three places where bisync does NOT do this:

1. When building the list of files that need to be transferred during `--resync`
2. When building the list of deltas during a non-resync, but only for so-called ["Unusual sync checks"](https://tip.rclone.org/bisync/#unusual-sync-checks)
3. When comparing Path1 to Path2 during `--check-sync`

1 and 2 end up being (mostly) harmless, because although it will try to copy the NFC and NFD versions as different files, the underlying `copy` function it calls is smart enough to realize they are the same file. It will therefore leave the files alone (if already equal) or update one of them (if not equal), but never create a duplicate.

3 is where the problem is most visible. Because `--check-sync` never normalizes, it thinks the NFC and NFD versions are different files. It therefore thinks something went wrong during the bisync run, because at the end of the run the Path1 file list does not exactly match Path2. Accordingly, it throws a critical error. (`--check-sync` is true by default, but can be overridden with `--check-sync=false`.)

The long-term solution is probably that `bisync` should support normalization in the same manner as `sync` (i.e. always normalize when comparing between paths, unless `--no-unicode-normalization` has been specified.) In the mean time, as a short-term solution, the issue can be bypassed by using `--check-sync=false`.

Also worth noting: in https://github.com/rclone/rclone/issues/5679#issuecomment-1689295692 we are contemplating whether `--check-sync` should perhaps just be replaced by outsourcing it to the more robust [rclone check](https://tip.rclone.org/bisync/#concurrent-modifications), which duplicates a lot of the same functionality, but does it better. That may ultimately be the answer here too, at least for item 3 (`--check-sync`). But I think items 1 and 2 do still need to be addressed regardless, as they are (at best) misleading to users (for instance, dry runs indicate that files will be transferred when they are not in fact new/changed and won't really be transferred.)

One more note: it seems that macOS will accept NFC filenames when created via terminal commands, but not when created via apps or Finder. This is why we don't tend to see these issues at first with files synced from cloud to mac via rclone, but the same files could present issues later if edited by an app or Finder. (This is also why `rclone touch NFC/ö` in the above repro will work, even on mac.)

#### How to use GitHub

### LINKED ISSUE #7282: Sync/Copy/Move: write list of transferred/deleted files
This is a feature request for the ability to access a complete list of files transferred/deleted by Sync/Copy/Move, as well as their basic attributes and any errors encountered. This is similar to [`rclone check`](https://rclone.org/commands/rclone_check/)'s ability to output lists for `MissingOnDst`, `MissingOnSrc`, etc.

While this would be a nice feature in general for many reasons, I am primarily interested because it's currently the blocking issue for [Bisync's biggest bug](https://github.com/rclone/rclone/issues/5676).

<details open>
<summary>

 ### Prior discussion (from #6797)</summary>

@nielash:
> Just wanted to mention, in case it's related to the work you're planning here, that it would be really helpful for [future `bisync` work](https://github.com/rclone/rclone/issues/5676) to have an easy way to access a complete list of transfers. Similar to [`accounting.Stats(ctx).Transferred()`](https://github.com/rclone/rclone/blob/25703ad20ed34a8990fc02913e17a74692050d21/fs/accounting/stats.go#L492-L506), but that one is limited to `MaxCompletedTransfers` (100) and doesn't include `modtime` or `hash`. My hope is to use this to avoid having to re-list each fs at the end of a bisync run to find out what happened during the `syncCopyMove` operations, as re-listing is more expensive and [error-prone](https://github.com/rclone/rclone/issues/6841).
>
> If this maybe already exists and I'm missing it, please let me know 😃 I see how to access the total number of transfers/errors, but we would need a list to reconcile it with the initial directory listings to build the "after" snapshot. If keeping the full transfer list in memory is a concern, we could potentially work with just a list of errored transfers (and infer that anything not on this list was successful and now matches what we passed in.) But this is not entirely foolproof, as it leaves open the possibility that files changed in between the listing and copying.

@ncw:

> That isn't quite what is under discussion here, but what I think you want is something like we have for `operations.Check` where you can pass in `io.Writer`s to have the file names written to.
> 
> ```
> type CheckOpt struct {
> 	Fdst, Fsrc   fs.Fs     // fses to check
> 	Check        checkFn   // function to use for checking
> 	OneWay       bool      // one way only?
> 	Combined     io.Writer // a file with file names with leading sigils
> 	MissingOnSrc io.Writer // files only in the destination
> 	MissingOnDst io.Writer // files only in the source
> 	Match        io.Writer // matching files
> 	Differ       io.Writer // differing files
> 	Error        io.Writer // files with errors of some kind
> }
> ```
> 
> If you read the docs for [rclone check](https://rclone.org/commands/rclone_check/) you'll see the bits about outputting file names to files. People have asked for exactly this for sync/copy/move and I think this internal inteface would suit your requirements too.
> 
> I should probably change the `io.Writer` into a `func(remote string) error` to explicitly say it will be called with a path each time. We'd need a way of signalling high level retries also as the functions potentially get called each high level retry.

@nielash:

> Yes, this would be great! (I used this recently in https://github.com/rclone/rclone/commit/5ca61ab70543abf740999b874ed65ec3ffc890cf.) For bisync purposes I think we'd also need the `size`, `modtime`, and `hash` in addition to the list of names (at least to the extent that `syncCopyMove` already has this info -- and I think it does, as that's how it knows if a file is "corrupted on transfer".)</details>

In addition to the above, it would also be important to know whether each operation was a copy or delete, and whether it had an error. The end goal is to update the initial listing, which uses this format:

https://github.com/rclone/rclone/blob/b750c50bfd4440e88cc66c718ca20e0892f6404f/cmd/bisync/listing.go#L27-L32

See also: 

- https://github.com/rclone/rclone/issues/3933
- https://github.com/rclone/rclone/issues/5676
- https://github.com/rclone/rclone/issues/6841
- [Forum discussion](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=5.%20Final%20listings%20should%20be%20created%20from%20initial%20snapshot%20%2B%20deltas%2C%20not%20full%20re%2Dscans%2C%20to%20avoid%20errors%20if%20files%20changed%20during%20sync)
- A [recent user post](https://forum.rclone.org/t/bisync-missed-syncing-changed-file-no-error-no-warning/41174) about this bug

#### How to use GitHub

### LINKED ISSUE #7332: bisync: performance
### setup

11,265 files - 290.743MiB (90MB in ~5 files the rest are smaller and more equal in size)

Results the same regardless of Wasabi or iDrive (iDrive typically performs a little faster in every test but only ~5-10% on average)

Remote is using crypt and base32768 encoding. 

BW: ~1.5MiB UP ~5MiB DOWN

### default/common flags

After testing best performing filters are: --transfers 64 --checkers 128 --fast-list

Along with this (--filter-from) or (--filters-file) is also used

e.g. --transfers 64 --checkers 128 --fast-list --filter-from/--filters-file ..

### check

~8s on average and between 6.8s to 9.6s

### ls

(using `time` as no info given with -v)

ls: 0m4.088s
lsl: 7m36.975s

### sync

upload:
	initial: 3m57.0s
	subsequent: 18.6s
	
download:
	initial: 2m9.0s
	subsequent: 17.8s

### copy

upload:
	initial: 4m0.0s
	subsequent 18.6s
	
download:
	initial: 2m6.5s
	subsequent: 16.8s

### bisync

These are some of the best and would say are usually a little longer by upto 5 mins this is likely when resync is run when data is already present vs copying the data itself.. There are no changes made to files when moving between a --resync and normal run.  

defaults:
	resync: 10m47.0s
	normal: 6m44.7s

defaults with ignore checksum: (--ignore-listing-checksum --ignore-checksum)
	resync: 10m36.5s
	normal: 6m31.9s

### overview

So as you can see bisync is orders of magnitude slower than `copy` or `sync` . `ls` is incredibly fast as is `sync` and `copy` in both directions (e.g. uses up almost 100% of bandwidth)

I only ran the test of `lsl` and `ls` at the very end and now thinking that bisync relies on lsl as this so incredibly slow (as I believe bisync does not use checksum so would need `lsl` info to compare?).

I can do a local to crypt (local) bisync in 15s (--resync) and 8.7s for subsequent. 

Is there something I am doing wrong? If not, are there plans/options available to rclone to improve the speed of bisync so that it is more in line with 2 x sync/copy commands (e.g. one upload, one download) or is this a limitation that is unlikely to be overcome? Also, if this is due to lsl taking so long is there a way to speed up lsl? 

Thanks to you all for your hard work and a special thank you @nielash for all you've been doing on bisync and continue to do. I as many others massively appreciate it and when this tool is out of experimental stages it will be a game changer for so many. If I had any free time or money right now I would help out but unfortunately the best I can do is give you some benchmarks and user feedback.

Thanks

nielash:

A suite of several related fixes tackling bisync's biggest pain points.

Highlights:
- bisync: merge copies and deletes, support `--track-renames` and `--backup-dir` -- fixes #5690 and #5685
- operations: add logger to log list of sync results -- fixes #7282
- sync: report list of synced paths to file -- see #7282
- bisync: generate final listing from sync results, not relisting - fixes #5676
- bisync: rollback listing on sync error
- bisync: generate listings concurrently with march -- fixes #7332
- sync: `--fix-case` flag to rename case insensitive dest -- fixes #4854
- bisync: handle unicode and case normalization consistently - fixes #7270
- bisync: optimize `--resync` performance -- partially addresses #5681
- bisync: add support for Google Docs - fixes #5696

Yes. See [this post](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636) and the individual issues linked above.

#### Checklist


nielash:
(I think the windows error is unrelated, possibly a side-effect of [recent Go security fix](https://github.com/golang/go/commit/9e933c189ca3a84f12995b3c799364a06abc4376).)

nielash:
@ncw Thought I'd bring https://github.com/rclone/rclone/pull/7410/commits/f65dc5f7191fbebbaf889c45331c4e181ade980e to your attention related to your [fix from yesterday](https://github.com/rclone/rclone/commit/acf1e2df84a350b7a86d7672d749dfb1ba090a44) regarding the `filepath.VolumeName` change in go1.21.4. The same issue appeared to be breaking `NewFs` on Windows in a way that wasn't previously showing up on tests. (I added one here that catches it.)

cmd/lsf/lsf_test.go
```diff
@@ -141,7 +141,7 @@ file3
 	items, _ := list.DirSorted(context.Background(), f, true, "")
 	var expectedOutput string
 	for _, item := range items {
-		expectedOutput += item.ModTime(context.Background()).Format("2006-01-02 15:04:05") + "\n"
+		expectedOutput += item.ModTime(context.Background()).Format("2006-01-02 15:04:05.000000000") + "\n"
```
ncw, line 144: This is a potentially backwards incompatible change.

I think `lsf` attempts to show the correct number of digits for the Precision in the Fs.

cmd/sync/sync.go
```diff
@@ -3,22 +3,142 @@ package sync
...
 	"github.com/rclone/rclone/fs/config/flags"
+	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/rclone/rclone/fs/sync"
 	"github.com/spf13/cobra"
 )
 
 var (
 	createEmptySrcDirs = false
+	combined           = ""
```
ncw, line 22: I would just make a LoggerOpt here pass that directly to `AddLoggerFlags`

backend/drive/drive.go
```diff
@@ -870,6 +870,11 @@ func (f *Fs) getFile(ctx context.Context, ID string, fields googleapi.Field) (in
 	return info, err
 }
 
+// Options returns remote-specific options for reference by commands
```
ncw, line 873: I don't like this - this breaks the backend encapsulation and will mean you need to `import "github.com/rclone/rclone/backend/drive` to use it which means that you ignore the users backend choices if they are embedding rclone.

The proper way to do this is to add a new method or a new Feature flag.

However I think you can do something different.

Detect whether you ever receive a file with size < 0 before giving your warning with a `sync.Once`.

This will mean that you catch the other cases because files of unknown length can happen in other backends not just Google Drive

- google drive
- google photos
- s3/gcs with Content-encoding Gzip files

I think this PR should be called `bisync: add support files with unknown length, including Google Docs - fixes #5696`

Generally in this commit `s/Google doc/files of unknown size/g;` :-)

cmd/bisync/deltas.go
```diff
@@ -264,6 +264,7 @@ func (b *bisyncRun) applyDeltas(ctx context.Context, ds1, ds2 *deltaSet) (change
 	// check is a read-only operation by its nature, so it's already "dry" in that sense.
 	ctxNew, ciCheck := fs.AddConfig(ctx)
 	ciCheck.DryRun = false
+	ciCheck.NoUpdateModTime = true
```
ncw, line 267: Please write why in the commit message!  (I can't comment on the commit messages)

backend/local/local.go
```diff
@@ -1447,6 +1447,10 @@ func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
 	if runtime.GOOS == "windows" {
 		s = filepath.ToSlash(s)
 		vol := filepath.VolumeName(s)
+		if vol == `\\?` {
+			// `\\?\C:`
+			vol = s[:6]
```
ncw, line 1452: That fix looks like it is an index error waiting to happen! 

Maybe wrap it in `if len(s) >= 6`?

Its a nasty fix - hopefully we can get rid of it soon!

cmd/bisync/cmd.go
```diff
@@ -50,6 +50,12 @@ type Options struct {
 	Resilient             bool
 	TestFn                TestFunc // test-only option, for mocking errors
 	Retries               int
+	GUI                   bool
```
ncw, line 53: Being able to show status to the user is a great idea.

However I'd like to consider the other operations too like sync etc.

At the moment you can build a pretty good gui with javascript using the rc.

So I think what you should do is take this commit off the series and stash it on a branch, then make an issue for discussion.

ncw (commented):
Wow that was a lot of commits! It contains some excellent work - lovely logging features, loads of bisync fixes and features - well done :-)

I'd like to release v1.65 within the next week or so, so how much of this do you want in that?

I have put some comments inline but there isn't anything serious in there. I'm happy to work on re-arranging the flags later. I don't want to merge the GUI now, but I will consider it for later.

So have a think about which commits you want in v1.65

From my point of view I think we should try to get the logging stuff in, and as many bisync fixes as we can.

PS smaller PRs in future please :-)

nielash:
Thanks! And sorry about the massive number of commits! 😅  Nearly all of them depend on earlier ones, so I wasn't sure how I could separate them into different PRs. 

I can take out the GUI for now, that's fine. As for the others, I'd like to get in as many as you'll allow! I can make some quick revisions here based on your comments. (I'll add some responses inline.)

backend/drive/drive.go
nielash: The main reason for not using size is that Directories also have a size of -1 (if using `--create-empty-src-dirs`). But we do have another way of detecting Dirs, so I think actually I can make this work. 

I had also wanted to not bother users with this warning if they're already using `--drive-skip-gdocs` (which we can't otherwise check). But I think you are right that we can assume they are not, if we see non-dir listings with size of -1.

fs/operations/operations.go
```diff
@@ -2231,3 +2269,326 @@ func SkipDestructive(ctx context.Context, subject interface{}, action string) (s
...
+		}
+	}
+	// should only make it this far if it's TransferError and both src and dst are nil
+	winner.Side = "none"
+	winner.Err = fmt.Errorf("unknown case -- can't determine winner. %v", err)
+	fs.Debugf(winner.Obj, "%v", winner.Err)
+	return winner
+}
+
+// DirObjInfo implements fs.ObjectInfo so that logger can log dirs
```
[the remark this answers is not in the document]
nielash: There is probably a better way -- but the problem I was trying to solve is that `LoggerFn` takes:
```
type LoggerFn func(ctx context.Context, sigil Sigil, src fs.ObjectInfo, dst fs.ObjectInfo, err error)
```

and `fs.Directory` does not satisfy `fs.ObjectInfo`. `fs.Dir` also does not work... I get:

```
cannot use dir (variable of type "github.com/rclone/rclone/fs".Dir) as "github.com/rclone/rclone/fs".ObjectInfo value in argument to s.logger: 
"github.com/rclone/rclone/fs".Dir does not implement "github.com/rclone/rclone/fs".ObjectInfo (missing method Fs)compiler[InvalidIfaceAssign]
```

cmd/lsf/lsf_test.go
nielash: I don't think it does, unless there's an option I'm missing:

```
rclone lsf a --format stp --separator ' '
5 2023-11-17 01:33:40 hi copy.txt
5 2023-11-17 01:30:58 hi.txt
```
vs. `lsl`:
```
rclone lsl a                            
        5 2023-11-17 01:33:40.782855000 hi copy.txt
        5 2023-11-17 01:30:58.202211000 hi.txt
```

You are right that it could potentially break things if users are scripting with the output and expecting a certain precision (although I'd have guessed they would use `lsjson` instead.)

backend/local/local.go
nielash: Will do! (And agreed.)

cmd/bisync/cmd.go
nielash: Ok, I'll remove it for now and make an issue for discussion. I do think there are several bisync-specific things that wouldn't make sense sharing a template with `sync` (for instance, the Path1 / Path2 columns with links to their respective listing files.) ([Demo](https://github.com/nielash/rclone/blob/bisync-logger/docs/static/img/bisync_gui_example.png)) It also displays a warning if a bisync was locked out by a prior run -- which is not a thing in sync.

The rc is very nice but I think it's a drawback that the server needs to be running in order to use it, vs. a static HTML file which can live anywhere (even another rclone remote). I also think the rc adds complexity for new users who want to get up and running quickly. But I'll move it to a branch for now for further discussion!

nielash:
@ncw Here's a revision. I think I followed all of your instructions, but let me know if I misunderstood anything! Also, don't hate me but I added a commit to fix integration tests, and two others that were pre-requisites for that. 

I also rebased and resolved conflicts.

fs/operations/operations.go
```diff
@@ -2231,3 +2269,326 @@ func SkipDestructive(ctx context.Context, subject interface{}, action string) (s
...
+	DirSlash   bool
+	// Recurse   bool
+	HashType  hash.Type
+	FilesOnly bool
+	DirsOnly  bool
+	Csv       bool
+	Absolute  bool
+}
+
+// AddLoggerFlags adds the logger flags to the cmdFlags command
```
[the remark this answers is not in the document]
nielash: Made this change (in https://github.com/rclone/rclone/pull/7410/commits/354d81aa95cec98cd3d26dd893b4eebc34a7181d)

fs/operations/operations.go
```diff
@@ -2231,3 +2269,326 @@ func SkipDestructive(ctx context.Context, subject interface{}, action string) (s
 	}
 	return skip
 }
+
```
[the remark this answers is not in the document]
nielash: Agreed! Made this change.

cmd/sync/sync.go
nielash: Made this change.

backend/drive/drive.go
nielash: Made this change.

backend/local/local.go
nielash: Made this change.

cmd/bisync/deltas.go
nielash: This is made moot by a different change, so I removed it.

fs/operations/operations.go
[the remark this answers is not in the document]
nielash: I didn't change this, but I'm happy to if you can point me towards a better solution.

cmd/lsf/lsf_test.go
nielash: I didn't change this, but I can if you think it's best.

Alternately I'm happy to implement what you described (variable precision per Fs), although that would still be a potentially backwards-incompatible change.

acegene:
> bisync: merge copies and deletes, support --track-renames and --backup-dir -- fixes https://github.com/rclone/rclone/issues/5690 and https://github.com/rclone/rclone/issues/5685

Regarding mostly the above (and to a lesser extent any bisync changes), I am trying to beta test. So far im using the version specified in [tag/v1.65.0-beta.7496.004c12643.bisync-logger-alpha](https://github.com/nielash/rclone/releases/tag/v1.65.0-beta.7496.004c12643.bisync-logger-alpha), any chance you could generate a more up-to-date delivery? Regardless, thanks for your work!

nielash:
@acegene thanks! If you're comfortable [installing from source](https://rclone.org/install/#source), you can [download](https://github.com/nielash/rclone/archive/refs/heads/bisync-logger.zip) the latest version from the [nielash:bisync-logger](https://github.com/nielash/rclone/tree/bisync-logger) branch and compile it. Otherwise I'm happy to upload some updated binaries later today or tomorrow!

nielash:
@acegene Here are some updated binaries if you'd like to test it out. Feedback is welcome :) 
[Bisync Feature Preview 2023-11-23](https://github.com/nielash/rclone/releases/tag/v1.65.0-beta.7552.7faf0e787.bisync-logger-alpha-2)

crocinsocks:
@nielash you are a legend, thank you!! Normal sync without changes takes around 30s whereas before it could take anywhere between 10-20 minutes.. Been using for a few days now without issue.

Reading the comments of --resilient changes I wasn't sure whether this would mean that an exited sync would be able to recover? From my tests it does not however I not using the changes from 3 days ago rather [bisync-logger-alpha-2](https://github.com/nielash/rclone/tree/v1.65.0-beta.7552.7faf0e787.bisync-logger-alpha-2 ). Am I wrong in thinking this and therefore should I open an issue specifically for this case? n.b. this is so a computer can be shutdown / rclone exited without breaking and requiring a --resync..

Also, just off our chats in the past should I open and start an issue for conflict renaming so it is not so ambiguous?

Thank you  🙏

nielash:
> @nielash you are a legend, thank you!! Normal sync without changes takes around 30s whereas before it could take anywhere between 10-20 minutes.. Been using for a few days now without issue.

So glad to hear it is working well for you! 😃 

> Reading the comments of --resilient changes I wasn't sure whether this would mean that an exited sync would be able to recover?

It depends on the reason for the exit. `--resilient` allows retrying when bisync has chosen to abort itself due to safety features such as failing `--check-access` or detecting a filter change. It does not currently cover external interruptions such as a user shutting down their computer in the middle of a sync. That said, it's something I would like to add support for, so yes, please open an issue!

> Also, just off our chats in the past should I open and start an issue for conflict renaming so it is not so ambiguous?

I was actually just sketching out a plan for this the other day! I'll open an issue today and tag you so we can discuss it.

nielash:
Hi @ncw, question for you -- I am working on some other bisync issues now and would like to submit a new PR for them since this one is too big already 😅  However, they build on the work I did here, so I'm not sure if there's a way to avoid duplicating all of these commits, as I don't think GH supports "dependent" PRs yet. Is there a better way to handle this?

acegene:
> Hi @ncw, question for you -- I am working on some other bisync issues now and would like to submit a new PR for them since this one is too big already 😅 However, they build on the work I did here, so I'm not sure if there's a way to avoid duplicating all of these commits, as I don't think GH supports "dependent" PRs yet. Is there a better way to handle this?

I could help with this point. Take a look at [this SO answer](https://stackoverflow.com/a/70178279). Essentially, if you start from this PR/branch (lets call it B1), you could create B2 that depends on it. Locally you could create the branch with:
```
git fetch
git checkout origin/B1
git checkout -b B2 # creates B2 with all the changes that have been pushed to B1
## Add any work you want to B2, then push upstream. then select B1 to merge into so it says "<AUTHOR> wants to merge X commits into B1 from B2".
```

Notes:
- If you have already started the work of B2 on some other branch, lets call it B2_beta, after creating B2 using the above, you could try to cherry-pick the new commits from B2_beta (those not present on B1) onto B2
- You know the process has been successful if the "duplicated" commits have matching commit hashes on each branch
- If new commits are pushed to B1, that you want in B2, you will need to do a rebase of B2 onto origin/B1

I am quite familiar with this type of process, let me know if you have questions or would like to hop in a discord call or something similar.

fs/operations/operations.go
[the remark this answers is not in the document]
ncw: LoggerFn should probably be taking an `fs.DirEntry` which is a superset of fs.Object and fs.Directory. If you want more information that is object only then you'd do a type assertion to `fs.Object`

cmd/lsf/lsf_test.go
ncw: This is the code that I was thinking of

https://github.com/rclone/rclone/blob/2f5685b40562cf764da673543245ddb13c049a4a/fs/operations/lsjson.go#L50

Maybe this is only used for `lsjson`?

I think we should try not to change the output of `lsf` unless it is guarded with an option.

fs/operations/operations.go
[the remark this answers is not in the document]
nielash: That worked perfectly -- thank you 🙂 Will send a revised version shortly.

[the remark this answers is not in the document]
nielash: Fixed in revised version.

cmd/lsf/lsf_test.go
nielash: Got it. Yeah, looks like this only used by `lsjson`, as `lsf` does not use RFC3339. I have guarded it with a flag in the new version, so that the default is unchanged and nothing should break.

ncw (approved):
Wow this is a monster :-)

I think we should merge it now so we can unblock the dev process. We've got time to fix up breakage before the next release.

I'll approve it and you can merge it!

Thank you and many apologies for the backlog.

nielash:
Awesome, thank you!! I will do the honors shortly and slay this "monster" 😆

PR #7416, opened 2023-11-12
Update Docs to show SMB remote supports modtime.md

asdffdsazqqq:
    Document that SMB remote supports modtime.

https://forum.rclone.org/t/how-to-access-smb-samba-with-rclone/42754


docs/content/overview.md
```diff
@@ -99,6 +99,8 @@ hash](https://static.hidrive.com/dev/0001).
 It combines SHA1 sums for each 4 KiB block hierarchically to a single
 top-level sum.
 
+FTP supports modtimes for the major FTP servers, and also others
+
```
albertony, line 103: I don't think you intended to include these lines?

ncw:
I dropped the extra bit from the commit and I've merged this now thank you :-)

PR #7429, opened 2023-11-18
Add check command option to consider modtime in addition to size and hash

albertony:

Existing `rclone check` command compares `size` and `hash`. This PR makes it optionally also consider `modtime`, in addition to those two.

The copy command already considers the `modtime`, and if different then it updates it even if everything else is equal. So I think making the check command also consider `modtime`makes it more in line with what is considered "different" by copy operations. Also, it was not previously possible to force-check all of `size`, `modtime` and `hash`, but one could run a copy to check/update timestamps, followed by a check command which would verify timestamps.

I know this is sort of a can of worms, with all the different options like `--size-only`, `--ignore-size`, `--checksum`, and it is not easy to make it clear for users which properties are actually considered when - I'm a little confused myself! And therefore, to try to avoid adding to the confusion with the new flag, I called it `--consider-modtime`. I'm not overly satisfied with that, but I found `--modtime` could be confusing as it could indicate **only** `modtime` is considered, so I tried to come up with something that makes it clear that it checks `modtime` **in addition** to everything else (`size` and `hash`, or just `size` if `--size-only`, or just hash if `--ignore-size`).

The original purpose of the check command was probably an integrity/corruption check, and then considering modtime may not be a natural part of this. And in future, perhaps we would want to include more metadata in the comparison? If the check command should be kept as dedicated integrity check, then maybe we should create a separate `rclone compare` command

The following show example output from a check command:

If size is different:

```
2023/11/18 15:09:43 DEBUG : change.log: size = 1344 (Local file system at //?/C:/Tools/npp)
2023/11/18 15:09:43 DEBUG : change.log: size = 1345 (Local file system at //?/C:/Tools/npp2)
2023/11/18 15:09:43 ERROR : change.log: size differ
```

If size is equal but hash is different:

```
2023/11/18 14:56:40 DEBUG : change.log: md5 = da48501e39999cb4abab66e20dbd4101 (Local file system at //?/C:/Tools/npp)
2023/11/18 14:56:40 DEBUG : change.log: md5 = 2578b5f7d1e4a710234bd9a0fcf37dad (Local file system at //?/C:/Tools/npp2)
2023/11/18 14:56:40 ERROR : change.log: md5 differ
```

If size and hash are equal, and running with `--consider-modtime` (if not, by default, the files are considered equal):
```
2023/11/18 15:14:25 DEBUG : change.log: modtime = 2023-11-18 15:14:19.536269 +0100 CET (Local file system at //?/C:/Tools/npp)
2023/11/18 15:14:25 DEBUG : change.log: modtime = 2023-10-14 18:00:26.9284192 +0200 CEST (Local file system at //?/C:/Tools/npp2)
2023/11/18 15:14:25 ERROR : change.log: modtime differ
```

#### Checklist


nielash:
Absolutely love this! 

If I may make a suggestion: I really liked [ncw's idea here](https://forum.rclone.org/t/compare-checksum-in-addition-to-not-instead-of-modtime/36724/2?u=nielash) of having a configurable sync key like `--compare size,modtime,checksum` (or perhaps `sth` like in `lsf`?) The new [`fs.Bits`](https://github.com/rclone/rclone/blob/master/fs/bits.go) could be a simple way to implement this without breaking any legacy flags (for example, if `--checksum` is set, it could be mapped internally to set `size,checksum`)

ncw:
I think this is a nice idea.

I think we should try to use the opportunity to unify the check and sync commands. The defaults should be different but you should be able to use the same flags to control them. 

I think what you are saying here is that `--consider-modtime` is effectively the default for `sync`/`copy`/`move` but as `--checksum` is the default for `check` we need a way of reversing that?

> If I may make a suggestion: I really liked [ncw's idea here](https://forum.rclone.org/t/compare-checksum-in-addition-to-not-instead-of-modtime/36724/2?u=nielash) of having a configurable sync key like `--compare size,modtime,checksum` (or perhaps `sth` like in `lsf`?) The new [`fs.Bits`](https://github.com/rclone/rclone/blob/master/fs/bits.go) could be a simple way to implement this without breaking any legacy flags (for example, if `--checksum` is set, it could be mapped internally to set `size,checksum`)

I was imagining that as an ordered list so you'd try `size`, `modtime` then `checksum`

> I had a vague idea to make the sync key configurable. At the moment it is size,modtime by default, size, checksum if --checksum is in use and size if --size-only is in use.

A quick inventory of rclone's flags - I think all of these affect the "sync key" in some way!

```
  -c, --checksum                                    Check for changes with size & checksum (if available, or fallback to size only).
      --ignore-size                                 Ignore size when skipping use mod-time or checksum
  -I, --ignore-times                                Don't skip files that match size and time - transfer all files
      --no-update-modtime                           Don't update destination mod-time if files identical
      --refresh-times                               Refresh the modtime of remote files
      --size-only                                   Skip based on size only, not mod-time or checksum
  -u, --update                                      Skip files that are newer on the destination
```

albertony:
> I think what you are saying here is that --consider-modtime is effectively the default for sync/copy/move but as --checksum is the default for check we need a way of reversing that?

Yes, I think so too... 😄 

> I think we should try to use the opportunity to unify the check and sync commands.

Sure, that would ideal, I do agree with that. However it is a potentially large refactor, I fear. At least if doing this for sync/copy/move operations, but also to some degree if only doing it for the check command in first iteration I think.

PR #7448, opened 2023-11-22
#7444 enable WebDAV methods with CORS

### LINKED ISSUE #7444: WebDAV partly unusable when CORS applied
This is a follow-up to #5078. First of all, thank you very much for implementing that request!
As my main goal was to access WebDAV via rclone from the browser (a PWA), that is unfortunately still not completely possible. Let's try to dig into details.

It is still not 100% clear to me what/where the problem lies, as the messages are different in different browsers.
The test case can be accessed from https://cashier-ii.alensiljak.eu.org/webdav-demo
This is the (PWA) page that issues the requests.

I'll provide here some general info that does not fit into the answers below.

As noted in #5078, these headers are included in the Response and I'm wondering if that is the source of the issue:
```
Access-Control-Allow-Origin  | *
Access-Control-Request-Method | POST, OPTIONS, GET, HEAD
```
This would explain the failure of PROPFIND. PUT also does not work.
There's a failure of POST, as well, but I suspect there is something wrong with how I'm using it (?).

N/A

Accessing the Rclone WebDAV server from a browser does not work as expected.

Initially I've tried a few WebDAV-specific libraries. These all use PROPFIND request, which seems to be blocked in this scenario.

After that, I've used the HttpClient directly, to issue HTTP commands. For the following commands:
- simple get (root)
- get a file
- propfind (root)
- post a file

```
GET http://localhost:8080

GET http://localhost:8080/rates.ledger

PROPFIND http://localhost:8080

POST http://localhost:8080/upload-test.txt
with string stream attached.
```
only the first two succeed. PROPFIND and POST fail. 
I may be using POST wrong. I'll adjust the request as we troubleshoot.

However, when other tools are used (non-browser, Total Commander on Android, curl, Windows Explorer), things seem fine. I believe the reason for this is that these do not care about CORS.

Let me know if you have questions or need clarifications. I'd be happy to help to try to narrow this down.

rclone v1.64.2
- os/version: Microsoft Windows 11 Pro for Workstations 22H2 (64 bit)
- os/kernel: 10.0.23570.1000 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.21.3
- go/linking: static
- go/tags: cmount

Windows 11, 64-bit

Local drive/folder, for this case.

```sh
rclone serve webdav . --allow-origin "*" -vv --log-file bug.log
```

```
2023/11/21 13:57:29 DEBUG : Creating backend with remote "."
2023/11/21 13:57:29 DEBUG : Using config file from "C:\\Users\\Alen\\AppData\\Roaming\\rclone\\rclone.conf"
2023/11/21 13:57:29 DEBUG : fs cache: renaming cache item "." to be canonical "//?/D:/tmp"
2023/11/21 13:57:29 INFO  : Local file system at //?/D:/tmp: poll-interval is not supported by this remote
2023/11/21 13:57:29 NOTICE: Warning: Allow origin set to *. This can cause serious security problems.
2023/11/21 13:57:29 NOTICE: Local file system at //?/D:/tmp: WebDav Server started on [http://127.0.0.1:8080/]
2023/11/21 13:57:43 INFO  : : 127.0.0.1:57907: Serving directory
2023/11/21 13:57:44 DEBUG : /rates.ledger: OpenFile: flags=O_RDONLY, perm=----------
2023/11/21 13:57:44 DEBUG : rates.ledger: Open: flags=O_RDONLY
2023/11/21 13:57:44 DEBUG : rates.ledger: >Open: fd=rates.ledger (r), err=<nil>
2023/11/21 13:57:44 DEBUG : /rates.ledger: >OpenFile: fd=rates.ledger (r), err=<nil>
2023/11/21 13:57:44 DEBUG : rates.ledger: ChunkedReader.openRange at 0 length 134217728
2023/11/21 13:57:44 DEBUG : rates.ledger: ChunkedReader.Read at 0 length 512 chunkOffset 0 chunkSize 134217728
2023/11/21 13:57:44 DEBUG : rates.ledger: ReadFileHandle.seek from 512 to 0 (fs.RangeSeeker)
2023/11/21 13:57:44 DEBUG : rates.ledger: ChunkedReader.RangeSeek from 512 to 0 length -1
2023/11/21 13:57:44 DEBUG : rates.ledger: ChunkedReader.Read at -1 length 541 chunkOffset 0 chunkSize 134217728
2023/11/21 13:57:44 DEBUG : rates.ledger: ChunkedReader.openRange at 0 length 134217728
2023/11/21 13:57:44 INFO  : /rates.ledger: GET from 127.0.0.1:57907
2023/11/21 13:57:46 INFO  : /: OPTIONS from 127.0.0.1:57907
2023/11/21 13:57:48 DEBUG : /upload-test.txt: OpenFile: flags=O_RDONLY, perm=----------
2023/11/21 13:57:48 DEBUG : /upload-test.txt: >OpenFile: fd=<nil>, err=file does not exist
2023/11/21 13:57:48 INFO  : /upload-test.txt: POST from 127.0.0.1:57907
2023/11/21 13:57:50 INFO  : Signal received: interrupt
2023/11/21 13:57:50 DEBUG : 7 go routines active
2023/11/21 13:57:50 INFO  : Exiting...
```

The log seems to indicate that there was no PROPFIND request at all. This may be blocked by the browser itself. It hints that this method might need to be explicitly allowed by the server.

I also don't see the GET request for "/".

#### How to use GitHub

alensiljak:
Fixes #7444

Enables the HTTP methods used with WebDAV protocol. It allows requests to rclone WebDAV issued from web browsers, which are currently blocked by CORS policy.

#7444

#### Checklist

### DIFF
--- a/lib/http/middleware.go
+++ b/lib/http/middleware.go
@@ -183,8 +183,8 @@ func MiddlewareCORS(allowOrigin string) Middleware {
 
 			if allowOrigin != "" {
 				w.Header().Add("Access-Control-Allow-Origin", allowOrigin)
-				w.Header().Add("Access-Control-Request-Method", "POST, OPTIONS, GET, HEAD")
 				w.Header().Add("Access-Control-Allow-Headers", "authorization, Content-Type")
+				w.Header().Add("Access-Control-Allow-Methods", "COPY, DELETE, GET, HEAD, LOCK, MKCOL, MOVE, OPTIONS, POST, PROPFIND, PROPPATCH, PUT, TRACE, UNLOCK")
 			}
 
 			next.ServeHTTP(w, r)
--- a/lib/http/middleware_test.go
+++ b/lib/http/middleware_test.go
@@ -323,8 +323,8 @@ func TestMiddlewareAuthCertificateUser(t *testing.T) {
 
 var _testCORSHeaderKeys = []string{
 	"Access-Control-Allow-Origin",
-	"Access-Control-Request-Method",
 	"Access-Control-Allow-Headers",
+	"Access-Control-Allow-Methods",
 }
 
 func TestMiddlewareCORS(t *testing.T) {


alensiljak:
Any tips about testing and documentation are welcome!

alensiljak:
I think this would be hard to test without a browser, considering it worked when using other clients, like Windows Explorer, Total Commander, Thunder Client, etc.

ncw:
There are some tests for this

https://github.com/rclone/rclone/blob/668711e43213bc9464c0b25c88621cd987898f08/lib/http/middleware_test.go#L324C36-L330

Which I suspect will fail! So you'll need to change them.

These just check for the existence of the correct headers - you could improve them if you wanted, though existence is likely good enough.

alensiljak:
Cool, will do!

alensiljak:
I've only updated the failing tests with the new headers. 
As mentioned, it makes no sense to test these methods without a browser as this case fails only then. Everything works when issued from a regular client.
Open for any suggestions.

ncw (approved):
This looks fine now - thank you :-)

It is tested as well as it was before - I think it is good enough.

PR #7452, opened 2023-11-24
dlna: also look at "Subs" subdirectory, don't swallow some subs, fix mime types

flokli:
Apparently it seems pretty common for subtitles to be put in a subdirectory called "Subs", rather than in the same directory as the media file itself.

This covers that usecase, by checking the returned listing for a directory called "Subs" to exist.

If it does, its child nodes are added to the list before they're being passed to mediaWithResources, allowing these subtitles to be discovered automatically.

Automatically find subtitles that live in a "Subs" directory alongside the video file, which seems fairly common.

Unknown

#### Checklist


flokli:
I added a test for the `mediaWithResources` function to test both the current behaviour, as well as the behaviour with the subtitles inside the `Subs` directory.

flokli:
I also tried this with an example of .idx/.srt files, and noticed the code was swallowing either one of .idx/.sub, and did always send text/srt as a mime type, which is only correct for .srt files. I added these to other individual commits that also extend the test cases. Happy to also move this to a followup PR, just let me know.

flokli:
Hmmh, curious, the test logs mention `application/x-subrip` rather than `text/srt`.

@dwalters (https://github.com/rclone/rclone/commit/6337cc70d320868f47995fdf89290dd845f9b47d), do you recall where `text/srt` comes from?

flokli:
I checked the mime database, and verified `application/x-subrip` again. The samsung TV here is apply with `application/x-subrip` or `text/srt`, so I don't think it matters, and I updated the test cases and the "mime database lite" we ship for some platforms.

flokli:
I dropped checking for specific mime types in the tests, it seems this is still too platform-specific. PTAL.

flokli:
poke @ncw :-) any opinion on that? I've been flying with this patch for a while now, but noticed this PR is still open. Would be nice to see it landing upstream.

cmd/serve/dlna/cds.go
```diff
@@ -130,6 +130,20 @@ func (cds *contentDirectoryService) readContainer(o object, host string) (ret []
 		return
 	}
 
+	// if there's a "Subs" child directory, add its children to the list as well,
+	// so mediaWithResources is able to find them.
+	for _, node := range dirEntries {
+		if node.Name() == "Subs" && node.IsDir() {
```
ncw, line 151: This should probably be a case insensitive compare so `strings.EqualFold`

ncw, line 151: Does this directory name need to be configured?

It certainly needs documenting though - a line in the docs about here maybe?

https://github.com/rclone/rclone/blob/70e8ad456fcc627f57c001aa82ac925bd9fb47a8/cmd/serve/dlna/dlna.go#L50

ncw (commented):
Sorry this didn't get reviewed before.

These look like nice changes. I put a small query inline but otherwise great - thank you :-)

cmd/serve/dlna/cds.go
flokli: They didn't mention the subtitle discovery yet at all, but yes, can add.

`Subs` is a convention I've seen so far quite recently, and so far nothing else seems to be commonly used. I don't think it makes a lot of sense to make this configurable.

flokli: Ok, will do.

flokli: Added a commit documenting the functionality.

flokli: Done

ncw (approved):

I'll merge now.

PR #7467, opened 2023-11-29
systemd: Fix detection and switch to the coreos package everywhere

darthShadow:

systemd: Fix detection and switch to the coreos package everywhere rather than having 2 separate libraries

https://forum.rclone.org/t/what-does-the-6-before-info-mean-and-will-it-always-be-6/43108/17

#### Checklist

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -23,7 +23,7 @@ import (
 	"github.com/rclone/rclone/vfs/vfscommon"
 	"github.com/rclone/rclone/vfs/vfsflags"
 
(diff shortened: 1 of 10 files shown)


ncw (approved):
Looks nice - great idea :-)

Let's see if it fixes the forum user's problem before merging.

I see I've accidentally merged some code which has broken the other os test :-(

And we have an unresolved security issue (which doesn't have a fix yet) which is breaking the lint tests.

ncw:
The user reported on the forum that this works fine!

https://forum.rclone.org/t/what-does-the-6-before-info-mean-and-will-it-always-be-6/43108/25

I fixed the conflict and I'll merge when the CI goes green - thank you :-)

PR #7493, opened 2023-12-07
Support WebDAV headers with CORS

### LINKED ISSUE #7492: PROPFIND with Depth fails on Serve WebDAV

n/a

1.65

Win 11, 64-bit

local drive

n/a

#### How to use GitHub

alensiljak:
Fixes #7492

It's a follow-up on previous fixes of WebDAV with CORS #7444, #7448. The direct issue is #7492.

#### Checklist

### DIFF
--- a/lib/http/middleware.go
+++ b/lib/http/middleware.go
@@ -183,8 +183,8 @@ func MiddlewareCORS(allowOrigin string) Middleware {
 
 			if allowOrigin != "" {
 				w.Header().Add("Access-Control-Allow-Origin", allowOrigin)
-				w.Header().Add("Access-Control-Allow-Headers", "Authorization, Content-Type, Depth, Destination, If, Lock-Token, Overwrite, TimeOut, Translate")
-				w.Header().Add("Access-Control-Allow-Methods", "COPY, DELETE, GET, HEAD, LOCK, MKCOL, MOVE, OPTIONS, POST, PROPFIND, PROPPATCH, PUT, TRACE, UNLOCK")
+				w.Header().Add("Access-Control-Allow-Headers", "authorization, Content-Type")
+				w.Header().Add("Access-Control-Request-Method", "POST, OPTIONS, GET, HEAD")
 			}
 
 			next.ServeHTTP(w, r)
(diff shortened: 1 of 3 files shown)


alensiljak:
This immediately fixes issues with the parameters passed as headers. Tested with PROPFIND.

I agree this should be localized to WebDAV server but am unsure where or how to move this code.
I reckon https://github.com/rclone/rclone/blob/2f5685b40562cf764da673543245ddb13c049a4a/cmd/serve/webdav/webdav.go#L197 would be a good start?
Any tips appreciated. The migration may be done in this pull request or separately. 
This is also a good candidate for a bugfix release.

alensiljak:
The failed checks do not seem to have anything to do with this change, from what I can see.

lib/http/middleware.go
```diff
@@ -183,7 +183,7 @@ func MiddlewareCORS(allowOrigin string) Middleware {
 
 			if allowOrigin != "" {
 				w.Header().Add("Access-Control-Allow-Origin", allowOrigin)
-				w.Header().Add("Access-Control-Allow-Headers", "authorization, Content-Type")
+				w.Header().Add("Access-Control-Allow-Headers", "Authorization, Content-Type, Depth, Destination, If, Lock-Token, Overwrite, TimeOut, Translate")
```
ncw, line 186: Can you put this in the `serve webdav` middleware please? We don't need it to apply to all of rclone's http server.

Likewise the webdav only `Access-Control-Allow-Methods` should be in there too.

I'm not sure exactly how to do that - it might require a bit of plumbing in rclone.

alensiljak: > I'm not sure exactly how to do that - it might require a bit of plumbing in rclone.

This is the dreadful part. I'm not really a Go developer and it may take me quite a while to figure that out. :(
I agree with that comment. We can also separate these issues. This above is a quick fix for allowing headers, which fixes the WebDAV functionality. Finding the correct way to limit it to `webdav serve` will take longer, it seems. I looked at the `serve webdav` code and do not understand much there, as I've mentioned in the related issue.

alensiljak:
This was a bit simpler than I thought. Please review. I had to expose two fields in the Server struct in order to access them in webdav server. Wondering if this is acceptable.

ncw (commented):
I think exposing `Mux` is a good idea. It makes the commit a bit messy so I'd probably do one commit where all you do is `mux` -> `Mux`.

The next commit you can make your changes.

The only thing I don't like about your change is the mostly duplicated `MiddleWareCORS` function - maybe we should factor the the one in libhttp into a routine which take a `methods` and `headers` parameters?

PR #7494, opened 2023-12-08
Using WebDAV with CORS

### LINKED ISSUE #7492: PROPFIND with Depth fails on Serve WebDAV

alensiljak:
Fixes #7492

This is new pull request instead of #7493. The commits have been squashed and separated thematically.
It's a follow-up on previous fixes of WebDAV with CORS #7444, #7448. The direct issue (a problem encountered) is #7492.

#### Checklist

### DIFF
--- a/cmd/serve/webdav/webdav.go
+++ b/cmd/serve/webdav/webdav.go
@@ -203,6 +203,11 @@ func newWebDAV(ctx context.Context, f fs.Fs, opt *Options) (w *WebDAV, err error
 		return nil, fmt.Errorf("failed to init server: %w", err)
 	}
 
+	// Add the CORS headers.
+	w.Server.Mux.Use(libhttp.MiddlewareCORS(w.Server.Cfg.AllowOrigin,
+		"COPY, DELETE, GET, HEAD, LOCK, MKCOL, MOVE, OPTIONS, POST, PROPFIND, PROPPATCH, PUT, TRACE, UNLOCK",
+		"Authorization, Content-Type, Depth, Destination, If, Lock-Token, Overwrite, TimeOut, Translate"))
+
 	webdavHandler := &webdav.Handler{
 		Prefix:     w.opt.HTTP.BaseURL,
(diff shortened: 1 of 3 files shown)


alensiljak:
This actually doesn't work now. Previously, in the separate method in webdav, I used .set() method for headers, changing the existing ones, whereas now they are just added again:

image

The browser seems to pick up the first one and the requests fail.

alensiljak:
Fixed with .Set() instead of .Add().

ncw:
This is looking nice now.

We could merge it like this...

However I note from your experiments above that `MiddlewareCORS` is being called twice (the `.Add` vs `.Set` is a giveaway). I don't think it is possible to remove a middleware.

Maybe what we want is to store the CORSMethods and CORSHeaders in the [http server config](https://github.com/rclone/rclone/blob/110d07548fbe18937933239d823da904826205b3/lib/http/server.go#L100) and then instantiate the [MiddlewareCORS](https://github.com/rclone/rclone/blob/110d07548fbe18937933239d823da904826205b3/lib/http/middleware.go#L169)  with a `*http.Config` instead of `allowOrigin string` (since that is in the `http.Config` already)  - then it can read everything it needs from the `http.Config` directly.

This would also save exposing `Mux` also.

The `CORSMethods` and `CORSHeaders` should be initialised to the current values in the [DefaultCfg](https://github.com/rclone/rclone/blob/110d07548fbe18937933239d823da904826205b3/lib/http/server.go#L135)

This would then stop `MiddlewareCORS` being called twice and make it an easy config option in the servers that care.

I hope you don't mind this iterative approach to design - sometimes you don't see the way until you've tried some wrong turnings first!

alensiljak:
Great! Thanks for the feedback and the pointers. I think I feel a bit more comfortable around these areas now.
Continuing in #7495. Is this what you had in mind?

PR #7495, opened 2023-12-09
webdav: WebDAV-specific methods and headers

### LINKED ISSUE #7492: PROPFIND with Depth fails on Serve WebDAV

alensiljak:
Fixes #7492

This is new pull request instead of #7493 & #7494. The commits have been squashed and separated thematically.
It's a follow-up on previous fixes of WebDAV with CORS #7444, #7448. 
The direct issue (a problem encountered) is #7492.

#### Checklist

### DIFF
--- a/lib/http/middleware.go
+++ b/lib/http/middleware.go
@@ -166,9 +166,9 @@ func MiddlewareAuthCustom(fn CustomAuthFn, realm string, userFromContext bool) M
 var onlyOnceWarningAllowOrigin sync.Once
 
 // MiddlewareCORS instantiates middleware that handles basic CORS protections for rcd
-func MiddlewareCORS(allowOrigin string) Middleware {
+func MiddlewareCORS(cfg *Config) Middleware {
 	onlyOnceWarningAllowOrigin.Do(func() {
-		if allowOrigin == "*" {
+		if cfg.AllowOrigin == "*" {
 			fs.Logf(nil, "Warning: Allow origin set to *. This can cause serious security problems.")
 		}
 	})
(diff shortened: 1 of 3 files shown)


lib/http/middleware.go
```diff
@@ -166,9 +166,9 @@ func MiddlewareAuthCustom(fn CustomAuthFn, realm string, userFromContext bool) M
 var onlyOnceWarningAllowOrigin sync.Once
 
 // MiddlewareCORS instantiates middleware that handles basic CORS protections for rcd
-func MiddlewareCORS(allowOrigin string) Middleware {
+func MiddlewareCORS(cfg Config) Middleware {
```
ncw, line 169: I would pass a pointer to Config here otherwise you are copying a big structure about.

ncw (commented):
Very good 👍

One minor suggestion inline.

Thank you.

lib/http/middleware.go
alensiljak: I did not know there were pointers in go. 🙈
Will check it out tomorrow.

alensiljak: ok, that was easy. Should be fine now?

alensiljak:
Hi. Any chance this goes into a release soon? Are there any particular reasons the fix is not being included?

fyears:
Hi I come across the same browser  CORS issue while using rclone webdav serve. It seems that this pr should solve the issue. Is it possible to get it merged?

ncw:
I think this was fixed in 5fba5025168aac0c9aa5dd767e51e0ee264ebbe5 which was released in v1.65.0

@fyears would you agree?

fyears:
hmmm so maybe i am meeting another issue, and let me debug it (again) with the latest version. Sorry for brothering

ncw:
@fyears no worries - it might not be fixed properly :-)

ncw:
Hi @alensiljak, thank you for this, and apologies it sat so long.

The WebDAV-specific CORS headers and methods this adds are now handled in `lib/http` - the `MiddlewareCORS` sends the `Depth`/`Destination`/`Lock-Token`/... headers and the `PROPFIND`/`PROPPATCH`/`LOCK`/... methods. So this is effectively covered already, and I'm going to close it as superseded.

Thanks again for the work, and please do reopen or file a fresh issue if you find a case that isn't handled.

Issue #5678, opened 2021-10-08
bisync: implement sessions


ivandeex:
## What problem are you trying to solve?

We should track critical filtering and change detection flags in sessions
and either add a flag named `--use-saved-options` (or `--continue`)...
or **by default** restore such flags from the _session json_,
allowing user to change such flags only when `--resync`.

Also beta `bisync` introduced the flag `--filters-file` to track hashsum of filter files.
We should remove it (it duplicates `--filter-from`) and track filter checksums in the session altogether.

## Implementation details

Currently bisync applies its own set of rules to adapt names of `source-target listings` to local OS.
Better use good old `filepath.ToSlash`, `filepath.FromSlash` as well as new `encoder.OS` from commit https://github.com/rclone/rclone/commit/3a2f748aebfe4763c375de45ff18584e8c7a886b.

### https://github.com/rclone/rclone/pull/5164#issuecomment-842407758 (Jwink3101)

Unlike mirroring or one-way sync (a la `sync`), true bidirectional sync cannot be stateless. However, I think where to store the state warrants more discussion. One thing often discussed on the forum and rclone subreddit is that you can pull your config to a VPS and run it there. If you store the state of past syncs on the syncing machine, you now need to keep track of that.

This is actually, in my opinion, one of the fundamental issues with rclone and bi-directional sync. You need a robust way to handle this and you can't just go into a deeper dir. (for example `rclone sync A: B:` works as does `rclone sync A:sub B:sub`). This is the primary reason I chose to for my tool to not be a pure CLI one and instead of a sync config file that dictated these things. 

### https://github.com/rclone/rclone/pull/5164#issuecomment-842512914 (ivandeex)

I thought in a different direction - to snapshot the current state of flags at the time of operation into bisync session (extending the idea of hashsumming the filters) and request a resync or at least warn if it changes.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843406665 (cjnaz)

Hashing the user's filter file in rclonesync addressed a bug where many files were deleted because they did not show up in the current state list and were in the prior history list, so they needed to get deleted on the other side.  The rclonesync algorithm evolved, so the specific fail case may not be there now.  In any case, damage control must be a prime design requirement.  `--backup-dir` sounds appropriate.  Are CLI filter options accepted by bisync?

### https://github.com/rclone/rclone/pull/5164#issuecomment-843460561 (ivandeex)

[...]

PR #7497, opened 2023-12-09
bisync: normalize session name to non-canonical (and docs edits) - fixes #7423

### LINKED ISSUE #7423: bisync: session name should normalize to non-canonical version for overridden configs
When calling bisync with a remote, but a leading `/` in the remote path like so

```bash
rclone bisync remote:/path/with/unneeded/slash /some/local/path --resync
```

the path will be normalized by Rclone and the command will be executed without throwing any error (or warning).

If you then try to run `bisync` without `--resync` however, it will not behave as expected. Running the exact same command without `--resync` will always yield:

```log
ERROR : Bisync critical error: cannot find prior Path1 or Path2 listings, likely due to critical error on prior run
```

This suggests that `bisync` failed previously, when in fact it didn't. It seems to name the listings after the normalized path, while checking for existing listings with the non-normalized path, as provided by the user.

This seems likely, because running the command again, but without the leading slash and `--resync`, will work:

```bash
rclone bisync remote:path/with/unneeded/slash /some/local/path
```

rclone v1.64.2
- os/version: arch Soaring (64 bit)
- os/kernel: 6.6.1-zen1-1-zen (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.3
- go/linking: dynamic
- go/tags: none

Arch Linux, 64 bit

Proton Drive

#### How to use GitHub

nielash:
(This PR is dependent on #7410 -- only 3 of the commits are new 🙂 )

Before this change, bisync used the "canonical" Fs name in the filename for its listing files, including any {hexstring} suffix. An unintended consequence of this was that if a user added a backend-specific flag from the command line (thus "overriding" the config), bisync would fail to find the listing files it created during the prior run without this flag, due to the path now having a {hexstring} suffix that wasn't there before (or vice versa, if the flag was present when the session was established, and later removed.) This would sometimes cause bisync to fail with a critical error (if no listing existed with the alternate name), or worse -- it would sometimes cause bisync to use an old, incorrect listing (if old listings with the alternate name DID still exist, from before the user changed their flags.) 

After this change, the issue is fixed by always normalizing the SessionName to the non-canonical version (no {hexstring} suffix), regardless of the flags. To avoid a breaking change, we first check if a suffixed listing exists. If so, we rename it (and overwrite the non-suffixed version, if any.) If not, we carry on with the non-suffixed version. (We should only find a suffixed version if created prior to this commit.) 

The result for the user is that the same pair of paths will always use the same .lst filenames, with or without backend-specific flags.

- #7423
- [Forum post](https://forum.rclone.org/t/bisync-bugs-and-feature-requests/37636#:~:text=7.%20Overridden%20config%20can%20cause%20bisync%20critical%20error%20requiring%20%2D%2Dresync)
- [User Report](https://forum.rclone.org/t/bisync-missed-syncing-changed-file-no-error-no-warning/41174/6?u=nielash)

#### Checklist

### DIFF
--- a/cmd/bisync/bilib/canonical.go
+++ b/cmd/bisync/bilib/canonical.go
@@ -2,12 +2,15 @@
 package bilib
 
 import (
+	"context"
 	"os"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strings"
 
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/operations"
 )
 
 // FsPath converts Fs to a suitable rclone argument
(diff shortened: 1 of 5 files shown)


ncw (approved):
The point of those `{suffix}` is a reminder that with different backend flags the contents of the remote may be different.

They were put in for the VFS caching to make sure we don't mix up data we shouldn't.

For example if you override the `--drive-root-folder-id` you'll get a different set of files. However a lot of the flags, eg `--drive-list-chunk` will not affect the files shown.

So by ignoring the `{...}` you are removing a safety check - we should be encouraging the users to put the config overrides into the config file so they remain static.

I see that bisync has exactly the same problem though with not being able to find previous listings.

Maybe we should merge your fix just with a warning to the user about getting the config overrides into the config file to make a static listing.

Happy to go with your decision here, so I shall approve this :-)

nielash:
Thanks! Very good points. I think the safety check would be more properly addressed for Bisync by #5678 Sessions -- i.e. the ability for Bisync to detect that important flags have changed (whether backend or not) and require some extra confirmation like `--force` to proceed. I think the current `{suffix}` implementation has a few flaws:

- it catches only backend-specific flags, even though other flags can also change the contents of the remote (for example `--max-age`, `--ignore-case-sync`)
- it doesn't distinguish between important changes (`--drive-root-folder-id`) and unimportant changes (`--drive-pacer-min-sleep`)
- putting the overrides in the config is not ideal for flags like `--drive-skip-gdocs` where sometimes I want it (`bisync`) and sometimes I don't (`lsl`). You'd have to either make two remotes or override the override with `--drive-skip-gdocs=false` when necessary, which is not terrible but slightly annoying 😄
- The safety check can accidentally result in less safe behavior -- for instance, with Bisync, if it results in accidentally using old suffixed listings from a year ago that are still lying around.

Bisync users trip over this one quite a bit, so I think for the moment the unintended consequences outweigh the intended ones, and this feature should be merged, to be followed shortly by a better safety check via #5678.

As an aside: part of the reason I started working on a draft of https://github.com/rclone/rclone/issues/2697#issuecomment-1888668988 (Profiles) is because I am considering this as a possible way to implement #5678, as ivandeex also alluded to on that ticket. Basically -- use the Profiles methods to store a Bisync session in the config, and then on each run Bisync can detect if there are important differences. But it would also be possible to avoid storing this in the config, if the Profiles feature is too big to take on right now (for instance, Bisync could just marshall the flag configuration to JSON and store in the `workdir` alongside its other listing files.) No pressure at all to look at it -- just thought I'd mention as it is relevant to this ticket!

nielash:
(rebased to clear out the duplicate commits)

ncw:
> Thanks! Very good points. I think the safety check would be more properly addressed for Bisync by #5678 Sessions -- i.e. the ability for Bisync to detect that important flags have changed (whether backend or not) and require some extra confirmation like `--force` to proceed. I think the current `{suffix}` implementation has a few flaws:
> 
> * it catches only backend-specific flags, even though other flags can also change the contents of the remote (for example `--max-age`, `--ignore-case-sync`)

Hmm, yes! Though people don't usually use those with `rclone mount` which was the main user of this.

> * it doesn't distinguish between important changes (`--drive-root-folder-id`) and unimportant changes (`--drive-pacer-min-sleep`)

Aye. There is no way to do this other than markup all the individual config items which I decided was too much work!

> * putting the overrides in the config is not ideal for flags like `--drive-skip-gdocs` where sometimes I want it (`bisync`) and sometimes I don't (`lsl`). You'd have to either make two remotes or override the override with `--drive-skip-gdocs=false` when necessary, which is not terrible but slightly annoying 😄
> * The safety check can accidentally result in less safe behavior -- for instance, with Bisync, if it results in accidentally using old suffixed listings from a year ago that are still lying around.
> 
> Bisync users trip over this one quite a bit, so I think for the moment the unintended consequences outweigh the intended ones, and this feature should be merged, to be followed shortly by a better safety check via #5678.

That sounds fine to me.

I prefer the evolution towards perfection, than the striving for perfection immediately approach!

> As an aside: part of the reason I started working on a draft of [#2697 (comment)](https://github.com/rclone/rclone/issues/2697#issuecomment-1888668988) (Profiles) is because I am considering this as a possible way to implement #5678, as ivandeex also alluded to on that ticket. Basically -- use the Profiles methods to store a Bisync session in the config, and then on each run Bisync can detect if there are important differences. But it would also be possible to avoid storing this in the config, if the Profiles feature is too big to take on right now (for instance, Bisync could just marshall the flag configuration to JSON and store in the `workdir` alongside its other listing files.) No pressure at all to look at it -- just thought I'd mention as it is relevant to this ticket!

OK! I'm catching up on the PRs so will get to it soon :-)

Issue #5675, opened 2021-10-08
bisync: investigate alternative approaches


ivandeex:
## Synopsis

@JWink3101 once noted in https://github.com/rclone/rclone/pull/5164#issuecomment-842407758

There are two fundamental ways to do bi-sync:

1. Compare each side independently to past state to develop a changelog. Propagate that change log
2. Compare current state on each side using past state *only* for resolution (i.e. is this new on one or deleted on the other? Is it modified on one? Both?)
    - Additional use for tracking renames though this isn't needed if you don't mind rclone's current approach requiring shared hashes or ModTIme

You are doing the first. This has the advantage of being robust to time differences but also breaks with rclone's current behavior of `--modify-window` and the more general behavior of looking at current state. With approach 2, you can handle ModTime's not agreeing **and/or size and/or hash** like regular rclone. 

### https://github.com/rclone/rclone/pull/5164#issuecomment-842407758 (JWink3101)

From bisync docs:

> Note that on some cloud storage systems it is not possible to have file timestamps that match _precisely_ between the local and other filesystems. Bisync's approach to this problem is by tracking the deltas against saved state on each side _separately_, and then applying the resulting changes on the other side.

There are two fundamental ways to do bi-sync:

1. Compare each side independently to past state to develop a changelog. Propagate that change log
2. Compare current state on each side using past state *only* for resolution (i.e. is this new on one or deleted on the other? Is it modified on one? Both?)
    - Additional use for tracking renames though this isn't needed if you don't mind rclone's current approach requiring shared hashes or ModTIme

Based on your comment, you are doing the first. This has the advantage of being robust to time differences but also breaks with rclone's current behavior of `--modify-window` and the more general behavior of looking at current state.

With approach 2, you can handle ModTime's not agreeing **and/or size and/or hash** like regular rclone. 

The sub-bullet on the second is also interesting as it *does* allow you to track renames even if hashes aren't compatible. But you could also track with rclone's current behavior of looking at the extra files. Then use the past state to decide which side did the moving.

Approach 1 also requires the whole idea of `--resync` while it is implicit if you don't have past state in approach 2.

### https://github.com/rclone/rclone/pull/5164#issuecomment-842512914 (ivandeex)

> there are two fundamental ways to do bi-sync: ...
> With approach 2, you can handle ModTime's not agreeing **and/or size and/or hash** like regular rclone.

[...]

PR #7498, opened 2023-12-09
bisync: full support for comparing `checksum`, `size`, `modtime` - fixes #5679 fixes #5683 fixes #5684

### LINKED ISSUE #5679: bisync: add modtime support
## What problem are you are trying to solve?

bisync delta engine and unit tests do not take backend time tolerance into account. this should be fixed.

### https://github.com/rclone/rclone/pull/5164#discussion_r634229878 (ivandeex)

the unit test runner is based on `mangleListing` which is a very far relative of [fstest.CompareItems](https://github.com/rclone/rclone/blob/v1.55.1/fstest/fstest.go#L356). It already has ability to ignore hashsums on demand. It still needs ability to compare times with per-backend tolerance or even ignore times if unsupported. Having that would let us run tests against backends without time support and probably lift the corresponding ban.

Tentatively this can be implemented as:
1. parse out the time filed and convert to `time.Time`
2. round time down to _doubled_ tolerance
3. put time field back into string
4. feed the result to `diff`.

Why not just parse and compare with tolerance?
Because `mangleResult` is two-fold. Besides comparing it can prepare a test state snapshot for storing in golden directory, which later can provide a **bigger picture** of what's going on.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843151817 (ivandeex)

With a feature _must_ come a method of testing it.

Rclone has a comprehensive `fstest` framework dedicated to peculiarities of the plethora of supported cloud providers. Planning to work on this patch I had the heart to keep the project test engine's nature (while translating it from python to golang and refactoring it to work natively/unattended on Windows and with a wider set of backends) and propose it as a new kid on the set of rclone approaches (see my inline comment above), instead of rewriting it from the grounds up for `fstest`.

The bisync test suite **lacks** the ability to compare imprecise/absent timestamps as of the time of this writing (I hope I can say it, _absent_ timestamps should be "comparable" too). It works correctly mostly because it compares logs from a particular filesystem **with itself**. Unless this ward is fixed, we _can't run tests_ against backends without modtime and consequently can't declare them supported. I have an idea on how to fix that (see another inline comment), but the effort will need time.

## modtime in delta engine

Try to extend `operations.Equal` into `operations.Compare` with more returned conditions like `Newer, Older, ...` (related to #4810) for direct use in bisync delta engine

### https://github.com/rclone/rclone/pull/5164#issuecomment-842407758 (JWink3101)

The fact that this only supports a small subset of remotes is disappointing, esp. some popular ones not supporting mod time. Others, like S3, are way more performant if you disable it.

This is an issue with the overall algorithm of some sync tools but there is no reason you *need* modtime to decide what to sync. You may need it for *conflicts* but if you have past state, then you can tell if a changed file-size is a modification on one side only.

### https://github.com/rclone/rclone/pull/5164#issuecomment-842512914 (ivandeex)

> This is an issue with the overall algorithm of some sync tools but there is no reason you _need_ modtime to decide what to sync.

Theoretically, no. The bisync change detection in its current state is all about [bit masks](https://github.com/rclone/rclone/pull/5164/files#diff-25f7d45d3740899fdbbb93ff87a682943b7085b9c2f7f67b9cc58756c90e3630R31) (as opposed to "rclonic" [Equal](https://github.com/rclone/rclone/blob/v1.55.1/fs/operations/operations.go#L167)). But it needs more testing.

## How to use GitHub

### LINKED ISSUE #5683: bisync: honor `--checksum` and `--ignore-checksum`

### LINKED ISSUE #5684: bisync: add size checks to the delta engine and unit tests
## Synopsis

Delta engine in the _bisync beta_ can only use `modtime` to detect changes.
We _could_ take file size into consideration too (especially in cases when modtime is not supported),
and add relevant test cases.

### https://github.com/rclone/rclone/pull/5164#issuecomment-842512914 (ivandeex)

> This is an issue with the overall algorithm of some sync tools but there is no reason you _need_ modtime to decide what to sync.

Theoretically, no. The bisync change detection in its current state is all about [bit masks](https://github.com/rclone/rclone/pull/5164/files#diff-25f7d45d3740899fdbbb93ff87a682943b7085b9c2f7f67b9cc58756c90e3630R31) (as opposed to "rclonic" [Equal](https://github.com/rclone/rclone/blob/v1.55.1/fs/operations/operations.go#L167)). But it needs more testing.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843406665 (cjnaz)

Detecting on size change only leaves out the case where a file is changed but is the same size.  This could happen when a process writes a time string to a file, for example.  Other than this case, in practice changed-but-same-size is relatively rare; however, I'd be submitting a bug report if such a file was missed in the sync.

### https://github.com/rclone/rclone/pull/5164#issuecomment-843426747 (JWink3101)

Yep. I agree. It is a bad method but it is the same one, to my knowledge, used my rclone already for remotes that do not provide ModTime or a hash

## How to use GitHub

nielash:
(This PR is dependent on #7410 and #7497 -- only 1 commit is new 🙂 )

Before this change, bisync could only detect changes based on `modtime`, and would refuse to run if either path lacked `modtime` support. This made bisync unavailable for many of rclone's backends. Additionally, bisync did not account for the Fs's precision when comparing modtimes, meaning that they could only be reliably compared within the same side -- not against the opposite side. `size` and `checksum` (even when available) were ignored completely for deltas. 

After this change, bisync now fully supports comparing based on any combination of `size`, `modtime`, and `checksum`, lifting the prior restriction on backends without `modtime` support. The comparison logic considers the backend's precision, hash types, and other features as appropriate. 

The comparison features optionally use a new `--compare` flag (which takes any combination of `size`,`modtime`,`checksum`) and even supports some combinations not otherwise supported in `sync` (like comparing all three at the same time.) By default (without the `--compare` flag), bisync inherits the same comparison options as `sync` (that is: `size` and `modtime` by default, unless modified with flags such as `--checksum` or `--size-only`.) If the `--compare` flag is set, it will override these defaults. 

If `--compare` includes `checksum` and both remotes support checksums but have no hash types in common with each other, checksums will be considered only for comparisons within the same side (to determine what has changed since the prior sync), but not for comparisons against the opposite side. If one side supports checksums and the other does not, checksums will only be considered on the side that supports them. When comparing with `checksum` and/or `size` without `modtime`, bisync cannot determine whether a file is newer or older -- only whether it is changed or unchanged. (If it is changed on both sides, bisync still does the standard equality-check to avoid declaring a sync conflict unless it absolutely has to.) 

Also included are some new flags to customize the checksum comparison behavior on backends where hashes are slow or unavailable. `--no-slow-hash` and `--slow-hash-sync-only` allow selectively ignoring checksums on backends such as `local` where they are slow. `--download-hash` allows computing them by downloading when (and only when) they're otherwise not available. Of course, this option probably won't be practical with large files, but may be a good option for syncing small-but-important files with maximum accuracy (for example, a source code repo on a `crypt` remote.) An additional advantage over methods like `cryptcheck` is that the original file is not required for comparison (for example, `--download-hash` can be used to bisync two different `crypt` remotes with different passwords.) 

Additionally, all of the above are now considered during the final `--check-sync` for much-improved accuracy (before this change, it only compared filenames!) 

Many other details are explained in [the included docs](https://github.com/nielash/rclone/blob/bisync-compare/docs/content/bisync.md#--compare).

- #5679 
- #5683
- #5684
- #5675

#### Checklist


ncw (approved):
This is an incredible bit of work - it takes bisync to the next level :-)

I've given it a light review and nothing terrible jumps out!

There are some things in here we should probably backport to sync/copy/move so we keep the bisync and sync experience as consistent as possible (eg `--compare`) but we can talk about that later!

Let's get this merged for v1.66!

nielash:
> This is an incredible bit of work - it takes bisync to the next level :-)

Thank you 😃 

> There are some things in here we should probably backport to sync/copy/move so we keep the bisync and sync experience as consistent as possible (eg `--compare`) but we can talk about that later!

Agreed -- that would be great!

Jwink3101:
It is bittersweet but this may get me to drop developing and using [syncrclone](https://github.com/Jwink3101/syncrclone) for the native solution. At the very least, I will have to re-evaluate the comparison table when I get a chance. 

Thanks for your hard work!

nielash:
> It is bittersweet but this may get me to drop developing and using [syncrclone](https://github.com/Jwink3101/syncrclone) for the native solution. At the very least, I will have to re-evaluate the comparison table when I get a chance.
> 
> Thanks for your hard work!

Thank you for the very kind words, @Jwink3101! Coincidentally, while working on this I read through a lot of the feedback that you gave ivandeex back when bisync was first developed, and I found it very helpful! Particularly your insights about how `modtime` isn't strictly necessary for change detection. 🙂

Issue #7470, opened 2023-11-29
bisync: recover from exit without `--resync`


crocinsocks:
It would be great if bisync could recover from an exit (SIGINT, SIGTERM, SIGKILL) without the need of a `--resync`.

e.g. powercut, Ctrl+C, shutdown

This would save the extra work to get it to play nice with everyday usage. Currently, have to run a script to wait for sync(s) to finish then shutdown/restart/etc. In some cases there is nothing that can be done to prevent this such as as system crash or power cut resulting in `--resync` being required. Or sometimes you just want a quick restart due to a kernel/application update or other issue and instead you have to wait or accept the consequences. 

Thanks

nielash:
Definitely agree this could and should be improved!

(For others reading this: [here's our previous discussion](https://github.com/rclone/rclone/issues/7332#issuecomment-1732881448) about it, with some good background info about why this is a particularly tough challenge for bisync.)

Some possible solution approaches include:

- "autosave" -- save progress state every X seconds (or after every X transfers) throughout a run, so in the event of a crash, the next run can resume from the last autosave. 
Pros: probably the most accurate method
Cons: may be complicated to build, may slow down the sync with lots of writes to disk, may not be practical with a non-local `--workdir` if that is eventually supported

- "replay queues" -- save the queue list at the beginning of each sync, and in the event of an interruption, the next run attempts to finish the prior run by "replaying" it, before starting the new run.
Pros: easier and less expensive than "autosave"
Cons: potentially not robust to changes that occurred after the interrupted run

- "graceful shutdown" -- upon receiving Ctrl+C, gracefully wrap up what we're doing at the earliest opportunity, cleaning up after ourselves to make things easy for next time.
Pros: relatively easy and clean
Cons: only works when we have some warning (won't help with powercut, sudden crash, etc.)

- "diff from last trusted listing" -- similar idea to https://github.com/rclone/rclone/pull/7410/commits/ef4b47a9824754824a05491c3d4c7a2ef97253d4, but revert the whole listing instead of individual files, and re-diff.
Pros: the easiest and perhaps most elegant option
Cons: there's [one particular edge case](https://github.com/rclone/rclone/issues/7332#:~:text=the%20one%20scenario,and%20move%20on.%22) it wouldn't handle as well as the others (not sure how important that is)

One question I have for you is: how important is it to handle the "crash" scenarios in addition to the ones where we have some warning? i.e. would a "graceful shutdown" option go far enough for you?

[...]

PR #7501, opened 2023-12-11
bisync: Graceful Shutdown, --recover from interruptions without --resync

nielash:
(This PR is dependent on #7410, #7497, and #7498 -- only 2 commits are new 🙂 )

Before this change, bisync had no mechanism to gracefully cancel a sync early and exit in a clean state. Additionally, there was no way to recover on the next run -- any interruption at all would cause bisync to require a --resync, which made bisync more difficult to use as a scheduled background process. 

This change introduces a "Graceful Shutdown" mode and --recover flag to robustly recover from even un-graceful shutdowns. Details are documented thoroughly in the commit messages, and [the docs](https://github.com/nielash/rclone/blob/bisync-recover/docs/content/bisync.md#--recover).

- #7470
- #7332
- [How to reduce need for bisync::resync?](https://forum.rclone.org/t/how-to-reduce-need-for-bisync-resync/31067)

#### Checklist

### DIFF
--- a/cmd/bisync/cmd.go
+++ b/cmd/bisync/cmd.go
@@ -48,10 +48,12 @@ type Options struct {
 	SaveQueues            bool // save extra debugging files (test only flag)
 	IgnoreListingChecksum bool
 	Resilient             bool
+	Recover               bool
 	TestFn                TestFunc // test-only option, for mocking errors
 	Retries               int
 	Compare               CompareOpt
 	CompareFlag           string
+	DebugName             string
 }
 
 // Default values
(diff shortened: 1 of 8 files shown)


ncw (approved):
This looks very nice :-)

I've thought about doing something similar for sync, so one CTRL-C does a graceful shutdown of the sync - finished transferring existing objects - kill with a second. I haven't done this because I was worried about backwards compatibility, so we can see how this works with bisync which will be useful.

Great work :-)

nielash:
> I've thought about doing something similar for sync, so one CTRL-C does a graceful shutdown of the sync - finished transferring existing objects - kill with a second. I haven't done this because I was worried about backwards compatibility, so we can see how this works with bisync which will be useful.
> 
> Great work :-)

That sounds great to me! I agree it would be very useful in sync. The backwards compatibility concern could be addressed by using some other signal for it, I suppose!

Issue #6437, opened 2022-09-12
sync fails to remove directories on the destination that do not exist on the source


eharris:
#### What is the problem you are having with rclone?

I have noticed that under some circumstances (in my case, using a crypted remote) `rclone sync` fails to remove directories that should be removed since they do not exist on the source.  In my particular circumstance, it seems like the problem is that `rclone sync` uses something like the `rmdir` command, which (somewhat silently) fails if the directory has any contents left, even broken contents that `rclone` is ignoring.

I think this could be fixed by making the `sync` use `purge` functionality instead of `rmdir` to remove directories that don't exist on the source.

Here is the output from the `rclone sync` command, which claims that it has deleted 23 directories.

```
> rclone sync --progress --bwlimit 1M --links --track-renames --track-renames-strategy size,leaf,modtime --transfers 10 --max-backlog 20 --tpslimit 40 /raid/audio odc:audio
Transferred:              0 B / 0 B, -, 0 B/s, ETA -
Checks:            296345 / 296345, 100%
Deleted:                0 (files), 23 (dirs)
Elapsed time:     13m50.5s
```

However, a subsequent run of the sync without any changes being made still shows it trying to remove the 23 directories it claimed it had removed on the last run.  Note that no warnings or errors were reported at all (which is a separate problem).

So I ran again with `-v` to see what was going on with the directories it thought it was removing and got messages like this:

```
2022-09-12 16:44:10 INFO  : work/new/waiting/Johann_Sebastian_Bach_Canadian_Brass_-_Art_Of_The_Fugue: Removing directory
2022-09-12 16:44:10 INFO  : work/new/waiting/Johann Sebastian Bach; Canadian Brass - Art of the Fugue: Removing directory
```

Note that again, even with `-v`, no warnings or errors are reported with the supposedly removed directories.

So I tried manually listing (with `lsl`) one of the problem directories:

```
> rclone lsl odc:audio/work/new/waiting/Johann_Sebastian_Bach_Canadian_Brass_-_Art_Of_The_Fugue
>
```

Apparently no contents.  So I tried `rmdir` on that directory:

```
> rclone rmdir odc:audio/work/new/waiting/Johann_Sebastian_Bach_Canadian_Brass_-_Art_Of_The_Fugue2022/09/12 16:48:06 ERROR : Attempt 1/3 failed with 1 errors and: folder not empty
2022/09/12 16:48:06 ERROR : Attempt 2/3 failed with 1 errors and: folder not empty
[...]

PR #7512, opened 2023-12-18
operations: --no-block-rmdir for disposable files blocking removal of otherwise empty directory

nielash:

Before this change, rclone commands that remove empty dirs (including `rmdir`, `rmdirs`, `sync`, and `bisync`) would silently fail to remove dirs if those dirs contained files that were excluded by filters. This can pose a problem for some automatically created system files, such as `.DS_Store` on macOS, which users often prefer to exclude from syncs -- including them makes for noisy logs, but excluding them can make it impossible to remove otherwise empty directories.

After this change, a new `--no-block-rmdir` flag allows specifying a comma-separated list of such files which rclone should consider "disposable" if they block removal of otherwise empty directories. When set, if rclone fails to remove a directory, it will check the contents of the directory (ignoring any filters.) If the directory contains ONLY files on the disposable list, rclone will delete the files and then remove the directory.

See partially-related #6437

#### Checklist


nielash:

ncw (commented):
This code looks neat and well implemented :-)

My immediate thoughts are

- Are users going to want to use a glob, eg `--no-block-rmdir *~` for all the emacs users in the world - I think probably not but we should discuss now.
- Can we think of a better name than `-no-block-rmdir`?
- How useful is this in the mac world? Can you describe use cases that this fixes?

nielash:
> * Are users going to want to use a glob, eg `--no-block-rmdir *~` for all the emacs users in the world - I think probably not but we should discuss now.

Possibly! I think that would be a useful feature, at the possible cost of adding some complexity.

> * Can we think of a better name than `-no-block-rmdir`?

Haha. Some ideas:

- `--ignorable`
- `--deletable` 
- `--removable`
- `--disposable`
- `--discardable`

Syncthing has a [similar feature](https://docs.syncthing.net/users/ignoring.html#:~:text=A%20pattern%20beginning%20with%20a%20(%3Fd)%20prefix%20enables%20removal%20of%20these%20files%20if%20they%20are%20preventing%20directory%20deletion.) but it doesn't seem to have a name.

> * How useful is this in the mac world? Can you describe use cases that this fixes?

Pretty useful, IMO. The main use case is for `.DS_Store` files, since the OS creates them in every folder whether you like it or not. Here are two examples:

- You have a `crypt` remote that is wrapping `local`. If you do nothing, the OS will create `.DS_Store` files which `crypt` will complain about (mixing crypt and non-crypt in the same dir). But if you exclude the `.DS_Store` files with filters, you'll never be able to delete directories through `crypt`.
- You are bisyncing `local` with `drive` on a regular basis. If you do nothing, your `drive` activity history will always be filled with noisy activity logs showing changes in folders which shouldn't have changed -- which could be especially confusing for any other users you're sharing the folder with. But if you exclude the `.DS_Store` files with filters, deleting a directory on `drive` will cause bisync to fail when it tries to delete the corresponding directory on `local`.

Note also that `.DS_Store` files are strictly forbidden on `dropbox`, so `dropbox` users really have no choice but to filter these out and deal with possibly undeleteable directories when syncing `dropbox` to `local`.

Here's another example where this came up recently on the forum:
- https://forum.rclone.org/t/excluded-files-are-excluded-also-in-the-destination-directory-and-never-removed/44487

ncw:
I just had a thought... We have a flag which almost does this already for `sync` that is [--delete-excluded](https://rclone.org/filtering/#delete-excluded-delete-files-on-dest-excluded-from-sync)

I don't know whether this flag works for `bisync` but maybe it should?

What do you think?

nielash:
> I just had a thought... We have a flag which almost does this already for `sync` that is [--delete-excluded](https://rclone.org/filtering/#delete-excluded-delete-files-on-dest-excluded-from-sync)

I think that is different, as here we don't want to delete ALL excluded files -- only the ones with a certain name. 

It is common to use filters to sync only a subset of folders within a drive (but the excluded folders should not get deleted.)

ncw:
> > I just had a thought... We have a flag which almost does this already for `sync` that is [--delete-excluded](https://rclone.org/filtering/#delete-excluded-delete-files-on-dest-excluded-from-sync)
> 
> I think that is different, as here we don't want to delete ALL excluded files -- only the ones with a certain name.
> 
> It is common to use filters to sync only a subset of folders within a drive (but the excluded folders should not get deleted.)

If you put all the directories that you wanted excluding into filters then `--delete-excluded` is exactly the same as `--no-block-rmdir`.

I'm not fond of the UI for `--no-block-rmdir` it seems a bit ad-hoc. At minimum we should make the parameters globs I think.

On the other hand the implementation is very simple and contained which is a big plus.

Do you think users will always have the directories in `--no-block-rmdir` excluded by filters? I wonder if we can combine that with the filtering somehow? Maybe `--exclude-and-delete glob` which could act exactly like `--exclude` but also fill in an internal "no block rmdir" list, (which could be extended to files too?).

What do you think about a name `--delete-excluded-dirs glob` instead of `--no-block-rmdir`?

Sorry, not quite happy with this yet for reasons I'm finding hard to pin down!

PS We can use `slices` now as our minimum is go1.21 :-)

nielash:
> If you put all the directories that you wanted excluding into filters then `--delete-excluded` is exactly the same as `--no-block-rmdir`.

I think I am still not understanding this point. How is it exactly the same?

With `--delete-excluded`, how do I distinguish between the excluded files that should be deleted and the excluded files which should NOT be deleted?

> What do you think about a name `--delete-excluded-dirs glob` instead of `--no-block-rmdir`?

I don't quite understand the "dirs" part, as we are talking about deleting files here, not dirs. (Yes it is for the broader purpose of not blocking removal of its parent dir, if that dir would be getting removed anyway for some other reason. But it is only the file that is excluded with filters -- not the dir. In fact, this feature is only concerned with dirs which are NOT themselves excluded.)

Note also that the goal here is not to `rmdirs` all empty directories. Rather, the goal is to allow individual directories on the `dst` to be removed when (and only when) the corresponding `src` directory has been removed, and when the only thing blocking that removal is the presence of an excluded system file with a particular name. 

> Sorry, not quite happy with this yet for reasons I'm finding hard to pin down!

I wonder if it would be helpful to take inspiration from Syncthing's [similar feature](https://docs.syncthing.net/users/ignoring.html#:~:text=A%20pattern%20beginning%20with%20a%20(%3Fd)%20prefix%20enables%20removal%20of%20these%20files%20if%20they%20are%20preventing%20directory%20deletion.)? (Search that page for `(?d)`)

PR #7514, opened 2023-12-18
check and cryptcheck: report directory differences with --report-dirs - fixes #6440

### LINKED ISSUE #6440: cryptcheck should also check and report directory differences

rclone v1.59.2-beta.6379.603efbfe7
- os/version: debian 11.4 (64 bit)
- os/kernel: 5.10.0-17-amd64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17
- go/linking: dynamic
- go/tags: none

`cryptcheck` currently only compares file differences.  I think it would be useful for it to also report directory differences, that is, if a directory exists on one side but not the other.

It seems logical that any detectable differences should be reported by `cryptcheck`, which would include directory checks.

This is a follow-on to #6437 and #6439.

#### How to use GitHub

nielash:

Before this change, `check` and `cryptcheck` would not inform the user of directory differences (for example, an empty directory that exists only on the src.)

After this change, a new `--report-dirs` flag allows alerting users of such differences.

Note that directories will be reported regardless of whether they are empty. They will be included in the total error count, but counted and summarized separately, and not included in output files. `--report-dirs` is ignored when `--checkfile` is in use.

- https://github.com/rclone/rclone/issues/6440

#### Checklist


cmd/check/check.go
```diff
@@ -41,6 +42,7 @@ func init() {
 // AddFlags adds the check flags to the cmdFlags command
 func AddFlags(cmdFlags *pflag.FlagSet) {
 	flags.BoolVarP(cmdFlags, &oneway, "one-way", "", oneway, "Check one way only, source files must exist on remote", "")
+	flags.BoolVarP(cmdFlags, &ReportDirs, "report-dirs", "", ReportDirs, "Report directory differences in addition to files", "")
```
albertony, line 45: Naming things... We seem to spell out words like "directory" for most flags, and even for crypt backend we have this rather long `--crypt-directory-name-encryption`. However, its not very consistent, as for config we have two flags `--temp-dir` and `--cache-dir`. So I'm thinking if this should be `--report-directories`, or could it simply be `--directories`?

albertony (commented):
I think this looks good. No test, but maybe that's ok here (opt-in, not the most central code). I have a small comment regarding name of the flag, but its not a stopper, more thinking out loud...

cmd/check/check.go
nielash: Either of those would be fine with me! My thought was that `dirs` is consistent with `--create-empty-src-dirs` (the closest comparable flag) and also that shorter is better when it comes to flag names. But I am happy to change it!

albertony: Ah, right, that is one command flag using the short, and plural, "dirs". I looked primarily at global and backend flags, I think. So it is even more mixed than I initially thought.

ncw: He, he!

`--report-dirs` I think is OK! At some point I want to put these flags on sync/copy/move and for that `--dirs` is too generic.

cmd/cryptcheck/cryptcheck.go
```diff
@@ -80,6 +80,8 @@ func cryptCheck(ctx context.Context, fdst, fsrc fs.Fs) error {
 	}
 	defer close()
 
+	opt.ReportDirs = check.ReportDirs
```
ncw, line 83: We shouldn't need this - it should be in the return of `check.GetCheckOpt` already.

fs/operations/check.go
```diff
@@ -38,6 +38,7 @@ type CheckOpt struct {
 	Fdst, Fsrc   fs.Fs     // fses to check
 	Check        checkFn   // function to use for checking
 	OneWay       bool      // one way only?
+	ReportDirs   bool      // report dir differences in addition to files
```
ncw, line 41: This changes the `rc` interface, so the docs need fixing up here

https://github.com/rclone/rclone/blob/94997d25d2d8fa37a3615af5598bd9fb50e19bd5/fs/operations/rc.go#L721

ncw (commented):
This is a nice feature - thank you for implementing it and sorry it has been hanging around so long!

I don't think this is reporting to the report files is it? And if we do that we should make sure directories get reported with a trailing `/` so users can tell them apart from files.

I think it should probably have a test.

I put some other things inline.

Thank you

nielash:
> I don't think this is reporting to the report files is it? And if we do that we should make sure directories get reported with a trailing `/` so users can tell them apart from files.

It does not currently report to the output files (this is mentioned in the commit message.) Do you think it should?

All the other comments made sense -- will fix and add a test.

cmd/cryptcheck/cryptcheck.go
nielash: I could have sworn there was a reason for needing to do this... but 10 months later I can't remember what it was! I guess I'll try removing it and see if anything breaks 😅

ncw:
> > I don't think this is reporting to the report files is it? And if we do that we should make sure directories get reported with a trailing `/` so users can tell them apart from files.
> 
> It does not currently report to the output files (this is mentioned in the commit message.) Do you think it should?

I think that would probably be more useful if it appears in the output files

> All the other comments made sense -- will fix and add a test.

:+1:

PR #7515, opened 2023-12-18
bisync: add overlapping paths check

nielash:
(This PR is dependent on https://github.com/rclone/rclone/pull/7410, https://github.com/rclone/rclone/pull/7497, https://github.com/rclone/rclone/pull/7498, and https://github.com/rclone/rclone/pull/7501 -- only 1 commit is new 🙂 )

Before this change, Bisync did not check to make sure that Path1 and Path2 do not overlap, nor did it check for overlaps with `--backup-dir`. While `sync` does check for these things, it can sometimes be fooled because of the way Bisync calls it with `--files-from` filters. Relying on sync could also leave a run in a half-finished state if it were to error in one direction but not the other (`--backup-dir` only checks for overlaps with the dest.) 

After this change, Bisync does its own check up front, so we can quickly return an error and exit before any changes are made.

- https://github.com/rclone/rclone/issues/5685

#### Checklist


ncw (approved):
Nice :-)

PR #7528, opened 2023-12-23
A proof of concept PR to programmatically upload to Google Photos with an XMP Title

cwwalter:

Due to some recent changes at Google Photos, the Title of photos (as entered in external programs like lightRoom, Apple Photos etc and embedded in the XMP headers) are no longer shown in the google photos interface.  The 'caption' information is written into an "Other" field in the interface but the title (or description) embedded in the file headers are now ignored.  The only way to add that information once it has been uploaded is manually in the web interface for each photo.  So, people who have spent time adding this information will not be able to see use it.

There are several long threads and discussions about this you can find on the web.

Here is a discussion of the issue on the google pages:

https://support.google.com/photos/thread/143046752?hl=en&msgid=145433598

and here is a more technical discussion in the exiftools forum.  The last post points out this can in fact be done programmatically on upload via the API.

https://exiftool.org/forum/index.php?topic=10161.msg80066#msg80066

This stack overflow post shows some detail:

https://stackoverflow.com/questions/72639040/programmatically-set-a-google-photos-caption

I decided to see if I could implement this in rclone.  

This is a proof of principle PR to show that it can be done.  I managed to do this by only 7 lines of modification to googlephotos.go.  Technically, this change adds an ImageTitle entry to the uploadedItem struct.  It then fills it using an exiftools wrapper API and the info is uploaded to google photos via the API along with the AlbumID and UploadToken.  But, here are some caveats:

- I have never programmed in go (or even really read a go program) before.  So, I just did this by looking at the pattern of other code syntax in the file using my general coding knowledge.

- This is not a complete solution.  I completely ignored the batching functionality and assumed there would be only one file per upload.  I'm not familiar with the overall structure of the codebase and batcher.  I think there will need be changes made to make it work generally. When using it, I am using the `--transfers 1` option to require this. (I needed this anyway as I need to make sure the files are uploaded and completed in a fixed order).

- In order to extract the EXIF header information I used the package https://github.com/barasher/go-exiftool that I found through searching.  I am familiar with exiftool which is why I used it, but this is likely overkill for solving this problem.  The capability to extract the title from the header of an image may even be included in the standard libraries for all I know.

I've confirmed this code works for me when using a command like:

```
rclone copy -v --check-first --order-by name --transfers 1 .  photos:album/Album_Name
```

With these small changes, the Title shows up in the "description" at the top of the info panel and also shows as an overlay on the image view. 

So, as I said I am offering this a proof of concept PR, and I hope it is useful to the rclone experts and it would be great if the functionality could be merged into a more complete solution and included in the distribution.

Thanks for rclone! It's really impressive.

I didn't see this discussed in the rclone forums when I looked.

#### Checklist

- [ X] I'm done, this Pull Request is ready for review :-)

### DIFF
--- a/backend/googlephotos/googlephotos.go
+++ b/backend/googlephotos/googlephotos.go
@@ -36,6 +36,8 @@ import (
 	"github.com/rclone/rclone/lib/rest"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+
+	"github.com/barasher/go-exiftool"
 )
 
 var (
@@ -992,6 +994,7 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 // input to the batcher
 type uploadedItem struct {
 	AlbumID     string // desired album
+	ImageTitle  string // Title taken from the EXIF XMP Title
 	UploadToken string // upload ID
 }
 
@@ -1009,6 +1012,7 @@ func (f *Fs) commitBatchAlbumID(ctx context.Context, items []uploadedItem, resul
 	for i := range items {
 		if items[i].AlbumID == albumID {
 			request.NewMediaItems = append(request.NewMediaItems, api.NewMediaItem{
+	            Description: items[i].ImageTitle,
 				SimpleMediaItem: api.SimpleMediaItem{
 					UploadToken: items[i].UploadToken,
 				},
@@ -1125,8 +1129,13 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 		return errors.New("empty upload token")
 	}
 
+	et, err := exiftool.NewExiftool()
+	fileInfos := et.ExtractMetadata(fileName)
+	exifTitle, err := fileInfos[0].GetString("Title")
+
 	uploaded := uploadedItem{
 		AlbumID:     albumID,
+		ImageTitle:  exifTitle,
 		UploadToken: uploadToken,
 	}
 


ncw:
I'll run the CI and you'll see some things you'll need to fix. I particular error handling - what if reading the exif info fails? I'd suggest it is just ignored in this case with a debug log.

> * This is not a complete solution.  I completely ignored the batching functionality and assumed there would be only one file per upload.  I'm not familiar with the overall structure of the codebase and batcher.  I think there will need be changes made to make it work generally. When using it, I am using the `--transfers 1` option to require this. (I needed this anyway as I need to make sure the files are uploaded and completed in a fixed order).

Did you try the batching? Looking at the code I think it should "just work" as you've put the Title in the ImageInfo.

> * In order to extract the EXIF header information I used the package https://github.com/barasher/go-exiftool that I found through searching.  I am familiar with exiftool which is why I used it, but this is likely overkill for solving this problem.  The capability to extract the title from the header of an image may even be included in the standard libraries for all I know.

A lighter-weight non-GPL library would be nice which doesn't shell out to exiftool - maybe https://github.com/dsoprea/go-exif ?

Are there any other things rclone should be adding to the image?

cwwalter:
> I'll run the CI and you'll see some things you'll need to fix. I particular error handling - what if reading the exif info fails? I'd suggest it is just ignored in this case with a debug log.
> 

Sorry... once the holiday ended I had little time to come back to this.  OK, given the batching situation below, I can try to figure out how to do this myself.  But, it will be a little bit until I can work on it.


> > * This is not a complete solution.  I completely ignored the batching functionality and assumed there would be only one file per upload.  I'm not familiar with the overall structure of the codebase and batcher.  I think there will need be changes made to make it work generally. When using it, I am using the `--transfers 1` option to require this. (I needed this anyway as I need to make sure the files are uploaded and completed in a fixed order).
> 
> Did you try the batching? Looking at the code I think it should "just work" as you've put the Title in the ImageInfo.
> 

Ah, you are right!  I thought the batching was done with the list of files first.  If there is only one file at a time when I do the operation it should work. I did some tests with concurrent uploads and it seemed to work properly.

> > * In order to extract the EXIF header information I used the package https://github.com/barasher/go-exiftool that I found through searching.  I am familiar with exiftool which is why I used it, but this is likely overkill for solving this problem.  The capability to extract the title from the header of an image may even be included in the standard libraries for all I know.
> 
> A lighter-weight non-GPL library would be nice which doesn't shell out to exiftool - maybe https://github.com/dsoprea/go-exif ?
> 
> Are there any other things rclone should be adding to the image?

I don't think so. In fact, I think this is the *only* thing we can add.  They are doing something a bit odd.  All the headers are preserved and will still be there if you try to download the file.  Most are reflected in the web display.  However, they have chosen to ignore (only) this 'description' entry, and the only way to add a description is either to explicitly add it on upload as done here, to do it manually in the web interface.  I think there are no other fields that can be added.

ncw:
I'm going to close this as shelling out to exiftool is not what we want, and also requires downloading the source (which may not be local).

I think the idea behind it is good though if we could find an exif library which could work

PR #7529, opened 2023-12-24
bisync: auto-resolve conflicts, customize `--resync` - fixes #7471 #5681

### LINKED ISSUE #7471: bisync: auto-resolve conflicts
This is a proposal for a new bisync feature to optionally auto-resolve sync conflicts (and customize bisync's change detection behavior more generally), inspired in part by discussions with @crocinsocks on #7332 and #5681.

### What problem are you are trying to solve?
Currently, when a file is new/changed on both paths (relative to the prior sync), and the versions on each side are not identical, bisync will keep both versions, renaming them with `..path1` and `..path2` suffixes, respectively. Many users have requested more control over how bisync handles such conflicts -- including an option to automatically select one version as the "winner" and rename or delete the "loser".

Similarly, users have requested more control over how conflicting files are handled during `--resync`. Currently the `path1` version prevails by default, but users have requested options to automatically select the winner based on characteristics such as `newer`, `older`, `larger`, and `smaller`.

Some users have additionally requested the ability to use the above-described auto-resolve features _as an alternative to needing `--resync` at all._ In other words, in an error scenario where bisync would otherwise demand a manual `--resync` to recover, this option would tell bisync to **automatically** run a `--resync` and auto-resolve any conflicts according to the rules described above. Such an option would maximize resilience at the possible expense of perfection (for example, [deleted files could occasionally re-appear](https://github.com/rclone/rclone/issues/5681#:~:text=Bidirectional%20sync%20can,has%20suddenly%20reappeared%3F).) Because bisync takes an approach of "safe by default", such an option would be disabled by default, but made available to advanced users via a new flag.

### How do you think rclone should be changed to solve that?

I propose the following new options:
```
--auto-resolve CHOICE     Automatically resolve conflicts by preferring the version that is: 
                            newer | older | larger | smaller | path1 | path2 (default: none)
--conflict-loser CHOICE   When --auto-resolve selects a winner, do this to the loser:
                            delete | num | pathname (default: num)
``` 
For example, suppose there is a sync conflict of `file` and the version on Path1 is newer.

- With `--auto-resolve newer`, `path2/file` is renamed to `path2/file..conflict1` (or, if that already exists, `path2/file..conflict2`, etc.) and copied to path1, and `path1/file` is copied to path2 (without renaming).
- If `--conflict-loser pathname` is added, path2 version is renamed to `path2/file..path2` instead of `path2/file..conflict1` (overwriting any `path2/file..path2` that already exists.)
- If `--conflict-loser delete` is added, path2 version is deleted instead of renamed.

Options specific to `--resync`:
```
--resync-prefer OPTION    During --resync only, prefer the version that is: newer | older | larger | smaller | path1 | path2 (default: path1)
--auto-resync             If set, automatically run --resync to recover after critical errors. WARNING: only use if you are SURE you know what you are doing! (default: false)
--max-lock DURATION       Consider lock files older than this to be "expired" (default: 1d if --auto-resync=true, otherwise 0 (never expire))
```
`--resync-prefer` and `--auto-resync` can be used together, and either one can also be used without the other.

`--max-lock` is a compromise necessary to balance `--auto-resync` with [lock files](https://rclone.org/bisync/#lock-file). Otherwise, if `--auto-resync` always respected lock files, a single interrupted run would break `--auto-resync` forever. But if `--auto-resync` always _ignored_ lock files, it could start resyncing while a prior non-resync is still running. The implementation of `--max-lock` could possibly include a mechanism for alive runs to renew themselves periodically, to determine with more certainty whether a lock file was left by an interrupted run or one that is still running, just taking awhile.

Note that the `newer` and `older` options require modtime support for both paths, and will be ignored if either remote lacks modtimes (or if the user has requested to ignore them with [`--checksum`](https://github.com/rclone/rclone/issues/5683) or [`--size-only`](https://github.com/rclone/rclone/issues/5684).) 

An open question is what should happen when a winner can't be determined by the requested method. For example: `--auto-resolve newer` where the sizes differ but modtimes are identical. Probably the easiest would be to revert to the current behavior (keep both, do `..path1` / `..path2` rename), but I am open to other ideas!

### Further thoughts

Bisync users tend to fall into one of two camps:

- User A demands perfection every time. If Bisync is ever unsure of what to do, User A would prefer that Bisync stop and ask for help rather than take its best guess and continue.
- User B wants to "set it and forget it", and values resilience above perfection. Often this user is syncing something like a movie collection which could be easily re-downloaded from the internet or backups if necessary, so total perfection isn't required, and sync conflicts may not be worth the time to resolve. User B prefers that Bisync take its best guess and continue, rather than stopping to bother them every time it has a question, or worse -- refusing to run again until it gets an answer.

Bisync in its current form is a great fit for User A, and likely to cause a bit of frustration to User B. The features discussed above are aimed at making both camps happy -- it is User A's way by default, but User B can make it behave their way by adding flags.

Fellow bisync users, feel free to comment here with any thoughts on the above! I'd love to consider any and all feedback before developing a proof-of-concept (which I hope to do shortly.)

### How to use GitHub

nielash:
(This PR is dependent on #7410, #7497, #7498, #7501, and #7515 -- only 5 commits are new 🙂 )

Currently, when a file is new/changed on both paths (relative to the prior sync), and the versions on each side are not identical, bisync will keep both versions, renaming them with `..path1` and `..path2` suffixes, respectively. Many users have requested more control over how bisync handles such conflicts -- including an option to automatically select one version as the "winner" and rename or delete the "loser".

Similarly, users have requested more control over how conflicting files are handled during `--resync`. Currently the `path1` version prevails by default, but users have requested options to automatically select the winner based on characteristics such as `newer`, `older`, `larger`, and `smaller`.

This change introduces support for such options. Documented in great detail in [the docs](https://github.com/nielash/rclone/blob/bisync-resolve/docs/content/bisync.md#--conflict-resolve-choice-conflict-resolve) and commit messages.

- #7471
- #5681

#### Checklist


ncw (approved):
I've given this a light review - it looks good to me :-)

Great feature!

PR #7534, opened 2023-12-30
local: add server-side copy with xattrs on macOS (part-fix #1710)

### LINKED ISSUE #1710: Add extended attributes

nielash:

Before this change, macOS-specific metadata was not preserved by rclone, even for local-to-local transfers (it does not use the "user." prefix, nor is Mac metadata limited to xattrs.) Additionally, rclone did not take advantage of APFS's native "cloning" functionality for fast and deduplicated transfers.

After this change, local (on macOS only) supports "server-side copy" similarly to other remotes, and achieves this by using (when possible) macOS's native APFS "cloning", which is the same underlying mechanism deployed when a user duplicates a file via the Finder UI. This has several advantages over the previous behavior:

- It is extremely fast (even large files can be cloned instantly)
- It is very efficient in terms of storage, as it automatically deduplicates when possible (i.e. so that having two identical files does not consume more storage than having just one.) (The concept is similar to a "hard link", but subsequent modifications will not affect the original file.)
- It preserves Mac-specific metadata to the maximum degree, including not only xattrs but also metadata not easily settable by other methods, including Finder and Spotlight params.

When server-side "clone" is not available (for example, on non-APFS volumes), it falls back to server-side "copy" (still preserving metadata but using more disk storage.) It is only used when both remotes are local (and not wrapped by other remotes, such as crypt.) The behavior of local on non-mac systems is unchanged.

- #1710 (addresses the local-to-local part only)

#### Checklist


nielash:

ncw:
Do you want to try to merge this for v1.67?

nielash:
> Do you want to try to merge this for v1.67?

Yes 🙂

nielash:
> I failed on this PR again - sorry! Can you rebase it please and we'll try to get it in 1.68.
> 
> Thank you :-)

Rebased, and also added [a fix](https://github.com/rclone/rclone/pull/7534/commits/12a0df965de938bf5d384c690a6e735a3e8aa7cd) spotted by one of the newer tests.

ncw (commented):
I think the implementation of this looks great - nothing to add there, however...

> * The concept is similar to a "hard link", but subsequent modifications will not affect the original file.

This makes me think that this should be behind a flag.

This is because if you copy a file from A to B and A and B share (some) disk blocks then they have shared fate. If the disk blocks of A get corrupted then so does B and for a program like rclone which people use for backup that is bad.

For the sftp backend we added

      --sftp-copy-is-hardlink                   Set to enable server side copies using hardlinks

We could also do this for the local backend (probably for all systems if we worked hard enough).

Your patch implements a better solution that hardlinks using the macOS native API to preserve more stuff which is great.

This can also be done on Linux with [ioctl_ficlonerange](https://man7.org/linux/man-pages/man2/ioctl_ficlonerange.2.html) and is supported by quite a few file systems.

So maybe what we should do is gate this behind a flag

      --local-copy-is-clone                   Set to enable server side copies using file cloning

And then we give the user the choice about shared fate of the files and leave the door open for Linux/other implementations.

Gating it behind a flag would remove the commits in the tests also!

We should enable this flag for the testing probably.

What do you think?

nielash:
> This is because if you copy a file from A to B and A and B share (some) disk blocks then they have shared fate. If the disk blocks of A get corrupted then so does B and for a program like rclone which people use for backup that is bad.

If I'm understanding the concern correctly, I think that would apply to hardlinks but _not_ to APFS cloning. [Here's a summary](https://eclecticlight.co/2020/04/14/copy-move-and-clone-files-in-apfs-a-primer/) of the difference:

> APFS clone files (‘copy on write clones’) are in between normal files and hard links. When you make a copy or duplicate of a file to the same volume, APFS creates a new file which shares the same data as the original. As the copy diverges, during editing, the changed data is saved separately, and the common data storage reduces until the files have no shared data at all. This all happens transparently, and there’s no easy way to tell whether two files share common data, nor how much overlaps. This only happens within the same file system and volume, but its effect is seen on free space, which affects all volumes within that container.
> 
> Apple explains this well in the man page for `clonefile()`:
> _The cloned file dst shares its data blocks with the src file but has its own copy of attributes, extended attributes and ACLs which are identical to those of the named file src with the exceptions [of ownership information, setuid and setgid bits]. Subsequent writes to either the original or cloned file are private to the file being modified (copy-on-write)._
> 
> Another approach is to contrast cloning with hard links:
> _With a hard link, two or more references in the file system are fixed to the same object in storage. Any changes made to the data in that object are therefore reflected identically no matter which reference you use to access the data. In a clone, the two references are actually to distinct objects in the file system. When you make changes to the data via one of those references, the data are split for the two references so that they will actually see quite different content._

The thought did occur to me that perhaps this should require the existing `--metadata` flag. I'm not opposed to that (or `--local-copy-is-clone`), but also am not sure it's necessary, as APFS cloning is already the default on modern macs when copy and pasting via the Finder UI, and so I think mac users would already expect this (and might find it cumbersome/unintuitive to require an extra flag to get default behavior.) I don't think Finder even lets you copy instead of cloning (nor can I think of a reason why you might want to, since subsequently modifying one doesn't modify the other.) And there is always `--disable Copy` in any event.

There is also some reverse risk to consider -- that problems like [this one](https://github.com/rclone/rclone/issues/1710#issue-261825215) could occur if mac users who are new to rclone don't realize that xattrs aren't copied by default.

@kapitainsky in case you have any thoughts to add (I think you are a mac user?)

kapitainsky:
> @kapitainsky in case you have any thoughts to add (I think you are a mac user?)

It is true that Finder creates clones by default but `rclone` is command line tool and IMO it should behave like command line `cp` on macOS which creates real copy and not a clone. The latter requires explicit flag:

```
-c    copy files using clonefile(2).  Note that if clonefile(2) is not supported
        for the target filesystem, then cp will fallback to using copyfile(2)
        instead to ensure the copy still succeeds.
```

So I would lean towards `--local-copy-is-clone` flag.

Also as ref-clones are supported by other systems (BTRFS, ZFS on Linux/BSD) maybe we could extend this functionality scope to other OSes as well?

ncw:
@nielash wrote:
> > This is because if you copy a file from A to B and A and B share (some) disk blocks then they have shared fate. If the disk blocks of A get corrupted then so does B and for a program like rclone which people use for backup that is bad.
> 
> If I'm understanding the concern correctly, I think that would apply to hardlinks but _not_ to APFS cloning. [Here's a summary](https://eclecticlight.co/2020/04/14/copy-move-and-clone-files-in-apfs-a-primer/) of the difference:

I think they are a cool idea and much better than hardlinks, but they do still have the shared fate problem if they share data blocks on the disk - there is no avoiding that.

For example, let's say you have a bad sector on the disk. That will affect all files that share that sector so even though you thought you'd made a copy of the file, both your copy and the original are corrupt.

That is me being paranoid about data certainly!

> The thought did occur to me that perhaps this should require the existing `--metadata` flag. I'm not opposed to that (or `--local-copy-is-clone`), but also am not sure it's necessary, as APFS cloning is already the default on modern macs when copy and pasting via the Finder UI, and so I think mac users would already expect this (and might find it cumbersome/unintuitive to require an extra flag to get default behavior.) I don't think Finder even lets you copy instead of cloning (nor can I think of a reason why you might want to, since subsequently modifying one doesn't modify the other.) And there is always `--disable Copy` in any event.

I think preserving the metadata is fine and won't cause the user a problem so happy for it to do so even without the `--metadata` flag.

@kapitainsky wrote:
> It is true that Finder creates clones by default but `rclone` is command line tool and IMO it should behave like command line `cp` on macOS which creates real copies and not clones. The latter requires explicit flag (from `man cp`):
> 
> ```
> -c    copy files using clonefile(2).  Note that if clonefile(2) is not supported
>       for the target filesystem, then cp will fallback to using copyfile(2)
>       instead to ensure the copy still succeeds.
> ```
> 
> So I would lean towards `--local-copy-is-clone` flag.

If we decide later that it is a good idea we can default it to `true`. Regardless some users are going to want to turn it off I think.

> But at the end the key is that clones functionality will be available - default or not IMO it is cosmetic and as long as documented either way is fine.
> 
> Also as ref-clones are supported by other systems (BTRFS, ZFS on Linux/BSD) maybe we could extend this functionality scope to other OSes as well?

I'd like to do that. We can start with the macOS implementation, get that merged and go from there. I think `clonefile(2)` is available directly on BSDs but Linux does it differently - since Linux 4.5 [FICLONE](https://man7.org/linux/man-pages/man2/ioctl_ficlonerange.2.html) is available (`ioctl(int dest_fd, FICLONE, int src_fd);`)

Is `--local-copy-is-clone` a good name for this? The linux term for this kind of copying is `reflink` but I don't think a lot of people know that term - I certainly didn't before I started researching this.

kapitainsky:
> That is me being paranoid about data certainly!

:) I was for a long the same but since I switched to ZFS on my servers I never look back. Using single disk as a mean of backup by creating "real clones" is disastrous practice anyway:). Reflink cloning makes my life so much easier (saving tones of time in the process) that I prefer to add data redundancy than not to have and not to use it. 

> Is `--local-copy-is-clone` a good name for this? The linux term for this kind of copying is `reflink` but I don't think a lot of people know that term - I certainly didn't before I started researching this.

I think "clone" is widely used term precisely describing what is happening when "reflink" rather describes how it is achieved.

e.g. when file level clones were added to OpenZFS, PR was referring to cloning - https://github.com/openzfs/zfs/pull/13392. And the same term is used in Ubuntu ZFS docs - https://manpages.ubuntu.com/manpages/oracular/man7/zfsconcepts.7.html

nielash:
> Is `--local-copy-is-clone` a good name for this?

It's a little long, but I can't think of anything better 🙂 

> Maybe `--local-copy-is-reflink-clone` will remove any disambiguity?

That's even longer 🤣 

> If we decide later that it is a good idea we can default it to `true`. Regardless some users are going to want to turn it off I think.

If that's the case, personally I find the `--local-copy-is-clone=false` syntax a little confusing and think something like `--local-no-clone` (default `false`) would be clearer. But I don't feel strongly.

I do think it would be best to decide now what the default should be (as opposed to releasing it one way with the intent to break it later.)

ncw:
Just a thought...

Will this code cope with copying data to/from an SDcard or similar formated with VFAT/EXFAT?

OK, let's go with the default as on and `--local-no-clone` which can also be spelled `--disable Copy`. I think the benefits probably outweigh the disadvantages.

nielash:
> Just a thought...
> 
> Will this code cope with copying data to/from an SDcard or similar formated with VFAT/EXFAT?

This should be fine, because we're using `COPYFILE_CLONE` and not `COPYFILE_CLONE_FORCE`.

```
/*
 * This routine implements the clonefileat functionality
 * for copyfile. There are 2 kinds of clone flags, namely
 * 1. COPYFILE_CLONE_FORCE which is a 'force' clone flag.
 * 2. COPYFILE_CLONE which is a 'best try' flag.
 * In both cases, we inherit the flags provided
 * to copyfile call and clone the file.
 * Both these flags are equivalent to
 * (COPYFILE_EXCL | COPYFILE_ACL | COPYFILE_STAT | COPYFILE_XATTR | COPYFILE_DATA)
 * With force clone flag set, we return failure if cloning fails,
 * however, in case of best try flag, we fallback to the copy method.
 */
```

Same for older non-APFS macs.

> OK, let's go with the default as on and `--local-no-clone` which can also be spelled `--disable Copy`. I think the benefits probably outweigh the disadvantages.

Sounds good to me. I'll make that change.

nielash:
Added this flag here: https://github.com/rclone/rclone/pull/7534/commits/05094fd0a9588d51561f7af464982c4afbdda1c6

ncw (approved):
This looks great now - thank you @nielash 

I will merge it now :-)

PR #7535, opened 2023-12-30
local: preserve macOS metadata on non-local remotes - fixes #1710

### LINKED ISSUE #1710: Add extended attributes

nielash:

Before this change, macOS-specific metadata was not preserved by rclone, even for local-to-local transfers (it does not use the "user." prefix, nor is Mac metadata limited to xattrs.)

After this change, macOS-specific metadata can be preserved by storing it as an rclone metadata value in "AppleDouble" format (only when the source is `local` on macOS and `--local-apple-double` is set). In the reverse direction (dest is macOS `local`) rclone will automatically "unpack" this metadata to its original form (for instance, as mac xattrs).

Note that these values can sometimes exceed the length limitations imposed by some backends (such as Drive) on metadata values. For this reason, an additional `--local-metadata-max-length` flag has been added to allow skipping values that would otherwise error.

Using the "AppleDouble" format in this way has an advantage over using Go's xattr package, as AppleDouble is the native and idiomatic format for storing macOS metadata and can preserve not just xattrs but also metadata such as Finder and Spotlight params. It is also (arguably) preferable to rsync's approach, which requires a separate `._`prefixed sidecar for every file (hidden on macOS but often visible and messy on other remotes, and which also runs the risk of resource forks becoming separated from their data forks.)

- #1710 

#### Checklist


nielash:

kapitainsky:
As it is something I always hoped to be possible I have tried to test using self build `nielash:macos-xattrs` branch.

Maybe I am doing something wrong but I expected that when I copy:

`rclone . dst: --local-apple-double`

then AppleDouble files will be created at destination and all xattrs restored when reversing copy direction.

But no AppleDouble files are created.

Do I misunderstand what this PR is doing? Of course I am on macOS with APFS filesystem.

nielash:
Hi @kapitainsky, thank you for trying it! 

Instead of creating separate AppleDouble files at the destination, what this is supposed to do is encode the _content_ of the AppleDouble into the `rcapple2` rclone metadata key of the copied file. So if you run `rclone lsjson --stat -M filename` on the non-local side, you should see an `rcapple2` key with a value. Then in the reverse direction (non-local to local), it should "unpack" that value and set all the real mac xattrs just like the source file.

The goal here was to avoid having to create a sidecar for every file on the destination, by essentially storing the content of the sidecar in the metadata of the real file.

However, this feature is not very useful right now for a few reasons: 1. not many rclone backends support `--metadata`, and 2. many of the backends that do support `--metadata` enforce size limits that are too small for our AppleDouble content. (For example, `drive` limits it to 124 bytes in UTF-8 encoding, counting both the key and the value, and even a very small AppleDouble file would be closer to 300 bytes.)

> It works when I copy local to local but not when to onedrive or S3...

I think `onedrive` does not yet support `--metadata`, so that one is expected. But S3 does support `--metadata`. I tried it just now with an `rclone serve s3` remote, and I think I see the problem: we are intentionally dropping binary metadata in s3 due to https://github.com/rclone/rclone/commit/90d23139f626bcaa3a7519698773faac86cb9afb. So I think to make this work with s3, we will have to base64 encode it, which unfortunately will make it even longer. I tried it just now and it worked. Here is a branch with this change: https://github.com/nielash/rclone/tree/macos-xattrs-base64 and examples:

Create a local file and set an xattr:
```
% rclone touch test.txt
% xattr -w hello world test.txt
% xattr -l test.txt
hello: world
```

Copy it to S3:
```
% rclone copyto test.txt serves3:xattrtest/test.txt -M --local-apple-double
% rclone lsjson --stat -M serves3:xattrtest/test.txt        
{
	"Path": "test.txt",
	"Name": "test.txt",
	"Size": 0,
	"MimeType": "text/plain; charset=utf-8",
	"ModTime": "2024-01-21T22:07:57.101075000-05:00",
	"IsDir": false,
	"Tier": "STANDARD",
	"Metadata": {
		"atime": "2024-01-21T22:07:57.101075-05:00",
		"btime": "2024-01-22T03:11:37Z",
		"content-type": "text/plain; charset=utf-8",
		"gid": "20",
		"mode": "100644",
		"mtime": "2024-01-21T22:07:57.101075-05:00",
		"rcapple2": "AAUWBwACAABNYWMgT1MgWCAgICAgICAgAAIAAAAJAAAAMgAAAF8AAAACAAAAkQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQVRUUgAAAAAAAACRAAAAjAAAAAUAAAAAAAAAAAAAAAAAAAABAAAAjAAAAAUAAAZoZWxsbwAAAAB3b3JsZA==",
		"tier": "STANDARD",
		"uid": "501"
	}
}
```

Copy it back and verify xattr:
```
% rclone copyto serves3:xattrtest/test.txt testFromS3.txt -M --local-apple-double
% xattr -l testFromS3.txt
hello: world
```

I wish I had a better answer for the size limit problem. I am curious what you think -- do you think it would be better to try a sidecar approach instead? (As AppleDouble files are inherently sidecars...) It could potentially be implemented as a new backend that wraps `local`. Although I think there would be some big challenges attempting to map 1 object to 2 objects in one direction, and 2 to 1 in the other. I can maybe study how `chunker` accomplishes that. There is also the problem of: what if the files get disconnected on the remote side? And do you really want to see a separate `._file` for every `file` on the remote?

There is also the option of archiving the 2 files together into 1... but then the file is somewhat less usable on the remote.

Thoughts? I am open to suggestions 🙂

kapitainsky:
Thank you for your detailed response. I did some additional tests and now it is 100% clear.

My opinion here is that it is nice to have feature but indeed in practice limited only to local storage.

I think ideally it should work with AppleDouble sidecar model as well. Among Apple users it is known mode of operations and people who need it would understand its limitations. Of course it should not be default as for casual users xattrs are black magic and "._" files only annoyance. But they are life saver when you really need it.

Indeed chunker today uses many to one model. BTW I use it in some cases to store hashes for all files (with `hash_type = sha1all` and large `chunk_size` limit)

nielash:
> So my vote goes for overlay utilising sidecar files solution. I could even imagine that such overlay framework could be used to solve many other existing rclone limitations - like for example storing hashes for crypt etc. macOS xattr storage across any remotes would be fantastic new but such remote could have many more usage in the future.

I like this idea a lot 🙂 

How important do you think it is to have an actual AppleDouble file on the remote vs. (say) a JSON file containing the AppleDouble content in addition to other things (hashes, timestamps, other arbitrary metadata etc.) The JSON version would make this much easier to extend to other things like you mention -- but it wouldn't be a literal AppleDouble file that other tools would recognize as such. Do you think that matters, so long as rclone knows how to restore it properly?

One other thing that occurs to me is that `chunker` only has to wrap the "many" side, whereas we'd need some control over both the source and dest in order for both packing and unpacking of AppleDouble metadata to work (it wouldn't matter as much for other kinds of metadata). I have an idea about how it could be done though.

I'm also thinking that something like the `--sequesterRsrc` option in `ditto` would be useful to make the sidecars less ugly on the remote side.

One big question is whether to make the assumption that `crypt` and `chunker` make  -- that users will not be manipulating the underlying files directly, or something more like `hasher` which does not make that assumption (because it has the luxury of being able to regenerate what it's storing at any time.)

kapitainsky:
> How important do you think it is to have an actual AppleDouble file on the remote vs. (say) a JSON file containing the AppleDouble content in addition to other things (hashes, timestamps, other arbitrary metadata etc.) The JSON version would make this much easier to extend to other things like you mention -- but it wouldn't be a literal AppleDouble file that other tools would recognize as such. Do you think that matters, so long as rclone knows how to restore it properly?

I think universality of JSON approach wins over macOS Apple double sidecar especially given that rclone would handle macOS xattrs anyway. And it would open the door for many interesting development in the future - as effectively it would allow to store any extra data. 


> One big question is whether to make the assumption that `crypt` and `chunker` make -- that users will not be manipulating the underlying files directly, or something more like `hasher` which does not make that assumption (because it has the luxury of being able to regenerate what it's storing at any time.)

My answer to it would be YES - not everything can be "idiot proof". I think that this would be rather advanced overlay for users who know why they are using it. And of course would be nice if it is nicely documented including all possible caveats.

kapitainsky:
One more random thought. 

It would be probably very handy if this solution could (on request) create AppleDouble sidecars e.g.

1. I copy from macOS local to cloud and all xattrs are preserved in JSON sidecars
2. I can copy data back from cloud to macOS local and xattrs will be restores
3. I could copy data from remote to non macOS local/ other cloud without JSON overlay and upon providing `--create-appledouble-sidecar` xattrs would be saved as `._` files.

nielash:
> It would be probably very handy if this solution could (on request) create AppleDouble sidecars

Agreed. And I think that part would be fairly easy.

nielash:
Ok I'm going to convert this to "draft" for now while I work on a better version of this.

In the meantime, I still think #7534 is ready for review (it covers the local-to-local use case only.)

Thanks again @kapitainsky this was very helpful 😃 If you think of other ideas feel free to let me know!

FurloSK:
I just wanted to add that adding this feature would greatly improve the quality of life for [Skim app](https://sourceforge.net/projects/skim-app/) users (a great PDF viewer and IMHO the best PDF note-taker there is on macOS).

**Skim stores all notes and highlights in xattr** (although it can also store their backup in a sidecar ".skim" file), **and without xattr support in Rclone, each attempt to save the PDF fails with a warning, which is very disturbing for user.**

So I thank you very much for your effort to make this work, and hope you will find some time soon to somehow finish the implementation.

All the best, Robert

ncw:
@nielash does this still need to be open?

nielash:
> @nielash does this still need to be open?

No, I think it can be closed, and a new PR can be opened when the right solution is ready.

Btw, what I said [here](https://github.com/rclone/rclone/issues/1710#issuecomment-2294830639) remains true:

> I got pretty far on that project (fully working prototype) but this was before a lot of recent metadata improvements (like directory support!) so I think a lot of it would need to be redone.

> It's still something I'd like to do at some point, but probably won't get to for awhile. If anyone reading this would like to take a stab at it sooner, I'm happy to share my draft as a starting point.

Happily, it looks like I should have more time to devote to rclone stuff in the near future. Finishing this one isn't first on the list, but it's not last, either. 😅

FurloSK:
So, now that this issue is closed, what should I do to subscribe to news about metadata saving? (see [my post](#issuecomment-3452053945) above).

Sorry for a dumb question, but I am not sure why was the issue closed – is it considered solved? Or "won't do"?

nielash:
> So, now that this issue is closed, what should I do to subscribe to news about metadata saving?

I would suggest subscribing to issue #1710 

> I am not sure why was the issue closed – is it considered solved? Or "won't do"?

The issue (#1710) remains open, but this PR (#7535) is closed because it's not the right solution for the problem.

https://github.com/rclone/rclone/pull/7535#issuecomment-1903133973
> this feature is not very useful right now for a few reasons: 1. not many rclone backends support `--metadata`, and 2. many of the backends that do support `--metadata` enforce size limits that are too small for our AppleDouble content. (For example, `drive` limits it to 124 bytes in UTF-8 encoding, counting both the key and the value, and even a very small AppleDouble file would be closer to 300 bytes.)

FurloSK:
ah, sorry, I totally overlooked that this is a pull, not an issue.
Thanks for quick reply!

PR #7541, opened 2024-01-03
oauthutil: avoid panic when `*token` and `*ts.token` are the same

rkonfj:
The field `raw` of `oauth2.Token` may be an uncomparable type(often map[string]interface{}), causing `*token != *ts.token` expression to panic(comparing uncomparable type ...).

The semantics of comparing whether two tokens are the same can be achieved by comparing accessToken, refreshToken and expire to avoid panic.

None.

#### Checklist


ncw (approved):
Good fix - thank you :-)

PR #7550, opened 2024-01-06
serve restic: Fix error handling

MichaelEischer:
We've discovered in https://github.com/restic/restic/issues/4612 that rclone returns a "not found" error instead of an "internal server error" if listing a directory fails. This is very problematic as since restic 0.16.0 a "not found" error while listing a directory is treated like an empty directory. Consequently, the `prune` command can end up believing that there are no snapshots if an error occurs while listing the snapshots and wipe the entire repository.

The second commit modifies the error handling while retrieving a file, such that it only returns "not found" if there's a corresponding error. A future restic version will stop retrying "not found" errors (see https://github.com/restic/restic/pull/4520) and therefore needs to be able to distinguish permanent errors like "not found" from other probably temporary ones.

I haven't added any tests as I'm not sure how to best test the error handling here.

No.

#### Checklist

- ~~[ ] I have added documentation for the changes if appropriate.~~

### DIFF
--- a/cmd/serve/restic/restic.go
+++ b/cmd/serve/restic/restic.go
@@ -408,7 +408,7 @@ func (s *server) deleteObject(w http.ResponseWriter, r *http.Request) {
 
 	if err := o.Remove(r.Context()); err != nil {
 		fs.Errorf(remote, "Delete request remove error: %v", err)
-		if err == fs.ErrorObjectNotFound {
+		if errors.Is(err, fs.ErrorObjectNotFound) {
 			http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
 		} else {
 			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


cmd/serve/restic/restic.go
```diff
@@ -341,7 +341,11 @@ func (s *server) serveObject(w http.ResponseWriter, r *http.Request) {
 	o, err := s.newObject(r.Context(), remote)
 	if err != nil {
 		fs.Debugf(remote, "%s request error: %v", r.Method, err)
-		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+		if err == fs.ErrorObjectNotFound {
```
ncw, line 344: This should probably use `errors.Is` as we do elsewhere in this code.

ncw (commented):
This looks great - see comment inline.

Does the restic `TestBackendRESTExternalServer` tests test this case? We run these tests against rclone on the integration test server.

I think you could add a test here like this

https://github.com/rclone/rclone/blob/1045f5412804c73fb729e7d1ab0a99713aef75f9/cmd/serve/restic/restic_privaterepos_test.go#L68-L76

Not sure why the restic tests are in lots of different files but if you added a similar test to that file that would be perfect.

cmd/serve/restic/restic.go
MichaelEischer: I've also added a commit that replaces the err comparison used when deleting a file.

MichaelEischer:
I've added two small tests to check the new error handling behavior. PTAL

ncw (approved):
Thank you, that looks great. I'll merge it now :-)

Issue #7542, opened 2024-01-05
No access to Linkbox files and dirs


gvitali:
#### The associated forum post URL from `https://forum.rclone.org`
n/a

#### What is the problem you are having with rclone?
Rclone successfully encodes and copies dirs and files to Linkbox, but then no way to sync or list or download it.

The problem is that Linkbox allows files and folders with names of up to 100 characters. But the search in their API works with names of up to 50 ASCII characters and 25 UTF-16 characters (for example, Cyrillic) only. Thus, normal operation is possible only with very short names. For example, using base32 encoding for names, the maximum name length can be only 8 UTF-16 characters, and for base32768 - 16 UTF-16 characters. Which is, of course, very little in real life.

#### What is your rclone version (output from `rclone version`)
rclone v1.65.0 and rclone v1.66.0-beta.7594.5fa13e3e3

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Arch Linux 64 bit

#### Which cloud storage system are you using? (e.g. Google Drive)
Linkbox

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

`./rclone ls -vv "telebox-encoded:photos from London"`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
<7>DEBUG : rclone: Version "v1.65.0" starting with parameters ["rclone" "ls" "-vv" "telebox-encoded:photos from London"]
<7>DEBUG : rclone: systemd logging support activated
<7>DEBUG : Creating backend with remote "telebox-encoded:photos from London"
<7>DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
<7>DEBUG : Creating backend with remote "telebox:files/encoded/347ftlbc73000l927l7hbud1i77qle7j8pfi9j01c5lu0bve5k8g"
<3>ERROR : : error listing: directory not found
<7>DEBUG : 5 go routines active
Failed to ls with 2 errors: last error was: directory not found
```

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

kapitainsky:
And what you want rclone to do about it?

gvitali:
[...]

PR #7551, opened 2024-01-06
linkbox: fix work with names longer than 8-25 Unicode chars.

gvitali:

LinkBox API does not allow searching by more than 25 Unicode characters in the name, for this reason it is currently impossible to work with files and folders named longer than 8 Unicode chars (if encoded in base32). This fix queries all files in a long named directory and checks names there one by one, thus solving the issue.

#7542

#### Checklist

### DIFF
--- a/backend/linkbox/linkbox.go
+++ b/backend/linkbox/linkbox.go
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	maxEntitiesPerPage = 1024
+	maxEntitiesPerPage = 1000
 	minSleep           = 200 * time.Millisecond
 	maxSleep           = 2 * time.Second
 	pacerBurst         = 1
@@ -220,7 +220,7 @@ type listAllFn func(*entity) bool
 //
 // If the name doesn't match this then do an dir list instead
 // N.B.: Linkbox doesn't support search by name that is longer than 50 chars
-var searchOK = regexp.MustCompile(`^[a-zA-Z0-9_ .]{1,50}$`)
+var searchOK = regexp.MustCompile(`^[a-zA-Z0-9_ -.]{1,50}$`)
 
 // Lists the directory required calling the user function on each item found
 //
--- a/backend/linkbox/linkbox.go
+++ b/backend/linkbox/linkbox.go
@@ -219,7 +219,8 @@ type listAllFn func(*entity) bool
 // Search is a bit fussy about which characters match
 //
 // If the name doesn't match this then do an dir list instead
-var searchOK = regexp.MustCompile(`^[a-zA-Z0-9_ .]+$`)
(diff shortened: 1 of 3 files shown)


backend/linkbox/linkbox.go
```diff
@@ -238,6 +239,14 @@ func (f *Fs) listAll(ctx context.Context, dirID string, name string, fn listAllF
 		// If name isn't good then do an unbounded search
 		name = ""
 	}
+
+	nameSearch := name
```
ncw, line 243: Instead of doing this you can just limit the maximum length that is OK in the `searchOK` regexp above

Something like

```go
var searchOK = regexp.MustCompile(`^[a-zA-Z0-9_ .]{1,50}$`)
```

fstest/fstests/fstests.go
```diff
@@ -701,11 +702,15 @@ func Run(t *testing.T, opt *Opt) {
 				{"trailing dot", "trailing dot."},
 				{"invalid UTF-8", "invalid utf-8\xfe"},
 				{"URL encoding", "test%46.txt"},
+				{"many chars", "quite-long-file-name-with-more-than-fifty-chars.txt"},
```
ncw, line 705: Does this make the tests pass?

I note these are failing in the integration tests at the moment. (I'm sure they were passing when we merged linkbox so something has changed!)

ncw (commented):
Very useful - thank you :-)

See comments inline.

backend/linkbox/linkbox.go
gvitali: Thank you for your comment. Yes, you're right, it's more concise.

fstest/fstests/fstests.go
gvitali: This caused an error for Linkbox before the fix, but worked after the fix. However, it is better to move this into a separate commit, as it affects the tests of other remotes.

gvitali:
During Linkbox synchronization I encountered two more problems and fixed them. The maximum number of objects on a page should be no more than 1000. Currently it is 1024, for this reason the listing always ends on the first page with the error “object not found”, rclone tries to upload the file again, Linkbox stores it with the name “filename(N)”, and so the storage fills up indefinitely.
In the second case, I added a hyphen to the list of allowed characters, which makes queries more optimized (no need to load all files in a directory for an entity with a hyphen).

azowrle:
Why is this still not merged? I really need to use Telebox with this fix

ncw:
I rebased this and squashed it down to two commits.

I'll merge when the CI goes green - thank you @gvitali

ncw:
I rebased this again and fixed the conflicts. Hopefully I'll remember to merge it this time!

PR #7559, opened 2024-01-12
Windows installation script

BlazerYoo:
Makes downloading and setting up `rclone.exe` simpler for Windows users.
1. Batch script allows users to select rclone version specific to their architecture.
2. Automatically downloads the selected rclone version and adds it to path.

No.

#### Checklist

### DIFF
--- a/docs/content/install.bat
+++ b/docs/content/install.bat
@@ -1,0 +1,99 @@
+@echo off
+setlocal enabledelayedexpansion
+
+
+REM Check for admin rights
+net session >nul 2>&1
+if %errorLevel% equ 0 (
+    echo Administrative privileges confirmed.
+    echo --------------------------------------------------------
+) else (
+    echo Please run this script as an administrator.
+    pause
+    exit /b 1
+)
+
+
+REM Get windows versions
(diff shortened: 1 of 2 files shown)


ncw:
What do you think @albertony - do you like this install method? It is analogous to our install.sh method for non Windows platforms.

I would have thought

    winget install Rclone.Rclone

would be the preferred installation method for Windows users though @BlazerYoo - when would Windows users want to use this method?

BlazerYoo:
I made this script for Windows users that do not have winget. This script should work for all Windows machines with CMD and Powershell.

albertony:
Some impressive batch-fu in that script!

Though, I could be wrong, but I'm not convinced something like this is that useful for the majority of Windows users...
- In contrast to linux, there are not that many alternatives to choose from, only 3 (amd64, 386 and arm64), and 99% of users should pick the amd64 version, i.e. don't need a script to help choosing.
- The script installs into fixed location `C:\rclone`. Rclone does not deserve its own root folder on my machines, I'm sorry to say! :) But if it did, I would put it in a directory capitalized as `Rclone`, as it better aligns with standard Windows directory names, while others may prefer the latter. In linux the path and name conventions are quite standardized, but in Windows not so much. Well, standard for installed programs is `C:\Program Files`, or rather `%LocalAppData%` these days, but plain copy-install binaries is a bit more individual. For a public install script like this I think I would have gone with `%LocalAppData%\rclone`, which would align with the default config location, in a similar (but not same) path `%AppData%\rclone`. However my point is that there is no standard to follow that would fit all.
- Rclone can easily be installed with winget or chocolatey, for those who prefer convenience over control/customization or whatever...

Maybe a wiki page, and/or a forum thread, would be a better place for it, for now?

Neustradamus:
@BlazerYoo: Any progress on your PR?

PR #7565, opened 2024-01-14
Seafile: Fix download/upload error when FILE_SERVER_ROOT is relative

DanielEgbers:

This PR fixes an error in the Seafile backend implementation.
A Seafile server can be configured to use a relative url as `FILE_SERVER_ROOT` (e.g.: `/seafhttp` instead `https://seafile.example.com/seafhttp`) in order to support more than one hostname/ip (see https://github.com/haiwen/seahub/issues/3398#issuecomment-506920360).
The current backend implementation always expected an absolute download/upload url, resulting in an `unsupported protocol scheme` error. With this PR it supports both absolute and relative.

No

#### Checklist

### DIFF
--- a/backend/seafile/webapi.go
+++ b/backend/seafile/webapi.go
@@ -577,7 +577,7 @@ func (f *Fs) getDownloadLink(ctx context.Context, libraryID, filePath string) (s
 	return result, nil
 }
 
-func (f *Fs) download(ctx context.Context, url string, size int64, options ...fs.OpenOption) (io.ReadCloser, error) {
+func (f *Fs) download(ctx context.Context, downloadLink string, size int64, options ...fs.OpenOption) (io.ReadCloser, error) {
 	// Check if we need to download partial content
 	var start, end int64 = 0, size
 	partialContent := false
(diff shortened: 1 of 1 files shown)


backend/seafile/webapi.go
```diff
@@ -606,9 +606,14 @@ func (f *Fs) download(ctx context.Context, url string, size int64, options ...fs
 	// Build the http request
 	opts := rest.Opts{
 		Method:  "GET",
-		RootURL: url,
 		Options: options,
 	}
+	parsedURL, _ := url.Parse(downloadLink)
```
ncw, line 611: This needs to check the error returned by `url.Parse` if err !=nil then `parsedURL` won't be valid.

backend/seafile/webapi.go
```diff
@@ -688,11 +693,16 @@ func (f *Fs) upload(ctx context.Context, in io.Reader, uploadLink, filePath stri
 
 	opts := rest.Opts{
 		Method:      "POST",
-		RootURL:     uploadLink,
 		Body:        formReader,
 		ContentType: contentType,
 		Parameters:  url.Values{"ret-json": {"1"}}, // It needs to be on the url, not in the body parameters
 	}
+	parsedURL, _ := url.Parse(uploadLink)
```
ncw, line 700: same here...

ncw (commented):
This looks like a nice fix - thank you.

I think the errors from `url.Parse` need checking though...

backend/seafile/webapi.go
DanielEgbers: added error check

backend/seafile/webapi.go
DanielEgbers: added error check

ncw (approved):

Will merge :-)

Issue #7571, opened 2024-01-17
Document rclone OS compatibility requirements


kapitainsky:
As discussed on the forum:

https://forum.rclone.org/t/rclone-1-65-1-runtime-exception-error-crash-immediately-after-running-the-command/44051/7

We should add to rclone docs information about rclone OS compatibility - mostly due to changing golang requirements not all rclone versions can work on some older OS.

The current (v1.65.1) requirements are:

```
OS		Minimum Version
================================
Linux		2.6.32+
macOS		10.15 (Catalina)+
Windows		10+, Server 2016+
```

For historical versions so far based on  - https://go.dev/doc/go1.20:

```
Windows:
rclone v1.63.1 is the last rclone version working for Windows 7, 8, Server 2008 and Server 2012.

macOS:
rclone v1.63.1 is the last rclone version working for macOS 10.13 High Sierra or 10.14 Mojave

```
We have to dig a bit deeper and identify as well the last Windows XP version and older macOS.

kapitainsky:
And here some archeology including the latest rclone version using particular golang version:

```
Go 1.16 is the last release that will run on macOS 10.12 Sierra. Go 1.17 will require macOS 10.13 High Sierra or later.
rclone v1.56.0
- go/version: go1.16.6


Go 1.14 is the last release that will run on macOS 10.11 El Capitan. Go 1.15 will require macOS 10.12 Sierra or later.
rclone v1.52.0
- go version: go1.14.3


Go 1.12 is the last release that will run on macOS 10.10 Yosemite. Go 1.13 will require macOS 10.11 El Capitan or later.
rclone v1.49.0
- go version: go1.12.9


Go 1.10 is the last release that will run on OS X 10.8 Mountain Lion or OS X 10.9 Mavericks. Go 1.11 will require OS X 10.10 Yosemite or later.
Go 1.10 is the last release that will run on Windows XP or Windows Vista. Go 1.11 will require Windows 7 or later.
rclone v1.42
- go version: go1.10
```

I do not think that it is practical to go further back in time.

ncw:
That is a great bit of research - thank you.  If we've got back to go1.10 that was released in Feb 2018 so 6 years ago. Rclone of 6 years ago is unlikely to work well with a lot of the cloud providers so that seems like a good place to stop.

> Given that current rclone version supports Linux kernels 2.6.32+ (from 2003) there is no need to dig Linux further neither IMO.

[...]

PR #7573, opened 2024-01-17
docs: add rclone OS requirements

kapitainsky:
Adds rclone OS requirements list and latest rclone versions known to be working with specific historical OS releases.

Discussed on the forum:
https://forum.rclone.org/t/rclone-1-65-1-runtime-exception-error-crash-immediately-after-running-the-command/44051

Fixes: #7571

### DIFF
--- a/docs/content/downloads.md
+++ b/docs/content/downloads.md
@@ -10,6 +10,17 @@ Rclone is single executable (`rclone`, or `rclone.exe` on Windows) that you can
 simply download as a zip archive and extract into a location of your choosing.
 See the [install](https://rclone.org/install/) documentation for more details.
 
+## Release {{% version %}} OS requirements {#osrequirements}
+
+| OS | Minimum Version | 
+|:-------:|:-------:|
+| Linux | kernel 2.6.32 |
+| macOS (Intel) | 10.15 (Catalina) |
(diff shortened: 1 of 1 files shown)


docs/content/downloads.md
```diff
@@ -10,6 +10,17 @@ Rclone is single executable (`rclone`, or `rclone.exe` on Windows) that you can
 simply download as a zip archive and extract into a location of your choosing.
 See the [install](https://rclone.org/install/) documentation for more details.
 
+## Release {{% version %}} {#release} OS requirements
```
ncw, line 13: The `{#release}` here is an anchor for linking to so `https://url/#anchor`  - it belongs at the end and it needs to be different, so either remove it or put it at the end and call it `{#osrequirements}` say then we'll be able to link to it on the forum.

ncw (commented):
This looks very nice - thank you.

I put a minor comment inline about the anchor tags - if you can fix that up I will merge :-)

kapitainsky:
> This looks very nice - thank you.
> 
> I put a minor comment inline about the anchor tags - if you can fix that up I will merge :-)

Have not noticed that it is anchor tag.

It is corrected now.

ncw (approved):
Perfect - thank you :-)

PR #7578, opened 2024-01-18
genfilters: interactive filter generator for rclone + bisync integration

nielash:
(This PR is dependent on #7410, #7497, #7498, #7501, #7515, #7529, and #7577 -- only 2 commits are new 🙂 )

This adds a new `genfilters` subcommand to interactively generate filters, and also integrates this feature into `bisync` to make editing a `--filters-file` easier and less error-prone.

Detailed docs included in the commits.

- https://forum.rclone.org/t/i-made-an-rclone-filters-generator/43478/9?u=nielash

#### Checklist


docs/static/img/genfilters_demo.gif
ncw, line 1: I'd prefer to keep binary files out of the repo if we can.

When I did a demo for ncdu I used asciinema which worked quite well but having it hosted externally isn't ideal! It looks much nicer than a gif though as you have proper fonts.

ncw (commented):
This is a fun idea :-)

I pulled it locally and had a play.

Some immediate thoughts for discussion

- Needs help - I expect `?` or `h` to bring up the help, likewise CTRL-L to refresh the screen.
- I'd like to be able to edit the include file and see what happens to the files in real time (maybe too hard!). So press TAB say to move to editing the left hand side and TAB to move back again. Or press `e` to edit the include file in your default editor (like the way `git commit` works).
- I thought the default output to a file, then opening that file wasn't very unixy! Printing the result to stdout would probably be a more unixy default
- If you provide an output file, then it should use that file as the default input I think, so genfilters effectively edits a filter file for you. Having to supply two flags is non optimal. Maybe the one flag should be called `--edit`, so you `rclone genfilters --edit myfilters remote:path`
- I think we should try to make it as consistent with ncdu as possible so left, right should be enter exit directory and we should use RETURN or SPACE maybe to toggle the selection (might be worth studying some other text ui progs to see what they use).
- Perhaps ncdu should export some things genfilters should use or we factor ncdu parts into a lib.

nielash:
> This is a fun idea :-)
> 
> I pulled it locally and had a play.

Thanks! 😃 

> * Needs help - I expect `?` or `h` to bring up the help, likewise CTRL-L to refresh the screen.

Sounds good to me.

> * I'd like to be able to edit the include file and see what happens to the files in real time (maybe too hard!). So press TAB say to move to editing the left hand side and TAB to move back again. Or press `e` to edit the include file in your default editor (like the way `git commit` works).

That might be difficult! But also sounds like a fun challenge. 🙂 I'll look into this!

Aside from the challenges of the text editor part, there's also the question of how to handle unexpected input -- syntax errors, non-existent paths, duplicated rules, etc. Error? Ignore? Fix? Also,`genfilters` does not currently output some of the more advanced options like wildcards and regex globs, so I'm not sure what it should do if it receives them in an input file. Convert them to static include/exclude file rules? Delete them? Attempt to preserve them? I'll give this some thought. 🙂

> * I thought the default output to a file, then opening that file wasn't very unixy! Printing the result to stdout would probably be a more unixy default

Haha, fair point! There is a `--no-open` flag, but I agree that should probably just be the default. Printing to stdout in the absence of `-o` also sounds fine to me.

> * If you provide an output file, then it should use that file as the default input I think

What if the input file doesn't exist though? (In other words, you have no file to start from and you want `genfilters` to generate one.)

My first draft of this actually had no `--input-file` feature, only output! Input was a later development. 

> * Having to supply two flags is non optimal. Maybe the one flag should be called `--edit`, so you `rclone genfilters --edit myfilters remote:path`

In addition to my above question about non-existent input, what if you want to "Save as" the output to a different path than the input? Perhaps the input and output flags should stay but an `--edit FILE` flag should be added as shorthand for `--input-file FILE -o FILE`? (sidenote: I tried to grab the `-i` shorthand but it was already in use for `--interactive`...)

> I think we should try to make it as consistent with ncdu as possible so left, right should be enter exit directory and we should use RETURN or SPACE maybe to toggle the selection (might be worth studying some other text ui progs to see what they use).

Sure, that makes sense to me. (Although we don't really "enter" directories the way `ncdu` does -- but I know what you mean!)

> Perhaps ncdu should export some things genfilters should use or we factor ncdu parts into a lib.

That would be helpful! 🙂 I frequently find myself wanting to reuse code from other rclone packages, but often they are in private functions so it becomes a question of whether to start making intrusive changes outside the package I'm working on... I sometimes wish more things were exported by default, but I also understand why they're not!

I'll play around with these ideas and work on a revised version!

docs/static/img/genfilters_demo.gif
nielash: Got it! I'll give that a try for the revised version.

ncw:
X-Reply-Style: better-late-than-never

> > This is a fun idea :-)
> > I pulled it locally and had a play.
> 
> Thanks! 😃
> 
> > * Needs help - I expect `?` or `h` to bring up the help, likewise CTRL-L to refresh the screen.
> 
> Sounds good to me.
> 
> > * I'd like to be able to edit the include file and see what happens to the files in real time (maybe too hard!). So press TAB say to move to editing the left hand side and TAB to move back again. Or press `e` to edit the include file in your default editor (like the way `git commit` works).
> 
> That might be difficult! But also sounds like a fun challenge. 🙂 I'll look into this!
> 
> Aside from the challenges of the text editor part, there's also the question of how to handle unexpected input -- syntax errors, non-existent paths, duplicated rules, etc. Error? Ignore? Fix? Also,`genfilters` does not currently output some of the more advanced options like wildcards and regex globs, so I'm not sure what it should do if it receives them in an input file. Convert them to static include/exclude file rules? Delete them? Attempt to preserve them? I'll give this some thought. 🙂

Good thought. If there is an error in the input file we should just print the error nicely and stop. If the user enters something incorrectly in the editor then we should display an error and make them edit it again or cancel.

> > * I thought the default output to a file, then opening that file wasn't very unixy! Printing the result to stdout would probably be a more unixy default
> 
> Haha, fair point! There is a `--no-open` flag, but I agree that should probably just be the default. Printing to stdout in the absence of `-o` also sounds fine to me.
> 
> > * If you provide an output file, then it should use that file as the default input I think
> 
> What if the input file doesn't exist though? (In other words, you have no file to start from and you want `genfilters` to generate one.)

The unix way would be to pipe the input file on stdin and pipe it to separate output file, or have an `--edit` flag

> My first draft of this actually had no `--input-file` feature, only output! Input was a later development.
> 
> > * Having to supply two flags is non optimal. Maybe the one flag should be called `--edit`, so you `rclone genfilters --edit myfilters remote:path`

That works for me.

(You should preserve the old file with `--edit` - rename it as a backup file maybe? And `--edit` should work with a a non-existent file.
 
> > I think we should try to make it as consistent with ncdu as possible so left, right should be enter exit directory and we should use RETURN or SPACE maybe to toggle the selection (might be worth studying some other text ui progs to see what they use).
> 
> Sure, that makes sense to me. (Although we don't really "enter" directories the way `ncdu` does -- but I know what you mean!)
> 
> > Perhaps ncdu should export some things genfilters should use or we factor ncdu parts into a lib.
> 
> That would be helpful! 🙂 I frequently find myself wanting to reuse code from other rclone packages, but often they are in private functions so it becomes a question of whether to start making intrusive changes outside the package I'm working on... I sometimes wish more things were exported by default, but I also understand why they're not!

Yes we should try to be deliberate about which bits are exported otherwise the codebase will degenerate into a tangled mess!

> I'll play around with these ideas and work on a revised version!

:-)

How did you get on?

nielash:
> How did you get on?

I haven't had a chance to play around with it yet, but I do still plan to! 😅

PR #7594, opened 2024-01-24
drive: backend query command

john-paulsmith:

This command executes a list query in Google Drive’s native query language and returns id, name pairs for any matches. It’s useful for locating files quickly in folders with a large number of files, where rclone’s normal list command is slow due to client-side filtering.

No

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -3420,6 +3420,53 @@ func (f *Fs) copyID(ctx context.Context, id, dest string) (err error) {
 	return nil
 }
 
+func (f *Fs) query(ctx context.Context, query string) (entries []string, err error) {
+	list := f.svc.Files.List()
+	if query != "" {
+		list.Q(query)
+	}
+
+	if f.opt.ListChunk > 0 {
+		list.PageSize(f.opt.ListChunk)
+	}
+	list.SupportsAllDrives(true)
+	list.IncludeItemsFromAllDrives(true)
+	if f.isTeamDrive && !f.opt.SharedWithMe {
+		list.DriveId(f.opt.TeamDriveID)
+		list.Corpora("drive")
+	}
+	// If using appDataFolder then need to add Spaces
+	if f.rootFolderID == "appDataFolder" {
+		list.Spaces("appDataFolder")
+	}
+
+	fields := fmt.Sprintf("files(%s),nextPageToken,incompleteSearch", f.getFileFields(ctx))
+
+	var results []string
+	for {
+		var files *drive.FileList
+		err = f.pacer.Call(func() (bool, error) {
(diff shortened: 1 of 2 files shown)


backend/drive/drive.go
```diff
@@ -3570,6 +3617,19 @@ Use the --interactive/-i or --dry-run flag to see what would be copied before co
...
+	Short: "List files using Google Drive query language",
+	Long: `This command lists files based on a query
+
+Usage:
+
+    rclone backend query drive: query
+    
+The query syntax is documented at [Google Drive Search query terms and 
+operators](https://developers.google.com/drive/api/guides/ref-search-terms).
+
```
ncw, line 3631: It would be great to have an example here.

backend/drive/drive.go
```diff
@@ -3570,6 +3617,19 @@ Use the --interactive/-i or --dry-run flag to see what would be copied before co
...
+Results are tab-separated id, name pairs.`,
```
ncw, line 3632: My instincts are to have a more fully featured output format than TSV like JSON as it is difficult to predict what people want from this sort of command in advance and adding things to JSON is much easier without breaking backwards compatibility.

I'd probably just return a `[]*drive.File` from `query` and then return that from the backend command which will JSON-ify it.

What do you think?

We could always have an option `-o format=tsv` ?

ncw (commented):
I think this is a very useful feature - thank you!

See inline comments.

backend/drive/drive.go
john-paulsmith: I've added a couple of examples, including one covering escaping.

backend/drive/drive.go
john-paulsmith: I didn't realise it would produce JSON for free! I've removed the TSV formatting and will adjust our calling code or pipe the output through jq as needed.

backend/drive/drive.go
```diff
@@ -3420,6 +3420,53 @@ func (f *Fs) copyID(ctx context.Context, id, dest string) (err error) {
...
+			return f.shouldRetry(ctx, err)
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute query: %w", err)
+		}
+		if files.IncompleteSearch {
+			fs.Errorf(f, "search result INCOMPLETE")
+		}
+		for _, item := range files.Files {
+			item.Name = f.opt.Enc.ToStandardName(item.Name)
```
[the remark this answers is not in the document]
john-paulsmith: I've removed this code now.

john-paulsmith:
Thanks very much for your review. I've pushed another commit incorporating these changes.

ncw (approved):

Sorry for the delay in merging.

PR #7599, opened 2024-01-27
Check: Report differences and error counts separately

albertony:

On code inspection I noticed that the reported number of differences and errors after a check operation would always the same. It reports `"%d differences found"` using `accounting.Stats(ctx).GetErrors()`, followed by `"%d errors while checking"` using the exact same parameter, only difference is that "differences" is always reported while "errors" is only reported if `>0`. I cannot see a reason why this should be of any importance.

#### Checklist


ncw (approved):
That definitely looks like a bug left over from some previous refactor - well spotted :-)

PR #7602, opened 2024-01-27
Changes to exit codes

albertony:

According to rclone [documentation](https://rclone.org/docs/#list-of-exit-codes), exit code 1 is for "syntax or usage" and 2 is for "not otherwise categorised" errors. However, in current implementation, exit code 2 will never be returned (*with one exception in bisync, see below), and in reality exit code 1 will be used for anything other than the "otherwise categorised" errors - exit code 1 is the *default* for unsuccessful returns.

There seem to be a more or less convention for exit codes in Linux/shell world,  maybe some specifically for bash:
- Values between 0 and 255
- 0 is for success
- 1 is for general errors
- 2 is usage errors, invalid options or missing arguments.
- Values above 125 are specifically used by the shell (bash), e.g. when program is terminated on response to a signal.

(https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html)
(https://www.redhat.com/sysadmin/exit-codes-demystified)

This PR does the following changes to exit codes from rclone:
- Explicitely checks for "syntax or usage" type errors from rclone and return exit code 2 in those cases. E.g. `rclone copy a`, `rclone copy a b c` or `rclone copy --unknownflag a b`.
- Explicitely return exit code 2 when cobra command processor returns before rclone's command handling, where the resolving of exit codes occur. E.g. when running `rclone unknowncommand`, or simply `rclone`, this now returns exit code 2 instead of 1.
- Keep using exit code 1 as *default* for unsuccessful returns, which needs no change in existing code, but an update to the documentation so that it correctly reflects this - now after the previous point by switching the description of exit code 1 and 2.

\* **bisync exception:** There was actually one case of exit code 2, in bisync, where it explicitely called `os.Exit` without going through `resolveExitCode`. I added a commit to change that to exit code 7, more in-line with rest of rclone, and updated the description of exit codes in bisync documentation according to these changes affecting exit codes 1, 2 and 7.

https://forum.rclone.org/t/minor-bug-return-code-from-rclone-check/44236
- That thread started out by questioning why the check command returned exit code 1, documented as "syntax or usage error", when differences were found. With the changes in this PR it will still return exit code 1, but this is now documented as "error not otherwise categorised", and in case of "syntax or usage error" one will get a separate exit code 2. So this somewhat improves the situation, although the suggestion in that thread was to introduce a (or maybe several?) new exit codes for reporting result from a check, which could still be considered..

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -50,7 +50,6 @@ var (
 	version       bool
 	// Errors
 	errorCommandNotFound    = errors.New("command not found")
-	errorUncategorized      = errors.New("uncategorized error")
 	errorNotEnoughArguments = errors.New("not enough arguments")
 	errorTooManyArguments   = errors.New("too many arguments")
 )
@@ -495,8 +494,6 @@ func resolveExitCode(err error) {
 		os.Exit(exitcode.DirNotFound)
 	case errors.Is(err, fs.ErrorObjectNotFound):
 		os.Exit(exitcode.FileNotFound)
-	case errors.Is(err, errorUncategorized):
-		os.Exit(exitcode.UncategorizedError)
 	case errors.Is(err, accounting.ErrorMaxTransferLimitReached):
 		os.Exit(exitcode.TransferExceeded)
 	case errors.Is(err, fssync.ErrorMaxDurationReached):
@@ -507,8 +504,10 @@ func resolveExitCode(err error) {
 		os.Exit(exitcode.NoRetryError)
 	case fserrors.IsFatalError(err):
 		os.Exit(exitcode.FatalError)
-	default:
+	case errors.Is(err, errorCommandNotFound), errors.Is(err, errorNotEnoughArguments), errors.Is(err, errorTooManyArguments):
 		os.Exit(exitcode.UsageError)
+	default:
+		os.Exit(exitcode.UncategorizedError)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


ncw (approved):
This looks very nice and unifies the bisync error codes nicely.

I think having 1 as uncategorized and 2 as syntax error is much more sensible as the "default" exit code tends to be 1 and hence uncategorized.

cmd/cmd.go
```diff
@@ -567,6 +566,7 @@ func Main() {
 		if strings.HasPrefix(err.Error(), "unknown command") && selfupdateEnabled {
 			Root.PrintErrf("You could use '%s selfupdate' to get latest features.\n\n", Root.CommandPath())
 		}
-		log.Fatalf("Fatal error: %v", err)
+		log.Printf("Fatal error: %v", err)
+		os.Exit(exitcode.UsageError)
```
albertony, line 570: @ncw What about this part, is it safe (enough) to assume that whatever ends up here should be exit code 2? I.e. whatever not ending up in resolveExitCode or in any other os.Exit (log.Fatalf) elsewhere. I think typically the `rclone nonexisting` case must be handled here, but not sure what else. The alternative would be to move the change into the "unknown command" if above, and perhaps extend it if needed, but what do you think?

ncw:
Do you want to merge this for 1.68?

albertony:
I guess we can do that - If you think its ok to do the change to exit code to 2 on return from main, which I had a little question mark on in inline comment. 

If you think its ok, feel free to merge if you want it done today, I'll logoff very soon.

ncw (approved):
I like this change - it tidies up the exit codes nicely.

I think we should merge for 1.69 though as we've probably put enough potentially breaking changes in 1.68 already (eg AWS S3 SDKv2) so we can cook this one for longer in the 1.69 release cycle.

ncw:
Yes please merge :-)

PR #7618, opened 2024-02-06
move: add `--leave-root` to control `--delete-empty-src-dirs` behavior - fixes #7478

### LINKED ISSUE #7478: --delete-empty-src-dirs improvement on rclone move (--leave-root=false)

https://forum.rclone.org/t/rclone-rcd-move-folder-doesnt-delete-empty-source-folder/15080

rclone v1.62.2
- os/version: ubuntu 22.04 (64 bit)
- os/kernel: 5.15.0-83-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.20.2
- go/linking: static
- go/tags: none

delete the root directory after completed move if the folder is empty. There will be some cases that can be left files when using filters.
`rclone move /local/test/dir remote:bucket/test/dir`
the folder `/local/test/dir` is leaved empty.

just remove the root directory if the folder is empty. `--leave-root=false` flag as @ncw noted.

```
rclone move /local/test/dir remote:bucket/test/dir --no-traverse --fast-list --delete-empty-src-dirs --leave-root=false -v -P
Error: unknown flag: --leave-root
Usage:
  rclone move source:path dest:path [flags]

Flags:
      --create-empty-src-dirs   Create empty source dirs on destination after move
      --delete-empty-src-dirs   Delete empty source dirs after move
  -h, --help                    help for move

2023/12/01 17:20:06 Fatal error: unknown flag: --leave-root
```

#### How to use GitHub

nielash:

Before this change, it was not always possible to delete an empty source root directory during `rclone move`. Furthermore, the behavior was inconsistent based on whether the move was performed server-side or via the fallback to copy + delete.

After this change, a `--leave-root` flag is added to control whether `--delete-empty-src- dirs` deletes an empty src root. (It has no effect if `--delete-empty-src-dirs=false` or if the src root isn't empty.)

Note that `--leave-root` defaults to `false`, which is arguably a breaking change, although also arguably not, since the prior behavior was already inconsistent (the root was, in fact, still deleted for server-side moves without filters.) Defaulting to false also makes it consistent with the identical option in `rclone rmdirs`.

This setting should be particularly helpful for users wanting to rename a directory in place.

Note that both `--delete-empty-src-dirs` and `--leave-root` are effectively ignored for server-side directory move operations without filters, which always use the equivalent of `--delete-empty-src-dirs=true --leave-root=false`. (This is not a change, but was not previously documented.)

- #7478 
- https://forum.rclone.org/t/rclone-rcd-move-folder-doesnt-delete-empty-source-folder/15080

#### Checklist


ncw (commented):
I missed reviewing this before - sorry!

I think changing the defaults is not without risk - there is just bound to be a user or two relying on the fact that --delete-empty-src-dirs doesn't delete the root sometimes.

However I agree that this is very inconsistent and makes the rename case impossible, so arguably a bug.

The `--leave-root` flag should probably be on the `rclone move` command directly like `--delete-empty-src-dirs` is rather being a global though.

What do you think?

Do you want to get this into 1.68?

nielash:
> The `--leave-root` flag should probably be on the `rclone move` command directly like `--delete-empty-src-dirs` is rather being a global though.

I think `--delete-empty-src-dirs` is only able to pull that off because the `moveDir` functions take a dedicated `deleteEmptySrcDirs` parameter... are you suggesting that we change the function signatures to add another one? That seems like the more invasive change to me...

ncw:
> > The `--leave-root` flag should probably be on the `rclone move` command directly like `--delete-empty-src-dirs` is rather being a global though.
> 
> I think `--delete-empty-src-dirs` is only able to pull that off because the `moveDir` functions take a dedicated `deleteEmptySrcDirs` parameter... are you suggesting that we change the function signatures to add another one? That seems like the more invasive change to me...

Yes that will be more invasive, certainly.

I don't think we want a --leave-root flag on the main binary which is only useful for move though (I know there lots already but I've been trying not to add more!)

We could change the function to take a moveOpts struct instead which is nice to use on the RC. And is the way the check functions work.

Or if we do want to make a global then it needs a more precise name like --leave-root-on-move but I'm not keen on that. 

I haven't looked through the consequences of changing the move functions yet though so there may be more to it.

It might be that a bigger reorg should be done with sync copy and move.

Thoughts?

PR #7620, opened 2024-02-08
vfs: fix unicode normalization on macOS - fixes #7072

### LINKED ISSUE #7072: new feature: VFS auto normalization
Immediate goal of this enhancement is to fix macOS `rclone mount` which at the moment does not work properly regardless of mounting method used - macOSFUSE or FUSE-T. Optionally normalized VFS can bring other benefits as well e.g. some less common or future filesystems rclone users can work with and normalization constraints imposed.

The background issue and possible solutions have been discussed on the forum:

https://forum.rclone.org/t/possible-special-character-encoding-issue-on-macos/39048/30

At the moment many cloud storage destinations (and rclone virtual crypt remote) allow to store both NFC and NFD encoded files/directories names. Also rclone allows to upload/download both forms.

This in general works trouble free on Linux and Windows but breaks rclone mount functionality in macOS - depending on mount options either only NFC or only NFD content is visible/accessible - this is [unfortunate](https://eclecticlight.co/2021/05/08/explainer-unicode-normalization-and-apfs/) consequence of Apple design decision to be different than rest of the world and use exclusively NFD in their APIs. We won't change it.

Experimentation and testing, including [attempts](https://github.com/macos-fuse-t/fuse-t/issues/16) to improve situation with FUSE-T (it has now - since v1.0.20 - `-o nfc` option which should help with NFC path) lead me to believe that last missing part of the puzzle is rclone VFS lack of normalization.

I would like to propose to add optional normalization to VFS.

New option:

--vfs-normalize OFF/NFC/NFD - with OFF being default 

When set to NFC or NFD it would apply characters normalization to served content.

Edge case to watch is situation where two names differ only by normalization and coexist in the same folder e.g.:

```
$ rclone lsf crypt:TEST

CONFLICTéééFILE.txt
CONFLICTéééFILE.txt
```

one file is NFC and other is NFD. Only solution I see for such case is use only one name and log ERROR.

nielash:

Before this change, the VFS layer did not properly handle unicode normalization, which caused problems particularly for users of macOS. While attempts were made to handle it with various `-o modules=iconv` combinations, this was an imperfect solution, as no one combination allowed both NFC and NFD content to simultaneously be both visible and editable via Finder.

After this change, the VFS supports `--no-unicode-normalization` (default `false`) via the existing `--vfs-case-insensitive` logic, which is extended to apply to both case insensitivity and unicode normalization form.

This change also adds an additional flag, `--vfs-block-norm-dupes`, to address a probably rare but potentially possible scenario where a directory contains multiple duplicate filenames after applying case and unicode normalization settings. In such a scenario, this flag (disabled by default) hides the duplicates. This comes with a performance tradeoff, as rclone will have to scan the entire directory for duplicates when listing a directory. For this reason, it is recommended to leave this disabled if not needed. However, macOS users may wish to consider using it, as otherwise, if a remote directory contains both NFC and NFD versions of the same filename, an odd situation will occur: both versions of the file will be visible in the mount, and both will appear to be editable, however, editing either version will actually result in only the NFD version getting edited under the hood. `--vfs-block-norm-dupes` prevents this confusion by detecting this scenario, hiding the duplicates, and logging an error, similar to how this is handled in `rclone sync`.

- #7072 
- https://forum.rclone.org/t/possible-special-character-encoding-issue-on-macos/39048/30
- https://forum.rclone.org/t/some-unicode-forms-break-mount-on-macos-with-fuse-t/36403

#### Checklist

### DIFF
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -116,13 +116,6 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 	for _, option := range opt.ExtraFlags {
 		options = append(options, option)
 	}
-	if runtime.GOOS == "darwin" {
-		if !findOption("modules=iconv", options) {
-			iconv := "modules=iconv,from_code=UTF-8,to_code=UTF-8-MAC"
-			options = append(options, "-o", iconv)
-			fs.Debugf(nil, "Adding \"-o %s\" for macOS", iconv)
-		}
-	}
 	return options
 }
 
--- a/vfs/dir.go
+++ b/vfs/dir.go
@@ -18,6 +18,7 @@ import (
 	"github.com/rclone/rclone/fs/operations"
 	"github.com/rclone/rclone/fs/walk"
 	"github.com/rclone/rclone/vfs/vfscommon"
+	"golang.org/x/text/unicode/norm"
 )
 
 // Dir represents a directory entry
@@ -466,7 +467,6 @@ func (d *Dir) AddVirtual(leaf string, size int64, isDir bool) {
 		node = f
 	}
 	d.addObject(node)
-
 }
 
 // delObject removes an object from the directory
@@ -512,6 +512,35 @@ func (d *Dir) _readDir() error {
 		// create directories on the fly
 	} else if err != nil {
 		return err
+	}
+
+	if d.vfs.Opt.BlockNormDupes { // do this only if requested, as it will have a performance hit
+		ci := fs.GetConfig(context.TODO())
+
+		// sort entries such that NFD comes before NFC of same name
+		sort.Slice(entries, func(i, j int) bool {
+			if entries[i] != entries[j] && fs.DirEntryType(entries[i]) == fs.DirEntryType(entries[j]) && norm.NFC.String(entries[i].Remote()) == norm.NFC.String(entries[j].Remote()) {
+				if norm.NFD.IsNormalString(entries[i].Remote()) && !norm.NFD.IsNormalString(entries[j].Remote()) {
+					return true
+				}
+			}
+			return entries.Less(i, j)
+		})
+
+		// detect dupes, remove them from the list and log an error
+		normalizedNames := make(map[string]struct{}, entries.Len())
+		filteredEntries := make(fs.DirEntries, 0)
+		for _, e := range entries {
+			normName := fmt.Sprintf("%s-%T", operations.ToNormal(e.Remote(), !ci.NoUnicodeNormalization, (ci.IgnoreCaseSync || d.vfs.Opt.CaseInsensitive)), e) // include type to track objects and dirs separately
+			_, found := normalizedNames[normName]
+			if found {
+				fs.Errorf(e.Remote(), "duplicate normalized names detected - skipping")
+				continue
+			}
+			normalizedNames[normName] = struct{}{}
+			filteredEntries = append(filteredEntries, e)
+		}
+		entries = filteredEntries
 	}
 
 	err = d._readDirFromEntries(entries, nil, time.Time{})
@@ -767,15 +796,18 @@ func (d *Dir) stat(leaf string) (Node, error) {
 	}
 	item, ok := d.items[leaf]
 
-	if !ok && d.vfs.Opt.CaseInsensitive {
-		leafLower := strings.ToLower(leaf)
+	ci := fs.GetConfig(context.TODO())
+	normUnicode := !ci.NoUnicodeNormalization
+	normCase := ci.IgnoreCaseSync || d.vfs.Opt.CaseInsensitive
+	if !ok && (normUnicode || normCase) {
+		leafNormalized := operations.ToNormal(leaf, normUnicode, normCase) // this handles both case and unicode normalization
 		for name, node := range d.items {
-			if strings.ToLower(name) == leafLower {
+			if operations.ToNormal(name, normUnicode, normCase) == leafNormalized {
 				if ok {
-					// duplicate case insensitive match is an error
-					return nil, fmt.Errorf("duplicate filename %q detected with --vfs-case-insensitive set", leaf)
+					// duplicate normalized match is an error
+					return nil, fmt.Errorf("duplicate filename %q detected with case/unicode normalization settings", leaf)
 				}
-				// found a case insensitive match
+				// found a normalized match
 				ok = true
 				item = node
 			}
--- a/vfs/vfs_case_test.go
+++ b/vfs/vfs_case_test.go
@@ -5,10 +5,12 @@ import (
 	"os"
 	"testing"
 
+	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fstest"
 	"github.com/rclone/rclone/vfs/vfscommon"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/text/unicode/norm"
 )
 
 func TestCaseSensitivity(t *testing.T) {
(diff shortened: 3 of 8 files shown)


kapitainsky:
Thank you very much for tackling this pesky issue with your PR! macOS users will be grateful:)

I have been testing this feature for the last few days and it works like a charm with whatever NFC/NFD objects I throw at mount or remote. And everything is accessible on macOS using both terminal and Finder.

This looks like true solution for years old problem with NFC/NFD on macOS.  It works perfectly with both macFUSE and FUSE-T. No more any snake oil `iconv` options needed - it never really fully worked anyway.

@ncw I see that #7072 is on `v1.66` [milestones list](https://github.com/rclone/rclone/milestone/54) (it was also on `v1.65`:)) so hopefully you can include it in the upcoming rclone release.

kapitainsky:
Is there any issue stopping this change from being merged?

kapitainsky:
I am sorry for pestering but as this is something causing issues in my daily rclone usage I keep a close eye on this possible fix.

@ncw Has this change have a chance to be included in the next rclone release? Or is there still something needed required to do it?

vfs/dir.go
```diff
@@ -514,6 +515,41 @@ func (d *Dir) _readDir() error {
...
+				}
+			}
+			return entries.Less(i, j)
+		})
+
+		// detect dupes, remove them from the list and log an error
+		normalizedNames := make(map[string]struct{}, entries.Len())
+		filteredEntries := make(fs.DirEntries, 0)
+		for _, e := range entries {
+			normName := fmt.Sprintf("%s-%T", operations.ApplyTransforms(ctx, e.Remote()), e) // include type to track objects and dirs separately
```
ncw, line 541: Maybe we should have a version of ApplyTransforms we say explicitly which transforms we want - might save mucking about with contexts

ncw (commented):
I think this looks sound :-)

I have some quibbles with the use of the global config though - see inline!

Thank you !

nielash:
@ncw very helpful feedback, thank you! I think `fs.AddConfig` will work well for the actual code, but I'm not sure it will work for passing options from the tests, because it seems that the VFS methods do not actually take a `ctx` function parameter -- they all just use `context.TODO()`. Should I try to change them to take a function param?

Another option is maybe to make a new VFS option, although it seems not ideal to have lots of options that kind of do the same thing (like we kind of already have with `vfs.Opt.CaseInsensitive` and `ci.IgnoreCaseSync`). 

Whatever you think is best is fine with me 🙂

nielash:
Here's a revised version that addresses most of the feedback. The production code now uses `fs.AddConfig`. The one change I did not fully make is changing the _test_ code to use `fs.AddConfig`, because of what I mentioned in my [previous comment](https://github.com/rclone/rclone/pull/7620#issuecomment-1971386456). For now I have addressed that concern in a different way (by restoring the prior value after the test), but I remain open to different solutions if you prefer!

vfs/dir.go
```diff
@@ -767,15 +805,22 @@ func (d *Dir) stat(leaf string) (Node, error) {
 	}
 	item, ok := d.items[leaf]
 
-	if !ok && d.vfs.Opt.CaseInsensitive {
-		leafLower := strings.ToLower(leaf)
+	ctx := context.TODO()
+	ci := fs.GetConfig(ctx)
```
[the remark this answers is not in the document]
nielash: Made these changes.

cmd/cmount/mount.go
```diff
@@ -116,13 +116,14 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
 	for _, option := range opt.ExtraFlags {
 		options = append(options, option)
 	}
-	if runtime.GOOS == "darwin" {
+	// disabled in favor of ci.NoUnicodeNormalization -- see https://github.com/rclone/rclone/issues/7072#issuecomment-1604745556
+	/* if runtime.GOOS == "darwin" {
 		if !findOption("modules=iconv", options) {
 			iconv := "modules=iconv,from_code=UTF-8,to_code=UTF-8-MAC"
 			options = append(options, "-o", iconv)
 			fs.Debugf(nil, "Adding \"-o %s\" for macOS", iconv)
```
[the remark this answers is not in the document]
nielash: Done :)

vfs/dir.go
```diff
@@ -514,6 +515,41 @@ func (d *Dir) _readDir() error {
 		return err
 	}
 
+	if d.vfs.Opt.BlockNormDupes { // do this only if requested, as it will have a performance hit
+		ciLock.Lock()
+		defer ciLock.Unlock()
+		ctx := context.TODO()
+		ci := fs.GetConfig(ctx)
```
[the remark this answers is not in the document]
nielash: Made this change.

vfs/vfs_case_test.go
```diff
@@ -86,6 +88,8 @@ func TestCaseSensitivity(t *testing.T) {
 	assert.NotEqual(t, err, ENOENT)
 
 	// Run the same set of checks with case-Sensitive VFS, for comparison.
+	ci := fs.GetConfig(ctx)
```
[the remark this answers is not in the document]
nielash: I just removed this as it's no longer needed.

vfs/dir.go
nielash: Interesting idea! I'm happy to make one. I didn't do it yet as I'm not sure how much it would buy us here, since we would still need to at least read the global config to get the `ci.IgnoreCaseSync` and `ci.NoUnicodeNormalization` values (and we'd still need some way of setting those values from the tests).

vfs/vfs_case_test.go
```diff
@@ -160,3 +164,34 @@ func assertFileAbsentVFS(t *testing.T, vfs *VFS, name string) {
...
+	opt := vfscommon.DefaultOpt
+	vfs := New(r.Fremote, &opt)
+	defer cleanupVFS(t, vfs)
+
+	// assert that both files are found under NFD-normalized names
+	assertFileDataVFS(t, vfs, norm.NFD.String(both), "data1")
+	assertFileDataVFS(t, vfs, nfd, "data2")
+
+	// change ci.NoUnicodeNormalization to true and verify that only file1 is found
+	ci := fs.GetConfig(ctx)
```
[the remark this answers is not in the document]
nielash: I addressed this in a different way for now -- see [comment here](https://github.com/rclone/rclone/pull/7620#issuecomment-1975656951) about why, and some possible alternatives.

vfs/dir.go
ncw: I think it is a bit of a code smell to to having to adjust a `ctx` - we should use function parameters for passing values to functions.

From the [Go docs](https://pkg.go.dev/context) this is what contexts are for:

> Package context defines the Context type, which carries deadlines, cancellation signals, and other request-scoped values across API boundaries and between processes.

We've wedged configuration in there but I think changing it instead of making a function with more parameters is probably against the spirit of what context is for!

ncw (commented):
I think your use of context in the tests is fine! The VFS doesn't have contexts and that is because the underlying Go interfaces don't take contexts which is unfortunate.

However on further reflection I think we should try to avoid having to change the context in the main code - see the notes I put inline! Reading it is fine and expected but changing it is not ideal.

nielash:
Makes sense! Here's a new version with this changed.

vfs/dir.go
nielash: Got it -- made this change.

vfs/dir.go
```diff
@@ -767,15 +796,18 @@ func (d *Dir) stat(leaf string) (Node, error) {
 	}
 	item, ok := d.items[leaf]
 
-	if !ok && d.vfs.Opt.CaseInsensitive {
-		leafLower := strings.ToLower(leaf)
+	_, ci := fs.AddConfig(context.TODO())
```
nielash, line 799: Should this be changed back to `fs.GetConfig` now? (Does it matter?)

ncw: Yes it should be just `fs.GetConfig` if we don't need to change the config

ncw (approved):

I'll merge it.

PR #7635, opened 2024-02-14
config: add options for OAuth functionality

### LINKED ISSUE #7634: OAuth 2.0 Compatibility for Docker / Headless Runtimes
https://forum.rclone.org/t/fully-automated-in-docker/44204

1.65.1

rclone cannot complete an OAuth 2.0 authentication when running headless or in docker for two main reasons:
1. In Docker, the default bind of the OAuth callback listener to localhost does not allow network connection to outside of the container. You must bind to 0.0.0.0 when inside a Docker container.
2. Generic to headless mode, something (a reverse proxy) needs to pass the callback back to rclone and it's specific port. This is only possible if rclone uses a redirect URL which matches the reverse proxy. 

Add global config options for both the callback listener address and redirect URL.

#### How to use GitHub

coanghel:
For Docker environments, need to bind to 0.0.0.0 instead of localhost
For rclone served by reverse proxy, should be able to override redirrect

Fixes #7634 

#### Checklist


ncw:
Can you drop the formatting changes from docs.md please? It's very difficult see what you changed!

Also why does this have a separate config for redirect URL? Will that be different to the bind address?

 Thank you for taking a look at this.

coanghel:
Done, sorry about that I didn't even notice Prettier decided to hijack the markdown formatting. I added .vscode to the gitignore, I hope that's okay.

The bind address can (and necessarily **will** in Docker) be different in Docker deployments. For example, the bind address to listen to incoming OAuth callbacks in a Docker container will be 0.0.0.0:PORT however the redirect address that must be sent to the OAuth provider (e.g. Microsoft or Google or whatever) needs to be the **public** address that eventually is routed to that bind address.

Unfortunately, I noticed that various different remotes use different redirect URLs from the main oauthutil.go. Rather than update every single remote type to standardize what is used from oauthutil.go, it made more sense to allow the user to do a global override since the redirect URL will be managed by a reverse proxy anyway.

coanghel:
Hi @ncw , just following up if everything looks right now

ncw:
Hi @coanghel sorry for the delay in merging this.

I'd like to merge for 1.70 - can you as a first step rebase it and fix the conflicts? I suggest you do that locally then force push the branch with a clean history.

coanghel:
@ncw this may have been done correctly... but I'm not sure.

PR #7647, opened 2024-02-25
adjusted cmd/bisync/log.go indent function to be more expressive abou…

KDreynolds:
#### Checklist

### DIFF
--- a/cmd/bisync/log.go
+++ b/cmd/bisync/log.go
@@ -35,9 +35,12 @@ func (b *bisyncRun) indent(tag, file, msg string) {
 
 	if tag == "Path1" {
 		tag = Color(terminal.CyanFg, "Path1")
+	} else {
+		tag = Color(terminal.BlueFg, tag)
 	}
-
 	msg = Color(terminal.MagentaFg, msg)
+	msg = strings.Replace(msg, "Queue copy to", Color(terminal.GreenFg, "Queue copy to"), -1)
+	msg = strings.Replace(msg, "Queue delete", Color(terminal.RedFg, "Queue delete"), -1)
 	file = Color(terminal.CyanFg, escapePath(file, false))
 	logf(nil, "- %-18s%-43s - %s", tag, msg, file)
 }
--- a/cmd/bisync/log.go
+++ b/cmd/bisync/log.go
@@ -29,14 +29,15 @@ func (b *bisyncRun) indent(tag, file, msg string) {
 		logf = fs.Logf
 	}
 
+	if b.opt.DryRun {
+		logf = fs.Logf
+	}
+
 	if tag == "Path1" {
 		tag = Color(terminal.CyanFg, "Path1")
-	} else {
-		tag = Color(terminal.BlueFg, tag)
 	}
+
 	msg = Color(terminal.MagentaFg, msg)
-	msg = strings.Replace(msg, "Queue copy to", Color(terminal.GreenFg, "Queue copy to"), -1)
-	msg = strings.Replace(msg, "Queue delete", Color(terminal.RedFg, "Queue delete"), -1)
 	file = Color(terminal.CyanFg, escapePath(file, false))
 	logf(nil, "- %-18s%-43s - %s", tag, msg, file)
 }
--- a/cmd/bisync/bisync_test.go
+++ b/cmd/bisync/bisync_test.go
@@ -94,7 +94,7 @@ var logReplacements = []string{
 
 // Some dry-run messages differ depending on the particular remote.
 var dryrunReplacements = []string{
-	`^(NOTICE: file5.txt: Skipped) (copy|update modification time) (as --dry-run is set [(]size \d+[)])$`,
+	`^(INFO: file5.txt: Skipped) (copy|update modification time) (as --dry-run is set [(]size \d+[)])$`,
 	`$1 copy (or update modification time) $3`,
 }
 
@@ -103,7 +103,7 @@ var dryrunReplacements = []string{
 var logHoppers = []string{
 	// Test case `dry-run` produced log mismatches due to non-deterministic
 	// order of captured dry-run info messages.
-	`NOTICE: \S+?: Skipped (?:copy|move|delete|copy \(or [^)]+\)|update modification time) as --dry-run is set \(size \d+\)`,
+	`NOTICE: \S+?: Skipped (?:copy|move|delete|copy \(or [^)]+\)|update modification time) to \S+? as --dry-run is set \(size \d+\)`,
 
 	// Test case `extended-filenames` detected difference in order of files
 	// with extended unicode names between Windows and Unix or GDrive,
--- a/cmd/bisync/bisync_test.go
+++ b/cmd/bisync/bisync_test.go
@@ -94,7 +94,7 @@ var logReplacements = []string{
 
 // Some dry-run messages differ depending on the particular remote.
 var dryrunReplacements = []string{
-	`^(INFO: file5.txt: Skipped) (copy|update modification time) (as --dry-run is set [(]size \d+[)])$`,
+	`^(NOTICE: file5.txt: Skipped) (copy|update modification time) (as --dry-run is set [(]size \d+[)])$`,
 	`$1 copy (or update modification time) $3`,
 }
 
@@ -103,7 +103,7 @@ var dryrunReplacements = []string{
 var logHoppers = []string{
 	// Test case `dry-run` produced log mismatches due to non-deterministic
 	// order of captured dry-run info messages.
-	`NOTICE: \S+?: Skipped (?:copy|move|delete|copy \(or [^)]+\)|update modification time) to \S+? as --dry-run is set \(size \d+\)`,
+	`NOTICE: \S+?: Skipped (?:copy|move|delete|copy \(or [^)]+\)|update modification time) as --dry-run is set \(size \d+\)`,
 
 	// Test case `extended-filenames` detected difference in order of files
 	// with extended unicode names between Windows and Unix or GDrive,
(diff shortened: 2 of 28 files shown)


cmd/bisync/log.go
```diff
@@ -29,16 +29,29 @@ func (b *bisyncRun) indent(tag, file, msg string) {
 		logf = fs.Logf
 	}
 
+	if b.opt.DryRun {
+		logf = fs.Logf
+		msg += " in dry-run mode"
```
nielash, line 34: The `logf` part is great! I'm not sure we need the `msg` part. I think it's already clear we're in dry run mode from other lines, like `file.txt: Skipped copy as --dry-run is set`

It would also read a little awkwardly in some places, like:

```
NOTICE: - Path2    Do queued copies to in dry-run mode - Path1
```

cmd/bisync/log.go
```diff
@@ -29,16 +29,29 @@ func (b *bisyncRun) indent(tag, file, msg string) {
 		logf = fs.Logf
 	}
 
+	if b.opt.DryRun {
+		logf = fs.Logf
```
nielash, line 33: Very nice 🙂

cmd/bisync/log.go
```diff
@@ -29,16 +29,29 @@ func (b *bisyncRun) indent(tag, file, msg string) {
...
+		msg += " in dry-run mode"
+	}
+
 	if tag == "Path1" {
 		tag = Color(terminal.CyanFg, "Path1")
+	} else if tag == "Path2" {
```
nielash, line 39: The idea here was to use different colors for Path1 and Path2, to make it easier to visually differentiate in the terminal. So I would suggest either removing this change or using a different color.

cmd/bisync/log.go
```diff
@@ -29,16 +29,29 @@ func (b *bisyncRun) indent(tag, file, msg string) {
...
+		tag = Color(terminal.CyanFg, "Path2")
 	} else {
 		tag = Color(terminal.BlueFg, tag)
 	}
+
 	msg = Color(terminal.MagentaFg, msg)
-	msg = strings.Replace(msg, "Queue copy to", Color(terminal.GreenFg, "Queue copy to"), -1)
-	msg = strings.Replace(msg, "Queue delete", Color(terminal.RedFg, "Queue delete"), -1)
+	msg = strings.Replace(msg, "Queue copy to", Color(terminal.GreenFg, "Queue copy to Path2"), -1)
```
nielash, line 46: This part needs another look, as it could be either direction here (1 to 2 or 2 to 1). So, as it stands now it would result in confusing lines like:

```
NOTICE: - Path2    Queue copy to Path2 Path1 in dry-run mode - file.txt
```

Here's an example of how this function is called:
https://github.com/rclone/rclone/blob/6e28edeb9aa1c18e27faf641729b2cbc259fe13a/cmd/bisync/deltas.go#L479

cmd/bisync/log.go
```diff
@@ -29,16 +29,29 @@ func (b *bisyncRun) indent(tag, file, msg string) {
...
+	msg = strings.Replace(msg, "Queue delete", Color(terminal.RedFg, "Queue delete from Path1"), -1)
 	file = Color(terminal.CyanFg, escapePath(file, false))
-	logf(nil, "- %-18s%-43s - %s", tag, msg, file)
+
+	if b.opt.DryRun && (strings.Contains(msg, "Queue copy to Path2") || strings.Contains(msg, "Queue delete from Path1")) {
```
nielash, line 50: I am confused about this... aren't both conditions doing the same thing? What was the intent of this change?

cmd/bisync/operations.go
```diff
@@ -224,7 +224,7 @@ func (b *bisyncRun) runLocked(octx context.Context) (err error) {
 		return err
 	}
 
-	fs.Infof(nil, "Synching Path1 %s with Path2 %s", quotePath(path1), quotePath(path2))
+	fs.Logf(nil, "Synching Path1 %s with Path2 %s", quotePath(path1), quotePath(path2))
```
nielash, line 227: Note that this change would affect all runs, not just dry-runs. I don't necessarily mind it (I usually use `-v` anyway), but I wonder if some users would find it unnecessary?

nielash (changes requested):
Thank you for this! I think it is very close 🙂 There are a few things that need another look -- I added some comments inline.

Also I note that some of the tests are failing, which is pretty common for anything that changes logging in Bisync -- it's an odd quirk of how the test suite is designed 😄 What you'll want to do is "goldenize" the tests -- the process is documented [here](https://rclone.org/bisync/#updating-golden-results). But I would save this as a last step, as some of them might end up resolving themselves as a result of the revisions.

Thanks again and let me know if you have questions!

cmd/bisync/log.go
KDreynolds: Sorry, this one is a mistake. My apologies.

cmd/bisync/log.go
KDreynolds: ill change to green, if that does not work, I can easily just remove.

cmd/bisync/log.go
KDreynolds: I'll remove the msg

cmd/bisync/operations.go
KDreynolds: I'll change this back so as to not make out of scope changes.

KDreynolds:
Ok I have made the suggested changes and attempted to address @nielash concerns. Please let me know if there is a better way or if this will work. Appreciate the chance to contribute!

cmd/bisync/log.go
```diff
@@ -29,14 +29,23 @@ func (b *bisyncRun) indent(tag, file, msg string) {
 		logf = fs.Logf
 	}
 
+	if b.opt.DryRun {
+		logf = fs.Logf
+	}
+
 	if tag == "Path1" {
 		tag = Color(terminal.CyanFg, "Path1")
+		msg = strings.Replace(msg, "Queue copy to", Color(terminal.GreenFg, "Queue copy to Path2"), -1)
```
nielash, line 38: What this would do now is:

```
NOTICE: - Path1    Queue copy to Path2 Path2 - file.txt
```

So the destination path is getting duplicated.

I'm not sure that any change is necessary here, since the info is already in the `msg`.

cmd/bisync/log.go
```diff
@@ -29,14 +29,23 @@ func (b *bisyncRun) indent(tag, file, msg string) {
...
+		msg = strings.Replace(msg, "Queue delete", Color(terminal.RedFg, "Queue delete from Path1"), -1)
+	} else if tag == "Path2" {
+		tag = Color(terminal.GreenFg, "Path2")
```
nielash, line 41: We currently try to use green/yellow/red for their universal meanings of new/modified/deleted or success/warning/error, so I would choose a different color (or leave the default of `BlueFg`).

It should also avoid `CyanFg` (already used for Path1) and `MagentaFg` (already used as the default `msg` color). So that is why `BlueFg` was one of the only options left 🙂

nielash (commented):
Thanks for the revision! Very close now 🙂 

In my opinion, the change from 32-35 is sufficient to fix the issue. So my advice would be: keep that change, remove the others, fix the tests, squash, and then we are good to go!

Also, you may want to [rebase on master](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#basing-your-changes-on-the-latest-master) before fixing the tests, as ncw recently made some changes to the tests in https://github.com/rclone/rclone/commit/f5f86786b27290964d3b6b29790f1d245f66aa1d

KDreynolds:
ok I may have just totally confused myself with the indent function at this point. my apologies, I hope this addresses the issues how you were describing @nielash , if not, I'm ready to keep plugging away. Thank you for your patience on this.

nielash (commented):
`log.go` is looking good to me now 👍

So I think all that's left now is a little bit of cleanup:

- "goldenize" to get the bisync tests to pass. This command will often get you most of the way there:

```
go test ./cmd/bisync -remote local -golden
```

You'll know when you got it right because all 10 [checks](https://github.com/rclone/rclone/pull/7647/checks) on this PR should turn green.

- let's try to cleanup the commit history a bit -- ideally it should be squashed into one commit containing only your changes (it looks like some unrelated changes got pulled in here too)

Thanks again for your help!

KDreynolds:
ok so i had attempted to rebase with master and then squash into one commit. Then i 'goldenized' using "go test ./cmd/bisync -remote local -golden". I still had one test in bisync failing when i pushed. This is probably the issue that is causing the failing checks. I will investigate further. 

As for the unrelated changes i may have accidentally squashed into this commit, any guidance on fixing this mess i have made would be sincerely appreciated.

nielash:
Nice job with the rebase! It looks to me like it worked to get rid of the unrelated changes 🙂 

I took a look at the test failures locally. For me, goldenizing fixed all but the `dry_run` test. On that one, I think the problem is that the regex change on this line:

https://github.com/rclone/rclone/blob/3ac06fbb708bd17a7f6ba2b4cc3a9969d9d4b504/cmd/bisync/bisync_test.go#L106

is no longer matching the log line it's supposed to find. I think that probably just reverting it should fix it, as I think we ended up reverting the change that required it in the first place.

I'm also now questioning if this should be reverted too:

https://github.com/rclone/rclone/blob/3ac06fbb708bd17a7f6ba2b4cc3a9969d9d4b504/cmd/bisync/bisync_test.go#L95-L97

as if anything the change would be from `INFO` to `NOTICE`, not the other way around... unfortunately it's a hard thing to test because the issue it fixes only happens on certain backends (not `local`), and whoever wrote that comment neglected to say which ones 😄 But I'm going to bet it's supposed to be `NOTICE`.

If there are any tests you can't get to pass, just get them as close as you can and then let me know and I can see if I can figure it out.

Also if you could squash your changes into one commit when you're done that would be ideal (although I think we can also take care of that during the merge, if you prefer!)

Thanks again!

nielash (approved):
I cleaned up the branch to fix the tests and they are passing now 🙂

This is now good to go, from my perspective! (Just pending final approval from @ncw)

I think the commits can be squashed during the merge process.

Thanks again for this contribution, @KDreynolds! I'm sure users will appreciate this fix 😃

KDreynolds:
Thank you for all your help @nielash . I really appreciate it!

ncw (approved):
LGTM :-)

Please squash and merge!

Issue #7625, opened 2024-02-10
New subcommand that allows exports from git-annex?


dmcardle:
Hi folks, is there any interest in a new rclone subcommand that acts as a [git-annex](https://git-annex.branchable.com/) remote? The term "remote" is unfortunately overloaded by git-annex and rclone; I'm describing an rclone subcommand that speaks the git-annex [external special remote protocol](https://git-annex.branchable.com/design/external_special_remote_protocol/), enabling git-annex to store and retrieve content from rclone remotes.

A similar project called [git-annex-remote-rclone](https://github.com/git-annex-remote-rclone/git-annex-remote-rclone) already exists, but my primary complaint is that it does not support [git-annex-export](https://git-annex.branchable.com/git-annex-export/), aka "tree exports". As a result, the content exported from git-annex to an rclone remote is opaque to humans -- just a bunch of files with names like "GPGHMACSHA512--${DIGEST}", not a browseable file tree.

I think there's also some potential performance improvements if we move the client into rclone. For instance, it could support git-annex's ASYNC protocol extension rather than transferring files one at a time. I assume there's also some overhead to exec-ing rclone once per file that could be eliminated with a more persistent client.

This is admittedly pretty niche, but I'd be pleasantly surprised if there's any community interest!

---

#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/new-subcommand-for-git-annex-remote/44546

#### What is your current rclone version (output from `rclone version`)?

N/a.

#### What problem are you are trying to solve?

Export content from git-annex.

#### How do you think rclone should be changed to solve that?

With a new subcommand.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
This certainly sounds interesting.

I don't know how the git annex protocol works but if it involves calling rclone many times then it would be most efficient to run an rclone server.

Is this something you are interested in writing?

dmcardle:
Glad to see there's some interest!

> Is this something you are interested in writing?

Yep, I'm already working on it!

@ncw Would you accept this as a new subcommand, or would you prefer it to be a separate binary?

[...]

PR #7654, opened 2024-02-29
cmd/gitannex: Add the gitannex command

dmcardle:
This PR adds a new subcommand named "gitannex", aka "git-annex-remote-rclone-builtin" when invoked via a symlink.
  
This accomplishes milestone 1 from issue #7625: "minimal support for the external special remote protocol".

-----

To add a shim program that enables git-annex to interact with rclone remotes.

* #7625 
* https://forum.rclone.org/t/new-subcommand-for-git-annex-remote/44546

#### Checklist

### DIFF
--- a/cmd/gitannex/gitannex.go
+++ b/cmd/gitannex/gitannex.go
@@ -1,0 +1,515 @@
+// Package gitannex provides the "gitannex" command, which enables [git-annex]
+// to communicate with rclone by implementing the [external special remote
+// protocol]. The protocol is line delimited and spoken over stdin and stdout.
+//
+// # Milestones
+//
+// (Tracked in [issue #7625].)
+//
+//  1. ✅ Minimal support for the [external special remote protocol]. Tested on
+//     "local" and "drive" backends.
+//  2. Add support for the ASYNC protocol extension. This may improve performance.
+//  3. Support the [simple export interface]. This will enable `git-annex
+//     export` functionality.
+//  4. Once the draft is finalized, support import/export interface.
+//
+// [git-annex]: https://git-annex.branchable.com/
+// [external special remote protocol]: https://git-annex.branchable.com/design/external_special_remote_protocol/
+// [simple export interface]: https://git-annex.branchable.com/design/external_special_remote_protocol/export_and_import_appendix/
+// [issue #7625]: https://github.com/rclone/rclone/issues/7625
+package gitannex
+
+import (
+	"bufio"
+	"context"
+	_ "embed"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/cache"
+	"github.com/rclone/rclone/fs/operations"
+	"github.com/spf13/cobra"
+)
+
+const subcommandName string = "gitannex"
+const uniqueCommandName string = "git-annex-remote-rclone-builtin"
+
+//go:embed gitannex.md
+var gitannexHelp string
+
+func init() {
+	os.Args = maybeTransformArgs(os.Args)
+	cmd.Root.AddCommand(command)
+}
+
+// maybeTransformArgs returns a modified version of `args` with the "gitannex"
+// subcommand inserted when `args` indicates that the program was executed as
+// "git-annex-remote-rclone-builtin". One way this can happen is when rclone is
+// invoked via symlink. Otherwise, returns `args`.
+func maybeTransformArgs(args []string) []string {
+	if len(args) == 0 || filepath.Base(args[0]) != uniqueCommandName {
+		return args
+	}
+	newArgs := make([]string, 0, len(args)+1)
+	newArgs = append(newArgs, args[0])
+	newArgs = append(newArgs, subcommandName)
+	newArgs = append(newArgs, args[1:]...)
+	return newArgs
+}
+
+// messageParser helps parse messages we receive from git-annex into a sequence
+// of parameters. Messages are not quite trivial to parse because they are
+// separated by spaces, but the final parameter may itself contain spaces.
+//
+// This abstraction is necessary because simply splitting on space doesn't cut
+// it. Also, we cannot know how many parameters to parse until we've parsed the
+// first parameter.
+type messageParser struct {
+	line string
+}
+
+// nextSpaceDelimitedParameter consumes the next space-delimited parameter.
+func (m *messageParser) nextSpaceDelimitedParameter() (string, error) {
+	m.line = strings.TrimRight(m.line, "\r\n")
+	if len(m.line) == 0 {
+		return "", errors.New("nothing remains to parse")
+	}
+
+	before, after, found := strings.Cut(m.line, " ")
+	if found {
+		if len(before) == 0 {
+			return "", fmt.Errorf("found an empty space-delimited parameter in line: %q", m.line)
+		}
+		m.line = after
+		return before, nil
+	}
+
+	remaining := m.line
+	m.line = ""
+	return remaining, nil
+}
+
+// finalParameter consumes the final parameter, which may contain spaces.
+func (m *messageParser) finalParameter() (string, error) {
+	m.line = strings.TrimRight(m.line, "\r\n")
+	if len(m.line) == 0 {
+		return "", errors.New("nothing remains to parse")
+	}
+
+	param := m.line
+	m.line = ""
+	return param, nil
+}
+
+// configDefinition describes a configuration value required by this command. We
+// use "GETCONFIG" messages to query git-annex for these values at runtime.
+type configDefinition struct {
+	name        string
+	description string
+	destination *string
+}
+
+// server contains this command's current state.
+type server struct {
+	reader *bufio.Reader
+	writer io.Writer
+
+	// When true, the server prints a transcript of messages sent and received
+	// to stderr.
+	verbose bool
+
+	extensionInfo                bool
+	extensionAsync               bool
+	extensionGetGitRemoteName    bool
+	extensionUnavailableResponse bool
+
+	configsDone            bool
+	configPrefix           string
+	configRcloneRemoteName string
+}
+
+func (s *server) sendMsg(msg string) {
+	msg = msg + "\n"
+	if _, err := io.WriteString(s.writer, msg); err != nil {
+		panic(err)
+	}
+	if s.verbose {
+		_, err := os.Stderr.WriteString(fmt.Sprintf("server sent %q\n", msg))
+		if err != nil {
+			panic(fmt.Errorf("failed to write verbose message to stderr: %w", err))
+		}
+	}
+}
+
+func (s *server) getMsg() (*messageParser, error) {
+	msg, err := s.reader.ReadString('\n')
+	if err != nil {
+		if len(msg) == 0 {
+			// Git-annex closes stdin when it is done with us, so failing to
+			// read a new line is not an error.
+			return nil, nil
+		}
+		return nil, fmt.Errorf("expected message to end with newline: %q", msg)
+	}
+	if s.verbose {
+		_, err := os.Stderr.WriteString(fmt.Sprintf("server received %q\n", msg))
+		if err != nil {
+			return nil, fmt.Errorf("failed to write verbose message to stderr: %w", err)
+		}
+	}
+	return &messageParser{msg}, nil
+}
+
+func (s *server) run() error {
+	// The remote sends the first message.
+	s.sendMsg("VERSION 2")
+
+	for {
+		message, err := s.getMsg()
+		if err != nil {
+			return fmt.Errorf("error receiving message: %w", err)
+		}
+
+		if message == nil {
+			break
+		}
+
+		command, err := message.nextSpaceDelimitedParameter()
+		if err != nil {
+			return fmt.Errorf("failed to parse command")
+		}
+
+		switch command {
+		//
+		// Git-annex requires that these requests are supported.
+		//
+		case "INITREMOTE":
+			err = s.handleInitRemote()
+		case "PREPARE":
+			err = s.handlePrepare()
+		case "EXPORTSUPPORTED":
+			// Indicate that we do not support exports.
+			s.sendMsg("EXPORTSUPPORTED-FAILURE")
+		case "TRANSFER":
+			err = s.handleTransfer(message)
+		case "CHECKPRESENT":
+			err = s.handleCheckPresent(message)
+		case "REMOVE":
+			err = s.handleRemove(message)
+		case "ERROR":
+			errorMessage, parseErr := message.finalParameter()
+			if parseErr != nil {
+				err = fmt.Errorf("error while parsing ERROR message from git-annex: %w", parseErr)
+				break
+			}
+			err = fmt.Errorf("received error message from git-annex: %s", errorMessage)
+
+		//
+		// These requests are optional.
+		//
+		case "EXTENSIONS":
+			// Git-annex just told us which protocol extensions it supports.
+			// Respond with the list of extensions that we want to use (none).
+			err = s.handleExtensions(message)
+		case "LISTCONFIGS":
+			s.handleListConfigs()
+		case "GETCOST":
+			// Git-annex wants to know the "cost" of using this remote. It
+			// probably depends on the backend we will be using, but let's just
+			// consider this an "expensive remote" per git-annex's
+			// Config/Cost.hs.
+			s.sendMsg("COST 200")
+		case "GETAVAILABILITY":
+			// Indicate that this is a cloud service.
+			s.sendMsg("AVAILABILITY GLOBAL")
+		case "CLAIMURL", "CHECKURL", "WHEREIS", "GETINFO":
+			s.sendMsg("UNSUPPORTED-REQUEST")
+		default:
+			err = fmt.Errorf("received unexpected message from git-annex: %s", message.line)
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Idempotently handle an incoming INITREMOTE message. This should perform
+// one-time setup operations, but we may receive the command again, e.g. when
+// this git-annex remote is initialized in a different repository.
+func (s *server) handleInitRemote() error {
+	if err := s.queryConfigs(); err != nil {
+		return fmt.Errorf("failed to get configs: %w", err)
+	}
+
+	remoteRootFs, err := cache.Get(context.TODO(), fmt.Sprintf("%s:", s.configRcloneRemoteName))
+	if err != nil {
+		s.sendMsg("INITREMOTE-FAILURE failed to open root directory of rclone remote")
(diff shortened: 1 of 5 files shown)


dmcardle:
@ncw This is ready for review!

First of all, would you be willing to accept this new command in contrib/ or would you prefer I publish as a separate project? I like it living in the rclone repo because it uses rclone as a library — and I get to abuse your CI :).

At first, I wanted to add it as a new subcommand, e.g `rclone gitannex`, but it turns out that git-annex is incapable of invoking a "special remote program" with arguments. Rather, it just executes `git-annex-remote-${REMOTE_TYPE}` by design. I considered making a wrapper shell script with the expected name, but that would not be cross-platform. Ultimately, I settled on this design, where we just have a separate binary living in contrib/.

I wouldn't ask you to include this command in releases because it would probably double the size. Also, it's only useful to a very small, technically inclined crowd who can probably figure out how to install it themselves.

Thanks!

contrib/gitannex/git-annex-remote-rclone-goyle/main.go
```diff
@@ -0,0 +1,93 @@
...
+   ***test the remote***. This command is very new and has not been tested on
+   many rclone backends. Caveat emptor!
+
+    $ git annex testremote my-rclone-remote
+
+Happy annexing!
+`,
+	Version: "0.1",
+	Annotations: map[string]string{
+		"versionIntroduced": "TODO",
```
dmcardle, line 79: I suppose I don't need this. It was originally included to align with rclone subcommands.

contrib/gitannex/git-annex-remote-rclone-goyle/main.go
```diff
@@ -0,0 +1,93 @@
...
+        rcloneremotename=SomeRcloneRemote       \
+        rcloneprefix=git-annex-content
+
+2. Before you trust the garrgoyle command with your precious data, be sure to
+   ***test the remote***. This command is very new and has not been tested on
+   many rclone backends. Caveat emptor!
+
+    $ git annex testremote my-rclone-remote
+
+Happy annexing!
```
dmcardle, line 75: I should mention that I followed my own instructions and I'm already using the command to back up my own content :)

.gitignore
```diff
@@ -2,6 +2,8 @@
 _junk/
 rclone
 rclone.exe
+/git-annex-remote-rclone-goyle
+/git-annex-remote-rclone-goyle.exe
```
dmcardle, line 6: The line above begins with slash to disambiguate from the contrib/ subdirectory with the same name.

This one doesn't need to begin with a slash, I suppose.

contrib/gitannex/git-annex-remote-rclone-goyle/main.go
```diff
@@ -0,0 +1,93 @@
...
+`,
+	Version: "0.1",
```
ncw, line 77: That is the rclone version the subcommand was introduced.

contrib/gitannex/git-annex-remote-rclone-goyle/main.go
```diff
@@ -0,0 +1,93 @@
+// # What even is this?
```
ncw, line 1: Do you think this should be an rclone command so it is always built in to rclone? So something like `rclone git annexe` or something (don't really know what should be there!).

If not then this could live in its own repo quite easily? I'd be happy to host that repo under the `rclone` org.

contrib/gitannex/gitannex.go
```diff
@@ -0,0 +1,447 @@
...
+	if err != nil {
+		s.sendMsg(fmt.Sprintf("TRANSFER-FAILURE %s %s failed to get local fs", argMode, argKey))
+		return fmt.Errorf("failed to get local fs: %w", err)
+	}
+
+	remoteFileName := argKey
+
+	switch argMode {
+	case "STORE":
+		err = operations.CopyFile(context.TODO(), remoteFs, localFs, remoteFileName, filepath.Base(argFile))
```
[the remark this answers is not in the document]
ncw: > This should probably be a two step process: (1) transfer to a temp file with a random component to avoid collisions, (2) rename the temp file to the original key.

Rclone will do this automatically when transferring to a backend where partially uploaded files are visible (eg the local backend or FTP).

> I don't know what happens on every rclone backend if two rename operations race

Depends on the backend but likely last one wins.

ncw (commented):
I've given the code a light skim :-) It looks very well written.

I think we should give some thought to whether we should make this an rclone command or not.

If we make it an rclone command then it will be built in and easier to deploy for the users (without having to download go). It will of course bloat the binary with stuff, but that is a bridge we crossed a long time ago!

If it isn't an rclone command then it should probably live in its own repository (say `rclone/gitannex`). If you like that idea I'd be happy to make that repository and give you r/w access to it.

Let me know what you think?

contrib/gitannex/git-annex-remote-rclone-goyle/main.go
dmcardle: I thought this should be a subcommand originally, but it turns out that git-annex wants to exec `git-annex-remote-${type}` with no arguments, so there goes that idea. It seems a little silly to add a subcommand to rclone that requires a separate wrapper script/command to be useful.

> If not then this could live in its own repo quite easily? I'd be happy to host that repo under the rclone org.

Thanks for the offer to create a new repo! Unfortunately, my employer requires me to jump through more hoops if I'm creating a new project than if I'm patching an existing project. (I think even more hoops if I'm not releasing it under their GitHub org.) I'm going to research what the options are a little more and see how hard it might be to release in the rclone org.

So ultimately, I tried to thread the needle by putting this in contrib/,  satisfying both technical and bureaucratic constraints. Another advantage of this approach is that the project inherit's rclone's CI configuration and other infrastructure. I suppose the  disadvantage is that I have to bug you folks when I want to merge a PR, but it's nice to have another set of eyes :)

ncw: > I thought this should be a subcommand originally, but it turns out that git-annex wants to exec `git-annex-remote-${type}` with no arguments, so there goes that idea. It seems a little silly to add a subcommand to rclone that requires a separate wrapper script/command to be useful.

Sorry, I forgot to mention the other part of the plan which would be to create a hardlink from `git-annex-remote-xxxx` to rclone then get rclone to start the correct subcommand. We already do this for [mount helpers](https://rclone.org/commands/rclone_mount/#rclone-as-unix-mount-helper) so this is easy to do.

> > If not then this could live in its own repo quite easily? I'd be happy to host that repo under the rclone org.
> 
> Thanks for the offer to create a new repo! Unfortunately, my employer requires me to jump through more hoops if I'm creating a new project than if I'm patching an existing project. (I think even more hoops if I'm not releasing it under their GitHub org.) I'm going to research what the options are a little more and see how hard it might be to release in the rclone org.

I could create the project under the rclone org then invite you to collaborate - perhaps that would satisfy your employers?

> So ultimately, I tried to thread the needle by putting this in contrib/, satisfying both technical and bureaucratic constraints. Another advantage of this approach is that the project inherit's rclone's CI configuration and other infrastructure. 

The CI is a lot of work to set up, I agree! If it is part of rclone it could also participate in the integration tests.

> I suppose the disadvantage is that I have to bug you folks when I want to merge a PR, but it's nice to have another set of eyes :)

:-)

dmcardle: > Sorry, I forgot to mention the other part of the plan which would be to create a hardlink from `git-annex-remote-xxxx` to rclone then get rclone to start the correct subcommand. We already do this for [mount helpers](https://rclone.org/commands/rclone_mount/#rclone-as-unix-mount-helper) so this is easy to do.

Oh, cool! I didn't realize this was an option!

I think making this a subcommand is the best option, then. This also means I don't need to jump through approval hoops.

Thanks for the pointer to the mount helpers. IIUC, it looks like fs/mount_helpers.go transforms args in its `init()` *before*  rclone's `Main()` function runs. I can probably pull the same trick in the new subcommand's `init()`.

ncw: > I think making this a subcommand is the best option, then. This also means I don't need to jump through approval hoops.

Great

> Thanks for the pointer to the mount helpers. IIUC, it looks like fs/mount_helpers.go transforms args in its init() before rclone's Main() function runs. I can probably pull the same trick in the new subcommand's init().

I can't remember exactly how that works, but I'm sure you'll figure it out!

.gitignore
dmcardle: Leaving this as is, it doesn't seem super important.

dmcardle:
Hey @ncw, this is ready for another look when you get a chance.

I factored it back into a subcommand (`rclone gitannex`). It now recognizes when rclone is invoked via a symlink (or whatever) named `git-annex-remote-rclone-goyle`.

Thanks!

dmcardle:
Just made a few little improvements, like more consistent use of the `messageParser` type and more test coverage.

joeyh:
WRT "rclone gitannex" vs "git-annex-remote-rclone-goyle" I would be open to considering doing something in git-annex to support the former without needing the latter. If that would be helpful enough to bother with. It might be as simple as a hardcoded special case.

dmcardle:
> WRT "rclone gitannex" vs "git-annex-remote-rclone-goyle" I would be open to considering doing something in git-annex to support the former without needing the latter. If that would be helpful enough to bother with. It might be as simple as a hardcoded special case.

That would be really useful, @joeyh! Although I would probably keep the symlink detection logic in this PR in order to support current releases of git-annex. I'd keep it around until a git-annex release with the new feature trickled down to Debian stable.

My taste would be to implement a more generic mechanism rather than adding a special case for `rclone gitannex`. What if `externaltype` could be repeated, so that `git annex initremote MyRemote type=external externaltype=rclone externaltype=gitannex ...` would cause git-annex to exec `rclone` with the additional `gitannex` arg?

cmd/gitannex/run-git-annex-testremote.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+#
+# End-to-end tests for garrgoyle. This script runs the `git-annex testremote`
+# suite against garrgoyle in an ephemeral git-annex repo.
+#
+# Assumptions:
+#
+#   * This system has an rclone remote configured named "garrgoyle-test-remote".
+#
+#   * If it uses rclone's "local" backend, /tmp/garrgoyle-test-remote exists.
```
yarikoptic, line 10: best to avoid any assumptions or otherwise would require per location where you would intend to run this test to set it up somehow etc. What about the approach we have in git-annex-remote-rclone to use temporary HOME with all needed setup? checkout 

https://github.com/git-annex-remote-rclone/git-annex-remote-rclone/blob/master/tests/all-in-one.sh#L22

which might also give you ideas on additional (it is never enough!!) testing etc.

cmd/gitannex/gitannex.go
```diff
@@ -0,0 +1,515 @@
...
+
+	"github.com/rclone/rclone/cmd"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/cache"
+	"github.com/rclone/rclone/fs/operations"
+	"github.com/spf13/cobra"
+)
+
+const commandName string = "gitannex"
+const garrgoyleName string = "git-annex-remote-rclone-goyle"
```
yarikoptic, line 43: `goyle` is cute but may be more boring `native` or `builtin` thus to signal it's intimate relationship to rclone itself?

ncw: I have to say I don't understand what "goyle" or "garrgoyle" means at all! Do we need to expose it to users?

cmd/gitannex/gitannex.md
```diff
@@ -0,0 +1,52 @@
...
+Installation on Linux
+---------------------
+
+1. Create a symlink and ensure it's on your PATH. For example:
+
+        ln -s "$(realpath rclone)" "$HOME/bin/git-annex-remote-rclone-goyle"
+
+2. Add a new external remote to your git-annex repo.
+
+   The new remote's type should be "rclone-goyle". If it is on your PATH,
```
ncw, line 15: Could the type just be `rclone` rather than `rclone-goyle`? As someone who knows very little about the protocols involved, the word `goyle` means nothing to me - is it important?

ncw (commented):
I think this is looking very good and ready to merge.

I'd like to make sure we get the naming right - that is the only thing that is difficult to change afterwards - your opinions on that would be great.

cmd/gitannex/gitannex.go
dmcardle: So I was trying to be clever with the name, but I've gotten pretty consistent feedback that the name is confusing!

We need to expose *some* name to users since they'll need to install a symlink somewhere (unless/until @joeyh adds a special-case to git-annex!).

I'll rebrand it as git-annex-remote-rclone-goyle and get rid of the "garrgoyle" name. Thanks, folks!

cmd/gitannex/gitannex.md
dmcardle: So if the name were just "rclone", then we would have a name collision with <https://github.com/git-annex-remote-rclone/git-annex-remote-rclone>. I'm working on changing the type to "rclone-builtin" right now, which seems reasonably self explanatory.

Back when I was thinking this would be a standalone command, rather than baked into rclone, I picked a globally-unique suffix that also made a semi-pun (git-annex-remote-rclone-goyle ==> "garrgoyle").

cmd/gitannex/run-git-annex-testremote.sh
dmcardle: @yarikoptic Very clever! In the interest of unblocking the PR, I'll punt on improving the test script for now, but this is the very next thing on my TODO list for the project. I really appreciate you taking a look at the PR.

yarikoptic: Traveling ATM, can't promise speedy review over weekend, but would be happy to help in general.

dmcardle:
> I think this is looking very good and ready to merge.
> 
> I'd like to make sure we get the naming right - that is the only thing that is difficult to change afterwards - your opinions on that would be great.

Hey @ncw, just got rid of the confusing name. Now it's `git-annex-remote-rclone-builtin` when invoked via symlink.  I think this is ready to go!

Also, thank you to @joeyh and @yarikoptic for taking a look!

ncw (approved):
Thank you for this great feature :-)

I'll merge it now.

joeyh:
@dmcardle good point that changes in git-annex would take quite a while to trickle down to eg debian stable. Of course, so will the new rclone.

I've opened a todo on git-annex https://git-annex.branchable.com/todo/external_special_remotes_not_using_git-annex-remote_in_name/

yarikoptic:
FWIW a shameless plug which might come handy:

- we provide non-official git-annex builds daily for the most of the OSes from https://github.com/datalad/git-annex/ 
- https://github.com/datalad/datalad-installer also has `rclone` among its components with a good range of "methods" of installation: https://github.com/datalad/datalad-installer/?tab=readme-ov-file#rclone  .

Overall -- might be easy to try the tandem as soon as we get `rclone` with it out.

dmcardle:
Neat! Thank you for the context, @yarikoptic!

cmd/gitannex/gitannex.md
```diff
@@ -0,0 +1,38 @@
+Rclone's gitannex subcommand enables git-annex to store and retrieve content
+from an rclone remote. It expects to be run by git-annex, not directly by users.
+It is an "external special remote program" as defined by git-annex.
+
+Installation on Linux
+---------------------
+
+1. Create a symlink and ensure it's on your PATH. For example:
+
+        ln -s "$(realpath rclone)" "$HOME/bin/git-annex-remote-rclone-builtin"
```
yarikoptic, line 10: I need to learn this dark magic (if it works ;-) yet to get a chance to try, but I have concerns ).

But ideally IMHO `rclone` itself should just install a shim itself in the PATH. The trickery that it would be different for *nix (shell script is enough) and Windows (needs .bat or .exe?)

dmcardle: Please do give it a try! I'm happy to make changes if there's something broken or risky.

The dark magic happens in gitannex.go's `maybeTransformArgs()`, which is called by `init()`. Basically, if `argv[0]` looks like "git-annex-remote-rclone-builtin", we insert the `gitannex` subcommand in the second position. There's actually already precedent in rclone for rewriting arguments like this: <https://rclone.org/commands/rclone_mount/#rclone-as-unix-mount-helper>.

It's decidedly a hack, but it means we don't have to write any platform-specific wrapper scripts. I think we can do the rough equivalent of `ln -s` on Windows with `mklink`, but I don't have a Windows machine handy to try it out. 

I see a few paths forward that enable us to remove the symlink step from the setup instructions:

* If git-annex adds a special case for this remote, i.e. @joeyh resolves this [todo](https://git-annex.branchable.com/todo/external_special_remotes_not_using_git-annex-remote_in_name/), users could just set the `type` or `externaltype` to whatever git-annex expects. No need for a symlink.
* We could write wrapper scripts for every supported platform. Honestly, this is probably not that hard, but I'm generally wary of subtleties introduced by shell scripts.
* We could write wrapper scripts for every supported platform and add a sub-subcommand (something like `rclone gitannex installwrapper`) that installs the platform-specific wrapper somewhere. That location could be alongside the `rclone` binary, but that makes me nervous. It could be a user-selected bin directory, but then the user still has to manage their PATH.
* We could write wrapper scripts for every supported platform and ship them in rclone releases OR we could even ship relative symlinks in the releases, I suppose. This solves the package manager issue, but I'm not sure how project maintainers would feel about adding files to the release.

yarikoptic: sorry -- my bad -- I was just too tired I guess. Instead of `realpath rclone` I saw `rclone gitannex` ;-)

Thanks for the analysis!  yeah - we need @ncw's input on this.  I guess such a wrapper could be done directly in `go` and just come with `rclone` package "natively" to not mess with "platform specific solutions" (no symlinks on windows really AFAIK).

ncw: > I see a few paths forward that enable us to remove the symlink step from the setup instructions:

You could get rclone to create the symlink on the first writable element in PATH it finds...

I'd avoid scripting if possible - Go is marvelously consistent between platforms whereas shell/powershell isn't!

If we can build the logic into rclone then that makes it a one binary deploy which is what we aim for.

joeyh:
@dmcardle I've implemented the rclone special remote in git-annex, so no symlink is needed and config looks like:

git-remote foo type=rclone encryption=none rcloneremotename=foo rcloneprefix=bar

Could you update the rclone documentation to reflect this?

dmcardle:
> Could you update the rclone documentation to reflect this?

Will do!

@joeyh, I have to admit I don't really understand the command you posted — is it supposed to be `git annex initremote` rather than `git-remote`?

FYI, I'm working on making this compatible with the various layout modes defined by git-annex-remote-rclone in #7754. Can you advise on how to switch an existing remote from `externaltype=rclone` to `type=rclone`? (Or is that a bad idea?) I'm hoping that users can transition to rclone's built-in remote without needing to copy their data.

joeyh:
Sorry, that was in fact supposed to start with git-annex initremote.

PR #7657, opened 2024-03-02
sync: don't set dir modtimes if already set

nielash:

Before this change, directory modtimes (and metadata) were always synced from src to dst, even if already in sync (i.e. their modtimes already matched.) This potentially required excessive API calls, made logs noisy, and was potentially problematic for backends that create "versions" or otherwise log activity updates when modtime/metadata is updated.

After this change, a new `DirsEqual` function is added to check whether dirs are equal based on a number of factors such as ModifyWindow and sync flags in use. If the dirs are equal, the modtime/metadata update is skipped.

For backends that require `setDirModTimeAfter`, the "after" sync is performed only for dirs that could have been changed by the sync (i.e. dirs containing files that were created/updated.)

Note that dir metadata (other than modtime) is not currently considered by DirsEqual, consistent with [how object metadata is synced](https://rclone.org/docs/#metadata) (only when objects are unequal for reasons other than metadata).

To sync dir modtimes and metadata unconditionally (the previous behavior), use `--ignore-times`.

- https://github.com/rclone/rclone/issues/6685#issuecomment-1971552577

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -141,6 +141,40 @@ func Equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object) bool {
 	return equal(ctx, src, dst, defaultEqualOpt(ctx))
 }
 
+// DirsEqual is like Equal but for dirs instead of objects.
+// It returns true if two dirs should be considered "equal" for the purposes of syncCopyMove
+// (in other words, true == "skip updating modtime/metadata for this dir".)
+// Unlike Equal, it does not consider size or checksum, as these do not apply to directories.
+func DirsEqual(ctx context.Context, src, dst fs.Directory, opt DirsEqualOpt) (equal bool) {
+	if dst == nil {
+		return false
+	}
+	ci := fs.GetConfig(ctx)
+	if ci.SizeOnly || ci.Immutable || ci.IgnoreExisting || opt.ModifyWindow == fs.ModTimeNotSupported {
+		return true
+	}
+	if ci.IgnoreTimes {
+		return false
+	}
+	if !(opt.SetDirModtime || opt.SetDirMetadata) {
+		return true
+	}
+	srcModTime, dstModTime := src.ModTime(ctx), dst.ModTime(ctx)
+	if srcModTime.IsZero() || dstModTime.IsZero() {
+		return false
+	}
+	dt := dstModTime.Sub(srcModTime)
+	if dt < opt.ModifyWindow && dt > -opt.ModifyWindow {
+		fs.Debugf(dst, "Directory modification time the same (differ by %s, within tolerance %s)", dt, opt.ModifyWindow)
+		return true
+	}
+	if ci.UpdateOlder && dt >= opt.ModifyWindow {
+		fs.Debugf(dst, "Destination directory is newer than source, skipping")
+		return true
+	}
+	return false
+}
+
 // sizeDiffers compare the size of src and dst taking into account the
 // various ways of ignoring sizes
 func sizeDiffers(ctx context.Context, src, dst fs.ObjectInfo) bool {
@@ -170,6 +204,13 @@ func defaultEqualOpt(ctx context.Context) equalOpt {
 		updateModTime:     !ci.NoUpdateModTime,
 		forceModTimeMatch: false,
 	}
+}
+
+// DirsEqualOpt represents options for DirsEqual function()
+type DirsEqualOpt struct {
+	ModifyWindow   time.Duration // Max time diff to be considered the same
+	SetDirModtime  bool          // whether to consider dir modtime
+	SetDirMetadata bool          // whether to consider dir metadata
 }
 
 var modTimeUploadOnce sync.Once
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -89,9 +89,10 @@ type syncCopyMove struct {
 	setDirMetadata         bool                   // if set we set the directory metadata
 	setDirModTime          bool                   // if set we set the directory modtimes
 	setDirModTimeAfter     bool                   // if set we set the directory modtimes at the end of the sync
-	setDirModTimeMu        sync.Mutex             // protect setDirModTimeMu
+	setDirModTimeMu        sync.Mutex             // protect setDirModTimes and modifiedDirs
 	setDirModTimes         []setDirModTime        // directories that need their modtime set
 	setDirModTimesMaxLevel int                    // max level of the directories to set
+	modifiedDirs           map[string]struct{}    // dirs with changed contents (if s.setDirModTimeAfter)
 }
 
 // For keeping track of delayed modtime sets
@@ -155,6 +156,7 @@ func newSyncCopyMove(ctx context.Context, fdst, fsrc fs.Fs, deleteMode fs.Delete
 		setDirMetadata:         ci.Metadata && fsrc.Features().ReadDirMetadata && fdst.Features().WriteDirMetadata,
 		setDirModTime:          fdst.Features().WriteDirSetModTime || fdst.Features().MkdirMetadata != nil || fdst.Features().DirSetModTime != nil,
 		setDirModTimeAfter:     fdst.Features().DirModTimeUpdatesOnWrite,
+		modifiedDirs:           make(map[string]struct{}),
 	}
 
 	s.logger, s.usingLogger = operations.GetLogger(ctx)
@@ -399,6 +401,11 @@ func (s *syncCopyMove) pairChecker(in *pipe, out *pipe, fraction int, wg *sync.W
 					fs.Errorf(pair.Dst, "Source and destination exist but do not match: %v", err)
 					s.processError(err)
 				} else {
+					if pair.Dst != nil {
+						s.markDirModifiedObject(pair.Dst)
+					} else {
+						s.markDirModifiedObject(src)
+					}
 					// If destination already exists, then we must move it into --backup-dir if required
 					if pair.Dst != nil && s.backupDir != nil {
 						err := operations.MoveBackupDir(s.ctx, s.backupDir, pair.Dst)
@@ -764,7 +771,6 @@ func (s *syncCopyMove) renameID(obj fs.Object, renamesStrategy trackRenamesStrat
 	if renamesStrategy.hash() {
 		var err error
 		hash, err := obj.Hash(s.ctx, s.commonHash)
-
 		if err != nil {
 			fs.Debugf(obj, "Hash failed: %v", err)
 			return ""
@@ -1079,6 +1085,26 @@ func (s *syncCopyMove) DstOnly(dst fs.DirEntry) (recurse bool) {
 	return false
 }
 
+// keeps track of dirs with changed contents, to avoid setting modtimes on dirs that haven't changed
+func (s *syncCopyMove) markDirModified(dir string) {
+	if !s.setDirModTimeAfter {
+		return
+	}
+	s.setDirModTimeMu.Lock()
+	defer s.setDirModTimeMu.Unlock()
+	s.modifiedDirs[dir] = struct{}{}
+}
+
+// like markDirModified, but accepts an Object instead of a string.
+// the marked dir will be this object's parent.
+func (s *syncCopyMove) markDirModifiedObject(o fs.Object) {
+	dir := path.Dir(o.Remote())
+	if dir == "." {
+		dir = ""
+	}
+	s.markDirModified(dir)
+}
+
 // copyDirMetadata copies the src directory modTime or Metadata to dst
 // or f if nil. If dst is nil then it uses dir as the name of the new
 // directory.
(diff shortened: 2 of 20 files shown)


fs/operations/operations.go
```diff
@@ -141,6 +141,40 @@ func Equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object) bool {
...
+// DirsEqual is like Equal but for dirs instead of objects.
+// It returns true if two dirs should be considered "equal" for the purposes of syncCopyMove
+// (in other words, true == "skip updating modtime/metadata for this dir".)
+// Unlike Equal, it does not consider size or checksum, as these do not apply to directories.
+func DirsEqual(ctx context.Context, src, dst fs.Directory, opt DirsEqualOpt) (equal bool) {
+	if dst == nil {
+		return false
+	}
+	ci := fs.GetConfig(ctx)
+	if ci.SizeOnly || ci.Immutable || ci.UseServerModTime || opt.ModifyWindow == fs.ModTimeNotSupported {
```
ncw, line 153: I'm not sure `ci.UseServerModTime` is correct here? The destination could well do nothing with that flag (eg sync s3 -> local)?

fs/operations/operations.go
```diff
@@ -141,6 +141,40 @@ func Equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object) bool {
...
+		return true
+	}
+	if ci.IgnoreTimes {
```
ncw, line 156: Should probably update the help for ignore-times!

  -I, --ignore-times   Don't skip files that match size and time - transfer all files

maybe swap `files` for `items` or `files/dirs`?

fs/operations/operations.go
```diff
@@ -141,6 +141,40 @@ func Equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object) bool {
...
+	if !(opt.SetDirModtime || opt.SetDirMetadata) {
+		return true
+	}
+	srcModTime, dstModTime := src.ModTime(ctx), dst.ModTime(ctx)
+	if srcModTime.IsZero() || dstModTime.IsZero() {
+		return false
+	}
+	dt := dstModTime.Sub(srcModTime)
+	if dt < opt.ModifyWindow && dt > -opt.ModifyWindow {
+		fs.Debugf(dst, "dir modification time the same (differ by %s, within tolerance %s)", dt, opt.ModifyWindow)
```
ncw, line 168: Capital `Directory modification...` would be perfect!

fs/operations/operations.go
```diff
@@ -141,6 +141,40 @@ func Equal(ctx context.Context, src fs.ObjectInfo, dst fs.Object) bool {
...
+		return true
+	}
+	if ci.UpdateOlder && dt >= opt.ModifyWindow {
+		fs.Debugf(dst, "Destination is newer than source, skipping")
```
ncw, line 172: We should probably mention `Destination directory is newer...`

fs/sync/sync.go
```diff
@@ -399,6 +402,11 @@ func (s *syncCopyMove) pairChecker(in *pipe, out *pipe, fraction int, wg *sync.W
 					fs.Errorf(pair.Dst, "Source and destination exist but do not match: %v", err)
 					s.processError(err)
 				} else {
+					if pair.Dst != nil {
+						s.markDirModified(path.Dir(filepath.ToSlash(pair.Dst.Remote())))
```
ncw, line 406: This should not need `filepath.ToSlash` here - all paths should be unix `/` delimited in rclone internals.

If they aren't then that is a bug somewhere else :-)

Please remove them all from this file! Also see later suggestion for a function `markDirModifiedObject`...

fs/sync/sync.go
```diff
@@ -92,6 +93,7 @@ type syncCopyMove struct {
 	setDirModTimeMu        sync.Mutex             // protect setDirModTimeMu
 	setDirModTimes         []setDirModTime        // directories that need their modtime set
 	setDirModTimesMaxLevel int                    // max level of the directories to set
+	modifiedDirs           map[string]struct{}    // dirs with changed contents (if s.setDirModTimeAfter)
```
ncw, line 96: ~~This needs a mutex to protect it `modifiedDirsMu` - it will be accessed concurrently~~

Edit: I see you are using the `setDirModTimeMu` which is fine - either annotate the help with that info, or rename the variable `setDirModTimeDirs` or something like that to group it with `setDirModTimeMu`

fs/sync/sync.go
```diff
@@ -1079,6 +1086,16 @@ func (s *syncCopyMove) DstOnly(dst fs.DirEntry) (recurse bool) {
 	return false
 }
 
+// keeps track of dirs with changed contents, to avoid setting modtimes on dirs that haven't changed
```
ncw, line 1089: I would make an alternate version of this `markDirModifiedObject` which just takes an `fs.Object` and does the path manipulation and calls `markDirModified`.

Note that `path.Dir` doesn't understand the convention rclone uses (`""` for root directory) so some care is needed using it.

You'll probably need something like

```go
dir := path.Dir(o.Remote())
if dir == "." {
    dir = ""
}
```

fs/sync/sync.go
```diff
@@ -1152,6 +1177,9 @@ func (s *syncCopyMove) setDelayedDirModTimes(ctx context.Context) error {
 			}
 			item := item
 			g.Go(func() error {
+				if _, ok := s.modifiedDirs[filepath.ToSlash(item.dir)]; !ok {
```
ncw, line 1180: It would be slightly more efficient to do this before `g.Go` so we don't start go routines we don't need to.

ncw (commented):
This is looking really nice - thank you :-)

I put some specific nits inline.

General comments:
- Please kill `path.ToSlash` with fire :-) These should not appear in core sync routines - they are supposed to translated to `/` at the edges and be `/` within rclone. If `\` are getting in to the core then that needs fixing, but not here!
- The bisync tests are quite annoying aren't they!
- I wonder if we could combine the `s.modifiedDirs` and `s.setDirModTimes` somehow... We are storing a lot of directory objects we don't need to potentially. Perhaps we don't know until the end of the sync.

nielash:
> * Please kill `path.ToSlash` with fire :-) These should not appear in core sync routines - they are supposed to translated to `/` at the edges and be `/` within rclone. If `\` are getting in to the core then that needs fixing, but not here!

Haha, will do! There definitely is some issue, I think specific to Windows -- I've run into it multiple times on this and other projects. But I agree this is not the proper fix for it. I'll take these out and try to see if I can find the true cause.

> * The bisync tests are quite annoying aren't they!

They sure are! I would estimate that my time on this PR was maybe 10% writing code and 90% debugging the bisync tests. But on the other hand, this is a good example of an issue worth knowing about that a more conventional unit test would probably miss. 

> * I wonder if we could combine the `s.modifiedDirs` and `s.setDirModTimes` somehow... We are storing a lot of directory objects we don't need to potentially. Perhaps we don't know until the end of the sync.

Yeah, that is basically the problem. At the point that `s.setDirModTimes` is set, we haven't yet recursed into the directory and so we don't know if there are changed files within it. It might be possible to do it after each directory is done (instead of waiting til the end of the sync), or at least release the ones we know we definitely won't need, but I'm not sure how much difference that would make, and would probably add a lot of complexity trying to coordinate that across channels. IMO it doesn't seem worth it.

fs/sync/sync.go
nielash: Got it. Thoughts on making this a new helper function in `fspath`? I frequently find myself reinventing the wheel on path manipulation stuff -- sometimes incorrectly! `fspath` usually has something pretty close but not exactly what I need.

fs/sync/sync.go
nielash: I was wondering about that, and whether that's outweighed by being able to do it concurrently.

fs/operations/operations.go
nielash: To be honest I think `--ignore-times` has a larger naming problem, as it still has an effect with `--checksum` and `--size-only`, and on backends that don't support modtimes to begin with. "Transfer unconditionally" is probably a better description of what it actually does.

fs/operations/operations.go
nielash: I think you're right, this one doesn't belong here. I think I was looking for something that does the oppose of `--ignore-times`, but that doesn't really exist.

I'm also thinking `ci.IgnoreExisting` probably ought to be added...right?

fs/sync/sync.go
nielash: I removed all the `filepath` usages here :) I think the error I was seeing the other day may just have been a transient fluke of GitHub Actions (I've been seeing more of those lately)

Here is an example of what I mean though about the `path` library causing issues on Windows with rclone's unix paths: [this test](https://github.com/nielash/rclone/commit/c575274b07c391861b154347a9d5382f9fe84204) fails [only on Windows](https://github.com/nielash/rclone/actions/runs/8153440329). (And I confirmed it also fails on a real PC, not just GH Actions)

I think the issue in this particular case is `path.Join` removing the double `//` and then rclone adding it back incorrectly. But there may be other issues as well.

I think it would be really handy to have a more comprehensive library of rclone-safe path manipulation functions!

fs/operations/operations.go
nielash: Made this change.

fs/operations/operations.go
nielash: Made this change.

fs/operations/operations.go
nielash: Made this change.

fs/operations/operations.go
nielash: Made this change.

fs/sync/sync.go
nielash: Made this change.

fs/sync/sync.go
nielash: I took a stab at this -- see what you think.

fs/sync/sync.go
nielash: Made this change.

fs/sync/sync.go
ncw: I think this is great, but `fspath.Dir` is not correct.

Perhaps put in what I suggested then in a separate PR we can find all the `path.Dir` and `path.Base` and work out whether we should be fixing them up with a new function?

ncw (commented):
I think this is looking great except for `fspath.Dir` which isn't correct!

See suggestions inline - thank you :-)

fs/fspath/path.go
```diff
@@ -353,3 +352,37 @@ func JoinRootPath(remote, filePath string) string {
...
+// "remote:some/path/" -> "remote:some/path"
+//
+// "some/path" -> "some"
+//
+// "/path" -> ""
+//
+// "path" -> ""
+//
+// "remote:path" -> "remote:"
+func Dir(remote string) string {
```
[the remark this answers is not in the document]
nielash: Don't the other `fspath` functions use `remote` the same way?

https://github.com/rclone/rclone/blob/dcdbad35543b5396f4f1cbb39afc0b4abfd8d0ec/fs/fspath/path.go#L272-L281

That said, I see what you mean about the `:` and agree that's a problem. I had hoped to make one universal function that could work for both a full `ConfigString` and a relative `.Remote()`, but I see now why that's not possible... if given `12:00am` there's really no way to tell if that's one relative directory or a dir named `00am` at remote `12`. 😕

I think you are right that this is better addressed in a separate PR, so I will take this out.

fs/fspath/path.go
```diff
@@ -353,3 +352,37 @@ func JoinRootPath(remote, filePath string) string {
...
+	dir := path.Dir(remote)
+	if dir == "/" { // ex. remote == "/somelocalpath"
+		return ""
```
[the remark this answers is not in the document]
nielash: Hmm ok, I didn't realize that. So just making sure I understand... `Dir("somedir")` should return `""`, and `Dir("/somedir")` should return `/`?

fs/sync/sync.go
nielash: Ok, made that change.

fs/fspath/path.go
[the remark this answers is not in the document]
ncw: > Don't the other `fspath` functions use `remote` the same way?

Ha, yes that is true! I think your `Dir` command is approximately correct for `fspath` but a more accurate implementation would use `fspath.Split`.

> That said, I see what you mean about the `:` and agree that's a problem. I had hoped to make one universal function that could work for both a full `ConfigString` and a relative `.Remote()`, but I see now why that's not possible... if given `12:00am` there's really no way to tell if that's one relative directory or a dir named `00am` at remote `12`. 😕
> 
> I think you are right that this is better addressed in a separate PR, so I will take this out.

:+1:

fs/fspath/path.go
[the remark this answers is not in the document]
ncw: Yes that is correct...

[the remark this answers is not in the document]
nielash: Ok. The new version here should do that (I copied and pasted your suggestion: https://github.com/nielash/rclone/blob/8aa0ee9d60c72f7d9dedff0b4a6d6b2e9f795ab3/fs/sync/sync.go#L1101-L1104 )

PR #7660, opened 2024-03-04
`combine`: fix `CopyDirMetadata` error on upstream root

nielash:

`operations.CopyDirMetadata` was failing with:

```
internal error: expecting directory string from combine root '' to have SetMetadata method: optional feature not implemented
```

if the `dst` was the root directory of a `combine` upstream. This is because combine was returning a `*fs.Dir`, which does not satisfy the `fs.SetMetadataer` interface.

Here's one proposed solution, but feel free to toss it if you have a different one in mind 🙂 

Also, I didn't check, but it seems possible that Union and other similar backends might need a similar fix.

No, I just noticed it tonight.

#### Checklist

### DIFF
--- a/fs/dir_wrapper.go
+++ b/fs/dir_wrapper.go
@@ -7,8 +7,9 @@ import (
 
 // DirWrapper wraps a Directory object so the Remote can be overridden
 type DirWrapper struct {
-	Directory        // Directory we are wrapping
-	remote    string // name of the directory
+	Directory           // Directory we are wrapping
+	remote       string // name of the directory
+	failSilently bool   // if set, ErrorNotImplemented should not be considered an error for this directory
 }
 
 // NewDirWrapper creates a wrapper for a directory object
@@ -20,6 +21,16 @@ func NewDirWrapper(remote string, d Directory) *DirWrapper {
 		Directory: d,
 		remote:    remote,
 	}
+}
+
+// NewLimitedDirWrapper creates a DirWrapper that should fail silently instead of erroring for ErrorNotImplemented.
+//
+// Intended for exceptional dirs lacking abilities that the Fs otherwise usually supports
+// (ex. a Combine root which can't set metadata/modtime, regardless of support by wrapped backend)
+func NewLimitedDirWrapper(remote string, d Directory) *DirWrapper {
+	dw := NewDirWrapper(remote, d)
+	dw.failSilently = true
+	return dw
 }
 
 // String returns the name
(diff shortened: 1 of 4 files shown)


MANUAL.md
```diff
@@ -11982,8 +11982,8 @@ Metadata is data about a file which isn't the contents of the file.
 Normally rclone only preserves the modification time and the content
 (MIME) type where possible.
 
-Rclone supports preserving all the available metadata on files (not
-directories) when using the `--metadata` or `-M` flag.
+Rclone supports preserving all the available metadata on files and
```
ncw, line 11985: MANUAL.md is auto generated so you'll need to edit the original which is `docs/content/docs.md`

fs/dir_wrapper.go
```diff
@@ -55,6 +66,10 @@ func (d *DirWrapper) Metadata(ctx context.Context) (Metadata, error) {
 func (d *DirWrapper) SetMetadata(ctx context.Context, metadata Metadata) error {
 	do, ok := d.Directory.(SetMetadataer)
 	if !ok {
+		if d.failSilently {
+			Debugf(d, "can't SetMetadata for this directory (%T from %v) -- skipping", d.Directory, d.Fs())
```
ncw, line 70: Log message should start with a capital!

fs/dir_wrapper.go
```diff
@@ -66,6 +81,10 @@ func (d *DirWrapper) SetMetadata(ctx context.Context, metadata Metadata) error {
 func (d *DirWrapper) SetModTime(ctx context.Context, t time.Time) error {
 	do, ok := d.Directory.(SetModTimer)
 	if !ok {
+		if d.failSilently {
+			Debugf(d, "can't SetModTime for this directory (%T from %v) -- skipping", d.Directory, d.Fs())
```
ncw, line 85: Log message should start with a capital!

ncw (commented):
This looks great - thanks for fixing up!

I put some very minor nits inline.

fs/dir_wrapper.go
nielash: Funny -- the linter has admonished me in the past for using capitals (and punctuation) so I think I had it in my head that this was a general style rule, but maybe that's only for errors, since they sometimes get wrapped?

As an aside, is it possible to add custom linter rules for stuff like this? Could be convenient to get that little reminder to fix it myself, and spare you the trouble 🙂

MANUAL.md
nielash: And I see you beat me to it :) So I've dropped this one.

fs/dir_wrapper.go
ncw: Looking through the code, of course there are zillions of logs which don't start with a capital! Maybe it is more nuanced than I said and more a matter of taste...

```
$ git grep -P 'fs.(Debug|Info|Notice|Error)f\(.*?, "[a-z]' | wc -l
691
$ git grep -P 'fs.(Debug|Info|Notice|Error)f\(.*?, "[A-Z]' | wc -l
1388
```

The [linter has lots of config](https://golangci-lint.run/usage/configuration/) but give the 2:3 ratio above I think we should leave alone!

ncw (approved):

Will merge now.

Issue #7652, opened 2024-02-28
Sync directory modification time and metadata on the root of the sync


ncw:
Currently rclone does not sync directory modification time and metadata on the root of the sync.

The syncing was implemented in #6685 and there @nielash noted that the root modification times and metadata weren't being synced.

This is quite hard to fix as there is no way of getting an `fs.Directory` object for the root from the backend at the moment.

This probably needs a `NewEntry` method to replace or extend `NewObject`, or it needs `NewFs` to return the root directory entry.

This is the same problem that `rclone lsjson --stat remote:` has - this has no information on the root directory.

PS When this is fixed, take out the notes in `cmd/{sync,copy,move}` pointing to this issue

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ppanengsuriyani-create:
Cara upload gimana

PR #7668, opened 2024-03-08
`fs/cache`: fix parent not getting pinned when remote is a file & `hasher`: look for cached hash if passed hash unexpectedly blank 

nielash:

Three related bug fixes identified on https://forum.rclone.org/t/hasher-with-gdrive-backend-does-not-return-sha1-sha256-for-old-files/44680/9?u=nielash

1. `fs/cache`: fix parent not getting pinned when remote is a file
2. `hasher`: fix error from trying to stop an already-stopped db
3. `hasher`: look for cached hash if passed hash unexpectedly blank

`2` and `3` are specific to `hasher`, but `1` might possibly explain some unexplained behavior in other backends.

Before this change, when `cache.GetFn` was called on a _file_ rather than a directory, two cache entries would be added (the file + its parent) but only one of them would get pinned if the caller then called `Pin(f)`. This left the other one exposed to expiration if the `ci.FsCacheExpireDuration` was reached. This was problematic because both entries point to the same Fs, and if one entry expires while the other is pinned, the `Shutdown` method gets erroneously called on an Fs that is still in use.

An example of the problem showed up in the `Hasher` backend, which uses the `Shutdown` method to stop the bolt db used to store hashes. If a command was run on a Hasher file (ex. `rclone md5sum --download hasher:somelargefile.zip`) and hashing the file took longer than the `--fs-cache-expire-duration` (5m by default), the bolt db was stopped before the hashing operation completed, resulting in an error.

This change fixes the issue by ensuring that the Pin/Unpin functions also Pin/Unpin any aliases in lockstep, so that pinning a file also pins the cache entry for its parent.

- https://forum.rclone.org/t/hasher-with-gdrive-backend-does-not-return-sha1-sha256-for-old-files/44680/9?u=nielash

#### Checklist

### DIFF
--- a/fs/cache/cache.go
+++ b/fs/cache/cache.go
@@ -4,6 +4,7 @@ package cache
 import (
 	"context"
 	"runtime"
+	"strings"
 	"sync"
 
 	"github.com/rclone/rclone/fs"
@@ -12,10 +13,11 @@ import (
 )
 
 var (
-	once  sync.Once // creation
-	c     *cache.Cache
-	mu    sync.Mutex            // mutex to protect remap
-	remap = map[string]string{} // map user supplied names to canonical names
+	once           sync.Once // creation
+	c              *cache.Cache
+	mu             sync.Mutex            // mutex to protect remap
+	remap          = map[string]string{} // map user supplied names to canonical names - [fsString]canonicalName
+	childParentMap = map[string]string{} // tracks a one-to-many relationship between parent dirs and their direct children files - [child]parent
 )
 
 // Create the cache just once
@@ -57,6 +59,39 @@ func addMapping(fsString, canonicalName string) {
 	mu.Unlock()
 }
 
+// addChild tracks known file (child) to directory (parent) relationships.
+// Note that the canonicalName of a child will always equal that of its parent,
+// but not everything with an equal canonicalName is a child.
+// It could be an alias or overridden version of a directory.
+func addChild(child, parent string) {
+	if child == parent {
+		return
+	}
+	mu.Lock()
+	childParentMap[child] = parent
+	mu.Unlock()
+}
+
+// returns true if name is definitely known to be a child (i.e. a file, not a dir).
+// returns false if name is a dir or if we don't know.
+func isChild(child string) bool {
+	mu.Lock()
+	_, found := childParentMap[child]
+	mu.Unlock()
+	return found
+}
+
+// ensures that we return fs.ErrorIsFile when necessary
+func getError(fsString string, err error) error {
+	if err != nil && err != fs.ErrorIsFile {
+		return err
+	}
+	if isChild(fsString) {
+		return fs.ErrorIsFile
+	}
+	return nil
+}
+
 // GetFn gets an fs.Fs named fsString either from the cache or creates
 // it afresh with the create function
 func GetFn(ctx context.Context, fsString string, create func(ctx context.Context, fsString string) (fs.Fs, error)) (f fs.Fs, err error) {
@@ -69,31 +104,39 @@ func GetFn(ctx context.Context, fsString string, create func(ctx context.Context
 		created = ok
 		return f, ok, err
 	})
+	f, ok := value.(fs.Fs)
 	if err != nil && err != fs.ErrorIsFile {
+		if ok {
+			return f, err // for possible future uses of PutErr
+		}
 		return nil, err
 	}
-	f = value.(fs.Fs)
 	// Check we stored the Fs at the canonical name
 	if created {
 		canonicalName := fs.ConfigString(f)
 		if canonicalName != canonicalFsString {
-			// Note that if err == fs.ErrorIsFile at this moment
-			// then we can't rename the remote as it will have the
-			// wrong error status, we need to add a new one.
-			if err == nil {
+			if err == nil { // it's a dir
 				fs.Debugf(nil, "fs cache: renaming cache item %q to be canonical %q", canonicalFsString, canonicalName)
 				value, found := c.Rename(canonicalFsString, canonicalName)
 				if found {
 					f = value.(fs.Fs)
 				}
 				addMapping(canonicalFsString, canonicalName)
-			} else {
-				fs.Debugf(nil, "fs cache: adding new entry for parent of %q, %q", canonicalFsString, canonicalName)
-				Put(canonicalName, f)
+			} else { // it's a file
+				// the fs we cache is always the file's parent, never the file,
+				// but we use the childParentMap to return the correct error status based on the fsString passed in.
+				fs.Debugf(nil, "fs cache: renaming child cache item %q to be canonical for parent %q", canonicalFsString, canonicalName)
+				value, found := c.Rename(canonicalFsString, canonicalName) // rename the file entry to parent
+				if found {
+					f = value.(fs.Fs) // if parent already exists, use it
+				}
+				Put(canonicalName, f)                        // force err == nil for the cache
+				addMapping(canonicalFsString, canonicalName) // note the fsString-canonicalName connection for future lookups
+				addChild(fsString, canonicalName)            // note the file-directory connection for future lookups
 			}
 		}
 	}
-	return f, err
+	return f, getError(fsString, err) // ensure fs.ErrorIsFile is returned when necessary
 }
 
 // Pin f into the cache until Unpin is called
@@ -111,7 +154,6 @@ func PinUntilFinalized(f fs.Fs, x interface{}) {
 	runtime.SetFinalizer(x, func(_ interface{}) {
 		Unpin(f)
 	})
-
 }
 
 // Unpin f from the cache
@@ -174,6 +216,9 @@ func PutErr(fsString string, f fs.Fs, err error) {
 	canonicalName := fs.ConfigString(f)
 	c.PutErr(canonicalName, f, err)
 	addMapping(fsString, canonicalName)
+	if err == fs.ErrorIsFile {
+		addChild(fsString, canonicalName)
+	}
 }
 
 // Put puts an fs.Fs named fsString into the cache
@@ -186,6 +231,7 @@ func Put(fsString string, f fs.Fs) {
 // Returns number of entries deleted
 func ClearConfig(name string) (deleted int) {
 	createOnFirstUse()
+	ClearMappingsPrefix(name)
 	return c.DeletePrefix(name + ":")
 }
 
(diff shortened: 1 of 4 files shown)


fs/cache/cache.go
```diff
@@ -99,7 +100,9 @@ func GetFn(ctx context.Context, fsString string, create func(ctx context.Context
 // Pin f into the cache until Unpin is called
 func Pin(f fs.Fs) {
 	createOnFirstUse()
-	c.Pin(fs.ConfigString(f))
+	configString := fs.ConfigString(f)
+	c.Pin(configString)
+	c.PinPrefix(strings.TrimSuffix(configString, "/")) // also pin its aliases, if any
```
ncw, line 105: I'm worried that c.PinPrefix would pin things it shouldn't here, so if you made `remote:path/dir` then made `remote/path/file.txt` then the latter would pin the former.

ncw (commented):
> Before this change, when cache.GetFn was called on a file rather than a
directory, two cache entries would be added (the file + its parent) 

I wonder if this is the actual bug - adding two entries.

The thinking behind caching two entries was that the backend instance returned when you call `NewFs("remote:dir")` and `NewFs("remote:dir/file.txt")`  is the same.

**However** this is not true - the result of the `Root()` call is different for the two backends, so they should be treated as two backends not one.

So I think a better fix for this might be to remove the caching of the parent in the pointing to a file case.

What do you think?

PS I've cherry-picked the two hasher fixes for the release. This fix can go in the release if done in time or in a point release.

nielash:
> So I think a better fix for this might be to remove the caching of the parent in the pointing to a file case.\n\nWhat do you think?

I agree! Adding two entries never totally made sense to me.

ncw:
> I agree! Adding two entries never totally made sense to me.

Do you want to have a go at that change? Then you can see if it fixes your hasher oddities.

nielash:
> Do you want to have a go at that change? Then you can see if it fixes your hasher oddities.

Sure! Looking at it now.

nielash:
I am realizing a slight problem here, which is that `Pin(f)` relies on `fs.ConfigString(f)` to get the name, and `fs.ConfigString(f)` will return the same string for both `remote:path` and `remote:path/file.txt`. So if the only cache entry is  `remote:path/file.txt`, `Pin(f)` will never find it. What do you think is the proper solution to that?

nielash:
> the result of the `Root()` call is different for the two backends

Also, it looks to me like this used to be true, but was changed very recently in https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4

As a result, if all you have is `f` it seems like there's no way of determining which file it should point to...

nielash:
Here's an alternate fix -- see what you think. https://github.com/rclone/rclone/pull/7668/commits/6c808e63fc00c6d235721e2aa6901b99cca04d90

It eliminates the duplicate entry, but the one it keeps is the parent rather than the file. This is because of what I mentioned above -- that after https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4 the `f.Root()` for `remote:path` and `remote:path/file.txt` are indistinguishable, so you can deduce a parent from a child but not the reverse.

This version should ensure that:
1. only one entry is added to the cache (the file's parent, not the file).
2. future lookups correctly find the entry regardless of whether they are called with the parent name or one of its children.
3. `fs.ErrorIsFile` is returned when (and only when) `fsString` points to a file	(preserving the fix from https://github.com/rclone/rclone/commit/8d5bc7f28b8459a0e04fdbfd2dca43c5d9ec4555).

If we wanted to do the reverse and cache the file but not its parent, I think what we'd probably have to do is either revert https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4 and have `f.Root()` return a different path, or add a new required method to `fs.Fs` which returns the file remote path or `""` if not a file. Additionally we'd have to double-check that `fs.ConfigString(f)` returns something different for files vs. dirs. Could potentially be useful for other reasons, but I'm not sure it's necessary here, as there's a less invasive workable solution.

fs/cache/cache.go
nielash: I think this would also be true of [`c.DeletePrefix`](https://github.com/rclone/rclone/blob/b7422a4fc8b747d18c01834bf5a14f1a92ca596c/fs/cache/cache.go#L189) which is where I got the idea, and there are probably ways to refine the targeting, but it's probably moot now as my proposed revision doesn't have a `PinPrefix`.

ncw:
> Here's an alternate fix -- see what you think. [6c808e6](https://github.com/rclone/rclone/commit/6c808e63fc00c6d235721e2aa6901b99cca04d90)

Wow, that is quite complicated! Probably too complicated for a point release - I'm really allergic to breaking stuff in a point release!

> It eliminates the duplicate entry, but the one it keeps is the parent rather than the file. This is because of what I mentioned above -- that after [c69eb84](https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4) the `f.Root()` for `remote:path` and `remote:path/file.txt` are indistinguishable, so you can deduce a parent from a child but not the reverse.

Note that commit is just one of many fixing the same problem in lots of backends!
 
> This version should ensure that:
> 
> 1. only one entry is added to the cache (the file's parent, not the file).
> 2. future lookups correctly find the entry regardless of whether they are called with the parent name or one of its children.
> 3. `fs.ErrorIsFile` is returned when (and only when) `fsString` points to a file	(preserving the fix from [8d5bc7f](https://github.com/rclone/rclone/commit/8d5bc7f28b8459a0e04fdbfd2dca43c5d9ec4555)).

I think those seem correct.

Can we do it more simply?

One thing I was thinking is that if we changed the contents of the cache from `fs.Fs` to something like

```go
struct {
    f fs.Fs
    files []string
}
```

Then when inserting a `ErrorIsFile` we'd just insert the parent but add an entry to `files`, and when looking up we'd look for the fsString as-is then look it up without the last path entry. If a member of `files` matches then we could return the `f` with `ErrorIsFile`.

> If we wanted to do the reverse and cache the file but not its parent, I think what we'd probably have to do is either revert [c69eb84](https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4) and have `f.Root()` return a different path, or add a new required method to `fs.Fs` which returns the file remote path or `""` if not a file. Additionally we'd have to double-check that `fs.ConfigString(f)` returns something different for files vs. dirs. Could potentially be useful for other reasons, but I'm not sure it's necessary here, as there's a less invasive workable solution.

I don't want to undo c69eb84573c85206ab028eda2987180e049ef2e4 as it fixed a small but important data loss bug.

The whole returning fs.ErrorIsFile is not very pleasant, but I haven't thought of anything better! One idea I had was that you always open a backend at the root with `NewFs` then you `Chroot` into the directory that you want, but that isn't much better.

nielash:
> I'm really allergic to breaking stuff in a point release!

I would argue that it is already broken and this patch un-breaks it 😄 But I hear you!

It seems plausible that the same issue could be breaking `copyto` and `moveto` operations that take longer than 5 minutes, for backends with a `Shutdown` method. (I haven't tested this)

> Can we do it more simply?

My first version (https://github.com/rclone/rclone/pull/7668/commits/c4ac032a0cb2e668f992998aad5ed2eacdeb904f) is still an option. It's much simpler. (Leave the duplicate but Pin/Unpin them in lockstep)

> One thing I was thinking is that if we changed the contents of the cache from `fs.Fs` to something like
> 
> ```go
> struct {
>     f fs.Fs
>     files []string
> }
> ```
> 
> Then when inserting a `ErrorIsFile` we'd just insert the parent but add an entry to `files`, and when looking up we'd look for the fsString as-is then look it up without the last path entry. If a member of `files` matches then we could return the `f` with `ErrorIsFile`.

That's very similar to what my newest version is doing. I think it would work -- but it might mean having to call `NewFs` more often, because we'd lose all the `files []string` knowledge every time the cache expires.

> I don't want to undo [c69eb84](https://github.com/rclone/rclone/commit/c69eb84573c85206ab028eda2987180e049ef2e4) as it fixed a small but important data loss bug.

I don't think we need to.



> The whole returning fs.ErrorIsFile is not very pleasant, but I haven't thought of anything better! One idea I had was that you always open a backend at the root with `NewFs` then you `Chroot` into the directory that you want, but that isn't much better.

Yeah I don't have any perfect solutions either. One idea might be to have `NewFs` and `cache.Get` return an optional remote string param like `cmd.NewFsFile`, which can be `""` if not pointing to a file. (Maybe also an `fs.Directory` to help with #7652)

My overall observation as someone still learning all of this is that paths are much easier to reason about on the CLI than in the code -- `f.Root()` does not always mean what I'd expect!

ncw:
HI @nielash I was just looking back through old pulls and came across this. There are some good fixes here - do you want go try to get them in 1.68?

We were having a discussion above about the fs cache and I haven't revised it yet, but I remember that it was all getting very complicated.

Anyway now would be a good time to merge stuff so would you like to rebase the above commits, maybe squash the revert and re-write if you are happy with it and check you are happy with the design. I'll then take a look and we can try to get it in for 1.68

Thank you

nielash:
Thanks for the reviews! I will take a look -- not sure if I'll get to all of them by the 1.68 deadline, but we'll see how far I get 😄 

I think this one is mostly done, unless you happen to like the following idea and think it's worth tackling as part of this:

> Yeah I don't have any perfect solutions either. One idea might be to have `NewFs` and `cache.Get` return an optional remote string param like `cmd.NewFsFile`, which can be `""` if not pointing to a file. (Maybe also an `fs.Directory` to help with #7652)

And of course I'll also rebase and clean up the commit history (I think the `hasher` ones are now moot as they were merged via cherry-pick)

ncw:
> Thanks for the reviews! I will take a look -- not sure if I'll get to all of them by the 1.68 deadline, but we'll see how far I get 😄

Great, thank you :-)
 
> I think this one is mostly done, unless you happen to like the following idea and think it's worth tackling as part of this:
> 
> > Yeah I don't have any perfect solutions either. One idea might be to have `NewFs` and `cache.Get` return an optional remote string param like `cmd.NewFsFile`, which can be `""` if not pointing to a file. (Maybe also an `fs.Directory` to help with #7652)

I think that changing what NewFs returns is probably a good idea, but we should make an issue and discuss first to make sure we are choosing the right change as doing it will be a lot of work!

Perhaps NewFs should return an `(fs.Fs, fs.DirEntry, error)` where DirEntry can be a file or a directory depending. having NewFs return the actual file or directory could make some operations more efficient. And that idea interacts with the fs Cache in various exciting ways too.

> And of course I'll also rebase and clean up the commit history (I think the `hasher` ones are now moot as they were merged via cherry-pick)

:+1: 

Let me know when you want me to take a look.

ncw:
Is this ready for merge @nielash ? It looks that way to me but wanted to check first :-)

nielash:
> Is this ready for merge @nielash ? It looks that way to me but wanted to check first :-)

Yes 🙂

ncw (approved):
LGTM :⁠-⁠)

Will merge now

PR #7670, opened 2024-03-12
`memory`: fix deadlock in `operations.Purge`

nielash:

Before this change, the `Memory` backend had the potential to deadlock under certain conditions, if the `ListR` callback required locking the `b.mu` mutex. This was the case with `operations.Purge`, because `Memory` has no `Purge` method, and the fallback option does:

https://github.com/rclone/rclone/blob/ca903b9872790921ae85dee316ca1800657281c4/fs/operations/operations.go#L1126

which potentially starts removing objects before the listing has completed.

This change fixes the issue by ensuring the mu is unlocked for the duration of `fn`.

No.

#### Checklist

### DIFF
--- a/backend/memory/memory.go
+++ b/backend/memory/memory.go
@@ -38,8 +38,7 @@ func init() {
 }
 
 // Options defines the configuration for this backend
-type Options struct {
-}
+type Options struct{}
 
 // Fs represents a remote memory server
 type Fs struct {
@@ -385,10 +384,22 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (err error) {
 	bucket, directory := f.split(dir)
 	list := walk.NewListRHelper(callback)
+	entries := fs.DirEntries{}
 	listR := func(bucket, directory, prefix string, addBucket bool) error {
-		return f.list(ctx, bucket, directory, prefix, addBucket, true, func(remote string, entry fs.DirEntry, isDirectory bool) error {
-			return list.Add(entry)
+		err = f.list(ctx, bucket, directory, prefix, addBucket, true, func(remote string, entry fs.DirEntry, isDirectory bool) error {
+			entries = append(entries, entry) // can't list.Add here -- could deadlock
+			return nil
 		})
+		if err != nil {
+			return err
+		}
+		for _, entry := range entries {
+			err = list.Add(entry)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
 	}
 	if bucket == "" {
 		entries, err := f.listBuckets(ctx)
--- a/backend/memory/memory_internal_test.go
+++ b/backend/memory/memory_internal_test.go
@@ -1,0 +1,40 @@
+package memory
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	_ "github.com/rclone/rclone/backend/local"
+	"github.com/rclone/rclone/fs/operations"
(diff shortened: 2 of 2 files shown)


backend/memory/memory.go
```diff
@@ -303,7 +303,9 @@ func (f *Fs) list(ctx context.Context, bucket, directory, prefix string, addBuck
 					}
 					_, found := dirs[dir]
 					if !found {
+						b.mu.RUnlock()
```
ncw, line 306: This is ranging over `b.objects` here - I don't think it is safe to release the lock as it gives the opportunity for other methods to grab it and mutate `b.objects`.

fs/operations/operations_test.go
```diff
@@ -550,6 +550,22 @@ func TestCat(t *testing.T) {
 	}
 }
 
+// test that Purge fallback does not result in deadlock from concurrently listing and removing
```
ncw, line 553: This test should probably be in the internal tests for memory backend

fs/operations/operations_test.go
```diff
@@ -550,6 +550,22 @@ func TestCat(t *testing.T) {
...
+func TestPurgeListDeadlock(t *testing.T) {
+	ctx := context.Background()
+	r := fstest.NewRunIndividual(t) // make new container (azureblob has delayed mkdir after rmdir)
+	r.Mkdir(ctx, r.Fremote)
+	r.Fremote.Features().Disable("Purge") // force fallback-purge
+
+	// make a lot of files to prevent it from finishing too quickly
+	for i := 0; i < 100; i++ {
```
ncw, line 561: Does this need to be in a go routine to tickle the bug?

ncw (commented):
This is very tricky to solve properly!

The problem is that `ListR` calls out of the backend with the lock held. We can avoid that by accumulating all the entries in memory first (which is a bit wasteful on memory) before calling the function that is passed into `ListR` with the entries in one go at the end.

ncw (commented):
This is a tricky problem to solve!

I think getting ListR to batch all the `entries` before calling the callback on them is probably the best solution. Not particularly nice as it uses more memory than it has to but OK...!

fs/operations/operations_test.go
nielash: My thought was to check whether any other backends have this same problem (now or in the future) -- although that seems unlikely as this is somewhat specific to `memory`.

Maybe there's a more efficient way to check it? The reason for creating so many files is that this is basically a race condition which would not occur with fewer files.

ncw: As you've probably noticed I try not to get the integration tests to do too much work! Some backends take several seconds to upload a file so uploading 100 files may increase the run time by 5 minutes for the slow backends.

I think this test is fine, but let's put it in the Internal tests for the memory backend as I think it is unlikely that other backends will have this problem in the future.

ncw (commented):
Code looks great now - comments about the test inline.

I don't want to remove ListR for memory as it stands in for a bucket based remote in the tests and they all implement ListR. If this was a normal backend then yes removing it would make perfect sense, but I view it more as an aid to testing.

Thank you :-)

fs/operations/operations_test.go
nielash: Ok. Full disclosure: bisync's `volatile` test does something very similar (uploading 100 files) which is how I found this issue originally. Should I put in a condition to skip that test on non-local?

ncw: Probably a good idea, yes!

ncw (approved):
Thank you - please merge :-)

PR #7680, opened 2024-03-21
docs: sftp, clarify `shell_type = none` and `ssh = ` behaviour

kapitainsky:
Discussed on the forum:

https://forum.rclone.org/t/can-rclone-be-made-to-work-with-an-sftp-server-confining-users-to-an-sftp-jail-and-no-login/44931

#### Checklist

- [ x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).

### DIFF
--- a/docs/content/sftp.md
+++ b/docs/content/sftp.md
@@ -279,6 +279,8 @@ option before first run, the auto-detection will be skipped, and if
 you set a different value later this will override any existing.
 Value `none` can be set to avoid any attempts at executing shell
 commands, e.g. if this is not allowed on the server.
+If you have `shell_type = none` in the configuration then
+the [ssh](#sftp-ssh) must not be set.
 
 When the server is [rclone serve sftp](/commands/rclone_serve_sftp/),
 the rclone SFTP remote will detect this as a Unix type shell - even


docs/content/sftp.md
```diff
@@ -279,6 +279,10 @@ option before first run, the auto-detection will be skipped, and if
 you set a different value later this will override any existing.
 Value `none` can be set to avoid any attempts at executing shell
 commands, e.g. if this is not allowed on the server.
+However, if you have “shell_type = none” in the configuration then you
```
albertony, line 282: > “shell_type = none”

This uses the pesky "curly quotes". I wouldn't use those. And maybe then even make it a code formatted instead of quoted, like other references to `shell_type` and `none` in the lines above.

docs/content/sftp.md
```diff
@@ -279,6 +279,10 @@ option before first run, the auto-detection will be skipped, and if
...
+should not have an ssh command such as
+`ssh = ssh -i path\ssh_id user@ip-address`, you should have `ssh = `
```
albertony, line 284: To any possible confusion, where one is reading this as they *must* explicitly set the `ssh` option empty, maybe just say it must not be set? Maybe even link to its documentation `[ssh](#sftp-ssh)` - which also have a relevant note on the same topic:

> Rclone ignores all the internal config if you use this option and expects you to configure the ssh binary with the user/host/port and any other options

albertony (commented):
Nice. I didn't know that `ssh` overrides the `shell_type` this way, when I wrote this section of the docs.. But, without looking at the implementation, it seems it is in fact [documented](https://rclone.org/sftp/#sftp-ssh) behavior:

> Rclone ignores all the internal config if you use this option and expects you to configure the ssh binary with the user/host/port and any other options you need.

docs/content/sftp.md
kapitainsky: yeap. will change

kapitainsky:
> Nice. I didn't know that `ssh` overrides the `shell_type` this way, when I wrote this section of the docs.. But, without looking at the implementation, it seems it is in fact [documented](https://rclone.org/sftp/#sftp-ssh) behavior:
> 
> > Rclone ignores all the internal config if you use this option and expects you to configure the ssh binary with the user/host/port and any other options you need.

Yes - now when I understand the issue this part makes sense. But proposed clarification makes it explicit and brings more clarity, which IMO is always good:).

albertony:
> But proposed clarification makes it explicit and brings more clarity, which IMO is always good:).

I agree!

> I have also amended it as per your suggestions.

Great, thanks. I think this is fine now! 🚢

albertony (approved):

PR #7683, opened 2024-03-22
corrected relative path to absolute into  google drive.md

phgachoud:
and added some links to the reference


ncw (approved):
Thank you - very useful changes :-)

PR #7688, opened 2024-03-24
`memory`: fix incorrect list entries when rooted at subdirectory

nielash:

Before this change, `List` would return incorrect directory paths (relative to the wrong root) if the Fs root pointed to a subdirectory. For example, listing dir `a/b/c/d` of remote `:memory:` would work correctly, but listing dir `c/d` of remote `:memory:a/b` would not, and would result in `Entry doesn't belong in directory %q (contains subdir)` errors.

This change fixes the issue and adds a test to detect any other backends that might have the same issue.

No.

#### Checklist

### DIFF
--- a/fstest/fstests/fstests.go
+++ b/fstest/fstests/fstests.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/cache"
 	"github.com/rclone/rclone/fs/config"
 	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/fspath"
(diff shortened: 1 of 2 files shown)


fstest/fstests/fstests.go
```diff
@@ -1205,6 +1206,46 @@ func Run(t *testing.T, opt *Opt) {
...
+				newF, err := cache.Get(ctx, subRemoteName+"/hello? sausage/êé")
+				assert.NoError(t, err)
+				nestedFile := fstest.Item{
+					ModTime: fstest.Time("2001-02-03T04:05:06.499999999Z"),
+					Path:    "a/b/c/d/e.txt",
+				}
+				_, _ = testPut(ctx, t, newF, &nestedFile)
+				depth := 10
+
+				objs, dirs, err := walk.GetAll(ctx, newF, "", true, depth)
```
ncw, line 1221: Can you use `fstest.CheckListing` to see whether everything is OK? That does retries and exponential backoff which `walk.GetAll` doesn't which is often needed on eventually consistent backends.

fstest/fstests/fstests.go
```diff
@@ -1205,6 +1206,46 @@ func Run(t *testing.T, opt *Opt) {
 				}, fs.GetModifyWindow(ctx, f))
 			})
 
+			// TestFsListRootedSubdir tests putting and listing with an Fs that is rooted at a subdirectory 2 levels down
+			TestFsListRootedSubdir := func(t *testing.T) {
```
ncw, line 1210: Can you cut this test down the the minimum possible to reproduce the bug? Remember that this test gets run on 100 different backends in the integration tests!

ncw (commented):
Good catch - well done :-)

See some inline comments about the tests.

Thank you.

fstest/fstests/fstests.go
nielash: I can. The idea was to also look for other similar errors that could conceivably pop up in the future, but maybe some of these are already covered by other tests.

fstest/fstests/fstests.go
ncw: Ok!

fstest/fstests/fstests.go
nielash: Made this change.

ncw (approved):
LGTM thank you :-)

PR #7692, opened 2024-03-25
`dropbox`: fix chunked uploads when size <= chunkSize

nielash:

Before this change, `dropbox` would read the same bytes twice (exceeding the file size and causing an error) if the total size was known and less than or equal to the `chunkSize` (i.e. small enough to complete in one chunk). This change fixes the issue.

No.

#### Checklist

### DIFF
--- a/backend/dropbox/dropbox.go
+++ b/backend/dropbox/dropbox.go
@@ -1592,6 +1592,11 @@ func (o *Object) uploadChunked(ctx context.Context, in0 io.Reader, commitInfo *f
 
 	chunkSize := int64(o.fs.opt.ChunkSize)
 	chunks, remainder := size/chunkSize, size%chunkSize
+	if size > -1 && size <= chunkSize {
+		chunkSize = size
+		chunks = 1
+		remainder = 0
+	}
 	if remainder > 0 {
 		chunks++
 	}
@@ -1605,6 +1610,9 @@ func (o *Object) uploadChunked(ctx context.Context, in0 io.Reader, commitInfo *f
 	}
 	appendArg := files.UploadSessionAppendArg{Cursor: &cursor}
 	for currentChunk := 1; ; currentChunk++ {
+		if int64(currentChunk) > chunks {
+			chunkSize = 0 // don't read the same bytes again
+		}
 		cursor.Offset = in.BytesRead()
 
 		if chunks < 0 {
@@ -1654,7 +1662,7 @@ func (o *Object) uploadChunked(ctx context.Context, in0 io.Reader, commitInfo *f
 			break
 		}
 
-		if size > 0 {
+		if size >= 0 {
 			// if size is known, check if next chunk is final
 			appendArg.Close = uint64(size)-in.BytesRead() <= uint64(chunkSize)
 			if in.BytesRead() > uint64(size) {


ncw:
How do you trigger this problem?

This was my attempt

    echo hello | rclone rcat -vv --streaming-upload-cutoff 0 dropbox:smol.txt

But that didn't work.

I tried uploading small files with batches but that didn't work either.

The fact that no-one has reported it before makes me think it isn't easy to trigger!

nielash:
> How do you trigger this problem?

We are hitting it on the bisync tests because of this bit of code, which is a workaround for Dropbox's inability to set modtime on an existing object without re-uploading:

https://github.com/rclone/rclone/blob/a67688dcc7aa973cc2a82e757f391e34e852914d/cmd/bisync/bisync_test.go#L1067-L1082

Here's an example that will trigger it with the updated tests:

```
go test ./cmd/bisync -case rclone_args -remote TestDropbox: 
```

Result:
```
Error:          test step failed
Test:           TestBisyncRemoteRemote/rclone_args
Messages:       step 10 failed: upload failed: expected 19 bytes in input, but have read 38 so far
```

Note that `size%chunkSize` would be `19%50331648`, which results in `0` chunks and remainder of `19`

nielash:
Ugh wait... I think the bug is on the test, not in Dropbox. Hold that thought... 🤦‍♂️

nielash:
> Ugh wait... I think the bug is on the test, not in Dropbox. Hold that thought... 🤦‍♂️

Ok I fixed the test on #7693. The issue was that `operations.Cat` was not filtered for a single file, so it was concatenating all the files in the directory. This caused `buf` to exceed the size passed in, which Dropbox correctly complained about.

I do think there may still be a small bug in `dropbox.uploadChunked`, as your example command results in lines like:
```
2024/03/27 04:42:20 DEBUG : smol.txt: Uploading chunk 1/0
2024/03/27 04:42:20 DEBUG : smol.txt: Uploading chunk 2/0
```

with the numerator exceeding the denominator. But I'm not sure if that is actually causing any harm? So perhaps this should just be closed.

ncw:
> > Ugh wait... I think the bug is on the test, not in Dropbox. Hold that thought... 🤦‍♂️
> 
> Ok I fixed the test on #7693. The issue was that `operations.Cat` was not filtered for a single file, so it was concatenating all the files in the directory. 

If you have the `fs.Object` then use `operations.Open` to open it for reading. If you don't have the object then use `f.NewObject` to get it first.

> This caused `buf` to exceed the size passed in, which Dropbox correctly complained about.

Whew!
 
> I do think there may still be a small bug in `dropbox.uploadChunked`, as your example command results in lines like:
> 
> ```
> 2024/03/27 04:42:20 DEBUG : smol.txt: Uploading chunk 1/0
> 2024/03/27 04:42:20 DEBUG : smol.txt: Uploading chunk 2/0
> ```
> 
> with the numerator exceeding the denominator. But I'm not sure if that is actually causing any harm? So perhaps this should just be closed.

The `/0` are indicating that the length of the file isn't known at that point so we don't know how many chunks we actually have. It isn't particularly nice to look at but it is cosmetic only.

I'll close this then (with some relief that Dropbox uploads are in fact fine!).

PR #7697, opened 2024-03-26
log: Added support to include date-time in log filename

HikaruHokkyokusei:
Added a flag `log-file-time-format` that allows specifying a date-time format that is used to replace the `:dt:` wildcard in the filename of the log file.

This change allows users to specify a Date-time Format string for a new flag `log-file-time-format`. A datetime string is generated using this format that replaces the wildcard `:dt:` in the log filename. This change is required for a specific scenario: -

By default, rclone always appends (not overwrite) logs to a file. This is the same with cases when a Windows Service is created to allow rclone to run in the background while mounting. Based on the logging level, these logs can be huge and one might not prefer them getting appended to the same file every time the mount command is used (say after the device is restarted). Making dynamic filename (in the context of Windows Service) is a tricky challenge, as the `%....%` variable templates are not replaced with the actual values.

This change allows users to now use dynamic filenames based on the date and time of each session of the rclone.

While the change has not been discussed in any issues, a couple of users have talked in the context of similar requirements in the forums.

https://forum.rclone.org/t/dynamically-create-log-file-with-date-in-the-name/27603
https://forum.rclone.org/t/rclone-win2019-scheduled-tasks-date-in-logfile/33232

#### Checklist

### DIFF
--- a/fs/log/log.go
+++ b/fs/log/log.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/rclone/rclone/fs"
 	"github.com/sirupsen/logrus"
@@ -16,11 +17,12 @@ import (
 
 // Options contains options for controlling the logging
 type Options struct {
-	File              string // Log everything to this file
-	Format            string // Comma separated list of log format options
-	UseSyslog         bool   // Use Syslog for logging
-	SyslogFacility    string // Facility for syslog, e.g. KERN,USER,...
-	LogSystemdSupport bool   // set if using systemd logging
+	File               string // Log everything to this file
+	FilenameTimeFormat string // Date time format that is used to replace the :dt: in logging filename
+	Format             string // Comma separated list of log format options
+	UseSyslog          bool   // Use Syslog for logging
+	SyslogFacility     string // Facility for syslog, e.g. KERN,USER,...
+	LogSystemdSupport  bool   // set if using systemd logging
 }
 
 // DefaultOpt is the default values used for Opt
@@ -116,6 +118,31 @@ func InitLogging() {
 
 	// Log file output
 	if Opt.File != "" {
+		if Opt.FilenameTimeFormat != "" {
+			if strings.Index(Opt.File, ":dt:") == -1 {
+				log.Fatalf("Datetime format has been provided without using :dt: wildcard in the log filename")
+			}
+
+			currTime := time.Now()
+			replacer := strings.NewReplacer(
+				"%A", currTime.Format("Monday"),
+				"%a", currTime.Format("Mon"),
+				"%Y", currTime.Format("2006"),
+				"%y", currTime.Format("06"),
+				"%B", currTime.Format("January"),
+				"%b", currTime.Format("Jan"),
+				"%m", currTime.Format("01"),
+				"%d", currTime.Format("02"),
+				"%H", currTime.Format("15"),
+				"%I", currTime.Format("03"),
+				"%M", currTime.Format("04"),
+				"%S", currTime.Format("05"),
+			)
+
+			timeStr := replacer.Replace(Opt.FilenameTimeFormat)
+			Opt.File = strings.ReplaceAll(Opt.File, ":dt:", timeStr)
+		}
+
 		f, err := os.OpenFile(Opt.File, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0640)
 		if err != nil {
 			log.Fatalf("Failed to open log file: %v", err)
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -1390,6 +1390,42 @@ The options mean
 
 During rmdirs it will not remove root directory, even if it's empty.
 
+### --log-file-time-format=FORMAT ###
+
+A format that is used to generate a datetime string. Any `:dt:` wildcard
+in the log filename specified via `--log-file=FILE` will be replaced with
+this datetime string.
+
+If `FORMAT` is `%Y-%m-%d`, and log-file `FILE` is `err_:dt:.log`, then
+all rclone output will be logged to the file `err_2024-03-26.log`
+assuming that the current date it `26 March 2024`.
+
+Following keywords in the `FORMAT` are recognised as below:
+
+`%A` - Weekday (E.g.: Monday)
+
+`%a` - Short Weekday (E.g.: Mon)
+
+`%Y` - Year (E.g.: 2006)
+
+`%y` - Short Year (E.g.: 06)
+
+`%B` - Month name (E.g.: January)
+
+`%b` - Short Month name (E.g.: Jan)
+
+`%m` - Month number (E.g.: 01)
+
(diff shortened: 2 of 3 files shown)


nielash:
> 1. Do we need to invent our own time format when we can just link to [the Go docs](https://pkg.go.dev/time#pkg-constants)?

I did something just like this recently (actually twice!) Maybe this should be factored out as a `lib` package?

https://github.com/rclone/rclone/blob/2bebbfaded9139921a919f1b384bf4f736c85d4f/fs/operations/operations.go#L2106-L2156 ([docs](https://rclone.org/commands/rclone_lsf/#:~:text=Other%20formats%20can%20be%20specified%20with%20the%20%2D%2Dtime%2Dformat%20flag.%20Examples%3A))


https://github.com/rclone/rclone/blob/2bebbfaded9139921a919f1b384bf4f736c85d4f/cmd/bisync/bilib/names.go#L107-L165 ([docs](https://rclone.org/bisync/#conflict-suffix:~:text=%2D%2Dconflict%2Dsuffix%20supports%20several%20dynamic%20date%20variables))

HikaruHokkyokusei:
@nielash Thanks for pointing out the specific code where you have implemented this.

@ncw , Here is why I believe we cannot reuse the code provided by @nielash : -
Since we will be using this time in the name of the file, where `/` and `:` might not be allowed on some platforms, our available options will be severely restricted after eliminating invalid options. One possibility is to use the same technique that rclone uses to replace `/`, `:`, etc. in the filenames with other similar Unicode characters like `／` and `∶` when it is syncing, but will that be worth it?

Also, thanks for pointing out [go-strftime](https://github.com/ncruces/go-strftime) repo. I had come across it before, but didn't know if we were allowed to use external libs. I believe this is the best approach so I will replace my implementation with `go-strftime` lib. What do you guys think?

---

I will take a look at other places that you have mentioned in point number 3 and let you know @ncw .

---
Yep, the idea of a dynamic wildcard (instead of just using `:dt:`) makes sense. I'll make the appropriate changes for that too. I will make it such that the flag format will look like this `--file-time-format WILDCARD~#~FORMAT`. The argument value will be split at `~#~` and will provide us with both, the `wildcard` and the `format`.

HikaruHokkyokusei:
So with this commit, I have taken care of the custom wildcard and replaced the formatting methodology with the `go-strftime` library.

@ncw , I looked into `--suffix` and `--backup-dir`. I don't want to make changes to them without having a deep understanding of what is being done there.

I do not know the frequency of how many times these values are referred to. The logging file was created once, so that was straightforward.

One idea I was setting the value for `ci.Suffix` and `ci.BackupDir` in the `SetFlags` function [here](https://github.com/rclone/rclone/blob/f62e7b5b30be8357e53fe0134baafe25cf552238/fs/config/configflags/configflags.go#L178) which is called at initialization time [here](https://github.com/rclone/rclone/blob/f62e7b5b30be8357e53fe0134baafe25cf552238/cmd/cmd.go#L393).

In this `SetFlags` function, I plan to call `log.Opt.FilenameTimeFormat` and `log.Opt.FilenameTimeWildcard` defined [here](https://github.com/HikaruHokkyokusei/rclone/blob/0bd81ab8d551e2bead3b71e97f38cfabe6fa6b9f/fs/log/log.go#L22).

This way, one specific value will be set for `--suffix` and `--backup-dir`. Is this enough? I see in the code that operations like `opt.OrigBackupDir = ci.BackupDir` are performed [here](https://github.com/HikaruHokkyokusei/rclone/blob/0bd81ab8d551e2bead3b71e97f38cfabe6fa6b9f/cmd/bisync/operations.go#L83). So, the question arises whether I should make set this value once when the rclone is being initialized, or to recalculate this suffix or backup-dir wherever they are used, which depends on the frequency of them being called and their use case.

nielash:
I think your solution is fine, but just to clarify one thing:

> Since we will be using this time in the name of the file, where `/` and `:` might not be allowed on some platforms, our available options will be severely restricted after eliminating invalid options.

This issue is specifically addressed in bisync's `--conflict-suffix` by inventing a new constant without colons or slashes:

https://github.com/rclone/rclone/blob/f62e7b5b30be8357e53fe0134baafe25cf552238/docs/content/bisync.md?plain=1#L628-L635

and also allowing the user to specify their own custom format:

https://github.com/rclone/rclone/blob/2bebbfaded9139921a919f1b384bf4f736c85d4f/cmd/bisync/bilib/names.go#L154

That said, I agree that `%Y-%m-%d` semantics might be more intuitive than `2006-01-02` to casual users who aren't Go developers! So perhaps [go-strftime](https://github.com/ncruces/go-strftime) is a better choice for this reason.

fs/log/log.go
```diff
@@ -114,8 +118,27 @@ func InitLogging() {
 
 	fs.LogPrintPid = strings.Contains(flagsStr, ",pid,")
 
+	if Opt.FilenameTimeFormat != "" {
+		tmp := strings.SplitN(Opt.FilenameTimeFormat, "~#~", 2)
+		if len(tmp) != 2 {
+			log.Fatalf("Invlaid Filename Time Formart provided via --file-time-format. Please refer docs for proper usage.")
```
nielash, line 124: typos -- `Invalid` and `Format`

HikaruHokkyokusei: Ahhh, for some reason JB IDE didn't point out typos in line. Can't believe I made such a stupid mistake. My bad. Fixed the typos.

Thanks for pointing out.

HikaruHokkyokusei:
@ncw , would be great if you could provide some insight about how to move forward with `--suffix` and `--backup-dir`, otherwise I believe this PR is good to be merged.

HikaruHokkyokusei:
@nielash @ncw , I think this PR can be merged unless we are waiting for something else.

ncw:
Sorry for the delay merging @HikaruHokkyokusei 

I think the code and implementation looks great but I'd like to discuss exactly what this looks like as these things are very difficult to change once released!

We have at the moment `--file-time-format WILDCARD~#~FORMAT` so the user might use something like `--file-time-format @~#~%Y-%m-d` which means for `@` substitute the time 2001-02-03 in `--log-file rclone.@.log`

I'm thinking ahead a bit here, but I'd like to re-use this syntax in `--backup-dir` and `--suffix` which would relieve a small but significant pain point in rclone. This syntax could also be used in remote strings on the command line which is something people often want to do - copy to a dated folder with `rclone copy .... src: dst:@date@/

So up for discussion
1. The name of the flag `--file-time-format` - maybe it should be more generic `--time-format` if we are thinking of re-using it with `--suffix` and `--backup-dir`, thought it does apply to files always.
2. Should it have a default? Say `--file-time-format @date@~#~%Y-%m-d`. I think probably not but I don't know.
3. I don't like the `~#~` separator. What is wrong with using `=`? That would mean users couldn't use `=` as part of the wildcard, but there are plenty of other characters which can be used and that makes it very similar to other rclone flags.
4. We could use @nielash standard forms here also if we had `--time-format @@=RFC3339` where the contents after the `=` was exactly one of those standard forms.

Any thoughts welcome (@nielash and @albertony also)

PR #7552, opened 2024-01-08
Adding a new backend implementation for uloz.to

iotmaestro:
#### What is the purpose of this change?

As noted in the title, this PR adds a new backend for uloz.to. In the past the server was focused on free-for-all sharing, and has recently pivoted into a more Dropbox-like business model. They offer affordable storage plans, especially around the 10 TB mark where other larger providers fall short.

The PR only implements the basic functionality for now. I plan to add support for more optional features but wanted to keep the initial PR as simple as possible (it's quite bloated already).

Disclaimer: I'm not affiliated with Uloz.to in any way but as a satisfied customer.

#### Was the change discussed in an issue or in the forum before?

No

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [x] I have added tests for all changes in this PR if appropriate.
- [x] I have added documentation for the changes if appropriate.
- [ ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages). - **apologies about this, let's squash merge with a better message** 
- [x] I'm done, this Pull Request is ready for review :-)

PR #7699, opened 2024-03-28
Remove the temporary Uloz.to filesize limit

iotmaestro:

Remove the workarounds for uloz.to's file size. 

The combination of large files, absence of `Content-Length` header, and chunked encoding triggers the issue. 

- files under 2.5G are uploaded fine as demonstrated. 
- setting the `Content-Length` either fixes the issue (non-negative value), or triggers a 400 if the file size is not known (-1). 
- removing chunked encoding does nothing (if `Content-Length` is set), or triggers a 415 (if `Content-Length` is not specified)

This currently means we still run into the issue with large files if rclone doesn't know their size beforehand. Such scenario is presumably very rare though, and explicitly setting the size to -1 isn't handled properly in rclone's multipart generation code. 

https://github.com/rclone/rclone/pull/7552

#### Checklist

### DIFF
--- a/backend/ulozto/ulozto.go
+++ b/backend/ulozto/ulozto.go
@@ -38,8 +38,6 @@ const (
 	maxSleep      = 2 * time.Second
 	decayConstant = 2 // bigger for slower decay, exponential
 	rootURL       = "https://apis.uloz.to"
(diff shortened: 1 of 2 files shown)


ncw (approved):
Looks good to me. Thank you.

PR #7701, opened 2024-03-28
s3: update Scaleway's configuration options - fixes #7507

### LINKED ISSUE #7507: Allowing "crypt" drives to reside in many other drives (not just 1 drive)

```
rclone v1.65.0
- os/kernel: 5.15.0-89-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.4
- go/linking: static
- go/tags: none
```

Link multiple small drives and limit them to 1 larger drive using rclone's crypt drive function

https://rclone.org/crypt/

I have an idea that rclone can create additional features of crypt drives, for example allowing crypt drives to reside in many other drives (not just 1 drive). Example: we have 5 mega.io accounts of 20GB, 3 Google Drive accounts of 15GB... they can combine them into (5 x 20GB) + (3 x 15GB) ~ 145GB of "crypt" drive space.
When uploading a file to a "virtual" crypt drive, it will divide the file into files for example 1MB, 10MB... written to "real" drives (5 mega.io accounts of 20GB, 3 accounts). Google Drive account 15GB...)
When the user needs to download, rclone will download small files (eg 1MB, 10MB...) that have been divided and encrypted, then reassemble and create the original file.
I think it's possible, it's pretty magical, it looks like we're creating a universe inside the little free storage packages of the storage providers.

#### How to use GitHub

lavigne958:

In order to handle special character, the configuration must specify rclone configuration to use `list_url_encode`.

Scaleway's S3 backend is configured to return the error 'BucketAlreadyOwnedByYou' when trying to create a new bucket you already own, in that matter set rclone to expect suche error using `use_already_exists`

Set upload size value to recommended values. 

closes #7507 

#### Checklist

### DIFF
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -3364,8 +3364,6 @@ acl = private
 upload_cutoff = 5M
 chunk_size = 5M
 copy_cutoff = 5M
-list_url_encode = true
-use_already_exists = true
 ```
 
 [C14 Cold Storage](https://www.online.net/en/storage/c14-cold-storage) is the low-cost S3 Glacier alternative from Scaleway and it works the same way as on S3 by accepting the "GLACIER" `storage_class`.
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -3359,10 +3359,13 @@ endpoint = s3.nl-ams.scw.cloud
 access_key_id = SCWXXXXXXXXXXXXXX
 secret_access_key = 1111111-2222-3333-44444-55555555555555
 region = nl-ams
-location_constraint =
+location_constraint = nl-ams
 acl = private
-server_side_encryption =
-storage_class =
+upload_cutoff = 5M
+chunk_size = 5M
+copy_cutoff = 5M
+list_url_encode = true
+use_already_exists = true
 ```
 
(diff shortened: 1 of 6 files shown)


albertony:
The MANUAL.* and rclone.1 files are autogenerated, so you should revert your changes in those files. Thanks.

lavigne958:
> The MANUAL.* and rclone.1 files are autogenerated, so you should revert your changes in those files. Thanks.

alright will do, where is the source of the content then please ? :thinking:

docs/content/s3.md
```diff
@@ -3359,10 +3359,13 @@ endpoint = s3.nl-ams.scw.cloud
...
 region = nl-ams
-location_constraint =
+location_constraint = nl-ams
 acl = private
-server_side_encryption =
-storage_class =
+upload_cutoff = 5M
+chunk_size = 5M
+copy_cutoff = 5M
+list_url_encode = true
```
ncw, line 3367: You don't need to put list_url_encode or use_already_exists here as they are set in a quirk now - I'd remove them from here

ncw (commented):
This looks great thank you.

Please see minor comment inline!

Thank you

docs/content/s3.md
lavigne958: true. done :heavy_check_mark:

ncw (approved):
This is great now - thank you :-)

I'll merge now.

ncw:
Thank you

PR #7710, opened 2024-03-30
Obscure ulozto password

ncw:

Make the password be obscured to line up with other rclone backends

@iotmaestro I decided to obscure the password for uloz.to to make it consistent with the other backends.

I'll merge this shortly, but I'l putting this PR here as a heads-up for you as it will break your configs. You'll need to `rclone obscure` your password in the config file.

This also fixes a few things in the docs.

(Oops this has some other things in, but just the uloz.to commit is relevant!)


ncw:
I merged this in 679f4fdfa93834068c3869b4f8646090f79b5656

iotmaestro:
@ncw Thanks for the fixes and for the heads up!

Issue #7665, opened 2024-03-08
bisync: add to integration tests


ncw:
[Here](https://github.com/rclone/rclone/issues/6685#issuecomment-1974747783) we decided that bisync should participate in the integration tests for all backends

This will be part of giving us the confidence to take bisync out of beta.

I note that the bisync tests are likely to be troublesome here and very likely this issue will cause us to have to re-work them.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

PR #7711, opened 2024-03-30
bisync: several fixes for integration tests

nielash:

Several fixes for the `bisync` integration tests:

- use unique `initdir` and `datadir` for each subtest so concurrent tests don't interfere with each other
- remove dots from dir names for bucket backends
- ignore messages specific to `cache` backend
- skip fix-case tests on backends that can't fix-case
- don't expect `{hashtype} differ` messages on backends with no hash types
- print timestamps in UTC local

More fixes will still be needed, but this should hopefully fix a good portion of them.

- https://github.com/rclone/rclone/issues/7665

#### Checklist


ncw (approved):
Looks fine - please merge.

The ulozto changes makes me think that we need to adjust that for for every backend. Can we get away from that in future?

nielash:
> The ulozto changes makes me think that we need to adjust that for for every backend. Can we get away from that in future?

The `bisync_vscode_debuggers_launch.json` file doesn't actually affect the tests -- it's just for convenience to make debugging in VS Code easier, as typing out all those combinations manually would be quite tedious! It is set up to be auto-generated when tests are goldenized, but there's no need to do it immediately after adding a new backend.

PR #7716, opened 2024-04-01
cmd/gitannex: Replace e2e test script with Go command

dmcardle:
This commit implements milestone 2.1 for the gitannex subcommand: https://github.com/rclone/rclone/issues/7625#issuecomment-1951403856

This rewrite makes a few improvements over the old shell script:

(1) It no longer uses the system's rclone.conf. Now, it writes the
    rclone.conf file in an ephemeral directory.

(2) It no longer makes any assumptions about /tmp.

I'm hoping that writing this in Go will enable more cross-platform support in the future, but for now we're still restricted to Unixy systems due to reliance on the HOME environment variable.

Issue #7625

Improve end-to-end testing for the gitannex subcommand.

* Milestone 2.1 on https://github.com/rclone/rclone/issues/7625#issuecomment-1951403856
* Implements a change recommended here: https://github.com/rclone/rclone/pull/7654#discussion_r1536265154

#### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
             quicktest: true
 
           - job_name: mac_amd64
-            os: macos-11
+            os: macos-latest
             go: '>=1.22.0-rc.1'
             gotags: 'cmount'
             build_flags: '-include "^darwin/amd64" -cgo'
@@ -58,7 +58,7 @@ jobs:
             deploy: true
 
           - job_name: mac_arm64
-            os: macos-11
+            os: macos-latest
             go: '>=1.22.0-rc.1'
             gotags: 'cmount'
             build_flags: '-include "^darwin/arm64" -cgo -macos-arch arm64 -cgo-cflags=-I/usr/local/include -cgo-ldflags=-L/usr/local/lib'
@@ -138,7 +138,7 @@ jobs:
           brew update
           brew install --cask macfuse
           brew install git-annex
-        if: matrix.os == 'macos-11'
+        if: matrix.os == 'macos-latest'
 
       - name: Install Libraries on Windows
         shell: powershell


dmcardle:
@yarikoptic I took a stab at improving the e2e test situation. PTAL!

FYI, I also have support for various repo layouts on my agenda (milestone 2.2 of https://github.com/rclone/rclone/issues/7625#issuecomment-1951403856). Once that happens, I want to add a loop to this e2e test for every repo layout, which will give us some confidence that "rclone gitannex" is self-consistent. I also want to add some compat tests, e.g. initialize a remote with frankencase, store foo on git-annex-remote-rclone and retrieve it from "rclone gitannex".

ncw:
How long do these tests take to run?

- We could be running them on the CI
- Or we could be running them as part of the integration tests. The integration tests run against all the different backends which might be overkill.

Either way they should be re-formatted as a Go test as that is how the CI testing and the integration testing is done. I'd suggest they just `t.Skip` if on the wrong architecture and if the requirements aren't installed.

dmcardle:
> How long do these tests take to run?
> 

A minute or two on my slow laptop, but I was planning to repeat with a bunch of parameters in the future, so potentially much longer.

> 
> - We could be running them on the CI
> 
> - Or we could be running them as part of the integration tests. The integration tests run against all the different backends which might be overkill.
> 
> 

Oh, I'll have to look into the integration tests. It's probably worth running the current unit tests against every backend, at least.

> 
> Either way they should be re-formatted as a Go test as that is how the CI testing and the integration testing is done. I'd suggest they just `t.Skip` if on the wrong architecture and if the requirements aren't installed.
> 
> 

One thing I'm not sure about is whether it's possible or advisable for a Go test to reach into the repo and build the rclone binary like this command does. If that doesn't sound concerning to you, I'll try rewriting as a proper Go test.

Or is there maybe some other trick that's not occurring to me? I don't think we want to do any Makefile trickery that would prevent `go test ./...` from doing its thing.

yarikoptic:
> It's probably worth running the current unit tests against every backend

I would owe you a drink if it happens! ;)  I now even wonder if then eventually we should add rclone testing as a downstream testing component to our daily https://github.com/datalad/git-annex builds (red now but usually it looks more appealing ;) ).

dmcardle:
Hey @ncw, this is ready for another look!

I've rewritten the e2e command as a proper Go test now. The test skips itself in a few situations:

* When "short" mode is enabled.
* When the OS is not supported (only Unixy OSes for now).
* When the rclone binary on the PATH reports a version that is not roughly the same as the VERSION file. This is just a rough check that should prevent misleading test results when someone runs `go test` locally.

A few questions for @ncw:

1. Should the new test skip itself when `git-annex` is not installed? If not, git-annex is effectively a required dependency for devs to install. Is there any documentation or scripts to update?

1. Are you concerned about the additional CI latency from `brew install git-annex` on macOS? It might be significant since it's building from source. Maybe there's a way to ask `brew` to install a binary package. Or maybe we can just have the GitHub action download a binary and verify its SHA256 sum.

1. The new e2e test skips itself when the rclone on PATH looks like the wrong version. It occurs to me that this could still produce misleading results when a dev checks out an old commit and runs `go test` without overriding the PATH. In this situation, the test wouldn't be able to differentiate between the released rclone binary and the locally built binary with the same version. Can I rely on the `-DEV` suffix being present when the rclone binary was not built for a release?

FYI @joeyh, I had to downgrade to protocol version 1. I discovered through CI that GitHub Actions' "ubuntu-latest" image (see the [image list](https://github.com/actions/runner-images)) ships with a version of git-annex from 2021 that doesn't support protocol version 2 :scream:. I scanned through the protocol spec and I believe downgrading to version 1 should be safe. However,  I want to add support for git-annex exports, which will require version 2. I suppose I'll cross this bridge when I come to it, and for now we can be compatible with older versions of git-annex.

dmcardle:
> 2\. Are you concerned about the additional CI latency from `brew install git-annex` on macOS? It might be significant since it's building from source. Maybe there's a way to ask `brew` to install a binary package. Or maybe we can just have the GitHub action download a binary and verify its SHA256 sum.

Looks like this was even more of an issue than I predicted. Both of the mac jobs failed after 60 minutes, so I'm assuming they timed out.

dmcardle:
Aha! Homebrew doesn't *need* to build things from source -- it has a binary cache, but cached binaries are called "bottles". We were just not hitting the cache because CI is using an old version of macOS.

Supporting evidence:

* There are no git-annex bottles for macOS 11 here: https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fgit-annex
* Further, the GitHub Actions "macos-11" image is deprecated according to <https://github.com/actions/runner-images?tab=readme-ov-file#available-images>.

@ncw, WDYT about upgrading to "macos-12"? Already tacked on another commit to this PR to give macos-12 a shot.

ncw:
> I've rewritten the e2e command as a proper Go test now. The test skips itself in a few situations:

Very nice
 
> * When "short" mode is enabled.
> * When the OS is not supported (only Unixy OSes for now).
> * When the rclone binary on the PATH reports a version that is not roughly the same as the VERSION file. This is just a rough check that should prevent misleading test results when someone runs `go test` locally.

Nice! We should probably make an `rclone version --json` to make it easier to parse, though you could use `rclone rc --loopback core/version`. Rclone has a version parser built in in `lib/version` which you might find useful.

> A few questions for @ncw:
> 
> 1. Should the new test skip itself when `git-annex` is not installed? If not, git-annex is effectively a required dependency for devs to install.

Yes I think it should skip. I know that, for example, the debian maintainers run these tests when they import a new code version.

> Is there any documentation or scripts to update?

If we want to put this in the integration tester (ie run it on all backends) then there is more stuff to do, but let's get this working first.

> 2. Are you concerned about the additional CI latency from `brew install git-annex` on macOS? It might be significant since it's building from source. Maybe there's a way to ask `brew` to install a binary package. Or maybe we can just have the GitHub action download a binary and verify its SHA256 sum.

It looks like you've fixed this now with the macOS upgrade and that step is taking 43 seconds in total for all the installs which seems fine.

> 3. The new e2e test skips itself when the rclone on PATH looks like the wrong version. It occurs to me that this could still produce misleading results when a dev checks out an old commit and runs `go test` without overriding the PATH. In this situation, the test wouldn't be able to differentiate between the released rclone binary and the locally built binary with the same version. Can I rely on the `-DEV` suffix being present when the rclone binary was not built for a release?

No. In the CI the code will be built with `make` so the version strings will look like this `v1.67.0-beta.7845.07e3beb67.fix-7689-empty-dirs` however I think the tests are not run with the same version string so they will look like `v1.67.0-DEV` or `v1.67.0` if building for the release.

I think this test will be a little approximate. Perhaps as belt and braces we could check the timestamp of the gitannex.go file vs the timestamp of the rclone binary - that would catch the dev editing the file and forgetting to rebuild rclone scenario which is probably the most common.

> Aha! Homebrew doesn't _need_ to build things from source -- it has a binary cache, but cached binaries are called "bottles". We were just not hitting the cache because CI is using an old version of macOS.
> 
> Supporting evidence:
> 
> * There are no git-annex bottles for macOS 11 here: https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fgit-annex
> * Further, the GitHub Actions "macos-11" image is deprecated according to https://github.com/actions/runner-images?tab=readme-ov-file#available-images.
> 
> @ncw, WDYT about upgrading to "macos-12"? Already tacked on another commit to this PR to give macos-12 a shot.

This always used to be `macos-latest` but in b447b0cd7833df0f7878be9911f3aa837abb7576 it got changed to macos-11 which at the time was in beta to get a newer version. So we could try changing this back to `macos-latest` which is what it should be. If that doesn't work then use `macos-12` and make an issue about it and I'll look into exactly how we get to `macos-latest`.

Looking at the [logs for a run](https://github.com/rclone/rclone/actions/runs/8562145909/job/23464807889?pr=7716)

I see this in the Run tests section

```
ok  	github.com/rclone/rclone/cmd/gitannex	36.085s
```

I looked through the code and it is looking like a very nice addition to the rclone test suite :-)
And this in the race section

```
ok  	github.com/rclone/rclone/cmd/gitannex	35.729s
```

That isn't adding 1 minute to the time though since Go is running these tests in parallel.

Those times look acceptable. I wouldn't want them to get much longer!

dmcardle:
Thanks, @ncw!

New changes:

* The test now skips itself when git-annex is not installed.
* Checking more version info.

  Thanks for the pointer to the version code. Now I don't need to read the VERSION file. Now I'm parsing a subset of the JSON output from "core/version" and just comparing to built-in constants.
* Upgraded CI to macos-latest.

ncw (approved):
This looks great to me now.

Are you ready for me to merge it?

dmcardle:
> This looks great to me now.
> 
> Are you ready for me to merge it?

Great, it's ready to go!

ncw:
Thank you :-)

PR #7722, opened 2024-04-04
docs: fix MANUAL formatting problems

pietvo:
1) Missing closing code backticks (```) in s3.md causes formatting problems 

2) Pandoc requires blank lines before ATX headings

I noticed big formatting problems in the manual. In the second half of the manual some text appeared as Markdown code, rather than formatted text. On the other hand some code (terminal in/output was not formatted as such but as inline text, almost unreadable. It appeared that a closing code mark (```) was missing.

Moreover, pandoc requires a blank line before the headings, although the markdown spec doesn't reuire it. So some headings were formatted as inline markdown code rather than as heading.

No.

#### Checklist

### DIFF
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -475,6 +475,7 @@ Mount option syntax includes a few extra options treated specially:
 - `vv...` will be transformed into appropriate `--verbose=N`
 - standard mount options like `x-systemd.automount`, `_netdev`, `nosuid` and alike
   are intended only for Automountd and ignored by rclone.
+
 ## VFS - Virtual File System
 
 This command uses the VFS layer. This adapts the cloud storage objects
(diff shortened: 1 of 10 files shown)


albertony (approved):

ncw:
I think to fix these ones (which are auto generated)

```
## VFS - Virtual File System
```

We need to add a new line to the start of this file https://github.com/rclone/rclone/blob/master/vfs/vfs.md or add a new line to the places it is used, eg

https://github.com/rclone/rclone/blob/1e3ab7acfd4dcbda863d0a15b06c51789ad0a075/cmd/serve/s3/s3.go#L52

albertony:
Ah, right, sorry I missed that. Perhaps adding it where it is used will make the reason for it more clear, less chance of a later "cleanup" re-introducing the issue?

albertony:
I'm looking into a suggestion, here: https://github.com/rclone/rclone/pull/7727 (not ready yet).

pietvo:
I also noted that the section `## VFS - Virtual File System` occurs 10 times in `MANUAL.md`, which seems a bit odd to me. But each time it is followed by a different `rclone` command. I wonder if this couldn't be done in a more efficient way.

PR #7727, opened 2024-04-05
Ensure empty line between text and a following heading

albertony:

To ensure there is an empty line between text and a following heading. Because some consumers (like pandoc) may otherwise not parse the Markdown correctly.

The challenge here is when combining different help string snippets, e.g. common http and vfs help into a specific serve command's help. I think we have tried to make sure string constants containing help snippets contain an empty line at the end, so that simple concatenation `help1 + help2` will always give correct result, but there was some not conforming to this. However an additional challenge is that some help snippets are read from markdown file using `go:embed`. To make the use of these consistent with others, without having to remember to add `\n` constants when concatenating these or something like that, I wrapped a function around the embedded markdown, like this:

```go
//go:embed help.md
var help string

// Help returns the formatted help string
func help() string {
	return strings.TrimSpace(help) + "\n\n"
}
```

Additional:
- Consistent no newline at beginning of help for commands etc.
  - Most already conformed to this, but there were some outliers that I cleaned up.
- Empty line after headings (between heading and body).
  - Most already conformed to this, but there were some outliers that I cleaned up.

https://github.com/rclone/rclone/pull/7722

#### Checklist

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -31,6 +31,11 @@ import (
 //go:embed mount.md
 var mountHelp string
 
+// help returns the help string cleaned up to simplify appending
+func help(commandName string) string {
+	return strings.TrimSpace(strings.ReplaceAll(mountHelp, "@", commandName)) + "\n\n"
+}
+
 // Options for creating the mount
 type Options struct {
 	DebugFUSE          bool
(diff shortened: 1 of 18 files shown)


ncw (approved):
Looks great - please merge.

Thank you :-)

PR #7731, opened 2024-04-05
 ncdu: Do not quit on Esc 

unresolv:

prevent quitting ncdu when pressing 'escape' or 'q' - dedicating that to closing the box

Ctrl-C will still quit

#### Checklist

### DIFF
--- a/cmd/ncdu/ncdu.go
+++ b/cmd/ncdu/ncdu.go
@@ -114,7 +114,8 @@ func helpText() (tr []string) {
 		" ^L refresh screen (fix screen corruption)",
 		" r recalculate file sizes",
 		" ? to toggle help on and off",
-		" q/ESC/^c to quit",
+		" ESC to close the menu box",
+		" q/^c to quit",
 	}...)
 	return
 }
@@ -990,7 +991,7 @@ outer:
 				}
 				switch c {
 				case key(tcell.KeyEsc), key(tcell.KeyCtrlC), 'q':
-					if u.showBox {
+					if u.showBox || c == key(tcell.KeyEsc) {
 						u.showBox = false
 					} else {
 						break outer


fireonlive:
I'd like this; I often accidentally close ncdu after referencing the help pop-up. Sadly, it's also almost always with a remote that takes a long time to generate the listing and with lots of API requests that I may be charged for.

ncw:
It was the intention that `rclone ncdu` copies exactly the way `rclone ncdu` works.

For `ncdu`
- the key `q` and `ESC` closes the dialog. 
- the key `q` and `CTRL-C` quits.

For `rclone ncdu`
- the key `q` and `ESC` closes the dialog. 
- the key `q` and `ESC` and `CTRL-C` quits.

So maybe we should make the change that `ESC` no longer quits and encourage people to use that to close the dialog.

That would be `ncdu` compatible.

I would certainly miss `q` to close `rclone ncdu` as that is what I normally use!

What do you think?

unresolv:
@ncw thank you for your feedback!

that sounds good to me, and makes the diff a little smaller too :)

So now, for Ctrl-C, q, Esc, we:
1. Check if the box is there, or we are pressing Esc, if so, we close the box.
2. Otherwise, we exit

This means Esc will never close rclone ncdu, but q and Ctrl-C will work the same as before (close the box, then quit)

albertony:
Docs here need an update as well:

https://github.com/rclone/rclone/blob/2ab2ec29f9230f2959ae1e4428b07a4c7080aaed/cmd/ncdu/ncdu.go#L117

ncw (commented):
This is looking good now, but please add the docs as @albertony suggested - thank you :-)

unresolv:
@ncdu @albertony thank you, edited the documentation, as there was no entry for closing the box I added one for ESC, and removed ESC from quit

albertony:
> edited the documentation, as there was no entry for closing the box I added one for ESC, and removed ESC from quit

Perfect!

> @ncdu [...] thank you

Yes, I can join in with a big thanks to ncdu 😝

unresolv:
Oops I meant @ncw :’) 😊

unresolv:
thank you very much @ncw!

Issue #1644, opened 2017-09-01
Chroot for Remotes


ncw:
Make a `Chroot` call which makes a copy of a remote with a new root.  This can be used to avoid inefficiencies calling NewFs multiple times (see #1640 for crypt), also for `--backup-dir`.

In an ideal world this would enable us to hoist the checking to see whether the remote points to a file out of all the remotes, simplifying all of them.

ivandeex:
@ncw Will `alias` suffice?

PR #7743, opened 2024-04-08
bisync: more fixes for integration tests

nielash:

More fixes for bisync integration tests:

- `operations`: fix `Move` ignoring `remote` when falling back to `Copy`
- fix endless loop if lockfile decoder returns a non-nil error that is not `io.EOF`
- make `tempDir` path shorter to avoid exceeding linux filename length limits
- use `fs.ConfigStringFull` instead of `bilib.StripHexString` to properly reverse connection string remotes
- `chunker`: fix `NewFs` when root points to composite multi-chunk file without metadata
- `chunker`: fix case-insensitive comparison on `local` without metadata

- #7665 

#### Checklist

### DIFF
--- a/fs/operations/operations.go
+++ b/fs/operations/operations.go
@@ -446,6 +446,7 @@ func move(ctx context.Context, fdst fs.Fs, dst fs.Object, remote string, src fs.
 	if doMove := fdst.Features().Move; doMove != nil && (SameConfig(src.Fs(), fdst) || (SameRemoteType(src.Fs(), fdst) && (fdst.Features().ServerSideAcrossConfigs || ci.ServerSideAcrossConfigs))) {
 		// Delete destination if it exists and is not the same file as src (could be same file while seemingly different if the remote is case insensitive)
 		if dst != nil {
+			remote = dst.Remote()
 			if !SameObject(src, dst) {
 				err = DeleteFile(ctx, dst)
 				if err != nil {
@@ -455,6 +456,10 @@ func move(ctx context.Context, fdst fs.Fs, dst fs.Object, remote string, src fs.
 				doMove = func(ctx context.Context, src fs.Object, remote string) (fs.Object, error) {
 					return MoveCaseInsensitive(ctx, fdst, fdst, remote, src.Remote(), false, src)
 				}
+			}
+		} else if needsMoveCaseInsensitive(fdst, fdst, remote, src.Remote(), false) {
+			doMove = func(ctx context.Context, src fs.Object, remote string) (fs.Object, error) {
+				return MoveCaseInsensitive(ctx, fdst, fdst, remote, src.Remote(), false, src)
 			}
 		}
 		// Move dst <- src
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -386,7 +386,7 @@ func (s *syncCopyMove) pairChecker(in *pipe, out *pipe, fraction int, wg *sync.W
 			}
 			// Fix case for case insensitive filesystems
 			if s.ci.FixCase && !s.ci.Immutable && src.Remote() != pair.Dst.Remote() {
-				if newDst, err := operations.Move(s.ctx, s.fdst, pair.Dst, src.Remote(), pair.Dst); err != nil {
+				if newDst, err := operations.Move(s.ctx, s.fdst, nil, src.Remote(), pair.Dst); err != nil {
 					fs.Errorf(pair.Dst, "Error while attempting to rename to %s: %v", src.Remote(), err)
 					s.processError(err)
 				} else {


backend/s3/s3.go
```diff
@@ -4623,6 +4623,13 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 		}
 		setFrom_s3CopyObjectInput_s3PutObjectInput(&req, ui.req)
 		req.MetadataDirective = aws.String(s3.MetadataDirectiveReplace)
+	} else if strings.EqualFold(srcPath, dstPath) { // prevent "trying to copy an object to itself..." error
```
ncw, line 4626: I'm not sure about this fix. S3 is supposed to be case insensitive so how does this manifest? Is it in `serve s3` on macOS maybe?

Also you are allowed to copy objects over themselves normally - that is how you set the metadata.

How do you reproduce the problem that this fixes?

cmd/bisync/bisync_test.go
```diff
@@ -275,7 +275,7 @@ func testBisync(t *testing.T, path1, path2 string) {
 
 	baseDir, err := os.Getwd()
 	require.NoError(t, err, "get current directory")
-	randName := "bisync-" + time.Now().Format("150405-") + random.String(5) // AzureBlob doesn't like dots
+	randName := "bs-" + time.Now().Format("150405-") + random.String(2) // some bucket backends don't like dots, keep this short to avoid linux errors
```
ncw, line 278: This kind of implies that you are creating top level directories (which are buckets in s3 compatibles) - these have many limits under s3 compatibles.

I don't think you should be creating top level directories though, you should be doing this inside ones created by `fstest.RandomRemote` or using `fstest.RandomRemote` to create them.

cmd/bisync/bisync_test.go
```diff
@@ -1897,3 +1899,13 @@ func checkError(t *testing.T, err error, msgAndArgs ...interface{}) {
 	}
 	assert.NoError(t, err, msgAndArgs...)
 }
+
+// for example, replaces TestS3{juk_h}:dir with TestS3,directory_markers=true:dir
+// because NewFs needs the latter
+func (b *bisyncTest) replaceHex(remote string) string {
+	if bilib.HasHexString(remote) {
+		remote = strings.ReplaceAll(remote, fs.ConfigString(b.parent1), fs.ConfigStringFull(b.parent1))
```
ncw, line 1907: Just note that `ConfigStringFull` often contains credentials so it does **not** want to go in any logs.

So I think this is probably not a good idea...

What are you trying to fix?

ncw (commented):
I have some questions inline :-)

Assuming the EOF test is the cause of the lockups we should get that merged sooner rather than later so If you split it into a new commit I can cherry-pick it.

backend/s3/s3.go
nielash: Here's an example of the issue from the integration tests: https://pub.rclone.org/integration-tests/2024-04-07-010011/s3-cmd.bisync-TestS3-5.txt (`ctrl+F` the word `InvalidRequest` to jump to the error)

So I think that rules out the possibility that it is specific to mac or `serve s3`.

I can reproduce it locally with AWS, and have verified that the same test passes with this patch. But that doesn't mean my patch is the correct fix 🙂 

It is possible that this issue affects only a subset of equal-folding names (like maybe unicode normalization form differences but not simple ASCII casing differences)

cmd/bisync/bisync_test.go
nielash: This is inside `fstest.RandomRemote` (for every remote except `local`.) For example, a path might end up looking like:

`TestS3:rclone-test-qemaser8watigiw0haguhij9/bisync-013220-biqaq/normalization/path1/file1.txt`

The listing filename ends up concatenating both paths, like:
`TestS3_rclone-test-qemaser8watigiw0haguhij9_bisync-013220-biqaq_extended_char_paths_path1_測試_Русский_____ě_áñ..TestS3_rclone-test-qemaser8watigiw0haguhij9_bisync-013220-biqaq_extended_char_paths_path2_測試_Русский_____ě_áñ.path1.lst`

which is why when we added `fstest.RandomRemote` a few days ago, it started causing `file name too long` errors locally (which I think is the underlying reason for the json decoding error which caused the endless loop.)

So basically I'm trying to make the path shorter by shortening the `bisync-013220-biqaq` part, to stay under that limit (and check errors properly so that worst case scenario it fails immediately.)

An alternative would be to use fewer than 24 random chars in `fstest.RandomRemote`...but I figured you wouldn't want to mess with that, and I don't think we have to.

cmd/bisync/bisync_test.go
nielash: Hmm, ok. The problem I'm trying to solve is this, let me know if there's a better solution:

The test case scenarios have lots of globs like
`copy-as {datadir/}file1.txt {path1/}subdir file20.txt`
which get translated into instructions to set the stage before the test.

The `{path1/}` variable usually ends up looking like `TestS3:rclone-test-qemaser8watigiw0haguhij9/bisync-013220-biqaq/normalization/path1/`, which is fine, as we can just pass that directly into `cache.Get` or `fs.NewFs` and all will be well.

But if the config is overridden then it will have a hexstring, like `TestS3{juk_h}:rclone-test-qemaser8watigiw0haguhij9/bisync-013220-biqaq/normalization/path1/`. In that case, we can't just pass it in directly, because `NewFs` will complain about `config name contains invalid characters`, because of those braces. Previously we just stripped the hexstring off, but this was really not a good idea, as then the Fs returned did not have the right options (for example, `directory_markers` changes whether or not we can do the `createemptysrcdirs` test.)

`NewFs` will happily accept `TestS3,directory_markers=true:dir`, so I am looking for a way to replace `TestS3{juk_h}:dir` with `TestS3,directory_markers=true:dir`. Or a different way of reverse engineering a `TestS3{juk_h}:dir` string into an `fs` with the right config.

By the way, is it intentional that `cache.Get` can store a canonical name if given an FsString but can't accept a canonical name to look up a cached Fs? That was surprising to me.

nielash:
@ncw I've split these into separate cherry-pickable commits. I'd recommend pushing https://github.com/rclone/rclone/pull/7743/commits/f2485b4814b7dd2eb60f6c5ecb121240ea591901 and https://github.com/rclone/rclone/pull/7743/commits/56bd493829e16a117bae5f82c8967ec6bc31cef1 together, as I think the former is the real cause of the error, and without it there will still be errors -- just faster ones. But, up to you!

Also I'm not seeing a test from yesterday...did it crash or something?

backend/s3/s3.go
ncw: AWS S3 [should ignore unicode normalization and case](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html).

I managed to extract the request and response from the tests using

    go test -v -remote TestS3: -debug -verbose -run '^TestBisyncRemoteRemote$/^(extended_char_paths|normalization)$' -dump-bodies 2>&1 | tee /tmp/bisync-s3.log


```
2024/04/09 11:55:35 DEBUG : HTTP REQUEST (req 0xc000bb4400)
2024/04/09 11:55:35 DEBUG : PUT /bisync-115522-pacup/normalization/path2/folder/e%CC%81e%CC%81e%CC%81o%CC%88.txt HTTP/1.1
Host: rclone-test-vocekot3vekazev9picehox8.s3.eu-west-2.amazonaws.com
User-Agent: rclone/
Content-Length: 0
Authorization: XXXX
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Copy-Source: rclone-test-vocekot3vekazev9picehox8/bisync-115522-pacup/normalization/path2/folder/e%CC%81e%CC%81e%CC%81o%CC%88.txt
X-Amz-Date: 20240409T115535Z
X-Amz-Metadata-Directive: COPY
Accept-Encoding: gzip

2024/04/09 11:55:35 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2024/04/09 11:55:35 DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
2024/04/09 11:55:35 DEBUG : HTTP RESPONSE (req 0xc0004b2800)
2024/04/09 11:55:35 DEBUG : HTTP/1.1 400 Bad Request
Connection: close
Transfer-Encoding: chunked
Content-Type: application/xml
Date: Tue, 09 Apr 2024 11:55:34 GMT
Server: AmazonS3
X-Amz-Id-2: aPQJqzOJW77MdBbSPC4JR2gC4FlP8SPMs1WZa/vfdXH2b/d75KQOJp5QrgYteGfBFXpVbiWxeIY=
X-Amz-Request-Id: 83HASN30WZKYJPC5

179
<Error><Code>InvalidRequest</Code><Message>This copy request is illegal because it is trying to copy an object to itself without changing the object's metadata, storage class, website redirect location or encryption attributes.</Message><RequestId>83HASN30WZKYJPC5</RequestId><HostId>aPQJqzOJW77MdBbSPC4JR2gC4FlP8SPMs1WZa/vfdXH2b/d75KQOJp5QrgYteGfBFXpVbiWxeIY=</HostId></Error>
```

Here you can see that the object mentioned in the `PUT` and the object in the `X-Amz-Copy-Source` are actually the same, which is why AWS is complaining (here is me proving it to myself with cut and paste and python!)

```
>>> "e%CC%81e%CC%81e%CC%81o%CC%88.txt" == "e%CC%81e%CC%81e%CC%81o%CC%88.txt"
True
```

Hopefully that will give you something to dig into - I have to go for a lunch appointment but can help more this afternoon.

cmd/bisync/bisync_test.go
ncw: OK that is fine - I just wanted to make sure you weren't creating buckets :-)

backend/s3/s3.go
ncw: I don't think changing this line is correct

https://github.com/rclone/rclone/blob/2ab2ec29f9230f2959ae1e4428b07a4c7080aaed/fs/operations/operations.go#L485

If we do that then Move will no longer obey its contract

https://github.com/rclone/rclone/blob/2ab2ec29f9230f2959ae1e4428b07a4c7080aaed/fs/operations/operations.go#L400-L401

Where is this called from? I suspect the caller should be setting `dst` to `nil`.

ncw:
> @ncw I've split these into separate cherry-pickable commits. I'd recommend pushing [f2485b4](https://github.com/rclone/rclone/commit/f2485b4814b7dd2eb60f6c5ecb121240ea591901) and [56bd493](https://github.com/rclone/rclone/commit/56bd493829e16a117bae5f82c8967ec6bc31cef1) together, as I think the former is the real cause of the error, and without it there will still be errors -- just faster ones. But, up to you!

Thank you - have done that now.
 
> Also I'm not seeing a test from yesterday...did it crash or something?

According to the logs... it finished at 15:05 having started at 01:00 :-(

It looks like the test results are available, they just didn't get linked properly - not sure why

https://pub.rclone.org/integration-tests/2024-04-09-010011/index.html

You can look at this directory to see all the tests we have

https://pub.rclone.org/integration-tests/

cmd/bisync/bisync_test.go
ncw: > Just note that `ConfigStringFull` often contains credentials so it does **not** want to go in any logs.

OK, I think I was being overly alarmist there!

Yes it is true that lots of rclone users construct on the fly remotes with secrets in, and these **will** appear in the output of `ConfigStringFull`, so we should try not to log those.

**However** none of the test remotes do that so I think it is perfectly safe to do this in test code only and a quick skim convinces me that your change is only used in test code. And in fact we do use this already in the b2 and s3 internal tests for an almost identical reason.

So with that proviso, I think what you've got is OK, and many apologies for overstating the problem!

> By the way, is it intentional that cache.Get can store a canonical name if given an FsString but can't accept a canonical name to look up a cached Fs? That was surprising to me.

I made the cache mostly for the rc/api so backends don't get created and destroyed too often. So its purpose was to cache things by the name the user gave it, and generally users don't change the name they use.

We could make the cache look up things by canonical name and that might not be a bad idea, but I suspect it would be of limited use outside the tests.

backend/s3/s3.go
nielash: It is called from [here](https://github.com/rclone/rclone/blob/04128f97ee2444253b636ed52c4138d1e754b8f3/fs/sync/sync.go#L389), and I don't know if passing `nil` there is the right fix either, due to this warning:

https://github.com/rclone/rclone/blob/04128f97ee2444253b636ed52c4138d1e754b8f3/fs/operations/operations.go#L403-L406

nielash:
> You can look at this directory to see all the tests we have
> 
> https://pub.rclone.org/integration-tests/

Just to clarify: my comment on the 9th was asking about the test from the 8th, which I still don't see. I see folders for `2024-04-07` and `2024-04-09`, but not `2024-04-08`. But probably not worth any further investigation, as it is out of date now anyway and more recent tests have completed.

cmd/bisync/bisync_test.go
nielash: Ok, thanks for clarifying! You are right that this is only used on the tests, so I think it's ok. But good to know about `ConfigStringFull` for my future knowledge.

> We could make the cache look up things by canonical name and that might not be a bad idea, but I suspect it would be of limited use outside the tests.

There are times when it might be useful for `chroot`-ing an Fs, but I think your proposal on #1644 is a more efficient solution to that problem.

backend/s3/s3.go
ncw: What calls `operations.Move`? I made a stack trace and here is the answer

```
Stack trace:
goroutine 6818 [running]:
github.com/rclone/rclone/fs/log.Stack({0x206f080, 0xc000acc120}, {0xc000a6f900, 0x4a9})
	rclone/rclone/fs/log/log.go:86 +0x6e
github.com/rclone/rclone/fs/operations.move({0x2714728, 0xc000bab590}, {0x2728090?, 0xc000c9e2c0?}, {0x2728100, 0xc000acc120}, {0xc00124c1e1, 0x17}, {0x2728100, 0xc000acc120}, ...)
	rclone/rclone/fs/operations/operations.go:427 +0x1e8
github.com/rclone/rclone/fs/operations.Move(...)
	rclone/rclone/fs/operations/operations.go:414
github.com/rclone/rclone/fs/sync.(*syncCopyMove).pairChecker(0xc00070c500, 0x0?, 0x0?, 0xffffffffffffffff?, 0xc000bf5f98?)
	rclone/rclone/fs/sync/sync.go:389 +0x445
created by github.com/rclone/rclone/fs/sync.(*syncCopyMove).startCheckers in goroutine 5104
	rclone/rclone/fs/sync/sync.go:512 +0x45
```

I think the fundamental problem is that the `dst` object should point to an already existing **destination** but it is pointing to the **source** object I think.  That is `dst.Remote()` should be equal to `remote`.

The source of the call is here

https://github.com/rclone/rclone/blob/04128f97ee2444253b636ed52c4138d1e754b8f3/fs/sync/sync.go#L389

And you can see that it is quite clearly being called with the same object as `src` and `dst`. I think if we replace `dst` with `nil` here all will be well.

I made a fix for that here 8584bef00614a5642478b9ea4935d2ccd2492661

[v1.67.0-beta.7864.8584bef00.fix-7743-case-norm-sync](https://beta.rclone.org/branch/fix-7743-case-norm-sync/v1.67.0-beta.7864.8584bef00.fix-7743-case-norm-sync/) on branch [fix-7743-case-norm-sync](https://github.com/rclone/rclone/tree/fix-7743-case-norm-sync) (uploaded in 15-30 mins)

I'd appreciate your thoughts before I merge it though.

ncw: I forgot to say, this fixes the bisync s3 tests above

    go test -v -remote TestS3:  -run '^TestBisyncRemoteRemote$/^(extended_char_paths|normalization)$'

ncw:
> Just to clarify: my comment on the 9th was asking about the test from the 8th, which I still don't see. I see folders for `2024-04-07` and `2024-04-09`, but not `2024-04-08`. But probably not worth any further investigation, as it is out of date now anyway and more recent tests have completed.

Yes that is the test I was looking for too! I got confused at one point and put a link to the wrong directory but I ~~removed~~ it later. It is on the server but never finished or made it to the website and is now consigned to the oblivion of history!

I've cherry picked all the commits here to master **except** the operations one which I don't think is correct (see above). My replacement for that is here 8584bef (see notes above!)

It looks like we are getting there :-) Land is in sight on this long voyage...

At some point we are going to discover that some of the backends are just too slow to run the bisync tests in a reasonable length of time and we are going to have to disable the tests on them.

In config.yaml some tests are like this

```yaml
 - backend:  "zoho"
   remote:   "TestZoho:"
   fastlist: false
   tests:
     - backend
```

Which means only run the backend tests.

We can extend this if we need to to maybe say to not run the bisync tests on certain backends.

```yaml
 - backend:  "xxx"
   remote:   "TestXxx:"
   fastlist: false
   tests:
     - !cmd/bisync
```

I'd like for the tests to be done in < 4 hours, ideally in 2 hours as when making stable releases I need to run the tests on the stable release branch and if the tester is spending all night and day on the main tests there is never a moment!

We might want to consider increasing the number of simultaneous tests also.

backend/s3/s3.go
nielash: I think this works for `s3` because it's not case-insensitive, but this would break things for backends that actually are case-insensitive. For example, this fails for me:

```
go test ./fs/sync -v -remote TestDropbox:  -run '^TestFixCase$'
```

I think what makes this confusing is that on a case-insensitive backend, `hello` and `HELLO` actually _are_ considered the same remote, so if `hello` exists then the dst `HELLO` must also exist, even though it really doesn't. I think that's why the `src` and `dst` were passed as the same object -- because looking for the `dst` will end up finding the `src`. And so if we pass `nil` then we are doing what the comment warned about -- passing `nil` when the destination does exist, according to case-insensitive logic.

nielash: With this additional change it seems to work: https://github.com/nielash/rclone/commit/cec22a5827dc172eb8b082f1336ac37388116386 

Maybe that's the missing piece?

nielash:
> At some point we are going to discover that some of the backends are just too slow to run the bisync tests in a reasonable length of time and we are going to have to disable the tests on them.

One other thought... currently it runs 3 combinations of each test for each remote:

`TestBisyncRemoteLocal`
`TestBisyncLocalRemote`
`TestBisyncRemoteRemote`

This is helpful as sometimes one test will pick up a problem that the other two miss (like this `s3` issue which only happens with server-side copy.) But perhaps one is enough for certain slow backends? Seems better than disabling the tests entirely, as that way we still have some chance of finding out if things break in the future.

backend/s3/s3.go
ncw: Yes I think that looks correct. Do you want to pull all the remaining bits into this PR and we can merge it?

ncw:
> > At some point we are going to discover that some of the backends are just too slow to run the bisync tests in a reasonable length of time and we are going to have to disable the tests on them.
> 
> One other thought... currently it runs 3 combinations of each test for each remote:
> 
> `TestBisyncRemoteLocal` `TestBisyncLocalRemote` `TestBisyncRemoteRemote`
> 
> This is helpful as sometimes one test will pick up a problem that the other two miss (like this `s3` issue which only happens with server-side copy.) But perhaps one is enough for certain slow backends? Seems better than disabling the tests entirely, as that way we still have some chance of finding out if things break in the future.

OK, that seems like a good idea.

We do allow the `-short` flag to be set in backends and tests and you could use that in bisync to only run the first of those tests by checking `testing.Short`

https://github.com/rclone/rclone/blob/2c7680050bb189efe1770f77585527619197b596/fstest/test_all/config.go#L19

https://github.com/rclone/rclone/blob/2c7680050bb189efe1770f77585527619197b596/fstest/test_all/config.go#L33

backend/s3/s3.go
nielash: Will do.

nielash:
> I'd like for the tests to be done in < 4 hours

One other observation is that the "retries" seem to be including some tests that already passed (or were skipped). For example, given the following failures:

```
TestBisyncRemoteLocal/resolve, TestBisyncLocalRemote/resolve, TestBisyncRemoteRemote/normalization, TestBisyncRemoteRemote/resolve
```

it retries this:
```
-test.run '^(TestBisyncLocalRemote|TestBisyncRemoteLocal|TestBisyncRemoteRemote)$/^(normalization|resolve)$'
```
instead of this:
```
-test.run '^(TestBisyncRemoteLocal\/resolve|TestBisyncLocalRemote\/resolve|TestBisyncRemoteRemote\/normalization|TestBisyncRemoteRemote\/resolve)$'
```

And also a timeout causes all of them to be retried 5 times, even if there were no failures (prior to the timeout):
https://pub.rclone.org/integration-tests/current/drive-cmd.bisync-TestDrive-5.txt

Not sure if that's an easy optimization, but it would probably speed things up considerably! (And prevent the timeouts in the first place)

nielash:
Also looks like we're still slightly over the 255 limit for the remotes with long names like `TestChunkerChunk50bSHA1QuickS3:` 🫤 I'll try to find something else to shorten...

ncw (approved):
This is looking good now. What a difficult bug to fix properly!

I'll merge this which will close this PR as I think it has gone on long enough!

I'll reply to the other things in a second.

ncw:
> One other observation is that the "retries" seem to be including some tests that already passed (or were skipped). For example, given the following failures:
> 
> ```
> TestBisyncRemoteLocal/resolve, TestBisyncLocalRemote/resolve, TestBisyncRemoteRemote/normalization, TestBisyncRemoteRemote/resolve
> ```
> 
> it retries this:
> 
> ```
> -test.run '^(TestBisyncLocalRemote|TestBisyncRemoteLocal|TestBisyncRemoteRemote)$/^(normalization|resolve)$'
> ```
> 
> instead of this:
> 
> ```
> -test.run '^(TestBisyncRemoteLocal\/resolve|TestBisyncLocalRemote\/resolve|TestBisyncRemoteRemote\/normalization|TestBisyncRemoteRemote\/resolve)$'
> ```

Hmm, that is probably a bug with the regex consolidation code which turns a list of failed tests into a regexp.

https://github.com/rclone/rclone/blob/5323a2189852b29e67dd1ff9cb726518e4612f59/fstest/test_all/run.go#L103-L105

It is being a bit too clever favouring shortness of regexp over correctness/efficiency!

Fixing that would probably be a good idea - have you got time to take a look at it? It has some test cases too which look, well, wrong!

https://github.com/rclone/rclone/blob/5323a2189852b29e67dd1ff9cb726518e4612f59/fstest/test_all/run_test.go#L62-L67

> And also a timeout causes all of them to be retried 5 times, even if there were no failures (prior to the timeout): https://pub.rclone.org/integration-tests/current/drive-cmd.bisync-TestDrive-5.txt

Unfortunately when we get a timeout we don't know how far we got through the test. Do we throw away the existing list of good/bad tests at that point? We shouldn't do that.

> Not sure if that's an easy optimization, but it would probably speed things up considerably! (And prevent the timeouts in the first place)

I don't think it is easy to skip tests by regexp so we can't just skip tests which passed if that is what you were thinking. Go doesn't support negative lookahead assertions `(?!...)` which is what you'd need I think (and there is a good reason for that as you can't implement it without opening the regexp engine up to exponential backtracking attacks).

> Also looks like we're still slightly over the 255 limit for the remotes with long names like `TestChunkerChunk50bSHA1QuickS3:` 🫤 I'll try to find something else to shorten...

:-(

nielash:
> Fixing that would probably be a good idea - have you got time to take a look at it? It has some test cases too which look, well, wrong!

I can try to take a look. But I think all I'd be doing is taking out the stuff that splits and consolidates by level and using the full path for each failed test... is that what you had in mind too or am I misunderstanding?

> I don't think it is easy to skip tests by regexp so we can't just skip tests which passed if that is what you were thinking.

I was thinking retry the failures instead of skip the successes... but maybe that's the same problem? I haven't really looked at how this code works yet so maybe once I do I'll quickly see why this isn't going to be possible 😄

ncw:
> > Fixing that would probably be a good idea - have you got time to take a look at it? It has some test cases too which look, well, wrong!
> 
> I can try to take a look. But I think all I'd be doing is taking out the stuff that splits and consolidates by level and using the full path for each failed test... is that what you had in mind too or am I misunderstanding?

I just had another look at this.

Looking at [the docs](https://pkg.go.dev/cmd/go#hdr-Testing_flags) 

> -run regexp
>     Run only those tests, examples, and fuzz tests matching the regular
>     expression. For tests, the regular expression is split by unbracketed
>     slash (/) characters into a sequence of regular expressions, and each
>     part of a test's identifier must match the corresponding element in
>     the sequence, if any. Note that possible parents of matches are
>     run too, so that -run=X/Y matches and runs and reports the result
>     of all tests matching X, even those without sub-tests matching Y,
>     because it must run them to look for those sub-tests.
>     See also -skip.

So `-run Test1/subtest` works but `-run Test1[/]subtest` doesn't in my testing so we need to match each level of the test indivdually.

```
~/go/src/github.com/rclone/rclone/fs/rc$ go test -v -run 'TestParamsGetDuration/string=wrong'
=== RUN   TestParamsGetDuration
=== RUN   TestParamsGetDuration/string=wrong
--- PASS: TestParamsGetDuration (0.00s)
    --- PASS: TestParamsGetDuration/string=wrong (0.00s)
PASS
ok  	github.com/rclone/rclone/fs/rc	0.009s

~/go/src/github.com/rclone/rclone/fs/rc$ go test -v -run 'TestParamsGetDuration(/)string=wrong'
testing: warning: no tests to run
PASS
ok  	github.com/rclone/rclone/fs/rc	0.008s

~/go/src/github.com/rclone/rclone/fs/rc$ go test -v -run 'TestParamsGetDuration\/string=wrong'
testing: warning: no tests to run
PASS
ok  	github.com/rclone/rclone/fs/rc	0.005s
```

I spent a little time building a trie to combine the regexp matches better which was fun but it doesn't work :-(

So I don't think it is possible using just `-run`. I guess we could execute the test suite multiple times which is a little inefficient but maybe better than what we are doing at the moment.

> > I don't think it is easy to skip tests by regexp so we can't just skip tests which passed if that is what you were thinking.
> 
> I was thinking retry the failures instead of skip the successes... but maybe that's the same problem? I haven't really looked at how this code works yet so maybe once I do I'll quickly see why this isn't going to be possible 😄

We do have `-skip` though which I didn't realise

> -skip regexp
>     Run only those tests, examples, fuzz tests, and benchmarks that
>     do not match the regular expression. Like for -run and -bench,
>     for tests and benchmarks, the regular expression is split by unbracketed
>     slash (/) characters into a sequence of regular expressions, and each
>     part of a test's identifier must match the corresponding element in
>     the sequence, if any.

which might help.

nielash:
> Note that possible parents of matches are
> run too, so that -run=X/Y matches and runs and reports the result
> of all tests matching X, even those without sub-tests matching Y,
> because it must run them to look for those sub-tests.

Argh, that's kind of annoying!

However I'm a little confused, because after playing around with it I'm finding that the following does what I want:

```
-run '^TestBisyncRemoteLocal/resolve|TestBisyncLocalRemote/resolve|TestBisyncRemoteRemote/normalization|TestBisyncRemoteRemote/resolve$'
```

(note the no parentheses and unescaped slashes)

I think, upon re-reading, that maybe what they mean is that `-run=X/Y` matches `X` and `X/Y` but not `X/Z`? Which is fine, at least for bisync's purposes.

> We do have `-skip` though which I didn't realise

That has potential! Perhaps we could reuse the `findFailures()` logic to look for `PASS` and `SKIP`, and then `-skip` them.

ncw:
I think I've corrected my mental model of how this works and the code which generates the regexps. I wrote some nice tests earlier and this new revision fixes some bugs in those tests so I think it is working well!

I've merged that fix so hopefully that will help a bit.

What do we need to do next? I'd like to reduce the number of bisync errors as 70 errors is too many to look through!

The top bisync error is `TestBisyncRemoteRemote/ext_paths` by the look of it.

If you find insoluble problems you can tell the tester that you don't care if they fail like this

https://github.com/rclone/rclone/blob/2cff5514aadbd02320eca6932737c9f892f29ab2/fstest/test_all/config.yaml#L209-L217

You noticed this the other day I think (in a message I haven't replied to yet, and have now lost!).

ncw:
> > The top bisync error is `TestBisyncRemoteRemote/ext_paths` by the look of it.
> 
> Yes, and that is the one caused by the filename being longer than 255 bytes. I think I have a fix for that which I'll submit shortly.

Hooray! :fireworks: 
 
> The other large pattern is tests simply timing out without actually failing (does your fix above address this case too?) For example, `TestDrive:` passes locally for me, and I suspect that if the `-timeout` were larger it would eventually pass on your end too. But instead it fails 5 times, because they all restart from the beginning, instead of from the last successful test, and then time out before they have the chance to pass.

My fix doesn't address that.

You can adjust the timeout per backend with the `extratime` parameter. It would be easy to add a similar parameter to the tests, so we allow bisync twice the time to complete. I'm not sure that is the right approach though.

How about splitting the bisync tests into 3 separate tests. It would be fairly easy to add a `testname` flag which would be the name of the test to run (in a suitable form to pass to `go test -run`). Splitting the tests would have the advantage that they can be run in parallel too.

> `TestSFTPRcloneSSH:` is failing because its `ConfigStringFull` has spaces in it due to `TestSFTPRcloneSSH,ssh='ssh -i /tmp/rclone-serve-sftp-ssh.key `... That's a bit tricky to solve gracefully.

You can just skip the test there if you want - you won't get any different behavior to `TestSFTPRclone`.

> `TestChunkerMailru:` is failing due to `Error listing: invalid character '<' looking for beginning of value` which I haven't done much investigation on yet. I'm inclined to blame it on `mailru` (we found with the `--fix-case` tests that it really struggles with special characters), but that doesn't quite make sense as `TestMailru:` is passing. Other `chunker` combinations are also passing, so it seems like it's something specific to the combination of `chunker` wrapping `mailru`. (EDIT: I now see someone else has reported the same issue: #7768)

I think that is the server overloading...

> The other failures seem like backend-specific bugs that I will try to narrow down (this will be easier when the more systemic problems are out of the way)

:-)

nielash:
> How about splitting the bisync tests into 3 separate tests. It would be fairly easy to add a `testname` flag which would be the name of the test to run (in a suitable form to pass to `go test -run`). Splitting the tests would have the advantage that they can be run in parallel too.

I'm not sure if I understand what you mean. I thought they were already 3 separate tests?

https://github.com/rclone/rclone/blob/30517698aa1f410314a0512ac1e2881044375b8c/cmd/bisync/bisync_test.go#L225-L257

> You can just skip the test there if you want - you won't get any different behavior to `TestSFTPRclone`.

Sounds good.

nielash:
@ncw this fixes `TestSFTPRclone:` for me, but I'm not sure if it's the right fix. Thoughts? https://github.com/nielash/rclone/commit/993afa9e7cddaf60eb913a5e83e14ac2f4600f20

ncw:
> > How about splitting the bisync tests into 3 separate tests. It would be fairly easy to add a `testname` flag which would be the name of the test to run (in a suitable form to pass to `go test -run`). Splitting the tests would have the advantage that they can be run in parallel too.
> 
> I'm not sure if I understand what you mean. I thought they were already 3 separate tests?

Yes they are. I meant split them up in config.yaml something like this

```yaml
 - path: cmd/bisync
   testname: TestRemoteLocal
 - path: cmd/bisync
   testname: TestLocalRemote
 - path: cmd/bisync
   testname: TestRemoteRemote
```

Not sure that is a great idea but there you go!

> @ncw this fixes `TestSFTPRclone:` for me, but I'm not sure if it's the right fix. Thoughts? [nielash@993afa9](https://github.com/nielash/rclone/commit/993afa9e7cddaf60eb913a5e83e14ac2f4600f20)

I'm not sure either of those fixes are correct. I suspect the UTF-8 fix is only needed for macOS - can you try running the `rclone serve sftp` it on a Linux VM?

Why do you think the `--dir-cache` is helpful?

Confused :-)

nielash:
> I'm not sure either of those fixes are correct. I suspect the UTF-8 fix is only needed for macOS - can you try running the `rclone serve sftp` it on a Linux VM?

I'll try it on a linux machine and see what happens 🙂 


> Why do you think the `--dir-cache` is helpful?

My working theory is that `d.Rename` fails here with `file does not exist`:

https://github.com/rclone/rclone/blob/29ed17d19cb48ed1b51da649001176e7e9698e69/vfs/dir.go#L1088-L1092

because `d._readDir` does not consider it "stale" and therefore skips checking it here:

https://github.com/rclone/rclone/blob/29ed17d19cb48ed1b51da649001176e7e9698e69/vfs/dir.go#L500-L508
and this is because `d.vfs.Opt.DirCacheTime` here:

https://github.com/rclone/rclone/blob/29ed17d19cb48ed1b51da649001176e7e9698e69/vfs/dir.go#L360

is 5 minutes by default.

This might be wrong, but if so then I can't explain why the `ext_paths` tests definitely pass with the change and fail without it. 🤔

nielash:
> I'll try it on a linux machine and see what happens 🙂

@ncw I've now tested it on Ubuntu 23.10 and the behavior is the same. Tests pass with the fixes, fail when either one is reverted.

nielash:
I am also finding that `--dir-cache-time 0` similarly fixes one of the `TestS3Rclone:` failures.

nielash:
@ncw Here are some notes and my guesses about some of the other failures -- mainly posting this in case anything rings a bell to you, as I am not very familiar with the oddities of some of these backends.

- `TestAzureFiles:` looks like something related to hashes / hash types. It should be finding a hash difference and it's not (maybe they are blank?)
- `TestBox:` seems to be some issue with copying a file over itself. 
```
ERROR : file1.txt: Failed to copy: Error "item_name_in_use" (409): Item with the same name already exists ({"conflicts":{"type":"file","id":"1509373873360","file_version":{"type":"file_version","id":"1657286950960","{hashtype}":"da39a3ee5e6b4b0d3255bfef95601890afd80709"},"sequence_id":"0","etag":"0","{hashtype}":"da39a3ee5e6b4b0d3255bfef95601890afd80709","name":"file1.txt"}})
```
- `TestCache:` looks like an encoding issue when a directory name has control chars.
- `TestChunkerChunk3bNometaLocal:` error is `fatal error: runtime.SetFinalizer: finalizer already set` which makes me think it's a bug related to `cache.PinUntilFinalized`
- `TestChunkerChunk50bMD5QuickS3:` this is the `file name too long` issue I thought I fixed like 3 times already 🤦‍♂️ The long name in question is:
```
TestChunkerChunk50bMD5QuickS3_rclone-test-tefunaw2lolanoc6sasofor8_050050qu_ext_paths_path1_測試Русский__ěáñ..TestChunkerChunk50bMD5QuickS3_rclone-test-tefunaw2lolanoc6sasofor8_050050qu_ext_paths_path2_測試Русский__ěáñ.path2.lst-new
``` 
How would you feel about abbreviating `TestChunkerChunk50bMD5QuickS3` in some way? Alternatively, would it mess things up greatly if in [`RandomRemoteName`](https://github.com/nielash/rclone/blob/9061e8185054a713f0c218b0e33e64993f19a514/fstest/fstest.go#L439) we changed `24` to a lower number? (I realize we'd have to make the same change [here](https://github.com/nielash/rclone/blob/9061e8185054a713f0c218b0e33e64993f19a514/fstest/fstest.go#L50).) 
- `TestChunkerChunk50bMailru:` this is failing because of #7768. Is there a clean way to set `--tpslimit 10` as the default for the backend? If not, I think I can set it on the tests via the global config -- but seems like doing it in the backend would be better.
- `TestChunkerChunk50bSHA1HashS3:` same `file name too long` error described above
- `TestChunkerChunk50bSHA1QuickS3:` same `file name too long` error described above
- `TestChunkerChunk50bYandex:` seems like a `yandex` issue relating to copying a file over itself.
```
ERROR : file1.txt: Failed to copy: couldn't copy file: [409 - DiskResourceAlreadyExistsError] Resource "disk:/rclone-test-lebikon0foyesex0womeyod4/073716le/basic/path1/file1.txt" already exists. (Ресурс "disk:/rclone-test-lebikon0foyesex0womeyod4/073716le/basic/path1/file1.txt" уже существует.)
```
- `TestChunkerMailru:` same #7768 issue mentioned above
- `TestChunkerNometaLocal:` same `finalizer already set` issue mentioned above. Oddly, I'm not able to repro it locally, which makes me wonder if it's because the real tests run more slowly and therefore end up running into [this unfixed bug](https://github.com/rclone/rclone/pull/7668#issuecomment-1995473924).
- `TestGoogleCloudStorage,directory_markers:` we are hitting `rateLimitExceeded`, and also there seems to be an issue like the one I found in `:memory:` with list entries having paths that are too short:
```
ERROR : subdir_rawchars_␙_�_�: Entry doesn't belong in directory "subdir_rawchars_␙_\x81_\xfe" (too short) - ignoring
```
This only happens with `directory_markers`.
- `TestGoogleCloudStorage:` It gets very closing to finishing but then hits `rateLimitExceeded`. What is the proper solution here?
- `TestJottacloud:` seems like unicode like `folder/éééö.txt` is getting auto-normalized in a way we are not currently looking for.
- `TestLinkbox:` seems to be a filename limitation:
```
subdir/.chk_file: Failed to copy: Update second step failed: Linkbox error 500: The file name needs to include a suffix, such as xxx.mp4
```
- `TestMailru:` same #7768 issue mentioned above
- `TestMega:` seems like a bug with `o.Remove` not actually removing the object, but not returning an error. That `modtime_write_test` file shouldn't still be there by the time the real test starts.
- `TestnStorage:` looks like we just need to ignore the `Implicitly create directory` logs -- should be an easy fix.
- `TestOracleObjectStorage:` seems like either the backend can't handle names with whitespace / control chars or we are not encoding them properly in the API call.
```
Message: Failed to parse json from response body due to: unexpected end of JSON input. With response body .
```
- `TestPikPak:` seems to be unable to open files for hashing
```
failed to open file file1.txt: can't download - no link to download
```
- `TestPutio:` something related to renaming a file:
```
ERROR : file1.txt: Failed to copy: unexpected status code (400) response while doing POST to https://api.put.io/v2/files/rename
```
- `TestQuatrix:` looks like the fs `.String()` is not including the subpath, same issue as I fixed in `hdfs`. Should be an easy fix.
- `TestS3,directory_markers:` likely the same issue mentioned above with `TestGoogleCloudStorage,directory_markers:` where the bucket directory_markers don't always return the paths correctly:
```
ERROR : subdir with␊white space.txt: Entry doesn't belong in directory "subdir with␊white space.txt" (too short) - ignoring
ERROR : subdir_rawchars_␙_�_�: Entry doesn't belong in directory "subdir_rawchars_␙_\x81_\xfe" (too short) - ignoring
```
- `TestS3Rclone:` one of the failures is resolved by `--dir-cache-time 0` (see discussion above). The other seems to be something odd related to modtimes returning `time.Now()` instead of `2000-01-01 00:00:00 +0000 UTC`. I am guessing there's a bug in `serve s3`, although there's a small possibility the problem is in `vfs`. Any ideas?
- `TestSFTPOpenssh:` seems to be something related to unicode encoding, or possibly modtimes not always updating as they should.
- `TestSFTPRclone:` see discussion [here](https://github.com/rclone/rclone/pull/7743#issuecomment-2070914934) about potential fix
- `TestSFTPRsyncNet:` looks like maybe we are not properly terminal-escaping newlines when we try to calculate md5s:
```
ERROR : file2 with
white space.txt.kotizop8.partial: Failed to calculate dst hash: failed to calculate md5 hash: failed to run "md5 -r /data2/home/59018/rclone-test-jiyegog4voqinat7lifojaj7/043414fi/extended_filenames/path2/subdir\\ with'\n'white\\ space.txt/file2\\ with'\n'white\\ space.txt.kotizop8.partial": md5: /data2/home/59018/rclone-test-jiyegog4voqinat7lifojaj7/043414fi/extended_filenames/path2/subdir with: No such file or directory: Process exited with status 1
```
- `TestStorj:` it seems like there's a bug where the bucket needs to be explicitly `Mkdir`'d in a way that other backends don't need. I've noticed this before. 
```
storj: bucket: uplink: bucket not found ("rclone-test-mepoqax3vuyuris8velinag1")
```
- `TestUlozto:` getting lots of `failed to make directory: Error 60001 (422): Validation failed.` and I'm not sure exactly what that means. I know it's a relatively new backend and lots of other tests are failing too, so I'm inclined to say the problem isn't bisync.
- `TestWebdavNextcloud:` Looks like the initial problem is related to hashes being unexpectedly blank after server-side copy/move. And then some of the retries fail because the container name is already in use (maybe it's not properly stopped after each try?)
- `TestWebdavOwncloud:` same blank hash issue as nextcloud. (although interestingly `TestBisyncRemoteRemote/rclone_args` passes for owncloud and not for nextcloud)
- `TestYandex:` same issue as `TestChunkerChunk50bYandex:` relating to copying a file over itself.

For what it's worth, it seems like very few (if any) of these are actually bisync problems -- by and large they are backend-specific issues or issues with how the tests are designed.

I'll try to dig in to these further, but if you have any hunches that might save me some time, I am all ears 🙂

PR #7751, opened 2024-04-10
Add go mod and sum to gitattributes for consistent line endings

albertony:

Not at all important, but it avoids some git noise.

Go consistently uses `LF` when writing `go.mod` and `go.sum` files. Git on Windows by default transforms text files to `CRLF`. This means git clone writes `go.mod` and `go.sum` as `CRLF`, if updating them with e.g. `go mod tidy` go will rewrite them with `LF`, and then git will show them as different and warn if adding them to staging:

```
warning: in the working copy of 'go.mod', LF will be replaced by CRLF the next time Git touches it
warning: in the working copy of 'go.sum', LF will be replaced by CRLF the next time Git touches it
```

#### Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
+# Go writes go.mod and go.sum with lf even on windows
+go.mod text eol=lf
+go.sum text eol=lf
+
 # Ignore generated files in GitHub language statistics and diffs
 /MANUAL.* linguist-generated=true
 /rclone.1 linguist-generated=true


ncw:
This certainly seems safe since users don't normally edit go.mod and go.sum files by hand.

I had no idea that git checks out things with CRLF endings on Windows though - I thought we'd moved on from that years ago. My immediate thought is we should have LF endings everywhere but maybe that will break stuff in Windows? It will certainly break peoples existing checkouts. If I had known about this setting a decade ago I would have set it so Windows get the exact same files as everyone else!

If we apply this then what will happen to Windows git users when they pull it? I imagine if they haven't modified go.sum and go.mod it will update properly, but what if they have? It will cause a conflict won't it?

I looked up [some docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) and it was suggesting that we mark the binary files as binary. We don't have a lot of binary files though and perhaps git is dealing with them already - how can we tell?

ncw (approved):
I did a bit of exploration with

    git ls-files --eol | grep -E '^(i/-text)'

To show binary files and with `-v` to show ascii files and I think git is doing the right thing.

I think also that this won't cause a problem so I think we should merge it - please do so when happy!

albertony:
Found two small anomalies that I suggest cleaning up here: https://github.com/rclone/rclone/pull/7775

PS: I'm slowly starting to disagree with myself... 😕 

> I certainly prefer keeping the native Windows line endings on regular source files myself,

I re-discovered that golangci-lint (the gofmt/goimports linters) assumes source files have `lf`, and will complain with its notorious "File is not goimports-ed (goimports)" otherwise - e.g. if trying to run it locally on Windows with workdir cloned using git for Windows with default settings (`core.autocrlf=true`). However still not convinced `lf` should be forced on everyone by the repo, I'm still thinking this is something that should be up to the individual developers/setups to decide how they want it.

ncw:
Well you know my opinion which was forged over many years of fighting with cross platform code on windows!

But I don't want to cause a problem for windows Devs.

PR #7754, opened 2024-04-10
cmd/gitannex: Add layout modes for compatibility

dmcardle:

This change adds support for the repo layouts defined by  [git-annex-remote-rclone](https://github.com/git-annex-remote-rclone/git-annex-remote-rclone). This should enable git-annex users with external remotes of type "rclone" to switch to "rclone-builtin" without needing to retransfer content. It also adds e2e compatibility tests in an attempt to verify this claim!

Issue #7625 

#### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ ifdef GOTAGS
 BUILDTAGS=-tags "$(GOTAGS)"
 LINTTAGS=--build-tags "$(GOTAGS)"
 endif
+LDFLAGS=--ldflags "-s -X github.com/rclone/rclone/fs.Version=$(TAG)"
 
 .PHONY: rclone test_all vars version
 
@@ -50,7 +51,7 @@ rclone:
 ifeq ($(GO_OS),windows)
 	go run bin/resource_windows.go -version $(TAG) -syso resource_windows_`go env GOARCH`.syso
 endif
-	go build -v --ldflags "-s -X github.com/rclone/rclone/fs.Version=$(TAG)" $(BUILDTAGS) $(BUILD_ARGS)
+	go build -v $(LDFLAGS) $(BUILDTAGS) $(BUILD_ARGS)
 ifeq ($(GO_OS),windows)
 	rm resource_windows_`go env GOARCH`.syso
 endif
@@ -59,7 +60,7 @@ endif
 	mv -v `go env GOPATH`/bin/rclone`go env GOEXE`.new `go env GOPATH`/bin/rclone`go env GOEXE`
 
 test_all:
-	go install --ldflags "-s -X github.com/rclone/rclone/fs.Version=$(TAG)" $(BUILDTAGS) $(BUILD_ARGS) github.com/rclone/rclone/fstest/test_all
+	go install $(LDFLAGS) $(BUILDTAGS) $(BUILD_ARGS) github.com/rclone/rclone/fstest/test_all
 
 vars:
 	@echo SHELL="'$(SHELL)'"
@@ -87,13 +88,13 @@ test:	rclone test_all
 
 # Quick test
 quicktest:
-	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) ./...
+	RCLONE_CONFIG="/notfound" go test $(LDFLAGS) $(BUILDTAGS) ./...
 
 racequicktest:
-	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -cpu=2 -race ./...
+	RCLONE_CONFIG="/notfound" go test $(LDFLAGS) $(BUILDTAGS) -cpu=2 -race ./...
 
 compiletest:
-	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) -run XXX ./...
+	RCLONE_CONFIG="/notfound" go test $(LDFLAGS) $(BUILDTAGS) -run XXX ./...
 
 # Do source code quality checks
 check:	rclone
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,7 +124,7 @@ jobs:
           sudo modprobe fuse
           sudo chmod 666 /dev/fuse
           sudo chown root:$USER /etc/fuse.conf
-          sudo apt-get install fuse3 libfuse-dev rpm pkg-config git-annex
+          sudo apt-get install fuse3 libfuse-dev rpm pkg-config git-annex git-annex-remote-rclone
         if: matrix.os == 'ubuntu-latest'
 
       - name: Install Libraries on macOS
@@ -137,7 +137,7 @@ jobs:
           brew untap --force homebrew/cask
           brew update
           brew install --cask macfuse
-          brew install git-annex
+          brew install git-annex git-annex-remote-rclone
         if: matrix.os == 'macos-latest'
 
       - name: Install Libraries on Windows
--- a/cmd/gitannex/e2e_test.go
+++ b/cmd/gitannex/e2e_test.go
@@ -225,13 +225,6 @@ func TestEndToEnd(t *testing.T) {
 		t.Run(string(mode), func(t *testing.T) {
 			t.Parallel()
 
-			// Create a temp directory and chdir there, just in case.
-			originalWd, err := os.Getwd()
-			require.NoError(t, err)
-			tempDir := t.TempDir()
-			require.NoError(t, os.Chdir(tempDir))
-			t.Cleanup(func() { require.NoError(t, os.Chdir(originalWd)) })
-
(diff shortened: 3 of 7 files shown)


dmcardle:
@ncw, PTAL!

FYI: @yarikoptic

dmcardle:
Wow, the for loop thing bit me again. Fixed in the latest push. https://go.dev/blog/loopvar-preview

yarikoptic:
> This should enable git-annex users with external remotes of type "rclone" to switch to "rclone-builtin" without needing to retransfer content.

I haven't had a chance to try yet, but I just want to say **Thank you**.  I wonder if there could be e2e test run where first repo is initialized to use `git-annex-remote-rclone` and then migrated to use `git-annex-remote-rclone-builtin` and ensured to still operate correctly -- can get a previously uploaded file and then `annex testremote`.  That would be very much appreciated before I try to migrate thousands of repos (https://github.com/dandisets and https://github.com/dandizarrs) covering 100s of TBs of data already in dropbox ;)

dmcardle:
> I haven't had a chance to try yet, but I just want to say **Thank you**. 

You're very welcome! I'm just working on this for fun, so it's a pleasant surprise that someone finds it useful!

> I wonder if there could be e2e test run where first repo is initialized to use `git-annex-remote-rclone` and then migrated to use `git-annex-remote-rclone-builtin` and ensured to still operate correctly -- can get a previously uploaded file and then `annex testremote`.

In case you didn't spot it, the new `TestEndToEndRepoLayoutCompat` test in `e2e_test.go` is doing something kind of like this. It creates two remotes  backed by the same local rclone remote, one `externaltype=rclone` and one `externaltype=rclone-builtin`. It copies files to the first remote and makes sure they're present on the second, etc.

You raise a good point, though. I think it's worth adding a migration test that runs `git annex testremote` after changing the externaltype of an existing remote. I think we can change externaltype with `git annex enableremote`, but I haven't tried it yet. This shouldn't be too hard, I'll try to squeeze it into this PR.

> That would be very much appreciated before I try to migrate thousands of repos (https://github.com/dandisets and https://github.com/dandizarrs) covering 100s of TBs of data already in dropbox ;)

Looks like nearly an exabyte! :exploding_head:

dmcardle:
@yarikoptic PTAL at the latest commit, which adds TestEndToEndMigration tests.

yarikoptic:
> test in `e2e_test.go` is doing something kind of like this. It creates two remotes backed by the same local rclone remote, one `externaltype=rclone`

that is great - thank you!  I have only vague memory on how I decided that there is none -- I remember searching for `git-annex-remote-rclone` and finding none -- but I see it all over now, so I guess my "search foo" was weak at that point.
But I still do not see any hit for installation of `git-annex-remote-rclone` in `.github/workflows`:
```
❯ git grep git-annex-remote-rclone -- .github/ || echo empty
empty
```

aren't e2e tests ran on CI? if so -- should get `git-annex-remote-rclone` installed to exercise testing you introduced.

dmcardle:
> aren't e2e tests ran on CI? if so -- should get `git-annex-remote-rclone` installed to exercise testing you introduced.

Ah! Great catch! I'll update this PR to install git-annex-remote-rclone on Linux and macOS. (Still punting on Windows.)

dmcardle:
@ncw PTAL :)

Also, is there a way to tell which tests were skipped on CI? What do you think about changing CI to run `go test -v`? It would produce a lot more output, at least for these new tests that run `git annex testremote`, but we'd still get a concise summary at the end.

The reason I ask is that I mistakenly believed these new tests were running on CI because I  saw something like this in the log, but it turned out that they were certainly being skipped due to the missing `git-annex-remote-rclone` binary.

```
ok      github.com/rclone/rclone/cmd/gitannex   0.266s
```

dmcardle:
Hey @ncw, looks like CI is skipping the gitannex end-to-end tests because the `rclone` binary on PATH has an unexpected version string. For example, from the [linux job's log](https://github.com/rclone/rclone/actions/runs/8721904020/job/23926626461?pr=7754#step:12:12846):

```
=== RUN   TestEndToEnd
    e2e_test.go:199: Skipping due to rclone version: expected version "v1.67.0-DEV", but got "v1.67.0-beta.7905.220bbe24d.merge"
--- SKIP: TestEndToEnd (0.07s)
=== RUN   TestEndToEndMigration
    e2e_test.go:199: Skipping due to rclone version: expected version "v1.67.0-DEV", but got "v1.67.0-beta.7905.220bbe24d.merge"
--- SKIP: TestEndToEndMigration (0.06s)
=== RUN   TestEndToEndRepoLayoutCompat
    e2e_test.go:199: Skipping due to rclone version: expected version "v1.67.0-DEV", but got "v1.67.0-beta.7905.220bbe24d.merge"
--- SKIP: TestEndToEndRepoLayoutCompat (0.07s)
```

I have been assuming that the value of `fs.Version` seen by tests would match the version in the newly-built `rclone` binary, but that's clearly not the case!

It looks like the tests construct "v.1.67.0-DEV" based on this logic: https://github.com/rclone/rclone/blob/30517698aa1f410314a0512ac1e2881044375b8c/fs/version.go#L8-L12

Meanwhile, the Makefile just entirely clobbers `fs.Version`:
https://github.com/rclone/rclone/blob/30517698aa1f410314a0512ac1e2881044375b8c/Makefile#L53

The latest commit changes the Makefile to clobber the version tag for tests, too. Let's see if that does the trick :)

Makefile
```diff
@@ -87,13 +88,13 @@ test:	rclone test_all
 
 # Quick test
 quicktest:
-	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) ./...
+	RCLONE_CONFIG="/notfound" go test -v $(LDFLAGS) $(BUILDTAGS) ./cmd/gitannex/...
```
dmcardle, line 91: Whoops, I didn't mean to restrict tests to `./cmd/gitannex/...`. I'll fix it next time I work on this PR.

dmcardle:
Interesting, the tests seem to run on Linux now, but mac_amd64 now skips the tests for a different reason:

```
2024-04-17T13:12:28.1429840Z === RUN   TestEndToEnd
2024-04-17T13:12:28.2422770Z     e2e_test.go:200: Skipping due to rclone version: expected tag string "none", but got "cmount"
2024-04-17T13:12:28.2424190Z --- SKIP: TestEndToEnd (0.10s)
2024-04-17T13:12:28.2426030Z === RUN   TestEndToEndMigration
```

dmcardle:
> Interesting, the tests seem to run on Linux now, but mac_amd64 now skips the tests for a different reason:
> 
> ```
> 2024-04-17T13:12:28.1429840Z === RUN   TestEndToEnd
> 2024-04-17T13:12:28.2422770Z     e2e_test.go:200: Skipping due to rclone version: expected tag string "none", but got "cmount"
> 2024-04-17T13:12:28.2424190Z --- SKIP: TestEndToEnd (0.10s)
> 2024-04-17T13:12:28.2426030Z === RUN   TestEndToEndMigration
> ```

Aha! I think what's happening is that CI is running something like `GOTAGS=cmount make rclone` and then `make quicktest` without any tags. The `rclone` binary thus tells us it was compiled with "cmount" tags, but the tests expect "none". @ncw is this behavior surprising to you or WAI?

As a workaround, I'll just log a warning rather than skipping e2e tests when the tags do not match.

dmcardle:
@ncw This is ready for review, more or less!

I'm still stumped by the GoTags not matching, but my workaround is to just run the e2e tests anyway.

Another questionable change in this PR is that I changed `go test` commands to `go test -v`. Totally happy to remove it, but it was necessary for me to figure out whether the gitannex e2e tests were actually running in CI.

dmcardle:
@ncw Sorry to nag, but PTAL!

Makefile
```diff
@@ -87,13 +87,13 @@ test:	rclone test_all
 
 # Quick test
 quicktest:
-	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) ./...
+	RCLONE_CONFIG="/notfound" go test -v $(BUILDTAGS) ./...
```
ncw: I'm not in favour of adding the `-v` here. Rclone has a lot of tests and having to look through the `-v` output in the CI is not fun! (279 lines vs 68057) So I'm not in favour of making it longer.

ncw (commented):
I had a look through that - all looking very nice.

Can you drop the commit adding `-v` to `make quicktest` and I'll merge what we've got - thank you :-)

Makefile
dmcardle: That's totally fair! It's not too hard to temporarily add the `-v` as needed. Done.

Issue #7745, opened 2024-04-09
Documentation: rclone_genautocomplete_bash.md example command looks incorrect for SLES/openSUSE/Rocky?


pawsey-kbuckley:
The `rclone_genautocomplete_bash.md` file renders, within this GitHub repo, as

"Logout and login again to use the autocompletion scripts, or source them directly

`. /etc/bash_completion`

If you supply a command line argument ..."

however, if you are sourcing the script (**singular**) that that doc, a couple of lines
above, tell you gets created by default, **directly**, then should the example
not be telling you to do this

`. /etc/bash_completion.d/rclone`

instead?

I am thinking that the documentation's choice, of sourcing a utility 
file `/etc/bash_completion` that, on **Ubuntu**, actually sources this

`/usr/share/bash-completion/bash_completion`

might cause some confusion, as it did, for me. on a couple of  **SLES**,
 **openSUSE** and **Rocky 9** systems we run here, where

`/etc/bash_completion`

doesn't exist, even though the

`/usr/share/bash-completion/bash_completion`

file does.

If one does, as I did for testing without root/sudo, create a single
automcompletion script, then the sourcing of it would want the full
path to the script specified anyway.

Maybe then, the text and example should read, as follows, where the
script is singular:

"Logout and login again to use the autocompletion script, or source it directly

`. /etc/bash_completion.d/rclone`

If you supply a command line argument ..."

ncw:
Thanks for spotting this.

The source code for the docs is here - this is the only place which needs changing.

https://github.com/rclone/rclone/blob/master/cmd/genautocomplete/genautocomplete_bash.go

Do you fancy editing it to make a correction? You can do it in the web browser by clicking the pencil icon on that page - it will then make an edit for us to review. (Or you could send a Pull Request if you know how and would prefer).

PR #7755, opened 2024-04-11
Remove Ubuntu-ism from text: clarify non-root use

pawsey-kbuckley:
As requested, here is a PR containing some changes that present
a fuller functionality of the `genautocomplete_bash` via its docs.

It's a bit more of a re-write than the changes that the issue I had
raised originally suggested would be needed, but I think I have
"covered all the bases" now.

### DIFF
--- a/cmd/genautocomplete/genautocomplete_bash.go
+++ b/cmd/genautocomplete/genautocomplete_bash.go
@@ -18,20 +18,31 @@ var bashCommandDefinition = &cobra.Command{
 	Long: `
 Generates a bash shell autocompletion script for rclone.
 
-This writes to /etc/bash_completion.d/rclone by default so will
-probably need to be run with sudo or as root, e.g.
+By default, when run without any arguments, 
 
-    sudo rclone genautocomplete bash
+    rclone genautocomplete bash
 
-Logout and login again to use the autocompletion scripts, or source
-them directly
+the generated script will be written to
 
-    . /etc/bash_completion
+/etc/bash_completion.d/rclone
 
-If you supply a command line argument the script will be written
-there.
+and so will probably need to be run as root, or with sudo.
+
+If you supply a path to a file as the command line argument, theh 
+the generated script will be written to that file, in which case
+you should not need root privileges.
 
 If output_file is "-", then the output will be written to stdout.
+
+If you have installed the script into the default location, you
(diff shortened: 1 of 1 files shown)


pawsey-kbuckley:
Oops,

forgot to explicitly mention the issue that this relates to:  https://github.com/rclone/rclone/issues/7745

cmd/genautocomplete/genautocomplete_bash.go
```diff
@@ -18,20 +18,31 @@ var bashCommandDefinition = &cobra.Command{
...
 
-    sudo rclone genautocomplete bash
+    rclone genautocomplete bash
 
-Logout and login again to use the autocompletion scripts, or source
-them directly
+the generated script will be written to
 
-    . /etc/bash_completion
+/etc/bash_completion.d/rclone
```
ncw, line 27: This should probably have 4 spaces at the start so it is rendered as code

cmd/genautocomplete/genautocomplete_bash.go
```diff
@@ -18,20 +18,31 @@ var bashCommandDefinition = &cobra.Command{
...
 
-If you supply a command line argument the script will be written
-there.
+and so will probably need to be run as root, or with sudo.
```
ncw, line 29: I think this needs a little more clarification, maybe

    ...and so rclone will probably need to be run as root...

As I wasn't sure whether it was the generated script or rclone that needed running as root reading that.

cmd/genautocomplete/genautocomplete_bash.go
```diff
@@ -18,20 +18,31 @@ var bashCommandDefinition = &cobra.Command{
...
+
+If you supply a path to a file as the command line argument, theh
```
ncw, line 31: Typo here `theh`

ncw (commented):

I put some minor niggles inline.

Note that if you want this to refer and close the issue then put "Fixes #7745" at the end of the commit message. I can do this when I merge it also if you prefer.

pawsey-kbuckley:
Have fixed up you concerns/notes in my working  copy and pushed,
so my branch should now be ready to merge.

I'll leave the workfllow Fixes addition to you, though.

ncw (approved):
Thank you :-)

Issue #7748, opened 2024-04-09
s3 serve: Improve documentation: instructions unclear


static-moonlight:
Currently, I am looking at rclone to either act as FTP or S3 server (I'm aware it's still experimental). The [documentation](https://rclone.org/commands/rclone_serve_s3/) describes this line for example:

```bash
rclone serve s3 --auth-key ACCESS_KEY_ID,SECRET_ACCESS_KEY remote:path
```

I am running it directly from the container, so for me it looks like this:

```bash
podman run --rm docker.io/rclone/rclone:1.66 serve s3 --auth-key ACCESS_KEY_ID,SECRET_ACCESS_KEY remote:path
```

However, I have no idea how to make this run:

```
[...] NOTICE: Config file "/config/rclone/rclone.conf" not found - using defaults
[...] Failed to create file system for "remote:path": didn't find section in config file
```

Do I need to substitute `remote:path` or `remote` or `path` with something? Do I need to provide an `rclone.conf` with some specific content? If yes, is there an example somewhere? I could not find anything. As far as I understand the documentation, the example config is for the client side? Or not?

I'd love to see the documentation improved to make it more clear how to use this.

albertony:
> As far as I understand the documentation, the example config is for the client side?

I think you are correct.

> podman run --rm docker.io/rclone/rclone:1.66 serve s3 --auth-key ACCESS_KEY_ID,SECRET_ACCESS_KEY remote:path

You need to supply the path that you want to serve, using syntax from https://rclone.org/docs/#syntax-of-remote-paths. If you want to serve something from local disk, then specify path to it:

```
podman run --rm docker.io/rclone/rclone:1.66 serve s3 --auth-key ACCESS_KEY_ID,SECRET_ACCESS_KEY /path/to/serve
```

ncw:
It might be an idea to have both examples as `/path/to/wherever` shows it works for local files but `remote:path` shows it works for remote things which is kind of rclone's special sauce.

@static-moonlight The documentation is here https://github.com/rclone/rclone/blob/master/cmd/serve/s3/serve_s3.md - you can click the pencil icon to suggest a change.

PR #7761, opened 2024-04-13
doc: add example how to run serve s3

static-moonlight:
Suggestion to add more information to the documentation about serve s3.
See also: #7748


ncw (approved):
Looks great - thank you :-)

Issue #7760, opened 2024-04-13
Files on Windows vfs lost when file names contain CJK punctuations


CyberPuffer:
#### What is the problem you are having with rclone?

As an Asian user, it's very common that we use CJK punctuations in file names.
And since rclone has a Windows restricted characters replacement mechanism by default, it bugs our Asian users a lot.
If you use both Windows and Linux devices, the file name is inconsistent between two platforms.
So I try to achieve consistent file names by disabling this mechanism completely using `--local-encoding None` `--s3-encoding None`
But it seams that vfs on Windows doesn't honour `--local-encoding` arguments and drop files

Reproduce steps:
1. Mount a volume on Windows with `--local-encoding None` and `--s3-encoding None`
2. Copy some files that name contain CJK punctuations which are replacements of Windows restricted characters
    eg: `foo：bar.txt`
3. Unmount volume and remount volume
4. Uploaded files lost

#### What is your rclone version (output from `rclone version`)

v1.66.0

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Windows 11, 64bit

#### Which cloud storage system are you using? (e.g. Google Drive)

S3 Compliant Storage (Cloudflare R2)

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
%ProgramData%\Programs\rclone\rclone.exe ^
    -vv ^
    --use-mmap ^
    --local-encoding None ^
    --s3-encoding None ^
    --s3-directory-markers ^
    --use-server-modtime ^
    --update ^
  mount Private: * ^
    --network-mode ^
    --vfs-cache-mode=writes ^
    --vfs-fast-fingerprint ^
    --vfs-read-chunk-size 5M ^
    --vfs-read-chunk-size-limit 80M ^
    --vfs-disk-space-total-size 1T ^
    --vfs-cache-max-age 1m ^
    --vfs-used-is-size ^
    --volname myCloud
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
[...]

Issue #7456, opened 2023-11-26
The design of encoding for restricted characters is counter-intuitive


URenko:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/cant-access-file-with-in-the-name/43068

#### What is the problem you are having with rclone?
Can't access file with `‛` ( U+201B, SINGLE HIGH-REVERSED-9 QUOTATION MARK, the character used to escape restricted characters) in the name.

#### What is your rclone version (output from `rclone version`)

```
rclone v1.64.2
- os/version: debian 12.2 (64 bit)
- os/kernel: 6.1.0-13-amd64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.3
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Linux, 64bit

#### Which cloud storage system are you using? (e.g. Google Drive)

Local and OneDrive, et.al. Here I take local as the example.

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
echo 'here is the content' > '‛'
rclone cat ‛
rclone cat ‛‛
rclone cat ‛‛‛
rclone cat ‛‛‛‛
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
rclone cat ‛ -vv
<7>DEBUG : rclone: Version "v1.64.2" starting with parameters ["rclone" "cat" "‛" "-vv"]
<7>DEBUG : rclone: systemd logging support activated
<7>DEBUG : Creating backend with remote "‛"
<7>DEBUG : Using config file from "/home/<myusername>/.config/rclone/rclone.conf"
<7>DEBUG : fs cache: renaming cache item "‛" to be canonical "/tmp/test/‛‛"
<3>ERROR : : error listing: directory not found
<7>DEBUG : 4 go routines active
Failed to cat with 2 errors: last error was: directory not found
```

---

After reading the code, I understand that the story is like this:

* For human input, rclone treated it as already be encoded as a way called "Standard", which means `EncodeZero | EncodeSlash | EncodeCtl | EncodeDel | EncodeDot`
* Then rclone will decode it and encode back with the encoding of this backend （`FromStandardPath`). And use the encoded path to access the backend.
[...]

Issue #6098, opened 2022-04-11
Rclone failed to process files whose filename or path contains "‛"


wegood9:
#### What is your rclone version (output from `rclone version`)
rclone v1.58.0

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
debian 11.2 (64 bit)

#### Which cloud storage system are you using? (e.g. Google Drive)
SFTP and local filesystem

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
`rclone copy /tmp/test /tmp/test2`
rclone copy /tmp/test remote:test`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
2022/04/11 12:14:16 DEBUG : rclone: Version "v1.58.0" starting with parameters ["rclone" "copy" "/tmp/test" "/tmp/test2" "-vvv"]
2022/04/11 12:14:16 DEBUG : Creating backend with remote "/tmp/test"
2022/04/11 12:14:16 DEBUG : Using config file from "/home/username/.config/rclone/rclone.conf"
2022/04/11 12:14:16 DEBUG : Creating backend with remote "/tmp/test2"
2022/04/11 12:14:16 DEBUG : Local file system at /tmp/test2: Waiting for checks to finish
2022/04/11 12:14:16 DEBUG : Local file system at /tmp/test2: Waiting for transfers to finish
2022/04/11 12:14:16 ERROR : test‛‛s: Failed to copy: failed to open source object: open /tmp/test/test‛‛s: no such file or directory
2022/04/11 12:14:16 DEBUG : tests: md5 = d41d8cd98f00b204e9800998ecf8427e OK
2022/04/11 12:14:16 INFO  : tests: Copied (new)
2022/04/11 12:14:16 ERROR : Attempt 1/3 failed with 1 errors and: failed to open source object: open /tmp/test/test‛‛s: no such file or directory
2022/04/11 12:14:16 ERROR : test‛‛s: Failed to copy: failed to open source object: open /tmp/test/test‛‛s: no such file or directory
2022/04/11 12:14:16 DEBUG : Local file system at /tmp/test2: Waiting for checks to finish
2022/04/11 12:14:16 DEBUG : tests: Size and modification time the same (differ by 0s, within tolerance 1ns)
2022/04/11 12:14:16 DEBUG : tests: Unchanged skipping
[...]

PR #7791, opened 2024-04-22
fix issues related to encoding

URenko:
Fix the way to disable encoding by introducing a new type `Raw` to solve #7456 and #6098.
See #7456 and the changes on the document for details.
The behavior of `None` was not changed to keep backward compatibility.

And fix vfs cache encoding described in #7760
The problem is when we create and write cache, we use the "OS" encoded ("OS" encoding is hardcoded for different operating systems, not specified by users).
But when we copy it to the remote (backend), we use local.Fs, which accept a name already encoded in "Standard" way.
So we need to send a "Standard" encoded name.

However, the "minimum setup" mentioned in #7760 has not been resolved. See my comment in it for details.
But it's an edge case.

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1568,32 +1568,47 @@ func (o *Object) SetMetadata(ctx context.Context, metadata fs.Metadata) error {
 }
 
 func cleanRootPath(s string, noUNC bool, enc encoder.MultiEncoder) string {
-	if runtime.GOOS != "windows" || !strings.HasPrefix(s, "\\") {
-		if !filepath.IsAbs(s) {
-			s2, err := filepath.Abs(s)
-			if err == nil {
-				s = s2
-			}
-		} else {
-			s = filepath.Clean(s)
-		}
-	}
+	var vol string
 	if runtime.GOOS == "windows" {
-		s = filepath.ToSlash(s)
-		vol := filepath.VolumeName(s)
+		vol = filepath.VolumeName(s)
 		if vol == `\\?` && len(s) >= 6 {
 			// `\\?\C:`
 			vol = s[:6]
 		}
-		s = vol + enc.FromStandardPath(s[len(vol):])
-		s = filepath.FromSlash(s)
-		if !noUNC {
-			// Convert to UNC
-			s = file.UNCPath(s)
-		}
-		return s
+		s = s[len(vol):]
 	}
-	s = enc.FromStandardPath(s)
+	// Don't use FromStandardPath. Make sure Dot (`.`, `..`) as name will not be reencoded
+	// Take care of the case Standard: ./．/‛． (the first dot means current directory)
+	if enc != encoder.Standard {
+		s = filepath.ToSlash(s)
+		parts := strings.Split(s, "/")
+		encoded := make([]string, len(parts))
+		changed := false
+		for i, p := range parts {
+			if (p == ".") || (p == "..") {
+				encoded[i] = p
+				continue
+			}
+			part := enc.FromStandardName(p)
+			changed = changed || part != p
+			encoded[i] = part
+		}
+		if changed {
+			s = strings.Join(encoded, "/")
+		}
+		s = filepath.FromSlash(s)
+	}
+	if runtime.GOOS == "windows" {
+		s = vol + s
+	}
+	s2, err := filepath.Abs(s)
+	if err == nil {
+		s = s2
+	}
+	if !noUNC {
+		// Convert to UNC. It does nothing on non windows platforms.
+		s = file.UNCPath(s)
+	}
 	return s
 }
 
--- a/docs/content/overview.md
+++ b/docs/content/overview.md
@@ -348,8 +348,8 @@ have a Windows file system with Unicode fullwidth characters
 remote rather than being translated to regular (halfwidth) `*`, `?` and `:`.
 
 The `--backend-encoding` flags allow you to change that. You can
-disable the encoding completely with `--backend-encoding None` or set
-`encoding = None` in the config file.
+disable the encoding completely with `--backend-encoding Raw` or set
+`encoding = Raw` in the config file.
 
 Encoding takes a comma separated list of encodings. You can see the
 list of all possible values by passing an invalid value to this
@@ -375,7 +375,7 @@ will show you the defaults for the backends.
 | LeftSpace | SPACE on the left of a string | `␠` |
 | LeftTilde | `~` on the left of a string | `～` |
 | LtGt | `<`, `>` | `＜`, `＞` |
-| None | No characters are encoded | |
+| None ¹ | NUL 0x00 | ␀ |
 | Percent | `%` | `％` |
 | Pipe | \| | `｜` |
 | Question | `?` | `？` |
@@ -386,6 +386,10 @@ will show you the defaults for the backends.
 | SingleQuote | `'` | `＇` |
 | Slash | `/` | `／` |
 | SquareBracket | `[`, `]` | `［`, `］` |
+
+¹ Encoding from NUL 0x00 to ␀ is always implicit except when using Raw.
+It was previously incorrectly documented as disabling encoding,
+and to maintain backward compatibility, its behavior has not been changed.
 
 ##### Encoding example: FTP
 
@@ -430,7 +434,7 @@ the default value but without `Colon,Question,Asterisk`:
 --local-encoding "Slash,LtGt,DoubleQuote,Pipe,BackSlash,Ctl,RightSpace,RightPeriod,InvalidUtf8,Dot"
 ```
 
-Alternatively, you can disable the conversion of any characters with `--local-encoding None`.
+Alternatively, you can disable the conversion of any characters with `--local-encoding Raw`.
 
 Instead of using command-line argument `--local-encoding`, you may also set it
 as [environment variable](/docs/#environment-variables) `RCLONE_LOCAL_ENCODING`,
--- a/lib/encoder/encoder.go
+++ b/lib/encoder/encoder.go
@@ -35,34 +35,35 @@ const (
 
 // Possible flags for the MultiEncoder
 const (
-	EncodeZero          MultiEncoder = 0         // NUL(0x00)
-	EncodeSlash         MultiEncoder = 1 << iota // /
-	EncodeLtGt                                   // <>
-	EncodeDoubleQuote                            // "
-	EncodeSingleQuote                            // '
-	EncodeBackQuote                              // `
-	EncodeDollar                                 // $
-	EncodeColon                                  // :
-	EncodeQuestion                               // ?
-	EncodeAsterisk                               // *
-	EncodePipe                                   // |
-	EncodeHash                                   // #
-	EncodePercent                                // %
-	EncodeBackSlash                              // \
-	EncodeCrLf                                   // CR(0x0D), LF(0x0A)
-	EncodeDel                                    // DEL(0x7F)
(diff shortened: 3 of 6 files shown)


ncw:
> The behavior of `None` was not changed to keep backward compatibility.

I think I'd prefer `None` to do what people expect and not do any encoding. What was your reasoning here?

URenko:
> > The behavior of `None` was not changed to keep backward compatibility.
> 
> I think I'd prefer `None` to do what people expect and not do any encoding. What was your reasoning here?

Backward compatibility.
Since encoding has been introduced for a long time. There must be many people already misused `None`.
If we change the behavior and they upgrade it without reading the changelog, we will break their storage. For example, they will need to resync tons of files (this is what happened to me when encoding was introduced. I don't what it to happen again to others : ).
And If they read the changelog, they will change to `Raw` and reorganize things.

URenko:
@ncw any progress or further concern? :-)

URenko:
fixed #7824
> Specifically, it causes statements like rclone copy <somewhere> . to spontaneously miss if . expands to a path with a Full Width replacement character.

iseahound:
I think you need to fix encoding once and for all. The main problem is that: `|` → `｜` → `‛｜` → `‛‛｜` → `‛‛‛｜` and so on where the encoding of one character as two characters instantly breaks the reversibility of the encoding. The example would be when a file system contains `‛｜`. Was it originally encoded as `｜`  or `‛｜`? (Note this can be formally stated as every surjection has a right inverse )

Here are some suggestions:

1. Add the names of the storage systems. For example, it'd be best to be able to do `--local-encoding=drive` or `--drive-encoding=Windows`.
2. Use a better escape character. Why not use the Unicode private use area as a rclone exclusive character, and prefix all escape sequences with that? `�‛｜`This preserves the current usage of repeating `‛‛‛‛‛‛`s. Admittedly, this does kick the can down the line, but anyone using private use characters is probably a super user of some sort and can read the documentation.
3. Document the replacement logic better. Currently as I understand it, the source and destination remotes both have a "stack of invalid characters". Should the two stacks be equal, no encoding is done. If the stacks are unequal, then encoding/decoding is done.

iseahound:
4. Also, instead of using `RAW` and having `None` as well, maybe adding `--destination-encoding=source` would be much clearer. If your destination remote is drive for example, having `--drive-encoding=source` would copy the files as-is since the two stacks of invalid character encodings cancel each other out.

URenko:
Hi @iseahound, thank you for your suggestions.

But first, let me clarify some of your misunderstandings.

> The main problem is that: `|` → `｜` → `‛｜` → `‛‛｜` → `‛‛‛｜` and so on where the encoding of one character as two characters instantly breaks the reversibility of the encoding. The example would be when a file system contains `‛｜`. Was it originally encoded as `｜` or `‛｜`? (Note this can be formally stated as every surjection has a right inverse )

The encoding of one character as multiple characters is inevitable.
For example, if you've been exposed to some programming languages, you'll know that the `\` → `\\` → `\\\\`.
It is not possible to construct a bijection between a single character in an incomplete character set and a single character in a complete character set, only between a single and a single/multiple character.
As for your problem, a filename `‛｜` (I mean if you directly look at the backend) in a backend means `｜`.


> Why not use the Unicode private use area as a rclone exclusive character, and prefix all escape sequences with that? `�‛｜`This preserves the current usage of repeating `‛‛‛‛‛‛`s. Admittedly, this does kick the can down the line, but anyone using private use characters is probably a super user of some sort and can read the documentation.

Repeating character `‛` is not the source of the problem.
Firstly, `‛`(U+201B, not the character you can directly type) is extremely uncommon, and I bet no one's original files will have that character before they touch rclone.
Secondly, repeating `‛` usually occurs because people misuse rclone's encoding system.

For example, if one has a `你好｜世界？` in the local Linux filesystem, he copies it to Onedrive with rclone (he will have `你好‛｜世界‛？` on Onedrive), and then he copies it from Onedrive to a local Windows filesystem with Onedrive (he will have `你好‛｜世界‛？` on the local Windows filesystem), and if he copies it from the local Windows filesystem to Onedrive with rclone, he will have `你好‛‛｜世界‛‛？` on Onedrive.
The correct way is to use rclone **from beginning to end**, including creating, investigating, copying, and deleting. Then he will always only see  `你好｜世界？`.

> Should the two stacks be equal, no encoding is done. If the stacks are unequal, then encoding/decoding is done.

Seems correct but not exactly right.
There is also an intermediate layer of standard encoding between the source backend and the target backend, so the way to make this short-circuit happen is to specify both the source and target encoding as standard. see #7456 .

URenko:
Yes if we could go back to 2019 I would definitely choose to develop another coding system from scratch, but it has been 5 years.
Remember also that `None` is the default encoding for many backends. (It led to my nightmare 5 years ago.)

In addition to a couple of incorrect implementations that I fixed, there are two main design issues (but not implementation issues) with the current coding system. They don't pose a real problem if the user can use them properly and fixing them is destructive.

- **Use full-width characters which are common in Chinese and Japanese, as escape characters (`|`→`｜`, `?`→`？`...).**
It is not easy for uses to use rclone **from beginning to end**.
So at the very least, users are likely to create files on their own (e.g. `你好｜世界？`).
Then rclone thinks it's encoded, i.e. it means `你好|世界?` (which always happens unless encoding is `Raw` or use the short-circuit trick).
- **Use encoding by default.**
If we do not use encoding by default, we will not run into the problem mentioned above.
If we could redesign the coding system. I'd recommend no default encoding and that the encoding be used as a layer of backend (like Hasher).
And when people run into encoding issues with uploading (which should be rare), they will look at the docs and then realize that there is a backend that can solve the problem.

But for now I'd like to keep backward compatibility as much as possible. And, at least for sure, we need a way to actually disable coding (`Raw`)

iseahound:
Standard encoding seems to have problems with orphaned escaped characters:

If source and destination encoding is both `windows` this works fine,
If source and destination encoding is both `None` then deletion cannot happen.

**test.bat (Save as UTF-8)**
```
@echo off
chcp 65001

mkdir a
mkdir b
echo "hello world" > b/‛｜.txt
rclone --config="" sync a/ b/ --local-encoding=None
pause
```

URenko:
Yes, there is an intermediate layer of "standard" encoding, so the short-circuit trick is to use "standard" for both source and destination.

But I think you have found another bug. 🙃

iseahound:
I have a simple setup of 2 cloud services and one local drive 🤣

URenko:
> There seems to be problems with orphaned files with escaped characters in the filename:
> 
> If source and destination encoding is both `windows` this works fine, If source and destination encoding is both `None` then deletion cannot happen.
> 
> **test.bat (Save as UTF-8)**
> 
> ```
> @echo off
> chcp 65001
> 
> mkdir a
> mkdir b
> echo "hello world" > b/‛｜.txt
> rclone --config="" sync a/ b/ --local-encoding=None
> pause
> ```
> 
> In other words, even if the two stacks "cancel each other out" there exists a difference in behavior. Using your standard `--local-encoding `None,Slash,Ctl,Del,Dot` seems to fix it for some reason...

Now I think this is not a bug (at least not a bug in the local backend or sync).
This is what happens:
None: `‛｜.txt` decoded to Raw: `‛｜.txt` encoded to Standard: `‛‛‛｜.txt` decoded to Raw: `‛｜.txt` encoded to None: `‛‛｜.txt`
The problem is `‛｜.txt` is ill-defined in the encoding context None.
Note that each encoding has its own ill-defined cases (except Raw), so when you specify an encoding, make sure that the filenames in the backend are already encoded (well-defined).
Raw is what you should use.

URenko:
Rebased to v1.67.0.

These bugs are dangerous because they may lead to unknowing file loss.
So before merging I build the binaries in [my build branch](https://github.com/URenko/rclone/tree/build) and they can be downloaded from [the GitHub action artifacts](https://github.com/URenko/rclone/actions/runs/9527112881).

vfs/vfscache/item.go
```diff
@@ -581,7 +582,7 @@ func (item *Item) _store(ctx context.Context, storeFn StoreFn) (err error) {
 	// defer log.Trace(item.name, "item=%p", item)("err=%v", &err)
 
 	// Transfer the temp file to the remote
-	cacheObj, err := item.c.fcache.NewObject(ctx, item.name)
+	cacheObj, err := item.c.fcache.NewObject(ctx, item.c.fcache.(*local.Fs).LocalToStandardPath(toOSPath(item.name)))
```
ncw, line 585: This is a fix for #7760  isn't it - it should say `Fixes #7760` in the commit messsage. Can you also write what the problem was and how it fixed it in the commit message too - thanks!

I think a better fix would be to create the local backend with the encoding that we want, something like this (untested)

```patch
diff --git a/vfs/vfscache/cache.go b/vfs/vfscache/cache.go
index 9bcfc90ee..0dc3b7630 100644
--- a/vfs/vfscache/cache.go
+++ b/vfs/vfscache/cache.go
@@ -212,7 +212,7 @@ func (c *Cache) createItemDir(name string) (string, error) {
 
 // getBackend gets a backend for a cache root dir
 func getBackend(ctx context.Context, parentPath string, name string, relativeDirPath string) (fs.Fs, error) {
-	path := fmt.Sprintf("%s/%s/%s", parentPath, name, relativeDirPath)
+	path := fmt.Sprintf(":local,encoding=%v:%s/%s/%s", encoder.OS, parentPath, name, relativeDirPath)
 	return fscache.Get(ctx, path)
 }
```

backend/local/local.go
```diff
@@ -1573,32 +1573,46 @@ func (o *Object) SetMetadata(ctx context.Context, metadata fs.Metadata) error {
...
 		}
-		s = vol + enc.FromStandardPath(s[len(vol):])
-		s = filepath.FromSlash(s)
-		if !noUNC {
-			// Convert to UNC
-			s = file.UNCPath(s)
+		s = s[len(vol):]
+	}
+	// Don't use FromStandardPath. Make sure Dot (`.`, `..`) as name will not be reencoded
+	if enc != encoder.Standard {
```
ncw, line 1586: Perhaps just mask out the `Dot` encoding from enc? That would make the code a lot simpler

something like this?

```
enc &= ^encoding.EncodeDot
```

If you can minimise this change I'm much more likely to  merge it!

ncw (commented):
Thank you for doing this - it has caused a lot of frustration for users.

The first two commits are great! The 3rd and 4th need a bit of work (see inline comments) but the ideas are good.

Can you write more in [the commit messages](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages) please, and reference any existing issues.

Thanks and excited to get this merged for 1.68

backend/local/local.go
URenko: Consider someone want to create `./．` (Raw) in the current directory, the Standard encoding is `．/‛．`.
Then consider local-encoding with EncodeDot.
So given Standard: `./．/‛．` (the first dot means current directory), if use `enc &= ^encoding.EncodeDot`, it will be:
Standard: `./．/‛．` decoded to Raw: `././．` encoded to enc: `././．`, then Abs() lead to `．`.
But what we want write to the local filesystem is Raw: `．/‛．`.

I've added a comment to avoid anyone accidentally optimizing it by mistake.

vfs/vfscache/item.go
URenko: Good idea!
Just need to place parameter in quotes.
Tested and working well.

ncw (approved):
I've given this one last review and I think it is looking great - thank you :-)

I'm going to merge this now.

PR #7793, opened 2024-04-22
fs: adding warning to user that rclone config is not encrypted by default. fixes #7314

### LINKED ISSUE #7314: Default to secure credentials storage in "rclone config" etc

`rclone config` defaults to storing credentials that can be used to access cloud storage in cleartext (weakly obfuscated) in a config file on disk at a well-known path. This is a high risk for credential-scraper attacks.

E.g. if `rclone config` is used to set up a GDrive connection with Oauth token/secret, `rclone` will store these and the token acquired using them in `~/.config/rclone/rclone.conf`.

It does not recommend the use of a `--password-command` or password-helper (like git uses), recommend encryption of the config file, etc. This is an insecure default behaviour; while the user's homedir is *hopefully* safe from attacks, credentials for remote access to other systems are high-value targets that should have extra protection to limit attackers' ability to springboard from a compromised machine to other targets.

`rclone config` should at display a prominent warning when an unencrypted configuration is used, recommend the "set configuration password" option be used, and direct the user to https://rclone.org/docs/#configuration-encryption for details.

IMO rclone should also print a short security warning to stderr at every invocation when using an unencrypted config, unless a flag/env-var is set to deliberately silence this.

Ideally `rclone` should natively support indirection of individual secrets and/or the rclone config file encryption secret into the operating system's secret store. Most OSes now provide a suitable secret store with SSO integration for user interactive login, timeouts and re-prompting, biometrics integration etc. OSX's Keychain and Linux (GNOME)'s `libsecret` are good examples. To keep OS dependencies down, this could be indirected via canned "password helper" commands, much as is done by git.

A discussion in https://github.com/rclone/rclone/issues/3386 points out that the `rclone --password-command` option exists, and could potentially be used to externalize secrets (or is it only for the config file encryption?), but this is difficult to find, and it's hard to figure out how to take an existing configuration with baked-in passwords and convert it for use this way. A post at https://forum.rclone.org/t/how-to-encrypt-a-configuration-file/11917 suggests I'm far from the only one to find this confusing.

#### How to use GitHub

KDreynolds:

adds a simple warning message to create new config flow so users know default behavior is to use unencrypted config files. 

issue #7314 

#### Checklist

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -265,6 +265,7 @@ defaults for questions as usual.
 Note that |bin/config.py| in the rclone source implements this protocol
 as a readable demonstration.
 `, "|", "`")
+
 var configCreateCommand = &cobra.Command{
 	Use:   "create name type [key value]*",
 	Short: `Create a new remote with name, type and options.`,
@@ -292,9 +293,17 @@ using remote authorization you would do this:
 		if err != nil {
 			return err
 		}
-		return doConfig(args[0], in, func(opts config.UpdateRemoteOpt) (*fs.ConfigOut, error) {
+		err = doConfig(args[0], in, func(opts config.UpdateRemoteOpt) (*fs.ConfigOut, error) {
 			return config.CreateRemote(context.Background(), args[0], args[1], in, opts)
 		})
+		if err != nil {
+			return err
+		}
+
+		// Append the security warning about unencrypted configurations
+		fmt.Fprintln(os.Stderr, "\nWARNING: By default, rclone configuration files are not encrypted, which may pose a security risk. It is highly recommended to encrypt your configuration file to protect sensitive information. For details on how to encrypt your configuration, visit https://rclone.org/docs/#configuration-encryption.")
+
+		return nil
 	},
 }
 


KDreynolds:
not sure if this is how we would want to do this or if this is still needed, but i took a crack at it.

cmd/config/config.go
```diff
@@ -292,9 +293,17 @@ using remote authorization you would do this:
...
 		}
-		return doConfig(args[0], in, func(opts config.UpdateRemoteOpt) (*fs.ConfigOut, error) {
+		err = doConfig(args[0], in, func(opts config.UpdateRemoteOpt) (*fs.ConfigOut, error) {
 			return config.CreateRemote(context.Background(), args[0], args[1], in, opts)
 		})
+		if err != nil {
+			return err
+		}
+
+		// Append the security warning about unencrypted configurations
```
ncw, line 303: We should probably only show this message if the config isn't encrypted then we can make a definitive statement "your config is not encrypted..."

To do that we need to export the encrypted state from `fs/config/crypt.go` something like this (untested)

```go
func IsEncrypted() bool {
    return len(configKey) != 0
```

I think also that this should be printed to Stderr like the rest of the config stuff, and I'd probably hard wrap it to 80 chars.

cmd/config/config.go
```diff
@@ -292,9 +293,17 @@ using remote authorization you would do this:
...
+		fmt.Fprintln(os.Stderr, "\nWARNING: By default, rclone configuration files are not encrypted, which may pose a security risk. It is highly recommended to encrypt your configuration file to protect sensitive information. For details on how to encrypt your configuration, visit https://rclone.org/docs/#configuration-encryption.")
```
ncw, line 304: I think this should probably only be on plain `rclone config` (which is `configCommand` above) not here which is `rclone config create` which is often used by scripts and we don't want to print extra stuff here.

ncw (commented):
Thanks for having a go at this.

I put some suggestions inline.

We need to be careful not to break scripting of the rclone config commands here.

PR #7795, opened 2024-04-22
bisync: another day, another set of integration test fixes

nielash:

See https://github.com/rclone/rclone/pull/7743#issuecomment-2071094755 and feel free to cherry-pick as you see fit if some of these warrant more discussion.

- https://github.com/rclone/rclone/pull/7743#issuecomment-2071094755

#### Checklist

### DIFF
--- a/backend/azureblob/azureblob.go
+++ b/backend/azureblob/azureblob.go
@@ -1088,7 +1088,7 @@ func (f *Fs) list(ctx context.Context, containerName, directory, prefix string, 
 			isDirectory := isDirectoryMarker(*file.Properties.ContentLength, file.Metadata, remote)
 			if isDirectory {
 				// Don't insert the root directory
-				if remote == directory {
+				if remote == f.opt.Enc.ToStandardPath(directory) {
 					continue
 				}
 				// process directory markers as directories
--- a/cmd/serve/s3/backend.go
+++ b/cmd/serve/s3/backend.go
@@ -213,6 +213,14 @@ func (b *s3Backend) GetObject(ctx context.Context, bucketName, objectName string
 	}, nil
 }
 
+// storeModtime sets both "mtime" and "X-Amz-Meta-Mtime" to val in b.meta.
+// Call this whenever modtime is updated.
+func (b *s3Backend) storeModtime(fp string, meta map[string]string, val string) {
+	meta["X-Amz-Meta-Mtime"] = val
+	meta["mtime"] = val
+	b.meta.Store(fp, meta)
+}
+
 // TouchObject creates or updates meta on specified object.
 func (b *s3Backend) TouchObject(ctx context.Context, fp string, meta map[string]string) (result gofakes3.PutObjectResult, err error) {
 	_, err = b.vfs.Stat(fp)
@@ -237,6 +245,7 @@ func (b *s3Backend) TouchObject(ctx context.Context, fp string, meta map[string]
 	if val, ok := meta["X-Amz-Meta-Mtime"]; ok {
 		ti, err := swift.FloatStringToTime(val)
 		if err == nil {
+			b.storeModtime(fp, meta, val)
 			return result, b.vfs.Chtimes(fp, ti, ti)
 		}
 		// ignore error since the file is successfully created
@@ -245,6 +254,7 @@ func (b *s3Backend) TouchObject(ctx context.Context, fp string, meta map[string]
 	if val, ok := meta["mtime"]; ok {
 		ti, err := swift.FloatStringToTime(val)
 		if err == nil {
+			b.storeModtime(fp, meta, val)
 			return result, b.vfs.Chtimes(fp, ti, ti)
 		}
 		// ignore error since the file is successfully created
@@ -307,6 +317,7 @@ func (b *s3Backend) PutObject(
 	if val, ok := meta["X-Amz-Meta-Mtime"]; ok {
 		ti, err := swift.FloatStringToTime(val)
 		if err == nil {
+			b.storeModtime(fp, meta, val)
 			return result, b.vfs.Chtimes(fp, ti, ti)
 		}
 		// ignore error since the file is successfully created
@@ -315,6 +326,7 @@ func (b *s3Backend) PutObject(
 	if val, ok := meta["mtime"]; ok {
 		ti, err := swift.FloatStringToTime(val)
 		if err == nil {
+			b.storeModtime(fp, meta, val)
 			return result, b.vfs.Chtimes(fp, ti, ti)
 		}
 		// ignore error since the file is successfully created
@@ -425,6 +437,7 @@ func (b *s3Backend) CopyObject(ctx context.Context, srcBucket, srcKey, dstBucket
 		if err != nil {
 			return result, nil
 		}
+		b.storeModtime(fp, meta, val)
 
 		return result, b.vfs.Chtimes(fp, ti, ti)
 	}
--- a/fs/sync/sync.go
+++ b/fs/sync/sync.go
@@ -1203,7 +1203,7 @@ func (s *syncCopyMove) setDelayedDirModTimes(ctx context.Context) error {
 			g.Go(func() error {
 				var err error
 				// if item.src is set must copy full metadata
-				if item.src != nil {
+				if item.src != nil && s.setDirMetadata {
 					_, err = operations.CopyDirMetadata(gCtx, s.fdst, item.dst, item.dir, item.src)
 				} else {
 					_, err = operations.SetDirModTime(gCtx, s.fdst, item.dst, item.dir, item.modTime)
--- a/fs/sync/sync_test.go
+++ b/fs/sync/sync_test.go
@@ -154,7 +154,7 @@ func testCopyMetadata(t *testing.T, createEmptySrcDirs bool) {
 	if features.ReadDirMetadata {
 		fstest.CheckEntryMetadata(ctx, t, r.Fremote, fstest.NewDirectory(ctx, t, r.Fremote, dirPath), dirMetadata)
 	}
-	if !createEmptySrcDirs {
+	if !createEmptySrcDirs || !features.CanHaveEmptyDirectories {
 		// dir must not exist
 		_, err := fstest.NewDirectoryRetries(ctx, t, r.Fremote, emptyDirPath, 1)
 		assert.Error(t, err, "Not expecting to find empty directory")
@@ -2310,15 +2310,19 @@ func testSyncBackupDir(t *testing.T, backupDir string, suffix string, suffixKeep
 
 	r.CheckRemoteItems(t, file1b, file2, file3a, file1a)
 }
+
 func TestSyncBackupDir(t *testing.T) {
 	testSyncBackupDir(t, "backup", "", false)
 }
+
 func TestSyncBackupDirWithSuffix(t *testing.T) {
 	testSyncBackupDir(t, "backup", ".bak", false)
 }
+
 func TestSyncBackupDirWithSuffixKeepExtension(t *testing.T) {
 	testSyncBackupDir(t, "backup", "-2019-01-01", true)
 }
+
 func TestSyncBackupDirSuffixOnly(t *testing.T) {
 	testSyncBackupDir(t, "", ".bak", false)
 }
@@ -2788,7 +2792,7 @@ func predictDstFromLogger(ctx context.Context) context.Context {
 }
 
 func DstLsf(ctx context.Context, Fremote fs.Fs) *bytes.Buffer {
-	var opt = operations.ListJSONOpt{
+	opt := operations.ListJSONOpt{
 		NoModTime:  false,
 		NoMimeType: true,
 		DirsOnly:   false,
--- a/fstest/fstest.go
+++ b/fstest/fstest.go
@@ -48,7 +48,7 @@ var (
 	// ListRetries is the number of times to retry a listing to overcome eventual consistency
 	ListRetries = flag.Int("list-retries", 3, "Number or times to retry listing")
 	// MatchTestRemote matches the remote names used for testing
-	MatchTestRemote = regexp.MustCompile(`^rclone-test-[abcdefghijklmnopqrstuvwxyz0123456789]{24}$`)
+	MatchTestRemote = regexp.MustCompile(`^rclone-test-[abcdefghijklmnopqrstuvwxyz0123456789]{12}$`)
 )
 
 // Initialise rclone for testing
@@ -266,7 +266,7 @@ func CheckListingWithRoot(t *testing.T, f fs.Fs, dir string, items []Item, expec
 	var objs []fs.Object
 	var dirs []fs.Directory
 	var err error
-	var retries = *ListRetries
+	retries := *ListRetries
 	sleep := time.Second / 2
 	wantListing := makeListingFromItems(items)
 	gotListing := "<unset>"
@@ -431,7 +431,7 @@ func RandomRemoteName(remoteName string) (string, string, error) {
 		if !strings.HasSuffix(remoteName, ":") {
 			remoteName += "/"
 		}
-		leafName = "rclone-test-" + random.String(24)
+		leafName = "rclone-test-" + random.String(12)
 		if !MatchTestRemote.MatchString(leafName) {
 			log.Fatalf("%q didn't match the test remote name regexp", leafName)
 		}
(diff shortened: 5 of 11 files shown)


nielash:
I wonder if the `TestSFTPRclone:` issue is related to the one identified on https://github.com/rclone/rclone/pull/7791?

> And fix vfs cache encoding described in https://github.com/rclone/rclone/issues/7760
The problem is when we create and write cache, we use the "OS" encoded ("OS" encoding is hardcoded for different operating systems, not specified by users).
But when we copy it to the remote (backend), we use local.Fs, which accept a name already encoded in "Standard" way.
So we need to send a "Standard" encoded name.

I've reverted it for now.

nielash:
This was a real tricky one, but I think I finally figured out the `TestSFTPRclone:` issue. https://github.com/rclone/rclone/pull/7795/commits/15ff8ab01eacf3fe918e35a844e1b55d6f44a9f5

cmd/bisync/bisync_test.go
```diff
@@ -939,6 +939,9 @@ func (b *bisyncTest) checkPreReqs(ctx context.Context, opt *bisync.Options) (con
 		b.fs2.Features().Disable("Copy") // API has longstanding bug for conflictBehavior=replace https://github.com/rclone/rclone/issues/4590
 		b.fs2.Features().Disable("Move")
 	}
+	if strings.Contains(strings.ToLower(fs.ConfigString(b.fs1)), "mailru") || strings.Contains(strings.ToLower(fs.ConfigString(b.fs2)), "mailru") {
```
ncw, line 942: At some point in the future when we can specify global config in backend config we should move this config from here into the `TestMailru` remote. Let's see how well it works here.

fstest/fstest.go
```diff
@@ -47,7 +47,7 @@ var (
 	// ListRetries is the number of times to retry a listing to overcome eventual consistency
 	ListRetries = flag.Int("list-retries", 3, "Number or times to retry listing")
 	// MatchTestRemote matches the remote names used for testing
-	MatchTestRemote = regexp.MustCompile(`^rclone-test-[abcdefghijklmnopqrstuvwxyz0123456789]{24}$`)
+	MatchTestRemote = regexp.MustCompile(`^rclone-test-[abcdefghijklmnopqrstuvwxyz0123456789]{12}$`)
```
ncw, line 50: This also needs to change

https://github.com/rclone/rclone/blob/0735f44f910c3bafe38a7e7e04b07ade48edc1b7/fstest/test_all/clean.go#L20

Though maybe that one should say `{12,24}` to clean the old and the new!

vfs/dir.go
```diff
@@ -371,7 +371,9 @@ func (d *Dir) renameTree(dirPath string) {
 	// Make sure the path is correct for each node
 	if d.path != dirPath {
 		fs.Debugf(d.path, "Renaming to %q", dirPath)
+		delete(d.parent.items, name(d.path))
```
ncw, line 374: This is a good fix, but it really needs a test. I try to keep the VFS coverage up as all these corner cases are really difficult to track down!

vfs/dir.go
```diff
@@ -418,6 +422,15 @@ func (d *Dir) rename(newParent *Dir, fsDir fs.Directory) {
 	}
 }
 
+// convert path to name
+func name(p string) string {
+	p = path.Base(p)
+	if p == "." {
+		p = "/"
```
ncw, line 429: That should probably be `p = ""` to fit with rclone conventions, not 100% sure!

cmd/serve/s3/backend.go
```diff
@@ -213,6 +213,14 @@ func (b *s3Backend) GetObject(ctx context.Context, bucketName, objectName string
 	}, nil
 }
 
+// storeModtime sets both "mtime" and "X-Amz-Meta-Mtime" to val in b.meta.
+// Call this whenever modtime is updated.
+func (b *s3Backend) storeModtime(fp string, meta map[string]string, val string) {
+	meta["X-Amz-Meta-Mtime"] = val
+	meta["mtime"] = val
```
ncw, line 220: It should only be necessary to set `X-Amz-Meta-Mtime` as the s3 backend turns it into the `mtime` metadata.

So I'm not 100% sure what this is needed for.

I think the `serve s3` code is a bit confused about exactly which parameter to use so we should probably unconfuse it rather than make it worse!

backend/azureblob/azureblob.go
```diff
@@ -1088,7 +1088,7 @@ func (f *Fs) list(ctx context.Context, containerName, directory, prefix string,
 			isDirectory := isDirectoryMarker(*file.Properties.ContentLength, file.Metadata, remote)
 			if isDirectory {
 				// Don't insert the root directory
-				if remote == directory {
+				if remote == f.opt.Enc.ToStandardPath(directory) {
```
ncw, line 1091: `directory` should already be in Standard encoding here since it ultimately came from the user rather than the backend, so I'm not sure what is going on here.

Same for the other backends.

ncw (commented):
Thank you @nielash there are some great fixes in there and apologies for taking ages to review them.

I'll cherry pick the ones that don't need discussion and we can discuss the rest!

See comments inline.

Thank you

backend/azureblob/azureblob.go
nielash: I'm not sure about that. For example here:

https://github.com/rclone/rclone/blob/e9e9feb21eeb200782180e8532f2a0a95f745740/backend/s3/s3.go#L4178

`directory` is returned from `f.split`, which does `FromStandardPath` on it:

https://github.com/rclone/rclone/blob/e9e9feb21eeb200782180e8532f2a0a95f745740/backend/s3/s3.go#L2823-L2826

Here's a test to reproduce the issue:

```
go test -timeout 3h -run '^TestBisyncRemoteRemote/extended_filenames$' github.com/rclone/rclone/cmd/bisync -remote TestS3,directory_markers: -v 
```

cmd/serve/s3/backend.go
nielash: Here's a test to reproduce:

```
go run ./fstest/test_all  -run '^TestBisyncRemoteRemote/ext_paths.*$' -timeout 30s -verbose -maxtries 1 -remotes TestS3Rclone:
```

If I remember correctly, I think the problem is that there are blocks like this which try `X-Amz-Meta-Mtime` first but fall back to using `mtime`:

https://github.com/rclone/rclone/blob/e9e9feb21eeb200782180e8532f2a0a95f745740/cmd/serve/s3/backend.go#L307-L321

And previously it was not always consistently saving the one it ended up using in memory.

And then there's also `Last-Modified` just to make things even more fun!

https://github.com/rclone/rclone/blob/e9e9feb21eeb200782180e8532f2a0a95f745740/cmd/serve/s3/backend.go#L194-L197

vfs/dir.go
nielash: I think I copied this from here:

https://github.com/rclone/rclone/blob/e9e9feb21eeb200782180e8532f2a0a95f745740/vfs/dir.go#L163-L172

vfs/dir.go
nielash: Ok, I'll try to add one! (Although FWIW, I found this because it was causing `TestSFTPRclone:` to fail!)

nielash: Also btw, I am now seeing a bunch of `expecting X to have MkdirMetadata method: optional feature not implemented` and `expecting X to have SetMetadata method` errors that I'm pretty sure were not there before...maybe related to the recent `--create-empty-src-dirs` change?

https://pub.rclone.org/integration-tests/2024-05-13-010010/sftp-cmd.bisync-TestSFTPRclone-5.txt

nielash: I added a test.

fstest/fstest.go
nielash: Made this change.

nielash:
@ncw heads up -- potentially important bug fix here (https://github.com/rclone/rclone/pull/7795/commits/1e8d42f3bbb49ce919aa93a28cf8f4cdf3725ac1) for a regression introduced by our recent `--create-empty-src-dirs` change (https://github.com/rclone/rclone/commit/10eb4742dd9da47342ef768576fefd7c5f149443)

(See [this log](https://pub.rclone.org/integration-tests/2024-05-18-010009/sftp-cmd.bisync-TestSFTPRclone-5.txt) for many examples of the problem)

backend/azureblob/azureblob.go
ncw: Yes I've checked the 3 backends in question and you are 100% right :-)

I'll cherry pick this commit now.

fs/sync/sync.go
```diff
@@ -1203,7 +1203,7 @@ func (s *syncCopyMove) setDelayedDirModTimes(ctx context.Context) error {
 			g.Go(func() error {
 				var err error
 				// if item.src is set must copy full metadata
-				if item.src != nil {
+				if item.src != nil && s.setDirMetadata {
```
ncw, line 1206: I have now fixed this is a slightly different way in 39f8d039fe07c0215dba8bd192edbadf53c3d091 to make it clearer what is going on rather than using the state of `item.src == nil` to signify metadata.

fs/sync/sync_test.go
```diff
@@ -154,7 +154,7 @@ func testCopyMetadata(t *testing.T, createEmptySrcDirs bool) {
 	if features.ReadDirMetadata {
 		fstest.CheckEntryMetadata(ctx, t, r.Fremote, fstest.NewDirectory(ctx, t, r.Fremote, dirPath), dirMetadata)
 	}
-	if !createEmptySrcDirs {
+	if !createEmptySrcDirs || !features.CanHaveEmptyDirectories {
```
ncw, line 157: I fixed this one in 76798d5bb1382ffd191a0b685178eee0751bf3b4 - I'd forgotten I hadn't merged your fix - sorry!

fstest/fstest.go
ncw: Thanks. I've cherry-picked this now.

vfs/dir.go
ncw: Thank you. I've cherry picked this :-)

cmd/serve/s3/backend.go
ncw: I think we should probably fix `serve s3` to be consistent and use just one of these values, but I decided to cherry-pick your patch for v1.67 - thank you :-)

ncw:
I've merged all these (or equivalents) now, so I shall close this PR - thank you :-)

Let's see how the integration tests do tomorrow!

ncw:
Much better!

36 test failures in total
- 19 bisync failures (ignoring QingStor where the remote is broken)
- 12 other failures (ignoring QingStor)

nielash:
> Much better!
> 
> 36 test failures in total

Getting closer! I think a lot of the remaining ones are issues I made guesses about [here](https://github.com/rclone/rclone/pull/7743#issuecomment-2071094755). I think they are all backend-specific issues now, as opposed to bisync issues.

ncw:
I removed the ones that are now fixed from your list and added one more, protondrive.

It looks like there are a few low hanging fruits there.

## General comments

- some backends can have blank hashes - the bisync tests need to be able to cope with that. A blank hash means "hash missing" and shouldn't be an error.
- more comments inline

## Current failures

- `TestAzureFiles:` looks like something related to hashes / hash types. It should be finding a hash difference and it's not (maybe they are blank?)
- `TestBox:` seems to be some issue with copying a file over itself. 
    - ncw: Yes this rings a bell. It is undefined in the integration tests whether copying or moving a file over an existing file is allowed or not. `operations.Copy` should always call `backend.Update` if there is an existing file which should allow it to be overwritten. Whether `backend.Move` allows this I think is undefined. If we want all backends to behave the same way then we need to make an integration test for it and fix the failures! Do you know what operation is being invoked here?
```
ERROR : file1.txt: Failed to copy: Error "item_name_in_use" (409): Item with the same name already exists ({"conflicts":{"type":"file","id":"1509373873360","file_version":{"type":"file_version","id":"1657286950960","{hashtype}":"da39a3ee5e6b4b0d3255bfef95601890afd80709"},"sequence_id":"0","etag":"0","{hashtype}":"da39a3ee5e6b4b0d3255bfef95601890afd80709","name":"file1.txt"}})
```
- `TestCache:` looks like an encoding issue when a directory name has control chars.
    - ncw: I don't care about cache as it is deprecated - we should probably stop it running bisync tests
- `TestChunkerChunk50bYandex:` seems like a `yandex` issue relating to copying a file over itself.
    - ncw: This is will be fixed when fixing the problem with the yandex backed I think
```
ERROR : file1.txt: Failed to copy: couldn't copy file: [409 - DiskResourceAlreadyExistsError] Resource "disk:/rclone-test-lebikon0foyesex0womeyod4/073716le/basic/path1/file1.txt" already exists. (Ресурс "disk:/rclone-test-lebikon0foyesex0womeyod4/073716le/basic/path1/file1.txt" уже существует.)
```
- `TestJottacloud:` seems like unicode like `folder/éééö.txt` is getting auto-normalized in a way we are not currently looking for.
- `TestLinkbox:` seems to be a filename limitation:
```
subdir/.chk_file: Failed to copy: Update second step failed: Linkbox error 500: The file name needs to include a suffix, such as xxx.mp4
```
- `TestMega:` seems like a bug with `o.Remove` not actually removing the object, but not returning an error. That `modtime_write_test` file shouldn't still be there by the time the real test starts.
    - ncw: there are a number of weird bugs in the mega backend due to it being a reverse engineered protocol
- `TestnStorage:` looks like we just need to ignore the `Implicitly create directory` logs -- should be an easy fix.
- `TestOracleObjectStorage:` seems like either the backend can't handle names with whitespace / control chars or we are not encoding them properly in the API call.
```
Message: Failed to parse json from response body due to: unexpected end of JSON input. With response body .
```
- `TestPikPak:` seems to be unable to open files for hashing
```
failed to open file file1.txt: can't download - no link to download
```
- `TestProtonDrive` - 	TestBisyncRemoteLocal/nomodtime, TestBisyncLocalRemote/ext_paths, TestBisyncLocalRemote/extended_filenames
- `TestPutio:` something related to renaming a file:
```
ERROR : file1.txt: Failed to copy: unexpected status code (400) response while doing POST to https://api.put.io/v2/files/rename
```
- `TestQuatrix:` looks like the fs `.String()` is not including the subpath, same issue as I fixed in `hdfs`. Should be an easy fix.
- `TestSFTPOpenssh:` seems to be something related to unicode encoding, or possibly modtimes not always updating as they should.
- `TestSFTPRsyncNet:` looks like maybe we are not properly terminal-escaping newlines when we try to calculate md5s:
```
ERROR : file2 with
white space.txt.kotizop8.partial: Failed to calculate dst hash: failed to calculate md5 hash: failed to run "md5 -r /data2/home/59018/rclone-test-jiyegog4voqinat7lifojaj7/043414fi/extended_filenames/path2/subdir\\ with'\n'white\\ space.txt/file2\\ with'\n'white\\ space.txt.kotizop8.partial": md5: /data2/home/59018/rclone-test-jiyegog4voqinat7lifojaj7/043414fi/extended_filenames/path2/subdir with: No such file or directory: Process exited with status 1
```
- `TestStorj:` it seems like there's a bug where the bucket needs to be explicitly `Mkdir`'d in a way that other backends don't need. I've noticed this before. 
```
storj: bucket: uplink: bucket not found ("rclone-test-mepoqax3vuyuris8velinag1")
```
- `TestWebdavNextcloud:` Looks like the initial problem is related to hashes being unexpectedly blank after server-side copy/move. And then some of the retries fail because the container name is already in use (maybe it's not properly stopped after each try?)
- `TestWebdavOwncloud:` same blank hash issue as nextcloud. (although interestingly `TestBisyncRemoteRemote/rclone_args` passes for owncloud and not for nextcloud)
- `TestYandex:` same issue as `TestChunkerChunk50bYandex:` relating to copying a file over itself.

nielash:
> * ncw: Yes this rings a bell. It is undefined in the integration tests whether copying or moving a file over an existing file is allowed or not. `operations.Copy` should always call `backend.Update` if there is an existing file which should allow it to be overwritten. Whether `backend.Move` allows this I think is undefined. If we want all backends to behave the same way then we need to make an integration test for it and fix the failures! Do you know what operation is being invoked here?

@ncw It appears to me to be a normal copy operation of a file with the same name as an existing file. It may be happening only when the dest file's size is 0 and the src is > 0. But I'm not totally sure of that, and it's hard to test as I don't have a Box account.

Issue #7783, opened 2024-04-19
rclone serve: support systemd socket activation


flokli:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/systemd-socket-activation-for-rclone-serve/45644

#### What is your current rclone version (output from `rclone version`)?
1.66.0

#### What problem are you are trying to solve?
I'd like to create some `rclone serve` processes, and use systemd socket activation to start them on demand.

#### How do you think rclone should be changed to solve that?
Essentially, sockets are passed as FDs directly to the process.

https://darkcoding.net/software/systemd-socket-activation-in-go/ describes a golang-only solution to implement this.

It'd be nice if rclone would support this too. In the presence of the `LISTEN_PID` env var, it could automatically bind to FD 3 - or the `--addr` CLI arg could get additional syntax for this, similar to the (undocumented in the `--help` output) `unix://` syntax.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

flokli:
This is implemented in https://github.com/rclone/rclone/pull/7801, PTAL.

PR #7801, opened 2024-04-24
http/sftp: support listening on passed FDs

flokli:
Instead of the listening addresses specified above, rclone will listen to all FDs passed by the service manager, if any (and ignore any arguments passed by `--{{ .Prefix }}addr`.

This allows rclone to be a socket-activated service. It can be configured as described in https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html

It's possible to test this interactively through `systemd-socket-activate`, firing of a request in a second terminal:

```
❯ systemd-socket-activate -l 8088 -l 8089 --fdname=foo:bar -- ./rclone serve webdav :local:test/
Listening on [::]:8088 as 3.
Listening on [::]:8089 as 4.
Communication attempt on fd 3.
Execing ./rclone (./rclone serve webdav :local:test/)
2024/04/24 18:14:42 NOTICE: Local file system at /home/flokli/dev/flokli/rclone/test: WebDav Server started on [sd-listen:bar-0/ sd-listen:foo-0/]
```

Closes https://github.com/rclone/rclone/issues/7783

https://github.com/rclone/rclone/issues/7783
https://forum.rclone.org/t/systemd-socket-activation-for-rclone-serve/45644

#### Checklist


flokli:
poke @ncw :-)

ncw:
Can you stub it out on plan 9?

flokli:
That's what I did upstream, in https://github.com/coreos/go-systemd/pull/440.

If this doesn't get merged there, I could add a little wrapper package in rclone, and do the plan9 stubbing there, but it'd be more code on the rclone side.

Is this the only thing blocking this, or do you have other feedback on the code?

ncw:
Well you could do the same thing in the rclone code with build tags - that is what I was suggesting. Though if your pr is merged that is the neatest way for rclone!

I have looked at the code which looks ok (I'll do a formal review at some point). I would like some explanation of how it works in the docs and example config of how you would use it with systemd would be very useful.

Ones thing that wasn't clear to me - does systemd start a new rclone instance for every incoming connection? Or does it keep things running for a while. I guess this might be in the systemd docs!

flokli:
Yeah, I'm also hoping PR gets merged into go-systemd, as essentially everyone linking against the activation package and building or this arch would do stubbing on their own, and there already is a windows stub...

Regarding systemd configuration, there's a lot of different config knobs, which is why I only linked to the systemd docs. You can bind on Unix sockets, TCP ports, ...

rclone keeps running, and normally gets started lazily on the first connection. If rclone had a parameter to exit the process after a specific time of inactivity, it would also get fully shut down, but that's a bit out of scope for this PR.

ncw:
Hmm, looks like we might have to go for build tags - not a lot of action on that PR...

flokli:
Alright, let me take a look at that

flokli:
I rebased this, and moved the wrapping into our own wrapper package, linking to the upstream issue. also added socket activation to the `serve sftp` command (which only has a single listener).

flokli:
@ncw sorry for the noise, but can you kick off CI once again?

flokli:
Poke - this should be ready now.

flokli:
I just stubled over https://github.com/rclone/rclone/pull/7392, which tried doing the same thing. Some feedback on what'd be needed to get either of the two accepted would be appreciated.

ncw:
Sorry for the delay on this @flokli 

Can you rebase it on master please and fixup any problems (might not be any but there have been lots of commits to master since this started).

flokli:
@ncw I rebased this, and successfully tested socket-activating a webdav and sftp server.

```
systemd-socket-activate -l 8088 -l 8089 --fdname=foo:bar -- ./rclone serve sftp :local:vfs/
systemd-socket-activate -l 2022 --fdname=foo:bar -- ./rclone serve sftp :local:vfs
```

… both from the root of the rclone checkout, after a `go build ./rclone.go`.

ncw (commented):
I think this code is looking nice - thank you.

> I successfully tested socket-activating a webdav and sftp server.
> 
> ```
> systemd-socket-activate -l 8088 -l 8089 --fdname=foo:bar -- ./rclone serve webdav :local:vfs/
> systemd-socket-activate -l 2022 --fdname=foo:bar -- ./rclone serve sftp :local:vfs
> ```
> 
> … both from the root of the rclone checkout, after a `go build ./rclone.go`.

Can you put these commands in the docs please? It is great to have some commands people can cut and paste to try the feature.

It would be nice also to have a test - do you think that is possible?

PS I think all the test failures are flakes so I've re-run them.

flokli:
> PS I think all the test failures are flakes so I've re-run them.

The windows lint is actually correct - will push a fix.

> Can you put these commands in the docs please? It is great to have some commands people can cut and paste to try the feature.

It really isn't how I'd expect people to run this in production usecases - where they'll probably configure systemd .service and .socket files. I can try to find a wording that makes sure this only makes sense for testing purposes.

---

Auxillary question - do you think we should roll this a bit differently, essentially switch to be socket-activated if we detect to be in such an environment (and have a CLI flag to disable it), but keep the existing listener syntax mostly?

ncw:
> > PS I think all the test failures are flakes so I've re-run them.
> 
> The windows lint is actually correct - will push a fix.

:+1: 
 
> > Can you put these commands in the docs please? It is great to have some commands people can cut and paste to try the feature.
> 
> It really isn't how I'd expect people to run this in production usecases - where they'll probably configure systemd .service and .socket files. I can try to find a wording that makes sure this only makes sense for testing purposes.

Great.

> Auxillary question - do you think we should roll this a bit differently, essentially switch to be socket-activated if we detect to be in such an environment (and have a CLI flag to disable it), but keep the existing listener syntax mostly?

How does this change the listener syntax? Do you mean where it ignores the command line args if it detects socket activation?

Not sure I understand the question (probably not enough coffee yet :-)

aanderse:
> It really isn't how I'd expect people to run this in production usecases - where they'll probably configure systemd .service and .socket files.

right, which is why an example `socket` and `service` file would probably be useful to some in documentation, right?

flokli:
> > It really isn't how I'd expect people to run this in production usecases - where they'll probably configure systemd .service and .socket files.
> 
> right, which is why an example `socket` and `service` file would probably be useful to some in documentation, right?

The docs already link to https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html, which links to a bunch of guides:

>  For more extensive descriptions see the "systemd for Developers" series: [Socket Activation](https://0pointer.de/blog/projects/socket-activation.html), [Socket Activation, part II](https://0pointer.de/blog/projects/socket-activation2.html), [Converting inetd Services](https://0pointer.de/blog/projects/inetd.html), [Socket Activated Internet Services and OS Containers](https://0pointer.de/blog/projects/socket-activated-containers.html).

There's plenty of explanation there, and I'm fairly certain people looking for this feature are already aware of how it works, or willing to do some research on it if they want it. Duplicating a small subset of examples into an individual rclone documentation item feels like a bit too much IMHO. This is the rclone documentation, we can mention we do support it, be helpful and link to some more upstream documentation, but we shouldn't try to document every other auxillary tool rclone can interface. We also don't describe how to use an sftp or http client - I see this in a similar league.

aanderse:
yeah i mean you and i find this topic immediately recognizable so it seems basic - i just wasn't sure this would be obvious to everyone


> I'm fairly certain people looking for this feature are already aware of how it works

actually that's a great point... while this won't be obvious to everyone of course it will likely be obvious to everyone looking for it

on that note... thank you for your contribution here @flokli! i was just looking for this feature 2 days ago and found this PR... somehow i wasn't surprised to see your name 😉

flokli:
@ncw friendly ping :-) is there anything left you want me to change here, or can we land this?

ncw (approved):
Sorry for the delay in review - been on vacation !

Let's merge this for v1.68

Thank you :-)

PR #7809, opened 2024-04-28
oauthutil: clear client secret if client ID is set

mikix:
When an external OAuth flow is being used (i.e. a client ID and an OAuth token are set in the config), a client secret should not be set. If one is, the server may reject a token refresh attempt.

But there's no way to clear out a backend's default client secret via configuration, since empty-string config values are ignored.

So instead, when a client ID is set, we should clear out any default client secret, since it wouldn't apply anyway.

No - I was just debugging why my OAuth tokens weren't being refreshed correctly when using the Drive backend, and it led me here.

I did file #7825 for tracking purposes though.

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -376,6 +376,9 @@ func overrideCredentials(name string, m configmap.Mapper, origConfig *oauth2.Con
 	ClientID, ok := m.Get(config.ConfigClientID)
 	if ok && ClientID != "" {
 		newConfig.ClientID = ClientID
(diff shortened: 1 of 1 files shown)


mikix:
For clarity on user impact of the bug I'm trying to fix, because that wasn't prominent in my description above:

If you pass Rclone a `CLIENT_ID` and a `TOKEN` today, Rclone fails to refresh the token once it expires (say, after an hour).

ncw (approved):
This looks like a great fix , thank you.

I'll merge that now :-)

PR #7820, opened 2024-05-02
Progress: add continuous stats printing

t-lo:

This change adds a new flag,`--stats--print-continuous`, which will cause progress stats to be printed continuously instead of a static block. Stats may be interleaved with log messages in that case.

I'm using a long-running (hours) rclone systemd job to regularly pull backups, with both `--progress` and `--stats-one-line`. This new option allows for checking progress via the rclone systemd unit's journal; before, output was mangled.

Not to my knowledge. A search did not come up with anything.

#### Checklist

- N/A I have added tests for all changes in this PR if appropriate.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -246,7 +246,7 @@ func Run(Retry bool, showStats bool, cmd *cobra.Command, f func() error) {
 		showStats = true
 	}
 	if ci.Progress {
-		stopStats = startProgress()
+		stopStats = startProgress(ci.StatsPrintContinuous)
 	} else if showStats {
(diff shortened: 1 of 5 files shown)


ncw:
The --progress output is designed for the terminal.

Maybe using --stats 1s without --progress would be enough to fix your problem?

t-lo:
Ah, I see. I did not realise one-line stats include progress by default. So `--stats=1s --stats-one-line --stats-log-level NOTICE` is close enough to what I'm looking for.

Closing this; thanks!

ncw:
Great - glad that works for you :-)

PR #7821, opened 2024-05-03
dropbox: add option to override root namespace

wfraser:

This lets you, for example, use shared folders without mounting them into your home namespace first, as long as you know their namespace ID.

(The --dropbox-shared-folders flag could thus be changed to not need to mount the shared folder first, but I'm not doing that here as it's a behavior change, who knows, maybe somebody relies on it.)

This also allows access to your Dropbox Backups hidden namespace once you figure out what its namespace ID is.

#### Checklist

### DIFF
--- a/backend/dropbox/dropbox.go
+++ b/backend/dropbox/dropbox.go
@@ -216,7 +216,10 @@ are supported.
 
 Note that we don't unmount the shared folder afterwards so the 
 --dropbox-shared-folders can be omitted after the first use of a particular 
-shared folder.`,
+shared folder.
+
+See also --dropbox-root-namespace for an alternative way to work with shared
+folders.`,
 			Default:  false,
 			Advanced: true,
 		}, {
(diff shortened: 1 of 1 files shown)


ncw (commented):
I think this change looks great thank you.

I'd just like to have a discussion about the naming.

In other backends we have called this feature like this

```
      --box-root-folder-id string          Fill in for rclone to use a non root folder as its starting point
      --drive-root-folder-id string        ID of the root folder
      --filefabric-root-folder-id string   ID of the root folder
      --onedrive-root-folder-id string     ID of the root folder
      --pcloud-root-folder-id string       Fill in for rclone to use a non root folder as its starting point (default "d0")
      --pikpak-root-folder-id string       ID of the root folder
      --sharefile-root-folder-id string    ID of the root folder
```

So I'm tempted to say we should do the same for the dropbox backend too and call this `--dropbox-root-folder-id` to fit in with the rest of rclone.

However dropbox have their very specific naming with namespaces so maybe it isn't a good idea. We could of course write lots of help explaining things.

What do you think?

I'll run the CI now too.

wfraser:
Thanks for the review!

I think calling it "folder ID" here would be more confusing, because it needs to specifically be a namespace ID and not a file/folder ID (which is a different kind of identifier).

ncw:
Thank you :-)

I'll merge this now.

2994dd42d75b0:
How to find namespace ID of Back in Dropbox?

PR #7840, opened 2024-05-13
Update faq.md

AtomicRobotMan0101:
trivial typo

#### Checklist

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -71,7 +71,7 @@ support partially uploading an object. You can't take an existing
 object, and change some bytes in the middle of it.
 
 It would be possible to make a sync system which stored binary diffs
-instead of whole objects like rclone does, but that would break the
+instead of whole objects like rsync does, but that would break the
 1:1 mapping of files on your hard disk to objects in the remote cloud
 storage system.
 
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -71,7 +71,7 @@ support partially uploading an object. You can't take an existing
 object, and change some bytes in the middle of it.
 
 It would be possible to make a sync system which stored binary diffs
-instead of whole objects like rsync does, but that would break the
+like rsync does, instead of whole objects, but that would break the
 1:1 mapping of files on your hard disk to objects in the remote cloud
 storage system.
 


docs/content/faq.md
```diff
@@ -71,7 +71,7 @@ support partially uploading an object. You can't take an existing
 object, and change some bytes in the middle of it.
 
 It would be possible to make a sync system which stored binary diffs
-instead of whole objects like rclone does, but that would break the
+instead of whole objects like rsync does, but that would break the
```
albertony, line 74: Hm.. I guess this depends on what you interpret the "like..." part to refer to, and probably the original was intended to be read with emphasis on "instead of the whole objects like rclone does". Maybe simply adding a comma would make that clearer?

> It would be possible to make a sync system which stored binary diffs`,`
> instead of whole objects like rclone does, but that would break the

albertony: Maybe I'm nitpicking here, and please know that your attempt at making the documentation clearer is appreciated. :)

In my opinion your changed text have a similar ambiguity as the original:

Original version:

> It would be possible to make a sync system which stored binary diffs instead of whole objects like rclone does, ...

Your version:

> It would be possible to make a sync system which stored binary diffs instead of whole objects like rsync does, ...

I think adding a comma or restructuring the sentence a bit is necessary to avoid confusion, e.g. one of these:

1. > It would be possible to make a sync system which stored binary diffs, instead of whole objects like rclone does, ...

2. > It would be possible to make a sync system which stored binary diffs, instead of whole objects, like rsync does, ...

3. > It would be possible to make a sync system which stored binary diffs like rsync does, instead of whole objects, ...

(I like 1 and 3 better than 2)

What do you think?

AtomicRobotMan0101: Clarity, as always. 

Ive been going over the code over the last few days and have a new appreciation of the project.

My (trivial) suggestion was only to aide understanding.

Three is excellent.

albertony: Great. If you want to change to that and push, I can merge it.

AtomicRobotMan0101 (commented):
a trivial amendment of the FAQ to aide in understanding.

albertony (approved):

PR #7865, opened 2024-05-24
webdav: introduce unix_socket_path

flokli:
This adds a new optional parameter to the backend, allowing to specify a path to a unix domain socket to connect to, instead the specified URL.

If the parameter is set, we use `fshttp.NewClientCustom` to modify the HTTP transport, to use a dialer connecting to the unix domain socket path specified for that backend.

The URL itself is still used for the rest of the HTTP client, allowing host and subpath to stay intact.

This allows using rclone with the webdav backend to connect to a WebDAV server provided at a Unix Domain socket:

```
RCLONE_WEBDAV_UNIX_SOCKET_PATH=/path/to/my.sock \
RCLONE_WEBDAV_URL=http://localhost \
rclone sync mydir :webdav:/somewhere
```

This allows using rclone with the webdav backend to connect to a WebDAV server provided at a Unix Domain socket.

https://forum.rclone.org/t/support-connecting-to-unix-sockets/46161

#### Checklist

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -159,7 +159,9 @@ Set to 0 to disable chunked uploading.
 			Help:     "Exclude ownCloud mounted storages",
 			Advanced: true,
 			Default:  false,
-		}},
+		},
+			fshttp.UnixSocketConfig,
+		},
 	})
 }
 
@@ -177,6 +179,7 @@ type Options struct {
 	ChunkSize          fs.SizeSuffix        `config:"nextcloud_chunk_size"`
 	ExcludeShares      bool                 `config:"owncloud_exclude_shares"`
 	ExcludeMounts      bool                 `config:"owncloud_exclude_mounts"`
+	UnixSocket         string               `config:"unix_socket"`
 }
 
 // Fs represents a remote webdav
@@ -458,7 +461,12 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		precision:   fs.ModTimeNotSupported,
 	}
 
-	client := fshttp.NewClient(ctx)
+	var client *http.Client
+	if opt.UnixSocket == "" {
+		client = fshttp.NewClient(ctx)
+	} else {
+		client = fshttp.NewClientWithUnixSocket(ctx, opt.UnixSocket)
+	}
 	if opt.Vendor == "sharepoint-ntlm" {
 		// Disable transparent HTTP/2 support as per https://golang.org/pkg/net/http/ ,
 		// otherwise any connection to IIS 10.0 fails with 'stream error: stream ID 39; HTTP_1_1_REQUIRED'
--- a/cmd/rc/rc.go
+++ b/cmd/rc/rc.go
@@ -23,14 +23,15 @@ import (
 )
 
 var (
-	noOutput  = false
-	url       = "http://localhost:5572/"
-	jsonInput = ""
-	authUser  = ""
-	authPass  = ""
-	loopback  = false
-	options   []string
-	arguments []string
+	noOutput   = false
+	url        = "http://localhost:5572/"
+	unixSocket = ""
+	jsonInput  = ""
+	authUser   = ""
+	authPass   = ""
+	loopback   = false
+	options    []string
+	arguments  []string
 )
 
 func init() {
@@ -38,6 +39,7 @@ func init() {
 	cmdFlags := commandDefinition.Flags()
 	flags.BoolVarP(cmdFlags, &noOutput, "no-output", "", noOutput, "If set, don't output the JSON result", "")
 	flags.StringVarP(cmdFlags, &url, "url", "", url, "URL to connect to rclone remote control", "")
+	flags.StringVarP(cmdFlags, &unixSocket, "unix-socket", "", unixSocket, "Path to a unix domain socket to dial to, instead of opening a TCP connection directly", "")
 	flags.StringVarP(cmdFlags, &jsonInput, "json", "", jsonInput, "Input JSON - use instead of key=value args", "")
 	flags.StringVarP(cmdFlags, &authUser, "user", "", "", "Username to use to rclone remote control", "")
 	flags.StringVarP(cmdFlags, &authPass, "pass", "", "", "Password to use to connect to rclone remote control", "")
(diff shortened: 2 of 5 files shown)


ncw (commented):
Rclone's http servers support the syntax `unix:///tmp/socket` (eg rclone serve http/webdav)

Maybe we should get `fs/fshttp` to support this syntax too and then any backend which uses it will automatically get socket support?

flokli:
> Rclone's http servers support the syntax `unix:///tmp/socket` (eg rclone serve http/webdav)
> 
> Maybe we should get `fs/fshttp` to support this syntax too and then any backend which uses it will automatically get socket support?

There's a distinction between the endpoint to use to connect to a resource, and the file path to use inside of there (what we call subpath).

In the case of http, both are expressed in the same URL, `https://myserver/foo/bar` does open a HTTPS connection to the host `myserver`, and uses `foo/bar` as subpath.

For unix domain sockets, so however cannot have both - assuming you have a connection to the same server, at a unix domain socket at `/path/to/socket`, and you use `unix:///path/to/socket`, you loose the ability to express a subpath (like `foo/bar`).

That's one of the reasons why I went with a separate parameter, similar to how `curl` does do it.

ncw:
That is a good point!

I would like to generalise this if we can though.

Once idea I saw was having urls of the form `unix:///path/to/socket|http://localhost/path` or maybe `unix:///path/to/socket#/path`. Not sure we'd get any of those to work with Go's machinery though :-(

I'd like this to work on the http and the webdav backends so we could put all the definitions of everything in `fs/fshttp` so a backend would include the config object, a line in Options and then call fshttp.NewClientWithSocket(ctx, f.opt.UnixSocket). 

What do you think?

PS `curl` calls this `--unix-socket` so we should call the parameter just `unix_socket`.

flokli:
Curl calls it --unix-socket, yes. Happy to rename the argument like this too.

> Once idea I saw was having urls of the form unix:///path/to/socket|http://localhost/path or maybe unix:///path/to/socket#/path. Not sure we'd get any of those to work with Go's machinery though :-(

Curl opted against trying to somehow encode both unix socket address and subpath in the same URL. We shouldn't try to define our own URL standard for this either. I'd prefer to stay close to their behavior too - keep a separate parameter. Do you agree with this?

> I'd like this to work on the http and the webdav backends so we could put all the definitions of everything in fs/fshttp so a backend would include the config object, a line in Options and then call fshttp.NewClientWithSocket(ctx, f.opt.UnixSocket).

I'll wrap my head around this soon, assuming we're on the same page regarding the user-facing API.

backend/webdav/webdav.go
```diff
@@ -159,6 +160,11 @@ Set to 0 to disable chunked uploading.
 			Help:     "Exclude ownCloud mounted storages",
 			Advanced: true,
 			Default:  false,
+		}, {
```
ncw, line 163: Make this block a `var UnixSocketPathConfig = fs.Option{...}` in `fs/fshttp` and replace it here with just `fshttp.UnixSocketPathConfig`

backend/webdav/webdav.go
```diff
@@ -458,7 +465,17 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		precision:   fs.ModTimeNotSupported,
 	}
 
-	client := fshttp.NewClient(ctx)
+	var client *http.Client
```
ncw, line 468: Move this code to `fs/fshttp` as `NewClientWithUnixSocket` maybe and call it here `fshttp.NewClientWithUnixSocket(ctx, opt.UnixSocket)`

backend/webdav/webdav.go
```diff
@@ -177,6 +183,7 @@ type Options struct {
 	ChunkSize          fs.SizeSuffix        `config:"nextcloud_chunk_size"`
 	ExcludeShares      bool                 `config:"owncloud_exclude_shares"`
 	ExcludeMounts      bool                 `config:"owncloud_exclude_mounts"`
+	UnixSocketPath     string               `config:"unix_socket_path"`
```
ncw, line 186: Rename as `UnixSocket` as discussed

fs/config.go
```diff
@@ -635,6 +635,7 @@ type ConfigInfo struct {
 	Inplace                    bool              `config:"inplace"`      // Download directly to destination file instead of atomic download to temp/rename
 	PartialSuffix              string            `config:"partial_suffix"`
 	MetadataMapper             SpaceSepList      `config:"metadata_mapper"`
+	UnixSocketPath             string            `config:"unix_socket_path"` // Optional unix socket to connect to
```
ncw, line 638: I don't think you want this global config so remove this

ncw (commented):
Yes I think we are on the same page regarding the user facing API. (I experimented with the URL way and the Go machinery got in the way so I think that change would be very invasive).

I've put some guidance on how I think the refactor should go inline.

It would be nice to have a test for this feature - maybe we can do one in `fs/fshttp`?

ncw:
I just remembered something...

We have support for unix sockets for `rclone rcd` (via the http server framework) so we should have support for `rclone rc` also to access it.

Did it get done? I don't know it is not in the docs so I don't think so. It certainly doesn't work.

So we should probably add `--unix-socket` as a parameter to `rclone rc` (in `cmd/rc`) and call the new `fshttp.ClientWithUnixSocket` there too.

backend/webdav/webdav.go
flokli: Done

backend/webdav/webdav.go
flokli: Done

backend/webdav/webdav.go
flokli: Done

fs/config.go
flokli: Done

flokli:
the linter step is complaining about files I didn't touch, so seems unrelated to this PR. This is ready for review!

ncw:
> the linter step is complaining about files I didn't touch, so seems unrelated to this PR. This is ready for review!

The linter got upgraded and found some new things! You'll need to rebase on master to fix the lint problems.

flokli:
> > the linter step is complaining about files I didn't touch, so seems unrelated to this PR. This is ready for review!
> 
> The linter got upgraded and found some new things! You'll need to rebase on master to fix the lint problems.

Done!

ncw (approved):
This looks very good - thank you :-)

I'll merge now.

ncw:
I just squashed that into two commits and gave the docs a tweak - I'll merge when the CI goes green. Thank you :-)

PR #7866, opened 2024-05-24
tree-wide: replace /bin/bash with /usr/bin/env bash

flokli:
The latter is more portable, while the former only works on systems where /bin/bash exists (or is symlinked appropriately).

Make more of `go test` work on systems where `/bin/bash` does not exist, like NixOS.

No, but I ran into it a bunch of times and decided to fix it :-)

#### Checklist


ncw (approved):
Thank you - will merge now :-)

PR #7877, opened 2024-05-29
Adding a command line option to use `FILE_FLAG_BACKUP_SEMANTICS` on Windows

chamilton-ccn:

This change adds a command line option: `--use-backup-semantics` that causes `rclone` to include `syscall.FILE_FLAG_BACKUP_SEMANTICS` on Windows OSs when making calls to `syscall.CreateFile`. This behavior is desired when `rclone` is running within a security context that has `SeBackupPrivilege` and/or `SeRestorePrivilege` token privileges __enabled__. _Note:_ simply including this flag ___does not___ enable those privileges for you, but chances are, if you're interested in this feature you already know how to do that :smile: . Without this feature, `rclone` cannot properly leverage special security groups such as `Backup Operators` who possess the aforementioned token privileges.

[Relevant forum post](https://forum.rclone.org/t/rclone-sebackupprivilege-file-flag-backup-semantics/45339)

#### Checklist


ncw:
This looks like a great fix for advanced users.

Having to add an additional parameter to file.Open is a bit ugly - this is supposed to mirror os.Open

What are they consequences of adding this flag to all the opens rclone does? Would that be bad? 

I note that you made it a global option rather than a local backend option which implies it should affect everything.

If it is a global option then it needs docs in docs/content/docs.md but I think this should be a local backend option.

What do you think?

ncw:
@chamilton-ccn wrote:
> > What are they consequences of adding this flag to all the opens rclone does? Would that be bad?
> 
> This is an excellent question -- one I've thought about a lot since ~March. **TL;DR:** I don't think there are any drawbacks to including this flag on every call to `CreateFile`.

Thanks for investigating that.

> > I note that you made it a global option rather than a local backend option which implies it should affect everything.
> 
> Correct. If this option is going to be used at all, it is implied that the security context running the backup might not have explicit access to every file it might need to back up, so this flag needs to be present in every call to `CreateFile`.

By affect everything I meant the other users of `lib/file` which aren't the local backend. The most obvious one is the caching that happens in the VFS which is used when you do `rclone mount --vfs-cache-mode full/writes`.

From what you said I don't think it matters whether we add the flag here or not.

> > If it is a global option then it needs docs in docs/content/docs.md but I think this should be a local backend option.
> 
> Sure, however if we include `FILE_FLAG_BACKUP_SEMANTICS` in every call to `CreateFile` (Windows-only, obviously) then this question and the previous one are moot.

With the small caveat that the local backend isn't the only user of `lib/file`

> > What do you think?
> 
> I think that including `FILE_FLAG_BACKUP_SEMANTICS` in every call to `CreateFile` whether it is needed or not **is the cleanest possible solution**. That avoids having to change the function signature of `file.Open`; it avoids having to add another command line option; and it seamlessly enables this behavior (i.e., it'll "just work") whenever/wherever appropriate. Hardcoding this flag was my [original attempt](https://github.com/ccn-support-services/rclone/commit/51021b33f980a99891df778a13f7d4aeb749218d) at using it with `rclone`, and that modification was tested, then ultimately used in production for at least two months without issue. So, if you're fine with that, I'll close this PR, implement the alternative, then submit another PR.

Yes I'm fine with that :-)

You can force push this PR or close this one and make a new one, whichever you prefer.

We should write this in the docs too somewhere but I'm struggling to find an appropriate place! Perhaps a note in the changelog will have to do (which I'll write when this hits a release).

> Also, thanks again for all your work on this project. `rclone` is an amazing tool and I'm happy to contribute in any way that I can.

Very happy to have your expertise in this area :-)

Also maybe you could persuade your company that they would like a [support contract](https://rclone.com/support) which can help you get answers quicker and keeps the rclone project sustainable?

chamilton-ccn:
> Thanks for investigating that.

Of course! Not to beat a dead horse but initially (months ago) I wasn't sure if including the flag when it's not needed would cause trouble -- in fact, I was willing to err on the side of "if it's not needed, and you include it, it _probably will_ cause trouble", however after a great deal of research, I realized that it's impossible to use the flag for its intended purpose without potentially including it when it isn't needed. As far as I can tell, it **only** has an impact in situations where:

-  The existing security context has `SeBackupPrivilege` (required for read operations) and/or `SeRestorePrivilege` (required for write operations) delegated ***and*** enabled 
- ... and the call to `CreateFile` would _otherwise_ result in an access denied error.

Consider the fact that even the most locked-down account in any Windows environment _still_ has read access (at a minimum) to at least _some_ files/directories, and backup software shouldn't make assumptions about the types/locations/etc. of files its users might decide to back up, so it's possible -- even _probable_ -- that any backup job might include files that the user normally has access to, and attempting to backup those files means the flag would be included when it's not needed. Nobody in their right mind would want that scenario to be a problem :smile: So the only reasonable conclusion I can come to is that this flag is helpful when the requirements for its use are met, but may as well not exist otherwise.

> By affect everything I meant the other users of lib/file which aren't the local backend. The most obvious one is the caching that happens in the VFS which is used when you do rclone mount --vfs-cache-mode full/writes.
>
> From what you said I don't think it matters whether we add the flag here or not.

Ah, got it! Yeah, I don't see any reason at all why this should be an issue.

> You can force push this PR or close this one and make a new one, whichever you prefer.

It's such a small change that the latter, from a fresh branch, is probably best. I'll have that done shortly!

> We should write this in the docs too somewhere but I'm struggling to find an appropriate place! Perhaps a note in the changelog will have to do (which I'll write when this hits a release).

Maybe even include a reference to this discussion? I think that would provide some much-needed context for those who aren't terribly familiar with the subject matter. 

> Very happy to have your expertise in this area :-)
>
> Also maybe you could persuade your company that they would like a support contract which can help you get answers quicker and keeps the rclone project sustainable?

More than happy to contribute! I'll work on that, but in the immediate term, I've made a donation to the project.

ncw:
> > By affect everything I meant the other users of lib/file which aren't the local backend. The most obvious one is the caching that happens in the VFS which is used when you do rclone mount --vfs-cache-mode full/writes.
> > From what you said I don't think it matters whether we add the flag here or not.
> 
> Ah, got it! Yeah, I don't see any reason at all why this should be an issue.

Great

> > You can force push this PR or close this one and make a new one, whichever you prefer.
> 
> It's such a small change that the latter, from a fresh branch, is probably best. I'll have that done shortly!

I ran the CI on the new PR - I'll merge when it goes green :-)

> > Very happy to have your expertise in this area :-)
> > Also maybe you could persuade your company that they would like a support contract which can help you get answers quicker and keeps the rclone project sustainable?
> 
> More than happy to contribute! I'll work on that, but in the immediate term, I've made a donation to the project.

Thank you and thank you :-)

chamilton-ccn:
> I ran the CI on the new PR - I'll merge when it goes green :-)

The `mac_amd64` build is failing and I'm pretty sure it's not related to anything in the PR, but related to [this](https://github.com/golang/go/issues/61229#issuecomment-1988965927) instead.

Issue #6038, opened 2022-03-15
Flag --use-json-log does not working as expected 


fgendorf:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/non-jsoned-output-when-using-use-json-log/27299/9

#### What is the problem you are having with rclone?
using follow command does not generating logfile in json format if fail, the log is plain text:
`rclone lsjson server: --low-level-retries 1 --contimeout 6s --use-json-log --log-file err.log` 

#### What is your rclone version (output from `rclone version`)
rclone v1.57.0
- os/version: ubuntu 20.04 (64 bit)
- os/kernel: 5.4.0-100-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17.2
- go/linking: static
- go/tags: none

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Linux Ubuntu 20.04

#### Which cloud storage system are you using? (e.g. Google Drive)
SFTP

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
with -vv the log has json, but the last line is not a json
```
{"level":"debug","msg":"Version \"v1.57.0\" starting with parameters [\"rclone\" \"lsjson\" \"server:\" \"--low-level-retries\" \"1\" \"--contimeout\" \"6s\" \"--use-json-log\" \"--log-file\" \"err.log\" \"-vv\"]","object":"rclone","objectType":"string","source":"cmd/cmd.go:408","time":"2022-03-15T08:36:52.234974-04:00"}
{"level":"debug","msg":"Creating backend with remote \"server:\"","source":"fs/newfs.go:28","time":"2022-03-15T08:36:52.235601-04:00"}
{"level":"debug","msg":"Using config file from \"/home/localroot/.config/rclone/rclone.conf\"","source":"config/config.go:353","time":"2022-03-15T08:36:52.235786-04:00"}
{"level":"debug","msg":"low level retry 1/1 (error couldn't connect SSH: dial tcp: lookup REDACT: no such host)","object":"pacer","objectType":"string","source":"fs/pacer.go:90","time":"2022-03-15T08:36:52.236656-04:00"}
{"level":"debug","msg":"Rate limited, increasing sleep to 200ms","object":"pacer","objectType":"string","source":"fs/pacer.go:46","time":"2022-03-15T08:36:52.23675-04:00"}
[...]

PR #7884, opened 2024-06-01
Gocritic lint

albertony:

Basic gocritic linting via golangci-lint was enabled in https://github.com/rclone/rclone/pull/8009.

This continues on that PR with additional features, primarily enabling custom linting rules with ruleguard via gocritic.

It then adds custom linting rules for log statements, with suggestions to use `fs.` instead of `log.`, which is required for `--use-json-log` to properly make log output in JSON format. This fixes https://github.com/rclone/rclone/issues/6038.

Note:
- After ruleguard files are changed, lint cache will not be automatically invalidated. GitHub Actions cache must be manually deleted, unless waiting for the transition to a new year-week number because then it changes the cache key and therefore will the next lint will run without cache. Running locally, command `golangci-lint cache clean` can be used to test changes to ruleguard files.
  > //Note that when used from golangci-lint (using the gocritic linter configured
  > // with the ruleguard check), because rule files are not handled by
  > // golangci-lint itself, changes will not invalidate the golangci-lint cache,
  > // and you must manually clean to cache (golangci-lint cache clean) for them to
  > // be considered, as explained here:
  > // https://www.quasilyte.dev/blog/post/ruleguard/#using-from-the-golangci-lint

#### Checklist

### DIFF
--- a/bin/not-in-stable.go
+++ b/bin/not-in-stable.go
@@ -29,7 +29,7 @@ func readCommits(from, to string) (logMap map[string]string, logs []string) {
 	cmd := exec.Command("git", "log", "--oneline", from+".."+to)
 	out, err := cmd.Output()
 	if err != nil {
-		log.Fatalf("failed to run git log %s: %v", from+".."+to, err)
+		log.Fatalf("failed to run git log %s: %v", from+".."+to, err) //nolint:gocritic // Don't include gocritic when running golangci-lint to avoid ruleguard suggesting fs. intead of log.
 	}
 	logMap = map[string]string{}
 	logs = []string{}
@@ -39,7 +39,7 @@ func readCommits(from, to string) (logMap map[string]string, logs []string) {
 		}
 		match := logRe.FindSubmatch(line)
 		if match == nil {
-			log.Fatalf("failed to parse line: %q", line)
+			log.Fatalf("failed to parse line: %q", line) //nolint:gocritic // Don't include gocritic when running golangci-lint to avoid ruleguard suggesting fs. intead of log.
 		}
 		var hash, logMessage = string(match[1]), string(match[2])
 		logMap[logMessage] = hash
@@ -52,12 +52,12 @@ func main() {
 	flag.Parse()
 	args := flag.Args()
 	if len(args) != 0 {
-		log.Fatalf("Syntax: %s", os.Args[0])
+		log.Fatalf("Syntax: %s", os.Args[0]) //nolint:gocritic // Don't include gocritic when running golangci-lint to avoid ruleguard suggesting fs. intead of log.
 	}
 	// v1.54.0
 	versionBytes, err := os.ReadFile("VERSION")
 	if err != nil {
-		log.Fatalf("Failed to read version: %v", err)
+		log.Fatalf("Failed to read version: %v", err) //nolint:gocritic // Don't include gocritic when running golangci-lint to avoid ruleguard suggesting fs. intead of log.
 	}
 	if versionBytes[0] == 'v' {
 		versionBytes = versionBytes[1:]
@@ -65,7 +65,7 @@ func main() {
 	versionBytes = bytes.TrimSpace(versionBytes)
 	semver := semver.New(string(versionBytes))
 	stable := fmt.Sprintf("v%d.%d", semver.Major, semver.Minor-1)
-	log.Printf("Finding commits in %v not in stable %s", semver, stable)
+	log.Printf("Finding commits in %v not in stable %s", semver, stable) //nolint:gocritic // Don't include gocritic when running golangci-lint to avoid ruleguard suggesting fs. intead of log.
 	masterMap, masterLogs := readCommits(stable+".0", "master")
 	stableMap, _ := readCommits(stable+".0", stable+"-stable")
 	for _, logMessage := range masterLogs {
--- a/bin/rules.go
+++ b/bin/rules.go
@@ -1,0 +1,51 @@
+// Ruleguard file implementing custom linting rules.
+//
+// Note that when used from golangci-lint (using the gocritic linter configured
+// with the ruleguard check), because rule files are not handled by
+// golangci-lint itself, changes will not invalidate the golangci-lint cache,
+// and you must manually clean to cache (golangci-lint cache clean) for them to
+// be considered, as explained here:
+// https://www.quasilyte.dev/blog/post/ruleguard/#using-from-the-golangci-lint
+//
+// Note that this file is ignored from build with a build constraint, but using
+// a different than "ignore" to avoid go mod tidy making dsl an indirect
+// dependency, as explained here:
+// https://github.com/quasilyte/go-ruleguard?tab=readme-ov-file#troubleshooting
+
+//go:build ruleguard
+// +build ruleguard
+
+// Package gorules implementing custom linting rules using ruleguard
+package gorules
+
+import "github.com/quasilyte/go-ruleguard/dsl"
+
+// Suggest rewriting "log.(Print|Fatal|Panic)(f|ln)?" to
+// "fs.(Printf|Fatalf|Panicf)", and do it if running golangci-lint with
+// argument --fix. The suggestion wraps a single non-string single argument or
+// variadic arguments in fmt.Sprint to be compatible with format string
+// argument of fs functions.
+//
+// Caveats:
+//   - After applying the suggestions, imports may have to be fixed manually,
+//     removing unused "log", adding "github.com/rclone/rclone/fs" and "fmt",
+//     and if there was a variable named "fs" or "fmt" in the scope the name
+//     clash must be fixed.
+//   - Suggested code is incorrect when within fs package itself, due to the
+//     "fs."" prefix. Could handle it using condition
+//     ".Where(m.File().PkgPath.Matches(`github.com/rclone/rclone/fs`))"
+//     but not sure how to avoid duplicating all checks with and without this
+//     condition so haven't bothered yet.
+func useFsLog(m dsl.Matcher) {
+	m.Match(`log.Print($x)`, `log.Println($x)`).Where(m["x"].Type.Is(`string`)).Suggest(`fs.Logf(nil, $x)`)
+	m.Match(`log.Print($*args)`, `log.Println($*args)`).Suggest(`fs.Logf(nil, fmt.Sprint($args))`)
+	m.Match(`log.Printf($*args)`).Suggest(`fs.Logf(nil, $args)`)
+
+	m.Match(`log.Fatal($x)`, `log.Fatalln($x)`).Where(m["x"].Type.Is(`string`)).Suggest(`fs.Fatalf(nil, $x)`)
+	m.Match(`log.Fatal($*args)`, `log.Fatalln($*args)`).Suggest(`fs.Fatalf(nil, fmt.Sprint($args))`)
+	m.Match(`log.Fatalf($*args)`).Suggest(`fs.Fatalf(nil, $args)`)
+
+	m.Match(`log.Panic($x)`, `log.Panicln($x)`).Where(m["x"].Type.Is(`string`)).Suggest(`fs.Panicf(nil, $x)`)
+	m.Match(`log.Panic($*args)`, `log.Panicln($*args)`).Suggest(`fs.Panicf(nil, fmt.Sprint($args))`)
+	m.Match(`log.Panicf($*args)`).Suggest(`fs.Panicf(nil, $args)`)
+}
--- a/bin/rules.go
+++ b/bin/rules.go
@@ -37,15 +37,15 @@ import "github.com/quasilyte/go-ruleguard/dsl"
 //     but not sure how to avoid duplicating all checks with and without this
 //     condition so haven't bothered yet.
 func useFsLog(m dsl.Matcher) {
-	m.Match(`log.Print($x)`, `log.Println($x)`).Where(m["x"].Type.Is(`string`)).Suggest(`fs.Logf(nil, $x)`)
-	m.Match(`log.Print($*args)`, `log.Println($*args)`).Suggest(`fs.Logf(nil, fmt.Sprint($args))`)
+	m.Match(`log.Print($x)`, `log.Println($x)`).Where(m["x"].Type.Is(`string`)).Suggest(`fs.Log(nil, $x)`)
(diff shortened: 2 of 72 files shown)


cmd/serve/webdav/webdav.go
```diff
@@ -243,19 +243,19 @@ func newWebDAV(ctx context.Context, f fs.Fs, opt *Options) (w *WebDAV, err error
 }
 
 // Gets the VFS in use for this request
-func (w *WebDAV) getVFS(ctx context.Context) (VFS *vfs.VFS, err error) {
+func (w *WebDAV) getVFS(ctx context.Context) (vf *vfs.VFS, err error) {
```
ncw, line 246: I don't like these VFS -> vf subs

backend/netstorage/netstorage.go
```diff
@@ -858,10 +858,10 @@ func shouldRetry(ctx context.Context, resp *http.Response, err error) (bool, err
 
 // callBackend calls NetStorage API using either rest.Call or rest.CallXML function,
 // depending on whether the response is required
-func (f *Fs) callBackend(ctx context.Context, URL, method, actionHeader string, noResponse bool, response interface{}, options []fs.OpenOption) (io.ReadCloser, error) {
+func (f *Fs) callBackend(ctx context.Context, endpoint, method, actionHeader string, noResponse bool, response interface{}, options []fs.OpenOption) (io.ReadCloser, error) {
```
ncw, line 861: I think URL really is a better name in this package

bin/rules.go
```diff
@@ -0,0 +1,30 @@
...
+//go:build ruleguard
+// +build ruleguard
+
+// Package gorules implementing custom linting rules using ruleguard
+package gorules
+
+import "github.com/quasilyte/go-ruleguard/dsl"
+
+func useFsLog(m dsl.Matcher) {
+	m.Match(`log.Fatal($*args)`).Suggest(`fs.Fatal($args)`)
```
ncw, line 24: These suggestions should probably say

    Suggest(`fs.Fatal(nil, $args)`)

ncw (commented):
Thanks for doing this :-)

I think gocritic makes good suggestions by-and-large.

Maybe we should disable the `captlocal` linter check as that seemed to create the most noise. URL is what they are actually called "Uniform Resource Locator" and "VFS" is "Virtual Filing System" so unless we can special case those I think we should drop that rule.

The other rules looked good and it found a bug and the code was easier to read at the end.

It would be good to have a framework where we can make our own tests - rclone is getting to be a big project now and linting for our rules seems like a good idea (like the log.Fatal -> fs.Fatal).

albertony:
> I think gocritic makes good suggestions by-and-large.

Yes, I think so too. I'm also considering looking more into its performance checks (https://go-critic.com/overview.html#checkers-from-the-performance-group), disabled by default, but seem interesting (at least educational for me to look into what they report).

Enabling more and more linters with golangci-lint must result in some considerable amount of overlapping checks, but I don't see it as easy to verify it properly (without creating a large suite of code with representative issues to test the linters on), so as long as the time to run them is acceptable I guess its ok..

> Maybe we should disable the `captlocal` linter check as that seemed to create the most noise. URL is what they are actually called "Uniform Resource Locator" and "VFS" is "Virtual Filing System" so unless we can special case those I think we should drop that rule.

I knew I didn't find a good alternative to those two, and because of that I did those changes in a separate commit.. 😌 However, I must say I find it strange that it is almost exclusively the netstorage backend code that *must* use the `URL` name in variables. Everywhere else we have coped fine without it. Could we alias the package "net/url" instead, e.g. "gourl" (netstorage already has aliased hash package as "gohash"), and then use lowercase `url` as variable names?

I picked "vf" as name of VFS variables simply because fs.Fs variables often is "f".

> It would be good to have a framework where we can make our own tests - rclone is getting to be a big project now and linting for our rules seems like a good idea (like the log.Fatal -> fs.Fatal).

Yes. I've only scratched the surface of it, but I think it has potential to be very useful. The ruleguard syntax seems very neat yet extremely powerful for making various custom lint checks! I just put my extremely simple ruleguard implementation in a file in the existing bin folder for now, but we could consider a separate top-level folder for such files if we expect to make more use of it?

fs/operations/operations.go
```diff
@@ -1264,12 +1264,10 @@ func Cat(ctx context.Context, f fs.Fs, w io.Writer, offset, count int64, sep []b
 			err = fs.CountError(err)
 			fs.Errorf(o, "Failed to send to output: %v", err)
 		}
-		if len(sep) >= 0 {
```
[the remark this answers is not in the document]
albertony: I see - pushed an update where I do that. However I don't think it makes a difference: Original code and my first change (removing the if) only leads to it possibly writing an empty byte array, which does nothing but is not an error either. If I'm not mistaken...

bin/rules.go
albertony: Of course. Fixed it.

albertony:
Rebased and fixed conflicts

ncw:
Let's try to get the fixes and the enabling of the linter merged otherwise we'll be in rebase hell forever!

Can we
1. disable the `captlocal` linter and drop the corresponding commits
2. rebase (again - sorry!)
3. Merge up to and including "enable custom linting rules with ruleguard via gocritic"

This might involve making a new PR for 1-3 and then rebasing this one on master it is merged.

> I improved the rule implementation some. If I'm not mistaken, the intention was to replace calls to any of the standard logging functions from log package with our own in fs package? And also I've tweaked the suggestion so that it gives a working suggestion for `Print` and `Prinln` variants, not just `Printf`.

Nice :-)

> In other words:
> 
> ```
> log.(Print|Fatal|Panic)(f|ln)?` -> `fs.(Printf|Fatalf|Panicf)
> ```

I think `log.Print*` should become `fs.Logf` which is the direct equivalent of `log.Print*` and I don't think we need the `fs.Printf` function - that is bound to confuse someone sooner or later! Otherwise exactly what I had in mind

```
log.(Print|Fatal|Panic)(f|ln)?\(` -> `fs.(Logf|Fatalf|Panicf)\(nil, `
```

> I also pushed commits implementing the missing fs functions, to be able to test the lint suggestions, e.g. execute `golangci-lint run --fix`

I was not aware of that option. Would have saved me a lot of time a couple of days ago when installing golangci-lint 1.60.1!

> and see if the result compiles. But not sure if this matches what you had in mind, in the issue that triggered this.

I think that sounds good. The end goal being that if the user is using `--use-json-log` then all the log output is in JSON format. (That probably doesn't need to include log.Panicf!)

> This is still a draft, and there is something with regards to caching that I need to considered (added a todo in the pr description to remember it)

Doesn't golangci-lint do caching for us?

albertony:
> This might involve making a new PR for 1-3

Done: https://github.com/rclone/rclone/pull/8009

ncw:
> Rebased this on branch gocritic-lint-basic which is basis of #8009, for now. When that PR is merged this will be rebased on master leaving only the additional - currently:

Merged now!

> ```
> fs: add log Printf, Fatalf and Panicf
> fs: refactor base log method name for improved consistency
> fs: refactor log statements to use common helper
> build: enable custom linting rules with ruleguard via gocritic
> ```
> 
> I also did this change:
> 
> > I think `log.Print*` should become `fs.Logf`
> 
> Regarding caching: The issue is that if the ruleguard rules change, golangci-lint will not detect this and may still use its existing cache and therefore not consider the changed rules. When running locally one have to run `golangci-lint cache clean` to make sure they are considered. From GitHub Actions it is not that straight forward, although those with privileges can manually delete cache entries from the Actions page, however the cache key contains year-week number so once a week the lint run will create a new cache and this will solve itself. Maybe this is good enough?

Perhaps we need a note to that effect in the ruleguard code so we see it when updating it and can click the clear cache button. Then that is probably OK.

albertony:
> Perhaps we need a note to that effect in the ruleguard code so we see it when updating it and can click the clear cache button. Then that is probably OK.

https://github.com/rclone/rclone/blob/a6f456ec4e34b82c71b4b1a464643398eae720b8/bin/rules.go#L3-L8

> Merged now!

Great.
I have rebased.

This PR now have many (more in log than in "Files changed" tab) lint issues, under "unchanged files with check annotations", such as:

> ruleguard: suggestion: fs.Fatalf(nil, "Unable to decode %q: %s", fn, err) (gocritic)

I've intentionally left these in for now, but if you think the suggestions now are good I can add a commit fixing these - let's try `golangci-lint run --fix` :)

ncw (approved):
This is looking great!

The ruleguard fixes all look like they will work, but I think some of them will trigger other lint warnings. 

Yes give the auto fix a try :-)

I guess you are intending to add the subsequent commits here so we don't merge a lint apocalypse ;-)

albertony:
> The ruleguard fixes all look like they will work, but I think some of them will trigger other lint warnings.
> 
> Yes give the auto fix a try :-)

Done. The autofix worked really well to apply the suggestions, however some additional manual work was needed to update imports (maybe that can be done with a simple command as well, I just triggered on each file the on-save auto cleanup that VSCode does and it fixed most of it). It is documented here:

https://github.com/rclone/rclone/blob/a6f456ec4e34b82c71b4b1a464643398eae720b8/bin/rules.go#L29-L38

After fixing the imports to make it compile, there were no additional lint issues - in my local run of golangci-lint for Windows, but there could be for other platforms - I'll run that on GitHub.

> I guess you are intending to add the subsequent commits here so we don't merge a lint apocalypse ;-)

Good point!

⏬ 

Added this as a single commit, noted that it fixes #6038 (I'm assuming that is correct), and pushed to let it run through all checks...

fs/log/redirect_stderr_windows.go
```diff
@@ -34,6 +35,6 @@ func setStdHandle(stdhandle int32, handle syscall.Handle) error {
 func redirectStderr(f *os.File) {
 	err := setStdHandle(syscall.STD_ERROR_HANDLE, syscall.Handle(f.Fd()))
 	if err != nil {
-		log.Fatalf("Failed to redirect stderr to file: %v", err)
+		fs.Fatalf(nil, "Failed to redirect stderr to file: %v", err)
```
albertony, line 38: @ncw Is it correct to do it here?

bin/not-in-stable.go
```diff
@@ -5,12 +5,12 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"regexp"
 
 	"github.com/coreos/go-semver/semver"
+	"github.com/rclone/rclone/fs"
```
albertony, line 13: @ncw Maybe it will not be correct to use fs. in this file, so it needs to be excluded, or?

albertony:
Hm, the linting now failed with a lot of issues from govet of the type:

> lib\http\template.go:71:18: printf: non-constant format string in call to github.com/rclone/rclone/fs.Fatalf (govet)
>                fs.Fatalf(nil, fmt.Sprint("Fatal error executing template", err))

I didn't get those locally before, but I may not have been on latest version of golangci-lint because after I ran go install just now and re-ran I get the same issues locally as well. So this is probably related to a change in golangci-lint version [v1.60.1](https://github.com/golangci/golangci-lint/releases/tag/v1.60.1).

ncw:
> I didn't get those locally before, but I may not have been on latest version of golangci-lint because after I ran go install just now and re-ran I get the same issues locally as well. So this is probably related to a change in golangci-lint version [v1.60.1](https://github.com/golangci/golangci-lint/releases/tag/v1.60.1).

Yes I fixed a load of those recently for golangci lint v1.60.1 and that's the error I thought would happen, I guess having fixed a load of them recently!

fs/log/redirect_stderr_windows.go
ncw: Hmm, not sure. Does it cause a stack overflow due to unbounded recursion? If not then the change is good. You might need to simulate a failure to test.

bin/not-in-stable.go
ncw: Yes probably exclude!

albertony:
> Yes I fixed a load of those recently for golangci lint v1.60.1 and that's the error I thought would happen, I guess having fixed a load of them recently!

I see! What is the best fix then? Is it to add alternative non-format variants of all the functions (Log in addition to Logf, etc)?  I have pushed a version with this (not entirely finished, naming things etc..). Then I updated the custom lint rules accordingly, reverted back the lint fix and re-ran the lint. This time the autofix fixed everything without creating new lint issues. These changes are in the last 3 commits, rest is same as before.

fs/log/redirect_stderr_windows.go
albertony: It seems to work fine. Commenting out the if condition, running with `--log-file` option, the message "Failed to redirect stderr to file" ends up in the file. No stack overflow.

ncw:
> I see! What is the best fix then? Is it to add alternative non-format variants of all the functions (Log in addition to Logf, etc)? I have pushed a version with this (not entirely finished, naming things etc..). Then I updated the custom lint rules accordingly, reverted back the lint fix and re-ran the lint. This time the autofix fixed everything without creating new lint issues. These changes are in the last 3 commits, rest is same as before.

Nice - that seems like the proper fix - well done :-)

Sorry due to my vacation I've lost momentum on this PR :-(

Do you want to try to get it merged for 1.68 - I'm hoping to release this weekend? If so can you rebase, make sure the tests all pass and I'll merge.

ncw:
Thank you :-)

albertony:
Thank you! :)

> Sorry due to my vacation I've lost momentum on this PR :-(

Well deserved! I was getting a bit impatient due to rather high risk of more conflicts, but seems that was not necessary.

> Do you want to try to get it merged for 1.68 - I'm hoping to release this weekend?

Yes, works for me. I rebased it, and I see you have already merged it. Thank you. 🥳

ncw:
In the hour between you rebasing this and me merging it, I merged something else first which caused the build to break when I merged this. Unfortunate but easily fixed!

I wish GitHub had a rebase a pr button, would make life much easier! Or even rebase this pr if there are no conflicts.

I should probably write a script to do that, wouldn't be too hard...

I'm glad we've merged this now 🙂

albertony:
> I wish GitHub had a rebase a pr button, would make life much easier! 

Is this what you wished for, or did you mean something else:

https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/#always-have-the-option-to-update

ncw:
That is exactly what I meant :-) I've enabled that now - thank you!

Issue #4683, opened 2020-10-14
Add support for S3-compatible object-lock properties


taupehat:
Please see [this forum thread](https://forum.rclone.org/t/requesting-support-for-s3-compatible-object-lock-properties/19675) for initial details including links to the API calls supported by both AWS and Backblaze.

Currently AWS and Backblaze (via their S3 API only) support object immutability. I am proposing a set of flags which would leverage this feature in supported object storage systems. A brief (and probably incomplete) overview of suggested flags follows:

### At bucket creation
` --s3-bucket-object-lock-enabled` - for "rclone mkdir" this enables object-locking for a bucket. It's important to note this _must_ be set at the time a bucket is created for objects inside that bucket to be lockable.

### At object creation/upload
`--s3-object-lock-mode` [GOVERNANCE|COMPLIANCE]
`--s3-object-lock-retain-until-date` [date]
`--s3-object-lock-retention-period` [days]
`--s3-object-lock-legal-hold` [ON|OFF] default off

### At object deletion
` --s3-bypass-governance-retention`

### Determine lock status
`rclone backend list-object-lock`

Additionally, the configurator for S3 would need to know about this feature and which APIs support it so a user would get big scary warnings if they tried to set object-locking in an incompatible provider. This probably means Backblaze should be added to the S3 configurator in addition to the current native B2 configuration.

buchmann:
Will this support also data migration/copy between different Object Storages ? Example from OnPremise Objectstorage to Cloud Objecstorage and keep the Object Lock Retention Period from Source to Target ?

UbuntuRunner:
Did I misunderstand something here or shouldn't the object lock feature be incorporated from version 1.54?

ncw:
This hasn't been implemented yet - if anyone would like to help it would be much appreciated.

monty-dev:
I wanted to provide some info that might help someone decide to take this on. I don't currently have the skill to contribute. 
Adding object lock appears to need 3 values: 

`ObjectLockLegalHoldStatus ` enum values `ON` `OFF`
`ObjectLockMode` enum values `GOVERNANCE` `COMPLIANCE`
`ObjectLockRetainUntilDate` time.Time 

It would be ideal if these values could be provided at config as a default, and also on the upload operation. 
sources:
[go sdk options](https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/upload_input.go#L124)
[...]

PR #7894, opened 2024-06-07
s3: object lock integration

IsSkyfalls:
Add object lock support for individual objects.

Object lock will be applied to newly added object. Governance mode can be bypassed with option.

#### Bucket creation:
`bucket_object_lock_enabled`: create with object-lock flag and turn the object-lock configuration on

#### Object:
`object_lock_mode`: same as issue
`object_lock_retain_until`: differs from issue. merged the originally proposed `until-date` and `period` into one option, reusing fs.Time to accept date/datetime/duration
`object_lock_legal_hold`: same as issue

#### Deletion:
`bypass_governance_retention`: same as issue

#### Quirk:
`manually_set_object_lock`: currently enable for all providers. most s3 providers other than aws requires setting the object lock and legal hold using separate requests. Ideally a maintainer can run `TestIntegrationObjectLocking` in `s3_test.go` on every supported provider to figure this out. 

See issue #4683

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -60,6 +60,7 @@ import (
 	"github.com/rclone/rclone/lib/readers"
 	"github.com/rclone/rclone/lib/rest"
 	"github.com/rclone/rclone/lib/version"
+	"golang.org/x/exp/slices"
 	"golang.org/x/net/http/httpguts"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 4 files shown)


IsSkyfalls:
The `fs.Time` changes felt like too much of a hack to me, but I didn't find any other way of doing this other than duplicating all the existing code. Maybe my go skills are just rusty. If someone have a better idea I'm happy to implement properly.

side note: I just saw the last issue #7893 is also about the same issue. talk about timing!

brainJamStark:
I think we can combine the 2 PR's. the PR that I submitted sets the default retention per bucket and this PR manages the individual object settings.

IsSkyfalls:
Awaiting for maintainer's input on the added flags here. Do we want the DefaultRetention in ObjectLockConfiguration settings to share the same config as object creation?

brainJamStark:
> Awaiting for maintainer's input on the added flags here. Do we want the DefaultRetention in ObjectLockConfiguration settings to share the same config as object creation?

The default bucket retention works with Days and Years whereas the individual object locks it a timestamp.

fiskn:
Hi, small comment and a suggestion.

The parameters state:

				Value: "COMPLIANCE",
				Help:  "Retention period can't be updated or bypassed.",

In COMPLIANCE mode, the retention period can be updated, but only to further in the future.

Could functionality be included to either allow unchanged objects to have their retention updated on each invocation, or a separate run mode to go through and update retention on all existing objects?

This would allow you to maintain a rolling 30 day object locked retention on the destination copy, but any deleted objects would be deleted from the destination once the retention period has elapsed. The use case would be a delayed/versioned immutable copy of a bucket, to protect from malicious actors.

PR #7895, opened 2024-06-07
 backend: pcloud: Implement ChunkWriter using block level API

gfelbing:

- [x] Implement object.SetModTime
  - This is an undocumented API feature (reverse-engineered), using server-side copy with same target, only changing metadata.
  - This was necessary for fileops based upload and a quick win for the backend.
- [x] Implement new option `upload_chunk_size`
  - Enables chunk-wise uploads of files.
  - Successful uploaded chunks are skipped when upload needs retry, thanks to block-wise hashing capability.

---

tl;dr

Uploading large files to pcloud using the uploadfile API is error prone, as any connection reset / package loss will cause the full file to be retried. 
Additionally, I can see that pcloud applies some sort of rate limiting, causing large uploads to be throttled down to 32kbps. 
(Their "support" claims that there is no such rate limit and blames my ISP, but by increasing the `--transfers` with multiple files, get get transfers times 32kbps throughput. Also 32kbps is such a round number, I doubt that its coincidence.)

In my case, this renders it almost impossible to upload any file larger than ~200MB via rclone.

Fortunately, pcloud offers a [fileops API](https://docs.pcloud.com/methods/fileops/), which enables chunked uploads on block level 🎉
This PR adds an advanced option `upload_chunk_size` to the pcloud backend.
When set, files will be uploaded in chunks using the fileops API. Otherwise, current behavior is used (backwards compatibility).

The rate limit was discussed [here](https://forum.rclone.org/t/how-to-handle-pcloud-rate-limit/31279/19), but the described work around only helps when uploading multiple smaller files by increasing `--transfers`.
For single large files, I haven't found any workaround.

Block level sync in general was requests in the forum [here](https://forum.rclone.org/t/pcloud-block-level-sync/43455)

#### Checklist

### DIFF
--- a/backend/pcloud/chunkwriter.go
+++ b/backend/pcloud/chunkwriter.go
@@ -1,0 +1,332 @@
+package pcloud
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha1"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"net/url"
+	"strconv"
+	"sync"
+
+	"github.com/rclone/rclone/backend/pcloud/api"
+	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/lib/rest"
+	"golang.org/x/sync/errgroup"
+)
+
+type chunkWriter struct {
+	client    *rest.Client
+	fs        *Fs
+	chunkSize int64
+	src       fs.ObjectInfo
+	remote    string
+	fd        int64
+	fileID    int64
+
+	mu        sync.RWMutex
+	byteCount int64
+}
+
+func (c *chunkWriter) GetBytesCount() int64 {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.byteCount
+}
+
+func (c *chunkWriter) AddBytesCount(count int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.byteCount = c.byteCount + count
+}
+
+func newChunkWriter(
+	ctx context.Context,
+	remote string,
+	src fs.ObjectInfo,
+	srcFs *Fs,
+) (info fs.ChunkWriterInfo, writer *chunkWriter, err error) {
+	client, err := srcFs.newSingleConnClient(ctx)
+	if err != nil {
+		return info, writer, fmt.Errorf("create client: %w", err)
+	}
+	// init an empty file
+	leaf, directoryID, err := srcFs.dirCache.FindPath(ctx, remote, true)
+	if err != nil {
+		return info, writer, fmt.Errorf("resolve src: %w", err)
+	}
+	openResult, err := fileOpenNew(ctx, client, srcFs, directoryID, leaf)
+	if err != nil {
+		return info, writer, fmt.Errorf("open file: %w", err)
+	}
+
+	info = fs.ChunkWriterInfo{
+		ChunkSize:         int64(srcFs.opt.ChunkSize),
+		Concurrency:       srcFs.opt.UploadConcurrency,
+		LeavePartsOnError: true,
+	}
+	writer = &chunkWriter{
+		client:    client,
+		fs:        srcFs,
+		chunkSize: info.ChunkSize,
+		src:       src,
+		remote:    remote,
+		fd:        openResult.FileDescriptor,
+		fileID:    openResult.Fileid,
+	}
+
+	return info, writer, nil
+}
+
+// Abort implements fs.ChunkWriter.
+func (c *chunkWriter) Abort(ctx context.Context) error {
+	o, err := c.fs.NewObject(ctx, c.remote)
+	if err != nil {
+		return fmt.Errorf("open aborted: %w", err)
+	}
+	if err := o.Remove(ctx); err != nil {
+		return fmt.Errorf("delete aborted: %w", err)
+	}
+	return nil
+}
+
+// Close implements fs.ChunkWriter.
+func (c *chunkWriter) Close(ctx context.Context) error {
+	// In case the file existed previosly and was larger than the newly written file, we need to truncate it to the right size.
+	byteCount := c.GetBytesCount()
+	fs.Debugf(c.src, "truncate target to %d bytes", byteCount)
+	if _, err := c.fileTruncate(ctx, byteCount); err != nil {
+		return fmt.Errorf("truncate file: %w", err)
+	}
+
+	// close fd
+	if _, err := c.fileClose(ctx); err != nil {
+		return fmt.Errorf("close fd: %w", err)
+	}
+
+	// There seems to be no option in the fileops API to set the modtime.
+	// Therefore, we apply the modtime via the regular file API.
+	o, err := c.fs.NewObject(ctx, c.remote)
+	if err != nil {
+		return fmt.Errorf("open newly copied object: %w", err)
+	}
+	if err := o.SetModTime(ctx, c.src.ModTime(ctx)); err != nil {
+		return fmt.Errorf("set modtime: %w", err)
+	}
+
+	return nil
+}
+
+// WriteChunk implements fs.ChunkWriter.
+func (c *chunkWriter) WriteChunk(ctx context.Context, chunkNumber int, reader io.ReadSeeker) (int64, error) {
+	return c.writeChunk(ctx, chunkNumber, reader)
+}
+
+// writeChunk writes c.chunkSize bytes from reader (offset 0) to the target file with offset c.chunkSize * chunkNumber.
+// This method is used by the exported `chunkWriter.WriteChunk` as well as `Object.Update`.
+func (c *chunkWriter) writeChunk(ctx context.Context, chunkNumber int, reader io.Reader) (int64, error) {
+	chunk := int64(chunkNumber)
+
+	// read chunk into buffer
+	buffer := make([]byte, c.chunkSize)
+	offset := chunk * c.chunkSize
+	eg, egCtx := errgroup.WithContext(ctx)
+
+	// read input buffer and calculate hash
+	inSHA1 := ""
+	eg.Go(func() error {
+		byteCount, err := io.ReadFull(reader, buffer)
+		buffer = buffer[:byteCount]
+		if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) && !errors.Is(err, io.EOF) {
+			return err
+		}
+		inSHA1Bytes := sha1.Sum(buffer)
+		inSHA1 = hex.EncodeToString(inSHA1Bytes[:])
+		return nil
+	})
+
+	// get target hash
+	outSHA1 := ""
+	eg.Go(func() error {
+		targetChecksum, err := c.fileChecksum(egCtx, offset, c.chunkSize)
+		if err != nil {
+			return err
+		}
+		outSHA1 = targetChecksum.SHA1
+		return nil
+	})
+
+	if err := eg.Wait(); err != nil {
+		if errors.Is(err, io.EOF) {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("check hashes of chunk %d: %w", chunk, err)
+	}
+	if outSHA1 == "" || inSHA1 == "" {
+		return 0, fmt.Errorf("block %d: expect both hashes to be filled: src: %q, target: %q", chunk, inSHA1, outSHA1)
+	}
+
+	byteCount := int64(len(buffer))
+	c.AddBytesCount(byteCount)
+	// check hash of chunk, skip if fits
+	if inSHA1 == outSHA1 {
+		fs.Debugf(c.src, "chunk %d matches", chunk)
+		return byteCount, nil
+	}
+
+	// upload buffered chunk with offset if necessary
+	if _, err := c.filePWrite(ctx, offset, buffer); err != nil {
+		return 0, err
+	}
+	fs.Debugf(c.src, "chunk %d written", chunk)
+	return byteCount, nil
+}
+
+// Call pcloud file_open using folderid and name with O_CREAT and O_WRITE flags, see [API Doc.]
+// [API Doc]: https://docs.pcloud.com/methods/fileops/file_open.html
+func fileOpenNew(ctx context.Context, c *rest.Client, srcFs *Fs, directoryID, filename string) (*api.FileOpenResponse, error) {
+	opts := rest.Opts{
+		Method:           "PUT",
+		Path:             "/file_open",
+		Parameters:       url.Values{},
+		TransferEncoding: []string{"identity"}, // pcloud doesn't like chunked encoding
+		ExtraHeaders: map[string]string{
+			"Connection": "keep-alive",
+		},
+	}
+	filename = srcFs.opt.Enc.FromStandardName(filename)
+	opts.Parameters.Set("name", filename)
+	opts.Parameters.Set("folderid", dirIDtoNumber(directoryID))
+	opts.Parameters.Set("flags", "0x0042") // O_CREAT, O_WRITE
+
+	result := &api.FileOpenResponse{}
+	err := srcFs.pacer.CallNoRetry(func() (bool, error) {
+		resp, err := c.CallJSON(ctx, &opts, nil, result)
+		err = result.Error.Update(err)
+		return shouldRetry(ctx, resp, err)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("open new file descriptor: %w", err)
+	}
+	return result, nil
+}
+
+// Call pcloud file_checksum, see [API Doc.]
+// [API Doc]: https://docs.pcloud.com/methods/fileops/file_checksum.html
+func (c *chunkWriter) fileChecksum(
+	ctx context.Context,
+	offset, count int64,
+) (*api.FileChecksumResponse, error) {
+	opts := rest.Opts{
+		Method:           "PUT",
+		Path:             "/file_checksum",
+		Parameters:       url.Values{},
+		TransferEncoding: []string{"identity"}, // pcloud doesn't like chunked encoding
+		ExtraHeaders: map[string]string{
+			"Connection": "keep-alive",
+		},
+	}
+	opts.Parameters.Set("fd", strconv.FormatInt(c.fd, 10))
+	opts.Parameters.Set("offset", strconv.FormatInt(offset, 10))
+	opts.Parameters.Set("count", strconv.FormatInt(count, 10))
+
+	result := &api.FileChecksumResponse{}
+	err := c.fs.pacer.CallNoRetry(func() (bool, error) {
+		resp, err := c.client.CallJSON(ctx, &opts, nil, result)
+		err = result.Error.Update(err)
+		return shouldRetry(ctx, resp, err)
+	})
+	if err != nil {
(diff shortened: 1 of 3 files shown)


ncw (commented):
This looks good - well done :-) Multipart uploading code is tricky to write!

I put some comments inline.

One thought - since pcloud seems to be able to upload chunks out of sequence, how about implementing this interface instead

https://github.com/rclone/rclone/blob/b778ec014208bcd11f2739344fea762712568ad5/fs/features.go#L176

This will unlock multi-thread uploads from the rclone core which means rclone will be able to parallel transfer blocks from s3 to pcloud for example, without reading them sequentially.

You can then use some already existing code to call your OpenChunkWriter from your Update - here is how the s3 backend does it:

https://github.com/rclone/rclone/blob/b778ec014208bcd11f2739344fea762712568ad5/backend/s3/s3.go#L6322

https://github.com/rclone/rclone/blob/b778ec014208bcd11f2739344fea762712568ad5/backend/s3/s3.go#L5983

Can you also check the integration tests all pass by using

    # at the root of the source code
    go install ./...
    test_all -remotes mypcloudremote:

Thanks

ncw:
> * [x]  Implement object.SetModTime
>   
>   * This is an undocumented API feature (reverse-engineered), using server-side copy with same target, only changing metadata.
>   * This was necessary for fileops based upload and a quick win for the backend.

Can you put this is a separate commit please. I don't mind having two commits in the PR but like to keep the features one per commit.

> * [x]  Implement new option `upload_chunk_size`
>   
>   * Enables chunk-wise uploads of files.
>   * Successful uploaded chunks are skipped when upload needs retry, thanks to block-wise hashing capability.

This should just be called `chunk_size` to make it consistent with the other backends. You might want to consider having an `upload_cutoff` also. I'd cut and paste these Option definitions from the s3 backend and alter as necessary!

      --s3-chunk-size SizeSuffix   Chunk size to use for uploading (default 5Mi)
      --s3-upload-cutoff SizeSuffix          Cutoff for switching to chunked upload (default 200Mi)

backend/pcloud/pcloud.go
```diff
@@ -146,30 +151,40 @@ we have to rely on user password authentication for it.`,
 			Help:       "Your pcloud password.",
 			IsPassword: true,
 			Advanced:   true,
+		}, {
+			Name:     "upload_chunk_size",
```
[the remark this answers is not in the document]
gfelbing: done

backend/pcloud/pcloud.go
```diff
@@ -146,30 +151,40 @@ we have to rely on user password authentication for it.`,
...
-	RootFolderID string               `config:"root_folder_id"`
-	Hostname     string               `config:"hostname"`
-	Username     string               `config:"username"`
-	Password     string               `config:"password"`
+	Enc             encoder.MultiEncoder `config:"encoding"`
+	RootFolderID    string               `config:"root_folder_id"`
+	Hostname        string               `config:"hostname"`
+	Username        string               `config:"username"`
+	Password        string               `config:"password"`
+	UploadChunkSize string               `config:"upload_chunk_size"`
```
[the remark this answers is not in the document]
gfelbing: done

backend/pcloud/pcloud.go
```diff
@@ -146,30 +151,40 @@ we have to rely on user password authentication for it.`,
...
-	dirCache     *dircache.DirCache // Map of directory path to directory id
-	pacer        *fs.Pacer          // pacer for API calls
-	tokenRenewer *oauthutil.Renew   // renew the token on expiry
+	name                string                                          // name of this remote
+	root                string                                          // the path we are working on
+	opt                 Options                                         // parsed options
+	features            *fs.Features                                    // optional features
+	srv                 *rest.Client                                    // the connection to the server
+	newSingleConnClient func(ctx context.Context) (*rest.Client, error) // creates a new client, accepting keep-alive headers, limited to single connection. Used (and necessary) for pcloud fileops API, as it binds the session to the underlying TCP connection (thanks, pcloud! grml.). File descriptors are only valid within the same connection and auto-closed when the connection is closed, hence we need a separate client (with single connection) for each fd to avoid all sorts of errors and race conditions.
```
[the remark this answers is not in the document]
gfelbing: Not really. I moved it into it's own method and cleaned it a little (re-uses now the tokensource)

backend/pcloud/pcloud.go
```diff
@@ -1180,6 +1418,111 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
...
+	chunkSize := int64(0)
+	if o.fs.opt.UploadChunkSize != "" {
+		parsed, err := strconv.ParseInt(o.fs.opt.UploadChunkSize, 10, 64)
+		if err != nil {
+			return fmt.Errorf("parse upload chunk size: %w", err)
+		}
+		chunkSize = parsed
+	}
+	if chunkSize != 0 && chunkSize < size {
+		c, err := o.fs.newSingleConnClient(ctx)
```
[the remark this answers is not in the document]
gfelbing: I guess this one is obsolete, as this is moved into the chunkreader and I added a upload_cutoff.

gfelbing:
Hi @ncw ! 
Thanks for your great review, the hint towards the ChunkWriter was great!
I implemented all your suggestions.

There is only one caveat:
In order to make the ChunkWriter work for pcloud, I had to extend the ChunkWriter.Close interface, so it returns the new object.
This is necessary to fix the hashes race condition: pcloud's `/checksumfile` (which should generally be used), uses cached checksums which do not match the actual one directly after the fd is closed.
In that case, I need to get the checksums via `/file_checksum`, which is uncached and should therefore only be used for parts or edited files.
I created a separate commit for the interface extension. What do you think?

ncw:
> Thanks for your great review, the hint towards the ChunkWriter was great!
> I implemented all your suggestions.

Great :-)

> In order to make the ChunkWriter work for pcloud, I had to extend the ChunkWriter.Close interface, so it returns the new object.
> This is necessary to fix the hashes race condition: pcloud's `/checksumfile` (which should generally be used), uses cached checksums which do not match the actual one directly after the fd is closed.
> In that case, I need to get the checksums via `/file_checksum`, which is uncached and should therefore only be used for parts or edited files.

You should be able to use the chunkwriter interface as-is. S3 needs to do the same thing and you can see how it does it here by type asserting the chunkwriter handle back into something that it understands.

https://github.com/rclone/rclone/blob/c8d6b02dd62ba11a5d6aa16585d84cff8b15b6f2/backend/s3/s3.go#L5984-L5992

gfelbing:
Hey @ncw 

about the interface issue:

> You should be able to use the chunkwriter interface as-is. S3 needs to do the same thing and you can see how it does it here by type asserting the chunkwriter handle back into something that it understands.
> 
> https://github.com/rclone/rclone/blob/c8d6b02dd62ba11a5d6aa16585d84cff8b15b6f2/backend/s3/s3.go#L5984-L5992

Maybe I didn't get it completely, but this would only work for the `Fs.Update` path, right?
I case of a multithreaded operation (e.g. when uploading from local fs to pcloud), the ChunkWriter is directly called by the operation.
Afterwards, it creates a new object, w/o a chance for the underlying Fs to get the connection to the chunked upload:
https://github.com/rclone/rclone/pull/7895/files#diff-57914b2dca5f9b6573400f20eb94657c8fcb57b692e0636af96d7a74ec228016L242

While this seems to work for S3 (as long as the API implementation doesn't have race conditions), pcloud has it's problems with that. How would you address that w/o changing the interface?

gfelbing:
So run some more experiments with the pCloud API:

- file_pwrite also works beyond file size boundaries, pcloud just increases the file size then
- it seems that the API invalidates it's cached hash when the **size** of a file changes

Given that knowledge, I was able to implement the ChunkWriter with the old interface, by moving the file_truncate call to the Close (it's still necessary, as we need to reduce the file size when it existed before and was larger).

gfelbing:
Hi @ncw , so after fixing the race condition, I went on with the integration tests.
Apparently, there was a bug in the test: The `TestMultithreadCopyAbort` only worked for backends that support only a single hash, as it calls `SupportOnly` with the first hash from the tested backend. 
I fixed the test in a separate commit.

That's that, I am looking forward to your final review!

gfelbing:
Here are the test results for the pcloud backend: 
[2024-06-09-135538.zip](https://github.com/user-attachments/files/15754460/2024-06-09-135538.zip)

ncw:
> Given that knowledge, I was able to implement the ChunkWriter with the old interface, by moving the file_truncate call to the Close (it's still necessary, as we need to reduce the file size when it existed before and was larger).

Great



> So after fixing the race condition, I went on with the integration tests. Apparently, there was a bug in the test: The `TestMultithreadCopyAbort` only worked for backends that support only a single hash, as it calls `SupportOnly` with the first hash from the tested backend. I fixed the test in a separate commit.
> 
> That's that, I am looking forward to your final review!

Why did that cause a problem? If one hash matches then that is considered to be fine. Lots of rclone processes call `GetOne` to find a hash which works.

With this code

https://github.com/rclone/rclone/blob/d2af11413920c7d4c1a33f7ff089aa9122ab1b57/fs/operations/multithread_test.go#L124-L125

The problem backend is the local backend so you'll find that if you run the fs/operations tests on the local backend they will take a very long time.

So I'd question why the hash patch is needed. Rclone should be able to pick one hash and have it work. Can you explain?

> Here are the test results for the pcloud backend: [2024-06-09-135538.zip](https://github.com/user-attachments/files/15754460/2024-06-09-135538.zip)

Excellent work - that is what I like to see :-)

gfelbing:
Hey @ncw ,

> So I'd question why the hash patch is needed. Rclone should be able to pick one hash and have it work. Can you explain?

Currently, the tests work only by accident, as all backends that implement `ChunkWriter` only support a single hash type.
With `pcloud` supporting two hashes and `ChunkWriter`, the tests fail because

1. [TestMultithreadCopyAbort](https://github.com/rclone/rclone/blob/d2af11413920c7d4c1a33f7ff089aa9122ab1b57/fs/operations/multithread_test.go#L124-L125) calls SupportOnly with the first hash of the tested backend during setup phase. This removes the second (and all other hashes) of the tested backend from the supported hash types.
2. [The Test Run](https://github.com/rclone/rclone/blob/d2af11413920c7d4c1a33f7ff089aa9122ab1b57/fstest/run.go#L261) calls `hash.NewMultiHasherTypes`, with all the Hashes of the tested backend. This one fails, as it contains the second, then "unsupported" hash.

In order to fix that, we have 2 options:

1. The one I implemented: Reducing the hashes to those supported by the tested backend. This reduces the CPU/Memory effort to only the hashes necessary: For all backends that only support a single hash, nothing changes. For pcloud, both hashes are tested. To me, this was the cleaner solution.
2. Workaround in run.go: In order to make the test work with only a single hash, the [test run](https://github.com/rclone/rclone/blob/d2af11413920c7d4c1a33f7ff089aa9122ab1b57/fstest/run.go#L261) also needs to be adjusted to call the `hash.NewMultiHasherTypes` with only the first hash of the backend. While this would fix the symptom, idk if there are more places that are broken the same way. Also, it is prone to introduce more issues like that.

So, choose your poison :)

ncw:
> > So I'd question why the hash patch is needed. Rclone should be able to pick one hash and have it work. Can you explain?
> 
> Currently, the tests work only by accident, as all backends that implement `ChunkWriter` only support a single hash type.

The problem that bit of code is trying to fix is that the local backend supports `OpenWriterAt` and therefore multipart copy. However it also supports **all** the hashes and without that fix (or with your patch) it takes >100s to run the tests! The fix adjusts the global hashes available which is rather a bodge.

I've merged what I think is the correct patch here to master which only enables the fix/bodge when the destination is the local backend.

d068e0b1a9e587ddc669e0a795d1dcee99defd74

This retains the 20s run time for local backends (which is the important bit) and I've tested it works with your pcloud changes.

So if you could drop the "operations: multithread_test: fix integration test for backends..." commit and rebase your PR on master and  then I think we are done :-)

Thanks for digging into this failure because its been causing a failure in the integration tests with the `combine` backend too!

gfelbing:
Hey @ncw , 
I rebased my PR to master and run the tests again: success.
I also think this is ready now :)

gfelbing:
(rebased against latest master)

backend/pcloud/chunkwriter.go
```diff
@@ -0,0 +1,337 @@
...
+		return fmt.Errorf("open newly copied object: %w", err)
+	}
+	if err := o.SetModTime(ctx, c.src.ModTime(ctx)); err != nil {
+		return fmt.Errorf("set modtime: %w", err)
+	}
+
+	return nil
+}
+
+// WriteChunk implements fs.ChunkWriter.
```
ncw, line 123: This function looks redundant now? Perhaps just rename `writeChunk`

backend/pcloud/chunkwriter.go
```diff
@@ -0,0 +1,337 @@
...
+	eg.Go(func() error {
+		targetChecksum, err := c.fileChecksum(egCtx, offset, c.chunkSize)
+		if err != nil {
+			return err
+		}
+		outSHA1 = targetChecksum.SHA1
+		return nil
+	})
+
+	if errors.Is(err, io.EOF) {
```
ncw, line 160: As far as I can see, err hasnt been set here?

backend/pcloud/chunkwriter.go
```diff
@@ -0,0 +1,337 @@
...
+		return fmt.Errorf("delete aborted: %w", err)
+	}
+	return nil
+}
+
+// Close implements fs.ChunkWriter.
+func (c *chunkWriter) Close(ctx context.Context) error {
+	// In case the file existed previosly and was larger than the newly written file, we need to truncate it to the right size.
+	byteCount := c.GetBytesCount()
+	fs.Debugf(c.src, "truncate target to %d bytes", byteCount)
```
ncw, line 100: Does this mean that we are now overwriting existing files?

This means that if a transfer fails we've clobbered an existing file - is that correct?

If so then we need to set this feature flag

        PartialUploads           bool // uploaded file can appear incomplete on the fs while it's being uploaded

so rclone knows to upload to a temporary file name and rename it over at the end.

backend/pcloud/chunkwriter.go
```diff
@@ -0,0 +1,337 @@
...
+	if err := eg.Wait(); err != nil {
+		return 0, fmt.Errorf("check hashes of chunk %d: %w", chunk, err)
+	}
+	if outSHA1 == "" || inSHA1 == "" {
+		return 0, fmt.Errorf("block %d: expect both hashes to be filled: src: %q, target: %q", chunk, inSHA1, outSHA1)
+	}
+
+	byteCount := int64(len(buffer))
+	c.AddBytesCount(byteCount)
+	// check hash of chunk, skip if fits
```
ncw, line 173: This check is checking to see if existing chunks in the file match to see if we can skip the upload.

If we turn on the `PartialUploads` flag then this test will no longer work.

Also if you uploading a big file that has lots of changes (say a compressed `.tar.gz`) then this will effectively download the file and upload it.

So I'm not terribly keen on this test anyway.

backend/pcloud/chunkwriter.go
```diff
@@ -0,0 +1,337 @@
...
+	if _, err := c.filePWrite(ctx, offset, buffer); err != nil {
+		return 0, err
+	}
+	fs.Debugf(c.src, "chunk %d written", chunk)
+	return byteCount, nil
+}
+
+// Call pcloud file_open using folderid and name with O_CREAT and O_WRITE flags, see [API Doc.]
+// [API Doc]: https://docs.pcloud.com/methods/fileops/file_open.html
+func fileOpenNew(ctx context.Context, c *rest.Client, srcFs *Fs, directoryID, filename string) (*api.FileOpenResponse, error) {
```
ncw, line 189: These file operations look very general.

In fact general enough to implement the `OpenWriterAt` interface

```
	// OpenWriterAt opens with a handle for random access writes
	//
	// Pass in the remote desired and the size if known.
	//
	// It truncates any existing object
	OpenWriterAt func(ctx context.Context, remote string, size int64) (WriterAtCloser, error)
```

That is a more general interface than the `OpenChunkWriter` .

Is it possible to implement this? Rclone will use this in preference to the `OpenChunkWriter` interface for multithread uploads.

What do you think?

ncw (commented):
This is looking very nice now.

I gave the code a thorough going through and I came up with some things to discuss - please see inline.

Thank you :-)

PS These multipart uploaders are really tricky to get right!

backend/pcloud/chunkwriter.go
gfelbing: Not exactly, as you can see from the reader interface. 
The methods are both in use:
- `WriteChunk` is the exported method fullfilling the `ChunkWriter` interface using `io.ReadSeeker`
- `writeChunk` is called by the `fs.Update` while only having a `io.Reader` at hand

backend/pcloud/chunkwriter.go
gfelbing: right, this is why I don't like pre-defined return variables. I'll fix it.

backend/pcloud/chunkwriter.go
gfelbing: I added a comment.

backend/pcloud/chunkwriter.go
gfelbing: This check is the essential part of the chunked upload in my use-case:

I use rclone with pcloud to backup many large media files from local.
As uploading of large files often fails with pcloud (e.g. by a server side ConnReset), so I need to have a way to resume failed uploads of large files.

The hash for pcloud isn't a problem here, as they luckily offer an API for partial hashes.
While I understand that uploading a large file from another non-local remote may cause downloading the whole file just for checking the issues: The download would happen anyway, as the upload is runs through the local RAM as well, while this check would at least save the upload (which is probably the critical thing in most cases).

We could however skip this check by checking the file size, e.g. by storing the originial file size in `Open` and skipping this check when offset+chunksize runs out of bounds.

Just skipping this check wouldn't be an option for me. Alternatively, is there another concept in rclone for resuming failed uploads?

What do you think?

gfelbing (commented):
Hey @ncw ,

thanks for the review!
I fixed two of your comments already and answered to the others.
I guess the main thing is the question around the hashing of single chunks.

backend/pcloud/chunkwriter.go
gfelbing: @ncw I also tried to get some more details on this:
IMHO PartialUploads would be bad, as we loose the continuation of chunked uploads (see other discussion).
Yet I could make any sorts of handling for existing files, e.g
- delete on failure (current solution)
- [revert on failure](https://docs.pcloud.com/methods/revisions/revertrevision.html)
- leave broken file

In order to have a good solution here, can you tell me what the desired behavior would be on the following cases:

- successful upload (Close is called) of a file that already exists
- failed upload (Abort is called) of a file that already exists
- successful upload (Close is called) of a file that doesn't exist
- failed upload (Abort is called) of a file that doesn't exist

(I deliberately included the obvious cases for sake for completeness)

gfelbing:
Hi @ncw , any chance that you could take a look again?

ncw:
I think this PR is obsoleted by #7949 - would you agree @gfelbing  if so please close - thank you.

gfelbing:
Agree, superseeded by #7949

PR #7897, opened 2024-06-09
sftp: use uint32 for mtime

liori:
The SFTP protocol (and the golang sftp package) internally uses uint32 unix time for expressing mtime. Hence it is a waste of memory to store it as 24-byte time.Time data structure in long-lived data structures. So despite that the golang sftp package uses time.Time as external interface, we can re-encode the value back to the original format and save memory.

Save memory when mounting sftp backends. In my specific circumstances this simple change saves hundreds of MB of RAM.

No.

#### Checklist

I do not believe tests are necessary, as the only values that make sense in the context of sftp would already be within the range of uint32. As such, there should be no externally observable changes.

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -561,7 +561,7 @@ type Object struct {
 	fs      *Fs
 	remote  string
 	size    int64       // size of the object
-	modTime time.Time   // modification time of the object
+	modTime uint32      // modification time of the object as unix time
 	mode    os.FileMode // mode bits from the file
 	md5sum  *string     // Cached MD5 checksum
 	sha1sum *string     // Cached SHA1 checksum
@@ -1957,7 +1957,7 @@ func (o *Object) Size() int64 {
 
 // ModTime returns the modification time of the remote sftp file
 func (o *Object) ModTime(ctx context.Context) time.Time {
-	return o.modTime
+	return time.Unix(int64(o.modTime), 0)
 }
 
 // path returns the native SFTP path of the object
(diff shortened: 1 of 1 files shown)


ncw:
So save you hundreds of MB you must have at least 10M objects in memory which is likely using 10GB of memory in total.

What is your use case for rclone?

I would have thought there were lower hanging fruit on the memory usage tree?

You patch looks like it will work fine. I'm not terribly keen on it but if you can demostrate a good need I'll go with it :-)

It might be cleaner to use `.Sys()` on the returned `FileInfo`, assert it to a [FileStat](https://pkg.go.dev/github.com/pkg/sftp#FileStat) and read the info from there? That at least saves the round trip into a time.Time and back and shows where the `uint32` value comes from.

liori:
Indeed, millions or tens of millions of smallish objects. There are three main sources of memory consumption in my case: mtime, hashes, and file paths. This patch helps with mtime. I've been disabling hashes for a long time anyway, but the 16 bytes spent on nil pointers still hurt almost as much as mtime—my local rclone compilation has them commented out, a real solution would probably provide two versions of the _Object_ structure. The third one… I have something in mind, haven't tested yet.

liori:
Ah, I think I got what you mean. Well, it would be best if the sftp package exposed a strongly typed accessor for the FileStat structure…

ncw:
> Ah, I think I got what you mean. Well, it would be best if the sftp package exposed a strongly typed accessor for the FileStat structure…

No just do the type assertion, you can check it it works or not so it is effectively strong typed at runtime and only cost a few nS.

Then copy the Mtime out of it so as to not keep the `FileStat` alive.

ncw:
Are you going to finish this off @liori ?

liori:
Yes, I do. It's just that I only have time for this kind of hobby every so often, sorry!

liori:
Sorry it took so long, wanted to check another approach first.

puellanivis (approved):
Can’t speak for its appropriateness in the rest of the code, but the relevant `sftp` internals being used seems appropriate.

ncw (approved):
This looks great now, thank you

PR #7908, opened 2024-06-13
`s3`: fix 405 error on `HEAD` for delete marker with `versionId`

nielash:

When getting an object by specifying a `versionId` in the request, if the specified version is a delete marker, it returns `405 (Method Not Allowed)`, instead of `404 (Not Found)` which would be returned without a `versionId`.
See https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeleteMarker.html

Before this change, we were only looking for `404` (and not `405`) to determine whether the object exists. This meant that in some circumstances (ex. when Versioning is enabled for the bucket and we have a non-null `X-Amz-Version-Id`), we deemed the object to exist when we should not have.

After this change, `405 (Method Not Allowed)` is treated the same as `404 (Not Found)` for the purposes of `headObject`.

- https://forum.rclone.org/t/bisync-rename-failed-method-not-allowed/45723/13

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -5422,7 +5422,7 @@ func (f *Fs) headObject(ctx context.Context, req *s3.HeadObjectInput) (resp *s3.
 	})
 	if err != nil {
 		if awsErr, ok := err.(awserr.RequestFailure); ok {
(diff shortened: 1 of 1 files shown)


ncw (approved):
Great fix - thank you :-)

I'll merge now.

PR #7929, opened 2024-07-01
Implement missing Docker volume plugin mount and VFS options

jonasgeiler:

I have noticed that some options were missing when using rclone as a Docker volume plugin, like `vfs-fast-fingerprint` or `vfs-cache-min-free-space`, so I went through the `mountlib.Options` and `vfscommon.Options` structs and implemented all the missing ones.
I hope that all these options are supposed to be supported, and they won't break something with Docker or similar, but at the end this would be a user problem I guess, since you don't *have* to set them :)

New mount options:
- `daemon-wait`
- `devname`
- `direct-io`
- `mount-case-insensitive`

New VFS options:
- `vfs-refresh`
- `vfs-cache-min-free-space`
- `vfs-block-norm-dupes`
- `vfs-fast-fingerprint`
- `vfs-disk-space-total-size`
- `no-seek`

All of these are even documented here: https://rclone.org/commands/rclone_serve_docker/#options
But they did not actually work until now.

Unfortunately I did not get any feedback on my forum post: https://forum.rclone.org/t/docker-volume-plugin-is-apparently-missing-some-vfs-options/45794
But since it's a pretty minor change I decided to just open a PR.

#### Checklist

### DIFF
--- a/cmd/serve/docker/options.go
+++ b/cmd/serve/docker/options.go
@@ -216,6 +216,14 @@ func getMountOption(mntOpt *mountlib.Options, opt rc.Params, key string) (ok boo
 		mntOpt.NoAppleXattr, err = opt.GetBool(key)
 	case "network-mode":
 		mntOpt.NetworkMode, err = opt.GetBool(key)
+	case "daemon-wait":
+		mntOpt.DaemonWait, err = opt.GetDuration(key)
+	case "devname":
+		mntOpt.DeviceName, err = opt.GetString(key)
+	case "direct-io":
+		mntOpt.DirectIO, err = opt.GetBool(key)
+	case "mount-case-insensitive":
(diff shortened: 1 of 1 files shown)


jonasgeiler:
Should I add tests for this? I could not find any existing tests for the two modified functions, unfortunately. Please let me know and I will see what I can do.

jonasgeiler:
@ncw, sorry for the ping but I saw that you recently made some changes in `cmd/serve/docker/options.go` which caused a merge conflict that I have fixed now, along with adding the new `vfs-read-chunk-streams` option.
I would be grateful if you could now take a look at this PR, since it's a very helpful but minor change!

jonasgeiler:
Fixed in 33d2747 - this PR not relevant anymore.

PR #7948, opened 2024-07-13
swift: Add list_chunks and retry_403 options

bodqhrohro:
Discussion: 

Blomp tends to return 403 errors on directory listing requests which are not really informative because such requests eventually succeed if repeated.

This commit adds two options to relieve this behaviour and make syncing large directories more reliable.

https://forum.rclone.org/t/too-many-errors-with-blomp-especially-in-directories-with-thousands-of-files/46769

#### Checklist

### DIFF
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -36,7 +36,6 @@ import (
 // Constants
 const (
 	directoryMarkerContentType = "application/directory" // content type of directory marker objects
-	listChunks                 = 1000                    // chunk size to read directory listings
 	defaultChunkSize           = 5 * fs.Gibi
 	minSleep                   = 10 * time.Millisecond // In case of error, start at 10ms sleep.
 	segmentsContainerSuffix    = "_segments"
@@ -308,6 +307,20 @@ of the Swift API that do not implement pagination as expected.  See
 also "fetch_until_empty_page".`,
 			Default:  0,
 			Advanced: true,
+		}, {
+			Name: "list_chunks",
+			Help: `Maximum amount of directory entries per listing request.
+
+Reduce this if you have errors with listing large directories.`,
+			Default:  1000,
+			Advanced: true,
+		}, {
+			Name: "retry_403",
+			Help: `Retry on 403 errors.
+
+Blomp is known to return 403 errors on directory listing requests.`,
+			Default:  false,
+			Advanced: true,
 		}}, SharedOptions...),
 	})
 }
@@ -340,6 +353,8 @@ type Options struct {
 	Enc                         encoder.MultiEncoder `config:"encoding"`
 	FetchUntilEmptyPage         bool                 `config:"fetch_until_empty_page"`
 	PartialPageFetchThreshold   int                  `config:"partial_page_fetch_threshold"`
+	ListChunks                  int                  `config:"list_chunks"`
+	Retry403                    bool                 `config:"retry_403"`
 }
 
 // Fs represents a remote swift server
@@ -411,7 +426,7 @@ var retryErrorCodes = []int{
 
 // shouldRetry returns a boolean as to whether this err deserves to be
 // retried.  It returns the err as a convenience
-func shouldRetry(ctx context.Context, err error) (bool, error) {
+func shouldRetry(ctx context.Context, err error, retry403 bool) (bool, error) {
 	if fserrors.ContextError(ctx, &err) {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


ncw (commented):
I think I would unconditionally retry on 403 errors only just in the listing code. Something like this (untested)

```diff
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -672,6 +672,10 @@ func (f *Fs) listContainerRoot(ctx context.Context, container, directory, prefix
 		var err error
 		err = f.pacer.Call(func() (bool, error) {
 			objects, err = f.c.Objects(ctx, container, opts)
+			// Retry 403 errors on listings as Blomp is fond of returning them
+			if swiftError, ok := err.(*swift.Error); ok && swiftError.StatusCode == 403 {
+				return true, err
+			}
 			return shouldRetry(ctx, err)
 		})
 		if err == nil {
```

Retrying 403 errors is a really bad idea in general so we should try to limit the fallout.

What value did you come up with for list_chunks?

Maybe we should be putting these tweaks in a Blomp quirks section.

bodqhrohro:
> I think I would unconditionally retry on 403 errors only just in the listing code.

I got them on other requests as well.

```
@bq:10:44:40:/tmp/dl$ rg -io 'Failed to.*forbidden' /tmp/rclone*|rg -v 'with.*errors'|sed 's/^[^:]\+://'|sort|uniq -c
    647 Failed to calculate dst hash: Operation forbidden
      4 Failed to copy: failed to get list segments of object: Operation forbidden
    878 Failed to copy: Operation forbidden
     31 Failed to remove partially transferred object: Operation forbidden
    157 Failed to set modification time: Operation forbidden
     28 Failed to sync: Operation forbidden
```

> Retrying 403 errors is a really bad idea in general

Yup, repeating genuine authentication errors may indeed lead to an IP ban or something theoretically, that's why I made this as an option which should be enabled only when credentials are tested and are definitely accepted already.

> What value did you come up with for list_chunks?

100 seems to work well mostly, I didn't really test other values.

bodqhrohro:
I also sometimes receive 404 and 405 errors, should they be eligible for retries too, huh? 404 might be genuine sometimes (but after write operations it's definitely questionable).
```
@bq:11:17:14:/tmp/dl$ rg -io 'Failed to.*object not found' /tmp/rclone*|rg -v 'with.*errors'|sed 's/^[^:]\+://'|sort|uniq -c
     18 Failed to calculate dst hash: object not found
      1 Failed to copy: error decoding metadata: Object Not Found
      1 Failed to copy: failed to open source object: Object Not Found
     37 Failed to copy: object not found
      1 failed to open source object: Object Not Found
      1 Failed to set modification time: object not found
     44 Failed to set modification time: Object Not Found
      1 Failed to sync: failed to open source object: Object Not Found
      4 Failed to sync: object not found
@bq:11:17:22:/tmp/dl$ rg -io 'Failed to.*Method.*' /tmp/rclone*|rg -v 'with.*errors'|sed 's/^[^:]\+://'|sort|uniq -c
      2 Failed to calculate dst hash: HTTP Error: 405: 405 Method Not Allowed
     29 Failed to copy: HTTP Error: 405: 405 Method Not Allowed
```

ncw:
Urg.

400 errors shouldn't be retried.

Retrying 404 errors is weird but perhaps OK in specific circumstances.

It giving 405 errors is weird too.

I think we need a different approach. Are these errors being dished out by a proxy rather than swift itself? Can we tell that from some headers? If we could reliably detect errors from the proxy then we could retry any of those.

Try running with `-vv --dump headers` and have a look.

bodqhrohro:
I had provided a sample for the 403 error in the forum thread and it doesn't seem there are any clues really, just `X-Openstack-Request-Id` and `X-Trans-Id` headers like everywhere.

bodqhrohro:
I'd rather wonder why 404 errors are sometimes in lowercase and sometimes in camelcase, like they come from different places.

ncw:
Something strange is going on here. I have a contact at Blomp and I've emailed them to see if they can shed any light on the problem.

bodqhrohro:
Ignoring 404 errors completely also blocks transfers somehow.

ncw:
I have told Blomp about this problem but I haven't heard back yet.

Do you want to change the `retry_403` option into `retry_errors` and change its type into `fs.CommaSepList` to support multiple error codes to make experimentation easier? I'd be willing to merge that and the list_chunks option.

Or do you think we should wait to hear back from Blomp. I'll give them a nudge now.

bodqhrohro:
> Do you want to change the `retry_403` option into `retry_errors` and change its type into `fs.CommaSepList` to support multiple error codes to make experimentation easier?

Yeah, that would be more flexible.

bodqhrohro:
I'm also concerned about the fact it's passed in so many places for different requests. Which is probably the cause of that 404 problem, as 404 is genuine for *some* requests, but not for any. Can that be made flexible too without overcomplicating it too much?

Issue #7896, opened 2024-06-08
backend: pcloud: support block level sync (upload only)


gfelbing:
#### The associated forum post URL from `https://forum.rclone.org`

- Block level sync in general was requests in the forum [here](https://forum.rclone.org/t/pcloud-block-level-sync/43455)
- It helps fixing the the rate limit issues discussed [here](https://forum.rclone.org/t/how-to-handle-pcloud-rate-limit/31279/19):
  - The described work around only helps when uploading multiple smaller files by increasing `--transfers`
  - For single large files, I haven't found any workaround.
  - Block level sync seems to be slightly faster, but the main point is: With this, we can resume failed uploads of large files (e.g. due to timeouts caused by rate limit)

#### What is your current rclone version (output from `rclone version`)?

- master

#### What problem are you are trying to solve?

- Implement object.SetModTime to reduce unecessary re-uploads on mtime mismatch
  - This is an undocumented API feature (reverse-engineered), using server-side copy with same target, only changing metadata.
  - This is necessary for block level support, otherwise uploaded files will have mtime from time of upload.
- Implement new option `upload_chunk_size`
  - Enables chunk-wise uploads of files.
  - Successful uploaded chunks are skipped when upload needs retry, thanks to block-wise hashing capability.

Focusing on the upload in this issue, block level download can easily be added in follow-up.

---

tl;dr

Uploading large files to pcloud using the uploadfile API is error prone, as any connection reset / package loss will cause the full file to be retried. 
Additionally, I can see that pcloud applies some sort of rate limiting, causing large uploads to be throttled down. 
(Their "support" claims that there is no such rate limit and blames my ISP. But by increasing the `--transfers` with multiple files, I can multiply the throughput by amount of transfers until I reach my ISPs upload lim. Also 32kbps is such a round number, I doubt that its coincidence.)

In my case, this renders it almost impossible to upload any file larger than ~200MB via rclone.

Fortunately, pcloud offers a [fileops API](https://docs.pcloud.com/methods/fileops/), which enables chunked uploads on block level 🎉
PR #7895  adds an advanced option `upload_chunk_size` to the pcloud backend.
When set, files will be uploaded in chunks using the fileops API. Otherwise, current behavior is used (backwards compatibility).

#### How do you think rclone should be changed to solve that?

I created a draft PR #7895 that implements it. 
[...]

PR #7949, opened 2024-07-13
pcloud: implement SetModTime and OpenWriterAt

gfelbing:

This is the second approach on #7896 by implementing OpenWriterAt instead of OpenChunkWriter (#7895), proposed [here](https://github.com/rclone/rclone/issues/7896#issuecomment-2227097288) 

- Issue #7896 

#### Checklist

### DIFF
--- a/backend/pcloud/pcloud.go
+++ b/backend/pcloud/pcloud.go
@@ -1094,9 +1094,49 @@ func (o *Object) ModTime(ctx context.Context) time.Time {
 
 // SetModTime sets the modification time of the local fs object
 func (o *Object) SetModTime(ctx context.Context, modTime time.Time) error {
-	// Pcloud doesn't have a way of doing this so returning this
-	// error will cause the file to be re-uploaded to set the time.
-	return fs.ErrorCantSetModTime
+	filename, directoryID, err := o.fs.dirCache.FindPath(ctx, o.Remote(), true)
+	if err != nil {
+		return err
+	}
+	return o.setModTime(ctx, fileIDtoNumber(o.id), filename, directoryID, modTime)
+}
+
+func (o *Object) setModTime(
+	ctx context.Context,
+	fileID, filename, directoryID string,
+	modTime time.Time,
+) error {
+	filename = o.fs.opt.Enc.FromStandardName(filename)
+	opts := rest.Opts{
+		Method:           "PUT",
+		Path:             "/copyfile",
+		Parameters:       url.Values{},
+		TransferEncoding: []string{"identity"}, // pcloud doesn't like chunked encoding
+		ExtraHeaders: map[string]string{
+			"Connection": "keep-alive",
+		},
+	}
+	opts.Parameters.Set("fileid", fileID)
+	opts.Parameters.Set("folderid", dirIDtoNumber(directoryID))
+	opts.Parameters.Set("toname", filename)
+	opts.Parameters.Set("tofolderid", dirIDtoNumber(directoryID))
+	opts.Parameters.Set("ctime", strconv.FormatInt(modTime.Unix(), 10))
+	opts.Parameters.Set("mtime", strconv.FormatInt(modTime.Unix(), 10))
+
+	result := &api.ItemResult{}
+	err := o.fs.pacer.CallNoRetry(func() (bool, error) {
+		resp, err := o.fs.srv.CallJSON(ctx, &opts, nil, result)
+		err = result.Error.Update(err)
+		return shouldRetry(ctx, resp, err)
+	})
+	if err != nil {
+		return fmt.Errorf("update mtime: copyfile: %w", err)
+	}
+	if err := o.setMetaData(&result.Metadata); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // Storable returns a boolean showing whether this object storable
--- a/backend/pcloud/pcloud.go
+++ b/backend/pcloud/pcloud.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
@@ -146,7 +147,8 @@ we have to rely on user password authentication for it.`,
 			Help:       "Your pcloud password.",
 			IsPassword: true,
 			Advanced:   true,
-		}}...),
+		},
+		}...),
 	})
 }
 
@@ -161,15 +163,16 @@ type Options struct {
 
 // Fs represents a remote pcloud
 type Fs struct {
-	name         string             // name of this remote
-	root         string             // the path we are working on
-	opt          Options            // parsed options
-	features     *fs.Features       // optional features
-	srv          *rest.Client       // the connection to the server
-	cleanupSrv   *rest.Client       // the connection used for the cleanup method
-	dirCache     *dircache.DirCache // Map of directory path to directory id
-	pacer        *fs.Pacer          // pacer for API calls
-	tokenRenewer *oauthutil.Renew   // renew the token on expiry
+	name         string                 // name of this remote
+	root         string                 // the path we are working on
+	opt          Options                // parsed options
+	features     *fs.Features           // optional features
+	ts           *oauthutil.TokenSource // the token source, used to create new clients
+	srv          *rest.Client           // the connection to the server
+	cleanupSrv   *rest.Client           // the connection used for the cleanup method
+	dirCache     *dircache.DirCache     // Map of directory path to directory id
+	pacer        *fs.Pacer              // pacer for API calls
+	tokenRenewer *oauthutil.Renew       // renew the token on expiry
 }
 
 // Object describes a pcloud object
@@ -317,6 +320,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		name:  name,
 		root:  root,
 		opt:   *opt,
+		ts:    ts,
 		srv:   rest.NewClient(oAuthClient).SetRoot("https://" + opt.Hostname),
 		pacer: fs.NewPacer(ctx, pacer.NewDefault(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
 	}
@@ -326,6 +330,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	f.features = (&fs.Features{
 		CaseInsensitive:         false,
 		CanHaveEmptyDirectories: true,
+		PartialUploads:          true,
 	}).Fill(ctx, f)
 	if !canCleanup {
 		f.features.CleanUp = nil
@@ -333,7 +338,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	f.srv.SetErrorHandler(errorHandler)
 
 	// Renew the token in the background
-	f.tokenRenewer = oauthutil.NewRenew(f.String(), ts, func() error {
+	f.tokenRenewer = oauthutil.NewRenew(f.String(), f.ts, func() error {
 		_, err := f.readMetaDataForPath(ctx, "")
 		return err
 	})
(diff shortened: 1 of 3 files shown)


fs/operations/copy.go
```diff
@@ -97,7 +98,12 @@ func (c *copy) checkPartial() (remoteForCopy string, inplace bool, err error) {
 	}
 	// Avoid making the leaf name longer if it's already lengthy to avoid
 	// trouble with file name length limits.
-	suffix := "." + random.String(8) + c.ci.PartialSuffix
+
+	// generate a stable random suffix: use the significant bytes of the hashed remote
+	hashBytes := sha256.Sum256([]byte(c.remote))
```
ncw, line 106: Making the suffix depend on the remote name will make it identical for all users. This means that if A and B are uploading at the same time then the resultant file won't be either A or B but a corrupted mix of both.

You could use `fs.Fingerprint(ctx, c.src, true)` to create a string which depends on the source in a reasonably robust fashion. I'm not sure this will solve your problem though.

ncw (commented):
I like the SetModTime and OpenWriterAt changes (though I didn't review the code in detail yet).

However I don't like the resume changes. Rclone backends don't resume, and it is very difficult to do and ensure data integrity (how are we 100% sure that the partial file uploaded is the same as the start of the file we are about to upload?).

Can you remove the resuming element from this please and we can get the SetModTime and OpenWriterAt changes in. Then we can discuss resume. There is lots of prior art for discussing resume in the issues!

Thank you :-)

fs/operations/copy.go
gfelbing: Thanks, I'll consider it when we work on the resume part.
I removed all resume-related changes from this PR, including the stable partial filename.

gfelbing:
@ncw , thanks for the first review.
I removed the resume related changes, I see your point of discussing it separately.
This PR would be done with the reduced scope then.
Tests are currently running, I'll upload the result as soon as they are done.

backend/pcloud/pcloud.go
```diff
@@ -326,14 +330,18 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	f.features = (&fs.Features{
 		CaseInsensitive:         false,
 		CanHaveEmptyDirectories: true,
+		OpenWriterAt: func(ctx context.Context, remote string, size int64) (fs.WriterAtCloser, error) {
+			return newWriterAt(ctx, remote, f, size)
```
ncw, line 334: Normally we make `OpenWriterAt` a method on `*Fs` - then it can easily read the config etc.

backend/pcloud/writer_at.go
```diff
@@ -0,0 +1,253 @@
...
+	ctx    context.Context
+	client *rest.Client
+	fs     *Fs
+	size   int64
+	remote string
+	fd     int64
+	fileID int64
+}
+
+func newWriterAt(
```
ncw, line 28: This should be a method on `*Fs`

ncw (commented):
I think that looks great - thank you :-)

Can you make the OpenWriterAt a method on Fs then it will fit with the other implementations and save a few lines of code. You can leave the implementation in `writer_at.go` - that is fine.

backend/pcloud/writer_at.go
gfelbing: done

backend/pcloud/pcloud.go
gfelbing: done

gfelbing:
@ncw thanks for the review! I implemented your suggestions.
While trying to run the tests, I ran into one final thing: I need to adjust the default chunk- & buffersize for the pcloud backend.
According to the docs, the backends are supposed to configure their natural chunk sizes themselves.
I tried to look up how others do it (e.g. smb, azurefiles), but I struggled to find the place were they override the values for `--multi-thread-chunk-size` and `--multi-thread-write-buffer-size`. Can you give me a pointer on how this works?

ncw:
> thanks for the review! I implemented your suggestions. 

Great :-)

> While trying to run the tests, I ran into one final thing: I need to adjust the default chunk- & buffersize for the pcloud backend. According to the docs, the backends are supposed to configure their natural chunk sizes themselves. I tried to look up how others do it (e.g. smb, azurefiles), but I struggled to find the place were they override the values for `--multi-thread-chunk-size` and `--multi-thread-write-buffer-size`. Can you give me a pointer on how this works?

The option for configuring natural chunk sizes is in the `OpenChunkWriter` interface. There isn't one in the `OpenWriterAt` interface as it is a more general interface.

Can you explain more about what are good and bad chunk sizes for pcloud and what are the consequences of getting it wrong?

gfelbing:
@ncw I see, then we'd back to OpenChunkWriter :)

Actually, I am more happy with the OpenWriterAt, as it leaves the chunking logic to operations.

It's just the parameters, that need to have better defaults for pcloud:

* Smaller cutoff: the fileops API seems to work way better for medium sized images (10MB and up) already, while the file API works better for smaller sizes, especially when there are lots of them.
* Smaller bufersize: As the fileops API relies on TCP connections for handling file descriptors, it shows that there are more hickups when having longer running requests on the same connection, hence it shouldn't be too large. On the other hand, small many small buffers seem to be very slow. My experiments show that 1-5MB seem to work well.
* Chunksize vs Buffersize: Given a buffersize of 1-5MB and the bottleneck being TCP connections, I don't see an advantage of having a different chunk- than bufersize. I'd set them equally.

That being said, looking at the code: How would you suggest to implement it? IMHO the easiest change would be to make `operations.openChunkWriterFromOpenWriterAt` public and use it for pclouds OpenChunkWriter with backend specific parameters.

ncw:
> Actually, I am more happy with the OpenWriterAt, as it leaves the chunking logic to operations.

:-)

> It's just the parameters, that need to have better defaults for pcloud:

Could you do some experiments and measure throughput for various file sizes? I'd like to see the throughput for the default vs a set of tuned parameters.

We could put some hints in the docs about optimum parameters once we've done this.

> That being said, looking at the code: How would you suggest to implement it? IMHO the easiest change would be to make `operations.openChunkWriterFromOpenWriterAt` public and use it for pclouds OpenChunkWriter with backend specific parameters.

That is a great idea. Do you want to have an experiment with that? We could potentially use that idea to tune the other backends too.

gfelbing:
@ncw , I wrote a script to experiment with the parameters:

- Having test file sets of 32*1M, 16*2M, 8*4M, 4*8M, 2*16M, 1*32M files with random content (each set being 32M large, so a simple time on the rclone will do for throughput)
- Upload each set for every combination of multi thread parameters:
  - cutoff 1M, 2M, 4M, 8M, 16M, 32M
  - buffer size 1M, 2M, 4M, 8M, 16M, 32M
  - chunk size 1M, 2M, 4M, 8M, 16M, 32M
  - streams 1, 2, 4, 8

This makes me upload 162GB of random data, which can take a while...

```bash
#!/usr/bin/env bash

set -ex

if [[ "$1" == "generate" ]]; then
  # generate test files
  for i in {001..32}; do mkdir -p 1M && dd if=/dev/urandom of=1M/$i.rand bs=1M count=1 iflag=fullblock; done
  for i in {001..16}; do mkdir -p 2M && dd if=/dev/urandom of=2M/$i.rand bs=2M count=1 iflag=fullblock; done
  for i in {001..8}; do mkdir -p 4M && dd if=/dev/urandom of=4M/$i.rand bs=4M count=1 iflag=fullblock; done
  for i in {001..4}; do mkdir -p 8M && dd if=/dev/urandom of=8M/$i.rand bs=8M count=1 iflag=fullblock; done
  for i in {001..2}; do mkdir -p 16M && dd if=/dev/urandom of=16M/$i.rand bs=16M count=1 iflag=fullblock; done
  for i in {001..1}; do mkdir -p 32M && dd if=/dev/urandom of=32M/$i.rand bs=32M count=1 iflag=fullblock; done
fi

ALL_SIZES="1M 2M 4M 8M 16M 32M"
ALL_STREAMS="1 2 4 8"

rclone mkdir pcloud:upload-test

for CUTOFF in $ALL_SIZES; do
	for BUFFER_SIZE in $ALL_SIZES; do
		for CHUNK_SIZE in $ALL_SIZES; do
			for STREAMS in $ALL_STREAMS; do
			  for FILES in $ALL_SIZES; do
          rclone delete pcloud:upload-test
          LOG_FILE="$CUTOFF-$BUFFER_SIZE-$CHUNK_SIZE-$STREAMS-$FILES.log"
          { time rclone copy -vv \
            --multi-thread-streams $STREAMS \
            --multi-thread-cutoff $CUTOFF \
            --multi-thread-chunk-size $CHUNK_SIZE \
            --multi-thread-write-buffer-size $BUFFER_SIZE \
            $FILES pcloud:upload-test; } \
            >> $LOG_FILE 2>&1 
			  done
			done
		done
	done
done
```

ncw:
Are the results in? How did it go?

gfelbing:
Not quite, I am on vacation this week, can only run the script when my Starlink is on my RV and stable to produce similar results :)
I'll guess I'll be through next week.

gfelbing:
@ncw script is still ongoing (let it run today for 6h).
Still at cutoff 1M, in the mid off buffer size 2M.
I already created a small script to evaluate the log files, so far it shows a strong correlation (0.69) between file size and time and no correlation between chunk size (-0.01) or concurrent streams (0,07) and time.

Chunk size is no surprise, as all had the same buffer size. That supports my thesis that they should be same.

Concurrent streams is interesting though, I am looking forward wether that changes with more data (especiall different buffer sizes).

I'll give update as soon as the script is finished and post the full data.

ncw:
Great - and good idea measuring the correlation. Interested to hear the full results :-)

gfelbing:
Hey! Just an update: Still working on it.

- I moved the experiment from my local computer to a VPS server, in order to be a/ faster due to higher upload speed and b/ producing more stable results.
- I ran the experiment with the master branch first, as a baseline. This went fine so far.
- With the faster connection, I can reproduce a race condition with pcloud way more often: It seems that there is some sort of server side cache between the fileops API and the regular file API. 

The latter one results in many upload errors - as the checks after the upload uses the file API, while the upload with OpenWriterAt uses the fileops API. Errors are usually size not matching or checksum wrong. 
The upload itself worked properly though - if I look at the deleted files (luckily pcloud has a trash/revisioning system), the uploaded files are correct in size and checksum.
I'll give an update as soon as I narrowed down the exact race condition and try to find a solution.

ncw:
Interesting and fairly common. A sleep and retry is the usual solution here.

gfelbing:
Thanks for the hint, but were to put it? With OpenWriterAt, the check that needs retry happens outside in operations, and within (e.g. during OpenWriterAt.Close) I have no information about the expected checksum of the whole file on which I could retry.

ncw:
> Thanks for the hint, but were to put it? With OpenWriterAt, the check that needs retry happens outside in operations, and within (e.g. during OpenWriterAt.Close) I have no information about the expected checksum of the whole file on which I could retry.

Can you detect the condition in the `Close` method of `OpenWriterAt`?

Maybe by reading the file size / modtime / id / something else and see if it is as you expect? If not retrying?

I'd try just putting a time.Sleep in the `Close` method to see if it can be fixed like that, which at least gives the scale of the problem. We should only use that as a fix as a last resort if we can't find something better though!

ncw:
I'd like to get this in v1.68, is it just the intermittent problem above that is outstanding?

gfelbing:
Yes, that's the only thing which is open.
Currently a bit busy, could check your suggested fix in a few days.

ncw:
No probs.

gfelbing:
> Can you detect the condition in the Close method of OpenWriterAt?

Unfortunately, the writer doesn't get any information about the full file being transmitted (e.g. size, hash). Therefore a check inside is currently not possible.

> Maybe by reading the file size / modtime / id / something else and see if it is as you expect? If not retrying?

Retrying inside the writer isn't really possible as well, as that would undermine the whole purpose of the writer itself, wouldn't it?

> I'd try just putting a time.Sleep in the Close method to see if it can be fixed like that, which at least gives the scale of the problem. We should only use that as a fix as a last resort if we can't find something better though!

It seems that this is sadly the only option currently. At least it does have good effect, the integration tests ([2024-08-17-15-46_test_all.tar.gz](https://github.com/user-attachments/files/16646679/2024-08-17-15-46_test_all.tar.gz)) went through successfully, the experiment above also continues to run.
I'll update the PR with the sleep and an according comment. IMHO this could be merged then, yet I'll still keep on working on a long term solution. But this can easily go into a follow up (e.g. I am currently experimenting with pclouds binary api instead of the http one, to have better control over the underlying connections).

ncw:
> > Can you detect the condition in the Close method of OpenWriterAt?
> 
> Unfortunately, the writer doesn't get any information about the full file being transmitted (e.g. size, hash). Therefore a check inside is currently not possible.
> 
> > Maybe by reading the file size / modtime / id / something else and see if it is as you expect? If not retrying?
> 
> Retrying inside the writer isn't really possible as well, as that would undermine the whole purpose of the writer itself, wouldn't it?

In the Close method, close the file using the special API. Then using the normal API read the size and or checksum and sleep and retry until correct. Only then return from the Close method.

Hopefully that method will work.

You can use the pacer to do the retries and error on too many tries.

gfelbing:
While getting the size/checksum is no problem, how do I determine if it's `correct`? The upload happened via the chunkwriter, which has never seen the source file size/checksum. Am I missing something?

gfelbing:
Ah, forgot that the WriterAt gets the size on creation time. 
That I can use to compare against. 
Updated the PR, looks good on first glance. Tests are currently running.

ncw:
How did the test go?

gfelbing:
Sorry for the cliffhanger :)
The edge-case for unknown (negative sizes) was missing.
Ran the tests again, worked: 
[2024-08-21-195153_test_all.tar.gz](https://github.com/user-attachments/files/16696214/2024-08-21-195153_test_all.tar.gz)

ncw:
Can you fix the lint error? I think you'll need to delete an unused method.

Then I'll merge - thank you :-)

gfelbing:
Updated PR to fix the linter issue (local run succeeded)

ncw (approved):
Thank you - super 🙂 

Will merge now

PR #7953, opened 2024-07-15
systemd: prevent unmount rc command from sending a STOPPING=1 sd-notify message

AThePeanut4:
#### What is the purpose of this change?

#7540 has been annoying me for quite a long time, so I decided to try my hand at a fix. It turned out to be surprisingly simple - just move the sd-notify calls out of `(m *MountPoint) Wait()` and into the command `Run` handlers used by the `mount` commands in `NewMountCommand`. 

~~I also made some minor changes to the `serve docker` command, so that the `READY=1` notification is only sent _after_ the socket is created and a connection is made. This means that if there are any units with an `After` dependency on the rclone unit, they will only be started once the socket exists and is accepting connections.~~ Split to #7954

#### Was the change discussed in an issue or in the forum before?

Fixes #7540

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [x] I have added tests for all changes in this PR if appropriate.
- [x] I have added documentation for the changes if appropriate.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [x] I'm done, this Pull Request is ready for review :-)

PR #7954, opened 2024-07-15
serve docker: send systemd READY=1 message after socket is ready

AThePeanut4:

In the `serve docker` command, the `READY=1` sd-notify notification is currently sent just before `NewDriver` returns. At this point, rclone has not yet created or started listening on the socket. This change moves the ready notification later, until after the socket has been created and opened. This means any units with an After dependency on the rclone service can reliably expect the served socket to exist and be accepting connections.

Split from #7953

#### Checklist

### DIFF
--- a/cmd/serve/docker/docker.go
+++ b/cmd/serve/docker/docker.go
@@ -4,6 +4,7 @@ package docker
 import (
 	"context"
 	_ "embed"
+	"net"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


jonasgeiler:
It would be cool if we could pair this PR with mine over at #7929. Both valuable improvements to the `serve docker` command in my opinion.

PR #7975, opened 2024-07-26
fs/log: decouple `logrus` dependency -fixes #3426

### LINKED ISSUE #3426: Eliminate the tight coupling with logrus

5195075

- pulling logrus as a dependency when it's not needed
- two places to configure the logging level, logrus and `fs` package

* Rclone should keep using it's own logging abstractions.
* Logrus integration code should go to a separate package ex. `fs/log/logrus` and no other package should import logrus directly.
* The `fs.LogPrintf` should be what `fs.LogPrint` is now and it should be possible to change it.
* The `fs/log/logrus` package should export a function compatible with `fs.LogPrintf`
* Logrus should be enabled in main package by assigning `fs.LogPrintf`.

ParasRaba155:
fs/log.go logging methods now use slog package instead of logrus, All the logrus config has been moved to fs/logrus package No external package will now be using logrus directly instead should use fs/logrus package

- Decouple the `logrus` from `fs/log.go`
- All the `logrus` helper functions moved to `fs/logrus` package

#3426

#### Checklist

### DIFF
--- a/cmd/bisync/bilib/output.go
+++ b/cmd/bisync/bilib/output.go
@@ -6,13 +6,13 @@ import (
 	"log"
 
 	"github.com/rclone/rclone/fs"
-	"github.com/sirupsen/logrus"
+	"github.com/rclone/rclone/fs/logrus"
 )
 
 // CaptureOutput runs a function capturing its output.
 func CaptureOutput(fun func()) []byte {
 	logSave := log.Writer()
-	logrusSave := logrus.StandardLogger().Writer()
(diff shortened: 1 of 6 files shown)


ncw:
Hi @ParasRaba155, thank you for this, and apologies it sat so long.

rclone no longer depends on `logrus` - it's gone from the source entirely, surviving only as an indirect transitive dependency in `go.mod`. So the decoupling this PR set out to do is effectively already done, and I'm going to close it on that basis.

Much appreciated regardless. If you spot anything still referencing it, do say and we'll take another look.

Issue #7818, opened 2024-05-01
Hasher: trust cached sum


giovi2:
Hello,
I'm opening this feature request issue following @nielash suggestion under [my post in the forum](https://forum.rclone.org/t/hasher-trust-cached-sum/45386).

I would like to ask for a new flag for the hasher backend that will make the database _read-only_ when downloading a file from the remote.
This might be useful to keep track of data corruption while working with backends that don't support checksums.

To further explain my point,
as it is now, while uploading a file to a remote, the hasher will hash on the fly and store it in the database.
If the hash of the file changes afterwards (be it for network issues after the sum or anything else linked to the remote) and you download it again, the hasher will check it against the one in the database and update the entry with the new hash, no retries are made (`-vv` shows as such).

So my idea was: instead of simply update the entry, rclone would retry to download the file until the given number of `--retries` (since the issue might still be during the downloading process and the file is actually fine on the remote) and eventually fail the transfer with the error "`hashes differ`."

This way the entry in the database won't be updated and you can read from logs that file changed from the one you *had* at the beginning.

In addition to this, @nielash suggested the addition of another interesting flag such as a _fully-read-only_ mode of the database working exclusively with manually imported hashes.

NiTr0gLiTcH:
hope this is still alive, it sounds useful

ncw:
Does someone want to have a go at this? Adding a `read_only` flag to the hasher backend sounds like a useful addition.

I would first add a new method to `DB` in `lib/kv`, say `ReadOnly(bool)` which sets a read only flag in `DB`.

https://github.com/rclone/rclone/blob/612c717ea0eed3f646095ed57620fd05b2a2c20b/lib/kv/bolt.go#L29

Then in this function, check the flag and return an error if `forWrite` was true

https://github.com/rclone/rclone/blob/612c717ea0eed3f646095ed57620fd05b2a2c20b/lib/kv/bolt.go#L135

Define the error `ErrReadOnly` here

https://github.com/rclone/rclone/blob/612c717ea0eed3f646095ed57620fd05b2a2c20b/lib/kv/types.go#L12

And write a test.

That should be only a few lines of code and will make definitely sure the db is read only.

In a separate commit define the flag `read_only` here

https://github.com/rclone/rclone/blob/612c717ea0eed3f646095ed57620fd05b2a2c20b/backend/hasher/hasher.go#L53

Add it here as `ReadOnly`

[...]

PR #7981, opened 2024-07-31
Feat/readonly db

ParasRaba155:

- Add a `read_only` flag in the `hasher` baekend

#7818 

#### Checklist

### DIFF
--- a/backend/hasher/hasher.go
+++ b/backend/hasher/hasher.go
@@ -50,6 +50,10 @@ func init() {
 			Advanced: true,
 			Default:  fs.SizeSuffix(0),
 			Help:     "Auto-update checksum for files smaller than this size (disabled by default).",
+		}, {
+			Name:    "read_only",
+			Default: false,
+			Help:    "Set the db in read only mode",
 		}},
 	})
 }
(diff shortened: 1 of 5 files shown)


ncw (commented):
Looks very good - well done :-)

Can you put `Fixes #7818`  in the commit message?

Are you happy with this fix in your testing?

ParasRaba155:
I will fix up the commit msg, but before that I think the test does its job but it's not adequate.
I wanted to perform something actual but the `Op` interface is implemented in `backed/hasher` so couldn't come up with something better.

Do you have any idea about better test, then please do share. I think this test just barely does the job

ncw:
I think the `lib/kv` test is solid.

You could test the `read_only` flag in `hasher_internal_test.go` - create a new backend with NewFs and just check that when you do something which should write to the db it returns the correct error. I think that is sufficient to prove that the read only flag is set in `lib/kv`.

nielash (commented):
Nice work. I added a couple inline comments for spots I think may need another look.

Also, have we thought through (and tested) what should happen when an object is server-side moved or deleted in read-only mode?

Regarding your comment [here](https://github.com/rclone/rclone/issues/7818#issuecomment-2260755506), if you did want to implement the "partial" mode, I think what you'd want to do is add a parameter to `putHashes` so that the write is performed when called from `Put`, but not when called from `Hash` or `Open`.

PR #7994, opened 2024-08-08
Revert "help: global flags help command now takes glob filter"

albertony:
This reverts commit c6352231e43c32bf139cea64d69a4528ef93dedb from https://github.com/rclone/rclone/pull/7404.

See https://github.com/rclone/rclone/pull/7404#issuecomment-2276089200 and below.

#### Checklist

### DIFF
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -12,7 +12,6 @@ import (
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/config/configflags"
 	"github.com/rclone/rclone/fs/config/flags"
-	"github.com/rclone/rclone/fs/filter"
 	"github.com/rclone/rclone/fs/filter/filterflags"
 	"github.com/rclone/rclone/fs/log/logflags"
 	"github.com/rclone/rclone/fs/rc/rcflags"
@@ -67,12 +66,12 @@ var (
 
 // Show the flags
 var helpFlags = &cobra.Command{
-	Use:   "flags [<filter>]",
+	Use:   "flags [<regexp to match>]",
 	Short: "Show the global flags for rclone",
 	Run: func(command *cobra.Command, args []string) {
 		command.Flags()
 		if len(args) > 0 {
-			re, err := filter.GlobStringToRegexp(args[0], false)
+			re, err := regexp.Compile(`(?i)` + args[0])
 			if err != nil {
 				log.Fatalf("Invalid flag filter: %v", err)
 			}


ncw:
I think the idea of using globs is probably better than using regexp for non developers but I don't think they should be rooted, so we should implicitly put `*` on the start and end if not present.

What do you think?

albertony:
I guess what you mean is keep it as it was, but disabling the `addAnchors` option, i.e. not wrapping the regex generated from glob with `^`/`$`, as I wrote a test for in `TestGlobStringToRegexpWithoutAnchors`?

One possible negative consequence is that it is cumbersome to restrict the matching to the whole string, if that is what the user wants. With paths you have the `/` path separator that is effective in this sense, i.e. `file.txt` can be changed to `/file.txt` to avoid matching `/somefile.txt`. However with "string globs", that has no special treatment of path separator `/` and double asterisk `**`, if a user want to avoid `string` matching `somestring` - this is not possible with the glob syntax, is it, so would have to switch to advanced mode regex `{{^string$}}`.. I think?

ncw:
> I guess what you mean is keep it as it was, but disabling the `addAnchors` option, i.e. not wrapping the regex generated from glob with `^`/`$`, as I wrote a test for in `TestGlobStringToRegexpWithoutAnchors`?

Yes

Searching the help is useless anchors also

> One possible negative consequence is that it is cumbersome to restrict the matching to the whole string, if that is what the user wants. With paths you have the `/` path separator that is effective in this sense, i.e. `file.txt` can be changed to `/file.txt` to avoid matching `/somefile.txt`. However with "string globs", that has no special treatment of path separator `/` and double asterisk `**`, if a user want to avoid `string` matching `somestring` - this is not possible with the glob syntax, is it, so would have to switch to advanced mode regex `{{^string$}}`.. I think?

I use this command to find flags that I want to see the detail of or can't quite remember the full name so I think showing more flags is OK here. So if you search for `header` you'll get `header-upload` and `header-download` which is probably what the user wanted.

Also can we get it not to show the usage?

```
$ rclone help flags checkers
Usage:
  rclone help flags [<filter>] [flags]

Flags:
      --group string   Only include flags from specific group
  -h, --help           help for flags
      --name           Apply filter only on flag names


# Performance Flags

Flags helpful for increasing performance.

      --checkers int   Number of checkers to run in parallel (default 8)


Additional help topics:

```

I know it did that before but the usage has grown and it has become distracting!  Something like this would be perfect

```
$ rclone help flags checkers
# Performance Flags

Flags helpful for increasing performance.

      --checkers int   Number of checkers to run in parallel (default 8)
```

Maybe we should just filter the flags not as part of showing the help and just show all the flags found all in one section.

```
$ rclone help flags checkers
      --checkers int   Number of checkers to run in parallel (default 8)
```

albertony:
Good points.

Regarding filtering:

I was thinking mostly about the `listremotes` command in my last thoughts about anchor in filters, where there are much less items to filter and I was thinking forced anchoring would be preferrable to be able to be accurate. I see your point with the `help` command. (I'm thinking the rules should be the same for both, so that we don't have several different rules around in different parts of rclone). I have now pushed a new version where its not adding the anchors, and also it is forcing ignore case. I find this resembles more of a plain text search, which I'm starting to like.. :) Search for `header` matches `header-upload` and `header-download`. And the glob (and regex) syntax can still be utilized for more accurate search like `s3*encoding`. However, I think it is a bit "fuzzy", easy to use in simple cases and for new users, but if challenged with more complex cases from advanced users it may be a bit difficult to defend - if you see what I mean.

Regarding help flags output:

I like your suggestions, and did some experimenting... I'm still using the existing cobra/template functionality, did not want to rewrite this command to output in some new way, and did not want to refactor the existing code too much either.

If only showing the actual flags, this leads to some strange indentation, because the flags for each group are indented independently:

```
      --max-backlog int   Maximum number of objects in sync or check backlog (default 10000)
      --max-backlog int   Maximum number of objects in sync or check backlog (default 10000)
      --log-file string            Log everything to this file
      --log-format string          Comma separated list of log format options (default "date,time")
      --log-level LogLevel         Log level DEBUG|INFO|NOTICE|ERROR (default NOTICE)
      --log-systemd                Activate systemd integration for the logger
      --stats-log-level LogLevel   Log level to show --stats output DEBUG|INFO|NOTICE|ERROR (default INFO)
      --syslog                     Use Syslog for logging
      --syslog-facility string     Facility for syslog, e.g. KERN,USER (default "DAEMON")
      --use-json-log               Use json log format
      --box-owned-by string          Only show items owned by the login (email address) passed in
      --s3-sdk-log-mode Bits         Set to debug the SDK (default Off)
      --swift-endpoint-type string   Endpoint type to choose from the service catalogue (OS_ENDPOINT_TYPE) (default "public")
      --swift-user string            User name to log in (OS_USERNAME)
      --swift-user-id string         User ID to log in - optional - most swift systems use user and leave this blank (v3 auth) (OS_USER_ID)
```

Instead I created new and refactored the template and output strings a bit, to get a more condensed output, and more "command-output-style" and less "markdown-docs-style". A couple of examples

First an example of existing, where "Usage" and "Flags" headers look nice in console, while the filter group header "Copy Flags" uses Markdown, so there is a mix.

```
> .\rclone.exe copy -h
Usage:
  rclone copy source:path dest:path [flags]

Flags:
      --create-empty-src-dirs   Create empty source dirs on destination after copy
  -h, --help                    help for copy


# Copy Flags

Flags for anything which can copy a file.

      --check-first                                 Do all the checks before starting transfers
  -c, --checksum                                    Check for changes with size & checksum (if available, or fallback to size only)
      --compare-dest stringArray                    Include additional server-side paths during comparison
      --copy-dest stringArray                       Implies --compare-dest but also copies files from paths into destination
      --cutoff-mode HARD|SOFT|CAUTIOUS              Mode to stop transfers when reaching the max transfer limit HARD|SOFT|CAUTIOUS (default HARD)
...
```

My new version changes it to a bit more condenced, not Markdown, and consistent:
```
> .\rclone.exe copy -h
Usage:
  rclone copy source:path dest:path [flags]

Flags:
      --create-empty-src-dirs   Create empty source dirs on destination after copy
  -h, --help                    help for copy

Flags for anything which can copy a file (flag group Copy):
      --check-first                                 Do all the checks before starting transfers
  -c, --checksum                                    Check for changes with size & checksum (if available, or fallback to size only)
      --compare-dest stringArray                    Include additional server-side paths during comparison
      --copy-dest stringArray                       Implies --compare-dest but also copies files from paths into destination
      --cutoff-mode HARD|SOFT|CAUTIOUS              Mode to stop transfers when reaching the max transfer limit HARD|SOFT|CAUTIOUS (default HARD)
...
```

Example from help flags with filter, where I still show the flag group headers, but not the other general help details:
```
> .\rclone.exe help flags log
Flags for anything which can copy a file (flag group Copy):
      --max-backlog int   Maximum number of objects in sync or check backlog (default 10000)

Flags used for `rclone check` (flag group Check):
      --max-backlog int   Maximum number of objects in sync or check backlog (default 10000)

Logging and statistics (flag group Logging):
      --log-file string            Log everything to this file
      --log-format string          Comma separated list of log format options (default "date,time")
      --log-level LogLevel         Log level DEBUG|INFO|NOTICE|ERROR (default NOTICE)
      --log-systemd                Activate systemd integration for the logger
      --stats-log-level LogLevel   Log level to show --stats output DEBUG|INFO|NOTICE|ERROR (default INFO)
      --syslog                     Use Syslog for logging
      --syslog-facility string     Facility for syslog, e.g. KERN,USER (default "DAEMON")
      --use-json-log               Use json log format

Backend-only flags (these can be set in the config file also) (flag group Backend):
      --box-owned-by string          Only show items owned by the login (email address) passed in
      --s3-sdk-log-mode Bits         Set to debug the SDK (default Off)
      --swift-endpoint-type string   Endpoint type to choose from the service catalogue (OS_ENDPOINT_TYPE) (default "public")
      --swift-user string            User name to log in (OS_USERNAME)
      --swift-user-id string         User ID to log in - optional - most swift systems use user and leave this blank (v3 auth) (OS_USER_ID)
```

Another example, where the new option `--group` is used to limit to a single flag group, and then I do actually drop the flag header:
```
> .\rclone.exe help flags log --group logging
      --log-file string            Log everything to this file
      --log-format string          Comma separated list of log format options (default "date,time")
      --log-level LogLevel         Log level DEBUG|INFO|NOTICE|ERROR (default NOTICE)
      --log-systemd                Activate systemd integration for the logger
      --stats-log-level LogLevel   Log level to show --stats output DEBUG|INFO|NOTICE|ERROR (default INFO)
      --syslog                     Use Syslog for logging
      --syslog-facility string     Facility for syslog, e.g. KERN,USER (default "DAEMON")
      --use-json-log               Use json log format
```

ncw:
> I was thinking mostly about the `listremotes` command in my last thoughts about anchor in filters, where there are much less items to filter and I was thinking forced anchoring would be preferrable to be able to be accurate. I see your point with the `help` command. (I'm thinking the rules should be the same for both, so that we don't have several different rules around in different parts of rclone). I have now pushed a new version where its not adding the anchors, and also it is forcing ignore case. I find this resembles more of a plain text search, which I'm starting to like.. :)

Yes, I think a plain text search is the right concept to be aiming for.

> Search for `header` matches `header-upload` and `header-download`. And the glob (and regex) syntax can still be utilized for more accurate search like `s3*encoding`. However, I think it is a bit "fuzzy", easy to use in simple cases and for new users, but if challenged with more complex cases from advanced users it may be a bit difficult to defend - if you see what I mean.

In general there will only be a few matches so I think a slightly fuzzy match giving a few more things to look at is fine.

> I like your suggestions, and did some experimenting... I'm still using the existing cobra/template functionality, did not want to rewrite this command to output in some new way, and did not want to refactor the existing code too much either.
> 
> If only showing the actual flags, this leads to some strange indentation, because the flags for each group are indented independently:

Yes :-(

> Instead I created new and refactored the template and output strings a bit, to get a more condensed output, and more "command-output-style" and less "markdown-docs-style".

We need to make sure we don't break `make commanddocs` which uses exactly that markdown style to make the online help! Eg [cat](https://rclone.org/commands/rclone_cat/). I'm fine with the output changing, but it still has to parse as markdown.

I think your examples look very nice and this is definitely going in the right direction.

albertony:
> If only showing the actual flags, this leads to some strange indentation, because the flags for each group are indented independently:

And duplicates, I forgot to mention, in case you did not notice: `--max-backlog` is included in both Copy and Check groups, so when listing flags without groups it just looks like a bug. Could of course handle these things if we wanted to, but not sure I do - now.. :)

> We need to make sure we don't break make commanddocs which uses exactly that markdown style to make the online help!

Yes, I thought of that. Haven't tested it, so not entirely sure, but I'm under the impression it is covered:
- Command docs markdown are generated separately without relying on the cobra templates: https://github.com/rclone/rclone/blob/bfa3278f3058aa9efea30c45ed97fbed601a9d78/cmd/gendocs/gendocs.go#L167
- Global flags page have its own template: https://github.com/rclone/rclone/blob/bfa3278f3058aa9efea30c45ed97fbed601a9d78/cmd/help.go#L236

ncw:
> > If only showing the actual flags, this leads to some strange indentation, because the flags for each group are indented independently:
> 
> And duplicates, I forgot to mention, in case you did not notice: `--max-backlog` is included in both Copy and Check groups, so when listing flags without groups it just looks like a bug. Could of course handle these things if we wanted to, but not sure I do - now.. :)

Flags are allowed to be in more than one group. I don't think we should attempt to deal with that though!

> > We need to make sure we don't break make commanddocs which uses exactly that markdown style to make the online help!
> 
> Yes, I thought of that. Haven't tested it, so not entirely sure, but I'm under the impression it is covered:

Whew! Worth a verify but good!

albertony:
Found an issue in the templates I've been tweaking already, so fixed that as well with a new commit.

The issue is the "Additional help topics:" header, which in existing version would always be shown for commands that had any inherited flags (flag groups), even if they had no help subcommands. It would appear as if it was a header for `Use "rclone [command] --help" for more information about a command.` etc. but it was not really. Related, this also meant that any commands without inherited flags but with help subcommands would not get these listed. Not sure there are any commands with such help subcommands, probably this is just something kept from a default template? 

Existing version:

```
> rclone.exe copy -h
...
Flags for listing directories (flag group Listing):
      --default-time Time   Time to show if modtime is unknown for files and directories (default 2000-01-01T00:00:00Z)
      --fast-list           Use recursive list if available; uses more memory but fewer transactions


Additional help topics:

```

```
> rclone.exe about -h
...
Flags:
      --full   Full numbers instead of human-readable
  -h, --help   help for about
      --json   Format output as JSON

```

Anyway, I fixed this so that "Additional help topics:" are only shown if there are help subcommands, and then these are listed, regardless of any inherited flags.

New version:

```
> rclone.exe copy -h
...
Flags for listing directories (flag group Listing):
      --default-time Time   Time to show if modtime is unknown for files and directories (default 2000-01-01T00:00:00Z)
      --fast-list           Use recursive list if available; uses more memory but fewer transactions

```

```
> rclone.exe about -h
...
Flags:
      --full   Full numbers instead of human-readable
  -h, --help   help for about
      --json   Format output as JSON

```

albertony:
This initially small revert-PR blew up into something different and bigger... I don't know if I should split out all the command output commits into a separate PR, and leave the single glob related commit here? I was about to do that, but then stopped because all our discussions are in this PR and I wasn't sure if it would be more or less confusing...

Just let me know if you want me to split it.

If not, you may do a review now.

By the way, during the work on this I stumbled across a couple of other somewhat related issues that I did create new PRs for:
- https://github.com/rclone/rclone/pull/8001
- https://github.com/rclone/rclone/pull/8002

ncw (approved):
I think all these changes look good and we should merge them before they bitrot!

Please merge when ready.

Thank you :-)

albertony:
Great, will do. 😆

I agree, at this point its better to get it merged and then see how the beta version works and the tip.rclone.org site looks after that, and we'll iterate further if more issues pops out - just ping me.

PS: There seem to be some small issue in the build: Lint overview page shows "Cannot open: File exists", and in the log:

> Error: /usr/bin/tar: ../../../go/pkg/mod/golang.org/x/telemetry/config@v0.28.0/doc.go: Cannot open: File exists
/usr/bin/tar: ../../../go/pkg/mod/golang.org/x/telemetry/config@v0.28.0/LICENSE: Cannot open: File exists
/usr/bin/tar: ../../../go/pkg/mod/golang.org/x/telemetry/config@v0.28.0/config.json: Cannot open: File exists
/usr/bin/tar: ../../../go/pkg/mod/golang.org/x/telemetry/config@v0.28.0/go.mod: Cannot open: File exists
/usr/bin/tar: Exiting with failure status due to previous errors
Warning: Failed to restore: "/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2
Cache not found for input keys: golangci-lint-ubuntu[22](https://github.com/rclone/rclone/actions/runs/10405275663/job/28815661408#step:5:23)-go1.23.0-202433-ce1bb5155b22806b05070b3b3b8d9e2922ed1c689e1be5d359e5593adecdc607, golangci-lint-ubuntu22-go1.[23](https://github.com/rclone/rclone/actions/runs/10405275663/job/28815661408#step:5:24).0-202433-

It is only the cache restore that fails, I think, so the linting (and build) proceeds regardless.

Probably related to the new go version 1.23, and seeing the reference to telemetry perhaps related to the fact that this now is a built-in toolchain command `go telemetry` while before external package `go run golang.org/x/telemetry/cmd/gotelemetry@latest`... Though I don't see what the actual problem is..

ncw:
> PS: There seem to be some small issue in the build: Lint overview page shows "Cannot open: File exists", and in the log:

I saw that. I don't think it is new. Everything seems to be working!

PR #8001, opened 2024-08-12
Minor improvements to command docs layout

albertony:

A few minor improvements to command docs layout:
- Command docs for root rclone command (https://rclone.org/commands/rclone/) did not outdent header levels, so it was inconsistent with the rest (including different toc that included the top level command etc).
- Command docs (except root rclone command) had the "# SEE ALSO" header as top level header, same as the primary command header (e.g. "# rclone copy"), and it would therefore not be shown in toc.
- Make option group headers sub-headers of the command's own "Options" header. Previously these were at the same level, e.g. copy command had headers: Headers, Copy Options, Important Options, etc. Then it could be a bit confusing what the first "Headers" section contains compared to the rest.
- The line "See the [global flags page](/flags/) for global options not listed here." was written at the end of all options, but then it was effectively part of the last command group header, e.g. "Listing Options", while it is in fact common to all options. Moved it up to the main "Options" header.

#### Checklist

### DIFF
--- a/cmd/gendocs/gendocs.go
+++ b/cmd/gendocs/gendocs.go
@@ -159,33 +159,33 @@ rclone.org website.`,
 				}
 				doc := string(b)
 
-				var out strings.Builder
-				if groupsString := cmd.Annotations["groups"]; groupsString != "" {
-					groups := flags.All.Include(groupsString)
-					for _, group := range groups.Groups {
-						if group.Flags.HasFlags() {
-							_, _ = fmt.Fprintf(&out, "\n### %s Options\n\n", group.Name)
-							_, _ = fmt.Fprintf(&out, "%s\n\n", group.Help)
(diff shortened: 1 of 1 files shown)


cmd/gendocs/gendocs.go
```diff
@@ -165,6 +165,9 @@ rclone.org website.`,
 					if name != "rclone.md" {
 						return fmt.Errorf("internal error: failed to find cut points: startCut = %d, endCut = %d", startCut, endCut)
 					}
+					if endCut >= 0 {
+						doc = doc[:endCut] + "### See Also" + doc[endCut+12:]
```
ncw, line 169: I think the "SEE ALSO" got cribbed from man pages but it markdown headings are so much better :-)

ncw (approved):
Looks great - thank you :-)

Please merge when ready.

PR #8010, opened 2024-08-15
Codeql fixes

albertony:

This fixes a few CodeQL issues, reported in the project's Security page on GitHub. It would be good to at least get rid of the ones reported as `Critical` here, to avoid scaring away any potential users..

This should fix 2 of 3 `Critical` issues:
- https://github.com/rclone/rclone/security/code-scanning/129
- https://github.com/rclone/rclone/security/code-scanning/130

It should also fix 5 issues of type "Incorrect conversion between integer types" (severity `High`), all of that type except one, as well as one of 4 issues of type "Size computation for allocation may overflow" (severity `High`).

#### Checklist

### DIFF
--- a/backend/local/metadata.go
+++ b/backend/local/metadata.go
@@ -72,12 +72,12 @@ func (o *Object) parseMetadataInt(m fs.Metadata, key string, base int) (result i
 	value, ok := m[key]
 	if ok {
 		var err error
-		result64, err := strconv.ParseInt(value, base, 64)
+		parsed, err := strconv.ParseInt(value, base, 0)
 		if err != nil {
 			fs.Debugf(o, "failed to parse metadata %s: %q: %v", key, value, err)
 			ok = false
 		}
-		result = int(result64)
+		result = int(parsed)
 	}
 	return result, ok
 }
--- a/cmd/serve/docker/options.go
+++ b/cmd/serve/docker/options.go
@@ -2,6 +2,7 @@ package docker
 
 import (
 	"fmt"
+	"math"
 	"strings"
 
 	"github.com/rclone/rclone/cmd/mountlib"
@@ -272,10 +273,22 @@ func getVFSOption(vfsOpt *vfscommon.Options, opt rc.Params, key string) (ok bool
 		err = getFVarP(&vfsOpt.Umask, opt, key)
 	case "uid":
 		intVal, err = opt.GetInt64(key)
-		vfsOpt.UID = uint32(intVal)
+		if err == nil {
+			if intVal >= 0 && intVal <= math.MaxUint32 {
+				vfsOpt.UID = uint32(intVal)
+			} else {
+				err = fmt.Errorf("key %q (%v) overflows uint32", key, intVal)
+			}
+		}
 	case "gid":
 		intVal, err = opt.GetInt64(key)
-		vfsOpt.GID = uint32(intVal)
+		if err == nil {
+			if intVal >= 0 && intVal <= math.MaxUint32 {
+				vfsOpt.UID = uint32(intVal)
+			} else {
+				err = fmt.Errorf("key %q (%v) overflows uint32", key, intVal)
+			}
+		}
 
 	// non-vfs options
 	default:
--- a/fs/config/obscure/obscure.go
+++ b/fs/config/obscure/obscure.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 )
 
 // crypt internals
@@ -47,6 +48,9 @@ func crypt(out, in, iv []byte) error {
 // This is done by encrypting with AES-CTR
 func Obscure(x string) (string, error) {
 	plaintext := []byte(x)
+	if math.MaxInt32-aes.BlockSize < len(plaintext) {
+		return "", fmt.Errorf("value too large")
+	}
 	ciphertext := make([]byte, aes.BlockSize+len(plaintext))
 	iv := ciphertext[:aes.BlockSize]
 	if _, err := io.ReadFull(cryptRand, iv); err != nil {
--- a/lib/encoder/encoder.go
+++ b/lib/encoder/encoder.go
@@ -190,7 +190,7 @@ func (mask *MultiEncoder) Set(in string) error {
 		if bits, ok := nameToEncoding[part]; ok {
 			out |= bits
 		} else {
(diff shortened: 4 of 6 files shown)


albertony:
The last `Critical` issue I'm not entirely sure is real, and if so how to best fix it. Any suggestions?
- https://github.com/rclone/rclone/security/code-scanning/88

albertony:
I also left one issue "Incorrect conversion between integer types" (severity `High`), which I actually think is false positive due to the fact that the code checks `abs > size` etc, but that CodeQL misses that fact because several 32-bit `int`s are converted to `int64` and back.. If you agree we can select "Dismiss alert" with reason "False positive" in the GitHub Security page:
- https://github.com/rclone/rclone/security/code-scanning/134

backend/local/metadata.go
```diff
@@ -72,12 +72,12 @@ func (o *Object) parseMetadataInt(m fs.Metadata, key string, base int) (result i
 	value, ok := m[key]
 	if ok {
 		var err error
-		result64, err := strconv.ParseInt(value, base, 64)
+		parsed, err := strconv.ParseInt(value, base, 32)
```
ncw, line 75: I don't think this is the right fix as `int` is 64 bit on most platforms but 32 on some.

I think the right fix is probably using `strconv.Atoi` which hopefully deals with the sizes of integers properly. Looking at [the source of Atoi](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/strconv/atoi.go;l=242) I think it does the right thing.

cmd/serve/docker/options.go
```diff
@@ -272,10 +273,22 @@ func getVFSOption(vfsOpt *vfscommon.Options, opt rc.Params, key string) (ok bool
 		err = getFVarP(&vfsOpt.Umask, opt, key)
 	case "uid":
 		intVal, err = opt.GetInt64(key)
-		vfsOpt.UID = uint32(intVal)
+		if err == nil {
+			if intVal > 0 && intVal <= math.MaxInt32 {
```
ncw, line 277: I think the comparison should be
```
if intVal >= 0 && intVal <= math.MaxUint32
```

cmd/serve/docker/options.go
```diff
@@ -272,10 +273,22 @@ func getVFSOption(vfsOpt *vfscommon.Options, opt rc.Params, key string) (ok bool
...
+				vfsOpt.UID = uint32(intVal)
+			} else {
+				err = fmt.Errorf("key %q (%v) overflows uint32", key, intVal)
+			}
+		}
 	case "gid":
 		intVal, err = opt.GetInt64(key)
-		vfsOpt.GID = uint32(intVal)
+		if err == nil {
+			if intVal > 0 && intVal <= math.MaxInt32 {
```
ncw, line 286: And this one too

lib/encoder/encoder.go
```diff
@@ -190,7 +190,7 @@ func (mask *MultiEncoder) Set(in string) error {
 		if bits, ok := nameToEncoding[part]; ok {
 			out |= bits
 		} else {
-			i, err := strconv.ParseInt(part, 0, 64)
+			i, err := strconv.ParseUint(part, 0, 32)
```
ncw, line 193: This is another 32 vs 64 bit problem with uint

Not sure what the correct solution here is as there is no `strconf.Atou`

Rlone will go wrong on 32 bit platforms if there are more than 32 options though so we could make MultiEncoder a `uint32` or `uint64` though I expect that might be quite a distruptive change!

fs/config/obscure/obscure.go
```diff
@@ -47,6 +48,9 @@ func crypt(out, in, iv []byte) error {
 // This is done by encrypting with AES-CTR
 func Obscure(x string) (string, error) {
 	plaintext := []byte(x)
+	if math.MaxInt32-aes.BlockSize < len(plaintext) {
```
ncw, line 51: I think if someone has put 4 GiB of text in a string we are obscuring then perhaps they deserve rclone to blow up! The fix is OK though.

ncw (commented):
I put some queries inline about these. Some of them are quite tricky!

ncw:
> The last `Critical` issue I'm not entirely sure is real, and if so how to best fix it. Any suggestions?
> 
> * [Uncontrolled data used in network request](/rclone/rclone/security/code-scanning/88)

The p.url comes from something received from the plex server. I remember having a discussion about this with someone. 

https://github.com/rclone/rclone/blob/705c72d2939a875978d77badad2b56f7e6eb6ca0/backend/cache/plex.go#L159-L164

That is fairly common in REST protocols though so it might just be complaining about the `fmt.Sprintf("%s%s"` which should really be some sort of `url.Join`.

> I also left one issue "Incorrect conversion between integer types" (severity `High`), which I actually think is false positive due to the fact that the code checks `abs > size` etc, but that CodeQL misses that fact because several 32-bit `int`s are converted to `int64` and back.. If you agree we can select "Dismiss alert" with reason "False positive" in the GitHub Security page:
> 
> * [Incorrect conversion between integer types](/rclone/rclone/security/code-scanning/134)

I think that is a false positive, yes and it should be dismissed. I'll leave the pressing the button to you.

backend/local/metadata.go
albertony: Right, I was too quick on this one, and thinking C++ types.

I can use `strconv.ParseInt` as well, supplying bit size `0` does the same thing - it defaults to `strconv.IntSize` and effectively returns an `int` value within the `int64` return type. It is documented that "Atoi is equivalent to ParseInt(s, 10, 0), converted to type int", however it is faster, it has  a "fast path" and then a slow path where it do call `ParseInt` internally. In this specific case, though, our function `parseMetadataInt` takes `base` as a parameter, so sending that directly into `parseInt` fits best.

lib/encoder/encoder.go
albertony: `strconv.ParseUint(part, 0, 0)` does the same.

albertony:
Amended according to your comments.

ncw (approved):
Looks great, thanks

Please merge:-)

albertony:
Thanks. I have verified that the results of the latest scan had the expected improvements - except one new issue popped out, because it was previously masked by one of the other issues that I fixed in this PR. Added a new PR to fix this as well, to get it out of the way: https://github.com/rclone/rclone/pull/8021

PR #8006, opened 2024-08-15
Add automatic alias redirects for command pages

albertony:
#### What is the purpose of this change?

This continues from PR https://github.com/rclone/rclone/pull/7606, which was fixing issue https://github.com/rclone/rclone/issues/7605. 

As discussed in the issue comments, that previous PR attempt missed the fact that the gendocs command must generate the alias front matter field. This is implemented here. With this in place, Hugo will automatically generate redirect pages, which immediately redirects to the primary page using the `http-equiv="refresh" html trick`, as explained in [How Hugo Aliases Work](https://bwaycer.github.io/hugo_tutorial.hugo/extras/aliases/).

The implementation got a bit dirty, to handle aliasing at any level of subcommand paths. Currently the `completion` command has alias `genautocomplete`, intended for `/commands/rclone_genautocomplete/` to be redirected to `/commands/rclone_completion/`, however the subcommand `bash` has no alias but we still want `/commands/rclone_genautocomplete_bash/` to be redirected to `/commands/rclone_completion_bash/`.

#### Checklist

- [ ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ ] I have added tests for all changes in this PR if appropriate.
- [ ] I have added documentation for the changes if appropriate.
- [X] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [X] I'm done, this Pull Request is ready for review :-)

Issue #7605, opened 2024-01-31
Documentation for subcommand genautocomplete not in line with completion


albertony:
The documentation of the [completion](https://rclone.org/commands/rclone_completion/) subcommand lists PowerShell, while the [genautocomplete](https://rclone.org/commands/rclone_genautocomplete/) does not. However `genautocomplete` is [now](https://github.com/rclone/rclone/commit/51a468b2bae4ca8e21760435211623a8199a9167) simply an alias for `complete`, and therefore both supports powershell and lists it in help command output.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
When I try `rclone gendocs /tmp/content` to make the docs it only makes the `rclone_completion.md` file, so I think the `rclone_genautocomplete.md` is stale.

Perhaps we should delete the content and put a [redirect like the tardigrade one](https://github.com/rclone/rclone/blob/master/docs/content/tardigrade.md) or maybe put [an alias](https://bwaycer.github.io/hugo_tutorial.hugo/extras/aliases/) in the completion file which would be quite easy.

albertony:
Right. Sounds like we should just delete the docs/content/commands/rclone_genautocomplete*.md then.

Regarding redirect/alias, is the reason simply to avoid dangling links somewhere? I was thinking that a possible advantage of a redirect page like used for tardigrade is that it makes more sense to keep listing it e.g. in "Storage Systems" menu, however it seems that is not done for tardigrade and then isn't the alias solution simpler (better)?

I added a suggestion here: https://github.com/rclone/rclone/pull/7606 (I found a few references to genautocomplete here and there in the docs which I replaced with completion as well).

albertony:
Hm, doesn't look like the aliasing worked: Shouldn't we expect https://tip.rclone.org/commands/rclone_genautocomplete/ to automatically redirect to https://tip.rclone.org/commands/rclone_completion/?

ncw:
I don't think we or cobra implements that.

We could change the config doc generator to generate something like this for aliases

```
---
title: "Tardigrade"
description: "Rclone docs for Tardigrade"
versionIntroduced: "v1.52"
---

[...]

PR #8014, opened 2024-08-16
Make tardigrade page auto redirect to storj page

albertony:

Relying on Hugo, interpreting the alias and autogenerate a page which automatically redirects (https://bwaycer.github.io/hugo_tutorial.hugo/extras/aliases/).

This is in line with https://github.com/rclone/rclone/pull/8006, which handled aliases for command pages.

It was suggested in https://github.com/rclone/rclone/issues/7605#issuecomment-2291029301.

#### Checklist


albertony:
Hopefully this time I did it correct on first attempt - editing the front matter fields directly in the markdown source file, and that there is not autogenerated magic like with the command docs that overwrites it... 😅 

PS: I wrote the alias value as a simple name:

```yaml
aliases:
  - tardigrade
```

However this is handled entirely by Hugo and assumed to be a url path, `tardigrade` is interpreted as `/tardigrade/`, but luckily the backend docs are on the site root, so the resulting url `https://rclone.org/tardigrade/` is correct. I liked the appearance of the alias value this way, seemingly only naming the alias backend without any assumptions on html served docs, similar to how the command aliases are defined in source:

```go 
Aliases: []string{"genautocomplete"},
```

Although those are processed by gendocs and made into paths when generating markdown:

```yaml
aliases:
  - /commands/rclone_genautocomplete/
```

Just my preference, but if you think its risk of confusion I can change it to make it (more) clear it is a path on the rclone web site:

```yaml
aliases:
  - /tardigrade/
```

ncw (approved):
I think this looks fine. It's a relative redirect which it perfectly valid.

Please merge :-)

PR #8022, opened 2024-08-17
local: fix incorrect conversion between integer types

albertony:

Third attempt at fixing this stubborn one... This time I pushed the PR branch to upstream, because then CodeQL will run on the PR, and I was able to verify that it actually fixed the issue before merging to master. The CodeQL default setup is to run on two events: "Push to master" and "Pull request to master", and apparently the latter one does not apply for pull request from a forked repo.

#### Checklist

### DIFF
--- a/backend/local/metadata.go
+++ b/backend/local/metadata.go
@@ -129,10 +129,13 @@ func (o *Object) writeMetadataToFile(m fs.Metadata) (outErr error) {
 	}
 	mode, hasMode := o.parseMetadataInt(m, "mode", 8)
 	if hasMode {
-		if mode >= 0 && uint(mode) <= math.MaxUint32 {
-			err = os.Chmod(o.path, os.FileMode(mode))
-			if err != nil {
-				outErr = fmt.Errorf("failed to change permissions: %w", err)
+		if mode >= 0 {
+			umode := uint(mode)
+			if umode <= math.MaxUint32 {
+				err = os.Chmod(o.path, os.FileMode(umode))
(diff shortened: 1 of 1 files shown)


backend/local/metadata.go
```diff
@@ -129,10 +129,13 @@ func (o *Object) writeMetadataToFile(m fs.Metadata) (outErr error) {
...
 	mode, hasMode := o.parseMetadataInt(m, "mode", 8)
 	if hasMode {
-		if mode >= 0 && uint(mode) <= math.MaxUint32 {
-			err = os.Chmod(o.path, os.FileMode(mode))
-			if err != nil {
-				outErr = fmt.Errorf("failed to change permissions: %w", err)
+		if mode >= 0 {
+			umode := uint(mode)
+			if umode <= math.MaxUint32 {
+				err = os.Chmod(o.path, os.FileMode(umode))
```
ncw, line 135: It's complaining about the o.path not the FileMode here. That data is supplied by the user, but the Paths in the show Paths are nuts!

So I think this new alert is a false positive

ncw (commented):
I think your change looks fine. See inline for why I think the new alert is false

backend/local/metadata.go
albertony: Yeah, it is not even new, the same o.path was reported previously as well:
- https://github.com/rclone/rclone/security/code-scanning/136

And there are currently 33 reported "Uncontrolled data used in path expression"...

Issue #594, opened 2016-07-20
copy large objects as-is when syncing swift-to-swift


treyd:
When syncing a container between two swift endpoints using rclone, if there is a large object in one of these containers (SLO or DLO), the whole object is downloaded and then uploaded as a single object into the destination cluster (segmented again if necessary).  This is problematic for two reasons: one, there are valid use cases where a Swift user will create a large object out of segments of arbitrary sizes, and the reassembly in the destination will break this segmentation scheme; two, if the segments are in the same container as the manifest object, the reassembled object plus the segments will be migrated (thus doubling the data in the destination).

It would be more desirable if rclone, when syncing between two swift containers, would recognize a large object and migrate the individual segments from the source to the destination and then re-create the manifest in the destination.

ncw:
This would be difficult to achieve with the architecture of rclone at the moment.  The two remotes (source swift and destination swift) know nothing about each other and have no way to communicate something like chunk sizes.

You could sync the _segments container and remake the manifest I suppose.

Why is this causing you a problem?

treyd:
For reference: http://docs.openstack.org/developer/swift/overview_large_objects.html

In Swift, there are two ways to create a large object.  One (static large object) is to use a manifest that is a JSON file that contains a list of paths to individual chunks, along with the size, byte range in the larger object and MD5 of the chunks.  The second (dynamic large object) doesn't use a strict manifest and instead has a 0-byte object with a metatag that specifies a path prefix for all the chunks, and relies on lexicographical ordering to determine the chunk ordering for the large object.

In both of these cases, the actual chunks can be of any size (not always the max object size according to doing a GET on cluster.example.com/info and looking at `swift.max_file_size`).  In addition, the paths to the chunks include the container, so the chunks can be anywhere - in the same container, even.  Not all clients use the same segmenting scheme (e.g. not everyone puts chunks in a separate *_segments container, some clients put the chunks in the same container).

The issue (for us) in how rclone currently handles large objects is when we're trying to copy data from one swift cluster to another where the user in the source cluster has used an application (or has written one) will segment uploaded files with different sized chunks and possibly put them in the same container.  The application (and the user) expects this large object to still be a large object in the destination cluster looking the same way they uploaded it (i.e. the chunks are all there and are all the same size and in the same place).  Also, if we are migrating a container that has large objects and the chunks are stored in the same container, the current implementation migrates both the object (which it reassembles) *and* the chunks, which doubles the amount of data that gets copied.

[...]

PR #8040, opened 2024-08-29
sftp: add symlink support - fixes #5011

### LINKED ISSUE #5011: SFTP: download dangling symlink

I'd like to use `rclone copy` to download some data from an sftp server (tested with https://hub.docker.com/r/atmoz/sftp), but I get errors when rclone tries to download a symlink pointing to a non existing file.
In my use case, the symlink is dangling on the remote sftp server, but it is expected to resolve correctly once the symlink is downloaded.

Do you see any way to download these all symlinks as symlinks, even if they are broken? I've looked at the code, but can't find the piece of code handling this download process.

1.54.0

Linux, alpine docker image 3.13

####  Which cloud storage system are you using? (e.g. Google Drive)

sftp

`rclone copy -vv --ignore-existing --config '/tmp/tmpv0vrn5xp' 'sftp:test-repo/subdir' '/repos/test_repo/subdir' --sftp-user 'foo' --sftp-pass 'xxxxxxxxxxx'`

```
# rclone copy -vv --ignore-existing --config '/tmp/tmpv0vrn5xp' 'sftp:test-repo/subdir' '/repos/test_repo/subdir' --sftp-user 'foo' --sftp-pass 'xxxxxxxxxxx'
2021/02/10 10:27:56 DEBUG : rclone: Version "v1.54.0" starting with parameters ["rclone" "copy" "-vv" "--ignore-existing" "--config" "/tmp/tmpv0vrn5xp" "sftp:test-repo/subdir" "/repos/test_repo/subdir" "--sftp-user" "foo" "--sftp-pass" "VD5gGrcFh5hpuOThBM-pl77-bmk"]
2021/02/10 10:27:56 DEBUG : Using config file from "/tmp/tmpv0vrn5xp"
2021/02/10 10:27:56 DEBUG : Creating backend with remote "sftp:test-repo/subdir"
2021/02/10 10:27:56 DEBUG : sftp://foo@sftp:22/test-repo/subdir: New connection 172.22.0.4:49172->172.22.0.8:22 to "SSH-2.0-OpenSSH_7.9p1 Debian-10+deb10u2"
2021/02/10 10:27:56 DEBUG : sftp://foo@sftp:22/test-repo/subdir: Using absolute root directory "/test-repo/subdir"
2021/02/10 10:27:56 DEBUG : Creating backend with remote "/repos/test_repo/subdir"
2021/02/10 10:27:56 DEBUG : subfile.txt: Destination exists, skipping
2021/02/10 10:27:56 DEBUG : subsubdir2/poutrelle.xml: Destination exists, skipping
2021/02/10 10:27:56 DEBUG : subsubdir2/subsubfile.txt: Destination exists, skipping
2021/02/10 10:27:56 DEBUG : subsubdir2/subsubsubdir/subsubsubdir2/a file: Destination exists, skipping
2021/02/10 10:27:56 DEBUG : sftp://foo@sftp:22/test-repo/subdir: New connection 172.22.0.4:49176->172.22.0.8:22 to "SSH-2.0-OpenSSH_7.9p1 Debian-10+deb10u2"
2021/02/10 10:27:57 DEBUG : subsubdir/poutrelle.tsv: Destination exists, skipping
2021/02/10 10:27:57 DEBUG : subsubdir/subsubfile.txt: Destination exists, skipping
2021/02/10 10:27:57 DEBUG : subsubdir/poutrelle.xml: Destination exists, skipping
2021/02/10 10:27:57 DEBUG : subsubdir/relative_symlink.tsv: Copied (new)
2021/02/10 10:27:57 DEBUG : Local file system at /repos/test_repo/subdir: Waiting for checks to finish
2021/02/10 10:27:57 DEBUG : Local file system at /repos/test_repo/subdir: Waiting for transfers to finish
2021/02/10 10:27:57 ERROR : subsubdir/absolute_symlink.tsv: Failed to copy: failed to open source object: Open failed: file does not exist
2021/02/10 10:27:57 ERROR : Attempt 1/3 failed with 1 errors and: failed to open source object: Open failed: file does not exist
[...2 other similar attempts...]
2021/02/10 10:27:57 INFO  : 
Transferred:   	         0 / 0 Bytes, -, 0 Bytes/s, ETA -
Errors:                 1 (retrying may help)
Checks:                24 / 24, 100%
Elapsed time:         0.6s

2021/02/10 10:27:57 DEBUG : 20 go routines active
2021/02/10 10:27:57 Failed to copy: failed to open source object: Open failed: file does not exist
```

NikitaCOEUR:

The primary goal of this change is to enhance rclone's SFTP backend by adding support for preserving symlink structures during file transfers. This is particularly useful for users who need to maintain the integrity of symbolic links when syncing files between systems.

Add a new flag, `--sftp-links`, to the SFTP backend of rclone. The `--sftp-links` flag allows users to copy symbolic links (symlinks) as symlinks instead of following them.
The added functionality preserves the symlink structure on the destination, provided both the source and destination remotes support symlinks
> [!Important]
> Currently supported only between two SFTP remotes

The validation process ensures that symlinks are copied correctly by checking the size and hash of the target file, except when the target is a directory.

This issue was opened in 2021 and still requested : https://github.com/rclone/rclone/issues/5011

#### Checklist

> [!TIP]
> Need Help for add unit tests, i never do that before ...
> - [ ] I have added tests for all changes in this PR if appropriate.

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -178,13 +178,13 @@ E.g. if shared folders can be found in directories representing volumes:
 E.g. if home directory can be found in a shared folder called "home":
 
     rclone sync /home/local/directory remote:/home/directory --sftp-path-override /volume1/homes/USER/directory
-	
+
 To specify only the path to the SFTP remote's root, and allow rclone to add any relative subpaths automatically (including unwrapping/decrypting remotes as necessary), add the '@' character to the beginning of the path.
 
 E.g. the first example above could be rewritten as:
 
 	rclone sync /home/local/directory remote:/directory --sftp-path-override @/volume2
-	
+
 Note that when using this method with Synology "home" folders, the full "/homes/USER" path should be specified instead of "/home".
 
 E.g. the second example above should be rewritten as:
@@ -232,6 +232,19 @@ E.g. the second example above should be rewritten as:
 			Help:     "Set to skip any symlinks and any other non regular files.",
 			Advanced: true,
 		}, {
+			Name:    "links",
+			Default: false,
+			Help: `Copy symlinks instead of following them.
+
+This permit to recreate the same symlink structure on the destination.
+
+Only works between two remotes with symlink support. [Currently only supported between two SFTP remotes].
+
+Symlink is validate if target file on destination and on source have same size and same hash. [Except if target is a directory, size and hash are not checked].
+`,
+			Advanced: true,
+		}, {
+
 			Name:     "subsystem",
 			Default:  "sftp",
 			Help:     "Specifies the SSH2 subsystem on the remote host.",
@@ -243,7 +256,7 @@ E.g. the second example above should be rewritten as:
 
 The subsystem option is ignored when server_command is defined.
 
-If adding server_command to the configuration file please note that 
+If adding server_command to the configuration file please note that
 it should not be enclosed in quotes, since that will make rclone fail.
 
 A working example is:
@@ -469,7 +482,7 @@ connection for every hash it calculates.
 			Name:    "socks_proxy",
 			Default: "",
 			Help: `Socks 5 proxy host.
-	
+
 Supports the format user:pass@host:port, user@host:port, host:port.
 
 Example:
@@ -523,6 +536,7 @@ type Options struct {
 	Md5sumCommand           string          `config:"md5sum_command"`
 	Sha1sumCommand          string          `config:"sha1sum_command"`
 	SkipLinks               bool            `config:"skip_links"`
+	TranslateSymlinks       bool            `config:"links"`
 	Subsystem               string          `config:"subsystem"`
 	ServerCommand           string          `config:"server_command"`
 	UseFstat                bool            `config:"use_fstat"`
@@ -568,13 +582,15 @@ type Fs struct {
 
 // Object is a remote SFTP file that has been stat'd (so it exists, but is not necessarily open for reading)
 type Object struct {
-	fs      *Fs
-	remote  string
-	size    int64       // size of the object
-	modTime uint32      // modification time of the object as unix time
-	mode    os.FileMode // mode bits from the file
-	md5sum  *string     // Cached MD5 checksum
-	sha1sum *string     // Cached SHA1 checksum
+	fs              *Fs
+	remote          string
+	size            int64       // size of the object
+	modTime         uint32      // modification time of the object as unix time
+	mode            os.FileMode // mode bits from the file
+	md5sum          *string     // Cached MD5 checksum
+	sha1sum         *string     // Cached SHA1 checksum
+	linkTarget      string      // If object isSymlink, this is the target
+	linkTargetIsDir bool        // If object isSymlink, this is true if the target is a directory
 }
 
 // conn encapsulates an ssh client and corresponding sftp client
@@ -852,6 +868,9 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if len(opt.SSH) != 0 && ((opt.User != currentUser && opt.User != "") || opt.Host != "" || (opt.Port != "22" && opt.Port != "")) {
 		fs.Logf(name, "--sftp-ssh is in use - ignoring user/host/port from config - set in the parameters to --sftp-ssh (remove them from the config to silence this warning)")
 	}
+	if opt.TranslateSymlinks && opt.SkipLinks {
+		return nil, errors.New("can't use --sftp-links and --sftp-skip-links together")
+	}
 	f.tokens = pacer.NewTokenDispenser(opt.Connections)
 
 	if opt.User == "" {
@@ -1115,6 +1134,12 @@ func NewFsWithConnection(ctx context.Context, f *Fs, name string, root string, m
 		// Disable server side copy unless --sftp-copy-is-hardlink is set
 		f.features.Copy = nil
 	}
+	if opt.TranslateSymlinks {
+		// Enable symlink translation Feature when --sftp-links is set
+		// Not used yet but may be used in the future on shared backend operations.
+		// Maybe to check if src and dst backend support this feature before proceeding with the operation.
+		f.features.TranslateSymlink = true
+	}
 	// Make a connection and pool it to return errors early
 	c, err := f.getSftpConnection(ctx)
 	if err != nil {
@@ -1331,6 +1356,18 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 				remote: remote,
 			}
 			o.setMetadata(info)
+			if o.IsSymlink() {
+				// Read the link target if it is a symlink to get path to real object and its size
+				linkTarget, linkTargetIsDir, sizeTarget, err := f.Readlink(ctx, o.path())
+				if err != nil {
+					// If we can't read the link target, log the error and continue
+					fs.Errorf(remote, "Readlink failed: %v", err)
+					continue
+				}
+				o.size = sizeTarget
+				o.linkTarget = linkTarget
+				o.linkTargetIsDir = linkTargetIsDir
+			}
 			entries = append(entries, o)
 		}
 	}
@@ -1340,14 +1377,36 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 // Put data from <in> into a new remote sftp file object described by <src.Remote()> and <src.ModTime(ctx)>
 func (f *Fs) Put(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) (fs.Object, error) {
 	err := f.mkParentDir(ctx, src.Remote())
+
+	// Init two variables to store symlink object and check if source object is a symlink
+	// If TranslateSymlinks is set
+	var srcSymlinkObject fs.Object
+	isSymlink := false
+
 	if err != nil {
 		return nil, fmt.Errorf("Put mkParentDir failed: %w", err)
+	}
+
+	if f.opt.TranslateSymlinks {
+		// If TranslateSymlinks is set, we need to check if source object is a symlink
+		if or, ok := src.(*fs.OverrideRemote); ok {
+			srcSymlinkObject = or.UnWrap()
+			isSymlink = srcSymlinkObject.(*Object).IsSymlink()
+		}
 	}
 	// Temporary object under construction
 	o := &Object{
 		fs:     f,
 		remote: src.Remote(),
 	}
+	// if source file is a symlink, we need to specify target path to temporary object
+	if isSymlink {
+		o.linkTarget = srcSymlinkObject.(*Object).linkTarget
+		o.size = srcSymlinkObject.(*Object).size
+		o.mode = srcSymlinkObject.(*Object).mode
+		o.linkTargetIsDir = srcSymlinkObject.(*Object).linkTargetIsDir
+	}
+
 	err = o.Update(ctx, in, src, options...)
 	if err != nil {
 		return nil, err
@@ -1822,7 +1881,9 @@ func (o *Object) Remote() string {
 // Hash returns the selected checksum of the file
 // If no checksum is available it returns ""
 func (o *Object) Hash(ctx context.Context, r hash.Type) (string, error) {
-	if o.fs.opt.DisableHashCheck {
+	// Check if HashCheck is disabled or
+	// if TranslateSymlinks is enabled and the target is a directory
+	if o.fs.opt.DisableHashCheck || (o.fs.opt.TranslateSymlinks && o.linkTargetIsDir) {
 		return "", nil
 	}
 	_ = o.fs.Hashes()
@@ -1987,6 +2048,11 @@ func (o *Object) setMetadata(info os.FileInfo) {
 	o.mode = info.Mode()
 }
 
+// IsSymlink returns true if the remote sftp file is a symlink
+func (o *Object) IsSymlink() bool {
+	return o.mode&os.ModeSymlink != 0
+}
+
 // statRemote stats the file or directory at the remote given
 func (f *Fs) stat(ctx context.Context, remote string) (info os.FileInfo, err error) {
 	absPath := remote
@@ -1997,7 +2063,13 @@ func (f *Fs) stat(ctx context.Context, remote string) (info os.FileInfo, err err
 	if err != nil {
 		return nil, fmt.Errorf("stat: %w", err)
 	}
-	info, err = c.sftpClient.Stat(absPath)
+	if f.opt.TranslateSymlinks {
+		// Lstat is used to get the info of the symlink itself instead of the target
+		// We use Lstat only if the user has requested --sftp-links flag
+		info, err = c.sftpClient.Lstat(absPath)
+	} else {
+		info, err = c.sftpClient.Stat(absPath)
+	}
 	f.putSftpConnection(&c, err)
 	return info, err
 }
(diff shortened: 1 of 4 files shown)


ncw:
What do you think of changing this to work the way the local backend does with the `--link` flag?

    -l, --links                           Translate symlinks to/from regular files with a '.rclonelink' extension

If this was compatible with the local backend then we'd be able to copy symlinks from local <-> sftp as well as sftp <-> sftp which would make this a lot more generally useful

NikitaCOEUR:
Thank you for your feedback! I'd be happy to try using the `--links` flag instead of `--sftp-links`. However, I'm not entirely sure how to achieve this because I added the option directly in the SFTP backend, which automatically created the `--sftp-links` flag. Could you please clarify how I can utilize `--links` in this context?

ncw:
I didn't mean use --links, I meant make --sftp-links work in the same way, making files with the contents of the link with suffix .rclonelink

NikitaCOEUR:
Ok I'll work on it.

NikitaCOEUR:
Hello,

I wanted to provide an update on my progress with the SFTP backend and symlink implementation for this PR. I've encountered a few challenges and would appreciate your feedback on the following points:

## Comparison of Symlink Sizes [Use target size instead of Symlink Size]
Instead of comparing the size of the symlink itself at the source and destination, I prefer comparing the size of the file pointed to by the symlink. This is because some SFTP servers correct symlinks with relative paths to absolute paths (e.g., `../directory/file` becomes `/full/path/directory/file`).
When verification occurs, the size comparison between the source and destination may differ since the symlink with the absolute path is larger than the one with the relative path. Rclone currently flags this as an issue.
Does this approach seem reasonable, or do you have other suggestions?

## Handling Symlinks Pointing to Directories [Ensure target is a directory on source and destination]
The same issue arises with symlinks pointing to directories. It is challenging to verify if the target directories are the same, as comparing sizes or hashing directories is impractical.
My current thought is to simplify the verification by ensuring that both source and destination symlinks point to directories, without further verification of their contents. What are your thoughts on this approach?

## Timestamp Management for Symlinks [Ignore Timestamp for symlink on sftp backend]
I've encountered some surprises regarding timestamp management for symlinks. Most SFTP servers don't seem to support modifying the access and modification times of symlinks directly.
Although Chtimes is available, it operates on the file pointed to by the symlink, not the symlink itself.
An equivalent to Lutimes would be needed but is not widely implemented (or possibly not supported at all with the SFTP protocol). 
Should we ignore timestamp modifications for symlinks, or should we consider adding a flag like SetModTime=False when TranslateSymlinks is enabled in the SFTP backend?

I’ve opened an issue on the lutimes topic in the SFTP library to gather more input: [Issue #594](https://github.com/pkg/sftp/issues/594).

----
I wanted to keep you informed of my progress. I’m working on this in my spare time, which may explain some delays in communication.

Thank you for your feedback.

puellanivis:
> An equivalent to Lutimes would be needed but is not widely implemented (or possibly not supported at all with the SFTP protocol).

I’ve responded to your request at the SFTP package itself, but I want to add here for quick reference: your suspicion here is correct. There is sadly no way within the protocol to operate on symlinks directly, except for `SSH_FXP_LSTAT` and `SSH_FXP_READLINK`.

All the workarounds I could find “escape-hatch” out of SFTP into the greater SSH protocol to execute an arbitrary command.

ncw:
Thanks for working on this @NikitaCOEUR 

> I wanted to provide an update on my progress with the SFTP backend and symlink implementation for this PR. I've encountered a few challenges and would appreciate your feedback on the following points:
> 
> ## Comparison of Symlink Sizes [Use target size instead of Symlink Size]
> Instead of comparing the size of the symlink itself at the source and destination, I prefer comparing the size of the file pointed to by the symlink. This is because some SFTP servers correct symlinks with relative paths to absolute paths (e.g., `../directory/file` becomes `/full/path/directory/file`).

That might cause a problem with continual syncing if we are syncing symlinks to the sftp server.

> When verification occurs, the size comparison between the source and destination may differ since the symlink with the absolute path is larger than the one with the relative path. Rclone currently flags this as an issue. Does this approach seem reasonable, or do you have other suggestions?

If we want to emulate the handling of the `.rclonelink` files in the local backend, then the size will need to be the size of the symlink itself as rclone will want to download the contents of the symlink rather than the file pointed to.

I guess we could if necessary mark the size of the symlinks as unknown (`-1`). The rest of rclone knows what to do in this case. This isn't ideal as rclone then won't see if the symlink size has changed but it would probably work. Maybe this should be an option if it isn't needed on all sftp servers.

> ## Handling Symlinks Pointing to Directories [Ensure target is a directory on source and destination]
> The same issue arises with symlinks pointing to directories. It is challenging to verify if the target directories are the same, as comparing sizes or hashing directories is impractical. My current thought is to simplify the verification by ensuring that both source and destination symlinks point to directories, without further verification of their contents. What are your thoughts on this approach?

I think this should just be the symlink size and the code shouldn't care if it is pointing to a file or directory.

> ## Timestamp Management for Symlinks [Ignore Timestamp for symlink on sftp backend]
> I've encountered some surprises regarding timestamp management for symlinks. Most SFTP servers don't seem to support modifying the access and modification times of symlinks directly. Although Chtimes is available, it operates on the file pointed to by the symlink, not the symlink itself. An equivalent to Lutimes would be needed but is not widely implemented (or possibly not supported at all with the SFTP protocol). Should we ignore timestamp modifications for symlinks, or should we consider adding a flag like SetModTime=False when TranslateSymlinks is enabled in the SFTP backend?

Hmm, we should certainly avoid setting the timestamp of the destination when we meant to set the timestamp of the symlink.

If we just ignore the SetModTime for the symlinks then potentially you'll have rclone trying to modify the time on each sync.

Returning `fs.ErrorCantSetModTime` in this case will probably cause rclone to do something sensible!

> I’ve opened an issue on the lutimes topic in the SFTP library to gather more input: [Issue #594](https://github.com/pkg/sftp/issues/594).
> 
> I wanted to keep you informed of my progress. I’m working on this in my spare time, which may explain some delays in communication.

No worries, work at whatever pace you need. I work in an entirely interrupt driven way so if you write something on the PR I'll look at it, be that in 1 day or week or month!

NikitaCOEUR:
Hi,

>Returning fs.ErrorCantSetModTime in this case will probably cause rclone to do something sensible!

This causes the recreation of each symlink every time. When we create a symlink, the ModTime is set to the same as the Created Time. When rclone is launched for the first time and the symlink was created by a preceding rclone run, the symlink on the destination and source appear similar, but the ModTime differs. Raising this error triggers the systematic recreation of the symlink. I believe we need to find an alternative because systematically recreating the symlink doesn't make sense to me.

I think we should completely ignore the ModTime check for symlink objects. But I’m not sure how to go about it yet.

Neustradamus:
Any progress on it?

pepa65:
Please just implement it by recreating the symlinks first. People choosing to run with this flag would foremost want the symlinks, regardless of whatever they might have pointed at in the source filesystem, and regardless of potential inefficiencies.

I think recreating all symlinks is not a huge cost, and would greatly increase the value of rclone as a whole to me.

NikitaCOEUR:
Hi everyone,

I took some time to revisit this and tried to implement the approach that ncw suggested for the local backend. I can confirm that it works, but I’ve run into some important limitations along the way.

That said, the scenario that I think will interest most people—synchronizing a local backend with an OpenSSH SFTP server, or synchronizing between two OpenSSH SFTP backends while keeping the same directory structure—should work. This is probably the use case that matters the most in practice.

It’s important to note that the SFTP protocol itself does not allow this feature to be implemented perfectly. There are inherent constraints (like path normalization, chroot environments, and lack of lutimes for symlinks) that can cause edge cases to fail.

I also want to apologize for the time I spent on this. The problems I encountered initially were quite demotivating, and I had even worked around them by adding a script to replicate symlinks between the servers I wanted to sync.

However, the recent comments made it clear that this feature isn’t only desired by me, so I hope this implementation can serve the purpose and be useful for others as well.

Thanks for taking the time to review this!

pepa65:
I checked out your repo and build the `feat/add-sftp-links` branch, and it accepts the flag `--sftp-links`, but for each symlink it reports: "Can't follow symlink without -L/--copy-links". So I tried both `--sftp-links` and `-l`, which gives no errors, but does seems to copy the whole file instead of the symlink. Then I tried `--sftp-links` and `-L`, also no errors, but also no symlinks. What am I doing wrong?

NikitaCOEUR:
What exactly was the command you executed?
From local backend to sftp backend ? 
Did you use the flag --sftp-links or directly --links ?

pepa65:
Started with: `rclone copy --sftp-links /var/www pp:/data/www` -- should I be using `--links`??

NikitaCOEUR:
`--sftp-links` enables `TranslateSymlinks` only for the SFTP backend, just like `--local-links` does for the local backend.  
To enable `TranslateSymlinks` for all backends that support it, you need to specify each corresponding flag (`--local-links` + `--sftp-links`) or simply use `--links`, which is the global flag that enables `--<backend>-links`.

I think that’s the problem you encountered, because source symlinks are not transmitted as `.rclonelink` files to the SFTP backend.

You should verify the symlinks on the destination side, as I noticed you're transferring them from /var/www to /data/www. Some SFTP servers may rewrite or normalize symlinks depending on their implementation. I ran into a few unexpected behaviors during my tests.

pepa65:
Hmmm. Local and remote are both ext4 filesystems, and the remote SFTP-server is `openssh-sftp-server 9.6p1`. What I would like to happen is that the local symlinks will translate into remote symlinks. How can I make that happen, just use `--links`?

NikitaCOEUR:
I think you can just use `--links` yes

pepa65:
Okay, it seems that copying to an existing remote copy doesn't overwrite existing files... When writing to a new remote directory, it works!

But this does have me concerned, that old/obsolete content does not get overwritten. What if I in the future replace a file with a symlink, will the remote keep having the file instead of the symlink?

ncw:
@NikitaCOEUR wrote:
> I also want to apologize for the time I spent on this. The problems I encountered initially were quite demotivating, and I had even worked around them by adding a script to replicate symlinks between the servers I wanted to sync.
> 
> However, the recent comments made it clear that this feature isn’t only desired by me, so I hope this implementation can serve the purpose and be useful for others as well.

Thanks for persevering with this, and I'm sorry for the demotivation.

I will give this a review now - thank you :-)

backend/sftp/sftp.go
```diff
@@ -910,6 +948,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	if len(opt.SSH) != 0 && ((opt.User != currentUser && opt.User != "") || opt.Host != "" || (opt.Port != "22" && opt.Port != "")) {
 		fs.Logf(name, "--sftp-ssh is in use - ignoring user/host/port from config - set in the parameters to --sftp-ssh (remove them from the config to silence this warning)")
 	}
+	// Override --sftp-links with --links if set
```
ncw: This needs to be called out in the docs like the relevant section in the local backend

> Note that `--sftp-links` just enables this feature for the sftp backend. `--links` and `-l` enable the feature for all supported backends (including `local`) and the VFS.

ncw (commented):
This is looking very good - thank you :-)

I put one minor comment inline and there are some lint errors for you to fix up too.

And some questions:

Do the integration tests still pass? If you run `test_all -backends sftp`? You'll need docker for that to work. (See the docs https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#backend-testing )

I guess `rclone serve sftp` doesn't support symlinks either so I think the integration test will fail there.

Great work :-)

NikitaCOEUR:
Indeed, after quickly re-testing to reproduce the issue reported by pepa65, I managed to trigger the same error using `rclone serve`.

I’m currently working on a detector that evaluates whether the SFTP backend supports symlinks, so that an error can be raised earlier if it doesn’t. The goal is to warn users when the backend cannot handle symlinks and that `--links` cannot be used with it.

I also noticed that when trying to replace a regular file with a symlink, the symlink is created, but it gets removed during the cleanup phase (because of the `.rclonelink` handling — this happens even between two local backends). Running the sync a second time creates the symlink correctly.
>Okay, it seems that copying to an existing remote copy doesn't overwrite existing files... When writing to a new remote directory, it works!
@pepa65  What exactly is your command? Overwrite “works” for me.

Finally, I saw that symlinks are always flagged for recreation because their ModTime cannot be compared (due to `lutime` lacks), and hash handling isn’t natively supported with the SFTP backend.
I’m considering “forcing” hash computation — not on the SFTP server side, but on the rclone client side — since the hash can be derived from the string representing the target path. This would only apply to TranslatedLink objects.

Should I add a new commit to this PR, or would it be better to handle this in a separate one?
Also, regarding your earlier comment — should that be a new commit or a force-push update too?

Note: I’m currently trying to run `test_all --backends sftp`, but I’m seeing a lot of
`NOTICE: Failed to open remote "TestSFTPRclone:": NewFs: couldn't connect SSH: dial tcp 127.0.0.1:28621: connect: connection refused`
and occasionally messages like `Finished OK in 504.613562ms (try 2/5).`

I’m not sure if this is expected behavior or if there’s an actual issue.

ncw:
A detector sounds like a good idea. You can check for that error in the integration tests and skip them.

> I also noticed that when trying to replace a regular file with a symlink, the symlink is created, but it gets removed during the cleanup phase (because of the `.rclonelink` handling — this happens even between two local backends). Running the sync a second time creates the symlink correctly.

Effectively there are two files with the same name which confuses the sync.

> Finally, I saw that symlinks are always flagged for recreation because their ModTime cannot be compared (due to lutime lacks), and hash handling isn’t natively supported with the SFTP backend.

We need to fix that. Just calculating the hash should do it though shouldn't it?

> Should I add a new commit to this PR, or would it be better to handle this in a separate one?

Yes add to this commit. Let's get everything we need for the feature in this PR.

> Also, regarding your earlier comment — should that be a new commit or a force-push update too?

Don't mind - however you like to work. It will all get squashed into one commit when it gets merged.

> Note: I’m currently trying to run test_all --backends sftp, but I’m seeing a lot of
> NOTICE: Failed to open remote "TestSFTPRclone:": NewFs: couldn't connect SSH: dial tcp 127.0.0.1:28621: connect: connection refused
> and occasionally messages like Finished OK in 504.613562ms (try 2/5).
> 
> I’m not sure if this is expected behavior or if there’s an actual issue.

Sounds like it is working!

It takes a while then opens an HTML page with the results.

The can't connect might just be startup things or maybe you have a local firewall?

`test_all` is usually run on Linux so gets the most testing there. It works on macOS I believe but is unlikely to work on Windows.

NikitaCOEUR:
Hi @ncw,

  I've implemented all the changes:

  Changes Made

  1. ✅ Lint errors - Fixed (0 issues)
  2. ✅ Documentation - Added the note about --sftp-links vs --links in the option help text (backend/sftp/sftp.go:285-287)
  3. ✅ Symlink capability detection - Added SymlinkValidator interface that:
    - Tests if the backend can actually create symlinks before sync operations
    - Provides clear early error messages when symlinks are enabled but not supported
    - Implemented for both local (Windows privilege check) and SFTP backends
  4. ✅ Hash-based comparison for symlinks - Modified CheckHashes() in fs/operations to:
    - Compute hashes for .rclonelink files even when filesystems don't declare common hash support
    - Prevents systematic recreation of symlinks on each sync
  5. ✅ SetModTime handling for symlinks - Added SetModTimeCapability interface:
    - SFTP backend declares when SetModTime is not supported (e.g., for symlinks due to lack of lutimes)
    - Equality check skips modtime verification when not supported
    - Prevents false "not equal" detections

  Integration Tests

  I'm having difficulties getting consistent results with `test_all -backends sftp`. 
  - A lot of connection errors ("connection refused")
  - These issues appear even on master branch, making it hard to validate if my changes introduce problems
  
I’m working on Windows through a WSL environment, and I’ve never encountered issues like the ones I’m seeing here. The tests seem to start a container or a service, but those are stopped or restarted between tests — though that doesn’t always seem to happen. When they aren’t running, the tests consistently return a “connection refused” error.
I tried to fix it but couldn’t find a solution. Could you check on your side whether everything looks fine, or if some integration tests might need to be adjusted?
Thanks.

pepa65:
My job with symlinks went fine.

ncw:
@NikitaCOEUR 

Thanks for the changes - will check those over in a bit

> I’m working on Windows through a WSL environment, and I’ve never encountered issues like the ones I’m seeing here. The tests seem to start a container or a service, but those are stopped or restarted between tests — though that doesn’t always seem to happen. When they aren’t running, the tests consistently return a “connection refused” error.
> I tried to fix it but couldn’t find a solution. Could you check on your side whether everything looks fine, or if some integration tests might need to be adjusted?

The integration tests work fine on Linux - you can see them running here https://integration.rclone.org

I suspect this is a problem with WSL, perhaps with docker images on WSL. I don't have time to fix the tests on WSL though. Maybe @roucc can test this for you?

PR #8041, opened 2024-08-30
zoho: use download server and large file upload API

### LINKED ISSUE #5995: Zoho WorkDrive backend can not upload large files (>250MB)
#### Output of `rclone version`
rclone v1.58.0-beta.5999.f22b703a5
- os/kernel: 5.4.103-1-pve (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.17.6
- go/linking: static
- go/tags: none

#### Describe the issue

Zoho WorkDrive backend can not upload large files (>250MB)
I believe this may be caused by a limit Zoho has put on their API, the 'upload' endpoint does not support files larger than 250MB, and instead another endpoint needs to be used 'uploadlargefile'

Reference:
https://workdrive.zoho.com/apidocs/v1/filesfolders/uploadfile
https://workdrive.zoho.com/apidocs/v1/filesfolders/uploadlargefile

buengese:

This started out as a continuation of #7007 and supersedes it but I decided to go a bit further. This also implements the newer "large file" upload API which seems to perform much better.

I've also decided to rework how upload works in general. Files are now uploaded with a temporary name and renamed to their final name after the upload completes. This should get rid of all the encoding related integration test failures since Zoho actually supports most special characters - just not in the upload api.
There is however a big downside to this, the upload + delete + rename method does break file version history.

Also happens to fix #5995

Nope

#### Checklist

### DIFF
--- a/backend/zoho/zoho.go
+++ b/backend/zoho/zoho.go
@@ -50,6 +50,7 @@ var (
 			"WorkDrive.team.READ",
 			"WorkDrive.workspace.READ",
 			"WorkDrive.files.ALL",
+			"ZohoFiles.files.ALL",
 		},
 		Endpoint: oauth2.Endpoint{
 			AuthURL:   "https://accounts.zoho.eu/oauth/v2/auth",
@@ -61,6 +62,7 @@ var (
 		RedirectURL:  oauthutil.RedirectLocalhostURL,
 	}
 	rootURL     = "https://workdrive.zoho.eu/api/v1"
+	downloadURL = "https://download.zoho.eu/v1/workdrive"
 	accountsURL = "https://accounts.zoho.eu"
 )
 
@@ -205,6 +207,7 @@ type Fs struct {
 	opt      Options            // parsed options
 	features *fs.Features       // optional features
 	srv      *rest.Client       // the connection to the server
+	downloadsrv *rest.Client    // the connection to the Download server
 	dirCache *dircache.DirCache // Map of directory path to directory id
 	pacer    *fs.Pacer          // pacer for API calls
 }
@@ -229,6 +232,7 @@ func setupRegion(m configmap.Mapper) error {
 		return errors.New("no region set")
 	}
 	rootURL = fmt.Sprintf("https://workdrive.zoho.%s/api/v1", region)
+	downloadURL = fmt.Sprintf("https://download.zoho.%s/v1/workdrive",region)
 	accountsURL = fmt.Sprintf("https://accounts.zoho.%s", region)
 	oauthConfig.Endpoint.AuthURL = fmt.Sprintf("https://accounts.zoho.%s/oauth/v2/auth", region)
 	oauthConfig.Endpoint.TokenURL = fmt.Sprintf("https://accounts.zoho.%s/oauth/v2/token", region)
@@ -405,6 +409,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		root:  root,
 		opt:   *opt,
 		srv:   rest.NewClient(oAuthClient).SetRoot(rootURL),
+		downloadsrv:   rest.NewClient(oAuthClient).SetRoot(downloadURL),
 		pacer: fs.NewPacer(ctx, pacer.NewDefault(pacer.MinSleep(minSleep), pacer.MaxSleep(maxSleep), pacer.DecayConstant(decayConstant))),
 	}
 	f.features = (&fs.Features{
@@ -1159,7 +1164,7 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 		Options: options,
 	}
 	err = o.fs.pacer.Call(func() (bool, error) {
-		resp, err = o.fs.srv.Call(ctx, &opts)
+		resp, err = o.fs.downloadsrv.Call(ctx, &opts)
 		return shouldRetry(ctx, resp, err)
 	})
 	if err != nil {
--- a/backend/zoho/zoho.go
+++ b/backend/zoho/zoho.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/pacer"
 	"github.com/rclone/rclone/lib/random"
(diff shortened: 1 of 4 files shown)


buengese:
I was thinking maybe we need an`encoder.IsSimpleName` or something like that checks if a name is only composed of alphanumeric characters (and some common special characters like '.', '-', '_'). That would to directly upload these files preserving version history where possible.

ncw:
(I haven't looked at the code yet!)

> I've also decided to rework how upload works in general. Files are now uploaded with a temporary name and renamed to their final name after the upload completes.

This is relatively new, so you may not have seen this, but this is what the `PartialUploads` feature flag does. It causes files to be uploaded with temporary names and renamed. Its for backends where the partially uploaded file is visible during upload. It may not be relevant here but I thought I'd mention it.

https://github.com/rclone/rclone/blob/af9f8ced805509c3efd264fc5be2a82ea1a000f2/fs/features.go#L38

> This should get rid of all the encoding related integration test failures since Zoho actually supports most special characters - just not in the upload api.

That is useful.

Note also the https://github.com/rclone/rclone/tree/fix-zoho-api branch which I've been working on with Zoho support [on the forum](https://forum.rclone.org/t/second-followup-on-the-older-topic-rclone-invokes-more-number-of-workdrive-s-files-listing-api-calls-which-exceeds-the-throttling-limit/45697/20) which helps enormously with rate limiting.

> There is however a big downside to this, the upload + delete + rename method does break file version history.

I know that Google Drive users care about this **a lot** - the preference of Zoho users is unknown, but I'd assume that they care.

BTW does this large file upload API work with the free zoho? Are there any file size limits? And will we need to add any scopes to the oauth?

> I was thinking maybe we need an`encoder.IsSimpleName` or something like that checks if a name is only composed of alphanumeric characters (and some common special characters like '.', '-', '_'). That would to directly upload these files preserving version history where possible.

That sounds like a good compromise. Encoded names are relatively rare.

Perhaps check if the encoded name is the same as the raw name? As surely this will vary by encoding won't it? (Eg some encodings don't like `.` on the end of the name)

buengese:
I've rebased this on your changes but my original plans don't really work out. Zoho APIs  seem to have a tendency to be very frustrating to use. It appears the large file upload API does not support overwriting existing files so inherently breaks file version.

However there is a third upload API - the chunked upload API - this one this support file versioning so maybe it's the way to forward.

buengese:
Ok I think this is done now.

A quick summary of the changes:
- The download server is now used for downloads.
- Upload Large files API is used for files larger than 10MB. The documentation suggest to use this for files larger than 250MB but in my testing it performs so much better that it makes sense to use it for almost everything. The only limitation this API seems to have is that it cannot upload files with size 0.
	-  2 new OAuth2 scopes are required to use this API. I've added an error message to notify the use that `rclone config reconnect` needs to be run.
- Encoding related Integration test failures should be fixed. 
	- The upload with temporary name, rename afterwards method was NOT needed. Instead we can just URL encode filenames. The API docs suggest this only works for URL parameters but in fact this also works for Headers and this means we can use it with the upload API.
	- Additionally, contrary to my original assumptions we can also use the`override-name-exists` header with the upload
	- I've disable the `invalid-utf8` test for this backend. The encoded version also causes problems and I don't think many eople have names with invalid UTF-8 to begin with. large file API. (This is also undocumented)
- I've attempted to improve the rate limiting situation a bit by forcing a 60s sleep on status 429. This isn't perfect and makes things slower but it gave me a complete successful (albeit slow) integration test run.
- I added support for private spaces which means the personal version of Zoho WorkDrive is now supported.

buengese:
rebased

backend/zoho/zoho.go
```diff
@@ -293,13 +293,19 @@ func shouldRetry(ctx context.Context, resp *http.Response, err error) (bool, err
...
+		fs.Errorf(nil, "zoho: missing OAuth Scope. Run rclone config reconnect to fix this issue.")
+		return false, err
+	}
+
 	if resp != nil && resp.StatusCode == 401 && len(resp.Header["Www-Authenticate"]) == 1 && strings.Contains(resp.Header["Www-Authenticate"][0], "expired_token") {
 		authRetry = true
 		fs.Debugf(nil, "Should retry: %v", err)
 	}
 	if resp != nil && resp.StatusCode == 429 {
 		fs.Errorf(nil, "zoho: rate limit error received, sleeping for 60s: %v", err)
```
ncw: This log is now wrong :-)

**Edit** I see you update this in a subsequent commit :-)

backend/zoho/zoho.go
```diff
@@ -710,21 +767,40 @@ func (f *Fs) upload(ctx context.Context, name string, parent string, size int64,
...
+			return nil, err
+		}
 
-	// Create the directory for the object if it doesn't exist
-	leaf, directoryID, err := f.dirCache.FindPath(ctx, remote, true)
-	if err != nil {
-		return nil, err
-	}
+		// use normal upload API for small sizes (<10MiB)
+		if size < 10*1024*1024 {
```
ncw: On other backends which do this we allow this to be configured with an upload cutoff

Eg s3

https://github.com/rclone/rclone/blob/2d1c2b1f76f44b4db5369bacdd4e5f76feecc128/backend/s3/s3.go#L2119-L2126

```
      --azureblob-upload-cutoff string       Cutoff for switching to chunked upload (<= 256 MiB) (deprecated)
      --b2-upload-cutoff SizeSuffix          Cutoff for switching to chunked upload (default 200Mi)
      --box-upload-cutoff SizeSuffix         Cutoff for switching to multipart upload (>= 50 MiB) (default 50Mi)
      --drive-upload-cutoff SizeSuffix       Cutoff for switching to chunked upload (default 8Mi)
      --hidrive-upload-cutoff SizeSuffix     Cutoff/Threshold for chunked uploads (default 96Mi)
      --oos-upload-cutoff SizeSuffix         Cutoff for switching to chunked upload (default 200Mi)
      --qingstor-upload-cutoff SizeSuffix    Cutoff for switching to chunked upload (default 200Mi)
      --s3-upload-cutoff SizeSuffix          Cutoff for switching to chunked upload (default 200Mi)
      --sharefile-upload-cutoff SizeSuffix   Cutoff for switching to multipart upload (default 128Mi)
```

It would probably be a good idea to add that here just in case users need to tweak it

ncw (commented):
I think this is looking great - thank you :-)

I think we should add an `upload_cutoff` to control the large file switchover point - users are just bound to want to adjust it.

Do you want to get this merged for v1.68? I'm planning to release it at the weekend hopefully.

buengese:
Sorry for the delays. I've addressed your feedback now. I think this is ready for merge now.

ncw (approved):

I'll merge now.

Keep an eye on the integration tests tomorrow: https://pub.rclone.org/integration-tests/current/

PR #8044, opened 2024-09-02
Use builtin MkdirAll with go1.22 instead of our own custom version for windows

albertony:

Starting with go1.22 the standard os.MkdirAll has improved its handling of volume names, and as part of that it now stops recursing into parent directory if it is a volume name (see: https://github.com/golang/go/commit/cd589c8a73415afbf94a8976f20cbed9d4061ba6). This is similar to what was our main change and reason for creating a custom version. When building with go1.22 or newer we can therefore stop using our custom version, with the advantage that we automatically get current and future relevant improvements from golang. To support building with go1.21 the existing custom version is still kept, and therefore also our wrapper function file.MkdirAll - but it now just calls os.MkdirAll with go1.22 or newer on Windows.

See #5401, #6420 and acf1e2df84a350b7a86d7672d749dfb1ba090a44 for details about the creation of our custom version of MkdirAll.

#### Checklist

### DIFF
--- a/lib/file/mkdir_other.go
+++ b/lib/file/mkdir_other.go
@@ -1,10 +1,10 @@
-//go:build !windows
+//go:build !windows || go1.22
 
 package file
 
 import "os"
 
-// MkdirAll just calls os.MkdirAll on non-Windows.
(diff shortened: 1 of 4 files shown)


ncw (approved):
Look good to me. Please merge :-)

Issue #8024, opened 2024-08-18
Insecure 


nipil:
I think there is a problem with the "one-command" verification
at https://rclone.org/release_signing/#verify-signatures-and-hashes-together

Problem : the one-command example might imply valid release even if the signature is wrong.

Example : even if the SHA256SUM is altered, gpg decrypts data and sha256 succesfully verifies binaries, exiting globally with 0 (OK) :

```
# tail SHA256SUMS
826463d9a2bc5e511e091c24be7d4bf6f2df396702662fb528498223ccb39b94  rclone-v1.67.0-vendor.tar.gz
458413bdd7a85cb8a19a7f955e25ac633fe1513f956b6bc09efd5ca51d44aa8a  rclone-v1.67.0-windows-386.zip
117b99441024607d6043e274c7fcbed64d07ad87347d17dd0a717bdc1c59716b  rclone-v1.67.0-windows-amd64.zip
54f263712d02bf2345eb5a3444aa4f07b990f5b4c6d02f1de892d1ff8028b50c  rclone-v1.67.0-windows-arm64.zip_
-----BEGIN PGP SIGNATURE-----

iF0EARECAB0WIQT79zfs6firGGBL0qyTk14C/ztU+gUCZmxxrwAKCRCTk14C/ztU
+kUlAJ9Z93y4PMxYSmowWOSkNvI2lEsgKgCbBoPCt1Y8lupPWzhKsfGkE/7k1aQ=
=ae4p
-----END PGP SIGNATURE-----

# gpg --no-default-keyring --keyring rclone.gpg --trust-model always --decrypt SHA256SUMS | sha256sum -c --ignore-missing
gpg: Signature made Fri Jun 14 16:37:03 2024 UTC
gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
gpg: BAD signature from "Nick Craig-Wood <nick@craig-wood.com>" [unknown]
rclone-v1.67.0-linux-amd64.deb: OK

root@4aecc0ab8110:/# echo $?
0
```

Cause : gpg exit code (1=error) during decrypt is lost during the piping :

```
# gpg --no-default-keyring --keyring rclone.gpg --trust-model always --verify SHA256SUMS
gpg: Signature made Fri Jun 14 16:37:03 2024 UTC
gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
[...]

PR #8050, opened 2024-09-04
docs: more secure two-step signature and hash validation

nipil:

Ensure that the status code return when verifying gpg signature is not swallowed when verifyin file hashes

Propose a fix for issue #8024 

Same result in the "good" case :

```
gpg: Signature made Fri Jun 14 18:37:03 2024
gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
gpg: Good signature from "Nick Craig-Wood <nick@craig-wood.com>" [unknown]
Primary key fingerprint: FBF7 37EC E9F8 AB18 604B  D2AC 9393 5E02 FF3B 54FA
rclone-v1.67.0.tar.gz: OK

$ echo $?
0
```

Same result in the "bad hash" case :

```
gpg: Signature made Fri Jun 14 18:37:03 2024
gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
gpg: Good signature from "Nick Craig-Wood <nick@craig-wood.com>" [unknown]
Primary key fingerprint: FBF7 37EC E9F8 AB18 604B  D2AC 9393 5E02 FF3B 54FA
rclone-v1.67.0.tar.gz: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
sha256sum: 'standard input': no file was verified

$ echo $?
1
```

And a new result in the "bad signature" case :

```
gpg: Signature made Fri Jun 14 18:37:03 2024
gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
gpg: BAD signature from "Nick Craig-Wood <nick@craig-wood.com>" [unknown]

$ echo $?
1
```

#### Another feedback

The page at https://downloads.rclone.org/ does not list hash filed (md5, sha1, sha256) for `*-current-*` files, while all `vX.YY.z` sub-directories have hash files. So no signature is available directly from "current files" set. Using `VERSION` as a first stop, to then go in nested directories works but might seem counter-intuitive in contrast with the "current" files

### DIFF
--- a/docs/content/release_signing.md
+++ b/docs/content/release_signing.md
@@ -149,7 +149,7 @@ $ rclone hashsum sha256 -C SHA256SUMS rclone-v1.63.1-windows-amd64.zip
 You can verify the signatures and hashes in one command line like this:
 
 ```
-$ gpg --decrypt SHA256SUMS | sha256sum -c --ignore-missing
+$ h=$(gpg --decrypt SHA256SUMS) && echo "$h" | sha256sum - -c --ignore-missing
 gpg: Signature made Mon 17 Jul 2023 15:03:17 BST
 gpg:                using DSA key FBF737ECE9F8AB18604BD2AC93935E02FF3B54FA
 gpg: Good signature from "Nick Craig-Wood <nick@craig-wood.com>" [ultimate]


ncw (approved):

The solutions I was thinking of involved `set -o pipefail` which don't really work on the command line.

PR #8051, opened 2024-09-04
cloudinary: Adding Cloudinary backend - fixes #7989

### LINKED ISSUE #7989: Cloudinary backend
rclone v1.67.0-DEV

Supporting migrations and integrations with the popular Cloudinary service

Adding a new Cloudinary backend

#### How to use GitHub

yuval-cloudinary:
Adding a new backend for [Cloudinary](https://cloudinary.com/)

https://github.com/rclone/rclone/issues/7989

#### Checklist


ncw:
Great work :-)

How many of the integration tests are passing? `test_all -backends cloudinary`?

I notice you use a cloudinary go package - can you run

    bin/test_backend_sizes.py cloudinary

To see how much size this is adding to the rclone binary?

I'll now put a few very quick comments on the code - not a full review yet, but some obvious bits

ncw:
Looks like you need to run `go mod tidy` too!

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+	}
+	if results.TotalCount == 0 || len(results.Assets) == 0 {
+		return nil, fs.ErrorObjectNotFound
+	}
+	asset := results.Assets[0]
+
+	o := &Object{
+		fs:           f,
+		remote:       remote,
+		size:         int64(asset.Bytes),
```
ncw, line 380: What is the type of `asset.Bytes`? I'm hoping it is `uint64` because if it is smaller that that then this is a bug waiting to happen!

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+		if err == nil && count == int64(cl) {
+			break
+		}
+		time.Sleep(time.Duration(i) * time.Second)
+	}
+	return resp.Body, err
+}
+
+func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, options ...fs.OpenOption) error {
+	srcImmutable := object.NewStaticObjectInfo(o.Remote(), src.ModTime(ctx), src.Size(), true, nil, o.Fs())
```
ncw, line 583: Why are  you making `srcImmutable`? what is wrong with just using `src`?

docs/content/cloudinary.md
```diff
@@ -0,0 +1,242 @@
+---
+title: "Cloudinary"
+description: "Rclone docs for Cloudinary backend"
+versionIntroduced: "v1.0"
```
ncw, line 4: This would need to be v1.69.0 (I don't think we'll make the v1.68 release which is coming up in a few days

docs/content/cloudinary.md
```diff
@@ -0,0 +1,242 @@
...
+## About Cloudinary
+
+[Cloudinary](https://cloudinary.com/) is an image and video API platform.
+Trusted by 1.5 million developers and 10,000 enterprise and hyper-growth companies as a critical part of their tech stack to deliver visually engaging experiences.
+
+## Accounts & Pricing
+
+To use this backend, you need to [create a free account](https://cloudinary.com/users/register_free) on Cloudinary. Start with a free plan with generous usage limits. Then, as your requirements grow, upgrade to a plan that best fits your needs. See [the pricing details](https://cloudinary.com/pricing).
+
+## Securing Your Credentials
```
ncw, line 20: This is good documentation, but it doesn't belong in the cloudinary backend docs. Can you remove it here, and consider making a PR to `docs/content/docs.md`?

docs/content/cloudinary.md
```diff
@@ -0,0 +1,242 @@
...
+[snip]
+XX / cloudinary.com
+\ (cloudinary)
+[snip]
+Storage> cloudinary
+
+Option cloud_name.
+You can find your cloudinary.com cloud_name in your [dashboard](https://console.cloudinary.com/pm/developer-dashboard)
+Enter a value.
+cloud_name> ****************************
```
ncw, line 103: Anything you've `***` here you might want to consider adding a `Sensitive` tag in the config. This means it will be redacted with `rclone config redacted`.

docs/content/cloudinary.md
```diff
@@ -0,0 +1,242 @@
...
+
+`rclone mkdir cloudinary-media-library:directory`
+
+List the contents of a directory.
+
+`rclone ls cloudinary-media-library:directory`
+
+### Modified time and hashes
+
+Cloudinary stores md5sum and timestamps natively
```
ncw, line 162: Judging by the code timestamps are not supported (ie rclone can't set them).

> Cloudinary does not support modification times, therefore
> syncing will default to `--size-only` checking.  Note that using
> `--update` will work.

ncw (commented):
This is a great start - thank you :-)

I've put some quick notes inline on things which jumped out.

Could it support more optional features? `Move` would enable things like `--backup-dir` and renaming files in `rclone mount` which makes users very happy!

This is going to generate a lot of lint errors which will need fixing, alas.

yuval-cloudinary:
> Great work :-)
> 
> How many of the integration tests are passing? `test_all -backends cloudinary`?
> 
> I notice you use a cloudinary go package - can you run
> 
> ```
> bin/test_backend_sizes.py cloudinary
> ```
> 
> To see how much size this is adding to the rclone binary?
> 
> I'll now put a few very quick comments on the code - not a full review yet, but some obvious bits

Thanks for the warm welcome :D

3 tests have failed, and 2 passed.

The failure of TestSyncOverlapWithFilter is due to the backend folder deletion that deletes folders with no assets regardless if there are empty folders in it. If needed, I can add a 2nd call to check for sub-folders at the expense of performance and the chance to hit the Admin API rate-limit. What would you recommend?

I need to further analyze the failures of TestParseSumFile, and TestFileSetModTime

The sizes are 
Total binary size 63.731 MiB
Backend,Size MiB
cloudinary,0.432

What would be an acceptable limit?

yuval-cloudinary:
> This is a great start - thank you :-)
> 
> I've put some quick notes inline on things which jumped out.
> 
> Could it support more optional features? `Move` would enable things like `--backup-dir` and renaming files in `rclone mount` which makes users very happy!
> 
> This is going to generate a lot of lint errors which will need fixing, alas.

Cloudinary support Move, and I implemented it and then discarded it. The issue is that Move conflicts with the duplicate files that Cloudinary allows, alas, with a different logic than rclone.

With rclone, duplicate files are distinguished by the modTime. With Cloudinary, duplicate files are distinguished by a tuple of delivery_type, resource_type, public_id. This tuple is much more functional than a timestamp, it is the component of a delivery URL that is attached to the file (many customers have their assets publicly available on websites and mobile apps). It's also much slower than modTime, as it involves a change in the storage layer. As a consequence,  I can't Move the public_id when I move the asset. But I also cannot keep it the same, as I'll overwrite the renamed/moved asset when uploading a new asset with the same public_id.

backend/cloudinary/cloudinary.go
yuval-cloudinary: Good catch, as this is an int. I'll request the SDK owner to allow larger types

docs/content/cloudinary.md
yuval-cloudinary: Both md5 and timestamps are set internally (for any successful Put) and are read-only.
I'm happy to rephrase or remove this section per your advice

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+		Options: []fs.Option{
+			{
+				Name:     "cloud_name",
+				Help:     "Cloudinary Environment Name",
+				Required: true,
+			},
+			{
+				Name:     "api_key",
+				Help:     "Cloudinary API Key",
+				Required: true,
```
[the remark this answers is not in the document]
yuval-cloudinary: Done

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+			},
+			{
+				Name:       "api_secret",
+				Help:       "Cloudinary API Secret",
+				Required:   true,
+				IsPassword: true,
+			},
+			{
+				Name: "upload_prefix",
+				Help: "Specify alternative data center",
```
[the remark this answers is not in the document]
yuval-cloudinary: I revised the description, this is usually considered essential

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+					encoder.EncodeDel |
+					encoder.EncodeCtl |
+					encoder.EncodeRightSpace |
+					encoder.EncodeInvalidUtf8 |
+					encoder.EncodeDot),
+			},
+			{
+				Name:     "eventually_consistent_delay",
+				Advanced: true,
+				Help:     "Wait N seconds for eventual consistency",
```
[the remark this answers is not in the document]
yuval-cloudinary: Thanks! Done

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+
+// Options defines the configuration for this backend
+type Options struct {
+	CloudName                 string               `config:"cloud_name"`
+	APIKey                    string               `config:"api_key"`
+	APISecret                 string               `config:"api_secret"`
+	UploadPrefix              string               `config:"upload_prefix"`
+	UploadPreset              string               `config:"upload_preset"`
+	Enc                       encoder.MultiEncoder `config:"encoding"`
+	EventuallyConsistentDelay string               `config:"eventually_consistent_delay"`
```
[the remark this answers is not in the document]
yuval-cloudinary: Done

backend/cloudinary/cloudinary.go
yuval-cloudinary: For the wrong reasons... I refactored the needed code, and it looks better. Done

docs/content/cloudinary.md
yuval-cloudinary: Done

docs/content/cloudinary.md
yuval-cloudinary: Done

docs/content/cloudinary.md
yuval-cloudinary: Thanks, I submitted https://github.com/rclone/rclone/pull/8053
I think it would be good to keep a reference in the Cloudinary backend doc

backend/cloudinary/cloudinary.go
ncw: `int64` is a good choice IMHO - it is the one that Go uses everywhere in its APIs. Some people prefer `uint64` but I don't think not being able to address files larger than 8 ExaBytes is a problem we are likely to have any time soon!

docs/content/cloudinary.md
ncw: You can link to other parts of the documentation using `[link](/docs/#anchor)` in markdown syntax

docs/content/cloudinary.md
ncw: Yes that is fine. If rclone can't set the timestamps itself, then we say timestamps are not supported, so you'll need the wording above. Normally rclone would check the timestamp is exactly equal on the source and the destination (this is the way rsync works too) but if we can't set the timestamp then we have to fall back to `--update` which usually works pretty well).

ncw:
> 3 tests have failed, and 2 passed.

If you'd like help analysing what happened, then zip up the test_all output directory and attach it, and I'll analyse it for you.

> The failure of TestSyncOverlapWithFilter is due to the backend folder deletion that deletes folders with no assets regardless if there are empty folders in it. If needed, I can add a 2nd call to check for sub-folders at the expense of performance and the chance to hit the Admin API rate-limit. What would you recommend?

Are you saying that `Fs.Rmdir` will delete a folder which has other empty folders within in? If so then we should check for sub folders.

Or are you saying the backend deletes folders if they have no files in automatically? In which case perhaps we should be setting `CanHaveEmptyDirectories: false` in the Feature flags? That behavior sounds very similar to s3 which doesn't really have directories at all. The tests then will stop caring about directories being present or not in that case.

> I need to further analyze the failures of TestParseSumFile, and TestFileSetModTime
> 
> The sizes are Total binary size 63.731 MiB Backend,Size MiB cloudinary,0.432
> 
> What would be an acceptable limit?

That is not too bad! The backends built with just `lib/rest` come in about 100k but some of the backends are monsters as they include massive SDKs! 432k puts you in at about 15/57 so OK but smaller would be nicer :-)

> > Could it support more optional features? `Move` would enable things like `--backup-dir` and renaming files in `rclone mount` which makes users very happy!
> > This is going to generate a lot of lint errors which will need fixing, alas.
> 
> Cloudinary support Move, and I implemented it and then discarded it. The issue is that Move conflicts with the duplicate files that Cloudinary allows, alas, with a different logic than rclone.
> 
> With rclone, duplicate files are distinguished by the modTime. With Cloudinary, duplicate files are distinguished by a tuple of delivery_type, resource_type, public_id. This tuple is much more functional than a timestamp, it is the component of a delivery URL that is attached to the file (many customers have their assets publicly available on websites and mobile apps). It's also much slower than modTime, as it involves a change in the storage layer. As a consequence, I can't Move the public_id when I move the asset. But I also cannot keep it the same, as I'll overwrite the renamed/moved asset when uploading a new asset with the same public_id.

Actually rclone doesn't deal with duplicate files at all well! It assumes that backends don't have duplicate file names in normal operation (as you can't have them on local file systems or most backends (eg s3)). It provides a tool `rclone dedupe` to help you rename duplicate files.

So most rclone users will expect there not to be duplicate files.

Are duplicate files common with cloudinary? If so then we should probably make sure that we make sure the file names are different. The [google photos backend does this](https://rclone.org/googlephotos/#duplicates) as duplicates are really common there.

....

One other thing I was thinking of - does the cloudinary SDK use rclone's http.Client or http.Transport? I don't think it does from a skim of the code. If you don't use rclone's http Client or Transport you'll miss out on a lot of rclone features such as accurate bandwidth limiting, QoS, `--dump headers`, `--dump bodies`, `--timeout`, `--contimeout` etc etc. This might be another request for the SDK team but being able to specify your own client/transport is essential. You'll see that all the other http SDK's rclone uses support this (some only after being nagged by me :-)

docs/content/cloudinary.md
yuval-cloudinary: Done

docs/content/cloudinary.md
yuval-cloudinary: Done

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+}
+
+// Fs represents a remote cloudinary server
+type Fs struct {
+	name     string
+	root     string
+	opt      Options
+	features *fs.Features
+	srv      *rest.Client // the connection to the server
```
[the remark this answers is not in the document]
yuval-cloudinary: Done

backend/cloudinary/cloudinary.go
```diff
@@ -0,0 +1,626 @@
...
+				fs.Logf(o, "Unsupported mandatory option: %v", option)
+			}
+		}
+	}
+	if key != "" && value != "" {
+		opts.ExtraHeaders = make(map[string]string)
+		opts.ExtraHeaders[key] = value
+	}
+	// Make sure that the asset is fully available
+	for i := 1; i <= 4; i++ {
```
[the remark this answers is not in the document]
yuval-cloudinary: Done

yuval-cloudinary:
> > 3 tests have failed, and 2 passed.
> 
> If you'd like help analysing what happened, then zip up the test_all output directory and attach it, and I'll analyse it for you.
> 
> > The failure of TestSyncOverlapWithFilter is due to the backend folder deletion that deletes folders with no assets regardless if there are empty folders in it. If needed, I can add a 2nd call to check for sub-folders at the expense of performance and the chance to hit the Admin API rate-limit. What would you recommend?
> 
> Are you saying that `Fs.Rmdir` will delete a folder which has other empty folders within in? If so then we should check for sub folders.
> 
> Or are you saying the backend deletes folders if they have no files in automatically? In which case perhaps we should be setting `CanHaveEmptyDirectories: false` in the Feature flags? That behavior sounds very similar to s3 which doesn't really have directories at all. The tests then will stop caring about directories being present or not in that case.
> 
> > I need to further analyze the failures of TestParseSumFile, and TestFileSetModTime
> > The sizes are Total binary size 63.731 MiB Backend,Size MiB cloudinary,0.432
> > What would be an acceptable limit?
> 
> That is not too bad! The backends built with just `lib/rest` come in about 100k but some of the backends are monsters as they include massive SDKs! 432k puts you in at about 15/57 so OK but smaller would be nicer :-)
> 
> > > Could it support more optional features? `Move` would enable things like `--backup-dir` and renaming files in `rclone mount` which makes users very happy!
> > > This is going to generate a lot of lint errors which will need fixing, alas.
> > 
> > 
> > Cloudinary support Move, and I implemented it and then discarded it. The issue is that Move conflicts with the duplicate files that Cloudinary allows, alas, with a different logic than rclone.
> > With rclone, duplicate files are distinguished by the modTime. With Cloudinary, duplicate files are distinguished by a tuple of delivery_type, resource_type, public_id. This tuple is much more functional than a timestamp, it is the component of a delivery URL that is attached to the file (many customers have their assets publicly available on websites and mobile apps). It's also much slower than modTime, as it involves a change in the storage layer. As a consequence, I can't Move the public_id when I move the asset. But I also cannot keep it the same, as I'll overwrite the renamed/moved asset when uploading a new asset with the same public_id.
> 
> Actually rclone doesn't deal with duplicate files at all well! It assumes that backends don't have duplicate file names in normal operation (as you can't have them on local file systems or most backends (eg s3)). It provides a tool `rclone dedupe` to help you rename duplicate files.
> 
> So most rclone users will expect there not to be duplicate files.
> 
> Are duplicate files common with cloudinary? If so then we should probably make sure that we make sure the file names are different. The [google photos backend does this](https://rclone.org/googlephotos/#duplicates) as duplicates are really common there.
> 
> ....
> 
> One other thing I was thinking of - does the cloudinary SDK use rclone's http.Client or http.Transport? I don't think it does from a skim of the code. If you don't use rclone's http Client or Transport you'll miss out on a lot of rclone features such as accurate bandwidth limiting, QoS, `--dump headers`, `--dump bodies`, `--timeout`, `--contimeout` etc etc. This might be another request for the SDK team but being able to specify your own client/transport is essential. You'll see that all the other http SDK's rclone uses support this (some only after being nagged by me :-)

test_all: We now have 2 tests failing after resolving the RmDir (TestSyncOverlapWithFilter) issue, please see the output:
[test-all.zip](https://github.com/user-attachments/files/16924022/test-all.zip)

The TestSyncOverlapWithFilter is resolved now. That was the first option where the backend ignores empty folders when testing if a folder is eligible for deletion.

Duplicate files in Cloudinary exist, and it's not clear how to deal with them. Initially, I thought to return an error if I discover more than 1 file, but this is not that robust (if a user searches for an asset in a certain time range, or if a user uploads a file and creates a duplicate inadvertently).

I can request the support of http.Client and http.Transport. Can you share an overview and/or a specific example?

ncw:
This failure looks straight forward

```
"./operations.test -test.v -test.timeout 1h0m0s -remote TestCloudinary: -test.run '^TestParseSumFile$'" - Starting (try 5/5)
=== RUN   TestParseSumFile
    run.go:180: Remote "Cloudinary root 'rclone-test-nogejog3xayo'", Local "Local file system at /var/folders/k2/39p_26g53xq0jb_7lss43cvc0000gq/T/rclone2137360952", Modify Window "876000h0m0s"
    check_test.go:313: 
        	Error Trace:	/Users/yuvalpapish/dev/rclone/fs/operations/check_test.go:313
        	            	object not found
        	Test:       	TestParseSumFile
```

Rclone wrote a test file, but NewObject didn't find it.

I suspect this is eventual consistency at work - is that likely? If so then the test needs changing to use `fstest.NewObject` instead.

https://github.com/rclone/rclone/blob/da70088b11e633999f2f920d3c603d97c292135e/fs/operations/check_test.go#L311-L313

The subsequent crash is because the object is nil, that implies that the `assert` above should be `require`


----

The next errors in the VFS seem to have a few different causes. The duplicate files is probably because you need `Fs.Update` to update an existing file if there is one, not to create a new one. This might mean searching for an existing file and deleting it after the update is successful, or using a different API to update an existing file.

One of the errors seems to be an eventual consistency error

Not sure what "can't upload empty files to this remote" errors are about but I don't think any of those files should be empty so I conjecture it is a backend bug.

---- 

> Duplicate files in Cloudinary exist, and it's not clear how to deal with them. Initially, I thought to return an error if I discover more than 1 file, but this is not that robust (if a user searches for an asset in a certain time range, or if a user uploads a file and creates a duplicate inadvertently).

Just return both of them and rclone will issue a warning.
 
> I can request the support of http.Client and http.Transport. Can you share an overview and/or a specific example?

Libraries which do HTTP requests on behalf of a user should allow either an http.Transport (of type [http.RoundTripper](https://pkg.go.dev/net/http#RoundTripper) or [http.Client](https://pkg.go.dev/net/http#Client) to be passed in. This enables their use in restrictive environments (eg all requests have to go via a SOCKs proxy). It also enables their use in Google appengine where it is mandatory to use the http.Transport supplied by the appengine runtime.

As far as rclone is concerned, using the rclone transport will make sure the rate limiting `--bwlimit` is properly applied, enable debugging of the requests with `--dump headers` and `--dump bodies`, enable that QoS is applied, enable connections via SOCKs proxies, or unix sockets, enable `--timeout` and `--contimeout` and a lot more things that I've forgotten.

Here is how it is applied in some SDKs in use by rclone

s3

https://github.com/rclone/rclone/blob/da70088b11e633999f2f920d3c603d97c292135e/backend/s3/s3.go#L3097

azureblob

https://github.com/rclone/rclone/blob/da70088b11e633999f2f920d3c603d97c292135e/backend/azureblob/azureblob.go#L643

google drive

https://github.com/rclone/rclone/blob/da70088b11e633999f2f920d3c603d97c292135e/backend/drive/drive.go#L1276-L1277

orable object storage

https://github.com/rclone/rclone/blob/da70088b11e633999f2f920d3c603d97c292135e/backend/oracleobjectstorage/client.go#L89

ncw:
How are you getting on with this @yuval-cloudinary ? Do you need any help?

yuval-cloudinary:
> rclone/backend/s3/s3.go

Thanks for this example!

Cloudinary SDK supports external client, using:
	cld.Admin.Client = *client
	cld.Upload.Client = *client
	
However, the Rest Client that I'm using to Open files stopped returning the expected data. This happens regardless of using a shared client or separate instances. Any ideas?

yuval-cloudinary:
> > Just return both of them and rclone will issue a warning.

I adjusted the code to unique file names. It seems reasonable to not support duplicate file names.

> I suspect this is eventual consistency at work - is that likely? If so then the test needs to change to use fstest.NewObject instead.
Should I handle this change? In this PR or a new one?

> The subsequent crash is because the object is nil, which implies that the assert above should be require
Should I handle this change? In this PR or a new one?

ncw:
@yuval-cloudinary 
> Cloudinary SDK supports external client, using: cld.Admin.Client = *client cld.Upload.Client = *client

Great

> However, the Rest Client that I'm using to Open files stopped returning the expected data. This happens regardless of using a shared client or separate instances. Any ideas?

Can you link to the code?

> > Not sure what "can't upload empty files to this remote" errors are about but I don't think any of those files should be empty so I conjecture it is a backend bug.
> 
> This seems like a valid error that comes from an attempt to write an empty file, and is blocked https://github.com/cloudinary-devs/rclone/blob/63899d6de847ec4b707f16fe72af77da62aab166/backend/cloudinary/cloudinary.go#L410

I don't think the tests should be uploading empty files, or if they do they should expect this error. I've lost track of which test this is coming from - can you remind me?

> > > Just return both of them and rclone will issue a warning.
> 
> I adjusted the code to unique file names. It seems reasonable to not support duplicate file names.

OK

Will this backend work with a free cloudinary account? If so I should probably start some testing myself.

We are nearly there - it would be great to get this merged for v1.69

yuval-cloudinary:
The tests are passing now with the rclone HTTP client, so that's great.

> Not sure what "can't upload empty files to this remote" errors are about but I don't think any of those files should be empty so I conjecture it is a backend bug.

>> This seems like a valid error that comes from an attempt to write an empty file, and is blocked https://github.com/cloudinary-devs/rclone/blob/63899d6de847ec4b707f16fe72af77da62aab166/backend/cloudinary/cloudinary.go#L410

>>> I don't think the tests should be uploading empty files, or if they do they should expect this error. I've lost track of which test this is coming from - can you remind me?

This is coming from the analysis of test_all in https://github.com/rclone/rclone/pull/8051#issuecomment-2345995662

Also, should I rebase my PR and test again? How do we handle the changes in the testing code itself? (https://github.com/rclone/rclone/pull/8051#issuecomment-2360207655)

> Will this backend work with a free cloudinary account? If so I should probably start some testing myself.

Yes, it will work effortlessly with a free tier.
Make sure to create an [upload preset](https://console.cloudinary.com/settings/upload) with the following settings, to get maximal compliance with the test use case.
Discard original file name: true
Overwrite: true 
Use filename or externally defined Public ID: true 
Unique filename: false 
Use filename as display name: true 
Use asset folder as the Public ID prefix: false 
Delivery type: upload

> We are nearly there - it would be great to get this merged for v1.69
Absolutely agree!

ncw:
I configured myself a free Cloudinary account and tested the backend. It works :-)

I noted these things:

- The walkthrough in the docs needs updating - there have been changes in the parameters since it was written
- I uploaded a load of files but none of them had an MD5 checksum and the backend says it supports MD5 checksums
- Do the settings in the upload profile need documenting?
- `api_secret` does not want `Password: true` as this obfuscates token. Suggest

```diff
--- a/backend/cloudinary/cloudinary.go
+++ b/backend/cloudinary/cloudinary.go
@@ -64,10 +64,10 @@ func init() {
 			{
 				Name:       "api_secret",
 				Help:       "Cloudinary API Secret",
 				Required:   true,
-				IsPassword: true,
+				Sensitive: true,
 			},
```

----

I tried the tests but they did not pass as I was getting

    Failed to rmdir: Rate Limit Exceeded. Limit of 500 api operations reached. Try again on 2024-09-27 12:00:00 UTC

Does this come back in a machine readable format? If so then rclone should obey it (you can return the `fserrors.NewErrorRetryAfter(duration)` error from shouldRetry)

---- 

> > Not sure what "can't upload empty files to this remote" errors are about but I don't think any of those files should be empty so I conjecture it is a backend bug.
> 
> > > This seems like a valid error that comes from an attempt to write an empty file, and is blocked https://github.com/cloudinary-devs/rclone/blob/63899d6de847ec4b707f16fe72af77da62aab166/backend/cloudinary/cloudinary.go#L410
> 
> > > > I don't think the tests should be uploading empty files, or if they do they should expect this error. I've lost track of which test this is coming from - can you remind me?
> 
> This is coming from the analysis of test_all in [#8051 (comment)](https://github.com/rclone/rclone/pull/8051#issuecomment-2345995662)

I looked in `test.zip` and there is only one occurance of "Can't upload zero length files" in `TestIntegration/FsMkdir/FsPutZeroLength` which checks the correct error is received and the test passes

Which test do you see giving that error?

> Also, should I rebase my PR and test again? How do we handle the changes in the testing code itself? ([#8051 (comment)](https://github.com/rclone/rclone/pull/8051#issuecomment-2360207655))

Please do make changes to the testing code as we've discussed above - just put them in a separate commit before your main backend commit.

Rebase and squash would be great. Use 2 commits if you are modifying testing code, 1 otherwise.

> > Will this backend work with a free cloudinary account? If so I should probably start some testing myself.
> 
> Yes, it will work effortlessly with a free tier. Make sure to create an [upload preset](https://console.cloudinary.com/settings/upload) with the following settings, to get maximal compliance with the test use case. Discard original file name: true Overwrite: true Use filename or externally defined Public ID: true Unique filename: false Use filename as display name: true Use asset folder as the Public ID prefix: false Delivery type: upload

Should this be documented in the docs?

yuval-cloudinary:
> I configured myself a free Cloudinary account and tested the backend. It works :-)
> 
> I noted these things:
> 
> * The walkthrough in the docs needs updating - there have been changes in the parameters since it was written
> * I uploaded a load of files but none of them had an MD5 checksum and the backend says it supports MD5 checksums
> * Do the settings in the upload profile need documenting?
> * `api_secret` does not want `Password: true` as this obfuscates token. Suggest
> 
> ```diff
> --- a/backend/cloudinary/cloudinary.go
> +++ b/backend/cloudinary/cloudinary.go
> @@ -64,10 +64,10 @@ func init() {
>  			{
>  				Name:       "api_secret",
>  				Help:       "Cloudinary API Secret",
>  				Required:   true,
> -				IsPassword: true,
> +				Sensitive: true,
>  			},
> ```
> 
> I tried the tests but they did not pass as I was getting
> 
> ```
> Failed to rmdir: Rate Limit Exceeded. Limit of 500 api operations reached. Try again on 2024-09-27 12:00:00 UTC
> ```
> 
> Does this come back in a machine readable format? If so then rclone should obey it (you can return the `fserrors.NewErrorRetryAfter(duration)` error from shouldRetry)

* The documentation was regenerated
* I replace the api_secret to be Sensitive
* The upload_preset settings I suggested were tailored to the test suite. In real life, the scenarios could be different
* I added parsing so we can enjoy fserrors.NewErrorRetryAfter(duration)

yuval-cloudinary:
Here is a revised test_all results, after rebasing to 1.68.1 and amending the test that crashed
[test_all.zip](https://github.com/user-attachments/files/17327649/test_all.zip)

ncw:
Oops, sorry, missed your updated.

I see this summary in your .zip which is nearly all of the tests passing! Tests you don't think will ever pass we can mark as such in `fstest/test_all/config.yaml`.

Do you think we can make the remaining ones pass? Do you want me to help?

I'd like to merge this backend for 1.69 if possible.

Thank you

image

yuval-cloudinary:
@ncw The most common failure seems to be the hashing, yet this is a rock-solid functionality used by many customers. It seems to me that this is the reported error, for different failures.

  Which tests do you think I should disable in the config.yaml file?

yuval-cloudinary:
@ncw I've adjusted the config.yaml so test_all passes without any issue

ncw:
I've rebased this and squashed the commits. I couldn't push it back to your branch as you haven't allowed that so I've merged it in 7e4b8e33f596da442e41fbc496450e75f5907d6c and 5151a663f04a497ecc86957eee361c640bdd8b1c

The integration tests will run daily and will appear here in due course: https://pub.rclone.org/integration-tests/current/

Thank you for a great contribution and sorry it has taken such a long time to merge.

ncw:
@yuval-cloudinary the integration test results are in and I get this :-)

    Rate Limit Exceeded. Limit of 500 api operations reached. Try again on 2025-01-09 05:00:00 UTC

Any hints on avoiding this rate limit? My cloud name is `durakvfdu`

yuval-cloudinary:
> Any hints on avoiding this rate limit? My cloud name is `durakvfdu`
The integration tests are fairly intensive ;)  I raised the limit to 15k calls/hr. Please let me know if this is still an issue

ncw:
Thanks @yuval-cloudinary - I'll see tomorrow!

PR #8080, opened 2024-09-18
`local`: fix `--copy-links` on macOS when cloning

nielash:

Before this change, `--copy-links` erroneously behaved like `--links` when using cloning on macOS, and cloning was not supported at all when using `--links`.

After this change, `--copy-links` does what it's supposed to, and takes advantage of cloning when possible, by copying the file being linked to instead of the link itself.

Cloning is now also supported in `--links` mode for regular files (which benefit most from cloning). symlinks in `--links` mode continue to be tossed back to be handled by rclone's special translation logic.

- https://forum.rclone.org/t/macos-local-to-local-copy-with-copy-links-causes-error/47671/5?u=nielash

#### Checklist

### DIFF
--- a/backend/local/clone_darwin.go
+++ b/backend/local/clone_darwin.go
@@ -6,6 +6,7 @@ package local
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"runtime"
 
 	"github.com/go-darwin/apfs"
@@ -22,12 +23,15 @@ import (
 //
 // If it isn't possible then return fs.ErrorCantCopy
 func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object, error) {
-	if runtime.GOOS != "darwin" || f.opt.TranslateSymlinks || f.opt.NoClone {
+	if runtime.GOOS != "darwin" || f.opt.NoClone {
 		return nil, fs.ErrorCantCopy
 	}
 	srcObj, ok := src.(*Object)
 	if !ok {
 		fs.Debugf(src, "Can't clone - not same remote type")
+		return nil, fs.ErrorCantCopy
+	}
+	if f.opt.TranslateSymlinks && srcObj.translatedLink { // in --links mode, use cloning only for regular files
 		return nil, fs.ErrorCantCopy
 	}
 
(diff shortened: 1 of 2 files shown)


backend/local/clone_darwin.go
```diff
@@ -44,11 +50,21 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 		return nil, err
 	}
 
-	err = Clone(srcObj.path, f.localPath(remote))
+	srcPath := srcObj.path
+	if f.opt.FollowSymlinks { // in --copy-links mode, find the real file being pointed to and pass that in instead
+		linkPath, err := os.Readlink(srcPath)
```
ncw, line 55: This won't work if the symlink points to another symlink...

Resolving symlinks is a pain as you have to detect loops also, so we should try to get the kernel to do it if we can.

I would have thought we want to run the Clone without `CLONE_NOFOLLOW`/`COPYFILE_NOFOLLOW` or something like that? Though you don't appear to be using those flags so there is a gap in my understanding!

Actually I found [filepath.EvalSymlinks](https://pkg.go.dev/path/filepath?utm_source=godoc#EvalSymlinks) which looks like it would do the job if we can't get Clone to do it

Here is [the source for the meat of filpath.EvalSymlinks](https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/path/filepath/symlink.go;l=16) - as you see it is quite non trivial resolving symlinks!

ncw (commented):
This looks like a great fix - thank you :-)

I think we need to improve the symlink resolution as you'll see inline.

Do you think this should go in the point release v1.68.1 once we are happy? (Update - I see you added the tag already :+1:)

backend/local/clone_darwin.go
nielash: > I would have thought we want to run the Clone without `CLONE_NOFOLLOW`/`COPYFILE_NOFOLLOW` or something like that?

This was my first instinct too, and there is a first draft [here](https://github.com/nielash/rclone/commit/736c31803b8dcb6b8ac7952fa45c811447d44098) where I tried that. The problem is... while it does fix the bug, it seems to just copy instead of clone.

It is possible that there is some combination of flags that would do what we want, but that one wasn't it. Any ideas about what it might be?

I also came across [this comment](https://github.com/apple-oss-distributions/copyfile/blob/ed3f0a8bf8b6bac6838c92c297afcc826fec75f4/copyfile.c#L736) which seems to suggest it might not be possible...

ncw: That comment in the kernel looks very relevant

    // or we're cloning, where we never follow symlinks.

Do you want to update with `filepath.EvalSymlinks` then instead of `os.Readlink` - that should be a robust solution if Clone can't do it for us?

nielash: Yes, that sounds right to me.

nielash: Made that change.

ncw (approved):

I'll merge it now.

I think we've probably got enough things for a point release now!

Issue #7033, opened 2023-05-28
Speed and ETA are incorrect after a rescan


SimplicityGuy:
#### What is the problem you are having with rclone?
During an initial backup, some files fail and there are errors reported. rclone correctly rescans the source after all files have been uploaded. However, after the rescan the speed and ETA metrics do now update correctly.

#### What is your rclone version (output from `rclone version`)
v1.62.2

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Debian 11.7 (64 bit), kernel 5.10.0-20-amd64

#### Which cloud storage system are you using? (e.g. Google Drive)
Backblaze B2

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
`rclone sync --log-file=ams.rclone.log /mnt/md0 backup-b2-crypt:/target/`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
_did not capture_ 😞

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

animosity22:
The estimates are just that, estimates.

When you re-run something, you can't directly compare as it does a backlog of items in a non deterministic order so your results can vary greatly.

https://rclone.org/docs/#max-backlog-n

Retries and lots of other things can influence as well and that's all in the debug log.

SimplicityGuy:
I get that they're estimates, but the speed is completely off. I'd expect at least the speed to be reflected correctly.

animosity22:
The "why" it might inaccurate would be in the debug logs as retries and other things will influence it.

If you want to capture a full debug log and share it, that would shed light on the answer.

SimplicityGuy:
Ok, next backup I do I'll capture logs.

ncw:
> During an initial backup, some files fail and there are errors reported. rclone correctly rescans the source after all files have been uploaded. However, after the rescan the speed and ETA metrics do now update correctly. For instance, this is the estimate after a rescan: `21.510 KiB/s, ETA 1w1d12h`, meanwhile before the rescan the estimate was `1.142 MB/s, ETA 1h20m`

[...]

PR #8081, opened 2024-09-18
stats: fix the speed not getting updated after a pause in the processing

darthShadow:

This shifts the behavior of the average loop to be a persistent loop that gets resumed/paused when transfers & checks are started/completed.

Previously, the averageLoop was stopped on completion of transfers & checks but failed to start again due to the protection of the sync.Once

#7033

https://forum.rclone.org/t/speed-statistic-not-updating/46689

https://forum.rclone.org/t/rclone-copy-speed-statistic-not-updating/41860

#### Checklist


ncw (commented):
Thanks for taking a look at this.

I think we can do it a lot simpler though with a flag in the `average` struct, something like this (untested)

```diff
diff --git a/fs/accounting/stats.go b/fs/accounting/stats.go
index e46f95e5c..280c8ff34 100644
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -71,8 +71,7 @@ type averageValues struct {
        speed   float64
        stop    chan bool
        stopped sync.WaitGroup
-       startOnce sync.Once
-       stopOnce  sync.Once
+       started bool
 }
 
 // NewStats creates an initialised StatsInfo
@@ -85,7 +84,7 @@ func NewStats(ctx context.Context) *StatsInfo {
                transferring: newTransferMap(ci.Transfers, "transferring"),
                inProgress:   newInProgress(ctx),
                startTime:    time.Now(),
-               average:      averageValues{stop: make(chan bool)},
+               average:      averageValues{},
        }
 }
 
@@ -362,20 +361,23 @@ func (s *StatsInfo) averageLoop() {
 func (s *StatsInfo) startAverageLoop() {
        s.mu.RLock()
        defer s.mu.RUnlock()
-       s.average.startOnce.Do(func() {
+       if !s.average.started {
+               s.average.stop = make(chan bool)
+               s.average.started = true
                s.average.stopped.Add(1)
                go s.averageLoop()
-       })
+       }
 }
 
 // Stop the average loop
 //
 // Call with the mutex held
 func (s *StatsInfo) _stopAverageLoop() {
-       s.average.stopOnce.Do(func() {
+       if s.average.started {
+               s.average.started = false
                close(s.average.stop)
                s.average.stopped.Wait()
-       })
+       }
 }
 
 // Stop the average loop
```

What do you think?

darthShadow:
Sure, I can give that a try.

However, I’m still not clear on exactly what the speed is supposed to represent on a mount. It can mean one of 2 things:
* The average speed of all files transferred across the lifetime of the mount.
* The average of the last set of transfers.

The latter seems to be the current approach based on the numbers, but both approaches make sense to me. It may also be that the calculation is not right when dealing with pausing/resuming transfers.

ncw:
> Sure, I can give that a try.

:+1:

> However, I’m still not clear on exactly what the speed is supposed to represent on a mount. It can mean one of 2 things:
> 
> * The average speed of all files transferred across the lifetime of the mount.
> * The average of the last set of transfers.
> 
> The latter seems to be the current approach based on the numbers, but both approaches make sense to me. It may also be that the calculation is not right when dealing with pausing/resuming the average loop.

Hmm interesting point! It is certainly the second option now, but what will it be when the resuming works. Looking at the code I think it will be the average of the last set of transfers still.

ncw:
Yes... It is probably still useful as a measure of how fast the mount can transfer stuff - hopefully!

darthShadow:
The tests are passing now, but I had to upgrade the RO lock to a RW lock, not sure if that matters here.

darthShadow:
@ncw Let me know if you have any comments or issues with this.

darthShadow:
@ncw Let me know if you have any issues with this approach and will consider merging it if the conflicts are resolved. I don't want to go through the trouble of rebasing the fix just to let it go stale again and again get conflicts later.

ncw:
Looks like we shouldn't be running the build and push docker stuff on PRs or branches!

darthShadow:
Yeah, looks like the default Github Token doesn't allow package write permissions for pull requests. We could modify the workflow to use a more permissive token, but then there is nothing preventing a rogue PR (from a trusted account) from overwriting a tagged image.

References:

* https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
* https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action

If it's fine with you, I will send a separate PR disabling the docker workflows for PR's.

ncw:
> Yeah, looks like the default Github Token doesn't allow package write permissions for pull requests. We could modify the workflow to use a more permissive token, but then there is nothing preventing a rogue PR (from a trusted account) from overwriting a tagged image.

Not good!

> If it's fine with you, I will send a separate PR disabling the docker workflows for PR's.

Good idea.

PR #8089, opened 2024-09-21
lib/batcher: Deprecate unused option: batch_commit_timeout

dmcardle:

To ~~delete~~ deprecate an unused option in lib/batcher/.

I found that lib/batcher/ defines a `batch_commit_timeout` config, which is picked up by the dropbox and googlephotos backends, but couldn't find a struct field anywhere that hooks up to it.

With `git log -S`, I found that the code using the `BatchCommitTimeout` field was deleted in a1a8c21c709c1051df18d2b94e9eeecc158c7bc2, and then the struct field itself was deleted in b94806a1430172419bd945340611068bcaf6ff00. But the option's definition stuck around and so did the documentation, e.g. [RCLONE_DROPBOX_BATCH_COMMIT_TIMEOUT](https://github.com/rclone/rclone/blob/f8d782c02de5e42c354f281b4f32c4128fa5db1e/docs/content/dropbox.md?plain=1#L466-L469)!

#### Checklist

### DIFF
--- a/lib/batcher/options.go
+++ b/lib/batcher/options.go
@@ -66,8 +66,9 @@ default based on the batch_mode in use.
 		Advanced: true,
 	}, {
 		Name:     "batch_commit_timeout",
-		Help:     `Max time to wait for a batch to finish committing`,
+		Help:     `Max time to wait for a batch to finish committing. (no longer used)`,
 		Default:  fs.Duration(10 * time.Minute),
 		Advanced: true,
+		Hide:     fs.OptionHideBoth,
 	}}
 }


dmcardle:
@ncw PTAL!

I ran `make doc` locally but it made a LOT of other changes. Is this something you typically do before a release rather than continuously?

ncw:
> I ran `make doc` locally but it made a LOT of other changes. Is this something you typically do before a release rather than continuously?

Yes just before a release normally :-)

lib/batcher/options.go
```diff
@@ -64,10 +63,5 @@ default based on the batch_mode in use.
 `, opt.DefaultTimeoutAsync, opt.DefaultTimeoutSync),
 		Default:  fs.Duration(0),
 		Advanced: true,
-	}, {
-		Name:     "batch_commit_timeout",
-		Help:     `Max time to wait for a batch to finish committing`,
-		Default:  fs.Duration(10 * time.Minute),
-		Advanced: true,
```
ncw, line 71: We don't normally remove options as it can break peoples setups.

We can hide them from the docs and configurator like this

https://github.com/rclone/rclone/blob/84cebb6872cd8786e86357407f9159ec4386ba85/backend/azureblob/azureblob.go#L369-L375

ncw (commented):
I think this flag was a feature that I was thinking of making but never did!

It would be easy to implement though - just add a [deadline](https://pkg.go.dev/context#WithDeadline) to the `ctx` here

https://github.com/rclone/rclone/blob/84cebb6872cd8786e86357407f9159ec4386ba85/lib/batcher/batcher.go#L148-L149

Though making that feature work is potentially a change in behaviour...

lib/batcher/options.go
dmcardle: Thanks, that makes sense. I tried to apply this approach. PTAL.

dmcardle:
@ncw Forgot about this PR. I've now changed it from deleting the config to *deprecating* it. PTAL!

ncw (approved):
This looks great now thank you - will merge.

I forgot about this PR too :-)

Issue #7192, opened 2023-07-31
Rclone encryption improvements discussion


tomekit:
We're running [S3Drive](https://s3drive.app/) (GUI for S3 on desktop, mobile, web) and recently aligned with Rclone's encryption scheme for better interoperability and features like drive mount and Webdav that we could get out of the box.

Prior to that we've researched research couple different encryption schemes and despite current Rclone encryption shortcomings (from out perspective) we've went for it, knowing that there is a room for improvements.

Disclaimer, I am not a cryptographer, but rather an engineer who by the sheer luck of project choices happen to be more involved with cryptography for the past couple years.

### 1. No envelope encryption
The biggest shortcoming so far that we see, is that every single Rclone file content is encrypted by a single master key (derived from user provided password). On it's own this doesn't pose direct security threat since every file has its own random 24 bytes nonce, but it limits how you can securely exchange the encrypted file.

One of our features is file sharing, this feature is native to S3 protocol. When encrypted shared file is passed to the receiver they need to obtain the file decryption key. With a current Rclone's setup the decryption key happen to be user's master key. By sharing a single file the encryption key to ALL FILES is revealed which is very undesirable from a security point of view.

Before we've implemented Rclone's encryption, we've been using AWS client-side encryption: https://docs.aws.amazon.com/amazon-s3-encryption-client/latest/developerguide/encryption-algorithms.html#crypto_features it worked well, but it wasn't in the end feasible for us, mostly due to lack of consistent hardware accelerated streamed AES-GCM support across different platforms (web included).

What was great about AWS's approach is how they've implemented key management and key wrapping.
In other words, each file had a different encryption key (wrapped DEK), which would then be encrypted by the master key.
Sharing a resource would reveal only that single key (DEK). By choosing a key wrapping algorithm it would be possible to create multiple different encryption keys (by simply reencrypting the DEK).

I believe it's something that Rclone could use.

This change would have to be reflected in the ciphertext heading and would increase the length of a ciphertext in the ballpark of 16-32 bytes (depending on the KW algorithm).

### 2. No truncation protection
Currently, you can remove certain bits from the ciphertext and no one would notice. There are multiple ways to address that, but given the Rclone's ciphertext format, the one that would make sense in my opinion is: https://github.com/miscreant/meta/wiki/STREAM
In other words it would be good to reserve 1 byte of nonce for the: `last_block` flag.

### 3. Variable encryption settings
Current Rclone's settings are sane defaults which work well on a typical hardware.
[...]

PR #8095, opened 2024-09-23
crypt.go Implement V2 cipher version with MD5 hash support

tomekit:

Cipher improvements and MD5 hash support to crypt.
Commit description: "In V2 version each file encrypted using separate CEK (wrapped with AES Keywrap RFC3394). Truncation protection uses last byte of nonce to indicate last encryption block. Added MD5 hash support for crypt.go (only if V2 cipher is used), MD5 of plaintext is encrypted and authenticated at the end of file. Added 4 reserved bytes at the beginning for future use. Added 1 hash type byte in the footer to allow upgrade to different than MD5 hash. cipher_version setting determines hash version used for writes. For reads we determine cipher version by reading header magic bytes, that's regardless of cipher_version setting. A Added TestNewEncrypterV2 test to demonstrate the outputs and make sure that outputs match the expected hardcoded values. Modified other tests to satisfy newly added params (e.g. cek). Related to: https://github.com/rclone/rclone/issues/7192"

https://github.com/rclone/rclone/issues/7192

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -98,11 +98,14 @@ const (
 	NameEncryptionStandard
 	NameEncryptionObfuscated
 )
+
+// CipherVersion
 const (
 	CipherVersionV1 = "1"
 	CipherVersionV2 = "2"
 )
 
+// BlockCipher
 const (
 	BlockCipherXSalsa20 = 0x01 // Used currently
 	BlockCipherAESGCM   = 0x02 // Not used, possible future use (increase web performance): https://github.com/rclone/rclone/issues/7192#issuecomment-2353617267
@@ -251,33 +254,29 @@ func (c *Cipher) setCipherVersion(cipherVersion string) {
 func (c *Cipher) getFileHeaderSize() int {
 	if c.version == CipherVersionV2 {
 		return fileHeaderSizeV2
-	} else { // CipherVersionV1
-		return fileHeaderSize
 	}
+	return fileHeaderSize
 }
 
 func (c *Cipher) getFileMagicSize() int {
 	if c.version == CipherVersionV2 {
 		return fileMagicSizeV2
-	} else { // CipherVersionV1
-		return fileMagicSize
 	}
+	return fileMagicSize
 }
 
 func (c *Cipher) getFileMagicBytes() []byte {
 	if c.version == CipherVersionV2 {
 		return fileMagicBytesV2
-	} else { // CipherVersionV1
-		return fileMagicBytes
 	}
+	return fileMagicBytes
 }
 
 func (c *Cipher) getFileNonceSize() int { // Effective nonce size. Both V1 and V2 use 24 bytes, but V2 consists 23+1 bytes nonce, where last byte is calculated dynamically depending if the processing block is last (truncation protection)
 	if c.version == CipherVersionV2 {
 		return fileNonceSizeV2
-	} else { // CipherVersionV1
-		return fileNonceSize
 	}
+	return fileNonceSize
 }
 
 // Key creates all the internal keys from the password passed in using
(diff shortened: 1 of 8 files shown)


tomekit:
I've rebased most recent master changes on top of my changes, so I can carry on working and start addressing these issues.

tomekit:
I've addressed a bug (double close() call) which was causing some tests to fail.

I've also addressed Linter issues.

By any chance can we run the workflow again and see if there is anything left?

ncw:
I've run the CI.

Note that you can do `make check` to do the linting (or run `golangci-lint run ./...`) and `make quicktest` is what the CI does to run the tests.

tomekit:
It seems CI passed this time. I've found one more issue, which is now fixed. I was sure I was testing it properly, not sure how it slipped through !

I've noticed that there is still an issue with empty files `copy`. I am going to debug and understand what's going on.
```
2024/09/24 15:38:07 ERROR : test.txt: Failed to calculate dst hash: Hash decryption error: failed to authenticate decrypted block - bad password?
2024/09/24 15:38:07 ERROR : test.txt: corrupted on transfer: md5 hashes differ src(Local file system at /home/tomek/Downloads) "d41d8cd98f00b204e9800998ecf8427e" vs dst(Encrypted drive 's3drive_enc_backendtest1:compare') ""
2024/09/24 15:38:08 ERROR : Attempt 1/3 failed with 2 errors and: corrupted on transfer: md5 hashes differ src(Local file system at /home/tomek/Downloads) "d41d8cd98f00b204e9800998ecf8427e" vs dst(Encrypted drive 's3drive_enc_backendtest1:compare') ""
```

If there is no content, then there is no ciphertext and auth (that's alright), but should there be footer with encrypted: `d41d8cd98f00b204e9800998ecf8427e` (MD5) hash of empty string? I am not exactly sure, but given that for empty file we keep header it's likely we should also make sure that footer is still there.

ncw:
> If there is no content, then there is no ciphertext and auth (that's alright), but should there be footer with encrypted: `d41d8cd98f00b204e9800998ecf8427e` (MD5) hash of empty string? I am not exactly sure, but given that for empty file we keep header it's likely we should also make sure that footer is still there.

Yes I think a 0 byte file should encrypt to HEADER + FOOTER

## Notes for the future

I note also the crypt format is documented here https://github.com/rclone/rclone/blob/master/docs/content/crypt.md#file-formats so this patch should probably explain the differences between v1 and v2 there.

We could probably do with some docs - maybe a new section

- how v2 deals with v1 files (eg size might be wrong)
- how to upgrade from v1 to v2

Adding these docs later is fine.

We should probably put some [backend commands](https://github.com/rclone/rclone/blob/109e4ed0ed816415f283314566b4815df4b443ae/backend/crypt/crypt.go#L890) in which can 

- show the CEK for a given file
- given an encrypted file and a CEK, decrypt it

This will enable sharing of encrypted files

I'm happy to help design those commands.

tomekit:
> We could probably do with some docs - maybe a new section
>    -  how v2 deals with v1 files (eg size might be wrong)
>    - how to upgrade from v1 to v2

I am going to prepare it in a few days.

> show the CEK for a given file

I've implemented `show-cek` command which returns hex encoded unwrapped CEK, that's 32 bytes, 64 characters when hex encoded.

> given an encrypted file and a CEK, decrypt it

Do we have any existing command (or tests) to decrypt a local file? It would be useful for me to look it up how things are done.

There are couple other changes that I've applied today.

After I've implemented MD5 hashes I haven't tested decryption. It was failing since footer was being included in the call to decryption function, it's now skipped: https://github.com/rclone/rclone/pull/8095/commits/e9c5ee4b485e352912453580feaa41e229848bf2

It seems that we may want to return empty string instead of: `hash.ErrUnsupported` for V1 objects, so hash verification doesn't fail if user decides to e.g. copy V1 object to another location whilst already having the CipherVersionV2 enabled for their remote: https://github.com/rclone/rclone/pull/8095/commits/b62d37ba6f8c636a45e3ae78c5c19389d7cc3f42

ncw:
> I've implemented `show-cek` command which returns hex encoded unwrapped CEK, that's 32 bytes, 64 characters when hex encoded.

Great.

> > given an encrypted file and a CEK, decrypt it
> 
> Do we have any existing command (or tests) to decrypt a local file? It would be useful for me to look it up how things are done.

I'd suggest decrypting to and from an rclone remote like this does

https://github.com/rclone/rclone/blob/498d9cfa8550421a0b49092e8d0ed3a7eebd1fa2/backend/drive/drive.go#L3727

> There are couple other changes that I've applied today.
> 
> After I've implemented MD5 hashes I haven't tested decryption. It was failing since footer was being included in the call to decryption function, it's now skipped: [e9c5ee4](https://github.com/rclone/rclone/commit/e9c5ee4b485e352912453580feaa41e229848bf2)

:+1:
 
> It seems that we may want to return empty string instead of: `hash.ErrUnsupported` for V1 objects, so hash verification doesn't fail if user decides to e.g. copy V1 object to another location whilst already having the CipherVersionV2 enabled for their remote: [b62d37b](https://github.com/rclone/rclone/commit/b62d37ba6f8c636a45e3ae78c5c19389d7cc3f42)

If v1 isn't configured then `Hashes()` should return empty set. If v2 is configured then `Hashes()` should return MD5.

If v1 is configured then `Hash` should always return unsupported. If v2 is configured and a hash isn't available then returning "" is the correct thing to do.

ncw:
I noticed one thing missing from the tests

Create a copy of this test

https://github.com/rclone/rclone/blob/498d9cfa8550421a0b49092e8d0ed3a7eebd1fa2/backend/crypt/crypt_test.go#L32-L33

But set the `v2` encryption and check everything still works.

That will run rclone's backend test suite which may find some more things which need fixing.

tomekit:
I've implemented: `backend decrypt` command: https://github.com/rclone/rclone/pull/8095/commits/41252c41f6b6f3347829203d6f39937ed7525d23

The example use case is:
```
backend show-cek crypt: test.txt
9e50074301baef7712fe1cd2cc56991b68ef3f82ee0c5c51c23510fab13d07cb

// Name encryption
backend encode crypt: test.txt
dM-_57OvFX3JLRAqjwX69Q

backend decrypt crypt_with_wrong_password: dM-_57OvFX3JLRAqjwX69Q 9e50074301baef7712fe1cd2cc56991b68ef3f82ee0c5c51c23510fab13d07cb ~/test_decrypted.txt
```

Please note that I've used remote: `crypt_with_wrong_password` for the last command. This is to test that `cek` is indeed properly injected instead of decrypter trying to unwrap it from the object header using master key.

> Create a copy of this test

Will do

> If v1 is configured then Hash should always return unsupported.

This will need fixing, I am going to get back to this

tomekit:
> If v1 isn't configured then Hashes() should return empty set. If v2 is configured then Hashes() should return MD5.
> 
> If v1 is configured then Hash should always return unsupported. If v2 is configured and a hash isn't available then returning "" is the correct thing to do.

It should now behave as you've described: https://github.com/rclone/rclone/pull/8095/commits/2718311b7128a332628b3f0ae5fbb4de841aeeb4

==
Regarding test failures after V2 cipher specific  test: https://github.com/tomekit/rclone/blob/envelope_encryption_7192_7/backend/crypt/crypt_test.go#L54 as you've suggested

There was an issue with empty file decryption after introducing MD5 footer, this was resolved here: https://github.com/rclone/rclone/pull/8095/commits/3c4f118063597a1f48a251a9f5793402d50ad2ff

The other issue of failing tests was that after introducing MD5 footer,  size and encrypted hash differences checks started to fail. On one side object was uploaded using this method: 
```
obj := uploadFile(t, localFs, path, contents)
```
which automatically picked up new V2 version changes (including MD5 footer).

On the other side of the test, object for comparison purposes was created using just:
```
enc, err := f.cipher.newEncrypter(wrappedIn, nil, nil)
_, err = io.Copy(&outBuf, enc)
```

This hasn't included the newly added footer hashes which had to be added to the test.

==

I will be slowly finishing the work week and taking hopefully well deserved holidays, I will still do some progress on crypt documentation changes over the next few days.

I also don't understand 100% consequences of recent changes (mostly MD5 footer) to decryption in `fillBuffer()`: https://github.com/tomekit/rclone/blob/envelope_encryption_7192_7/backend/crypt/cipher.go#L1061

I understand that we need to substract length of footer (HashEncryptedSizeWithHeader) in some cases, however I am not sure how this footer affects the: `isLastBlock` variable derived from `n < blockDataSize` and whether condition `n <= blockHeaderSize` to throw `ErrorEncryptedFileBadHeader` still stands.
In some cases it may happen that 65K ciphertext block contains only footer or that part of such footer sits in blockN and the other part in blockN+1.
I think adding some more decryption tests with variable ciphertext sizes in the range of 65k - 100 and 65k + 100 might be beneficial to confirm there aren't any issues in this area. I will try to get to that once I am back.

I would appreciate if you could take time to review my recent output and if there is anything in specific that needs implementing/changing/improving. Thanks a lot!

tomekit:
Moved to: https://github.com/rclone/rclone/pull/8105

Issue #3511, opened 2019-09-04
Server side copy - Box - item_name_in_use 409 error with existing files


a6:
Linked from [forum posting](https://forum.rclone.org/t/server-side-copy-box-item-name-in-use-409-error-with-existing-files/11365).

If I attempt a server side copy from one box path to another, any file existing in both paths that has been updated in the src path will not be updated in the dst path. Instead, an error is thrown:

    DEBUG : : Sizes differ (src 32 vs dst 25)
    ERROR : : Failed to copy: Error "item_name_in_use" (409): Item with the same name already exists

The command is something like:

    rclone copy boxRemote:path1 boxRemote:path2

where path2 is intended to have a copy of all files that exist in path1.

The first time this copy is called, everything works because there are no existing files of path1 in path2. However, if later, a file is updated in path1 and then the above copy command is attempted, this `item_name_in_use` error is thrown.

This is with rclone 1.48 on Ubuntu x64 16.04.

I suspect this is not desired behavior, because the same types of copy commands work just fine if src is local instead of a box remote.

ncw:
I reproduced this like this with the latest beta

```
$ date -Is | rclone rcat boxcom:test1/file1.txt
$ date -Is | rclone rcat boxcom:test2/file1.txt
$ rclone copy -vv --retries 1 --low-level-retries 1 boxcom:test2 boxcom:test1
2019/09/05 13:03:01 DEBUG : rclone: Version "v1.49.0-024-g5dad88ae-fix-3506-names-beta" starting with parameters ["rclone" "copy" "-vv" "--retries" "1" "--low-level-retries" "1" "boxcom:test2" "boxcom:test1"]
2019/09/05 13:03:01 DEBUG : Using config file from "/home/ncw/.rclone.conf"
2019/09/05 13:03:02 DEBUG : file1.txt: Modification times differ by -8s: 2019-09-05 05:02:38 -0700 -0700, 2019-09-05 05:02:30 -0700 -0700
2019/09/05 13:03:02 INFO  : box root 'test1': Waiting for checks to finish
2019/09/05 13:03:02 DEBUG : file1.txt: SHA-1 = b28a9df4b94db8bc1fb4ad48a3b0137d8d6f3348 (box root 'test2')
2019/09/05 13:03:02 DEBUG : file1.txt: SHA-1 = f261a573e4a89950c5e8d5d22367570274a987d8 (box root 'test1')
[...]

PR #8100, opened 2024-09-25
box: fix server-side copying a file over existing dst (and other bisync test fixes)

nielash:

Several fixes for bisync integration tests, most of which were discussed [here](https://github.com/rclone/rclone/pull/7795#issuecomment-2163295857).

- https://github.com/rclone/rclone/issues/3511
- https://forum.rclone.org/t/rclone-bisync-box-failed-to-create-file-system-is-a-directory-not-a-file/47908/19?u=nielash

#### Checklist

### DIFF
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -43,6 +43,7 @@ import (
 	"github.com/rclone/rclone/lib/jwtutil"
 	"github.com/rclone/rclone/lib/oauthutil"
 	"github.com/rclone/rclone/lib/pacer"
+	"github.com/rclone/rclone/lib/random"
 	"github.com/rclone/rclone/lib/rest"
 	"github.com/youmark/pkcs8"
 	"golang.org/x/oauth2"
@@ -256,7 +257,6 @@ func getQueryParams(boxConfig *api.ConfigJSON) map[string]string {
 }
 
 func getDecryptedPrivateKey(boxConfig *api.ConfigJSON) (key *rsa.PrivateKey, err error) {
-
 	block, rest := pem.Decode([]byte(boxConfig.BoxAppSettings.AppAuth.PrivateKey))
 	if len(rest) > 0 {
 		return nil, fmt.Errorf("box: extra data included in private key: %w", err)
@@ -619,7 +619,7 @@ func (f *Fs) CreateDir(ctx context.Context, pathID, leaf string) (newID string, 
 		return shouldRetry(ctx, resp, err)
 	})
 	if err != nil {
-		//fmt.Printf("...Error %v\n", err)
+		// fmt.Printf("...Error %v\n", err)
 		return "", err
 	}
 	// fmt.Printf("...Id %q\n", *info.Id)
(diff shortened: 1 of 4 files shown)


backend/box/box.go
```diff
@@ -966,6 +965,15 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 		return nil, err
 	}
 
+	// check if dest already exists
+	item, err := f.preUploadCheck(ctx, leaf, directoryID, src.Size())
+	if err != nil {
+		return nil, err
+	}
+	if item != nil { // dest already exists, fall back to manual copy
+		return nil, fs.ErrorCantCopy
```
ncw, line 974: Could we just delete the existing file here?

Or if paranoid
- rename existing file with random suffix
- do copy
- if successful delete existing file
- if unsucessful rename exsting file back

The success/unsuccess could be done in a `defer` function which looks at the result of `err`.

fstest/test_all/config.yaml
```diff
@@ -395,6 +395,10 @@ backends:
  - backend:  "cache"
    remote:   "TestCache:"
    fastlist: false
+   ignore:
```
ncw, line 398: This will cause the tests to get run and ignored which is fine but wastes CPU.

Perhaps we should expand the syntax of `config.yaml` slightly and add an `IgnoreTests` option?

https://github.com/rclone/rclone/blob/498d9cfa8550421a0b49092e8d0ed3a7eebd1fa2/fstest/test_all/config.go#L37

ncw (commented):
Thanks for doing this. I wonder how many integration tests those new tests are going to break :grimacing: - will you help me fix them up?

I put some comments inline

Thank you

backend/box/box.go
nielash: I like the paranoid option. I suppose there's still a slight risk of leaving things in a messy state if the rename-back fails, but I don't know if that's enough of a concern to worry about.

fstest/test_all/config.yaml
nielash: Good point. I do think that would be a handy option. Or, if it's a bigger change than you want to make right now, I could handle it with a `t.Skip` in the bisync test code.

nielash:
> I wonder how many integration tests those new tests are going to break 😬 - will you help me fix them up?

My guess: not as many as you'd think, or we'd have already seen them show up on the bisync tests. But it's possible! Do you have a way of running the integration tests on a branch other than `master`? So that we could evaluate the damage before deciding whether to merge?

backend/box/box.go
ncw: We do the paranoid option in one of the other backends I think.

Perhaps less failure prone would be
- do new copy to random suffix
- if successful delete existing then rename copy
    - or maybe just rename the copy over the existing if that is supported
- if unsucessful probably nothing to do

fstest/test_all/config.yaml
ncw: In 84cebb6872cd8786e86357407f9159ec4386ba85 I merged an `ignoretests` parameter :-)

ncw:
> > I wonder how many integration tests those new tests are going to break 😬 - will you help me fix them up?
> 
> My guess: not as many as you'd think, or we'd have already seen them show up on the bisync tests.

Good point

> But it's possible! Do you have a way of running the integration tests on a branch other than `master`? So that we could evaluate the damage before deciding whether to merge?

I do, yes. The code has to be on a branch pushed to the main rclone repo.

nielash:
Revised version here reflecting the changes we discussed.

> > But it's possible! Do you have a way of running the integration tests on a branch other than `master`? So that we could evaluate the damage before deciding whether to merge?
> 
> I do, yes. The code has to be on a branch pushed to the main rclone repo.

I see. I'm happy to make a branch in the main repo if you want to do a test run. (But I'm not too worried about it.)

ncw (approved):
Looks perfect, thank you :⁠-⁠)

Please merge

ncw:
Hi @nielash !

Just been looking through the integration tests for the new tests added here

Here is a condensed version of the failures

| backend | tests failing |
|---------|---------------|
| dropbox |	TestServerSideCopyOverSelf |
| gofile | TestServerSideCopyOverSelf |
| onedrive | TestServerSideCopyOverSelf |
| pikpak | TestServerSideCopyOverSelf |
| putio | TestServerSideCopyOverSelf |
| sugarsync | TestServerSideCopyOverSelf |
| yandex | TestServerSideCopyOverSelf |
| ulozto | TestServerSideMoveOverSelf |


Do you fancy having a go at any of those? I'm going to try doing a few to see how hard it is! I'm going to start with the `gofile` backend since I wrote that one quite recently

ncw:
@nielash I've done dropbox and gofile - its not proving hard, so I'll do the others :-)

Issue #7935, opened 2024-07-07
sftp: read ssh config file ~/.ssh/config


milahu:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/sftp-support-ssh-configuration-file/34395

#### What is your current rclone version (output from `rclone version`)?

rclone 1.66.0

- os/version: nixos 24.05 (Uakari) (64 bit)
- os/kernel: 6.6.30 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.22.2
- go/linking: dynamic
- go/tags: cmount

#### What problem are you are trying to solve?

replace rsync with rclone, without rclone config file

```diff
-rsync -raPv example.com:somedir/ somedir/
+rclone sync -Pv sftp:example.com:somedir/ somedir/
```

#### How do you think rclone should be changed to solve that?

like rsync, rclone should [read](https://github.com/kevinburke/ssh_config) my `~/.ssh/config` file
and use the settings from the `example.com` section

```
Host example.com
  HostName somesubdomain.example.com
  User someuser
  IdentitiesOnly yes
  IdentityFile ~/.ssh/id_rsa
```

currently i have to pass these options manually to rclone

```sh
rclone \
  --config /dev/null \
  sync \
  --progress \
  --verbose \
  --sftp-host somesubdomain.example.com \
  --sftp-user someuser \
  --sftp-key-file ~/.ssh/id_rsa \
  example.com,type=sftp:somedir/ somedir/
```

`type=sftp` is needed to fix

> Failed to create file system for "sftp:example.com:somedir/": didn't find section in config file

`--config /dev/null` is needed
so rclone does not add this config to `~/.config/rclone/rclone.conf`

#### protocol type

syntax sugar: let me set type like a url schema prefix

```diff
-example.com,type=sftp:somedir/
+sftp:example.com:somedir/
```

... or let me set type with `--type sftp`

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
Rclone can use the SSH binary directly if you want it to use your SSH config.

Check out https://rclone.org/sftp/#sftp-ssh

milahu:
[...]

PR #8102, opened 2024-09-25
flag  --use-ssh-config loads ssh config from ~/.ssh/config temporarily into environment with precedence for sftp backend

kivi:

Adding flag `--use-ssh-config`, which enables usage of configurations from ~/.ssh/config.

Example: `rclone ls --use-ssh-config remote-host:some-directory` 

Any other command other than `ls` should also work.

It was discussed in a github issue: https://github.com/rclone/rclone/issues/7935

#7935 

#### Checklist

- [ ?] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).


fs/config/configfile/configfile_ssh.go
```diff
@@ -0,0 +1,91 @@
...
+
+func ConvertToIniKey(customKey string) string {
+	if iniKey, found := keyMapping[strings.ToLower(customKey)]; found {
+		return iniKey
+	}
+
+	return customKey
+}
+
+type SshConfig map[string]map[string]string
```
milahu, line 28: one key can have multiple values

```
$ ssh -G example.com | grep ^identityfile
identityfile ~/.ssh/id_rsa
identityfile ~/.ssh/id_ecdsa
identityfile ~/.ssh/id_ecdsa_sk
identityfile ~/.ssh/id_ed25519
identityfile ~/.ssh/id_ed25519_sk
identityfile ~/.ssh/id_xmss
```

see also https://github.com/kevinburke/ssh_config

<blockquote>

Certain directives can occur multiple times for a host (such as `IdentityFile`),
so you should use the `GetAll` or `GetAllStrict` directive to retrieve those
instead.

```go
files := ssh_config.GetAll("myhost", "IdentityFile")
```

</blockquote>

kivi: Thanks @milahu. I will consider that in the next iteration.
At the moment my biggest problem it, its writing the merged config into rclone config file. So maybe it would be better to move the ssh config into `backend/ftp` but there I experienced some problems, so this is just a draft atm.

ncw:
Thank you @kivi for producing some code to look at.

I've looked through it and I like this iteration better.

I was thinking about not saving the info in the config file, but that made me think rclone already has a mechanism for that.

Instead of internally creating a config file, you could set environment variables. You can configure rclone backends entirely with environment variables - see [the docs on environment variables](https://rclone.org/docs/#config-file).

This would have a couple of advantages

- we could decouple it from config loading - it can essentially live in its own package in `lib/usesshconfig` say and only be called when the `--use-ssh-config` flag is set.
- we then don't have to worry about the merged config getting saved

What do you think?

kivi:
@ncw thanks for taking the time. I agree, that it should NOT save the into any config file. But that's the current behavior, when I do create that internal config file. 
As far I could grasp the rclone code, that was the only way for me get an initial working draft. 
I am honestly not understanding how `backend/sftp.go` is working [`NewFs`](https://github.com/rclone/rclone/blob/master/backend/sftp/sftp.go#L839C6-L839C11) is called and somehow the dependencies are injected. Where is that config for the chosen host is coming from? How can I add the config needed (or is using enviroment vars, as you suggested, a possible solution?)

`Init` in `backend/sftp` describes the parameters, and refers to `NewFs` - what is actually happening inside `Init` or with `Init`. What pattern is that and how is that actually working? I you could give me some more hints, so I could speed up towards a common solution?

And yes I will check/try that with rclone env vars.

kivi:
@ncw 

> we could decouple it from config loading - it can essentially live in its own package in lib/usesshconfig say and only be called when the --use-ssh-config flag is set.

we would still load the ~/.ssh/config file right? So decoupling in that sense sounds definitely the way we should take.

ncw:
> we would still load the ~/.ssh/config file right? So decoupling in that sense sounds definitely the way we should take.

Yes, load the ~/.ssh/config file but into environment variables.

fs/config/configfile/configfile_ssh.go
kivi: @milahu how would you provide more than one `key_file` using rclone sftp? I think it is not possible with the current rclone version.

kivi:
> > we would still load the ~/.ssh/config file right? So decoupling in that sense sounds definitely the way we should take.
> 
> Yes, load the ~/.ssh/config file but into environment variables.

Ok, I could it get it running and it's not overwriting the rclone config file. So that is an improvement.
But why not use `Option` of `sftp`, why set the env vars, which is then picked up from another function, that fills the config or `Option` of sftp.go.

ncw:
> > Yes, load the ~/.ssh/config file but into environment variables.
> 
> Ok, I could it get it running and it's not overwriting the rclone config file. So that is an improvement.

Great

> But why not use `Option` of `sftp`, why set the env vars, which is then picked up from another function, that fills the config or `Option` of sftp.go.

The environment variables end up in the `Option` in `sftp` eventually.

Setting environment variables decouples the config nicely which is a big plus.

kivi:
> The environment variables end up in the `Option` in `sftp` eventually.
> 
> Setting environment variables decouples the config nicely which is a big plus.

I don't agree that's coupling with the Env function you have. If you follow the comment from @milahu above, using multiple `key_file`'s would be impossible. Or any Options could not be set multiple, except changing the Env-handling of `rclone`.

I mean I could commit that could we can live with that for now. But I can not call this decoupling, imho... but certainly an improvement.

ncw:
> If you follow the comment from @milahu above, using multiple `key_file`'s would be impossible.

If the sftp backend allowed multiple key_files (which it doesn't) you can easily [specify them with environment variables](https://rclone.org/docs/#options-1):

> Options that can appear multiple times (type stringArray) are treated slighly differently as environment variables can only be defined once. In order to allow a simple mechanism for adding one or many items, the input is treated as a [CSV encoded](https://godoc.org/encoding/csv) string. For example

kivi:
> If the sftp backend allowed multiple key_files (which it doesn't) you can easily [specify them with environment variables](https://rclone.org/docs/#options-1):


I am aware of that. So that's why it's better to split this feature up. The current PR works, let me know if you are fine with that so I clean it up.

@milahu I would suggest creating a new issue requiring `rclone` to support multiple KeyFiles respectively IdentityFiles.

milahu:
> @milahu I would suggest creating a new issue requiring `rclone` to support multiple KeyFiles respectively IdentityFiles.

please help yourself : )

obviously this would be required for full support of the ssh config format
and this should be easy to implement in rclone

kivi:
> > @milahu I would suggest creating a new issue requiring `rclone` to support multiple KeyFiles respectively IdentityFiles.
> 
> please help yourself : )
> 
> obviously this would be required for full support of the ssh config format and this should be easy to implement in rclone

Sure, I could do that create that ticket. I would then copy your words from this issue and create a new one. ;)

fs/config/configfile/configfile_ssh.go
kivi: @milahu I would split that as well into another issue. So one for multiple `IndentityFile`'s and another one for default values. (question, can you think about more default values, that might be interesting to use with rclone?)

Again, @ncw this another reason I think parsing the ssh config into Env does not for me like decoupled enough... even though this would work. Now for getting this kind of feature would imply to parse all given ssh configs and default IdentityFile's into rclone env, without actually the need of all config that has been loaded into Envs.

kivi:
@ncw I was wondering how to document this. I added a new flag `--use-ssh-config` in `config/configflags/configflags.go`, which I hope was the right place to do so. And there is new directory `lib/sshconfig/use_sshconfig[_test].go`

Other flags are appearing multiple time in the Manual.

buengese:
Hey, I think this feature is a great idea in principle but I think and might be a bit more complicated than what is implemented so far.

First of all I'm personally not a fan of turning the ssh config into environment variables and than parsing these again. While this results in a nice decoupling it feels like an ugly hack to me. I think it would be great if we had a way to dynamically hook into the config system and load additional config data sources tough I haven't really thought about how to implement this.

The more important issue I see here is that sshconfig `Host` directives are pattern matching and naively generating one backend from each `Host` and it's nodes will not work for more complicated ssh config files.

kivi:
Hello @buengese,

Thanks for joining. I am also not happy with turning ssh config into env vars. Contrary to your arguments, I think it is coupling, but not ugly. First of all, it's a new feature, so there's no need to decouple. But we can try to avoid unnecessary dependencies. An example would be if extending rclone to process multiple `key_file`'s would result in an env var with a comma-separated list of `identity_files`'s from ssh config.

What do you mean by more complicated than what is currently implemented? 

What do you mean by naively generating a backend?

By the way, in my first commits I tired to load ssh config into internal configfile of rclone, but that makes it to write the loaded ssh config; into rclone's config. Which is what we don't want.

I tried to see if I could just add it to the sftp backend. But the way config's are loaded into sftp backend looks a bit magic to me. I think I have explained this in previous comments.

buengese:
Sorry I might have been a bit unclear so I'll try to make an example. What I mean with naive parsing is that you create one remote per host entry in ssh config, and set it's configuration parameters using the attributes of this ssh host entry. This will work in simple cases. If you have one exact matching pattern for each host entry and define all relevant parameters on that host entry.

i.e.

```
Host one
  HostName 127.1.1.20
  User onerclone
  PreferredAuthentications publickey
  IdentityFile ~/.ssh/id_ed123
Host tworclone 
  HostName 127.1.1.40
  User vik
  LocalForward localhost:8090 127.0.0.1:8190
  IdentityFile ~/.ssh/id_rsa
  ```
  
  This will create the remotes `one` and `tworclone and both of them will work. 
  But my ssh config file looks more like this. 

```
Host laptop
	HostName 10.188.67.21
	User bngs
	ForwardAgent yes

Host switchdmz.s switchdmz
        User admin
	KexAlgorithms diffie-hellman-group1-sha1
	HostKeyAlgorithms ssh-rsa
	Ciphers aes256-cbc

Host ping.sp
	User root

Host vm
	HostName 10.211.55.11
	User parallels

Host work
	HostName redacted
        ProxyJump redacted
	User bngs

Host *.bngs.io *.bngs.xyz
        User root
        ForwardAgent yes
        CertificateFile ~/.ssh/all.pem
        IdentityFile ~/.ssh/signed.pub

Host *.sp
        HostName %hline.de
        ForwardAgent true
        ProxyJump bngs@fob.s

Host *.s
        HostName %hpline.de
        ForwardAgent true

Host *
        IdentityAgent /Users/bngs/Library/Containers/com.maxgoedjen.Secretive.SecretAgent/Data/socket.ssh
```

The current way of handling this file will generate a bunch of useless remotes. 

```
laptop:
ping.sp:
vm:
work:
*.s:
centdeck:
switchraum.s:
*.bngs.io:
*.sp:
*:
```

And none of them would work because all of them require multiple entries to be parsed.

**Comment by kivi** [comment]
@buengese Thanks for clarifying. I agree, this is could be improved a lot. Also, I was initially thinking to implement it into the `sftp backend`.  The parsing is done with `github.com/kevinburke/ssh_config`, which means, the parser should work fine. The mapper is "naive" and therefore the result seem to be "ugly" because it strips away all the richness config of your ssh config. 

Well, ideally it is done in `backend sftp` or let's say it is triggered, where we already know the target and/or source `host` and we can it load into the rclone config's from a new LoadSsshConfig(host) (without overiting and extending the rclone config file).

How we landed here...

When I tried to extend the config, it worked fine, but then unfortunately it magically extends your ~/.config/rclone/rclone.conf - and once it's overwritten that, but you change your ~/.ssh/config it would just ignore it. Then you would need some kind of kind of force flag (--use-sshConfig-force) but that also a very bad UX and it would confuse users.

If I knew how to inject the right piece ssh config into rclone config, without automagically extending the rclone config file, then I would fix it. Just couldn't figure it out yet.

Looks like @ncw is happier if it is done with ENV. 


@buengese, any idea? I am happy to rewrite and looking forward to use my ssh configs in a full extend with rclone.

**Comment by ncw** [comment]
@kivi
> I was wondering how to document this. I added a new flag `--use-ssh-config` in `config/configflags/configflags.go`, which I hope was the right place to do so. And there is new directory `lib/sshconfig/use_sshconfig[_test].go`

That makes it a global flag which I think is correct.

> Other flags are appearing multiple time in the Manual.

The only bit of manual you need to edit is `docs/content/docs.md` and put a section for the flag in there.

@buengese wrote:
> First of all I'm personally not a fan of turning the ssh config into environment variables and than parsing these again. While this results in a nice decoupling it feels like an ugly hack to me. I think it would be great if we had a way to dynamically hook into the config system and load additional config data sources tough I haven't really thought about how to implement this.

I don't particularly like the env vars either, but I wanted to stop this PR turning into a re-write the config system PR!

The right place to put the config is here

https://github.com/rclone/rclone/blob/75257fc9cdcaba71aba29e733c75c253fcf1540d/fs/configmap.go#L106

The `config.AddGetter` calls add the config sources as documented here: https://rclone.org/docs/#precedence

This will still potentially save config to the config file but it will always be overridden by the sshConfig fetcher.

This could also solve @buengese problem with the wildcards - you only ever generate the config for remotes the user actually requests and you can resolve the wildcards at that point.

If you want to try that @kivi you'll need to implement the `configmap.Getter` interface

https://github.com/rclone/rclone/blob/75257fc9cdcaba71aba29e733c75c253fcf1540d/fs/config/configmap/configmap.go#L24-L30

I would initialise (load the config) in this function if needed

https://github.com/rclone/rclone/blob/75257fc9cdcaba71aba29e733c75c253fcf1540d/fs/config.go#L652

And do something like this

```diff
@@ -119,6 +119,11 @@ func ConfigMap(prefix string, options Options, configName string, connectionStri
 		config.AddGetter(&regInfoValues{options, false}, configmap.PriorityNormal)
 	}
 
+	// ssh config specific variables
+	if configName != "" {
+		config.AddGetter(sshconfig.New(configName), configmap.PriorityNormal)
+	}
+
 	// remote specific environment vars
 	if configName != "" {
 		config.AddGetter(configEnvVars(configName), configmap.PriorityNormal)
```

Your `sshconfig.New` function should return the `configmap.Getter` - but if it hasn't been initialised then it should always return `false`.

kivi:
@ncw thanks.

```sh
fs/config/configflags/configflags.go:39:2: var-naming: var useSshConfig should be useSSHConfig (revive)
	useSshConfig    bool
...

lib/sshconfig/use_sshconfig.go:65:6: var-naming: func mapSshToRcloneConfig should be mapSSHToRcloneConfig (revive)
func mapSshToRcloneConfig(r io.Reader) (*sshConfig, error) {

```

Do you know why I need to rename the variable names? Do I need to use capitcal letters for Ssh? This error is shown, when it runs `golangci-lint`. So it will on github actions as well.

ncw:
Yes SSH needs to be in caps as it is an abbreviation. Though it isn't!

That linter is possibly a bit too fussy, but go with the flow!

kivi:
Oh, for some reason I have missed your comment. 

> If you want to try that @kivi you'll need to implement the `configmap.Getter` interface 

Yes, the example you are pointing out in your comment above. I did already and it works. To enable it, when `--use-ssh-config` flag is given, I would need to pass the flag down or somehow decroate the configmap. The `configmap` function has already many conditions and acts depending on the arguments you pass. And here it is even called twice in the same function 

https://github.com/rclone/rclone/blob/75257fc9cdcaba71aba29e733c75c253fcf1540d/fs/newfs.go#L80

`ConfigFs()` does one time `ConfigMap()` directly and calls `ParseRemote()`

```go
func ConfigFs(path string) (fsInfo *RegInfo, configName, fsPath string, config *configmap.Map, err error) {
	// Parse the remote path
	fsInfo, configName, fsPath, connectionStringConfig, err := ParseRemote(path)
	if err != nil {
		return
	}
	config = ConfigMap(fsInfo.Prefix, fsInfo.Options, configName, connectionStringConfig)
	return
}
```


`ConfigMap()` is also called in `ParseRemote()` 

```go
func ParseRemote(path string) (fsInfo *RegInfo, configName, fsPath string, connectionStringConfig configmap.Simple, err error) {
	parsed, err := fspath.Parse(path)
	if err != nil {
		return nil, "", "", nil, err
	}
	configName, fsPath = parsed.Name, parsed.Path
	var fsName string
	var ok bool
	if configName != "" {
		if strings.HasPrefix(configName, ":") {
			fsName = configName[1:]
		} else {
			m := ConfigMap("", nil, configName, parsed.Config)
			fsName, ok = m.Get("type")
			if !ok {
				return nil, "", "", nil, ErrorNotFoundInConfigFile
			}
```


I was debugging and configmap was called on different occasions. Through `fs.cache` without cache and I could not figure out how to pass down the flag `--use-ssh-config`. 

That's why I commented what I think about the configmap. But as I said, maybe we finish this with env first and rework the configmap maybe for intentions like being able to use multiple `key_file` or even a default key_file without breaking functionality.

kivi:
> Yes SSH needs to be in caps as it is an abbreviation. Though it isn't!
> 
> That linter is possibly a bit too fussy, but go with the flow!

ok. I think that's it.
This can be merged into main.

After merge, I would go on with configmap and then later enable multiple identiyfiles/key-files if  that is popular.

kivi:
Hi @ncw, are you going to merge this?

CaptaiNiveau:
Hey! I'd really like to have this feature in rclone. I have a fairly extensive ssh config and this would make it much more seamless to use. It'll be even nicer once I start using smallstep, since that automatically manages the ssh config file, making this feature even more necessary.

What's blocking this from being merged (besides the conflicts that came up since this was drafted)?

PR #8105, opened 2024-09-26
crypt: Implement V2 cipher version with MD5 hash support #2

tomekit:
This is continuation of previous pull request: https://github.com/rclone/rclone/pull/8095 based from: https://github.com/tomekit/rclone/tree/envelope_encryption_7192_7 branch with rebased/cleaned up commit messages. 

Hi @ncw, I've thought it will be good idea to clean up previous commits and prepare something that's cleaner to merge into Rclone codebase at some point. Just because I've wanted to leave older branch untouched with its history (no force commits), I've left it as it is for to not break your review workflow.

===

Cipher improvements and MD5 hash support to crypt.
Commit description: "In V2 version each file encrypted using separate CEK (wrapped with AES Keywrap RFC3394). Truncation protection uses last byte of nonce to indicate last encryption block. Added MD5 hash support for crypt.go (only if V2 cipher is used), MD5 of plaintext is encrypted and authenticated at the end of file. Added 4 reserved bytes at the beginning for future use. Added 1 hash type byte in the footer to allow upgrade to different than MD5 hash. cipher_version setting determines hash version used for writes. For reads we determine cipher version by reading header magic bytes, that's regardless of cipher_version setting. A Added TestNewEncrypterV2 test to demonstrate the outputs and make sure that outputs match the expected hardcoded values. Modified other tests to satisfy newly added params (e.g. cek). Related to: https://github.com/rclone/rclone/issues/7192"

https://github.com/rclone/rclone/issues/7192

#### Checklist

### DIFF
--- a/backend/crypt/cipher.go
+++ b/backend/crypt/cipher.go
@@ -1,6 +1,7 @@
 package crypt
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/aes"
@@ -18,9 +19,11 @@ import (
 	"unicode/utf8"
 
 	"github.com/Max-Sum/base32768"
+	keywrap "github.com/nickball/go-aes-key-wrap"
 	"github.com/rclone/rclone/backend/crypt/pkcs7"
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
+	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/lib/readers"
 	"github.com/rclone/rclone/lib/version"
 	"github.com/rfjakob/eme"
@@ -35,9 +38,20 @@ const (
 	fileMagicSize       = len(fileMagic)
 	fileNonceSize       = 24
 	fileHeaderSize      = fileMagicSize + fileNonceSize
-	blockHeaderSize     = secretbox.Overhead
-	blockDataSize       = 64 * 1024
-	blockSize           = blockHeaderSize + blockDataSize
+	fileEncryptedSuffix = ".bin" // V1 cipher by default sets the suffix (only if name encryption disabled), but user can disable it completely by setting it to `none`.
+
+	fileMagicV2             = "RCLONE\x00\x01"
+	fileMagicSizeV2         = len(fileMagicV2)
+	fileNonceSizeV2         = 23
+	fileReservedBytesV2Size = 4 // Make sure it matches the: `fileReservedBytesV2`
+	fileHeaderSizeV2        = fileMagicSizeV2 + fileNonceSizeV2 + fileWrappedCekSize + fileReservedBytesV2Size
+	fileCekSize             = 32
+	fileWrappedCekSize      = fileCekSize + 8 // Some overhead of AES RFC 3394 Key Wrapping
+
+	blockHeaderSize      = secretbox.Overhead
+	blockDataSize        = 64 * 1024
+	blockSize            = blockHeaderSize + blockDataSize
+	lastBlockFlag   byte = 1 // Corresponds to: `last_block` https://github.com/miscreant/meta/wiki/STREAM
 )
 
 // Errors returned by cipher
@@ -47,8 +61,9 @@ var (
 	ErrorNotAMultipleOfBlocksize = errors.New("not a multiple of blocksize")
 	ErrorTooShortAfterDecode     = errors.New("too short after base32 decode")
 	ErrorTooLongAfterDecode      = errors.New("too long after base32 decode")
-	ErrorEncryptedFileTooShort   = errors.New("file is too short to be encrypted")
+	ErrorEncryptedFileTooShort   = errors.New("file is too short to be decrypted")
 	ErrorEncryptedFileBadHeader  = errors.New("file has truncated block header")
+	ErrorEncryptedCekInvalid     = errors.New("file encryption key is invalid")
 	ErrorEncryptedBadMagic       = errors.New("not an encrypted file - bad magic string")
 	ErrorEncryptedBadBlock       = errors.New("failed to authenticate decrypted block - bad password?")
 	ErrorBadBase32Encoding       = errors.New("bad base32 filename encoding")
@@ -62,7 +77,9 @@ var (
 
 // Global variables
 var (
-	fileMagicBytes = []byte(fileMagic)
+	fileMagicBytes      = []byte(fileMagic)
+	fileMagicBytesV2    = []byte(fileMagicV2)
+	fileReservedBytesV2 = []byte{BlockCipherXSalsa20, 0x00, 0x00, 0x00} // Make sure it matches the: `fileReservedBytesV2Size`
 )
 
 // ReadSeekCloser is the interface of the read handles
@@ -84,6 +101,18 @@ const (
 	NameEncryptionOff NameEncryptionMode = iota
 	NameEncryptionStandard
 	NameEncryptionObfuscated
+)
+
+// CipherVersion
+const (
+	CipherVersionV1 = "1"
+	CipherVersionV2 = "2"
+)
+
+// BlockCipher
+const (
+	BlockCipherXSalsa20 = 0x01 // Used currently
+	BlockCipherAESGCM   = 0x02 // Not used, possible future use (increase web performance): https://github.com/rclone/rclone/issues/7192#issuecomment-2353617267
 )
 
 // NewNameEncryptionMode turns a string into a NameEncryptionMode
@@ -181,16 +210,18 @@ type Cipher struct {
 	dirNameEncrypt  bool
 	passBadBlocks   bool // if set passed bad blocks as zeroed blocks
 	encryptedSuffix string
+	version         string
+	exactSize       bool
 }
 
 // newCipher initialises the cipher.  If salt is "" then it uses a built in salt val
-func newCipher(mode NameEncryptionMode, password, salt string, dirNameEncrypt bool, enc fileNameEncoding) (*Cipher, error) {
+func newCipher(mode NameEncryptionMode, password, salt string, dirNameEncrypt bool, enc fileNameEncoding, version string) (*Cipher, error) {
 	c := &Cipher{
-		mode:            mode,
-		fileNameEnc:     enc,
-		cryptoRand:      rand.Reader,
-		dirNameEncrypt:  dirNameEncrypt,
-		encryptedSuffix: ".bin",
+		mode:           mode,
+		fileNameEnc:    enc,
+		cryptoRand:     rand.Reader,
+		dirNameEncrypt: dirNameEncrypt,
+		version:        version,
 	}
 	c.buffers.New = func() any {
 		return new([blockSize]byte)
@@ -218,6 +249,49 @@ func (c *Cipher) setEncryptedSuffix(suffix string) {
 // Call to set bad block pass through
 func (c *Cipher) setPassBadBlocks(passBadBlocks bool) {
 	c.passBadBlocks = passBadBlocks
+}
+
+func (c *Cipher) setCipherVersion(cipherVersion string) {
+	c.version = cipherVersion
+}
+
+func (c *Cipher) setExactSize(exact bool) {
+	c.exactSize = exact
+}
+
+func getFileHeaderSize(cipherVersion string) int {
+	if cipherVersion == CipherVersionV2 {
+		return fileHeaderSizeV2
+	}
+	return fileHeaderSize
+}
+
+func getFileFooterSize(cipherVersion string) int {
+	if cipherVersion == CipherVersionV2 {
+		return int(HashEncryptedSizeWithHeader)
+	}
+	return 0
+}
+
+func getFileMagicSize(cipherVersion string) int {
+	if cipherVersion == CipherVersionV2 {
+		return fileMagicSizeV2
+	}
+	return fileMagicSize
+}
+
+func getFileMagicBytes(cipherVersion string) []byte {
+	if cipherVersion == CipherVersionV2 {
+		return fileMagicBytesV2
+	}
+	return fileMagicBytes
+}
+
+func getFileNonceSize(cipherVersion string) int { // Effective nonce size. Both V1 and V2 use 24 bytes, but V2 consists 23+1 bytes nonce, where last byte is calculated dynamically depending if the processing block is last (truncation protection)
+	if cipherVersion == CipherVersionV2 {
+		return fileNonceSizeV2
+	}
+	return fileNonceSize
 }
 
 // Key creates all the internal keys from the password passed in using
@@ -619,6 +693,8 @@ func (c *Cipher) NameEncryptionMode() NameEncryptionMode {
 
 // nonce is an NACL secretbox nonce
 type nonce [fileNonceSize]byte
+type cek [fileCekSize]byte
+type wrappedCek [fileWrappedCekSize]byte
 
 // pointer returns the nonce as a *[24]byte for secretbox
 func (n *nonce) pointer() *[fileNonceSize]byte {
@@ -627,18 +703,33 @@ func (n *nonce) pointer() *[fileNonceSize]byte {
 
 // fromReader fills the nonce from an io.Reader - normally the OSes
 // crypto random number generator
-func (n *nonce) fromReader(in io.Reader) error {
-	read, err := readers.ReadFill(in, (*n)[:])
-	if read != fileNonceSize {
+func (n *nonce) fromReader(in io.Reader, expectedNonceSize int) error {
+	read, err := readers.ReadFill(in, (*n)[:expectedNonceSize])
+	if read != expectedNonceSize {
 		return fmt.Errorf("short read of nonce: %w", err)
 	}
 	return nil
 }
 
-// fromBuf fills the nonce from the buffer passed in
-func (n *nonce) fromBuf(buf []byte) {
+func (n *cek) fromReader(in io.Reader) error {
+	read, err := readers.ReadFill(in, (*n)[:])
+	if read != fileCekSize {
+		return fmt.Errorf("short read of CEK: %w", err)
+	}
+	return nil
+}
+
+func (n *wrappedCek) fromBuf(buf []byte) {
 	read := copy((*n)[:], buf)
-	if read != fileNonceSize {
+	if read != fileWrappedCekSize {
+		panic("buffer to short to read wrapped CEK")
+	}
+}
+
+// fromBuf fills the nonce from the buffer passed in
+func (n *nonce) fromBuf(buf []byte, expectedNonceSize int) {
+	read := copy((*n)[:], buf)
+	if read != expectedNonceSize {
 		panic("buffer to short to read nonce")
 	}
 }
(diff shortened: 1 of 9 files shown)


tomekit:
I've modified: `crypt.md` to add information about V2 cipher:
https://github.com/rclone/rclone/pull/8105/commits/0b5a5fc0447e99f13de1a13d2ed3f4e98cc05f84

TODOs for myself:
- Prepare documentation about data migration,
- Mention about decrypted size issues for V1 objects.

tomekit:
I was trying to prepare some docs about migration and was playing with `copy` command myself. I've faced couple issues and I am not exactly sure what would be the best way forwards and whether dedicated migration command is required.

My first migration approach was to set up two crypt remotes with exactly same config, except `cipher_version` which would be `1` for `source_crypt` and `2` for `dest_crypt`.

I've added single 1 byte file: `old/test.txt` and then executed: `copy -vvv source_crypt: dest_crypt:`. 
It worked fine migrating object to V2 format:
```
2024/09/28 13:14:34 DEBUG : rclone: Version "v1.69.0-DEV" starting with parameters ["/home/tomek/.cache/JetBrains/GoLand2024.2/tmp/GoLand/___copy_source_crypt__dest_crypt_" "copy" "-vvv" "--checksum" "source_crypt:" "dest_crypt:"]
2024/09/28 13:14:34 DEBUG : Creating backend with remote "source_crypt:"
2024/09/28 13:14:34 DEBUG : Using config file from "/home/tomek/.config/rclone/rclone.conf"
2024/09/28 13:14:34 DEBUG : Creating backend with remote "s3drive_auto_hYyoJhP3mTCBDMXB4BBtE3NmbfgElRFgFiM5bGFdXBk:backendtest1"
2024/09/28 13:14:34 DEBUG : Creating backend with remote "dest_crypt:"
2024/09/28 13:14:34 DEBUG : old/test.txt: Bad size for decrypt: file has truncated block header
2024/09/28 13:14:34 DEBUG : Encrypted drive 'dest_crypt:': Waiting for checks to finish
2024/09/28 13:14:34 DEBUG : old/test.txt: Bad size for decrypt: file has truncated block header
2024/09/28 13:14:34 DEBUG : old/test.txt: Bad size for decrypt: file has truncated block header
2024/09/28 13:14:34 DEBUG : old/test.txt: Bad size for decrypt: file has truncated block header
2024/09/28 13:14:34 DEBUG : old/test.txt: Sizes differ (src 1 vs dst 0)
2024/09/28 13:14:34 DEBUG : Encrypted drive 'dest_crypt:': Waiting for transfers to finish
2024/09/28 13:14:35 DEBUG : old/test.txt: md5 = d5eb4226dd830c666eba0ef13e14c5b1 OK
2024/09/28 13:14:35 INFO  : old/test.txt: Copied (replaced existing)
2024/09/28 13:14:35 INFO  : 
Transferred:              1 B / 1 B, 100%, 0 B/s, ETA -
Elapsed time:         1.6s
```

The issue is that after retry, it will copy that file again:
```
2024/09/28 13:16:33 DEBUG : old/test.txt: Sizes differ (src 77 vs dst 1)
```
since size derivation is based on config assumption (`cipher_version`), not per individual object (this would require HTTP call for every object).

Adding: `--ignore-size` to the copy command wouldn't really help, as in that case migration wouldn't copy objects, as mod date on both ends would match just fine.

... so there is no easy way for `rclone copy` to "retry" migration from last known point. This reduce chances of success and might create duplicate versions if remote is versioned (e.g. S3 with versioning enabled).

==

Additionally there maybe a serious issue with current dynamic call to `setCipherVersion`. During migration, using `copy` operation between same remotes this may lead to unintended data loss when `verify` method kicks in and throws `corrupted on transfer` and then: `Removing failed copy`.

I've got reproducible steps and whilst I am not 100% sure about the exact cause I presume that's consequence of parallel execution and dynamic call to `setCipherVersion`. If two objects (having different cipher version) are processed in parallel, then: `setCipherVersion` call might change the overall context.

Given object A (V2) and object B(V1), Rclone during execution will start with assumed version (based on config), then whichever per-object operation calls: `setCipherVersion` will make Rclone switch to that version and affect all future operations, until: `setCipherVersion` is called again. In some cases this will lead to `decryptedSize`  returning different values before copy and after copy (verification) leading to "data corruption" which would delete that from destination remote. If destination remote is the same as source remote (my migration approach), then this leads to data loss.

I think we may need to:
a) Provide reliable migration mechanism, either by modifying copy and implementing explicit cipher detection (querying magic bytes) or find some other command + combination of configuration flags which can satisfy this requirement.

b) Get rid of: `setCipherVersion` detection in a current unsafe form. We should explicitly follow `cipher_version` in the config and provide strict behavior.

Given the a) and b), user would be forced to migrate their data before they can switch `cipher_version`.

The alternative to forced migration would be implementing Cipher version on the object level and whenever there is a cipher version conditional we would then workout what to do based on combination of Rclone configuration and object level cipher version. Perhaps we could abuse: `meta     fs.Metadata` field in: `StaticObjectInfo` to sneak cipher version through without making the non-crypt abstractions being aware of crypt.

Do you have any thoughts on that?

ncw:
> since size derivation is based on config assumption (`cipher_version`), not per individual object (this would require HTTP call for every object)

Ideas

1. Perhaps the easiest thing to do might be to make another flag which does that extra read on every file. Say `accurate_v1_size` and set that during the migration? This might be useful in other scenarios too.

2. We could also use the file name to indicate v1 vs v2. v1 files generally have the `.bin` extension (not always though) so we could use `.2.bin` or something like that for v2 files.

3. We could erase the size on the objects until we have read them. This means returning `-1` from `Size()` until we know for definite whether it is a v1 object or v2 object. Some of the backends work like this. This probably isn't a lot better than `--ignore-size` though.

tomekit:
> Perhaps the easiest thing to do might be to make another flag which does that extra read on every file. Say accurate_v1_size and set that during the migration? This might be useful in other scenarios too.

That's what I did by adding: `exact_size` (happy to rename it though). 
https://github.com/rclone/rclone/pull/8105/commits/0538517cee791116275c32096e02752069ab5ea7

and some basic migration documentation: https://github.com/rclone/rclone/pull/8105/commits/0538517cee791116275c32096e02752069ab5ea7#diff-b9ee403b94441fb6df9e132c4164e0d8e72cf1b2156358392408810a5da6043f

I haven't mentioned it here just yet, as I need more time testing it more thoroughly. 

> We could also use the file name to indicate v1 vs v2. v1 files generally have the .bin extension (not always though) so we could use .2.bin or something like that for v2 files.

I wouldn't rely on `.bin` extension being an exact proof of V1 version, however I like the idea of trying to amend the naming convention to reliably detect the V2 version (without HTTP call involved). Do you think that we could add e.g.: `2.bin` suffix regardless if filepath encryption is enabled or not? As far I am concerned at the moment the `.bin` suffix gets added for V1 only if name_encryption is disabled.
Given that we would be able to 100% reliably detect the V2 version, we could then assume that if version isn't present then it must be V1.

tomekit:
I've implemented additional approach (.2.bin suffix), but have some decision paralysis as to which direction is better.

In this pull request, new cipher version is stored only in the header (magic bytes).
Even though I wouldn't recommend to migrate files in-place. It's technically possible, by using: `--no-check-dest` flag which will "force" copy objects. Without this flag Rclone wouldn't copy as modtime/size would be the same. 
In that case both: `source_crypt` and `dest_crypt` are the same and point to the same location, except that `dest_crypt` has: `cipher_version = 2` set.
```
copy -vvv --crypt-exact-size --no-check-dest source_crypt: dest_crypt:
```

It's also possible to skip the above flags and simply use:
```
copy -vvv source_crypt: dest_crypt:
```

however it may throw some warnings (for small files): 
```
2024/10/08 12:33:08 DEBUG : old/test.txt: Bad size for decrypt: file is too short to be decrypted
2024/10/08 12:33:08 DEBUG : old/test.txt: Bad size for decrypt: file is too short to be decrypted
2024/10/08 12:33:08 DEBUG : old/test.txt: Bad size for decrypt: file is too short to be decrypted
2024/10/08 12:33:08 DEBUG : old/test.txt: Bad size for decrypt: file is too short to be decrypted
```
and it will always repeat transfer (even for migrated items), since calculated size on both sides (for same file) will differ.

E.g. 1 byte file V1 to V2 migration shows:
```
2024/10/08 12:33:08 DEBUG : old/test.txt: Sizes differ (src 1 vs dst 0)
```
It can be seen, that on the src side size is displayed correctly, on the dst side it would be negative, so Rclone uses `0`.

subsequent run for already migrated file will show:
```
2024/10/08 12:35:48 DEBUG : old/test.txt: Sizes differ (src 77 vs dst 1)
```
Now it's the `dst` side showing correct size for V2 object, however `src` assumes V1 therefore showing incorrect value.

### .2.bin suffix

In a different branch (https://github.com/tomekit/rclone/commit/5b637f09cf9126fd6cfdab14a0704dc08ab0ee55) I've experimentally implemented approach where we always append the: `.2.bin` suffix to the V2 objects.

It assumes that everything that has: `.2.bin` suffix is V2, everything else is V1. I've removed: `exact_size` as it's no longer needed (correct size can be derived from object path).

I've then introduced: `--force_migration_crypt_v1`, so when `copy` is used to migrate objects from V1 to V2 it skips `modtime` and `size` checks as long as the source object is V1.

The consequence is that in-place (that is within same remote) migration isn't quite possible, as it would create duplicates.

For instance, V1 object: `FABcDX4weka4enCoMNvtQw/-KEliBza8NvE1KdiwJL3GA` would be copied to: `
FABcDX4weka4enCoMNvtQw/-KEliBza8NvE1KdiwJL3GA.2.bin`. In that case  `rclone ls` would be able to decrypt both filepaths to e.g.: `folder/test.txt` showing two duplicates.

I don't think that no in-place migration is a huge issue though, I probably wouldn't recommend it anyway due to safety issues (if something goes wrong, e.g. verify method spots the problem and then deletes the `dst`, but since `dst` is pointing to same location as `src` it would mean user data was deleted. Technically we could add flag to make `verify` only throw some warning instead of doing hard delete, but I am not sure if this is the right path anyway).

Another change is that: `suffix` parameter is no longer supported/functional. Simply speaking if `cipher_version` is set to `2` it would always set: `suffix` internally to: `.2.bin`.

==
I would appreciate if you could let me know if you prefer the approach with the: `.2.bin` suffix or without.

tomekit:
I've started adding more tests: https://github.com/rclone/rclone/commit/75a5a534864dcdf65abe2f272453f0d760f1b8aa

I've then realized that hash wrapping logic: `wrapReaderCalculatePlaintextHash` and `wrapReaderAppendPlaintextHash` had to be repeated in couple functions and tests, so instead I've placed it inside the: `encryptData` (https://github.com/rclone/rclone/commit/34a77f687a968fa4bebcde1bac94a82a9cc4e6ff) and in subsequent commits modified all calls, so it uses `encryptData` directly instead of creating the `newEncrypter` every time, duplicating the code and reducing testability.

I've decided to push on top of: `envelope_encryption_7192_8_suffix` branch which keeps the approach with `.2.bin` suffix added to the encrypted filepath. Unless there is any issue with the increased length of the filepath I think I've liked this approach better.
If by any chance we decide to not use V2 specific suffix, then I still can "backport" these changes onto: `envelope_encryption_7192_8` branch.

I don't really want to rush any decisions here, but once we know which direction (suffix or not) to go, then I could then polish things, add more tests, update documentation according to the chosen approach... and hopefully submit for a final review.

dzoep:
Hello,

as a part of improving data confidentiality for one big academic
provider of object storage for research, we are working on extending
the rclone crypt remote with full key management capabilities.

The approach proposed in this (and previous) PR and in the issue #7192
is a solid basis for both a rich-featured key management system (KMS)
and the added benefit of the ability to share single object/file is
great. CEK forms the necessary low-level basis for any advanced
functionality.

However, for any kind of KMS, the ability to re-encrypt CEKs with
different master key is important. With the current design, this would
require overwriting the object header, which - I am afraid - will not
play well with for example S3 versioning enabled.

It would be great, if the final design could separate the object
meta-data and object itself into two objects. Implementing this
naively would probably introduce some problems on its own, so I
understand we need to design it carefully, however, I think it is
worth the hassle and also I am going to spend quite some time on such
design anyway.

As with original crypt remote for which we have performed an in-depth
cryptanalysis, I am already looking into the proposed usage of the AES
Keywrap mechanism and into the proposed hash support. With the latter,
I think the required functionality lies in providing the user with
full AEAD (authenticated encryption with associated data) guarantees
for the whole encrypted objects - like the secretbox provides for
individual chunks.

Anyway, thank you @tomekit for all the good work - I have to repeat
it: this is a very solid basis for some quite important features the
users typically need.


Dominik

tomekit:
Thanks for your feedback and kind words!

> It would be great, if the final design could separate the object meta-data and object itself into two objects.
> Implementing this naively would probably introduce some problems on its own

I was considering this approach, but got scared by different consistency guarantees of different clouds and back-ends, after all Rclone aims to be compatible with most/all of them.

I've assumed that making sure that metadata file is consistent with raw data file is not an easy task. If by any chance raw file update succeeds, but metadata write files, what would happen? Would user end up with undecryptable blob (dataloss)?
Just because of this problem, additional metadata files may pose significant overhead, at least additional PUT and additional verification call (at worst some additional copy/delete operations to make sure that data updates stays safe and atomic) which may affect performance and final usage pricing of per-request billed storages like S3.

> the ability to re-encrypt CEKs with different master key is important. With the current design, this would require overwriting the object header

In our storage product ([S3Drive](https://s3drive.app/)) we plan to support multiple master keys and re-encrypt CEKs (for the purposes of sharing files) by building the metadata storage layer ourselves on top of Rclone crypt. In this layer we can keep multiple information associated with the object. 
We've went that way, as by having control of this layer we can provide certain guarantees that would be hard/impossible to achieve with the native metadata approach with Rclone.
The issue is some complexity, because now data layer and metadata layer needs to be maintained separately, but we're aware of that.

The above could be replicated natively with Rclone using separate metadata file (in addition to object encryption header store in file contents). 

Unfortunately with this approach it's still not possible to implement the main master key invalidation (as it's stored as part of the file header) without modifying the whole object. Given that decrypting resource requires at least two things: 
a) access to encrypted raw data (protected by e.g. API key or other factors),
b) decrypting file with a valid master key,

I would still assume it won't be a necessity very often, as user can always easily change just `a)` in order to "cut" the adversary out of the access.

...and if by any chance user MUST invalidate the master key, they could always migrate data (exact steps we need to find out, as in some cases inplace migration might be impossible/risky). 
In order to reduce burden related to key invalidation, user from start could use multiple `crypt` vaults to segregate data (e.g. by importance and/or size), so if one master key is leaked by any chance, then it would only affect some data not all of it.

> however, I think it is worth the hassle and also I am going to spend quite some time on such design anyway.

If you have some idea I would be happy to provide some comments (although I am far from being an expert here). 

I am also well aware that this all is getting somewhat big, complex and fragile. Even best proposal might just end up being "some nice Github discussion" if there is no resource to deliver. We (S3Drive) are happy to spend more time on this, although I am sure we won't be able to implement completely new concepts (storing metadata in a separate file is certainly a new one), so unless there is some brave contributor we would have to pass.

Given that Rclone crypt as such supports versioning (_with some caveats*_) perhaps we could release improvements in chunks (obviously we need to rule out as much design issues early, as changes are still costly).
If we aim too high from start we risk being stuck in a perfect solution which won't end up being implemented.

Sorry if this sounds overly negative, being on this topic for quite a while I am just trying to be realistic.

### Just an idea
How about wrapping master key, with yet another key that would be stored in the root of crypt vault?
In that approach user would need some encryption material (key or password) stored in their config, which would then be able to unwrap the master key stored in some file e.g. `.metadata`.
If user wanted to change their key/password, then only key inside `.metadata` would have to be rewrapped.
That's in principle how some privacy-friendly storage providers (implementing E2E encryption) allow password change in an e-mail/password accounts.

Would that help by any chance in the use case you have in mind?

The risk is that if `.metadata` file is lost in the cloud somehow, user loses ability to decrypt all of their vault dataset.
What I like with current Rclone approach is that it's rather robust and corruption is unlikely, as everything is "stored in one place".

_*related to encrypted/decrypted file size if we modify the storage layout - this wouldn't be an issue if we've decided to use version suffix as proposed above_

dzoep:
> Thanks for your feedback and kind words!
> 
> > It would be great, if the final design could separate the object meta-data and object itself into two objects.
> > Implementing this naively would probably introduce some problems on its own
> 
> I was considering this approach, but got scared by different consistency guarantees of different clouds and back-ends, after all Rclone aims to be compatible with most/all of them.

Yes, a consistent behavior across the multitude of rclone-supported storage providers is a major headache when thinking about this.

> 
> I've assumed that making sure that metadata file is consistent with raw data file is not an easy task. If by any chance raw file update succeeds, but metadata write files, what would happen? Would user end up with undecryptable blob (dataloss)? Just because of this problem, additional metadata files may pose significant overhead, at least additional PUT and additional verification call (at worst some additional copy/delete operations to make sure that data updates stays safe and atomic) which may affect performance and final usage pricing of per-request billed storages like S3.

All of this is true, however there have already been some ideas proposed. For example, as the meta-data is rather small, there may be a way to always maintain a single metadata object for multiple data objects with some stronger mechanism for assuring consistency (update metadata first, use native or custom versioning mechanism...). Other proposals include using object metadata, which can be (to some extent) updated for storing "CryptV2" metadata (including wrapped CEK). I do not have enough expertise to judge whether this can be reliably used with other back-ends other than S3.

> 
> > the ability to re-encrypt CEKs with different master key is important. With the current design, this would require overwriting the object header
> 
> In our storage product ([S3Drive](https://s3drive.app/)) we plan to support multiple master keys and re-encrypt CEKs (for the purposes of sharing files) by building the metadata storage layer ourselves on top of Rclone crypt. In this layer we can keep multiple information associated with the object. We've went that way, as by having control of this layer we can provide certain guarantees that would be hard/impossible to achieve with the native metadata approach with Rclone. The issue is some complexity, because now data layer and metadata layer needs to be maintained separately, but we're aware of that.
> 
> The above could be replicated natively with Rclone using separate metadata file (in addition to object encryption header store in file contents).
> 
> Unfortunately with this approach it's still not possible to implement the main master key invalidation (as it's stored as part of the file header) without modifying the whole object. Given that decrypting resource requires at least two things: a) access to encrypted raw data (protected by e.g. API key or other factors), b) decrypting file with a valid master key,
> 
> I would still assume it won't be a necessity very often, as user can always easily change just `a)` in order to "cut" the adversary out of the access.

I understand that from a non-cryptographic perspective this is perfectly valid. The issue here is when there are legal obligations to make stronger guarantee in case of raw data leaks. Which happens to be the case for many kinds of data especially within EU. 

> 
> ...and if by any chance user MUST invalidate the master key, they could always migrate data (exact steps we need to find out, as in some cases inplace migration might be impossible/risky). In order to reduce burden related to key invalidation, user from start could use multiple `crypt` vaults to segregate data (e.g. by importance and/or size), so if one master key is leaked by any chance, then it would only affect some data not all of it.

Yes, that is true when there is a limited number of users and more levels of confidentiality of the data. The "full KMS" solution is needed when it is the other way around. For example a large data set on which a team of hundreds is working and from time to time new members are added (easy) and some removed (this is the tough one). The key revocation feature does not guarantee that any data which the user once had access to would be inaccessible (from purely cryptographic perspective they should be treated like the have a copy), however on any newly uploaded data there should be guarantee that the user cannot read them without access to the new CEKs - typically via rotated master key. And re-encrypting the whole bucket every other week would be VERY expensive if it was necessary to re-upload all the objects again. I fully agree with your assessment that this is also about the final usage pricing - no matter how elegant the solution will be, if it will incur significant operations costs increase, people will (and rightly) complain.

> 
> > however, I think it is worth the hassle and also I am going to spend quite some time on such design anyway.
> 
> If you have some idea I would be happy to provide some comments (although I am far from being an expert here).
> 
> I am also well aware that this all is getting somewhat big, complex and fragile. Even best proposal might just end up being "some nice Github discussion" if there is no resource to deliver. We (S3Drive) are happy to spend more time on this, although I am sure we won't be able to implement completely new concepts (storing metadata in a separate file is certainly a new one), so unless there is some brave contributor we would have to pass.

As I wrote earlier, this PR is already a great contribution. I have to be careful with my words here but I can at least publiclz say that we have already allocated resources for getting it a bit further and I hope all the rclone users will benefit from this.

> 
> Given that Rclone crypt as such supports versioning (_with some caveats*_) perhaps we could release improvements in chunks (obviously we need to rule out as much design issues early, as changes are still costly). If we aim too high from start we risk being stuck in a perfect solution which won't end up being implemented.

I think this work will have to be addressed and released in chunks (a bit of pun intended) anyway.

> 
> Sorry if this sounds overly negative, being on this topic for quite a while I am just trying to be realistic.

Absolutely no worries, this is a complex issue and there is no silver bullet. We are involved since early spring and as a cryptologist on the project, I am trying since the beginning (and will continue to) keep realistic expectations from everyone.

> ### Just an idea
> 
> How about wrapping master key, with yet another key that would be stored in the root of crypt vault? In that approach user would need some encryption material (key or password) stored in their config, which would then be able to unwrap the master key stored in some file e.g. `.metadata`. If user wanted to change their key/password, then only key inside `.metadata` would have to be rewrapped. That's in principle how some privacy-friendly storage providers (implementing E2E encryption) allow password change in an e-mail/password accounts.

This would - sadly - not work. As the user can extract the key and keep it aside. It would be kind of security by obscurity which I cannot recommend.

However on a brighter note, I've looked into the RFC3394 usage here and although I was a bit skeptical at the beginning, the usage provides all the guarantees it should and therefore I concluded that using AES keywrap for CEKs is sound from cryptographic perspective. Good job there!

> 
> Would that help by any chance in the use case you have in mind?
> 
> The risk is that if `.metadata` file is lost in the cloud somehow, user loses ability to decrypt all of their vault dataset. What I like with current Rclone approach is that it's rather robust and corruption is unlikely, as everything is "stored in one place".
> 
> _*related to encrypted/decrypted file size if we modify the storage layout - this wouldn't be an issue if we've decided to use version suffix as proposed above_

Yes, it may actually handle certain (non-cryptographic) metadata as well.

I still need to look into the HMAC and effective nonce truncation (the last byte) before I jump to any conclusions or suggestions.

I hope this also does not sound too negative - it is definitely not meant that way!

tomekit:
> Yes, that is true when there is a limited number of users and more levels of confidentiality of the data. The "full KMS" solution is needed when it is the other way around. For example a large data set on which a team of hundreds is working and from time to time new members are added (easy) and some removed (this is the tough one). 

Does this need to be handled on the cryptographic layer? I would've assumed that if proper Access Control is implemented within the organization, then the user that we want to remove from org (e.g. past employee) would simply lose raw read access to encrypted file. Regardless of chosen cryptographic solution they wouldn't be able to access file anyway.

> I still need to look into the HMAC 
If that's any help for you, that's using: https://en.wikipedia.org/wiki/ChaCha20-Poly1305#Salsa20-Poly1305_and_XSalsa20-Poly1305 

In this PR I haven't modified this, however I've added hash of unencrypted file is stored in the file footer, which is then encrypted and authenticated using additional HMAC, also stored in the end.

> and effective nonce truncation (the last byte) 
I've followed the approach used here: https://github.com/miscreant/meta/wiki/STREAM 

and used some "advice from SO: https://crypto.stackexchange.com/questions/78466/which-is-the-best-algorithm-for-large-scale-file-encryption-other-than-aes/78471#78471
and https://crypto.stackexchange.com/a/106992/70896

Neustradamus:
About MD5, it is not secure:
- https://en.wikipedia.org/wiki/MD5#Security

tomekit:
Hi @ncw,

It's been a while since we've initially proposed this: https://github.com/rclone/rclone/issues/7192

We weren't able to spend more time on this back then to implement missing bits / docs / migration path, however I would be happy to spend more time finalizing it, so we can have it merged to mainline Rclone.

We're running the proposed V2 crypt with success in our project (S3Drive) since early 2024. 

In 2025 we've then applied two fixes related to decryption:
https://github.com/tomekit/rclone/commit/7ffdcc62aac8b3ca98368394385f0d6d95df1fa2
https://github.com/tomekit/rclone/commit/8ee971a1423aa186991c00b5902ac3383459fe44
which affected some specific file sizes (e.g. 65503). 

I've now squashed merge all changes against master and force pushed to branch being subject of this pull request: `envelope_encryption_7192_8`, so we can have clean state and carry on using same pull request ID.
I've backed up `envelope_encryption_7192_8` before force push as: https://github.com/tomekit/rclone/tree/envelope_encryption_7192_8_stash2 just in case.

Regarding crypt version migration path, I've had success migrating V1 to V2 by moving data from backend A to backend B, however once I've used in-place migration: https://github.com/rclone/rclone/pull/8105/changes#diff-b9ee403b94441fb6df9e132c4164e0d8e72cf1b2156358392408810a5da6043fR903
I've had mixed results and troubles, including data-loss. I honestly spent more than enough time back then to get it right. I can assume that either in-place migration is too tricky, or I am not capable of providing reliable migration path.
At this stage I can suggest not recommending in-place migration and only recommend migration to different back-end / folder.

By any chance, would you be able to have a glance at the proposed code and suggest what else is required? What test cases would you add on top of proposed in this PR?

Your help is very much appreciated here!

PR #8117, opened 2024-10-06
rc/job:  use mutex for adding listeners thread safety

hayden-pan:
Fix in extreme cases, when the job is executing finish(), the listener added by calling OnFinish() will never be executed.

This change should not cause compatibility issues, as consumers should not make assumptions about whether listeners will be run in a new goroutine.

Fix the concurrency issues when I use rclone rc package as a library in my project.

I searched and got nothing.

#### Checklist

### DIFF
--- a/cmd/serve/dlna/dlna_util.go
+++ b/cmd/serve/dlna/dlna_util.go
@@ -108,7 +108,7 @@ func (lrw *loggingResponseWriter) logRequest(code int, err interface{}) {
 		err = ""
 	}
 
-	fs.LogPrintf(level, lrw.request.URL, "%s %s %d %s %s",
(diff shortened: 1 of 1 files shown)


ncw (commented):
Well spotted.

Can you add a test which fails (maybe under the race detector) without your patch?

hayden-pan:
> Well spotted.
> 
> Can you add a test which fails (maybe under the race detector) without your patch?

Ok, I maked another pull request #8124 , please help check : )

ncw:
Thanks!

PS next time just update the existing pull request. Either add a new commit or force push it - thanks!

hayden-pan:
> Thanks!
> 
> PS next time just update the existing pull request. Either add a new commit or force push it - thanks!

Sorry, I am not familiar with the GitHub PR workflow

PR #8125, opened 2024-10-10
Disable Automatic Authentication Skipping for Unix Sockets

mozlima:

This commit removes the bypass of authentication for requests coming from Unix sockets in the rclone serve functionality. This change addresses security concerns when behind a reverse proxy, where skipping authentication can expose sensitive endpoints. It ensures that authentication checks are consistently applied

#### Checklist

### DIFF
--- a/lib/http/context.go
+++ b/lib/http/context.go
@@ -36,12 +36,6 @@ func IsAuthenticated(r *http.Request) bool {
 	return false
 }
 
-// IsUnixSocket checks if the request was received on a unix socket, used to skip auth & CORS
-func IsUnixSocket(r *http.Request) bool {
-	v, _ := r.Context().Value(ctxKeyUnixSock).(bool)
-	return v
-}
-
 // PublicURL returns the URL defined in NewBaseContext, used for logging & CORS
 func PublicURL(r *http.Request) string {
 	v, _ := r.Context().Value(ctxKeyPublicURL).(string)
--- a/lib/http/middleware.go
+++ b/lib/http/middleware.go
@@ -57,11 +57,6 @@ func NewLoggedBasicAuthenticator(realm string, secrets goauth.SecretProvider) *L
 func basicAuth(authenticator *LoggedBasicAuth) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			// skip auth for unix socket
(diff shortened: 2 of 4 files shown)


ncw:
> This commit removes the bypass of authentication for requests coming from Unix sockets in the rclone serve functionality. This change addresses security concerns when behind a reverse proxy, where skipping authentication can expose sensitive endpoints. It ensures that authentication checks are consistently applied

This is a breaking change so I'm going to need a bit more explanation on exactly why it causes a security problem.

Unix socket permissions are controlled with file system permissions so if the permissions you set on the unix sockets are too wide then that is a problem with the setup I would have thought?

Please explain exactly how this is causing an issue for you - thank you :-)

mozlima:
When you configure a reverse proxy and let rclone handle authentication, it quietly disables the authentication process if it's listening on a Unix socket. As a result, no authentication is enforced

ncw:
Can you describe what reverse proxy you used and how you set it up please. Also what rclone service were you proxying?

Thanks

mozlima:
@ncw  

```
# /tmp/Caddyfile
:8060 {
  route {
    reverse_proxy unix//var/tmp/rclone.sock
  }
}

```
```
caddy run --environ --config /tmp/Caddyfile
```

```
rclone -vv serve restic --htpasswd  .htpasswd --addr 'unix:///var/tmp/rclone.sock' /tmp
```
```
curl -X POST 'http://fake:fake@myhost.org:8060/fake/test?create=true'
```

```
ls /tmp/fake
```

ncw:
Ah, OK I see what you mean now - thank you very much for the example. I didn't get that the reverse proxy was talking to `rclone serve` via unix sockets before.

We need to be careful about backwards compatibility though as this feature is used a lot in rclone integrations.

So a unix socket needs to have no auth if no other auth flags were supplied like `--htpasswd  /tmp/htpasswd` otherwise we'll definitely break existing setups.

Can you test those scenarios please and make sure it doesn't need auth?

What do you think about writing a test for the unix socket auth? Check that without auth options it is unauthenticated but is authenticated with auth options?

ncw (approved):
Thank you, this looks perfect now.

I had to rebase it manually as there was something really odd about the PR - when I rebased it half the code disappeared! I had to fish the changes out your PR with git diff and use patch to put them back in.

No idea why that was - I think your code ended up committed in one of the merges rather than a separate commit.

Anyway once the CI goes green I will merge - thank you :-)

PR #8138, opened 2024-10-16
sftp: allow inline ssh public certificate for sftp

mimiteto:
Currently rclone allows us to specify the path to a public ssh certificate file.
That works great for cases where we can specify key path, like local envs.
If users are using rclone with [volsync](https://github.com/backube/volsync/tree/main/docs/usage/rclone) there currently is a limitation that users can specify only the rclone config file. With this change users can pass the public certificate in the same fashion as tehy can with `key_file`.

This change allows reading ssh certificates from within the config file.

Nope.

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -100,6 +100,11 @@ in the new OpenSSH format can't be used.`,
 			IsPassword: true,
 			Sensitive:  true,
 		}, {
+			Name: "pubkey",
+			Help: `SSH public certificate for public certificate based authentication.
+Set this if you have a signed certificate you want to use for authentication.
+If specified will override pubkey_file.`,
+		}, {
 			Name: "pubkey_file",
 			Help: `Optional path to public key file.
 
@@ -511,6 +516,7 @@ type Options struct {
 	KeyPem                  string          `config:"key_pem"`
 	KeyFile                 string          `config:"key_file"`
 	KeyFilePass             string          `config:"key_file_pass"`
+	PubKey                  string          `config:"pubkey"`
 	PubKeyFile              string          `config:"pubkey_file"`
 	KnownHostsFile          string          `config:"known_hosts_file"`
 	KeyUseAgent             bool            `config:"key_use_agent"`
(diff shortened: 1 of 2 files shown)


backend/sftp/sftp.go
```diff
@@ -904,6 +910,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 
 	keyFile := env.ShellExpand(opt.KeyFile)
 	pubkeyFile := env.ShellExpand(opt.PubKeyFile)
+	pubKey := env.ShellExpand(opt.PubKey)
```
albertony, line 913: This is not documented above, like for `pubkey_file`, i.e. `+ env.ShellExpandHelp`. But I'm not sure it should be used at all? Expanding a leading `~` into user's home directory path does not make sense here. Or is there a reason I don't see?

backend/sftp/sftp.go
```diff
@@ -997,13 +1004,21 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		}
 
 		// If a public key has been specified then use that
-		if pubkeyFile != "" {
-			certfile, err := os.ReadFile(pubkeyFile)
-			if err != nil {
-				return nil, fmt.Errorf("unable to read cert file: %w", err)
+		if pubkeyFile != "" || pubKey != "" {
+			pubKeyRaw := []byte(pubKey)
```
albertony, line 1008: Should it use `strconv.Unquote` the same way as is done for `opt.KeyPem` above, which I'm assuming is to handle backslash escaped special characters?

backend/sftp/sftp.go
mimiteto: That is an error on my side. Should be like that.

backend/sftp/sftp.go
albertony: I see. This is probably more specific to the PEM-encoded private key, which "should be on a single line with line endings replaced with '\n'" (as documented), that's why that needs the Unquote part. So I guess you can just forget about my suggestion. 😔

mimiteto:
@albertony is there anything else to be done on this PR?

albertony:
Not from my side, no, good work. I'll let @ncw do the final review, though.

ncw (approved):

Will merge now.

PR #8139, opened 2024-10-16
s3: Storj provider: fix server-side copy of files bigger than 5GB

kaloyan-raev:

Like some other S3-compatible providers, Storj does not currently implements UploadPartCopy and returns NotImplemented errors for multi-part server side copies.

This patch works around the problem by raising --s3-copy-cutoff for Storj to the maximum. This means that rclone will never use multi-part copies for files in Storj. This includes files larger than 5GB which (according to AWS documentation) must be copied with multi-part copy. This works fine for Storj.

See https://github.com/storj/roadmap/issues/40

No

#### Checklist


ncw (approved):
Nice fix - thank you :-)

PR #8140, opened 2024-10-19
docs: instructs avoiding inline comments in a filter-file

tgfisher:

Inline comments in a "--filter-from" file effectively remove the filter on that line (with directories in my hands). As far as I can tell this behavior isn't documented/warned and doesn't throw errors.

[PR requested](https://forum.rclone.org/t/inline-comments-effectively-remove-filter-in-filter-from/48301) in forum.

#### Checklist

### DIFF
--- a/docs/content/filtering.md
+++ b/docs/content/filtering.md
@@ -505,8 +505,8 @@ processed in.
 Arrange the order of filter rules with the most restrictive first and
 work down.
 
-Avoid inline comments. These can remove the filtering action of the
-associated filter.
+Inline comments are not supported. These comments alter the filtering action of the
+associated filter. _Use `-vv --dump filters` to see how they appear in the final regexp._
 
 E.g. for `filter-file.txt`:
 
@@ -515,7 +515,7 @@ E.g. for `filter-file.txt`:
     + *.jpg
     + *.png
     + file2.avi
-    - /dir/tmp/** # inline comment nullifies this exclusion
+    - /dir/tmp/** # WARNING! The inline comment nullifies this exclusion.
     - /dir/Trash/**
     + /dir/**
     # exclude everything else
--- a/docs/content/filtering.md
+++ b/docs/content/filtering.md
@@ -505,8 +505,7 @@ processed in.
 Arrange the order of filter rules with the most restrictive first and
 work down.
 
-Inline comments are not supported. These comments alter the filtering action of the
-associated filter. _Use `-vv --dump filters` to see how they appear in the final regexp._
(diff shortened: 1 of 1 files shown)


tgfisher:
- [x] stronger language

docs/content/filtering.md
```diff
@@ -505,13 +505,17 @@ processed in.
...
+associated filter. _Use `-vv --dump filters` to see how they appear in the final regexp._
+
 E.g. for `filter-file.txt`:
 
     # a sample filter rule file
     - secret*.jpg
     + *.jpg
     + *.png
     + file2.avi
+    - /dir/tmp/** # WARNING! The inline comment nullifies this exclusion.
```
albertony, line 518:
```suggestion
    - /dir/tmp/** # WARNING! This text will be treated as part of the path.
```

docs/content/filtering.md
```diff
@@ -505,13 +505,17 @@ processed in.
 Arrange the order of filter rules with the most restrictive first and
 work down.
 
+Inline comments are not supported. These comments alter the filtering action of the
+associated filter. _Use `-vv --dump filters` to see how they appear in the final regexp._
```
albertony, line 509:
```suggestion
Lines starting with # or ; are ignored, and can be used to write comments. Inline comments are not supported. _Use `-vv --dump filters` to see how they appear in the final regexp._
```

albertony (commented):
Thanks.

There is definitely room for improvement here, I agree.

But I'm thinking - this line:

> \- /dir/tmp/** # WARNING! The inline comment nullifies this exclusion.

actually excludes a literal path `/dir/tmp/** # WARNING! The inline comment nullifies this exclusion.`.

Simple as that. So, with all due respect, I think your description is a bit misleading:

> These comments alter the filtering action of the associated filter.

> \# WARNING! The inline comment nullifies this exclusion.

The concept of comments is not really mentioned in the docs, except in examples. But in the section for `--files-from` the following description can be found:

> Lines starting with # or ; are ignored.

I totally understand that, at for least someone with programing experience, it is easy to compare with comments in other languages, and then it could be easy to assume inline comments are supported as well.

I think, ideally, there should be a section explaining this comment / ignored lines feature for filter-related files. And document that it applies to all of them:
  - `--files-from`
  - `--include-from`
  - `--exclude-from`
  - `--filter-from`

You don't have to do that as part of this PR, though. I added suggestions for adjusting your changes, if we do something like that we could merge that for now. What do you think?

tgfisher:
Thanks @albertony. Your comments/changes are very nice. Absolutely agree with them and I appreciate the help! I see now that my version was too colloquial and kinda sloppy.... "These comments alter the filtering action [_from what you would expect given the_] associated filter.", haha.

albertony:
Great! I'll merge it now. Thank you for your effort.

PR #8149, opened 2024-10-24
Remove publish app mention

madalinignisca:
This is not required to use the application for yourself or a small group of trusted people.

Removes the mention to publish the application in Google Cloud as it is not required for personal use, including if id and secret is shared with a small group of trusted people included in the testing users.

N/A

#### Checklist

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -1797,7 +1797,8 @@ create and delete files with RClone.
 "Save and continue" to add test users. Be sure to add your own account to
 the test users. Once you've added yourself as a test user and saved the
 changes, click again on "Credentials" on the left panel to go back to
-the "Credentials" screen.
+the "Credentials" screen. You can add up to 100 test users, to whom you
+must share the id and secret to be able to use it.
 
    (PS: if you are a GSuite user, you could also select "Internal" instead
 of "External" above, but this will restrict API use to Google Workspace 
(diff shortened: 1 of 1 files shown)


docs/content/drive.md
```diff
@@ -1814,10 +1815,7 @@ then select "OAuth client ID".
    If you chose "Internal" you don't need to publish and can skip straight to
    Step 11 but your destination drive must be part of the same Google Workspace.)
 
-10. Go to "Oauth consent screen" and then click "PUBLISH APP" button and confirm.
-   You will also want to add yourself as a test user.
-
-11. Provide the noted client ID and client secret to rclone.
+10. Provide the noted client ID and client secret to rclone.
```
ncw, line 1818: I'd prefer it that you left this step but put a note that it isn't necessary for such and such a reason. How does that sound?

madalinignisca: Sounds OK for me. My thoughts would be that Google might reject and disable apps for some people if purely this publish becomes a normal step in any situation, not only rclone, as per too many requests that never end in paid projects for them.

madalinignisca: Actually, we can forget about this. I did not know about expiring grant.

madalinignisca:
I did not know about the expiring grants and the note after instructions makes more sense.

PR #8191, opened 2024-11-17
fs/config: fix certain logs ignoring `--quiet`

### LINKED ISSUE #8190: Option `--quiet` fails to prevent logging

When using `rclone config update` the tool will print to stdout the configuration of the remote you updated even if you have the `--quiet` flag set.

```
rclone v1.69.0-DEV
- os/version: nixos 24.11 (64 bit)
- os/kernel: 6.6.59 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.23.3
- go/linking: dynamic
- go/tags: none
```

`rclone config -q update ari key value`

```
➜  ~ rclone config -q update ari key value
[ari]
host = ari
key_file = /home/jess/.ssh/id_ed25519
type = sftp
user = jess
key = value
```

#### How to use GitHub

ttrssreal:
Hi, I'm packaging this downstream and have a script printing logs when it shouldn't.
Is `fs.Logf` the right method to use in this context?
Thanks!

Fixes #8190

See #8190

#### Checklist

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -303,6 +303,9 @@ func doConfig(name string, in rc.Params, do func(config.UpdateRemoteOpt) (*fs.Co
 	if err != nil {
 		return err
 	}
+	if updateRemoteOpt.NoOutput {
+		return nil
+	}
 	if !(updateRemoteOpt.NonInteractive || updateRemoteOpt.Continue) {
 		config.ShowRemote(name)
 	} else {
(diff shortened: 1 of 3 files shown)


ncw (changes requested):
Sorry for the delay reviewing this.

I don't think this is quite the right approach as replacing the `fmt.Printf` with `fs.Logf` will decorate the printout with timestamps which will look very different.

https://github.com/rclone/rclone/blob/2c72842c10016e8b7a33507bf7a70f2fe03b4444/cmd/config/config.go#L306-L308

I think what we want to do is just not call `config.ShowRemote(name)` here.

I don't think this should be based on a log level though - these aren't logs.

So maybe a new flag for `config update` how about `--no-output`?

What do you think?

ttrssreal:
Okie sounds good!
Is this what you mean?

ncw (approved):
Perfect - thank you :-)

PR #8199, opened 2024-11-20
drive: added moveid function to backend

spencexyz:

I'm moving files based on their label from Google Drive to GCS. To do that I create an array of files and then need to move them by ID. I can't use the standard move function because files can have the same name in Drive which would create a conflict when passing in the absolute path of the file to be moved. There was already a function to copy files by ID, I've added one to move files by ID so that they are deleted from their origin after being copied to GCS.

These discussions relate to copying files by ID, not moving them. I couldn't find any about moving.
https://github.com/rclone/rclone/issues/3625
https://forum.rclone.org/t/copy-files-using-drive-file-id/18881/11

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -3559,6 +3559,41 @@ func (f *Fs) copyID(ctx context.Context, id, dest string) (err error) {
 	return nil
 }
 
+// move file with id to dest (original file is deleted, unlike with copy)
+func (f *Fs) moveID(ctx context.Context, id, dest string) (err error) {
+	info, err := f.getFile(ctx, id, f.getFileFields(ctx))
+	if err != nil {
+		return fmt.Errorf("couldn't find id: %w", err)
+	}
+	if info.MimeType == driveFolderType {
+		return fmt.Errorf("can't copy directory use: rclone copy --drive-root-folder-id %s %s %s", id, fs.ConfigString(f), dest)
+	}
+	info.Name = f.opt.Enc.ToStandardName(info.Name)
+	o, err := f.newObjectWithInfo(ctx, info.Name, info)
+	if err != nil {
+		return err
+	}
+	destDir, destLeaf, err := fspath.Split(dest)
+	if err != nil {
+		return err
+	}
+	if destLeaf == "" {
+		destLeaf = path.Base(o.Remote())
+	}
+	if destDir == "" {
+		destDir = "."
+	}
+	dstFs, err := cache.Get(ctx, destDir)
+	if err != nil {
+		return err
+	}
+	_, err = operations.Move(ctx, dstFs, nil, destLeaf, o)
+	if err != nil {
+		return fmt.Errorf("move failed: %w", err)
+	}
+	return nil
+}
+
 // Run the drive query calling fn on each entry found
 func (f *Fs) queryFn(ctx context.Context, query string, fn func(*drive.File)) (err error) {
 	list := f.svc.Files.List()
@@ -3792,6 +3827,29 @@ attempted if possible.
 Use the --interactive/-i or --dry-run flag to see what would be copied before copying.
 `,
 }, {
+	Name:  "moveid",
+	Short: "Move files by ID",
+	Long: `This command copies files by ID
+
+Usage:
+
+    rclone backend moveid drive: ID path
+    rclone backend moveid drive: ID1 path1 ID2 path2
+
+It moves the drive file with ID given to the path (an rclone path which
+will be passed internally to rclone copyto). The ID and path pairs can be
+repeated.
+
+The path should end with a / to indicate copy the file as named to
+this directory. If it doesn't end with a / then the last path
+component will be used as the file name.
+
+If the destination is a drive backend then server-side copying will be
+attempted if possible.
+
+Use the --interactive/-i or --dry-run flag to see what would be copied before copying.
+`,
+}, {
 	Name:  "exportformats",
 	Short: "Dump the export formats for debug purposes",
 }, {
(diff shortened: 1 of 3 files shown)


ncw (changes requested):
This looks great thank you.

I put some minor things inline.

You might want to consider factoring `copyID` into `copyOrMoveID` and passing a an action name "move" or "copy" and a function pointer "operations.Copy" or "operations.Move" to remove the code duplication.

backend/drive/drive.go
```diff
@@ -3559,6 +3559,41 @@ func (f *Fs) copyID(ctx context.Context, id, dest string) (err error) {
...
 }
 
+// move file with id to dest (original file is deleted, unlike with copy)
+func (f *Fs) moveID(ctx context.Context, id, dest string) (err error) {
+	info, err := f.getFile(ctx, id, f.getFileFields(ctx))
+	if err != nil {
+		return fmt.Errorf("couldn't find id: %w", err)
+	}
+	if info.MimeType == driveFolderType {
+		return fmt.Errorf("can't copy directory use: rclone copy --drive-root-folder-id %s %s %s", id, fs.ConfigString(f), dest)
```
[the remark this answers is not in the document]
spencexyz: Just checked and that command does work. Thanks for catching these errors!

spencexyz:
> This looks great thank you.
> 
> I put some minor things inline.
> 
> You might want to consider factoring `copyID` into `copyOrMoveID` and passing a an action name "move" or "copy" and a function pointer "operations.Copy" or "operations.Move" to remove the code duplication.

Thanks for the feedback. I've gone ahead and combined them like you suggested. Can you take a look and let me know if this is what you had in mind?

spencexyz:
@ncw would you happen to have time to take a look at this in the next week or so?

backend/drive/drive.go
```diff
@@ -3969,16 +3978,16 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
 			dir = arg[0]
 		}
 		return f.unTrashDir(ctx, dir, true)
-	case "copyid":
-		if len(arg)%2 != 0 {
-			return nil, errors.New("need an even number of arguments")
+	case "copyOrMoveID":
```
ncw, line 3981: I think we need to keep the "copyid" command for backwards compatibility, so I'd suggest you make this

```go
case "copyid", "moveid":
```

The command definitions will need duplicating though.

ncw (commented):
I like the deduplication of code - nice :-)

I think we need to undo part of it though as we need to keep the separate "copyid" and "moveid" commands for backwards compatibility. See inline.

Sorry for the delay reviewing this - bit of a PR pile up right now!

backend/drive/drive.go
spencexyz: Sounds good, I went ahead and made those changes.

spencexyz:
@ncw thanks for taking the time to look at this!

I see what you mean, good point. I've gone ahead and made those changes. Let me know how these changes look.

backend/drive/drive.go
```diff
@@ -3969,16 +3994,16 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
...
-			return nil, errors.New("need an even number of arguments")
+	case "copyid", "moveid":
+		if len(arg)%2 == 0 {
+			return nil, errors.New("need an odd number of arguments")
 		}
 		for len(arg) > 0 {
 			id, dest := arg[0], arg[1]
-			arg = arg[2:]
-			err = f.copyID(ctx, id, dest)
+			arg = arg[3:]
```
ncw, line 4003: This should say `arg = arg[2:]` to pop the used arguments off for the `for` loop

backend/drive/drive.go
```diff
@@ -3969,16 +3994,16 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
...
+			err = f.copyOrMoveID(ctx, name, id, dest)
 			if err != nil {
-				return nil, fmt.Errorf("failed copying %q to %q: %w", id, dest, err)
+				return nil, fmt.Errorf("failed moving %q to %q: %w", id, dest, err)
```
ncw, line 4006: Probably needs a `%s` and `name` argument

backend/drive/drive.go
```diff
@@ -3776,11 +3782,9 @@ Result:
 Usage:
 
     rclone backend copyid drive: ID path
-    rclone backend copyid drive: ID1 path1 ID2 path2
```
ncw, line 3779: The pairs can still be repeated so you can undo these changes

backend/drive/drive.go
```diff
@@ -3789,7 +3793,28 @@ component will be used as the file name.
...
+}, {
+	Name:  "moveid",
+	Short: "Move files by ID",
+	Long: `This command moves files by ID
+
+Usage:
+
+    rclone backend moveid drive: ID path
+
+It moves the drive file with ID given to the path (an rclone path which
```
ncw, line 3807: This should have the text about repeated pairs also

backend/drive/drive.go
```diff
@@ -3969,16 +3994,16 @@ func (f *Fs) Command(ctx context.Context, name string, arg []string, opt map[str
 			dir = arg[0]
 		}
 		return f.unTrashDir(ctx, dir, true)
-	case "copyid":
-		if len(arg)%2 != 0 {
-			return nil, errors.New("need an even number of arguments")
+	case "copyid", "moveid":
+		if len(arg)%2 == 0 {
+			return nil, errors.New("need an odd number of arguments")
```
ncw, line 3999: This should be `even` as name is not a parameter

ncw (changes requested):
This is looking great thank you :-)

There are some changes that need to be undone I think - I've marked those inline.

Thank you

spencexyz:
@ncw thanks for your help and patience with this PR. This is my first time contributing to a project like this and you've been really helpful.

I made those changes, let me know what you think.

ncw (approved):

ncw:
That looks great now thank you @spencexyz 

I just rebased and squashed the branch and dropped the `drive.md` change which we don't need. Hopefully the CI will go green and I can merge :-)

Thank you very much for contributing this, and very happy to have more of your contributions.

ncw:
Thank you

Issue #8208, opened 2024-11-26
when installed from snap, unable to mount OneDrive


divinity76:
#### The associated forum post URL from `https://forum.rclone.org`

N/A

#### What is the problem you are having with rclone?

When rclone is installed from snap, it is unable to mount OneDrive.
If rclone is installed from apt or from https://github.com/rclone/rclone/releases `rclone-v1.68.2-linux-amd64.deb` (the *exact* same version that Snap offered during my testing) , OneDrive mount works just fine.

#### What is your rclone version (output from `rclone version`)

`snap install rclone` (broken): 1.68.2
`apt install ./rclone-v1.68.2-linux-amd64.deb` from https://github.com/rclone/rclone/releases  (working): 1.68.2
`apt-get install rclone` (working): 1.60.1

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

WSL-Ubuntu24.04 x64 running on WSL2 running on Windows 10 x64

#### Which cloud storage system are you using? (e.g. Google Drive)

OneDrive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

```
rclone mount onedrive: /remote/onedrive/ -vvv
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
$ namei -l /remote/onedrive/
f: /remote/onedrive/
drwxr-xr-x root root /
drwxr-xr-x hans hans remote
drwxr-xr-x hans hans onedrive
$ id
uid=1000(hans) gid=1000(hans) groups=1000(hans),27(sudo),33(www-data),100(users),113(docker)
$ rclone mount onedrive: /remote/onedrive/ -vvv
2024/11/26 20:47:54 DEBUG : rclone: Version "v1.68.2" starting with parameters ["/snap/rclone/522/bin/rclone" "mount" "onedrive:" "/remote/onedrive/" "-vvv"]
2024/11/26 20:47:54 DEBUG : Creating backend with remote "onedrive:"
2024/11/26 20:47:54 DEBUG : Using config file from "/home/hans/snap/rclone/522/.config/rclone/rclone.conf"
[...]

PR #8210, opened 2024-11-26
cmd/mountlib: better snap mount error message

divinity76:

mounting will always fail when rclone is installed from the snap package manager.
But the error message generated when trying to mount from a snap install was not
very good. Improve the error message.
Should prevent more bugreports like https://github.com/rclone/rclone/issues/8208
(I spent an hour on that bugreport.. Time wasted that wouldn't be if the error message was better)

discussed in issue #8208.

#### Checklist

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -373,6 +373,9 @@ func (m *MountPoint) Mount() (mountDaemon *os.Process, err error) {
 
 	m.ErrChan, m.UnmountFn, err = m.MountFn(m.VFS, m.MountPoint, &m.MountOpt)
 	if err != nil {
+	        if len(os.Args) > 0 && strings.HasPrefix(os.Args[0], "/snap/") {
+        	    return nil, fmt.Errorf("mounting is not supported when running from snap")
+        	}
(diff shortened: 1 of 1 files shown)


divinity76:
Haven't found a way to actually test this because snap use mount tricks and read-only filesystem to protect it's install files, to the point where even root can't touch the filesystem the normal way, 
quote:
```
$ sudo bash -c 'cat ./rclone > /snap/rclone/current/bin/rclone'
bash: line 1: /snap/rclone/current/bin/rclone: Read-only file system
```

But the code seems trivial enough to me.. I'm requesting review.

ncw (approved):
This looks very useful, thank you and will undoubtedly help people in the future.

I'll merge when the ci goes green.

divinity76:
@ncw thanks, was just a formatting issue, should be fixed now but the CI's didn't auto-rerun when I pushed the formatted version.. 
>1 workflow awaiting approval
>This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Seems a maintainer needs to manually approve that the CI's re-run?

ncw (approved):
Thank you - looks great now :-)

PR #8218, opened 2024-11-30
plcoud: openWriterAt: make client/fd handling atomic - fixes #8147

### LINKED ISSUE #8147: OpenWriterAt implementations causes problems when copying bigger files
See discussion in 

https://forum.rclone.org/t/copy-sync-of-large-files-to-pcloud-always-failing/48334

Workarrounds are the options:

--multi-thread-streams=1

or

--disable OpenWriterAt

gfelbing:

This PR stabilizes the situation of bigger files and multi-threaded uploads.
The root cause boils down to the old "fun" property of pclouds fileops API: sessions are bound to TCP connections.
This forces us to use a http client with only a single connection underneath.
With large files, we reuse the same connection for each chunk.
If that connection interrupts (e.g. because we are talking through the internet), all chunks will fail.
The probability for latter one increases with larger files.
As the point of the whole multi-threaded feature was to speed-up large files in the first place, this change pulls the client creation (and hence connection handling) into each chunk.
This should stabilize the situation, as each chunk (and retry) gets its own connection.

Additionally, the default buffersize of multi-threaded uploads are not really suitable for this use-case.
For the use case here, a buffersize equal to the chunk-size would be suitable, as we just want to split the chunks, and can upload those in total. Also, the overhead for each writeAt is quite high, it doesn't make sense to make this smaller.

> @ncw unfortunatelly, I don't see an "easy" way to change the default buffersize when just implementing OpenWriterAt.
> AFAIK, the only way to do it would be to switch to OpenChunkWriter, or is there a way for a backend to change the default of global config?

#8147

#### Checklist

### DIFF
--- a/backend/pcloud/pcloud.go
+++ b/backend/pcloud/pcloud.go
@@ -399,14 +399,15 @@ func (f *Fs) OpenWriterAt(ctx context.Context, remote string, size int64) (fs.Wr
 	if err != nil {
 		return nil, fmt.Errorf("open file: %w", err)
 	}
+	if _, err := fileClose(ctx, client, f.pacer, openResult.FileDescriptor); err != nil {
+		return nil, fmt.Errorf("close file: %w", err)
+	}
 
 	writer := &writerAt{
 		ctx:    ctx,
-		client: client,
 		fs:     f,
 		size:   size,
 		remote: remote,
-		fd:     openResult.FileDescriptor,
 		fileID: openResult.Fileid,
 	}
 
--- a/backend/pcloud/writer_at.go
+++ b/backend/pcloud/writer_at.go
@@ -18,21 +18,14 @@ import (
 // writerAt implements fs.WriterAtCloser, adding the OpenWrtierAt feature to pcloud.
 type writerAt struct {
 	ctx    context.Context
-	client *rest.Client
 	fs     *Fs
 	size   int64
 	remote string
-	fd     int64
 	fileID int64
 }
 
 // Close implements WriterAt.Close.
 func (c *writerAt) Close() error {
-	// close fd
-	if _, err := c.fileClose(c.ctx); err != nil {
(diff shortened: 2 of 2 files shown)


gfelbing:
With this fix, I am uploading a 16.2GB large file using Starlink+VPN with ~1MBit/s upload.
My options are `--multi-thread-write-buffer-size 16M --multi-thread-chunk-size 16M --multi-thread-streams 8`.
ATM, after 2.5h, I am currently at 3.2GB, ~200/1039 chunks finished. 
So far all chunks were successfully uploaded.

ncw:
Hi @gfelbing 

I just noticed the pcloud integration tests are failing like this

```
"go test -v -timeout 1h0m0s -remote TestPcloud: -verbose -test.run '^TestIntegration$/^FsMkdir$/^FsOpenWriterAt$|^TestIntegration$/^FsMkdir$/^FsPutFiles$/^(FsDirMove|FsListDirFile2|FsListLevel2|FsListR|FsListRDirFile2|FsListRLevel2|FsPurge|FsPurgeRoot)$|^TestIntegration$/^FsMkdir$/^FsPutFiles$/^FromRoot$/^(List|ListEntries)$'" - Starting (try 5/5)
=== RUN   TestIntegration
    fstests.go:438: Using remote "TestPcloud:"
=== RUN   TestIntegration/FsMkdir/FsOpenWriterAt
    fstests.go:790: 
        	Error Trace:	/home/rclone/go/src/github.com/rclone/rclone/fstest/fstests/fstests.go:790
        	            	open file: open new file descriptor: pcloud error: Access denied. You do not have permissions to perform this operation. (2003)
        	Test:       	TestIntegration/FsMkdir/FsOpenWriterAt
```

With a new access denied message.

Any ideas? I think this is related to the code you did for this, but I'm not 100% sure.

PR #8222, opened 2024-12-02
torrent backend

jkaberg:

This PR adds an remote to view and read/seek .torrent files located in an root directory in/on the target directory - the contents of the torrent files is exposed. This is my first pass at golang and adding an remote to rclone so I thought it would be appropriate to get some feedback before I finish up. It's mostly working, having some minor issues with seeking that I'm sorting out but should be ready for an initial review. 

Not that I'm aware. 

#### Checklist


jkaberg:
I'll revisit this later, the current approach will be simplified.

ncw:
I had a brief skim through the code. Very nice! Good tests which are important for a read only backend since we can't use rclone's integration test suite.

Once thing I noted - try not to upgrade the version of go from 1.21 in the `go.mod` - this may mean you have to pick some slightly older dependencies.

Another thought - I have an `archive` backend cooking at the moment - https://github.com/rclone/rclone/tree/archive-backend - which does the same thing - finding things with special extensions (eg `.zip` or `.torrent`) and looking inside them. Just a thought - maybe we should merge the torrent backend into the archive backend? The individual archiver packages implement the `fs.Fs` interface but don't have to worry about the virtual file system bit, eg [the zip package handler](https://github.com/rclone/rclone/blob/archive-backend/backend/archive/zip/zip.go).

PR #8228, opened 2024-12-04
update pcloud doc

marsjane:

update pcloud doc to avoid puzzling token error when use remote rclone authorize

No. Just I met this when use the doc and want to save other ppl's time

#### Checklist

### DIFF
--- a/docs/content/pcloud.md
+++ b/docs/content/pcloud.md
@@ -1,7 +1,7 @@
 ---
 title: "pCloud"
 description: "Rclone docs for pCloud"
-versionIntroduced: "v1.40"
+versionIntroduced: "v1.39"
 ---
 
 # {{< icon "fa fa-cloud" >}} pCloud
--- a/docs/content/pcloud.md
+++ b/docs/content/pcloud.md
@@ -1,7 +1,7 @@
 ---
 title: "pCloud"
 description: "Rclone docs for pCloud"
-versionIntroduced: "v1.39"
+versionIntroduced: "v1.40"
 ---
 
 # {{< icon "fa fa-cloud" >}} pCloud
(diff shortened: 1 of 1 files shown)


docs/content/pcloud.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "pCloud"
 description: "Rclone docs for pCloud"
-versionIntroduced: "v1.39"
+versionIntroduced: "v1.40"
```
albertony, line 4: Why this change?

albertony (commented):
Thanks. Looks good to me, except I have one question.

docs/content/pcloud.md
marsjane: aha sry I just thought I may need to update version if it has a new update, let me revert this part back

albertony (approved):

marsjane:
> I think the should have happened automatically. Did you cut and paste the rclone authorise command along with the base64 characters when you ran it? Those characters should include the hostname.
> 
> It is possible this is broken - it is supposed to work without manual editing of the config
Yes I understand it is totally automatic when I set it up on a normal PC, I tried it do set up all correctly.
But the another case is I config the rclone on a nas, which has no website browser, so I connect to it via ssh to config it, when it ask for the token config, I just run the rclone authorise command on another pc and copy back the shown token config, sth like: (there is no hostname in it)
```
Paste the following into your remote machine --->
{"access_token":"","token_type":"","expiry":""}
<---End paste
```
In this case if you don't set the hostname manually you will get token error.

ncw (approved):
Yes you are right @marsjane 

I will merge this - thank you :-)

PR #8229, opened 2024-12-04
Fix loading of authorized keys file with comment on last line

### LINKED ISSUE #8227: [sftp] rclone sftp serve cannot parse authorized_keys file ends with comments

When I tried to run `rclone sftp serve` with authorized_keys below, rclone cannot load the public keys. (`2024/12/04 11:11:05 NOTICE: Loaded 0 authorized keys from "/home/pusnow/.ssh/authorized_keys"`)

```
ssh-ed25519 <MY_PUBLIC_KEY1>
ssh-ed25519 <MY_PUBLIC_KEY2>
ecdsa-sha2-nistp256 <MY_PUBLIC_KEY3>
ssh-ed25519 <MY_PUBLIC_KEY4>
ecdsa-sha2-nistp256 <MY_PUBLIC_KEY5>
# comment
```

Removing the comment temporary fixes the problem.

```
rclone v1.68.2-DEV
- os/version: fedora 41 (64 bit)
- os/kernel: 6.11.6-300.fc41.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.23.3
- go/linking: static
- go/tags: none
```

Fedora 41, 64bit, x86_64

rclone serve sftp

`rclone serve sftp  $(pwd)`

```
2024/12/04 11:17:46 DEBUG : rclone: Version "v1.68.2-DEV" starting with parameters ["rclone" "-vv" "serve" "sftp" "/home/pusnow"]
2024/12/04 11:17:46 DEBUG : Creating backend with remote "/home/pusnow"
2024/12/04 11:17:46 DEBUG : Using config file from "/home/pusnow/.config/rclone/rclone.conf"
2024/12/04 11:17:46 INFO  : Local file system at /home/pusnow: poll-interval is not supported by this remote
2024/12/04 11:17:46 NOTICE: Loaded 0 authorized keys from "/home/pusnow/.ssh/authorized_keys"
2024/12/04 11:17:46 INFO  : 
Errors:                 1 (retrying may help)
Elapsed time:         0.0s

2024/12/04 11:17:46 DEBUG : 4 go routines active
2024/12/04 11:17:46 NOTICE: Failed to sftp: no authorization found, use --user/--pass or --authorized-keys or --no-auth or --auth-proxy
```

#### How to use GitHub

albertony:

Existing:

In serve sftp function `loadAuthorizedKeys`, after reading the authorized keys file, `ssh.ParseAuthorizedKey` from `golang.org/x/crypto` is called with the file content as input to parse a single authorized key. It silently skips invalid lines and comment lines until it is able to parse a key, and returns it together with remaining text. If it is unable to find a key, it returns an error. We repeat the call `ssh.ParseAuthorizedKey` with any remaining file content, in a loop, while collecting all parsed keys. However, this means that when reaching a point in the file content where there is remaining text but from which no valid keys can be parsed, such as comment lines at the end of the authorized keys file, we will end up getting an error return. The error is returned by `loadAuthorizedKeys`, it is not reported back to user unless authorized keys file were explicitly used with `--authorized-keys` option, but one will get a log entry "NOTICE: Loaded 0 authorized keys from ...".

New:

Ignores the error returned by `ssh.ParseAuthorizedKey`. The only error that can be returned is that no key was found. This makes sense, since it already silently skips invalid lines and comment lines and returns success as long as it can find one key. If one wants to ensure a key was found, then one can just count the number of returned keys instead.

Fixes #8227

#### Checklist

### DIFF
--- a/cmd/serve/sftp/server.go
+++ b/cmd/serve/sftp/server.go
@@ -143,8 +143,13 @@ func (s *server) serve() (err error) {
 		authKeysFile := env.ShellExpand(s.opt.AuthorizedKeys)
 		authorizedKeysMap, err = loadAuthorizedKeys(authKeysFile)
 		// If user set the flag away from the default then report an error
-		if err != nil && s.opt.AuthorizedKeys != Opt.AuthorizedKeys {
-			return err
+		if s.opt.AuthorizedKeys != Opt.AuthorizedKeys {
+			if err != nil {
+				return err
+			}
+			if len(authorizedKeysMap) == 0 {
+				return fmt.Errorf("failed to parse authorized keys")
+			}
 		}
 		fs.Logf(nil, "Loaded %d authorized keys from %q", len(authorizedKeysMap), authKeysFile)
 	}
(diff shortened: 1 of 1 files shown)


ncw (approved):
This looks OK to me :-)

Issue #4152, opened 2020-04-20
Change Notifications for Local Filesystems


darthShadow:
#### What is your current rclone version (output from `rclone version`)?

N/A

#### What problem are you are trying to solve?

Change Notifications for Local Filesystem Mounts

#### How do you think rclone should be changed to solve that?

Add [fsnotify](https://github.com/fsnotify/fsnotify) integration to support change events from Local Filesystems

There does exist a similar issue: https://github.com/rclone/rclone/issues/249 but that's limited to running commands like sync, copy, etc. based on changes. This is more of a broader-level integration for local mounts.

darthShadow:
A couple of caveats with the above integration:

* fsnotify doesn't have in-built support for recursive watches until issues [18](https://github.com/fsnotify/fsnotify/issues/18) & [41](https://github.com/fsnotify/fsnotify/issues/41) are resolved.
    - A workaround would be to do it manually, as implemented in [rfsnotify](https://github.com/farmergreg/rfsnotify/blob/master/rfsnotify.go)
* There are limits to the number of directory watches in every OS.
	- For Linux, it can be increased via `sysctl -w fs.inotify.max_user_watches=<NUMBER>`
	- This needs to be figured out for Mac & Windows.

Sunil-P:
I think this is [relevant](https://krypted.com/mac-os-x/maximum-files-in-mac-os-x/) for Mac.

ncw:
It probably wouldn't be too difficult to build a prototype based on fsnotify + rfsnotify. Maybe we should have a go at doing that and see how bad the resource usage is for large directory trees.

@darthShadow - fancy having a go at this?

darthShadow:
Sorry, but I very much doubt I will have the time to work on it in the next couple of weeks.
I have already added it to my TODO list, however, and will see if I can at least prototype it and check the resource usage for different sizes of directory trees.

ivandeex:
@ncw @darthShadow 

[fsnotify](https://github.com/fsnotify/fsnotify) does _not_ look like a good candidate as it fails to implement recursive notify handles (required by rclone) for as long as _6 years_ (sic!). The _fsnotify_ author has admitted the fact and recommended other projects in https://github.com/fsnotify/fsnotify/issues/18#issuecomment-538689153.

[...]

Issue #249, opened 2015-12-11
Feature request: Listen filesystem events and sync


dbcm:
It would be nice to have rclone in daemon mode, listening some paths and when something change, rclone sync it.

ncw:
A nice idea.  Could use this library: https://github.com/howeyc/fsnotify

dbcm:
\o/

dibu28:
+1

timofonic:
Nice. I already suggested it too. 

I think this issue was posted earlier. My subject contained "daemon mode", a more explicit expression. 

Anyway, all of you explained it better and provided an interesting candidate piece of software to make the task easier. Thanks! 

On December 11, 2015 7:58:58 PM CET, dibu28 <notifications@github.com> wrote:
>+1
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/ncw/rclone/issues/249#issuecomment-164017813

-- 
Enviado desde mi dispositivo Android con K-9 Mail. Por favor disculpa mi brevedad.

timofonic:
I have an issue with using Google Drive over Ext4 under Windows. I use dual and hate it, but I have no option because of homework.

Here's the relevant issue over Ext4 by using ext2fsd

: 
https://sourceforge.net/p/ext2fsd/discussion/143329/thread/37c888a1/?limit=25
<img width="577" alt="googledrivebigeyecantbelied" src="https://cloud.githubusercontent.com/assets/49494/11811379/7c17e3aa-a335-11e5-8925-7408418abe44.PNG">
<img width="480" alt="googledrivehardlinktoext4directory" src="https://cloud.githubusercontent.com/assets/49494/11811380/7c1d0a06-a335-11e5-8784-e3284d5fe3ef.PNG">
<img width="573" alt="google-drive-wants-ntfs" src="https://cloud.githubusercontent.com/assets/49494/11811381/7c25b8a4-a335-11e5-8923-cadd1785dd6a.PNG">

I found CloudSyncHelper might be useful, but I'm unable to understand how to install it. I submitted an issue:
https://github.com/XElementSoftware/CloudSyncHelper/issues/6

zeshanb:
Having doubts this CloudSyncHelper is what it's claiming to be:
https://github.com/XElementSoftware/CloudSyncHelper/blob/master/deployment/CloudSyncHelper.xml

timofonic:
@zeshanb 

[...]

PR #8258, opened 2024-12-20
ChangeNotify() for local backend using fsnotify

lawmurray:

This is an initial implementation of `ChangeNotify()` for the local backend, using [fsnotify](https://github.com/fsnotify/fsnotify). I expect that some further discussion and iteration and testing will be needed, but hope this implementation is a useful starting point for that.

Related issues: #249 #4152

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate. **Ran `rclone test changenotify` but further tests may be warranted yet**

#### Further notes

fsnotify does not support recursive watching of a directory, and it seems that it will not (fsnotify/fsnotify#18) until there is support for all platforms that it targets. There is some work for specific platforms, such as Linux with inotify, although this is not yet released (fsnotify/fsnotify#472).

So at this stage, rather than relying on fsnotify for recursive support, I've implemented that in rclone itself, using `filepath.WalkDir()` (which is what fsnotify/fsnotify#472 is doing anyway), including watching new directories when created. Watching a directory is initiated before listing and recursing into it, which should eliminate race conditions if changes are made while the watchers are established (but further thinking and testing would be worthwhile!).

I've tested this on Linux only. It will need testing on other platforms.

An interactive test routine is to run `rclone test changenotify . -vv` to watch the current working directory, then perform some file operations and see what is logged. The debug logging is verbose for this purpose (it reports all events; this could be reduced eventually). Something like `mkdir -p a/b/c` makes for an interesting test (turns out that an event is produced only for `a`, so it's necessary to then recursively watch `a` to keep the watch list up to date). Perhaps some fuzzing tests with random timings would be useful for race conditions, which may be platform specific.

I hope this is useful, and very happy for others to help with this as well, of course.

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1305,11 +1305,13 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 				entryPath := event.Name
 				entryType := fs.EntryObject
 
-				info, err := os.Stat(entryPath)
-				if err != nil {
-					fs.Debugf(f, "Failed to stat %s", entryPath)
-				} else if info.IsDir() {
-					entryType = fs.EntryDirectory
+				if event.Op != fsnotify.Remove {
+					info, err := os.Stat(entryPath)
+					if err != nil {
+						fs.Debugf(f, "Failed to stat %s", entryPath)
+					} else if info.IsDir() {
+						entryType = fs.EntryDirectory
+					}
 				}
 
 				switch event.Op {
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1302,36 +1302,44 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 					return
 				}
 
+				entryPath := event.Name
+				entryType := fs.EntryObject
+
+				info, err := os.Stat(entryPath)
+				if err != nil {
+					fs.Debugf(f, "Failed to stat %s", entryPath)
+				} else if info.IsDir() {
+					entryType = fs.EntryDirectory
+				}
+
 				switch event.Op {
 				case fsnotify.Create:
-					fs.Debugf(f, "Create: %s", event.Name)
-					info, err := os.Stat(event.Name)
-					if err != nil {
-						fs.Debugf(f, "Failed to stat %s", event.Name)
-					} else if info.IsDir() {
-						err := watcher.Add(event.Name)
+					fs.Debugf(f, "Create: %s", entryPath)
+					if entryType == fs.EntryDirectory {
+						err := watcher.Add(entryPath)
 						if err != nil {
-							fs.Debugf(f, "Failed to start watching %s", event.Name)
+							fs.Debugf(f, "Failed to start watching %s", entryPath)
 						} else {
-							fs.Debugf(f, "Started watching %s", event.Name)
+							fs.Debugf(f, "Started watching %s", entryPath)
 						}
 					}
 				case fsnotify.Write:
-					fs.Debugf(f, "Write: %s", event.Name)
+					fs.Debugf(f, "Write: %s", entryPath)
 				case fsnotify.Remove:
-					fs.Debugf(f, "Remove: %s", event.Name)
-					// No need to stop watching directories when removed, handled by
-					// fsnotify
+					// No need to stop watching removed directories, handled by fsnotify
+					fs.Debugf(f, "Remove: %s", entryPath)
 				case fsnotify.Rename:
-					fs.Debugf(f, "Rename: %s", event.Name)
+					fs.Debugf(f, "Rename: %s", entryPath)
 				case fsnotify.Chmod:
-					fs.Debugf(f, "Chmod: %s", event.Name)
+					fs.Debugf(f, "Chmod: %s", entryPath)
 				}
+				notifyFunc(entryPath, entryType)
+
 			case err, ok := <-watcher.Errors:
 				if !ok {
 					return
 				}
-				fs.Debugf(f, "Received error: %s", err.Error())
+				fs.Debugf(f, "Error: %s", err.Error())
 			}
 		}
 	}()
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1277,22 +1277,7 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 			return
 		}
 		defer watcher.Close()
-
-		// Add root path, recursively
-		//
-		// For a directory, WalkDir() makes the callback before listing, so
-		// watching for changes can begin before the initial listing
-		filepath.WalkDir(f.root, func(path string, d os.DirEntry, err error) error {
-			if d.IsDir() {
-				err := watcher.Add(path)
-				if err != nil {
-					fs.Debugf(f, "Failed to start watching %s", path)
-				} else {
-					fs.Debugf(f, "Started watching %s", path)
-				}
-			}
-			return nil
-		})
+		f.watchPath(watcher, f.root)
 
 		// Process events and errors
 		for {
@@ -1318,12 +1303,7 @@ func (f *Fs) ChangeNotify(ctx context.Context, notifyFunc func(string, fs.EntryT
 				case fsnotify.Create:
 					fs.Debugf(f, "Create: %s", entryPath)
 					if entryType == fs.EntryDirectory {
-						err := watcher.Add(entryPath)
-						if err != nil {
-							fs.Debugf(f, "Failed to start watching %s", entryPath)
-						} else {
-							fs.Debugf(f, "Started watching %s", entryPath)
-						}
+						f.watchPath(watcher, entryPath)
 					}
 				case fsnotify.Write:
 					fs.Debugf(f, "Write: %s", entryPath)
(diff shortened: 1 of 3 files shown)


lawmurray:
Looking into these failed tests.

fstest/fstests/fstests.go
```diff
@@ -876,18 +877,18 @@ func Run(t *testing.T, opt *Opt) {
 			require.NoError(t, err)
 
 			pollInterval := make(chan time.Duration)
-			dirChanges := map[string]struct{}{}
```
lawmurray, line 879: This change fixes some tests for the union backend, as multiple goroutines now write to these maps concurrently.

lawmurray:
While not yet ready for a merge, this should be in a good state for feedback and discussion now. I've expanded the comments to explain some of the caveats of implementing `ChangeNotify()` with fsnotify. Assuming those caveats are acceptable, I may need a little help on the failing tests (just focused on the [Linux tests](https://github.com/rclone/rclone/actions/runs/12445623349/job/34747135355?pr=8258) for now):

1. `FAIL: TestMount`: does this assume that `ChangeNotify()` is not implemented? The `changeNotify()` at `rclone/vfs/dir.go:285` looks like it invalidates the cache so that it may be correct for the extra directory to appear now.
2. `FAIL: TestRcPollInterval`: if I understand correctly this test looks to check that the polling interval is respected, but in doing so assumes that the implementation uses polling? This one does not.

Still trying to understand these tests some more.

lawmurray:
I now have [all tests passing](https://github.com/lawmurray/rclone/actions/runs/12870466766), just on a different branch to avoid creating more noise here. There are still known issues that need discussion if this is to be successful.

I did make quite a mess of this pull request sorry! I'll get it tidied up, or perhaps just close and open a new one, but hope to have this ready for review soon.

ncw:
> I now have [all tests passing](https://github.com/lawmurray/rclone/actions/runs/12870466766), just on a different branch to avoid creating more noise here. There are still known issues that need discussion if this is to be successful.

Great - well done :-)

> I did make quite a mess of this pull request sorry! I'll get it tidied up, or perhaps just close and open a new one, but hope to have this ready for review soon.

What I would do is backup the branch then squash all the commits down to one to clean the history, rebase on master then force push. That will give the best experience to review without opening a new PR.

Can you put a list of open issue that need discussing at that point then we can work out what to do with them.

Thank you

backend/local/changenotify_other.go
```diff
@@ -0,0 +1,275 @@
+//go:build !windows
```
lawmurray, line 1: While there is some commonality between the implementation for Windows and other platforms, I've separated them out with conditional compilation to keep the logic clearer on reading.

fstest/fstests/fstests.go
lawmurray, line 1: The union backend tests fail due to concurrent writes to maps. I've added a mutex to fix this. An alternative is `sync.Map` but it makes error messages harder to parse when expected vs actual results differ.

vfs/rc_test.go
```diff
@@ -97,7 +98,15 @@ func TestRcPollInterval(t *testing.T) {
 	}
 	out, err := call.Fn(context.Background(), nil)
 	require.NoError(t, err)
-	assert.Equal(t, rc.Params{}, out)
```
lawmurray, line 100: My understanding of this test is that it's just checking the reported polling interval. That changes for a backend with `ChangeNotify` enabled, so I have updated the expected output to match.

lawmurray:
> What I would do is backup the branch then squash all the commits down to one to clean the history, rebase on master then force push. That will give the best experience to review without opening a new PR.
> 
> Can you put a list of open issue that need discussing at that point then we can work out what to do with them.

Thanks @ncw. I've done all that and I think it's now ready for review. I've deleted posts of mine that are no longer relevant and updated my initial post with a detailed description of the known issues that need discussion. Do let me know if you have questions on any of that.

darthShadow:
Should we consider adding a dedupe buffer for write events? Otherwise writing of large files will create a lot of events, even with a larger notify interval.

Ideally we would have this PR (https://github.com/fsnotify/fsnotify/pull/651) merged already, so it's handled automatically, atleast for Linux, but since it's not, we will need to implement the dedupe ourselves similar to the logic here: https://github.com/fsnotify/fsnotify/blob/main/cmd/fsnotify/closewrite.go

ncw (commented):
I put a few comments on the code inline.

I think probably my major concern is performance. We are putting recursive listeners on all the directories in the Linux code. That is going to use a lot of resources.

When I try it on my home directory it uses over 1GB of memory and the watchers start failing with

```
2025/02/26 18:04:53 ERROR : Local file system at /home/ncw: Failed to start watching go/pkg/mod/golang.org/x/tools@v0.23.0/go/ssa/testdata/src/encoding/xml, already removed? no space left on device
```

For an industrial strength tool like rclone, I think it should be able to monitor any size of directory.

Currently there are only two users of ChangeNotify

- cmd/test/changenotify/changenotify.go
- vfs/vfs.go

The VFS use being the most important - it is used to invalidate directories in the directory cache.

So I would certainly consider changing the interface if that makes it easier to implement.

If we do want to use it for bisync or continuous syncing we should have a think about what interface we want there too.

lawmurray:
@darthShadow

> Should we consider adding a dedupe buffer for write events? Otherwise writing of large files will create a lot of events, even with a larger notify interval.

Sure. If I understand the logic in the link, it's basically, "don't send a write event if another write event happens in the next 100ms". The code looks to launch a goroutine for every path, I'll try and explore working it into the polling implementation instead, so many goroutines looks a little heavy for the purpose to me.

backend/local/changenotify_windows.go
```diff
@@ -0,0 +1,222 @@
...
+
+// Hack to enable recursive watchers in fsnotify, which are available for
+// Windows and Linux (although not quite what is needed for Linux here), but
+// not yet Mac, hence there is no public interface for this in fsnotify just
+// yet. This is currently only needed for, and enabled for, Windows builds.
+//
+// Setting fsnotify.enableRecurse to true enables recursive handling: paths
+// that end with with \... or /... as watched recursively.
+
+//go:linkname enableRecurse github.com/fsnotify/fsnotify.enableRecurse
```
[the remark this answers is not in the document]
lawmurray: Was not aware that it has been deprecated actually, but definitely aware of its naughtiness :laughing: 

It's not ideal, obviously, and meant to be a temporary hack to keep moving. Let me see where upstream is at on potentially providing a public API for this. Looks like they have at least one recent question on the topic (fsnotify/fsnotify#673).

lawmurray:
@ncw

> When I try it on my home directory it uses over 1GB of memory and the watchers start failing with
> 
> ```
> 2025/02/26 18:04:53 ERROR : Local file system at /home/ncw: Failed to start watching go/pkg/mod/golang.org/x/tools@v0.23.0/go/ssa/testdata/src/encoding/xml, already removed? no space left on device
> ```
I'll have to profile, but I think the memory use here is likely for the `known` map rather than for the watchers. The `known` map is keyed by the full path of each file and directory, whereas a more memory-efficient tree-like data structure that mimics the file system would reduce memory use a lot. Even better though, the only reason `known` exists at all is bookkeeping for the second argument to `ChangeNotify()`, so an interface change could get rid of it completely and I think there won't be this memory issue at all (which I agree is bad btw, 1 GB is way too much---just didn't want to optimize this data structure until having the conversation on interface).

The error I'll need to check out... `no space left on device` seems odd, unless that relates to the max number of watchers with inotify. This is, unfortunately, always an issue on Linux, I've even seen it with the official Dropbox client. You can (writing here for testing purposes, but should definitely be included in documentation down the line):

1. Check how many directories that will need watchers (one watcher is required per directory, files don't matter):
    ```
    find . -type d | wc -l
    ```
2. Work out maximum number of watchers allowed at present:
    ```
    sudo sysctl -a | grep fs.inotify.max
    ````
    giving by default on e.g. Ubuntu 24.04
    ```
    fs.inotify.max_user_instances = 128  
    fs.inotify.max_user_watches = 65536
    ```
    I believe the former is basically the number of processes that can create watchers, the latter the number of files/directories watched.

    (Incidentally, other distributions, like Arch, are more generous, configured with 1024/524288 by default).

3. Increase the allowed number temporarily:
    ```
    sudo sysctl -n fs.inotify.max_user_instances=1024
    sudo sysctl -n fs.inotify.max_user_watches=524288
    ```
    or write the same to `/etc/sysctl.conf` to make it permanent.

> So I would certainly consider changing the interface if that makes it easier to implement.
>
> If we do want to use it for bisync or continuous syncing we should have a think about what interface we want there too.

I'll follow up with some proposed next steps.

lawmurray:
By way of next step, what I'm thinking to do is:

1. Add a new value to the enum `EntryType`, say `EntryUncertain`, for use when a backend cannot reliably or efficiently determine the entry type. The idea is to not require code changes for backends that have an existing implementation that can reliably identify the entry type.
2. Update tests to allow the use of `EntryUncertain`. Some of them collect notifications separately for `EntryObject` and `EntryDirectory`. Ideally this is preserved and the tests will still check that the entry type is correct, if used, but will also allow the use of `EntryUncertain`.
3. Update the local backend to use `EntryUncertain`, see where we're at in terms of memory use, etc.

Sound okay? I understand this won't address everything yet, but it seems like a good next step to me.

Issue #8248, opened 2024-12-15
metrics: `rclone rc` binding metrics endpoint


hiddenmarten:
#### The associated forum post URL from `https://forum.rclone.org`
I don't have any, just wanted to create an issue to link my future PR with.

#### What is the problem you are having with rclone?
`rclone rc` trying to bind metrics endpoint if environment variable `RCLONE_METRICS_ADDR` is set

#### What is your rclone version (output from `rclone version`)
v1.68.2

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Linux 

#### Which cloud storage system are you using? (e.g. Google Drive)
Doesn't matter in that case.

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
`RCLONE_METRICS_ADDR="0.0.0.0:9000" rclone rc -vv rc/noop`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

NOTE: `└──╼ $ RCLONE_METRICS_ADDR="0.0.0.0:9000" rclone rcd` was run in advance.

```
└──╼ $ RCLONE_METRICS_ADDR="0.0.0.0:9000" rclone rc -vv rc/noop
2024/12/15 22:21:21 DEBUG : Setting  metrics_addr="0.0.0.0:9000" from environment variable RCLONE_METRICS_ADDR
2024/12/15 22:21:21 DEBUG : Setting --metrics-addr "0.0.0.0:9000" from environment variable RCLONE_METRICS_ADDR="0.0.0.0:9000"
2024/12/15 22:21:21 DEBUG : rclone: Version "v1.68.2" starting with parameters ["rclone" "rc" "-vv" "rc/noop"]
2024/12/15 22:21:21 CRITICAL: Failed to start metrics server: failed to init server: listen tcp 0.0.0.0:9000: bind: address already in use
```

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

hiddenmarten:
_Duplicate explanation of use cases from PR for the sake of context:_

I encountered this issue when running rclone in Kubernetes. Let me provide a simplified Deployment manifest to explain the context:

```yaml
[...]

PR #8263, opened 2024-12-23
metrics: disable the metrics server if the main server is not configured to start

hiddenmarten:

Fix issue: https://github.com/rclone/rclone/issues/8248

No, it wasn't.

#### Checklist

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -429,11 +429,12 @@ func initConfig() {
 		fs.Fatalf(nil, "Failed to start remote control: %v", err)
 	}
 
-	// Start the metrics server if configured
-	_, err = rcserver.MetricsStart(ctx, &rc.Opt)
-	if err != nil {
-		fs.Fatalf(nil, "Failed to start metrics server: %v", err)
-
+	// Start the metrics server if configured and not running the "rc" command
+	if os.Args[1] != "rc" {
+		_, err = rcserver.MetricsStart(ctx, &rc.Opt)
+		if err != nil {
+			fs.Fatalf(nil, "Failed to start metrics server: %v", err)
+		}
 	}
 
 	// Setup CPU profiling if desired


hiddenmarten:
Hi, @ncw.

I've finished with the implementation, may I get a review? :)

ncw (commented):
I'm not sure this fix is correct.

Don't we want to be able to enable the metrics server if --rc isn't enabled?

This change will disable the metrics server if --rc is not passed.

What problem are you trying to fix?

hiddenmarten:
I'm trying to resolve this issue.
https://github.com/rclone/rclone/issues/8248

The goal is to ensure that the dedicated metrics server runs only if the main server is enabled. This approach makes sense to me, as it prevents the metrics server from being enabled during one-time operations, such as `rclone rc`.

ncw:
Sorry didn't notice the issue you made where you explained everything nicely ☺️

Currently the way you enable the metrics is by setting the metrics address. Not sure why you are setting that if you don't want metrics?

We decided we didn't need a metrics enable flag but maybe that is what you want?

hiddenmarten:
> "Not sure why you are setting that if you don't want metrics?"

I encountered this issue when running rclone in Kubernetes. Let me provide a simplified Deployment manifest to explain the context:

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      containers:
        - name: rclone
          image: rclone/rclone:1.68.2
          args:
            - rcd
          env:
            - name: RCLONE_METRICS_ADDR
              value: ":9000"
          livenessProbe:
            exec:
              command:
                - rclone rc rc/noop
```

In this setup, the liveness probe fails because the `rclone rc rc/noop` command does not work correctly when the `RCLONE_METRICS_ADDR` environment variable is set. 

As a workaround, we can modify the liveness probe command like this: `RCLONE_METRICS_ADDR="" rclone rc rc/noop`

However, this approach raises a concern: it's not immediately clear from the manifest why we are overriding the `RCLONE_METRICS_ADDR` variable. This lack of clarity could lead to confusion for others reading the configuration.

Another potential scenario to consider: If someone sets `RCLONE_METRICS_ADDR` globally in their `~/.bashrc` or `~/.zshrc` file, they may inadvertently prevent rclone from working correctly when running commands locally from the CLI. 

> We decided we didn't need a metrics enable flag but maybe that is what you want?

Probably, I think it's also a good approach. Tell me if you prefer to just have a separate flag for enabling metrics and I'll update this PR.

ncw:
Thanks for explaining more. Here is my understanding

You are running

    RCLONE_METRICS_ADDR=:9000 rclone rcd -vv

And you'd like to see it is alive. Due to the way the environment variables are handled, you end up running

    RCLONE_METRICS_ADDR=:9000 rclone rc rc/noop

Which gives the error

    CRITICAL: Failed to start metrics server: failed to init server: listen tcp :9000: bind: address already in use

So it is starting the metrics server on the `rclone rc` command that you are having a problem with.

So I guess the solutions are

1. `RCLONE_METRICS_ADDR="" rclone rc rc/noop` - as you stated above
2. Disable the metrics server on `rclone rc`
3. Add a --metrics-enabled flag then use it `rclone rc rc/noop --metrics-enabled=false`

Any others?

What do you think?

hiddenmarten:
You're absolutely correct, I have to work a bit on my explanation if it took a few iterations to be understandable. :)

My plan was to implement a second option since it doesn't require any effort from a user perspective during `rclone rc` call.
But I'm ready to implement any solution you find the best in the current situation.

ncw:
> You're absolutely correct, I have to work a bit on my explanation if it took a few iterations to be understandable. :)

:-)
 
> My plan was to implement a second option since it doesn't require any effort from a user perspective during `rclone rc` call. But I'm ready to implement any solution you find the best in the current situation.

That sounds OK. Your current PR does it for all commands which isn't OK.

hiddenmarten:
Hi @ncw,

I've updated the PR based on our discussion. I hope the changes are satisfactory for now.

P.S. The failed test doesn't seem to be related to the changes I made - it’s likely just a flaky test.

ncw (commented):
That's a creative fix - definitely has character! :-)

It would look better in `cmd/cmd.go` as at least we are parsing the command line there rather than in rcserver.

https://github.com/rclone/rclone/blob/fdff1a54ee67977b755a8c6fcfa33befad6e41b4/cmd/cmd.go#L432-L437

TBH I don't think there is a neat fix for this so anything we do is going to look ugly.

hiddenmarten:
> That's a creative fix - definitely has character! :-)

For sure! 😄

> It would look better in cmd/cmd.go, as at least we are parsing the command line there rather than in rcserver.

I had thought about that, honestly, but later realized that we control how the server is started in the rcserver package. That’s why I decided to put it in rcserver. Thanks for guiding me on this matter. 😊

> TBH, I don’t think there is a neat fix for this, so anything we do is going to look ugly.

Yep, I even tried to find a way to get it from the Cobra package but failed, unfortunately. That’s why I had to get "creative." 😅

hiddenmarten:
@ncw Please, check it out when you have time 🙏

ncw (approved):
Looks great - thank you :-)

PR #8273, opened 2024-12-25
webdav: added support for Digest authentication scheme (auto-detected) for server type 'other'

laurento:

For the webdav server type 'other', this PR adds support for the Digest authentication scheme.
It also auto-detects if the server offers Digest authentication and, if available, prioritizes it over Basic authentication.

My PR aims to offer a possible solution to the long-standing issue #2110 

#### Checklist

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -40,6 +40,7 @@ import (
 	"github.com/rclone/rclone/lib/rest"
 
 	ntlmssp "github.com/Azure/go-ntlmssp"
+	"github.com/icholy/digest"
 )
 
 const (
@@ -409,6 +410,24 @@ func (o *Object) filePath() string {
 	return o.fs.filePath(o.remote)
 }
 
+// checkForDigestAuth issues an unauthenticated request to the server to
+// determine if 'Digest' is among the supported authentication schemes.
+// Multiple challenges may be specified in a single response header
+// or provided in separate headers in the same response.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate
+func checkForDigestAuth(url string) bool {
+	resp, err := http.Get(url)
+	if err == nil {
+		for _, challenge := range resp.Header["Www-Authenticate"] {
+			if strings.Contains(challenge, "Digest") {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 // NewFs constructs an Fs from the path, container:path
 func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, error) {
 	// Parse config into Options struct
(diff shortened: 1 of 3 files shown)


backend/webdav/webdav.go
```diff
@@ -482,6 +501,18 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			rt: ntlmssp.Negotiator{RoundTripper: t},
 		}
 	}
+	if opt.Vendor == "other" {
+		// If available, prefer Digest authentication.
+		if checkForDigestAuth(opt.URL) {
+			fs.Debugf(nil, "using Digest authentication scheme")
+			client.Transport = &digest.Transport{
```
ncw, line 508: Can this use rclone's transport? Not using rclone 's transport will cause some of rclone's features to stop working

backend/webdav/webdav.go
```diff
@@ -482,6 +501,18 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 			rt: ntlmssp.Negotiator{RoundTripper: t},
 		}
 	}
+	if opt.Vendor == "other" {
+		// If available, prefer Digest authentication.
+		if checkForDigestAuth(opt.URL) {
```
ncw, line 506: This does an extra http transaction each time the backend is started which isn't ideal.

Could we fallback to digest if the transaction fails? That should give us the www authenticate header.

Or we could store the preference or digest Auth in the config and not probe for it, or maybe combine with above suggestion?

ncw (commented):
Thanks for doing this. Which system are you accessing?

I put some things to think about inline.

backend/webdav/webdav.go
laurento: Fair enough :) I changed it accordingly and it seems to work just fine. I'll push it with my next commit.

backend/webdav/webdav.go
laurento: As a matter of fact, that was my first idea/attempt. Initially, I didn't find a great place where to implement it. I'm not very familiar with rclone codebase.

However, after looking around, I'll try again in the coming days by implementing the fallback approach into `shouldRetry()`. I would appreciate any guidance you may have.

I don't have a strong opinion against adding a dedicated config param, but I think it would be nicer to automatically adapt to the server's authentication schemes. Let me give it a try. Let's see how it looks.

laurento:
[Cloudme](https://www.cloudme.com) webdav requires Digest authentication, and thus I started looking into it.

backend/webdav/webdav.go
ncw: Yes I agree auto adapting would be great.

shouldRetry sounds like a good place to put the logic.

Note that you can store automatically discovered stuff in the config file easily so that could be an option.

hoclun-rigsep:
fwiw I merged and built this myself on macOS with FUSE support and it seems to work fine!

PR #8278, opened 2024-12-29
Update drive.md

wlklch:
clarification of mount for internal google filetypes

#### Checklist

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -1699,7 +1699,7 @@ see User rate limit exceeded errors, wait at least 24 hours and retry.
 You can disable server-side copies with `--disable copy` to download
 and upload the files if you prefer.
 
-### Limitations of Google Docs
+### Limitations of Google Docs, Sheets, etc.
 
 Google docs will appear as size -1 in `rclone ls`, `rclone ncdu` etc,
 and as size 0 in anything which uses the VFS layer, e.g. `rclone mount`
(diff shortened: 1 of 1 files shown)


docs/content/drive.md
```diff
@@ -1720,6 +1720,8 @@ correct size and be downloadable. Whether it will work on not depends
 on the application accessing the mount and the OS you are running -
 experiment to find out if it does work for you!
 
+Affected are not only google docs but also google sheets (and probably other types in an internal google format) - the unknown filesize is only a symptom of a deeper problem. Although mount reports those files with the corresponding office-file-extension, they are of in the google internal format (after all, this is a mount, not sync!). If you double click such a file, your installed office (Writer, rsp. Word ...) tries to open it and fails - because it is not in office but in google format. Only when you copy/paste the file it will be converted to the corresponding office-format and you can work locally with it.
```
ncw, line 1723: This is actually incorrect. Some systems can deal with the unknown file size and load the file in converted format. The internal google format is **never** exported - there is no API to do that.

So I suggest re-wording this as

> Note that this affects all types of Google Office Suite file, such as Google Docs, Sheets, Slides, Drawings, Forms, etc. The rclone documents refer to these generically as "Google docs".

wlklch: This is exactly what I describe: download works on my system.
However, I want to clarify a different point. I want to detect, whether a file on the google drive ist the same as a local copy, in a reliable and in a net work efficient way. 

-  I understand, that rclone sync does this. 
- my usual way to check it, is to compare mtTime and filesize which fails (and I try to explain this in my text, and that this failure is not a malfunction of rclone, but has a deeper reason)
-  However,  my  real problem is not a program of mine, but a 3. party software (rdiff-backup) which  fails to "sync" on the mount of a google drive. It took me a long to understand this failure and my text proposal should speed up the process  in future

PR #8279, opened 2024-12-29
install.sh: ensure curl follows redirects in installation script

rkoyama1623:

Added the `-L` option to curl commands in the installation script to handle HTTP redirects properly. 
This resolves issues where the install script fails when some proxy server is used.

Not discussed yet.

#### Checklist

### DIFF
--- a/docs/content/install.sh
+++ b/docs/content/install.sh
@@ -53,9 +53,9 @@ export XDG_CONFIG_HOME=config
 #check installed version of rclone to determine if update is necessary
 version=$(rclone --version 2>>errors | head -n 1)
 if [ -z "$install_beta" ]; then
-    current_version=$(curl -fsS https://downloads.rclone.org/version.txt)
+    current_version=$(curl -fsSL https://downloads.rclone.org/version.txt)
 else
-    current_version=$(curl -fsS https://beta.rclone.org/version.txt)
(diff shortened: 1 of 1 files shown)


ncw:
What proxy servers require this? It seems like this would enable MITM attacks.

rkoyama1623:
In a network with a certain security system in place, it was not possible to download files without curl's -L option.
For security policy reasons, I will not describe the details of the proxy server.
This URL shows the similar case of mine: https://github.com/nvm-sh/nvm/issues/2152

However, as you pointed out, this change creates a vulnerability to MITM attacks.
I apologize for the inappropriate pull request due to my lack of security knowledge.

As mentioned in the issue above, I have recognized that this problem is on the network side, and I would like to close this change.

> Hijacking SSL is pretty fraught, especially if HSTS is used, so hopefully the solution is for them to "not do that" :-)
> [LINK](https://github.com/nvm-sh/nvm/issues/2152#issuecomment-1092057818)

ncw:
@rkoyama1623 Thanks for the explanation - much appreciated. I think I don't want to add the `-L` parameter in this case as it weakens security (very slightly). If someone else has this problem in the future they will find this issue and we can re-open the discussion if necessary.

PR #8283, opened 2025-01-02
accounting: add per ctx token bucket

shaofeng66:

The motivation of this change is, we use rclone as imported library in our go project, and there could be simultaneously operations such as multiple sync.CopyDir() calls in a single process, and each of them need have different bandwidth limit. We can not achive that by setting ConfigInfo.BwLimit b/c the TokenBucket in token_bucket.go is a global variable. Since each sync.CopyDir() have different context, we can use the context to pass different bwlimit to transfer.

expose a token bucket wrapper type, so that client code can set a limiter in context to limit the transfer bandwidth in context scope. this is useful when rclone is used as imported library.

The client code can be written as:
```golang
var bwp fs.BwPair
bwp.Set(fmt.Sprintf("%dB", bwlimit))

var tbw accounting.TokenBucketWrapper
tbw.SetBwLimit(bwp)

ctx = context.WithValue(ctx, accounting.TokenBucketWrapperPerContextKey, &tbw)

sync.CopyDir(ctx1, dstFS, srcFs, false)
```

#### Checklist

### DIFF
--- a/fs/accounting/accounting.go
+++ b/fs/accounting/accounting.go
@@ -344,6 +344,16 @@ func (acc *Account) limitPerFileBandwidth(n int) {
 	}
 }
 
+// Account for n bytes from the current context bandwidth limit (if any)
+func (acc *Account) limitPerContextBandwidth(n int) {
+	value := acc.ctx.Value(TokenBucketWrapperPerContextKey)
+	if value != nil {
+		if tbw, ok := value.(*TokenBucketWrapper); ok {
+			tbw.LimitBandwidth(TokenBucketSlotAccounting, n)
+		}
+	}
+}
+
 // Account the read and limit bandwidth
 func (acc *Account) accountRead(n int) {
 	// Update Stats
@@ -356,6 +366,7 @@ func (acc *Account) accountRead(n int) {
 
 	TokenBucket.LimitBandwidth(TokenBucketSlotAccounting, n)
 	acc.limitPerFileBandwidth(n)
+	acc.limitPerContextBandwidth(n)
 }
 
 // read bytes from the io.Reader passed in and account them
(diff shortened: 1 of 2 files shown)


ncw:
Did you try

      --bwlimit-file BwTimetable   Bandwidth limit per file in KiB/s, or use suffix B|K|M|G|T|P or a full timetable

This can be set in the ctx via `fs.AddConfig()` - and should do what you want I think.

https://github.com/rclone/rclone/blob/a389a2979b0a4726e180478f51074bc4a64a9826/fs/config.go#L590

shaofeng66:
> Did you try
> 
> ```
>   --bwlimit-file BwTimetable   Bandwidth limit per file in KiB/s, or use suffix B|K|M|G|T|P or a full timetable
> ```
> 
> This can be set in the ctx via `fs.AddConfig()` - and should do what you want I think.
> 
> https://github.com/rclone/rclone/blob/a389a2979b0a4726e180478f51074bc4a64a9826/fs/config.go#L590

looks like the "--bwlimit-file" config is per file bwlimit, but if a sync.CopyDir() call copy multiple files with multipe transfers(default 4?), the total bw would be the sum of all transfers? What I want is a bwlimit of a sync.CopyDir() call, no matter how many transfers it involves.

ncw:
I see what you mean.

I don't want to add API which can only be used by using rclone as a library as that isn't one of the primary ways of using rclone.

So we need to re-work this so it can work over the rc API too. Perhaps we could set a bandwidth limit as part of a stats group as that is easy to supply to rc calls.

What do you think?

ncw:
I'm going to close this as I haven't heard from you for a year.

Feel free to reopen or resubmit a fixed version

Thank you

PR #8297, opened 2025-01-09
build: update `github.com/shirou/gopsutil` to v4

Juneezee:

v4 is the latest version with bug fixes and enhancements. While there are 4 breaking changes in v4, they do not affect us because we do not use the impacted functions.

Note: Due to the library's versioning policy, v4.24.5 is the first v4 release https://github.com/shirou/gopsutil/releases/tag/v4.24.5.

#### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,7 +64,7 @@ require (
 	github.com/rfjakob/eme v1.1.2
 	github.com/rivo/uniseg v0.4.7
 	github.com/rogpeppe/go-internal v1.12.0
-	github.com/shirou/gopsutil/v3 v3.24.5
+	github.com/shirou/gopsutil/v4 v4.24.12
 	github.com/sirupsen/logrus v1.9.3
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
 	github.com/spf13/cobra v1.8.1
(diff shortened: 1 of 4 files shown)


ncw (approved):
This looks great - thank you.

Will merge now.

ncw:
In the [RELEASE.md](https://github.com/rclone/rclone/blob/master/RELEASE.md#update-dependencies) there are instructions on how to update the dependencies.

What is notably missing from these a check if modules have got a new major version. Any ideas on how to do that?

Juneezee:
> In the [RELEASE.md](https://github.com/rclone/rclone/blob/master/RELEASE.md#update-dependencies) there are instructions on how to update the dependencies.
> 
> What is notably missing from these a check if modules have got a new major version. Any ideas on how to do that?

@ncw From Go Modules point of view, `github.com/shirou/gopsutil/v3` and `github.com/shirou/gopsutil/v4` are two different modules. This is by design according to https://go.dev/doc/modules/release-workflow#breaking:

> In the module’s go.mod file, revise the module path to append the new major version number, as in the following example:
>
> ```
> example.com/mymodule/v2
> ```
>
> Given that the module path is the module’s identifier, this change effectively creates a new module. It also changes the package path, ensuring that developers won’t unintentionally import a version that breaks their code. Instead, those wanting to upgrade will explicitly replace occurrences of the old path with the new one.

This means that `go get -u ./...` does not update to the latest module path for us.

At the time of writing this comment, Dependabot is not capable of updating new module path https://github.com/dependabot/dependabot-core/issues/2213.

On the other hand, Renovate is able to detect and update to the new module path, as well as updating the import paths in our source code (https://docs.renovatebot.com/golang/):

> If the user has enabled the option gomodUpdateImportPaths in the [postUpdateOptions](https://docs.renovatebot.com/configuration-options/#postupdateoptions) array, then Renovate uses [mod](https://github.com/marwan-at-work/mod) to update import paths on major updates, which can update any Go source file

ncw:
Interesting, thank you. I don't really want to install a bot - we do the upgrades on a cycle and depends it is far too noisy (we tried it)

A command line tool would be ideal -  that could be run as part of the upgrade process to see if there were any major upgrades.

This looks like it might be the tool - what do you think? 

https://github.com/icholy/gomajor

Juneezee:
> This looks like it might be the tool - what do you think?
> 
> https://github.com/icholy/gomajor

Looks great! I didn’t know a tool like that existed.

PR #8299, opened 2025-01-11
feat: vfs shutdown should close poll chan

izouxv:

vfs shutdown shoud close poll chan


ncw (commented):
Seems reasonable... Did this cause you an actual problem?

izouxv:
yes,  it will run many goroutine.

izouxv:
> Seems reasonable... Did this cause you an actual problem?

the chan need close

ncw (approved):
Thank you :-)

PR #8302, opened 2025-01-13
Create a skeleton of a Cryptomator backend

tbodt:

Support Cryptomator encrypted vault format!

The implementation in this PR is far from done, but I want to merge it to get past the point of there only being one person who can work on it. But I do intend to keep working on it, don't expect it to just sit in this state.

Let me know what level of e.g. test and documentation coverage would be required for this to be merged.

#6954 

#### Checklist


fhilgers:
I originally wanted to create a lot of integration tests to verify my implementation against the reference client, but other projects got in the way and I never got to it. That was the main reason I did not want to put it in tree.

I just saw I also forgot to add a license to the repository. I'll fix that later today and add an MIT license. Feel free to take the code and move it in tree, if want.

I am also open for a PR that exposes the raw lower level API instead of having everything as internal.

If you have questions/suggestions, feel free to ask/ping me again.

tbodt:
Thanks for your reply! Embarrassed to say I didn't notice the missing license, but thanks for fixing that.

As I mentioned in the last comment, I would prefer an approach of copying code over here from the other repo, just so that me and anyone else who ends up working on this doesn't have to deal with making 2 PRs.

tbodt:
Just finished copying most of the code and tests from the library into this PR. Definitely a lot more test coverage than before! Also added support for SIV_GCM. As for the backend itself, that'll be harder to reuse because of removing the Vault abstraction.

tbodt:
Well, I got carried away, it's not a skeleton anymore :p it's got all the features of the last backend PR and a few more. I also wrote a test against the cryptomator CLi.

upbqdn:
Any chance this gets merged anytime soon?

ncw:
Hi @tbodt this is looking good :-)

I'm not sure why the tests aren't passing though - any ideas?

Do the tests use cryptomator to build archives and get rclone to read them and vice versa? That would be best if we can manage it but we'll need to provide cryptomator to the CI somehow. That only needs to run on one platform.

tbodt:
Yes, I did write tests that use the real cryptomator and I remember they were passing when i originally wrote them. I'll have to dig into the failures when I have time. Also I need to implement cryptomator's special handling of long file names.

tbodt:
Turns out the CI failures are only because of missing ListP implementation. Just added and pushed.

PR #8314, opened 2025-01-16
docs: explain the stringArray flag parameter descriptor

christophberger:
This PR adds a brief description of the `stringArray` parameter descriptor to flags.md, pointing out that there is actually no such thing as a string array for flag parameters, and that the user would have to repeat the flag for each string. 

See also: https://forum.rclone.org/t/questions-about-include-and-filter-and-stringarray/1638/2

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -619,6 +619,11 @@ it to `false`.  It is also possible to specify `--boolean=false` or
 parsed as `--boolean` and the `false` is parsed as an extra command
 line argument for rclone.
 
+Options documented to take a `stringArray` parameter accept multiple 
+values. To pass more than one value, repeat the option; for example: 
+`--include value1 --include value2`.
+
+
 ### Time or duration options {#time-option}
 
 TIME or DURATION options can be specified as a duration string or a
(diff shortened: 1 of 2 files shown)


albertony:
Good idea.

Maybe there should be a complete section defining the meaning of *all* the different value types mentioned for options? We do to some degree have that in https://rclone.org/docs/#options, although not complete, and also there are some additional more or less relevant supplementary information in a couple of other places, such as https://rclone.org/docs/#options-1 and https://rclone.org/docs/#human-readable. These do to a very little degree mention stringArray, so your text could be added there as well. Or, perhaps we should improve the https://rclone.org/docs/#options text and link to that from https://rclone.org/flags/ to avoid duplication...  🤔  What do you think?

christophberger:
@albertony Good point, I wasn't aware of the `docs/#options` section. I can add a description of `stringArray` to, or after, [this paragraph](https://rclone.org/docs/#options:~:text=Options%20that%20take,argument%20for%20rclone.) and link to that from `/flags`.

christophberger:
Done

albertony (approved):

PR #8316, opened 2025-01-17
Cleanup documentation of options in the usage page

albertony:

Currently in docs.md we have:
1. Inconsistent types used in docs.md compared to what is used internally, in flags.md, command line help etc.
    - In flags.md and command line help etc. we use the type names from code, while in docs.md we have used more custom identifiers, perhaps intended as a more human-readable form? But the inconsistency may just confuse more.
2. Inconsistent use of format `--option value` vs `--option=value` in headers.
    - In flags.md and command line help etc. we use `--option value`, so probably best to stick with that.
3. Inconsistent use of `--option` vs `--option=value` (and `--option=value` ) for options that do take a value in headers.
   - In flags.md and command line help etc. we consistently use `--option value` as the format.
    - Not really sure what I prefer here. On one hand using the short `--option` form for all options, regardless if pure (boolean) flag or one that takes a parameter value, looks best in doc pages. On the other hand we may not currently describe good enough if/what value it takes, so keeping that there may still give relevant detail. For now I've (mostly) kept the `--option value` header format where it was used before, but haven't gone through all existing `--option` headers to extend them if they do take a value.
5. Inconsistent use of option with value type, e.g. `--option string`, vs option with actual value, e.g. `--dump mapper`, in headers.
    - After some cleanup the only ones left are the `--dump` examples. I've added double quotes in an attempt at indicating what is actual values, e.g. `--dump "headers"`, although not sure how good the result is...

PS: Draft, with a lot of individual commits to be able to easily adjust. Will clean up later - perhaps squash everything into one commit?

#### Checklist


docs/content/docs.md
```diff
@@ -1066,7 +1066,7 @@ Mode to run dedupe command in.  One of `interactive`, `skip`, `first`,
 `newest`, `oldest`, `rename`.  The default is `interactive`.  
 See the dedupe command for more information as to what these options mean.
 
-### --default-time TIME
+### --default-time Time
```
ncw, line 1077: Add an `=` in this one?

docs/content/changelog.md
```diff
@@ -1013,7 +1013,7 @@ instead of of `--size-only`, when `check` is not available.
         * It can be disabled with the [--inplace](/docs/#inplace) flag.
         * This behaviour will also happen if the backend is wrapped, for example `sftp` wrapped with `crypt`.
     * The [s3](/s3/#s3-directory-markers), [azureblob](/azureblob/#azureblob-directory-markers) and [gcs](/googlecloudstorage/#gcs-directory-markers) backends now support directory markers so empty directories are supported (Jānis Bebrītis, Nick Craig-Wood)
-    * The [--default-time](/docs/#default-time-time) flag now controls the unknown modification time of files/dirs (Nick Craig-Wood)
+    * The [--default-time](/docs/#default-time-duration) flag now controls the unknown modification time of files/dirs (Nick Craig-Wood)
```
ncw, line 1016: Should probably add `{#id}` to the `--default-time` docs.

albertony:
@ncw if you want to take just a really quick look, and you have any immediate thoughts, perhaps something comes to mind that could lead to problems, let me know so I can adjust and/or avoid digging further into that direction. 😄 I'll ping you for a formal review later.

docs/content/docs.md
```diff
@@ -2735,56 +2735,56 @@ the body of the request will be gunzipped before showing it.
 
 The available flags are:
 
-#### --dump headers
+##### --dump "headers"
```
ncw, line 2746: I see what you are getting at here, but that is going to encourage the users to type `--dump "headers"` isn't it? Rather than the more succinct `--dump headers`

docs/content/docs.md
```diff
@@ -630,13 +631,21 @@ although some may also have different parameters such as `DumpFlags`
 or just `string` and the help text explains that this will be
 interpreted as a list.
 
-### Time or duration options {#time-option}
+### Time options
 
-Options that take a `Duration` can be specified as a duration string
-or a time string.
+Options that take a `Time` parameter must be specified as an absolute
```
ncw, line 636: I don't think is quite correct. `Time` options can take durations too, so `--default-time 1y` refers to a time 1 year ago.

`Duration` options can take times also so you can say `--min-age 2025-01-01`.

I don't think the original was quite correct either, but that is why it was only one section rather than two!

docs/content/docs.md
albertony: Not sure what you mean... I've tried to make headers consistently use `--option value` instead of `--option=value` for now, so this follows that. Do you mean we should swap all of these? Or is there something special about this specific one?

docs/content/docs.md
albertony: Exactly, I agree, that was what I meant with "not sure how good the result is"... 😆

docs/content/docs.md
albertony: Right. I did have a vague suspicion this could be a reason for the existing combined explanation. Will fix, i.e. change back to something more like it was.

docs/content/docs.md
albertony: Pushed a different variant: Make these inline code format, which should emphasize that they are examples:

```markdown
##### `--dump filters`
```

albertony: A possible other alternative is to just use the parameter value in these headers:

> #### --dump DumpFlags
>
> The `--dump` flag takes a comma separated list of flags to dump info about.
> ...
> 
> The available flags are:
>
> ##### headers
> ...
> ##### bodies
> ...


Or, if we don't need the automatic anchors for linking, just list them in the description for the main `--dump` option. I can't see that we currently link to any of them anyway. Something like:

> #### --dump DumpFlags
>
> The `--dump` flag takes a comma separated list of flags to dump info about.
>
>The available flags are:
>
>- headers
>    - Dump HTTP headers with `Authorization:` lines removed. May
>      still contain sensitive info.  Can be very verbose.  Useful for
>      debugging only. Use `--dump auth` if you do want the `Authorization:` headers.
>- bodies
>    - Dump HTTP headers and bodies - may contain sensitive info.  Can be
>      very verbose.  Useful for debugging only.
>    - Note that the bodies are buffered in memory so don't use this for
>      enormous files.

I'm leaning towards this last alternative. I find it to be more in line with what we do for other such "enum-like" options, e.g. https://rclone.org/docs/#log-level-level

docs/content/docs.md
albertony: > I don't think the original was quite correct either

I found one error when testing these options a bit, but created a separate PR suggesting fixing functionality to make it match the documentation: https://github.com/rclone/rclone/pull/8318.

albertony:
@ncw you may take a more formal review now - or remind me if you think there are existing comments I need to look more into first.

> We should probably be putting more {#id} on titles whenever we reference them probably was a general thought from reading the changes.

I have not done this (yet). Whenever I have changed headers, I've tried to be thorough and update all links that uses the autogenerated anchor. So maybe changing to id can be left for later?

docs/content/docs.md
ncw: The other changes in this commit all use `=` but I think you change them all later!

docs/content/docs.md
```diff
@@ -1475,14 +1475,14 @@ will only work well for the bucket based backends (eg s3, b2,
 azureblob, swift) but these are the only backends likely to have
 millions of entries in a directory.
 
-### --log-file=FILE
+### --log-file
 
-Log all of rclone's output to FILE.  This is not active by default.
+Log all of rclone's output to file.  This is not active by default.
```
ncw: This no longer makes sense without the `FILE` parameter. Could be "Log all of rclone's output to **a** file"?

Same with all the other changes in this commit (The `FILE` removal commit)

ncw (approved):
This is looking fantastic :-)

Let's get it merged before you have to spend ages rebasing it again!

At some point I would like to make sure we have an anchor for every internal reference we make but that can be for another time.

I think my only note of significance was the one with removing the FILE out of the header made the help no longer make sense.

I'm going to approve this - please merge when you are ready :-)

albertony:
Thank you!

> I think my only note of significance was the one with removing the FILE out of the header made the help no longer make sense.

Nice catch. There were also a couple similar for DIR. I've fixed them now.

> I'm going to approve this - please merge when you are ready :-)

I'm ready to do this, but one question: Do you prefer squashing in this case? There are 25 commits which I've tried to scope to each set of changes, but it may be a bit much to merge as is? I could squash all to one. Or group them into a handful of larger ones (at least try to, but may be getting into conflicts).

ncw:
I'm happy for you to merge these individually - each one is nicely self contained.

albertony:
Agreed. I'll merge it then (before more conflicts). And then hopefully someone will test it on tip.rclone.org (I will do so myself as well, of course).

PR #8318, opened 2025-01-17
config: make parsing of duration options consistent

albertony:

Currently, when parsing some `Duration` options, `time.Duration` is used instead of `fs.Duration`. This means suffix is required, and only `ms`, `s`, `m` and `h` are supported. For other options, where `fs.Duration` is used, then suffix is optional and defaults to `s`, while additional suffixes `d`, `w`, `M` and `y` are supported, as well as absolute times which will be interpreted as duration relative to now.

Example, attempting to specify `--contimeout 3` leads to error:

> Error: invalid argument "3" for "--contimeout" flag: parsing "3" as time.Duration failed: time: missing unit in duration "3"

#### Checklist


albertony:
First commit seems enough for parsing to go through `fs.Duration` and work. Second commit refactors to make the types used in `ConfigOptionsInfo` and `ConfigInfo` consistent, if we prefer this?

albertony:
Thanks. It does actually build at each commit. But I don't remember exactly the reasoning behind the 3 separate commits, so I think squashing is better anyway... 😊 I will squash merge it!

PR #8319, opened 2025-01-18
Clarify usage of SSE-C options

kellytrinh:
explain a bit more AWS requirements on the key/options

Added IDrive as supported backend only if enabled on their side.

Rclone currently accepts incorrect parameters to these options and the error messages are not helpful.

Added documentation so folks know what is expected.

Yes, I mentioned it here

https://forum.rclone.org/t/doucmentation-suggestion-s3-sse-c-key-requirement/49562

and was pointed here to create PR.

#### Checklist

Above not applicalbe.  Just look at the diff; this is a small change that obviously helps improve docs

### DIFF
--- a/docs/content/s3.md
+++ b/docs/content/s3.md
@@ -1271,13 +1271,13 @@ Properties:
 
 #### --s3-sse-customer-algorithm
 
-If using SSE-C, the server-side encryption algorithm used when storing this object in S3.
+If using SSE-C, the server-side encryption algorithm used when storing this object in S3.  As per AWS S3 standard, this must be "AES256" if SSE-C is used.
 
 Properties:
 
 - Config:      sse_customer_algorithm
 - Env Var:     RCLONE_S3_SSE_CUSTOMER_ALGORITHM
-- Provider:    AWS,Ceph,ChinaMobile,Minio
+- Provider:    AWS,Ceph,ChinaMobile,Minio, IDrive (if enabled via dashboard)
 - Type:        string
 - Required:    false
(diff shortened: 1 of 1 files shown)


docs/content/s3.md
```diff
@@ -1271,13 +1271,13 @@ Properties:
...
 
-If using SSE-C, the server-side encryption algorithm used when storing this object in S3.
+If using SSE-C, the server-side encryption algorithm used when storing this object in S3.  As per AWS S3 standard, this must be "AES256" if SSE-C is used.
 
 Properties:
 
 - Config:      sse_customer_algorithm
 - Env Var:     RCLONE_S3_SSE_CUSTOMER_ALGORITHM
-- Provider:    AWS,Ceph,ChinaMobile,Minio
+- Provider:    AWS,Ceph,ChinaMobile,Minio, IDrive (if enabled via dashboard)
```
ncw, line 1280: This also comes from here

https://github.com/rclone/rclone/blob/87eaf376298436c7991ff6975184961bb86bc895/backend/s3/s3.go#L1970-L1972

You can have IDrive in or out - you can't have a comment! If you need a comment then put it in the [IDrive section in s3.md](https://github.com/rclone/rclone/blob/master/docs/content/s3.md#idrive-e2-idrive-e2)

The same applies to the changes below.

ncw (changes requested):
Nice change thank you :-)

Unfortunately it isn't quite in the right place - I've indicated inline where it should go.

Thanks

PR #8350, opened 2025-01-30
smb: implement true multi-threaded upload

jgiannuzzi:
#### Checklist

### DIFF
--- a/backend/smb/filepool.go
+++ b/backend/smb/filepool.go
@@ -1,0 +1,91 @@
+package smb
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/cloudsoda/go-smb2"
+	"golang.org/x/sync/errgroup"
+)
+
+type smbChunkWriterFile struct {
+	*smb2.File
+	c *conn
+}
+
+func (w *smbChunkWriter) getFile(ctx context.Context) (f *smbChunkWriterFile, err error) {
+	w.poolMu.Lock()
+	if len(w.pool) > 0 {
+		f = w.pool[0]
+		w.pool = w.pool[1:]
+	}
+	w.poolMu.Unlock()
+
+	if f != nil {
+		return f, nil
+	}
+
+	w.f.addSession() // Show session in use
+
+	c, err := w.f.getConnection(ctx, w.share)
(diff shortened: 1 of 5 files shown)


ncw:
Can we fix OpenWriterAt to be multi-threaded? That would probably be a better solution as it maps what smb can do better than the OpenChunkWriter interface?

jgiannuzzi:
Good point, I'll try that approach! When I started working on multithreaded uploads for `smb`, I thought that implementing `OpenChunkWriter` was the only way to achieve that goal, but as I've now spent more time in the code base, I totally see what you mean.
If you don't mind, I'd like to keep this PR open as a draft until I come up with the new approach. That should hopefully be this coming week.

ncw:
Sounds great - thank you. 

I haven't checked, but I think the sftp backend might have this problem too.

backend/smb/filepool.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/cloudsoda/go-smb2"
+	"golang.org/x/sync/errgroup"
+)
+
```
ncw, line 11: Can you factor w.PoolMu and w.pool out of smbChunkWriter into a new type (`filepool` say) and define it here. It might needs some other stuff like an `f fs.Fs` but I think it is independent of `smbChunkWriter`.

You can make a newFilepool method then this factoring will become a robust re-usable type that is easy to reason about.

backend/smb/filepool.go
```diff
@@ -0,0 +1,91 @@
...
+type smbChunkWriterFile struct {
```
ncw, line 12: This then should be called `poolFile` or maybe `filepoolFile` or maybe just `file`.

jgiannuzzi:
Thanks for your comments @ncw! This PR was not ready, I had only pushed it after rebasing out of habit. I started the work on a different branch following a reusable approach and will open a new PR once I'm done.

ncw:
@jgiannuzzi no worries! Ping me on the PR when you want another review.

jgiannuzzi:
Superseded by #8719

PR #8354, opened 2025-01-30
Update rclone_serve_nfs.md

zackees:
NFS serve is not available for Windows, but there is no mention of this in the docs. Infact, the use of Windows in the text suggests that it is available.

This change adds a small message to let the user know that nsf serve is not available for Windows.


ncw (approved):
Thank you for noticing this and fixing the docs - appreciated :-)

PR #8360, opened 2025-02-04
Fix 7974 out of memory sync with s3 list prefix

alxfv:
see https://forum.rclone.org/t/s3-listobjects-v2-performance-optimization-for-a-flat-structures/49860


alxfv:
sorry, wanted to create draft PR in my fork

PR #8364, opened 2025-02-07
Update rclone_purge.md

zackees:
From the forums it's `--transfers` which controls the concurrency of this operation. So let's put this in the readme.


docs/content/commands/rclone_purge.md
```diff
@@ -7,6 +7,8 @@ description: "Remove the path and all of its contents."
 
 Remove the path and all of its contents.
 
+The concurrency of this operation is controlled by the `--transfers` global flag.
```
ncw, line 10: I just checked the source code - it is actually `--checkers` which controls the concurrency here.

Also note that the backend may implement this command directly in which case `--checkers` is ignored.

zackees:
Thanks for the note! I've added this to the documentation.

ncw (approved):
Looks great - thank you :-)

ncw:
Actually, just as I was merging it, I realised that this is in the wrong place. The file you edited is auto generated from here so can you edit this instead - sorry should have noticed that before!

https://github.com/rclone/rclone/blob/e175c863aa9950b67cebd18d95c56ed547a012f7/cmd/purge/purge.go#L19-L26

zackees:
Move it over here

https://github.com/rclone/rclone/pull/8382

ncw:
Thanks! I've merged #8382 so I'll close this one.

PR #8366, opened 2025-02-10
bisync: fix listings missing concurrent modifications - fixes #8359

### LINKED ISSUE #8359: bisync: modtime of latest/updated file gets changed matching modtime of the previous revision

Using `bisync`, where the remote is Synology NAS SFTP, the modification time of file downloaded file is sometimes not updated properly on one of the nodes and eventually the "old" modtime gets propagated to all the nodes.

Note that the content / checksum of the file is correctly propagated as expected. The confusion comes from the fact that the modtime of the file ends up being the date of the previous revision of the file, and that is propagated to all the nodes.

After spending some time on this, I got back to a working behavior by changing the pattern to prevent pulling any conflict files and removing rclone generated conflict files on all my nodes. In other words, the bisync pattern I used was pulling conflict files wich would themselves conflict with other conflict files from other nodes. I didn't care much about those files as I could confirm the main file(s) were ok.

However, my understanding of the problem is that those conflicts would somehow interfere with the bisync operation in a way that made the moditfication time not be updated while the content of the file was.

I understand that bisync is in beta and I'm happy to report that the feature is useful to me.

```
# on amd64 Linux (Ubuntu)
$ stat ~/REDACTED
  File: /home/XXX/REDACTED
  Size: 98584     	Blocks: 200        IO Block: 4096   regular file
Device: 252,1	Inode: 16664833    Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/REDACTED)   Gid: ( 1000/REDACTED)
Access: 2025-02-03 01:50:02.254891607 -0500
Modify: 2025-01-31 23:57:40.000000000 -0500
Change: 2025-02-03 01:35:01.522276874 -0500
 Birth: 2025-02-03 01:35:01.503276785 -0500
```

```
# in termux (Android)
$ stat ~/REDACTED
  File: /data/data/com.termux/files/home/XXX/REDACTED
  Size: 98584     	Blocks: 200        IO Block: 4096   regular file
Device: 0,201	Inode: 83456       Links: 1
Access: (0660/-rw-rw----)  Uid: (10275/ u0_a275)   Gid: ( 1023/media_rw)
Context: u:object_r:fuse:s0
Access: 2025-02-03 00:38:52.000000000 -0500
Modify: 2025-01-31 23:57:40.483101109 -0500
Change: 2025-02-03 01:23:57.860059979 -0500
 Birth: -
```

```
rclone v1.69.0-termux
- os/version: unknown
- os/kernel: 5.15.123-android13-8-00033-g39606e3309fc-ab10974680 (aarch64)
- os/type: android
- go/version: go1.23.4
- go/linking: dynamic
- go/tags: noselfupdate
```

Note: I also use amd64 Linux clients. The client above was the one I notice that did not set proper modtime to the downloaded file. That bogus modtime somehow got propagated back.

arm64 Android, amd64 Linux

SFTP (Synology NAS)

```
rclone bisync REDACTED:PATH /data/data/com.termux/files/home/REDACTED --include /REDACTED* --compare size,modtime,checksum --resilient --recover --max-lock 2m --conflict-resolve newer --verbose
```
config:
```
[REDACTED]
type = sftp
host = REDACTED
user = REDACTED
port = REDACTED
pass = REDACTED
path_override = /volume1/
description = REDACTED
shell_type = unix
md5sum_command = none
sha1sum_command = none
```

I think I provided above insightful information on the problem. If you feel however you still need my help to reproduce the issue, I will take the time top reproduce the issue, this time with the log level you request as it seemed I could reproduce  the issue systematically.

#### How to use GitHub

nielash:

Before this change, there was a bug affecting listing files when:

- a given bisync run had changes in the 2to1 direction
AND
- the run had NO changes in the 1to2 direction
AND
- at least one of the changed files changed AGAIN during the run (specifically, after the initial march and before the transfers.)

In this situation, the listings on one side would still retain the prior version of the changed file, potentially causing conflicts or errors. 

This change fixes the issue by making sure that if we're updating the listings on one side, we must also update the other. (We previously tried to skip it for efficiency, but this failed to account for the possibility that a changed file could change again during the run.)

- https://forum.rclone.org/t/avoiding-bisync-times-are-equal-conflicts-with-newer/46636/32?u=nielash
- #8359 

#### Checklist


ncw (approved):
This is actually quite a small change but the test stuff sure does blow it up!

It looks good to me - thank you :-)

I'll schedule this for v1.69.1

Issue #8370, opened 2025-02-11
Rclone silently hides Dropbox Paper files (and other “non-downloadable” files)


vasi:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/rclone-silently-hides-dropbox-paper-files-and-other-non-downloadable-files/49996

#### What is the problem you are having with rclone?

When used with Dropbox, rclone is not showing "non-downloadable" files. These are files like Google Docs, Dropbox Paper docs, etc. They are documented in the [Dropbox API](https://developers.dropbox.com/dbx-file-access-guide#non-downloadable-files).

On other files, rclone works fine. For now, rclone can't actually _do_ anything with non-downloadable files, since they can't be copied as normal. But it should at least not hide them. I almost panicked when they were mysteriously missing!

Perhaps eventually we can make rclone "copy" these files via the export API, but let's consider that a separate question for the moment.

#### What is your rclone version (output from `rclone version`)

```
✗ rclone version
rclone v1.69.0
- os/version: linuxmint 22.1 (64 bit)
- os/kernel: 6.8.0-53-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.23.4
- go/linking: static
- go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

Linux Mint 22.1, 64-bit x86_64

#### Which cloud storage system are you using? (e.g. Google Drive)

Dropbox

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

Listing a directory that contains only a single Dropbox Paper file.

```
rclone lsf dropbox:rclonetest
```

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

```
$ rclone lsf dropbox:rclonetest
2025/02/11 00:01:42 DEBUG : rclone: Version "v1.69.0" starting with parameters ["rclone" "lsf" "-vv" "dropbox:rclonetest"]
2025/02/11 00:01:42 DEBUG : Creating backend with remote "dropbox:rclonetest"
2025/02/11 00:01:42 DEBUG : Using config file from "/home/vasi/.config/rclone/rclone.conf"
2025/02/11 00:01:42 DEBUG : 6 go routines active
2025/02/11 00:01:42 INFO  : Dropbox root 'rclonetest': Committing uploads - please wait...
```

#### Debugging

[...]

PR #8371, opened 2025-02-11
Dropbox exports

vasi:

Some files in Dropbox are considered ["non-downloadable"](https://developers.dropbox.com/dbx-file-access-guide#non-downloadable-files). Examples of such files are web links, and Dropbox Paper documents. One can't use the files/download API for these, instead one has to use the files/export API.

Previously, rclone just skipped these in listing, and failed to cat or copy or sync them. See #8370 

With this PR, such files are successfully listed, and can be cat'ted or copied or synced. It also adds a `--dropbox-export-formats` argument, so the user can choose their preferred format, like with the `drive` backend.

I file [this issue](https://github.com/rclone/rclone/issues/8370), and [this forum post](https://forum.rclone.org/t/rclone-silently-hides-dropbox-paper-files-and-other-non-downloadable-files/49996) earlier today. No discussion yet.

#### Checklist


docs/content/dropbox.md
```diff
@@ -492,6 +492,25 @@ Properties:
 - Type:        string
 - Required:    false
 
+#### --dropbox-export-formats
+
+Comma separated list of preferred formats for exporting files
```
ncw, line 497: This part of the file is auto generated from the `export_formats` definition, so put this nice help in there.

backend/dropbox/dropbox.go
```diff
@@ -247,7 +247,13 @@ folders.`,
 			Help:     "Specify a different Dropbox namespace ID to use as the root for all paths.",
 			Default:  "",
 			Advanced: true,
-		}}...), defaultBatcherOptions.FsOptions("For full info see [the main docs](https://rclone.org/dropbox/#batch-mode)\n\n")...),
+		}, {
+			Name:     "export_formats",
+			Help:     "Comma separated list of preferred formats for exporting files",
```
ncw, line 252: Put the help you've put in `docs/content/dropbox.md` in here.

backend/dropbox/dropbox.go
```diff
@@ -247,7 +247,13 @@ folders.`,
...
+			Default:  "",
```
ncw, line 253: Define the Default as `fs.CommaSepList{}` and the config parser will parse it for you into a `[]string`

backend/dropbox/dropbox.go
```diff
@@ -264,6 +270,7 @@ type Options struct {
 	PacerMinSleep fs.Duration          `config:"pacer_min_sleep"`
 	Enc           encoder.MultiEncoder `config:"encoding"`
 	RootNsid      string               `config:"root_namespace"`
+	ExportFormats string               `config:"export_formats"`
```
ncw, line 273: .. so fs.CommaSepList here instead of string

backend/dropbox/dropbox.go
```diff
@@ -1586,6 +1617,18 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 		return
 	}
 
+	if o.non_downloadable {
+		arg := files.ExportArg{Path: o.id, ExportFormat: o.export_format}
+		err = o.fs.pacer.Call(func() (bool, error) {
+			var exportResult *files.ExportResult
+			exportResult, in, err = o.fs.srv.Export(&arg)
+			o.bytes = int64(exportResult.ExportMetadata.Size)
```
ncw, line 1625: These should go after the `o.fs.pacer.Call` when you are sure it has suceeded otherwise you'll certainly case a `nil` pointer exception.

ncw (commented):
This looks very promising - thank you for doing this :-)

I think this should only show the files we have an export format for, otherwise they should be skipped.

> There is no support for changing the paths to add extensions, as drive does. It would add a lot of complexity in NewObject() and other places.

I suspect that is going to confuse the users - if you export a file called `MyDocument` as a markdown file called `MyDocument` then a lot of users will have no idea what to do with it.

I put some other comments inline and the linter found a few things too.

Thanks

docs/content/dropbox.md
vasi: Oh fascinating, I'll do that.

vasi:
> I think this should only show the files we have an export format for, otherwise they should be skipped.

I'm not quite catching your drift. Do you mean:
1. We should only list files that have _some_ `ExportAs` value?
2. We should only list files with formats on an allowlist? Eg; if Dropbox started adding "word" as an export format, we should ignore it until someone updates rclone?
3. We should only list files with formats the user explicitly mentioned in export_formats?

vasi:
> I suspect that is going to confuse the users - if you export a file called `MyDocument` as a markdown file called `MyDocument` then a lot of users will have no idea what to do with it.

So far, all exportable Dropbox files have fairly unusual extensions already: `.web` for a weblink, `.paper` for a Paper doc, `.papert` for a Paper template. So I was imagining users would just set `.paper` on their machine to open in the Markdown editor or browser. But it's true that might be confusing to users.

But let's say we do want to do extension mapping:
* The listing code should be pretty straightforward.
* My impression from reading the drive.go code is that NewObject will now have to start listing the parent directory, to find files with the right mapping. Or at least doing a get_metadata on a predicted base path. Is that correct? 
* I think that's the only changes to implementation we'd need. Am I missing anything?

ncw:
> > I suspect that is going to confuse the users - if you export a file called `MyDocument` as a markdown file called `MyDocument` then a lot of users will have no idea what to do with it.
> 
> So far, all exportable Dropbox files have fairly unusual extensions already: `.web` for a weblink, `.paper` for a Paper doc, `.papert` for a Paper template. So I was imagining users would just set `.paper` on their machine to open in the Markdown editor or browser. But it's true that might be confusing to users.

Ah, OK. If they have sensible extensions are good - that is a little less confusing.

> But let's say we do want to do extension mapping:
> 
> * The listing code should be pretty straightforward.
> * My impression from reading the drive.go code is that NewObject will now have to start listing the parent directory, to find files with the right mapping. Or at least doing a get_metadata on a predicted base path. Is that correct?

Yes it should try a `get_metadata`.

> * I think that's the only changes to implementation we'd need. Am I missing anything?

Do you think we'll ever be able to import these files back to dropbox? It might be worth thinking about that now.

> > I think this should only show the files we have an export format for, otherwise they should be skipped.
> 
> I'm not quite catching your drift. Do you mean:
> 
> 1. We should only list files that have _some_ `ExportAs` value?
> 2. We should only list files with formats on an allowlist? Eg; if Dropbox started adding "word" as an export format, we should ignore it until someone updates rclone?
> 3. We should only list files with formats the user explicitly mentioned in export_formats?

Sorry I wasn't very clear was I!

For the "non-downloadable" files we should only show them in listings if there is a method of exporting them. That means the user has to have configured the export formats and there be one for that kind of file. We set sensible default export formats for the google drive backend.

Judging by my experience with the Google drive backend people will want to skip downloading the "non-downloadable" files.

We have these two flags in the drive backend to control this:

      --drive-show-all-gdocs   Show all Google Docs including non-exportable ones in listings
      --drive-skip-gdocs       Skip google documents in all listings

vasi:
Ok, I've addressed all lints and comments. Thank you for review!

As requested:
* We now hide files with no known export format. Only files exportable to markdown or HTML are visible.
* We now expose files with an added extension (".md" or ".html"). This means we have to look up both the with-extension and without-extension filenames, but we can do it in parallel so it shouldn't greatly impact efficiency.

I've also removed web-links from this PR. The "url" export format is a JSON object containing a `url` field, which I don't think corresponds to any real-life file format or extension. It's plausible that we later add an internal-to-rclone conversion of this to eg: desktop files, but let's make that a follow up so as not to overly complicate this PR.

I'm a bit uncomfortable about one aspect of adding extensions. It's possible (though a bit odd) for a user to have an actual Dropbox file named "foo.paper.md", as well as a Paper doc named "foo.paper", in which case the real-file will hide the existence of the Paper doc. Has this sort of thing been a problem with Drive?

vasi:
Woops, our comments overlapped a bit. I'll try to address the interesting bits.

> Do you think we'll ever be able to import these files back to dropbox? It might be worth thinking about that now.

I was wrong that this didn't exist! We have APIs [/files/paper/create](https://www.dropbox.com/developers/documentation/http/documentation#files-paper-create) and /files/paper/update that let us do this.

There are a several awkward bits:
* The filename _must_ end in `.paper`, so we'd have to munge some names.
* Conversion can be lossy, even across an export+create round-trip.
* It's not unusual for users to have .md or .html files as normal-files in Dropbox, so we'd have to be extremely careful not to convert these implicitly.
* Some users may still be on the old pre-2020 Paper API, I myself was just moved to the new API last month.
* I have no idea how embedded images are supposed to work with these API.

Anyhow: Do you think we should integrate this into the same PR, or save it for a follow-up?

> For the "non-downloadable" files we should only show them in listings if there is a method of exporting them. That means the user has to have configured the export formats and there be one for that kind of file. 

Ah, that makes sense. Do you think it's reasonable to keep the default `html,markdown`, but then allow the user to set it to empty-string to indicate that they don't want to sync these files? Or do we need extra flags like Drive has?

ncw:
> As requested:
> 
> * We now hide files with no known export format. Only files exportable to markdown or HTML are visible.
> * We now expose files with an added extension (".md" or ".html"). This means we have to look up both the with-extension and without-extension filenames, but we can do it in parallel so it shouldn't greatly impact efficiency.

Great

> I've also removed web-links from this PR. The "url" export format is a JSON object containing a `url` field, which I don't think corresponds to any real-life file format or extension. It's plausible that we later add an internal-to-rclone conversion of this to eg: desktop files, but let's make that a follow up so as not to overly complicate this PR.

OK

> I'm a bit uncomfortable about one aspect of adding extensions. It's possible (though a bit odd) for a user to have an actual Dropbox file named "foo.paper.md", as well as a Paper doc named "foo.paper", in which case the real-file will hide the existence of the Paper doc. Has this sort of thing been a problem with Drive?

This hasn't really been a problem with google drive. In Drive the files generally don't have extensions, so "Document" downloads to "Document.xlsx" and I can't remember any complaints about people having both "Document" and "Document.xlsx" files which sounds more likely than "foo.paper" and "foo.paper.md".

> > Do you think we'll ever be able to import these files back to dropbox? It might be worth thinking about that now.
> 
> I was wrong that this didn't exist! We have APIs [/files/paper/create](https://www.dropbox.com/developers/documentation/http/documentation#files-paper-create) and /files/paper/update that let us do this.
> 
> There are a several awkward bits:
> 
> * The filename _must_ end in `.paper`, so we'd have to munge some names.

That might be an argument for removing the `.paper` extension.  So we get

"document.paper" on dropbox and "document.md" on the local disk. That looks more usable than asking the user to name their documents "document.paper.md".

We can't have different names for import and export

If we do "document.paper" (dropbox) <-> "document.paper.md" (local) that works fine, but if we have "document.md" (local) -> "document.paper" (dropbox) -> "document.paper.md" (local) then that is going to cause trouble.

> * Conversion can be lossy, even across an export+create round-trip.

That is the same with google drive. Fine if we document it.

> * It's not unusual for users to have .md or .html files as normal-files in Dropbox, so we'd have to be extremely careful not to convert these implicitly.

In google drive, importing files is set with 

      --drive-import-formats string      Comma separated list of preferred formats for uploading Google docs

And this is blank by default so no documents are uploaded by default. I think this needs to be a deliberate choice by the user otherwise we will have the trouble you are referring to.

> * Some users may still be on the old pre-2020 Paper API, I myself was just moved to the new API last month.

Not sure of the consequences of this.

> * I have no idea how embedded images are supposed to work with these API.
> 
> Anyhow: Do you think we should integrate this into the same PR, or save it for a follow-up?

I think a follow up, but I thought it would be worth thinking about it now.

> > For the "non-downloadable" files we should only show them in listings if there is a method of exporting them. That means the user has to have configured the export formats and there be one for that kind of file.
> 
> Ah, that makes sense. Do you think it's reasonable to keep the default `html,markdown`, but then allow the user to set it to empty-string to indicate that they don't want to sync these files? Or do we need extra flags like Drive has?

I think the default `html,markdown` is fine. The google drive implementation will choose an export format for you if one isn't found in the export list. Should we be doing that too? That makes the `--drive-skip-gdocs` flag necessary though.

vasi:
> This hasn't really been a problem with google drive. In Drive the files generally don't have extensions, so "Document" downloads to "Document.xlsx" and I can't remember any complaints about people having both "Document" and "Document.xlsx" files which sounds more likely than "foo.paper" and "foo.paper.md".

Perfect, I won't worry about that then.

> That might be an argument for removing the `.paper` extension.

Yes, your argument is convincing!

> > * Some users may still be on the old pre-2020 Paper API, I myself was just moved to the new API last month.
> Not sure of the consequences of this.

Those users won't see any Paper docs in their Dropbox files, any Paper docs they have are in a weird separate storage. And I don't think they can import Paper docs either, but I currently have no way of testing that! I don't think it matters for now, but we'll have to think about it when we implement import.

> I think the default `html,markdown` is fine. The google drive implementation will choose an export format for you if one isn't found in the export list. Should we be doing that too? That makes the `--drive-skip-gdocs` flag necessary though.

Hmm. I do think consistency with Drive would be useful. My understanding is that would give us the following flags for this PR:
* `--dropbox-skip-exports`: Completely ignore non-downloadable files
* `--dropbox-show-all-exports`: Show all non-downloadable files, but never download them
* `--dropbox-export-formats=""`: Pick a default export format for non-download files (at least when it's a format we know how to deal with)
* `--dropbox-export-formats="..."`: Pick formats from the list, in order. If none are supported for a particular file, then use a default for that file

This feels a bit complex, but the argument for consistency is strong.

Current TODOs for me:
[ ] Get `.paper` extensions removed
[ ] Look into the Windows build failure, though I'm pretty sure it's spurious
[ ] See how it feels to implement the Drive-like flags
[ ] Make issues for "support weblinks" and "support import", to at least collect our thoughts.

vasi:
Ok, all of the above is done.

* This PR now supports Drive-like options `export_formats`, `show_all_exports` and `skip_exports`
    * For consistency with the Drive options, export_formats now takes file extension (like `md`) rather than the named formats the Dropbox API uses (like `markdown`)* Paper files (and Paper Templates) now lose their .paper/.papert extensions when being exposed as .html/.md
* I wasn't able to reproduce the Windows build failure when running a [build on my fork](https://github.com/vasi/rclone/actions/runs/13317520322/job/37195054843). I'm assuming for now that it's spurious.
* Issues are created for the follow-up PRs, with some implementation hints: #8391  and #8392
    * One thing to consider from the import issue: There's multiple document types that map to the same extension! When someone imports foo.md, do we create foo.paper or foo.papert?
    * My instinct is to just always prefer foo.paper, since templates aren't super common. But we probably want to hash this out now, since some solutions could involve changing our name-mapping from this PR.
* I abandoned looking into metadata for now. An option like `--include=*.md` provided to the sync command is good enough for me when doing bulk-backups of my Paper docs.


Adding all these features didn't complicate things too much. The only parts that feel a bit suspect to me are:
* The "API format" vs "extension" split feels a bit complex. I defined separate types for these, so at least it's easy to keep them straight in the code.
* Finding the Dropbox-API-path that corresponds to a filename is harder than I'd like. A `foo.md` could come from a literal file `foo.md`, a Paper `foo.paper`, or a template `foo.papert`. (Or even some new non-exportable format that shows up and can export markdown, named just `foo`.) So we have to check all of these!

Some further notes:
* I tested server-side copies in `show_all_exports` mode, and they work!
* It turns out that Dropbox doesn't update the FileMetadata quickly for Paper docs, perhaps only every 30 minutes. So if I sync, make a change, then sync again, my change won't be recorded since the modification-time will be identical. We probably want to document this, and also see if it's true about other file types in Dropbox.
* Has anybody ever asked for multiple export formats of the same doc in Drive? Dropbox's HTML and markdown exports are both _slightly_ lossy, in different ways, so for backup purposes it'd be nice to have both. I suppose multiple `sync` commands isn't that much trouble though.

ncw (commented):
I gave the code a go through - it is looking nice :-)

It would be nice to have some tests (I don't know if you've tried running the dropbox integration tests `go test -v -remote yourdrobpoxremote:`, add `-short` for a bit less stuff.

The ones from the drive backend are here

https://github.com/rclone/rclone/blob/6560ea9bdc515dd4f611f81b501f08899a5589a0/backend/drive/drive_internal_test.go#L646-L659

Though these seem to depend on importing a document...

ncw:
> Those users won't see any Paper docs in their Dropbox files, any Paper docs they have are in a weird separate storage. And I don't think they can import Paper docs either, but I currently have no way of testing that! I don't think it matters for now, but we'll have to think about it when we implement import.

That sounds OK.

> Hmm. I do think consistency with Drive would be useful. My understanding is that would give us the following flags for this PR:
> 
> * `--dropbox-skip-exports`: Completely ignore non-downloadable files
> * `--dropbox-show-all-exports`: Show all non-downloadable files, but never download them
> * `--dropbox-export-formats=""`: Pick a default export format for non-download files (at least when it's a format we know how to deal with)
> * `--dropbox-export-formats="..."`: Pick formats from the list, in order. If none are supported for a particular file, then use a default for that file
> 
> This feels a bit complex, but the argument for consistency is strong.

You've renamed `docs` to `exports` in these flags. I tend to think of all the Google exportable files as google docs which includes documents, sheets, presentations etc, but I've noticed that terminology confuses people I use it on, so `exports` is better and consistent with the other flags which is nice.

Probably the most esoteric flag is `dropbox-show-all-exports` but as you noted below it is very useful for server side copies.

> Ok, all of the above is done.

Great work :-)

> * This PR now supports Drive-like options `export_formats`, `show_all_exports` and `skip_exports`

Nice!
   
>   * For consistency with the Drive options, export_formats now takes file extension (like `md`) rather than the named formats the Dropbox API uses (like `markdown`)* Paper files (and Paper Templates) now lose their .paper/.papert extensions when being exposed as .html/.md

:+1:

> * I wasn't able to reproduce the Windows build failure when running a [build on my fork](https://github.com/vasi/rclone/actions/runs/13317520322/job/37195054843). I'm assuming for now that it's spurious.

If it happens again I'll look at it, but there are various flakes that turn up in the rclone tests. Every now and again I go on a purge to get rid of them!

> * Issues are created for the follow-up PRs, with some implementation hints: [Dropbox: Support weblinks #8391](https://github.com/rclone/rclone/issues/8391)  and [Dropbox: Support import of Paper documents #8392](https://github.com/rclone/rclone/issues/8392)
>   
>   * One thing to consider from the import issue: There's multiple document types that map to the same extension! When someone imports foo.md, do we create foo.paper or foo.papert?

I don't think we have that problem in drive.

>   * My instinct is to just always prefer foo.paper, since templates aren't super common. But we probably want to hash this out now, since some solutions could involve changing our name-mapping from this PR.

We could have a ".template.md" extension? That would solve the problem.

> * I abandoned looking into metadata for now. An option like `--include=*.md` provided to the sync command is good enough for me when doing bulk-backups of my Paper docs.

Great

> Adding all these features didn't complicate things too much. The only parts that feel a bit suspect to me are:
> 
> * The "API format" vs "extension" split feels a bit complex. I defined separate types for these, so at least it's easy to keep them straight in the code.

In google drive we fetch these from the server

> * Finding the Dropbox-API-path that corresponds to a filename is harder than I'd like. A `foo.md` could come from a literal file `foo.md`, a Paper `foo.paper`, or a template `foo.papert`. (Or even some new non-exportable format that shows up and can export markdown, named just `foo`.) So we have to check all of these!

:-( As long as we try not to slow down the common case too much.
 
> Some further notes:
> 
> * I tested server-side copies in `show_all_exports` mode, and they work!

:-)

> * It turns out that Dropbox doesn't update the FileMetadata quickly for Paper docs, perhaps only every 30 minutes. So if I sync, make a change, then sync again, my change won't be recorded since the modification-time will be identical. We probably want to document this, and also see if it's true about other file types in Dropbox.

Seems reasonable.

> * Has anybody ever asked for multiple export formats of the same doc in Drive? Dropbox's HTML and markdown exports are both _slightly_ lossy, in different ways, so for backup purposes it'd be nice to have both. I suppose multiple `sync` commands isn't that much trouble though.

No. I think just run the sync twice is fine.

I gave the code a run through - looking good! I forgot to say that this needs some docs too.

vasi:
> We could have a ".template.md" extension? That would solve the problem.

Or even `.papert.md`?
 
> > * The "API format" vs "extension" split feels a bit complex. I defined separate types for these, so at least it's easy to keep them straight in the code.
> In google drive we fetch these from the server

Unfortunately Dropbox has no equivalent of Drive's `About` call, we can only figure this out on a per-file basis. And it doesn't use something standard-ish like MIME types, instead we just have to _know_ that the format Dropbox calls `markdown` has the extension `.md`. :disappointed:  

> > A `foo.md` could come from a literal file `foo.md`, a Paper `foo.paper`, or a template `foo.papert`.So we have to check all of these!
> 
> :-( As long as we try not to slow down the common case too much.

We check for them in parallel, so it shouldn't slow anything down unless we're hitting API rate limits. https://github.com/rclone/rclone/pull/8371/files#diff-7070c850317ae70144b7c0544c568a183ab48d05f0210e96160c3ecf3746ff22R691-R769

TODOs for me:
- [ ] Fix nits:
    - [ ] My "Google Docs" copy-paste error
    - [ ] The bad variable name `knownAPIExportFormats`
- [ ] Try running the backend tests, thanks for clarifying how to do that!
    - [ ] Look into adding new tests. Maybe I'll do an import here, but just with non-generalized code?
- [ ] Document how exports work for Dropbox.
- [ ] Look into whether slow updates of modTime happen for other Dropbox files. Depending on the result, decide where to document the slow updates for Paper.
-

ncw:
> > We could have a ".template.md" extension? That would solve the problem.
> 
> Or even `.papert.md`?

Sounds good!

>   Look into adding new tests. Maybe I'll do an import here, but just with non-generalized code?

If it isn't too hard then that is probably the best test and it will be learning for the full import.

vasi:
So I ran the tests, and one failed. But it also fails for me on master! It's `TestIntegration/FsMkdir/FsPutFiles/PublicLink`. I'll investigate a bit, but would love to know if this works for other people.

vasi:
Test failure turned out to be unrelated, fix in #8397

vasi:
Ok, all the above tasks are done. Nits fixed, a simple test written, docs written.

I could try to skip the test if the user is on the pre-2020 Paper API and is not allowed to create Paper docs? But maybe such users should just create a new test account.

ncw (approved):
This looks perfect now, thank you .

I'll just run the ci 

Anything else to add or shall I merge?

vasi:
Nothing else on my part. Do you need me to squash commits so we only have one with a message in the house style?

ncw:
> Nothing else on my part. Do you need me to squash commits so we only have one with a message in the house style?

I'll do that now - thank you very much for a great feature :-)

Issue #8383, opened 2025-02-13
Back Blaze Multi threaded upload will have a recoverable error - fails anyway


zackees:
Continuation of:

https://github.com/rclone/rclone/issues/8315#issuecomment-2655173223

PR for fix: https://github.com/rclone/rclone/pull/8384

Almost like clockwork, rclone will fail at 3% for this file below. Even when the b2 concurrency value is set to just one.

@ncw mentioned that it looks like a proxy error with the HTML being sent back as an error code, or vice versa.

Looking at this logs, retry=False was being triggered for this error, borking the entire multi-part download consisting of 4k upload parts. Yes, if any one of those upload parts fails then the entire thing fails.

The reason is this: low-level retries are ONLY enabled if and only if, the error type and message can evaluate for true in list of predicate that will be run on the error. If no rule returns true, then the retries=False is returned.

This means that *any* unexpected error not covered by the white list will bypass the low level retries. That's what happened in the B2 case.

My fix does not address the bigger picture - it only adds this specific error type to the white list.

It may be useful to consider that retry=True be returned for any 500 "internal error" status code returned for an upload part.

```
650 failed: multi-thread copy: failed to write chunk: Post "https://pod-020-3011-01.backblaze.com/b2api/v1/b2_upload_part/4_zb87219cbda6098ab96340e17_f228491889785e70a_d20250212_m204420_c002_v0203011_t0019_u01739393060115/0054": multi-thread copy: failed to write chunk: invalid character '<' looking for beginning of value
2|run2  | Transferred:              4.402 GiB / 7.901 TiB, 0%, 8.663 MiB/s, ETA 1w4d1h
2|run2  | Checks:                 3 / 3, 100%, Listed 312
2|run2  | Transferred:            0 / 22, 0%
2|run2  | Elapsed time:      3m48.3s
2|run2  | Transferring:
2|run2  |  * zlib2-data/pilimi-zlib2-15000000-15679999.tar:  1% /364.036Gi, 8.561Mi/s, 11h
2|run2  | Transferred:              4.402 GiB / 7.901 TiB, 0%, 8.663 MiB/s, ETA 1w4d1h
2|run2  | Checks:                 3 / 3, 100%, Listed 312
2|run2  | Transferred:            0 / 22, 0%
2|run2  | Elapsed time:      3m48.5s
2|run2  | Transferring:
[...]

PR #8384, opened 2025-02-13
Allows B2 Retry on Recoverable error

zackees:
Fixes https://github.com/rclone/rclone/issues/8383

Tested on `Ubuntu 24.10`

This turns out to be a very major bug on the B2 backend.

This patch definitely fixes it as I see that retry=True is now set for the error when it comes. 

The next (recoverable) error that jumps out is this:

`no tomes available`

Which led me to this:

https://forum.duplicati.com/t/upload-to-backblaze-b2-no-tomes-available-connection-reset-by-peer/16621

Which led me to this:

https://www.backblaze.com/blog/b2-503-500-server-error/

## In a nutshell

B2 API is supposed to occasionally give 500 internal error codes. It will resolve with additional retries. Since Rclone doesn't do a retry for this specific type of error, it fails.

This error is the *probably* the reason my datalakes has been failing the src->dst file audits in every one of the data repos. I'll know more next week.

### DIFF
--- a/fs/fserrors/error.go
+++ b/fs/fserrors/error.go
@@ -378,14 +378,15 @@ func Cause(cause error) (retriable bool, err error) {
 // This is incredibly ugly - if only errors.Cause worked for all
 // errors and all errors were exported from the stdlib.
 var retriableErrorStrings = []string{
-	"use of closed network connection", // internal/poll/fd.go
-	"unexpected EOF reading trailer",   // net/http/transfer.go
-	"transport connection broken",      // net/http/transport.go
-	"http: ContentLength=",             // net/http/transfer.go
-	"server closed idle connection",    // net/http/transport.go
-	"bad record MAC",                   // crypto/tls/alert.go
-	"stream error:",                    // net/http/h2_bundle.go
-	"tls: use of closed connection",    // crypto/tls/conn.go
+	"use of closed network connection",                     // internal/poll/fd.go
+	"unexpected EOF reading trailer",                       // net/http/transfer.go
+	"transport connection broken",                          // net/http/transport.go
+	"http: ContentLength=",                                 // net/http/transfer.go
+	"server closed idle connection",                        // net/http/transport.go
+	"bad record MAC",                                       // crypto/tls/alert.go
+	"stream error:",                                        // net/http/h2_bundle.go
+	"tls: use of closed connection",                        // crypto/tls/conn.go
+	"invalid character '<' looking for beginning of value", // From Zach Vorhies
 }
 
 // Errors which indicate networking errors which should be retried


zackees:
This is definitely working now. I see that retry=True is now set for the error when it comes. 

The next error that jumps out is this:

`no tomes available`

Which led me to this:

https://forum.duplicati.com/t/upload-to-backblaze-b2-no-tomes-available-connection-reset-by-peer/16621

Which led me to this:

https://www.backblaze.com/blog/b2-503-500-server-error/

## In a nutshell

B2 knows they will give a false error. B2 advises to just retry. Since Rclone doesn't do a retry for this specific type of error, it fails.

This error is the reason why I've been dealing with all kinds of missing files throughout all my downloaders.

ncw:
Great analysis. I don't think this is the quite the right fix though.

The fix needs to go in this function in the b2 backend

https://github.com/rclone/rclone/blob/e27bf8b738b79f31afa7e938f069565710e42875/backend/b2/b2.go#L391-L393

The actual error you are seeing broken down into parts is this

```
Error sending chunk 25 (retry=false):
Post "https://pod-020-3011-15.backblaze.com/b2api/v1/b2_upload_part/4_zb87219cbda6098ab96340e17_f228491889785e70a_d20250212_m204420_c002_v0203011_t0019_u01739393060115/0045":
multi-thread copy: failed to write chunk:
invalid character '<' looking for beginning of value:
&url.Error{Op:"Post", URL:"https://pod-020-3011-15.backblaze.com/b2api/v1/b2_upload_part/4_zb87219cbda6098ab96340e17_f228491889785e70a_d20250212_m204420_c002_v0203011_t0019_u01739393060115/0045", Err:(*fmt.wrapError)(0xc007674740)}
```

This is from `encoding/json` and I think it is probably from this line

https://github.com/rclone/rclone/blob/e27bf8b738b79f31afa7e938f069565710e42875/backend/b2/upload.go#L300

We need some more info to fix properly.

Can you run your transfer with `-vv --dump responses`. This will make a lot of log, but if you can find the HTTP RESPONSE before the error and post that and the corresponding REQUEST (match up the hex IDs) then we can work out exactly what the best fix is.

BTW Rclone should already be retrying 5xx error codes

https://github.com/rclone/rclone/blob/e27bf8b738b79f31afa7e938f069565710e42875/backend/b2/b2.go#L362-L364

zackees:
A thoroughly broken B2 pipeline seems to be important enough to let this one line banger in as part of the set of string matching lines you are using for exactly this purpose. It's tested at scale. A broken B2 pipeline is serious.

Here's the challenge i'm facing - rclone is crashing before I can get the response payloads of the degenerate case. Your --dump is inlining binary data. Maybe I'm not smart enough to figure out why the process crashes when i `| tee` out the log file.

All we really need is logging for the case of `retries=False` is returned from low level retries predicate. That means the status code + a truncated response content when retries=False is returned.

Then I'll inspect again for the specific response pattern that is triggering a bypass of the low-level-retries algorithm.

ncw:
Try just `--dump headers` that shouldn't make binary in the log and may well be enough to find the problem.

zackees:
> Try just `--dump headers` that shouldn't make binary in the log and may well be enough to find the problem.

Done. I was able to repro the error and sent the compressed log files to your inbox.

ncw:
I isolated the request causing the problem - I'll reply on the issue  #8383

PR #8386, opened 2025-02-13
Sqlite

marcelowa:

Add a new backend: SQLite

#### Checklist

- [x ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).


ncw:
This is a fun idea!

Unfortunately there is a problem with github.com/mattn/go-sqlite3 as it requires CGO. We build rclone without CGO support to maximise portability.

It does get built with CGO support on Windows and macOS to use `libfuse`. However we don't link the rclone binary against libfuse which means as long as you don't run `rclone mount` you don't need to install fuse.

It looks like `go-sqlite3` includes the sqlite code within it - it doesn't seem to be using the shared library which is good.

There are alternatives to github.com/mattn/go-sqlite3

- https://github.com/cvilsmeier/sqinn-go which uses an sqlite subprocess
- https://pkg.go.dev/modernc.org/sqlite which has translated the SQLite C source to Go automatically

However rclone does include a very powerful built in database [lib/kv](https://pkg.go.dev/github.com/rclone/rclone@v1.69.0/lib/kv) which is based on https://pkg.go.dev/go.etcd.io/bbolt

If you re-worked your db to use that package then it would be easy to merge.

Another idea is you keep this as a plugin: https://pkg.go.dev/github.com/rclone/rclone@v1.69.0/lib/plugin

I'll run the CI now and we can see what blows up.

darthShadow:
https://github.com/ncruces/go-sqlite3 is also an option. It uses the WASM build of sqlite.

marcelowa:
@ncw thanks for the great feedback!
Although "cvilsmeier/sqinn-go" doesn't look like an active project (last commit in 2023), "modernc.org/sqlite" and "ncruces/go-sqlite3" which @darthShadow  suggested look like viable options, I'll check both of them!
I also saw that the lint didn't pass with small issues to fix, I'll fix the issues

marcelowa:
I ended up using "ncruces/go-sqlite3" because "modernc.org/sqlite" didn't behave well

marcelowa:
@ncw can you please run ci again?
Thanks

ncw:
@marcelowa shall we close this in favour of #8445 ?

pyhedgehog:
bolt is go-only format. It's not equal replacement for sqlite3 that is supported worldwide.

PR #8390, opened 2025-02-13
adds logging during failure condition for low-level-retry

zackees:
This is extra logging to inspect the low level retry failure on the b2 backend when using `-v`

See also:
https://github.com/rclone/rclone/issues/8383


ncw:
I'm not sure we need this now we have a log with `--dump headers`

zackees:
My response is that -v is common and logging retry failures is very very important. When a low level retry fails unexpectedly the upload will fail. The larger the upload the higher chance it has to fail.

But hey, it's your code base not mine.

PR #8406, opened 2025-02-19
build: cache go build & test

darthShadow:

Some more CI caching changes, this time for the go builds and tests.

N/A

#### Checklist


darthShadow:
Timings can be seen here: https://github.com/rclone/rclone/actions?query=branch%3Afeat%2Fcache-go-build-test

Android builds have gone down from ~8m to ~2m.
The main `build.yml` builds have gone down from ~24m to ~4m.

.github/workflows/build.yml
```diff
@@ -168,44 +219,95 @@ jobs:
...
+      - name: Print rclone Version
         shell: bash
         run: |
           rclone version
 
-      - name: Run tests
+      - name: Update mtime
+        shell: bash
+        run: |
+          go run ./bin/go-test-cache/main.go
```
darthShadow, line 268: This is necessary because Golang caches the tests based on the modified timestamps of the file.

Reference: https://github.com/golang/go/issues/58571

It would be cleaner to do it with the newly added GOCACHEPROG but there's no compatible implementation for Github Actions Cache with that yet.

Makefile
```diff
@@ -88,13 +88,13 @@ test:	rclone test_all
 
 # Quick test
 quicktest:
-	RCLONE_CONFIG="/notfound" go test $(LDFLAGS) $(BUILDTAGS) ./...
+	RCLONE_CONFIG="/notfound" go test $(BUILDTAGS) ./...
```
darthShadow, line 91: This is necessary because Golang reuses the tests cache only if the flags remain the same and having the dynamic version as part of the flags was causing the cache to never be re-used.

.github/workflows/build.yml
```diff
@@ -168,44 +219,95 @@ jobs:
...
+            ${{ env.GOMODCACHE }}
+          key: ${{ env.CACHE_PREFIX }}-modcache-${{ hashFiles('**/go.mod') }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ env.CACHE_PREFIX }}-modcache
+
+      # Both load & update the cache when on default branch
+      - name: Load Go Build & Test Cache
+        id: go-cache
+        uses: actions/cache@v4
+        if: github.ref_name == github.event.repository.default_branch && github.event_name != 'pull_request'
```
darthShadow, line 235: This ensures that we both restore & update the caches on the main branch but only restore it on the other branches or pull requests.

Makefile
ncw: Nice. Can you do the same thing to the other tests in the makefile which are next to this one.

darthShadow: I did apply the same changes to the `racequicktest` & `compiletest` too below. I left `test_all` as that's not run on the Github CI anyway. I don't think I missed anything?

darthShadow:
Any interest in merging this for 1.70? I don't want to go through the effort of rebasing and testing it, just to let it go stale again.

Issue #8298, opened 2025-01-09
http: rclone rcd name from `content-disposition` header


hiddenmarten:
#### The associated forum post URL from `https://forum.rclone.org`
I haven't found any topic in the forum exactly on this matter.
Similar topic was raised [here](https://github.com/rclone/rclone/pull/5485), but it's about `copyurl`, while my issue is about `rcd`. 

#### What is your current rclone version (output from `rclone version`)?
1.68.2 (checked on master branch)

#### What problem are you are trying to solve?
Link to file:
https://drive.google.com/uc?export=download&id=1FoZaCXy3ibXMfrRxwVlnMGPnbVy9-dmI

Create http configuration file:
```bash
curl -X POST http://localhost:5572/config/create \
-H "Content-Type: application/json" \
-d '
{
  "name": "http-no-esc",
  "type": "http",
  "parameters": {
    "url": "https://drive.google.com",
    "no_escape": true
  }
}'
```

Send a stat request:
```bash
curl -X POST http://localhost:5572/operations/stat \
-H "Content-Type: application/json" \
-d '
{
  "fs": "http-no-esc:",
  "remote": "uc?export=download&id=1FoZaCXy3ibXMfrRxwVlnMGPnbVy9-dmI"
}'
```

Result (attention to the Path and Name fields):
```json
{
        "item": {
                "Path": "uc?export=download\u0026id=1FoZaCXy3ibXMfrRxwVlnMGPnbVy9-dmI",
                "Name": "uc?export=download\u0026id=1FoZaCXy3ibXMfrRxwVlnMGPnbVy9-dmI",
                "Size": 76966,
                "MimeType": "image/jpeg",
                "ModTime": "2024-10-21T10:15:58Z",
                "IsDir": false
        }
}
```

The actual name provided by header: `content-disposition: attachment; filename="test_img.jpg"`,
but it seems that this header is not being respected by the rclone in rcd mode.

#### How do you think rclone should be changed to solve that?
I suppose the rclone should have some kind of "smart mode" to recognize the name of the file from such header. 
I'm not sure should it be enabled ot not by default, but from my perspective it should be. 

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

[...]

PR #8413, opened 2025-02-22
http: metadata, content-disposition

hiddenmarten:

The objective is to extract the file name from the content-disposition header. This name should replace the one in stat, and a content-disposition handler should be provided via serve.

https://github.com/rclone/rclone/issues/8298

#### Checklist


hiddenmarten:
Hi @ncw!

May I get a review, please?

hiddenmarten:
Bump @ncw .
:)

ncw (commented):
Sorry for the delay reviewing.

Adding metadata to the HTTP backend seems like a good idea, though if we were starting from that then there are lot of other things we could add other than Content-Disposition.

I don't like the changes in lsjson/operations - any parsing should be done in the HTTP backend. So set another metadata key, eg `content-disposition-filename`. I don't think any changes should be needed to lsjson/operations.

The http serve changes are interesting, but again maybe should support more metadata than just content-disposition.

A minimum set of metadata for the http backend and server might be something like

- cache-control
- content-disposition
- content-encoding
- content-language
- content-type

What do you think?

hiddenmarten:
Just pulled an actual master branch and rebased on it, sorry.

hiddenmarten:
Hi @ncw,

The changes are ready for review. Here’s what’s been done:

- Updated field names to match the S3 backend. (This change can be reverted if necessary.)
- Added system metadata fields, as per your suggestion.
- Reverted the modifications in `lsjson/operations` and updated the `Remote()` method to use the `contentDispositionFilename` when available.

Please let me know if you need any further improvements or changes.

hiddenmarten:
@ncw ping :)

hiddenmarten:
@ncw ping :)

PR #8426, opened 2025-03-03
Documentation - Add note for using `rclone cat` for slicing out a byte range from a file, embedded in the copy-to documentation.

zackees:
Knowing that cat is the tool for byte slicing would have saved me a lot of time from emulating this via mount with caching minimum and direct-io.

Updating the docs with this bread-crumb so that users in the future can easily discover that `cat --offset X --count Y` is the proper sub command to get a byte slice out of large remote file.


ncw (approved):
Thank you - very useful for those who come after :-)

PR #8429, opened 2025-03-06
fix: vfs dir crash

izouxv:
(cherry picked from commit 5fcfe04fc167a7c707184c6f92422c360622f1ef)

crash fix.


darthShadow:
Please specify more details regarding your usage of this and exactly how it crashed and possibly a testcase to reproduce the crash.
I don't see any way `d.cleanupTimer` can be nil at that point in the code.

Issue #6585, opened 2022-11-27
Backblaze B2 Native API Server-Side Encryption


pl4nty:
#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.60.1
- os/version: alpine 3.16.3 (64 bit)
- os/kernel: 5.15.68.1-microsoft-standard-WSL2 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.19.3
- go/linking: static
- go/tags: none
```

#### What problem are you are trying to solve?

Backblaze B2 [server-side encryption](https://www.backblaze.com/b2/docs/server_side_encryption.html) is currently available in rclone via the S3 API, but S3 uses MD5 which prevents the use of `--checksum` with SHA1 remotes (eg OneDrive personal).

#### How do you think rclone should be changed to solve that?

The B2 Native API uses SHA1 and supports server-side encryption, so implementing it in rclone would allow the above combination of features.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

pl4nty:
I'm new to both Go and rclone, but happy to attempt implementing this. Might take a few months though, due to the holiday season and other projects

ncw:
@pl4nty if you'd like to have a go that would be great. I skimmed the docs page and it looks like it follows quite closesly the S3 implementation, so I would copy the s3 options over into the b2 backend and then plumb them in to the API calls.

I don't think it should be too tricky.

Go is an easy language to pick up (it was designed to be easy to pick up). If you are an experienced dev then [the go tour](https://go.dev/tour/welcome/1) will get you up to speed and it doesn't take very long.

pl4nty:
The deprecation of SHA1 for OneDrive will make my usecase redundant, but I'm still interested in implementing this. Will just be a lower priority

PR #8436, opened 2025-03-08
b2: Add Server-Side encryption support

fries1234:

This commit adds SSE-C (Server-Side Encryption - Customer) support to the B2 native backend. The server uses a customer provided AES-256 key to encrypt the files when you upload them to the bucket, and then it discards your key from the servers RAM after you're done uploading.

The option names and descriptions are based off the S3 backend implementation as the way S3 and B2 does SSE-C is pretty similar.

I saw it discussed in #6585.

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -8,7 +8,9 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/md5"
 	"crypto/sha1"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


fries1234:
just a thought: I wonder if the Base64 string getting converted into a Go UTF-8 string changes the data a tiny bit messing up the md5 hash compared to directly decoding the base64 string and hashing from the binary? I used openssl to generate 32 random bytes for an AES-256 key, which isn't exactly a UTF-8 compatible string.

fries1234:
All though thinking about it again, that might break compatibility the S3 compatible B2 API (even if rclone doesn't officially support B2 as a provider in its s3 backend, it still does work if you put the provider as AWS).

fries1234:
bump 2.0

fries1234:
bump 3.0

backend/b2/b2.go
```diff
@@ -1395,14 +1467,25 @@ func (f *Fs) copy(ctx context.Context, dstObj *Object, srcObj *Object, newInfo *
 	}
 
 	opts := rest.Opts{
-		Method: "POST",
-		Path:   "/b2_copy_file",
+		Method:       "POST",
+		Path:         "/b2_copy_file",
+		ExtraHeaders: map[string]string{},
```
ncw, line 1472: This doesn't appear to be used?

ncw (commented):
This looks very neat and well done - thank you.

Have you tried running the integration tests with SSE set? That might throw some more things up.

I put one very minor thing inline!

fries1234:
I just ran the integration tests with SSE enabled and they seem to pass!

fries1234:
bump

roucc (approved):
Thanks for doing this. Sorry this was left behind!

PR #8450, opened 2025-03-14
feat: add support for password commands in crypt backend

hrntknr:

crypt backend to allow passwords to be retrieved from the command. This means that passwords can be dynamically retrieved from standard input, secret manager, gpg, etc.

No.

#### Checklist

### DIFF
--- a/docs/content/crypt.md
+++ b/docs/content/crypt.md
@@ -478,6 +478,8 @@ Properties:
 
 Password or pass phrase for encryption.
 
+crypt-password or crypt-password-command is required.
+
 **NB** Input to this must be obscured - see [rclone obscure](/commands/rclone_obscure/).
 
 Properties:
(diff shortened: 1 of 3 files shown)


docs/content/crypt.md
albertony, line 1: Your changes in this file is are in an autogenerated section, so you must make sure the help texts in `backend/crypt/crypt.go` contains everything, and you can omit the changes here in `docs/content/crypt.md` from the PR.

hrntknr: Thanks for your review.
I've integrated all the updates exclusively into backend/crypt/crypt.go. The edits in docs/content/crypt.md have been reverted.

albertony:
There is a very similar global flag related to rclone's configuration file:

```
--password-command SpaceSepList       Command for supplying password for encrypted configuration
```

With this implementation:

https://github.com/rclone/rclone/blob/6a7e68aaf2d56df069eb31a46f2f6015a9429b24/fs/config/crypt.go#L173-L205

Did you look at that? Considered re-using the existing function - with some refactoring?

Also, in contrast to that version, your version executes the supplied command in `cmd /c` on Windows and `sh -c` on Unix. I'm not sure that is always the best strategy. E.g. I'm thinking there is a risk of challenges getting the escaping/quoting of arguments correct, or what was your argument for doing that? Wouldn't it be safer and more flexible to not do this, and if required the user can supply a command string which includes the "cmd /c" prefix etc?

hrntknr:
Refactored in [8c88d50](https://github.com/rclone/rclone/pull/8450/commits/8c88d5046ac240b463a0543556358d6eda07e745) to use common code with password-command and removed processes like `sh -c` and `cmd /c`.
Please review again.

hrntknr:
@albertony Please check again.

PR #8455, opened 2025-03-18
box: allow to configure with config file contents

dsander:
Especially when using rclone via rc it is helpful to configure the box backend using the contents of the config file instead of heaving to upload the file to the server that is running rclone.

This allows the Box backend to be configured using the contents of the Box config.json file. It was done in a similar way to how the [Google Cloud Storage](https://github.com/rclone/rclone/blob/44e83d77d754aef3b4408edf62f5943d1c979a8c/backend/googlecloudstorage/googlecloudstorage.go#L106) does it.

https://forum.rclone.org/t/box-support-configuring-with-box-config-file-contents/50309

#### Checklist


docs/content/box.md
```diff
@@ -314,6 +314,19 @@ Properties:
...
+
+Leave blank normally.
+
+Properties:
+
+- Config:      box_config_credentials
+- Env Var:     RCLONE_BOX_BOX_CONFIG_CREDENTIALS
+- Type:        string
+- Required:    false
+
```
albertony, line 329: This is in an autogenerated section, so you can remove this change from your PR.

dsander: Done, sorry I missed the marker in the file above.

albertony: No worries. It happens all the time :) I think we have a linter check for it now, so it would have been "flagged" when running the workflow. I just wanted to give you a chance to fix it first.

backend/box/box.go
```diff
@@ -113,6 +111,11 @@ func init() {
 		}, {
 			Name: "box_config_file",
 			Help: "Box App config.json location\n\nLeave blank normally." + env.ShellExpandHelp,
+		}, {
+			Name:      "box_config_credentials",
```
roucc: This will make the flag box-box-config-credentials, we added the other box ones e.g. box_config_file by mistake and we are trying to avoid that again! Please could you adjust it to just be "config_credentials" thanks! Otherwise I think it is ready to merge.

roucc (changes requested):

backend/box/box.go
dsander: Sure, no problem, done.

dsander:
Awesome, thanks!

PR #8459, opened 2025-03-19
sftp: respect --no-check-dest parameter

sam-mt:

I need to upload files to sftp folder that does not allow list, read, only upload.
SFTP backend attempts to get object stats after upload. This will fail in case when uploading to folder that has only write permissions (no read, list, etc.) with "permission denied" error.
Change makes Update method to return "best guess" stats when --no-check-dest flag is used.

Did not find previous referrences

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -2214,6 +2214,15 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 
 	// Stat the file after the upload to read its stats back if o.fs.opt.SetModTime == false
 	if !o.fs.opt.SetModTime {
+		ci := fs.GetConfig(ctx)
+		if ci.NoCheckDest {
+			fs.Debugf(o, "--no-check-dest is set, so returning best guess")
+			o.modTime = uint32(src.ModTime(ctx).Unix())
+			o.size = src.Size()
+			o.mode = os.FileMode(0666) // regular file
+			return nil
+		}
+
 		err = o.stat(ctx)
 		if err == fs.ErrorObjectNotFound {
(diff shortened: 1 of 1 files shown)


backend/sftp/sftp.go
```diff
@@ -2214,6 +2214,15 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
...
 	if !o.fs.opt.SetModTime {
+		ci := fs.GetConfig(ctx)
+		if ci.NoCheckDest {
+			fs.Debugf(o, "--no-check-dest is set, so returning best guess")
+			o.modTime = uint32(src.ModTime(ctx).Unix())
+			o.size = src.Size()
+			o.mode = os.FileMode(0666) // regular file
+			return nil
+		}
+
```
ncw, line 2225: There is a very similar block of code just below this which GitHub won't let me attach a note to!

This makes me think this would be more general purpose if we could just detect the error that the stat returns in your case and fall into the block below.

This code was put in by 9ee3ad70e9 and there is a nice explanation there.

Maybe we should ignore all errors here and that would be a more general purpose solution?

sam-mt: I tried to avoid calling stat at all, at least performance reasons.
I did tests with different servers and in some cases I got "permission denied" and in some cases "file not found", so I cannot guarantee reliable behavior here.

My main motivation is to skip any extra checks in case when `--no-check-dest` is explicitly set.

to de-duplicate the code I can add `setDefaultMetadata` method to `Object`
@ncw - what do you think?

sam-mt: in addition. handling stat refresh if SetModTime is false could be moved inside `SetModTime` method and on `Update` level - keep only NoCheckDest.

PR #8474, opened 2025-03-25
fuse version query

simonmcnair:
I think it would be nice if the readme clarifies that fuse3 vs fuse is required, or fuse is required and not fuse3.

I can't think why fuse would be chosen over fuse3 ? 

It would be good to explicitly state if fuse3 is not the correct version.

#### Checklist

- [*] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).

### DIFF
--- a/docs/content/docker.md
+++ b/docs/content/docker.md
@@ -45,7 +45,7 @@ on the host.
 The _FUSE_ driver is a prerequisite for rclone mounting and should be
 installed on host:
 ```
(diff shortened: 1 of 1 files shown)


ncw (approved):
This looks great - thanks - rclone is happiest with fuse3 but will work with fuse2 so I'll merge this.

Thank you

PR #8488, opened 2025-04-04
operations: fix call fmt.Errorf with wrong err

alingse:

a simple bug fix, the `err` is nil, should use `er2`

#### Checklist

### DIFF
--- a/fs/operations/multithread.go
+++ b/fs/operations/multithread.go
@@ -314,7 +314,7 @@ func (w *writerAtChunkWriter) WriteChunk(ctx context.Context, chunkNumber int, r
 	case *bufio.Writer:
 		er2 := w.Flush()
 		if er2 != nil {
-			return -1, fmt.Errorf("multi-thread copy: flush failed: %w", err)
+			return -1, fmt.Errorf("multi-thread copy: flush failed: %w", er2)
 		}
 	}
 	return n, nil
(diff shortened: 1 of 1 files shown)


albertony (approved):
Nice catch, thank you.

fs/operations/multithread.go
```diff
@@ -314,7 +314,7 @@ func (w *writerAtChunkWriter) WriteChunk(ctx context.Context, chunkNumber int, r
 	case *bufio.Writer:
 		er2 := w.Flush()
 		if er2 != nil {
-			return -1, fmt.Errorf("multi-thread copy: flush failed: %w", err)
+			return -1, fmt.Errorf("multi-thread copy: flush failed: %w", er2)
```
albertony, line 317: On the other hand; there was no reason for the existing code to create a separate error variable, I think.

So an alternative (better?) fix is this:

```go
err = w.Flush()
if err != nil {
	return -1, fmt.Errorf("multi-thread copy: flush failed: %w", err)
```

What do you think?

alingse: yes, I think should use a easy way. I will update it

fs/operations/multithread.go
```diff
@@ -312,8 +312,8 @@ func (w *writerAtChunkWriter) WriteChunk(ctx context.Context, chunkNumber int, r
 	// if we were buffering, flush to disk
 	switch w := writer.(type) {
 	case *bufio.Writer:
-		er2 := w.Flush()
-		if er2 != nil {
+		err := w.Flush()
```
albertony, line 315:
```suggestion
		err = w.Flush()
```

A small additional adjustment is necessary, or else build with fail with:

> no new variables on left side of :=

PR #8494, opened 2025-04-08
update docs to options/local

Its-Just-Nans:

`options/config` isn't correct

I don't think so, it's more like a typo

#### Checklist

### DIFF
--- a/fs/rc/config.go
+++ b/fs/rc/config.go
@@ -49,7 +49,7 @@ Parameters:
 
 Note that these are the global options which are unaffected by use of
 the _config and _filter parameters. If you wish to read the parameters
-set in _config then use options/config and for _filter use options/filter.
+set in _config then use options/local and for _filter use options/filter.
 
 This shows the internal names of the option within rclone which should
 map to the external options very easily with a few exceptions.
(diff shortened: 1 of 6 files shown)


Its-Just-Nans:
@ncw Any changes for this documentation related MR?

ncw:
This only needs to change ‎fs/rc/config.go all the others are auto generated.

Thanks

Its-Just-Nans:
Fixed

fs/rc/config.go
```diff
@@ -49,7 +49,7 @@ Parameters:
 
 Note that these are the global options which are unaffected by use of
 the _config and _filter parameters. If you wish to read the parameters
-set in _config then use options/config and for _filter use options/filter.
+set in _config then use options/local and for _filter use options/filter.
```
ncw: I think the `options/filter` is wrong too.

`options/local` returns both `config` and `filter` settings.

So I think this should say something like

> If you wish to read the parameters set in `_config` or `_filter` use `options/local`.

What do you think?

ncw (commented):
Thanks for fixing this. Reviewing I see that there is another mistake - do you mind fixing that one too?

fs/rc/config.go
Its-Just-Nans: Okay 
fixed

ncw (approved):
Thank you :-)

PR #8496, opened 2025-04-09
add rc call to unlock config

Its-Just-Nans:

Add a call to unlock the configuration file

https://forum.rclone.org/t/librclone-rpc-encryption-configuration/50761
https://forum.rclone.org/t/askpassword-in-remote-control/50768/2

#### Checklist

### DIFF
--- a/fs/config/rc.go
+++ b/fs/config/rc.go
@@ -11,6 +11,37 @@ import (
 
 func init() {
 	rc.Add(rc.Call{
+		Path:         "config/unlock",
+		Fn:           rcConfigPassword,
+		Title:        "Unlock the config file.",
+		AuthRequired: true,
+		Help: `
+Unlocks the config file if it is locked.
+
+Parameters:
+
+- 'config_password' - password to unlock the config file
+
+A good idea is to disable AskPassword before making this call
+`,
+	})
+}
+
+// Unlock the config file
+// A good idea is to disable AskPassword before making this call
+func rcConfigPassword(ctx context.Context, in rc.Params) (out rc.Params, err error) {
+	configPass, err := in.GetString("config_password")
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


Its-Just-Nans:
note that I think that this call need to set `AskPassword` to `false` but I didn't really did that part

Its-Just-Nans:
Adding a docs page

Is `config/unlock` is good or `config/password` would be better ?

ncw:
> Adding a docs page
> 
> Is `config/unlock` is good or `config/password` would be better ?

I like `config/unlock`. Note that the docs go in the source code- don't edit the the docs file directly as that is auto generated, just put your docs in the `Help` field

Its-Just-Nans:
fixed

Its-Just-Nans:
@ncw Any update on this one ?

fs/config/rc.go
```diff
@@ -9,6 +9,37 @@ import (
...
+// A good idea is to disable AskPassword before making this call
+func rcConfigPassword(ctx context.Context, in rc.Params) (out rc.Params, err error) {
+	configPass, err := in.GetString("config_password")
+	if err != nil {
+		return nil, err
+	}
+	if SetConfigPassword(configPass) != nil {
+		return nil, errors.New("failed to set config password")
+	}
+	return DumpRcBlob(), nil
```
ncw: I think you just need `return nil, nil` here - you don't want to return the config.

Its-Just-Nans: What about returning true
to symbolize a success

ncw: The error code will tell the user whether it is successful or not. The convention we use is either to return `nil` or an empty `rc.Params{}` which could have something else in one day.

Its-Just-Nans: Okay fixed

Its-Just-Nans:
yes fixed

PR #8498, opened 2025-04-10
listremotes should send an empty array

Its-Just-Nans:

The remote call `config/listremotes` sends nil

#### Checklist


albertony (approved):
Thanks. Makes sense to me.
Not sure if breaking anyone's existing workflow is a concern or not, what do you think?

Its-Just-Nans:
> Thanks. Makes sense to me. Not sure if breaking anyone's existing workflow is a concern or not, what do you think?

On my side, I think this change can break existing workflow "only" if the client does direct array access like `list[0]`
But since the endpoint sent `null` before, a direct array access is dangerous in any case and could break with ou without this change.

Its-Just-Nans:
Hi @ncw 

What about this MR now?

Fixing a little ambiguous endpoint by always sending an array (even if empty)

PR #8502, opened 2025-04-10
add jobs stats

Its-Just-Nans:

Add jobs stats in `job/list`

#### Checklist

### DIFF
--- a/fs/rc/jobs/job.go
+++ b/fs/rc/jobs/job.go
@@ -184,6 +184,21 @@ func (jobs *Jobs) IDs() (IDs []int64) {
 	return IDs
 }
 
+func (jobs *Jobs) Stats() (running []int64, finished []int64) {
+	jobs.mu.RLock()
+	defer jobs.mu.RUnlock()
+	running = []int64{}
+	finished = []int64{}
+	for job_id := range jobs.jobs {
+		if jobs.jobs[job_id].Finished {
+			finished = append(finished, job_id)
+		} else {
+			running = append(running, job_id)
+		}
+	}
+	return running, finished
+}
+
 // Get a job with a given ID or nil if it doesn't exist
 func (jobs *Jobs) Get(ID int64) *Job {
 	jobs.mu.RLock()
@@ -412,6 +427,9 @@ Results:
 func rcJobList(ctx context.Context, in rc.Params) (out rc.Params, err error) {
 	out = make(rc.Params)
 	out["jobids"] = running.IDs()
+	current, finished := running.Stats()
+	out["running"] = current
+	out["finished"] = finished
 	out["executeId"] = executeID
 	return out, nil
 }
--- a/fs/rc/jobs/job_test.go
+++ b/fs/rc/jobs/job_test.go
@@ -377,6 +377,8 @@ func TestRcJobList(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, out1)
 	assert.Equal(t, []int64{1}, out1["jobids"], "should have job listed")
+	assert.Equal(t, 1, out1["running"], "should have running job")
+	assert.Equal(t, 0, out1["finished"], "should not have finished job")
 
 	_, _, err = NewJob(ctx, longFn, rc.Params{"_async": true})
 	assert.NoError(t, err)


fs/rc/jobs/job_test.go
```diff
@@ -377,6 +377,8 @@ func TestRcJobList(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, out1)
 	assert.Equal(t, []int64{1}, out1["jobids"], "should have job listed")
+	assert.Equal(t, 1, out1["running"], "should have running job")
+	assert.Equal(t, 0, out1["finished"], "should not have finished job")
```
ncw, line 381: This test seems to think `running` and `finished` are counts, but your `Stats` function returns lists of IDs.

ncw (commented):
This is great - thank you.

There seems to be some confusion (tests vs code) as to whether this returns a number of a list of IDs

Can you put docs for the changes here please

https://github.com/rclone/rclone/blob/e0c99d620340f974255820882ab2802c6694a9ab/fs/rc/jobs/job.go#L403-L407

fs/rc/jobs/job_test.go
Its-Just-Nans: indeed
I fixed it

Its-Just-Nans:
Hi @ncw 

What do you think of this improvement ?

Which able the user to have more informations on the jobs

Its-Just-Nans:
It could be great if this MR makes it into https://github.com/rclone/rclone/milestone/59 :)

Its-Just-Nans:
Hi @ncw 

Any update on this one?

roucc (approved):
Hi, sorry this got left behind, but I will merge it now, thanks!

fs/rc/jobs/job.go
```diff
@@ -412,6 +430,9 @@ Results:
 func rcJobList(ctx context.Context, in rc.Params) (out rc.Params, err error) {
 	out = make(rc.Params)
 	out["jobids"] = running.IDs()
+	runningIDs, finishedIDs := running.Stats()
+	out["running_ids"] = runningIDs
+	out["finished_ids"] = finishedIDs
```
nkiryanov, line 435: Sorry to jump in, but I noticed these keys use `snake_case` instead of `camelCase` like the rest of the project.
Would it make sense to align them before this makes it into a release?

Its-Just-Nans: Hi

Thanks for the review and nice catch

The current function also return "jobids" instead of "jobIds" so in this particular case it seems that there's no real choices.

Also if we agree on casing, we should update #8496

I agree that having a defined case is good
But also think that in json snake case is more common I would say

nkiryanov: >The current function also return "jobids" instead of "jobIds" so in this particular case it seems that there's no real choices.
> Also if we agree on casing, we should update #8496

Ah, you're right - I'm not very familiar with the codebase and missed that `snake_case` is already used in other parts.
My bad, sorry for the noise! :)

Its-Just-Nans: > Ah, you're right - I'm not very familiar with the codebase and missed that `snake_case` is already used in other parts.


Actually #8496 is also a recent MR. So I think we can still change that too

roucc: @ncw What do you think?

ncw: @nkiryanov I think you are right - the snake case sticks out a bit.

We currently use `CamelCase` `lowerCamelCase` and a bit of `smashedtogether` but there are no `snake_case` identifiers in the rc at the moment.

For JSON `lowerCamelCase` is by far the most popular format.

I will update the ones in 8335596 to `lowerCamelCase`.

The only other one is `config_password` in 30d8ab5f2fffa273f98ef05d97c44916bfeb35aa has been in a release now so we can't change that unless we want to do backwards compatibility. I might just do that also.

ncw: I've done those in 4eccc401685927e3f5dd73b8fc654a3c6c20efe5 and 6d4c625bfb87aad942b80e32e98da1ddd31e14d0

Its-Just-Nans: > I've done those in 4eccc401685927e3f5dd73b8fc654a3c6c20efe5 and 6d4c625bfb87aad942b80e32e98da1ddd31e14d0

Hi

What about doing also `jobids` to `jobIds` in
- `job/list`
- `job/status`

(With a background compatibility like config_password ofc)

ncw: I don't really want to duplicate the output. Those ones have been around for a long time too so I think we'll just have to live with them :grimacing:

Its-Just-Nans: I agree for `job/list`

But I think that for `job/status`, we need the same as https://github.com/rclone/rclone/commit/6d4c625bfb87aad942b80e32e98da1ddd31e14d0


I made https://github.com/rclone/rclone/pull/8995

PR #8503, opened 2025-04-12
googlephotos: take into account duplicated ids as well, reduce iterations

ferferga:

I'm mainly using rclone with the `media/all` folder to `sync` my photos and every time I saw a lot of `Duplicate object whatever {AAA...}.jpg found in source - ignoring`. When listing photos, we took into account the fact that duplicated filenames could exist in an account, but not the fact that the API might return duplicated ids (I guess that, when lisiting all photos, if a photo exists in multiple albums, is in the archive, etc), but those IDs are always unique! It's wrong to have duplicated ids!

With this change, we skip duplicated ids, leaving the old algorithm untouched (duplicated filenames are still prefixed with the ID). I also found it was possible to reduce the iterations we did in the code, improving performance.

I performed a sync with a version built with these changes and no files were deleted from the destination — proving they were completely skipped before and no changes to the previous algorithm happened. If I do `reclone lsf GPhotos:media/all`, photos with duplicated filenames but different IDs are still being prefixed correctly.

For reference, this was my GPhotos remote config, in case you're not able to reproduce the issue — I think the main culprit is `include_archived = true`:

```
[GPhotos]
type = google photos
read_only = true
read_size = false
start_year = 1800
include_archived = true
...token, client and secrets excluded
```

No, I was seeing this issue myself when syncing and went right ahead on fixing it.

#### Checklist


backend/googlephotos/googlephotos.go
```diff
@@ -654,24 +654,30 @@ func (f *Fs) listDir(ctx context.Context, prefix string, filter api.SearchFilter
...
+	// Handle duplicate items
+	filenameCount := map[string]int{}
+	seenIDs := map[string]bool{}
+	newEntries := make([]fs.DirEntry, 0, len(entries))
 	for _, entry := range entries {
 		o, ok := entry.(*Object)
-		if ok {
-			dupes[o.remote]++
+		if !ok {
+			newEntries = append(newEntries, entry)
```
ferferga, line 665: Not sure if necessary to append in case ok is false or just continue is enough, but left it here to better match the original behaviour. Please let me know if this should be safe to remove.

backend/googlephotos/googlephotos.go
```diff
@@ -654,24 +654,35 @@ func (f *Fs) listDir(ctx context.Context, prefix string, filter api.SearchFilter
...
+	newEntries := make([]fs.DirEntry, 0, len(entries))
 	for _, entry := range entries {
 		o, ok := entry.(*Object)
-		if ok {
-			duplicated := dupes[o.remote] > 1
-			if duplicated || o.remote == "" {
-				o.remote = addFileID(o.remote, o.id)
-			}
+		if !ok {
+			newEntries = append(newEntries, entry)
```
ferferga, line 673: Not sure if necessary to append in case ok is false or just continue is enough, but left it here to better match the original entries' contents. Please let me know if this should be safe to remove.

backend/googlephotos/googlephotos.go
```diff
@@ -654,24 +654,38 @@ func (f *Fs) listDir(ctx context.Context, prefix string, filter api.SearchFilter
...
```
ferferga, line 675: Not sure if necessary to append in case ok is false or just continue is enough, but left it here to better match the original entries' contents. Please let me know if this should be safe to remove and we can simply `continue`.

backend/googlephotos/googlephotos.go
```diff
@@ -654,24 +654,38 @@ func (f *Fs) listDir(ctx context.Context, prefix string, filter api.SearchFilter
...
```
ferferga, line 675: Not sure if necessary to append in case ok is false or just continue is enough, but left it here to better match the original entries' contents. Please let me know if this should be safe to remove and we can simply `continue`.

ferferga:
Hey @ncw, sorry for pinging, but I've seen that all old PRs have at least a comment from you or some other maintainer, and this one has been sitting idle for quite some time, probably forgotten, hence the ping :).

May I have a review here please?

PR #8529, opened 2025-04-30
Zus backend

Hitenjain14:

Add zus storage as a backend

#### Checklist


Hitenjain14:
Thanks for taking a look! I’ve run the tests and they’re all passing. Regarding the binary size, I’ll check to see if there’s room for reduction.
As for the cryptographic signatures, Zus uses the BLS scheme, and currently there’s no pure Go implementation available for it.

ncw:
> Thanks for taking a look! I’ve run the tests and they’re all passing.

Great.

> Regarding the binary size, I’ll check to see if there’s room for reduction.

:+1: 

> As for the cryptographic signatures, Zus uses the BLS scheme, and currently there’s no pure Go implementation available for it.

We don't allow CGO in the rclone build as it breaks the one binary, deploy everywhere users enjoy, so we are going to have to think of a way around that.

Judging from the errors https://github.com/herumi/bls-go-binary/ is the library.

We were trying to build this on our Windows test machine but we couldn't get it to build there - can you get this to build on Windows?

ncw:
I tried to make this work following the instructions in the zus.md.

I got as far as here

```
$ rclone lsf TestZus: -vv
2025/05/02 12:04:11 CRITICAL: Failed to create file system for "TestZus:": [conf]missing config file: /home/ncw/.zcn/config.yaml
```

Where does this extra config file come from? Rclone backends need to stand alone and not need extra tools to configure them. All the config should live in the `.rclone.conf` file if at all possible so you can copy it from one machine to another and have a complete working setup.

ncw:
> > Where does this extra config file come from? Rclone backends need to stand alone and not need extra tools to configure them. All the config should live in the `.rclone.conf` file if at all possible so you can copy it from one machine to another and have a complete working setup.
> 
> As the backend requires users to own an allocation, they would have setup the wallet and related config required to initialise the sdk, do you suggest we ask the user to enter their mnemonics and config data rather than reading from default folder?

I think the docs need some steps that someone who knows nothing about zus could use to make rclone work. I have no idea how to make the missing config file!

In an ideal world, yes, you would enter your config info directly into rclone. Maybe that is too hard, I don't know.

Some backends have a mode that you can use a config file **or** existing auth (eg env_auth in the S3 backend).

Take a look at the Storj docs (which is a very similar system) to see what I think your docs should look like if you want to make zus accessible to more people: https://rclone.org/storj/ Note how there is a clear separation of the S3 access and the direct access and pros and cons for each.

PR #8533, opened 2025-05-02
googlephotos: update readOnly/Write scopes - fixes #8434

### LINKED ISSUE #8434: googlephotos: plan to remove
Google is removing the scopes that rclone uses to backup photos. So, we should just remove the code to avoid annoyance and user reports.

https://developers.google.com/photos/support/updates

As part of our changes to the Google Photos APIs, we have made the following updates.

As shown on the [updated Authorization page](https://developers.google.com/photos/overview/authorization#library-api-scopes), the following scopes will be removed from the Library API after March 31, 2025:

photoslibrary.readonly

germancasares:

Google Photos has updated their allowed scopes and now the only capability that RClone can offer is to upload new photos. It will also be able to download photos, but only those that Rclone uploaded.

https://github.com/rclone/rclone/issues/8434

#### Checklist

### DIFF
--- a/backend/googlephotos/googlephotos.go
+++ b/backend/googlephotos/googlephotos.go
@@ -43,6 +43,7 @@ var (
 	errAlbumDelete = errors.New("google photos API does not implement deleting albums")
 	errRemove      = errors.New("google photos API only implements removing files from albums")
 	errOwnAlbums   = errors.New("google photos API only allows uploading to albums rclone created")
+	errReadOnly    = errors.New("can't upload files in read only mode")
 )
 
 const (
@@ -52,19 +53,31 @@ const (
 	listChunks                  = 100 // chunk size to read directory listings
 	albumChunks                 = 50  // chunk size to read album listings
 	minSleep                    = 10 * time.Millisecond
-	scopeReadOnly               = "https://www.googleapis.com/auth/photoslibrary.readonly"
-	scopeReadWrite              = "https://www.googleapis.com/auth/photoslibrary"
-	scopeAccess                 = 2 // position of access scope in list
+	scopeAppendOnly             = "https://www.googleapis.com/auth/photoslibrary.appendonly"
+	scopeReadOnly               = "https://www.googleapis.com/auth/photoslibrary.readonly.appcreateddata"
+	scopeReadWrite              = "https://www.googleapis.com/auth/photoslibrary.edit.appcreateddata"
 )
 
 var (
+	// scopes needed for read write access
+	scopesReadWrite = []string{
+		"openid",
+		"profile",
+		scopeAppendOnly,
+		scopeReadOnly,
+		scopeReadWrite,
+	}
+
+	// scopes needed for read only access
+	scopesReadOnly = []string{
+		"openid",
+		"profile",
+		scopeReadOnly,
+	}
+
 	// Description of how to auth for this app
 	oauthConfig = &oauthutil.Config{
-		Scopes: []string{
-			"openid",
-			"profile",
-			scopeReadWrite, // this must be at position scopeAccess
-		},
+		Scopes:       scopesReadWrite,
 		AuthURL:      google.Endpoint.AuthURL,
 		TokenURL:     google.Endpoint.TokenURL,
 		ClientID:     rcloneClientID,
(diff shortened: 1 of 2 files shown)


ncw:
Thanks for doing this @germancasares 

I'm having to redo the approval for the current rclone client access and I've requested the `photoslibrary.appendonly` scope as part of that. They have made me drop the old scopes too. However these have stopped working now anyway.

So I can't merge this until the approval comes though. Or I suppose I could since google photos is broken anyway...

Reading [the docs](https://developers.google.com/photos/overview/authorization#library-api-scopes)  they make me think  we need `photoslibrary.readonly.appcreateddata` too to be able to read stuff we've uploaded. Is that the case? Can you check?

I don't think we'll be able to delete photos either after this change even if we uploaded them which makes the syncing not work very well.

This also needs a rework of the rclone docs https://github.com/rclone/rclone/blob/master/docs/content/googlephotos.md

germancasares:
> Thanks for doing this @germancasares
> 
> I'm having to redo the approval for the current rclone client access and I've requested the `photoslibrary.appendonly` scope as part of that. They have made me drop the old scopes too. However these have stopped working now anyway.
> 
> So I can't merge this until the approval comes though. Or I suppose I could since google photos is broken anyway...
> 
> Reading [the docs](https://developers.google.com/photos/overview/authorization#library-api-scopes) they make me think we need `photoslibrary.readonly.appcreateddata` too to be able to read stuff we've uploaded. Is that the case? Can you check?
> 
> I don't think we'll be able to delete photos either after this change even if we uploaded them which makes the syncing not work very well.
> 
> This also needs a rework of the rclone docs https://github.com/rclone/rclone/blob/master/docs/content/googlephotos.md

I do use rclone but with my own clientID, so in theory I could use it without approval. I understand that the GPhotos backend needs an overhaul now, so unfortunate that Google is closing down the capabilities of the APIs. I didn't even consider the role to remove images uploaded by Rclone. I couldn't find anything related to that on their blog post: https://developers.google.com/photos/support/updates.

Syncing was my only use case for the Google Photos backend, oh well I guess this will speed up my transition to Immich. Thank you so much for maintaining Rclone and I'll keep checking in to see if there is any other way to use Google Photos with Rclone.

backend/googlephotos/googlephotos.go
```diff
@@ -52,8 +52,8 @@ const (
 	listChunks                  = 100 // chunk size to read directory listings
 	albumChunks                 = 50  // chunk size to read album listings
 	minSleep                    = 10 * time.Millisecond
-	scopeReadOnly               = "https://www.googleapis.com/auth/photoslibrary.readonly"
-	scopeReadWrite              = "https://www.googleapis.com/auth/photoslibrary"
+	scopeReadOnly               = "https://www.googleapis.com/auth/photoslibrary.appendonly"
+	scopeReadWrite              = "https://www.googleapis.com/auth/photoslibrary.appendonly"
```
sergeivolodin, line 56: Shouldn't it be
```go
scopeReadOnly               = "https://www.googleapis.com/auth/photoslibrary.readonly.appcreateddata"
scopeReadWrite              = "https://www.googleapis.com/auth/photoslibrary.readonly.appcreateddata https://www.googleapis.com/auth/photoslibrary.appendonly"
```
?

They separated the permissions so that read and write are separate permissions. There's no "ReadWrite" permission anymore. To sync photos and videos, both are though. Also, `scopeReadOnly` should definitely point to `readonly.appcreateddata` (as `appendonly` isn't a read permission).

See https://developers.google.com/photos/overview/authorization#library-api-scopes

Your fix didn't work for me, but mine did.

Did sync work for you with your fix?

sergeivolodin (changes requested):
Suggested a change better reflecting the new permissions

ncw:
@germancasares @sergeivolodin I've updated the PR with the correct scopes for read/write access and read only access and tested that they work.

I've also updated the docs just to warn people that rclone can only download files it uploaded.

I'll merge this when the CI completes.

Thank you both for your help.

backend/googlephotos/googlephotos.go
ncw: I've incorporated that into a new version and I've tested read only and read write and all seems to be working now.

stantyan:
@ncw Hi, is it possible to somehow start using this fixed version now or we need to wait for the next release? I am using TrueNAS Scale that uses rclone for various sync tasks, so does it mean in my case I need to wait until TrueNAS will update to the fixed Google Photos rclone version too?

ncw:
@stantyan If you can use the rclone beta on your TrueNAS then you'll have the new features. I'm afraid I know nothing about the NAS though so you'll have to ask elsewhere for how to do that.

mrtolkien:
I still get the same issue on 1.70 that I had on 1.69:
image

The update did make it so I was able to create a read-only remote, but I still can't seem to create a R/W remote anymore even after updating `rclone`

Issue #290, opened 2016-01-10
"Failed to create file system for "remote:": Didn't find section in config file"


nfshr:
rclone was set up as advised in the docs via auto-config for google-drive (choice 9, "drive"). Set up went ok and token was received. Config was left at default when possible. Now i get the following err when trying to connect:

$ rclone ls remote:
2016/01/10 19:16:52 Failed to create file system for "remote:": Didn't find section in config file

Any Ideas why this happens? Couldn't find anything in the docs/Google.

OS: Arch; rclone: community/rclone 1.26-1

klauspost:
Replace `remote` with the name you gave your drive.

Use `rclone config` to see the name of your remote.

nfshr:
Oh, ok, thx! :)

venkatnaidu:
rclone installed ubuntu . when i run rclone lsd remote:     
giving this error 
Failed to create file system for "remote:": Didn't find section in config file
Current remotes:

Name                 Type
====                 ====
test-wed-bulk        s3
wed-bulk-upload      s3

venkatnaidu@ecadmin-Vostro-460:~$ rclone ls test-wed-bulk:
2016/03/30 16:20:26 S3 bucket : Can't list objects at root - choose a bucket using lsd

Transferred:            0 Bytes (   0.00 kByte/s)
Errors:                 1
Checks:                 0
Transferred:            0
Elapsed time:           0

xinmeng1:
thx

Aireezz:
I'm having this same issue yet I have specified it with the name I gave the drive.

Drive name = encrypted
Files located under

Main folder -Media
 sub folders -
                    Tv
                    Movies
                    Music

i run this command            rclone mount encrytped:media /mnt/googledrive &


I have also tried specifying a sub folder like this.   rclone mount encrytped:media/tv /mnt/googledrive &

No luck, any thoughts?

Flash619:
I have the exact same issue when using a crypt remote type. If I try secret:myfolder/ It get "didn't find section in config file".

razzmatazzrr:
What is this part in the rclone config about?

service_account_file> /home/foo/myJSONfile.json # This is where the JSON file goes!

Can I leave this blank? I've set a root_folder_id. All I want to do is copy and delete jpgs from a folder on my raspberry pi, however when I try run this...

rclone -v move /home/pi/timelapse remote:tent --delete-after --include *.jpg --include *.log

[...]

PR #8536, opened 2025-05-03
docs: Clarify remote naming convention

JellyJoe198:

If this was here I would not have gotten "Failed to create file system for "remote:": Didn't find section in config file" from using the wrong name for my remote, as I didn't know that had to be the name you chose.

I replaced  
"Once configured you can then use `rclone` like this,"  
with  
"Once configured you can then use `rclone` like this, (replace `remote` with the name you gave your remote)"  

on all docs except `union.md` and `koofr.md` as they have a different format

#290

#### Checklist

### DIFF
--- a/docs/content/alias.md
+++ b/docs/content/alias.md
@@ -81,7 +81,7 @@ q) Quit config
 e/n/d/r/c/s/q> q
 ```
 
-Once configured you can then use `rclone` like this,
+Once configured you can then use `rclone` like this, (replace `remote` with the name you gave your remote)
 
 List directories in top level in `/mnt/storage/backup`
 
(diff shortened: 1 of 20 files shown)


docs/content/alias.md
```diff
@@ -81,7 +81,7 @@ q) Quit config
 e/n/d/r/c/s/q> q
 ```
 
-Once configured you can then use `rclone` like this,
+Once configured you can then use `rclone` like this, (replace `remote` with the name you gave your remote)
```
albertony, line 84: Good idea, thank you!

I'm thinking... I find the trailing comma of the original text a bit strange, and with your added parenthesis this is perhaps even more true. Not a big deal, of course, but what about something like the following instead?

```suggestion
Once configured you can then use `rclone` like this (replace `remote` with the name you gave your remote):
```

JellyJoe198: Yeah that sounds better, I'll do that. Also I've only used rsync on a OneDrive remote so far but I think this should work for all of these files. It might be even more clear to say "drive" for OneDrive, google drive, etc. and different terminology for others.

PR #8542, opened 2025-05-07
Add FAQ for getting rid of the 'Config file not found' message

geerlingguy:

To make it more obvious to users who don't need to configure remotes that they can create an empty `rclone.conf` file to get rid of the annoying notices after every command like:

```
2025/05/06 17:42:15 NOTICE: Config file "/Users/jgeerling/.config/rclone/rclone.conf" not found - using defaults
```

Yes, see https://forum.rclone.org/t/notice-about-missing-rclone-conf-is-annoying/51116

#### Checklist

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -22,6 +22,20 @@ See the [remote setup docs](/remote_setup/) for more info.
 
 This has now been documented in its own [remote setup page](/remote_setup/).
 
+### How can I get rid of the "Config file not found" notice?
+
+If you see a notice like 'NOTICE: Config file "rclone.conf" not found', this
+means you have not configured any remotes.
+
+If you need to configure a remote, see the [config help docs](/docs/#configure).
+
+If you are using rclone entirely with [on the fly remotes](/docs/#backend-path-to-dir),
+you can create an empty config file to get rid of this notice, for example:
+
+```
+rclone config touch
+```
+
 ### Can rclone sync directly from drive to s3 ###
 
 Rclone can sync between two remote cloud storage systems just fine.


geerlingguy:
Fixed a couple typos / specified the example command is for Linux.

docs/content/faq.md
```diff
@@ -22,6 +22,20 @@ See the [remote setup docs](/remote_setup/) for more info.
...
+
+If you need to create a remote, see the [remote setup page](/remote_setup/).
+
+If you are using rclone without remotes, you can create an empty config file
+to get rid of this notice, for example on Linux:
+
+```
+touch ~/.config/rclone/rclone.conf
+```
+
```
ncw, line 38: You can also use `--config ""` which saves making a file.

docs/content/faq.md
```diff
@@ -22,6 +22,20 @@ See the [remote setup docs](/remote_setup/) for more info.
 
 This has now been documented in its own [remote setup page](/remote_setup/).
 
+### How can I get rid of the "Config file not found" notice?
+
+If you see a notice like 'NOTICE: Config file "rclone.conf" not found', this
+means you have not configured any remotes.
+
+If you need to create a remote, see the [remote setup page](/remote_setup/).
```
ncw, line 30: I think this would be a better link https://rclone.org/docs/#configure

(Without the `https://rclone.org` prefix for the docs, but you've figured that out already :-)

docs/content/faq.md
```diff
@@ -22,6 +22,20 @@ See the [remote setup docs](/remote_setup/) for more info.
...
+
+If you are using rclone without remotes, you can create an empty config file
```
ncw, line 32: Strictly speaking remotes configured on the fly on the command line are still remotes.

So perhaps something like this might be better:

If you are using rclone entirely with [on the fly remotes](https://rclone.org/docs/#backend-path-to-dir), you can create an empty config file

ncw (commented):
Thank you for doing this.

I put some points for discussion inline.

docs/content/faq.md
albertony: I agree. And if you want the conf file on disk, then [rclone config touch](https://rclone.org/commands/rclone_config_touch/) is a simpler and cross-platform command where you don't have to figure out what the path is.

geerlingguy: Thanks; I've updated the suggested command to this instead. Seems like the best option.

docs/content/faq.md
geerlingguy: Fixed this as well. Thanks!

Issue #8183, opened 2024-11-15
Support for Hetzner Cloud S3


maxpain:

ncw:
Hetzner  s3 works fine with s3.

It would be nice to have a provider to simplify config. If someone wants to add one - follow the guide here: https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#adding-a-new-s3-provider

spiffytech:
I've submitted PR #8551 to add an S3 provider for Hetzner object storage.

tysoncung:
I'm interested in working on this issue. Could you provide more context about what you're looking for? Any additional details about requirements or constraints would be helpful.

ncw:
This has been merged now

PR #8551, opened 2025-05-12
#8183: Add S3 provider for Hetzner object storage

spiffytech:

This PR closes issue #8183: adding an S3 provider for Hetzner Object Storage.

#8183 

### Hetzner test performance

Hetzner only offers European regions for their object storage.

When I ran the integration tests from my home in North Carolina, or my VPS in Virginia, they blew past the 10-minute test timeout (especially the chunk tests).

When I used a Hetzner VPS in Finland, they ran well under the limit.

Yay, trans-Atlantic latency!

### Hetzner `useAlreadyExists` behavior

The existing integration tests appear to assume that if a bucket is being asynchronously deleted, and you request creating a new bucket with the same name, the provider honor the bucket create command (cancelling the deletion?).

Hetzner doesn't do this. You'll get a create error (which the S3 `fs` code swallows), then the bucket still gets deleted, and then all the tests error out because they try to use a bucket that doesn't exist.

I considered waiting until the API reported the bucket name is available for reuse, but we can't count on every provider doing that swiftly. Hetzner, at least, is slow to report that the name is free. If we were recreating the bucket into a different region, AWS [is known to](https://serverfault.com/questions/554374/how-long-before-an-s3-bucket-can-be-created-with-same-name-after-deletion) take over an hour.

Instead, I altered the tests to use a new bucket name each time they request a bucket. With this change, the tests pass under Hetzner and AWS. I don't know enough about what's being tested to know if this breaks any assumptions.

**Please let me know if this isn't the right way to handle this!**

I saw some tests used `f.MkDir()` and other use `operations.MkDir`. I'm not sure why they're different, so I left the latter alone.

rclone-s3 (2)

rclone-hetzner

#### Checklist


ncw:
> The existing integration tests appear to assume that if a bucket is being asynchronously deleted, and you request creating a new bucket with the same name, the provider honor the bucket create command (cancelling the deletion?).

I'm not particularly happy about altering the integration tests just for one provider.

The azureblob backend has a similar problem too and what it does is check the create error (hopefully it is specific enough) and sleep and retry. Could we use that technique here too?

spiffytech:
Very understandable.

Note that this PR creates a new S3 provider, not a separate backend like azureblob. The proposal to retry would require changing the behavior for all S3 providers, or special-casing Hetzner Object Storage. Or maybe a new quirk? Are any of these choices acceptable, or do we need to look at a different approach?

Hetzner returns `BucketAlreadyExists` for buckets pending deletion. `s3.go` `makeBucket` explicitly swallows that error if quirk `useAlreadyExists` is `false` (which it is for Hetzner). Is it appropriate for other S3 providers to stop swallowing the error?

```go
// Current code
case "BucketAlreadyExists", "BucketNameUnavailable":
    if f.opt.UseAlreadyExists.Value {
        // We can trust BucketAlreadyExists to mean not owned by us, so make it non retriable
        err = fserrors.NoRetryError(err)
    } else {
        // We can't trust BucketAlreadyExists to mean not owned by us, so ignore it
        err = nil
    }
}
```
I did some quick tests to validate the proposal. azureblob triggers `Pacer` retries using `f.shouldRetry()`. For Hetzner, this doesn't work without further changes: `Pacer` retries only a few times, and pretty quickly. It hits the retry limit and errors out before the `CreateBucket` call succeeds.

I tried hacking a `for` loop into `s3.go` `makeBucket` with a `time.Sleep()`, just to see if the idea works at all. The good news is all tests pass except one.

Test `FsListRootedSubdir` fails because it relies on the current error-swallowing behavior. The test calls `cache.Get` ➜ `GetFn` against a bucket that already exists and is _not_ pending deletion. So the loop just spins forever waiting for `CreateBucket` to stop returning `BucketAlreadyExists`. My read on the code is `GetFn` is expected to be idempotent, and AFAICT Hetzner cannot handle this scenario idempotently.

I don't know how we'd make `cache.Get`'s idempotent behavior support Hetzner Object Storage. The error for a permanent bucket is the same as for one that'll disappear any second. And `cache.Get` can't just ask "does this bucket exist" because we don't know if it's pending deletion.

**Next steps**
So, I need input on where to go from here:
1. How should `s3.go` know to apply Hetzner-specific behavior?
    - A new quirk seems the least intrusive. Is this an appropriate use of quirks?
2. Presuming Pacer is the right layer to apply the retries, what's the idiomatic way to make pacer more lenient for this narrow situation?
3. How should we preserve `cache.Get`'s idempotence with Hetzner? Can it even be done?

n.b. (1) and (2) sound more invasive than changing the integration tests, but (3) seems like a genuine problem we have to solve regardless.

spiffytech:
I've pushed an update backing out the changes to the integration tests, and adding a retry for `CreateBucket`. This is WIP and isn't ready to merge, as noted above.

ncw:
> * How should `s3.go` know to apply Hetzner-specific behavior?
>   
>   * A new quirk seems the least intrusive. Is this an appropriate use of quirks?

Absolutely. That is what quirks are for. Make sure you name it for what it does, not fixHetzner.

> * How can we preserve `CreateBucket` idempotence with Hetzner?

Good question!

We could just decide that one test failure is ok. Does it work if you just run that test with the -run flag? If so the integration tester would run tes tests, get one failure, and retry that test.

> * Can we have `makeBucket` use more retries than S3 `Pacer`'s default?

The AWS SDK reties 10 times for most things anyway which is why we only use 2 by default.

You could just have a bit of manual logic, or make a new pacer, whichever you fancy. I think if the tests pass today it probably isn't worth worrying about too much though.

spiffytech:
> Does it work if you just run that test with the -run flag?

Nope: because an `Fs` is initialized at the top of the test suite, running just `FsListRootedSubdir` still fails to create the bucket.

> I think if the tests pass today it probably isn't worth worrying about too much though.

Sounds good. YAGNI!

---

I've pushed an update that adds the quirk. Given we're willing to accept one test failure, I think this is ready for review!

jumbi77:
Can i may ask the status of this? Maybe any hints from @ncw what should be done/fixed to get this integrated?

Much thanks to all participants

jumbi77:
Just FYI linking #8343 as outdated duplicate ?

Neustradamus:
@spiffytech: Any progress on your PR?

spiffytech:
This PR code-complete and ready for review.

roucc:
@spiffytech I have rebased your pull request as it had many conflicts. I also removed the bucket checking mechanism as it is unlikely users will actually cause this problem, it is only seen in the tests. I ran the tests on a bucket and they all passed.

```sh
go test -v -remote hetzners3:rclone-testing -short
```

I'll merge this when the tests pass. Thanks for doing this, sorry it has taken us such a long time to merge it.

spiffytech:
Great, thanks for pushing it over the finish line!

PR #8554, opened 2025-05-13
feat: support symlink detection when reading directories

huangnauh:
#### Checklist

### DIFF
--- a/cmd/mount/dir.go
+++ b/cmd/mount/dir.go
@@ -130,6 +130,12 @@ func (d *Dir) ReadDirAll(ctx context.Context) (dirents []fuse.Dirent, err error)
 		if node.IsDir() {
 			dirent.Type = fuse.DT_Dir
 		}
+		switch node := node.(type) {
+		case *vfs.File:
+			if node.IsSymlink() {
+				dirent.Type = fuse.DT_Link
+			}
+		}
 		dirents = append(dirents, dirent)
 	}
 	itemsRead = len(dirents)


Neustradamus:
@ncw: Have you seen this PR?

ncw:
@huangnauh @Neustradamus sorry missed this one.

Can either of you explain what it fixes please? So when using `rclone mount ` with `XXX` flags, I do `XXX` and I see `YYY` but I was expecting `ZZZ`? Thanks

darthShadow:
I think when the changes were made for adding symlink support for the various mount commands, we missed setting proper flags in readdir results, specifically only for the mount command. It was added for cmount & mount2, but got missed in mount.

mount2 implementation: https://github.com/rclone/rclone/blob/master/cmd/mount2/node.go#L257
cmount implementation: https://github.com/rclone/rclone/blob/master/cmd/cmount/fs.go#L146

mount symlink changes: https://github.com/rclone/rclone/commit/db1ed6969318a3112118f552611b24a9236b7be6

It missed the update for readdir results too.

ncw:
@darthShadow I think the fix looks correct.

I'd just like to know what it breaks because in testing `rclone mount` with symlinks works fine.

trapexit:
The problem is likely the difference between d_type being set in readdir and stat. At the very least d_type should be set to DT_UNKNOWN in Linux but d_type is not supported on all platforms and where it is software needs to handle DT_UNKNOWN. IE... they shouldn't assume readdir gives you the file type. Should be picked up from stat.

ncw:
Thanks for your input @trapexit 

I will merge this as it is correct. Just nosy as to what it broke!

PR #8571, opened 2025-05-26
fs: fix goroutine leak and improve stats accounting process

### LINKED ISSUE #8570: Goroutine leak in `accounting.(*StatsInfo).averageLoop`
https://forum.rclone.org/t/goroutine-leak-in-accounting-statsinfo-averageloop/51321

Monitoring my S3 rclone mount showed that goroutines count is infinitely increasing.

This can be seen by plotting the go_goroutines metric, here, roughly 24 hours after starting the mount:
Image

Profiling using `go tool pprof -dot http://localhost:5574/debug/pprof/goroutine` gave me this diagram:

Image

From the diagram, we can clearly see that the `averageLoop` the one goroutine kept from behind deleted.
When running the tool multiple times, the number of currently running goroutines is increased.

Reading [the code](https://github.com/rclone/rclone/blob/15510c66d401e6737a2b15a0dba988e04ffb4ddd/fs/accounting/stats.go#L331), I'm suspecting that the `a.stop` channel is sometimes not closed. This would explain why the goroutines aren't being freed.

I'm currently building it myself to debug the issue so that's the latest `master` commit:

```
rclone v1.70.0-beta.8745.15510c66d
- os/version: alpine 3.21.3 (64 bit)
- os/kernel: 5.15.0-140-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.24.3
- go/linking: static
- go/tags: none
```

Amazon S3 Object Storage

```bash
rclone mount config-name:/my-bucket /data \
  --config=/config/rclone.conf \
  --uid=1000 \
  --gid=1000 \
  --cache-dir=/cache \
  --use-mmap \
  --allow-other \
  --umask=002 \
  --rc \
  --rc-no-auth \
  --rc-addr=:5574 \
  --rc-enable-metrics \
  --vfs-cache-mode writes \
  --allow-non-empty
```

#### How to use GitHub

quantumsheep:
Fixes #8570

The issue https://github.com/rclone/rclone/issues/8570 describe a goroutine leak of `accounting.(*StatsInfo).averageLoop`.

These changes does 3 things:
1. Stop automatically starting the average loop when calling `NewStats`.
   * This was an issue in `accounting.(*statsGroups).sum` where it called `NewStats` without stopping the average loop. Stopping the loop in this function could be an alternative fix to the issue.
2. Completly stop and restart the loop instead of pausing it. This simplifies a lot the code.
3. Use a `context.Context` instead of a `chan bool` to ask the loop to stop. This is also for simplification.

Proof that it works well (you can compare to [the issue](https://github.com/rclone/rclone/issues/8570)'s image):

Image

I posted on the forum but currently, no discussion was made.

https://forum.rclone.org/t/goroutine-leak-in-accounting-statsinfo-averageloop/51321

#### Checklist

### DIFF
--- a/fs/accounting/stats.go
+++ b/fs/accounting/stats.go
@@ -68,6 +68,7 @@ type StatsInfo struct {
 
 type averageValues struct {
 	mu      sync.Mutex
+	period  float64
 	lpBytes int64
 	lpTime  time.Time
 	speed   float64
@@ -328,8 +329,6 @@ func (s *StatsInfo) calculateTransferStats() (ts transferStats) {
 }
 
 func (s *StatsInfo) averageLoop(ctx context.Context) {
-	var period float64
-
 	ticker := time.NewTicker(averagePeriodLength)
 	defer ticker.Stop()
 
@@ -347,11 +346,11 @@ func (s *StatsInfo) averageLoop(ctx context.Context) {
 				avg = float64(a.lpBytes) / elapsed
 			}
 
-			if period < averagePeriod {
-				period++
+			if a.period < averagePeriod {
+				a.period++
 			}
 
-			a.speed = (avg + a.speed*(period-1)) / period
+			a.speed = (avg + a.speed*(a.period-1)) / a.period
 			a.lpBytes = 0
 			a.lpTime = now
 


ncw:
I think this problem was introduced in 46b498b86a4ba690707a93369ede32e11ae039c3 by @darthShadow 

>     stats: fix the speed not getting updated after a pause in the processing
> 
>     This shifts the behavior of the average loop to be a persistent loop
>     that gets resumed/paused when transfers & checks are started/completed.
> 
>     Previously, the averageLoop was stopped on completion of
>     transfers & checks but failed to start again due to the protection of
>     the sync.Once

@quantumsheep wrote

> These changes does 3 things:
> 
> 1. Stop automatically starting the average loop when calling `NewStats`.
>    
>    * This was an issue in `accounting.(*statsGroups).sum` where it called `NewStats` without stopping the average loop. Stopping the loop in this function could be an alternative fix to the issue.

Do you think that is the cause of the go routine leak?


> 2. Completly stop and restart the loop instead of pausing it. This simplifies a lot the code.

I'm concerned stopping and starting the loop will cause the stats to be different. Looking at the code the only difference would be that we would soft start the moving average again.

> 3. Use a `context.Context` instead of a `chan bool` to ask the loop to stop. This is also for simplification.

That looks nice.

> Proof that it works well (you can compare to [the issue](https://github.com/rclone/rclone/issues/8570)'s image):

@darthShadow - any thoughts on this?

PR #8573, opened 2025-05-27
drive: fixed the issue where Google Docs file name was truncated, causing panic or unparseable characters

jaycee88:
When downloading Google documents, the remote does not contain the suffix, and the extension is '.docx'. The code here will cause panic or the file name will be truncated (if the file name is Chinese or Japanese, there will be unparseable characters)

9f7df015cbd14d1f07ae0f6b60ea02bf
86a48ca000ff88e2d611e171d2f1efe5

#### Checklist

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -1698,7 +1698,7 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
 		return nil, err
 	}
 
-	remote = remote[:len(remote)-len(extension)]
+	remote = strings.TrimSuffix(remote, extension)
 	obj, err := f.newObjectWithExportInfo(ctx, remote, info, extension, exportName, exportMimeType, isDocument)
 	switch {
 	case err != nil:


backend/drive/drive.go
```diff
@@ -1495,7 +1495,7 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
 		return nil, err
 	}
 
-	remote = remote[:len(remote)-len(extension)]
+	remote = strings.TrimSuffix(remote, extension)
```
ncw, line 1498: I think this must be a bug in `f.getRemoteInfoWithExport(ctx, remote)`  - why is it returning extension if remote doesn't have extension?

ncw (commented):
I'm not 100% sure this is the correct fix. Your code looks good but I think `remote` should always have `extension` - see inline.

How do you replicate this? Can you give me step by step instructions please :-)

Thank you.

jaycee88:
> I'm not 100% sure this is the correct fix. Your code looks good but I think `remote` should always have `extension` - see inline.
> 
> How do you replicate this? Can you give me step by step instructions please :-)
> 
> Thank you.

Sorry, it turns out the issue was caused by our modification to the list behavior. In our version, the listed Google Docs use the original file name without appending the .docx extension. However, in the native rclone implementation, Google Docs are listed with the .docx extension appended, which avoids the error.

PR #8579, opened 2025-06-01
cmount: add support for fuse3

jfantinhardesty:

Adds support for fuse3 to the cmount command on Linux and NetBSD based on the recent release from cgofuse https://github.com/winfsp/cgofuse/releases/tag/v1.6.0. This is done by using the `fuse3` build tag which cgofuse uses to link to libfuse3. To build rclone with fuse3 you now can do `go build -tags=cmount,fuse3`

(https://github.com/rclone/rclone/issues/6632)

#### Checklist

### DIFF
--- a/cmd/cmount/mount.go
+++ b/cmd/cmount/mount.go
@@ -34,73 +34,6 @@ func init() {
 	buildinfo.Tags = append(buildinfo.Tags, "cmount")
 }
 
-// mountOptions configures the options from the command line flags
-func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.Options) (options []string) {
-	// Options
-	options = []string{
-		"-o", fmt.Sprintf("attr_timeout=%g", time.Duration(opt.AttrTimeout).Seconds()),
-	}
-	if opt.DebugFUSE {
-		options = append(options, "-o", "debug")
-	}
-
-	if runtime.GOOS == "windows" {
-		options = append(options, "-o", "uid=-1")
-		options = append(options, "-o", "gid=-1")
-		options = append(options, "--FileSystemName=rclone")
-		if opt.VolumeName != "" {
-			if opt.NetworkMode {
-				options = append(options, "--VolumePrefix="+opt.VolumeName)
-			} else {
-				options = append(options, "-o", "volname="+opt.VolumeName)
-			}
-		}
-	} else {
-		options = append(options, "-o", "fsname="+device)
-		options = append(options, "-o", "subtype=rclone")
-		options = append(options, "-o", fmt.Sprintf("max_readahead=%d", opt.MaxReadAhead))
-		// This causes FUSE to supply O_TRUNC with the Open
-		// call which is more efficient for cmount.  However
-		// it does not work with cgofuse on Windows with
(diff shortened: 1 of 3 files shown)


ncw (commented):
Thanks for doing this :-)

I don't like the duplicated `mountOption` function though.

I'd suggest you set constant in your two fuse3 and !fuse3 file say `isFuse3` and then use this to select what to do in `mountOptions` in `mount.go`. A compile time constant will be completely optimized by the compiler so this is nice and efficient and will be easier to maintain I think.

What do you think?

jfantinhardesty:
Great suggestion, that is a lot cleaner. I made those changes so it should be easier to read now.

cmd/cmount/mount.go
```diff
@@ -58,14 +58,17 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
...
+		// The following options are not supported by fuse3
+		if !isFuse3 {
+			options = append(options, "-o", fmt.Sprintf("max_readahead=%d", opt.MaxReadAhead))
+			// This causes FUSE to supply O_TRUNC with the Open
+			// call which is more efficient for cmount.  However
+			// it does not work with cgofuse on Windows with
+			// WinFSP so cmount must work with or without it.
+			options = append(options, "-o", "atomic_o_trunc")
+			if opt.DaemonTimeout != 0 {
+				options = append(options, "-o", fmt.Sprintf("daemon_timeout=%d", int(time.Duration(opt.DaemonTimeout).Seconds())))
```
darthShadow: I believe this is a MacOS only option (on both fuse2 & fuse3), so perhaps this could be moved to the `darwin` if-block below.

cmd/cmount/mount.go
```diff
@@ -58,14 +58,17 @@ func mountOptions(VFS *vfs.VFS, device string, mountpoint string, opt *mountlib.
...
-		// This causes FUSE to supply O_TRUNC with the Open
-		// call which is more efficient for cmount.  However
-		// it does not work with cgofuse on Windows with
-		// WinFSP so cmount must work with or without it.
-		options = append(options, "-o", "atomic_o_trunc")
-		if opt.DaemonTimeout != 0 {
-			options = append(options, "-o", fmt.Sprintf("daemon_timeout=%d", int(time.Duration(opt.DaemonTimeout).Seconds())))
+		// The following options are not supported by fuse3
+		if !isFuse3 {
+			options = append(options, "-o", fmt.Sprintf("max_readahead=%d", opt.MaxReadAhead))
```
darthShadow, line 63: This seems like it should be set in the init response but I don't see any way to set those for cgofuse.  Same with maxwrite, which we default to 1MB for mount2, but skip here.

jfantinhardesty: Yes, I think that a change would need to be added to cgofuse since in fuse3 these options can no longer be specified on the command line.

cmd/cmount/mount.go
jfantinhardesty: Good point, I moved these to the darwin section.

Neustradamus:
@ncw: Any progress on this PR?

PR #8589, opened 2025-06-08
sftp: Add ability to preserve hardlinks within a given synchronizatio…

ab2525:
…n session.

Add support for hardlinks https://github.com/rclone/rclone/issues/6160

[](https://github.com/rclone/rclone/issues/6160)

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/pkg/sftp"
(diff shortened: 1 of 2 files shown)


ncw:
This looks interesting :-)

Let me see if I understand what it does

- rclone uploads a file A
- Later it tries to upload a file B
- If B is hardlinked to A then make a hardlink instead of uploading
- What if B is different to A?

Is that about right? I think the way we do updates shouldn't break hardlinks (maybe it should...)

That probably means that you don't actually need to make a link in step 3 as it is already made (or I've mis-understood something).

Something looks wrong with the code though. You shouldn't be using "syscall" as that isn't portable between OSes.

The sftp library has its own [FileStat](https://pkg.go.dev/github.com/pkg/sftp#FileStat) structure and I'd hope that you can find inode numbers in the `StatExtended` field.

theVerySharpFlat:
> Is that about right? I think the way we do updates shouldn't break hardlinks (maybe it should...)

It seems hardlinks are broken during local to local rclone copy operations without --inplace as the file assumes the inode of the partial file rclone writes to

PR #8604, opened 2025-06-16
local: hardlinks

theVerySharpFlat:
Adds support for hardlinks with the local backend as per #6160 

#6160 
#8589

> If you require using it, you have options.
> 
> 1) Code it

 _Originally posted by @animosity022 in [#6160](https://github.com/rclone/rclone/issues/6160#issuecomment-2772345217)_

#### Checklist

### DIFF
--- a/fs/hlinktracker.go
+++ b/fs/hlinktracker.go
@@ -17,7 +17,8 @@ type HLinkRootInfo struct {
 
 // A generic hardlink tracker for use in filesystems that support hardlinks
 type HLinkTracker struct {
-	hardlinks sync.Map // map[HLinkInfo]*HLinkRootInfo
+	hardlinks   sync.Map // map[HLinkInfo]*HLinkRootInfo
+	dstToSrcMap sync.Map // map[HLinkInfo]HLinkInfo
 }
 
 func (t *HLinkTracker) RegisterHLinkRoot(ctx context.Context, src Object, fsrc Hardlinker, dst Object, fdst Hardlinker, dstPath string, willTransfer bool) (bool, error) {
@@ -39,6 +40,28 @@ func (t *HLinkTracker) RegisterHLinkRoot(ctx context.Context, src Object, fsrc H
 	info, ok := rawInfo.(*HLinkRootInfo)
 	if !ok {
 		return false, fmt.Errorf("HLinkTracker hardlinks map unexpectedly returned non-HLinkRootInfo value")
+	}
+
+	var dstLinkInfo any
+	dstHasLinkInfo := false
+	if !isExisting && dst != nil && !willTransfer {
+		dstLinkInfo, dstHasLinkInfo = fdst.HLinkID(ctx, dst)
+
+		if !dstHasLinkInfo {
+			return false, fmt.Errorf("destination %v unexpectedly has no hlink info", dst)
+		}
+	} else if isExisting && info.remoteHLinkInfo != nil {
+		dstHasLinkInfo = true
+		dstLinkInfo = info.remoteHLinkInfo
+	}
+
+	if dstHasLinkInfo {
+		srcInodeInfoFromDst, haveSrcInodeInfoFromDst := t.dstToSrcMap.LoadOrStore(dstLinkInfo, srcLinkInfo)
+
+		if (haveSrcInodeInfoFromDst && !isExisting) || (haveSrcInodeInfoFromDst && isExisting && info.remoteHLinkInfo != nil && (srcInodeInfoFromDst != srcLinkInfo)) {
+			willTransfer = true
+			isExisting = false
+		}
 	}
 
 	defer info.lock.Unlock()


theVerySharpFlat:
I'd like to add tests, but my check for if the destination filesystem is local depends on the local backend, creating a cyclic dependency. Is there a recommended way to add this functionality to the copy/move routines without modifying every single backend in the project?

ncw:
I think what we need to do is think of a new Fs method which implements this.

You won't have any problems with recursive includes at that point and it means we can use the same implementation for sftp and local.

I'm not sure exactly what the minimum interface is though - perhaps you could write something up and we could discuss.

theVerySharpFlat:
> I think what we need to do is think of a new Fs method which implements this.
> 
> You won't have any problems with recursive includes at that point and it means we can use the same implementation for sftp and local.
> 
> I'm not sure exactly what the minimum interface is though - perhaps you could write something up and we could discuss.

Sounds good. I just wanted the maintainer goahead before I extended the scope of my changes

i30817:
Hardlinks don't work (more specifically, can't be created) in android without root. I'm just mentioning it because it's a common failure mode of trying to do exact mirroring Unix filesystems in unrooted android, not that I tested your code fails, or even understand it.

theVerySharpFlat:
> Hardlinks don't work (more specifically, can't be created) in android without root

interesting...is this a filesystem limitation or just typical android sandboxing?

i30817:
It's android sandbox. Android itself uses extfat (for sdcards) and ext4fs. It was a unpleasant surprise to me too when I tried to use a dedup rust program (another that wasn't aware of this) and got several errors. 

Android sandbox in general is full of filesystem limitations that severely limit android appeal for Linux usecases, google and the hardware companies really want to keep the good stuff to themselves to monetize it (including mount and annoyingly, FUSE and -i think- extended attributes). Therefore you can't system mount a networked filesystem or loop device, so no opening isos, squash files etc. Applications can work around _that_ but only by reimplementing the protocol client inside the app, and only for themselves (sandbox) which is why VLC absurdly has a nfs and samba client, and people want a webdav client too (the user only sees them in the form of using specially formatted urls - nfs has bugs if you use ':' in music files and those songs get skipped, probably because the uri parser they're using gets confused about this hack).

Can't have users use webdav by themselves instead of paying a remote service for it, or make their small 256gb drive fit in more data with a specialized squash drive instead of buying a new device, now can they lol. 

It's why RSAF is so useful, in spite of it being a slow hack, it's using rclone and storage access framework to work around those limitation in many apps that use the native android filechooser or the SAF API in general (if it's more than one file they need). I'm excited for the new version of rclone because it has a userspace squash mount in the works, so it _should_ work with RSAF soon, which will allow me to mount a 100gb squash file holding scummvm games in a 128gb tablet.

theVerySharpFlat:
> It's why RSAF is so useful, in spite of it being a slow hack

I'm doing large volume data migration for a compute cluster where performance is critical. Hopefully I'll land some performance PRs in the coming weeks (though none of my previous PRs have been merged up to this point)

> I'm excited for the new version of rclone because it has a userspace squash mount in the works, so it _should_ work with RSAF soon, which will allow me to mount a 100gb squash file holding scummvm games in a 128gb tablet.

Best of luck!

theVerySharpFlat:
@ncw This PR is feature complete and has been used in production with validation every step of the way for the past month with no problems except those addressed in https://github.com/rclone/rclone/pull/8604/commits/e411e890dd19f6bd03fd205cce4483aa44257f47

Do you mind giving this a review? In terms of merging, do we mind using github's squash and merge feature?

In terms of design, I created the `Hardlinker` FS interface for all remotes that want to support this. Then, I created a reference implementation called hlinktracker which remotes can use in implementing `Hardlinker` if they so choose. Integrating this into SFTP shouldn't be terribly hard given the similarities with the `local` backend---however, more exotic backends like gdrive might want to roll their own implementation.

JakubOrsula:
>Integrating this into SFTP shouldn't be terribly hard given the similarities with the local backend---however, more exotic backends like gdrive might want to roll their own implementation.

I reviewed your code as far as I can see you are using 
```
UnixHLinkInfo{
    dev: uint64,    // Device ID (st.Dev)
    ino: uint64,    // Inode number (st.Ino)
}
```
as key to your map. Which means hard links are preserved only across same device. I guess it works if you are running large raids, but for cross-device transfer it won't work as the device id will be different.

Same goes for sftp, inode numbers will not be preserved between source and destination.

Otherwise it is very nice and clean implementation 👍

theVerySharpFlat:
[04e48ce](https://github.com/rclone/rclone/pull/8604/commits/04e48ce0f178d7076e6353c2c66a4d346dd0d76e) fixes a bug where links are broken on the source by a program that maintains the modtime

This leads to a situation where structures like this
```
a-b c-d (src)
a-b-c-d (dest)
```
Are maintained even though the link state is wrong. Thus, we add a reverse map which we use to look for inconsistencies. @JakubOrsula you probably want this commit in your fork. Thanks for working with me!

JakubOrsula:
@theVerySharpFlat thanks for the commit. I have tried in my free time to incorporate it, and good news is that hardlinking works with sftp with some caveats.

You can create hardlinks on almost all sftps. What you can't do is get inode numbers or anything else relevant for UnixHLinkInfo structure. This is sort of fine, as most users are just looking for ways to push their local hardlinks to remote and not the other way around. Of course, if you have sftp with shell access you can preserve hardlinks during the download. If not it is downloaded as copies.

I have left one comment in the PR regarding needing to register the links if they do not need to be transferred. 

Are you going to try to get your changes merged? I would like to get the sftp changes merged but since I am building on top of you I would like to not interfere in that.

Your latest commit is relying on the remote info, which I do not have in the case of chrooted, shell-less sftp so I am not using it yet.

fs/hlinktracker.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// If this case is true, we have fully transferred the root and can immediately link
+		if info.remoteHLinkInfo != nil {
+			if info.remoteHLinkInfo != dstLinkInfo {
+				dstPathNoSuffix := strings.TrimSuffix(dstPath, ".rclonelink")
+				srcPathNoSuffix := strings.TrimSuffix(info.remotePath, ".rclonelink")
+				Debugf(fdst, "performing hardlink %v->%v", srcPathNoSuffix, dstPathNoSuffix)
+
+				err := fdst.HLink(ctx, srcPathNoSuffix, dstPathNoSuffix)
```
JakubOrsula, line 90: @theVerySharpFlat why do the hardlink here even when the `willTransfer` is `false`? I guess it is very cheap operation on local fs, but over sftp it is more expensive.

theVerySharpFlat: Due to this condition: `info.remoteHLinkInfo != dstLinkInfo`. Generally, rclone's will transfer is dictated by modification time. Thus, if you have a situation as follows
```
SRC:     A-B
DST:     A B
```
where modification times are identical on the source and destination, this conditional ensures A and B on the destination are linked together.

theVerySharpFlat:
> You can create hardlinks on almost all sftps. What you can't do is get inode numbers or anything else relevant for UnixHLinkInfo structure. This is sort of fine, as most users are just looking for ways to push their local hardlinks to remote and not the other way around.

In this case, I guess your only option is to return a UUID for every destination HLinkInfo and force relinking (unideal, but what else can you do?)

> Are you going to try to get your changes merged? I would like to get the sftp changes merged but since I am building on top of you I would like to not interfere in that.

I have been, but the maintainer has been silent.

> Your latest commit is relying on the remote info, which I do not have in the case of chrooted, shell-less sftp so I am not using it yet.

Yeah. You can put in a flag that, depending on the backend, disables the reverse check and forces relinking regardless.

fs/hlinktracker.go
JakubOrsula: Well, it is at the price of some unneeded syscalls, but I understand it is reasonable on local, maybe on sftp as well. I will experiment and see what are the real world numbers. Thanks for the response.

theVerySharpFlat: Hmmm. I don't see how link(2) is uneeded here given that inconsistent state between the source and destination is unacceptable. Certainly, this check is irrelevant on backends that don't support destination hlinkinfo. Can you elaborate?

Rid:
@JakubOrsula is the build at https://github.com/JakubOrsula/rclone/tree/feature/sftp-hardlinks in a working state? I'd like to use hardlinks with SFTP and it doesn't look like this is going to be merged any time soon.

JakubOrsula:
Yep it works, feel free to use it. 

It has some caveats, without shell access it preserves the links only when uploading to sftp.

PR #8611, opened 2025-06-18
Avoid failed to transform error messages when running any rclone command

### LINKED ISSUE #8610: Irrelevant `ERROR` messages when running any `rclone` command
### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/invoke-of-1-70-0-complains-of-echo-not-found/

`ERROR` messages about failed transformation are printed when running any `rclone` command.

Windows 11, 64 bit

```
PS C:\> rclone version
2025/06/18 00:00:00 ERROR : Failed to transform: : error running command "echo stories": exec: "echo": executable file not found in %PATH%
2025/06/18 00:00:00 ERROR : stories/The Quick Brown Fox!.txt: number of path segments must match: stories/The Quick Brown Fox!.txt (1),  (0)
rclone v1.70.0
- os/version: Microsoft Windows 11 Home 24H2 24H2 (64 bit)
- os/kernel: 10.0.26100.4351 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.24.4
- go/linking: static
- go/tags: cmount
```

#### How to use GitHub

albertony:
Fixes #8610

Very much a quickfix.

#### Checklist


albertony:
Maybe instead we could have used a GOOS conditional.

On the other hand, isn't it a bit unfortunate that the rclone need to run through all the example transforms at every startup?

ncw:
Thanks for looking at this @albertony . Annoying to have missed this in the release!

I don't think it should be running through the transform examples at all though.

I'm going to fix this properly by generating the docs in advance like we do in various other places which is a bigger change but will work better I think.

PR #8614, opened 2025-06-18
Docs improvement transform

albertony:

Mostly cleanup of the CLI help text output formatting:
- Use `## header` style instead of `## header ##`, which is more inline with rest of the docs (at least we are moving in that direction).
  - E.g. https://github.com/rclone/rclone/blob/9d464e8e9a56a988bbb5f701006173c9bbefeb6a/cmd/sync/sync.go#L185
- One empty line between sections, not two or three.
- One empty line before code blocks.
- Avoid leading space character for all except the first encoding mask entry.
- Partly as a result of the above changes, avoid using `fmt.Sprintln` for plain string types.

Note that the first commit, the one in `lib/transform/cmap.go`, I am not too sure about: I noticed in the help output that there was an empty initial element:

>Char maps:
>\```
>
>IBM-Code-Page-037
>IBM-Code-Page-437
>IBM-Code-Page-850
>...
>\```

I concluded this was not intentionally part of the returned string list, tests are ok without it, so I removed it. But maybe I'm wrong, and should instead just have trimmed it from the help output?

#### Checklist


ncw:
I fixed the windows "echo" problem - this involved generating the docs with `go generate` and embedding them.

This unfortunately has created a conflict in this code, but apart from renaming `help.go` to `gen_help.go` I think your changes should apply easily.

albertony:
Great. I've rebased and fixed up the PR, so it should be ready for a review.

PR #8619, opened 2025-06-19
local sparse copying

theVerySharpFlat:

Ensure sparse blocks are maintained as well as possible in local to local transfers.

To do so, I use the FS_IOC_FIEMAP ioctl on linux to list out all of the non-sparse extents, which are then passed to the multi threaded transfer system, which has been slightly modified to accept non-constant size chunks.

https://forum.rclone.org/t/possible-to-use-rclone-with-sparse/45189

#### Checklist

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -25,6 +25,7 @@ import (
 	"github.com/rclone/rclone/fs/filter"
 	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/hash"
+	"github.com/rclone/rclone/fs/sparse"
 	"github.com/rclone/rclone/lib/encoder"
 	"github.com/rclone/rclone/lib/file"
 	"github.com/rclone/rclone/lib/readers"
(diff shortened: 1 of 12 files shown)


ncw:
We can't use CGO in rclone - it complicates the multiplatform build too much.

Is there another way of doing this?

Perhaps we should just check for zero bytes in the writer and create sparse files then? I'm pretty sure that is what rsync does with

       --sparse, -S             turn sequences of nulls into sparse blocks

That would be a much smaller patch.

theVerySharpFlat:
> We can't use CGO in rclone - it complicates the multiplatform build too much.
> 
> Is there another way of doing this?
> 
> Perhaps we should just check for zero bytes in the writer and create sparse files then? I'm pretty sure that is what rsync does with
> 
> ```
>    --sparse, -S             turn sequences of nulls into sparse blocks
> ```
> 
> That would be a much smaller patch.

Yeah, but let's admit, it's a little less cool. That's probably a good thing though. I'll get to work on that

theVerySharpFlat:
@ncw [1673acc](https://github.com/rclone/rclone/pull/8619/commits/1673accbba0c7319d3b9a629cecee88402cc4bd6) implements rsync-style sparse preservation.

This thing depends on openChunkWriterfromOpenWriterAt for multithreaded copies and WriterAt in general. For single threaded copies, I've only gone in and implemented this for the local backend by changing the backend-specific update function. Is this alright? I think the usage of fs/operations in backend/local might create a circular dependency in the tests (again). IMHO, it would be the smoothest to force multithreaded transfers (default 1 thread) when --sparse is specified.

Second, tests for sparseness may be hard to do as sparseness is very fs/os version dependent. How do we want to proceed given this?

theVerySharpFlat:
@ncw this PR is ready for review!

theVerySharpFlat:
reopening in a new PR as the scope and implementation of this has expanded

PR #8638, opened 2025-06-25
feat(sftp-http-proxy): add proxy-authorization http header

ndessart:

Add HTTP proxy authentication support to the SFTP HTTP proxy backend.
Thinking about it, this is more a bugfix rather than a feature (the username:password syntax for the proxy is already documented).

No

#### Checklist

- [ ] I have added tests for all changes in this PR if appropriate. --> _there is currently no tests for the http proxy._

### DIFF
--- a/docs/content/sftp.md
+++ b/docs/content/sftp.md
@@ -1186,6 +1186,12 @@ URL for HTTP CONNECT proxy
 
 Set this to a URL for an HTTP proxy which supports the HTTP CONNECT verb.
 
+Supports the format http://user:pass@host:port, http://host:port, http://host.
+
+Example:
+
+    http://myUser:myPass@proxyhostname.example.com:8000
+
 
 Properties:
 
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -204,6 +204,12 @@ Example:
 			Help: `URL for HTTP CONNECT proxy
 
 Set this to a URL for an HTTP proxy which supports the HTTP CONNECT verb.
+
+Supports the format http://user:pass@host:port, http://host:port, http://host.
+
+Example:
+
+    http://myUser:myPass@proxyhostname.example.com:8000
 `,
 			Advanced: true,
 		}, {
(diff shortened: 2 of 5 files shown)


ncw:
> Thinking about it, this is more a bugfix rather than a feature (the username:password syntax for the proxy is already documented).

Does the user:pass syntax not work?

Do you have a link to an RFC or similar for the proxy-authorization http header?

Thanks

ndessart:
> Does the user:pass syntax not work?

No it currently does not work, I've learned it the hard way!
It is parsed by rclone but not sent to the HTTP proxy.

I've confirmed this issue using wireshark and comparing the requests generated by `curl --proxy $HTTP_PROXY` and rclone before identifying the root cause in rclone/lib/proxy/http.go.

> Do you have a link to an RFC or similar for the proxy-authorization http header?

Sure: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Proxy-Authorization

ndessart:
Update: I've read the contribution guidelines and updated this PR accordingly.

Please let me know if I you need me to rebase and squash or anything else. Thank you

docs/content/sftp.md
```diff
@@ -1071,6 +1071,11 @@ URL for HTTP CONNECT proxy
 
 Set this to a URL for an HTTP proxy which supports the HTTP CONNECT verb.
 
+Supports the format http://user:pass@host:port, http://host:port, http://host.
```
ncw, line 1189: Can you change the ftp docs too please which support this.

ncw (commented):
This looks great - thank you.

Yes, now I think about it we need to send the user/pass to the proxy otherwise it won't work!

This needs a doc change in the FTP backend too - I think that is all.

docs/content/sftp.md
ndessart: Done

ndessart:
> This needs a doc change in the FTP backend too - I think that is all.

I've just added the ftp-http-proxy documentation

ncw (approved):
Sorry for the delay @ndessart 

I've rebased this and put the docs in the correct place so it is ready to merge.

Thank you :-)

PR #8641, opened 2025-06-26
fix: add configurable bar for name truncation for partial files

theVerySharpFlat:

Currently, rclone truncates paths over a hundred characters to fit a partial suffix. Sometimes, this comes around and bites you when it turns out the ending, say 10 characters, are the only thing distinguishing one file being transferred from another. Thus, I have added this flag to make this limit not hardcoded.

#### Checklist

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -2116,6 +2116,12 @@ Suffix length limit is 16 characters.
 
 The default is `.partial`.
 
+### --partial-pathmax PATHMAX
+
+Rclone truncates the names of partial files by the length of the partial suffix (formatted as `.hash.partial`) if the base path exceeds `--partial-pathmax` characters in attempt to respect remote path maximums. If `--partial-pathmax` is set to 0, truncation to fit the suffix is disabled entirely.
+
+The default is 100 characters
+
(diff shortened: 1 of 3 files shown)


ncw:
Rclone adds a hash of the file fingerprint to create a unique suffix. I guess if you have a lot of files with the same size/modtime/hash this could cause a problem.

Perhaps rclone should include the filename in the hash too? That might be an easier fix.

theVerySharpFlat:
> Rclone adds a hash of the file fingerprint to create a unique suffix. I guess if you have a lot of files with the same size/modtime/hash this could cause a problem.
> 
> Perhaps rclone should include the filename in the hash too? That might be an easier fix.

Yeah, our usecase is tens of terrabytes of data spread across tens to hundreds of millions of files where checksumming is a bottleneck, so we have that disabled. We also have very long and similar filenames

I'd rather have the thing fail due to exceeding PATH_MAX instead of writing to the same partial file concurrently

theVerySharpFlat:
> Perhaps rclone should include the filename in the hash too? That might be an easier fix.

In the fingerprint, yes. I'll sub a PR for that

theVerySharpFlat:
closing in favor of #8643

PR #8643, opened 2025-06-26
fs: add remote path to object fingerprint

theVerySharpFlat:
follow up to #8641, adds remote path to fingerprints in order to distinguish from other similar files, especially when hashes aren't available.

#### Checklist


theVerySharpFlat:
This is probably a bandaid for a larger problem. In reality, we should somehow atomically check if a file exists *and* create it if it doesn't exist and fail if it does (and reroll the name deterministically, perhaps), both as a single atomic operation. IIRC, there's a metadata lock in the local backend that we could use, but this is more of a backend-agnostic problem

PR #8648, opened 2025-06-28
march: fix deadlock in --no-traverse when all transfers skipped using sequence numbering

jeremy:

When every source object already exists at the destination and --no-traverse is set, march deadlocked after 100 entries because dstChan lagged srcChan. Buffer all sources and destinations, then replay them through balanced channels. Introduced in v1.70.0 with callback-based syncing.

See #8647 for the revert-to-old-behavior fix and repro script. This PR attempts a fix that retains streaming behavior.

* Replaces slice-accumulating sorter with O(checkers) streaming
* Each incoming src entry gets a monotonic sequence number; a map buffers out-of-order destination HEAD results so dstChan is replayed in sequence order
* Directories get a zero-cost placeholder to guarantee 1:1 src/dst event flow without blocking

Using repro script from #8647:
```bash
Using rclone: ./rclone
Version: rclone v1.71.0-DEV

Creating 200 test files...
Initial copy to destination...
Testing --no-traverse with all files existing (10s timeout)...
Checks:               200 / 200, 100%, Listed 200
Elapsed time:         0.0s

✅ PASS: Command completed successfully
```

* https://github.com/rclone/rclone/pull/8647
* https://forum.rclone.org/t/1-70-1-appears-to-hang-when-using-files-from-rclone-ls-no-traverse-and-ignore-existing/51720
* https://forum.rclone.org/t/deadlock-when-using-no-traverse-max-duration-and-files-from-with-many-files-100/51650

#### Checklist

### DIFF
--- a/fs/march/march.go
+++ b/fs/march/march.go
@@ -420,39 +420,119 @@ func (m *March) processJob(job listDirJob) ([]listDirJob, error) {
 	// for each item in the srcList to head dst object
 	if m.NoTraverse && !m.NoCheckDest {
 		originalSrcChan := srcChan
-		srcChan = make(chan fs.DirEntry, 100)
-		ls, err := list.NewSorter(m.Ctx, m.Fdst, list.SortToChan(dstChan), m.dstKey)
-		if err != nil {
-			return nil, err
-		}
+		srcChan = make(chan fs.DirEntry, ci.Checkers*2)
 
+		// Use sequence numbers for ordering - simpler than sorter
+		// This assumes source entries arrive in the same order as destination keys
+		// would be sorted, which is true for all current backends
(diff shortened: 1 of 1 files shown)


jeremy:
Nice approach! Dig passing the "slot" channel down the pipeline so we get ordering for free, by construction. No notes.

ncw:
> Nice approach! Dig passing the "slot" channel down the pipeline so we get ordering for free, by construction. No notes.

Thanks for the review @jeremy . Sorry about not merging your code - I started of noodling with the chan of chans idea to give you feedback on a simpler implementation but I couldn't get it to work and by the time I did get it to work I had implemented it completely!

PR #8669, opened 2025-07-08
Use consistent markdown heading syntax

albertony:
I think consistently using `#` prefixed header format for headers - at any level - is best. I find the alternative format for level 1 and 2 headers, `====` and `-----`, is often just confusing - at least the level 2 which is typically mixed in between other level headers. And at least when, as was the case in the usage docs, there is a mix of both - inconsistent use of both variants of level 2 header format.

The situation in docs.md was:
- The level 1 header is written as:
  > \# Usage
- Some level 2 headers:
  > Configure
  > \---------
- Other level 2 headers:
  > \## Metadata support {#metadata}
- And, obviously, all lower level headers in the only supported format:
  > \### Valid remote names

Edit: When running [markdownlint](https://github.com/DavidAnson/markdownlint) with default configuration, this is covered by the [MD003](https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md003.md) rule:

> MD003/heading-style: Heading style [Expected: atx; Actual: setext]

#### Checklist

### DIFF
--- a/docs/content/changelog.md
+++ b/docs/content/changelog.md
@@ -49,7 +49,7 @@ description: "Rclone Changelog"
 * New Features
     * Add [`--max-connections`](/docs/#max-connections-n) to control maximum backend concurrency (Nick Craig-Wood)
     * Add [`--max-buffer-memory`](/docs/#max-buffer-memory) to limit total buffer memory usage (Nick Craig-Wood)
-    * Add transform library and [`--name-transform`](/docs/#name-transform-command-xxxx) flag (nielash)
+    * Add transform library and [`--name-transform`](/docs/#name-transform-stringarray) flag (nielash)
     * sync: Implement [`--list-cutoff`](/docs/#list-cutoff) to allow on disk sorting for reduced memory use (Nick Craig-Wood)
     * accounting: Add listed stat for number of directory entries listed (Nick Craig-Wood)
     * backend: Skip hash calculation when the hashType is None (Oleksiy Stashok)
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -12,8 +12,7 @@ what the [basic syntax](#basic-syntax) looks like, describes the
 various [subcommands](#subcommands), the various [options](#options),
 and more.
 
-Configure
----------
+## Configure
 
 First, you'll need to configure rclone.  As the object storage systems
 have quite complicated authentication these are kept in a config file.
@@ -92,8 +91,7 @@ See the following for detailed instructions for
 - [Zoho WorkDrive](/zoho/)
 - [The local filesystem](/local/)
 
-Basic syntax
------
+## Basic syntax
 
 Rclone syncs a directory tree from one storage system to another.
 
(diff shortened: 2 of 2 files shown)


albertony:
I sneaked in an additional commit: Change parameter type to the more consistent `stringArray` for the new `--name-transform` option, that has been introduced recently and therefore was missed by https://github.com/rclone/rclone/pull/8316.

albertony:
And another one.. 😮‍💨
Fixing the header level of `--dump`, which was effectively a sub-section of `--memprofile` (but even before my recently merged PR it was at the same level, but then subsection of `--cpuprofile`).

darthShadow:
Perhaps we should be using some kind of linting tool for markdown too, to keep any changes consistent. Something like https://github.com/DavidAnson/markdownlint with https://github.com/DavidAnson/markdownlint-cli2 via https://github.com/DavidAnson/markdownlint-cli2-action.

albertony:
That is a very good idea! And that "toolset" from David Anson seem to be a very good candidate, thanks for the tip!

We have a lot of markdown text coming from go source files, which it will probably not be able to cover out of the box, but covering the main .md files would be a good start.

albertony:
The same author has also created an extension for VSCode (https://github.com/DavidAnson/vscode-markdownlint), which is also based on markdownlint-cli2, and is a perfect supplement to the GitHub Action (markdownlint-cli2-action).

When running it with default configuration on current master, I see that my header changes here are covered by the [MD003](https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md003.md) rule:

> MD003/heading-style: Heading style [Expected: atx; Actual: setext]

albertony:
I will take a chance and merge this PR without waiting for review, I think it is pretty safe, because then I can make a new PR adding linter and fix reported issues from that etc, which I think will be more clean approach..

ncw:
Very happy with this :-)

And great idea for a linter.

I've been gradually pulling out the huge chunks of markdown into .md files which get embedded. They are much easier to edit like that and will lint too!

albertony:
Sounds good, and that we are moving in a right direction.

I'm not sure yet what to do with markdown text in source codes, and the resulting checked in generated markdown files (like the command docs) and generated sections of markdown files (like in the backend docs). I mean, how a PR lint check should treat issues in these etc.. But to start with I am just considering the main, plain markdown files, like docs.md, install.md etc. And then we can iterate on that.

I've found one limitation, though: The linter does not always match exactly what Hugo does when rendering html. For example, header anchors in hugo strips off leading `-`. When linking to a header with text `--config string` we therefore uses anchor `#config-string` in Markdown, which gets us to https://tip.rclone.org/docs/#config-string. But this does not work on GitHub's rendering, where it should have been `#--config-string` (https://github.com/rclone/rclone/blob/master/docs/content/docs.md#--config-string), and because of this the linter complains about it: ~`MD038/no-space-in-code: Spaces inside code span elements`~ edit: `MD051/link-fragments: Link fragments should be valid`.

ncw:
> But to start with I am just considering the main, plain markdown files, like docs.md, install.md etc. And then we can iterate on that.

Sounds sensible.


This error



> `MD038/no-space-in-code: Spaces inside code span elements`.

https://github.com/DavidAnson/markdownlint/blob/main/doc/md038.md

Doesn't seem to be about anchors?

albertony:
Sorry - copy/paste error in that comment! 😊 

Should be:

> MD051/link-fragments: Link fragments should be valid

PR #8671, opened 2025-07-09
fs: correct conditions of DoneTransferring. fixes #8670

### LINKED ISSUE #8670: The fs/accounting/stats module fails to function correctly when transfers==1.
The stats counter reseting in accounting/stats.go is triggered when.
The transfer/checker worker counter all is 0.

When transfers=1 is configured, after the checker completes its scan and exits,
the transfer switching triggers this reset condition.

https://github.com/rclone/rclone/blob/b7c6268d3e3f33ab3c6ca2bc0d2f98a60ed72105/fs/accounting/stats.go#L830-L835

This should include a check to see if the transfer queue is empty.

Another related bug: the state management for stats.average
is also broken. The _stopAverageLoop method doesn't reset
the s.average.started flag.

Transfer speed and ETA time is not update correctly.

```
rclone v1.71.0.1
- os/version: darwin 12.7.6 (64 bit)
- os/kernel: 21.6.0 (x86_64)
- os/type: darwin
- os/arch: amd64
- go/version: go1.24.4
- go/linking: dynamic
- go/tags: none
```

LInux & Mac OSX

```
rclone copy -Pv --transfers=1 --bwlimit=10m a b
```

is:

Fix issue #8670 

#### Checklist


ncw:
@darthShadow do you want to give this a review?

fs/accounting/stats.go
```diff
@@ -829,7 +830,8 @@ func (s *StatsInfo) DoneTransferring(remote string, ok bool) {
 		s.transfers++
 		s.mu.Unlock()
 	}
-	if s.transferring.empty() && s.checking.empty() {
+
+	if s.transferring.empty() && s.checking.empty() && s.transferQueue == 0 {
```
darthShadow: This goes back again to the earlier point of how we want to calculate the averages. Whether only for the duration that something is actually transferring or even during the period when something is queued but not actually being transferred (possibly with `--check-first`?)

is: I can roughly get your point... The problem is that after _stopAverageLoop, the loop cannot be restarted because s.average.started was not reset.

I'll revert the changes here, keep the _stopAverageLoop modifications, and resubmit the PR.

darthShadow (approved):

fs/accounting/stats.go
darthShadow: Thanks. I will keep this conversation open for @ncw to take a look in case he feels differently but otherwise, it looks fine and I have approved the change.

ncw:
Thanks for doing this @is and thanks for the review @darthShadow 

I'll merge this now.

PR #8678, opened 2025-07-11
Markdown format lint

albertony:

The purpose is to make sure our Markdown files are using valid syntax. But also, the more challenging task: That they use consistent syntax where there are multiple Markdown dialects or there are multiple valid alternatives.

I ran a linter tool on a selected set of Markdown files, adjusted configuration to something I found reasonable, fixed the reported problems and committed the changed Markdown files. Then I added a step to our pull request lint check that runs the same automatically.

The linter used in CI is the GitHub Action created by David Anson: https://github.com/DavidAnson/markdownlint-cli2-action. It is based on the command line tool https://github.com/DavidAnson/markdownlint-cli2, which in turn is based on the linter library https://github.com/DavidAnson/markdownlint, which in turn is built around the parser library https://github.com/micromark/micromark. There is also a VSCode extension https://github.com/DavidAnson/vscode-markdownlint, which is based on the same markdownlint-cli2 tool as the GitHub Action, and therefore also shares the same linter rule configuration.

Linter rules are configured in `.markdownlint.yml` at the root.

I should probably also write something in the contributing documentation, but I can do that once this is settled.

PS: Currently contains the changes from https://github.com/rclone/rclone/pull/8676 to avoid git conflicts, if not that gets merged first I can remove them.

Based on suggestion in https://github.com/rclone/rclone/pull/8669#issuecomment-3055648783, and following discussion.

#### Checklist


albertony:
I just had an idea: What if we executed the Markdown updating locally first, couldn't we then perform linting on all the markdown files - including those that are autogenerated and contain autogenerated sections?

I think we need to build the rclone binary then, since the command docs updating procedure needs to run it, but with github caching and all the previous golangci-lint steps, maybe creating the rclone binary is not that much extra effort (i.e. time to run the lint job).

Shouldn't doing the following be enough to generate all relevant Markdowns (as `MANUAL.md` target depends on `commanddocs`, `backenddocs` and `rcdocs`)?

```sh
make && make MANUAL.md
```

What do you think @ncw?

albertony:
Another thing... 

I found the reason for the header anchors format not being compatible with GitHub. We have configured Hugo to use the the legacy [blackfriday](https://gohugo.io/functions/urls/anchorize/#blackfriday) variant of anchorize function:

https://github.com/rclone/rclone/blob/5e539c6a7209c2e797966b6478d382bc3d495386/docs/config.json#L30

This is:

> Provided for backwards compatibility with Hugo v0.59.1 and earlier. This option will be removed in a future release.

Hugo version [0.60](https://github.com/gohugoio/hugo/releases/tag/v0.60.0), released Nov 27, 2019 replaced the Markdown processor, from "Blackfriday" to the current "Goldmark":

> [Goldmark](https://github.com/yuin/goldmark/) by [@yuin](https://github.com/yuin) is now the new default library used for Markdown in Hugo. It's CommonMark compliant and GitHub flavored, and both fast and flexible. Blackfriday, the old default, has served us well, but there have been formatting and portability issues that were hard to work around. The "CommonMark compliant" part is the main selling feature of Goldmark, but with that you also get attribute syntax on headers and code blocks (for code blocks you can turn on/off line numbers and highlight line ranges), strikethrough support and an improved and configurable implementation of TableOfContents.

Switching to the default anchorize function, should let us lint including the rule "MD051/link-fragments". The default is named "github":

> Compatible with GitHub. This is the default.

Getting a verification of the links is perhaps one of the more important advantages of doing linting on the Markdowns. And with the statement that the legacy option we use are being removed, maybe we should consider switching to the default? Would be some work, of course, but with help of the linter and some magic regex searches etc, should be doable. The major downside is that direct links to headers on rclone.org from external sites would break... 

Edit: I added a commit to this PR where I updated the section about writing documentation in the contributing docs, with a note regarding the above finding and also some more other details..

ncw:
> I just had an idea: What if we executed the Markdown updating locally first, couldn't we then perform linting on all the markdown files - including those that are autogenerated and contain autogenerated sections?
> 
> I think we need to build the rclone binary then, since the command docs updating procedure needs to run it, but with github caching and all the previous golangci-lint steps, maybe creating the rclone binary is not that much extra effort (i.e. time to run the lint job).

Building the rclone binary takes not a lot of time.
 
> Shouldn't doing the following be enough to generate all relevant Markdowns (as `MANUAL.md` target depends on `commanddocs`, `backenddocs` and `rcdocs`)?
> 
> ```shell
> make && make MANUAL.md
> ```

`make doc` is what I would usually type. That should build rclone as part of `rcdocs`

Ideally I'd be able to run this locally too as I rely on `make check` to tell me if I've messed something up, say `make docscheck` or something like that.

I can work on that later if you want; it will be slightly complicated by the fact that `make doc` changes checked in files.

albertony:
> I've looked carefully through those changes and they look fantastic to me :-)

Thanks! 😊 

> I see how many ways I've abused markdown over the years and I see the linter caught some nice bugs too.

Well, we can blame some on Markdown historically lacking a formal syntax definition, and also that there are different "dialects", and to some degree ambiguity (like writing lists and code blocks), so this is not so much about fixing invalid Markdown but more about making it consistent across the project.

> Building the rclone binary takes not a lot of time.
> 
> > Shouldn't doing the following be enough to generate all relevant Markdowns (as `MANUAL.md` target depends on `commanddocs`, `backenddocs` and `rcdocs`)?
> > ```shell
> > make && make MANUAL.md
> > ```
> 
> `make doc` is what I would usually type. That should build rclone as part of `rcdocs`

Ah yes, only need the single command, rclone will be built as a dependency. Actually, `make MANUAL.md` is the minimal command when only Markdown content are relevant. The `make doc` in addition generates `rclone.1`, `MANUAL.html` and `MANUAL.txt`, which shouldn't matter much, however that relies on pandoc command so if only done to be able to do markdown linting we might want to save the few seconds by not installing pandoc on the GitHub runner and run it, although probably not a big difference.

> Ideally I'd be able to run this locally too as I rely on `make check` to tell me if I've messed something up, say `make docscheck` or something like that.

I see. There are extensions for VSCode, Sublime, Vim and Emacs, listed in <https://github.com/DavidAnson/markdownlint>. However, it is not the same as being able to run a `make docscheck` command, I see your point. It would be easy to just run the official CLI tool <https://github.com/DavidAnson/markdownlint-cli2>, unfortunately it is a Node.js tool so need to be `npm install`-ed (a single-binary executable would be nice).

> I think we should merge this ASAP before it bitrots - please do so when happy!

Let's do this. I've rebased this PR on master, and added a small note in the contribution docs regarding the markdown linter, so I will merge it soon...

> I can work on that later if you want; it will be slightly complicated by the fact that make doc changes checked in files.

And yes, lets iterate further on the remaining parts later. Having a strategy for the generated parts would be really nice. Unless there are regressions or we see any other important issues that are important, I will probably not be able to start on new tasks for a couple of weeks, so feel free to do so.. 🌴

Issue #8683, opened 2025-07-13
Abnormal Memory Consumption After Upgrading to AWS Go SDK v2 While Mounting Ceph S3


VVoidV:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/abnormal-memory-consumption-in-rclone-after-upgrading-to-aws-go-sdk-v2-while-mounting-ceph-s3/51926

#### What is the problem you are having with rclone?
The memory usage keeps growing. After creating approximately 400,000 files and then deleting them all, the RSS memory usage remains at 5.0GB. However, vfs/stats shows that there are almost no files left in the file system.

#### What is your rclone version (output from `rclone version`)
```
[root@lbr-rclone ~]# rclone version
rclone v1.70.3

os/version: almalinux 9.5 (64 bit)
os/kernel: 5.14.0-503.23.2.el9_5.x86_64 (x86_64)
os/type: linux
os/arch: amd64
go/version: go1.24.4
go/linking: static
go/tags: none
```

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
```
os/version: almalinux 9.5 (64 bit)
os/kernel: 5.14.0-503.23.2.el9_5.x86_64 (x86_64)
```

#### Which cloud storage system are you using? (e.g. Google Drive)
```
ceph RGW
```

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

`/usr/bin/rclone mount xxx:xxx /mnt/rc --vfs-cache-mode full --cache-dir /cache --log-file /var/log/rc.log --log-level info --uid 1000 --gid 1000 --allow-other --rc --rc-addr=127.0.0.1:5572`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

`nothing interesting threre...`

#### vfs/stats
```
{
        "diskCache": {
                "bytesUsed": 3295,
                "erroredFiles": 0,
                "files": 1,
                "hashType": 1,
                "outOfSpace": false,
                "path": "/cache/vfs/xxx/xxx",
                "pathMeta": "/cache/vfsMeta/xxx/xxx",
                "uploadsInProgress": 0,
                "uploadsQueued": 0
        },
        "fs": "xxx:xxxx",
        "inUse": 1,
        "metadataCache": {
                "dirs": 9,
                "files": 14
        },
        "opt": {
                "BlockNormDupes": false,
                "CacheMaxAge": 3600000000000,
[...]

PR #8684, opened 2025-07-14
backend/s3: Fix memory leak by cloning strings #8683

VVoidV:
To addresses a potential memory leak in the S3 backend where 
strings extracted from large API responses were keeping the entire 
response in memory. The issue occurs because Go strings share underlying 
memory with their source, preventing garbage collection of large XML 
responses even when only small substrings are needed.

#8683 

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -4458,7 +4458,7 @@ func (f *Fs) list(ctx context.Context, opt listOpt, fn listFn) error {
 		}
 		foundItems += len(resp.Contents)
 		for i, object := range resp.Contents {
-			remote := deref(object.Key)
+			remote := *stringClone(deref(object.Key))
 			if urlEncodeListings {
 				remote, err = url.QueryUnescape(remote)
 				if err != nil {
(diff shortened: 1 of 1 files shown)


VVoidV:
After exploring the code and analyzing the heap profile, I found that some of our strings in `func (o *Object) setMetaData(resp *s3.HeadObjectOutput) ` are not using the stringClone method, which might be the cause of the continuous memory growth.

I ran the same test mentioned in the issue. With the fixed version on Go 1.24, without disabling THP, the memory usage has significantly improved compared to 1.70.3, dropping from 4.2GB to around 1.2GB.

#### 1.70.3
profile: 
[rclone.heap.gz](https://github.com/user-attachments/files/21219754/rclone.heap.gz)

rclone heap

#### fixed version
profile: 


rclone-fix heap

VVoidV:
According to the call graph above, a considerable amount of memory is still being allocated during request parsing. I'll take a closer look tomorrow.

ncw:
Nice, well spotted. This has been a problem in the past.

You should probably use stringClone everywhere rather than strings.Clone for consistency, but I think it may be in our lowest go version now so that is a job for another pr/commit if so.

VVoidV:
> Nice, well spotted. This has been a problem in the past.
> 
> You should probably use stringClone everywhere rather than strings.Clone for consistency, but I think it may be in our lowest go version now so that is a job for another pr/commit if so.

1. All have been modified to use stringClone.
2. Added a few additional stringClone calls.

ncw:
Is this working to your satisfaction now @VVoidV ? If so please set the PR to "Ready for review". What you've done so far looks correct to me.

PR #8692, opened 2025-07-17
feat(crypt): implement OpenChunkWriter support

darthShadow:
Add chunked writing support to crypt backend with proper encryption block alignment and nonce management. Enables parallel uploads for encrypted files when underlying backend supports OpenChunkWriter or OpenWriterAt.

NOTE: Most of this code has been written by Claude after significant number of hours spent discussing on the approach and handling of both OpenChunkWriter & OpenWriterAt in the underlying FS.

All the tests seem to be passing.

Implements OpenChunkWriter for the crypt backend

N/A

#### Checklist

### DIFF
--- a/backend/crypt/crypt.go
+++ b/backend/crypt/crypt.go
@@ -2,12 +2,15 @@
 package crypt
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/rclone/rclone/fs"
@@ -296,6 +299,11 @@ func NewFs(ctx context.Context, name, rpath string, m configmap.Mapper) (fs.Fs, 
 
 	// Enable ListP always
 	f.features.ListP = f.ListP
+
+	// Enable OpenChunkWriter if underlying backend supports it or OpenWriterAt
+	if wrappedFs.Features().OpenChunkWriter != nil || wrappedFs.Features().OpenWriterAt != nil {
+		f.features.OpenChunkWriter = f.OpenChunkWriter
+	}
 
 	return f, err
 }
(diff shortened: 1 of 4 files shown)


backend/crypt/crypt.go
```diff
@@ -297,6 +300,11 @@ func NewFs(ctx context.Context, name, rpath string, m configmap.Mapper) (fs.Fs,
 	// Enable ListP always
 	f.features.ListP = f.ListP
 
+	// Enable OpenChunkWriter if underlying backend supports it or OpenWriterAt
```
ncw, line 303: Because of the `Fill` above, the `OpenChunkWriter` feature is likely to be set already, so this should clear it if it can't be implemented I think.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
 	return do(ctx)
 }
 
+// OpenChunkWriter opens a ChunkWriter for chunked writing
+func (f *Fs) OpenChunkWriter(ctx context.Context, remote string, src fs.ObjectInfo, options ...fs.OpenOption) (info fs.ChunkWriterInfo, writer fs.ChunkWriter, err error) {
+	// Check if data encryption is disabled
+	if f.opt.NoDataEncryption {
+		// For unencrypted data, just delegate to underlying backend
+		do := f.Fs.Features().OpenChunkWriter
+		if do == nil {
```
ncw, line 1152: Probably what we should do here is declare support for `OpenWriterAt` if the underlying backend supports it **and** NoDataEncryption is set.

The multipart copy routines have a very similar adaptor already and it would be better to use that one.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+		// Check if underlying backend supports OpenWriterAt
+		openWriterAt := f.Fs.Features().OpenWriterAt
+		if openWriterAt == nil {
+			return info, nil, fs.ErrorNotImplemented
+		}
+		// Use adapter to convert OpenWriterAt to OpenChunkWriter
+		do = f.openChunkWriterFromOpenWriterAt(openWriterAt)
+	}
+
+	// Generate a random nonce for this file
```
ncw, line 1175: Can't we use `cipher.newEncrypter` to do this which creates the file header and nonce?

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+		LeavePartsOnError: underlyingInfo.LeavePartsOnError,
+	}
+
+	return info, cryptWriter, nil
+}
+
+// calculateOptimalChunkSize calculates chunk size aligned to encryption block boundaries
+func (f *Fs) calculateOptimalChunkSize(baseChunkSize int64, fileSize int64) int64 {
+	if baseChunkSize <= 0 {
+		baseChunkSize = 5 * 1024 * 1024 // 5MB default
```
ncw, line 1216: This should probably be configurable with a `--crypt-chunk-size` parameter.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+	}
+
+	// Align to encryption block boundaries to avoid partial blocks
+	blocksPerChunk := (baseChunkSize + blockDataSize - 1) / blockDataSize
+	alignedChunkSize := blocksPerChunk * blockDataSize
+
+	// Ensure reasonable parallelism for smaller files
```
ncw, line 1223: Does this make a difference? I'd be inclined to remove this code unless it demonstrates a significant improvement as none of the other multi-chunk backends do this.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+	chunkSize int64
+	chunks    int
+	writerAt  fs.WriterAtCloser
+	f         fs.Fs
+	closed    bool
+
+	// Track chunk ordering for variable-sized chunks (encryption)
+	chunkOffsets  map[int]int64  // chunk number -> file offset
+	pendingChunks map[int][]byte // buffered chunks waiting to be written
+	mu            sync.Mutex
```
ncw, line 1298: Our convention is to put the mutex above all the things which need protecting, so likely any mutable state in `writerAtChunkWriter`

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+	nextOffset    int64
+	nextChunk     int // next chunk number we expect to write
+}
+
+// WriteChunk writes chunkNumber from reader
+func (w *writerAtChunkWriter) WriteChunk(ctx context.Context, chunkNumber int, reader io.ReadSeeker) (int64, error) {
+	fs.Debugf(w.remote, "writing chunk %v", chunkNumber)
+
+	// Read the chunk data into memory
+	chunkData, err := io.ReadAll(reader)
```
ncw, line 1308: This should be using `lib/pool/RW` for its memory management then it will fit into the rest of rclone's memory usage and will mmap memory with --use-mmap etc.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+	if err != nil {
+		return 0, err
+	}
+
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	// If this is the next chunk we're waiting for, write it immediately
```
ncw, line 1316: Why does this re-order the chunks? The whole point of this interface is that it can be called concurrently.

Provided you are careful with the nonce I don't see why you shouldn't just write the chunks as they come in.

backend/crypt/crypt.go
```diff
@@ -1135,6 +1143,428 @@ func (f *Fs) Shutdown(ctx context.Context) error {
...
+func getKeys(m map[int][]byte) []int {
+	keys := make([]int, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
+// Abort the chunk writing
+func (w *writerAtChunkWriter) Abort(ctx context.Context) error {
```
ncw, line 1416: This should call the underlying chunk writers abort also

ncw (commented):
I gave this a quick once over - see comments inline :-)

I'm not sure I understand why you are buffering the blocks and writing them in-order - that will kill the performance gain I would have thought.

Thinking about this further, the header is a bit of a problem isn't it.

The OpenChunkWriter expects all chunks to be the same size, except for the last one.

Not sure how we can square that with the initial 32 byte header then the 64k chunks.

So I can see this working fine for an underlying backend which implements OpenWriterAt but not one which implements OpenChunkWriter as there is no size of chunks which lets you write the 32 byte header.

Maybe that is what the buffering of chunks is about.

Issue #8114, opened 2024-10-03
Support for FileJump?


closewall:
Will rclone add support for FileJump? It appears to have an open API with supplied documentation: https://drive.filejump.com/api-docs

albertony:
I wouldn't hold my breath.. 🙂 This is the 81. open issue regarding support for a new backend. Unless you plan to implement it yourself, or sponsor it making it more interesting for someone else to invest time on it, it is unlikely to happen any time soon. Sorry.

masrlinu:
I am on my way to add support for FileJump ☺️

masrlinu:
Listing a directory already almost works, but I still have problems. So whoever wants to help:

https://forum.rclone.org/t/help-with-adding-new-backend-filejump-com/48127

But by the way, ftp (which is supported by rclone) is on the roadmap:
https://drive.filejump.com/pages/roadmap
I asked the filejump support. They answered:
"It's under testing phase shall be released soon"

masrlinu:
We are so close to supporting the new backend FileJump 😊 But I still need some help!!

Listing the folder works now, for example with
```
./rclone lsjson filejump:
```
I implemented all methods of the backend, but there are still some errors. It would be good, if somebody with experience could take a look on them and maybe correct my [code](https://github.com/masrlinu/rclone).

Here is this [issue in the rclone-forum](https://forum.rclone.org/t/help-we-are-so-close-to-supporting-the-new-backend-filejump-but-i-need-some-help/48244).

By the way, if you want to check the Api, you can just go to [filejump.com](https://filejump.com) and check the network connections in the developer tools, while you do something. The requests are very easy to understand.

masrlinu:
Syncing via
```
./rclone sync /tmp/tosync/ filejump:/tosync
```
works now, but the files get copied twice, I still have to find out, why.

ncw:
@masrlinu are you running the integration tests? If you can make the integration tests pass then the backend will work perfectly. The way I develop a new backend is just to plod through the integration tests making them all pass.

Make [these ones work first](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#unit-tests) and then [the full tests](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#integration-tests)

Also take a look at the other things in the writing a new backend docs if you haven't already.

GhsVilela:
> Syncing via
> 
> ```
> ./rclone sync /tmp/tosync/ filejump:/tosync
> ```
> 
[...]

PR #8693, opened 2025-07-17
Implement backend for FileJump (#8114)

masrlinu:

This PR implements a backend for the storage provider FileJump.com.

Yes, this backend was requested in issue #8114.

#### Checklist


ncw:
How are the integration tests going? Does `go test -v` in the backend directory work? How about `test_all`?

See here for more info: https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#writing-a-new-backend

masrlinu:
Yes, everything works. I just tested it again. But I had to start it with a bigger timeout. I tested it with
test_all -backends filejump -timeout 10h

It finished in 1h48min.

Here is the output of the test:
[2025-07-21-140856.zip](https://github.com/user-attachments/files/21352390/2025-07-21-140856.zip)

ncw:
> Yes, everything works. I just tested it again. But I had to start it with a bigger timeout. I tested it with test_all -backends filejump -timeout 10h
> 
> It finished in 1h48min.

That is a bit slow, but not the slowest backend!

> Here is the output of the test: [2025-07-21-140856.zip](https://github.com/user-attachments/files/21352390/2025-07-21-140856.zip)

That looks really good. The bisync tests took the longest.

We can add a time modifier in the `config.yaml` for the integration tests

https://github.com/rclone/rclone/blob/ff9cbab5fa36fef07f9237dcef073e16b0c72de8/fstest/test_all/config.yaml#L201

This multiplies the normal timeout (which is 1h) by that factor. Looking at the timings none of the tests took longer than 2h so that should be fine.

This backend was made using the public REST API at https://eu.filejump.com/api-docs - is that right?

I'll give the code a review now.

backend/filejump/api/types.go
```diff
@@ -0,0 +1,119 @@
...
+		return ""
+	}
+	return strconv.Itoa(i.ID)
+}
+
+// ModTime returns the modification time of the item
+func (i *Item) ModTime() (t time.Time) {
+	// Parse UpdatedAt first
+	if i.UpdatedAt != "" {
+		// Try multiple time formats that FileJump might use
```
ncw: Do we need multiple formats?

If so you should probably factor this parsing out as I can see it repeated below

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+
+func init() {
+	fs.Register(&fs.RegInfo{
+		Name:        "filejump",
+		Description: "FileJump",
+		NewFs:       NewFs,
+		Options: []fs.Option{{
+			Name:     "access_token",
+			Help:     "You should create an API access token here: https://drive.filejump.com/account-settings",
+			Required: true,
```
ncw, line 55: This needs `Sensitive: true`

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+}
+
+// Object describes a filejump object
+type Object struct {
+	fs              *Fs
+	remote          string
+	hasMetaData     bool
+	size            int64
+	modTime         time.Time
+	preserveModTime bool // Flag to indicate if modTime should be preserved
```
ncw: What is this for?

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+values.Set("param2", "wert2")
+values.Set("EntryIds", fmt.Sprintf("[%s]", dir))
+values.Set("DeleteForever", strconv.FormatBool(true)))
+
+result, err := CallJSON[Response, struct{}](f, ctx, "GET", "/file-entries/delete", &values, nil)
+if err != nil {
+    return err
+}
+*/
+func CallJSON[T any, B any](f *Fs, ctx context.Context, method string, path string, params *url.Values, body *B) (*T, error) {
```
ncw: Interesting use of generics. I don't think they are strictly needed, but I see it allows you return the `*T` from the function (rather than pass it in) which looks nice.

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+	// }
+	// fs.Logf(nil, logMsg)
+
+	var result T
+	opts := rest.Opts{
+		Method: method,
+		Path:   path,
+		ExtraHeaders: map[string]string{
+			"Accept":          "application/json",
+			"Accept-Encoding": "gzip, deflate, br, zstd",
```
ncw: Can Go deal with those other formats? I only ever use `gzip`.

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+values.Set("EntryIds", fmt.Sprintf("[%s]", dir))
+values.Set("DeleteForever", strconv.FormatBool(true)))
+
+result, err := CallJSON[Response, struct{}](f, ctx, "GET", "/file-entries/delete", &values, nil)
+if err != nil {
+    return err
+}
+*/
+func CallJSON[T any, B any](f *Fs, ctx context.Context, method string, path string, params *url.Values, body *B) (*T, error) {
+	// Eingabeparameter in einer Zeile
```
ncw: Can you comment in English please :-)

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+// Size returns the size of an object in bytes
+func (o *Object) Size() int64 {
+	// fs.Logf(nil, "Size: Hole Größe für '%s'", o.remote)
+	err := o.readMetaData(context.TODO())
+	if err != nil {
+		// fs.Logf(o, "Failed to read metadata: %v", err)
+		return -1
+	}
+	return o.size
+}
```
ncw, line 2235: Please put the Fs **and** Object interface assertions at the end.

Eg

https://github.com/rclone/rclone/blob/ff9cbab5fa36fef07f9237dcef073e16b0c72de8/backend/s3/s3.go#L7060-L7076

backend/filejump/filejump_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/rclone/rclone/backend/filejump"
+	"github.com/rclone/rclone/fstest/fstests"
+)
+
+// TestIntegration runs integration tests against the remote
+func TestIntegration(t *testing.T) {
+	fstests.Run(t, &fstests.Opt{
+		RemoteName:               "filejump:",
+		NilObject:                (*filejump.Object)(nil),
+		UnimplementableFsMethods: []string{"OpenWriterAt", "OpenChunkWriter"},
```
ncw: You can remove this line - it is only needed for Wrapping fs (eg crypt)

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+		// If delete fails, try to remove the copied object to avoid duplicates
+		_ = dstObj.Remove(ctx)
+		return nil, err
+	}
+	
+	return dstObj, nil
+}
+
+
+// Check the interfaces are satisfied
```
ncw: Please put this at the end as that is where we look for it!

docs/content/filejump.md
```diff
@@ -0,0 +1,215 @@
...
+
+When you do this, rclone won't be able to see any files in the root folder or any folders above the one you specified.
+
+### Invalid refresh token
+
+If you get an error about an invalid refresh token, this is usually because you have revoked the API access token in your FileJump account settings. You will need to create a new token and update your rclone config.
+
+### Modification times
+
+FileJump doesn't support setting custom modification times via the API. The modification time will be set to the upload time by the server. This means that `--update` and similar flags that rely on modification times may not work as expected.
```
ncw: Actually the `--update` flag is precisely for this situation.

Remove

> This means that `--update` and similar flags that rely on modification times may not work as expected.

and add

> You may want to use the [--update](/docs/#u-update) flag for a more accurate sync.

docs/content/filejump.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Duplicates
+
+FileJump allows duplicate file names in the same directory. When rclone uploads a file, it will first remove any existing files with the same name to avoid duplicates.
+
+### Restricted filenames
+
+FileJump has a minimum filename length requirement of 3 characters. Rclone will automatically pad shorter filenames with spaces to meet this requirement.
```
ncw: That is the first time I've seen that restriction in a cloud storage system!

I think this needs to be a little bit clearer

> Rclone will automatically pad shorter filenames with spaces on upload to meet this requirement. It will also remove the spaces on download. This means if you happen to have a 1 or 2 character file name ending in a space then syncing may go wrong.

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+
+// type Object interface:
+
+// SetModTime sets the metadata on the object to set the modification date
+func (o *Object) SetModTime(ctx context.Context, t time.Time) error {
+	// fs.Logf(nil, "SetModTime wurde aufgerufen")
+	// FileJump doesn't support setting custom modification times via API
+	// The modification time is set by the server when the file is uploaded
+	// Return this error to indicate that setting mod time requires re-upload
+	return fs.ErrorCantSetModTimeWithoutDelete
```
ncw: You should just return `fs.ErrorCantSetModTime` here. The `fs.ErrorCantSetModTimeWithoutDelete` is for backends which can only set modification times when they upload a file.

docs/content/filejump.md
```diff
@@ -0,0 +1,215 @@
...
+### Filename restrictions
+
+- Minimum filename length: 3 characters (automatically padded by rclone)
+- Standard filename encoding restrictions apply
+
+### Modification times
+
+- FileJump doesn't support setting custom modification times
+- Upload time is used as the modification time
+- `--update` flag behavior may be affected
```
ncw: Using the `--update` flag can improve sync accuracy.

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1680 @@
...
+	}
+
+	type ResultDelete struct {
+		Status string `json:"status,omitempty"`
+	}
+
+	iDir, _ := strconv.Atoi(rootID)
+	request := RequestDelete{
+		EntryIds:      []int{iDir},
+		DeleteForever: true,
```
ncw: Generally backends default to storing deletes in the trash and use a `hard_delete` flag to override.

Eg

      --jottacloud-hard-delete   Delete files permanently rather than putting them into the trash
      --mega-hard-delete         Delete files permanently rather than putting them into the trash

We started doing it that way round after one user deleted all of their files by accident! So safety first, but overridable by the user.

fstest/test_all/config.yaml
```diff
@@ -162,6 +162,8 @@ backends:
  - backend:  "gofile"
    remote:   "TestGoFile:"
    fastlist: true
+ - backend:  "filejump"
+   remote:   "TestFileJump:"
```
ncw: Might want extra time here as discussed above!

ncw (changes requested):
This is looking really great. The integration tests all passing is fantastic.

I've put some notes inline - nothing serious.

I think you've got some lint things to fix too.

Thank you

masrlinu:
> This backend was made using the public REST API at https://eu.filejump.com/api-docs - is that right?

Yes, that is right. But Google Chrome even shows the typescript of the original Code of the website and it is easy to see in the network tab in the browser after every request, what exactly happened.

So there are two different ways how to upload a file. One is by requesting an upload url and upload it there (directly to the underlying wasabi backend). The website does it this way. And the other way to upload a file is via the public REST API. It looks like a wrapper, which uploads the file later to the wasabi backend.

I did it like the website. With big files, it maybe should be faster, because the files don't use the wrapper. But maybe for short files, I should use the REST API.

As I am at the moment on holidays and can just program on my phone, I will look into your change review later.

But actually this artificial intelligence agent:

https://www.atlassian.com/blog/announcements/rovo-dev-command-line-interface

helped me A LOT with programming. I told it to make my code pass the tests and the same day, my code passed ALL the tests!

Sometimes, I had to guide it, but it is unbelievable, you should test it 😄 It is completely free until the end of this month.

And if you like, while I am on holidays, you can update my code, however you like. But don't worry, just if you have time ☺️

ncw:
@masrlinu thanks for the update. I'll mark this for v1.71 which is due 23rd August. So enjoy your holiday and don't feel you need to look at this until you get back :-)

masrlinu:
Hi @ncw! I've addressed all the suggestions, fixed the linter issues, updated the docs, and fixed the filejump-specific workspace handling. The latest changes also include support for move, dir-move, cleanup, about, and publiclink. Let me know what you think!
PS: all tests with test_all passed again

backend/filejump/api/types.go
masrlinu: Now we just use one format

backend/filejump/filejump.go
masrlinu: I removed it :-)

backend/filejump/filejump.go
masrlinu: I changed it to just use gzip

masrlinu:
I've implemented some fixes. The FileJump cleanup logic wasn't working properly when the trash contained too many files, so I've updated it to delete trash files in batches of 100. I also addressed a few other issues. All integration tests are passing:
[rclone-integration-tests.zip](https://github.com/user-attachments/files/21683014/rclone-integration-tests.zip)

docs/content/filejump.md
```diff
@@ -0,0 +1,258 @@
+---
+title: "FileJump"
+description: "Rclone docs for FileJump"
+versionIntroduced: "v1.69"
```
ncw: This should be v1.71

docs/content/filejump.md
```diff
@@ -0,0 +1,258 @@
...
+---
```
ncw, line 6: Suggest you put

```
status: Beta
```

In the headers then you'll get a Beta badge like this one

https://rclone.org/bisync/

backend/filejump/filejump_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"testing"
+
+	"github.com/rclone/rclone/backend/filejump"
+	"github.com/rclone/rclone/fstest/fstests"
+)
+
+// TestIntegration runs integration tests against the remote
+func TestIntegration(t *testing.T) {
+	fstests.Run(t, &fstests.Opt{
+		RemoteName: "filejump:",
```
ncw: Can you call this `TestFileJump:` so it fits in with the naming scheme of the other test remotes.

docs/content/filejump.md
```diff
@@ -0,0 +1,258 @@
...
+Type of storage to configure.
+Choose a number from below, or type in your own value
+[snip]
+XX / FileJump
+   \ "filejump"
+[snip]
+Storage> filejump
+** See help for filejump backend at: https://rclone.org/filejump/ **
+
+You should create an API access token here: https://drive.filejump.com/account-settings
```
ncw: This URL doesn't seem to work. It gives me "These credentials do not match our records"

https://eu.filejump.com/account-settings

Did work though.

However the token did not work giving these errors

```
2025/08/15 18:09:21 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2025/08/15 18:09:21 DEBUG : HTTP REQUEST (req 0xc00044a140)
2025/08/15 18:09:21 DEBUG : GET /api/v1/drive/file-entries?perPage=1000 HTTP/1.1
Host: drive.filejump.com
User-Agent: rclone/v1.71.0-beta.8915.df83968c9.pr-8693-filejump
Content-Length: 4
Accept: application/json
Accept-Encoding: gzip
Authorization: XXXX
Content-Type: application/json

null
2025/08/15 18:09:21 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2025/08/15 18:09:22 DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
2025/08/15 18:09:22 DEBUG : HTTP RESPONSE (req 0xc00044a140)
2025/08/15 18:09:22 DEBUG : HTTP/1.1 401 Unauthorized
Transfer-Encoding: chunked
Access-Control-Allow-Origin: *
Cache-Control: no-cache, private
Connection: keep-alive
Content-Type: application/json
Date: Fri, 15 Aug 2025 17:09:22 GMT
Server: nginx
Vary: Authorization

1e
{"message":"Unauthenticated."}
0
```

I made the token work (for listing anyway) with this

```diff
diff --git a/backend/filejump/filejump.go b/backend/filejump/filejump.go
index 564873e04..347f3630a 100644
--- a/backend/filejump/filejump.go
+++ b/backend/filejump/filejump.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-       apiBaseURL          = "https://drive.filejump.com/api/v1"
+       apiBaseURL          = "https://eu.filejump.com/api/v1"
        defaultUploadCutoff = 50 * 1024 * 1024
        smallFileCutoff     = 15 * 1024 * 1024 // 15 MiB
 )
```

but it didn't appear to work for uploads

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1969 @@
...
+	// Input parameters in one line
+	logMsg := fmt.Sprintf("CallJSON: %s %s", method, path)
+	if params != nil && len(*params) > 0 {
+		logMsg += fmt.Sprintf(" params=%s", params.Encode())
+	}
+	if body != nil {
+		bodyJSON, _ := json.Marshal(body)
+		logMsg += fmt.Sprintf(" body=%s", string(bodyJSON))
+	}
+	fs.Debugf(f, "%v", logMsg)
```
ncw: Can you remove the debugging out of this function please.

You can see the transactions with `-vv --dump headers/bodies/requests/responses`

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1969 @@
...
+// deserve to be retried.  It returns the err as a convenience
+func shouldRetry(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	if fserrors.ContextError(ctx, &err) {
+		return false, err
+	}
+	authRetry := false
+
+	if resp != nil && resp.StatusCode == 401 && strings.Contains(resp.Header.Get("Www-Authenticate"), "expired_token") {
+		authRetry = true
+		fs.Debugf(nil, "Should retry: %v", err)
```
ncw: No need for this debug - the pacer will debug

backend/filejump/filejump.go
```diff
@@ -0,0 +1,1969 @@
...
+		// If no MIME type was found, return default
+		mimeType = "application/octet-stream"
+	}
+
+	return extension, mimeType
+}
+
+func (o *Object) upload(ctx context.Context, in io.Reader, leaf, directoryID string, size int64, modTime time.Time, options ...fs.OpenOption) (err error) {
+	// Upload via S3 presigned URL
+	fs.Debugf(o, "Uploading file via S3 presigned URL")
```
ncw: I'd probably remove this one too

ncw (commented):
I thought I'd give this a go so I went through the docs configuring it like a normal user.

I found a few things - see inline.

With the change for eu server I still couldn't get it to work though. It returned this on uploads

```
(error HTTP error 500 (500 Internal Server Error) returned body: "{\n    \"message\": \"There was an issue. Please try again later.\"\n}")
```

I was doing this test

```
rclone test makefiles --files 100 --max-file-size 2M /tmp/100files
rclone copy -Pvv /tmp/100files TestFileJump:100files
```

docs/content/filejump.md
masrlinu: Oh, I found out the reason. It is, because filejump has three different servers for different accounts. My account was an old one, so I also have to ask for the Server in the options.

masrlinu:
So much work!! The server eu.filejump.com uses a different protocol for file storage. After a lot of effort, I finally got it working on both drive.filejump.com and eu.filejump.com. However, on eu.filejump.com two tests still fail:
```
            --- FAIL: TestIntegration/FsMkdir/FsPutFiles/ObjectOpenSeek (0.42s)
            --- FAIL: TestIntegration/FsMkdir/FsPutFiles/ObjectOpenRange (1.02s)
```
It seems that range reads/seek might not be supported there. For now, I’d like to mark this as unsupported in the current version and maybe add proper support in a future release. Is there a clean way to declare this as unsupported and have those tests skipped?

ncw:
> So much work!! The server eu.filejump.com uses a different protocol for file storage

:-(

> After a lot of effort, I finally got it working on both drive.filejump.com and eu.filejump.com. 

Well done!

> However, on eu.filejump.com two tests still fail:
> 
> ```
>             --- FAIL: TestIntegration/FsMkdir/FsPutFiles/ObjectOpenSeek (0.42s)
>             --- FAIL: TestIntegration/FsMkdir/FsPutFiles/ObjectOpenRange (1.02s)
> ```
> 
> It seems that range reads/seek might not be supported there. For now, I’d like to mark this as unsupported in the current version and maybe add proper support in a future release. Is there a clean way to declare this as unsupported and have those tests skipped?

There are an awful lot of things that rely on Range requests - it isn't an optional feature. So many things will be broken if it doesn't work properly (eg `rclone mount`, multithread copy, retry network failures, etc, etc).

I have in the past bodged it in by opening the object and discarding bytes but I don't recommend that as it is a huge waste of bandwidth. All the current backend support either Range requests or seeking to a point before opening.

masrlinu:
I checked it — the backend at eu.filejump.com doesn’t support range requests. Pretty bad. I worked around it by downloading the full file and simulating the ranges. Currently running test_all.

backend/filejump/filejump.go
```diff
@@ -49,8 +51,8 @@ func init() {
 			Required:  true,
 			Sensitive: true,
 		}, {
-			Name:     "api_domain",
-			Help:     "Enter the domain name of your FileJump server, as shown in your browser’s address bar when you’re logged in (e.g., drive.filejump.com, app.filejump.com, eu.filejump.com).",
+			Name:     "server",
+			Help:     "Enter the number for your FileJump server (check your browser's address bar when logged in): 1 = drive.filejump.com, 2 = app.filejump.com, 3 = eu.filejump.com",
```
ncw: This should be using config examples like this one

https://github.com/rclone/rclone/blob/f2eb5f35f63dc980356917241402b6522d4936d7/backend/s3/s3.go#L403-L412

If you do that then rclone config and the GUIs will make menus of the choices.

Make the value be `drive.filejump.com` etc then you don't need to look up the numbers in the code.

This should probably be called `api_domain`

masrlinu:
I updated the backend as you suggested and also updated the documentation. All tests run via test_all passed for drive.filejump.com and eu.filejump.com. I couldn’t test it on the server app.filejump.com, since I don’t have an account there and can’t create one — account registration is only possible on the newest server, eu.filejump.com.

backend/filejump/filejump.go
```diff
@@ -292,7 +292,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	}
 
 	// Get the API domain and handle server selection errors
-	apiDomain, err := opt.GetAPIDomain()
+	apiDomain, err = opt.GetAPIDomain()
```
ncw: Please don't make this a global - it will stop the user running more than one FileJump backend at once.

It should live in the `Fs` structure.

backend/filejump/filejump.go
```diff
@@ -1486,6 +1486,66 @@ func (o *Object) SetModTime(ctx context.Context, t time.Time) error {
 	return fs.ErrorCantSetModTime
 }
 
+// RangeReader implementiert io.ReadCloser und simuliert Range-Requests
```
ncw: Comments in English please

backend/filejump/filejump.go
```diff
@@ -1533,12 +1626,19 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
 			_ = resp.Body.Close()
 			return nil, fmt.Errorf("download failed with status: %s", resp.Status)
 		}
 
-		return resp.Body, nil
+		// Wenn keine Range-Option vorhanden ist, gib den kompletten Body zurück
+		if rangeOption == nil {
+			return resp.Body, nil
+		}
+
```
ncw, line 1552: This is way more complicated than it needs to be.

First copy offset bytes to discard with io.CopyN


Next wrap the handle in `readers.NewLimitedReadCloser` (see `lib/readers`)

Eg

https://github.com/rclone/rclone/blob/ecd52aa809775e68c6745bcbf8987d93fdafb7d8/backend/imagekit/imagekit.go#L645-L653

In fact that shows how to apply a Range header and correct if the Range header was ignored - you should probably do that.

backend/filejump/filejump.go
```diff
@@ -51,9 +49,16 @@ func init() {
 			Required:  true,
 			Sensitive: true,
 		}, {
-			Name:     "server",
+			Name:     "api_domain",
 			Help:     "Enter the number for your FileJump server (check your browser's address bar when logged in): 1 = drive.filejump.com, 2 = app.filejump.com, 3 = eu.filejump.com",
```
ncw: The help should just say `Your FileJump server (check your URL bar)` or something like that - leave the mechanics of the choice up to the config user interface which might be a drop down menu.

docs/content/filejump.md
```diff
@@ -62,6 +70,17 @@ d) Delete this remote
 y/e/d> y
 ```
 
+### Server selection (api_domain)
+
+During configuration you will be asked to select your FileJump server (api_domain). Choose the number that matches the domain you see in your browser when logged in:
```
ncw: Put this in the help for the `api_domain` and it will appear in the docs automatically, **and** be in the config interface

backend/filejump/filejump.go
masrlinu: ok, thank you! It is already solved, because now I directly get it from the Options.

backend/filejump/filejump.go
masrlinu: Thank you, it was already corrected in the new version :-)

backend/filejump/filejump.go
masrlinu: I updated my logic like you said :-)

masrlinu:
I updated everything like you said!
The tests
TestIntegration/FsMkdir/FsPutFiles/ObjectOpenSeek
TestIntegration/FsMkdir/FsPutFiles/ObjectOpenRange
pass for both backends :-)

masrlinu:
Please read https://github.com/rclone/rclone/issues/8114#issuecomment-3481523530, before you merge!

DWomack:
I am trying build your version that supports drive.filejump.com.  
To make this as easy as possible, I am trying to use docker compose.  Also trying to do this on a raspberrypi that has access to symetric fiber internet.  Here is the compose file.

services:
  rclone-builder:
    image: golang:1.24-alpine
    container_name: rclone-builder
    working_dir: /build
    volumes:
      - ./rclone-bin:/output
    environment:
      - CGO_ENABLED=0
      - GOOS=linux
      - GOARCH=arm64
    command: sh -c "apk add --no-cache git make bash && git clone https://github.com/rclone/rclone.git . && git fetch origin pull/8693/head:pr-8693 && git checkout pr-8693 && make && cp rclone /output/ && /output/rclone version && /output/rclone help backends | grep -i filejump"

So on a rpi 4, this runs and creates the rclone file, but when I execute the rclone binary with rclone config, I don't see filejump.

I'm trying to do this in docker compose so the developement environment is portable.

The new www.filejump.com website will finally pull files from drive.filejump.com, sort of.  It hasn't been very reliable. I have lots of dupes now.  That's the reason I want the rclone solution to work with drive.filejump.com.

If you could send me a pre built amd64 rclone that works with the variants of filejump I would appreciate it.  I have dropbox, and filerule.com.  I think filerule.com is by the filejump people.

DWomack:
I finally got it to build and filejump shows up as a backend.  Yeah!!!  Now to test.

This was on an amd64 ubuntu box.

DWomack:
Just want to say thanks for your work.
I have it installed on two Pi and one amd64 ubuntu. 
I have 360 gig to move from drive.filejump.com to the new filejump.com via webdav and to filerule.com which is also webdav
I learned to use docker compose containers to build rclone.  Great lesson.
Thank you for the time you spent with this.

masrlinu:
You're welcome 😄 If you copy the data to two different clouds, instead of reading twice from the source, you can create a [union](https://rclone.org/union/), then you will have a backend, which acts like one but is actually two backends. You should configure it correctly, then if you copy files to there, they actually will be saved on both backends and are just read one time. And also for later, you can mount this union on your computer and the files will always be on both backends.

DWomack:
My collocated pi that has symmetric fiber is chugging along syncing my old drive.filejump.com to the new FileJump.com WebDAV. Data moving from old to new. 

I have tried to do the same with filerule.com via WebDAV, but it is a bit dodgy. Their WebDAV interface has some issues.  

I have been using Claude.ai to debug. It says run this test.  I do and then it interprets the results. Now run this test...

I have found that davfs2 in Linux should work against filerule.com but doesn't. 

I really appreciate your work on the backend for rclone.

PR #8699, opened 2025-07-20
fs: allow global variables to be overriden or set on backend creation

### LINKED ISSUE #8563: Support `--ca-cert` functionality as config option

```console
paul@monkeywrench:~$ rclone version
rclone v1.60.1-DEV
- os/version: debian 12.11 (64 bit)
- os/kernel: 6.1.0-35-amd64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.19.8
- go/linking: dynamic
- go/tags: none
paul@monkeywrench:~$ 
```

I am trying to connect to a WebDAV endpoint that uses a (server) TLS certificate issue by a CA that is not in my computer's default trust store.

This CA is not part of CA/B, so it is not automatically part of my computers default trust store.

Also I don't want to trust the CA outside of this WebDAV endpoint, so I don't want to add this CA to my default trust store.

rclone already supports this, via the `--ca-cert` command-line argument.  However, I don't want to have to type this argument whenever I'm using rclone.  Instead, I would like to have this option as part of my rclone configuration for this remote.

The configuration file could accept an extra option, `ca_cert` perhaps.

#### How to use GitHub

ncw:
This allows backend config to contain

- `override.var` - set var during remote creation only
- `global.var` - set var in the global config permanently

Fixes #8563

(this also adds the --http-proxy flag which was the initial impetus for this)

### DIFF
--- a/docs/content/docs.md
+++ b/docs/content/docs.md
@@ -437,6 +437,103 @@ Do not use single character names on Windows as it creates ambiguity with Window
 drives' names, e.g.: remote called `C` is indistinguishable from `C` drive. Rclone
 will always assume that single letter name refers to a drive.
 
+## Adding global configuration to a remote
+
+It is possible to add global configuration to the remote configuration which
+will be applied just before the remote is created.
+
+This can be done in two ways. The first is to use `override.var = value` in the
+config file or the connection string for a temporary change, and the second is
+to use `global.var = value` in the config file or connection string for a
+permanent change.
+
+This is explained fully below.
+
+### override.var
+
+This is used to override a global variable **just** for the duration of the
+remote creation. It won't affect other remotes even if they are created at the
+same time.
+
+This is very useful for overriding networking config needed for just for that
+remote. For example, say you have a remote which needs `--no-check-certificate`
+as it is running on test infrastructure without a proper certificate. You could
+supply the `--no-check-certificate` flag to rclone, but this will affect **all**
+the remotes. To make it just affect this remote you use an override. You could
+put this in the config file:
+
+```ini
+[remote]
+type = XXX
+...
+override.no_check_certificate = true
+```
+
+or use it in the connection string `remote,override.no_check_certificate=true:`
+(or just `remote,override.no_check_certificate:`).
+
+Note how the global flag name loses its initial `--` and gets `-` replaced with
+`_` and gets an `override.` prefix.
+
+Not all global variables make sense to be overridden like this as the config is
+only applied during the remote creation. Here is a non exhaustive list of ones
+which might be useful:
+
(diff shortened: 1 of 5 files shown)


ncw:
@nielash and @albertony I would appreciate your review on this.

Do the docs make sense? Will users understand what the feature does? Is it a good way of implementing it?

I don't want this to interfere with future `--profile` developments for example so I really want to get this right first time!

albertony:
I think this can work nicely. As a feature for experienced users at least, it seems potentially very useful. I think the docs explains it well. However, there is definitely a possibility for users to confuse themselves by combining regular command line options, with perhaps overlapping set of them set in one of the remotes and maybe some additional overlapping set even in the other remote. But it seems you have added debug logging to make it possible to track such cases... So 👍 from me after a first read through.

nielash:
I took a brief look at it and I think it is sensible. I like its simplicity. I think it would be compatible with the [`--profile`](https://github.com/rclone/rclone/issues/2697#issuecomment-1888668988) idea and they serve different enough purposes that it is useful to have both. Also, I'd hate to hold up a feature that is ready to ship now with theoretical concerns about a future feature which I might not have time to finish for a while! 

My one thought (and I haven't had a chance to do a deep dive on it) is that I wonder if this would mess anything up when we have to call `NewFs` (or `cache.Get`) internally during a command, although I think that's probably pretty rare? But I think this is making the assumption that we only call `NewFs` before running a command, and I'm not sure if that's true (but maybe it is?) Would that run the risk of changing the global config in the middle of a sync? 

My quick answer though is that I think it's a good and simple proposal, and it is worth taking the easy win now instead of holding out for perfection later!

ncw:
Thanks for the reviews @albertony and @nielash 

I'll merge this now.

ncw:
@nielash Sorry for forgot to answer this

> My one thought (and I haven't had a chance to do a deep dive on it) is that I wonder if this would mess anything up when we have to call `NewFs` (or `cache.Get`) internally during a command, although I think that's probably pretty rare? But I think this is making the assumption that we only call `NewFs` before running a command, and I'm not sure if that's true (but maybe it is?) Would that run the risk of changing the global config in the middle of a sync?

I think for the `rc` and the commands we always make the backends first - all the internal calls take a `fs.Fs` rather than a config string. I think the one exception to this is `--backup-dir` which is made on the fly when needed. I think nearly all uses of `--backup-dir` use the same remote as the destination though which will have been instantiated.

Possibly a corner case worth documenting, but I think it is very unlikely to be a problem in practice.
 
> My quick answer though is that I think it's a good and simple proposal, and it is worth taking the easy win now instead of holding out for perfection later!

:-)

PR #8708, opened 2025-07-23
bisync: deglobalize to fix concurrent runs via rc - fixes #8675

### LINKED ISSUE #8675: Runtime error when run two bisync jobs with _async

https://forum.rclone.org/t/runtime-error-when-run-two-bisync-jobs-with-async/51906

When I run the first rcone rc bisync job with the async=true flag, it runs normally, but if I run another bisync task at the same time, the application throws an error "panic: runtime error: invalid memory address or nil pointer dereference" and stops.
If I use the resync=true flag, the error does not occur.

rclone v1.70.3
os/version: ubuntu 24.04 (64 bit)
os/kernel: 6.8.0-63-generic (x86_64)
os/type: linux
os/arch: amd64
go/version: go1.24.4
go/linking: static
go/tags: none

ubuntu 64bit, windows 10 64bit

alfresco webdav (other)

./rclone  rcd --no-check-certificate --rc-user=admin --rc-pass=admin -vvv --log-file /tmp/rclone.log
./rclone rc sync/bisync --rc-user=admin --rc-pass=admin path2="TR:/07_Acts_with_contractors" path1="/home/ilya/csync_test/07_Acts_with_contractors" createEmptySrcDirs=true  _async=true
./rclone rc sync/bisync --rc-user=admin --rc-pass=admin path2="TR:/08_Minutes" path1="/home/ilya/csync_test/08_Minutes" createEmptySrcDirs=true  _async=true

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xae78bc]

goroutine 1730 [running]:
github.com/rclone/rclone/fs/accounting.StatsGroupFromContext(...)
        github.com/rclone/rclone/fs/accounting/stats_groups.go:264
github.com/rclone/rclone/fs/accounting.Stats({0x0, 0x0})
        github.com/rclone/rclone/fs/accounting/stats_groups.go:273 +0x1c
github.com/rclone/rclone/cmd/bisync.(*bisyncRun).ForObject(0xc000d76000, {0x2c3a8e0, 0xc000d35560}, 0x1)
        github.com/rclone/rclone/cmd/bisync/march.go:132 +0x6b
github.com/rclone/rclone/cmd/bisync.(*bisyncRun).parse(0xc000d76000?, {0x75eb253c92e8?, 0xc000d35560?}, 0x1?)
        github.com/rclone/rclone/cmd/bisync/march.go:111 +0x107
github.com/rclone/rclone/cmd/bisync.(*bisyncRun).Match(0xc000d76000, {0x25eeec0?, 0xc000cd4380?}, {0x75eb25505800, 0xc00091e8a0}, {0x75eb253c92e8, 0xc000d35560})
        github.com/rclone/rclone/cmd/bisync/march.go:89 +0x1e5
github.com/rclone/rclone/fs/march.(*March).processJob(0xc000df8480, {{0xc0004f5920, 0x2a}, {0xc0009d7950, 0x2a}, 0x7ffffffc, 0x7ffffffc, 0x0, 0x0})
        github.com/rclone/rclone/fs/march/march.go:491 +0xd46
github.com/rclone/rclone/fs/march.(*March).Run.func1()
        github.com/rclone/rclone/fs/march/march.go:169 +0x1e5
created by github.com/rclone/rclone/fs/march.(*March).Run in goroutine 1559
        github.com/rclone/rclone/fs/march/march.go:159 +0x172

#### How to use GitHub

nielash:

Before this change, bisync used some global variables, which could cause errors if running multiple concurrent bisync runs through the `rc`. (Running normally from the command line was not affected.)

This change deglobalizes those variables so that multiple bisync runs can be safely run at once, from the same rclone instance.

- #8675
- https://forum.rclone.org/t/runtime-error-when-run-two-bisync-jobs-with-async/51906

#### Checklist

### DIFF
--- a/cmd/bisync/bisync_test.go
+++ b/cmd/bisync/bisync_test.go
@@ -176,6 +176,7 @@ var (
 	// Flag -refresh-times helps with Dropbox tests failing with message
 	// "src and dst identical but can't set mod time without deleting and re-uploading"
 	argRefreshTimes = flag.Bool("refresh-times", false, "Force refreshing the target modtime, useful for Dropbox (default: false)")
+	ignoreLogs      = flag.Bool("ignore-logs", false, "skip comparing log lines but still compare listings")
 )
 
 // bisyncTest keeps all test data in a single place
@@ -261,6 +262,25 @@ func TestBisyncRemoteRemote(t *testing.T) {
 	require.NoError(t, err)
 	defer cleanup()
 	testBisync(t, remote, remote)
+}
+
+// make sure rc can cope with running concurrent jobs
+func TestBisyncConcurrent(t *testing.T) {
+	oldArgTestCase := argTestCase
+	*argTestCase = "basic"
+	*ignoreLogs = true // not useful to compare logs here because both runs will be logging at once
+	t.Cleanup(func() {
+		argTestCase = oldArgTestCase
+		*ignoreLogs = false
+	})
+
+	t.Run("test1", testParallel)
+	t.Run("test2", testParallel)
+}
+
+func testParallel(t *testing.T) {
+	t.Parallel()
+	TestBisyncRemoteRemote(t)
 }
 
 // TestBisync is a test engine for bisync test cases.
@@ -1429,6 +1449,9 @@ func (b *bisyncTest) compareResults() int {
 		resultText := b.mangleResult(b.workDir, file, false)
 
 		if fileType(file) == "log" {
+			if *ignoreLogs {
+				continue
+			}
 			// save mangled logs so difference is easier on eyes
 			goldenFile := filepath.Join(b.logDir, "mangled.golden.log")
 			resultFile := filepath.Join(b.logDir, "mangled.result.log")
--- a/cmd/bisync/checkfn.go
+++ b/cmd/bisync/checkfn.go
@@ -249,7 +249,7 @@ func (b *bisyncRun) resyncTimeSizeEqual(ctxNoLogger context.Context, src fs.Obje
 		// note that arg order is path1, path2, regardless of src/dst
 		path1, path2 := b.resyncWhichIsWhich(src, dst)
 		if sizeDiffers(path1.Size(), path2.Size()) {
-			winningPath := b.resolveLargerSmaller(path1.Size(), path2.Size(), path1.Remote(), path2.Remote(), b.opt.ResyncMode)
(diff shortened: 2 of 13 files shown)


nielash:
(rebased to remove duplicate commits)

cmd/bisync/bisync_test.go
```diff
@@ -264,6 +265,25 @@ func TestBisyncRemoteRemote(t *testing.T) {
 	testBisync(t, remote, remote)
 }
 
+// make sure rc can cope with running concurrent jobs
```
ncw, line 282: This is going to be quite hard work for the integration tester I think! I looked at a random remote and it takes 15 minutes or so.

We might need to limit this test to only running on local

nielash: Agreed, and I can go ahead and make that change, no problem.

I also wonder what you'd think of adding a new `fstest` skip category for tests like this which would be impractical to run nightly but which we also don't want to run _never_. Perhaps these extended tests could instead be run once a month? And by developers when submitting new backends, which I think is the more important use case. 

My thought is that, in practice, users do of course use bisync on non-local remotes, and it is plausible that particular backends could have the same kinds of data race issues that make them non-thread-safe for concurrent `rc` runs. That would be a backend issue, not a bisync issue per se, but the backend developer is unlikely to be aware of it absent tests like this. (For example, my [discovery last week](https://github.com/rclone/rclone/pull/8739/commits/ff8a7d6f8480fe1cab89f647b4430270762e0bcc) that `azurefiles` has been erasing hashes when modtime is set, and none of the other tests caught it.) IMO, it is preferable for the backend developer to fix these kinds of issues at the outset than for us to have to fix them later with more limited knowledge of the backend in question. 😅 

In the meantime, here is a commit to skip on non-local: https://github.com/nielash/rclone/commit/afba15d436da03eac444726485528475ef6a4e93 Up to you if you want to merge this at the same time as the others, or after trying it once and seeing what happens 😄

ncw: I think running this test on local alone should be enough to find >99% of problems. Let's merge this as is, then can you a day later merge the local only patch then we can have one run with the full tests. I'm not sure we run the integration tester with the race detector though so maybe this isn't as useful there as it should be.

We probably should run the integration tester with the race detector though...

As for your skip idea - I don't really like the idea of tests that only get run sometimes. There is enough noise in the integration test results that we probably would notice!

Really we need to scale out the integration tests, keep the results in a database and make them runnable on GitHub. A big project though.

PR #8719, opened 2025-07-31
smb: Improve multithreaded upload performance using multiple connection.

sudiptob2:
In the current design, `OpenWriterAt` provides the interface for random-access writes, and `openChunkWriterFromOpenWriterAt` wraps this interface to enable parallel chunk uploads using multiple goroutines. A global **connection pool** is already in place to manage SMB connections across files. However, only one connection is used per file, which makes multiple goroutines compete for the connection during multithreaded writes.

The proposed uses creates separate connections for each goroutine, which allows **True Parallelism** by giving each goroutine its own SMB connection


ncw:
Fixing OpenWriterAt seems like a better solution than implementing OpenChunkWriter

Did you try the integration tests? And with the race detector?

sudiptob2:
@ncw Thanks for taking a look!
For now, I have tried the following tests. Need some guidance if there is some other way I need to test it locally.

```bash
# 1. passing on CI
make quicktest 

# 2. passing on CI
make racequicktest 

# 3. Passed 
go run ./fstest/test_all -backends smb -race -verbose
```

sudiptob2:
@ncw, just following up, do you have any feedback on this PR? Please let me know if I need to make any improvements.

Issue #7350, opened 2023-10-03
Reduce RAM usage by relying on temp files


maxim:
#### The associated forum post URL from `https://forum.rclone.org`

https://forum.rclone.org/t/1-64-0-and-latest-beta-1-65-0-both-crash-on-rclone-crypt-sync-b2-on-large-files/41987

#### What is your current rclone version (output from `rclone version`)?

```
rclone v1.65.0-beta.7391.55c3c221b.fix-b2-upload-url-lock
- os/version: ubuntu 22.04 (64 bit)
- os/kernel: 5.15.0-84-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.1
- go/linking: static
- go/tags: none
```

#### What problem are you are trying to solve?

Rclone currently stores chunks in RAM. Recently, RAM usage increased for B2 users, when a refactor changed the RAM formula from

`--transfers * --b2-chunk-size`

to

`--transfers * --b2-chunk-size * --b2-upload-concurrency`

A possible way to reduce RAM usage across the board is to move most chunk manipulations into temporary files on disk.

#### How do you think rclone should be changed to solve that?

To summarize the forum post, we could add an alternative implementation of `io.ReadSeeker` interface, that reads disk (file handles) rather than objects in RAM.

In a more abstract framing: any chunk-data operation that writes/reads RAM could instead do so with temporary files on disk. This includes activity such as caching data coming from network, and hashing data to calculate checksums. Every chunk manipulation should be done via streaming to/from temporary files rather than RAM.

This behavior could be exposed to an end user as an option such as `--low-memory` or the opposite `--slow-disk`, depending on which mode ends up being the default.

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
I see 3 possible buffering schemes

1. `RAM`: Buffer in RAM like we have at the moment
2. `Disk`: Buffer on disk as suggested here
3. `None`: Don't buffer, but re-read from the source if necessary.

The `b2` backend only rewinds the stream if there is a retry. Retries are rare, so we could use scheme `None` when uploading to `b2`. (We need a backend feature to identify this, or maybe a parameter in the chunking protocol.)

[...]

Issue #7886, opened 2024-06-02
Rclone sync behaving badly, forcibly clearing RAM cache


zcalusic:
#### What is the problem you are having with rclone?

I started experiencing performance issues after running an rclone sync. I noticed that carefully cached content in RAM is forcibly expunged after the rclone sync completes. This behavior was implemented in the following commit: https://github.com/rclone/rclone/commit/f39655093401c8aa4cbfcd56a16c446655fafe77.

As a result, rclone is now behaving as if it is the only user of the data and assumes that the data won't be needed again, which is a mistaken assumption. Rclone is NOT the owner of the data that is synced and therefore should NOT decide to expunge it from the RAM just because it doesn't need it anymore!

Please remove the fadvise(POSIX_FADV_DONTNEED) calls from rclone. Alternatively, if anyone finds this "feature" useful (which is unlikely), make it optional, with the default setting being not to interfere with memory management.

The Linux kernel memory management subsystem is quite capable of freeing unused memory, especially if data has been read only once. In such cases, the data is in the inactive page cache list and is immediately freeable as soon as it is needed for something else.

Observed under rclone v1.66.0, Linux OS, 64 bit

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
In the case of doing transfers where the total file size is greater than the page cache I can't see an advantage in overwriting the page cache with data which is unlikely to be used again.

The patch you referenced fixed a real pain point for rclone users.

Maybe you could write more about your specific problem - perhaps we should make the fadvise calls optional on a flag?

zcalusic:
Yes, as I wrote in my original report, you can make fadvise() calls optional, and it works for me. The important thing is that such an aggressive method of clearing RAM can be and is turned off by default.

[...]

PR #8720, opened 2025-07-31
Increase copy speed by re-enabling buffering for multithreaded copies from local filesystems

vitorog:
Make multithreaded copies from local to S3 fast again (by re-enabling buffering)!

From v1.64.2 to v1.65+, the performance of multithreaded copies from local to S3 decreased. 
After some investigation and debugging, it seems this is a side effect of this [commit](https://github.com/rclone/rclone/commit/1f9a79ef09fe9f1720040073992f7dc06030cebf?diff=split#diff-57914b2dca5f9b6573400f20eb94657c8fcb57b692e0636af96d7a74ec228016R146) / issue https://github.com/rclone/rclone/issues/7350 - more specifically disabling buffering when copying from local filesystems.

I executed some tests in a **m6a.2xlarge** AWS EC2 instance (Network up to 12.5 Gbps and EBS up to 10 Gbps) and here are the results:

Copy of a 64 GB file from the local filesystem to S3
`time ./rclone copyto --s3-no-check-bucket --ignore-checksum --s3-disable-checksum --progress --s3-upload-cutoff=0 --multi-thread-cutoff=256M --multi-thread-streams 20 --disable=copy --no-check-dest <Local File> <S3 Bucket>`

| Buffering | Time | Avg Speed |
|:---|---|---:|
| `disabled` | 18 min 46s | ~58 MB/s |
| `enabled` | 9m 10s | ~121 MB/s |

I'm not completely sure about the memory consumption implications - but alternatively if it cannot be enabled by default - could we consider making it configurable?

What do you think? Looking forward to some input and feedback!

No

#### Checklist

### DIFF
--- a/fs/operations/multithread.go
+++ b/fs/operations/multithread.go
@@ -141,9 +141,9 @@ func multiThreadCopy(ctx context.Context, f fs.Fs, remote string, src fs.Object,
 		noBuffering = true
 		usingOpenWriterAt = true
 	} else if src.Fs().Features().IsLocal {
-		// If the source fs is local we don't need to buffer
-		fs.Debugf(src, "multi-thread copy: disabling buffering because source is local disk")
-		noBuffering = true
+		// If the source fs is local, buffering can greatly increase performance of multithread uploads from local
+		fs.Debugf(src, "multi-thread copy: enabling buffering because source is local disk")
+		noBuffering = false
 	} else if f.Features().ChunkWriterDoesntSeek {
 		// If the destination Fs promises not to seek its chunks
 		// (except for retries) then we don't need buffering.


ncw:
Hmm, interesting. Fundamentally the disk should read at the same speed into the s3 multipart buffer (how it is at the moment) or into a memory buffer (like it used to be). Given that disk read speeds > network speeds why is this making a difference?

My guess is that it is because the s3 backend reads each block 3 times (once to MD5 it, once to sign it and once to send it). Before it read once off disk and twice out of RAM. Now it is reading 3 times off disk.

The OS should have cached the 2nd and 3rd reads though but...we may well have disabled that with fadvise.

Try this patch and see if it make a difference.

```patch
diff --git a/backend/local/local.go b/backend/local/local.go
index 14effd2a9..adc4568c3 100644
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -1350,7 +1350,7 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 	if err != nil {
 		return
 	}
-	wrappedFd := readers.NewLimitedReadCloser(newFadviseReadCloser(o, fd, offset, limit), limit)
+	wrappedFd := readers.NewLimitedReadCloser(fd, limit)
 	if offset != 0 {
 		// seek the object
 		_, err = fd.Seek(offset, io.SeekStart)
```

Disabling fadvise was discussed in https://github.com/rclone/rclone/issues/7886 - maybe we should.

Second guessing the OS is probably a bad idea since I'm sure the linux kernel developers are better at memory management than me :-)

vitorog:
hi, @ncw , thanks for your reply. 
Your explanation actually makes a lot of sense (way better than the theory I came up with 😅). 
I tried your patch and got basically the same improvement (~2x speed increase) 🎉

Same test (v1.70.3 with fadvise disabled):
```
Transferred:       64.297 GiB / 64.297 GiB, 100%, 110.439 MiB/s, ETA 0s
Elapsed time:      8m50.0s

real    8m50.098s
```

So my impression based on the issue you linked is that the "proper" way forward here would be to make fadvise configurable - if that's the case I would close this PR. 

It may be beyond my linux / golang skills, but I could give a try at https://github.com/rclone/rclone/issues/7886 and making it configurable.

vitorog:
Closing this PR as there is a bigger discussion about this topic here: https://github.com/rclone/rclone/pull/8723

vitorog:
hi, @ncw , I'm re-opening this issue because after running more tests in a different setup, it seems fadvise is not the only thing impacting speeds 😅

I'm running a setup in Kubernetes where some SMB volumes are mounted in the nodes. I'm using rclone to transfer files from this storage to S3. Since the volumes are mounted in the nodes, rclone uses the "local" backend (not the SMB one).

Running the same test (64 GB file upload to S3 - g4dn.4xlarge EC2 instances in AWS Outpost):

| Setting | Avg Speed |
|:---|---:|
| `fadvise disabled` |  ~165 MB/s |
| `buffering re-enabled` | ~310 MB/s |
| `buffering re-enabled + fadvise disabled` | ~347 MB/s |

I think since we are reading data from a network disk (SMB), disabling buffering has a huge impact when doing the multithreaded uploads to S3. 
What do you think?

ncw:
This is the conclusion that the restic project came to - disable fadvise and have RAM buffing for max performance.

I don't want to balloon the memory usage of rclone though there was a reason we did this - issue #7350 - that issue has a lot of good stuff in - it is worth a read.

Adaping the table from #7350 this is what is actually implemented now

- `RAM`: Buffer in RAM
- `None`: Don't buffer, but re-read from the source if necessary.

The logic is `RAM` unless
1. the destination supports `OpenWriterAt`, eg local, azurefiles, smb, pcloud => `None`
2. the source is local => `None`
3. the destination supports `OpenChunkWriter` and promises not to seek its chunks except for retries, eg b2 => `None`

| Source backend | Destination backend| Buffering |
| --- | --- | --- |
| local | any | None¹ |
| any | local/azurefiles/smb/pcloud | None² |
| any | b2 | None |
| any | s3/azureblob/oos | RAM |

And the notes from before

¹ Needs performance testing to see if it slows stuff down a lot! Might need to be RAM **Yes it does slow stuff down**

² It works like this at the moment as the local backend never needs retries. (OpenWriterAt doesn't read the data twice)

Does that mean we should make it configurable? I hesitate to add yet another configuration flag for the poor users though. #7350 suggests a --low-memory flag which would make sure we used disk buffering for local reads.

Perhaps a more targeted flag like `--multi-thread-low-memory` which if true uses `None` strategy, so make the rules

The logic is `RAM` unless
1. the destination supports `OpenWriterAt`, eg local, azurefiles, smb, pcloud => `None`
2. the source is local **and** `--multi-thread-low-memory` == `true` => `None`
3. the destination supports `OpenChunkWriter` and promises not to seek its chunks except for retries, eg b2 => `None`

We do now have `--max-buffer-memory` which can be used to control how much memory rclone is using and that does work pretty well which could be used instead of `--multi-thread-low-memory`.

What do you think @vitorog ?

darthShadow:
Strictly for SMB (https://linux.die.net/man/8/mount.cifs), there also seem to be 2 options available that the users can tune themselves:

* The FS-Cache via the `fsc` option
* Setting `cache` to `loose` rather than the default value of `strict`

I wonder if any of these allow you to get the same performance as that of RAM caching via rclone without needing any changes.

PR #8721, opened 2025-08-01
accounting: populate transfer snapshot with "what" value

TimothyBJacobs:

The transfer snapshot contains useful information about the result of a transfer, but without knowing _what_ that transfer was doing, it makes it harder to reason about since a file can go through many different tranfers.

In my case, we're monitoring the active transfers, and want to be able to better distinguish between the type of failures.

No
#### Checklist

### DIFF
--- a/fs/accounting/transfer.go
+++ b/fs/accounting/transfer.go
@@ -17,6 +17,7 @@ type TransferSnapshot struct {
 	Size        int64     `json:"size"`
 	Bytes       int64     `json:"bytes"`
 	Checked     bool      `json:"checked"`
+	What        string    `json:"what"`
 	StartedAt   time.Time `json:"started_at"`
 	CompletedAt time.Time `json:"completed_at,omitempty"`
 	Error       error     `json:"-"`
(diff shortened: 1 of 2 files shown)


ncw (commented):
This is great - thank you :-)

A couple of thoughts

1) This is a user visible change as it will change the output of the `core/transferred` stats, so the documentation needs updating there

https://github.com/rclone/rclone/blob/d4e68bf66b0b98bd35bf3eb1aab4fed6db317e75/fs/accounting/stats_groups.go#L175

(I think that is the right place - please double check!)

2) If `what` is "" then it is a transfer, so maybe when we make the transfer snapshot we should replace "" with "transferring" (I think the other things are like that "checking")

What do you think?

TimothyBJacobs:
Thanks for reviewing!

> This is a user visible change as it will change the output of the core/transferred stats, so the documentation needs updating there

Oops, missed that. Updated. I believe I've now checked through all the usages of the `TransferSnapshot`. It's also used in the `bisync` cmd as debug output, but doesn't look like there is accompanying documentation about that shape.

> If what is "" then it is a transfer, so maybe when we make the transfer snapshot we should replace "" with "transferring" (I think the other things are like that "checking")

Yeah, that makes sense to me. Updated.

TimothyBJacobs:
@ncw Could you take another look at this?

ncw:
Thank you :-)

PR #8723, opened 2025-08-03
Support disabling fadvise with "--disable-fadvise" flag

vitorog:

This change adds a new flag ("--disable-fadvise"), that allow disabling fadvise for the local backend. 
Using fadvise causes performance issues when doing multipart copies to S3, for example. It also could impact some use cases where rclone copies data that it doesn't own (for example, database backups on live databases).

Tested copying a 64GB file from local to S3 in a m6a.2xlarge AWS EC2 instance:
`time ./rclone copyto --s3-no-check-bucket --ignore-checksum --s3-disable-checksum --progress --s3-upload-cutoff=0 --multi-thread-cutoff=256M --multi-thread-streams 20 --disable=copy --no-check-dest <Local File> <S3 Bucket>`

| Flag | Time | Avg Speed |
|:---|---|---:|
| `no flag` | 18 min 46s | ~58 MB/s |
| `--disable-fadvise` | 8m 37s | ~126 MB/s |

https://github.com/rclone/rclone/issues/7886
https://github.com/rclone/rclone/pull/8720

#### Checklist


darthShadow:
Does it make sense to only disable the `unix.FADV_DONTNEED` calls while still keeping the `unix.FADV_SEQUENTIAL` calls? I believe the `unix.FADV_SEQUENTIAL` calls will still be useful as kernel hints for readahead.

darthShadow:
It may also be worth considering whether we want to make this opt-out instead of opt-in, so something like `--enable-fadvise-clear`.

I really don't think rclone should be clearing the  page cache of other apps (if it is copying the underlying files) and as has been said before: "unused RAM is wasted RAM". Another point to that argument is that even restic has stopped doing this and removed all code related to it: https://github.com/restic/restic/pull/670.

ncw:
Thanks for doing this @vitorog 

@darthShadow wrote
> Does it make sense to only disable the `unix.FADV_DONTNEED` calls while still keeping the `unix.FADV_SEQUENTIAL` calls? I believe the `unix.FADV_SEQUENTIAL` calls will still be useful as kernel hints for readahead.

Hmm don't know. We are specifically not reading the files sequentially in the multipart upload case we are reading a block, rewinding, reading the block again, rewinding reading the blockagain , then reading the next block...

Not all backends do this though.

I'm probably favouring not supplying `unix.FADV_SEQUENTIAL`  either as it isn't true a lot of the time.

@darthShadow wrote:
> It may also be worth considering whether we want to make this opt-out instead of opt-in, so something like `--enable-fadvise-clear`.
> 
> I really don't think rclone should be clearing the page cache of other apps (if it is copying the underlying files) and as has been said before: "unused RAM is wasted RAM". Another point to that argument is that even restic has stopped doing this and removed all code related to it: [restic/restic#670](https://github.com/restic/restic/pull/670).

I agree, this should be opt in, or we should just delete the code like they did in restic. Restic is a very similar use case.

The fact that we don't do this on any other OSes is a good argument for removing it completely I think.

Thoughts @darthShadow and @vitorog ?

darthShadow:
> unix.FADV_SEQUENTIAL

I would think it's still useful for a potential case in the future where we avoid needing to actually re-seek in the file multiple times and can instead do it with an in-memory buffer (probably with the pool.RW implementation?). Reference from the linked restic issue (https://github.com/restic/restic/issues/666#issuecomment-259815592) about how it was problematic for them:

> After a while I figured out that reading data twice is a bad idea, mostly because the data may have changed between computing the chunk boundaries and saving the data to the repo.

I would even like to double-down on the sequential approach by also asking the kernel to prefetch the next block of data into cache by adding the `unix.FADV_WILLNEED` flag (something like https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/lib/fs/fadvise_unix.go#L12), but I think that can be done in a different PR.

---

> unix.FADV_DONTNEED

> just delete the code

I am in favor of deleting the code, but I am not sure of the history of the issue and why it was required at all in the first place. Is the original scenario no longer a concern?

@mmatczuk Since you are the original PR author for this, any thoughts about the removal?

vitorog:
Thanks @ncw and @darthShadow  for all the input and discussion. Thinking about my use case (multipart uploads to S3), it seems fadvise is not required at all, so I'm also fine with removing it, but I have no idea about the implications for others...
I implemented the change as "opt-out" to not change the current behaviour - but after reading all discussions and issues I also agree it should be "opt-in" instead (if we decide to keep it).

Happy to support if I can with what is decided 😅

paf0186:
Hey!  I'm a developer on the Lustre parallel file system, so I'm a linux kernel type.  I'd like to remove this code, , since the many parallel fadvise calls are killing performance.

For vitorog, fadvise is never required - it's advisory, and the system can ignore it (though not at the FS leve).  It would only ever be a performance enhancement.

The way rclone is using fadvise - after every r/w cycle - it ignores the usual page reclaim mechanisms and tries to do page reclaim from every rclone thread (in the OS sense, the OS level schedulable entities) at the same time.  This can be bad for performance, since normally the reclaim would be handled by a dedicated background thread at the OS level (kswapd, generally, which manages much more than swap).

It looks something like this on Lustre, with many threads calling in over and over to free pages piling up on a particular spinlock, because there are so many trying to act at once rather than a single thread taking on more of the work (the same work, but all done from one thread is much faster):

`   26     69.29%  rclone           [kernel.kallsyms]           [k] native_queued_spin_lock_slowpath
   25             |
   24             ---native_queued_spin_lock_slowpath
   23                |
   22                |--67.73%--__lock_text_start
   21                |          |
   20                |           --67.69%--__wake_up_common_lock
   19                |                     |
   18                |                      --67.69%--osc_page_delete
   17                |                                cl_page_delete0
   16                |                                cl_page_delete
   15                |                                ll_releasepage
   14                |                                invalidate_inode_page
   13                |                                __invalidate_mapping_pages
   12                |                                generic_fadvise
   11                |                                ksys_fadvise64_64
   10                |                                __x64_sys_fadvise64
    9                |                                do_syscall_64
    8                |                                entry_SYSCALL_64_after_hwframe
    7                |                                0x40714e
    6                |                                |`

We've got an improvement to the locking in Lustre we're going to make, but it's only coming up as a hotspot because rclone is hitting fadvise from many threads.  So I would definitely support removing fadvise - it's trying to take into the hands of the application something that's really for the file system to manage.  The standard method for freeing the page cache is much faster and avoids this sort of contention.

I'm going to refresh this PR to make it just remove fadvise and not have an option; let me know what else I can do to help here.  (I'm also OK with giving it a command line option if that's preferred, but I definitely think it should at least be off-by-default.)

paf0186:
I've opened #9280 which removes all fadvise calls entirely rather than adding a flag. The discussion here seemed to converge on removal being the right approach (consistent with restic's decision), so that PR just deletes the code.

If reviewers prefer to keep FADV_SEQUENTIAL while only removing FADV_DONTNEED, happy to adjust.

ncw:
I have merged #9280 which removes fadvise unconditonally which is my preferred approach. I will close this now - thank you for getting the ball rolling @vitorog

PR #8744, opened 2025-08-12
Multiple symlink resolution

mathstuf:

Support config files living at locations found by multiple symlink hops.

No.

#### Checklist

### DIFF
--- a/fs/config/configfile/configfile_test.go
+++ b/fs/config/configfile/configfile_test.go
@@ -387,12 +387,12 @@ func TestConfigFileSaveSymlinkAbsolute(t *testing.T) {
 	}
 
 	t.Run("Absolute", func(t *testing.T) {
-		link := filepath.Join(linkDir, "configfilelink")
+		link := filepath.Join(linkDir, "configfileabslink")
 		target := filepath.Join(testDir, "b", "configfiletarget")
 		testSymlink(t, link, target, target)
 	})
 	t.Run("Relative", func(t *testing.T) {
-		link := filepath.Join(linkDir, "configfilelink")
+		link := filepath.Join(linkDir, "configfilerellink")
 		target := filepath.Join("b", "c", "configfiletarget")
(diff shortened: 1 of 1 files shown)


ncw (commented):
This is a nice idea.

What about using [filepath.EvalSymlinks](https://pkg.go.dev/path/filepath#EvalSymlinks) that should simplify the logic?

mathstuf:
Ah, that does look a lot simpler. I'll try that out. I'll also add a test for a broken symlink.

mathstuf:
That breaks a lot of tests…

mathstuf:
I tried to use the `filepath.EvalSymlinks` API but that ends up breaking other tests. If the simplification is wanted, I think it needs a separate PR to also consider the behavior change(s) covered by other tests expecting specific errors.

ncw:
> I tried to use the `filepath.EvalSymlinks` API but that ends up breaking other tests. If the simplification is wanted, I think it needs a separate PR to also consider the behavior change(s) covered by other tests expecting specific errors.

I would prefer to use `filepath.EvalSymlinks` if possible as following symlinks properly is non trivial so I'd rather get the standard library to do it for us.

mathstuf:
I think that loses error reporting fidelity as most of the `configfile` test failure modes start reporting different error codes. But if that is acceptable, I can push harder on that API call.

ncw:
@mathstuf start pulling on that thread and see what unravels - we can review after.

mathstuf:
I don't think it's going to work out. I believe the issue is that if the file doesn't exist, we get *no* information about any intermediate path resolution in the meantime. We would have to `evalSymlinks` on each path component until we get a non-no-exist error and then paste back the path components stripped on the way there to get the absolute path where the configuration file *would* exist if the intermediate directory components existed (as normal directories) and the file config file name (as a normal file). I don't think it would be any simpler than the existing logic and almost certainly more expensive (due to repeated `EvalSymlinks` calls each starting from scratch rather than chasing from what we know we have and stopping when things fall apart).

Issue #4523, opened 2020-08-22
Destination DIR deleted on using --exclude-if-present flag


SubhashBose:
#### What is the problem you are having with rclone?
if I have a directory tree like this __both in local and remote__ 

```
dir1/
dir1/file1.txt
dir1/dir2/
dir1/dir2/file2.txt
```

Now I want rlone to simply exclude the __dir1__ during sync from tracking any modification either in remote or local. So I use the flag `--exclude-if-present .rcignoredir` add a file in the tree.

```
dir1/.rcignoredir
```

Now during sync, rclone in an attempt to exclude the directory, it deletes the entire __dir1__ tree in the remote/destination. I expected it not to delete anything from destination, but simply ignore any changes in __dir1__ on both sides.

I know I can do this with `--exclude /dir1` flag, which also works as I expected, i.e ignoring dir1 from sync without deleting anything from remote. 

So, I find this behavior inconsistent between `--exclude-if-present` flag and all other filterings flags (https://rclone.org/filtering/). For other flags, what I understand, rclone will only delete files from remote ONLY if I enable `--delete-excluded` additionally.

#### What is your rclone version (output from `rclone version`)
```
rclone v1.51.0
- os/arch: linux/amd64
- go version: go1.13.7
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Linux

####  Which cloud storage system are you using? (eg Google Drive)
Irrespective of any storage. But I am using SFTP for destination and source is a local directory

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)  

```
rclone sync  --checksum /local/dir/ remote: --exclude-if-present .rcignoredir
```

#### The rclone config contents with secrets removed.  

I think it's not needed

#### A log from the command with the `-vv` flag  

```
2020/08/21 21:19:16 DEBUG : rclone: Version "v1.51.0" starting with parameters ["rclone" "--config" "rclone.conf" "sync" "-P" "-vv" "--checksum" "./" "remote:" "--exclude" ".rcfilter" "--filter-from" ".rcfilter" "--exclude-if-present" ".rcignoredir" "--exclude" "upload_site.sh"]
2020/08/21 21:19:16 DEBUG : Using RCLONE_CONFIG_PASS password.
2020/08/21 21:19:16 DEBUG : Using config file from "rclone.conf"
2020/08/21 21:19:18 DEBUG : --removed-- to "SSH-2.0-OpenSSH_4.3"
[...]

PR #8746, opened 2025-08-13
fs: accept directory names in exclude-if-present

### LINKED ISSUE #8745: Accept directory names in –exclude-if-present

https://forum.rclone.org/t/accept-directory-names-in-exclude-if-present/52248/1

rclone v1.70.3
- os/version: darwin 15.3.2 (64 bit)
- os/kernel: 24.3.0 (arm64)
- os/type: darwin
- go/version: go1.24.4
- go/linking: dynamic
- go/tags: cmount

*copy from forum post*

I often treat directory and file names interchangeable in copy operations (often with a leading ‘/’ but with roughly same meaning).

I want to copy all directories inside my projects directory but not the ones that are git tracked as they’d have git upstream.

Specifying a directory name in ``--exclude-if-present`` errors out:

```
$ rclone ls test --exclude-if-present '.git'

2025/08/13 03:49:59 ERROR : error listing: is a directory not a file
2025/08/13 03:49:59 NOTICE: Failed to ls with 2 errors: last error was: is a directory not a file
```

The error is not very clear about what is a directory and not a file, probably having an error like ``.git is a directory not a file`` would make more sense here anyway.

Accept argument to ``--exclude-if-present`` to be any of directory or file

#### How to use GitHub

Hritik14:

Fixes: https://github.com/rclone/rclone/issues/8745

Forum: https://forum.rclone.org/t/accept-directory-names-in-exclude-if-present/52248/1
Issue: Fixes: https://github.com/rclone/rclone/issues/8745

#### Checklist


ncw (commented):
This would also need to change `ListContainsExcludeFile` I think.

These could do with tests and docs also, and the help for --exclude-if-present would need adjusting.

I am slightly concerned that this is a backwards incompatible change though.

Perhaps we should require directories in --exclude-if-present to have trailing `/` so you'd write `--exclude-if-present .ignoreFile --exclude-if-present .ignoreDir/` which would fix the backwards incompatibility problem.

What do you think?

Hritik14:
> This would also need to change ListContainsExcludeFile I think.

I'm not sure of the changes there. I suppose ``obj, ok := entry.(fs.Object)`` fails because dir is not considered an object ?

> require directories in --exclude-if-present to have trailing /

Makes a lot of sense. Added this.

> These could do with tests and docs also, and the help for --exclude-if-present would need adjusting.

I'm happy to make those changes after we agree on an implementation :)

fs/filter/filter.go
```diff
@@ -491,12 +491,16 @@ func (f *Filter) DirContainsExcludeFile(ctx context.Context, fremote fs.Fs, remo
 	if len(f.Opt.ExcludeFile) > 0 {
 		for _, excludeFile := range f.Opt.ExcludeFile {
 			exists, err := fs.FileExists(ctx, fremote, path.Join(remote, excludeFile))
-			if err != nil {
-				return false, err
-			}
 			if exists {
 				return true, nil
 			}
+			if err == fs.ErrorIsDir {
```
ncw: This looks like the right approach. The logic needs to go like this though otherwise it ignores errors on the exists check

```go
if err != nil {
    if errors.Is(err, fs.ErrorIsDir) && strings.HasSuffix(excludeFile, "/") {
      return true, nil
    }
    return false, err
}
if exists {
    return true, nil
}
```

ncw (commented):
This looks like the right approach now (see notes inline).

Please carry on with docs and tests - thank you :-)

ncw:
I think ListContainsExcludeFile needs to look something like this (untested)

```go
// ListContainsExcludeFile checks if exclude file is present in the list.
func (f *Filter) ListContainsExcludeFile(entries fs.DirEntries) bool {
	if len(f.Opt.ExcludeFile) == 0 {
		return false
	}
		basename := path.Base(entry.Remote())
		switch entry.(type) {
		case fs.Object:
			if slices.Contains(f.Opt.ExcludeFile, basename) {
				return true
			}
		case fs.Directory:
			if slices.Contains(f.Opt.ExcludeFile, basename+"/") {
				return true
			}
		}
	}
	return false
}
```

fs/filter/filter.go
Hritik14: above errors out in case exclude file doesn't have a `/` but name corresponds to a dir.

Eg:
```
test2
├── node-project
│   └── lol
└── python-project
    └── .venv/
```

```
rclone ls /Users/neo/tmp/test2 --exclude-if-present '.venv' --exclude-if-present '.git'
```

```
2025/08/30 18:59:27 ERROR : error listing: directory not found
2025/08/30 18:59:27 NOTICE: Failed to ls with 2 errors: last error was: directory not found
```

I wonder if we should even check for `/` suffix, since, rclone just crashes if it encounters a ignore-file name that exists as a directory in the fs.
Anyone relying on this behaviour is unlikely.

Hritik14: above errors out in case exclude file doesn't have a `/` but name corresponds to a dir.

Eg:
```
test2
├── node-project
│   └── lol
└── python-project
    └── .venv/
```

```
rclone ls /Users/neo/tmp/test2 --exclude-if-present '.venv' --exclude-if-present '.git'
```

```
2025/08/30 19:05:19 ERROR : error listing: is a directory not a file
2025/08/30 19:05:19 NOTICE: Failed to ls with 2 errors: last error was: is a directory not a file
```

I wonder if we should even check for `/` suffix, since, rclone just crashes if it encounters a ignore-file name that exists as a directory in the fs.
Anyone relying on this behaviour is unlikely.

Hritik14: above errors out in case exclude file doesn't have a `/` but name corresponds to a dir.

Eg:
```
test2
├── node-project
│   └── lol
└── python-project
    └── .venv/
```

```
rclone ls /Users/neo/tmp/test2 --exclude-if-present '.venv' --exclude-if-present '.git'
```

```
2025/08/30 19:05:19 ERROR : error listing: is a directory not a file
2025/08/30 19:05:19 NOTICE: Failed to ls with 2 errors: last error was: is a directory not a file
```

I wonder if we should even check for `/` suffix, since, rclone just crashes if it encounters a ignore-file name that exists as a directory in the fs.
Anyone relying on this behaviour is unlikely.

ncw: Ha, yes you are right. Existing rclone has a bug! I was trying to preserve the existing behaviour, but let's fix the bug

So that makes your original solution nearly right but I think we need to add errors.Is, so

```patch
-                       if err == fs.ErrorIsDir {
+                       if errors.Is(err, fs.ErrorIsDir) {
                                return strings.HasSuffix(excludeFile, "/"), nil
                        }
```

Can you do that then add tests and docs (in `docs/content/filtering.md` also, and the help for --exclude-if-present would need adjusting.

https://github.com/rclone/rclone/blob/9e200531b1490000656031c42fff4d95477e1b46/fs/filter/filter.go#L36

Maybe something like

"Exclude directories if file or dir/ is present"

Thanks

Hritik14: > "Exclude directories if file or dir/ is present"

given it was a bug, do we even require the trailing `/` for directories ? most of the other commands do not distinguish against file/dir in this manner, (say copy).

ncw:
The options

1. `--exclude-if-present name` should exclude the containing directory if `name` is a **file or directory**
2. `--exclude-if-present name` should exclude the containing directory if `name` is a **file only** and we use `--exclude-if-present name/` to exclude a directory

What I normally do here is look for prior art to see what other programs do. Here is a selection of programs with an `--exclude-if-present` flag or similar. I didn't actually try these, I looked at the docs so they may not be 100%

| Program | Exclude can be a Directory |
|---|---|
| borg | Yes |
| restic | No |
| duplicity | No |
| rdiff-backup | Yes |
| tar | No |
| bupstash | No |

There isn't a strong consensus for whether files and directories should be treated differently or not.

Given the fact that currently if `--exclude-if-present` finds a directory with that name it errors the program, I think we are free to make our own choice.

If we choose option 2) if you want files and directories you can always add `--exclude-if-present name --exclude-if-present name/` which makes it more flexible.

@albertony @nielash do you have an opinion?

nielash:
> @albertony @nielash do you have an opinion?

I lean slightly towards option 2, for a few reasons.

- It is more consistent with the "[How filter rules are applied to directories](https://rclone.org/filtering/#directory_filter)" section of the docs
- It is more flexible (to your point about including the flag twice)
- Treating files and dirs as interchangeable could lead to unexpected results (suppose I want it to apply to dirs but not to files with the same name)

I hear the point about `rclone copy` accepting either a dir or file as arguments, but that's slightly different from _filtering rules,_ which do care about the trailing slash. (I also have always found the `copy` behavior a bit odd, as the [docs](https://rclone.org/commands/rclone_copy/#synopsis) specifically instruct: "To copy single files, use the `copyto` command instead.")

Also, it may or may not be worth revisiting #4523 at the same time.

albertony:
> I lean slightly towards option 2, for a few reasons.

I agree - with everything you said there!

fs/filter/filter.go
ncw: I think the consensus is `--exclude-if-present file` and `--exclude-if-present dir/`.

Do you want to finish of the PR assuming that and we can get it merged - thank you

Hritik14: Yes, I agree with the reasoning by @nielash. I'm a little occupied with my day job but I do intend to finish the rest of it soon.

Hritik14:
The current implementation had a bug where having a `/` in the filter would look for both file and dir - thanks to the tests.

here's how it looked
```
; tree test2 -a
test2
├── node-project
│   └── lol
└── python-project
    └── .venv
        └── file

4 directories, 2 files
; /Users/neo/go/bin/rclone ls test2 --exclude-if-present lol/
        0 python-project/.venv/file
; /Users/neo/go/bin/rclone ls test2 --exclude-if-present lol
        0 python-project/.venv/file
```

I've fixed this behavior in the latest commit, although the a test fails right now since ``mockfs`` does not support adding directories and exclusively works on root. I might need to implement directories in mockfs for this test to pass unless you can suggest a better way ?

Hritik14:
@ncw I've added the req mockfs impl and docs. It's ready for review!

docs/content/filtering.md
```diff
@@ -915,13 +915,13 @@ format.
 
 Useful for debugging.
 
-## Exclude directory based on a file
+## Exclude directory based on a file or directory
```
ncw, line 933: Typo: "directtories" → "directories"

ncw (commented):
Thanks for the update - this is looking pretty good! A few issues to address before merging.

The overall approach is correct and the `ListContainsExcludeFile` changes are clean. See inline comments for specific issues.

docs/content/filtering.md
```diff
@@ -915,13 +915,13 @@ format.
...
 
 The `--exclude-if-present` flag controls whether a directory is
 within the scope of an rclone command based on the presence of a
-named file within it. The flag can be repeated to check for
-multiple file names, presence of any of them will exclude the
-directory.
+named file or a directory within it. The flag can be repeated to
+check for multiple file names or directtories, presence of any
+of them will exclude the directory.
```
ncw, line 939: Typo: "directtories" → "directories"

fstest/mockfs/mockfs.go
```diff
@@ -106,21 +114,36 @@ func (f *Fs) Features() *fs.Features {
...
-		return f.rootDir, nil
+	for _, entry := range f.entries {
+		remote := entry.Remote()
+		if !strings.HasPrefix(remote, dir) {
+			continue
+		}
+		remaining := remote[len(dir):]
+		// Only direct children
+		if !strings.Contains(remaining, "/") {
+			entries = append(entries, entry)
```
ncw, line 125: Bug: the prefix matching here is broken for non-root directories.

Two issues:
1. **False prefix match**: If `dir="foo"` and an entry has remote `"foobar/file"`, `strings.HasPrefix("foobar/file", "foo")` is true — incorrectly matching a sibling directory.
2. **Separator not stripped**: When `dir="subdir"` and entry is `"subdir/file"`, `remaining="/file"` contains `/`, so it gets excluded when it should be a direct child.

The fix should check against `dir + "/"` as the prefix when `dir != ""`:

```go
func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err error) {
	prefix := ""
	if dir != "" {
		prefix = dir + "/"
	}
	for _, entry := range f.entries {
		remote := entry.Remote()
		if !strings.HasPrefix(remote, prefix) {
			continue
		}
		remaining := remote[len(prefix):]
		// Only direct children
		if !strings.Contains(remaining, "/") {
			entries = append(entries, entry)
		}
	}
	if len(entries) == 0 && dir != "" {
		return nil, fs.ErrorDirNotFound
	}
	return entries, nil
}
```

The current tests only test `dir=""` so they don't catch this. Please add a test for listing a subdirectory.

fs/filter/filter.go
```diff
@@ -491,12 +495,16 @@ func (f *Filter) DirContainsExcludeFile(ctx context.Context, fremote fs.Fs, remo
...
 			exists, err := fs.FileExists(ctx, fremote, path.Join(remote, excludeFile))
+			if exists {
+				return !strings.HasSuffix(excludeFile, "/"), nil
+			}
+			if errors.Is(err, fs.ErrorIsDir) {
+				return strings.HasSuffix(excludeFile, "/"), nil
+			}
 			if err != nil {
 				return false, err
 			}
```
ncw, line 522: This `DirContainsExcludeFile` approach relies on `NewObject` returning `fs.ErrorIsDir` when the remote points to a directory. However, bucket-based backends (S3, B2, GCS, Azure Blob, Swift, etc.) do **not** return `ErrorIsDir` from `NewObject` — they return either `ErrorObjectNotFound` or `ErrorNotAFile`, both of which cause `FileExists` to return `(false, nil)`, so the `errors.Is(err, fs.ErrorIsDir)` branch is never reached.

This means `--exclude-if-present dir/` will behave differently depending on backend type:

**How it works — two code paths:**
1. `ListContainsExcludeFile` — called in `DirSorted`/`DirSortedFn` when a directory's entries have already been listed. It checks if the current batch of entries contains the exclude marker. If found, the directory's contents are excluded. This works on **all** backends because it checks the entry type from the `List` response.

2. `DirContainsExcludeFile` (this function) — called via `IncludeDirectory` in `filterDir` for each subdirectory encountered. It proactively checks whether a child directory should be recursed into **before** listing it. This is the primary exclusion mechanism — it filters the directory out of the parent listing entirely.

**Consequence on bucket-based backends:**
- `DirContainsExcludeFile` fails to detect the directory marker → returns false → the excluded directory is **not filtered out** of the parent listing
- The walker then recurses into the directory, calls `List`, and `ListContainsExcludeFile` catches it → returns empty contents
- **Net result**: the directory itself still appears in listings and may be created (empty) on the destination during sync, but its contents are correctly excluded

**Consequence on hierarchical backends** (local, drive, sftp, etc.):
- `DirContainsExcludeFile` detects the directory via `ErrorIsDir` → returns true → the directory is filtered out of the parent listing entirely
- The directory does not appear in listings and is not created on the destination

Please add a note in the docs mentioning that on bucket-based backends (S3, B2, GCS, Azure Blob, Swift, etc.), `--exclude-if-present dir/` will exclude the directory's contents but the empty directory itself may still appear in listings. On hierarchical backends (local, drive, sftp, etc.) the directory is excluded completely.

fs/types.go
```diff
@@ -113,7 +113,7 @@ type ObjectInfo interface {
 }
 
 // DirEntry provides read only information about the common subset of
-// a Dir or Object.  These are returned from directory listings - type
+// a Directory or Object.  These are returned from directory listings - type
```
ncw: Unrelated cosmetic change — please drop this from the PR or put it in a separate commit.

Hritik14:
@ncw added tests and other suggestions, please review :)

Hritik14:
@ncw gentle ping

Hritik14:
@ncw gentle ping :)

Hritik14:
@ncw gentle ping

PR #8750, opened 2025-08-14
azureblob: add metadata and tags support across upload and copy paths

cliff-openai:

This change adds first-class metadata support to the Azure Blob backend, including headers, user metadata, tags, and modtime overrides, and wires it through uploads and server-side copies.

There is a behavior change in that rclone will now set the "mtime" custom metadata when doing server side copies to azure and the `--metadata` argument is given.

Highlights
- Map standard headers: cache-control, content-disposition, content-encoding, content-language, content-type → corresponding x-ms-blob-* HTTP headers.
- Map user metadata: any non-reserved keys (excluding x-ms-*) are sent as blob user metadata. Keys are normalized to lowercase for consistency.
- Support tags: parse `x-ms-tags` as a comma-separated list of key=value pairs and apply them on uploads and copies.
- Support mtime override: accept `mtime` in metadata (RFC3339/RFC3339Nano) to override the stored modtime persisted in user metadata.

Upload integration
- Both singlepart and multipart uploads call a common mapper to translate `--metadata`, `--metadata-set`, and `--metadata-mapper` output into Azure HTTP headers, user metadata, and tags.
- Ensures `mtime` is recorded in user metadata, honoring overrides.
- Preserves and merges computed MD5 when enabled.

Tests
- Add backend/azureblob/azureblob_internal_test.go covering:
  - Singlepart/multipart uploads with metadata mapping.
  - Singlepart/multipart copies with metadata mapping and Content-Type fallback behavior.
  - Validation helpers asserting headers and user metadata on the blob.

Not directly, there is https://github.com/rclone/rclone/pull/8399 which seems abandoned (a smaller, less-complete version of this change).

I am willing to create an issue for this if it would be helpful.

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate.  (although I haven't tested every corner case, let me know if I should add more testing)

### DIFF
--- a/backend/azureblob/azureblob_internal_test.go
+++ b/backend/azureblob/azureblob_internal_test.go
@@ -5,13 +5,19 @@ package azureblob
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
+	"net/http"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
 	"github.com/rclone/rclone/fs"
+	"github.com/rclone/rclone/fs/object"
 	"github.com/rclone/rclone/fstest"
 	"github.com/rclone/rclone/fstest/fstests"
+	"github.com/rclone/rclone/fstest/testserver"
 	"github.com/rclone/rclone/lib/random"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


ncw:
I gave this a quick look - looks like good work - thank you :-)

It doesn't seem to set any Flags in Features, eg

https://github.com/rclone/rclone/blob/d9a36ef45cbdb80bfd870d0ad1cbca2bea2cf021/backend/s3/s3.go#L3974-L3976

It also isn't setting the `MetadataInfo` in the backend initialization

https://github.com/rclone/rclone/blob/d9a36ef45cbdb80bfd870d0ad1cbca2bea2cf021/backend/s3/s3.go#L228-L230

Can you set those things, then run the integration tests and see if they pass as the integration tests won't do anything to do with metadata unless the feature flags are set.

cliff-openai:
Thanks for the early feedback @ncw !
Sorry that I didn't notice that you had sent feedback until just now.

I think that maybe I did as you asked, and that it passes tests.

roucc:
Hi @cliff-openai, I have ran the `test_all -remotes TestAzureBlob:` command (after upgrading it to V2), and have gotten 1 test failure:

```
"go test -v -timeout 1h0m0s -remote TestAzureBlob: -test.run '^TestIntegration$/^FsMkdir$/^FsPutFiles$'" - Starting (try 5/5)
=== RUN   TestIntegration
    fstests.go:438: Using remote "TestAzureBlob:"
2025/10/30 16:19:57 NOTICE: TestAzureBlob: Starting server
    fstests.go:146: 
        	Error Trace:	/home/dougal/rclone/fstest/fstests/fstests.go:146
        	            				/home/dougal/rclone/fstest/fstests/fstests.go:162
        	            				/home/dougal/rclone/fstest/fstests/fstests.go:224
        	            				/home/dougal/rclone/fstest/fstests/fstests.go:955
        	            	PUT https://rclone.blob.core.windows.net/rclone-test-wagukaf7yefi/file name.txt
        	            	--------------------------------------------------------------------------------
        	            	RESPONSE 400: 400 The metadata specified is invalid. It has characters that are not permitted.
        	            	ERROR CODE: InvalidMetadata
        	            	--------------------------------------------------------------------------------
        	            	﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidMetadata</Code><Message>The metadata specified is invalid. It has characters that are not permitted.
        	            	RequestId:835c1d3b-601e-0060-07b9-49c062000000
        	            	Time:2025-10-30T16:19:57.1785915Z</Message></Error>
        	            	--------------------------------------------------------------------------------
        	Test:       	TestIntegration/FsMkdir/FsPutFiles
        	Messages:   	Put
=== NAME  TestIntegration/FsMkdir
    fstests.go:2817: Warning: this should produce fs.ErrorDirNotFound
--- FAIL: TestIntegration (0.20s)
    --- FAIL: TestIntegration/FsMkdir (0.12s)
        --- FAIL: TestIntegration/FsMkdir/FsPutFiles (0.08s)
FAIL
exit status 1
FAIL	github.com/rclone/rclone/backend/azureblob	0.203s
"go test -v -timeout 1h0m0s -remote TestAzureBlob: -test.run '^TestIntegration$/^FsMkdir$/^FsPutFiles$'" - Finished ERROR in 692.895239ms (try 5/5): exit status 1: Failed [TestIntegration/FsMkdir/FsPutFiles]
```

Any ideas?
Other than this it is ready to merge, thank you!

cliff-openai:
Gah, this took me too long to figure out.

The problem is that I had the fix in my local worktree and I had just never committed/pushed it.

You can see the fix/description in https://github.com/rclone/rclone/pull/8750/commits/488cfe07ad58e13dae0d01d1162799bed3dda566, but ultimately the problem is that the test was using a metadata key of `rclone-test`, and Azure Blob Storage does not support hyphens in metadata key values.

backend/azureblob/azureblob_internal_test.go
```diff
@@ -148,4 +154,437 @@ func (f *Fs) testWriteUncommittedBlocks(t *testing.T) {
 func (f *Fs) InternalTest(t *testing.T) {
 	t.Run("Features", f.testFeatures)
 	t.Run("WriteUncommittedBlocks", f.testWriteUncommittedBlocks)
+	t.Run("Metadata", f.testMetadataPaths)
+}
+
+// Standalone runner for metadata path tests to allow easy filtering with -run
+func TestAzureMetadataPaths(t *testing.T) {
```
ncw: The unit tests are failing because there is no valid config for azure blob when running in the GitHub CI.

We do the integration tests on a separate machine with valid configs for all providers (and a whole suite of other tests).

So this test needs to Skip if there isn't a valid config.

However I think you probably need to delete `TestAzureMetadataPaths` completely and just call `testAzureMetadataPaths` from `InternalTest`

That will ensure it is only called when there is a valid config.

You should be able to filter these tests (with a longer path admittedly) with `-run` just fine.

cliff-openai:
Thanks! I’ve dropped the standalone TestAzureMetadataPaths helper and now call testMetadataPaths only from InternalTest, so the metadata suite stays within the integration harness and hopefully skips the CI config requirement.

ncw:
Hi @cliff-openai - sorry for the delay looking at this.

I ran the full integration test suite but unfortunately we are still seeing failures in the `backend/azureblob` tests

```
cd backend/azureblob
go test -v -timeout 1h0m0s -remote TestAzureBlob: -run '^TestIntegration$/^FsMkdir$/^FsPutFiles$/^Internal$/^Metadata$/^(CopyMultipart|CopyMultipartTags|CopyMultipart_MetadataOnly)$'
```

Here is the log: [azureblob.log](https://github.com/user-attachments/files/23658944/azureblob.log)

So it looks like something is broken with the Metadata CopyMultipart tests

```
                    --- FAIL: TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata/CopyMultipart (4.86s)
                    --- FAIL: TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata/CopyMultipart_MetadataOnly (4.93s)
                    --- FAIL: TestIntegration/FsMkdir/FsPutFiles/Internal/Metadata/CopyMultipartTags (4.91s)
```

Any ideas?

cliff-openai:
ok, I believe that I reproduced the failure and fixed it.

I'm sorry about the many round trips required on this review!

PR #8756, opened 2025-08-18
backend/compress: add zstd compression

A1ex3:

Added support for reading and writing seekable Zstd-compressed archives using [github.com/klauspost/compress/zstd](https://github.com/klauspost/compress) and [github.com/SaveTheRbtz/zstd-seekable-format-go/pkg](https://github.com/SaveTheRbtz/zstd-seekable-format-go) (ZSTD seekable compression format implementation in Go).

Updated `Go` version from 1.24.0 to 1.24.4 to meet the requirements of `github.com/SaveTheRbtz/zstd-seekable-format-go/pkg`.

#### Checklist

### DIFF
--- a/backend/compress/compress.go
+++ b/backend/compress/compress.go
@@ -529,6 +529,8 @@ func isCompressible(r io.Reader, compressionMode int) (bool, error) {
 		if err != nil {
 			return false, err
 		}
+	case Uncompressed:
+		return false, nil
 	default:
 		return false, fmt.Errorf("unsupported compression mode %d", compressionMode)
 	}
--- a/backend/compress/compress.go
+++ b/backend/compress/compress.go
@@ -2,10 +2,8 @@
 package compress
 
 import (
-	"bufio"
 	"bytes"
 	"context"
-	"crypto/md5"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
@@ -21,7 +19,6 @@ import (
 
 	"github.com/buengese/sgzip"
 	"github.com/gabriel-vasile/mimetype"
-	"github.com/klauspost/compress/zstd"
 
 	"github.com/rclone/rclone/fs"
 	"github.com/rclone/rclone/fs/accounting"
@@ -125,6 +122,47 @@ this limit will be cached on disk.`,
 	})
 }
 
+// compressionModeHandler defines the interface for handling different compression modes
+type compressionModeHandler interface {
+	// processFileNameGetFileExtension returns the file extension for the given compression mode
+	processFileNameGetFileExtension(compressionMode int) string
+
+	// newObjectGetOriginalSize returns the original file size from the metadata
+	newObjectGetOriginalSize(meta *ObjectMetadata) (int64, error)
+
+	// isCompressible checks the compression ratio of the provided data and returns true if the ratio exceeds
+	// the configured threshold
+	isCompressible(r io.Reader, compressionMode int) (bool, error)
+
+	// putCompress compresses the input data and uploads it to the remote, returning the new object and its metadata
+	putCompress(
+		ctx context.Context,
+		f *Fs,
+		in io.Reader,
+		src fs.ObjectInfo,
+		options []fs.OpenOption,
+		mimeType string,
+	) (fs.Object, *ObjectMetadata, error)
+
+	// openGetReadCloser opens a compressed object and returns a ReadCloser in the Open method
+	openGetReadCloser(
+		ctx context.Context,
+		o *Object,
+		offset int64,
+		limit int64,
+		cr chunkedreader.ChunkedReader,
+		closer io.Closer,
+		options ...fs.OpenOption,
+	) (rc io.ReadCloser, err error)
+
+	// putUncompressGetNewMetadata returns metadata in the putUncompress method for a specific compression algorithm
+	putUncompressGetNewMetadata(o fs.Object, mode int, md5 string, mimeType string, sum []byte) (fs.Object, *ObjectMetadata, error)
+
+	// This function generates a metadata object for sgzip.GzipMetadata or SzstdMetadata.
+	// Warning: This function panics if cmeta is not of the expected type.
+	newMetadata(size int64, mode int, cmeta any, md5 string, mimeType string) *ObjectMetadata
+}
+
 // Options defines the configuration for this backend
 type Options struct {
 	Remote           string        `config:"remote"`
@@ -138,12 +176,13 @@ type Options struct {
 // Fs represents a wrapped fs.Fs
 type Fs struct {
 	fs.Fs
-	wrapper  fs.Fs
-	name     string
-	root     string
-	opt      Options
-	mode     int          // compression mode id
-	features *fs.Features // optional features
+	wrapper     fs.Fs
+	name        string
+	root        string
+	opt         Options
+	mode        int                    // compression mode id
+	features    *fs.Features           // optional features
+	modeHandler compressionModeHandler // compression mode handler
 }
 
 // NewFs constructs an Fs from the path, container:path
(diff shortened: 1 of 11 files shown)


kapitainsky:
Great idea to introduce this modern compression algorithm. My question is why you limit compression levels to just 4 options? I think it would be much more useful to support all standard levels 1 to 19. I do not see any reason why to use only 4 arbitrary selected levels. Or is it something enforced by zstd golang implementation?

For example myself I often use level 17 for online backup software as its speed on my computer nicely matches my max Internet uplink speed. I do not see why to remove option for fine tuning like this when it is available.

A1ex3:
The `Go` zstd library [github.com/klauspost/compress/zstd](https://github.com/klauspost/compress) supports only four compression levels:
```go
// EncoderLevel predefines encoder compression levels.
// Only use the constants made available, since the actual mapping
// of these values are very likely to change and your compression could change
// unpredictably when upgrading the library.
type EncoderLevel int

const (
	speedNotSet EncoderLevel = iota

	// SpeedFastest will choose the fastest reasonable compression.
	// This is roughly equivalent to the fastest Zstandard mode.
	SpeedFastest

	// SpeedDefault is the default "pretty fast" compression option.
	// This is roughly equivalent to the default Zstandard mode (level 3).
	SpeedDefault

	// SpeedBetterCompression will yield better compression than the default.
	// Currently it is about zstd level 7-8 with ~ 2x-3x the default CPU usage.
	// By using this, notice that CPU usage may go up in the future.
	SpeedBetterCompression

	// SpeedBestCompression will choose the best available compression option.
	// This will offer the best compression no matter the CPU cost.
	SpeedBestCompression

	// speedLast should be kept as the last actual compression option.
	// The is not for external usage, but is used to keep track of the valid options.
	speedLast
)
```

The `EncoderLevelFromZstd` function allows converting a standard zstd level to one of the levels supported by the library. However, there is practically no benefit from this conversion:
```go
// EncoderLevelFromZstd will return an encoder level that closest matches the compression
// ratio of a specific zstd compression level.
// Many input values will provide the same compression level.
func EncoderLevelFromZstd(level int) EncoderLevel {
	switch {
	case level < 3:
		return SpeedFastest
	case level >= 3 && level < 6:
		return SpeedDefault
	case level >= 6 && level < 10:
		return SpeedBetterCompression
	default:
		return SpeedBestCompression
	}
}
```

kapitainsky:
> The `Go` zstd library [github.com/klauspost/compress/zstd](https://github.com/klauspost/compress) supports only four compression levels:

Thank you for clarification. More granular levels would be nice but definitely are not critical. zstd with only 4 levels will make day and night difference compared to gzip anyway.

ncw (commented):
This looks like a useful feature - thank you :-)

All those switch statements

```go
switch {
	case Gzip:
	case Zstd:
	default:
}
```

Have made the code rather untidy though.

What do you think about factoring it into two implementations, defined by an interface. You can then separate the implementations into zstd.go and gzip.go (say)

That should restore a bit of sanity to the code.

A1ex3:
@ncw  Thanks for the suggestion! I've refactored the compression handling - almost everything has been moved behind an interface, with separate implementations in `gzip_handler.go`, `zstd_handler.go`, `uncompressed_handler.go` and `unknown_handler.go`. The codebase should now be much cleaner.

ncw:
The refactor looks great - thank you :-)

Can you add another target to https://github.com/rclone/rclone/blob/master/fstest/test_all/config.yaml

```
 - backend:  "compress"
   remote:   "TestCompressZstd:"
   fastlist: false
```

`TestCompressZstd` can be something like

```
[TestCompressZstd]
type = compress
remote = /tmp/rclone-compress-test-zstd
mode = zstd
```

And then check the integration tests all pass with `test_all -remotes TestCompressZstd:,TestCompress:`

Once we merge that will go into the integration tester and be run once per day.

Would it be possible to lose some dependencies? I think github.com/SaveTheRbtz/zstd-seekable-format-go uses go.uber.org/atomic which can be done in stdlib now I think and go.uber.org/zap - would it be possible to use the stdlib log/slog? That will integrate with rclone's logging system.

I will try to give the code another review in a day or two.

A1ex3:
> Can you add another target to https://github.com/rclone/rclone/blob/master/fstest/test_all/config.yaml
> ```
>  - backend:  "compress"
>    remote:   "TestCompressZstd:"
>    fastlist: false
> ```
> `TestCompressZstd` can be something like
>```
> [TestCompressZstd]
> type = compress
> remote = /tmp/rclone-compress-test-zstd
> mode = zstd
>```
> And then check the integration tests all pass with `test_all -remotes TestCompressZstd:,TestCompress:`

@ncw Added a target in `fstest/test_all/config.yaml` and ran integration tests with the following configuration:
```
[TestCompress]
type = compress
remote = /tmp/rclone-compress-test-gzip
mode = gzip
level = -1

[TestCompressZstd]
type = compress
remote = /tmp/rclone-compress-test-zstd
mode = zstd
level = 2
```
Results:
image

I decided to fork [github.com/SaveTheRbtz/zstd-seekable-format-go](https://github.com/SaveTheRbtz/zstd-seekable-format-go) and removed logging, since it was mostly needed for debugging the library itself.
In addition:
- replaced `go.uber.org/atomic` with `sync/atomic`
- removed `go.uber.org/*` dependencies
- lowered the minimal Go version to 1.24.0

Now rclone uses [github.com/a1ex3/zstd-seekable-format-go/pkg](https://pkg.go.dev/github.com/a1ex3/zstd-seekable-format-go/pkg).

Additionally, fixed discovered issues with `zstd` in `backend/compress`.

ncw:
Hi @A1ex3 sorry for the delay in looking at this. Your changes are great thank you :-) We ran the integration tests one last time locally and the bisync tests are failing - which they weren't for you earlier. Can you help?

Thanks

Nick

A1ex3:
Hi @ncw,
I’ve just merged the latest changes from `master` locally into `feat-compress-zstd` branch and re-ran the integration tests for `TestCompress` and `TestCompressZstd` - both passed successfully.
Could you please share the test report or steps to reproduce the failing bisync tests?
Also, do they fail for a specific compression type or for all of them?

gordan-bobic:
Is this likely to get merged imminently? I could really do with this feature.

roucc (approved):
Hi @A1ex3, I've rebased this and ran the tests and they all pass. Not sure what the problem with the bisync tests was before but @ncw did some work to fix them yesterday. Sorry this has taken so long and I'll merge this once the CI passes, thanks!

A1ex3:
Hi @roucc, thanks! Glad everything’s working properly now.

kapitainsky:
@A1ex3 could you please have a look at:

https://forum.rclone.org/t/compress-findings/53859

and provide your feedback?

gordan-bobic:
This is expected. Here are some related testing results from ZFS back when they implemented zstd. You need at least zstd-6 to match gzip-9, but zstd is massively faster.
zfs-compression-2

Regarding what the compression levels mean, I'm going to hazard a guess they are directly 1:1 mappings onto what the underlying zstd golang library does:
https://pkg.go.dev/github.com/klauspost/compress/zstd
So your question should probably be directed here:
https://github.com/klauspost/compress

kapitainsky:
> This is expected.

Sorry but what is expected? I have tested using stand-alone gzip and zstd and this golang implementation behaves at least weird. (see details in the linked forum thread)

I have used silesia files set for test and already zstd -3 produces better results than gzip -9. But in case of this golang implementation even `SpeedBestCompression` is worse that gzip. Does it mean that it is not using even zstd level 3?

gordan-bobic:
Like I said - your questions on this should really be directed here:
https://github.com/klauspost/compress

PR #8770, opened 2025-08-22
Fix #8766: Do not log successful unmount as an error

### LINKED ISSUE #8766: Successfull unmount should not be logged as an error
When I quit an `rclone mount --rc ...` process using `rclone rc core/quit exitCode=0`, I still get an error logged, e.g.

```
2025/08/21 16:50:50 ERROR : /data/target-1: Unmounted rclone mount
```

This line explains that:

https://github.com/rclone/rclone/blob/4af2f01abc97ed0fc499b5fbc415a2ac7328e5fb/cmd/mountlib/mount.go#L399

I think, this should be changed to `fs.Infof()` or `fs.Logf()`.

tvogel:

Fix #8766

Yes, acknowledged in #8766.

#### Checklist

### DIFF
--- a/cmd/mountlib/mount.go
+++ b/cmd/mountlib/mount.go
@@ -396,7 +396,7 @@ func (m *MountPoint) Wait() error {
 			if err := m.Unmount(); err != nil {
 				fs.Errorf(m.MountPoint, "Failed to unmount: %v", err)
 			} else {
-				fs.Errorf(m.MountPoint, "Unmounted rclone mount")
+				fs.Logf(m.MountPoint, "Unmounted rclone mount")
 			}
 		})
 	}


ncw:
@nielash there is an interesting unrelated bisync test failure in the actions run here you might want to take a quick look at

nielash:
> @nielash there is an interesting unrelated bisync test failure in the actions run here you might want to take a quick look at

Thanks. This was very tricky to debug, as I can't reliably reproduce it, but I think the issue is this: https://github.com/nielash/rclone/commit/fbfc4dcbc52e23dd584e6768265fc7d6102b560f

ncw:
@nielash that looks plausible. Can you trigger it with a sleep or maybe even a sync point in the tests?

nielash:
That's exactly what I did. But it would still fail only some of the time.

ncw:
An alternate fix might be not to count errors in the cleanup function. We could do this by calling `ctx = accounting.WithStatsGroup(ctx, group)` to make a new context which will count the stats not in the global stats counter. That might be easier and would fix other potential problems in the future. What do you think?

nielash:
I definitely think that would be a helpful thing!

While we're dreaming big, I'd also love something similar for logging, so that (for example) the `fs.Infof` calls from parallel subtests could log to separate outputs, or different contexts could use different log levels. (Looks like it's [hard-coded](https://github.com/rclone/rclone/blob/b010591c96d9b4451bfda0cedf5eead1f710b4e4/fs/log.go#L176) at the moment.)

ncw:
> I definitely think that would be a helpful thing!
> 
> While we're dreaming big, I'd also love something similar for logging, so that (for example) the `fs.Infof` calls from parallel subtests could log to separate outputs, or different contexts could use different log levels. (Looks like it's [hard-coded](https://github.com/rclone/rclone/blob/b010591c96d9b4451bfda0cedf5eead1f710b4e4/fs/log.go#L176) at the moment.)

The only reason this hasn't been done is that we would need to change all `fs.Logf`, `fs.Debugf` etc calls and add a context which would be quite a lot of work given that there are > 2000 instances and we'd need to make sure each one had a `ctx` which would likely mean quite a bit of context propagation. I think there is an issue about it somewhere. It is one of those things that we really should do sometime though!

nielash:
> The only reason this hasn't been done is that we would need to change all `fs.Logf`, `fs.Debugf` etc calls and add a context which would be quite a lot of work given that there are > 2000 instances and we'd need to make sure each one had a `ctx` which would likely mean quite a bit of context propagation. I think there is an issue about it somewhere. It is one of those things that we really should do sometime though!

Makes sense! Could potentially set them all to `context.TODO()` to start, and fix them over time as needed.

ncw:
I tried the simple fix c185ed7f5d3ee229549226d4938cbb3ec39ffbf0 but it failed after 5 tries https://github.com/rclone/rclone/actions/runs/17301504427/job/49135392679 which looks like the same problem but I'm not 100% sure

nielash:
> I tried the simple fix [c185ed7](https://github.com/rclone/rclone/commit/c185ed7f5d3ee229549226d4938cbb3ec39ffbf0) but it failed after 5 tries https://github.com/rclone/rclone/actions/runs/17301504427/job/49135392679 which looks like the same problem but I'm not 100% sure

This tweak fixes it for me... https://github.com/nielash/rclone/commit/336c1da267b3f72fc0eab4b38bb481bf79f444d5
https://github.com/nielash/rclone/actions/runs/17340896008

ncw:
Thanks @nielash - I've merged those two commits now so cross fingers!

ncw:
@nielash Ahhhh, that has killed a lot of [integration tests](https://pub.rclone.org/integration-tests/current/) :-(

nielash:
> @nielash Ahhhh, that has killed a lot of [integration tests](https://pub.rclone.org/integration-tests/current/) :-(

Ahh, sorry! Luckily it is the same error over and over, and should be easy to fix.

PR #8773, opened 2025-08-25
backend/ulozto: implement the about functionality for uloz.to

metlos:

This PR implements the about info for the uloz.to remote. Because the provider doesn't report the total size of the available space, this is solved by simply declaring the total size as a configuration value of the remote.

The used size is properly detected and the free size is then simply computed as the difference between the two.

No.

#### Checklist

### DIFF
--- a/backend/ulozto/ulozto.go
+++ b/backend/ulozto/ulozto.go
@@ -49,7 +49,6 @@ type Options struct {
 	RootFolderSlug string               `config:"root_folder_slug"`
 	Enc            encoder.MultiEncoder `config:"encoding"`
 	ListPageSize   int                  `config:"list_page_size"`
-	TotalSize      fs.SizeSuffix        `config:"total_size"`
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


backend/ulozto/ulozto.go
```diff
@@ -97,7 +98,13 @@ any root slug set.`,
 				Advanced: true,
 				Default:  encoder.Display | encoder.EncodeInvalidUtf8 | encoder.EncodeBackSlash,
 			},
-		}})
+			{
```
ncw: Rclone has lots of backends which support About but don't know the total size.

There are various mechanisms for dealing with this

      --vfs-disk-space-total-size SizeSuffix   Specify the total space of disk (default off)

So I don't think we need this here.

Can you remove the `total_size` and just report unknown.

ncw (commented):
This looks useful thank you :-)

See comments inline.

backend/ulozto/ulozto.go
metlos: Makes sense, removed in https://github.com/rclone/rclone/pull/8773/commits/4e7cc1dc07104325f96dbcecef38bfa8af1fcce3. 

I used the config option to be able to specify the size when using the remote as part of a union. Is there another way of specifying the size of a remote so that the size is defined for a union that the remote would be part of? As far as I understood, mount-time options cannot be used in the list of remotes of a union.

PR #8778, opened 2025-08-26
Restrict the markdown languages to use for code blocks

albertony:
#### Checklist


ncw:
This looks sensible to me :-)

Note 85deea82e4acc5e09f85a41be3f60355f0f9be2d which might make your life easier!

ncw (approved):
This is great, thank you.

Please merge.

ncw:
I was writing docs for a command recently and I came across the old annoyance of not being able to have back quotes in back quoted strings. We use strings.ReplaceAll for this normally to swap | for \` but it made me wonder if we should have an official way to do this. (I also needed to make a markdown table with | which was annoying!)

Perhaps it could be run on all the command (and RC) help automatically (so no need to wrap it) and there could be one version for the html docs and one for the cli where it expands out the `[link](/whatever)` to it's full html so it works better in the terminal.

Just musing anyway...

albertony:
> This is great, thank you.
> 
> Please merge.

Thanks. I just found one small thing I'm looking into...

In Hugo, an indented code block, and fenced code block without a language, or an unrecognized language, is rendered differently from one with language identifier "text", "plain" or any other supported languages I've found. Any supported languages will be rendered with the light text on black background, while when no language is specified there will be dark gray text on light gray background.

image

This in contrast to GitHub, where code blocks without language looks identical to code blocks with "text" language specified (to my eyes, at least). If we want to allow this default/none-styled code blocks, then I can add one more allowed language identifier which is one not recognized by Hugo (or others to avoid confusion). I'm come up with "none" so far.

But on the other hand, the code blocks with black background, maybe they stand out a bit too much? Especially for large sections like the config example, e.g. in https://rclone.org/compress/. For small command examples I like the look, it makes them stand out like a small console window. But with these large blocks maybe it is a bit too much..?

Hugo supports different styles for the code blocks: https://gohugo.io/quick-reference/syntax-highlighting-styles. The default is "monokai", and seem to be a dark mode style. Our web side is more a light theme, so the "monokailight" style could be better:

image

And if we did switch to something like this, then the "none" language mentioned above is not so important, as now "text" does in fact look more like the default (not identical, it is a bit lighter).

There are also styles "github" and "github-dark", if we wanted colors more close to what GitHub uses, but I haven't looked too much into them, or others.

ncw:
I think changing the default style sounds like the best plan so we don't have to have the "none" language.

albertony:
Agreed. I extracted my above description into a separate PR that only changes the style, to "monokailight":
- https://github.com/rclone/rclone/pull/8935

PR #8783, opened 2025-08-28
Add some more details about supported regex syntax

albertony:

The name transform syntax description says `regex=/pattern/replacement/`, but the value is supposed to be a plain string "pattern/replacement" and not the typical regex literal syntax with enclosed slashes, so I think it is less confusing to new users without the outer slashes.

Also added a short text about the regular expression syntax, and mentioning that capturing group variables can be used in the replacement.

Based on:
- https://pkg.go.dev/regexp#pkg-overview
- https://pkg.go.dev/regexp#Regexp.Expand
- https://pkg.go.dev/regexp/syntax

#### Checklist


nielash (approved):
Good catch, thank you 🙂

ncw:
Thank you @albertony :-)

Issue #7210, opened 2023-08-04
chunker overlay's reporting of transfer information is not super useful always, and chunker interact with --transfers in a funky way


left1000:
![chunkercouldbemorreliable](https://github.com/mikf/gallery-dl/assets/16432525/e074e86c-ffc6-40af-9728-5e7bddf006c1)

As you can see here the chunker overlay uploads chunkers then renames them via a server-side move which doubles the reported bandwidth and or filesize transferred and makes bandwidth reporting inaccurate. Now I could do some math to determine the bandwidth usage myself, BUT, if the filesize transfered data measurement is wrong, my math will result in a wrong average bandwidth used result also.

Also maybe even more critically... I find that the chunker overlay almost randomly multiplies the --transfers flag.

I think the correct function would be for a manually entered --transfers flag to overwrite some of chunker's choices. 
If I set --transfers 8 and then upload 8 files that are broken into 16 total ChunkerOverlayFileChunks; it will make rclone function as if I had set -transfers 16... And if I'm trying to carefully calibrate my rclone copy/sync command in order to properly fit inside some draconian api call restriction throttling.. I'd rather in the above example for rclone to treat --transfers 8 like --transfers 4. But I can't just type --transfers 4 because if I'm uploading even larger files maybe the 8 files are broken into 80 ChunkerOverlayFileChunks and even if I entered --transfers 4 it would behave as if I'd entered --transfers 40
So ideally when a chunker overlay is told to upload 8 files at once, that should be 8 ChunkerOverlayFileChunks. AKA, --transfers 8 always means to transfer 8 source files at once. It also usually means to transfers 8 destination files at once. But in a chunker overlay those two 8's are not equal. --transfers 8 in rclone needs to always transfer 8 destination files. Which is almost always 8 source files. But inside a chunkeroverlay, 8 destination files could be anywhere from 1-8 source files. Does that make any sense to anyone but me?

TLDR: the chunker overlay is harder to control, and harder to monitor than other overlays. I think if some genius reads this message, they could think of a way to improve it :)

kapitainsky:
Transfers reporting is already fixed and will be included in v1.64. If you want you can try it now. Download fixed version from here:

https://beta.rclone.org/branch/fix-7183-server-side-accounting/v1.64.0-beta.7191.b3c480be2.fix-7183-server-side-accounting/

left1000:
[...]

PR #8791, opened 2025-09-01
bisync: ignore expected "nothing to transfer" differences on tests

nielash:

Fixes for bisync integration tests broken by https://github.com/rclone/rclone/commit/9e200531b1490000656031c42fff4d95477e1b46.

- https://github.com/rclone/rclone/pull/8770#issuecomment-3241792807

#### Checklist

### DIFF
--- a/cmd/bisync/bisync_test.go
+++ b/cmd/bisync/bisync_test.go
@@ -177,6 +177,7 @@ var (
 	// "src and dst identical but can't set mod time without deleting and re-uploading"
 	argRefreshTimes = flag.Bool("refresh-times", false, "Force refreshing the target modtime, useful for Dropbox (default: false)")
 	ignoreLogs      = flag.Bool("ignore-logs", false, "skip comparing log lines but still compare listings")
+	argPCount       = flag.Int("pcount", 2, "number of parallel subtests to run for TestBisyncConcurrent") // go test ./cmd/bisync -race -pcount 10
 )
 
 // bisyncTest keeps all test data in a single place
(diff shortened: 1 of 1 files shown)


ncw (approved):
This looks great to me - please merge when happy :-)

Thank you.

nielash:
@ncw looks like that fixed it, with two exceptions:

- `koofr`: very likely due to its `src and dst identical but can't set mod time without deleting and re-uploading` behavior, but would need a koofr account to verify. I can put a fix in for that, and maybe you can test it?
- `chunker`: very likely due to #7210 (inflated transfer count due to internal chunk transfers). Ideally that would be fixed in the backend... but I suppose we could also single out `chunker` by name on the test to ignore this log line. Thoughts? (For that matter, we could also just ignore it on all non-local... although that might be a bit heavy-handed.)

Technically `onedrive` also failed, although it does not look bisync-related (`oauth2: "invalid_grant"`)

And interestingly, `gofile` actually passed today! 🤔

ncw:
@nielash wrote:
> looks like that fixed it, with two exceptions:

Great :-)

> * `koofr`: very likely due to its `src and dst identical but can't set mod time without deleting and re-uploading` behavior, but would need a koofr account to verify. I can put a fix in for that, and maybe you can test it?

I'll email you some test credentials.

> * `chunker`: very likely due to [chunker overlay's reporting of transfer information is not super useful always, and chunker interaction with --transfers in a funky way #7210](https://github.com/rclone/rclone/issues/7210) (inflated transfer count due to internal chunk transfers). Ideally that would be fixed in the backend... 

Hmm, the transfers are real transfers so the user might wish to see them but that issue is all about the user being confused by them which is another valid option! I wonder if we could mask them by having a special chunker stats group to do the sub-transfers with... Perhaps getting a bit overcomplicated.

> but I suppose we could also single out `chunker` by name on the test to ignore this log line. Thoughts? (For that matter, we could also just ignore it on all non-local... although that might be a bit heavy-handed.)

I'd probably do the simplest thing and ignore it for `chunker`. There are already quite a few things like that in the bisync tests aren't there?

> Technically `onedrive` also failed, although it does not look bisync-related (`oauth2: "invalid_grant"`)

The credentials had expired. Likely because I put 2FA on the account to try to make a "Personal Vault" appear to investigate another issue! I've refreshed them now.

> And interestingly, `gofile` actually passed today! 🤔

I think it has quite heavy rate limiting.

PR #8794, opened 2025-09-02
test speed: adds command to test a specified remotes speed

### LINKED ISSUE #3198: Request: Speedtest
Would be really happy to have some kind of "rclone speedtest gdrive" option.

(Today i can't use Plex .. don't know why. Can't find the bottleneck .. Seems to be GoogleDrive, since my server has zero CPU and 1GB connection (no drop down). Could give useful insights).

roucc:
Run speed test to try and work in a given time budget, uploading randomly created files to the remote then downloading them again.

Fixes #3198

#### Checklist

### DIFF
--- a/cmd/test/makefiles/speed.go
+++ b/cmd/test/makefiles/speed.go
@@ -42,6 +42,13 @@ func init() {
 	flags.FVarP(speedFlags, &large, "large", "", "Size of large files", "")
 	flags.BoolVarP(speedFlags, &useJSON, "json", "", useJSON, "Output only results in JSON format", "")
 
+	addCommonFlags(speedFlags)
+}
+
+func logf(text string, args ...any) {
+	if !useJSON {
+		fmt.Printf(text, args...)
+	}
 }
 
 var speedCmd = &cobra.Command{
@@ -57,7 +64,9 @@ var speedCmd = &cobra.Command{
 	uses the results of an initial test to determine how many files to use in
 	each subsequent test.
 
-	It is recommended to use -q flag for a simpler output.
+	It is recommended to use -q flag for a simpler output. e.g.:
+	
+	    rlone test speed remote: -q
 
 	**NB** This command will create and delete files on the remote in a randomly
 	named directory which should be tidied up after.
@@ -73,9 +82,7 @@ var speedCmd = &cobra.Command{
 
 		// initial test
 		size := fs.SizeSuffix(1024 * 1024)
-		if !useJSON {
-			fmt.Printf("Running initial test for 4 files of size %v\n", size)
-		}
+		logf("Running initial test for 4 files of size %v\n", size)
 		stats, err := speedTest(ctx, 4, size, args[0])
 		if err != nil {
 			return fmt.Errorf("speed test failed: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+	test.Command.AddCommand(speedCmd)
+
+	speedFlags := speedCmd.Flags()
+	flags.FVarP(speedFlags, &testTime, "test-time", "", "Length for each test to run", "")
+	flags.IntVarP(speedFlags, &fcap, "file-cap", "", fcap, "Maximum number of files to use in each test", "")
+	flags.FVarP(speedFlags, &small, "small", "", "Size of small files", "")
+	flags.FVarP(speedFlags, &medium, "medium", "", "Size of medium files", "")
+	flags.FVarP(speedFlags, &large, "large", "", "Size of large files", "")
+	flags.BoolVarP(speedFlags, &useJSON, "json", "", useJSON, "Output only results in JSON format", "")
+
```
ncw, line 44: Can you factor this bit of code into a function - say `addCommonFlags`

https://github.com/rclone/rclone/blob/0019e18ac3a673714e851b2a97d63a8ad0a77654/cmd/test/makefiles/makefiles.go#L64-L72

Then call it here `addCommonFlags(speedFlags)` so this command has the flags for controlling the contents of the test files too.

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+	This command runs a series of uploads and downloads to the remote, measuring
+	and printing the speed of each test using varying file sizes and numbers of
+	files.
+
+	Test time can be innaccurate with small file caps and large files. As it
+	uses the results of an initial test to determine how many files to use in
+	each subsequent test.
+
+	It is recommended to use -q flag for a simpler output.
```
ncw: Could you put an example of how to run the test here please?

Eg

```sh
rclone test speed remote: -q
```

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+		stats, err := speedTest(ctx, 4, size, args[0])
+		if err != nil {
+			return fmt.Errorf("speed test failed: %w", err)
+		}
+
+		var results []*Stats
+
+		// main tests
+		for _, size := range []fs.SizeSuffix{small, medium, large} {
+			combinedSpeed := stats.Upload.Speed + stats.Download.Speed
```
ncw: I don't think this calculation is quite right... Maybe this should be the average speed if the intention is that the total test for upload and download should hit the `testTime`?

It might be best to do the upload and download calculations separately as they are likely to be quite different.

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+			numberOfFiles := int((float64(combinedSpeed) * time.Duration(testTime).Seconds()) / (2 * float64(size)))
+			if !useJSON {
+				fmt.Println()
```
ncw: This code is essentially duplicated in the else of this if statement.

I think you are doing this to avoid the print statements in the `--json` case.

How about making your own `logf` function which doesn't print if `--json` is set - then you could combine these two branches and tidy up the other if jsonLOG statements

Something like this (untested)

```go
func logf(text string, args ...any) {
    if !jsonLOG {
        fmt.Printf(text, args...)
    }
}
```

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+	fremote := cmd.NewFsDir([]string{tempDirPath})
+	aErr := io.EOF
+	defer atexit.OnError(&aErr, func() {
+		err := operations.Purge(ctx, fremote, "")
+		if err != nil {
+			fs.Debugf(fremote, "Failed to remove temp dir %q: %v", tempDirPath, err)
+		}
+	})()
+
+	flocalDir, err := os.MkdirTemp("", "rclone-makefiles-local-")
```
ncw: This should probably say "rclone-speedtest-local-"

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil, fmt.Errorf("failed to create local temp dir: %w", err)
+	}
+	defer atexit.OnError(&aErr, func() { _ = os.RemoveAll(flocalDir) })()
+
+	flocal, err := cache.Get(ctx, flocalDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create local fs: %w", err)
+	}
+
+	fdownloadDir, err := os.MkdirTemp("", "rclone-makefiles-download-")
```
ncw: s/makefiles/speedtest/ here too

cmd/test/makefiles/speed.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	err = operations.CheckDownload(ctx, &opt)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check redownloaded files were identical: %w", err)
+	}
+
+	return &stats, nil
+}
+
+// docs
```
ncw: spare comment?

ncw (commented):
Thank you for doing this @roucc - it looks really good.

I've put some suggestions for improvements inline.

Thank you :-)

Issue #8458, opened 2025-03-19
segfaults frequently with rclone bisync


sigmaSd:
#### What is the problem you are having with rclone?

segfaults when running rclone bisync , not always but frequently

#### What is your rclone version (output from `rclone version`)

rclone v1.68.2
- os/version: fedora 41 (64 bit)
- os/kernel: 6.12.11-200.fc41.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.23.4
- go/linking: dynamic
- go/tags: none

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)

linux 64bits (fedora 41)

#### Which cloud storage system are you using? (e.g. Google Drive)

google drive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)

rclone bisync  ~/Notes gdrive:notes_vault --exclude ".obsidian/" --exclude=".git/" --progress --color never

#### A log from the command

<details>
 *                  Notes/TODO/Main.md.conflict1:100% /463, 462/s, 0sTransferred:           1.797 KiB / 1.797 KiB, 100%, 153 B/s, ETA 0s
Checks:                18 / 18, 100%
Renamed:                2
Transferred:            4 / 4, 100%
Server Side Moves:      2 @ 920 B
Elapsed time:         8.8spanic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x55824f202280]

goroutine 1 gp=0xc0000061c0 m=7 mp=0xc000100708 [running]:
panic({0x5582507ba2e0?, 0x558251c75f00?})
        /usr/lib/golang/src/runtime/panic.go:804 +0x168 fp=0xc000b94948 sp=0xc000b94898 pc=0x55824dc47888
runtime.panicmem(...)
        /usr/lib/golang/src/runtime/panic.go:262
runtime.sigpanic()
        /usr/lib/golang/src/runtime/signal_unix.go:917 +0x359 fp=0xc000b949a8 sp=0xc000b94948 pc=0x55824dc4a939
github.com/rclone/rclone/cmd/bisync.(*bisyncRun).modifyListing(0xc0007c4000, {0x558250bec590, 0x558251dfa600}, {0x558250bfff10, 0xc00084e008}, {0x558250c00ed0, 0xc0007075e0}, {0xc000476808, 0x6, 0x9}, ...)
[...]

PR #8803, opened 2025-09-06
bisync: fix error handling for renamed conflicts

nielash:

Before this change, rclone could crash during `modifyListing` if a rename's `srcNewName` is known but not found in the `srcList` (`srcNewName != "" && new == nil`). This scenario should not happen, but if it does, we should print an error instead of crashing.

On #8458 there is a report of this possibly happening on `v1.68.2`. It is unknown what the underlying issue was, and whether it still exists in the latest version, but if it does, the user will now see an error and debug info instead of a crash.

- #8458

#### Checklist


ncw (approved):
LGTM - thank you

PR #8850, opened 2025-09-24
Correcting azure file copy code to wait till copy completes before re…

### LINKED ISSUE #8848: [Azure Files] Copy() wrapper not polling for copy to succeed/fail before returning
I am a developer from Azure Files. We are investigating an issue reported by a customer which seems to show that the client is using rclone and is doing a copyfile from file A to file B followed by a renamefile from file C to file B.

It seems that in Azure Files copy() wrapper routine as referred here : https://github.com/rclone/rclone/blob/master/backend/azurefiles/azurefiles.go#L1316 does not poll for the completion of the copy before returning.
In similar code for Azure Blob, for the copy() wrapper routine, https://github.com/rclone/rclone/blob/master/backend/azureblob/azureblob.go#L1987
we can see that the rclone client is polling whether copyStatus is CopyStatusTypePending or not, as shown below, which is missing for Azure FIles.

	// Poll for completion if necessary
	//
	// The for loop is never executed for same storage account copies.
	copyStatus := startCopy.CopyStatus
	getOptions := blob.GetPropertiesOptions{}
	pollTime := 100 * time.Millisecond
	for copyStatus != nil && string(*copyStatus) == string(container.CopyStatusTypePending) {
		time.Sleep(pollTime)
		var getMetadata blob.GetPropertiesResponse
		err = f.pacer.Call(func() (bool, error) {
			getMetadata, err = dstBlobSVC.GetProperties(ctx, &getOptions)
			return f.shouldRetry(ctx, err)
		})
		if err != nil {
			return nil, err
		}
		copyStatus = getMetadata.CopyStatus
		pollTime = min(2*pollTime, time.Second)
	}

Due to this bug on the client, we can run into an edge case which can make the file inaccessible for the client. 

rclone version shows up on the server as v1.71.0
This seems to be a day 0 issue from when the azure files support was added. i.e v1.65.0 from 2023-11-26 onwards

At this point, it is not clear from how the customer is using rclone

Azure Files

From the signature that I see on the server, the customer is doing a copyfile followed by renamefile using REST.

I do not have that information.

#### How to use GitHub

vibhansa-msft:
Fixes #8848 

yes.

#### Checklist

### DIFF
--- a/backend/azurefiles/azurefiles.go
+++ b/backend/azurefiles/azurefiles.go
@@ -1313,10 +1313,29 @@ func (f *Fs) Copy(ctx context.Context, src fs.Object, remote string) (fs.Object,
 	}
 	srcURL := srcObj.fileClient().URL()
 	fc := f.fileClient(remote)
-	_, err = fc.StartCopyFromURL(ctx, srcURL, &opt)
+	startCopy, err := fc.StartCopyFromURL(ctx, srcURL, &opt)
 	if err != nil {
 		return nil, fmt.Errorf("Copy failed: %w", err)
 	}
+
+	// Poll for completion if necessary
+	//
+	// The for loop is never executed for same storage account copies.
+	copyStatus := startCopy.CopyStatus
+	var properties file.GetPropertiesResponse
+	pollTime := 100 * time.Millisecond
+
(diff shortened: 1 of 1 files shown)


bhansalivikas:
@ncw  Can you please look at this PR. This bug is leading to data corruption issue when someone tries to copy data from one Azure Files account to another.

PR #8855, opened 2025-09-25
Jottacloud whitelabel authentication improvements

### LINKED ISSUE #8852: rclone stopped working with Jottacloud legacy login
https://forum.rclone.org/t/elkjop-cloud-config/52526

I have chosen jottacloud and it fails here

Option config_type.
Select authentication type.
Choose a number from below, or type in an existing value of type string.
Press Enter for the default (standard).
/ Standard authentication.
1 | Use this if you're a normal Jottacloud user.
\ (standard)
/ Legacy authentication.
2 | This is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.
\ (legacy)
/ Telia Cloud authentication.
3 | Use this if you are using Telia Cloud (Sweden).
\ (telia_se)
/ Telia Sky authentication.
4 | Use this if you are using Telia Sky (Norway).
\ (telia_no)
/ Tele2 Cloud authentication.
5 | Use this if you are using Tele2 Cloud.
\ (tele2)
/ Onlime Cloud authentication.
6 | Use this if you are using Onlime Cloud.
\ (onlime)
config_type> 2

Do you want to create a machine specific API key?

Rclone has it's own Jottacloud API KEY which works fine as long as one
only uses rclone on a single machine. When you want to use rclone with
this account on more than one machine it's recommended to create a
machine specific API key. These keys can NOT be shared between
machines.
y) Yes
n) No (default)
y/n> y

Error: failed to register device: Post "https://api.jottacloud.com/auth/v1/register?device_id=rclone-44vt04llECptGXr3FwurU": read tcp 192.168.1.195:52037->185.179.129.36:443: wsarecv: En eksisterende forbindelse blev tvangsafbrudt af en ekstern vært.

If i choose no instead of yes I get:

Option config_password.
Password (only used in setup, will not be stored)
Choose an alternative below. Press Enter for the default (n).
y) Yes, type in my own password
g) Generate random password
n) No, leave this optional password blank (default)
y/g/n> y
Enter the password:
password:
Confirm the password:
password:

2025/09/15 17:12:53 CRITICAL: Reveal failed: input too short when revealing password - is it obscured?

Which cloud storage system are you using? (eg Google Drive)
Elkjøp cloud

rclone v1.71.0

Linux and Linux Docker (binhex-rclone)

Jottacloud White Label Products like Elkjøp cloud, LetsGoCloud, etc. with lagacy login

rclone config

see above

albertony:

Elkjøp with all its subsidiaries have finally migrated their whitelabel services away from the legacy authentication method, and seem to be using OAuth 2.0 OpenAPI Connect in the same variant as Telia and Tele2 currently does.

This includes:
- Elkjøp Cloud (cloud.elkjop.no)
- Elgiganten Sweden (cloud.elgiganten.se)
- Elgiganten Denmark (cloud.elgiganten.dk)
- Giganti Cloud (cloud.gigantti.fi)
- ELKO Cloud (cloud.elko.is)

I also added a new whitelabel provider from the company MediaMarkt, called Let's Go Cloud (letsgo.jotta.cloud).

Adding all of these new alternatives to the config process in the same manner as Tele2 and Telia was implemented, would lead to a lot of clutter, both in code and in the configuration UX, in my opinion. I therefore decided to do some refactoring of the code, and restructuring of the configuration flow. This includes no longer hardcoding two complete URLs, representing the Authorization and Token endpoints, for each of these services. These are very similar for all of the whitelabel services, the domain name is the main difference. But also, I decided to not only centralize the generation of these URLs, but also to use the OpenID provider configuration endpoint to retrieve them. This should be at a well-known URL ending on `/.well-known/openid-configuration`, according to the OpenID standard. We already use such an endpoint for the standard authentication, although there the well-known URL are parsed from the login token.

The previous configuration flow included the following part:

```
Option config_type.
Select authentication type.
Choose a number from below, or type in an existing string value.
Press Enter for the default (standard).
   / Standard authentication.
 1 | Use this if you're a normal Jottacloud user.
   \ (standard)
   / Legacy authentication.
 2 | This is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.
   \ (legacy)
   / Telia Cloud authentication.
 3 | Use this if you are using Telia Cloud.
   \ (telia)
   / Tele2 Cloud authentication.
 4 | Use this if you are using Tele2 Cloud.
   \ (tele2)
   / Onlime Cloud authentication.
 5 | Use this if you are using Onlime Cloud.
   \ (onlime)
```

This has now been replaced by the following, split into 2 steps:
```
Option config_type.
Select authentication type.
Choose a number from below, or type in an existing value of type string.
Press Enter for the default (standard).
   / Standard authentication.
 1 | Use this if you're a normal Jottacloud user.
   \ (standard)
   / Whitelabel authentication.
 2 | Use this if you are using the service offered by a third party such as Telia, Tele2, Onlime, Elkjøp, etc.
   \ (whitelabel)
   / Legacy authentication.
 3 | This is only required for certain whitelabel versions of Jottacloud and not recommended for normal users.
   \ (legacy)
config_type> 2

Option whitelabel_type.
Select whitelabel type.
Choose a number from below, or type in an existing value of type string.
Press Enter for the default (telia_se).
   / Telia Cloud authentication.
 1 | Use this if you are using Telia Cloud (Sweden).
   \ (telia_se)
   / Telia Sky authentication.
 2 | Use this if you are using Telia Sky (Norway).
   \ (telia_no)
   / Tele2 Cloud authentication.
 3 | Use this if you are using Tele2 Cloud.
   \ (tele2)
   / Onlime Cloud authentication.
 4 | Use this if you are using Onlime Cloud.
   \ (onlime)
   / Elkjøp Cloud authentication.
 5 | Use this if you are using Elkjøp Cloud (Norway).
   \ (elkjop)
   / Elgiganten Cloud authentication.
 6 | Use this if you are using Elgiganten Cloud (Sweden).
   \ (elgiganten_se)
   / Elgiganten Cloud authentication.
 7 | Use this if you are using Elgiganten Cloud (Denmark).
   \ (elgiganten_dk)
   / Giganti Cloud authentication.
 8 | Use this if you are using Giganti Cloud.
   \ (gigantti)
   / ELKO Cloud authentication.
 9 | Use this if you are using ELKO Cloud.
   \ (elko)
   / Let's Go Cloud authentication.
10 | Use this if you are using Let's Go Cloud.
   \ (letsgo)
whitelabel_type>
```
This PR also includes the fix suggested in https://github.com/rclone/rclone/issues/8852#issuecomment-3333333963, for the `Reveal failed: input too short when revealing password - is it obscured` bug when using legacy authentication and answering `y` to the question `Do you want to create a machine specific API key?`.

Fixes #8852

https://forum.rclone.org/t/elkjop-cloud-config/52526

#### Checklist


backend/jottacloud/jottacloud.go
```diff
@@ -61,27 +61,25 @@ const (
 	defaultTokenURL = "https://id.jottacloud.com/auth/realms/jottacloud/protocol/openid-connect/token"
 	defaultClientID = "jottacli"
 
+	whitelabelClientID = "desktop"
+
+	whitelabelTeliaSeDomain = "cloud-auth.telia.se"
```
ncw: These could probably be a `map[tag]struct{ domain, realm string}` couldn't they?

That would simplify the code a bit below.

You could also build the `[]fs.OptionExample` passed to the `fs.ConfigChooseExclusiveFixed("whitelabel_type_done", ...` function programatically from this struct. Might need to have a name in for that too.

ncw (commented):
Thanks for doing this - great to sort out the mess :-)

I put an idea for sorting it out a bit more inline.

maschhoff:
Great Job! 
Testet on 1.72. beta jotta cloud fix with LetsGoCloud and get if its opens the browser:

invalid_client - Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The requested OAuth 2.0 Client does not exist.

It seems that the authentication is different:
Here an example:

REQUEST:
https://letsgo.jotta.cloud/auth/realms/letsgo/oauth2/auth?client_id=desktop-win&code_challenge=yZTVPLnGfdduWhRpYXNcQEqoeJyckkndRtZbVRhpucEgLkGHKFkgIzOlVfuMsbWt&code_challenge_method=plain&redirect_uri=http%3A%2F%2Flocalhost%3A53682&response_type=code&scope=offline_access%20offline%20openid&state=eftijxyfFIMsAoOFRPEvkOndQLnGxCmH

RESPONSE:
[http://localhost:53682/?code=ory_ac_9hQnZWy91XQ54ayKNJSMlGqBcTJzzA6ECcUKy1PxNeE.V5L25pZMBIQkrzdh3LtHf4BZYd9rBa8byf8aEt8WDBg&scope=openid+offline_access&state=eftijxyfFIMsAoOFRPEvkOndQLnGxCmH]

seems code_challenge and method can be left in the request its plain PKCE
The response code look wirred

backend/jottacloud/jottacloud.go
```diff
@@ -66,24 +67,32 @@ const (
...
-	onlimeCloudTokenURL = "https://cloud-auth.onlime.dk/auth/realms/onlime_wl/protocol/openid-connect/token"
-	onlimeCloudAuthURL  = "https://cloud-auth.onlime.dk/auth/realms/onlime_wl/protocol/openid-connect/auth"
-	onlimeCloudClientID = "desktop"
 )
 
+func getWhitelabelServices() map[string]struct {
+	name, domain, realm, clientID string
+	scopes                        []string
+} {
+	return map[string]struct {
```
ncw, line 76: I'd probably just make this a global variable rather than a function to return it.

ncw (approved):

Happy for you to merge it when you and the testers are happy.

backend/jottacloud/jottacloud.go
albertony: My reasoning was that this is only used during config, so it felt unnecessary to generate the map and keep it in memory at all times. But maybe this is not correct or a relevant consideration?

I see that similar data is global variable in s3, but there it is also referenced from init (simply to generate a string to use in the fs.RegInfo.Description), so then I guess the difference is somewhat smaller.

That being said, I'm considering generating the fs description in Jottacloud as well (in a future pr), so that changes the assumption a bit as well...

ncw: I hadn't thought of it being generated Vs being static and in memory all the time, good point . The function may use more code, I don't know. You do whichever was you feel the most comfortable with 🙂

PR #8865, opened 2025-09-30
swift: If storage_policy isn't set, use the root containers policy

### LINKED ISSUE #8858: If a segments container is created for Swift, it uses the default (or rclone configured) storage policy

N/A

1.71.1

I'd expect that if rclone needs to create a segments container it would use the same storage policy of the base container. It doesn't, it uses either the policy configured for the remote in rclone.conf (etc) or the default policy configured for Swift.

X-Storage-Policy for the base container is fetched, and is available to rclone.

How this exhibits is our customers may naively create a multi region replicated container for the segments when they intended it to be single region - as used for the base container.

I just ran into this while testing some other behaviour.

If storage_policy isn't set locally, then X-Storage-Policy from the base container should be used when creating the segments container.

#### How to use GitHub

puck:
Ensure that if we need to create a segments container it uses the same storage policy as the root container.

Fixes #8858

#### Checklist

### DIFF
--- a/backend/swift/swift.go
+++ b/backend/swift/swift.go
@@ -561,6 +561,21 @@ func (f *Fs) setRoot(root string) {
 	f.rootContainer, f.rootDirectory = bucket.Split(f.root)
 }
 
+// Fetch the base container's policy to be used if/when we need to create a
+// segments container to ensure we use the same policy.
+func (f *Fs) fetchStoragePolicy(ctx context.Context, container string) (fs.Fs, error) {
+	err := f.pacer.Call(func() (bool, error) {
+		var rxHeaders swift.Headers
+		_, rxHeaders, err := f.c.Container(ctx, container)
+
+		f.opt.StoragePolicy = rxHeaders["X-Storage-Policy"]
+		fs.Debugf(f, "Auto set StoragePolicy to %s", f.opt.StoragePolicy)
+
+		return shouldRetryHeaders(ctx, rxHeaders, err)
+	})
+	return nil, err
+}
+
 // NewFsWithConnection constructs an Fs from the path, container:path
 // and authenticated connection.
 //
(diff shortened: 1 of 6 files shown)


puck:
I've fixed the lint issue.

backend/swift/swift.go
```diff
@@ -1132,6 +1133,24 @@ func (f *Fs) newSegmentedUpload(ctx context.Context, dstContainer string, dstPat
...
+		if f.opt.StoragePolicy == "" {
+			err = f.pacer.Call(func() (bool, error) {
+				var rxHeaders swift.Headers
+				_, rxHeaders, err := f.c.Container(ctx, f.rootContainer)
+
+				f.opt.StoragePolicy = rxHeaders["X-Storage-Policy"]
+				fs.Debugf(f, "Auto set StoragePolicy to %s", f.opt.StoragePolicy)
+
+				return shouldRetryHeaders(ctx, rxHeaders, err)
+			})
```
ncw: I think it would be worth factoring this out into a function `setStoragePolicyIfUnset` say so you can call it explicitly in the tests.

backend/swift/swift_test.go
```diff
@@ -195,4 +196,9 @@ func (f *Fs) testCopyLargeObject(t *testing.T) {
 	require.Equal(t, obj.Size(), objTarget.Size())
 }
 
+func (f *Fs) testPolicyDiscovery(t *testing.T) {
+	// Default policy for SwiftAIO is Policy-0.
+	assert.Equal(t, "Policy-0", f.opt.StoragePolicy)
```
ncw: This test, as written, relies on the other tests being run first. This doesn't always happen especially in the integration tester. So I'd factor the setting out and call it explicitly (see above).

This could fail in the integration tests if we had any swift storage backends that weren't SAIO but we don't currently!

puck: Ah, I wasn't aware of the detail about tests being able to run in different orders. I had considered including it one of the existing tests, but that felt dirty.

I'll split out the code above into a different function and test that. I'll have a go tonight.

ncw: Ping me on the issue when you want me to take another look - thanks

backend/swift/swift.go
puck: Done.

backend/swift/swift_test.go
puck: Done.

PR #8866, opened 2025-09-30
s3: support providing `x-amz-object-lock-*` headers in `--header-upload`

wouterj:

This adds basic support for the features suggested by #4683. Instead of adding the CLI options, this change at least allows passing the required headers through `--header-upload`. We cannot configure lock settings globally on the bucket, so would need to rely on a tool like rclone to provide this.

I couldn't find any test for `prepareUpload()` and I don't have a lot of Go experience. Do you want any test to cover these new options and would it be possible to get some guidance on how and where to add these if necessary? :)

https://github.com/rclone/rclone/issues/4683#issuecomment-1011109273

#### Checklist


ncw:
@wouterj we have now merged proper cli object lock support in 7d0a8bf850b93d3f70a50f47eab489295e6058d9

We can merge this too as I don't think it will interfere - what do you think?

PR #8869, opened 2025-10-02
Remote setup doc improvements

albertony:

I found the https://rclone.org/remote_setup page a bit outdated, so I updated it, as well as some small updates to the related https://rclone.org/commands/rclone_authorize/ page and others.

PS: The https://rclone.org/remote_setup page is not listed in the "Docs" drop down menu, it is only linked to from the FAQ page and some backend docs pages. Should we list it?

#### Checklist

### DIFF
--- a/docs/content/remote_setup.md
+++ b/docs/content/remote_setup.md
@@ -6,7 +6,7 @@ description: "Configuring rclone up on a remote / headless machine"
 # Configuring rclone on a remote / headless machine
 
 Some of the configurations (those involving oauth2) require an
-Internet connected web browser.
+internet-connected web browser.
 
 If you are trying to set rclone up on a remote or headless box with no
 browser available on it (e.g. a NAS or a server in a datacenter) then
--- a/docs/content/remote_setup.md
+++ b/docs/content/remote_setup.md
@@ -8,20 +8,21 @@ description: "Configuring rclone up on a remote / headless machine"
 Some of the configurations (those involving oauth2) require an
 internet-connected web browser.
 
-If you are trying to set rclone up on a remote or headless box with no
-browser available on it (e.g. a NAS or a server in a datacenter) then
-you will need to use an alternative means of configuration.  There are
-two ways of doing it, described below.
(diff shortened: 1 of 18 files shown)


docs/content/remote_setup.md
```diff
@@ -70,50 +73,57 @@ y/e/d>
...
 Configuration file is stored at:
 /home/user/.rclone.conf
 ```
 
-Now transfer it to the remote box (scp, cut paste, ftp, sftp, etc.) and
-place it in the correct place (use `rclone config file` on the remote
-box to find out where).
+Finally, transfer the file to the remote machine (scp, cut paste, ftp, sftp, etc.)
+and place it in the correct location (use [rclone config file](/commands/rclone_config_file)
+on the remote box to find out where).
```
ncw: You replaced `box` with `machine` earlier which seems like a good change :-)

ncw (approved):
Very nice changes - thank you.

I spotted one missed box -> machine change.

> PS: The https://rclone.org/remote_setup page is not listed in the "Docs" drop down menu, it is only linked to from the FAQ page and some backend docs pages. Should we list it?

Yes I think we probably should.

albertony:
Thank you.

> > PS: The https://rclone.org/remote_setup page is not listed in the "Docs" drop down menu, it is only linked to from the FAQ page and some backend docs pages. Should we list it?
> 
> Yes I think we probably should.

I think so too. :) Although to some degree I felt it is a bit too niche thing to defend such a "top level status", compared to more important topics like "Installation", "Usage" and "Changelog". But on the other hand, so is perhaps "Bugs" and "FAQ". And on yet another hand (third..?) it is perhaps more the case that "Usage" should be split up, and perhaps a new page "Configuring" should be created... Oh well.. 😅 

I've added a commit that adds it to the navbar.html. I chose to name it "Remote Setup" and placed it *after* "Remote Control", but feel free to suggest better alternatives..

ncw (approved):
"Remote Setup" looks great :-)

PR #8874, opened 2025-10-03
vfs: fix chunker integration test

roucc:
Chunker failing in integration tests for the past few days after my zip file commit.

chunker50b remote causing 5Mib / 50b files needed to be made.


ncw (approved):
Looks great - thank you. Please merge when the tests pass.

PR #8878, opened 2025-10-04
Jottacloud authentication improvements

albertony:

Continuing a bit from:
- https://github.com/rclone/rclone/pull/8855

Before this change, Jottacloud (jottacloud.com) could only be used with the special "personal login token" flow (which they use for their official CLI), while white-label services used a more traditional oauth flow with credentials through web browser etc. But Jottacloud can also be used with the same traditional oauth flow. Some may prefer this, as they are unfamiliar with the "personal login token" approach, and also it can be a bit cumbersome with the manual "side-flow", where they have to enter their password twice etc etc.

Here I add Jottacloud as an option to the "white-label service" list.. But then the name is misleading, so I renamed that.. I named this authentication option "traditional", referring to the traditional oauth flow, in lack of a better word, so now we have for Jottacloud: standard (with login token), traditional (traditional oauth flow with web browser), and legacy (with password).

I think it is natural to have Jottacloud as the first, and default, entry of the list of services when selecting this "traditional" option. Therefore I refactored the (whitelabel) service configurations from map to slice.

I updated documentation accordingly, and this includes some improvements to the existing in that the notes regarding stale token when reusing config, and "My logged in devices", previously were written specifically for the standard authentication, while in fact also (may, I tested with official Jottacloud where it does) applies to the traditional/whitelabel authentication.

See commit messages for some more details.

#### Checklist

### DIFF
--- a/backend/jottacloud/jottacloud.go
+++ b/backend/jottacloud/jottacloud.go
@@ -17,7 +17,7 @@ import (
 	"net/url"
 	"os"
 	"path"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


docs/content/jottacloud.md
```diff
@@ -35,34 +35,26 @@ Paths are specified as `remote:path`
...
-need to replace the token for one of them, using the [config reconnect](https://rclone.org/commands/rclone_config_reconnect/)
-command, which requires you to generate a new personal login token and supply
-as input. If you do not do this, the token may easily end up being invalidated,
-resulting in both instances failing with an error message, something along
-the lines of:
+## Authentication
+
+Authentication in Jottacloud is in general based on OAuth 2.0 and OpenID
+Connect (OIDC). There are different variants to choose from, depending on which
+service you are using, e.g. a white-label service may only support one specific.
```
ncw: ...one specific **method**,

Maybe?

docs/content/jottacloud.md
```diff
@@ -79,36 +71,101 @@ DEBUG : remote: got fatal oauth error: oauth2: "invalid_grant" "Session doesn't
...
+refresh token (family).
+
+You can get some overview of your active tokens in your service's web user
+interface, if you navigate to "Settings" and then "Security" (in which case
+you end up at <https://www.jottacloud.com/web/secure> or similar). Down on
+that page you have a section "My logged in devices", which contains a list
+of active tokens. From the right side of that list you can click a button
+("X") to invalidate the refresh token, which means you will still have access
+using an existing access token until that expires, but you will not be able
+to refresh it. Note that using the different authentication variants have
```
ncw: s/have/has/

docs/content/jottacloud.md
```diff
@@ -79,36 +71,101 @@ DEBUG : remote: got fatal oauth error: oauth2: "invalid_grant" "Session doesn't
...
+different effect, and different white-label services may behave differently.
+
+### Standard
+
+This is an OAuth variant designed for command-line applications. It is
+primarily supported by the official service (jottacloud.com), but may also be
+supported by some of the white-label services. The information necessary to be
+able to perform authentication, like domain name and endpoint to connect to,
+are found automatically (it is encoded into the supplied login token, described
+next), so you do not need to specify service.
```
ncw: specify **the** service

backend/jottacloud/jottacloud.go
```diff
@@ -69,29 +69,29 @@ const (
...
+	name     string
+	domain   string
+	realm    string
+	clientID string
+	scopes   []string
+}
+
+func getServices() []service {
+	return []service{
+		{"jottacloud", "Jottacloud", "id.jottacloud.com", "jottacloud", "desktop", []string{"openid", "jotta-default", "offline_access"}}, // Chose client id "desktop" here, will be identified as "Jottacloud for Desktop" in "My logged in devices", but could have used "jottacli" here as well.
```
ncw, line 85: My experience with the S3 backends leads me to think a comment here would be good as to what order the backends should be in, so "Jottacloud" first with the remainder sorted alphabetically by **key** or **name** not sure

ncw (approved):
Hey this looks great - thank you :-)

I made a few minor notes - see inline.

Please merge when ready!

albertony:
Thank you! I will take your suggestions in.

Regarding the docs; I'm still a bit unsure about how the traditional (web-based) oauth actually works in Jottacloud, ref recent comments in #8852, so I may have to adjust the doc description more before I merge this. I'm not that familiar with many of the technicalities of oauth, and how it is normal for services (i.e. other backends) to behave, so I'm not sure if Jottacloud behavior is normal or at least similar to some others. Do you have some thoughts?
- The facts:
  - Jottacloud backend uses a fixed client_id and no client_secret.
  - Jottacloud service enforces "refresh token rotation" and "automatic reuse detection" (well... not really a documented fact, but a evidence shows...).
- My observations:
   - I create 2 new remotes from scratch, for each of them selecting the traditional web-based oauth, and `Yes (default)` when asked `Use web browser to automatically authenticate rclone with remote?`, entering account username password in web browser, etc.
   - All good, both remotes work - for 1 hour.
   - After 1 hour I refresh the first of them, successfully, and it still works after that.
   - Then I try to refresh the other, and it fails (invalid_grant: Session doesn't have required client).
   - The first one, which I successfully refreshed, continue to work for another hour, but when I attempt to refresh it a second time then it also fails (invalid_grant: Session doesn't have required client).
   - I have to run through reconnect (replace token) to get them to work again, but only for one hour before back to square 1.
     - EDIT: Actually, I only have to reconnect one remote, after I have done reconnect on the first, then not only does the first remote work again, but also the second remote! But only for another hour.
- Other users observe similar situation when using one rclone remote and also has the official desktop client running.
- Most of this is very similar to the known issue when using the "personal login token" method to setup one remote, then just copy that remote config to another. That case ends up with the same error - previously "Stale token", but now I get the same "Session doesn't have required client" description also for this. As I understand it this is an expected result from the "refresh token rotation" and "automatic reuse detection" features, because we end up attempting to refresh using an old and invalidated refresh token that has already been refreshed and replaced.
- It looks like when I create the second (and more) remote with this web-based oauth, I'm actually not getting a new "root" refresh token, I'm in fact getting the same token, or at least in the same "token family". The fact that I only get a single entry in the "My logged in devices" list in the web interface also points in this direction (with two remotes created from separate personal login tokens, I get one entry for each of them).  And then it seems as if the "refresh token rotation" and "automatic reuse detection" features flags this use case as well, and invalidates the entire token set. (EDIT: But the fact that reconnecting one remote also fixes the other is then confusing, that perhaps points to some other explanation?). This would then mean there is effectively a limitation that only a single token can be created. Maybe this is tied to the machine, so a single token per machine, I don't know. Is this normal, since all use same client_id or something? Or at least something you have seen before? Or is it another strange feature in Jottacloud's oauth implementation?

PR #8888, opened 2025-10-09
fs: remove unnecessary Seek call on log file

aneeshusa:

We were seeing a (non-fatal) error in our logs:
```
Failed to seek log file to end: seek /proc/1/fd/1: illegal seek
```

Because we open the log file with O_APPEND,
we don't need to manually seek to the end.
As https://pkg.go.dev/os#File.Seek also confirms
that the behavior of `Seek` is not specified
if the file has been opened with O_APPEND,
remove the `Seek` call.

#### Checklist

I manually tested this by building a new rclone binary and confirm the spurious error goes away (but logging output is emitted to the log-file as normal).

### DIFF
--- a/fs/log/log.go
+++ b/fs/log/log.go
@@ -216,10 +216,6 @@ func InitLogging() {
 			if err != nil {
 				fs.Fatalf(nil, "Failed to open log file: %v", err)
 			}
-			_, err = f.Seek(0, io.SeekEnd)
-			if err != nil {
-				fs.Errorf(nil, "Failed to seek log file to end: %v", err)
-			}
 			redirectStderr(f)
 			w = f
 		} else {


ncw:
@aneeshusa this change looks sensible to me, thank you.

Because I can't remember why it is there can you confirm it works properly on Windows, macOS and Linux? So check that the log gets written to the end of an existing file. I can't think why it wouldn't be but better to check! 

If you can confirm that then I'll merge.

Thank you

aneeshusa:
Thanks for the quick review! I confirmed the fix on Linux and macOS. I don't have access to a Windows machine but read the Go stdlib and the Windows API docs and confirmed the way we use the API is correct.

PR #8890, opened 2025-10-09
Improved debug logs from token refresh

albertony:

While testing various OAuth related scenarios in the Jottacloud backend, and digging into what I claim is a bug with a fix in #8887, I found that the debug logs from the token refresh operations were a bit lacking.

- It is not extremely clear from the debug logs that a token refresh is actually being performed. E.g. a sequence like this does not say it very explicitly:
    ```
    DEBUG : xxweb: Loaded invalid token from config file - ignoring
    DEBUG : Saving config "token" in section "xxweb" of the config file
    DEBUG : Keeping previous permissions for config file: -rw-rw-rw-
    DEBUG : xxweb: Saved new token in config file
    ```

- The log message "Loaded invalid token from config file - ignoring" can be misleading. It can be followed by "Loaded new refresh token from config file" and then it is not really *doing nothing*, because it will actually use the token, both access and refresh part, even though the access token is (still) not valid.

  https://github.com/rclone/rclone/blob/3afa563eaf6c42960de1a948820e2bdaeda69fc9/lib/oauthutil/oauthutil.go#L253-L262

- Background refresher `oauthutil.Renew` performs a backend-specified operation that requires token, and will then end up in regular token refresh functionality in `TokenSource.Token()`. Logging token refresh successful/failed status messages from both places will give duplicate information. It is enough to report it from the regular token retrieval function, I think. But it could make sense for the background-refresher to still report any errors, just in case there is an error in the specified operation that did not end up in an error logged from the regular token refresh code. I got into one such situation with Jottacloud, although it turned out to be a bug in the backend implementation initializing the background refresher too early (see #8887), it still is an example of what might happen. The last log line here shows the error message:
    ```
    jottacloud root '': Background refresher detected expired token - 0 uploads in progress - refreshing
    xxweb: Token expired
    xxweb: Did not find new token in config file
    xxweb: Token refresh successful
    Saving config "token" in section "xxweb" of the config file
    Keeping previous permissions for config file: -rw-rw-rw-
    xxweb: Saved new token in config file
    jottacloud root '': Background token refresh failed: read metadata failed: error 400: org.springframework.security.core.userdetails.UsernameNotFoundException: Username not found in url! (Bad Request)
    ```
- Before I discovered the mentioned bug in Jottacloud backend implementation, it would log "Token expired but no uploads in progress - doing nothing" at startup, which was confusing because it lead me to think there was no token refresh being performed, while actually there was. This message was from the background process, but when running a command that needs a token it will perform token refresh regardless while retrieving the token normally. This overlaps a bit with the first point; that it is not clear from the debug logs that a token refresh is actually being performed. Also, after fixing the backend bug, the token is refreshed normally before the background refresher is started, so the "Token expired but no uploads in progress - doing nothing" is no longer logged during startup. In any case; I think a small tweak of the log message texts from the background refresher a bit, making it more clear that this is in fact work done from a background process, will avoid more confusion.

With this PR the log from regular token refresh at startup, instead of this:

```
DEBUG : xxweb: Loaded invalid token from config file - ignoring
DEBUG : Saving config "token" in section "xxweb" of the config file
DEBUG : Keeping previous permissions for config file: -rw-rw-rw-
DEBUG : xxweb: Saved new token in config file
```

Will be like this:

```
DEBUG : xxweb: Token expired
DEBUG : xxweb: There was not a new token in config file
DEBUG : xxweb: Token refresh successful
DEBUG : Saving config "token" in section "xxweb" of the config file
DEBUG : Keeping previous permissions for config file: -rw-rw-rw-
DEBUG : xxweb: Saved new token in config file
```

And if the refresh was triggered by the background refresher on a long-running operation, it will now be:

```
DEBUG : jottacloud root '': Background refresher detected expired token - 2 uploads in progress - refreshing
DEBUG : xxweb: Token expired
DEBUG : xxweb: There was not a new token in config file
DEBUG : xxweb: Token refresh successful
DEBUG : Saving config "token" in section "xxweb" of the config file
DEBUG : Keeping previous permissions for config file: -rw-rw-rw-
DEBUG : xxweb: Saved new token in config file
```

See also:
- #8887

#### Checklist

### DIFF
--- a/lib/oauthutil/oauthutil.go
+++ b/lib/oauthutil/oauthutil.go
@@ -250,9 +250,7 @@ func (ts *TokenSource) reReadToken() (changed bool) {
 		return false
 	}
 
-	if !newToken.Valid() {
-		fs.Debugf(ts.name, "Loaded invalid token from config file - ignoring")
-	} else {
+	if newToken.Valid() {
 		fs.Debugf(ts.name, "Loaded fresh token from config file")
 		changed = true
 	}
@@ -264,6 +262,8 @@ func (ts *TokenSource) reReadToken() (changed bool) {
 	if changed {
 		ts.token = newToken
 		ts.tokenSource = nil // invalidate since we changed the token
+	} else {
+		fs.Debugf(ts.name, "No updated token found in the config file")
 	}
 	return changed
 }
@@ -319,6 +319,8 @@ func (ts *TokenSource) Token() (*oauth2.Token, error) {
 		return ts.token, nil
 	}
 
+	fs.Debug(ts.name, "Token expired")
+
 	// Try getting the token a few times
 	for i := 1; i <= maxTries; i++ {
 		// Try reading the token from the config file in case it has
@@ -344,6 +346,7 @@ func (ts *TokenSource) Token() (*oauth2.Token, error) {
 
 		token, err = ts.tokenSource.Token()
 		if err == nil {
+			fs.Debug(ts.name, "Token refresh successful")
 			break
 		}
 		if newErr := maybeWrapOAuthError(err, ts.name); newErr != err {
--- a/lib/oauthutil/renew.go
+++ b/lib/oauthutil/renew.go
@@ -47,16 +47,14 @@ func (r *Renew) renewOnExpiry() {
 		}
 		uploads := r.uploads.Load()
 		if uploads != 0 {
-			fs.Debugf(r.name, "Token expired - %d uploads in progress - refreshing", uploads)
+			fs.Debugf(r.name, "Background refresher detected expired token - %d uploads in progress - refreshing", uploads)
 			// Do a transaction
 			err := r.run()
-			if err == nil {
-				fs.Debugf(r.name, "Token refresh successful")
-			} else {
-				fs.Errorf(r.name, "Token refresh failed: %v", err)
+			if err != nil {
+				fs.Errorf(r.name, "Background token refresher failed: %v", err)
 			}
 		} else {
-			fs.Debugf(r.name, "Token expired but no uploads in progress - doing nothing")
+			fs.Debugf(r.name, "Background refresher detected expired token but no uploads in progress - doing nothing")
 		}
 	}
 }


lib/oauthutil/oauthutil.go
```diff
@@ -264,6 +262,8 @@ func (ts *TokenSource) reReadToken() (changed bool) {
 	if changed {
 		ts.token = newToken
 		ts.tokenSource = nil // invalidate since we changed the token
+	} else {
+		fs.Debugf(ts.name, "There was not a new token in config file")
```
albertony: One could argue that this is not necessary to log. If re-reading the config file resulted in no relevant new information then no need to log that, just continue with the token refresh and that will be logged as usual.

lib/oauthutil/renew.go
```diff
@@ -47,16 +47,14 @@ func (r *Renew) renewOnExpiry() {
...
-			if err == nil {
-				fs.Debugf(r.name, "Token refresh successful")
-			} else {
-				fs.Errorf(r.name, "Token refresh failed: %v", err)
+			if err != nil {
+				fs.Errorf(r.name, "Background refresher failed: %v", err)
 			}
 		} else {
-			fs.Debugf(r.name, "Token expired but no uploads in progress - doing nothing")
+			fs.Debugf(r.name, "Background refresher detected expired token but no uploads in progress - doing nothing")
```
albertony, line 57: Same with this, not extremely useful information. But perhaps when looking into those rare cases it have some value. 🤷‍♀️

lib/oauthutil/oauthutil.go
ncw: I think the log is comforting to have :-)

Perhaps a slight reword?

"No updated token found in the config file"

lib/oauthutil/renew.go
```diff
@@ -47,16 +47,14 @@ func (r *Renew) renewOnExpiry() {
...
-			fs.Debugf(r.name, "Token expired - %d uploads in progress - refreshing", uploads)
+			fs.Debugf(r.name, "Background refresher detected expired token - %d uploads in progress - refreshing", uploads)
 			// Do a transaction
 			err := r.run()
-			if err == nil {
-				fs.Debugf(r.name, "Token refresh successful")
-			} else {
-				fs.Errorf(r.name, "Token refresh failed: %v", err)
+			if err != nil {
+				fs.Errorf(r.name, "Background refresher failed: %v", err)
```
ncw: That should probably mention "token" just in case, so "Background token refresher failed..."

lib/oauthutil/renew.go
ncw: I like it.

lib/oauthutil/oauthutil.go
albertony: Great, thanks! I didn't like my original text that much, but gave up finding something better. Your suggestion is much better!

lib/oauthutil/renew.go
albertony: Good point. I found that too many "refresh" and "token" words in these log messages became something of a mouthful, "Background token refresher detected expired token ... - refreshing", and then ended up choosing "Background refresher" as the name of this functionality, and used that for all of them. But you are correct in that other two does mention token later on, "Background refresher detected expired token", and I agree that it could slightly reduce risk of confusion to add it also in this third message

PR #8904, opened 2025-10-18
VFS metadata persistence and visibility enhancements

### LINKED ISSUE #7729: WebDAV: Add support for metadata using PROPFIND/PROPPATCH
rclone v1.66.0-DEV
- os/version: debian 12.5 (64 bit)
- os/kernel: 6.1.0-14-amd64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.21.8
- go/linking: dynamic
- go/tags: cmount

WebDAV (according to https://rclone.org/overview/) does not currently support metadata, but this seems like it should be possible using PROPFIND/PROPPATCH.

Add support for metadata using PROPFIND/PROPPATCH.

#### How to use GitHub

denis-kudelin:

- introduce granular `--vfs-persist-metadata` modes so operators can choose which POSIX-like attributes (owner, mode, times) to persist
- add `--vfs-hide-metadata` to hide sidecar files from directory listings while keeping direct access
- make metadata sidecar JSON encoding stable (numbers as strings, times in RFC3339) and ensure owner updates succeed even when direct backend chown is unsupported
- document the new flags and behaviour, including sidecar semantics

- fixes #7729

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ directories to and from different cloud storage providers.
 - Seagate Lyve Cloud [:page_facing_up:](https://rclone.org/s3/#lyve)
 - SeaweedFS [:page_facing_up:](https://rclone.org/s3/#seaweedfs)
 - Selectel Object Storage [:page_facing_up:](https://rclone.org/s3/#selectel)
+- Servercore Object Storage [:page_facing_up:](https://rclone.org/s3/#servercore)
 - SFTP [:page_facing_up:](https://rclone.org/sftp/)
 - SMB / CIFS [:page_facing_up:](https://rclone.org/smb/)
 - Spectra Logic [:page_facing_up:](https://rclone.org/s3/#spectralogic)
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -183,6 +183,9 @@ var providerOption = fs.Option{
 		Value: "Selectel",
 		Help:  "Selectel Object Storage",
 	}, {
+		Value: "Servercore",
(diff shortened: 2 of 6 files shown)


ncw:
Can you describe what you are trying to achieve with this patch? I'm not sure altering the VFS like this is the right approach.

denis-kudelin:
> Can you describe what you are trying to achieve with this patch? I'm not sure altering the VFS like this is the right approach.

Goal: persist POSIX-like attributes (owner/mode/times) across mounts and optionally hide sidecar metadata; make chmod/chown succeed on remotes that don’t support them.

Why VFS: this is where FUSE attr ops (chmod/chown/utime) are handled today, so one backend‑agnostic implementation avoids per‑remote hacks (WebDAV/S3/Drive/etc.). Defaults remain unchanged; the feature is opt‑in via `--vfs-persist-metadata` (granular modes) and `--vfs-hide-metadata`. Sidecars use a stable JSON encoding.

ncw:
Rclone has a metadata system already which can store general purpose metadata on some backends already.

I think the ideas in your PR would be better split up:

- make the VFS export the metadata from the backend, eg https://rclone.org/s3/#metadata that we already have
- make a wrapper backend which deals with sidecar files for backends which don't have metadata support.

Note that rclone doesn't currently have a way to set metadata except when uploading files which may need to be added.

Adding metadata sidecars direct to the vfs is putting them in the wrong layer in my opinion.

For a large feature like this is would have been useful if you had opened an issue so we could discuss first.

Thanks

PR #8908, opened 2025-10-21
log: Avoid side effects by using own logger instance - fixes #8907

### LINKED ISSUE #8907: Log package interferes with application logging

N/A

When importing the rclone log package, this package unconditionally modifies the default `slog` handler on import. This might be surprising to applications using `slog` for logging expecting the default handler with the corresponding log-level, format etc to be present.
Of course the application _may_ reset the default `slog` handler explicitly but this will also change the format of the rclone log-messages which may be unintented and unexpected.

Furthermore, applications like [Telegraf](https://github.com/influxdata/telegraf) may want to add extra information to the raw rclone log message. To do so, the `rclone/fs/log` package allows to override the handler behavior e.g. by specifying an output function. However, this construct is very fragile as using the default `slog` logger in this function causes the whole program to hang. The reason is that the `rclone/fs/log` overrides the default slog handler. This handler may then be called from the `fs` code through logging a message. In turn the new handler calls the output function which can use the default slog-logger and thus calls the default handler which is pointing to the `rclone/fs/log`  handler. The code below demonstrates the issue:

```go
package main

import (
	"context"
	"log/slog"

	"github.com/rclone/rclone/fs"
	"github.com/rclone/rclone/fs/log"
)

func main() {
	// This logging line is already influenced by the import of fs/log which
	// sets the slog-handler potentially causing side effects in any application
	// using it.
	slog.Info("This is the slog logger!")

	// Setting the handler output to anything using the default slog logger
	// will cause an infinite loop as the default handler will call this
	// function which in turn will call the default handler during processing.
	log.Handler.SetOutput(func(level slog.Level, text string) {
		slog.Log(context.Background(), level, "[myprefix] "+text)
	})

	// This line will trigger the infinite loop mentioned above.
	fs.Log(nil, "A message from rclone FS...")
}
```

v1.7.2

N/A

N/A

N/A

N/A

#### How to use GitHub

srebhan:

Commit dfa4d94 replaced the old logging with the more modern `slog` package. However, this change is intrusive for applications using `slog` as their default logging facility, causing side effects and a fragile construct when using `slog` logging with custom output functions.

This PR sets up a custom logger, avoiding side effects from overriding the `slog` default logger.

resolves #8907 

#### Checklist

### DIFF
--- a/fs/log/slog.go
+++ b/fs/log/slog.go
@@ -36,11 +36,8 @@ func defaultHandler() *OutputHandler {
 	// Create our handler
 	h := NewOutputHandler(os.Stderr, opts, logFormatDate|logFormatTime)
 
-	// Set the slog default handler
-	slog.SetDefault(slog.New(h))
-
-	// Make log.Printf logs at level Notice
-	slog.SetLogLoggerLevel(fs.SlogLevelNotice)
+	// Setup the logger
+	fs.SetLogger(h)
 
 	return h
 }
--- a/fs/log/slog.go
+++ b/fs/log/slog.go
@@ -36,8 +36,17 @@ func defaultHandler() *OutputHandler {
 	// Create our handler
 	h := NewOutputHandler(os.Stderr, opts, logFormatDate|logFormatTime)
 
-	// Setup the logger
+	// Always set rclone's internal logger so rclone logging works
 	fs.SetLogger(h)
+
+	// Optionally set the process-wide default logger so that
+	// log.Print/log.Fatal and slog.Default() also use rclone's handler.
+	// Library consumers can set fs.InitialiseDefaultLogger = false to
+	// prevent this.
+	if fs.InitialiseDefaultLogger {
+		slog.SetDefault(slog.New(h))
+		slog.SetLogLoggerLevel(fs.SlogLevelNotice)
+	}
 
 	return h
 }
--- a/fs/log/slog.go
+++ b/fs/log/slog.go
@@ -27,6 +27,12 @@ var Handler = defaultHandler()
 // This will be adjusted by InitLogging to be the configured levels
 // but it is important we have a logger running regardless of whether
 // InitLogging has been called yet or not.
+//
+// Note that this only sets rclone's private logger via fs.SetLogger
+// so that importing this package has no side effects on the
+// process-wide default slog logger. The CLI and other rclone-as-a-
+// program entry points pick up the default logger redirection in
+// InitLogging instead.
 func defaultHandler() *OutputHandler {
 	// Default options for default handler
 	opts := &slog.HandlerOptions{
@@ -36,17 +42,8 @@ func defaultHandler() *OutputHandler {
 	// Create our handler
 	h := NewOutputHandler(os.Stderr, opts, logFormatDate|logFormatTime)
 
-	// Always set rclone's internal logger so rclone logging works
+	// Set rclone's internal logger so rclone logging works
 	fs.SetLogger(h)
-
-	// Optionally set the process-wide default logger so that
-	// log.Print/log.Fatal and slog.Default() also use rclone's handler.
-	// Library consumers can set fs.InitialiseDefaultLogger = false to
-	// prevent this.
-	if fs.InitialiseDefaultLogger {
-		slog.SetDefault(slog.New(h))
-		slog.SetLogLoggerLevel(fs.SlogLevelNotice)
-	}
 
 	return h
 }
--- a/fs/log.go
+++ b/fs/log.go
@@ -9,6 +9,10 @@ import (
 	"slices"
 )
 
+// logger represents the slog logging facility and should be overriden by
+// the fs/log handling code.
+var logger *slog.Logger = slog.Default()
+
 // LogLevel describes rclone's logs.  These are a subset of the syslog log levels.
 type LogLevel = Enum[logLevelChoices]
 
@@ -134,7 +138,7 @@ func LogLevelToSlog(level LogLevel) slog.Level {
 }
 
 func logSlog(level LogLevel, text string, attrs []any) {
-	slog.Log(context.Background(), LogLevelToSlog(level), text, attrs...)
+	logger.Log(context.Background(), LogLevelToSlog(level), text, attrs...)
 }
 
 func logSlogWithObject(level LogLevel, o any, text string, attrs []any) {
@@ -302,4 +306,9 @@ func PrettyPrint(in any, label string, level LogLevel) {
 		return
 	}
 	LogPrintf(level, label, "\n%s\n", string(inBytes))
+}
+
+// SetLogger overrides the slog logger using the specified handler
+func SetLogger(h slog.Handler) {
+	logger = slog.New(h)
 }
--- a/fs/log.go
+++ b/fs/log.go
@@ -9,9 +9,22 @@ import (
 	"slices"
 )
 
-// logger represents the slog logging facility and should be overriden by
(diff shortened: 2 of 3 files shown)


fs/log/slog.go
```diff
@@ -36,11 +36,8 @@ func defaultHandler() *OutputHandler {
 	// Create our handler
 	h := NewOutputHandler(os.Stderr, opts, logFormatDate|logFormatTime)
 
-	// Set the slog default handler
-	slog.SetDefault(slog.New(h))
```
ncw, line 40: If we don't call this then rclone or any of its dependencies won't get properly logged if they are using `log.Print` or `log.Fatal`.

From the docs

> Setting a logger as the default with
> 
>     slog.SetDefault(logger)
> 
> will cause the top-level functions like Info to use it. SetDefault also updates
> the default logger used by the log package, so that existing applications that
> use log.Printf and related functions will send log records to the logger's
> handler without needing to be rewritten.
>

fs/log/slog.go
```diff
@@ -36,11 +36,8 @@ func defaultHandler() *OutputHandler {
...
-
-	// Make log.Printf logs at level Notice
-	slog.SetLogLoggerLevel(fs.SlogLevelNotice)
```
ncw, line 43: I don't see an equivalent for this line in your patch? Unless the Default logger is Notice level which I don't see in the docs?

ncw (commented):
Thanks for fixing this. Using rclone as a library is probably the least well supported way of using rclone but we need to play nice if people do :-)

I put some queries inline.

ncw:
Any thoughts on the review notes @srebhan ? I think this is an important fix but I want to make sure we get it right.

fs/log/slog.go
srebhan: I do understand, however, we are using rclone as a library and overriding the default slog logger will mess with all our logging and causes the issues mentioned in #8907.
How about we are setting this handler the default from the rclone application? I can add a function to `fs`...

Just for protocol, using a **default** logging facility in a library is bad style and should be avoided. Add a logger interface and allow overriding the logger in the library and then set that logger from the application is a much better pattern as it allows the application to choose how to log stuff.

fs/log/slog.go
srebhan: Well this line overrides the log-bridge of slog! This should not be done within the library but rather in the application so others can use the library without their logging being changed...

srebhan:
I responded to your comments @ncw. I see your point of this PR changing behavior but the current approach is somewhat limited due to the structure. I could put up a PR reworking logging such that the logger can be customized from the application and is not defined in the library itself. What do you think?

srebhan:
@ncw is there any way to push this forward? Currently this issue holds-up bumping the version in Telegraf...

ncw:
@srebhan there's a fundamental tension here.  Rclone was designed as a command line application. It has a API interface in the remote control interface. Both of those are well supported interfaces for rclone. linking with rclone directly and using it as a package is very much a third class citizen. The changes we've got so far in the pull request break rclone as a library and as an API server. So I think what we need to do is put a global flag in the log package `var InitialiseLogger = true` which only makes those logging changes when that flag is `false`.  So you could clear the flag in your initialisation and then rclone won't initialise its log package in the way it does at the moment. How does that sound to you?

nielash:
Just wanted to note that `go1.26.0` added a new [`slog.MultiHandler`](https://pkg.go.dev/log/slog#MultiHandler) type. Does that help with anything here? (If not, ignore and carry on!)

ncw:
@srebhan I pushed a commit on your branch with the InitialiseLogger flag idea. Does that work for you?

@nielash 
> Just wanted to note that `go1.26.0` added a new [`slog.MultiHandler`](https://pkg.go.dev/log/slog#MultiHandler) type. Does that help with anything here? (If not, ignore and carry on!)

Nice find! I think we could use that in the implementation when we log to multiple places at once. It could theoretically let both rclone and the host app share the default logger, but it adds complexity and still doesn't solve the fundamental issue that a library shouldn't impose its handler on the process default. I think opt-out will work better.

srebhan:
@ncw this works for me. We also usually add a log-handler so we can provide traces into the library if things go wrong...

ncw:
@srebhan 
> this works for me. We also usually add a 
log-handler so we can provide traces into the library if things go wrong...

I'm not sure what you mean here. Can you explain more? Thanks

srebhan:
@ncw in Telegraf we customize the default `slog` handler etc in our main process and set the default logger via `slog.SetDefault` to our handler to catch all messages that would otherwise bypass our logging facility.

Now if _any_ library we import also calls `slog.SetDefault` it breaks _any_ main program that redefines the default logger, in our case Telegraf. With your commit you are adding a conditional `fs.InitialiseDefaultLogger` which can prevent `rclone` is calling `slog.SetDefault`.
However, I cannot see how Telegraf can set this flag _before_ the `fs/log` package is imported because the `fs` package itself will import `fs/log` during indirect imports. This means if we in Telegraf do

```golang
import "github.com/rclone/rclone/fs"
```

the default `fs.InitialiseDefaultLogger` flag will be `true` and some `import` in the `fs` package will import `github.com/rclone/rclone/fs/log` which will then call `slog.SetDefault` because the flag above is `true`.

This will break Telegraf's logging as now all our messages pass through your logger/handler! And we cannot prevent this as we need to import `fs` to set the  `fs.InitialiseDefaultLogger` flag but if we import `fs` the `fs/log` package is imported before we can change the flag.

The correct thing to do would be to use a logger instance in the `fs` library (as my commit did) and _not_ set the default logger **in the `fs` library** but do so in your **main** application!

srebhan:
@ncw any update? This is a minimal code-example to trigger the issue (see #8907):

```go
package main

import (
	"context"
	"log/slog"

	"github.com/rclone/rclone/fs"
	"github.com/rclone/rclone/fs/log"
)

func main() {
	// This logging line is already influenced by the import of fs/log which
	// sets the slog-handler potentially causing side effects in any application
	// using it.
	slog.Info("This is the slog logger!")

	// Setting the handler output to anything using the default slog logger
	// will cause an infinite loop as the default handler will call this
	// function which in turn will call the default handler during processing.
	log.Handler.SetOutput(func(level slog.Level, text string) {
		slog.Log(context.Background(), level, "[myprefix] "+text)
	})

	// This line will trigger the infinite loop mentioned above.
	fs.Log(nil, "A message from rclone FS...")
}
```
your change to this PR breaks the setup again.

srebhan:
@ncw can we please restore my original PR and get this merged? I made sure this has no negative effect on rclone itself as it is backward compatible. This PR sits there for a while and Telegraf is now seeing some vulnerabilities due to an old rclone dependency and we cannot upgrade...

ncw:
Sorry for the delay looking at this @srebhan 

I've added another commit to the branch which restores the default logging but only if you call `InitLogging` and not on package init(). I think this should fix your concerns and the combined diff looks sensible now.

Let me know what you think.

Thanks

srebhan:
@ncw no worries. Your fix works for me! Can we please merge this and get a release so we can fix the CVE in Telegraf?

ncw (approved):
Thanks for reviewing @srebhan - will merge now.

srebhan:
@ncw do you have any ETA for a release containing this commit?

ncw:
It is scheduled for v1.75 which you can find the release date in the milestone.

I didn't want to put it in the point release as I deemed it a bit too risky. Alas it just missed the v1.74 release.

PR #8911, opened 2025-10-22
New backend: piqlConnect

trgwii:

I currently work for Piql, and we see rclone as an excellent tool that we already use in company and it's also a tool that I use personally to back up things to Jottacloud. [piqlConnect](https://piql.com/connect/) is a relatively new cloud storage platform whose main purpose is to facilitate uploads / transfers of files to Piql to be encoded onto [piqlFilm](https://piql.com/technology/film/).

This PR adds MVP support for the piqlConnect. I am fairly new to rclone, and Golang in general, so I may need some help to steer this closer to a production-ready backend. I took inspiration from the Box backend, but I didn't manage to integrate everything, notably pacer-related code is still missing from this implementation, but probably should be added.

I also may need to switch over the Azure backend parts to actually use Azure, but I think as the piqlConnect API evolves a bit there may be a desire to make the Azure SAS URL's that are in use more opaque.

I am also pretty confused about how to properly handle situations like when the root path passed into NewFs is a path to a file and not a directory (like with `rclone cat TestPiqlConnect:Workspace/Foo/bar/baz.txt`).

I briefly asked @buengese and he urged me to create a draft PR first. I felt like maybe the rclone forum is more user-focused and is a bit slow when I'm looking for backend implementation-related questions.

I have managed to pass a subset of the encoding tests, and I wanna make ensure the entire test suite passes.

Not that I know of.

#### Checklist

### DIFF
--- a/backend/piqlconnect/piqlconnect.go
+++ b/backend/piqlconnect/piqlconnect.go
@@ -2,6 +2,7 @@
 package piqlconnect
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/hex"
@@ -294,6 +295,24 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 	return f.listPackages(ctx, segments[0])
 }
 
+func makeBlockListStringForPut(i int) bytes.Buffer {
+	var blocklist bytes.Buffer
+	blocklist.Grow(256)
+	blocklist.WriteString(`<?xml version="1.0" encoding="utf-8"?><BlockList>`)
+	for j := 1; j <= i; j++ {
+		blocklist.WriteString("<Uncommitted>")
+		var src [20]byte
+		dec := strconv.AppendInt(src[:0], int64(j), 10)
+
+		var dst [28]byte
+		b64 := base64.StdEncoding.AppendEncode(dst[:0], dec)
+		blocklist.Write(b64)
+		blocklist.WriteString("</Uncommitted>")
+	}
+	blocklist.WriteString("</BlockList>")
+	return blocklist
+}
+
 // Put the object
 //
 // Copy the reader in to the new object which is returned.
@@ -346,16 +365,13 @@ func (f *Fs) Put(ctx context.Context, in io.Reader, src fs.ObjectInfo, optons ..
 			}
 		}
 
-		blocklist := "<?xml version=\"1.0\" encoding=\"utf-8\"?><BlockList>"
-		for j := 1; j <= i; j++ {
-			blocklist += "<Uncommitted>" + base64.StdEncoding.EncodeToString([]byte(strconv.Itoa(j))) + "</Uncommitted>"
-		}
-		blocklist += "</BlockList>"
-		req, err := http.NewRequestWithContext(ctx, "PUT", results[0]+"&comp=blocklist", strings.NewReader(blocklist))
+		blocklist := makeBlockListStringForPut(i)
+		len := blocklist.Len()
+		req, err := http.NewRequestWithContext(ctx, "PUT", results[0]+"&comp=blocklist", &blocklist)
 		if err != nil {
 			return nil, err
 		}
-		req.ContentLength = int64(len(blocklist))
+		req.ContentLength = int64(len)
 		resp, err := f.httpClient.Do(req)
 		if err != nil {
 			return nil, err


ncw:
> I am also pretty confused about how to properly handle situations like when the root path passed into NewFs is a path to a file and not a directory (like with `rclone cat TestPiqlConnect:Workspace/Foo/bar/baz.txt`).

If you pass a file into `NewFs` the backend should return an `Fs` pointing to the directory above, and the special error `fs.ErrorIsFile`. The rest of rclone will then know what to do. It is the backends job to work out if `baz.txt` in your example is a file or a directory.

> I have managed to pass a subset of the encoding tests, and I wanna make ensure the entire test suite passes.

Great! Keep going :-)

The tests guide you through what you need to fix, so fix the first failing test and repeat until everything is fixed. You'll find that the early tests fixes will fix lots of later tests.

Let me know if you need more help.

trgwii:
I've hit something of a conflict, I can make almost all the encoding tests pass with using encoder Decode for List and NewObject, but in order to make some of the integration tests pass (TestIntegration/FsMkdir/FsEncoding) I have to revert to using ToStandardPath, which regresses the encoding tests.

Are there any guarantees for the encoding of the input string parameters of calls to my List, NewObject, Put, Open, etc?
Are there any requirements for the returned encoding of entry names from my List?

My initial intuition was that everything coming in is in "Standard encoding", but encoding tests seem to prefer "no decoding" (aka "decoded"), while integration tests want "Standard encoding".

I would love some clarification on what the rules are :)

backend/piqlconnect/api/types.go
```diff
@@ -0,0 +1,79 @@
+// Package api has type definitions for piqlconnect
+package api
+
+// Item describes a folder or a file as returned by /api/files
+// Folder items always end with "/" and have size 0
+type Item struct {
+	Id        string `json:"id"`
```
romshark: Nitpick: https://www.reddit.com/r/golang/comments/168r2fc/id_or_id/

Also applies to all other identifier in this code.

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+			azureResp, err := f.httpClient.Do(req)
+			if err != nil {
+				return nil, err
+			}
+			if azureResp.StatusCode < 200 || azureResp.StatusCode > 299 {
+				azureRespBytes, err := io.ReadAll(azureResp.Body)
+				if err != nil {
+					return nil, err
+				}
+				return nil, fmt.Errorf("invalid http response code from azure: " + azureResp.Status + "\n" + string(azureRespBytes))
```
romshark:
```suggestion
				return nil, fmt.Errorf("invalid http response code from azure: %s\n%s", azureResp.Status, string(azureRespBytes))
```

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+		resp, err := f.httpClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode < 200 || resp.StatusCode > 299 {
+			respBytes, err := io.ReadAll(resp.Body)
+			if err != nil {
+				return nil, err
+			}
+			return nil, fmt.Errorf("invalid http response code from azure: " + resp.Status + "\n" + string(respBytes))
```
romshark:
```suggestion
			return nil, fmt.Errorf("invalid http response code from azure: %s\n%s", resp.Status, string(respBytes))
```

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+		azureResp, err := f.httpClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		if azureResp.StatusCode < 200 || azureResp.StatusCode > 299 {
+			azureRespBytes, err := io.ReadAll(azureResp.Body)
+			if err != nil {
+				return nil, err
+			}
+			return nil, fmt.Errorf("invalid http response code from azure: " + azureResp.Status + "\n" + string(azureRespBytes))
```
romshark:
```suggestion
			return nil, fmt.Errorf("invalid http response code from azure: %s\n%s", azureResp.Status, string(azureRespBytes))
```

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+		resp, err := f.client.CallJSON(ctx, &rest.Opts{Method: "POST", Path: "/folders"}, &createFolder, nil)
+		if err != nil {
+			return err
+		}
+		folderIdBytes, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		folderId := string(folderIdBytes)
+		_ = folderId
```
romshark: what's this for?

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+		OrganisationId: f.organisationId,
+		PackageId:      f.packageIdCache[segments[1]],
+		FolderPaths:    [1]string{strings.Join(segments[2:], "/")},
+		Recursive:      true,
+	}
+	_, err := f.client.CallJSON(ctx, &rest.Opts{Method: "POST", Path: "/folders/delete"}, &removeFolder, nil)
+	if err != nil {
+		return err
+	}
+	return nil
```
romshark:
```suggestion
	_, err := f.client.CallJSON(ctx, &rest.Opts{Method: "POST", Path: "/folders/delete"}, &removeFolder, nil)
	return err
```

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,722 @@
...
+}
+
+// setMetaData sets the metadata from info
+func (o *Object) setMetaData(info *api.Item) error {
+	if info.Size == 0 && info.Path[len(info.Path)-1] == '/' {
+		return fs.ErrorIsDir
+	}
+	modTime, err := time.Parse(time.RFC3339, info.UpdatedAt)
+	if err != nil {
+		return err
```
romshark:
```suggestion
		return fmt.Errorf("parsing updatedAt as RFC3339: %w", err)
```

Consider wrapping errors in general to give the error message more context.

backend/piqlconnect/piqlconnect.go
trgwii: Remnants I think, removed.

trgwii:
I should probably also say that it is possible for anyone who wants to help test this backend to create a temporary test account at https://test.app.piql.com/
That instance is connected to Stripe in test mode, so if you are asked for card details you can enter a [Stripe test card](https://docs.stripe.com/testing#testing-interactively)
After signing in and creating a team, you can visit https://test.app.piql.com/settings/security to generate an API key that the rclone backend can use.

Edit: API documentation is also available at: https://piql.com/docs/

trgwii:
I don't think draft PR's show up in anyone's feeds, so I marked it ready for review because I believe I still need some guidance on implementing the encoding in all the right places in a sensible way.

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,744 @@
...
+	packageRelativePath := f.opt.Enc.FromStandardPath(strings.Join(segments[2:], "/"))
+	values.Set("path", packageRelativePath)
+	resp, err := f.client.CallJSON(ctx, &rest.Opts{Path: "/files", Parameters: values}, nil, &files)
+	if err != nil {
+		if resp.StatusCode == 403 {
+			return nil, fs.ErrorIsFile
+		}
+		if resp.StatusCode == 404 {
+			return nil, fs.ErrorDirNotFound
+		}
```
romshark:
```suggestion
		switch resp.StatusCode {
		case 403:
			return nil, fs.ErrorIsFile
		case 404:
			return nil, fs.ErrorDirNotFound
		}
```

you may also use `StatusForbidden` and `StatusNotFound` from https://pkg.go.dev/net/http

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,744 @@
...
+			dir := fs.NewDir(f.absolutePathToRclone(file.Path[0:len(file.Path)-1]), modTime)
+			entries = append(entries, dir)
+		} else {
+			var o fs.Object
+			o, err = f.newObjectWithInfo(ctx, f.absolutePathToRclone(file.Path), &file)
+			if err != nil {
+				return nil, err
+			}
+			entries = append(entries, o)
+		}
```
romshark, line 238: since `entries` is always at least the size of `files`, preallocate before the loop with `entries = make(fs.DirEntries, 0, len(files))`

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,744 @@
...
+	segments := getPathSegments(absolutePath)
+	values := url.Values{}
+	values.Set("organisationId", f.organisationID)
+
+	ps := []api.Package{}
+	_, err = f.client.CallJSON(ctx, &rest.Opts{Path: "/packages", Parameters: values}, nil, &ps)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range ps {
```
romshark, line 253: `entries` can be preallocated with `make(fs.DirEntries, 0, len(ps))`. There are more cases in this file where preallocation is possible.

trgwii: I don't want to preallocate this one, as it is possible you have some offline packages etc, making the resulting slice shorter than the preallocated length.

romshark: Ah, right, I missed the fact that there's no `else` case here, sorry :)

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,736 @@
...
+			if azureResp.StatusCode < 200 || azureResp.StatusCode > 299 {
+				azureRespBytes, err := io.ReadAll(azureResp.Body)
+				if err != nil {
+					return nil, err
+				}
+				return nil, fmt.Errorf("invalid http response code from azure: %s\n%s", azureResp.Status, string(azureRespBytes))
+			}
+		}
+
+		blocklist := "<?xml version=\"1.0\" encoding=\"utf-8\"?><BlockList>"
```
romshark:
```suggestion
		blocklist := `<?xml version="1.0" encoding="utf-8"?><BlockList>`
```
You can use raw string literals in such cases to make them more readable.

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,736 @@
...
+		for j := 1; j <= i; j++ {
+			blocklist += "<Uncommitted>" + base64.StdEncoding.EncodeToString([]byte(strconv.Itoa(j))) + "</Uncommitted>"
+		}
+		blocklist += "</BlockList>"
+		req, err := http.NewRequestWithContext(ctx, "PUT", results[0]+"&comp=blocklist", strings.NewReader(blocklist))
```
romshark: this looks rather inefficient. Consider adding a `func makeBlocklist(...) io.Reader` that is at least `optimized` (or `optimized2` if you care for performance more than code readability) from the benchmark below.

WARNING: I didn't test `optimized2`, it's an example that you can do even better.

```
package main

import (
	"bytes"
	"encoding/base64"
	"io"
	"runtime"
	"strconv"
	"strings"
	"testing"
)

func original(i int) io.Reader {
	blocklist := "<?xml version=\"1.0\" encoding=\"utf-8\"?><BlockList>"
	for j := 1; j <= i; j++ {
		blocklist += "<Uncommitted>" + base64.StdEncoding.EncodeToString([]byte(strconv.Itoa(j))) + "</Uncommitted>"
	}
	blocklist += "</BlockList>"
	return strings.NewReader(blocklist)
}

func optimized(i int) io.Reader {
	var blocklist bytes.Buffer
	blocklist.WriteString(`<?xml version="1.0" encoding="utf-8"?><BlockList>`)
	for j := 1; j <= i; j++ {
		blocklist.WriteString("<Uncommitted>")
		blocklist.WriteString(base64.StdEncoding.EncodeToString([]byte(strconv.Itoa(j))))
		blocklist.WriteString("</Uncommitted>")
	}
	blocklist.WriteString("</BlockList>")
	return &blocklist
}

func optimized2(i int) io.Reader {
	var blocklist bytes.Buffer

	estimatedValuesLen := i * 40
	blocklist.Grow(len(`<?xml version="1.0" encoding="utf-8"?>
	<BlockList>
		<Uncommitted>
		</Uncommitted>
	</BlockList>`) + estimatedValuesLen)

	blocklist.WriteString(`<?xml version="1.0" encoding="utf-8"?><BlockList>`)
	for j := 1; j <= i; j++ {
		blocklist.WriteString("<Uncommitted>")

		var src [20]byte
		dec := strconv.AppendInt(src[:0], int64(j), 10)

		var dst [64]byte
		b64 := base64.StdEncoding.AppendEncode(dst[:0], dec)
		blocklist.Write(b64)

		blocklist.WriteString("</Uncommitted>")
	}
	blocklist.WriteString("</BlockList>")
	return &blocklist
}

func BenchmarkOriginal(b *testing.B) {
	var r io.Reader
	for b.Loop() {
		r = original(8)
	}
	runtime.KeepAlive(r)
}

func BenchmarkOptimized(b *testing.B) {
	var r io.Reader
	for b.Loop() {
		r = optimized(8)
	}
	runtime.KeepAlive(r)
}

func BenchmarkOptimized2(b *testing.B) {
	var r io.Reader
	for b.Loop() {
		r = optimized2(8)
	}
	runtime.KeepAlive(r)
}

```

at i<10 the difference is not relevant, but for large blocklists the difference is significant.

at i=8:
```
pkg: benchfmt
cpu: Apple M4 Pro
BenchmarkOriginal-14             3313053               349.7 ns/op          1952 B/op         18 allocs/op
BenchmarkOptimized-14            4701933               256.2 ns/op          1040 B/op         13 allocs/op
BenchmarkOptimized2-14           8276617               142.6 ns/op           496 B/op          2 allocs/op
PASS
ok      benchfmt        4.242s
```

at i=1024:
```
cpu: Apple M4 Pro
BenchmarkOriginal-14                 895           1252738 ns/op        17270026 B/op       2976 allocs/op
BenchmarkOptimized-14              41932             28548 ns/op           73512 B/op       1960 allocs/op
BenchmarkOptimized2-14             79392             15009 ns/op           49200 B/op          2 allocs/op
PASS
ok      benchfmt        4.002s
```

trgwii: I think optimizing this part is a bit silly, since it has to do N web requests prior anyway, but it's implemented anyway :)

backend/piqlconnect/piqlconnect.go
```diff
@@ -0,0 +1,752 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if azureResp.StatusCode < 200 || azureResp.StatusCode > 299 {
+				azureRespBytes, err := io.ReadAll(azureResp.Body)
+				if err != nil {
+					return nil, err
+				}
+				return nil, fmt.Errorf("invalid http response code from azure: %s\n%s", azureResp.Status, string(azureRespBytes))
+			}
```
romshark, line 365: I see this is a recurring pattern. Consider moving this to a function:

```go
func checkResponseOK(resp *http.Response) error {
	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
		return nil
	}
	b, err := io.ReadAll(resp.Body)
	if err != nil {
		return fmt.Errorf("unexpected http response status from azure: %s", resp.Status)
	}
	return fmt.Errorf("unexpected http response status from azure: %s (%s)", resp.Status, b)
}
```

and then use it like that everywhere:
```go
azureResp, err := f.httpClient.Do(req)
if err != nil {
	return nil, err
}
if err := checkResponseOK(azureResp); err != nil {
	return nil, err
}
```

The original code also returns the `io.ReadAll` error instead of the actual cause (unexpected response status). `checkResponseOK` instead just skips the body.

Alternatively, you could wrap all HTTP calls into something like `do(r *http.Request, acceptedStatusCodes ...int) error`

P.S.
I'm not sure what response bodies you get there from Azure, might be a bit risky to put it all in the error message.

PR #8912, opened 2025-10-22
webdav: use SpaceSepList to parse bearer token command

hunshcn:

`strings.Split` is not safe for command. shellquota is need.

Nope.

#### Checklist

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -194,7 +194,7 @@ type Options struct {
 	User               string               `config:"user"`
 	Pass               string               `config:"pass"`
 	BearerToken        string               `config:"bearer_token"`
-	BearerTokenCommand string               `config:"bearer_token_command"`
+	BearerTokenCommand fs.SpaceSepList      `config:"bearer_token_command"`
 	Enc                encoder.MultiEncoder `config:"encoding"`
 	Headers            fs.CommaSepList      `config:"headers"`
 	PacerMinSleep      fs.Duration          `config:"pacer_min_sleep"`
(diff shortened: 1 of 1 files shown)


ncw:
> `strings.Split` is not safe for command. shellquota is need.

The results of the `strings.Split` is never passed to the shell - there is no safety issue here.

However using `strings.Split` means that we can't pass arguments with spaces in.

In other parts of rclone we've used

      --password-command SpaceSepList       Command for supplying password for encrypted configuration

> The argument to this flag should be a command with an optional space separated list of arguments. If one of the arguments has a space in then enclose it in ", if you want a literal " in an argument then enclose the argument in " and double the ". See CSV encoding (https://godoc.org/encoding/csv) for more info.

So maybe we should use that here too?

hunshcn:
In my scene

```
[webdav]
type = webdav
bearer_token_command = /bin/bash -c 'echo $KEY1:$KEY2'
```

SpaceSepList can not handle it. So I think shellquota is need.

ncw:
> In my scene
> 
> ```
> [webdav]
> type = webdav
> bearer_token_command = /bin/bash -c 'echo $KEY1:$KEY2'
> ```
> 
> SpaceSepList can not handle it. So I think shellquota is need.

This would work just fine with SpaceSepList

```
bearer_token_command = /bin/bash -c "echo $KEY1:$KEY2"
```

Or maybe this if you really need the single quotes to be given to bash

```
bearer_token_command = /bin/bash -c "'echo $KEY1:$KEY2'"
```

hunshcn:
Oh you are right. I agree to use SpaceSepList. Although there is a certain learning cost in using it, it is meaningful to be consistent with other similar commands.

I have update this mr using SpaceSepList.

PR #8914, opened 2025-10-23
s3: make it easier to add new S3 providers

roucc:
Before this change, you had to modify a fragile data-structure containing all providers.  This often led to things being out of order, conflicts whilst merging and mistakes.  As well as the changes for one provider being in different places across the file.

After this change, new providers are defined in an easy to edit YAML file, one per provider.

#### Checklist


roucc:
The converter is in this PR but it will likely need to be removed and archived somewhere else, just in case.

ncw (approved):
Thanks for doing this @roucc , it is a great improvement.

I've double checked 
- all `rclone config` the same before and after
- all quirks the same before Vs after 
- binary size is Ok a few 10s of K bigger
- memories size - a few more K - fine
- startup time is 1 or 2 mS out of 30. Also fine.

I'll merge now.

Issue #8905, opened 2025-10-20
Add flag for transferring ACL when do rclone copy command


ivashnikita:
#### What is your current rclone version (output from `rclone version`)?

rclone v1.72.0-DEV

#### What problem are you are trying to solve?

Hi! It seems that I cannot transfer object ACLs using `rclone copy` right now. I saw a flag `-M` for transferring metadata, but rclone doesn't transfer ACLs with it. I'd probably like to implement this functionality. Are there any conceptual objections?

#### How do you think rclone should be changed to solve that?

Add support for ACL migration and possibly verification of ACL migration correctness

#### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

ncw:
You didn't say which backend you are using. I'm going to assume s3 from now on.

If we want it to work then we need to put the ACL into the metadata and from there then copy with `-M` will preserve the ACL.

We need to think of a way of serializing the ACLs into a single string (that is all the metadata supports). We've used JSON format in the drive/onedrive backend but a simpler format may be appropriate.

This probably needs to be opt with a flag in as it will take an extra call per object read and written. We've done this with a bitmask option like this in drive. So an option called `metadata_acl` or something like that with `off|read|write|read,write`

https://github.com/rclone/rclone/blob/484d955ea8fed315261fc7d60b0a0d83f282cfbd/backend/drive/drive.go#L686-L700

If you would like to implement this, hopefully that is enough guidance. If not please ask for more instructions :-)

Thank you

ivashnikita:
Yes, we're talking about S3. It seems like ACL migration should be enabled with a separate flag for greater control, while the `-M` flag is better left separately for metadata. Regarding serialization, it seems unnecessary in our case; the main question is about the correct use of PutObjectAcl . But I think we can discuss the details when I submit a PR. 

Thanks for the prompt response!

ncw:
This really needs to go through the metadata otherwise you'll be building a solution which only works for s3 -> s3 copies.

You might say that surely that is the only thing that will work as ACLs aren't portable, however

[...]

PR #8915, opened 2025-10-23
s3: add transfer and check objects acls logic

ivashnikita:

Added the ability to transfer ACL objects using rclone copy and verify ACL objects using rclone check.

https://github.com/rclone/rclone/issues/8905

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -68,6 +68,46 @@ import (
 	"github.com/rclone/rclone/lib/version"
 )
 
+// rwChoices type for fs.Bits
+type rwChoices struct{}
+
+func (rwChoices) Choices() []fs.BitsChoicesInfo {
+	return []fs.BitsChoicesInfo{
+		{Bit: uint64(rwOff), Name: "off"},
+		{Bit: uint64(rwRead), Name: "read"},
+		{Bit: uint64(rwWrite), Name: "write"},
+		{Bit: uint64(rwFailOK), Name: "failok"},
+	}
+}
+
+// rwChoice type alias
+type rwChoice = fs.Bits[rwChoices]
+
+const (
+	rwRead rwChoice = 1 << iota
+	rwWrite
+	rwFailOK
+	rwOff rwChoice = 0
+)
+
+// Examples for the options
+var rwExamples = fs.OptionExamples{{
+	Value: rwOff.String(),
+	Help:  "Do not read or write the value",
+}, {
+	Value: rwRead.String(),
+	Help:  "Read the value only",
+}, {
+	Value: rwWrite.String(),
+	Help:  "Write the value only",
+}, {
+	Value: rwFailOK.String(),
+	Help:  "If writing fails log errors only, don't fail the transfer",
+}, {
+	Value: (rwRead | rwWrite).String(),
+	Help:  "Read and Write the value.",
+}}
+
 // Register with Fs
 func init() {
 	fs.Register(addProvidersToInfo(&fs.RegInfo{
@@ -704,6 +744,21 @@ In this case, you might want to try disabling this option.
 			Advanced: true,
 			Default:  false,
 		}, {
+			Name: "metadata_acl",
+			Help: `Control whether object ACLs should be transferred or checked.
+
+Object access can be controlled through various methods, including ACLs.
+
+This flag transfers object ACLs during copy operations and check ACLs during check operations.
+Note that when bucket owners differ, the ACL owner is updated to the new bucket owner.
+Other grantees in the ACL that are not the source bucket owner remain unchanged.
+
+Also note that this flag is relevant if the metadata flag is enabled.
+`,
+			Advanced: true,
+			Default:  rwOff,
+			Examples: rwExamples,
+		}, {
 			Name:     "sts_endpoint",
 			Help:     "Endpoint for STS (deprecated).\n\nLeave blank if using AWS to use the default endpoint for the region.",
 			Advanced: true,
@@ -1085,6 +1140,7 @@ type Options struct {
 	UseDualStack                bool                 `config:"use_dual_stack"`
 	LocationConstraint          string               `config:"location_constraint"`
 	ACL                         string               `config:"acl"`
+	MetadataACL                 rwChoice             `config:"metadata_acl"`
 	BucketACL                   string               `config:"bucket_acl"`
 	RequesterPays               bool                 `config:"requester_pays"`
 	ServerSideEncryption        string               `config:"server_side_encryption"`
@@ -1187,6 +1243,8 @@ type Object struct {
 	meta         map[string]string // The object metadata if known - may be nil - with lower case keys
 	mimeType     string            // MimeType of object - may be ""
 	versionID    *string           // If present this points to an object version
+	acl          *string           // Object ACL
+	bucketOwner  *string           // Bucket owner
 
 	// Metadata as pointers to strings as they often won't be present
 	storageClass       *string // e.g. GLACIER
@@ -4370,7 +4428,63 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 		})
 	}
 
+	if o.fs.opt.MetadataACL.IsSet(rwRead) {
+		aclString, err := o.getACLString(ctx)
+		if err != nil {
+			fs.Logf(o, "Failed to get acl: %v", err)
+			return resp.Body, nil
+		}
+		o.acl = &aclString
+
+		bucketOwner, err := o.getBucketOwner(ctx)
+		if err != nil {
+			fs.Logf(o, "Failed to get bucket owner: %v", err)
+			return resp.Body, nil
+		}
+		o.bucketOwner = &bucketOwner
+	}
+
 	return resp.Body, nil
+}
+
+func (o *Object) getACLString(ctx context.Context) (string, error) {
+	bucket, bucketPath := o.split()
+
+	resp, err := o.fs.c.GetObjectAcl(ctx, &s3.GetObjectAclInput{
+		Bucket: &bucket,
+		Key:    &bucketPath,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	objectACLBytes, err := json.Marshal(types.AccessControlPolicy{
+		Grants: resp.Grants,
+		Owner:  resp.Owner,
+	})
+	if err != nil {
+		return "", nil
+	}
+
+	return string(objectACLBytes), nil
+}
+
+func (o *Object) getBucketOwner(ctx context.Context) (string, error) {
+	bucket, _ := o.split()
+
+	resp, err := o.fs.c.GetBucketAcl(ctx, &s3.GetBucketAclInput{
+		Bucket: &bucket,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	ownerBytes, err := json.Marshal(resp.Owner)
+	if err != nil {
+		return "", err
+	}
+
+	return string(ownerBytes), nil
 }
 
 var warnStreamUpload sync.Once
@@ -4771,8 +4885,10 @@ func (o *Object) uploadSinglepartPresignedRequest(ctx context.Context, req *s3.P
 
 // Info needed for an upload
 type uploadInfo struct {
-	req       *s3.PutObjectInput
-	md5sumHex string
+	req         *s3.PutObjectInput
+	aclReq      *s3.PutObjectAclInput
+	bucketOwner *types.Owner
+	md5sumHex   string
 }
 
 // Prepare object for being uploaded
@@ -4800,6 +4916,14 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
 			ui.req.StorageClass = types.StorageClass(strings.ToUpper(tier))
 		}
 	}
+
+	if o.fs.opt.MetadataACL.IsSet(rwWrite) {
+		ui.aclReq = &s3.PutObjectAclInput{
+			Bucket: &bucket,
+			Key:    &bucketPath,
+		}
+	}
+
 	// Fetch metadata if --metadata is in use
 	meta, err := fs.GetMetadataOptions(ctx, o.fs, src, options)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


fs/operations/check.go
```diff
@@ -4,15 +4,18 @@ import (
...
 	"fmt"
 	"io"
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
 
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
```
ncw: Your pull request shouldn't need to change anything outside `backend/s3` so remove all changes under `fs`

backend/s3/s3.go
```diff
@@ -3605,6 +3663,21 @@ func (o *Object) Size() int64 {
 	return o.bytes
 }
 
+// ACL returns an object ACL and a bucket owner object
+func (o *Object) ACL(ctx context.Context) (string, string, error) {
```
ncw: I don't think this function is called from `backend/s3` and we don't want to add a new public method to Object so I think it can be deleted.

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
 		})
 	}
 
+	if o.fs.opt.MetadataACL.IsSet(rwRead) {
+		aclString, err := o.getACLString(ctx)
+		if err != nil {
+			fs.Logf(o, "Failed to get acl: %v", err)
+			return resp.Body, nil
```
ncw: Why isn't this just returning an error?

return nil, fs.Errorf("failed to read ACL: %w", err)

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
+		}
+		o.acl = &aclString
+
+		bucketOwner, err := o.getBucketOwner(ctx)
+		if err != nil {
+			fs.Logf(o, "Failed to get bucket owner: %v", err)
```
ncw: Why isn't this just returning an error?

return nil, fs.Errorf("failed to get bucket owner: %w", err)

backend/s3/s3.go
```diff
@@ -667,6 +707,21 @@ In this case, you might want to try disabling this option.
 			Help:     `Suppress setting and reading of system metadata`,
 			Advanced: true,
 			Default:  false,
+		}, {
+			Name: "metadata_acl",
+			Help: `Control whether object ACLs should be transferred or checked.
```
ncw: This should be something like this I think

```
Control whether object ACLs should be read or written to the metadata

In conjunction with the --metadata flag this can be used to read or
write ACLs or preserve ACLs dring a copy operation with "read,write"

Note that when writing ACLs if bucket owners differ, the ACL owner
is updated to the new bucket owner. Other grantees in the ACL that
are not the source bucket owner remain unchanged.
```

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
 
+func (o *Object) getACLString(ctx context.Context) (string, error) {
+	bucket, bucketPath := o.split()
+
+	resp, err := o.fs.c.GetObjectAcl(ctx, &s3.GetObjectAclInput{
+		Bucket: &bucket,
+		Key:    &bucketPath,
+	})
+	if err != nil {
+		return "", err
```
ncw: Please wrap the error `fmt.Errorf("get ACL failed: %w", err)`

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
+	}
+
+	objectACLBytes, err := json.Marshal(types.AccessControlPolicy{
+		Grants: resp.Grants,
+		Owner:  resp.Owner,
+	})
+	if err != nil {
+		return "", nil
```
ncw: This has ignored the error

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
+}
+
+func (o *Object) getBucketOwner(ctx context.Context) (string, error) {
+	bucket, _ := o.split()
+
+	resp, err := o.fs.c.GetBucketAcl(ctx, &s3.GetBucketAclInput{
+		Bucket: &bucket,
+	})
+	if err != nil {
+		return "", err
```
ncw: Wrap the error please

backend/s3/s3.go
```diff
@@ -3994,9 +4067,65 @@ func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.Read
...
+	}
+
+	ownerBytes, err := json.Marshal(resp.Owner)
+	if err != nil {
+		return "", err
```
ncw: Wrap the error please

backend/s3/s3.go
```diff
@@ -4416,6 +4554,24 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
 			ui.req.ContentType = pv
 		case "x-amz-tagging":
 			ui.req.Tagging = pv
+		case "x-amz-acl":
+			if !o.fs.opt.MetadataACL.IsSet(rwWrite) {
+				continue
+			}
+			var acl types.AccessControlPolicy
+			if err := json.Unmarshal([]byte(v), &acl); err != nil {
+				fs.Debugf(o, "failed to parse metadata %s: %q: %v", k, v, err)
```
ncw: This should return a wrapped error

backend/s3/s3.go
```diff
@@ -4416,6 +4554,24 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
...
+			}
+			ui.aclReq.AccessControlPolicy = &acl
+		case "bucket-owner":
+			if !o.fs.opt.MetadataACL.IsSet(rwWrite) {
+				continue
+			}
+			var owner types.Owner
+			if err := json.Unmarshal([]byte(v), &owner); err != nil {
+				fs.Debugf(o, "failed to parse metadata %s: %q: %v", k, v, err)
```
ncw: This should return a wrapped error

backend/s3/s3.go
```diff
@@ -4434,6 +4590,19 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
 		}
 	}
 
+	if o.fs.opt.MetadataACL.IsSet(rwWrite) {
+		resp, err := o.fs.c.GetBucketAcl(ctx, &s3.GetBucketAclInput{
+			Bucket: &bucket,
+		})
+		if err != nil {
+			fs.Logf(o, "failed to get bucket acl: %v", err)
```
ncw: Return the error

backend/s3/s3.go
```diff
@@ -4434,6 +4590,19 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
...
+		}
+		ui.aclReq.AccessControlPolicy, err = o.mapACL(ui.aclReq.AccessControlPolicy, ui.bucketOwner, resp.Owner)
+		if err != nil {
+			fs.Logf(o, "failed to map acl: %v", err)
```
ncw: Return the error

backend/s3/s3.go
```diff
@@ -4538,6 +4707,57 @@ func (o *Object) prepareUpload(ctx context.Context, src fs.ObjectInfo, options [
 	return ui, nil
 }
 
+// We need to transfer ownership of objects to the owner of the dst bucket
+// if the objects were owned by the owner of the src bucket
+func (o *Object) mapACL(srcACL *types.AccessControlPolicy, srcBucketOwner, dstBucketOwner *types.Owner) (*types.AccessControlPolicy, error) {
+	if srcACL == nil {
+		return nil, fmt.Errorf("src acl is nil")
```
ncw: Be more explicit about the error - what should the user do here? Or is it an internal error which should never happen?

backend/s3/s3.go
```diff
@@ -4732,6 +4971,7 @@ var (
 	_ fs.CleanUpper      = &Fs{}
 	_ fs.OpenChunkWriter = &Fs{}
 	_ fs.Object          = &Object{}
+	_ fs.ObjectWithACL   = &Object{}
```
ncw: Please remove

ncw (changes requested):
This code looks like it is getting there :-)

The metadata system should be able to deal with all the metadata transfers so you don't need to change anything except in backend/s3

Can you also update the system metadata please

https://github.com/rclone/rclone/blob/ee92673e1b060f83da2c7253b2bbc9f3d43a680c/backend/s3/s3.go#L842-L843

Are you planning to put `acl` and `bucket-owner` in since you are reading `bucket-owner`?

See inline for comments - mostly about error handling!

Thank you and looking forward to the next version.

backend/s3/s3.go
ivashnikita: There are various difficulties with ACLs, for example, when we have read permissions for an object but not read permissions for the ACL. In this case, we have two options: either move the object without the ACL and notify the user of the error, or don't move the object at all. Your approach seems better: either the object is fully moveable or it isn't. I fixed it based on your comments

backend/s3/s3.go
ivashnikita: This sounds better

backend/s3/s3.go
ivashnikita: fixed it

backend/s3/s3.go
ivashnikita: I deleted it

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: fixed

backend/s3/s3.go
ivashnikita: It seems the second case is an internal error. If the user has set the ACL transfer flag, then the user must ensure that the ACLs are set. For example, a user transfers files without ACLs from the file system to S3 and encounters this error. Basically, we can only report that there is no ACL, and it's hard to say what the user should do - get an ACL from somewhere or remove the flag. There's a wide range of actions here

backend/s3/s3.go
ivashnikita: did it

fs/operations/check.go
ivashnikita: It seems that in this case I won't be able to verify ACLs using `rclone check`, or will I? If not, then we can simply address this in a separate PR, and then resolve the migration issue for now.  I delete everything outside of `backend/s3`

backend/s3/s3.go
ivashnikita: fixed

ivashnikita:
Hi! I've corrected it based on your comments

ivashnikita:
@ncw Just following up on this PR! Looking forward to your comments. Thanks!

ivashnikita:
Hey @ncw! Wondering about review timelines for this PR. 
We're excited about these changes and looking forward to getting them merged.

ivashnikita:
@ncw Hi! It looks like this PR may have stalled. Could we revisit the question of transferring the ACL?

fs/operations/check.go
ncw: `rclone check` isn't for verifying metadata - only file contents so we'd need to think of a new mechanism for this.

ncw (approved):
Thank you - this is looking super :-)

I will merge it now.

ivashnikita:
It seems you need to restart the build/windows job. It failed due to the external service returning the error "Back-end server is at capacity"

ivashnikita:
@ncw Hey! Could you take a look at the pipeline so we can deliver the changes as planned? :-)

ivashnikita:
@ncw Hi! Just a friendly ping on the PR in case it slipped through. Is the failing CI job blocking things right now?

ivashnikita:
@ncw @albertony Hi! Sorry to bother you - just checking in on the PR status. It looks like the changes haven’t been merged yet. Is there anything else needed before it can be merged?

ivashnikita:
@ncw Hey! Sorry for the ping - I know there’s a lot to take care of. Since you’ve approved the changes, could we get them merged and delivered when you have a moment? That would be great - thanks!

ivashnikita:
@ncw Hey! It looks like this PR got lost. You approved it and wanted to merge it. Let's finish this story :-)

ivashnikita:
@ncw Hi! This one's rebased, approved and mergeable. Shall we land it?

daskol:
Honestly, why use `rclone` or invest time in improving it when there is no feedback from the maintainers for months, especially when modern AI tools can help build a similar solution from scratch in a couple of days and fix bugs along the way?

mazen-isstech:
> Honestly, why use `rclone` or invest time in improving it when there is no feedback from the maintainers for months, especially when modern AI tools can help build a similar solution from scratch in a couple of days and fix bugs along the way?

It seems to be actively maintained, judging by the commits. It is also a battle tested tool that, and as far as I can see, is reliable.

Vibe coding a critical tool like rclone, which requires a lot of code, using probabilistic models in a couple of days would not have been a good idea without a ton of human code review on top of it. The tool and the test suite would both have to be reviewed. That's not a trivial task.

PR #8771, opened 2025-08-24
More Markdown format lint

albertony:
This is a continuation of https://github.com/rclone/rclone/pull/8678, to make sure our Markdown docs uses a valid and consistent format. Here I've started to work through the command docs and backend docs, and have quite many issues, mostly things like this:

- MD007/ul-indent: Unordered list indentation [Expected: 0; Actual: 2]
- MD046/code-block-style: Code block style [Expected: fenced; Actual: indented]
- MD040/fenced-code-language: Fenced code blocks should have a language specified
- MD012/no-multiple-blanks: Multiple consecutive blank lines [Expected: 1; Actual: 2]
- MD004/ul-style: Unordered list style [Expected: dash; Actual: asterisk]
- MD032/blanks-around-lists: Lists should be surrounded by blank lines
- MD013/line-length: Line length [Expected: 80; Actual: 153]

Since the files in focus this time are partially or entirely generated from source code, the process was to: build rclone, generate markdown, lint, fix issues in the sources of the markdown text, and repeat. This was done as a local and manual process for now, but the future plan is to get the linting of these running on GitHub pull requests alongside our other markdown linters, e.g. as I did a PoC on with https://github.com/rclone/rclone/pull/8704.

The work in this area is not done, but with these changes there are risk of git conflicts, so I wanted to ship a batch as soon as possible, and see if we are able to merge it before I continue with a new iteration.

First of all we need to see the result of the checks for edits in autogenerated sections, that runs on the PR. Then I will run some more tests locally, to verify that there are no regressions in the command line help and the generated HTML pages, which are based on the same sources.

Next iteration (which I now plan as a new PR after this is merged) will start by going through the backend options - I've only updated the static parts of the backend docs so far in this PR.

Note that with the command docs, there is an issue with the cobra library we use, it generates parts of the markdown and currently the linter reports a couple of issues in that. I have changed it locally for now, to be able to lint the generated command docs without issues, but will have to think about what to do with this later.

#### Checklist

- [ ] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [ ] I have added tests for all changes in this PR if appropriate.
- [ ] I have added documentation for the changes if appropriate.
- [ ] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ ] I'm done, this Pull Request is ready for review :-)

PR #8934, opened 2025-10-31
Markdown lint fixes

albertony:

Another bulk of markdown lint fixes.

Mostly in code not currently covered by our linter in ci, due to autogenerated parts. We had #8704 as a sort of proof of concept for handling these, so I will probably pick up that at some point, but before something like that can go live there are a lot of lint problems to fix, so I thought I'd get rid of *some* of them...

There is one change worth nothing: I've changed the code block language that are used for typical command examples (which there are a lot of) from `sh` to `console`. I introduced many of them myself in https://github.com/rclone/rclone/pull/8771, when converting code block style from indented to fenced, but I'm starting to think I picked the wrong one then. The `sh` is for shell script, and does some syntax colorization that not always makes sense in general console. One issue is backslashes. If you see the following windows specific examples in https://rclone.org/commands/rclone_mount/, you'll see that it interprets the backslash as escape characters:

```
rclone mount remote:path/to/files C:\path\parent\mount
rclone mount remote:path/to/files \\cloud\remote
```

Also in https://rclone.org/s3/ you can see that integers in "random" places are colorized.

In https://rclone.org/commands/rclone_backend/ a single `=` sign is colorized.

The downside of using the `console` language block is that the syntax highlighting seem to be very limited. I guess that was the reason I did not like it initially, and chose `sh` instead. There are some comments in the project sites of several of the relevant renderers, criticizing it, and suggesting improvements, but there does not seem to be much progress. Some renderers have rules detecting the prompt prefix to color command and output differently, but not consistent as some only look for a single `$` while others look for `:~`, and there is a lint rule [MD014](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md) that says that prompt prefix should only be used when output is included, and many of our examples are just a command without output, so it is not a clear case anyway... In reality it seems `console` has very little effect, and is mostly comparable to just `text`. I will look more into this topic later, but I think for now it is better to do the switch and adjust back later if needed. I've kept `sh` for some examples, e.g. where there are several complex lines, typical linux commands and arguments, and also if there are `#` prefixed comments, which is nice to get colored.

Related to https://github.com/rclone/rclone/pull/8778.

#### Checklist

### DIFF
--- a/docs/content/azureblob.md
+++ b/docs/content/azureblob.md
@@ -1044,7 +1044,7 @@ Properties:
 
 ### Custom upload headers
 
-You can set custom upload headers with the `--header-upload` flag. 
+You can set custom upload headers with the `--header-upload` flag.
 
 - Cache-Control
 - Content-Disposition
@@ -1053,19 +1053,21 @@ You can set custom upload headers with the `--header-upload` flag.
 - Content-Type
 - X-MS-Tags
 
-Eg `--header-upload "Content-Type: text/potato"` or `--header-upload "X-MS-Tags: foo=bar"`
+Eg `--header-upload "Content-Type: text/potato"` or
+`--header-upload "X-MS-Tags: foo=bar"`.
 
 ## Limitations
 
 MD5 sums are only uploaded with chunked files if the source has an MD5
 sum.  This will always be the case for a local to azure copy.
 
-`rclone about` is not supported by the Microsoft Azure Blob storage backend. Backends without
-this capability cannot determine free space for an rclone mount or
-use policy `mfs` (most free space) as a member of an rclone union
+`rclone about` is not supported by the Microsoft Azure Blob storage backend.
+Backends without this capability cannot determine free space for an rclone
+mount or use policy `mfs` (most free space) as a member of an rclone union
 remote.
 
-See [List of backends that do not support rclone about](https://rclone.org/overview/#optional-features) and [rclone about](https://rclone.org/commands/rclone_about/)
+See [List of backends that do not support rclone about](https://rclone.org/overview/#optional-features)
+and [rclone about](https://rclone.org/commands/rclone_about/).
 
 ## Azure Storage Emulator Support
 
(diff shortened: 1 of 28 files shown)


ncw:
A thought for you from the forum: https://forum.rclone.org/t/unable-to-get-any-color-despite-color-always/52930 - this confusion was caused by the colouring in the docs not matching the colouring that rclone actually does.

In particular look at  the first example here: https://rclone.org/commands/rclone_dedupe/

This is generated with an indented block

    $ rclone lsl drive:dupes
      6048320 2016-03-05 16:23:16.798000000 one.txt
      6048320 2016-03-05 16:23:11.775000000 one.txt
       564374 2016-03-05 16:23:06.731000000 one.txt
      6048320 2016-03-05 16:18:26.092000000 one.txt
      6048320 2016-03-05 16:22:46.185000000 two.txt
      1744073 2016-03-05 16:22:38.104000000 two.txt
       564374 2016-03-05 16:22:52.118000000 two.txt

So not sure what language that is set to. But something to think about as we don't want to cause user confusion,

albertony:
> In particular look at the first example here: https://rclone.org/commands/rclone_dedupe/

Yes, another good (bad) example of "random" colouring...

> This is generated with an indented block

Hm, I see that the checked in docs/content/commands/rclone_dedupe.md has indented block, but in [source](https://github.com/rclone/rclone/blob/becc068d363d86091612df66de190bdbb342fb75/cmd/dedupe/dedupe.go#L71) it is fenced block with sh as language, and when I serve it using Hugo locally I see that how it looks on rclone.org matches fenced with sh block and not indented block.

ncw:
Ah yes, we didn't run make docs yet after your last set of changes.

I think using `console` instead of `sh` will help with this problem and would have fixed the forum users confusion.

ncw (approved):
I had a brief look through these and I think they are a good idea. Thanks for persevering with this. Please merge when ready.

PR #8942, opened 2025-11-04
Improve autogenerated marker in backend docs

albertony:

The initial reason was to automatically add suppression of markdownlint line length error reports. While looking at this I found that it made sense to change the backend docs autogenerated markers from Hugo shortcodes to HTML comments, and then some additional adjustments followed. See the individual commits (messages) for more details of exactly what was done.

The changes hasn't been discussed in any detail before, there is no issue for this, so this is more or less just a suggestion from me that comes as a PR.

Some background:

The existing format looks like this:

```
{{< rem autogenerated options start" - DO NOT EDIT - instead edit fs.RegInfo in backend/archive/archive.go then run make backenddocs" >}}
### Standard options
```

The first line here is not recognized by Markdown editors, so it will treat it as any other line of text. Also, it has two markdownlint issues, if we were to enable it on backend docs as is:
- MD013/line-length: Line length
- MD022/blanks-around-headings: Headings should be surrounded by blank lines

This uses a custom `rem` Hugo shortcode template. It was introduced in 74b8cbfb841c1b5491d5f840b2c541e91177bd6c, to avoid warnings "raw HTML omitted while rendering" without having to enable the Hugo option for unsafe HTML parsing enabled. However, since then (I think); 
- Hugo has introduced a built-in comments shortcode:

  > Code within a comment is not parsed, executed, or displayed.
  
  ```
  {{/* This is a comment. */}}
  
  {{/*
  This is a block comment.
  */}}
  ```
  
  ```
  {{- /* This is a comment with adjacent whitespace removed. */ -}}
  
  {{- /*
  This is a block comment with
  adjacent whitespace removed.
  */ -}}
  
  ```
- HTML comments are now allowed in Hugo without enabling unsafe HTML parsing.

  > Hugo strips HTML comments when rendering a page

  However with one caveat:
 
  > ... but first evaluates any template code within the HTML comment delimiters. Depending on the template code within the HTML comment delimiters, this could cause unexpected results or fail the build.
  > Do not attempt to use HTML comment delimiters to comment out template code.

I chose HTML comment, as that is something that Markdown editors typically understand and highlights (with color). Also, rhe markdownlint supression is HTML comment, so this makes it more consistent.

https://github.com/rclone/rclone/pull/8937#discussion_r2490341744

#### Checklist


albertony:
Great, thanks! My assumptions were correct then. :)

I can squash the commits, no problem.

PS: Regarding the commit:

> Avoid impression that make backenddocs has to be run and results committed, since we have a lint check which will then report error because we want to prevent manual changes in autogenerated sections.

It was a very subtle change, don't know if it made too much of a difference, but I could not come up with something while keeping it short enough to fit in one line and within the view of "readers". Maybe there should be a section in the CONTRIBUTING.md which explains more explicit what to do (and not do), and the marker comment could refer to it. Not sure. Probably a task for later.

PR #8952, opened 2025-11-09
backend/ulozto: Fix downloads returning HTML error page

aliaj1:
The uloz.to backend was failing to download files, instead returning an HTML page with a "Slow download" message. This was caused by recent changes in the uloz.to API.

This commit fixes the issue by making the following changes to the download process:

1.  The `hash` received from the download link API is now appended as a query parameter to the download URL.
2.  The download is now performed using the authenticated `rest` client to ensure premium access is recognized.
3.  The `DeviceID` is now generated dynamically for each download request to avoid potential rate-limiting of a static ID.

These changes together resolve the download issue and allow files to be downloaded correctly again.


jwh:
Works perfectly, thanks!

PR #8954, opened 2025-11-09
fix: load config with errors

Its-Just-Nans:

The call `config/dump` fails poorly. This MR fix that by sending an error

fix:
- https://github.com/rclone/rclone/issues/8955

No

#### Checklist


Its-Just-Nans:
Hi @ncw 

I've updated my MR to create the `LoadedDataWithErr()` function.

This functions is used in `DumpRcBlob()`.

This will wrap
- the rc call `config/dump`
- the CLI `rclone config dump`

Issue #2428, opened 2018-08-01
Feature Request: Support B2 application keys


stevesbrain:
Hey - since the release of the B2 [application keys](https://www.backblaze.com/b2/docs/application_keys.html), it would be great if rclone can support per-bucket keys! As far as I can tell, the only modification needed is to make the call to b2_list_buckets only try and list the bucket we're attempting to use:

> In an application key restricted to one bucket, listing buckets is only allowed if you provide the bucket's name or ID in the b2_list_buckets request. 
> https://www.backblaze.com/b2/docs/application_keys.html

As it stands currently, the following error occurs when attempting to perform an action with a single-bucket key:

```
$ rclone ls b2-api:/ -vv
2018/08/01 09:55:36 DEBUG : rclone: Version v1.42 starting with parameters [rclone ls b2-api:/ -vv]
2018/08/01 09:55:36 DEBUG : Using config file from /home/user/.config/rclone/rclone.conf
2018/08/01 09:55:38 DEBUG : B2 bucket : Unauthorized: Unknown 401  (401 unauthorized)
2018/08/01 09:55:38 DEBUG : pacer: Rate limited, increasing sleep to 20ms
2018/08/01 09:55:38 DEBUG : pacer: low level retry 1/10 (error Unknown 401  (401 unauthorized))
```
Rclone config is the same as when using master ID + key, except only using application key + ID instead. 

Thanks :)

hit-xchvpit:
Second this one. Thought it was just some mistyped ID/key until I found this issue. Even created a second app key for testing but still the same result.

Remote tree listing test also throws an error, and possibly with every custom app key authentication:
```
$ rclone tree remote: -vv
2018/08/01 17:15:57 DEBUG : rclone: Version "v1.42" starting with parameters ["rclone" "tree" "remote:" "-vv"]
2018/08/01 17:16:06 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2018/08/01 17:16:07 DEBUG : pacer: Rate limited, increasing sleep to 20ms
2018/08/01 17:16:07 DEBUG : pacer: low level retry 1/10 (error Invalid authorization (401 unauthorized))
2018/08/01 17:16:07 DEBUG : pacer: Rate limited, increasing sleep to 40ms
2018/08/01 17:16:07 DEBUG : pacer: low level retry 2/10 (error Invalid authorization (401 unauthorized))
[...]

PR #8978, opened 2025-11-18
b2: support authentication with new bucket restricted application keys

DianaNites:

Backblaze has updated its b2_authorize_account API endpoint, newly created application keys are now "multi-bucket" keys, capable of being limited to multiple buckets.
These keys can only be used with the v4 endpoint,
not v1 which returns an HTTP 400.

This commit switches authorization to the v4 endpoint, and allows such keys to work with any one of the allowed buckets.

Without this change newly created application keys with even one bucket restriction no longer work. I discovered this when creating a new key with the backblaze cli tool restricted to one bucket which worked fine with rclone before, and now fails to authenticate.

With this change rclone now works with these newly created v4 keys and passes tests. No other endpoints needed to be changed.

No

#### Checklist

### DIFF
--- a/backend/b2/b2.go
+++ b/backend/b2/b2.go
@@ -1079,51 +1079,68 @@ type listBucketFn func(*api.Bucket) error
 
 // listBucketsToFn lists the buckets to the function supplied
 func (f *Fs) listBucketsToFn(ctx context.Context, bucketName string, fn listBucketFn) error {
-	var bucketID string
-	for _, b := range f.info.APIs.Storage.Allowed.Buckets {
-		if b.Name == bucketName {
-			bucketID = b.ID
-			break
+	responses := make([]api.ListBucketsResponse, len(f.info.APIs.Storage.Allowed.Buckets))[:0]
+
+	for i := range f.info.APIs.Storage.Allowed.Buckets {
+		b := &f.info.APIs.Storage.Allowed.Buckets[i]
+		// Empty names indicate a bucket that no longer exists, this is non-fatal
+		// for multi-bucket API keys.
+		if b.Name == "" {
+			continue
 		}
-	}
-	var account = api.ListBucketsRequest{
-		AccountID: f.info.AccountID,
-		BucketID:  bucketID,
-	}
-	if bucketName != "" && account.BucketID == "" {
-		account.BucketName = f.opt.Enc.FromStandardName(bucketName)
+		// When requesting a specific bucket skip over non-matching names
+		if bucketName != "" && b.Name != bucketName {
+			continue
+		}
+
+		var account = api.ListBucketsRequest{
+			AccountID: f.info.AccountID,
+			BucketID:  b.ID,
+		}
+		if bucketName != "" && account.BucketID == "" {
+			account.BucketName = f.opt.Enc.FromStandardName(bucketName)
+		}
+
+		var response api.ListBucketsResponse
+		opts := rest.Opts{
+			Method: "POST",
+			Path:   "/b2_list_buckets",
+		}
+		err := f.pacer.Call(func() (bool, error) {
+			resp, err := f.srv.CallJSON(ctx, &opts, &account, &response)
+			return f.shouldRetry(ctx, resp, err)
+		})
+		if err != nil {
+			return err
+		}
+		responses = append(responses, response)
 	}
 
-	var response api.ListBucketsResponse
-	opts := rest.Opts{
-		Method: "POST",
-		Path:   "/b2_list_buckets",
-	}
-	err := f.pacer.Call(func() (bool, error) {
-		resp, err := f.srv.CallJSON(ctx, &opts, &account, &response)
-		return f.shouldRetry(ctx, resp, err)
-	})
-	if err != nil {
-		return err
-	}
 	f.bucketIDMutex.Lock()
 	f.bucketTypeMutex.Lock()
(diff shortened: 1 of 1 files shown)


backend/b2/b2.go
```diff
@@ -607,17 +607,28 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
...
+	// If this is a key limited to number of bucket, all must exist already,
+	// and one of them must be ours.
+	if f.rootBucket != "" && len(f.info.APIs.Storage.Allowed.Buckets) != 0 {
+		buckets := f.info.APIs.Storage.Allowed.Buckets
+		var rootFound = false
+		var rootID string
+		for _, b := range buckets {
+			allowedBucket := f.opt.Enc.ToStandardName(b.Name)
+			if allowedBucket == "" {
+				return nil, errors.New("bucket that application key is restricted to no longer exists")
```
ncw: This probably doesn't want to be a fatal error any more - there might be lots of buckets and if our bucket is still around then that is OK. Probably I'd turn that into a `fs.Debugf` and `continue` the loop

backend/b2/b2.go
```diff
@@ -1067,9 +1078,16 @@ type listBucketFn func(*api.Bucket) error
 
 // listBucketsToFn lists the buckets to the function supplied
 func (f *Fs) listBucketsToFn(ctx context.Context, bucketName string, fn listBucketFn) error {
+	var bucketID string
+	for _, b := range f.info.APIs.Storage.Allowed.Buckets {
```
ncw: I'm not sure this is 100% correct. If this is called with `bucketName == ""` then it should list all the buckets we are allowed to look at and I don't think it is doing that.

Otherwise this function is used for looking up bucket IDs

ncw: So `rclone lsd b2:` should show all the buckets that the key is allowed to see

DianaNites: I thought of that, but it doesnt seem to be possible on the b2 API side because only the key management APIs have changed, the list bucket API still only takes one bucket ID or name, and presumably still errors if one is not given.

On the rclone-side it could just put this entire function in a loop and call it for *every* bucket ID, but that seemed a bit much and didnt know if it would mess with other things somehow.

backend/b2/b2.go
DianaNites: Can do, didnt want to change the existing behavior.

backend/b2/b2.go
ncw: I was hoping that if `bucketName == ""` then `listBucketsToFn` could call `fn` for all buckets in `f.info.APIs.Storage.Allowed.Buckets` without doing any API calls. However we don't have the Type of the bucket available if we do that.

Does `rclone lsd b2:` show all the buckets that the key can access? I think it probably will since I would hope "/b2_list_buckets" would filter them. That is my main concern.

DianaNites: I havent tested that and before seeing your reply I have already done a commit that does the loop over all allowed buckets.

/b2_list_buckets however is [documented](https://www.backblaze.com/apidocs/b2-list-buckets) to fail, not filter, however. "When using an authorization token that is restricted to a bucket, you must include the bucketId or bucketName of that bucket in the request, or the request will be denied."

DianaNites: That issue seems to be the entire reason list buckets provides the bucket ID at all, in fact https://github.com/rclone/rclone/issues/2428

DianaNites: I have now tested rclone lsd, without a549a9c5fe4e23e5275ac31b9e0ef5df8c01fc93, list buckets does indeed return 401 Unauthorized on the root backend.

With the changes, it returns the buckets supported by the key, with the intended semantics of one list call per existing bucket, only when listing the root, no extra API calls for directly listing a bucket

DianaNites:
Rebased onto master

ncw (approved):
I think this looks perfect now - thank you :-)

I'll merge it once the CI goes green.

ncw:
@DianaNites unfortunately this has broken the integration tests, eg

```console
$ cd backend/b2
$ go test -v -run TestIntegration/FsMkdir/FsPutFiles/FsListDirRoot
=== RUN   TestIntegration
    fstests.go:438: Using remote "TestB2:"
=== RUN   TestIntegration/FsMkdir/FsPutFiles/FsListDirRoot
    fstests.go:1050: 
        	Error Trace:	/home/ncw/go/src/github.com/rclone/rclone/fstest/fstests/fstests.go:1050
        	Error:      	[]string{} does not contain "rclone-test-rokahir7fiya"
        	Test:       	TestIntegration/FsMkdir/FsPutFiles/FsListDirRoot
        	Messages:   	Remote leaf not found
--- FAIL: TestIntegration (6.98s)
    --- FAIL: TestIntegration/FsMkdir (5.58s)
        --- FAIL: TestIntegration/FsMkdir/FsPutFiles (4.32s)
            --- FAIL: TestIntegration/FsMkdir/FsPutFiles/FsListDirRoot (0.45s)
FAIL
exit status 1
FAIL	github.com/rclone/rclone/backend/b2	6.989s
```

Have you got time to investigate?

Thanks

(The integration tests run off a remote called `TestB2` which has a full access key).

DianaNites:
I thought it passed on merge? I'll do my best to investigate soon

For my part i've been using my patch without issue across both old and new bucket-limited keys, so nothing immediately stands out

PR #9013, opened 2025-12-06
Add support for piped config

### LINKED ISSUE #9012: Unable to pass configuration to rclone via pipe / stdin

Unable to pass configuration to rclone via pipe / stdin.

Without a way to pipe the configuration and / or secrets, there is no way to use rclone on remote machines unless one also uploads secrets that can be easily stolen by anyone with procfs access (environment variables or command line arguments), or directly from the remote machine's filesystem.

Example of actual use case:

`ssh machine "rclone --config /dev/stdin sha1sum --download remote:" < rclone.conf > cryptfs-hashes.log`

Typical use would be to launch a VM close to where the storage is, such as on AWS for S3 backed storage, or Hetzner VM for data on a storage box, so that the 10GbE network links between storage and rclone can be used, rather than a ~100mbit connection to a far away laptop from which the storage related commands for rclone are being issued. I.e. low latency and high throughput advantage, especially important for hashing remote files since rclone's crypt backend doesn't support storing nor retrieving hash metadata for files.

```
rclone v1.72.0
- os/version: ubuntu 22.04 (64 bit)
- os/kernel: 6.8.0-87-generic (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.25.4
- go/linking: static
- go/tags: none
```

* Ubuntu 22.04.5 LTS, 64bit
* Alpine Linux v3.22, 64bit
* Windows 10, 64bit

N/A

`echo -e "[remote]\ntype = test" | rclone --config /dev/stdin listremotes`

```
2025/12/06 07:20:34 DEBUG : rclone: Version "v1.72.0" starting with parameters ["./rclone" "-vv" "--config" "/dev/stdin" "listremotes"]
2025/12/06 07:20:34 CRITICAL: Failed to load config file "/dev/stdin": seek /dev/stdin: illegal seek
```

#### How to use GitHub

jojje:

resolves #9012

No documentation update made since this change simply looses the constraint on what file-like objects can be used to pass the information. Previously only seekable file objects were supported, but this change now allows also non-seekable streams, such as `stdin` or named pipes, which is what users would expect to work by default. For instance `/dev/stdin`.

See issue #9012

#### Checklist

### DIFF
--- a/fs/config/configfile/configfile_test.go
+++ b/fs/config/configfile/configfile_test.go
@@ -2,6 +2,7 @@ package configfile
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


jojje:
Ok, this is as tight as I can make the change.
If you'd want me to move the test to some other file, let me know. The current file was what seemed most appropriate though, since it was either this one, or some crypt related test. What tipped the scale for this one was the domain affinity of the issue; about config file, not crypto stuff. It just so happened the bug originated from an orthogonal concern.

PR #9016, opened 2025-12-07
s3: use DeleteObjectsInput API to delete files in batch - fixes #8160

### LINKED ISSUE #8160: Use S3 DeleteObjects api when deleting multiple objects/versions e.g. with `purge`

https://forum.rclone.org/t/any-reason-rclone-does-not-use-s3-deleteobjects-when-deleting-multiple-objects/48476

```
rclone v1.68.1
- os/version: darwin 14.7.1 (64 bit)
- os/kernel: 23.6.0 (arm64)
- os/type: darwin
- go/version: go1.23.1
- go/linking: dynamic
- go/tags: none
```

Faster deletes e.g. when purging an S3 bucket.

Use the S3 `DeleteObjects` api which is a single `POST` request that accepts a list of up to 1000 object keys/versions in the body. That should be faster than sending 1000 individual `DELETE` requests.
https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html

Nick replied in the forum:
> The s3 backend should definitely be using that in its [Purge 1](https://github.com/rclone/rclone/blob/b9207e57274cd6e7c488c5bd751a058ba0f8b3b9/backend/s3/s3.go#L5466-L5471) implementation (which recursively deletes a directory).
> Elsewhere the s3 backend could be doing that, but it would have to use [lib/batcher 1](https://github.com/rclone/rclone/blob/master/lib/batcher/batcher.go) which is reasonably complicated but not impossible.

#### How to use GitHub

lobshunter:

This PR speeds up the `purge` command on S3 backends by using DeleteObjects API as discussed in #8160. Another change that makes purging faster is setting `quiet: true` in the API request.

In my personal use case(purging ~1M of objects), the time cost drops from several hours to ~20 minutes.

#8160

#### Checklist

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -330,25 +330,25 @@ If empty it will default to the environment variable "AWS_PROFILE" or
 		}, {
 			Name: "role_arn",
 			Help: `ARN of the IAM role to assume.
-			
+
 Leave blank if not using assume role.`,
 			Advanced: true,
 		}, {
 			Name: "role_session_name",
 			Help: `Session name for assumed role.
-			
+
 If empty, a session name will be generated automatically.`,
 			Advanced: true,
 		}, {
 			Name: "role_session_duration",
 			Help: `Session duration for assumed role.
-			
+
 If empty, the default session duration will be used.`,
 			Advanced: true,
 		}, {
 			Name: "role_external_id",
 			Help: `External ID for assumed role.
-			
+
 Leave blank if not using an external ID.`,
 			Advanced: true,
 		}, {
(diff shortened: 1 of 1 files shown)


backend/s3/s3.go
```diff
@@ -842,6 +842,7 @@ const (
 	maxUploadParts      = 10000 // maximum allowed number of parts in a multi-part upload
 	minChunkSize        = fs.SizeSuffix(1024 * 1024 * 5)
 	defaultUploadCutoff = fs.SizeSuffix(200 * 1024 * 1024)
+	purgeBatchSize      = 1000 // the max number of objects allowed in a DeleteObjects request
```
lobshunter, line 845: This number comes from AWS [document](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html). 

> The request can contain a list of up to 1,000 keys that you want to delete.

It also works on Alibaba Cloud, but I'm kind of worried some uncommon S3 providers may set a lower limit on this. Should I set a lower value or add a new config option?

PR #9024, opened 2025-12-10
FileLu: Fix memory leak during download and add multipart upload support

kingston125:

This update resolves a memory leak caused by reading entire file downloads into memory (body, _ := io.ReadAll(resp.Body)).
It also adds support for uploading large files using multipart upload.

No

#### Checklist


backend/filelu/filelu_object.go
```diff
@@ -59,80 +58,63 @@ func (f *Fs) NewObject(ctx context.Context, remote string) (fs.Object, error) {
...
-			if offset > int64(len(currentContents)) {
-				return false, fmt.Errorf("offset %d exceeds file size %d", offset, len(currentContents))
-			}
-			currentContents = currentContents[offset:]
-		}
-		if count > 0 && count < int64(len(currentContents)) {
-			currentContents = currentContents[:count]
+	// 6) Validate response codes
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+		body, _ := io.ReadAll(resp.Body)
```
ncw: This is a memory leak waiting to happen here - loading the whole reponse into memory is not a good idea

ncw (changes requested):
Can you identify the actual problem please? 

> The issue was caused by buffers not being released correctly when handling streamed responses.
> The fix ensures that memory is freed promptly after each read operation, improving stability and preventing excessive RAM usage during large downloads.

That doesn't tell me what the problem is. Go has a garbage collector.

Once you've identified the actual problem can you send me a fix to **just** fix that please? This patch rewrites the whole function changing everything for no good reason. It also has removed the pacer retries which are essential for reliability.

I don't mind looking at code that people have made with AI but I expect them to review it and tidy it first. Definitely remove the `1)` `2)` `3)` from the comments - they get out of date very quickly.

Maybe rewrite the whole function using the rest library which can do most of the work for you. Here is the `Open` function from the `box` backend which deals with range requests etc and should just work for you with minimal adaptation.

```go
func (o *Object) Open(ctx context.Context, options ...fs.OpenOption) (in io.ReadCloser, err error) {
	if o.id == "" {
		return nil, errors.New("can't download - no id")
	}
	fs.FixRangeOption(options, o.size)
	var resp *http.Response
	opts := rest.Opts{
		Method:  "GET",
		Path:    "/files/" + o.id + "/content",
		Options: options,
	}
	err = o.fs.pacer.Call(func() (bool, error) {
		resp, err = o.fs.srv.Call(ctx, &opts)
		return shouldRetry(ctx, resp, err)
	})
	if err != nil {
		return nil, err
	}
	return resp.Body, err
}
```

Thank you

backend/filelu/filelu_object.go
kingston125: Thank you for identifying the issue. I’ve updated the code and kept the pacer.call, only fixing the io.ReadAll.

kingston125:
Hi @ncw 
The old code had an issue when uploading large files, so we also added multipart upload support for the FileLu backend.

kingston125:
Happy New Year, @ncw ! Wishing you a productive and successful new year, and continued growth and improvements for Rclone. I really appreciate all the hard work you put into the project.

If everything looks good, could you please merge this? If not, please let me know what I need to work on.

Thank you.

PR #9040, opened 2025-12-14
Fix headers hierarchy for rclone_mount.md

darkdragon-001:

Similar to "Mounting on macOS", "Mounting on Linux" needs to be a sub-heading of the general "rclone". This also fixes the missing anchor overlay on hover.

#### Checklist

### DIFF
--- a/docs/content/commands/rclone_mount.md
+++ b/docs/content/commands/rclone_mount.md
@@ -336,7 +336,7 @@ full new copy of the file.
 When mounting with `--read-only`, attempts to write to files will fail *silently*
 as opposed to with a clear warning as in macFUSE.
 
-# Mounting on Linux
+## Mounting on Linux
 
 On newer versions of Ubuntu, you may encounter the following error when running
 `rclone mount`:


albertony:
Good catch, thanks!

I'm not sure about the header levels below the "Mounting on Linux", though. I think maybe they need to be as they were. E.g. "Limitations" section is not Linux-specific. What do you think?

darkdragon-001:
@albertony I reverted the other changes. PTAL.

darkdragon-001:
@albertony do you know when the documentation gets updated? On https://rclone.org/commands/rclone_mount/ it still shows v1.33 while v1.37 including this change has been released in the meantime.

albertony:
That label shows when the command was first introduced. So the documentation are updated regularly:

> After your changes have been merged, you can verify them on [tip.rclone.org](https://tip.rclone.org/). This site is updated daily with the current state of the master branch at 07:00 UTC. The changes will be on the main [rclone.org](https://rclone.org/) site once they have been included in a release.

(https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md)

darkdragon-001:
@albertony Then it still doesn't work. Do you have an idea how to make this heading appear in the table of contents (only visible in the desktop layout)?

albertony:
Hm, yes, I think we messed up. The change was done in an autogenerated file. Sorry for not noticing before. We have a linter rule for that now, so if the same was done today I think the PR checks would fail for this reason.

It needs to be done here:

https://github.com/rclone/rclone/blob/master/cmd/mountlib/mount.md

And there is a quirk here, if I remember correctly: The header is already `## Mounting on Linux`, but one level is removed as part of the process of merging this into the final page. I cannot check it to be 100% sure right now, but you see that another header is `### Mounting on macOS`.

Are you willing to do a second attempt with a new PR?

darkdragon-001:
@albertony Thanks for your explanation. Next attempt in #9211.

PR #9051, opened 2025-12-18
drive: add optional cycle detection for shortcut loops

fperron-kt88:

This change adds an **optional cycle detection mechanism** to the Google Drive backend to prevent infinite recursion during recursive listings when directory shortcuts form cycles.

Shared Drives can contain shortcut graphs such as:
- Folder A → shortcut to Folder B
- Folder B → shortcut back to Folder A

When using recursive operations (e.g. `lsjson --recursive` or `--fast-list`), these cycles can cause traversal to loop indefinitely.  
The new flag allows users to safely enable cycle detection when working with such Drive layouts.

Cycle detection is **disabled by default** to preserve existing traversal behavior.

---

### What does this change do?

- Adds a new advanced backend flag: `--drive-cycle-detection` (default: `false`)
- When enabled, tracks visited Google Drive directory IDs during `ListR` traversal
- Skips already-visited directories to prevent infinite recursion
- Emits debug-level log messages when cycles are detected and skipped
- Adds a unit test validating flag parsing and default behavior

When the flag is disabled, there is **no change in behavior or performance**.

---

### Was the change discussed in an issue or the forum before?

No.  
This addresses an observed traversal hang caused by cyclic shortcut graphs in Shared Drives.

---

### Backward compatibility

Cycle detection is opt-in to avoid changing traversal semantics for users who rely on repeated visits through multiple shortcut paths.

- The flag defaults to `false`, preserving existing behavior
- No performance impact when disabled
- When enabled, some directory paths may be visited only once even if reachable via multiple shortcut paths

---

#### Unit tests

- Added TestCycleDetectionFlag to validate flag parsing and default behavior:
  - Default value is `false`
  - Flag parsing correctly enables cycle detection

Run with:
```bash
go test ./backend/drive -run TestCycleDetectionFlag
```
---

### Integration testing

Traversal behavior was validated manually using a personal test Google Drive with cyclic shortcuts.

Due to the requirement for authenticated Drive credentials and specific shortcut graph layouts, integration tests are not included in CI. Existing Drive integration tests are unaffected and will be skipped unless a TestDrive: remote is configured.

### Checklist

[x]  I have read the contribution guidelines
[x]  I have added tests for this change where appropriate (unit only here, manual in a test env)
[ ]  I have added documentation for this change (flag is advanced and self-describing)
[x]  All commit messages follow the rclone house style
[x] I’m done; this Pull Request is ready for review

### DIFF
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -2208,6 +2208,8 @@ func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (
 		return err
 	}
 	directoryID = actualID(directoryID)
+	visited := sync.Map{}
+	visited.Store(directoryID, true)
 
 	mu := sync.Mutex{} // protects in and overflow
 	wg := sync.WaitGroup{}
@@ -2239,8 +2241,14 @@ func (f *Fs) ListR(ctx context.Context, dir string, callback fs.ListRCallback) (
 	// Send the entry to the caller, queueing any directories as new jobs
 	cb := func(entry fs.DirEntry) error {
 		if d, isDir := entry.(fs.Directory); isDir {
-			job := listREntry{actualID(d.ID()), d.Remote()}
-			sendJob(job)
+			id := actualID(d.ID())
+			if _, alreadyVisited := visited.Load(id); !alreadyVisited {
+				visited.Store(id, true)
+				job := listREntry{id, d.Remote()}
+				sendJob(job)
+			} else {
+				fs.Debugf(f, "Skipping directory %q due to cycle detection", d.Remote())
+			}
 		}
 		mu.Lock()
 		defer mu.Unlock()
--- a/backend/drive/drive.go
+++ b/backend/drive/drive.go
@@ -752,6 +752,14 @@ two accounts.
 				Value: "true",
 				Help:  "Get GCP IAM credentials from the environment (env vars or IAM).",
 			}},
+		}, {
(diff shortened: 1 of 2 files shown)


fperron-kt88:
bump. any comment? should I open a bug first?

fperron-kt88:
Hello Nick Craig-Wood,

I guess that the PR was closed after a period of inactivity.

Can you help me figure-out why this PR was overlooked? I tried to interact,
but got no real traction in doing so…

Any feedback would be appreciated!

(personal details edited out... sorry..., also fixed a typo in appreciated above)

ncw:
Sorry, nothing to do with the quality of your work and everything to do with the massive pr queue.

This came up in another issue and I had an insight that we could use the data we already have in the directory cache to solve the problem. I then went hunting for related issues and I found this PR :-(

I'm sorry this didn't get merged. It is no reflection on you, and we'd be happy to receive further pull requests from you we will try to merge more quickly.

If in the future you want a review on a pull request add a comment. This generates a notification which should be seen.

Thank you and sorry 

Nick

fperron-kt88:
Hello Nick, thanks for the clarification. All good then, thanks for the tip: I'll be more active on the comments in the future ;-)

Can I have a pointer to the actual solution to the problem, I'd point to it here to leave a trace and signal to what amounts to a dupe?

ncw:
It is in the text above

> ncw closed this in 0baac15

PR #9055, opened 2025-12-21
fix(hasher): prevent double-download when syncing with auto_size

darthShadow:

When syncing from hasher with auto_size larger than file size, files were downloaded twice: once by fs.Fingerprint() calling Hash() which triggered updateHashes(), and again for the actual transfer.

Root cause: checkPartial() calls Fingerprint(ctx, src, fast=true) for generating stable partial upload suffixes. Fingerprint checks SlowHash before calling Hash(), but hasher didn't set SlowHash even when computing hashes requires downloading the entire file.

Fix:
- Set SlowHash=true when passHashes is empty (no fast pass-through hashes from base). Must be set AFTER Mask() to avoid being ANDed to false.
- Add singleflight to deduplicate concurrent updateHashes() calls for the same file, preventing redundant downloads from parallel Hash() requests.

When passHashes is non-empty (e.g., S3 with MD5), SlowHash remains false and Fingerprint uses the fast pass-through hash. When passHashes is empty (e.g., crypt/SFTP base), Fingerprint skips hashing entirely. Explicit rclone hashsum commands still work as they call Hash() directly.

N/A

#### Checklist

### DIFF
--- a/backend/hasher/hasher.go
+++ b/backend/hasher/hasher.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/fs/list"
 	"github.com/rclone/rclone/lib/kv"
+	"golang.org/x/sync/singleflight"
 )
 
 // Register with Fs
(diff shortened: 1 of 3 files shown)


darthShadow:
Ping @ncw

darthShadow:
Another ping for @ncw

PR #9056, opened 2025-12-21
fix(size): display results even when listing errors occur

darthShadow:

Previously, any error during directory enumeration (e.g., a subdirectory being deleted mid-operation) would cause the size command to exit without displaying results, even though partial counts were successfully computed.

Remove the early return on error and ensure results are always printed before returning. The error is still returned for proper exit code handling, so scripts checking exit status continue to work correctly.

For JSON output, if both Count and JSON encoding fail, the Count error takes priority as it's more meaningful to the user.

N/A

#### Checklist

### DIFF
--- a/cmd/size/size.go
+++ b/cmd/size/size.go
@@ -58,14 +58,19 @@ of the size command.`,
 			}
 
 			results.Count, results.Bytes, results.Sizeless, err = operations.Count(context.Background(), fsrc)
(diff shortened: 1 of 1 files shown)


darthShadow:
Ping @ncw

darthShadow:
Another ping for @ncw

PR #9057, opened 2025-12-22
backend/dropbox,mega,swift: fix integer underflow in free space calculation

IsaccoBenedetti:
Sometimes a remote reports that used space is slightly greater than total space. When this happens, the calculation `free := total - used` was resulting in a negative integer. 
Since it was stored in an unsigned int, this caused the free space to be reported as an enormous amount, and also caused issues when used in combination with a union remote with policies like mfs, as rclone would try to upload all new files to the full remote.
I simply added a small check to ensure that if used > total the free space is reported as 0 instead of underflowing.

Not that I know of.

#### Checklist

- [x] I have added tests for all changes in this PR if appropriate. -> tested manually, unit tests probably unnecessary for such a small fix
- [x] I have added documentation for the changes if appropriate. -> unnecessary


ncw:
I guess you are seeing this in rclone mount?

It might be better to fix that code (in cmd/mountlib I think), perhaps as well.

You could make your fix a little briefer by using `max(0, total - used)`

The convention in rclone is that backends only put stuff they actually know in Usage and the higher layers fill in the blanks.

That is another possibility - remove the calculations and make sure the mountlib code deals with used > total properly.

What do you think?

IsaccoBenedetti:
Hi ncw, and thank you for your review!

I'm sorry, I'm not completely familiar neither with Go, nor with rclone codebase, so bear with me if I'm missing something, but I don't think it's just a mount problem: you can easily reproduce the bug using rclone about. For example, this is before and after I upload a file that would make my dropbox free account full:
```
PS C:\Users\isacc\Projects\rclone\rclone> rclone about dropbox:
Total:   2 GiB
Used:    1.867 GiB
Free:    135.764 MiB
PS C:\Users\isacc\Projects\rclone\rclone> rclone about dropbox:
Total:   2 GiB
Used:    2.124 GiB
Free:    8 EiB
```
The same problem was also happening in rclone webUI. I'm actually not even using mount, I'm serving the union drive via SFTP.

Actually, since it's possible this same issue is also present in other backends as well, it would be nice if there was a centralized place in which this check could be made for all of them. I haven't seen an apt place, but you surely know the codebase better than me, so if you can think of any suggestions that would be great. As things stand, I wanted to keep things as simple as possible so I've limited myself only to those backends I can properly test.

About your max(0, total - used) suggestion: I tried, but it doesn't work if we keep the variables unsigned ints. What happens is that first used and total get subtracted, underflowing and resulting in a huge number. Then, max selects the huge number, as it should.
Here's the output with my if statement:
```
PS C:\Users\isacc\Projects\rclone\rclone> go build
PS C:\Users\isacc\Projects\rclone\rclone> ./rclone about dropbox:
Total:   2 GiB
Used:    2.124 GiB
Free:    0 B
```
And here's the one using max:
```
PS C:\Users\isacc\Projects\rclone\rclone> go build
PS C:\Users\isacc\Projects\rclone\rclone> ./rclone about dropbox:
Total:   2 GiB
Used:    2.124 GiB
Free:    8 EiB
```

Alternatively, maybe we could make those variables signed ints, then it should work, but it may be more invasive.

IsaccoBenedetti:
Hi ncw! Just a quick update: I've been running this patch on my VPS for about a month now and it completely solved the issue both with the one of the disks filling up and the rest staying empty, and the error in free space reporting. I also haven't noticed any side effects or regressions. Standing by if you need any changes.

Thanks!

Issue #4998, opened 2021-02-05
Mount: When the cache space is insufficient, repeated strings will be generated in the log file.


BlackMickey:
#### What is the problem you are having with rclone?
When the cache space is insufficient, duplicate strings will be generated in the log file..

#### What is your rclone version (output from `rclone version`)
rclone v1.54.0
- os/arch: linux/amd64
- go version: go1.15.7

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
Debian 10, arm64 , 64 bit

####  Which cloud storage system are you using? (e.g. Google Drive)
Google Drive

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
rclone mount gdrive: /mnt/gdrive \
--config='~/.config/rclone/rclone.conf' \
--allow-other \
--read-only \
--vfs-cache-mode=full
--vfs-cache-max-size=12G \
--vfs-cache-poll-interval=1m \
--cache-dir=/cache  \
--log-level=INFO \
 --log-file=/tmp/rclone.log

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
```
2021/02/06 02:16:12 NOTICE: Serving remote control on http://localhost:5572/
2021/02/06 02:16:12 INFO  : Starting transaction limiter: max 8 transactions/s with burst 8
2021/02/06 02:20:12 INFO  : vfs cache: cleaned: objects 0 (was 0) in use 0, to upload 0, uploading 0, total size 0 (was 0)
2021/02/06 02:21:12 INFO  : vfs cache: cleaned: objects 1 (was 1) in use 1, to upload 0, uploading 0, total size 288M (was 288M)
2021/02/06 02:22:12 INFO  : vfs cache: cleaned: objects 2 (was 2) in use 2, to upload 0, uploading 0, total size 1.241G (was 1.241G)
...
2021/02/06 02:33:12 INFO  : vfs cache: cleaned: objects 15 (was 15) in use 15, to upload 0, uploading 0, total size 12.310G (was 12.310G)
2021/02/06 02:33:31 INFO  : Test.mkv: vfs cache: downloader: error count now 1: vfs reader: failed to write to cache file: write /cache/vfs/gdrive/Test.mkv: no space left on device
2021/02/06 02:33:31 ERROR : Test.mkv: vfs cache: failed to download: vfs reader: failed to write to cache file: write /cache/vfs/gdrive/Test.mkv: no space left on device
[...]

PR #9061, opened 2025-12-25
fix: Mount: When the cache space is insufficient, repeated string

majiayu000:
## Summary

Fix for #4998: Mount: When the cache space is insufficient, repeated strings will be generated in the log file.

## Changes

- `vfs/vfscache/downloaders/downloaders.go`
- `test_all.log`

## Test Plan

- [x] Existing tests pass
- [ ] Manual verification

### DIFF
--- a/test_all.log
+++ b/test_all.log
@@ -1,0 +1,1 @@
+bash: test_all: command not found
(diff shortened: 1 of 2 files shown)


majiayu000:
Closing this PR due to issues found during review: accidentally included test_all.log file and the fix may be incomplete. Will submit a proper fix.

Issue #6242, opened 2022-06-13
Preserve folder creation dates after rclone copy from S3 / MinIO buckets


mirekphd:
#### What is your current rclone version (output from `rclone version`)?
/ $ rclone --version
rclone v1.58.1-DEV
- os/version: alpine 3.15.4 (64 bit)
- os/kernel: 3.10.0-957.27.2.el7.x86_64 (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.18.2
- go/linking: dynamic
- go/tags: none

#### What problem are you are trying to solve?
Restore backup data from remote S3-compat. bucket (MinIO 2022), while preserving the original folder creation times (currently only works for files).

The command I'm using for copying data from the remote backup:
```
rclone --progress --check-first --create-empty-src-dirs --multi-thread-streams=$NTHREADS --s3-upload-concurrency=$MULTIPART_NTHREADS --s3-max-upload-parts=$MAX_PARTS --s3-chunk-size=$CHUNK_SIZE --s3-upload-cutoff=$CHUNKED_UPLOADS_CUTOFF --max-size=$MAX_FILE_SIZE --s3-disable-checksum=$DISABLE_CHECKSUM --s3-use-multipart-etag=$USE_MULTIPART_ETAG --s3-endpoint $SOURCE_URL copy minio:$SOURCE_BUCKET_NAME $TARGET_LOCAL_FOLDER
```

Illustration of the current behavior after data is restored from a MinIO bucket:
```
/ $ ls -lant /home/jovyan/python-for-data-science-knowledge-tests/tests-20210118/test-for-
bartek/
total 804
drwxr-xr-x    4 65534    65534          200 Jun 13 09:08 .
drwxr-xr-x    6 65534    65534           88 Jun 13 09:08 ..
drwxr-xr-x    2 65534    65534          163 Jun 13 09:08 .ipynb_checkpoints
drwxr-xr-x    2 65534    65534           96 Jun 13 09:08 input
-rw-r--r--    1 65534    65534       412622 Apr 27 11:53 SOLUTION-FOR-stacked-regressions-top-4-on-leaderboard.ipynb
-rw-r--r--    1 65534    65534       407205 Apr 27 11:50 EXAM-TEMPLATE-stacked-regressions-top-4-on-leaderboard.ipynb
/ $ date
Mon Jun 13 09:27:26 UTC 2022
```

#### How do you think rclone should be changed to solve that?
[...]

Issue #1669, opened 2017-09-10
Sync directory mtimes (folder modtime)


rodalpho:
Per https://forum.rclone.org/t/change-directory-mtime-on-gdrive/3797

I’m syncing a bunch of backup directories to an encrypted GDrive.

The contents of those directories frequently change, and on my desktop this also changes the directories’ last modified time. So I can go in nautilus or finder and sort by date modified and see the directories with the newest files on top.

Rclone doesn’t sync directory mtimes, so they all have the mtime of when they were first synced to GDrive.

***ncw note*** when this gets done we will need to update `rclone touch` too.

reticentnz:
Hi @ncw, 

Thought I should mention that this used to (and may still) work on ACD, but doesn't seem to on Gdrive. I don't know if this was implemented server side (ie ACD updates the time of the folder when files are changed within it) or whether this was something rclone was pushing. In any case, I thought I should mention this in case you didn't know, and that some of the code/work required could be lifted from the ACD module.

It's caused some issues with applications that use this folder date as a quick way of checking whether files have been updated.

Looking forward to the fix.

Cheers

CottonEaster:
I too would find this feature useful, on Google Drive.

aatreya:
Looking forward to this as well. I have a bunch of archived data I'd like to transfer to Google Drive, but it's important that the directory modification times don't get changed, so I can't do that until this is supported. :(

Anything I can do to help with this effort?

ncw:
@aatreya what needs to be done is that an optional interface for setting directory times is added do all directory objects emitted by backends that are supported.  It is quite a big job unfortunately but not an impossible sized one if you want to help?

vicmarto:
I would find this feature useful too, but with the SFTP backend.

duaneg3:
This would be useful on the local backend as well ...

I have a version which works for local files:

1) add fs.DirDateHash variable  (fs/dir.go) and store date for every dir 
    (create AddDir() which call from backend/local/local.go::List())
 2) modify local/local.go::Mkdir to set file mod time  (it is an interface routine)
    (calls new lib/files/mkdir_windows.go::MkdirAll3())
 3) modify syn/syn.go::run() to modify the directory date using operations.Mkdir() 
     (just sets the modtime in this case)
     (similar location as --delete-empty-src-dirs)

[...]

PR #9064, opened 2025-12-27
s3: use LastModified from directory marker objects for directory modtime

datahorders:

This PR fixes directory timestamp handling in the S3 backend when listing directory marker objects.

**The Problem**

When S3-compatible backends return directory marker objects (zero-byte objects with keys ending in `/`) in ListObjectsV2 responses, these objects include a valid `LastModified` timestamp. However, the S3 backend was ignoring this timestamp and creating directory entries with Go's zero time value (`time.Time{}`), which displays as `2000-01-01 00:00:00` regardless of the actual timestamp returned by the server.

In `itemToDirEntry`, directory entries were unconditionally created with:

```go
d := fs.NewDir(remote, time.Time{}).SetSize(size)
```

This discards the `LastModified` field from the API response even when it contains a valid timestamp.

**The Fix**

```go
modTime := time.Time{}
if object.LastModified != nil {
    modTime = *object.LastModified
}
d := fs.NewDir(remote, modTime).SetSize(size)
```

If the object has a `LastModified` timestamp, use it. Otherwise, fall back to zero time (preserving existing behavior for backends that do not provide timestamps).

**Before the fix:**
```
$ rclone lsd s3-remote:bucket
           0 2000-01-01 00:00:00        -1 folder1
           0 2000-01-01 00:00:00        -1 folder2
           0 2000-01-01 00:00:00        -1 folder3
```

**After the fix:**
```
$ rclone lsd s3-remote:bucket
           0 2025-12-26 17:28:09        -1 folder1
           0 2025-12-18 13:46:18        -1 folder2
           0 2024-12-28 17:46:02        -1 folder3
```

**Background**

This was discovered while building an S3-compatible gateway that indexes directories as marker objects with proper filesystem mtimes. The S3 API correctly returned directory markers with `LastModified`:

```xml
<Contents>
  <Key>folder1/</Key>
  <LastModified>2025-12-26T17:28:09.470Z</LastModified>
  <Size>0</Size>
</Contents>
```

But `rclone lsd` showed `2000-01-01 00:00:00` for all directories.

**Compatibility**

This change is backward compatible:

- If `object.LastModified` is nil, behavior is unchanged (zero time used)
- If `object.LastModified` is set, the directory now uses that timestamp
- Standard AWS S3 does not typically return directory markers in `Contents`, so most users syncing with AWS will see no change
- S3-compatible backends that do return directory markers (MinIO, Ceph RGW, custom gateways, etc.) will now display correct timestamps

This is related to the broader topic of directory timestamp handling discussed in:

- #6242 - Preserve folder creation dates after rclone copy from S3 / MinIO buckets
- #1669 - Sync directory mtimes (folder modtime)

Those issues focus on writing directory timestamps during sync operations. This PR addresses a simpler problem: reading timestamps that already exist on directory marker objects returned by the server.

No prior issue was opened for this specific bug since the fix is straightforward.

#### Checklist

- [x] I have read the contribution guidelines
- [x] I have added tests for all changes in this PR if appropriate
- [x] I have added documentation for the changes if appropriate
- [ ] All commit messages are in house style
- [x] I'm done, this Pull Request is ready for review

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -1707,6 +1707,10 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 		f.features.WriteDirSetModTime = true
 		f.features.ReadDirMetadata = true
 		f.features.UserDirMetadata = true
+	} else {
+		// Disable directory metadata features when DirectoryMarkers is not set
+		f.features.MkdirMetadata = nil
+		f.features.DirSetModTime = nil
 	}
 	// f.listMultipartUploads()
 	if !opt.UseMultipartUploads.Value {


ncw:
Looks very nice - thank you.

I ran the CI.

Did you try the integration tests after the change?

Ideally we'd implement the backend method Mkdirmetadata too

https://github.com/rclone/rclone/blob/7c1660214d68c10f622951416148b7616e55bed1/fs/features.go#L91
 
That's how timestamps gets set on directories in rclone.

datahorders:
Yes, I ran the integration tests against MinIO with directory_markers enabled - all passing now.

I've implemented MkdirMetadata along with full directory metadata support:

  - New Directory struct implementing fs.Directory, fs.Metadataer, fs.SetMetadataer, and fs.SetModTimer
  - MkdirMetadata creates directory markers with metadata (stores mtime in x-amz-meta-mtime)
  - DirSetModTime for setting directory modification times
  - Updated itemToDirEntry to return the new Directory type when DirectoryMarkers is enabled
  - Enabled the WriteDirMetadata, ReadDirMetadata, WriteDirSetModTime, and UserDirMetadata feature flags

Passing tests:
  - FsMkdirMetadata
  - ReadDirMetadata
  - WriteDirMetadata
  - WriteDirSetModTime
  - FsDirSetModTime

Let me know if you need any other testing.

acme:
I was quite excited about this pull request until I realised that it only works if the content was uploaded to S3 with the `directory_markers` rclone feature enabled. Unfortunately that is off by default so my remotes don't have directory markers.

Issue #8300, opened 2025-01-11
rclone sync files with identical size again and again


geekifan:
#### What is the problem you are having with rclone?

I have two backends (both seafile). I want to sync files from backend SeafileA to backend SeafileB. But I find that every time I run `rclone sync SeafileA:/src SeafileB:/dst`, some files that have already been synchronized to B are synchronized again, even though the sizes of these files are exactly the same (I checked with `rclone size`, and the number of bytes is exactly the same).

#### What is your rclone version (output from `rclone version`)
rclone v1.68.2
- os/version: arch (64 bit)
- os/kernel: 6.8.12-4-pve (x86_64)
- os/type: linux
- os/arch: amd64
- go/version: go1.23.3
- go/linking: dynamic
- go/tags: none

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
ArchLinux x64

#### Which cloud storage system are you using? (e.g. Google Drive)
Seafile

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
rclone sync SeafileA:/src SeafileB:/dst -P --bwlimit off:4M

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)

Some details are omitted for privacy reasons
```
2025/01/11 17:34:08 INFO  : Starting bandwidth limiter at off:4Mi Byte/s
2025/01/11 17:34:08 DEBUG : rclone: Version "v1.68.2" starting with parameters ["rclone" "sync" "SeafileA:/src" "SeafileB:/dst" "-P" "--bwlimit" "off:4M" "-vv" "--log-file" "debug.log"]
2025/01/11 17:34:08 DEBUG : Creating backend with remote "SeafileA:/src"
2025/01/11 17:34:08 DEBUG : Using config file from "/root/.config/rclone/rclone.conf"
2025/01/11 17:34:08 DEBUG : Seafile server version 11.0.14
2025/01/11 17:34:08 DEBUG : Creating backend with remote "SeafileB:/dst"
2025/01/11 17:34:08 DEBUG : Seafile server version 11.0.16
2025/01/11 17:34:09 DEBUG : A.docx: Sizes identical
2025/01/11 17:34:09 DEBUG : crypto.pptx: Sizes identical
2025/01/11 17:34:09 DEBUG : seafile library 'SeafileB' path 'dst': Waiting for checks to finish
2025/01/11 17:34:09 DEBUG : B.svg: Sizes identical
[...]

PR #9066, opened 2025-12-27
backend/seafile: add modification time support - fixes #8337

### LINKED ISSUE #8337: Seafile Modification Time Support
Seafile recently released version 12.0, and it includes a small but notable change: [the file upload API](https://seafile-api.readme.io/reference/post_seafhttp-upload-api-upload-token-ret-json-1) now supports setting the modification time (mtime). Is this change worth integrating into Rclone to enable Seafile to support mtime, thereby bringing more robust synchronization and new features like bisync?

Image

clocken:

This adds support for preserving file modification time on upload via the API change introduced in 12.0

See feature request #8337

#### Checklist

### DIFF
--- a/backend/seafile/object.go
+++ b/backend/seafile/object.go
@@ -97,7 +97,7 @@ func (o *Object) Update(ctx context.Context, in io.Reader, src fs.ObjectInfo, op
 			return err
 		}
 
-		uploaded, err := o.fs.upload(ctx, in, uploadLink, o.pathInLibrary)
+		uploaded, err := o.fs.upload(ctx, in, uploadLink, o.pathInLibrary, src.ModTime(ctx))
 		if err == ErrorInternalDuringUpload {
 			// This is a temporary error, try again with a new upload link
 			continue
--- a/backend/seafile/seafile.go
+++ b/backend/seafile/seafile.go
@@ -494,8 +494,7 @@ func (f *Fs) String() string {
 
 // Precision of the ModTimes in this Fs
 func (f *Fs) Precision() time.Duration {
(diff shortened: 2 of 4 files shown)


roucc:
Do the seafile integration tests pass and is this supported by all versions of seafile? Thanks!

comqas:
Hit the exact problem this fixes, from the read side rather than the write side (so worth noting alongside #8300, which turned out to be a different root cause).

Case: a one-way `rclone sync seafile:lib/dir localdir` used to keep a local mirror current before reading a file's content. A real edit changed a file's text but happened to leave the byte size identical (net-zero: one character added on one line, one removed on another). Plain `sync` — no flags — reported "nothing to transfer" and kept serving the stale local copy; `--checksum` made no difference either, since with `Hashes()` returning `hash.None` it silently falls back to `--size-only`, same outcome. With `Precision()` returning `ModTimeNotSupported`, size is genuinely the *only* signal `sync` has to go on here, so a same-size edit is invisible to it by construction, not an edge case.

Worked around it locally with `--ignore-times` (forces every file through every run — correct, but expensive against a large directory), but that's clearly just papering over the same gap this PR closes properly.

Server is a self-hosted `seafileltd/seafile-mc:13.0-latest`, so the API capability this depends on is already there. Happy to test this branch against a real instance if that's useful for getting it un-stuck — it's been open since December with no review yet.

mdashlw:
I can confirm this PR works fine with Seafile 13. I’ve been using it for a while. I don’t care to test older versions, and maybe that’s a requirement for getting it merged, I don’t know.

PR #9079, opened 2026-01-05
oracleobjectstorage, sftp: eliminate unnecessary heap allocation

### LINKED ISSUE #9078: small refactoring to reduce heap allocation

This change is small and easy to apply I think.

v1.72.1

Removing unnecessary heap allocation.

Just move the location to declare variables.

1. 3 variables in `WaitForStateContext`

https://github.com/rclone/rclone/blob/28c187b9b49a9785a7f47a802c74e8a8aedf11c6/backend/oracleobjectstorage/waiter.go#L63-L64
https://github.com/rclone/rclone/blob/28c187b9b49a9785a7f47a802c74e8a8aedf11c6/backend/oracleobjectstorage/waiter.go#L87

Currently, these variables will be allocated in heap, which consumes more memory. They are all used in `go func()` but not used in other places. So they can be declared in `go func()`.

2. change `=` to `:=`, which is a tiny change. 

https://github.com/rclone/rclone/blob/28c187b9b49a9785a7f47a802c74e8a8aedf11c6/cmd/serve/sftp/connection.go#L294

Because of `=`, the assignment op, `err` variable declared in `handleChannel` will be allocated in heap, which is unnecessary. Change `=` to `:=` to declare `err` in the function literal will avoid allocating it in heap. This is really a small change but it does work. 😂

#### How to use GitHub

Lslightly:
Move the declaration location of variables to eliminate heap
allocation which may make rclone faster and reduce memory usage slightly.

Fixes #9078

Eliminate unnecessary heap allocation.

See #9078.

#### Checklist

### DIFF
--- a/backend/ftp/ftp.go
+++ b/backend/ftp/ftp.go
@@ -901,7 +901,6 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
 		}
 		resultchan <- result
 	}(c)
-	c = nil
 
 	// Wait for List for up to Timeout seconds
 	timer := time.NewTimer(f.ci.TimeoutOrInfinite())


backend/ftp/ftp.go
```diff
@@ -892,15 +892,15 @@ func (f *Fs) List(ctx context.Context, dir string) (entries fs.DirEntries, err e
...
+	go func(c *ftp.ServerConn) {
 		result, err := c.List(f.dirFromStandardPath(path.Join(f.root, dir)))
 		f.putFtpConnection(&c, err)
 		if err != nil {
 			errchan <- err
 			return
 		}
 		resultchan <- result
-	}()
+	}(c)
```
Lslightly, line 903: A extreme solution might be adding `c = nil` or `drop(&c)` to move the ownership of `c` to goroutine. But the lifespan of `c` still depends on GC. So adding `c = nil` manually does not make much sense I think.

roucc (approved):
Thanks, I'll merge now!

Issue #8870, opened 2025-10-03
Can't sync with Proton Drive


julianfairfax:
#### The associated forum post URL from `https://forum.rclone.org`
https://forum.rclone.org/t/proton-drive-unable-sync-copy-anything/52556

#### What is the problem you are having with rclone?
I can't sync with Proton Drive, folders are created but files aren't uploaded

#### What is your rclone version (output from `rclone version`)
rclone v1.71.1

#### Which OS you are using and how many bits (e.g. Windows 7, 64 bit)
fedora 42 (64 bit)

#### Which cloud storage system are you using? (e.g. Google Drive)
Proton Drive (protondrive)

#### The command you were trying to run (e.g. `rclone copy /tmp remote:tmp`)
`rclone sync ~/Test protondrive:/Test -vv --protondrive-replace-existing-draft=true`

#### A log from the command with the `-vv` flag (e.g. output from `rclone -vv copy /tmp remote:tmp`)
```
2025/10/03 09:45:32 DEBUG : rclone: Version "v1.71.1" starting with parameters ["rclone" "sync" "/home/user/Test" "protondrive:/Test" "-vv" "--protondrive-replace-existing-draft=true"]
2025/10/03 09:45:32 DEBUG : Creating backend with remote "/home/user/Test"
2025/10/03 09:45:32 DEBUG : Using config file from "/home/user/.config/rclone/rclone.conf"
2025/10/03 09:45:32 DEBUG : Creating backend with remote "protondrive:/Test"
2025/10/03 09:45:32 DEBUG : protondrive: detected overridden config - adding "{kpshc}" suffix to name
2025/10/03 09:45:32 DEBUG : proton drive root link ID 'Test': Has cached credentials
2025/10/03 09:45:35 DEBUG : proton drive root link ID 'Test': Used cached credential to initialize the ProtonDrive API
2025/10/03 09:45:36 DEBUG : fs cache: renaming cache item "protondrive:/Test" to be canonical "protondrive{kpshc}:Test"
2025/10/03 09:45:36 DEBUG : Test.txt: Need to transfer - File not found at Destination
2025/10/03 09:45:36 DEBUG : proton drive root link ID 'Test': Waiting for checks to finish
2025/10/03 09:45:36 DEBUG : proton drive root link ID 'Test': Waiting for transfers to finish
[...]

PR #9080, opened 2026-01-06
protondrive: implement shouldRetry instead of always returning false

tomholford:
## Summary
Implement `shouldRetry`, replacing the `return false, err` stub.

## Problem
`shouldRetry` returned `false` unconditionally, which made protondrive the only rclone backend that disables pacer-level retries entirely. Genuine transport-level transients (TCP resets, brief 5xx, network blips) silently went un-retried.

Note: this is **not** the cause of the recent Proton Drive upload failures tracked in #8870 — that was the missing block-verification step (now fixed by @tdawe1's merge). See the discussion thread on this PR for context.

## Changes
- Implement `shouldRetry` following the pattern used by box, swift, dropbox, etc.
- Use `errors.As` to unwrap `*proton.APIError` (typed-error approach)
- Retry transient storage block errors (`Code=200501`)
- Retry server errors (5xx, except 503)
- Skip 429 and 503 — handled inside `go-proton-api`'s resty layer via `catchTooManyRequests` / `catchRetryAfter`, which honours `Retry-After`
- Fall back to `fserrors.ShouldRetry` for non-API errors

## Test plan
- [x] `golangci-lint` passes
- [x] `go build` and `go test ./backend/protondrive/...` pass
- [x] Table-driven unit tests cover: nil, cancelled context, typed `*proton.APIError` for `Code=200501` / 5xx / 4xx, wrapped errors via `errors.As`, non-API fallback
- [x] Manual upload to a real Proton Drive account: retry path triggers as expected on transient `Code=200501`

## Notes
Reviewed `shouldRetry` across other backends (box, swift, dropbox, ftp, putio, filescom). This implementation follows the same shape: context check, typed-error decisions, fallback to `fserrors.ShouldRetry`.

### DIFF
--- a/backend/protondrive/protondrive.go
+++ b/backend/protondrive/protondrive.go
@@ -22,6 +22,7 @@ import (
 	"github.com/rclone/rclone/fs/config/configmap"
 	"github.com/rclone/rclone/fs/config/configstruct"
 	"github.com/rclone/rclone/fs/config/obscure"
+	"github.com/rclone/rclone/fs/fserrors"
 	"github.com/rclone/rclone/fs/fshttp"
 	"github.com/rclone/rclone/fs/hash"
 	"github.com/rclone/rclone/lib/dircache"
@@ -250,8 +251,31 @@ type Object struct {
 
 // shouldRetry returns a boolean as to whether this err deserves to be
 // retried.  It returns the err as a convenience
+//
+// 429 rate-limit and 503 responses are retried inside go-proton-api's resty
+// layer (see catchTooManyRequests / catchRetryAfter) using the Retry-After
+// header, so we do not retry them again here.
 func shouldRetry(ctx context.Context, err error) (bool, error) {
-	return false, err
+	if fserrors.ContextError(ctx, &err) {
+		return false, err
+	}
+	if err == nil {
+		return false, nil
+	}
+	var apiErr *proton.APIError
+	if errors.As(err, &apiErr) {
+		// Transient storage block error ("Please retry").
+		if apiErr.Code == 200501 {
+			fs.Debugf(nil, "Retrying storage block error: %v", err)
+			return true, err
+		}
+		// Server errors. 503 is already handled by the SDK; retry the rest.
+		if apiErr.Status >= 500 && apiErr.Status < 600 && apiErr.Status != 503 {
+			return true, err
+		}
+		return false, err
+	}
+	return fserrors.ShouldRetry(err), err
 }
 
 //------------------------------------------------------------------------------
--- a/backend/protondrive/protondrive_internal_test.go
+++ b/backend/protondrive/protondrive_internal_test.go
@@ -1,8 +1,14 @@
 package protondrive
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"regexp"
 	"testing"
+
+	"github.com/rclone/go-proton-api"
+	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


ncw:
Looking good so far - carry on please :-)

tomholford:
> Looking good so far - carry on please :-)

Thank you, just did a manual test and it worked as expected. Marked as ready for review 🤞  Next going to take a crack at the upstream issue in the Proton API Bridge

the-narwhal:
Hey @tomholford just wanted to flag a couple of things in case it helps get this moving.

 The PR now has a conflict: `ae5d388` (merged yesterday) added content to `protondrive_internal_test.go`, which this PR also targets.

 One other thing worth knowing: `go-proton-api` already handles 429 internally at the resty level via `catchRetryAfter`/`catchTooManyRequests`. Would that make the `Status=429` check in `shouldRetry` redundant? 
 
 Also, since `proton.APIError` is a proper typed error with `Code` and `Status` fields, you could use `errors.As` to unwrap it directly instead of string matching, something like:
 
```go
 var apiErr *proton.APIError
 if errors.As(err, &apiErr) {
     if apiErr.Code == 200501 { return true, err }
     if apiErr.Status >= 500  { return true, err }
     return false, err
 }
 return fserrors.ShouldRetry(err), err
 ```

That said, the string-matching approach works fine too, just flagging it as an option. Either way, the core fix is solid and worth getting merged.

ncw:
@tomholford have you seen this cause a problem in the wild or is it more of a theoretical problem?

If you've seen an actual problem in the wild, then adjust the commit message to say `protondrive: fix lack of retries on XXX causing YYY`

tomholford:
@ncw :

Real-world. Hit it during automated nightly backups to Proton Drive — wrote up the full picture in https://github.com/rclone/rclone/issues/8870#issuecomment-3980917915, including the 1.39 GiB Gitea backup that finally went through once shouldRetry, block verification, and draft cleanup were all in place. Other users in that same thread show identical `Code=200501` log lines.

Worth flagging: now that @tdawe1's block-verification patches are merged, the dominant cause of upload failure is fixed, so this PR is the defense-in-depth piece — it stops a single transient `422 Code=200501 ("Please retry")` mid-transfer from killing an otherwise-healthy upload.

Reworded the commit subject and force-pushed. Same push also addresses @the-narwhal's review: `shouldRetry` now uses `errors.As` with `*proton.APIError` instead of string matching, redundant `Status=429` branch dropped (handled by go-proton-api's resty layer via `catchTooManyRequests`/`catchRetryAfter`), and the test conflict from `ae5d388` is resolved.

tomholford:
@ncw:

After thinking about this some more, I want to walk back part of my previous comment. The `Code=200501` storm I was hitting was almost certainly the block-verification bug surfacing as Proton's generic "Please retry" code, not transport-level transient errors. Two signals point that way: user adriy-be in #8870 tried `RCLONE_LOW_LEVEL_RETRIES=20` and it didn't help (transient errors would've been masked by that), and uploads recovered for everyone right after and tdawe1's merge with no retry-related changes. So in my own case retries probably wouldn't have made a difference — I conflated the symptom (a 422 telling clients to retry) with the cause (a protocol-level rejection).

That said, the stub is still worth fixing. `return false, err` makes protondrive the only rclone backend that disables pacer-level retries entirely; every other backend (box, swift, dropbox, ftp, …) at minimum falls back to `fserrors.ShouldRetry(err)` so genuine transport-level transients (TCP resets, brief 5xx) get retried. This PR just brings protondrive in line with that pattern.

Retitled the commit to match: `protondrive: implement shouldRetry instead of always returning false`.

ncw (approved):

Will merge now.

Issue #3770, opened 2019-11-28
Partial download extension


opsoyo:
#### What is your current rclone version (output from `rclone version`)?
1.50.2

#### What problem are you are trying to solve?
Partial download extension, like in [Aria](https://aria2.github.io/)

#### How do you think rclone should be changed to solve that?

1. Adding an argument to allow a custom partial extension (eg. --partial-extension=part),
2. or a bool argument to enable the partial extension (eg. --partial-extension),
3. or a permanent partial extension hardcoded.

This was discussed in the [forum](https://forum.rclone.org/t/adding-extension-to-partial-file/6346), but I can't find the issue created to develop it.

My personal situation utilizes mergerfs to join a local writable drive with a readonly version of my crypt. This allows wget/curl/aria/etc to skip completed transfers which have already been sent to cloud storage or have already been downloaded locally. Simultaneously, I have an rclone systemd script reading from that local drive only to process the uploading task. This provides a fairly seamless experience when reading or writing data, unless it's readonly on the cloud.

I'm proposing a .part / .incomplete / .unconfirmed to allow separate upload instances of rclone to ignore files which have not finished downloading locally.

ncw:
I think what you are asking is that when rclone is doing a transfer instead of writing direct to `filename` it writes to `filename.partial` then renames `filename.partial` to `filename` at the end of the transfer.

Taking a cue from rsync, what it does is make a `filename~123123` file if I remember rightly, then it renames that to `filename` at the end of the sync.  Perhaps an extension like `.partial.123123` might be the best way - you could still filter on that.

Thoughts
- not all backends can rename files
- most backends the upload and replacement of a file is atomic (eg you can't have a half written file)
    - notable exceptions - local, sftp
   - so this is only really useful with local or sftp
- adding things to the file name means rclone is more likely to hit path name limits

Rsync has an option called `--partial` which if you set it it means that it won't delete partially uploaded files on error.  rclone could potentially have this too.

Rsync also has an option called `--partial-dir` which is a directory that files get uploaded into before being moved into the correct place.

What do you think?

opsoyo:
> I think what you are asking is that when rclone is doing a transfer instead of writing direct to `filename `it writes to `filename.partial` then renames `filename.partial` to `filename` at the end of the transfer.

You've read my mind! 

> not all backends can rename files...

[...]

PR #9092, opened 2026-01-11
add --partial-dir option for partial file storage

davidvella:

Add --partial-dir flag to store partial files in a separate directory during transfer, similar to rsync's --partial-dir option. 

Use Case: We use fsnotify to listen for new files in a directory that is synced by rclone. With the current behavior, partial files are written to the destination directory with a .partial suffix. As these files are being written, fsnotify generates a large number of events which consumes significant CPU resources. With --partial-dir, partial files are written to a separate directory that we don't monitor. We only receive fsnotify events when the completed file is moved to the final destination, significantly reducing CPU usage and eliminating noise from in-progress transfers. 

Example:
`rclone sync remote:bucket /data --partial-dir /tmp/rclone-partial`

This stores partial files at /tmp/rclone-partial/path/to/file.HASH.partial during transfer, then moves them to /data/path/to/file on completion.

I think it was previously discussed [here](https://github.com/rclone/rclone/issues/3770) but wasn't picked up.

#### Checklist


fs/config.go
```diff
@@ -550,6 +550,11 @@ var ConfigOptionsInfo = Options{{
 	Default: ".partial",
 	Help:    "Add partial-suffix to temporary file name when --inplace is not used",
 	Groups:  "Copy",
+}, {
+	Name:    "partial_dir",
+	Default: "",
+	Help:    "put a partially transferred file into DIR",
```
albertony: Two minor comments here:

1. Should start with capital letter.
2. I would change from "DIR" to something else, e.g. "this directory" or something else. I see that existing option `--backup-dir` also uses "DIR", but I think that comes from a time when it was listed as `--backup-dir DIR`  and then it makes sense, while now it is listed as `--backup-dir string` (after inconsistencies in how we describe argument types was cleaned up).

davidvella: Done - changed to 'Directory to store partial files during transfer'

davidvella:
@albertony Just checking if there’s anything else needed to get this merged, or if you’re waiting on something from my end

albertony:
Not waiting for you, no. Waiting for someone to do a proper full review. But unfortunately there is a long backlog...

ncw:
@davidvella I had a look through the code - looks very nice!

Just thinking about the decision as to store the partial using the complete path. This means that the partial directory will accumulate empty directories which I think the user probably won't be expecting.

Though I just read the docs for `rsync` and it does exactly the same thing!

So we should probably add this to the docs. Something like

> Rclone uses `--partial-dir` as a scratch space and may leave empty directories behind once files have been transferred. If you want to remove empty directories use `rclone rmdir` to remove them.

The docs should also probably note 

> The partial directory might need to be on the same file system as files need to be moved from the partial directory to the destination.

The docs should also note 

> If the partial directory is inside the sync tree then it should be excluded from the sync with a filter otherwise the sync will misbehave.

Thanks

davidvella:
> > The partial directory might need to be on the same file system as files need to be moved from the partial directory to the destination.



> @davidvella I had a look through the code - looks very nice!
> 
> Just thinking about the decision as to store the partial using the complete path. This means that the partial directory will accumulate empty directories which I think the user probably won't be expecting.
> 
> Though I just read the docs for `rsync` and it does exactly the same thing!
> 
> So we should probably add this to the docs. Something like
> 
> > Rclone uses `--partial-dir` as a scratch space and may leave empty directories behind once files have been transferred. If you want to remove empty directories use `rclone rmdir` to remove them.
> 
> The docs should also probably note
> 
> > The partial directory might need to be on the same file system as files need to be moved from the partial directory to the destination.
> 
> The docs should also note
> 
> > If the partial directory is inside the sync tree then it should be excluded from the sync with a filter otherwise the sync will misbehave.
> 
> Thanks

Thanks for the feedback! @ncw I've updated the docs to include:

- Empty directories being left behind and using `rclone rmdirs` to clean them up (note: changed from `rmdir` since `rmdir` expects the directory to already be empty, whereas `rmdirs` recursively removes empty directories)
- Same backend requirement for the move operation
- Filter exclusion if partial directory is inside the sync tree

I also clarified that `--partial-dir` is relative to the destination root. As the code currently stands in `copy.go:116`, it uses `path.Join(c.ci.PartialDir, remoteForCopy)` which creates a relative path within the destination backend - so absolute paths like `/tmp/partial` end up being interpreted as a path relative to the destination rather than an actual absolute filesystem path.

As things stand this is okay for my use case. However, as it stands its doesn't support absolute paths to align with rsync's behavior. Would that be worth implementing, or is keeping it relative to the destination sufficient?

davidvella:
Hi, @ncw following up on this. I think I've addressed the previous comments. Happy to adjust further.

ncw:
I think paths relative to the transfer root is the rclone way.

I suppose it could also be a full backend specifier like --backup-dir which would remove the ambiguity and bring it into line with --backup-dir and friends.

This makes the move logic slightly more complicated.

What do you think? Which is most rclone/ rsync?

davidvella:
@ncw Thanks for the suggestion. I've thought about it, but I think I'd prefer to keep the current simpler approach where it's relative to the destination. 
Supporting full backend specifiers would require adding validation logic to check it's on the same backend (like --backup-dir does), which adds complexity. The current code is straightforward - it just uses path.Join within the destination Fs. 
I don't think the behaviour has to match rsync exactly. As long as the documentation is clear that --partial-dir is relative, I think that's sufficient. I've updated the docs to make this explicit. 
Happy to revisit if there's a strong use case for absolute paths, but for now I think the simplicity is worth the trade-off.

davidvella:
Hi, @ncw following up on this. I think it is okay to merge?

davidvella:
Hi, @ncw following up

ncw:
Hi @davidvella - sorry I didn't have time to merge this for 1.74

I've been turning over in my mind the absolute vs relative path thing and I think it should take a full backend specifier like `--backup-dir`/`--compare-dest`/`--copy-dest` do.

1. The PR's own example doesn't behave the way it reads which says to me users are going to find this confusing. The description says:

> `rclone sync remote:bucket /data --partial-dir /tmp/rclone-partial`
> This stores partial files at `/tmp/rclone-partial/path/to/file.HASH.partial`...

But with the current implementation, the partials actually land at `/data/tmp/rclone-partial/...` - inside the destination tree. That's almost certainly not what a user typing `/tmp/rclone-partial` expects, and it directly defeats the use case of avoiding an fsnotify storm on the synced tree.

2. It eliminates a real UX difficulty. With relative-to-root, anything inside the sync tree breaks the sync unless the user remembers to add a filter rule. The docs already have to warn about this.

3. Consistency. Rclone already has a clear convention for "directory of files used during sync" - `--backup-dir`, `--compare-dest`, `--copy-dest` are all backend specifiers. Rclone has enough flags that consistency is very important.

4. The implementation cost isn't bad. `operations.BackupDir()` (fs/operations/operations.go:1899) already does the same-config / overlap / can-server-side-move validation we'd need. The main change is the move call at copy.go:357 - `c.dstFeatures.Move(ctx, newDst, c.remote)` would become `operations.Move(ctx, c.f, ..., partialFs, ...)`, which already handles same-backend-different-Fs moves. That's maybe 30-40 extra lines plus a `partialDir fs.Fs` field on the copy struct, and we get all the validation for free.

If we ship the relative-to-root version now and try to retrofit later, we either break behavior or end up supporting both modes forever which would be unfortunate.

What do you think?

PR #9095, opened 2026-01-13
docs: fix googlephotos custom client_id instructions

hyusap:
 
The Google Photos "Making your own client_id" documentation was incomplete. It referenced the Google Drive guide but only mentioned changing the OAuth scopes, without noting that users must also enable a different API.
 
This change clarifies the two key differences from the Google Drive setup:
1. At step 3, enable the "Photos Library API" instead of the "Google Drive API"
2. At step 5, use the Google Photos specific scopes instead of the Drive scopes
 
Before this fix, users following the documentation would get errors because the Photos Library API was not enabled in their Google Cloud Console project.
 
No prior discussion - this was a documentation gap discovered while following the setup instructions.
 
#### Checklist
 
### DIFF
--- a/docs/content/googlephotos.md
+++ b/docs/content/googlephotos.md
@@ -659,8 +659,14 @@ second that each client_id can do set by Google.
 If there is a problem with this client_id (eg quota too low or the
 client_id stops working) then you can make your own.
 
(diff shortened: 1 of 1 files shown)


ncw:
Thank you

PR #9100, opened 2026-01-15
webdav: set both types of hashes if available

WeidiDeng:

Set both type of hashes if the backend webdav server supports.

I have noticed sometimes rclone will report `hashes could not be checked` warning even though these files used to have one for my owncloud compatible webdav server.

The problem is that when updating modified time, rclone will append checksums in the `PROPPATCH` request. However, at most one checksum will be added.

For owncloud, this means only sha1 hash will be present. But when checking files between 2 file systems, rclone will only choose one common hash to compare in addition to the usual size and modified time. For local file system and remote webdav server, md5 is chosen. But because rclone removed md5 when setting the modified time, it will report `hashes could not be checked`.

No.

#### Checklist

### DIFF
--- a/backend/doi/doi.go
+++ b/backend/doi/doi.go
@@ -77,7 +77,7 @@ The DOI provider can be set when rclone does not automatically recognize a suppo
 			Name: "doi_resolver_api_url",
 			Help: `The URL of the DOI resolver API to use.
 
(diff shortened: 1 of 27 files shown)


WeidiDeng:
@ncw Do you have time to review this pr?

WeidiDeng:
@ncw Do you have time to take a look?

PR #8537, opened 2025-05-04
Filen Backend

Enduriel:
#### What is the purpose of this change?

Add Support for [Filen](https://filen.io/)

#### Was the change discussed in an issue or in the forum before?

Discussed in issue #6728

#### Known (potential) Issues

- The backend still fails a single test `TestBisyncRemoteRemote/extended_filenames`, I've not been able to figure out why, so I'd appreciate some help with debugging this. [Here's](https://github.com/user-attachments/files/20026962/2025-05-04-101750.zip) a zip of the report of my latest run.
- I'm not 100% clear on whether I can integrate with rclone's standard multithreading approach. The way filen uploads work is that uploads are always made in 1 MiB chunks, this didn't seem to integrate well with the multithreading assumptions rclone makes (especially because this means non-multithreaded speeds are awful), so there's currently a somewhat different multithreading system which can be configured as specified in the documentation.

#### Other Questions
- In the longer term, we're considering transitioning to a unified rust SDK for all our products, I saw there was some discussion on whether it'd be feasible to use an underlying rust library for a backend [here](https://forum.rclone.org/t/would-calling-gpl-licensed-rust-be-acceptable/48365/3), and I just wanted to clarify if you guys think it would be possible as a long term goal to potential use that.

#### Checklist

- [x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).
- [x] I have added tests for all changes in this PR if appropriate.
- [x] I have added documentation for the changes if appropriate.
- [x] All commit messages are in [house style](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#commit-messages).
- [ ] I'm done, this Pull Request is ready for review :-)

PR #9102, opened 2026-01-16
filen: swap to blake3 hashes

Enduriel:

After adding the filen backend to rclone in #8537, and with people testing our fork finding odd hash mismatches, yesterday we identified an issue causing hash generation using our TypeScript SDK to have a potential race condition meaning that (since we don't add any client fingerprinting to uploads) all old hashes are potentially wrong, and therefore useless.

This change updates the Go SDK version to a version that uses blake3 for hashing (since it seems to be better for this use-case anyway), I will be doing the same work for our Rust SDK which all our clients will soon be migrating to.

We only identified this now because the TypeScript SDK never actually validated file hashes on download, whereas the newer SDKs do. File integrity hasn't been compromised here, the issue is just that we can't trust the hash mismatches.

Tests run on the latest version: [2026-01-15-163419.zip](https://github.com/user-attachments/files/24667146/2026-01-15-163419.zip)

No

#### Checklist

### DIFF
--- a/backend/filen/filen.go
+++ b/backend/filen/filen.go
@@ -226,7 +226,7 @@ func (f *Fs) Precision() time.Duration {
 
 // Hashes returns the supported hash sets.
 func (f *Fs) Hashes() hash.Set {
-	return hash.Set(hash.SHA512)
+	return hash.Set(hash.BLAKE3)
 }
 
 // Features returns the optional features of this Fs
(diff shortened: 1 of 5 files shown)


Enduriel:
@ncw, sorry. I forgot there are places in the docs that list the relevant hashes for the backend. Should be updated everywhere now.

Unfortunately since we are a fully encrypted cloud storage provider, we cannot calculate the hashes server side, this means every hash we want to see when listing has to be computed on the client before upload. This means we cannot support an arbitrary number of hashes, and because there's potentially some cool stuff we could do if we know that no two files with the same hash can be identical, we need a stronger guarantee than what MD5 provides.

PR #9106, opened 2026-01-18
sftp: properly escape quotes in PowerShell

### LINKED ISSUE #9098: `copy` fails with SFTP remote that uses PowerShell when paths contain unicode single quotes
I ran into problems using `copy` with an SFTP Windows remote that used PowerShell when filenames contain unicode single quotes. It fails at the hashing stage after copying the file. This is with that remote as either the source or the destination. I tested `‘`, `’`, `‚`, and `‛`. I expect this to be a problem on other platforms as well, as long as PowerShell is the shell. 

I narrowed the cause down to [here](https://github.com/rclone/rclone/blob/9ec918f1374630e7d4614389ca013c0c0bd452a1/backend/sftp/sftp.go#L2060). Escaping like this is insufficient. PowerShell treats 5 characters (the 4 I listed and the usual `'`) as single quotes, as you can see [here](https://github.com/PowerShell/PowerShell/blob/1241ad2794cc84c18554e36ce658bfd86781cf8e/src/System.Management.Automation/engine/parser/CharTraits.cs#L267-L271).

I'm not sure why single quotes are used for quoting with PowerShell by rclone (it didn't seem to be explained in the original issue) but I suppose it doesn't really matter since PowerShell is cross-platform, so escaping has to be done either way. I *think* fixing this is as simple as updating the function I linked to to prefix all 5 kinds of single quote in one pass.

---
Just a note for testing or anything, I had a few more issues with `‛` but I know that's a known problem (#7456) so I used `--local-encoding None,Slash,Ctl,Del,Dot` as described there to work around those, but was still left with the issue I'm describing here.

#### How to use GitHub

benbryant0:
Fixes #9098

I'm not that experienced with Go so if this isn't done in the best way, please change it!

#### Checklist

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -2052,12 +2052,24 @@ func (o *Object) Hash(ctx context.Context, r hash.Type) (string, error) {
 	return hashString, nil
 }
 
+// This should match PowerShell's definition of single quotes
+// https://github.com/PowerShell/PowerShell/blob/1241ad2794cc84c18554e36ce658bfd86781cf8e/src/System.Management.Automation/engine/parser/CharTraits.cs#L267-L271
+var powerShellQuoteReplacer = sync.OnceValue(func() *strings.Replacer {
+	return strings.NewReplacer(
+		"'", "''",
+		"‛", "'‛",
+		"‘", "'‘",
+		"’", "'’",
+		"‚", "'‚",
+	)
+})
+
 // quoteOrEscapeShellPath makes path a valid string argument in configured shell
 // and also ensures it cannot cause unintended behavior.
 func quoteOrEscapeShellPath(shellType string, shellPath string) (string, error) {
 	// PowerShell
 	if shellType == "powershell" {
-		return "'" + strings.ReplaceAll(shellPath, "'", "''") + "'", nil
+		return "'" + powerShellQuoteReplacer().Replace(shellPath) + "'", nil
 	}
 	// Windows Command Prompt
 	if shellType == "cmd" {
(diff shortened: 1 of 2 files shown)


backend/sftp/sftp.go
```diff
@@ -2052,12 +2052,24 @@ func (o *Object) Hash(ctx context.Context, r hash.Type) (string, error) {
 	return hashString, nil
 }
 
+// This should match PowerShell's definition of single quotes
+// https://github.com/PowerShell/PowerShell/blob/1241ad2794cc84c18554e36ce658bfd86781cf8e/src/System.Management.Automation/engine/parser/CharTraits.cs#L267-L271
+var powerShellQuoteReplacer = sync.OnceValue(func() *strings.Replacer {
```
albertony, line 2057: (Why) is the sync.OnceValue wrapper needed?

albertony:
Wow, I had no idea, never seen this in documentation or other literature I've come across! Also, I find it quite surprising that Microsoft actually chose to do it this way! However, since they did, I agree it makes sense for rclone to escape them all. Thanks.

backend/sftp/sftp.go
benbryant0: Not needed. I assumed it'd be preferable not to allocate & instantiate the replacer until it was used, since this is defined at the package level. If I understand correctly, that would mean it was always allocated any time rclone ran at all, even though this is only needed in a very specific case.

benbryant0:
> Wow, I had no idea, never seen this in documentation or other literature I've come across! Also, I find it quite surprising that Microsoft actually chose to do it this way! However, since they did, I agree it makes sense for rclone to escape them all. Thanks.

Yup :( I went down a rabbit hole thinking all sorts of other things must've been the cause before I worked out it was this. It's extra weird since a unicode single quote will act as a closing quote for the usual ASCII one. They don't even need to match! So weird, but maybe there's a reason for it. I didn't look into it beyond solving it.

backend/sftp/sftp.go
albertony: Ok, I see. Not sure it is necessary, but let's wait for Nick or someone else stopping by for a proper review.

benbryant0:
Came back around to this to rebase it, and it looks like the exact same solution was implemented once it was discovered this was also a security issue: https://github.com/rclone/rclone/commit/e122fba1a57641b63a580aa26c026903a84e2e88

PR #9114, opened 2026-01-22
config: add --config-pass-cache flag for TTL-based password caching

Siddharth-magesh:
This PR adds an optional `--config-pass-cache` flag which enables a TTL-based,
in-memory cache for the decrypted config password when the password is supplied
via `--password-command`.

The feature is strictly opt-in and requires the user to explicitly provide a
valid duration. Existing behavior is unchanged unless the flag is set.

---

### Key Features

- Cache is in-memory only and is never written to disk
- Cache is scoped to the current rclone process
- TTL-based expiration using standard Go duration syntax
- Reduces repeated invocations of `--password-command` during long-running
  commands (e.g. mount, serve, chained operations)

---

### Security Considerations

- Caching is enabled only when `--password-command` is used
- No caching occurs when the password is provided via `RCLONE_CONFIG_PASS`
- Cache is automatically cleared when the TTL expires or the process exits
- No background daemon or persistent storage is introduced

---

To improve usability when using encrypted configs with `--password-command`
by allowing an explicit, time-limited, in-memory cache of the decrypted config
key, while preserving rclone’s default security behavior.

No. This PR is intentionally minimal and conservative in scope. I am happy to
open or link a follow-up issue if maintainers prefer prior discussion.

---

#### Checklist

- [x] I have read the contribution guidelines.
- [x] I have added tests for the changes in this PR.
- [x] I have added documentation for the changes.
- [x] This Pull Request is ready for review.
- [x] All commit messages follow house style.

### DIFF
--- a/backend/mega/mega.go
+++ b/backend/mega/mega.go
@@ -17,12 +17,10 @@ Improvements:
 
 import (
 	"context"
-	"crypto/tls"
 	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"path"
 	"slices"
 	"strings"
@@ -256,25 +254,7 @@ func NewFs(ctx context.Context, name, root string, m configmap.Mapper) (fs.Fs, e
 	defer megaCacheMu.Unlock()
 	srv := megaCache[opt.User]
 	if srv == nil {
-		// srv = mega.New().SetClient(fshttp.NewClient(ctx))
-
-		// Workaround for Mega's use of insecure cipher suites which are no longer supported by default since Go 1.22.
-		// Relevant issues:
-		// https://github.com/rclone/rclone/issues/8565
-		// https://github.com/meganz/webclient/issues/103
-		clt := fshttp.NewClient(ctx)
-		clt.Transport = fshttp.NewTransportCustom(ctx, func(t *http.Transport) {
-			var ids []uint16
-			// Read default ciphers
-			for _, cs := range tls.CipherSuites() {
-				ids = append(ids, cs.ID)
-			}
-			// Insecure but Mega uses TLS_RSA_WITH_AES_128_GCM_SHA256 for storage endpoints
-			// (e.g. https://gfs302n114.userstorage.mega.co.nz) as of June 18, 2025.
-			t.TLSClientConfig.CipherSuites = append(ids, tls.TLS_RSA_WITH_AES_128_GCM_SHA256)
-		})
-		srv = mega.New().SetClient(clt)
-
+		srv = mega.New().SetClient(fshttp.NewClient(ctx))
 		srv.SetRetries(ci.LowLevelRetries) // let mega do the low level retries
 		srv.SetHTTPS(opt.UseHTTPS)
 		srv.SetLogger(func(format string, v ...any) {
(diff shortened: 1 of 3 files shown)


ncw:
I'm pretty sure that if --password-command is set, then we call store it in memory anyway so I'm not sure of the purpose of this PR.

Please explain - thanks!

Siddharth-magesh:
When `--password-command` is used, the decrypted config password is held in memory only **transiently**, but the **password command is re-executed each time the config is loaded or decrypted**. This happens in paths that call config load/decrypt logic repeatedly (e.g. during `fs.NewFs`, remotes being reopened, or long-running commands like `mount` and `serve`).

This PR adds an **explicit, opt-in, TTL-bounded in-memory cache** for the *decrypted* config password **after the first successful invocation**, so subsequent config accesses reuse it within the same process.

From a security standpoint:

- Caching is **opt-in only** and applies **only to `--password-command`**
- The cache is **in-memory, process-scoped, and TTL-limited**
- It is cleared automatically on TTL expiry or process exit
- Default behavior is unchanged unless the flag is set

So while the password already exists in memory briefly, this change makes reuse **explicit, time-limited, and user-controlled**, while avoiding repeated command execution.

ncw:
In the source code

We read the password for --password-command with this code

https://github.com/rclone/rclone/blob/6b9f77459ae24b056a30591c3204531b48555c3d/fs/config/crypt.go#L181-L184

This is called from here

https://github.com/rclone/rclone/blob/6b9f77459ae24b056a30591c3204531b48555c3d/fs/config/crypt.go#L87-L88

But only if `configKey` is not set.

When the password is read we set config key

https://github.com/rclone/rclone/blob/6b9f77459ae24b056a30591c3204531b48555c3d/fs/config/crypt.go#L92-L94

Which sets configKey here

https://github.com/rclone/rclone/blob/6b9f77459ae24b056a30591c3204531b48555c3d/fs/config/crypt.go#L283

So I think the password is already cached.

If you disagree, please provide instructions on how to demonstrate the password isn't cached. This should be something I can replicate easily.

Siddharth-magesh:
I agree with your reading of the code, and you’re right that `configKey` is cached once `SetConfigPassword()` is called.

The distinction this PR is trying to make is that what’s cached today is the **derived `configKey`**, not the **result of `--password-command` itself**. If the process needs to decrypt config before `configKey` has been initialised (e.g. in flows where config is reloaded early or multiple times), `GetPasswordCommand()` may still be invoked again.

The intent here is to:

* Cache the **output of `--password-command` explicitly**, not just the derived key
* Make that caching **opt-in and TTL-bounded**, rather than implicit and unbounded for the process lifetime
* Avoid repeated executions of external password commands in long-running or multi-phase commands

That said, if `configKey` is in fact always initialised early enough to prevent repeated calls in all supported flows, then this PR may be redundant. I’m **happy to close it** if that’s the case, or to add a small reproducer (e.g. logging or a test that forces multiple config loads) if you think a specific scenario should already be covered.

ncw:
Do a test. Make a --password-command that logs something and see if you can get it to be called multiple times from a single rclone process.

Siddharth-magesh:
Confirmed via testing: password command is invoked exactly once per process. `configKey` caching in `SetConfigPassword()` and the dataLoaded gaurd in `LoadedData()` already prevent repeated invocations.
 
Closing as redundant - thanks for the clarification.

PR #9124, opened 2026-01-27
pikpak: add proxy_url option for download acceleration

nickhoo555:
## Summary
Add a new configuration option proxy_url for the PikPak backend that allows users to specify a reverse proxy URL for accelerating downloads.

## Problem
Some users experience slow direct download speeds from PikPak servers due to network conditions or geographic location.

## Solution
When proxy_url is configured, download URLs are routed through the specified proxy.

## Usage
Via command line, config file, or environment variable RCLONE_PIKPAK_PROXY_URL.

## Testing
Tested with Cloudflare Worker reverse proxy, confirmed working with significant speed improvement.

### DIFF
--- a/backend/pikpak/pikpak.go
+++ b/backend/pikpak/pikpak.go
@@ -222,6 +222,11 @@ Fill in for rclone to use a non root folder as its starting point.
 			Help:     "Use original file links instead of media links.\n\nThis avoids issues caused by invalid media links, but may reduce download speeds.",
 			Advanced: true,
 		}, {
+			Name:     "proxy_url",
+			Default:  "",
+			Help:     "Cloudflare Worker proxy URL for accelerating downloads.\n\nIf set, download URLs will be proxied through this URL.\nFormat: https://your-worker.your-subdomain.workers.dev\n\nThe proxy will receive requests like:\nhttps://your-worker.workers.dev/https://original-download-url",
+			Advanced: true,
+		}, {
 			Name:     "hash_memory_limit",
 			Help:     "Files bigger than this will be cached on disk to calculate hash if required.",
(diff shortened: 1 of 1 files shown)


wiserain:
URL format seems to be specific for certain method. Can I ask which tool you used for proxy?

nickhoo555:
I used a Cloudflare Worker as the reverse proxy. Here's the basic Worker code:

```javascript
addEventListener('fetch', event => {
  event.respondWith(handleRequest(event.request))
})

async function handleRequest(request) {
  // Extract the target URL from the path (e.g., /https://example.com/file)
  let url = request.url.substr(8)
  url = decodeURIComponent(url.substr(url.indexOf('/') + 1))
  
  if (url.indexOf('://') === -1) {
    url = 'https://' + url
  }

  // Forward the request with original headers
  const headers = new Headers(request.headers)
  headers.delete('host')
  
  const response = await fetch(url, {
    method: request.method,
    headers: headers,
    body: request.body
  })

  return new Response(response.body, {
    status: response.status,
    headers: response.headers
  })
}
```

The proxy URL format `https://proxy.example.com/https://original-url` is a common pattern used by many CF Worker proxies. This format is simple and widely adopted.

If there are other popular proxy URL formats that should be supported, I'm happy to add them as well.

wiserain:
I like the direction of this PR, but I'm concerned that the name `--pikpak-proxu-url` might be too specific and potentially misleading since it's actually modifying the URL rather than acting as a network proxy.

How about a template approach? For example, an option like `download_url_format`.
- If it contains a placeholder like `{url}` or `%s`, rclone replaces it with the actual download link.
- If not, it can just behave like the current implementation.

This makes the feature more general (for mirrors, rewrites, or proxies) without the ambiguity of the word 'proxy'.

PR #9147, opened 2026-02-04
feat: add csvrun command and fix box backend issues

jacquesvandermerwe:
## Summary
This PR introduces a new `csvrun` command for performing bulk copy/move operations defined in a CSV file. It also includes fixes for the Box backend and a PowerShell helper script.

### Changes
- **New Command (`rclone csvrun`)**: 
  - Reads a CSV file with columns: `SourceRemote`, `SourcePath`, `TargetRemote`, `TargetPath`, `[Mode]`.
  - Supports `copy` and `move` operations.
  - Optimizes performance by reusing `Fs` instances (connections) for multiple files on the same remote.
  - Supports parallel execution via `--transfers`.
- **Box Backend Fix**:
  - Fixed an issue in `timingReader` where `fetchDuration` was incorrectly calculated because EOF wasn't always signaled. This ensures accurate bottleneck analysis in reports.
- **Script Fix**:
  - `Get-OneDriveIDs.ps1`: Replaced hardcoded remote name with a parameter (defaults to `mysharepoint`).
- **Internal**:
  - Exported `MoveOrCopyFile` in `fs/operations` to allow use by `csvrun`.

## Test Plan
- Verified `csvrun` compiles and runs.
- Verified `timingReader` fix with unit tests.
- Generated documentation using `gendocs`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Get-OneDriveIDs.ps1
+++ b/Get-OneDriveIDs.ps1
@@ -1,5 +1,7 @@
 # --- CONFIGURATION ---
-$RemoteName = "BC_EMEA_SPO"  # Name of your rclone remote
+param(
+    [string]$RemoteName = "mysharepoint"
+)
 # ---------------------
 
 # 1. Get the Access Token from rclone config
--- a/backend/box/box.go
+++ b/backend/box/box.go
@@ -949,15 +949,17 @@ type timingReader struct {
 	io.Reader
 	ReadDone chan time.Time
 	once     sync.Once
+	lastRead time.Time
 }
 
 func (tr *timingReader) Read(p []byte) (n int, err error) {
 	n, err = tr.Reader.Read(p)
+	tr.lastRead = time.Now()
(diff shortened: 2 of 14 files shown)


ncw:
> Optimizes performance by reusing Fs instances (connections) for multiple files on the same remote.

The approved way of doing this is to use `rclone rcd` then use `rclone rc` to issue commands.

This PR then becomes a very simple script to read the CSV and issue the `rclone rc` commands.

Or with a slightly more complicated script, you could read the CSV and issue a [batch rc command](https://rclone.org/rc/#job-batch)  with will control the concurrency for you.

So I don't think this needs to be part of rclone itself.

If you want to do reporting on file transfers then use the logger flags: https://rclone.org/commands/rclone_copy/#logger-flags - no need to patch the backend.

Issue #2350, opened 2018-06-13
rclone to Qnap Webdav fails with subfolders that contain multiple files.


reyes136:
#### What is the problem you are having with rclone?
When trying to transfer files to a qnap nas with webdav access setup the transfer fails if the folder i'm trying to transfer has another folder in it with more then 1 file. Also any encrypted transfers fail.

If I copy over 1 folder that contains 1 folder and 1 file it transfers ok.
Also I can use cyberduck to transfer folders with as many sub folders and files with not issues and I can even create an encrypted vault and transfer files to that. So I beleive that the Qnap Webdav is set up correctly.

#### What is your rclone version (eg output from `rclone -V`)

```
rclone v1.41-101-gf9a1a7e7
- os/arch: windows/amd64
- go version: go1.10
```

#### Which OS you are using and how many bits (eg Windows 7, 64 bit)
Windows 10 64 bit

####  Which cloud storage system are you using? (eg Google Drive)
Qnap TS-451+

#### The command you were trying to run (eg `rclone copy /tmp remote:tmp`)
rclone>rclone copy C:\Users\Enlight\Desktop\test_folder1 qnap: -vv

#### A log from the command with the `-vv` flag (eg output from `rclone -vv copy /tmp remote:tmp`)

```
c:\Program Files\rclone>rclone copy C:\Users\Enlight\Desktop\test_folder1 qnap: -vv
2018/06/12 22:10:47 DEBUG : rclone: Version "v1.41-101-gf9a1a7e7" starting with parameters ["rclone" "copy" "C:\\Users\\Enlight\\Desktop\\test_folder1" "qnap:" "-vv"]
2018/06/12 22:10:47 DEBUG : Using config file from "C:\\Users\\Enlight\\.config\\rclone\\rclone.conf"
2018/06/12 22:10:48 INFO  : webdav root '': Waiting for checks to finish
2018/06/12 22:10:48 INFO  : webdav root '': Waiting for transfers to finish
2018/06/12 22:10:48 ERROR : test1/015_10.jpg: Failed to copy: Update mkParentDir failed: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>403 Forbidden</title>
</head><body>
<h1>Forbidden</h1>
<p>You don't have permission to access /enlight/test1/
on this server.<br />
</p>
<hr>
<address>Apache Server at hurtlocker.myqnapcloud.com Port 5011</address>
[...]

PR #9152, opened 2026-02-07
WebDAV: Permit redirects on PROPFIND for metadata

bbockelm:

The WebDAV implementation already permits redirects on PROPFIND for listing paths in the `listAll` method but does not permit this for metadata in `readMetaDataForPath`.  This results in a strange experience for endpoints that heavily use redirects -

```
rclone lsl endpoint:
```

functions and lists `hello_world.txt` in its output but

```
rclone lsl endpoint:hello_world.txt
```

Fails with a HTTP 307.

The git history for this setting indicates this was done to avoid an issue where redirects cause a verb change to GET in the Go HTTP client; it does not appear to be problematic with HTTP 307 (issue #2350, solved by @ncw in https://github.com/rclone/rclone/commit/1550f70865245f917c63330039b18876bb214c8f).

No - but it seems trivial.  Would be happy to open an issue if deeper discussion is desired

#### Checklist

I would appreciate guidance if / what tests and documentation are needed.

### DIFF
--- a/backend/webdav/webdav.go
+++ b/backend/webdav/webdav.go
@@ -348,7 +348,7 @@ func (f *Fs) readMetaDataForPath(ctx context.Context, path string, depth string)
 		ExtraHeaders: map[string]string{
 			"Depth": depth,
 		},
-		NoRedirect: true,
+		CheckRedirect: rest.PreserveMethodRedirectFn,
 	}
 	if f.hasOCMD5 || f.hasOCSHA1 {
 		opts.Body = bytes.NewBuffer(owncloudProps)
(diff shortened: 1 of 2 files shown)


ncw:
This will break #2350 so it would be nice to find a fix which doesn't.

A 307 redirect is correct and the verb remains unchanged, but the qnap uses 301 which does cause Go to change the verb. This behaviour is annoying but apparently RFC dictated.

It might be better to implement the CheckRedirect function in the opt and change the verb back there.

If you need more of a hint, I'll find some code when I'm at my laptop, but searching the code for CheckRedirect should give you examples.

Maybe this function should go in lib/rest so we have a standard way of overriding the verb change on http redirects when needed.

bbockelm:
That approach sounds fine to me!

However, this is all sounding a bit fiddly - is there a good example of an integration test that I can build this into?

bbockelm:
@ncw - apologies for being bothersome but is there any way I can help here?  I think I was able to tweak the approach correctly.

PR #9173, opened 2026-02-12
s3: add Object Lock support

### LINKED ISSUE #4683: Add support for S3-compatible object-lock properties

chris081519-crypto:

Add comprehensive S3 Object Lock support with the following new options:

- `--s3-object-lock-mode`: Set Object Lock mode (GOVERNANCE/COMPLIANCE/copy)
- `--s3-object-lock-retain-until-date`: Set retention date (RFC3339/duration/copy)
- `--s3-object-lock-legal-hold-status`: Set legal hold (ON/OFF/copy)
- `--s3-bypass-governance-retention`: Bypass governance mode for deletions
- `--s3-bucket-object-lock-enabled`: Enable Object Lock when creating buckets
- `--s3-object-lock-set-after-upload`: Set retention via separate API calls

The special value `copy` preserves the source object's setting when used with `--metadata`, enabling scenarios like cloning from COMPLIANCE to GOVERNANCE while keeping the original retention date.

Fixes #4683
See also #7894, #7893, #8866

#### Checklist

- [ x] I have read the [contribution guidelines](https://github.com/rclone/rclone/blob/master/CONTRIBUTING.md#submitting-a-new-feature-or-bug-fix).


ncw (commented):
This looks very nice, thank you. Nice metadata integration.

Do you think you could make an integration test? 

They go in here

https://github.com/rclone/rclone/blob/33859568d6793684493db80bbe861c0f2bac5ed2/backend/s3/s3_internal_test.go#L501

The test needs to skip on S3 providers which don't support object lock. Not sure how to achieve that. If we can't then we could skip on provider which isn't ideal, or better add a new quirk to say whether object lock is supported or not.

Thanks

Nick

chris081519-crypto:
Hi Nick,
Thanks for the review! I've added integration tests in `s3_internal_test.go` (`InternalTestObjectLock`).

**Skip detection:** The test creates a temporary bucket with `ObjectLockEnabledForBucket: true`. If that fails, the test skips. If it succeeds, a `GetObjectLockConfiguration` call verifies that Object Lock is actually functional — this catches S3-compatible servers (like `rclone serve s3`) that silently accept the flag without implementing Object Lock.

I chose this approach (creating a new bucket) rather than enabling Object Lock on the existing test bucket (like `InternalTestVersions` does with versioning), because enabling Object Lock on existing buckets wasn't supported by most providers for a long time — and rclone itself doesn't do that either.

**Test coverage:**
- **Retention** — upload with GOVERNANCE mode + retain-until-date, verify via metadata
- **LegalHold** — upload with legal hold ON, toggle to OFF via API, verify round-trip
- **SetAfterUpload** — lock applied post-upload via `PutObjectRetention` / `PutObjectLegalHold`
- **BypassGovernanceRetention** — used during cleanup to delete governance-locked objects

The temporary bucket (with all versions) is cleaned up after the test.

Tested locally against MinIO and SeaweedFS — passes on both.

If you'd like the tests structured differently or have any other feedback, happy to adjust!

Regards,
Chris

chris081519-crypto:
Had to adjust my previous comment, as while testing minio and seaweed, I missed testing against rclone serve s3. CI should pass now.

PR #9182, opened 2026-02-14
googlephotos: implement moveToTrash functionality for Google Photos

### LINKED ISSUE #7956: Google Photos album for deleted media
I was disappointed to see that the Google Photos API does not support media removal. Rclone does the best it can by removing the media items from the albums, but the media is still kept around without any associated album, counting against the storage quota.

Unfortunately, this makes it very hard to find dangling media that should have been removed, as I have to manually delete them in the web UI.

I propose adding the removed media to a "Trash" album to make it easier to find and delete them. This could be an advanced configuration of the Google Photos remote, with the album name configurable to avoid collisions with existing albums.

Additionally, the same concept could be applied to removed albums, as the API does not support album removal either.

Cprakhar:
This PR implements a new feature to delete and move files from albums to "Rclone Trash" album for deletion of media from everywhere. As the new Google Photos API doesn't support deletion from media, as discussed in #7956 and [here](https://rclone.org/googlephotos/#deleting-files) 

Closes #7956 

#### Checklist


backend/googlephotos/googlephotos.go
```diff
@@ -1237,6 +1238,42 @@ func (o *Object) Remove(ctx context.Context) (err error) {
...
+	err = o.moveToTrash(ctx, fileName)
+	if err != nil {
+		return fmt.Errorf("couldn't move item to trash: %w", err)
+	}
+	return nil
+}
+
+// MoveToTrash moves the object to the trash album
+func (o *Object) moveToTrash(ctx context.Context, fileName string) (err error) {
+	trashAlbumPath := path.Join("album", trashAlbum)
```
ncw: When `root = "album/MyAlbum"` (user pointed rclone at a specific album), this sets `trashAlbumPath = "Rclone Trash"`.

Then Mkdir is called, which does `absPath = path.Join("album/MyAlbum", "Rclone Trash") =  album/MyAlbum/Rclone Trash"`.

This matches `^album/(.+)$` with `match[1] = "MyAlbum/Rclone Trash"`, so it creates an album literally named `"MyAlbum/Rclone Trash"` instead of `"Rclone Trash"`. The same problem affects the Put call - the file gets uploaded to the wrong album.

backend/googlephotos/googlephotos.go
```diff
@@ -1237,6 +1238,42 @@ func (o *Object) Remove(ctx context.Context) (err error) {
...
+		trashAlbumPath = trashAlbum
+	}
+
+	err = o.fs.Mkdir(ctx, trashAlbumPath)
+	if err != nil {
+		return err
+	}
+
+	remote := path.Join(trashAlbumPath, fileName)
+	in, err := o.Open(ctx)
```
ncw: This downloads and re-uploads every single byte of every deleted photo/video over the network!

The Google Photos API has `albums/:batchAddMediaItems` which can add an existing media item to a new album by ID, without any download/upload. This would be instant and use zero bandwidth. The media item still exists in the library after `batchRemoveMediaItems` - it's just unlinked from the album.

Add to trash first, then remove from the source album for safety

backend/googlephotos/googlephotos.go
```diff
@@ -1237,6 +1238,42 @@ func (o *Object) Remove(ctx context.Context) (err error) {
...
+	if err != nil {
+		return fmt.Errorf("couldn't move item to trash: %w", err)
+	}
+	return nil
+}
+
+// MoveToTrash moves the object to the trash album
+func (o *Object) moveToTrash(ctx context.Context, fileName string) (err error) {
+	trashAlbumPath := path.Join("album", trashAlbum)
+	if strings.HasPrefix(o.fs.root, "album") {
```
ncw: The `strings.HasPrefix(o.fs.root, "album")` check is too broad - it would also match a hypothetical root like "album-something" (unlikely but sloppy). A prefix check of "album/" or an exact match on "album" would be more correct.

backend/googlephotos/googlephotos_test.go
```diff
@@ -173,6 +173,18 @@ func TestIntegration(t *testing.T) {
 				entries, err := f.List(ctx, albumName)
 				require.NoError(t, err)
 				assert.Equal(t, 0, len(entries))
+
+				// Check if transferred to trash
+				entries, err = f.List(ctx, "album/"+trashAlbum)
```
ncw, line 181: The test only checks the case where the root is empty (listing "album/"+trashAlbum). It doesn't test the broken case where root = "album/SomeAlbum", which is the exact scenario that's buggy.

ncw (changes requested):
Thank you for implementing this feature

I think this feature should be configurable as it represents a change in behaviour for rclone which we try to avoid.

  - A config option like `use_trash = true/false` (default `false`) to opt in would be perfect
  - The option's Help field should explain that it copies removed items to an "Rclone Trash" album before removing them from the source album
  - Possibly a configurable trash album name too, though that's lower priority

See inline for some other comments on the implementation

Thank you :-)

Cprakhar:
@ncw, I have incorporated your suggestions. If you have time, take a look.

Cprakhar:
@ncw Any plan to merge it?

PR #9189, opened 2026-02-17
Update Proton Drive app-version header to rclone specific one

dlaumen:
Hi

I'm an engineer in the Proton Drive team. We would like to ask you to make a minor change in the way you interact with the API.

1. Change the domain to drive-api.proton.me
    - There are changes planned to the infrastructure so that Drive API endpoints are only available on the drive-api host, this'll ensure your client will keep working for you and users of this client in the future.
    - See https://github.com/rclone/go-proton-api/pull/1
2. Change the app-version header
    -  We have added support for third-party app-version headers: `external-drive-<project>@<semver-version>`, e.g. `external-drive-rclone@1.0.0`
    - The project name must conform to `-[a-z_]+`
    - The version can be any semver-compliant version number, you can inject your own version numbering if you want
    - This ensures that this client remains functional even if we perform a force-upgrade of one of our official clients, which would prevent a specific app-version from using the API
    - This would also help us with our metrics, as they currently report your client as the official MacOS client, which causes problems for our team when debugging because errors from the official client get mixed in with errors from this project

We would be very grateful if you could adopt these changes, even though we cannot offer more support for third parties at this moment.

### DIFF
--- a/docs/content/drive.md
+++ b/docs/content/drive.md
@@ -2008,6 +2008,9 @@ Scroll down and click "+ Add users". Add yourself as a test user and press save.
 
 10. Provide the noted client ID and client secret to rclone.
 
+11. Run the web-based authorization flow from within `rclone config`, by answering
+    "Y" when it asks "Already have a token - refresh?".
+
 Be aware that, due to the "enhanced security" recently introduced by
 Google, you are theoretically expected to "submit your app for verification"
 and then wait a few weeks(!) for their response; in practice, you can go right


ncw:
Hi @dlaumen , thanks for getting in touch.

We always try to work with the vendors so I don't see any problem with merging this.

You have `external-drive-rclone@1.0.0` in your identifier string.

Do you want the current version of rclone? This is in `fs.Version` eg `v1.74.0-beta.9474.f8e1d74b9` or `v1.73.0` which should be semver compliant (perhaps without the leading `v`)?

dlaumen:
Thank you for the followup @ncw !

I didn't know how to dynamically get a version number, and for us it's currently mostly important to be able to distinguish official versus 3rd party clients.

I did just double-checked against our parser for this version, but unfortunately that won't work. It's apparently slightly stricter than what semver allows in regards to the pre-release identifiers, we only allow integers. We do support a +build

The regex for the whole version string:
```
/^(external-drive)+(-[a-z_]+)+@[0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?-((stable|beta|RC|alpha)(([.-]?\d+)*)?)?([.-]?dev)?(\+.*)?$/i
```

So this could work: `external-drive-rclone@1.74.0-beta.9474+f8e1d74b9` or simply `external-drive-rclone@1.74.0-beta.9474` or `external-drive-rclone@1.74.0-beta`. For us it doesn't matter that much how detailed you send a version number.

Am I also correct in assuming this app_version is configurable/overridable by the user? 
We currently have some workarounds/exceptions in place on the API for rclone to avoid breaking changes on uploads, but if a user sets a custom app_version and we can no longer detect the request comes from rclone then the upload will fail. We will keep this exception in place for rclone until after the Proton Drive SDK is deemed stable and 3rd parties had a chance to implement it.  So it might maybe be better if the user cannot override this and break uploads without knowing/understanding why?

PR #9190, opened 2026-02-18
internxt: fix Entry doesn't belong in directory errors on windows

jzunigax2:
Fix directory listing for the internxt backend on windows

https://forum.rclone.org/t/internxt-backend-returns-backslash-paths-for-subdirectories-entry-doesnt-belong-in-directory-too-short/53379

#### Checklist

### DIFF
--- a/backend/internxt/internxt.go
+++ b/backend/internxt/internxt.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net"
 	"path"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


jzunigax2:
> That looks like a good fix to me. Does it work?

hey @ncw, it does work however I did not notice an issue on integration tests. The TestTransform subtests are now consistently failling originally I wasn't sure if it was related to these changes but just tried on master without this commit and got the same failure. 

I now believe the pacer short-circuit in [8b85ffbf0](https://github.com/rclone/rclone/commit/8b85ffbf033cff39f1c861eb3ee15b2f691f6685) (skipping the token when sleepTime=0) likely exposed this behavior. Previously the implicit serialization through the pacer token spaced out requests enough to mask an eventual consistency issue. 

I'll address this, I see some backends like onedrive have an arbitrary sleep for similar eventual consistency issues. In this case, folder creation returns a valid UUID but the folder isn't immediately available for file metadata creation, so concurrent uploads to a new subdirectory all hit "Folder not found (404)"

ncw:
@jzunigax2 Yes I think you are right - there are a lot of eventual consistency type failures in the Internxt backend integration tests at the moment.

It would be great if you could fix them, but feel free to take this fix out of draft and I can merge if you want?

jzunigax2:
@ncw good news, bumped internxt/rclone-adapter version in this PR now integration tests should be passing

ncw:
I've merged this to master now which means it will be in [the latest beta](https://beta.rclone.org/) in 15-30 minutes and released in v1.74  and in v1.73.2 (I think this is suitable for a point release - would you agree?)

jzunigax2:
great, thanks! and yeah I agree this would suit a point release

